PR #95, opened 2016-06-30
Use Promise for Commands

peppsac:
Commands are async, so Promise are a good fit for this usage.

It will ease the transition to remove logic code from Providers.
Simplified example:

``` js
    // calling site
    interCommand.request(args, node);

    // provider
    Provider.prototype.executeCommand = function(command){
        return this.getColorTextures(tile, services).then(function(result)
        {
            this.setTexturesLayer(result, destination);
        }.bind(tile));
```

Would become:

``` js
    // calling site
    interCommand.request(args, node).then(function(textures) {
          node.setTexturesLayer(result,destination);
    });

    // provider
    Provider.prototype.executeCommand = function(command){
        return this.getColorTextures(tile, services).then(function(result)
        {
            command.resolve(result);
        };
```

### DIFF
--- a/src/Core/Commander/InterfaceCommander.js
+++ b/src/Core/Commander/InterfaceCommander.js
@@ -4,7 +4,7 @@
  * Description: Cette Classe construit une commande. Cette Command ensuite pousser dans une file d'attente.
  */
 
-define('Core/Commander/InterfaceCommander', ['Core/Commander/ManagerCommands', 'Core/Commander/Command'], function(ManagerCommands, Command) {
+define('Core/Commander/InterfaceCommander', ['Core/Commander/ManagerCommands', 'Core/Commander/Command', 'when'], function(ManagerCommands, Command, when) {
 
     function InterfaceCommander(type) {
         //Constructor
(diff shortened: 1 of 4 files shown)


peppsac:
I added a 2nd commit that modifies WMTS_Provider and TileProvider to take advantages of these Promise.

qdnguyen:
Cool but if you can add web workers in this module to parallelize tasks, it is even better!

peppsac:
> Cool but if you can add web workers in this module to parallelize tasks, it is even better!

For now this PR is focused on getting the groundwork done so we can improve the code later -and yes Web Workers would be nice to have :-)

gcebelieu:
Be careful, promise is ES6 and doesn't seem to be supported by IE11 (see : http://caniuse.com/#feat=promises).
To add this support, a promise polyfill (https://github.com/lahmatiy/es6-promise-polyfill) should be added.

tbroyer:
@gcebelieu This is using `when` promises, not native promises. But IMO eventually promises should be moved to native ones with a polyfill.

BTW, https://github.com/stefanpenner/es6-promise is better maintained and seems to be the go-to polyfill.

peppsac:
See https://github.com/iTowns/itowns2/issues/97

gcebelieu:
thanks, this sounds good to me :)

PR #99, opened 2016-07-04
Rework layer params

peppsac:
1st commit fixes a typo (except if it's not a typo, in case it needs a comment)

2nd commit commit message :

```
 Rework color layers parameters handling

 Parameters are only initialized once: when the tile has not texture.
 So let's init these param in the TileProvider, rather than hiding it
 in the first getColorTextures call.

 TileMesh.setParamsColor method has been renamed setColorLayerParameters.
```

### DIFF
--- a/src/Core/Commander/Providers/TileProvider.js
+++ b/src/Core/Commander/Providers/TileProvider.js
@@ -145,18 +145,34 @@ define('Core/Commander/Providers/TileProvider', [
             tile.WMTSs = [];
 
             // TODO passer mes layers colors?
-            var paramsColor = [];
-
+            var paramMaterial = [];
+            var textureCount = 0;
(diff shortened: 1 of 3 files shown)


peppsac:
One open question though: what is the `pit` parameter : 
https://github.com/iTowns/itowns2/pull/99/files#diff-ca95311157eb5bcc4cfca16f08efb160R163

peppsac:
Another reason to move these out of the WMTS specific code is because these parameters are exclusively used by the fragment shader:

``` glsl
uniform vec4        paramLayers[8];
uniform vec2        paramBLayers[8];
```

That makes them rendering parameters and as such they are probably useful for other image providers (like WMS).
If they end up being specific to WMTS then there's some refactoring to do (LayeredMaterial? shaders? Providers?) to be able to use any kind of Provider.

PR #100, opened 2016-07-04
Replace when.js with native promises and a polyfill

### LINKED ISSUE #97: Use a polyfill for Promise
Why: we use Promise extensively but not all browsers support this feature.

https://github.com/stefanpenner/es6-promise is a good candidate

See issue #93 and #95 for more context.

tbroyer:
Tested non-regression in Chrome (51/Linux), Firefox (49/Dev Edition/Linux) and Edge (13).
Couldn't actually test that it works in IE11 (the only browser actually needing the polyfill) due to other errors.

Note: "bundles" the polyfill in the library for now; we may want to revisit this in the future.

Close #97 

r: @peppsac


peppsac:
Looks good.

A few places still reference `when.js`:
- packages.json
- LICENSE.md (and maybe you could add an entry from es6-promise here as well?)
- ManagerCommands comment (line 11)
- Ori.js
- ProjectiveTexturingMaterial.js
- KMZLoader

tbroyer:
Totally removed `when.js`, and also refactored a few things (all related to promises). Tell me if you'd prefer them to be split in separate PRs.

Note: I amended my first commit to get a clean history.

src/Core/Commander/Providers/PanoramicProvider.js
```diff
@@ -72,24 +70,18 @@
      * @param {type} distance
      * @returns {Promise}
      */
-    PanoramicProvider.prototype.getMetaDataFromPos = function(longitude, latitude, distance){
-
-        var that = this;
-        if(!_panoramicsMetaData){
+    PanoramicProvider.prototype.getMetaDataFromPos = function(longitude, latitude/*, distance*/){
```
tbroyer: Wasn't sure what to do here: entirely remove the unused `distance` parameter or just comment it out?

FWIW, it was passed to `getClosestPanoInMemory` which didn't _expect_ it; and I inlined `getClosestPanoInMemory` here in the promise's `then`.

nosy-b: Yes it was originally set to ensure a max distance. Doesn't matter anymore

tbroyer: OK, removed. (and directly amended the commit, given the nature and scope of the change)

peppsac:
r+, merging.

Thanks !

PR #94, opened 2016-06-30
Add earlyDropFunction attribute to Commands

peppsac:
Allow the user to specify a function to decide if a Command
must be cancelled.

Note that children creation commands are tagged as non-cancellable.
That's because children creation only happens when a node has no children.

If 1 of the child creation request is cancelled the node
will be in an invalid state: 3 children and no way to create
the last one.

Fix #82 'Tiles are not displayed'

@gchoqueux r?

PR #101, opened 2016-07-05
Cleanup commands and quadtree

peppsac:
Note 1: first commit is from PR #94 
Note 2: do not merge yet as this introduces a regression when zooming quickly (tile blinking)

### DIFF
--- a/src/Scene/Quadtree.js
+++ b/src/Scene/Quadtree.js
@@ -19,10 +19,26 @@ define('Scene/Quadtree', [
 
 
     function commandQueuePriorityFunction(cmd) {
-        var requester = cmd.requester;
-        return requester.sse ?
-            Math.floor(requester.isVisible() ? requester.sse * requester.sse * 100000 : 1.0) :
-            1.0;
+        var node = cmd.requester;
+
+        // We know that 'node' is visible because commands can only be
+        // issued for visible nodes.
+        //
+        // Prioritize subdivision request
+        if (cmd.layer instanceof Quadtree) {
+            return 10000;
+        } else {
+            if (!node.loaded) {
+                return 1000;
+            } else {
+                // TODO: this magic value comes from NodeProcess
+                if (6.0 < node.sse) {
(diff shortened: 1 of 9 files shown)


peppsac:
Rebased on master.

src/Scene/Quadtree.js
```diff
@@ -12,14 +12,39 @@
...
     'Scene/Layer',
+    'Core/Commander/InterfaceCommander',
     'Core/Geographic/Quad',
     'Renderer/NodeMesh'
-], function(Layer, Quad, NodeMesh) {
-
+], function(Layer, InterfaceCommander, Quad, NodeMesh) {
+
+
+    function commandQueuePriorityFunction(cmd) {
```
Jeremy-Gaillard, line 21: I would move this function to NodeProcess since the priority of a command depends more on the nature of the node than the structure in which it is.
Edit: just noticed this would also require passing the priority value directly to the interfaceCommander instead of using a function

gchoqueux:
41e2d96447c2013390a117acc3551c580f010801, the policy priorities needs thinking.
There's not trivial solution and there are lot of parameters : subdivision, request texture, visibility, point of interest, anticipation of movement. Ideally, a user can customize this priorities (like c5db57ad0cc18bc31d2235300fd3d3a85efef05e). I add issue  #136

src/Scene/Quadtree.js
peppsac, line 21: Makes sense but I would keep it in `Quadtree` until we have another implementation (Octree maybe?) that uses NodeProcess and this priority function.
WDYT?

peppsac:
> the policy priorities needs thinking.

Agreed. The changes proposed here are a first step in this direction.

gchoqueux:
r- remove regression when zooming quickly (tile blinking)

peppsac:
@gchoqueux blinking issue is fixed (see last commit). r?

gchoqueux:
r+

PR #126, opened 2016-07-06
added note in doc for debian users with npm 1.4

vmora:
check npm version is >2.0

otherwise the `npm start -- --port 3000` won't work


peppsac:
Looks good to me. Merged.

PR #44, opened 2016-02-17
Move from AMD and CommonJS to ES6 modules

tbroyer:
This is an (almost) automated conversion:
- first "unname" the AMD modules,
  
  ```
  sed -i "s|define.'[^']*',|define(|g" $(git grep --name-only  -e "define[(]'[^']*',")
  ```
- then convert them to ES6 modules through amdtoes6 (use `npm install amd-to-es6` to install the tool).
  
  ```
  amdtoes6 -d src/ -o src/ -b
  ```
- finally, convert CommonJS modules to ES6 modules through esnext (use `npm install esnext` to install the tool).
  
  ```
  find src/ -type f -name "*.js" \
      -not -name "*.min.js" -not -name "optimer_regular.js" \
      -exec esnext -I -w modules.commonjs {} \;
  ```

Issue #122, opened 2016-07-05
NPM package iTowns


gchoqueux:
Make npm package for iTowns

tbroyer:
Prerequisite: either move to commonjs (might not be enough though), or update our "bundling" policy and/or move our dependencies as devDependencies (if we bundle them in a UMD module, like we do today, then we no longer need to declare them as dependencies, and should no longer do it).

autra:
I guess this is superseded by https://github.com/iTowns/itowns2/issues/243

PR #131, opened 2016-07-06
Amd and commonjs to es6

vmora:
This is #44 again.

### DIFF
--- a/doc/API_Doc/ApiGlobe.js.html
+++ b/doc/API_Doc/ApiGlobe.js.html
@@ -33,7 +33,7 @@
  */
 
 
-define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
+define( [
        'Core/Commander/Interfaces/EventsManager',
        'Scene/Scene',
        'Scene/NodeProcess',
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\"",
     "test": "npm run lint && npm run build",
     "build": "webpack -p",
-    "start": "webpack-dev-server -d --inline --hot"
+    "start": "webpack-dev-server -d --inline --hot",
+    "prepublish": "babel -d dist/ src/"
   },
   "repository": {
     "type": "git",
@@ -29,6 +30,12 @@
     "simd": "^2.0.0"
   },
   "devDependencies": {
+    "babel-cli": "^6.5.1",
+    "babel-core": "^6.5.1",
+    "babel-loader": "^6.2.2",
+    "babel-plugin-transform-runtime": "^6.5.0",
+    "babel-preset-es2015": "^6.5.0",
+    "babel-runtime": "^6.5.0",
     "eslint": "^2.5.3",
     "eslint-loader": "^1.3.0",
     "js-beautify": "^1.6.2",
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["es2015"],
+  "plugins": ["transform-runtime"]
+}
--- a/src/Core/Commander/Command.js
+++ b/src/Core/Commander/Command.js
@@ -4,38 +4,34 @@
  * Description: Cette object contient une commande à executer. Elle porte également les buffers résultants.
  */
 
-define('Core/Commander/Command', [], function() {
+function Command() {
+    //Constructor
 
+    this.name = null;
+    this.priority = 0.0; //Math.floor((Math.random()*100));
+    this.state = null;
+    this.inParallel = null;
+    this.inBuffers = null;
+    this.outBuffers = null;
+    this.paramsFunction = {};
+    this.processFunction = null;
+    this.async = null;
+    this.force = null;
+    this.type = null;
+    this.addInHistory = null;
+    this.source = null;
+    this.requester = null;
+    this.provider = null;
 
-    function Command() {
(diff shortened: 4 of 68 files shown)


doc/API_Doc/ApiGlobe.js.html
```diff
@@ -33,7 +33,7 @@ <h1 class="page-title">Source: ApiGlobe.js</h1>
  */
 
 
-define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
+define( [
```
peppsac, line 36: Is this `define` needed?

tbroyer, line 36: This is an artifact of the `sed` script being used to transform "named defines" to "unnamed defines" prior to running `amdtoes6` (which doesn't support named defines). The `sed` script modified that file, but `amdtoes6` obviously didn't.

IMO, either we:
- ignore it, leave the commit like that; we'll eventually remove the generated doc from `master` anyway
- remove `doc` from `master` first, then rebase that change on top
- exclude that file from the commit

package.json
```diff
@@ -7,7 +7,8 @@
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\"",
     "test": "npm run lint && npm run build",
     "build": "webpack -p",
-    "start": "webpack-dev-server -d --inline --hot"
+    "start": "webpack-dev-server -d --inline --hot",
+    "prepublish": "babel -d dist/ src/"
```
tbroyer, line 11: This will transform ES6 sources to ES5 with CommonJS imports, so it's a "legal" NPM module (see #122); it should target `lib/` then, not `dist/` (which contains the output from `webpack`).

`main` above should then be updated to point to `lib/Main.js`.

And `.npmignore` should ignore `dist/` (we'll publish the ES5 sources to NPM, not the output from `webpack`), and `.gitignore` should also ignore `lib/`.

(see #44 where this was all done)

tbroyer, line 11: An alternative would be to use CommonJS for now and stick with ES5: no need for that `prepublish` calling `babel`, just update the `main` to point to `src/Main.js` and add `dist/` to `.npmignore`.

vmora:
There is a need for decision here, since we have mixed AMD and commonjs imports. 

Here's my vote:
1) es6 import, since it's the fututre
2) commonjs, good enought and easy to transform afterward
3) AMD, last choice for me since it's error prone (2 lists that should be in the same order... but sometimes aren't)

ping @iTowns/core-devs

package.json
tbroyer, line 11: Thinking more about it: we don't follow NPM / Node.js rules for imports (non-relative names –i.e. those not starting with `./` or `../`– are resolved against `node_modules` sibling folders walking up the hierarchy; so `require("Core/Commander/Command")` wouldn't work in "pure Node.js", it'd have to be either `require("itowns2/lib/Core/Command/Command")` to resolve to the file from an _installed_ `itowns2` NPM module; or `require("./Command")` –or `import Command from './Command'`– when called from our `InterfaceCommander`). See http://www.commonjs.org/specs/modules/1.0/ and https://nodejs.org/api/modules.html for the details.

Our current usage of referencing files by their _absolute_ path “relative” to our `src` root only works because we configured Webpack that way, but it won't/wouldn't work if we publish our files to NPM and try to use the module from another project (if using Webpack or Rollup –and probably also Browserify–, one could configure it to make it work, but that's additional burden).

So I think we should just remove that `prepublish` script, and revisit things when we want to tackle the NPM issue (see #44).  
(FWIW, I also think we should move to using relative import paths, like everyone else in the JS/Node.js land, but again, that's another matter that should be addressed in a distinct PR)

package.json
```diff
@@ -29,6 +30,12 @@
     "simd": "^2.0.0"
   },
   "devDependencies": {
+    "babel-cli": "^6.5.1",
```
tbroyer, line 33: I'm really not sure `babel-cli` is needed.

AFAICT, the `babel` executable is installed as part of `babel-core`, and `babel-cli` is meant to be installed globally (`npm install -g babel-cli`) so it's available on your command-line anywhere, and will "sniff" the current module to actually call the locally-installed `babel` (at `node_modules/.bin/babel`) if it exists; defaulting to the globally-installed one otherwise.

vpicavet:
May be related : https://github.com/mrdoob/three.js/pull/9310

peppsac:
> May be related : mrdoob/three.js#9310

This has been merged. I think it's a great example to follow for itowns.

I'm going to work on this done soon, unless:
- someone thinks it's a bad ida.
- someone else want to do it :)

@iTowns/reviewers : thoughts?

tbroyer:
I just updated #44 with code that appears to work (using `npm start`)

I removed everything somehow related to publishing the module on the NPM registry (e.g. transforming ES6 to ES5 before publishing)

vmora:
> thoughts?

+1

peppsac:
@tbroyer PR #44 's branch seems outdated?

tbroyer:
Hmm, [`amdtoes6` branch](https://github.com/iTowns/itowns2/tree/amdtoes6) has been updated, but the PR being closed it wasn't updated. Let me create a new PR.

tbroyer:
#169

peppsac:
PR #169 is now merged, we can close this one - thanks @vmora and @vpicavet for your help :)

vmora:
@peppsac, @tbroyer thanks.

PR #132, opened 2016-07-07
Rework UV shader parameters

peppsac:
Using descriptive names for uv/uv1 helps understanding the rendering
architecture.


peppsac:
This PR is now ready.

@gchoqueux r?

gchoqueux:
The UVs's tile contain geographic projection. Currently, there are 2 projections : PM (Pseudo Mercator) and WGS84. The geometry vertex use WGS84. Ideally, the tile geometry shouldn't have notions of geographic projections but only generic UVs. The tile should be build according to the required projection systems. 

I add new issue #138 

r+

PR #139, opened 2016-07-07
Test on Windows through AppVeyor, and use NPM prepublishOnly rather than prepublish

tbroyer:
Actually, I'm not really sure about this, because AppVeyor has a really strange concept of "account" vs "user".
The checks currently are at https://ci.appveyor.com/project/tbroyer/itowns2/build/1.0.1 (notice that it's linked to my account, not to some sort of organization). I can rename my "account" to "iTowns", and according to their documentations I (the "user")could be a "member" of another, personal, "account"; but I'm really not sure they'd let me **create** that other account.

So, well, this PR adds AppVeyor configuration, and that's what's important.  
But ultimately maybe someone who only intends to use AppVeyor for iTowns should sign-up there, rename his account to `iTowns` and create the job.  
In the mean time, I'm OK with having itowns2 under my personal AppVeyor account (as I intend to eventually use AppVeyor for other projects, I'm not going to rename my account), so if everyone's OK with seeing my name in the AppVeyor URLs, then that's OK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "build": "webpack -p",
     "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
-    "prepublish": "npm run build && npm run transpile"
+    "prepublishOnly": "npm run build && npm run transpile"
   },
   "files": [
     "*.md",
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,11 @@
+# See https://www.appveyor.com/docs/lang/nodejs-iojs
+install:
+  - ps: Install-Product node stable
+  - node --version
+  - npm --version
+  - npm ci
+
+test_script:
+  - npm test
+
+build: off


tbroyer:
@iTowns/reviewers Should we move this forward? or should I delete the hooks that currently build each push and PR? (and fails early by lack of `appveyor.yml` configuration)

gmaillet:
I'm not sure that AppVeyor is necessary for now if it's only for the building process (with no real tests).

tbroyer:
FWIW:
- we had a bug on Windows wrt builds (6a768b599adfbe405febbc4826feb832b5b9442d & 645087b5fd6b7b48b25ff4360369f671989023e9) so we could at least make sure that people on Windows can still build the project
- we're in the process of adding tests, so testing on Windows would be good. Also, `headless-gl` is said to work out-of-the-box on AppVeyor, contrary to Travis: https://github.com/stackgl/headless-gl#how-can-i-use-headless-gl-with-a-continuous-integration-service, so we could expect WebGL to work on IE11: https://www.appveyor.com/docs/installed-software#selenium

That said, I'm OK with delaying adding AppVeyor until we _do_ have tests!

autra:
@tbroyer do you mind if I close this? It has been open for 2 years, and I'm not sure it's really relevant any more (we currently use pupeeter for browser tests). WDYT?

tbroyer:
Not being a Windows users, I don't have a strong opinion on this.

I however still think it _could_ be useful to have automated feedback from Windows:
 * that developers on Windows still have a working environment
 * that tests pass on Windows

How about first trying to update that PR?

tbroyer:
Fwiw, updated PR ran OK on AppVeyor: https://ci.appveyor.com/project/tbroyer/itowns2/build/1.0.595

package.json
```diff
@@ -13,7 +13,7 @@
     "build": "webpack -p",
     "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
-    "prepublish": "npm run build && npm run transpile"
+    "prepublishOnly": "npm run build && npm run transpile"
```
autra, line 16: From the commit message:

> Note that in CI (Travis and AppVeyor), "npm test" will
run "npm run build", so this scenario is still "tested".

I wholeheartedly support this change, although if I'm not mistaken, the `transpile` part is not run. If we switch to `prepublishOnly`, we should either add `transpile` to `build` or to `test`, or maybe both btw (is there a script name for that?)

autra (commented):
> Fwiw, updated PR ran OK on AppVeyor: https://ci.appveyor.com/project/tbroyer/itowns2/build/1.0.595

Oh so we might be nearer merging that than I initially thought, great!

> In the mean time, I'm OK with having itowns2 under my personal AppVeyor account (as I intend to eventually use AppVeyor for other projects, I'm not going to rename my account), so if everyone's OK with seeing my name in the AppVeyor URLs, then that's OK.

I don't mind. What's the worst that could happen? That you close your account? In this case, we'll just switch I guess. Or maybe AppVeyor *did* introduce something like organization account in the meantime?

Where is the url configured btw?

autra:
also, @tbroyer apparently something is wrong in the AppVeyor repo, because github webhook receives `{"message":"PR and project repositories do not match."}`. Maybe it's missing the renaming itowns -> itowns2 ?

tbroyer:
> I don't mind. What's the worst that could happen? That you close your account? In this case, we'll just switch I guess. Or maybe AppVeyor did introduce something like organization account in the meantime?

They did! https://www.appveyor.com/docs/team-setup/
I added @iTowns/psc as "admin" and @iTowns/reviewers as "user".
Builds will now be at https://ci.appveyor.com/project/iTowns/itowns (only change I made from the default configuration: I checked "ignore branches without `appveyor.yml`")

I also removed the webhook and deleted the project from my personal account.

package.json
tbroyer: Isn't transpiling already done when compiling? I mean, Babel runs as part of the Webpack build; it does not output ES5 to disk, but this is mostly equivalent in all other aspects, right?

autra: I guess it is, you're right. Actually, I thought we needed the `lib/` folder for unit tests, but actually sources are compiled on the fly by mocha apparently. I think I mixed up with the old version of integration tests, before we used puppeeter.

zarov:
It seems to block some PR, why is that ? See for example https://github.com/iTowns/itowns/pull/798

autra (approved):
LGTM! Glad to see this happening.

@tbroyer can you edit your commit messages by prefixing them with `test:` for the first one and `chore:` for the second? We're trying to adopt a convention for them to automatically generate changelogs, see #518.

Thanks again for resurrecting this!

tbroyer:
> can you edit your commit messages

Done

PR #141, opened 2016-07-07
Use requestAnimationFrame for update loop

Jeremy-Gaillard:
Replaced setTimeout by requestAnimationFrame.

Removed calls to scene update (scene.wait()) from managerCommands and c3Dengine and clarified the update mechanism.

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -221,9 +221,7 @@ ApiGlobe.prototype.createSceneGlobe = function(coordCarto, viewerDiv) {
     this.viewerDiv = viewerDiv;
 
     viewerDiv.addEventListener('globe-built', function() {
-
         if (loaded == false) {
-
             loaded = true;
             viewerDiv.dispatchEvent(eventLoaded);
         } else {
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -271,20 +271,11 @@ ApiGlobe.prototype.createSceneGlobe = function(coordCarto, viewerDiv) {
 
     map.tiles.init(map.layersConfiguration.getGeometryLayers()[0]);
 
-    //!\\ TEMP
-    //this.scene.wait(0);
-    //!\\ TEMP
-
     return this.scene;
-
 };
 
 ApiGlobe.prototype.update = function() {
-
-    //!\\ TEMP
-    this.scene.wait(0);
-    //!\\ TEMP
-
+    this.scene.notifyChange();
 };
 
 // ApiGlobe.prototype.setLayerAtLevel = function(baseurl,layer/*,level*/) {
--- a/src/Core/Commander/ManagerCommands.js
+++ b/src/Core/Commander/ManagerCommands.js
@@ -9,28 +9,64 @@ import PriorityQueue from 'js-priority-queue';
 
 var instanceCommandManager = null;
 
+function _instanciateQueue() {
+    return {
+        storage: new PriorityQueue({
+            comparator: function(a, b) {
+                var cmp = b.priority - a.priority;
+                // Prioritize recent commands
+                if (cmp === 0) {
+                    return b.timestamp - a.timestamp;
+                }
+                return cmp;
+            }
+        }),
+        counters: {
+            executing: 0,
+            executed: 0,
+            failed: 0,
+            cancelled: 0
+        },
+        execute: function(cmd, provider) {
+            this.counters.executing++;
+
+            var p = provider.executeCommand(cmd);
+
+            if (!p) throw new Error('arg');
+            return p.then(
+                function() {
+                    this.counters.executing--;
+                    this.counters.executed++;
+                }.bind(this),
+                function() {
+                    this.counters.executing--;
+                    this.counters.executed++;
+                    this.counters.failed++;
+                }.bind(this)
+            );
+        }
+    };
+}
+
 function ManagerCommands(scene) {
     //Constructor
     if (instanceCommandManager !== null) {
         throw new Error("Cannot instantiate more than one ManagerCommands");
     }
 
-    this.queueAsync = new PriorityQueue({
-        comparator: function(a, b) {
-            var cmp = b.priority - a.priority;
-            // Prioritize recent commands
-            if (cmp === 0) {
-                return b.timestamp - a.timestamp;
-            }
-            return cmp;
-        }
-    });
+    this.defaultQueue = _instanciateQueue();
+    this.hostQueues = new Map();
 
-    this.queueSync = null;
-    this.loadQueue = [];
     this.providers = {};
-    this.history = null;
+
     this.eventsManager = new EventsManager();
+    this.maxConcurrentCommands = 16;
+    this.maxCommandsPerHost = 6;
+    this.counters = {
+        runningCommands: 0,
+        executedCommands: 0,
+        hostCommands: {}
+    };
 
     if (!scene)
         throw new Error("Cannot instantiate ManagerCommands without scene");
@@ -41,8 +77,53 @@ function ManagerCommands(scene) {
 
 ManagerCommands.prototype.constructor = ManagerCommands;
 
+ManagerCommands.prototype.runCommand = function(command, queue) {
+    var provider = this.providers[command.layer.protocol];
+
+    if (!provider) {
+        throw new Error('No known provider for layer', command.layer.id);
+    }
+
+    queue.execute(command, provider).then(function() {
+        this.scene.notifyChange();
+
+        // try to execute next command
+        if (queue.counters.executing < this.maxCommandsPerHost) {
+            let cmd = this.deQueue(queue);
+            if (cmd) {
+                return this.runCommand(cmd, queue);
+            }
+        }
+    }.bind(this));
+
+
+}
+
 ManagerCommands.prototype.addCommand = function(command) {
-    this.queueAsync.queue(command);
+    // parse host
+    let layer = command.layer;
+
+    let host = layer.url ? new URL(layer.url).host : undefined;
+
+    // init queue if needed
+    if (host && !(this.hostQueues.has(host))) {
+        this.hostQueues.set(host, _instanciateQueue());
+    }
+
+    let q = host ? this.hostQueues.get(host) : this.defaultQueue;
+
+    // execute command now if possible
+    if (q.counters.executing < this.maxCommandsPerHost) {
+        if (host) {
+            this.runCommand(command, q);
+        } else {
+            Promise.resolve(true).then(function() {
+                this.runCommand(command, q);
+            }.bind(this));
+        }
+    } else {
+        q.storage.queue(command);
+    }
 };
 
 
@@ -55,50 +136,44 @@ ManagerCommands.prototype.getProtocolProvider = function(protocol) {
 };
 
 ManagerCommands.prototype.commandsLength = function() {
-    return this.queueAsync.length;
+    return 0;//return this.queueAsync.length;
 };
 
 ManagerCommands.prototype.isFree = function() {
     return this.commandsLength() === 0;
 };
 
-ManagerCommands.prototype.runAllCommands = function() {
-
-
-    if (this.commandsLength() === 0) {
-        return Promise.resolve(0);
-    }
-
-    return Promise.all(this.arrayDeQueue(16))
-        .then(function() {
-            // if (this.commandsLength() <= 16)
-            this.scene.wait(1);
-            // else
-            //     this.scene.renderScene3D();
-            return this.runAllCommands();
-        }.bind(this));
+ManagerCommands.prototype.resetExecutedCommandsCount = function() {
+    this.counters.executedCommands = 0;
 };
 
-ManagerCommands.prototype.arrayDeQueue = function(number) {
+ManagerCommands.prototype.commandsWaitingExecutionCount = function() {
+    let sum = this.defaultQueue.storage.length;
 
-    var nT = number === undefined ? this.queueAsync.length : number;
-
-    var arrayTasks = [];
-
-    while (this.queueAsync.length > 0 && arrayTasks.length < nT) {
-        var command = this.deQueue();
-
-        if (command) {
-            var layer = command.layer;
-            var provider = this.providers[layer.protocol];
-            if (provider) {
-                arrayTasks.push(provider.executeCommand(command));
-            }
-        }
+    for (var q of this.hostQueues) {
+        sum += q[1].storage.length;
     }
-
-    return arrayTasks;
+    return sum;
 };
+
+ManagerCommands.prototype.commandsRunningCount = function() {
+    let sum = this.defaultQueue.counters.executing;
+
+    for (var q of this.hostQueues) {
+        sum += q[1].counters.executing;
+    }
+    return sum;
+};
+
+ManagerCommands.prototype.commandsCancelledCount = function() {
+    let sum = this.defaultQueue.counters.cancelled;
+
+    for (var q of this.hostQueues) {
+        sum += q[1].counters.cancelled;
+    }
+    return sum;
+};
+
 
 ManagerCommands.prototype.getProviders = function() {
     var p = [];
@@ -107,21 +182,22 @@ ManagerCommands.prototype.getProviders = function() {
         p.push(this.providers[protocol]);
     }
     return p;
-}
+};
+
 
 
 /**
  */
-ManagerCommands.prototype.deQueue = function() {
-
-    while (this.queueAsync.length > 0) {
-        var cmd = this.queueAsync.peek();
+ManagerCommands.prototype.deQueue = function(queue) {
+    var st = queue.storage;
+    while (st.length > 0) {
+        var cmd = st.dequeue();
 
         if (cmd.earlyDropFunction && cmd.earlyDropFunction(cmd)) {
-            this.queueAsync.dequeue();
+            queue.counters.cancelled++;
             cmd.reject(new Error('command canceled ' + cmd.requester.id + '/' + cmd.layer.id));
         } else {
-            return this.queueAsync.dequeue();
+            return cmd;
         }
 
     }
@@ -131,32 +207,10 @@ ManagerCommands.prototype.deQueue = function() {
 
 /**
  */
-ManagerCommands.prototype.removeCanceled = function() {
-    //TODO: Implement Me
-
-};
-
-/**
- */
 ManagerCommands.prototype.wait = function() {
-    //TODO: Implement Me
     this.eventsManager.wait();
 };
 
-/**
- */
-ManagerCommands.prototype.forecast = function() {
-    //TODO: Implement Me
-
-};
-
-/**
- * @param object
- */
-ManagerCommands.prototype.addInHistory = function( /*object*/ ) {
-    //TODO: Implement Me
-
-};
 
 export default function(scene) {
     instanceCommandManager = instanceCommandManager || new ManagerCommands(scene);
--- a/src/Core/Commander/ManagerCommands.js
+++ b/src/Core/Commander/ManagerCommands.js
@@ -27,12 +27,15 @@ function _instanciateQueue() {
             failed: 0,
             cancelled: 0
         },
-        execute: function(cmd, provider) {
-            this.counters.executing++;
+        execute: function(cmd, provider, executingCounterUpToDate) {
+            if (!executingCounterUpToDate) {
+                this.counters.executing++;
+            }
 
-            var p = provider.executeCommand(cmd);
+            // If the provider returns a Promise, use it to handle counters
+            // Otherwise use a resolved Promise.
+            var p = provider.executeCommand(cmd) || Promise.resolve();
 
-            if (!p) throw new Error('arg');
             return p.then(
                 function() {
                     this.counters.executing--;
@@ -62,11 +65,6 @@ function ManagerCommands(scene) {
     this.eventsManager = new EventsManager();
     this.maxConcurrentCommands = 16;
     this.maxCommandsPerHost = 6;
-    this.counters = {
-        runningCommands: 0,
-        executedCommands: 0,
-        hostCommands: {}
-    };
 
     if (!scene)
         throw new Error("Cannot instantiate ManagerCommands without scene");
@@ -77,15 +75,19 @@ function ManagerCommands(scene) {
 
 ManagerCommands.prototype.constructor = ManagerCommands;
 
-ManagerCommands.prototype.runCommand = function(command, queue) {
+ManagerCommands.prototype.runCommand = function(command, queue, executingCounterUpToDate) {
     var provider = this.providers[command.layer.protocol];
 
     if (!provider) {
         throw new Error('No known provider for layer', command.layer.id);
     }
 
-    queue.execute(command, provider).then(function() {
-        this.scene.notifyChange();
+    queue.execute(command, provider, executingCounterUpToDate).then(function() {
+        // notify scene that one command ended.
+        // We allow the scene to delay the update/repaint up to 100ms
+        // to reduce CPU load (no need to perform an update on completion if we
+        // know there's another one ending soon)
+        this.scene.notifyChange(100);
 
         // try to execute next command
         if (queue.counters.executing < this.maxCommandsPerHost) {
@@ -114,13 +116,16 @@ ManagerCommands.prototype.addCommand = function(command) {
 
     // execute command now if possible
     if (q.counters.executing < this.maxCommandsPerHost) {
-        if (host) {
-            this.runCommand(command, q);
-        } else {
-            Promise.resolve(true).then(function() {
-                this.runCommand(command, q);
-            }.bind(this));
-        }
+        // increment before
+        q.counters.executing++;
+
+        var runNow = function() {
+            this.runCommand(command, q, true);
+        }.bind(this);
+
+        // We use a setTimeout to defer processing but we avoid the
+        // queue mechanism
+        window.setTimeout(runNow, 0);
     } else {
         q.storage.queue(command);
     }
@@ -135,23 +140,11 @@ ManagerCommands.prototype.getProtocolProvider = function(protocol) {
     return this.providers[protocol];
 };
 
-ManagerCommands.prototype.commandsLength = function() {
-    return 0;//return this.queueAsync.length;
-};
-
-ManagerCommands.prototype.isFree = function() {
-    return this.commandsLength() === 0;
-};
-
-ManagerCommands.prototype.resetExecutedCommandsCount = function() {
-    this.counters.executedCommands = 0;
-};
-
 ManagerCommands.prototype.commandsWaitingExecutionCount = function() {
-    let sum = this.defaultQueue.storage.length;
-
+    let sum = this.defaultQueue.storage.length
+        + this.defaultQueue.counters.executing;
     for (var q of this.hostQueues) {
-        sum += q[1].storage.length;
+        sum += q[1].storage.length + q[1].counters.executing;
     }
     return sum;
 };
@@ -165,15 +158,16 @@ ManagerCommands.prototype.commandsRunningCount = function() {
     return sum;
 };
 
-ManagerCommands.prototype.commandsCancelledCount = function() {
-    let sum = this.defaultQueue.counters.cancelled;
+ManagerCommands.prototype.resetCommandsCount = function(type) {
 
+    let sum = this.defaultQueue.counters[type];
+    this.defaultQueue.counters[type] = 0;
     for (var q of this.hostQueues) {
-        sum += q[1].counters.cancelled;
+        sum += q[1].counters[type];
+        q[1].counters[type] = 0;
     }
     return sum;
 };
-
 
 ManagerCommands.prototype.getProviders = function() {
     var p = [];
--- a/src/Core/Commander/Providers/TileProvider.js
+++ b/src/Core/Commander/Providers/TileProvider.js
@@ -105,7 +105,9 @@ TileProvider.prototype.executeCommand = function(command) {
     tile.updateMatrix();
     tile.updateMatrixWorld();
 
-    return command.resolve(tile);
+    command.resolve(tile);
+
+    return Promise.resolve();
 };
 
 export default TileProvider;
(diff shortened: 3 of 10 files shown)


src/Core/Commander/ManagerCommands.js
```diff
@@ -82,16 +82,14 @@ define('Core/Commander/ManagerCommands', [
...
             return Promise.all(this.arrayDeQueue(16))
-                .then(function () {
-                    // if (this.commandsLength() <= 16)
-                    this.scene.wait(1);
-                    // else
-                    //     this.scene.renderScene3D();
-                    return this.runAllCommands();
+                .then(function() {
+
+                    return Promise.resolve(false);
```
tbroyer: `return false;` is enough

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -246,7 +246,7 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
     ApiGlobe.prototype.update = function() {
 
         //!\\ TEMP
-        this.scene.wait(0);
+        this.scene.step();//wait(0);
```
peppsac: Please remove dead code.

src/Core/Commander/ManagerCommands.js
```diff
@@ -82,16 +82,14 @@ define('Core/Commander/ManagerCommands', [
...
+
                 }.bind(this));
```
peppsac: Is  `.bind(this)` still needed?

Jeremy-Gaillard:
Applied reviewers' suggestions.

Don't merge yet, there is still a problem with the globe loaded event.

Jeremy-Gaillard:
Ready for review.

peppsac:
So I gave this branch a quick test.
2 issues so far : 
- it takes almost 20 sec to get the first complete rendering of the globe. This is a regression compared to master (3 sec)
- high cpu load during that time and for a little longer (enough to turn on the fans on my laptop :) )

Jeremy-Gaillard:
- What do you mean exactly by complete rendering? Are you waiting for the globe loaded event or the first fully textured globe? In the first case, I do not have such a difference.
- Working on it. I just saw that there are far too much rendering pass.

peppsac:
> What do you mean exactly by complete rendering? Are you waiting for the globe loaded event or the first fully textured globe? In the first case, I do not have such a difference.

Yes, the first fully texture globe. Do you observe the same thing as me?

Jeremy-Gaillard:
Yes. I think it's a problem with the request for the higher resolution tiles being completed before the lower resolution ones.

The performance issues seem to come from the rendering/updating happening each time a command is completed. On the master, 16 commands are executed before each new render.

Jeremy-Gaillard:
I pushed a commit where I wait for multiple commands to be executed before updating. The performance should be far better.

Edit: don't bother yet, can't zoom without freezing

Jeremy-Gaillard:
Should be working now. Needs rebasing however.

src/Scene/Scene.js
```diff
@@ -126,32 +129,8 @@ define('Scene/Scene', [
...
-                }
-
-            }.bind(this));
-
-        this.renderScene3D();
-
-    };
-
     Scene.prototype.realtimeSceneProcess = function() {
+        // rename sceneUpdate/layerUpdate?
```
peppsac: Indeed `realtimeSceneProcess` isn't clear. `layersUpdate` ?
Probably worth doing it in this PR since you're cleaning the update loop.

src/Scene/Scene.js
```diff
@@ -161,7 +140,7 @@ define('Scene/Scene', [
                 var sLayer = layer.children[sl];
 
                 if (sLayer instanceof Quadtree)
-                    this.browserScene.browse(sLayer, this.currentCamera(), process, NO_SUBDIVISE);
+                    this.browserScene.browse(sLayer, this.currentCamera(), process, NO_SUBDIVIDE);
```
peppsac: Consider fixing the coding style as well (add `{... }` around if/else blocks)

src/Scene/Scene.js
```diff
@@ -180,15 +159,49 @@ define('Scene/Scene', [
         this.gfxEngine.update();
     };
 
-    Scene.prototype.wait = function(timeWait) {
+    Scene.prototype.notifyChange = function() {
```
peppsac: - add documentation for this method
- Maybe `scene.needsUpdate()` is clearer than `scene.notifyChange()`? (or `scene.invalidate()`?)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -246,7 +246,7 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
     ApiGlobe.prototype.update = function() {
 
         //!\\ TEMP
-        this.scene.wait(0);
+        this.scene.step(0);
```
peppsac: Are the `//!\\ TEMP` comments still needed?
And I think it should be a `this.scene.notifyChange()`

src/Core/Commander/ManagerCommands.js
```diff
@@ -77,44 +80,35 @@ define('Core/Commander/ManagerCommands', [
             return this.commandsLength()===0;
         };
 
-        ManagerCommands.prototype.runAllCommands = function() {
-
-
-            if (this.commandsLength() === 0)
-            {
-                return Promise.resolve(0);
+        ManagerCommands.prototype.fillCommandPool = function() {
```
peppsac: `fillCommandPool` is describing _how_ ManagerCommands does its job but fails to describe _what_ this method is supposed to do.
IMHO `managerCommands.update()`, `managerCommands.execCommands()` are better. WDYT?

src/Core/Commander/ManagerCommands.js
```diff
@@ -77,44 +80,35 @@ define('Core/Commander/ManagerCommands', [
...
+
+            var command;
+            var that = this;
```
peppsac: I think @tbroyer advised against this pattern and recommended using `bind(this)`

src/Core/Commander/ManagerCommands.js
```diff
@@ -77,44 +80,35 @@ define('Core/Commander/ManagerCommands', [
...
+            function launchCommand(cmd) {
+                var providers = that.getProviders(cmd.layer);
+                for (var i = 0; i < providers.length; i++) {
+                    that.currentCommandsCount++;
```
peppsac: `currentCommandsCount` -> `runningCommandsCount`?

src/Core/Commander/ManagerCommands.js
```diff
@@ -77,44 +80,35 @@ define('Core/Commander/ManagerCommands', [
...
+                    var p = providers[i].executeCommand(cmd);
+                    if(p) {
+                        p.then(function() {
+                            that.currentCommandsCount--;
+                            that.executedCommands++;
```
peppsac: `executedCommands` could have the same `Count` suffix.
Or actually I'd prefer them to leave in a sub-object:

```
this.counters.currentCommands--
this.counters.executedCommands++
```

src/Core/Commander/ManagerCommands.js
```diff
@@ -77,44 +80,35 @@ define('Core/Commander/ManagerCommands', [
...
-                var command = this.deQueue();
-                if(command)
-                {
-
-                    // TEMP
-
-                    var providers = this.getProviders(command.layer);
-                    for (var i = 0; i < providers.length; i++)
-                        arrayTasks.push(providers[i].executeCommand(command));
+            for(var i = this.currentCommandsCount; i < this.maxConcurrentCommands; i++) {
```
peppsac: Consider adding a short comment explaining what is going on.

src/Scene/Scene.js
```diff
@@ -75,6 +75,9 @@ define('Scene/Scene', [
 
         this.gLDebug = gLDebug;
         this.gfxEngine = c3DEngine(this,positionCamera,viewerDiv, debugMode,gLDebug);
+        this.needsUpdate = true;
+        this.forceUpdate = true;
```
peppsac: What are these boolean doing? I hope we can get rid of one of them because they seem redundant...

src/Scene/Scene.js
```diff
@@ -180,15 +159,49 @@ define('Scene/Scene', [
...
+            this.needsUpdate = false;
+        }
+    };
+
+    Scene.prototype.step = function(timestamp) {
 
-        var waitTime = timeWait ? timeWait: 20;
+        if(this.forceUpdate || this.needsUpdate && (this.managerCommand.executedCommands >= 16 || (timestamp - this.lastTimestamp) > 200 && this.managerCommand.executedCommands !== 0)) {
+            this.forceUpdate = false;
+            this.managerCommand.executedCommands = 0;
```
peppsac: Why are you modifying managerCommand internal state here?

peppsac: Except if `executedCommands` means `executedCommandsSinceLastRendering`.
But in this case, consider adding a method to ManagerCommands instead (`resetExecutedCommandsCount`)

src/Scene/Scene.js
```diff
@@ -180,15 +159,49 @@ define('Scene/Scene', [
...
+            this.browserScene.browse(quadtree,this.currentCamera(), process, CLEAN);
+            this.viewerDiv.dispatchEvent(event);
+            this.needsUpdate = false;
+        }
+    };
+
+    Scene.prototype.step = function(timestamp) {
 
-        var waitTime = timeWait ? timeWait: 20;
+        if(this.forceUpdate || this.needsUpdate && (this.managerCommand.executedCommands >= 16 || (timestamp - this.lastTimestamp) > 200 && this.managerCommand.executedCommands !== 0)) {
```
peppsac: This big conditional statement, mixing && and ||, needs to be cleaned/simplified.
At the very least, introduce some simple boolean:

``` js
var atLeast16commandsExecutedSinceLastRender = this.managerCommand.executedCommands >= 16;
var timeout5fps = (timestamp - this.lastTimestamp) > 200; // not sure why you want at least 1 command executed here?
```

src/Scene/Scene.js
```diff
@@ -180,15 +159,49 @@ define('Scene/Scene', [
...
+            // run commands
+            this.runCommands();
 
-        window.clearInterval(this.timer);
+            // render scene
+            console.log("render");
+            this.renderScene3D();
 
-        this.timer = window.setTimeout(this.quadTreeRequest.bind(this), waitTime,this.layers[0].node.tiles, this.layers[0].process);
+            this.lastTimestamp = timestamp;
```
peppsac: Rename to `this.lastRendertime` or `this.lastUpdateTime`

peppsac:
I left a few comments, but overall great job on this PR @Jeremy-Gaillard :+1:

src/Core/Commander/ManagerCommands.js
```diff
@@ -40,6 +40,9 @@ define('Core/Commander/ManagerCommands', [
             this.providerMap = {};
             this.history = null;
             this.eventsManager = new EventsManager();
+            this.currentCommandsCount = 0;
+            this.maxConcurrentCommands = 16;
```
peppsac: I'm not sure that hardcoding a command count is a good idea.
I believe the purpose is two-fold:
- limit the number of parallel network fetches (btw, 16 is too much. On Chrome lots of requests are in 'queued' state, see https://developers.google.com/web/tools/chrome-devtools/profile/network-performance/understanding-resource-timing )
- keep a high enough framerate

For the first point maybe we need a global network fetch limit (16?) + a per-host limit (Chrome documentation suggests 6).
For the second case we need a update duration limit (50 ms if we want to keep 20fps for instance).
Last but not least: some commands are not doing network job (it'll be more visible once #149 is merged).

tbroyer: I haven't really followed the discussion (and haven't looked at the patch besides those few lines), but to keep a high framerate, the idea is generally to flip a flag whenever there's a need to update the "rendering", and in a `requestAnimationFrame` callback to re-render based on the flag.

I have no idea what commands are and how they're used in iTowns, so I can't really comment on how things should work wrt queuing and executing them, but things I've seen elsewhere related to similarly-named patterns are to:
- use [`setImmediate`](https://developer.mozilla.org/en/docs/Web/API/Window/setImmediate) or similar to defer processing (see https://www.nczonline.net/blog/2013/07/09/the-case-for-setimmediate/ )
- run as many commands as possible, timing them until you reach 15ms or so, to keep a 60fps, and deferring the rest to a new `setImmediate` (see https://www.nczonline.net/blog/2009/08/11/timed-array-processing-in-javascript/)

`requestAnimationFrame` should only be used for things that directly affect rendering (so rather as a side-effect of commands IIUC), not as a general-purpose task scheduler.

peppsac: `setImmediate` is non-standard so maybe we should prefer a standard based implementation.
This post https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/Hn3GxRLXmR0/XP9xcY_gBPQJ explains why Chrome didn't implement `setImmediate` but also details several alternatives.
The last `tick` function in his post matches what we need; it's just missing a call to `requestAnimationFrame`

src/Scene/Scene.js
Jeremy-Gaillard: The two boolean have different meaning: needsUpdate indicates that the scene is not in a stable state (still needs subdivision or textures...), forceUpdate means the scene has to be rendered, even if the scene has not changed.
The scene can need an update but not update for performance reasons: typically, we want to wait for a certain amount of commands to be executed before rendering and updating the scene.

I clarified all this in a new commit.

src/Core/Commander/ManagerCommands.js
tbroyer: `requestIdleCallback`? https://twitter.com/jacobrossi/status/667047851415969792 & http://permalink.gmane.org/gmane.org.w3c.webperf/757 (not more standard, but with public support from several vendors: https://www.chromestatus.com/feature/5572795866021888 / https://platform-status.mozilla.org/#request-idle-callback)

FWIW, that `tick` function is the one I described above, and more or less what can be found in Nicholas Zakas' blog entry I linked to ;-)

Jeremy-Gaillard: I'll look at these links and see if I can adapt them to our case.

src/Scene/Scene.js
peppsac: What `force` really means is `forceRendering` right?

Jeremy-Gaillard: Yes, the name has changed in the latest version : https://github.com/iTowns/itowns2/pull/141/files#diff-a6b95dde1d808f9cea55c3fc36fc7f2bR190

Jeremy-Gaillard:
Rebased and ready for review.

src/Core/Commander/ManagerCommands.js
```diff
@@ -127,7 +133,16 @@ define('Core/Commander/ManagerCommands', [
...
+                var requester = cmd.requester;
+                if (requester.disposed) {
+                    this.queueAsync.dequeue();
+                } else if (cmd.earlyDropFunction && cmd.earlyDropFunction(cmd)) {
+                    while (requester.children.length > 0) {
+                        var child = requester.children[0];
+                        child.dispose();
+                        requester.remove(child);
+                    }
+                    requester.pendingSubdivision = false;
```
peppsac: why did you restore this code?

src/Scene/Scene.js
```diff
@@ -180,15 +160,63 @@ define('Scene/Scene', [
...
+        // request new frame in case the rendering loop was on hold
+        // force rendering to ensure interactivity
+        requestAnimationFrame(function(timestamp) { this.step(timestamp, true); }.bind(this));
+    };
 
-        var waitTime = timeWait ? timeWait: 20;
+    /**
+     * Subdivide layers and clean them if necessary
+     */
+    Scene.prototype.processLayers = function() {
```
peppsac: `processLayers` is misleading. `processQuadtreeLayer` ?

Though both `updateLayers` and `processLayers` are confusing.
Consider adding some documentation/comments (esp. : why do we need 2 functions? why updateLayers is only needed if `forceRender` is true, etc)

src/Core/Commander/ManagerCommands.js
```diff
@@ -71,43 +76,44 @@ define('Core/Commander/ManagerCommands', [
...
+                if(p) {
+                    var promise = p.executeCommand(cmd);
+                    if(promise) {
+                        promise.then(function() {
+                            this.counters.runningCommands--;
+                            this.counters.executedCommands++;
+                        }.bind(this));
+                    } else {
+                        this.counters.runningCommands--;
+                        this.counters.executedCommands++;
```
peppsac: Still no per-host command limit ? (see my earlier comment on why it's important)

src/Core/Commander/ManagerCommands.js
Jeremy-Gaillard: It's not exactly restored. It is part of this commit a1135483f65c3c07f0568434241697eabbe985c3 that fixed some commands being launched for disposed nodes.
With the addition of the early drop function, the bug seems to have disappeared, I'll remove the commit.

src/Scene/Scene.js
Jeremy-Gaillard: I'll try to find better names.
updateLayers updates the state of the layer for visualisation (tile visibility, sse...)
processLayers creates commands and cleans the quadtree.

src/Core/Commander/ManagerCommands.js
Jeremy-Gaillard: Looking at it.

Jeremy-Gaillard:
Added a limit to the number of commands per host.
I still need to add a few comments before being ready to merge.

src/Core/Commander/ManagerCommands.js
Jeremy-Gaillard: So, I don't think this mechanism is really suited to our case. If I understand your links correctly, the goal of the tick function is to execute the maximum of atoms in 16ms. However, we don't have a lot to do at each frame, we just create new promises. The only big operation is the tree browsing which can't (easily) be cut into multiple atoms.
However, it may be useful in the future, when multiple quadtrees (or other data structures) will need updating.

src/Core/Commander/ManagerCommands.js
```diff
@@ -71,43 +79,80 @@ define('Core/Commander/ManagerCommands', [
...
-            var arrayTasks = [];
+        ManagerCommands.prototype.executeCommands = function() {
+
+            var command;
+            var launchCommand = function(cmd) {
+                this.counters.runningCommands++;
+                var layer = cmd.layer;
+                var host;
+                if(layer.url) {
+                    host = new URL(layer.url).host;
```
peppsac: Consider doing this operation just once when the command is created (in InterfaceCommander maybe?).

src/Core/Commander/ManagerCommands.js
```diff
@@ -52,6 +59,7 @@ define('Core/Commander/ManagerCommands', [
 
         ManagerCommands.prototype.addCommand = function(command) {
             this.queueAsync.queue(command);
```
peppsac: I'd suggest doing this instead:

``` js
if (commandIsRunnable(command)) {
  executeCommand(command);
} else {
  this.queueAsync.queue(command);
}
```

(no need to queue/dequeue an immediately runnable command).

You can also remove the call to `this.executeCommands()` which become useless with this change.

src/Core/Commander/ManagerCommands.js
```diff
@@ -71,43 +79,80 @@ define('Core/Commander/ManagerCommands', [
...
 
-        ManagerCommands.prototype.arrayDeQueue = function(number) {
-
-            var nT = number === undefined ? this.queueAsync.length : number;
-
-            var arrayTasks = [];
+        ManagerCommands.prototype.executeCommands = function() {
+
+            var command;
+            var launchCommand = function(cmd) {
```
peppsac: This function with nested functions looks really complex. We should aim to simplify the code.

src/Core/Commander/ManagerCommands.js
```diff
@@ -71,43 +79,80 @@ define('Core/Commander/ManagerCommands', [
...
+                        }.bind(this));
+                    } else {
+                        commandEnd();
+                    }
+                } else {
+                    commandEnd();
+                }
+            }.bind(this);
+
+            var skippedCommands = [];
```
peppsac: We could have a sorted commands queue per host so we don't have to implement this skipping mechanism.
Or if we believe that having a single queue is useful (for global sorting for instance) then we could consider changing the underlying storage to better support our use case.

src/Core/Commander/ManagerCommands.js
```diff
@@ -71,43 +79,80 @@ define('Core/Commander/ManagerCommands', [
...
 
-            while (this.queueAsync.length > 0 && arrayTasks.length < nT) {
-                var command = this.deQueue();
+                var commandEnd = function() {
+                    this.counters.runningCommands--;
+                    this.counters.executedCommands++;
+                    if(host) {
+                        this.counters.hostCommands[host]--;
+                        if(this.counters.hostCommands[host] === 0) {
+                            this.counters.hostCommands[host] = undefined;
```
peppsac: What's the point of replacing `0` by `undefined`?

src/Scene/Scene.js
```diff
@@ -50,8 +50,8 @@ define('Scene/Scene', [
 
     var instanceScene = null;
     var event = new Event('globe-built');
-    var NO_SUBDIVISE = 0;
-    var SUBDIVISE = 1;
+    var NO_SUBDIVIDE = 0;
+    var SUBDIVIDE = 1;
```
peppsac: While fixing typo you can also replace `var` with `const`

src/Core/Commander/ManagerCommands.js
peppsac: Or in `addCommand`.
And at the same time you could init the `this.counters.hostCommands[host]` to remove all the `if(this.counters.hostCommands[host] === undefined)` checks.

peppsac:
I updated this PR based, on PR #176 and #178 .
It also includes rework of error handling which will move to a different PR at some point.
This is not ready for merging yet, but you can already give it a try.

peppsac:
Branch updated. Still not ready for merging (waiting for PR #176 to land first)

peppsac:
Updated the branch - it's not ready for merging yet.

The last commit is a debug commit, just to share with you what I'm using and gather feedback: do you think it would be useful to merge something like this in itowns?

Here's what it looks like:
tools

You can test it using: `DEV_BUILD=true npm start`

gchoqueux:
Some color layers aren't refined when the camera stops to move.
And there are error messages in NodeProcess.js:155
Uncaught (in promise) TypeError: Cannot read property 'childrenLoaded' of null(…)

gchoqueux:
There are error messages in NodeProcess.js:155
_Uncaught (in promise) TypeError: Cannot read property 'childrenLoaded' of null(…)_

gchoqueux:
The event `'globe-loaded'` is broken. It emits before all layers are loaded

src/Renderer/LayeredMaterial.js
```diff
@@ -20,7 +20,7 @@ var nbLayer = 2;
 var vector = new THREE.Vector3(0.0, 0.0, 0.0);
 var vector2 = new THREE.Vector2(0.0, 0.0);
 var vector4 = new THREE.Vector4(0.0, 0.0, 0.0, 0.0);
-var showDebug = false;
+var showDebug = true;
```
gchoqueux: `var showDebug = true;` replace by false

gchoqueux:
Some time, color layers flick for refinement.
Tiles do goings and comings in their ancestors : [Video 1](https://drive.google.com/file/d/0B5xBYEUN9JvFUWQ1NWtWTnpHdXc/view?usp=sharing) 
This bug appears after large zoom [Video 2](https://drive.google.com/file/d/0B5xBYEUN9JvFWVdWTmg4cENxT00/view?usp=sharing), the tiles doesn't use the good ancestor.

peppsac:
@gchoqueux I fixed some of the issues you reported : 
- `error messages in NodeProcess.js:155`
- `The event 'globe-loaded' is broken`

I do not experiment flickering here though. If you have a way to reproduce the problem I'be interested to learn about it :-)

gchoqueux:
Debug tools needs new PR. 
and The Travis CI build failed with 'Debug tools'

peppsac:
Meh... debug tools are not part of this PR - I just forgot to not push them.
Anyway, do you still reproduce your issues with this version of the code?

gchoqueux:
> debug tools are not part of this PR

ok no problem

> Anyway, do you still reproduce your issues with this version of the code?

I add comment and new video in [`Issue comment`](https://github.com/iTowns/itowns2/pull/141#issuecomment-252026191)

peppsac:
@gchoqueux branch updated.

peppsac:
Updated again.

gchoqueux:
Textures color aren't loaded.
Selected tile have material.nbTextures[Elevation] = 4 ??

capture du 2016-10-27 15-26-07

gchoqueux:
Tile doesn't load, children are builded but not loaded

capture du 2016-10-27 15-44-03

peppsac:
Does it happen each time?

gchoqueux:
no

gchoqueux:
There are performance problems, particularly during loading's globe 
**TimeLine on master** : There aren't many rendering and time loading is 1,2 second (without evaluate scripting)
master

**TimeLine on updateLoop** :  There are a lot of rendering and time loading is 8,5 second (without evaluate scripting)
loop

**Others examples** start loading (start center different)
- Master Loading is 1,2 second
  master2
- updateLoop Loading is 3.4 second
  loop2

peppsac:
We really need to start working on a test framework to prevent regressions (bug or performance)...
Anyway I pushed a new patch that fix performance problems.

index.html
```diff
@@ -543,6 +543,12 @@
                         "maxTileRow": 9375,
                         "minTileCol": 15914,
                         "maxTileCol": 20507
+                    },
```
gchoqueux: Add a mechanism in the API, if  a new `tileMatrixSetLimits` is needed then it's added automatically. 
And remove this :

``` js
  "20": {}
```

src/Core/Commander/ManagerCommands.js
```diff
@@ -55,49 +140,40 @@ ManagerCommands.prototype.getProtocolProvider = function(protocol) {
 };
 
 ManagerCommands.prototype.commandsLength = function() {
-    return this.queueAsync.length;
+    return 0;//return this.queueAsync.length;
```
gchoqueux: what is the use of this function?

src/Core/Commander/Providers/TileProvider.js
```diff
@@ -105,7 +105,9 @@ TileProvider.prototype.executeCommand = function(command) {
     tile.updateMatrix();
     tile.updateMatrixWorld();
 
-    return command.resolve(tile);
+    command.resolve(tile);
+
+    return Promise.resolve();
```
gchoqueux: why empty Promise?

src/Scene/BrowseTree.js
```diff
@@ -137,53 +98,38 @@ BrowseTree.prototype.browse = function(tree, camera, process, layersConfig, opti
...
-BrowseTree.prototype._clean = function(node, level, process, camera) {
+BrowseTree.prototype._browseNonDisplayableNode = function(node, level, process, camera, params) {
     // update node's sse value
     node.sse = camera.computeNodeSSE(node);
+    node.setDisplayed(false);
 
     var sse = process.checkNodeSSE(node);
 
+    if (!sse && !node.loaded) {
+        process.refineNodeLayers(node, camera, params);
```
gchoqueux, line 126: Why refine node if the node isn't displayed?

src/Scene/NodeProcess.js
```diff
@@ -292,95 +302,135 @@ function updateNodeImagery(quadtree, node, layersConfig, force) {
...
+    node.setDisplayed(false);
+    node.setSelected(false);
+
+    node.setVisibility(isVisible);
+
+    if (isVisible) {
+        // update node's sse value
+        node.sse = camera.computeNodeSSE(node);
+
+        var sse = this.checkNodeSSE(node);
```
gchoqueux: let or var?

src/Scene/BrowseTree.js
```diff
@@ -272,4 +218,10 @@ BrowseTree.prototype.updateMobileMappingLayer = function(layer, camera) {
...
 };
 
-export default BrowseTree;
+BrowseTree.prototype.updateQuadtree = function(layer, layersConfiguration, camera) {
+    var quadtree = layer.node.tiles;
+
+    this.browse(quadtree, camera, layer.process, layersConfiguration);
+};
+
+export { BrowseTree };
```
gchoqueux: why you replace `export default BrowseTree;` in BrowseTree.js

gchoqueux (changes requested):

gchoqueux:
this PR could be separated into two : _'requestAnimationFrame'_ and _'rework ProcessNode and browseTree'_

index.html
peppsac: This issue is a WMTS one. I'm going to modify `tileInsideLimit` and level instead.
Basically: tileInsideLimit will not test tile.level against zoom.max ; but before fetching a texture we'll clamp the requested level to zoom.max.
WDYT?

src/Core/Commander/ManagerCommands.js
peppsac: Now it's useless indeed. I'll remove it.

src/Core/Commander/Providers/TileProvider.js
peppsac: I will revert this change.

src/Scene/BrowseTree.js
peppsac: I should add a comment. It's meant to make sure that no tile is stuck in the `!loaded` state.

src/Scene/BrowseTree.js
peppsac: Good catch, I'll revert this too

src/Scene/NodeProcess.js
peppsac: I should indeed transform `var`s in `let`s

peppsac:
> this PR could be separated into two : 'requestAnimationFrame' and 'rework ProcessNode and browseTree'

I'm not so sure as there are lots of dependencies between both (change in Scene/BrowseTree where driven by changes in requestAnimationFrame and the other way around).

But let's put this issue aside for a while, I'd rather focus on the content of the PR itself :-)

gchoqueux (approved):

PR #143, opened 2016-07-08
Prevent creation of the same command multiple times

Jeremy-Gaillard:
Refinement commands were created even when there was already a command under way.
A pending attribute has been added to the node to prevent this from happening.

### DIFF
--- a/src/Scene/Node.js
+++ b/src/Scene/Node.js
@@ -24,6 +24,7 @@ define('Scene/Node', [], function() {
         this.screenSpaceError = 0.0;
         this.loaded = false;
         this.pendingSubdivision = false;
+        this.pendingLayers = {};
         this.visible = true;
         this.layer = null;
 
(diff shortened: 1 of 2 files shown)


src/Scene/Node.js
```diff
@@ -24,6 +24,7 @@ define('Scene/Node', [], function() {
         this.screenSpaceError = 0.0;
         this.loaded = false;
         this.pendingSubdivision = false;
+        this.pendingLayers = {};
```
peppsac, line 27: Given that `getDownScaledLayer` will always return the same id _until_ it has been updated, maybe we can use a simple `bool` here?
(just a suggestion, doens't matter much)

peppsac:
Can you rebase this branch please (I'll merge it after)?

src/Scene/Node.js
Jeremy-Gaillard, line 27: I think we should support this behaviour, we may want to send the commands for all the layers concurrently in the future.

Jeremy-Gaillard:
Done.

peppsac:
LGTM, merging

PR #149, opened 2016-07-12
Refacto providers layers

peppsac:
Quite a big PR, touching a lot of places.

The 3 commits   0e7a421,    362f5ef,  6f54449 have a circular dependency so you can't test the first 2 alone. I kept individual commits anyway to ease the review.

There are some open questions for the future (e.g: `colorTerrain` Layer in Globe.js) but I'd prefer to get some feedback on the work already done before digging further.

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -105,19 +105,21 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
 
     ApiGlobe.prototype.moveLayerUp = function(layer){
 
-        this.scene.getMap().moveLayerUp(layer);
+        this.scene.getMap().layersConfiguration.moveLayerUp(layer);
+        this.scene.getMap().updateLayersOrdering();
         this.scene.renderScene3D();
     };
 
     ApiGlobe.prototype.moveLayerDown = function(layer){
 
-        this.scene.getMap().moveLayerDown(layer);
+        this.scene.getMap().layersConfiguration.moveLayerDown(layer);
+        this.scene.getMap().updateLayersOrdering();
         this.scene.renderScene3D();
     };
 
     ApiGlobe.prototype.moveLayerToIndex = function(layer,newId){
-
-        this.scene.getMap().moveLayerToIndex(layer,newId);
+        this.scene.getMap().layersConfiguration.moveLayerToIndex(layer,newId);
+        this.scene.getMap().updateLayersOrdering();
         this.scene.renderScene3D();
     };
 
@@ -127,6 +129,7 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
         {
             eventLayerRemoved.layer = id;
             this.viewerDiv.dispatchEvent(eventLayerRemoved);
+            this.scene.getMap().updateLayersOrdering();
             this.scene.renderScene3D();
             return true;
         }
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -72,10 +72,22 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
 
     };
 
+
     ApiGlobe.prototype.getWMTSProvider = function()
     {
-        return this.scene.managerCommand.getProvider(this.scene.getMap().tiles).providerWMTS;
-    };
+        return this.scene.managerCommand.getProtocolProvider('wmts');
+    }
+
+    /**
+    * This function gives a chance to the matching provider to pre-process some
+    * values for a layer.
+    */
+    function preprocessLayer(layer, provider) {
+        if (provider.preprocessDataLayer) {
+            layer.tileInsideLimit = provider.tileInsideLimit.bind(provider);
+            provider.preprocessDataLayer(layer);
+        }
+    }
 
     /**
     * This function adds an imagery layer to the scene. The layer id must be unique. The protocol rules wich parameters are then needed for the function.
@@ -83,19 +95,12 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
     * @param {Layer} layer.
     */
     ApiGlobe.prototype.addImageryLayer = function(layer) {
+        preprocessLayer(layer, this.scene.managerCommand.getProtocolProvider(layer.protocol));
 
         var map = this.scene.getMap();
-        var manager = this.scene.managerCommand;
-        var provider;
 
-        // get provider in function of provider
-        if(layer.protocol === 'wmts' || layer.protocol === 'wmtsc')
-            provider = this.getWMTSProvider();
-
-        provider.addLayer(layer);
-        var colorLayer = map.addColorLayer(layer.id)
-        manager.addLayer(colorLayer,provider);
-
+        map.tiles.colorLayers.push(layer);
+        map.colorTerrain.services.push(layer.id);
     };
 
     ApiGlobe.prototype.moveLayerUp = function(layer){
@@ -120,7 +125,6 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
 
         if(this.scene.getMap().removeColorLayer(id))
         {
-            this.getWMTSProvider().removeLayer(id);
             eventLayerRemoved.layer = id;
             this.viewerDiv.dispatchEvent(eventLayerRemoved);
             this.scene.renderScene3D();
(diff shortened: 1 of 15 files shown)


src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -130,6 +152,22 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
         return false;
     };
 
+
+    /**
+    * Add an elevation layer to the map. Elevations layers are used to build the terrain, if there is some overlapped the best resolution is taken, if resolution is equals, the first one is used.
```
Jeremy-Gaillard: "if there is some overlap the best resolution is taken, if the resolutions are equal, the first one is used"

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -122,29 +74,19 @@ define('Core/Commander/Providers/WMTS_Provider', [
 
         };
 
-        WMTS_Provider.prototype.removeLayer = function(idLayer)
+        WMTS_Provider.prototype.removeLayer = function(/*idLayer*/)
```
Jeremy-Gaillard, line 72: Is there a point in keeping this function?

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -318,48 +225,26 @@ define('Core/Commander/Providers/WMTS_Provider', [
...
             //var service;
-            var destination = command.paramsFunction.layer.description.style.layerTile;
+            var destination = command.paramsFunction.destination;
             var tile = command.requester;
 
-            if(destination === 1)
-            {
-                return this.getColorTextures(tile,command.paramsFunction.layer.services).then(function(result)
-                {
+            if(destination === 1) {
```
Jeremy-Gaillard, line 226: We should rename destination to something that makes more sense in this context. TextureType?

Test for textureType === "elevation" or textureType === "imagery" instead of 0/1.

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -318,48 +225,26 @@ define('Core/Commander/Providers/WMTS_Provider', [
...
+                });
             }
         };
 
 
         WMTS_Provider.prototype.getZoomAncestor = function(tile,layer) {
 
             var levelParent = tile.getLevelNotDownScaled();
-            return (levelParent < layer.zoom.min ? tile.level : levelParent) + (layer.tileMatrixSet === 'PM' ? 1 : 0);
+            return (levelParent < layer.zoom.min ? tile.level : levelParent) + (layer.options.tileMatrixSet === 'PM' ? 1 : 0);
```
Jeremy-Gaillard, line 242: Replace the test by a levelOffset attribute in the layer options?

src/Scene/NodeProcess.js
```diff
@@ -135,6 +197,67 @@ define('Scene/NodeProcess',
...
+                promises.push(quadtree.interCommand.request(args, node, refinementCommandCancellationFn));
+            }
+        }
+
+        return Promise.all(promises).then(function(colorTextures) {
+            var textures = [];
+            for (var j=0; j<colorTextures.length; j++) {
+                textures = textures.concat(colorTextures[j]);
+            }
+            node.setTexturesLayer(textures, 1);
```
Jeremy-Gaillard, line 213: Wouldn't node.setTexturesLayer(colorTextures, 1) achieve the same result? Without the need to create a new array.

I think we should also split setTexturesLayer into setElevationTextures and setColorTextures to get rid of these annoying "0" and "1".

src/Scene/Quadtree.js
```diff
@@ -50,6 +50,9 @@ define('Scene/Quadtree', [
         this.tileType = type;
         this.minLevel = 2;
         this.maxLevel = 17;
+        this.wgs84TileLayer = null;
+        this.colorLayers = [];
+        this.elevationLayers = [];
```
Jeremy-Gaillard: These attributes should be moved elsewhere (maybe at the globe level?), the quadtree might be used for other data types (point clouds, WFS data...).

src/Core/Commander/Providers/WMTS_Provider.js
peppsac, line 72: Good catch. I'll remove it

src/Core/Commander/Providers/WMTS_Provider.js
peppsac, line 242: Which test?

Jeremy-Gaillard, line 242: (layer.options.tileMatrixSet === 'PM' ? 1 : 0)

src/Scene/NodeProcess.js
peppsac, line 213: `colorTextures` is an array (because of `Promise.all(promises)`) of textures (if the Provider returns a single Promise) or array of textures (is the Provider returns a Promise.all of multiple promises).
Hence the need to concat the various textures in a single array.

Regarding your second suggestion: it could be a good idea. Or at the very least change the name of the function to get rid of 'Layer'

src/Scene/Quadtree.js
peppsac: Yeah, they won't last long here.

Currently NodeProcess knows about these layer (e.g: `for (j = 0; j < quadtree.colorLayers.length; j++) {` so if you move them straight to Globe you'll add a dependency between NodeProcess and Globe.

I can do this in this PR, but I'd prefer to get this one merged, and do another PR later to fix this issue.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
peppsac: This comment comes from the old code and is wrong :)
I'll fix it.

src/Scene/Quadtree.js
peppsac: I'm going to push another commit that moves layer knowledge and configuration elsewhere.

peppsac:
I updated the branch with a new object `LayersConfiguration`, owned by `Globe`.
This object holds everything related to layers and replace `services` and most of the job done by `Layer`

gchoqueux:
Nice job
- Remove console  log
- I would like keep the dead code (WMTS) in appropriate object (I do it, if you want) 
- Why you parse all providers to get the needed provider, why doesn't you use map 'protocol-provider'
- Can you comment `ApiGlobe.prototype.preprocessLayer`
- I imagine you want merge `updateNodeImagery` and `updateNodeElevation` in one function like `updateNodeRasterData` or 'updateNode'? because the mechanism is the same 
- Can you confirm if the creation of the mesh through a command (in nodeProcess)
- Why in `TileProvider` the tile is create visible?

gchoqueux:
Bug : all levels WMTS aren't displayed, the load is limited to a certain level

peppsac:
> Remove console log

Done (a5ff1a7)

>  I would like keep the dead code (WMTS) in appropriate object (I do it, if you want)

I'd rather not: dead code adds no value but adds maintenance cost.
If you ever need to get it back you'll simply revert commit f0304eb

> Why you parse all providers to get the needed provider, why doesn't you use map 'protocol-provider'

Changed ManagerCommand.providers to be a protocol to provider map (ac48014)

> Can you comment ApiGlobe.prototype.preprocessLayer

Done (ac48014)

peppsac:
>    I imagine you want merge updateNodeImagery and updateNodeElevation in one function like updateNodeRasterData or 'updateNode'? because the mechanism is the same

They're similar but not identical. IMHO they could probably be merged in the future, but not in this PR.

>    Can you confirm if the creation of the mesh through a command (in nodeProcess)

I don't understand this question :-/

>    Why in TileProvider the tile is create visible?

Good catch, I need to make sure it's still needed (was needed at some point to fix a bug)

peppsac:
> Bug : all levels WMTS aren't displayed, the load is limited to a certain level

Can you be more specific? Or post a screenshot? Or steps to reproduce?

vpicavet:
-1 for keeping dead code : removing code is the best way to improve a software, no code no bug no maintainance.
If this part needs some improvement or reuse of some sort, open an issue and refer to the commit corresponding to the removal.

gchoqueux:
I doesn't want to keep dead code, but transform it in live code

gchoqueux:
>    Why in TileProvider the tile is create visible?
> Good catch, I need to make sure it's still needed (was needed at some point to fix a bug)

the bug is fixed?

peppsac:
> the bug is fixed?

Yes, so there's no need to call `setVisible(true)` by default

gchoqueux:
> Bug : all levels WMTS aren't displayed, the load is limited to a certain level
> Can you be more specific? Or post a screenshot? Or steps to reproduce?

The wmts stops to load

refacto_providers_layers : there isn't loading activity 
capture du 2016-07-13 11-02-22

Master branch:
capture du 2016-07-13 11-03-40

gchoqueux:
> Can you confirm if the creation of the mesh through a command (in nodeProcess)
> I don't understand this question :-/

The tileMesh is created by a command?

peppsac:
> The tileMesh is created by a command?

Yes, see: https://github.com/iTowns/itowns2/blob/refacto_providers_layers/src/Scene/NodeProcess.js#L92

gchoqueux:
> Can you comment ApiGlobe.prototype.preprocessLayer

what is the preprocess?

peppsac:
Ok, missing subdivision level bug fixed

peppsac:
> what is the preprocess?

It's simply a mean to give Providers a chance to do some precomputation on a layer.
WMTS_Provider uses it to compute customUrl or zoom attributes

gchoqueux:
Ok it needs just a rebase

peppsac:
Cleaned & rebased!

gchoqueux:
r+

PR #150, opened 2016-07-12
Wms support

peppsac:
WMS support based on the work done by @qdnguyen 

Depends on PR #149

### DIFF
--- a/index.html
+++ b/index.html
@@ -384,6 +384,27 @@ and open the template in the editor.
         //     fx :        2.5,
         // });
 
+        itowns.viewer.addImageryLayer({
+                                        url       : "https://wxs.ign.fr/gratuit/geoportail/v/wms",
+                                        protocol  : 'wms',
+                                        version   : '1.3.0',
+                                        id        : 'testwms',
+                                        name      : 'REGION.2016',
+                                        style      : "",
+                                        projection : "EPSG:4326",
+                                        transparent : true,
+                                        // min long, min lat, max long, max lat
+                                        bbox      : [-61.80983300000002, -21.38963079935857, 55.83665389999998, 51.0890011990783],
+                                        featureInfoMimeType : "",
+                                        dateTime  : "",
+                                        heightMapWidth : 256,
+                                        waterMask      : false,
+                                        options: {
+                                            mimetype  : "image/png",
+                                            tileMatrixSet: 'WGS84G'
+                                        }
+                                    });
+
         itowns.viewer.addElevationLayer({
             protocol:   "wmts",
             id:         "IGN_MNT",
--- a/src/Core/Commander/Providers/IoDriver_Image.js
+++ b/src/Core/Commander/Providers/IoDriver_Image.js
@@ -35,11 +35,9 @@ IoDriver_Image.prototype.read = function(url) {
 
 
         image.addEventListener('error', function( /*event*/ ) {
-
-            this.src = '';
+            // TODO: real error managament
             resolve(undefined);
-
-        }.bind(this), false);
+        }, false);
 
         image.crossOrigin = '';
         image.src = url;
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -11,6 +11,7 @@ import IoDriver_Image from 'Core/Commander/Providers/IoDriver_Image';
 import IoDriverXML from 'Core/Commander/Providers/IoDriverXML';
 import defaultValue from 'Core/defaultValue';
 import THREE from 'THREE';
+import Projection from 'Core/Geographic/Projection';
 import CacheRessource from 'Core/Commander/Providers/CacheRessource';
 import BoundingBox from 'Scene/BoundingBox';
 
@@ -27,6 +28,7 @@ function WMS_Provider(/*options*/) {
     this.cache = CacheRessource();
     this.ioDriverImage = new IoDriver_Image();
     this.ioDriverXML = new IoDriverXML();
+    this.projection = new Projection();
 
     this.getTextureFloat = function(buffer){
         // Start float to RGBA uint8
@@ -95,15 +97,22 @@ WMS_Provider.prototype.tileInsideLimit = function(tile,layer) {
     return bboxRegion.intersect(bbox);
 };
 
-WMS_Provider.prototype.getColorTexture = function(tile, layer) {
+WMS_Provider.prototype.getColorTexture = function(tile, layer, parameters) {
     if (!this.tileInsideLimit(tile,layer) || tile.material === null) {
         return Promise.resolve();
     }
 
-    var result = {pitch : new THREE.Vector3(0, 0, 1)};
+    var pitch = parameters.ancestor ?
+        this.projection.WMS_WGS84Parent(tile.bbox, parameters.ancestor.bbox) :
+        new THREE.Vector3(0, 0, 1);
 
-    var url = this.url(tile.bbox, layer);
+    var bbox = parameters.ancestor ?
+        parameters.ancestor.bbox :
+        tile.bbox;
 
+    var url = this.url(bbox, layer);
+
+    var result = { pitch };
     result.texture = this.cache.getRessource(url);
 
     if (result.texture !== undefined) {
@@ -132,7 +141,6 @@ WMS_Provider.prototype.getColorTexture = function(tile, layer) {
 
     }.bind(this)).catch(function(/*reason*/) {
             result.texture = null;
-
             return result;
         });
 
@@ -193,7 +201,7 @@ WMS_Provider.prototype.executeCommand = function(command) {
 
     var func = supportedFormats[layer.format];
     if (func) {
-        return func(tile, layer).then(function(result) {
+        return func(tile, layer, command.paramsFunction).then(function(result) {
             return command.resolve(result);
         });
     } else {
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -12,6 +12,7 @@ import IoDriverXML from 'Core/Commander/Providers/IoDriverXML';
 import defaultValue from 'Core/defaultValue';
 import THREE from 'THREE';
 import CacheRessource from 'Core/Commander/Providers/CacheRessource';
+import BoundingBox from 'Scene/BoundingBox';
 
 /**
  * Return url wmts MNT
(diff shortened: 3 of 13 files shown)


qdnguyen:
IO_driver selection in providers should be more generic?  #157

peppsac:
PR ready for review.

gchoqueux:
OK i review

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -30,39 +32,235 @@ define('Core/Commander/Providers/WMS_Provider', [
...
             this.ioDriverImage = new IoDriver_Image();
             this.ioDriverXML = new IoDriverXML();
 
-            this.baseUrl = options.url || "";
-            this.layer = options.layer || "";
-            this.format = defaultValue(options.format, "image/jpeg");
-            this.srs = options.srs || "";
-            this.width = defaultValue(options.width, 256);
-            this.height = defaultValue(options.height, 256);
+            this._ready       = false;
```
gchoqueux: `this._ready` is unused

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -30,39 +32,235 @@ define('Core/Commander/Providers/WMS_Provider', [
...
+        };
+        WMS_Provider.prototype.customUrl = function(url,coord)
+        {
+
+            //convert radian to degree, lon is added a offset of Pi
+            //to align axis to card center
+            //var bbox = (coord.minCarto.longitude - Math.PI)* 180 / Math.PI + "," + coord.minCarto.latitude* 180 / Math.PI + "," +
+            //           (coord.maxCarto.longitude - Math.PI )*180/ Math.PI + "," + coord.maxCarto.latitude* 180 / Math.PI;
+
+            var bbox = coord.minCarto.latitude * 180.0 / Math.PI +
```
gchoqueux: The coord to BBox must be put into the Core/Geographic
or
Understand, why coord are incorrect.

gchoqueux:
Good Feature
- `WMS_Provider` and `WMTS_Provider` will need to factor. there is a lot of redundancy between the two objects. We can imagine RasterDataTileProvider.

src/Core/Math/Rectangle.js
```diff
@@ -3,14 +3,61 @@
  * Class: Rectangle
  */
 
-function Rectangle() {
```
gchoqueux, line 6: `Scene/BoundingBox` is very similar to `Core/Math/Rectangle`
It needs to merge in one object

src/Core/Commander/Providers/WMS_Provider.js
qdnguyen: I dont understand what you mean but this coord is given by tileProvider and i transform it to map center which is origin of wms layers.

gchoqueux:
Why there are `urlGlobalIR` and `urlClouds` in WMS
must be removed and replace by custom `WMS_Provider` like WMTS_Provider's example with openStreetMap

src/Core/Math/Rectangle.js
qdnguyen, line 6: rectangle is used in region check with only 4 coords: west, east, south, north. It's faster than bbox check. 
From rectangle we can have the bbox in 2D but in theory rectangle is not bbox. I refer to keep them separately to make code more clear.

qdnguyen: Instead of RasterDataTileProvider, we can also select io_driver according to input info when layer is added? so that providers contain only an instance of io_driver?  regardless of data format?

qdnguyen:
ok, i have cleaned urlGlobalIR and urlClouds, we can request image from these servers with our new wms_provider.

src/Core/Commander/Providers/WMS_Provider.js
qdnguyen: this variable was deleted in commit 0c23027

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux:
``` js
var bbox = coord.minCarto.latitude * 180.0 / Math.PI +
"," +
(coord.minCarto.longitude - Math.PI)* 180.0 / Math.PI +
","+
coord.maxCarto.latitude* 180.0 / Math.PI +
"," +
(coord.maxCarto.longitude - Math.PI )*180.0 / Math.PI;
```

This part must to move in Core and the question is 'Why we need to convert longitude `(coord.maxCarto.longitude - Math.PI )`'. May be we should correct longitude in tile

qdnguyen: As i said, origin of wms layers is at map center, because in WGS84 the prime meridian pass through GreenWich, which is not the case in our itowns by now. I have open an issue some months ago but it's not resolved yet. This problem touch to every module, so for now we can not move it to core, but one time core is modified, we can delete this code. #19

qdnguyen:
> WMS_Provider and WMTS_Provider will need to factor. there is a lot of redundancy between the two objects. We can imagine RasterDataTileProvider.

Instead of RasterDataTileProvider, we can also select io_driver according to input info when layer is added? so that providers contain only an instance of io_driver? regardless of data format? But this work should be done after merged because it touch to many providers, not only wms_provider.

src/Core/Commander/Providers/WMS_Provider.js
peppsac: If issue #19 is the sole reason why you transform `longitude` please add a comment in the code:

``` js
// shifting longitude because of issue #19
```

And please remove dead code.

qdnguyen: done!

gchoqueux:
there is a display artifact  (because of the lens texture)
Can you remove it?

capture du 2016-07-19 11-43-54

qdnguyen:
i dont know where it comes from, but it's not from wms module!

gchoqueux:
> i dont know where it comes from, but it's not from wms module!

->

> because of the lens texture

There isn't this bug in master

gchoqueux:
There's white line on equator
capture du 2016-07-21 13-51-44

peppsac:
I pushed a fix for the lensflare issue.

peppsac:
I fixed the white line bug and  have rebased the branch on master.
@qdnguyen do you mind if I force push this branch?

qdnguyen:
yes, do it, thanks!

peppsac:
Done.
@gchoqueux is there something left that prevents merging this branch?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -30,71 +32,221 @@ define('Core/Commander/Providers/WMS_Provider', [
...
 
-            this.baseUrl = options.url || "";
-            this.layer = options.layer || "";
-            this.format = defaultValue(options.format, "image/jpeg");
-            this.srs = options.srs || "";
-            this.width = defaultValue(options.width, 256);
-            this.height = defaultValue(options.height, 256);
+            this.getTextureFloat;
+
+            if(this.support)
```
Jeremy-Gaillard: this.support is always undefined.

gchoqueux:
There's a problem with wms textures?
capture du 2016-07-26 12-08-06

peppsac:
It's an error handling issue, and it's not specific to WMS.
The example WMS server returns an error sometimes, and itowns doesn't handle this very well atm.

src/Core/Commander/Providers/WMS_Provider.js
peppsac: did you push your change? (I don't see it)

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -30,71 +32,221 @@ define('Core/Commander/Providers/WMS_Provider', [
...
+        WMS_Provider.prototype.tileInsideLimit = function(tile,layer) {
+                var bbox = tile.bbox;
+                // shifting longitude because of issue #19
+                var west =  layer.bbox[0]*Math.PI/180.0 + Math.PI;
+                var east =  layer.bbox[2]*Math.PI/180.0 + Math.PI;
+                var bboxRegion = new BoundingBox(west, east, layer.bbox[1]*Math.PI/180.0, layer.bbox[3]*Math.PI/180.0, 0, 0, 0);
+            return bboxRegion.intersect(bbox);
+        };
+
+        WMS_Provider.prototype.getColorTextures = function(tile, layer) {
```
peppsac: This function is not needed, please get rid of it.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -30,71 +32,221 @@ define('Core/Commander/Providers/WMS_Provider', [
...
+
+            // bug #74
+            //var limits = layer.tileMatrixSetLimits[coWMTS.zoom];
+            // if (!limits || !coWMTS.isInside(limits)) {
+            //     var texture = -1;
+            //     this.cache.addRessource(url, texture);
+            //     return Promise.resolve(texture);
+            // }
+            // -> bug #74
+
```
peppsac: Remove dead code please.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -30,71 +32,221 @@ define('Core/Commander/Providers/WMS_Provider', [
...
+            }.bind(this));
         };
 
 
+        WMS_Provider.prototype.executeCommand = function(command){
+            //var service;
+            var destination = command.paramsFunction.destination;
+            var tile = command.requester;
+
+            if(destination === 1) {
```
peppsac: IMHO the test should depend solely on `layer.mimetype`:
- if it's a color image format (eg: png, jpg) use `getColorTexture`
- if it's an `xbil` format then use `getElevationTexture` (which should probably have a different name).

src/Core/Commander/Providers/WMS_Provider.js
qdnguyen: it is needed in preprocessLayer function of ApiGlobe module

peppsac: I meant the `getColorTextures` function

qdnguyen: for data selection in providers, i have created an issue #157 but we must do it right now in this PR? because we must do for all providers, not only wms_provider.

src/Core/Commander/Providers/WMS_Provider.js
qdnguyen: Issue #157

peppsac:
Rebased on master

peppsac:
And added a workaround for an error in IoDriver_Image.
@gchoqueux : r?

qdnguyen:
@peppsac  did you rebase this PR on master? i did clean up some codes but i have not commited yet.

peppsac:
Yes it's rebased.

qdnguyen:
ok, it was not a big thing, please push it :)

src/Core/Commander/Providers/IoDriver_Image.js
```diff
@@ -35,11 +35,9 @@ IoDriver_Image.prototype.read = function(url) {
 
 
         image.addEventListener('error', function( /*event*/ ) {
-
-            this.src = '';
+            // TODO: real error managament
```
tbroyer, line 38: See #170

index.html
```diff
@@ -384,6 +384,27 @@
         //     fx :        2.5,
         // });
 
+        itowns.viewer.addImageryLayer({
+                                        url       : "https://wxs.ign.fr/gratuit/geoportail/v/wms?",
```
tbroyer: `WMS_Provider` uses `layer.url + "?SERVICE=WMS…"` which results in a doubled `??` in the URL. That doesn't seem to cause problems to the server but is probably not what was intended.

I don't know WMS enough but either the `?` here should be removed, or `WMS_Provider` should look at whether `layer.url` contains a `?` already and conditionally append with a `?` or a `&` (or possibly no separator if `layer.url` ends with one of those)

gchoqueux:
All levels of WMS are loaded. WMTS uses ancestors, it isn't the same things for WMS?
It's better to load only the necessary levels.

gchoqueux:
What is the policy to display WMS in function of level's tile?
It's possible to set this rule?

gchoqueux:
If you add only wms in index.html, the navigator don't stop to request WMS images.

qdnguyen:
the notion of level is not exist in wms, the service generates the level according to bbox, i think we can limit LODs by checking bbox size but i'm not sur about this, may be @gcebelieu or @elias75015 can reply? Anyway, without getcapabilities, we have less infos!

p/s: the downsload is constrained by checking west, east, south, north but i dont think it's enough to limit LODs

gchoqueux:
in 'Globe Loaded' Event, some requests WMS are launched and nothing is displayed. why?
All queries fail, it's possible to test before launch a request?

gchoqueux:
> the notion of level is not exist in wms

Yes, it's true
But the WMS images are linked with tiles and these tiles have a level.
WMS_provider cuts in tiles and it's the same WMTS's mechanism.
It must be possible to use the textures's ancestors

qdnguyen:
> But the WMS images are linked with tiles and these tiles have a level.
> WMS_provider cuts in tiles and it's the same WMTS's mechanism.

For now, wms depends on wmts tile bbox, but in general, you can request it with any bbox which will be taken into account by wms to generate LODs. 

> It must be possible to use the textures's ancestors

what do you mean 'texture's ancestors'? is it a quadtree? yes, we can create a quadtree for wms which cut tile with a specifique tilingScheme. And maxLevel, minLevel of quadtree can be defined by getcap?

qdnguyen:
> in 'Globe Loaded' some requests WMS are launched and nothing is displayed. why?

I dont understand your question, wms layers are added through  itowns.viewer.addImageryLayer, not addEventListener 'Globe Loaded'

peppsac:
Regarding,

> If you add only wms in index.html, the navigator don't stop to request WMS images.

and 

> What is the policy to display WMS in function of level's tile?

Actually providers shouldn't be aware of `level` at all - `NodeProcess` should probably take care of this.

And the issue you mention is probably caused by this very issue, as this code is used to determine if we should download another texture:

``` js
var level = this.Textures[id][0].level;
```

I'll work on another PR to land before this one, to move all this `level` management to `NodeProcess`.
(it's also related to error management, because as you can see only the first texture is tested to determine if we need to refine the color layer)

gchoqueux:
WMS_SUPPORT2 Warning console
- /src/Scene/NodeProcess.js  307:33  error  'err' is defined but never used  no-unused-vars
- /src/Renderer/LayeredMaterial.js
  - 294:5  error  Unreachable code  no-unreachable
  - 295:9  error  Unreachable code  no-unreachable
  - 297:9  error  Unreachable code  no-unreachable

gchoqueux:
> in 'Globe Loaded' some requests WMS are launched and nothing is displayed. why?

In start loading, There's WMS requests. They return transparent texture. Perhaps, it needs to crop bbox's request with global bbox. I will investigate on the subject

peppsac:
> In start loading, There's WMS requests. They return transparent texture. Perhaps, it needs to crop bbox's request with global bbox.

The bbox is big because it covers France + overseas territory.
But the server returns non-transparent texture only for smaller bbox (level 4 or 5)

gchoqueux:
> But the server returns non-transparent texture only for smaller bbox (level 4 or 5)

The `MinScaleDenominator` and `MaxScaleDenominator` elements define the range of scales for which it is appropriate to generate a map of a Layer. 

But there's no info to scale denominator in GetCapabilities.

gchoqueux:
Nice : 
- WMS don't load all level texture.
- On the Start, there's no loop requesting

gchoqueux:
`TileMesh.prototype.changeSequenceLayers` doesn't work anymore. `sequence` is undefined.

peppsac:
Branch updated

gchoqueux:
r+
I create a new issue for 'scale denominator'

Issue #155, opened 2016-07-13
THREE.LineBasicMaterial() : 'linewidth' attribut doesn't work on Windows


vcoindet:
I tried to create lines with Three.js this way :

```
var material = new THREE.LineBasicMaterial({
       color: 0xffffff, opacity: 1, linewidth: 5
});

var line = new THREE.Line(geometry, material); 
```

But the attribut 'linewidth' doesn't have any effect on Windows, it's always returning a line with the width of 1px.
It seems that it's a limitation of ANGLE.

Is there any alternative way to change the width of the line ?

nosy-b:
Yes,
Three js got a tube geometry (but expensive as it creates many triangles!).
Or you can write a specific shader for a buffer geometry when we detect
non-native gl browser.

qdnguyen:
I propose that we should create our lineShaderMaterial, i may need this shader for wfs also, i can see this problem next week if you want!

vcoindet:
That would be great @qdnguyen !

qdnguyen:
@vcoindet this was done in my new commits, can you review it please?

gchoqueux:
A solution is added in PR #158

Issue #114, opened 2016-07-05
Add Gpx feature


gchoqueux:
- Create Api functions for gpx (add,remove,visibility,opacity)
- Create gpx provider

gchoqueux:
feature in add in merged PR #158

Issue #119, opened 2016-07-05
Develop Globe.getCurrentAltitude(longitude, latitude)


gchoqueux:
This function returns altitude of the current MNT displayed.

qdnguyen:
I need also this function to get alti for my wfs data #115 

![sans titre](https://cloud.githubusercontent.com/assets/3224540/17402654/9f04485a-5a54-11e6-97f7-003779769be4.png)

gchoqueux:
move in #118

PR #158, opened 2016-07-18
Gpx

vcoindet:
#114 :

GPX_Provider created : it parses GPX files (gets Waypoints and Tracks points) and creates Three geometies that can be added to the scene (points and lines). 
It can be called with the 'loadGPX' function wich is in ApiGlobe.js.

One bug is still to be fixed on Windows : 'linewidth' attribut doesn't have any effect on Windows (see #155 ).


nosy-b:
Super!
We'll check for LineWidth and other style representation for features soon. Do you have a screenshot of a GPX visualization?

vcoindet:
gpx

qdnguyen:
how do you activate this layer? i did not see on GUI?

vcoindet:
gpx2

vcoindet:
You have to call it in index.html

```
viewerDiv.addEventListener('globe-loaded',function(){
                    console.log("Globe Loaded");
                    itowns.viewer.loadGPX("URL_of_your_file");
                } ,
 false);
```

qdnguyen:
Can you publish the data in https://github.com/iTowns/iTowns2-sample-data ? juste one file to test.

otherwise, why the data is loaded in addEventListener? we can not have a provider like wfs?

vcoindet:
You don't have to put in an addEventListener for this one.
The event 'globe-loaded' is here for some other functions, like 'setCenter' where you have to wait for the globe to be full loaded or you will have some accuracy errors. Here it's not necessary.

qdnguyen:
But i dont understand how you deal with many files? or just only one file we need in itowns?

vcoindet:
For now, you can load only one file at the time, or call the function several times.

vcoindet:
I just found a new problem in my code : 
I'm using the tag "ele" in gpx files to get the elevation, but sometime this tag doesn't exist, so my algorithm doesn't work.

This will be fixed when 'Globe.getCurrentAltitude(longitude, latitude)' is done. (see issue #119)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -639,6 +641,18 @@ define('Core/Commander/Interfaces/ApiInterface/ApiGlobe', [
     };
 
 
+    ApiGlobe.prototype.loadGPX = function(url) {
+        //providerGPX should be instanced elsewhere
```
peppsac: > -        //providerGPX should be instanced elsewhere 

Indeed. All other providers are instanced in `createSceneGlobe`

src/Core/Commander/Providers/GPX_Provider.js
```diff
@@ -0,0 +1,75 @@
...
+                // ------------------------------------
+                // Getting the waypoint points
+                // ------------------------------------
+                var wpt = result.getElementsByTagName("wpt");
+                var geometry_p = new THREE.Geometry();
+
+                for (var i = 0; i < wpt.length; i++)
+                    geometry_p.vertices.push(this.ellipsoid.cartographicToCartesian(new CoordCarto().setFromDegreeGeo(Number(wpt[i].attributes.lon.nodeValue),Number(wpt[i].attributes.lat.nodeValue),Number(wpt[i].getElementsByTagName("ele")[0].childNodes[0].nodeValue))));
+
+                var material_p = new THREE.PointsMaterial( {color: 0x00ff00, size: 50, sizeAttenuation : true} );
```
peppsac: You probably don't need to create 1 PointMaterial per point. 
Suggestion: move Materials creation before the `for (var i = 0; i < wpt.length; i++)` loop

peppsac: Oops nevermind I misread.
New coment : please use `{ }` for each block :)

src/Core/Commander/Providers/GPX_Provider.js
```diff
@@ -0,0 +1,75 @@
...
+
+                var point = new THREE.Points( geometry_p, material_p );
+
+                // ------------------------------------
+                //Getting the track points
+                // ------------------------------------
+                var trkpt = result.getElementsByTagName("trkpt");
+                var geometry = new THREE.Geometry();
+
+                for (var j = 0; j < trkpt.length; j++)
```
peppsac: Add braces `{ }`

src/Core/Commander/Providers/GPX_Provider.js
```diff
@@ -0,0 +1,75 @@
...
+                var line = new THREE.Line(geometry, material);
+                line.add(point);
+
+                return line;
+            }.bind(this));
+
+        };
+
+        return GPX_Provider;
+
```
peppsac: Lots of unneeded blank lines in this file. Can you remove them please?

peppsac:
What about making a real Provider (one that can execute commands)?

Alternatively : split the parseGpx function to a GpxUtil.js file and do:

``` js
xmlDriver.read(url).then(function(xml) {
    this.scene.getMap().gpxTracks.children[0].add(GPXUtils.GpxToGeometry(xml));
}.bind(this));
```

in Globe.js

src/Core/Commander/Providers/GPX_Provider.js
qdnguyen: i think you should use group = new Object3D and return group instead of line.

qdnguyen:
finaly, we can see line width styled with texture under Angle. 
autoroute1

qdnguyen:
autoroute1

src/Core/Commander/Providers/GPX_Provider.js
qdnguyen: done in my code!

qdnguyen:
add ItownsPoint module, i will use this for potree also!
bb

gchoqueux:
GPX is broken, for the moment

qdnguyen:
because my last commit is changed?

gchoqueux:
r?

peppsac:
- Indentation is incorrect in several places
- Missing explanations about _why_ you need to reimplement Points/Lines
- No answer to my comment about the GpxProvider: https://github.com/iTowns/itowns2/pull/158#issuecomment-234537490

qdnguyen:
> Indentation is incorrect in several places

Can you capture a photo please? at some zoom level, line is disappeared but i dont know why? 

> Missing explanations about why you need to reimplement Points/Lines

For line, you can see some explainations here #155. Another reason I styled line is that we can add street name as what they did in google street view :

sans titre

For points, we dont need to create a specifique shader to show texture but i did it because i will extend this shader for potree also. 

> No answer to my comment about the GpxProvider: #158 (comment)

I leave @vcoindet to reply you this question because it relate to his work but he is in vacation now.

gchoqueux:
I agree this comment : https://github.com/iTowns/itowns2/pull/158#issuecomment-234537490
I'm working on it.

gchoqueux:
> I agree this comment : https://github.com/iTowns/itowns2/pull/158#issuecomment-234537490

Update branch : Recast Gpx provider

peppsac:
I pushed a commit that take advantages of modules.
The GpxUtils module now exports a single function `loadGpx` and all others are private.

Another question : GpxUtils.js should probably move elsewhere (`src/utils/` ?) but this can wait until we have at least two _utils_ modules.

PR #163, opened 2016-07-19
Handle errors when reading a json object inside the IODriver_JSON

mlamure:
What kind of message can be thrown when the error is caught by the catch statement ?

Right now, there is a console statement which create an error : Unexpected console statement  no-console.

### DIFF
--- a/src/Core/Commander/Providers/IoDriver_JSON.js
+++ b/src/Core/Commander/Providers/IoDriver_JSON.js
@@ -20,6 +20,15 @@ define('Core/Commander/Providers/IoDriver_JSON', ['Core/Commander/Providers/IoDr
     IoDriver_JSON.prototype.read = function(url) {
         return fetch(url).then(function(response) {
             return response.json().then(
+                function(response) {
+                    if(response.status >= 200 && response.status < 300)
+                        return response;
+                    else{
+                        var error = new Error(response.statusText);
+                        error.response = response;
+                        throw error;
+                    }
+                }).then(
                 function(data){
                     return data;
                 }).catch(function(error) {
--- a/src/Core/Commander/Providers/IoDriver_JSON.js
+++ b/src/Core/Commander/Providers/IoDriver_JSON.js
@@ -19,7 +19,12 @@ define('Core/Commander/Providers/IoDriver_JSON', ['Core/Commander/Providers/IoDr
 
     IoDriver_JSON.prototype.read = function(url) {
         return fetch(url).then(function(response) {
-            return response.json();
+            return response.json().then(
+                function(data){
+                    return data;
+                }).catch(function(error) {
+                    console.log('request failed', error);
+                });
         });
     };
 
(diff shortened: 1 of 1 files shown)


tbroyer:
Well, the real question is: how do we want to handle the error?

(also note that `IoDriver_JSON` does not check the HTTP status code of the response)

Currently, the error handling is left to the caller and "bubbles up" all the way up to `MobileMappingLayer` (through `WFS_Provider`, `BuildingBox_Provider`, and `PanoramicProvider`).

IMO, there needs to be a global reflection on error handling, not just for that specific case.

mlamure:
I will had a test for the HTTP status in the IODriver_JSON inside my branch.

I can open an issue to discuss the 'global reflection on error' matter if you want.

src/Core/Commander/Providers/IoDriver_JSON.js
```diff
@@ -20,6 +20,15 @@ define('Core/Commander/Providers/IoDriver_JSON', ['Core/Commander/Providers/IoDr
     IoDriver_JSON.prototype.read = function(url) {
         return fetch(url).then(function(response) {
             return response.json().then(
+                function(response) {
+                    if(response.status >= 200 && response.status < 300)
```
tbroyer: This should be done **before** calling `.json()`. Something like:

``` js
return fetch(url).then(function(response) {
  if (response.status >= 200 && response.status < 300) {
    return response.json();
  } else {
    // throw
  }
})
```

src/Core/Commander/Providers/IoDriver_JSON.js
```diff
@@ -19,7 +19,21 @@ define('Core/Commander/Providers/IoDriver_JSON', ['Core/Commander/Providers/IoDr
...
+                    else{
+                        var error = new Error(response.statusText);
+                        error.response = response;
+                        throw error;
+                    }
+                }).then(
+                function(data){
+                    return data;
+                }).catch(function(error) {
+                    console.log('request failed', error);
```
tbroyer: Also, note that this means that the method returns a _fulfilled_ `Promise` (of `undefined`), so it would be complicated for callers of this method to handle errors.

I strongly believe that the error should not be caught here, and we should leave error handling to the "end" of the chain as much as possible (`MobileMappingLayer` currently).

Note that when a promise has been _rejected_, all the `then()` are skipped; only the next `catch()` (or `onRejected` callback passed as a second argument tp a `then()`) will be called, so all intermediary steps can assume things went well in their `then()` (this is assuming this change is not merged, where each `then()` would have to check whether the argument it receives is `undefined`)

mlamure: OK, I had not understand what you meant before.
I will try to put the catch statement at the end of the chain and test the status code before the json() function.

mlamure:
I did try the code I just commit with my feature provider and I had no problem to catch the error at the end of the chain.
I still don't know what to put inside the catch statement. Do you have any idea ?

tbroyer:
Superseded by #170 

wrt the final `catch()` to properly handle errors; `WMTS_Provider` uses a `-1` texture for example, and I think we shouldn't do anything if we don't know what to do (Chrome will log it; unfortunately not Firefox); at least for now!

qdnguyen:
is this PR still active? otherwise we can close it!

PR #168, opened 2016-07-25
Replace size by ellipsoid in most constructors.

Jeremy-Gaillard:
Size is too generic of a term and handing an ellipsoid to a class has far more sense than handing a 3-dimensional vector.

This work will also help the future implementation of the planar mode.


peppsac:
LGTM

PR #169, opened 2016-07-25
Move from AMD and CommonJS to ES6 modules

tbroyer:


tbroyer:
Note: A few things have an `instanceof` check, that mrdoob/three.js#9310 said wasn't supported. My quick smoketest didn't show any issue, but that code might not have been exercised and could need to be reworked.

peppsac:
Seems to work well and this PR really improves code readability: so :+1: 

I should have merged this one before merging PR #168 - sorry. Can you update it again and I'll merge it?

tbroyer:
Rebased and reprocessed.

src/Core/Commander/Providers/IoDriver_Image.js
```diff
@@ -4,49 +4,47 @@
...
-            image.addEventListener('error', function(/*event*/) {
+        image.addEventListener('error', function( /*event*/ ) {
 
-                this.src = '';
-                resolve(undefined);
+            this.src = '';
+            resolve(undefined);
 
-            }.bind(this), false);
+        }.bind(this), false);
```
peppsac, line 42: `this` is now undefined (on master `this` was `Window`).

We should probably double-check all the `.bind(this)` patterns to make sure they are needed and working as expected.

tbroyer, line 42: If `this` was `Window` or the global object, then it was **not** working as expected, as the function actually expected `image`. It should have probably used `event.eventTarget.src = ''` or `image.src = ''`.

I'll have a (quick) look at the 43 `.bind(this)` we have in the code (not that much actually, I would have thought there'd be more of them).

peppsac, line 42: > it was not working as expected

Yep, I agree

> I'll have a (quick) look at the 43 .bind(this)

Great, thanks

tbroyer, line 42: Given the `resolve(undefined)` that should be a `reject(new Error("…"))`, I'd rather fix this in a subsequent PR if you don't mind. It's not an issue with this PR (there's another `this` issue in `Ori` where `onload` is _missing_ a `bind(this)`). The rest looks OK.

peppsac:
Let's merge this branch - @tbroyer would you mind updating it again?

peppsac:
Thanks, merging.

PR #170, opened 2016-07-27
Network error handling

tbroyer:


peppsac:
Good work.
We need to make sure that Error are properly handled, mainly in NodeProcess.
But maybe you already checked this?

tbroyer:
> But maybe you already checked this?

No, this is only a first improvement step to propagate errors (rather than actually "handle" them). I updated `WMTS_Provider` as I changed the way it communicated the error (from `undefined` to a rejected promise), but that's it.

I quickly tested and Chrome didn't log about unhandled rejected promises (this is how I found out about `WMTS_Provider`), but I probably didn't have HTTP errors either.
(note: Firefox unfortunately doesn't say anything when rejected promises are left unhandled)

peppsac:
Ok.

As soon as PR #150 is merged it'll be easier to test: the WMS server regularly produce errors.

tbroyer:
I merged the branch locally, resolving the conflict on `IoDriver_Image` to use this branch, and things went OK (errors from `IoDriver_Image` when the server responds with an XML error message) are properly handled in `WMS_Provider` to `null` the `result.texture`.

peppsac:
This PR looks good. Can you rebase it?

Issue #165, opened 2016-07-21
Freeze layer


gchoqueux:
Add Freeze property in Layer
Freeze is boolean. If Freeze equals false, the Layer can't be queried at Provider.
For example a WMTS Layer stops loading textures.

peppsac:
Will be implemented in PR #176

PR #176, opened 2016-08-03
Cleanup and elevation

peppsac:
3 commits:
- the first one simplifies frustum culling code
- the 2nd makes TileMesh.distance computation more generic
- the last one clean up elevation texture handling

### DIFF
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -97,18 +97,10 @@ WMS_Provider.prototype.tileInsideLimit = function(tile,layer) {
     return tile.level > 2 && layer.bbox.intersect(tile.bbox);
 };
 
-WMS_Provider.prototype.getColorTexture = function(tile, layer, parameters) {
+WMS_Provider.prototype.getColorTexture = function(tile, layer, bbox, pitch) {
     if (!this.tileInsideLimit(tile,layer) || tile.material === null) {
         return Promise.resolve();
     }
-
-    var pitch = parameters.ancestor ?
-        this.projection.WMS_WGS84Parent(tile.bbox, parameters.ancestor.bbox) :
-        new THREE.Vector3(0, 0, 1);
-
-    var bbox = parameters.ancestor ?
-        parameters.ancestor.bbox :
-        tile.bbox;
 
     var url = this.url(bbox, layer);
 
@@ -140,20 +132,26 @@ WMS_Provider.prototype.getColorTexture = function(tile, layer, parameters) {
         return result;
 
     }.bind(this)).catch(function(/*reason*/) {
-            result.texture = null;
-            return result;
-        });
+        result.texture = null;
+        return result;
+    });
 
 };
 
-WMS_Provider.prototype.getXbilTexture = function(tile, layer) {
-    var url = this.url(tile.bbox, layer);
+WMS_Provider.prototype.getXbilTexture = function(tile, layer, bbox, pitch) {
+    var url = this.url(bbox, layer);
 
     // TODO: this is not optimal: if called again before the IoDriver resolves, it'll load the XBIL again
     var textureCache = this.cache.getRessource(url);
 
-    if (textureCache !== undefined)
-        return Promise.resolve(textureCache);
+    if (textureCache !== undefined) {
+        return Promise.resolve(textureCache ? {
+            pitch,
+            texture: textureCache.texture,
+            min: textureCache.min,
+            max: textureCache.max
+        } : null);
+    }
 
 
     // bug #74
@@ -167,13 +165,11 @@ WMS_Provider.prototype.getXbilTexture = function(tile, layer) {
 
     return this._IoDriver.read(url).then(function(result) {
         if (result !== undefined) {
-
-            //TODO USE CACHE HERE ???
-
             result.texture = this.getTextureFloat(result.floatArray);
             result.texture.generateMipmaps = false;
             result.texture.magFilter = THREE.LinearFilter;
             result.texture.minFilter = THREE.LinearFilter;
+            result.pitch = pitch;
 
             // In RGBA elevation texture LinearFilter give some errors with nodata value.
             // need to rewrite sample function in shader
@@ -181,7 +177,7 @@ WMS_Provider.prototype.getXbilTexture = function(tile, layer) {
 
             return result;
         } else {
-            var texture = -1;
+            var texture = null;
             this.cache.addRessource(url, texture);
             return texture;
         }
@@ -191,6 +187,7 @@ WMS_Provider.prototype.getXbilTexture = function(tile, layer) {
 WMS_Provider.prototype.executeCommand = function(command) {
     var layer = command.paramsFunction.layer;
     var tile = command.requester;
+    var ancestor = command.paramsFunction.ancestor;
 
     var supportedFormats = {
         'image/png':           this.getColorTexture.bind(this),
@@ -201,7 +198,16 @@ WMS_Provider.prototype.executeCommand = function(command) {
 
     var func = supportedFormats[layer.format];
     if (func) {
-        return func(tile, layer, command.paramsFunction).then(function(result) {
+        var pitch = ancestor ?
+            this.projection.WMS_WGS84Parent(tile.bbox, ancestor.bbox) :
+            new THREE.Vector3(0, 0, 1);
+
+        var bbox = ancestor ?
+            ancestor.bbox :
+            tile.bbox;
+
+
+        return func(tile, layer, bbox, pitch).then(function(result) {
             return command.resolve(result);
         });
     } else {
--- a/src/Core/Commander/Providers/WMTS_Provider.js
+++ b/src/Core/Commander/Providers/WMTS_Provider.js
@@ -57,7 +57,6 @@ WMTS_Provider.prototype.customUrl = function(url, tilematrix, row, col) {
     urld = urld.replace('%COL', col.toString());
 
     return urld;
-
 };
 
 WMTS_Provider.prototype.removeLayer = function( /*idLayer*/ ) {
@@ -112,16 +111,42 @@ WMTS_Provider.prototype.url = function(coWMTS, layer) {
  * @param {type} coWMTS : coord WMTS
  * @returns {WMTS_Provider_L15.WMTS_Provider.prototype@pro;_IoDriver@call;read@call;then}
  */
-WMTS_Provider.prototype.getXbilTexture = function(tile, layer) {
-    var coWMTS = tile.tileCoord;
+WMTS_Provider.prototype.getXbilTexture = function(tile, layer, parameters) {
+    var cooWMTS = tile.matrixSet[layer.options.tileMatrixSet][0];
+    var pitch = new THREE.Vector3(0.0, 0.0, 1.0);
 
-    var url = this.url(coWMTS, layer);
+    if (parameters.ancestor) {
+        // account for possible level offset between coords and tile level
+        // (e.g for PM texture cooWMTS.level = tile.level + 1)
+        var levelOffset = cooWMTS.zoom - tile.level;
+
+        cooWMTS = this.projection.WMTS_WGS84Parent(
+            cooWMTS,
+            this.computeLevelToDownload(tile, parameters.ancestor, layer) + levelOffset,
+            pitch);
+    }
+
+    var url = this.url(cooWMTS, layer);
 
     // TODO: this is not optimal: if called again before the IoDriver resolves, it'll load the XBIL again
     var textureCache = this.cache.getRessource(url);
 
-    if (textureCache !== undefined)
-        return Promise.resolve(textureCache);
+    if (textureCache !== undefined) {
+        if (textureCache) {
+            let minmax = this._IoDriver.computeMinMaxElevation(
+                textureCache.floatArray,
+                256, 256,
+                pitch);
+            return Promise.resolve(
+                {
+                    pitch,
+                    texture: textureCache.texture,
+                    min: minmax.min,
+                    max: minmax.max
+                });
+        }
+        return Promise.resolve(null);
+    }
 
 
     // bug #74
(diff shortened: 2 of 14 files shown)


src/Scene/NodeProcess.js
```diff
@@ -276,47 +268,37 @@ function updateNodeElevation(quadtree, node, elevationLayers) {
     // Here we only make sure that the tile with level == levelElevation == 11 has its elevation texture.
     // Also see TileMesh.setTextureElevation
     var tileNotDownscaled = (node.level === node.levelElevation) ?
-        node :
+        null :
         node.getParentLevel(node.levelElevation);
```
gchoqueux: `node.getParentLevel(node.levelElevation)` it is not necessarily the ancestor not down scaled,
it may be less than `node.levelElevation`. If there's big zoom (level 2 to 20), it's not necessary to load all `levelElevation`.

gchoqueux:
``` javascript
function refinementCommandCancellationFn(cmd) {
     return cmd.requester.parent.childrenLoaded() &&
         cmd.requester.visible === false &&
         2 <= cmd.requester.level;
}
```

Constant 2 is linked with `tileMatrixSet`. It should be replaced by something like  `Quatree.minTileMatrixSet`.

src/Globe/TileMesh.js
```diff
@@ -29,9 +29,11 @@ import GlobeDepthMaterial from 'Renderer/GlobeDepthMaterial';
...
 import RendererConstant from 'Renderer/RendererConstant';
 
-var groupelevation = [14, 11, 7, 3];
-var l_ELEVATION = 0;
-var l_COLOR = 1;
+const groupelevation = [14, 11, 7, 3];
+const l_ELEVATION = 0;
+const l_COLOR = 1;
+
+const NO_ELEVATION = -2;
```
gchoqueux:
``` javascript
const NO_ELEVATION = -2;
```

Texture's Elevation has null, it's not sufficient to check if there's no elevation?

src/Globe/TileMesh.js
```diff
@@ -46,7 +48,9 @@ function TileMesh(params, builder, geometryCache) {
     this.geometry = defaultValue(geometryCache, new TileGeometry(params, builder));
     this.normal = params.center.clone().normalize();
 
-    this.distance = params.center.length();
+    var worldNormal = new THREE.Vector3(0, 0, 1).applyQuaternion(this.OBB().getWorldQuaternion());
```
gchoqueux, line 49: Why there's this change?

src/Globe/TileMesh.js
```diff
@@ -80,7 +84,7 @@ function TileMesh(params, builder, geometryCache) {
     }
 
     // Layer
-    this.currentElevation = -1;
+    this.currentElevation = NO_ELEVATION;
```
gchoqueux: > Texture's Elevation has null, it's not sufficient to check if there's no elevation?

Same comment
It's not possible to remove `this.currentElevation`?
or remove one state `-1` or `NO_ELEVATION`?
Maybe we need add level in elevation texture, like image texture.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -191,6 +187,7 @@ WMS_Provider.prototype.getXbilTexture = function(tile, layer) {
 WMS_Provider.prototype.executeCommand = function(command) {
     var layer = command.paramsFunction.layer;
     var tile = command.requester;
+    var ancestor = command.paramsFunction.ancestor;
 
     var supportedFormats = {
```
gchoqueux, line 192: `supportedFormats` it could be a private variable of the object

src/Renderer/LayeredMaterial.js
```diff
@@ -241,18 +241,21 @@ LayeredMaterial.prototype.setTexture = function(texture, layer, slot, pitScale)
...
 };
 
-LayeredMaterial.prototype.getIdLayer = function(slot) {
-    for (var l = 0; l < this.paramLayers.length; l++) {
-        if (slot === this.paramLayers[l].x)
-            return l;
+LayeredMaterial.prototype.slotToLayerIndex = function(slot) {
+    for (var l = 1; l < this.uniforms.nColorLayer.value; l++) {
+        if (slot < this.paramLayers[l].x) {
+            return l-1;
```
gchoqueux, line 262: Add comment : why yo substract one

src/Globe/TileMesh.js
peppsac, line 49: See commit message:

> Previous computation assumed we were using a globe.
> The new one produces the same result for a globe, but also
> work in other situations (planar mode for instance).

src/Renderer/LayeredMaterial.js
```diff
@@ -277,23 +280,20 @@ LayeredMaterial.prototype.getLayerTextureOffset = function(layer) {
...
+        }
     }
 };
 
-LayeredMaterial.prototype.getDelta = function() {
-    if (this.paramLayers[0])
-        // TODO: Fix Me, this function always used 1st layer
-        return 0; // this.paramLayers[0].y;
-    else
+LayeredMaterial.prototype.getDelta = function(layerIndex) {
```
gchoqueux: This function returns always 0?

src/Scene/Node.js
```diff
@@ -90,14 +90,17 @@ Node.prototype.update = function() {
  * @param {type} level
  * @returns {undefined}
  */
-Node.prototype.getParentLevel = function(level) {
+Node.prototype.getNodeAtLevel = function(level) {
```
gchoqueux, line 93: rename `getNodeAtLevel` in `getParentAtLevel`?

src/Renderer/LayeredMaterial.js
peppsac, line 262: Because `this.paramLayers[l].x` is an offset (layer N start at index this.paramLayers[N].x)
So if `slot < this.paramLayers[N].x` that means it belongs to the previous layer, hence the `-1`

src/Renderer/LayeredMaterial.js
peppsac: Yeah, because currently it only produces infinite requests.
Can you explain me what it's supposed to do?

src/Scene/Node.js
peppsac, line 93: I changed the name because I also changed the behavior.
If `this.level == level` then it returns `this` - so it's not always the parent

src/Scene/NodeProcess.js
```diff
@@ -215,23 +199,27 @@ function findAncestorWithValidTextureForLayer(node, layerId) {
...
 
-    for (var i = 0; i < colorLayers.length; i++) {
-        var layer = colorLayers[i];
+    const colorLayers = layersConfig.getColorLayers();
+    for (let i = 0; i < colorLayers.length; i++) {
+        let layer = colorLayers[i];
 
-        if (layer.tileInsideLimit(node, layer)) {
-            var args = {
+        // only request visible downscaled color layers
```
gchoqueux: see Freeze layer #165 
Freeze option layer allows not to load data and let visible layer

src/Globe/TileMesh.js
gchoqueux, line 49: Ok
It's possible to move compute normal in `TileGeometry` and implicitly in `BuilderEllipsoidTile` and `BuilderPlanarTile`. I don't understand why you need to project center to compute distance.

src/Scene/NodeProcess.js
peppsac: I think not loading a non-visible layer (regardless of its frozen status) makes sense.
So I can change the code to load only visible and non-frozen layer. Sounds good ?

src/Renderer/LayeredMaterial.js
gchoqueux: it is a bad thing
originally it's delta level between _**WGS84G layer**_ and _**PM layer**_. 
I had chosen : `PMLayer.level === tile.level + 1` 

For the same tile Pm layer is more accurate that WGS84G layer.

src/Scene/Node.js
gchoqueux, line 93: `getAncestorAtLevel` ?

src/Scene/NodeProcess.js
gchoqueux: yes, it's good

gchoqueux:
When you switch visibility's layer, you have to move the camera to reload layer.
should update the quadtree in change visibility?

src/Renderer/LayeredMaterial.js
peppsac: I don't think we need this at all.

peppsac:
Here's a diff to help visualize texture loading:

``` diff
    1 diff --git a/src/Scene/NodeProcess.js b/src/Scene/NodeProcess.js
    2 index e49db57..06ccfe1 100644
    3 --- a/src/Scene/NodeProcess.js
    4 +++ b/src/Scene/NodeProcess.js
    5 @@ -234,12 +234,15 @@ function updateNodeImagery(quadtree, node, layersConfig, force) {
    6              node.materials[RendererConstant.FINAL].getLevelLayerColor(1, slot) === -1) {
    7              ancestor = findAncestorWithValidTextureForLayer(node, layer.id);
    8              args.ancestor = ancestor;
    9 +            if (!node._f) node._f = {};
   10 +            node._f[layer.id] = [];
   11          }
   12  
   13          promises.push(quadtree.interCommand.request(args, node, refinementCommandCancellationFn).then(
   14              function(result) {
   15                  let level = (ancestor ? ancestor.level : node.level) +
   16                      (layer.options.levelOffset || 0);
   17 +                node._f[layer.id].push(level);
   18  
   19                  // Assign .level to texture
   20                  if (Array.isArray(result)) {
   21 @@ -275,6 +278,8 @@ function updateNodeElevation(quadtree, node, layersConfig, force) {
   22  
   23      var ancestor = null;
   24  
   25 +    if (!node._e) node._e = [];
   26 +
   27      if (currentElevation < 0) {
   28          // no texture: use elevation texture from 1st non-downscaled parent
   29          var n = node.getNodeAtLevel(node.levelElevation);
   30 @@ -312,6 +317,7 @@ function updateNodeElevation(quadtree, node, layersConfig, force) {
   31  
   32                  if (terrain && terrain.texture) {
   33                      terrain.texture.level = ancestor.level + (layer.options.levelOffset || 0);
   34 +                    node._e.push(terrain.texture.level);
   35                  }
   36  
   37                  node.setTextureElevation(terrain);
```

Then use _S_ + left click to inspect a tile.

gchoqueux:
Question : 
it's normal that all levels are loaded?

gchoqueux:
Same problems that #176 :  
- The tile refinement is too important : https://github.com/iTowns/itowns2/pull/179#issuecomment-238540887 
- Bad choice texture elevation : https://github.com/iTowns/itowns2/pull/179#issuecomment-238249931

gchoqueux:
> it's normal that all levels are loaded?

Seems to have been solved

peppsac:
PR updated.
I still can't reproduce the issues from this comment https://github.com/iTowns/itowns2/pull/176#issuecomment-239144001
Does it still happen for you @gchoqueux?

src/Renderer/LayeredMaterial.js
```diff
@@ -270,7 +273,9 @@ LayeredMaterial.prototype.setParam = function(param) {
     this.uniforms.nColorLayer.value = param.length;
     for (var l = 0; l < param.length; l++) {
         this.layerIdToIndex[param[l].idLayer] = l;
-        this.paramLayers[l] = new THREE.Vector4(param[l].layerTexturesOffset, param[l].tileMT === 'PM' ? 1 : 0, param[l].visible, param[l].opacity);
+        this.paramLayers[l] = new THREE.Vector4(
+            param[l].layerTexturesOffset,
+            param[l].tileMT === 'PM' ? 1 : 0, param[l].visible, param[l].opacity);
```
Jeremy-Gaillard, line 278: Could you remove the reference to 'PM' while you're at it? Seems way too specific to be in the material code.

According to the globe fragment shader, parameter.y indicates whether the layer is in wgs84 projection or not. I suggest replacing tileMT by a appropriately-named boolean (or adding a new attribute if you want to keep tileMT).

peppsac:
PR updated. All known issues are fixed.

src/Globe/TileMesh.js
```diff
@@ -195,37 +196,6 @@ TileMesh.prototype.setSelected = function(select) {
     this.materials[RendererConstant.FINAL].setSelected(select);
 };
 
-TileMesh.prototype.parseBufferElevation = function(image, minMax, pitScale) {
```
gchoqueux, line 198: This function computes the bounding box's tile.
This bbox is used to compute bounding sphere.
This bounding sphere is used to compute the screen space error. 
If this function is removed the screen space error is too big, and tiles are subdivided or displayed unnecessarily.
we encountered performance problems without this optimization. 
Furthermore the bouding spheres will be used for collision.

src/Renderer/LayeredMaterial.js
peppsac: I plan to work on this topic soon. I'll take care of the PM/WGS84 then.

peppsac:
PR rebased and updated. Ready for review.

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -112,16 +111,42 @@ WMTS_Provider.prototype.url = function(coWMTS, layer) {
  * @param {type} coWMTS : coord WMTS
  * @returns {WMTS_Provider_L15.WMTS_Provider.prototype@pro;_IoDriver@call;read@call;then}
  */
-WMTS_Provider.prototype.getXbilTexture = function(tile, layer) {
-    var coWMTS = tile.tileCoord;
+WMTS_Provider.prototype.getXbilTexture = function(tile, layer, parameters) {
+    var cooWMTS = tile.matrixSet[layer.options.tileMatrixSet][0];
```
gchoqueux, line 115: When I delete a `addImageryLayer` (with `name: 'ORTHOIMAGERY.ORTHOPHOTOS.CRS84'` by example) in html file [(here)](https://github.com/iTowns/itowns2/blob/e615c05969b802ed88949d7cd1e93bf0b286b905/index.html#L174).
There's a error `tile.matrixSet[layer.options.tileMatrixSet]` is `undefined`.
There's not this error in master 94a98e455813fc36a878852d887067369782bb7c

gchoqueux:
No updating/reloading when a layer going from invisible to visible (Necessary change in PR #179)

src/Core/Commander/Providers/WMTS_Provider.js
peppsac: This bug is fixed.

PR #178, opened 2016-08-05
Fix error longitude on globe

gchoqueux:
This PR solves the problem : the 'zero meridian' doesn't cross Greenwich.(see PR #19)

This PR recasts BoundingBox and change BBox's Globe -> longitude(-180,180),latitude(-90,90)


src/Scene/BoundingBox.js
```diff
@@ -5,32 +5,37 @@
...
 /**
  *
- * @param {type} minLongitude : longitude minimum
- * @param {type} maxLongitude : longitude maximum
- * @param {type} minLatitude  : latitude minimum
- * @param {type} maxLatitude  : latitude maximum
+ * @param {type} west : longitude minimum
+ * @param {type} east : longitude maximum
+ * @param {type} south  : latitude minimum
+ * @param {type} north  : latitude maximum
```
peppsac, line 18: What about storing 4 floats in a array, and add an additionnal parameter (`unit`?).
This way we could handle _degrees_ and _metres_ more easily in the codebase.

In the same way, `south` would be a function, returning either `coords[0]` (if `unit` is degrees) or convert `coords[0]` from `unit` to degrees.
Note that for a first step we should **not** support conversion and simply throw an error.

src/Core/Commander/Providers/BuildingBox_Provider.js
```diff
@@ -83,7 +83,7 @@ BuildingBox_Provider.prototype.generateMesh = function(elements, bbox, altitude)
     var ellipsoid = new Ellipsoid(new THREE.Vector3(6378137, 6356752.3142451793, 6378137));
     var features = elements.features;
     var altitude_ground = altitude - 1.5; //35;  // truck height
-    //var cPano = new CoordCarto().setFromDegreeGeo(p1.x  ,p1.z, z_min );
+    //var cPano = new GeoCoordinate().setFromDegree(p1.x  ,p1.z, z_min );
```
peppsac: Remove dead code

src/Core/Geographic/CoordStars.js
```diff
@@ -97,9 +97,9 @@ var CoordStars = {
         var dayMilliSec = 24 * 3600000;
 
         var longitude = sun.ascension + ((date % dayMilliSec) / dayMilliSec) * -360 + 180; // cause midday
-        var coSunCarto = ellipsoid.cartographicToCartesian(new CoordCarto().setFromDegreeGeo(longitude, lat, 50000000));
-        //  console.log(ellipsoid.cartographicToCartesian(new CoordCarto().setFromDegreeGeo(48.85, 2.35, 0)));
-        //  console.log(ellipsoid.geodeticSurfaceNormalCartographic(new CoordCarto().setFromDegreeGeo(48.85, 2.35, 0)));
+        var coSunCarto = ellipsoid.cartographicToCartesian(new GeoCoordinate().setFromDegree(longitude, lat, 50000000));
+        //  console.log(ellipsoid.cartographicToCartesian(new GeoCoordinate().setFromDegree(48.85, 2.35, 0)));
+        //  console.log(ellipsoid.geodeticSurfaceNormalCartographic(new GeoCoordinate().setFromDegree(48.85, 2.35, 0)));
```
peppsac: Dead code

src/Core/Geographic/GeoCoordinate.js
```diff
@@ -0,0 +1,204 @@
...
+/**
+ *
+ * @param {type} defaultValue
+ * @returns {CoordCarto_L9.GeoCoordinate}
+ */
+import defaultValue from 'Core/defaultValue';
+import mE from 'Core/Math/MathExtented';
+
+/* eslint-disable */
+var UNIT_RADIAN = 0;
```
peppsac: Use `const` instead of `var`?
Also I don't see the point of disabling the linter. Remove `COORD_ALT` instead if it's not used.

src/Core/Geographic/GeoCoordinate.js
```diff
@@ -0,0 +1,204 @@
...
+var UNIT_METER = 2;
+
+var COORD_LONG = 0;
+var COORD_LAT = 1;
+var COORD_ALT = 2;
+/* eslint-enable */
+
+function GeoCoordinate(longitude, latitude, altitude,unit) {
+
+
```
peppsac: Unneeded blank lines

src/Core/Geographic/GeoCoordinate.js
```diff
@@ -0,0 +1,204 @@
...
+GeoCoordinate.UNIT_DEGREE = UNIT_DEGREE;
+
+GeoCoordinate.prototype.constructor = GeoCoordinate;
+
+GeoCoordinate.prototype.longitude = function() {
+
+	return this.coordinate[this.angleUnit][COORD_LONG];
+};
+
+var _setCoordinateValue = function(angleUnit,coordinate,id,value)
```
peppsac: Any advantages over the `function _setCoordinateValue(angleUnit,coordinate,id,value)` syntax?

peppsac:
I don't really like the design of the `GeoCoordinate`:
- I don't like having a `this.angleUnit` state + `longitude()/latitude()` methods that depends on it
- `this.coordinate[][]` is a complex construction. Why not simply have a 3-coords array?
- I believe that having const methods (as in C++ semantics) would be clearer than having methods (like`setUnit` or `toDegree`) modifying this
- `new GeoCoordinate().setFromDegree(x, y, z)` is a heavy syntax and could be simplified 

Here's what I had in mind:

``` js
var c = new GeoCoordinate(UNIT_DEGREES, long, lat, alt);
var longInDegrees = c.longitude(UNIT_DEGREES); // doesn't modify c
var latInRadians = c.latitude(UNIT_RADIANS); // doesn't modify c
var x = c.x(UNIT_METERS); // x and longitude methods are aliases.
var d = c.convert(UNIT_RADIANS); // doesn't modify c
```

For a while, we would assume that WGS84 is always used (when using angles) or L93 (when using meters).
Later, all these methods could take an additionnal argument: srs (or crs):

``` js
var c = new GeoCoordinate(WGS84, UNIT_DEGREES, long, lat, alt);
var longInDegrees = c.longitude(WGS84, UNIT_DEGREES);
var x = c.x(L93, UNIT_METERS);
var d = c.convert(EPSG3960);
```

What do you think?

gchoqueux:
> I don't like having a `this.angleUnit` state + `longitude()/latitude()` methods that depends on it

 `this.angleUnit` is removed

> `this.coordinate[][]` is a complex construction. Why not simply have a 3-coords array?

 `this.coordinate[][]` is replaced by typed array

> I believe that having const methods (as in C++ semantics) would be clearer than having methods (like`setUnit` or `toDegree`) modifying this

`setUnit` and `toDegree` are removed. 

`GeoCoordinate.longitude(unit)` return longitude in the specified unit, default is in radian
`GeoCoordinate.latitude(unit)` return latitude in the specified unit, default is in radian

``` js
var coordinate = new GeoCoordinate(longitude, latitude, altitude);
var longitudeInDegree = coordinate.longitude(UNIT_DEGREE); 
var latitudeInRadian = coordinate.latitude(UNIT_RADIAN);
latitudeInRadian = coordinate.latitude(); // return is in radian
```

> `new GeoCoordinate().setFromDegree(x, y, z)` is a heavy syntax and could be simplified 

``` js
var coordinate_A = new GeoCoordinate(longitude, latitude, altitude,RADIAN);
var coordinate_B = new GeoCoordinate(longitude, latitude, altitude); // radian is default unit
var coordinate_C = new GeoCoordinate(longitude, latitude, altitude,DEGREE);
```

nosy-b:
Ok I'm looking through the updated code. Seems to work without bugs
A few remarks:
- It can be a bit frustrating to loose the coordinate when you create them in degree. 
  For example you create the geo coord for paris : `var posWGS84 = new GeoCoordinate(2.34, 48.86, 0, 1);`
  Then you will automatically convert it in Rad loosing the initial easy to read degree coord. (Specifying the 4rth parameter by a number for unit type is not intuitive either. )
- Paris returned absolute coordinate is: {x: 20626143.989646945, y: 23615140.058829017, z: -845198.8006852799} but as mentionned in #19 Z should be positive.

gchoqueux:
> loosing the initial easy to read degree coord

There's not loosing if you use `posWGS84.longitude(UNIT.DEGREE);` , it return in degree.

> Specifying the 4rth parameter by a number for unit type is not intuitive either.

you need to import `UNIT from 'Core/Geographic/GeoCoordinate'` to have intuitive parameter

``` js
import GeoCoordinate,{UNIT} from 'Core/Geographic/GeoCoordinate';
var posWGS84 = new GeoCoordinate(2.34, 48.86, 0,UNIT.DEGREE);
```

> Paris returned absolute coordinate is: {x: 20626143.989646945, y: 23615140.058829017, z: -845198.8006852799} but as mentionned in #19 Z should be positive.

It's normal for the moment, Y is up of the system.
Next line Example to have layer in Z up
https://github.com/iTowns/itowns2/blob/master/src/Globe/Globe.js#L81

This feature needs a new PR.

PR #180, opened 2016-08-30
Fix horizon occlusion bug

peppsac:
In horizonCulling `points[i]` was being modified. As a result occlusion
result was broken.

This commit:
- fix the issue (33 culled tiles instead of 2 at camera startup position)
- replace custom function by Three.js implementation
- reduce the number of temporary THREE.Vector3

### DIFF
--- a/src/Scene/NodeProcess.js
+++ b/src/Scene/NodeProcess.js
@@ -394,32 +394,25 @@ NodeProcess.prototype.prepare = function(camera) {
  * @returns {undefined}
  */
 NodeProcess.prototype.preHorizonCulling = function(camera) {
-
-    this.cV = MathExt.divideVectors(camera.position(), this.r);
-
-    this.vhMagnitudeSquared = MathExt.lenghtSquared(this.cV) - 1.0;
-
+    this.cV.copy(camera.position()).divide(this.r);
+    this.vhMagnitudeSquared = this.cV.lengthSq() - 1.0;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


Jeremy-Gaillard:
Seems OK.

You should remove the THREE dependency from MathExtended, it causes an error when building.

peppsac:
@Jeremy-Gaillard thanks. Fixed the build error.

src/Scene/NodeProcess.js
```diff
@@ -429,43 +422,23 @@ NodeProcess.prototype.pointHorizonCulling = function(point) {
...
 
 NodeProcess.prototype.horizonCulling = function(node) {
 
     // horizonCulling Oriented bounding box
     var points = node.OBB().pointsWorld;
-    center.setFromMatrixPosition(node.matrixWorld);
     var isVisible = false;
     for (var i = 0, max = points.length; i < max; i++) {
-        var point = points[i].add(center);
+        point.setFromMatrixPosition(node.matrixWorld).add(points[i]);
```
gchoqueux: May use `addVectors(v1,v2)` 

``` javascript
nodePosition.setFromMatrixPosition(node.matrixWorld);
for (var i = 0, max = points.length; i < max; i++) {
        point.addVectors(nodePosition,points[i]);

        if (!this.pointHorizonCulling(point)) {
            isVisible = true;
            break;
        }
    }
```

gchoqueux:
r+

PR #187, opened 2016-09-14
Change Three v79 to v80

gchoqueux:
the Three library now uses ES6 Modules

### DIFF
--- a/src/Renderer/ThreeExtented/GlobeControls.js
+++ b/src/Renderer/ThreeExtented/GlobeControls.js
@@ -6,9 +6,7 @@
 //    Zoom - middle mouse, or mousewheel / touch: two finger spread or squish
 //    Pan - right mouse, or arrow keys / touch: three finter swipe
 
-/* global document,window*/
-
-import THREE from 'three';
+import * as THREE from 'THREE';
 import Sphere from 'Core/Math/Sphere';
 import CustomEvent from 'custom-event';
 
@@ -177,10 +175,11 @@ var snapShotCamera;
 
 /////////////////////////
 
+/* globals document,window */
+
 function GlobeControls(camera, domElement, engine) {
 
     var scene = engine.scene;
-
     this.camera = camera;
     snapShotCamera =  new SnapCamera(camera);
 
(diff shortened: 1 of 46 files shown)


src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -225,8 +224,9 @@ function GlobeControls(camera, domElement, engine) {
 
 
     if(enableTargetHelper)
-
+    {
```
peppsac, line 232:
``` js
if (enableTargetHelper) {
```

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -445,7 +445,9 @@ function GlobeControls(camera, domElement, engine) {
             var delta = coord.altitude()  - (bbox.top() + radiusCollision);
 
             if(delta<0)
+            {
```
peppsac:
``` js
if (delta < 0) {
```

peppsac (commented):
I'm not sure that using `import * as THREE from 'THREE';` is a good idea.
I'd prefer to explicitely list imported elements ; what do you think?

gchoqueux:
> I'm not sure that using import \* as THREE from 'THREE'; is a good idea.

Yes, I think too `import * as THREE from 'THREE';` is not a best idea.

For example in https://github.com/iTowns/itowns2/blob/ThreeV80/src/Renderer/BasicMaterial.js#L8

``` js
import * as THREE from 'THREE';
```

becomes

``` js
import {RawShaderMaterial,Color,Matrix4} from 'THREE';
```

But we lose name's module

``` js
```

becomes

``` js
RawShaderMaterial.call(this);
```

I didn't find import method to load specific member and keeping name's module :(
I think, it's better to keep name's module `THREE`

tbroyer:
What's wrong with `import * as THREE from 'three'`? That's exactly what the [pull request making the change to Three.js](https://github.com/mrdoob/three.js/pull/9310) suggests.

peppsac:
In most language wildcard imports are considered a bad practice, hence my question.
In Javascript these imports are scoped thanks to the `as XXX` syntax so I suppose that's not an issue.

So it looks good but I need to test before giving my approval :-)

peppsac:
LGTM

PR #190, opened 2016-09-28
Geoid

mbredif:
This PR regroups all the modifications that where necessary to build the [geoid demo](http://www.itowns-project.org/demo-geoid/). 
# Extended api
- ApiGlobe.createSceneGlobe(coordCarto, viewerDiv, params) : added a params dictionnary.
  - params.elevationZooms (Int Array : [14,11,7,3]) : are the subset of elevation zooms to be used (formerly hardcoded groupelevation array is the default), implemented [here](http://github.com/iTowns/itowns2/blob/geoid/src/Core/Commander/Providers/TileProvider.js#L77-L89)
  - params.atmosphere (boolean : true) : build a globe with/without an atmosphere
  - params.zFactor (Numeric : 1) : initial zFactor
- new addImageryLayer param options:
  - "zoom : { min:1, max:4 }", used in [SSE computation](http://github.com/iTowns/itowns2/blob/geoid/src/Scene/NodeProcess.js#L81). Previously a minzoom=2 was hardcoded and maxzoom was not checked.
  - "opacity" (Numeric : 1)
  - "visible" (boolean : true)
- new api : ApiGlobe.getLayerOpacity(layerid)
- new api : ApiGlobe.setZFactor(zFactor) and  ApiGlobe.getZFactor() to manipulate the elevation exageration
- a [screenshot](http://github.com/iTowns/itowns2/blob/geoid/index.html#L106-L168) function that is able to create screenshots larger than 4096*4096 using tiled rendering, (now in the index.html, to be integrated in the js code)
# ZFactor implementation

correct culling of exagerated elevation required the following modifications:
- [fix bug with tiles where elevation<-10](http://github.com/iTowns/itowns2/blob/geoid/src/Core/Commander/Providers/IoDriver_XBIL.js#L38).
- [clamping removed due to tiles with negative elevation. Better nodata handling is required](http://github.com/iTowns/itowns2/blob/geoid/src/Renderer/Shader/GlobeVS.glsl#L66)
- [increase skirt height to prevent cracks](http://github.com/iTowns/itowns2/blob/geoid/src/Globe/TileGeometry.js#L216)
- OBB is [now cached](http://github.com/iTowns/itowns2/blob/geoid/src/Scene/NodeProcess.js#L362) in NodeProcess.frustumCullingOBB
- OBB now takes into account [on the fly](http://github.com/iTowns/itowns2/blob/geoid/src/Globe/TileMesh.js#L341-L345) the ZFactor to ensure correct Frustum and Horizon Culling.


tbroyer (commented):
Do we really want to replace the index.html with the geoid demo?

vpicavet:
demos should preferably go a specific repository, we should only keep one simple index.html in the main repo

mbredif:
Ok, I reverted the index.html to a regular globe.
I also added support for noDataValue and created elevation layer states similar to color layer states (zFactor and noDataValue).
- Not sure about performance
- The elevation layer values used in the demo has other bogus values than -99999. Are they specified?
- I get CORS errors when working on npm dev server : 
  `Image from origin 'http://wxs.ign.fr' has been blocked from loading by Cross-Origin Resource Sharing policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8080' is therefore not allowed access. The response had HTTP status code 403.`

vpicavet:
As for the CORS errors, you should first make sure that CORS is allowed on wxs.ign.fr web services.

gcebelieu:
> As for the CORS errors, you should first make sure that CORS is allowed on wxs.ign.fr web services

normaly it should, but it seems that it is not allways the case... we're investigating

gcebelieu:
Finaly the CORS problem wasn't one : it was a consequence of geoportal key rights problem. When wxs.ign.fr answers in 403 there is no 'Access-Control-Allow-Origin' header set (useless) ; when it answers in 200 (data is returned) it is correctly set.

The key rights problem has been solved. You should'nt have CORS messages any more.

mbredif:
Thanks for the CORS error : it is fixed now.
What about the nodata values? -99999 seems to be the no-data value almost everywhere but not quite. What are the other codes ?

gchoqueux:
Warning, @qdnguyen , this branch needs to rebase on master and a cleaning on history (squash commit) !!

qdnguyen:
yes, i will do it ASAP when i finish with IRTF data.

peppsac:
@mbredif, @qdnguyen: this PR has some useful stuff but it should be splitted in smaller ones to ease the review/merge process.

gchoqueux:
moved in #511

PR #191, opened 2016-09-30
Layer update strategies

peppsac:
Implement different (and configurable) update strategies for layer.

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,321 +34,319 @@ and open the template in the editor.
         <script src="dist/itowns2.js"> </script>
         <script type="text/javascript">
 
-               var initCenter = { longitude:2.3465, latitude: 48.88, altitude: 25000000};
+            var initCenter = { longitude:2.3465, latitude: 48.88, altitude: 25000000};
 
-               // iTowns namespace defined here
+            // iTowns namespace defined here
+            var viewerDiv = document.getElementById("viewerDiv");
 
-               var viewerDiv = document.getElementById("viewerDiv");
+            var scene = itowns.viewer.createSceneGlobe(initCenter, viewerDiv) ;
 
-               itowns.viewer.createSceneGlobe(initCenter, viewerDiv) ;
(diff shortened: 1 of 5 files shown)


index.html
```diff
@@ -34,321 +34,331 @@
...
+                }
+            });
 
-                itowns.viewer.addImageryLayer({
+            itowns.viewer.addImageryLayer({
                 protocol:   "wmts",
                 id:         "ScanEX",
                 url:        "http://wxs.ign.fr/va5orxd0pgzvq3jxutqfuy0b/geoportail/wmts",
                 fx :        2.5,
+                updateStrategy: {
```
gchoqueux, line 272: should be added default value when there's no updateStrategy

gchoqueux (changes requested):

gchoqueux:
No updating/reloading when a layer going from invisible to visible

nosy-b:
Same bug as the master:
When going **outside of France** even with only 'Ortho' imagery layer activated it tries to download other layers data deactivated(ex: SCAN-EXPRESS.STANDARD)

peppsac:
> Same bug as the master:
> When going outside of France even with only 'Ortho' imagery layer activated it tries to download other layers data deactivated(ex: SCAN-EXPRESS.STANDARD)

It's expected.
To be `loaded` a tile needs to have all of its textures, regardless of the status of the layer.
So while I agree with you that this is _wrong_, a proper fix would be to rework the `child.texturesNeeded` mechanism.

Note : a tile expects to have (at least) 1 texture per layer where `tileInsideLimit` returns true.

index.html
peppsac: Fixed.

peppsac:
> No updating/reloading when a layer going from invisible to visible

Fixed as well (calling `update()` instead of `scene.renderScene3D()` in `ApiGlobe.prototype.setLayerVisibility`)

gchoqueux:
r+

PR #197, opened 2016-10-21
Rework on index.html, addImageryLayerFromJSON, addElevationLayerFromJSON, GUIGlobe

vcoindet:
**In ApiGlogbe.js :** 
- addElevationLayerFromJSON()
  Allows to load a single Imagery Layer from a JSON file
- addImageryLayersFromJSONArray()
  Allows to load Imagery Layers from an array composed of JSON files
- addImageryLayerFromJSON()
  Allows to load a single Imagery Layer from a JSON file
- addElevationLayersFromJSONArray() 
  Allows to load Elevation Layers from an array composed of JSON files

**GuiTools** 
In examples/GUI/
New class to create a new menu using dat.gui

**JS and JSON layers**
In examples/layers

**In index.html**
Rework on the all file using the previous functions. This makes the index.html file more readable...


src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -100,14 +101,48 @@ function preprocessLayer(layer, provider) {
...
 };
 
+/**
+ * This function adds an imagery layer to the scene using a JSON file. The layer id must be unique. The protocol rules wich parameters are then needed for the function.
+ * @constructor
+ * @param {Layer} layer.
+ */
+
+ApiGlobe.prototype.addImageryLayerFromJSON = function(url) {
+     return this.readJSON(url).then(function(result){
```
peppsac: Why not use  `JSONDriver.read(url)` directly instead of adding a one-liner function?

src/Core/Commander/Providers/JSON_Provider.js
```diff
@@ -0,0 +1,13 @@
...
+ * Description: Parseur de JSON
+ */
+
+import Provider from 'Core/Commander/Providers/Provider';
+import IoDriver_JSON from 'Core/Commander/Providers/IoDriver_JSON';
+
+function JSON_Provider(ellipsoid) {
+    //Constructor
+
+}
```
peppsac: This file is not needed

examples/GUI/GUIGlobe.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * Generated On: 2015-10-5
+ * Class: GUIGlobe
+ * Description: Classe pour créer un menu.
+ */
+
+
+function GUIGlobe() {
```
peppsac: Why _Globe_ suffix? This is a helper class to build a GUI to display/manipulate layers, so what do you think of `LayerGUI` ?

examples/GUI/GUIGlobe.js
```diff
@@ -0,0 +1,69 @@
...
+    //Constructor
+
+    this.gui = new dat.GUI( { autoPlace : false});
+    this.gui.domElement.id = 'menuDiv';
```
peppsac: Don't hardcode the element id. Add it as a parameter to the constructor instead.

examples/GUI/GUIGlobe.js
```diff
@@ -0,0 +1,69 @@
...
+ * Generated On: 2015-10-5
+ * Class: GUIGlobe
+ * Description: Classe pour créer un menu.
+ */
+
+
+function GUIGlobe() {
+    //Constructor
+
+    this.gui = new dat.GUI( { autoPlace : false});
```
peppsac: `dat` will be undefined.
Maybe it's time to try to integrate it as a npm dev-dependency for itowns?

peppsac:
Instead of moving layers declaration content to separate files to cleanup index.html, supporting standard `GetCapabilities` method for WMTS layers would be a huge benefit.

We could then use:

``` js
            itowns.viewer.addImageryLayer({
                protocol:   "wmts",
                id:         "Ortho",
                url:        "http://wxs.ign.fr/va5orxd0pgzvq3jxutqfuy0b/geoportail/wmts",
                name: "ORTHOIMAGERY.ORTHOPHOTOS",
                updateStrategy: {
                    type: 0, /* see LayerUpdateStrategy.js */
                    options: {}
                }
            });
```

Instead of having to copy-paste 10s of config lines.

examples/GUI/GUIGlobe.js
vcoindet: It's not only to display layers, it's also used to create section in the menu. See the "addGUI" function which is used to create the 'RealisticLighting' section.
This class might be used to do other things in the future, so I'm not sure that "layerGUI" is a good name...

vcoindet:
I totally agree, supporting standard GetCapabilities method for WMTS layers would be a huge benefit.
But I think we should do both ways (GetCapabilities and the method proposed in this PR). I think it could be usefull for users

examples/GUI/GUIGlobe.js
peppsac: I don't really see the point of `GUIGlobe`.
What are the benefits of introduce this class over using `dat.gui` directly?
From my point of view, there's no need to introduce a new class if you just a new a few simple helpers methods in a single place...

peppsac:
Sure. But it's a good opportunity to implement `GetCapabilities` :)

vcoindet:
Yes I totally agree ;)

examples/GUI/GUIGlobe.js
vcoindet: I will probably need some help on that point... I guess there is some work to do on webpack.config et package.json but I'm not very comfortable with the package.json...

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: ok

src/Core/Commander/Providers/JSON_Provider.js
gchoqueux: removed

examples/GUI/GUIGlobe.js
gchoqueux: @tbroyer Need help to use dat.gui npm dependency

gchoqueux:
branch updated and rebased

autodeploy.sh
```diff
@@ -34,16 +34,22 @@ git clone $REPO --single-branch --branch $TARGET_BRANCH out
...
+git rm -rf dist index.html API_Doc examples/layers examples/GUI
 popd
 # Copy build results
 cp -R dist out/
 # generate the API documentation
 npm run doc -- -d out/API_Doc
 # Copy demo
 cp index.html out/
+# Copy layers
+cp -R examples/layers out/
```
tbroyer: Won't that create an `out/layers/` rather than `out/examples/layers/`? (as expected by the `git rm` at lines 37 and 52)

gchoqueux: updated
question @tbroyer : how use dat.gui and npm?

tbroyer:
> question @tbroyer : how use dat.gui and npm?

There's `dat-gui` for that 3 year-old (sic!) v0.5, or `dat.gui` for the more recent 0.6.0 and 0.6.1.
Given that it's for examples only, it'd have to be a `dev` dependency, not a standard one (the iTowns library itself doesn't depend on it); but it'd be useless to depend on it from NPM unless we also process examples with Webpack (using multiple `entry` entries), which may not be worth it.

vcoindet:
branch updated and rebased

examples/GUI/GuiTools.js
```diff
@@ -0,0 +1,71 @@
...
+		folder.add( { frozen: false }, 'frozen').onChange(function(value) {
+			this.api.setLayerFreeze(layer.id, value);
+		}.bind(this));
+};
+
+GuiTools.prototype.addElevationLayerGUI = function (layer) {
+
+		var folder = this.elevationGui.addFolder(layer.id);
+		folder.add( { frozen: false }, 'frozen').onChange(function(value) {
+			this.api.setLayerFreeze(this, value);
```
peppsac:
```
GuiTools.js:38 Uncaught TypeError: Cannot read property 'setLayerFreeze' of undefined
```

examples/layers/JSLayers/IGN_MNT.js
```diff
@@ -0,0 +1,73 @@
+itowns.viewer.addElevationLayer({
```
peppsac: AFAIK `JS/Layers/*.js` files aren't used. What's the point of adding them then?

examples/GUI/GuiTools.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Generated On: 2015-10-5
+ * Class: GuiTools
+ * Description: Classe pour créer un menu.
+ */
+
+/* global viewerDiv*/
+
+function GuiTools(api,domId) {
+    //Constructor
```
peppsac: Avoid useless comments. Same comment applies to the comments at the top of the file

examples/GUI/GuiTools.js
vcoindet: Sorry, I forgot to correct this one... Will do it

peppsac:
@vcoindet  didn't answer my question : 

> I don't really see the point of GUIGlobe.
> What are the benefits of introduce this class over using dat.gui directly?
> From my point of view, there's no need to introduce a new class if you just a new a few simple helpers methods in a single place...

vcoindet:
Well, the purpose here was to make a more clean index.html. I think the way we organise our menu/GUI should be just an exemple for users, I don't think we should impose our own menu/GUI.

examples/layers/JSLayers/IGN_MNT.js
vcoindet: that's right, we should probably remove all the js layer files as we don't use it. I will do it !

examples/GUI/GuiTools.js
vcoindet: done

examples/layers/JSLayers/IGN_MNT.js
vcoindet: JS files removed. done

examples/GUI/GuiTools.js
vcoindet: done

peppsac:
LGTM.

PR #199, opened 2016-11-02
Cleanup: remove webpack module aliases

tbroyer:
Those aliases were legacy from the RequireJS era to ease the
migration to webpack.


peppsac:
Thanks, merging.

Issue #192, opened 2016-10-04
Bad handling of wmts layer boundaries when mixing them. (Test with world DTM)


nosy-b:
I added in the index an elevationLayer for the whole world and an elevationLayer for high precision DTM just for France.
- The order of declaration of the elevation layers is important. (But we wish to have an automatic use of the best precision layer available at any position)
- The update strategy group has to be the same for both even though one has less levels than the other.
- Test on tileMatrixSetLimits is not done (minTileRow etc). And even if it was, no country is squared. On the boundary we usually meet Tiles with NoDataValue and no tiles available.
- Same problem as already mentioned(#191). If deactivated layers (imagery) it can still try in some case to request them and end up to too many bad requests.

Two big use cases: 
- Evolving around the Alpes. It works fine with no break now, but when zooming more out of France it tries to download HighRes French DTM.
- Deactivating all imagery Layers but Ortho, then go to Everest Mountain, it dies on trying to download many tiles of deactivated imagery layers (not available out of France)

~The Branch for those tests is [https://github.com/iTowns/itowns2/tree/worldDTM](https://github.com/iTowns/itowns2/tree/worldDTM)~

I 'm going to work on this (starting by handling deactivated imagery layer as discussed in #191) but as you did some work on it recently @peppsac , @gchoqueux , don't hesitate to look too.

peppsac:
My 2 cents:

> The order of declaration of the elevation layers is important. (But we wish to have an automatic use of the best precision layer available at any position)

Yes because itowns can only use 1 texture for elevation, so higher priority layer should be listed 1st.
What do you have in mind to fix this?

> (not available out of France)

It shouldn't try to download them if the layer had a proper bounding box (instead of simply relying on the tile.level).
Also note that I have a set of commits to handle error properly - with the default strategy being: 3 retries then stop trying to download a texture.
See https://github.com/iTowns/itowns2/pull/141/commits/4469b742bfee58bf1bd07fdc005817a5921efaa3 and https://github.com/iTowns/itowns2/pull/141/commits/2cb7c59db2547354b3aa0b1e99eacc1a8dca870a

PR #200, opened 2016-11-09
Rework LayeredMaterial

gchoqueux:
- **Add `Style` object to WFS**
- **Rework LayeredMaterial**
  - All textures and parameters are ordered by display order.
  - `this.Layers` contains all layers by display order
  - The slots's texture management is done internally, now, to set texture uses `layerId`
  - Fix bug : Remove color layer doesn't dispose textures colors
- **Uniform changes** (`LayeredMaterial` and `BasicMaterial`)
  - use`THREE.Uniform` in material
  - `int RTC`          becomes `bool useRTC`
  - `int lightingOn` becomes `bool lightingOn`
  - `int debug`       becomes `bool debug`
  - `int lightingOn` becomes `bool lightingOn`
  - `int selected`    becomes `bool selected`
- **Remove slot texture** in `NodeProcess` is replaced by `layerId`
- **Change Skirt size** influence performance rendering, minimize pixel shader pass
- **Fix typos and add Binary Tools**


src/Globe/Globe.js
```diff
@@ -208,6 +208,17 @@ Globe.prototype.updateLayersOrdering = function() {
     this.tiles.children[0].traverse(cO);
 };
 
+
+Globe.prototype.removeColorLayer = function(layer) {
+
+    var cO = function(object) {
+        if (object.changeSequenceLayers)
```
peppsac: Add missing `{  }`

peppsac: Didn't you intend to write `if (object.removeColorLayer)` ?

src/Globe/TileMesh.js
```diff
@@ -320,42 +281,25 @@ TileMesh.prototype.loadingCheck = function() {
...
 };
 
-TileMesh.prototype.removeLayerColor = function(idLayer) {
-    var nbTextures = this.materials[RendererConstant.FINAL].nbLoadedTextures();
-    this.materials[RendererConstant.FINAL].removeLayerColor(idLayer);
-    this.texturesNeeded -= nbTextures - this.materials[RendererConstant.FINAL].nbLoadedTextures();
+TileMesh.prototype.removeColorLayer = function(idLayer) {
+    var texturesCount = this.materials[RendererConstant.FINAL].loadedTexturesCount();
+    this.materials[RendererConstant.FINAL].removeColorLayer(idLayer);
+    this.texturesNeeded -= texturesCount - this.materials[RendererConstant.FINAL].loadedTexturesCount();
```
peppsac: This is wrong.
If the textures for this layer aren't yet loaded  this tile is going to be stuck in the `!loaded` state because `this.texturesNeeded` will be incorrect.

`texturesNeeded` should be updated the same way it's initialized in `NodeProcess`.
A simpler alternative would be to disallow layer removal for `!loaded` tiles.

src/Renderer/LayeredMaterial.js
```diff
@@ -177,152 +187,248 @@ LayeredMaterial.prototype.dispose = function() {
...
+    this.layerTexturesCount[index] = count;
+};
+
+LayeredMaterial.prototype.loadedTexturesCount = function() {
+    return this.texturesCount[l_ELEVATION] + this.texturesCount[l_COLOR];
 };
 
-LayeredMaterial.prototype.setNbLayersColor = function(n) {
-    this.uniforms.nColorLayer.value = n;
+LayeredMaterial.prototype.downScaledColorLayer = function(layer,level) {
```
peppsac: Maybe rename `isColorLayerDownscaled`?

src/Renderer/LayeredMaterial.js
```diff
@@ -177,152 +187,248 @@ LayeredMaterial.prototype.dispose = function() {
...
+    return this.Textures[l_COLOR][this.getLayerTextureOffset(layer)].level < level;
+};
+
+LayeredMaterial.prototype.downScaledLayer = function(layerType,level) {
```
peppsac: Maybe rename `isLayerTypeDownscaled`?

src/Renderer/LayeredMaterial.js
```diff
@@ -177,152 +187,248 @@ LayeredMaterial.prototype.dispose = function() {
...
+
+    return false;
 };
 
+LayeredMaterial.prototype.getColorLayerLevel = function(colorLayerId) {
 
-LayeredMaterial.prototype.getLayerLevel = function(id, slot) {
-    var level = this.Textures[id][slot || 0].level;
+    let index = this.indexOfColorLayer(colorLayerId);
+    if(index ===-1) {
```
peppsac, line 417: I'm not sure that hiding the "colorLayerId is invalid" error is a good idea.
Either throw an error, or return an invalid level value (`-1` ?)

src/Scene/NodeProcess.js
```diff
@@ -95,36 +95,43 @@ NodeProcess.prototype.subdivideNode = function(node, camera, params) {
                 var j;
 
                 child.matrixSet = [];
-
+                let tileMatrixSet,texturesCount;
```
peppsac: Nope. Replacing `var` by `let` is a good idea but don't move the declaration to a bigger scope.

src/Renderer/LayeredMaterial.js
```diff
@@ -177,152 +187,248 @@ LayeredMaterial.prototype.dispose = function() {
...
-LayeredMaterial.prototype.getNbColorTexturesLayer = function(layerIndex) {
-    return (this.paramLayers[layerIndex + 1].x || this.nbTextures[1]) - (this.paramLayers[layerIndex].x || 0);
+LayeredMaterial.prototype.pushLayer = function(param) {
+
+    let newIndex = this.getColorLayersCount();
+    let offset = newIndex === 0 ? 0 : this.getTextureOffsetByLayerIndex(newIndex-1) + this.getTextureCountByLayerIndex(newIndex-1);
+
+    this.uniforms.paramLayers.value[newIndex] = new THREE.Vector4();
+
+    this.setLayerOffset(newIndex,offset);
```
peppsac: You could rename `setLayerOffset` : `setTextureOffsetByLayerIndex` to match its getter

src/Scene/NodeProcess.js
```diff
@@ -176,16 +183,17 @@ NodeProcess.prototype.refineNodeLayers = function(node, camera, params) {
         updateNodeImagery
     ];
 
-    for (let i=0; i<2; i++) {
-        if (node.pendingLayers[i] === undefined
-            && (!node.loaded || node.downScaledLayer(i))) {
+    // WARNING don't use declaration function to reset flag
+    // var resetPendingFlag = function(id){node.pendingLayers[id] = undefined;};
```
peppsac: Remove this

src/Scene/NodeProcess.js
```diff
@@ -176,16 +183,17 @@ NodeProcess.prototype.refineNodeLayers = function(node, camera, params) {
...
 
-            node.pendingLayers[i] = true;
+    for (let typeLayer=0; typeLayer<2; typeLayer++) {
```
peppsac, line 185: It's probably a good call to rename `i` -> `layerType` but it could probably be defered to a specific cleanup PR.

src/Scene/NodeProcess.js
```diff
@@ -205,18 +213,12 @@ NodeProcess.prototype.hideNodeChildren = function(node) {
  */
 function findAncestorWithValidTextureForLayer(node, layerType, layer) {
     var parent = node.parent;
-    if (parent && parent.material && parent.material.getLayerTextureOffset) {
-        var slot = layerType == l_ELEVATION ? 0 : parent.material.getLayerTextureOffset(layer.id);
-        if (slot < 0) {
-            return null;
+    if (parent && parent.material && parent.material.getLayerLevel) {
+        var level = parent.material.getLayerLevel(layerType, layer);
+        if (level >= 0) {
```
peppsac: I'd prefer to keep the other order `0 <= level`
I like to only use `<` and `<=` and never use `>`. Of course in single statement it doesn't matter, but in more complex case :

``` js
if (a >= -2 && a <= 10)
if (-2 <= a && a <= 10)
```

I think the 2nd version is easier to read.

src/Renderer/LayeredMaterial.js
```diff
@@ -177,152 +187,248 @@ LayeredMaterial.prototype.dispose = function() {
...
+            if (this.Textures[l_COLOR][offset].level < level) {
+                return true;
+            }
+        }
+    }
+
+    return false;
 };
 
+LayeredMaterial.prototype.getColorLayerLevel = function(colorLayerId) {
```
peppsac: `getColorLevelByLayerId` ?
`getColorLayerLevelById` ?

src/Renderer/LayeredMaterial.js
```diff
@@ -41,132 +43,140 @@ var getColorAtIdUv = function(nbTex) {
...
+    this.offsetScale[l_COLOR] = Array(nbSamplers);
+    this.offsetScale[l_ELEVATION] = [vector];
+    fillArray(this.offsetScale[l_COLOR],vector) ;
 
     this.Textures[0] = [emptyTexture];
     this.Textures[1] = Array(nbSamplers);
-    this.paramLayers = Array(8);
-    this.paramBLayers = Array(8);
+    var paramLayers = Array(8);
+    this.layerTexturesCount = Array(8);
```
peppsac, line 118: This is `colorLayerTexturesCount`.
Also it's a bit annoying to duplicate texture count in `layerTextureCount` and `texturesCount` :(

src/Scene/NodeProcess.js
```diff
@@ -324,18 +322,9 @@ function updateNodeElevation(quadtree, node, layersConfig, force) {
...
-            if (currentElevation === -1) {
-                let a = findAncestorWithValidTextureForLayer(node, l_ELEVATION, layer);
-
-                if (a != null) {
-                    bestLayer = layer;
-                    ancestor = a;
-                    break;
-                }
-            }
+        let a = findAncestorWithValidTextureForLayer(node, l_ELEVATION);
```
peppsac: `ancestor = findAncestorWithValidTextureForLayer(node, l_ELEVATION);`

peppsac: Hmm you don't set `bestLayer`, is this on purpose?

src/Scene/NodeProcess.js
```diff
@@ -344,31 +333,29 @@ function updateNodeElevation(quadtree, node, layersConfig, force) {
...
     // Again, LayeredMaterial's 1 elevation texture limitation forces us to `break` as soon
     // as one layer can supply a texture for this node. So ordering of elevation layers is important.
-    if (bestLayer == null) {
-        for (let i = 0; i < elevationLayers.length; i++) {
-            let layer = elevationLayers[i];
-
-            if (!layer.tileInsideLimit(node, layer)) {
-                continue;
-            }
+    for (let i = elevationLayers.length - 1; i >= 0; i--) {
```
peppsac, line 330: I don't really mind changing the order here, but if it's important add a comment stating why we must browse elevation layer in this order.

src/Scene/NodeProcess.js
```diff
@@ -415,24 +402,19 @@ NodeProcess.prototype.processNode = function(node, camera, params) {
         if (sse && params.tree.canSubdivideNode(node)) {
             // big screen space error: subdivide node, display children if possible
             this.subdivideNode(node, camera, params);
-        }
-
-        if (!hidden)  {
+        } else if (!hidden && node.downScaledLayer(1)) {
```
peppsac: Don't use `1` magic value.
Also, are you not interested in refining elevation layers?

peppsac (changes requested):

src/Globe/TileMesh.js
gchoqueux: I rework this part

src/Renderer/LayeredMaterial.js
gchoqueux: I prefer work this in another PR

src/Scene/NodeProcess.js
gchoqueux: `bestLayer` is found in loop [here](https://github.com/iTowns/itowns2/blob/6826b6c61d485dc7f9023ca3a282111c81da9c88/src/Scene/NodeProcess.js#L353)
Maybe rework this in other PR with `LayeredMaterial.prototype.getColorLayerLevelById`

peppsac: bestLayer was supposed to be set at the same time than the ancestor. Or the 1st pass thing is useless (it's supposed to retrieve ancestor inital texture)

gchoqueux:
Updated and rebased

src/Scene/NodeProcess.js
gchoqueux: In `LayeredMaterial`, for the moment _elevation layer_ haven't the same mechanism that _color layer_ . _Elevation layer_ isn't generic like _color layer_. It's not possible for the moment to find `bestLayer` with `ancestor`. Because the TileMesh, LayeredMaterial or texture haven't information on  _elevation layer_'s identifiant. 
We need a new PR to make _Elevation layer_ more generic and and to superimpose _elevation layers_.

peppsac: Ok I get your point.
I agree that _elevation_ handling needs to be reworked in different PR (maybe issue #192 is relevant btw).

gchoqueux:
Rebased and fix problem ancestor elevation

gchoqueux:
updated and rebased

src/Scene/NodeProcess.js
```diff
@@ -324,51 +318,41 @@ function updateNodeElevation(quadtree, node, layersConfig, force) {
...
-            let targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node.level, currentElevation, layer.updateStrategy.options);
-            if (targetLevel <= currentElevation) {
-                continue;
-            }
+        let targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node.level, currentElevation, layer.updateStrategy.options);
+        if (targetLevel <= currentElevation) {
+            continue;
+        }
 
+        if(ancestor === null) {
```
peppsac, line 349: This change is weird and the behavior is not obvious.
I understand why you did this but this is acceptable only because it's meant to be temporary.

Add a comment explaining what is going on (either here or when you define `ancestor` line 321):

``` js
// ancestor is not enough: we also need to know from which layer we're going to request the elevation texture (see how this is done for color texture).
// Right now this is done in the `for` loop below but this is hacky because there's no real warranty that bestLayer and ancestor really match.
// FIXME: we need to be able to set both ancestor and bestLayer at the same time
```

src/Globe/Globe.js
```diff
@@ -208,6 +208,17 @@ Globe.prototype.updateLayersOrdering = function() {
     this.tiles.children[0].traverse(cO);
 };
 
+Globe.prototype.removeColorLayer = function(layer) {
+
+    var cO = function(object) {
+        if (object.removeColorLayer) {
+            object.removeColorLayer(layer);
+        }
+    }.bind(this);
```
peppsac: Is `bind(this)` needed?

src/Globe/TileMesh.js
```diff
@@ -308,54 +269,38 @@ TileMesh.prototype.OBB = function() {
...
-            if (layer && this.getIndexLayerColor(layer) === -1)
-                newSequence.splice(i, 1);
-        }
-    } else
-        newSequence = sequence;
+    var layerCount = this.materials[RendererConstant.FINAL].getColorLayersCount();
 
-    var sequenceMaterial = [];
+    // Quit if there is only one layer
+    if (layerCount < 2)
```
peppsac: Missing `{ }`

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,400 @@ var getColorAtIdUv = function(nbTex) {
...
+
+    // Remove visibility Parameters
+    this.uniforms.visibility.value.splice(layerIndex, 1);
+    this.uniforms.visibility.value.push(true);
 
-    for (var j = layerIndex, mx = this.paramLayers.length; j < mx; j++)
-        this.paramLayers[j].x -= nbTextures;
+    // Dispose Layers textures
+    for (i = offset, max = offset + texturesCount; i < max; i++) {
+        if (this.textures[l_COLOR][i] instanceof THREE.Texture)
```
peppsac: Missing `{ }`

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,400 @@ var getColorAtIdUv = function(nbTex) {
...
-    this.paramBLayers.splice(layerIndex, 1);
-    this.paramLayers.push(vector4);
-    this.paramBLayers.push(vector2);
+    let layerIndex = this.indexOfColorLayer(layer);
 
-    for (var i = startIdTexture, max = startIdTexture + nbTextures; i < max; i++) {
-        if (this.Textures[1][i] instanceof THREE.Texture)
-            this.Textures[1][i].dispose();
+    if (layerIndex === -1) {
+        return;
```
peppsac, line 231: At some point we'll need to stop silently ignoring errors and decide on a strategy (log errors ? throw ?)

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,400 @@ var getColorAtIdUv = function(nbTex) {
...
+        this.textures[l_COLOR].push(emptyTexture);
+        this.offsetScale[l_COLOR].push(vector);
     }
 
-    // fill the end's sequence
-    sequence[this.uniforms.nColorLayer.value] = this.uniforms.nColorLayer.value;
-    this.layerIdToIndex.idLayer = undefined;
-};
+    // Update slot start texture layer
+    for (var j = layerIndex, mx = this.getColorLayersCount(); j < mx; j++)
```
peppsac: Missing `{ }`

peppsac: Also use `let i` instead of `var j`

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,400 @@ var getColorAtIdUv = function(nbTex) {
...
-    if (texture === null) {
-        var l = this.slotToLayerIndex(slot);
-        if (l >= 0) {
-            this.paramLayers[l].z = 0;
+};
+
+LayeredMaterial.prototype.setTexturesLayer = function(textures, layerType, layer) {
+    var index = this.indexOfColorLayer(layer);
+    var slotOffset = this.getTextureOffsetByLayerIndex(index);
+    for (var i = 0, max = textures.length; i < max; i++) {
```
peppsac: Always use `let` for new code

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,400 @@ var getColorAtIdUv = function(nbTex) {
...
+    this.offsetScale[layerType][slot] = offsetScale ? offsetScale : new THREE.Vector3(0.0, 0.0, 1.0);
 };
 
-LayeredMaterial.prototype.slotToLayerIndex = function(slot) {
-    for (var l = 1; l < this.uniforms.nColorLayer.value; l++) {
-        if (slot < this.paramLayers[l].x) {
-            return l-1;
+LayeredMaterial.prototype.setColorLayerParameters = function(params) {
+    if(this.getColorLayersCount() === 0) {
+        for (var l = 0; l < params.length; l++) {
```
peppsac: style: `let` and `if (` (missing space)

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,400 @@ var getColorAtIdUv = function(nbTex) {
...
+LayeredMaterial.prototype.isLayerTypeDownscaled = function(layerType,level) {
+
+    if (layerType === l_ELEVATION) {
+        if (this.textures[l_ELEVATION][0].level < 0) {
+            return false;
+        } else {
+            return this.textures[l_ELEVATION][0].level < level;
+        }
+    } else if (layerType === l_COLOR) {
+        for (var index = 0, max = this.colorLayersId.length; index < max; index++) {
```
peppsac: `let` :)

peppsac (changes requested):

src/Renderer/LayeredMaterial.js
gchoqueux: It's necessary to choose a solution

gchoqueux:
updated and rebased

src/Globe/TileGeometry.js
```diff
@@ -213,15 +214,16 @@ TileGeometry.prototype.computeBuffers = function(params, builder) {
     }
 
     var iStart = idVertex;
-    var rmax = 5000;
-    var r = Math.max(rmax, Math.pow(rmax, 1 / params.zoom));
+
+    //TODO: WARNING beware size'skirt influence performance
+    var rmax = 50;
+    var r = Math.pow(rmax, 1 + 1.25/ params.zoom);
```
peppsac: Why did you change the formula to `1+1.25/zoom`?

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,402 @@ var getColorAtIdUv = function(nbTex) {
...
-    for (var i = startIdTexture, max = startIdTexture + nbTextures; i < max; i++) {
-        if (this.Textures[1][i] instanceof THREE.Texture)
-            this.Textures[1][i].dispose();
+    if (layerIndex === -1) {
+        return;
     }
 
-    this.Textures[1].splice(startIdTexture, nbTextures);
-    this.uniforms.nColorLayer.value--;
+    var offset = this.getTextureOffsetByLayerIndex(layerIndex);
```
peppsac: `var` -> `let`

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,402 @@ var getColorAtIdUv = function(nbTex) {
...
+    for (let i = offset, max = offset + texturesCount; i < max; i++) {
+        if (this.textures[l_COLOR][i] instanceof THREE.Texture) {
+            this.textures[l_COLOR][i].dispose();
+        }
+    }
 
-    // Rebuild sequence
-    var sequence = this.uniforms.layerSequence.value;
-    var limit = false;
+    var removedTexturesLayer = this.textures[l_COLOR].splice(offset, texturesCount);
```
peppsac: `var` -> `let`

src/Renderer/LayeredMaterial.js
```diff
@@ -41,288 +43,402 @@ var getColorAtIdUv = function(nbTex) {
...
 
-    for (var l = 0; l < this.uniforms.nColorLayer.value; l++) {
-        if (limit || sequence[l] === layerIndex) {
-            limit = true;
-            sequence[l] = sequence[l + 1];
+    for (let i = removedTexturesLayer.length - 1; i >= 0; i--) {
+        if(-1 < removedTexturesLayer[i].level) {
+            loadedTexturesLayerCount++;
         }
+    }
```
peppsac: For loops like this one can be shortened using `reduce` and arrow functions.

Something like this (not tested):

``` js
let loadedTexturesLayerCount = removedTexturesLayer.reduce((sum, texture) => sum + (-1 < texture.level), 0);
```

peppsac:
I left a few comments regarding style.
But I have a more important one: I think this PR introduces a regression in texture loading/flickering.

Here's an example of the issue. During zooming you can see the time where the child tile used a worse quality texture than its parent (red square):

crappy gif showing the bug

src/Globe/TileGeometry.js
gchoqueux: I simplify the formula. 
Skirt's size influence performance. For the moment, I reduce the size to increase performance (pixel shader performance)
But this calculation isn't correct. A correct calculation must to minimize the size without crack between tiles. This size must be take into account the bbox's size.

src/Renderer/LayeredMaterial.js
gchoqueux: ok, I change this

gchoqueux:
> I think this PR introduces a regression in texture loading/flickering.

Seems fixed

gchoqueux:
updated and rebased

peppsac (approved):

peppsac:
LGTM, merging.

PR #202, opened 2016-11-10
Error handling

peppsac:
This PR implements error handling for layers loading.

It introduces a new `LayerUpdateState` object holding fetch info: error count, last error timestamp etc.
`NodeProcess` can them use this to decide if it can tries to download texture from a given layer.

Currently there's a single error strategy implemented: it simply retries after longer and longer wait period. We could later implement different strategies if needed.

### DIFF
--- a/src/Core/Commander/ManagerCommands.js
+++ b/src/Core/Commander/ManagerCommands.js
@@ -22,9 +22,13 @@ function _instanciateQueue() {
             },
         }),
         counters: {
+            // commands in progress
             executing: 0,
+            // commands successfully executed
             executed: 0,
+            // commands failed
             failed: 0,
+            // commands cancelled
             cancelled: 0,
         },
         execute(cmd, provider, executingCounterUpToDate) {
@@ -36,15 +40,15 @@ function _instanciateQueue() {
             // Otherwise use a resolved Promise.
             var p = provider.executeCommand(cmd) || Promise.resolve();
 
-            return p.then(() => {
+            return p.then((result) => {
                 this.counters.executing--;
+                cmd.resolve(result);
                 this.counters.executed++;
-            },
-                        () => {
-                            this.counters.executing--;
-                            this.counters.executed++;
-                            this.counters.failed++;
-                        });
+            }, (err) => {
+                this.counters.executing--;
+                cmd.reject(err);
+                this.counters.failed++;
+            });
         },
     };
 }
@@ -119,8 +123,7 @@ ManagerCommands.prototype.addCommand = function addCommand(command) {
         }.bind(this);
 
         // We use a setTimeout to defer processing but we avoid the
-        // queue mechanism
-        // TODO: Why not use Promise?
+        // queue mechanism (why setTimeout and not Promise? see tasks vs microtasks priorities)
         window.setTimeout(runNow, 0);
     } else {
         q.storage.queue(command);
--- a/src/Core/Commander/ManagerCommands.js
+++ b/src/Core/Commander/ManagerCommands.js
@@ -43,6 +43,7 @@ function _instanciateQueue() {
             return p.then((result) => {
                 this.counters.executing--;
                 cmd.resolve(result);
+                // only count successul commands
                 this.counters.executed++;
             }, (err) => {
                 this.counters.executing--;
@@ -88,7 +89,7 @@ ManagerCommands.prototype.runCommand = function runCommand(command, queue, execu
         // We allow the scene to delay the update/repaint up to 100ms
         // to reduce CPU load (no need to perform an update on completion if we
         // know there's another one ending soon)
-        this.scene.notifyChange(100);
+        this.scene.notifyChange(100, true);
 
         // try to execute next command
         if (queue.counters.executing < this.maxCommandsPerHost) {
--- a/src/Core/Commander/ManagerCommands.js
+++ b/src/Core/Commander/ManagerCommands.js
@@ -173,6 +173,17 @@ ManagerCommands.prototype.getProviders = function getProviders() {
 };
 
 /**
+ * Custom error thrown when cancelling commands. Allows the caller to act differently if needed.
+ */
+function CancelledCommandException(command) {
+    this.command = command;
+}
+
+CancelledCommandException.prototype.toString = function toString() {
+    return `Cancelled command ${this.command.requester.id}/${this.command.layer.id}`;
+};
+
+/**
  */
 ManagerCommands.prototype.deQueue = function deQueue(queue) {
     var st = queue.storage;
@@ -181,7 +192,7 @@ ManagerCommands.prototype.deQueue = function deQueue(queue) {
 
         if (cmd.earlyDropFunction && cmd.earlyDropFunction(cmd)) {
             queue.counters.cancelled++;
-            cmd.reject(new Error(`command canceled ${cmd.requester.id}/${cmd.layer.id}`));
+            cmd.reject(new CancelledCommandException(cmd));
         } else {
             return cmd;
         }
@@ -196,6 +207,7 @@ ManagerCommands.prototype.wait = function wait() {
     this.eventsManager.wait();
 };
 
+export { CancelledCommandException };
 
 export default function (scene) {
     instanceCommandManager = instanceCommandManager || new ManagerCommands(scene);
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -126,14 +126,13 @@ WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbo
     var textureCache = this.cache.getRessource(url);
 
     if (textureCache !== undefined) {
-        return Promise.resolve(textureCache ? {
+        return Promise.resolve({
             pitch,
             texture: textureCache.texture,
             min: textureCache.min,
             max: textureCache.max,
-        } : null);
+        });
     }
-
 
     // bug #74
     // var limits = layer.tileMatrixSetLimits[coWMTS.zoom];
@@ -143,25 +142,18 @@ WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbo
     //     return Promise.resolve(texture);
     // }
     // -> bug #74
-
     return this._IoDriver.read(url).then((result) => {
-        if (result !== undefined) {
-            result.texture = this.getTextureFloat(result.floatArray);
-            result.texture.generateMipmaps = false;
-            result.texture.magFilter = THREE.LinearFilter;
-            result.texture.minFilter = THREE.LinearFilter;
-            result.pitch = pitch;
+        result.texture = this.getTextureFloat(result.floatArray);
+        result.texture.generateMipmaps = false;
+        result.texture.magFilter = THREE.LinearFilter;
+        result.texture.minFilter = THREE.LinearFilter;
+        result.pitch = pitch;
 
-            // In RGBA elevation texture LinearFilter give some errors with nodata value.
-            // need to rewrite sample function in shader
-            this.cache.addRessource(url, result);
+        // In RGBA elevation texture LinearFilter give some errors with nodata value.
+        // need to rewrite sample function in shader
+        this.cache.addRessource(url, result);
 
-            return result;
-        } else {
-            var texture = null;
-            this.cache.addRessource(url, texture);
-            return texture;
-        }
+        return result;
     });
 };
 
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -188,7 +188,7 @@ WMS_Provider.prototype.executeCommand = function executeCommand(command) {
             tile.bbox;
 
 
-        return func(tile, layer, bbox, pitch).then(result => command.resolve(result));
+        return func(tile, layer, bbox, pitch);
     } else {
         return Promise.reject(new Error(`Unsupported mimetype ${layer.format}`));
     }
--- a/src/Core/Commander/Providers/WMTS_Provider.js
+++ b/src/Core/Commander/Providers/WMTS_Provider.js
@@ -159,15 +159,6 @@ WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, pa
         result.texture.generateMipmaps = false;
         result.texture.magFilter = THREE.LinearFilter;
         result.texture.minFilter = THREE.LinearFilter;
-
-        // To compare with level tile
-        result.texture.url = result.url;
-
-        // In RGBA elevation texture LinearFilter give some errors with nodata value.
-        // need to rewrite sample function in shader
-        // result.texture.magFilter = THREE.NearestFilter;
-        // result.texture.minFilter = THREE.NearestFilter;
-
         this.cache.addRessource(url, { texture: result.texture, floatArray: result.floatArray });
 
         return result;
--- a/src/Core/Commander/Providers/WMTS_Provider.js
+++ b/src/Core/Commander/Providers/WMTS_Provider.js
@@ -226,7 +226,7 @@ WMTS_Provider.prototype.executeCommand = function executeCommand(command) {
 
     var func = supportedFormats[layer.options.mimetype];
     if (func) {
-        return func(tile, layer, command.paramsFunction).then(result => command.resolve(result));
+        return func(tile, layer, command.paramsFunction);
     } else {
         return Promise.reject(new Error(`Unsupported mimetype ${layer.options.mimetype}`));
     }
--- a/src/Core/Commander/Providers/WMTS_Provider.js
+++ b/src/Core/Commander/Providers/WMTS_Provider.js
@@ -13,6 +13,8 @@ import Fetcher from 'Core/Commander/Providers/Fetcher';
 import * as THREE from 'three';
 import CacheRessource from 'Core/Commander/Providers/CacheRessource';
 
+const SIZE_TEXTURE_TILE = 256;
+
 function WMTS_Provider(options) {
     // Constructor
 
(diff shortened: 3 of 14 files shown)


nosy-b:
When zooming in the app with default layers it raises:
 `Uncaught (in promise) TypeError: Cannot read property 'scene' of undefined
    at NodeProcess.js:413`
Three is still r80, is that normal?

nosy-b:
- Camera rotation (ctrl-mouseMove) is not working as before. Sometimes it doesn't take the projected center of the screen as pivot.

peppsac:
I fixed the `Uncaught (in promise) TypeError: Cannot read property 'scene' of undefined at NodeProcess.js:413` problem.

nosy-b:
This PR is a good idea because it is necessary to efficiently handle requests when no data is available but:
- In _LayerUpdateState.js_ 
  `const PAUSE_BETWEEN_ERRORS = [ 1000.0, 3000.0, 7000.0, 6000.0 ];`
  is not longer and longer ;)
- The _PAUSE_BETWEEN_ERRORS_  has no impact. It will still infinitely loop when you go outside of available data. The only way to kill this is to return false on update_error but it also block the nodeprocess.

The test I usually use: 
Zoom strongly to italy -> it launches infinite tile requests 
Zoom to french  DOM-TOM where high resolution data is available for the default index services->too much tiles in command to process actual visible ones

peppsac:
> In LayerUpdateState.js const PAUSE_BETWEEN_ERRORS = [ 1000.0, 3000.0, 7000.0, 6000.0 ]; is not longer and longer ;)

The last value was supposed to be `60000`...
And yes it's not strictly `longer and longer` - but a 60 sec pause beween retries seems like a good first step.

> The PAUSE_BETWEEN_ERRORS has no impact. It will still infinitely loop when you go outside of available data. The only way to kill this is to return false on update_error but it also block the nodeprocess.

Do you mean that it doesn't wait before retrying a request?

I'll try your testing strategy (thanks for sharing it) :)

nosy-b:
> Do you mean that it doesn't wait before retrying a request?

Yes it will request the same tiles (unavailable) infinitely at high speed

We should make the master version with a default behaviour that simply cut the request on _update_error_ or a too long _pending loading_ so _nodeprocess_ don't retry any command and don't go down the tree of this tile.

peppsac:
I pushed a fix for 2 typos in the code, it should work better now :)

peppsac:
Keep in mind that it's a first step toward error handling. The goal is to take a first stab at it and agree on a plan.

Also I know @autra is working on a PR unifying network accesses: it'll help with error handling.

nosy-b:
Ok I'll check future @autra PR.
For your last fix, it does stop the unlimited request when no data is available that's a good point!
But if you pause action with no data and zoom in again it will try new requests where ideally it should not go down the tree if the leaf has no data available. (Also still always impossible to see Ile de La Réunion also for example).
For the camera rotation bug do you know where it comes from, maybe not this PR? It happens when zooming from far and then rotating the camera, sometimes the rotation is not centered and if you do it again without moving it'll be centered.

autra:
> For the camera rotation bug do you know where it comes from, maybe not this PR? It happens when zooming from far and then rotating the camera, sometimes the rotation is not centered and if you do it again without moving it'll be centered.

FWIW I have this bug on master too, I seem to always go back to Africa ;-)

src/Core/Commander/ManagerCommands.js
```diff
@@ -178,7 +180,13 @@ ManagerCommands.prototype.getProviders = function() {
     return p;
 };
 
+function CancelledCommandException(command) {
+    this.command = command;
+}
 
+CancelledCommandException.prototype.toString = function() {
+    return 'Cancelled command ' + this.command + ' for node ' + this.command.requester;
+}
```
gchoqueux: add semi colon

src/Core/Commander/ManagerCommands.js
```diff
@@ -178,7 +180,13 @@ ManagerCommands.prototype.getProviders = function() {
     return p;
 };
 
+function CancelledCommandException(command) {
```
gchoqueux, line 182: Can you explain the addition of this object?

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -133,20 +133,17 @@ WMTS_Provider.prototype.getXbilTexture = function(tile, layer, parameters) {
...
-                    pitch,
-                    texture: textureCache.texture,
-                    min: minmax.min,
-                    max: minmax.max
-                });
-        }
-        return Promise.resolve(null);
+        let minmax = this._IoDriver.computeMinMaxElevation(
+            textureCache.floatArray,
+            256, 256,
```
gchoqueux: Create constant like :
```js
const SIZE_TEXTURE_TILE = 256;
```

src/Renderer/LayeredMaterial.js
```diff
@@ -320,9 +320,9 @@ LayeredMaterial.prototype.setNbLayersColor = function(n) {
 
 
 LayeredMaterial.prototype.getLayerLevel = function(id, slot) {
-    var level = this.Textures[id][slot || 0].level;
```
gchoqueux: var or let?

src/Scene/BrowseTree.js
```diff
@@ -119,10 +119,13 @@ BrowseTree.prototype._browseNonDisplayableNode = function(node, level, process,
...
     node.setDisplayed(false);
 
-    var sse = process.checkNodeSSE(node);
-
-    if (!sse && !node.loaded) {
+    const sse = process.checkNodeSSE(node);
+    if (!node.loaded) {
         // Make sure this node is not stuck in a !loaded state
+        // TODO: we could be smarter and do this only for visible tiles for instance
+        // but we need to be careful due to the number of possible states for a tile (pendingSubdivision,
```
gchoqueux, line 121: :+1:

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
+const UPDATE_IDLE = 0;
```
gchoqueux: little comment for this new object

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+        case UPDATE_PENDING: {
+            return false;
+        }
+        case UPDATE_ERROR:
+        default: {
+            let errorDuration = this.waitDurationUntilNextTry();
+            return errorDuration <= (timestamp - this.lastErrorTimestamp);
+        }
+    }
+}
```
gchoqueux: Missing semicolon

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+
+LayerUpdateState.prototype.waitDurationUntilNextTry = function() {
+    if (this.state != UPDATE_ERROR) {
+        return 0;
+    }
+    let idx =
+        Math.max(0, Math.min(this.errorCount, PAUSE_BETWEEN_ERRORS.length) - 1);
+
+    return PAUSE_BETWEEN_ERRORS[idx];
+}
```
gchoqueux: Missing semicolon

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+
+LayerUpdateState.prototype.try = function() {
+    this.state = UPDATE_PENDING;
+}
```
gchoqueux: Missing semicolon

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+
+LayerUpdateState.prototype.success = function() {
+    this.lastErrorTimestamp = 0;
+    this.state = UPDATE_IDLE;
+}
```
gchoqueux: Missing semicolon

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+
+LayerUpdateState.prototype.failure = function(timestamp) {
+    this.lastErrorTimestamp = timestamp;
+    this.state = UPDATE_ERROR;
+    this.errorCount++;
+}
```
gchoqueux: Missing semicolon

src/Scene/NodeProcess.js
```diff
@@ -290,23 +274,33 @@ function updateNodeImagery(quadtree, node, layersConfig, force) {
...
+                node.layerUpdateState[layer.id].success();
+
                 return result;
+            },
+            function(err) {
+                if (err instanceof CancelledCommandException) {
+                    node.layerUpdateState[layer.id].success();
+                } else {
+                    node.layerUpdateState[layer.id].failure(Date.now());
+                    scene.notifyChange(node.layerUpdateState[layer.id].waitDurationUntilNextTry(), false);
```
gchoqueux: We could instead dispatch an event system to notify a change.

gchoqueux:
This PR is in progress? it needs to rebase?

peppsac:
I fixed all your comments on a local branch except the last one (using an event).
And of course this PR needs a rebase...

(I removed the `ready` flag)

gchoqueux:
You will push your changes?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -145,12 +142,12 @@ WMS_Provider.prototype.getXbilTexture = function(tile, layer, bbox, pitch) {
     var textureCache = this.cache.getRessource(url);
 
     if (textureCache !== undefined) {
-        return Promise.resolve(textureCache ? {
+        return Promise.resolve({
```
autra, line 129: In this case, I would replace the if by 

```js
if (textureCache) {
```
Covering also the null case (and false and empty string). Btw the old code was resolving with null when textureCache was null. Was it used somewhere?

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -133,20 +133,17 @@ WMTS_Provider.prototype.getXbilTexture = function(tile, layer, parameters) {
     var textureCache = this.cache.getRessource(url);
 
     if (textureCache !== undefined) {
```
autra, line 132: `if (textureCache)` ?

src/Scene/LayerUpdateState.js
autra: Also, you can consider doing an "enum" (well, js-style):

```js
const UPDATE_STATE = {
    IDLE: 0,
    PENDING: 1,
    // ...
};
```

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+const PAUSE_BETWEEN_ERRORS = [ 1000.0, 3000.0, 7000.0, 60000.0 ];
+
+function LayerUpdateState() {
+    this.state = UPDATE_IDLE;
+    this.lastErrorTimestamp = 0;
+    this.errorCount = 0;
+
+}
+
+LayerUpdateState.prototype.constructor = LayerUpdateState;
```
autra: Only needed if you assign something to the prototype, which is not the case here.

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+        case UPDATE_ERROR:
+        default: {
+            let errorDuration = this.waitDurationUntilNextTry();
+            return errorDuration <= (timestamp - this.lastErrorTimestamp);
+        }
+    }
+}
+
+LayerUpdateState.prototype.waitDurationUntilNextTry = function() {
+    if (this.state != UPDATE_ERROR) {
```
autra: It's a good practice to use `!==` and `===` by default. `==` *can* be useful, but js implicit conversion rules are kind of... surprising sometimes, so if we can avoid them, it's better.

src/Scene/NodeProcess.js
```diff
@@ -432,7 +456,7 @@ NodeProcess.prototype.processNode = function(node, camera, params) {
     }
 
     return wasVisible || isVisible;
-}
+};
```
autra, line 452: <3

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+        case UPDATE_ERROR:
+        default: {
+            let errorDuration = this.waitDurationUntilNextTry();
+            return errorDuration <= (timestamp - this.lastErrorTimestamp);
+        }
+    }
+}
+
+LayerUpdateState.prototype.waitDurationUntilNextTry = function() {
```
autra: Not very fan of this method name. You're returning a time right? You're not actually waiting. `getWaitTimeUntilNextTry`?

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,57 @@
...
+    if (this.state != UPDATE_ERROR) {
+        return 0;
+    }
+    let idx =
+        Math.max(0, Math.min(this.errorCount, PAUSE_BETWEEN_ERRORS.length) - 1);
```
autra, line 40: If I'm understanding correctly, you never give up, but always try at least once per min right?

src/Scene/LayerUpdateState.js
autra: Ok actually `wait` was a noun :-) I think starting by `get` is a bit clearer.

src/Scene/LayerUpdateState.js
peppsac: Yes

src/Scene/LayerUpdateState.js
peppsac: I don't use `get`  for method that aren't accessors.
But I agree with your comment so I changed the name to: `secondsUntilNextTry()`

src/Core/Commander/Providers/WMS_Provider.js
peppsac: `null` value for elevation textures is gone with commit `Elevation textures can now have 3 states`.
 It was a special value used for error handling (sort-of).

src/Scene/NodeProcess.js
peppsac: Changing from `notifyChange` function to an event belongs to another PR (if we decide to do the change).

peppsac:
Long overdue branch update done :-) !

src/Scene/NodeProcess.js
```diff
@@ -280,22 +289,32 @@ function updateNodeImagery(quadtree, node, layersConfig, force) {
...
+                node.layerUpdateState[layer.id].success();
+
                 return result;
+            },
+            (err) => {
+                if (err instanceof CancelledCommandException) {
+                    node.layerUpdateState[layer.id].success();
+                } else {
+                    node.layerUpdateState[layer.id].failure(Date.now());
+                    scene.notifyChange(node.layerUpdateState[layer.id].secondsUntilNextTry() * 1000, false);
```
gchoqueux, line 302: Can you comment for next PR?
to replace `scene.notifyChange` by event.

```js
foo.dispatchEvent(eventLayerUpdateFailure);
```

src/Scene/Node.js
```diff
@@ -21,7 +21,7 @@ function Node() {
     this.screenSpaceError = 0.0;
     this.loaded = false;
     this.pendingSubdivision = false;
```
gchoqueux, line 24: it's possible to remove `this.pendingSubdivision` and use `this.layerUpdateState = {};`?

src/Core/Commander/ManagerCommands.js
```diff
@@ -84,7 +89,7 @@ ManagerCommands.prototype.runCommand = function runCommand(command, queue, execu
         // We allow the scene to delay the update/repaint up to 100ms
         // to reduce CPU load (no need to perform an update on completion if we
         // know there's another one ending soon)
-        this.scene.notifyChange(100);
+        this.scene.notifyChange(100, true);
```
gchoqueux, line 92: just for info : hard code value
needs TODO comment?

src/Renderer/LayeredMaterial.js
```diff
@@ -369,12 +369,20 @@ LayeredMaterial.prototype.isColorLayerDownscaled = function isColorLayerDownscal
...
     if (layerType === l_ELEVATION) {
-        if (this.textures[l_ELEVATION][0].level < 0) {
+        const tex = this.textures[l_ELEVATION][0];
+        // 3 possible cases
+        //   - initialization (no texture)
+        if (tex === undefined) {
+            return true;
+        }
+        //   - blank texture (eg: empty xbil texture)
+        if (tex.level === -1) {
```
gchoqueux: just for info : hard code value
needs TODO comment?

src/Scene/LayerUpdateState.js
```diff
@@ -0,0 +1,60 @@
...
+    ERROR: 2,
+};
+const PAUSE_BETWEEN_ERRORS = [1.0, 3.0, 7.0, 60.0];
+
+/**
+ * LayerUpdateState is the update state of a layer, for a given object (e.g tile).
+ * It stores information to allow smart update decisions, and especially network
+ * error handling.
+ */
+function LayerUpdateState() {
```
gchoqueux, line 13: You can try ES6 Class, if you want

src/Scene/NodeProcess.js
```diff
@@ -12,12 +12,13 @@ import Projection from 'Core/Geographic/Projection';
...
+import LayerUpdateState from 'Scene/LayerUpdateState';
+import { CancelledCommandException } from 'Core/Commander/ManagerCommands';
 
 export const SSE_SUBDIVISION_THRESHOLD = 6.0;
 
-function NodeProcess(camera, ellipsoid, bbox) {
-    // Constructor
-
+function NodeProcess(scene, camera, ellipsoid, bbox) {
+    this.scene = scene;
```
gchoqueux, line 22: TODO in new PR, remove `this.scene`

gchoqueux (changes requested):

src/Scene/Node.js
peppsac: Yes, but it requires quite a lot of refactoring because `pendingSubdivision` is used in several places that doesn't know about layers (eg TileMesh, Node), while `layerUpdateState` needs a layer.

I'll probably do it later though.

src/Core/Commander/ManagerCommands.js
peppsac: There's a 3 line comment above this value, explaining the rationale.

gchoqueux: `this.scene.notifyChange(100);` 
use `export const delayUpdate = 100`
and This value (100) may change depending on the computer

src/Scene/Node.js
gchoqueux: add TODO in the code and new issue

src/Core/Commander/ManagerCommands.js
peppsac: This change comes from an earlier commit (4218023e).

gchoqueux (approved):

Issue #203, opened 2016-11-14
ghpage and gh-pages & gh-pages not updated


vcoindet:
There are two gh-pages that might be in conflict.

**1 - gh-page:** 
This one is the original one (I think)
It's related to the demo website (http://www.itowns-project.org/itowns2/), but the demo website is not updated, same thing for the API documentation (http://www.itowns-project.org/itowns2/API_Doc/) where we should find the documentation of the function _setZoomLevel()_ (for example), but it's not the case...

**2 - ghpage:**
**What is this one ?**
It looks like it's a clone of the master...

**So my point is:** 
The autodeploy.sh is supposed to deploy **dist**, **index.html**, **API_Doc**, and **examples** in gh-pages so the demo is always updated.
But it seems that none of the last PR merged in the master (#194 and #197) were deployed correctly in the gh-pages branch as none of  **dist**, **index.html**, **API_Doc**, and **examples** are updated in here...

Why the last modifications are not in the demo website/gh-pages ? I think something went wrong but I don't understand what...

tbroyer:
Regarding `gh-pages`, there are jsdoc errors preventing the "deployment": https://travis-ci.org/iTowns/itowns2/jobs/174826498#L481
Because it's an `after_success` it does not affect the build result.
We may want to migrate to a `deploy`: https://docs.travis-ci.com/user/deployment/script and/or do the jsdoc during the build phase.

As for the `ghpage` branch, GitHub tells me (from my dashboard) that @gchoqueux created it 3 days ago (on 2016-11-11T11:58:27Z exactly)

BTW, we're still running Travis on Node 4 and 5; shouldn't we update to 6 and 7?

vcoindet:
OK, I couldn't find the travis log (#jobs965.1) until you post it here. 
I found the error, it seems that jsdoc doesn't support "<type>", I will correct it (the error is in ApiGlobe.js, as expected).

For the other points you describe, we should probably discuss it with the PSC. What do you think @gchoqueux @vpicavet @nosy-b ?

vcoindet:
Doc corrected in PR #204

tbroyer:
Updating Node.js and "linting" the JSDoc as part of `npm test` in PR #206

PR #206, opened 2016-11-15
Update Travis CI configuration

tbroyer:
Use up-to-date versions of Node.js and "lint" the JSDoc as part of the build.

We could test with Node.js 4 (still supported in LTS ['til April 2017](https://github.com/nodejs/LTS/)) using versions `lts/argon` and `lts/boron` (or `4` and `6` explicitly), but then we'll have to update the build again in April to remove `lts/argon` which would then be unsupported, then in October to add the next LTS release, and repeat that dance every 6 months. I think it's enough to test on the latest LTS (and latest non-LTS).

Related to #203

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - "4"
-  - "5"
+  - "lts/*"
+  - "node"
 after_success:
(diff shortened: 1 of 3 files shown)


peppsac:
LGTM merged (usgin the `Rebase and Merge` github button)

PR #209, opened 2016-11-17
iTowns JavaScript Style

gchoqueux:
### This PR is to discuss and work on **iTowns JavaScript Style**
**Tools used**
* [JSCS](http://jscs.info/) is a code style linter and formatter for your style guide
* Add jscs configuration file for iTowns (This preset is airbnb's preset modified)

**Sublime's Packages**
* [SublimeLinter-jscs](https://packagecontrol.io/packages/SublimeLinter-jscs)

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,92 @@
+{
+    "disallowSpacesInNamedFunctionExpression": {
+        "beforeOpeningRoundBrace": true
+    },
+    "disallowSpacesInFunctionDeclaration": {
+        "beforeOpeningRoundBrace": true
+    },
+    "disallowSpacesInsideBrackets": true,
+    "disallowEmptyBlocks": true,
+    "disallowSpacesInCallExpression": true,
+    "disallowSpacesInsideArrayBrackets": true,
+    "disallowSpacesInsideParentheses": true,
+    "disallowQuotedKeysInObjects": true,
+    "disallowSpaceAfterObjectKeys": true,
+    "requireSpaceBeforeObjectValues": true,
+    "disallowSpaceAfterPrefixUnaryOperators": true,
+    "disallowSpaceBeforePostfixUnaryOperators": true,
+    "disallowSpaceBeforeBinaryOperators": [
+        ","
+    ],
+    "disallowMixedSpacesAndTabs": true,
+    "disallowTrailingWhitespace": true,
+    "requireTrailingComma": { "ignoreSingleLine": true },
+    "requireSpaceAfterComma": true,
+    "disallowYodaConditions": true,
+    "disallowKeywords": [ "with" ],
+    "disallowKeywordsOnNewLine": ["else"],
+    "disallowMultipleLineBreaks": true,
+    "disallowMultipleLineStrings": true,
+    "disallowMultipleVarDecl": true,
+    "disallowSpaceBeforeComma": true,
+    "disallowSpaceBeforeSemicolon": true,
+    "disallowSpacesInsideTemplateStringPlaceholders": true,
+    "disallowArrayDestructuringReturn": true,
+    "requireSpaceBeforeBlockStatements": true,
+    "requireParenthesesAroundIIFE": true,
+    // Original next is uncommented
+    /*
+    "requireSpacesInAnonymousFunctionExpression": {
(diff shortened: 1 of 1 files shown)


peppsac:
The homepage of JSCS lint states that `JSCS has merged with ESLint!`

Maybe we should use directly `ESLint` instead?

peppsac:
Another candidate is clang-format. See http://probst.io/blog/2015/01/15/clang-format-for-javascript/ for some examples.

Advantages:
- fast :)
- easy to integrate in text editor (at least on Linux/Mac)
- you can use the same tool (and rule!) to both format your code and as a pre-commit check / or CI test (see http://stackoverflow.com/questions/22866609/can-clang-format-tell-me-if-formatting-changes-are-necessary#comment63404587_22866610 for a possible way to check)

To test: `clang-format -style="{BasedOnStyle: Mozilla, IndentWidth: 4}" src/Scene/NodeProcess.js`

tbroyer:
We already have ESLint (use `npm run lint -- --fix` to "fix" the code to conform to the configuration), which is already quite fast and with good integration within editors (in Visual Studio Code, I get on-the-fly linting, and automatic reformat/fix on save). If you want a pre-commit hook, it shouldn't be hard to modify the sample hook to call `npm run lint` –to lint everything–, or possibly `eslint`/`node_modules/.bin/eslint` with the list of added/modified JS files)
I don't see a clear advantage to clang-format, and it would mean keeping ESLint and clang-format configurations in sync.

Anyway, I think we can close this PR as JSCS is basically a no-go nowadays (and we already have ESLint).

peppsac:
Let's leave this PR open so we can continue the discussion.

I've tested the `--fix` option of eslint, both on a single file (NodeProcess.js) and the whole project.
It has been unable to fix many errors - maybe my setup is wrong but I'm a bit disappointed by the tool.

I suggested `clang-format` as a replacement to `eslint` : I don't want to maintain 2 config files :) (at least it can fix all the errors it reports :p)

peppsac:
To be clear: I don't care very much about the tool used.

But I think it's important to have a plan to go from the current state (thousands of errors) to the desired states (no error and a way to make sure they don't come back).

tbroyer:
I could agree if ESLint was only about formatting, but it's not: http://eslint.org/docs/rules/
We can use both tools if we remove our (2!) rules about formatting in ESLint: use ESLint for possible errors and other non-formatting-related guidelines, and clang-format to format the code.
Wrt to your failed experiment: we did not configure ESLint for **any** format style besides forbidding trailing spaces and requiring that files end with an EOL.
(also note that we have a `.editorconfig` for LF end-of-lines and indent size and style; that one we'd have to either update or possibly delete if we used clang-format)

---

BTW, we can continue to discuss on a closed PR :-)
I was concerned that we might merge it by mistake.

autra:
My 2 cents: eslint is the de-facto tool in the js world so we won't be wrong if we stick with it. That being said, I'm going to have a look at clang-format this afternoon to weight the pro/cons a bit.

Also, I have started to have a look at the airbnb conf for eslint. It still needs a bit of tinkering to work correctly for us (it needs to be correctly interfaced with webpack for instance). I'm currently looking at it and I'll try to make a PR soon, WDYT?

peppsac:
> we did not configure ESLint for any format style

Yep. But I used the airbnb configuration during my tests.
But as @autra found, the config needs some adjustements to work on iTowns code.

Looking forward to an updated PR :-)

nosy-b:
@autra I think it's a good point!

autra:
About clang-format : AFAIK it is not really a linter but just a formatter. So it won't catch nearly as much things as eslint (random example : consistent return, default case in switch statement etc...).

If we realize eslint is bad at automatic formatting, it could still be good to think about clang-format to fill the void though. I'm sure it won't be that hard to configure it similarly to eslint. 

More info after my attempt at eslinting the whole project ;-)

autra:
Ok, sorry if my comments are a bit off: I was in the train, and didn't get half your comments before submitting mine :-/

peppsac:
Closing (again):
* as stated earlier discussion can continue even if the PR was closed
* see PR #212 is another implementation

PR #210, opened 2016-11-17
Fix race condition in the initialisation of the Scene.

autra:
Doing Quadtree.init with the geometry layer triggers and update of the
scene, which later causes subdivision to be made to the first 3 levels.
When this subdivision is made, we want to be absolutely sure that
LayerConfiguration is complete. Otherwise, layers textures won't be
downloaded if they are added after this initialisation.

FYI, this race conditions was triggered (most of the time) in Firefox but not in chromium.

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -95,6 +95,23 @@ function preprocessLayer(layer, provider) {
 }
 
 /**
+ * Init the geometry layer of the Scene.
+ */
+ApiGlobe.prototype.init = function () {
+    var map = this.scene.getMap();
+    map.tiles.init(map.layersConfiguration.getGeometryLayers()[0]);
+};
+
+/**
+ * Add the geometry layer to the scene.
+ */
+ApiGlobe.prototype.addGeometryLayer = function(layer) {
+    preprocessLayer(layer, this.scene.managerCommand.getProtocolProvider(layer.protocol));
+    var map = this.scene.getMap();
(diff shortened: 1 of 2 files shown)


src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -95,6 +95,23 @@ function preprocessLayer(layer, provider) {
 }
 
 /**
+ * Init the geometry layer of the Scene.
+ */
+ApiGlobe.prototype.init = function () {
+    var map = this.scene.getMap();
```
peppsac: `let` instead of `var`

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -95,6 +95,23 @@ function preprocessLayer(layer, provider) {
...
+    map.tiles.init(map.layersConfiguration.getGeometryLayers()[0]);
+};
+
+/**
+ * Add the geometry layer to the scene.
+ */
+ApiGlobe.prototype.addGeometryLayer = function(layer) {
+    preprocessLayer(layer, this.scene.managerCommand.getProtocolProvider(layer.protocol));
+    var map = this.scene.getMap();
```
peppsac: Use `let`

peppsac (approved):
Looks good.

autra:
Updated. @gchoqueux @nosy-b or @tbroyer WDYT?

nosy-b:
LGTM

autra:
@nosy-b wait I didn't squash my commits :astonished:  :grinning:

PR #211, opened 2016-11-18
Net refacto

autra:
Some simplifications of the network layer. I'm wondering if my Fetcher.js should go in another directory (for example `src/Core/Web`) ?


autra:
@iTowns/reviewers thanks for the review

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -0,0 +1,41 @@
...
+            res = resolve;
+            rej = reject;
+        });
+
+        let texture = textureLoader.load(url, res, () => {}, rej);
+        return {texture, promise};
+    }
+}
+
+export default Fetcher;
```
tbroyer: Instead of `export default Fetcher` how about exporting each function instead?

``` js
export function json(url) {
   …
}
export function xml(url) {
   …
}
export function texture(url) {
   …
}
```

tbroyer (commented):
I approve those changes but will let others "formally" approve them as there might have been reasons to go with all those `IoDriver`s.

src/Core/Commander/Providers/Fetcher.js
peppsac: Or maybe use a verb instead of a noun : `Fetch.xml` or `Get.xml` ?

peppsac (changes requested):
This PR breaks GPX loading - bisecting indicates that `3ee59564f900617f542162cfd63d962973fb32f3 is the first bad commit`.

To reproduce you can put `agpxfile.gpx` in dist and then: 
`itowns.viewer.loadGPX('http://localhost:8080/dist/agfpxfile.gpx')`

peppsac:
Here's a quick fix:

``` js
-            return new window.DOMParser().parseFromString(response.text(), 'text/xml');
+            return response.text().then(text => new window.DOMParser().parseFromString(text, 'text/xml'));
```

src/Core/Commander/Providers/Fetcher.js
autra: > how about exporting each function instead?

Oh yeah indeed :-)

> use a verb instead of a noun : Fetch.xml or Get.xml

you mean fetchXml, fetchJson?

peppsac: What's the benefit of exporting functions?

I'd prefer `let x = Get.xml(url);` over `let x = xml(url);`.

But you can do the same using `import * as Get from 'Fetcher'`

autra:
@iTowns/reviewers updated. WARNING this PR needs rebasing before merging :-)

autra:
And I forgot to change the export, so definitely not ready to be merged actually :-)

src/Core/Commander/Providers/Fetcher.js
autra: So yeah, after thinking about this a bit: we must find the right granularity for our exports, neither too big nor too small. Generally speaking, I believe the right level is the *service* as opposed to a single function here. So I would prefer to keep the PR as is.

Then for the naming issue, I like `Fetcher` because we can then refer to it as an entity. Although `Get` is also nice because it gives the code a native language feel, I still have a slight preference for the former, but no strong opinion.

That being said, whether we choose one or the other, I'm not very fond of `import * as Get from 'Fetcher'`: I think it will only be confusing for the reader, who will have to scroll up to check which javascript object we are really talking about.

TL;DR I prefer my way :-D

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -22,7 +22,9 @@ let Fetcher = {
     xml: function(url) {
         return fetch(url).then(response => {
             checkResponse(response);
-            return new window.DOMParser().parseFromString(response.text(), 'text/xml');
+            return response.text();
+        }).then(text => {
+            return new window.DOMParser().parseFromString(text, 'text/xml');
```
tbroyer: nit: you could remove the braces and the `return` here.

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -0,0 +1,43 @@
...
+    if (response.status < 200 || response.status >= 300) {
+        var error = new Error(`Error loading ${response.url}: status ${response.status}`);
+        error.status = response.status;
+        throw error;
+    }
+}
+
+let Fetcher = {
+
+    json: function(url) {
```
tbroyer: nit: how about the "method definition" syntax?

```js
{
  json(url) { … }
  xml(url) { … }
  texture(url) { … }
}
```

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -0,0 +1,43 @@
...
+
+function checkResponse(response) {
+    if (response.status < 200 || response.status >= 300) {
+        var error = new Error(`Error loading ${response.url}: status ${response.status}`);
+        error.status = response.status;
+        throw error;
+    }
+}
+
+let Fetcher = {
```
tbroyer: nit: does the object need to be named?

```js
export default {
  …
};
```

src/Core/Commander/Providers/Fetcher.js
tbroyer: Wrt. naming, `Fetch` vs. `Fetcher`, can't decide…

tbroyer (approved):
SGTM, only suggesting stylistic tweaks.

(BTW, sorry for the spam, I keep clicking "Add single comment" instead of "start review"; don't know why…)

src/Core/Commander/Providers/Fetcher.js
autra: Yep, and actually I need to do it this way for the linter in #212 :-)

autra:
@tbroyer @peppsac updated. Don't forget, I need to squash before merging :-)

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -39,5 +39,3 @@ export default {
         return {texture, promise};
     }
 }
-
-export default Fetcher;
```
tbroyer: I believe that changed line should go in the previous commit.

autra: Yep, updated.

peppsac (approved):
LGTM - except I still prefer the Get.xml() form but it's not a big deal :)

autra:
I have squashed the fixup. Is this ready to be merged?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -134,7 +133,7 @@ ApiGlobe.prototype.addImageryLayer = function(layer) {
  */
 
 ApiGlobe.prototype.addImageryLayerFromJSON = function(url) {
-     return JSONDriver.read(url).then(function(result){
+     return Fetcher.json(url).then(function(result){
```
gchoqueux: add space before {

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -238,7 +237,7 @@ ApiGlobe.prototype.addElevationLayer = function(layer) {
  */
 
 ApiGlobe.prototype.addElevationLayersFromJSON = function(url) {
-     return JSONDriver.read(url).then(function(result){
+     return Fetcher.json(url).then(function(result){
```
gchoqueux: add space before {

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -0,0 +1,41 @@
...
+        let res, rej;
+        let promise = new Promise(function(resolve, reject) {
+            res = resolve;
+            rej = reject;
+        });
+
+        let texture = textureLoader.load(url, res, () => {}, rej);
+        return {texture, promise};
+    }
+}
```
gchoqueux: add semicolon

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -109,32 +106,21 @@ WMS_Provider.prototype.getColorTexture = function(tile, layer, bbox, pitch) {
...
-            result.texture = texture;
-        else
-        {
-            result.texture = new THREE.Texture(image);
-            result.texture.needsUpdate = true;
-            result.texture.generateMipmaps = false;
-            result.texture.magFilter = THREE.LinearFilter;
-            result.texture.minFilter = THREE.LinearFilter;
-            result.texture.anisotropy = 16;
-            result.texture.url = url;
```
gchoqueux, line 122: Why you remove `result.texture.url = url;`?

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -200,39 +197,26 @@ WMTS_Provider.prototype.getColorTexture = function(coWMTS, pitch, layer) {
...
-        if (texture)
-            result.texture = texture;
-        else {
-            result.texture = new THREE.Texture(image);
-            result.texture.needsUpdate = true;
-            result.texture.generateMipmaps = false;
-            result.texture.magFilter = THREE.LinearFilter;
-            result.texture.minFilter = THREE.LinearFilter;
-            result.texture.anisotropy = 16;
-            result.texture.url = url;
```
gchoqueux, line 216: Why you remove `result.texture.url = url;`?

src/Core/Commander/Providers/Fetcher.js
```diff
@@ -0,0 +1,41 @@
+import { TextureLoader } from 'three';
```
gchoqueux, line 1: I know that THREE is present everywhere in code, but it'll be a good idea to limit its expansion in iTowns.

gchoqueux (changes requested):

gchoqueux:
There's a bug in globe loading : some tiles flick (texture->blue->texture).

src/Core/Commander/Providers/Fetcher.js
autra: What worries you exactly?

gchoqueux: This is not yet the case for the moment, but a good principle will be not to contaminate our code with that of THREE.js. (Especially for loading)
The Fetcher can return an image and not a texture.

autra:
> There's a bug in globe loading : some tiles flick (texture->blue->texture).

Yep indeed. I'm looking at it.

src/Core/Commander/Providers/WMS_Provider.js
autra: I thought it was not used anywhere, but I'll double-check

autra: So yeah, I couldn't see any use of this parameter in the code, but I may have missed something. Don't hesitate to point potential problems to me.

src/Core/Commander/Providers/Fetcher.js
autra: > (Especially for loading)

What do you mean by that? Importing it one or n times won't change the loading time, it's just a scope thing AFAIK (hopefully so, otherwise `import` syntax would have to be avoided at all cost!)

>  The Fetcher can return an image and not a texture.

That would mean nearly exactly duplicating what Three.js already does (see [TextureLoader](https://github.com/mrdoob/three.js/blob/6c7f000734f8579da37fb39e5c2e9e5e2dfb14f8/src/loaders/TextureLoader.js)). So a bigger code (so slower to load), that we need to maintain.

I feel like I'm missing the point: why importing and using an utility function is "contaminating the code"? And why do we want to limit the "expansion of three.js in itowns"? (I don't know what you mean by "expansion" and "contamination" btw). Do we plan on being agnostic regarding the WebGL lib?

Or do you mean you feel this function should be elsewhere?

peppsac: I'm with @autra on this one.
I prefer to take full advantage of our chosen framework (three.js). We already have way bigger dependencies (Mesh, Materials, Vectors) so I don't see the value of *not* using the `TextureLoader` function.
Actually I'd rather drop our limited reimplementation of three.js features: `Camera` and `Scene` for instance but that's another story :)

autra:
@gchoqueux I corrected the tile flicking and the nits.

src/Core/Commander/Providers/Fetcher.js
gchoqueux: > We already have way bigger dependencies (Mesh, Materials, Vectors)

It's true and normal, the use of THREE should only encapsulate in objects in Renderer folder.
I would like to have the minimum use of THREE in folders Scene, Core and Globe. It would be better in terms of design.

Ok for the moment, it's not very important, so keep textureLoader in Fetcher

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux: Ok, remove `url`

src/Core/Commander/Providers/WMTS_Provider.js
gchoqueux: Ok, remove `url`

gchoqueux (approved):

gchoqueux:
r+

PR #212, opened 2016-11-21
Eslint conf

autra:
Following the discussion in https://github.com/iTowns/itowns2/pull/209, here is a conf proposal for eslint.

What I did for the first commit: 
- update eslint (new version is required for airbnb conf) and eslint-loader, and install  `eslint-plugin-import`, `eslint-import-resolver-webpack` (so that eslint undertands our import statement with webpack) and [`eslint-config-airbnb-base`](https://www.npmjs.com/package/eslint-config-airbnb-base). I preferred this to [`eslint-config-airbnb`](https://www.npmjs.com/package/eslint-config-airbnb) because the latter only adds linter conf for React.
- run `eslint "src/**/*.js" "test/**/*.js" --fix` (for some reason, the --fix option does not work with `npm run lint`)
- disable the rules that were not fixed by --fix to have a 0 error state in the project.

So for first commit, the 2 really important changes are in `package.json` and `.eslintrc.js`. That being said, I guess it would be good to have a quick look at the other files : the --fix process got a bit confused in some places during the process, and I had to fix 2 or 3 silly dangling commas by hand. I will go through them, but 2 pairs of eyes is always better than one :-)

Commit 2 and 3 fixes `no-useless-concat` and `no-lonely-if` errors, as examples. I will comment on each rules, giving a short explanation of what they are doing. 

Most notably, this PR changes the indentation to be 2 spaces instead of 4.


tbroyer:
> run `eslint "src/**/*.js" "test/**/*.js" --fix` (for some reason, the --fix option does not work with `npm run lint`)

Did you run it as `npm run lint --fix` (won't work) or `npm run lint -- --fix`?

.eslintrc.js
```diff
@@ -1,21 +1,84 @@
...
-    // Stylistic rules
-    "eol-last": "error",
-    "no-trailing-spaces": "error",
+    // TODO reactivate all these rules
+
+    // The following rule should be no more restrictive than
+    //'no-plusplus': ['error', { 'allowForLoopAfterthoughts': true }],
+    // to allow for i++ in for loops (even though we should mostly use for..of
+    // anyway)
+    'no-plusplus': 'off',
```
autra, line 25: I'm not sure we want to reactivate this one, see http://eslint.org/docs/rules/no-plusplus

Honestly, we just need to be careful not to put a var and a ++ in 2 different lines. WDYT?

.eslintrc.js
```diff
@@ -1,21 +1,84 @@
...
+
+    'func-names': 'off',
+    'no-mixed-operators': 'off',
+    'no-use-before-define': 'off',
+    'no-underscore-dangle': 'off',
```
autra, line 51: In Itowns, we had used this one as a way of saying "this field in private". I'm in favor of keeping this though, because usually we can enforce the privacy of something by moving it in a closure or outside of the exported object.

.eslintrc.js
```diff
@@ -1,21 +1,84 @@
...
+    'eqeqeq': 'off',
+    // what len ? Airbnb does 100. github wraps line above 80
+    'max-len': 'off',
+    'no-param-reassign': 'off',
+    'no-else-return': 'off',
+    'prefer-const': 'off',
+    'one-var': 'off',
```
autra, line 58: So I'm not even sure which one of `never` (several var, let, const) or `always` (only one `var`, `let` and `const` in a scope) we want to choose...

autra:
> Did you run it as npm run lint --fix (won't work) or npm run lint -- --fix?

The former, thanks for the tip !

.eslintrc.js
tbroyer: Fwiw, my vote goes for `never`.

.eslintrc.js
peppsac: Agreed.

.eslintrc.js
peppsac: +1 for `never`

peppsac:
Random comments:
* 2 spaces indentation. I'm fine with this change but maybe this should explicitely stated in the PR summary.
* I'd love to have a rule that forbids mixing `<` and `>` operators in an expression. Is it possible and is this something other dev would want?
* why `NodeProcess(camera,...)` (no space before `(`) and `NodeProcess.prototype.backFaceCulling = function (node...)` ?
* is there a column-width limit?
* Some comments are weirdly indentend, e.g in `NodeProcess.js`:
```js
    }
    if (!force) {
            // does this tile needs a new texture?
      if (!node.downScaledColorLayer(layer.id)) {
```

autra:
> * I'd love to have a rule that forbids mixing `<` and `>` operators in an expression. Is it possible and is this something other dev would want?

you mean something that forbid condition like `foo < bar && bazz > bar` and the like? I don't know if this exists. If you find (or implement) it, I'll be happy to add it :-)

> * why `NodeProcess(camera,...)` (no space before `(`) and `NodeProcess.prototype.backFaceCulling = function (node...)` ?

I think it is : space before keyword, no space before identifier.

> * is there a column-width limit?

Currently disable (need to fix the code, apparently --fix was not implemented for this rule. And I'm actually not sure it is easily doable because of automatic semicolon insertion), but at 100 char in airbnb conf. See comments in `.eslintrc`

> * Some comments are weirdly indentend, e.g in `NodeProcess.js`:
> ```js
>     }
>     if (!force) {
>             // does this tile needs a new texture?
>       if (!node.downScaledColorLayer(layer.id)) {
> ```

I hoped no one would notice :-) Apparently --fix does not touch the comments... I'll see what I can do.

.eslintrc.js
gchoqueux: +1 for never

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -12,36 +12,35 @@ import WMTS_Provider from 'Core/Commander/Providers/WMTS_Provider';
...
+let loaded = false;
+const eventLoaded = new CustomEvent('globe-loaded');
+const eventRange = new CustomEvent('rangeChanged');
+const eventCenter = new CustomEvent('centerchanged');
+const eventOrientation = new CustomEvent('orientationchanged');
+const eventLayerAdded = new CustomEvent('layeradded');
+const eventLayerRemoved = new CustomEvent('layerremoved');
+const eventLayerChanged = new CustomEvent('layerchanged');
+const eventLayerChangedVisible = new CustomEvent('layerchanged:visible');
+const eventLayerChangedOpacity = new CustomEvent('layerchanged:opacity');
```
gchoqueux: Several events aren't constant

.eslintrc.js
gchoqueux: agreed

gchoqueux:
I would like to keep 4 spaces indentation
* In general, it's much more readable. 
* For this PR, It's more difficult to find errors, like [this comment](https://github.com/iTowns/itowns2/pull/212#discussion_r89072755)

Add this rule for 4 spaces indentation
```js
'indent': ["error", 4],
```

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
peppsac: IMHO if they are not constant they shouldn't be declared as top scope variables.

gchoqueux: Ok
Where?

.eslintrc.js
autra: Ok, I'm putting never in the next version

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
peppsac: Do:
```js
this.viewerDiv.dispatchEvent(new CustomEvent('layeradded'));
```
or
```js
const LayerAddedEvent = 'layeradded'; // top-scope
[...]
this.viewerDiv.dispatchEvent(new CustomEvent(LayerAddedEvent)); // fn-scope
```
Instead of:
```js
var eventLayerAdded = new CustomEvent('layeradded'); // top-scope
[...]
this.viewerDiv.dispatchEvent(eventLayerAdded); // fn-scope
```

autra: Er, just stop me if I'm mistaken, but these variables are not reassigned, are they? So they *are* const. Const does *not* mean you can't do `myobj.prop = 'foo'`, it just means you can't do `myobj = 'bar` a second time.

autra:
For the indentation, I'll argue it's a matter of getting used to it. I do prefer 2 spaces (I find 4 spaces too much, especially with a line width limit, which we will have). But as it is also a matter of taste, I let you @iTowns/reviewers guys decide among yourselves :-)

autra:
> For this PR, It's more difficult to find errors

I don't see why? Github does highlight things correctly, and has different colors depending if the change is only white spaces or anything else. I'm not sure I see your points, could you elaborate pls?

Btw, I'm pretty sure [the comment you linked](https://github.com/iTowns/itowns2/pull/212#discussion_r89072755) is actually valid code.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
peppsac: Yeah, javascript `const` should mean `immutable` but...
So I'd suggest to not modify your PR for this particular point :)

gchoqueux:
It's possible to use tab and to set the size of the tab in its editor?

peppsac:
There are only 2 reasonnable options:
* 4 spaces: because it's used today in itowns codebase
* 2 spaces: which seems to be the most used settings (airbnb, google, mozilla, etc)

2 space if my favorite, since it reduces lines width (but if we make this change we need to update .editorconfig)

autra:
I pushed a new version. Sorry, I didn't make fixup commits, but that's near impossible considering the number of affected files. For this reason, I'll be grateful if we manage to merge that soon (although let's take time to debate on these important decisions) because the rebase process is quite painful :-) 

ATM I have let indentation at 2 spaces, I'll change if we decide otherwise. Also, I have removed the 2 following commits correcting 2 rules. it was only cluttering this PR.

I have also keep the `const` modifications for now, I'll change it too when our discussion will be over. Thanks!

tbroyer:
Let the bikeshedding begin!

Put differently: you don't have to agree with everything in a codestyle, you only have to agree to following it (and having a tool that fixes your code to conform to the codestyle greatly helps); there's a [Go Proverb](https://go-proverbs.github.io/) that says _“Gofmt's style is no one's favorite, yet gofmt is everyone's favorite.”_
JavaScript tentatively has such a thing with [Standard JS](http://standardjs.com/), or you can go with a renowned codestyle (AirBnB, Google, etc.) **and keep it as-is**.
Any other choice will inevitably lead to never-ending discussions (bikeshedding).

IIUC, @autra disabled many rules from the AirBnB codestyle because the code needs fixing, but the goal would (should!) be to re-enable **all** of them, without exception (unless everyone can agree such as with  `no-plusplus`; well, maybe… I'm not even sure…)

gchoqueux:
@autra 
> For this PR, It's more difficult to find errors

Yes, github does highlight things correctly, but there's enormous diff because of 2 spaces indendations. 
It takes a lot of time to parse the code.

autra:
And btw: I've disabled `no-plusplus` ;-)

autra:
@gchoqueux oh I see. Regardless of the result of this debate, maybe `git diff -w  --word-diff-regex=[^[:space:]]` can help?

gchoqueux:
> never-ending discussions 

To avoid the never-ending discussions there's the vote

gchoqueux:
>  `git diff -w  --word-diff-regex=[^[:space:]]`

Yes, but it's not possible in github reviewing?

autra:
Ok, let's vote maybe ? So dear @iTowns/reviewers, do you prefer:

1. 2 spaces indentation
1. 4 spaces indentation

- [x] @peppsac : 2 spaces
- [x] @tbroyer : 2 spaces
- [x] @gchoqueux : 4 spaces
- [x] @nosy-b : 2 spaces

autra:
> Yes, but it's not possible in github reviewing?

No. A pity if you ask me :-)

tbroyer:
> > Yes, but it's not possible in github reviewing?
>
> No. A pity if you ask me :-)

Actually, yes there is, but hidden: add `?w=1` to the URL.
See https://github.com/blog/967-github-secrets

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: Sorry, I didn't know that it's possible to add key to constant object. 
So, it's possible to use `const`for the events.

autra:
A tie! I guess the status quo (aka 4 spaces) wins in this case, right @tbroyer @peppsac ?

autra:
So I added a commit that revert to 4 spaces. 

Beware, this PR needs squashing before being mergedI

nosy-b:
I didn't participate actively to  this PR but personally I think it would be good to simply follow an existing popular coding style such as airbnb without modification. So I could move to the 2 spaces easily ;)

autra:
@nosy-b so do you vote 2 or 4? I'm going to have lunch. In the meantime, iif you want to change your vote, you can. 

When I go back, I propose that if it's still a tie, we keep the current majoritary style, ie 4 spaces.

peppsac:
LGTM! (regardless of the 2-spaces / 4-spaces setting )

peppsac:
Merged, thanks everyone :)

autra:
Now, we "just" need to reactivate every rules and correct all the things.

(we were at ~6000 errors before I turned them off, just sayin' ;-) )

nosy-b:
Temporarly corrected the 1b35dc9c1fcfadf201d79325abf29884d4ed33de with 'linebreak-style': 'off' which was causing hundreds of errors on win
 [Better solutions might exist](http://eslint.org/docs/rules/linebreak-style)

PR #195, opened 2016-10-11
Web Feature Service part 1

mlamure:
This PR contains the following changes :
- Support multiple layers for wfs
- Add 3D Building (RGE in 3D)
- Support multiple functionalities (addFeature, removeFeature, pickFeature, add/geTileMatrixSets)
- Support standard geojson parsing (point, linestring, polygon, multipolygon, etc)
- Add different type of features (point box, line)
- Support the change of type of feature depending on the level or on the selected area (manage the change with the interface / the selected area can be changed by pressing D with a right click on the mouse)

PR #213, opened 2016-11-21
Layer management

Jeremy-Gaillard:
This pull request aims at allowing multiple layers to be added to a scene without having to modify the scene.js file. This will be especially useful for mixing data from various data sources, since we will now be able to have multiple, independent quadtrees (or any other data structure) in the same scene.

Layers are paired with an updater. The updater can freely choose to implement a number of functions (update, updateMaterial, selectNode) depending on the layer's requirements.

Additionally, in this PR:
- BrowseTree is now a static class
- Selected node and fog configuration moved to Scene.js
- Removed or commented out dead code
- Moved RTC matrix calculation from c3DEngine to Camera


peppsac:
If `BrowseTree` only supports `QuadTree` structutre, maybe it could be a good idea to merge it in `UpdaterQuadTree`?

Jeremy-Gaillard:
BrowseTree should work with any tree structure, but your point still stands. I'll merge the two files later today.

Jeremy-Gaillard:
@peppsac Done.

gchoqueux:
I'll watch this week this PR

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -345,8 +346,9 @@ ApiGlobe.prototype.createSceneGlobe = function (coordCarto, viewerDiv) {
     this.scene = Scene(coordinate, ellipsoid, viewerDiv, debugMode, gLDebug);
 
     var map = new Globe(ellipsoid, gLDebug);
+    var updater = new GlobeUpdater(ellipsoid);
 
-    this.scene.add(map);
+    this.scene.setMap(map, updater);
```
gchoqueux, line 351: Why change `add` and use `setMap`, It should have no difference when you add a `node` or `globe` (`Globe` inherits `node`)

src/Renderer/Camera.js
```diff
@@ -174,4 +174,30 @@ Camera.prototype.getFrustumLocalSpace = function (position, quaternion) {
 };
 
 
+Camera.prototype.getRTCMatrixFromCenter = function (center) {
```
gchoqueux, line 177: These are `nodeProcess`, like occlusion, horizon occlusion...
The format of a process is `foo(node, camera, params)` to be able to be added generically in a `updater` or `browseTree`. The idea is to have a generic `updater` to which have added some `nodeProcess`, but not having several updater.
This is not the best place, because it is not the camera job to calculate a `RTC` matrix for a node. This should be a `nodeProcess` as a static function

src/Scene/DefaultUpdater.js
```diff
@@ -0,0 +1,32 @@
+/**
```
gchoqueux, line 1: The idea is to have a generic `updater` or 'browseTree' to which have added some static functions `nodeProcess`, but not having several updater. The 'nodeProcess' are applied to each node and the traverse stops when a process returns false or undefined. There is also a notion of 'nodeProcess' order.

**example**
```js
const processList = [
  nodeProcess.occlusion,
  nodeProcess.occlusionHorizon,
  nodeProcess.SSE,
  nodeProcess.refineNode,
];
var updaterQuadtree = new updater(processList);
var quadTree = new QuadTree();

// link updater to quadtree for update 
scene.apply(quadTree,updaterQuadtree);
```
We can even imagine a process tree to be applied to each node. The children process will be performed according to the parent process.

src/Scene/Scene.js
```diff
@@ -44,23 +37,22 @@ function Scene(coordinate, ellipsoid, viewerDiv, debugMode, gLDebug) {
...
+
     this.map = null;
 
     this.cameras = null;
-    this.selectNodes = null;
+    this.selectedNodeId = null;
     this.managerCommand = ManagerCommands(this);
     this.orbitOn = false;
 
-    this.stylesManager = new StyleManager();
```
gchoqueux, line 54: I need `style` for the wfs ( #195)

src/Scene/Scene.js
```diff
@@ -94,18 +86,6 @@ Scene.prototype.getPickPosition = function (mouse) {
     return this.gfxEngine.getPickingPositionFromDepth(mouse);
 };
 
-Scene.prototype.getStyle = function (name) {
```
gchoqueux, line 97: I need `Style` this for the wfs ( #195)

gchoqueux (changes requested):

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
Jeremy-Gaillard: In my opinion:
- Either you treat map differently for specific purpose (snap-to-ground for instance) and you need to differentiate the two functions
- Either globe is just another layer, and you remove the map attribute

src/Renderer/Camera.js
Jeremy-Gaillard: This function does not compute the RTC matrix for a node but for a position in space, relative to the current camera instance.

It could be more aptly named, like "computeRTCMatrixAtPoint".

I think Camera is a good place for this function, since other computing functions related to the camera (such as SSE) are in the same class.

src/Scene/Scene.js
Jeremy-Gaillard: My bad, I'll put it back.

src/Scene/DefaultUpdater.js
Jeremy-Gaillard: I see. I think I can implement the modular node process, but I have more doubts about the updater: all the layers are not handled the same way. There are different strategies to determine which nodes have to be processed: browse a tree, compute geometrically for a gird, etc. I think we need a different updater for each one of these strategies.

src/Renderer/Camera.js
gchoqueux: I think move `SSE` from `Camera`

gchoqueux: `computeRTCMatrixAtPoint` is better name

gchoqueux: These functions must be static, generic and group in nodeProcess

src/Scene/DefaultUpdater.js
gchoqueux: > I think we need a different updater for each one of these strategies.

`const process` allows to manage the different structures.
You can instance a new `Updater` with different process for each data structure. The difference is the `processList`, it must be adapt to the structure.

And I return to the static and generic `processNode` (like SSE and RTC) to be given to the `Updater`.

src/Renderer/Camera.js
gchoqueux: see [comment on Updater](https://github.com/iTowns/itowns2/pull/213#discussion_r91319381)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: > Either globe is just another layer, and you remove the map attribute

Why not

src/Scene/NodeProcess.js
```diff
@@ -15,11 +13,8 @@ import { l_ELEVATION, l_COLOR } from 'Renderer/LayeredMaterial';
 
 export const SSE_SUBDIVISION_THRESHOLD = 6.0;
 
-function NodeProcess(camera, ellipsoid, bbox) {
+function NodeProcess(ellipsoid) {
```
peppsac, line 16: IMHO: this class is `GlobeTileProcess`.
Maybe later we'll be able to extract some generic codes from this, but currently it only serves 1 purpose: process a Tile from a Globe.

src/Scene/MobileMappingUpdater.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Generated On: 2016-11-15
+ * Class: DefaultUpdater
+ * Description: Updaters for MobileMappingLayers.
+ */
+
+function MobileMappingUpdater() {
```
peppsac, line 7: Is MobileMapping supported yet? if not I'd rather not add untested class like this one.

src/Scene/GlobeUpdater.js
```diff
@@ -0,0 +1,71 @@
...
+import Quadtree from 'Scene/Quadtree';
+import TreeUpdater from 'Scene/TreeUpdater';
+import Layer from 'Scene/Layer';
+import DefaultUpdater from 'Scene/DefaultUpdater';
+import MobileMappingLayer from 'MobileMapping/MobileMappingLayer';
+import MobileMappingUpdater from 'Scene/MobileMappingUpdater';
+
+
+function GlobeUpdater(ellipsoid) {
+    this.treeUpdater = new TreeUpdater(new NodeProcess(ellipsoid));
```
peppsac, line 17: GlobeUpdate **is** a TreeUpdater. `defaultUpdater` and `mobileMappingUpdater` should go away

src/Scene/Scene.js
```diff
@@ -150,20 +130,15 @@ Scene.prototype.scheduleUpdate = function () {
...
-                this.browserScene.updateMobileMappingLayer(sLayer, this.currentCamera());
-            } else if (sLayer instanceof Layer) {
-                this.browserScene.updateLayer(sLayer, this.currentCamera());
-            }
-        }
+        var updater = this.layers[l].updater;
+        params.layer = this.layers[l].layer;
+        params.layersConfig = params.layer.layersConfiguration;
+        // Is implemented for Globe, Quadtree, Layer and MobileMappingLayer.
+        if (updater.update)
```
peppsac, line 140:
```
if (updater.update) {
    updater.update(params);
}
```

src/Scene/Scene.js
```diff
@@ -215,18 +190,14 @@ Scene.prototype.scene3D = function () {
...
+    this.layers.push({ layer, updater });
+    this.gfxEngine.add3DScene(layer.getMesh());
+};
 
-    this.layers.push({
-        node,
-        process: nodeProcess,
-    });
-    this.gfxEngine.add3DScene(node.getMesh());
+Scene.prototype.setMap = function (layer, updater) {
```
peppsac, line 198: Add a comment answering these questions: why/when should I use `setMap` ? why/when should I use `add` ?

Also is this `this.map` thing necessary at all?

src/Scene/TreeUpdater.js
```diff
@@ -0,0 +1,151 @@
...
+        applyFunctionToChildren(n => this._browseNonDisplayableNode(n, node.level + 2, process, camera, params), node);
+    }
+};
+
+TreeUpdater.prototype._browseNonDisplayableNode = function (node, level, process, camera, params) {
+    // update node's sse value
+    node.sse = camera.computeNodeSSE(node);
+    node.setDisplayed(false);
+
+    var sse = process.checkNodeSSE(node);
```
peppsac, line 117: This `TreeUpdater` could be modified to work with any `Layer`.
It just needs to recursively traverse the tree until `process()` function returns false.
And all the visibility, SSE stuff should move to helper classes or Globe processor.

peppsac (commented):
(comments from my review are based on this architecture comment: https://github.com/iTowns/itowns2/issues/217#issuecomment-266241017)

src/Scene/NodeProcess.js
Jeremy-Gaillard: The name is indeed too generic for its actual purpose, I'll change it.

src/Scene/Scene.js
Jeremy-Gaillard: I'll be removing this.map. See: https://github.com/iTowns/itowns2/pull/213#discussion_r91311375

peppsac:
Superseded by #239

PR #218, opened 2016-12-07
Animation and Damping

gchoqueux:
### New Class for animation features
* `AnimationPlayer` to control animation
* `Animation` default animation
* `AnimationExpression` to create animation with function
### Add Animation features in globe control
* double click event : it triggers a animation to zoom to mouse position
* damping moves
### Add Animation features in Api's functions
* `setCameraOrientation()`
* `setTilt()`
* `setHeading()`
* `resetTilt()`
* `resetHeading()`
* `setCenter()`
* `setCenterAdvanced()` (2 animations for moment)
* `setRange()`
* `setZoomLevel()`
* `setZoomScale()`
* `setAnimationEnabled()`
* `isAnimationEnabled()`

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,9 +8,9 @@ and open the template in the editor.
     <head>
         <title>Globe</title>
 
-	<style type="text/css">
+    <style type="text/css">
             html {height: 100%}
-	    body { margin: 0; overflow:hidden; height:100%}
+        body { margin: 0; overflow:hidden; height:100%}
 
             #viewerDiv {
                 margin : auto auto;
@@ -32,47 +32,51 @@ and open the template in the editor.
     <body>
         <div id="viewerDiv"></div>
         <script src="examples/GUI/GuiTools.js"></script>
-        <script src="dist/itowns2.js"> </script>
+        <script src="dist/itowns2.js"></script>
         <script type="text/javascript">
 
             /* global itowns,document,GuiTools*/
 
-            var positionOnGlobe = { longitude:2.3465, latitude: 48.88, altitude: 25000000};
+            const positionOnGlobe = { longitude: 2.3465, latitude: 48.88, altitude: 25000000 };
 
             // iTowns namespace defined here
-            var viewerDiv = document.getElementById('viewerDiv');
-            var menuGlobe = new GuiTools(itowns.viewer,'menuDiv');
+            const viewerDiv = document.getElementById('viewerDiv');
+            const menuGlobe = new GuiTools(itowns.viewer, 'menuDiv');
             itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
 
-            let wgs84TileLayer = {
+            const wgs84TileLayer = {
                 protocol: 'tile',
-                id: 'wgs84'
+                id: 'wgs84',
             };
 
-			var urlImageryLayers = [], urlElevationLayers = [];
+            const urlImageryLayers = [];
+            const urlElevationLayers = [];
 
             urlImageryLayers.push('examples/layers/JSONLayers/Ortho.json');
             urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
             urlImageryLayers.push('examples/layers/JSONLayers/ScanEX.json');
             urlImageryLayers.push('examples/layers/JSONLayers/Region.json');
 
-		  	urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT.json');
+            urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT.json');
             urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT_HIGHRES.json');
 
             itowns.viewer.addGeometryLayer(wgs84TileLayer);
-            let imageryPromise = itowns.viewer.addImageryLayersFromJSONArray(urlImageryLayers);
-            let elevationPromise = itowns.viewer.addElevationLayersFromJSONArray(urlElevationLayers);
+            const imageryPromise = itowns.viewer.addImageryLayersFromJSONArray(urlImageryLayers);
+            const elevationPromise = itowns.viewer.addElevationLayersFromJSONArray(urlElevationLayers);
 
-            imageryPromise.then((layersColors) => menuGlobe.addImageryLayersGUI(layersColors));
-            elevationPromise.then((layersElevations) => {menuGlobe.addElevationLayersGUI(layersElevations);});
+            imageryPromise.then((layersColors) => { menuGlobe.addImageryLayersGUI(layersColors); });
+            elevationPromise.then((layersElevations) => { menuGlobe.addElevationLayersGUI(layersElevations); });
 
             Promise.all([imageryPromise, elevationPromise]).then(() => itowns.viewer.init());
 
-            menuGlobe.addGUI('RealisticLighting',false,function(newValue) {itowns.viewer.setRealisticLightingOn(newValue);});
+            menuGlobe.addGUI('RealisticLighting', false,
+                (newValue) => { itowns.viewer.setRealisticLightingOn(newValue); });
 
-           	itowns.viewer.addEventListener('globe-loaded', function() {
-            	console.log("Globe Loaded");
-        	});
-        </script>
+            itowns.viewer.addEventListener('globe-loaded', () => {
+                // eslint-disable-next-line no-console
+                console.info('Globe Loaded');
+            });
+
+</script>
     </body>
 </html>
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "paralleljs": "^0.2.1",
     "simd": "^2.0.0",
     "string_format": "^0.0.6",
-    "three": "^0.82.0",
+    "three": "^0.83.0",
     "whatwg-fetch": "^1.0.0",
     "custom-event": "^1.0.0"
   },
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -17,11 +17,12 @@ import CustomEvent from 'custom-event';
 import Fetcher from 'Core/Commander/Providers/Fetcher';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from 'Scene/LayerUpdateStrategy';
 
-var loaded = false;
+var sceneIsLoaded = false;
 var eventLoaded = new CustomEvent('globe-loaded');
 var eventRange = new CustomEvent('rangeChanged');
 var eventCenter = new CustomEvent('centerchanged');
 var eventOrientation = new CustomEvent('orientationchanged');
+var eventPan = new CustomEvent('panchanged');
 var eventLayerAdded = new CustomEvent('layeradded');
 var eventLayerRemoved = new CustomEvent('layerremoved');
 var eventLayerChanged = new CustomEvent('layerchanged');
@@ -30,12 +31,24 @@ var eventLayerChangedOpacity = new CustomEvent('layerchanged:opacity');
 var eventLayerChangedIndex = new CustomEvent('layerchanged:index');
 var eventError = new CustomEvent('error');
 
+var enableAnimation = false;
+
+const defer = function defer() {
+    const deferedPromise = {};
+    deferedPromise.promise = new Promise((resolve, reject) => {
+        deferedPromise.resolve = resolve;
+        deferedPromise.reject = reject;
+    });
+    return deferedPromise;
+};
+
 function ApiGlobe() {
     // Constructor
     this.scene = null;
     this.commandsTree = null;
     this.projection = new Projection();
     this.viewerDiv = null;
+    this.callback = null;
 }
 
 ApiGlobe.prototype.constructor = ApiGlobe;
@@ -320,8 +333,10 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
     this.viewerDiv = viewerDiv;
 
     viewerDiv.addEventListener('globe-built', () => {
-        if (loaded == false) {
-            loaded = true;
+        if (sceneIsLoaded === false) {
+            sceneIsLoaded = true;
+            this.scene.currentControls().updateCameraTransformation();
+            this.scene.updateScene3D();
             viewerDiv.dispatchEvent(eventLoaded);
         } else {
             viewerDiv.dispatchEvent(eventError);
@@ -330,9 +345,6 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
 
     var gLDebug = false; // true to support GLInspector addon
     var debugMode = false;
-
-    // gLDebug = true; // true to support GLInspector addon
-    // debugMode = true;
 
     var ellipsoid = new Ellipsoid({
         x: 6378137,
@@ -358,6 +370,11 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
     this.scene.managerCommand.addProtocolProvider('tile', new TileProvider(ellipsoid));
     this.scene.managerCommand.addProtocolProvider('wms', new WMS_Provider({ support: map.gLDebug }));
 
+    this.sceneLoadedDeferred = defer();
+    this.addEventListener('globe-loaded', () => {
+        this.sceneLoadedDeferred.resolve();
+        this.sceneLoadedDeferred = defer();
+    });
 
     return this.scene;
 };
@@ -365,16 +382,6 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
 ApiGlobe.prototype.update = function update() {
     this.scene.notifyChange();
 };
-
-// ApiGlobe.prototype.setLayerAtLevel = functionsetLayerAtLevel(baseurl,layer/*,level*/) {
-//     // TODO CLEAN AND GENERIC
-//     var wmtsProvider = new WMTS_Provider({url:baseurl, layer:layer});
-//     this.scene.managerCommand.providerMap[4] = wmtsProvider;
-//     this.scene.managerCommand.providerMap[5] = wmtsProvider;
-//     this.scene.managerCommand.providerMap[this.scene.layers[0].node.meshTerrain.id].providerWMTS = wmtsProvider;
-//     this.scene.browserScene.updateNodeMaterial(wmtsProvider);
-//     this.scene.renderScene3D();
-// };
 
 ApiGlobe.prototype.showClouds = function showClouds(value, satelliteAnimation) {
     this.scene.getMap().showClouds(value, satelliteAnimation);
@@ -385,7 +392,7 @@ ApiGlobe.prototype.setRealisticLightingOn = function setRealisticLightingOn(valu
     this.scene.setLightingPos();
     this.scene.gfxEngine.setLightingOn(value);
     this.scene.getMap().setRealisticLightingOn(value);
-    this.scene.browserScene.updateMaterialUniform('lightingOn', value ? 1 : 0);
+    this.scene.browserScene.updateMaterialUniform('lightingOn', value);
     this.scene.renderScene3D();
 };
 
@@ -470,13 +477,14 @@ ApiGlobe.prototype.getCenter = function getCenter() {
  * Sets orientation angles of the current camera, in degrees.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/9qr2mogh/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
  * @constructor
- * @param {Orientation} Param - The angle of the rotation in degrees.
+ * @param      {object}   orientation  The angle of the rotation in degrees
+ * @param      {boolean}  isAnimated   Indicates if animated
+ * @return     {Promise}   { description_of_the_return_value }
  */
-
-ApiGlobe.prototype.setCameraOrientation = function setCameraOrientation(orientation /* param,pDisableAnimationopt*/) {
-    this.setHeading(orientation.heading);
-    this.setTilt(orientation.tilt);
-    this.viewerDiv.dispatchEvent(eventOrientation);
+ApiGlobe.prototype.setCameraOrientation = function setCameraOrientation(orientation, isAnimated) {
+    return this.scene.currentControls().setOrbitalPosition(undefined, orientation.heading, orientation.tilt, isAnimated).then(() => {
+        this.viewerDiv.dispatchEvent(eventOrientation);
+    });
 };
 
 /**
@@ -551,17 +559,38 @@ ApiGlobe.prototype.getRangeFromEllipsoid = function getRangeFromEllipsoid() {
 };
 
 /**
+ * Sets the animation enabled.
+ * @constructor
+ * @param      {boolean}  enable  The enable
+ */
+ApiGlobe.prototype.setAnimationEnabled = function setAnimationEnabled(enable) {
+    enableAnimation = enable;
+};
+
+/**
+ * Determines if animation enabled.
+ *
+ * @return     {boolean}  True if animation enabled, False otherwise.
+ */
+ApiGlobe.prototype.isAnimationEnabled = function isAnimationEnabled() {
+    return enableAnimation;
+};
+
+/**
  * Change the tilt.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/p6t76zox/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
  * @constructor
  * @param {Angle} Number - The angle.
- * @param {Boolean} [pDisableAnimation] - Used to force the non use of animation if its enable.
+ * @param      {boolean}  isAnimated  Indicates if animated
+ * @return     {Promise}
  */
-
-ApiGlobe.prototype.setTilt = function setTilt(tilt /* , bool*/) {
+ApiGlobe.prototype.setTilt = function setTilt(tilt, isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
     eventOrientation.oldTilt = this.getTilt();
-    this.scene.currentControls().setTilt(tilt);
-    this.viewerDiv.dispatchEvent(eventOrientation);
+    return this.scene.currentControls().setTilt(tilt, isAnimated).then(() => {
+        this.viewerDiv.dispatchEvent(eventOrientation);
+        this.scene.notifyChange(1);
+    });
 };
 
 /**
@@ -569,35 +598,38 @@ ApiGlobe.prototype.setTilt = function setTilt(tilt /* , bool*/) {
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/rxe4xgxj/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
  * @constructor
  * @param {Angle} Number - The angle.
- * @param {Boolean} [pDisableAnimation] - Used to force the non use of animation if its enable.
+ * @param      {boolean}  isAnimated  Indicates if animated
+ * @return     {Promise}
  */
-
-ApiGlobe.prototype.setHeading = function setHeading(heading /* , bool*/) {
+ApiGlobe.prototype.setHeading = function setHeading(heading, isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
     eventOrientation.oldHeading = this.getHeading();
-    this.scene.currentControls().setHeading(heading);
-    this.viewerDiv.dispatchEvent(eventOrientation);
+    return this.scene.currentControls().setHeading(heading, isAnimated).then(() => {
+        this.viewerDiv.dispatchEvent(eventOrientation);
+        this.scene.notifyChange(1);
+    });
 };
 
 /**
  * Resets camera tilt -> sets the tilt to 0°.
  * @constructor
- * @param {Boolean} [pDisableAnimation] - Used to force the non use of animation if its enable.
+ * @param      {boolean}  isAnimated  Indicates if animated
+ * @return     {Promise}
  */
-
-ApiGlobe.prototype.resetTilt = function resetTilt(/* bool*/) {
-    this.scene.currentControls().setTilt(0);
-    this.viewerDiv.dispatchEvent(eventOrientation);
+ApiGlobe.prototype.resetTilt = function resetTilt(isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
+    return this.scene.currentControls().setTilt(0, isAnimated);
 };
 
 /**
  * Resets camera heading -> sets the heading to 0°.
  * @constructor
- * @param {Boolean} [pDisableAnimation] - Used to force the non use of animation if its enable.
+ * @param      {boolean}  isAnimated  Indicates if animated
+ * @return     {Promise}
  */
-
-ApiGlobe.prototype.resetHeading = function resetHeading(/* bool*/) {
-    this.scene.currentControls().setHeading(0);
-    this.viewerDiv.dispatchEvent(eventOrientation);
+ApiGlobe.prototype.resetHeading = function resetHeading(isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
+    return this.scene.currentControls().setHeading(0, isAnimated);
 };
 
 /**
@@ -613,17 +645,31 @@ ApiGlobe.prototype.computeDistance = function computeDistance(p1, p2) {
     return this.scene.getEllipsoid().computeDistance(new GeoCoordinate().copy(p1), new GeoCoordinate().copy(p2));
 };
 
+ApiGlobe.prototype.setSceneLoaded = function setSceneLoaded() {
+    sceneIsLoaded = false;
+    return this.sceneLoadedDeferred.promise;
+};
+
 /**
  * Changes the center of the scene on screen to the specified coordinates.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/x06yhbq6/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
  * @constructor
  * @param {coordinates} coordinates - Properties : longitude and latitude
+ * @param      {boolean}  isAnimated  Indicates if animated
+ * @return     {Promise}
  */
-ApiGlobe.prototype.setCenter = function setCenter(coordinates) {
+ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
     eventCenter.oldCenter = this.getCenter();
-    var position3D = this.scene.getEllipsoid().cartographicToCartesian(new GeoCoordinate(coordinates.longitude, coordinates.latitude, 0, UNIT.DEGREE));
-    this.scene.currentControls().setCenter(position3D);
-    this.viewerDiv.dispatchEvent(eventCenter);
+    const position3D = this.scene.getEllipsoid().cartographicToCartesian(new GeoCoordinate(coordinates.longitude, coordinates.latitude, 0, UNIT.DEGREE));
+    position3D.range = coordinates.range;
+    return this.scene.currentControls().setCenter(position3D, isAnimated).then(() => {
+        this.scene.notifyChange(1);
+        return this.setSceneLoaded().then(() => {
+            this.scene.currentControls().updateCameraTransformation();
+            this.viewerDiv.dispatchEvent(eventCenter);
+        });
+    });
 };
 
 /**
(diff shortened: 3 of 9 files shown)


index.html
```diff
@@ -32,47 +32,52 @@
...
 
             itowns.viewer.addGeometryLayer(wgs84TileLayer);
-            let imageryPromise = itowns.viewer.addImageryLayersFromJSONArray(urlImageryLayers);
-            let elevationPromise = itowns.viewer.addElevationLayersFromJSONArray(urlElevationLayers);
+            const imageryPromise = itowns.viewer.addImageryLayersFromJSONArray(urlImageryLayers);
+            const elevationPromise = itowns.viewer.addElevationLayersFromJSONArray(urlElevationLayers);
 
-            imageryPromise.then((layersColors) => menuGlobe.addImageryLayersGUI(layersColors));
-            elevationPromise.then((layersElevations) => {menuGlobe.addElevationLayersGUI(layersElevations);});
+            imageryPromise.then((layersColors) => { menuGlobe.addImageryLayersGUI(layersColors); });
```
autra, line 67: Just chiming in because I did the linter things : is this required by the linter? (It shouldn't !)

index.html
```diff
@@ -32,47 +32,52 @@
...
-            menuGlobe.addGUI('RealisticLighting',false,function(newValue) {itowns.viewer.setRealisticLightingOn(newValue);});
+            menuGlobe.addGUI('RealisticLighting', false,
+                (newValue) => { itowns.viewer.setRealisticLightingOn(newValue); });
 
-           	itowns.viewer.addEventListener('globe-loaded', function() {
-            	console.log("Globe Loaded");
-        	});
-        </script>
+            itowns.viewer.addEventListener('globe-loaded', () => {
+                /* eslint-disable no-alert, no-console */
```
autra: Same question: do you still have warnings/errors about `console.log`? airbnb style should allow them if I did things correctly.

And now I'm done and you guys can talk about important stuff ;-)

index.html
gchoqueux: yes

index.html
gchoqueux: Yes, there are errors with `console.log`, this is one of the first rules we added. I prefer like this

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -44,51 +46,152 @@ var CONTROL_KEYS = {
     S: 83,
 };
 
+// TODO: can be optimize for some uses
+var presiceSlerp = function (qb, t) {
```
peppsac: Why not use Quaternion.slerp from three.js?

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -1032,14 +1194,54 @@ GlobeControls.prototype.constructor = GlobeControls;
...
+
+GlobeControls.prototype.setRange = function (pRange, isAnimated) {
+    const deltaRange = pRange - this.getRange();
+    return this.moveOrbitalPosition(deltaRange, 0, 0, isAnimated);
+};
+
+GlobeControls.prototype.setOrbitalPosition = function (range, heading, tilt, isAnimated) {
+    const deltaPhi = ((tilt || this.getTilt()) * Math.PI / 180 - this.getTiltRad());
+    const deltaTheta = ((heading || this.getHeading()) * Math.PI / 180 - this.getHeadingRad());
+    const deltaRange = (range || this.getRange()) - this.getRange();
```
peppsac: `range || 0` ? (the same applies to the 2 previous lines)

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+        'Play',
+        'End',
+        'Pause',
+    ];
+
+/**
+ * AnimationPlayer
+ * It can play, pause or stop animation
+ */
+function AnimationPlayer(dom) {
```
autra: This is a very good candidate to use es6 classes :-)

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+    this.dom = dom;
+    this.id = null;
+    this.keyframe = 0;
+    this.frameEvent = new CustomEvent('frameAnimation');
+    this.stopEvent = new CustomEvent('stopAnimation');
+    this.endEvent = new CustomEvent('endAnimation');
+    this.animation = null;
+    this.resolve = null;
+    this.promise = new Promise((r) => { this.resolve = r; });
```
autra: btw, you can also write

```js
this.promise = new Promise(r => this.resolve = r);
```

Here, the `=` makes it not that nice, but when it's just a method call, it can be clearer. Just so you know :-)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -634,33 +630,37 @@ ApiGlobe.prototype.setCenter = function (coordinates) {
...
  */
-
-ApiGlobe.prototype.setCenterAdvanced = function (pPosition /* , pDisableAnimationopt*/) {
-    this.setCenter(pPosition.position);
-    this.setRange(pPosition.range);
-    this.setHeading(pPosition.heading);
-    this.setTilt(pPosition.tilt);
+ApiGlobe.prototype.setCenterAdvanced = function (pPosition, isAnimated) {
+    return this.setCenter(pPosition, isAnimated).then(() => {
+        var p = this.scene.currentControls().setOrbitalPosition(undefined, pPosition.heading, pPosition.tilt, isAnimated);
```
autra: `const p`? or just return directly...

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -44,51 +46,152 @@ var CONTROL_KEYS = {
...
+    root.scale = 1.0 - (1.0 - root.sphericalTo.radius / root.spherical.radius) * progress;
+    root.sphericalDelta.theta = root.sphericalTo.theta;
+    root.sphericalDelta.phi = root.sphericalTo.phi;
+};
+
+// Animations
+var animationDampingMove = new AnimatedExpression(120, ctrl, dampingMoveAnimatedExpression, 'Damping Move');
+var animationZoomCenter = new AnimatedExpression(45, ctrl, zoomCenterAnimatedExpression, 'Zoom Center');
+var animationOrbit = new AnimatedExpression(30, orbit, animationOrbitExpression, 'set Orbit');
+var OrbitDamping = new Animation(60, {}, 'orbit damping');
```
autra: `orbitDamping` ?

src/Scene/AnimationPlayer.js
autra: btw AnimationPlayer is a singleton right? In this case, maybe we can just have a `getInstance` method that would construct one or return the existing one. 

Then, we can put all the private state in a closure. This way you can avoid all the `bind(this)` below by creating these functions once and for all in the closure. Then you can return a plain object with all the public methods as fields (no need to populate the prototype in this case).

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * { Animation }
+ *
+ * @class      Animation (name)
+ * @param      {Number}  duration  The animation's duration
+ * @param      {Object}  root      The root object to animate
+ * @param      {String}  name      The animation's name
+ */
+function Animation(duration, root, name) {
```
autra: Why do you need this class? Why not using only `AnimatedExpression`? (and renaming to `Animation`)

Disclaimer, I'm not fan of these inheritance techniques in javascript. They don't actually enforce anything and the language is not made for this. They really look like they are imported from Java or C++ too much to feel ok ;-)

In js, we usually use just plain object constructed with `{ .. }` and document what we expect for each functions.

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+    this.duration = duration || 0;
+    this.root = root;
+    this.name = name;
+}
+
+Animation.prototype = Object.create(Animation.prototype);
+
+Animation.prototype.animate = function () {
```
autra: Following my previous comment: especially because this construction requires making empty function just *anywhere*.

src/Renderer/ThreeExtented/GlobeControls.js
autra: and use const?

src/Scene/AnimationPlayer.js
autra: > They don't actually enforce anything

to elaborate on that: 
- they don't enforce signature (`Object.create` does not "freeze" the object, and does not force a method to be defined. Methods can be added, removed, redefined)
- you cannot (easily) enforce type check when calling a method (so your method don't actually care to get a simple object or an object defined with new and a prototype) 

So for these reasons, you gain nothing more than if you just passed an object looking like

```js
{ 30, elem, 'name', animationFn}
```

The code would behave exactly the same, with exactly the same guarantees, but would be a lot more readable and easier to work with.

Here, Animation and AnimatedExpression are actually **data** and not **services**. They don't behave on their own, they don't need a prototype, they are just carrying some info + a callback. So I would
-  remove `Animation` and `AnimatedExpression` declarations entirely.
- replace `this.animation.animate(this.keyframe)` by `this.animation.expression(this.animation.root, keyFrame / this.animation.duration)` (it's a one-liner anyway) [on line 153](https://github.com/iTowns/itowns2/pull/218/files#diff-3629b694323d3d642bb4c6ecbebbaadbR153)
- all the AnimatedExpression declaration on GlobeControl would just be some plain objects that would be stored on `this.animation` of `AnimationPlayer` 
- Document what animation should look like in `AnimationPlayer.play`

I know this is a regular pattern in Itowns, but it's not idiomatic in js (at least a *lot* less than in Java) and we should abandon it. It is useful in other languages, but in js it's just a pain. 

Of course if I missed a reason why it's done this way, please do tell me :-)

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -958,13 +1108,22 @@ function GlobeControls(camera, domElement, engine) {
         keyS = false;
     };
 
+    // Callback launched when player is stopped
+    this.stopAnimation = function () {
```
autra: If it is to be called each time an animation is finished, please consider having this callback as a parameter when constructing the `AnimationPlayer`. This way you won't need to add
```js
player.play().then(() => this.stopAnimation())
```
Or something similar each time. Then this callback would be called each time you call `this.resolve` in AnimationPlayer.

Btw, promises are not made for recurring events. They are made for one-time resolvable event, for which they are great. But as soon as your need can be expressed as

> I want to do A every time B happens

They are poor candidates. Callbacks are still the right solution for this.

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * { Stop current animation }
+ *
+ * @return  {Promise}  Promise is resolved when animation is stopped or finished
+ */
+AnimationPlayer.prototype.stop = function () {
+    setPlayerState.bind(this)(PLAYER_STATE.STOP);
+    finishAnimation.bind(this)();
+    this.dom.dispatchEvent(this.stopEvent);
+    return this.promise;
```
autra: So I have a problem with `this.promise`:

- you init it in the constructor and capture the resolve function in `this.resolve`
- you return the promise only in `stop`
- you change `this.resolve` in `play` but not `this.promise`
- you resolve it in `finishAnimation`

As I expect `play` to be called before `stop` or `finishAnimation`, I don't see how `this.promise` could ever be resolved...

autra (commented):
Just passing by for my own culture, leaving some remarks. It's in no way exhaustive :-)

src/Scene/AnimationPlayer.js
gchoqueux: `AnimationPlayer` isn't a singleton. 

> Then, we can put all the private state in a closure. This way you can avoid all the bind(this) below by creating these functions once and for all in the closure. Then you can return a plain object with all the public methods as fields (no need to populate the prototype in this case).

a little example?

src/Scene/AnimationPlayer.js
gchoqueux: I rectify

src/Scene/AnimationPlayer.js
gchoqueux: > Why do you need this class?

There will be `keyFrameAnimation` inherit from `Animation` 

> `{ 30, elem, 'name', animationFn}`

Ok

> remove `Animation` and `AnimatedExpression` declarations entirely and replace `this.animation.animate(this.keyframe)` by `this.animation.expression(this.animation.root, keyFrame / this.animation.duration)`

I would prefer not to

> all the AnimatedExpression declaration on GlobeControl would just be some plain objects that would be stored on this.animation of AnimationPlayer

I don't understand

> Document what animation should look like in `AnimationPlayer.play`

Ok

src/Renderer/ThreeExtented/GlobeControls.js
gchoqueux: > `orbitDamping`

I change to `dampingOrbitalMvt` : damping orbital movement

src/Renderer/ThreeExtented/GlobeControls.js
gchoqueux: ok

gchoqueux:
@autra I push new commit
When using, does the animation work well?

index.html
tbroyer: You can use
```js
console.info('Globe Loaded');
```
instead, or
```js
console.info('Globe Loaded'); // eslint-disable-line no-console
```
(I don't think you need to disable `no-alert`)

src/Scene/AnimationPlayer.js
autra: > AnimationPlayer isn't a singleton.

Aren't you instanciating it only from GlobeControls's constructor?

> a little example?

Sure! if you never instanciate more than one AnimationPlayer, it could look like this

```js
// all the const here
const FRAMERATE = 60; 
// ....

// the state variables here. The one that are currently attached to `this` 
// in the constructor except this.dom
let id = null;
// ... 

// then goes all the functions, 
function resetTimer() {
    if (id) {
       clearInterval(id);
        id = undefined;
    }
    keyframe = 0;
}

function isStopped() {
    setPlayerState(PLAYER_STATE.STOP);
    finishAnimation();
    this.dom.dispatchEvent(stopEvent); // you can still use this.dom
    return promise; // I'm not sure you want a promise here
}
// ...

// then you export what you need to be visible
export default  {
    dom: null, // or document, as a default value 
    resetTimer,
    isStopped,
    //.... 
}
```
The client code just need to do `AnimationPlayer.dom = domElement;`. Here is [a working example of something similar](https://jsfiddle.net/8g68esnp/2/) (the `var myLib = (function() {...})();` emulates export and import mechanism of es6.

Imho this has several general advantages:
- what is private is *really* private: you can't access from the outside because webpack wraps all this in closures (actually you don't need to make a closure explicitly here). Btw closure is the real and only way to make things private in js. 
- it's less verbose

And additionally here, you get rid of `this`, so you don't need to `.bind` your functions (and it's faster too, because IIRC `bind` creates a new function each time it is called). The less you worry about this, the better :-)

If you need several instance of `AnimationPlayer` (but your code does not, in its current form), it's still possible, but you would need to wrap all the instance dependant fields in a closure, something like:

```js
// all the const
const ....

getInstance(dom) {
    // instanciate all instance dependant fields
    let id = ...;
    // ...
   
    // declare all the functions that depends on these field
    // they capture local variables in their closure
    function isStopped() { /* .... */ }
    return {
         isStopped,
         play,
         // ....
    };
}
export default {
    getInstance,
    // and maybe some "static" functions
}
```
I hope I'm clearer :-)

src/Scene/AnimationPlayer.js
autra: >  >    all the AnimatedExpression declaration on GlobeControl would just be some plain objects that would be stored on this.animation of AnimationPlayer
>
> I don't understand

Just saying that in my opinion you don't need classes and inheritance to pass a 4-flelds object to you `play` method (AnimatedExpression.animate does not really add values). 

Btw inheritance does not exist in javascript per-se. Yes, you can inherit prototype, but it does not make sense to do so with empty function. 

doing
```js
Animation.prototype.animate = function () {};
AnimatedExpression.prototype = Object.create(Animation.prototype);
AnimatedExpression.prototype.animate = function (keyFrame) {    /* actually do something */ };
```

is the same as doing
```js
AnimatedExpression.prototype.animate = function() {};
AnimatedExpression.prototype.animate = function(keyFrame) { /* do the real thing */ };
```
It's not very useful :-) You don't need `Animation`.

autra:
>  When using, does the animation work well?

Yes, except that I still have the bug when everything goes upside-down from time-to-time (but I had that before your patch)

Damping is the continuation of the movement when you release the mouse right? 

If so, it works only one way with the ctrl key (when increasing the angle diff to the normal), but not the other way (when going back near the normal), is it what you wanted? I can't trigger any damping with any other actions (but maybe that's intended). It's a bit laggy at the beginning of the animation though.. I'm testing on firefox

autra:
> If so, it works only one way with the ctrl key (when increasing the angle diff to the normal), but not the other way (when going back near the normal), is it what you wanted? I can't trigger any damping with any other actions (but maybe that's intended). It's a bit laggy at the beginning of the animation though.. I'm testing on firefox

Correction, things seems to be working as expected. I don't why I couldn't trigger the damping in my first tests.

src/Scene/AnimationPlayer.js
gchoqueux: OK thanks for your explanations

> Aren't you instanciating it only from GlobeControls's constructor?

Yes for the moment, but there will be other uses (like earth's rotation, object3D animation...)

src/Scene/AnimationPlayer.js
gchoqueux: but i use Animation [here](https://github.com/iTowns/itowns2/blob/05fa85953e12d4f9691e48bcde5e07d850f659c8/src/Renderer/ThreeExtented/GlobeControls.js#L182)

gchoqueux:
> Damping is the continuation of the movement when you release the mouse right?

Yes

src/Scene/AnimationPlayer.js
gchoqueux: I have pushed commit : ES6 Class for Animation

src/Scene/AnimationPlayer.js
gchoqueux: I have removed `Animation.prototype.animate`

gchoqueux:
@autra I have fix all your comments
and add ES6 Class for Animation

gchoqueux:
Rebased and Ready

gchoqueux:
Updated

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+const debugMsg =
+    [
+        'Stop',
+        'Play',
+        'End',
+        'Pause',
+    ];
+
+// if debugAnimation is true console.log are enabled to sniff animation'state
+var _DEBUG = null;
```
autra: nits: let

autra: And there are other occurrences of var below. It's not critical though, they behave nearly the same. But let has a Temporal Dead Zone, which is useful to fail fast.

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
+const FRAMERATE = 60;
+const INTERVAL_TIME = 1000 / FRAMERATE;
+// if is true console.log are enabled to sniff animation'state
+const debugAnimation = false;
```
autra, line 4: Maybe we can have a global that would put us into debug mode. It would be `false` by default and could be made true on dev mode. Then we could have a `debug.js` class that would contain our _DEBUG (or we could overwrite console.info). It would avoid to have a different debug var for each class.

I'll try to make a PR with this maybe this week or the next

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+        setPlayerState.bind(this)(PLAYER_STATE.PAUSE);
+        clearInterval(this.id);
+        this.id = undefined;
+    }
+
+    /**
+     * { this function is executed with each frame }
+     */
+    frame() {
+        if (this.keyframe === this.animation.duration) {
```
autra: Maybe 
```js
if (this.keyframe < this.animation.duration) {
    // code in current else
} else {
    // code in current if
}
```
It would allow to skip frames without risks.

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,203 @@
...
+ * It can play, pause or stop animation
+ */
+class AnimationPlayer {
+    constructor(dom) {
+        this.dom = dom;
+        this.id = null;
+        this.keyframe = 0;
+        this.frameEvent = new CustomEvent('frameAnimation');
+        this.stopEvent = new CustomEvent('stopAnimation');
+        this.endEvent = new CustomEvent('endAnimation');
```
autra: So when you fire an event like this, it gets a bunch of other attributes (`target`, `currentTarget`, `cancelable`...) from the browser and then is given to event listeners. For this reason, I'm not very comfortable with having static ones like that. I think it's safer to instanciate a new CustomEvent each time you dispatch it. It would also reduce the number of fields we have here (at the very least, these should be module constants).

package.json
```diff
@@ -29,7 +29,7 @@
     "paralleljs": "^0.2.1",
     "simd": "^2.0.0",
     "string_format": "^0.0.6",
-    "three": "^0.82.0",
+    "three": "^0.83.0",
```
autra, line 32: This could be merged immediately if it has its own PR :-)

index.html
```diff
@@ -32,47 +32,51 @@
...
 
-           	itowns.viewer.addEventListener('globe-loaded', function() {
-            	console.log("Globe Loaded");
-        	});
-        </script>
+            itowns.viewer.addEventListener('globe-loaded', () => {
+                // eslint-disable-next-line no-console
+                console.info('Globe Loaded');
+            });
+
```
autra, line 79: All this trivial change (space, formatting etc...) would be merged very quickly if they had their own PR. It would also make this one easier to review imho.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -613,17 +630,38 @@ ApiGlobe.prototype.computeDistance = function computeDistance(p1, p2) {
     return this.scene.getEllipsoid().computeDistance(new GeoCoordinate().copy(p1), new GeoCoordinate().copy(p2));
 };
 
+ApiGlobe.prototype.setSceneLoaded = function computeDistance(callback) {
```
autra: Why is the function name `computeDistance`?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -613,17 +630,38 @@ ApiGlobe.prototype.computeDistance = function computeDistance(p1, p2) {
...
+    loaded = false;
+    this.callback = () => {
```
autra: If you call 2 times `setSceneLoaded` before `globe-loaded` fires, only one of the 2 callbacks would get called. Actually, if you want to execute some piece of code once after the next `globe-loaded` event, promises are a good way to do so. More precisely, you can use what is commonly called **deferred**:

They are built this way (they are just wrappers around promise and their reject and resolve functions).
```js
// in some utils module maybe?
function defer() {
    let defer = {};
    defer.promise = new Promise((resolve, reject) => {
        defer.resolve = resolve;
        defer.reject = reject;
    });
    return defer;
}
```
Then instead of `this.callback` you will instanciate a defer, fire and replace it each time globe-loaded fires:
```js
this.globeLoadedDeferred = defer(); // precise names are better
this.addEventListener('globe-loaded', () => {
    this.globeLoadedDeferred.resolve();
    this.globeLoadedDeferred = defer();
});
```
And each time you currently call `setSceneLoaded`:
```js
this.globeLoadedDeferred.promise.then(<your callback>);
```
This way, each registered  callbacks since last `globe-loaded` would get called the next time the event fires. No need to add and remove listeners.

Also, by returning promises, you can chain them (see below).

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -613,17 +630,38 @@ ApiGlobe.prototype.computeDistance = function computeDistance(p1, p2) {
...
-ApiGlobe.prototype.setCenter = function setCenter(coordinates) {
+ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
     eventCenter.oldCenter = this.getCenter();
-    var position3D = this.scene.getEllipsoid().cartographicToCartesian(new GeoCoordinate(coordinates.longitude, coordinates.latitude, 0, UNIT.DEGREE));
-    this.scene.currentControls().setCenter(position3D);
-    this.viewerDiv.dispatchEvent(eventCenter);
+    const position3D = this.scene.getEllipsoid().cartographicToCartesian(new GeoCoordinate(coordinates.longitude, coordinates.latitude, 0, UNIT.DEGREE));
+    position3D.range = coordinates.range;
+    return new Promise((r) => {
```
autra: With the above change,  this will become
```js
    return this.scene.currentControls().setCenter(position3D, isAnimated).then(() => {
        this.scene.notifyChange(1);
        return this.globeLoadedDeferred.promise.then(() => {
            this.scene.currentControls().updateCameraTransformation();
            this.viewerDiv.dispatchEvent(eventCenter);
        });
    });
```
And calls to `setCenter().then` would also get executed after this then.

autra: Less boilerplate, isn't it?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -634,33 +672,39 @@ ApiGlobe.prototype.setCenter = function setCenter(coordinates) {
...
  */
-ApiGlobe.prototype.setRange = function setRange(pRange /* , bool anim*/) {
+ApiGlobe.prototype.setRange = function setRange(pRange, isAnimated) {
+    isAnimated = isAnimated || this.isAnimationEnabled();
     eventRange.oldRange = this.getRange();
-    loaded = false;
-    this.scene.currentControls().setRange(pRange);
-    this.addEventListener('globe-loaded', updateTargetCamera(this));
+
+    return new Promise((r) => {
```
autra: Same modifications as in `setCenter`

autra (commented):
I have not finished my review, but as I won't be available the next 3 days, I'm submitting this part right now. I'll get back to it asap.

src/Scene/AnimationPlayer.js
gchoqueux: Ok global debug with options (shader, animation, tile, command...)

index.html
gchoqueux: yes

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: oops

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: Nice

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: thanks

gchoqueux:
updated

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/**
+ * { Animation }
+ *
+ * @class      Animation (name)
+ * @param      {Number}  duration  The animation's duration
+ * @param      {Object}  root      The root object to animate
+ * @param      {String}  name      The animation's name
+ */
```
peppsac, line 180: 10 lines of periphrases do not make a good documentation.

If you add documentation, make it useful:
* explain the purpose of Animation: move objects? ik based animation? lerp between values? generic?
* specify units (duration is in seconds? days? µs ?)
* why|when should I use (not) `Animation`?

peppsac: btw, this comment applies to all the new comments added in this PR.

Documentation is hard but code review helps improving it.

peppsac: Last comment about comments: where does this strange layout (`{ Animation `} instead of `Animation`) comes from?

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,204 @@
...
+     * { Start animation }
+     *
+     * @param      {Animation} The animation to play
+     * @return     {Promise}  Promise is resolved when animation is stopped or finished
+     */
+    play(animation, callBack) {
+        this.callBack = callBack;
+        this.animation = animation;
+        if (this.animation.animate) this.callBackFrame = () => this.animation.animate(this.keyframe);
+        setPlayerState.bind(this)(PLAYER_STATE.PLAY);
```
peppsac: I'm not sure why there are so much `bind(this)` in this code ?

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,204 @@
...
+    this.callBackFrame = () => {};
+    if (this.callBack) {
+        this.callBack();
+        this.callBack = null;
+    }
+    this.resolve();
+};
+
+const setPlayerState = function setPlayerState(state) {
+    this.state = state;
```
peppsac:
```js
const setPlayerState = function setPlayerState(player, state) {
   player.state = state;
```
is less clumsy and doesn't require a `bind(this)`

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,204 @@
...
+    // Public functions
+
+    /**
+     * { Start animation }
+     *
+     * @param      {Animation} The animation to play
+     * @return     {Promise}  Promise is resolved when animation is stopped or finished
+     */
+    play(animation, callBack) {
+        this.callBack = callBack;
```
peppsac: Why do you mix `callbacks` and `promises`?
IMHO you should drop `this.callBack` unless you have a strong argument

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,204 @@
...
+ * It can play, pause or stop animation
+ */
+class AnimationPlayer {
+    constructor(dom) {
+        this.dom = dom;
+        this.id = null;
+        this.keyframe = 0;
+        this.animation = null;
+        this.resolve = null;
+        this.promise = new Promise(r => this.resolve = r);
```
peppsac: Not needed

peppsac (changes requested):

src/Renderer/ThreeExtented/GlobeControls.js
gchoqueux: Three.js `Quaternion.slerp` considers an angle less than 0.001 to zero ([see](https://github.com/mrdoob/three.js/blob/master/src/math/Quaternion.js#L443))

src/Scene/AnimationPlayer.js
gchoqueux: yes

src/Scene/AnimationPlayer.js
gchoqueux: @autra in this [comment](https://github.com/iTowns/itowns2/pull/218#discussion_r92003633) recommended to use callback

But Effectively the promise isn't used in play function

src/Scene/AnimationPlayer.js
gchoqueux: I remove the `bind(this)`

src/Scene/AnimationPlayer.js
gchoqueux: How would you initialize it?

src/Scene/AnimationPlayer.js
autra: What @peppsac means is that `this.callback` and `this.promise` are redundant to each other (they are both called/fired at the end of the animation). I suggested to use callback for each frame. I guess you need only `this.promise` and `this.callbackFrame` right?

gchoqueux: > I suggested to use callback for each frame

This has been the case since the beginning, [it was for callback in animation's ending](https://github.com/iTowns/itowns2/pull/218#discussion_r92003633). I see this

autra: I don't know why I said that earlier and I can't remember how the code was at this time. We might have misunderstood each other :-)

So anyway: one promise for the end of the animation and a callback for each frame. Sounds good to you guys?

gchoqueux: > one promise for the end of the animation and a callback for each frame. Sounds good to you guys?

Yes, I revert to this code

gchoqueux:
updated

autra:
Ok I think I see where the misunderstanding lies.

I think AnimationPlayer.promise is not at the right place. It should be in Animation (and won't ever be null, you init it when you create the animation, resolve it when it's finished). It's a state information about an animation, not the player. Same goes for callbackFrame, it's an information about Animation, not AnimationPlayer. And same goes for state. Anyway, I think everything would be so much simpler if your animations could play themselves (so actually if AnimationPlayer and Animation would be merged). 

```js
let animation = new Animation(..., callbackFrame);
animation.play().then(...).catch(...);
...
// later
animation.cancel(); // catch is executed 
// or
animation.pause();
animation.resume();
```

gchoqueux:
> so actually if AnimationPlayer and Animation would be merged

Later, I would like to evolve `AnimationPlayer` to play animations simultaneously.

```js
AnimationPlayer.play(animation_A, animation_B);
```
To synchronize the playing, it would take a manager like the AnimationPlayer

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * Gets the interval time.
+     *
+     * @return {Number}  The interval time is the time between consecutive frames
+     */
+    static get intervalTime() {
+        return INTERVAL_TIME;
+    }
+
+    isPlayed() {
```
peppsac: isPlaying

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
...
+        this.waitTimer = null;
+    }
+
+    /**
+     * Gets the interval time.
+     *
+     * @return {Number}  The interval time is the time between consecutive frames
+     */
+    static get intervalTime() {
+        return INTERVAL_TIME;
```
peppsac: * `frameDuration` would be a better name
* you don't know the framerate so assuming it'll be 60 FPS might be wrong...
* last but not least, this function is unused so you could remove it

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
+const FRAMERATE = 60;
+const INTERVAL_TIME = 1000 / FRAMERATE;
```
peppsac: FRAME_DURATION ?

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
...
+
+    // Public functions
+
+    /**
+     * { Start animation }
+     *
+     * @param      {Animation} The animation to play
+     * @return     {Promise}  Promise is resolved when animation is stopped or finished
+     */
+    play(animation) {
```
peppsac, line 116: AnimationPlayer only support 1 animation at a time, so you must check the value of `this.state`

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
+const FRAMERATE = 60;
+const INTERVAL_TIME = 1000 / FRAMERATE;
+// if is true console.log are enabled to sniff animation'state
+const debugAnimation = false;
+
+// player statut
+const PLAYER_STATE = {
+    // player is stopped
+    STOP: 0,
```
peppsac, line 9: What's the difference between `STOP` and `PAUSE`? (there's no `resume` function)

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -651,121 +765,155 @@ function GlobeControls(camera, domElement, engine) {
...
+
+        // Launch damping movement for :
+        //      * CONTROL_STATE.ORBIT
+        //      * CONTROL_STATE.MOVE_GLOBE
+        if (this.enableDamping) {
+            if (state === CONTROL_STATE.ORBIT && (sphericalDelta.theta > EPS || sphericalDelta.phi > EPS)) {
+                player.play(dampingOrbitalMvt).then(() => this.stopAnimation());
+            } else if (state === CONTROL_STATE.MOVE_GLOBE && lastRotation.length === 2 && !(lastRotation[1].equals(lastRotation[0]))) {
+                ctrl.qDelta.setFromUnitVectors(lastRotation[1], lastRotation[0]);
+                player.play(animationDampingMove).then(() => this.stopAnimation());
```
peppsac: Why do you stop the animation after it ended?

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * { The animation is played after a number of frames }
+     *
+     * @param      {Animation}  animation    The animation to play
+     * @param      {Number}  waitingTime  The waiting time before start animation (time in frame)
+     */
+    playLater(animation, waitingFrame) {
+        this.resolveWait = null;
```
peppsac, line 132: Check `this.state`

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * { this function is executed with each frame }
+     */
+    frame() {
+        if (this.keyframe < this.animation.duration) {
+            this.callBackFrame();
+            this.keyframe++;
+            this.dom.dispatchEvent(frameEvent);
```
peppsac, line 162: `event` and `callback` provide the same feature.

Why don't you drop one (callback)?

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,231 @@
...
+ * Animation is play by the AnimationPlayer during the time of duration
+ * During playback, the AnimationPlayer emits event for each frame
+ * Animation is used to execute a callback to each frame
+ * @class      Animation
+ * @param      {Number}  duration  The animation's duration in number of frames. FRAMERATE is number of frames in one seconde.
+ * @param      {String}  name      The animation's name. It's used for debug message.
+ */
+class Animation {
+    constructor(params) {
+        this.duration = params.duration || 0;
```
peppsac: Maybe pick a more sensible default value, 60 (1 sec) ? Or throw an error if `params.duration` is unset.

src/Scene/AnimationPlayer.js
peppsac: (this comment also applies to `stop` and `pause` methods)

peppsac (changes requested):
While the animation works, there are still a few things to fix before merging.

src/Scene/AnimationPlayer.js
gchoqueux: I add resume()

src/Scene/AnimationPlayer.js
gchoqueux: I remove it

src/Scene/AnimationPlayer.js
gchoqueux: ok

src/Renderer/ThreeExtented/GlobeControls.js
gchoqueux: it's bad name... 
I rename `reinitiateControls`, this function reinitiates somes controls's params.

src/Scene/AnimationPlayer.js
gchoqueux: `callBackFrame` is function to animate object
`frameEvent` is used to resfresh display.

src/Scene/AnimationPlayer.js
peppsac: I'd prefer to remove `PAUSE` state until there's a real need for it.

gchoqueux: ok

src/Scene/AnimationPlayer.js
gchoqueux: If another animation is playing, it's stopped and the new animation is played
(comment add in code)

src/Renderer/ThreeExtented/GlobeControls.js
```diff
@@ -958,13 +1104,22 @@ function GlobeControls(camera, domElement, engine) {
         keyS = false;
     };
 
+    // Callback launched when player is stopped
+    this.reinitiateControls = function reinitiateControls() {
```
peppsac: `resetControls` ?

src/Scene/AnimationPlayer.js
```diff
@@ -0,0 +1,214 @@
...
+        // needed to return promise to wait sync
+        return Promise.resolve();
+    }
+
+    /**
+     * { this function is executed with each frame }
+     */
+    frame() {
+        if (this.keyframe < this.animation.duration) {
+            this.callBackFrame();
```
peppsac: I'd suggest removing `this.callBackFrame` and simply do:
```js
if (animation.animate) {
    animation.animate(this.keyframe;
}
```

peppsac (approved):

PR #220, opened 2016-12-12
Fix func_names linter errors

autra:
Having some time, I took upon myself to correct one linter error this Monday. I hope I'll be able to do this once a week.

 This was more or less automatic, except for one important point: naming functions add names to the current scope. Therefore we need to be careful not to shadow any existing declaration by mistake. For local names, I guess it's not a problem because the linter would detect it (as the other declaration won't be used any more). However global names, for instance those attached to the `window` object can still be shadowed. I have been careful of those, but I might have missed something so please double-check! 

Pinging @iTowns/reviewers

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,7 +45,6 @@ module.exports = {
 
     // TODO reactivate all the following rules
 
-    'func-names': 'off',
     'no-mixed-operators': 'off',
     'no-use-before-define': 'off',
     'no-underscore-dangle': 'off',
--- a/src/Core/Commander/Command.js
+++ b/src/Core/Commander/Command.js
@@ -28,7 +28,7 @@ Command.prototype.constructor = Command;
 
 /**
  */
-Command.prototype.instance = function () {
+Command.prototype.instance = function instance() {
     // TODO: Implement Me
 
 };
--- a/src/Core/Commander/Command.js
+++ b/src/Core/Commander/Command.js
@@ -26,11 +26,4 @@ function Command() {
 
 Command.prototype.constructor = Command;
 
-/**
- */
-Command.prototype.instance = function instance() {
-    // TODO: Implement Me
-
-};
-
 export default Command;
--- a/src/Core/Commander/InterfaceCommander.js
+++ b/src/Core/Commander/InterfaceCommander.js
@@ -13,7 +13,7 @@ function InterfaceCommander(type, priorityFunction) {
     this.type = type;
 
     if (!Date.now) {
-        this.timestamp = function () {
+        this.timestamp = function timestamp() {
             return new Date().getTime();
         };
     } else {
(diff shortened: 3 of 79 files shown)


src/Core/Commander/Command.js
```diff
@@ -28,7 +28,7 @@ Command.prototype.constructor = Command;
 
 /**
  */
-Command.prototype.instance = function () {
+Command.prototype.instance = function instance() {
```
peppsac: Maybe you could remove this empty method instead?

src/Core/Commander/InterfaceCommander.js
```diff
@@ -13,7 +13,7 @@ function InterfaceCommander(type, priorityFunction) {
     this.type = type;
 
     if (!Date.now) {
-        this.timestamp = function () {
+        this.timestamp = function timestamp() {
```
peppsac: According to https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/now Date.now should be available in all navigators we target (including IE >= 9)...
So what about removing the `timestamp` function insteand?

peppsac (commented):
Too bad you didn't transform the code to use `class` first because the diff would have been way smaller :)

autra:
> Too bad you didn't transform the code to use class first because the diff would have been way smaller :)

Not sure if I have the courage for that :-D

autra:
@iTowns/reviewers let me ping you again :-) I know it's dull, but I guess we'd better merge this quickly to avoid having to rebase it and review it again, thanks!

peppsac (approved):
LGTM, will merge later today (so other reviewers can have a look if they want).

.eslintrc.js
```diff
@@ -45,7 +45,6 @@ module.exports = {
 
     // TODO reactivate all the following rules
 
-    'func-names': 'off',
```
gchoqueux, line 48: Can you put this file `.eslintrc.js ` in separate commit? 
To facilitate cherry-pick and rebase.

src/Core/Commander/InterfaceCommander.js
```diff
@@ -11,34 +11,26 @@ function InterfaceCommander(type, priorityFunction) {
...
  * @return  {[object Object]}
  */
-InterfaceCommander.prototype.buildCommand = function () {
+InterfaceCommander.prototype.buildCommand = function buildCommand() {
     // TODO: Implement Me
     this._builderCommand();
 };
 
-InterfaceCommander.prototype.request = function (parameters, requester, earlyDropFunction) {
+InterfaceCommander.prototype.request = function request(parameters, requester, earlyDropFunction) {
```
gchoqueux, line 26: What is the interest of declaring twice the function's name?

gchoqueux (changes requested):

.eslintrc.js
autra: Not really. Otherwise one of the commit would have a broken build. But I don't see how it would change anything for cherry-pick and rebase? You would have the same conflics anyway right?

src/Core/Commander/InterfaceCommander.js
autra: a good explanation from [the rule's page](http://eslint.org/docs/rules/func-names)

> If you leave off the function name then when the function throws an exception you are likely to get something similar to anonymous function in the stack trace. If you provide the optional name for a function expression then you will get the name of the function expression in the stack trace. 

Behind that is the fact that each function object has a `name` parameter, which is populated when you declare `function foo`, but not when you declare `let foo = function`. It's this attribute that is used in stacktraces and stuff. Note that ES6 arrow functions infere the name from the context (so it's not needed for them, most of the time).

.eslintrc.js
gchoqueux: Ok

src/Core/Commander/InterfaceCommander.js
gchoqueux: ok

gchoqueux (approved):

Issue #60, opened 2016-03-07
providers and copyright


massea:
What is the politic about copyrights ?
For the moment, we use IGN otho and elevation but we will also use global elevation and global imagery (and for my part, bathymetry and tidal data) and it will be necessary to cite the sources.

nosy-b:
Yes you're right we should mention the geographic sources at the bottom of the screen

nosy-b:
(Each active provider **domain** should be inserted into the Copyright List and displayed.)

gcebelieu:
More generaly, I think itowns should provide the ability to display mention of resources providers through a widget like openlayers (http://openlayers.org/en/v3.14.2/apidoc/ol.Attribution.html) or leaflet (http://leafletjs.com/reference.html#control-attribution) do.
The responsibility to display (or not) ressources references is left to itowns users.

vpicavet:
+1 @gcebelieu 
Let's transform this ticket into "create a widget for resources attributions"
Meanwhile we can list providers and attribution somewhere in the doc or code.

gcebelieu:
sounds good to me.
@vpicavet @nosy-b @gchoqueux @massea In the same time, another issue may also be created (I'll do it) : "Data resources (elevation, images, features and others ) should not be hardcoded in itowns2 code but only in examples".

gcebelieu:
These 2 issues will be handled in : #62

PR #221, opened 2016-12-14
IPR / Attribution

vcoindet:
IPR Attribution

In ApiGlobe.js

getLayersIPR() -> return an array of IPR with no duplicate
setLayerIPR(param)
removeDuplicateFomArray(originalArray, prop)

Note that the name of the IPR is mandatory

### DIFF
--- a/examples/layers/JSONLayers/DARK.json
+++ b/examples/layers/JSONLayers/DARK.json
@@ -3,11 +3,9 @@
     "id": "DARK",
     "customUrl": "http://a.basemaps.cartocdn.com/dark_all/%TILEMATRIX/%COL/%ROW.png",
     "options": {
-    	"ipr": {
+    	"attribution": {
     		"name":"CARTO",
-    		"url": "https://carto.com/",
-    		"logo":"",
-    		"description":""
+    		"url": "https://carto.com/"
     	},
         "tileMatrixSet": "PM",
         "mimetype": "image/png"
--- a/examples/layers/JSONLayers/DARK.json
+++ b/examples/layers/JSONLayers/DARK.json
@@ -3,6 +3,12 @@
     "id": "DARK",
     "customUrl": "http://a.basemaps.cartocdn.com/dark_all/%TILEMATRIX/%COL/%ROW.png",
     "options": {
+    	"ipr": {
+    		"name":"CARTO",
+    		"url": "https://carto.com/",
+    		"logo":"",
+    		"description":""
+    	},
         "tileMatrixSet": "PM",
         "mimetype": "image/png"
     }
--- a/index.html
+++ b/index.html
@@ -58,6 +58,8 @@ and open the template in the editor.
             urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
             urlImageryLayers.push('examples/layers/JSONLayers/ScanEX.json');
             urlImageryLayers.push('examples/layers/JSONLayers/Region.json');
+            /* urlImageryLayers.push('examples/layers/JSONLayers/DARK.json');
+            urlImageryLayers.push('examples/layers/JSONLayers/OPENSM.json'); */
 
             urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT.json');
             urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT_HIGHRES.json');
--- a/index.html
+++ b/index.html
@@ -58,8 +58,6 @@ and open the template in the editor.
             urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
             urlImageryLayers.push('examples/layers/JSONLayers/ScanEX.json');
             urlImageryLayers.push('examples/layers/JSONLayers/Region.json');
-            /* urlImageryLayers.push('examples/layers/JSONLayers/DARK.json');
-            urlImageryLayers.push('examples/layers/JSONLayers/OPENSM.json'); */
 
             urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT.json');
             urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT_HIGHRES.json');
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -917,35 +917,18 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
 
 /**
  * Get the Attribution of all layers in the scene.
- * @return {array}  An array of attribution.
+ * @return {map}  A Map of attribution.
  */
 
 ApiGlobe.prototype.getLayersAttribution = function getLayersAttribution() {
     const lc = this.scene.getMap().layersConfiguration;
     const map = new Map();
     [...lc.getColorLayers(), ...lc.getElevationLayers()].forEach((l) => {
-        if (l.options.attribution != undefined) {
+        if (l.options.attribution) {
             map.set(l.options.attribution.name, l.options.attribution);
         }
     });
     return map;
-};
-
-/**
- * Set the attribution for a layer.
- * @param {param} Param - The id of the layer and the new IPR {id, attribution}.
- */
-
-ApiGlobe.prototype.setLayerAttribution = function setLayerAttribution(param) {
-    if (param.attribution.name === undefined) {
-        throw new Error(`Missing parameter "param.attribution.name" : ${param.id}`);
-    }
-    const lc = this.scene.getMap().layersConfiguration;
-    [...lc.getColorLayers(), ...lc.getElevationLayers()].forEach((l) => {
-        if (param.id === l.id) {
-            l.options.attribution = param.attribution;
-        }
-    });
 };
 
 ApiGlobe.prototype.selectNodeById = function selectNodeById(id) {
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -915,6 +915,59 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
     this.viewerDiv.removeEventListener('layerchanged:index', this.callbackLayerChanged, false);
 };
 
+/**
+ * Get the IPR (Intellectual Property Rights) of all layers in the scene.
+ * @constructor
+ * @return {array}  An array of IPR.
+ */
+
+ApiGlobe.prototype.getLayersIPR = function getLayersIPR() {
+    var IPR = [];
+    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
+    for (var i = 0; i < colorLayers.length; i++) {
+        IPR.push(colorLayers[i].options.ipr);
+    }
+    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
+    for (var j = 0; j < elevationLayers.length; j++) {
+        IPR.push(elevationLayers[j].options.ipr);
+    }
+    IPR = this.removeDuplicatesFromArray(IPR, 'name');
+    return IPR;
+};
+
+ApiGlobe.prototype.removeDuplicatesFromArray = function removeDuplicatesFromArray(originalArray, prop) {
+    var newArray = [];
+    var lookupObject = {};
+    for (var i in originalArray) {
+        lookupObject[originalArray[i][prop]] = originalArray[i];
+    }
+    for (i in lookupObject) {
+        newArray.push(lookupObject[i]);
+    }
+    return newArray;
+};
+
+/**
+ * Set the IPR for a layer.
+ * @constructor
+ * @param {param} Param - The id of the layer and the new IPR {id, ipr}.
+ */
+
+ApiGlobe.prototype.setLayerIPR = function setLayerIPR(param) {
+    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
+    for (var i = 0; i < colorLayers.length; i++) {
+        if (param.id === colorLayers[i].id) {
+            colorLayers[i].options.ipr = param.ipr;
+        }
+    }
+    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
+    for (var j = 0; j < elevationLayers.length; j++) {
+        if (param.id === elevationLayers[j].id) {
+            elevationLayers[j].options.ipr = param.ipr;
+        }
+    }
+};
+
 ApiGlobe.prototype.selectNodeById = function selectNodeById(id) {
     this.scene.selectNodeId(id);
     this.scene.update();
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -916,56 +916,36 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
 };
 
 /**
- * Get the IPR (Intellectual Property Rights) of all layers in the scene.
- * @constructor
- * @return {array}  An array of IPR.
+ * Get the Attribution of all layers in the scene.
+ * @return {array}  An array of attribution.
  */
 
-ApiGlobe.prototype.getLayersIPR = function getLayersIPR() {
-    var IPR = [];
-    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
-    for (var i = 0; i < colorLayers.length; i++) {
-        IPR.push(colorLayers[i].options.ipr);
-    }
-    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
-    for (var j = 0; j < elevationLayers.length; j++) {
-        IPR.push(elevationLayers[j].options.ipr);
-    }
-    IPR = this.removeDuplicatesFromArray(IPR, 'name');
-    return IPR;
-};
-
-ApiGlobe.prototype.removeDuplicatesFromArray = function removeDuplicatesFromArray(originalArray, prop) {
-    var newArray = [];
-    var lookupObject = {};
-    for (var i in originalArray) {
-        lookupObject[originalArray[i][prop]] = originalArray[i];
-    }
-    for (i in lookupObject) {
-        newArray.push(lookupObject[i]);
-    }
-    return newArray;
+ApiGlobe.prototype.getLayersAttribution = function getLayersAttribution() {
+    const lc = this.scene.getMap().layersConfiguration;
+    const map = new Map();
+    [...lc.getColorLayers(), ...lc.getElevationLayers()].forEach((l) => {
+        if (l.options.attribution != undefined) {
+            map.set(l.options.attribution.name, l.options.attribution);
+        }
+    });
+    return map;
 };
 
 /**
- * Set the IPR for a layer.
- * @constructor
- * @param {param} Param - The id of the layer and the new IPR {id, ipr}.
+ * Set the attribution for a layer.
+ * @param {param} Param - The id of the layer and the new IPR {id, attribution}.
  */
 
-ApiGlobe.prototype.setLayerIPR = function setLayerIPR(param) {
-    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
-    for (var i = 0; i < colorLayers.length; i++) {
-        if (param.id === colorLayers[i].id) {
-            colorLayers[i].options.ipr = param.ipr;
-        }
+ApiGlobe.prototype.setLayerAttribution = function setLayerAttribution(param) {
+    if (param.attribution.name === undefined) {
+        throw new Error(`Missing parameter "param.attribution.name" : ${param.id}`);
     }
-    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
-    for (var j = 0; j < elevationLayers.length; j++) {
-        if (param.id === elevationLayers[j].id) {
-            elevationLayers[j].options.ipr = param.ipr;
+    const lc = this.scene.getMap().layersConfiguration;
+    [...lc.getColorLayers(), ...lc.getElevationLayers()].forEach((l) => {
+        if (param.id === l.id) {
+            l.options.attribution = param.attribution;
         }
-    }
+    });
 };
 
 ApiGlobe.prototype.selectNodeById = function selectNodeById(id) {
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -69,6 +69,8 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
 
     layer.bbox_url = layer.bbox_url || 'swne';
     layer.format = layer.options.mimetype || 'image/png';
+    layer.options.ipr = layer.options.ipr || '';
+    layer.crs = layer.projection || 'EPSG:4326';
     layer.width = layer.heightMapWidth || 256;
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -69,7 +69,6 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
 
     layer.bbox_url = layer.bbox_url || 'swne';
     layer.format = layer.options.mimetype || 'image/png';
-    layer.crs = layer.projection || 'EPSG:4326';
     layer.width = layer.heightMapWidth || 256;
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -69,7 +69,6 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
 
     layer.bbox_url = layer.bbox_url || 'swne';
     layer.format = layer.options.mimetype || 'image/png';
-    layer.options.ipr = layer.options.ipr || '';
     layer.crs = layer.projection || 'EPSG:4326';
     layer.width = layer.heightMapWidth || 256;
     layer.version = layer.version || '1.3.0';
--- a/src/Core/Commander/Providers/WMTS_Provider.js
+++ b/src/Core/Commander/Providers/WMTS_Provider.js
@@ -73,7 +73,6 @@ WMTS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer
         options.mimetype = options.mimetype || 'image/png';
         options.style = options.style || 'normal';
         options.projection = options.projection || 'EPSG:3857';
-        options.ipr = options.ipr || '';
         var newBaseUrl = `${layer.url
             }?LAYER=${options.name
             }&FORMAT=${options.mimetype
--- a/src/Core/Commander/Providers/WMTS_Provider.js
+++ b/src/Core/Commander/Providers/WMTS_Provider.js
@@ -73,6 +73,7 @@ WMTS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer
         options.mimetype = options.mimetype || 'image/png';
         options.style = options.style || 'normal';
         options.projection = options.projection || 'EPSG:3857';
+        options.ipr = options.ipr || '';
         var newBaseUrl = `${layer.url
             }?LAYER=${options.name
             }&FORMAT=${options.mimetype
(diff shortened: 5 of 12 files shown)


vcoindet:
rebased and ready

vcoindet:
Have to do some changes

vcoindet:
Ready

index.html
```diff
@@ -55,6 +55,8 @@
             urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
             urlImageryLayers.push('examples/layers/JSONLayers/ScanEX.json');
             urlImageryLayers.push('examples/layers/JSONLayers/Region.json');
+            /* urlImageryLayers.push('examples/layers/JSONLayers/DARK.json');
+            urlImageryLayers.push('examples/layers/JSONLayers/OPENSM.json'); */
```
autra: Maybe remove these comments?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
     this.viewerDiv.removeEventListener('layerchanged:index', this.callbackLayerChanged, false);
 };
 
+/**
+ * Get the IPR (Intellectual Property Rights) of all layers in the scene.
+ * @constructor
```
autra: Not a constructor.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+ * @return {array}  An array of IPR.
+ */
+
+ApiGlobe.prototype.getLayersIPR = function getLayersIPR() {
+    var IPR = [];
+    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
+    for (var i = 0; i < colorLayers.length; i++) {
+        IPR.push(colorLayers[i].options.ipr);
```
autra: You could use `Array.prototype.map` (or at least a `for...of` loop).

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    }
+    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
+    for (var j = 0; j < elevationLayers.length; j++) {
+        IPR.push(elevationLayers[j].options.ipr);
+    }
+    IPR = this.removeDuplicatesFromArray(IPR, 'name');
```
autra: Why not using a Set?

autra: It would avoid to implement another function just for that

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
 
+/**
+ * Get the IPR (Intellectual Property Rights) of all layers in the scene.
+ * @constructor
+ * @return {array}  An array of IPR.
+ */
+
+ApiGlobe.prototype.getLayersIPR = function getLayersIPR() {
+    var IPR = [];
+    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
```
autra: Use `const` for both vars.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    }
+    for (i in lookupObject) {
+        newArray.push(lookupObject[i]);
+    }
+    return newArray;
+};
+
+/**
+ * Set the IPR for a layer.
+ * @constructor
```
autra: not a constructor

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+ * @param {param} Param - The id of the layer and the new IPR {id, ipr}.
+ */
+
+ApiGlobe.prototype.setLayerIPR = function setLayerIPR(param) {
```
autra: Do we really need this method if it is specified in the json? These are mainly static info anyway right?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    if (param.ipr.name === undefined) {
+        throw new Error(`Missing parameter "param.ipr.name" : ${param.id}`);
+    }
+    var colorLayers = this.scene.getMap().layersConfiguration.getColorLayers();
+    for (var i = 0; i < colorLayers.length; i++) {
```
autra: For..of loops

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -67,12 +67,20 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     }
 
     layer.format = layer.options.mimetype || 'image/png';
+    layer.options.ipr = layer.options.ipr || '';
+    layer.options.ipr.description = layer.options.ipr.description || '';
+    layer.options.ipr.url = layer.options.ipr.url || '';
+    layer.options.ipr.logo = layer.options.ipr.logo || '';
+    layer.options.ipr.name = layer.options.ipr.name;
```
autra: well, I guess we don't need this line :-)

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -67,12 +67,20 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
...
     layer.crs = layer.projection || 'EPSG:4326';
     layer.width = layer.heightMapWidth || 256;
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
     layer.bbox = layer.bbox || new BoundingBox();
+    if (layer.options.ipr.name === undefined) {
+        throw new Error(`Missing parameter "layer.options.ipr.name" in layer : ${layer.id}`);
```
autra: Is this really a requirement? Is it possible for a data source to *not* have any known Intellectual Property?

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -71,6 +71,13 @@ WMTS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer
         options.mimetype = options.mimetype || 'image/png';
         options.style = options.style || 'normal';
         options.projection = options.projection || 'EPSG:3857';
+        options.ipr.description = options.ipr.description || '';
+        options.ipr.url = options.ipr.url || '';
+        options.ipr.logo = options.ipr.logo || '';
+        options.ipr.name = options.ipr.name;
```
autra: So all this defensive code will be virtually identical for every provider. Why not moving it in `ApiGlobe.js:preprocessLayer`? This method has already some similar initialization code and it would avoid duplicate codes.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    }
+    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
+    for (var j = 0; j < elevationLayers.length; j++) {
+        IPR.push(elevationLayers[j].options.ipr);
+    }
+    IPR = this.removeDuplicatesFromArray(IPR, 'name');
+    return IPR;
+};
+
+ApiGlobe.prototype.removeDuplicatesFromArray = function removeDuplicatesFromArray(originalArray, prop) {
```
autra: This should not be part of the api. It should be in a `utils` module somewhere. Or at least not in the prototype. But I actually think you don't need this, see my above comments.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,62 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    for (var i = 0; i < colorLayers.length; i++) {
+        IPR.push(colorLayers[i].options.ipr);
+    }
+    var elevationLayers = this.scene.getMap().layersConfiguration.getElevationLayers();
+    for (var j = 0; j < elevationLayers.length; j++) {
+        IPR.push(elevationLayers[j].options.ipr);
+    }
+    IPR = this.removeDuplicatesFromArray(IPR, 'name');
+    return IPR;
+};
```
autra: So to summarize, another possible implementation:

```js 
ApiGlobe.prototype.getLayersIPR = function getLayersIPR() {
    const lc = this.scene.getMap().layersConfiguration; // just a shortcut
    const result = new Set();
   // the spread operator allows for quick concatenation 
    return [...lc.getColorLayers(), ...lc.getElevationLayers()].forEach(l => result.add(l.options.ipr));
   // or return new Array(result) if you really need an array.
   // A Set seems more suitable here to me here, though.
};
```

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
autra: I mean, we're not supposed to change them during the execution right?

autra (commented):
Hey \o 

Just passing by, left some comments where I think things can be fancier :-)

index.html
vcoindet: Yes, I forgot to remove them (I used them for my tests)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: Well, it is required by the IGN's geoportal

src/Core/Commander/Providers/WMS_Provider.js
vcoindet: I will check that, I will discuss this with the IGN's Geoportal

src/Core/Commander/Providers/WMTS_Provider.js
vcoindet: OK, I will check that

index.html
vcoindet: done

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: done

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: Wasn't exactly what I wanted, I used a Map instead.
I also used `for...of` loop

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: removed

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: done

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: Some layers might not have any IPR specified in the json, so it's interesting to have a method to set an IPR for those layers.

src/Core/Commander/Providers/WMS_Provider.js
vcoindet: removed

src/Core/Commander/Providers/WMS_Provider.js
vcoindet: removed

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: Note that I specified the attribut `param.ipr.name` mandatory in this method. If one choose to specify an IPR, this one must have one argument at least.

vcoindet:
rebased and ready

examples/layers/JSONLayers/DARK.json
```diff
@@ -3,6 +3,12 @@
     "id": "DARK",
     "customUrl": "http://a.basemaps.cartocdn.com/dark_all/%TILEMATRIX/%COL/%ROW.png",
     "options": {
+    	"ipr": {
+    		"name":"CARTO",
+    		"url": "https://carto.com/",
+    		"logo":"",
```
peppsac: What's the point of adding empty properties (logo, description)?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -73,7 +73,6 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
     layer.bbox = layer.bbox || new BoundingBox();
-
```
peppsac: Unrelated change (it adds noise to the PR so I'd rather not have this change here)

peppsac (changes requested):

autra:
Actually... @vcoindet can you explain to me the rationale of adding such apis in Itowns? I actually don't see why itowns should be aware of such a thing. It is the client application (here the geoportal) which adds layers to itowns (which has the ambition of being a framework, let's recall that), so this geoportal is perfectly aware of these IPR informations and can display them independently (they are not part of WMTS nor WMS specs right?). It can even query the servers independently of itowns.

I really feel that the reasoning is backward here. Itowns should not be (in my humble opinion) a helper for the geoportal. Rather, it should focus on it's goal : be able to display all the layers (of all the possible type) as easily as possible, and allow to configure stuff like cameras, controls etc. Thoughts?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,50 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+ * @return {array}  An array of IPR.
+ */
+
+ApiGlobe.prototype.getLayersIPR = function getLayersIPR() {
+    const lc = this.scene.getMap().layersConfiguration;
+    const colorLayers = lc.getColorLayers();
+    const map = new Map();
+    for (const i in colorLayers) {
+        map.set(colorLayers[i].options.ipr.name, colorLayers[i].options.ipr);
+    }
```
autra: I meant 

```js 
for (const layer of lc.getColorLayers()) {
    map.set(layer.options.ipr.name, layer.options.ipr);
}
```
- use for..of loops
- Matter of personal opinion, but no need to have a variable `colorLayers` when you are accessing the field only once.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +852,50 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    const elevationLayers = lc.getElevationLayers();
+    for (const j in elevationLayers) {
+        map.set(elevationLayers[j].options.ipr.name, elevationLayers[j].options.ipr);
+    }
+    map.delete(undefined);
```
autra: ?

nosy-b:
@autra I share your point of view but adding an optional credit properties in the layer description will not transform iTowns into a heavy library.
Yes, French Geoportal or other website using itowns should then handle themself how they want the credit to be displayed.

vcoindet:
Well, I guess that itowns should be able to display all the layers, but ain't we supposed to mention their sources ?
I agree that some layers won't specify their IPR in their files, that's why I removed the defensive code WMS_Provider et WMTS_Provider.

It won't be useful for the geoportal only, as one is supposed to mention the sources of the layers displayed on iTowns... That's why I developed these methods.

gcebelieu:
@autra , @nosy-b  I agree with @vcoindet answer : IPR feature goes further than a simple geoportal need : you may have to quote geographical data originators to meet their conditions of use when they exist. 

If this may reassure you : this feature is proposed by many frameworks such as CesiumJS, OpenLayers or Leaflet...

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: ok, I see

autra:
> If this may reassure you : this feature is proposed by many frameworks such as CesiumJS, OpenLayers or Leaflet...

Because they display it. If we want to display it too, then I suggest adding the displaying code to this PR. It's a logically linked change and would make the rationale more obvious :-) If we do not, then really we don't need this API imho.

Also, let's call the object `attribution` and not `IPR`, like Cesium, Leaflet etc... all do. `attribution` is now apparently a de facto standard and it would make more sense for newcomers who are used to these frameworks.

Furthermore, another reason why it's important to have the displaying code: it affects the api design. For example at the moment you're returning a map. Whether or not this is relevant will depend on how we use this API. FYI, Cesium, Leaflet and OpenLayers all gives back html and not a Map. 

Concerning the IPR internals, they use a simpler model: Cesium and Leaflet have only a text field (a description). OpenLayers adds an href. Do we really need a logo and 2 text fields? Maybe description and href are enough (and indeed you did fill `description` and `logo` to empty strings in your PR), WDYT?

I guess you still need a defensive code in preprocessLayer, as attributions are optionals: they might not be set in the json.

autra:
Also not sure `ApiGlobe.js` is the right place for that. There should be some `layer` objects like OpenLayers has where you can add these attributions, but I guess rationalizing the api is not the goal of this PR...

examples/layers/JSONLayers/DARK.json
vcoindet: removed

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: done

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: Some IPR might not be specified, so I chose to remove it from the map object

vcoindet:
Rebased and Ready

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -917,6 +917,39 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
     this.viewerDiv.removeEventListener('layerchanged:index', this.callbackLayerChanged, false);
 };
 
+/**
+ * Get the IPR (Intellectual Property Rights) of all layers in the scene.
```
autra: Let's update the comments too

examples/layers/JSONLayers/DARK.json
```diff
@@ -3,6 +3,10 @@
     "id": "DARK",
     "customUrl": "http://a.basemaps.cartocdn.com/dark_all/%TILEMATRIX/%COL/%ROW.png",
     "options": {
+    	"ipr": {
```
autra: let's call them "attribution" ?

src/Core/Commander/Providers/WMS_Provider.js
autra: +1, would pollute `git log -- src/Core/Commander/Providers` for example.

vcoindet:
rebased and ready

autra:
IIRC after discussing with the geoportal team, we just need to expose an event containing `zoom` and `extent` changes, then they would calculate which attribution to display (according to their metadata) themselves. So I guess we don't need this PR any more right?

gcebelieu:
> So I guess we don't need this PR any more right?

well, there is still a need that can be reusable for itowns framework. Let me expose it :

I think iTowns API should give the possibility to attach one or several Attribution(s) object(s)  to a layer object (the object structure may be discussed further). 

The Attribution Widget role is to aggregate all attributions attached to layers currently visible on the globe (this aggregation is processed at every layer change event that occurs on the globe) and display an aggregated result on a given DOM target. The aggregation may be smart enough to avoid displaying the same attribution twice.

The final application (here geoportail) will listen to zoom and center changed events in order to dynamicaly feed Attribution property of displayed layers. The Attribution Widget will then refresh itself automaticaly.

What do you think ?

autra:
> I think iTowns API should give the possibility to attach one or several Attribution(s) object(s) to a layer object (the object structure may be discussed further).

Why not? but:

> The Attribution Widget role is to aggregate all attributions attached to layers currently visible on the globe (this aggregation is processed at every layer change event that occurs on the globe) and display an aggregated result on a given DOM target. The aggregation may be smart enough to avoid displaying the same attribution twice.

So not a widget, just a div on one corner of Itowns, like Leaflet does for instance. In this case, I repeat myself: this PR should include the displaying code.

> The final application (here geoportail) will listen to zoom and center changed events in order to dynamicaly feed Attribution property of displayed layers. The Attribution Widget will then refresh itself automaticaly.

So what you are suggesting is 
construct time
1/ itowns provides a template file (Widget.js) 
2/ Geoportail instanciate a class MyWidget.js
3/ geoportail feeds this instance to itowns
4/ itowns remembers it

then runtime !
5/ user zoom
6/ itowns sends an event
7/ geoportal code (in MyWidget probably) get the event and extract a metadata
8/ geoportal sends the metadata *back* to itowns
9/ itowns orders the widget to refresh, calling a function on it.

Don't you see how this is a nonsense technically? Especially steps 7, 8 and 9 ? What I'm saying is simple: you don't need itowns to manage your widgets, and you'll be better off. Free yourself :-) Really, I'm not speaking only as a itowns dev. I'm also a user of itowns on other projects. Forcing myself to implement (nonsense in javascript) a "class" whose structure is forced on me by the framework is a pain.

Beside it is: 
1/ Orthogonal to the current PR (most of all, you want zoom/center changed events).
2/ highly specific to geoportail needs, and imho this dynamic behaviour should not be in Itowns core. It's currently not clear and debatable that this need would be shared by other client apps.

Again, you just need to have a `ExtendChangedEvent`, and your client code would just be

```js
itowns.viewer.addEventListener('extend-changed-event', function(ev) {
    let extend = ev.extend;
    let zoom = ev.zoom
    /* your code in step 7 */
   /* your code in function called during step 9 */
});
```
Don't you see how it is *simpler*, for everybody?

autra:
That being said 

> highly specific to geoportail needs, and imho this dynamic behaviour should not be in Itowns core. 

is my opinion. I guess the best course of action for this feature request would be to open an issue tagged with `enhancement`.

gcebelieu:
> So not a widget, just a div on one corner of Itowns, like Leaflet does for instance. In this case, I repeat myself: this PR should include the displaying code.

Yeah, I totaly agree with that : this widget (or whatever you want to call it) should also display its result in a div.

> So what you are suggesting is
> construct time
> ... 
> Don't you see how this is a nonsense technically? Especially steps 7, 8 and 9 ? What I'm saying is simple: you don't need itowns to manage your widgets, and you'll be better off. Free yourself :-) Really, I'm not speaking only as a itowns dev. I'm also a user of itowns on other projects. Forcing myself to implement (nonsense in javascript) a "class" whose structure is forced on me by the framework is a pain.


I was not writting about a generic Widget (we've already discussed and agreed about that : "let's start with events and see" ) : I was just trying to explain what could be the perimeter of that attribution widget in iTowns API and how Geoportail (which has more complex attributions management) may handle with it (and event).

> I guess the best course of action for this feature request would be to open an issue tagged with enhancement

This issue has been created almost a year ago (#60 with an agreement of @vpicavet ) and conducted to this PR. We certainely may refine it and discuss about the perimeter, but please, let's go on forward on that subject.

vcoindet:
Rebased and Ready

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -917,6 +917,39 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+/**
+ * Get the Attribution of all layers in the scene.
+ * @return {array}  An array of attribution.
+ */
+
+ApiGlobe.prototype.getLayersAttribution = function getLayersAttribution() {
+    const lc = this.scene.getMap().layersConfiguration;
+    const map = new Map();
+    [...lc.getColorLayers(), ...lc.getElevationLayers()].forEach((l) => {
+        if (l.options.attribution != undefined) {
```
gchoqueux: if (l.options.attribution != undefined)

replace by

if (!l.options.attribution)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -917,6 +917,39 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
     this.viewerDiv.removeEventListener('layerchanged:index', this.callbackLayerChanged, false);
 };
 
+/**
+ * Get the Attribution of all layers in the scene.
+ * @return {array}  An array of attribution.
```
gchoqueux: Return map not Array

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -917,6 +917,39 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    return map;
+};
+
+/**
+ * Set the attribution for a layer.
+ * @param {param} Param - The id of the layer and the new IPR {id, attribution}.
+ */
+
+ApiGlobe.prototype.setLayerAttribution = function setLayerAttribution(param) {
+    if (param.attribution.name === undefined) {
```
gchoqueux: if (!param.attribution.name)

gchoqueux (changes requested):

gchoqueux:
I rebase this branch.

@peppsac Have you seen the changes?

vcoindet:
changes done @gchoqueux  and @peppsac 
rebased and ready

vcoindet:
rebased

vcoindet:
rebased

vcoindet:
rebased and ready @gchoqueux and @peppsac

gchoqueux (approved):

vcoindet:
If no one has any further change requested, I propose that this PR be merged by wenesday the 15th.
@peppsac @gchoqueux WDYT ?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -69,6 +69,7 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
 
     layer.bbox_url = layer.bbox_url || 'swne';
     layer.format = layer.options.mimetype || 'image/png';
+    layer.crs = layer.projection || 'EPSG:4326';
```
peppsac: Unrelated change, please remove.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -915,6 +915,39 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+    });
+    return map;
+};
+
+/**
+ * Set the attribution for a layer.
+ * @param {param} Param - The id of the layer and the new IPR {id, attribution}.
+ */
+
+ApiGlobe.prototype.setLayerAttribution = function setLayerAttribution(param) {
```
peppsac: I'm not sure this function is useful at all.
Is there a use case for modifying attribution at runtime ?
And also :
```js
api.setLayerAttribution({id: layer.id, attribution: { ...});
```
can be written in a simpler way:
```js
layer.attribution = { ... }
```
WDYT?

src/Core/Commander/Providers/WMS_Provider.js
vcoindet: Removed

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: I agree with you, it's simplier to do that way, I removed the function.

vcoindet:
Rebased and ready @peppsac

peppsac (approved):

peppsac:
LGTM, merging.

PR #222, opened 2016-12-14
Widget class

vcoindet:
Adding Widget class in example

Widget class with constructor and method (Widget.js)

In Widget.js
- getName()
- getElement()
- getOptions()
- setOptions(pOptions)
- listenToMap(pEventName, pCallback)
- getGlobe()
- setGlobe(map)

Method in widgetAPI.js
- addWidget(widget)
- getWidgets()
- removeWidget(widget)

To make the examples work : put the following lines in index.html :
- < link rel="stylesheet" href="examples/widgets/widget.css" > in the 'head' tag
- <script src="examples/widgets/pickPositionWidget.js"> </script>
  <script src="examples/widgets/rangeWidget.js"> </script> 
  Right before the closure of the 'body' tag
- <script src="examples/widgets/Widget.js"> </script>
  <script src="examples/widgets/widgetAPI.js"> </script>
  Just after the line <script src="dist/itowns.js"></script>
- const widgetTab = new widgetAPI();

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-*	text=auto
+*   text=auto
 *.js eol=lf
 *.css eol=lf
 *.glsl eol=lf
--- a/examples/widgets/Widget.js
+++ b/examples/widgets/Widget.js
@@ -1,0 +1,104 @@
+/**
+ * Generated On: 2016-11-22
+ * Class: Controls
+ * Description: Create a widget.
+ */
+
+function Widget(pName, pElement, options) {
+    this.pName = pName;
+    this.pElement = pElement;
+    this.options = options;
+    this._map = null;
+}
+
+/**
+ * Return the name of the widget.
+ * @constructor
+ * @return     {name}  The name.
+ */
+
+Widget.prototype.getName = function getName() {
+    return this.pName;
+};
+
+/*
+ * Return the element used by the widget to display its GUI.
+ * @constructor
+ * @return     {element}  The element.
+*/
+
+Widget.prototype.getElement = function getElement() {
+    return this.pElement;
+};
+
+/**
+ * Return the options of the widget.
+ * @constructor
+ * @return     {object}  Object.
+ */
+
+Widget.prototype.getOptions = function getOptions() {
+    this.options = this.options || {};
+    return this.options;
+};
+
+/**
+ * Change the options of the widget.
+ * @constructor
+ * @param {object} pOptions - The new options of the conrtol.
+ */
+
+Widget.prototype.setOptions = function setOptions(pOptions) {
+    this.pName = pOptions.name;
+    this.pElement = pOptions.element;
+    this.options = pOptions.options;
+};
+
+/**
+ * Listen to an event linked to the map.
+ * @constructor
+ * @param {string} Eventname - The name of the event.
+ * @param {callback} Callback - The callback that is called when the event is heard.
+ */
+
+Widget.prototype.listenToMap = function listenToMap(pEventName, pCallback) {
+    document.getElementById('viewerDiv').addEventListener(pEventName, pCallback, false);
+};
+
+/**
+ * Remove an event linked to the map.
+ * @constructor
+ * @param {string} Eventname - The name of the event.
+ * @param {callback} Callback - The callback that is called when the event is heard.
+ */
+
+Widget.prototype.removeFromMap = function removeFromMap(pEventName, pCallback) {
+    document.getElementById('viewerDiv').removeEventListener(pEventName, pCallback, false);
+};
+
+/**
+ * Get the Map associated with the widget. Undefined if the widget is not added to a map.
+ * @constructor
+ */
+
+Widget.prototype.getMap = function getMap() {
+    return this._map;
+};
+
+/**
+ * Associate a map to a widget.
+ * @constructor
+ */
+
+Widget.prototype.setMap = function setMap(map) {
+    if (this._map) {
+        this.options.div.removeChild(this.pElement);
+    }
+    this._map = map;
+    if (this._map) {
+        var element = this.pElement;
+        this.options.div.appendChild(element);
+    }
+};
+
+/* export default Widget; */
--- a/examples/widgets/pickPositionWidget.js
+++ b/examples/widgets/pickPositionWidget.js
@@ -1,0 +1,37 @@
+/* global itowns */
+/* global Widget */
+/* global widgetTab */
+
+const pickWidget = document.createElement('div');
+pickWidget.id = 'pickWidget';
+
+const widgetsP = document.getElementById('widgets');
+
+widgetsP.appendChild(pickWidget);
+
+const widget = new Widget(
+    'Pick',
+    document.getElementById('pickWidget'),
+    {
+        div: widgetsP,
+    });
+
+const callback = function () {
+    widgetTab.addWidget(widget);
+    if (document.getElementById('pickList')) {
+        document.getElementById(widget.pElement.id).removeChild(document.getElementById('pickList'));
+    }
+    const tab = ['longitude', 'latitude', 'altitude'];
+    const coords = itowns.viewer.pickPosition().coordinate;
+    const list = document.createElement('ul');
+    list.id = 'pickList';
+    for (var j = 0; j < coords.length; j++) {
+        var item = document.createElement('li');
+        item.id = tab[j];
+        item.innerHTML = `${tab[j]} : ${coords[j]}`;
+        list.appendChild(item);
+    }
+    document.getElementById(widget.pElement.id).appendChild(list);
+};
+
+widget.listenToMap('mousedown', callback);
--- a/examples/widgets/rangeWidget.js
+++ b/examples/widgets/rangeWidget.js
@@ -1,0 +1,31 @@
+/* global itowns */
+/* global Widget */
+/* global widgetTab */
+
+const rangeWidget = document.createElement('div');
+rangeWidget.id = 'rangeWidget';
+
+const widgetsR = document.getElementById('widgets');
+
+widgetsR.appendChild(rangeWidget);
+
+const widgetR = new Widget(
+    'Range',
+    document.getElementById('rangeWidget'),
+    { div: widgetsR });
+
+const callbackR = function () {
+    widgetTab.addWidget(widgetR);
+    if (document.getElementById('rangeList')) {
+        document.getElementById(widgetR.pElement.id).removeChild(document.getElementById('rangeList'));
+    }
+    const coords = itowns.viewer.getRange();
+    const list = document.createElement('ul');
+    list.id = 'rangeList';
+    const item = document.createElement('li');
+    item.innerHTML = `range : ${coords}`;
+    list.appendChild(item);
+    document.getElementById(widgetR.pElement.id).appendChild(list);
+};
+
+widgetR.listenToMap('mousedown', callbackR);
--- a/examples/widgets/widgetAPI.js
+++ b/examples/widgets/widgetAPI.js
@@ -1,0 +1,34 @@
+/* global Widget */
+/* global itowns */
+
+function widgetAPI() {
+    this.widgets = new Set();
+}
+
+/**
+ * Add a widget.
+ */
+
+widgetAPI.prototype.addWidget = function addWidget(widget) {
+    widget.setMap(itowns.viewer.scene.getMap());
+    this.widgets.add(widget);
+};
+
+/**
+ * Returns all widgets.
+ * @return     {array}  The array of widgets.
+ */
+
+widgetAPI.prototype.getWidgets = function getWidget() {
+    return this.widgets;
+};
+
+/**
+ * Remove a widget.
+ * @param {object} Widget - The Widget object.
+ */
+
+widgetAPI.prototype.removeWidget = function removeWidget(widget) {
+    widget.setMap();
+    this.widgets.delete(widget);
+};
--- a/index.html
+++ b/index.html
@@ -8,29 +8,41 @@ and open the template in the editor.
     <head>
         <title>Globe</title>
 
-    <style type="text/css">
+    	<style type="text/css">
             html {height: 100%}
-        body { margin: 0; overflow:hidden; height:100%}
+
+    	    body {
+                margin: 0;
+                overflow:hidden;
+                height:100%
+            }
 
             #viewerDiv {
                 margin : auto auto;
                 width: 100%;
                 height: 100%;
                 padding: 0;
+                position: relative;
                 /*margin-top: 50vh;
                 transform: translateY(-50%);*/
             }
-            #menuDiv {position: absolute; top:0px; margin-left: 0px;}
-
-
+            #menuDiv {
+                position: absolute;
+                top:0px;
+                margin-left: 0px;
+            }
         </style>
         <meta charset="UTF-8">
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.6.1/dat.gui.min.js"></script>
+        <script src="examples/widgets/Widget.js"></script>
     </head>
     <body>
-        <div id="viewerDiv"></div>
+        <div id="viewerDiv">
+            <div id="widgets"></div>
+        </div>
+
         <script src="examples/GUI/GuiTools.js"></script>
         <script src="dist/itowns.js"></script>
         <script type="text/javascript">
@@ -42,6 +54,7 @@ and open the template in the editor.
             // iTowns namespace defined here
             const viewerDiv = document.getElementById('viewerDiv');
             const menuGlobe = new GuiTools(itowns.viewer, 'menuDiv');
+
             itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
 
             const wgs84TileLayer = {
@@ -76,7 +89,6 @@ and open the template in the editor.
                 // eslint-disable-next-line no-console
                 console.info('Globe Loaded');
             });
-
 </script>
     </body>
 </html>
--- a/examples/widgets/widgets.css
+++ b/examples/widgets/widgets.css
@@ -1,0 +1,10 @@
+#widgets {
+      position: absolute;
+      top: 1rem;
+      right: 1rem;
+      width: 20rem;
+      box-shadow: 1px 1px 10px #555;
+      font: 11px Lucida Grande,sans-serif;
+      color: #eee;
+      padding: 0.5rem;
+}


vcoindet:
rebased and ready

examples/controls/control.css
```diff
@@ -0,0 +1,12 @@
+#controlDiv {
+      position:absolute;
+      top:10px;
+      right:10px;
+      width:150px;
+      height:150px;
```
autra: nits: space before `:`

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +857,45 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
     this.viewerDiv.removeEventListener('layerchanged:index', this.callbackLayerChanged, false);
 };
 
+/**
+ * Add a control.
+ * @constructor
```
autra: Not a constructor! (and below too)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +857,45 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+ * @param {object} Control - The Control object.
+ */
+
+ApiGlobe.prototype.addControl = function addControl(control) {
+    var index = this.controls.indexOf(control);
```
autra: Maybe a Set is a more natural data structure here.

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * Generated On: 2016-11-22
+ * Class: Controls
+ * Description: Cette Classe construit un control.
```
autra: Should be in English

peppsac:
@vcoindet can you please squash or reword your commits? (5 commits use the same text)

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+
+function Control(pName, pElement, options) {
+    this.pName = pName;
+    this.pElement = pElement;
+    this.options = options;
+    this.map_ = null;
+    this.api = ApiGlobe;
+}
+
+Control.prototype.constructor = Control;
```
autra: you don't need this if you don't overwrite the `prototype`.

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+ * Description: Cette Classe construit un control.
+ */
+
+import ApiGlobe from 'Core/Commander/Interfaces/ApiInterface/ApiGlobe';
+
+function Control(pName, pElement, options) {
+    this.pName = pName;
+    this.pElement = pElement;
+    this.options = options;
+    this.map_ = null;
```
autra: Is it supposed to be a private field? In this case, the convention is to name it `this._map`.

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+ * @constructor
+ * @param {string} Eventname - The name of the event.
+ * @param {callback} Callback - The callback that is called when the event is heard.
+ */
+
+Control.prototype.listenToMap = function listenToMap(pEventName, pCallback) {
+    document.getElementById('viewerDiv').addEventListener(pEventName, pCallback, false);
+};
+
+Control.prototype.getGlobe = function getGlobe() {
```
autra: Might not be a globe in all cases (we want to make the geometry configurable in the future right?)

src/Core/Commander/Interfaces/Control.js
```diff
@@ -0,0 +1,67 @@
+/**
```
autra: One of `src/Core/Commander/Interfaces/Control.js` and `src/Core/Commander/Interfaces/Control/Control.js` should be removed I guess.

index.html
```diff
@@ -42,6 +53,7 @@
             // iTowns namespace defined here
             var viewerDiv = document.getElementById('viewerDiv');
             var menuGlobe = new GuiTools(itowns.viewer,'menuDiv');
+
```
autra: Not very critical, but in general it's best to remove unrelated changes from a PR. (And submit them elsewhere if needed).

index.html
```diff
@@ -68,7 +80,7 @@
 
             Promise.all([imageryPromise, elevationPromise]).then(() => itowns.viewer.init());
 
-            menuGlobe.addGUI('RealisticLighting',false,function(newValue) {itowns.viewer.setRealisticLightingOn(newValue);});
+            menuGlobe.addGUI('RealisticLighting',false,(newValue) => {itowns.viewer.setRealisticLightingOn(newValue);});
```
autra: Same remark, unrelated change.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -660,6 +663,7 @@ ApiGlobe.prototype.setRange = function setRange(pRange /* , bool anim*/) {
     eventRange.oldRange = this.getRange();
     loaded = false;
     this.scene.currentControls().setRange(pRange);
+    this.viewerDiv.dispatchEvent(eventRange);
```
autra: Be careful, events should not be used as static variable. THe browser sets some fields (`targets`, `innerTargets`). By using the same ref for every events, you risk some collision. It's best to create a new instance at each `dispatch`.

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+    return this.map_;
+};
+
+Control.prototype.setGlobe = function setGlobe(map) {
+    if (this.map_) {
+        document.getElementById(this.options.div.id).removeChild(this.pElement);
```
autra: If I'm not mistaken `this.options.div` is already the same dom element returned by `document.getElementById(this.options.div.id)`.

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+    this.map_ = map;
+    if (this.map_) {
+        var element = this.pElement;
+        document.getElementById(this.options.div.id).appendChild(element);
```
autra: same here.

src/Core/Commander/Interfaces/Control/Control.js
autra: `getMap` and `setMap`?

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+/**
+ * Return the options of the control.
+ * @constructor
+ * @return     {object}  Object.
+ */
+
+Control.prototype.getOptions = function getOptions() {
+    this.options = this.options || '';
```
autra: you meant `this.options = this.options || {};` right?

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Listen to an event linked to the globe.
+ * @constructor
+ * @param {string} Eventname - The name of the event.
+ * @param {callback} Callback - The callback that is called when the event is heard.
+ */
+
+Control.prototype.listenToMap = function listenToMap(pEventName, pCallback) {
+    document.getElementById('viewerDiv').addEventListener(pEventName, pCallback, false);
```
autra: there should be a `removeEventListener` somewhere, as apparently you can remove controls. Failure to do so would prevent the browser to dispose of your object.

examples/controls/control.css
autra: Consider using `rem` units instead of px to accomodate with different screen size and density.

examples/controls/control.css
```diff
@@ -0,0 +1,12 @@
...
+      background-color: white;
+      border-style: solid;
+      border-color: black;
+      border-width: 1px;
+      box-shadow: 1px 1px 10px #555;
```
autra: (for border and boxes, some people prefer staying with px, they think it's nicer)

examples/controls/control.css
```diff
@@ -0,0 +1,12 @@
+#controlDiv {
```
autra: This file should be called `controls.css`, and this div `controlsDiv`. Or even better `#controls`. The fact it is a div is an implementation detail.

examples/controls/control.css
```diff
@@ -0,0 +1,12 @@
...
+      top:10px;
+      right:10px;
+      width:150px;
+      height:150px;
+      background-color: white;
+      border-style: solid;
+      border-color: black;
+      border-width: 1px;
+      box-shadow: 1px 1px 10px #555;
+}
```
autra: I actually suggest:

```css
position: absolute;
top: 1rem;
right: 1rem;
width: 20rem;
box-shadow: 1px 1px 10px #555;
font: 11px Lucida Grande,sans-serif;
color: #eee;
padding: 0.5rem;
```

And separating the fields with `<br>` instead of `;` in your callbacks (in `pickposition_controls`).

Result:

screen shot 2017-01-16 at 15 22 27

examples/controls/pickPosition_control.js
```diff
@@ -0,0 +1,21 @@
+var pickControl = document.createElement('div');
```
autra: For the filename: either choose snake_case or CamelCase. (the rest of the project use CamelCase if I'm not mistaken).

examples/controls/pickPosition_control.js
```diff
@@ -0,0 +1,21 @@
...
+
+var controlDiv = document.getElementById('controlDiv');
+
+controlDiv.appendChild(pickControl);
+
+var control = new itowns.viewer.Control(
+    'Pick',
+    document.getElementById('pickControl'),
+    {
+        div: document.getElementById('controlDiv')
```
autra: `div: controlDiv`. You have it just above.

examples/controls/pickPosition_control.js
```diff
@@ -0,0 +1,21 @@
...
+    }
+);
+
+var callback = function () {
+    itowns.viewer.addControl(control);
+    document.getElementById(control.pElement.id).innerHTML = "longitude : " + itowns.viewer.pickPosition().coordinate[0] * 180 / 3.14 + " ; latitude : " + itowns.viewer.pickPosition().coordinate[1] * 180 / 3.14 + " ; altitude : " +  itowns.viewer.pickPosition().coordinate[2];
+};
+
+itowns.viewer.addEventListener('mousedown', callback);
```
autra: So I don't understand: how is `Control` useful if you do all the logic outside of it? You're not even using `listenToMap`... (though I think a callback in the `options` constructor parameter would be a better idea than listenToMap).

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -852,6 +857,45 @@ ApiGlobe.prototype.removeEventListenerLayerChanged = function removeEventListene
...
+ * @constructor
+ * @param {object} Control - The Control object.
+ */
+
+ApiGlobe.prototype.removeControl = function removeControl(control) {
+    var index = this.controls.indexOf(control);
+    if (index > -1) {
+        this.controls.splice(index, 1);
+    }
+    control.setGlobe();
```
autra: Let's be explicit `control.setGlobe(null);`

But it sounds like an abuse to me. I would prefer having a `control.dispose` method that would remove listeners and stuff.

index.html
```diff
@@ -68,7 +80,7 @@
...
 
            	itowns.viewer.addEventListener('globe-loaded', function() {
             	console.log("Globe Loaded");
```
autra: Actually, we could have several example html, one basic like it is now, another with all the controls etc... WDYT?

src/Core/Commander/Interfaces/Control/Control.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * Generated On: 2016-11-22
+ * Class: Controls
+ * Description: Cette Classe construit un control.
+ */
+
+import ApiGlobe from 'Core/Commander/Interfaces/ApiInterface/ApiGlobe';
```
autra: Not used anywhere. Also, cyclic dependencies should be avoided as much as possible (it creates difficult-to-maintain code).

autra (commented):
As nobody was looking at your PR, I took upon myself to do it :-) 

Can you explain to me what is the exact role of `Controls`? Is it to display information from some events? if so, maybe we can find another name (controls is for controlling stuff. We're not really controlling them here). `indicators`? `sensors`?

Also, my general feeling about this is that it is quite a rigid way to give information to client code. Don't we just want some events (optionally embedding useful informations, like new values) and let client code (geoportal, anything) deal with that how they see fit? 

As it is now, if I just want to store range info somewhere (for history or telemetry purpose for example), I cannot do that without displaying them. But maybe I didn't get the rationale here.

examples/controls/control.css
vcoindet: OK I agree

examples/controls/control.css
vcoindet: Will work on that

examples/controls/control.css
vcoindet: I don't usually use `<br>` for semantic reasons. I'd rather use `<li>`. I agree that `;` wasn't a good solution, but it was just an example to test the control class

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: You're probably right, didn't know this method at the time. Will change that. Or Maybe a `new Map()` can do the job too.

src/Core/Commander/Interfaces/Control/Control.js
vcoindet: You're absolutely right, I forget to remove it

src/Core/Commander/Interfaces/Control/Control.js
vcoindet: OK

src/Core/Commander/Interfaces/Control/Control.js
vcoindet: Well, I'm not sure of that. I have to test it, I will let you know

src/Core/Commander/Interfaces/Control.js
vcoindet: Oups, my mistake, the first one has to be removed...

examples/controls/pickPosition_control.js
vcoindet: lowerCamelCase if I'm not mistaken too. The best is that i choose lowerCamelCase.

examples/controls/pickPosition_control.js
vcoindet: absolutely !

examples/controls/control.css
autra: yep :-) Indeed, `li` are even better :-)

src/Core/Commander/Interfaces/Control/Control.js
vcoindet: Yep, I guess you're right, I hesitated between those two choices, but I only thought about 3D and not 2D

examples/controls/pickPosition_control.js
autra: Actually yes, lowerCamelCase, thanks for translating for me ;-)

vcoindet:
To be clear on the purpose of the Controls class : 
A control is a widget with a DOM element which position should be fixed on the screen (determined by using CSS) . It can involve inputs or be informational only.
You can use it for simple custom controls by creating the element with listeners, for example displaying the IPRs on the screen : IPRs can change when you remove a layer (event `layerchanged:visible`) or when you change the center of the map.

Here the control class is inspired on the OpenLayers 3's control class.

All the methods implemented in the Controls class can be used to help you display your control on the screen. So basically, it is used to create DOM elements and all controls inherits of this class.
Some examples of controls : menu, IPR, scale bar, mini Globe, zoom and dezoom button, etc... (see the screenshot for IPRs below)
ipr

index.html
vcoindet: I totally agree, as controls like the menu, IPR, range, picking, etc... should be examples.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: So I guess we should change that for all dispatched events. Maybe in an another PR ? WDYT ?

autra:
Thanks for the context !

So actually, those Controls would exactly do dat.gui's job + displaying read-only informations, right? If I understand it correctly, why don't we use dat.js for this? It's already a very good state-of-the-art lib, with more than 1000 stars on github... and we are already using it! So it would mean less maintenance for us, and a better integration with the currently enabled controls. 

For example, with the current master, with a small modification to GuiTools.js (return the result of `this.gui..add` on line 74), I can achieve a range spy just with these following lines in index.html:
```js
let myUi = menuGlobe.addGUI('range', {itowns.viewer.getRange()}, (newValue) => { /* do nothing */ });
myUi.domElement.children[0].disabled = true; // make this read-only
itowns.viewer.addEventListener('rangeChanged', () => {
    console.log('new range', itowns.viewer.getRange());
    menuGlobe.range = itowns.viewer.getRange();
    myUi.updateDisplay();
});
```
So with some polishing (creating a new method in GuiTools or modifiying `addGUI` by adding a `readonly` boolean parameter and an event name to listen to), we already have everything we need, no need to create another module for that (+ some css etc).

Note that the example above only works if you manually execute `itowns.viewer.setRange(...)` in the console, but that's because we don't correctly fire `rangechanged` events when we do it from GlobeControls.js. So this needs to get fixed anyway (rangechanged events should be fired from GlobeControls I guess).

So to summarize, I thing that:
- we should fix our events in ApiGlobe and ensure they are fired when they should.
- we should use dat.gui for these controls.

One last thing:

> A control is a widget

So let's call them Widgets? Because for us, controls are a different thing (mouse and keyboard actions), ex GlobeControls.js (and I hope to make them configurable and to have FPSControls, PanoramicControls etc...).

src/Core/Commander/Interfaces/Control/Control.js
vcoindet: Yep, you're right, so I changed it

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
autra: yep, you're right :-)

vcoindet:
Commit squashed, Rebased and Ready
Changed the words 'controls' by 'widgets' as suggested by @autra

vcoindet:
Rebased and Ready

Put everything in examples, please read the description of this PR to make it work correctly.

.gitattributes
```diff
@@ -1,4 +1,4 @@
-*	text=auto
+*   text=auto
```
gchoqueux, line 1: reason?

examples/widgets/Widget.js
```diff
@@ -0,0 +1,104 @@
...
+};
+
+/**
+ * Associate a map to a widget.
+ * @constructor
+ */
+
+Widget.prototype.setMap = function setMap(map) {
+    if (this._map) {
+        this.options.div.removeChild(this.pElement);
```
gchoqueux, line 95: why remove and re-add this.pElement?

examples/widgets/Widget.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+    this._map = map;
+    if (this._map) {
+        var element = this.pElement;
```
gchoqueux, line 99: remove `var element` and replace `this.options.div.appendChild(element);` by `this.options.div.appendChild(this.pElement);`

examples/widgets/pickPositionWidget.js
```diff
@@ -0,0 +1,37 @@
+/* global itowns */
```
gchoqueux, line 1: rename by PickPositionWidget.js

examples/widgets/Widget.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * Generated On: 2016-11-22
+ * Class: Controls
+ * Description: Create a widget.
+ */
+
+function Widget(pName, pElement, options) {
```
gchoqueux, line 7: export Widget

examples/widgets/pickPositionWidget.js
```diff
@@ -0,0 +1,37 @@
...
+/* global Widget */
```
gchoqueux, line 2: import Widget

examples/widgets/pickPositionWidget.js
```diff
@@ -0,0 +1,37 @@
...
+/* global widgetTab */
```
gchoqueux, line 3: remove global  widgetTab 

User must do widgetTab.add(pickPositionWidget);

Do example to test pickPositionWidget

examples/widgets/pickPositionWidget.js
```diff
@@ -0,0 +1,37 @@
...
+        div: widgetsP,
+    });
+
+const callback = function () {
+    widgetTab.addWidget(widget);
+    if (document.getElementById('pickList')) {
+        document.getElementById(widget.pElement.id).removeChild(document.getElementById('pickList'));
+    }
+    const tab = ['longitude', 'latitude', 'altitude'];
+    const coords = itowns.viewer.pickPosition().coordinate;
```
gchoqueux, line 25: remove global api call, pass in api in constructor

examples/widgets/rangeWidget.js
```diff
@@ -0,0 +1,31 @@
+/* global itowns */
```
gchoqueux, line 1: no global variable and use import export, see comment for pick

rename file by RangeWidget.js

gchoqueux: add example

examples/widgets/widgetAPI.js
```diff
@@ -0,0 +1,34 @@
+/* global Widget */
+/* global itowns */
+
+function widgetAPI() {
```
gchoqueux, line 4: widgetApi is not really necessary

gchoqueux (changes requested):

examples/widgets/pickPositionWidget.js
```diff
@@ -0,0 +1,37 @@
...
+    const list = document.createElement('ul');
+    list.id = 'pickList';
+    for (var j = 0; j < coords.length; j++) {
```
gchoqueux, line 28: replace by for ... in

gchoqueux: sorry replace by for ... of

examples/widgets/Widget.js
autra: I might be late to the party, but you cannot really do that: the examples are neither handled by webpack nor by babel. Imho they should be seen as ready-to-use snippets of code for client apps. Therefore, they shouldn't rely on webpack, babel or anything linked to our build system. They should be valid, browser-ready javascript for our target browsers.

This actually means that:
- we need to formally decide which our target browsers (and versions) are (because it defines which js api/functions/syntax we can safely use)
- we need a special `.eslintrc` for the `examples` folder

examples/widgets/pickPositionWidget.js
autra: (if for..of is supported by our target browsers, we won't use babel here I guess)

examples/widgets/Widget.js
vcoindet: Sorry for the late answer... You're not too late ! As you mentionned it, it's not possible to do that in `examples`.
In fact, I thought maybe we could do a new `index.html` in `/examples/` for each widget's examples, but it won't solve the problem.
I agree with your solution, I think it's the best we can think about right now. But I don't have any knowledge on `.eslintrc`, so do you agree to help me to do a special `.eslintrc` for the `examples` folder ?

PR #225, opened 2017-01-10
Remove JavaTools, that wasn't actually doing anything

autra:
Slice actually allocates a new array. Therefore the original array was left untouched. The proper way to deallocate an array is to assign [] or null or undefined to the var and let the garbage collector do its job.

As this code is very old, I guess emptying those array wasn't very useful after all. So I prefer to keep the same behaviour as before and to not empty those arrays.

@iTowns/reviewers please r?


tbroyer (approved):

src/Renderer/LayeredMaterial.js
```diff
@@ -172,14 +171,6 @@ LayeredMaterial.prototype.dispose = function dispose() {
             }
         }
     }
-
-    var jT = new JavaTools();
-
-    jT.freeArray(this.textures[0]);
```
gchoqueux, line 178: > The proper way to deallocate an array is to assign [] or null or undefined to the var and let the garbage collector do its job.

it's necessary to write this?

```js
this.textures[0] = null;
this.uniforms.dTextures_00.value = null;
this.uniforms.dTextures_01.value = null;
```

gchoqueux:
why you merge from master 44a5ff7dc0338b6a046335e77c10cf18396abed4?

src/Renderer/LayeredMaterial.js
peppsac: No it's not needed.
One word was missing in @autra sentence: "The proper way to **explicitely** deallocate ...".
Even without setting these to `null` the GC will realize they are unreachable and thus candidate for deallocation.

autra: @peppsac actually, when they are fields, they might never go out of scope, so it might be necessary to assign null, undefined, or better, an empty array to deallocate the previous one.

@gchoqueux I didn't do it here, because it was working before, and I was afraid of the side-effect if I did fix it. ("if it ain't broken, don't fix it"). I wanted to stay iso with previous *real* behaviour. It would always be time to optimize later if we feel we need it.

autra: But yeah, we might never need to explicitly deallocate them.

autra: Anyway, `dispose` is never called, and *generally*, it's better to let the gc dispose of everything (if there is no event listeners). It will optimize natively, pretty sure it is always better than doing it explicitly. And if not, it's a browser bug :-)

PR #226, opened 2017-01-16
Linter fixes

peppsac:
A batch of fixes for the linter rules.

I'm trying a new thing for git commits: using a prefix describing the type of commit: chore, docs, feat, fix, refactor, style, or test.

Also I added an exception for 2 lines breaking the 'no-loop-func': let me know if you have a better fix :)


autra:
>  the type of commit: chore, docs, feat, fix, refactor, style, or test.

Could you detail when we should use one or the other? (or put the link where you got the idea?)

autra:
>  the type of commit: chore, docs, feat, fix, refactor, style, or test.

Could you detail when we should use one or the other? (or put the link where you got the idea?)

vpicavet:
@peppsac if you want this to become a general rule ( sounds good), please propose a PR on the CONTRIBUTING.md file

peppsac:
Here's a short article about what prefixes can be used, and when: http://karma-runner.github.io/1.0/dev/git-commit-msg.html

@vpicavet good idea, I'll open a PR

src/Core/Commander/Providers/BuildingBox_Provider.js
```diff
@@ -81,12 +81,12 @@ BuildingBox_Provider.prototype.generateMesh = function generateMesh(elements, bb
     var altitude_ground = altitude - 1.5; // 35;  // truck height
 
     for (var r = 0; r < features.length; r++) {
-        var hauteur = (features[r].properties.hauteur + suppHeight) || 0;
-        var z_min = altitude_ground; // features[r].properties.z_min;  // altitude_ground // force altitude ground
+        const hauteur = (features[r].properties.hauteur + suppHeight) || 0;
+        const z_min = altitude_ground; // features[r].properties.z_min;  // altitude_ground // force altitude ground
         var polygon = features[r].geometry.coordinates[0][0];
 
+        var arrPoint2D = [];
```
autra: In this case, this should be `const` right? (or let?)

src/Core/Commander/Providers/BuildingBox_Provider.js
```diff
@@ -167,10 +167,10 @@ BuildingBox_Provider.prototype.generateMesh = function generateMesh(elements, bb
     if (this.rtcOn) {
         firstPos = _geometry.vertices[0].clone();
         // create pivot from 1st pos vertex
-        for (var i = 0; i < _geometry.vertices.length; ++i) {
+        for (let i = 0; i < _geometry.vertices.length; ++i) {
```
autra: When we see this, let's have the reflex to change for a `for..of` loops.

src/Globe/TileGeometry.js
```diff
@@ -232,7 +232,7 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
             return id;
         };
 
-        buildUVSkirt = function buildUVSkirt() {
+        buildUVSkirt = function buildUVSkirt(id) {
```
autra, line 235: I'm curious: why this change? I'm not against it, but what was the linter saying?

autra: Because we're still capturing other variables here...

src/MobileMapping/Sensor.js
```diff
@@ -87,6 +87,7 @@ Sensor.prototype.getMatOrientationCapteur = function getMatOrientationCapteur()
         case 2:
             return ori2;
         case 3:
+        default:
```
autra, line 90: Not logically equivalent to previous code. Can you elaborate why this works?

src/MobileMapping/Shader.js
```diff
@@ -323,7 +325,7 @@ var Shader = {
 
         '         gl_FragColor = vec4(colorpoint,alpha);',
 
-        '      }	',
+        '      }',
```
autra, line 328: Wow. What *was* this char?

src/Scene/BrowseTree.js
```diff
@@ -173,15 +173,16 @@ BrowseTree.prototype.updateLayer = function updateLayer(layer, camera) {
 
     var root = layer.children[0];
 
-    for (var c = 0; c < root.children.length; c++) {
-        var node = root.children[c];
+    for (let c = 0; c < root.children.length; c++) {
+        const node = root.children[c];
 
         var cRTC = function getCRtcFn() {
```
autra, line 179: You don't have a no-loop-func error here?

src/Renderer/ProjectiveTexturingMaterial.js
```diff
@@ -241,14 +241,16 @@ var ProjectiveTexturingMaterial = {
...
             _infos.cam = Ori.sensors[i].infos; // console.log(_infos.cam);
-            m = idmask[i];
+            const m = idmask[i];
             if (m >= 0) {
+                // eslint-disable-next-line no-loop-func
                 this.loadTexture(Ori.getMask(i), {}, (tex, m) => {
                     _shaderMat.uniforms.mask.value[m] = tex;
                 }, m);
             }
+            // eslint-disable-next-line no-loop-func
```
autra: I did

```diff
+        function setMaskOnLoad(tex, m) {
+            _shaderMat.uniforms.mask.value[m] = tex;
+        }
+
+        function setTextureOnLoad(tex, i) {
+            _shaderMat.uniforms.texture.value[i] = tex;
+        }
+
         _infos.pano = panoInfo;
         _infos.lod = _infos.lods[0];
         for (i = 0; i < N; ++i) {
             _infos.cam = Ori.sensors[i].infos; // console.log(_infos.cam);
             m = idmask[i];
             if (m >= 0) {
-                this.loadTexture(Ori.getMask(i), {}, (tex, m) => {
-                    _shaderMat.uniforms.mask.value[m] = tex;
-                }, m);
+                this.loadTexture(Ori.getMask(i), {}, setMaskOnLoad, m);
             }
-            this.loadTexture(_infos.url, _infos, (tex, i) => {
-                _shaderMat.uniforms.texture.value[i] = tex;
-            }, i);
+            this.loadTexture(_infos.url, _infos, setTextureOnLoad, i);
         }
         this.changePanoTextureAfterloading(panoInfo, posFrameWithPivot, rot, 1);
```

If I'm not mistaken, the only variables captured by these closures are from outside the loop. So putting the function declaration outside of it should cause no problems.

src/Scene/BrowseTree.js
autra: Oh I get it! It's the `const` declaration, that correctly scopes the variable and avoid the coding error described in the doc where your closure doesn't contain what you believe it does.

autra (commented):
Nice :-) I suggested a fix for the disabled `no-loop-func` errors.

Generally speaking, I think it's better to avoid declaring func in loops for perf issue, but I guess that's outside the scope of this PR. Let's keep it focused on linter errors :-) (and be done with this quick!)

src/MobileMapping/Shader.js
autra: Oh, right it's a tab. On the split view, it was displayed as a 0-lenght char

src/Globe/TileGeometry.js
peppsac: That's because `id` is modified inside the loop, while others variables aren't.

src/MobileMapping/Sensor.js
peppsac: Definining an arbitrary default value can't hurt, can it? (the alternative would be to raise an error I guess)

src/Renderer/ProjectiveTexturingMaterial.js
peppsac: Thanks

src/MobileMapping/Sensor.js
autra: It depends on how normal the "this.orientation is not 0, 1, 2 or 3" is... I honestly don't know here so I leave that to your fine judgment :-)

autra (approved):
LGTM :-)

gchoqueux (commented):
LGTM

PR #227, opened 2017-01-17
Linter fixes

autra:
Another PR for linter. It also activates the linter on the examples/ folder.

This PR removes a bunch of unused files. Less code, less errors to fix. Also `npm run build` is now about 1sec faster on my machine after these changes.

@iTowns/reviewers


src/Core/System/Capabilities.js
```diff
@@ -38,22 +38,7 @@ Capabilities.prototype.ioFile = function ioFile() {
 };
 
 Capabilities.prototype.isInternetExplorer = function isInternetExplorer() {
-    return /* @cc_on!@*/ false || !!document.documentMode;
+    return false || !!document.documentMode;
```
peppsac, line 41: The `/* @cc_on!*/` is a trick to detect IE < 11 - so it may be needed.
(but as IE < 11 isn't supported by three.js maybe this check is useless anyway)

src/Core/defaultValue.js
```diff
@@ -7,7 +7,7 @@
 
 import { Vector3 } from 'three';
 
-var defaultValue = function defaultValue(value, def) {
+const defaultValue = function defaultValue(value, def) {
```
peppsac, line 10: What about removing this `defaultValue` thing completely?
Most of the time `a = defaultValue(foo, bar)` can be replaced with `a = foo || bar`

src/Scene/Description/Style.js
```diff
@@ -1,154 +0,0 @@
-/**
```
peppsac, line 1: I think this class was recently introduczed, and it's going to be used soon.
The same goes for `Palette.js`, `StyleManager.js` and maybe `Description.js`.

@gchoqueux can you confirm?

Note that this is the 2nd time someone tries to remove it so it confirms that introducing *soon-to-be-used* code is a baaaad idea.

peppsac (commented):
Looks good.
Regarding Style/Description/... I don't mind removing them and reintroduce them later, when they actually serve a purpose.

src/Scene/Description/Style.js
autra: > Note that this is the 2nd time someone tries to remove it so it confirms that introducing soon-to-be-used code is a baaaad idea.

All wip code should be in a branch imho :-)

src/Core/defaultValue.js
autra: Yep, I hesitated, but decided against it, because `defaultValue` has quite a few use. I thought it would need another Pr.

src/Core/System/Capabilities.js
autra: Yep I guess so :-)

src/Scene/Description/Style.js
gchoqueux: Yes this code is going to be used soon

autra: > Yes this code is going to be used soon

Then whoever wants them just need to run
```bash
git checkout acea086 -- src/Scene/Description
```

That will bring them back in the tree. In the meantime, let's not put the burden to maintain dead code on *all* contributors (I have already 2 commits in master that touches these files).

src/Scene/Scene.js
```diff
@@ -94,18 +89,6 @@ Scene.prototype.getPickPosition = function getPickPosition(mouse) {
     return this.gfxEngine.getPickingPositionFromDepth(mouse);
 };
 
-Scene.prototype.getStyle = function getStyle(name) {
```
gchoqueux, line 97: I need this for WFS

src/Scene/Scene.js
```diff
@@ -51,12 +49,9 @@ function Scene(coordinate, ellipsoid, viewerDiv, debugMode, gLDebug) {
     this.managerCommand = ManagerCommands(this);
     this.orbitOn = false;
 
-    this.stylesManager = new StyleManager();
```
gchoqueux, line 54: I need this for WFS

gchoqueux (changes requested):

peppsac:
So let's move forward: @autra can you restore the following files :
* `Description.js`
* `DescriptionManager.js`
* `Palette.js`
* `Style.js`
* `StyleManager.js`

But from now I'll be careful : no more soon-to-be-used code will be merged, because it causes unnecessary maintenance issues.

gchoqueux:
And restore scene.js with style function...

autra:
> So let's move forward: @autra can you restore the following files :

> And restore scene.js with style function...

So be it! Files restored :-) Please check that it's ok though.

autra:
Ok this time, it's really ready. thanks @peppsac for being careful.

gchoqueux (approved):

PR #233, opened 2017-01-23
Upgrade packages

gchoqueux:
Make 'npm install' after this upgrade

* dependencies

  * es6-promise     3.2.1  ->   4.0.5
  * babel-runtime   6.9.2  ->   6.20.0
  * imports-loader  0.6.5  ->   0.7.0
  * jszip           2.5.0  ->   3.1.3
  * simd            2.0.0  ->   2.0.0
  * three           0.83.0 ->   0.84.0
  * whatwg-fetch    1.0.0  ->   2.0.2
  * custom-event    1.0.0  ->   1.0.1

* devDependencies

  * babel-cli           6.11.4  ->  6.22.2
  * babel-core          6.11.4  ->  6.22.1
  * babel-loader        6.2.4   ->  6.2.10
  * babel-preset-es2015 6.9.0   ->  6.22.0
  * eslint              3.10.2  ->  3.14.0
  * webpack             1.12.12 ->  1.14.0
  * webpack-dev-server  1.14.1  ->  1.16.2
  * babel-plugin-transform-runtime  6.9.0  ->   6.22.0
  * eslint-import-resolver-webpack  0.7.0  ->   0.8.1


src/Core/Commander/Providers/Fetcher.js
```diff
@@ -2,6 +2,8 @@ import { TextureLoader } from 'three';
 
 const textureLoader = new TextureLoader();
 
+textureLoader.setCrossOrigin('');
```
peppsac: Why is this change needed?

gchoqueux: new THREE.js v84 throw cross origin error when loads images

peppsac: Then it needs a comment stating this.
But why do you use `''` and not the real origin?

gchoqueux: I don't know these mechanisms
Have you idea?

autra: If I'm not mistaken, `setCrossOrigin` [forwards to](https://github.com/mrdoob/three.js/blob/dev/src/loaders/ImageLoader.js#L71) img tag [crossOrigin property](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin). Possible values are `anonymous` and `with-credentials`. (empty is equivalent to `anonymous` if I'm not mistaken).

I guess we probably want `with-credentials`? Just in case there is a http auth on the wms/wmts/whatever server...

tbroyer: This is https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-crossorigin (`TextureLoader` delegates to `ImageLoader` which simply creates a `new Image()`)

Based on that, `setCrossOrigin('')` is equivalent to `setCrossOrigin('anonymous')`, and only `''`, `'anonymous'` and `'use-credentials'` are valid values.

The real change in r84 is that `ImageLoader` used to delegate to a `FileLoader`/`XHRLoader` (the `new Image` path in r83 was only triggered [when `crossOrigin` was not `undefined`](https://github.com/mrdoob/three.js/commit/73a4ca56b51dcd9480bcbf62b22c904dcb6c8ffb), or when loading a `data:` URL; prior to r83, only `data:` URLs triggered the `new Image` path, everything else going through the `XHRLoader`), but that was removed in r84: https://github.com/mrdoob/three.js/commit/e7f7d330b975b1e7d3b760a3b943976406bfa814

So, cross-origin images used to work until r84 because they were loaded as blobs _via_ an XHR, but they're now loaded "as images", so they respect the canvas security policy wrt cross-origin media.

tbroyer: > I guess we probably want with-credentials? Just in case there is a http auth on the wms/wmts/whatever server...

Maybe this should be configurable per source?

In any case, the XHR wasn't previously done with `withCredentials=true` (in iTowns), so if this change is made, it should be separate from the Three.js update from this PR.

gchoqueux: So it suits you for this PR?

tbroyer: > So it suits you for this PR?

As far as I'm concerned, yes. Maybe use `anonymous` instead of the empty string to make it more explicit?

peppsac: +1 for `anonymous`

gchoqueux: 'anonymous' is equivalent ''?

peppsac: Yes, but the second one is explicit (ie: doesn't require the reader to know that `''` maps to `'anonymous'`)

gchoqueux: ok I replace

gchoqueux: Done

peppsac:
Thanks, merging.

PR #238, opened 2017-01-25
Debug mode

autra:
This PR contains a POC of debug infos and charts. It adds a `__DEBUG__` var that is true only on dev mode. All the code inside `if (__DEBUG__)` blocks are removed in prod mode by the minifier.

@iTowns/reviewers  please r?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,6 +44,10 @@ module.exports = {
     }],
     'one-var': ['error', 'never'],
     // TODO reactivate this rule once a proper npm package is made
+    // a good configuration might be:
+    /*'import/no-extraneous-dependencies': ['error', {
+      devDependencies: ['test/**', 'tests/**', 'examples/**'],
+    }],*/
     'import/no-extraneous-dependencies': 'off',
 
     // TODO reactivate all the following rules
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -68,5 +68,8 @@ module.exports = {
     'brace-style': 'off',
     'new-cap': 'off',
     'no-continue': 'off',
+  },
+  "globals": {
+    "__DEBUG__": false
   }
 }
--- a/examples/index.html
+++ b/examples/index.html
@@ -1,0 +1,82 @@
+<!DOCTYPE html>
+<!--
+To change this license header, choose License Headers in Project Properties.
+To change this template file, choose Tools | Templates
+and open the template in the editor.
+-->
+<html>
+    <head>
+        <title>Globe</title>
+
+    <style type="text/css">
+            html {height: 100%}
+        body { margin: 0; overflow:hidden; height:100%}
+
+            #viewerDiv {
+                margin : auto auto;
+                width: 100%;
+                height: 100%;
+                padding: 0;
+                /*margin-top: 50vh;
+                transform: translateY(-50%);*/
+            }
+            #menuDiv {position: absolute; top:0px; margin-left: 0px;}
+
+
+        </style>
+        <meta charset="UTF-8">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <script src="examples/GUI/GuiTools.js"></script>
+        <script src="dist/itowns2.js"></script>
+        <script type="text/javascript">
+
+            /* global itowns,document,GuiTools*/
+
+            const positionOnGlobe = { longitude: 2.3465, latitude: 48.88, altitude: 25000000 };
+
+            // iTowns namespace defined here
+            const viewerDiv = document.getElementById('viewerDiv');
+            const menuGlobe = new GuiTools(itowns.viewer, 'menuDiv');
+            itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
+
+            const wgs84TileLayer = {
+                protocol: 'tile',
+                id: 'wgs84',
+            };
+
+            const urlImageryLayers = [];
+            const urlElevationLayers = [];
+
+            urlImageryLayers.push('examples/layers/JSONLayers/Ortho.json');
+            urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
+            urlImageryLayers.push('examples/layers/JSONLayers/ScanEX.json');
+            urlImageryLayers.push('examples/layers/JSONLayers/Region.json');
+
+            urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT.json');
+            urlElevationLayers.push('examples/layers/JSONLayers/IGN_MNT_HIGHRES.json');
+
+            itowns.viewer.addGeometryLayer(wgs84TileLayer);
+            const imageryPromise = itowns.viewer.addImageryLayersFromJSONArray(urlImageryLayers);
+            const elevationPromise = itowns.viewer.addElevationLayersFromJSONArray(urlElevationLayers);
+
+            imageryPromise.then((layersColors) => { menuGlobe.addImageryLayersGUI(layersColors); });
+            elevationPromise.then((layersElevations) => { menuGlobe.addElevationLayersGUI(layersElevations); });
+
+            Promise.all([imageryPromise, elevationPromise]).then(() => itowns.viewer.init());
+
+            menuGlobe.addGUI('RealisticLighting', false,
+                (newValue) => { itowns.viewer.setRealisticLightingOn(newValue); });
+
+            itowns.viewer.addEventListener('globe-loaded', () => {
+                // eslint-disable-next-line no-console
+                console.info('Globe Loaded');
+            });
+
+</script>
+    </body>
+</html>
--- a/index.html
+++ b/index.html
@@ -33,6 +33,7 @@ and open the template in the editor.
         <div id="viewerDiv"></div>
         <script src="examples/GUI/GuiTools.js"></script>
         <script src="dist/itowns.js"></script>
+        <script src="dist/debug.js"></script>
         <script type="text/javascript">
 
             /* global itowns,document,GuiTools*/
@@ -42,7 +43,8 @@ and open the template in the editor.
             // iTowns namespace defined here
             const viewerDiv = document.getElementById('viewerDiv');
             const menuGlobe = new GuiTools(itowns.viewer, 'menuDiv');
-            itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
+            const scene = itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
+            itowns.debug = new Debug(scene);
 
             const wgs84TileLayer = {
                 protocol: 'tile',
--- a/index.html
+++ b/index.html
@@ -27,7 +27,7 @@ and open the template in the editor.
         <meta charset="UTF-8">
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.6.1/dat.gui.min.js"></script>
+        <script src="node_modules/dat.gui/build/dat.gui.min.js"></script>
     </head>
     <body>
         <div id="viewerDiv"></div>
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
-    "start": "webpack-dev-server -d --inline --hot",
+    "start": "NODE_ENV=development webpack-dev-server -d --inline --hot",
     "prepublish": "npm run build"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "eslint": "^3.14.0",
     "eslint-config-airbnb-base": "^11.0.1",
     "eslint-import-resolver-webpack": "^0.8.1",
+    "dat.gui": "^0.6.1",
     "eslint-loader": "^1.6.1",
     "eslint-plugin-import": "^2.2.0",
     "imports-loader": "^0.7.0",
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "eslint": "^3.14.0",
     "eslint-config-airbnb-base": "^11.0.1",
     "eslint-import-resolver-webpack": "^0.8.1",
+    "chart.js": "^2.4.0",
     "dat.gui": "^0.6.1",
     "eslint-loader": "^1.6.1",
     "eslint-plugin-import": "^2.2.0",
--- a/src/Renderer/LayeredMaterial.js
+++ b/src/Renderer/LayeredMaterial.js
@@ -20,7 +20,6 @@ emptyTexture.level = EMPTY_TEXTURE_LEVEL;
 const layerTypesCount = 2;
 var vector = new THREE.Vector3(0.0, 0.0, 0.0);
 var vector4 = new THREE.Vector4(0.0, 0.0, 0.0, 0.0);
-var showDebug = false;
 var fooTexture;
 
 export const l_ELEVATION = 0;
@@ -87,8 +86,9 @@ const LayeredMaterial = function LayeredMaterial(id) {
     this.fragmentShaderHeader += `const int   TEX_UNITS   = ${nbSamplers.toString()};\n`;
     this.fragmentShaderHeader += pitUV;
 
-    if (showDebug)
-        { this.fragmentShaderHeader += '#define DEBUG\n'; }
+    if (__DEBUG__) {
+        this.fragmentShaderHeader += '#define DEBUG\n';
+    }
 
     // see GLOBE FS
     this.fragmentShaderHeader += getColorAtIdUv(nbSamplers);
--- a/src/Renderer/Shader/GlobeFS.glsl
+++ b/src/Renderer/Shader/GlobeFS.glsl
@@ -25,7 +25,6 @@ uniform vec3        lightPosition;
 
 // Options global
 uniform bool        selected;
-uniform bool        debug;
 uniform bool        lightingOn;
 
 varying vec2        vUv_WGS84;
@@ -33,6 +32,7 @@ varying float       vUv_PM;
 varying vec3        vNormal;
 
 #if defined(DEBUG)
+    uniform bool showOutline;
     const float sLine = 0.008;
 #endif
 
@@ -51,7 +51,7 @@ void main() {
     #endif
 
     #if defined(DEBUG)
-         if(vUv_WGS84.x < sLine || vUv_WGS84.x > 1.0 - sLine || vUv_WGS84.y < sLine || vUv_WGS84.y > 1.0 - sLine)
+         if (showOutline && (vUv_WGS84.x < sLine || vUv_WGS84.x > 1.0 - sLine || vUv_WGS84.y < sLine || vUv_WGS84.y > 1.0 - sLine))
              gl_FragColor = CRed;
          else
     #endif
@@ -145,8 +145,4 @@ void main() {
             gl_FragColor.rgb *= light;
         }
     }
-
-    if(debug)
-       gl_FragColor = vec4( 1.0, 1.0, 0.0, 1.0);
-
 }
--- a/utils/debug/Debug.js
+++ b/utils/debug/Debug.js
@@ -1,0 +1,29 @@
+/* global menuGlobe */
+
+/**
+ * Create a debug instance attached to an itowns instance
+ *
+ * @Constructor
+ * @param {Scene} scene the itowns Scene
+ * @return {Debug} a debug instance
+ */
+// disabling eslint errors as it is the exported constructor
+function Debug(scene) {
+    const gui = menuGlobe.gui.addFolder('Debug Tools');
+
+    const state = {
+        showOutline: false,
+        wireframe: false,
+    };
+    // tiles outline
+    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
+        scene.map.layersConfiguration.getGeometryLayers()[0].showOutline = newValue;
+        scene.map.tiles.children[0].traverse((object) => {
+            if (object.materials) {
+                object.materials[0].uniforms.showOutline = { value: newValue };
+            }
+        });
+        scene.renderScene3D();
+    });
+}
+window.Debug = Debug;
--- a/utils/debug/Debug.js
+++ b/utils/debug/Debug.js
@@ -1,4 +1,5 @@
 /* global menuGlobe */
+import Chart from 'chart.js';
 
 /**
  * Create a debug instance attached to an itowns instance
@@ -9,12 +10,95 @@
  */
 // disabling eslint errors as it is the exported constructor
 function Debug(scene) {
+    // CHARTS
+    // create charts div
+    const chartDiv = document.createElement('div');
+    chartDiv.id = 'chart-div';
+    chartDiv.style = 'position: absolute; bottom: 0; left: 0; width: 100vw; height: 20rem; background-color: white; display: none';
+
+    scene.viewerDiv.appendChild(chartDiv);
+
+    // line graph for nb elements
+    const nbObjectsCanvas = document.createElement('canvas');
+    nbObjectsCanvas.heigth = '20rem';
+    nbObjectsCanvas.width = '40rem';
+    nbObjectsCanvas.id = 'nb-objects';
+    chartDiv.appendChild(nbObjectsCanvas);
+
+    const timestamp = Date.now();
+    const nbObjectsDataset = { label: 'Number of object in Scene', data: [{ x: 0, y: 0 }] };
+    const nbObjectsChartLabel = ['0s'];
+    const nbObjectsChart = new Chart('nb-objects', {
+        type: 'line',
+        data: {
+            labels: nbObjectsChartLabel,
+            datasets: [nbObjectsDataset],
+        },
+        options: {
+            animation: { duration: 10 },
+            scales: {
+                yAxes: [{
+                    display: true,
+                    ticks: {
+                        suggestedMin: 0, // minimum will be 0, unless there is a lower value.
+                    },
+                }],
+            },
+        },
+    });
+
+    function updateNbObjectsChart() {
+        function countElem(node) {
+            if (!node) {
+                return 0;
+            }
+            let count = 1; // this node
+            if (node.children) {
+                for (const child of node.children) {
+                    count += countElem(child);
+                }
+            }
+            return count;
+        }
+
+        const timeInS = Math.floor((Date.now() - timestamp) / 1000);
+
+        const limit = 50;
+        nbObjectsChartLabel.push(`${timeInS}s`);
+        if (nbObjectsChartLabel.length > limit) {
+            nbObjectsChartLabel.shift();
+        }
+        nbObjectsDataset.data.push({ x: timeInS, y: countElem(scene.gfxEngine.scene3D) });
+        if (nbObjectsDataset.data.length > limit) {
+            nbObjectsDataset.data.shift();
+        }
+        nbObjectsChart.update();
+    }
+
+    // hook that to scene.update
+    const oldUpdate = Object.getPrototypeOf(scene).update.bind(scene);
+    scene.update = function debugUpdate() {
+        oldUpdate();
+        updateNbObjectsChart();
+    };
+
+    // DEBUG CONTROLS
     const gui = menuGlobe.gui.addFolder('Debug Tools');
 
     const state = {
         showOutline: false,
         wireframe: false,
+        displayCharts: false,
     };
+
+    // charts
+    gui.add(state, 'displayCharts').name('Display charts').onChange((newValue) => {
+        if (newValue) {
+            chartDiv.style.display = 'flex';
+        } else {
+            chartDiv.style.display = 'none';
+        }
+    });
 
     function applyToNodeFirstMaterial(cb) {
         scene.getMap().tiles.children[0].traverse((object) => {
--- a/utils/debug/Debug.js
+++ b/utils/debug/Debug.js
@@ -15,15 +15,30 @@ function Debug(scene) {
         showOutline: false,
         wireframe: false,
     };
-    // tiles outline
-    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
-        scene.map.layersConfiguration.getGeometryLayers()[0].showOutline = newValue;
-        scene.map.tiles.children[0].traverse((object) => {
+
+    function applyToNodeFirstMaterial(cb) {
+        scene.getMap().tiles.children[0].traverse((object) => {
             if (object.materials) {
-                object.materials[0].uniforms.showOutline = { value: newValue };
+                cb(object.materials[0]);
             }
         });
         scene.renderScene3D();
+    }
+
+    // tiles outline
+    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
+        scene.map.layersConfiguration.getGeometryLayers()[0].showOutline = newValue;
+        applyToNodeFirstMaterial((material) => {
+            material.uniforms.showOutline = { value: newValue };
+        });
+    });
+
+    // tiles wireframe
+    gui.add(state, 'wireframe').name('Wireframe').onChange((newValue) => {
+        scene.map.layersConfiguration.getGeometryLayers()[0].wireframe = newValue;
+        applyToNodeFirstMaterial((material) => {
+            material.wireframe = newValue;
+        });
     });
 }
 window.Debug = Debug;
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -10,12 +10,15 @@ var definePlugin = new webpack.DefinePlugin({
 });
 
 module.exports = {
-  entry: [ 'es6-promise', 'whatwg-fetch', 'custom-event', path.resolve(__dirname, 'src/Main.js') ],
+  entry: {
+      itowns: [ 'es6-promise', 'whatwg-fetch', 'custom-event', path.resolve(__dirname, 'src/Main.js') ],
+      debug: [ path.resolve(__dirname, 'utils/debug/Debug.js') ]
+  },
   devtool: 'source-map',
   output: {
     path: path.resolve(__dirname, 'dist'),
-    filename: 'itowns.js',
-    library: 'itowns',
+    filename: '[name].js',
+    library: '[name]',
     libraryTarget: 'umd',
     umdNamedDefine: true
   },
@@ -25,7 +28,8 @@ module.exports = {
         test: /\.js$/,
         include: [
           path.resolve(__dirname, 'src'),
-          path.resolve(__dirname, 'test')
+          path.resolve(__dirname, 'test'),
+          path.resolve(__dirname, 'utils')
         ],
         loader: 'eslint'
       }
@@ -35,7 +39,8 @@ module.exports = {
         test: /\.js$/,
         include: [
           path.resolve(__dirname, 'src'),
-          path.resolve(__dirname, 'test')
+          path.resolve(__dirname, 'test'),
+          path.resolve(__dirname, 'utils')
         ],
         loader: 'babel'
       },
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,9 +1,13 @@
 var path = require('path');
-
+var webpack = require('webpack');
 
 // THREE js replace
 //"three": "^0.74.0" -> "three": "mrdoob/three.js#35a5994828da7cebc0d8442062f784b3f9e1f818",
 //                                               #idcommit
+
+var definePlugin = new webpack.DefinePlugin({
+    __DEBUG__: JSON.stringify(process.env.NODE_ENV === 'development')
+});
 
 module.exports = {
   entry: [ 'es6-promise', 'whatwg-fetch', 'custom-event', path.resolve(__dirname, 'src/Main.js') ],
@@ -62,5 +66,6 @@ module.exports = {
   },
   devServer: {
     publicPath: '/dist/'
-  }
+  },
+  plugins: [definePlugin]
 };
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -11,6 +11,8 @@ The following people have contributed to iTowns 2.
 * [Oslandia](http://www.oslandia.com)
   * [Vincent Picavet](https://github.com/vpicavet)
   * [Jérémy Gaillard](https://github.com/Jeremy-Gaillard)
+  * [Pierre-Éric Pelloux-Prayer](https://github.com/peppsac)
+  * [Augustin Trancart](https://github.com/autra)
 
 * [AtolCD](http://www.atolcd.com)
   * [Thomas Broyer](https://github.com/tbroyer)
--- a/autodeploy.sh
+++ b/autodeploy.sh
@@ -41,7 +41,8 @@ cp -R dist out/
 # generate the API documentation
 npm run doc -- -d out/API_Doc
 # Copy demo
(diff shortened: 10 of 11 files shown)


index.html
```diff
@@ -27,7 +27,7 @@
         <meta charset="UTF-8">
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.6.1/dat.gui.min.js"></script>
+        <script src="node_modules/dat.gui/build/dat.gui.min.js"></script>
```
tbroyer, line 30: `autodeploy.sh` would need to be updated after such a change.

A better approach would probably be to move that "example" into its own script that'd be processed by webpack (another `entrypoint`) so it could bundle dat.gui.

.eslintrc.js
```diff
@@ -43,6 +43,10 @@ module.exports = {
       }
     }],
     'one-var': ['error', 'never'],
+    'import/no-extraneous-dependencies': ['error', {
+      /* the following folders can import devDependencies */
+      devDependencies: ['test/**', 'tests/**', 'src/debug/**', 'examples/**'],
```
tbroyer: `src/debug` shouldn't import dev dependencies (though optional dependencies would be OK I believe).

src/debug/Debug.js
```diff
@@ -0,0 +1,112 @@
+import dat from 'dat.gui/build/dat.gui';
```
tbroyer: Why is this not using `dat.gui` here? Why use `dat.gui/build/dat.gui`?

src/debug/Debug.js
```diff
@@ -0,0 +1,112 @@
...
+import LayeredMaterial from 'Renderer/LayeredMaterial';
+import Chart from 'chart.js/src/chart';
```
tbroyer: Chart.js' `package.json` has `"main": "src/chart.js",` so this should just be `from 'chart.js'`, with the exact same behavior in practice.

src/debug/Debug.js
```diff
@@ -0,0 +1,112 @@
...
+    // tiles outline
+    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
+        LayeredMaterial.prototype._showOutline = newValue;
+        var cO = function c0(object) {
+            if (object.materials) {
+                object.materials[0].uniforms.showOutline.value = newValue;
+            }
+        };
+
+        window.itowns.viewer.scene.getMap().tiles.children[0].traverse(cO);
```
tbroyer: Couldn't this use `import whatever from '../Main'` then `whatever.scene.…`?

webpack.config.js
```diff
@@ -1,10 +1,14 @@
 var path = require('path');
-
+var webpack = require('webpack');
 
 // THREE js replace
 //"three": "^0.74.0" -> "three": "mrdoob/three.js#35a5994828da7cebc0d8442062f784b3f9e1f818",
 //                                               #idcommit
 
+var definePlugin = new webpack.DefinePlugin({
+    __DEBUG__: JSON.stringify(JSON.parse(process.env.NODE_ENV === 'development'))
```
tbroyer: Is the `JSON.parse` really needed here‽ Looks to me like `JSON.stringify` would be enough.

.eslintrc.js
autra: Why? src/debug is used only for dev mode. That's the point of this folder.

src/debug/Debug.js
autra: Yeah I should have left a comment. Because we then need to process dat's files through babel, and we use incompatible babel configs with dat.gui :-/ It's possible I didn't find the way out though. If you have better, I'd be glad to hear it!

src/debug/Debug.js
autra: Indeed, correcting it.

src/debug/Debug.js
autra: Then I would need to remove the `import Debug` in Main.js (I don't know if ES6 imports can be circulars, but I would prefer to avoid it). Importing Debug from main and attaching it to the api in debug mode made more sense to me, WDYT?

index.html
autra: I would prefer to have webpack deal only with the framework and not touch anything linked to "client app" (which `index.html` really is). Actually to be a bit more precise, here is the way I see things:

- index.html is indeed an example, but one that targets developers (it's the entrypoint when you do `npm start`). It then makes sense to reference dev dependencies in it.
- We could add  a bunch of other use cases in `examples/` (or better `demos/`). Then one, some or all of these demos would be included in the github pages of this repo. These client apps would do their own stuff about whatever ui, external js they need (use CDN, have a js/ folder, whatever).

So what I propose for this PR: copy the old version of `index.html` in `demos/`, use this new one for the github page and keep the new version of `index.html` here, for developers.

@peppsac is working on a plan mode, so this would be another demos, and I plan on adding others with wmts from other sources (from USA National Map for example). 

WDYT?

webpack.config.js
autra: Indeed!

autra:
@tbroyer I pushed more commits, the last one illustrates my comment above.

Beware! this branch needs squashing (the fixup commits) before any merge.

.eslintrc.js
tbroyer: Wouldn't one goal be to make the sources usable "as-is", without webpack preprocessing? (or even simpler, and could –almost– be done today already: after conversion to ES5 through Babel)

Then developers using iTowns2 through NPM could choose to enable debug or not **in their projects**, and would have issues with the code not finding some dependencies. Now move those to `optionalDependencies` and everything's OK: want to enable debug? then add this dependency from our set of optional dependencies.

Now maybe I'm mistaken and that doesn't make sense; still it "smells" IMO that something in `src/` could depend on dev dependencies.

src/debug/Debug.js
tbroyer: Oh, OK, so more accurately, dat.gui's `package.json` is plain broken, referencing in its `main` a file using ES6 imports and needing Babel.

src/debug/Debug.js
tbroyer: Ah right, circular dependencies…

So, does `Debug` really need to be exposed in the API? (see comment above about using/enabling it from projects that use iTowns)

And if it does then could the code possibly be split to break the circular dependency? or could the public API be changed? (the only reason we'd need to import `Main` here is because it holds an `ApiGlobe` instance; that instance could be created "elsewhere", with `Main` importing it like it impors `Debug`, and `Debug` then importing that "elsewhere" too; or the public API could be changed to expose the `ApiGlobe` constructor and let users create their own instance of it, and `Debug` would be changed to receive an `ApiGlobe` instance to act on, as well as maybe a container `<div>`).

Feel free to punt those decisions to a later time though (add TODOs in the code or open an issue to track it).

.eslintrc.js
autra: I think I agree. Let me rephrase to be extra sure:

1. Either debug mode is not available *at all* for client app, and then it is not in `src/debug` but in `debug/` for example and `debug/` is added to `npmignore`
1. Or it is available to client app, and in this case, its dependencies are in `optionalDependencies`.

And you prefer option 2., is that right?

src/debug/Debug.js
autra: Yeah, and I tried to make babel have a pass on it (which is already a hack), but without success :-/

.eslintrc.js
autra: Btw our npm stuffs are broken too, if I'm not mistaken:

- we ignore src/
- we don't include any `dist/` folder even though we reference it in the main of our package.json.

Shouldn't we include src, and include dist too? (and update dist each time we release?)

src/debug/Debug.js
autra: > So, does Debug really need to be exposed in the API? (see comment above about using/enabling it from projects that use iTowns)

It's *nice* to do so. it's nice to be able to call debug methods from the webconsole.

> the public API could be changed to expose the ApiGlobe constructor and let users create their own instance of it

Actually, yes, this should be done: you want to be able to embed several instances of itowns in one webpage (no reason to limit to only one). In this case, creating the debug UI should be done in a per instance basis. So it's ApiGlobe that will import Debug, expose a `createDebugDisplay` on each instance. This method would attach a debug object to one instance of itowns. WDYT?

.eslintrc.js
tbroyer: Because `src/` contains code that use ES6 imports **and** rely on some Webpack configuration (for modules resolution for example), it should be in `.npmignore` (and it is, so we're fine).  
Then, AFAIK, `dist/` would be included in the package because it's not ignored (and it is generated by `npm build` during any "NPM release process", so there shouldn't be cases where it's not there).

What I had in mind at some point was to run Babel on `src/` to output ES5+commonjs to, say, `lib/`, then we could change our `main` to `lib/Main.js` (for people using iTowns through NPM and with Browserify, Rollup, or Webpack; so their transformations/bundling/optimizations can act on the real sources rather than the webpack-generated bundle). IIRC, this is not possible today mostly (or entirely) due to the module paths in `import` statements being relative to `src/` rather than relative to the script that imports them, and the importing of `.glsl` files (we'd have to use `require('fs').readFileSync(…)` instead, and use a BRFS or gslify transform during the Babel transpilation, or require that everyone does that BRFS or gslify transformation on their side; see https://gist.github.com/mattdesl/aaf759da84cc44c22305).

src/debug/Debug.js
tbroyer: Sounds good/better, but I'm not familiar with iTowns' codebase besides what's needed to _build_ it. From what I can grasp in a few tens of seconds reading the code though, SGTM.

.eslintrc.js
autra: > Then, AFAIK, dist/ would be included in the package because it's not ignored (and it is generated by npm build during any "NPM release process"

Oh right I didn't know `npm install` was done automatically for npm release. But I still think we should commit it when doing a git/github release, because it supports an npm installation directly from our github repo. Are these workflows compatible?

autra:
@tbroyer actually, I think it's better not to give the debug mode for external client for now. We could still add it afterwards, but I don't feel like committing to yet another public API for that. I will amend this PR with this in mind. 

So what I did:
- move `src/debug` to `debug/`
- stop doing `import debug`: the application still has debug code (inside `if (__DEBUG)` but do not depend on Debug.js anymore. Instead it can expose api only in debug mode that would get used by Debug.js. I think it's saner this way.

Please let me know what you think of this, if I forgot something or if I misunderstood you :-)

index.html
peppsac: If you expect people using `index.html` to install `devDependencies` then this must be documented in `README.md` or `BUILDING.md`.

webpack.config.js
```diff
@@ -1,16 +1,23 @@
 var path = require('path');
-
+var webpack = require('webpack');
 
 // THREE js replace
 //"three": "^0.74.0" -> "three": "mrdoob/three.js#35a5994828da7cebc0d8442062f784b3f9e1f818",
 //                                               #idcommit
 
+var definePlugin = new webpack.DefinePlugin({
+    __DEBUG__: JSON.stringify(process.env.NODE_ENV === 'development')
```
peppsac, line 9: I would prefer `dev`:  it's easier to write :)
Btw you may want to match `DEV` and `dev`.

peppsac: Nevermind, I didn't see that `npm start` was starting in dev mode.

peppsac (approved):
LGTM (but I'll wait for @tbroyer approval before merging)

.eslintrc.js
```diff
@@ -67,5 +71,8 @@ module.exports = {
     'brace-style': 'off',
     'new-cap': 'off',
     'no-continue': 'off',
+  },
+  "globals": {
+    "__DEBUG__": true
```
tbroyer: Should this be `false` to disallow setting `__DEBUG__`? (so it's only set by Webpack's DefinePlugin)

(the eslint-config-airbnb-base config enables `no-global-assign` at error level to get us covered, but IIUC `true` would bypass it)

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
+// importing the built version instead of root because of incompatibilities with babel build
```
tbroyer: I'd prefer _"because dat.gui is badly packaged and exposes ES6"_

index.html
gchoqueux: Keep old index.html in [README.md](https://github.com/iTowns/itowns2/blame/master/README.md#L29)  
> if you want to play with a demonstration, please see [http://www.itowns-project.org/itowns2/])

So change `autodeploy.sh`

and Add, if you want : dev link demonstration in readme.md

src/Renderer/LayeredMaterial.js
```diff
@@ -154,6 +153,18 @@ const LayeredMaterial = function LayeredMaterial(id) {
     // this.wireframe = true;
 
     this.colorLayersId = [];
+
+    if (__DEBUG__) {
+        this.uniforms.showOutline = {
+            type: 'i',
```
gchoqueux: use bool, three.js 
```
this.uniforms.showOutline = {  value: LayeredMaterial.prototype._showOutline }
```

The type is no longer needed

src/Renderer/LayeredMaterial.js
```diff
@@ -154,6 +153,18 @@ const LayeredMaterial = function LayeredMaterial(id) {
...
+            value: LayeredMaterial.prototype._showOutline,
+        };
+        this.uniforms.borderColor = {
+            type: 'v3',
```
gchoqueux: use THREE.Color
```
diffuseColor: { value: new THREE.Color()}
```

The type is no longer needed

src/Renderer/Shader/GlobeFS.glsl
```diff
@@ -33,7 +33,9 @@ varying float       vUv_PM;
 varying vec3        vNormal;
 
 #if defined(DEBUG)
+    uniform int showOutline;
```
gchoqueux: bool...

src/Renderer/Shader/GlobeFS.glsl
```diff
@@ -51,7 +53,7 @@ void main() {
     #endif
 
     #if defined(DEBUG)
-         if(vUv_WGS84.x < sLine || vUv_WGS84.x > 1.0 - sLine || vUv_WGS84.y < sLine || vUv_WGS84.y > 1.0 - sLine)
+         if (showOutline > 0 && (vUv_WGS84.x < sLine || vUv_WGS84.x > 1.0 - sLine || vUv_WGS84.y < sLine || vUv_WGS84.y > 1.0 - sLine))
```
gchoqueux: replace by
`if (showOutline && (vUv_WGS84.x < sLine || vUv_WGS84.x > 1.0 - sLine || vUv_WGS84.y < sLine || vUv_WGS84.y > 1.0 - sLine))`

gchoqueux (changes requested):
Dev mode is default mode, isn't it?

autra:
> Dev mode is default mode, isn't it?

When running `npm start`, yes. `npm run build` default to a production build.

index.html
autra: I'm not sure what you mean. I did change `autodeploy.sh` right?

> Keep old index.html in README.md

I didn't change `README.md`. 

As for a dev demonstration, I thought we could let it for people doing `npm start`. They would be the one interested in it anyway.

autra:
@gchoqueux I updated LayeredMaterials and GlobeFS according to your suggestions. Thanks!

demos/index.html
```diff
@@ -0,0 +1,82 @@
+<!DOCTYPE html>
```
gchoqueux: There's example folder, remove demos folder
I would like this index.html in root ghpage (no dev mode)
You add link to dev mode in README.md

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
+// importing the built version instead of root because dat.gui is badly packaged and exposes ES6
```
gchoqueux: move debug folder in examples Folder

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
...
+        updateNbObjectsChart();
+    };
+
+    // DEBUG CONTROLS
+    const gui = new dat.GUI();
+
+    const state = {
+        showOutline: false,
+        wireframe: false,
+        displayCharts: true,
```
gchoqueux: set displayCharts to false

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
...
+        if (newValue) {
+            chartDiv.style.display = 'flex';
+        } else {
+            chartDiv.style.display = 'none';
+        }
+    });
+
+    // tiles outline
+    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
+        var cO = function c0(object) {
```
gchoqueux: factorize in API or Scene

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
...
+                object.materials[0].uniforms.showOutline.value = newValue;
+            }
+        };
+
+        scene.getMap().tiles.children[0].traverse(cO);
+        scene.renderScene3D();
+    });
+
+    // tiles wireframe
+    gui.add(state, 'wireframe').name('Wireframe').onChange((newValue) => {
```
gchoqueux: factorize in API or Scene (same [comment](https://github.com/iTowns/itowns2/pull/238/files#r99136332))

gchoqueux (changes requested):
add option in GUI for this [mode](https://github.com/iTowns/itowns2/blob/master/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js#L354)

debug/Debug.js
peppsac: no no, debug belongs to itowns core.

debug/Debug.js
peppsac: what do you want to factorize exactly?

debug/Debug.js
gchoqueux: yes 
move in src/Core

demos/index.html
autra: >  I would like this index.html in root ghpage (no dev mode)

It's already the case. `demo/index.html` is the one deployed (see change in autodeploy). This index has debug mode disable.

autra: Ok for the move.

debug/Debug.js
autra: Debug is neither examples nor src. 

Everything that is in src/ is bundled in the npm package. `debug` won't. 

Examples contains example layers, snippets, demos etc. 

What is in `debug/` is more some internal/debug/dev tools. So it should be in its own folder, called either `debug`, `dev`, `tools` or...

autra:
@gchoqueux I pushed corrections, waiting for your answers.

autra:
Oh I forgot 

> add option in GUI for this mode

Can I call my joker card on this one? I have added them https://github.com/iTowns/itowns2/issues/245, but supporting a dynamic switch on these would need some more work (you need to have a `remove` function in Node, you have to dynamically change a lot of things that are currently in c3dEngine's constructor). I would prefer to have another PR for this. This one enables the basic debug infra + some examples in order to get started. Ofc we would need to work more on that later.

autra:
Just pushed a rename of the last 2 commits to comply with the new way of naming.

index.html
```diff
@@ -42,7 +43,8 @@
             // iTowns namespace defined here
             const viewerDiv = document.getElementById('viewerDiv');
             const menuGlobe = new GuiTools(itowns.viewer, 'menuDiv');
-            itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
+            let scene = itowns.viewer.createSceneGlobe(positionOnGlobe, viewerDiv);
```
gchoqueux: const scene

debug/Debug.js
gchoqueux: move to:
Utils/debug

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
...
+    // hook that to scene.update
+    const oldUpdate = Object.getPrototypeOf(scene).update.bind(scene);
+    scene.update = function debugUpdate() {
+        oldUpdate();
+        updateNbObjectsChart();
+    };
+
+    // DEBUG CONTROLS
+    const gui = new dat.GUI();
+
```
gchoqueux: add :
`gui.close();`

debug/Debug.js
gchoqueux: factorize `gui.add` showOutline, wireframe....

gchoqueux (changes requested):
uniform bool debug; in globeFS.glsl can be removed

debug/Debug.js
autra: What does the additional level of hierarchy bring?

debug/Debug.js
autra: Why?

debug/Debug.js
autra: It's a bit unclear to me. Please elaborate.

debug/Debug.js
gchoqueux: Not to be congested by interfaces

debug/Debug.js
gchoqueux: in utils, it will have other folders : build, converters, exporters, debug

debug/Debug.js
gchoqueux:
```js
gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
        var cO = function c0(object) {
            if (object.materials) {
                Object.getPrototypeOf(object.materials[0])._showOutline = newValue;
                object.materials[0].uniforms.showOutline.value = newValue;
            }
        };

        scene.getMap().tiles.children[0].traverse(cO);
        scene.renderScene3D();
    });
```

Redundant code -->

```js
// tiles wireframe
gui.add(state, 'wireframe').name('Wireframe').onChange((newValue) => {
        var cO = function c0(object) {
            if (object.materials) {
                Object.getPrototypeOf(object.materials[0])._wireframe = newValue;
                object.materials[0].wireframe = newValue;
            }
        };

        scene.getMap().tiles.children[0].traverse(cO);
        scene.renderScene3D();
});
```

autra: Not exactly :-)

`object.materials[0].uniforms.showOutline.value = newValue;` vs `object.materials[0].wireframe = newValue;`

I didn't see any viable factorization because of that. Can you write the code you had in mind here plz?

gchoqueux: 1. factorization
Ideally this part should be factorized and encapsulated in the Core and gui.add may be factorized in gui [examples](https://github.com/iTowns/itowns2/blob/master/examples/GUI/GuiTools.js)
```js
Ideally this part should be factorized and encapsulated in the Core
const setPropertyMaterial = (newValue, property, fn) => {
    const root = scene.getMap().tiles.children[0];
    Object.getPrototypeOf(root.children[0].materials[0])[`_${property}`] = newValue;
    const cO = function c0(object) {
        if (object.materials) {
            fn(object, newValue);
        }
    };
    root.traverse(cO);
    scene.renderScene3D();
};
const setShowOutLine = (object, newValue) => { object.materials[0].uniforms.showOutline.value = newValue; };
gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
    setPropertyMaterial(newValue, 'showOutline', setShowOutLine);
});
const setWireframe = (object, newValue) => { object.materials[0].wireframe = newValue; };
gui.add(state, 'wireframe').name('Wireframe').onChange((newValue) => {
    setPropertyMaterial(newValue, 'wireframe', setWireframe);
});
```

2. prototype change

I prefer add layer tiles's option than change Prototype, to have option debug by layer. I prefer remove this prototype change (Moreover the modification is made for each tile and it is not necessary) :
```js
Object.getPrototypeOf(root.children[0].materials[0])[`_${property}`] = newValue;
```

debug/Debug.js
```diff
@@ -0,0 +1,130 @@
...
+function Debug(scene) {
+    // CHARTS
+    // create charts div
+    const chartDiv = document.createElement('div');
+    chartDiv.id = 'chart-div';
+    chartDiv.style = 'position: absolute; bottom: 0; left: 0; width: 100vw; height: 20rem; background-color: white; display: none';
+
+    scene.viewerDiv.appendChild(chartDiv);
+
+    // line graph for nb elements
```
gchoqueux: when you display charts, there's bug with navigation

debug/Debug.js
autra: > Ideally this part should be factorized and encapsulated in the Core

This does not belong to Core imho. The idea is for the Core not to know anything about debug stuff. It's better to limit the use of `__DEBUG__` as much as we can. 

> I prefer add layer tiles's option than change Prototype, to have option debug by layer.

How could this work, as LayeredMaterial knows nothing about the layer? I'm not sure I understand.

>     Object.getPrototypeOf(root.children[0].materials[0])[`_${property}`] = newValue;

is it guaranteed that root.children[0].materials is defined?

debug/Debug.js
autra: Can you describe the bug please?

debug/Debug.js
autra: To elaborate on why I put that in the prototype: the current default value of wireframe and showOutline change at runtime (when we tick the boxes) and is to be shared by all instances of LayeredMaterial. Having them at the layer creation is not enough (and imho, the layerconfiguration should not depend on debug stuff). For these reasons, the prototype is a really good place for these booleans. If you're concerned about the fact we do it multiple times, I agree it's not necessary. I can make a test that will prevent any further assignment after the first one if you want, but I think we don't really care (setting boolean might even be less costly than making a test).

gchoqueux: how do you use the material to have one layer in `wireframe` (tiles by example) and other layer in solid (building by example)?

gchoqueux: > This does not belong to Core imho
I would prefer :
 * Factorize debug GUI in Examples/Gui 
 * encapsulate set wireframe and outline in api
 * move GUI debug in folder in original bottom layers.

It seems to me that the prototype modification also changes BasicMaterial

debug/Debug.js
gchoqueux: When you display charts, there's bug with navigation
Mouse navigation does not work properly

gchoqueux: Same problem without debug

gchoqueux: need a new PR to fix this

debug/Debug.js
autra: > how do you use the material to have one layer in wireframe (tiles by example) and other layer in solid (building by example)?

At the moment, the current behaviour of the checkbox is to activate it for all layers. It would always be time to change the location of this boolean when/if someone implements a wireframe checkbox for each layers. Let's be concerned about current behaviour and not future/hypothetic ones.

> Factorize debug GUI in Examples/Gui 

Debug is not examples right? It's an internal tools imo.

> encapsulate set wireframe and outline in api

For this, I would say it's outside the scope of this PR. This PR is concerned with debug tools to be used internally by us. We don't know yet if this belongs to the public API. Let's wait to see if there is a need (it's always either to add things to an API, but very difficult to remove them once it's there). It won't be difficult to add it in the future anyway.

> move GUI debug in folder in original bottom layers.

I don't understand what you're asking.

> It seems to me that the prototype modification also changes BasicMaterial

Nope, why would it?

peppsac:
I retested this PR and it contains some useful features and I'd really like to get this merged quickly.

So, could you summarize what are the issues you believe are blocking this PR @gchoqueux?

peppsac:
FWIW, the merits of this PR is to get developer oriented tools to help us building itowns.
I had a private branch doing the same thing for a long time and it's really helpful :)

This PR is definitely _not_ the place to modify the public API of iTowns or to perform cleanup of existing code: this kind of change deserve their own PR.

src/Renderer/LayeredMaterial.js
```diff
@@ -154,6 +154,12 @@ const LayeredMaterial = function LayeredMaterial(id) {
     // this.wireframe = true;
 
     this.colorLayersId = [];
+
+    if (__DEBUG__) {
+        this.uniforms.showOutline = { value: LayeredMaterial.prototype._showOutline };
```
gchoqueux: remove prototype use and layer configuration to pass property

src/Renderer/Shader/GlobeFS.glsl
```diff
@@ -33,7 +33,9 @@ varying float       vUv_PM;
 varying vec3        vNormal;
 
 #if defined(DEBUG)
+    uniform bool showOutline;
     const float sLine = 0.008;
+    uniform vec3 borderColor;
```
gchoqueux: `borderColor` isn't used

src/Renderer/LayeredMaterial.js
```diff
@@ -154,6 +154,12 @@ const LayeredMaterial = function LayeredMaterial(id) {
...
+        this.uniforms.borderColor = { value: new THREE.Color(1, 1, 1) };
```
gchoqueux: `borderColor` isn't used

utils/debug/Debug.js
```diff
@@ -0,0 +1,128 @@
...
+                cb(object.materials[0]);
+            }
+        });
+        scene.renderScene3D();
+    }
+
+    // tiles outline
+    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
+        applyToNodeFirstMaterial((material) => {
+            Object.getPrototypeOf(material)._showOutline = newValue;
```
gchoqueux: move this function `setShowLine` (or `enableBorderLine`) in the Core

utils/debug/Debug.js
```diff
@@ -0,0 +1,128 @@
...
+            material.uniforms.showOutline.value = newValue;
+        });
+    });
+
+    // tiles wireframe
+    gui.add(state, 'wireframe').name('Wireframe').onChange((newValue) => {
+        applyToNodeFirstMaterial((material) => {
+            Object.getPrototypeOf(material)._wireframe = newValue;
+            material.wireframe = newValue;
```
gchoqueux: move this function `enableWireframe` in the Core

gchoqueux (changes requested):

utils/debug/Debug.js
gchoqueux: It's not only for debug, there are lines on desktop globe world earth or map

utils/debug/Debug.js
gchoqueux: It's not only for debug, some users may have an interest in using the wireframe

utils/debug/Debug.js
peppsac: This should be done in another PR titled "feature: add capabilites to draw tile outlines". This new PR would move the  appropriate code to Core but also add options to control border color, thickness and maybe other attributes etc

src/Renderer/LayeredMaterial.js
peppsac: Yes. Note that currently it would be a configuration option of the **geometry** layer - because that's the way the shader work currently.
It would be nice to display border (with different color) per imagery layer, but again that's a different PR.

utils/debug/Debug.js
peppsac: Same answer: this PR is about debug features, so I'd rather keep the scope tight instead of pushing API changes.
And IMHO I don't really see what kind of users would be interested in enabling wireframe rendering?

src/Renderer/Shader/GlobeFS.glsl
autra: Yep indeed.

utils/debug/Debug.js
gchoqueux: I preferred in this PR

utils/debug/Debug.js
gchoqueux: I preferred `enableWireframe()` in this PR

utils/debug/Debug.js
peppsac: Nope. Seriously. This PR is about debug code.
You're free to submit a PR after this one that adds this feature though.

utils/debug/Debug.js
peppsac: Nope. See my previous comment.

gchoqueux: But the change of prototype doesn't suit me

peppsac: @autra is following your suggestion and moves the debug variables in LayersConfiguration, so it should be fine.

gchoqueux: These are not necessarily debug variables, users can use them

peppsac: Currently (on master) they are.
This PR simply enables a debug control to toggle it at runtime.
Later you'll be able to remove the `if (__DEBUG__) ...` code surrounding them if you want to add API support for them.

gchoqueux: I prefer, in this PR, that the variables wireframe and outline is not only for the debug

peppsac: Honestly it gets tiring: personal preference isn't an argument.

I stand by my point:
- on master you can't currently display wireframes or outline
- this PR is about adding debug  code to itowns, not about modifying user visible API

=> it doesn't make sense to expect this PR to modify the API (nor does it makes sense to add these variables in non-debug mode, because the functions handling them are only availble in debug mode)

package.json
```diff
@@ -39,6 +39,8 @@
     "eslint": "^3.14.0",
     "eslint-config-airbnb-base": "^11.0.1",
     "eslint-import-resolver-webpack": "^0.8.1",
+    "chart.js": "^2.4.0",
+    "dat.gui": "^0.6.1",
```
gchoqueux, line 43: it's the moment to use this package `dat.gui`

examples/index.html
```diff
@@ -0,0 +1,82 @@
...
+                transform: translateY(-50%);*/
+            }
+            #menuDiv {position: absolute; top:0px; margin-left: 0px;}
+
+
+        </style>
+        <meta charset="UTF-8">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.6.1/dat.gui.min.js"></script>
```
gchoqueux: at is the purpose of this line?

package.json
autra: Well, that's what I'm doing.

examples/index.html
gchoqueux: it's possible to use package `dat.gui`?

autra: Yep, in autodeploy.sh, I can copy it from node_modules to the gh_pages and use this one here, no problem.

utils/debug/Debug.js
gchoqueux: - Somes users can use wireframe for study alignment, occlusion, arrangement (use cartography architecture,  and debug). Many applications offer wireframe mode without using a debug mode, because they have a different purpose than debugging. 
- Somes users can use outline, it's a classic mapping tool ([example](http://mondecarte.o.m.f.unblog.fr/files/2014/11/carte-du-monde-9.jpg))
- Modifying prototypes does not allow wireframe by entity
- The option wireframe must be placed in BasicMaterial

peppsac: Point 1 and 2: good to know. That'll be fine for a later PR

Point 3: I already wrote that @autra is changing its PR to not modify the prototype. But it still won't allow to toggle wireframe by entity

Point 4: what is `the option wireframe` ? I guess it's different from `material.wireframe` since it already exists (it's a three.js feature: https://github.com/mrdoob/three.js/blob/dev/src/materials/ShaderMaterial.js#L38)

gchoqueux: > Point 1 and 2: good to know. That'll be fine for a later PR

It is added in this PR the right time to do it, and move in BasicMaterial

autra:
PR Updated. I don't touch LayeredMaterial.prototype any more.

It seems we won't reach consensus on the API change. For me, this means the API change can wait. Please note that this PR does not prevent it for the future. Better to merge what is working than to try to do too much in one PR right?

gchoqueux:
[`uniform int debug;`](https://github.com/autra/itowns2/blob/master/src/Renderer/Shader/GlobeFS.glsl#L32) It no longer makes sense

package.json
gchoqueux: it's not possible to use `import`?

utils/debug/Debug.js
```diff
@@ -0,0 +1,128 @@
...
+            if (object.materials) {
+                cb(object.materials[0]);
+            }
+        });
+        scene.renderScene3D();
+    }
+
+    // tiles outline
+    gui.add(state, 'showOutline').name('Show tiles outline').onChange((newValue) => {
+        scene.map.layersConfiguration.getGeometryLayers()[0].showOutline = newValue;
```
gchoqueux, line 114: Nice, you remove prototype trick

As previously reported :

> - Somes users can use wireframe for study alignment, occlusion, arrangement (use cartography architecture,  and debug). Many applications offer wireframe mode without using a debug mode, because they have a different purpose than debugging. 
> - Somes users can use outline, it's a classic mapping tool ([example](http://mondecarte.o.m.f.unblog.fr/files/2014/11/carte-du-monde-9.jpg))

These properties are not only for debug and can be encapsulated in the core.

package.json
autra: dat.gui is used in GuiTools that does not use webpack (as it is an example) and in examples/index.html that also import it directly. What do you want exactly?

gchoqueux: I was asking the question (like the dat.gui is added package) if you could use in `GuiTools`
```js
import dat from 'dat.gui' //  in GuiTools
```

autra: And that what I answered to.

> dat.gui is used in GuiTools that does not use webpack

GuiTools is not processed. Therefore, pure javascript only. So no you can't.

utils/debug/Debug.js
autra: I've read this comment.

I'm not here to discuss the relevance of having it in the API, I'm just saying it does not belong to this PR. Why are you trying to force that on me *now*? Especially considering that

1. we clearly disagree on this matter
1. Nothing in this PR prevents these additions for later work

So there is clearly a debate, and when there is such a debate, the right thing to do is to move it on its own issue/PR. It should not prevent this one to get merged.

Anyway, because there is no consensus, and because it's not the purpose of this PR, I won't do it here, sorry.

gchoqueux: > - Somes users can use wireframe for study alignment, occlusion, arrangement (use cartography architecture,  and debug). Many applications offer wireframe mode without using a debug mode, because they have a different purpose than debugging. 
> - Somes users can use outline, it's a classic mapping tool ([example](http://mondecarte.o.m.f.unblog.fr/files/2014/11/carte-du-monde-9.jpg))

> These properties are not only for debug and can be encapsulated in the core.

I don't understand what disturbs on these points?

autra: > I don't understand what disturbs on these points?

It does not matter here. What matters is 

1. it is unrelated to the current PR
1. there is no consensus yet.

So : let's move the discussion elsewhere (and I'll give my reasons there) and move on here.

autra: Btw you didn't answer my question.

gchoqueux: Wireframe and outline/grid are properties of appearance that can be used as cartographic tools for users, notions of style / semantics and debug. This PR relates these properties in 2 commits. I don't find in the comments contrary arguments on the field of action of these properties. So there is an interest to factorize these functions in this PR or to splitter this PR and create a new one with the 2 commits concerned.

**Tools that seem to mainly concern the debug** :
* Charts with count nodes
* Console output when selecting tile (not unified with this PR and __DEBUG__ )
* Current camera helper and global view (not unified with this PR and __DEBUG__ )
* BBox helper (not unified with this PR and __DEBUG__ )
* Animation debug message (not unified with this PR and __DEBUG__ )

There's also other unanswered comment :
[uniform int debug](https://github.com/iTowns/itowns2/pull/238#issuecomment-280332466)

gchoqueux (approved):

gchoqueux:
Ok, we will later make another PR for wireframe and grid
Need rebase

autra:
@gchoqueux thanks! PR rebased.

PR #241, opened 2017-01-27
Prepare publishing alpha to npm

autra:
This PR:
- adds a prepublish script to build dist/ before publishing
- adds entries to .npmignore to have a clean package for client apps.

The idea is to publish this 0.0.0-alpha release on npmjs, then create a client repo `itowns-geoportail` that will use this release, in order for the Geoportail team (with @vcoindet probably) to experiment their stuff with free hands :-)

It would also be our first go as a real framework in the npm world. 

@tbroyer I'm especially interested in your feedbacks
pinging @iTowns/reviewers too.


.npmignore
```diff
@@ -1 +1,14 @@
+autodeploy.sh
+.babelrc
+data/
```
autra: this one is especially important I guess.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "itowns2",
-  "version": "1.0.0",
+  "name": "itowns",
+  "version": "0.0.0-alpha",
   "description": "A JS/WebGL framework for 3D geospatial data visualization",
   "main": "dist/itowns2.js",
```
autra: Should I rename this js too?

package.json
```diff
@@ -9,7 +9,8 @@
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
```
autra, line 11: @tbroyer I found out this `build` is not executed on `npm pack`, hence the `prepublish` below.

autra:
Referencing https://github.com/iTowns/itowns2/issues/122 here.

Btw @tbroyer should we move our dependencies to devDependencies then? As we only publish aggregated, babelized js for now, I guess we don't need them right?

.npmignore
tbroyer: It would be easier to use a whitelist in `package.json`'s `files` than maintain this blacklist here; WDYT?

package.json
tbroyer: Yes, I couldn't find any clear documentation of what happens when within NPM lifecycles, and was mislead by some documentation.

package.json
```diff
@@ -9,7 +9,8 @@
...
-    "start": "webpack-dev-server -d --inline --hot"
+    "start": "webpack-dev-server -d --inline --hot",
+    "prepublish": "webpack -p"
```
tbroyer: Should probably be `npm run build` here.

tbroyer (commented):
> Btw @tbroyer should we move our dependencies to devDependencies then? As we only publish aggregated, babelized js for now, I guess we don't need them right?

Yes, we should (and then it starts to become unmaintainable).

But I think we should maybe instead try to have a webpack build that keeps our dependencies "external", such that code using iTowns *and* same dependencies as ours (ThreeJS particularly, as it's pretty big) wouldn't have them duplicated. But again, I'm not familiar enough with the code to know whether it'd be a problem in practice (is ThreeJS completely an "implementaion detail" of iTowns or is it exposed in the API?)

Even better would be to run Babel and deploy that, and keep webpack only for an "all included" script that you can download –e.g. from GitHub releases, or "archived" in our gh-pages– and drop in your server (for cases where you don't use NPM).

For an early alpha, publishing the webpack build would be OK I guess.

autra:
> Yes, we should (and then it starts to become unmaintainable).

What do you mean by unmaintainable? 

> But I think we should maybe instead try to have a webpack build that keeps our dependencies "external", such that code using iTowns and same dependencies as ours (ThreeJS particularly, as it's pretty big) wouldn't have them duplicated.

What do you mean by "external"? In another bundle? Like that, people could add them or not in their index.html or bundling system? Not sure if this would work in practice. Would our bundle find stuff attached to window? I need to check the generated code.

> But again, I'm not familiar enough with the code to know whether it'd be a problem in practice (is ThreeJS completely an "implementaion detail" of iTowns or is it exposed in the API?)

It's exposed in the API through properties. I'm not sure what you're thinking about. What I can tell is that client code would probably need to use the exact same version of three.js as we do.

So I'm:
- moving everything to devDependencies
- making a whitelist in package.json

autra:
And if/when this gets merged, I'm gonna open an issue to generate ES5 source files and move back devDependencies into dependencies. sounds good to you @tbroyer?

package.json
autra: I did rename it.

.eslintrc.js
```diff
@@ -43,6 +43,8 @@ module.exports = {
       }
     }],
     'one-var': ['error', 'never'],
+    // TODO reactivate this rule once a proper npm package is made
+    'import/no-extraneous-dependencies': 'off',
```
autra, line 47: Reason: we import devDependencies into src/ :-/

tbroyer:
> > Yes, we should (and then it starts to become unmaintainable).
> 
> What do you mean by unmaintainable?

That we're mixing dependencies used by the code and dependencies used to build the code (well, this is already an "issue" with NPM anyway as you'd put all test-related dependencies into the same `devDependencies` bucket too…)

> > But I think we should maybe instead try to have a webpack build that keeps our dependencies "external", such that code using iTowns and same dependencies as ours (ThreeJS particularly, as it's pretty big) wouldn't have them duplicated.
> 
> What do you mean by "external"? In another bundle? Like that, people could add them or not in their index.html or bundling system? Not sure if this would work in practice. Would our bundle find stuff attached to window? I need to check the generated code.

I meant that we'd want to build a "batteries included" bundle that one could just reference from a CDN (e.g. GitHub Pages) and/or download and drop on a server, and use without additional dependencies; **and** (for NPM use) another bundle of our code only, with all external dependencies still resolved from NPM, for those who'll use iTowns as an NPM dependency and will use their own Webpack/Browserify/Rollup setup to create a bundle of their application.  
…either a bundle of our code, or just our code after Babel processing (without bundling, but then GLSL files need to be loaded differently, because we couldn't then rely on webpack loaders).

BTW, by "external", I specifically meant: https://webpack.github.io/docs/configuration.html#externals (so, one UMD bundle with everything included like we do today, and one CommonJS bundle with only our code, and everything else external; note that this means using 2 distinct `webpack.config.js`)

> What I can tell is that client code would probably need to use the exact same version of three.js as we do.

It might be a good idea to "fix" the version of Three in the `package.json` then (remove the `^` form the version specifier)

> So I'm:
> 
>  * moving everything to devDependencies
>  * making a whitelist in package.json

SGTM

autra:
Ok thanks for your explanations. I tried to capture these in https://github.com/iTowns/itowns2/issues/243 (the glsl issue is not explicitly mentioned, but we're gonna bump into it sooner or later anyway).

So we already have the "battery included" build, #243 is about the other one. 


  > > What I can tell is that client code would probably need to use the exact same version of three.js as we do.
>
> It might be a good idea to "fix" the version of Three in the package.json then (remove the ^ form the version specifier)

Probably, but I'm still not sure. And I guess @gchoqueux would know this better than me... So I will let you change it if you think it's necessary Gerald ?

So I guess this PR is ready then?

autra:
Actually wait, I still think my change in `src/Main.js` is bizarre.

autra:
ok actually I think it's the best way to expose itowns.viewer (or use `expose-loader` plugin, but I guess it's overkill here).

tbroyer (approved):

Issue #236, opened 2017-01-23
Elevation texture level problem


gchoqueux:
It would appear that since PR #202 , the maximum level for the texture elevation is 3.

gchoqueux:
To fix the bug [here](https://github.com/iTowns/itowns2/blob/master/src/Renderer/LayeredMaterial.js#L385), (fix in future PR _remove ancestor_)
```js
return tex.level < this.level;
```
to
```js
return tex.level < level;
```

PR #244, opened 2017-01-31
Remove ancestor

gchoqueux:
### Refactor
**Remove node's ancestor**
The ancestor gave the textures downscaled WMTS and WMS. Now, ancestor is replaced by direct parent. This replacement is effective for color and elevation.

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -243,6 +243,10 @@ ApiGlobe.prototype.removeImageryLayer = function removeImageryLayer(id) {
  */
 
 ApiGlobe.prototype.addElevationLayer = function addElevationLayer(layer) {
+    if (layer.protocol === 'wmts' && layer.options.tileMatrixSet !== 'WGS84G') {
+        throw new Error('Only WGS84G tileMatrixSet is currently supported for WMTS elevation layers');
+    }
+
     preprocessLayer(layer, this.scene.scheduler.getProtocolProvider(layer.protocol));
     const map = this.scene.getMap();
     if (this.getLayerById(layer.id)) {
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -147,13 +147,15 @@ ApiGlobe.prototype.addGeometryLayer = function addGeometryLayer(layer) {
 
 ApiGlobe.prototype.addImageryLayer = function addImageryLayer(layer) {
     preprocessLayer(layer, this.scene.scheduler.getProtocolProvider(layer.protocol));
-    const map = this.scene.getMap();
     if (this.getLayerById(layer.id)) {
-      // eslint-disable-next-line no-console
+        // eslint-disable-next-line no-console
         console.error(`Error : id "${layer.id}" already exist, WARNING your layer isn't added`);
     } else {
-        map.layersConfiguration.addColorLayer(layer);
-        this.viewerDiv.dispatchEvent(eventLayerAdded);
+        this.scene.getMap().layersConfiguration.addColorLayer(layer);
+        this.scene.notifyChange(1, true);
+        this.setSceneLoaded().then(() => {
+            this.viewerDiv.dispatchEvent(eventLayerAdded);
+        });
     }
 };
 
@@ -248,13 +250,15 @@ ApiGlobe.prototype.addElevationLayer = function addElevationLayer(layer) {
     }
 
     preprocessLayer(layer, this.scene.scheduler.getProtocolProvider(layer.protocol));
-    const map = this.scene.getMap();
     if (this.getLayerById(layer.id)) {
       // eslint-disable-next-line no-console
         console.error(`Error : id "${layer.id}" already exist, WARNING your layer isn't added`);
     } else {
-        map.layersConfiguration.addElevationLayer(layer);
-        this.viewerDiv.dispatchEvent(eventLayerAdded);
+        this.scene.getMap().layersConfiguration.addElevationLayer(layer);
+        this.scene.notifyChange(1, true);
+        this.setSceneLoaded().then(() => {
+            this.viewerDiv.dispatchEvent(eventLayerAdded);
+        });
     }
 };
 
--- a/src/Core/Commander/Providers/OGCWebServiceHelper.js
+++ b/src/Core/Commander/Providers/OGCWebServiceHelper.js
@@ -1,0 +1,84 @@
+import * as THREE from 'three';
+import Fetcher from './Fetcher';
+import CacheRessource from './CacheRessource';
+import IoDriver_XBIL from './IoDriver_XBIL';
+import Projection from '../../Geographic/Projection';
+
+
+export const SIZE_TEXTURE_TILE = 256;
+
+// CacheRessource is necessary for neighboring PM textures
+// The PM textures overlap several tiles WGS84, it is to avoid net requests
+// Info : THREE.js have cache image https://github.com/mrdoob/three.js/blob/master/src/loaders/ImageLoader.js#L25
+const cache = CacheRessource();
+const ioDXBIL = new IoDriver_XBIL();
+const projection = new Projection();
+
+const getTextureFloat = function getTextureFloat(buffer) {
+    const texture = new THREE.DataTexture(buffer, SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE, THREE.AlphaFormat, THREE.FloatType);
+    texture.needsUpdate = true;
+    return texture;
+};
+
+export default {
+    ioDXBIL,
+    getColorTextureByUrl(url) {
+        const cachedTexture = cache.getRessource(url);
+
+        if (cachedTexture) {
+            return Promise.resolve(cachedTexture);
+        }
+
+        const { texture, promise } = Fetcher.texture(url);
+
+        texture.generateMipmaps = false;
+        texture.magFilter = THREE.LinearFilter;
+        texture.minFilter = THREE.LinearFilter;
+        texture.anisotropy = 16;
+
+        return promise.then(() => {
+            cache.addRessource(url, texture);
+            return texture;
+        });
+    },
+    getXBilTextureByUrl(url, pitch) {
+        const textureCache = cache.getRessource(url);
+
+        if (textureCache !== undefined) {
+            const { min, max } = ioDXBIL.computeMinMaxElevation(
+                textureCache.image.data,
+                SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
+                pitch);
+            return Promise.resolve({ pitch, textureCache, min, max });
+        }
+
+        return ioDXBIL.read(url).then((result) => {
+            // TODO  RGBA is needed for navigator with no support in texture float
+            // In RGBA elevation texture LinearFilter give some errors with nodata value.
+            // need to rewrite sample function in shader
+            result.texture = getTextureFloat(result.floatArray);
+            result.texture.generateMipmaps = false;
+            result.texture.magFilter = THREE.LinearFilter;
+            result.texture.minFilter = THREE.LinearFilter;
+            result.pitch = pitch;
+
+            cache.addRessource(url, result.texture);
+
+            return result;
+        });
+    },
+    computeTileMatrixSetCoordinates(tile, layer) {
+        // Are WMTS coordinates ready?
+        if (!tile.wmtsCoords) {
+            tile.wmtsCoords = {};
+        }
+
+        const tileMatrixSet = layer.options.tileMatrixSet || 'WGS84G';
+        if (!(tileMatrixSet in tile.wmtsCoords)) {
+            const tileCoord = projection.WGS84toWMTS(tile.bbox);
+
+            tile.wmtsCoords[tileMatrixSet] =
+                projection.getCoordWMTS_WGS84(tileCoord, tile.bbox, tileMatrixSet);
+        }
+    },
+};
--- a/src/Core/Commander/Providers/OGCWebServiceHelper.js
+++ b/src/Core/Commander/Providers/OGCWebServiceHelper.js
@@ -11,6 +11,7 @@ export const SIZE_TEXTURE_TILE = 256;
 // The PM textures overlap several tiles WGS84, it is to avoid net requests
 // Info : THREE.js have cache image https://github.com/mrdoob/three.js/blob/master/src/loaders/ImageLoader.js#L25
 const cache = CacheRessource();
+const cachePending = new Map();
 const ioDXBIL = new IoDriver_XBIL();
 const projection = new Projection();
 
@@ -37,43 +38,52 @@ export default {
         texture.anisotropy = 16;
 
         return promise.then(() => {
-            cache.addRessource(url, texture);
+            if (!cache.getRessource(url)) {
+                cache.addRessource(url, texture);
+            }
             return texture;
         });
     },
-    getXBilTextureByUrl(url, pitch) {
+    getXBilTextureByUrl(url) {
         const textureCache = cache.getRessource(url);
 
         if (textureCache !== undefined) {
-            const { min, max } = ioDXBIL.computeMinMaxElevation(
-                textureCache.image.data,
-                SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
-                pitch);
-            return Promise.resolve({ pitch, textureCache, min, max });
+            return Promise.resolve({ texture: textureCache });
         }
 
-        return ioDXBIL.read(url).then((result) => {
+        const promiseXBil = (cachePending.get(url) || ioDXBIL.read(url)).then((result) => {
             // TODO  RGBA is needed for navigator with no support in texture float
             // In RGBA elevation texture LinearFilter give some errors with nodata value.
             // need to rewrite sample function in shader
+
+            // loading concurrence
+            const textureConcurrence = cache.getRessource(url);
+            if (textureConcurrence) {
+                cachePending.delete(url);
+                return textureConcurrence;
+            }
+
             result.texture = getTextureFloat(result.floatArray);
             result.texture.generateMipmaps = false;
             result.texture.magFilter = THREE.LinearFilter;
             result.texture.minFilter = THREE.LinearFilter;
-            result.pitch = pitch;
-
             cache.addRessource(url, result.texture);
+            cachePending.delete(url);
 
             return result;
         });
+
+        cachePending.set(url, promiseXBil);
+
+        return promiseXBil;
     },
-    computeTileMatrixSetCoordinates(tile, layer) {
+    computeTileMatrixSetCoordinates(tile, tileMatrixSet) {
         // Are WMTS coordinates ready?
         if (!tile.wmtsCoords) {
             tile.wmtsCoords = {};
         }
 
-        const tileMatrixSet = layer.options.tileMatrixSet || 'WGS84G';
+        tileMatrixSet = tileMatrixSet || 'WGS84G';
         if (!(tileMatrixSet in tile.wmtsCoords)) {
             const tileCoord = projection.WGS84toWMTS(tile.bbox);
 
--- a/src/Core/Commander/Providers/TileProvider.js
+++ b/src/Core/Commander/Providers/TileProvider.js
@@ -17,6 +17,8 @@ import Provider from './Provider';
 import Projection from '../../Geographic/Projection';
 import BuilderEllipsoidTile from '../../../Globe/BuilderEllipsoidTile';
 import TileGeometry from '../../../Globe/TileGeometry';
+import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from './OGCWebServiceHelper';
+import { EMPTY_TEXTURE_ZOOM, l_ELEVATION } from '../../../Renderer/LayeredMaterial';
 
 function TileProvider() {
     Provider.call(this, null);
@@ -41,6 +43,8 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
         bbox,
         level: (command.level === undefined) ? (parent.level + 1) : command.level,
         segment: 16,
+        parentMaterial: parent.material,
+        parentWmtsCoords: parent.wmtsCoords,
     };
 
     const geometry = new TileGeometry(params, this.builder);
@@ -61,6 +65,19 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
     tile.updateMatrixWorld();
     tile.OBB().parent = tile;   // TODO: we should use tile.add(tile.OBB())
     tile.OBB().update();
+
+    // update bbox if node herits texture elevation from parent
+    if (tile.material.getElevationLayerLevel() > EMPTY_TEXTURE_ZOOM) {
+        const textureElevation = tile.material.getLayerTextures(l_ELEVATION)[0];
+        const { min, max } = OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
+                textureElevation.image.data,
+                SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
+                tile.material.offsetScale[0][0]);
+
+        if (min && max) {
+            tile.setBBoxZ(min, max);
+        }
+    }
 
     return Promise.resolve(tile);
 };
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -28,7 +28,7 @@ function WMS_Provider(/* options*/) {
 
     this.getTextureFloat = function getTextureFloat(buffer) {
         // Start float to RGBA uint8
-        var texture = new THREE.DataTexture(buffer, 256, 256, THREE.AlphaFormat, THREE.FloatType);
+        const texture = new THREE.DataTexture(buffer, 256, 256, THREE.AlphaFormat, THREE.FloatType);
 
         texture.needsUpdate = true;
         return texture;
@@ -73,6 +73,9 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
+    layer.bbox = layer.bbox || new BoundingBox();
+    layer.options = {};
+    layer.options.tileMatrixSet = layer.tileMatrixSet || 'WGS84G';
 
     layer.customUrl = `${layer.url
                   }?SERVICE=WMS&REQUEST=GetMap&LAYERS=${layer.name
@@ -90,14 +93,29 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
     return tile.level > 2 && layer.bbox.intersect(tile.bbox);
 };
 
+function computeTileWMTSCoordinates(tile, wmtsLayer, projection) {
+    // Are WMTS coordinates ready?
+    if (!tile.wmtsCoords) {
+        tile.wmtsCoords = {};
+    }
+
+    const tileMatrixSet = wmtsLayer.options.tileMatrixSet;
+    if (!(tileMatrixSet in tile.wmtsCoords)) {
+        const tileCoord = projection.WGS84toWMTS(tile.bbox);
+
+        tile.wmtsCoords[tileMatrixSet] =
+            projection.getCoordWMTS_WGS84(tileCoord, tile.bbox, tileMatrixSet);
+    }
+}
+
 WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, bbox, pitch) {
     if (!this.tileInsideLimit(tile, layer) || tile.material === null) {
         return Promise.resolve();
     }
 
-    var url = this.url(bbox.as('EPSG:4326'), layer);
+    const url = this.url(bbox.as('EPSG:4326'), layer);
+    const result = { pitch };
 
-    var result = { pitch };
     result.texture = this.cache.getRessource(url);
 
     if (result.texture !== undefined) {
@@ -111,6 +129,8 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, b
     result.texture.magFilter = THREE.LinearFilter;
     result.texture.minFilter = THREE.LinearFilter;
     result.texture.anisotropy = 16;
+    result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
+    result.texture.bbox = bbox;
 
     return promise.then(() => {
         this.cache.addRessource(url, result.texture);
@@ -120,10 +140,10 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, b
 };
 
 WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbox, pitch) {
-    var url = this.url(bbox, layer);
+    const url = this.url(bbox, layer);
 
     // TODO: this is not optimal: if called again before the IoDriver resolves, it'll load the XBIL again
-    var textureCache = this.cache.getRessource(url);
+    const textureCache = this.cache.getRessource(url);
 
     if (textureCache !== undefined) {
         return Promise.resolve({
@@ -134,19 +154,12 @@ WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbo
         });
     }
 
-    // bug #74
-    // var limits = layer.tileMatrixSetLimits[coWMTS.zoom];
-    // if (!limits || !coWMTS.isInside(limits)) {
-    //     var texture = -1;
-    //     this.cache.addRessource(url, texture);
-    //     return Promise.resolve(texture);
-    // }
-    // -> bug #74
     return this._IoDriver.read(url).then((result) => {
         result.texture = this.getTextureFloat(result.floatArray);
         result.texture.generateMipmaps = false;
         result.texture.magFilter = THREE.LinearFilter;
         result.texture.minFilter = THREE.LinearFilter;
+        result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
         result.pitch = pitch;
 
         // In RGBA elevation texture LinearFilter give some errors with nodata value.
@@ -158,62 +171,37 @@ WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbo
 };
 
 WMS_Provider.prototype.executeCommand = function executeCommand(command) {
-    var layer = command.layer;
-    var tile = command.requester;
-    var ancestor = command.ancestor;
+    const layer = command.layer;
+    const tile = command.requester;
 
-    var supportedFormats = {
+    computeTileWMTSCoordinates(tile, layer, this.projection);
+
+    const supportedFormats = {
         'image/png': this.getColorTexture.bind(this),
         'image/jpg': this.getColorTexture.bind(this),
         'image/jpeg': this.getColorTexture.bind(this),
         'image/x-bil;bits=32': this.getXbilTexture.bind(this),
     };
 
-    var func = supportedFormats[layer.format];
+    const func = supportedFormats[layer.format];
     if (func) {
-        var pitch = ancestor ?
-            this.projection.WMS_WGS84Parent(tile.bbox, ancestor.bbox) :
-            new THREE.Vector3(0, 0, 1);
+        const searchInParent = tile.materials[0].getColorLayerLevelById(layer.id) < 0 && tile.parent.materials[0].getColorLayerLevelById(layer.id) > -1;
+        let pitch = new THREE.Vector3(0, 0, 1);
+        let bbox = tile.bbox;
 
-        var bbox = ancestor ?
-            ancestor.bbox :
-            tile.bbox;
-
+        if (searchInParent) {
+            const texture = tile.parent.material.getLayerTextures(layer.type, layer.id)[0];
+            if (texture) {
+                bbox = texture.bbox;
+                pitch = this.projection.WMS_WGS84Parent(tile.bbox, bbox);
+                return Promise.resolve({ pitch, texture });
+            }
+        }
 
         return func(tile, layer, bbox, pitch);
     } else {
         return Promise.reject(new Error(`Unsupported mimetype ${layer.format}`));
     }
-};
-
-
-/**
- * Returns a texture from the WMS stream with the specified bounding box
- * @param {BoundingBox} bbox: requested bounding box
- * @returns {WMS_Provider_L15.WMS_Provider.prototype@pro;_IoDriver@call;read@call;then}
- */
-WMS_Provider.prototype.getTexture = function getTexture(bbox) {
-    if (bbox === undefined)
-        { return Promise.resolve(-2); }
-
-    var url = this.url(bbox);
-
-    // TODO: this is not optimal: if called again before ioDriverImage resolves, it'll load the image again
-    var textureCache = this.cache.getRessource(url);
-
-    if (textureCache !== undefined) {
-        return Promise.resolve(textureCache);
-    }
-    return this.ioDriverImage.read(url).then((image) => {
-        var result = {};
-        result.texture = new THREE.Texture(image);
-        result.texture.generateMipmaps = false;
-        result.texture.magFilter = THREE.LinearFilter;
-        result.texture.minFilter = THREE.LinearFilter;
-        result.texture.anisotropy = 16;
-        this.cache.addRessource(url, result.texture);
-        return result.texture;
-    });
 };
 
 export default WMS_Provider;
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -6,12 +6,8 @@
 
 
 import * as THREE from 'three';
-import Provider from './Provider';
-import IoDriver_XBIL from './IoDriver_XBIL';
-import Fetcher from './Fetcher';
-import Projection from '../../Geographic/Projection';
-import CacheRessource from './CacheRessource';
 import BoundingBox from '../../../Scene/BoundingBox';
+import WMTS_Provider, { computeTileWMTSCoordinates } from './WMTS_Provider';
 
 /**
  * Return url wmts MNT
@@ -20,22 +16,11 @@ import BoundingBox from '../../../Scene/BoundingBox';
  * @param {String} options.format: image format (default: format/jpeg)
  * @returns {Object@call;create.url.url|String}
  */
-function WMS_Provider(/* options*/) {
-    // Constructor
-    Provider.call(this, new IoDriver_XBIL());
-    this.cache = CacheRessource();
-    this.projection = new Projection();
-
-    this.getTextureFloat = function getTextureFloat(buffer) {
-        // Start float to RGBA uint8
-        const texture = new THREE.DataTexture(buffer, 256, 256, THREE.AlphaFormat, THREE.FloatType);
-
-        texture.needsUpdate = true;
-        return texture;
-    };
+function WMS_Provider(options) {
+    WMTS_Provider.call(this, options);
 }
 
-WMS_Provider.prototype = Object.create(Provider.prototype);
+WMS_Provider.prototype = Object.create(WMTS_Provider.prototype);
 
 WMS_Provider.prototype.constructor = WMS_Provider;
 
@@ -93,21 +78,6 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
     return tile.level > 2 && layer.bbox.intersect(tile.bbox);
 };
 
-function computeTileWMTSCoordinates(tile, wmtsLayer, projection) {
-    // Are WMTS coordinates ready?
-    if (!tile.wmtsCoords) {
-        tile.wmtsCoords = {};
-    }
-
-    const tileMatrixSet = wmtsLayer.options.tileMatrixSet;
-    if (!(tileMatrixSet in tile.wmtsCoords)) {
-        const tileCoord = projection.WGS84toWMTS(tile.bbox);
-
-        tile.wmtsCoords[tileMatrixSet] =
-            projection.getCoordWMTS_WGS84(tileCoord, tile.bbox, tileMatrixSet);
-    }
-}
-
 WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, bbox, pitch) {
     if (!this.tileInsideLimit(tile, layer) || tile.material === null) {
         return Promise.resolve();
@@ -116,56 +86,18 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, b
     const url = this.url(bbox.as('EPSG:4326'), layer);
     const result = { pitch };
 
-    result.texture = this.cache.getRessource(url);
-
-    if (result.texture !== undefined) {
-        return Promise.resolve(result);
-    }
-
-    const { texture, promise } = Fetcher.texture(url);
-    result.texture = texture;
-
-    result.texture.generateMipmaps = false;
-    result.texture.magFilter = THREE.LinearFilter;
-    result.texture.minFilter = THREE.LinearFilter;
-    result.texture.anisotropy = 16;
-    result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
-    result.texture.bbox = bbox;
-
-    return promise.then(() => {
-        this.cache.addRessource(url, result.texture);
-        result.texture.needsUpdate = true;
+    return this.getColorTextureByUrl(url).then((texture) => {
+        result.texture = texture;
+        result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
+        result.texture.bbox = bbox;
         return result;
     });
 };
 
 WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbox, pitch) {
     const url = this.url(bbox, layer);
-
-    // TODO: this is not optimal: if called again before the IoDriver resolves, it'll load the XBIL again
-    const textureCache = this.cache.getRessource(url);
-
-    if (textureCache !== undefined) {
-        return Promise.resolve({
-            pitch,
-            texture: textureCache.texture,
-            min: textureCache.min,
-            max: textureCache.max,
-        });
-    }
-
-    return this._IoDriver.read(url).then((result) => {
-        result.texture = this.getTextureFloat(result.floatArray);
-        result.texture.generateMipmaps = false;
-        result.texture.magFilter = THREE.LinearFilter;
-        result.texture.minFilter = THREE.LinearFilter;
+    return this.getXBilTextureByUrl(url, pitch).then((result) => {
         result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
-        result.pitch = pitch;
-
-        // In RGBA elevation texture LinearFilter give some errors with nodata value.
-        // need to rewrite sample function in shader
-        this.cache.addRessource(url, result);
-
         return result;
     });
 };
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -78,33 +78,42 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
     return tile.level > 2 && layer.bbox.intersect(tile.bbox);
 };
 
-WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, bbox, pitch) {
+WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer) {
     if (!this.tileInsideLimit(tile, layer) || tile.material === null) {
         return Promise.resolve();
     }
 
-    const url = this.url(bbox.as('EPSG:4326'), layer);
+    const url = this.url(tile.bbox.as('EPSG:4326'), layer);
+    const pitch = new THREE.Vector3(0, 0, 1);
     const result = { pitch };
 
     return this.getColorTextureByUrl(url).then((texture) => {
         result.texture = texture;
         result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
-        result.texture.bbox = bbox;
+        result.texture.bbox = tile.bbox;
         return result;
     });
 };
 
-WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, bbox, pitch) {
-    const url = this.url(bbox, layer);
-    return this.getXBilTextureByUrl(url, pitch).then((result) => {
+WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer) {
+    const url = this.url(tile.bbox, layer);
+    return this.getXBilTextureByUrl(url, new THREE.Vector3(0, 0, 1)).then((result) => {
         result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
         return result;
     });
 };
 
 WMS_Provider.prototype.executeCommand = function executeCommand(command) {
-    const layer = command.layer;
+    const parentTextures = command.parentTextures;
     const tile = command.requester;
+
+    if (parentTextures) {
+        const texture = parentTextures[0];
+        const pitch = this.projection.WMS_WGS84Parent(tile.bbox, texture.bbox);
+        return Promise.resolve({ pitch, texture });
+    }
+
+    const layer = command.layer;
 
     computeTileWMTSCoordinates(tile, layer, this.projection);
 
@@ -116,24 +125,13 @@ WMS_Provider.prototype.executeCommand = function executeCommand(command) {
     };
 
     const func = supportedFormats[layer.format];
+
     if (func) {
-        const searchInParent = tile.materials[0].getColorLayerLevelById(layer.id) < 0 && tile.parent.materials[0].getColorLayerLevelById(layer.id) > -1;
-        let pitch = new THREE.Vector3(0, 0, 1);
-        let bbox = tile.bbox;
-
-        if (searchInParent) {
-            const texture = tile.parent.material.getLayerTextures(layer.type, layer.id)[0];
-            if (texture) {
-                bbox = texture.bbox;
-                pitch = this.projection.WMS_WGS84Parent(tile.bbox, bbox);
-                return Promise.resolve({ pitch, texture });
-            }
-        }
-
-        return func(tile, layer, bbox, pitch);
+        return func(tile, layer);
     } else {
         return Promise.reject(new Error(`Unsupported mimetype ${layer.format}`));
     }
 };
+
 
 export default WMS_Provider;
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -8,24 +8,6 @@
 import * as THREE from 'three';
 import BoundingBox from '../../../Scene/BoundingBox';
 import OGCWebServiceHelper from './OGCWebServiceHelper';
-import { UNIT } from '../../Geographic/Coordinates';
-
-const WMS_WGS84Parent = function WMS_WGS84Parent(bbox, bboxParent) {
-    const dim = bbox.dimensions(UNIT.RADIAN);
-    const dimParent = bboxParent.dimensions(UNIT.RADIAN);
-    const scale = dim.x / dimParent.x;
-
-    const x =
-        Math.abs(bbox.west(UNIT.RADIAN) - bboxParent.west(UNIT.RADIAN)) /
-        dimParent.x;
-    const y =
-        Math.abs(
-            bbox.south(UNIT.RADIAN) + dim.y -
-            (bboxParent.south(UNIT.RADIAN) + dimParent.y)) /
-        dimParent.y;
-
-    return new THREE.Vector3(x, y, scale);
-};
 
 /**
  * Return url wmts MNT
@@ -113,14 +95,7 @@ WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer) {
 };
 
 WMS_Provider.prototype.executeCommand = function executeCommand(command) {
-    const parentTextures = command.parentTextures;
     const tile = command.requester;
-
-    if (parentTextures) {
-        const texture = parentTextures[0];
-        const pitch = WMS_WGS84Parent(tile.bbox, texture.bbox);
-        return Promise.resolve({ pitch, texture });
-    }
 
     const layer = command.layer;
     const supportedFormats = {
--- a/src/Core/Commander/Providers/WMS_Provider.js
+++ b/src/Core/Commander/Providers/WMS_Provider.js
@@ -7,7 +7,25 @@
 
 import * as THREE from 'three';
 import BoundingBox from '../../../Scene/BoundingBox';
-import WMTS_Provider, { computeTileWMTSCoordinates } from './WMTS_Provider';
+import OGCWebServiceHelper from './OGCWebServiceHelper';
+import { UNIT } from '../../Geographic/Coordinates';
+
+const WMS_WGS84Parent = function WMS_WGS84Parent(bbox, bboxParent) {
+    const dim = bbox.dimensions(UNIT.RADIAN);
+    const dimParent = bboxParent.dimensions(UNIT.RADIAN);
+    const scale = dim.x / dimParent.x;
+
+    const x =
+        Math.abs(bbox.west(UNIT.RADIAN) - bboxParent.west(UNIT.RADIAN)) /
+        dimParent.x;
+    const y =
+        Math.abs(
+            bbox.south(UNIT.RADIAN) + dim.y -
+            (bboxParent.south(UNIT.RADIAN) + dimParent.y)) /
+        dimParent.y;
+
+    return new THREE.Vector3(x, y, scale);
+};
 
 /**
  * Return url wmts MNT
@@ -16,13 +34,8 @@ import WMTS_Provider, { computeTileWMTSCoordinates } from './WMTS_Provider';
  * @param {String} options.format: image format (default: format/jpeg)
  * @returns {Object@call;create.url.url|String}
  */
-function WMS_Provider(options) {
-    WMTS_Provider.call(this, options);
+function WMS_Provider() {
 }
-
-WMS_Provider.prototype = Object.create(WMTS_Provider.prototype);
-
-WMS_Provider.prototype.constructor = WMS_Provider;
 
 WMS_Provider.prototype.url = function url(bbox, layer) {
     const box = bbox.as(layer.projection);
@@ -46,6 +59,9 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     if (!layer.bbox) {
         throw new Error('bbox is required');
     }
+    if (!layer.projection) {
+        throw new Error('projection is required');
+    }
 
     layer.bbox = new BoundingBox(
         layer.projection,
@@ -58,9 +74,6 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
-    layer.bbox = layer.bbox || new BoundingBox();
-    layer.options = {};
-    layer.options.tileMatrixSet = layer.tileMatrixSet || 'WGS84G';
 
     layer.customUrl = `${layer.url
                   }?SERVICE=WMS&REQUEST=GetMap&LAYERS=${layer.name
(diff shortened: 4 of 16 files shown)


src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -109,22 +109,23 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
+        if (parentZoom > -1) {
+            coordWMTS = this.projection.WMTS_WGS84Parent(
+                coordWMTS,
+                Math.max(layer.zoom.min, parentZoom),
+                pitch);
+        }
     }
 
-    var url = this.url(cooWMTS, layer);
+    var url = this.url(coordWMTS, layer);
```
peppsac: When `searchInParent` is true you could skip the `this.url()` call and directly use parent's texture.

If this is done on all providers, we could remove their usage of `CacheResources` which IMHO is a good idea:
* the browser already cache network request
* we never evict data from this cache

This would allow to effectively delete unused THREE.Texture which is currently not done (even after calling material.dispose() because CacheResource still own a reference).

gchoqueux: Yes, this is the direction I want to take. 
I can do the job now?

`material.dispose()` frees VRAM and not RAM (CacheResource have reference in RAM)

peppsac: Yes, it would be nice if you have the time to implement it.
Also you may want to squash the 3 `remove ancestor in loading` commits.

gchoqueux: CacheRessource is necessary for neighboring PM textures

gchoqueux:
updated

index.html
```diff
@@ -53,9 +53,9 @@
             const urlElevationLayers = [];
 
             urlImageryLayers.push('examples/layers/JSONLayers/Ortho.json');
-            urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
-            urlImageryLayers.push('examples/layers/JSONLayers/ScanEX.json');
-            urlImageryLayers.push('examples/layers/JSONLayers/Region.json');
+            // urlImageryLayers.push('examples/layers/JSONLayers/OrthosCRS.json');
```
peppsac: Why did you remove those?
Also your commit title misses a "commit_type:" prefix

src/Core/Commander/Providers/TileProvider.js
```diff
@@ -44,6 +47,7 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
         bbox,
         level: (command.level === undefined) ? (parent.level + 1) : command.level,
         segment: 16,
+        light: command.layer.light,
```
peppsac: I don't think `TileProvider` needs to be aware of lighting.
Maybe you could move lights handling in NodeProcess?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -22,22 +18,11 @@ import { UNIT } from 'Core/Geographic/GeoCoordinate';
...
-
-    this.getTextureFloat = function getTextureFloat(buffer) {
-        // Start float to RGBA uint8
-        var texture = new THREE.DataTexture(buffer, 256, 256, THREE.AlphaFormat, THREE.FloatType);
-
-        texture.needsUpdate = true;
-        return texture;
-    };
+function WMS_Provider(options) {
+    WMTS_Provider.call(this, options);
```
peppsac: Instead of inherinting from `WMTS_Provider`, maybe we could build a helper class for all those protocols (WFS, WMS, WMTS)?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -73,6 +58,7 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
     layer.bbox = layer.bbox || new BoundingBox();
+    layer.options = { tileMatrixSet: 'WGS84G' };
```
peppsac: Why this change?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -95,125 +81,57 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, b
...
     if (func) {
-        var pitch = ancestor ?
-            this.projection.WMS_WGS84Parent(tile.bbox, ancestor.bbox) :
-            new THREE.Vector3(0, 0, 1);
-
-        var bbox = ancestor ?
-            ancestor.bbox :
-            tile.bbox;
-
+        const searchInParent = tile.materials[0].getColorLayerLevelById(layer.id) < 0 && tile.parent.materials[0].getColorLayerLevelById(layer.id) > -1;
```
peppsac: No, you can't depend on `tile.materials[0].getColorLayerLevelById(layer.id)` to do this.
Actually you can't depend on `LayeredMaterial` at all.
Add the needed params to the `command` instead (for instance the parent's texture for this layer).

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -104,62 +114,69 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
-            cooWMTS,
-            this.computeLevelToDownload(tile, parameters.ancestor, layer) + levelOffset,
-            pitch);
+WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer) {
+    let coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
+    const pitch = new THREE.Vector3(0.0, 0.0, 1.0);
+    const searchInParent = tile.material.getElevationLayerLevel() < 0;
+
+    if (searchInParent || coordWMTS.zoom > layer.zoom.max) {
+        const parentZoom = tile.parent.material.getElevationLayerLevel();
```
peppsac: idem: I'd rather not introduce a dependency to `LayeredMaterial` here.

src/Globe/Atmosphere.js
```diff
@@ -96,95 +93,28 @@ function Atmosphere(ellipsoid) {
...
-        },
-        tSkyboxDiffuse: {
-            type: 't',
-            value: new THREE.Texture(),
-        },
-        fNightScale: {
-            type: 'f',
-            value: 1.0,
-        },
+        v3LightPosition: { value: LIGHTING_POSITION.clone().normalize() },
```
peppsac: Nice change, but it would have been easier to review if you had split all the code cleanup in a separate PR.

src/Renderer/LayeredMaterial.js
```diff
@@ -420,4 +419,20 @@ LayeredMaterial.prototype.getLayerLevel = function getLayerLevel(layerType, laye
...
+    }
+
+    const index = this.indexOfColorLayer(layer);
+
+    if (index !== -1) {
+        const count = this.getTextureCountByLayerIndex(index);
+        const textureIndex = this.getTextureOffsetByLayerIndex(index);
+        return this.textures[l_COLOR].slice(textureIndex, textureIndex + count);
+    } else {
+        return [];
```
peppsac:
```js
throw new Error(`Requesting textures for invalid layer : ${layer}`);
```
Also replace function param name : `layer` -> `layerId`

src/Renderer/c3DEngine.js
```diff
@@ -28,6 +28,7 @@ function c3DEngine(scene, positionCamera, viewerDiv, debugMode, gLDebug) {
     this.viewerDiv = viewerDiv;
     this.debug = debugMode;
     this.scene3D = new THREE.Scene();
+    this.scene3D.autoUpdate = false;
```
peppsac: This change is unrelated to the current PR so I'd prefer that you submit a separate PR for this.

src/Scene/LayersConfiguration.js
```diff
@@ -4,6 +4,7 @@
  * Description: Le layer est une couche de données. Cette couche peut etre des images ou de l'information 3D. Les requètes de cette couche sont acheminées par une interfaceCommander.
  *
  */
+import { l_COLOR, l_ELEVATION } from 'Renderer/LayeredMaterial';
```
peppsac: I'd rather not do this.
If you really need to be able to do `layer.type = l_ELEVATION;` then the `l_ELEVATION` constant probably belong to this file.
But I'd rather not have this at all - and it's unrelated to the current PR scope anyway ?

peppsac (changes requested):
The new mechanism works well, but I left some general architecture comments.

peppsac:
Btw, you could have done 4 different PRs, 1 for each "theme":

* fix (Loading texture) : error loading elevation #236 
* fix (Layer) : error in option layer for lighting
* removes layers for demo Front page
* and the 2 providers refactoring commit

They would have been easier to review and faster to merge (hint: it's not too late to do this :))

(yes my own current PR has 17 commits ranging from layer refactoring to typo fixing but... do as I say not as I do :))

index.html
gchoqueux: The external returns on the demo are:
  * Loading speed (due to the 4 layers)
  * Unnecessary layers: region (ugly), CRS (duplicate with ortho) and scan (limited to France)

src/Core/Commander/Providers/TileProvider.js
gchoqueux: completely agree

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux: yes, I also thought about a imageTiledProvider  to provide image tiled for WMTS, WMS or big texture.

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux: Because I need to know how the wms is tiled

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux: ok

src/Renderer/LayeredMaterial.js
gchoqueux: ok

index.html
peppsac: Ok (just copy-paste your comment in the commit message)

src/Core/Commander/Providers/WMS_Provider.js
peppsac: But: 
- they might not be tiled like this. For instance the Lyon opendata wms server is in meter and you can query any bbox inside its extents so the `tileMatrixSet` makes no sense in this case.
- if you add an option use the `layer.foo = layer.foo || my_default_option`

peppsac:
Looks better with the last commit. Can you split this PR in 4 PR (see comment https://github.com/iTowns/itowns2/pull/244#issuecomment-279326570) so I can review/merge ?

gchoqueux:
rebased

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -73,6 +58,9 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
+    layer.bbox = layer.bbox || new BoundingBox();
```
peppsac: Empty `BoundingBox` can't be built anymore.
IMHO a WMS layer requires a boundingbox to work correctly so we'd better throw an error if it's missing.
Of course we'll be able to improve this when `GetCapabilities` is implemented :)

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -90,130 +78,60 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
...
 
-    var url = this.url(bbox.as('EPSG:4326'), layer);
-
-    var result = { pitch };
-    result.texture = this.cache.getRessource(url);
-
-    if (result.texture !== undefined) {
-        return Promise.resolve(result);
-    }
+    const url = this.url(tile.bbox.as('EPSG:4326'), layer);
```
peppsac: `tile.bbox.as(layer.projection)`

src/Core/Commander/Providers/WMS_Provider.js
peppsac: Of course the same goes for `projection` attribute: it's mandatory.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -90,130 +78,60 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
...
-    result.texture.magFilter = THREE.LinearFilter;
-    result.texture.minFilter = THREE.LinearFilter;
-    result.texture.anisotropy = 16;
-
-    return promise.then(() => {
-        this.cache.addRessource(url, result.texture);
-        result.texture.needsUpdate = true;
+    return this.getColorTextureByUrl(url).then((texture) => {
+        result.texture = texture;
+        result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
```
peppsac: Why is this needed? WMS has nothing to do with `coordWMTS` right?

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -90,130 +78,60 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
...
-        result.pitch = pitch;
-
-        // In RGBA elevation texture LinearFilter give some errors with nodata value.
-        // need to rewrite sample function in shader
-        this.cache.addRessource(url, result);
-
+WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer) {
+    const url = this.url(tile.bbox, layer);
+    return this.getXBilTextureByUrl(url, new THREE.Vector3(0, 0, 1)).then((result) => {
+        result.texture.coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
```
peppsac: Same, `coordWMTS` is irrelevant here.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -90,130 +78,60 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
...
+    if (parentTextures) {
+        const texture = parentTextures[0];
+        const pitch = this.projection.WMS_WGS84Parent(tile.bbox, texture.bbox);
+        return Promise.resolve({ pitch, texture });
+    }
+
+    const layer = command.layer;
 
-    var supportedFormats = {
+    computeTileWMTSCoordinates(tile, layer, this.projection);
```
peppsac: Unneeded.

src/Core/Commander/Providers/WMS_Provider.js
peppsac: I'd rather not introduce this inheritance.
So I belive you need to either:
- introduce a `OGCWebServiceHelper.js` file holding helper methods (see `Fetcher.js` for a good example)
- live a few more weeks with some duplicate codes between `WMTS` and `WMS`, until someone is willing to refactor this code.

src/Scene/NodeProcess.js
```diff
@@ -398,16 +337,15 @@ function updateNodeElevation(scene, quadtree, node, layersConfig, force) {
     if (bestLayer !== null) {
         node.layerUpdateState[bestLayer.id].newTry();
 
+        const searchInParent = (!node.isElevationLayerLoaded() && node.parent.isElevationLayerLoaded()) || (node.level > bestLayer.zoom.max);
```
peppsac: Can you add a comment explaining why the test is different to the color one?

peppsac (changes requested):

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux: projection?

gchoqueux: ah ok  the CRS

src/Core/Commander/Providers/WMS_Provider.js
gchoqueux: WMS is tiled and works like WMTS.
I need to know his correspondence with tile coordinate

src/Scene/NodeProcess.js
gchoqueux: Because elevation is composed of 2 layers

gchoqueux:
I need WMTS coordinate  to search in parent texture

src/Core/Commander/Providers/WMS_Provider.js
Jeremy-Gaillard: WMS is not a tiled protocol. If you need a coordWMTS attribute, I suggest moving the assignment to NodeProcess or another place which works exclusively with tiles.

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,129 @@
...
+            texture,
+            min: minmax.min,
+            max: minmax.max,
+        });
+};
+
+const getTextureFloat = function getTextureFloat(buffer) {
+    // Start float to RGBA uint8
+    // var bufferUint = new Uint8Array(buffer.buffer);
+    // var texture = new THREE.DataTexture(bufferUint, 256, 256);
```
peppsac: Maybe remove this commented code

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -73,6 +58,8 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
+    layer.options = {};
+    layer.options.tileMatrixSet = layer.tileMatrixSet || 'WGS84G';
```
peppsac: See previous comment thread https://github.com/iTowns/itowns2/pull/244#discussion_r104377515
(TL;DR: *WMS is not a tiled protocol and has nothing to do with WMTS coords*)

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -4,45 +4,13 @@
...
-    if (this.support)
-        { this.getTextureFloat = function getTextureFloat() {
-            return new THREE.Texture();
-        }; }
-    else
-        { this.getTextureFloat = function getTextureFloat(buffer) {
-            // Start float to RGBA uint8
-            // var bufferUint = new Uint8Array(buffer.buffer);
-            // var texture = new THREE.DataTexture(bufferUint, 256, 256);
+import Tish from './TiledImageTools';
```
peppsac: `Tish` ??

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -109,147 +73,65 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
-    if (textureCache !== undefined) {
-        const minmax = this._IoDriver.computeMinMaxElevation(
-            textureCache.floatArray,
-            SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
+WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, parentTextures) {
+    let coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
+    const pitch = new THREE.Vector3(0.0, 0.0, 1.0);
+
+    if (parentTextures) {
+        coordWMTS = Tish.WMTS_WGS84Parent(
```
peppsac: This only works with `WGS84G` tileMatrixSet right? (As you assume there's only one texture)
If this is right, add throw an error when this condition is false.

src/Renderer/LayeredMaterial.js
```diff
@@ -394,26 +394,41 @@ LayeredMaterial.prototype.isLayerTypeDownscaled = function isLayerTypeDownscaled
...
+    }
+
+    const index = this.indexOfColorLayer(layerId);
+
+    if (index !== -1) {
+        const count = this.getTextureCountByLayerIndex(index);
+        const textureIndex = this.getTextureOffsetByLayerIndex(index);
+        return this.textures[l_COLOR].slice(textureIndex, textureIndex + count);
+    } else {
+        throw new Error('No known layer :', layerId);
```
peppsac:
```js
throw new Error(`Invalid layer id "${layerId}"`);
```

src/Scene/NodeProcess.js
```diff
@@ -253,48 +236,25 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
...
         }
 
-        let ancestor = null;
-
-        const currentLevel = node.materials[RendererConstant.FINAL].getColorLayerLevelById(layer.id);
-        // if this tile has no texture (level == -1), try use one from an ancestor
-        if (currentLevel === -1) {
-            ancestor = findAncestorWithValidTextureForLayer(node, l_COLOR, layer);
-        } else {
-            var targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node.level, currentLevel, layer.updateStrategy.options);
```
peppsac, line 263: You're dropping all network strategy code.
Why?

src/Scene/NodeProcess.js
```diff
@@ -253,48 +236,25 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
...
         const command = {
             /* mandatory */
             layer,
             requester: node,
             priority: nodeCommandQueuePriorityFunction(node),
             earlyDropFunction: refinementCommandCancellationFn,
-            /* specific params */
-            ancestor,
-            /* redraw only if we're aren't using a texture from our parent */
-            redraw: (ancestor == null),
```
peppsac, line 283: Why do you remove `redraw` param ?

src/Scene/NodeProcess.js
```diff
@@ -253,48 +236,25 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
...
                 if (Array.isArray(result)) {
-                    for (let j = 0; j < result.length; j++) {
-                        result[j].texture.level = level;
-                    }
-
                     node.setTexturesLayer(result, l_COLOR, layer.id);
                 } else if (result.texture) {
-                    result.texture.level = level;
+                    if (!result.texture.coordWMTS) {
+                        result.texture.coordWMTS = node.wmtsCoords[layer.options.tileMatrixSet][0];
```
peppsac: Eeek. If this is mandatory this can be handled by the provider.

src/Scene/NodeProcess.js
peppsac: I meant a comment in the *code*, not the PR.

peppsac (changes requested):

src/Scene/NodeProcess.js
gchoqueux: [except for WMS, see comment](https://github.com/iTowns/itowns2/pull/244#discussion_r104423818)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -199,6 +199,10 @@ ApiGlobe.prototype.removeImageryLayer = function removeImageryLayer(id) {
  */
 
 ApiGlobe.prototype.addElevationLayer = function addElevationLayer(layer) {
+    if (layer.protocol === 'wmts' && layer.options.tileMatrixSet !== 'WGS84G') {
+        throw new Error('Elevation layer with wmts protocol support only WGS84G projection');
```
autra: If I understand correctly, this is a current limitation of itowns (well, in its globe mode at least). To make this clearer, I suggest:

> Only WSG84 tileMatrixSet is currently supported for WMTS elevation layers

(If I'm not mistaken, WSG84 is a geometry/tile matrix set, not a projection)

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+
+const getTextureFloat = function getTextureFloat(buffer) {
+    const texture = new THREE.DataTexture(buffer, SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE, THREE.AlphaFormat, THREE.FloatType);
+    texture.needsUpdate = true;
+    return texture;
+};
+
+export default {
+    cropXbilTexture(texture, pitch) {
+        return _cropXbilTexture(texture, pitch);
```
autra: This proxy is not needed, you can rename `_cropXbilTexture` in `cropXbilTexture` and expose this directly,
```js
export default {
    cropXbilTexture,
    getColorTextureByUrl(url) {
          /* ... */  
    }
    /* ... */
```

Btw, defining named function and exposing them in default object instead of using the property shortcut allows you to reference them directly if they need to call each other. Otherwise, a method B would need to use  `this.methodA`. Referencing them directly is good, because it allows functions to be passed around as ref without worrying about `this`, so avoiding the need to `bind`. 

To be clearer, here in this particular case:

```js 
function a() { /*...*/ }
function b() { /* ... */}
export default {
    a,
    b,
}
```

is better than 
```js
export default {
    a() { /*...*/ },
    b() {/* ...*/ }
}
```
because in the first case, b can call a() even if you take it out of the object (by doing `let myB = obj.b`). In the second case, you need to use `this.a()` in b, therefore you need to use `bind` if you want to pass this method around. As `TiledImageTools` is a helper (has no instances) the first pattern can be used easily and is more robust

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
+import * as THREE from 'three';
+import Fetcher from './Fetcher';
+import CacheRessource from './CacheRessource';
+import IoDriver_XBIL from './IoDriver_XBIL';
+import Projection from '../../Geographic/Projection';
+import CoordWMTS from '../../Geographic/CoordWMTS';
+import { UNIT } from '../../Geographic/Coordinates';
+
+const SIZE_TEXTURE_TILE = 256;
```
autra: If I'm not mistaken SIZE_TEXTURE_TILE is dependent of the TileMatrixSet. I know that the only TileMatrixSets we currently support use this value, but I'd rather not make things worse here, and use the parametized values in functions that uses this instead (layer.width and layer.height). It would allow to support other TileMatrixSets more easily.

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+};
+
+export default {
+    cropXbilTexture(texture, pitch) {
+        return _cropXbilTexture(texture, pitch);
+    },
+    getColorTextureByUrl(url) {
+        const textureCached = cache.getRessource(url);
+
+        if (textureCached !== undefined) {
```
autra: `if (textureCached)` ?

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+        texture.minFilter = THREE.LinearFilter;
+        texture.anisotropy = 16;
+
+        return promise.then(() => {
+            cache.addRessource(url, texture);
+            return texture;
+        });
+    },
+    getXBilTextureByUrl(url, pitch) {
+        const textureCache = cache.getRessource(url);
```
autra: `const cachedTexture`

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+
+        if (textureCache !== undefined) {
```
autra: `if (cachedTexture)`

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+            return _cropXbilTexture(textureCache, pitch);
+        }
+
+        return ioDXBIL.read(url).then((result) => {
+            // RGBA is needed for navigator with no support in texture float
```
autra: Is this a TODO, a FIXME or a general comment explaining something?

autra: (if it's a TODO, as I suspect it is: 
```js
// TODO RGBA texture must be used for browsers without float texture support
```
)

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+            // In RGBA elevation texture LinearFilter give some errors with nodata value.
+            // need to rewrite sample function in shader
```
autra: Is this a TODO? Should be in the said shader then?

That being said, I know this comes from `WMTS_Provider`, so I'm not requiring these comments to be fixed here (it's nice if you do though).

src/Scene/NodeProcess.js
```diff
@@ -99,6 +100,7 @@ NodeProcess.prototype.subdivideNode = function subdivideNode(node, camera, param
                 // update Imagery wmts
                 for (const layer of colorLayers) {
                     if (layer.tileInsideLimit(child, layer)) {
+                        TiledImageTools.computeTileWMTSCoordinates(child, layer);
```
autra: This is only for WMTS layer isn't it? if so, why are you doing this here?

src/Scene/NodeProcess.js
```diff
@@ -119,6 +121,7 @@ NodeProcess.prototype.subdivideNode = function subdivideNode(node, camera, param
                 const elevationLayers = params.layersConfig.getElevationLayers();
                 let canHaveElevation = false;
                 for (const layer of elevationLayers) {
+                    TiledImageTools.computeTileWMTSCoordinates(child, layer);
```
autra: Ok, we only support wmts elevation, but still, I still feel this doesn't belong here.

src/Scene/NodeProcess.js
autra: @Jeremy-Gaillard as you requested this change, any objection to have this in `WMTS_Provider`, for example in `executeCommand`?

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+
+        const tileMatrixSet = wmtsLayer.options.tileMatrixSet || 'WGS84G';
+        if (!(tileMatrixSet in tile.wmtsCoords)) {
+            const tileCoord = projection.WGS84toWMTS(tile.bbox);
+
+            tile.wmtsCoords[tileMatrixSet] =
+                projection.getCoordWMTS_WGS84(tileCoord, tile.bbox, tileMatrixSet);
+        }
+    },
+    WMTS_WGS84Parent(cWMTS, levelParent, pitch) {
```
autra: Should probably stay in `WMTS_Provider`.

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+        const r = (cWMTS.row - (cWMTS.row % diff)) * invDiff;
+        const c = (cWMTS.col - (cWMTS.col % diff)) * invDiff;
+
+        pitch.x = cWMTS.col * invDiff - c;
+        pitch.y = cWMTS.row * invDiff - r;
+        pitch.z = invDiff;
+
+        return new CoordWMTS(levelParent, r, c);
+    },
+    WMS_WGS84Parent(bbox, bboxParent) {
```
autra: Should probably stay in `WMS_Provider`.

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+            result.texture.magFilter = THREE.LinearFilter;
+            result.texture.minFilter = THREE.LinearFilter;
+            result.pitch = pitch;
+
+            cache.addRessource(url, result.texture);
+
+            return result;
+        });
+    },
+    computeTileWMTSCoordinates(tile, wmtsLayer) {
```
autra: See my comments in NodeProcess. The way we solve this discussion will determine the location of this method.

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+import Fetcher from './Fetcher';
+import CacheRessource from './CacheRessource';
+import IoDriver_XBIL from './IoDriver_XBIL';
+import Projection from '../../Geographic/Projection';
+import CoordWMTS from '../../Geographic/CoordWMTS';
+import { UNIT } from '../../Geographic/Coordinates';
+
+const SIZE_TEXTURE_TILE = 256;
+
+// CacheRessource is necessary for neighboring PM textures
```
autra: Can you elaborate?

Is it because you need to get several PM texture to cover a WSG84 tile?

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+// Info : THREE.js have cache image https://github.com/mrdoob/three.js/blob/master/src/loaders/ImageLoader.js#L25
```
autra: Yes, and unfortunately, if I'm not mistaken, they fail to evict it too :-(

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+            dimParent.x;
+        const y =
+            Math.abs(
+                bbox.south(UNIT.RADIAN) + dim.y -
+                (bboxParent.south(UNIT.RADIAN) + dimParent.y)) /
+            dimParent.y;
+
+        return new THREE.Vector3(x, y, scale);
+    },
+};
```
autra: So this factorizes code from both `WMS_Provider` and `WMS_Provider`. It cannot be called `TiledSomething`.

@peppsac 's proposal (`OGCWebServiceHelper`) sounds good to me. Feel free to propose something else though.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -20,23 +16,9 @@ import BoundingBox from '../../../Scene/BoundingBox';
...
-
-        texture.needsUpdate = true;
-        return texture;
-    };
+function WMS_Provider() {
 }
 
-WMS_Provider.prototype = Object.create(Provider.prototype);
-
 WMS_Provider.prototype.constructor = WMS_Provider;
```
autra: This line is now unneeded.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -90,130 +76,52 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
...
-            tile.bbox;
-
+    const func = supportedFormats[layer.format];
 
-        return func(tile, layer, bbox, pitch);
+    if (func) {
+        return func(tile, layer);
     } else {
         return Promise.reject(new Error(`Unsupported mimetype ${layer.format}`));
     }
```
autra, line 114: Refacto proposal if you want, from line 111:

```js
switch (layer.format) {
    case 'image/png':
    case 'image/jpg':
    case 'image/jpeg':
        return this.getColorTexture(tile, layer);
    case 'image/x-bil;bits=32':
        return this.getXbilTexture(tile, layer);
    default:
        return Promise.reject(new Error(...));
```

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -4,45 +4,13 @@
...
-
-            texture.needsUpdate = true;
-            return texture;
-        }; }
+function WMTS_Provider() {
 }
 
-WMTS_Provider.prototype = Object.create(Provider.prototype);
-
 WMTS_Provider.prototype.constructor = WMTS_Provider;
```
autra: Not needed any more

src/Globe/TileMesh.js
```diff
@@ -159,19 +159,34 @@ TileMesh.prototype.setBBoxZ = function setBBoxZ(min, max) {
...
     if (textures) {
-        this.material.setTexturesLayer(textures, layerType, layer);
+        this.material.setTexturesLayer(textures, layerType, layerId);
     }
     this.loadingCheck();
 };
 
-TileMesh.prototype.isColorLayerDownscaled = function isColorLayerDownscaled(layer) {
+TileMesh.prototype.getLayerTextures = function getLayerTextures(layerType, layerId) {
     var mat = this.materials[RendererConstant.FINAL];
```
autra: Proposal: in this method an the three below: use `const mat`, or even better, get rid of the intermediary var. The engine will optimize it away anyway, but it would be faster when correcting linter errors.

src/Renderer/LayeredMaterial.js
```diff
@@ -362,29 +362,29 @@ LayeredMaterial.prototype.getLoadedTexturesCount = function getLoadedTexturesCou
...
 };
 
-LayeredMaterial.prototype.isColorLayerDownscaled = function isColorLayerDownscaled(layer, level) {
-    return this.textures[l_COLOR][this.getLayerTextureOffset(layer)].level < level;
+LayeredMaterial.prototype.isColorLayerDownscaled = function isColorLayerDownscaled(layerId, level) {
+    if (this.textures[l_COLOR][this.getLayerTextureOffset(layerId)]) {
+        return this.textures[l_COLOR][this.getLayerTextureOffset(layerId)].coordWMTS.zoom < level;
+    } else {
+        return false;
+    }
```
autra: nits
```js
return this.textures[l_COLOR][this.getLayerTextureOffset(layerId)] &&
    this.textures[l_COLOR][this.getLayerTextureOffset(layerId)].coordWMTS.zoom < level;
```
```

File: src/Renderer/LayeredMaterial.js [hunk 47]
```diff
@@ -362,29 +362,29 @@ LayeredMaterial.prototype.getLoadedTexturesCount = function getLoadedTexturesCou
...
-        return tex.level < level;
+        return tex.coordWMTS.zoom < level;
     } else if (layerType === l_COLOR) {
         // browse each layer
         for (let index = 0, max = this.colorLayersId.length; index < max; index++) {
             const offset = this.getTextureOffsetByLayerIndex(index);
-            if (this.textures[l_COLOR][offset].level < level) {
+            const texture = this.textures[l_COLOR][offset];
+            if (texture.coordWMTS.zoom < level) {
                 return true;
```
autra: nits : 
```js
return this.textures[l_color][offset].coordWMTS.zoom < level;
```

src/Scene/NodeProcess.js
```diff
@@ -253,48 +237,36 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
             }
         }
 
-        let ancestor = null;
-
+        node.layerUpdateState[layer.id].newTry();
```
autra: There is still a case where we `continue` below, can you explain why you mark a new try here and not after this `continue`?

src/Scene/NodeProcess.js
```diff
@@ -253,48 +237,36 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
...
-                    for (let j = 0; j < result.length; j++) {
-                        result[j].texture.level = level;
-                    }
-
                     node.setTexturesLayer(result, l_COLOR, layer.id);
                 } else if (result.texture) {
-                    result.texture.level = level;
+                    if (!result.texture.coordWMTS) {
+                        result.texture.coordWMTS = node.wmtsCoords[layer.options.tileMatrixSet || 'WGS84G'][0];
+                    }
```
autra, line 253: Maybe this belongs to `WMTS_Provider.executeCommand`

src/Scene/NodeProcess.js
```diff
@@ -398,16 +354,18 @@ function updateNodeElevation(scene, quadtree, node, layersConfig, force) {
     if (bestLayer !== null) {
         node.layerUpdateState[bestLayer.id].newTry();
 
+        // Elevation layer search in parent, from the moment it exceeds its maximum zoom
+        const searchInParent = (!node.isElevationLayerLoaded() && node.parent.isElevationLayerLoaded()) || (node.level > bestLayer.zoom.max);
```
autra: let's invert the shortcut here and test `node.level > bestLayer.zoom.max` first. It's more readable and would perform better, because isElevationLayer can potentially access some pretty big indices on array. I don't know if it is something noticeable here, but inverting this condition is cheap so...

src/Scene/NodeProcess.js
```diff
@@ -418,13 +376,9 @@ function updateNodeElevation(scene, quadtree, node, layersConfig, force) {
...
 
-                if (terrain.texture) {
-                    terrain.texture.level = (ancestor || node).level;
-                }
-
                 if (terrain.max === undefined) {
-                    terrain.min = (ancestor || node).bbox.bottom();
-                    terrain.max = (ancestor || node).bbox.top();
+                    terrain.min = (node.parent || node).bbox.bottom();
+                    terrain.max = (node.parent || node).bbox.top();
```
autra: shouldn't it be `(node || node.parent)`?

src/Core/Commander/Providers/TiledImageTools.js
```diff
@@ -0,0 +1,126 @@
...
+
+const SIZE_TEXTURE_TILE = 256;
+
+// CacheRessource is necessary for neighboring PM textures
+// Info : THREE.js have cache image https://github.com/mrdoob/three.js/blob/master/src/loaders/ImageLoader.js#L25
+const cache = CacheRessource();
+const ioDXBIL = new IoDriver_XBIL();
+const projection = new Projection();
+
+const _cropXbilTexture = function _cropXbilTexture(texture, pitch) {
```
autra: Actually, I don't see the point of this method: it only calls `computeMinMaxElevation` and returns an object containing unchanged results from this method, and *unchanged* parameters. It makes the API of this file unclear.

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -109,147 +73,65 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
 
-    var func = supportedFormats[layer.options.mimetype];
+    const func = supportedFormats[layer.options.mimetype];
     if (func) {
-        return func(tile, layer, command);
+        return func(tile, layer, parentTextures);
     } else {
         return Promise.reject(new Error(`Unsupported mimetype ${layer.options.mimetype}`));
     }
 };
```
autra: See refacto proposal in `WMS_Provider`

src/Scene/NodeProcess.js
```diff
@@ -253,48 +237,36 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
...
-
         const command = {
             /* mandatory */
             layer,
             requester: node,
             priority: nodeCommandQueuePriorityFunction(node),
             earlyDropFunction: refinementCommandCancellationFn,
-            /* specific params */
-            ancestor,
+            parentTextures: searchInParent ? node.parent.getLayerTextures(l_COLOR, layer.id) : undefined,
```
autra: When we search in `parentTextures`, things are completely synchronous. So we don't need to pass through the scheduler. Not doing so will have several advantages:

- it would be clearer to the reader that things are synchrone if it is another method call.
- it would be easier to go multi-thread with Workers. If we go multithread, we would have no guarantee the parentTexture will be split before children receives the requested new tiles.
- It would make `provider.executeCommand` simpler.

autra: Basically, instead of computing a `searchInParent` boolean, let's do the work there, and avoid the command scheduling in this case.

Scheduling is only useful when we have some network operations to do (because then we want to parallelize things, to wait for completion, to handle errors etc...)

src/Core/Commander/Providers/TiledImageTools.js
autra: let's call `computeMinMaxElevation direction and in providers

```js
const {min, max} = ioDXBIl.computeMinMax(/* ... */);
return Promise.resolve({pitch, texture, min, max});
```

src/Core/Commander/Providers/TiledImageTools.js
autra: But actually, I think we don't need this method anyway.

autra (changes requested):
Things are mostly looking good :-) There is a lot of comments, but the PR is big, so...

src/Core/Commander/Providers/TiledImageTools.js
gchoqueux: yes, need new PR

src/Core/Commander/Providers/TiledImageTools.js
gchoqueux: The PM textures overlap several tiles WGS84, it is to avoid net requests

src/Core/Commander/Providers/TiledImageTools.js
gchoqueux: that is to say?

src/Core/Commander/Providers/TiledImageTools.js
gchoqueux: Ok, I took note of the remark, but i factorize in `TiledImageTools.js `

autra: Why? It makes no sense to do that. `TiledImageTools` is not a promise factory. Factorization is only good when it's useful. And it's not useful to avoid a `Promise.resolve`. Adding a method just for that contribute to make the code harder to read and to maintain.

src/Core/Commander/Providers/TiledImageTools.js
gchoqueux: The shader `GlobeFS.glsl` already takes into account the comment

src/Scene/NodeProcess.js
gchoqueux: I need WMTS coordinate here because the **WMS** is also required in tile

src/Core/Commander/Providers/WMS_Provider.js
peppsac: Note that the elevation texture special case handling doesn't really belong to this provider.
For instance, I might have a WMS server supplying grayscale image to use as a MNT (and I've done this at least twice for demos built with itowns).

So deducing image usage from image format is wrong. It does'nt require fixing here, but keep this in mind while refactoring.

src/Scene/NodeProcess.js
gchoqueux: This is done here because we need to know the WMTS coordinates to know if we require new texture or not

src/Scene/NodeProcess.js
gchoqueux: Same 
> This is done here because we need to know the WMTS coordinates to know if we require new texture or not

src/Scene/NodeProcess.js
gchoqueux: Yes it is true, it should be seen in a new PR

We must not lose sight of the prioritization of the requests with the scheduler.

src/Scene/NodeProcess.js
gchoqueux: I need this for WMS too. because WMS is requested by tile

src/Renderer/LayeredMaterial.js
gchoqueux: I don't think because we must check all the textures

src/Core/Commander/Providers/TiledImageTools.js
gchoqueux: I will look at

gchoqueux:
thanks for reviewing @autra 
PR updated

src/Scene/NodeProcess.js
```diff
@@ -355,7 +355,7 @@ function updateNodeElevation(scene, quadtree, node, layersConfig, force) {
         node.layerUpdateState[bestLayer.id].newTry();
 
         // Elevation layer search in parent, from the moment it exceeds its maximum zoom
-        const searchInParent = (!node.isElevationLayerLoaded() && node.parent.isElevationLayerLoaded()) || (node.level > bestLayer.zoom.max);
+        const searchInParent = (!node.isElevationLayerLoaded() && node.parent.isElevationLayerLoaded()) || (bestLayer.zoom.max < node.level);
```
autra: No I meant:
```js
const searchInParent = (bestLayer.zoom.max < node.level) || (!node.isElevationLayerLoaded() && node.parent.isElevationLayerLoaded());

Comment (autra) [reply src/Scene/NodeProcess.js]: This way, node.isElevationLayerLoaded is not called if first condition is true.

File: src/Core/Commander/Providers/TiledImageTools.js (hunk 40 above)
Comment (autra) [reply src/Core/Commander/Providers/TiledImageTools.js]: THREE.js uses cache, but never evict it, never check its validity, doesn't have any limits in size. It's really bad. I'm just commenting on the `Info` comment. No action needed in this PR :-)

File: src/Core/Commander/Providers/TiledImageTools.js (hunk 39 above)
Comment (autra) [reply src/Core/Commander/Providers/TiledImageTools.js]: Ok then.

File: src/Core/Commander/Providers/TiledImageTools.js (hunk 28 above)
Comment (autra) [reply src/Core/Commander/Providers/TiledImageTools.js]: Ok, it's coming from old code. I might change this in the `getCapabilities` PR.

File: src/Renderer/LayeredMaterial.js (hunk 47 above)
Comment (autra) [reply src/Renderer/LayeredMaterial.js]: Yes, sorry I meant
```js
const texture = this.textures[l_COLOR][offset];
return textures.level < level && textures.coordWMTS.zoom < level;
```
`if (condition) return true` is always equivalent to `return condition`.

btw why do we need to test textures.level *and* textures.coordWMTS.level ? Isn't it redundant ?

src/Scene/NodeProcess.js
autra: Ok I see. WMTS is a protocol. You're speaking about our *TileMatrixSet*. Let's call that `computeTileMatrixSetCoordinates`. Same for the other occurrences.

autra: I do hope we will support other tileMatrixSets than WSG84G in the future.

src/Scene/NodeProcess.js
autra: > it should be seen in a new PR

This is something you add in *this* PR. I understand we delay refactoring of existing code, but here, this is not the case. Moreover, it seems to me that the goal of this PR was to remove ancestor, in order to remove the transitional state where a node has children, but they are not ready yet. If we achieve this original goal, this would greatly simplify the code and perform better. For this, this children preparation must be synchronous. 

Yes the scheduler prioritize things, but here, there is no notion of priority : it should be done now. Also, this scheduling has runtime cost (using an external lib, sorting the queue...). We should use it only when we must.

autra:
> PR updated

Still some actions needed
- https://github.com/iTowns/itowns2/pull/244#discussion_r111219933
- https://github.com/iTowns/itowns2/pull/244#discussion_r111165042

And I have added some remarks and answers.

src/Scene/NodeProcess.js
gchoqueux: You really want to change? Because wmts coordinate is not a protocol.
It already supports PM

src/Renderer/LayeredMaterial.js
gchoqueux: we test only :`texture.coordWMTS.zoom < level`

gchoqueux:
new update

src/Renderer/LayeredMaterial.js
autra: Ok ok yes, I miread the diff in my second comment. I thought that's what you meant by 

> I don't think because we must check all the textures

So we're back at my first comment, which is actually correct: 

> nits :
> ```js
> return this.textures[l_color][offset].coordWMTS.zoom < level;
> ```

And 

> `if (condition) return true` is always equivalent to `return condition;`

Do you see what I mean?

autra: Ok, the return false is after the loop, my mistake. Sorry for the noise :-)

src/Scene/NodeProcess.js
autra: [WMTS *is* a protocol](https://en.wikipedia.org/wiki/Web_Map_Tile_Service). 

Yes, we support WSG84G TileMatrixSet + we kind of support PM TileMatrixSet. But actually, everything fits together because our geometry follow the same tile scheme. It is specific to the globe, and even more : to a IGN-homemade TileMatrixSet. So this still needs renaming, but I'm gonna start a new discussion in the right place (where this method is defined) proposing something in my next review.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -199,6 +199,10 @@ ApiGlobe.prototype.removeImageryLayer = function removeImageryLayer(id) {
  */
 
 ApiGlobe.prototype.addElevationLayer = function addElevationLayer(layer) {
+    if (layer.protocol === 'wmts' && layer.options.tileMatrixSet !== 'WGS84G') {
+        throw new Error('Only WSG84 tileMatrixSet is currently supported for WMTS elevation layers');
```
autra: `Only WSG84G...` sorry I might have mispelled it in my previous review.

src/Core/Commander/Providers/OGCWebServiceHelper.js
```diff
@@ -0,0 +1,84 @@
...
+            result.texture.magFilter = THREE.LinearFilter;
+            result.texture.minFilter = THREE.LinearFilter;
+            result.pitch = pitch;
+
+            cache.addRessource(url, result.texture);
+
+            return result;
+        });
+    },
+    computeTileMatrixSetCoordinates(tile, wmtsLayer) {
```
autra: I'm not requiring anything here as it is a big topic, just sharing some thoughts:
- all these wmtsFoo variables and methods should get renamed to tileMatrixSetFoo
- In the future `WGS84ToWMTS` might be a `coordinate.toTileMatrixSetCoordinates( tileMatrixSetId)` or a `bbox.toTileMatrixSetCoordinates(tileMatrixSetId` (knowing that a coordinate embed its crs). This would allow us to support custom tileMatrixSet.

src/Core/Commander/Providers/WMS_Provider.js
```diff
@@ -73,6 +74,7 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.version = layer.version || '1.3.0';
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
+    layer.options = {};
```
autra: Are you sure about that? layer.options might not be empty (ex `Region.json`)

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -109,147 +86,69 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
-
-        cooWMTS = this.projection.WMTS_WGS84Parent(
-            cooWMTS,
-            this.computeLevelToDownload(tile, parameters.ancestor, layer) + levelOffset,
+WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, parentTextures) {
+    let coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
+    const pitch = new THREE.Vector3(0.0, 0.0, 1.0);
+
+    if (parentTextures) {
+        coordWMTS = WMTS_WGS84Parent(
```
autra: If I'm not mistaken, `WMTS_WGS84Parent` returns a new object, but here you're not doing anything with `coordWMTS` before returning. Likely a mistake ?

src/Renderer/LayeredMaterial.js
```diff
@@ -394,26 +391,41 @@ LayeredMaterial.prototype.isLayerTypeDownscaled = function isLayerTypeDownscaled
...
+    return texture ? texture.coordWMTS.zoom : EMPTY_TEXTURE_ZOOM;
 };
 
 LayeredMaterial.prototype.getElevationLayerLevel = function getElevationLayerLevel() {
-    return this.textures[l_ELEVATION][0].level;
+    return this.textures[l_ELEVATION][0].coordWMTS.zoom;
 };
 
-LayeredMaterial.prototype.getLayerLevel = function getLayerLevel(layerType, layer) {
+LayeredMaterial.prototype.getLayerLevel = function getLayerLevel(layerType, layerId) {
```
autra: If I'm not mistaken, you removed the only calls to this method, it's now unused. Let's remove it.

autra: The switch is now entirely in NodeProcess I guess.

src/Renderer/LayeredMaterial.js
```diff
@@ -394,26 +391,41 @@ LayeredMaterial.prototype.isLayerTypeDownscaled = function isLayerTypeDownscaled
 };
 
 LayeredMaterial.prototype.getColorLayerLevelById = function getColorLayerLevelById(colorLayerId) {
-    let index = this.indexOfColorLayer(colorLayerId);
-    // TODO: hiding the "colorLayerId is invalid", it's problem, needs new PR
+    const index = this.indexOfColorLayer(colorLayerId);
     if (index === -1) {
-        index = 0;
+        return -1;
```
autra:
```js
throw new Error(`Invalid layer id ${colorLayerId}`)
``` 
?

src/Core/Commander/Scheduler.js
```diff
@@ -120,7 +120,10 @@ Scheduler.prototype.execute = function execute(command) {
     const q = host ? this.hostQueues.get(host) : this.defaultQueue;
 
     // execute command now if possible
-    if (q.counters.executing < this.maxCommandsPerHost) {
+    if (forceRunNow) {
```
autra: Not sure about this. Wouldn't it be simpler to simply call `provider.executeCommand` if `searchInParent` is true in NodeProcess.processNodeImagery or Elevation?

src/Scene/NodeProcess.js
autra: ok

autra (changes requested):
One more step and I guess we'll be good :-)

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -109,147 +86,69 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
+            coordWMTS,
+            parentTextures[0].coordWMTS.zoom,
             pitch);
+        const { min, max } = OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
+                parentTextures[0].image.data,
+                SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
+                pitch);
+        return Promise.resolve({ pitch, texture: parentTextures[0], min, max });
```
peppsac: I'd suggest to *not* use `Promise` where not needed.
Just return your object if this task can be performed directly.
You'll to adapt `Scheduler` line:
```
var p = provider.executeCommand(cmd) || Promise.resolve();
```
to test if the result from `executeCommand` is a Promise or a plain object.

src/Core/Commander/Scheduler.js
```diff
@@ -101,7 +101,7 @@ Scheduler.prototype.runCommand = function runCommand(command, queue, executingCo
     });
 };
 
-Scheduler.prototype.execute = function execute(command) {
+Scheduler.prototype.execute = function execute(command, forceRunNow) {
```
peppsac: Rather than adding a boolean which makes code reading harder, I'd prefer a 2nd function.
```js
sheduler.execute(cmd, false);
sheduler.execute(cmd, true);
```
is less clear than:
```js
sheduler.execute(cmd);
sheduler.executeNow(cmd);
```
You could even change the return type for each function.
`execute` would return a Promise, while `executeNow` would return the result (and fail with an error if the requested command cannot be executed in a synchronous fashion).

peppsac:
Maybe we could simplify the process of inheriting its parent's textures for a node.

If texture had another attribute describing the tile area covered, we could deduce the offset and scale to use for a child without asking the original provider (and it's a trivial function to write).

With this change, we wouldn't even need to add a `forceRunNow` option to the scheduler.

Does this make sense or am I missing something?

src/Core/Commander/Scheduler.js
autra: I'm in favor of this change. let's not add more complexity to `execute`

src/Renderer/LayeredMaterial.js
gchoqueux: no, it isn't a error.

autra: Then `return EMPTY_TEXTURE_ZOOM`? My point is that we shouldn't return a magic number.

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -109,147 +86,69 @@ WMTS_Provider.prototype.url = function url(coWMTS, layer) {
...
-
-        cooWMTS = this.projection.WMTS_WGS84Parent(
-            cooWMTS,
-            this.computeLevelToDownload(tile, parameters.ancestor, layer) + levelOffset,
+WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, parentTextures) {
+    const coordWMTS = tile.wmtsCoords[layer.options.tileMatrixSet][0];
+    const pitch = new THREE.Vector3(0.0, 0.0, 1.0);
+
+    if (parentTextures) {
+        WMTS_WGS84Parent(
```
autra: Well all right. I see. This WMTS_WGS84Parent function is really bad. It has side-effects AND returns a value. This is very confusing and error-prone. I'm not requiring anything here as it is just a copy-paste, but we might want to refacto that in the future.

autra (dismissed):
Still https://github.com/iTowns/itowns2/pull/244/files#r111940993 and last @peppsac comment to address/answer and we will be done here I guess :-)

src/Renderer/LayeredMaterial.js
```diff
@@ -76,7 +90,7 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
 };
 /* eslint-enable */
 
-const LayeredMaterial = function LayeredMaterial(id) {
+const LayeredMaterial = function LayeredMaterial(material, coord) {
```
peppsac: The rest look good so I'll focus only on this part.

Here's what I had in mind:

- Replace LayeredMaterial `offsetScale` with:
```js
// express what part of the texture (source) is drawn,
// and where (destination) it is drawn.
// (mapping.source is our current `offsetScale`)
// (mapping.destination is currently baked into UV attributes of a tile)
mapping = {
    source: {
        origin: new THREE.Vector2(0, 0),
        size: 1,
    },
    destination: {
        origin: new THREE.Vector2(0, 0),
        size: 1,
    },
}
```

- replace `pitch` result from providers with a `mapping` object. Currently
  only PM textures will have a non-default destination

- replace `coord` in `function LayeredMaterial(material, coord)` with tile's mapping attribute,
  as well as it's parent's mapping attribute.
  With this change, `offsetToParent` can be written without any dependencies on WMTS logic.

What do you think ?

gchoqueux: I would like to do. We need to talk about this.

-  `mapping` will not bear directly into uniforms

gchoqueux: > dependencies on WMTS logic.

```js
var emptyTexture = new THREE.Texture();
emptyTexture.coordWMTS = { zoom: EMPTY_TEXTURE_ZOOM };
```

There was already a dependency with WMTS. But it should also be deleted

gchoqueux: As some projections are not linear (like PM), it is simpler to do the calculations in the same system and use WMTS coordinate.

gchoqueux: I think to keep WMTS logic.
it's more simple to work in same system (specially for non-linear systems)
What do you think?

src/Core/Commander/Scheduler.js
peppsac: `forceRunNow` option is never used now (which is good), so please remove it.

src/Scene/Quadtree.js
```diff
@@ -24,7 +24,7 @@ function Quadtree(type, schemeTile, link) {
     this.schemeTile = schemeTile;
     this.tileType = type;
     this.minLevel = 2;
-    this.maxLevel = 17;
+    this.maxLevel = 18;
```
peppsac: This change is unrelated to the PR (as is the removal of imagery layers).
You should be extra careful to not mix these kind of changes in a refactoring PR.

src/Renderer/LayeredMaterial.js
peppsac: Comment not addressed.

src/Renderer/LayeredMaterial.js
```diff
@@ -146,11 +160,56 @@ const LayeredMaterial = function LayeredMaterial(id) {
...
 
-    this.setUuid(id || 0);
-    this.wireframe = false;
-    // this.wireframe = true;
-
     this.colorLayersId = [];
+    this.elevationLayersId = [];
+
+    if (material instanceof LayeredMaterial) {
+        // Getting inheritance from elevation
```
peppsac: May I suggest this comment instead:
```js
// inherit parent's elevation texture
```
(and same for the similar ones below)

src/Renderer/LayeredMaterial.js
```diff
@@ -146,11 +160,56 @@ const LayeredMaterial = function LayeredMaterial(id) {
...
+        if (material.getElevationLayerLevel() > -1) {
```
peppsac: `!= EMPTY_TEXTURE_ZOOM` ?

src/Renderer/LayeredMaterial.js
```diff
@@ -146,11 +160,56 @@ const LayeredMaterial = function LayeredMaterial(id) {
...
+
+        // Getting inheritance from lighting
+        this.uniforms.lightPosition.value = material.uniforms.lightPosition.value;
+        this.uniforms.lightingEnabled.value = material.uniforms.lightingEnabled.value;
+
+        // Getting inheritance from color
+        let offsetTextures = 0;
+        for (let i = 0; i < material.colorLayersId.length; i++) {
+            const layerId = material.colorLayersId[i];
+            if (material.getColorLayerLevelById(layerId) > -1) {
```
peppsac: `!= EMPTY_TEXTURE_ZOOM` ?

peppsac: The whole loop could benefit from a comment explaining its purpose (and "how" it is achieved).

src/Renderer/LayeredMaterial.js
peppsac: I disagree :)
But that something we can fix later, so I won't block any longer on this point.

peppsac (changes requested):
A few comments to address (please make sure you reply to all of them, so reviewers know that you took a look at them).
As wrote in another comment, this is my last "Request changes" review for this PR - next time I'll "Approve" it so we can move forward.

src/Core/Commander/Providers/OGCWebServiceHelper.js
```diff
@@ -37,43 +38,50 @@ export default {
...
             result.texture.magFilter = THREE.LinearFilter;
             result.texture.minFilter = THREE.LinearFilter;
-            result.pitch = pitch;
-
             cache.addRessource(url, result.texture);
 
             return result;
         });
+
+        cachePending[url] = promiseXBil;
```
peppsac: `cachePending` will grow indefinitely.
You could probably remove elements from it after the line:
```js
cache.addRessource(url, result.texture);
```

src/Globe/Globe.js
```diff
@@ -202,6 +202,16 @@ Globe.prototype.removeColorLayer = function removeColorLayer(layer) {
     this.tiles.children[0].traverse(cO);
 };
 
+Globe.prototype.update = function update() {
+    var cO = function cO(object) {
+        if (object.loaded !== undefined) {
+            object.loaded = false;
```
peppsac: What is the purpose of this function?

src/Scene/NodeProcess.js
```diff
@@ -187,17 +158,25 @@ NodeProcess.prototype.hideNodeChildren = function hideNodeChildren(node) {
...
-    //
-    if (!node.loaded) {
-        // Prioritize lower-level (ie: bigger) non-loaded nodes
-        // because we need them to be loaded to be able
-        // to subdivide.
-        return 1000 - node.level;
-    } else if (node.isDisplayed()) {
+
+
+    // NOTE : node.loaded is removed
```
peppsac: Remove this - no need to keep comments refering to removed code

src/Scene/NodeProcess.js
```diff
@@ -187,17 +158,25 @@ NodeProcess.prototype.hideNodeChildren = function hideNodeChildren(node) {
...
+    //     // Prioritize lower-level (ie: bigger) non-loaded nodes
+    //     // because we need them to be loaded to be able
+    //     // to subdivide.
+    //     return 1000 - node.level;
+    // } else
+
+    // Priorisation of the most degraded tiles
+    const priority = node.level * Math.round(node.sse);
+
+    if (node.isDisplayed()) {
```
peppsac: Again: this is a functionnal change in a refacto PR.
Avoid this please.

src/Scene/NodeProcess.js
```diff
@@ -225,6 +205,25 @@ function updateNodeImagery(scene, quadtree, node, layersConfig, force) {
             continue;
         }
 
+        const material = node.materials[RendererConstant.FINAL];
+
+
+        if (material.indexOfColorLayer(layer.id) === -1) {
```
peppsac, line 199: `=== EMPTY_TEXTURE_ZOOM`?

gchoqueux: is index not level

src/Renderer/LayeredMaterial.js
```diff
@@ -76,7 +90,7 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
 };
 /* eslint-enable */
 
-const LayeredMaterial = function LayeredMaterial(id) {
+const LayeredMaterial = function LayeredMaterial(material, coordDestination, coordSource) {
```
gchoqueux: @peppsac , you disagree, me too I prefer a most generic thing, as you suggest.
I ask for another opinion because I think it isn't possible with this `mapping` method,  due to nonlinear projections (PM, Lambert...) This method requires several unnecessary conversions. On the other hand it is perfect in the WGS84 system
```js
mapping = {
    source: {
        origin: new THREE.Vector2(0, 0),
        size: 1,
    },
    destination: {
        origin: new THREE.Vector2(0, 0),
        size: 1,
    },
}
```

src/Scene/BoundingBox.js
```diff
@@ -138,9 +138,16 @@ BoundingBox.prototype.offsetScale = function offsetScale(bbox) {
...
-    var pitX = Math.abs(bbox.west() - this.west()) / this._dimension.x;
-    var pitY = Math.abs(bbox.north() - this.north()) / this._dimension.y;
-    var scale = bbox._dimension.x / this._dimension.x;
+
+    const dimension = {
+        x: Math.abs(this.east() - this.west()),
+        y: Math.abs(this.north() - this.south()),
+    };
+
+    var pitX = Math.abs(bbox.west() - this.west()) / dimension.x;
```
peppsac: I don't think you want to use `Math.abs` here.
Instead you should check that inputs are valid (ie : that `bbox` is contained in `this`) and throw an error otherwise

peppsac: Also I'd prefer `originX` and `originY` to `pitX` and `pitY`

src/Core/Commander/Providers/OGCWebServiceHelper.js
```diff
@@ -0,0 +1,92 @@
...
+import Projection from '../../Geographic/Projection';
+
+
+export const SIZE_TEXTURE_TILE = 256;
+
+// CacheRessource is necessary for neighboring PM textures
+// The PM textures overlap several tiles WGS84, it is to avoid net requests
+// Info : THREE.js have cache image https://github.com/mrdoob/three.js/blob/master/src/loaders/ImageLoader.js#L25
+const cache = CacheRessource();
+const cachePending = [];
```
peppsac: How about using a [Map](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Map) instead?

peppsac (commented):
Mostly good, just a few things to address and we're good to go

peppsac (approved):

Issue #223, opened 2017-01-03
WGS84


Jeremy-Gaillard:
Hi,

Is there a reason why the iTowns globe doesn't use the standard WGS84 axes? [In the WGS specification](http://www.unoosa.org/pdf/icg/2012/template/WGS_84.pdf), the z axis goes through the north pole, while in iTowns, it is the y axis. This is pretty annoying when trying to import external data, since it requires rotating the axes in the providers.

gchoqueux:
In iTowns, Three.js use Y axe is axe up.
You can import your data in this [`layerWGS84Zup`](https://github.com/iTowns/itowns2/blob/master/src/Globe/Globe.js#L40) 
And uncomment
```js
// this.add(this.layerWGS84Zup);
```

qdnguyen:
i think this problem was resolved longtime ago? #19

Jeremy-Gaillard:
There is a workaround using `layerWGS84Zup`. But not having the proper WGS84 axes by default can easily lead to errors, so we should fix it.

PR #239, opened 2017-01-25
Refacto layer; warning big PR :)

peppsac:
This PR is quite big, but it's not intended to be merged as is. I want to iterate on this work but it'll be faster if you can all give early feedback.

Here's the wording from the most important commit:
```
refactor: deep rework of layers and threejs integration
    
    See issues #217, #124, #40 for related discussion.
    
    This massive commit:
      * unifies layers handling (geometry layers are no longer special citizen)
      * split processing in smaller units: these units will be more easy to reuse. For
    example implementing Plane support will be easier because we're now able to share
    the Tile and LayeredMaterial concepts without being forced to use Globe based
    concept (e.g horizontalCulling).
      * limits the duplication of feature between three.js and iTowns: we get the best
    of three.js and we remove incoherencies (e.g our Node were slightly different)
      * breaks the Globe|Quadtree centric vision of iTowns to ease the implementation of
    other types of vizualisaiotn (oriented images, plane mode, pointcloud, etc)
```

The last 7 commits are from WFS part 1 PR #195 and are just included as an example.

PR #246, opened 2017-02-06
Fix WGS84

Jeremy-Gaillard:
This PR is based on #239 

It fixes the issue #223 (bad WGS84 axes) and slightly refactors the TileMesh class.


peppsac:
Could you submit the last commit ( 	38b4f33) in its own PR ?
And another PR for cd91e9f
This way they could be merged quickly.

Jeremy-Gaillard:
Done. Check #254 for first two commits and this one for WGS84 fix (this PR still depends on #254 though).

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -7,17 +7,21 @@
...
+function OBB(min, max, lookAt, translate, up) {
     THREE.Object3D.call(this);
     this.box3D = new THREE.Box3(min, max);
 
     this.natBox = this.box3D.clone();
 
     this.quaInv = this.quaternion.clone().inverse();
 
+    if (up) {
+        this.up.set(up.x, up.y, up.z);
```
gchoqueux: why you set UP?

gchoqueux (commented):
It looks good, I wait #254 to test

src/Renderer/ThreeExtended/OBB.js
Jeremy-Gaillard: The lookAt function needs to know which way is up. Up by default is (0,1,0). Since we changed the "main axis" from y to z, we also need to change the up vector when initialising our OBB.

Jeremy-Gaillard:
Rebased & ready

src/Renderer/ThreeExtended/OBB.js
gchoqueux: yes, I know
but it's possible to instantiate the box by swapping the min and max and translations in `builderEllipsoidTile`.
It would also be necessary to have the BBOXhelper to verify

src/Globe/BuilderEllipsoidTile.js
```diff
@@ -120,14 +120,14 @@ BuilderEllipsoidTile.prototype.OBB = function OBBFn(params) {
     }
 
     maxHeight *= 0.5;
-    var width = Math.abs(maxV.z - minV.z) * 0.5;
+    var width = Math.abs(maxV.y - minV.y) * 0.5;
     var height = Math.abs(maxV.x - minV.x) * 0.5;
     var delta = height - Math.abs(cardin3DPlane[5].x);
     var max = new THREE.Vector3(width, height, maxHeight);
```
gchoqueux, line 128: must be modifiy to remove up param in constructor

peppsac:
@Jeremy-Gaillard any update on this PR?

Jeremy-Gaillard:
If there's no rush, I'd prefer waiting for #239 to be merged since it'll considerably facilitate the implementation of the requested changes.

src/Globe/BuilderEllipsoidTile.js
Jeremy-Gaillard: Fixed. The problem came from the fact that the default up vector was (0,1,0) instead of (0,0,1). I changed it at the initialisation of the scene, in c3DEngine.

Jeremy-Gaillard:
Rebased and ready.

gchoqueux (approved):
OK
Thanks

PR #248, opened 2017-02-09
Package lib

### LINKED ISSUE #243: Generate ES5 source files in lib/ and add them to npm package
At the moment, in the npm package, we can't bundle something else than the dist folder, because our sources needs specific webpack and babel configuration. For this reason, we need to stop declare any dependencies and use devDependencies instead in our package.json, otherwise client will download these dependencies when doing `npm install`, even though they would already be included in the bundle in the `dist/` folder. This is PR #241.

Once #241 is merged, it would be nice to:
- generate ES5 files in a `lib/` folder that can be used without babel.
- Bundle these generated source files.
- move back prod dependencies to `dependencies` tag in `package.json`

This would allow client app to correctly use npm with our projects and directly require source files. The dist/ folder would be published as github release.

autra:
Here is the PR that fixes #243 

I've put a lot of explanations on each commit messages. Please do read them!

What I like:
- now our npm package are bundle/compiler agnostics. Client code can use the bundle like an old webpage, or they can require each module separately without relying on our babel config, or even on webpack loaders for glsl (as they are inlined)
- No change from our previous dev workflow
- Minimal addition to our devDependencies, and the ability to put back our real dependencies in "dependencies" tag in `package.json`

What I like less:
- There is now a bit of magic involved (webpack and npm calls babel to process js files, which in turns uses a webpack loader in `webpack-babel.config.js` to inline glsl files)

What still needs to be done:
- Build an itowns-full.js containing our code + all the dependencies, and an itowns.js without dependencies + vendor.js. Then github releases would contain these 3 files, but npm packages might contain only itowns.js (as they would have dependencies through npm)

@tbroyer this is my attempt to realize [agnostic.md](https://gist.github.com/mattdesl/aaf759da84cc44c22305) goals.

@iTowns/reviewers please r?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,7 @@
 {
-  "presets": ["es2015"],
-  "plugins": ["transform-runtime"]
+    "presets": ["es2015"],
+    "plugins": [
+        "transform-runtime",
+        [ "babel-plugin-webpack-loaders", { "config": "./webpack-babel.config.js", "verbose": false }]
+    ]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
+// please consider modifying babel-loader configuration in webpack.config.js
 {
     "presets": ["es2015"],
     "plugins": [
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
-    "prepublish": "npm run build && babel src --out-dir lib"
+    "prepublish": "npm run build && BABEL_DISABLE_CACHE=1 babel src --out-dir lib"
   },
   "files": [
     "*.md",
@@ -42,6 +42,7 @@
     "babel-core": "^6.22.1",
     "babel-loader": "^6.2.10",
     "babel-plugin-transform-runtime": "^6.22.0",
+    "babel-plugin-webpack-loaders": "^0.8.0",
     "babel-preset-es2015": "^6.22.0",
     "cross-env": "^3.1.4",
     "eslint": "^3.14.0",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "itowns",
   "version": "0.0.0-alpha",
   "description": "A JS/WebGL framework for 3D geospatial data visualization",
-  "main": "dist/itowns.js",
+  "main": "lib/Main.js",
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc src/Core/Commander/Interfaces/ApiInterface/*",
@@ -10,11 +10,12 @@
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build && babel src --out-dir lib"
   },
   "files": [
     "*.md",
     "dist",
+    "lib",
     "doc",
     "examples"
   ],
@@ -27,16 +28,22 @@
     "url": "https://github.com/iTowns/itowns2/issues"
   },
   "homepage": "https://itowns.github.io/",
+  "dependencies": {
+    "babel-runtime": "^6.20.0",
+    "custom-event": "^1.0.1",
+    "es6-promise": "^4.0.5",
+    "js-priority-queue": "^0.1.5",
+    "jszip": "^3.1.3",
+    "three": "^0.84.0",
+    "whatwg-fetch": "^2.0.2"
+  },
   "devDependencies": {
     "babel-cli": "^6.22.2",
     "babel-core": "^6.22.1",
     "babel-loader": "^6.2.10",
     "babel-plugin-transform-runtime": "^6.22.0",
     "babel-preset-es2015": "^6.22.0",
-    "babel-runtime": "^6.20.0",
     "cross-env": "^3.1.4",
-    "custom-event": "^1.0.1",
-    "es6-promise": "^4.0.5",
     "eslint": "^3.14.0",
     "eslint-config-airbnb-base": "^11.0.1",
     "eslint-import-resolver-webpack": "^0.8.1",
@@ -45,16 +52,9 @@
     "eslint-loader": "^1.6.1",
     "eslint-plugin-import": "^2.2.0",
     "imports-loader": "^0.7.0",
-    "js-priority-queue": "^0.1.5",
     "jsdoc": "^3.4.3",
-    "jszip": "^3.1.3",
-    "paralleljs": "^0.2.1",
     "raw-loader": "^0.5.1",
-    "simd": "^2.0.0",
-    "string_format": "^0.0.6",
-    "three": "^0.84.0",
     "webpack": "^1.14.0",
-    "webpack-dev-server": "^1.16.2",
-    "whatwg-fetch": "^2.0.2"
+    "webpack-dev-server": "^1.16.2"
   }
 }
--- a/webpack-babel.config.js
+++ b/webpack-babel.config.js
@@ -1,0 +1,19 @@
+var path = require('path');
+
+module.exports = {
+  output: {
+    libraryTarget: 'commonjs2',
+    umdNamedDefine: true
+  },
+  module: {
+    loaders: [
+       {
+        test: /\.glsl$/,
+        include: [
+          path.resolve(__dirname, 'src'),
+        ],
+        loader: 'raw'
+      },
+    ],
+  },
+};
--- a/webpack-babel.config.js
+++ b/webpack-babel.config.js
@@ -7,7 +7,8 @@ module.exports = {
   },
   module: {
     loaders: [
-       {
+      // please consider modifying corresponding loaders in webpack.config.js too
+      {
         test: /\.glsl$/,
         include: [
           path.resolve(__dirname, 'src'),
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -43,13 +43,17 @@ module.exports = {
           path.resolve(__dirname, 'utils')
         ],
         loader: 'babel',
+        // Please consider modifying .babelrc too
+        // .babelrc is used for transpiling src/ into lib/ in the prepublish
+        // phase, see package.json
         query: {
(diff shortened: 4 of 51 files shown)


webpack.config.js
```diff
@@ -57,7 +49,7 @@ module.exports = {
     ]
   },
   resolve: {
-    root: path.resolve(__dirname, 'src'),
+    root: path.resolve(__dirname, ''),
```
tbroyer: Should `resolve.root` be deleted instead? (or set to an empty array) AFAICT, we now want "module names" to only be resolved as Node modules.

webpack.config.js
```diff
@@ -35,14 +35,6 @@ module.exports = {
         ],
         loader: 'babel'
       },
-       {
-        test: /\.glsl$/,
```
tbroyer: AIUI, this is no longer need because it's now taken care of by the `babel-plugin-webpack-loaders`, but the README for that project says that it shouldn't be used during a Webpack compilation: https://github.com/istarkov/babel-plugin-webpack-loaders#warning (that plugin runs Webpack synchronously during the Babel transformation, which is itself run by Webpack for bundling)

Maybe it'd be better to have different Babel configurations depending on the usage: when calling babel-cli to produce CommonJS, then use `.babelrc` that calls the babel-plugin-webpack-loader with the babel-webpack.config.js; and when calling webpack, use the raw-loader in webpack.config.js and a different babel config that does not use the babel-plugin-webpack-loader (by inlining the babel config in the babel-loader's `query` and setting `babelrc: false`).

tbroyer (commented):
I really like your work to transform import paths; this should probably be split out into its own commit (along with the `resolve.root` change –see comment below about it though), possibly even its own PR.

autra:
> [...] transform import paths; this should probably be split out into its own commit (along with the resolve.root change –see comment below about it though), possibly even its own PR.

I would let this changeset into this Pr, because the reason I'm doing this is solely because of the lib/ stuff. Otherwise, we wouldn't care and would stay with absolute import, as they are easier to deal with internally, right? I'm putting them into their own commit.

webpack.config.js
autra: yep, I've seen that but as we are using it for dev and I didn't see any noticeable performance hit, I thought it was not worth the effort. Also I wanted prod and dev build to stay as close as possible.

I did the measurement again, and this time my build is 6sec longer! So I guess the warning is real :-) I'm amending this right now.

autra:
@tbroyer I've updated my PR with your required change. Pro: build is now as fast as before. Cons: now, we have 2 conf duplications : babel conf is in `.babelrc` and `webpack.config.js`, glsl loader conf is in `webpack.config.js` and `webpack-babel.config.js`. Do you have an idea to mitigate that?

tbroyer:
Well, Webpack configs are JS so they can `require()` other files; this means `webpack.config.js` could possibly `require('./.babelrc')` (no sure how it'll deal with the "JSON5" syntax; worst case would be using `require('json5')` and explicitly parsing the `.babelrc` file) and then _massage_ the data to remove the specific values we don't want to use in Webpack. And similarly `webpack.config.js` could `require('webpack-babel.config.js')` (or an intermediate, shared config) to merge it in. That'd remove duplications but would likely make the configuration more complex. Not sure it'd be worth it (a comment in the files telling you to keep files in sync could possibly work better from a maintenance point of view)

autra:
@tbroyer I've added some comments. I'm not very comfortable with that, as I always find that nobody reads comments I've ever left in code, but the alternative is meh, so let's try that out.

.babelrc
```diff
@@ -1,4 +1,7 @@
 {
-  "presets": ["es2015"],
-  "plugins": ["transform-runtime"]
+    "presets": ["es2015"],
```
tbroyer, line 3: Add a comment here ([this is _JSON5_](https://babeljs.io/docs/usage/babelrc/) yay!) that this file is only used when transpiling src/ to lib/ in the prepublish phase, and should be kept in sync with the babel-loader configuration (`query`) in webpack.config.js?

package.json
```diff
@@ -2,19 +2,20 @@
...
+  "main": "lib/Main.js",
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc src/Core/Commander/Interfaces/ApiInterface/*",
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
     "start": "webpack-dev-server -d --inline --hot",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build && BABEL_DISABLE_CACHE=1 babel src --out-dir lib"
```
tbroyer: Should this run `npm run test` instead of `npm run build`?

webpack-babel.config.js
```diff
@@ -0,0 +1,20 @@
+var path = require('path');
+
+module.exports = {
+  output: {
+    libraryTarget: 'umd',
```
tbroyer: Why UMD? The target is NodeJS here, so CommonJS (from the [`babel-plugin-webpack-loaders` examples](https://github.com/istarkov/minimal-example-for-babel-plugin-webpack-loaders/blob/c901553c80024b4e7fdf961228b0ef55de53f230/webpack.config.js#L4), this “NEEDS TO” be `commonjs2`)

package.json
autra: Or build should be `npm run lint && webpack -p && npm run doc`?

webpack-babel.config.js
autra: Changing...

autra: Actually answering the `why UMD` question: because it works and because we use UMD in `webpack.config.js`. But let's follow the instructions you linked.

autra: Btw, are we sure the target is nodejs? ppl will `import` or `require` our sources from a browser right? Or maybe I misunderstood something?

tbroyer: Files from `lib/` will be used like any other NPM module; i.e. passed through Webpack, Browserify, Rollup, or equivalent (or in case there's a very good browser-environment emulation –jsdom?– that can do WebGL, then it could be NodeJS proper), which will rewrite the CommonJS imports.

People using iTowns by other means will use the output from Webpack (in `dist/`, as a UMD module, so compatible with RequireJS/AMD **and** direct use in a `<script src=>`, and also incidentally compatible with CommonJS)

autra:
@tbroyer updated. WDYT?

package.json
```diff
@@ -2,19 +2,20 @@
...
+  "main": "lib/Main.js",
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc src/Core/Commander/Interfaces/ApiInterface/*",
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
     "start": "webpack-dev-server -d --inline --hot",
-    "prepublish": "npm run build"
+    "prepublish": "npm run lint && npm run doc && npm run build && BABEL_DISABLE_CACHE=1 babel src --out-dir lib"
```
tbroyer: Where does this (`npm run doc`) generate the documentation? shouldn't it rather be `npm run doclint` instead (only to ensure the jsdoc is OK)?  
Do we package the doc in the NPM module? (and if so, should we? I'm not quite sure…)

package.json
```diff
@@ -2,19 +2,20 @@
...
   },
   "files": [
     "*.md",
     "dist",
+    "lib/*",
```
tbroyer: Why use `/*` here? other directories are listed as simple entries without `/*` suffix.

tbroyer (approved):
Good to go, with a couple small comments. Feel free to not address them.

package.json
autra: > Where does this (npm run doc) generate the documentation?

I thought it would generate it in `doc`, but actually it does it in `out`, except in `autodeploy.sh` for publishing to gh_pages.

So we don't actually package it. I'm not sure we need to, so I'm going to remove this `npm run doc` from the prepublish phase.

> shouldn't it rather be npm run doclint instead (only to ensure the jsdoc is OK)?

it is already run in the `test` phase (so run for each PR).  If we don't package the doc, I guess we don't need to lint it there.

Thanks for the remark about the doc, I now thing we should just do:
`npm run build && BABEL_DISABLE_CACHE=1 babel src --out-dir lib`. 

WDYT?

package.json
autra: Removing it.

package.json
autra: So this conversation came back, as I reverted some change. I actually don't know :-) 

Maybe test should run `lint` and `doclint` only? And then prepublish could run `build`, `test` and babel? But as we run lint and tests from a PR, it's not very important for it to do so (though it gives another layer of security of course). WDYT?

autra:
@tbroyer Only the `prepublish` script to finalize :-)

package.json
```diff
@@ -27,31 +28,30 @@
     "url": "https://github.com/iTowns/itowns2/issues"
   },
   "homepage": "https://itowns.github.io/",
+  "dependencies": {
```
gchoqueux, line 31: There is a return on the `dependencies`? 
[https://github.com/iTowns/itowns2/pull/241/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L23](https://github.com/iTowns/itowns2/pull/241/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L23)

gchoqueux (approved):

gchoqueux (commented):
change commit name 
> type (scope): subject

package.json
tbroyer: Yes, that's exactly what this PR is about: consuming the NPM module like any other library, with its own dependencies, rather than (necessarily) as a bundle (and then bundle things yourself using webpack or browserify; and properly deduplicate those dependencies used by your own code and by iTowns)

package.json
tbroyer: OK, let's use `npm run build` (I kind-of like having everything properly linted and tested before releasing, which is why I proposed using `npm run test`; but then we'd probably want to go the last mile of also checking that the current checkout is "clean", and properly tagged as as the appropriate version, etc. Those might be better in a "release script" than in the `prepublish` script)

tbroyer (approved):

gchoqueux:
Everyone agrees to merge this one?

autra:
Rebased 

> Everyone agrees to merge this one?

I don't know, but at least nobody seems to be against it ;-)

PR #250, opened 2017-02-15
Fix realistic lighting

gchoqueux:


src/Scene/LayersConfiguration.js
```diff
@@ -18,6 +18,12 @@ function LayersConfiguration() {
     this.geometryLayers = [];
     this.colorLayers = [];
     this.elevationLayers = [];
+    this.lightingLayers = [];
```
peppsac, line 21: Why don't you store the light position in `Globe` class?

src/Scene/NodeProcess.js
```diff
@@ -85,6 +85,7 @@ NodeProcess.prototype.subdivideNode = function subdivideNode(node, camera, param
                 /* specific params */
                 bbox: bboxes[i],
                 type: quadtree.type,
+                light: params.layersConfig.lightingLayers[0],
```
peppsac: I would instead do this 10 lines below:
```js
quadtree.scheduler.execute(command).then((child) => 
    child.material.setLightingOn(params.light.enable);
    child.material.uniforms.lightPosition.value = params.light.position;
```
this way TileProvider doesn't need to be aware of Lighting parameters

src/Scene/LayersConfiguration.js
gchoqueux: because it's sun position

peppsac: You should rename it sunPosition then :)

peppsac: On a more serious tone: even if layers are losely defined actually, I don't believe their purpose is to store sun position.

My suggestion to store it in `Globe` is simply based on the fact that this class owns all globe related parameters/var.

gchoqueux: The lighting isn't only linked to the globe

peppsac: Currently it *is* a globe-only feature.
But while I understand that we might want lighting on other scenario this doesn't mean it makes sense to store lighting as a layer.

If you don't like storing this in `Globe`, what about these 2 proposals:
  * store lighting position in `Scene` ?
  * store lighting position in imagery layers attributes (alongside `visible`, `frozen` attributes etc) ?

peppsac: (the last option allow to enable|disable lighting per-layer which might be useful)

gchoqueux: The `lightingLayer` allows you to enable or disable all the lights for all other layers

Furthermore each light independently (Moreover each light can be enable)

By adding an `enableLighting` option for each layer and mesh, it is possible to disable the illumination independently for each layer or mesh.

peppsac: The shader code doesn't support more than 1 light. So planning for more lights might be overkill.

Even if you want to make sure we support more than 1 light,  all of this can still be achieved without adding a `lightingLayer` (e.g: an array of lights in `Scene`).

Can you state why you believe that defining "light layers" is the best place for this?

peppsac: To be perfectly clear: you PR is about "fixing realistic lighting" so I'd prefer you to take the shortest path to do this, instead of planning for the future.

I won't oppose you any longer on this point and if you insist I don't mind merging this code - especially since PR #239 will rework layers anyway.

gchoqueux: Ok, we'll see the #239  review.
and now?

src/Core/Commander/Providers/TileProvider.js
```diff
@@ -30,8 +30,7 @@ TileProvider.prototype = Object.create(Provider.prototype);
 
 TileProvider.prototype.constructor = TileProvider;
 
-TileProvider.prototype.preprocessLayer = function preprocessLayer(/* layer*/) {
-    /* no-op */
+TileProvider.prototype.preprocessDataLayer = function preprocessDataLayer(/* layer*/) {
```
peppsac: Actually let's remove this function completely since it's useless

src/Scene/LayersConfiguration.js
peppsac: I still think it's broken to add a layer simply to be able to initialize a uniform value.
Note that this is the same problem @autra has in its PR debug ("how to initialize the showOutline value?")...

Anyway, I believe #239 will allow to make something simpler, such as allowing the user to define a function called each time a new time a new tile is created.

So I propose that we merge both PR as is, and revisit the 2 issues in the refactoring PR.
Makes sense?

gchoqueux: so I must review the 239

peppsac: Not really.
We can simply acknowledge that the current implementation is not perfect to implement these features. And then fix these issues in PR 239 (or later).

gchoqueux:
rebased

gchoqueux:
rebased

peppsac:
There's an issue when you zoom:
capture d ecran de 2017-03-02 14-32-04
capture d ecran de 2017-03-02 14-32-11

If you turn lighting off and on again, the problem disappear.

src/Renderer/BasicMaterial.js
```diff
@@ -51,7 +51,7 @@ function BasicMaterial(color) {
         uuid: { value: 0 },
         debug: { value: false },
         selected: { value: false },
-        lightOn: { value: true },
+        enableLighting: { value: true },
```
peppsac: This uniform is a status, not an action. So `lightingOn` or `lightingEnabled` would be a better name.

src/Scene/Quadtree.js
```diff
@@ -79,6 +79,7 @@ Quadtree.prototype.requestNewTile = function requestNewTile(geometryLayer, bbox,
         bbox,
         type: this.type,
         level: 0,
+        light: lightingLayer,
```
peppsac, line 82: I still don't like this change, but PR #239 will help improving this  :)

peppsac (changes requested):

peppsac (approved):

PR #251, opened 2017-02-15
feature: allow command to indicate they don't need a redraw

peppsac:
There's no need to cause a redraw if we know that the visual
result won't change.


gchoqueux (commented):
OK

gchoqueux (approved):

PR #253, opened 2017-02-15
refactor: improve GeoCoordinate and force explicit usage of units

peppsac:
Explicit control of units allows to quickly uncover bugs (expecting Degrees
but get Radians etc).


examples/layers/JSONLayers/Region.json
```diff
@@ -7,7 +7,8 @@
 	"style"      : "",
 	"projection" : "EPSG:4326",
 	"transparent" : true,
-	"bbox"      : [-61.80983300000002, -21.38963079935857, 55.83665389999998, 51.0890011990783],
+    "unit"      : 1,
```
gchoqueux: It is more explicit to write
`"unit"      : "degree"`

I prefer to consider that by default the bbox is in degree and remove this option from Region.json

src/Core/Geographic/GeoCoordinate.js
```diff
@@ -18,60 +21,73 @@ export const UNIT = {
...
 }
 
 GeoCoordinate.prototype.constructor = GeoCoordinate;
 
 GeoCoordinate.prototype.longitude = function longitude(unit) {
-    return getCoordinateValue(unit, this.coordinate, COORD.LONG);
+    return convert(this.unit, unit, this.coordinate[COORD.LONG]);
 };
 
+GeoCoordinate.prototype.x = GeoCoordinate.prototype.longitude;
```
gchoqueux: There is a semen problem and therefore of object :

There are several types of coordinates (see link) : 

- [geocentric rectangular coordinates](https://en.wikipedia.org/wiki/ECEF) X,Y, Z (center is ellipsoid center) which should be the Cartesian coordinates of three.js (not actually presently due to axis inversions #223)
- [Geographic coordinate](https://en.wikipedia.org/wiki/Geographic_coordinate_system) like `GeoCoordinate` (longitude, latitude and altitude)
- [planes coordinates](https://en.wikipedia.org/wiki/Geographic_coordinate_system#Cartesian_coordinates)  ENU and NED (not yet in iTowns)

So, there's no `x, y or z` in geographic coordinate

and (x,y,z) are geocentric or planes?

src/Core/Geographic/GeoCoordinate.js
```diff
@@ -18,60 +21,73 @@ export const UNIT = {
...
 GeoCoordinate.prototype.latitude = function latitude(unit) {
-    return getCoordinateValue(unit, this.coordinate, COORD.LAT);
+    return convert(this.unit, unit, this.coordinate[COORD.LAT]);
 };
 
+GeoCoordinate.prototype.y = GeoCoordinate.prototype.latitude;
+
 GeoCoordinate.prototype.setLatitude = function setLatitude(latitude, unit) {
-    setCoordinateValue(unit, this.coordinate, COORD.LAT, latitude);
+    this.coordinate[COORD.LAT] = convert(unit, this.unit, latitude);
```
gchoqueux: no default unit? and why force unit?

src/Core/Math/Ellipsoid.js
```diff
@@ -21,8 +22,8 @@ function Ellipsoid(size) {
 }
 
 Ellipsoid.prototype.geodeticSurfaceNormalCartographic = function geodeticSurfaceNormalCartographic(coordCarto) {
-    var longitude = Math.PI * 2 - coordCarto.longitude();
-    var latitude = coordCarto.latitude();
+    var longitude = Math.PI * 2 - coordCarto.longitude(UNIT.RADIAN);
```
gchoqueux, line 25: no default unit?

gchoqueux (changes requested):

src/Core/Geographic/GeoCoordinate.js
```diff
@@ -18,60 +21,73 @@ export const UNIT = {
...
+
 GeoCoordinate.prototype.setLongitude = function setLongitude(longitude, unit) {
-    setCoordinateValue(unit, this.coordinate, COORD.LONG, longitude);
+    this.coordinate[COORD.LONG] = convert(unit, this.unit, longitude);
+    return this;
 };
 
 GeoCoordinate.prototype.latitude = function latitude(unit) {
-    return getCoordinateValue(unit, this.coordinate, COORD.LAT);
+    return convert(this.unit, unit, this.coordinate[COORD.LAT]);
```
gchoqueux: > I don't like having a this.angleUnit state + longitude()/latitude() methods that depends on it ([comment in 178](https://github.com/iTowns/itowns2/pull/178#issuecomment-243469528 ))

I thought you did not like doing too much like that?

src/Core/Geographic/GeoCoordinate.js
peppsac: Ok, makes sense.
But then I don't want a `GeoCoordinate` object, I want a `Coordinates` objects.
Internally it'll be one of the 3 listed above but it would all on-the-fly conversion to others if needed.

examples/layers/JSONLayers/Region.json
peppsac: Why `degree` by default? Is this the most common case?

src/Core/Geographic/GeoCoordinate.js
peppsac: See commit description: `Explicit control of units allows to quickly uncover bugs (expecting Degrees but get Radians etc)`

src/Core/Geographic/GeoCoordinate.js
peppsac: This comment was refering to these methods: `c.longitude()` which would return longitude in the *current unit*, but this current unit might be unknown to the caller.

I prefer `c.longitude(my_unit)` where the caller control the unit of the result.

examples/layers/JSONLayers/Region.json
gchoqueux: `"projection" : "EPSG:4326"` 

The projection defines the units

src/Core/Geographic/GeoCoordinate.js
gchoqueux: (x,y,z) are geocentric or planes?

peppsac: geocentric at the moment.

examples/layers/JSONLayers/Region.json
peppsac: Well projection covers a lot more thing.
Actually I was about to remove this attribute since we don't use it.
And then, *later*, restore it but to use it completely (using proj4 to transform from 1 proj to another).

src/Core/Geographic/GeoCoordinate.js
gchoqueux: Actually **geocentric rectangular coordinates** use three.Vector3. Why create a another object?

examples/layers/JSONLayers/Region.json
gchoqueux: If you want to know the unit of the angles the standard says that you have to use the type of projection. Therefore it's not necessary to add `unit`.

src/Core/Geographic/GeoCoordinate.js
peppsac: My point is to have a single itowns `Coordinates` object used through the whole codebase. Why ?

Because it allows to mix sources (e.g WMS and WMTS server) who use different projection system transparently.

Currently the only example of this in itowns codebase is WMS, and it only supports using a WMS source that is in GeoGraphic Coords in `Degrees` and map it on a GeoGraphic Coords geometry using `Radians`.

examples/layers/JSONLayers/Region.json
peppsac: Does itowns codebase support projection|crs anywhere? (hint: the projection defined in Region.json is *wrong*, according to the `GetCapabilities` it should be `<CRS>EPSG:3857</CRS>`).

But OK, I'll remove `unit` and add a list of know crs and deduce the unit from the projection.

src/Core/Geographic/GeoCoordinate.js
gchoqueux: the WMS protocol says unit is defined in EPSG Projection. It should be possible to take it into account. 

`Coordinates` would be exactly what?

peppsac: See https://github.com/iTowns/itowns2/pull/253#discussion_r101466111

gchoqueux: Coordinates will be at the same time a cartesian object and a spherical object, it is not common in terms of object. Generally the basic objects are separated, for example :

**THREE**

[`THREE.Vector3`](https://threejs.org/docs/?q=vect#Reference/Math/Vector3) and [`THREE.Spherical`](https://threejs.org/docs/?q=spher#Reference/Math/Spherical) 
 
and some specific methods to set object from another : 
`THREE.Vector3.setFromSpherical(sperical)`
`THREE.Spherical.setFromVector3(vector3)`

**Cesium**

[`Cesium.Cartesian3`](https://cesiumjs.org/Cesium/Build/Documentation/Cartesian3.html) and 
[`Cesium.Cartographic`](https://cesiumjs.org/Cesium/Build/Documentation/Cartographic.html)

Methods

`Cesium.Cartesian3.fromCartesian(cartesian, ellipsoid, result)`
`Cesium.Cartesian3.fromRadians(longitude, latitude, height, ellipsoid, result)`


There is `Projection` object  (Which is a bit in _iTowns_) in Cesium to convert object 

[GeographicProjection](https://cesiumjs.org/Cesium/Build/Documentation/GeographicProjection.html)

peppsac: We don't need to build raw Vector3 and Spherical because three.js already have them.

The Coordinates object I'm proposing is an object that store coordinates (3 floats for instance) **AND** the crs (ie: how to interpret the coordinates).

It's more expressive to use: `Coordinates(12, 45, 'wgs84')` than it is to use `Spherical(12, 45)`, or `Coordinates(12345, 34444, 5566, 'epsg:3946')` than `Vector3(12345, 34444, 5566)`.

And you get the ability to project in another crs in a transparent fashion (using `proj4.js` under the hood):
```js
Coordinates(12345, 34444, 5566, 'epsg:3946').longitude('wgs84')
```

gchoqueux: > We don't need to build raw `Vector3` and `Spherical` because three.js already have them

Yes, it does not help us to add `Vector3` and `Spherical`. They are in the commentary to illustrate the concepts of object-oriented programming. 

> `Coordinates(12, 45, 'wgs84')` and `Coordinates(12345, 34444, 5566, 'epsg:3946')`   

How do you get different constructor signatures?
you must write  `Coordinates(12, 45, 0,  'wgs84')`  and set convention altitude

> `Coordinates(12345, 34444, 5566, 'epsg:3946')`

On the other hand there is no altitude convention in epsg: 3946

> Coordinates(12345, 34444, 5566, 'epsg:3946').longitude('wgs84')

You can give an example of Cartesian geocentric?



Overall Coordinate seems to be fine

autra: > How do you get different constructor signatures?

Using [Rest Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) (as we are using ES6, better use them instead of `arguments`). I would then suggest to put the CRS as first argument. Signature would be 
```js
function Coordinates(crs, ...coordinates)
```

gchoqueux: OK

It would be necessary to rename `GeoCoordinates` in `Coordinates`

For later, there will remain the conventions of altitude
Currently it represents the distance with the ellipsoid

You can give just an example of Cartesian geocentric?

gchoqueux:
Uncaught Error: Cannot build a GeoCoordinate without a unit
--> at Projection.cartesianToGeo (Projection.js:190)

peppsac:
So, here's a new commit renaming `GeoCoordinates` -> `Coordinates`.
It also try to fix all usages of `Coordinates`.

This is not the final version. Current open questions:
* maybe we could limit to crs with a EPSG code, and therefore use a simple `int` instead of a `string` to declare which crs is used.
* I couldn't find the correct EPSG code for WGS84 using radians units so I hacked one... any better suggestions?

examples/layers/JSONLayers/Region.json
gchoqueux: no change?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -353,17 +352,12 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
...
     var debugMode = false;
 
-    var ellipsoid = new Ellipsoid({
-        x: 6378137,
-        y: 6356752.3142451793,
-        z: 6378137,
-    });
-
-    var coordinate = new GeoCoordinate().copy(coordCarto, UNIT.DEGREE);
+    var coordinate = new Coordinates('EPSG:4326',
```
gchoqueux: Not to burden the writings
add somes facades :

```js
import { EPSG_4326 } from '../../../Geographic/Coordinates';
var coordinate = new EPSG_4326(coordCarto.longitude, coordCarto.latitude, coordCarto.altitude);
```

peppsac:
Pushed a few more cleaning commits.

The last commit is also an example of how we could integrate `proj4` and the possibilities it opens (I added a `EPSG:3946` wms layer to the globe).

examples/layers/JSONLayers/Region.json
peppsac: Fixed :)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -668,7 +664,8 @@ ApiGlobe.prototype.setSceneLoaded = function setSceneLoaded() {
 ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
     eventCenter.oldCenter = this.getCenter();
-    const position3D = this.scene.getEllipsoid().cartographicToCartesian(new GeoCoordinate(coordinates.longitude, coordinates.latitude, 0, UNIT.DEGREE));
+    const position3D = new Coordinates('EPSG:4326', coordinates.longitude, coordinates.latitude, 0)
+        .as('EPSG:4328').xyz();
```
gchoqueux: [4328](http://spatialreference.org/ref/epsg/4328/) is deprecated use [4978](http://spatialreference.org/ref/epsg/4978/)

what returns `position3D.longitude(DEGREE) `?

src/Core/Geographic/Coordinates.js
```diff
@@ -0,0 +1,170 @@
+/**
```
gchoqueux, line 1: Git considers it as an addition as a renaming

src/Core/Geographic/Coordinates.js
```diff
@@ -0,0 +1,170 @@
...
+
+export const UNIT = {
+    RADIAN: 0,
+    DEGREE: 1,
+    METER: 2,
+};
+
+export const supportedCRS = {
+    'EPSG:4326': UNIT.DEGREE,
+    'EPSG:4326R': UNIT.RADIAN,
```
gchoqueux: 4326R, there's no such thing? 
just for radian?

src/Core/Geographic/Coordinates.js
```diff
@@ -0,0 +1,170 @@
...
+ * Generated On: 2015-10-5
+ * Class: Coordinates
+ * Description: Coordonnées cartographiques
+ */
+
+import * as THREE from 'three';
+import mE from '../Math/MathExtended';
+import Ellipsoid from '../Math/Ellipsoid';
+
+export function ellipsoidSizes() {
```
gchoqueux: it's better name : `ellipsoidWGS84Rayons`
and x, y and z are rayons on WGS84 ellipsoid

src/Globe/Globe.js
```diff
@@ -55,15 +54,15 @@ function Globe(ellipsoid, gLDebug) {
...
+    this.atmosphere = this.NOIE ? new Atmosphere(ellipsoid) : undefined;
     this.clouds = new Clouds();
 
     var material = new BasicMaterial(new THREE.Color(1, 0, 0));
 
     var geometry = new THREE.SphereGeometry(5);
     var batiment = new THREE.Mesh(geometry, material);
 
-    var position = this.ellipsoid.cartographicToCartesian(new GeoCoordinate(0, 48.87, 200, UNIT.DEGREE));
+    var position = new Coordinates('EPSG:4326', 0, 48.87, 200).as('EPSG:4328');
```
gchoqueux: it doesn't need `xyz()`

gchoqueux (changes requested):

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
peppsac: It throws an error (see https://github.com/iTowns/itowns2/pull/253/files#diff-7c37ee347b0423d12e6624421e03f4caR145)

src/Core/Geographic/Coordinates.js
peppsac: Yes, hence my comment https://github.com/iTowns/itowns2/pull/253#issuecomment-282790796.
Any better suggestion ?
The  alternative I had in mind is to simply not support EPSG:4326 + radians. In this case the user will have to convert to degrees on its own.

src/Globe/Globe.js
peppsac: Yes it does (well I can even remove this line since `position` is overwritten in the next line)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
peppsac: I pushed a commit with a slightly different syntax:
```js
var coordinate = new C.EPSG_4326(lon, lat, alt);
````

File: src/Core/Math/Ellipsoid.js (hunk 4 above)
Comment (peppsac) [reply src/Core/Math/Ellipsoid.js:25]: Nope, because this function only supports `radians`

File: src/Core/Geographic/Coordinates.js [hunk 12]
```diff
@@ -18,6 +19,9 @@ export function ellipsoidSizes() {
 
 const ellipsoid = new Ellipsoid(ellipsoidSizes());
 
+proj4.defs('EPSG:3946',
```
Jeremy-Gaillard: Suggestion:
- Remove EPSG:3946 from Coordinates.js
- Add a function "defineCrs(name, unit, projString)"
- Call this function in index.html

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: ok nice

src/Core/Commander/Providers/GpxUtils.js
```diff
@@ -19,15 +18,15 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
 }
 
-function _gpxPtToCartesian(pt, ellipsoid) {
+function _gpxPtToCartesian(pt) {
     var longitude = Number(pt.attributes.lon.nodeValue);
     var latitude = Number(pt.attributes.lat.nodeValue);
     var elevation = Number(pt.getElementsByTagName('ele')[0].childNodes[0].nodeValue);
 
-    return ellipsoid.cartographicToCartesian(new GeoCoordinate(longitude, latitude, elevation, UNIT.DEGREE));
+    return new Coordinates('EPSG:4326', longitude, latitude, elevation).as('EPSG:4328').xyz();
```
gchoqueux: EPSG:4328 is deprecated

peppsac: Done

src/Core/Geographic/Coordinates.js
peppsac: I'm not sure I want to wrap `proj4` at all.
I think I'd prefer the user to directly call `proj4.defs(...)` and transform `supportedCRS` to a function capable of querying `proj4` definition.

peppsac: I implemented this change in commit b2de20e 
For now I've temporarly moved `proj4.defs(...)` in ApiGlobe, but it could be in `index.html`

gchoqueux:
For one bbox `WGS84`, in green the projection in `EPSG:3946` in red the projection of min and max BBOX. In the present state, [the bbox in request WMS](https://github.com/iTowns/itowns2/pull/253/files#diff-4d5db84d9c2e552c5b657933784391daR44) is too small. 

In blue box is the correct bbox, to obtain the entire projected area. It is then necessary to re-project the result in the WGS84 tile

wgs84_lambert

src/Core/Geographic/Coordinates.js
```diff
@@ -0,0 +1,213 @@
...
+
+// Only support explicit conversions
+function _convert(coordsIn, newCrs) {
+    if (newCrs === coordsIn.crs) {
+        return new Coordinates(newCrs, coordsIn._values);
+    } else {
+        // TODO: of course we don't want to keep this code.
+        // This is mostly a POC until we move to something like proj4.js
+        if (newCrs === 'EPSG:4978') {
+            const cartes = ellipsoid.cartographicToCartesian(coordsIn);
```
gchoqueux: You consider that the unit is in radian

src/Core/Geographic/Coordinates.js
gchoqueux: `EPSG: 4326R` does not exist, as you said, try to avoid this notation.

maybe add `unit` properties : 

```js
    assertCrsIsValid(crs);
    this.crs = crs;
    // Set unit by default unit 
    this.unit = supportedCRS[this.crs];
    for (let i = 0; i < coordinates.length; i++) {
        this._values[i] = coordinates[i];
 }
}
Coordinates.prototype.unit = function unit() {
    return this.unit;
};
Coordinates.prototype.asUnit = function asUnit(unit) {
    const cC = this.clone();
    cC.unit = unit;
    return cC;
};
Coordinates.prototype.clone = function clone() {
    return new Coordinates(this.crs, this._values[0], this._values[1], this._values[2]);
};
```

and use like this 

```js
const coordinates = new Coordinates('EPSG:4326', 10, 40, 0).asUnit(UNIT.RADIAN);
```

peppsac:
Thanks for the details about `WGS84` and `EPSG:3946`.
I won't modify the code right now because it's quite a big change, and the `EPSG:3946` was more an example of what could be achievable.

src/Core/Geographic/Coordinates.js
peppsac: Ooop, indeed. Fixing this.

peppsac: Err nope actually. `ellipsoid.cartographicToCartesian` uses `longitude(RADIAN)` and `latitude(RADIAN)` so we're good.

gchoqueux (approved):

gchoqueux:
ok, it's good

PR #254, opened 2017-02-16
Tile mesh refactoring

Jeremy-Gaillard:
Small changes to tile mesh management.

- Removed dead code and useless attributes
- Reworked frustum culling
- Tile geometry building moved from TileMesh  to TileProvider

### DIFF
--- a/src/Core/Commander/Providers/TileProvider.js
+++ b/src/Core/Commander/Providers/TileProvider.js
@@ -16,6 +16,7 @@
 import Provider from 'Core/Commander/Providers/Provider';
 import Projection from 'Core/Geographic/Projection';
 import BuilderEllipsoidTile from 'Globe/BuilderEllipsoidTile';
+import TileGeometry from 'Globe/TileGeometry';
 
 function TileProvider(ellipsoid) {
     Provider.call(this, null);
@@ -46,7 +47,9 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
         segment: 16,
     };
 
-    var tile = new command.type(params, this.builder);
+    const geometry = new TileGeometry(params, this.builder);
+
+    var tile = new command.type(geometry, params);
 
     tile.setUuid(this.nNode++);
     tile.link = parent.link;
--- a/src/Core/Commander/Providers/TileProvider.js
+++ b/src/Core/Commander/Providers/TileProvider.js
@@ -60,6 +60,8 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
     parent.add(tile);
     tile.updateMatrix();
     tile.updateMatrixWorld();
+    tile.OBB().parent = tile;   // TODO: we should use tile.add(tile.OBB())
+    tile.OBB().update();
 
     return Promise.resolve(tile);
 };
--- a/src/Globe/TileMesh.js
+++ b/src/Globe/TileMesh.js
@@ -15,7 +15,6 @@
  * @returns {EllipsoidTileMesh_L20.TileMesh}
  */
 import NodeMesh from 'Renderer/NodeMesh';
-import TileGeometry from 'Globe/TileGeometry';
 import BoundingBox from 'Scene/BoundingBox';
 import * as THREE from 'three';
 import LayeredMaterial, { l_ELEVATION } from 'Renderer/LayeredMaterial';
@@ -23,7 +22,7 @@ import GlobeDepthMaterial from 'Renderer/GlobeDepthMaterial';
 import MatteIdsMaterial from 'Renderer/MatteIdsMaterial';
 import RendererConstant from 'Renderer/RendererConstant';
 
-function TileMesh(params, builder, geometryCache) {
+function TileMesh(geometry, params) {
     // Constructor
     NodeMesh.call(this);
 
@@ -33,7 +32,7 @@ function TileMesh(params, builder, geometryCache) {
     this.level = params.level;
     this.bbox = params.bbox || new BoundingBox();
 
-    this.geometry = geometryCache || new TileGeometry(params, builder);
+    this.geometry = geometry;
     this.normal = params.center.clone().normalize();
 
     // TODO Why move sphere center
--- a/src/Globe/TileMesh.js
+++ b/src/Globe/TileMesh.js
@@ -10,8 +10,6 @@
  * @param {type} TileGeometry
  * @param {type} BoundingBox
  * @param {type} THREE
- * @param {type} OBBHelper
- * @param {type} SphereHelper
  * @param {type} LayeredMaterial
  * @param {type} GeoCoordinate
  * @returns {EllipsoidTileMesh_L20.TileMesh}
@@ -20,8 +18,6 @@ import NodeMesh from 'Renderer/NodeMesh';
 import TileGeometry from 'Globe/TileGeometry';
 import BoundingBox from 'Scene/BoundingBox';
 import * as THREE from 'three';
-import OBBHelper from 'Renderer/ThreeExtended/OBBHelper';
-import SphereHelper from 'Renderer/ThreeExtended/SphereHelper';
 import LayeredMaterial, { l_ELEVATION } from 'Renderer/LayeredMaterial';
 import GlobeDepthMaterial from 'Renderer/GlobeDepthMaterial';
 import MatteIdsMaterial from 'Renderer/MatteIdsMaterial';
@@ -39,10 +35,6 @@ function TileMesh(params, builder, geometryCache) {
 
     this.geometry = geometryCache || new TileGeometry(params, builder);
     this.normal = params.center.clone().normalize();
-
-    var worldNormal = new THREE.Vector3(0, 0, 1).applyQuaternion(this.OBB().getWorldQuaternion());
-    // distance to globe center
-    this.distance = params.center.clone().projectOnVector(worldNormal).length();
 
     // TODO Why move sphere center
     this.centerSphere = new THREE.Vector3().addVectors(this.geometry.boundingSphere.center, params.center);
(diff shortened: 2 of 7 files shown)


src/Globe/TileMesh.js
```diff
@@ -72,28 +63,6 @@ TileMesh.prototype = Object.create(NodeMesh.prototype);
 
 TileMesh.prototype.constructor = TileMesh;
 
-TileMesh.prototype.buildHelper = function buildHelper() {
```
gchoqueux, line 75: It's good that this is to be removed from here, where do you move it?

src/Scene/NodeProcess.js
```diff
@@ -473,16 +473,16 @@ NodeProcess.prototype.processNode = function processNode(node, camera, params) {
...
 NodeProcess.prototype.frustumCullingOBB = function frustumCullingOBB(node, camera) {
-    // position in local space
-    var position = node.OBB().worldToLocal(camera.position().clone());
-    position.z -= node.distance;
-
-    quaternion.multiplyQuaternions(node.OBB().inverseQuaternion(), camera.camera3D.quaternion);
-
-    return camera.getFrustumLocalSpace(position, quaternion).intersectsBox(node.OBB().box3D);
+    obbMatrixWorld.multiplyMatrices(node.matrixWorld, node.OBB().matrix);
+    const m = camera.camera3D.matrixWorldInverse.clone();
```
gchoqueux: you can remove a clone `matrixWorldInverse` with `premultiply()`
```js
obbMatrixWorld.premultiply(camera.camera3D.matrixWorldInverse);
```

gchoqueux (changes requested):
Navigation and loading are more than the master

src/Globe/TileMesh.js
Jeremy-Gaillard: I'm anticipating #239 where the helpers are built in a debug layer. See [here](https://github.com/iTowns/itowns2/blob/c179f064324f18f73c4b1e23ded7250a9978d4f8/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js#L370).

src/Scene/NodeProcess.js
Jeremy-Gaillard: Good catch, changing it.

src/Globe/TileMesh.js
gchoqueux: ok

src/Scene/NodeProcess.js
```diff
@@ -479,16 +479,15 @@ NodeProcess.prototype.processNode = function processNode(node, camera, params) {
...
 
 NodeProcess.prototype.frustumCullingOBB = function frustumCullingOBB(node, camera) {
-    // position in local space
-    var position = node.OBB().worldToLocal(camera.position().clone());
-    position.z -= node.distance;
-
-    quaternion.multiplyQuaternions(node.OBB().inverseQuaternion(), camera.camera3D.quaternion);
-
-    return camera.getFrustumLocalSpace(position, quaternion).intersectsBox(node.OBB().box3D);
+    obbMatrixWorld.multiplyMatrices(node.matrixWorld, node.OBB().matrix);
```
gchoqueux: error result `frustumCullingOBB` :
new frustrum
capture du 2017-02-20 18-11-01

old frustumCullingOBB
capture du 2017-02-20 18-13-58

Jeremy-Gaillard: Fixed

src/Scene/NodeProcess.js
```diff
@@ -479,16 +479,19 @@ NodeProcess.prototype.processNode = function processNode(node, camera, params) {
...
-var quaternion = new THREE.Quaternion();
+const frustum = new THREE.Frustum();
+const obbMatrixWorld = new THREE.Matrix4();
 
 NodeProcess.prototype.frustumCullingOBB = function frustumCullingOBB(node, camera) {
-    // position in local space
-    var position = node.OBB().worldToLocal(camera.position().clone());
-    position.z -= node.distance;
+    // Move camera in OBB local space
+    obbMatrixWorld.copy(camera.camera3D.matrixWorldInverse);
```
gchoqueux: Avoiding a copy
```js
obbMatrixWorld.multiplyMatrices(camera.camera3D.matrixWorldInverse, node.matrixWorld);
obbMatrixWorld.multiply(node.OBB().matrix);
obbMatrixWorld.premultiply(camera.camera3D.projectionMatrix);
frustum.setFromMatrix(obbMatrixWorld);
```

peppsac:
@Jeremy-Gaillard you could have almost split this PR in 2:
  - rework frustum culling
  - TileMesh refactoring / cleanup

And in the rework frustum culling, it could have been interesting to implement a debug bar chart, displaying how many tiles are culled and visible per level.
(Would have been easier to verify that the culling behavior didn't change with your patch)

If you're interested in doing this I had a graph like this in my old debug branch (see https://github.com/peppsac/itowns2/commit/d8dffd0a494458364d902da6f82d88e02531169f).

src/Scene/NodeProcess.js
gchoqueux: another optimisation and also more logical :

```js
const frustum = new THREE.Frustum();
const obbMatrixWorld = new THREE.Matrix4();
const modelViewMatrix = new THREE.Matrix4();
NodeProcess.prototype.frustumCullingOBB = function frustumCullingOBB(node, camera) {
    // Move camera in OBB local space
    // must be precompute in node.OBB().matrixWorld in tileProvider and when add texture elevation
    obbMatrixWorld.multiplyMatrices(node.matrixWorld, node.OBB().matrix);
    ///  must precompute in modelViewMatrix in camera
    modelViewMatrix.multiplyMatrices(camera.camera3D.projectionMatrix, camera.camera3D.matrixWorldInverse);
    obbMatrixWorld.premultiply(modelViewMatrix);
    frustum.setFromMatrix(obbMatrixWorld);
    return frustum.intersectsBox(node.OBB().box3D);
};
```

Which give 

```js
const frustum = new THREE.Frustum();
const obbMatrixWorld = new THREE.Matrix4();
NodeProcess.prototype.frustumCullingOBB = function frustumCullingOBB(node, camera) {
    // Move camera in OBB local space
    obbMatrixWorld.multiply(camera.camera3D.modelViewMatrix, node.OBB().matrixWorld);
    frustum.setFromMatrix(obbMatrixWorld);
    return frustum.intersectsBox(node.OBB().box3D);
};
```

Jeremy-Gaillard: Ideally, the OBB should be a child of the node, so the world matrix is updated each time the node matrix changes. This will be possible after #239 is merged.

In the meantime, I will just add a comment to remind us of doing it later.

I'll add the camera matrix precomputing in the prepare function.

gchoqueux: Question : where in #239, the world matrix is updated?

Jeremy-Gaillard: Three does it automatically (see matrixAutoUpdate attribute). However, in the current version, we can't add other objects to the node's children without breaking the quadtree processes.

Jeremy-Gaillard:
@peppsac I'll look into it.

src/Scene/NodeProcess.js
gchoqueux: Now there's :
`this.scene3D.autoUpdate = false;`

I don't know which is priority between `scene.autoUpdate` and `object.matrixAutoUpdate`

I have to do update matrix world since this change, that's why I want to know where is the update matrix world (for not RTC) in the 239

Jeremy-Gaillard: [Three's scene documentation](https://threejs.org/docs/index.html?q=scene#Reference/Scenes/Scene) seems to indicate that autoUpdate has priority.

It would be a good idea to remove all the `matrixAutoUpdate = false` since they are useless.

Jeremy-Gaillard:
I added a debug bar chart that shows the number of visible tiles per level.

I also optimised the culling process, as suggested by @gchoqueux.

src/Core/Commander/Providers/TileProvider.js
```diff
@@ -60,6 +63,8 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
     parent.add(tile);
     tile.updateMatrix();
     tile.updateMatrixWorld();
+    tile.OBB().parent = tile;
```
gchoqueux: tile is parent, but OBB isn't a child.
Comment to change in `tile.add`

src/Scene/NodeProcess.js
```diff
@@ -21,6 +21,7 @@ function NodeProcess(scene, camera, ellipsoid) {
 
     this.r = ellipsoid.size || new THREE.Vector3();
     this.cV = new THREE.Vector3();
+    this.cameraViewMatrix = new THREE.Matrix4();
```
gchoqueux: yo can move this.cameraViewMatrix in camera 
`camera.viewMatrix`
and add function `camera.updateViewMatrix()`

utils/debug/Debug.js
```diff
@@ -18,21 +18,38 @@ function Debug(scene) {
 
     scene.viewerDiv.appendChild(chartDiv);
 
+    const leftChart = document.createElement('div');
```
gchoqueux, line 21: nice :+1:

gchoqueux (changes requested):
* the horizon culling is broken
* Are you sure that OBB.matrixWorld is update with the new texture elevation

src/Core/Commander/Providers/TileProvider.js
Jeremy-Gaillard: Unfortunately, we can't do that without breaking iTowns (browsetree goes through EVERY child of the node, so it crashes when it tries to process the OBB). Hopefully, #239 will allow us to do it. I left a comment in the meantime.

src/Scene/NodeProcess.js
Jeremy-Gaillard: Done, but I reused the camera.update() function instead.

Jeremy-Gaillard:
@gchoqueux 
* "the horizon culling is broken"
Fixed

* "Are you sure that OBB.matrixWorld is update with the new texture elevation"
It seems so : TileMesh.setTextureElevation -> TileMesh.setBBoxZ -> OBB.addHeight -> OBB.update

utils/debug/Debug.js
peppsac: Looks nice indeed, and will definitely be useful.

One more suggestion (if it's easy to implement): I'd love to have both `visible` and `displayed` properties on the same graph :)

Jeremy-Gaillard: Done!

gchoqueux:
All seems good, nice

gchoqueux:
need a rebase and validation of @peppsac

gchoqueux (approved):

PR #266, opened 2017-02-24
refacto (GPX) : Improvements and add example

gchoqueux:
- Use THREE.MeshLine to fix linewidth
Due to limitations in the ANGLE layer, with the WebGL renderer on Windows platforms linewidth will always be 1 regardless of the set value. 

- add pins for way points 
- fix no elevation Gpx 
- Add gpx example

capture


gchoqueux:
rebased

src/Renderer/Lines.js
```diff
@@ -144,11 +144,13 @@ ItownsLine.prototype.process = function process() {
...
 };
 
-ItownsLine.prototype.createQuad = function createQuad(pt1, pt2) {
+Lines.prototype.createQuad = function createQuad(pt1, pt2) {
     // Définition propre a chaque géométrie
     var geometry = new THREE.BufferGeometry();
 
     // les 6 points
+
+    // var indices  =  new Int16Array([0,1,2,3,4,5]);
```
peppsac: Remove dead code

src/Renderer/Shader/LineVS.glsl
```diff
@@ -29,16 +6,18 @@ attribute float side;
...
-uniform mat4   modelViewMatrix;
+uniform mat4 mVPMatRTC;
+uniform mat4 projectionMatrix;
+uniform mat4 modelViewMatrix;
 
-uniform float	THICKNESS;	// the thickness of the line in pixels
-uniform float	MITER_LIMIT;	// 1.0: always miter, -1.0: never miter, 0.75: default
-uniform vec2	WIN_SCALE;	// the size of the viewport in pixels
+uniform float THICKNESS;  // the thickness of the line in pixels
+uniform bool MITER_LIMIT;    // 1.0: always miter, -1.0: never miter, 0.75: default
```
peppsac: `bool` type conflicts with comment.

src/Renderer/LineMaterial.js
```diff
@@ -0,0 +1,44 @@
...
+    this.fragmentShader = this.fragmentShaderHeader + LineFS;
+    this.vertexShader = this.vertexShaderHeader + LineVS;
+
+    this.wireframe = false;
+
+    const texture = options.texture ? Fetcher.texture.load(options.texture) : undefined;
+
+    this.uniforms.THICKNESS = { value: options.linewidth };
+    this.uniforms.MITER_LIMIT = { value: true };
+    this.uniforms.WIN_SCALE = { value: new THREE.Vector2(window.innerWidth, window.innerHeight) };
```
peppsac: This code is wrong (window size might be different of itowns container size).
And the container size can change so this uniforms probably needs to be updated before rendering.
(this is'nt something you'll want to do in this PR though)

peppsac: Actually simply doing something similar to:
```
this.uniforms.WIN_SCALE = { value: c3DEngine().viewerDiv.size() }
```
should be enough (ie: pointing the uniforms to an Vector2 object stored elsewhere and kept up to date)

peppsac (changes requested):
Only a few minor comments to address

src/Renderer/LineMaterial.js
```diff
@@ -0,0 +1,44 @@
...
+    if (options === undefined) {
+        throw new Error('options is required');
+    }
+
+    this.fragmentShader = this.fragmentShaderHeader + LineFS;
+    this.vertexShader = this.vertexShaderHeader + LineVS;
+
+    this.wireframe = false;
+
+    const texture = options.texture ? Fetcher.texture.load(options.texture) : undefined;
```
peppsac: `Fetcher.texture(options.texture)` ?

peppsac (changes requested):
This PR breaks GPX rendering.
On master `itowns.viewer.loadGPX(...)` works, but the gpx isn't visible in this PR.

gchoqueux:
rebased

src/Renderer/LineMaterial.js
peppsac: You missed this comment.
You probably want to replace this line by:
```js
const texture = options.texture ? Fetcher.texture(options.texture).texture : undefined;
```

gchoqueux: ok, I fix

src/Core/Commander/Providers/GpxUtils.js
```diff
@@ -99,6 +101,13 @@ function _gpxToMesh(gpxXML) {
         gpxMesh.add(wayPts);
     }
 
+    gpxMesh.matrixWorld.elements = new Float64Array(16);
+    gpxMesh.matrix.elements = new Float64Array(16);
```
peppsac: Add a comment explaining why you're overriding `Float32Array` with `64` ones.

src/Renderer/PointMaterial.js
```diff
@@ -13,18 +14,19 @@ const ItownsPointMaterial = function ItownsPointMaterial(options) {
     this.vertexShader = PointVS;
     this.fragmentShader = PointFS;
 
-    var texture = new THREE.TextureLoader().load(options.texture);
+    const texture = options.texture ? Fetcher.texture(options.texture) : undefined;
```
peppsac: Still not correct.
From my original comment :
```js
const texture = options.texture ? Fetcher.texture(options.texture).texture : undefined;
//                                                                ^^^^^^^^
```

src/Renderer/c3DEngine.js
```diff
@@ -32,6 +32,7 @@ function c3DEngine(scene, positionCamera, viewerDiv, debugMode, gLDebug) {
     this.scene3D.sortObjects = false;
     this.width = this.debug ? viewerDiv.clientWidth * 0.5 : viewerDiv.clientWidth;
     this.height = viewerDiv.clientHeight;
+    this.viewerDiv.size = new THREE.Vector2(this.width, this.height);
```
peppsac: Don't modify `viewerDiv` please (this isn't what my original comment suggested).

peppsac: Instead replace `c3DEngine`'s widhth/height property with a size and use it in like this:
```js
this.uniforms.WIN_SCALE = { value: c3DEngine().size };
```

peppsac (changes requested):
Thanks for the PR.
Do you have any idea why a .gpx line is only visible if you zoom close to the ground ? (level 13 tiles)

examples/gpx/ULTRA2009.gpx
```diff
@@ -0,0 +1,10683 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
autra: Why not using https://github.com/iTowns/iTowns2-sample-data ? This repo was meant for that right?

autra (dismissed):
Is this PR still relevant? (nearly all the affected files are gone) If not I guess we can close it

autra:
For reference, https://github.com/iTowns/itowns2/pull/345 (this PR obsoletes this the most).

gchoqueux:
Line's thickness doesn't work on windows

peppsac:
> Line's thickness doesn't work on windows

What about trying to get this fixed upstream (`three.js`) ?

gchoqueux:
Updated 

Reviewer?

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -84,6 +148,14 @@ function _gpxToMesh(gpxXML) {
         gpxMesh.add(wayPts);
     }
 
+    gpxMesh.matrixWorld.elements = new Float64Array(16);
+    gpxMesh.matrix.elements = new Float64Array(16);
```
peppsac: You don't need to replace `.elements` anymore.

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -84,6 +148,14 @@ function _gpxToMesh(gpxXML) {
...
+    gpxMesh.position.copy(gpxXML.center);
+    gpxMesh.updateMatrix();
```
peppsac: Not needed (called by `updateMatrixWorld` if `matrixAutoUpdate` is `true` which is the default value)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -84,6 +148,14 @@ function _gpxToMesh(gpxXML) {
...
+    gpxMesh.updateMatrixWorld(true);
+    gpxMesh.matrixAutoUpdate = false;
+    gpxMesh.matrixWorldNeedsUpdate = false;
```
peppsac: Not needed either (set by `updateMatrixWorld`)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -24,40 +24,104 @@ function _gpxPtToCartesian(pt) {
...
     var wayPts = _gpxToWayPointsArray(gpxXML);
 
     if (wayPts.length) {
-        const positions = new Float32Array(wayPts.length * 3);
+        const points = new THREE.Group();
+
         for (var i = 0; i < wayPts.length; i++) {
-            const pos = _gpxPtToCartesian(wayPts[i]);
-            pos.toArray(positions, 3 * i);
+            if (gpxXML.center === undefined) {
```
peppsac: You could move this out of the `for` loop to improve readability.

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -24,40 +24,104 @@ function _gpxPtToCartesian(pt) {
...
+                gpxXML.center = _gpxPtToCartesian(wayPts[0]);
+            }
+            const position = _gpxPtToCartesian(wayPts[i]).sub(gpxXML.center);
+            const mesh = new THREE.Mesh(geometryPoint, materialPoint);
+            mesh.position.copy(position);
+            mesh.lookAt(gpxXML.center.clone().negate());
+            mesh.updateMatrixWorld();
```
peppsac: You can save this one as well because at some point you're going to call `updateMatrixWorld`on the `points` which will then call `gpxToWayPointsMesh` recursively.

examples/gpx.js
```diff
@@ -0,0 +1,28 @@
+/* global itowns, document, renderer */
+// # Simple Globe viewer
```
peppsac: Fix comment please

examples/gpx.js
```diff
@@ -0,0 +1,28 @@
...
+// This layer is defined in a json file but it could be defined as a plain js
+// object. See Layer* for more info.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+// Add two elevation layers.
+// These will deform iTowns globe geometry to represent terrain elevation.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+
+exports.view = globeView;
+exports.initialPosition = positionOnGlobe;
```
peppsac: Unused?

examples/gpx.html
```diff
@@ -0,0 +1,71 @@
...
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script type="text/javascript">
+            var renderer;
+            var exports = {};
+        </script>
+        <script src="gpx.js"></script>
+        <script type="text/javascript">
+            /* global itowns, document, GuiTools, globeView, promises */
+            var menuGlobe = new GuiTools('menuDiv');
```
peppsac: I don't think the `GuiTools` are useful in this example so I'd remove them.

This will help shortening your example to the bare minimum needed to display your gpx file.

peppsac (changes requested):
Thanks.

I left a first round of comments, mainly small style issues.

You should include a comment in the source code explaining why you're not using Points/Line in the simplest way.

peppsac:
Btw, I know it's out the of scope of your PR but the GPX support should support a different destination CRS than `EPSG:4978`.

Also could you add a screenshot of what GPX rendering looks like with your changes?

gchoqueux:
i'll add CRS support

examples/gpx.html
```diff
@@ -0,0 +1,65 @@
...
+        <meta charset="UTF-8">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="GUI/dat.gui/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <script src="GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
```
peppsac: You can remove this line.

examples/gpx.html
```diff
@@ -0,0 +1,65 @@
...
+            @media (max-width: 600px) {
+                #menuDiv {
+                    display: none;
+                }
+            }
+        </style>
+        <meta charset="UTF-8">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="GUI/dat.gui/dat.gui.min.js"></script>
```
peppsac: You can remove this line

examples/gpx.html
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            #viewerDiv {
+                margin : auto auto;
+                width: 100%;
+                height: 100%;
+                padding: 0;
+            }
+
+            #menuDiv {
```
peppsac: You can remove all the `menuDiv` stuff here

examples/gpx.html
```diff
@@ -0,0 +1,65 @@
...
+        <script type="text/javascript">
+            var renderer;
+            var exports = {};
+        </script>
+        <script src="gpx.js"></script>
+        <script type="text/javascript">
+            /* global itowns, document, GuiTools, globeView, promises */
+            // Listen for globe full initialisation event
+            globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
+                // eslint-disable-next-line no-console
```
peppsac: `html` files are not linter so this is not needed.

examples/gpx.html
```diff
@@ -0,0 +1,65 @@
...
+                console.info('Globe initialized');
+                itowns.loadGpx('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/ULTRA2009.gpx', globeView.referenceCrs).then((gpx) => {
```
peppsac: You can't use arrow functions here (must only use ES5 features)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +16,154 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
     // Getting the waypoint points
-    var wayPts = _gpxToWayPointsMesh(gpxXML);
+    var wayPts = _gpxToWayPointsMesh(gpxXML, crs);
 
     if (wayPts) {
         gpxMesh.add(wayPts);
     }
 
+    gpxMesh.position.copy(gpxXML.center);
+    gpxMesh.updateMatrix();
```
peppsac: Call to `updateMatrix` isn't necessary.

gchoqueux:
updated

examples/gpx.html
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            #viewerDiv {
+                margin : auto auto;
+                width: 100%;
+                height: 100%;
+                padding: 0;
+            }
+
+            @media (max-width: 600px) {
```
peppsac: You forgot to remove this as well

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +16,153 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+    gpxMesh.updateMatrixWorld();
+    // gpxMesh is static data, it doens't need matrix update
+    gpxMesh.matrixAutoUpdate = false;
+
     return gpxMesh;
 }
 
-export default function loadGpx(urlFile, networkOptions) {
-    return Fetcher.xml(urlFile, networkOptions).then(gpxXML => _gpxToMesh(gpxXML));
+export default function loadGpx(urlFile, crs, networkOptions) {
```
peppsac: Don't you think you should include `networkOptions` in a `options` object that would also allow to configure a few things:
- use 'pin' or not
- line size (because with your change the user cannot simply set : `line.material.linewidth`)

peppsac: Last but not least, but while reworking this file, maybe you could export it as:
```js
export default {
   load(urlFile, crs, options) {
   ... 
   }
};
```
This way the user would do:
```js
itowns.GpxUtils.load(...);
```

peppsac:
Another suggestion for the linewidth issue would be to use: https://github.com/spite/THREE.MeshLine/

gchoqueux:
@peppsac updated
Thanks, I replace by MeshLine with depth log path for shader

gchoqueux:
updated
add options and doc

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
         return points;
     } else {
         return null;
     }
 }
 
-function _gpxToWTrackPointsMesh(gpxXML) {
+function updatePath(renderer, scene, camera) {
+    const distance = getDistance(this, camera);
+    this.material.depthTest = distance < this.geometry.boundingSphere.radius * 2;
```
peppsac, line 79: Why do you need to disable depth test?
Is there a valid reason or is this a workaround for a bug in iTowns?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "lib/Main.js",
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
-    "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js",
+    "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js src/Core/Geographic/Coordinates.js src/Core/Scheduler/Providers/GpxUtils.js",
```
peppsac: `src/Core/Geographic/Coordinates.js` is now listed twice.

package.json
```diff
@@ -40,7 +40,8 @@
     "proj4": "^2.4.3",
     "text-encoding-utf-8": "^1.0.1",
     "three": "^0.86.0",
-    "whatwg-fetch": "^2.0.2"
+    "whatwg-fetch": "^2.0.2",
+    "three.meshline": "^1.0.3"
```
peppsac: Did you add this one manually (because it's not properly sorted)?

peppsac: Also, you didn't update `package-lock.json` (that requires a recent enough version of npm)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -5,8 +5,12 @@
  */
 
 import * as THREE from 'three';
+import Line from 'three.meshline';
 import Fetcher from './Fetcher';
 import Coordinates from '../../Geographic/Coordinates';
+import Capabilities from '../../System/Capabilities';
+// eslint-disable-next-line
```
peppsac: Why do you need to disable eslint here?

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -5,8 +5,12 @@
...
+import $3dTiles_Provider, { patchMaterial } from './3dTiles_Provider';
```
peppsac: You should only import `patchMaterial`:
```js
import { patchMaterial } from './3dTiles_Provider'; 
```

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -99,7 +99,7 @@ function getBox(volume, inverseTileTransform) {
 
 const rePosition = new RegExp('gl_Position.*(?![^]*gl_Position)');
 const reMain = new RegExp('[^\\w]*main[^\\w]*(void)?[^\\w]*{');
-function patchMaterial(material) {
+export function patchMaterial(material) {
```
peppsac: Maybe you could have moved this function to its own utility file.
At the very least change its name to really reflect what it's doing (`patchMaterialForLogDepthSupport`?)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+    return new Coordinates('EPSG:4326', longitude, latitude, elevation).as(crs).xyz();
 }
 
-function _gpxToWayPointsMesh(gpxXML) {
+const geometryPoint = new THREE.BoxGeometry(1, 1, 80);
+const materialPoint = new THREE.MeshBasicMaterial({ color: 0xffffff });
+const point = new THREE.Vector3();
+
+function getDistance(object, camera) {
+    point.setFromMatrixPosition(object.matrixWorld).add(object.geometry.boundingSphere.center);
```
peppsac: You don't account for scaling and rotation here.
```js
object.geometry.boundingSphere.center.clone().applyMatrix4(object.matrixWorld);
```
would be correct I think.

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+    return camera.position.distanceTo(point);
```
peppsac: This ignores camera's transformation as well (but this isn't the only place in itowns that assumes that the camera has no parent)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
         return null;
     }
 }
 
-function _gpxToWTrackPointsMesh(gpxXML) {
+function updatePath(renderer, scene, camera) {
+    const distance = getDistance(this, camera);
+    this.material.depthTest = distance < this.geometry.boundingSphere.radius * 2;
+    const size = renderer.getSize();
+    this.material.uniforms.resolution.value.set(size.width, size.height);
```
peppsac, line 81: We should probably provide an easy way to register things that needs to be updated on resolution change (material like this one, render targets, etc).
(not for this PR though)

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+    this.material.needsUpdate = true;
```
peppsac: I don't think you need this. Did you test without it?

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+
+function _gpxToWTrackPointsMesh(gpxXML, crs, lineWidth = 12) {
     var trackPts = _gGpxToWTrackPointsArray(gpxXML);
 
     if (trackPts.length) {
-        const positions = new Float32Array(trackPts.length * 3);
-        for (var i = 0; i < trackPts.length; i++) {
-            const pos = _gpxPtToCartesian(trackPts[i]);
-            pos.toArray(positions, 3 * i);
+        if (gpxXML.center === undefined) {
```
peppsac: `gpxXML.center = gpxXML.center || _gpxPtToCartesian(trackPts[0], crs);` ?

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
     var wayPts = _gpxToWayPointsArray(gpxXML);
 
     if (wayPts.length) {
-        const positions = new Float32Array(wayPts.length * 3);
-        for (var i = 0; i < wayPts.length; i++) {
-            const pos = _gpxPtToCartesian(wayPts[i]);
-            pos.toArray(positions, 3 * i);
+        const points = new THREE.Group();
+
+        if (gpxXML.center === undefined) {
```
peppsac: `gpxXML.center = gpxXML.center || _gpxPtToCartesian(trackPts[0], crs);` ?

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: There's visibility problem with DTM for long distances

sans

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+        pathMesh.onBeforeRender = updatePath;
+
+        return pathMesh;
     } else {
         return null;
     }
 }
 
-function _gpxToMesh(gpxXML) {
+function _gpxToMesh(gpxXML, crs, options) {
```
peppsac: `, options = {}) {`

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +20,173 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+     * @param {string} urlFile  The url of gpx file
+     * @param {string} crs - The default CRS of Three.js coordinates. Should be a cartesian CRS.
+     * @param {Object=} options Optional properties.
+     * @param {boolean=} [options.enablePin=true] draw pin for way points
+     * @param {NetworkOptions=} options.networkOptions Options for fetching resources over network
+     * @param {number=} [options.lineWidth=12] set line width to track line
+     * @return {THREE.Mesh} Three.js Mesh see {@link https://threejs.org/docs/#api/objects/Mesh}
+     * @example
+     * // How add gpx object
+     * itowns.GpxUtils.load(url).then((gpx) => {
```
peppsac: Missing `crs` arg.

src/Core/View.js
```diff
@@ -28,7 +28,7 @@ import Scheduler from './Scheduler/Scheduler';
  * @constructor
  * @example
  * // How add gpx object
- * itowns.loadGpx(url).then((gpx) => {
+ * itowns.GpxUtils.load(url).then((gpx) => {
```
peppsac: Missing `crs`

peppsac (changes requested):

src/Core/Scheduler/Providers/GpxUtils.js
peppsac: Yep.
But my question was: did you look for the root cause (for instance playing with the near/far values of our camera) before settling on this fix?

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: I have to write $3dTiles_Provider to import patchMaterial otherwise there's a error
and i not use 3dTiles_Provider so i disable eslint

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: I have to write $3dTiles_Provider to import patchMaterial otherwise there's a error

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: yes

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: yes

src/Core/Scheduler/Providers/GpxUtils.js
peppsac: Or is this because the elevation in the gpx and in the DEM doesn't match?

src/Core/Scheduler/Providers/3dTiles_Provider.js
gchoqueux: in a new PR

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: Forget about displacement map

src/Core/Scheduler/Providers/3dTiles_Provider.js
peppsac: Nope: this PR exports a previously private function so you must either split it to a new file or nename it here.

gchoqueux: ok to rename

src/Core/Scheduler/Providers/GpxUtils.js
gchoqueux: yes, the reason the gpx and in the DEM doesn't match
if gpx matchs with DEM, there's depth problem

gchoqueux:
updated

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +19,172 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
     }
 }
 
-function _gpxToMesh(gpxXML) {
+function _gpxToMesh(gpxXML, options) {
     if (!gpxXML) {
         return undefined;
     }
 
+    options = options || {};
```
peppsac: You can remove this, if you make sure that `options` is always defined.
And the simplest way to do this is to fix the only entry point for this module:
```js
load(urlFile, crs, options = {}) {
...
```

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +19,172 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+     * itowns.GpxUtils.load(url, viewer.referenceCrs).then((gpx) => {
+     *      if (gpx) {
+     *         viewer.scene.add(gpx);
+     *         viewer.notifyChange(true);
+     *      }
+     * });
+     *
+     */
+    load(urlFile, crs, options) {
+        options = options || {};
```
peppsac: As stated earlier: move this to function signature.

src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -16,78 +19,172 @@ function _gGpxToWTrackPointsArray(gpxXML) {
...
+        options.crs = crs;
+        return Fetcher.xml(urlFile, options ? options.networkOptions : {}).then(gpxXML => _gpxToMesh(gpxXML, options));
```
peppsac: Then you can directly use: `.xml(urlFile, options.networkOptions)`

src/Core/Scheduler/Providers/GpxUtils.js
peppsac: Still looks like a hack to me but we can live with this for a while (#449 could help here).

peppsac (approved):
I've left a few nits to address before merging but otherwise it looks good.
Thanks!

PR #267, opened 2017-02-24
Replace CVML by Earcut

gchoqueux:
new module for triangualisation


src/Core/Commander/Providers/BuildingBox_Provider.js
```diff
@@ -70,8 +70,7 @@ BuildingBox_Provider.prototype.getData = function getData(bbox, altitude) {
 };
 
 BuildingBox_Provider.prototype.generateMesh = function generateMesh(elements, bbox, altitude) {
-    // console.log(elements);
-
+    var roofGeometry = new THREE.Geometry(); // for the roof
```
peppsac, line 73: What's the difference with `var geometry = new THREE.Geometry(); // for the roof ` 2 lines below?

src/Core/Commander/Providers/BuildingBox_Provider.js
```diff
@@ -99,7 +98,8 @@ BuildingBox_Provider.prototype.generateMesh = function generateMesh(elements, bb
                 const vector3_1 = new THREE.Vector3(pgeo1.x, pgeo1.y, pgeo1.z); // - x temporary, bug
                 const vector3_2 = new THREE.Vector3(pgeo2.x, pgeo2.y, pgeo2.z);
 
-                arrPoint2D.push(CVML.newPoint(p1.z, p1.x)); // -pgeo1.x, pgeo1.z)); //for roof
+
+                arrPoint2D.push(p1.z, p1.x);
```
peppsac, line 102: Why not simply make `p1` a Vector2 and doing `arrPoint2D.push(p1)`?
It would simplify the loop code below to:
```js
    const pt1 = arrPoint2D[triangles[w]];
    const pt2 = arrPoint2D[triangles[w + 1]];
    const pt3 = arrPoint2D[triangles[w + 2]];
```
Or even:
```js
    const [pt1, pt2, pt3] = arrPoint2D.slice(w, w+3);
```

gchoqueux:
BuildingBox_Provider, it's unfortunately obsolete (like MobileMapping), it was added as prototyping. 
It would be necessary to remove mobileMapping folder.

peppsac:
Actually I have restored mobilemapping feature in a branch.
As soon as #239 is merged I can submit a PR for this (and clean unneeded code from MobileMapping folder).

gchoqueux:
The BuildingBox_Provider's code parse of the geojson and there is duplicate with the PR Feature #195 that will centralize this code.

gchoqueux:
That's why I'm not optimizing this part

peppsac:
Ok.

peppsac (approved):
Looks good then.

PR #268, opened 2017-02-28
BABEL_DISABLE_CACHE for windows

vcoindet:
Put "cross-env" before "BABEL_DISABLE_CACHE" in package.json so #248 it will work on windows too ! ;)

To be tested on other environment than windows before merge


tbroyer (approved):

tbroyer:
Verified locally on Ubuntu, and on Travis: https://travis-ci.org/iTowns/itowns2/jobs/206135441#L156

Issue #260, opened 2017-02-17
pan(pVector)


vcoindet:
No repeatability of the pan : 
If one sets an pan of {x : 1000, y : 1000} and then a pan of {x : -1000, y : -1000}, one doesn't come back on the original position

gchoqueux:
fix in #273

elias75015:
Looks fine on #273 :+1:

Issue #255, opened 2017-02-16
setCenterAdvanced problems


vcoindet:
- SetCenterAdvanced doesn't have the same behavior with or without animation 
        -> Without animation : the range taken into account
        -> Without animation : bug

- The zoom level and scale level don't seem to be taken into account at all...

gchoqueux:
fix in #273

PR #273, opened 2017-03-17
Fixes bad camera positions and change options WMTS

gchoqueux:
Fixes camera bad positions when you use :

* setCameraTargetGeoPosition() #255
* setRange()
* pan() see #260
* createSceneGlobe()

Move WMTS's options in layer.options 

Add little doc addImageryLayer and addElevationLayer

**Warning, this PR doesn't fix incorrect position with setZoomLevel and setZoomScale**

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -16,7 +16,8 @@ import Fetcher from '../../Providers/Fetcher';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from '../../../../Scene/LayerUpdateStrategy';
 
 var sceneIsLoaded = false;
-var eventLoaded = new CustomEvent('globe-loaded');
+export const INITIALIZED_EVENT = 'initialized';
+
 var eventRange = new CustomEvent('rangeChanged');
 var eventOrientation = new CustomEvent('orientationchanged');
 var eventPan = new CustomEvent('panchanged');
@@ -26,7 +27,6 @@ var eventLayerChanged = new CustomEvent('layerchanged');
 var eventLayerChangedVisible = new CustomEvent('layerchanged:visible');
 var eventLayerChangedOpacity = new CustomEvent('layerchanged:opacity');
 var eventLayerChangedIndex = new CustomEvent('layerchanged:index');
-var eventError = new CustomEvent('error');
 
 var enableAnimation = false;
 
@@ -320,15 +320,13 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
     // Deplacer plus tard
 
     this.viewerDiv = viewerDiv;
+    this.sceneLoadedDeferred = defer();
 
     viewerDiv.addEventListener('globe-built', () => {
-        if (sceneIsLoaded === false) {
+        if (!sceneIsLoaded) {
             sceneIsLoaded = true;
-            this.scene.currentControls().updateCameraTransformation();
-            this.scene.updateScene3D();
-            viewerDiv.dispatchEvent(eventLoaded);
-        } else {
-            viewerDiv.dispatchEvent(eventError);
+            this.sceneLoadedDeferred.resolve();
+            this.sceneLoadedDeferred = defer();
         }
     }, false);
 
@@ -355,10 +353,10 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
     this.scene.scheduler.addProtocolProvider('tile', new TileProvider());
     this.scene.scheduler.addProtocolProvider('wms', new WMS_Provider({ support: map.gLDebug }));
 
-    this.sceneLoadedDeferred = defer();
-    this.addEventListener('globe-loaded', () => {
-        this.sceneLoadedDeferred.resolve();
-        this.sceneLoadedDeferred = defer();
+    this.setSceneLoaded().then(() => {
+        this.scene.currentControls().updateCameraTransformation();
+        this.scene.updateScene3D();
+        this.viewerDiv.dispatchEvent(new CustomEvent(INITIALIZED_EVENT));
     });
 
     return this.scene;
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -97,10 +97,54 @@ ApiGlobe.prototype.addGeometryLayer = function addGeometryLayer(layer) {
 };
 
 /**
- * This function adds an imagery layer to the scene. The layer id must be unique. The protocol rules wich parameters are then needed for the function.
- * @constructor
- * @param {Layer} layer.
+ * The intellectual property rights
+ * @typedef {Object} Attribution
+ * @property {string} name
+ * @property {string} url
  */
+
+/**
+ * Options to wms protocol
+ * @typedef {Object} OptionsWms
+ * @property {Attribution} attribution The intellectual property rights for the layer
+ * @property {string} name
+ * @property {string} mimetype
+ */
+
+/**
+ * Options to wtms protocol
+ * @typedef {Object} OptionsWmts
+ * @property {Attribution} attribution The intellectual property rights for the layer
+ * @property {string} name
+ * @property {string} mimetype
+ * @property {string} tileMatrixSet
+ * @property {Array.<Object>} tileMatrixSetLimits The limits for the tile matrix set
+ * @property {number} tileMatrixSetLimits.minTileRow Minimum row for tiles at the level
+ * @property {number} tileMatrixSetLimits.maxTileRow Maximum row for tiles at the level
+ * @property {number} tileMatrixSetLimits.minTileCol Minimum col for tiles at the level
+ * @property {number} tileMatrixSetLimits.maxTileCol Maximum col for tiles at the level
+ * @property {Object} [zoom]
+ * @property {Object} [zoom.min] layer's zoom minimum
+ * @property {Object} [zoom.max] layer's zoom maximum
+ */
+
+/**
+ * Layer
+ * @typedef {Object} Layer
+ * @property {string} id Unique layer's id
+ * @property {string} layer.protocol wmts and wms (wmtsc for custom deprecated)
+ * @property {string} layer.url Base URL of the repository or of the file(s) to load
+ * @property {Object} layer.updateStrategy strategy to load imagery files
+ * @property {OptionsWmts|OptionsWms} layer.options WMTS or WMS options
+ */
+
+/**
+ * This function adds an imagery layer to the scene. The layer id must be unique.
+ * The protocol rules wich parameters are then needed for the function.
+ * @constructor
+ * @param {Layer} Layer
+ */
+
 ApiGlobe.prototype.addImageryLayer = function addImageryLayer(layer) {
     preprocessLayer(layer, this.scene.scheduler.getProtocolProvider(layer.protocol));
     const map = this.scene.getMap();
@@ -195,7 +239,7 @@ ApiGlobe.prototype.removeImageryLayer = function removeImageryLayer(id) {
  * The layer id must be unique amongst all layers already inserted.
  * The protocol rules which parameters are then needed for the function.
  * @constructor
- * @param {Layer} layer.
+ * @param {Layer} layer
  */
 
 ApiGlobe.prototype.addElevationLayer = function addElevationLayer(layer) {
@@ -456,7 +500,7 @@ ApiGlobe.prototype.setCameraOrientation = function setCameraOrientation(orientat
 /**
  * Pick a position on the globe at the given position.
  * @constructor
- * @param {Number | MouseEvent} x|event - The x-position inside the Globe element or a mouse event.
+ * @param {number | MouseEvent} x|event - The x-position inside the Globe element or a mouse event.
  * @param {number | undefined} y - The y-position inside the Globe element.
  * @return {Position} position
  */
@@ -505,7 +549,7 @@ ApiGlobe.prototype.getHeading = function getHeading() {
  * Returns the "range": the distance in meters between the camera and the current central point on the screen.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/Lbt1vfek/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
  * @constructor
- * @return {Number} number
+ * @return {number} number
  */
 
 ApiGlobe.prototype.getRange = function getRange() {
@@ -642,13 +686,13 @@ ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosit
  * The zoom level and the scale can't be set at the same time.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/7yk0mpn0/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
  * @param {Position} position
- * @param {Number}  position.longitude  Coordinate longitude WGS84 in degree
- * @param {Number}  position.latitude  Coordinate latitude WGS84 in degree
- * @param {Number}  [position.tilt]  Camera tilt in degree
- * @param {Number}  [position.heading]  Camera heading in degree
- * @param {Number}  [position.range]  The camera distance to the target center
- * @param {Number}  [position.level]  level,  ignored if range is set
- * @param {Number}  [position.scale]  scale,  ignored if the zoom level or range is set. For a scale of 1/500 it is necessary to write 0,002.
+ * @param {number}  position.longitude  Coordinate longitude WGS84 in degree
+ * @param {number}  position.latitude  Coordinate latitude WGS84 in degree
+ * @param {number}  [position.tilt]  Camera tilt in degree
+ * @param {number}  [position.heading]  Camera heading in degree
+ * @param {number}  [position.range]  The camera distance to the target center
+ * @param {number}  [position.level]  level,  ignored if range is set
+ * @param {number}  [position.scale]  scale,  ignored if the zoom level or range is set. For a scale of 1/500 it is necessary to write 0,002.
  * @param {boolean}  isAnimated  Indicates if animated
  * @return {Promise}
  */
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -643,15 +643,34 @@ ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosit
  * The level has to be between the [getMinZoomLevel(), getMaxZoomLevel()].
  * The zoom level and the scale can't be set at the same time.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/7yk0mpn0/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
- * @param {Position} pPosition - The detailed position in the scene.
- * @param      {boolean}  isAnimated  Indicates if animated
- * @return     {Promise}
+ * @param {Position} position
+ * @param {Number}  position.longitude  Coordinate longitude WGS84 in degree
+ * @param {Number}  position.latitude  Coordinate latitude WGS84 in degree
+ * @param {Number}  [position.tilt]  Camera tilt in degree
+ * @param {Number}  [position.heading]  Camera heading in degree
+ * @param {Number}  [position.range]  The camera distance to the target center
+ * @param {Number}  [position.level]  level,  ignored if range is set
+ * @param {Number}  [position.scale]  scale,  ignored if the zoom level or range is set. For a scale of 1/500 it is necessary to write 0,002.
+ * @param {boolean}  isAnimated  Indicates if animated
+ * @return {Promise}
  */
-ApiGlobe.prototype.setCameraTargetGeoPositionAdvanced = function setCameraTargetGeoPositionAdvanced(pPosition, isAnimated) {
+ApiGlobe.prototype.setCameraTargetGeoPositionAdvanced = function setCameraTargetGeoPositionAdvanced(position, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
-    return this.setCameraTargetGeoPosition(pPosition, isAnimated).then(() => {
-        const p = this.scene.currentControls().setOrbitalPosition(undefined, pPosition.heading, pPosition.tilt, isAnimated);
-        return p;
+    if (position.level) {
+        position.range = this.getRangeFromZoomLevel(position.level);
+    } else if (position.scale) {
+        position.range = this.getRangeFromScale(position.scale);
+    }
+    return this.setCameraTargetGeoPosition(position, isAnimated).then(() => {
+        position.range = position.range || this.getRange();
+        position.tilt = position.tilt || this.getTilt();
+        position.heading = position.heading || this.getHeading();
+        return this.scene.currentControls().setOrbitalPosition(position.range, position.heading, position.tilt, isAnimated).then(() => {
+            this.scene.notifyChange(1);
+            return this.setSceneLoaded().then(() => {
+                this.scene.currentControls().updateCameraTransformation();
+            });
+        });
     });
 };
 
@@ -710,10 +729,14 @@ ApiGlobe.prototype.getZoomLevel = function getZoomLevel() {
  * @return     {Promise}
  */
 ApiGlobe.prototype.setZoomLevel = function setZoomLevel(zoom, isAnimated) {
-    zoom = Math.max(this.getMinZoomLevel(), zoom);
-    zoom = Math.min(this.getMaxZoomLevel(), zoom);
-    const distance = this.scene.getMap().computeDistanceForZoomLevel(zoom, this.scene.currentCamera());
-    return this.setRange(distance, isAnimated);
+    const range = this.getRangeFromZoomLevel(zoom);
+    return this.setRange(range, isAnimated);
+};
+
+ApiGlobe.prototype.getRangeFromZoomLevel = function getRangeFromZoomLevel(zoom) {
+    // FIXME : The distance computed is incorrect, (Fixed in PR 279)
+    const range = this.scene.getMap().computeDistanceForZoomLevel(zoom, this.scene.currentCamera());
+    return range;
 };
 
 /**
@@ -781,6 +804,11 @@ ApiGlobe.prototype.getZoomScale = function getZoomScale(pitch) {
  * @return     {Promise}
  */
 ApiGlobe.prototype.setZoomScale = function setZoomScale(zoomScale, pitch, isAnimated) {
+    const range = this.getRangeFromScale(zoomScale);
+    return this.setRange(range, isAnimated);
+};
+
+ApiGlobe.prototype.getRangeFromScale = function getRangeFromScale(zoomScale, pitch) {
     // Screen pitch, in millimeters
     pitch = (pitch || 0.28) / 1000;
 
@@ -823,7 +851,7 @@ ApiGlobe.prototype.setZoomScale = function setZoomScale(zoomScale, pitch, isAnim
     }
 
     const range = distance - rayon;
-    return this.setRange(range, isAnimated);
+    return range;
 };
 
 /**
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -259,14 +259,14 @@ ApiGlobe.prototype.addElevationLayersFromJSONArray = function addElevationLayers
  */
 ApiGlobe.prototype.getMinZoomLevel = function getMinZoomLevel(index) {
     var layer = this.getImageryLayers()[index];
-    if (layer && layer.zoom) {
-        return layer.zoom.min;
+    if (layer && layer.options.zoom) {
+        return layer.options.zoom.min;
     } else {
         var layers = this.getImageryLayers();
         let min = Infinity;
         for (var i = layers.length - 1; i >= 0; i--) {
-            if (layers[i].zoom) {
-                min = Math.min(min, layers[i].zoom.min);
+            if (layers[i].options.zoom) {
+                min = Math.min(min, layers[i].options.zoom.min);
             }
         }
         return min;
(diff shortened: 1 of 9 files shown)


peppsac:
Please move the test commit out of this PR (big topic so it deserves its own PR).

nosy-b:
I just started reviewing with method` ApiGlobe.getCenter()`
In index.html it seems the initial position of camera is never properly used but is rounded up. 
Try for example to start centerd on Paris` const positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };` 
Then ` ApiGlobe.getCenter()` will return {longitude: 2.3513230022207057, latitude: 49.00976271512532} which seems to be the real position as zooming get you north of Paris.

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -67,30 +67,32 @@ WMTS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer
...
+            const arrayLimits = Object.keys(options.tileMatrixSetLimits);
+            const size = arrayLimits.length;
+            const maxZoom = Number(arrayLimits[size - 1]);
+            const minZoom = maxZoom - size + 1;
 
-        layer.zoom = {
-            min: minZoom,
-            max: maxZoom,
-        };
+            layer.zoom = {
```
elias75015: Shouldn't it be in "layer.options" ?

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -634,15 +642,26 @@ ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
...
 ApiGlobe.prototype.setCenterAdvanced = function setCenterAdvanced(pPosition, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
     return this.setCenter(pPosition, isAnimated).then(() => {
-        const p = this.scene.currentControls().setOrbitalPosition(undefined, pPosition.heading, pPosition.tilt, isAnimated);
-        return p;
+        if (pPosition.range === undefined) {
+            if (pPosition.level) {
+                pPosition.range = this.getRangeFromZoomLevel(pPosition.level);
+            } else if (pPosition.scale) {
+                pPosition.range = this.getRangeFromScale(pPosition.scale);
```
elias75015: - Only with the param "isAnimated" set to false : When the scale parameter is given, the view is tilted to 90°, even without the parameter "tilt" given to the function.
- I think it is more explicit to give the scale param as a fraction denominator than in decimal (ex : 500 instead of 0.002)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -634,15 +642,26 @@ ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
...
+ * @return {Promise}
  */
 ApiGlobe.prototype.setCenterAdvanced = function setCenterAdvanced(pPosition, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
     return this.setCenter(pPosition, isAnimated).then(() => {
-        const p = this.scene.currentControls().setOrbitalPosition(undefined, pPosition.heading, pPosition.tilt, isAnimated);
-        return p;
+        if (pPosition.range === undefined) {
+            if (pPosition.level) {
+                pPosition.range = this.getRangeFromZoomLevel(pPosition.level);
```
elias75015: The function takes well into account the level param (the view seems to be correctly positioned), but the getZoomLevel doesn't return the same value as the one given in the setCenterAdvenced function.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -428,7 +428,9 @@ ApiGlobe.prototype.getCameraLocation = function getCameraLocation() {
 
 ApiGlobe.prototype.getCenter = function getCenter() {
     var controlCam = this.scene.currentControls();
-    return this.projection.cartesianToGeo(controlCam.getTargetCameraPosition());
+    const tg = controlCam.getTargetCameraPosition();
+    const coordinates = new Coordinates('EPSG:4978', tg.x, tg.y, tg.z).as('EPSG:4326');
+    return { longitude: coordinates.longitude(UNIT.DEGREE), latitude: coordinates.latitude(UNIT.DEGREE) };
```
elias75015: Not sure about the length of the returned values (does it make sense ?) : 
latitude : 47.995276640947594
longitude : 1.0015039089052769

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -660,7 +679,7 @@ ApiGlobe.prototype.setRange = function setRange(pRange, isAnimated) {
 
     return this.scene.currentControls().setRange(pRange, isAnimated).then(() => {
         this.scene.notifyChange(1);
-        this.setSceneLoaded().then(() => {
+        return this.setSceneLoaded().then(() => {
             this.scene.currentControls().updateCameraTransformation();
             this.viewerDiv.dispatchEvent(eventRange);
```
elias75015, line 735: If I do a setRange(10000), the view position is OK, but the getRange() does not return 10000.

The getRange seems to return old range value. Ex :
- setRange(10000) --> getRange() = 25011386.530042846
- setRange(200000) --> getRange() =10013.066997543165 (almost 10000)

gchoqueux:
> Then ApiGlobe.getCenter() will return {longitude: 2.3513230022207057, latitude: 49.00976271512532} which seems to be the real position as zooming get you north of Paris.

Fix in last commit, It was due to an ellipsoidal/sphere approximation

src/Core/Commander/Providers/WMTS_Provider.js
gchoqueux: ok, I move in options

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: Yes maybe, especially since I do not know the precision of my calculation

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: getZoomLevel will be fixed in #279

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: > I think it is more explicit to give the scale param as a fraction denominator than in decimal (ex : 500 instead of 0.002)

The scale of a map is the ratio,  we must be find a another name to this parameter

gchoqueux: > Only with the param "isAnimated" set to false : When the scale parameter is given, the view is tilted to 90°, even without the parameter "tilt" given to the function.

Try with the new commit ` Fix (Viewer) :  add default value for setCenterAdvanced() `

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: try with `then` promise,  you do getRange() just after setRange(), While the setRange is not yet completed 

setRange(10000).then(() => getRange())

nosy-b:
LGTM

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -260,13 +303,13 @@ ApiGlobe.prototype.addElevationLayersFromJSONArray = function addElevationLayers
 ApiGlobe.prototype.getMinZoomLevel = function getMinZoomLevel(index) {
     var layer = this.getImageryLayers()[index];
     if (layer && layer.zoom) {
```
autra: I guess it should be `layer.options.zoom`.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -260,13 +303,13 @@ ApiGlobe.prototype.addElevationLayersFromJSONArray = function addElevationLayers
...
-        return layer.zoom.min;
+        return layer.options.zoom.min;
     } else {
         var layers = this.getImageryLayers();
         let min = Infinity;
         for (var i = layers.length - 1; i >= 0; i--) {
             if (layers[i].zoom) {
```
autra: same here.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -628,7 +672,14 @@ ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosit
...
         .as('EPSG:4978').xyz();
-    position3D.range = coordinates.range;
+    if (coordinates.range) {
+        // error delta ellipsoid
+        const currentTargetPosition = this.getCameraTargetGeoPosition();
+        const oldPosition3D = new C.EPSG_4326(currentTargetPosition.longitude, currentTargetPosition.latitude, 0)
+            .as('EPSG:4978').xyz();
+        const diff = position3D.length() - oldPosition3D.length();
+        position3D.range = coordinates.range + diff;
+    }
```
autra: Suggestion: put this piece of code in GlobeControls.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -643,15 +694,37 @@ ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosit
...
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/7yk0mpn0/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
- * @param {Position} pPosition - The detailed position in the scene.
- * @param      {boolean}  isAnimated  Indicates if animated
- * @return     {Promise}
- */
-ApiGlobe.prototype.setCameraTargetGeoPositionAdvanced = function setCameraTargetGeoPositionAdvanced(pPosition, isAnimated) {
+ * @param {Position} position
+ * @param {Number}  position.longitude  Coordinate longitude WGS84 in degree
+ * @param {Number}  position.latitude  Coordinate latitude WGS84 in degree
+ * @param {Number}  [position.tilt]  Camera tilt in degree
```
autra: should not be surrounded by `[]`, right?

autra (commented):
Leaving some comments, as there are still some `layer.zoom` in the code. You might want to use `git grep` to make sure nothing is left.

Btw, why not keeping `layer.zoom` and not using `layer.options.zoom`? Do we really need the additional level (as more or less everything are layer options anyway)? 

Thanks

peppsac (dismissed):
Regardless of the changes themselves (no time for review atm) this PR has several problems:
- no PR summary (and the title is vague)
- 7 commits with broad topics
- commits messages are either empty or meaningless (e.g : `Now setRange() return Promise`)

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: position.tilt is optionnal

gchoqueux:
> layer.options.zoom

zoom is specific to WMTS, so I moved zoom in `wmts_options` (see doc)

but this is not perfect, `attribution` should out of options

src/Core/Commander/Providers/WMTS_Provider.js
```diff
@@ -241,9 +243,9 @@ WMTS_Provider.prototype.computeLevelToDownload = function computeLevelToDownload
     var lvl = ancestor ? ancestor.level : tile.level;
 
     return Math.min(
-        layer.zoom.max,
+        layer.options.zoom.max,
```
elias75015, line 244: A lot of 404 CORS errors are launched errors when we zoom beyond the zoom.max given as parameter of the layer.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -643,15 +694,37 @@ ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosit
...
+ * @param {Number}  [position.heading]  Camera heading in degree
+ * @param {Number}  [position.range]  The camera distance to the target center
+ * @param {Number}  [position.level]  level,  ignored if range is set
+ * @param {Number}  [position.scale]  scale,  ignored if the zoom level or range is set. For a scale of 1/500 it is necessary to write 0,002.
+ * @param {boolean}  isAnimated  Indicates if animated
+ * @return {Promise}
+ */
+ApiGlobe.prototype.setCameraTargetGeoPositionAdvanced = function setCameraTargetGeoPositionAdvanced(position, isAnimated) {
```
elias75015: Not totally working for the moment (tested in web dev console) :
- range parameter is strangely handled (value given doesn't fit with the view)
- heading and tilt parameters not taken into account

gchoqueux:
@autra @peppsac  : remarks take into account

gchoqueux:
@elias75015 Branch is updated

gchoqueux:
@peppsac when will you be available?

peppsac:
@gchoqueux not this week, maybe next one.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: I agree with @gchoqueux

autra: Yes yes, me too :-) Thanks for the precision

gchoqueux:
@autra @peppsac, this PR needs reviewing

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -16,7 +16,8 @@ import Fetcher from '../../Providers/Fetcher';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from '../../../../Scene/LayerUpdateStrategy';
 
 var sceneIsLoaded = false;
-var eventLoaded = new CustomEvent('globe-loaded');
+var eventLoaded = new CustomEvent('dataloaded');
+var eventInitialized = new CustomEvent('initialized');
```
autra: Create events where you dispatch them. They can't be used several time safely

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -97,10 +97,49 @@ ApiGlobe.prototype.addGeometryLayer = function addGeometryLayer(layer) {
...
+ * @property {Object} [zoom.min] layer's zoom minimum
+ * @property {Object} [zoom.max] layer's zoom maximum
+ */
+
+/**
+ * This function adds an imagery layer to the scene. The layer id must be unique.
+ * The protocol rules wich parameters are then needed for the function.
  * @constructor
- * @param {Layer} layer.
+ * @param {Layer} layer
```
autra: Missing the `Layer` typedef

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -97,10 +97,49 @@ ApiGlobe.prototype.addGeometryLayer = function addGeometryLayer(layer) {
...
- * This function adds an imagery layer to the scene. The layer id must be unique. The protocol rules wich parameters are then needed for the function.
+ * The intellectual property rights
+ * @typedef {Object} Attribution
+ * @property {String} name
+ * @property {String} url
+ */
+
+/**
+ * Options to wms protocol
+ * @typedef {Object} options_wms
```
autra: By tradition, types are in CamelCase (and variables are also camelCase in itowns anyway).

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -97,10 +97,49 @@ ApiGlobe.prototype.addGeometryLayer = function addGeometryLayer(layer) {
 };
 
 /**
- * This function adds an imagery layer to the scene. The layer id must be unique. The protocol rules wich parameters are then needed for the function.
+ * The intellectual property rights
+ * @typedef {Object} Attribution
+ * @property {String} name
```
autra: the type should be the same string returned by the `typeof` operator, so `string` and not `String`

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -97,10 +97,49 @@ ApiGlobe.prototype.addGeometryLayer = function addGeometryLayer(layer) {
...
+
+/**
+ * Options to wtms protocol
+ * @typedef {Object} options_wmts
+ * @property {Attribution} attribution The intellectual property rights for the layer
+ * @property {String} name
+ * @property {String} mimetype
+ * @property {String} tileMatrixSet
+ * @property {Array.<Object>} tileMatrixSetLimits The limits for the tile matrix set
+ * @property {Number} tileMatrixSetLimits.minTileRow Minimum row for tiles at the level
```
autra: `number`

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
autra: Actually, for all the events, but particularly for the events you modify in this PR: the only things it makes sense to put in `const` (not `var` btw) are the string itself:

```js
const DATA_LOADED_EVENT = 'dataloaded';

/* later */

this.dispatchEvent(new CustomEvent(DATA_LOADED_EVENT));

/* elsewhere */

itowns.viewer.addEventListener(itowns.viewer.DATA_LOADED_EVENT, /* ... */);
```

Webpack will then warn if there is a typo.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -643,15 +689,37 @@ ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosit
...
+ * @param {boolean}  isAnimated  Indicates if animated
+ * @return {Promise}
+ */
+ApiGlobe.prototype.setCameraTargetGeoPositionAdvanced = function setCameraTargetGeoPositionAdvanced(position, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
-    return this.setCameraTargetGeoPosition(pPosition, isAnimated).then(() => {
-        const p = this.scene.currentControls().setOrbitalPosition(undefined, pPosition.heading, pPosition.tilt, isAnimated);
-        return p;
+    return this.setCameraTargetGeoPosition(position, isAnimated).then(() => {
+        if (position.range === undefined) {
```
autra: `if (position.range)`

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -478,18 +478,27 @@ function GlobeControls(camera, domElement, engine) {
...
 
+            if (enableTargetHelper) {
+                pickingHelper.visible = true;
+                cameraTargetOnGlobe.visible = true;
+            }
+
             return position;
         };
-    }());
+    }(this.pickingHelper));
```
autra: Why do you need this closure here? `this.pickingHelper` won't change, I don't see the point of capturing engineGFX, and position is not really an internal state here, as you never read from it.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -478,18 +478,27 @@ function GlobeControls(camera, domElement, engine) {
...
 
-        return function getPickingPosition(coords)
-        {
             position = engineGfx.getPickingPositionFromDepth(coords);
             engineGfx.renderScene();
 
+            if (enableTargetHelper) {
+                pickingHelper.visible = true;
+                cameraTargetOnGlobe.visible = true;
+            }
```
autra: Can you explain why you need these 2 `if` here? It looks odd to have to do that.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -669,6 +679,9 @@ function GlobeControls(camera, domElement, engine) {
         spherical.setFromVector3(offset);
         state = CONTROL_STATE.NONE;
         lastRotation = [];
+        if (enableTargetHelper) {
+            this.dispatchEvent(this.changeEvent);
```
autra, line 679: Note: typically why I'm pretty sure the dependency between c3DEngine and Controls should not exist.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1122,13 +1135,15 @@ function GlobeControls(camera, domElement, engine) {
...
     // update object camera position
-    this.updateCameraTransformation = function updateCameraTransformation(controlState)
+    this.updateCameraTransformation = function updateCameraTransformation(controlState, forceNoUpdateTarget)
     {
         const bkDamping = this.enableDamping;
         this.enableDamping = false;
         state = controlState || CONTROL_STATE.ORBIT;
         update();
-        updateCameraTargetOnGlobe.bind(this)();
+        if (forceNoUpdateTarget === undefined || forceNoUpdateTarget) {
```
autra: Please use positive logic : `forceUpdate`, `preventUpdate`, whatever you prefer. Negative naming is difficult to read and error-prone. Indeed, here when forceNoUpdateTarget is true, we *do* update the target on globe, which seems contradictory to me.

Btw, why do you need to bypass the update of the camera in some case ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1269,6 +1283,15 @@ GlobeControls.prototype.getCameraTargetPosition = function getCameraTargetPositi
 GlobeControls.prototype.setCameraTargetPosition = function setCameraTargetPosition(position, isAnimated) {
     const center = this.getCameraTargetPosition();
 
+    if (position.range) {
+        // Compensation of the altitude from the approximation of the ellipsoid by a sphere
```
autra, line 1283: Then the question is: where do we approximate the ellipsoid by a sphere that in turn make this hack necessary?

src/Renderer/c3DEngine.js
```diff
@@ -15,7 +15,7 @@ import RendererConstant from './RendererConstant';
 
 var instance3DEngine = null;
 
-function c3DEngine(scene, positionCamera, viewerDiv, debugMode, gLDebug) {
+function c3DEngine(scene, positionCamera, targetCamera, viewerDiv, debugMode, gLDebug) {
```
autra: This is specific to GlobeControls, so to a Globe. I would prefer not to add that as a parameter to c3DEngine.

src/Scene/Scene.js
```diff
@@ -39,12 +39,17 @@ const RENDERING_ACTIVE = 1;
  * - remove debug boolean, replace by if __DEBUG__ and checkboxes in debug UI
  * - Scene (and subobjects) should be instanciable several times.
  */
-function Scene(crs, positionCamera, viewerDiv, debugMode, gLDebug) {
+function Scene(crs, coordinate, viewerDiv, debugMode, gLDebug) {
```
autra: Why this change? you don't need it and it's less explicit for the caller.

src/Scene/Scene.js
```diff
@@ -39,12 +39,17 @@ const RENDERING_ACTIVE = 1;
...
     if (instanceScene !== null) {
         throw new Error('Cannot instantiate more than one Scene');
     }
     this.referenceCrs = crs;
 
+    const positionCamera = coordinate.as(crs);
+    const coordinateTargetCamera = coordinate.clone();
+    coordinateTargetCamera._values[2] = 0;
```
autra: undescored properties are meant to be considered as private. You can't use it here. You either need to move this code or to add a function to Coordinate (`fromLonLatHeight` maybe)

autra (changes requested):

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: It is necessary to distinguish range equal zero and undefined

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: if, there are targets helper, we must hide these targets for the picking

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: Yes indeed

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: Yes it is removed in PR #239

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: Because the target must wait for the data to be loaded (in particular the relief data)

src/Renderer/c3DEngine.js
gchoqueux: Yes, positionCamera and targetCamera are specific GlobeControl
There are removed in PR #239

src/Main.js
```diff
@@ -4,13 +4,14 @@
...
 
-import ApiGlobe from './Core/Commander/Interfaces/ApiInterface/ApiGlobe';
+import ApiGlobe, { INITIALIZED_EVENT } from './Core/Commander/Interfaces/ApiInterface/ApiGlobe';
 // browser execution or not ?
 const scope = typeof window !== 'undefined' ? window : {};
 const itowns = scope.itowns || {
     viewer: new ApiGlobe(),
 };
 scope.itowns = itowns;
+itowns.viewer.INITIALIZED_EVENT = INITIALIZED_EVENT;
```
gchoqueux: @autra how export INITIALIZED_EVENT?

gchoqueux:
@autra PR updated

src/Main.js
autra: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

autra: And you want to have a look at Main.js

gchoqueux:
@autra PR updated

src/Renderer/c3DEngine.js
autra: Where?

autra:
@gchoqueux still some unanswered comments (https://github.com/iTowns/itowns2/pull/273/files/eaea9f8672ed35e22a10deca40bb8bb1b3a5b813#diff-df61a67eee71572718ef066564a3a0af). Thanks.

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: if there are  targets helper, we hide targets helper before picking
and after picking if there are  targets helper we display targets helper for rendering

gchoqueux: The use of THREE.js layers in PR @239 will simplify this point

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -322,13 +367,9 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
...
     viewerDiv.addEventListener('globe-built', () => {
-        if (sceneIsLoaded === false) {
+        if (!sceneIsLoaded) {
             sceneIsLoaded = true;
-            this.scene.currentControls().updateCameraTransformation();
-            this.scene.updateScene3D();
-            viewerDiv.dispatchEvent(eventLoaded);
-        } else {
-            viewerDiv.dispatchEvent(eventError);
+            viewerDiv.dispatchEvent(new CustomEvent(DATA_LOADED_EVENT));
```
autra: The event DATA_LOADED_EVENT is only listened to in this function. So you can remove it and put the 2 calls in the listener here. We don't need to add an event.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1122,13 +1131,15 @@ function GlobeControls(camera, domElement, engine) {
...
     // update object camera position
-    this.updateCameraTransformation = function updateCameraTransformation(controlState)
+    this.updateCameraTransformation = function updateCameraTransformation(controlState, preventUpdate)
     {
         const bkDamping = this.enableDamping;
         this.enableDamping = false;
         state = controlState || CONTROL_STATE.ORBIT;
         update();
-        updateCameraTargetOnGlobe.bind(this)();
+        if (preventUpdate === undefined || preventUpdate) {
```
autra: If preventUpdate is true, `updateCameraTargetOnGlobe` is called, which seems contradictory to me.

Did you want to 
```js
if (!preventUpdate) { // 
   // update
}
```

Or even better, do you know javascript [now has default parameters value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).

I then suggest to have the following signature: 
```js
function updateCameraTransformation(controlState, updateCameraTarget = true)
```

 and simply use `if (updateCameraTarget)` here.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -478,18 +479,22 @@ function GlobeControls(camera, domElement, engine) {
...
+        const position = engine.getPickingPositionFromDepth(coords);
+        engine.renderScene();
 
-            return position;
-        };
-    }());
+        if (enableTargetHelper) {
+            pickingHelper.visible = true;
+            cameraTargetOnGlobe.visible = true;
+        }
```
autra, line 494: Can you elaborate on these 2 ifs? I see no difference when I remove them. The `targetHelpers` are the axis on the globe where you clicked and where the target is, right? Picking is done with s+click right?

autra (changes requested):
2 last remarks, one question, and we're good to go.

Before I merge, please split your review commits, and squash them in the relevant commits.

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: if 'enableTargetHelper' is enable there's error in picking position because helpers write in depth buffer. Picking position is do every time you click on globe.  s+click right is for selection.

> The targetHelpers are the axis on the globe where you clicked and where the target is, right

yes

gchoqueux:
updated

autra:
LGTM. A few nitpicks on commit messages with proposals:

>  fix (Camera) : fix setRange() doesn't return Promise
> 
>  return Promise in setRange()

Change to:
> Fix (Camera): Return a promise in setRange

In third commit:

> hide helperViewer when pick position

to 
> hide helperViewer when picking position to avoid wrong depth testing

In fifth:

> fix (GlobeControls) : fix initial target position is incorrect

to 
> Fix (GlobeControls): Correct initial target position

In sixth: there is a obsolete explanation about data-loaded (removed), and change:

> Fix (Camera): fix update target position after setRange()

to 
> Fix (Camera): fix target position update after setRange()

(if I understand well what you meant)

gchoqueux:
commits message updated

autra (approved):

autra:
Thanks!

PR #276, opened 2017-03-22
Api tests

pjjmunier:


peppsac:
This PR needs a big update.

Note that I'm not a big fan of generic error messages. I prefer customized ones, that offer help to the user on how to fix their error. For instance:
```js
Parameter(s) "%var%" not correctly formatted
```
is less heplful than:
```js
Parameter XXX not correctly formatted (expected "longitude,latitude,altitude" but got "12, 3")
```

So in my opinion you should drop `ErrorFormater` and `ParameterError` and use the template literals syntax.

PR #269, opened 2017-03-01
New events and events corrections

vcoindet:
**Adding the following events :**
- animationstarted
- animationstopped
- zoomchanged

**Improvements on the following events :**
- centerchanged (called through user interaction)
- orienationchanged (called through user interaction, access to oldOrientation)
- layerchanged (access to old and state of the layer)
- layerchanged:visible (access to oldvisibility)
- layerchanged:opacity (access to oldopacity)
- layerchanged:index (access to oldindex)
- layeradded (access to the id of the layer added)

PR #277, opened 2017-03-24
`camera-target-updated` (was `center-changed`) event 

autra:
This PR correctly set and fire a `camera-target-updated` when the target of the camera on the Globe changes. This event was called `center-changed`, but arguably the new name is clearer.

This is WIP because it's based on peep's work in #239. As #239 is a core work, let's try to merge it quickly please!

Notifying @pjjmunier @elias75015 and @gcebelieu this is the first PR fixing the events you need, according to [this list](https://github.com/iTowns/itowns2/issues/272#issuecomment-287779100). Don't hesitate to test!

Notifying also @vcoindet : I believe this PR overlaps a bit was you did in #269, but it is only about one event, and pushed things a bit farther up (notably: it does not only fire when explicitly calling the API, but also when user interacts with the mouse, which is the intended behaviour I believe).

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -339,7 +339,9 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
 
     var coordinate = new C.EPSG_4326(coordCarto.longitude, coordCarto.latitude, coordCarto.altitude);
 
-    this.scene = Scene(coordinate, viewerDiv, debugMode, gLDebug);
+    // FIXME: the scene is not really in EPSG:4978 atm, some axis are inverted, see
+    // https://github.com/iTowns/itowns2/pull/246
+    this.scene = Scene('EPSG:4978', coordinate, viewerDiv, debugMode, gLDebug);
 
     var map = new Globe(gLDebug);
 
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -19,7 +19,6 @@ import { STRATEGY_MIN_NETWORK_TRAFFIC } from '../../../../Scene/LayerUpdateStrat
 var sceneIsLoaded = false;
 var eventLoaded = new CustomEvent('globe-loaded');
 var eventRange = new CustomEvent('rangeChanged');
-var eventCenter = new CustomEvent('centerchanged');
 var eventOrientation = new CustomEvent('orientationchanged');
 var eventPan = new CustomEvent('panchanged');
 var eventLayerAdded = new CustomEvent('layeradded');
@@ -439,9 +438,9 @@ ApiGlobe.prototype.getCameraLocation = function getCameraLocation() {
  * @return {Position} position
  */
 
-ApiGlobe.prototype.getCenter = function getCenter() {
+ApiGlobe.prototype.getCameraTargetGeoPosition = function getCameraTargetGeoPosition() {
     var controlCam = this.scene.currentControls();
-    return this.projection.cartesianToGeo(controlCam.getTargetCameraPosition());
+    return this.projection.cartesianToGeo(controlCam.getCameraTargetPosition());
 };
 
 /**
@@ -620,22 +619,22 @@ ApiGlobe.prototype.setSceneLoaded = function setSceneLoaded() {
 /**
  * Changes the center of the scene on screen to the specified coordinates.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/x06yhbq6/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
- * @constructor
- * @param {coordinates} coordinates - Properties : longitude and latitude
- * @param      {boolean}  isAnimated  Indicates if animated
- * @return     {Promise}
+ * @param {Object} coordinates - The globe coordinates in EPSG_4326 projection to aim to
+ * @param {number} coordinates.latitude
+ * @param {number} coordinates.longitude
+ * @param {number} coordinates.range
+ * @param {boolean}  isAnimated - if the movement should be animated
+ * @return {Promise} A promise that resolves when the next 'globe-loaded' event fires.
  */
-ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
+ApiGlobe.prototype.setCameraTargetGeoPosition = function setCameraTargetGeoPosition(coordinates, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
-    eventCenter.oldCenter = this.getCenter();
     const position3D = new C.EPSG_4326(coordinates.longitude, coordinates.latitude, 0)
         .as('EPSG:4978').xyz();
     position3D.range = coordinates.range;
-    return this.scene.currentControls().setCenter(position3D, isAnimated).then(() => {
+    return this.scene.currentControls().setCameraTargetPosition(position3D, isAnimated).then(() => {
         this.scene.notifyChange(1);
         return this.setSceneLoaded().then(() => {
             this.scene.currentControls().updateCameraTransformation();
-            this.viewerDiv.dispatchEvent(eventCenter);
         });
     });
 };
@@ -646,14 +645,13 @@ ApiGlobe.prototype.setCenter = function setCenter(coordinates, isAnimated) {
  * The level has to be between the [getMinZoomLevel(), getMaxZoomLevel()].
  * The zoom level and the scale can't be set at the same time.
  * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/7yk0mpn0/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
- * @constructor
  * @param {Position} pPosition - The detailed position in the scene.
  * @param      {boolean}  isAnimated  Indicates if animated
  * @return     {Promise}
  */
-ApiGlobe.prototype.setCenterAdvanced = function setCenterAdvanced(pPosition, isAnimated) {
+ApiGlobe.prototype.setCameraTargetGeoPositionAdvanced = function setCameraTargetGeoPositionAdvanced(pPosition, isAnimated) {
     isAnimated = isAnimated || this.isAnimationEnabled();
-    return this.setCenter(pPosition, isAnimated).then(() => {
+    return this.setCameraTargetGeoPosition(pPosition, isAnimated).then(() => {
         const p = this.scene.currentControls().setOrbitalPosition(undefined, pPosition.heading, pPosition.tilt, isAnimated);
         return p;
     });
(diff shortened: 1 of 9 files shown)


autra:
Btw, the event is available on the current controls (currently available at `itowns.viewer.scene.currentControls()`,  though I hope we would simplify that in the future). Amending the doc to reflect that.

gchoqueux:
> I believe this PR overlaps a bit was you did in #269 

It may have been necessary to agree with @vcoindet before running. it's the case?

autra:
> It may have been necessary to agree with @vcoindet before running. it's the case?

Oh I didn't see the last changes on @vcoindet's PR, my bad :-) Anyway, it was faster (and probably better) to show some code than explaining what I had in mind, because my PR also contains some renaming. Btw, don't worry, I don't care which PR gets merged, really :-) I do care about having something that works on master. So if you want, I can comment on @vcoindet's PR to help things move on.

gchoqueux:
OK, We will mix the code, if necessary

vcoindet:
OK I saw your code @autra, it seems to be exactly what I did in #269...
I don't care wich PR is merged, but I'd like to test your code first to be sure it does the same thing. ;-)

autra:
We can get yours merged if you want to @vcoindet, I'm completely ok with that :-) Do you want me to comment there?

elias75015:
@autra thanks for your PR. We will quickly test it, but it seems that a great part of your code is corresponding to what @vcoindet did on his PR #269.

As seen with Victor, he will mix the code of both proposals. Once this work done, we will test the event handling more deeply :)

autra:
No problem :-) btw what do you guys think about the rename of center?

elias75015:
We're okay with the rename to camera-target-updated, it makes sense :+1:  

The most important is that the event is correctly triggered and returns to the user info expected !

vcoindet:
I still didn't had time to test, but I'd rather chose your PR @autra as you already rename the center. I will remove what I've done in mine concerning the center event.
I'm OK with the new names btw, we still understand what this is about ! ;-)

autra:
Yeah I thought center was too generic...

okay, whatever you want guys :-) 

> As seen with Victor

For cases like that, it may be better to discuss  on the open (irc channels, github etc...) than in private @elias75015 . That's the way open-source projects do :-)

vcoindet:
OK I tested it. Works fine for me.

I guess we can merge this PR, I will adapte my PR in consequences.

vcoindet:
Can you please wait before merging this PR ?

We would like to discuss the point that `oldCameraTargetPosition` and `newCameraTargetPosition` are in cartesian and it seems a bit strange for a globe. I think we should be able to propose those values in degrees for this case.

gcebelieu:
Shouldn't it be in the current coordinates system of the map / globe along with the CRS id in wich it is returned ?

elias75015:
Indeed, I think the most common behaviour is to return values in the coordinates system of the map .

autra:
The logic here is to return coordinates from Three.js (which happens to be cartesian in this use-case, but might not in other examples).  The idea is also to support multiple data-sources each having its own CRS. In this case, the only common denominator is to return Three.js coordinates, and let the user convert according to his needs. Imho, it's the simplest yet most flexible solution. As [shown in `Debug.js`](https://github.com/iTowns/itowns2/pull/277/files#diff-dd325aead9f06453147e0a0a07638301R239), the conversion is super-easy anyway.

Anyway, after having thought about that, I even think we shouldn't give back any informations in our events as a first iteration of this. We don't really know which informations are relevant for the client anyway. Some client code might just want to be notified in some case, maybe some wants only the new position etc. 

So maybe the right first step would be to remove all info and let client code query `getCameraTargetPosition` if they really want coordinates in their event handler. Later, if we realize that all or most listeners also wants some other infos, we could add them. Thoughts?

autra:
Also, embedding nothing might be better for performance for events that are fired very often (during moves for instance), though I don't know yet if this performance hit is significant.

autra:
Actually, I think I can embed 2 Coordinates object in the event, It may be the best thing to do at the end and will allow conversion between different CRS.

Btw, yes, this PR is not ready to get merged, simply because it depends on another one (#239)

elias75015:
I think it is not really a problem for us if the coordinates are not directly in the CRS of the globe (geographic), if you think it is better that way. 
In theory, we can convert them pretty easily by using the "_convert" or the "cartesianToGeo"  functions (see Coordinates.js and Projection.js). We still need to test these functions.

Anyway, we should at least have coordinates informations embed in the event.

autra:
> In theory, we can convert them pretty easily by using the "_convert"

You mean `Coordinates.as`.

> Anyway, we should at least have coordinates informations embed in the event.

Well, it's not strictly necessary (because you can query them through methods in the api), but I'll embed Coordinates objects anyway.

Btw, Coordinates objects are the "right" way of using these informations in Itowns. Under the hood, they delegate to proj4 for the conversions.

gchoqueux:
> The logic here is to return coordinates from Three.js

In Ellipsoid, the logic isn't the cartesian. In itowns, ellipsoid building is based on spherical coordinate system, we build our ellipsoid in a spherical system and we convert it to cartesian for THREE.js. In this case,  WGS84 is spherical coordinate system. 

This system is not practical and will always be converted to place objects or make measurements.
The data returned must be the most suitable for the system in which we work.

autra:
> The data returned must be the most suitable for the system in which we work.

What I've said earlier is not relevant any more anyway, but please note we don't always have ellipsoid, and we don't always work in WGS84.

gchoqueux:
Yes, quite, but an adapted coordinate system must always be the case

autra:
> an adapted coordinate system must always be the case

Your sentence is a bit unclear. I *think* we agree, but can you rephrase please?

autra:
Rebased on master.

src/Core/Geographic/Coordinates.js
```diff
@@ -106,6 +106,15 @@ function _convert(coordsIn, newCrs) {
                                    cartesian.x, cartesian.y, cartesian.z);
         }
 
+        if (coordsIn.crs === 'EPSG:4978' && newCrs === 'EPSG:4326') {
+            const geo = ellipsoid.cartesianToCartographic({
+                x: coordsIn._values[0],
+                y: coordsIn._values[1],
+                z: coordsIn._values[2],
+            });
+            return new Coordinates(newCrs, geo.latitude, geo.longitude, geo.h);
```
gchoqueux: * You have reversed longitude and latitude
* `cartesianToCartographic` return in Radian and EPSG:4326 is in degree

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1162,23 +1170,23 @@ function GlobeControls(camera, domElement, engine) {
     window.addEventListener('keydown', onKeyDown.bind(this), false);
     window.addEventListener('keyup', onKeyUp.bind(this), false);
 
-    // Initialisation Globe Target and movingGlobeTarget
+    // Initialisation camera arget on globe and movingCameraTargetOnGlobe
```
gchoqueux: typo : target

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -327,7 +324,7 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(coordCarto, view
 
     var coordinate = new C.EPSG_4326(coordCarto.longitude, coordCarto.latitude, coordCarto.altitude);
 
-    this.scene = Scene(coordinate, viewerDiv, debugMode, gLDebug);
+    this.scene = Scene('EPSG:4978', coordinate, viewerDiv, debugMode, gLDebug);
```
gchoqueux: Warning, bad WGS84 axes, see #246, Y is up axis in iTowns and WGS84 Z is the up axis
Maybe add a comment to wait fix in #246
As you pay attention to the order of the axes?

utils/debug/Debug.js
```diff
@@ -219,5 +221,34 @@ function Debug(scene) {
...
+            if (newValue) {
+                eventFolder = gui.addFolder('Events');
+
+                // camera-target-updated event
+                const initialPosition = C.fromXYZ(scene.referenceCrs, controls.getCameraTargetPosition()).as('EPSG:4326');
+                state.cameraTargetUpdated = `lat: ${initialPosition.latitude()} lon: ${initialPosition.longitude()}`;
+                const cameraTargetUpdatedController = eventFolder.add(state, 'cameraTargetUpdated').name('camera-target-updated');
+                const cameraTargetListener = (ev) => {
+                    const positionGeo = ev.newCameraTargetPosition.as('EPSG:4326');
+                    state.cameraTargetUpdated = `lat: ${positionGeo.latitude()} lon: ${positionGeo.longitude()}`;
```
gchoqueux: Displays less decimal for readability

gchoqueux (changes requested):

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
autra: Not sure how it is related to my change. I only added a parameter that was hardcoded before. I don't really deal with axes there.

autra:
updated

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
gchoqueux: Yes, but it was not explicitly stated that the scene respected the 'EPSG:4978', because precisely it was not respected. So, it doesn't hurt to leave a little warning

autra: In this case, I need to think about it. Maybe then `EPSG:4978` is simply and plainly incorrect and I need to wait for #246 to get merged for this to make any sense. I won't but a `// warning: this is wrong` in the code :-) 

At any rate, no time for this now, I'll see next week.

src/Core/Geographic/Coordinates.js
```diff
@@ -106,6 +106,15 @@ function _convert(coordsIn, newCrs) {
                                    cartesian.x, cartesian.y, cartesian.z);
         }
 
+        if (coordsIn.crs === 'EPSG:4978' && newCrs === 'EPSG:4326') {
+            const geo = ellipsoid.cartesianToCartographic({
+                x: coordsIn._values[0],
+                y: coordsIn._values[1],
+                z: coordsIn._values[2],
+            });
+            return new Coordinates(newCrs, geo.longitude, geo.latitude, geo.h);
```
gchoqueux, line 115: `cartesianToCartographic` returns in radian and `'EPSG:4326'` is in DEGREE

src/Core/Geographic/Projection.js
```diff
@@ -166,33 +166,6 @@ Projection.prototype.UnitaryToLatitudeWGS84 = function UnitaryToLatitudeWGS84(v,
     return bbox.south(UNIT.RADIAN) + v * dim.y;
 };
 
-Projection.prototype.cartesianToGeo = function cartesianToGeo(position) {
-    // TODO: warning switch coord
-    var p = position.clone();
-    p.x = position.x;
-    p.y = position.z;
```
gchoqueux, line 173: You must inverse the axis
In debug console, the coordinates radian is wrong

gchoqueux (changes requested):

autra:
@gchoqueux updated.

gchoqueux (approved):

PR #279, opened 2017-04-03
Fix tile's geometric error 

gchoqueux:
The geometric error is calculated to have a correct texture display.
For the projection of a texture's texel to be less than or equal to one pixel
To be completely correct one would have to take into account the number of segments of the tile but there will be losses of performance

This PR also fixes the getZoomLevel

### DIFF
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -924,48 +924,11 @@ ApiGlobe.prototype.getZoomScale = function getZoomScale(pitch) {
     // TODO: Why error div size height in Chrome?
     // Screen pitch, in millimeters
     pitch = (pitch || 0.28) / 1000;
-
-    // To compute scale, we must to calculate the maximum vertical distance (in meter) perceived by the camera
-    // the maximum vertical distance 2xHS (look at the explanations below 'HS segment')
-    // There's two state
-    //     * Globe is inside the frustrum camera
-    //     * Globe intersects with the frustrum camera
     const camera = this.scene.currentCamera();
-    const center = this.scene.currentControls().getCameraTargetPosition();
-    const rayon = center.length();
-    const range = center.distanceTo(camera.camera3D.position);
-    // compute distance camera/globe's center
-    const distance = rayon + range;
-    // Three points C,G and S
-    // C : Camera's position
-    // G : Globe's center
-    // S : The furthest interesection[camera verical frustrum, globe surface] from line CG
-    // HS is triangle CSG's altitude going through S and H is in GC segment
-    // alpha is angle GCS
-    // phi is angle CSG
-    const alpha = camera.FOV / 180 * Math.PI * 0.5;
-    const phi = Math.PI - Math.asin(distance / rayon * Math.sin(alpha));
-    // projection is projection segment HS on camera
-    let projection;
-
-    if (isNaN(phi)) {
-        // Globe is inside the frustrum camera
-        projection = distance * 2 * Math.tan(alpha);
-    } else {
-        // Globe intersects with the frustrum camera
-
-        // develop operation
-        // {
-        //     var beta = Math.PI - ( phi + alpha);
-        //     projection = rayon * Math.sin(beta) * 2.0;
-        // }
-        // factorisation ->
-        projection = 2.0 * rayon * Math.sin(phi + alpha);
-    }
-
-    const zoomScale = camera.height * pitch / projection;
-
-    return zoomScale;
+    const FOV = camera.FOV / 180 * Math.PI * 0.5;
+    // projection one unit on screen
+    const unitProjection = camera.height / (2 * this.getRange() * Math.tan(FOV));
+    return pitch * unitProjection;
 };
 
 /**
@@ -986,45 +949,11 @@ ApiGlobe.prototype.getRangeFromScale = function getRangeFromScale(zoomScale, pit
     // Screen pitch, in millimeters
     pitch = (pitch || 0.28) / 1000;
 
-    // To set scale, we must to calculate the maximum vertical distance (in meter) perceived by the camera
-    // the maximum vertical distance 2xHS (look at the explanations below 'HS segment')
-    // projection is projection segment HS on camera
-    // There's two state
-    //     * Globe is inside the frustrum camera
-    //     * Globe intersects with the frustrum camera
-
     const camera = this.scene.currentCamera();
-    const projection = camera.height * pitch / zoomScale;
-    const rayon = this.scene.currentControls().getCameraTargetPosition().length();
     const alpha = camera.FOV / 180 * Math.PI * 0.5;
-    // distance camera/globe's center
-    let distance;
-    // Three points C,G and S
-    // C camera's position
-    // G globe's center
-    // S = the furthest interesection[camera verical frustrum, globe surface] from line CG
-    // HS is triangle CSG's altitude going through S and H is in GC segment
-    // alpha is angle GCS
-    // phi is angle CSG
-    // beta is angle SGC
-    const sinBeta = projection / (2 * rayon);
+    // Invert one unit projection (see getZoomScale)
+    const range = pitch * camera.height / (zoomScale * 2 * Math.tan(alpha));
 
-    if (sinBeta < 1.0) {
-        // Globe is inside the frustrum camera
-        const beta = Math.asin(sinBeta);
-        // develop operation
-        //  {
-        //      let phi = Math.PI - ( beta + alpha);
-        //      distance  = rayon * Math.sin(phi) / Math.sin(alpha) ;
-        //  }
-        //  factorisation ->
-        distance = rayon * Math.sin(beta + alpha) / Math.sin(alpha);
-    } else {
-        // Globe is inside the frustrum camera
-        distance = rayon / Math.tan(alpha) * sinBeta;
-    }
-
-    const range = distance - rayon;
     return range;
 };
 
--- a/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
+++ b/src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
@@ -473,7 +473,7 @@ ApiGlobe.prototype.createSceneGlobe = function createSceneGlobe(globeLayerId, co
         }
     };
 
-    const SSE_SUBDIVISION_THRESHOLD = 6.0;
+    const SSE_SUBDIVISION_THRESHOLD = 1.0;
 
 
     // init globe layer with default parameter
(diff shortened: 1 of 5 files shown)


src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -688,7 +688,10 @@ ApiGlobe.prototype.pan = function pan(pVector) {
  * @return     {number}  The zoom level.
  */
 ApiGlobe.prototype.getZoomLevel = function getZoomLevel() {
-    return this.scene.getMap().getZoomLevel();
+    const theoricalZoom = this.scene.getMap().computeTheoricalZoom(this.scene.currentCamera(), this.getRange());
+    // const idNode = this.scene.gfxEngine.screenCoordsToNodeId();
+    // const realZoom = this.scene.getMap().getZoomLevelFromIdnode(idNode);
+    return theoricalZoom;
```
elias75015: The getZoomLevel method works well when we move the view with the mouse and keyboard controls.
However, I see 2 points :
- when I do a setZoomLevel(1) and then a getZoomLevel(), it returns 2. 

- I think the zoom levels should begin to 0 and finish to 21, ie  minZoomLevel should be 0 instead of 1.

qdnguyen: >when I do a setZoomLevel(X) and then a getZoomLevel(), it does not always return X, but sometimes X+1 (ex : setZoomLevel(1) -> getZoomLevel() = 2

In 3D, there is no concept of getZoomLevel and setZoomLevel as in 2D, because we can have many LoDs in scene depending on tile position. If we must implement set and getZoomLevel, it may  have the problem as you said.

elias75015: Sorry Quoc, I edited my previous comment : the problem happens only for the level 1, for the others levels, it finally seems to be fine. 
Maybe there is a little problem in the formula we use to calculate the zoom level ?

gchoqueux:
Warning bug with getZoomLevel(3) or  getRange without animation.
The camera have bad move after call getZoomLevel or getRange

gchoqueux:
@elias75015 and @pjjmunier problem fixed

I change distance max camera just to test with level 0, but there a bug atmosphere.
I revert distance max camera  before merge, But it will not be possible to reach level 0

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
elias75015: Tell me if I'm wrong : the zoom level returned is calculated by retrieving the level of the last requested tiles in the WMTS pyramid. 
How the zoom level can be calculated in the case there is no WMTS layer added to the globe ?

- It is maybe safer to calculate the zoomLevel using the scale / resolution of the view ? (see https://geoservices.ign.fr/documentation/geoservices/wmts.html#r%C3%A9solutions---niveaux-de-zoom)

For information purposes,  OpenLayers offers a way to change the default maxResolution and zoomFactor of the view, in case layers added are in other projection that Spherical Mercator (EPSG:3857) : see https://openlayers.org/en/latest/doc/tutorials/concepts.html

gchoqueux: The zoom level returned is calculated with best tile's level (in WGS84) whose projection (on screen) doesn't exceed 256 pixels. There is always tiles so no problem if there is not WMTS.

These tiles are geometric tiles.

The scheme tile geometric is WGS84, so a one WMTS image tile in WGS84 is applied to a geometric of the same level.  

On the other hand, several PM tiles of level (N+1) are applied to a geomertic tile of level N.

The scale  is also correct.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -767,6 +767,10 @@ ApiGlobe.prototype.getZoomScale = function getZoomScale(pitch) {
         projection = 2.0 * rayon * Math.sin(phi + alpha);
     }
 
+    // More simple compute
+    // const unitProjection = camera.height / (2 * range * Math.tan(alpha));
+    // const simpleZoomScale= pitch * unitProjection;
+
```
peppsac: This commit can be dropped or squashed in another

src/Globe/TileMesh.js
```diff
@@ -155,6 +156,7 @@ TileMesh.prototype.setBBoxZ = function setBBoxZ(min, max) {
         var trans = this.normal.clone().setLength(delta.y);
 
         this.geometry.boundingSphere.radius = Math.sqrt(delta.x * delta.x + this.oSphere.radius * this.oSphere.radius);
+        this.geometricError = this.geometry.boundingSphere.radius / SIZE_TEXTURE_TILE;
```
peppsac: You should compute `geometricError` at a single place.
I'd suggest keeping the computation here, and add a check in `TileProvider` to make sure that `tile.geometricError` is defined.

src/Globe/Globe.js
```diff
@@ -219,7 +221,12 @@ Globe.prototype.getZoomLevel = function getZoomLevel() {
 
 
 Globe.prototype.computeDistanceForZoomLevel = function computeDistanceForZoomLevel(zoom, camera) {
-    return camera.preSSE * Math.pow(this.tiles.minLevel, (this.tiles.maxLevel - zoom + 1)) / SSE_SUBDIVISION_THRESHOLD;
+    const delta = Math.PI / Math.pow(2, zoom - 1);
+    const chord = 2.0 * 6378137 * Math.sin(delta * 0.5);
+    const radius = chord * 0.5;
+    const error = radius / 256;
+    // return camera.preSSE * Math.pow(this.tiles.minLevel, (this.tiles.maxLevel - zoom + 1)) / SSE_SUBDIVISION_THRESHOLD;
+    return camera.preSSE * error / SSE_SUBDIVISION_THRESHOLD;
```
peppsac: Remove dead code.

Also the whole function isn't really clear. What are you computing? What is `chord`? What is `256`? `6378137`?

src/Renderer/c3DEngine.js
```diff
@@ -430,11 +430,15 @@ var unpack1K = function unpack1K(color, factor) {
...
 c3DEngine.prototype.screenCoordsToNodeId = function screenCoordsToNodeId(mouse) {
+    if (mouse === undefined)
+        { mouse = new THREE.Vector2(Math.floor(this.width / 2), Math.floor(this.height / 2)); }
+
     var camera = this.camera.camera3D;
 
     camera.updateMatrixWorld();
 
     var buffer = this.renderTobuffer(mouse.x, this.height - mouse.y, 1, 1, RendererConstant.ID);
+    this.renderScene();
```
peppsac, line 405: This fix may be useful, but I don't understand how it's related to the current commit?
(I don't need this change to be removed though)

peppsac (changes requested):
Thanks for the PR.

Open question: is the `geometricError` property really needed? (you could simply have a local variable in `computeNodeSSE` instead)

gchoqueux:
> Open question: is the geometricError property really needed? (you could simply have a local variable in computeNodeSSE instead)

`geometricError` isn't unique, it depends on the size of the tile and the error of the data

src/Globe/TileMesh.js
```diff
@@ -159,10 +161,17 @@ TileMesh.prototype.setBBoxZ = function setBBoxZ(min, max) {
...
         this.geometry.boundingSphere.radius = Math.sqrt(delta.x * delta.x + this.oSphere.radius * this.oSphere.radius);
+        this.updateGeometricError();
         this.centerSphere = new THREE.Vector3().addVectors(this.oSphere.center, trans);
     }
 };
 
+TileMesh.prototype.updateGeometricError = function updateGeometricError() {
+    // The geometric error is calculated to have a correct texture display.
+    // For the projection of a texture's texel to be less than or equal to one pixel
+    this.geometricError = this.geometry.boundingSphere.radius / SIZE_TEXTURE_TILE;
```
peppsac, line 166: Is our `geometricError` the same concept than 3d-tiles' `geometricError`?
It looks very similar, and maybe we should adopt the same semantic.
For 3d-tiles, the definition is: `The geometricError property is a nonnegative number that defines the error, in meters, introduced if this tile is rendered and its children are not`.
Maybe you should explain the logic of the formula used, but IMHO:
- depending on a texture size to compute a geometric error is a bit strange
- on the other hand, why doesn't this depends on the number of segments / vertices of a given node? (a node with more segments can be more precise)

Here's a drawing showing `geometricError` for 3d-tiles (in black) ([source](http://cesiumjs.org/publications.html#rendering-the-whole-wide-world)):
capture d ecran de 2017-05-04 08-49-51

gchoqueux: Yes is the same, 

Overall, the geometric error for DMT could be SizeTile / segment Count. 
But with the method, the SSE SUBDIVISION THRESHOLD must be larger, because the projection of a segment will be at the maximum of a pixel. We would then have much more segment to display.

Here, as the images are represented in texels (image + geometry). This calculation considers that a texel must not exceed one pixel.

This calculation of meaning because the other on the master does not make sense.
The best would be to make it parametric

src/Process/GlobeTileProcessing.js
```diff
@@ -104,7 +106,7 @@ export function globeSubdivisionControl(minLevel, maxLevel, sseThreshold) {
 
         const sse = computeNodeSSE(context.camera, node);
 
-        return sseThreshold < sse;
+        return SSE_SUBDIVISION_THRESHOLD < sse;
```
peppsac, line 109: I don't see the point of this change ? (and the 2 above as well)

src/Globe/TileMesh.js
peppsac: Not 100% sure I agree but I'll trust you on this one :)

src/Process/GlobeTileProcessing.js
```diff
@@ -132,3 +134,30 @@ export function globeSchemeTileWMTS(type) {
...
+    return isNaN(zoom) ? 0 : Math.round(zoom);
+}
+
+export function computeDistanceCameraFromTileZoom(zoom) {
+    const delta = Math.PI / Math.pow(2, zoom);
+    const circleChord = 2.0 * ellipsoidSizes().x * Math.sin(delta * 0.5);
+    const radius = circleChord * 0.5;
+    const error = radius / SIZE_TEXTURE_TILE;
+
+    return preSSE * error / (SSE_SUBDIVISION_THRESHOLD * 0.5) + radius;
```
peppsac, line 162: Oh. I guess that's why you store `SSE_SUBDIVISION_THRESHOLD` above.

peppsac (approved):
Ok, merging. Thanks.

PR #283, opened 2017-04-07
Webpack2

autra:
Webpack 1 is deprecated, so let's update. First commit fixes the indentation of `webpack.config.js`, so looking at second commit only might be better to review changes.

@iTowns/reviewers please r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
   "homepage": "https://itowns.github.io/",
   "dependencies": {
     "babel-runtime": "^6.20.0",
-    "custom-event": "^1.0.1",
     "earcut": "^2.1.1",
     "es6-promise": "^4.0.5",
     "js-priority-queue": "^0.1.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "babel-core": "^6.22.1",
     "babel-loader": "^6.2.10",
```
tbroyer: Update all Webpack loaders to their latest version that explicitly support Webpack 2? (eslint-loader → 1.7.1 ; imports-loader → 0.7.1) Should we use the 7.0.0-beta.1 for babel-loader or live with the deprecation warning with 6.x?

tbroyer (changes requested):

package.json
autra: Resuming work on this. Yes, I'm upgrading everything. babel-loader 7 is now stable fortunately.

tbroyer (approved):

PR #284, opened 2017-04-10
Initialization imagery and elevation layer

gchoqueux:
Move all initializations of imagery and elevation, in dedicated functions.
Because the tiles wasn't initialized to add new layer after globe loaded.


gchoqueux:
moved in #244

PR #287, opened 2017-04-13
Jsdoc new template & new examples

vcoindet:
Two points in this PR : 

- Adding a new template (minami) for jsdoc
   -> removing all `@constructor` in ApiGlobe so all methods will now be considered as methods and not classes by jsdoc
   -> Modification in package.json, please make sure this is correct as I have no real way to test it.

- Creating new JSFidlle examples embedded in ApiGlobe

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc src/Core/Commander/Interfaces/ApiInterface/*",
-    "doclint": "npm run doc -- -t templates/silent",
+    "doclint": "npm run doc -- -t ../minami",
     "test": "npm run lint && npm run build && npm run doclint",
     "build": "webpack -p",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
@@ -56,6 +56,7 @@
     "eslint-plugin-import": "^2.2.0",
     "imports-loader": "^0.7.0",
     "jsdoc": "^3.4.3",
+    "minami": "^1.0.24",
     "raw-loader": "^0.5.1",
     "webpack": "^1.14.0",
     "webpack-dev-server": "^1.16.2"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc src/Core/Commander/Interfaces/ApiInterface/*",
-    "doclint": "npm run doc -- -t templates/silent",
+    "doclint": "npm run doc -- -t ../minami",
```
autra, line 9: The template should be in the `doc` script, not doclint. This is the one that will get deployed, see `autodeploy.sh`.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
```diff
@@ -128,7 +130,7 @@ ApiGlobe.prototype.addImageryLayerFromJSON = function addImageryLayerFromJSON(ur
 
 /**
  * This function adds an imagery layer to the scene using an array of JSON files. The layer id must be unique. The protocol rules wich parameters are then needed for the function.
- * @constructor
+ * <iframe width="100%" height="400" src="//jsfiddle.net/iTownsIGN/763trrk8/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
```
autra, line 133: Nice! Can you put an explicit `https` scheme? It would allow to test locally, when we access the html with the `file:///` scheme

package.json
autra: And the `doclint` can keep the `silent` template, it would suppress its output, as its goal is just to test that the jsdoc is correct. (hint: it's correct only in ApiGlobe. We need to reactivate it for every js files, which would require some work to correct invalid syntax)

autra (changes requested):
Thanks for the effort! It's a very good first step toward a nice documentation. It's a mandatory step for releasing, although not the most exciting to do...

After this get merged, the doc TODO would be:
- generate doc for every js files (at least those exposed, but it should be most of them)
- correct the syntax (because jsdoc will fail on some)
- correct the documentation, for example in this file, there are some incorrect parameters name.

I'm just putting this TODO here in case someone wants to continue the hard work :-)

package.json
vcoindet: So I guess, you would like me to do something like this : 

```
"doc": "jsdoc src/Core/Commander/Interfaces/ApiInterface/* -t node_modules/minami/",
```

I guess the part  `-t node_modules/minami/` could be put in `autodeploy.sh` too, I don't know what is the best.

src/Core/Commander/Interfaces/ApiInterface/ApiGlobe.js
vcoindet: `https` won't work as the itowns lib is loaded in http in JSFiddle...
But the `HTTP` protocol works fine, so I can put 'http' on every URLs. WDYT ?

PR #288, opened 2017-05-02
Perf improvment

peppsac:
Quick PR that improves itowns perf with 2 minor fixes.

1st commit : reduce the redraw count, making itowns use less CPU without affecting the visual quality
2nd commit : simplification of command cancellation function, thanks to PR #244


gchoqueux (approved):
ok

PR #291, opened 2017-05-03
WMTS tilematrixsetlimits

peppsac:
Use tileMatrixSetLimits in WMTS_Provider tileInsideLimit function.

At some point elevation special code to handle targetLevel should probably be used for color as well.


gchoqueux (approved):
OK, 
It will be necessary to refactor the part on targetLevel

PR #294, opened 2017-05-09
feature: replace TileMesh OBBHelper by a debug layer

peppsac:
With this PR OBBHelper are now living in a separate layer, and we can use three.js `layers` feature to toggle visibility.


utils/debug/Debug.js
```diff
@@ -279,5 +272,90 @@ function Debug(scene) {
             }
         };
     })());
+
+    // hook that to scene.update
+    const oldUpdate = Object.getPrototypeOf(scene).update.bind(scene);
```
gchoqueux, line 277: It's a little tricky
Need to add two custom functions in scene: preUpdate () and postUpdate ()

gchoqueux (approved):

PR #299, opened 2017-05-10
fix (ellipsoid) : error longitude in compute Cartesian to Cartographic

gchoqueux:
Fix error in #246

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -12,8 +12,8 @@ import Ellipsoid from '../Math/Ellipsoid';
 export function ellipsoidSizes() {
     return {
         x: 6378137,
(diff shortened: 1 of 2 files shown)


gchoqueux:
@nosy-b ?

nosy-b:
Ok, but as this function can be used many time to convert arrays of coordinates, should we try to optimize it a bit using factorization?
`e*e` is computed 5 times for example

gchoqueux:
updated
and add fix on wrong ellipsoid size

gchoqueux:
@nosy-b 
updated
and add fix on wrong ellipsoid size

nosy-b (approved):
I tested using an external converter for Cartesian to Geodetic and reverse, that sounds good.

Issue #185, opened 2016-09-02
3D Tiles support


vpicavet:
The 3D Tiles specification is starting to take shape and approaches a stable version : 
- https://github.com/AnalyticalGraphicsInc/3d-tiles

As the specification also enters OGC proposed work as community standard, it will be important to support this protocol at some point.
- https://www.gim-international.com/content/news/ogc-proposed-work-item-for-3d-tiles-as-community-standard

The spec will take the following parts into account for planned version 1.0 : 
- [x] tileset.json
- [x] Batched 3D Model (b3dm)
- [ ] Instanced 3D Model (i3dm)
- [x] Point Cloud (pnts)
- [ ] Vector Data
- [ ] Composite (cmpt)
- [ ] Declarative Styling

Cesium already implements the draft version of the specification : 
- https://github.com/AnalyticalGraphicsInc/cesium/issues/3241

Some issues which must be solved before being able to implement 3D Tiles :  
- [x] Vector data clamping #181 ( and related vector tiles support #182 )
- [x] Tile creation and update #71 
- [x] gltf support in THREE.js

Do not hesitate to comment and add information on this large issue.

Jeremy-Gaillard:
Working on it, starting with the tileset.json import.

Jeremy-Gaillard:
tileset.json support:
- [x] asset
  - [x] version
  - [x] tilesetVersion
  - [x] gltfUpAxis
- [x] bounding volume
  - [x] box
  - [x] region
  - [x] sphere
- [x] geometric error
- [x] viewer request volume
- [x] refine (add or replace)
- [x] transform
- [x] content
  - [x] bounding volume
  - [x] url
     - ~~additional tileset.json are not supported~~ Done
     - ~~the url extension should be used to determine the tile type~~ 3d-tiles specification has been updated to remove this constraint
- [x] children

Known issues:
- [x] the transformation matrix should apply to the box and sphere bounding volumes and viewer request volumes
- [ ] if a viewer request volume is defined, the loading of the content should be delayed until the camera is in the volume (currently, only the culling part is done)

Jeremy-Gaillard:
Batched 3D Model (b3dm) support:
- [x] Geometry
- [x] Batch table
- [ ] Feature table

Additional issues:
- Three doesn't seem able to handle the batchId shader attribute

vpicavet:
@nikosaul is working on this issue

qdnguyen:
[...]

PR #304, opened 2017-05-17
3d-tiles

Jeremy-Gaillard:
Early support for 3d-tiles. To try the demo (examples/3dtiles.html) you need to run the 3d-tiles sample server provided [here](https://github.com/AnalyticalGraphicsInc/3d-tiles-samples).

PR for visibility and feedback, I'm still working on a few bugs.

What coverage of the 3d-tiles spec should we have before merging? See #185 for progress.

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -42,73 +42,50 @@
 
             itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(result => globe.addLayer(result));
 
-            // TODO: temporary, move elsewhere
-            fetch('http://localhost:8003/tilesets/TilesetWithDiscreteLOD/tileset.json').then((response) => {
-                response.json().then((tileset) => {
-                    const tileset2dict = function(root, dict) {
-                        let id = 0;
-                        const recurse = function(node) {
-                            dict[id] = node;
-                            node.tileId = id;
-                            id++
-                            for(let child in node.children) {
-                                recurse(node.children[child], dict)
-                            }
-                        }
-                        recurse(root);
+            // add 3d-tiles layer
+            const $3dTilesLayer = new itowns.GeometryLayer('3d-tiles');
+
+            $3dTilesLayer.preUpdate = (context, layer) => {
+                if(layer.root === undefined) {
+                    itowns.init3dTilesLayer(context, layer);
+                    return [];
+                }
+                itowns.pre3dTilesUpdate(context, layer);
+                return [layer.root];
+            };
+            $3dTilesLayer.update = itowns.process3dTilesNode(
+                itowns.$3dTilesCulling,
+                itowns.$3dTilesSubdivisionControl
+            );
+            $3dTilesLayer.url = 'http://localhost:8003/tilesets/TilesetWithDiscreteLOD/tileset.json';
+            $3dTilesLayer.protocol = '3d-tiles'
+            $3dTilesLayer.overrideMaterials = true;  // custom cesium shaders are not functional
+
+            itowns.View.prototype.addLayer.call(globe, $3dTilesLayer);
+
+            // debug
+            const debug3dt = new itowns.GeometryLayer('debug-3d-tiles');
+
+            debug3dt.update = function(context, layer, node) {
+                var n = node.children.filter(n => n.layer == debug3dt.id);
+                if (node.visible) {
+                    if (n.length == 0) {
+                        n = new itowns.THREE.BoxHelper(node.boundingVolume.box);
+                        n.layer = debug3dt.id;
+                        node.add(n);
+                        n.updateMatrix();
+                        n.updateMatrixWorld();
+                    } else {
+                        n = n[0];
                     }
+                    n.visible = true;
+                }
+                if (n.length > 0) {
+                    n[0].visible = false;
+                }
+            }
 
-                    let tileIndex = {};
-                    tileset2dict(tileset.root, tileIndex);
-
-                    // add 3d-tiles layer
-                    const $3dTilesLayer = new itowns.GeometryLayer('3d-tiles');
-
-                    $3dTilesLayer.preUpdate = (context, layer) => {
-                        if(layer.root === undefined) {
-                            itowns.init3dTilesLayer(context, layer);
-                            return [];
-                        }
-                        itowns.pre3dTilesUpdate(context, layer);
-                        return [layer.root];
-                    };
-                    $3dTilesLayer.update = itowns.process3dTilesNode(
-                        itowns.$3dTilesCulling,
-                        itowns.$3dTilesSubdivisionControl
-                    );
-                    $3dTilesLayer.url = 'http://localhost:8003/tilesets/TilesetWithDiscreteLOD/';
-                    $3dTilesLayer.protocol = '3d-tiles'
-                    $3dTilesLayer.tileIndex = tileIndex;
-                    $3dTilesLayer.tilesetRoot = tileset.root
-                    $3dTilesLayer.overrideMaterials = true;  // custom cesium shaders are not functional
-
-                    itowns.View.prototype.addLayer.call(globe, $3dTilesLayer);
-
-                    // debug
-                    const debug3dt = new itowns.GeometryLayer('debug-3d-tiles');
-
-                    debug3dt.update = function(context, layer, node) {
-                        var n = node.children.filter(n => n.layer == debug3dt.id);
-                        if (node.visible) {
-                            if (n.length == 0) {
-                                n = new itowns.THREE.BoxHelper(node.boundingVolume.box);
-                                n.layer = debug3dt.id;
-                                node.add(n);
-                                n.updateMatrix();
-                                n.updateMatrixWorld();
-                            } else {
-                                n = n[0];
-                            }
-                            n.visible = true;
-                        }
-                        if (n.length > 0) {
-                            n[0].visible = false;
-                        }
-                    };
-
-                    itowns.View.prototype.addLayer.call(globe, debug3dt, $3dTilesLayer);
-                })
-            });
+            itowns.View.prototype.addLayer.call(globe, debug3dt, $3dTilesLayer);
 
 </script>
     </body>
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -1,0 +1,115 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>3d-tiles example</title>
+
+    <style type="text/css">
+            html {height: 100%}
+        body { margin: 0; overflow:hidden; height:100%}
+
+            div {
+                margin : auto auto;
+                width: 100%;
+                padding: 0;
+                height: 100%
+            }
+            #menuDiv {position: absolute; top:0px; margin-left: 0px;}
+
+
+        </style>
+        <meta charset="UTF-8">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="/node_modules/dat.gui/build/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <div id="viewerDiv2"></div>
+        <script src="/examples/GUI/GuiTools.js"></script>
+        <script src="/dist/itowns.js"></script>
+        <script src="/dist/debug.js"></script>
+        <script type="text/javascript">
+            /* global itowns,document,GuiTools*/
+
+            const positionOnGlobe = { longitude: -75.65044586, latitude: 40.062840764, altitude: 100000 };
+
+            // iTowns namespace defined here
+            const viewerDiv = document.getElementById('viewerDiv');
+
+            const menuGlobe = new GuiTools('menuDiv');
+            const globe = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            menuGlobe.view = globe;
+
+            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(result => globe.addLayer(result));
+
+            // TODO: temporary, move elsewhere
+            fetch('http://localhost:8003/tilesets/TilesetWithDiscreteLOD/tileset.json').then((response) => {
+                response.json().then((tileset) => {
+                    const tileset2dict = function(root, dict) {
+                        let id = 0;
+                        const recurse = function(node) {
+                            dict[id] = node;
+                            node.tileId = id;
+                            id++
+                            for(let child in node.children) {
+                                recurse(node.children[child], dict)
+                            }
+                        }
+                        recurse(root);
+                    }
+
+                    let tileIndex = {};
+                    tileset2dict(tileset.root, tileIndex);
+
+                    // add 3d-tiles layer
+                    const $3dTilesLayer = new itowns.GeometryLayer('3d-tiles');
+
+                    $3dTilesLayer.preUpdate = (context, layer) => {
+                        if(layer.root === undefined) {
+                            itowns.init3dTilesLayer(context, layer);
+                            return [];
+                        }
+                        itowns.pre3dTilesUpdate(context, layer);
+                        return [layer.root];
+                    };
+                    $3dTilesLayer.update = itowns.process3dTilesNode(
+                        itowns.$3dTilesCulling,
+                        itowns.$3dTilesSubdivisionControl
+                    );
+                    $3dTilesLayer.url = 'http://localhost:8003/tilesets/TilesetWithDiscreteLOD/';
+                    $3dTilesLayer.protocol = '3d-tiles'
+                    $3dTilesLayer.tileIndex = tileIndex;
+                    $3dTilesLayer.tilesetRoot = tileset.root
+                    $3dTilesLayer.overrideMaterials = true;  // custom cesium shaders are not functional
+
+                    itowns.View.prototype.addLayer.call(globe, $3dTilesLayer);
+
+                    // debug
+                    const debug3dt = new itowns.GeometryLayer('debug-3d-tiles');
+
+                    debug3dt.update = function(context, layer, node) {
+                        var n = node.children.filter(n => n.layer == debug3dt.id);
+                        if (node.visible) {
+                            if (n.length == 0) {
+                                n = new itowns.THREE.BoxHelper(node.boundingVolume.box);
+                                n.layer = debug3dt.id;
+                                node.add(n);
+                                n.updateMatrix();
+                                n.updateMatrixWorld();
+                            } else {
+                                n = n[0];
+                            }
+                            n.visible = true;
+                        }
+                        if (n.length > 0) {
+                            n[0].visible = false;
+                        }
+                    };
+
+                    itowns.View.prototype.addLayer.call(globe, debug3dt, $3dTilesLayer);
+                })
+            });
+
+</script>
+    </body>
+</html>
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -60,6 +60,8 @@
             $3dTilesLayer.url = 'http://localhost:8003/tilesets/TilesetWithDiscreteLOD/tileset.json';
             $3dTilesLayer.protocol = '3d-tiles'
             $3dTilesLayer.overrideMaterials = true;  // custom cesium shaders are not functional
+            $3dTilesLayer.type = 'geometry';
+            $3dTilesLayer.visible = true;
 
             itowns.View.prototype.addLayer.call(globe, $3dTilesLayer);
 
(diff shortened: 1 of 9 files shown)


examples/3dtiles.html
```diff
@@ -0,0 +1,119 @@
...
+                    return [];
+                }
+                itowns.pre3DTilesUpdate(context, layer);
+                return [layer.root];
+            };
+            threeDTilesLayer.update = itowns.process3DTilesNode(
+                itowns.threeDTilesCulling,
+                itowns.threeDTilesSubdivisionControl
+            );
+            threeDTilesLayer.url = 'http://localhost:8003/tilesets/TilesetWithDiscreteLOD/tileset.json';
```
peppsac: Where can I get test files like this one?

Jeremy-Gaillard: The test files are those provided by AGI in their [3d-tiles-samples repository](https://github.com/AnalyticalGraphicsInc/3d-tiles-samples). Just run the provided node server locally and it should work.

qdnguyen: i dit it, but any b3dm is loaded, how can we see the dragon model? we must move to a specific place?

Jeremy-Gaillard: The camera is a bit misplaced, you should see the model if you move slightly to the top-right of the screen. I will fix this ASAP.

Jeremy-Gaillard:
Fixed position of the camera and bad mesh orientation.

Jeremy-Gaillard:
I'd like to remove glTFLoader from the code base and add it as a dependency. The loader is currently in the examples folder of three (https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js). Any idea how I could do this? (ping @autra)

peppsac:
Hmm visibility testing is broken for the globe :s
(the problem was there before the changes from 6e5ae5bf689a5564fe22e07a12c9f27a90059bbe )

We definitely need to fix this...

peppsac:
Globe occlusion issue fixed in PR https://github.com/iTowns/itowns2/pull/307

src/Core/Scheduler/Providers/ThreeDTiles_Provider.js
```diff
@@ -0,0 +1,162 @@
...
+ */
+import * as THREE from 'three';
+import Provider from './Provider';
+import CacheRessource from './CacheRessource';
+import B3dmLoader from '../../../Renderer/ThreeExtended/B3dmLoader';
+import Fetcher from './Fetcher';
+import FeatureMaterial from '../../../Renderer/FeatureMaterial';
+
+
+export function ThreeDTilesIndex(tileset) {
```
peppsac: I think I'd prefer `_3dTilesXXXX` names over, `ThreeDTilesXXX`.

Cons:
   * `_` is supposed to be reserved for private usage
Pros:
  * names that match the procol/spec name
  * _3dTilesIndex is more readable than ThreeDTilesIndex

WDYT?

src/Renderer/Shader/FeatureVS.glsl
```diff
@@ -0,0 +1,74 @@
...
+uniform int        RTC;
+varying float      light;
+
+attribute vec3 color;
+varying vec3 vColor;
+
+// IE error : Initializer for const variable must initialize to a constant value
+//const vec3 dir =  normalize(vec3(1.0,1.0,0.5));
+
+void main()
```
peppsac: I fail to see why makes this file needed?

Jeremy-Gaillard: It's to transfer color information to the fragment shader.

I think I'll remove feature material and shaders from this PR before it's merged, they aren't really useful or relevant to this PR.

src/Core/Scheduler/Providers/ThreeDTiles_Provider.js
Jeremy-Gaillard: I agree with ThreeDTiles being an awkward name. I can't think of better names than yours, so I'll use them next time I refactor.

examples/3dtiles.html
qdnguyen: no thing shows when i move the camera, no connection between itowns and 3d-tiles sample server, have any ideas?

Jeremy-Gaillard: Is your branch up-to-date? Did you open the right demo (localhost:8080/examples/3dtiles.html)?

autra:
> I'd like to remove glTFLoader from the code base and add it as a dependency. The loader is currently in the examples folder of three (https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js). Any idea how I could do this? (ping @autra)

You can import things with a path starting with the package name. `import GltfLoader from three/examples/js/loaders/GLTFLoader` should work (I haven't tested).

examples/3dtiles.html
qdnguyen: ok, thanks! it work now, i can see dragon model! but i found some bugs, the globe disappear when we zoom in and the left mouse button is bloked. For second bug, if we use right mouse button to move the camera, after that, the left mouse come back. Anyway, good job, i will test with other models and give you more feedbacks.

src/Core/Scheduler/Providers/ThreeDTiles_Provider.js
autra: AFAIK variables in javascript can start with most unicode character. We *could* find a char that looks like a 3 in other language, it would be funny for everyone involved ;-)

Jokes apart, why not following jquery example and start with `$`? `$3DTiles` etc... It is somehow familiar for a js dev, and has the advantage of not looking like a private var.

autra: Other possibility, roman numerals: `iiiDTiles` (but I think I still prefer the `$`-prefixed version)

Jeremy-Gaillard:
@autra thanks for the answer. Unfortunately, it doesn't work perfectly.

I tried to do this:
```javascript 
import GlTFLoader from 'three/examples/js/loaders/GLTFLoader';
``` 
However, since the examples don't use ES6 imports, it only sets the value of THREE.GLTFLoader, and since THREE isn't defined in my class, it fails.

The following code works:
```javascript 
import GlTFLoader from 'three/examples/js/loaders/GLTFLoader';
...
this.glTFLoader = new THREE.GLTFLoader();
``` 
But I have a jslint error since GlTFLoader isn't used in my code. I guess I could deactivate this specific warning, but it seems a bit hacky. Do you have a better idea?

src/Core/Scheduler/Providers/ThreeDTiles_Provider.js
Jeremy-Gaillard: $3dTiles suits me. I'll make the change.

Jeremy-Gaillard:
Updated.

I think the branch is in a stable enough state to be considered for merging now.

autra:
> The following code works:
> 
> ```js
> import * as THREE from 'three';
> import GlTFLoader from 'three/examples/js/loaders/GLTFLoader';
> ...
> this.glTFLoader = new THREE.GLTFLoader();
> ```
> But I have a jslint error since GlTFLoader isn't used in my code. I guess I could deactivate this specific warning, but it seems a bit hacky. Do you have a better idea?

Yep, it works because of the `imports` loader in webpack.config.js. Also, `import  'three/examples/js/loaders/GLTFLoader'` [should work too](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and pass eslint.

Jeremy-Gaillard:
Indeed it does. Fixed it.

src/Core/Scheduler/Providers/$3dTiles_Provider.js
```diff
@@ -0,0 +1,153 @@
...
+ * Description:
+ */
+import * as THREE from 'three';
+import Provider from './Provider';
+import B3dmLoader from '../../../Renderer/ThreeExtended/B3dmLoader';
+import Fetcher from './Fetcher';
+import BasicMaterial from '../../../Renderer/BasicMaterial';
+
+
+export function $3dTilesIndex(tileset, urlPrefix) {
```
peppsac: The name `$3dTilesIndex` is fine but I'd suggest renaming the files and remove their `$` prefix.

src/Core/Scheduler/Providers/$3dTiles_Provider.js
```diff
@@ -0,0 +1,153 @@
+/**
+ * Created On: 2016-11-8
+ * Class: $3dTiles_Provider
+ * Description:
+ */
```
peppsac: Please remove this

src/Core/Scheduler/Providers/$3dTiles_Provider.js
```diff
@@ -0,0 +1,153 @@
...
+        const e = center.x + box[3];
+        const s = center.y - box[7];
+        const n = center.y + box[7];
+        const b = center.z - box[11];
+        const t = center.z + box[11];
+
+        return { box: new THREE.Box3(new THREE.Vector3(w, s, b), new THREE.Vector3(e, n, t)) };
+    } else if (boundingVolume.sphere) {
+        // TODO
+        return { sphere: undefined };
```
peppsac: Maybe all the unimplemented code path should `throw new Error()..)` instead of silently failing?

src/Core/Scheduler/Providers/$3dTiles_Provider.js
```diff
@@ -0,0 +1,153 @@
...
+
+function getBox(boundingVolume) {
+    if (boundingVolume.region) {
+        // TODO: unsupported
+        return { region: boundingVolume.region };
+    } else if (boundingVolume.box) {
+        // TODO: only works for axis aligned boxes
+        const box = boundingVolume.box;
+        const center = new THREE.Vector3(box[0], box[1], box[2]);
+        const w = center.x - box[3];
```
peppsac: Could be worth adding a comment to explain what `box` contain?
It looks like 4 (x, y, z) values, but something like:
```js
// box contains 4 (x, y, z) values, in this order: top-left, bottom-left, top-right, bottom-right
```
helps understanding the code

src/Core/Scheduler/Providers/$3dTiles_Provider.js
```diff
@@ -0,0 +1,153 @@
...
+            if (result !== undefined) {
+                let func;
+                const magic = textDecoder.decode(new Uint8Array(result, 0, 4));
+                if (magic[0] === '{') {
+                    result = JSON.parse(textDecoder.decode(new Uint8Array(result)));
+                    const newPrefix = url.slice(0, url.lastIndexOf('/') + 1);
+                    layer.tileIndex.extendTileset(result, metadata.tileId, newPrefix);
+                } else if (magic == 'b3dm') {
+                    func = supportedFormats.b3dm;
+                }
```
peppsac: Add something like:
```js
} else {
Promise.reject(`unsupported magic code ${magic}`);
}
```

src/Process/$3dTilesProcessing.js
```diff
@@ -0,0 +1,174 @@
...
+        requestNewTile(context.view, context.scheduler, layer, tileset.root, undefined).then(
+            (tile) => {
+                context.view.scene.add(tile);
+                tile.updateMatrixWorld();
+                layer.root = tile;
+            });
+    });
+}
+
+export function process3dTilesNode(cullingTest, subdivisionTest) {
```
peppsac: The whole file looks a lot like `TiledNodeProcessing`.
How about using `TiledNodeProcessing` instead? Adapt it to cover both use-case and maybe rename it `TiledStructureProcessing` ?)

examples/3dtiles.html
```diff
@@ -0,0 +1,117 @@
...
+            const globe = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            menuGlobe.view = globe;
+
+            const urlImageryLayers = [];
+            const urlElevationLayers = [];
+
+            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(result => globe.addLayer(result));
+
+            itowns.Fetcher.json('layers/JSONLayers/IGN_MNT.json').then(result => globe.addLayer(result));
+            itowns.Fetcher.json('layers/JSONLayers/IGN_MNT_HIGHRES.json').then(result => globe.addLayer(result));
```
peppsac: Are those needed?
Try to keep examples short, so I'd suggest removing eveything not mandatory for your `3d-tiles` sample:
- elevation layer(s)
- `RealisticLighting` toggle
- `globe.addEventListener(..)`

examples/3dtiles.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
+<!--
+To change this license header, choose License Headers in Project Properties.
+To change this template file, choose Tools | Templates
+and open the template in the editor.
+-->
```
peppsac: Remove this too

examples/3dtiles.html
```diff
@@ -0,0 +1,117 @@
...
+            html {height: 100%}
+        body { margin: 0; overflow:hidden; height:100%}
+
+            div {
+                margin : auto auto;
+                width: 100%;
+                padding: 0;
+                height: 100%
+                /*margin-top: 50vh;
+                transform: translateY(-50%);*/
```
peppsac: And this :)

peppsac (changes requested):
Thanks, this looks great :)

I've left a comments, all small things, except the one on $3dTilesProcessing.js.

src/Process/$3dTilesProcessing.js
Jeremy-Gaillard: I'd prefer to wait before factorizing. This PR does not cover the full tileset spec, and some of the remaining options that are not yet supported will change the process (I'm mainly thinking about the second, tight bounding box that allows to cull the geometry of the tile without culling the tile itself).

Jeremy-Gaillard:
Pushed modifications asked by @peppsac. If everything is OK, I'll rebase and the PR will be ready to be merged.

src/Renderer/ThreeExtended/B3dmLoader.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const textDecoder = new TextDecoder('utf-8');
+B3dmLoader.prototype.parse = function parse(buffer) {
+    if (!buffer) {
+        throw new Error('No array buffer provided.');
+    }
+
+    const array = new Uint8Array(buffer);
+    const view = new DataView(buffer);
```
peppsac: Using :
```js
// 4 first bytes defined magic type
const array = new Uint8Array(array, 0, 4);
const view = new DataView(buffer, 4);
```
You can remove `byteOffset += 4;` and remove the `subarray` thing.
Actually you could avoid the `array` variable completelty and do:
```js
b3dmHeader.magic = textDecoder.decode(new Uint8Array(buffer, 0, 4));
```
?

Jeremy-Gaillard:
Rebased and ready. The visibility buttons no longer work, but the problem doesn't seem to be specific to 3d-tiles, I guess a recent PR broke it?

peppsac (approved):
Looks good, merging. Thanks 👍

examples/3dtiles.html
```diff
@@ -0,0 +1,94 @@
...
+                itowns.pre3dTilesUpdate(context, layer);
+                return [layer.root];
+            };
+            $3dTilesLayer.update = itowns.process3dTilesNode(
+                itowns.$3dTilesCulling,
+                itowns.$3dTilesSubdivisionControl
+            );
+            $3dTilesLayer.url = 'http://localhost:8003/tilesets/TilesetWithDiscreteLOD/tileset.json';
+            $3dTilesLayer.protocol = '3d-tiles'
+            $3dTilesLayer.overrideMaterials = true;  // custom cesium shaders are not functional
```
elemoine, line 62: How is this related to Cesium?

examples/3dtiles.html
```diff
@@ -0,0 +1,94 @@
...
+            $3dTilesLayer.type = 'geometry';
+            $3dTilesLayer.visible = true;
+
+            itowns.View.prototype.addLayer.call(globe, $3dTilesLayer);
```
elemoine, line 66: Why not just do `globe.addLayer($3dTilesLayer)`?

examples/3dtiles.html
```diff
@@ -0,0 +1,94 @@
...
+                        n = n[0];
+                    }
+                    n.visible = true;
+                }
+                if (n.length > 0) {
+                    n[0].visible = false;
+                }
+            }
+
+            itowns.View.prototype.addLayer.call(globe, debug3dt, $3dTilesLayer);
```
elemoine, line 90: Any reason for adding `$3dTilesLayers` twice?

elemoine (commented):
A few post-merge questions, if you like answering them :)

examples/3dtiles.html
Jeremy-Gaillard: Cesium uses specific uniforms in its shader to record certain parameters (the sun's position for example). The 3d-tiles samples given by AGI use these uniforms in their shader.

The optimal solution would be to have a compatibility mode where we handle the cesium uniforms in our materials. In the meantime, I just implemented a way to override the 3d-tiles materials.

examples/3dtiles.html
Jeremy-Gaillard: The GlobeView class [overrides the addLayer function](https://github.com/iTowns/itowns2/blob/master/src/Core/Prefab/GlobeView.js#L256). Instead of adding a layer to the scene, it attaches the layer to the globe tile layer.

If we want to add a layer that is independent from the tile layer, we have to use this syntax.

examples/3dtiles.html
Jeremy-Gaillard: This line doesn't add `$3dTilesLayers` again, it attaches `debug3dt` to `$3dTilesLayers`.

examples/3dtiles.html
elemoine: Ok. I find this awkward and hard to understand to example reader.

elemoine:
Thank you for your answers @Jeremy-Gaillard.

src/Process/3dTilesProcessing.js
```diff
@@ -0,0 +1,174 @@
...
+
+let preSSE;
+export function pre3dTilesUpdate(context) {
+    // pre-sse
+    const hypotenuse = Math.sqrt(context.camera.width * context.camera.width + context.camera.height * context.camera.height);
+    const radAngle = context.camera.FOV * Math.PI / 180;
+
+     // TODO: not correct -> see new preSSE
+    // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
+    const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
```
elemoine, line 68: From the [three.js doc](https://threejs.org/docs/index.html#api/cameras/PerspectiveCamera) I understand that `context.camera.FOV` (`radAngle`) is the vertical field of view. So I wonder why `HYFOV` is not calculated in terms of `camera.height` instead of `camera.width`:

    const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.height);

?

Also, I wonder why you use the hypotenuse FOV for the SSE calculation. Why not using the vertical FOV as in Cesium? Just curious…

src/Process/3dTilesProcessing.js
```diff
@@ -0,0 +1,174 @@
...
+function computeNodeSSE(camera, node) {
+    if (node.boundingVolume.region) {
+        // TODO
+    }
+    if (node.boundingVolume.box) {
+        // TODO: compute proper distance
+        const worldCoordinateCenter = node.boundingVolume.box.getCenter();
+        worldCoordinateCenter.applyMatrix4(node.matrixWorld);
+        const distance = camera.camera3D.position.distanceTo(worldCoordinateCenter);
+        return preSSE * (node.geometricError / distance);
```
elemoine, line 81: I also don't understand this. Could you please explain how you got to that expression? The [calculation](https://github.com/AnalyticalGraphicsInc/cesium/blob/38dab936cff23e8f99fb79a3873adfe9400d3bab/Source/Scene/QuadtreePrimitive.js#L688) in Cesium makes more sense to me, but I'd like to understand your reasoning. Thanks.

Jeremy-Gaillard: These functions are transpositions of the terrain tile SSE computations (we'll factorize them in the future). Most of this code was not originally written by me, so I may not be able to answer fully to your questions.

- I think we use the hypotenuse FOV because it is proportional to both the screen height and width
- ~~I remember checking the calculation myself, and it did match with [what I saw elsewhere](http://folk.uio.no/inftt/Div/visualization2.pdf). However, the calculation could be more precise (we compute the SSE for the point at the center of the box instead of the box itself).~~ I checked again, and it seems we're multiplying by `(2.0 * Math.tan(HYFOV * 0.5))` instead of dividing by it (like it is done in Cesium and the presentation I linked). This seems to be a mistake, do you agree @gchoqueux @peppsac?

PR #305, opened 2017-05-17
feature: allow geometry layer to implement partial update

peppsac:
Before this commit there was no link between an update starting and the cause
of this update being fired.

Now the source of notifyChange() calls are stored, and the layer.preUpdate()
method receive them as an argument.
This way preUpdate can choose a better starting point from the update.

As an example I've modified GlobeView/PlanarView to use this feature: updates
now start at the common ancestor (if any) of all the notifyChanges() sources.

This PR also modifies the debug graph to make it more useful.

Here's a screencapture of the graphs, after zooming on the globe :

capture d ecran de 2017-05-17 14-57-12

The `grey` line is the starting level of the `update` function (if level > 0 it means a partial update).
The `green` line is `update` duration in milliseconds.
One can observe that update duration is directly linked to the starting level, which makes sense :)

### DIFF
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -8,8 +8,6 @@ function MainLoop(scheduler, engine) {
     this.needsRedraw = false;
     this.scheduler = scheduler;
     this.gfxEngine = engine; // TODO: remove me
-
-    this._viewsToUpdate = new Set();
 }
 
 MainLoop.prototype = Object.create(EventDispatcher.prototype);
@@ -54,7 +52,7 @@ MainLoop.prototype._update = function _update(view) {
 
     for (const geometryLayer of view.getLayers((x, y) => !y)) {
         context.geometryLayer = geometryLayer;
-        const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer);
+        const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer, view._changeSources);
         updateElements(context, geometryLayer, elementsToUpdate);
     }
 };
@@ -81,6 +79,7 @@ MainLoop.prototype._step = function _step(view) {
         document.title = document.title.substr(0, document.title.length - 2);
     }
     this.renderingState = RENDERING_PAUSED;
+    view._changeSources.clear();
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/Core/MainLoop.js
```diff
@@ -54,14 +52,14 @@ MainLoop.prototype._update = function _update(view) {
 
     for (const geometryLayer of view.getLayers((x, y) => !y)) {
         context.geometryLayer = geometryLayer;
-        const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer);
+        const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer, changeSources);
```
gchoqueux: why not put ?
```js
const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer, view._changeSources);
```

and in `_step ` keep :

```js
this._update(view);
```

src/Core/Prefab/PlanarView.js
```diff
@@ -56,12 +56,56 @@ function PlanarView(viewerDiv, boundingbox) {
     const tileLayer = new GeometryLayer('planar');
     const initLayer = initTiledGeometryLayer(planarSchemeTile(boundingbox));
 
-    tileLayer.preUpdate = (context, layer) => {
+    function _commonAncestorLookup(a, b) {
```
gchoqueux, line 59: Why not start factoring common functions between  globe and planar?

src/Core/View.js
```diff
@@ -108,14 +110,24 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+            this._changeSources.add(changeSource);
+            this.mainLoop.scheduleViewUpdate(this, needsRedraw);
+        }, delay);
     } else {
+        this._changeSources.add(changeSource);
         this.mainLoop.scheduleViewUpdate(this, needsRedraw);
     }
 };
 
+View.prototype._readAndResetChangeSources = function _readAndResetChangeSources() {
```
gchoqueux: unused function?

gchoqueux (changes requested):

src/Core/MainLoop.js
peppsac: Indeed, changing.

src/Core/Prefab/PlanarView.js
peppsac: Let's wait a bit between factorizing again... I'm fine with small scope copy-paste because it's easier to delete than a wrong abstraction.

src/Core/View.js
peppsac: Oops, legacy code from a different implementation. Removing it.

gchoqueux (approved):

PR #306, opened 2017-05-18
refacto (GlobeView) : move specific globe's functions from ApiGlobe to GlobeView

gchoqueux:
**Move or refacto all functions from ApiGlobe.js**

* feature new function to add an editable property to a layer (and automatic dispatch event when property is changed)


src/Core/Prefab/GlobeView.js
```diff
@@ -8,18 +8,80 @@ import RendererConstant from '../../Renderer/RendererConstant';
...
+ * @property {Array.<Object>} tileMatrixSetLimits The limits for the tile matrix set
+ * @property {number} tileMatrixSetLimits.minTileRow Minimum row for tiles at the level
+ * @property {number} tileMatrixSetLimits.maxTileRow Maximum row for tiles at the level
+ * @property {number} tileMatrixSetLimits.minTileCol Minimum col for tiles at the level
+ * @property {number} tileMatrixSetLimits.maxTileCol Maximum col for tiles at the level
+ * @property {Object} [zoom]
+ * @property {Object} [zoom.min] layer's zoom minimum
+ * @property {Object} [zoom.max] layer's zoom maximum
+ */
+
```
peppsac: I don't see why you're adding these comments here?

src/Core/Prefab/GlobeView.js
```diff
@@ -186,12 +248,21 @@ function GlobeView(viewerDiv, coordCarto) {
...
-        viewerDiv.dispatchEvent(new CustomEvent('globe-built'));
+        this.dispatchEvent({ type: 'globe-built' });
     });
 
+    const fn = () => {
+        this.removeEventListener('globe-built', fn);
+        this.dispatchEvent({ type: 'initialized' });
+    };
+
+    this.addEventListener('globe-built', fn);
```
peppsac: This looks like a convoluted way to emit `initialized` on the first time `globe-built` is being fired.
I guess you could get rid of `globe-built` completely.
(and use `GLOBE_INITIALIZED` instead of `initialized`)

src/Core/Prefab/GlobeView.js
```diff
@@ -186,12 +248,21 @@ function GlobeView(viewerDiv, coordCarto) {
...
+
     window.addEventListener('resize', () => {
         this.controls.updateCamera(this.camera, this.viewerDiv.clientWidth, this.viewerDiv.clientHeight);
     }, false);
+
+    this.notifyChange(0, true);
```
peppsac, line 214: Is this really needed?

src/Core/Prefab/GlobeView.js
```diff
@@ -204,7 +275,213 @@ GlobeView.prototype.addLayer = function addLayer(layer) {
...
+    this.dispatchEvent({ type: GLOBE_VIEW_EVENTS.LAYER_ADDED });
+    return layer;
+};
+
+/**
+ * Return the list of all layers in the scene in the order of how they are stacked on top of each other.
+ * @return     {layer}  The Layers.
+ */
+// REMOVE?
+GlobeView.prototype.getImageryLayers = function getImageryLayers() {
```
peppsac: Agreed with your comment: remove this function please (it brings no value)

src/Core/Prefab/GlobeView.js
```diff
@@ -346,4 +623,147 @@ GlobeView.prototype.changeRenderState = function changeRenderState(newRenderStat
...
+ * Add an elevation layer to the map using an array of JSON files.
+ * Elevations layers are used to build the terrain.
+ * Only one elevation layer is used, so if multiple layers cover the same area, the one
+ * with best resolution is used (or the first one is resolution are identical).
+ * The layer id must be unique amongst all layers already inserted.
+ * The protocol rules which parameters are then needed for the function.
+ * @param {Layer} layer.
+ * @return     {layer}  The Layers.
+ */
+GlobeView.prototype.addElevationLayersFromJSONArray = function addElevationLayersFromJSONArray(urls) {
```
peppsac: The functions  `add*FromJSONArray` don't belong to itowns IMHO.

And I feel the same regarding the `add*Layer` actually.
iTowns provides 1 way to add a layer (`view.addLayer`) which cover all our needs - the specific functions should live in the application using itowns.

peppsac (changes requested):
Stopping my review to make a broader point.
`GlobeView` is supposed to contain code that match 2 criteria:
  - it's useful when building a globe application
  - *and* this code wouldn't make sense in a non-globe itowns application

Most of the code you moved from ApiGlobe fails to meet these criteria.

As an example, take `addImageryLayer`: either we want itowns to know about color layer, in which case GlobeView is not the correct place to add this code. Or we want to keep itowns unaware of layer types, in which case this code belong to the app.

Layers ordering on the other hand could probably lives in itowns, but not in GlobeView. Maybe this code could live in a separate file that does just this?

While I'm all for removing ApiGlobe, I don't want GlobeView to become the new ApiGlobe :)

gchoqueux:
@peppsac 
Yes, I will continue to clean

src/Core/Prefab/GlobeView.js
gchoqueux: if there's not `tileMatrixSetLimits`

src/Core/Prefab/GlobeView.js
gchoqueux: yes

src/Core/Prefab/GlobeView.js
gchoqueux: removed

src/Core/Prefab/GlobeView.js
gchoqueux: removed

src/Core/Prefab/GlobeView.js
peppsac: My original question wasn't precise enough.
Let me try again: "why is this added, since until now it worked fine without it?"
And after reading more of the changes, I can answer to my question: "Because until now, there was a call to apiGlobe.update() which actually does the this.notifyChange(..) for you"

src/Core/View.js
```diff
@@ -58,7 +59,38 @@ function View(crs, viewerDiv, mainLoop) {
...
+function _defineLayerProperty(view, layer, propertyName, defaultValue) {
+    // todo : default value, test if setter and getter exist
+    var property = layer[propertyName] == undefined ? defaultValue : layer[propertyName];
+    Object.defineProperty(layer,
+        propertyName,
+        { get: () => property,
+            set: (newValue) => {
+                if (property !== newValue) {
+                    property = newValue;
+                    const event = { type: 'layer-changed' };
```
peppsac: Maybe you could use: 
```js
const event = { type: `layer-${property}-changed` };
```
?

src/Core/View.js
```diff
@@ -58,7 +59,38 @@ function View(crs, viewerDiv, mainLoop) {
...
+    if (layer.type == 'geometry' || layer.type == 'debug') {
+        layer.threejsLayer = view.mainLoop.gfxEngine.getUniqueThreejsLayer();
+        if (layer.visible) {
+            view.camera.camera3D.layers.enable(layer.threejsLayer);
+        } else {
+            view.camera.camera3D.layers.disable(layer.threejsLayer);
+        }
+    }
+}
+
```
peppsac: I think this function could live in the `Layer.js` file.

src/Core/View.js
```diff
@@ -82,15 +114,30 @@ function _preprocessLayer(view, layer, provider) {
     // probably not the best place to do this
     if (layer.type == 'color') {
         layer.frozen = layer.frozen || false;
-        layer.visible = layer.visible == undefined ? true : layer.visible;
-        layer.opacity = layer.opacity == undefined ? 1.0 : layer.opacity;
-        layer.sequence = 0;
+        _defineLayerProperty(view, layer, 'visible', true);
```
peppsac: With this change a user can't declare a layer to be hidden by default. 
I'm fine with this but you need to warn the user that your ignoring/overriding `visible` and `opacity`

src/Core/View.js
```diff
@@ -82,15 +114,30 @@ function _preprocessLayer(view, layer, provider) {
...
+        _defineLayerProperty(view, layer, 'opacity', 1.0);
+        layer.addEventListener('layer-changed', () => view.notifyChange(0, true));
```
peppsac: I'm not sure that we want to auto-connect the property changed event(s) to `notifyChange`.

I'd leave this task to the user.

src/Core/View.js
peppsac: And you could use the same semantic than you used in `GlobeControls`: old/new values.

src/Core/View.js
```diff
@@ -82,15 +114,30 @@ function _preprocessLayer(view, layer, provider) {
...
         layer.frozen = layer.frozen || false;
+    } else if (layer.type == 'geometry' || layer.type == 'debug') {
+        _defineLayerProperty(view, layer, 'visible', true);
+        layer.addEventListener('layer-changed', (e) => {
+            if (e.visible !== undefined) {
+                if (e.visible) {
+                    view.camera.camera3D.layers.enable(layer.threejsLayer);
+                } else {
+                    view.camera.camera3D.layers.disable(layer.threejsLayer);
+                }
```
peppsac: IMHO these lines:
```js
                if (e.visible) {
+                    view.camera.camera3D.layers.enable(layer.threejsLayer);
+                } else {
+                    view.camera.camera3D.layers.disable(layer.threejsLayer);
+                }
```
should be in the property setter - not in the event listener (the only question is how to get access to the `view`).

And I'd remove the `notifyChange` (for the same reason: it's up to the user)

peppsac: Though we can keep the `notifyChange` but in this case, I'd fire them directly in the `set` function, not in an event listener.

src/Core/View.js
gchoqueux: if `layer.visible == false` the layer is hidden. The object's properties  take precedence over the default value.

src/Core/View.js
gchoqueux: to discuss.
Both should be possible

src/Core/View.js
gchoqueux: The problem is that you can't use the layer with several `View`

src/Core/Layer/Layer.js
```diff
@@ -1,3 +1,22 @@
...
+            set: (newValue) => {
+                if (property !== newValue) {
+                    const event = { type: 'layer-changed', previous: {}, new: {} };
+                    event.previous[propertyName] = property;
+                    event.new[propertyName] = newValue;
+                    property = newValue;
+                    layer.dispatchEvent(event);
+                }
+            } });
+};
```
peppsac: Atlernative proposal:
```js
export const defineLayerProperty = function defineLayerProperty(layer, propertyName, defaultValue, onChange) {
    var property = layer[propertyName] == undefined ? defaultValue : layer[propertyName];
    Object.defineProperty(layer,
        propertyName,
        {
            get: () => property,
            set: (newValue) => {
                if (property !== newValue) {
                    const event = { type: 'layer-changed', previous: {}, new: {} };
                    event.previous[propertyName] = property;
                    event.new[propertyName] = newValue;
                    property = newValue;
                    if (onChange) {
                        onChange(layer, propertyName);
                    }
                    layer.dispatchEvent(event);
                }
            },
        });
};
```
I don't like public events (`layer-changed` here) being used for internal changes (calling `_syncThreejsLayer`) because if a user listen for the same event, you can't garantee the calling order.
So my proposal introduces a simple `onChange` callback to cover this. 
WDYT?

peppsac: Regarding event naming, I think :
```js
layer.addEventListener('visible-property-changed', (evt) => {
    console.log(`new value: ${evt.new}`);
}
```
is more readable / intuitive than:
```js
layer.addEventListener('layer-changed', (evt) => {
    if ('visible' in evt.new) {
        console.log(`new value ${evt.new.visible}`);
    }
}

File: src/Core/Prefab/GlobeView.js (hunk 1 above)
Comment (peppsac) [reply src/Core/Prefab/GlobeView.js]: But, what are they documenting? Layers?

File: src/Core/Prefab/GlobeView.js [hunk 12]
```diff
@@ -174,26 +231,158 @@ function GlobeView(viewerDiv, coordCarto, options) {
...
+            layerId,
+        });
+
+        return true;
+    } else {
+        throw new Error(`${layerId} isn't color layer`);
+    }
+};
+
+// QUESTION: new file for ordering functions?
```
peppsac: Yes (already wrote this in a previous comment) :)

peppsac (changes requested):

src/Core/Prefab/ColorLayersOrdering.js
```diff
@@ -0,0 +1,72 @@
+import { ImageryLayers } from '../Layer/Layer';
```
peppsac: I'd move this file either in `../Layer` or in `../../Rendering/` (because it's heavily related to `LayeredMaterial`)

src/Core/Layer/Layer.js
```diff
@@ -1,3 +1,25 @@
...
+
+
+export const defineLayerProperty = function defineLayerProperty(layer, propertyName, defaultValue, onChange) {
+    var property = layer[propertyName] == undefined ? defaultValue : layer[propertyName];
+    Object.defineProperty(layer,
+        propertyName,
+        { get: () => property,
+            set: (newValue) => {
+                if (property !== newValue) {
+                    const event = { type: `${propertyName}-layer-changed`, previous: {}, new: {} };
```
peppsac: `layer` is redundant in event name `${propertyName}-layer-changed` since the user will register on a layer anyway.

`${propertyName}-changed` or `${propertyName}-property-changed` seem better (I prefer the -property- version). WDYT?

src/Core/View.js
```diff
@@ -90,15 +105,21 @@ function _preprocessLayer(view, layer, provider) {
     // probably not the best place to do this
     if (layer.type == 'color') {
         layer.frozen = layer.frozen || false;
```
peppsac: Why don't you add `defineLayerProperty` for `frozen` and `sequence` as well ?

index.html
```diff
@@ -39,39 +39,29 @@
...
+            menuGlobe.view = globeView;
 
-            const apiGlobe = new itowns.ApiGlobe();
-            const menuGlobe = new GuiTools(apiGlobe, 'menuDiv');
-            apiGlobe.createSceneGlobe('globe', positionOnGlobe, viewerDiv);
-            // itowns.debug = new Debug(itowns.viewer.scene);
+            itowns.Fetcher.json('examples/layers/JSONLayers/Ortho.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('examples/layers/JSONLayers/OrthosCRS.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('examples/layers/JSONLayers/ScanEX.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('examples/layers/JSONLayers/Region.json').then(result => globeView.addLayer(result));
```
peppsac, line 49: Nice!

src/Core/Prefab/GlobeView.js
peppsac: Oh I get it. You're documenting `addLayer`.
Could you move this above `addLayer` please?

src/Main.js
```diff
@@ -1,21 +1,24 @@
 // Modules using `default` export must be imported...
 import * as THREE from 'three';
 import proj4 from 'proj4';
-import ApiGlobe, { INITIALIZED_EVENT } from './Core/Scheduler/Interfaces/ApiInterface/ApiGlobe';
 import { CONTROL_EVENTS } from './Renderer/ThreeExtended/GlobeControls';
 import View from './Core/View';
-import GlobeView from './Core/Prefab/GlobeView';
+import GlobeView, { GLOBE_VIEW_EVENTS } from './Core/Prefab/GlobeView';
+import { ColorLayersOrdering } from './Core/Prefab/ColorLayersOrdering';
```
peppsac: You can replace this line and the matching export below by:
`export { ColorLayersOrdering } from './Core/Prefab/ColorLayersOrdering'; `

peppsac (changes requested):
Looks mostly good, thanks :)

peppsac (approved):

peppsac:
Thanks for removing ApiGlobe :-)

PR #307, opened 2017-05-18
fix: disable horizon culling for level-1 tiles

peppsac:
The algorithm doesn't seem to work in this case and it
doesn't make sense to test them anyway.

### DIFF
--- a/src/Process/GlobeTileProcessing.js
+++ b/src/Process/GlobeTileProcessing.js
@@ -31,6 +31,7 @@ export function preGlobeUpdate(context) {
 }
 
 function pointHorizonCulling(pt) {
+    // see https://cesiumjs.org/2013/04/25/Horizon-culling/
     var vT = pt.divide(radius).sub(cV);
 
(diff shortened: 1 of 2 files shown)


src/Process/GlobeTileProcessing.js
```diff
@@ -64,7 +64,7 @@ function frustumCullingOBB(node, camera) {
 }
 
 export function globeCulling(node, camera) {
-    return !(frustumCullingOBB(node, camera) && horizonCulling(node));
+    return !(frustumCullingOBB(node, camera) && (node.level < 2 || horizonCulling(node)));
```
gchoqueux: Replace 2 by process's parameter.
`globeCulling` with `node.level < 2 ` can be important with a different `scheme tile`

gchoqueux (changes requested):

src/Process/GlobeTileProcessing.js
peppsac: Fixed.

gchoqueux (approved):

PR #309, opened 2017-05-19
Generate texture atlases at runtime

peppsac:
1st commit:
    Texture atlas usually perform better than individual textures.
    It also allows to reduce the number of Sampler2D to the number of layer,
    instead of the number of textures.
    
    I didn't merged all the layers in a single atlas, because updating this
    huge texture is slow (at least using the naive way of building a THREE.Texture
    from an atlas. It may be feasible to access directly WebGL functions like
    glTexSubImage2D).
    
    Currently the texture packer is naive and probably only works for our test
    data, but it should be easy to implement a more robust solution.
    
    Last but not least, without this commit iTowns crashes on my phone :)

~The 2nd commit leverages on the 1st and implements color textures transition.~ Edit: animation moved to a later PR

~Note that this PR is not quite ready for merging (layer deletion not implemented etc...) but you can test it and give feedback.~ Edit: ready!

Thanks!

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -366,21 +366,23 @@ GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
     // update the picking ray with the camera and mouse position
     const selectedId = this.screenCoordsToNodeId(mouse);
 
+    let selected;
     for (const n of this.wgs84TileLayer.level0Nodes) {
         n.traverse((node) => {
             // only take of selectable nodes
             if (node.setSelected) {
-                node.setSelected(node.id === selectedId);
+                // node.setSelected(node.id === selectedId);
 
                 if (node.id === selectedId) {
                     // eslint-disable-next-line no-console
                     console.info(node);
+                    selected = node;
                 }
             }
         });
     }
 
-    this.notifyChange(true);
+    return selected;
 };
 
 GlobeView.prototype.screenCoordsToNodeId = function screenCoordsToNodeId(mouse) {
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -375,27 +375,6 @@ GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
                 if (node.id === selectedId) {
                     // eslint-disable-next-line no-console
                     console.info(node);
-
-                    if (__DEBUG__) {
-                        const container = document.getElementById('canvasDebug');
-                        if (container) {
-                            while (container.firstChild) {
-                                container.removeChild(container.firstChild);
-                            }
-
-                            for (const layer of node.material.colorLayersId) {
-                                const index = node.material.indexOfColorLayer(layer);
-                                const texture = node.material.uniforms.atlasTextures.value[index];
-                                const canvas = texture.image;
-                                for (const txt of [`${layer}`, `${texture.uv.length} images`, `${canvas.width}x${canvas.height}`]) {
-                                    const s = document.createElement('div');
-                                    s.textContent = txt;
-                                    container.appendChild(s);
-                                }
-                                container.appendChild(canvas);
-                            }
-                        }
-                    }
                 }
             }
         });
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -375,6 +375,27 @@ GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
                 if (node.id === selectedId) {
                     // eslint-disable-next-line no-console
                     console.info(node);
+
+                    if (__DEBUG__) {
+                        const container = document.getElementById('canvasDebug');
+                        if (container) {
+                            while (container.firstChild) {
+                                container.removeChild(container.firstChild);
+                            }
+
+                            for (const layer of node.material.colorLayersId) {
+                                const index = node.material.indexOfColorLayer(layer);
+                                const texture = node.material.uniforms.atlasTextures.value[index];
+                                const canvas = texture.image;
+                                for (const txt of [`${layer}`, `${texture.uv.length} images`, `${canvas.width}x${canvas.height}`]) {
+                                    const s = document.createElement('div');
+                                    s.textContent = txt;
+                                    container.appendChild(s);
+                                }
+                                container.appendChild(canvas);
+                            }
+                        }
+                    }
                 }
             }
         });
--- a/src/Core/Scheduler/Providers/Fetcher.js
+++ b/src/Core/Scheduler/Providers/Fetcher.js
@@ -86,4 +86,17 @@ export default {
             return response.arrayBuffer();
         });
     },
+
+    img(url, options = {}) {
+        return new Promise((resolve, reject) => {
+            var image = new Image();
+
+            image.onload = () => resolve(image);
+
+            image.onerror = () => reject(new Error(`Error loading ${url}`));
+
+            image.crossOrigin = options.crossOrigin;
+            image.src = url;
+        });
+    },
 };
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -26,15 +26,13 @@ function initNodeImageryTexturesFromParent(node, parent, layer) {
 
                 if (c.isInside(parentCoords)) {
                     const result = c.offsetToParent(parentCoords);
-
                     const p = atlas.uv[i];
-
                     const idx = coords.indexOf(c);
                     offsetScale[idx] = p.clone();
                     offsetScale[idx].x += result.x * p.z;
                     offsetScale[idx].y += result.y * p.w;
                     offsetScale[idx].z *= result.z;
-                    offsetScale[idx].w *= result.z;
+                    offsetScale[idx].w *= result.w;
 
                     found++;
                     break;
@@ -164,7 +162,6 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     }
 
     const material = node.material;
-    let index = material.indexOfColorLayer(layer.id);
 
     // Initialisation
     if (node.layerUpdateState[layer.id] === undefined) {
@@ -200,11 +197,12 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
             };
 
             material.pushLayer(paramMaterial);
+
+            initNodeImageryTexturesFromParent(node, node.parent, layer);
+
             const imageryLayers = context.view.getLayers(l => l.type === 'color');
             const sequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
             material.setSequence(sequence);
-
-            initNodeImageryTexturesFromParent(node, node.parent, layer);
         }
     }
 
@@ -217,6 +215,9 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     // to avoid mixing layer's network updates and layer's params
     // Update material parameters
     const layerIndex = material.indexOfColorLayer(layer.id);
+    if (layerIndex < 0) {
+        return;
+    }
     material.setLayerVisibility(layerIndex, layer.visible);
     material.setLayerOpacity(layerIndex, layer.opacity);
 
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -1,4 +1,5 @@
 import { l_ELEVATION, l_COLOR, EMPTY_TEXTURE_ZOOM } from '../Renderer/LayeredMaterialConstants';
+import { acquireTexture, releaseTexture } from '../Renderer/LayeredMaterial';
 import { chooseNextLevelToFetch } from '../Core/Layer/LayerUpdateStrategy';
 import LayerUpdateState from '../Core/Layer/LayerUpdateState';
 import { ImageryLayers } from '../Core/Layer/Layer';
@@ -11,30 +12,46 @@ const MAX_RETRY = 4;
 function initNodeImageryTexturesFromParent(node, parent, layer) {
     if (parent.material && parent.material.getColorLayerLevelById(layer.id) > EMPTY_TEXTURE_ZOOM) {
         const coords = node.getCoordsForLayer(layer);
-        const offsetTextures = node.material.getLayerTextureOffset(layer.id);
 
-        let textureIndex = offsetTextures;
+        const indexInNode = node.material.indexOfColorLayer(layer.id);
+        const indexInParent = parent.material.indexOfColorLayer(layer.id);
+
+        const offsetScale = node.material.offsetScale[l_COLOR][layer.id];
+
+        let found = 0;
+        const atlas = parent.material.uniforms.atlasTextures.value[indexInParent];
         for (const c of coords) {
-            for (const texture of parent.material.getLayerTextures(l_COLOR, layer.id)) {
-                if (c.isInside(texture.coords)) {
-                    const result = c.offsetToParent(texture.coords);
-                    node.material.textures[l_COLOR][textureIndex] = texture;
-                    node.material.offsetScale[l_COLOR][textureIndex] = result;
-                    textureIndex++;
+            for (let i = 0; i < atlas.coords.length; i++) {
+                const parentCoords = atlas.coords[i];
+
+                if (c.isInside(parentCoords)) {
+                    const result = c.offsetToParent(parentCoords);
+
+                    const p = atlas.uv[i];
+
+                    const idx = coords.indexOf(c);
+                    offsetScale[idx] = p.clone();
+                    offsetScale[idx].x += result.x * p.z;
+                    offsetScale[idx].y += result.y * p.w;
+                    offsetScale[idx].z *= result.z;
+                    offsetScale[idx].w *= result.z;
+
+                    found++;
                     break;
                 }
             }
         }
 
         if (__DEBUG__) {
-            if ((textureIndex - offsetTextures) != coords.length) {
+            if (found != coords.length) {
                 /* eslint-disable */
-                console.error(`non-coherent result ${textureIndex} ${offsetTextures} vs ${coords.length}. ${coords}`);
+                console.error(`non-coherent result ${textureIndex} ${found} vs ${coords.length}. ${coords}`);
                 /* eslint-enable */
             }
         }
-        const index = node.material.indexOfColorLayer(layer.id);
-        node.material.layerTexturesCount[index] = coords.length;
+
+        releaseTexture(node.material.uniforms.atlasTextures.value[indexInNode]);
+        node.material.uniforms.atlasTextures.value[indexInNode] = acquireTexture(atlas);
         node.material.loadedTexturesCount[l_COLOR] += coords.length;
     }
 }
@@ -58,7 +75,7 @@ function initNodeElevationTextureFromParent(node, parent, layer) {
         // extract min-max from the texture (too few information), we instead chose
         // to use parent's min-max.
         const useMinMaxFromParent = node.level - texture.coords.zoom > 6;
-        if (!useMinMaxFromParent) {
+        if (!useMinMaxFromParent && layer.options.mimetype.indexOf('x-bil') > 0) {
             const { min, max } = OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
                 texture.image.data,
                 SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
@@ -147,6 +164,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     }
 
     const material = node.material;
+    let index = material.indexOfColorLayer(layer.id);
 
     // Initialisation
     if (node.layerUpdateState[layer.id] === undefined) {
@@ -243,6 +261,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
         priority: nodeCommandQueuePriorityFunction(node),
         earlyDropFunction: refinementCommandCancellationFn,
         targetLevel,
+        rawImage: true, // color textures use atlas so we request an img, not a texture
     };
 
     return context.scheduler.execute(command).then(
@@ -252,9 +271,9 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
             }
 
             if (Array.isArray(result)) {
-                node.setTexturesLayer(result, l_COLOR, layer.id);
+                node.setTexturesLayer(result, l_COLOR, layer);
             } else if (result.texture) {
-                node.setTexturesLayer([result], l_COLOR, layer.id);
+                node.setTexturesLayer([result], l_COLOR, layer);
             } else {
                 // TODO: null texture is probably an error
                 // Maybe add an error counter for the node/layer,
--- a/src/Renderer/AtlasBuilder.js
+++ b/src/Renderer/AtlasBuilder.js
@@ -2,6 +2,16 @@ import * as THREE from 'three';
 import Capabilities from '../Core/System/Capabilities';
 import fit from './Packer';
 
+const availableCanvas = [];
+
+function getCanvas() {
+    if (availableCanvas.length) {
+        return availableCanvas.pop();
+    }
+    const canvas = document.createElement('canvas');
+    return canvas;
+}
+
 /**
  * Build a texture atlas from N images.
  *
@@ -18,7 +28,8 @@ import fit from './Packer';
  */
 export default function pack(images, uvs, needsPixelSeparation) {
     // pick an available canvas, or build a new one
-    const atlasCanvas = document.createElement('canvas');
+    const atlasCanvas = getCanvas();
+
     // Use a 1 pixel border to avoid color bleed when sampling at the edges
     // of the texture
     const colorBleedHalfOffset = (!needsPixelSeparation || images.length == 1) ? 0 : 1;
@@ -129,6 +140,11 @@ export default function pack(images, uvs, needsPixelSeparation) {
     atlas.minFilter = THREE.LinearFilter;
     atlas.anisotropy = 1;
     atlas.uv = uv;
+
+    atlas.onUpdate = () => {
+        availableCanvas.push(atlasCanvas);
+        atlas.onUpdate = undefined;
+    };
 
     return atlas;
 }
--- a/src/Renderer/AtlasBuilder.js
+++ b/src/Renderer/AtlasBuilder.js
@@ -1,0 +1,134 @@
+import * as THREE from 'three';
+import Capabilities from '../Core/System/Capabilities';
+import fit from './Packer';
+
+/**
+ * Build a texture atlas from N images.
+ *
+ * We use a classic 2D Bin Packing algorithm to assign each individual image a
+ * location in the resulting texture.
+ * Then this texture is created using a <canvas>,  onto which we draw all images.
+ * In the end we return a THREE.CanvasTexture and an array 'uv' of Vector4, describing
+ * the position/size of each input images in the atlas.
+ * @param {array} images - an array of <img>
+ * @param {array} uvs - an array of coordinates indicating what part of the image we should keep
+ * @param {boolean} needsPixelSeparation - does this atlas need to use a anti color bleed pixel
+ * between images
+ * @return {THREE.CanvasTexture}
+ */
+export default function pack(images, uvs, needsPixelSeparation) {
+    // pick an available canvas, or build a new one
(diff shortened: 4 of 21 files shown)


elemoine:
FWIW OpenLayers uses the bin packing algorithm described in http://www.blackpawn.com/texts/lightmaps/. See https://github.com/openlayers/openlayers/blob/master/src/ol/style/atlas.js for the OpenLayers implementation.

peppsac:
@elemoine yep, this algorithm seems to be widly used. I was just too lazy to do it so I adapted a simple existing code. 
Note that we're not too constrained anyway: our textures are all power of 2 (mostly 256x256 pixels) but as we need to add a border around them to prevent color bleeding, we end up with 258x258 pixels.
And as the end result must be a power of 2 texture as well, we often waste a lot of space (e.g: 2 256x256 textures need to be stored in a 1024x512 atlas)

qdnguyen:
>Texture atlas usually perform better than individual textures.
It also allows to reduce the number of Sampler2D to the number of layer,
instead of the number of textures.

Other solution is to use mega texture (virtual texture ) and update it through glTexSubImage2D, but i dont know how to support all LoDs of wmts with mega texture.

peppsac:
@qdnguyen yeah, mega texture is an option, but I'm not familiar with them so I'll stick with atlases :)

peppsac:
OK, this PR is ready now :)

peppsac:
Updated and added a debug tool to see the atlas from the selected tile.

gchoqueux:
little problem with OBB 
capture du 2017-06-09 15-40-51

peppsac:
Updated: fixed the OBB bug and an issue with `img` and `THREE.Texture`

gchoqueux:
I do a performance test :
`view.controls.setRange(50000,true)`
time on master : 1.5s and Atlas 2.5s

peppsac:
> time on master : 1.5s and Atlas 2.5s

What are you measuring exactly? Here's what I do:
```
console.log('>>>'); globeView.controls.setRange(50000,true).then(()=>{console.log('<<<');})
```
My branch is a bit slower (4.5sec vs 4 sec) but it's expected since I do a bit more work than the default `three.js` implementation.

peppsac:
Also in the general case the bottleneck will probably be the network accesses.

peppsac:
Rebased and ready to review.

src/Core/Scheduler/Providers/Fetcher.js
```diff
@@ -40,6 +40,19 @@ export default {
...
 
+    img(url) {
+        return new Promise((resolve, reject) => {
+            var image = new Image();
+
+            image.onload = () => resolve(image);
+
+            image.onerror = () => reject(new Error(`Error loading ${url}`));
+
+            image.crossOrigin = '';
```
autra: Not requiring it here, but we should probably make this configurable in the future (as it depends on server config).

utils/debug/Debug.js
```diff
@@ -261,6 +262,21 @@ function Debug(view, viewerDiv) {
         }
     });
 
+    const canvasDiv = document.createElement('div');
+    canvasDiv.id = 'canvasDebug';
+    canvasDiv.style = 'position: absolute; top: 0; right: 0; width: 25vw; background-color: white; display: none; border-width: 0 0 3px 3px; border-style:solid; border-color: black;';
```
autra: + something like `height: 100%; overflow: auto` to make it scrollable if it overflows?

(I would prefer `width: 25%` but it depends what you want to do)

autra: Nice tool btw!

test/itowns-testing.js
```diff
@@ -40,7 +40,17 @@ global.URL = function URL(str) {
     return url.parse(str);
 };
 global.document = new function _d() {
```
autra, line 46: I'm really starting to think we need something like [mock-browser](https://www.npmjs.com/package/mock-browser). It could be a good middleground between this and a headless browser wdyt? If you agree, I'll make a PR with this.

src/Renderer/Packer.js
```diff
@@ -0,0 +1,45 @@
+// original code from https://github.com/jakesgordon/bin-packing
+// MIT License
+
+import { Math as _Math } from 'three';
```
autra, line 4: Curious why you renamed `Math`?

src/Renderer/Packer.js
```diff
@@ -0,0 +1,45 @@
...
+
+function fit(blocks, w, h) {
```
autra: Could you document this method plz?

src/Renderer/LayeredMaterial.js
```diff
@@ -382,69 +304,55 @@ LayeredMaterial.prototype.setLayerFx = function setLayerFx(index, fx) {
...
-    if (layerType === l_ELEVATION) {
-        return this.textures[l_ELEVATION];
+function _updateAtlas(textures, offsetScale, keepAlphaChannel) {
+    const images = [];
+    const uvs = [];
+
+    for (let i = 0; i < textures.length; i++) {
+        const img = textures[i];
+        images.push(img);
+        uvs.push(offsetScale[i]);
```
autra: Did I miss something or you are just doing a shallow copy of textures and offsetScale here?

autra: Did I miss something or you are just doing a shallow copy of textures and offsetScale here?

src/Renderer/LayeredMaterial.js
```diff
@@ -382,69 +304,55 @@ LayeredMaterial.prototype.setLayerFx = function setLayerFx(index, fx) {
...
-        return this.textures[l_COLOR].slice(textureIndex, textureIndex + count);
-    } else {
-        throw new Error(`Invalid layer id "${layerId}"`);
+    if (!keepAlphaChannel) {
+        atlas.format = THREE.RGBFormat;
     }
-};
+
+    atlas.uv = uv;
+    return { atlas, uv };
```
autra: I would say one of these 2 lines is redundant.

src/Renderer/LayeredMaterial.js
```diff
@@ -382,69 +304,55 @@ LayeredMaterial.prototype.setLayerFx = function setLayerFx(index, fx) {
...
 LayeredMaterial.prototype.getColorLayerLevelById = function getColorLayerLevelById(colorLayerId) {
     const index = this.indexOfColorLayer(colorLayerId);
     if (index === -1) {
         return EMPTY_TEXTURE_ZOOM;
     }
-    const slot = this.getTextureOffsetByLayerIndex(index);
-    const texture = this.textures[l_COLOR][slot];
-
-    return texture ? texture.coords.zoom : EMPTY_TEXTURE_ZOOM;
+    return this.uniforms.atlasTextures.value[index].coords[0].zoom;
```
autra, line 415: What is this 0?

src/Renderer/LayeredMaterial.js
```diff
@@ -139,218 +82,210 @@ const LayeredMaterial = function LayeredMaterial(options) {
...
-                        total++;
-                    }
-                }
-                if (total != count) {
-                    throw new Error(`${node.id} - Invalid total texture count. Found: ${total}, expected: ${count} for ${layer.id}`);
-                }
-            }
-        };
-    }
+    const atlasTextures = Array(8);
```
autra: nits `new Array(8)` ?

autra: Doesn't change anything here especially, but it's a matter of good habits for the most... dangerous part of js language :-) It works only because the writer of `Array` has coded for this use. In general, I'd like to see the `constructor starts with uppercase and should be called with new, normal function starts with lowercase and are called normally` rule enforced in itowns.

src/Renderer/LayeredMaterial.js
```diff
@@ -139,218 +82,210 @@ const LayeredMaterial = function LayeredMaterial(options) {
...
     }
 
+    // build new arrays
+    _swapIndices(this.colorLayersId, newIndices);
+    _swapIndices(this.uniforms.paramLayers.value, newIndices);
+    _swapIndices(this.uniforms.atlasTextures.value, newIndices);
+
     this.uniforms.colorLayersCount.value = this.getColorLayersCount();
+
+    if (updateFragmentShader || updateFragmentShader == undefined) {
```
autra: Looks like you want to use a [default parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) value for `updateFragmentShader` instead.

src/Core/Prefab/GlobeView.js
```diff
@@ -327,6 +327,27 @@ GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
...
+                                const canvas = texture.image;
+                                for (const txt of [`${layer}`, `${texture.uv.length} images`, `${canvas.width}x${canvas.height}`]) {
+                                    const s = document.createElement('div');
+                                    s.textContent = txt;
+                                    container.appendChild(s);
+                                }
+                                container.appendChild(canvas);
+                            }
+                        }
+                    }
```
autra: That's cool!

src/Renderer/AtlasBuilder.js
```diff
@@ -0,0 +1,94 @@
+import * as THREE from 'three';
+import Capabilities from '../Core/System/Capabilities';
+import fit from './Packer';
+
+export default function pack(images, uvs) {
```
autra: Can you document this function plz? (I struggle to understand it without it, and some comments below, maybe).

autra: My understanding is that you pack all the `images` into one big texture (with the help of the canvas thing below) and that uvs contains the individual coordinates. Is that right?

src/Renderer/AtlasBuilder.js
```diff
@@ -0,0 +1,94 @@
...
+
+        blocks.push({
+            index: i,
+            w: sWidth + 2 * colorBleedHalfOffset,
+            h: sHeight,
+            empty: replaceWithEmpty,
+        });
+    }
+
+    blocks.sort((a, b) => Math.max(a.w, a.h) < Math.max(b.w, b.h));
```
autra: Why do you need to sort? (I feel like I don't really understand your data layout)

src/Renderer/LayeredMaterial.js
```diff
@@ -139,218 +82,210 @@ const LayeredMaterial = function LayeredMaterial(options) {
...
-    // refill remove textures
-    for (let i = 0, max = texturesCount; i < max; i++) {
-        this.textures[l_COLOR].push(emptyTexture);
-        this.offsetScale[l_COLOR].push(vector);
-    }
+    // for (let i = offset, max = offset + texturesCount; i < max; i++) {
+    //     if (this.textures[l_COLOR][i] instanceof THREE.Texture) {
+    //         this.textures[l_COLOR][i].dispose();
+    //     }
+    // }
```
autra: I guess we can remove these?

autra (changes requested):

peppsac:
Thanks for the review @autra
I won't reply to your comments here but instead I'll amend the code to fix all the issues you noted.

test/itowns-testing.js
peppsac: Worth a try at least :)

src/Renderer/LayeredMaterial.js
peppsac: Because all coords of one atlas will have the same level/zoom anyway

examples/planar.html
```diff
@@ -57,7 +57,7 @@
         </script>
         <script src="planar.js"></script>
         <script type="text/javascript">
-            var animationEnabled = true;
+            var animationEnabled = false;
```
autra: Do we need the diffs on this file in this PR?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -9,30 +9,48 @@ import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from '../Core/Scheduler/Provid
 function initNodeImageryTexturesFromParent(node, parent, layer) {
     if (parent.materials && parent.materials[RendererConstant.FINAL].getColorLayerLevelById(layer.id) > EMPTY_TEXTURE_ZOOM) {
         const coords = node.getCoordsForLayer(layer);
```
autra: If i'm not mistaken, this function always returns an Array of size 1, or did I miss something?

autra: It seems it is the case. Maybe we can then avoid the `for...off` below?

autra: Ok I forgot about PM. I get it.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -160,9 +179,15 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
         const imageryLayers = context.view.getLayers(l => l.type === 'color');
         const sequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
-        material.setSequence(sequence);
+        // do not update fragment shader yet
+        material.setSequence(sequence, false);
 
         initNodeImageryTexturesFromParent(node, node.parent, layer);
+
+        // do it now - after we set the atlas textures
+        material._updateFragmentShader();
```
autra: Then `updateFragmentShader` should not be considered as private, and be renamed accordingly.

autra: Can we indicate to initNodeImageryTexturesFromParent the correct sequence from the start? It would save the trouble of calling setSequence here, and would remove the need of the updateFragment flag in setSequence.

src/Renderer/AtlasBuilder.js
autra: I meant a block comment on top explaining the purpose of this method, how it achieves its goal and what it returns. I'm not even requiring you use jsdoc tags, as I know you like them so much (but do use `/**` to make it appear in the generated doc though plz).

src/Renderer/AtlasBuilder.js
```diff
@@ -0,0 +1,117 @@
...
+        const x = block.fit.x + colorBleedHalfOffset;
+        const y = block.fit.y;
+
+        if (!block.empty) {
+            src.x *= img.width;
+            src.y *= img.height;
+            src.width *= img.width;
+            src.height *= img.height;
+
+            // draw the whole image image
```
autra: nits: one `image` is enough.

src/Renderer/LayeredMaterial.js
```diff
@@ -139,218 +82,209 @@ const LayeredMaterial = function LayeredMaterial(options) {
...
-        for (let l = 0; l < params.length; l++) {
-            this.pushLayer(params[l]);
-        }
+LayeredMaterial.prototype.setElevationTexture = function setElevationTexture(texture, offsetScale) {
+    if (this.textures[l_ELEVATION][0] === undefined || this.textures[l_ELEVATION][0].image === undefined) {
+        this.loadedTexturesCount[l_ELEVATION] += 1;
     }
+
+    texture._ownerCount = (texture._ownerCount || 0) + 1;
+    // BEWARE: array [] -> size: 0; array [10]="wao" -> size: 11
```
autra: I don't really see the point of this comment here? But anyway, maybe we want a map instead then.

src/Renderer/LayeredMaterial.js
```diff
@@ -139,218 +82,209 @@ const LayeredMaterial = function LayeredMaterial(options) {
...
+    this.offsetScale[l_ELEVATION][0] = offsetScale || new THREE.Vector3(0.0, 0.0, 1.0);
 };
 
 LayeredMaterial.prototype.pushLayer = function pushLayer(param) {
     const newIndex = this.getColorLayersCount();
-    const offset = newIndex === 0 ? 0 : this.getTextureOffsetByLayerIndex(newIndex - 1) + this.getTextureCountByLayerIndex(newIndex - 1);
 
-    this.uniforms.paramLayers.value[newIndex] = new THREE.Vector4();
+    const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
+    const nbSamplers = Math.min(maxTexturesUnits - 1, 16 - 1) - 1;
```
autra, line 355: Why can't we create more than 14 samplers if `maxTexturesUnits` is bigger than 16?

src/Renderer/Packer.js
autra: Ok Forgot about `Math` in js

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -9,30 +9,48 @@ import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from '../Core/Scheduler/Provid
...
-            for (const texture of parent.materials[RendererConstant.FINAL].getLayerTextures(l_COLOR, layer.id)) {
-                if (c.isInside(texture.coords)) {
-                    const result = c.offsetToParent(texture.coords);
-                    node.material.textures[l_COLOR][textureIndex] = texture;
-                    node.material.offsetScale[l_COLOR][textureIndex] = result;
-                    textureIndex++;
+            for (let i = 0; i < atlas.coords.length; i++) {
+                const parentCoords = atlas.coords[i];
+
+                if (c.isInside(parentCoords)) {
```
autra, line 27: So this is true if c is strictly inside parentCoords, right? But with PM it's possible that c overlaps with one of the parent texture without being entirely in it right? We simply ignore this case?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -53,7 +71,7 @@ function initNodeElevationTextureFromParent(node, parent, layer) {
         // extract min-max from the texture (too few information), we instead chose
         // to use parent's min-max.
         const useMinMaxFromParent = node.level - texture.coords.zoom > 6;
-        if (!useMinMaxFromParent) {
+        if (!useMinMaxFromParent && layer.options.mimetype.indexOf('x-bil') > 0) {
```
autra, line 76: What was the problem?

src/Renderer/LayeredMaterial.js
autra: (Not requiring it here though)

autra (changes requested):
A few comments, many questions :-) Overall this looks a lot simpler than previous code.

Thanks!

src/Renderer/LayeredMaterial.js
peppsac: Sorry I forgot to fix this one.
Is this rule available in eslint?

src/Process/LayeredMaterialNodeProcessing.js
peppsac: (you're commenting on existing code that I simply updated)

And I *think* a PM texture will always be fully contained in its parent textures which are PM as well (would need to double-check this though)

src/Renderer/LayeredMaterial.js
autra: it's http://eslint.org/docs/rules/new-cap, currently disabled in our conf (but apparently they have exceptions for `Array`, so let's forget about this one)

autra: And it's currently disabled because we don't follow this rule at all :-/

src/Process/LayeredMaterialNodeProcessing.js
peppsac: IMHO it's private to itowns, not to `LayeredMaterial`, hence the `_` prefix.

But I can remove it if you prefer.

examples/planar.html
peppsac: No sorry, I'll remove this.

src/Process/LayeredMaterialNodeProcessing.js
peppsac: The code under the `if` assumes a xbil texture:
```
const { min, max } = OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
```

src/Renderer/LayeredMaterial.js
peppsac: (again, this is existing code)

I'll remove it.

src/Renderer/LayeredMaterial.js
peppsac: (this code was there before as well)

I think the reasonning is that sometimes `Capabilities.getMaxTextureUnitsCount();` reported incorrect values, so the `16` was a reasonnable maximum.

This could be investigated and maybe removed...

src/Renderer/AtlasBuilder.js
peppsac: You're not requiring them, but our linter does.

src/Process/LayeredMaterialNodeProcessing.js
peppsac: Thanks for the suggestion: I fixed the code using your idea.

src/Renderer/AtlasBuilder.js
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * We use a classic 2D Bin Packing algorithm to assign each individual image a
+ * location in the resulting texture.
+ * Then this texture is created using a <canvas>,  onto which we draw all images.
+ * In the end we return a THREE.CanvasTexture and an array 'uv' of Vector4, describing
+ * the position/size of each input images in the atlas.
+ * @param {array} images - an array of <img>
+ * @param {array} uvs - an array of coordinates indicating what part of the image we should keep
+ * @return {THREE.CanvasTexture}
+ */
```
autra: Thanks!

autra (dismissed):

autra:
LGTM, thanks!

nosy-b:
Ok I check to see performance issues. It should be slower with texture atlas as it implies more JS slow canvas work. I tell you before 5:30 pm!

nosy-b:
Ok, I didn't see a significant speed difference which is cool (although I didn't measure it precisely) but the memory footprint is much more, going above 1.2 Go after 45 sec of active navigation vs 0.5 Go with the master.
I would suggest to work on that after the release.

peppsac:
@nosy-b :

> the memory footprint is much more, going above 1.2 Go after 45 sec of active navigation vs 0.5 Go with the master.

That's interesting. How did you measure this? (using dev tools memory tab or something else ?)

nosy-b:
With the OS task manager (config win + chrome). It looks like creating texture combining others increases the memory footprint constantly, maybe you keep references to original tiles and new combining textures

autra:
> With the OS task manager

Not sure it's a reliable way of measuring memory footprint. Navigators keeps some chunks of memory allocated from time to time for future use, or maybe it didn't GC immediately because your system still had some memory left etc. I suggest to use firefox  or chrome memory tab in devtools. In firefox, you can force a GC, CC and reclaim as much memory as it is possible in the `about:memory` page. I think it will yield more accurate results.

peppsac:
So, using the task manager of Chrome (see https://developers.google.com/web/tools/chrome-devtools/memory-problems/) I could indeed confirm that the total memory used by my iTowns tab was growing.

I found and fixed 2 different bugs related to memory management, but the issue was still there.

And then I found these 2 bugs: https://bugs.chromium.org/p/chromium/issues/detail?id=242215 and https://bugs.chromium.org/p/chromium/issues/detail?id=626082 which seem related to our issue.

The second bug mentions that the issue has been (partially?) fixed 2 weeks ago :)

Testing with Chrome 61 confirmed that the problem seems fixed.

I'll try to see if I can mitigate the issue (I can maybe reused old canvas instead of creating new ones)...

peppsac:
I just pushed 2 different approaches to limit the number of canvas used and thus the memleak.

autra:
@peppsac which approach gave you the best results? Maybe we can add a // XXX or // FIXME comment explaining why this change, and linking to the bug reports ? I'll see if I reproduce in firefox.

peppsac:
The second approach creates less `canvas` but is more fragile since it depends on 2 `three.js` behavior:
- a texture is only uploaded once (which is true if its .version field never change, which is true in our case)
- an uploaded texture has a `dispose` listener

I don't mind depending on three.js behavior, but we should probably introduce a comment/tag (`// three.js version: r0.86`?) everytime we do this to make sure this is well tested when we upgrade three.js.

autra:
I've measuring memory on Firefox, and couldn't reproduce. While memory usage with this PR is consistently slightly higher than on master (around 10%), I saw no leaks.

How I tested:
- I loaded Itowns
- Waited for globe to be loaded, then took a memory snapshot
- zoomed on Paris, waited for globe to be loaded, then took another snapshot.

On master, I'm around 195-200MB. On this branch, around 215-220MB. @peppsac is this slight increase expected?

Anyhow, that confirms the chromium bug I guess. @peppsac I let you choose the approach you prefer to limit memory usage, but I don't think we should block this PR because of this bug. Maybe we can add some comments, maybe also a warning in the console with a link to the bug report for chromium developers?

src/Core/Prefab/GlobeView.js
peppsac: And broken with the last patch, because I recycle the canvas :)

autra: Damn'!

autra (approved):

gchoqueux (changes requested):
Wait before merge, please
I would like reviewing this PR

gchoqueux:
there's a problem with video memory
This PR uses more video memory that master branch

statut |  Gpu system bytes | Gpu dedicated bytes | Gpu committed bytes
------------ | ------------ | ------------- | -------------
 Master init |  21M  |  157M  | 177M
 Master range 500000 |  19M  | 175M  | 191M
Atlas init  |  26M |  195M | 219M
 Atlas range 500000 |  19M  |  260 M |  276M

May be textures atlas isn't completely filled.

Capture textures in video memory, There are empty areas
capture
capture

peppsac:
> Capture textures in video memory, There are empty areas

Textures must use power of 2 sizes, so these empty areas are expected.

>  This PR uses more video memory that master branch

Video memory usage is not an end goal in itself.
That being said, yes this PR increases memory usage, but on the other hand it simplifies shader code and texture management.
It also opens the way to implement feature like animated transition.

We can refine `AtlasBuilder` to improve memory usage but I prefered to first land a naive version that performs OK and then iterate from there.

gchoqueux:
Unfortunate this isn't a small increase in memory. 
Which can be fixed by filling in the empty slots in atlas

peppsac:
> Unfortunate this isn't a small increase in memory.

So what?
It's a compromise: increased mem usage vs shader cost and other params (I forgot texture sampler count in my previous comment). 

> Which can be fixed by filling in the empty slots in atlas

I won't fill empty slots in atlas, since it's not possible.
But we can reduce GPU-object count by merging multiple atlases into one for instance.
We can also investigate the need for the color-bleed-preventing pixel for PM texture... hmm maybe I should stack them vertically since they are consecutives images... I'll give it a try.

peppsac:
Pushed an update that should bring back memory usage to something closer to master.

Changes implemented:
- if a layer has a single image -> then skip the atlas mechanism and directly return a `THREE.Texture` as `TextureLoader` used to do for us
- if a layer has several textures: currently it only happens for PM layers so I added a preparation step to avoid the need to add a pixel line between individual images. So now a PM layer has this kind of atlas:
capture d ecran de 2017-07-19 15-19-44

@gchoqueux WDYT?

gchoqueux:
I'll review, don't merge before my review, please

gchoqueux:
I look at it this morning

gchoqueux:
When you do serveral back and forth, there is a memory leak in the Vram. (Chrome windows v59)

`globeView.controls.setRange(500000,true)` and `globeView.controls.setRange(25100000,true) `

The dispose texture is more complicated?

gchoqueux:
there's VRAM leak on  Chrome windows v61

peppsac:
Updated to rebase on master and added a fix for `getXbilTextureByUrl`

Regarding the leak: IMHO as long as the textures are all released (= three.js texture count doesn't keep increasing) I don't see how the bug could be in itowns :-/

peppsac:
(moved the `fix` commit to #401 )

peppsac:
Branch updated.
I re-enabled a temporary debug tool to see which texture are used for each tile (only in index.html demo).

There's more work to do, but are we interested in getting this merged or not?
It should help with #654  and #583 but I'd like some feedback.

autra:
I confirm that it helps a lot for #654. Planar examples works perfectly on a sony xperia tablet. Globe nearly works: I needed to use OPENSM.json layer: when near the ground, it works perfectly, when far away, texture are not displaying (but that might be because of the layer, I don't know).

Orthographic doesn't work anymore on both desktop and tablet with this PR though...

tested with both firefox and chrome

At least the Firefox doesn't crash any more.

peppsac:
Self-closing this one. I'll reopen it later, when I have some time to work on it.

Issue #181, opened 2016-09-01
Vector data clamping to ground


vpicavet:
Right now vector data are displayed, but not clamped to the ground.
We will need to be able to clamp the data for 2D vector data visualization : 
- [x] Points
- [x] Lines
- [x] Polygons

@vmora already implemented a clamping algorithm previously, he may be able to help.

Clamping should be usable by all loaders for vector data, like WFS ( #115 ), GPX ( #114 #158 ), KML ( #24 ), GeoJSON, SHP and vector tiles.

gchoqueux:
Some functions to add
- [ ] map.getCurrentAltitude(longitude, latitude)
  This function returns altitude of the current MNT displayed
- [ ] process.PlaceMeshOnMapSurface(node,map) 
  This process place the node/object3d on 3d surface of map

gchoqueux:
This features are in itowns

PR #312, opened 2017-05-23
Clamp to ground

peppsac:
feature: add placeObjectOnGround module
    
    Allows to place an object on the ground.
    Works by reading the elevation texture used for display.
    
    Support both float texture (DataTexture) and color texture.

### DIFF
--- a/examples/clamptoground.html
+++ b/examples/clamptoground.html
@@ -1,0 +1,91 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Clamp to Ground demo</title>
+
+        <style type="text/css">
+            html {height: 100%}
+            body { margin: 0; overflow:hidden; height:100%}
+
+            #viewerDiv {
+                margin : auto auto;
+                width: 100%;
+                height: 100%;
+                padding: 0;
+            }
+        </style>
+        <meta charset="UTF-8">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <script src="/dist/itowns.js"></script>
+        <script type="text/javascript">
+            const positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };
+
+            // iTowns namespace defined here
+            const viewerDiv = document.getElementById('viewerDiv');
+            const globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+
+            itowns.Fetcher.json('/examples/layers/JSONLayers/Ortho.json').then(result => globeView.addLayer(result));
+
+            itowns.Fetcher.json('/examples/layers/JSONLayers/IGN_MNT.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('/examples/layers/JSONLayers/IGN_MNT_HIGHRES.json').then(result => globeView.addLayer(result));
+
+
+            const clamped = new itowns.THREE.Group();
+
+            const l = new itowns.GeometryLayer('outline');
+            l.type = 'geometry';
+            l.preUpdate = (context, layer, sources) => {
+                return clamped.children || [];
+            };
+            l.update = (context, layer, obj) => {
+                if (obj.geometry instanceof itowns.THREE.ConeGeometry) {
+                    itowns.placeObjectOnGround(globeView.referenceCrs, globeView.wgs84TileLayer, globeView.wgs84TileLayer.level0Nodes, obj, {
+                        offset: 250,
+                    });
+                } else {
+                    itowns.placeObjectOnGround(globeView.referenceCrs, globeView.wgs84TileLayer, globeView.wgs84TileLayer.level0Nodes, obj, {
+                        modifyGeometry: true,
+                        offset: 10,
+                    });
+                }
+                // update RTC matrix
+                obj.material.setMatrixRTC(
+                        context.engine.getRTCMatrixFromNode(obj, context.camera));
+            };
+            itowns.View.prototype.addLayer.call(globeView, l);
+
+            // setup click listener
+            viewerDiv.addEventListener('contextmenu', (evt) => {
+                const carto = new itowns.Coordinates('EPSG:4978', globeView.getPickingPositionFromDepth({ x: evt.clientX, y: evt.clientY }));
+
+                var geometry = new itowns.THREE.ConeGeometry(100, 500);
+                var material = new itowns.BasicMaterial(new itowns.THREE.Color(0xff0000));
+                material.side = itowns.THREE.DoubleSide;
+                var cone = new itowns.THREE.Mesh(geometry, material);
+                cone.frustumCulled = false;
+                cone.updateMatrix();
+                cone.layers.set(l.threejsLayer);
+
+                const p = carto.xyz();
+                cone.lookAt(p);
+                cone.rotateX(-Math.PI * 0.5);
+                cone.position.copy(p);
+
+                cone.updateMatrix();
+                cone.updateMatrixWorld(true);
+
+                clamped.add(cone);
+
+                globeView.notifyChange(true);
+                return false;
+            });
+
+            globeView.scene.add(clamped);
+
+        </script>
+    </body>
+</html>
--- a/src/Process/PlaceOnGroundProcessing.js
+++ b/src/Process/PlaceOnGroundProcessing.js
@@ -1,0 +1,131 @@
+import * as THREE from 'three';
+import Coordinates from '../Core/Geographic/Coordinates';
+import Extent from '../Core/Geographic/Extent';
+import RendererConstant from '../Renderer/RendererConstant';
+import { l_ELEVATION } from '../Renderer/LayeredMaterial';
+
+function tileAt(pt, tile) {
+    if (tile.extent) {
+        if (!tile.extent.pointIsInside(pt)) {
+            return undefined;
+        }
+
+        for (const c of tile.children) {
+            const t = tileAt(pt, c);
+            if (t) {
+                return t;
+            }
+        }
+        if (tile.getLayerTextures(l_ELEVATION)[0].coords.zoom > -1) {
+            return tile;
+        }
+        return undefined;
+    }
+}
+
+let _canvas;
+function _readTextureValueAt(layer, texture, x, y) {
+    if (texture.image.data) {
+        return texture.image.data[y * texture.image.width + x];
+    } else {
+        if (!_canvas) {
+            _canvas = document.createElement('canvas');
+            _canvas.width = 1;
+            _canvas.height = 1;
+        }
+        const ctx = _canvas.getContext('2d');
+        ctx.drawImage(texture.image, x, y, 1, 1, 0, 0, 1, 1);
+        const d = ctx.getImageData(0, 0, 1, 1);
+        return THREE.Math.lerp(
+            layer.materialOptions.colorTextureElevationMinZ,
+            layer.materialOptions.colorTextureElevationMaxZ,
+            d.data[0] / 255);
+    }
+}
+
+function _updateSinglePoint(referenceCrs, layer, nodes, crs, toModify, zOffset) {
+    const pt = new Coordinates(referenceCrs, toModify).as(crs);
+    pt._values[2] = 0;
+
+    let tileWithValidElevationTexture = null;
+    for (const n of nodes) {
+        tileWithValidElevationTexture = tileAt(pt, n);
+        if (tileWithValidElevationTexture) {
+            break;
+        }
+    }
+
+    if (tileWithValidElevationTexture) {
+        const src = tileWithValidElevationTexture.getLayerTextures(l_ELEVATION)[0];
+        const width = src.image.width;
+        const height = src.image.height;
+
+        const tmp = new Extent(pt.crs, pt, pt);
+
+        // Assuming that tiles are split in 4 children, we lookup the parent that
+        // really owns this texture
+        const stepsUpInHierarchy = Math.log2(1.0 /
+            tileWithValidElevationTexture.materials[RendererConstant.FINAL].offsetScale[l_ELEVATION][0].z);
+        for (let i = 0; i < stepsUpInHierarchy; i++) {
+            tileWithValidElevationTexture = tileWithValidElevationTexture.parent;
+        }
+
+        const offset = tmp.offsetToParent(tileWithValidElevationTexture.extent);
+
+        const u = offset.x;
+        const v = offset.y;
+
+        // read elevation texture, using bi-linear filtering
+        const up = Math.max(0, u * (width - 1) - 0.5);
+        const vp = Math.max(0, v * (height - 1) - 0.5);
+
+        const u1 = Math.floor(up);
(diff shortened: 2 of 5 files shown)


peppsac:
Related issue: #181

peppsac:
Example screenshot from the demos:

capture d ecran de 2017-05-23 18-25-00

capture d ecran de 2017-05-23 18-34-10


Red cylinder is automatically placed above the ground, white/black cylinder is auto placed under the ground. Both adapt their position when the elevation texture changes.

examples/clamptoground.html
```diff
@@ -0,0 +1,115 @@
...
+                itowns.placeObjectOnGround(globeView.referenceCrs, globeView.wgs84TileLayer, globeView.wgs84TileLayer.level0Nodes, line);
+            };
+            itowns.View.prototype.addLayer.call(globeView, l);
+
+            for (const pt of [[4, 45], [5, 48]]) {
+                var group = new itowns.THREE.Group();
+
+                // red cylinder above the ground
+                var geometry = new itowns.THREE.BoxGeometry( 10, 10, 500);
+                var material = new itowns.THREE.MeshBasicMaterial( {color: 0xff0000} );
```
gchoqueux: Use `/src/Renderer/BasicMaterial.js` material with RTC (relative to camera), to remove `jittering`

examples/clamptoground.html
```diff
@@ -0,0 +1,115 @@
...
+                group.lookAt(p);
+                group.position.copy(p);
+                group.updateMatrix();
+
+                group.updateMatrixWorld(true);
+                contour.add(group);
+            }
+
+            globeView.scene.add(contour);
+            globeView.controls.setCameraTargetGeoPosition({longitude:4, latitude: 45}).then(() => {
```
gchoqueux: replace by :
```js
globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, () => {
                console.info('Globe initialized');
                menuGlobe.addImageryLayersGUI(globeView.getLayers(l => l.type === 'color'));
                menuGlobe.addElevationLayersGUI(globeView.getLayers(l => l.type === 'elevation'));
                globeView.controls.setCameraTargetGeoPosition({ longitude: 2.351323, latitude: 48.856712 }).then(() => globeView.controls.setCameraTargetGeoPositionAdvanced({ longitude: 4, latitude: 45, range: 10000 }));
 });
```

examples/clamptoground.html
```diff
@@ -0,0 +1,115 @@
...
+                var material = new itowns.THREE.MeshBasicMaterial( {color: 0xff0000} );
+                var cylinder = new itowns.THREE.Mesh( geometry, material );
+                cylinder.frustumCulled = false;
+                cylinder.position.z = 250;
+                cylinder.updateMatrix();
+                cylinder.layers.set(l.threejsLayer);
+                group.add(cylinder);
+
+                // black cylinder under the ground
+                material = new itowns.THREE.MeshBasicMaterial( {color: 0x000000} );
```
gchoqueux: format text file :  space...

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+            return undefined;
+        }
+
+        for (const c of tile.children) {
+            const t = tileAt(pt, c);
+            if (t) {
+                return t;
+            }
+        }
+        if (tile.materials[0].textures[0][0].coords.zoom > -1) {
```
gchoqueux: use `getLayerTextures`

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+            layer.materialOptions.colorTextureElevationMaxZ,
+            d.data[0] / 255);
+    }
+}
+
+function _updateSinglePoint(referenceCrs, layer, nodes, crs, toModify) {
+    const pt = new Coordinates(referenceCrs, toModify).as(crs);
+    pt._values[2] = 0;
+
+    let smallest = null;
```
gchoqueux: Find a more explicit name

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+    for (const n of nodes) {
+        smallest = tileAt(pt, n);
+        if (smallest) {
+            break;
+        }
+    }
+
+    if (smallest) {
+        const src = smallest.materials[0].textures[0][0];
```
gchoqueux: use function in LayeredMaterial

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+        const width = src.image.width;
+        const height = src.image.height;
+
+        const tmp = new Extent(pt.crs, pt, pt);
+
+        const goUp = Math.log2(1.0 / smallest.materials[0].offsetScale[0][0].z);
```
gchoqueux: use `l_Elevation`

gchoqueux: `goUp` find better name
`goUpInParent` of 'goDownInZoom'

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+        for (let i = 0; i < goUp; i++) {
```
gchoqueux: Factorize in `/src/Process/TiledNodeProcessing.js` : 
`getElevationFromTile(tile, coords)`

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return undefined;
+    }
+}
+
+let _canvas;
+function _readTextureValueAt(layer, texture, x, y) {
+    if (texture.image.data) {
+        return texture.image.data[y * texture.image.width + x];
+    } else {
```
gchoqueux, line 30: separate `readTexture` and `readImage`

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+        if (!_canvas) {
+            _canvas = document.createElement('canvas');
+            _canvas.width = 1;
+            _canvas.height = 1;
+        }
+        const ctx = _canvas.getContext('2d');
+        ctx.drawImage(texture.image, x, y, 1, 1, 0, 0, 1, 1);
+        const d = ctx.getImageData(0, 0, 1, 1);
+        return THREE.Math.lerp(
+            layer.materialOptions.colorTextureElevationMinZ,
```
gchoqueux, line 40: `colorTextureElevationMinZ`, What's this?

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+        }
+        if (tile.materials[0].textures[0][0].coords.zoom > -1) {
+            return tile;
+        }
+        return undefined;
+    }
+}
+
+let _canvas;
+function _readTextureValueAt(layer, texture, x, y) {
```
gchoqueux, line 27: move in tools file texture

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,126 @@
...
+
+    const crs = tiles[0].extent.crs();
+
+    if (!modifyGeometry) {
+        _updateSinglePoint(referenceCrs, layer, tiles, crs, obj.position);
+        obj.updateMatrix();
+        obj.updateMatrixWorld(true);
+    } else {
+        const geometry = obj.geometry;
+        if (geometry instanceof THREE.Geometry) {
```
gchoqueux, line 122: it's just for a surface?

gchoqueux (changes requested):
The object could be added to the tile where it lands, and refresh only when the tile is no longer displayed.

peppsac:
> The object could be added to the tile where it lands, and refresh only when the tile is no longer displayed.

You also want to refresh when the tile's elevation texture change.
Also we don't have a 'this tile got hidden' event, so even if we keep a reference to the tile we still need to perform a refresh everytime the scene changes. 

But I can add something like this:
```js
if (obj.tileAnchor.material.visible && obj.tileAnchor.materials[0].textures[0].id == obj.currentElevationTextureId) {
  // nothing changed: skip update
}
```

examples/clamptoground.html
peppsac: Why ?

src/Process/PlaceOnGroundProcessing.js
peppsac: See a5b326029171531086ff4be8af286f86dc7c5a6b

(color textures have values between 0 and 1, so to get a float elevation I added min/max z)

src/Process/PlaceOnGroundProcessing.js
peppsac: If you use it with  `modifyGeometry = true` yes.

src/Process/PlaceOnGroundProcessing.js
peppsac: What do you want to factorize exactly ?

examples/clamptoground.html
peppsac: Fixed.
But RTC is more or less unusable: it requires the user to compute/update RTC matrix. We need to either automatize this process or think of a better mechanism.

examples/clamptoground.html
peppsac: Done

src/Process/PlaceOnGroundProcessing.js
peppsac: Done

src/Process/PlaceOnGroundProcessing.js
peppsac: I disagree.
We'll see when another use case emerge - maybe it'll make sense to move this code in a separate place, maybe not.

src/Process/PlaceOnGroundProcessing.js
peppsac: I didn't do this yet, because I don't really see the benefits of this change.
If you really think it's important I'll do it though.

src/Process/PlaceOnGroundProcessing.js
peppsac: Done (`tileWithValidElevationTexture`)

src/Process/PlaceOnGroundProcessing.js
peppsac: Done

src/Process/PlaceOnGroundProcessing.js
peppsac: Done

peppsac (commented):
Branch updated.

examples/clamptoground.html
gchoqueux: We need to wait `itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED` before to move target because we use depth renderer  with globe

src/Process/PlaceOnGroundProcessing.js
gchoqueux: rename to `_readTextureOrImageValueAt`

peppsac: In both case I'm reading values from a texture.
But this texture can be a data texture or an image based texture.

examples/clamptoground.html
peppsac: Hmm ok.
I don't really understand why (looks like a bug to me) but it's not important for this PR so I'll change this.

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,131 @@
...
+        ctx.drawImage(texture.image, x, y, 1, 1, 0, 0, 1, 1);
+        const d = ctx.getImageData(0, 0, 1, 1);
+        return THREE.Math.lerp(
+            layer.materialOptions.colorTextureElevationMinZ,
+            layer.materialOptions.colorTextureElevationMaxZ,
+            d.data[0] / 255);
+    }
+}
+
+function _updateSinglePoint(referenceCrs, layer, nodes, crs, toModify, zOffset) {
```
gchoqueux, line 46: It would seem clearer to me, to move the texture reading logic in material or Tile. `PlaceOnGroundProcessing.js` would contain only the clamping logic.
```js
if (tileWithValidElevationTexture) {
        const z = tileWithValidElevationTexture.material.getElevation(pt);
        // or 
        const z = tileWithValidElevationTexture.getElevation(pt);

        pt._values[2] = z + zOffset;

        toModify.copy(pt.as(referenceCrs).xyz());
        return tileWithValidElevationTexture;
}
```
In addition, the `getElevation(coords)` function can also be used for `pickAltitude`.

src/Process/PlaceOnGroundProcessing.js
```diff
@@ -0,0 +1,131 @@
...
+
+        // Assuming that tiles are split in 4 children, we lookup the parent that
+        // really owns this texture
+        const stepsUpInHierarchy = Math.log2(1.0 /
+            tileWithValidElevationTexture.materials[RendererConstant.FINAL].offsetScale[l_ELEVATION][0].z);
+        for (let i = 0; i < stepsUpInHierarchy; i++) {
+            tileWithValidElevationTexture = tileWithValidElevationTexture.parent;
+        }
+
+        const offset = tmp.offsetToParent(tileWithValidElevationTexture.extent);
```
gchoqueux, line 73: It might be simpler to look for a tile where there's no offset?

gchoqueux (changes requested):

gchoqueux:
ok to do this :

```js
if (obj.tileAnchor.material.visible && obj.tileAnchor.materials[0].textures[0].id == obj.currentElevationTextureId) {
  // nothing changed: skip update
}
```

src/Process/PlaceOnGroundProcessing.js
peppsac: No we need the one that's displayed (I need to fix this as well).
We want to get the same result that the shader - so we need to sample at the same point (and the current logic is broken, since we should interpolate between the 3 vertices of the trianges used to rasterize the given point)...

src/Process/PlaceOnGroundProcessing.js
peppsac: Ok I'll split the texture reading part to a separate file

src/Process/PlaceOnGroundProcessing.js
gchoqueux: I think you would have the same result with the ancestor because in finality you look in the same texture.
Should we not take into account the size of the segments?

peppsac: No, the result will be different because the density of vertices depends on the tile.

e.g: let's say you want to read the value at `x:9, y:0`.
In the tile owning the texture (scale = 1), the nearest vertices will be at `0, 0` and `15, 0` (because the tile has 16 vertices, spread over 256 pixels).
In its child texture (scale = 0.5), nearest vertices will be at `8, 0`,  `15, 0` so even if you read from the same texture you'll get different results.

peppsac:
Closing for now. I'll re-open it when the last issues are sorted out.

PR #316, opened 2017-05-31
feature: remove delay param from notifyChange

peppsac:
Removed because it's a rarely used feature, and with 2 different use case, and both
can be implemented in a better way.

- reduce CPU usage (see Scheduler usage): iTowns partial update support cover this use-case,
with the benefit of not adding any latency.

- rescheduling an update after an error (see Layered...Processing.js): the user can use setTimeout
directly

### DIFF
--- a/examples/planar.html
+++ b/examples/planar.html
@@ -72,7 +72,7 @@ and open the template in the editor.
             itowns.proj4.defs('EPSG:3946',
                 '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
 
-            const menuGlobe = new GuiTools(itowns.viewer, 'menuDiv');
+            const menuGlobe = new GuiTools('menuDiv');
 
             const bbox = new itowns.Extent(
                 'EPSG:3946',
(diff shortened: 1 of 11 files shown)


src/Core/MainLoop.js
```diff
@@ -58,6 +58,14 @@ MainLoop.prototype._update = function _update(view) {
 };
 
 MainLoop.prototype._step = function _step(view) {
+    const willRedraw = this.needsRedraw;
+
+    // Reset internal state before calling _update (so future calls to View.notifyChange()
+    // can properly change it)
+    this.needsRedraw = false;
+    this.renderingState = RENDERING_PAUSED;
+    view._changeSources.clear();
```
autra: Why did you move these 2 lines? Should they be called *after* calling `this._update` below?

examples/planar.html
```diff
@@ -158,7 +158,7 @@
                 view.camera.camera3D.updateMatrix();
                 view.camera.camera3D.updateMatrixWorld(true);
 
-                view.notifyChange(20, true);
+                view.notifyChange(true);
```
autra, line 161: You're not using setTimeout here, can you elaborate why it's not needed?

autra (changes requested):

examples/planar.html
peppsac: The delay was a left over from a test.
Here we want a continuous animation, so we want to keep the mainLoop running permanently.

src/Core/MainLoop.js
peppsac: Indeed I need to keep a copy of changeSources() before clearing it. Thanks :)

autra (approved):
LGTM, thanks :-)

PR #318, opened 2017-05-31
doc (jsdoc): event documentation improvements

vcoindet:
Documentation for the events : 

- sequence-property-changed

- opacity-property-changed

- visible-property-changed

- layer-removed

- layer-added

- initialized

- layers-order-changed

- pan-changed

- orientation-changed

- range-changed

- camera-target-changed

### DIFF
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -1,5 +1,35 @@
 import { EventDispatcher } from 'three';
 
+/**
+ * Fires when layer sequence change (meaning when the order of the layer changes in the view)
+ * @event Layer#sequence-property-changed
+ * @property new {object}
+ * @property new.sequence {number} the new value of the layer sequence
+ * @property previous {object}
+ * @property previous.sequence {number} the previous value of the layer sequence
+ * @property target {Layer} dispatched on layer
+ * @property type {string} sequence-property-changed
+*/
+/**
+ * Fires when layer opacity change
+ * @event Layer#opacity-property-changed
+ * @property new {object}
+ * @property new.opacity {object} the new value of the layer opacity
+ * @property previous {object}
+ * @property previous.opacity {object} the previous value of the layer opacity
+ * @property target {Layer} dispatched on layer
+ * @property type {string} opacity-property-changed
+*/
+/**
+ * Fires when layer visibility change
+ * @event Layer#visible-property-changed
+ * @property new {object}
+ * @property new.visible {object} the new value of the layer visibility
+ * @property previous {object}
+ * @property previous.visible {object} the previous value of the layer visibility
+ * @property target {Layer} dispatched on layer
+ * @property type {string} visible-property-changed
+*/
 
 export const defineLayerProperty = function defineLayerProperty(layer, propertyName, defaultValue, onChange) {
     const existing = Object.getOwnPropertyDescriptor(layer, propertyName);
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -18,6 +18,43 @@ import { updateLayeredMaterialNodeImagery, updateLayeredMaterialNodeElevation } 
 import { globeCulling, preGlobeUpdate, globeSubdivisionControl, globeSchemeTileWMTS, globeSchemeTile1 } from '../../Process/GlobeTileProcessing';
 import BuilderEllipsoidTile from './Globe/BuilderEllipsoidTile';
 
+/**
+ * Fires when the view is completely loaded. Controls and view's functions can be called then.
+ * @event GlobeView#initialized
+ * @property target {view} dispatched on view
+ * @property type {string} initialized
+ */
+/**
+ * Fires when a layer is added
+ * @event GlobeView#layer-added
+ * @property layerId {string} the id of the layer
+ * @property target {view} dispatched on view
+ * @property type {string} layers-added
+ */
+/**
+ * Fires when a layer is removed
+ * @event GlobeView#layer-removed
+ * @property layerId {string} the id of the layer
+ * @property target {view} dispatched on view
+ * @property type {string} layers-added
+ */
+/**
+ * Fires when the layers oder has changed
+ * @event GlobeView#layers-order-changed
+ * @property new {object}
+ * @property new.sequence {array}
+ * @property new.sequence.0 {number} the new layer at position 0
+ * @property new.sequence.1 {number} the new layer at position 1
+ * @property new.sequence.2 {number} the new layer at position 2
+ * @property previous {object}
+ * @property previous.sequence {array}
+ * @property previous.sequence.0 {number} the previous layer at position 0
+ * @property previous.sequence.1 {number} the previous layer at position 1
+ * @property previous.sequence.2 {number} the previous layer at position 2
+ * @property target {view} dispatched on view
+ * @property type {string} layers-order-changed
+ */
+
 
 /**
  * Globe's EVENT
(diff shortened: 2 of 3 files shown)


src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,59 @@
 import { EventDispatcher } from 'three';
 
+/**
+ * Layer's sequence (index) event
+ * @event Layer#sequence-property-changed
+ * @property sequence-property-changed {string} emit after the index of a layer has changed
```
tbroyer: From the example, it looks like the event has no property named `sequence-property-changed`, but 5 properties: `layerId`, `new`, `previous`, `target`, and `type` (and `new.sequence` and `previous.sequence`)

http://usejsdoc.org/tags-event.html

src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,59 @@
...
+ * @example
+ * const layer = viewer.getLayers(l => l.id == 'layerId')[0];
+ * layer.addEventListener('sequence-property-changed', (event) => { console.log(event); });
+ * // returns:
```
tbroyer: "prints" rather than "returns"?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -241,6 +241,7 @@ let enableEventPositionChanged = true;
 
 /**
  * globe controls events
+ * @event
  * @property PAN_CHANGED {string} emit after camera pan
```
tbroyer: Should those be changed to individual `@event` for each possible event (`GlobeControls#pan-changed`?), each with its own set of `@property` to document the event's properties? (instead of, or in addition to, the examples)

…and those `@property` could possibly be kept to document the _constants_ one could use (instead of the event name as string), but I think `@event` here is wrong.

vcoindet:
Modified and rebased.
WDYT now @tbroyer ?

tbroyer (approved):
Haven't verified what it'll look like in the generated doc, but this at least seem to follow the correct usage of the tags.

src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,77 @@
...
+ * Layer's sequence (index) event
+ * @event Layer#sequence-property-changed
+ * @property layerId {string} the id of the layer
+ * @property new {object}
+ * @property new.sequence {number} the new value of the layer sequence
+ * @property previous {object}
+ * @property previous.sequence {number} the previous value of the layer sequence
+ * @property target {Layer} dispatched on layer
+ * @property type {string} sequence-property-changed
+ * @example
```
autra: It's nice to have examples, but imho these gives little value to the doc. It's too verbose for the gain. We can remove them.

src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,77 @@
...
 
+/**
+ * Layer's sequence (index) event
+ * @event Layer#sequence-property-changed
+ * @property layerId {string} the id of the layer
+ * @property new {object}
+ * @property new.sequence {number} the new value of the layer sequence
+ * @property previous {object}
+ * @property previous.sequence {number} the previous value of the layer sequence
+ * @property target {Layer} dispatched on layer
```
autra, line 10: You don't need the layerId if layer is in target.

src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,77 @@
 import { EventDispatcher } from 'three';
 
+/**
+ * Layer's sequence (index) event
```
autra: `sequence` is not really clear from an outside POV. We are talking about the order in the stack right? Maybe add a little phrase explaining that.

src/Core/Layer/Layer.js
```diff
@@ -10,7 +82,7 @@ export const defineLayerProperty = function defineLayerProperty(layer, propertyN
             { get: () => property,
                 set: (newValue) => {
                     if (property !== newValue) {
-                        const event = { type: `${propertyName}-property-changed`, previous: {}, new: {} };
+                        const event = { layerId: layer.id, type: `${propertyName}-property-changed`, previous: {}, new: {} };
```
autra: Probably not needed, see above comment

src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,77 @@
...
+ * //   new: Object
+ * //       sequence:    'new sequence'
+ * //   previous: Object
+ * //       sequence:    'previous sequence'
+ * //   target: Layer
+ * //   type: "sequence-property-changed"
+ *
+*/
+/**
+ * Layer's opacity event
```
autra: > Fires when layer opacity change

src/Core/Layer/Layer.js
```diff
@@ -1,5 +1,77 @@
...
+ * //   new: Object
+ * //       opacity:    'new opacity'
+ * //   previous: Object
+ * //       opacity:    'previous opacity'
+ * //   target: Layer
+ * //   type: "opacity-property-changed"
+ *
+*/
+/**
+ * Layer's visibility event
```
autra: Fires when...

src/Core/Prefab/GlobeView.js
```diff
@@ -18,6 +18,74 @@ import { updateLayeredMaterialNodeImagery, updateLayeredMaterialNodeElevation }
 import { globeCulling, preGlobeUpdate, globeSubdivisionControl, globeSchemeTileWMTS, globeSchemeTile1 } from '../../Process/GlobeTileProcessing';
 import BuilderEllipsoidTile from './Globe/BuilderEllipsoidTile';
 
+/**
+ * Emit when the globe is initialized
```
autra: What does that exactly means? (what is initialized? What can the user expect when this fires?)

src/Core/Prefab/GlobeView.js
```diff
@@ -27,27 +95,22 @@ import BuilderEllipsoidTile from './Globe/BuilderEllipsoidTile';
  * @property COLOR_LAYERS_ORDER_CHANGED {string} emit when  color layers order change
  * @example
  * viewer.addEventListener(itowns.GLOBE_VIEW_EVENTS.LAYER_ADDED, (event) => console.log(event));
- * // or
- * viewer.addEventListener('layer-added', (event) => console.log(event));
- * // returns:
- * // Object {type: "layer-added", target: GlobeView}
+ * // prints:
```
autra: Same. These examples are trivial. Not worthy to put them here.

autra (changes requested):

src/Core/Layer/Layer.js
vcoindet: I think it's quite good to have some examples in the doc. Maybe those ones are a bit too detailed, maybe I can remove the prints part and leave just how to call the function. WDYT ?

autra: I think this example is too trivial to be interesting. Do we need to explain to js dev how to call `addEventListener`? What devs will look for is the name of the possible events, and what properties these `event` contains when they fire.

vcoindet: OK for me

vcoindet:
Rebased and ready @autra 
Regarding your comments, I also removed examples in GlobeControls.js concerning the events.

autra:
Thanks @vcoindet. What about my comment on `layerID`?

vcoindet:
Sorry, forgot to remove it from the doc !
Doing it right now

vcoindet:
done @autra

autra (approved):

autra:
Thanks! 

( for commit messages, we're following [this](https://seesparkbox.com/foundry/semantic_commit_messages), so it should be prefixed by `docs` and not doc. it's better to be exact if we want to grep the log one day)

vcoindet:
OK noticed, my mistake

PR #319, opened 2017-05-31
Fix debug and renderer

peppsac:
1st commit: fix debug handling  - reduce itowns.js size by ~40%

2nd commit: minor fix to make sure external renderer really works in itowns

Thanks!

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -87,6 +87,8 @@
                 }
             }
 
+            new debug.Debug(globe, viewerDiv);
+
             itowns.View.prototype.addLayer.call(globe, debug3dt, $3dTilesLayer);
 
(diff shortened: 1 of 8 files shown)


examples/3dtiles.html
```diff
@@ -87,6 +87,8 @@
                 }
             }
 
+            debug.Debug(globe, viewerDiv);
```
autra: it should be used as a constructor normally: `new debug.Debug`

autra (changes requested):
LGTM

examples/3dtiles.html
peppsac: OK, fixed.

autra (approved):

PR #320, opened 2017-05-31
Use newer version of dat.gui

autra:
Now it is installable from github. No need to pass through npm.

@iTowns/reviewers please r?


autra:
Closing this for now, as apparently the npm config of dat.gui does not work with node 8 (depends on a node-sass version that does not work with node 8 yet).

I might just copy dat.gui.min.js somewhere in the code then :-/

autra:
Reopening as it seems the issue will get fixed. I will forget if I leave that closed. Don't merge now :-)

autra:
To track: https://github.com/sass/node-sass/issues/1985

autra:
@mbredif why did you assign @dgricci here? It looks like a mistake, so I'm assigning myself. Don't hesitate to revert (and explain why) if I missed something.

gchoqueux:
There is still no longer slider for opacity?
Like speed in this [example](https://workshop.chromeexperiments.com/examples/gui/#1--Basic-Usage)

autra:
@gchoqueux I want to update dat.gui for this very reason.

gchoqueux:
cool
Rebase please
and I will merge

autra:
Updated, but I'm hesitating... The problem:

`dat.gui` depends on `sassify` which in turns depends on `node-sass`3.8. There's prebuilt binary for it for node 6 (and `npm install` downloads them), but for node 8, there is no pre-built binary for node-sass 3.8. So `npm install` build it. The problem is that it is very long (ok that'll happen only the first time, but still). 

To fix this, sassify would need to update to node-sass 4, and *then* dat.gui would need to update sassify. Regarding the fact that the last sassify commit was nearly a year ago, that's unlikely to happen soon... All  of that for a bit of UI for debug purpose...

Possible solutions:
- live with it
- push upstream to make things change
- embed dat.gui.js directly in examples/
- other solution ?

peppsac:
Given that `dat.gui.min.js` is only 50KB I'm fine with copying it directly to examples and remove the dat.gui npm module.

autra:
@iTowns/reviewers please r?

peppsac (approved):
LGTM, thanks.

Issue #314, opened 2017-05-26
Documentation tool change proposal


peppsac:
I wrote some documentation over the past few days and I feel that jsdoc isn't quite the tool I need.

As I may not be the only feeling this way I propose to discuss this issue here to settle the choice on our documentation tool.

My main issue regarding jsdoc is its special syntax: it clutters the code with information only useful to their tool (@param, @constructor, etc).

I took a quick look at other tools, and found [docco](https://github.com/jashkenas/docco) which is really nice.
Basically it puts no constraint on the programmer, and simply generate a documentation site from the comments in the code (see http://coffeescript.org/#literate for the inspiration).

https://github.com/iTowns/itowns2/tree/documentation is my attempt at using docco for itowns. You can test by using the new command `npm run doc2` or you can see the result here: http://37.187.164.233/itowns-docs/

A few notes:
- I've modified 2 examples, so they are documented and functionnal on the API doc site
- take a look at the source code of the commented files when using Docco (e.g https://github.com/iTowns/itowns2/blob/documentation/src/Core/View.js)
- I modified docco to add a few short cut for our use case:
   * internal autolink: if you write "Camera*" it will automatically insert a link to the documentation of Camera
   * sidebar generation (as request by @vcoindet :)) and fuzzysearch (using implementation from https://blog.forrestthewoods.com/reverse-engineering-sublime-text-s-fuzzy-match-4cffeed33fdb#.7i6md382t)
  * public function signature extraction
- docco is a 200 lines project, so it's very easy to adapt :)
  
Let me know what you think :)

vcoindet:
Nice work !
I have to say, I wasn't conviced at first by the result, but I think we can do something better with docco than with jsdoc. It seems to be cleaner with docco ! ;)

I have a few observations on the result : 

- jsdoc natively propose tables to describe parameters of the functions and I think it's more clean and more readable. It is probably possible to do tables with docco (as it is in mardown) but the effort will more important and the maintenance too...
- thanks for the sidebar : do you think its position can be fixed, so when we scrolled it would always appeared on the screen ?
- fuzzysearch is a really good thing, do you think we can use it for all methods (such as addLayer for example) ?
- the "show/hide code" is better than in jsdoc ! ;)

peppsac:
Thanks for your feeback @vcoindet :)

- tables for arguments: yes you can use the markdown syntax for tables.

| name        | type   |  Description                       |
|-------------|--------|------------------------------------|
[...]

PR #321, opened 2017-06-01
Turn on ESLint valid-jsdoc rule (as warning for now)

tbroyer:
Don't require descriptions for parameter and return tags:
type matters most, and description can sometimes be redundant
and useless, making it annoying and noisy.

#314

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,6 +43,11 @@ module.exports = {
       }
     }],
     'one-var': ['error', 'never'],
+    'valid-jsdoc': ['error', {
+      requireReturn: false,
+      requireParamDescription: false,
+      requireReturnDescription: false,
+    }],
     // TODO reactivate this rule once a proper npm package is made
     // a good configuration might be:
     /*'import/no-extraneous-dependencies': ['error', {
--- a/src/Core/AnimationPlayer.js
+++ b/src/Core/AnimationPlayer.js
@@ -80,7 +80,6 @@ const setPlayerState = function setPlayerState(player, state) {
 };
 
 /**
- * AnimationPlayer
  * It can play, pause or stop Animation or AnimationExpression (See below).
  * AnimationPlayer is needed to use Animation or AnimationExpression
  * AnimationPlayer emits events :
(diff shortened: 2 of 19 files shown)


vcoindet:
SGTM.

In our case, it returns a lot of warning as we put some jsdoc syntax in some files that won't appear in the final doc... maybe these should be just normal comments to avoid these warnings.

tbroyer:
> maybe these should be just normal comments to avoid these warnings.

or fixed? ;-)

autra:
> descriptions for parameter and return tags:
type matters most

I disagree with this, especially in js. Often, a good js code is one where type is either so evident it's not worth mentioning, or where it does not really matter. Ok, sometimes you need to be precise about it, but let's not make it compulsory. If we want a typed language, let's use typescript :-)

Btw I'm actually tempted to try esdoc instead of jsdoc for its ability to infer these types (I need to check the extent of this ability though).

Also, I disagree with the very principle of warnings, especially when it generates so many of them. By experience, they will never get corrected and *will* get ignored. To me, it's error or nothing. This politics would also have the nice side-effect of not putting the burden on other devs, especially the one who has nice enough to correct them (I'm speaking about me there, I have corrected enough eslint problem to have a say on this :-) )

autra:
> or fixed? ;-)

You've said it.

vcoindet:
yeah it can be fixed too, I will correct them in a new PR ! ;)

tbroyer:
> To me, it's error or nothing. […] (I'm speaking about me there, I have corrected enough eslint problem to have a say on this :-) )

OK, let's do that: I'll fix the jsdoc comments / make them non-jsdoc and update the rule to `'error'`.

vcoindet:
OK, so I'll let you take care of it

tbroyer:
Forgot to respond to this comment:

> I disagree with this, especially in js. Often, a good js code is one where type is either so evident it's not worth mentioning, or where it does not really matter. Ok, sometimes you need to be precise about it, but let's not make it compulsory. If we want a typed language, let's use typescript :-)

It's not about us, it's about our users.

I'd be OK with switching from `valid-jsdoc` to the `eslint-plugin-jsdoc` which allows not requiring parameter types (https://www.npmjs.com/package/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-require-param-type).

gchoqueux (approved):
Thanks

gchoqueux:
Rebase and I merge

autra:
thanks @tbroyer

PR #323, opened 2017-06-01
3d-tiles improvements and points early support

NikoSaul:
This pull request features:
- Pnts format early support
- Support for region and sphere bounding volumes
- Partial support of viewerRequestVolume

Another tileset has been added to the 3d-tiles example for testing purposes.

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import Provider from './Provider';
 import B3dmLoader from '../../../Renderer/ThreeExtended/B3dmLoader';
+import PntsLoader from '../../../Renderer/ThreeExtended/PntsLoader';
 import Fetcher from './Fetcher';
 import BasicMaterial from '../../../Renderer/BasicMaterial';
 import OBB from '../../../Renderer/ThreeExtended/OBB';
@@ -94,6 +95,12 @@ $3dTiles_Provider.prototype.b3dmToMesh = function b3dmToMesh(data, layer) {
     });
 };
 
+$3dTiles_Provider.prototype.pntsParse = function pntsParse(data) {
+    return new Promise((resolve) => {
+        resolve(PntsLoader.parse(data));
+    });
+};
+
 function configureTile(tile, layer, metadata) {
     tile.frustumCulled = false;
     tile.loaded = true;
@@ -124,6 +131,7 @@ $3dTiles_Provider.prototype.executeCommand = function executeCommand(command) {
 
         const supportedFormats = {
             b3dm: this.b3dmToMesh.bind(this),
+            pnts: this.pntsParse.bind(this),
         };
 
         return Fetcher.arrayBuffer(url).then((result) => {
@@ -136,6 +144,8 @@ $3dTiles_Provider.prototype.executeCommand = function executeCommand(command) {
                     layer.tileIndex.extendTileset(result, metadata.tileId, newPrefix);
                 } else if (magic == 'b3dm') {
                     func = supportedFormats.b3dm;
+                } else if (magic == 'pnts') {
+                    func = supportedFormats.pnts;
                 } else {
                     Promise.reject(`Unsupported magic code ${magic}`);
                 }
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -113,6 +113,7 @@ function configureTile(tile, layer, metadata) {
     tile.tileId = metadata.tileId;
     tile.additiveRefinement = (metadata.refine === 'add');
     tile.boundingVolume = getBox(metadata.boundingVolume);
+    tile.viewerRequestVolume = metadata.viewerRequestVolume ? getBox(metadata.viewerRequestVolume) : undefined;
 }
 
 const textDecoder = new TextDecoder('utf-8');
@@ -150,6 +151,7 @@ $3dTiles_Provider.prototype.executeCommand = function executeCommand(command) {
                     Promise.reject(`Unsupported magic code ${magic}`);
                 }
                 if (func) {
+                    // TODO: request should be delayed if there is a viewerRequestVolume
                     return func(result, layer).then((content) => {
                         tile.add(content);
                         tile.traverse(setLayer);
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -3,6 +3,9 @@ import Provider from './Provider';
 import B3dmLoader from '../../../Renderer/ThreeExtended/B3dmLoader';
 import Fetcher from './Fetcher';
 import BasicMaterial from '../../../Renderer/BasicMaterial';
+import OBB from '../../../Renderer/ThreeExtended/OBB';
+import Extent from '../../Geographic/Extent';
+import MathExtended from '../../Math/MathExtended';
 
 
 export function $3dTilesIndex(tileset, urlPrefix) {
@@ -46,13 +49,18 @@ $3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(/
 
 };
 
-function getBox(boundingVolume) {
-    if (boundingVolume.region) {
-        throw new Error('Region bounding volume not supported yet');
-        // return { region: boundingVolume.region };
-    } else if (boundingVolume.box) {
+function getBox(volume) {
+    if (volume.region) {
+        const region = volume.region;
+        const extent = new Extent('EPSG:4326', MathExtended.radToDeg(region[0]), MathExtended.radToDeg(region[2]), MathExtended.radToDeg(region[1]), MathExtended.radToDeg(region[3]));
+        const box = OBB.extentToOBB(extent, region[4], region[5]);
+        box.position.copy(box.centerWorld);
+        box.updateMatrix();
+        box.updateMatrixWorld();
+        return { region: box };
+    } else if (volume.box) {
         // TODO: only works for axis aligned boxes
-        const box = boundingVolume.box;
+        const box = volume.box;
         // box[0], box[1], box[2] = center of the box
         // box[3], box[4], box[5] = x axis direction and half-length
         // box[6], box[7], box[8] = y axis direction and half-length
@@ -66,9 +74,9 @@ function getBox(boundingVolume) {
         const t = center.z + box[11];
 
         return { box: new THREE.Box3(new THREE.Vector3(w, s, b), new THREE.Vector3(e, n, t)) };
-    } else if (boundingVolume.sphere) {
-        throw new Error('Sphere bounding volume not supported yet');
-        // return { sphere: undefined };
+    } else if (volume.sphere) {
+        const sphere = new THREE.Sphere(new THREE.Vector3(volume.sphere[0], volume.sphere[1], volume.sphere[2]), volume.sphere[3]);
+        return { sphere };
     }
 }
 
--- a/src/Main.js
+++ b/src/Main.js
@@ -19,7 +19,6 @@ export { Extent };
 export { Coordinates };
 export { Fetcher };
 
-
 // Others can be directly exported
 export { UNIT } from './Core/Geographic/Coordinates';
 export { processTiledGeometryNode, initTiledGeometryLayer } from './Process/TiledNodeProcessing';
--- a/src/Main.js
+++ b/src/Main.js
@@ -19,6 +19,7 @@ export { Extent };
 export { Coordinates };
 export { Fetcher };
 
+
 // Others can be directly exported
 export { UNIT } from './Core/Geographic/Coordinates';
 export { processTiledGeometryNode, initTiledGeometryLayer } from './Process/TiledNodeProcessing';
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -46,6 +46,27 @@ function subdivideNode(context, layer, node) {
 }
 
 export function $3dTilesCulling(node, camera) {
+    // For viewer Request Volume https://github.com/AnalyticalGraphicsInc/3d-tiles-samples/tree/master/tilesets/TilesetWithRequestVolume
+    if (node.viewerRequestVolume) {
+        const nodeViewer = node.viewerRequestVolume;
+        if (nodeViewer.region) {
+            // TODO
+            return true;
+        }
+        if (nodeViewer.box) {
+            // TODO
+            return true;
+        }
+        if (nodeViewer.sphere) {
+            const worldCoordinateCenter = nodeViewer.sphere.center.clone();
+            worldCoordinateCenter.applyMatrix4(node.matrixWorld);
+            // To check the distance between the center sphere and the camera
+            if (!(camera.camera3D.position.distanceTo(worldCoordinateCenter) <= nodeViewer.sphere.radius)) {
+                return true;
+            }
+        }
+    }
+
     // For bounding volume
     if (node.boundingVolume) {
         const boundingVolume = node.boundingVolume;
@@ -66,7 +87,7 @@ let preSSE;
 export function pre3dTilesUpdate(context) {
     // pre-sse
     const hypotenuse = Math.sqrt(context.camera.width * context.camera.width + context.camera.height * context.camera.height);
-    const radAngle = context.camera.FOV * Math.PI / 180;
+    const radAngle = context.camera.camera3D.fov * Math.PI / 180;
 
      // TODO: not correct -> see new preSSE
     // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -46,13 +46,18 @@ function subdivideNode(context, layer, node) {
 }
 
 export function $3dTilesCulling(node, camera) {
-    if (node.boundingVolume.region) {
-        // TODO
-    }
-    if (node.boundingVolume.box) {
-        return !camera.isBox3DVisible(node.boundingVolume.box, node.matrixWorld);
-    } else if (node.boundingVolume.sphere) {
-        // TODO return !camera.isSphereVisible(node.boundingVolume.sphere, node.matrixWorld);
+    // For bounding volume
+    if (node.boundingVolume) {
+        const boundingVolume = node.boundingVolume;
+        if (boundingVolume.region) {
+            return !camera.isBox3DVisible(boundingVolume.region.box3D, boundingVolume.region.matrixWorld);
+        }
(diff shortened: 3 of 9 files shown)


src/Main.js
```diff
@@ -8,6 +8,7 @@ import PlanarView from './Core/Prefab/PlanarView';
 import Extent from './Core/Geographic/Extent';
 import Coordinates from './Core/Geographic/Coordinates';
 import Fetcher from './Core/Scheduler/Providers/Fetcher';
+import OBBHelper from '../utils/debug/OBBHelper';
```
peppsac: OBBHelper is a debug feature, so it should be exported only in `debug.js`
To do so, simply add:
```js
export { OBBHelper };
```
to `Debug.js` (since it's the entry point of debug.js, see `webpack.config.js`)

Then, you can use `new debug.OBBHelper(...)`

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -48,8 +52,14 @@ $3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(/
 
 function getBox(boundingVolume) {
     if (boundingVolume.region) {
-        throw new Error('Region bounding volume not supported yet');
-        // return { region: boundingVolume.region };
+        const region = boundingVolume.region;
+        const radToDeg = 180 / Math.PI;
```
peppsac: `MathExtended.js` have functions to perform rad <-> deg conversions so let's use them instead.

src/Process/3dTilesProcessing.js
```diff
@@ -46,13 +46,40 @@ function subdivideNode(context, layer, node) {
...
+        if (nodeViewer.box) {
+            // TODO
+            return true;
+        }
+        if (nodeViewer.sphere) {
+            const worldCoordinateCenter = nodeViewer.sphere.center.clone();
+            worldCoordinateCenter.applyMatrix4(node.matrixWorld);
+            // To check the distance between the sphere and the camera
+            const distance = Math.max(0.0, camera.camera3D.position.distanceTo(worldCoordinateCenter) - nodeViewer.sphere.radius);
+            if (!(distance <= nodeViewer.sphere.radius)) {
```
peppsac: This test seems wrong. I would write:
```js
if (camera.camera3D.position.distanceTo(worldCoordinateCenter) <= nodeViewer.sphere.radius) {
   return true;
}
```
Am I missing something?

src/Renderer/Camera.js
```diff
@@ -107,4 +107,19 @@ Camera.prototype.box3DSizeOnScreen = function box3DSizeOnScreen(box3d, matrixWor
     return c.applyMatrix4(m);
 };
 
+Camera.prototype.isSphereVisible = function isSphereVisible(sphere, matrixWorld) {
```
peppsac: Maybe you could extract the shared code between this function and isBox3dVisible to a private function ?

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -48,8 +52,14 @@ $3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(/
...
+        const extent = new Extent('EPSG:4326', region[0] * radToDeg, region[2] * radToDeg, region[1] * radToDeg, region[3] * radToDeg);
```
peppsac: Does 3d-tiles enforces EPSG:4326 usage?

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -97,4 +98,74 @@ OBB.prototype._cPointsWorld = function _cPointsWorld(points) {
     return points;
 };
 
+// get oriented bounding box of tile
+OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
```
peppsac, line 102: If you move this function here, you might want to add some asserts at the beginning. At the very least: it'll only work with Geographic coordinates so check `extent`'s crs.

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,119 @@
+import * as THREE from 'three';
+
+function PntsLoader() {
```
peppsac: AFAICT this is not useful.
You can direclty do:
```js
export default {
   parse: function parse(buffer, layer) { ...}
}
```

peppsac (changes requested):
Thanks ! Happy to see 3d-tiles points support progressing.

I did a quick review, leaving mostly stylistic comments.

I'll do another more in depth review later. One of the open question is: I have a branch ready which can display PotreeConverter binary points format ; so we need to think about what can be shared and what needs to be specific to each implementation...

src/Core/Scheduler/Providers/3dTiles_Provider.js
NikoSaul: Yes it does: https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/schema/boundingVolume.schema.json#L19

vpicavet: That's actually annoying, since we would like to use 3d tiles for projected data, and avoid re-projections as much as possible. Could we implement support for any projection in iTowns, and try to get this support in the standard afterwards ?

Jeremy-Gaillard: There are multiple bounding volumes types. Regions are used for globe portions, but there's the simpler box type that can be used in local coordinates.

vpicavet: ok perfect then

src/Core/Scheduler/Providers/3dTiles_Provider.js
NikoSaul: Ho yes thank you, i don't see the MathExtented class :+1:

src/Process/3dTilesProcessing.js
NikoSaul: No you're right, it's good I corrected this problem

NikoSaul:
Thank you for your first return, I tried to fix everything, I hope it is good.

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,113 @@
...
+            throw new Error('Invalid pnts file.');
+        }
+    },
+};
+
+function parseFeatureBinary(array, byteOffset, FTJSONLength, buffer, layer) {
+    const subArrayJson = decodeFromCharCode(array.subarray(byteOffset, FTJSONLength + byteOffset));
+    const parseJSON = JSON.parse(subArrayJson);
+
+    const view = new Float32Array(buffer);
```
peppsac: This is doing a full copy of buffer.
You should pass the arrayBuffer to this function and create directly the positionArray below:
```
const positionArray = new Float32Array(buffer,  byteOffsetPos, lengthFeature * 3);
```
(and the same is true for `colorArray`)

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,113 @@
...
+    let lengthFeature;
+
+    if (parseJSON.POINTS_LENGTH) {
+        lengthFeature = parseJSON.POINTS_LENGTH;
+    }
+
+    if (parseJSON.POSITION)
```
peppsac: Nit: make sure to follow the code style adopted : the `{` should be on the same line (for `if` and function declarations)

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,113 @@
...
+    {
+        const byteOffsetPos = (parseJSON.POSITION.byteOffset + subArrayJson.length + byteOffset) / 4;
+        const positionArray = new Float32Array(view.subarray(byteOffsetPos, (lengthFeature * 3) + byteOffsetPos));
+        if (parseJSON.RGB) {
+            const byteOffsetCol = parseJSON.RGB.byteOffset + subArrayJson.length + byteOffset;
+            const colorArray = new Uint8Array(array.subarray(byteOffsetCol, (lengthFeature * 3) + byteOffsetCol));
+            return createGeometryPoints(positionArray, colorArray, lengthFeature, layer);
+        }
+    }
```
peppsac: Consider throwing an error for unsupported configurations.

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,113 @@
...
+function createGeometryPoints(positions, colors, numPoints, layer)
+{
+    const geometry = new THREE.BufferGeometry();
+    geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+    geometry.addAttribute('color', new THREE.BufferAttribute(colors, 3, true));
+
+    const material = new THREE.PointsMaterial({ size: 0.05, vertexColors: THREE.VertexColors, sizeAttenuation: true });
+    const points = new THREE.Points(geometry, material);
+
+    points.layers.set(layer);
```
peppsac: The `layer` param doesn't seem to be a three.js layer (`func(result, command.layer)` in 3dTilesProvider) so this won't work.

You already set the layers properly 3dTilesProvider I believe you can simply remove the call the `points.layer.set(...)` here and all references to layer (I don't think this file should be aware of layers anyway).

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,113 @@
...
+/*
+    read Feature table json header
+*/
+/*
+function readTestFT(array, byteOffset, FTJSONLength) {
+    console.log(decodeFromCharCode(array.subarray(byteOffset, FTJSONLength + byteOffset))); // 28 - 8 byte of header
+}
+*/
+
+function decodeFromCharCode(value) {
```
peppsac: I think you can replace this function (or at least its content) with: `String.fromCharCode(...value)`. WDYT?

src/Renderer/Camera.js
```diff
@@ -81,19 +81,25 @@ Camera.prototype.setRotation = function setRotation(rotation) {
 
 const temp = new THREE.Vector3();
 const frustum = new THREE.Frustum();
-const obbViewMatrix = new THREE.Matrix4();
-Camera.prototype.isBox3DVisible = function isBox3DVisible(box3d, matrixWorld) {
+const localViewMatrix = new THREE.Matrix4();
+function localTranslationFrutum(matrixWorld, camera) {
```
peppsac: Typo: localTranslationFrustum

src/Main.js
```diff
@@ -19,6 +19,7 @@ export { Extent };
 export { Coordinates };
 export { Fetcher };
 
+
```
peppsac: Nit: do not add an empty line :)

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -32,6 +36,7 @@ function $3dTiles_Provider(/* options*/) {
 
     Provider.call(this);
     this.b3dmLoader = new B3dmLoader();
+    this.pntsLoader = PntsLoader;
```
peppsac: This isn't really needed now that PntsLoader isn't an object anymore.

peppsac (changes requested):
Almost ready for merging.

When you'll rebase your commits, please fix the prefix to match  the ones from https://seesparkbox.com/foundry/semantic_commit_messages (and use lowercase). Thanks :)

qdnguyen:
Nice to hear  region and sphere are supported in 3dTiles_Provider but i did not see any points on globe? The data file (points.pnts) was downloaded correctly.

NikoSaul:
@qdnguyen The file pnts load in the building like that: [here](https://github.com/AnalyticalGraphicsInc/3d-tiles-samples/tree/master/tilesets/TilesetWithRequestVolume) :)

qdnguyen:
Ok, thank, but please tell me where is its position on globe that you can visualize it? i did not see neither building nor points.

NikoSaul:
@qdnguyen With the 3dtiles.html you are in good position but you have to active on your left the Resquet volume ;) 

Else the position is:
 ```` 
const positionOnGlobe = { longitude: -75.65044586, latitude: 40.062840764, altitude: 100000 };

src/Renderer/Camera.js
NikoSaul: Ho yes my bad :-1:

qdnguyen:
Can we support a b3dm file with embeded resources such as texture in this pr? 

i got an error from gltf loader when i try to load my b3dm file with embeded texture.
GLTFLoader.js:980 Uncaught (in promise) TypeError: Cannot read property 'binary_glTF' of undefined
at GLTFLoader.js:980
at _each (GLTFLoader.js:592)
at GLTFLoader.js:978
at

qdnguyen:
B3dmLoader should be generic, it should support header length of 28,24,20 bytes. The same for gltfloader et gltf2loader.

Jeremy-Gaillard:
There's no such things as 20 or 24 bytes header for b3dm. [The specification](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel) indicates 28 bytes, if your files have smaller header length, they use an out-of-date version of the spec. I recommend updating your files rather than iTowns.

There is no way to differentiate between your out-of-date file and a correct one: the version attribute in the header is 1 regardless. 3d-tiles is still in active development, we must expect unversionned changed until the first release.

qdnguyen:
it's not about my file, it is what cesium used to generate 3dtiles example, ref https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/blob/master/generator/lib/createB3dm.js

In fact, almost b3dm convertor such as gltf2glb, gltf to b3dm still use a header of 20 bytes.

Otherwise, should we support glb in 3dtiles if b3dm header is not used? it's juste a minor modification in 3dtileprovider?

Jeremy-Gaillard:
The script you linked has the right header length. Even if it wasn't the case, I don't think we should try to support deprecated versions of b3dm in this early stage of development.

In any case, b3dm is not in the scope of this PR, so it will not be done here.

qdnguyen:
b3dm is a part of 3dtile, its loader should be improved in this pr, no? 

I'm beginer of b3dm format but i think the version (4 bytes) param can be used to do it?

Jeremy-Gaillard:
> b3dm is a part of 3dtile, its loader should be improved in this pr, no?

The 3d-tiles standard is far too large to fit in one PR. The scope of the PR is defined in @NikoSaul's opening comment.

> I'm beginer of b3dm format but i think the version (4 bytes) param can be used to do it?

This was what I was trying to explain in my previous comment. The 3d-tiles specification is not stable, and no "fixed" version has been released yet. The version byte is "1" and has never changed yet because of that.

NikoSaul:
@peppsac I think it's good now, if you can confirm :)

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,98 @@
...
+            if (pntsHeader.BTBinaryLength > 0) {
+                throw new Error('For pnts loader, BTBinaryLength: not yet managed');
+            }
+        } else {
+            throw new Error('Invalid pnts file.');
+        }
+    },
+};
+
+function parseFeatureBinary(array, byteOffset, FTJSONLength, view) {
```
peppsac: `view` and `array` are unneeded here, you can just pass the original buffer.

```js
const subArrayJson = textDecoder.decode(new Uint8Array(array.subarray(byteOffset, FTJSONLength + byteOffset))); 
```
becomes:
```js
const subArrayJson = textDecoder.decode(new Uint8Array(array, byteOffset, FTJSONLength));
```
and you can directly replace `view` with `buffer` below.

IMHO this improves lisibility.

src/Renderer/ThreeExtended/PntsLoader.js
```diff
@@ -0,0 +1,98 @@
+import * as THREE from 'three';
+
+const textDecoder = new TextDecoder('utf-8');
+export default {
+    parse: function parse(buffer) {
+        if (!buffer) {
+            throw new Error('No array buffer provided.');
+        }
+        const array = new Uint8Array(buffer);
```
peppsac: Same comment, remove array it's not useful.

peppsac (approved):

src/Renderer/Camera.js
```diff
@@ -77,43 +77,68 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
-const obbViewMatrix = new THREE.Matrix4();
+const localViewMatrix = new THREE.Matrix4();
 const tempMatrix = new THREE.Matrix4();
 const tempBox3d = new THREE.Box3();
-
-function _prepareBox3AndMatrix(box3d, matrixWorld, visibilityTestingOffset) {
+const tempSphere3d = new THREE.Sphere();
+function centerVolume(volumes3D) {
+    if (volumes3D.type === 'Sphere') {
+        temp = volumes3D.center.negate();
```
peppsac: `negate()` will modify the `volumes3D` center.

What about this proposal:
```js
// Frustum intersection are not reliable when big numbers are involved,
// so shift the volume toward the origin.
// To compensate for this modification, we'll also modify the matrixWorld
// of the object.
function shiftVolumeTowardOrigin(originalVolume, tempContainer) {
    if (originalVolume.center) {
        // THREE.Sphere objects have a .center property
        temp.copy(originalVolume.center);
    } else if (originalVolume.getCenter) {
        // THREE.Box objects have a .getCenter method
        originalVolume.getCenter(temp);
    } else {
         throw new Error(`Unsupported volume object ${volumes3D}`);
    }
    // temp is -center
    temp.negate();
    // shift the volumes3D toward origin
    tempContainer.copy(originalVolume);
    tempContainer.translate(temp);
}
```

Then use it llike this:
```js
shiftVolumeTowardOrigin(box3d, tempBox3d);
```
ou

```js
shiftVolumeTowardOrigin(sphere, tempSphere3d);
```

WDYT? (warning: I didn't test my proposal so it may be completely broken :))

src/Renderer/Camera.js
```diff
@@ -77,43 +77,68 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
+    localViewMatrix.multiplyMatrices(this._viewMatrix, tempMatrix);
+    frustum.setFromMatrix(localViewMatrix);
 
     return frustum.intersectsBox(tempBox3d);
 };
 
+Camera.prototype.isSphereVisible = function isSphereVisible(sphere, matrixWorld) {
+    sphere.type = 'Sphere';
+    centerVolume(sphere);
+    _prepareBox3AndMatrix(matrixWorld, this._visibilityTestingOffset);
```
peppsac: Forgot to remove the call to `_prepareBox3AndMatrix` ?

peppsac (changes requested):

peppsac (approved):
I believe we can now merge this PR.
Thanks for the hard work :)

As a side note regarding 3d-tiles: it'd be nice to work on a PR to transform the 3d-tiles.html examples into a simple functional test (see globe_test.js for instance).

NikoSaul:
Thank to you for take your time for this PR :)

PR #326, opened 2017-06-07
layers color's visibility and extent

gchoqueux:


autra:
Let's split that in several PRs to ease review please. At least the miniglobe and the example widget should have their own. The refacto commit about extent seems to have unrelated things in it. Let's put them in other commits/PRs. I guess it's ok to have all the one-line fixes in the same bugfix PR.

Making more PR can seem slower, but it's actually faster:
- the reviews are easier to do, and reviewers will be more likely to review (we won't need to plan 1/2h to do so)
- trivial changes are not blocked by more complex changes
- If we disagree on a particular point, the other PRs can still move on, instead of being blocked by an unresolved discussion

Thanks!

PR #327, opened 2017-06-07
layer colors's visibility and extent

gchoqueux:
Refacto and features to get layers color's visibility and extent 

* Refacto (WMTS/WMS) : all exent in options
* Feature (extent) : add functions `set` and `add`
* Add min and max option for WMS

### DIFF
--- a/examples/layersColorVisible.html
+++ b/examples/layersColorVisible.html
@@ -1,0 +1,80 @@
+<!-- Update menu with colors layers really displayed in viewer -->
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Color layers Visible</title>
+        <style type="text/css">
+            html {height: 100%}
+            body { margin: 0; overflow:hidden; height:100%}
+
+            #viewerDiv {
+                margin : auto auto;
+                width: 100%;
+                height: 100%;
+                padding: 0;
+            }
+            #menuDiv {position: absolute; top:0px; margin-left: 0px;}
+        </style>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="../node_modules/dat.gui/build/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <script src="GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script type="text/javascript">
+            /* global itowns,document,GuiTools*/
+
+            const positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };
+
+            // iTowns namespace defined here
+            const viewerDiv = document.getElementById('viewerDiv');
+            const menuGlobe = new GuiTools('menuDiv');
+            const globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            menuGlobe.view = globeView;
+
+            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('layers/JSONLayers/Region.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('layers/JSONLayers/IGN_MNT.json').then(result => globeView.addLayer(result));
+
+            function getLayersColorVisible(node, layers) {
+                if (!node || !node.visible) {
+                    return;
+                }
+                if (node.level) {
+                    // if node.material.visible is true then the node is diplayed in viewer
+                    // if node.material.colorLayersId is defined then the node have color layers
+                    if (node.material.visible && node.material.colorLayersId) {
+                        // get all node's color layers
+                        const tileColorLayers = node.material.colorLayersId;
+                        layers.id = [...new Set(layers.id.concat(tileColorLayers))];
+                    }
+                }
+                if (node.children) {
+                    for (const child of node.children) {
+                        getLayersColorVisible(child, layers);
+                    }
+                }
+            }
+
+            globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, () => {
+                // eslint-disable-next-line no-console
+                console.info('Globe initialized');
+                menuGlobe.addImageryLayersGUI(globeView.getLayers(l => l.type === 'color'));
+                menuGlobe.addElevationLayersGUI(globeView.getLayers(l => l.type === 'elevation'));
+            });
+
+            // Use preRender to update menu before rendering
+            globeView.preRender = () => {
+                const layers = { id: [] };
+                // globeView.scene is THREE.js scene
+                getLayersColorVisible(globeView.scene, layers);
+                const colorLayers = globeView.getLayers(l => l.type == 'color');
+
+                colorLayers.forEach(layer =>
+                    menuGlobe.hideFolder(layer.id, layers.id.indexOf(layer.id) == -1));
+            };
+        </script>
+    </body>
+</html>
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -6,6 +6,13 @@ import Coordinates, { crsToUnit, crsIsGeographic, assertCrsIsValid, convertValue
  * It can use explicit coordinates (e.g: lon/lat) or implicit (WMTS coordinates)
  */
 
+const CARDINAL = {
+    WEST: 0,
+    EAST: 1,
+    SOUTH: 2,
+    NORTH: 3,
+};
+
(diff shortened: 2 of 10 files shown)


src/Core/Geographic/Extent.js
```diff
@@ -263,4 +271,33 @@ Extent.prototype.intersect = function intersect(bbox) {
              this.north() <= other.south(this._internalStorageUnit));
 };
 
+Extent.prototype.set = function set(west, east, south, north) {
```
peppsac, line 274: Any reason why you're adding these 2 functions (they don't seem used)?

gchoqueux: It's so that the users can concatenate the extents (visible tiles for example)

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -69,7 +69,7 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
 };
 
 WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
-    return layer.extent.intersect(tile.extent);
+    return tile.level >= layer.options.zoom.min && layer.extent.intersect(tile.extent);
```
peppsac: You need to add the same test for `zoom.max`

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -37,7 +37,7 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     if (!layer.name) {
         throw new Error('layerName is required.');
     }
-    if (!layer.bbox) {
+    if (!layer.options && !layer.options.extent) {
```
peppsac: I think you want a `||` here.

peppsac: Also please fix the error message below: 'layer.options.extent property is required'.
Btw, it's a bit strange to move `extent` in `options` since it's a mandatory parameter.

IMHO, the `options` thing should go away, and all layer properties should live at the top level. What do you think?

src/Core/Geographic/Extent.js
```diff
@@ -263,4 +271,33 @@ Extent.prototype.intersect = function intersect(bbox) {
...
+    this._values[CARDINAL.WEST] = west;
+    this._values[CARDINAL.EAST] = east;
+    this._values[CARDINAL.SOUTH] = south;
+    this._values[CARDINAL.NORTH] = north;
+};
+
+Extent.prototype.add = function add(extent) {
```
peppsac: `add` isn't a good name IMHO. May I suggest `merge` ? or `concat` ?

peppsac: You should check that `extent.crs()` is the same as this.crs() and throw otherwise.

src/Core/Geographic/Extent.js
```diff
@@ -263,4 +271,33 @@ Extent.prototype.intersect = function intersect(bbox) {
...
+    const west = extent.west(this._internalStorageUnit);
+    if (west < this.west()) {
+        this._values[CARDINAL.WEST] = west;
+    }
+
+    const east = extent.east(this._internalStorageUnit);
+    if (east > this.east()) {
+        this._values[CARDINAL.EAST] = extent.east(this._internalStorageUnit);
```
peppsac: `... = east;`

src/Core/Geographic/Extent.js
peppsac: I'm a bit reluctant to add code that isn't used at least once in our code base.
(we can merge it anyway if you *know* it'll be useful)

examples/layersColorVisible.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Globe</title>
```
peppsac: `title`  is wrong. 
Maybe we should start adding a short comment at the top of our examples, to explain what they are about?

(also please fix indentation)

peppsac (changes requested):

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: yes, i move in top level

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: done

src/Core/Geographic/Extent.js
gchoqueux: done

src/Core/Geographic/Extent.js
gchoqueux: done

examples/layersColorVisible.html
gchoqueux: done

gchoqueux:
@peppsac branch updated

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -37,17 +37,21 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
...
     }
     if (!layer.projection) {
         throw new Error('projection is required');
     }
 
     layer.extent = new Extent(
         layer.projection,
-        layer.bbox[0], layer.bbox[1],
-        layer.bbox[2], layer.bbox[3]);
+        layer.extent.west, layer.extent.east,
```
peppsac: I think you can directly write: `layer.extent = new Extent(layer.projection, layer.extent);`

src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -52,6 +53,9 @@ WMTS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer
                 max: maxZoom,
             };
         }
+        if (layer.extent) {
```
peppsac: This `extent` is unused right? (since in WMTS we on TileMatrixSet)

src/Core/View.js
```diff
@@ -224,6 +229,9 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
 View.prototype.notifyChange = function notifyChange(needsRedraw, changeSource) {
     this._changeSources.add(changeSource);
     this.mainLoop.scheduleViewUpdate(this, needsRedraw);
+    if (needsRedraw) {
+        this.dispatchEvent({ type: VIEW_EVENTS.UPDATED });
```
peppsac: Can you document the purpose of this event?
And/or give example use case for it?

examples/layersColorVisible.html
```diff
@@ -0,0 +1,63 @@
+<!-- this example outputs in console, the array of colors layers really displayed in viewer -->
```
peppsac: IMHO examples should demo an interesting visual feature.
I don't think this one is worth it.

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: done

src/Core/View.js
gchoqueux: It's interesting to know when the viewer is updated to synchronize with other applications
See examples with `layersColorVisible.html `, that I modified to couple the viewer and the menu

examples/layersColorVisible.html
gchoqueux: This example allows a user to discover itowns' structure
I added some comments

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: removed

gchoqueux:
@peppsac updated

examples/layersColorVisible.html
peppsac: Sorry but I still don't think it's worth adding this to the examples.
But I agree the content is useful: it should be in the documentation.

src/Core/View.js
peppsac: 1. I'm wary of using events to *synchronize* things. Are you sure your listener will be called in a sync manner? Or will they be called after the update is performed? or after the rendering? or "it depends"?
2. The name and usage of your event disagree with each other:
```
itowns.GLOBE_VIEW_EVENTS.UPDATED, isDisplayed
```
(note that there's already a way to plug a function just before rendering: `view.preRender`)

gchoqueux: i remove itowns.GLOBE_VIEW_EVENTS.UPDATED and i use preRender

gchoqueux:
@peppsac  updated

peppsac (approved):
Ok thanks.

PR #329, opened 2017-06-07
Miniglobe

gchoqueux:
**Add feature to render : add alphaClearValue in viewer option**
alphaClearValue : set alpha value in renderer.setClearColor(color, alpha)

**example : add a mini globe prefab**
The mini-globe allows to have a farther away from the globe to quickly find position.

The mini-globe is handle of geometry Layer Globe, it displays only the level 2 tiles.
capture du 2017-06-07 17-57-59

**Add example scale widget in index.html**
This example shows how to display and refresh a widget scale on the screen

### DIFF
--- a/examples/globe.html
+++ b/examples/globe.html
@@ -20,6 +20,18 @@
                 padding: 0;
             }
 
+            #miniDiv {
+                display: block;
+                margin-bottom: 20px;
+                margin-right: 20px;
+                position: absolute;
+                width:100px;
+                height:100px;
+                left: 20;
+                bottom: 0;
+                color: white;
+            }
+
             #menuDiv {
                 position: absolute;
                 top:0px;
@@ -59,6 +71,8 @@
     <body>
         <div id="viewerDiv">
             <span class="divScaleWidget"> Scale </span>
+        </div>
+        <div id="miniDiv">
         </div>
         <script src="GUI/GuiTools.js"></script>
         <script src="../dist/itowns.js"></script>
--- a/examples/globe.html
+++ b/examples/globe.html
@@ -25,6 +25,25 @@
                 top:0px;
                 margin-left: 0px;
             }
+            .divScaleWidget {
+                border: 2px solid black;
+                border-top: none;
+                text-align: center;
+                display: block;
+                background-image: linear-gradient(rgba(200, 200, 200, 0.3), rgba(200, 200, 200, 0.3));
+                margin-bottom: 20px;
+                margin-right: 20px;
+                position: absolute;
+                width:200px;
+                height:18px;
+                color: black;
+                font-family: 'Open Sans',
+                sans-serif;
+                font-size: 16px;
+                right: 0;
+                bottom: 0;
+            }
+
 
             @media (max-width: 600px) {
                 #menuDiv {
@@ -38,7 +57,9 @@
         <script src="GUI/dat.gui/dat.gui.min.js"></script>
     </head>
     <body>
-        <div id="viewerDiv"></div>
+        <div id="viewerDiv">
+            <span class="divScaleWidget"> Scale </span>
+        </div>
         <script src="GUI/GuiTools.js"></script>
         <script src="../dist/itowns.js"></script>
         <script src="../dist/debug.js"></script>
@@ -51,6 +72,23 @@
             /* global itowns, document, GuiTools, globeView, promises */
             var menuGlobe = new GuiTools('menuDiv');
             menuGlobe.view = globeView;
+            var divScaleWidget = document.querySelectorAll('.divScaleWidget')[0];
+
+            function updateScaleWidget() {
+                var value = globeView.controls.pixelsToMeters(200);
+                value = Math.floor(value);
+                var digit = Math.pow(10, value.toString().length - 1);
+                value = Math.round(value / digit) * digit;
+                var pix = globeView.controls.metersToPixels(value);
+                var unit = 'm';
+                if (value >= 1000) {
+                    value /= 1000;
+                    unit = 'km';
+                }
+                divScaleWidget.innerHTML = `${value} ${unit}`;
+                divScaleWidget.style.width = `${pix}px`;
+            }
+
             // Listen for globe full initialisation event
             globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 // eslint-disable-next-line no-console
@@ -59,6 +97,10 @@
                     menuGlobe.addImageryLayersGUI(globeView.getLayers(function (l) { return l.type === 'color'; }));
                     menuGlobe.addElevationLayersGUI(globeView.getLayers(function (l) { return l.type === 'elevation'; }));
                 });
+                updateScaleWidget();
+            });
+            globeView.controls.addEventListener(itowns.CONTROL_EVENTS.RANGE_CHANGED, () => {
+                updateScaleWidget();
             });
         </script>
     </body>
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -3,18 +3,54 @@
 
 // Define initial camera position
 var positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };
+var promises = [];
+var miniView;
+var minDistance = 10000000;
+var maxDistance = 30000000;
 
 // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
 var viewerDiv = document.getElementById('viewerDiv');
+var miniDiv = document.getElementById('miniDiv');
 
 // Instanciate iTowns GlobeView*
 var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, { renderer: renderer });
-
-var promises = [];
-
 function addLayerCb(layer) {
     return globeView.addLayer(layer);
 }
+
+// Dont' instance mini viewer if it's Test env
+if (!renderer) {
+    miniView = new itowns.GlobeView(miniDiv, positionOnGlobe, {
+        // `limit globe' subdivision level:
+        // we're don't need a precise globe model
+        // since the mini globe will always be seen from a far point of view (see minDistance above)
+        maxSubdivisionLevel: 2,
+        // Don't instance default controls since miniview's camera will be synced
+        // on the main view's one (see globeView.onAfterRender)
+        noControls: true,
+    });
+
+    // Set a 0 alpha clear value (instead of the default '1')
+    // because we want a transparent background for the miniglobe view to be able
+    // to see the main view "behind"
+    miniView.mainLoop.gfxEngine.renderer.setClearColor(0x000000, 0);
+
+    // update miniview's camera with the globeView's camera position
+    globeView.onAfterRender = function onAfterRender() {
+        // clamp distance camera from globe
+        var distanceCamera = globeView.camera.camera3D.position.length();
+        var distance = Math.min(Math.max(distanceCamera * 1.5, minDistance), maxDistance);
+        var camera = miniView.camera.camera3D;
+        // Update target miniview's camera
+        camera.position.copy(globeView.controls.moveTarget()).setLength(distance);
+        camera.lookAt(globeView.controls.moveTarget());
+        miniView.notifyChange(true);
+    };
+
+    // Add one imagery layer to the miniview
+    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) { miniView.addLayer(layer); });
+}
+
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -239,7 +239,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     }
 
     this._renderState = RendererConstant.FINAL;
-
+    const renderer = this.mainLoop.gfxEngine.renderer;
     this.preRender = () => {
         const v = new THREE.Vector3();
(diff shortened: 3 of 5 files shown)


peppsac:
`feature (miniGlobe)` and `example (scale widget)` are not good commit messages.

Here's an example of what I expected to read instead:
```
feature: add a mini globe prefab

Mini-globe is a regularly asked feature so this commit adds some helper code to achieve this feature.

The mini-globe is implemented as a [...] because [...].
[ explain the how, or the choices made, or the missing feature, etc]
```

Also this PR lacks a screenshot :P

gchoqueux:
@peppsac Done

gchoqueux:
updated

examples/globe.html
```diff
@@ -51,6 +72,26 @@
...
+
+            function updateScaleWidget() {
+                var pitch = 0.28;
+                var size = 200 * pitch;
+                var scaled = globeView.controls.getScale();
+                var mm = size / scaled / 1000;
+                mm = Math.floor(mm);
+                var digit = Math.pow(10, mm.toString().length - 1);
+                var mmr = Math.round(mm / digit) * digit;
+                var pix = Math.floor(mmr * scaled / pitch * 1000);
```
peppsac: IHMO all this code indicates that `getScale` is a poor API.
Maybe you could improve it based on this experience, e.g:
```js
var m = globeView.controls.pixelsToSI(200);
// m.value = 1.23 and m.unit = 'km'
```

src/Core/Prefab/Globe/MiniGlobe.js
```diff
@@ -0,0 +1,86 @@
...
+    const pointMesh = new THREE.Mesh(point, new THREE.MeshBasicMaterial({ color: 0xffffff, side: THREE.DoubleSide, opacity: 0.85, transparent: true }));
+    targetMesh.add(pointMesh);
+    targetMesh.position.set(0, 0, -500);
+    shadowMesh.translateZ(-10);
+    targetMesh.add(shadowMesh);
+    targetMesh.layers.set(layer);
+    shadowMesh.layers.set(layer);
+    pointMesh.layers.set(layer);
+
+    return targetMesh;
```
peppsac: Is this mesh supposed to indicate on the mini-globe the visible area in the big one? 
Because here it doesn't really match (the area displayed in the big globe is bigger than the area indicated by this mesh)

peppsac (commented):
Did you try to implement this feature without modifying itowns first?

I believe there's everything needed already: instanciate 2 GlobeView with different subdivision and layer settings, each one rendering in its own `div`.

src/Core/Prefab/Globe/MiniGlobe.js
gchoqueux: it's just target, it shows camera's target

examples/globe.js
```diff
@@ -3,26 +3,52 @@
...
 var positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };
+var promises = [];
 
 // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
 var viewerDiv = document.getElementById('viewerDiv');
+var miniDiv = document.getElementById('miniDiv');
 
 // Instanciate iTowns GlobeView*
 var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, { renderer: renderer });
+var miniView = new itowns.GlobeView(miniDiv, positionOnGlobe, {
```
peppsac: Cound you add comments to your changes so this file can serve as documentation/tutorial?

examples/globe.js
```diff
@@ -3,26 +3,52 @@
...
+    camera.position.copy(globeView.controls.moveTarget()).setLength(distance);
+    camera.lookAt(globeView.controls.moveTarget());
+    miniView.notifyChange(true);
+};
 
 function addLayerCb(layer) {
     return globeView.addLayer(layer);
 }
+
+function addLayerCbM(layer) {
```
peppsac: Suggestion: maybe than adding a duplicate function we could use the following syntax instead:
```js
promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(globeView.addLayer.bind(globeView)));
```
WDYT?

examples/globe.js
```diff
@@ -3,26 +3,52 @@
...
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
 promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCbM));
 // Add two elevation layers.
 // These will deform iTowns globe geometry to represent terrain elevation.
 promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
 promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
 
+globeView.miniview = miniView;
```
peppsac: If it's not used in the `globe_test.js` file then you don't need to export this.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1651,6 +1651,38 @@ GlobeControls.prototype.getScale = function getScale(pitch) {
...
+ * @property {string} unit distance unit
+ */
+
+/**
+ * Convert pixel in length (meter or kilometer)
+ * @param      {number}  pixel   The pixel
+ * @param      {number}  pitch   The pitch
+ * @return     {Length} length
+ */
+GlobeControls.prototype.pixelsToMeter = function getScale(pixel, pitch) {
```
peppsac: Use a all plural form: `pixelsToMeters`

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1651,6 +1651,38 @@ GlobeControls.prototype.getScale = function getScale(pitch) {
...
+ * @typedef {Object} Length
+ * @property {number} value distance value
+ * @property {string} unit distance unit
+ */
+
+/**
+ * Convert pixel in length (meter or kilometer)
+ * @param      {number}  pixel   The pixel
+ * @param      {number}  pitch   The pitch
+ * @return     {Length} length
```
peppsac: This documentation is wrong (`meter or kilometer`) and a typical example of why structured / linted documentation easily leads to useless documentation (ping issue #314).
Sure all fields are documented, have their type specified but how can `@param      {number}  pixel   The pixel`  be useful to anyone? 
Or `The pitch`??

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1651,6 +1651,38 @@ GlobeControls.prototype.getScale = function getScale(pitch) {
...
+    return size / scaled / 1000;
+};
+
+/**
+ * convert length in meter in pixel
+ * @param      {number}  value   length in meter
+ * @param      {number}  pitch   The pitch
+ * @return     {number}  pixel
+ */
+GlobeControls.prototype.meterTopixels = function getScale(value, pitch) {
```
peppsac: Same comments apply: plural + documentation.

src/Core/Prefab/GlobeView.js
```diff
@@ -250,9 +250,9 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
         if (len < lim) {
             var t = Math.pow(Math.cos((lim - len) / (lim - v.x * size * 0.9981) * Math.PI * 0.5), 1.5);
             var color = new THREE.Color(0x93d5f8);
-            this.mainLoop.gfxEngine.renderer.setClearColor(color.multiplyScalar(1.0 - t));
+            this.mainLoop.gfxEngine.renderer.setClearColor(color.multiplyScalar(1.0 - t), options.alphaClearValue);
```
peppsac: This is a new feature in a `add example`commit. I should live in a separate PR or at the minimum in a separate commit.

src/Renderer/ThreeExtended/GlobeControls.js
peppsac: And as with `alphaClearColor`: "This is a new feature in a add examplecommit. I should live in a separate PR or at the minimum in a separate commit."

peppsac (changes requested):
Looking way better (code-wise).
Still a few things to adjust please.

examples/globe.html
```diff
@@ -51,6 +72,23 @@
             /* global itowns, document, GuiTools, globeView, promises */
             var menuGlobe = new GuiTools('menuDiv');
             menuGlobe.view = globeView;
+            var divScaleWidget = document.querySelectorAll('.divScaleWidget')[0];
+
+            function updateScaleWidget() {
+                var value = globeView.controls.pixelsToMeter(200);
```
peppsac: pixelsToMeter -> pixelsToMeter**s**

examples/globe.html
```diff
@@ -51,6 +72,23 @@
...
+                value = Math.floor(value);
+                var digit = Math.pow(10, value.toString().length - 1);
+                value = Math.round(value / digit) * digit;
+                var pix = globeView.controls.meterTopixels(value);
```
peppsac: meter**s**To**P**ixels

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1651,6 +1651,31 @@ GlobeControls.prototype.getScale = function getScale(pitch) {
...
+    return size / scaled / 1000;
+};
+
+/**
+ * Projection on screen in pixels of length in meter on globe
+ * @param      {number}  value Length in meter on globe
+ * @param      {number}  pitch Screen pitch, in millimeters
+ * @return     {number}  projection in pixels on screen
+ */
+GlobeControls.prototype.metersTopixels = function metersTopixels(value, pitch) {
```
peppsac: metersTo**P**ixels

peppsac: While fixing the typo, you could also extract this feature to a specific commit.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1651,6 +1651,31 @@ GlobeControls.prototype.getScale = function getScale(pitch) {
 };
 
 /**
+ * To convert the projection in meters on the globe of a number of pixels of screen
+ * @param      {number} pixels count pixels to project
+ * @param      {number} pitch Screen pitch, in millimeters
```
peppsac: Could you rename this paramter `pixelPitch` and expand its documentation to explain the default value (`pixel pitch in millimeters (default = 0.28 mm)`

examples/globe.js
```diff
@@ -3,18 +3,48 @@
...
 // Instanciate iTowns GlobeView*
 var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, { renderer: renderer });
-
-var promises = [];
-
 function addLayerCb(layer) {
     return globeView.addLayer(layer);
 }
+
+if (!renderer) {
```
peppsac: Is this really needed?
If so, maybe add a comment explaining why you do this (I suppose you want to detect if you're in a test env or not?)

examples/globe.js
```diff
@@ -3,18 +3,48 @@
...
+    // Instanciate new Globe viewer to add mini view
+    // maxSubdivisionLevel limits display level to zoom/level at 2
+    // alphaClearValue at 0 sets opacity rendering to totaly transparent
+    miniView = new itowns.GlobeView(miniDiv, positionOnGlobe, {
+        renderer: renderer,
```
peppsac: No you don't want to reuse the same renderer, because you want a new *Canvas* being inserted in `miniDiv`

examples/globe.js
```diff
@@ -3,18 +3,48 @@
...
-var promises = [];
-
 function addLayerCb(layer) {
     return globeView.addLayer(layer);
 }
+
+if (!renderer) {
+    // Instanciate new Globe viewer to add mini view
+    // maxSubdivisionLevel limits display level to zoom/level at 2
+    // alphaClearValue at 0 sets opacity rendering to totaly transparent
```
peppsac: The 2 lines could be replaced by something explaining the intent for each parameter:

```js
miniView = new itowns.GlobeView(miniDiv, positionOnGlobe, {
        renderer: renderer,
        // `limit globe' subdivision level: we're don't need a precise globe model since the mini globe will always be seen from a far point of view (see minDistance above)
        maxSubdivisionLevel: 2,
        // don't create default controls since miniview's camera will be synced on the main view's one
        noControls: true,
        // force a 0 alpha clear value (instead of the default '1') because we want a transparent background for the miniglobe view to be able to see the main view "behind"
       alphaClearValue: 0
    });
```

src/Renderer/c3DEngine.js
```diff
@@ -18,6 +18,9 @@ function c3DEngine(rendererOrDiv, options = {}) {
     if (options.alpha === undefined) {
         options.alpha = true;
     }
+    if (options.alphaClearValue === undefined) {
```
peppsac: Beware: `c3DEngine` constructor receive `options.renderer` object as `options`.

In this PR you sometimes write in `options.renderer.alphaClearValue` (like now) and sometimes in `options.alphaClearValue` (like in `GlobeView.preRender`).

Last but not least: I *think* we should reserve `options` for fixed options = stuff that you can't change dynamically.
For this `alphaClearValue` I *believe* you maybe doesn't even need this option.
Here's an alternative implementation:
-  this.renderer.setClearColor(0x030508, 0); // <- do this once when creating the miniglobe
- change `setClearColor`calls to: `this.renderer.setClearColor(0x030508, renderer.getClearAlpha()); `

gchoqueux:
Thank you for these tips
Sorry I squash commits

peppsac (approved):
Ok, looks good now. Thanks!

PR #332, opened 2017-06-08
fix: Check elevation values before replacing parent : LayeredMaterialNodeProcessing

nosy-b:
This small fix allows for better handling of multiple DTM, it answers #192.

- WMTS can send some tiles with no real value in it through a code like -99999. We check this before replacing the elevation parent tile so we don't create holes and can keep other less precise DTM tile, typically around their borders
- This PR also modifies the index to handle a world DTM and a more precise for France.

It is not elegant at all, it should be handled with LayerState for example with a state specifying when an elevation tile is its max level.

The same should be done for Imagery as sometimes WMTS send white images, not errors.

### DIFF
--- a/examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
+++ b/examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
@@ -7,7 +7,7 @@
 	"updateStrategy": {
 		"type": 1,
 		"options": {
-			"groups": [3, 7, 11, 14]
+			"groups": [11, 14]
 		}
 	},
 	"options": {
@@ -19,36 +19,6 @@
 		"mimetype": "image/x-bil;bits=32",
 		"tileMatrixSet": "WGS84G",
 		"tileMatrixSetLimits": {
-			"6": {
-				"minTileRow": 13,
-				"maxTileRow": 36,
-				"minTileCol": 62,
-				"maxTileCol": 80
-			},
-			"7": {
-				"minTileRow": 27,
-				"maxTileRow": 79,
-				"minTileCol": 84,
-				"maxTileCol": 167
-			},
-			"8": {
-				"minTileRow": 55,
-				"maxTileRow": 146,
-				"minTileCol": 248,
-				"maxTileCol": 320
-			},
-			"9": {
-				"minTileRow": 110,
-				"maxTileRow": 292,
-				"minTileCol": 497,
-				"maxTileCol": 640
-			},
-			"10": {
-				"minTileRow": 221,
-				"maxTileRow": 585,
-				"minTileCol": 994,
-				"maxTileCol": 1281
-			},
 			"11": {
 				"minTileRow": 442,
 				"maxTileRow": 1267,
--- a/examples/layers/JSONLayers/WORLD_DTM.json
+++ b/examples/layers/JSONLayers/WORLD_DTM.json
@@ -1,0 +1,69 @@
+
+{
+    "type":       "elevation",
+    "protocol":   "wmts",
+    "id":         "MNT_WORLD_SRTM3",
+    "url":        "http://wxs.ign.fr/va5orxd0pgzvq3jxutqfuy0b/geoportail/wmts",
+    "noDataValue": -99999,
+    "updateStrategy": {
+        "type": 1,
+        "options": {
+            "groups": [3, 7, 9]
+        }
+    },
+    "options": {
+        "name": "ELEVATION.ELEVATIONGRIDCOVERAGE.SRTM3",
+        "mimetype": "image/x-bil;bits=32",
+        "tileMatrixSet": "WGS84G",
+        "tileMatrixSetLimits": {
+            "3": {
+                "minTileRow": 1,
+                "maxTileRow": 6,
+                "minTileCol": 0,
+                "maxTileCol": 16
+            },
+            "4": {
+                "minTileRow": 2,
+                "maxTileRow": 12,
+                "minTileCol": 0,
+                "maxTileCol": 32
+            },
+            "5": {
+                "minTileRow": 5,
+                "maxTileRow": 25,
+                "minTileCol": 0,
+                "maxTileCol": 64
+            },
+            "6": {
+                "minTileRow": 10,
+                "maxTileRow": 51,
+                "minTileCol": 0,
+                "maxTileCol": 128
+            },
+            "7": {
+                "minTileRow": 20,
+                "maxTileRow": 103,
+                "minTileCol": 0,
+                "maxTileCol": 256
+            },
+            "8": {
+                "minTileRow": 41,
+                "maxTileRow": 207,
+                "minTileCol": 0,
+                "maxTileCol": 512
+            },
+            "9": {
+                "minTileRow": 82,
+                "maxTileRow": 415,
+                "minTileCol": 0,
+                "maxTileCol": 1024
+            },
+            "10": {
+                "minTileRow": 164,
+                "maxTileRow": 830,
+                "minTileCol": 0,
+                "maxTileCol": 2048
+            }
+        }
+    }
+}
--- a/index.html
+++ b/index.html
@@ -48,7 +48,7 @@ and open the template in the editor.
             itowns.Fetcher.json('examples/layers/JSONLayers/ScanEX.json').then(result => globeView.addLayer(result));
             itowns.Fetcher.json('examples/layers/JSONLayers/Region.json').then(result => globeView.addLayer(result));
 
-            itowns.Fetcher.json('examples/layers/JSONLayers/IGN_MNT.json').then(result => globeView.addLayer(result));
+            itowns.Fetcher.json('examples/layers/JSONLayers/WORLD_DTM.json').then(result => globeView.addLayer(result));
             itowns.Fetcher.json('examples/layers/JSONLayers/IGN_MNT_HIGHRES.json').then(result => globeView.addLayer(result));
 
             menuGlobe.addGUI('RealisticLighting', false,
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -68,6 +68,32 @@ function initNodeElevationTextureFromParent(node, parent, layer) {
     }
 }
 
+function getIndiceWithPitch(i, pitch, w) {
+    // Return corresponding indice in parent tile using pitch
+    const currentX = (i % w) / w;  // normalized
+    const currentY = Math.floor(i / w) / w; // normalized
+    const newX = pitch.x + currentX * pitch.z;
+    const newY = pitch.y + currentY * pitch.z;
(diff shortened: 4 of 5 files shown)


index.html
```diff
@@ -47,8 +47,8 @@
             itowns.Fetcher.json('examples/layers/JSONLayers/OrthosCRS.json').then(result => globeView.addLayer(result));
             itowns.Fetcher.json('examples/layers/JSONLayers/ScanEX.json').then(result => globeView.addLayer(result));
             itowns.Fetcher.json('examples/layers/JSONLayers/Region.json').then(result => globeView.addLayer(result));
-
-            itowns.Fetcher.json('examples/layers/JSONLayers/IGN_MNT.json').then(result => globeView.addLayer(result));
+
```
peppsac: Whitespace change

src/Core/Layer/LayerUpdateState.js
```diff
@@ -3,6 +3,7 @@ const UPDATE_STATE = {
     PENDING: 1,
     ERROR: 2,
     DEFINITIVE_ERROR: 3,
+    DEEPEST_LEVEL: 4,
```
peppsac: This doesn't seem to be used anywhere?

src/Core/Scheduler/Providers/IoDriver_XBIL.js
```diff
@@ -49,10 +49,10 @@ IoDriver_XBIL.prototype.computeMinMaxElevation = function computeMinMaxElevation
             }
         }
     }
-
     if (max === -1000000 || min === 1000000) {
         return { min: undefined, max: undefined };
     }
+
```
peppsac: Unneeded whitespace changes

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -236,11 +236,14 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
...
     }
 
+
     if (node.layerUpdateState[layer.id] === undefined) {
         node.layerUpdateState[layer.id] = new LayerUpdateState();
         initNodeElevationTextureFromParent(node, node.parent, layer);
         currentElevation = material.getElevationLayerLevel();
     }
+
+
```
peppsac: Whitespaces...

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -281,6 +284,8 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
             }
 
             node.layerUpdateState[layer.id].success();
+            // eslint-disable-next-line no-console
+            // if (terrain.texture.image.data[0] < 0) console.log('AAAAAAAAAAAAAAAAAAAFCefzfzefzefezfezfzefzef');
```
peppsac: Maybe you should remove this :)

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -289,7 +294,7 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
                 terrain.texture.flipY = false;
                 terrain.texture.needsUpdate = true;
             }
-            node.setTextureElevation(terrain);
+            if (terrain.texture.image.data[0] && terrain.texture.image.data[terrain.texture.image.data.length - 1] > -99999) node.setTextureElevation(terrain);
```
peppsac: Hmm I'm not really a fan of this proposal. I've seen some place where data[0] was valid but data[15] wasn't for instance.

Also have you considered this alternative: allow several textures layers?
They wouldn't blend their values like colors layers but instead the first elevation texture with a valid value would be used.
So in your case, you would declare `IGN_MNT_HIGHRES` as the first elevation layer, and `MNT_WORLD_SRTM3` as the second one and it would act as a fallback when `HIGH_RES` gives no value.

WDYT?

peppsac (commented):
Thanks for taking care of this issue.
I've left a few minor style comment and a general design one :)

src/Process/LayeredMaterialNodeProcessing.js
autra: Oh nice one. Taking this to replace my `coucou` or `wtf` ;-)

src/Process/LayeredMaterialNodeProcessing.js
autra: (nits, formatting: this line is really long, prefer the
```js
if (...) {
    ....
}
``` 
form)

examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
```diff
@@ -7,7 +7,7 @@
 	"updateStrategy": {
 		"type": 1,
 		"options": {
-			"groups": [3, 7, 11, 14]
+			"groups": [11, 14]
```
autra, line 10: Just  a little question: why `IGN_MNT_HIGHRES.json` is modified at all? Let's either keep it as it is if we plan on use it, or let's remove it entirely.

src/Process/LayeredMaterialNodeProcessing.js
nosy-b: The test is not complete, I wanted to test diagonal extremity pixels. More like:
```
let terrainData = terrain.texture.image.data;
if (terrainData[0] > -99999 && terrainData[terrainData.length - 1] > -99999) 
        node.setTextureElevation(terrain);
```

It's a quick fix to avoid using non-real elevation tiles value that create holes.
As you asked @autra the IGN_MNT_HIGHRES DTM is complementary with the WORLD_DTM now that replaces the low res IGN_MNT (We should rename all MNT with DTM also).
So when declaring them, their TileMatrixSet have an empty union. I should try to see how it works when we don't take care of that.

@peppsac Yes I think it should be done as we do for color layers, with an order sequence for elevation layers. To avoid overloading the number of textures, the active elevation layer choice should be done before sending elevation to the shader.

src/Core/Layer/LayerUpdateState.js
nosy-b: Yes it comes from a test, I should remove it

src/Process/LayeredMaterialNodeProcessing.js
nosy-b: _+_ This fix goes with a rapid V2 release and can be kept in the future. 
_+_ Very few users will mix many DTM on the same region
 _-_   It will certainly need a more strong and generic approach as we mentioned above (2.1?)

peppsac: Can you update the PR with the whitespaces fixes plus:
- a test of the 4 corners of the terrainData
- a comment stating that this is a compromise and that we should implement multi elevation layer instead
- make sure this works with texture images (see planar example)

With these changes this PR should be mergeable.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -92,6 +92,16 @@ function refinementCommandCancellationFn(cmd) {
     return !cmd.requester.isDisplayed();
 }
 
+function checkNodeElevationTextureValidity(texture) {
+    // We check if the elevation texture has some significant values through corners
+    const tData = texture.image.data;
+    const l = tData.length;
+    return tData[0] > -99999 &&
```
gchoqueux: use `noDataValue` in layer option

examples/layers/JSONLayers/WORLD_DTM.json
```diff
@@ -0,0 +1,69 @@
...
+{
+    "type":       "elevation",
+    "protocol":   "wmts",
+    "id":         "MNT_WORLD_SRTM3",
+    "url":        "http://wxs.ign.fr/va5orxd0pgzvq3jxutqfuy0b/geoportail/wmts",
+    "noDataValue": -99999,
+    "updateStrategy": {
+        "type": 1,
+        "options": {
+            "groups": [3, 7, 10]
```
gchoqueux: may be use 
for WORLD
`"groups": [3, 7]`
and for HIGH RES
`"groups": [10, 14]`

gchoqueux (dismissed):

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -281,15 +291,17 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
             }
 
             node.layerUpdateState[layer.id].success();
-
```
peppsac: Nit: unneeded change

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -281,15 +291,17 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
...
             if (terrain.texture && terrain.texture.flipY) {
                 // DataTexture default to false, so make sure other Texture types
                 // do the same (eg image texture)
                 // See UV construction for more details
                 terrain.texture.flipY = false;
                 terrain.texture.needsUpdate = true;
             }
-            node.setTextureElevation(terrain);
+            if (!terrain.texture.image.data || checkNodeElevationTextureValidity(terrain.texture, layer.noDataValue))
```
peppsac: Maybe `checkXbilTextureValidity` would be more appropriate?

And what about calling: 
```js
node.layerUpdateState[layer.id].failure(Date.now(), true);
```
to indicate that this `layer` will never supply a valid texture for this node?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -281,15 +291,17 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
...
+                // Quick check to avoid using elevation texture with no significant values
+                // We should later implement multi elevation layer to choose the one to use at each level
+                { node.setTextureElevation(terrain); }
```
peppsac: Style issue:
```js
if (...) {
  // Quick check ...
  // ...
  node.setTextureElevation(terrain);
}
```

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -92,6 +92,16 @@ function refinementCommandCancellationFn(cmd) {
     return !cmd.requester.isDisplayed();
 }
 
+function checkNodeElevationTextureValidity(texture, noDataValue) {
+    // We check if the elevation texture has some significant values through corners
+    const tData = texture.image.data;
+    const l = tData.length;
+    return tData[0] > noDataValue &&
+           tData[l - 1] > noDataValue &&
+           tData[Math.sqrt(l) - 1] > noDataValue &&
```
peppsac, line 127: I'd prefer `tData[width - 1]` and `tData[height * width]` instead of `Math.sqrt` usage but it's up to you

nosy-b:
Ok, I changed a bit to handle more cleanly those no data value containing tiles. 
They now use parent values for the pixel with no-data-values when we detect that it's not a normal tile so it doesn"t slow down the rendering.
I 'll rebase

peppsac (approved):
Seems good.
Can you rebase this PR please?

nosy-b:
@peppsac rebased, thx.

PR #333, opened 2017-06-12
refactor: replace RTC feature with matrix manipulation

peppsac:
~This commit replaces the R(elative)T(o)C(enter) feature, with a simpler version
based on three.js callbacks and matrices manipulations.~

~RTC is useful when manipulating big coordinates, but instead of computing a matrix
per-object cancelling this object translation, we instead cancel the camera position
value from all matrixWorld.~

~Pro:~
- ~automatic: you only need to call `patchObjectForRTC` once and you're done (instead of
  updating the RTC matrix each frame)~
- ~reduces the amount of uniforms: before we used a bool (useRTC) and a mat4~

~Cons:~
- ~prevent user to use onBeforeRender/onAfterRender (but this could be mitigated,
  I didn't implement it because it's not worth to add complexity now)~
- ~all the objects must be patched or they won't get rendered. This one is annoying,
  but the API could be adapted (for instance: add an itowns method to add an object
  to the scene that would do the job for the user. Note that the original scene.add
  method would still be here for advanced users)~

~It also increments three.js version because starting from v0.85, `onBeforeRender`
callbacks are called before computing camera.matrixWorldInverse/obj.modelViewMatrix)~

~WDYT?~

Edit to match latest version of the code:

Starting from version 0.85 three.js uses double for matrices elements.
As a consequence we don't need to bother with RTC feature at all.

Note that we could have done this earlier, if instead of patching matrixWorld
we had also patched matrixWorldInverse...


Jeremy-Gaillard:
I like the idea a lot. Since we no longer need the RTC uniforms, does this mean we can use Three's default shaders transparently?

I think we still need a way to support RTC because there are so much resources that rely on it (cesium's 3d-tiles generator for example) that we want to be able to view. Maybe you could change the onBeforeRender function to detect if the object to be rendered has an RTC attribute and change the matrix computation in this case?

peppsac:
I'm not a 3d-tiles expert but I believe you can support the `rtcCenter` attribute by adding an intermediate object:
```js
var mesh = ...; // create mesh from 3d-tiles
var rtc = new THREE.Object3D();
rtc.position.copy(rtcCenter);
mesh.add(rtc);
scene.add(rtc);
```
Wouldn't this be enough?

Jeremy-Gaillard:
Yeah you're right, it should work if we prepare the nodes correctly.

I have no further objections, this PR looks good to me.

peppsac:
Branch updated.

peppsac:
Pushed a new version that will work transparently for all objects (no need to call `patchObjectForRTC`).

@Jeremy-Gaillard (and others) can you test and confirm it works with your setup?

Jeremy-Gaillard:
The 3d-tiles example is broken. The debug layer generates errors ("updateMatrixWorld is not a function"), and even if I remove it, the dragon model does not appear, despite being loaded.

peppsac:
Update the branch.
Please review, this is an important feature of itowns...

src/Renderer/Shader/TileVS.glsl
```diff
@@ -73,7 +71,7 @@ void main() {
             vPosition = vec4( position ,1.0 );
         }
 
-        mat4 projModelViewMatrix = useRTC ? mVPMatRTC : projectionMatrix * modelViewMatrix;
+        mat4 projModelViewMatrix = projectionMatrix * modelViewMatrix;
```
nosy-b: No more need for `mat4 projModelViewMatrix`.
Compute directly gl_Position as usual:  `gl_Position = projectionMatrix * modelViewMatrix * vPosition;`

src/Renderer/Shader/TileDepthVS.glsl
```diff
@@ -22,7 +24,8 @@ void main() {
             vPosition = vec4(position, 1.0);
         }
 
-        gl_Position = mVPMatRTC * vPosition;
+        mat4 projModelViewMatrix = projectionMatrix * modelViewMatrix;
```
nosy-b: No more need for mat4 projModelViewMatrix.
Compute directly gl_Position as usual: gl_Position = projectionMatrix * modelViewMatrix * vPosition;

src/Renderer/Camera.js
```diff
@@ -4,26 +4,74 @@
  * Description: La camera scene, interface avec la camera du 3DEngine.
```
nosy-b, line 4: In english

package.json
```diff
@@ -36,7 +36,7 @@
     "js-priority-queue": "^0.1.5",
     "jszip": "^3.1.3",
     "proj4": "^2.4.3",
-    "three": "^0.84.0",
+    "three": "^0.85.0",
```
nosy-b: If updating THREE version, we should go to the last release: r86

peppsac: I can give it a try and see if it works correctly; r86 was released yesterday :)

src/Renderer/Camera.js
peppsac: Sure this needs fixing. Except it has nothing to do with this PR :)

src/Renderer/Shader/TileDepthVS.glsl
peppsac: OK, I don't mind using the one-line version, I'll update the PR.

peppsac:
Branch updated, please r?

src/Renderer/Camera.js
```diff
@@ -4,24 +4,13 @@
...
 
     this.camera3D = options.camera ? options.camera : new THREE.PerspectiveCamera(30, this.ratio);
 
-
-    if (!options.camera) {
-        // do not modify a camera we don't own fully
-        // /!\ WARNING Matrix JS are in Float32Array
-        this.camera3D.matrixWorld.elements = new Float64Array(16);
-        this.camera3D.matrixAutoUpdate = false;
-        this.camera3D.rotationAutoUpdate = false;
```
autra, line 22: It's now the client's responsibility to do that if he/she wants to, that's it?

nosy-b:
@Jeremy-Gaillard @qdnguyen, is it working fine with 3D-tiles?

autra (approved):

Jeremy-Gaillard:
Everything looks good for 3d-tiles.

src/Renderer/Camera.js
peppsac: No.
The first line is useless now that three.js matrices use double.
The 2nd/3rd lines are optimization that bring no performance gain but bring troubles: the user have to remember to call `camera.updateMatrix`, otherwise the camera will never move.

PR #336, opened 2017-06-16
Fix: correct style for test + timeout

autra:


autra:
@iTowns/reviewers This fixes our build. please r?

nosy-b:
_continuous-integration/travis-ci/push — The Travis CI build failed_
?

autra:
The log would tell you it's not the same issue ;-) I'm fixing the other one as well anyway.

peppsac (approved):
Thanks

tbroyer:
@autra `lts/*` is as correct as `--lts`: https://github.com/creationix/nvm#long-term-support (actually _more_ correct, as `--lts` is an `nvm install` argument, whereas `lts/*` is the actual version specifier)

Can you point to a Travis CI build log that failed because of `lts/*` vs `--lts`?

autra:
@tbroyer https://travis-ci.org/iTowns/itowns2/jobs/242006919

It seems to be a recent change somewhere.

autra:
That being said, @tbroyer, apparently I don't reproduce in every PRs

tbroyer:
@autra It was a network error; I restarted that job and it correctly picked Node 6, without error.

autra:
Ok, perfect then! Only the timeout change is needed in this then. Amending.

peppsac:
@tbroyer where did you see it was a network error?

And how did you restart the job?

Thanks :)

peppsac:
So it seems increasing the timeout even more is needed.
I'm not really happy with this change but I'm even less happy to block all PR.

autra:
@peppsac the alternative is to disable the test for now... But if 10sec on jenkins is acceptable from a user POV for this test, then I guess we can merge.

peppsac:
For now we can merge the 10s timeout, assuming that tests are slow on Travis.
If a timeout occurs on a developper machine we'll have to deep digger and find the root issue.
So yes please merge :)

tbroyer:
> @tbroyer where did you see it was a network error?

The error message was saying nvm couldn't download the `lts/*` version (or something like that) and that it could be due to the network. Given that restarting the job succeeded, I concluded that it indeed was a network error.

> And how did you restart the job?

There's a “restart job” button on the top-right hand of the job page, or on the right end of a job line in the build page (and there's a “restart build” button on the top-right hand).

tbroyer:
> For now we can merge the 10s timeout, assuming that tests are slow on Travis.

How about using a huge timeout only on Travis? (check for [environment variables](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables), e.g. `CI=true`) or maybe even disabling that specific test on Travis?

peppsac:
> The error message was saying nvm couldn't download [...]

Missed this :)

> There's a “restart job” button on the top-right  [...]

I'm not logged in Travis, that may explain why I don't see it.

> How about using a huge timeout only on Travis? (check for environment variables, e.g. CI=true) 

Why not - though 10s is already really big.

> or maybe even disabling that specific test on Travis?

Nope.
Let's not start disabling *some* tests: IMHO a test is either reliable (= working everywhere) or removed :)

PR #337, opened 2017-06-19
feature: improve isBox3DVisible and box3DSizeOnScreen (reduce mem. al…

peppsac:
…loc)

Rewrite these 2 functions to avoid temporary values.


autra (approved):
LGTM

PR #338, opened 2017-06-19
Refactor planar sse

peppsac:
Planar mode sse used a simplistic computation until now.

This PR proposes to change it to a visibilty based one: we subdivide based one the on-screen size of each tile.

WDYT?

### DIFF
--- a/src/Core/Prefab/PlanarView.js
+++ b/src/Core/Prefab/PlanarView.js
@@ -43,7 +43,6 @@ function PlanarView(viewerDiv, boundingbox, options) {
             node.material.wireframe = layer.wireframe || false;
         }
     };
-    const SSE_SUBDIVISION_THRESHOLD = 6.0;
 
     const tileLayer = new GeometryLayer('planar');
     const initLayer = initTiledGeometryLayer(planarSchemeTile(boundingbox));
@@ -101,7 +100,7 @@ function PlanarView(viewerDiv, boundingbox, options) {
     tileLayer.update =
         processTiledGeometryNode(
             planarCulling,
-            planarSubdivisionControl(4, SSE_SUBDIVISION_THRESHOLD),
+            planarSubdivisionControl(options.maxSubdivisionLevel || 5),
             nodeInitFn);
     tileLayer.builder = new PlanarTileBuilder();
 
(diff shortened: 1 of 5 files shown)


src/Process/PlanarTileProcessing.js
```diff
@@ -9,47 +8,24 @@ export function planarCulling(node, camera) {
...
+    const dim = {
+        x: 0.5 * (onScreen.max.x - onScreen.min.x),
+        y: 0.5 * (onScreen.max.y - onScreen.min.y),
+    };
 
-    if (ratio >= 0.25) {
-        return 7;
-    }
-    return 1;
+    return (dim.x > 0.3 && dim.y >= 0.3);
```
autra: Can you elaborate on the magic constants (0.3 here and the multiplication by 0.5 above) and about the `>` for x and `>=` for y?

`0.3` is the equivalent of the old `threshold`, isn't it?

src/Core/Prefab/PlanarView.js
```diff
@@ -101,7 +100,7 @@ function PlanarView(viewerDiv, boundingbox, options) {
     tileLayer.update =
         processTiledGeometryNode(
             planarCulling,
-            planarSubdivisionControl(4, SSE_SUBDIVISION_THRESHOLD),
+            planarSubdivisionControl(options.maxSubdivisionLevel || 5),
```
autra, line 103: Maybe obvious question, but why did you change the default maxLevel from 4 to 5?

autra (commented):
Nice! It feels more natural and better to adopt a perspective approach to this problem. I just have a few questions.

src/Process/PlanarTileProcessing.js
peppsac: `0.5` is needed because `box3DSizeOnScreen` result goes from `-1` to `+1` (for each axis), so we get back to `[0, 1]`.

The `>` vs `>=` is a typo, I'll fix this.

The `0.3` value are `threshold` that looks good on the sole example of planar mode that we have.
I can make them configurable if you want but honestly this PR is a first iteration and we will probably need to adjust it later (for instance to depend on the resolution of the images we get from the layer).

src/Core/Prefab/PlanarView.js
peppsac: `4`  or `5` are both hand-picked values that look good.
Again, we probably want something smarter (e.g depending on the area covered?) but that's for a another PR...

autra (approved):
LGTM

Issue #278, opened 2017-03-30
Add a createGlobe API to pass an existing Three.js scene


mattrei:
Hi,

I am totally impressed by the itowns2 globe example and the API in general. Looks wonderful to me.

However I have one design request. I would like to use the globe example on an already defined THREEJs Scene. Currently the `createSceneGlobe`? API expects a _div_ and creates the renderer and scene itself.
I think it is a great idea to also have `createGlobe` API where I can pass a `scene` object myself.

Having a quick glance at the well structured source, it may be possible to add feature without having a big impact on the code. 

What do you think about it?

Thx,
Matthias

gchoqueux:
Hello
Very pleased that you like it.

For the moment, it is not possible to create a globe alone. There is a specific scene graph path in our API. Maybe we will later merge the two scene graph paths.
In any case it will not be possible in the first release.

But soon there will be tools to integrate your own 3D objects.

autra:
Hey @mattrei, it sounds very interesting! Can you tell us a bit more about your use case? Let us see what we can do for you!

mattrei:
hi  @gchoqueux  and  @autra,
thanks for the interest!

My use case is quite simple: I am currently developing for my thesis an educational web application, where pupils can interact with a globe. Currently my globe, and thats fine for the requirements so far, is quite static. As everything is developed in Three.js my globe is visually just a sphere with a earth like texture popped on it.

However your example impressed me quite a bit, because its basically a Cesium3D globe, built on the Three.js stack. The Cesium3D globe is not compatible within Three.js.

That said, it would be great, probably for many applications, not just mine, to have an interactive globe as provided in your examples, where you users can zoom in and zoom out accordingly and the raster tiles get updated accordingly. 
However It just can't use a Three.js scene provided by you, however I would love to have a way to add the globe as an Object3D to my scene.

Hope I could clarify my use case.

Thx for your efforts and interest!

gchoqueux:
Sure enough it would be nice if we can add the globe to a scene three.js.
But this isn't foreseen in the first release. On the other hand we will quickly propose tools to add objects to the ground.

> However It just can't use a Three.js scene provided by you

Why?

autra:
@mattrei you can already get a ref to the THREE js scene we instanciate with `itowns.viewer.scene.gfxEngine.scene3D`. Also, the scene is instanciated in `c3DEngine.js`, maybe it's not that hard to patch itowns to pass an already instanciated scene to c3DEngine.

[...]

PR #341, opened 2017-06-19
feature: support user supplied camera

peppsac:
Allowing the user to supply its own camera makes itowns integration easier.

Should help to fix issue #278 

r?

### DIFF
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -8,17 +8,19 @@
 
 import * as THREE from 'three';
 
-function Camera(width, height) {
+function Camera(width, height, options = {}) {
     this.ratio = width / height;
-    this.FOV = 30;
 
-    this.camera3D = new THREE.PerspectiveCamera(this.FOV, this.ratio);
+    this.camera3D = options.camera ? options.camera : new THREE.PerspectiveCamera(30, this.ratio);
 
-    // /!\ WARNING Matrix JS are in Float32Array
-    this.camera3D.matrixWorld.elements = new Float64Array(16);
 
-    this.camera3D.matrixAutoUpdate = false;
-    this.camera3D.rotationAutoUpdate = false;
+    if (!options.camera) {
+        // do not modify a camera we don't own fully
+        // /!\ WARNING Matrix JS are in Float32Array
+        this.camera3D.matrixWorld.elements = new Float64Array(16);
+        this.camera3D.matrixAutoUpdate = false;
+        this.camera3D.rotationAutoUpdate = false;
+    }
 
     this._viewMatrix = new THREE.Matrix4();
     this._visibilityTestingOffset = new THREE.Vector3();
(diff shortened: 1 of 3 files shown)


src/Renderer/Camera.js
```diff
@@ -35,18 +37,16 @@ Camera.prototype.resize = function resize(width, height) {
...
+    if (this.camera3D.aspect != this.ratio) {
+        this.camera3D.aspect = this.ratio;
+        if (this.camera3D.isOrthographicCamera) {
+            const halfH = (this.camera3D.right - this.camera3D.left) * 0.5 / this.ratio;
+            this.camera3D.top = halfH;
+            this.camera3D.bottom = -halfH;
+        }
     }
+
+    this.camera3D.updateProjectionMatrix();
```
autra, line 49: Should this be put in the `if (this.camera3D.aspect !== this.ratio)`above ?

src/Renderer/Camera.js
```diff
@@ -35,18 +37,16 @@ Camera.prototype.resize = function resize(width, height) {
...
-
-    if (this.cameraHelper) {
-        var dir = new THREE.Vector3(0, 0, -1);
-        var quaternion = new THREE.Quaternion();
-        quaternion.setFromAxisAngle(new THREE.Vector3(0, 1, 0), this.HFOV / 2);
-        dir.applyQuaternion(quaternion);
-
-        this.arrowHelper.setDirection(dir);
-        this.cameraHelper.update();
+    if (this.camera3D.aspect != this.ratio) {
```
autra: nits `!==`

src/Renderer/Camera.js
```diff
@@ -8,17 +8,19 @@
 
 import * as THREE from 'three';
 
-function Camera(width, height) {
+function Camera(width, height, options = {}) {
     this.ratio = width / height;
-    this.FOV = 30;
```
autra, line 13: ❤

src/Renderer/Camera.js
```diff
@@ -35,18 +37,16 @@ Camera.prototype.resize = function resize(width, height) {
...
-        var quaternion = new THREE.Quaternion();
-        quaternion.setFromAxisAngle(new THREE.Vector3(0, 1, 0), this.HFOV / 2);
-        dir.applyQuaternion(quaternion);
-
-        this.arrowHelper.setDirection(dir);
-        this.cameraHelper.update();
+    if (this.camera3D.aspect != this.ratio) {
+        this.camera3D.aspect = this.ratio;
+        if (this.camera3D.isOrthographicCamera) {
+            const halfH = (this.camera3D.right - this.camera3D.left) * 0.5 / this.ratio;
```
autra: why not `const halfH = height * 0.5` ? Seems exactly the same thing for me. Am I right assuming `this.camera3D.right - this.camera3D.lef === width`? Also, I was expecting `camera3D.{right,left}` to be updated here?

autra (changes requested):

src/Renderer/Camera.js
peppsac: No, because when using orthographic camera you need to call `updateProjectionMatrix` when you change one of the `.left` / `.rightt` / `.bottom` / `.top` property

src/Renderer/Camera.js
peppsac: Yes I'll change to `height * 0.5`.

And no need to update right/left: I chose to keep right/left constant when resizing, so I just need to update top/bottom to match the new view ratio.

peppsac: Err sorry I was wrong :)
`height` is a pixel value while `this.camera3D.right` - `this.camera3D.left` are using CRS value, ie: the geographic area visible that will be drawn on the screen.

src/Renderer/Camera.js
peppsac: Fixed

src/Renderer/Camera.js
autra: ok, thanks for the explanation!

src/Renderer/Camera.js
autra: But you are changing them only in this `if`, or did I miss something?

peppsac: They can be changed outside as well (e.g: in a control)

autra: ok thanks!

autra (approved):

PR #345, opened 2017-06-27
refactor: remove BasicMaterial

peppsac:
BasicMaterial main purpose was to implement RTC.
Now that RTC is gone, BasicMaterial can be removed as well.
Key benefit: we can use regular three.js features, like Lights or MeshLambertMaterial!

And since this PR touches a lot of shaders, I also took advantage of three.js glsl


peppsac:
I'd like to remove `ItownsPointMaterial` and `ItownsLineMaterial` as well which aren't really used for now.
`three` already have configurable materials for drawing points and lines, so I'd rather use them.

WDYT?

peppsac:
I pushed another commit:
-  it removes ItownsPoint* and ItownsLine*
- replaces them with three.js features

src/Renderer/Shader/TileFS.glsl
```diff
@@ -36,19 +38,25 @@ varying vec3        vNormal;
...
+#endif
 
 void main() {
+    #include <logdepthbuf_fragment>
 
-    #if defined(USE_LOGDEPTHBUF) && defined(USE_LOGDEPTHBUF_EXT)
-
-	   gl_FragDepthEXT = log2(vFragDepth) * logDepthBufFC * 0.5;
+    #if defined(MATTE_ID_MODE)
+        gl_FragColor = packDepthToRGBA(float(uuid) / (256.0 * 256.0 * 256.0));
```
autra, line 50: `packDepthToRGBA` supposes a float between 0 and 1, that's why?

src/Main.js
```diff
@@ -27,6 +27,7 @@ export { GeometryLayer, ImageryLayers } from './Core/Layer/Layer';
 export { process3dTilesNode, init3dTilesLayer, $3dTilesCulling, $3dTilesSubdivisionControl, pre3dTilesUpdate } from './Process/3dTilesProcessing';
 export { STRATEGY_MIN_NETWORK_TRAFFIC, STRATEGY_GROUP, STRATEGY_PROGRESSIVE, STRATEGY_DICHOTOMY } from './Core/Layer/LayerUpdateStrategy';
 
+export { default as loadGpx } from './Core/Scheduler/Providers/GpxUtils';
```
autra: (GpxUtils seems to be broken on master btw)

autra (approved):
Looks good to me, I just took advantage of the review to ask one question :-)

src/Renderer/Shader/TileFS.glsl
peppsac: yes

PR #348, opened 2017-06-28
Fetch options

autra:
Give the user the possibility to give options to network request, to handle cors, credentials and other stuff.

The second commit fixes an issue I only have with Firefox dev edition: the view is not automatically refreshed after I load the globe.

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -135,7 +135,7 @@ $3dTiles_Provider.prototype.executeCommand = function executeCommand(command) {
             pnts: this.pntsParse.bind(this),
         };
 
-        return Fetcher.arrayBuffer(url).then((result) => {
+        return Fetcher.arrayBuffer(url, layer.networkOptions).then((result) => {
             if (result !== undefined) {
                 let func;
                 const magic = textDecoder.decode(new Uint8Array(result, 0, 4));
--- a/src/Core/Scheduler/Providers/Fetcher.js
+++ b/src/Core/Scheduler/Providers/Fetcher.js
@@ -2,8 +2,6 @@ import { TextureLoader } from 'three';
 
 const textureLoader = new TextureLoader();
 
-textureLoader.setCrossOrigin('anonymous');
-
 function checkResponse(response) {
     if (response.status < 200 || response.status >= 300) {
         var error = new Error(`Error loading ${response.url}: status ${response.status}`);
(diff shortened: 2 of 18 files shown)


autra:
@iTowns/reviewers please r?

src/Core/Scheduler/Providers/Fetcher.js
```diff
@@ -14,23 +12,55 @@ function checkResponse(response) {
...
-        return fetch(url).then((response) => {
+    /**
+     * Little wrapper over fetch to get some json
+     *
+     * @param {string} url
+     * @param {Object} options - fetch options (passed directly to fetch)
+     *
+     * @return {Promise}
+     */
+    json(url, options = {}) {
```
peppsac, line 23: Can't we keep the same default value:

```js
options = { 'crossOrigin': 'anonymous' }
```

Or even better, default to `crossOrigin: anonymous` unless the user explicitely set a different `crossOrigin` value.

autra: This default was a mistake in the first place, if you ask me. Let's default to same default the browsers have, aka nothing. We're not here to rewrite the web for the user in my opinion.

peppsac: Ok, I didn't realized that "nothing" was the default for browser.

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -135,7 +135,7 @@ $3dTiles_Provider.prototype.executeCommand = function executeCommand(command) {
             pnts: this.pntsParse.bind(this),
         };
 
-        return Fetcher.arrayBuffer(url).then((result) => {
+        return Fetcher.arrayBuffer(url, layer.fetchOptions).then((result) => {
```
peppsac: Why did you use sometimes `fetchOptions` and sometimes `textureLoaderOptions` name?

I'd prefer a single name: e.g `fetchOptions` or `networkOptions`

autra: There's a different set of options for `Fetcher.texture` (which uses TextureLoader under the hood) and `Fetcher.{json,xml}`. They don't overlap, but I thought it was clearer that way. Now I don't really know, your call :-)

tbroyer: I'd say +1 for `fetchOptions` vs. `textureLoaderOptions`; they are completely different, with no overlap, and different ways to say the same thing.

They need to be documented in `Core/View.js` though (`LayerOptions` type def)

peppsac: Both are a way to give network usage options, so using the same name (`networkOptions`) makes sense.

The doc can then explain  that the valid values depend on the content of the layer.

autra:
updated, please let me know what you think

peppsac (approved):
LGTM, thanks

PR #351, opened 2017-06-29
fix: recursively cleanup all nodes

peppsac:
Makes sure we don't produce orphan THREE.Object3D.

@iTowns/reviewers r?


autra (approved):
LGTM :-)

PR #355, opened 2017-06-29
Add log depth buffer support for imported b3dm models

Jeremy-Gaillard:
Imported model do not use log depth buffer. To have the same depth buffer logic for all our meshes (terrain and b3dm), we have to edit the shaders ourselves (or until three is patched).

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -6,6 +6,8 @@ import Fetcher from './Fetcher';
 import OBB from '../../../Renderer/ThreeExtended/OBB';
 import Extent from '../../Geographic/Extent';
 import MathExtended from '../../Math/MathExtended';
+import Capabilities from '../../System/Capabilities';
+import PrecisionQualifier from '../../../Renderer/Shader/Chunk/PrecisionQualifier.glsl';
 
 
 export function $3dTilesIndex(tileset, urlPrefix) {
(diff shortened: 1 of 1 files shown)


src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -80,12 +81,39 @@ function getBox(volume) {
...
     return this.b3dmLoader.parse(data).then((result) => {
         const init = function f_init(mesh) {
             mesh.frustumCulled = false;
-            if (layer.overrideMaterials) {
-                mesh.material = new THREE.MeshLambertMaterial(0xffffff);
+            if (mesh.material) {
+                if (layer.overrideMaterials) {
+                    mesh.material = new THREE.MeshLambertMaterial(0xffffff);
+                } else if (Capabilities.isLogDepthBufferSupported()) {
+                    patchMaterial(mesh.material);
```
peppsac: This is needed only when the GLTFLoader uses a `RawShaderMaterial` - in all other cases three.js does it for us.

Adding a test condition `mesh.material.isRawShaderMaterial` should do the job.

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -80,12 +81,39 @@ function getBox(volume) {
...
+function patchMaterial(material) {
+    // Add vertex shader log depth buffer header
+    material.vertexShader = `#include <logdepthbuf_pars_vertex>\n#define EPSILON 1e-6\n${material.vertexShader}`;
+    // Add log depth buffer code snippet after last gl_Position modification
+    let re = rePosition.exec(material.vertexShader);
+    let idx = re[0].length + re.index;
+    material.vertexShader = `${material.vertexShader.slice(0, idx)}\n#include <logdepthbuf_vertex>\n${material.vertexShader.slice(idx)}`;
+
+    // Add fragment shader log depth buffer header
+    material.fragmentShader = `precision highp float;\n#include <logdepthbuf_pars_fragment>\n${material.fragmentShader}`;
```
peppsac: You could use `src/Renderer/Shader/Chunk/PrecisionQualifier.glsl` here.

src/Core/Scheduler/Providers/3dTiles_Provider.js
peppsac: Maybe you could also make sure the shader doesn't already implement this?
(maybe looking for `USE_LOGDEPTHBUF`, `#include <logdepthbuf` strings ? or `gl_FragDepthEXT` writes in the fragment shader?

In all cases I'd suggest also to:
- log a warning indicating that we patched the file
- add a way for the user to disable this patch (an option in layer for instance ?)

Jeremy-Gaillard:
Updated. Implemented all your suggestions.

peppsac:
Hijicking this PR for a Travis question: it seems this PR only have 1 Travis check while others have 2 (see #309 for instance).
Any idea when/why did that change?

tbroyer:
The Travis jobs are still queued, that's probably why we only see one line. But there are 2 jobs: [for the branch](https://travis-ci.org/iTowns/itowns2/builds/248375872), and [for the PR](https://travis-ci.org/iTowns/itowns2/builds/248375890)

peppsac:
@tbroyer thanks for the details

peppsac (approved):
I'm not sure if it's a good idea at all or if the patching process is robust enough...
Anyway, if it fixes at least one bug and doesn't cause new ones, we can merge this fix for now and adapt it later.

peppsac:
(I've manually cancelled then restarted the builds because they seemed stuck)

Issue #353, opened 2017-06-29
Annoying Damping effect when we want camera to be static


nosy-b:
The damping effect is nice but it appears that sometimes the tweening comes when we wanted the camera to be static. Case:
Fast mouse-move-leftClick for camera rotation then stop for a very short time and release click -> camera will move with damping although it should be stopped.

This occur in GlobeControls.js on MouseUp when comparing last camera positions (`lastRotation[1].equals(lastRotation[0]`). Value samples are not made often/fast enough so they differ and the motion happens.

Solution 1: Change the dampingFactor but it might reduce too much the tweening effect for some users desire.
Solution 2: Update more often the camera position difference
Other solutions might exists more direct.

PR #356, opened 2017-06-29
fix: damping effect when we want camera fixed after movement

nosy-b:
It answers issue #353 so now when moving the camera precisely it should not tween after the mouseup.

### DIFF
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -155,6 +155,8 @@ let enableAnimation = true;
 var player = null;
 // Save 2 last rotation globe for damping
 var lastRotation = [];
+// Save the last time of mouse move for damping
+var lastTimeMouseMove = 0;
 
 // Expression used to damp camera's moves
 var dampingMoveAnimatedExpression = (function getDampMoveAniExprFn() {
@@ -887,6 +889,7 @@ function GlobeControls(view, target, domElement, viewerDiv, radius, getPickingPo
                     quatGlobe.setFromUnitVectors(normalizedIntersection, tSphere.picking.normal);
                     // backups last move globe for damping
                     lastRotation.push(normalizedIntersection.clone());
+                    lastTimeMouseMove = Date.now();
                     // Remove unnecessary movements backups
                     if (lastRotation.length > 2) {
                         lastRotation.splice(0, 1);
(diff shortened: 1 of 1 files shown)


src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -994,7 +997,7 @@ function GlobeControls(view, target, domElement, viewerDiv, radius, getPickingPo
         if (this.enableDamping) {
             if (state === CONTROL_STATE.ORBIT && (sphericalDelta.theta > EPS || sphericalDelta.phi > EPS)) {
                 player.play(dampingOrbitalMvt).then(() => this.resetControls());
-            } else if (state === CONTROL_STATE.MOVE_GLOBE && lastRotation.length === 2 && !(lastRotation[1].equals(lastRotation[0]))) {
+            } else if (state === CONTROL_STATE.MOVE_GLOBE && lastRotation.length === 2 && (Date.now() - lastTimeMouseMove < 50) && !(lastRotation[1].equals(lastRotation[0]))) {
```
peppsac, line 1000: Can you add a comment explaining your intent here?

nosy-b: I choose explicit variable name this time ;) 
We basically check time between the last mouse move and the mouse up to detect if tweening is wanted or not

peppsac: "We basically check time between the last mouse move and the mouse up" => this is indeed obvious from the variable naming :)

"if tweening is wanted or not" => the logic may be obvious to you, but from an outsider point of view it isn't, so it's worth adding a little comment.
Here's what I expected to read (assuming I understood correctly the test):
```js
// disable animation is mouse up event occurs less than 50ms after the last mouse move
```

nosy-b: It stayed in the documentation style of this class ;)  (which is a temporary class from what @gchoqueux said).
Though I agree it's never bad to put too much comments, I'll rebase

nosy-b: @peppsac commented and rebased

tbroyer: An alternative would be to extract the check into a function:

```js
/**
 * Returns whether to animate the XXX.
 *
 * Animation is only enabled if mouse is moving (approximation).
 */
function shouldAnimate() {
  return (Date.now() - lastTimeMouseMove < 50) && !(lastRotation[1].equals(lastRotation[0]));
}

…

 } else if (state === CONTROL_STATE.MOVE_GLOBE && lastRotation.length === 2 && shouldAnimate()) {
```

peppsac: Didn't you forget a word at the beginning of your comment (`disable` I suppose)?

And yes `GlobeControls` would benefit from being completely rewritten, taking advantages of itowns `Coordinates`.

peppsac: While I approve the idea of moving this to a separate function, I'd prefer a more concise version:

```js
function shouldAnimate() {
  // animate only if XXXXX and YYY
  return (Date.now() - lastTimeMouseMove < 50) && !(lastRotation[1].equals(lastRotation[0]));
}
```
But well, @nosy-b its your code, so as long there's a comment you can write it the way you like :)

nosy-b: `// animation since mouse up event occurs less than 50ms after the last mouse move` comment is ok: the test passed so we animate.

peppsac (approved):

PR #358, opened 2017-06-29
Fix: correct network options for all examples and for TMS_Provider

autra:
This is a followup from #348 where I forgot a lot of things... Sorry for that!

Apparently, we don't need any cors options for ArrayBuffer fetching.

I didn't change the default of the web (so by default, fetches have mode and credentials not set, and img have crossOrigin not set), because apparently it helps to prevent CRSF attacks.

### DIFF
--- a/src/Core/Scheduler/Providers/WMS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMS_Provider.js
@@ -101,7 +101,7 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer) {
 
 WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer) {
     const url = this.url(tile.extent.as(layer.projection), layer);
-    return this.getXBilTextureByUrl(url, new THREE.Vector3(0, 0, 1));
+    return OGCWebServiceHelper.getXBilTextureByUrl(url, layer.networkOptions);
 };
 
 WMS_Provider.prototype.executeCommand = function executeCommand(command) {
--- a/test/itowns-testing.js
+++ b/test/itowns-testing.js
@@ -28,9 +28,10 @@ global.fetch = function _fetch(url) {
             counters.fetch.push(url);
         }
         res({
-            buffer: () => content || {},
-            json: () => JSON.parse(content),
-            arrayBuffer: () => content || {},
+            buffer: () => Promise.resolve(content || {}),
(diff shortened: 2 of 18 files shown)


src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -101,7 +101,7 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer) {
 
 WMS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer) {
     const url = this.url(tile.extent.as(layer.projection), layer);
-    return this.getXBilTextureByUrl(url, new THREE.Vector3(0, 0, 1));
+    return OGCWebServiceHelper.getXBilTextureByUrl(url, layer.networkOptions);
```
autra, line 104: I will be surprised if this still works but well...

autra:
@iTowns/reviewers please r?

src/Core/Scheduler/Providers/WMS_Provider.js
peppsac: This method has never been used, nor been tested: it only exists because `WMS_Provider` used `WMTS_Provider` as a template.

peppsac (approved):
Thanks for the fix.

And yes, our testing process is really poor: regressions still occur on a regular basis...

autra:
> And yes, our testing process is really poor: regressions still occur on a regular basis...

And it's fragile: there it is failing without any good reason.

autra:
I *must* have done something wrong, because here it is *consistently* failing! Looking into it.

autra:
I'm actually not sure *at all* my commit really fixed this. @peppsac please r? the last commit.

Btw, I have started to integrate mock-browser. I think it'll save us from having to reimplement correct dom behaviour for each API.

test/itowns-testing.js
```diff
@@ -28,9 +28,10 @@ global.fetch = function _fetch(url) {
             counters.fetch.push(url);
         }
         res({
-            buffer: () => content || {},
-            json: () => JSON.parse(content),
-            arrayBuffer: () => content || {},
+            buffer: () => Promise.resolve(content || {}),
+            json: () => Promise.resolve(JSON.parse(content)),
+            arrayBuffer: () => Promise.resolve(content || {}),
+            ok: true,
```
peppsac, line 34: Yup, looks better than the previous implementation. Thanks

PR #362, opened 2017-06-30
Examples formatting

autra:
First pass at cleaning the examples (there will be others I'm sure).

They will be part of our vitrine for our users, let's make them shiny and polished!

@Jeremy-Gaillard please have a look at the 3dtiles examples. I have integrated the debug functions into Debug.js. 

About Debug.js, I think it should be splitted soon...

@iTowns/reviewers please r?

### DIFF
--- a/examples/multiglobe.html
+++ b/examples/multiglobe.html
@@ -6,23 +6,31 @@ and open the template in the editor.
 -->
 <html>
     <head>
-        <title>Globe</title>
+        <title>Itowns - multiglobe</title>
 
-    <style type="text/css">
-            html {height: 100%}
-        body { margin: 0; overflow:hidden; height:100%}
+        <style type="text/css">
+            html {
+                height: 100%;
+            }
+
+            body {
+                margin: 0;
+                overflow: hidden;
+                height: 100%;
+            }
 
             div {
                 margin : auto auto;
                 width: 100%;
                 padding: 0;
                 height: 50%
-                /*margin-top: 50vh;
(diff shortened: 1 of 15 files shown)


peppsac (dismissed):
Looks good, r+

Jeremy-Gaillard:
Thanks for the work @autra.

Just two remarks:
- Please set one of the layer visible by default (3d-tiles-discrete-lod)
- The text from the debug bbox of the request volume tileset stays displayed even after hiding the debug layer

autra:
@Jeremy-Gaillard updated. I've also updated the starting position

@peppsac please r? again, especially the last commit

Thanks!

autra:
@peppsac please r?

examples/multiglobe.html
```diff
@@ -8,21 +8,29 @@
     <head>
         <title>Globe</title>
```
peppsac: You might want to change the title as well

examples/planar.html
```diff
@@ -3,109 +3,33 @@
         <title>Globe</title>
```
peppsac: Change title ?

examples/postprocessing.html
```diff
@@ -3,9 +3,16 @@
     <head>
         <title>Globe</title>
```
peppsac: Title...

peppsac (commented):
+1
You only need to fix a few `<title>` and we'll be able to merge this PR.

peppsac (approved):

PR #363, opened 2017-07-03
Remove useless events and add a preprocessed promise

autra:
We dont need to fire events for synchronous methods. It's useless, but
even more, it's confusing for users.

This commit adds a promise to deal with the only asynchronous case:
preprocessLayer. This *will* get asynchronous in the future when we'll
support GetCapabilities call and other metadata fetching methods.

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -1,7 +1,6 @@
 import * as THREE from 'three';
 
 import View from '../View';
-import { COLOR_LAYERS_ORDER_CHANGED } from '../../Renderer/ColorLayersOrdering';
 import RendererConstant from '../../Renderer/RendererConstant';
 import GlobeControls from '../../Renderer/ThreeExtended/GlobeControls';
 import { unpack1K } from '../../Renderer/LayeredMaterial';
@@ -24,51 +23,14 @@ import BuilderEllipsoidTile from './Globe/BuilderEllipsoidTile';
  * @property target {view} dispatched on view
  * @property type {string} initialized
  */
-/**
- * Fires when a layer is added
- * @event GlobeView#layer-added
- * @property layerId {string} the id of the layer
- * @property target {view} dispatched on view
- * @property type {string} layers-added
- */
-/**
- * Fires when a layer is removed
- * @event GlobeView#layer-removed
- * @property layerId {string} the id of the layer
- * @property target {view} dispatched on view
- * @property type {string} layers-added
- */
-/**
- * Fires when the layers oder has changed
- * @event GlobeView#layers-order-changed
- * @property new {object}
- * @property new.sequence {array}
- * @property new.sequence.0 {number} the new layer at position 0
- * @property new.sequence.1 {number} the new layer at position 1
- * @property new.sequence.2 {number} the new layer at position 2
- * @property previous {object}
- * @property previous.sequence {array}
- * @property previous.sequence.0 {number} the previous layer at position 0
- * @property previous.sequence.1 {number} the previous layer at position 1
- * @property previous.sequence.2 {number} the previous layer at position 2
- * @property target {view} dispatched on view
- * @property type {string} layers-order-changed
- */
-
 
 /**
  * Globe's EVENT
  * @property GLOBE_INITIALIZED {string} emit one time when globe is initialized
- * @property LAYER_ADDED {string} emit when layer id added in viewer
- * @property LAYER_REMOVED {string} emit when layer id removed in viewer
- * @property COLOR_LAYERS_ORDER_CHANGED {string} emit when  color layers order change
  */
 
 export const GLOBE_VIEW_EVENTS = {
     GLOBE_INITIALIZED: 'initialized',
-    LAYER_ADDED: 'layer-added',
-    LAYER_REMOVED: 'layer-removed',
-    COLOR_LAYERS_ORDER_CHANGED,
 };
 
 /**
@@ -271,15 +233,7 @@ GlobeView.prototype.addLayer = function addLayer(layer) {
         }
         layer.update = updateLayeredMaterialNodeElevation;
     }
-    const layerId = layer.id;
-    View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
-
-    this.dispatchEvent({
-        type: GLOBE_VIEW_EVENTS.LAYER_ADDED,
-        layerId,
-    });
-
-    return layer;
+    return View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
 };
 
 /**
--- a/src/Renderer/ColorLayersOrdering.js
+++ b/src/Renderer/ColorLayersOrdering.js
@@ -13,9 +13,7 @@ function updateLayersOrdering(geometryLayer, imageryLayers) {
     }
 }
 
-export const COLOR_LAYERS_ORDER_CHANGED = 'layers-order-changed';
-
-export const ColorLayersOrdering = {
+const ColorLayersOrdering = {
     /**
      * Moves up in the layer list. This function has no effect if the layer is moved to its current index.
      * @function moveLayerUp
@@ -28,13 +26,8 @@ export const ColorLayersOrdering = {
         const imageryLayers = view.getLayers(l => l.type === 'color');
         const layer = view.getLayers(l => l.id === layerId)[0];
         if (layer) {
-            const previousSequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
             ImageryLayers.moveLayerUp(layer, imageryLayers);
             updateLayersOrdering(view.wgs84TileLayer, imageryLayers);
-            view.dispatchEvent({ type: COLOR_LAYERS_ORDER_CHANGED,
-                previous: { sequence: previousSequence },
-                new: { sequence: ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers) },
-            });
             view.notifyChange(true);
         } else {
             throw new Error(`${layerId} isn't color layer`);
@@ -52,20 +45,17 @@ export const ColorLayersOrdering = {
         const imageryLayers = view.getLayers(l => l.type === 'color');
         const layer = view.getLayers(l => l.id === layerId)[0];
         if (layer) {
-            const previousSequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
             ImageryLayers.moveLayerDown(layer, imageryLayers);
             updateLayersOrdering(view.wgs84TileLayer, imageryLayers);
-            view.dispatchEvent({ type: COLOR_LAYERS_ORDER_CHANGED,
-                previous: { sequence: previousSequence },
-                new: { sequence: ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers) },
-            });
             view.notifyChange(true);
         } else {
             throw new Error(`${layerId} isn't color layer`);
         }
     },
     /**
-     * Moves a specific layer to a specific index in the layer list. This function has no effect if the layer is moved to its current index.
+     * Moves a specific layer to a specific index in the layer list. This
+     * function has no effect if the layer is moved to its current index.
+     *
      * @function moveLayerToIndex
      * @param      {View}  view the viewer
      * @param      {string}  layerId   The layer's idendifiant
@@ -77,16 +67,12 @@ export const ColorLayersOrdering = {
         const imageryLayers = view.getLayers(l => l.type === 'color');
         const layer = view.getLayers(l => l.id === layerId)[0];
         if (layer) {
-            const previousSequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
             ImageryLayers.moveLayerToIndex(layer, newIndex, imageryLayers);
             updateLayersOrdering(view.wgs84TileLayer, imageryLayers);
-            view.dispatchEvent({ type: COLOR_LAYERS_ORDER_CHANGED,
-                previous: { sequence: previousSequence },
-                new: { sequence: ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers) },
-            });
             view.notifyChange(true);
         } else {
             throw new Error(`${layerId} isn't color layer`);
         }
     },
 };
+export default ColorLayersOrdering;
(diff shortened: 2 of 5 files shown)


src/Renderer/ColorLayersOrdering.js
```diff
@@ -13,9 +13,7 @@ function updateLayersOrdering(geometryLayer, imageryLayers) {
     }
 }
 
-export const COLOR_LAYERS_ORDER_CHANGED = 'layers-order-changed';
-
-export const ColorLayersOrdering = {
+const ColorLayersOrdering = {
```
peppsac, line 16: Didn't you forget to modify GuiTools:
```js
examples/GUI/GuiTools.js:        view.addEventListener('layers-order-changed', () => {
```
?

autra: yep!

autra:
@peppsac updated

gcebelieu:
Hi,

As iTowns user, I don't find confusing the events that are proposed to be removed. On the contrary, they are very useful to our developments to integrate iTowns into our geoportal projects : 

* Widget library for iTowns : 
https://github.com/IGNF/geoportal-extensions/blob/itowns_controls/src/Itowns/Controls/LayerSwitcher.js#L178
...

* iTowns integration into our SDK :
https://github.com/IGNF/geoportal-sdk/blob/itownsControls/src/itowns/IT.js#L883 
...

Why remove them ?

Our typical usage is a widget that reflects layers stack and order on the globe (LayerSwitcher). This layers stack may be modified by actions coming from other components than our widget (external API call made by an application using our widget and iTowns). How can the widget be informed that such layer has been removed / added / moved up ou down without listening to those events ?

autra:
> Our typical usage is a widget that reflects layers stack and order on the globe (LayerSwitcher). This layers stack may be modified by actions coming from other components than our widget (external API call made by an application using our widget and iTowns). How can the widget be informed that such layer has been removed / added / moved up ou down without listening to those events ?

You've said it: actions coming from "other components". So "other components" should fire the event, because it is responsible for it, NOT itowns. You can see "other components" as the `Controller`, itowns as the `Model` and the widget as the `View` layer of a MVC arch, if you want. The Controller should update the widget.

It's not like its a complicated change: just add `this.dispatchEvent(myEvent)` in`IT.js` [here](https://github.com/IGNF/geoportal-sdk/blob/itownsControls/src/itowns/IT.js#L246)! Hell, you can even do:
```js
this.libMap.dispatchEvent({
        type: 'layer-added',
        layerId: layer.id,
});
``` 
at this place if you don't want IT.js to implement EventDispatcher, without the event being defined in itowns. You wouldn't even need to modify LayerSwitcher !! It's dirty but it'll work. Itowns is designed to make you able to do *whatever* you want, and really javascript is made that way btw.

Really, I've told you 2 times in person (once in January, once in March) and several times in issues and PR in this repo (too lazy to link to these comments right now). Is it really necessary to be so afraid of a one line change in your side? Honestly I'll make the PR if it's too hard for you guys ;-) (I'm just provoking you here ;-) ). 

And PROMISE if it turns out it's not possible to do it this way, I'll  add it back to itowns, I *swear*.

gchoqueux:
I'll watch this PR

peppsac:
There's a single "user" of iTowns - it's not written as a multiple entry point framework.
This "user" can be an application (like our examples) or your SDK.

A single user mean: no one can change iTowns state but you.

In the SDK case you wrap itowns and hide it from the user, so the app using the SDK isn't going to directly use iTowns: they're using iTowns through your API.

So what @autra propose makes sense IMHO: use your API to add all the events you need or to directly update your UI.

gchoqueux:
Adding a layer isn't really synchronous, it's interesting to know when all the tiles have been updated and display the new layer. In this case, the simplest is the use of events (especially for the synchronization of interfaces)

src/Core/Prefab/GlobeView.js
```diff
@@ -24,51 +23,14 @@ import BuilderEllipsoidTile from './Globe/BuilderEllipsoidTile';
...
  * Globe's EVENT
  * @property GLOBE_INITIALIZED {string} emit one time when globe is initialized
- * @property LAYER_ADDED {string} emit when layer id added in viewer
- * @property LAYER_REMOVED {string} emit when layer id removed in viewer
- * @property COLOR_LAYERS_ORDER_CHANGED {string} emit when  color layers order change
  */
 
 export const GLOBE_VIEW_EVENTS = {
     GLOBE_INITIALIZED: 'initialized',
-    LAYER_ADDED: 'layer-added',
```
gchoqueux, line 69: It is not so simple to withdraw it eventually without a little consultation.
A more judicious event will be `newLayerHasBeenDiplayed`

gchoqueux (changes requested):

autra:
> Adding a layer isn't really synchronous

No it's not. Where is the asynchronous action?

peppsac:
@gchoqueux currently the event is fired right after adding the layer.
```js
GlobeView.prototype.addLayer = function addLayer(layer) {
    [...]
    this.dispatchEvent({
        type: GLOBE_VIEW_EVENTS.LAYER_ADDED,
        layerId,
    });
}
```
so it has little to no value.

Adding another event like `newLayerHasBeenDiplayed` might be useful, but it's unrelated to this PR.

gchoqueux:
This event GLOBE_VIEW_EVENTS.LAYER_ADDED is actually not very sensible, I propose to put it in **deprecated** for the transition while waiting for newLayerHasBeenDiplayed or removedLayerHasBeenHidden

gchoqueux:
the event should be `newLayerHasBeenLoadedAndDiplayed `

gcebelieu:
@autra : we know how to dispatch events. This is not the point and that's why we're discussing another time about this (this time in july ;) )

> There's a single "user" of iTowns - it's not written as a multiple entry point framework.
> This "user" can be an application (like our examples) or your SDK.
> 
> A single user mean: no one can change iTowns state but you.

That's the point : why this restriction ? This "single user" approach prevents us to provide additionnal tools (our geoportal extension for iTowns but this might inspire others) to be used together with iTowns by a third party application.
We aggreed to realize the widgets we need in a separate project but this requires a minimum glue from iTowns Core to works together. Those events are IMHO part of this glue.

I would also say : no one can tell about iTowns state but iTowns. It doesn't make sense to me to implement events management about iTowns state in a separate extension : we had a similar discussion about the API.

> In the SDK case you wrap itowns and hide it from the user, so the app using the SDK isn't going to directly use iTowns: they're using iTowns through your API.
> 
> So what @autra propose makes sense IMHO: use your API to add all the events you need or to directly update your UI.

In the SDK case I share this vision with you because of the wrap you mentionned. The point is for the extension. See above

autra:
@gcebelieu 

>> A single user mean: no one can change iTowns state but you.
>
> That's the point : why this restriction ?

You missed @peppsac's point completely. On this particular instance, itowns is not going to remove or add layers behind the client app back. That would be weird right? It's not an approach we chose, it's a fact of the web: itowns is used by one webpage, not several at the same time. You can take user = webpage here.

>  It doesn't make sense to me to implement events management about iTowns state in a separate extension

Well, you can't ask itowns to make the glue between your sdk and your extension right? That's an implementation detail about *your* case. How will your sdk and your extension communicate? You need to solve this problem regardless of whether or not this PR gets merged.

> this requires a minimum glue from iTowns Core to works together. Those events are IMHO part of this glue.

No there are not. Again you failed to understand. They give you nothing. `addLayer` is currently a synchronous method. The glue you're talking about is the glue between your sdk and your extension, not itown's job. The sdk *knows* already when a layer is added. It just need to communicate it to your extension.

> This "single user" approach prevents us to provide additionnal tools (our geoportal extension for iTowns but this might inspire others) to be used together with iTowns by a third party application.

Where and how? The patch to adapt your sdk to this PR is this one:

```diff
diff --git i/src/itowns/IT.js w/src/itowns/IT.js
index ccda837..5164212 100644
--- i/src/itowns/IT.js
+++ w/src/itowns/IT.js
@@ -244,6 +244,10 @@ function (
             // we add the layer and refresh the itowns viewer
             // this will launch the addedLayer callback (dans "IT._onLayerChanged")
             this.libMap.addLayer(layer) ;
+            this.libMap.dispatchEvent({
+                type: 'layer-added',
+                layerId: layer.id,
+            });
 
         }
     } ;
@@ -617,6 +621,10 @@ function (
         layerIds.forEach(function (_layerId) {
             if (this._getLayersObj(_layerId) && Array.isArray(this._getLayersObj(_layerId)) && this._getLayersObj(_layerId).length > 0) {
                 this.libMap.removeLayer(_layerId) ;
+                this.dispatchEvent({
+                    type: 'layer-removed',
+                    layerId: _layerId,
+                });
             } else {
                 console.log("[IT]  : Impossible to remove " + _layerId + " - not Found");
             }

```

It fixes both your sdk and your extension. Where are you prevented to do anything? Note that it's one way (the less disruptive) to solve your communication problem between your sdk and your extension.

So try this patch please, and let me know if any difficulties arise.

gcebelieu:
@autra  you're also missing my point : the problem is not communication between our SDK and our extension, but the use of iTowns + our extension by another application. This application would then have to dispatch a "layer-added" event (as you suggest) each time it adds a layer and this event should be supported, documented by our extension whereas it deals with inner iTowns state, not one of our widget's : *this* is weird to me.

If I missed @peppsac's point I'd prefer *he* tells me where.

Finally, I don't understand this argument : 

> We dont need to fire events for synchronous methods. It's useless

I don't see why events should be reserved to asynchrous maters. Why are other widely used mapping libraries trigger events for simple properties changes. Did they also missed something ?

peppsac:
@gcebelieu my point was: your SDK creates itowns (https://github.com/IGNF/geoportal-sdk/blob/itownsControls/src/itowns/IT.js#L78) and I assumed that you don't expect your user to interact directly with the itowns instance and that you expect your users to go through your API.
Am I wrong?
If it's true, there's no benefit for firing an event when calling `addLayer` in iTowns since your SDK is the one calling `addLayer`.

Another concern of mine with these events is that they only work for a narrow use case: you need to use `GlobeView` and attach layer to the globe geometry to get them. So they won't get fired a if you use a `PlanarView`, or if you create a globe from a plain `View`, or if you add a geometry layer, etc.

So maybe in the future we'll need an `addLayer` event, but it will be quite different from the current one.

So instead of making a first release with an event tagged as *deprecated* (as suggested by @gchoqueux ), I suggest taking a step back and:
- removing the event before the first release
- go back to the drawing board and build one (or several) events that fits better itowns' mission ("visualizing any SIG data")
- implement them

In the meantime you can implement @autra suggestion and wait for the outcome.
WDYT?

gcebelieu:
Thanks @peppsac  for this clarification step.

My turn to clarify my position : 

What I'm trying to explain is that we have **two use cases** : the geoportal-sdk one and the geoportal-extension.

1. the **geoportal-sdk** : yes, you've got it : this is a wrapper around iTowns and geoportal-extensions ; it provides it's own narrowed API compared to iTowns framework and yes in this case we could handle events by ourselves because we wrap them into ours. We aggree : end of discussion on that point :+1: 

2. the **geoportal-extensions** : this is a javascript library that provides widgets to be used together with iTowns API. We're not wrapping anything in that case : it is aimed to people who wants to develop web applications using iTowns framework (with all its possibilities) together with geoportal oriented widgets or widgets that are not integrated into iTowns Core because you - I mean : iTowns people - didn't  consider them to be useful to iTowns Core (remember previous discussions).

In that case, some widgets behaviour depends on iTowns state change : layers stack, camera position, camera elevation. ... My opinion and my experience to handle this (we have developped those widgets for OpenLayers, Leaflet and VirtualGeo) is to listen to events reflecting the map or globe view state change : there are always events of that kind ; I don't see anything useless or heretic in that way of making things.

I don't think this is geoportal-extensions role to implement events that relates the iTowns globe state : that would mean that a third party developer using iTowns API and geoportal-extension widgets should explicitly trigger "layer-added" event (for instance : there are other cases) each time it does a iTowns addLayer call to have widgets work as expected. This is non sense to me.

My other argument is that these events are there ; the API has been stabilised those last months and, thanks to that, we made great progres in our integration of iTowns. Our widgets work thanks to events proposed and I don't understand why two days before releasing date there is a sudden urge to remove them from API.

I think that those events have all legitimacy to be in iTowns and should remain as such in that release (I understand @gchoqueux deprecation proposal as a renaming proposal, but he can explain this better than me) not even tagged as deprecated for the moment : this can wait further releases and discussions.

Mkonini:
I agree with @gcebelieu proposition, it is compliant with @peppsac's proposition some days ago :"If something important is missing don't worry, itowns 2.0.1 will be release a few weeks later, and then 2.0.2, etc etc." 
So I'm for keeping stabilized things as they are, improvements should be possible in next releases.

peppsac:
@Mkonini not quite. I was talking about missing features, so in the context of this discussion my previous argument should be understood as `If layer events are missing don't worry [from itowns 2.0.0], itowns 2.0.1 will be release a few weeks later`

But keeping these half-baked event in 2.0.0 makes it difficult to clean them later: as soon as they're part of a released version, any user could argue that he/she wants to keep them because they're useul for him/her.

@gcebelieu I hear you, but in my opinion you forget important things:
- itowns is not released yet, so yes, big API changes are to be expected - even if you organizsation is part of itowns you can't drive it toward your unique current purpose (globe)
- these events are limited: as I pointed out, they work only for one specific use case which happens to be *exactly* the only one you use: the geoportal-extensions only support the GlobeView as far as I can tell. As previously written, I'd prefer to remove them for now, and implement them properly (minimum requirement: they work regardless of the type of View and for all layers) later.
- keeping them just put some maintenance burden on iTowns. If we have some users depending on them, we'll have to keep them for a while, just because we didn't took the chance to rework them before the first release.

Anyway, if you feel so unhappy about the removal of these events now, I don't mind keeping them **for now**.
@autra ?

autra:
I believe I have understand you fairly well @gcebelieu. I know about your use case, I've seen your code. I still disagree with this. For instance, I consider that a lot of openlayers events are mistakes (for instance change:maxResolution on layer groups...).

>  I don't understand why two days before releasing date there is a sudden urge to remove them from API.

Well, considering I've been ignored on this point before, you can hardly call that a sudden urge, but ok fair point: the release is basically now, and it's too late. Let' move on.

gcebelieu:
Thanks @autra and @peppsac for this agreement on keeping these events as is for now.

PR #364, opened 2017-07-03
chore: rework examples and stop commit builds/example to this repo

peppsac:
gh-pages branch uses a lot of space, so moving this out to the organization
repository will help reduce the git repository size.

Use this PR to fix all examples path as well, so they work both after a git clone
and when used for the website.

@iTowns/reviewers r?

I'm not sure if the `deploy_key` can push to other repo in the `iTowns` org.
Maybe @tbroyer knows?

### DIFF
--- a/autodeploy.sh
+++ b/autodeploy.sh
@@ -9,7 +9,6 @@ if [ "$TRAVIS_NODE_VERSION" != "node" ]; then
 fi
 
 SOURCE_BRANCH="master"
-TARGET_BRANCH="gh-pages"
 
 if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
     echo "Skipping deploy for a pull request"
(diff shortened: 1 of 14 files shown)


autodeploy.sh
```diff
@@ -56,11 +64,11 @@ git config user.name "$COMMIT_AUTHOR_NAME"
 git config user.email "$COMMIT_AUTHOR_EMAIL"
 
 # Commit the "changes", i.e. the new version.
-git add dist index.html API_Doc examples dat.gui.min.js
+git add itowns2
 git commit -m "Deploy to GitHub Pages: ${SHA}"
```
autra: `Deploy from <repo>` ? As it would be a commit in itowns.github.io repo.

autodeploy.sh
```diff
@@ -56,11 +64,11 @@ git config user.name "$COMMIT_AUTHOR_NAME"
...
 
 chmod 600 deploy_key
 eval `ssh-agent -s`
 ssh-add deploy_key
```
autra: you're in `out/`, but you've copied the deploy_key in `out/itowns2`.

autra (approved):
Apart from the `deploy_key` path, LGTM. I didn't check in details all the path, but we'll be able to check the examples once deployed.

peppsac:
Fixed both.

autra (approved):
Let's see what this will do :-)

tbroyer:
I'm late to the party but I disagree with this change. This is moving the issue from a totally automated branch (that could be force-pushed) to a repo that contains handwritten content too.

If we don't want to deploy the dist files, we could include them in the npm package and use unpkg.org to "host" them; with some small automated editing of the examples HTML files to change the links during auto-deploy.

autra:
The goal is to have very few handwritten content in itowns.github.io. Btw, the clone of this repo is fast nowadays, but I didn't see a lot of contributions to it :-) 

So yeah, automatic doc from itowns2 (+ maybe a landing page or 2), or we can expect no doc at all (of value I mean). That was the logic of this change.

autra:
To answer more in details:

>  This is moving the issue from a totally automated branch (that could be force-pushed) to a repo that contains handwritten content too.

Yes, this is moving the issue from a repo where it is a real issue to a repo where we don't really care, thus mitigating the issue. 

I agree that the mix of handwritten and generated doc is not ideal. However, there are 2 things to take into account:
- the last activity in itowns.github.io is on november 2016....
- we are pushing (not even force-pushing) generated content to a single, well-identified folder (`itowns2/`). I'll document it in `itowns.github.io`.

> If we don't want to deploy the dist files, we could include them in the npm package and use unpkg.org to "host" them; with some small automated editing of the examples HTML files to change the links during auto-deploy.

So to avoid a simple push in a folder, you propose to rely on npm.org, unpkg.org and some sed command? Come on :-) Btw, that would necessitate to publish a package for each push on master. For me, that's a big no: I prefer automatic `git push` than automatic `npm publish`.

To answer your concern on the ML:

> If itowns.github.io contains manually-maintained files, then if we want good documentation we will care about clone performance.

The relevant documentation (the one that will get updated often) will be in itowns/itowns2. So yeah, that's exactly my point :-)

Anyway, I'm not saying that this is a golden solution, don't hesitate to push something if you have a better solution!

tbroyer:
> So to avoid a simple push in a folder, you propose to rely on npm.org, unpkg.org and some sed command? Come on :-) Btw, that would necessitate to publish a package for each push on master. For me, that's a big no: I prefer automatic `git push` than automatic `npm publish`.

Sorry, I thought we had an automated `npm publish` already. Otherwise (almost) agree with everything you said.

/me needs :coffee: and :sleeping_bed: :zzz:

PR #368, opened 2017-07-04
Fix: correct type when cleaning ssh key

autra:
Last fix for autodeploy :-)

@iTowns/reviewers please r?


peppsac (approved):
Looks good, again :)

PR #369, opened 2017-07-04
3d-tiles additive refinement fix

Jeremy-Gaillard:
Tiles were never hidden after they were loaded when using additive refinement.


peppsac (approved):
Thanks (please rebase)

Jeremy-Gaillard:
Rebased and ready.

PR #371, opened 2017-07-04
Fix: embed GLTFLoader to fix our ./lib folder

autra:
The loading of this GLTFLoader was done by a custom webpack
configuration on our side, and wasn't transpiled in the lib/ folder. As
B3dmLoader was using it, our lib/ folder was essentially broken.

### DIFF
--- a/src/Renderer/ThreeExtended/GLTFLoader.js
+++ b/src/Renderer/ThreeExtended/GLTFLoader.js
@@ -1,0 +1,2219 @@
+// This file is copy pasted from THREE
+/* CUSTOM ITOWNS */
+/* eslint-disable */
+import * as THREE from 'three';
+/* END CUSTOM ITOWNS */
+
+/**
+ * @author Rich Tibbett / https://github.com/richtr
+ * @author mrdoob / http://mrdoob.com/
+ * @author Tony Parisi / http://www.tonyparisi.com/
+ * @author Takahiro / https://github.com/takahirox
+ */
+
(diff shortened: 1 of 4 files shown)


test/itowns-testing.js
```diff
@@ -102,7 +103,7 @@ global.requestAnimationFrame = (cb) => {
 };
 
 // eslint-disable-next-line import/no-dynamic-require
-const itowns = require(`${process.env.PWD}/dist/itowns`);
+const itowns = require(`${process.env.PWD}/lib/Main.js`);
```
autra, line 106: This will allow to test that our `./lib` folder is not broken :-)

autra:
@iTowns/reviewers please r? I think this is necessary for the release.

src/Renderer/ThreeExtended/GLTFLoader.js
```diff
@@ -0,0 +1,2219 @@
+// This file is copy pasted from THREE
```
gchoqueux, line 1: There is no way to link or do a pre build to avoid copy/paste this file

autra: Not in an easy way unfortunately. It's a quick fix for the release. If I find something better, I'll do a PR after the release, but the most important is to have a npm package that works for this release. WDYT?

peppsac (approved):

autra:
@gchoqueux anything else here?

PR #379, opened 2017-07-07
fix: 3dTilesProcessing outdated notifyChange call

Jeremy-Gaillard:


src/Process/3dTilesProcessing.js
```diff
@@ -33,14 +33,14 @@ function subdivideNode(context, layer, node) {
                     node.add(tile);
                     tile.updateMatrixWorld();
                     if (node.additiveRefinement) {
```
autra, line 35: For my personal culture, why do we `notifyChange` only if node.additiveRefinement is true?

autra: Ok forget it. With the change line 43, it's clear.

autra (approved):

PR #381, opened 2017-07-10
fix (camera): use scale component when using matrixWorld

peppsac:
- 1st commit fixes Camera's method that were incorrectly ignoring scale
- ~2nd commit : add a 2d camera to allow the user to draw in pixel coords. This is something that could easily be done by a user, but we can save them the boilerplate code to do this so I think it belongs to itowns.~
- ~3rd commit : add new methods to camera, to compute contour and area of a Box3. As stated in the commit message we ideally would like to support any kind of geometry so this is only the first step.~

~As a quick demo of commits 2 and 3, here's a quick screenshot showing the contour (red) of OBB on top of its OBBHelper (green):~

export

~(I have other unmerged branches that really using these features)~

Edit: went back to my first PR that simply fix the scale issue.

@iTowns/reviewers r?

### DIFF
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -12,7 +12,6 @@ function Camera(width, height, options = {}) {
     this.camera3D = options.camera ? options.camera : new THREE.PerspectiveCamera(30, this.ratio);
 
     this._viewMatrix = new THREE.Matrix4();
-    this._visibilityTestingOffset = new THREE.Vector3();
     this.width = width;
     this.height = height;
 }
@@ -35,7 +34,9 @@ Camera.prototype.resize = function resize(width, height) {
         }
     }
 
-    this.camera3D.updateProjectionMatrix();
+    if (this.camera3D.updateProjectionMatrix) {
+        this.camera3D.updateProjectionMatrix();
+    }
 };
 
 Camera.prototype.update = function update() {
@@ -43,14 +44,7 @@ Camera.prototype.update = function update() {
     this.camera3D.updateMatrixWorld();
 
     // keep our visibility testing matrix ready
-    this._visibilityTestingOffset.setFromMatrixPosition(this.camera3D.matrixWorld);
-    const c = this.camera3D.matrixWorld;
-    // cancel out translation
-    c.setPosition({ x: 0, y: 0, z: 0 });
-    this._viewMatrix.getInverse(c);
-    this._viewMatrix.premultiply(this.camera3D.projectionMatrix);
-    // restore translation
-    c.setPosition(this._visibilityTestingOffset);
+    this._viewMatrix.multiplyMatrices(this.camera3D.projectionMatrix, this.camera3D.matrixWorldInverse);
 };
 
 Camera.prototype.getDistanceFromOrigin = function getDistanceFromOrigin() {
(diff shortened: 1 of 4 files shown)


nosy-b:
Do you have an example of use to justify the necessity to integrate the contour computation on screen with another package and another rendering pass by default (2nd&3rd commit)?

peppsac:
> Do you have an example of use to justify the necessity to integrate the contour computation on screen with another package

Compute the area on screen of a rectangular region.

If you use `Camera.box3DSizeOnScreen` the error can be big. e.g: twice as big for a square if the camera is in the diagonal axis (if my maths are correct...)

> another rendering pass by default

You misread the code I believe: I didn't add "another rendering pass" (in the common sense of the term, which is: redraw the whole scene from geometries / rendertarget).
I added a default empty scene and a second camera.
So by default, the cost is a call to `WebGLRenderer.render()` with an empty scene, which is almost free.
We could make it free by checking if `scene2D` has any children before calling `render`, but I'm not sure it's worth it.

The usage of the scene could be: it's sometimes easier to reason in term of screen coordinates, rather than 3D.
For instance, if you want to draw something (e.g text) on top of some object, without fighting the depth buffer or being affected by the perspective. 
Or if you want to draw some kind of GUI and you don't want to use the DOM for some reason. etc

src/Renderer/Camera.js
```diff
@@ -65,60 +80,119 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
+        points[7].set(tmp.box3d.max.x, tmp.box3d.max.y, tmp.box3d.max.z).applyMatrix4(tmp.matrix); // 111
+        // copy pasted / adapted from Box3.applyMatrix4 - end
+
+        for (const pt of points) {
+            // translate/scale to [0, width]x[0, height]
+            pt.x = this.width * (pt.x + 1) * 0.5;
+            pt.y = this.height * (1 - pt.y) * 0.5;
+            if (clampToScreen) {
+                pt.x = Math.min(this.width, Math.max(0, pt.x));
+                pt.y = Math.min(this.height, Math.max(0, pt.y));
```
mbredif: Are you sure you want to clamp each point in isolation before computing the convex hull ? The result will not be the same as the more intuitive clipping of the geometry.

src/Renderer/Camera.js
```diff
@@ -65,60 +80,119 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+    ];
+    return function box3DContourOnScreen(box3d, matrixWorld, clampToScreen = true) {
+        if (box3d.isEmpty()) {
```
gchoqueux: what do you think of this method?
http://jsfiddle.net/Eskel/g593q/5/

peppsac: The jsfiddle method is useful when you only need to render the contour of a solid object.

The `box3dContourOnScreen` returns the geometry of the contour - and when extended to handle any kind of geometry it will work with things like pointcloud for instance.

src/Renderer/Camera.js
peppsac: I'm not sure to understand the problem of clipping each point individually.
Can you show me a drawing of your suggestion to see how the result differs?

mbredif: Here is a slightly simpler clipping example with a triangle instead of a projected box. (image from yaldex.com)
triangle clipping 
Your code instead project orthogonally outside points to the screen boundary, which will result in the underestimated triangle (P5*, proj(P2,[P0*,P1*]), proj(P1,[P2*,P3*]), where proj(p,s) denotes the orthogonal projection of point p on segment s. Is it really what you want ?

src/Renderer/Camera.js
gchoqueux: yes, it's for polygonal object with normals. But, in the end, this method could still work. This is a comment for reflection.

On the other hand I doubt the performance of your method on large clouds of points unless you also use the oriented bounding box.

src/Renderer/Camera.js
peppsac: Ah ok I get it, and you're right: I need a to clip the geometry properly :)

gchoqueux:
can help : https://pdfs.semanticscholar.org/1f59/8266e387cf367702d16acf5a4e02cc72cb99.pdf

peppsac:
I'm going to remove the commit introducing `box3dContourOnScreen` from this PR since it has a few issues and I don't want to block the 2 others.

Any objections regarding the camera2D/scene2D stuff?

gchoqueux:
There's no example of using camera2D?

May be use in post processing example

gchoqueux:
may be use in post processing example

peppsac:
Removed the new feature commits.

src/Renderer/Camera.js
```diff
@@ -65,60 +65,66 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
     // shift the volumes3D toward origin
     tempContainer.copy(volumes3D);
-    tempContainer.translate(temp);
+    tempContainer.translate(tmp.v);
 
     // modify position: substract camera.position and add box3d.min
-    tempMatrix.elements[12] -= visibilityTestingOffset.x + temp.x;
-    tempMatrix.elements[13] -= visibilityTestingOffset.y + temp.y;
-    tempMatrix.elements[14] -= visibilityTestingOffset.z + temp.z;
+    tmp.matrix.elements[12] -= visibilityTestingOffset.x + tmp.v.x * tmp.scale.x;
```
gchoqueux: and rotate?

peppsac: Indeed. I'm going to see if I can simplify these functions.

peppsac:
Simplified all Camera functions. @gchoqueux r?

src/Renderer/Camera.js
```diff
@@ -65,60 +59,48 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
     tempContainer.copy(volumes3D);
-    tempContainer.translate(temp);
-
-    // modify position: substract camera.position and add box3d.min
-    tempMatrix.elements[12] -= visibilityTestingOffset.x + temp.x;
-    tempMatrix.elements[13] -= visibilityTestingOffset.y + temp.y;
-    tempMatrix.elements[14] -= visibilityTestingOffset.z + temp.z;
 }
 
 Camera.prototype.isBox3DVisible = function isBox3DVisible(box3d, matrixWorld) {
```
gchoqueux: you can write 

```js
Camera.prototype.isBox3DVisible = function isBox3DVisible(box3d, matrixWorld) {
    tmp.viewMatrix.multiplyMatrices(this._viewMatrix, matrixWorld);
    tmp.frustum.setFromMatrix(tmp.viewMatrix);
    return tmp.frustum.intersectsBox(box3d);
};
```

src/Renderer/Camera.js
```diff
@@ -65,60 +59,48 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
 
-    localViewMatrix.multiplyMatrices(this._viewMatrix, tempMatrix);
-    frustum.setFromMatrix(localViewMatrix);
-    return frustum.intersectsBox(tempBox3d);
+    tmp.viewMatrix.multiplyMatrices(this._viewMatrix, tmp.matrix);
+    tmp.frustum.setFromMatrix(tmp.viewMatrix);
+    return tmp.frustum.intersectsBox(tmp.box3d);
 };
 
 Camera.prototype.isSphereVisible = function isSphereVisible(sphere, matrixWorld) {
```
gchoqueux: you can write 

```js
Camera.prototype.isSphereVisible = function isSphereVisible(sphere, matrixWorld) {
    tmp.viewMatrix.multiplyMatrices(this._viewMatrix, matrixWorld);
    tmp.frustum.setFromMatrix(tmp.viewMatrix);
    return tmp.frustum.intersectsSphere(sphere);
};
```

src/Renderer/Camera.js
```diff
@@ -65,60 +59,48 @@ Camera.prototype.setRotation = function setRotation(rotation) {
     this.camera3D.quaternion.copy(rotation);
 };
 
-const temp = new THREE.Vector3();
-const frustum = new THREE.Frustum();
-const localViewMatrix = new THREE.Matrix4();
-const tempMatrix = new THREE.Matrix4();
-const tempBox3d = new THREE.Box3();
-const tempSphere3d = new THREE.Sphere();
+const tmp = {
```
gchoqueux: some temp properties can be deleted

src/Renderer/Camera.js
```diff
@@ -65,60 +59,48 @@ Camera.prototype.setRotation = function setRotation(rotation) {
...
-    return frustum.intersectsSphere(tempSphere3d);
+    tmp.viewMatrix.multiplyMatrices(this._viewMatrix, tmp.matrix);
+    tmp.frustum.setFromMatrix(tmp.viewMatrix);
+    return tmp.frustum.intersectsSphere(tmp.sphere3d);
 };
 
 Camera.prototype.box3DSizeOnScreen = function box3DSizeOnScreen(box3d, matrixWorld) {
-    _prepareMatrix(matrixWorld, this._visibilityTestingOffset, box3d, tempBox3d);
-    tempMatrix.premultiply(this._viewMatrix);
+    _prepareMatrix(matrixWorld, this._visibilityTestingOffset, box3d, tmp.box3d);
```
gchoqueux: may be remove totaly `_prepareMatrix`

gchoqueux (changes requested):

peppsac:
Branch updated.

src/Renderer/Camera.js
```diff
@@ -43,14 +43,8 @@ Camera.prototype.update = function update() {
     this.camera3D.updateMatrixWorld();
 
     // keep our visibility testing matrix ready
-    this._visibilityTestingOffset.setFromMatrixPosition(this.camera3D.matrixWorld);
-    const c = this.camera3D.matrixWorld;
-    // cancel out translation
-    c.setPosition({ x: 0, y: 0, z: 0 });
-    this._viewMatrix.getInverse(c);
+    this._viewMatrix.copy(this.camera3D.matrixWorldInverse);
```
gchoqueux: replace 
```js
    this._viewMatrix.copy(this.camera3D.matrixWorldInverse);
    this._viewMatrix.premultiply(this.camera3D.projectionMatrix);
```

by

```js
    this._viewMatrix.multiplyMatrices(this.camera3D.projectionMatrix, this.camera3D.matrixWorldInverse);
```

gchoqueux (changes requested):

NikoSaul:
This PR fixes a culling bug (objects not displayed when they should be) 
[Video showing the bug](https://drive.google.com/open?id=0Bxq_xC-O1pQxUEpMaWtFaWU3a00)
:+1:

src/Renderer/Camera.js
peppsac: Done.

gchoqueux (approved):

PR #383, opened 2017-07-10
fix (test) : process.env.PWD is undefined in Windows

gchoqueux:
`npm run test` fails on Windows because` process.env.PWD` is undefined

### DIFF
--- a/test/itowns-testing.js
+++ b/test/itowns-testing.js
@@ -102,6 +102,10 @@ global.requestAnimationFrame = (cb) => {
     }
 };
 
+// Fix: process.env.PWD is undefined for Windows
+if (!process.env.PWD) {
+    process.env.PWD = process.cwd();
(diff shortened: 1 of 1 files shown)


peppsac (approved):
Thanks for the fix 👍

test/itowns-testing.js
```diff
@@ -102,6 +102,10 @@ global.requestAnimationFrame = (cb) => {
     }
 };
 
+// Fix: process.env.PWD is undefined for Windows
```
tbroyer, line 105: I may be missing something but wouldn't the proper fix have been to replace the usage of `process.env.PWD` and use `process.cwd()` instead?

```js
const itowns = require(`${process.cwd()}/lib/Main.js`)
```

or is there a good reason to use `process.env.PWD` here?

PR #386, opened 2017-07-12
Support parent object that isn't a THREE.Scene

peppsac:
Allows the user to specify an object onto which the created Object3D from a Layer will be added (instead of being added to the root of the scene).

I started fixing all the places that assumed that the matrixWorld of their top most parent is an identity matrix.
There's still work to be done in Atmosphere shaders (it's using `cameraPosition`), GlobeControls (because it's moving the camera as if the globe is centered in (0, 0, 0) and if the globe size is constant) and probably many others.

@iTowns/reviewers r?

### DIFF
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -53,11 +53,22 @@ export const defineLayerProperty = function defineLayerProperty(layer, propertyN
     }
 };
 
-function GeometryLayer(i) {
+function GeometryLayer(id, object3d) {
+    if (!id) {
+        throw new Error('Missing id parameter (GeometryLayer must have a unique id defined)');
+    }
+    if (!object3d || !object3d.isObject3D) {
+        throw new Error('Missing/Invalid object3d parameter (must be a three.js Object3D instance)');
+    }
     this._attachedLayers = [];
 
+    Object.defineProperty(this, 'object3d', {
+        value: object3d,
+        writable: false,
+    });
+
     Object.defineProperty(this, 'id', {
-        value: i,
+        value: id,
         writable: false,
     });
 }
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -135,7 +135,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
         }
     }
 
-    const wgs84TileLayer = new GeometryLayer('globe');
+    const wgs84TileLayer = new GeometryLayer('globe', options.object3d || this.scene);
     const initLayer = initTiledGeometryLayer(globeSchemeTileWMTS(globeSchemeTile1));
     wgs84TileLayer.preUpdate = (context, layer, changeSources) => {
         SubdivisionControl.preUpdate(context, layer);
@@ -144,7 +144,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
         if (layer.level0Nodes === undefined) {
             initLayer(context, layer);
         }
-        preGlobeUpdate(context);
+        preGlobeUpdate(context, layer);
         if (changeSources.has(undefined) || changeSources.size == 0) {
             return layer.level0Nodes;
         }
@@ -203,7 +203,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     const sun = new THREE.DirectionalLight();
     sun.position.set(-0.5, 0, 1);
     sun.updateMatrixWorld(true);
-    this.scene.add(sun);
+    wgs84TileLayer.object3d.add(sun);
 
     this.addLayer(wgs84TileLayer);
 
@@ -216,7 +216,8 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     this.atmosphere.traverse((obj) => { obj.layers.set(atmosphereLayer); });
     this.camera.camera3D.layers.enable(atmosphereLayer);
 
-    this.scene.add(this.atmosphere);
+    wgs84TileLayer.object3d.add(this.atmosphere);
+    this.atmosphere.updateMatrixWorld(true);
 
 
     // Configure controls
@@ -232,11 +233,14 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     this._renderState = RendererConstant.FINAL;
 
     this.preRender = () => {
-        var len = this.camera.position().length();
-        var lim = size * 1.1;
+        const v = new THREE.Vector3();
+        v.setFromMatrixPosition(wgs84TileLayer.object3d.matrixWorld);
+        var len = v.distanceTo(this.camera.position());
+        v.setFromMatrixScale(wgs84TileLayer.object3d.matrixWorld);
+        var lim = v.x * size * 1.1;
 
         if (len < lim) {
-            var t = Math.pow(Math.cos((lim - len) / (lim - size * 0.9981) * Math.PI * 0.5), 1.5);
+            var t = Math.pow(Math.cos((lim - len) / (lim - v.x * size * 0.9981) * Math.PI * 0.5), 1.5);
             var color = new THREE.Color(0x93d5f8);
             this.mainLoop.gfxEngine.renderer.setClearColor(color.multiplyScalar(1.0 - t));
         } else if (len >= lim) {
--- a/src/Core/Scheduler/Providers/TileProvider.js
+++ b/src/Core/Scheduler/Providers/TileProvider.js
@@ -54,7 +54,7 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
     tile.layers.set(command.threejsLayer);
 
     if (parent) {
-        parent.worldToLocal(params.center);
+        params.center.sub(parent.geometry.center);
     }
 
     tile.position.copy(params.center);
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -26,10 +26,9 @@ function TileMesh(geometry, params) {
     this.geometry = geometry;
     this.normal = params.center.clone().normalize();
 
-    // TODO Why move sphere center
-    this.centerSphere = new THREE.Vector3().addVectors(this.geometry.boundingSphere.center, params.center);
+    this.boundingSphereOffset = new THREE.Vector3();
 
-    this.oSphere = new THREE.Sphere(this.centerSphere.clone(), this.geometry.boundingSphere.radius);
+    this.oSphere = new THREE.Sphere(this.geometry.boundingSphere.center.clone(), this.geometry.boundingSphere.radius);
 
     this.material = new LayeredMaterial(params.materialOptions);
 
@@ -79,6 +78,11 @@ TileMesh.prototype.dispose = function dispose() {
     this.geometry.dispose();
     this.geometry = null;
     this.material = null;
+};
+
+TileMesh.prototype.updateMatrixWorld = function updateMatrixWorld(force) {
+    THREE.Mesh.prototype.updateMatrixWorld.call(this, force);
+    this.geometry.OBB.update();
 };
 
 TileMesh.prototype.isVisible = function isVisible() {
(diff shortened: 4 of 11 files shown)


src/Core/Layer/Layer.js
```diff
@@ -53,11 +53,18 @@ export const defineLayerProperty = function defineLayerProperty(layer, propertyN
...
-function GeometryLayer(i) {
+function GeometryLayer(id, object) {
+    if (!id) {
+        throw new Error('Missing id parameter (GeometryLayer must have a unique id defined)');
+    }
+    if (!object || !object.isObject3D) {
+        throw new Error('Missing/Invalid object parameter (must be a three.js Object3D instance)');
+    }
     this._attachedLayers = [];
+    this.object = object;
```
autra: Maybe find a more explicit name for `this.object`?

src/Core/Layer/Layer.js
```diff
@@ -53,11 +53,18 @@ export const defineLayerProperty = function defineLayerProperty(layer, propertyN
     }
 };
 
-function GeometryLayer(i) {
+function GeometryLayer(id, object) {
+    if (!id) {
+        throw new Error('Missing id parameter (GeometryLayer must have a unique id defined)');
+    }
+    if (!object || !object.isObject3D) {
+        throw new Error('Missing/Invalid object parameter (must be a three.js Object3D instance)');
```
autra: What about not making object mandatory and defaulting  to `scene` when needed? It would be a bit more convenient for users (though less convenient for you right now), WDYT?

src/Core/Prefab/GlobeView.js
```diff
@@ -229,11 +230,14 @@ function GlobeView(viewerDiv, coordCarto, options) {
     this._renderState = RendererConstant.FINAL;
 
     this.preRender = () => {
-        var len = this.camera.position().length();
-        var lim = size * 1.1;
+        const v = new THREE.Vector3();
+        v.setFromMatrixPosition(wgs84TileLayer.object.matrixWorld);
+        var len = v.distanceTo(this.camera.position());
+        v.setFromMatrixScale(wgs84TileLayer.object.matrixWorld);
+        var lim = v.x * size * 1.1;
```
autra: Can you elaborate on `v.x`? A convenient approximation?

src/Core/Prefab/GlobeView.js
```diff
@@ -229,11 +230,14 @@ function GlobeView(viewerDiv, coordCarto, options) {
...
 
         if (len < lim) {
-            var t = Math.pow(Math.cos((lim - len) / (lim - size * 0.9981) * Math.PI * 0.5), 1.5);
+            var t = Math.pow(Math.cos((lim - len) / (lim - v.x * size * 0.9981) * Math.PI * 0.5), 1.5);
```
autra, line 243: I don't really understand this, so I'll trust you on this one.

src/Core/TileMesh.js
```diff
@@ -81,6 +81,11 @@ TileMesh.prototype.dispose = function dispose() {
     this.material = null;
 };
 
+TileMesh.prototype.updateMatrixWorld = function updateMatrixWorld(force) {
+    THREE.Mesh.prototype.updateMatrixWorld.call(this, force);
```
autra, line 84: maybe `super.updateMatrixWorld(force)` will work here.

autra (commented):
Some questions and suggestions, nothing very important :-)

src/Core/Layer/Layer.js
peppsac: Yeah, I tried a few alternatives (`root`, `anchor`), but none of them were really good so I went for the dumbest :)

The alternative is to transform the `GeometryLayer` in a `THREE.Object3D`, that could be added anywhere (another object, a scene, etc).

src/Core/Layer/Layer.js
peppsac: I don't have access to the `scene` here so I prefered to require the `object` param.

src/Core/Prefab/GlobeView.js
peppsac: The previous code did this:
```js
const size = ellipsoidSizes().x;
[...]
var lim = size * 1.1;
```
So the new code does the same computation, but it doesn't assume that:
1) the globe is in (0,0,0) (hence the need to use `distanceTo` instead of `position.length()`)
2) the ellipsoidSize is fixed (that's why we use `setFromMatrixScale`)

src/Core/Prefab/GlobeView.js
peppsac: `v` is a temp vector, and at this point it holds the `scale` of the object owning the globe.
So we scale the ellipsoid size approximation (`size`) when using it to compute `t`.

src/Core/Layer/Layer.js
autra: I think `object3D`, while not ideal, is a bit better. `parent` would work too I guess? `layerParent` ? `object3DAnchor`?

src/Core/Layer/Layer.js
autra: good for me.

autra (approved):

src/Core/Layer/Layer.js
gchoqueux: or transform `GeometryLayer ` in `THREE.Group`

src/Core/Prefab/GlobeView.js
```diff
@@ -198,7 +198,8 @@ function GlobeView(viewerDiv, coordCarto, options) {
     this.atmosphere.traverse((obj) => { obj.layers.set(atmosphereLayer); });
     this.camera.camera3D.layers.enable(atmosphereLayer);
 
-    this.scene.add(this.atmosphere);
+    wgs84TileLayer.object.add(this.atmosphere);
```
gchoqueux: It doesn't make sense to add the sun and atmosphere to `wgs84TileLayer`

gchoqueux (changes requested):

src/Core/Prefab/GlobeView.js
peppsac: I don't add them to the scene, they're added to the `object` where the `tiles` of `wgs84TileLayer` will be added.

peppsac: And I'm a bit reluctant to transform `GeometryLayer` in a `THREE.Group` for now.
No particular reason though, so we might revisit this decision later.

peppsac:
Updated.

Note: `GlobeControls` won't work if you use a different scale/position for the parent object of itowns

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -13,13 +13,15 @@
...
 
+import { Matrix4 } from 'three';
 import Provider from './Provider';
 import TileGeometry from '../../TileGeometry';
 import TileMesh from '../../TileMesh';
 import { CancelledCommandException } from '../Scheduler';
 
 function TileProvider() {
     Provider.call(this, null);
+    this.tempMatrix = new Matrix4();
```
autra: either this._tempMatrix, or better: make it a private module variable (`const tempMatrix` outside constructor).

autra (approved):
LGTM! (just one nits + the `super` things if you have nothing else to do).

@gchoqueux as you seem to be off, I leave you until tomorrow morning for eventual comments, then I'll merge.

src/Core/TileMesh.js
peppsac: Nope:
`'super' outside of function or class`

src/Core/Scheduler/Providers/TileProvider.js
peppsac: Done.

autra:
@gchoqueux any comments on this? I've said I'll merge it this morning, but as it seems to be important for you, I can wait till tonight.

gchoqueux:
i'll review this morning

gchoqueux:
Leave a resonable delay when you want to merge

src/Core/Prefab/GlobeView.js
```diff
@@ -193,7 +193,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     const sun = new THREE.DirectionalLight();
     sun.position.set(-0.5, 0, 1);
     sun.updateMatrixWorld(true);
-    this.scene.add(sun);
+    wgs84TileLayer.object3d.add(sun);
```
gchoqueux, line 206: The sun, atmosphere object doesn't really make sense in the layer wgs84TileLayer
It would be necessary to create a new geometric layer for these objects.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -54,7 +56,29 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
...
+        // then
+        //    N^-1 * M -> parent local to command.layer.object3d local
+        // then we inverse this:
+        //    (N^-1 * M)^-1 =
+        //        M^-1 * N -> command.layer.object3d local to parent local
+
+        // tempMatrix holds M^-1
+        tempMatrix.getInverse(parent.matrixWorld);
+        // tempMatrix now holds M^-1 * N
+        tempMatrix.multiply(command.layer.object3d.matrixWorld);
```
gchoqueux: Normally, `layer.object3d` transformation should already be inherited via children.

I'll delete `tempMatrix.multiply(command.layer.object3d.matrixWorld)`

gchoqueux (changes requested):
is it possible merge layer.level0Nodes and layer.object3d?

src/Core/Prefab/GlobeView.js
peppsac: I don't know so I won't fix this here.

src/Core/Scheduler/Providers/TileProvider.js
peppsac: I believe you misread the comment: it explains in lengthy details why it's needed.

TileGeometry and BuilderEllipsoidTile ignore completely the matrix transformation stuff, so they build as if the globe is always at (0,0,0), with no rotation nor scaled applied.
Hence the need to change the `parent.worldToLocal(...)` code.

peppsac:
> is it possible merge layer.level0Nodes and layer.object3d?

I can remove completely `level0Nodes` but then I'll need to replace all `layer.level0Nodes` with:
```js
layer.object3d.children.filter(n => n.layer == 'globe')
```

So while it's possible I don't think it's necessary.

src/Core/Prefab/GlobeView.js
gchoqueux: It is not necessary to change it now.
return to : 
`this.scene.add(sun);`

autra: For sun I don't know, but the atmosphere is linked to the ellipsoid right? It *must* be added to the same root object as the WGS84 layer imo.

peppsac: Yes, that's why I won't change it back to `this.scene.add`

```js
wgs84TileLayer.object3d.add(sun);
```
Means exactly what @autra said "add the sun to the same root object".
It doesn't mean: "add the sun to the wgs84 layer".

src/Process/GlobeTileProcessing.js
```diff
@@ -70,10 +72,13 @@ export function globeCulling(minLevelForHorizonCulling) {
 }
 
 function computeNodeSSE(camera, node) {
+    const v = new THREE.Vector3();
+    v.setFromMatrixScale(node.matrixWorld);
     const boundingSphere = node.geometry.boundingSphere;
     const distance = Math.max(
         0.0,
-        camera.camera3D.position.distanceTo(node.centerSphere) - boundingSphere.radius);
+        camera.camera3D.position.distanceTo(
```
gchoqueux: Why don't take into account the rotations and translations inherited by the parents?

gchoqueux: if layer.object3d have rotation node.centerSphere is incorrect

peppsac: It should probably be: `node.centerSphere.clone().applyMatrix4(node.matrixWorld)` here

(though I belive `centerSphere` and `oSphere` should both be removed...)

src/Core/Scheduler/Providers/TileProvider.js
gchoqueux: I may be a way to explore to simplify the code

if you replace  here https://github.com/iTowns/itowns/blob/master/src/Core/TileGeometry.js#L46

`this.center = builder.Center(params); `
 by
`this.center = builder.Center(params).clone(); `

so you can write

```js
if (parent) {
      params.center.sub(parent.geometry.center);
}
```

peppsac: It looks correct. Updating.

gchoqueux (approved):

Issue #387, opened 2017-07-12
Skirt size adaptation leaves holes


nosy-b:
Skirt size computation is linear to params level which induces bad cartesian size adaptation which result in holes:
![skirtbug](https://user-images.githubusercontent.com/158514/28122277-0a773f08-671e-11e7-9475-3e4a4035d828.jpg)

peppsac:
Closing since #388 is merged.

PR #388, opened 2017-07-12
fix: skirt size now depends on tile size

nosy-b:
It answers Issue #387
We now compute skirt size depending on the real tile size to keep same ratio at every zoom level to avoid holes and keep light fragment shading

### DIFF
--- a/src/Core/TileGeometry.js
+++ b/src/Core/TileGeometry.js
@@ -191,6 +191,9 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
             skirt = skirt.concat(verticesRow.slice().reverse());
         }
     }
+    // We compute the actual size of tile segment to use later for the skirt.
+    const segmentSize = new THREE.Vector3().fromArray(scratchBuffers.position).distanceTo(
+        new THREE.Vector3().fromArray(scratchBuffers.position, 3));
 
     if (!params.disableSkirt) {
         skirt = skirt.concat(skirtEnd.reverse());
@@ -222,12 +225,10 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
     var iStart = idVertex;
 
     // TODO: WARNING beware skirt's size influences performance
-    // Fix Me: Compute correct the skirt's size : minimize the size without crack between tiles
-    // This size must be take into account the bbox's size
-    // For the moment, I reduce the size to increase performance (pixel shader performance)
-
+    // The size of the skirt is now a ratio of the size of the tile.
+    // To be perfect it should depend on the real elevation delta but too heavy to compute
     if (!params.disableSkirt) {
-        const r = 5 * ((20 - params.level) + 10);
+        const r = segmentSize; // Tile size divided by the number of segments
 
         var buildIndexSkirt = function buildIndexSkirt() {};
         var buildUVSkirt = function buildUVSkirt() {};


src/Core/TileGeometry.js
```diff
@@ -222,12 +225,10 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
...
     // TODO: WARNING beware skirt's size influences performance
-    // Fix Me: Compute correct the skirt's size : minimize the size without crack between tiles
-    // This size must be take into account the bbox's size
-    // For the moment, I reduce the size to increase performance (pixel shader performance)
-
+    // The size of the skirt is now a ration of the size of the tile.
+    // To be perfect it should depend on the real elevation delta but too heavy to compute
     if (!params.disableSkirt) {
-        const r = 5 * ((20 - params.level) + 10);
+        const r = segmentSize; // Tile size divided by the number of segments
```
peppsac: How using `segmentSize` will be better than any other arbitrary formula?

IMHO the issue you're referencing is caused by the different subdivision level of 2 neighbors, so maybe we should also adjust the subdivision algorithm / SSE to prevent this problem.

src/Core/TileGeometry.js
```diff
@@ -222,12 +225,10 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
     var iStart = idVertex;
 
     // TODO: WARNING beware skirt's size influences performance
-    // Fix Me: Compute correct the skirt's size : minimize the size without crack between tiles
-    // This size must be take into account the bbox's size
-    // For the moment, I reduce the size to increase performance (pixel shader performance)
-
+    // The size of the skirt is now a ration of the size of the tile.
```
peppsac: Typo: ratio

src/Core/TileGeometry.js
```diff
@@ -146,13 +146,13 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
...
         var uv_pm = builder.getUV_PM ? builder.getUV_PM(params) : undefined;
 
+
         for (x = 0; x <= widthSegments; x++) {
             u = x / widthSegments;
 
             builder.uProjecte(u, params);
 
             var vertex = builder.VertexPosition(params, params.projected);
-
```
peppsac: Remove whitespaces changes.

src/Core/TileGeometry.js
```diff
@@ -146,13 +146,13 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
 
         var uv_pm = builder.getUV_PM ? builder.getUV_PM(params) : undefined;
 
+
```
peppsac: Remove whitespaces changes.

src/Core/TileGeometry.js
```diff
@@ -191,6 +191,9 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
             skirt = skirt.concat(verticesRow.slice().reverse());
         }
     }
+    // We compute the actual size of tile segment to use later for the skirt.
+    const segmentSize = new THREE.Vector3(scratchBuffers.position[0], scratchBuffers.position[1], scratchBuffers.position[2]).distanceTo(
```
peppsac:
```js
const segmentSize = new THREE.Vector3().fromArray(scratchBuffers.position).distanceTo(new THREE.Vector3().fromArray(scratchBuffers.position, 3));
```
?

peppsac (changes requested):
Left a few comments.

Could you also post a before/after screenshot, with tile outines enabled to "see" the change?

Thanks!

src/Core/TileGeometry.js
nosy-b: Using segment size allows the ration skirt size/tile size to stay the same at each level of the tree.
The previous simplistic formula doesn't do that and produce huge skirts (heavy frag shading) when close to the ground and very small when high in sky (creating holes)

nosy-b: (using only params.level (even squared) can't give you the real cartesian size of the wanted skirt)

nosy-b:
Two screenshots of before/after.
We can see that small holes are no more and also that having a constant Skirt/Tile size ratio avoid having supersized skirt (slowing a bit the rendering) when close to ground.
beforeafterskirtchange
beforeafterskirtchangewireframe

nosy-b:
@peppsac changes ok

peppsac (approved):
Thanks for the screenshots :)

PR #389, opened 2017-07-12
Remove ES6 syntax from examples

autra:
`examples/` are not processed by babel. Therefore they should be written only with ES5 syntax until we drop support from those older browsers.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,8 +57,11 @@ module.exports = {
 
     // TODO reactivate all the following rules
 
+    // maybe 'no-mixed-operators': ['error', { allowSamePrecedence: true }],
     'no-mixed-operators': 'off',
     'no-use-before-define': 'off',
+    // should probably be
+    // 'no-underscore-dangle': ['error', { allowAfterThis: true, allowAfterSuper: true }],
     'no-underscore-dangle': 'off',
     'eqeqeq': 'off',
     // what len ? Airbnb does 100. github wraps line above 80
--- a/examples/.eslintrc.js
+++ b/examples/.eslintrc.js
@@ -1,0 +1,61 @@
+module.exports = {
+    'extends': [
+        'eslint-config-airbnb-base',
+        'eslint-config-airbnb-base/rules/strict',
+    ],
+    parserOptions: {
+        ecmaVersion: 5,
+        sourceType: 'script',
+        ecmaFeatures: {
+            impliedStrict: true
+        }
+    },
+    env: {
+        browser: true,
+        es6: false,
+        amd: true,
+        commonjs: true
+    },
+    rules: {
+        'no-plusplus': 'off',
+        // this option sets a specific tab width for your code
+        // http://eslint.org/docs/rules/indent
+        indent: ['error', 4, {
+            SwitchCase: 1,
+            VariableDeclarator: 1,
+            outerIIFEBody: 1,
+            // MemberExpression: null,
+            // CallExpression: {
+            // parameters: null,
+            // },
+            FunctionDeclaration: {
+                parameters: 1,
+                body: 1
+            },
+            FunctionExpression: {
+                parameters: 1,
+                body: 1
+            }
+        }],
+        'one-var': ['error', 'never'],
+        'valid-jsdoc': ['error', {
+            requireReturn: false,
+            requireParamDescription: false,
+            requireReturnDescription: false,
+        }],
+
+        // deactivated rules for es5
+        'no-var': 'off',
+        'prefer-arrow-callback': 'off',
+        'object-shorthand': 'off',
+        'no-param-reassign': ['error', { 'props': false }],
+        'no-mixed-operators': ['error', { allowSamePrecedence: true }],
+
+        // deactivated rules for `examples/`
+        'no-console': 'off',
+
+        // TODO reactivate all the following rules
+        'no-underscore-dangle': 'off',
+
+    }
+}
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -42,18 +42,18 @@
         <script src="../dist/debug.js"></script>
         <script type="text/javascript">
             /* global itowns,document,GuiTools*/
-            const positionOnGlobe = { longitude: -75.61, latitude: 40.04, altitude: 50000 }
+            var positionOnGlobe = { longitude: -75.61, latitude: 40.04, altitude: 50000 }
             // iTowns namespace defined here
-            const viewerDiv = document.getElementById('viewerDiv');
+            var viewerDiv = document.getElementById('viewerDiv');
 
-            const globe = new itowns.GlobeView(viewerDiv, positionOnGlobe);
-            const menuGlobe = new GuiTools('menuDiv', globe, 300);
+            var globe = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var menuGlobe = new GuiTools('menuDiv', globe, 300);
 
-            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(result => globe.addLayer(result));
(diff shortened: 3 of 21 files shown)


examples/.eslintrc.js
```diff
@@ -0,0 +1,58 @@
...
+        // deactivated rules for es5
+        'no-var': 'off',
+        'prefer-arrow-callback': 'off',
+        'object-shorthand': 'off',
+        'no-param-reassign': ['error', { 'props': false }],
+        'no-mixed-operators': ['error', { allowSamePrecedence: true }],
+
+        // TODO reactivate all the following rules
+        'no-underscore-dangle': 'off',
+
```
tbroyer: There's no `"root": true` in this file, so it's merged with the _project global_ `.eslintrc.js`, which means those _overrides_ are useless, and the removed lines here do not _reactivate_ the corresponding rules.

We should probably go for simplicity here and simply override the `parserOptions` and `env`.

.eslintrc.js
```diff
@@ -57,8 +57,11 @@ module.exports = {
 
     // TODO reactivate all the following rules
 
+    // maybe 'no-mixed-operators': ['error', { allowSamePrecedence: true }],
```
tbroyer, line 60: Should this be in another PR?

examples/3dtiles.html
```diff
@@ -42,18 +42,18 @@
...
-            const menuGlobe = new GuiTools('menuDiv', globe, 300);
+            var globe = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var menuGlobe = new GuiTools('menuDiv', globe, 300);
 
             itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(result => globe.addLayer(result));
 
             // function use :
             // For preupdate Layer geomtry :
-            const preUpdateGeo = (context, layer) => {
+            var preUpdateGeo = (context, layer) => {
```
tbroyer: This lambda needs to be converted into a function.

examples/GUI/GuiTools.js
```diff
@@ -25,58 +25,64 @@ dat.GUI.prototype.hideFolder = function hideFolder(name, value) {
...
+        return seq.indexOf(a.id) < seq.indexOf(b.id);
+    });
+    for (i = 0; i < sortedLayers.length; i++) {
+        this.addImageryLayerGUI(sortedLayers[i]);
     }
 };
 
 GuiTools.prototype.addElevationLayersGUI = function addElevationLayersGUI(layers) {
-    for (var i = 0; i < layers.length; i++) {
+    var i;
```
tbroyer, line 84: Cannot the `var` stay into the `for`?

examples/globe.html
```diff
@@ -38,21 +38,21 @@
...
         <script src="globe.js"></script>
         <script type="text/javascript">
             /* global itowns, document, GuiTools, globeView, promises */
-            const menuGlobe = new GuiTools('menuDiv');
+            var menuGlobe = new GuiTools('menuDiv');
             menuGlobe.view = globeView;
             // Listen for globe full initialisation event
-            globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, () => {
+            globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 // eslint-disable-next-line no-console
```
tbroyer, line 51: The examples-specific `.eslintrc.js` could maybe disable the `no-console` rule?

webpack.config.js
```diff
@@ -5,6 +5,11 @@ var definePlugin = new webpack.DefinePlugin({
     __DEBUG__: JSON.stringify(process.env.NODE_ENV === 'development'),
 });
 
+var providePlugin = new webpack.ProvidePlugin({
```
tbroyer, line 8: Should we do the same for the promise and fetch polyfills? Or should this be changed to just a new entry in the `itowns` entry point?

utils/debug/Debug.js
```diff
@@ -8,7 +8,7 @@ import View from '../../src/Core/View';
 
 function createChartContainer(chartDiv, width, left, chartId) {
     const div = document.createElement('div');
-    div.style = `position: absolute; bottom: 0; left: ${left}vw; width: ${width}vw; height: 20rem; background-color: white;`;
+    div.style.cssText = `position: absolute; bottom: 0; left: ${left}vw; width: ${width}vw; height: 20rem; background-color: white;`;
```
tbroyer, line 11: Which browser does not support setting `style` directly?

autra: Some webkit-based, like midori or Epiphany. Safari is or maybe was affected at some point too, if I believe comments on the web (I have no way of testing though)

examples/.eslintrc.js
autra: >  those overrides are useless

They are not. Try to remove `prefer-arrow-callback` above.

> the removed lines here do not reactivate the corresponding rules.

They do. Try adding a very long line in a js file in examples.

The merging capability is apparently not as clever as you think it is (or I completely missed what you're saying). According to my test, the `rules` property of this file completely shadowed the `rules` property of the main `.eslintrc`.

> We should probably go for simplicity here and simply override the parserOptions and env.

Not enough, according to my tests. Eslint still complains about classic `function` callback until I disabled the `prefer-arrow-callback` above.

examples/GUI/GuiTools.js
autra: We have a linter rule for that. (`var-on-top` or something like this)

It's annoying, but useful too, as this `i` will get hoisted anyway.

examples/globe.html
autra: (If you ask me, `no-console` would be disabled everywhere)

Why not?

webpack.config.js
autra: I don't know. I haven't touched Promise and fetch because using itowns without either of them is quite hard, so it made sense to me to expose them in window. For TextEncoder and TextDecoder, it's only used internally so I didn't want to expose them needlessly.

autra: So, as I didn't know, I went for the status quo for this PR (already big and painful, I may say! ;-) )

examples/3dtiles.html
autra: Good catch!

examples/.eslintrc.js
tbroyer: I haven't had much experience with several eslintrc files, and mostly (if not only) redefining only the parser options and environment (and a much less strict ruleset), do I may just have misunderstood the docs or read too much into them.

examples/globe.html
tbroyer: To avoid leaving debug logs in production builds?

webpack.config.js
tbroyer: With this reasoning, maybe fetch could follow the same pattern, as it's not directly exposed. But let's not do that here anyway.

tbroyer (commented):
I believe there's a lambda to turn into a function, but otherwise consider this PR approved.

.eslintrc.js
autra: It's included here, because it matches with the rules in examples/.eslintrc.js

autra:
updated. Fixed a lambda catched by @tbroyer plus another one, and one little indentation mistake.

@iTowns/reviewers please r?

tbroyer (approved):

webpack.config.js
autra: I gave it a thought, and I actually think we shouldn't publicly polyfill *at all*, as we are not an app, but a module. So I think I'll use babel-runtime + babel-plugin-transform-runtime instead, or use polyfill and the ProvidePlugin (not sure which one is best yet).

Issue #87, opened 2016-06-16
Incorrect bounding box (OBB) size


peppsac:
Steps to reproduce:
- enable OBB display
- zoom on a moutain area

Result :
- for some tiles the bottom of the bounding box will be higher that the lowest point of the tile. 

Examples (see red squares):
![obb_size_bug](https://cloud.githubusercontent.com/assets/2198295/16108222/38bdd134-33a2-11e6-9e4d-cfe47974cec9.png)
- cause visibility issue because the culling code uses this OBB to determine visibility of a tile

Expected:
- the OBB should encompass the whole tile

gchoqueux:
Yes, the computing of bounding box is sampled in DEM texture. The function used is `TileMesh.prototype.parseBufferElevation = function(image,minMax,pitScale)`. 
The errors are small, this does not justify all the problems of culling.

peppsac:
Small or not, it will cause holes anyway (see https://github.com/iTowns/itowns2/issues/82#issuecomment-225939592 for an example).

Before filling this bug I tried various changes in the parseBufferElevation (including setting `inc` to 1) and it didn't make the problem go away.

Maybe it's an issue in the OBB construction? Alas the code in `BuilderEllipsoidTile.prototype.OBB` and `THREE.OBB.prototype.addHeight` isn't easy to grasp and lacks comments to document the maths being used.
Do you think you could take some time to document what is being done in these functions (and remove the `// TODO à vérifier --->`)?

peppsac:
I pushed a branch to add a unit test to identify issues related to OBB computation: https://github.com/iTowns/itowns/tree/test_geometry_obb

I'd like to merge it, but currently my test fails so either:
- my test is broken
- OBB computation is wrong sometimes

(ping @gchoqueux since you wrote the OBB code)

gchoqueux:
I'll look at it

PR #392, opened 2017-07-13
Frame requester and First person controls.

autra:
This PR adds the notion of FrameRequester: a module can subscribe to each request animation frame. Last commit adds a first person controls that uses this feature.

I preferred to make one PR because FPControls gives meaning to this FrameRequester, but I can split it if you want. Maybe a simpler example can be found to illustrate this.

@iTowns/reviewers please r?

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -83,7 +83,7 @@ export const GLOBE_VIEW_EVENTS = {
  * @param {object} coordCarto
  * @param {object=} options - see {@link View}
  */
-function GlobeView(viewerDiv, coordCarto, options) {
+function GlobeView(viewerDiv, coordCarto, options = {}) {
     THREE.Object3D.DefaultUp.set(0, 0, 1);
     const size = ellipsoidSizes().x;
     // Setup View
@@ -213,26 +213,11 @@ function GlobeView(viewerDiv, coordCarto, options) {
     const positionTargetCamera = positionCamera.clone();
     positionTargetCamera.setAltitude(0);
 
-    this.controls = new GlobeControls(
-        this,
-        positionTargetCamera.as('EPSG:4978').xyz(),
-        this.mainLoop.gfxEngine.renderer.domElement,
-        viewerDiv,
-        size,
-        this.getPickingPositionFromDepth.bind(this));
-    this.controls.rotateSpeed = 0.25;
-    this.controls.zoomSpeed = 2.0;
-    this.controls.minDistance = 30;
-    this.controls.maxDistance = size * 8.0;
-
-    this.controls.addEventListener('change', () => {
-        this.camera.update();
-        this.notifyChange(true);
-    });
-
-    this.controls.addEventListener('selectClick', (event) => {
-        this.selectNodeAt(event.mouse);
-    }, false);
+    if (options.noControls) {
+        this.camera.camera3D.lookAt(positionTargetCamera.as('EPSG:4978').xyz());
+    } else {
+        this.controls = new GlobeControls(this, positionTargetCamera.as('EPSG:4978').xyz(), size);
+    }
 
     this._renderState = RendererConstant.FINAL;
 
@@ -257,10 +242,6 @@ function GlobeView(viewerDiv, coordCarto, options) {
     };
 
     this.mainLoop.addEventListener('command-queue-empty', fn);
-
-    window.addEventListener('resize', () => {
-        this.controls.updateCamera(this.camera, viewerDiv.clientWidth, viewerDiv.clientHeight);
-    }, false);
 
     this.notifyChange(true);
 }
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -147,23 +147,31 @@ function GlobeView(viewerDiv, coordCarto, options) {
         }
         let commonAncestor;
         for (const source of changeSources.values()) {
-            if (!commonAncestor) {
-                commonAncestor = source;
-            } else {
-                commonAncestor = _commonAncestorLookup(commonAncestor, source);
-                if (!commonAncestor) {
-                    return layer.level0Nodes;
-                }
+            if (source.isCamera) {
+                // if the change is caused by a camera move, no need to bother
+                // to find common ancestor: we need to update the whole tree:
+                // some invisible tiles may now be visible
+                return layer.level0Nodes;
             }
-            if (commonAncestor.material == null) {
-                commonAncestor = undefined;
+            if (source.layer === layer.id) {
+                if (!commonAncestor) {
+                    commonAncestor = source;
+                } else {
+                    commonAncestor = _commonAncestorLookup(commonAncestor, source);
+                    if (!commonAncestor) {
+                        return layer.level0Nodes;
+                    }
+                }
+                if (commonAncestor.material == null) {
+                    commonAncestor = undefined;
+                }
             }
         }
         if (commonAncestor) {
             this._latestUpdateStartingLevel = commonAncestor.level;
             return [commonAncestor];
         } else {
-            return [];
+            return layer.level0Nodes;
         }
     };
 
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -14,7 +14,7 @@ import { GeometryLayer, Layer, defineLayerProperty } from './Layer/Layer';
 import Scheduler from './Scheduler/Scheduler';
 
 /**
- * Constructs an Itowns Scene instance
+ * Constructs an Itowns View instance
  *
  * @param {string} crs - The default CRS of Three.js coordinates. Should be a cartesian CRS.
  * @param {HTMLElement} viewerDiv - Where to instanciate the Three.js scene in the DOM
@@ -56,6 +56,7 @@ function View(crs, viewerDiv, options = {}) {
         this.mainLoop.gfxEngine.getWindowSize().y,
         options);
 
+    this._frameRequesters = [];
     this._layers = [];
 
     window.addEventListener('resize', () => {
@@ -276,6 +277,51 @@ View.prototype.getLayers = function getLayers(filter) {
         }
     }
     return result;
+};
+
+/**
+ * @typedef {object} FrameRequester
+ * @property {Function(dt, updateLoopRestarted)} update - Method that will be called each
+ * time the MainLoop updates. This function will be given as parameter the
+ * delta (in ms) between this update and the previous one, and whether or not
+ * we just started to render again. This update is considered as the "next"
+ * update if view.notifyChange was called during a precedent update. If
+ * view.notifyChange has been called by something else (other micro/macrotask,
+ * UI events etc...), then this update is considered as being the "first".
+ *
+ * This means that if a FrameRequester.update function wants to animate
+ * something, it should keep on calling view.notifyChange until its task is
+ * done.
+ *
+ * Implementors of FrameRequester.update should keep in mind that this function
+ * will be potentially called at each frame, thus care should be given about
+ * performance.
+ *
+ * Typical FrameRequesters are controls, module wanting to animate moves or UI
+ * elements etc... Basically anything that would want to call
+ * requestAnimationFrame.
+ */
+/**
+ * Add a frame requester to this view.
+ *
+ * FrameRequesters can activate the MainLoop update by calling view.notifyChange.
+ *
+ * @param {FrameRequester} frameRequester
+ * @param {Function} frameRequester.update - update will be called at each
+ * MainLoop update with the time delta between last update, or 0 if the
+ * MainLoop has just been relaunched.
+ */
+View.prototype.addFrameRequester = function addFrameRequester(frameRequester) {
+    this._frameRequesters.push(frameRequester);
+};
+
+/**
+ * Remove a frameRequester.
+ *
+ * @param {FrameRequester} frameRequester
+ */
+View.prototype.removeFrameRequester = function removeFrameRequester(frameRequester) {
+    this._frameRequesters.slice(this._frameRequesters.indexOf(frameRequester), 1);
 };
 
 export default View;
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -230,12 +230,6 @@ if (enableTargetHelper) {
 var _handlerMouseMove;
 var _handlerMouseUp;
 
-let _getPickingPosition;
-
-// Pseudo collision
-const radiusCollision = 50;
-
-
 // Event
 let enableEventPositionChanged = true;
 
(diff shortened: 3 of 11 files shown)


src/Renderer/c3DEngine.js
```diff
@@ -38,11 +38,13 @@ function c3DEngine(rendererOrDiv) {
         this.renderer.setSize(this.width, this.height);
     }.bind(this);
 
-    //
+    var canvas = document.createElement('canvas');
+    canvas.tabIndex = -1;
+    canvas.addEventListener('click', () => canvas.focus());
```
autra: A bit hacky, but I didn't find better :-)

src/Core/Prefab/GlobeView.js
```diff
@@ -84,6 +84,7 @@ export const GLOBE_VIEW_EVENTS = {
  * @param {object=} options - see {@link View}
  */
 function GlobeView(viewerDiv, coordCarto, options) {
+    options = options || {};
```
peppsac: Maybe use the default parameter syntax for GlobeView constructor instead?

src/Core/Prefab/GlobeView.js
```diff
@@ -205,26 +206,9 @@ function GlobeView(viewerDiv, coordCarto, options) {
     const positionTargetCamera = positionCamera.clone();
```
peppsac, line 213: This could be moved in the `if (!options.noControls) {` block I think.

And maybe you should always set the camera position to `coordCarto` - even if the user didn't ask for `GlobeControls` to be used?

src/Renderer/ThreeExtended/FPControls.js
```diff
@@ -0,0 +1,144 @@
...
+    if (delta < 0) {
+        this.moves.add(MOVEMENTS.wheelup);
+    } else {
+        this.moves.add(MOVEMENTS.wheeldown);
+    }
+
+    this.view.notifyChange(false);
+}
+
+class FPControls extends THREE.EventDispatcher {
```
peppsac: What about `FirstPersonControls` name?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -356,22 +350,19 @@ function defer() {
...
+ * @param {GlobeView} view
  * @param {*} target
- * @param {HTMLElement} domElement
- * @param {HTMLElement} viewerDiv
  * @param {number} radius
- * @param {function(THREE.Vector2):THREE.Vector2} getPickingPosition
+ * @param {options} options
  */
-function GlobeControls(view, target, domElement, viewerDiv, radius, getPickingPosition) {
+function GlobeControls(view, target, radius, options) {
```
peppsac:
```js
function GlobeControls(view, target, radius, options = {}) {
```
?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -554,7 +543,7 @@ function GlobeControls(view, target, domElement, viewerDiv, radius, getPickingPo
         } else if (this.camera instanceof THREE.OrthographicCamera) {
             this.camera.zoom = Math.max(this.minZoom, Math.min(this.maxZoom, this.camera.zoom * dollyScale));
             this.camera.updateProjectionMatrix();
-            this.dispatchEvent(this.changeEvent);
+            this._view.notifyChange(true);
```
peppsac: What about using: `this._view.notifyChange(true, this.camera);` to indicate to the update cycle that the change was caused by a camera movement? (same comments applies to FPControls obviously)?

src/Renderer/c3DEngine.js
peppsac: You could add a comment explaining why you did add this?

peppsac: Also, the canvas doesn't need to be created at all if `renderer` is supplied by the user.

peppsac (commented):
Thanks, this looks useful.

I quickly skimmed through the diff and left a few minor comments.
I'll probably do a follow-up  review next week.

src/Core/View.js
```diff
@@ -278,4 +279,49 @@ View.prototype.getLayers = function getLayers(filter) {
     return result;
 };
 
+/**
+ * @typedef {object} FrameRequester
+ * @property {Function(dt, firtLoop)} update - Method that will be called each
```
peppsac: typo: `firtLoop` => `firstLoop`. Maybe `updateLoopRestarted` is a better name too?

I'm still wondering if this boolean is really useful..

autra: What I don't like is that it's so easily messed up if controls (for instance) don't plug them well into this system. It relies on FrameRequesters to call `notifyChange` in the same microtask as `MainLoop._step` to work correctly (which GlobeControls does not yet). Otherwise, firstLoop will always be true.

Apart from that, I use it in `FPControls`, so it's at least *a bit* useful...

src/Renderer/c3DEngine.js
autra: The alternative would be to use viewerDiv instead of canvas for these events. WDYT?

src/Core/View.js
peppsac: The usage in `FPContols` could be moved in `MainLoop`:
```js
frameRequester.update(this._firstLoop ? 16 : dt);
```
But let's keep it for now. What do you think of my name proposal?

src/Renderer/c3DEngine.js
peppsac: I have no strong feeling regarding this.

Or maybe you could move this below the `renderer` creation and use `renderer.domElement` instead?

src/Core/View.js
autra: > What do you think of my name proposal?

I agree, ameding now.

> The usage in FPContols could be moved in MainLoop:

I don't think so, because now dt is the *real* interval between 2 requestAnimationFrame. So it's never undefined any more. I think it's less magic and more useful this way.

src/Renderer/c3DEngine.js
autra: I have a better idea: I think i'll add some options to define the focus policy in my FPControls instead. By default, it'll do nothing and let the client completely deal with focus, but it'll have a `focusOnClick` and `focusOnHover` options.

autra:
@peppsac updated. 

For the changeSource of `view.notifyChange`, there are some places where I didn't pass the camera, please check if it makes sense.

autra (commented):
EDIT: ignore, wrong tab :-)

autra:
For some reasons, my comments disappeared :-/ Let me do that again

autra:
@peppsac updated

peppsac (approved):
LGTM :)

gchoqueux (dismissed):
there's visibilty bug in globe

autra:
> there's visibilty bug in globe

As usual, I need you to be a bit more specific please :-)

On a more general basis, please be aware that this PR will uncover bugs, not necessarily directly caused by it, but just because GlobeView is not made to use anything else than GlobeControls right now. Can you reproduce using index.html?

gchoqueux:
I made a quick remark to alert, I'll come back soon

gchoqueux:
> Visibilty bug in globe

Means that there are tiles on the globe that are displayed or hidden when they should not

It is very simple to reproduce this bug:
zoom and de-zoom quickly when the camera pyramid intersects tiles

Warning: an invisible bug in the master isn't a bug

gchoqueux:
There's also an error message in `GlobeControls.updateCamera` :
Uncaught TypeError: Can not read property 'clientWidth' of undefined

peppsac:
@gchoqueux : you can be more helpful by following simple steps (here and in all your reviews):
- avoid 1 line comments ('it doesn't work')
- always specify steps to reproduce, or at the very least include a description of the issue (instead of waiting for the PR owner to ask for them)
- you can also do some bit of investigation. You're the top contributor of itowns, so you can probably guess sometimes where the culprit code is. E.g: 'if I remove these 2 lines from your changeset the bug disappear' etc

peppsac:
@gchoqueux @autra : I think the issue is caused by the cleanup of GlobeControls.

You replaced:
```js
-    this.controls.addEventListener('change', () => {
-        this.camera.update();
-        this.notifyChange(true);
-    });
```
(see  https://github.com/iTowns/itowns/pull/392/files#diff-caf5227be4f9e2f283d82df32a678e4bL220 )
with direct calls to `view.notifyChange()` in `GlobeControls`.

This is a good thing, but it uncovered a bug in `MainLoop`: the camera is updated *after* the step() function and *before* the render function.

I think a good fix would be to update the cameraat the beginning of step: this way the update will proceed with the camera at the correct position.

Moving this block at the beginning of `step` should be enough:
```js
    const dim = this.gfxEngine.getWindowSize();
    view.camera.resize(dim.x, dim.y);
    view.camera.update();
```

autra:
> Warning: an invisible bug in the master isn't a bug

I disagree on this. Will we close geoportail (or others, for that matter) bugs when we cannot reproduce on one of the html in `examples/`? Anyway, does not matter so much here, as I can reproduce it on master. So not a bug on this PR.

autra:
Beside, I'm not even sure what the bug is because I still don't have a gif or a video showing it.

Let me do some guesswork. Is it the blue triangle on top of the screen in this gif?

bug_maybe

Because if so, this is a clean master.

peppsac:
Your gif looks seriously like issue #87 
Could you display bounding-boxes to be sure?

autra:
My gif indeed looks like #87.

autra:
Added one commit fixing (hopefully) @gchoqueux's issue. Also amended `Feature: add first person controls` with a correct camera update in `FirstPersonControls.update` for the same reason. Thanks @peppsac for the explanations.

@peppsac @gchoqueux WDYT?

src/Core/Prefab/GlobeView.js
```diff
@@ -163,7 +163,7 @@ function GlobeView(viewerDiv, coordCarto, options) {
             this._latestUpdateStartingLevel = commonAncestor.level;
             return [commonAncestor];
         } else {
-            return [];
+            return layer.level0Nodes;
```
peppsac, line 174: Nice. But the purpose is defeated by https://github.com/iTowns/itowns/pull/392/files#diff-caf5227be4f9e2f283d82df32a678e4bL150

The above loop assumed that `changeSources` content were all `TileMesh`.
With this PR this is no longer the case (which is a good thing), but you need add a safeguard in the loop for this case.

Options:
- add `if (source.isCamera) { ... }` but this will break as soon as a non camera object is used
- add a `isTileMesh` property ?
- test for `isMesh` and `node.level` existence?

autra: I see. It seems that L158 saves me a little though (when sources does not have materials).

Actually, I think the best safeguard would be to test `source.layer === 'globe'`, because of line 137. WDYT?

autra:
Ok updated again. Please look especially at 9b7b96d.

autra:
Updated again.

peppsac:
LGTM now

autra:
@gchoqueux WDYT?

gchoqueux:
I'll do the review as soon as possible.

Otherwise first person should have a notion of gravity?

autra:
> Otherwise first person should have a notion of gravity?

Actually, maybe a better name for my control would be `FlyControls`.

autra:
To elaborate on this. I *do* have another controls that have a notion of verticality (if that's what you meant by "gravity"). At the moment, it works only for planar view (or other scene that assumes a planar world, like local pointcloud for instance). I need to make it able to change its vertical axe for the globe. I'll submit it later when this is done.

autra:
Just pushed a commit that renames it to `FlyControls`. I can squash it to the original commit if you want.

autra:
Ok so the commits are tidy and commit messages should be ok now. I've added a help div on the 2 example pages, renamed and changed title of globe_fp.html (now globe_fly.html

autra:
Updated: the commits are tidy and commit messages should be ok now. I've added a help div on the 2 example pages, renamed and changed title of globe_fp.html (now globe_fly.html).

peppsac:
Thanks for all the changes @autra 
Now it's time to move forward: I'm merging this PR.

gchoqueux:
Why did you merge before my review?

https://github.com/iTowns/itowns/pull/392#issuecomment-316155803

peppsac:
Because @autra answered all your comments and I have other branches waiting on this one to be merged.

And I didn't know when you'll be available again to review (knowing that you're not working 100% on iTowns currently). 

Feel free to contact me privately to talk about this.

Meanwhile we have some PR waiting for a review: #397, #396, #386 and #309

gchoqueux:
This https://github.com/iTowns/itowns/pull/392#issuecomment-316018782 has been disregarded. 
This bug is in the master when you resize the windows, but it isn't very important. On the other hand, it shows that to merge quickly and rejecting the review isn't the best method.

gchoqueux:
And if you want to know when I'm available : Ask me ;)

peppsac:
Where are the details of the issue you reported in this comment?
Sorry but it's again a "something is broken" comment - not helpful and easy to forget.

> And if you want to know when I'm available : Ask me ;)

Sure. But as you weren't on IRC we assumed we weren't working today.

gchoqueux:
> Where are the details of the issue you reported in this comment?

I guess if you read this review, you would have asked for details.

**There's also an error message in GlobeControls.updateCamera :
Uncaught TypeError: Can not read property 'clientWidth' of undefined**

Isn't detailed enough? What do not you understand?
This  `GlobeControls.updateCamera` has been modified in this PR?

> I didn't know when you'll be available again to review

Sorry I thought you did not know, So ask the question about when I'll be available

gchoqueux:
If you want to accelerate a PR, it's more appropriate to use this method : https://github.com/iTowns/itowns/pull/221#issuecomment-285635382

autra:
@gchoqueux 

> There's also an error message in GlobeControls.updateCamera :
Uncaught TypeError: Can not read property 'clientWidth' of undefined

I believe this is fixed (I tested both on chrome and Firefox). Please let me know if you still encounter this issue, I'll make a follow-up PR.

> Sorry I thought you did not know, So ask the question about when I'll be available

No, that's not how it works. It's not our responsibility to organize our work around your schedule. If you're off, you're off. If you're working, you're working. And if you're working, you are also connected to IRC, please (where you can ask us to wait if you want).

But you can't just come, put a "change requested" with a "wait for my review" like you did in other PR, then we have to wait for you to come back before anything can move. When @peppsac or I are on PTO, on other leave or on other projects, we don't request you to wait for us. Does this make sense to you?

> If you want to accelerate a PR, it's more appropriate to use this method : #221 (comment)

And I didn't put a "wait for my review" there because I wasn't working on itowns at that point, even though I wouldn't have merged this.

Here, @peppsac only chose to merge because your review didn't express a disagreement in the principle of it. In short, you said nothing that can't be addressed in minor follow-up PRs. Again feel free to open issues or ping me on IRC if you want such follow-ups.

autra:
To follow-up

> This #392 (comment) has been disregarded.

I believe it hasn't, but I'm double-checking right now.

autra:
To follow-up again @gchoqueux, and to elaborate on why you should always give *precise* step to reproduce:

There was actually 2 situations where this error fired: 
- one at startup, that is fixed
- one at resize, still there indeed

You were speaking about the one when resizing, but only gave the information it was so *yesterday*. So :

> There's also an error message in GlobeControls.updateCamera :
> Uncaught TypeError: Can not read property 'clientWidth' of undefined
> 
> Isn't detailed enough? What do not you understand?

No, an error in the console is *not* "detailed enough". It's useless without steps to reproduce. And also:

> I guess if you read this review, you would have asked for details.

That's again not how it works. We are not here to do a guessing game. You need to provide the details, and especially the steps to reproduce. This will guarantee a better experience for your reviewee.

A must-read in bug reporting https://developer.mozilla.org/en-US/docs/Mozilla/QA/Bug_writing_guidelines#Writing_a_clear_summary

And I quote:

> Steps to reproduce are the **most important part of any bug report**. If a developer is able to reproduce the bug, the bug is very likely to be fixed. If the steps are unclear, it might not even be possible to know whether the bug has been fixed.

PR #394, opened 2017-07-17
Fix: correctly prepare lib/ for running test

autra:
We now run the test using directly `lib/Main.js`, so it doesn't make sens to run `npm run build` before testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,11 @@
     "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js",
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint && npm run build && npm run test-examples",
-    "test-examples": "mocha test/globe_test.js && mocha test/planar_test.js && mocha test/postprocessing_test.js && mocha test/externalscene_test.js",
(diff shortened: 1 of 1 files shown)


autra:
@iTowns/reviewers please r?

package.json
```diff
@@ -7,11 +7,12 @@
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js",
     "doclint": "npm run doc -- -t templates/silent",
-    "test": "npm run lint && npm run build && npm run test-examples",
+    "test": "npm run lint && npm run transpile && npm run test-examples",
```
tbroyer: We may want to modify `.travis.yml` to run `npm run build` in addition to `npm run test` then.

I'd rather keep the `npm run build` here, and add `npm run transpile` to the list (either here or within the `test-examples` script instead)

package.json
```diff
@@ -7,11 +7,12 @@
...
     "test-examples": "mocha test/globe_test.js && mocha test/planar_test.js && mocha test/postprocessing_test.js && mocha test/externalscene_test.js",
     "build": "webpack -p",
+    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
```
tbroyer: :+1:

autra:
@iTowns/reviewers updated

tbroyer (approved):

PR #396, opened 2017-07-19
feature: add new functions to prevent subdivisions when textures are missing

peppsac:
Avoid the following issues:
- in planar.html: subdivision happens before texture fetches. The visual experience is that each small tile gets its elevation/color texture one after another, instead of refining. Example on planar.html:
b

- if you start with the camera at a zoom level higher than allowed by the WMTS server, the visible tiles will never gets a texture since:
  * tile level is > to WMTS max level -> no texture for tile.
  * tile's parent are hidden (because subdisvision) -> no texture fetched for parent. Even if parent level is <= WMTS max level and then this parent's texture could be used by its displayed children  
Example on 3dtiles.html (when lowering initial altitude to 5000):

  a

### DIFF
--- a/src/Process/SubdivisionControl.js
+++ b/src/Process/SubdivisionControl.js
@@ -1,0 +1,27 @@
+export default {
+    preUpdate: (context, layer) => {
+        context.colorLayers = context.view.getLayers(
+            (l, a) => a && a.id == layer.id && l.type == 'color');
+        context.elevationLayers = context.view.getLayers(
+            (l, a) => a && a.id == layer.id && l.type == 'elevation');
+    },
+
+    hasEnoughTexturesToSubdivide: (context, layer, node) => {
+        // Prevent subdivision if node is covered by at least one elevation layer
+        // and if node doesn't have a elevation texture yet.
+        for (const e of context.elevationLayers) {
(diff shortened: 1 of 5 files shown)


src/Process/SubdivisionControl.js
```diff
@@ -0,0 +1,27 @@
...
+    hasEnoughTexturesToSubdivide: (context, layer, node) => {
+        // Prevent subdivision if node is covered by at least one elevation layer
+        // and if node doesn't have a elevation texture yet.
+        for (const e of context.elevationLayers) {
+            if (e.tileInsideLimit(node, e) && !node.isElevationLayerLoaded()) {
+                return false;
+            }
+        }
+
+        // Prevent subdivision if missing color texture
```
autra, line 18: should we still subdivise if we have an elevation texture, even if we don't have color textures yet?

peppsac: Currently I'm requiring that all layers have a texture. Maybe we can adjust this in the future or split the conditions in 2 separate functions.

autra (approved):
Greatly improve the experience when we start itowns with a low altitude, thanks!

PR #397, opened 2017-07-19
Add new example landing page

peppsac:
This commit adds a landing page showing all examples.

I'm sure it can be made prettier but the first questions are: do we want such a page? or do we want a similar page but using screenshots instead of interactive iframes?

WDYT?


examples/3dtiles.html
```diff
@@ -26,6 +26,11 @@
                 top:0px;
                 margin-left: 0px;
             }
+            @media (max-width: 600px) {
+                #menuDiv {
+                    display: none;
+                }
+            }
```
autra, line 33: ♥

autra (commented):
> the first questions are: do we want such a page? or do we want a similar page but using screenshots instead of interactive iframes?

Yes we want such a page, definitely. 

That being said, interactive iframes won't scale. Maybe we can have the same approach as [on itowns' website](http://www.itowns-project.org/#demo) (a play button). That way, the user will choose how many itowns will run on the same page (and will probably stop before their computer blow off!). I'm pretty sure we can auto-generate screenshots from travis with a headless browser one day, but I'm fine to have black rectangles for now. 

> I'm sure it can be made prettier

Yes :-)

peppsac:
Ok, I'll replace auto-on iframes by iframes + play button and maybe a screenshot.

peppsac:
Done.
Feel free to push commits on this branch to improve it - I'll do a rebase/squash when we're happy with the result.

examples/index.html
```diff
@@ -19,37 +19,37 @@
         (click on title to open fullscreen)
         </p>
         <div class="exampleContainer">
-            <h3><a href='./orthographic.html'>Orthographic</a></h3>
+            <h3><a href='./orthographic.html' target='_blank'>Orthographic</a></h3>
```
autra: Not really important, but I prefer the old behaviour (without `target='_blank'`) as it allows both behaviour (normal click, and middle click or ctrl+click). We can let user choose :-)

peppsac:
Another alternative would be to generate this page using jekyll and the itowns.github.io template.

autra:
> Another alternative would be to generate this page using jekyll and the itowns.github.io template.

yep. I'll have a go when I have bandwidth if you want. In the meantime, this is better than nothing, and I can link this page from itowns.github.io. So I think we can merge this :-)

autra (approved):

examples/index.html
nosy-b: Yes it is better for nerds but many people are not ;)

nosy-b:
@peppsac Yes it should use the itowns.github.io template because it will be accessible from it

examples/planar.html
```diff
@@ -43,6 +43,13 @@
             #help > ul {
                 padding: 0 1rem;
             }
+
+            @media (max-width: 600px) {
+                #help {
```
autra, line 48: maybe also hide the controls?

autra:
I've just rebased it. I suggest to merge it as it is for now. Later I'll try to have a better solution (an auto-generated one for instance).

Imo an ideal solution would be:
- with example lists auto-generated
- with screenshots auto-generated (a bit harder)
- integrated with jekyll

PR #400, opened 2017-07-20
feature: add point cloud support

peppsac:
This PR adds basic support for pointclouds in itowns without using Potree.

What's working:
- support for PotreeConverter ouput
- support for https://github.com/Oslandia/lopocs point cloud server
- edl rendering
- various debug options
- picking

What's not done:
- refactor 3dtiles point support to be able to share more code
- and probably 100s of other things...

I've pushed a new dataset https://github.com/iTowns/iTowns2-sample-data/tree/master/pointclouds/eglise_saint_blaise_arles (thanks to Maxime Seguin).

Important note: this code is still WIP and is likely to move quickly, so there's no need to be over pedantic during reviews.

Screenshot:

capture d ecran de 2017-07-20 17-03-53


examples/pointcloud.html
```diff
@@ -0,0 +1,157 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Point Cloud Viewer</title>
+
+        <style type="text/css">
+            html {height: 100%;}
+            body { margin: 0; overflow:hidden; height:100%}
```
autra: nits: formatting

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
+import * as THREE from 'three';
+import Provider from './Provider';
+import Fetcher from './Fetcher';
+import PointsMaterial from '../../../Renderer/PointsMaterial';
+
+function PointCloudProvider() {
```
autra: you can use class syntax here

```js
class PointCloudProvider extends Provider {
  constructor() {
      super();
  } 

  preprocessDataLayer() { ... }
  
  // ...
}

File: src/Core/View.js [hunk 3]
```diff
@@ -44,7 +44,7 @@ function View(crs, viewerDiv, options = {}) {
 
     this.referenceCrs = crs;
 
-    this.mainLoop = options.mainLoop || new MainLoop(new Scheduler(), new c3DEngine(options.renderer || viewerDiv));
+    this.mainLoop = options.mainLoop || new MainLoop(new Scheduler(), new c3DEngine((options.renderer && options.renderer.domElement) ? options.renderer : viewerDiv, options.renderer));
```
autra: The last argument to c3Engine should be options I guess. 

This line is a bit hard to parse :-)

src/Renderer/PointsMaterial.js
```diff
@@ -0,0 +1,36 @@
...
+    // this.uniforms.color = { type: 'v3', value: '0xFFFFFF' };
+    this.uniforms.size = new Uniform(size);
+    this.uniforms.resolution = new Uniform(new Vector2(window.innerWidth, window.innerHeight));
+    this.uniforms.pickingMode = new Uniform(false);
+    this.uniforms.density = new Uniform(0.01);
+
+    if (__DEBUG__) {
+        this.uniforms.useDebugColor = new Uniform(false);
+        this.uniforms.debugColor = new Uniform(new Color());
+        this.defines = {
```
autra: Can we add a define instead of replacing everything?

src/Renderer/PointsMaterial.js
```diff
@@ -0,0 +1,36 @@
...
+            DEBUG: 1,
+        };
+    }
+}
+
+PointsMaterial.prototype = Object.create(RawShaderMaterial.prototype);
+PointsMaterial.prototype.constructor = PointsMaterial;
```
autra: Class syntax ?

src/Renderer/c3DEngine.js
```diff
@@ -38,12 +38,23 @@ function c3DEngine(rendererOrDiv) {
...
+    // pick default options
+    if (options.antialias === undefined) {
+        options.antialias = true;
+    }
+    if (options.alpha === undefined) {
+        options.alpha = true;
+    }
+    if (options.logarithmicDepthBuffer === undefined) {
+        options.logarithmicDepthBuffer = this.gLDebug || NOIE;
+    }
```
autra: I would do

```js
options.antialias = options.antialias || true;
// ...
```

just after var NOIE on line 13. It's better to initialize default values before doing anything else.

examples/pointcloud.html
autra: I have actually pushed some reformatting of this file here: https://github.com/autra/itowns/commit/980215607c98808c53a2048d8f891720738ce807 Feel free to squash it here.

examples/pointcloud.js
```diff
@@ -0,0 +1,167 @@
+/* global itowns, pointcloud, GuiTools, view, menuGlobe, debug */
```
autra: generally, we should write valid es5 here.

autra: (So no const, no let, no arrow functions...)

autra: Also, if we auto-generate tuto from examples page, we would need to comment this file. Feel free to do it here or not though.

examples/pointcloud.js
```diff
@@ -0,0 +1,167 @@
...
+    document.getElementsByTagName('body').className = '';
+
+    const viewerDiv = document.getElementById('viewerDiv');
+
+    itowns.THREE.Object3D.DefaultUp.set(0, 0, 1);
+
+    itowns.proj4.defs('EPSG:3946',
+        '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+    window.menuGlobe = new GuiTools('menuDiv');
```
autra: We should either pass a bigger width, or not display full urls in pointcloud debug section.

examples/pointcloud.js
```diff
@@ -0,0 +1,167 @@
...
+        const buffer = view.mainLoop.gfxEngine.renderViewTobuffer(
+                view, view.mainLoop.gfxEngine.fullSizeRenderTarget,
+                mouse.x, mouse.y, 1, 1);
+
+        // see PointCloudProvider and the construction of unique_id
+        const objId = (buffer[0] << 8) | buffer[1];
+        const index = (buffer[2] << 8) | buffer[3];
+
+        const i = (objId << 16) | index;
+        // eslint-disable-next-line no-console
```
autra: in examples/ the no-console rule is disabled.

examples/pointcloud.js
```diff
@@ -0,0 +1,167 @@
...
+
+                o.material.enablePicking(false);
+            }
+        });
+
+        selectionGroup.visible = true;
+        view.notifyChange(true);
+    });
+
+    const flyControls = new itowns.FlyControls(view, { focusOnClick: true });
```
autra: Note to myself: submit my other control that preserve vertical (better for pointcloud)

examples/pointcloud.js
```diff
@@ -0,0 +1,167 @@
...
+
+
+    function placeCamera(position, lookAt) {
+        view.camera.camera3D.position.set(position.x, position.y, position.z);
+        view.camera.camera3D.lookAt(lookAt);
+        view.camera.update();
+    }
+
+    view.addLayer(pointcloud).addEventListener('preprocessed', () => {
+        // eslint-disable-next-line no-new
```
autra: Do you really need that?

autra: I guess we should disable it in examples/ then...

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
...
+    Provider.call(this);
+}
+
+PointCloudProvider.prototype = Object.create(Provider.prototype);
+PointCloudProvider.prototype.constructor = PointCloudProvider;
+
+// PointCloud
+// from potree
+function createChildAABB(aabb, childIndex) {
```
autra: What is AABB?

autra: What's the purpose of this method?

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
...
+        colors[4 * i + 1] = view.getUint8(offset + 13);
+        colors[4 * i + 2] = view.getUint8(offset + 14);
+        colors[4 * i + 3] = 255;
+
+        offset += 16;
+    }
+    return [positions, colors];
+}
+
+function readPositionAndColorCustomBinFormat(ab, numPoints) {
```
autra: Somewhere, even though it is quite obvious, we need an explanation of what this custom bin format is, and how to generate it (link to your PotreeConverter branch ?).

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
...
+                    stack.push(item);
+                }
+            }
+        }
+
+        return root;
+    });
+}
+
+function readPositionAndColorBinFormat(ab, numPoints) {
```
autra: either `PotreeBin` or a comment saying it is the potree bin format?

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
...
+        layer.file = 'cloud.js';
+    }
+    layer.fetchOptions = layer.fetchOptions || {};
+
+    Fetcher.json(`${layer.url}/${layer.file}`, layer.fetchOptions).then((cloud) => {
+        layer.metadata = cloud;
+
+        let bbox;
+
+        if (layer.metadata.scale) {
```
autra: Here the presence of scale seems to indicate the format. Shouldn't it be the other way around? The format is specified and *then* we apply custom scale logic according to it?

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
...
+
+        parseOctree(
+                layer,
+                layer.metadata.hierarchyStepSize,
+                { baseurl: `${layer.url}/${cloud.octreeDir}/r`, name: '', bbox })
+        .then((root) => {
+            // eslint-disable-next-line no-console
+            console.log('LAYER metadata:', root);
+            layer.root = root;
+            layer.dispatchEvent({ type: 'preprocessed' });
```
autra: I prefer to have a preprocessed promise attached to the layer object, because if you arrive too late to attach a listener, you can still do `layer.preprocessedPromise.then(...)`.

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,282 @@
...
+        });
+    });
+};
+
+
+PointCloudProvider.prototype.executeCommand = function executeCommand(command) {
+    const layer = command.layer;
+    var node = command.requester;
```
autra: nits const or let

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,291 @@
...
+        new THREE.Vector3(h.x, -h.y, h.z),
+        new THREE.Vector3(-h.x, -h.y, h.z),
+        new THREE.Vector3(-h.x, -h.y, -h.z),
+        new THREE.Vector3(-h.x, -h.y, h.z),
+        new THREE.Vector3(-h.x, h.y, -h.z),
+        new THREE.Vector3(-h.x, h.y, h.z),
+        new THREE.Vector3(h.x, h.y, -h.z),
+        new THREE.Vector3(h.x, h.y, h.z),
+        new THREE.Vector3(h.x, -h.y, -h.z),
+        new THREE.Vector3(h.x, -h.y, h.z));
```
autra, line 33: Can you explain this to me? It's called cube, but it has 3 times the amount of vertices of a cube, with duplicates... I feel I'm missing something :-)

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,291 @@
...
+                        0, Math.floor(shouldBeLoaded * elt.obj.geometry.attributes.position.count));
+                    layer.counters.pointCount += elt.obj.realPointCount;
+                    layer.counters.displayedCount += Math.floor(shouldBeLoaded * elt.obj.geometry.attributes.position.count);
+                    elt.obj.material.uniforms.size.value = layer.pointSize;
+                } else if (!elt.promise) {
+                    elt.promise = context.scheduler.execute({
+                        layer,
+                        requester: elt,
+                        view: context.view,
+                        priority: 1.0 / elt.name.length, // surfaceOnScreen,
```
autra, line 231: Have you tried with `surfaceOnScreen`?

autra: Actually, surfaceOnScreen produces more "hole" during the initial loading, but if you move close to the door, it's a lot better than 1.0 / elt.name.length (which seems to take *ages* to do anything). It makes sense I guess, as the former should produce something near a pre-order walk,  while the latter is more post-order.

I guess it would help to be able to cancel requests maybe?

autra: (I'm not really requesting anything here, just thinking out loud).

I've tried `surfaceOnScreen / elt.name.length`, but it seems to behave like `surfaceOnScreen`.

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,291 @@
...
+            // not visible
+            if (!elt.notVisibleSince) {
+                elt.notVisibleSince = Date.now();
+            }
+        }
+
+        if (shouldBeLoaded >= 0.9 && elt.children && elt.children.length) {
+            return elt.children;
+        }
+        return undefined;
```
autra, line 263: return [] ?

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,291 @@
...
+                if (pts.material.visible) {
+                    pts.geometry.setDrawRange(0, pts.geometry.drawRange.count * reduction);
+                }
+            }
+            layer.counters.displayedCount *= reduction;
+        }
+
+        const now = Date.now();
+
+        for (let i = 0; i < layer.group.children.length; i++) {
```
autra: I find inversed while loop more natural when you remove elements inside the loop. It avoids the need for the lost-between-curly-braces `i--` on line 288.

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,291 @@
...
+                                layer.bboxes.children.splice(j, 1);
+                                obj.boxHelper.material.dispose();
+                                obj.boxHelper.geometry.dispose();
+                                obj.boxHelper = null;
+                                break;
+                            }
+                        }
+                    }
+                }
+                i--;
```
autra: It looks so sad alone here!

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,291 @@
...
+                obj.geometry.dispose();
+                obj.material = null;
+                obj.geometry = null;
+                obj.owner.obj = null;
+
+                if (__DEBUG__) {
+                    if (obj.boxHelper) {
+                        for (let j = 0; j < layer.bboxes.children.length; j++) {
+                            if (layer.bboxes.children[j].id == obj.boxHelper.id) {
+                                layer.bboxes.children.splice(j, 1);
```
autra: Won't this mess your loop?

src/Renderer/Shader/EDLPassOneFS.glsl
```diff
@@ -0,0 +1,83 @@
...
+uniform float cameraNear;
+uniform float cameraFar;
+
+uniform int n;
+uniform int directions;
+uniform float radius;
+
+varying vec2 vUv;
+
+#if 1
```
autra: ?

src/Renderer/Shader/PointsFS.glsl
```diff
@@ -0,0 +1,32 @@
...
+varying vec4 vColor;
+uniform bool pickingMode;
+
+#ifdef DEBUG
+uniform bool useDebugColor;
+uniform vec3 debugColor;
+#endif
+
+void main() {
+    #if 1
```
autra: Did you want an option here?

src/Renderer/Shader/PointsFS.glsl
```diff
@@ -0,0 +1,32 @@
...
+    #endif
+
+    gl_FragColor = vColor;
+#ifdef DEBUG
+    if (useDebugColor && !pickingMode) {
+        gl_FragColor = mix(vColor, vec4(debugColor, 1.0), 0.5);
+    }
+#endif
+
+#if defined(USE_LOGDEPTHBUF) && defined(USE_LOGDEPTHBUF_EXT)
```
autra: not really important, but we could import `src/renderers/shaders/ShaderChunk/logdepthbuf_fragment.glsl` from three?

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+        float pointSize = 1.0;
+        float slope = tan(1.0 / 2.0);
+        float projFactor =  -0.5 * resolution.y / (slope * mvPosition.z);
+
+        float z = min(0.5 * -gl_Position.z / gl_Position.w, 1.0);
+        gl_PointSize = max(3.0, min(10.0, 0.05 * projFactor));// min(10.0, 3.0 / (z * z)); //min(8.0, 0.05 * (resolution.y * 0.5 / - mvPosition.z));
+    }
+
+#ifdef USE_LOGDEPTHBUF
```
autra: use `logdepthbuf_vertex.glsl` from three?

utils/debug/PointCloudDebug.js
```diff
@@ -0,0 +1,192 @@
...
+    if (elt.name.length > hierarchyNode.length) {
+        return elt.name.startsWith(hierarchyNode);
+    } else if (elt.name.length < hierarchyNode.length) {
+        return hierarchyNode.startsWith(elt.name);
+    } else {
+        return hierarchyNode === elt.name;
+    }
+}
+
+function PointCloudDebug(view, datUi) {
```
autra: Actually I'm wondering why we are exposing a constructor, as we don't expose any API OOP style. Maybe we can expose a `generatePointCloudDebugTools` instead. (it would prevent the no-new above too).

utils/debug/PointCloudDebug.js
```diff
@@ -0,0 +1,192 @@
...
+    sticky.add(layer, 'dbgStickyNode').name('Sticky node name').onChange(update);
+    sticky.add(layer, 'dbgDisplayChildren').name('Display children of sticky node').onChange(update);
+    sticky.add(layer, 'dbgDisplayParents').name('Display parents of sticky node').onChange(update);
+
+    // bbox
+    layer.debugUI.add(layer, 'dbgDisplaybbox').name('Display bounding boxes').onChange(update);
+
+    // point count
+    const pcount = layer.debugUI.addFolder('Point count');
+    pcount.add(layer, 'dbgEnablePointCount').name('Enable point cound viz').onChange(update);
```
autra: typo: `Enable point count...`

utils/debug/PointCloudDebug.js
```diff
@@ -0,0 +1,192 @@
...
+    pcount.add(layer, 'dbgPointCountThreshold1').name('Threshold 1').min(1).max(1000000)
+        .onChange(update);
+    pcount.add(layer, 'dbgPointCountThreshold2').name('Threshold 2').min(1).max(1000000)
+        .onChange(update);
+    pcount.add(layer, 'dbgDisplayType', ['all', 'red', 'green', 'blue']).onChange(update);
+
+    const dump = {
+        fn: () => {
+            const d = visibilityDump(layer.root);
+            // eslint-disable-next-line no-console
```
autra: I guess I should disable no-console in debug/ folder too!

utils/debug/PointCloudDebug.js
```diff
@@ -0,0 +1,192 @@
...
+                    elt.obj.material.uniforms.debugColor.value.setRGB(0, 1, 0);
+                    if (layer.dbgDisplayType != 'all' && layer.dbgDisplayType != 'green') {
+                        elt.obj.material.visible = false;
+                    }
+                } else {
+                    elt.obj.material.uniforms.debugColor.value.setRGB(0, 0, 1);
+                    if (layer.dbgDisplayType != 'all' && layer.dbgDisplayType != 'blue') {
+                        elt.obj.material.visible = false;
+                    }
+                }
```
autra: Just by curiosity, how do you use this debug tool? What problem has it helped you to solve?

autra (changes requested):
DISCLAIMER: I didn't review EDL yet. If it's not too complicated, I would advocate to have a separate PR for it.

I think it's a really good start, and even though I requested some changes, I don't have any major objections. On overall I think we could better format the examples to make it public ready, and comment a bit more the code to be more accessible to people not familiar with it.

That being said, as I participated (a bit) in the elaboration of this code, I might not be the best reviewer here. Does someone want to step in?

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: A[xis]A[ligned]B[ounding]B[ox]

It's related to PotreeConverter file structure. And this specific function compute the bounding-box of a node's child based on its index.

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: Actually, I plan to move all the file parsing to other files (BinLoader or Driver or whatever) when I'll refactor to support `pnts` format.

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: The point here was to be able to experiment both in itowns and lopocs while having the minimum code impact.

I could change that, but again, all the file format specific code will move in separate files soon.

autra: Ok makes sense. In the future we might either have different protocol for pointclouds (like we have different protocol value for wmts, wms despite all being layer colors), or a "format" paramater?

src/Process/PointCloudProcessing.js
peppsac: Draws a cube as lines. So 1 cube = 12 lines = 24 vertices.

src/Process/PointCloudProcessing.js
peppsac: One thing lacking here is command cancelation.

And yes priority function needs tweaking :)

src/Process/PointCloudProcessing.js
peppsac: Not needed, as MainLoop does:
```js
function updateElements(context, geometryLayer, elements) {
    if (!elements) {
        return;
    }
```

src/Process/PointCloudProcessing.js
peppsac: I don't think so. Anyway I'll simplify these 2 loops.

src/Renderer/PointsMaterial.js
peppsac: Done.

src/Process/PointCloudProcessing.js
autra: Yes I know it's not needed, just found it more consistent and safer. Not very important though.

src/Process/PointCloudProcessing.js
peppsac: Done

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: Done

src/Renderer/Shader/EDLPassOneFS.glsl
peppsac: Debug code left-over. I removed it.

src/Process/PointCloudProcessing.js
autra: By using splice with a forward for loop, you're going to jump above some elements right?

src/Renderer/Shader/PointsFS.glsl
peppsac: Done.

src/Renderer/Shader/PointsVS.glsl
peppsac: Done

src/Renderer/c3DEngine.js
peppsac: Done

utils/debug/PointCloudDebug.js
peppsac: Let's say you have a scene using 2.5M points. This tool helps you understand the distribution: is there a big node with 2M point, or are there lots of 10k points nodes, etc

src/Core/View.js
peppsac: Ok, I'll rewrite it using a few more lines :)

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: Done.

peppsac: Done

peppsac:
Pushed an update that should cover most/all of @autra comments.

examples/pointcloud.js
```diff
@@ -0,0 +1,164 @@
...
+
+    window.menuGlobe = new GuiTools('menuDiv');
+
+    window.view = new itowns.View('EPSG:3946', viewerDiv, { renderer: { logarithmicDepthBuffer: false } });
+
+    const info = document.createElement('span');
+    document.getElementById('info').appendChild(info);
+    document.getElementById('info').appendChild(document.createElement('br'));
+
+    window.pointcloud = new itowns.GeometryLayer('pointcloud');
```
Jeremy-Gaillard: Missing `view.scene` parameter

examples/pointcloud.html
```diff
@@ -0,0 +1,162 @@
...
+        <div id="viewerDiv"></div>
+        <div id="info"></div>
+
+        <div class="centered">
+            <div id="lopocs"> </div>
+            <div id="itowns"> </div>
+            <h2>Point cloud absolute URL: </h2>
+            <h4>(file.js from PotreeConverter or Lopocs url)</h4>
+            <input id="url" type="url" value="https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/eglise_saint_blaise_arles/eglise_saint_blaise_arles.js">
+            <input id="submit" type="submit" value="Se Connecter">
```
Jeremy-Gaillard: French button text

examples/pointcloud.js
```diff
@@ -0,0 +1,164 @@
...
+    const viewerDiv = document.getElementById('viewerDiv');
+
+    itowns.THREE.Object3D.DefaultUp.set(0, 0, 1);
+
+    itowns.proj4.defs('EPSG:3946',
+        '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+    window.menuGlobe = new GuiTools('menuDiv');
+
+    window.view = new itowns.View('EPSG:3946', viewerDiv, { renderer: { logarithmicDepthBuffer: false } });
```
Jeremy-Gaillard: Why is the logarithmic depth buffer deactivated if we are in the EPSG:3946 reference system (where we should have very large coordinate numbers)?

examples/pointcloud.js
```diff
@@ -0,0 +1,164 @@
...
+
+        selectionGroup.visible = true;
+        view.notifyChange(true);
+    });
+
+    const flyControls = new itowns.FlyControls(view, { focusOnClick: true });
+
+    menuGlobe.gui.add(pointcloud, 'octreeDepthLimit', -1, 20).name('Depth limit')
+        .onChange(() => view.notifyChange(true));
+    menuGlobe.gui.add(pointcloud, 'pointBudget', 500000, 15000000).name('Max point count')
```
Jeremy-Gaillard: 500000 seems too high of a lower bound...

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,274 @@
...
+        }
+        layer.fetchOptions = layer.fetchOptions || {};
+
+        Fetcher.json(`${layer.url}/${layer.file}`, layer.fetchOptions).then((cloud) => {
+            layer.metadata = cloud;
+
+            let bbox;
+
+            if (layer.metadata.scale) {
+                layer.metadata.customBinFormat = layer.metadata.pointAttributes == 'CIN';
```
Jeremy-Gaillard: nit: ===

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,274 @@
...
+            layer.file = 'cloud.js';
+        }
+        layer.fetchOptions = layer.fetchOptions || {};
+
+        Fetcher.json(`${layer.url}/${layer.file}`, layer.fetchOptions).then((cloud) => {
+            layer.metadata = cloud;
+
+            let bbox;
+
+            if (layer.metadata.scale) {
```
Jeremy-Gaillard: Why is scale the attribute tested in this condition?

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,274 @@
...
+            max.max(v);
+        }
+        tightbbox = new THREE.Box3(min, max);
+    }
+
+    var geometry = new THREE.BufferGeometry();
+    geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+    geometry.addAttribute('color', new THREE.BufferAttribute(colors, 4, true));
+
+    // generate unique id for picking
```
Jeremy-Gaillard: For factorization purposes, it may be worth it to use the same method for picking (of multiple geometries packed in the same geometry object) that is used in 3d-tiles: a `batch_id` that is unique to each object in the tile and another id unique to the tile.

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,300 @@
...
+
+    const area = THREE.ShapeUtils.area(contour);
+
+    return Math.abs(area);
+}
+
+function initBoundingBox(elt, layer) {
+    const size = elt.tightbbox.getSize();
+    elt.obj.boxHelper = new THREE.LineSegments(
+    cube(size),
```
Jeremy-Gaillard: nit: indent these tow lines?

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        if (layer.counters.displayedCount > layer.pointBudget) {
+            const reduction = layer.pointBudget / layer.counters.displayedCount;
+            for (const pts of layer.group.children) {
+                if (pts.material.visible) {
+                    pts.geometry.setDrawRange(0, pts.geometry.drawRange.count * reduction);
+                }
+            }
+            layer.counters.displayedCount *= reduction;
```
Jeremy-Gaillard, line 278: Pretty sure this will result in rounding errors. It is safer to compute the number of points hidden for each element and subtract them from `displayedCount`

Jeremy-Gaillard: Also, hiding the same ratio of points for each node is suboptimal since the points in the nodes higher in the hierarchy are supposed to hold more information. I don't know if this has a big visual impact.

Jeremy-Gaillard (commented):
Partial review, I'll finish it early next week.

General remarks:
- Add point cloud example to the example index page
- postUpdate should be in another commit/PR

examples/pointcloud.html
peppsac: Fixed.

examples/pointcloud.js
peppsac: It caused rendering issues. I'll add a TODO to make sure we revisit this decision.

examples/pointcloud.js
peppsac: For all the pointclouds I tested, 500k was a very low value.
I replaced `500000` by `1` anyway.

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: I'll add a comment to explain why (to differentiate between loposc & potreeconverter).

src/Process/PointCloudProcessing.js
peppsac: Regarding rounding errors: it's not relevant. We're talking about 100k or 1M of points. So a rounding errors, if any, are negligible.

And yes maybe we wan't something smarter and hide points differently. The main advantadge of the current algo is: it's working and it's dead simple.

src/Core/MainLoop.js
```diff
@@ -69,6 +69,9 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
         context.geometryLayer = geometryLayer;
         const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer, updateSources);
         updateElements(context, geometryLayer, elementsToUpdate);
+        if (geometryLayer.postUpdate) {
```
peppsac: Moved this part to https://github.com/iTowns/itowns/pull/407

src/Renderer/PointCloudEDLRenderer.js
```diff
@@ -0,0 +1,170 @@
...
+
+    this.edlMaterial = new THREE.ShaderMaterial({
+        uniforms: {
+            depthTexture: { value: null },
+            depthTextureHalfRes: { value: null },
+            depthTextureQuarterRes: { value: null },
+            subsample: { value: false },
+            resolution: { value: new THREE.Vector2(256, 256) },
+            cameraNear: { value: 0.01 },
+            cameraFar: { value: 100 },
```
Jeremy-Gaillard: Far and near uniforms do not match the camera's values.

Edit: no impact, but maybe a bit confusing?

src/Renderer/Shader/EDLPassOneFS.glsl
```diff
@@ -0,0 +1,77 @@
...
+        zp1 = readDepth(depthTextureHalfRes, vUv);
+        zp2 = readDepth(depthTextureQuarterRes, vUv);
+    }
+    float s = 0.0;
+
+    const int max_k = 16;
+    const int max_n = 4;
+
+    float step = 2.0 * 3.1415926 / float(directions);
+    for (int i=0; i<max_k; i++) {
```
Jeremy-Gaillard: You could use a constant by editing the shader in the js script. I'm not sure how well the GPU handles these kinds of hacks, performance wise...

src/Renderer/PointCloudEDLRenderer.js
```diff
@@ -0,0 +1,170 @@
...
+    this.quad.frustumCulled = false;
+    this.edlScene.add(this.quad);
+    this.edlCam = new THREE.OrthographicCamera(-1, 1, 1, -1, 0, 10);
+
+    this.edlMaterial = new THREE.ShaderMaterial({
+        uniforms: {
+            depthTexture: { value: null },
+            depthTextureHalfRes: { value: null },
+            depthTextureQuarterRes: { value: null },
+            subsample: { value: false },
```
Jeremy-Gaillard: Subsample seems a bit misleading to me... I first thought it was used to save resources.

Maybe rename to multiscale_sampling or something like that?

src/Renderer/Shader/BasicVS.glsl
```diff
@@ -0,0 +1,6 @@
+varying vec2 vUv;
```
Jeremy-Gaillard: The attributes and uniforms are not declared. Does Three automatically add them to the shader? If that's the case, we should be consistent in our shader implementation and only declare what is necessary (or, on the contrary, declare everything).

examples/pointcloud.html
```diff
@@ -0,0 +1,162 @@
...
+
+        <div class="centered">
+            <div id="lopocs"> </div>
+            <div id="itowns"> </div>
+            <h2>Point cloud absolute URL: </h2>
+            <h4>(file.js from PotreeConverter or Lopocs url)</h4>
+            <input id="url" type="url" value="https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/eglise_saint_blaise_arles/eglise_saint_blaise_arles.js">
+            <input id="submit" type="submit" value="Connect">
+            <div id="result">
+                <h3>Nuages disponibles:</h3>
```
autra: Should be in English

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -0,0 +1,45 @@
...
+varying vec4 vColor;
+
+void main() {
+    if (pickingMode) {
+        vColor = unique_id;
+    } else {
+        vColor = vec4(color, 1.0);
+    }
+    vec4 mvPosition = modelViewMatrix * vec4( position, 1.0 );
+    mat4 projModelViewMatrix = useRTC ? mVPMatRTC: projectionMatrix * modelViewMatrix;
```
Jeremy-Gaillard: RTC is no longer relevant AFAIK

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -0,0 +1,45 @@
...
+
+    if (size > 0.01) {
+        gl_PointSize = size;
+    } else {
+        float pointSize = 1.0;
+        float slope = tan(1.0 / 2.0);
+        float projFactor =  -0.5 * resolution.y / (slope * mvPosition.z);
+
+        float z = min(0.5 * -gl_Position.z / gl_Position.w, 1.0);
+        gl_PointSize = max(3.0, min(10.0, 0.05 * projFactor));// min(10.0, 3.0 / (z * z)); //min(8.0, 0.05 * (resolution.y * 0.5 / - mvPosition.z));
```
Jeremy-Gaillard: This doesn't seem completely sorted out yet.

peppsac:
Removed EDL from this PR. I'll submit a different PR later.

utils/debug/PointCloudDebug.js
```diff
@@ -0,0 +1,192 @@
...
+    layer.dbgDisplaybbox = false;
+
+    layer.dbgEnablePointCount = false;
+    layer.dbgPointCountThreshold1 = 10000;
+    layer.dbgPointCountThreshold2 = 100000;
+    layer.dbgDisplayType = 'all';
+
+    // UI
+    const update = () => this.view.notifyChange(true);
+    layer.debugUI = this.gui.addFolder(`${layer.index}: ${layer.url}`);
```
Jeremy-Gaillard: This produces a name far too long for the GUI.

layer.index is undefined

Jeremy-Gaillard (commented):
Part 2. This review is (unfortunately) mostly on EDL.

peppsac:
@Jeremy-Gaillard thanks for your comments on EDL. I'll update the code with them, even if it has been removed from this PR (sorry :-/)

examples/pointcloud.html
```diff
@@ -0,0 +1,156 @@
...
+                        showPointcloud(url, undefined, table, false);
+                    });
+
+                    td2.appendChild(i);
+                    tr.appendChild(td);
+                    tr.appendChild(td2);
+                    return tr;
+                }
+
+                document.getElementById('submit').addEventListener('click', () => {
```
autra: No arrow function in examples

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,283 @@
...
+        min = new THREE.Vector3().copy(cmin).add(xHalfLength).add(yHalfLength);
+        max = new THREE.Vector3().copy(cmax).add(xHalfLength).add(yHalfLength);
+    }
+
+    return new THREE.Box3(min, max);
+}
+
+
+function parseOctree(layer, hierarchyStepSize, root) {
+    return fetch(`${root.baseurl}/r${root.name}.hrc`, layer.fetchOptions).then(response => response.arrayBuffer()).then((blob) => {
```
autra: I think you meant `Fetcher.arrayBuffer` here.

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,283 @@
...
+    var geometry = new THREE.BufferGeometry();
+    geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+    geometry.addAttribute('color', new THREE.BufferAttribute(colors, 4, true));
+
+    // generate unique id for picking
+    const ids = new Uint8Array(4 * numPoints);
+    const base_id = nextuuid++;
+    if (numPoints > 0xffff) {
+        // TODO: fixme
+        // throw new Error('unhandled value');
```
autra: It could be a good idea to throw something, or at least warn that picking won't work correctly, if you don't want to fix this here.

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,283 @@
...
+    points.tightbbox = tightbbox;
+    points.materials = {
+        color: material,
+    };
+
+    return points;
+}
+
+function loadBin(layer, url, readPositionAndColor, hasBbox) {
+    return fetch(url, layer.fetchOptions).then(foo => foo.arrayBuffer()).then(ab =>
```
autra: `Fetcher.arrayBuffer`

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -0,0 +1,283 @@
...
+        const node = command.requester;
+
+        // Query HRC if we don't have children metadata yet.
+        if (node.childrenBitField && node.children.length === 0) {
+            parseOctree(layer, layer.metadata.hierarchyStepSize, node).then(() => command.view.notifyChange(false));
+        }
+
+        const extension = layer.metadata.customBinFormat ? 'cin' : 'bin';
+
+        const url = `${node.baseurl}/r${node.name}.${extension}?isleaf=${command.isLeaf ? 1 : 0}`;
```
autra: Can you elaborate on this `isleaf` parameter? I guess it's for lopocs, but doesn't it know if a node is leaf already?

src/Process/PointCloudProcessing.js
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        // Start updating from hierarchy root
+        return [layer.root];
+    },
+
+    update(context, layer, elt) {
+        const { shouldBeLoaded } = shouldDisplayNode(context, layer, elt);
+
+        const boundingBoxEnabled = context.view.camera.camera3D.layers.mask & layer.bboxes.layers.mask;
```
autra: Should be inside a `if (__DEBUG__)`? 

I'm thinking out loud, but maybe in the future, we could officially support the display of bounding boxes, WDYT?

peppsac:
PR rebased and commits squashed.

examples/pointcloud.html
```diff
@@ -0,0 +1,156 @@
...
+                function createResultRow(serverUrl, table) {
+                    var tr = document.createElement('tr');
+                    var td = document.createElement('td');
+                    td.textContent = table;
+                    var td2 = document.createElement('td');
+                    var i = document.createElement('input');
+                    i.value = 'Display';
+                    i.type = 'submit';
+
+                    i.addEventListener('click', () => {
```
autra: No arrow function in examples (html are not linted).

examples/pointcloud.html
```diff
@@ -0,0 +1,156 @@
...
+                    var url = document.getElementById('url').value;
+
+                    // assume that "url ends with .js" == potreeconverter
+                    if (url.endsWith('.js')) {
+                        var lastSlashIndex = url.lastIndexOf('/');
+                        var filename = url.substr(lastSlashIndex + 1);
+
+                        showPointcloud(url.substr(0, lastSlashIndex), filename, undefined, true);
+                    } else {
+                        itowns.Fetcher.json(url + '/infos/sources').then((j) => {
```
autra: no arrow function

autra: A good way to test your html is to use midori. It currently does not support arrow functions :-)

Ideally, I'd fix IE11 and we would test with it.

src/Core/View.js
```diff
@@ -44,7 +44,17 @@ function View(crs, viewerDiv, options = {}) {
 
     this.referenceCrs = crs;
 
-    this.mainLoop = options.mainLoop || new MainLoop(new Scheduler(), new c3DEngine(options.renderer || viewerDiv));
+    let engine;
+    // options.renderer can be 2 separate things:
+    //   - an actual renderer (in this case we don't use viewerDiv)
+    //   - options for the renderer to be created
```
autra, line 52: This comment should be in the doc above. Suggestion:

```js
/**
 * ...
 * @param {?(WebGLRenderer|object)} options.renderer - {@link WebGLRenderer} instance to use, otherwise 
 * a default one will be constructed. In this case, if options.renderer is an object, it will be used to 
 * configure the renderer (see {@link c3DEngine}.  If not present, a new <canvas> will be created and 
 * added to viewerDiv (mutually exclusive with mainLoop)
```

src/Process/PointCloudProcessing.js
autra: Or maybe get rid of this variable, it is used only once.

src/Renderer/Shader/PointsVS.glsl
autra: Yep, let's remove it.

autra (changes requested):
With these last comments addressed, I'll approve this.

examples/pointcloud.html
peppsac: Except I don't have IE11 on my machine :)

examples/pointcloud.html
```diff
@@ -112,40 +112,46 @@
...
-                            // clean table
-                            while (dd.firstChild) {
-                                dd.removeChild(dd.firstChild);
-                            }
-                            var i;
-                            for (i = 0; i < j.length; i++) {
-                                dd.appendChild(createResultRow(url, j[i].table));
-                            }
-                        });
+                        itowns.Fetcher.json(url + '/infos/sources').then(function (metadata) { buildLopocsResultTable(metadata) });;
```
autra: nits `itowns.Fetcher.json(url + '/infos/sources').then(buildLopocsResultTable);`

autra: (and one `;`)

autra (approved):
LGTM.

Just a little nits to fix and a squash, and we're good :-)

examples/pointcloud.html
peppsac: Fixed (and I added an entry to `examples/index.html`)

src/Process/PointCloudProcessing.js
gchoqueux: in OBBHelper, there's method to draw cube with lines

peppsac: I know but `OBBHelper` is designed around `OBB` so it won't work for my use-case without some refactoring.

gchoqueux: You can be inspired by OBBHelper,  to draw cube with only 8 vertex

PR #402, opened 2017-07-21
fix: don't block subdivision on frozen layers

peppsac:


autra (approved):

autra:
Thanks!

Mkonini:
I think this causes a regression, when you zoom to max, globe disappears!

autra:
@Mkonini I can reproduce before @peppsac's commit, so I confirm the bug, but its unrelated to this PR.

PR #404, opened 2017-07-21
fix: 3d-tiles replace refinement behaviour properly hides tile content

Jeremy-Gaillard:


src/Process/3dTilesProcessing.js
```diff
@@ -144,8 +144,8 @@ export function process3dTilesNode(cullingTest, subdivisionTest) {
         if (node.parent.pendingSubdivision && !node.parent.additiveRefinement) {
             node.visible = false;
             // TODO: node.setDisplayed(false)?
-            if (node.material) {
-                node.material.visible = false;
+            if (node.content) {
```
peppsac: Maybe copy-paste the commit comment here. Or maybe add a `setTileVisible()` function, including this comment.
WDYT?

Jeremy-Gaillard: I created a setDisplayed function.

I also simplified the code a bit by removing the calls to setDisplayed when the tile was not visible.

src/Process/3dTilesProcessing.js
```diff
@@ -138,15 +138,17 @@ export function init3dTilesLayer(context, layer) {
     });
 }
 
+function setDisplayed(node, display) {
+    if (node.content) {
```
peppsac: You should copy-paste a comment here explaining why your not simply doing `node.material.visible = false`

Jeremy-Gaillard: Done

peppsac (approved):

PR #407, opened 2017-07-24
feature: add a postUpdate method to GeometryLayer

peppsac:
Allows the user to perform an action when a GeometryLayer update is finished.
(this is different from preRender since an update may not request a render)

### DIFF
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -71,6 +71,8 @@ function GeometryLayer(id, object3d) {
         value: id,
         writable: false,
     });
+
+    this.postUpdate = () => {};
 }
 
 GeometryLayer.prototype = Object.create(EventDispatcher.prototype);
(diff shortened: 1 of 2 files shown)


src/Core/Layer/Layer.js
```diff
@@ -71,6 +71,8 @@ function GeometryLayer(id, object3d) {
         value: id,
         writable: false,
     });
+
+    this.postUpdate = () => {};
```
autra, line 75: we don't have a `this.preUpdate = () => {};`. Let's either add it or have none of the 2 and guard the method call in `MainLoop` (I midly prefer the second possibility, more resilient to user-provided layers).

autra (changes requested):

src/Core/Layer/Layer.js
peppsac: The `preUpdate` method is mandatory while the `postUpdate` isn't.

IMHO implementing default optionnal methods eases readability of the `MainLoop` but if you prefer I can revert to `if (layer.postMethod) layer.postMethod` style.

autra: No it's ok I get it. I forgot we transform object passed to `addLayer` to `Layer` instance anyway.

autra (approved):

PR #408, opened 2017-07-24
style: fix new linters errors

peppsac:
See https://travis-ci.org/iTowns/itowns/jobs/256790269


autra:
For reference, the build was broken only in npm3 because it does not honor package-lock.json: the travis build picked a new version of eslint-config-airbnb-base, which contains https://github.com/airbnb/javascript/commit/00c9e52aec17b9a44fbf03494a931346e8f090de. That commits makes our conf to report this error. 

That's also why you won't be able to reproduce it locally if:
- you are on npm 5
- you are on npm 3 but still have the old version on your `node_modules/` folder.

@peppsac I've pushed a commit updating package.json and package-lock.json with eslint-config-airbnb-base to 11.3.0.

peppsac:
Perfect, thanks.

autra (approved):

PR #409, opened 2017-07-24
Chore: add a template for gh issues

autra:
This PR adds a template for our issues. I separated this from the PR template after I had some remarks about it. It will be easier to reach agreement on individual parts.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,40 @@
+<!--- Provide a general summary of the issue in the Title above -->
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+* Version used:
+* Browser Name and version:
+* Operating System and version (desktop or mobile):
+* Link to your project:
+
+## Context
+<!--- For bugs, Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
+<!--- How has this bug affected you? -->
+<!--- If you have errors or a stacktrace in the console, here is a good place to put it. -->
+<!--- Please consider using a pastebin service or create a gist.-->
+<!--- Some screenshots are always greatly appreciated here! -->
+
+<!--- For improvement suggestions, please tell us what you are trying to accomplish -->
+<!--- Providing context helps us come up with a solution that is most useful in the real world -->
+
+## Steps to Reproduce (for bugs)
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug. Include code to reproduce, if relevant -->
+1.
+2.
+3.
+4.
+
+## Expected Behavior
+<!--- If you're describing a bug, tell us what you think should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+
+## Actual Behavior
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
+<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+
+## Possible Cause/Fix/Solution
+<!--- Not obligatory, but suggest a cause or reason for the bug if relevant, -->
+<!--- or ideas how to implement the addition or change -->
+<!--- In any case, please consider opening a PR and become an iTowns developer! -->
+


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Steps to Reproduce
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug include code to reproduce, if relevant -->
+1.
+2.
+3.
+4.
+
+## Context
```
tbroyer: This section appears twice

tbroyer (changes requested):
I would reorder the sections, with the steps to reproduce and environment early on, and the expected and actual behavior just after the steps to reproduce. Something like:

 * Environment
 * Steps to Reproduce
 * Expected Behavior
 * Actual Behavior
 * Possible Fix
 * Context

(possibly move "context" up in the list, maybe between "environment" and "steps to reproduce")

Also, maybe rename "possible fix" to something hinting at the cause rather than the fix, and/or debugging instead (i.e. what the reporter found during debug and might be the cause of the issue –without necessarily suggesting a fix, otherwise he could make a PR– or other similar comments). Maybe simply use "Possible Cause" rather than "Possible Fix".

Those are only suggestions. The reason I "request changes" is the duplicate "context" section.

ISSUE_TEMPLATE.md
autra: Let's rename one of those. Do you have a suggestion? (I thought of calling the first one `Description` but...)

tbroyer: Or just "merge" them? Unless you really think it's important to have 2 distinct sections.

autra:
I've used your order @tbroyer, with the 2 context sections merged, just below `Environment`

@iTowns/reviewers please r?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Expected Behavior
+<!--- Tell us what should happen -->
+
+## Actual Behavior
+<!--- Tell us what happens instead -->
+
+## Possible Cause/Fix
+<!--- Not obligatory, but suggest a cause or reason for the bug -->
+<!--- Do you have a fix? Why not open a Pull Request then? :-) -->
```
tbroyer: Might possibly be read as passive-aggressive by some people, despite the smiley.

I'd only talk about the cause. If the reporter has a suggestion for a fix, we could hope that he'll put it there too ("I think it comes from X, and changing it to Y might be enough to fix the bug")

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+## Steps to Reproduce
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug include code to reproduce, if relevant -->
+1.
+2.
+3.
+4.
+
+## Expected Behavior
+<!--- Tell us what should happen -->
```
tbroyer: "what you think should happen" :wink:

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+* Version used:
+* Browser Name and version:
+* Operating System and version (desktop or mobile):
+* Link to your project:
+
+## Context
+<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
+<!--- How has this bug affected you? What were you trying to accomplish? -->
+<!--- If you have errors or a stacktrace in the console, here is a good place
+to put it. Please consider using a pastebin service or create a gist.-->
```
tbroyer: nit: strange to use a multiline comment here; inconsistent with the rest of the template (e.g. steps to reproduce)

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
+<!--- Provide a general summary of the issue in the Title above -->
```
tbroyer, line 1: Should we add some comment about what to do if one wants to request an enhancement rather than report a bug? ("if you want to request an enhancement / a new feature, then feel free to disregard the following template", maybe followed by something like "please explain why it would be a good idea, and concentrate on the needs this would solve rather than how it would solve them and/or sketching an API")

tbroyer (commented):
Some comments; take them mostly as suggestions.

I'd rather let others review too before merging, but LGTM

ISSUE_TEMPLATE.md
autra: There was 2 choices in the template generator I used to quick-start this: one was "template focused on issue", the other one was "template focused on improvement". I chose to focus on issues, because I feel it's usually the hardest to decrypt from reporters. That being said, we should definitely say something about improvement yes.

Or maybe have each section be phrased to take both possibilities into account? I'll try to see if it can be done easily (and without too much verbosity).

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+## Context
+<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
+<!--- How has this bug affected you? What were you trying to accomplish? -->
+<!--- If you have errors or a stacktrace in the console, here is a good place
+to put it. Please consider using a pastebin service or create a gist.-->
+<!--- Some screenshots are always greatly appreciated here! -->
+
+## Steps to Reproduce
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug include code to reproduce, if relevant -->
```
peppsac: Replace `reproduce this bug include code to reproduce ...` with:
- `reproduce this bug. Include code to reproduce ...` or
- `reproduce this bug including code to reproduce ...`

ISSUE_TEMPLATE.md
autra: It was intended as an encouraging "go ahead, you can do it!", but apparently fails at it. What if I rephrase this way?

> Have a fix? Please open a PR and become an iTowns developer!

peppsac: Agreed: I'd only keep the 1st sentence.

peppsac (approved):

ISSUE_TEMPLATE.md
tbroyer: @autra Much better!

autra:
@iTowns/reviewers I rephrased this trying to make it usable for improvements. Please r?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,40 @@
...
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+
+## Actual Behavior
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
+<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+
+## Possible Cause/Fix/Solution
+<!--- Not obligatory, but suggest a cause or reason for the bug if relevant, -->
+<!--- or ideas how to implement the addition or change -->
+<!--- In these case, please consider opening a PR and become an iTowns developer!--->
```
tbroyer: "This case" or "These cases"

autra: or just "Please consider opening a PR and becom*ing* an iTowns developer!"

autra:
Pushed a new version, I only changed  the last sentence (the bit about becoming an iTowns developer). please r?, @iTowns/reviewers

autra:
@tbroyer do you have other comments? 

other @iTowns/reviewers don't hesitate to comment to. I think this'll be useful to anybody (including myself).

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,40 @@
...
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+
+## Actual Behavior
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
+<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+
+## Possible Cause/Fix/Solution
+<!--- Not obligatory, but suggest a cause or reason for the bug if relevant, -->
+<!--- or ideas how to implement the addition or change -->
+<!--- In any case, please consider opening a PR to become an iTowns developer! -->
```
tbroyer: nit: “…consider opening a PR **and** become an iTowns developer” ?

tbroyer (approved):

ISSUE_TEMPLATE.md
autra: Yep, done :-)

PR #410, opened 2017-07-24
Chore: add a PR template for gh

autra:


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+## Motivation and Context
+<!--- Why is this change required? What problem does it solve? -->
+<!--- If it fixes an open issue, please link to the issue here. -->
+
+## How Has This Been Tested?
+<!--- Please describe in detail how you tested your changes. -->
+<!--- Include details of your testing environment, and the tests you ran to -->
+<!--- see how your change affects other areas of the code, etc. -->
+
+## Screenshots (if appropriate):
```
tbroyer: nit: remove colon (other section titles don't end with a colon)

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Types of changes
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
```
tbroyer: I don't think using a "todo list" / "task list" markup is a good idea for such questions.

Maybe instruct to put a `:heavy_check_mark:` ( :heavy_check_mark: ) or `:white_check_mark:` ( :white_check_mark: ) in front of the lines that apply (and either leave the others without mark, or put a `:x:` :x: in front of them)

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
+<!--- Provide a general summary of your changes in the Title above -->
+
+## Description
+<!--- Describe your changes in detail -->
+
+## Motivation and Context
```
peppsac: I'm not sure it's a good idea to split `Description` and `Motivation / Context`.
What's the expected gain of this split over a unique `Description` section?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+<!--- Why is this change required? What problem does it solve? -->
+<!--- If it fixes an open issue, please link to the issue here. -->
+
+## How Has This Been Tested?
```
peppsac: Can you give us an example of what you would write for a PR here? (eg for https://github.com/iTowns/itowns/pull/392)

PULL_REQUEST_TEMPLATE.md
peppsac: This kind of list is redundant IMHO.
You already asked the developer to provide: a summary, a detailed description + context. If a PR introduces a new behavior the descriptions should be enough to understand what's going on: no need to require box-ticking.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Screenshots (if appropriate):
+
+## Types of changes
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+
+## Checklist:
```
peppsac: Meh.
`My code follows the code style of this project.` => we have a linter for this.
`My change requires a change to the documentation.` and `I have updated the documentation accordingly` => we only have an API documentation, and this API doc has a linter as well.
`I have read the **CONTRIBUTING** document.` => what are you expecting from this line?
`I have added tests to cover my changes.` => we don't have tests, so this one can be removed for now.
` All new and existing tests passed.` => Travis already does the check.

So I'd remove this check list completely. If you want to list some useful hints maybe write them in the `CONTRIBUTING` file?

peppsac (commented):
Thanks for starting the PR.

I've left a few comments, but the summary is: I'd only keep `Description` + `Screenshots`.

Note that while having a PR template is all nice and well, it also add some more friction to the PR process.
Right now, we're more in need of commiters than in need of formatting PRs properly.

autra:
> Note that while having a PR template is all nice and well, it also add some more friction to the PR process.
> Right now, we're more in need of commiters than in need of formatting PRs properly.

Right, I agree with that. Maybe this template is too verbose for PRs. I'll rewrite and come up with a shorter proposal. Maybe a set of recommandations in a comment is enough, without having a real template WDYT?

PULL_REQUEST_TEMPLATE.md
autra: Tested on Firefox 55 (this is enough to raise reviewer's awareness on IE compatibility for instance). Maybe we can ask for this information only.

PULL_REQUEST_TEMPLATE.md
autra: > this API doc has a linter as well.

The linter has nothing to do with this (check only the syntax, not missing+outdated doc). But anyway, it's more the responsibility of the reviewer to check that. Actually, we should have a checklist for reviewers instead (browser compat ? Doc ? etc...)!

autra:
@iTowns/reviewers updated. It's much simpler now.

Note that the very relevance of a PR template can be discussed here too. I don't mind to close this if you guys find it useless.

tbroyer:
> Note that the very relevance of a PR template can be discussed here too. I don't mind to close this if you guys find it useless.

Yeah, with this last version, I'm not sure it's really worth it… Could be even further simplified by just including comments to ask for screenshots and/or link to issue if appropriate.

I personally like how GitHub defaults to using the commit message when the PR contains only one commit, and I suppose this wouldn't be the case anymore with a PR template.

autra:
Good, it gives me the occasion to state what I have in mind. 

I find that generally we don't really give enough context to our PR. Also, we have the tendency to have PR that contains several features, or mix features with bug fixes. This template might help us realise this: if we cannot really describe our PR in a very specific way, it means it should be splitted.

True: we reviewer have also the tendency of not really looking at commit messages too (and check if they are good enough), but as most of our PR are not single commits, I expect the gain of a template to outweight the loss of not having single commits PR automatically grab the commit message.

And writing this, I realize that I would probably have written it in the ` Motivation and Context` section, if I had this template :-)

autra:
So actually, I feel this template has more value now that it has before I decreased its length.

autra:
@iTowns/reviewers any comments?

tbroyer (approved):

Issue #411, opened 2017-07-24
White screen when zooming


vcoindet:
When the globe is ready and you zoom with the scroll wheel, everything suddenly disappears when your close to level 13 (like we were under the globe) and all we can see is a white screen.
The same thing happens when using the `globeView.controls.setZoom()` for levels higher than 9. (same bahavior for `setScale()`, `setRange()` and `setOrbitalPosition()`, which is logical).

It seems that this bug appeared in the following commit https://github.com/iTowns/itowns/commit/6c7d9b6b54c1a06116f5b0962511224385591bc3

peppsac:
Nope, it was broken before (I just tried  ea65fc0).

To test you can either do this after the initial loading of `index.html`:
- `globeView.controls.getRange()` should return `25.000.000` - in buggy versions it returns more than 25M
- or `globeView.controls.setRange(1000)` that will display a white screen. But if you then do `globeView.controls.setRange(10000)`, wait for a bit until tiles are loaded and then do `globeView.controls.setRange(1000)` it'll work

The problem is probably in `GlobeView.prototype.getPickingPositionFromDepth` or `GlobeControl`'s `updateCameraTargetOnGlobe` (or both).

vcoindet:
Hum... My bad sorry ! ;)

You're effectively right, I guess it is getting worse with the commit I mentionned...
In commit https://github.com/iTowns/itowns/commit/6c7d9b6b54c1a06116f5b0962511224385591bc3 : starts to bug around zoom levels 10 - 13
In commit https://github.com/iTowns/itowns/commit/ea65fc03d0a75896f1508793e3b0707d30efe5d3 : starts to bug around zoom levels 15 - 18

peppsac:
Indeed 6c7d9b6 introduces a bug in `unpack1K` function.

With the fix from #412, we're back at the same error level than before - that *might* be caused by precision issues...

vcoindet:
> that might be caused by precision issues...

After several tests I arrived to the same conclusion, but we have this kind of problems since a long time ago now... Every time the camera is far from the globe, the altitude accuracy is not that good

peppsac:
@gchoqueux why did you close this issue? (you stated in #472 that it **didn't** resolve this issue)

gchoqueux:
it's automatic closing with in the  #472 description (with the `resolve` key)

PR #412, opened 2017-07-25
fix: missing 4th component value of unpack1K factors

peppsac:
The original formula in three.js shader is:
  const vec4 UnpackFactors = UnpackDownscale / vec4( PackFactors, 1. );
so the 4th component should be 'UnpackFactors' which is:
  255 / 256

Partial fix for #411

### DIFF
--- a/src/Renderer/LayeredMaterial.js
+++ b/src/Renderer/LayeredMaterial.js
@@ -32,7 +32,7 @@ export function unpack1K(color, factor) {
         UnpackDownscale / (256.0 * 256.0 * 256.0),
         UnpackDownscale / (256.0 * 256.0),
         UnpackDownscale / 256.0,
-        1.0);
+        UnpackDownscale);
     return bitSh.dot(color) * factor;
 }
 


vcoindet:
Thanks,
We are effectively back to the same error level than before, not perfect, but better ! ;)

peppsac:
More details:

https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/packing.glsl#L17 is the function used to store the depth value in a rgba field.

https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/packing.glsl#L23 is the inverse function.
But we need it in javascript, so I translated it to javascript here: https://github.com/iTowns/itowns/commit/6c7d9b6b54c1a06116f5b0962511224385591bc3#diff-f5c5a4b625fe789476515783a6e21153R30 (see PR https://github.com/iTowns/itowns2/pull/345) but I messed up the 4th component :)

autra (approved):
Thanks!

PR #413, opened 2017-07-25
Feature: add a preprocessed promise to layer

autra:
Preprocessing is asynchrone for some case (for example in #400, and it will also be the case when we'll support GetCapabilities). Therefore, here is a proposition for dealing with this asynchronism.

I chose to implement it with a promise because:

1. it fires once and only once
2. If part of the code wants to register *after* the preprocessed event, it can still do so. In contrast, with an event, if it arrives too late, it will never get the event.

For these reasons, I think promises are the way to go here.

@iTowns/reviewers please r?


src/Core/View.js
```diff
@@ -180,6 +185,11 @@ function _preprocessLayer(view, layer, provider) {
  * Add layer in viewer.
  * The layer id must be unique.
  *
+ * This function calls `preprocessDataLayer` of the relevant provider with this
+ * layer. A this operation can be asynchrone it attaches a
```
Jeremy-Gaillard: Typo: "A this operation"

tbroyer: Also, asynchrone → asynchronous

autra: I should probably remove this bit anyway:

```js
/**
 * ...
 * This function calls `preprocessDataLayer` of the relevant provider with this
 * layer and set `layer.preprocessedPromise` to a promise that resolves when the 
 * preprocessing operation is done.

```

WDYT?

peppsac: Sounds good.

peppsac (approved):
LGTM

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -45,8 +45,17 @@ $3dTiles_Provider.prototype.removeLayer = function removeLayer() {
 
 };
 
-$3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer() {
-
+$3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer) {
+    if (layer.preprocessedPromise) {
```
peppsac: Couldn't this be done once and for all in Scheduler?

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -45,8 +45,17 @@ $3dTiles_Provider.prototype.removeLayer = function removeLayer() {
...
+        return layer.preprocessedPromise;
+    }
+    return Fetcher.json(layer.url, layer.networkOptions).then((tileset) => {
+        const urlPrefix = layer.url.slice(0, layer.url.lastIndexOf('/') + 1);
+        layer.tileset = tileset;
+        layer.tileIndex = new $3dTilesIndex(tileset, urlPrefix);
```
peppsac: Maybe you could unexport `$3dTilesIndex` now that it's used only here?

peppsac (approved):
LGTM.
Maybe @Jeremy-Gaillard will want to double-check.

src/Core/Scheduler/Providers/3dTiles_Provider.js
autra: You mean in View?

peppsac: Yes :)

autra:
@peppsac done. Waiting for @Jeremy-Gaillard input then.

src/Process/3dTilesProcessing.js
```diff
@@ -120,21 +117,13 @@ function computeNodeSSE(camera, node) {
...
-        const urlPrefix = layer.url.slice(0, layer.url.lastIndexOf('/') + 1);
-        layer.tileIndex = new $3dTilesIndex(tileset, urlPrefix);
-        layer.asset = tileset.asset;
-        requestNewTile(context.view, context.scheduler, layer, tileset.root, undefined).then(
-            (tile) => {
-                layer.object3d.add(tile);
-                tile.updateMatrixWorld();
-                layer.root = tile;
-            });
+    layer.preprocessedPromise.then((layer) => {
```
Jeremy-Gaillard: There's nothing preventing this code from being called multiple times (by this https://github.com/iTowns/itowns/blob/208fdad43c3553b283694db97b9f24e028c17a26/examples/3dtiles.html#L61). What are the implications of a promise being redefined multiple times?

More broadly, shouldn't the preprocessedPromise be completed before the layer starts updating?

peppsac: IMHO `init3dTilesLayer` should be removed completely and its code should move completely in `preprocessLayer`.

And maybe `MainLoop` should be modified to only update layers that have a resolved `preprocessedPromise`.

autra: > What are the implications of a promise being redefined multiple times?

It's not redefined, you just add another callback to its completion. Here it means that the root tiles could be requested several time. I should probably guard the call to requestNewTile (or check if a check is done there).

> More broadly, shouldn't the preprocessedPromise be completed before the layer starts updating?

Imho, not necessarily. We just have to make sure preupdate gives no element back in this case.

> IMHO init3dTilesLayer should be removed completely and its code should move completely in preprocessLayer.

not sure it is possible, but I'll give it a try.

> And maybe MainLoop should be modified to only update layers that have a resolved preprocessedPromise.

if preupdate returns an empty list, it's not necessary. I like the fact preupdate is guaranteed to be called at each loop. WDYT?

Jeremy-Gaillard: > Imho, not necessarily. We just have to make sure preupdate gives no element back in this case.

> if preupdate returns an empty list, it's not necessary. I like the fact preupdate is guaranteed to be called at each loop. WDYT?

It seems really strange that you would do some kind of processing on a layer that has not been completely defined yet.

peppsac: Actually maybe we should keep `init3dTilesLayer`, but replace its usage in `preUpdate` with something like this:
(done in `3dtiles.html` ?)
```js
layer.preprocessedPromise.then((layer) => {
   itowns.$3dtiles.init3dTilesLayer(layer);
}
```

autra:
@peppsac @Jeremy-Gaillard here is another proposition. I kept the existing api for now. I still don't see why user should explicitly call `init`, as it's required anyway.

peppsac:
@autra can you adapt GlobeView to your proposal please?

Jeremy-Gaillard:
Looks good to me. I think it is fine without an explicit init function. Both preprocessDataLayer and init are part of the initialisation process, so it doesn't bother me that the user only calls one.

autra:
@peppsac done

src/Core/View.js
```diff
@@ -107,8 +107,21 @@ function _preprocessLayer(view, layer, provider) {
             layer.tileTextureCount = provider.tileTextureCount.bind(provider);
         }
 
-        if (provider.preprocessDataLayer) {
-            provider.preprocessDataLayer(layer);
+        if (!layer.preprocessedPromise) {
```
peppsac: What about `layer.onReady`?

src/Core/View.js
```diff
@@ -107,8 +107,21 @@ function _preprocessLayer(view, layer, provider) {
...
+                if (preprocessingResult && preprocessingResult.then) {
+                    layer.preprocessedPromise = preprocessingResult;
+                } else {
+                    layer.preprocessedPromise = Promise.resolve();
+                }
+            } else {
+                layer.preprocessedPromise = Promise.resolve();
+            }
+            layer.preprocessedPromise.then(() => {
+                layer.ready = true;
```
peppsac: Too bad we can't query the state of a Promise: this forces us to add a bool just to keep track of resolved promises :(

peppsac (commented):
Looks mostly good now.

src/Core/View.js
autra: `onFoo` is traditionnally more for event. Maybe `whenReady`? It will read "layer when ready then".

src/Core/View.js
autra: You read my mind :-)

autra: The alternative is to wrap every code that needs this boolean in whenReady.then callback (if they accept to run in another microtask).

autra:
@peppsac updated with "whenReady". Let me know if you want another name.

autra:
Refactored pointcloud to use whenReady + add package-lock.json update from #400

peppsac (commented):
Would it make sense to adapt `GlobeView.prototype.addLayer` to fire the `GLOBE_VIEW_EVENTS.LAYER_ADDED` when the layer is ready?

autra:
> Would it make sense to adapt GlobeView.prototype.addLayer to fire the GLOBE_VIEW_EVENTS.LAYER_ADDED when the layer is ready?

Do you want me to tell you what would make sense regarding `LAYER_ADDED`? ;-)

More seriously, what'll make sense is to have it in `View.addLayer` instead. Can we keep that for another PR?

peppsac:
> More seriously, what'll make sense is to have it in View.addLayer instead. Can we keep that for another PR?

Ok, let's postpone this change to a `LAYER_ADDED` specific PR.

src/Core/View.js
```diff
@@ -217,11 +234,14 @@ function _preprocessLayer(view, layer, provider) {
...
  *
  * // Add Elevation Layer
- * view.addLayer({
+ * var layer = view.addLayer({
  *      type: 'elevation',
  *      id: 'iElevation',
  * });
  *
+ * // React to preprocessing
+ * layer.whenReady.then(() => { ... });
```
peppsac: Maybe you should modify `addLayer` to return `layer.whenReady` instead of `layer`. WDYT?

examples/pointcloud.js
```diff
@@ -84,7 +76,19 @@ function showPointcloud(serverUrl, fileName, lopocsTable) {
...
+                pointcloud.counters.displayedCount.toLocaleString() + ' (' +
+                Math.floor(100 * pointcloud.counters.displayedCount / pointcloud.counters.pointCount) + '%) (' +
+                view.mainLoop.gfxEngine.renderer.info.memory.geometries + ')';
+        };
+
+
         view.notifyChange(true);
     }
-    view.addLayer(pointcloud).addEventListener('preprocessed', onLayerPreprocessed);
+    view.addLayer(pointcloud).whenReady.then(onLayerPreprocessed);
```
peppsac: `onLayerReady` ?

With the previous suggestion, the whole line would become:
```js
view.addLayer(pointcloud).then(onLayerReady); 
```

src/Process/GlobeTileProcessing.js
```diff
@@ -115,19 +114,19 @@ export const globeSchemeTile0 = 0;
 export const globeSchemeTile1 = 1;
 
 export function globeSchemeTileWMTS(type) {
-    const schemeT = new SchemeTile();
+    const schemeT = [];
```
peppsac, line 117: +1

src/Core/View.js
autra: If the user didn't pass an instance of Layer to `addLayer` but a simple object with options, we create one for him (in `_preprocessLayer` I think). This instance needs to be returned (that's more or less the same pattern as THREE TextureLoader). 

We could separate layer instanciation from layer addition in the future, or keep it that way, I don't have any strong opinion on the matter for now.

peppsac (approved):

src/Core/View.js
peppsac: Right.
I'll go with the more used pattern: if in all examples we always use `addLayer.whenReady` then I'd return the promise directly (and the user can still retrieve the `layer` instance using its `id`).
Otherwise let's keep `addLayer` as is.

autra: An alternative would be to resolve whenReady with the layer. 

It does not solve the use case of needing the layer object before it is ready though. Querying `view.getLayers` just after `view.addLayer` seems cumbersome to me. Do you think this use case will be often encountered? If not, I guess we can indeed return the promise.

autra:
I have a preference for the current form for several reasons:

-  it keeps both possibility, at the cost of being slightly more verbose in the case you want only the promise.
- `view.addLayer(...).then` suggest the promise resolves when the layer is added. But actually, it's added synchronously (available in `getLayers` immediately after). It's the preprocessing + the initialisation process which are asynchronous.

peppsac:
Ok let's keep the current version.

src/Core/View.js
```diff
@@ -216,15 +234,18 @@ function _preprocessLayer(view, layer, provider) {
...
- * view.addLayer({
+ * // Add Elevation Layer and do something once it's ready
+ * var layer = view.addLayer({
  *      type: 'elevation',
  *      id: 'iElevation',
- * });
+ * }).then(() => { .... });
+ *
+ * // One can also attach a callback to the same promise with a layer instance.
+ * layer.whenReady.then(() => { ... });
```
peppsac, line 244: Forgot to adapt to your latest change.

examples/pointcloud.js
```diff
@@ -84,7 +76,19 @@ function showPointcloud(serverUrl, fileName, lopocsTable) {
...
+                pointcloud.counters.displayedCount.toLocaleString() + ' (' +
+                Math.floor(100 * pointcloud.counters.displayedCount / pointcloud.counters.pointCount) + '%) (' +
+                view.mainLoop.gfxEngine.renderer.info.memory.geometries + ')';
+        };
+
+
         view.notifyChange(true);
     }
-    view.addLayer(pointcloud).addEventListener('preprocessed', onLayerPreprocessed);
+    view.addLayer(pointcloud).then(onLayerPreprocessed);
```
peppsac: If you don't mind I believe onLayerReady is better name :)

src/Core/View.js
autra: It's still valid, isn't it?

autra: (I added a .then on line 241. Might not be very clear though)

examples/pointcloud.js
autra: yep

src/Core/View.js
peppsac: Oh right. My bad :)

autra:
rebased and ready.

peppsac (approved):
LGTM (this is the 4th time I write this on this PR...)

PR #416, opened 2017-07-27
feature: add helper functions to create Globe and Planar layers

peppsac:
GlobeView/PlanarView are nice but when you need to create several Planar layers
you have to duplicate a lot of code.

This commit splits layer creation to a helper functions that are exposed to the
user.

Example of what can be done easily with this change (and thanks to PR #386):
1 globe layer and 6 planar layers


gchoqueux (approved):
LGTM

PR #417, opened 2017-07-27
Refacto: replace TextDecoder polyfill with a lighter one

autra:
EDIT: now this PR only change the dependencie to one that only supports utf-8, but is much lighter.

text-encoding dependency is very big. Not very worthy for how we use it. Inspiration: https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js#L665 (thanks @Jeremy-Gaillard)

This brings back our bundle size from 1.4Mo to ~800ko.

@iTowns/reviewers please r?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5377,10 +5377,10 @@
       "integrity": "sha1-IGvo4YiGC1FEJTdebxrom/sB/Y0=",
       "dev": true
     },
-    "text-encoding": {
-      "version": "0.6.4",
-      "resolved": "https://registry.npmjs.org/text-encoding/-/text-encoding-0.6.4.tgz",
-      "integrity": "sha1-45mpgiV6J22uQou5KEXLcb3CbRk="
+    "text-encoding-utf-8": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/text-encoding-utf-8/-/text-encoding-utf-8-1.0.1.tgz",
+      "integrity": "sha1-Uepsen6y+09nRnt2NzVmH1YDSS0="
     },
     "text-table": {
       "version": "0.2.0",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "jszip": "^3.1.3",
     "monotone-convex-hull-2d": "^1.0.1",
     "proj4": "^2.4.3",
-    "text-encoding": "^0.6.4",
+    "text-encoding-utf-8": "^1.0.1",
     "three": "^0.86.0",
     "whatwg-fetch": "^2.0.2"
   },
--- a/src/utils/polyfill.js
+++ b/src/utils/polyfill.js
@@ -1,0 +1,7 @@
+import {
+  TextDecoder as TextDecoderPolyfill,
+  TextEncoder as TextEncoderPolyfill,
+} from 'text-encoding-utf-8';
+
+export const TextDecoder = typeof global.TextDecoder === 'function' ? global.TextDecoder : TextDecoderPolyfill;
+export const TextEncoder = typeof global.TextEncoder === 'function' ? global.TextEncoder : TextEncoderPolyfill;
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -6,8 +6,8 @@ var definePlugin = new webpack.DefinePlugin({
 });
 
 var providePlugin = new webpack.ProvidePlugin({
-    TextDecoder: ['text-encoding', 'TextDecoder'],
-    TextEncoder: ['text-encoding', 'TextEncoder'],
+    TextDecoder: [path.resolve(__dirname, 'src/utils/polyfill'), 'TextDecoder'],
+    TextEncoder: [path.resolve(__dirname, 'src/utils/polyfill'), 'TextEncoder'],
 });
 
 module.exports = {
@@ -26,7 +26,8 @@ module.exports = {
     plugins: [
         definePlugin,
         providePlugin,
-        new webpack.optimize.CommonsChunkPlugin({ name: 'itowns' })],
+        new webpack.optimize.CommonsChunkPlugin({ name: 'itowns' }),
+    ],
     module: {
         rules: [
             {


package-lock.json
```diff
@@ -3979,6 +3979,14 @@
       "integrity": "sha1-w2GT3Tzhwu7SrbfIAtu8d6gbHA8=",
       "dev": true
     },
+    "monotone-convex-hull-2d": {
```
tbroyer: What's the reason for all those additions, for a deleted dependency in `package.json`?

src/utils/Ponyfills.js
```diff
@@ -0,0 +1,15 @@
+// This file is intented to have multiple exports
```
Jeremy-Gaillard: ?

src/utils/Ponyfills.js
```diff
@@ -0,0 +1,15 @@
...
+// eslint-disable-next-line import/prefer-default-export
+export function convertUint8ArrayToString(array) {
+    if (window.TextDecoder !== undefined) {
+        return new TextDecoder().decode(array);
+    }
+
+    // Avoid the String.fromCharCode.apply(null, array) shortcut, which
+    // throws a "maximum call stack size exceeded" error for large arrays.
```
tbroyer: This could at least process the array in chunks, rather than one element at a time.

Maybe it's not worth the added complexity though, if that approach is fast enough.

src/utils/Ponyfills.js
```diff
@@ -0,0 +1,15 @@
...
+    var s = '';
+    for (var i = 0, il = array.length; i < il; i++) {
+        s += String.fromCharCode(array[i]);
```
tbroyer: This is not equivalent to `new TextDecoder().decode(…)` if there are values outside the 0–127 range; because `TextDecoder` defaults to UTF-8 when no encoding is given.

I don't know the usecases, and which inputs will be passed to this function, but's probably worth at least adding a comment (jsdoc?) that values are expected to all fall into the 0–127 range (i.e. be US-ASCII).

Maybe the fast path could also use `new TextEncoded('us-ascii')`? (it'll actually use `windows-1252`, but at least shows the intent)

src/utils/Ponyfills.js
autra: Ponyfill

autra: ([Ponyfills are a thing](https://github.com/sindresorhus/ponyfill))

src/utils/Ponyfills.js
autra: @Jeremy-Gaillard do you know if we'll have value outside the 0-127 range?

src/utils/Ponyfills.js
autra: It's more a fallback for older browser, so yeah, I'm not very concerned about perf here.

package-lock.json
autra: Ah yes, it's because #400 forgot to modify it. #413 fixes it in its last commit, but this one will pick this diff until #413 is merged. Anyway, this is auto-generated, not sure if it is worthy to review it TBH.

src/utils/Ponyfills.js
autra: As this is taken nearly verbatim from THREE, I'll expect this to work correctly for gltf at least.

Jeremy-Gaillard: There does not seem to be any restrictions in the specifications, so it could be any valid JSON object. Though, the overwhelming majority of files will not have any problematic characters.

autra: > This is not equivalent to new TextDecoder().decode(…) if there are values outside the 0–127 range; because TextDecoder defaults to UTF-8 when no encoding is given.

Wait. Why is that so @tbroyer? [String.fromCharCode documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode) says it `returns a string created by using the specified sequence of Unicode values.`

package-lock.json
tbroyer: > Anyway, this is auto-generated, not sure if it is worthy to review it TBH.

It is. If only to detect when it's missing (like in #400). But also in case it differs from what the `package.json` says (like here; or if you use `git add -p` or equivalent to selectively include dependencies in your `package.json` when crafting your commits, etc.)

src/utils/Ponyfills.js
tbroyer: Yep, Unicode code points, so 160 is read as a U+00A0 (no-break space), not as an invalid UTF-8 byte (`10100000` denotes a continuation byte, so it's invalid unless preceeded by a `11xxxxxx` byte) which `TextDecoder` would replace with a U+FFFD (replacement character).

Try `new TextDecoder().decode(Uint8Array.of(160))` vs. `String.fromCharCode(160)`

package-lock.json
autra: > It is. If only to detect when it's missing (like in #400).

Yeah right

> also in case it differs from what the package.json says (like here

Here, the commit put it back in sync with the current state of package.json. But ok, I'll make sure #413 is merged before this one, that'll fix everything and the diff will be pretty :-)

tbroyer: Note that I was only replying to the fact that the lock file should be reviewed. I have my answer regarding those changes and I'm OK to have them in this PR.

…well, provided they are in a commit on their own (as in #413), so +1 to getting #413 merged before this PR :wink:

src/utils/Ponyfills.js
autra: The question is: which one is correct in the context of 3dTiles ? @Jeremy-Gaillard do you have some input on this?

peppsac: Given that `GLTFLoader` already uses this workaround I'd use it as well in iTowns (for now).

Can you add a one-time warning though (something like `warning: fallbacks on String.fromCharCode method. Open an issue in itowns issue tracker if it doesn't work as expected`)?

Jeremy-Gaillard: If it's any help, the glTF specification explicitly refers TextDecoder https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_binary_glTF#content

autra: If we only need UTF-8, we might be able to use https://www.npmjs.com/package/text-encoding-shim (but it seems it also uses fromCharCode at some point, I need to check if it is equivalent to what we've done here).

tbroyer: It's the same as here but adds a pass of `escape()` + `decodeURIComponent()` to "decode UTF-8". Adds some overhead but is safe. We could then either use that shim or the code from this PR.

BTW, JSON is defined to be a stream of UTF-8 bytes IIRC.

autra: I could use also https://github.com/arv/text-encoding-utf-8 but it's not very active (it's a fork from text-encoding with only utf-8 support)

autra: Ok not using https://www.npmjs.com/package/text-encoding-shim because it uses `escape`, which is deprecated. I'm going to test https://github.com/arv/text-encoding-utf-8

peppsac: `text-encoding-utf-8` looks like a good choice indeed

autra:
@iTowns/reviewers  and especially @tbroyer please r?

tbroyer (approved):

PR #418, opened 2017-07-27
fix: don't enable debug features in non debug builds

peppsac:


nosy-b (approved):
Super, all good now at http://www.itowns-project.org/itowns/examples/pointcloud.html

PR #420, opened 2017-07-27
add Style for deployment before Jekyll template

nosy-b:
Temporary style link in example's index with the CSS present in https://github.com/iTowns/itowns.github.io for deployment.
In the near future, it should be templated using Jekyll


autra (approved):
THer

autra:
> THer

I meant LGTM ;-)

PR #423, opened 2017-07-28
Fix add layer

peppsac:
@autra this is a followup of PR 413

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -276,14 +276,14 @@ GlobeView.prototype.addLayer = function addLayer(layer) {
         layer.update = updateLayeredMaterialNodeElevation;
     }
     const layerId = layer.id;
-    View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
+    const layerPromise = View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
 
     this.dispatchEvent({
         type: GLOBE_VIEW_EVENTS.LAYER_ADDED,
         layerId,
     });
 
-    return layer;
+    return layerPromise;
 };
 
 /**
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -100,8 +100,12 @@ const _syncThreejsLayer = function _syncThreejsLayer(layer, view) {
 function _preprocessLayer(view, layer, provider) {
     if (!(layer instanceof Layer) && !(layer instanceof GeometryLayer)) {
         const nlayer = new Layer(layer.id);
-        delete layer.id;
+        // nlayer.id is read-only so delete it from layer before Object.assign
+        const tmp = layer;
+        delete tmp.id;
         layer = Object.assign(nlayer, layer);
+        // restore layer.id in user provider layer object
+        tmp.id = layer.id;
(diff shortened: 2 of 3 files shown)


peppsac:
Commit 1:
```
    fix: do not remove id from user layer object
    
    Deleting id is mandatory for Object.assign to succeed, but we must
    restore it before returning the object to the user.
```

Commit 12
```
    fix: make sure addLayer functions have a consistent API
    
    This is a follow up of PR 413.
    
    GlobeView and PlanarView override View.addLayer so make sure to return
    the original result.
    
    Also fixed View's _preprocessLayer to make sure that the last promise in
    layer.whenReady promises chain returns the actual layer.
```

src/Core/View.js
```diff
@@ -100,8 +100,12 @@ const _syncThreejsLayer = function _syncThreejsLayer(layer, view) {
 function _preprocessLayer(view, layer, provider) {
     if (!(layer instanceof Layer) && !(layer instanceof GeometryLayer)) {
         const nlayer = new Layer(layer.id);
-        delete layer.id;
+        // nlayer.id is read-only so delete it from layer before Object.assign
+        const tmp = layer;
+        delete tmp.id;
         layer = Object.assign(nlayer, layer);
+        // restore layer.id in user provider layer object
+        tmp.id = layer.id;
```
autra, line 108: We have a to-be-reactivated linter rule that forbids modifying parameters. Guess it'll be useful (because bug like that are hard to find).

Also, I'm wondering if all this part shouldn't be in Layer constructor? Then we won't even need to do this stunt: we'll just copy layer fields before creating the read-only property. WDYT?

src/Core/Prefab/GlobeView.js
```diff
@@ -276,14 +276,14 @@ GlobeView.prototype.addLayer = function addLayer(layer) {
...
-    View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
+    const layerPromise = View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
 
     this.dispatchEvent({
         type: GLOBE_VIEW_EVENTS.LAYER_ADDED,
         layerId,
     });
 
-    return layer;
+    return layerPromise;
```
autra, line 286: You could even do

```js
this.dispatchEvent({
    type: GLOBE_VIEW_EVENTS.LAYER_ADDED,
    layerId,
});

return View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
```
it's just mono-thread and synchronous ;-)

src/Core/View.js
```diff
@@ -120,17 +124,17 @@ function _preprocessLayer(view, layer, provider) {
...
+                if (!(preprocessingResult && preprocessingResult.then)) {
+                    preprocessingResult = Promise.resolve();
                 }
             } else {
-                layer.whenReady = Promise.resolve();
+                preprocessingResult = Promise.resolve();
             }
-            layer.whenReady.then(() => {
+            // the last promise in the chain must return the layer
+            layer.whenReady = preprocessingResult.then(() => {
```
autra, line 137: Thanks!

autra (approved):
Thanks for these fixes!

Issue #357, opened 2017-06-29
Split layers rendering on screen as option


nosy-b:
A nice enhancement would be to split the rendering between layers depending on the position on the screen.
So layers should have a property that specify if split is on and its side
example:
![split](https://user-images.githubusercontent.com/158514/27700374-b384ce2e-5cfd-11e7-99bc-df3ca185d2a8.jpg)

peppsac:
You can probably already to this using either:
- 2 View
- or a custom `view.render` method, using the following pseudo-code:
```
enable only layer1
set viewport to the left half width of the screen
render view
enable only layer2
set viewport to the right half width of the screen
render view
```

I think the 2nd method is the easiest, and it would actually make a good example :)

nosy-b:
Ya that's an idea but it should be better in a single pass. Just adding a uniform to tell where is the split position to the tiles shader do the job. (would need more work to set specific layer disposition)
I did it as a test on a branch splitLayers, sadly no time to make it clean and add it to examples directory but should be soon! (after the holidays next week :))
![splitlayer](https://user-images.githubusercontent.com/158514/27745784-fbf35668-5dc4-11e7-9759-4cb35bd5a879.jpg)

peppsac:
Are you're suggesting to add an option to the core tile shader just for this feature?

nosy-b:
It might look a bit messy yes, as it looks more like a post process feature.
But why not add the possibility to the Tile Shader to know if some part of the screen are not usable? In that way we avoid a double pass or a double view, it's even faster than the normal rendering in fact
WDYT?

peppsac:
Forgot to say in my earlier comment but it looks nice indeed.

I don't think performance is what is important here.
The core issue at play is I don't want to bloat the core shaders with random features (render layer X in black and white ? use color as transparency for this another layer ? apply blur for this layer ? etc etc).

I don't expect such a big hit of doing 2 half-rendering anyway - we don't have much geometries.
So I stand by my point: this kind of feature should be implemented using `renderer.setScissor`.

But, if we can find a way to provide an easy customisation of the shaders used for the user I'd be happy to merge it
 Maybe allowing the user to override key part of the shader, like this function: https://github.com/iTowns/itowns2/pull/309/files#diff-0ce27fb5c06a05269926d6968eb4ab32R46 (this is from the atlas PR because this code has been simplified a bit).

peppsac:
Can we close this issue?

PR #424, opened 2017-07-28
Splitted rendering example

peppsac:
This PR add a new example based on issue #357:

split


peppsac:
Ping @nosy-b

nosy-b (changes requested):
Looks good! But OpenSM looks like mixed with white, it's too enlightened. 
Has to do with the frag shader when mixing diffuseColor and layerColor maybe?

peppsac:
The rendering looks identical to this layer original site (original on the top of the screenshot):
capture d ecran de 2017-07-28 16-00-50

Note that I changed the OPENSM layer url since the oold one wasn't working anymore.

nosy-b:
My bad, it's just a light OSM rendering and my screen sucks :)
splitcalib

nosy-b (approved):

PR #426, opened 2017-07-28
refactor: rework Debug tools

peppsac:
- split debug code to various files (1 per chart, 1 per geom type)
- clarify things belonging to layers or views
- support dynamic creation/hiding of graphs

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -98,7 +98,10 @@
             itowns.View.prototype.addLayer.call(globe, $3dTilesLayerRequestVolume);
 
             // Add the UI Debug
-            new debug.Debug(globe, viewerDiv);
+            const d = new debug.Debug(globe, menuGlobe.gui);
+            debug.createTileDebug(menuGlobe.gui, globe, globe.wgs84TileLayer, d);
+            debug.create3dTilesDebugUI(menuGlobe.gui, globe, $3dTilesLayerDiscreteLOD, d);
+            debug.create3dTilesDebugUI(menuGlobe.gui, globe, $3dTilesLayerRequestVolume, d);
 
         </script>
     </body>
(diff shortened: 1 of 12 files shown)


examples/3dtiles.html
```diff
@@ -98,7 +98,10 @@
             itowns.View.prototype.addLayer.call(globe, $3dTilesLayerRequestVolume);
 
             // Add the UI Debug
-            new debug.Debug(globe, viewerDiv);
+            const d = new debug.Debug(globe, menuGlobe.gui);
+            new debug.TileDebug(menuGlobe.gui, globe, globe.wgs84TileLayer, d);
```
autra: If we don't need the created object, and as it seems it never uses `this`, let's expose regular function `createTileDebugUi` instead of constructors.

autra (commented):
Didn't make an extensive review, but looks good on first reading.

examples/3dtiles.html
peppsac: I use it in other debug modules (see TileDebug)

peppsac: I eventually understood your comment and applied your suggested modification :)

utils/debug/TileDebug.js
```diff
@@ -0,0 +1,148 @@
...
+    gui.add(layer, 'wireframe').name('Wireframe').onChange((newValue) => {
+        layer.wireframe = newValue;
+
+        applyToNodeFirstMaterial(view, layer.object3d, layer.id, (material) => {
+            material.wireframe = newValue;
+        });
+    });
+
+    // TileObjectChart visibility
+    gui.add(state, 'objectChart').name('Object chart').onChange((newValue) => {
```
gchoqueux, line 55: I would move `Object Chart` and `visibilty Chart` in Debug tools folder

peppsac:
These charts are `per layer` charts instead of `per view`: that's why their visibility options is in their respective chart folder.

gchoqueux:
Ok, I found it more ergonomic

gchoqueux (approved):

PR #427, opened 2017-07-28
fix: use CDN for OrbitControls for cubic_planar example

peppsac:
@autra please?

### DIFF
--- a/examples/cubic_planar.html
+++ b/examples/cubic_planar.html
@@ -61,8 +61,7 @@
         <script type="text/javascript">
             window.THREE = itowns.THREE;
         </script>
-        <script src="../node_modules/three/examples/js/controls/OrbitControls.js"></script>
-        <script src="GUI/dat.gui/dat.gui.min.js"></script>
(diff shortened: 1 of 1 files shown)


examples/cubic_planar.html
```diff
@@ -61,8 +61,7 @@
         <script type="text/javascript">
             window.THREE = itowns.THREE;
         </script>
-        <script src="../node_modules/three/examples/js/controls/OrbitControls.js"></script>
-        <script src="GUI/dat.gui/dat.gui.min.js"></script>
+        <script src="https://cdn.rawgit.com/mrdoob/three.js/dev/examples/js/controls/OrbitControls.js"></script>
```
tbroyer: Maybe use https://unpkg.com/three@0.86.0/examples/js/controls/OrbitControls.js instead? (or least choose a version that matches our version of three.js, just in case)

peppsac: Good idea, thanks for the suggestion. I modified the PR.

tbroyer (approved):

PR #430, opened 2017-08-01
refactor: rework Tiles removal and cleanup

peppsac:
This commit removes TileMesh specific cleanup code and instead
introduce ObjectRemovalHelper that can be used on any object.

Another change is that attached layers get called on just-removed
elements. This allows each layer to perform its own cleanup.

### DIFF
--- a/src/Process/ObjectRemovalHelper.js
+++ b/src/Process/ObjectRemovalHelper.js
@@ -1,0 +1,70 @@
+export default {
+    /**
+     * Cleanup obj to release three.js allocated resources
+     * @param {Object3D} obj object to release
+     */
+    cleanup(obj) {
+        if (typeof obj.dispose === 'function') {
+            obj.dispose();
+        } else {
+            if (obj.geometry) {
+                obj.geometry.dispose();
+                obj.geometry = null;
+            }
+            if (obj.material) {
+                obj.material.dispose();
+                obj.material = null;
+            }
+        }
+    },
+
+    /**
+     * Remove obj's children belonging to layerId layer.
+     * Neither obj nor its children will be disposed!
+     * @param {String} layerId The id of the layer that objects must belong to. Other object are ignored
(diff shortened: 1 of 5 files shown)


peppsac:
I need to fix one last issue and update `TileDebug.js` as well before this is ready.

peppsac:
OK fixed.
You can test the difference by enabling the BoundingBox on the index.html sample and display the geometry count chart.

Then type:
```
globeView.controls.setRange(5000, true)
globeView.controls.setRange(25000000, true)
```
in the console and observe the variation in the geometry count chart.

Ideally, after going back at the initial position the geometry count should be the same as when itowns was started (= 90).

On master, this count never gets back to 90.
This PR fixes this problem.

peppsac:
@gchoqueux r?

src/Process/ObjectRemovalHelper.js
```diff
@@ -0,0 +1,71 @@
...
+        obj.remove(...toRemove);
+        return toRemove;
+    },
+
+    /**
+     * Recursively remove obj's children belonging to layerId layer.
+     * All removed obj will have their geometry/material disposed.
+     * @param {String} layerId The id of the layer that objects must belong to. Other object are ignored
+     * @param {Object3D} obj The Object3D we want to clean
+     * @param {Boolean} callDispose Whether or not the initial obj should be disposed as well.
```
gchoqueux: callDispose isn't used

gchoqueux (changes requested):
OK 
just one comment

src/Process/ObjectRemovalHelper.js
peppsac: Fixed!

peppsac: For typo like this you can leave a comment but approve the PR.
This way I can fix it and merge directly without waiting for your approval.

gchoqueux (approved):

PR #431, opened 2017-08-01
WFS support

peppsac:
~(still WIP)~
~(based on PR #430 )~
Edit: Ready!

This PR implements basic WFS support.
It's a refactoring / reimplementation of previous WFS PR (PR #195 )

The current focus is:
- write a working WFS provider and cleanup its options
- write a GeoJSON -> three.js simple parser
- implement the required glue between geojson and itowns

What isn't in the scope of this PR:
- positionning geometry based on altitude (because it requires a feature that doesn't exist yet in itowns)
- extruding geometry: it's unrelated to WFS or geojson

I've added a small example, displaying 2 different WFS stream:
wfs


peppsac:
Updated the branch with auto-merge of `multi*`geometries and `FeatureCollections` in a single mesh.

peppsac:
If you want to try WFS on `cubic_planar.html` you can use the following diff: https://gist.github.com/peppsac/d02deee6d44fe5587761eaafd47e728e

peppsac:
WFS on cubic_planar:

WFS on cubic_planar

examples/wfs.js
```diff
@@ -0,0 +1,150 @@
...
+view.camera.camera3D.lookAt(new itowns.THREE.Vector3(1840839, 5172718, 0));
+
+// Request redraw
+view.notifyChange(true);
+
+function colorFunctionLine(layer, node, feature) {
+    feature.traverse(function _(n) {
+        var rgb;
+        if (n.material) {
+            if (n.properties) {
```
gchoqueux: When `n.properties` is define?

examples/wfs.js
```diff
@@ -0,0 +1,150 @@
...
+        }
+    });
+}
+
+view.addLayer({
+    update: itowns.FeatureProcessing.update(colorFunctionLine),
+    url: 'https://download.data.grandlyon.com/wfs/rdata?',
+    protocol: 'wfs',
+    version: '2.0.0',
+    axisOrder: 'wsne',
```
gchoqueux: `axisOrder` is set by the WFS version?

http://docs.geoserver.org/latest/en/user/services/wfs/basics.html

src/Core/Scheduler/Providers/WFS_Provider.js
```diff
@@ -4,61 +4,132 @@
...
-    return Fetcher.json(url, this.networkOptions);
+WFS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer) {
+    if (!layer.typeName) {
+        throw new Error('layer.typeName is required.');
+    }
+
+    layer.axisOrder = layer.axisOrder || 'swne';
+    layer.format = layer.options.mimetype || 'json';
+    layer.crs = layer.projection || 'EPSG:4326';
+    layer.version = layer.version || '1.3.0';
```
gchoqueux: `1.3.0` isn't correct version

http://www.opengeospatial.org/standards/wfs

src/Process/FeatureProcessing.js
```diff
@@ -0,0 +1,104 @@
...
+                    node.layerUpdateState[layer.id].success();
+                    if (!node.parent) {
+                        ObjectRemovalHelper.removeChildrenAndCleanupRecursively(layer.id, result);
+                        return;
+                    }
+                    // We don't use node.matrixWorld here, because feature coordinates are
+                    // expressed in crs coordinates (which may be different than world coordinates,
+                    // if node's layer is attached to an Object with a non-identity transformation)
+                    const tmp = node.extent.center().xyz().negate();
+                    applyOffset(result, tmp);
```
gchoqueux, line 77: Maybe we could return geometries (centred on (0, 0, 0)) with associated coordinates and create mesh here with the right transformation. And remove `applyOffset`

src/Core/Scheduler/Providers/WFS_Provider.js
```diff
@@ -4,61 +4,132 @@
...
-function WFS_Provider(options) {
+function WFS_Provider() {
     this.cache = CacheRessource();
-    this.baseUrl = options.url || '';
-    this.networkOptions = options.networkOptions;
-    this.layer = options.layer || '';
-    this.typename = options.typename || '';
-    this.format = options.format === undefined ? 'json' : options.format;
-    this.epsgCode = options.epsgCode || 4326;
+    this.batiments = [];
```
gchoqueux: `this.batiments` is unused

Do you have a mechanism to avoid duplicates geometry, because WFS queries return data that overlap extents:

see dirty patch, with `this.batiments`

https://github.com/iTowns/itowns/pull/195/files#diff-96c0e22190f6d3f57d176e91dfd7d1ffR130

examples/wfs.js
peppsac: Feature properties are copied here: https://github.com/iTowns/itowns/pull/431/files#diff-6e5ed93be78417d38f4b8eb6d6f2a5e1R230

examples/wfs.js
peppsac: Not really.
But since:
- at the moment we only handle data in the long/lat order (https://github.com/iTowns/itowns/pull/431/files#diff-6e5ed93be78417d38f4b8eb6d6f2a5e1R38)
- and when using EPSG code it seems that long/lat is the default

I propose to remove the `axisOrder` thing for now, and always assume `long/lat` or `x/y` both for building the url and for the returned data. WDYT?

src/Core/Scheduler/Providers/WFS_Provider.js
peppsac: See https://github.com/iTowns/itowns/pull/431/files#diff-6e5ed93be78417d38f4b8eb6d6f2a5e1R202
I only keep the geometries which have their 1st point inside the tile that requested the features.

So no duplicate.
This filtering mechanism is basic but it works for my tests data. We can extend / improve it layer (for instance by allowing the user to supply a `filterFeature` function)

src/Process/FeatureProcessing.js
peppsac: The `GeoJSON2Three` module returns objects with correct transformations and itowns fixes them when importing them in its scene.
I think it's a better idea than tweaking GeoJson coords directly.

But we can revisit this decision later if this code becomes problematic (performance wise maybe).

src/Core/Scheduler/Providers/WFS_Provider.js
peppsac: Correct, switching to 2.0.2

examples/wfs.js
gchoqueux: But feature properties aren't copied in THREE.Mesh or THREE.Group?

examples/wfs.js
```diff
@@ -0,0 +1,148 @@
...
+            color = new itowns.THREE.Color(0x5555ff);
+        } else if (feature.properties.id.indexOf('bati_industriel') === 0) {
+            color = new itowns.THREE.Color(0xff5555);
+        } else {
+            color = new itowns.THREE.Color(0xeeeeee);
+        }
+
+        featureVertices = mesh.featureVertices[i];
+        if (featureVertices) {
+            for (j = 0; j < featureVertices.count; j++) {
```
gchoqueux: If all the points have the same color it's better to change the material's color

peppsac: I have 1 mesh (so 1 material) but N features in it.
That's why I specify color per vertex and not per mesh.

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -0,0 +1,296 @@
...
+
+    // instanciate mesh
+    let result;
+    switch (vertices_and_indices.type) {
+        case 'point':
+            result = new THREE.Points(geometry);
+            break;
+        case 'linestring':
+            result = new THREE.LineSegments(geometry);
+            result.material.side = THREE.DoubleSide;
```
gchoqueux: why you use THREE.DoubleSide?

peppsac: I removed this.

examples/wfs.js
peppsac: They are copied in the returned Mesh: either as mesh.properties (for a single feature) or in mesh.features (for a feature collection)

gchoqueux (approved):
Ok, Thanks

PR #435, opened 2017-08-02
refacto: rework Camera's API

peppsac:
This commit changes the API of Camera as follow:
  - remove duplicate three.js-proxy function (like setRotation)
  - make this class GIS-aware: expose a function to retrieve the position in a
given crs
  - hide resize function and call it from update (since all the callers of resize
effectively call them in pair)

### DIFF
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -1,45 +1,42 @@
 /**
- * Generated On: 2015-10-5
- * Class: Camera
- * Description: La camera scene, interface avec la camera du 3DEngine.
+ * Wrapper around three.js camera to expose some geographic helpers.
  */
 
 import * as THREE from 'three';
+import Coordinates from '../Core/Geographic/Coordinates';
 
-function Camera(width, height, options = {}) {
-    this.ratio = width / height;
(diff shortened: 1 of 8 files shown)


src/Renderer/Camera.js
```diff
@@ -1,62 +1,65 @@
...
-    if (this.camera3D.updateProjectionMatrix) {
-        this.camera3D.updateProjectionMatrix();
+    if (camera.camera3D.updateProjectionMatrix) {
+        camera.camera3D.updateProjectionMatrix();
     }
-};
+}
+
+Camera.prototype.update = function update(width, height) {
+    resize(this, width, height);
```
gchoqueux, line 38: add test to update without resize (globe control example is broken )
```js
if (width && height) {
        resize(this, width, height);
}
```

gchoqueux (changes requested):

src/Renderer/Camera.js
peppsac: I fixed all the callers of `camera.update()` instead.

gchoqueux (approved):

PR #439, opened 2017-08-03
fix (layer) : be able to attach a layer to multiple parents

gchoqueux:
It's possible to attach same color or elevation to multiples geometries layers.
It's used in miniglobe #329 to display same color layer with 2 different scales

### DIFF
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -81,6 +81,8 @@ GeometryLayer.prototype.constructor = GeometryLayer;
 GeometryLayer.prototype.attach = function attach(layer) {
     if (!layer.update) {
         throw new Error(`Missing 'update' function -> can't attach layer ${layer.id}`);
+    } else if (this._attachedLayers.filter(l => l.id === layer.id).length > 0) {
+        throw new Error(`Layer is already attach layer ${layer.id}`);
     }
     this._attachedLayers.push(layer);
 };
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -274,19 +274,23 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
 GlobeView.prototype = Object.create(View.prototype);
 GlobeView.prototype.constructor = GlobeView;
 
-GlobeView.prototype.addLayer = function addLayer(layer) {
-    if (layer.type == 'color') {
-        const colorLayerCount = this.getLayers(l => l.type === 'color').length;
-        layer.sequence = colorLayerCount;
-        layer.update = updateLayeredMaterialNodeImagery;
-    } else if (layer.type == 'elevation') {
-        if (layer.protocol === 'wmts' && layer.options.tileMatrixSet !== 'WGS84G') {
-            throw new Error('Only WGS84G tileMatrixSet is currently supported for WMTS elevation layers');
+GlobeView.prototype.addLayer = function addLayer(layer, parent) {
+    const duplicate = this.getLayers(l => l.id == layer.id);
+    if (duplicate.length === 0) {
+        if (layer.type == 'color') {
+            const colorLayerCount = this.getLayers(l => l.type === 'color').length;
+            layer.sequence = colorLayerCount;
+            layer.update = updateLayeredMaterialNodeImagery;
+        } else if (layer.type == 'elevation') {
+            if (layer.protocol === 'wmts' && layer.options.tileMatrixSet !== 'WGS84G') {
+                throw new Error('Only WGS84G tileMatrixSet is currently supported for WMTS elevation layers');
+            }
+            layer.update = updateLayeredMaterialNodeElevation;
         }
-        layer.update = updateLayeredMaterialNodeElevation;
     }
+
     const layerId = layer.id;
-    const layerPromise = View.prototype.addLayer.call(this, layer, this.wgs84TileLayer);
+    const layerPromise = View.prototype.addLayer.call(this, layer, parent === undefined ? this.wgs84TileLayer : parent);
 
     this.dispatchEvent({
         type: GLOBE_VIEW_EVENTS.LAYER_ADDED,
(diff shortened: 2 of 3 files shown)


src/Core/Layer/Layer.js
```diff
@@ -79,7 +79,7 @@ GeometryLayer.prototype = Object.create(EventDispatcher.prototype);
 GeometryLayer.prototype.constructor = GeometryLayer;
 
 GeometryLayer.prototype.attach = function attach(layer) {
-    if (!layer.update) {
+    if (!layer.update || this._attachedLayers.filter(l => l.id === layer.id).length > 0) {
```
peppsac: You can't reuse the same error message for 2 different errors.

src/Core/Prefab/GlobeView.js
```diff
@@ -274,19 +274,23 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
...
-GlobeView.prototype.addLayer = function addLayer(layer) {
-    if (layer.type == 'color') {
-        const colorLayerCount = this.getLayers(l => l.type === 'color').length;
-        layer.sequence = colorLayerCount;
-        layer.update = updateLayeredMaterialNodeImagery;
-    } else if (layer.type == 'elevation') {
-        if (layer.protocol === 'wmts' && layer.options.tileMatrixSet !== 'WGS84G') {
-            throw new Error('Only WGS84G tileMatrixSet is currently supported for WMTS elevation layers');
+GlobeView.prototype.addLayer = function addLayer(layer, parent) {
+    const duplicate = this.getLayers((l => l.id == layer.id));
```
peppsac: You can remove one parenthesis pair (`getLayers(( ...));`  => `getLayers(...);`)

peppsac: You can remove one parenthesis pair (`getLayers(( ...));`  => `getLayers(...);`)

src/Core/Prefab/GlobeView.js
```diff
@@ -274,19 +274,23 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
...
+    if (duplicate.length === 0) {
```
peppsac: And what happens if `duplicate.length` is not 0?

src/Core/Prefab/GlobeView.js
```diff
@@ -274,19 +274,23 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
...
+        if (layer.type == 'color') {
```
peppsac: And what happens if `duplicate.length` is not 0?

src/Core/View.js
```diff
@@ -250,9 +250,10 @@ function _preprocessLayer(view, layer, provider) {
  * @param {Layer=} parentLayer
  * @return {Promise} a promise resolved with the new layer object when it is fully initialized
  */
+
 View.prototype.addLayer = function addLayer(layer, parentLayer) {
     const duplicate = this.getLayers((l => l.id == layer.id));
-    if (duplicate.length > 0) {
+    if (duplicate.length > 0 && !parentLayer) {
```
peppsac, line 256: The previous test was incorrect, but this one is wrong too.
I believe what we want is not instanciate 2 `Layer` objects with the same id.
So I'd suggest replace:
`&& !parentLayer` with `&& !(layer instanceof Layer)`.
WDYT?

peppsac: (I'm not 100% sure so let's discuss it here before changing your code)

peppsac: We probably should clarify what `id` means for itowns...
IMHO, it's used to identify layers *and* it doesn't make sense to attach the same layer multiple times.

peppsac (changes requested):
Before merging this, we should make sure that everyone has the same opinion of what is an `id`.

Also, your commit title is unclear: I can't tell if the fix is intended to be able to attach a layer to different parents or to make this impossible.

Last but not least: could you add a description in the commit message of why you want to make this change?

src/Core/Prefab/GlobeView.js
gchoqueux: `duplicate.length === 0`  means this layer never added and need preparation. 
`duplicate.length != 0` It is managed in `View.addLayer`, it can be attached to a different layer

src/Core/View.js
gchoqueux: If you want to display the same WMTS on a plan and globe.
We need to duplicate the color layer for the moment. 
I can make a layer clone. WDYT?

src/Core/Prefab/GlobeView.js
peppsac: I don't think so.
If a user does this:
```js
view.addLayer({id: 'my_layer', ....});
view.addLayer({id: 'my_layer', ....});
```
Then the second layer needs preprocessing (or an error depending on the point of view);

A preprocessed layer can be identified because:
- it's a `Layer` (and not an object)
- and it has a `layer.whenReady` property

src/Core/View.js
peppsac: No need for a `clone` method IMHO.

When you want to reuse the same datasource you have 2 alternatives:
- do like `cubic_planar.js` demo (https://github.com/iTowns/itowns/blob/master/examples/cubic_planar.js#L113 : the same *layer* is attached to several geometry layer but with a different id each time (so 1 `Layer` is only attached once). 
Advantage:
    * it already works
    * an `id` identifies a specific Layer and its attach "point"  (ie: you can use this `id` anywhere in itowns and it will only affect one thing

- we can allow attaching the same `Layer`object multiple times.

  Advantages:
    * Shorter syntax

  Risks:
    * one `id` now refers to distinct visual objects (so modifying 1 will affect others)
    * the syntax might be a bit weird, since the original `addLayer` do not return the `Layer` object but only its `whenReady` promise.

gchoqueux: I think I'll close this PR, because we have to think about it.

But I think that the rendering properties {visible, color, elevation, opacity} and the data {wmts, image, wms} should be separated. Like materials and textures, the texture can be reused in several materials. WDYT?

peppsac: I dunno.
That's why I'm writing `examples`: to experiment with itowns API and discover where it needs improvments/changes :)

But yes we can close this one for now.

PR #440, opened 2017-08-04
Reduce memory usage

peppsac:
3 commits regarding memory usage.

The first 2 reduces default memory usage of iTowns from over 130MB to 50MB.

The last one modifies API to match three.js/cesium convention to allow a caller of a method to specify a `target` where the result should be written instead of forcing a new allocation each time.
The effects of this last commit are difficult to measure but IMHO reducing temporary allocations count should be a win because it reduces pressure on the GC.


src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -30,21 +39,24 @@ BuilderEllipsoidTile.prototype.Prepare = function Prepare(params) {
...
 BuilderEllipsoidTile.prototype.VertexPosition = function VertexPosition(params) {
-    params.cartesianPosition = new C.EPSG_4326_Radians(
-        params.projected.longitudeRad,
-        params.projected.latitudeRad).as('EPSG:4978');
-    return params.cartesianPosition;
+    this.tmp.coords[0]._values[0] = params.projected.longitudeRad;
+    this.tmp.coords[0]._values[1] = params.projected.latitudeRad;
+
+    this.tmp.coords[0].as('EPSG:4978', this.tmp.coords[1]).xyz(this.tmp.position);
+    this.tmp.normal.copy(this.tmp.position).normalize();
```
gchoqueux, line 53: It would be necessary to fix this normal, because is normal to sphere and not ellipsoid.

I started in this commit https://github.com/iTowns/itowns/commit/75a5b87a2a9f907a08c557bf3ba5e48e9955efb5#diff-e74011929d1712ee6adcb01a09dd097aR51

The correct normal is computed in `Ellipsoid.cartographicToCartesian`.
You can change now or I do a new PR

peppsac: A new PR is best for this fix.

src/Core/Prefab/Globe/Atmosphere.js
```diff
@@ -151,7 +153,6 @@ function Atmosphere() {
     this.skyDome.mesh.material.depthWrite = false;
     this.add(this.skyDome.mesh);
```
gchoqueux: `this.skyDome.mesh` have wrong camera's layer

peppsac: Yes, this needs fixing.

peppsac: Fixed in #442

src/Core/Prefab/GlobeView.js
```diff
@@ -220,7 +219,6 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
 
     // Atmosphere
     this.atmosphere = new Atmosphere();
-    this.clouds = new Clouds();
     this.atmosphere.add(this.clouds);
```
gchoqueux: remove this line

peppsac: Oops... fixed

src/Core/Prefab/Globe/Atmosphere.js
gchoqueux: In `_initRealisticLighning` the new objects need to set camera's layer?

peppsac: Correct, I'll fix this.

gchoqueux (approved):

PR #445, opened 2017-08-08
  feature: add new FirstPersonControls

peppsac:
(depends on #444)

This control is very similar to FlyControls but the major difference is that
in this control the camera cannot roll.
This is handy in app using a planar mode where the roll movement is never wanted.

This control can also be used to create a PanoramaControls, just set moveSpeed to 0.

At some point we'll probably refactor all our controls under a single class exposing
the different behavior.

Edit:
Here are the advantages I see at including controls in iTowns (instead of using existing ones):

- integrates better with iTowns architecture. Especially the existing controls are expecting a continuous update loop while we have a pausable one (so our controls use .notifyChange when needed)
-   in a second step we'll be able to hide the different controls under a single configurable Control thingy - instead of asking our user to find existing controls and plug them in itowns
-  last but not least: merging our controls doesn't prevent the user to use some existing three.js controls anyway.

### DIFF
--- a/src/Renderer/ThreeExtended/FirstPersonControls.js
+++ b/src/Renderer/ThreeExtended/FirstPersonControls.js
@@ -1,0 +1,131 @@
+import * as THREE from 'three';
+
+const MAX_FOV = 90;
+
+// Note: we could use existing three.js controls (like https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/FirstPersonControls.js)
+// but including these controls in itowns allows use to integrate them tightly with itowns.
+// Especially the existing controls are expecting a continuous update loop while we have a pausable one (so our controls use .notifyChange when needed)
+
+function onDocumentMouseDown(event) {
+    event.preventDefault();
+    this._isUserInteracting = true;
+
+    this._onMouseDownMouseX = event.clientX;
+    this._onMouseDownMouseY = event.clientY;
+    this._onMouseDownPhi = this._phi;
+    this._onMouseDownTheta = this._theta;
+}
+
+function onDocumentMouseMove(event) {
+    if (this._isUserInteracting === true) {
+        const fovCorrection = this.camera.fov / MAX_FOV; // 1 at MAX_FOV
+        this._phi = (this._onMouseDownMouseX - event.clientX) * 0.13 * fovCorrection + this._onMouseDownPhi;
+        this._theta = (event.clientY - this._onMouseDownMouseY) * 0.13 * fovCorrection + this._onMouseDownTheta;
+        this.view.notifyChange(false);
+    }
+}
+
+function onDocumentMouseUp() {
+    this._isUserInteracting = false;
+}
+
+function onKeyUp(e) {
+    const move = MOVEMENTS[e.keyCode];
+    if (move) {
+        this.moves.delete(move);
+        this.view.notifyChange(true);
+        e.preventDefault();
+    }
+}
+
+function onKeyDown(e) {
+    const move = MOVEMENTS[e.keyCode];
+    if (move) {
+        this.moves.add(move);
+        this.view.notifyChange(false);
+        e.preventDefault();
+    }
+}
+
+const MOVEMENTS = {
+    38: { method: 'translateZ', sign: -1 }, // FORWARD: up key
+    40: { method: 'translateZ', sign: 1 }, // BACKWARD: down key
+    37: { method: 'translateX', sign: -1 }, // STRAFE_LEFT: left key
+    39: { method: 'translateX', sign: 1 }, // STRAFE_RIGHT: right key
+    33: { method: 'translateY', sign: 1 }, // UP: PageUp key
+    34: { method: 'translateY', sign: -1 }, // DOWN: PageDown key
+};
+
+
+class FirstPersonControls extends THREE.EventDispatcher {
+    constructor(view, options = {}) {
+        super();
+        this.camera = view.camera.camera3D;
+        this.view = view;
+        this.moves = new Set();
+        this.moveSpeed = options.moveSpeed || 10; // backward or forward move speed in m/s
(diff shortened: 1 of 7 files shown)


src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -0,0 +1,129 @@
...
+        this.moveSpeed = options.moveSpeed || 10; // backward or forward move speed in m/s
+        this._isUserInteracting = false;
+        this._onMouseDownMouseX = 0;
+        this._onMouseDownMouseY = 0;
+        this._onMouseDownLon = 0;
+        this._onMouseDownLat = 0;
+
+        // init from camera rotation
+        this.camera.rotation.reorder('ZYX');
+        this._lon = -THREE.Math.radToDeg(this.camera.rotation.z);
```
gchoqueux: `_long` and `_lat` are more geographic params
May be use `_ph`i or `_theta`  (or `heading` and `tilt`)

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -0,0 +1,129 @@
...
+        this._lat = THREE.Math.radToDeg(this.camera.rotation.x) + 90;
+        this.updateAngles();
+
+        const domElement = view.mainLoop.gfxEngine.renderer.domElement;
```
gchoqueux: It would be necessary to think about factoring this `view.mainLoop.gfxEngine.renderer` in a PR
It begins to be used several times
`view.renderer()`

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -0,0 +1,129 @@
...
+
+        this.view.notifyChange(true);
+    }
+
+    update(dt, updateLoopRestarted) {
+        // if we are in a keypressed state, then update position
+
+        // dt will not be relevant when we just started rendering, we consider a 1-frame move in this case
+        if (updateLoopRestarted) {
+            dt = 16;
```
gchoqueux, line 114: dt is delta Time?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -0,0 +1,129 @@
...
+
+        for (const move of this.moves) {
+            if (move.method === 'translateY') {
+                this.camera.position.z += move.sign * this.moveSpeed * dt / 1000;
+            } else {
+                this.camera[move.method](move.sign * this.moveSpeed * dt / 1000);
+            }
+        }
+
+        if (this.moves.size > 0 || this._isUserInteracting) {
```
gchoqueux: `> 0 ` is necessary?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -0,0 +1,129 @@
...
+    this._onMouseDownMouseX = event.clientX;
+    this._onMouseDownMouseY = event.clientY;
+    this._onMouseDownLon = this._lon;
+    this._onMouseDownLat = this._lat;
+}
+
+function onDocumentMouseMove(event) {
+    if (this._isUserInteracting === true) {
+        const fovCorrection = this.camera.fov / MAX_FOV; // 1 at MAX_FOV
+        this._lon = (this._onMouseDownMouseX - event.clientX) * 0.13 * fovCorrection + this._onMouseDownLon;
```
gchoqueux: what is 0.13?

gchoqueux (commented):
it's good
I added some comments

I think `FirstPersonControls` doesn't represent the type of control because the camera isn't clamped to the terrain.

When we advance with a movement of mouse,
I feel that the right and left are reversed.

gchoqueux:
there's npm packages control: 

https://www.npmjs.com/package/three-first-person-controls

https://www.npmjs.com/package/three-fly-controls

gchoqueux:
Example with three-first-person-controls : 
https://threejs.org/examples/?q=terrain#webgl_geometry_terrain

THREE calls it first person control, But the difference is minimal with fly control 
https://threejs.org/examples/?q=fly#misc_controls_fly

src/Renderer/ThreeExtended/FirstPersonControls.js
peppsac: Yes.

src/Renderer/ThreeExtended/FirstPersonControls.js
peppsac: Sure it could have been written `if (this.moves.size || ...)` but I don't think the `> 0` harms readability.

peppsac:
> I think FirstPersonControls doesn't represent the type of control because the camera isn't clamped to the terrain.

Clamping is a different concept IMHO. FirstPersonControls simply refers to the classic control scheme: forward/backward/strafe and orient camera with mouse.

Regarding `three-first-person-controls`: I'm aware of the existing controls.
Here are the advantages I see at including similar controls in iTowns:
- integrates better with iTowns architecture. Especially the existing controls are expecting a continuous update loop while we have a pausable one (so our controls use `.notifyChange` when needed)
- in a second step we'll be able to hide the different controls under a single configurable `Control` thingy - instead of asking our user to find existing controls and plug them in itowns
- last but not least: merging our controls doesn't prevent the user to use some existing three.js controls anyway.

gchoqueux:
the advantages must be added in the PR and the FirstPersonControls file

src/Renderer/ThreeExtended/FirstPersonControls.js
peppsac: Fixed.

peppsac:
> the advantages must be added in the PR and the FirstPersonControls file

Done

gchoqueux (approved):

PR #446, opened 2017-08-08
3dtiles debug bounding

NikoSaul:


utils/debug/3dTilesDebug.js
```diff
@@ -31,14 +31,17 @@ export default function create3dTilesDebugUI(datDebugTool, view, layer) {
...
-                    helper.updateMatrixWorld();
                 }
                 // 3dtiles with box
                 if (node.boundingVolume.box) {
                     const size = node.boundingVolume.box.getSize();
+                    const center = node.boundingVolume.box.getCenter();
                     const g = new THREE.BoxGeometry(size.x, size.y, size.z);
                     const material = new THREE.MeshBasicMaterial({ wireframe: true });
                     helper = new THREE.Mesh(g, material);
+                    helper.position.x += center.x;
```
peppsac: You can replace this with: 
```js
node.boundingVolume.box.getCenter(helper.position);
```

src/Process/3dTilesProcessing.js
```diff
@@ -96,7 +96,6 @@ export function pre3dTilesUpdate(context, layer) {
 function computeNodeSSE(camera, node) {
     if (node.boundingVolume.region) {
         const worldCoordinateCenter = node.boundingVolume.region.centerWorld;
-        worldCoordinateCenter.applyMatrix4(node.matrixWorld);
```
peppsac: This won't work if your 3d-tiles object is attached to an object with a non-identity transform.
That's because `centerWorld` isn't in world coords (aka three.js coords): it's in WGS84-world-coords.

peppsac: I'd suggest you to use the same pattern as `TileProvider`:
- OBB expressed in tile local coords
- and add OBB as a tile's child (see TiledNodeProcessing also)

peppsac: And to transform in tile local coords I think you'd only need to apply the inverse of the transform defined in `tileset.json`...

peppsac (approved):

PR #447, opened 2017-08-09
fix (Fog) : re active fog in globeView

gchoqueux:


src/Core/Prefab/GlobeView.js
```diff
@@ -247,6 +248,8 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
         v.setFromMatrixScale(wgs84TileLayer.object3d.matrixWorld);
         var lim = v.x * size * 1.1;
 
+        this.fogDistance = mfogDistance * Math.pow((len - size * 0.99) * 0.25 / size, 1.5);
```
peppsac: Where do these formula come from?

src/Process/TiledNodeProcessing.js
```diff
@@ -126,7 +126,7 @@ export function processTiledGeometryNode(cullingTest, subdivisionTest) {
 
             if (node.material.visible) {
                 // update uniforms
-                node.setFog(1000000000);
+                node.setFog(context.view.fogDistance);
```
peppsac, line 129: I'm not sure it's the best place to store `fogDistance`...
Maybe it would make sense in `Camera` but I have no strong opinion about this.

peppsac (commented):
This `fog` feature is only active when log depth buffer is enabled:
```glsl
        #if defined(USE_LOGDEPTHBUF) && defined(USE_LOGDEPTHBUF_EXT)
            float depth = gl_FragDepthEXT / gl_FragCoord.w;
            float fogIntensity = 1.0/(exp(depth/distanceFog));
        #else
            float fogIntensity = 1.0;
        #endif
```
and for `GlobeView`.

Don't you think it would a good idea to fix these 2 limitations?

Last but not least: your PR lacks a before|after screenshot :)

src/Core/Prefab/GlobeView.js
gchoqueux: The old formula was deleted from code

gchoqueux:
This PR restores a broken stuff
We will see to add other features

peppsac (approved):
You're restoring a broken feature but:
- it's still broken in PlanarMode so now it's half-broken. Is it really better?
- the used formula could used 1 or 2 lines of comments :-/
```js
fog = size * 160 * Math.pow((len - size * 0.99) * 0.25 / size, 1.5)
```

Anyway, feel free to merge this if you need it.

peppsac:
@gchoqueux could you update this PR please? (I assume you still want to merge it?)

peppsac:
Btw, this PR broke all planar modes examples.

gchoqueux:
the problem is identified:  `planar.fogDistance` is undefined

now how fix it in PR #550

autra:
Ok thanks.

When something is broken, I think we should adopt the following process:
- revert the faulty commit (this PR for instance)
- create a PR with the fixed commit *only* (don't mix it with features or other fix, like in #550)

That way, master (and our website) stays functional at all time, and we guarantee that the fix is merged quickly. Now, things are broken, and we have to wait until the feature is ready to get merged to have the fix. Not good!

PR #449, opened 2017-08-09
feature: add DEMUtils module

peppsac:
This module has helpers methods to read the elevation used by the displayed tiles
at some specific coords.

It doesn't return the absolute correct value but instead the displayed one, which
is what we really want most of the time to be able to position objects correctly.

It supports both float texture (xbil) and color textures.

See also issue #181 and PR #312 for the previous attempt at solving this.

Here's an example. It's a simple plane that got its geometry modified in one call using:
```js
DEMUtils.placeObjectOnGround(
    view.tileLayer,
    view.referenceCrs,
    obj,
    { modifyGeometry: true, offset: 0 });
```
capture d ecran de 2017-08-09 17-05-36

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -461,6 +461,34 @@ Coordinates.prototype.as = function as(crs, target) {
     return _convert(this, crs, target);
 };
 
+/**
+ * Returns the normalized offset from top-left in extent of this Coordinates
+ * e.g: extent.center().offsetInExtent(extent) would return (0.5, 0.5).
+ * @param {Extent} extent
+ * @param {Vector2} target optional Vector2 target. If not present a new one will be created
+ * @return {Vector2} normalized offset in extent
+ */
+Coordinates.prototype.offsetInExtent = function offsetInExtent(extent, target) {
+    if (this.crs != extent.crs()) {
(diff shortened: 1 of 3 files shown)


peppsac:
Updated the PR:
- DEMUtils now support an optionnal cache to avoid the cost of reading again and again in texture
- added a `Coordinates.offsetInExtent` method (instead of abusing `Extent.offsetInParent`)

peppsac:
If you want to play with this here's a demo: https://gist.github.com/peppsac/8197e9419a6f77637daf0919f49a7dac

Here's what it does:
- 100 objects positionned by changing their matrix
- 1 plane using a THREE.Geometry positionned by modifing geometry.vertices
- 1 ring using a THREE.BufferGeometry positionned by modifing geometry.attributes.position

src/Core/Geographic/Coordinates.js
```diff
@@ -461,6 +461,27 @@ Coordinates.prototype.as = function as(crs, target) {
     return _convert(this, crs, target);
 };
 
+Coordinates.prototype.offsetInExtent = function offsetInExtent(extent, target) {
```
gchoqueux: documention?

gchoqueux (approved):

PR #214, opened 2016-11-21
Web Feature Service part 2

nosy-b:
What this PR does:

* Load/parse a KML (a small sample can be found in exemples dir) 
* clamp lines and polygon to the ground
* generate icons and text boxes for 2D point 
* handle picking on polygons showing its attributes

_This Pull Request should have come after WFS integration #195  but that's a way to put pressure ;)_

PR #450, opened 2017-08-10
feature (Imagery Layer) : add tool/provider to convert vector data to image

gchoqueux:
This allows you to generate color layers from vector datas files
Support files : `gpx`, `geojson`, `kml`

Add a new provider : `VectorToTexture_Provider`
it's convert vector datas to textures
it supports only `geosjon`, for the other formats (kml, gpx) it uses package `togeojson`

I open this PR to discuss to draw vector data via Textures
It's use refacto of #214, #431 and #195 

GPX
gpx

Geojson
geojson

KML
kml

Planar mode : KML + Geojson
planar


gchoqueux:
This PR can be reviewed
Examples must be moved to new files and the data moved to the samples repository.

examples/planar.js
```diff
@@ -54,6 +54,41 @@ view.addLayer({
         mimetype: 'image/jpeg',
     },
 });
+
+itowns.Fetcher.xml('./lyon.kml').then(function _(file) {
+    var kmlLayer = {
+        type: 'color',
+        file: file,
+        protocol: 'vectortocolor',
```
peppsac: +1 much better than adding a fake `wms` protocol

src/Core/Geographic/Coordinates.js
```diff
@@ -423,7 +423,19 @@ Coordinates.prototype.z = function z() {
...
+
+/**
+ * The function converts Coordinates to Vector3
+ * Vector3.x: longitude or x value
+ * Vector3.y: latitude or y value
+ * Vector3.z: altitude or z value
+ * @param      {Vector3}  optionalTarget  if there's optionalTarget the result is copy in it
+ * @return     {Vector3}  return the convertion to Vectot3
+ */
+Coordinates.prototype.toVector3 = function toVector3(optionalTarget) {
```
peppsac: Why do you need to introduce this function?

src/Core/Geographic/Extent.js
```diff
@@ -336,4 +336,28 @@ Extent.prototype.merge = function merge(extent) {
     }
 };
 
+/**
+ * mergeCoordinate performs the minimum extension for
+ * a coordinates to belong to the extended
+ *
+ * @param {Coordinates} coordinates  The coordinates to belong
+ */
+Extent.prototype.mergeCoordinate = function mergeCoordinate(coordinates) {
```
peppsac: I'd suggest using the same verb than three.js for easier understanding. [Box3](https://threejs.org/docs/#api/math/Box3) use `expandByPoint` so I propose `expandByCoordinate`.

src/Core/Scheduler/Providers/VectorToColor_Provider.js
```diff
@@ -0,0 +1,136 @@
...
+        layer.extent = getExtentFromGpxFile(layer.file);
+        layer.geojson = GeoJSON2Three.parse(layer.projection, geojson, layer.extent, options);
+    }
+};
+
+VectorToColor_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
+    return tile.level >= layer.options.zoom.min && tile.level <= layer.options.zoom.max && layer.extent.intersect(tile.extent);
+};
+
+VectorToColor_Provider.prototype.getColorTexture = function getColorTexture(tile, layer) {
```
peppsac: `getColorTexture` name is a copy-paste from other providers.
Maybe consider a more descriptive name?

src/Core/Scheduler/Providers/VectorToColor_Provider.js
```diff
@@ -0,0 +1,136 @@
...
+        const west = bound.getAttribute('minlon');
+        const east = bound.getAttribute('maxlon');
+        const south = bound.getAttribute('minlat');
+        const north = bound.getAttribute('maxlat');
+        return new Extent('EPSG:4326', west, east, south, north);
+    }
+    return new Extent('EPSG:4326', -180, 180, -90, 90);
+}
+
+function VectorToColor_Provider() {
```
peppsac: We started using the `class` syntax in other providers so you might want to use it here as well.

src/Core/Scheduler/Scheduler.js
```diff
@@ -91,9 +92,9 @@ Scheduler.prototype.initDefaultProviders = function initDefaultProviders() {
     this.addProtocolProvider('tms', new TMS_Provider());
     this.addProtocolProvider('potreeconverter', PointCloudProvider);
     this.addProtocolProvider('wfs', new WFS_Provider());
+    this.addProtocolProvider('vectortocolor', new VectorToColor_Provider());
```
peppsac: What do you think of `rasterizer` instead of `vectortocolor`?

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -291,13 +332,15 @@ function readFeatureCollection(crsIn, crsOut, json, filteringExtent) {
 }
 
 export default {
-    parse(crsOut, json, filteringExtent) {
-        const crsIn = readCRS(json);
+    parse(crsOut, json, filteringExtent, options = { toMesh: true }) {
```
peppsac: I need to think more about this but I don't like this change because using a module named `GeoJSON2Three` to *not* produce a three object is kind of weird.

You probably want to split this file in 3 files: a Geojson parser producing *something*, a `somethingToThree` module and a `somethingToTexture` module.

peppsac (commented):
Looks good!

I did a quick first review, I'll get back to this PR soon.

index.html
```diff
@@ -34,32 +34,77 @@
...
+                });
+
+                itowns.Fetcher.json(urlGeojson).then(function (file) {
+                    const geojsonLayer = {
+                        type: 'color',
+                        file,
+                        protocol: 'vectortocolor',
+                        id: 'ariege',
+                        projection: 'EPSG:4326',
+                        crsFile: 'EPSG:4326',
```
peppsac: `crsFile` looks like a duplicate of `projection`. Are they different?

src/Core/Geographic/Extent.js
```diff
@@ -336,4 +336,28 @@ Extent.prototype.merge = function merge(extent) {
...
+    const coords = coordinates.as(this.crs());
+    const longitude = this._internalStorageUnit == 2 ? coords.x(this._internalStorageUnit) : coords.longitude(this._internalStorageUnit);
```
peppsac: Please replace `== 2` by `== UNIT.METER`

src/Core/Geographic/Coordinates.js
gchoqueux: to build vector3 with degree unit Coordinates

src/Core/Geographic/Extent.js
So I change `merge` by `expandByExtent`?

src/Core/Geographic/Coordinates.js
peppsac: Here's an alternative name then: `rawValues()`
This would return a `Vector3` but as a user its clear that there's no conversion occuring.

src/Core/Geographic/Extent.js
peppsac: See `Box3`: they have a `union` method to merge 2 boxes so yes, you could rename `merge` -> `union`

src/Core/Scheduler/Providers/VectorToColor_Provider.js
gchoqueux: Ok

Ps : There's no provider using class

src/Core/Scheduler/Scheduler.js
gchoqueux: yes

index.html
gchoqueux: Yes I'm not sure about this.
It is more the projection of destination,
I can rename it to `crsOut` 
or automatically assign it when adding the layer in the `addLayer` function

src/Core/Scheduler/Providers/VectorToColor_Provider.js
peppsac: > Ps : There's no provider using class

Yeah.
But here's the process you're going to go through:
- convert `VectorToColor_Provider` to `class` syntax
- get errors from eslint because your methods won't be using `this`
- rewrite `VectorToColor_Provider` without `class` or `prototype` (see `PointCloudProvider`)

index.html
peppsac: `projection` should keep it's current meaning = `crs` of the data of this layer.
I don't think you need to introduce another property...

```js
    layer.geojson = GeoJSON2Three.parse(layer.projection, layer.file, layer.extent, options); 
```
is wrong, because the first parameter should be the attached geometry crs.

```js
options.crsIn = options.crsIn || readCRS(json);
```
Should use layer.projection instead.

IMHO you should try to add gpx file to the Lyon example to see it works properly.

src/Renderer/ThreeExtended/GeoJSON2Three.js
gchoqueux: I agree with you
We need to valid architecture before I start the changes

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -39,66 +40,82 @@ function readCoordinates(crsIn, crsOut, coordinates) {
...
+
     return out;
 }
 
-// Helper struct that returns an object { type: "", vertices: [...], indices: [...]}:
+// Helper struct that returns an object { type: "", vertices: [...], indices: [...], extent}:
 // - type is the geom type
 // - vertices is an array of THREE.Vector3
 // - indices is optional, and are currently only used for polygons
+// - extent is optional, it's extent vertices
```
peppsac: `vertices' extent`

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -39,66 +40,82 @@ function readCoordinates(crsIn, crsOut, coordinates) {
...
 // Multi-* geometry types are merged in one.
 const GeometryToVertices = {
-    point(crsIn, crsOut, coordinates, filteringExtent) {
-        const coords = readCoordinates(crsIn, crsOut, coordinates);
+    point(crsIn, crsOut, coordinates, filteringExtent, options) {
+        const extent = options.buildExtent ? new Extent(crsOut, Infinity, -Infinity, Infinity, -Infinity) : 0;
```
peppsac: Maybe use `undefined` instead of `0` when `buildExtent` is `false`.

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -141,10 +166,10 @@ const GeometryToVertices = {
         return result;
     },
 
-    multiPolygon(crsIn, crsOut, coordinates, filteringExtent) {
+    multiPolygon(crsIn, crsOut, coordinates, filteringExtent, options) {
         let result;
         for (const polygon of coordinates) {
-            const p = this.polygon(crsIn, crsOut, polygon, filteringExtent);
+            const p = this.polygon(crsIn, crsOut, polygon, options);
```
peppsac: polygon function signature is: `polygon(crsIn, crsOut, coordinates, filteringExtent, options)` so there's something wrong here.

peppsac: And this `options` object seems only to be used to store `buildExtent` or `toMesh`... cf my earlier comment: I don't think this is a good idea.

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -39,66 +40,82 @@ function readCoordinates(crsIn, crsOut, coordinates) {
...
+            for (const vertex of contour) {
+                vertices2.push(vertex.rawValues());
+            }
         }
 
-        const triangles = Earcut(vertices, null, 3);
-        // TODO: handle holes
-
-        return { type: 'polygon', vertices: vertices2, indices: triangles };
+        return { type: 'polygon', vertices: vertices2, indices: triangles, extent };
```
peppsac: Don't you need `indices` when `options.toMesh` is false?

src/Renderer/ThreeExtended/GeoJSON2Three.js
```diff
@@ -258,28 +283,44 @@ function readFeatureCollection(crsIn, crsOut, json, filteringExtent) {
             lines: collec.filter(c => c.geometry.type === 'linestring'),
             polygons: collec.filter(c => c.geometry.type === 'polygon'),
         };
-
-        const result = new THREE.Group();
+        const result = options.toMesh ? new THREE.Group() : { children: [], add: function add(c) { this.children.push(c); } };
```
peppsac: This line screams "refactor me please" :)

src/Renderer/ThreeExtended/GeoJSON2Three.js
peppsac: I'd start with the suggestion I made earlier and see if it works (using the object defined at https://github.com/iTowns/itowns/pull/450/files#diff-6e5ed93be78417d38f4b8eb6d6f2a5e1R52 as the intermediate representation)

peppsac (commented):
Still some work to do:
- for crs: https://github.com/iTowns/itowns/pull/450#discussion_r135230136
- and in the geojson to ... code

src/Core/Geographic/Coordinates.js
gchoqueux: ok

src/Core/Geographic/Extent.js
gchoqueux: done

src/Core/Scheduler/Providers/VectorToColor_Provider.js
gchoqueux: done

src/Core/Scheduler/Providers/VectorToColor_Provider.js
gchoqueux: done

src/Renderer/ThreeExtended/GeoJSON2Three.js
gchoqueux: done

src/Renderer/ThreeExtended/GeoJSON2Three.js
gchoqueux: done

index.html
gchoqueux: I push a proposition https://github.com/iTowns/itowns/pull/450/commits/25fd9556ef713c0487135d3391f25c47269a6a18

gchoqueux: with gpx test in planar mode

src/Renderer/ThreeExtended/GeoJSON2Three.js
gchoqueux: @peppsac : Ok, done, I split GeoJSON2Three, the Pr is ready

gchoqueux:
@peppsac PR is updated

examples/planar_vector.js
```diff
@@ -0,0 +1,123 @@
...
+    colorTextureElevationMaxZ: 240,
+};
+
+view.camera.setPosition(new itowns.Coordinates('EPSG:3946', extent.west(), extent.south(), 2000));
+// Then look at extent's center
+view.camera.camera3D.lookAt(extent.center().xyz());
+
+// instanciate controls
+// eslint-disable-next-line no-new
+new itowns.FirstPersonControls(view, { focusOnClick: true, moveSpeed: 1000 });
```
autra: Just a reminder, we have now a PlanarControls, if you prefer (I'm fine with either of them, your call :-)).

src/Core/Geographic/Coordinates.js
```diff
@@ -423,11 +423,34 @@ Coordinates.prototype.z = function z() {
...
+
+/**
+ * The function returns the coordinates's values without convertion in Vector3
+ * Vector3.x: longitude or x value
+ * Vector3.y: latitude or y value
+ * Vector3.z: altitude or z value
+ * @param      {Vector3}  optionalTarget  if there's optionalTarget the result is copy in it
+ * @return     {Vector3}  return the coordinates's values without convertion in Vector3
+ */
+Coordinates.prototype.rawValues = function rawValues(optionalTarget) {
```
autra: Can you elaborate on why this function? (It seems to me you only use it once)

src/Core/Geographic/Extent.js
```diff
@@ -336,4 +336,29 @@ Extent.prototype.merge = function merge(extent) {
     }
 };
 
+/**
+ * expandByPoint performs the minimum extension for
+ * a coordinates to belong to the extended
```
autra: I suggest 

> expandByPoint perfoms the minimal extension for the point to belong to this Extent object

src/Core/Prefab/GlobeView.js
```diff
@@ -276,6 +276,9 @@ GlobeView.prototype.addLayer = function addLayer(layer) {
         const colorLayerCount = this.getLayers(l => l.type === 'color').length;
         layer.sequence = colorLayerCount;
         layer.update = updateLayeredMaterialNodeImagery;
+        if (layer.protocol === 'rasterizer') {
+            layer.reprojection = 'EPSG:4326';
```
autra, line 280: can you elaborate why `EPSG:4326`? Intuitively I would say that reprojection should always be the view CRS. Is there any case where this would not be the case?

If not, then all the uses of `layer.reprojection` can be replaced by `view.referenceCrs` in `VectorToColor_Provider.preprocessDataLayer` (you'll need to add view as second parameter there. All `preprocessDataLayer` are called with it, but it is ignored in this particular case).

gchoqueux:
@autra can you valid the last 2 works from @peppsac https://github.com/iTowns/itowns/pull/450#pullrequestreview-58633748 ?

Still some work to do:
- for crs: https://github.com/iTowns/itowns/pull/450#discussion_r135230136 -> see 8713049e
- and in the geojson to ... code -> see 25fd955

src/Core/Prefab/PlanarView.js
```diff
@@ -153,6 +153,9 @@ PlanarView.prototype.constructor = PlanarView;
 PlanarView.prototype.addLayer = function addLayer(layer) {
     if (layer.type == 'color') {
         layer.update = updateLayeredMaterialNodeImagery;
+        if (layer.protocol === 'rasterizer') {
+            layer.reprojection = this.referenceCrs;
```
autra, line 157: Ok here you use the referenceCrs. But I believe you don't need an additional property, see above (I'm pretty sure we always have a reference to the view in preUpdate and update methods anyway).

src/Core/Scheduler/Providers/VectorToColor_Provider.js
```diff
@@ -0,0 +1,129 @@
+/**
+ * Class: VectorToColor_Provider
```
autra: As the protocol was renamed from `vectortocolor` to `rasterizer`, I suggest renaming this file to `RasterizerProvider` too.

src/Core/Scheduler/Providers/VectorToColor_Provider.js
```diff
@@ -0,0 +1,129 @@
...
+        const east = bound.getAttribute('maxlon');
+        const south = bound.getAttribute('minlat');
+        const north = bound.getAttribute('maxlat');
+        return new Extent('EPSG:4326', west, east, south, north);
+    }
+    return new Extent('EPSG:4326', -180, 180, -90, 90);
+}
+
+function createTextureFromVector(tile, layer) {
+    if (tile.material === null) {
```
autra: `if (tile.material)`? (could be undefined maybe)

src/Core/Scheduler/Providers/VectorToColor_Provider.js
```diff
@@ -0,0 +1,129 @@
...
+
+        if (!(layer.extent instanceof Extent)) {
+            layer.extent = new Extent(layer.projection, layer.extent);
+        }
+
+        if (!layer.options.zoom) {
+            layer.options.zoom = { min: 5, max: 21 };
+        }
+
+        layer.format = layer.options.mimetype || 'vector/kml';
```
autra: Nits: maybe we can remove the default value here. We would have thrown an exception if `mimetype` is empty anyway, and I find this a bit confusing.

src/Core/Scheduler/Providers/VectorToColor_Provider.js
autra: `if (tile.material)` (could be undefined maybe?)

examples/globe_vector.js
```diff
@@ -0,0 +1,67 @@
...
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+// Add two elevation layers.
+// These will deform iTowns globe geometry to represent terrain elevation.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+
+promises.push(itowns.Fetcher.xml(urlKml).then(function _(file) {
+    var kmlLayer = {
+        type: 'color',
+        file: file,
```
autra: What about having a `url` field instead of `file` and fetch it in `preprocessDataLayer`? As globeView.addLayer returns the `whenReady` promise, this would be changed into

```js
promises.push(globeView.addlayer(kmlLayer));
```

Thinking out loud (not for this PR), but we could do the same for color and elevation layer: in layer preprocessing, if layer.capabilities is not set, query `getCapabilities`. If it's set and a string, threat it as a url, if it's set and an object, use it as capabilities description.

src/Renderer/ThreeExtended/GeoJSON2Feature.js
```diff
@@ -0,0 +1,237 @@
...
+                const codeStart = json.crs.properties.name.indexOf(':', epsgIdx + 5);
+                if (codeStart > 0) {
+                    return `EPSG:${json.crs.properties.name.substr(codeStart + 1)}`;
+                }
+            }
+        }
+        throw new Error(`Unsupported CRS type '${json.crs}'`);
+    }
+    // assume default crs
+    return 'EPSG:4236';
```
autra: typo from old code, I guess it should be `EPSG:4326`.

src/Renderer/ThreeExtended/GeoJSON2Feature.js
```diff
@@ -0,0 +1,237 @@
...
+ */
+import Coordinates from '../../Core/Geographic/Coordinates';
+import Extent from '../../Core/Geographic/Extent';
+
+function readCRS(json) {
+    if (json.crs) {
+        if (json.crs.type.toLowerCase() == 'epsg') {
+            return `EPSG:${json.crs.properties.code}`;
+        } else if (json.crs.type.toLowerCase() == 'name') {
+            const epsgIdx = json.crs.properties.name.toLowerCase().indexOf('epsg:');
```
autra, line 14: I'm pretty sure this parsing is wrong, seeing https://geojson.org/geojson-spec.html#named-crs

Anyway, the crs part is not even in the spec proposal any more. So I wouldn't even bother and I would remove `readCRS` entirely and use option.crsIn every time. I know this is old code, so it could be done here or in another PR, as you wish.

src/Renderer/ThreeExtended/GeoJSON2Feature.js
autra: But yeah, easiest would be to remove this function imho.

src/Renderer/ThreeExtended/GeoJSON2Feature.js
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Helper struct that returns an object { type: "", coordinates: [...], extent}:
+// - type is the geom type
+// - Coordinates is an array of Coordinate
+// - extent is optional, it's coordinates's extent
+// Multi-* geometry types are merged in one.
+const GeometryToCoordinates = {
+    point(crsIn, crsOut, coordsIn, filteringExtent, options) {
+        const extent = options.buildExtent ? new Extent(crsOut, Infinity, -Infinity, Infinity, -Infinity) : undefined;
```
autra, line 55: Don't you want to initialize your extent with 0, 0, 0, 0 instead, as it will be expanded by `readCoordinates`?

src/Renderer/ThreeExtended/GeoJSON2Feature.js
```diff
@@ -0,0 +1,237 @@
...
+        let coordinates = readCoordinates(crsIn, crsOut, coordsIn, extent);
+        if (filteringExtent) {
+            coordinates = coordinates.filter(c => filteringExtent.isPointInside(c));
+        }
+        return { type: 'point', coordinates, extent };
+    },
+    polygon(crsIn, crsOut, coordsIn, filteringExtent, options) {
+        const extent = options.buildExtent ? new Extent(crsOut, Infinity, -Infinity, Infinity, -Infinity) : undefined;
+        const coordinates = readCoordinates(crsIn, crsOut, coordsIn, extent);
+        if (filteringExtent && !filteringExtent.isPointInside(coordinates[0])) {
```
autra, line 65: I'm confused here. Why this line isn't the same as line 58?

src/Core/Scheduler/Providers/VectorToColor_Provider.js
```diff
@@ -0,0 +1,129 @@
...
+        return tile.level >= layer.options.zoom.min && tile.level <= layer.options.zoom.max && layer.extent.intersect(tile.extent);
+    },
+    executeCommand(command) {
+        const tile = command.requester;
+
+        const layer = command.layer;
+        const supportedFormats = {
+            'vector/kml': createTextureFromVector,
+            'vector/gpx': createTextureFromVector,
+            'vector/geojson': createTextureFromVector,
```
autra: if we are always calling the same method, `supportedFormats` can be an array of format, and code becomes

```js
if (!supportedFormat.includes(layer.format) {
    return Promise.reject(...); 
}

return createTextureFromVector(...);
```

autra (changes requested):
Taking this over, because @peppsac is off. Mostly good, I think we are nearly there. I particularly like the examples :-)

src/Core/Geographic/Coordinates.js
gchoqueux: thanks i forgot to remove this unused function

src/Core/Geographic/Extent.js
gchoqueux: thanks for suggestion

src/Core/Prefab/GlobeView.js
gchoqueux: `layer.projection` is the file's projection, This is the projection of the source as the wmts, wms...
`layer.reprojection` is texture tile projection.

In globe's `referenceCrs` is cartesian geocentric system and texture projection is planar. 
In planar mode `referenceCrs` and texture projection, both are planar.

The layer must know the original projection and the destination projection, to rasterize vector data to texture

src/Core/Scheduler/Providers/VectorToColor_Provider.js
gchoqueux: maybe `RasterProvider` is better

examples/planar_vector.js
gchoqueux: ok

src/Core/Geographic/Coordinates.js
gchoqueux: removed

src/Core/Prefab/PlanarView.js
gchoqueux: see https://github.com/iTowns/itowns/pull/450#discussion_r136289546

examples/globe_vector.js
gchoqueux:
```js
promises.push(globeView.addlayer(kmlLayer));
```
--> Done

src/Renderer/ThreeExtended/GeoJSON2Feature.js
gchoqueux: in another PR

src/Renderer/ThreeExtended/GeoJSON2Feature.js
gchoqueux: done I change  in `EPSG:4326`

src/Renderer/ThreeExtended/GeoJSON2Feature.js
gchoqueux: if we extent with coordinates  (1,1) the extent becomes (0, 1, 0, 1) and it's wrong

src/Renderer/ThreeExtended/GeoJSON2Feature.js
gchoqueux: This filtering is added and used  WFS PR to remove somes polygons and lines that aren't tile's extent

gchoqueux:
@autra updated

src/Core/Prefab/GlobeView.js
autra: Ok, I forgot that referenceCRS is actually the CRS of the THREE coordinates, used to place the camera for instance. Thanks for the explanations!

src/Core/Scheduler/Providers/VectorToColor_Provider.js
autra: Yes `RasterProvider` is good.

src/Renderer/ThreeExtended/GeoJSON2Feature.js
autra: Oh all right, I misread `[-Infinity, Infinity, -Infinity, Infinity]`

gchoqueux:
@autra branch updated

autra (approved):
LGTM, thanks!

PR #451, opened 2017-08-11
feature: add StaticProvider

peppsac:
Sometimes it's handy to be able to use a single image as a layer (mainly want you
want to quick test without the hassle of setting up a wms server for instance).

This provider is really simple, and just compute the offset in the image to use
to match the tile extent.

Example usage:
```js
    view.addLayer({
        url: 'http://localhost:8080/foo.png',
        type: 'color',
        protocol: 'image',
        id: 'myimage',
    });
```

### DIFF
--- a/src/Core/Scheduler/Providers/StaticProvider.js
+++ b/src/Core/Scheduler/Providers/StaticProvider.js
@@ -1,0 +1,132 @@
+import { Vector3 } from 'three';
+import Extent from '../../Geographic/Extent';
+import OGCWebServiceHelper from './OGCWebServiceHelper';
+import Fetcher from './Fetcher';
+import { l_COLOR } from '../../../Renderer/LayeredMaterial';
+
+// select the smallest image entirely covering the tile
+function selectBestImageForExtent(images, extent) {
+    let selection;
+    for (const entry of images) {
+        if (extent.isInside(entry.extent)) {
+            if (!selection) {
+                selection = entry;
+            } else {
+                const d = selection.extent.dimensions();
+                const e = entry.extent.dimensions();
+                if (e.x <= d.x && e.y <= d.y) {
+                    selection = entry;
+                }
+            }
+        }
+    }
+    return selection;
+}
+
+function getColorTexture(tile, layer) {
+    if (!layer.tileInsideLimit(tile, layer)) {
+        return Promise.reject(`Tile '${tile}' is outside layer bbox ${layer.extent}`);
+    }
+    if (!tile.material) {
+        return Promise.resolve();
+    }
+
+    if (!layer.images) {
+        return Promise.reject();
+    }
+
+    const selection = selectBestImageForExtent(layer.images, tile.extent);
+
+    if (!selection) {
+        return Promise.reject(
+            new Error(`No available image for tile ${tile}`));
+    }
+
+    const url = layer.url.href.substr(0, layer.url.href.lastIndexOf('/') + 1) + selection.image;
+    return OGCWebServiceHelper.getColorTextureByUrl(url, layer.networkOptions).then((texture) => {
+        // adjust pitch
+        const result = {
+            texture,
+            pitch: new Vector3(0, 0, 1),
+        };
+
+        result.texture.extent = selection.extent;
+        result.texture.coords = selection.extent;
+        if (!result.texture.coords.zoom || result.texture.coords.zoom > tile.level) {
+            result.texture.coords.zoom = tile.level;
+            result.texture.file = selection.image;
+        }
+        // TODO: modify TileFS to handle tiles with ratio != image's ratio
+        result.pitch = tile.extent.offsetToParent(selection.extent);
+
+
+        return result;
+    });
+}
+
+
+/**
+ * This provider uses no protocol but instead download static images directly.
+ *
+ * It uses as input 'image_filename: extent' values and then tries to find the best image
+ * for a given tile using the extent property.
+ */
+export default {
+    preprocessDataLayer(layer) {
+        if (!layer.extent) {
+            throw new Error('layer.extent is required');
+        }
+
+        if (!(layer.extent instanceof Extent)) {
+            layer.extent = new Extent(layer.projection, ...layer.extent);
+        }
+
+        layer.options = {};
+        layer.canTileTextureBeImproved = this.canTileTextureBeImproved;
+        layer.url = new URL(layer.url, window.location);
+        return Fetcher.json(layer.url.href).then((metadata) => {
+            layer.images = [];
+            // eslint-disable-next-line guard-for-in
+            for (const image in metadata) {
+                const extent = new Extent(layer.projection, ...metadata[image]);
+                layer.images.push({
+                    image,
+                    extent,
+                });
+            }
+        });
+    },
+
+    tileInsideLimit(tile, layer) {
+        if (!layer.images) {
+            return false;
+        }
+        for (const entry of layer.images) {
+            if (tile.extent.isInside(entry.extent)) {
+                return true;
+            }
+        }
+
+        return false;
+    },
+
+    canTileTextureBeImproved(layer, tile) {
+        const s = selectBestImageForExtent(layer.images, tile.extent);
+        if (!s) {
+            return false;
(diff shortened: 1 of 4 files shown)


src/Core/Scheduler/Providers/SingleImageProvider.js
```diff
@@ -0,0 +1,52 @@
...
+
+    tileInsideLimit(tile, layer) {
+        // Only fetch level 0 (full res texture): node with level > 0 will inherit the
+        // texture from their parent
+        return tile.level === 0 && layer.extent.intersect(tile.extent);
+    },
+
+    getColorTexture(tile, layer) {
+        if (!this.tileInsideLimit(tile, layer)) {
+            return Promise.reject(`Tile '${tile}' is outside layer bbox ${layer.extent}`);
```
autra: This will produce something like 

> Tile '[object Object]' is outside layer bbox '[object Object]'

Also, usually, people reject with an Error object, not a single string, which could allow you to attach tile and layer to this error object maybe? Not sure what is best here.

autra (commented):
Is it supposed to work with the globe? Is the following result expected?

selection_038

peppsac:
I'm going to refactor this one to support another use case: instead of using a single image, the user could give this provider a list of images and their respective extent.
This would make it easier to use itowns without the hassle of setting up an imagery server.

peppsac:
PR updated (depends on PR #485 )

I've refactored the code a bit and now this provider takes a list of image with their respective extent.

Note: this provider only makes sense for planar tiles.

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+function selectBestImageForExtent(images, extent) {
+    let selection;
+    for (const entry of images) {
+        if (extent.isInside(entry.extent)) {
+            if (!selection) {
+                selection = entry;
+            } else {
+                const d = selection.extent.dimensions();
+                const e = entry.extent.dimensions();
+                if (e.x <= d.x && e.y <= d.y) {
```
autra, line 17: Is it possible to have a e.x > d.x but e.y < d.y? In this case, maybe we should choose the smallest one.

And btw, shouldn't we discriminate on the resulting px per meter instead? That's how I'd define "best" intuitively...

autra: For my second point, I realised you can't do that before downloading the image (at least not without additional metadata) so we can forget it.

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+        const mat = tile.material;
+        const idx = mat.getLayerTextureOffset(layer.id);
+        const currentTexture = mat.textures[1][idx];
+        if (!currentTexture.file) {
+            return true;
+        }
+        return currentTexture.file != s.image;
+    },
+
+    getColorTexture(tile, layer) {
```
autra: I didn't see where you're using it elsewhere. If you're not, let's not export this.

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+        if (!(layer.extent instanceof Extent)) {
+            layer.extent = new Extent(layer.projection, ...layer.extent);
+        }
+
+        layer.options = {};
+        layer.canTileTextureBeImproved = this.canTileTextureBeImproved;
+        layer.url = new URL(layer.url, window.location);
+        return Fetcher.json(layer.url.href).then((metadata) => {
+            layer.images = [];
+            // eslint-disable-next-line guard-for-in
```
autra, line 89: eslint complains because you didn't check `if (metadata.hasOwnProperty(image))` inside the for-loop.

Note that you *might* need to use `Object.prototype.hasOwnProperty.call(metadata, image)` instead if it does not work, see https://eslint.org/docs/rules/guard-for-in

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+            return true;
+        }
+        return currentTexture.file != s.image;
+    },
+
+    getColorTexture(tile, layer) {
+        if (!this.tileInsideLimit(tile, layer)) {
+            return Promise.reject(`Tile '${tile}' is outside layer bbox ${layer.extent}`);
+        }
+        if (tile.material === null) {
```
autra: maybe we can use `==` to also cover undefined? Or alternatively, as `tile.material` won't ever be 0, just `if (tile.material)`?

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        if (!layer.images) {
+            return Promise.reject();
+        }
+
+        const selection = selectBestImageForExtent(layer.images, tile.extent);
+
+        if (!selection) {
+            return Promise.reject('pas dimage');
```
autra: Let's reject with a `Error` object (and let's translate that ;-) )

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+    },
+
+    canTileTextureBeImproved(layer, tile) {
+        const s = selectBestImageForExtent(layer.images, tile.extent);
+        if (!s) {
+            return false;
+        }
+        const mat = tile.material;
+        const idx = mat.getLayerTextureOffset(layer.id);
+        const currentTexture = mat.textures[1][idx];
```
autra: `mat.textures[l_COLOR][idx]`

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -0,0 +1,130 @@
...
+        if (!currentTexture.file) {
```
autra: Does this mean we have done the request but haven't received the response yet?

autra: Actually, I'm not sure this situation is possible. Can you elaborate?

autra (changes requested):
Nice! This will be useful indeed.

Can you provide a simple example with this PR?

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: The help page isn't helpful: `This rule is aimed at preventing unexpected behavior that could arise from using a for in loop` : Sure, but what "unexpected behavior" could possibly happen? (especially in this case where metadata is a json file)?

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Textures are always not null in iTowns (see https://github.com/iTowns/itowns/blob/master/src/Renderer/LayeredMaterial.js#L133).
So in this provider I'm storing the image name  in an additional property (`file`) to be able to compare this filename to the best choice available.
It's the most naive way to implement this but it works.

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Correct. I'll fix this.

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Yep, thanks.

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Oops 👍

src/Core/Scheduler/Providers/StaticProvider.js
autra: Yes, in this particular case, we don't really risk anything. It's just that I prefer to add a guard than to disable the linter. But I should really have prefixed this with `nits`, it's not very important.

peppsac:
PR updated.
I don't have any data to build a public example yet, but I'll open a PR soon to add an example.

autra (approved):

PR #453, opened 2017-08-22
fix: fix a bug with the points, they don't have the z-buffer logarith…

NikoSaul:
fix: fix a bug with the points, they don't have the z-buffer logarithmic active


peppsac (approved):

chrislcs:
Can confirm the fix is working, thanks

PR #454, opened 2017-08-25
Feature : added new PlanarControls

EmmanuelSchmuck:
PlanarControls provides ergonomic camera controls for the planar mode : drag, pan, orbit,
zoom and animated expression to smoothly move and orient the camera.

Modified planar example to use these controls
Update contributors

### DIFF
--- a/src/Renderer/ThreeExtended/PlanarControls.js
+++ b/src/Renderer/ThreeExtended/PlanarControls.js
@@ -4,10 +4,9 @@
 * Ctrl + left mouse : rotate (orbit) around the camera's focus point.
 * Scroll wheel : zooms toward cursor position (animated).
 * Middle mouse button (wheel click) : 'smart zoom' at cursor location (animated).
-* S : go to start view (animated)
+* Y : go to start view (animated)
 * T : go to top view (animated)
 * How to use : instanciate PlanarControls after camera setup (setPosition and lookAt)
-* or you can also setup the camera with options.startPosition and options.startLook
 */
 
 import * as THREE from 'three';
@@ -16,8 +15,8 @@ import * as THREE from 'three';
 const keys = {
     CTRL: 17,
     SPACE: 32,
-    S: 83,
     T: 84,
+    Y: 89,
 };
 
 const mouseButtons = {
@@ -84,6 +83,12 @@ function PlanarControls(view, options = {}) {
 
     // should be less than 90 deg (90 = parallel to the ground)
     this.maxZenithAngle = (options.maxZenithAngle || 82.5) * Math.PI / 180;
+
+    // starting camera position and orientation target are setup before instanciating PlanarControls
+    // using: view.camera.setPosition() and view.camera.lookAt()
+    // startPosition and startQuaternion are stored to be able to return to the start view
+    this.startPosition = this.camera.position.clone();
+    this.startQuaternion = this.camera.quaternion.clone();
 
     // prevent the default contextmenu from appearing when right-clicking
     // this allows to use right-click for input without the menu appearing
@@ -490,7 +495,7 @@ function PlanarControls(view, options = {}) {
     * Triggers an animated movement (travel) to set the camera to starting view
     */
     this.goToStartView = function goToStartView() {
-        this.initiateTravel(this.startPosition, 'auto', this.startLook, true);
+        this.initiateTravel(this.startPosition, 'auto', this.startQuaternion, true);
     };
 
     /**
@@ -585,14 +590,6 @@ function PlanarControls(view, options = {}) {
     PlanarControls.prototype = Object.create(THREE.EventDispatcher.prototype);
     PlanarControls.prototype.constructor = PlanarControls;
 
-    // starting position and lookAt target can be set outside this class, before instanciating PlanarControls
-    // or they can be set with options : startPosition and startLookAt
-    this.startPosition = options.startPosition || this.camera.position.clone();
-    this.startLook = options.startLook || this.camera.quaternion.clone();
-
-    this.camera.position.copy(this.startPosition);
-    this.camera.quaternion.copy(this.startLook);
-
     // event listeners for user input
     this.addInputListeners();
 }
@@ -628,8 +625,7 @@ var onMouseDown = function onMouseDown(event) {
 };
 
 /**
-* Catch the event when a touch on the mouse is uped. Reinit the state of the controller and disable.
-* the listener on the move mouse event.
+* Catch the event when a touch on the mouse is uped.
 * @param {event} event : the current event
 */
 var onMouseUp = function onMouseUp(event) {
@@ -693,7 +689,7 @@ var onKeyDown = function onKeyDown(event) {
     if (event.keyCode === keys.T) {
         this.goToTopView();
     }
-    if (event.keyCode === keys.S) {
+    if (event.keyCode === keys.Y) {
         this.goToStartView();
     }
     if (event.keyCode === keys.SPACE) {
@@ -725,7 +721,6 @@ var onMouseWheel = function onMouseWheel(event) {
 var onContextMenu = function onContextMenu(event) {
     event.preventDefault();
 };
-
 
 /**
 * smoothing function (sigmoid) : based on h01 Hermite function
--- a/src/Renderer/ThreeExtended/PlanarControls.js
+++ b/src/Renderer/ThreeExtended/PlanarControls.js
@@ -1,0 +1,746 @@
+/** Description: Camera controls adapted for a planar view, with animated movements
+* Left mouse button : "drag" the ground, translating the camera on the (xy) world plane.
+* Right mouse button : translate the camera on local x and world z axis (pan)
+* Ctrl + left mouse : rotate (orbit) around the camera's focus point.
+* Scroll wheel : zooms toward cursor position (animated).
+* Middle mouse button (wheel click) : 'smart zoom' at cursor location (animated).
+* S : go to start view (animated)
+* T : go to top view (animated)
+* How to use : instanciate PlanarControls after camera setup (setPosition and lookAt)
+* or you can also setup the camera with options.startPosition and options.startLook
+*/
+
+import * as THREE from 'three';
+
+// event keycode
+const keys = {
+    CTRL: 17,
+    SPACE: 32,
+    S: 83,
+    T: 84 };
+
+const mouseButtons = {
+    LEFTCLICK: THREE.MOUSE.LEFT,
+    MIDDLECLICK: THREE.MOUSE.MIDDLE,
+    RIGHTCLICK: THREE.MOUSE.RIGHT,
+};
+
+// control state
+const STATE = {
+    NONE: -1,
+    DRAG: 0,
+    PAN: 1,
+    ROTATE: 2,
+    TRAVEL: 3,
+};
+
+/**
+* PlanarControls Constructor
+* Numerical values have been adjusted for the example provided in examples/planar.html
+* Most of them can be changed with the options parameter
+* @param {PlanarView} view : the itowns view (planar view)
+* @param {options} options : optional parameters.
+*/
+function PlanarControls(view, options = {}) {
+    this.view = view;
+    this.camera = view.camera.camera3D;
+    this.domElement = view.mainLoop.gfxEngine.renderer.domElement;
+    this.position = this.camera.position;
+
+    this.rotateSpeed = options.rotateSpeed || 2.0;
+
+    // minPanSpeed when close to the ground, maxPanSpeed when close to maxAltitude
+    this.maxPanSpeed = options.maxPanSpeed || 10;
+    this.minPanSpeed = options.minPanSpeed || 0.05;
+
+    // animation duration for the zoom
+    this.zoomTravelTime = options.zoomTravelTime || 0.2;
+
+    // zoom movement is equal to the distance to the zoom target, multiplied by zoomFactor
+    this.zoomInFactor = options.zoomInFactor || 0.25;
+    this.zoomOutFactor = options.zoomOutFactor || 0.4;
+
+    // pan movement is clamped between maxAltitude and groundLevel
+    this.maxAltitude = options.maxAltitude || 12000;
+
+    // approximate ground altitude value
+    this.groundLevel = options.groundLevel || 200;
+
+    // min and max duration in seconds, for animated travels with 'auto' parameter
+    this.autoTravelTimeMin = options.autoTravelTimeMin || 1.5;
+    this.autoTravelTimeMax = options.autoTravelTimeMax || 4;
+
+    // max travel duration is reached for this travel distance
+    this.autoTravelTimeDist = options.autoTravelTimeDist || 20000;
+
+    // after a smartZoom, camera height above ground will be between these two values
+    this.smartZoomHeightMin = options.smartZoomHeightMin || 75;
+    this.smartZoomHeightMax = options.smartZoomHeightMax || 500;
+
+    // if set to true, animated travels have 0 duration
+    this.instantTravel = options.instantTravel || false;
+
+    this.minZenithAngle = options.minZenithAngle || 0 * Math.PI / 180;
+
+    // should be less than 90 deg (90 = parallel to the ground)
+    this.maxZenithAngle = (options.maxZenithAngle || 82.5) * Math.PI / 180;
+
+    // prevent the default contextmenu from appearing when right-clicking
+    // this allows to use right-click for input without the menu appearing
+    this.domElement.addEventListener('contextmenu', onContextMenu.bind(this), false);
+
+    // add this PlanarControl instance to the view's framerequesters
+    // with this, PlanarControl.update() will be called each frame
+    this.view.addFrameRequester(this);
+
+    this.state = STATE.NONE;
+    this.isCtrlDown = false;
+
+    // mouse movement
+    this.mousePosition = new THREE.Vector2();
+    this.lastMousePosition = new THREE.Vector2();
+    this.deltaMousePosition = new THREE.Vector2(0, 0);
+
+    // drag movement
+    this.dragStart = new THREE.Vector3();
+    this.dragEnd = new THREE.Vector3();
+    this.dragDelta = new THREE.Vector3();
+
+    // camera focus point : ground point at screen center
+    this.centerPoint = new THREE.Vector3(0, 0, 0);
+
+    // camera rotation
+    this.phi = 0.0;
+
+    // animated travel
+    this.travelEndPos = new THREE.Vector3();
+    this.travelStartPos = new THREE.Vector3();
+    this.travelStartRot = new THREE.Quaternion();
+    this.travelEndRot = new THREE.Quaternion();
+    this.travelAlpha = 0;
+    this.travelDuration = 0;
+    this.travelUseRotation = false;
+    this.travelUseSmooth = false;
+
+    // time management
+    this.deltaTime = 0;
+    this.lastElapsedTime = 0;
+    this.clock = new THREE.Clock();
+
+    // eventListeners handlers
+    this._handlerOnKeyDown = onKeyDown.bind(this);
+    this._handlerOnKeyUp = onKeyUp.bind(this);
+    this._handlerOnMouseDown = onMouseDown.bind(this);
+    this._handlerOnMouseUp = onMouseUp.bind(this);
+    this._handlerOnMouseMove = onMouseMove.bind(this);
+    this._handlerOnMouseWheel = onMouseWheel.bind(this);
+
+    /**
+    * PlanarControl update
+    * Updates the view and camera if needed, and handles the animated travel
+    */
+    this.update = function update() {
+        this.deltaTime = this.clock.getElapsedTime() - this.lastElapsedTime;
+        this.lastElapsedTime = this.clock.getElapsedTime();
+
+        if (this.state === STATE.TRAVEL) {
+            this.handleTravel(this.deltaTime);
+        }
+        if (this.state !== STATE.NONE) {
+            this.view.camera.update(window.innerWidth, window.innerHeight);
+            this.view.notifyChange(true);
+        }
+    };
+
+    /**
+    * Initiate a drag movement (translation on xy plane) when user does a left-click
+    * The movement value is derived from the actual world point under the mouse cursor
+    * This allows the user to 'grab' a world point and drag it to move (eg : google map)
+    */
+    this.initiateDrag = function initiateDrag() {
+        this.state = STATE.DRAG;
+
+        // the world point under mouse cursor when the drag movement is started
+        this.dragStart.copy(this.getWorldPointAtScreenXY(this.mousePosition));
+
+        // the difference between start and end cursor position
+        this.dragDelta.set(0, 0, 0);
+    };
+
+    /**
+    * Handle the drag movement (translation on xy plane) when user moves the mouse while in STATE.DRAG
+    * The drag movement is previously initiated when user does a left-click, by initiateDrag()
+    * Compute the drag value and update the camera controls.
+    * The movement value is derived from the actual world point under the mouse cursor
+    * This allows the user to 'grab' a world point and drag it to move (eg : google map)
+    */
+    this.handleDragMovement = function handleDragMovement() {
+        // the world point under the current mouse cursor position, at same altitude than dragStart
+        this.dragEnd.copy(this.getWorldPointFromMathPlaneAtScreenXY(this.mousePosition, this.dragStart.z));
+
+        // the difference between start and end cursor position
+        this.dragDelta.subVectors(this.dragStart, this.dragEnd);
+
+        // new camera position
+        this.position.add(this.dragDelta);
+
+        // request update
+        this.update();
+    };
+
+    /**
+    * Initiate a pan movement (local translation on xz plane) when user does a righ-click
+    */
+    this.initiatePan = function initiatePan() {
+        this.state = STATE.PAN;
+    };
+
+    /**
+    * Handle the pan movement (translation on local x / world z plane) when user moves the mouse while in STATE.PAN
+    * The drag movement is previously initiated when user does a right-click, by initiatePan()
+    * Compute the pan value and update the camera controls.
+    */
+    this.handlePanMovement = function handlePanMovement() {
+        // normalized (betwwen 0 and 1) distance between groundLevel and maxAltitude
+        const distToGround = THREE.Math.clamp((this.position.z - this.groundLevel) / this.maxAltitude, 0, 1);
+
+        // pan movement speed, adujsted according to altitude
+        const panSpeed = THREE.Math.lerp(this.minPanSpeed, this.maxPanSpeed, distToGround);
+
+        // lateral movement (local x axis)
+        this.position.copy(this.camera.localToWorld(new THREE.Vector3(panSpeed * -1 * this.deltaMousePosition.x, 0, 0)));
+
+        // vertical movement (world z axis)
+        const newAltitude = this.position.z + panSpeed * this.deltaMousePosition.y;
+
+        // check if altitude is valid
+        if (newAltitude < this.maxAltitude && newAltitude > this.groundLevel) {
+            this.position.z = newAltitude;
+        }
+
+        // request update
+        this.update();
+    };
+
+    /**
+    * Initiate a rotate (orbit) movement when user does a right-click or ctrl + left-click
+    */
+    this.initiateRotation = function initiateRotation() {
+        this.state = STATE.ROTATE;
+
+        const screenCenter = new THREE.Vector2(0.5 * window.innerWidth, 0.5 * window.innerHeight);
+
+        this.centerPoint.copy(this.getWorldPointAtScreenXY(screenCenter));
+
+        const r = this.position.distanceTo(this.centerPoint);
+        this.phi = Math.acos((this.position.z - this.centerPoint.z) / r);
+    };
+
+    /**
+    * Handle the rotate movement (orbit) when user moves the mouse while in STATE.ROTATE
+    * the movement is an orbit around 'centerPoint', the camera focus point (ground point at screen center)
+    * The rotate movement is previously initiated in initiateRotation()
+    * Compute the new position value and update the camera controls.
+    */
+    this.handleRotation = function handleRotation() {
+        // angle deltas
+        // deltaMousePosition is computed in onMouseMove / onMouseDown s
+        const thetaDelta = -this.rotateSpeed * this.deltaMousePosition.x / window.innerWidth;
+        const phiDelta = -this.rotateSpeed * this.deltaMousePosition.y / window.innerHeight;
+
+        // the vector from centerPoint (focus point) to camera position
+        const offset = this.position.clone().sub(this.centerPoint);
+
+        const quat = new THREE.Quaternion().setFromUnitVectors(this.camera.up, new THREE.Vector3(0, 0, 1));
+        const quatInverse = quat.clone().inverse();
+
+        if (thetaDelta !== 0 || phiDelta !== 0) {
+            if ((this.phi + phiDelta >= this.minZenithAngle)
+            && (this.phi + phiDelta <= this.maxZenithAngle)
+            && phiDelta !== 0) {
+                // rotation around X (altitude)
+                this.phi += phiDelta;
+                offset.applyQuaternion(quat);
+
+                const rotationXQuaternion = new THREE.Quaternion();
+                const vector = new THREE.Vector3();
+
+                vector.setFromMatrixColumn(this.camera.matrix, 0);
+                rotationXQuaternion.setFromAxisAngle(vector, phiDelta);
+                offset.applyQuaternion(rotationXQuaternion);
+                offset.applyQuaternion(quatInverse);
+            }
+            if (thetaDelta !== 0) {
+                // rotation around Z (azimuth)
+
+                const rotationZQuaternion = new THREE.Quaternion();
+                rotationZQuaternion.setFromAxisAngle(new THREE.Vector3(0, 0, 1), thetaDelta);
+                offset.applyQuaternion(rotationZQuaternion);
+            }
+        }
+
+        this.position.copy(offset).add(this.centerPoint);
+
+        this.camera.lookAt(this.centerPoint);
+
+        this.update();
+    };
+
+    /**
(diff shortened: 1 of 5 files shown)


src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,762 @@
...
+const travelEndRot = new THREE.Quaternion();
+let travelAlpha = 0;
+let travelDuration = 0;
+let travelUseRotation = false;
+let travelUseSmooth = false;
+
+// time management
+let deltaTime = 0;
+let lastElapsedTime = 0;
+const clock = new THREE.Clock();
```
peppsac: How about storing all these inside the `PlanarControls` instance so a user can instanciate several controls?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,762 @@
...
+        state = STATE.DRAG;
+
+        // the world point under mouse cursor when the drag movement is started
+        dragStart.copy(this.getWorldPointAtScreenXY(mousePosition));
+
+        // the difference between start and end cursor position
+        dragDelta.set(0, 0, 0);
+    };
+
+    /** =========================================================================================================================
```
peppsac: I don't mind the `=========` separator but this is something that should be done everywhere or nowhere (as all other stylistic choices).
So I'd suggest removing them from this PR and open a specific PR to convert itowns to this style (but I'm not sure it would be merged...)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,762 @@
...
+    // p between 1.0 and 1.5
+    return Math.pow((value * value * (3 - 2 * value)), 1.20);
+};
+
+/** =========================================================================================================================
+* clamp value between 0 and 1
+* @param {number} value : the value to be smoothed, between 0 and 1
+* @returns {number}
+*/
+var clamp01 = function clamp01(value) {
```
peppsac: You should directly use `THREE.Math.clamp(value, 0, 1)` instead of adding a one liner function.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,762 @@
...
+* T : go to top view (animated)
+* How to use : instanciate PlanarControls after camera setup (setPosition and lookAt)
+* or you can also setup the camera with options.startPosition and options.startLookAt
+*/
+
+import * as THREE from 'three';
+
+// event keycode
+const keys = { CTRL: 17, SPACE: 32, S: 83, T: 84 };
+const mouseButtons = { LEFTCLICK: THREE.MOUSE.LEFT, MIDDLECLICK: THREE.MOUSE.MIDDLE, RIGHTCLICK: THREE.MOUSE.RIGHT };
```
peppsac: Minor style comment:
```js
 const mouseButtons = {
    LEFTCLICK: THREE.MOUSE.LEFT,
    MIDDLECLICK: THREE.MOUSE.MIDDLE,
    RIGHTCLICK: THREE.MOUSE.RIGHT,
};
```
is prefered.

peppsac (commented):
Thanks for the PR and welcome to iTowns project :)

I did a very quick first review.
I (or another reviewer) will have do a more in depth one before merging but the code is looking good mostly and is well documented, nice!

EmmanuelSchmuck:
Thank you for the review @peppsac , I made the modifications you suggested.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,746 @@
...
+* Numerical values have been adjusted for the example provided in examples/planar.html
+* Most of them can be changed with the options parameter
+* @param {PlanarView} view : the itowns view (planar view)
+* @param {options} options : optional parameters.
+*/
+function PlanarControls(view, options = {}) {
+    this.view = view;
+    this.camera = view.camera.camera3D;
+    this.domElement = view.mainLoop.gfxEngine.renderer.domElement;
+    this.position = this.camera.position;
```
peppsac: `this.position` isn't really useful: let's use `this.camera.position` directly instead.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,746 @@
...
+    this._handlerOnMouseUp = onMouseUp.bind(this);
+    this._handlerOnMouseMove = onMouseMove.bind(this);
+    this._handlerOnMouseWheel = onMouseWheel.bind(this);
+
+    /**
+    * PlanarControl update
+    * Updates the view and camera if needed, and handles the animated travel
+    */
+    this.update = function update() {
+        this.deltaTime = this.clock.getElapsedTime() - this.lastElapsedTime;
```
peppsac: Why don't you use the `dt` provided as an argument to your update function (see https://github.com/iTowns/itowns/blob/master/src/Core/MainLoop.js#L63)?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,746 @@
...
+        this.lastElapsedTime = this.clock.getElapsedTime();
+
+        if (this.state === STATE.TRAVEL) {
+            this.handleTravel(this.deltaTime);
+        }
+        if (this.state !== STATE.NONE) {
+            this.view.camera.update(window.innerWidth, window.innerHeight);
```
peppsac: You don't need to call this: it's done for you in view's update process (https://github.com/iTowns/itowns/blob/master/src/Core/MainLoop.js#L96)

peppsac (commented):
I tested this controls quickly and it seems to be working good.

I left a few comments and then it'll be ready for merging.

EmmanuelSchmuck:
@peppsac : modified again following your suggestions. Also, my inputs no longer call update() but notifyChange() instead, much cleaner.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    this.travelEndRot = new THREE.Quaternion();
+    this.travelAlpha = 0;
+    this.travelDuration = 0;
+    this.travelUseRotation = false;
+    this.travelUseSmooth = false;
+
+    // time management
+    this.deltaTime = 0;
+    this.lastElapsedTime = 0;
+    this.clock = new THREE.Clock();
```
autra: `this.clock` is not used.

autra: Actually, all three "time management" vars are not used, if I'm not mistaken.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+
+    // eventListeners handlers
+    this._handlerOnKeyDown = onKeyDown.bind(this);
+    this._handlerOnKeyUp = onKeyUp.bind(this);
+    this._handlerOnMouseDown = onMouseDown.bind(this);
+    this._handlerOnMouseUp = onMouseUp.bind(this);
+    this._handlerOnMouseMove = onMouseMove.bind(this);
+    this._handlerOnMouseWheel = onMouseWheel.bind(this);
```
autra: As everything is defined in the constructor, you don't really need to attach these handlers on `this`. They can be local variables. That's better because they are truly isolated then (and being able to really isolate things is a very good reason to do this way, and not to use `prototype` or ES6 classes :-) )

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+
+    /**
+    * PlanarControl update
+    * Updates the view and camera if needed, and handles the animated travel
+    * @param {number} dt : deltatime given by mainLoop
+    * @param {boolean} updateLoopRestarted : given by mainLoop
+    */
```
autra: Not sure we want to document this. It's just the implementation of a FrameRequester. It's not meant to be called by the outside world. Better not have it in documentation.

src/Renderer/ThreeExtended/PlanarControls.js
autra: This is also true for every local state you don't want to expose. I'm thinking about:
- this.view
- this.camera
- all `travelFoo` vars
- all `drag` vars
-  this.mousePosition, this.lastMousePosition and this.deltaMousePosition
- isCtrlDown

I'm pretty sure these should be hidden.

In contrast, I believe keeping all the option properties visible is perfect :-)

For `this.state`, I have no strong opinion about it.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    this.update = function update(dt, updateLoopRestarted) {
+        // dt will not be relevant when we just started rendering, we consider a 1-frame move in this case
+        if (updateLoopRestarted) {
+            dt = 16;
+        }
+        if (this.state === STATE.TRAVEL) {
+            this.handleTravel(dt / 1000);
```
autra: What about reasoning in ms everywhere?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+};
+
+/**
+* smoothing function (sigmoid) : based on h01 Hermite function
+* returns a value between 0 and 1
+* @param {number} value : the value to be smoothed, between 0 and 1
+* @returns {number}
+*/
+var smooth = function smooth(value) {
+    // p between 1.0 and 1.5
```
autra: p is 1.20?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    }
+    if (event.keyCode === keys.Y) {
+        this.goToStartView();
+    }
+    if (event.keyCode === keys.SPACE) {
+        this.initiateSmartZoom(event);
+    }
+    if (event.keyCode === keys.CTRL) {
+        this.isCtrlDown = true;
+    }
```
autra: nits: switch-case?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+};
+
+/**
+* Catch and manage the event when a key is up.
+* @param {event} event : the current event
+*/
+var onKeyUp = function onKeyUp(event) {
+    if (event.keyCode === keys.CTRL) {
+        this.isCtrlDown = false;
```
autra: I believe you don't need `isCtrlDown`, use event.ctrlKey instead (available for [MouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/ctrlKey) and [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/ctrlKey)).

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    this.deltaMousePosition.copy(this.mousePosition).sub(this.lastMousePosition);
+
+    this.lastMousePosition.copy(this.mousePosition);
+
+    // notify change if moving
+    if (this.state !== STATE.NONE) {
+        this.view.notifyChange(true);
+    }
+    if (this.state === STATE.ROTATE) {
+        this.handleRotation();
```
autra: So we call `handleRotation` at each mouse event. I have the feeling that doing all this work in `update` instead would yield the same result with better perf (actually that's one of the reason I wrote the FrameRequester logic : to give a painless mechanism to throttle input events). The logic is to set the state in event handlers, then deal with them (movements, animation etc... in `update`). 

I know `FlyControls` doesn't do things this way, but it should have done so imho (my bad). Or maybe I missed something?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    this.mousePosition.set(event.clientX, event.clientY);
+
+    if (this.state === STATE.TRAVEL) {
+        return;
+    }
+
+    this.lastMousePosition.copy(this.mousePosition);
+
+    if (event.button === mouseButtons.LEFTCLICK) {
+        if (this.isCtrlDown) {
```
autra: Use `event.ctrlKey` instead (see below).

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        }
+        else if (this.state === STATE.PAN) {
+            this.domElement.style.cursor = 'cell';
+        }
+        else if (this.state === STATE.TRAVEL) {
+            this.domElement.style.cursor = 'wait';
+        }
+        else if (this.state === STATE.ROTATE) {
+            this.domElement.style.cursor = 'move';
+        }
```
autra: switch-case block?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        this.domElement.addEventListener('mousewheel', this._handlerOnMouseWheel, false);
+        // For firefox
+        this.domElement.addEventListener('MozMousePixelScroll', this._handlerOnMouseWheel, false);
+    };
+
+    /**
+    * removes all the input event listeners (desactivate the controls)
+    */
+    this.removeInputListeners = function removeInputListeners() {
+        window.removeEventListener('keydown', this._handlerOnKeyDown, true);
```
autra: I agree this looks more practical in our examples, but please consider that one html page could have several itowns instance, or other js actions using keys... So I'll listen to `domElement` events instead.

src/Renderer/ThreeExtended/PlanarControls.js
autra: To elaborate: sometimes you change the position of the camera in an event (ex `handleRotation`), sometimes you change it in `update` (ex the drag logic). I believe that we should do calculations in the `update` macrotask (meaning: in the same macrotask of the `requestAnimationFrame`) as much as possible, not in the events macrotasks. So here, you'd only need to record positions.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+
+    /**
+    * returns the world point (xyz) under the posXY screen point
+    * the point belong to an abstract mathematical plane of specified altitude (doesnt use actual geometry)
+    * @param {THREE.Vector2} posXY : the mouse position in screen space (unit : pixel)
+    * @param {number} altitude : the altitude (z) of the mathematical plane
+    * @returns {THREE.Vector3}
+    */
+    this.getWorldPointFromMathPlaneAtScreenXY = function getWorldPointFromMathPlaneAtScreenXY(posXY, altitude) {
+        const vector = new THREE.Vector3();
```
autra: Let's put this temp var in a closure to avoid instanciating it at each frame (by experience, creating these temp vars is very costly and the cause of every movement stuttering I had when implementing controls). We only need one per functions.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    /**
+    * returns the world point (xyz) under the posXY screen point
+    * if geometry is under the cursor, the point in obtained with getPickingPositionFromDepth
+    * if no geometry is under the cursor, the point is obtained with getWorldPointFromMathPlaneAtScreenXY
+    * @param {THREE.Vector2} posXY : the mouse position in screen space (unit : pixel)
+    * @returns {THREE.Vector3}
+    */
+    this.getWorldPointAtScreenXY = function getWorldPointAtScreenXY(posXY) {
+        // the returned value
+        const pointUnderCursor = new THREE.Vector3();
```
autra: Put this temp var in a closure.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    };
+
+    /**
+    * returns the world point (xyz) under the posXY screen point
+    * the point belong to an abstract mathematical plane of specified altitude (doesnt use actual geometry)
+    * @param {THREE.Vector2} posXY : the mouse position in screen space (unit : pixel)
+    * @param {number} altitude : the altitude (z) of the mathematical plane
+    * @returns {THREE.Vector3}
+    */
+    this.getWorldPointFromMathPlaneAtScreenXY = function getWorldPointFromMathPlaneAtScreenXY(posXY, altitude) {
```
autra: I have the feeling this might be better in PlanarView (same thing for `getWorldPointAtScreenXY`). I'm not sure about that though. WDYT?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        // the world point under mouse cursor when the drag movement is started
+        this.dragStart.copy(this.getWorldPointAtScreenXY(this.mousePosition));
+
+        // the difference between start and end cursor position
+        this.dragDelta.set(0, 0, 0);
+    };
+
+    /**
+    * Handle the drag movement (translation on xy plane) when user moves the mouse while in STATE.DRAG
+    * The drag movement is previously initiated when user does a left-click, by initiateDrag()
```
autra: > The drag movement is previously initiated when user does a left-click, by initiateDrag()

Maybe replace by

> The drag movement is previously initiated by initiateDrag()

because the fact that drag movement is done by left-click is an ergonomy detail (and could be made configurable in the future).

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    */
+    this.handlePanMovement = function handlePanMovement() {
+        // normalized (betwwen 0 and 1) distance between groundLevel and maxAltitude
+        const distToGround = THREE.Math.clamp((this.camera.position.z - this.groundLevel) / this.maxAltitude, 0, 1);
+
+        // pan movement speed, adujsted according to altitude
+        const panSpeed = THREE.Math.lerp(this.minPanSpeed, this.maxPanSpeed, distToGround);
+
+        // lateral movement (local x axis)
+        this.camera.position.copy(this.camera.localToWorld(new THREE.Vector3(panSpeed * -1 * this.deltaMousePosition.x, 0, 0)));
```
autra: Use `this.camera.position.set(x, y, z)` directly.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+
+    /**
+    * Initiate a rotate (orbit) movement when user does a right-click or ctrl + left-click
+    */
+    this.initiateRotation = function initiateRotation() {
+        this.state = STATE.ROTATE;
+
+        const screenCenter = new THREE.Vector2(0.5 * window.innerWidth, 0.5 * window.innerHeight);
+
+        this.centerPoint.copy(this.getWorldPointAtScreenXY(screenCenter));
```
autra: You could pass directly `{x: 0.5 * window.innerWidth, y: 0.5 * window.innerHeight}` to `getWorldPointAtScreenXY`, I *think* it'll be cheaper than to create a Vector2... or modify getWorldPointAtScreenXY to take directly x and y? Caller would not need to create an object this way.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        }
+    };
+
+    /**
+    * Initiate a rotate (orbit) movement when user does a right-click or ctrl + left-click
+    */
+    this.initiateRotation = function initiateRotation() {
+        this.state = STATE.ROTATE;
+
+        const screenCenter = new THREE.Vector2(0.5 * window.innerWidth, 0.5 * window.innerHeight);
```
autra: You want the center of the canvas, not the window (remember, the canvas can be integrated in whatever webpage). You already have a ref to the canvas (`domElement`) and you'll need to use `clientHeight` and `clientWidth` (I guess). Please also change other instances of windows.innerWidth in this file.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    /**
+    * Handle the rotate movement (orbit) when user moves the mouse while in STATE.ROTATE
+    * the movement is an orbit around 'centerPoint', the camera focus point (ground point at screen center)
+    * The rotate movement is previously initiated in initiateRotation()
+    * Compute the new position value and update the camera controls.
+    */
+    this.handleRotation = function handleRotation() {
+        // angle deltas
+        // deltaMousePosition is computed in onMouseMove / onMouseDown s
+        const thetaDelta = -this.rotateSpeed * this.deltaMousePosition.x / window.innerWidth;
```
autra: same, I'm pretty sure you want the canvas clientWidth.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        const phiDelta = -this.rotateSpeed * this.deltaMousePosition.y / window.innerHeight;
+
+        // the vector from centerPoint (focus point) to camera position
+        const offset = this.camera.position.clone().sub(this.centerPoint);
+
+        const quat = new THREE.Quaternion().setFromUnitVectors(this.camera.up, new THREE.Vector3(0, 0, 1));
```
autra: Create the quaternion in a closure.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        }
+    };
+
+    /**
+    * Triggers a 'smart zoom' animated movement (travel) toward the point under mouse cursor
+    * The camera will be smoothly moved and oriented close to the target, at a determined height and distance
+    */
+    this.initiateSmartZoom = function initiateSmartZoom() {
+        // point under mouse cursor
+        const pointUnderCursor = this.getWorldPointAtScreenXY(this.mousePosition);
```
autra: Should we limit the smartzoom action to a click in the plane limits?

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        // direction of the movement, projected on xy plane and normalized
+        const dir = new THREE.Vector3();
+        dir.copy(pointUnderCursor).sub(this.camera.position);
+        dir.z = 0;
+        dir.normalize();
+
+        const distanceToPoint = this.camera.position.distanceTo(pointUnderCursor);
+
+        // camera height (altitude above ground) at the end of the travel
+        const targetHeight = THREE.Math.lerp(this.smartZoomHeightMin, this.smartZoomHeightMax, Math.min(distanceToPoint / 5000, 1));
```
autra: 5000 seems to be a magic number... Beside, I would have expected something involving only z, not all components of the distance, can you elaborate on this calculus?

autra: Or seeing what happens below, maybe `targetHeight` could be named differently?

Btw, "I found it nicer this way" is a perfect valid answer here :-)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        // start position (current camera position)
+        this.travelStartPos.copy(this.camera.position);
+
+        // start rotation (current camera rotation)
+        this.travelStartRot.copy(this.camera.quaternion);
+
+        // setup the end rotation :
+
+        // case where targetOrientation is a quaternion
+        if (typeof targetOrientation.w !== 'undefined') {
```
autra: Then the test should be `if (targetOrientation instanceof THREE.Quaternion)`?

autra: (it would also work if `targetOrientation` is undefined this way)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    */
+    this.initiateTravel = function initiateTravel(targetPos, travelTime, targetOrientation, useSmooth) {
+        this.state = STATE.TRAVEL;
+        this.view.notifyChange(true);
+        // the progress of the travel (animation alpha)
+        this.travelAlpha = 0;
+        // update cursor
+        this.updateMouseCursorType();
+
+        this.travelUseRotation = !(targetOrientation === 'none');
```
autra: I would test `targetOrientation instanceOf THREE.Quaternion || targetOrientation instanceOf THREE.Vector3`, it would also cover `undefined`, `null`, whatever string and invalid objects the caller gave you. Here the caller needs to remember to pass `'none'` to explicitly disable rotation.

Not using `isVector3` because this would throw an error if undefined or null. (I don't know why Vector3 implementors felt the need to have this boolean, it's useless imho).

autra: (I assume you don't want a rotation to the default value of `travelEndRot` if `targetOrientation` is invalid)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+
+        // end position
+        this.travelEndPos.copy(targetPos);
+
+        // beginning of the travel duration setup
+
+        if (this.instantTravel) {
+            this.travelDuration = 0;
+        }
+
```
autra: nits: remove empty line (a reader could miss the `else if` below)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+        // case where travelTime is set to 'auto' : travelDuration will be a value between autoTravelTimeMin and autoTravelTimeMax
+        // depending on travel distance and travel angular difference
+        else if (travelTime === 'auto') {
+            // a value between 0 and 1 according to the travel distance. Adjusted by autoTravelTimeDist parameter
```
autra: then `autoTravelTimeDist` is a factor, not a real distance? Maybe we could rename it then.

src/Renderer/ThreeExtended/PlanarControls.js
autra: (and you need to use canvas.clientHeight)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,736 @@
...
+    /**
+    * Triggers an animated movement (travel) to set the camera to top view, above the focus point, at altitude=distanceToFocusPoint
+    */
+    this.goToTopView = function goToTopView() {
+        const topViewPos = new THREE.Vector3();
+        const targetQuat = new THREE.Quaternion();
+        const screenCenter = new THREE.Vector2(0.5 * window.innerWidth, 0.5 * window.innerHeight);
+
+        topViewPos.copy(this.getWorldPointAtScreenXY(screenCenter));
+        topViewPos.z += Math.min(this.maxAltitude, this.camera.position.distanceTo(topViewPos));
```
autra: Can you explain this line to me? Why isn't it always `maxAltitude`?

autra (changes requested):
Taking over this, as @peppsac is off. Sorry for the inconvenience to have another reviewer with different expectations stepping in :-/

On overall I really like it! The ergonomy (and the code) is good :-) I made a lot of comments, but most are minor.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -505,16 +526,20 @@ function PlanarControls(view, options = {}) {
...
-        vector.set((posXY.x / window.innerWidth) * 2 - 1, -(posXY.y / window.innerHeight) * 2 + 1, 0.5);
-        vector.unproject(this.camera);
-        const dir = vector.sub(this.camera.position).normalize();
-        const distance = (altitude - this.camera.position.z) / dir.z;
-        const pointUnderCursor = this.camera.position.clone().add(dir.multiplyScalar(distance));
-
-        return pointUnderCursor;
-    };
+        let dir = new THREE.Vector3();
+        let pointUnderCursor = new THREE.Vector3();
```
autra: It's not a big deal but, `dir` does not need to be declared here. You can declare it when you first use it. Anyway, `dir` and `vector` will store the same pointer, because `unproject`, `sub`, `normalize` and `multiplyScalar` all modify the vector3 inplace and returns it (see for instance [how `sub` is implemented](https://github.com/mrdoob/three.js/blob/master/src/math/Vector3.js#L161)).

On the other hand, `pointUnderCursor` **must not** be declared here, otherwise we would always return the same pointer, which is dangerous. We need to create a new instance each time, which is done by the `clone` anyway. So in short, just
```js
return this.camera.clone().add(dir.multiplyScalar(distance));
``` 
instead of line 538-540 is fine (that being said, you can keep `pointUnderCursor` if you find it clearer, just declare it with const on line 538).

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -523,67 +548,78 @@ function PlanarControls(view, options = {}) {
     * @param {THREE.Vector2} posXY : the mouse position in screen space (unit : pixel)
     * @returns {THREE.Vector3}
     */
-    this.getWorldPointAtScreenXY = function getWorldPointAtScreenXY(posXY) {
+    this.getWorldPointAtScreenXY = (() => {
         // the returned value
         const pointUnderCursor = new THREE.Vector3();
+        return (posXY) => {
+            // check if there is valid geometry under cursor
+            if (typeof this.view.getPickingPositionFromDepth(posXY) !== 'undefined') {
```
autra: Sorry I missed that in my previous review: let's put `this.view.getPickingPositionFromDepth(posXY)` in a variable to avoid calling it 2 times.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -212,10 +223,10 @@ function PlanarControls(view, options = {}) {
         const panSpeed = THREE.Math.lerp(this.minPanSpeed, this.maxPanSpeed, distToGround);
 
         // lateral movement (local x axis)
-        this.camera.position.copy(this.camera.localToWorld(new THREE.Vector3(panSpeed * -1 * this.deltaMousePosition.x, 0, 0)));
+        this.camera.position.copy(this.camera.localToWorld(new THREE.Vector3(panSpeed * -1 * deltaMousePosition.x, 0, 0)));
```
autra:
```js
this.handlePanMovement  = (() => {
    const tmp = new THREE.Vector3();
    
    return () => {
         // ....
         tmp.set(panSpeed * -1 * deltaMousePosition.x, 0, 0);
         this.camera.position.copy(this.camera.localToWorld(tmp));
         // ...
   }
})();
```
should prevent one instanciation.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -243,53 +252,58 @@ function PlanarControls(view, options = {}) {
...
+            // angle deltas
+            // deltaMousePosition is computed in onMouseMove / onMouseDown s
+            const thetaDelta = -this.rotateSpeed * deltaMousePosition.x / this.domElement.clientWidth;
+            const phiDelta = -this.rotateSpeed * deltaMousePosition.y / this.domElement.clientHeight;
+
+            // the vector from centerPoint (focus point) to camera position
+            const offset = this.camera.position.clone().sub(centerPoint);
+
+            quat.setFromUnitVectors(this.camera.up, new THREE.Vector3(0, 0, 1));
+            quatInverse = quat.clone().inverse();
```
autra: what about `quat.setFromUnitVectors(this.camera.up, new THREE.Vector3(0, 0, 1)).inverse()` and using `quat` instead of quatInverse below? It saves a `clone()`.

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -243,53 +252,58 @@ function PlanarControls(view, options = {}) {
...
-            }
-            if (thetaDelta !== 0) {
-                // rotation around Z (azimuth)
-
-                const rotationZQuaternion = new THREE.Quaternion();
-                rotationZQuaternion.setFromAxisAngle(new THREE.Vector3(0, 0, 1), thetaDelta);
-                offset.applyQuaternion(rotationZQuaternion);
+    this.handleRotation = (() => {
+        const quat = new THREE.Quaternion();
+        let quatInverse = new THREE.Quaternion();
```
autra: Please "cache" here:

- `new THREE.Vector3(0, 0, 1)` on line 268
- `rotationXQuaternion` and `vector` on line 279-280
- `rotationZQuaternion` on line 290

src/Renderer/ThreeExtended/PlanarControls.js
autra: And it's totally my fault, because I was the one to ask you about `pointUnderCursor`. Sorry :-/

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -523,67 +548,78 @@ function PlanarControls(view, options = {}) {
...
+        return (posXY) => {
+            // check if there is valid geometry under cursor
+            if (typeof this.view.getPickingPositionFromDepth(posXY) !== 'undefined') {
+                pointUnderCursor.copy(this.view.getPickingPositionFromDepth(posXY));
+            }
+            // if not, we use the mathematical plane at altitude = groundLevel
+            else {
+                pointUnderCursor.copy(this.getWorldPointFromMathPlaneAtScreenXY(posXY, this.groundLevel));
+            }
+            return pointUnderCursor;
```
autra: same thing as above. Sorry, my fault :-/

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -0,0 +1,743 @@
...
+            case STATE.ROTATE:
+                this.domElement.style.cursor = 'move';
+                break;
+            default:
+                break;
+        }
+    };
+
+    PlanarControls.prototype = Object.create(THREE.EventDispatcher.prototype);
+    PlanarControls.prototype.constructor = PlanarControls;
```
autra: You're not using `dispatchEvent`, so you don't need the 2 lines above.

autra (changes requested):
Nearly there :-)

autra (approved):

autra:
Yay thanks \o/

autra:
Nice first contrib btw !

EmmanuelSchmuck:
Thank you :)

PR #457, opened 2017-08-28
Refactor: fail fast if a protocol is given but unknown

### LINKED ISSUE #455: addLayer protocol parameter "wms" given in uppercase throws unhandled errors
If we add a WMS layer with the globe.addLayer function, the parameter "protocol" needs to be in lowerCase.
If not, all the other tiles queries are freezed (for any layer), and the following error is launched : 
```
Uncaught TypeError: m.tileInsideLimit is not a function
    at Object.hasEnoughTexturesToSubdivide
```

These errors could be handled and not completely freeze the view.
Moreover, maybe the API documentation could be more explicit about the importance of the case of the "protocol" parameter (or even better, the function could accept any case for the protocol parameter ! ).

autra:
Partially fixes #455 

Making a typo in layer.protocol yields unrelated error message and this could be painful to application devs. Let's be nice and warn them as soon as possible.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -259,7 +259,11 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         layer.extent = parentLayer.extent;
     }
 
-    layer = _preprocessLayer(this, layer, this.mainLoop.scheduler.getProtocolProvider(layer.protocol));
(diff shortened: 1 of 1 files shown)


autra:
@iTowns/reviewers please r?

gchoqueux (approved):

PR #458, opened 2017-08-29
fix (api): getZoom debug

vcoindet:
With the miniGlobe, getZoom used to always return 0.

getZoom() needs the preSSE wich is calculated with the size of the canvas. But with the miniGlobe, preSSE was calculated with the size of the miniGlobe's canvas, so it was wrong.

To solve this I added IDs to the canvas elements, so now preSSE is calculated with the good size.

### DIFF
--- a/src/Process/GlobeTileProcessing.js
+++ b/src/Process/GlobeTileProcessing.js
@@ -12,6 +12,18 @@ let SSE_SUBDIVISION_THRESHOLD;
 
 const worldToScaledEllipsoid = new THREE.Matrix4();
 
+function _preSSE(view) {
+    const canvasSize = view.mainLoop.gfxEngine.getWindowSize();
+    const hypotenuse = canvasSize.length();
+    const radAngle = view.camera.camera3D.fov * Math.PI / 180;
+
+     // TODO: not correct -> see new preSSE
+    // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
+    const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / canvasSize.x);
+
+    preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
+}
+
 export function preGlobeUpdate(context, layer) {
     // We're going to use the method described here:
     //    https://cesiumjs.org/2013/04/25/Horizon-culling/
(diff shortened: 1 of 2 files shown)


gchoqueux:
You must add param `view` in `computeTileZoomFromDistanceCamera` to compute the good `preSSE`

vcoindet:
Thanks @gchoqueux !
What do you think now ?

src/Process/GlobeTileProcessing.js
```diff
@@ -12,6 +12,20 @@ let SSE_SUBDIVISION_THRESHOLD;
...
+    const hypotenuse = canvasSize.length();
+    const radAngle = view.camera.camera3D.fov * Math.PI / 180;
+
+     // TODO: not correct -> see new preSSE
+    // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
+    const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / canvasSize.x);
+
+    preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
+
+    return preSSE;
```
gchoqueux: unused return?

vcoindet: removed

gchoqueux (approved):

src/Process/GlobeTileProcessing.js
```diff
@@ -12,6 +12,18 @@ let SSE_SUBDIVISION_THRESHOLD;
...
+function _preSSE(view) {
+    const canvasSize = view.mainLoop.gfxEngine.getWindowSize();
+    const hypotenuse = canvasSize.length();
+    const radAngle = view.camera.camera3D.fov * Math.PI / 180;
+
+     // TODO: not correct -> see new preSSE
+    // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
+    const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / canvasSize.x);
+
+    preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
```
autra, line 24: Here `preSSE` is shared between all views, therefore between all globe layers. As it depends on the canvas size, therefore on the renderer, therefore on the view, there is a risk of invalid value (basically if someone uses it before calling `_preSSE()`). I suggest making preSSE a field on `view` instead of a module variable. This way, you don't even need to call `_presSSE` on `computeTileZoomFromDistanceCamera`, `view.preSSE` will always be correct :-)

gchoqueux:
yes, 
preSSE is computed with view and camera. It could return in the fields of the camera (in the case where there will be several camera and one viewer)

src/Process/GlobeTileProcessing.js
gchoqueux: yes, 
preSSE is computed with view and camera. It could return in the fields of the camera (in the case where there will be several camera and one viewer)

PR #459, opened 2017-08-29
fix (3dtiles): fix SSE computation

NikoSaul:
Use distance to bounding volume instead of distance to center of the bounding volume
Issue: #185

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -1,5 +1,3 @@
-import * as THREE from 'three';
-
 function requestNewTile(view, scheduler, geometryLayer, metadata, parent) {
     const command = {
         /* mandatory */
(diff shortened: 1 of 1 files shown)


nguyenign:
It seems to me that it continues to download all the tiles!

NikoSaul:
Can you give me your tileset.json please? 
Else this PR doesn't fix all problems for the issue #185 but just the sse.

nguyenign:
you can find the download link in this issue, but if it's not enough i can send you a bigger 3dtiles.

Jeremy-Gaillard:
Your problem isn't really a bug. We tried making your tileset work and we identified two issues:
- Missing `refine: "ADD` fields in some of the tiles
- The SSE refinement threshold is much higher in Cesium (16 by default, compared to 1 in iTowns)

So it works properly, it's just that the building disappear only when you're very far away.

We will make a PR to change the SSE threshold and make it customisable.

nguyenign:
- We need to define only one time refine, then it's optional! 

"The refine property is a string that is either "REPLACE" for replacement refinement or "ADD" for additive refinement. It is required for the root tile of a tileset; it is optional for all other tiles. When refine is omitted, it is inherited from the parent tile." 

- i think we should use dynamic sse as the link i sent you in chat.

can you please commit your modif?

autra (approved):

autra:
LGTM Thanks!

Jeremy-Gaillard:
> We need to define only one time refine, then it's optional!

My bad, I'll add it to the list of thing to fix.

nguyenign:
ok jeremy but do attention, refine is not fixed, we can have many small 3dtiles in a big 3dtiles with which use different refine method.

PR #460, opened 2017-08-29
feature (3dtiles): sliders for sse Threshold

NikoSaul:
feature (3dtiles): sse subdivision threshold can be manually defined

- Default threshold is now 16 instead of 1
- This commit also adds a new slider to modify the threshold, for each tileset in the 3d-tiles example

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -130,6 +130,7 @@
             $3dTilesLayerRequestVolume.overrideMaterials = true;  // custom cesium shaders are not functional
             $3dTilesLayerRequestVolume.type = 'geometry';
             $3dTilesLayerRequestVolume.visible = true;
+            $3dTilesLayerRequestVolume.sseThreshold = 1;            
             // add an event for have information when you move your mouse on a building
             itowns.View.prototype.addLayer.call(globe, $3dTilesLayerRequestVolume).then(function _() { window.addEventListener('mousemove', picking, false); })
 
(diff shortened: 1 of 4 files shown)


autra (approved):
LGTM

nguyenign:
with help of slider, i saw that the performance is better with sse =25. 

Other questions : 
- do we have a cache for 3dtiles? it seem that the performance is degraded when we navigate, the 3dtiles take too much ressource. 
- what boundingVolume you use for frustum culling? tile's content or tile? the boundingVolume in tile's content is better for frustum culling because it's bbox of tile's element (ex. building). Other is defined for spatial subdivision (ex.octree, quatree). I saw many building are not shown in scene, specially on frustum border. 
- some buildingSome building appears and disappears when we navigate.

autra:
> do we have a cache for 3dtiles? it seem that the performance is degraded when we navigate, the 3dtiles take too much ressource.

Let's profile the application before proposing a cache that might not be relevant to the issue. I agree it could have better perf, but we don't know yet what is slow.

nguyenign:
>Let's profile the application before proposing a cache that might not be relevant to the issue. I agree it could have better perf, but we don't know yet what is slow.

the rendering is slow after a period of navigation because resource is occupied a lot by "unused" data?

Jeremy-Gaillard:
> do we have a cache for 3dtiles? it seem that the performance is degraded when we navigate, the 3dtiles take too much ressource.

No cache.

> what boundingVolume you use for frustum culling? tile's content or tile? the boundingVolume in tile's content is better for frustum culling because it's bbox of tile's element (ex. building). Other is defined for spatial subdivision (ex.octree, quatree). I saw many building are not shown in scene, specially on frustum border.

We do not use the content bounding box for culling yet (as indicated in #185)

> some buildingSome building appears and disappears when we navigate.

This may be caused by the refine parameter not being inherited.

Issue #153, opened 2016-07-12
Rework GlobeControl


gchoqueux:
GlobeControl needs recasting.
- [x] remove scope
- [x] add inertia in move globe
- [x] refactoring
- [ ] debuging
- [x] clean up
- [ ]  Recast touch for globe
- [x]  Add damping move and  animate move camera
- [x]  Add real collision
- [ ] Bugs
  - Fix target problem with pan and panoramic (when target isn't on globe)
  - Fix problem with space

gchoqueux:
moved in #545

PR #465, opened 2017-08-30
feature: add collision handling for camera with terrain

nosy-b:
This PR adds a test for collision between the camera and the terrain (geometry layer) for all Views (Planar, Globe...) (issue #153).

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -1,9 +1,3 @@
-/**
- * Generated On: 2015-10-5
- * Class: Scene
- * Description: La Scene est l'instance principale du client. Elle est le chef orchestre de l'application.
- */
-
 /* global window, requestAnimationFrame */
 import { Scene, EventDispatcher, Vector2 } from 'three';
 import Camera from '../Renderer/Camera';
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -4,6 +4,7 @@
 
 import * as THREE from 'three';
 import Coordinates from '../Core/Geographic/Coordinates';
+import DEMUtils from '../utils/DEMUtils';
 
 function Camera(crs, width, height, options = {}) {
     Object.defineProperty(this, 'crs', { get: () => crs });
@@ -98,6 +99,31 @@ Camera.prototype.box3SizeOnScreen = function box3SizeOnScreen(box3, matrixWorld)
         tmp.box3.applyMatrix4(this._viewMatrix);
     }
     return tmp.box3;
+};
+
+ /**
+ * Test for collision between camera and a geometry layer (DTM/DSM) to adjust camera position
+ * It could be modified later to handle an array of geometry layers
+ * TODO Improve Coordinates class to handle altitude for any coordinate system (even projected one)
+ * @param {view} view where we test the collision between geometry layers and the camera
+ * @param {elevationLayer} elevationLayer (DTM/DSM) used to test the collision with the camera. Could be another geometry layer
+ * @param {minDistanceCollision} minDistanceCollision the minimum distance allowed between the camera and the surface
+ */
+Camera.prototype.adjustAltitudeToAvoidCollisionWithLayer = function adjustAltitudeToAvoidCollisionWithLayer(view, elevationLayer, minDistanceCollision) {
+    // We put the camera location in geographic by default to easily handle altitude. (Should be improved in Coordinates class for all ref)
+    const camLocation = view.camera.position().as('EPSG:4326');
+    if (elevationLayer !== undefined) {
+        const elevationUnderCamera = DEMUtils.getElevationValueAt(elevationLayer, camLocation);
+        if (elevationUnderCamera != undefined) {
+            const difElevation = camLocation.altitude() - (elevationUnderCamera.z + minDistanceCollision);
+            // We move the camera to avoid collisions if too close to terrain
+            if (difElevation < 0) {
+                camLocation.setAltitude(elevationUnderCamera.z + minDistanceCollision);
+                view.camera.camera3D.position.copy(camLocation.as(view.referenceCrs).xyz());
+                view.notifyChange(true);
+            }
+        }
+    }
 };
 
 export default Camera;
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -16,7 +16,6 @@ import { computeTileZoomFromDistanceCamera, computeDistanceCameraFromTileZoom } 
 // Recast touch for globe
 // Fix target problem with pan and panoramic (when target isn't on globe)
 // Fix problem with space
-// Add real collision
 
 // FIXME:
 // when move globe in damping orbit, there isn't move!!
@@ -410,6 +409,10 @@ function GlobeControls(view, target, radius, options = {}) {
     this.minAzimuthAngle = -Infinity; // radians
     this.maxAzimuthAngle = Infinity; // radians
 
+    // Set collision options
+    this.handleCollision = typeof (options.handleCollision) !== 'undefined' ? options.handleCollision : true;
+    this.minDistanceCollision = 100;
+
     // Set to true to disable use of the keys
     this.enableKeys = true;
 
@@ -573,6 +576,10 @@ function GlobeControls(view, target, radius, options = {}) {
     const axisX = new THREE.Vector3(1, 0, 0);
 
     var update = function update() {
+        // We test if camera collide to geometry layer or too close to ground and ajust it's altitude in case
+        if (this.handleCollision) { // We check distance to the ground/surface geometry. (Could be another geometry layer)
+            this._view.camera.adjustAltitudeToAvoidCollisionWithLayer(this._view, view.getLayers(layer => layer.type === 'geometry')[0], this.minDistanceCollision);
+        }
         // MOVE_GLOBE
         // Rotate globe with mouse
         if (state === CONTROL_STATE.MOVE_GLOBE) {
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -236,6 +236,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
         this.camera.camera3D.lookAt(positionTargetCamera.as('EPSG:4978').xyz());
     } else {
         this.controls = new GlobeControls(this, positionTargetCamera.as('EPSG:4978').xyz(), size);
+        this.controls.handleCollision = typeof (options.handleCollision) !== 'undefined' ? options.handleCollision : true;
     }
 
     this._renderState = RendererConstant.FINAL;
--- a/src/Renderer/ThreeExtended/PlanarControls.js
+++ b/src/Renderer/ThreeExtended/PlanarControls.js
@@ -88,6 +88,10 @@ function PlanarControls(view, options = {}) {
     this.focusOnMouseOver = options.focusOnMouseOver || true;
     this.focusOnMouseClick = options.focusOnMouseClick || true;
 
+    // Set collision options
+    this.handleCollision = typeof (options.handleCollision) !== 'undefined' ? options.handleCollision : true;
+    this.minDistanceCollision = 30;
+
     // starting camera position and orientation target are setup before instanciating PlanarControls
     // using: view.camera.setPosition() and view.camera.lookAt()
     // startPosition and startQuaternion are stored to be able to return to the start view
@@ -148,6 +152,10 @@ function PlanarControls(view, options = {}) {
 
     // Updates the view and camera if needed, and handles the animated travel
     this.update = function update(dt, updateLoopRestarted) {
+        // We test if camera collide to geometry layer or too close to ground and ajust it's altitude in case
+        if (this.handleCollision) { // We check distance to the ground/surface geometry. (Could be another geometry layer)
+            this.view.camera.adjustAltitudeToAvoidCollisionWithLayer(this.view, view.getLayers(layer => layer.type === 'geometry')[0], this.minDistanceCollision);
+        }
         // dt will not be relevant when we just started rendering, we consider a 1-frame move in this case
         if (updateLoopRestarted) {
             dt = 16;


src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -642,6 +642,58 @@ function GlobeControls(view, target, radius, options = {}) {
...
+            }
+        }
+
+        // We first test if camera is very close to the tile bounding box
+        if (currentTile) {
+            const alti = currentTile.OBB().z.max + this.minDistance;
+            const distance = camLocation.altitude() - alti;
+            // If inside, we then need more precision: the exact elevation under camera location
+            if (distance < 0) {
+                const corners = currentTile.OBB()._cPointsWorld(currentTile.OBB()._points());
```
autra: this is actually `currentTile.OBB().pointsWorld()`

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -642,6 +642,58 @@ function GlobeControls(view, target, radius, options = {}) {
...
+                const AB = corners[0].clone().sub(corners[1]);
+                const AC = corners[2].clone().sub(corners[1]);
+                const pNormedProjectedOnAB = pNormed.clone().projectOnVector(AB);
+                const pNormedProjectedOnAC = pNormed.clone().projectOnVector(AC);
+                let u = pNormedProjectedOnAB.length() / AB.length();
+                let v = pNormedProjectedOnAC.length() / AC.length();
+                const offsetScale = currentTile.material.uniforms.offsetScale_L00.value[0];
+                u = offsetScale.x + u * offsetScale.z;
+                v = offsetScale.y + v * offsetScale.z;
+                if (currentTile.material.textures[0][0].image) {
```
autra: `if (currentTile.isElevationLayerLoaded())` should work.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -642,6 +642,58 @@ function GlobeControls(view, target, radius, options = {}) {
...
+                    const w = currentTile.material.textures[0][0].image.width;
```
autra: Not a big deal, but I find it more readable to write:
```js
const image = currentTile.getLayerTextures(l_ELEVATION)[0].image;
const w = image.width;`
```
and later line 687
```js
const elevationUnderCamera = image.data[indiceElevation];
```

More readable, and *should* be more stable because internal storage could change, but I expect `getLayerTextures` to stay.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -642,6 +642,58 @@ function GlobeControls(view, target, radius, options = {}) {
...
+                    const indiceElevation = Math.floor(v * w) * w + Math.floor(u * w);
```
autra: `elevationIndex`?

autra: In rigor, shouldn't it be 
```js
const indiceElevation = Math.floor(v * h) * w + Math.floor(u * w);
```
with `h = image.height`? (please check this though, I might have mixed up the axis).

In practice, I don't know if we'll have non-squared image, but I also find it easier to understand this way.

autra (dismissed):
Looking mostly good :-) My comments are mainly cosmetics, except the last one (not sure if it has great consequences though).

peppsac:
Adding features to `GlobeControls` (aka: globe only feature) is a bad idea, except for features that really requires a globe.

- we already have https://github.com/iTowns/itowns/blob/master/src/utils/DEMUtils.js to extract altitude information from textures. Please use/adapt it instead of adding new (globe specific) code.
- the collision with tile feature should probably live in `Camera` (or in another shared module) in order to be usable by everyone: PlanarView or View.

nosy-b:
@peppsac Right, I'm working on it

nosy-b:
@iTowns/reviewers Updated as generic method for all type of view with a geometry layer such as globe or planar

src/Core/View.js
```diff
@@ -291,8 +295,31 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+ */
+View.prototype.checkCollisionWithGeometryLayer = function checkCollisionWithGeometryLayer() {
+    const camLocation = new Coordinates(this.referenceCrs, this.camera.camera3D.position).as('EPSG:4326');
+    if (this._layers[0] && this._layers[0] instanceof GeometryLayer) {
+        const elevationUnderCamera = DEMUtils.getElevationValueAt(this._layers[0], camLocation);
+        if (elevationUnderCamera != undefined) {
+            const difElevation = camLocation.altitude() - (elevationUnderCamera.z + this.minDistanceCollision);
+            // We move the camera to avoid collisions if too close to terrain
+            if (difElevation < 0) {
+                camLocation.setAltitude(elevationUnderCamera.z + this.minDistanceCollision);
```
peppsac: I'm not sure all controls instances support external changes of camera's position.

src/Core/View.js
```diff
@@ -291,8 +295,31 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+    }
+};
+
+
+/**
+ * Test for collision between camera and geometry layer (DTM/DSM) to adjust camera position
+ */
+View.prototype.checkCollisionWithGeometryLayer = function checkCollisionWithGeometryLayer() {
+    const camLocation = new Coordinates(this.referenceCrs, this.camera.camera3D.position).as('EPSG:4326');
+    if (this._layers[0] && this._layers[0] instanceof GeometryLayer) {
```
peppsac: You can't really expect the first layer to always be the geometry one (even if it's true in GlobeView/PlanarView)

peppsac (changes requested):
Now that this PR uses `DEMUtils` it's clearer that the code handling camera<->layer collision doesn't belong to `View`.

I can think of 2 alternatives:
- add a function to `Camera`: `adjustAltitudeToAvoidCollisionWithLayer` and use this function from controls. 
- or add collision handling code to controls classes (because they're in charge of camera's position)

In both cases, collision handling would be optional, and enabled by default in Globe/PlanarView.

WDYT?

nosy-b:
@peppsac Agreed, the code location for collision handling is not ideal... but I liked the idea that any controls doesn't have to add a line to handle it. 
I change it now , commit very soon

autra:
My 2 cents: let's enable it by default for our plane and globe examples while keeping in mind that some applications do *not* want collision detection (I'm working for an application using itowns, where some data are actually underground :-) ).

nosy-b:
It's enabled by default for globeview and planarview but the default camera setting is set to not care about collision as we said.
@peppsac  WDYT? Ok for 2.1? Could be improved maybe later with smarter collision tracking with inertial speed variation for example.

src/Renderer/Camera.js
```diff
@@ -13,6 +14,7 @@ function Camera(crs, width, height, options = {}) {
     this._viewMatrix = new THREE.Matrix4();
     this.width = width;
     this.height = height;
+    this.handleCollision = false;
```
peppsac: For now I'd prefer this option to exist only on the controls that support it.
So I'd add a `handleCollision` to `GlobeControls` and `PlanarControls` and some code in `PlanerView`/`GlobeView` constructor. For instance:
```
this.controls.handleCollision = options.handleCollision || false;
```

src/Renderer/Camera.js
```diff
@@ -100,4 +102,25 @@ Camera.prototype.box3SizeOnScreen = function box3SizeOnScreen(box3, matrixWorld)
...
 };
 
+ /**
+ * Test for collision between camera and geometry layer (DTM/DSM) to adjust camera position
+ * @param {view} view where we test the collision between geometry layers and the camera
+ * @param {minDistanceCollision} minDistanceCollision the minimum distance allowed between the camera and the surface
+ */
+Camera.prototype.adjustAltitudeToAvoidCollisionWithLayer = function adjustAltitudeToAvoidCollisionWithLayer(view, minDistanceCollision) {
+    const camLocation = new Coordinates(view.referenceCrs, view.camera.camera3D.position).as('EPSG:4326');
+    const elevationLayer = view.getLayers(layer => layer.type === 'geometry')[0];
```
peppsac: You should pass the `layer` as a parameter to this function.
It should be more robust than assuming that it's the first `geometry` one.

peppsac: So this also indicates that instead of storing a boolean `handleCollision` you probably want to store a list of layers that you want the camera to collide with.
So my earlier comment example code would become:
```
this.controls.collisionLayers = options.handleCollision ? [wgs84TileLayer] : [];
```

WDYT?

nosy-b:
@peppsac it's not that simple as PlanarView doesn't reference its controller as opposed to GlobeView.
Nevertheless let's give it a try!

The collision options now sits in PlanarControls/GlobeControls. It's On by default but can be set in their respective options. While instantiating globview it is possibile to set it too.
This collision works with the (unique) geometry layer of the different views but it is now possible to pass any layer as a parameter. 
A very simple upgrade would make it work with an array of layers but to keep it clean I left it like that as we have a specific way to compute the elevation with elevation tiles and we might in the future check collision differently (against 3D tiles buildings for examples)

peppsac:
You commited `Segmentation.js` which is unrelated to this PR :)

Also `this.controls.handleCollision = options.handleCollision || true;` will always result in `this.controls.handleCollision = true`...

nosy-b:
Oups :)

nosy-b:
@peppsac Are you ok for merge?

src/Renderer/Camera.js
```diff
@@ -100,4 +101,26 @@ Camera.prototype.box3SizeOnScreen = function box3SizeOnScreen(box3, matrixWorld)
...
 
+ /**
+ * Test for collision between camera and a geometry layer (DTM/DSM) to adjust camera position
+ * It could be modified later to handle an array of geometry layers
+ * @param {view} view where we test the collision between geometry layers and the camera
+ * @param {elevationLayer} elevationLayer (DTM/DSM) used to test the collision with the camera. Could be another geometry layer
+ * @param {minDistanceCollision} minDistanceCollision the minimum distance allowed between the camera and the surface
+ */
+Camera.prototype.adjustAltitudeToAvoidCollisionWithLayer = function adjustAltitudeToAvoidCollisionWithLayer(view, elevationLayer, minDistanceCollision) {
+    const camLocation = new Coordinates(view.referenceCrs, view.camera.camera3D.position).as('EPSG:4326');
```
peppsac: `EPSG:4326` is wrong here (are you sure you tried in planar mode :) ?)

But the good news is: there's a ready to use function that does what you need: 
```js
const camLocation = view.camera.position(); 
```

src/Renderer/Camera.js
```diff
@@ -100,4 +101,26 @@ Camera.prototype.box3SizeOnScreen = function box3SizeOnScreen(box3, matrixWorld)
...
+    if (elevationLayer !== undefined) {
```
peppsac: IMHO, either you:
- test for `undefined` and throw a useful error
- document that `elevationLayer` cannot be undefined

But testing and then ignoring `undefined` seems like a bad idea.

src/Renderer/Camera.js
```diff
@@ -100,4 +101,26 @@ Camera.prototype.box3SizeOnScreen = function box3SizeOnScreen(box3, matrixWorld)
...
+        const elevationUnderCamera = DEMUtils.getElevationValueAt(elevationLayer, camLocation);
+        if (elevationUnderCamera != undefined) {
+            const difElevation = camLocation.altitude() - (elevationUnderCamera.z + minDistanceCollision);
+            // We move the camera to avoid collisions if too close to terrain
+            if (difElevation < 0) {
+                camLocation.setAltitude(elevationUnderCamera.z + minDistanceCollision);
+                view.camera.camera3D.position.copy(camLocation.as(view.referenceCrs).xyz());
```
peppsac, line 122: cf: https://github.com/iTowns/itowns/pull/465#discussion_r138268173 
"I'm not sure all controls instances support external changes of camera's position."

Did you check?

peppsac: You probably should use `view.notifyChange(...)` here since you modify the camera position.

peppsac (commented):
Sorry but there a still a few things to address before merging.

src/Renderer/Camera.js
nosy-b: No EPSG:4326 is geographic which is needed.

nosy-b: (And yes of course I tested in planar and it works)

src/Renderer/Camera.js
nosy-b: The three different controls we use support it.
But yes I thought in the future about having a function in camera that just compute a new position if collision so the controller itself can modify the camera position using this value.

src/Renderer/Camera.js
peppsac: Good to know that it's working but... it doesn't make sense, right?
Why would you transform from a planar CRS to geographic to query elevation in a layer with a planar CRS?

src/Renderer/Camera.js
peppsac: Ok, let's keep it this way if it's working fine with the available controls.

src/Renderer/Camera.js
nosy-b: I understand your comment but here, it's not an error. If undefined it means there is no collision it's not a proper error

src/Renderer/Camera.js
nosy-b: Yes it looks weird:) That's because we can easily manipulate the altitude in geographic coord using the coordinates class.
But yes I can simplify it to 
```javascript
const camLocation = view.camera.position().as('EPSG:4326');
```
and better, we could also modify Coordinates class so we handle altitude as a height if Coord is not geographic. This would avoid coordinate transform when in projected coordinate system

src/Renderer/Camera.js
nosy-b: (We don't need view.notifyChange as this function is called on a control update which then notify the change)

src/Renderer/Camera.js
nosy-b: I'll rebase and add a comment to the function so it's clearer

peppsac: Ok, should be enough for now.

src/Renderer/Camera.js
peppsac: IMHO it makes no sense to call this function without an elevationLayer.
So cluttering the code with unneeded `if X !== undefined` isn't bringing us any value.

So I'd prefer this function to have a clear contract ("I require an elevationLayer") and leave it to the user the task of not calling it with an invalid value.

src/Renderer/Camera.js
peppsac: > We don't need view.notifyChange [...]

You need to call `view.notifyChange` because you're modifying the camera's position.
The fact that the controls update method may call `view.notifyChange` too is irrelevant.

nosy-b: yes i did it in fact it's rebased

nosy-b: yes i did it in fact it's rebased

peppsac (approved):
I still have a few concerns with this PR but I prefer to move forward and come back to this subject later.

It's a nice feature to have anyway so thanks for the work you put into it :)

PR #467, opened 2017-09-01
fix (api): pan correction

vcoindet:
## Description
Small PR to correct the pan function in GlobeControls.
changed element.clientHeight to element.clientWidth for the left pan.

### DIFF
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -518,7 +518,7 @@ function GlobeControls(view, target, radius, options = {}) {
             targetDistance *= Math.tan((this.camera.fov / 2) * Math.PI / 180.0);
 
             // we actually don't use screenWidth, since perspective camera is fixed to screen height
-            this.panLeft(2 * deltaX * targetDistance / element.clientHeight);
+            this.panLeft(2 * deltaX * targetDistance * this.camera.aspect / element.clientWidth);
             this.panUp(2 * deltaY * targetDistance / element.clientHeight);
         } else if (this.camera instanceof THREE.OrthographicCamera) {
             // orthographic


src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -518,7 +518,7 @@ function GlobeControls(view, target, radius, options = {}) {
             targetDistance *= Math.tan((this.camera.fov / 2) * Math.PI / 180.0);
 
             // we actually don't use screenWidth, since perspective camera is fixed to screen height
-            this.panLeft(2 * deltaX * targetDistance / element.clientHeight);
+            this.panLeft(2 * deltaX * targetDistance / element.clientWidth);
```
autra: actually I'm not sure about that. `targetDistance` is calculated from the camera fov, which is the *vertical* opening of the camera. Hence we normalize with `element.clientHeight`. 

For this reason (I believe), according to my tests, the mouse follow the ground better with the old code when you do some left-right pan. Thoughts?

autra (changes requested):

src/Renderer/ThreeExtended/GlobeControls.js
autra: To precise things after our irc discussion, what I've said above is only true because `camera.aspect` is correctly set to `width / height`. I guess this won't work if someone wants to play with a different aspect (have a skewed scene). The correct calculus should then be:

```js
this.panLeft(2 * deltaX * targetDistance * camera.aspect / element.clientWidth);
```

(sanity check, if camera.aspect is `element.clientWidth / element.clientHeight`, then it's the same as `2 * deltaX * targetDistance / element.clientHeight`).

Do you want to amend your PR with this calculus?

vcoindet: OK, I changed the file as you suggested.
Note that it doesn't change anything in our case right now, but it will work correctly if one allows to change the ratio `width / height`

autra (approved):

autra:
Thanks!

PR #468, opened 2017-09-05
feature: add removeLayer function to View

Jeremy-Gaillard:
Added a function to remove a layer from a view.

How to test:
- Open examples/planar.html
- Run the command `itowns.View.prototype.removeLayer.call(view, 'planar')`

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -282,6 +282,27 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
 };
 
 /**
+ * Removes a layer from the view
+ * @param {string} layerId - the id of the layer to remove
+ */
+View.prototype.removeLayer = function removeLayer(layerId) {
+    const layers = this.getLayers((l => l.id == layerId));
+    if (layers.length === 0) {
+        throw new Error(`No layer with id '${layerId}' to remove`);
+    }
+    const toRemove = this.scene.children.filter((c => c.layer == layerId));
+    for (const obj of toRemove) {
+        const idx = this.scene.children.indexOf(obj);
+        this.scene.children.splice(idx, 1);
+    }
+
+    const idx = this._layers.indexOf(layers[0]);
+    this._layers.splice(idx, 1);
+
+    this.notifyChange(true);
+};
+
+/**
  * Notifies the scene it needs to be updated due to changes exterior to the
  * scene itself (e.g. camera movement).
  * non-interactive events (e.g: texture loaded)


Jeremy-Gaillard:
I'm having trouble re-instantiating the layer after removing it, so don't merge yet.

Jeremy-Gaillard:
Never mind, the problem was elsewhere. The PR should work fine.

src/Core/View.js
```diff
@@ -282,6 +282,27 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
 /**
+ * Removes a layer from the view
+ * @param {string} layerId - the id of the layer to remove
+ */
+View.prototype.removeLayer = function removeLayer(layerId) {
+    const layers = this.getLayers((l => l.id == layerId));
+    if (layers.length === 0) {
+        throw new Error(`No layer with id '${layerId}' to remove`);
+    }
+    const toRemove = this.scene.children.filter((c => c.layer == layerId));
```
autra, line 293: If I'm not mistaken, obj are not necessarily added to `this.scene` any more (see GeometryLayer constructor).

autra (changes requested):
I'm wondering if we can really do things that way for 2 reasons:
1. Layer can have parents. Btw have you looked at `layer.detach`? At the minimum `removeLayer` should properly detach from parent. What happens to children of the layer you're removing? They should get removed as well maybe?
1. Maybe we should call (or better: schedule a call for the next update loop) a `layer.cleanUp` method to give a chance to the layer to clean itself (remove cache, objects to the scene etc...). This call would recursively call cleanUp on children layers. The reason for this: intuitively I would think this cleaning process is particular to each protocol (to each *Processing actually).

I feel that things are not that simple here. On the other hand, what exact problem are you trying to solve right now? If you only need to disable/hide a layer, this could be very simple to implement (if not done already).

Jeremy-Gaillard:
> Layer can have parents

The removeLayer function (in its current form) should only be used on layers that do not depend on a parent (i.e. only those in the _layers array). But you are right about the children needing to be properly detached.

> Maybe we should call a layer.cleanUp method

It would be a good idea indeed.

> I feel that things are not that simple here

I agree, part of the reason I created this PR was to gather feedback.

> what exact problem are you trying to solve right now?

I'm trying to give the user the possibility to reload a dataset with a different configuration. Technically, it means replacing a 3D Tiles layer with another 3D Tiles layer. Disabling or hiding the previous layer(s) doesn't seem very practical for this use case.

src/Core/View.js
Jeremy-Gaillard: True. I should use `layer.object3d` instead of `this.scene`.

autra:
Basically, if I understand correctly, removeLayer should undo what is relevant to undo in `addLayer`, `Layer.{attach,preUpdate,update,postUpdate}` and `Provider.preprocessLayer` (either in itself, or in layer.cleanup). That way, if we call again `addLayer` with the removed layer object, everything would work as expected. Maybe some info can stay (metadata gathered in preprocessLayer for instance, if we think they'll stay relevant), but on overall, itowns should go back to (nearly) the same state it has before the first `addLayer`. What do you think about these principles?

> I'm trying to give the user the possibility to reload a dataset with a different configuration

Interesting! Could you be a bit more specific about what you mean by "reloading a dataset with a different config"? Which configuration options are you talking about?

Jeremy-Gaillard:
Your description seems correct. Though I don't think it's necessary to keep any metadata at all.

> Could you be a bit more specific about what you mean by "reloading a dataset with a different config"

Basically, it's changing the way the data is represented (lod, data type, ...). I'll share a demo of my work soon.

gchoqueux:
Warning `GlobeView` have a method `removeLayer`, it's necessary to factorize the code

peppsac:
@Jeremy-Gaillard maybe we could close this PR in favor of #546 ?

Jeremy-Gaillard:
Indeed we can. Closing.

PR #470, opened 2017-09-07
Feature : pick raster vector

gchoqueux:
## Description

Add tools to pick/read Coordinate.
Add tools to test if Coordinate is in polygon, over line or point.
Add example pick raster vector

## Motivation and Context

After add rasterizer in #450 to show vector datas,  
It's interesting to be able to interact with this data, like picking and extract the descriptions

pickingraster


peppsac:
Thanks for the PR (and the screenshot :))

To ease the review, could you:
- split the 2 fixes commit to other PRs?
- add some details to the commit message of ` fix (Depth) : correct calculation of depth` to explain the proposed changes and what was the problem? Does this fix issue #411?
- split the `use cache depth buffer` as well

Thanks!

gchoqueux:
@peppsac
#470  splited in #471, #472 and #474

gchoqueux:
this PR is ready, because #471, #472 and #474 are merged

src/Core/Geographic/Extent.js
```diff
@@ -232,22 +232,23 @@ Extent.prototype.dimensions = function dimensions(unit) {
  * Return true if coord is inside the bounding box.
  *
  * @param {Coordinates} coord
+ * @param {number} delta value to expand extent
  * @return {boolean}
  */
-Extent.prototype.isPointInside = function isPointInside(coord) {
+Extent.prototype.isPointInside = function isPointInside(coord, delta) {
```
peppsac: I'd use `epsilon = 0` rather than `delta`.

And the jsdoc could be replaced with: `coord is inside the extent (+/- epsilon)`.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1670,6 +1670,18 @@ GlobeControls.prototype.pixelsToMeters = function pixelsToMeters(pixels, pixelPi
 };
 
 /**
+ * To convert the projection in degree WGS84 on the globe of a number of pixels of screen
```
peppsac: This seems to describe the opposite of what the function does?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1670,6 +1670,18 @@ GlobeControls.prototype.pixelsToMeters = function pixelsToMeters(pixels, pixelPi
...
+ * @param      {number} pixels count pixels to project
+ * @param      {number} pixelPitch Screen pixel pitch, in millimeters (default = 0.28 mm / standard pixel size of 0.28 millimeters as defined by the OGC)
+ * @return     {number} projection in degree on globe
+ */
+GlobeControls.prototype.pixelsToDegrees = function pixelsToDegrees(pixels, pixelPitch = 0.28) {
+    const chord = this.pixelsToMeters(pixels, pixelPitch);
+    const rayon = ellipsoidSizes().x;
+    return 2 * Math.asin(chord / (2 * rayon)) * 180 / Math.PI;
```
peppsac: Please use `THREE.Math.radToDeg`: easier to read than `180 / Math.PI`.
And use english var names -> `rayon` -> `radius`...
Btw, I can't see how  this formula can be correct on an ellipsoid with different radius values on each axes?

peppsac: Last but not least: this method doens't belong to `*Controls`.
`GlobeView` is a better place or a new `GlobeUtils` module if there are other utils functions like this one.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
+function overLine(point, vs, precision) {
```
peppsac: What is `vs`?
I'm all for short variable names in small scope (e.g: `for(let i ...)`) but function arguments should have explicit names.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
...
+    const x0 = point._values[0];
+    const y0 = point._values[1];
+    for (var i = 0, j = vs.length - 1; i < vs.length; j = i++) {
```
peppsac: This funny for-loop construct is surely justified but please add a comment to explain the intent.

src/Renderer/ThreeExtended/FeaturesUtils.js
peppsac: Try to stick to the convention and being method names with a verb please.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
...
+
+function insidePolygon(point, vs) {
+    // ray-casting algorithm based on
+    // http://www.ecse.rpi.edu/Homepages/wrf/Research/Short_Notes/pnpoly.html
+
+    const x = point._values[0];
+    const y = point._values[1];
+
+    var inside = false;
+    for (var i = 0, j = vs.length - 1; i < vs.length; j = i++) {
```
peppsac: Same comment on this funny loop.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
...
+        const xi = vs[i]._values[0];
+        const yi = vs[i]._values[1];
+        const xj = vs[j]._values[0];
+        const yj = vs[j]._values[1];
+
+        var intersect = ((yi > y) != (yj > y))
+            && (x < (xj - xi) * (y - yi) / (yj - yi) + xi);
```
peppsac: Please add a one line (or more!) comment describing what is tested here

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
...
+        if (intersect) inside = !inside;
```
peppsac: Style: 
```js
if (intersect) {
    inside = !inside;
}
```

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    return inside;
+}
+
+export default {
+    getFeaturesAtCoordinate(coordinate, collection, precision = 0.1) {
```
peppsac: Documentation ?

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,114 @@
...
+                }
+                /* eslint-disable guard-for-in */
+                for (const id in features.featureVertices) {
+                    const polygon = features.featureVertices[id];
+                    if (polygon.extent && !polygon.extent.isPointInside(coordinate, precision)) {
+                        continue;
+                    }
+                    const properties = collection.features[id].properties.properties;
+                    const coordinates = features.coordinates.slice(polygon.offset, polygon.offset + polygon.count);
+                    if (features.type == 'linestring' && overLine(coordinate, coordinates, precision)) {
```
peppsac: You could probably extract a method to test a given feature in an external function and reuse it in both cases (features collection or single feature).

examples/globe_vector.html
```diff
@@ -49,15 +79,98 @@
...
+                            var visible = false;
+                            var precision = globeView.controls.pixelsToDegrees(precisionPx);
+                            for (var i=0; i < layers.length; i++) {
+                                var layer = layers[i];
+                                    var result = itowns.FeaturesUtils.getFeaturesAtCoordinate(geoCoord, layer.feature, precision);
+                                    var id = 0;
+                                    if (result.points.length || result.lines.length || result.polygons.length) {
+                                        visible = true;
+
+                                        for (var polygon of result.polygons) {
```
peppsac: I don't think you can use `for ... of` in examples (they're not compiled to ES5)

examples/globe_vector.html
```diff
@@ -49,15 +79,98 @@
...
+                                            var color = polygon.properties.fill || layer.style.fill;
+                                            var stroke = polygon.properties.stroke || layer.style.stroke;
+                                            var name = `polygon${id}`;
+                                            var symb = `<span id=${name} >&#9724</span>`;
+                                            tooltip.innerHTML += `${symb}   ${polygon.properties.name || polygon.properties.nom || polygon.properties.description || layer.name}<br />`;
```
peppsac: Same for string templates...

examples/globe_vector.html
```diff
@@ -49,15 +79,98 @@
...
+                    var precisionPx = 5;
+                    function buildToolTip(geoCoord, e) {
+                        tooltip.innerHTML = '';
+                        tooltip.style.visibility = 'hidden';
+                        if (geoCoord) {
+                            var visible = false;
+                            var precision = globeView.controls.pixelsToDegrees(precisionPx);
+                            for (var i=0; i < layers.length; i++) {
+                                var layer = layers[i];
+                                    var result = itowns.FeaturesUtils.getFeaturesAtCoordinate(geoCoord, layer.feature, precision);
```
peppsac: `geoCoord` and `precision in pixels` looks like a weird mix. I'd either go full pixels or full SIG-coord...

peppsac (commented):
Thanks for updating this PR.

I did a first quick review - hopefully most of my comments should be easy to fix and I'll be able to do a more in-depth review.

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: ok
eslint doesn't return error

examples/globe_vector.html
gchoqueux: precision is in degree

src/Core/Geographic/Extent.js

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: There are several functions to move, I'll do a special PR

gchoqueux:
@peppsac updated

examples/globe_vector.html
peppsac: Then this comment is out of date?
` +                    // precision in pixels `

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+export default {
+    /**
+     * filters the features that are under the coordinate
+     *
+     * @param      {Coordinates}  coordinate  the coordinate for the filter condition
+     * @param      {Features}  collection  features collection to filter
+     * @param      {number}  precision  tolerance around the coordinate
```
peppsac: What's the unit? The same as coordinate?

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
+function pointIsOverLine(point, linePoints, precision) {
```
peppsac: In all this file you can replace precision with epsilon too.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+    const x0 = point._values[0];
+    const y0 = point._values[1];
+    // in first j is last point of line and then it'll be (i - 1)
```
peppsac: Ok. 
How about adding: `// for each segment of the line (j is i -1)` 

Also, your code assumes that this is a closed line; I'm not sure this is a valid assumption.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+
+        const Xp = x0 - x1;
+        const Yp = y0 - y1;
+
+        const x21 = x2 - x1;
+        const y21 = y2 - y1;
+        const n = Math.sqrt(x21 * x21 + y21 * y21);
+        const scalar = (Xp * x21 + Yp * y21) / n;
+
+        if (scalar >= -precision && scalar <= n + precision) {
```
peppsac: You still didn't explain the algorithm used.
But this line is weirdly non-symetric (`-precision` on the left but `n + precision` on the right). Are you sure it's correct?

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+    const x0 = point._values[0];
+    const y0 = point._values[1];
+    for (var i = 0; i < points.length; ++i) {
+        const x1 = points[i]._values[0];
+        const y1 = points[i]._values[1];
+        const xP = x0 - x1;
+        const yP = y0 - y1;
+        const n = Math.sqrt(xP * xP + yP * yP);
+        if (n < precision) {
+            return i;
```
peppsac: You're not returning the closest point.
You're returning the first point that matches `distance(point, points[i]) < precision`.

peppsac: (you can avoid a bunch of Math.sqrt as well by comparing against precisionSquared instead)

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+            inside = !inside;
+        }
+    }
+
+    return inside;
+}
+
+
+function getFeatureAtCoordinate(coordinate, feature, coordinates, precision, properties, result) {
+    if (feature.type == 'linestring' && pointIsOverLine(coordinate, coordinates, precision)) {
```
peppsac: If you only use `feature.type` then pass `featureType` as a function argument directly.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+export default {
+    /**
+     * filters the features that are under the coordinate
+     *
+     * @param      {Coordinates}  coordinate  the coordinate for the filter condition
+     * @param      {Features}  collection  features collection to filter
+     * @param      {number}  precision  tolerance around the coordinate
+     * @return     {array}  array of filters features
+     */
+    getFeaturesAtCoordinate(coordinate, collection, precision = 0.1) {
```
peppsac: I don't think `collection` is the correct name of this argument. `features`?

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,122 @@
...
+        const result = { points: [], lines: [], polygons: [] };
```
peppsac: Why do you return only the `geometric` information and not the `features`?

peppsac (changes requested):

gchoqueux:
`// precision in pixels ` is for `precisionPx`
Then `precisionPx` is converted into degree with `pixelsToDegrees`

examples/globe_vector.html
peppsac: Ah ok.

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: I add comment

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: > You're not returning the closest point.

I change that

> you can avoid a bunch of Math.sqrt as well by comparing against precisionSquared instead

yes, thanks

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: I change that, to no closed line

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: I return coordinates and the associated properties.
I can't return features because our object can be a fusion of several features.
This function filters out unwanted features

gchoqueux:
@peppsac updated

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,146 @@
...
+    let squaredEpsilon = epsilon * epsilon;
+    let closestPoint;
+    for (var i = 0; i < points.length; ++i) {
+        const x1 = points[i]._values[0];
+        const y1 = points[i]._values[1];
+        const xP = x0 - x1;
+        const yP = y0 - y1;
+        const n = xP * xP + yP * yP;
+        if (n < squaredEpsilon) {
+            closestPoint = points[i];
```
peppsac, line 61: ~Do you want to return the indice `i` as the function name suggests or the point?~

Edit: fixed

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * filters the features that are under the coordinate
+     *
+     * @param      {Coordinates}  coordinate  the coordinate for the filter condition
+     * @param      {Features}  features  features to filter
+     * @param      {number}  epsilon  tolerance around the coordinate (in coordinate's unit)
+     * @return     {array}  array of filters features
+     */
+    getFeaturesAtCoordinate(coordinate, features, epsilon = 0.1) {
+        const result = { points: [], lines: [], polygons: [] };
```
peppsac: I don't see the point of returning the features grouped by geometry types?
Why don't you simply return a list of matching features?

(and please do not force push in the middle of a review, it makes it hard to have a discussion because previous comments are hidden)

peppsac: And `getXXX` name scheme is reserved for getter imho. `extract`? `filter`?

peppsac: Hmmm... You're not looping over `features` so I'm really wondering: what is the intent of this fonction? 
Can a user pass an array of features?
Or are only you extracting geometry/geometries that match a given coordinate from a **single** feature?

peppsac (changes requested):

peppsac:
I forgot one point in my review: now that we have a automated framework in place, it would be a good idea to test functions like the ones you introduce in `FeatureUtils`

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: > I don't see the point of returning the features grouped by geometry types?

Ok, for a list of typed features

> You're not looping over features

I loop over all features through, features.geometries have coordinates and features.features have properties [see here](https://github.com/iTowns/itowns/pull/470/files/9d7cb65a3791c5a3d7889563079740772cfddb06#diff-489ba48df1c76dca6a092ba77ffc0b64R133)

> And getXXX name scheme is reserved for getter imho. extract? filter?

ok for filter

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,146 @@
...
+                if (feature.extent && !feature.extent.isPointInside(coordinate, epsilon)) {
+                    continue;
+                }
+                /* eslint-disable guard-for-in */
+                for (const id in feature.featureVertices) {
+                    const polygon = feature.featureVertices[id];
+                    if (polygon.extent && !polygon.extent.isPointInside(coordinate, epsilon)) {
+                        continue;
+                    }
+                    const properties = features.features[id].properties.properties;
```
peppsac: Where does this double `properties.properties` come from?

gchoqueux: I don't sure understand your question
It comes from your PR on the features

gchoqueux:
@peppsac  updated

src/Renderer/ThreeExtended/FeaturesUtils.js
peppsac: The first `properties` comes from the PR yes (https://github.com/iTowns/itowns/blob/master/src/Renderer/ThreeExtended/GeoJSON2Feature.js#L164): it's where all custom properties are copied from the geojson.

But I'm afraid that the second comes from your data - like this https://github.com/iTowns/itowns/pull/470/files#diff-8966359153adaa7d05280e380a2a2dbbR13

Does this PR still work if you rename 'properties' in your example geojson to 'orange' for instance?

test/featureUtils_unit_test.js
```diff
@@ -0,0 +1,48 @@
+/* global describe, assert, it, itowns */
+require('./itowns-testing.js');
+var example = require('../examples/loadGeojson.js');
```
peppsac: Why don't you directly load `../examples/geojson/simple.geojson` from here?

And `itowns-testing.js` is definetely not adequate for unit testing. You should directly import the required modules (see `coordinate_unit_test.js`).

But besides this: 👍 for adding tests!

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: `properties` is in [geojson's specification](https://gist.github.com/virtualandy/1233401#file-geojson-spec-1-1-L231)

test/featureUtils_unit_test.js
gchoqueux: updated unit test

src/Renderer/ThreeExtended/FeaturesUtils.js
peppsac: Oohh I didn't know that :)

In this case, the parser should be fixed, no need to create this properties.properties oddity.
It'll probably wait another PR though....

peppsac: But since we have 2 PR in progress (this one + #519) using this attribute it would be nice to fix this before merging them.

test/featureUtils_unit_test.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+        res({
+            json: () => Promise.resolve(JSON.parse(content)),
+            ok: true,
+        });
+    });
+    return pr;
+};
+
+const promiseGeoJson = Fetcher.json('./geojson/simple.geojson').then(function _(result) {
```
peppsac: No need to use Fetcher (and mock fetch): juste use `fs.readFile` + `JSON.parse` directly.

peppsac: Or even simpler : `const json = require('./geojson/simple.geojson');` should work...

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: I'll add a PR

test/featureUtils_unit_test.js
gchoqueux: update

examples/globe_vector.html
```diff
@@ -29,14 +29,44 @@
                     display: none;
                 }
             }
+            .couponcode:hover .coupontooltip {
```
autra: Where do these "coupon" comes from?

test/featureUtils_unit_test.js
```diff
@@ -0,0 +1,44 @@
...
+
+const assert = require('assert');
+const geojson = require('../examples/geojson/simple.geojson.json');
+
+const feature = GeoJSON2Feature.parse('EPSG:4326', geojson, undefined, { buildExtent: true });
+
+describe('FeaturesUtils', function () {
+    it('should correctly parse geojson', (done) => {
+        assert.equal(feature.geometries.length, 3);
+        done();
```
autra: (nits: you don't need to call done or even have it in the parameter list in synchrone tests)

src/Renderer/ThreeExtended/FeaturesUtils.js
gchoqueux: see #527

examples/globe_vector.html
gchoqueux: They are used to display the feature’s properties

autra: Yes I know :-) I was asking about the meaning of "coupon": it doesn't seem to make sense in our context. These classes should be renamed.

gchoqueux:
@peppsac  PR rebased

gchoqueux:
reviewers, this PR can be merged?

examples/globe_vector.html
```diff
@@ -47,17 +74,99 @@
...
+                            tooltip.style.top = e.pageY + 'px';
+                        }
+                    }
+
+                    function pickPosition(e) {
+                        buildToolTip(globeView.controls.pickGeoPosition(e.clientX, e.clientY), e);
+                    }
+                    // precision in pixels
+                    var precisionPx = 5;
+                    function buildToolTip(geoCoord, e) {
```
peppsac: Could you move this code to a utility `examples/tooltip.js` file? It would help readability of this example and allo code sharing with other examples.

src/Renderer/ThreeExtended/FeaturesUtils.js
```diff
@@ -0,0 +1,145 @@
...
+            && (x < (xj - xi) * (y - yi) / (yj - yi) + xi);
+        if (isIntersect) {
+            inside = !inside;
+        }
+    }
+
+    return inside;
+}
+
+function filterFeatureUnderCoordinate(coordinate, type, coordinates, epsilon, properties, result) {
```
peppsac: `properties` and `result` are only useful because this function modifies directly `result`.
I think it would be better to modify the signature to:
`isFeatureUnderCoordinate(coordinate, featureType, featureCoords, epsilon)`
and make it return `true` or `false`.

Then the caller becomes:
```js
for (const id in feature.featureVertices) {
    const polygon = feature.featureVertices[id];
    if (polygon.extent && !polygon.extent.isPointInside(coordinate, epsilon)) {
        continue;
    }
    const coordinates = feature.coordinates.slice(polygon.offset, polygon.offset + polygon.count);
    if (isFeatureUnderCoordinate(coordinate, feature.type, coordinates, epsilon) {
    	result.push({
    		coordinates,
    		type: feature.type,
    		properties: features.features[id].properties,
    	});
    }
}

peppsac (commented):
Mostly good. 
Only 2 comments to address and we're good to go.

gchoqueux:
@peppsac updated

peppsac (approved):
LGTM 👍

PR #471, opened 2017-09-11
fix (scheduler): improve layers with relative url support

gchoqueux:
## Description

Added an exception when `Url` was instantiated
## Motivation and Context

Prior to this commit using a relative url (like ./foo.gpx) would produce 
an error.
This is fixed by this commit by using the optional 'base' parameter of URL
 constructor that specifies the base address if the first parameter is a 
relative URL.
(see https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -122,8 +122,7 @@ Scheduler.prototype.execute = function execute(command) {
 
     // parse host
     const layer = command.layer;
-
-    const host = layer.url ? new URL(layer.url).host : undefined;
+    const host = layer.url ? new URL(layer.url, document.location).host : undefined;
 
     command.promise = new Promise((resolve, reject) => {
         command.resolve = resolve;


src/Core/Scheduler/Scheduler.js
```diff
@@ -123,7 +123,13 @@ Scheduler.prototype.execute = function execute(command) {
     // parse host
     const layer = command.layer;
 
-    const host = layer.url ? new URL(layer.url).host : undefined;
+    let host;
+    try {
+        host = layer.url ? new URL(layer.url).host : undefined;
+    } catch (ex) {
+        host = undefined;
```
autra: We should at least test for the type of exception to only catch the one described in [the doc](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and let other type of expression go up. 

Doing so is a good reflexe to have, because here we are catching everything, but other errors can happen that are unrelated to local file url problem (ex what if layer is null?). In these cases, we need to report these errors instead of assuming host = undefined.

autra (changes requested):
`new URL` can only fail with an invalid url. I feel this is a configuration problem and not a code problem. Can you give me an example of failing url to clarify what you intend to fix? Thanks.

src/Core/Scheduler/Scheduler.js
gchoqueux: >  for example with local file paths

promises.push(globeView.addLayer({
    type: 'color',
    url: './ULTRA2009.gpx',
    protocol: 'rasterizer',
    id: 'Gpx',
}));

autra: I think `url` should always be a valid url. Either the user gives a valid string, or a valid url object (for instance `new URL('foo.gpx', window.location.origin + window.location.pathname)`) in `url`.

If you really want to allow strings like `./foo.gpx`, we could test if the string contains `://` and if not, construct the URL ourselves with the correct basepath, like just above. The idea is to avoid catching all exceptions here.

peppsac: Maybe we could simply always do: `new URL(layer.url, document.location)`?

autra: Indeed!

gchoqueux: thanks for your advice @autra @peppsac 
PR updated

src/Core/Scheduler/Scheduler.js
```diff
@@ -119,11 +119,20 @@ Scheduler.prototype.runCommand = function runCommand(command, queue, executingCo
 Scheduler.prototype.execute = function execute(command) {
     // TODO: check for mandatory commands fields
 
-
     // parse host
     const layer = command.layer;
+    let host;
 
-    const host = layer.url ? new URL(layer.url).host : undefined;
+    if (layer.url) {
```
autra: no actually, what we meant is 
```js
const host = layer.url ? new URL(layer.url, document.location).host : undefined;
```
This should work in all cases.

src/Core/Scheduler/Scheduler.js
```diff
@@ -119,11 +119,20 @@ Scheduler.prototype.runCommand = function runCommand(command, queue, executingCo
...
+        if (layer.url.includes('://')) {
+            host = new URL(layer.url).host;
+        } else {
+            // local file
```
autra: the problem is not with local files (`file://local/path` works fine with the old code), but with relative path (without scheme). Can you update your commit message reflecting that please? 

Btw in this message, let's not use the word "problem", when there's a fix, it's obvious there was a problem ;-)  I would leave out the `:` too and the second `fix`, in order to make it readable aloud. Maybe simply `Fix getting host with relative url in scheduler` will do?

peppsac: IMHO the `fix:` prefix is useful (for example see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type)

But you could reword the part after `:` to avoid the repetition.
And add a description in your commit message, it's more important than the PR text (hint: github could close tomorrow and all PR text will be lost but the git commit message would remain).

```
fix (scheduler): improve layers with relative url support

Prior to this commit using a relative url (like ./foo.gpx) would produce 
an error.
This is fixed by this commit by using the optional 'base' parameter of URL
 constructor that specifies the base address if the first parameter is a 
relative URL.
(see https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)

```

gchoqueux: updated

src/Core/Scheduler/Scheduler.js
gchoqueux: updated

autra (approved):

autra:
LGTM, I'm waiting for travis, then I'll merge.

PR #472, opened 2017-09-11
fix (Depth) : correct calculation of depth

### LINKED ISSUE #411: White screen when zooming

gchoqueux:
## Description

Use of depth buffers instead of calculating Z in the shader TileFS.glsl
this PR doesn't resolve #411 
## Motivation and Context

Before `DEPTH_MODE` returns Z value between 0 and camera.far, as `packDepthToRGBA` packs values between 0.0 and 1.0, Z was divided by a big constant.
It is therefore more logical to use the depth (between 0.0 and 1.0) and convert it out of the shader

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -389,7 +389,6 @@ const screen = new THREE.Vector2();
 const pickWorldPosition = new THREE.Vector3();
 const ray = new THREE.Ray();
 const direction = new THREE.Vector3();
-const depthRGBA = new THREE.Vector4();
 GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse) {
     const dim = this.mainLoop.gfxEngine.getWindowSize();
     mouse = mouse || dim.clone().multiplyScalar(0.5);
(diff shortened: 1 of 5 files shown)


peppsac:
`screenCoordsToNodeId` uses the `factor` argument so this change will break it.

gchoqueux:
@peppsac updated

src/Core/Prefab/GlobeView.js
```diff
@@ -427,13 +428,25 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
     direction.applyMatrix4(matrix);
     direction.sub(ray.origin);
 
-    var angle = direction.angleTo(ray.direction);
+    const angle = direction.angleTo(ray.direction);
 
     depthRGBA.fromArray(buffer).divideScalar(255.0);
 
-    var depth = unpack1K(depthRGBA, 100000000.0) / Math.cos(angle);
+    let orthoZ = 0;
```
peppsac: Maybe you could move this function to `c3DEngine`, something like: `c3DEngine.depthBufferValueToZ(depthRGBA)`?

gchoqueux: done

peppsac (approved):

PR #474, opened 2017-09-11
feature (GlobeView) : use cache depth buffer

gchoqueux:
## Description

Add cache depth buffer, this buffer is updated when all datas is loaded and no camera movement

## Motivation and Context

This allows you to read depth information without rendering again when the viewer is paused.
and thus to read the geographical positions without new rendering (`GlobeControls.readGeoPosition` )

### DIFF
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -1,7 +1,7 @@
 import { EventDispatcher } from 'three';
 
-const RENDERING_PAUSED = 0;
-const RENDERING_SCHEDULED = 1;
+export const RENDERING_PAUSED = 0;
+export const RENDERING_SCHEDULED = 1;
 
 function MainLoop(scheduler, engine) {
     this.renderingState = RENDERING_PAUSED;
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 
 import View from '../View';
+import { RENDERING_PAUSED } from '../MainLoop';
 import { COLOR_LAYERS_ORDER_CHANGED } from '../../Renderer/ColorLayersOrdering';
 import RendererConstant from '../../Renderer/RendererConstant';
 import GlobeControls from '../../Renderer/ThreeExtended/GlobeControls';
@@ -239,8 +240,17 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     }
 
     this._renderState = RendererConstant.FINAL;
+    this._fullSizeDepthBuffer = null;
+
     const renderer = this.mainLoop.gfxEngine.renderer;
     this.preRender = () => {
+        // WARNING, if the prerender is re-defined by the user,
+        // These mechanisms no longer work
+        // TODO: need to fix it
+        if (this._fullSizeDepthBuffer != null) {
+            // clean depth buffer
+            this._fullSizeDepthBuffer = null;
+        }
         const v = new THREE.Vector3();
         v.setFromMatrixPosition(wgs84TileLayer.object3d.matrixWorld);
         var len = v.distanceTo(this.camera.camera3D.position);
@@ -384,30 +394,43 @@ GlobeView.prototype.screenCoordsToNodeId = function screenCoordsToNodeId(mouse) 
     return Math.round(unpack);
 };
 
+GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, height) {
+    const g = this.mainLoop.gfxEngine;
+    const previousRenderState = this._renderState;
+    this.changeRenderState(RendererConstant.DEPTH);
+    const buffer = g.renderViewTobuffer(this, g.fullSizeRenderTarget, x, y, width, height);
+    this.changeRenderState(previousRenderState);
+    return buffer;
+};
+
 const matrix = new THREE.Matrix4();
 const screen = new THREE.Vector2();
 const pickWorldPosition = new THREE.Vector3();
 const ray = new THREE.Ray();
 const direction = new THREE.Vector3();
 GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse) {
-    const dim = this.mainLoop.gfxEngine.getWindowSize();
+    const l = this.mainLoop;
+    const viewPaused = l.scheduler.commandsWaitingExecutionCount() == 0 && l.renderingState == RENDERING_PAUSED;
+    const g = l.gfxEngine;
+    const dim = g.getWindowSize();
+    const camera = this.camera.camera3D;
+
     mouse = mouse || dim.clone().multiplyScalar(0.5);
+    mouse.x = Math.floor(mouse.x);
+    mouse.y = Math.floor(mouse.y);
 
-    var camera = this.camera.camera3D;
+    const prev = camera.layers.mask;
+    camera.layers.mask = 1 << this.wgs84TileLayer.threejsLayer;
 
-    // Prepare state
-    const prev = this.camera.camera3D.layers.mask;
-    this.camera.camera3D.layers.mask = 1 << this.wgs84TileLayer.threejsLayer;
-
-    const previousRenderState = this._renderState;
-    this.changeRenderState(RendererConstant.DEPTH);
-
-    // Render to buffer
-    var buffer = this.mainLoop.gfxEngine.renderViewTobuffer(
-        this,
-        this.mainLoop.gfxEngine.fullSizeRenderTarget,
-        mouse.x, dim.y - mouse.y,
-        1, 1);
+    // Render/Read to buffer
+    let buffer;
+    if (viewPaused) {
+        this._fullSizeDepthBuffer = this._fullSizeDepthBuffer || this.readDepthBuffer(0, 0, dim.x, dim.y);
+        const id = ((dim.y - mouse.y - 1) * dim.x + mouse.x) * 4;
+        buffer = this._fullSizeDepthBuffer.slice(id, id + 4);
+    } else {
+        buffer = this.readDepthBuffer(mouse.x, dim.y - mouse.y - 1, 1, 1);
+    }
 
     screen.x = (mouse.x / dim.x) * 2 - 1;
     screen.y = -(mouse.y / dim.y) * 2 + 1;
@@ -427,13 +450,11 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
     direction.sub(ray.origin);
 
     const angle = direction.angleTo(ray.direction);
-    const orthoZ = this.mainLoop.gfxEngine.depthBufferRGBAValueToOrthoZ(buffer, camera);
+    const orthoZ = g.depthBufferRGBAValueToOrthoZ(buffer, camera);
     const length = orthoZ / Math.cos(angle);
 
     pickWorldPosition.addVectors(camera.position, ray.direction.setLength(length));
 
-    // Restore initial state
-    this.changeRenderState(previousRenderState);
     camera.layers.mask = prev;
 
     if (pickWorldPosition.length() > 10000000)
--- a/src/Core/Prefab/PlanarView.js
+++ b/src/Core/Prefab/PlanarView.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 
 import View from '../View';
+import { RENDERING_PAUSED } from '../MainLoop';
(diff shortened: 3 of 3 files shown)


gchoqueux:
reviewers?

autra:
> This allows you to read depth information without rendering again when the viewer is paused.

Did we have a perf problem with this extra rendering before?

gchoqueux:
On non-powerful machines rendering with every mouse movement is problematic in terms of performance. 
Moreover they are render useless

gchoqueux:
several `readRenderTargetPixels` is also not very powerful...
reviewing?

src/Core/Prefab/GlobeView.js
```diff
@@ -240,8 +241,32 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
...
+    this.wgs84TileLayer = wgs84TileLayer;
+
+    const fn = () => {
+        this.mainLoop.removeEventListener('command-queue-empty', fn);
+        this.dispatchEvent({ type: GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED });
+    };
+
+    this.mainLoop.addEventListener('command-queue-empty', fn);
+
+    this.callbacksOnAfterRender = new Map();
```
peppsac: I'm concerned with this change.

IMHO, either this is a good feature and you should rework other parts of iTowns to support this.
Or this is more a hack and we shouldn't merge this.

Maybe my suggestion https://github.com/iTowns/itowns/pull/476#issuecomment-330150491 could help here: remove the current callback and replace with attachable `FrameRequester`.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1778,4 +1797,8 @@ GlobeControls.prototype.reset = function reset() {
     this.updateCameraTransformation();
 };
 
+GlobeControls.prototype.getState = function getState() {
```
peppsac: Exposing `state` for 1 user `this.controls.getState() == -1` is meh.

Maybe you could expose something more meaningful (eg `function isControlIdle`) ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1762,6 +1764,23 @@ GlobeControls.prototype.pickGeoPosition = function pickGeoPosition(mouse, y) {
     return new Coordinates('EPSG:4978', pickedPosition).as('EPSG:4326');
 };
 
+
+GlobeControls.prototype.readGeoPosition = function readGeoPosition(mouse, y) {
```
peppsac: Why are you adding this function?

src/Core/Prefab/GlobeView.js
```diff
@@ -389,8 +411,13 @@ const screen = new THREE.Vector2();
 const pickWorldPosition = new THREE.Vector3();
 const ray = new THREE.Ray();
 const direction = new THREE.Vector3();
-GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse) {
-    const dim = this.mainLoop.gfxEngine.getWindowSize();
+GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse, readFullBuffer) {
```
peppsac: What's the usage of this `readFullBuffer` parameter?

src/Core/Prefab/GlobeView.js
```diff
@@ -399,15 +426,24 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
...
-    this.changeRenderState(RendererConstant.DEPTH);
-
-    // Render to buffer
-    var buffer = this.mainLoop.gfxEngine.renderViewTobuffer(
-        this,
-        this.mainLoop.gfxEngine.fullSizeRenderTarget,
-        mouse.x, dim.y - mouse.y,
-        1, 1);
+    let buffer;
+    if (viewPaused) {
```
peppsac: Can't you simplify this code? Or extract this to a utility method maybe?

peppsac (changes requested):
Overall this PR seems a bit complicated, just to add a cache.

Did you consider making this cache feature work for the planar mode as well?

src/Core/Prefab/GlobeView.js
gchoqueux: removed

src/Core/Prefab/GlobeView.js
gchoqueux: removed

src/Core/Prefab/GlobeView.js
gchoqueux: ok

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: removed

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: ok

gchoqueux:
@peppsac 
PR cleaned and updated 
Do you want cache in planar mode? if you want add this feature

src/Core/Prefab/GlobeView.js
peppsac: Thanks, it's much more readable.

src/Core/Prefab/GlobeView.js
```diff
@@ -399,15 +418,14 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
...
+        fullSizeDepthBuffer = fullSizeDepthBuffer || this.getDepthBuffer(0, 0, dim.x, dim.y);
+        const id = ((dim.y - mouse.y) * dim.x + mouse.x) * 4;
```
peppsac: Are you sure about this?
If `mouse.y` is `0`, you're going to read `dim.y * dim.x + mouse.x` which is probably past the end of the buffer.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1778,4 +1780,8 @@ GlobeControls.prototype.reset = function reset() {
     this.updateCameraTransformation();
 };
 
+GlobeControls.prototype.isControlIdle = function isControlIdle() {
```
peppsac: `isIdle` is enough IMHO.

peppsac:
What do you mean by: `Do you want cache in planar mode? if you want add this feature` ?

src/Core/Prefab/GlobeView.js
gchoqueux: yes it's normal , the y is reversed in buffer
see in master, it's reversed : https://github.com/iTowns/itowns/pull/474/files/193b6928ce34b331577af08120514e066359ff5a#diff-caf5227be4f9e2f283d82df32a678e4bL408

peppsac: I'm not asking about the reversed y axis.

Your buffer size is `dim.x * dim.y * 4`
If mouse is: `{ x: 10, y: 0 }` you're going to start reading `(dim.y * dim.x + 10) * 4` which is out of bounds (ie: I guess the correct formula is `((dim.y - mouse.y - 1) * dim.x + mouse.x) * 4`)

gchoqueux:
updated

Do you want this feature in planar mode?

peppsac:
Yes, I think it would make sense to have feature parity in planar and globe mode.

src/Core/Prefab/GlobeView.js
gchoqueux: yes, thanks

gchoqueux:
updated, add in planar mode

src/Core/Prefab/GlobeView.js
```diff
@@ -71,6 +71,7 @@ export const GLOBE_VIEW_EVENTS = {
     COLOR_LAYERS_ORDER_CHANGED,
 };
 
+let fullSizeDepthBuffer = null;
```
peppsac: You should store this in the GlobeView instance (prefixed with a `_` to mark it as private) because a user can create several GlobeViews (the same applies to PlaneView).

src/Core/Prefab/GlobeView.js
```diff
@@ -384,13 +385,24 @@ GlobeView.prototype.screenCoordsToNodeId = function screenCoordsToNodeId(mouse)
     return Math.round(unpack);
 };
 
+GlobeView.prototype.getDepthBuffer = function getDepthBuffer(x, y, width, height) {
```
peppsac: I'd prefer `readDepthBuffer` since `getXXX` is reserved for trivial getter functions.

src/Core/Prefab/GlobeView.js
```diff
@@ -384,13 +385,24 @@ GlobeView.prototype.screenCoordsToNodeId = function screenCoordsToNodeId(mouse)
...
+};
+
 const matrix = new THREE.Matrix4();
 const screen = new THREE.Vector2();
 const pickWorldPosition = new THREE.Vector3();
 const ray = new THREE.Ray();
 const direction = new THREE.Vector3();
 GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse) {
-    const dim = this.mainLoop.gfxEngine.getWindowSize();
+    const viewPaused = this.mainLoop.scheduler.commandsWaitingExecutionCount() == 0 && (!this.controls || this.controls.isIdle());
```
peppsac: Couldn't you replace `the controls.isIdle` test with a something testing if the mainloop is halted (ie : mainLoop.renderingState == RENDERING_SCHEDULED)?

This way you wouldn't have to add a `isIdle` function to all controls.

src/Core/Prefab/GlobeView.js
```diff
@@ -399,15 +411,16 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
...
-        1, 1);
+    // Render/Read to buffer
+    let buffer;
+    if (viewPaused) {
+        fullSizeDepthBuffer = fullSizeDepthBuffer || this.getDepthBuffer(0, 0, dim.x, dim.y);
+        const id = ((dim.y - mouse.y - 1) * dim.x + mouse.x) * 4;
+        buffer = fullSizeDepthBuffer.slice(id, id + 4);
+    } else {
+        fullSizeDepthBuffer = null;
+        buffer = this.getDepthBuffer(mouse.x, dim.y - mouse.y, 1, 1);
```
peppsac: `dim.y - mouse.y - 1`?

peppsac (commented):
Thanks for adapting the code for PlanarView as well.

gchoqueux:
@peppsac  thanks for comments
updated

peppsac (approved):
Looks good!

gchoqueux:
there's little problem, i need to fix it

gchoqueux:
@peppsac 
finally this test  to known if the view is idled, doesn't work :
`const viewPaused = l.scheduler.commandsWaitingExecutionCount() == 0 && l.renderingState == RENDERING_PAUSED;`

I rollback and and I add that we dontt have to code `isIdle`
`const viewPaused = l.scheduler.commandsWaitingExecutionCount() == 0 && (!this.controls || (this.controls.isIdle && this.controls.isIdle()));`

peppsac:
>  finally this test to known if the view is idled, doesn't work [...]

What is the problem exactly?

gchoqueux:
The problem is that we read in the cache while it's obsolete

peppsac:
> The problem is we read in the cache while it's obsolete

Sure, but this is the end result. Did did you investigate why? It should work so understanding the underlying issue could be valuable (this could expose a bug in itowns or your PR)

gchoqueux:
@peppsac I fix with `RENDERING_PAUSED`

src/Core/MainLoop.js
```diff
@@ -142,7 +142,10 @@ MainLoop.prototype._renderView = function _renderView(view) {
         // use default rendering method
         this.gfxEngine.renderView(view);
     }
-
+    if (view._fullSizeDepthBuffer != null) {
```
peppsac: Why didn't you put this code in `GlobeView.preRender` ?

gchoqueux: because it will be erased if the user redefines `preRender`

peppsac: We already have this problem because we already do some stuff in Globe.preRender.

You should put depth buffer cache invalidation in preRender *and* add a todo/issue to state that this is a problem that should get fixed.

gchoqueux: updated

peppsac:
Thanks :)

PR #475, opened 2017-09-11
fix (subdivision) : no stop subdivision in case error loading

gchoqueux:
## Description

Don't stop subdivision tile in case error loading
## Motivation and Context

If some wmts, wms datas don't load, the tiles are no longer subdivided and block the display of the other downloaded data.

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -5,6 +5,9 @@ import { ImageryLayers } from '../Core/Layer/Layer';
 import { CancelledCommandException } from '../Core/Scheduler/Scheduler';
 import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from '../Core/Scheduler/Providers/OGCWebServiceHelper';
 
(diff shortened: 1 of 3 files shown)


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -243,10 +243,13 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                     // eslint-disable-next-line no-console
                     console.warn(`Imagery texture update error for ${node}: ${err}`);
                 }
-                node.layerUpdateState[layer.id].failure(Date.now());
-                window.setTimeout(() => {
-                    context.view.notifyChange(false, node);
-                }, node.layerUpdateState[layer.id].secondsUntilNextTry() * 1000);
+                const definitiveError = node.layerUpdateState[layer.id].errorCount > 4;
```
autra: Let's introduce a `MAX_RETRY` (or something similar) constant.

autra (approved):
With the one nits corrected, LGTM. Thanks!

PR #476, opened 2017-09-13
fix: make sure camera's matrix is only updated by iTowns

peppsac:
## Description
We must make sure that camera.matrixWorld is not modified in between mainloop.update
and mainloop.render otherwise we'll have visual inconsistencies.

e.g: there were holes in the planar view when rotating quickly the camera because of
this issue (three.js calls camera.updateWorldMatrix at the beginning of its rendering
code).
camera

### DIFF
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -95,6 +95,16 @@ MainLoop.prototype._step = function _step(view, timestamp) {
 
     view.camera.update(dim.x, dim.y);
 
+    // Disable camera's matrix auto update to make sure the camera's
+    // world matrix is never updated mid-update.
+    // Otherwise inconsistencies can appear because object visibility
+    // testing and object drawing could be performed using different
+    // camera matrixWorld.
+    // Note: this is required at least because WEBGLRenderer calls
+    // camera.updateMatrixWorld()
+    const oldAutoUpdate = view.camera.camera3D.matrixAutoUpdate;
+    view.camera.camera3D.matrixAutoUpdate = false;
+
     // update data-structure
     this._update(view, updateSources, dt);
 
@@ -117,6 +127,8 @@ MainLoop.prototype._step = function _step(view, timestamp) {
     if (__DEBUG__) {
         document.title = document.title.substr(0, document.title.length - 2);
     }
+
+    view.camera.camera3D.matrixAutoUpdate = oldAutoUpdate;
 };
 
 MainLoop.prototype._renderView = function _renderView(view) {


gchoqueux:
Why there isn't the same phenomenon with the globe view?

peppsac:
> Why there isn't the same phenomenon with the globe view?

I believe that it's less visible because its a sphere (so no hard edges) but I don't see why it wouldn't happen on the globe as well.

gchoqueux:
I can't reproduce it on the globe even when I'm very close to the ground (which is a plane surface)
while I easily reproduce the bug on the planar

peppsac:
Looked a bit more into the issue:
- planar controls update is called by the Mainloop (as a FrameRequester), after camera.update call but before view.render so if planar controls modifies the camera's matrix it'll cause some issue if three.js updates matrixWorld
- globecontrols  update is called at various place (mouse move event, animation player update, etc): as a result the call flow is harder to predict - but as it's only updated on events it cannot interrupt the update/render flow which would explain why you can't repro the issue on the globe.

autra (approved):
I'm not a big fan of monkey-patching, but I guess we don't have much choice here. Thanks for the patch!

@gchoqueux do you have any more remarks on this?

gchoqueux (commented):
I do not agree with this patch. 
As it removes an important function `updateMatrixWorld` from the `THREE.camera` that can be called through the `resize` of the camera itowns. I think that if there is this bug is that there is something else that is badly done.

I don't agree that there is the same bug on the globe, I would like to have a procedure to reproduce it.

Does `matrixAutoUpdate` have any impact on the camera?

peppsac:
@gchoqueux `resize` is always called in `update`.
And `update` calls the real `Camera.updateMatrixWorld` method.

So this patch doesn't remove "an important function". It simply change the process from "camera.matrixWorld" can be updated at random time to "camera.matrixWorld is only updated once"

```
Camera.prototype.update = function update(width, height) {
    resize(this, width, height);
    // update matrix
    THREE.Camera.prototype.updateMatrixWorld.call(this.camera3D);
```

Regarding the (non)globe issue: https://github.com/iTowns/itowns/pull/476#issuecomment-329491400 details why you can't reproduce this issue.

It can be summarized as: `GlobeControls` isn't properly integrated in `iTowns` so instead of using the `FrameRequester` method and update the camera at known time, it instead does things when it wants (on events or setTimeout). 
So basically this non-integration bug prevent the camera inconsistencies bug.

gchoqueux:
I'll deepen this PR

peppsac:
btw, this PR is more a workaround to fix a very visible issue.

A more long term solution would require to:
- make sure all controls use the FrameRequester mechanism
- define several attach points for FrameRequester, e.g:
```js
addFrameRequester(foo, FRAME_ANCHOR.CONTROLS);
addFrameRequester(foo, FRAME_ANCHOR.PRE_VIEW_UPDATE);
addFrameRequester(foo, FRAME_ANCHOR.POST_VIEW_UPDATE);
addFrameRequester(foo, FRAME_ANCHOR.PRE_VIEW_RENDER);
addFrameRequester(foo, FRAME_ANCHOR.POST_VIEW_RENDER);
```
and each anchor point would be documented with explicit contract (ie: you can only modify camera's matrix in `FRAME_ANCHOR.CONTROLS`).

gchoqueux (changes requested):
As I said in this commentary
https://github.com/iTowns/itowns/pull/476#pullrequestreview-63020410

I think it's possible to fix this bug with `matrixAutoUpdate`

I don't agree with this type of modification that modifies the framework of THREE.js while there is probably a suitable solution in the framework

peppsac:
Modified the workaround to use `matrixAutoUpdate` instead of patching `updateMatrixWorld`.

gchoqueux (approved):
Thanks, it's better

PR #479, opened 2017-09-15
updated THREE to r87

nosy-b:
## Description
Updated THREE to r87

## Motivation and Context
Up to date as the 15th of September to the last THREE JS revision
 No problem running all examples.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "monotone-convex-hull-2d": "^1.0.1",
     "proj4": "^2.4.3",
     "text-encoding-utf-8": "^1.0.1",
-    "three": "^0.86.0",
+    "three": "^0.87.0",
     "three.meshline": "^1.0.3",
     "togeojson": "^0.16.0",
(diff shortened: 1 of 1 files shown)


autra:
@nosy-b did you check https://github.com/mrdoob/three.js/releases for incompatible change?

autra:
@Jeremy-Gaillard new version of three removed Gltf 1.0 support and renamed GLTF2Loader to GLTFLoader. Do we replace the loader we committed in itowns according to you?

nosy-b:
@autra Yes, didn't see anything worrying.

autra:
@nosy-b yep. Just the one thing about gltf loader, but I guess it could be done later.

autra (approved):

Jeremy-Gaillard:
@autra the current 3D Tiles specification still references glTF 1.0 and all the 3D Tiles samples (which are used in our example) still use it. For now, we can't replace the glTF Loader, we may even need to keep it for a while until 3D Tiles-related tools are updated.

PR #482, opened 2017-09-18
fix (3tiles): handle 'refine' parameter inheritance properly 

NikoSaul:
fix (3tiles): handle 'refine' parameter inheritance properly 
A node must be inherited of the parent's value unless explicitly specified 
See issue #185


peppsac (approved):

peppsac:
Thanks!

PR #483, opened 2017-09-20
refactor: add Extent.intersects method and use three.js wording

peppsac:
The old 'intersect' method is renamed 'intersects' (to match three.js wording
for Box2 for instance: 'intersectsBox').

A new 'intersection' method is added and return the intersection between 2 extents.

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -296,13 +296,30 @@ Extent.prototype.offsetScale = function offsetScale(bbox) {
  * @param {type} bbox
  * @returns {Boolean}
  */
-Extent.prototype.intersect = function intersect(bbox) {
+Extent.prototype.intersectsExtent = function intersectsExtent(bbox) {
     const other = bbox.as(this.crs());
     return !(this.west() >= other.east(this._internalStorageUnit) ||
              this.east() <= other.west(this._internalStorageUnit) ||
              this.south() >= other.north(this._internalStorageUnit) ||
              this.north() <= other.south(this._internalStorageUnit));
 };
+
+/**
+ * @documentation: Return the intersection of this extent with another one
+ * @param {type} other
+ * @returns {Boolean}
+ */
+Extent.prototype.intersect = function intersect(other) {
+    if (!this.intersectsExtent(other)) {
+        return new Extent(this.crs(), 0, 0, 0, 0);
+    }
+    return new Extent(this.crs(),
+        Math.max(this.west(), other.west(this._internalStorageUnit)),
+        Math.min(this.east(), other.east(this._internalStorageUnit)),
+        Math.max(this.south(), other.south(this._internalStorageUnit)),
+        Math.min(this.north(), other.north(this._internalStorageUnit)));
+};
+
 
 Extent.prototype.set = function set(west, east, south, north) {
     this._values[CARDINAL.WEST] = west;
--- a/src/Core/Scheduler/Providers/Raster_Provider.js
+++ b/src/Core/Scheduler/Providers/Raster_Provider.js
@@ -137,7 +137,7 @@ export default {
         }
     },
     tileInsideLimit(tile, layer) {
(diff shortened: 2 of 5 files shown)


src/Core/Geographic/Extent.js
```diff
@@ -296,14 +296,31 @@ Extent.prototype.offsetScale = function offsetScale(bbox) {
...
              this.south() >= other.north(this._internalStorageUnit) ||
              this.north() <= other.south(this._internalStorageUnit));
 };
 
+/**
+ * @documentation: Return the intersection of this extent with another one
+ * @param {type} other
+ * @returns {Boolean}
+ */
+Extent.prototype.intersect = function intersect(other) {
```
autra, line 312: Is this the THREE.js nomenclature too? If not, I suggest to find another name to avoid the confusion with intersects. `getIntersection`?

src/Core/Geographic/Extent.js
```diff
@@ -296,14 +296,31 @@ Extent.prototype.offsetScale = function offsetScale(bbox) {
...
+    if (!this.intersects(other)) {
+        return new Extent(this.crs(), 0, 0, 0, 0);
```
autra: Do we have a `Extent.isEmpty` method? Or maybe we could return the same constant every time `EMPTY_EXTENT`, allowing client code to test `extent === EMPTY_EXTENT`? Feel free to skip this if you think nobody'll use it though.

src/Core/Geographic/Extent.js
```diff
@@ -296,14 +296,31 @@ Extent.prototype.offsetScale = function offsetScale(bbox) {
...
+    }
+    return new Extent(this.crs(),
+        Math.max(this.west(), other.west(this._internalStorageUnit)),
+        Math.min(this.east(), other.east(this._internalStorageUnit)),
+        Math.max(this.south(), other.south(this._internalStorageUnit)),
+        Math.min(this.north(), other.north(this._internalStorageUnit)));
```
autra, line 320: I'm wondering if there are some funny CRS where the axis would go the other way ? (where west would be the max, and east the min)

autra (changes requested):
Just some little nits to address/questions to answer :-)

src/Core/Geographic/Extent.js
peppsac: Yes it is:
`Box3.intersect` => ` Returns the intersection of this and box [...]`
`Box3.intersectsXXX` => `  Determines whether or not this box intersects XXX`

src/Core/Geographic/Extent.js
peppsac: I have no strong feelings about this one... so I'd suggest to keep the code as is, and see if it becomes an issue?

src/Core/Geographic/Extent.js
autra: So our `Extent.intersects` should be `Extent.intersectsExtent`? Pro: clearer, less likely to be mixed with `intersect`. Cons: a bit silly maybe :-)

src/Core/Geographic/Extent.js
autra: Fine by me

src/Core/Geographic/Extent.js
peppsac: Ok, let's use `Extent.intersectsExtent`

peppsac: Done

src/Core/Geographic/Extent.js
autra: FWIW leaflet fixed a similar bug https://github.com/Leaflet/Leaflet/issues/4886, and this CRS https://epsg.io/9843-method 

These could be just a curiosity though. Also, this might not be the place here to support such CRS

gchoqueux (commented):
it's good for me
@autra

autra (approved):

autra:
Good for me too! Sorry I missed the update

PR #484, opened 2017-09-20
feature: add command.targetLevel support to WMS_Provider

peppsac:
If targetLevel is defined and < to node.level we request the texture for the parent that has level == targetLevel and then use this texture for node.

It's already supported for WMTS provider but it was missing for WMS.

This PR also modifies the `planar` example to use this feature.

### DIFF
--- a/src/Core/Scheduler/Providers/WMS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMS_Provider.js
@@ -5,7 +5,6 @@
  */
 
 
-import * as THREE from 'three';
 import Extent from '../../Geographic/Extent';
 import OGCWebServiceHelper from './OGCWebServiceHelper';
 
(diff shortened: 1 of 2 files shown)


gchoqueux:
> It's already supported for WMTS 

I'am not sure, no?
https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Providers/WMTS_Provider.js#L171

gchoqueux:
it's supported only in elevation WMTS?

peppsac:
> it's supported only in elevation WMTS?

Indeed. I thought we already added the missing code for the color layers as well.

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -74,26 +73,41 @@ WMS_Provider.prototype.tileInsideLimit = function tileInsideLimit(tile, layer) {
...
     if (!this.tileInsideLimit(tile, layer)) {
         return Promise.reject(`Tile '${tile}' is outside layer bbox ${layer.extent}`);
     }
     if (tile.material === null) {
         return Promise.resolve();
     }
 
-    const coords = tile.extent.as(layer.projection);
+    let extent = tile.extent;
+    if (targetLevel != undefined) {
```
autra: nits: can we invert this if-else ? (I find positive logic more readable)

autra (approved):

src/Core/Scheduler/Providers/WMS_Provider.js
peppsac: Done

PR #485, opened 2017-09-20
feature: allow layer custom 'can this tile get a better texture' mechanism

peppsac:
The default mechanism is based on a simple approach: if texture.level and
node.level don't match -> try to fetch a new texture.

This works well for providers that can compute meaningful level attributes,
but breaks if it's not possible.

This commit allows layer to have their own function.

(PR coming later with a Provider using this feature)

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -192,9 +192,16 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     }
 
     // does this tile needs a new texture?
-    if (!node.isColorLayerDownscaled(layer)) {
+    if (layer.canTileTextureBeImproved) {
+        // if the layer has a custom method -> use it
+        if (!layer.canTileTextureBeImproved(layer, node)) {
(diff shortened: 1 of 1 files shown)


gchoqueux:
A small example of use case, please

peppsac:
An example will come when I'll update my PR #451:
```js
    canTileTextureBeImproved(layer, tile) {
        const s = selectBestImageForExtent(layer.images, tile.extent);
        if (!s) {
            return false;
        }
        const mat = tile.material;
        const idx = mat.getLayerTextureOffset(layer.id);
        const currentTexture = mat.textures[1][idx];
        if (!currentTexture.file) {
            return true;
        }
        return currentTexture.file != s.image;
    },
```
Basically we pick the best image for a given tile but images' extent aren't aligned on tiles' extent.
So a given image might be the best match for tiles are different levels.

autra (approved):

autra:
Actually, before merging, can you document this possiblity in layer's documentation plz?

autra:
Ok nevermind it'll be only used internally.

PR #487, opened 2017-09-22
Feature geometry extruded

gliegard:
## Description
- Add style in the layer, to specify how to build mesh from feature.
- In style, add `extrude` polygon
- Fix coordinate conversion to make, for example, WFS in _EPSG:3946_ working on the globe.
- Update the example (WFS on planar), to have polygon extruded.
- Add an example : WFS extruded on the globe.

## Motivation and Context
This change allow the user to specify the way feature from WFS is treated. Feature polygons can be extruded. It works in planar mode and in the globe.

## Screenshots
screenshot-pr-feature-extruded


peppsac:
Thanks for the PR.

You might want to do a few things to ease the review:
- move each `fix: XXXX` commit to a separate PR. They'll get merged faster this way.
- move `feature (Coordinate) re-use normal vector already calculated` to a new PR as well
- keep only the WFS/extrusion related code in this PR.

PR #491, opened 2017-09-25
fix (WFS_Provider) : Log error in the console

gliegard:
## Description
Fix WFS Provider, error is not caught.

## Motivation and Context
When an error occurs in the Promise, we can't see it.
With this change, the error in caugth and we can see the error stack in the console.


peppsac:
Please do not omit commit messages (PR description are nice but aren't stored in the project .git history).

Regarding the change: *if* we want to log the Providers errors, we might want to do it for all providers.
So you should probably move the error handler in `Scheduler` (somewhere around https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Scheduler.js#L48).

gliegard:
Ok, I put the changes in the scheduler, and drop the changes on the WFS Provider

PR #493, opened 2017-09-25
fix (Scheduler) Error caught from the providers is logged in the console

gliegard:
## Description
Add log in the console, when error occurs in the Providers.

## Motivation and Context
When an error occurs in the Promise, we can't see it.
With this change, the error is logged in the console.

I didn't manage to reOpen #491

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -56,7 +56,7 @@ function _instanciateQueue() {
                 this.counters.executing--;
                 cmd.reject(err);
                 this.counters.failed++;
-                if (__DEBUG__) {
+                if (__DEBUG__ && this.counters.failed < 3) {
                     // eslint-disable-next-line no-console
                     console.error(err);
                 }
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -56,6 +56,10 @@ function _instanciateQueue() {
                 this.counters.executing--;
                 cmd.reject(err);
                 this.counters.failed++;
+                if (__DEBUG__) {
+                    // eslint-disable-next-line no-console
+                    console.error(err);
+                }
             });
         },
     };


src/Core/Scheduler/Scheduler.js
```diff
@@ -56,6 +56,8 @@ function _instanciateQueue() {
                 this.counters.executing--;
                 cmd.reject(err);
                 this.counters.failed++;
+                // eslint-disable-next-line no-console
+                console.error(err.stack);
```
autra: Please do directly `console.error(err)`. The stack will be displayed, but we'll also have the error message. Also, I'd prefer to have this LOGGER only in debug mode, as user can attach to errors with cmd.promise.

autra (changes requested):

src/Core/Scheduler/Scheduler.js
gliegard: ok, how can i know I'm in debug mode in the code ?

gliegard: it's done

gliegard:
I update the code, and rebase, it's ready to review !

src/Core/Scheduler/Scheduler.js
peppsac: > Also, I'd prefer to have this LOGGER only in debug mode, as user can attach to errors with cmd.promise.

The command are issued by itowns (eg. `FeatureProcessing`) so an external user can't attach to these promises.
What about logging the N first errors for a given layer?
(eventually adding an option to layers to opt-out from this behavior ?)

gliegard: Good idea to avoid being spam by errors !
I update the code. The new behavior is if you see two same errors, it means many errors occurs.

gliegard: > What about logging the N first errors for a given layer?

Ok, I show only two errors.

> (eventually adding an option to layers to opt-out from this behavior ?)

As errors are only shown in Debug mode, I think we don't have to do it now.

Is it ok for you ?

peppsac: Yep, thanks.

gliegard:
@itowns-reviewers

autra (approved):

autra:
LGTM thanks !

gliegard:
Thanks

PR #494, opened 2017-09-26
feature (Coordinate) store normal vector already computed

gliegard:
## Description
Store normal vector in the coordinate object when we already have computed it.

## Motivation and Context
It's a preparation work for a feature : Extrude feature polygon, to have a box instead of a flat polygon.
To extrude a polygon, I need the Normal vector.

In planar mode, it's ok, it's always the Vecor(0, 0, 1).
In globe mode, we always have to convert cartographic coordinate to cartesian position, and during this conversion, we compute the normal vertor.

I suggest that we store the normal vector in the coordinate object, during the conversion. So I will use the normal vector later in the extrusing process, without computing it a second times.

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -136,8 +136,11 @@ function _convert(coordsIn, newCrs, target) {
         }
     } else {
         if (coordsIn.crs === 'EPSG:4326' && newCrs === 'EPSG:4978') {
-            const cartesian = ellipsoid.cartographicToCartesian(coordsIn);
-            return target.set(newCrs, cartesian);
+            const normal = ellipsoid.geodeticSurfaceNormalCartographic(coordsIn);
+            const cartesian = ellipsoid.cartographicToCartesian(coordsIn, normal);
+            target.set(newCrs, cartesian);
+            target._normal = normal;
+            return target;
         }
 
         if (coordsIn.crs === 'EPSG:4978' && newCrs === 'EPSG:4326') {
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -224,7 +224,31 @@ export function convertValueToUnit(unitIn, unitOut, value) {
 function Coordinates(crs, ...coordinates) {
     this._values = new Float64Array(3);
     this.set(crs, ...coordinates);
+
+    Object.defineProperty(this, 'surfaceNormal',
+        {
+            configurable: true,
+            get: () => {
+                const result = this._normal || this.computeSurfaceNormal();
+                return result;
+            },
+        });
 }
+
+Coordinates.prototype.computeSurfaceNormal = function computeSurfaceNormal() {
+    // compute a normal only for cartesian coordinate.
+    if (this._internalStorageUnit != UNIT.METER) {
+        throw (new Error('You can only compute cartesian normal when unit is METER'));
+    }
+    // In globe mode (EPSG:4978), we compute the normal.
+    if (this.crs == 'EPSG:4978') {
+        this._normal = ellipsoid.geodeticSurfaceNormal(this);
+        return this._normal;
+    }
+    // In planar mode, normal is the up vector.
+    this._normal = THREE.Object3D.DefaultUp;
+    return this._normal;
+};
 
 Coordinates.prototype.set = function set(crs, ...coordinates) {
     _crsToUnitWithError(crs);
--- a/src/Core/Math/Ellipsoid.js
+++ b/src/Core/Math/Ellipsoid.js
@@ -19,7 +19,19 @@ function Ellipsoid(size) {
     this.size = new THREE.Vector3(size.x, size.y, size.z);
 
     this._radiiSquared = new THREE.Vector3(size.x * size.x, size.y * size.y, size.z * size.z);
+
+    this._oneOverRadiiSquared = new THREE.Vector3(size.x === 0.0 ? 0.0 : 1.0 / (size.x * size.x),
+        size.y === 0.0 ? 0.0 : 1.0 / (size.y * size.y),
+        size.z === 0.0 ? 0.0 : 1.0 / (size.z * size.z));
 }
+
+Ellipsoid.prototype.geodeticSurfaceNormal = function geodeticSurfaceNormal(cartesian) {
+    var result = new THREE.Vector3(
+        cartesian.x() * this._oneOverRadiiSquared.x,
+        cartesian.y() * this._oneOverRadiiSquared.y,
+        cartesian.z() * this._oneOverRadiiSquared.z);
+    return result.normalize();
+};
 
 Ellipsoid.prototype.geodeticSurfaceNormalCartographic = function geodeticSurfaceNormalCartographic(coordCarto) {
     var longitude = coordCarto.longitude(UNIT.RADIAN);
--- a/src/Core/Math/Ellipsoid.js
+++ b/src/Core/Math/Ellipsoid.js
@@ -56,10 +56,10 @@ Ellipsoid.prototype.setSize = function setSize(size) {
 };
 
 
-Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto) {
+Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto, normal) {
     // var n;
     var k = new THREE.Vector3();
-    var n = this.geodeticSurfaceNormalCartographic(coordCarto);
+    var n = normal.clone() || this.geodeticSurfaceNormalCartographic(coordCarto);
 
     k.multiplyVectors(this._radiiSquared, n);
 
--- a/test/coordinate_unit_test.js
+++ b/test/coordinate_unit_test.js
@@ -1,4 +1,5 @@
 /* global describe, it */
+import * as THREE from 'three';
 import proj4 from 'proj4';
 import assert from 'assert';
 import Coordinates, { UNIT } from '../src/Core/Geographic/Coordinates';
@@ -41,7 +42,7 @@ describe('Coordinate conversions', function () {
         assertCoordEqual(coord1, coord3);
     });
     // This case happend in iTowns when we convert the tile extent (4326 radian) to a target WFS server (EPSG:3946 for example) to request Lyon bus line in WFS.
-    it('Coordinate conversion from EPSG:4326 Radian (tiles extent) to EPSG:3946 (Lyon WFS) and back to EPSG:4326 (degrees)', function () {
+    it('should correctly convert from EPSG:4326 Radian (tiles extent) to EPSG:3946 (Lyon WFS) and back to EPSG:4326 (degrees)', function () {
         // geographic example for EPSG 4326 in degrees
         var longIn = 4.82212;
         var latIn = 45.723722;
@@ -58,5 +59,31 @@ describe('Coordinate conversions', function () {
         // verify coordinates
         assertFloatEqual(longIn, coord3.longitude());
         assertFloatEqual(latIn, coord3.latitude());
+    });
+});
+
+describe('Coordinate surface normale property', function () {
+    it('should correctly compute a surface normal ', function () {
+        const coord0 = new Coordinates('EPSG:4326', 15.0, 12.0).as('EPSG:4978');
+        const normal0 = coord0.surfaceNormal;
+
+        assertFloatEqual(normal0.x, 0.944818029);
+        assertFloatEqual(normal0.y, 0.253163227);
+        assertFloatEqual(normal0.z, 0.207911690);
+
+        const coord1 = new Coordinates('EPSG:4978', 6027050.95, 1614943.43, 1317402.53);
+        const normal1 = coord1.surfaceNormal;
+
+        assertFloatEqual(normal0.x, normal1.x);
+        assertFloatEqual(normal0.y, normal1.y);
+        assertFloatEqual(normal0.z, normal1.z);
+    });
+    it('should correctly return the default up vector for planar mode ', function () {
+        const coord0 = new Coordinates('EPSG:4326', 15.0, 12.0);
+        coord0._internalStorageUnit = UNIT.METER;
+
+        const normal0 = coord0.surfaceNormal;
+
+        assert.equal(normal0, THREE.Object3D.DefaultUp);
     });
 });


src/Core/Math/Ellipsoid.js
```diff
@@ -30,7 +30,8 @@ Ellipsoid.prototype.geodeticSurfaceNormalCartographic = function geodeticSurface
     var y = cosLatitude * Math.sin(longitude);
     var z = Math.sin(latitude);
 
-    var result = new THREE.Vector3(x, y, z);
+    result = result || THREE.Vector3();
```
peppsac: Missing `new` keyword.

src/Core/Math/Ellipsoid.js
```diff
@@ -44,10 +45,10 @@ Ellipsoid.prototype.setSize = function setSize(size) {
 };
 
 
-Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto) {
+Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto, normal) {
```
peppsac, line 59: You could rename the output parameter: `outNormal` to be more explicit.

Or alternatively (and I'd prefer this approach): you could `return { xyz:  k.add(n), normal: n} ` and adapt the other user of this function

peppsac (commented):
Looking good but it feels a bit weird to store a `normal` in a THREE.Vector3...

src/Core/Math/Ellipsoid.js
gliegard: > Or alternatively (and I'd prefer this approach): you could return { xyz: k.add(n), normal: n} and adapt the other user of this function.

I have updated the code with this approach, and rebase because some changes was not needed any more. Thanks for the suggestion, it's more elegant.

gliegard:
@itowns-reviewers

peppsac:
Sorry but I'm still not sure it's a good idea: with this change a `Coordinates` object will sometimes have a `normal` property, sometimes not.
*Maybe* we want to add a function/property on `Coordinates` (`surfaceNormal` ?) which would be  unitialized and computed when read?

```js
var normal;
Object.defineProperty(coord,
    'surfaceNormal',
    { get: () => {
    	normal = normal || computeSurfaceNormal();
    	return normal;
    });
```

Thoughts?

gliegard:
> Sorry but I'm still not sure it's a good idea: with this change a Coordinates object will sometimes have a normal property, sometimes not.

Yes I understand, but IMHO, I think it's a good compromise. The idea is :
- In globe mode, we have to convert coordinate, so we store the normal vector.
- In planar mode, normal is not defined, because it's always the Up vector.

> Maybe we want to add a function/property on Coordinates (surfaceNormal ?) which would be unitialized and computed when read?

Im' not sure, because I don't want to compute the normal after the conversion.

peppsac:
> In globe mode, we have to convert coordinate, so we store the normal vector.

In globe mode you'll only have the normal if you start from EPSG:4326 coords and convert  them to EPSG:4978. 
So for instance: `new Coordinates('EPSG:4326', X, Y).normal` would be `undefined` which is unfortunate.

Again: modifying a base object (Coordinates here) to add a property that is valid only in a very specific case and for one usage is problematic so we need to figure out a better solution.

> Im' not sure, because I don't want to compute the normal after the conversion.

You could adapt `cartographicToCartesian` to add a 2nd parameter: the precomputed normal.
And your code would become:
```js
              [...]
              const cartesian = ellipsoid.cartographicToCartesian(coordsIn, coordsIn.surfaceNormal); 
              [...]
```
`cartographicToCartesian` could be adapted like this:
```js
Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto, normal) {
    var k = new THREE.Vector3();
    var n = normal || this.geodeticSurfaceNormalCartographic(coordCarto);
```

WDYT?

gliegard:
I've tried the Object.defineProperty approach.

src/Core/Geographic/Coordinates.js
```diff
@@ -192,8 +194,30 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+            set: (normal) => {
+                this._normal = normal;
+            },
+        });
 }
 
+Coordinates.prototype.computeNormal = function computeNormal() {
+    // In globe mode, (EPSG:4978), coordinate has been converted and normal is already computed.
+    if (this.crs == '4978') {
+        throw (new Error('Should not happend'));
```
peppsac: `Should not happend` is not helpful as an error message :-)

Anyway, there's no justification for this potential error. Why don't you compute the normal if requested?

src/Core/Geographic/Coordinates.js
```diff
@@ -192,8 +194,30 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+    }
+    // In planar mode, normal is the up vector.
+    this._normal = THREE.Object3D.DefaultUp;
```
peppsac: I believe this is wrong. In planar mode, the `normal` is always (0, 0, 1) since planar coord systems all use x,y.

src/Core/Geographic/Coordinates.js
```diff
@@ -192,8 +194,30 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
     this.set(crs, ...coordinates);
+
+    Object.defineProperty(this, 'normal',
+        {
+            configurable: true,
+            get: () => {
+                const result = this._normal || this.computeNormal();
+                return result;
+            },
+            set: (normal) => {
```
peppsac: I don't think you want a setter: the surface normal at some point is non-mutable.
You shouldn't even make `computeNormal` part of the prototype: it should be a private function of the module.

src/Core/Geographic/Coordinates.js
```diff
@@ -192,8 +194,30 @@ export function convertValueToUnit(unitIn, unitOut, value) {
 function Coordinates(crs, ...coordinates) {
     this._values = new Float64Array(3);
     this.set(crs, ...coordinates);
+
+    Object.defineProperty(this, 'normal',
```
peppsac: I don't like `normal` name: too close to the generic 3D/math term.
I suggested `surfaceNormal` name but maybe there's an even better GIS-related name to express the "surface of an object without any DEM/MNT applied".

peppsac (commented):
I left a few comments regarding your new code.

I'm still not sure that this is a good idea (note that the suggestion in https://github.com/iTowns/itowns/pull/494#issuecomment-332934753 may be a better alternative)

autra (dismissed):
Let's not add a normal property on Coordinates, it doesn't make sense. Specifying an optional normal parameter in `cartographicToCartesian` is fine by me, and would cover the use case, I believe. 

If storing this value for a longer lifetime than a simple scope is necessary, let's move the problem up up one level, probably in the currently manipulated `Object3D`? But if it's just for loadKMZ, we don't need to do so anyway.

Also, as a general remark, in your `motivation and context` section, you don't really explain what's the *real* motivation. Is it a perf problem? If so where and how? What's your use case? Have you profiled? All these things are useful to assess the criticity of a PR for us reviewers. That being said, I don't deny this could be useful.

Thanks!

src/Core/Geographic/Coordinates.js
gliegard: Compute the normal AFTER the conversion is difficult.
I can add code to compute.

gliegard:
> You could adapt cartographicToCartesian to add a 2nd parameter: the precomputed normal.
And your code would become:

> Let's not add a normal property on Coordinates, it doesn't make sense. Specifying an optional normal parameter in cartographicToCartesian is fine by me, and would cover the use case, I believe.

Specifiing an optional normal parameter in cartographicToCartesian doesn't help in my case.
Because when Coordinate is already converted in EPSG:4978 (globe), we don't call cartographicToCartesian a second time since it's not a cartographic coordinate anymore.

> If storing this value for a longer lifetime than a simple scope is necessary, let's move the problem up up one level, probably in the currently manipulated Object3D? But if it's just for loadKMZ, we don't need to do so anyway.

The normal is computed in coordinate conversion (method coordinate.as('EPSG:4978') ). 
I need the normal after this conversion, in Feature2Mesh, just before creating an Object3D.
It's not for loadKMZ, I updated the motivation and context.

> Also, as a general remark, in your motivation and context section, you don't really explain what's the real motivation. Is it a perf problem? If so where and how? What's your use case? Have you profiled? All these things are useful to assess the criticity of a PR for us reviewers. That being said, I don't deny this could be useful.

Thanks, It was not very clear, I updated the 'movivation and context' section.

autra:
I see. But still: normals should not be in a Coordinates object. I suspect you want to have it in a geometry.

So actually : 

> It's a preparation work for a feature : Extrude feature polygon, to have a box instead of a flat polygon.
> To extrude a polygon, I need the Normal vector.

Do you already have a working branch for that? Because there's a lot of possible way to get normals, and three.js offers at least one (you can [compute normals of a geometry](https://threejs.org/docs/#api/core/Face3.normal)). It seems a more natural way to me to achieve what you want.

peppsac:
Honestly this is an optimisation feature that doesn't fit quite well in itowns' concepts yet.

So I'd propose that you implement this optimisation locally first, with something like this:
```js
normal = ellipsoid.geodeticSurfaceNormalCartographic(coord);
position = ellipsoid.cartographicToCartesian(coord, normal); // and modify cartographicToCartesian to accept an optional normal arg
```
instead of:
```js
position = coord.as('EPSG:4978');
```

Then if it's useful in several places we can try to design a better solution.

WDYT?

gliegard:
> But still: normals should not be in a Coordinates object

I don't agree. Coordinate object cant convert geographic coordinate to cartesian position (3D) using the Ellipsoid. Normal vector is in the 3D space like cartesian position.

> Honestly this is an optimisation feature that doesn't fit quite well in itowns' concepts yet.

It's not an optimisation, it's just avoid computing something already computed.

> Do you already have a working branch for that? Because there's a lot of possible way to get normals, and three.js offers at least one (you can compute normals of a geometry). It seems a more natural way to me to achieve what you want.

Yes I have a working branch origin/Feature_geometry_extruded

> So I'd propose that you implement this optimisation locally first, with something like this:
normal = ellipsoid.geodeticSurfaceNormalCartographic(coord);
position = ellipsoid.cartographicToCartesian(coord, normal); // and modify cartographicToCartesian to accept an optional normal arg
instead of:
position = coord.as('EPSG:4978');

It's not a good idea to modify the calling code, and I'm not sure it's a geographic coordinate at this point. I prefer the normal storing to be transparent. 

If I add some code to avoid throwing the error, I think it can be usefull for everybody to have the normal vector here. 

WDYT ?

peppsac:
> It's not a good idea to modify the calling code, and I'm not sure it's a geographic coordinate at this point. I prefer the normal storing to be transparent.

I'd prefer that you show how useful this feature is before trying to integrate it more into itowns.
If it's only used once (in FeatureProcessing), then it'll be just fine to stay contained at this place.
If it's used in several place **then** it'll be a good time to think about how we can generalize this idea.

I'm pretty sure you can easily adapt the calling site to distinguish between geographic coords (probably because the unit will be deg/rad) and other cases:

```js
if (isGeographic(...)) {
    normal = ellipsoid.geodeticSurfaceNormalCartographic(coord);
    position = ellipsoid.cartographicToCartesian(coord, normal);
} else {
    position = coord.as(...);
    normal = new THREE.Vector3(0, 0, 1);
}
```
This seems like a reasonnable first step to me: it covers your use case ("don't compute the normal twice") and doesn't fall in the early-generalization trap either.

gchoqueux:
it should be noted that when during the creation of the tiles, for each vertex, the normal to the surface is recalculated unnecessarily.

gliegard:
> I'm pretty sure you can easily adapt the calling site to distinguish between geographic coords (probably because the unit will be deg/rad) and other cases:

It's not so easy, for example, here is the use case 
WFS_Provider call GeoJSON2Feature, that read and convert coordinates (here we have the normal), and create a list of features. Then WFS_Provider call Feature2Mesh to convert a list of feature to 3D mesh.

And I work in Feature2Mesh to extrude a polygon. So I have features made of coordinates, and I've lost the normal vector of each coordinate.

gliegard:
I think the normal vector is an important part of the coordinate because it's a part of it's orientation.

src/Core/Geographic/Coordinates.js
peppsac: Why? If the user request the normal, you can go back to `4326` and compute the normal from there.

No need to throw.

peppsac:
For feature extrusion you only need 1 normal per object right so.. it would make sense to compute the normal when performing extrusion - it's not going to be costly anyway.

Again: I'm not against this PR per se. But I'm not convinced by the current version so how about merging your extrusion work instead and then come back to this PR?

autra:
> For feature extrusion you only need 1 normal per object

+1. Again, this should not be in a `Coordinates` instance (representing a point). I'll try to have a look at your branch.

gchoqueux:
it's an extrusion on the local normal. For example on the globe the walls follow the plumb bob.

autra:
> it's an extrusion on the local normal. For example on the globe the walls follow the plumb bob.

Isn't it completely overkill (if not wrong) for buildings?

gchoqueux:
see [cesium example](https://cesiumjs.org/Cesium/Apps/Sandcastle/index.html?src=Geometry%20and%20Appearances.html&label=Showcases)

autra:
> see cesium example

These are not buildings.

Actually, @gliegard, you're not really answering my question in your description

> So I will use the normal vector later in the extruding process, without computing it a second times.

*Why* don't you want to avoid this second computation? Is it a perf problem? If so where (initial load time, responsiveness)? Did you profile and assess that the slow thing was the normal computation? If yes what is causing this slowness? (my intuition: unlikely it'll cause a problem as it is not in a update, except for the `new Vector3() in `geodeticSurfaceNormalCartographic`, and that could be easily avoided with a closure).

Until it's proven that we need to cache this normal, we're only loosing time trying to kill an extra computation we're not even sure is causing any trouble. And it's at the cost of readability and maintenance.

So here's the course of action we need to follow:

1. Please explain why you think you can't work with polygon normals here (and really you do in my opinion).
1. If you convince us (why not after all? If it's useful, we definitely want it :-) ):
    1. Let's first try with computing the normal again when we need it. Measure the perf and see what these calls cost us
    1. if we assess that the cost is to high (but again I doubt it), let's try to understand *why* it is slow? 
    1. Devise a solution from here. It could very well be something similar to what you're proposing here, or something else entirely, according to what you've found.

Let's not loose time on problems that don't exist.

gchoqueux:
Indeed they are not buildings, but there is also the visualization of the data that can be extruded (Region, country ...)

autra:
> Indeed they are not buildings, but there is also the visualization of the data that can be extruded (Region, country ...)

ok fair enough, this can indeed be useful.

gchoqueux:
By example [here](https://github.com/iTowns/itowns/blob/c6a44eb30db27d9f3bf562707660c3953d2d7d9e/src/Core/Prefab/Globe/BuilderEllipsoidTile.js#L53) there are many recalculations of normals, so this is less efficient

src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +227,28 @@ export function convertValueToUnit(unitIn, unitOut, value) {
 function Coordinates(crs, ...coordinates) {
     this._values = new Float64Array(3);
     this.set(crs, ...coordinates);
+
+    Object.defineProperty(this, 'surfaceNormal',
```
peppsac, line 231: Let's steal the wording from your other PR: `geodeticSurfaceNormal` is a good name.

src/Core/Geographic/Coordinates.js
```diff
@@ -137,7 +137,10 @@ function _convert(coordsIn, newCrs, target) {
     } else {
         if (coordsIn.crs === 'EPSG:4326' && newCrs === 'EPSG:4978') {
             const cartesian = ellipsoid.cartographicToCartesian(coordsIn);
-            return target.set(newCrs, cartesian);
+            target.set(newCrs, cartesian.position);
+            // store the normal vector of the point.
+            target._surfaceNormal = cartesian.normal;
```
peppsac: Again.
Please reverse the logic here so we don't need to ever modify `target._surfaceNormal` outside of the property def.
```js
const cartesian = ellipsoid.cartographicToCartesian(coordsIn, coordsIn.geodeticSurfaceNormal);
target.set(newCrs, cartesian.position);
```

src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +227,28 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+        {
+            configurable: true,
+            get: () => {
+                const result = this._normal || this.computeNormal();
+                return result;
+            },
+        });
 }
 
+Coordinates.prototype.computeNormal = function computeNormal() {
```
peppsac: Do not expose this function please (ie : remove it from Coordinates prototype).

src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +227,28 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+    // In globe mode, (EPSG:4978), coordinate has been converted and normal is already computed.
+    if (this.crs == 'EPSG:4978') {
```
peppsac: You also need a special case for  `crs == 'EPSG:4326'`.

src/Core/Scheduler/Providers/KML_Provider.js
```diff
@@ -22,7 +22,7 @@ KML_Provider.prototype.loadKMZ = function loadKMZ(longitude, latitude) {
             var child = result.scene.children[0];
             var coorCarto = result.coorCarto;
 
-            var position = this.ellipsoid.cartographicToCartesian(coorCarto);
+            var position = this.ellipsoid.cartographicToCartesian(coorCarto).position;
             coorCarto.altitude = 0;
             var normal = this.ellipsoid.geodeticSurfaceNormalCartographic(coorCarto);
```
peppsac: As previously stated, modify this line for:
```js
    var normal = normal || this.ellipsoid.geodeticSurfaceNormalCartographic(coorCarto); 
```

src/Core/Geographic/Coordinates.js
peppsac: Are you sure `geodetic` is a real word? Or do you mean `geodesic...`?

peppsac (changes requested):
I don't care (yet) about the performance improvment that this PR will possibly bring.

But being able as a user to easily query the geodesic surface normal is useful to orient object correctly in globe mode so let's focus on this point first.

autra:
> But being able as a user to easily query the geodesic surface normal is useful to orient object correctly in globe mode so let's focus on this point first.

Ok, fair enough as well.

> By example here there are many recalculations of normals, so this is less efficient

Yes, but it's not necessarily a problem. To put it in another perspective, it could be less efficient computation-wise, but more efficient maintenance-wise (because yielding a clearer code). For example in this case, I did the homework, and here are my results using the Chrome performance devtool:

selection_051

I get similar results for the tile builder (1.5% of the globe construction time). In both case, it's a premature optimization in my opinion. 

That being said, looking at the profile data, I think we can improve the performance of the GeoJSON parser by inverting the logic of filtering of `GeometryToCoordinates.<type>` : at the moment, we parse all the Coordinates, convert *all* of them (which is quite slow according to the profile), *then* filter with `filteringExtent`. Converting the extent the other way round and filtering *before* parsing and converting the coordinates will likely cut down the parsing time a lot. We would create less object too, reducing the GC time (which is actually the slowest operation in all the parsing). I think this would yield the best results in this case. WDYT @gliegard ?

gliegard:
Ok, maybe it's a premature optimisation, we'll come back here later :)
Thanks.

PR #495, opened 2017-09-27
fix (GeoJSON2Feature) : filteringExtent parameter was missing

gliegard:
## Description
Fix (GeoJSON2Feature)  filteringExtent parameter was missing for multiLineString conversion method.

## Motivation and Context
Prior to this change, in the WFS Planar example, we could see a TypeError in the console.


gliegard:
@itowns-reviewers

peppsac (approved):
Thanks.
Could you rewrite your commit title using: `fix (Feature): ...` before merging?

PR #496, opened 2017-09-27
fix (Extent) during conversion, internal storage unit was not given

gliegard:
## Description
Fix extent conversion

## Motivation and Context
During an extent conversion, internal storage unit was not used.

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -111,7 +111,11 @@ Extent.prototype.as = function as(crs) {
         let south = Infinity;
         let east = -Infinity;
         let west = Infinity;
+        // loop over the coordinates
         for (let i = 0; i < cardinals.length; i++) {
+            // add the internalStorageUnit to the coordinate.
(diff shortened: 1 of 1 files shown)


src/Core/Geographic/Extent.js
```diff
@@ -112,6 +112,7 @@ Extent.prototype.as = function as(crs) {
         let east = -Infinity;
         let west = Infinity;
         for (let i = 0; i < cardinals.length; i++) {
+            cardinals[i]._internalStorageUnit = this._internalStorageUnit;
```
peppsac: It seems correct but would require a comment because it "looks" wrong (because we overwrite `cardinals[i]`  at the next line).

gliegard: ok, i have updated the code.

peppsac:
Your git commit title format is incorrect (see https://github.com/iTowns/itowns/pull/498#issuecomment-332889446)

peppsac (approved):
Thanks!

gliegard:
Thanks.

PR #498, opened 2017-09-27
fix (Coordinate) debug conversion to EPSG:4978

gliegard:
## Description
- fix coordination conversion with proj4

## Motivation and Context
- There is a bug to convert anything to 4978 with proj4 : https://github.com/proj4js/proj4js/issues/195
- If we have EPSG:4326 in Radian, we have to convert it in Degree before using Proj4.

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -9,6 +9,8 @@ import proj4 from 'proj4';
 import mE from '../Math/MathExtended';
 import Ellipsoid from '../Math/Ellipsoid';
 
+proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
+
 const projectionCache = {};
 
 export function ellipsoidSizes() {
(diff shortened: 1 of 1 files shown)


peppsac (commented):
Looks ok but you need to fix your commit message:
1. fix (Coordinate): workaround proj4 EPSG:4978 bug
2. the PR text should be your commit message content

gliegard:
ok, it's done.

peppsac:
You forgot the `:` (and no space between the closing `)` and the `:`) in your commit message.
It's important to be able to automatize (partially) the writing of the changelog.

gliegard:
ok

src/Core/Geographic/Coordinates.js
```diff
@@ -148,8 +150,23 @@ function _convert(coordsIn, newCrs, target) {
...
 
         if (coordsIn.crs in proj4.defs && newCrs in proj4.defs) {
-            const p = instanceProj4(coordsIn.crs, newCrs).forward([coordsIn._values[0], coordsIn._values[1]]);
-            return target.set(newCrs, p[0], p[1], coordsIn._values[2]);
+            // little hack due to our choice to use Radian for the EPSG:4326 projection (instead of the standard EPSG:4326 in Degree)
+            const needToHack = coordsIn.crs === 'EPSG:4326' && coordsIn._internalStorageUnit == UNIT.RADIAN;
+            const val0 = needToHack ? mE.radToDeg(coordsIn._values[0]) : coordsIn._values[0];
+            const val1 = needToHack ? mE.radToDeg(coordsIn._values[1]) : coordsIn._values[1];
+
+            // there is a bug for converting anything to 4978 with proj4
```
peppsac: Do converting from 4978 works correctly? The linked issue seems to indicate that both projections `to` and `from` 4978 aren't supported.

gliegard: I don't know, I don't have an example to test this case.

gliegard: Ok, I tried and converting From 4978 doesn't work correctly.

gliegard:
updated

src/Core/Geographic/Coordinates.js
```diff
@@ -148,8 +150,38 @@ function _convert(coordsIn, newCrs, target) {
         }
 
         if (coordsIn.crs in proj4.defs && newCrs in proj4.defs) {
-            const p = instanceProj4(coordsIn.crs, newCrs).forward([coordsIn._values[0], coordsIn._values[1]]);
-            return target.set(newCrs, p[0], p[1], coordsIn._values[2]);
+            var val0 = coordsIn._values[0];
+            var val1 = coordsIn._values[1];
```
peppsac: Use `let` rather than `var`.

The whole function looks nicer than in your previous iterations, so good job 👍

src/Core/Geographic/Coordinates.js
```diff
@@ -148,8 +150,38 @@ function _convert(coordsIn, newCrs, target) {
...
+                } else if (coordsIn._internalStorageUnit == UNIT.RADIAN && refUnit == UNIT.DEGREE) {
+                    val0 = coordsIn.longitude(UNIT.DEGREE);
+                    val1 = coordsIn.latitude(UNIT.DEGREE);
+                }
+            }
+
+            // there is a bug for converting anything from and to 4978 with proj4
+            // https://github.com/proj4js/proj4js/issues/195
+            if (newCrs == 'EPSG:4978') {
+                // the workaround is to use an intermediate projection, like EPSG:4326
```
peppsac: This comment applies to both the `if` and `else if` case, so you could move it 1 line above.

test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
+/* global describe, it */
```
peppsac: Nice to have, but I think this should move in a new PR.

Or if nobody is against it, you could reword the PR title / commits to move from a 'fix bug' PR to a 'feature: add Coordinates conversion unit testing' PR.

autra: > Nice to have, but I think this should move in a new PR.

As it is testing the new hopefully bug-free behaviour, why do you want it in another PR?

> feature: add Coordinates conversion unit testing

a unit test is not a feature imo. I'm not against separating the bug fix from the test in 2 commits (although I'd prefer one commit), but it should be prefixed with 'chore' (or 'test' as it is currently), right?

gliegard:
Ok, I've separated the PR !

peppsac (approved):
Looking good, let's move forward 👍

PR #499, opened 2017-09-28
fix (3dtiles): do not try to subdivide a tile without children

peppsac:
This commits add a check to prevent an infinite loop when using 3dtiles.

Without it the 'process3dTilesNode' would try to subdivide tiles that
don't have children and on completion of the subdivision (ie: instantly) would
issue a 'view.notifyChange()' call.


autra (approved):
Thanks!

PR #501, opened 2017-09-29
feature: add support for transparent 3d-tiles geometries

autra:
This is a follow up of #488, doing the same work for 3dTiles.


src/Core/Prefab/Globe/Atmosphere.js
```diff
@@ -65,6 +65,7 @@ function Atmosphere() {
         side: THREE.FrontSide,
         blending: THREE.AdditiveBlending,
         transparent: true,
+        depthWrite: false,
```
autra, line 68: The atmosphere can hide transparent object if its frontier is between the camera and the object.

peppsac (approved):
Looks good to me

PR #502, opened 2017-10-03
Unit test coordinate conversion

gliegard:
## Description
- add unit test in 'npm test' command
- add unit tests for Coordinate object

## Motivation and Context
Test coordinate conversions, fixed here #498


test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
...
+function assertFloatEqual(float1, float2) {
+    assert.equal(Number(float1).toFixed(5), Number(float2).toFixed(5));
+}
+
+// Assert two coordinates obects are equals.
+function assertCoordEqual(coord1, coord2) {
+    assert.equal(coord1._internalStorageUnit, coord2._internalStorageUnit);
+    assertFloatEqual(coord1._values[0], coord2._values[0]);
+    assertFloatEqual(coord1._values[1], coord2._values[1]);
+    assertFloatEqual(coord1._values[2], coord2._values[2]);
```
autra: You should compare the crs too.

test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+describe('Coordinate conversions', function () {
+    this.timeout(10000);
```
autra: this smells. Why do you need that?

test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
...
+    it('Coordinate conversion from EPSG:4326 to EPSG:4978 and back to EPSG:4326', function (done) {
+        var coord1 = new Coordinates('EPSG:4326', 15, 12);
+        var coord2 = coord1.as('EPSG:4978');
+        var coord3 = coord2.as('EPSG:4326');
+        assertCoordEqual(coord1, coord3);
+        done();
```
autra: you don't need done at all for synchronous tests.

test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
...
+function assertCoordEqual(coord1, coord2) {
+    assert.equal(coord1._internalStorageUnit, coord2._internalStorageUnit);
+    assertFloatEqual(coord1._values[0], coord2._values[0]);
+    assertFloatEqual(coord1._values[1], coord2._values[1]);
+    assertFloatEqual(coord1._values[2], coord2._values[2]);
+}
+
+describe('Coordinate conversions', function () {
+    this.timeout(10000);
+    it('Coordinate conversion from EPSG:4326 to EPSG:4978 and back to EPSG:4326', function (done) {
```
autra: the `it` style in mocha should read like real English. So something like:
```js
it('should correctly convert from EPSG:4326 to EPSG:4978 and back to EPSG:4326')
```
(we can do better than that probably, but you get the idea)

test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
...
+    it('Coordinate conversion from EPSG:4326 Radian (tiles extent) to EPSG:3946 (Lyon WFS) and back to EPSG:4326 (degrees)', function (done) {
+        // geographic example for EPSG 4326 in degrees
+        var longIn = 4.82212;
+        var latIn = 45.723722;
+        // let's define an input coordinate EPSG:4326 in radian.
+        var coord1 = new Coordinates('EPSG:4326', longIn / 180 * Math.PI, latIn / 180 * Math.PI);
+        coord1._internalStorageUnit = UNIT.RADIAN;
+        // convert coordinate in EPSG:3946
+        var coord2 = coord1.as('EPSG:3946');
+        // and convert back to EPSG:4626 stadanrd in degree.
```
autra: typo `standard`

test/coordinate_unit_test.js
```diff
@@ -0,0 +1,56 @@
...
+    it('Coordinate conversion from EPSG:4326 to EPSG:4978 and back to EPSG:4326', function (done) {
+        var coord1 = new Coordinates('EPSG:4326', 15, 12);
+        var coord2 = coord1.as('EPSG:4978');
+        var coord3 = coord2.as('EPSG:4326');
+        assertCoordEqual(coord1, coord3);
+        done();
+    });
+    it('Coordinate conversion from EPSG:3946 (Lyon WFS bus example) to EPSG:4978 (Globe) and back to EPSG:3946', function (done) {
+        var coord1 = new Coordinates('EPSG:3946', 1500000, 5100000, 12);
+        var coord2 = coord1.as('EPSG:4978');
```
autra: can we assert coord2 too? You can use epsg.io to get the expected value in EPSG:4978.

autra (changes requested):
Thanks for adding unit tests in itowns :-) I've left some comments, but the principle is good imo.

test/coordinate_unit_test.js
gliegard: thanks, i'don't need that :)

test/coordinate_unit_test.js
gliegard: Good idea, but 
- epsg.io doesn't work with an altitude input (here it's 12)
- and epsg.io doesn't give Z value in output, so the coordinate comparison fail.

gliegard: And I's not the purpose of this test : In this test, I want to convert a coordinate to another crs, and convert back to the first crs. I prefer this test to be simple and easy to read like this.

gliegard:
updated

test/coordinate_unit_test.js
autra: > epsg.io doesn't work with an altitude input (here it's 12)

I'm sure there's other way of finding converted altitude :-)

> And I's not the purpose of this test : In this test, I want to convert a coordinate to another crs, and convert back to the first crs. I prefer this test to be simple and easy to read like this.

The test would also pass if `.as` does absolutely nothing, that's why I thing it's more robuts if we also test that it *actually* do something. 

For a start, you could just assert on x and y.

autra: Note that this remark is also true for the other tests, some of which do not set a z.

autra:
For this message commit

> feature (npm): add unit tests

Not a feature (reserved for itowns feature), use `chore` or `test` instead.

gliegard:
updated

test/coordinate_unit_test.js
gliegard: as I want to keep the tests simple,
I added another test to test simple conversion to another crs and verify values.

autra (approved):

autra:
thanks!

gliegard:
thanks!

Issue #399, opened 2017-07-20
Window resize causes camera to jump in examples/orthographic.html


nosy-b:
When changing window size, the camera target location of the view jumps far away
It is the only example with this bug.
I can reproduce by following these steps:
- Launch examples/orthographic.html
- Move the map
- Go FullScreen

Expected results: Map center location stay the same
Actual results: Map center location moves far

peppsac:
Can you add [detailled repro steps](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Bug_writing_guidelines#Writing_precise_steps_to_reproduce) please?

peppsac:
Sorry it took so long to fix, I missed the edit in your original post.

peppsac:
Fixed with #503

PR #503, opened 2017-10-04
fix (Camera): don't reset camera's position on resize

peppsac:
Camera's position y component was set to -h/h, which centered the view around 0.
This commit adjust camera's height to the new ratio but also remember the old
camera's y position.

Fixes issue https://github.com/iTowns/itowns/issues/399


nosy-b (approved):
Works good now!

PR #505, opened 2017-10-05
feature (Coordinate) compute normal of the coordinate in cartesian.

gliegard:
## Description
- feature (Coordinate) compute normal of the coordinate in cartesian.
- add corresponding unit test

## Motivation and Context
Preparation work for a next feature : the extrusion of polygons.

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -224,6 +224,29 @@ export function convertValueToUnit(unitIn, unitOut, value) {
 function Coordinates(crs, ...coordinates) {
     this._values = new Float64Array(3);
     this.set(crs, ...coordinates);
+
+    Object.defineProperty(this, 'geodesicNormal',
+        {
+            configurable: true,
+            get: () => {
+                this._normal = this._normal || computeGeodesicNormal(this);
+                return this._normal;
+            },
+        });
+}
+
+const planarNormal = new THREE.Vector3(0, 0, 1);
+
+function computeGeodesicNormal(coord) {
+    if (coord.crs == 'EPSG:4326') {
+        return ellipsoid.geodeticSurfaceNormalCartographic(coord);
+    }
+    // In globe mode (EPSG:4978), we compute the normal.
+    if (coord.crs == 'EPSG:4978') {
+        return ellipsoid.geodeticSurfaceNormal(coord);
+    }
+    // In planar mode, normal is the up vector.
+    return planarNormal;
 }
 
 Coordinates.prototype.set = function set(crs, ...coordinates) {
--- a/src/Core/Math/Ellipsoid.js
+++ b/src/Core/Math/Ellipsoid.js
@@ -19,7 +19,19 @@ function Ellipsoid(size) {
     this.size = new THREE.Vector3(size.x, size.y, size.z);
 
     this._radiiSquared = new THREE.Vector3(size.x * size.x, size.y * size.y, size.z * size.z);
+
+    this._oneOverRadiiSquared = new THREE.Vector3(size.x === 0.0 ? 0.0 : 1.0 / (size.x * size.x),
+        size.y === 0.0 ? 0.0 : 1.0 / (size.y * size.y),
+        size.z === 0.0 ? 0.0 : 1.0 / (size.z * size.z));
 }
+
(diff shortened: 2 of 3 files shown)


src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +224,32 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+        {
+            configurable: true,
+            get: () => {
+                const result = this._normal || this.computeSurfaceNormal();
+                return result;
+            },
+        });
 }
 
+Coordinates.prototype.computeSurfaceNormal = function computeSurfaceNormal() {
```
peppsac: Move this function out of the prototype: the user don't need to call it explicitely so let's not expose it.

src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +224,32 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+    // compute a normal only for cartesian coordinate.
+    if (this._internalStorageUnit != UNIT.METER) {
```
peppsac: Please don't throw an error.
You just need to handle the `4326` case (using `ellipsoid.cartographicToCartesian`) and you're all set.

src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +224,32 @@ export function convertValueToUnit(unitIn, unitOut, value) {
...
+        throw (new Error('You can only compute cartesian normal when unit is METER'));
+    }
+    // In globe mode (EPSG:4978), we compute the normal.
+    if (this.crs == 'EPSG:4978') {
+        this._normal = ellipsoid.geodeticSurfaceNormal(this);
+        return this._normal;
+    }
+    // In planar mode, normal is the up vector.
+    this._normal = THREE.Object3D.DefaultUp;
```
peppsac: As stated earlier, you must use `new THREE.Vector3(0, 0, 1)` not `DefaultUp`

src/Core/Geographic/Coordinates.js
```diff
@@ -224,8 +224,32 @@ export function convertValueToUnit(unitIn, unitOut, value) {
 function Coordinates(crs, ...coordinates) {
     this._values = new Float64Array(3);
     this.set(crs, ...coordinates);
+
+    Object.defineProperty(this, 'surfaceNormal',
+        {
+            configurable: true,
+            get: () => {
+                const result = this._normal || this.computeSurfaceNormal();
```
peppsac: And as you use a internal member (`_normal`) to store the value, you can also copy the normal when it makes sense (ie : when going from 4326 to 4978 and the otherway around).

So add something similar to what you did here: https://github.com/iTowns/itowns/pull/494/files#diff-7c37ee347b0423d12e6624421e03f4caR142

```js
target._normal = coordsIn.surfaceNormal;
```

and you can then close PR #494 :)

peppsac: And I still prefer the `geodesicNormal` or `geodesicSurfaceNormal` terms because they contain an important piece of information: this normal isn't computed using a DEM/MNT.

peppsac (changes requested):

src/Core/Geographic/Coordinates.js
gliegard: In planar view, DefaultUp is defined as the normal to the plan.
https://github.com/iTowns/itowns/blob/061d19e370e68e22af9e23d82ff4fc9ffc76d28c/src/Core/Prefab/PlanarView.js#L121

It's more generic to use DefaultUp. WDYT ?

peppsac: Nope.
This settings is ignored when tiled are built: gis coordinates are always mapped on x,y so z is always the normal, regardless of DefaultUp.

gliegard:
updated

peppsac:
Did you forget about https://github.com/iTowns/itowns/pull/505#discussion_r143223683? Or do you want to implent this in another PR?

gliegard:
No I didn't forget, I tried but that introduced a strange bug..
I prefer implement that in another PR.

peppsac (approved):
Thanks, looks good to me.

gliegard:
Thanks.

PR #507, opened 2017-10-06
3dtiles micro fixes

peppsac:
2 small 3d-tiles commits:

    feat (3dtiles): allow layer.overrideMaterials to be a material
    
    If layer.overrideMaterials is a material, all meshes for this layer
    will get a clone() of it as their material.
    If it's 'true', all meshes will get a default MeshLambertMaterial mat.
    Else they'll get the GLTFLoader supplied mesh.
    
    This commit also adds the missing material.dispose() call.

Second one:

    fix (3dtiles): use distance to as commands priority value
    
    Distance may not be the best metric to determine priority but it's better
    than a constant value.

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -133,7 +133,13 @@ $3dTiles_Provider.prototype.b3dmToMesh = function b3dmToMesh(data, layer) {
             mesh.frustumCulled = false;
             if (mesh.material) {
                 if (layer.overrideMaterials) {
-                    mesh.material = new THREE.MeshLambertMaterial(0xffffff);
+                    mesh.material.dispose();
+                    if (typeof (layer.overrideMaterials) === 'object' &&
+                        layer.overrideMaterials.isMaterial) {
+                        mesh.material = layer.overrideMaterials.clone();
+                    } else {
+                        mesh.material = new THREE.MeshLambertMaterial({ color: 0xffffff });
+                    }
                 } else if (Capabilities.isLogDepthBufferSupported()
                             && mesh.material.isRawShaderMaterial
                             && !layer.doNotPatchMaterial) {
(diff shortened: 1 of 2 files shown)


src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -133,7 +133,13 @@ $3dTiles_Provider.prototype.b3dmToMesh = function b3dmToMesh(data, layer) {
             mesh.frustumCulled = false;
             if (mesh.material) {
                 if (layer.overrideMaterials) {
-                    mesh.material = new THREE.MeshLambertMaterial(0xffffff);
+                    mesh.material.dispose();
```
autra, line 136: In the future, it would be cool to be able to pass parameters to the gltf parser.

autra (approved):

src/Process/3dTilesProcessing.js
```diff
@@ -4,7 +4,7 @@ function requestNewTile(view, scheduler, geometryLayer, metadata, parent) {
         view,
         requester: parent,
         layer: geometryLayer,
-        priority: 10000,
+        priority: parent ? 1.0 / (parent.distance + 1) : 100,
```
Jeremy-Gaillard, line 7: I'm a bit worried about how the priority will interact with other layers. With values this small, 3d-tiles tiles will always be downloaded after terrain tiles, regardless of the precision of the data.

This is part of a broader issue about how to homogenize priority values though, so you can go ahead and merge.

Jeremy-Gaillard (approved):

src/Process/3dTilesProcessing.js
autra: Yep, on the other hand, when we mixed pointcloud with 3dTiles, the old value made itowns wait for *all* 3dtiles building to be downloaded before even downloading the first point ;-) I'll say this is better than before, if not ideal.

peppsac: IMHO priorities only make sense relative to other commands for the same layer and provider.

Currently the scheduler uses the priority as an absolute value for a given host which is wrong... I have plans to fix this, I just need some free time :)

Jeremy-Gaillard: I agree. Good to know you have plans for improving priority management.

PR #508, opened 2017-10-06
feat (3dtiles): improve gltf line support

peppsac:
Map all gltf line modes to the proper THREE.Line* class.

From three's documentation:
    THREE.Line is using gl.LINE_STRIP
    THREE.LineSegments [...] is rendered using gl.LINES
    THREE.LineLoop [...] is rendered using gl.LINE_LOOP

And from GLTF spec:
    {
        "enum": [ 1 ],
        "description": "LINES"
    },
    {
        "enum": [ 2 ],
        "description": "LINE_LOOP"
    },
    {
        "enum": [ 3 ],
        "description": "LINE_STRIP"
    },
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/mesh.primitive.schema.json#L28

### DIFF
--- a/src/Renderer/ThreeExtended/GLTFLoader.js
+++ b/src/Renderer/ThreeExtended/GLTFLoader.js
@@ -80,8 +80,6 @@ THREE.GLTFLoader = ( function () {
 
 			}
 
-			console.time( 'GLTFLoader' );
-
 			var parser = new GLTFParser( json, extensions, {
 
 				path: path || this.path,
@@ -90,8 +88,6 @@ THREE.GLTFLoader = ( function () {
 			} );
 
 			parser.parse( function ( scene, scenes, cameras, animations ) {
-
-				console.timeEnd( 'GLTFLoader' );
 
 				var glTF = {
 					"scene": scene,
@@ -428,6 +424,8 @@ THREE.GLTFLoader = ( function () {
 		SAMPLER_2D: 35678,
 		TRIANGLES: 4,
 		LINES: 1,
+		LINE_LOOP: 2,
+		LINE_STRIP: 3,
 		UNSIGNED_BYTE: 5121,
 		UNSIGNED_SHORT: 5123,
 
@@ -1685,7 +1683,9 @@ THREE.GLTFLoader = ( function () {
 
 						group.add( meshNode );
 
-					} else if ( primitive.mode === WEBGL_CONSTANTS.LINES ) {
+						} else if ( primitive.mode === WEBGL_CONSTANTS.LINES
+							|| primitive.mode === WEBGL_CONSTANTS.LINE_STRIP
+							|| primitive.mode === WEBGL_CONSTANTS.LINE_LOOP ) {
 
 						var geometry = new THREE.BufferGeometry();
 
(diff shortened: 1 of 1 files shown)


src/Renderer/ThreeExtended/GLTFLoader.js
```diff
@@ -1726,8 +1726,15 @@ THREE.GLTFLoader = ( function () {
...
-
-							meshNode = new THREE.Line( geometry, material );
+							if (primitive.mode === WEBGL_CONSTANTS.LINES ) {
+								meshNode = new THREE.LineSegments( geometry, material );
+							} else if (primitive.mode === WEBGL_CONSTANTS.LINE_STRIP ) {
+								meshNode = new THREE.Line( geometry, material );
+							} else if (primitive.mode === WEBGL_CONSTANTS.LINE_LOOP) {
+								meshNode = new THREE.LineLoop( geometry, material );
+							} else {
+								meshNode = new THREE.Line( geometry, material );
```
Jeremy-Gaillard: Shouldn't you throw an error instead if the mode is unknown ?

src/Renderer/ThreeExtended/GLTFLoader.js
```diff
@@ -80,8 +80,6 @@ THREE.GLTFLoader = ( function () {
 
 			}
 
-			console.time( 'GLTFLoader' );
```
Jeremy-Gaillard, line 83: How about surrounding these with `if(__DEBUG__)` instead of removing them?

src/Renderer/ThreeExtended/GLTFLoader.js
peppsac: We only enter here if:
```js
if ( primitive.mode === WEBGL_CONSTANTS.LINES
+							|| primitive.mode === WEBGL_CONSTANTS.LINE_STRIP
+							|| primitive.mode === WEBGL_CONSTANTS.LINE_LOOP )
```
is true anyway. If you want I can remove the else completely?

src/Renderer/ThreeExtended/GLTFLoader.js
peppsac: Because it's very noisy when you have a big tileset.

Is this timing info important for you?

src/Renderer/ThreeExtended/GLTFLoader.js
Jeremy-Gaillard: You could remove the last else and do this:

```js
else { // WEBGL_CONSTANTS.LINE_LOOP)
    meshNode = new THREE.LineLoop( geometry, material );
}
```

src/Renderer/ThreeExtended/GLTFLoader.js
Jeremy-Gaillard: It has been useful a few times, but I'm starting to work with glTF2, so you can go ahead and delete it.

src/Renderer/ThreeExtended/GLTFLoader.js
peppsac: Updated.

autra (approved):
Looks good to me

PR #511, opened 2017-10-06
Geoid2

qdnguyen:
Description

This PR is needed to visualize geoid color and DTM.

The changes were done 
- Add zfactor to deform geoid DTM
- Add geoid example

Screenshots 

image


src/Renderer/Shader/TileVS.glsl
```diff
@@ -14,6 +14,8 @@ uniform sampler2D   dTextures_00[1];
 uniform vec3        offsetScale_L00[1];
 uniform int         loadedTexturesCount[8];
 
+uniform float       zFactor;
```
peppsac, line 17: Please move all the `zFactor` stuff to a specific commit.

src/Renderer/Shader/TileVS.glsl
```diff
@@ -59,7 +59,8 @@ void main() {
                 }
 
             #elif defined(DATA_TEXTURE_ELEVATION)
-                float   dv  = max(texture2D( dTextures_00[0], vVv ).w, 0.);
+                float   dv  = texture2D( dTextures_00[0], vVv ).w;
+                dv *=zFactor;
```
peppsac: You should be doing `dv *=zFactor;` below (just before `vPosition = vec4( position +  vNormal  * dv ,1.0 )`).
This way it would work with all elevation data sources.

peppsac (changes requested):
Thanks for the rebase :)

Please use proper wording for your commits (see http://codeinthehole.com/writing/a-useful-template-for-commit-messages/ for some help) and split them in separate topics (e.g: 1 for the zFactor feature and 1 for the new example)

src/Renderer/Shader/TileVS.glsl
qdnguyen: i did it before like you said and it work but i dont know why it does not work now if i put it before vPosition = vec4( position + vNormal * dv ,1.0 )). I think three is a problem with vNormal

.gitignore
```diff
@@ -6,3 +6,5 @@
 /node_modules/
 /dist/
 /lib/
+/nbproject/private/
```
peppsac: This doesn't belong to the itowns's `.gitignore` - gitignore is only populated with folders/files that are by-product or temp files of itowns.

examples/geoid.html
```diff
@@ -0,0 +1,71 @@
+<!DOCTYPE html>
+<!--
```
peppsac: Please remove this generated header.

examples/geoid.html
```diff
@@ -0,0 +1,71 @@
...
+To change this license header, choose License Headers in Project Properties.
+To change this template file, choose Tools | Templates
+and open the template in the editor.
+-->
+<html>
+    <head>
+        <title>Globe</title>
```
peppsac: And fix the title :)

examples/geoid.html
```diff
@@ -0,0 +1,71 @@
...
+            var viewerDiv = document.getElementById('viewerDiv');
+            var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var menuGlobe = new GuiTools('menuDiv', globeView);
+            menuGlobe.view = globeView;
+
+            var promises = []
+            promises.push(itowns.Fetcher.json('layers/JSONLayers/GeoidTexture.json').then(function(result) { return globeView.addLayer(result); }));
+            //promises.push(itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(function(result) { return globeView.addLayer(result); }));
+            promises.push(itowns.Fetcher.json('layers/JSONLayers/GeoidMNT.json').then(function(result) { return globeView.addLayer(result); }));
+            //promises.push(itowns.Fetcher.json('layers/JSONLayers/IGN_MNT.json').then(function(result) { return globeView.addLayer(result); }));
```
peppsac: And remove dead code in all this file.

nbproject/project.properties
```diff
@@ -0,0 +1,3 @@
+files.encoding=UTF-8
```
peppsac: This `nbproject` folder must not be added to iTowns.

src/Renderer/Shader/TileVS.glsl
peppsac: Hmm.. I don't know what is the underlying issue here, but you need to fix this so the `z factor` feature works independently of the elevation data source.

utils/debug/Main.js
```diff
@@ -2,3 +2,4 @@ export { default as Debug } from './Debug';
 export { default as PointCloudDebug } from './PointCloudDebug';
 export { default as createTileDebugUI } from './TileDebug';
 export { default as create3dTilesDebugUI } from './3dTilesDebug';
+
```
peppsac: Do not add empty lines please.

src/Renderer/Shader/TileVS.glsl
```diff
@@ -69,6 +70,7 @@ void main() {
             #endif
 
             vPosition = vec4( position +  vNormal  * dv ,1.0 );
+
```
peppsac: Do not add empty lines please.

src/Core/TileMesh.js
```diff
@@ -169,7 +169,11 @@ TileMesh.prototype.center = function center() {
 };
 
 TileMesh.prototype.OBB = function OBB() {
-    return this.geometry.OBB;
+    var zFactor = this.material.uniforms.zFactor.value;
+    var OBB = this.geometry.OBB;
+        OBB.box3D.min.z *= zFactor;
```
peppsac: Indentation is wrong here.

src/Renderer/LayeredMaterial.js
```diff
@@ -491,5 +493,9 @@ LayeredMaterial.prototype.setSelected = function setSelected(selected) {
     this.uniforms.selected.value = selected;
 };
 
+LayeredMaterial.prototype.setLayerZFactor = function setLayerZFactor(zFactor) {
```
peppsac: Is this method useful? It's just a one-line method and you don't use it https://github.com/iTowns/itowns/pull/511/files#diff-caf5227be4f9e2f283d82df32a678e4bL523 so I'd suggest removing it for now.

peppsac (commented):
Quick review, focusing mostly on style issues.

.gitignore
autra: (You can add your own ignore stuff in the .gitignore in your home folder)

src/Renderer/Shader/TileVS.glsl
nguyenign: i set zFactor value by default to 1, so it will not impact on elevation data.

nbproject/project.properties
nguyenign: thanks for your advice

src/Renderer/Shader/TileVS.glsl
peppsac: My point is: we want this to work regardless of the data source used for elevation.
As an example, `zFactor` feature should work on planar.html demo.

src/Core/TileMesh.js
nguyenign: don't understant! We need it to adapt Bbox to geoid tiles'bbox, zFactor is set by default to 1 so it will not have an impact on  DTM

src/Renderer/LayeredMaterial.js
nguyenign: yes, i used it in updateLayeredMaterialNodeElevation to deform geoid

src/Renderer/Shader/TileVS.glsl
nguyenign: done

utils/debug/Main.js
nguyenign: done

src/Core/TileMesh.js
peppsac: The [code indentation](https://en.wikipedia.org/wiki/Indentation_style) is wrong.

Btw, the code is wrong too since you're going to modify `OBB.box3D.min.z` (and max) each time the `OBB()` function is called.
So for instance if `min.z = 1.0` and `zFactor = 3.0` initially, `min.z` will be: `3` then `9` then `27`...

src/Renderer/LayeredMaterial.js
peppsac: I meant: it's a one-liner and you only use once.
So I'd prefer you to remove it (and replace it's usage in `updateLayeredMaterialNodeElevation` with `material.uniforms.zFactor.value = layer.zFactor;`)

src/Renderer/Shader/TileVS.glsl
qdnguyen: ok, done

src/Renderer/LayeredMaterial.js
qdnguyen: done

peppsac:
Just tested:
- you should upload your data to https://github.com/iTowns/iTowns2-sample-data, so you'll be able to use a URL that works for everyone (see https://github.com/iTowns/itowns/blob/master/examples/pointcloud.html#L105 for an example)
- you have a merge commit in your PR: you need to do a rebase
- your commits do not respect the git messages light convention we're trying to adopt: see https://github.com/iTowns/itowns/issues/518

Thanks!

src/Core/TileMesh.js
qdnguyen: is there a posibility that in TileMesh i can have a notification when zFactor is changed ?

qdnguyen: @peppsac is there a copy function of OBB? what do you think if i make a copy of OBB and modify it?

peppsac: IMHO, I don't think you want to build a copy of the OBB.
You should be able to use the OBB.scale attribute to store the `zFactor` (ie: `scale = (1, 1, zFactor)`).

nguyenign: Thank @peppsac for help, i tried it before but it did not work (i have holes as you can see in image), i think i need to modify OBB max min to adapt bbox for SSE.

image

peppsac: My reasonning is the following: `OBB` is an `Object3D`. If you can't get `Object3D` properties (like scale) to work then it's a bug in `OBB`. 
And this bug needs to be fixed (instead of adding another function to `OBB`).

nguyenign: No, there is no bug and i understand what you said but I tried the scale and it did not work, because the problem is Box3D in OBB. Do not update it, the calculation of SSE will be wrong and the results as you can see in the picture above.

Btw, i also do not think the code above is wrong, because in the case there is no geoid, zFactor is always equal to 1, there is no effect on MNT. In geoid case, when zFactor is greater than 1, i saw it runs just fine as you can see in picture below.

image

nguyenign: please dont take into account my last comment, i forgot update OBB, it work now by using OBB.scale and OBB.update .

nguyenign:
ok, pr is ready to merge

src/Core/Prefab/GlobeView.js
```diff
@@ -107,7 +107,7 @@ export function createGlobeLayer(id, options) {
         }
     }
 
-    const wgs84TileLayer = new GeometryLayer(id, options.object3d || new THREE.Group());
+    const wgs84TileLayer = new GeometryLayer(id, options.object3d);
```
peppsac: Why did you changed this?

src/Core/Prefab/GlobeView.js
```diff
@@ -194,6 +194,8 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     // Setup View
     View.call(this, 'EPSG:4978', viewerDiv, options);
 
+    options.object3d = options.object3d || this.scene;
```
peppsac: Same question: why did you change this?

peppsac: I don't think you need to modify GlobeView, except for adding the `setZFactor` method.

src/Core/TileMesh.js
```diff
@@ -169,7 +169,11 @@ TileMesh.prototype.center = function center() {
 };
 
 TileMesh.prototype.OBB = function OBB() {
-    return this.geometry.OBB;
+    var zFactor = this.material.uniforms.zFactor.value;
+    var OBB = this.geometry.OBB;
+        OBB.scale.z = zFactor;
```
peppsac, line 175: Code indentation is still wrong (https://github.com/iTowns/itowns/pull/511#discussion_r145419514).
And please use `const` instead of `var`.

utils/debug/Main.js
```diff
@@ -1,4 +1,4 @@
 export { default as Debug } from './Debug';
 export { default as PointCloudDebug } from './PointCloudDebug';
 export { default as createTileDebugUI } from './TileDebug';
-export { default as create3dTilesDebugUI } from './3dTilesDebug';
+export { default as create3dTilesDebugUI } from './3dTilesDebug';
```
peppsac, line 4: You don't need to modify this file.

peppsac (changes requested):
This PR is improving, thanks.

You still need to fix a few minor issues.
Then you can rebase you commits on master and:
- keep only 2 commits:
    * 1 adding the `zFactor` feature
    * 1 adding an example
- make sure your commit wording matches the template from #518

src/Core/Prefab/GlobeView.js
nguyenign: it's changed automatically when i did rebase, i will change it back!

src/Core/TileMesh.js
nguyenign: what wrong? it work for me!

nguyenign:
ok, it's ready to merge now :)

nosy-b:
@qdnguyen Any news on that nice feature?

nguyenign:
no, as i said it's ready to merge!

nosy-b:
@qdnguyen Could you rebase?

nguyenign:
rebased!

nosy-b:
You have some lint problems
  _### 302:1  error  Trailing spaces not allowed                                   no-trailing-spaces
  303:5  error  Expected exception block, space or tab after '//' in comment  spaced-comment_

Plus, could you add the example.html in the example/index.html page with the right screenshot as it's a new feature?

gchoqueux:
moved in #1174

PR #512, opened 2017-10-06
feat (3dtiles): implement tiles cleaning

peppsac:
This commit is based on the cleaning implementation of PointCloudProcessing.

The basic idea is: tag invisible tiles and removed them after a delay.

It's a bit rough, but it's a good start to open discussion.

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -78,6 +78,9 @@ $3dTiles_Provider.prototype.removeLayer = function removeLayer() {
 
 $3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer, view, scheduler) {
     layer.sseThreshold = layer.sseThreshold || 16;
+    layer.cleanupDelay = layer.cleanupDelay || 1000;
+
+    layer._cleanableTiles = [];
     return Fetcher.json(layer.url, layer.networkOptions).then((tileset) => {
         layer.tileset = tileset;
         const urlPrefix = layer.url.slice(0, layer.url.lastIndexOf('/') + 1);
@@ -196,7 +199,6 @@ $3dTiles_Provider.prototype.pntsParse = function pntsParse(data) {
 
 function configureTile(tile, layer, metadata, parent) {
     tile.frustumCulled = false;
-    tile.loaded = true;
     tile.layer = layer.id;
 
     // parse metadata
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -75,6 +75,7 @@ function _subdivideNodeSubstractive(context, layer, node) {
                     if (node.additiveRefinement) {
                         context.view.notifyChange(true);
                     }
+                    layer.tileIndex.index[tile.tileId].loaded = true;
                 }));
         }
         Promise.all(promises).then(() => {
@@ -123,7 +124,74 @@ export function $3dTilesCulling(camera, node, tileMatrixWorld) {
     return false;
 }
 
+// Cleanup all 3dtiles|three.js starting from a given node n.
+// n's children can be of 2 types:
+//   - have a 'content' attribute -> it's a tileset and must
+//     be cleaned with cleanup3dTileset()
+//   - doesn't have 'content' -> it's a raw Object3D object,
+//     and must be cleaned with _cleanupObject3D()
+function cleanup3dTileset(layer, n, depth = 0) {
+    // If this layer is not using additive refinement, we can only
+    // clean a tile if all its neighbours are cleaned as well because
+    // a tile can only be in 2 states:
+    //   - displayed and no children displayed
+    //   - hidden and all of its children displayed
+    // So here we implement a conservative measure: if T is cleanable
+    // we actually only clean its children tiles.
+    const canCleanCompletely = n.additiveRefinement || depth > 0;
+
+    for (let i = 0; i < n.children.length; i++) {
+        // skip non-tiles elements
+        if (!n.children[i].content) {
+            if (canCleanCompletely) {
+                n.children[i].traverse(_cleanupObject3D);
+            }
+        } else {
+            cleanup3dTileset(layer, n.children[i], depth + 1);
+        }
+    }
+
+
+    if (canCleanCompletely) {
+        if (n.dispose) {
+            n.dispose();
+        }
+        delete n.content;
+        layer.tileIndex.index[n.tileId].loaded = false;
+        n.remove(...n.children);
+
+        // and finally remove from parent
+        if (depth == 0 && n.parent) {
+            n.parent.remove(n);
+        }
+    } else {
+        const tiles = n.children.filter(n => n.tileId != undefined);
+        n.remove(...tiles);
+    }
+}
+
+// This function is used to cleanup a Object3D hierarchy.
+// (no 3dtiles spectific code here because this is managed by cleanup3dTileset)
+function _cleanupObject3D(n) {
+    // all children of 'n' are raw Object3D
+    for (const child of n.children) {
+        _cleanupObject3D(child);
+    }
+    // free resources
+    if (n.material) {
+        n.material.dispose();
+    }
+    if (n.geometry) {
+        n.geometry.dispose();
+    }
+    n.remove(...n.children);
+}
+
 export function pre3dTilesUpdate(context, layer) {
+    if (!layer.visible) {
+        return [];
+    }
+
     // pre-sse
     const hypotenuse = Math.sqrt(context.camera.width * context.camera.width + context.camera.height * context.camera.height);
     const radAngle = context.camera.camera3D.fov * Math.PI / 180;
@@ -132,6 +200,23 @@ export function pre3dTilesUpdate(context, layer) {
     // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
     context.camera.preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
+
+    // once in a while, garbage collect
+    if (Math.random() > 0.98) {
+        // Make sure we don't clean root tile
+        layer.root.cleanableSince = undefined;
+
+        // Browse
+        const now = Date.now();
+
+        for (const elt of layer._cleanableTiles) {
+            if ((now - elt.cleanableSince) > layer.cleanupDelay) {
+                cleanup3dTileset(layer, elt);
+            }
+        }
+        layer._cleanableTiles = layer._cleanableTiles.filter(n => (layer.tileIndex.index[n.tileId].loaded && n.cleanableSince));
+    }
+
     return [layer.root];
 }
 
@@ -173,6 +258,7 @@ export function init3dTilesLayer(view, scheduler, layer) {
                 delete layer.tileset;
                 layer.object3d.add(tile);
                 tile.updateMatrixWorld();
+                layer.tileIndex.index[tile.tileId].loaded = true;
                 layer.root = tile;
             });
 }
@@ -183,6 +269,13 @@ function setDisplayed(node, display) {
     // node.material.visible
     if (node.content) {
         node.content.visible = display;
+    }
+}
+
+function markForDeletion(layer, elt) {
+    if (!elt.cleanableSince) {
+        elt.cleanableSince = Date.now();
+        layer._cleanableTiles.push(elt);
     }
 }
 
@@ -198,9 +291,11 @@ export function process3dTilesNode(cullingTest, subdivisionTest) {
         const isVisible = cullingTest ? (!cullingTest(context.camera, node, node.matrixWorld)) : true;
         node.visible = isVisible;
 
-        let returnValue;
 
         if (isVisible) {
+            node.cleanableSince = undefined;
+
+            let returnValue;
             if (node.pendingSubdivision || subdivisionTest(context, layer, node)) {
                 subdivideNode(context, layer, node, cullingTest);
                 // display iff children aren't ready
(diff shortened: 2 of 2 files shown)


src/Process/3dTilesProcessing.js
```diff
@@ -80,6 +80,10 @@ export function $3dTilesCulling(node, camera) {
 }
 
 export function pre3dTilesUpdate(context, layer) {
```
Jeremy-Gaillard: It's unfortunate that we can only clean when needsUpdate is true, and not in iTowns' downtime. But this would require some architectural changes.

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +92,49 @@ export function pre3dTilesUpdate(context, layer) {
     // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
     context.camera.preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
+
+    // once in a while, garbage collect
+    if (Math.random() < 0.98) {
```
Jeremy-Gaillard: Is there a reason for using a random test instead of a fixed waiting time?

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +92,49 @@ export function pre3dTilesUpdate(context, layer) {
...
+        const now = Date.now();
+        layer.root.traverse((n) => {
+            if (n.layer != layer.id) {
+                return;
+            }
+
+            // Browse children, and remove old ones before
+            // 'traverse' method meets them.
+            for (let i=0; i<n.children.length; i++) {
+                const c =n.children[i];
```
Jeremy-Gaillard: Missing space

src/Process/3dTilesProcessing.js
```diff
@@ -80,6 +80,10 @@ export function $3dTilesCulling(node, camera) {
 }
 
 export function pre3dTilesUpdate(context, layer) {
+    if (!layer.visible) {
```
Jeremy-Gaillard: I'm wondering if we should still clean the layer if the layer is not visible. If I'm not mistaken, tiles that were visible before hiding the layer should not have been marked for deletion, so we avoid removing all the tiles inadvertently.

src/Process/3dTilesProcessing.js
peppsac: Nope, the random() solution allows to not maintain a state (`lastCleanupTime` or similar).

Another option would be to maintain a `cleanable` array with all potentially removable objects are referenced.
This array could be browsable at each update loop easily.

src/Process/3dTilesProcessing.js
peppsac: Yeah, this could be an improvment.. I'll keep this in mind for a future PR

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +100,51 @@ export function pre3dTilesUpdate(context, layer) {
     // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
     context.camera.preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
+
+    // once in a while, garbage collect
```
autra, line 204: If I read correctly, this `once in a while` is actually 98% of the time ?

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +100,51 @@ export function pre3dTilesUpdate(context, layer) {
...
+                            o.geometry.dispose();
+                        }
+                        // we can't remove 'o' from 'c' yet,
+                        // because 'traverse' first applies the callback to a node
+                        // and then to its children.
+                    });
+
+                    // remove c children recursively
+                    removeChildren(c);
+                    delete c.content;
```
autra: AFAIK `c.content` can actually be a `Scene` object (when it is a parsed gltf for instance), and this scene contains *another* hierarchy of 3D objects you need to clean...

autra: You want to add this logic to `removeChildren` too I guess

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +100,51 @@ export function pre3dTilesUpdate(context, layer) {
...
+                    n.remove(c);
+                    i--;
```
autra: Can we use a descendant for loop instead? Or  a `while (i--)`?

src/Process/3dTilesProcessing.js
```diff
@@ -165,13 +234,17 @@ export function process3dTilesNode(cullingTest, subdivisionTest) {
             if ((node.material === undefined || node.material.visible)) {
```
autra: Unrelated to current PR, but why do we set `visible = false` for all children (on line 236) when `node.material.visible == true` ? I would have expected the contrary.

autra (changes requested):

src/Process/3dTilesProcessing.js
peppsac: Yep... I'll fix this typo :)

Do you think we can keep this random based version for now, or should we move to another mechanism?

src/Process/3dTilesProcessing.js
peppsac: (I refactored this part a bit)
But even with the current version, the full hierarchy should be traversed for `dispose()` and then removed.

src/Process/3dTilesProcessing.js
peppsac: This code looks really fishy...
@Jeremy-Gaillard can you explain what is this code doing?

src/Process/3dTilesProcessing.js
autra: Both : we can keep that as a first experiment, but we should think about a garbage-collecting mechanism, possibly unified for every type of layers. 

Some thoughts:
- we should do a full garbage collect when the user does nothing, meaning after we have stopped the update loop (with a 200ms timeout or something similar).
- maybe it's not enough (because if the user keeps moving, we still want to clean from time to time). In this case, we might want to clean regularly, while keeping the load down. Maybe clean only things that have been hidden for a long time, or very far from being visible again? But how to calculate this metrics? Or maybe a time-based metric: we measure previous garbage-collecting process, and depending on the time it took, we define a number of objects/layers/things we can clean at each update without affecting frame rate too much.

src/Process/3dTilesProcessing.js
Jeremy-Gaillard: This is strange indeed. `node.material` is always undefined, so the condition is useless.

Logically, we should be hiding only the nodes that we will not be visiting. So, the code from line 235 to 240 should be at line 231. I tested it on your branch and it works, I'll push a fix.

peppsac:
ping: PR updated.

src/Process/3dTilesProcessing.js
```diff
@@ -79,7 +79,51 @@ export function $3dTilesCulling(node, camera) {
...
+        n.dispose();
+    }
+    delete n.content;
+    layer.tileIndex.index[n.tileId].loaded = false;
+    n.remove(...n.children);
+}
+
+// This function is used to cleanup a Object3D hierarchy.
+// (no 3dtiles spectific code here because this is managed by cleanup3dTileset)
+function _cleanupObject3D(n) {
```
autra, line 175: Nits: we can drop the _ prefix here.

src/Process/3dTilesProcessing.js
```diff
@@ -79,7 +79,51 @@ export function $3dTilesCulling(node, camera) {
...
+            n.children[i].traverse(_cleanupObject3D);
+        } else {
+            cleanup3dTileset(layer, n.children[i]);
+        }
+    }
+    if (n.dispose) {
+        n.dispose();
+    }
+    delete n.content;
+    layer.tileIndex.index[n.tileId].loaded = false;
```
Jeremy-Gaillard: `loaded` is an attribute of the tile. By going through the tileIndex, you get the tile metadata (a portion of the tileset.json). I presume this is not intended?

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +132,20 @@ export function pre3dTilesUpdate(context, layer) {
...
+    if (Math.random() > 0.98) {
+        // Browse
+        const now = Date.now();
+
+        for (const elt of layer._cleanableTiles) {
+            if ((now - elt.cleanableSince) > (layer.cleanupDelay || 1000)) {
+                cleanup3dTileset(layer, elt);
+            }
+        }
+        layer._cleanableTiles = layer._cleanableTiles.filter(n => (layer.tileIndex.index[n.tileId].loaded && n.cleanableSince));
```
Jeremy-Gaillard: See previous comment. The filter condition is always false, since `layer.tileIndex.index[n.tileId].loaded` is either undefined or false.

Jeremy-Gaillard (changes requested):

src/Process/3dTilesProcessing.js
Jeremy-Gaillard: Ok, saw your other PR, this one should come afterwards.

peppsac: Ah yes I forgot to port the `.loaded = true` to this PR.

src/Process/3dTilesProcessing.js
```diff
@@ -79,7 +79,51 @@ export function $3dTilesCulling(node, camera) {
...
+            cleanup3dTileset(layer, n.children[i]);
+        }
+    }
+    if (n.dispose) {
+        n.dispose();
+    }
+    delete n.content;
+    layer.tileIndex.index[n.tileId].loaded = false;
+    n.remove(...n.children);
+}
```
autra: Don't you want to call `cleanupObject3D` on n too'?

src/Process/3dTilesProcessing.js
```diff
@@ -88,6 +132,20 @@ export function pre3dTilesUpdate(context, layer) {
...
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
     context.camera.preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
+
+    // once in a while, garbage collect
+    if (Math.random() > 0.98) {
+        // Browse
+        const now = Date.now();
+
+        for (const elt of layer._cleanableTiles) {
+            if ((now - elt.cleanableSince) > (layer.cleanupDelay || 1000)) {
```
autra: Can we make this option more discoverable? (document it... somewhere?)

autra (commented):
Submitting some nits now, before you address @Jeremy-Gaillard review.

src/Process/3dTilesProcessing.js
autra: So I guess the question is: where do we document options that are specific for one type of layers?

I guess in this case, it should be it the `Processing` module. Maybe we can document what layer options this `Processing` file use on top of it?

src/Process/3dTilesProcessing.js
peppsac: The `_` is to indicate that this a `private` function within this module: you're not supposed to use it except from `cleanup3dTileset`.
But I can remove it if you feel it's useless.

src/Process/3dTilesProcessing.js
peppsac: No, tiles don't have geometry or material. They're only a container for mesh (stored in children and content) and other tiles (in children).

src/Process/3dTilesProcessing.js
autra: As you're not exposing it, it can't be used outside of this module anyway. For me the `_` prefix is useful when you have to attach it to `this` for some reason (I don't think it's ever *strictly* necessary to attach things to `this`, but from time to time, it saves a lot of headache to do so).

src/Process/3dTilesProcessing.js
peppsac: This is an important comment, but IMHO we should make a specific issue/PR to properly document all layers options.

Jeremy-Gaillard: We usually initialize these kind of options in preprocessDataLayer, which could be a first step in making this parameter more discoverable.

peppsac: Followed your suggestion and moved this to `preprocessDataLayer`

src/Process/3dTilesProcessing.js
peppsac: Fixed.

peppsac:
Updated with one fix: I forgot to remove tiles from their parent when removed.

src/Process/3dTilesProcessing.js
```diff
@@ -79,7 +80,74 @@ export function $3dTilesCulling(node, camera) {
...
+//     and must be cleaned with _cleanupObject3D()
+function cleanup3dTileset(layer, n, depth = 0) {
+    // If this layer is not using additive refinement, we can only
+    // clean a tile if all its neighbours are cleaned as well because
+    // a tile can only be in 2 states:
+    //   - displayed and no children displayed
+    //   - hidden and all of its children displayed
+    // So here we implement a conservative measure: if T is cleanable
+    // we actually only clean its children tiles.
+    const canCleanCompletely = layer.additiveRefinement || depth > 0;
```
Jeremy-Gaillard: Additive refinement is handled at the tile level. You can have different type of refinement in the same layer.

```js
n.additiveRefinement
```

Jeremy-Gaillard (changes requested):
Looks good. Only one small fix before merging.

src/Process/3dTilesProcessing.js
peppsac: Thanks fixed.

Jeremy-Gaillard (approved):

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -78,6 +78,9 @@ $3dTiles_Provider.prototype.removeLayer = function removeLayer() {
 
 $3dTiles_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer, view, scheduler) {
     layer.sseThreshold = layer.sseThreshold || 16;
+    layer.cleanupDelay = layer.cleanupDelay || 1000;
```
autra, line 81: +1 much better,  until we have some docs

autra (approved):

Issue #478, opened 2017-09-14
EPSG:3857 (PseudoMercator) projection not handled for WMS layers


elias75015:
## Context
WMS layers can't be added in PseudoMercator (EPSG:3857), but only in WGS84 (EPSG:4326). 

If we change the projection property of a WMS layer object to "EPSG:3857", the following error is launched and the WMS layer isn't displayed :

> non-coherent result 0 0 vs 1. [object Object]

On the other hand, there is no problem with WMTS layers added in EPSG:3857.

## Steps to Reproduce
1. Change the projection property of a WMS layer to "EPSG:3857" (itowns/examples/layers/JSONLayers/Region.json for instance)
2. Add it to the globe (see itowns/index.html)

## Expected Behavior
The region layer is displayed.

## Actual Behavior
The region layer is not display and error is launched in the dev console.

peppsac:
FYI I've started working on this.
I'll also fix the `split.html` example in the process...

elias75015:
Hello,

in your opinion, is it possible for the patch to be delivered within the next 2 weeks ?

peppsac:
Hu. No idea, I didn't had any time to continue my work on this for the past 2 weeks.

I pushed my current branch here: https://github.com/iTowns/itowns/tree/fix_tms

There are some oddities in `OGCWebServierHelper.computeTMSCoordinates` that I don't understand yet  and at least another issue regarding texture count (see `TileFS.glsl` hack-change).

Feel free to push more commits to improve this branch.

Edit: forgot to write that both `split.html` and `orthographic.html` work on this branch...

PR #513, opened 2017-10-06
Tile geometry projection support

peppsac:
This PR improves itowns handling of having multiple tile geometries with different CRS.

Until now, geometries are *assumed* to use the same CRS as the view which is often the case.
I believe we should support displaying other CRS too, as long as the transformation from 1 crs to the other is possible.
If not possible or not implemented we should fail early instead of silently accepting data that we can't display.

With that being said, this PR implements tile geometry projection and adds a smallish example to illustrate a possible use.
The example could be improved to look nicer and allow the user to type any CRS instead of hardcoding it: help is welcome, feel free to push improvements directly to the branch :-)

Here's the result with CRS 5514 (see https://epsg.io/5514):
capture d ecran de 2017-10-06 17-26-36

### DIFF
--- a/src/Core/Prefab/Planar/PlanarTileBuilder.js
+++ b/src/Core/Prefab/Planar/PlanarTileBuilder.js
@@ -16,21 +16,27 @@ PlanarTileBuilder.prototype.constructor = PlanarTileBuilder;
 // init projected object -> params.projected
 PlanarTileBuilder.prototype.Prepare = function Prepare(params) {
     params.nbRow = Math.pow(2.0, params.zoom + 1.0);
-    params.projected = new THREE.Vector3();
+    params.projected = new Coordinates(params.extent.crs(), 0, 0, 0);
 };
 
 
 // get center tile in cartesian 3D
 PlanarTileBuilder.prototype.Center = function Center(params) {
     params.extent.center(this.tmp.coords);
-    params.center = new THREE.Vector3(this.tmp.coords.x(), this.tmp.coords.y(), 0);
+    if (params.extent.crs() !== params.crs) {
+        this.tmp.coords = this.tmp.coords.as(params.crs);
+    }
+    params.center = this.tmp.coords.xyz();
     return params.center;
 };
 
 // get position 3D cartesian
 PlanarTileBuilder.prototype.VertexPosition = function VertexPosition(params) {
-    this.tmp.position.set(params.projected.x, params.projected.y, 0);
-    return this.tmp.position;
+    if (params.extent.crs() !== params.crs) {
+        return params.projected.as(params.crs).xyz();
+    } else {
+        return params.projected.xyz();
+    }
 };
 
 // get normal for last vertex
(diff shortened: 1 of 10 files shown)


gchoqueux:
there's a problem with picking mouse.
the planar layer `epsgLayer` must have the same THREE layer, for the `getPickingPositionFromDepth` to function

gchoqueux:
in example with globeView
`createPlanarLayer` return geometry layer which is not plan, because it is converted into the crs of the globe. On the other hand the layer is instantiated with a planar projection.
it could also be created with `createGlobeLayer`?
if yes it would be necessary to factorize to have a single function
WDYT?

peppsac:
> in example with globeView createPlanarLayer return geometry layer which is not plan, because it is converted into the crs of the globe. 

This geometry layer uses a planar CRS but it is displayed in a the globe's CRS so its underlying geometry uses `EPSG:4978`.
So IMHO, it makes sense to create it using `createPlanarLayer`.

gchoqueux:
`createPlanarLayer` causes tile visibility problems in the globe

gchoqueux:
is it possible to attach a imagery layer CRS 5514 to see the result?

peppsac:
Updated, with OSM data displayed on the projected layer.

capture d ecran de 2017-10-20 14-42-22

gchoqueux:
the projection is incorrect

peppsac:
> the projection is incorrect

Could you elaborate please?

From my perspective there's a visual artifact but maybe it's due to the 3857 CRS being used for the OpenStreepMap layer (using the fix from #478 ). Or are talking about something else?

gchoqueux:
I can't make the example work. the conversion projection are not compatible.
Could you check?

peppsac:
I'm not sure what's wrong here: #478 is supposed to fix 3857 for WMS layer and this PR (try to) use this feature.

gchoqueux:
how use PR's example?

peppsac:
> how use PR's example?

Open `epsg_finder.html`, click on 'View'.
The result should be similar to https://github.com/iTowns/itowns/pull/513#issuecomment-338195874

gchoqueux:
there are errors of the conversions of coordinates, when I try example

peppsac:
> there are errors of the conversions of coordinates, when I try example

That means you managed to reproduce the problem.
Now what I'd need is your input on how you think this could get fixed.

gchoqueux:
it's worse than that, I don't see the planar layer

peppsac:
So this PR depends on PR #561 to fix a `TileGeometry` bug.

src/Core/Prefab/Planar/PlanarTileBuilder.js
```diff
@@ -40,17 +46,33 @@ PlanarTileBuilder.prototype.VertexNormal = function VertexNormal() {
 
 // coord u tile to projected
 PlanarTileBuilder.prototype.uProjecte = function uProjecte(u, params) {
-    params.projected.x = params.extent.west() + u * (params.extent.east() - params.extent.west());
+    params.projected._values[2] = 100;
```
zarov, line 49: Couldn't you use `Coordinates#setAltitude` ? I'm not fond of accessing properties that are supposed to be private. Same question for next lines.

zarov (commented):
Could you rebase your work ? Also any view on how to fix the bad coordinates conversion ?

peppsac:
I'm closing this PR for now - I'll reopen it when I'll have some time to fix the remaining issues.

PR #514, opened 2017-10-06
First person control fix

autra:
## Description

Some fixes and refacto in  FirstPersonControls. 

This is also a first step in making FirstPersonControls and FlyControls as alike as possible

### DIFF
--- a/src/Renderer/ThreeExtended/FirstPersonControls.js
+++ b/src/Renderer/ThreeExtended/FirstPersonControls.js
@@ -20,7 +20,7 @@ function onDocumentMouseMove(event) {
         // (where deltaH is the vertical amount we moved, and H the renderer height)
         // we loosely approximate tan(x) by x
         const pxToAngleRatio = THREE.Math.degToRad(this.camera.fov) / this.view.mainLoop.gfxEngine.height;
-        this.camera.rotation.z = (this._onMouseDownMouseX - event.clientX) * pxToAngleRatio + this._onMouseDownRotZ;
+        this.camera.rotation.z = (event.clientX - this._onMouseDownMouseX) * pxToAngleRatio + this._onMouseDownRotZ;
         this.camera.rotation.x = (event.clientY - this._onMouseDownMouseY) * pxToAngleRatio + this._onMouseDownRotX;
         this.view.notifyChange(false);
     }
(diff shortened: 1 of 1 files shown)


src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -12,15 +10,18 @@ function onDocumentMouseDown(event) {
...
     if (this._isUserInteracting === true) {
-        const fovCorrection = this.camera.fov / MAX_FOV; // 1 at MAX_FOV
-        this._phi = (this._onMouseDownMouseX - event.clientX) * 0.13 * fovCorrection + this._onMouseDownPhi;
-        this._theta = (event.clientY - this._onMouseDownMouseY) * 0.13 * fovCorrection + this._onMouseDownTheta;
+        // in rigor we have tan(theta) = tan(cameraFOV) * deltaH / H
+        // (where deltaH is the vertical amount we moved, and H the renderer height)
+        // we loosely approximate tan(x) by x
+        const pxToAngleRatio = THREE.Math.degToRad(this.camera.fov) / this.view.mainLoop.gfxEngine.height;
+        this.camera.rotation.z = (event.clientX - this._onMouseDownMouseX) * pxToAngleRatio + this._onMouseDownRotZ;
+        this.camera.rotation.x = (event.clientY - this._onMouseDownMouseY) * pxToAngleRatio + this._onMouseDownRotX;
```
peppsac, line 24: I'm not sure about this change. I kinda liked that `camera.matrix` was only modified only in the `mainLoop.update` process.

autra: On the other hand, can we require every FrameRequesters not to touch the `camera.matrix` either directly or indirectly? Seems a hard requirement to me. Does this already cause trouble?

peppsac: This code is not run in a FrameRequester context, but in a `mouse-move` event.
Anyway, the real issue is to not modify the camera.matrixWorld between the update and render and this shouldn't happen because events are queued after the current script execution (right?).

So we can probably merge this and gets the benefit of a simplified code.

autra: > This code is not run in a FrameRequester context, but in a mouse-move event

Yes sorry, I wasn't clear: can we require client code (whether `GlobeControls` or anything) not to modify the camera outside the update? 

> not modify the camera.matrixWorld between the update and render and this shouldn't happen because events are queued after the current script execution (right?).

Not sure to understand this, setting rotation.{x,y,z} does not modify camera.matrixWorld right?

autra:
@peppsac commit messages updated.

peppsac (approved):

PR #515, opened 2017-10-06
Fix (controls): compute more correcly rotations in FlyControls

autra:
## Description
Use a better (albeit still approximate) heuristic to calculate horizontal and vertical movement with mouse drag.

## Motivation and Context

2 advantages to this patch:
- it unifies things with FirstPersonControls
- it removes a magic and opaque constant

### DIFF
--- a/src/Renderer/ThreeExtended/FlyControls.js
+++ b/src/Renderer/ThreeExtended/FlyControls.js
@@ -1,8 +1,5 @@
 import * as THREE from 'three';
 
-const maxFOV = 90;
-const rotationFactor = 0.0022; // determined empirically
-
 const MOVEMENTS = {
     38: { method: 'translateZ', sign: -1 }, // FORWARD: up key
     40: { method: 'translateZ', sign: 1 }, // BACKWARD: down key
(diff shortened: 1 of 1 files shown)


peppsac (commented):
Does this PR change the camera behavior (what do you mean by "more correctly")?
Also your commit message is empty (it should contain the text of youy PR).

autra:
> what do you mean by "more correctly"

I'm probably wrong to say that. I thought the old behaviour was incorrect for screen size different from mine, because this size wasn't taken into account in the calculation, but actually I was wrong, because the new calculus is *also* independant from screen size: `pointerX - this._onMouseDownMouseX` can actually be seen as a "percentage of screen I moved", which would be for vertical displacement for instance "mouse drag in screen percentage * screen height in px".

So if I calculate this constant, it is equivalent of having a magic number of `0.0015` with the old code. I believe the new code is a better approximation, partly also because I understand it better. 

Not sure if this really answer your question though. I need to change my commit message, it seems.

autra:
@peppsac Commit message changed.

peppsac (approved):
Thanks :)

Issue #518, opened 2017-10-09
Adopt angular convention for git commit messages


peppsac:
As we adopted a code style for the js code, I'd like to explicitely adopt a style for git messages.

IMHO the style proposed at https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md is good and has an extra benefit: it allows to simplify changelog generation.

The important bits are:
- the template to follow:
```
<type>(<scope>): <subject>
<BLANK LINE>
<body>
<BLANK LINE>
<footer>
```
- the allowed types:
  * feat, perf, fix => these 3 are special because they are pickup when building the changelog
  * docs, chore, style, refactor, test

Last but not least, we should choose a set of allowed scopes to allow grouping of relevant changes in the changelog.
~Here's a first proposal~
Second version:
- core (when modifying MainLoop / View / core logic)
- gis (Coordinates, Extent, Ellipsoid)
- controls (GlobeControls, etc)
- ~process (or update) (when modifiying TileNodeProcessing..)~
- ~providers~
- examples
- protocols (wfs, wmts, etc changes)
- 3dtiles
- pointcloud
- vector (not using 'feature' here because it's already used for 'type' prefix)
- geometry (e.g when adding features to TileVS)
- debug

peppsac:
(just edited the first post)

zarov:
Good idea, I propose to add a file at the root of the repo, similar as this one https://gist.github.com/adeekshith/cd4c95a064977cdc6c50.

By doing `git config commit.template .git-commit-template.txt`, people will have the template with explanations on how to write their commit message, when doing a `git commit`. Maybe we could "force it" in a npm task ? Like a `init` (in `scripts` section) that would do this command, or at the beginning of `start` (but ugly...).

autra:
Or, as a first step, we could start by amending https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md. 

I'm all to have a commit template file in the repo, and have the `git config` documented in CONTRIBUTING.md, but I'm not comfortable having it installed automatically (I do already use a custom commit template). That being said, if the majority of you want it installed automatically, I'll live with it.

tbroyer:
+1 to the commit template, -1 to installing it automatically.

Would it be possible to have a check in Travis for PRs targeting the master branch? (`$TRAVIS_BRANCH == "master") or possibly for all PRs. Similar to https://github.com/angular/angular.js/blob/master/scripts/travis/build.sh#L22-L28

[...]

PR #519, opened 2017-10-09
Feature geometry extruded

gliegard:
## Description
- Add style in the layer, to specify how to build mesh from feature.
- In style, add extrude function to get extrusion amount from WFS server.
- Update the example (WFS on planar), to have polygon extruded.
- Add an example : WFS extruded on the globe.

## Motivation and Context
This change allow the user to specify the way feature from WFS is treated. Feature polygons can be extruded. It works in planar mode and in the globe.

screenshot-pr-feature-extruded3


src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -42,49 +140,77 @@ function coordinatesToMesh(coordinates) {
...
             /* eslint-disable guard-for-in */
             for (const id in coordinates.featureVertices) {
-                const polygon = coordinates.featureVertices[id];
-                const contour = vertices.slice(polygon.offset * 3, (polygon.offset + polygon.count) * 3);
-                const triangles = Earcut(contour, null, 3);
+                // extract contour coodinates and properties of one feature
+                const { contour, propertie } = extractFeature(coordinates, properties, id);
+                // get altitude and extrude amount from properties
+                const altitudeBottom = style.altitude ? style.altitude(propertie) : 0;
+                const extrudeAmount = style.extrude ? style.extrude(propertie) : 0;
```
peppsac: I'm not a huge fan of this design. Here's what I had in mind:
- add a callback in the layer, called for each new feature. Actually this callback exists - but is misnamed `colorFunction`
- expose some helper functions to perform feature polygon extrusion easily



This way, you won't have to complexify even more this part and the user is free to implement whatever feature.

gliegard: I don't think it's a good idea to do the extrusion in a callback because
- do the extrusion means creating a mesh
- at this point (in the callback), the mesh is already created. So I will have to destroy the mesh and create another one ?? It's not a good idea.
- and Feature2Mesh seems to be the good place to convert a feature to a mesh.

If you think it's complex, I can extract methods.

gliegard: I don't think it's a good idea to do the extrusion in a callback because
- do the extrusion means creating a mesh
- at this point (in the callback), the mesh is already created. So I will have to destroy the mesh and create another one ?? It's not a good idea.
- and Feature2Mesh seems to be the good place to convert a feature to a mesh.

If you think it's complex, I can extract methods.

gliegard: I don't think it's a good idea to do the extrusion in a callback because
- do the extrusion means creating a mesh.
- at this point (in the callback), the mesh is already created. So I will have to destroy the mesh and create another one ?
- and Feature2Mesh seems to be the good place to convert a feature to a mesh.

If you think it's complex, I can extract methods.

peppsac: The problem is you're burying deep inside iTowns something the user might want to control precisely.
IMHO the callback should be right here:
https://github.com/iTowns/itowns/pull/519/files#diff-b1820acc0c4f07c7e51a605b08ae2cc1R101

This code could become something like:
```js
callback(GeoJSON2Feature.parse(crs, geojson, tile.extent));
```

And then the user could do whatever fits its usecase:
- convert to 2D mesh (using e.g `Feature2Mesh.convert`)
- convert to extruded mesh (using a new function: `Feature2Mesh.convertAndExtrude` or whatever)
- something entirely different (build its own meshes, filter heavily, etc etc)

gliegard: ah ok, I understand. I try this approach.

gliegard: I've updated the code to extract a callback to choose extrusion of not.
But I still need to get altitude and and extrude amount from properties.

gliegard:
@itowns-reviewers

examples/globe_wfs_extruded.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    itowns.FeatureProcessing.assignColorsToFeatureCollection(
+        featureCollection, featureCollection.children[0], colors);
+
+    featureCollection.children[0].material.linewidth = 5;
+}
+
+globeView.addLayer({
+    update: itowns.FeatureProcessing.update(colorFunctionLine),
```
peppsac: `colorFunctionLine` is used once when creating the mesh so `colorFunctionLine` should be an argument of `convertAndExtrude`.

examples/globe_wfs_extruded.js
```diff
@@ -0,0 +1,122 @@
...
+    convert: itowns.Feature2Mesh.convertAndExtrude,
+    url: 'https://download.data.grandlyon.com/wfs/rdata?',
+    protocol: 'wfs',
+    version: '2.0.0',
+    id: 'tcl_bus',
+    typeName: 'tcl_sytral.tcllignebus',
+    level: 14,
+    projection: 'EPSG:3946',
+    style: {
+        altitude: function altitude() { return 180; },
```
peppsac: `altitude` property seems out of place in `style`. Maybe this function should be a parameter of `convertAndExtrude` instead.

examples/globe_wfs_extruded.js
```diff
@@ -0,0 +1,122 @@
...
+
+    itowns.FeatureProcessing.assignColorsToFeatureCollection(
+        featureCollection, featureCollection.children[0], colors);
+
+    featureCollection.children[0].material.linewidth = 5;
+}
+
+globeView.addLayer({
+    update: itowns.FeatureProcessing.update(colorFunctionLine),
+    convert: itowns.Feature2Mesh.convertAndExtrude,
```
peppsac: Do you really extrude line as well?

examples/globe_wfs_extruded.js
peppsac: And of course, the same comment is valid for `extrude`

src/Core/Scheduler/Providers/WFS_Provider.js
```diff
@@ -98,7 +97,8 @@ WFS_Provider.prototype.getFeatures = function getFeatures(crs, tile, layer) {
     if (result.feature !== undefined) {
         return Promise.resolve(result);
     }
-    return Fetcher.json(url, layer.networkOptions).then(geojson => assignLayer(Feature2Mesh.convert(GeoJSON2Features.parse(crs, geojson, tile.extent)), layer));
+
+    return Fetcher.json(url, layer.networkOptions).then(geojson => assignLayer(layer.convert(GeoJSON2Features.parse(crs, geojson, tile.extent), layer.style), layer));
```
peppsac: You're making `layer.convert` mandatory: you should check for its existence in preprocessLayer or define a sensible default value.

peppsac (commented):
Thanks for the update.

This review is focused on the API changes.

examples/globe_wfs_extruded.js
gliegard: Nope, only polygons

examples/globe_wfs_extruded.js
gliegard: This code come from example/wfs.js. 
Creating the mesh and coloring the mesh is separated in the code.
I don't want to refactor / re design this part of iTowns now.

peppsac: This `colorFunctionLine` stuff was written under this simple assumption: the geojson -> mesh feature wouldn't not be customisable, except for the color part.

Now you're breaking this assumption: this is a good thing, but you can't avoid refactoring the code that now doesn't make sense.

The color selection code is only used at mesh creation time which is now the responsability of your `convert` function: you should adapt the old code to the new feature.

peppsac (changes requested):
This PR needs some change, see: https://github.com/iTowns/itowns/pull/519#discussion_r146345705

examples/globe_wfs_extruded.js
gliegard: I understand that, it's a good idea, but for me, it's a big work to refactor that.
Can you be more precise on the refactor you want ?

peppsac: Move the mesh creation dependencies from `update` to `convert`.
So your `convert` function could receive an object with properties allowing the user to select the value of these attributes:
- color: could be either a THREE.Color or a function or undefined (= you pick a random color for a given layer)
- altitude: could be a constant, a function or undefined (= 0)
- extrude: could be a constant, a function or undefined (= no extrusion)

I don't expect this code shuffling to be a big refacto and it should greatly improve the API.

gliegard: I 've update the code.

Now the color function is very easy to read and write !!

But now, the material is not accessible for the user. 
For example, for the Lyon bus lines, I had to put the material change in Feature2Mesh.
https://github.com/iTowns/itowns/pull/519/files#diff-2a78502be7a11a150dcce98113215193R145

So, I plan to add another callback to update the material (I'll do it tomorrow).
Good idea ? @peppsac

src/Process/FeatureProcessing.js
```diff
@@ -23,7 +23,7 @@ function applyOffset(obj, offset) {
 }
 
 export default {
-    update(colorFunction) {
+    update() {
```
peppsac: You can remove this closure now.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,346 @@
...
+
 export default {
-    convert(feature) {
-        if (feature.geometries) {
-            return featureCollectionToThree(feature);
-        } else {
-            return featureToThree(feature);
-        }
+
+    convertNoElevation() {
```
peppsac: IMHO, you could have a single entry point here: `Feature2Mesh.convert` with a single `options` parameter.
See https://github.com/iTowns/itowns/pull/519#discussion_r146487033 

Example usages:
```js
Feature2Mesh.convert({ color: new THREE.Color(0xff0000), altitude: myAltitudeFunction });
Feature2Mesh.convert();
Feature2Mesh.convert({ color: myColorFunction, altitude: 150, extrude: myExtrudeFn });
```

examples/wfs.js
```diff
@@ -123,6 +106,9 @@ view.addLayer({
         south: 45.437,
         north: 46.03,
     },
+    style: {
+        extrude: function extrude(properties) { return properties.hauteur; },
```
peppsac: This is useless now, right?

examples/globe_wfs_extruded.js
peppsac: For tiles and points we already have a callback: https://github.com/iTowns/itowns/blob/master/src/Process/TiledNodeProcessing.js#L47 and https://github.com/iTowns/itowns/blob/master/src/Process/PointCloudProcessing.js#L236 so it would make sense to add the same thing for FeatureProcessing.

On the other hand, you can also add another argument to your convert function: `material` that could be either:
- a `Material` (like we do for 3dtiles: https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Providers/3dTiles_Provider.js#L170)
- or a collection of material property names, which could either constant or functions. For instance:
```js
convert({ material: new THREE.LineDashedMaterial() });
// or
convert({ material: { linewidth: 5, linecap: myLineCapFunction } });
```
WDYT?

src/Process/FeatureProcessing.js
gliegard: Sorry, I don't understand.

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: ok, good idea.

src/Process/FeatureProcessing.js
peppsac:
```js
 export default {
    update(context, layer, node) {
        if (!node.parent && node.children.length) {
        ...
```

gliegard:
updated

examples/wfs.js
```diff
@@ -43,32 +43,27 @@ view.camera.camera3D.position.set(1839739, 5171618, 910);
...
+    return new itowns.THREE.Color(rgb[0] / 255, rgb[1] / 255, rgb[2] / 255);
 }
 
 view.addLayer({
-    update: itowns.FeatureProcessing.update(colorFunctionLine),
+    update: itowns.FeatureProcessing.update,
+    convert: itowns.Feature2Mesh.convert({
+        onMeshCreated: setMaterialLineWith,
+        color: colorLine,
+        altitude: undefined,
```
peppsac: You don't need to explicitely set altitude to ` undefined`: you can just omit altitude.

examples/wfs.js
```diff
@@ -87,29 +82,27 @@ view.addLayer({
...
+function extrudeBuildings(properties) {
+    return properties.hauteur;
 }
 
 view.addLayer({
-    update: itowns.FeatureProcessing.update(colorFunctionBuildings),
+    type: 'geometry',
+    update: itowns.FeatureProcessing.update,
+    convert: itowns.Feature2Mesh.convert({
+        onMeshCreated: undefined,
```
peppsac: The `onMeshCreated` property could be a top level layer property: that's the way it's done for points and tiles.

src/Core/Scheduler/Providers/WFS_Provider.js
```diff
@@ -98,7 +98,11 @@ WFS_Provider.prototype.getFeatures = function getFeatures(crs, tile, layer) {
     if (result.feature !== undefined) {
         return Promise.resolve(result);
     }
-    return Fetcher.json(url, layer.networkOptions).then(geojson => assignLayer(Feature2Mesh.convert(GeoJSON2Features.parse(crs, geojson, tile.extent)), layer));
+
+    const defaultCallbacks = { onMeshCreated: undefined, color: undefined, altitude: undefined, extrude: undefined };
```
peppsac: You can remove `defaultCallbacks`. And modify ` Feature2Mesh.convert` signature to be: `Feature2Mesh.convert(options = {})`

src/Process/FeatureProcessing.js
```diff
@@ -23,105 +23,77 @@ function applyOffset(obj, offset) {
...
-            if (featureVertices) {
-                for (let j = 0; j < featureVertices.count; j++) {
-                    const baseIdx = 3 * (featureVertices.offset + j);
-                    colorAttribute[baseIdx + 0] = colors[i].r * 255;
-                    colorAttribute[baseIdx + 1] = colors[i].g * 255;
-                    colorAttribute[baseIdx + 2] = colors[i].b * 255;
+        context.scheduler.execute(command).then((result) => {
+            // if request return empty json, WFS_Provider.getFeatures return undefined
+            if (result) {
+                node.layerUpdateState[layer.id].success();
```
peppsac: Here is a good place to call `layer.onMeshCreated`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,345 @@
 import * as THREE from 'three';
 import Earcut from 'earcut';
 
-function coordinatesToMesh(coordinates) {
-    if (!coordinates) {
-        return;
+function isFunction(functionToCheck) {
+    var getType = {};
+    return functionToCheck && getType.toString.call(functionToCheck) === '[object Function]';
```
peppsac: Couldn't this be simplified as `typeof(functionToCheck) === 'function'` ?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,345 @@
...
 
 export default {
-    convert(feature) {
-        if (feature.geometries) {
-            return featureCollectionToThree(feature);
-        } else {
-            return featureToThree(feature);
-        }
+
+    convert(callbacks) {
```
peppsac: As I wrote earlier: use `convert(options = {}) {`  instead.

peppsac (commented):
This looks better. I left a few more comments on the API-related functions but I think it'll be good next time.

gliegard:
updated !

examples/globe_wfs_extruded.js
```diff
@@ -0,0 +1,108 @@
...
+        altitude: altitudeBuildings,
+        extrude: extrudeBuildings }),
+    url: 'http://wxs.ign.fr/72hpsel8j8nhb5qgdh07gcyp/geoportail/wfs?',
+    protocol: 'wfs',
+    version: '2.0.0',
+    id: 'wfsBuilding',
+    typeName: 'BDTOPO_BDD_WLD_WGS84G:bati_remarquable,BDTOPO_BDD_WLD_WGS84G:bati_indifferencie,BDTOPO_BDD_WLD_WGS84G:bati_industriel',
+    level: 14,
+    projection: 'EPSG:4326',
+    // extent: {
```
peppsac: Remove dead code please.

examples/wfs.js
```diff
@@ -43,32 +43,27 @@ view.camera.camera3D.position.set(1839739, 5171618, 910);
...
+    var rgb = properties.couleur.split(' ');
+    return new itowns.THREE.Color(rgb[0] / 255, rgb[1] / 255, rgb[2] / 255);
 }
 
 view.addLayer({
-    update: itowns.FeatureProcessing.update(colorFunctionLine),
+    update: itowns.FeatureProcessing.update,
+    convert: itowns.Feature2Mesh.convert({
+        color: colorLine,
+        altitude: undefined,
```
peppsac: Remove the `undefined`  properties.

src/Process/FeatureProcessing.js
```diff
@@ -23,105 +23,86 @@ function applyOffset(obj, offset) {
...
-                    if (colorFunction) {
-                        colorFunction(layer, node, result);
+        context.scheduler.execute(command).then((result) => {
+            // if request return empty json, WFS_Provider.getFeatures return undefined
+            if (result) {
+                // call onMeshCreated callback if needed
+                if (layer.onMeshCreated) {
+                    let mesh = result;
+                    if (result instanceof THREE.Group) {
+                        mesh = result.children[0];
```
peppsac: I don't think this is a good idea because the user won't be able to access to the other children.
You should call `onMeshCreated(result)` (and document *somewhere* what type of objects the user can expect)

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,338 @@
...
+ *
+ * @param  {structure with coordinate[] and featureVertices[]} coordinates - representation of the features
+ * @param  {properties[]} properties - properties of the features
+ * @param  {number} id - id of the feature
+ * @return {Coordinate[], propertie[] } {contour, properties}
+ */
+function extractFeature(coordinates, properties, id) {
+    const featureVertices = coordinates.featureVertices[id];
+    const contour = coordinates.coordinates.slice(featureVertices.offset, featureVertices.offset + featureVertices.count);
+    const propertie = properties[id].properties;
```
peppsac: `property`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,338 @@
...
 
-    // build indice and instanciate mesh
-    let result;
+        // set indices
+        const line = coordinates.featureVertices[id];
+        // TODO optimize indices lines
+        // is the same array each time
+        for (let i = line.offset; i < line.offset + line.count - 1; ++i) {
+            indices.push(i);
+            indices.push(i + 1);
```
peppsac, line 166: It seems you could just do `indices.push(i)` use a `THREE.Line` (instead of `LineSegments`).
Any reason for not doing this?

peppsac: But you'd have to create 1 `Line` per feature, instead of a 1 `LineSegment`.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,338 @@
...
+    let colors = [];
+    const geometry = new THREE.BufferGeometry();
+    let offset = 0;
+    /* eslint-disable guard-for-in */
+    for (const id in coordinates.featureVertices) {
+        // extract contour coodinates and properties of one feature
+        const { contour, propertie } = extractFeature(coordinates, properties, id);
+        // get altitude and extrude amount from properties
+        const altitudeBottom = getAltitude(options, propertie);
+        // console.log('ALT : ', altitudeBottom);
```
peppsac: Remove commented code please.

examples/globe_wfs_extruded.js
```diff
@@ -0,0 +1,108 @@
...
+// This layer is defined in a json file but it could be defined as a plain js
+// object. See Layer* for more info.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+
+// Add two elevation layers.
+// These will deform iTowns globe geometry to represent terrain elevation.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+
+function setMaterialLineWith(mesh) {
```
peppsac: Typo: `...Width`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,338 @@
...
-        vertices[offset + 2] = coordinate._values[2];
+    const vertices = new Array(contour.length * 3);
+    // loop over contour coodinates
+    for (const coordinate of contour) {
+        // convert coordinate to position
+        const vec = coordinate.xyz();
+        // get the normal vector.
+        const normal = coordinate.geodesicNormal;
+        // move the vertex following the normal, to put the point on the good altitude
+        vec.add(normal.clone().multiplyScalar(altitude));
```
peppsac, line 71: You're extruding along the normal and I don't think this is good idea: what about extruding along the Z axis then rotating your mesh so their local Z is aligned with the geodesicNormal at their position?

peppsac (dismissed):
Here's a small snippet which you could integrate in wfs.js:
```diff
+let meshes = [];
+const scaler = {
+    update: (dt) => {
+        if (meshes.length) {
+            view.notifyChange(true);
+        }
+        for (const mesh of meshes) {
+            mesh.scale.z = Math.min(
+                1.0, mesh.scale.z + 0.016);
+            mesh.updateMatrixWorld(true);
+        }
+        meshes = meshes.filter(m => m.scale.z < 1);
+    },
+};
+
+view.addFrameRequester(scaler);
+
 view.addLayer({
     type: 'geometry',
     update: itowns.FeatureProcessing.update,
     convert: itowns.Feature2Mesh.convert({
         color: colorBuildings,
         extrude: extrudeBuildings }),
+    onMeshCreated: (mesh) => {
+        mesh.scale.z = 0.01;
+        meshes.push(mesh);
+    },
```

Note: the same snippet doesn't work on globe_wfs because of the extrusion logic (see https://github.com/iTowns/itowns/pull/519#discussion_r147135443)

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: > Any reason for not doing this?

It is the existing code, I din't change it, just refactor.

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: No, It's an extrusion local to each points.

Imagine that you want to extrude a bigger polygon on the globe. For example, if we want to extrude a polygon representing region, a country or a continent. In this case, your extrusion on Z won't work.

peppsac: Then, use a rotation that is an average of all the normals, for instance.
And extrude along the diff between the normal at one point and this average.

WDYT?

peppsac: (or the normal at the center, it doesn't really matter,just pick the easiest to compute)

gliegard: Your animation works on planar, because Z axis make sense on planar.

But I don't want to make a wrong approximation to handle this case not adapted at all for globe.

If it's really important for you, I can add the normal vector to the mesh object, to let the user do this kind of scale and add this pretty scale example to the examples.

WDYT ?

peppsac: The thing is: you're building meshes in global coordinates if you extrude them this way... and it's not the way three.js works (ie: the whole matrix/matrixWorld thing is built on one assumption: each element of the scene uses local coordinates which are then transformed using `matrixWorld`)

gliegard: The existing code create meshes in global coordinates.

Your suggestion is interesting, but it's a new feature : put the mesh in local coordinates, adding rotations...

So we'll do that in another PR.

peppsac: > The existing code create meshes in global coordinates.

Not really. They're created in global coords but then vertices position are patched back to local coordinates (see `applyOffset` in FeatureProcessing`).

This wasn't the best technical option, but it was working and didn't cause issue because geometries were flat anyway.

IMHO you just have to compute a matrix that make sense:
- matrix's translation will probably identical to the `offset` from `applyOffset`
- matrix's rotation could be the geodesicNormal at node.center.
Then you can inverse this matrix and apply it to every vertices position you compute.

gliegard: Again, we can do that in another PR.

peppsac: Will you do this other PR? When?

Because I fear that the current implement fits perfectly your use-case and that you'll never bother fixing it (or it'll always be a low-priority item for you)...

gliegard: It's not a fix, it's a new feature.

The feature you speak about "Have the mesh in a local coordinate" was not in iTowns before. 

Adding extrusion to the polygons doesn't necessary mean add this new feature "Have the mesh in a local coordinate".

We can add an issue to speak more about this subject, but it's not the purpose of this PR.

peppsac: I disagree - mainly because I'm pretty sure this "other PR" will never happen.

Anyway I won't block on this specific point anymore since this PR implements a useful feature.

Mkonini: Yes indeed it's not the purpose of this PR, furthermore the new feature PR (when, who) is out of scope this code review. HTH

gliegard: I add this issue https://github.com/iTowns/itowns/issues/558

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: You can change the way we create lines, changing LineSegment to Line if you want in another PR, but it's not the purpose of this PR. This PR is about polygon extrusion.

peppsac: Good thing that I didn't asked you to change this then :)

gliegard:
@itowns-reviewers

examples/globe_wfs_extruded.js
peppsac: Still there.

examples/wfs.js
peppsac: Unaddressed comment

peppsac:
Can you add the snippet from https://github.com/iTowns/itowns/pull/519#pullrequestreview-72184139 to `wfs.js` please.

examples/globe_wfs_extruded.js
gliegard: ok

examples/wfs.js
gliegard: ok

gliegard:
> Can you add the snippet from #519 (review) to wfs.js please

I didn't add your snippet
- It doesn't pass the lint
- If you want to add it, consider coding this other feature #558 in another PR, and you will add the snippet in globe mode and planar mode.

gliegard:
That being said, I've updated the code.

autra:
> - It doesn't pass the lint

FWIW on examples folder, js code should be valid ES5 (should run in browser without babel), so change:
- `let` and `const` to `var`
- `for..of` to regular for loop
- arrow functions to regular `function foo(...) {...}`. 

With this it should pass the linter.

CONTRIBUTORS.md
```diff
@@ -5,8 +5,10 @@ The following people have contributed to iTowns 2.
 * [IGN](http://www.ign.fr)
   * [Gérald Choqueux](https://github.com/gchoqueux)
   * [Alexandre Devaux](https://github.com/nosy-b)
-  * [Victor C](https://github.com/VictorCo)
+  * [Victor Coindet](https://github.com/VictorCo)
   * [Quoc Dinh Nguyen](https://github.com/qdnguyen)
+  * [Grégoire Maillet](https://github.com/gmaillet)
+  * [Guillaume Liégard](https://github.com/gliegard)
```
autra, line 11: Congrats to both of you btw \o/

peppsac:
One last comment from me: please fix your git commits message (see https://github.com/iTowns/itowns/issues/518 for the template touse).

I'm stopping reviewing this PR, @autra will finish the reviewing process.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
 export default {
-    convert(feature) {
-        if (feature.geometries) {
-            return featureCollectionToThree(feature);
-        } else {
-            return featureToThree(feature);
-        }
+
+    convert(options = {}) {
+        return function _(feature) {
```
autra: This is the name that will get displayed in stacktraces. Use `_convert` instead of `_` for instance.

examples/globe_wfs_extruded.html
```diff
@@ -0,0 +1,69 @@
...
+            #menuDiv {
+                position: absolute;
+                top:0px;
+                margin-left: 0px;
+            }
+            @media (max-width: 600px) {
+                #menuDiv {
+                    display: none;
+                }
+            }
```
autra, line 31: Please format correctly the CSS  (no space before `:`, a space after, and no unit if the value is 0). I know you copy-pasted, but I'm fed up with submitting formatting PR before each release. Thanks!

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: But actually: it's never useful to make a closure around arguments, as js support function currying. Remove this closure for a simple function of signature `convert(options= {}, feature)` and replace every occurences of:
```js
 const convertFeature2Mesh = layer.convert ? layer.convert : Feature2Mesh.convert({});
```
by
```js
const convertFeature2Mesh = layer.convert || Feature2Mesh.convert.bind(Feature2Mesh, {});
```

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
 import * as THREE from 'three';
 import Earcut from 'earcut';
 
-function coordinatesToMesh(coordinates) {
-    if (!coordinates) {
-        return;
+function isFunction(functionToCheck) {
```
autra: Erf, unnecessary abstraction. And even if it were useful, it's not its place. Please remove this.

(FWIW, delete this function then run `sed -i "s;isFunction\(([^)]*)\);typeof \1 === 'function';" src/**/*.js -r`. That will do the trick )

gliegard:
updated

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+        if (isFunction(options.extrude)) {
+            return options.extrude(properties);
+        } else {
+            return options.extrude;
+        }
+    }
+    return 0;
+}
+
+function randomColor() {
```
autra, line 26: Same thing, not its place. But let's keep this as it is now, I'll make a PR for that (I have already a branch with a colorgenerator somewhere. I'll retrofit this code in that future PR).

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+    for (const id in coordinates.featureVertices) {
+        const { contour, property } = extractFeature(coordinates, properties, id);
+        // get altitude from properties
+        const altitude = getAltitude(options, property);
+        const newVertices = coordinatesToVertice(contour, altitude);
+        vertices = vertices.concat(newVertices);
+
+        // assign color to each point
+        const colorArray = createColorArray(contour.length, getColor(options, property), 255);
+        colors = colors.concat(colorArray);
```
autra, line 138: You're not doing anything with this color array right?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+        // get altitude from properties
+        const altitude = getAltitude(options, property);
+        const newVertices = coordinatesToVertice(contour, altitude);
+        vertices = vertices.concat(newVertices);
 
-    // build indice and instanciate mesh
-    let result;
+        // set indices
+        const line = coordinates.featureVertices[id];
+        // TODO optimize indices lines
```
autra: What's the meaning of this TODO?

src/Core/Scheduler/Providers/WFS_Provider.js
```diff
@@ -98,7 +98,10 @@ WFS_Provider.prototype.getFeatures = function getFeatures(crs, tile, layer) {
     if (result.feature !== undefined) {
         return Promise.resolve(result);
     }
-    return Fetcher.json(url, layer.networkOptions).then(geojson => assignLayer(Feature2Mesh.convert(GeoJSON2Features.parse(crs, geojson, tile.extent)), layer));
+
+    const convertFeature2Mesh = layer.convert ? layer.convert : Feature2Mesh.convert({});
```
autra: let's assign `layer.convert` with the result of this evaluation, to avoid re-creating a function each time getFeatures is called.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+        // fill the vertices array at the offset position
+        vec.toArray(vertices, offset);
+        // increment the offset
         offset += 3;
     }
+    return vertices;
+}
+
+/*
+ * Helper function to extract, for a given feature id, the feature contour coordinates, and it's properties.
```
autra: `it's` -> `its`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+ *
+ * @param  {structure with coordinate[] and featureVertices[]} coordinates - representation of the features
```
autra: is this type valid jsdoc? If not, consider making this a comment only : this function is not exposed, and jsdoc will break when we'll add this file to the documentation.

autra:
Ok thanks, I'll submit my review before tonight.

examples/globe_wfs_extruded.js
```diff
@@ -0,0 +1,102 @@
...
+
+// Add one imagery layer to the scene
+// This layer is defined in a json file but it could be defined as a plain js
+// object. See Layer* for more info.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+
+// Add two elevation layers.
+// These will deform iTowns globe geometry to represent terrain elevation.
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
```
autra, line 29: I guess ou only need on elevation layer for this example

autra: (Don't forget the comment above)

examples/wfs.js
```diff
@@ -87,29 +82,48 @@ view.addLayer({
...
+
+view.addFrameRequester(scaler);
 view.addLayer({
-    update: itowns.FeatureProcessing.update(colorFunctionBuildings),
+    type: 'geometry',
+    update: itowns.FeatureProcessing.update,
+    convert: itowns.Feature2Mesh.convert({
+        color: colorBuildings,
+        extrude: extrudeBuildings }),
+    onMeshCreated: function scaleZ(mesh) {
```
autra, line 123: Cute! <3

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+        indices.push(i + length);
+        indices.push(i + 1);
+        // second triangle indices
+        indices.push(i + 1);
+        indices.push(i + length);
+        indices.push(i + length + 1);
+    }
+}
+
+function coordinateToPoint(coordinates, properties, options) {
```
autra: `coordinatesToPoints`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+}
+
+/*
+ * Convert coordinates to vertices positionned at a given altitude
+ *
+ * @param  {Coordinate[]} contour - Coordinates of a feature
+ * @param  {number} altitude - Altitude of the feature
+ * @return {Vector3[]} vertices
+ */
+function coordinatesToVertice(contour, altitude) {
```
autra: `coordinatesToVertices`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+        // assign color to each point
+        const colorArray = createColorArray(contour.length, getColor(options, property), 255);
+        colors = colors.concat(colorArray);
+    }
+
+    geometry.addAttribute('position', new THREE.BufferAttribute(new Float32Array(vertices), 3));
+    return new THREE.Points(geometry);
+}
+
+function coordinateToLine(coordinates, properties, options) {
```
autra: `coordinatesToLines`

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -1,90 +1,336 @@
...
+    let vertices = [];
+    const indices = [];
+    let colors = [];
+    const geometry = new THREE.BufferGeometry();
 
-    geometry.addAttribute('position', new THREE.BufferAttribute(vertices, 3));
+    /* eslint-disable guard-for-in */
```
autra: Use `Object.hasOwnProperty` and don't disable eslint

autra: Same thing in other instance in this file.

autra: Ok, some of them are copy-pasta, but at *least* use `eslint-disable-next-line` instead of disabling the rule for the whole file.

examples/globe_wfs_extruded.html
```diff
@@ -0,0 +1,69 @@
...
+            menuGlobe.view = globeView;
+            // Listen for globe full initialisation event
+            globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
+                // eslint-disable-next-line no-console
+                console.info('Globe initialized');
+                Promise.all(promises).then(function () {
+                    menuGlobe.addImageryLayersGUI(globeView.getLayers(function (l) { return l.type === 'color'; }));
+                    menuGlobe.addElevationLayersGUI(globeView.getLayers(function (l) { return l.type === 'elevation'; }));
+
+                    globeView.controls.setTilt(75, true);
```
autra: Please consider using a lower angle. With this angle, a lot of tiles are visible near the horizon which slows down everything, and the example is doing a lot of jittering when this happens. I feel 45 is enough to get the feel of it. WDYT?

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: To elaborate on that: one reason why the jsdoc is only generated on a couple of files is that there are so many invalid js doc everywhere that reactivating it for the whole project is very cumbersome (I have tried to do so several times).

autra (changes requested):
Ok so I won't focus (too much) on the API thing, I guess you guys have already discussed enough about this. I'm merely interesting in the last clean before the merge. 

About commits:
- please put them in the correct order (polygon extrusion commit should come before the planar update using onMeshCreated and convert)
- please use the convention described [here](https://github.com/iTowns/itowns/issues/518). 

So actually, I've rewritten commit messages for you (in topological order):

- 1st commit
 
> refactor(examples): change controls in wfs planar example

- 2nd

> feat(wfs): add convert and onMeshCreated customisation for wfs layers
>
> If not empty, convert will be called instead of Feature2Mesh.convert to get THREE.js Meshes from the parsed GeoJSON.
> 
> onMeshCreated will be called just after the creation of a mesh.
>
> Together, these 2 mechanisms alow to customize the mesh creation. Possible applications include coloring according to some feature properties, meshes extrusion etc.

These should be ideally splitted in 2 commits, but let's forget that. Btw, what is the added value of onMeshCreated over convert in the current state of the code? I feel like the former could be abandoned to use the latter in all cases...

- 3rd

> refactor(examples): update wfs planar to use `layer.{convert,onMeshCreated}`

- 4th

> feat(examples): add polygon extrusion on globe example


One last remark, this code performs quite badly, and that's a real pity:  a *lot* of objects are created needlessly. Notably `coordinatesToVertice` create 2 Vector3 in a loop and one array, and is itself called in a loop in `coordinateToPolygonExtruded` which - by the way - discard and recreate `vertices` 2 times per loop (so bigger each time)... All of that cause a *huge* pressure on the GC: in some of my tests, GC time took more than 30% on the execution time (and profiling show that `coordinateToPolygonExtruded` is indeed the top offender)! And we can really feel it when using these examples...

I'm not blocking on that because I think this PR has been held long enough, but that's a thing 1/ to keep in mind when writing this kind of code and 2/ that we might want to revisit in the future.

¡Animo! We're nearly done here.

examples/globe_wfs_extruded.js
gliegard: I don't agree, it's a globe example, it represents the world.

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: What do you mean ?

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: I don't know, It was an existing comment, so I let the comment.

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: I've followed the first comment (call the function _convert) ; because the other suggestion doesn't work with the example.

gliegard:
Thanks @peppsac for all your suggestions and the work we've done on the API.
Thanks @autra for your suggestions too.

The code is updated.

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: This colors variable is not used in the current code. You're missing a geometry.addAttribute('color', ...) after line 141 probably.

examples/globe_wfs_extruded.js
autra: What do you want to show with this example? "Hey see I'm able to have wfs data on a terrain with elevation, and I can extrude them too. And also, it's on a real globe. And it's fast!". Point 1 and 2 are already done, and for that you need only elevations where your wfs data are. The less layer we have, the better point 3 will be achieved. Hence the fact that saving one layer loading is worthy here.

Anyway, whatever you want.

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: Well of course it does, with some little tweaks (you might want to bind on itowns.FeatureToMesh, or even on null, it should work too), but ok, fine by me. After all, maybe the api is clearer that way.

autra:
Ok one last thing to do([here](https://github.com/iTowns/itowns/pull/519#discussion_r148131782)), and this'll get merged first thing on Thursday morning (or whatever is your next work day) \o/

don't hesitate to ping me on irc as soon as this is ready.

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: Oh thanks !!

gliegard:
updated !

autra (approved):

autra:
Thanks for all the effort :-)

gliegard:
Thanks !! :)

PR #504, opened 2017-10-04
fix (View) : missing camera.update after resize

vcoindet:
## Description

Just added the following line : 
`this.camera.update(newSize.x, newSize.y);`
It was missing since PR #435 

## Motivation and Context

After resizing the screen, the globe wasn't exactly following the mouse.

PR #521, opened 2017-10-09
fix (GlobeControls) : moving camera.update call from GlobeControls to…

vcoindet:
… MainLoop so it won't be missed after a resize

## Description

removed 
`window.addEventListener('resize', this.updateCamera.bind(this), false);`
as it wasn't always working when the screen was resized...

Modification added in MainLoop with FrameRequester.

## Motivation and Context

#504 Improvement

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -71,7 +71,6 @@ function View(crs, viewerDiv, options = {}) {
         // the container's size. Otherwise we use window' size.
         const newSize = new Vector2(viewerDiv.clientWidth, viewerDiv.clientHeight);
         this.mainLoop.gfxEngine.onWindowResize(newSize.x, newSize.y);
-        this.camera.update(newSize.x, newSize.y);
         this.notifyChange(true);
     }, false);
 
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -456,7 +456,17 @@ function GlobeControls(view, target, radius, options = {}) {
         sizeRendering.FOV = this.camera.fov;
     };
 
-    window.addEventListener('resize', this.updateCamera.bind(this), false);
+    const self = this;
+    const resizeHandler = {
+        update() {
+            const dim = self._view.mainLoop.gfxEngine.getWindowSize();
+            const sizeDiff = (dim.width != sizeRendering.width || dim.height != sizeRendering.height);
+            if (sizeDiff) {
+                self.updateCamera();
+            }
+        },
+    };
+    this._view.addFrameRequester(resizeHandler);
 
     this.getAutoRotationAngle = function getAutoRotationAngle() {
         return 2 * Math.PI / 60 / 60 * this.autoRotateSpeed;


src/Core/MainLoop.js
```diff
@@ -92,8 +92,26 @@ MainLoop.prototype._step = function _step(view, timestamp) {
 
     // update camera
     const dim = this.gfxEngine.getWindowSize();
+    const foo = (dim.width != view.camera.width || dim.height != view.camera.height);
```
autra: Let's find a better name than `foo`. `foo` is Sacred™, let's use it only to praise the Gods of Coding ;-)

src/Core/MainLoop.js
```diff
@@ -92,8 +92,26 @@ MainLoop.prototype._step = function _step(view, timestamp) {
...
+
+    const animator = {
+        t: 0,
+        update: (dt, ignore) => {
+            this.t = ignore ? 0 : Math.min(1, dt / 1000 + this.t);
+
+            if (this.t < 1) {
+                if (foo) {
+                    view.controls.updateCamera();
```
autra: `view.controls` exists only for `GlobeView`.

`view.controls.updateCamera` exists only for GlobeControls. On overall, this is a GlobeControls issue. It should be in GlobeControls. Actually, I think GlobeControls should be a registered FrameRequester and this should happen in its update method.

autra: And I think `foo` calculation should be moved inside the frame requester to ease this.

src/Core/MainLoop.js
```diff
@@ -92,8 +92,26 @@ MainLoop.prototype._step = function _step(view, timestamp) {
 
     // update camera
     const dim = this.gfxEngine.getWindowSize();
+    const foo = (dim.width != view.camera.width || dim.height != view.camera.height);
+
+    const animator = {
+        t: 0,
+        update: (dt, ignore) => {
+            this.t = ignore ? 0 : Math.min(1, dt / 1000 + this.t);
```
autra: What about setting `this.t = 0` at the same time we call `view.controls.updateCamera` and testing `Date.now() - this.t > 1000` on line 102? I find it easier to read.

src/Core/MainLoop.js
```diff
@@ -92,8 +92,26 @@ MainLoop.prototype._step = function _step(view, timestamp) {
...
+
+            if (this.t < 1) {
+                if (foo) {
```
autra: If foo is false, you want to do nothing right? Not even increasing `this.t` I think.

autra (changes requested):

src/Core/MainLoop.js
vcoindet: oops... sorry about that, didn't realize it was still here !

src/Core/MainLoop.js
vcoindet: I agree, let's do that ! ;)

src/Core/MainLoop.js
vcoindet: Yes that's right, do nothing if false.

vcoindet:
@autra : WDYT now ?

peppsac:
@vcoindet what about this alternative.
Add something similar to `GlobeControls.js` constructor:
```js
const resizeHandler = {
    update() {
        const dim = this._view.mainLoop.gfxEngine.getWindowSize();
        const sizeDiff = (dim.width != sizeRendering.width 
            || dim.height != sizeRendering.height);  
        if (sizeDiff) {
             this.updateCamera();
        }
    }
};
view.addFrameRequester(resizeHandler);
```

It feels like it would perform the same task (call updateCamera after resize) in a much simpler way.
WDYT?

vcoindet:
seems good, but I can't have access to `this` in the `update()` function. Unless I do that : 
```
const self = this;
const resizeHandler = {
    update() {
          const dim = self._view.mainLoop.gfxEngine.getWindowSize();
          const sizeDiff = (dim.width != sizeRendering.width || dim.height != sizeRendering.height);
          if (sizeDiff) {
              self.updateCamera();
          }
      },
  };
this._view.addFrameRequester(resizeHandler);
```

vcoindet:
I pushed a new commit, WDYT @peppsac and @autra ?

autra (approved):
LGTM. This code can be simplified once we accept simple function as FrameRequester.

autra:
Thanks!

PR #522, opened 2017-10-10
feat(core): add the ability to colorize pointclouds

autra:
## Description

This actually moves a debug feature into core: it replaces `debugColor`
and `useDebugColor` with `customColor` and `useCustomColor`.

## Motivation and Context

I think it is a useful feature to have in a production context: 
- users can have several pointclouds displayed at the same time, and it's convenient to allow to separate them visually.
- When clouds have neither information of color nor intensity, colorizing them allows to see applied visual improvements. We don't have any on itowns right now, but for example Eye-Dome lightning add some black outlines. You can't see them if the cloud is black too.


peppsac (approved):
Thanks!

Looks good for me with just one nit: could you move the whole text of your PR in the commit message before merging?

autra:
@peppsac updated

peppsac (approved):

PR #523, opened 2017-10-10
feat (3dtiles): use different subdiv. logic for add and sub refinement

peppsac:
This commit changes the way additive refinement is handled for 3dtiles.
Until now we had a single option: when a tile reached its sse threshold,
we would download and display all of its children.

In the additive refinement we can do better and only download the children
that will be visible. This reduces network/mem usage, for the same visual
result.

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -14,21 +14,51 @@ import { init3dTilesLayer } from '../../../Process/3dTilesProcessing';
 function $3dTilesIndex(tileset, baseURL) {
     let counter = 0;
     this.index = {};
-    const recurse = function recurse_f(node, baseURL) {
+    const inverseTileTransform = new THREE.Matrix4();
+    const recurse = function recurse_f(node, baseURL, parent) {
+        // compute transform (will become Object3D.matrix when the object is downloaded)
+        node.transform = node.transform ? (new THREE.Matrix4()).fromArray(node.transform) : undefined;
+
+        // The only reason to store _worldFromLocalTransform is because of extendTileset where we need the
+        // transform chain for one node.
+        node._worldFromLocalTransform = node.transform;
+        if (parent && parent._worldFromLocalTransform) {
+            if (node.transform) {
+                node._worldFromLocalTransform = new THREE.Matrix4().multiplyMatrices(
+                    parent._worldFromLocalTransform, node.transform);
+            } else {
+                node._worldFromLocalTransform = parent._worldFromLocalTransform;
+            }
+        }
+
+        // getBox only use inverseTileTransform for volume.region so let's not
+        // compute the inverse matrix each time
+        if ((node.viewerRequestVolume && node.viewerRequestVolume.region)
+            || (node.boundingVolume && node.boundingVolume.region)) {
+            if (node._worldFromLocalTransform) {
+                inverseTileTransform.getInverse(node._worldFromLocalTransform);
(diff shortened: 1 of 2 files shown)


src/Process/3dTilesProcessing.js
```diff
@@ -7,41 +9,86 @@ function requestNewTile(view, scheduler, geometryLayer, metadata, parent) {
...
-
         Promise.all(promises).then(() => {
             node.pendingSubdivision = false;
             context.view.notifyChange(true);
         });
     }
 }
 
-export function $3dTilesCulling(node, camera) {
+export function $3dTilesCulling(node, camera, overrideMatrixWorld) {
```
Jeremy-Gaillard: Not a big fan of this override. I would prefer something like this instead: 
```js
export function $3dTilesCulling(viewerRequestVolume, boundingVolume, matrixWorld, camera)
```

Jeremy-Gaillard (changes requested):
Very quick review, but this seems nice. It also fixes the issue with the viewerRequestVolume not being properly handled.

src/Process/3dTilesProcessing.js
peppsac: I changed the function, and fixed the special case of boundingVolume.region. Let me know what you think.

src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -170,21 +191,18 @@ function configureTile(tile, layer, metadata, parent) {
...
         tile.additiveRefinement = parent ? (parent.additiveRefinement) : false;
     }
-    tile.parentFromLocalTransform = tile.transform;
-    tile.worldFromLocalTransform = new THREE.Matrix4().multiplyMatrices(parent ? parent.worldFromLocalTransform : new THREE.Matrix4(), tile.parentFromLocalTransform);
-    const m = new THREE.Matrix4();
-    m.getInverse(tile.worldFromLocalTransform);
-    tile.viewerRequestVolume = metadata.viewerRequestVolume ? getBox(metadata.viewerRequestVolume, m) : undefined;
-    tile.boundingVolume = getBox(metadata.boundingVolume, m);
+    tile.viewerRequestVolume = metadata.viewerRequestVolume;
+    tile.boundingVolume = metadata.boundingVolume;
```
Jeremy-Gaillard, line 214: Can you remove these? It is confusing to have the same information stored at two different places. I'd suggest only keeping a link to the metadata.

Jeremy-Gaillard (changes requested):
You forgot to change the call to recurse in the extendTileset function. It should be something like this:
```js
recurse(tileset.root, baseURL, this.index[nodeId]);
```

src/Core/Scheduler/Providers/3dTiles_Provider.js
peppsac: Both prperties are used often, so removing them *could* have some perf impact.
I'd prefer to not remove them for now.

peppsac:
Fixed the `recurse`call.

src/Core/Scheduler/Providers/3dTiles_Provider.js
Jeremy-Gaillard: That not exactly what I meant. My point was there should be a clearer separation between the metadata part of the object (bounding volume, etc.) and Object 3D part (transformation). Presently, both are mixed and sometimes we call culling on the metadata part, sometimes on the Object 3D part.

Jeremy-Gaillard (approved):

PR #524, opened 2017-10-11
fix (provider): add elevation textures support to StaticProvider

peppsac:
Make sure elevation textures works with StaticProvider.

### DIFF
--- a/src/Core/Scheduler/Providers/StaticProvider.js
+++ b/src/Core/Scheduler/Providers/StaticProvider.js
@@ -2,7 +2,7 @@ import { Vector3 } from 'three';
 import Extent from '../../Geographic/Extent';
 import OGCWebServiceHelper from './OGCWebServiceHelper';
 import Fetcher from './Fetcher';
-import { l_COLOR } from '../../../Renderer/LayeredMaterial';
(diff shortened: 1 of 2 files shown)


src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -111,13 +112,18 @@ export default {
...
+            return false;
+        }
         const s = selectBestImageForExtent(layer.images, tile.extent);
         if (!s) {
             return false;
         }
         const mat = tile.material;
         const idx = mat.getLayerTextureOffset(layer.id);
-        const currentTexture = mat.textures[l_COLOR][idx];
+        const currentTexture = layer.type === 'color' ?
```
gchoqueux: you can use `mat.getLayerTextures`

autra (approved):
I forgot you requested my review on this. With the nit from @gchoqueux, LGTM.

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Thanks, fixed.

PR #526, opened 2017-10-11
fix: restore PM layers texture display

### LINKED ISSUE #525: Some layers are not displayed
##  Environment

## Context

I try to update the geoportal project with the last version of the iTowns library (branch master). 
I encounter a bug : some layers are not displayed, even if in the network tab of the Chrome development console the queries are correctly launched, and responses are "200".

The issue happens on the main iTowns example. The Scan Express layer is not displayed (see below to reproduce).

1. Open the main itowns example (index.html) in a web browser.
2. In the GuiTools Menu, set as "not visible" all the color layers except the "ScanEX" one.

## Expected Behavior

The Scan Express layer should be displayed.

## Actual Behavior

The Scan Express layer is not displayed (even if queries are correctly launched)

scanex_notdisplayed

I tried to checkout on old master commits, to identify the origin of the problem. The commit b2e616c seems to have introduced this bug. Before it, the problems described above were not visible.
If the bug is confirmed, maybe it could be interesting to add tests on the build to check if layers were correctly added/displayed.

peppsac:
This was broken in PR https://github.com/iTowns/itowns/pull/492

This commit fixes the issue by comparing the texture count to the
correct index.

Fixes #525

### DIFF
--- a/src/Renderer/Shader/TileFS.glsl
+++ b/src/Renderer/Shader/TileFS.glsl
@@ -98,7 +98,7 @@ void main() {
                     int pmTextureCount = int(paramsA.y);
                     int textureIndex = int(paramsA.x) + (projWGS84 ? 0 : pmSubTextureIndex);
 
-                    if (!projWGS84 && pmTextureCount <= textureIndex) {
+                    if (!projWGS84 && pmTextureCount <= pmSubTextureIndex) {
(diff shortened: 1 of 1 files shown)


gchoqueux (approved):
thanks for fix

PR #530, opened 2017-10-13
fix (geometry): remove tile's layer

gchoqueux:
## Description
remove the specified `layerUpdateState` when layer is removed

## Motivation and Context
When you remove and re-add the same layer, the tiles retained the state of the first layer


peppsac (approved):
Thanks

Issue #529, opened 2017-10-13
Plugins ?


mbredif:
I do not know if this issue implies documentation or development, but what is the standard way of adding custom functionalities to itowns ?
In other words, simple projects may only rely on js code in the html page but more more complex project (custom providers, custom materials...) may require to tweak the internals of itowns. Is it required to then fork the project or is there some kind of plugin architecture ?

autra:
*Ideally* we should not have to fork the project.

As Oslandia, we often do that for client projects. My current process is:
- install itowns as npm dependency `npm install itowns --save`
- use whatever bundler (webpack, rollup...) to import individual itowns module into your files. 
- create your own module. You can very well create your own material, own provider (and execute `view.mainLoop.scheduler.addProtocolProvider` etc... A lot of things are already possible, but of course we might miss some API. Don't hesitate to open issues or PR for them!

The process is quite similar as to using THREE in your project. Just use itowns in your project the way we use THREE in itowns... Well I would say that if our package wasn't currently broken !

Indeed, I just realized this morning (while you were writing this probably) that things are a bit funky these times:
- if you want to use three, you should use itowns.THREE and not three directly. It's because we should declare THREE as a peerDependency instead. I'm fixing this asap.
- Our npm package is currently broken because THREE.MeshLine - one of our dependencies -  doesn't declare THREE neither as peerDependency nor as dependency... So sadly it works on itowns because we directly depend on THREE, but not on project that depends on itowns. I expect the above point to fix that as well.
- project that depends on itowns and import it via a package bundler MUST declare a global `__DEBUG__` var somewhere. I need to fix that too. They won't have this problem if they use the bundles in `dist/`.

autra:
BTw the package is broken in 2.1, not in 2.0 (we don't use MeshLine there).

autra:
@mbredif FYI: 
- #531 fixes the last problem
- #532 the first.

For the problem with MeshLine, one workaround is to install `three` in your client with `npm install --save three`, even if you don't use it.

I'll try to push a client example on our organization to get you started.

So again, *ideally*, client code won't need to fork itowns or copy the bundles into their repo, but will use it as any npm dependency.

mbredif:
That is great, thanks for the comments and the fixes !
Is this process of using itowns as an npm dependency documented somewhere, maybe as a tutorial ?
It would help a lot to get people on board !

autra:
[...]

PR #531, opened 2017-10-13
Remove debug code from transpiled sources.

autra:
## Description
This contains 2 commits:

- the first one correctly remove debug code from transpiled sources

- the second is a basic refactoring

## Motivation and Context

Before this patch, people using itowns as a module (as it should be used) had to manually define __DEBUG__ in their own module bundler config, otherwise they had some undefined error. This commit correctly strips itowns out of these debug code during transpilation.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,6 +3,16 @@
     "presets": ["es2015"],
     "plugins": [
         "transform-runtime",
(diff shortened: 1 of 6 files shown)


autra:
Referencing https://github.com/iTowns/itowns/issues/529 here.

tbroyer (approved):
:shipit:

peppsac:
Thanks @autra

PR #532, opened 2017-10-13
Fix npm package

autra:
## Description
These are a collection of small fixes in our packaging system.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,7 +14,7 @@ var providePlugin = new webpack.ProvidePlugin({
 
 module.exports = {
     entry: {
-        itowns: ['es6-promise', 'whatwg-fetch', path.resolve(__dirname, 'src/Main.js')],
+        itowns: ['es6-promise', 'whatwg-fetch', path.resolve(__dirname, 'src/MainBundle.js')],
         debug: [path.resolve(__dirname, 'utils/debug/Main.js')],
     },
     devtool: 'source-map',
@@ -54,7 +54,7 @@ module.exports = {
                 // .babelrc is used for transpiling src/ into lib/ in the prepublish
                 // phase, see package.json
                 options: {
-                    presets: ['es2015'],
+                    presets: [['es2015', { modules: false } ]],
                     plugins: ['transform-runtime'],
                     babelrc: false,
                 },
--- a/examples/externalscene.html
+++ b/examples/externalscene.html
@@ -34,6 +34,7 @@
         <script type="text/javascript">
             var renderer;
             var exports = {};
+            var THREE = itowns.THREE;
         </script>
         <script src="externalscene.js"></script>
     </body>
--- a/examples/externalscene.js
+++ b/examples/externalscene.js
@@ -1,7 +1,7 @@
-/* global itowns, document, renderer */
+/* global itowns, THREE, document, renderer */
 var positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };
 
-var scene = new itowns.THREE.Scene();
+var scene = new THREE.Scene();
 
 // iTowns namespace defined here
 var viewerDiv = document.getElementById('viewerDiv');
--- a/examples/planar.html
+++ b/examples/planar.html
@@ -72,6 +72,7 @@
         <script src="../dist/itowns.js"></script>
         <script type="text/javascript">
             var renderer; var exports = {};
+            var proj4 = itowns.proj4;
         </script>
         <script src="planar.js"></script>
     </body>
--- a/examples/planar.js
+++ b/examples/planar.js
@@ -1,4 +1,4 @@
-/* global itowns, document, renderer */
+/* global itowns, document, renderer, proj4 */
 // # Planar (EPSG:3946) viewer
 
 var extent;
@@ -6,7 +6,7 @@ var viewerDiv;
 var view;
 
 // Define projection that we will use (taken from https://epsg.io/3946, Proj4js section)
-itowns.proj4.defs('EPSG:3946',
+proj4.defs('EPSG:3946',
     '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
 
 // Define geographic extent: CRS, min/max X, min/max Y
--- a/examples/postprocessing.html
+++ b/examples/postprocessing.html
@@ -67,6 +67,7 @@
         <script type="text/javascript">
             var renderer;
             var exports = {};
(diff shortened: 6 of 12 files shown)


tbroyer (dismissed):
This is a somewhat breaking change, so should be highlighted in the release notes for the next release (for those who wouldn't use three, or meshline, in their apps)

autra:
> This is a somewhat breaking change, so should be highlighted in the release notes for the next release (for those who wouldn't use three, or meshline, in their apps)

Normally, they shouldn't have to do anything (three and three.meshline are still in `dependencies`)

tbroyer:
> (three and three.meshline are still in dependencies)

Oops, missed it.
Shouldn't they be removed then though? What's the goal of adding them as peer dependencies but keeping them as dependencies? Making sure the correct version is being used? (given that almost all update of three has breaking changes, I'm not sure the configuration is correct there then)

autra:
My understanding (to be taken with a grain of salt) is that if a project A declares B as `peerDependencies`, B used to be installed automatically when you do `npm install` in A folder. Apparently, as of npm 5, that has changed (according to my tests) and it no longer installs anything on its own. I think the rationale is that you might want to declare you're compatible with a specifi version of a package, without actually explicitly depends on this package.

> Making sure the correct version is being used?

On downstream projects yes. It only displays warnings when there's a mismatch though (they should really be errors with a -f flag imho, but nobody cares about my npm opinions  it seems!). 

Long story short, as we still need three to be in itowns' node_modules, we need it in dependencies as well. 

Onto the version specifier question: in semver, all 0.x versions are considered major. Therefore, while ^1.0 would include every 1.x versions, ^0.77 would only include 0.77.x versions. From https://docs.npmjs.com/misc/semver on caret range:

> Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

And [the npmjs.com semver calculator](https://semver.npmjs.com/) confirms that.

Npm is fun!

peppsac (commented):
It looks good to me but I'll leave the final word to you or @tbroyer on this topic :)

(nit: there's a typo to correct in one commit title: `dependencie`-> `dependency`)

autra:
> (nit: there's a typo to correct in one commit title: dependencie-> dependency)

indeed, corrected :-)

autra:
Wait before merging, there's actually a problem by doing so: in a project depending on itowns, three would get installed and imported 2 times: one in node_modules/three, and one in node_modules/three/itowns/node_modules/three, causing a lot of problem (Object3D's id being wrong, code using `foo instanceOf Vector3` becoming unreliable etc).

tbroyer:
> there's actually a problem by doing so: in a project depending on itowns, three would get installed and imported 2 times: one in node_modules/three, and one in node_modules/three/itowns/node_modules/three, causing a lot of problem

Yes, that's because it's still a "normal" dependency.

A "peer dependency" would fix this, at the expense of asking **everyone** to explicitly add a dependency on tree (a breaking change, but possibly not a big deal; I mean, everyone using React in the browser for instance knows they have to depend on both `react` and `react-dom`; whereas `react-dom` could have had a "normal" dependency on `react`, but use a peer dependency instead, probably for the same reasons we'd like to use a peer dependency for three too)

autra:
Yep, but actually, I think I have now the good setup: having them both in peerDependencies and **dev**Dependencies (would get installed too on `npm i` in itowns).

Yes, we actually don't want to hide THREE at all, and I don't think it's bad to require users to install it separately. That also means we should remove itowns.THREE someday.

peppsac:
> That also means we should remove itowns.THREE someday

Isn't this PR a good time to get rid of `itowns.THREE`?

autra:
> Isn't this PR a good time to get rid of itowns.THREE?

I need to think of a way to keep being able to use three in examples then.

tbroyer:
Use another entrypoint script than `Main.js` when bundling with Webpack?
that reexports everything from `Main.js` **and** three and proj4 (btw, should we do the same for proj4 that we do with three in this PR?); that way, NPM users would use `Main.js` (modified to remove three) and their local/peer three, and examples (and those people using the Webpack bundle) would have to use `itowns.THREE`.

The alternative would be to have the bundle depend on an "ambient" three (i.e. force users of the bundle to first include three in the page – like people are used to include jquery before they can use any jquery plugin).

BTW, there seems to be a single use of `itowns.THREE` outside of examples (in `utils/debug/PointCloudDebug.js`)

autra:
> (btw, should we do the same for proj4 that we do with three in this PR?)

Done.

> Isn't this PR a good time to get rid of itowns.THREE?

Well, that's really not that simple actually. I need to persuade itowns.js to take THREE (and all our peerDependencies really) as an external. It's quite easy of course, but it also means `itowns.js` won't include our peerDependencies by default. In turns, this means that if we want to still provide a drop-in bundle, I need to generate a itowns_full.js too. I have something that nearly works, but currently it involves adding 2 more webpack configs :-) Worth a PR on its own, right?

autra:
I just pushed a commit that put proj4 in peerDependencies. @iTowns/reviewers please r?

tbroyer (approved):
> Well, that's really not that simple actually.

How about my proposal to have the "drop-in bundle" ship with three and proj4 and provide them at `itowns.THREE` and `itowns.proj4`? (exactly like we do today; just that those properties wouldn't exist when using itowns as an npm dependency)

> Worth a PR on its own, right?

I'm fine with this (though if you'd like to go with the –probably simpler– proposal above, then maybe do it in this PR)

autra:
I didn't want to bother exposing `itowns.THREE` and `itowns.proj4` for user of the full bundle, so I was planning to completely remove these from itowns for everybody. Those wanted to use them directly would either use npm or our "light" bundle. WDYT?

autra:
no strong opinion either way though.

autra:
@tbroyer I ended up implementing your suggestion. I have been too greedy at first: I wanted a clean module use case *and* a full bundle *and* a bundle relying on ambiant peer dependencies, but that's apparently impossible to do with one webpack configuration. So let's be more reasonable here :-) WDYT now?

I still might do an ambient build (without peerDependencies) one day if we feel the need (and I find the motivation) for it. Thanks for your suggestion anyway :-)

@iTowns/reviewers updated.

webpack.config.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
                 // .babelrc is used for transpiling src/ into lib/ in the prepublish
                 // phase, see package.json
                 options: {
-                    presets: ['es2015'],
+                    presets: [['es2015', { modules: false } ]],
```
autra, line 57: Babel support for `export * from 'foo'` seems broken, so I let webpack deals with that.

peppsac (approved):
LGTM

tbroyer (approved):

PR #535, opened 2017-10-13
fix (controls) : fixes for GlobeControls

### LINKED ISSUE #300: setCameraTargetGeoPosition doesn't work in index.html
`setCameraTargetGeoPosition` doesn't work when called in `index.html` when `isAnimated` is settted to `true`
This might be a synchronization problem.

Works fine when `isAnimated` is sette to `false`
Works fine in console.

### LINKED ISSUE #411: White screen when zooming

### LINKED ISSUE #486: setCameraTargetGeoPosition : gap with the wanted position
globeView.controls.setCameraTargetGeoPosition can introduce a gap between the wanted position and the arrival position, and can even change the tilt and/or the heading.

## Your Environment

* Version used: itowns 2.1
* Browser Name and version: Chrome 60.0 & Firefox 55.0.2 (64 bits)
* Operating System and version: Ubuntu 16.04

1 - Change the zoom
2 - change the tilt and the heading of the globe with the mouse
3 - In chrome and firefox console : 
```
globeView.controls.setCameraTargetGeoPosition({
       longitude: 5.369789,
       latitude: 43.296482
 })
```

-> You're not on the wanted position, a gap is introduced (sometimes a big gap). It can even change the tilt and the heading.

## Expected Behavior
With those coordinates, one should be on Marseille's port

## Actual Behavior
Not on the Marseille's port

gchoqueux:
**Fixes multi bug in globeControls**

- fix (control): bad camera's heading convention in `globeControls`

- fix (controls): in `GlobeControls.setCameraTargetPosition` the final position isn't the desired position
The problem was due to altitude changes during the displacement
fix #486 

- fix (controls): in `GlobeControls.setOrbitalPosition`, the final position isn't the desired position
The problem was due to altitude changes during the displacement

- fix (controls): fix error target with pick depth position
For the large ranges pick depth position is too imprecise
More robust calculation with `DEMUtils`

- fix (controls): add target update target for wheel event
Target wasn't updated after range change
fix #411 

- fix (controls): wait initialization before move camera
if you change camera's position before `init`,
the movement doesn't oppose
fix #300

- fix (controls): add `THREE.layer` for helper axis
Because helper fakes pick position
because Helper axis write in depth buffer

### DIFF
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -94,7 +94,7 @@ const EPS = 0.000001;
 const rotateStart = new THREE.Vector2();
 const rotateEnd = new THREE.Vector2();
 const rotateDelta = new THREE.Vector2();
-const spherical = new THREE.Spherical(1.0, 0.01, Math.PI * 0.5);
+const spherical = new THREE.Spherical(1.0, 0.01, 0);
 const snapShotSpherical = new THREE.Spherical(1.0, 0.01, Math.PI * 0.5);
 const sphericalDelta = new THREE.Spherical(1.0, 0, 0);
 const sphericalTo = new THREE.Spherical();
@@ -128,8 +128,8 @@ var animatedScale = 0.0;
 const positionObject = (function getPositionObjectFn()
 {
     const quaterionX = new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(1, 0, 0), Math.PI / 2);
-    return function positionObject(newPosition, object)
-    {
+    return function positionObject(newPosition, object) {
+        object.up = THREE.Object3D.DefaultUp;
         object.position.copy(newPosition);
         object.lookAt(newPosition.clone().multiplyScalar(1.1));
         object.quaternion.multiply(quaterionX);
@@ -1603,7 +1603,7 @@ GlobeControls.prototype.getTilt = function getTilt() {
  * @return {Angle} number - The angle of the rotation in degrees.
  */
 GlobeControls.prototype.getHeading = function getHeading() {
-    return spherical.theta * 180 / Math.PI;
+    return (THREE.Math.radToDeg(spherical.theta) + 360) % 360;
 };
 
 GlobeControls.prototype.getTiltRad = function getTiltRad() {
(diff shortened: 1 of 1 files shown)


peppsac:
Some of the commits are missing details about what is being fixed: this is needed to understand what's going on (or open an issue and mark your commit as fixing this issue).
Example: `setOrbitalPosition position error` is not enough: what kind of error? where does it came from? etc etc...

gchoqueux:
@peppsac  updated

gchoqueux:
reviewer?

autra:
Can you please split this? Especially I would like all the depth precision issue in its own PR. 

GlobeControls is becoming a clear example of spaghetti code. It should be refactored/rewritten imho. At the very least it should embrace the FrameRequester logic. WDYT?

gchoqueux:
@peppsac  @autra, yes of course, globe controls requires a total re factorization. I'll do it soon.

nosy-b:
I agree with the fact that GlobeControls is pretty messy. It needs a complete rework, I created an issue for that matter: #545 

But in the mean time this PR removes lots of actual annoying bugs without side effects nor touching any other classes so I propose that we move ahead with this last bug fix before a complete refactor.

autra:
> I propose that we move ahead with this last bug fix before a complete refactor.

Yes, of course. Can we have the depth precision issue in its own PR please? I feel this is the tricky part. We'll merge the other fixes quickly, and it would be easier to review the depth precision part too.

nosy-b (approved):
It corrects efficiently many bugs. 
It can be merged like this as an exception (7 commits) because we agree that a quick fix was needed and a complete rework of this class is programmed next and will need some fresh energy!

autra:
> as an exception

All right.

Next time you are in a hurry please at least take the time to drop a PM on IRC. I can understand it, and would have reviewed taking this into account (or let you the review, no problem) and we'd merged quickly.

Basically, in this precise case, I would have liked to review the commit messages only: we can always change the code, but *not* the commit messages. They are very useful when we need to understand something, or when making the changelog during a release. As PR authors, it's always difficult to make good ones when we submit them, because as we did the code, it's obvious for us. So imho commit messages should always been read by another pair of eyes than the submitter :-)

Thanks for your understanding.

PR #536, opened 2017-10-18
fix (protocols): support wms in EPSG:3857 projection

### LINKED ISSUE #478: EPSG:3857 (PseudoMercator) projection not handled for WMS layers

gchoqueux:
## Description
wms protocol supports  in `EPSG:3857` projection

## Motivation and Context
wms protocol only supported the `EPSG:4326`
fix #478

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -1,6 +1,8 @@
 import * as THREE from 'three';
 import Coordinates, { crsToUnit, crsIsGeographic, assertCrsIsValid, convertValueToUnit, reasonnableEpsilonForUnit } from '../Geographic/Coordinates';
+import Projection from '../Geographic/Projection';
 
+const projection = new Projection();
 /**
  * Extent is a SIG-area (so 2D)
  * It can use explicit coordinates (e.g: lon/lat) or implicit (WMTS coordinates)
@@ -90,7 +92,35 @@ Extent.prototype.as = function as(crs) {
     assertCrsIsValid(crs);
 
     if (_isTiledCRS(this._crs)) {
-        throw new Error('Unsupported yet');
+        if (this._crs == 'WMTS:PM') {
+            // Convert this to the requested crs by using 4326 as an intermediate state.
+            const nbCol = Math.pow(2, this.zoom);
+            const size = 360 / nbCol;
+            // convert column PM to longitude EPSG:4326 degree
+            const west = 180 - size * (nbCol - this.col);
+            const east = 180 - size * (nbCol - (this.col + 1));
+            const nbRow = nbCol;
+            const sizeRow = 1.0 / nbRow;
+            // convert row PM to Y PM
+            const Yn = 1 - sizeRow * (nbRow - (this.row));
+            const Ys = 1 - sizeRow * (nbRow - (this.row + 1));
+            // convert Y PM to latitude EPSG:4326 degree
+            const north = THREE.Math.radToDeg(projection.YToWGS84(Yn));
+            const south = THREE.Math.radToDeg(projection.YToWGS84(Ys));
+            // create intermediate EPSG:4326 and convert in new crs
+            return new Extent('EPSG:4326', { west, east, south, north }).as(crs);
+        } else if (this._crs == 'WMTS:WGS84G' && crs == 'EPSG:4326') {
+            const nbRow = Math.pow(2, this.zoom);
+            const size = 180 / nbRow;
+            const north = size * (nbRow - this.row) - 90;
+            const south = size * (nbRow - (this.row + 1)) - 90;
+            const west = 180 - size * (2 * nbRow - this.col);
+            const east = 180 - size * (2 * nbRow - (this.col + 1));
+
+            return new Extent(crs, { west, east, south, north });
+        } else {
+            throw new Error('Unsupported yet');
+        }
     }
 
     if (this._crs != crs) {
--- a/src/Core/Scheduler/Providers/WMS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMS_Provider.js
@@ -4,7 +4,7 @@
  * Description: Provides data from a WMS stream
  */
 
-
+import * as THREE from 'three';
 import Extent from '../../Geographic/Extent';
 import OGCWebServiceHelper from './OGCWebServiceHelper';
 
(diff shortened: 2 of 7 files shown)


src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -134,4 +152,18 @@ WMS_Provider.prototype.executeCommand = function executeCommand(command) {
...
 };
 
+WMS_Provider.prototype.getColorTextures = function getColorTextures(tile, layer, targetLevel) {
+    if (tile.material === null) {
+        return Promise.resolve();
+    }
+    const promises = [];
+    const bcoord = tile.getCoordsForLayer(layer);
+
+    for (const coordWMTS of bcoord) {
```
peppsac: Why a WMS provider would want to return several image for a single tile?

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -69,13 +82,21 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
...
+        }
+        if (c.zoom < layer.options.zoom.min || c.zoom > layer.options.zoom.max) {
+            return false;
+        }
+    }
+    return true;
 };
 
-WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, targetLevel) {
+WMS_Provider.prototype.getColorTexture = function getColorTexture(coordWMTS, layer, tile, targetLevel) {
```
peppsac: `coordWMTS` name seems out of place in a WMTS_Provider. Can you explain why you use this name?

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -97,20 +115,20 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, t
         }
         extent = parentAtLevel.extent;
         targetLevel = parentAtLevel.level;
+        const pow = 1 / (coordWMTS.zoom - targetLevel);
+        const row = Math.floor(Math.pow(coordWMTS.row, pow));
```
peppsac: `row` and `column` stuff seems out of place as well.
Any comments to explain what is going on here?

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -57,6 +63,13 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
 
+    if (layer.projection == 'EPSG:3857') {
+        layer.options.tileMatrixSet = 'PM';
```
peppsac: From an outsider look this looks wrong: why WMS would need knowledge of tileMatrixSet?

src/Core/Geographic/Extent.js
```diff
@@ -90,7 +92,37 @@ Extent.prototype.as = function as(crs) {
...
+            const east = 180 - size * (nbCol - (this.col + 1));
+            const nbRow = nbCol;
+            const sizeRow = 1.0 / nbRow;
+            const Yn = 1 - sizeRow * (nbRow - (this.row));
+            const Ys = 1 - sizeRow * (nbRow - (this.row + 1));
+            const north = THREE.Math.radToDeg(projection.YToWGS84(Yn));
+            const south = THREE.Math.radToDeg(projection.YToWGS84(Ys));
+            const minCoord = new Coordinates('EPSG:4326', west, south).as(crs);
+            const maxCoord = new Coordinates('EPSG:4326', east, north).as(crs);
+            return new Extent(crs, {
```
peppsac: Maybe add a comment explaining what you're doing because it's not obvious.

peppsac: Also it seems weird to reference 4326 here and have a completely different computation 5 lines below for 4326?

peppsac (commented):
I trust you that it's correct, but comments would help a lot.

Also, while working on 3857 it would be nice to fix `split.html` example...

src/Core/Geographic/Extent.js
gchoqueux: i add comment
and fix error state crs

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: Previously the `tileMatrixSet` was implicit because the projections were identical. Now, it's necessary to specify in which `tileMatrixSet` the textures are placed

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: i rool back on that

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: it's same reason that PM tileMatrixSet

gchoqueux:
@peppsac  updated

gchoqueux:
> Also, while working on 3857 it would be nice to fix split.html example

It's access to image problem : `the access has been blocked by CORS policy`

peppsac:
> It's access to image problem : the access has been blocked by CORS policy

Hmm wmtsc pseudo protocol is in transient state. It's not tested in WMTS_Provider preprocess 
layer  but it's still declared as supported in Scheduler... But ok, we'll see this point later.

src/Core/Geographic/Extent.js
```diff
@@ -90,7 +92,34 @@ Extent.prototype.as = function as(crs) {
     assertCrsIsValid(crs);
 
     if (_isTiledCRS(this._crs)) {
-        throw new Error('Unsupported yet');
+        if (this._crs == 'WMTS:PM') {
```
peppsac, line 95: Can you add something describing the intent here - which seems to be: convert `this` to the requested crs by using `4326` as an intermediate state.

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -57,6 +67,16 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
     layer.style = layer.style || '';
     layer.transparent = layer.transparent || false;
 
+    // add tileMatrixSet destination to tile
```
peppsac: I still think that this doesn't belong to WMS_Provider.

Let me know if I understand correctly: the info you're adding here are helping to convert extent from the geometry to extent(s) in the layer's crs?

In which case maybe it would make sense to move this to `TileMesh.getCoordsForLayer`... WDYT?

gchoqueux:
@peppsac @elias75015 updated
add new : fix error conversion latitude 4326 to 3857

peppsac (commented):
Looks ok now. Can you rebase before merging please?

gchoqueux:
@peppsac rebased

peppsac (approved):

PR #538, opened 2017-10-19
feature (protocols): force loading step of group strategy

gchoqueux:
## Description
Force loading step of group strategy and don't use parent texture, if level is step of group

## Motivation and Context
When node use parent's texture elevation which is distant in the hierarchy,
OBB's size can be too big and it can be subdivided unnecessarily and the joints between the tiles are too different (see screenshot)

capture du 2017-10-19 11-27-38

**New steps to reproduce**
* Start localhost/index.htlm with only ortho layer (size `viewport 1100x1100`)
* when globe is loaded call in console : 
`globeView.controls.setCameraTargetGeoPositionAdvanced({ longitude:  6.886991921801097, latitude: 45.87929727669077, range: 2000});`

**Result in master, ~45 tiles of level 17  and ~420 wmts requests**
submaster

**Result in PR, ~40 tiles of level 16 and ~340 wmts requests**
downsub

the error comes from inheritance of textures elevations with a large geometric error and which induces more subdivisions

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -158,7 +158,10 @@ export function createGlobeLayer(id, options) {
 
     function subdivision(context, layer, node) {
         if (SubdivisionControl.hasEnoughTexturesToSubdivide(context, layer, node)) {
-            return globeSubdivisionControl(2, options.maxSubdivisionLevel || 17, options.sseSubdivisionThreshold || 1.0)(context, layer, node);
+            return globeSubdivisionControl(2,
+                options.maxSubdivisionLevel || 18,
+                options.sseSubdivisionThreshold || 1.0,
+                options.maxDeltaElevationLevel || 4)(context, layer, node);
         }
         return false;
     }
--- a/src/Core/Prefab/PlanarView.js
+++ b/src/Core/Prefab/PlanarView.js
@@ -9,7 +9,7 @@ import { GeometryLayer } from '../Layer/Layer';
 
 import { processTiledGeometryNode } from '../../Process/TiledNodeProcessing';
 import { updateLayeredMaterialNodeImagery, updateLayeredMaterialNodeElevation } from '../../Process/LayeredMaterialNodeProcessing';
-import { planarCulling, planarSubdivisionControl } from '../../Process/PlanarTileProcessing';
+import { planarCulling, planarSubdivisionControl, prePlanarUpdate } from '../../Process/PlanarTileProcessing';
 import PlanarTileBuilder from './Planar/PlanarTileBuilder';
 import SubdivisionControl from '../../Process/SubdivisionControl';
 
@@ -55,6 +55,8 @@ export function createPlanarLayer(id, extent, options) {
     tileLayer.preUpdate = (context, layer, changeSources) => {
         SubdivisionControl.preUpdate(context, layer);
 
+        prePlanarUpdate(context, layer);
+
         if (__DEBUG__) {
             layer._latestUpdateStartingLevel = 0;
         }
@@ -98,7 +100,8 @@ export function createPlanarLayer(id, extent, options) {
 
     function subdivision(context, layer, node) {
         if (SubdivisionControl.hasEnoughTexturesToSubdivide(context, layer, node)) {
-            return planarSubdivisionControl(options.maxSubdivisionLevel || 5)(context, layer, node);
+            return planarSubdivisionControl(options.maxSubdivisionLevel || 5,
+                options.maxDeltaElevationLevel || 4)(context, layer, node);
         }
         return false;
     }
--- a/src/Process/GlobeTileProcessing.js
+++ b/src/Process/GlobeTileProcessing.js
@@ -42,6 +42,15 @@ export function preGlobeUpdate(context, layer) {
 
     // pre-sse
     context.camera.preSSE = _preSSE(context.view);
+
+    const elevationLayers = context.view.getLayers((l, a) => a && a.id == layer.id && l.type == 'elevation');
+    context.maxElevationLevel = -1;
+    for (const e of elevationLayers) {
+        context.maxElevationLevel = Math.max(e.options.zoom.max, context.maxElevationLevel);
+    }
+    if (context.maxElevationLevel == -1) {
+        context.maxElevationLevel = Infinity;
+    }
 }
 
 function pointHorizonCulling(pt) {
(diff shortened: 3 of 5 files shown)


peppsac:
I'd like to try to reproduce the issue here.
Can you wrtie specific steps to reproduce it (for instance camera position/rotation of the above screenshot) ?

zarov:
@gchoqueux could you provide a camera position / instructions where the problem occurs, so we can see where the fix operates ?

gchoqueux:
~~it isn't the same consequences as the subdivisions. 
But the heritage altitudes (low zoom) also causes culling bug. because bounding box isn't representative (the altitudes with a low zoom have a large error)~~

This bug isn't the same issue exposed in this PR.
It comes from a calculation error of the bounding box (caused by the sub-sampling of searches in the elevation data)

culling3

**Steps to reproduce in globe**
```js
// 1 execute in console
globeView.controls.setCameraTargetGeoPositionAdvanced({ longitude:  6.886991921801097, latitude: 45.87929727669077, range: 2000});
// wait ending and  execute in console
globeView.controls.setCameraTargetGeoPositionAdvanced({ longitude:  6.886991921801097, latitude: 45.87929727669077, tilt: 70, heading: 180,  range: 1500});
```

gchoqueux:
@iTowns/reviewers branch ready and updated

src/Process/SubdivisionControl.js
```diff
@@ -4,9 +4,25 @@ export default {
...
+        context.forceElevation = [];
+        for (const e of context.elevationLayers) {
+            if (e.updateStrategy.type == 1 && e.updateStrategy.options.forceLoadingStep && e.updateStrategy.options.groups) {
+                context.forceElevation = context.forceElevation.concat(e.updateStrategy.options.groups);
+            }
+        }
+        context.forceElevation.sort((a, b) => a - b);
+    },
     hasEnoughTexturesToSubdivide: (context, layer, node) => {
+        if (context.forceElevation.length) {
```
peppsac: Can you document what are the next 5 lines doing, please?

examples/layers/JSONLayers/WORLD_DTM.json
```diff
@@ -8,7 +8,8 @@
     "updateStrategy": {
         "type": 1,
         "options": {
-            "groups": [3, 7, 9]
+            "groups": [3, 7, 9],
+            "forceLoadingStep": true
```
peppsac: It seems you added `forceLoadingStep = true` for all layers using the Group strategy.
Could it still make sense to use the Group stragegy without `forceLoadingStep`?

src/Process/SubdivisionControl.js
```diff
@@ -4,9 +4,25 @@ export default {
             (l, a) => a && a.id == layer.id && l.type == 'color');
         context.elevationLayers = context.view.getLayers(
             (l, a) => a && a.id == layer.id && l.type == 'elevation');
-    },
 
+        context.forceElevation = [];
+        for (const e of context.elevationLayers) {
+            if (e.updateStrategy.type == 1 && e.updateStrategy.options.forceLoadingStep && e.updateStrategy.options.groups) {
```
peppsac: (replace `1` with `STRATEGY_GROUP`)

src/Process/SubdivisionControl.js
```diff
@@ -4,9 +4,25 @@ export default {
...
+                context.forceElevation = context.forceElevation.concat(e.updateStrategy.options.groups);
+            }
+        }
+        context.forceElevation.sort((a, b) => a - b);
```
peppsac: Can you comment / use a more descriptive variable name?
I think `forceElevation` is a sorted array of all defined valid level for elevation layers using the group strategy. Is that correct?

src/Process/SubdivisionControl.js
peppsac: My guess is: you prevent subdivision of a node if its elevation level is < to max(valid group levels <= node.level)?

So you need to explain what you do and why you need to do it (maybe with an example?)

peppsac (changes requested):
Thanks for the repro steps.

Your fix needs a bit more comments.
Also did you consider adding a unit test for this issue?

examples/layers/JSONLayers/WORLD_DTM.json
gchoqueux: completely agree

src/Main.js
```diff
@@ -24,3 +24,6 @@ export { default as GeoJSON2Features } from './Renderer/ThreeExtended/GeoJSON2Fe
 export { default as FeaturesUtils } from './Renderer/ThreeExtended/FeaturesUtils';
 export { CONTROL_EVENTS } from './Renderer/ThreeExtended/GlobeControls';
 export { default as DEMUtils } from './utils/DEMUtils';
+// Export for test
+// Question: how export TileMesh only for the test?
+export { default as TileMesh } from './Core/TileMesh';
```
gchoqueux: How export `TileMesh` only for the test?
I modify `TileMesh` function ([here](https://github.com/iTowns/itowns/pull/538/files#diff-7c875b4c6fb106f12afe62df279296c5R7)) only for the test to not pollute the code and use.

autra: You can import TileMesh in tests without adding it in Main.js, you just need the correct path, see  #594 for instance (and the existing unit tests).

autra: And in #594 you have also some functions exported only for testing.

gchoqueux: @autra thanks!

gchoqueux:
branch updated

src/Process/SubdivisionControl.js
```diff
@@ -1,12 +1,33 @@
...
+                context.allGroupElevationLevels = context.allGroupElevationLevels.concat(e.updateStrategy.options.groups);
+            }
+        }
+        context.allGroupElevationLevels.sort((a, b) => a - b);
+    },
     hasEnoughTexturesToSubdivide: (context, layer, node) => {
+        // Prevent to subdivise the node if the current elevation level
+        // is below the target level in group strategy.
+        // we must avoid a tile, with level 20, inherits a level 3 elevation texture.
+        // The induced geometric error is much too large and distorts the SSE
```
peppsac: I fail to see why this is specific to the `STRATEGY_GROUP`?
Basically you want to prevent subdivision if `node.level - elevationLevel` is too big.

And IMHO, the bug should be fixed in the SSE logic (https://github.com/iTowns/itowns/blob/master/src/Process/LayeredMaterialNodeProcessing.js#L60 is another place that got patched for the same SSE logic issue).

gchoqueux: > I fail to see why this is specific to the STRATEGY_GROUP?

indeed, we can extend this logic to all strategies.

> And IMHO, the bug should be fixed in the SSE logic (https://github.com/iTowns/itowns/blob/master/src/Process/LayeredMaterialNodeProcessing.js#L60 is another place that got patched for the same SSE logic issue).

On the other hand, if this logic is placed at this point, the tile will be displayed first without any elevation texture, it will be a regression..

gchoqueux:
@peppsac  
I extend this logic to all strategy and moved it SSE processing

gchoqueux:
This PR is updated

gchoqueux:
**New steps to reproduce** in title

Bug related here https://github.com/iTowns/itowns/pull/538#issuecomment-351053532
doesn't correspond to the modification of this PR

peppsac (approved):
Ok, let's merge this PR: it's an acceptalbe workaround for a bug in the SSE mechanism.

PR #539, opened 2017-10-19
Improve lots of layer handling

peppsac:
This PR improves iTowns usage when displaying many geometry layers.

Commit 1:
```
feat (core): only use threeJsLayer feature when necessary

Using three.js' layers visibility has a big drawback: we're limited to 31 visibility
groups.
It's quite a low amount, so this commit changes the approach used for visibility:
  - GeometryLayer uses 'layer.object3d.visible' instead
  - layers attached to GeomLayer (like the TileDebug one) still use threeJsLayer,
    because their elements are attached to another layer's objects so they don't
    all share a common parent

Another change included is that now it doesn't make sense to use view.scene as a
default object3d parameter, so this commit switches to a Group instead.
```

Commit 2:
```
feat (core): simplify layer/object3d API

This commit simplifies API usage: layer.object3d is automatically added to the
View's scene - unless it already has a parent.
```

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -111,7 +111,6 @@
             $3dTilesLayerDiscreteLOD.visible = true;
 
             itowns.View.prototype.addLayer.call(globe, $3dTilesLayerDiscreteLOD);
-            globe.scene.add($3dTilesLayerDiscreteLOD.object3d);
 
             // Create a new Layer 3d-tiles For Viewer Request Volume
             // -----------------------------------------------------
@@ -133,7 +132,6 @@
             $3dTilesLayerRequestVolume.sseThreshold = 1;
             // add an event for have information when you move your mouse on a building
             itowns.View.prototype.addLayer.call(globe, $3dTilesLayerRequestVolume).then(function _() { window.addEventListener('mousemove', picking, false); })
-            globe.scene.add($3dTilesLayerRequestVolume.object3d);
 
             // Add the UI Debug
             var d = new debug.Debug(globe, menuGlobe.gui);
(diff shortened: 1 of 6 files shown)


examples/3dtiles.html
```diff
@@ -96,7 +96,7 @@
 
             // Create a new Layer 3d-tiles For DiscreteLOD
             // -------------------------------------------
-            var $3dTilesLayerDiscreteLOD = new itowns.GeometryLayer('3d-tiles-discrete-lod', globe.scene);
+            var $3dTilesLayerDiscreteLOD = new itowns.GeometryLayer('3d-tiles-discrete-lod', new itowns.THREE.Group());
```
autra, line 99: What about having Layer.object3D defaulted to a new THREE.Group [here](https://github.com/iTowns/itowns/pull/539/files#diff-61e012c3e00bdc510be33d9eac684b2aR61) instead of throwing?

autra (approved):
Just one minor proposal, but otherwise LGTM

examples/3dtiles.html
peppsac: If we silently create a THREE.Group the user will still need to add it to the scene. And if it's not added to the scene the layer will be invisible.

So I'd prefer throw here to make sure the user is aware that displaying the layer relies on a three object.

gchoqueux (approved):

examples/3dtiles.html
autra: > If we silently create a THREE.Group the user will still need to add it to the scene

Nope, because we add it [here](https://github.com/iTowns/itowns/pull/539/files#diff-d297dcf7b9d0e696ece4b1f6e1d7c69dR313) since this patch. It seems reasonable to have a `THREE.Group` by default to me, as you nearly always do:

```js
pointcloud = new itowns.GeometryLayer('pointcloud', new itowns.THREE.Group());
```
It'll just save the extra argument. (not a big deal though)

PR #544, opened 2017-10-20
fix (protocols): use correct WMS parameter names using the version

peppsac:
SRS/CRS naming and axis order changed with version 1.3.0 so let's use
the correct value by default.


gchoqueux:
you can rebase?

peppsac:
Rebased

gchoqueux (approved):

PR #546, opened 2017-10-24
Layer management

peppsac:
## Description

This PR tries to improve iTowns API for layer management.

## Motivation and Context
Our API is currently a bit confusing so I tried to find something simpler that would support all our use cases.

Here are the proposed simplifications:
- top level / geometry layers are added to a *View* but other layers (eg color layers) are added to geometry layers
- instead of having a unique entry point to add layers `View.addLayer`,  this PR adds different entry point for each supported layer type
- `GlobeView.wgs84TileLayer` and `PlanarView.tileLayer` become `XxxxView.baseLayer`

This PR can be summarized with an example:
```js
// before
globeView.addLayer(colorLayer);
itowns.View.prototype.addLayer.call(globeView, anotherGeomLayer);

// after
globeView.baseLayer.addColorLayer(colorLayer);
globeView.addLayer(anotherGeomLayer);
```

I also cherry-picked the content of PR #468 to add the ability to remove a layer:

```js
mylayer.removeFeatureLayer('tcl_bus'); // removes feature layer
view.removeLayer(layer); // removes a top level layer
```

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -88,7 +88,7 @@
             var globe = new itowns.GlobeView(viewerDiv, positionOnGlobe);
             var menuGlobe = new GuiTools('menuDiv', globe, 300);
 
-            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(function (result) { return globe.addLayer(result) });
+            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(globe.baseLayer.addColorLayer);
 
             // function use :
             // For preupdate Layer geomtry :
@@ -96,46 +96,26 @@
 
             // Create a new Layer 3d-tiles For DiscreteLOD
             // -------------------------------------------
-            var $3dTilesLayerDiscreteLOD = new itowns.GeometryLayer('3d-tiles-discrete-lod', new itowns.THREE.Group());
+            var $3dTilesLayerDiscreteLOD = itowns.create3dTiles(
+                '3d-tiles-discrete-lod', {
+                    url: 'https://raw.githubusercontent.com/AnalyticalGraphicsInc/3d-tiles-samples/master/tilesets/TilesetWithDiscreteLOD/tileset.json'
+                });
 
-            $3dTilesLayerDiscreteLOD.preUpdate = preUpdateGeo;
-            $3dTilesLayerDiscreteLOD.update = itowns.process3dTilesNode(
-                itowns.$3dTilesCulling,
-                itowns.$3dTilesSubdivisionControl
-            );
-            $3dTilesLayerDiscreteLOD.name = 'DiscreteLOD';
-            $3dTilesLayerDiscreteLOD.url = 'https://raw.githubusercontent.com/AnalyticalGraphicsInc/3d-tiles-samples/master/tilesets/TilesetWithDiscreteLOD/tileset.json';
-            $3dTilesLayerDiscreteLOD.protocol = '3d-tiles'
-            $3dTilesLayerDiscreteLOD.overrideMaterials = true;  // custom cesium shaders are not functional
-            $3dTilesLayerDiscreteLOD.type = 'geometry';
-            $3dTilesLayerDiscreteLOD.visible = true;
-
-            itowns.View.prototype.addLayer.call(globe, $3dTilesLayerDiscreteLOD);
+            globe.addLayer($3dTilesLayerDiscreteLOD);
 
             // Create a new Layer 3d-tiles For Viewer Request Volume
             // -----------------------------------------------------
 
-            var $3dTilesLayerRequestVolume = new itowns.GeometryLayer('3d-tiles-request-volume', new itowns.THREE.Group());
-
-            $3dTilesLayerRequestVolume.preUpdate = preUpdateGeo;
-            $3dTilesLayerRequestVolume.update = itowns.process3dTilesNode(
-                itowns.$3dTilesCulling,
-                itowns.$3dTilesSubdivisionControl
-            );
-
-            $3dTilesLayerRequestVolume.name = 'RequestVolume';
-            $3dTilesLayerRequestVolume.url = 'https://raw.githubusercontent.com/AnalyticalGraphicsInc/3d-tiles-samples/master/tilesets/TilesetWithRequestVolume/tileset.json';
-            $3dTilesLayerRequestVolume.protocol = '3d-tiles'
-            $3dTilesLayerRequestVolume.overrideMaterials = true;  // custom cesium shaders are not functional
-            $3dTilesLayerRequestVolume.type = 'geometry';
-            $3dTilesLayerRequestVolume.visible = true;
-            $3dTilesLayerRequestVolume.sseThreshold = 1;
-            // add an event for have information when you move your mouse on a building
-            itowns.View.prototype.addLayer.call(globe, $3dTilesLayerRequestVolume).then(function _() { window.addEventListener('mousemove', picking, false); })
+            var $3dTilesLayerRequestVolume = itowns.create3dTiles(
+                '3d-tiles-request-volume', {
+                    url: 'https://raw.githubusercontent.com/AnalyticalGraphicsInc/3d-tiles-samples/master/tilesets/TilesetWithRequestVolume/tileset.json'
+                });
+            globe.addLayer($3dTilesLayerRequestVolume).then(function _() { window.addEventListener('mousemove', picking, false);
+            });
 
             // Add the UI Debug
             var d = new debug.Debug(globe, menuGlobe.gui);
-            debug.createTileDebugUI(menuGlobe.gui, globe, globe.wgs84TileLayer, d);
+            debug.createTileDebugUI(menuGlobe.gui, globe, globe.baseLayer, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, globe, $3dTilesLayerDiscreteLOD, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, globe, $3dTilesLayerRequestVolume, d);
 
--- a/examples/cubic_planar.js
+++ b/examples/cubic_planar.js
@@ -94,8 +94,7 @@ for (index = 0; index < wmsLayers.length; index++) {
 
     view.addLayer(tileLayer);
 
-    view.addLayer({
-        update: itowns.updateLayeredMaterialNodeImagery,
+    tileLayer.addColorLayer({
         url: 'https://download.data.grandlyon.com/wms/grandlyon',
         networkOptions: { crossOrigin: 'anonymous' },
         type: 'color',
@@ -107,10 +106,9 @@ for (index = 0; index < wmsLayers.length; index++) {
         options: {
             mimetype: 'image/jpeg',
         },
-    }, tileLayer);
+    });
 
-    view.addLayer({
-        update: itowns.updateLayeredMaterialNodeElevation,
+    tileLayer.addElevationLayer({
         url: 'https://download.data.grandlyon.com/wms/grandlyon',
         type: 'elevation',
         protocol: 'wms',
@@ -123,7 +121,7 @@ for (index = 0; index < wmsLayers.length; index++) {
         options: {
             mimetype: 'image/jpeg',
         },
-    }, tileLayer);
+    });
 
     // Since the elevation layer use color textures, specify min/max z
     tileLayer.materialOptions = {
--- a/examples/cubic_planar.js
+++ b/examples/cubic_planar.js
@@ -97,7 +97,6 @@ for (index = 0; index < wmsLayers.length; index++) {
     tileLayer.addColorLayer({
         url: 'https://download.data.grandlyon.com/wms/grandlyon',
         networkOptions: { crossOrigin: 'anonymous' },
-        type: 'color',
         protocol: 'wms',
         version: '1.3.0',
         id: 'wms_imagery' + wms + index,
@@ -110,7 +109,6 @@ for (index = 0; index < wmsLayers.length; index++) {
 
     tileLayer.addElevationLayer({
         url: 'https://download.data.grandlyon.com/wms/grandlyon',
-        type: 'elevation',
         protocol: 'wms',
         networkOptions: { crossOrigin: 'anonymous' },
         version: '1.3.0',
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -14,9 +14,6 @@ var miniDiv = document.getElementById('miniDiv');
 
 // Instanciate iTowns GlobeView*
 var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, { renderer: renderer });
-function addLayerCb(layer) {
-    return globeView.addLayer(layer);
-}
 
 // Dont' instance mini viewer if it's Test env
 if (!renderer) {
@@ -48,17 +45,19 @@ if (!renderer) {
     };
 
     // Add one imagery layer to the miniview
-    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) { miniView.addLayer(layer); });
+    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) {
+        miniView.baseLayer.addColorLayer(layer);
+    });
 }
 
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(globeView.baseLayer.addColorLayer));
 // Add two elevation layers.
 // These will deform iTowns globe geometry to represent terrain elevation.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(globeView.baseLayer.addElevationLayer));
+promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(globeView.baseLayer.addElevationLayer));
 
 exports.view = globeView;
 exports.initialPosition = positionOnGlobe;
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -45,9 +45,7 @@ if (!renderer) {
     };
 
     // Add one imagery layer to the miniview
-    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) {
-        miniView.baseLayer.addColorLayer(layer);
-    });
+    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(miniView.baseLayer.addColorLayer);
 }
 
 // Add one imagery layer to the scene
--- a/examples/wfs.js
+++ b/examples/wfs.js
@@ -22,10 +22,10 @@ viewerDiv = document.getElementById('viewerDiv');
 
 // Instanciate PlanarView*
 view = new itowns.PlanarView(viewerDiv, extent, { renderer: renderer });
-view.tileLayer.disableSkirt = true;
+view.baseLayer.disableSkirt = true;
 
 // Add an WMS imagery layer (see WMS_Provider* for valid options)
-view.addLayer({
+view.baseLayer.addColorLayer({
     url: 'https://download.data.grandlyon.com/wms/grandlyon',
     networkOptions: { crossOrigin: 'anonymous' },
     type: 'color',
@@ -59,7 +59,7 @@ function colorLine(properties) {
     return new itowns.THREE.Color(rgb[0] / 255, rgb[1] / 255, rgb[2] / 255);
 }
 
-view.addLayer({
+view.baseLayer.addFeatureLayer({
     update: itowns.FeatureProcessing.update,
     convert: itowns.Feature2Mesh.convert({
         color: colorLine }),
@@ -80,7 +80,7 @@ view.addLayer({
     options: {
         mimetype: 'geojson',
     },
-}, view.tileLayer);
+});
 
 function colorBuildings(properties) {
     if (properties.id.indexOf('bati_remarquable') === 0) {
@@ -114,8 +114,7 @@ scaler = {
 };
 
 view.addFrameRequester(scaler);
-view.addLayer({
-    type: 'geometry',
+view.baseLayer.addFeatureLayer({
     update: itowns.FeatureProcessing.update,
     convert: itowns.Feature2Mesh.convert({
         color: colorBuildings,
@@ -142,6 +141,6 @@ view.addLayer({
     options: {
         mimetype: 'json',
     },
-}, view.tileLayer);
+});
 
 exports.view = view;
--- a/src/Core/DefaultGeometryLayers.js
+++ b/src/Core/DefaultGeometryLayers.js
@@ -187,7 +187,7 @@ export function createPlane(id, extent, options) {
  * @function
  * @param {string} id - identifer (must be unique amongst all layers id)
  * @param {Object} options
- * @param {URL} options.url - URL of the tileset
+ * @param {string} options.url - URL of the tileset
  * @param {number} [options.sseThreshold = 16] - refinement threshold
  * @param {number} [options.cleanupDelay = 1000] - delay (in ms) after which unused
  * tiles are removed.
@@ -242,6 +242,7 @@ export function createPointcloud(id, options) {
     const layer = new GeometryLayer(id, options.object3d || new THREE.Group());
     layer.file = options.file;
     layer.protocol = 'potreeconverter';
+    layer.type = 'pointcloud';
     layer.url = options.url.href;
     layer.table = options.lopocsTable;
 
--- a/src/Core/DefaultGeometryLayers.js
+++ b/src/Core/DefaultGeometryLayers.js
@@ -1,0 +1,194 @@
+import * as THREE from 'three';
+
+import { GeometryLayer } from './Layer/Layer';
+import { processTiledGeometryNode } from '../Process/TiledNodeProcessing';
+import SubdivisionControl from '../Process/SubdivisionControl';
+
+import { globeCulling, preGlobeUpdate, globeSubdivisionControl, globeSchemeTileWMTS, globeSchemeTile1 } from '../Process/GlobeTileProcessing';
+import BuilderEllipsoidTile from './Prefab/Globe/BuilderEllipsoidTile';
+
+import { planarCulling, planarSubdivisionControl } from '../Process/PlanarTileProcessing';
+import PlanarTileBuilder from './Prefab/Planar/PlanarTileBuilder';
+
+import { process3dTilesNode, $3dTilesCulling, $3dTilesSubdivisionControl, pre3dTilesUpdate } from '../Process/3dTilesProcessing';
+
+function _commonAncestorLookup(a, b) {
+    if (!a || !b) {
+        return undefined;
+    }
+    if (a.level == b.level) {
+        if (a.id == b.id) {
+            return a;
+        } else if (a.level != 0) {
+            return _commonAncestorLookup(a.parent, b.parent);
+        } else {
+            return undefined;
+        }
+    } else if (a.level < b.level) {
+        return _commonAncestorLookup(a, b.parent);
+    } else {
+        return _commonAncestorLookup(a.parent, b);
+    }
+}
+
+function _tiledNodeInitFunction(layer, parent, node) {
+    node.material.setLightingOn(layer.lighting.enable);
+    node.material.uniforms.lightPosition.value = layer.lighting.position;
+
+    if (layer.noTextureColor) {
+        node.material.uniforms.noTextureColor.value.copy(layer.noTextureColor);
+    }
+
+    if (__DEBUG__) {
+        node.material.uniforms.showOutline = { value: layer.showOutline || false };
+        node.material.wireframe = layer.wireframe || false;
+    }
+}
+
+function _tiledPreUpdate(preUpdateSpecialisation) {
+    return function _(context, layer, changeSources) {
+        SubdivisionControl.preUpdate(context, layer);
+
+        if (__DEBUG__) {
+            layer._latestUpdateStartingLevel = 0;
+        }
+
+        if (preUpdateSpecialisation) {
+            preUpdateSpecialisation(context, layer);
+        }
+        if (changeSources.has(undefined) || changeSources.size == 0) {
+            return layer.level0Nodes;
+        }
+
+        let commonAncestor;
+        for (const source of changeSources.values()) {
+            if (source.isCamera) {
+                // if the change is caused by a camera move, no need to bother
+                // to find common ancestor: we need to update the whole tree:
+                // some invisible tiles may now be visible
+                return layer.level0Nodes;
+            }
+            if (source.layer === layer.id) {
+                if (!commonAncestor) {
+                    commonAncestor = source;
+                } else {
+                    commonAncestor = _commonAncestorLookup(commonAncestor, source);
+                    if (!commonAncestor) {
+                        return layer.level0Nodes;
+                    }
+                }
+                if (commonAncestor.material == null) {
+                    commonAncestor = undefined;
+                }
+            }
+        }
+        if (commonAncestor) {
+            if (__DEBUG__) {
+                layer._latestUpdateStartingLevel = commonAncestor.level;
+            }
+            return [commonAncestor];
+        } else {
+            return layer.level0Nodes;
+        }
+    };
+}
+
+export function createGlobe(id, options) {
+    // Configure tiles
+    const wgs84TileLayer = new GeometryLayer(id, options.object3d || new THREE.Group());
+    wgs84TileLayer.schemeTile = globeSchemeTileWMTS(globeSchemeTile1);
+    wgs84TileLayer.extent = wgs84TileLayer.schemeTile[0].clone();
+    for (let i = 1; i < wgs84TileLayer.schemeTile.length; i++) {
+        wgs84TileLayer.extent.union(wgs84TileLayer.schemeTile[i]);
+    }
+    wgs84TileLayer.preUpdate = _tiledPreUpdate(preGlobeUpdate);
+
+    function subdivision(context, layer, node) {
+        if (SubdivisionControl.hasEnoughTexturesToSubdivide(context, layer, node)) {
+            return globeSubdivisionControl(2, options.maxSubdivisionLevel || 17, options.sseSubdivisionThreshold || 1.0)(context, layer, node);
+        }
+        return false;
+    }
+
+    wgs84TileLayer.update = processTiledGeometryNode(globeCulling(2), subdivision);
+    wgs84TileLayer.builder = new BuilderEllipsoidTile();
+    wgs84TileLayer.onTileCreated = _tiledNodeInitFunction;
+    wgs84TileLayer.type = 'geometry';
+    wgs84TileLayer.protocol = 'tile';
+    wgs84TileLayer.visible = true;
+    wgs84TileLayer.lighting = {
+        enable: false,
+        position: { x: -0.5, y: 0.0, z: 1.0 },
+    };
+
+    return wgs84TileLayer;
+}
+
+
+export function createPlane(id, extent, options) {
+    const tileLayer = new GeometryLayer(id, options.object3d || new THREE.Group());
+    tileLayer.extent = extent;
+    tileLayer.schemeTile = [extent];
+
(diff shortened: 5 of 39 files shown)


examples/globe.js
```diff
@@ -48,17 +45,19 @@ if (!renderer) {
     };
 
     // Add one imagery layer to the miniview
-    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) { miniView.addLayer(layer); });
+    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) {
```
mbredif: why not the following ?
` itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(miniView.baseLayer.addColorLayer); `

peppsac: You're right. I pushed a fix, thanks.

gcebelieu:
Does this PR also preserves any backward compatibility with the being removed old API (considered as deprecated), I mean, at least for next release minor iTowns release for instance ?

More generaly, are there rules defined in iTowns to handle API changes ? If not, I think it would be important to establish such rules.

WDYT ?

peppsac:
> Does this PR also preserves any backward compatibility with the being removed old API (considered as deprecated), I mean, at least for next release minor iTowns release for instance ?

This PR doesn't really remove any old API (well except for one bit but I'll restore it).
The idea is to add an easier to use API, built on top of the existing API.

> More generaly, are there rules defined in iTowns to handle API changes ?
> If not, I think it would be important to establish such rules.

I dunno. We could open an issue to discuss this topic.

Mkonini:
I agree, we must open an issue to decide some rules about how API evolutions will be handled, before starting any modification on it.

peppsac:
Don't forget that rules imply an overhead and that itowns is developped by a small group of dev.

I don't think we have the manpower nor the infrastructure to maintain several parallel branches.
The only safe rule we can commit to is to bump itowns major version number on important API changes (ie: the release with this PR merged would be 2.2.0, and a release without could be 2.1.1).

Mkonini:
My opinion: rules do not create any overhead, inversely having no rules contribute to increase technical debt, so overhead.
You're proposing a rule, let's consider that as a first step to go ahead in the discussion surely not the last.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,96 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+    };
+
+    layer.removeLayer = (layerOrId) => {
+        const layerId = layerOrId.id === undefined ? layerOrId : layerOrId.id;
+        const toRemove = view.getLayers(l => l.id === layerId)[0];
+        if (layer.type === 'color') {
+            return layer.removeColorLayer(toRemove);
+        } else if (layer.type === 'elevation') {
+            return layer.removeElevationLayer(toRemove);
+        } else {
```
autra: so feature layers are actually everything that are neither elevation nor color right?

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,96 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+            }
+            return true;
+        } else {
+            // eslint-disable-next-line no-console
+            console.error(`${featureLayerOrId} isn't a feature layer`);
+            return false;
+        }
+    };
+
+    layer.removeLayer = (layerOrId) => {
```
autra, line 126: Maybe only set removeLayer if it does not already exists. That way we give a chance to user to implement their own if they have other layer types.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,96 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+
+        elevationLayer.update = elevationLayer.update || updateLayeredMaterialNodeElevation;
+
+        return view.addLayer(elevationLayer, layer);
+    };
+    layer.removeElevationLayer = (elevationLayerOrId) => {
+        const layerId = elevationLayerOrId.id === undefined ? elevationLayerOrId : elevationLayerOrId.id;
+        const elevationLayer = view.getLayers(l => l.id === layerId)[0];
+        if (elevationLayer && elevationLayer.type === 'elevation' && layer.detach(elevationLayer)) {
+            // TODO: cleanup elevation textures
```
autra, line 97: Please consider implementing this here if not too complex.

gcebelieu:
> This PR doesn't really remove any old API (well except for one bit but I'll restore it).
> The idea is to add an easier to use API, built on top of the existing API.

OK, thanks for the explanation. Great if the removed bit is to be restored (even at least in a deprecated status).

autra (approved):
This looks good to me, with a few comments to address.

Concerning the whole "breaking change" part: if I understand correctly, the only change here is that `createGlobeLayer` is renamed to `createGlobe`, but I guess people use `new GlobeLayer` and not `createGlobe` directly (right @gcebelieu)? Did I miss another change? 

So this discussion is not really relevant here... But as we started it: little reminder that @peppsac has been advocating to use Angular convention [here](https://github.com/iTowns/itowns/issues/518). One of the goal was to automatically generate changelogs with a list of breaking changes if existing, making migration from one version to another easy, if not trivial. In this topic, documentation and communication are key, really.

peppsac:
The `createGlobeLayer` -> `createGlobe` is not the important part - actually I'm not entirely sold on this idea.
The real API change is the one shown in the top post: instead of using `view.addLayer` for all layers, "attached layers" now are added to their parent layer throuh the specialized `addXXXLayer`.

src/Core/Scheduler/Providers/TileProvider.js
peppsac: Right now this PR is only about code shuffling: I took the 'remove color layer' code from GlobeView but 'elevation layer' removal isn't really supported yet.

src/Core/Scheduler/Providers/TileProvider.js
peppsac: This `removeLayer` function is only used to ease writing the code that removes a base layer and all it's attached layer. I can remove it completely if you think it's confusing?

src/Core/Scheduler/Providers/TileProvider.js
peppsac: Not really :)
But feature layer don't have a type currently, hence the current implementation.
We should probably enforce the "must-have-a-type-property" requirement for all layers though...

autra:
> The real API change is the one shown in the top post

+1 on this API change btw. It's a lot clearer, and makes it explicit that geometry layers and color + elevation layers are not quite the same (the former being a base for the latter).

You left some instances of `this._view.wgs84TileLayer` in GlobeControls that should be replaced by `baseLayer` (i have errors in console).

peppsac:
>  You left some instances of this._view.wgs84TileLayer in GlobeControls that should be replaced by baseLayer (i have errors in console).

Thanks @autra : fixed.

I pushed a commit that restore the globeView.add(layer) feature + adds a deprecated warning.
So this PR shouldn't break any working code.

Mkonini:
Let us the time to verify, wait for @gcebelieu answer to @autra question before merge!

autra:
@Mkonini no need for that, [I have my answer](https://github.com/IGNF/geoportal-extensions/blob/master/src/Itowns/GlobeViewExtended.js#L28)

That being said, I'm interested in other reviewers input on this depreciation.

elias75015:
Hi. I just tried this PR with the main iTowns example. 
It looks like the removeLayer function is broken.

If we do a : 
`globeView.removeLayer("Ortho")`

we got the following error : 

> View.js?d297:394 Uncaught TypeError: Cannot read property 'id' of undefined
>     at eval (TileProvider.js?4857:64)
>     at GlobeView.getLayers (View.js?d297:397)
>     at GeometryLayer.layer.removeColorLayer (TileProvider.js?4857:64)
>     at GlobeView.removeImageryLayer [as removeLayer] (GlobeView.js?caf5:233)
>     at <anonymous>:1:11

autra:
@elias75015 set your layer type correctly, see https://github.com/iTowns/itowns/pull/546/files#diff-caf5227be4f9e2f283d82df32a678e4bR232

peppsac:
@elias75015 thanks for testing this PR!

I pushed a fix regarding layer removal + I added tests to cover this as well.
Let me know if it works now.

elias75015:
Thanks for the fix, it works now.

However, I meet another problem with the PR (reproducible on the itowns globe sample). If we add a listener on the globe for the layer_added event to list the layers :  

`globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.LAYER_ADDED, function(layerid) {
		console.log(globeView.getLayers());}
);`

Then we remove/add a layer : 
```
globeView.removeLayer("Ortho");

globeView.addLayer({
    "type": "color",
    "protocol":   "wmts",
    "id":         "Ortho",
    ...
});
```

The just added layer is not returned with the other layers returned by the getLayers(); we put in the listener.

peppsac:
@elias75015 indeed. I fixed this too.

elias75015:
@peppsac Thanks! On my side, nothing else to add.

examples/3dtiles.html
```diff
@@ -88,54 +88,34 @@
...
-            $3dTilesLayerDiscreteLOD.type = 'geometry';
-            $3dTilesLayerDiscreteLOD.visible = true;
-
-            itowns.View.prototype.addLayer.call(globe, $3dTilesLayerDiscreteLOD);
+            var $3dTilesLayerDiscreteLOD = itowns.create3dTiles(
+                '3d-tiles-discrete-lod', {
+                    url: 'https://raw.githubusercontent.com/AnalyticalGraphicsInc/3d-tiles-samples/master/tilesets/TilesetWithDiscreteLOD/tileset.json'
+                });
+
+            globe.addLayer($3dTilesLayerDiscreteLOD);
```
gchoqueux, line 104: I notice this variable name isn't correct, it can be confusing with `baseLayer` that represents the globe
it will be better to have `globeView`,

examples/cubic_planar.js
```diff
@@ -94,8 +94,7 @@ for (index = 0; index < wmsLayers.length; index++) {
 
     view.addLayer(tileLayer);
 
-    view.addLayer({
-        update: itowns.updateLayeredMaterialNodeImagery,
+    tileLayer.addColorLayer({
```
gchoqueux, line 97: the name `addColorLayer` is really necessary, because layer are typed.
we could imagine :
 - Either we remove the `layer.type: color` with `addColorLayer`, since it's redundant
 - Either we keep the `layer.type: color` with  `addLayer` which differentiates it from the layer types

src/Core/DefaultGeometryLayers.js
```diff
@@ -0,0 +1,194 @@
+import * as THREE from 'three';
```
gchoqueux, line 1: Could you add explanatory comments for non comment functions and add documentation for export functions?

src/Core/DefaultGeometryLayers.js
```diff
@@ -0,0 +1,194 @@
...
+            return undefined;
+        }
+    } else if (a.level < b.level) {
+        return _commonAncestorLookup(a, b.parent);
+    } else {
+        return _commonAncestorLookup(a.parent, b);
+    }
+}
+
+function _tiledNodeInitFunction(layer, parent, node) {
```
gchoqueux, line 34: There is only the material that initializes in this function. the name and parameters are they well suited?

src/Core/DefaultGeometryLayers.js
```diff
@@ -0,0 +1,194 @@
...
+            return [commonAncestor];
+        } else {
+            return layer.level0Nodes;
+        }
+    };
+}
+
+export function createGlobe(id, options) {
+    // Configure tiles
+    const wgs84TileLayer = new GeometryLayer(id, options.object3d || new THREE.Group());
```
gchoqueux: why not call it `globe`?

src/Core/DefaultGeometryLayers.js
```diff
@@ -0,0 +1,194 @@
...
+        enable: false,
+        position: { x: -0.5, y: 0.0, z: 1.0 },
+    };
+
+    return wgs84TileLayer;
+}
+
+
+export function createPlane(id, extent, options) {
+    const tileLayer = new GeometryLayer(id, options.object3d || new THREE.Group());
```
gchoqueux: same suggestion: renamed `tileLayer` to `plane`?

src/Core/DefaultGeometryLayers.js
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export function create3dTiles(id, options) {
+    if (options.url) {
+        options.url = new URL(options.url, document.location);
+    }
+    if (!options.url) {
+        throw new Error('options.url must point to the tileset to be used');
+    }
+    const layer = new GeometryLayer(id, options.object3d || new THREE.Group());
```
gchoqueux, line 204: Same suggestion: renamed `layer` to `3dTiles` or `3dTilesLayer`?

src/Core/Prefab/GlobeView.js
```diff
@@ -207,7 +98,7 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     this.camera.camera3D.updateProjectionMatrix();
     this.camera.camera3D.updateMatrixWorld(true);
 
-    const wgs84TileLayer = createGlobeLayer('globe', options);
+    const wgs84TileLayer = createGlobe('globe', options);
```
gchoqueux, line 101: will it be clearer to do?

```js
this.globe = createGlobe('globe', options);
```

or to make a getter `globe` that returns `this.baseLayer`, 
but I feel that `baseLayer` isn't necessary

src/Core/Prefab/GlobeView.js
```diff
@@ -286,29 +177,47 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
 GlobeView.prototype = Object.create(View.prototype);
 GlobeView.prototype.constructor = GlobeView;
 
-GlobeView.prototype.addLayer = function addLayer(layer) {
+GlobeView.prototype._preAddLayer = function _preAddLayer(layer) {
```
gchoqueux: could you comment on the purpose of this function?

gchoqueux: It's possible to factorize in `viewer` or in other place the common things between `planarView` and `globeView`

src/Core/Prefab/PlanarView.js
```diff
@@ -144,28 +143,47 @@ function PlanarView(viewerDiv, extent, options = {}) {
     this._renderState = RendererConstant.FINAL;
     this._fullSizeDepthBuffer = null;
 
-    this.tileLayer = tileLayer;
+    this.baseLayer = tileLayer;
```
gchoqueux, line 146: same comment
`this.plane = createPlane(...);`

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,100 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
         throw new Error(`Cannot init tiled layer without schemeTile for layer ${layer.id}`);
     }
 
+    layer.addColorLayer = (colorLayer, addToView = true) => {
+        if (colorLayer.protocol === 'rasterizer') {
+            colorLayer.reprojection = 'EPSG:4326';
```
gchoqueux: redundancy with `_preAddLayer` is for backward compatibility?

gchoqueux: it could be factorize?

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,100 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
         throw new Error(`Cannot init tiled layer without schemeTile for layer ${layer.id}`);
     }
 
+    layer.addColorLayer = (colorLayer, addToView = true) => {
```
gchoqueux, line 37: you need the documentation

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,100 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+            const colorLayerCount = view.getLayers(l => l.type === 'color').length;
+            colorLayer.sequence = colorLayerCount;
+            colorLayer.update = updateLayeredMaterialNodeImagery;
+        }
+
+        if (addToView) {
+            return view.addLayer(colorLayer, layer);
+        }
+    };
+    layer.removeColorLayer = (colorLayerOrId) => {
```
gchoqueux, line 53: same, you need the documentation

src/Core/View.js
```diff
@@ -318,6 +322,45 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+            elt.material.dispose();
+        }
+        elt.parent = null;
+    });
+    layer.object3d.remove(...layer.object3d.children);
+
+    this.scene.remove(layer.object3d);
+
+    if (layerId == this.baseLayer.id) {
+        delete this.baseLayer;
```
gchoqueux, line 373: `baseLayer` is it really useful? 
it isn't already in `this._layers`

gchoqueux (changes requested):

examples/3dtiles.html
peppsac: I agree but you don't have to wait for this PR to fix this: you can right now open a new PR to rename variables.

examples/cubic_planar.js
peppsac: I clearly prefer explicit (`addColorLayer`) method names than generic ones and this PR moves the API in this direction.

And yes the `type` attribute will become a read-only property (rather than a user defined property)

src/Core/DefaultGeometryLayers.js
peppsac: I'll do this later (but before merging of course) when we agree that the proposed API is fine.

src/Core/DefaultGeometryLayers.js
peppsac: The paramers come from `TileNodedProcessing`: _tiledNodeInitFunction is just a callback

peppsac:
So it seems you're all fine with the basic direction of this PR.

I'm going to put a bit more work into it, especially to ease the documentation of all the `addXXXLayer` functions.
I'll post an update when it's ready to review.

peppsac:
I pushed an update, improving the documentation.
WDYT?

src/Core/Layer/Layer.js
```diff
@@ -175,4 +182,77 @@ const ImageryLayers = {
...
+function CanDisplayColorLayer() {}
+/**
+ * Add a color layer to display on this geometry
+ *
+ * @function
+ * @name CanDisplayColorLayer#addColorLayer
+ * @param {Object} colorLayer - property bag defining the layer to display.
+ * @param {string} colorLayer.id - unique identifier of this layer
+ * @param {string} colorLayer.protocol - one of wms, wmts, wmtsc, tms, static.
+ * Each protocol has a set of specific options that are documented in XXX.
```
Jeremy-Gaillard: "in XXX"?

src/Core/Layer/Layer.js
```diff
@@ -175,4 +182,77 @@ const ImageryLayers = {
...
+function CanDisplayElevationLayer() {}
+/**
+ * Use an elevation data source to deform the underlying geometry
+ *
+ * @function
+ * @name CanDisplayElevationLayer#addElevationLayer
+ * @param {Object} elevationLayer - property bag defining the layer to display.
+ * @param {string} elevationLayer.id - unique identifier of this layer
+ * @param {string} elevationLayer.protocol - one of wms, wmts, wmtsc, static.
+ * Each protocol has a set of specific options that are documented in XXX.
```
Jeremy-Gaillard: Same as above

src/Core/Layer/Layer.js
```diff
@@ -53,6 +53,13 @@ export const defineLayerProperty = function defineLayerProperty(layer, propertyN
     }
 };
 
+/**
+ * A GeometryLayer defines a geometry: globe, plane or pointcloud.
+ * This geometry can then be used to display other datas, like color layers.
```
Jeremy-Gaillard: No "s" at the end of data.

src/Core/Layer/Layer.js
```diff
@@ -53,6 +53,13 @@ export const defineLayerProperty = function defineLayerProperty(layer, propertyN
...
+ * @constructor
+ * @param {string} id - unique id
+ * @param {Object} object3d - the THREE.Object3D will be the parent of the object hierarchy
```
Jeremy-Gaillard: "the THREE.Object3D **that** will be the parent of the **layer's** object hierarchy"

src/Core/Layer/Layer.js
```diff
@@ -175,4 +182,77 @@ const ImageryLayers = {
...
+ * @param {({Layer}|string)} - the layer to remove or its id
+ */
+
+ /**
+ * Interface implemented by {@link GeometryLayer} instances capable of displaying a feature layers
+ * @interface
+ */
+function CanDisplayFeatureLayer() {}
+/**
+ * Display a feature layers on the underlying geometry
```
Jeremy-Gaillard: I think we need a definition of what a feature layer is. I guess it is a layer made of data fetched from a wfs (or wfs-like) source?

src/Core/DefaultGeometryLayers.js
```diff
@@ -93,7 +93,21 @@ function _tiledPreUpdate(preUpdateSpecialisation) {
     };
 }
 
-export function createGlobe(id, options) {
+/**
+ * Creates a Globe geometry, which can be used to draw other layers (color, elevation, ...)
```
Jeremy-Gaillard: "Creates a **g**lobe geometry, **on** which other layers (color, elevation, ...) **can be drawn**"

src/Core/DefaultGeometryLayers.js
```diff
@@ -93,7 +93,21 @@ function _tiledPreUpdate(preUpdateSpecialisation) {
...
+ *
+ * @function
+ * @param {string} id - identifer (must be unique amongst all layers id)
+ * @param {Object} [options]
+ * @param {Object} [options.object3d=THREE.Group] - The THREE.Object3D that will be the parent of all Object3D built
+ * by this layer
+ * @param {number} [options.maxSubdivisionLevel=17] - The geometry subdivision max depth
```
Jeremy-Gaillard: "subdivision **'s**"

src/Core/DefaultGeometryLayers.js
```diff
@@ -124,7 +138,21 @@ export function createGlobe(id, options) {
...
+/**
+ * Creates a planar geometry, which can be used to draw other layers (color, elevation, ...)
+ *
+ * @function
+ * @param {string} id - identifer (must be unique amongst all layers id)
+ * @param {Extent} extent - the extent of the planar geometry
+ * @param {Object} [options]
+ * @param {Object} [options.object3d=THREE.Group] - The THREE.Object3D that will be the parent of all Object3D built
+ * by this layer
+ * @param {number} [options.maxSubdivisionLevel=5] - The geometry subdivision max depth
```
Jeremy-Gaillard: "subdivision **'s**"

src/Core/DefaultGeometryLayers.js
```diff
@@ -124,7 +138,21 @@ export function createGlobe(id, options) {
     return wgs84TileLayer;
 }
 
-
+/**
+ * Creates a planar geometry, which can be used to draw other layers (color, elevation, ...)
```
Jeremy-Gaillard: "Creates a planar geometry, **on** which other layers (color, elevation, ...) **can be drawn**"

src/Core/DefaultGeometryLayers.js
```diff
@@ -153,6 +181,19 @@ export function createPlane(id, extent, options) {
     return tileLayer;
 }
 
+/**
+ * Creates a 3d-tiles layer to display a tileset.
```
Jeremy-Gaillard: "Creates a layer that can display a 3D Tiles tileset"

src/Core/DefaultGeometryLayers.js
```diff
@@ -153,6 +181,19 @@ export function createPlane(id, extent, options) {
...
+ *
+ * @function
+ * @param {string} id - identifer (must be unique amongst all layers id)
+ * @param {Object} options
+ * @param {URL} options.url - url to the tileset
```
Jeremy-Gaillard: "url **of** the tileset"

src/Core/DefaultGeometryLayers.js
```diff
@@ -168,14 +209,28 @@ export function create3dTiles(id, options) {
...
 }
 
+/**
+ * Creates a geometry layer to display a pointcloud.
+ * The pointcloud data must have been prepared using PotreeConverter.
+ *
+ * @function
+ * @param {string} id - identifer (must be unique amongst all layers id)
+ * @param {Object} options
+ * @param {string} options.url URL to the folder containing options.file
```
Jeremy-Gaillard: "URL **of**"

Note: "URL to" may be correct, but seems awkward.

Note 2: we should choose between "url" and "URL". "URL" seems to be the correct form.

Jeremy-Gaillard (commented):
Nice work on the refactoring. I made a few comments on the wording of the documentation.

src/Core/Layer/Layer.js
peppsac: They're not documented yet (not in master nor in this PR)

peppsac:
Thanks @Jeremy-Gaillard for the review. I merged all your comments.

I pushed another commit to remove the need to specify the `type` of a layer when using the `addXXXLayer` function.

gcebelieu:
> Let us the time to verify, wait for @gcebelieu answer to @autra question before merge!

Thank's, I'm fine with @elias75015 review

src/Core/Scheduler/Providers/TileProvider.js
peppsac: Done see: https://github.com/iTowns/itowns/pull/546/files#diff-61e012c3e00bdc510be33d9eac684b2aR193

peppsac:
Branch updated.

examples/cubic_planar.js
gchoqueux: Before there  was already `addImageryLayer`, `addElevationLayer` and `addFeatureLayer`.
What was the motivation to change in `addLayer` and what's new to change the name again.
Because these changes impact the work of users

src/Core/Prefab/GlobeView.js
gchoqueux: so, `baseLayer` is necessary

gchoqueux: so, `baseLayer` is necessary?

peppsac: Yes it is.
It's a shortcut to the "default" geometry layer (ie : the globe when using GlobeView).

examples/cubic_planar.js
peppsac: > Before there was already ...

Possibly yes.

> What was the motivation to change in addLayer

The goal was to make itowns different from a globe renderer.
And at the time, adding a generic `addLayer` look like a good idea.

>  Because these changes impact the work of users

Sure. But cumbersome APIs "impact the work of users" as well.

Since we introduced `addLayer` we built more apps using iTowns and in the process realized that the API could be improved, which is the goal of this PR.

src/Core/Layer/Layer.js
```diff
@@ -175,4 +182,74 @@ const ImageryLayers = {
...
 
+/**
+ * Interface implemented by {@link GeometryLayer} instances capable of displaying color layers
+ * @interface
+ */
+/**
+ * Add a color layer to display on this geometry
+ *
+ * @function
+ * @name CanDisplayColorLayer#addColorLayer
```
gchoqueux, line 193: what is `CanDisplayColorLayer`?
where is documentation?

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,110 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
         throw new Error(`Cannot init tiled layer without schemeTile for layer ${layer.id}`);
     }
 
+    layer.addColorLayer = (colorLayer, addToView = true) => {
```
gchoqueux, line 37: I think this isn't the best place to overload the tileLayer (with `addColorLayer`....) the layer here because `TileProvider` provides tiles and pre-process tiles.
We could imagine a separate file that would be used in `createGlobe` and `createPlanar`

gchoqueux (changes requested):

examples/cubic_planar.js
gchoqueux: Ok, technically is there a reason that requires to rename or move function?

src/Core/Layer/Layer.js
peppsac: `CanDisplayColorLayer` *is* the documentation (see jsdoc virtual comments http://usejsdoc.org/tags-interface.html#virtual-comments)

src/Core/Scheduler/Providers/TileProvider.js
peppsac: That's what I thought initially, but actually adding `addColorLayer` and others functions in `TileProvider` makes sense: these features works when using TileMesh + LayeredMaterial which is exactly what `TileProvider` builds.

examples/cubic_planar.js
peppsac: We could keep the generic `addLayer` name but IMHO it improves the API to have names that match the type of layers you can add.

src/Core/Scheduler/Providers/TileProvider.js
gchoqueux: `tileProvider` provides the data (image, elevation...) container and not the data.
A movie screen provider doesn't handle movies.
`tileProvider` should not handle anything other than tiles

src/Core/Layer/Layer.js
gchoqueux: yes thanks,
but `{@link CanDisplayColorLayer}` links on nothing

gchoqueux:
@elias75015  you have time to take a look at the doc?

examples/cubic_planar.js
gchoqueux: What annoys me going back and forth in the API, which impacts the users.
Especially since previously the names of the API were clear

src/Core/Scheduler/Providers/TileProvider.js
peppsac: I think you got it backwards.
`TileProvider` doesn't provide the color layer support.
TileProvider simply knows that layers using the `tile` protocol support color/elevation/feature layer so it adds the `addXXXLayer` method, that's it.

src/Core/Layer/Layer.js
peppsac: Oh right. This is fixed now.

src/Core/Scheduler/Providers/TileProvider.js
gchoqueux: `TileProvider` should only provide tiles and provide the first tiles during the preprocess.
I don't see why a tile provider add functions to add another layer.

I find it too specific because why we could not add a feature / color layer to anything other than tiles

peppsac: Let's imagine we add a new feature to itowns: colorize a point cloud based on a WMS service.

To express this new feature, the PointCloudProvider would add a addColorLayer to a pointcloud layer.

This way, layers have specific methods for all addXXXLayer they support + specific documentation coming from jsdoc virtual comment.
And it seems clearer as a user, than a generic `addLayer` interface that will alwyas be available but with a chaning behavior depending on the underlying layer.

gchoqueux: The code considers that the tile has a `layeredMaterial`. The tile will not have a layered material all the time and the tile should not be constrained to a single material. This code is more specific to the material than to the tile. The only role of `tileProvider` is to provide mesh.

peppsac: > The code considers that the tile has a layeredMaterial

No this code assumes that tiles produced by TileProvider supports color layers.

> The tile will not have a layered material all the time and the tile should not be constrained to a single material. 

This has been the case since the beginning of iTowns and I'm not aware of any plans to remove the mutual dependency.
So unless I've missed something I think we can reasonnably assume that TileProvider / TileMesh and LayeredMaterial are heavily linked.

> The only role of tileProvider is to provide mesh.

Did you read the code I added in TileProvider? This doesn't change TileProvider role, it's simply about extending layers using the `tile` protocol to inform the user they can add color (and other) layers to it.
Would it make you happy if I moved the `addXXXLayer`/`removeXXXLayer` code to `DefaultGeometryLayers.js`?

gchoqueux:
I think it's reasonable to add test, isn't it?

peppsac:
> I think it's reasonable to add test, isn't it?

I don't understand sorry :/

gchoqueux:
in addition to the previous remarks,
It must be ensured that the functions of adding layers, removing layers, changing layer orders ... are not buggy. So we need functional / unit tests

peppsac:
> So we need functional / unit tests

Hold on. Before starting a new discussion I'd like to settle the current one, aka: https://github.com/iTowns/itowns/pull/546#discussion_r152269900

elias75015:
> @elias75015 you have time to take a look at the doc?

What I think about the interfaces doc :  
- the options specific to a protocol should be exposed in the generated documentation

> Each protocol has a set of specific options that are documented in their respective providers.

- it could be interesting to add a little snippet to show how to add/remove a layer (as it is done in the doc for other methods of the exposed API)

src/Core/Scheduler/Providers/TileProvider.js
gchoqueux: it's better to place in `DefaultGeometryLayers.js`, and in addition to factoring common things in `createGlobe` and `createPlanar`

examples/wfs.js
```diff
@@ -22,10 +22,10 @@ viewerDiv = document.getElementById('viewerDiv');
...
 view = new itowns.PlanarView(viewerDiv, extent, { renderer: renderer });
-view.tileLayer.disableSkirt = true;
+view.baseLayer.disableSkirt = true;
 
 // Add an WMS imagery layer (see WMS_Provider* for valid options)
-view.addLayer({
+view.baseLayer.addColorLayer({
     url: 'https://download.data.grandlyon.com/wms/grandlyon',
     networkOptions: { crossOrigin: 'anonymous' },
     type: 'color',
```
zarov, line 31: You forgot to remove this one ;)

src/Core/Prefab/PlanarView.js
```diff
@@ -144,28 +143,45 @@ function PlanarView(viewerDiv, extent, options = {}) {
...
+ * @deprecated
+ * @return {Promise} see View.addLayer
+ */
+PlanarView.prototype.addLayer = function addLayer(layer) {
+    if (layer instanceof GeometryLayer) {
+        return View.prototype.addLayer.call(this, layer);
+    }
+    if (!this._warnAddLayerDeprecated) {
+        // eslint-disable-next-line no-console
+        console.warn('globeView.addLayer(colorLayer) has been deprecated.\n' +
```
zarov, line 174: `globeView` -> `planarView`

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -32,6 +34,110 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+            }
+            return true;
+        } else {
+            // eslint-disable-next-line no-console
+            console.error(`${featureLayerOrId} isn't a feature layer`);
+            return false;
+        }
+    };
+
+    layer.removeLayer = (layerOrId) => {
```
zarov, line 126: I don't understand having this method without `addLayer`. Why not have a similar `addLayer` with shortcuts to `addXXXXLayer` like below ? Or then you need to deprecate this method imho.

zarov (commented):
I agree for moving to `addXXXXLayer`, but there is one thing I don't understand.
> `GlobeView.wgs84TileLayer` and` PlanarView.tileLayer` become `XxxxView.baseLayer`

is quite a good thing, good uniformization, but why then do we need to go through the `XxxxView.baseLayer.addXXXXLayer` instead of `XxxxView.addXXXXLayer`, like we used to have before ?

peppsac:
> XxxxView.baseLayer.addXXXXLayer instead of XxxxView.addXXXXLayer, like we used to have before 

I initially removed the `View.addLayer` shortcut because it made it complicated to add a layer to the view when you didn't want to attach it to `baseLayer` (you had to use `itowns.View.prototype.addLayer.call(...)`)

Now that the `addLayer` is gone and replaced by specialized `addXXXLayer` we could reintroduce the shortcuts for all `baseLayer.addXXXLayer` methods.

Is this what you had in mind?

zarov:
> (you had to use itowns.View.prototype.addLayer.call(...))

Oh I see, it is better like this then. Yes, that's what I got in mind.

peppsac:
Closing: I'll propose a new PR later with an improved API, inspired from MapBox style spec (splitting datasource, style and displayable objects)

PR #547, opened 2017-10-24
fix (protocol) : stop using layer's options tileMatrixSet in WMS provider

gchoqueux:
WMS layer haven't tileMatrixSet, this PR removes this inconsistency

problem when you use PR #513

### DIFF
--- a/src/Core/Scheduler/Providers/WMS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMS_Provider.js
@@ -32,13 +32,7 @@ WMS_Provider.prototype.url = function url(bbox, layer) {
 };
 
 WMS_Provider.prototype.tileTextureCount = function tileTextureCount(tile, layer) {
-    if (tile.extent.crs() == layer.projection) {
-        return 1;
-    } else {
-        const tileMatrixSet = layer.options.tileMatrixSet;
-        OGCWebServiceHelper.computeTileMatrixSetCoordinates(tile, tileMatrixSet);
-        return tile.getCoordsForLayer(layer).length;
-    }
+    return tile.extent.crs() == layer.projection ? 1 : tile.getCoordsForLayer(layer).length;
 };
 
 WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer) {


peppsac (commented):
Commit message should be `fix (protocols)` (not `protocol`).

Also `fix the undef layer...` is unclear.
Here's another wording, which seems clearer to me:
```
fix (protocols): stop using tileMatrixSet for WMS protocol

tileMatrixSet are only valid for WMTS protocol so the WMS_Provider can't use them (because they will probably be undefined)
```

gchoqueux:
ok, I change message
`tileMatrixSet`  is valid for `wms`, it's projection that determines its `tileMatrixSet`

peppsac:
`TileMatrixSet` is missing from the WMS spec (http://www.opengeospatial.org/standards/wms) so I don't see what link do you see between tilematrixset and wms?

gchoqueux:
in itowns, the `wms` uses the wmts's tile matrix set.

peppsac:
Looks more like a bug than a feature to me; mainly because WMTS / TileMatrixSet only makes sense for the globe.

gchoqueux:
i don't understand

peppsac:
Don't you think that:

> in itowns, the wms uses the wmts's tile matrix set.

And:

> WMS layer haven't tileMatrixSet, this PR removes this inconsistency

are a bit contradictory?

And can you explain to me how WMTS TileMatrixSet are going to help when using WMS on a planar geometry (for instance: drawing a `EPSG:3857` layer on a `EPSG:3946` geometry)?

Mkonini:
@gchoqueux , @peppsac 
To facilitate understanding maybe it would be better to open an issue to discuss about, and btw see if the proposed fix is worthly to be merged.

zarov:
@gchoqueux could you explain where the WMS uses the `tileMatrixSet` of the WMTS in iTowns ? So we can better understand it, as this PR should be merged by now.

gchoqueux:
In [wms_provider](https://github.com/iTowns/itowns/blob/a616629d9001a6a988b40598cd4575c60372fbd9/src/Core/Scheduler/Providers/WMS_Provider.js#L135)  the wms texture is branded by tile's coord. it corresponds to the tile matrix set. this allows us to know if the texture is accurate enough for the tile. 

On the other hand this PR stops to use tileMatrix set in layer's option.

peppsac (commented):
Rebasing / force-pushing the PR makes the conversation useless because the matching code changes aren't accessible now.

Anyway, this PR looks good now.

zarov (approved):

zarov:
Thanks @gchoqueux and @peppsac

PR #548, opened 2017-10-24
feature (controls) : add collision camera/tiles's obb to globeControls

gchoqueux:
## Description
add collision between camera's bounding and tiles's oriented bounding box. Depending on the distance of the camera with obbs, we add a slowdown or constrain to the movement. this constraint or deceleration is suitable for two types of movement MOVE_GLOBE and ORBIT. This constraint or deceleration. It's inversely proportional to the camera / obb distance

## Motivation and Context
the actual collision idoesn't work properly because it uses only the line / tile intersection.
In addition it adds a translation on normal surface without to take into account the movements of the globe control

### DIFF
--- a/src/Renderer/ThreeExtended/OBB.js
+++ b/src/Renderer/ThreeExtended/OBB.js
@@ -98,6 +98,25 @@ OBB.prototype._cPointsWorld = function _cPointsWorld(points) {
     return points;
 };
 
+/**
+ * Determines if the sphere is above the XY space of the box
+ *
+ * @param      {Sphere}   sphere  The sphere
+ * @return     {boolean}  True if sphere is above the XY space of the box, False otherwise.
+ */
+OBB.prototype.isSphereAboveXYBox = function isSphereAboveXYBox(sphere) {
+    const localSpherePosition = this.worldToLocal(sphere.position);
+    // get obb closest point to sphere center by clamping
+    const x = Math.max(this.box3D.min.x, Math.min(localSpherePosition.x, this.box3D.max.x));
+    const y = Math.max(this.box3D.min.y, Math.min(localSpherePosition.y, this.box3D.max.y));
+
+    // this is the same as isPointInsideSphere.position
+    const distance = Math.sqrt((x - localSpherePosition.x) * (x - localSpherePosition.x) +
+                           (y - localSpherePosition.y) * (y - localSpherePosition.y));
+
+    return distance < sphere.radius;
+};
+
 // Allocate these variables once and for all
 const tmp = {
     epsg4978: new Coordinates('EPSG:4978', 0, 0),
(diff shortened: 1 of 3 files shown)


Mkonini:
This PR is an improvement of PR #465 it allows to get rid of mostly remaining bugs, it would be good to have it.  @nosy-b , @peppsac , @autra : wolud you please review it ?

peppsac:
Commit title doesn't match #518.

And I believe there's a typo / missing words in the commit message ('This constraint or deceleration' ?)

Regarding the content of the PR: GlobeControls is nicely reaching 2000 lines of stateful code so I can't safely review this PR in a couple of hours (and I though that #535 was the last quick-fix-before-reactoring-PR?).

nosy-b:
It does improve collision handling but only in globe view not planar. 
Plus, I agree it also keeps adding code in a deprecated Class.
@gchoqueux Could you place `getDistanceSphereObbUp` and its friends out of GlobeControls for a more generic and future usage?

gchoqueux:
updated

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -98,6 +98,18 @@ OBB.prototype._cPointsWorld = function _cPointsWorld(points) {
     return points;
 };
 
+OBB.prototype.sphereIsAboveXYBox = function sphereIsAboveXYBox(sphere) {
```
peppsac: Functions name start with a verb (`isXXX`).

Can you add the jsdoc too please?

gchoqueux: done

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -98,6 +98,24 @@ OBB.prototype._cPointsWorld = function _cPointsWorld(points) {
...
 
+/**
+ * Determines if the sphere is above the XY space of the box
+ *
+ * @param      {Sphere}   sphere  The sphere
+ * @return     {boolean}  True if sphere is above the XY space of the box, False otherwise.
+ */
+OBB.prototype.isSphereAboveXYBox = function isSphereAboveXYBox(sphere) {
+    // get obb closest point to sphere center by clamping
+    var x = Math.max(this.box3D.min.x, Math.min(sphere.position.x, this.box3D.max.x));
```
peppsac: You don't need to use `this.matrixWorld` here?

peppsac: (and please use const where possible)

peppsac: And please could you add a unit test to cover this feature?

peppsac (changes requested):

src/Renderer/ThreeExtended/OBB.js
gchoqueux: yes, it's better, I move transform local position here

> And please could you add a unit test to cover this feature?

I don't have time for the moment

gchoqueux:
updated

src/Renderer/ThreeExtended/OBB.js
gchoqueux: I add the unit test

peppsac (approved):
Thanks for the test.

I quickly tested the controls and it seems to work.

PR #550, opened 2017-10-25
feature (core): support fog for none logarithmic depth buffer 

gchoqueux:
add support fog for none logarithmic depth buffer

### DIFF
--- a/src/Renderer/Shader/TileFS.glsl
+++ b/src/Renderer/Shader/TileFS.glsl
@@ -76,10 +76,11 @@ void main() {
 
         #if defined(USE_LOGDEPTHBUF) && defined(USE_LOGDEPTHBUF_EXT)
             float depth = gl_FragDepthEXT / gl_FragCoord.w;
-            float fogIntensity = 1.0/(exp(depth/distanceFog));
         #else
-            float fogIntensity = 1.0;
+            float depth = gl_FragCoord.z / gl_FragCoord.w;
         #endif
+
+        float fogIntensity = 1.0/(exp(depth/distanceFog));
 
         vec4 diffuseColor = CWhite;
         bool validTexture = false;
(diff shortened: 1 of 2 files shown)


src/Core/TileMesh.js
```diff
@@ -85,7 +85,9 @@ TileMesh.prototype.changeState = function changeState(state) {
 };
 
 TileMesh.prototype.setFog = function setFog(fog) {
-    this.material.setFogDistance(fog);
+    if (fog != undefined) {
```
peppsac: In which situation would `setFog(...)` would be called with `fog = undefined` (ie: are you sure the bug isn't elsewhere?)

peppsac (changes requested):
Wording of your git commits do not follow #518

src/Renderer/Shader/TileFS.glsl
```diff
@@ -78,7 +78,8 @@ void main() {
             float depth = gl_FragDepthEXT / gl_FragCoord.w;
             float fogIntensity = 1.0/(exp(depth/distanceFog));
         #else
-            float fogIntensity = 1.0;
+            float depth = gl_FragCoord.z / gl_FragCoord.w;
+            float fogIntensity = 1.0/(exp(depth/distanceFog));
```
autra: this line could be moved out of the #if/#else

autra (dismissed):
LGTM, just one nits if you care to correct.

For general information, I tested with Epiphany on Ubuntu 16.04: it doesn't support log depth buffer.

gchoqueux:
There are several browsers don't support depth buffer

src/Core/TileMesh.js
gchoqueux: Because the fogDistance isn't always defined, we can give it a default value. 
it's reason break on planar View

peppsac: Could you instead fix the caller to not call `setFog` with an undefined value please?

gchoqueux: done

peppsac (approved):
Thanks.

PR #551, opened 2017-10-26
feat (core): support texture with different x/y scales

peppsac:
Until now we assumed that textures proportion matched geometry proportion.
This commits breaks this assumption by using a 2 floats instead of 1 for
specifying the region used.


src/Core/Geographic/Extent.js
```diff
@@ -194,7 +194,12 @@ Extent.prototype.offsetToParent = function offsetToParent(other) {
     const scale =
```
gchoqueux: you could standardize variable names in the caller `scaleX` instead of `scale`

gchoqueux (approved):

src/Core/Geographic/Extent.js
peppsac: Fixed.

PR #552, opened 2017-10-26
fix (geometry): compute vertices/triangles count correctly

peppsac:
The vertices and triangles counts must be lower if the skirt is disabled.

### DIFF
--- a/src/Core/TileGeometry.js
+++ b/src/Core/TileGeometry.js
@@ -73,8 +73,8 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
     // segments count :
     // Tile : (nSeg + 1) * (nSeg + 1)
     // Skirt : 8 * (nSeg - 1)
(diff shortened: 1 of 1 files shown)


src/Core/TileGeometry.js
```diff
@@ -86,7 +86,7 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
         outBuffers.index = cachedBuffers.index;
         outBuffers.uv.wgs84 = cachedBuffers.uv.wgs84;
     } else {
-        scratchBuffers.index = new Uint32Array(triangles * 3 * 2);
```
gchoqueux, line 89: there was too much index in all case?

gchoqueux (approved):

src/Core/TileGeometry.js
peppsac: No I move the `* 2` in the triangle count computation.
The indices count was correct but is better expressed as `num_triangles * 3`

gchoqueux: ok

Issue #545, opened 2017-10-20
Rework GlobeControls


nosy-b:
The GlobeControls class has evolved a lot since the creation of itowns2. It is of great help but it is getting very messy with 2000 lines of delicious spaghetti.

A complete refactorization is necessary.

autra:
Yes!

A basic idea: the new `GlobeControls` should be a FrameRequester. We use the event listeners to only set states (panning, zooming, etc...) and relevant variables (mouseDown positions and stuff). We do the heavy work in the `update` function of this GlobeControls. See other controls for inspiration.

It might be necessary to add other anchors for FrameRequester in MainLoop (being called before rendering, after rendering, before update, after update etc...) to be able to do everything we want (without having to setup timeout, promises on command-empty, or dummy frame requesters).

Last but not least, I think we can work incrementally by having a GlobeControls2.js, and replacing the former GlobeControls once we reach feature parity, instead of trying to rewrite everything in one PR.

gchoqueux:
Since #795, Globe Controls has been greatly refactoring and debugging. 
It still remains 
- FrameRequester
- Collision refacto

PR #553, opened 2017-10-26
fix (controls): positioning error during camera movements

gchoqueux:
There was positioning error during camera movements with `globeControls`

**First commit**
In `setCameraTargetPosition` without animation, the final position was wrong because the transformation was simply not applied.
Can be tested through the function `setCameraTargetGeoPosition`

**second commit**
Other positioning error come from with double click.
Because it's in conflict with the simple click that calls the `setCameraTargetPosition` too.
For double click event:
  * with animation enabled problem is fixed
  * with animation disabled is forced to enable

### DIFF
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -1105,6 +1105,12 @@ function GlobeControls(view, target, radius, options = {}) {
     var ondblclick = function ondblclick(event) {
         if (this.enabled === false) return;
 
+        if (!this.isAnimationEnabled()) {
+             // eslint-disable-next-line no-console
+            console.warn('double click without animation is disabled, waiting fix in future refactoring');
+            return;
+        }
+
         // Double click throws move camera's target with animation
         if (!currentKey) {
             ptScreenClick.x = event.clientX - event.target.offsetLeft;
@@ -1114,7 +1120,7 @@ function GlobeControls(view, target, radius, options = {}) {
 
             if (point) {
                 animatedScale = 0.6;
-                this.setCameraTargetPosition(point, this.isAnimationEnabled());
+                this.setCameraTargetPosition(point);
             }
         }
     };
@@ -1619,8 +1625,11 @@ GlobeControls.prototype.setCameraTargetPosition = function setCameraTargetPositi
         return player.play(animationZoomCenter).then(() => {
             this.resetControls();
             this.waitSceneLoaded().then(() => {
-                this.updateCameraTransformation();
-                ctrl.targetGeoPosition = null;
+                animatedScale = 0;
+                if (player.isStopped()) {
+                    this.updateCameraTransformation();
+                    ctrl.targetGeoPosition = null;
+                }
             });
         });
     } else {
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -1428,6 +1428,12 @@ function GlobeControls(view, target, radius, options = {}) {
     window.addEventListener('keydown', onKeyDown.bind(this), false);
     window.addEventListener('keyup', onKeyUp.bind(this), false);
 
+    // Reset key/mouse when window loose focus
+    window.addEventListener('blur', () => {
+        onKeyUp.bind(this)();
+        onMouseUp.bind(this)();
+    });
+
     // Initialisation Globe Target and movingGlobeTarget
     setCameraTargetObjectPosition(target);
     movingCameraTargetOnGlobe.copy(target);
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -1626,6 +1626,7 @@ GlobeControls.prototype.setCameraTargetPosition = function setCameraTargetPositi
     } else {
         ctrl.progress = 1.0;
         quatGlobe.setFromUnitVectors(vFrom, vTo);
+        this.updateCameraTransformation(this.states.MOVE_GLOBE, false);
         this._view.wgs84TileLayer.postUpdate = () => {
             clampToGround(ctrl);
             this.updateCameraTransformation(this.states.MOVE_GLOBE, false);


autra:
> In PR #545 

#545 is an issue, you meant #535 maybe ?

gchoqueux:
yes, thanks #535

autra (changes requested):
Commit message should preferably explain why the change (and not what, most of the time).

So what is the actual bug? Can you give me reproduction steps, expected and actual results please? I tried to double-click everywhere with animation disabled (this leads to this new code being executed), but couldn't see the difference before and after your patch: this double-click is still completely broken in both cases...

gchoqueux:
@autra 
Thanks, for the bug with double-click.
I updated PR's message and the commit

gchoqueux:
@autra PR is updated

nosy-b:
Weird behaviour when leaving the focus (opening a terminal) then going back to itowns, the click is no more listened, only with CTRL - CLICK I can see that the app is not freezed.
(I specified enableAnimation to false, it still animates translation also...)

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1114,7 +1114,7 @@ function GlobeControls(view, target, radius, options = {}) {
 
             if (point) {
                 animatedScale = 0.6;
-                this.setCameraTargetPosition(point, this.isAnimationEnabled());
+                this.setCameraTargetPosition(point, true);
```
autra: This was for testing right? If so should be remoed

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1619,13 +1619,17 @@ GlobeControls.prototype.setCameraTargetPosition = function setCameraTargetPositi
...
+                if (player.isStopped()) {
+                    this.updateCameraTransformation();
+                    ctrl.targetGeoPosition = null;
+                }
             });
         });
     } else {
         ctrl.progress = 1.0;
         quatGlobe.setFromUnitVectors(vFrom, vTo);
+        this.updateCameraTransformation(this.states.MOVE_GLOBE, false);
```
autra, line 1644: MOVE_GLOBE is the default state?

autra (changes requested):
One leftover+a question  and a check on  @nosy-b's comments.

After this is done, I'll do one last test and I guess we can merge that. I'm not sure to really understand what's going on here though...

autra:
@nosy-b just tested on master and reproduced the 2 bugs you mentioned. They are not caused by this PR.

gchoqueux:
@nosy-b fix focus problem

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: Default state is `NONE`, in `setCameraTargetPosition uses `MOVE_GLOBE` to move target's position`

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: I added comment, there's conflict simple/double click without animation. I'll fix it in the refacto.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1422,6 +1425,12 @@ function GlobeControls(view, target, radius, options = {}) {
     window.addEventListener('keydown', onKeyDown.bind(this), false);
     window.addEventListener('keyup', onKeyUp.bind(this), false);
 
+    // Reset key/mouse when window loose focus
+    window.onblur = () => {
```
autra: `this.domElement.addEventListener('blur',...)`: this will fix it it another element takes focus somehow too. Also, maybe the webpage in which itowns is integrated uses also this blur event. Here you're overriding whatever might already be in `window.onblur`.

Also, now that the review process has begun, I'd prefer you open new PR for new fixes that come on the way.

src/Renderer/ThreeExtended/GlobeControls.js
autra: Ok, not very intuitive (we might want another name for the refacto), but fair enough.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1114,7 +1114,10 @@ function GlobeControls(view, target, radius, options = {}) {
 
             if (point) {
                 animatedScale = 0.6;
-                this.setCameraTargetPosition(point, this.isAnimationEnabled());
+                // FIXME: single click and double click are in conflict with no animation
+                // with this.setCameraTargetPosition
+                // force animation at true to tempory fix the problem
+                this.setCameraTargetPosition(point, true);
```
autra: If it works with animation, but not without, it looks like a race condition. Anyhow I'm skeptical of this fix. Adding an animation when animations are disabled to fix a bug is... not exactly the most straightforward thing I can think of. I'm sure there's a way to make it work with animation disabled. 

What is the conflict exactly?  Can you describe the bug in more details?

autra (changes requested):

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: I don't want to spend more time on this version,
I will correct that in the refacto

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: I change in `window.addEventListener('blur'`...
`this.domElement.addEventListener` doesn't work

autra: Ok works for me. (it won't fix it if another element takes focus in the page, but that's not the goal of this PR)

src/Renderer/ThreeExtended/GlobeControls.js
autra: Then if you don't want to spend more time here, let's deactivate the double-click action when enableAnimation is false. We can't really merge code that lies! (doing an animation when enableAnimation is false)

If this does not suit you, then I'll try to find a solution when I have time, but you'll have to wait for me. If you want to speed up this process, you can answer my questions 

> What is the conflict exactly? Can you describe the bug in more details?

If you don't want to wait, you'll have to propose an alternative solution :-)

autra: Anyway, maybe it's better to put the problematic commit into its own PR. This way, I'll merge this one now.

gchoqueux: I disable without animation and I add message warning

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1114,7 +1120,10 @@ function GlobeControls(view, target, radius, options = {}) {
 
             if (point) {
                 animatedScale = 0.6;
-                this.setCameraTargetPosition(point, this.isAnimationEnabled());
+                // FIXME: single click and double click are in conflict with no animation
+                // with this.setCameraTargetPosition
+                // force animation at true to tempory fix the problem
```
autra: This comment is now outdated. Let's remove it.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1105,6 +1105,12 @@ function GlobeControls(view, target, radius, options = {}) {
     var ondblclick = function ondblclick(event) {
         if (this.enabled === false) return;
 
+        if (!this.isAnimationEnabled()) {
+             // eslint-disable-next-line no-console
+            console.warn('double click without animation is disabled, waiting fix in new factorization');
```
autra: `waiting fix in future refactoring`?

autra (changes requested):
The new diff is on the wrong commit (is in last commit, should be on second). Also, second commit message needs to be updated. After this review, it'll be ready to merge.

gchoqueux:
@autra done

gchoqueux:
@autra we merge it?

autra:
@gchoqueux I think you forgot to adress

> The new diff is on the wrong commit (is in last commit, should be on second). Also, second commit message needs to be updated. After this review, it'll be ready to merge.

So I took the liberty to amend your commits. Please double-check if it's ok for you, then we can merge.

autra (dismissed):

autra:
(I have a backup of the old commits if needed)

autra:
Ok the diff are alright, but the commit message needs some amending. You can take my proposals or not, but at least on second commit message:

>   * with animation disabled is forced to enable

should be changed, as it's not the case anymore. Thanks.

autra (approved):
Here we go!

PR #554, opened 2017-10-26
Fix color layer processing

peppsac:
LayerUpdateState were under-utilized and this lead to many useless computation.
For instance 'tileInsideLimit' was called each time we tried to update a tile, while this could be done once.

The 1st commit of this PR shuffles the code to move all the 'one-time' init at the top of the function.

(the 2nd commit is a followup of a616629d9001a6a988b40598cd4575c60372fbd9 where I forgot 1 user of pitch values)


autra (approved):
LGTM.

PR #555, opened 2017-10-27
feat (protocols): improve StaticProvider.tileInsideLimit performance

peppsac:
When a StaticProvider has many images to choose from it becomes quite costly
to browse all of them each time we need to know if a tile is covered by the
layer.
This commit adds a shortcut: if the node already has a LayerUpdateState object
for this layer, this means we already determined the node is indeed covered by
the layer.


autra (approved):
Yep, this'll be better :-)

PR #556, opened 2017-10-27
Split OBB.extentToOBB function

peppsac:
This PR splits OBB.extentToOBB function in 2, moving the real computation
from xyz coords to OBB.cardinalsXYZToOBB.
This will allow to reuse the same code in a different context than the  EPSG:4326 globe.

It also add OBB unit testing code - with some of them being disabled because not all geometry vertices are inside the computed bounding box (see #87). But the tests were used to verify that the function splitting didn't introduce more errors.

### DIFF
--- a/src/Renderer/ThreeExtended/OBB.js
+++ b/src/Renderer/ThreeExtended/OBB.js
@@ -126,8 +126,8 @@ const tmp = {
     minV: new THREE.Vector3(),
     translate: new THREE.Vector3(),
     cardinal3D: new THREE.Vector3(),
-    planeZ: new THREE.Quaternion(),
-    qRotY: new THREE.Quaternion(),
+    transformNormalToZ: new THREE.Quaternion(),
+    alignTileOnWorldXY: new THREE.Quaternion(),
     tangentPlaneAtOrigin: new THREE.Plane(),
     zUp: new THREE.Vector3(0, 0, 1),
 };
@@ -144,8 +144,6 @@ OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
 
     // Calcule the center world position with the extent.
     extent.center(tmp.cardinals[8]);
-    const centerWorld = tmp.cardinals[8].as('EPSG:4978', tmp.epsg4978).xyz();
-    tmp.normal.copy(centerWorld).normalize();
 
     const bboxDimension = extent.dimensions(UNIT.RADIAN);
     const phiStart = extent.west(UNIT.RADIAN);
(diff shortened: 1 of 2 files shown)


src/Renderer/ThreeExtended/OBB.js
```diff
@@ -175,41 +173,55 @@ OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
     tmp.cardinals[7]._values[0] = phiStart;
     tmp.cardinals[7]._values[1] = thetaStart + bboxDimension.y * 0.5;
 
-    var cardin3DPlane = [];
+    const cardinalsXYZ = [];
+    const centersLongitude = tmp.cardinals[8].longitude(UNIT.RADIAN);
+    for (var i = 0; i < tmp.cardinals.length; i++) {
```
gchoqueux: you could use `for (const ... of ...)`

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -175,41 +173,55 @@ OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
...
+        cardinalsXYZ.push(tmp.cardinals[i].as('EPSG:4978').xyz());
+    }
+
+    return this.cardinalsXYZToOBB(cardinalsXYZ, centersLongitude, true, minHeight, maxHeight);
+};
 
+OBB.cardinalsXYZToOBB = function cardinalsXYZToOBB(cardinals, centerPhi, isEllipsoid, minHeight = 0, maxHeight = 0) {
```
gchoqueux: you could add some documentation

gchoqueux: `centerPhi`, phi isn't too suitable for Cartesian parameters

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -175,41 +173,55 @@ OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
...
-    var halfMaxHeight = 0;
+
+    let halfMaxHeight = 0;
+    tmp.normal.copy(cardinals[8]).normalize();
     tmp.tangentPlaneAtOrigin.set(tmp.normal, 0);
 
     tmp.planeZ.setFromUnitVectors(tmp.normal, tmp.zUp);
     tmp.qRotY.setFromAxisAngle(
-        new THREE.Vector3(0, 0, 1), -tmp.cardinals[8].longitude(UNIT.RADIAN));
+        new THREE.Vector3(0, 0, 1), -centerPhi);
```
gchoqueux: Why it's necessary to have `centerPhi` in parameter, we could assume that it is in the `cardinals[8]`

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -175,41 +173,55 @@ OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
...
     tmp.qRotY.multiply(tmp.planeZ);
 
-    for (var i = 0; i < tmp.cardinals.length; i++) {
-        tmp.cardinals[i].as('EPSG:4978', tmp.epsg4978).xyz(tmp.cardinal3D);
-        cardin3DPlane.push(tmp.tangentPlaneAtOrigin.projectPoint(tmp.cardinal3D));
-        const d = cardin3DPlane[i].distanceTo(tmp.cardinal3D.sub(centerWorld));
+    let point5InPlaneX;
+    for (let i = 0; i < cardinals.length; i++) {
```
gchoqueux, line 209: you could use `for (const ... of ...)` ?

test/obb_unit_test.js
```diff
@@ -20,3 +25,53 @@ describe('OBB', function () {
...
+    const vec = new THREE.Vector3();
+    for (let i = 0; i < geom.attributes.position.count; i++) {
+        vec.fromArray(geom.attributes.position.array, 3 * i);
+
+        vec.applyMatrix4(inverse);
+        if (!geom.OBB.box3D.containsPoint(vec)) {
+            failing++;
+        }
+    }
+    assert.equal(geom.attributes.position.count - failing, geom.attributes.position.count, 'All points should be inside OBB');
```
gchoqueux, line 51: why not write ?
```js
assert.equal(failing, 0, 'All points should be inside OBB');
```

test/obb_unit_test.js
```diff
@@ -20,3 +25,53 @@ describe('OBB', function () {
...
+}
+
+describe('Ellipsoid tiles OBB computation', function () {
+    const builder = new BuilderEllipsoidTile();
+
+    xit('IGNORED see #87 - should compute globe-level 0 OBB correctly', function () {
```
gchoqueux, line 57: I do a new PR to refacto tile and OBB for ellipsoid.
In particular to place the normal of tile's center on the Z axis and use the bounding box THREE.js

gchoqueux (changes requested):

src/Renderer/ThreeExtended/OBB.js
peppsac: Because `cardinals` are `xyz` coordinates, not angular.

test/obb_unit_test.js
peppsac: With my version, if the test is failing you 2 info: the number of failing vertices and the total number of vertices.

src/Renderer/ThreeExtended/OBB.js
peppsac: I use the `i` in the for loop so I can't use `for of`

peppsac:
Rebased and updated.

src/Renderer/ThreeExtended/OBB.js
it's rotation on z axis

gchoqueux: it's even a rotation around the normal of the tile
I will change to rotating `centerPhi`  to `rotationOnNormal`

peppsac:
@gchoqueux could you take a look at this PR please?

gchoqueux:
[there's a comment not taken into account](https://github.com/iTowns/itowns/pull/556#discussion_r149351246)

src/Renderer/ThreeExtended/OBB.js
peppsac: I changed for `centerLongitude` because that's what it is from the user of the function point of view.

gchoqueux: `centerLongitude` has less meaning in planar.
`qRotY` is incorrect because it's a rotation on Z.
this rotation is done on the normal of the tile, so I will change the names in this direction.
`centerPhi` to `rotationOnNormal` 
`qRotY` to `quaternionOnNormal`

peppsac: > centerLongitude has less meaning in planar.

Doesn't matter since this method is not meant to be used in planar mode.

>  qRotY is incorrect because it's a rotation on Z.

I'll fix this.

>  centerPhi to rotationOnNormal

No. This is not an internal variable but a function parameter name.
And `rotationOnNormal` doesn't make sense for the users of this function - while tile's longitude is quite easy to understand.

>  qRotY to quaternionOnNormal

I don't think repeating the type of the variable in the var name is useful.
I can change to `rotationOnNormal` if you want even if I don't think this name is useful.
I'd prefer something describing the intent: ie what does this rotation means?
For instance if the point of this rotation is to align x/y axes of the tile to the world's X/Y axes, then I'd name it "alignLocalXYAxesToWorldAxes".

gchoqueux: > `alignLocalXYAxesToWorldAxes`

the axes of the tiles are already aligned. Against this rotation on the normal allows to align the plane of symmetry of the tile on the Y axis.

may be `alignPlaneSymmetryToYAxis`

* ok for `centerLongitude`

peppsac: `align the plane of symmetry of the tile on the Y axis` is unclear. 
Do you meant "to align the axis of symetry of the tile on the Y axis?"
(with the axis of symetry being the line going through cardinals 1-5?)
If it's not what you meant, could you make a simple drawing?

gchoqueux: To align orthogonal symmetric plane of the tile's geometry on the y axis.
there isn't axis of symmetry  of the tile.

gchoqueux (approved):

gchoqueux:
thanks

PR #559, opened 2017-10-27
feat (controls): add PanoramaControls

peppsac:
Add a control mode tailored for panoramas.
Includes:
  - fov and rotation limits to prevent the user to see "outside" of the panorama
  - rotations takes the camera up vector into account. So it works as expected
even if camera up vector isn't (0, 0, 1)

### DIFF
--- a/src/Renderer/ThreeExtended/FirstPersonControls.js
+++ b/src/Renderer/ThreeExtended/FirstPersonControls.js
@@ -4,32 +4,83 @@ import * as THREE from 'three';
 // but including these controls in itowns allows use to integrate them tightly with itowns.
 // Especially the existing controls are expecting a continuous update loop while we have a pausable one (so our controls use .notifyChange when needed)
 
-function onPointerDown(event, pointerX, pointerY) {
+
+// Mouse movement handling
+function onDocumentMouseDown(event, pointerX, pointerY) {
     event.preventDefault();
-    this._isUserInteracting = true;
+    this._isMouseDown = true;
 
     this._onMouseDownMouseX = pointerX;
     this._onMouseDownMouseY = pointerY;
-    this._onMouseDownRotZ = this.camera.rotation.z;
-    this._onMouseDownRotX = this.camera.rotation.x;
+
+    this._stateOnMouseDown = this._state.snapshot();
+}
+
+function limitRotation(camera3D, rot, verticalFOV) {
+    // Limit vertical rotation (look up/down) to make sure the user cannot see
+    // outside of the cone defined by verticalFOV
+    const limit = THREE.Math.degToRad(verticalFOV - camera3D.fov) * 0.5;
+    return THREE.Math.clamp(rot, -limit, limit);
 }
 
 function onPointerMove(pointerX, pointerY) {
-    if (this._isUserInteracting === true) {
+    if (this._isMouseDown === true) {
         // in rigor we have tan(theta) = tan(cameraFOV) * deltaH / H
         // (where deltaH is the vertical amount we moved, and H the renderer height)
         // we loosely approximate tan(x) by x
         const pxToAngleRatio = THREE.Math.degToRad(this.camera.fov) / this.view.mainLoop.gfxEngine.height;
-        this.camera.rotation.z = (pointerX - this._onMouseDownMouseX) * pxToAngleRatio + this._onMouseDownRotZ;
-        this.camera.rotation.x = (pointerY - this._onMouseDownMouseY) * pxToAngleRatio + this._onMouseDownRotX;
-        this.view.notifyChange(false);
+
+        // update state based on pointer movement
+        this._state.rotateY = ((pointerX - this._onMouseDownMouseX) * pxToAngleRatio) + this._stateOnMouseDown.rotateY;
+        this._state.rotateX = limitRotation(
+            this.camera,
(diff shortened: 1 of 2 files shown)


gchoqueux:
`FlyControls`, `PanoramaControls`, `FirstPersonControls`, `PlanarControls` do similar things, it must be possible to factorize. (like the head movement for example)
Or, fov and rotation limits could be factorize and be customizable.

What is the contribution of `PanoramaControls` that can't be added to an existing one?

peppsac:
They all do slightly different things, so currently I believe it's better to keep them separated instead of having a big `Controls` with a lot of options.

peppsac:
I pushed a commit that improve FirstPersonControls instead of adding a new control class.

WDYT?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -4,32 +4,84 @@ import * as THREE from 'three';
...
     this._onMouseDownMouseX = pointerX;
     this._onMouseDownMouseY = pointerY;
-    this._onMouseDownRotZ = this.camera.rotation.z;
-    this._onMouseDownRotX = this.camera.rotation.x;
+
+    this._stateOnMouseDown = this._state.snapshot();
+}
+
+function limitRotation(camera3D, rot, verticalFOV) {
+    if (camera3D.fov / 2 + Math.abs(THREE.Math.radToDeg(rot)) >
```
gchoqueux: maybe use : 
```js
if (camera3D.fov + 2 * Math.abs(THREE.Math.radToDeg(rot)) > verticalFOV)
```

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -4,32 +4,84 @@ import * as THREE from 'three';
...
+        verticalFOV / 2) {
+        const s = Math.sign(rot);
```
gchoqueux: `const s` is useful?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -4,32 +4,84 @@ import * as THREE from 'three';
...
 
     this._onMouseDownMouseX = pointerX;
     this._onMouseDownMouseY = pointerY;
-    this._onMouseDownRotZ = this.camera.rotation.z;
-    this._onMouseDownRotX = this.camera.rotation.x;
+
+    this._stateOnMouseDown = this._state.snapshot();
+}
+
+function limitRotation(camera3D, rot, verticalFOV) {
```
gchoqueux, line 19: could you add comment
it's for head rotation?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -4,32 +4,84 @@ import * as THREE from 'three';
...
+    if (camera3D.fov / 2 + Math.abs(THREE.Math.radToDeg(rot)) >
+        verticalFOV / 2) {
+        const s = Math.sign(rot);
+        rot = s * THREE.Math.degToRad(verticalFOV - camera3D.fov) / 2;
+    }
+    return rot;
```
gchoqueux: could be replaced by : 
```js
const limit = THREE.Math.degToRad(verticalFOV - camera3D.fov) * 0.5;
rot= THREE.Math.clamp(-limit, limit);
```

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -4,32 +4,84 @@ import * as THREE from 'three';
...
+function onDocumentMouseWheel(event) {
+    let delta = 0;
+    if (event.wheelDelta !== undefined) {
+        delta = -event.wheelDelta;
+    // Firefox
+    } else if (event.detail !== undefined) {
+        delta = event.detail;
     }
+
+    this.camera.fov = Math.max(10,
```
gchoqueux: could you use `THREE.Math.clamp`

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -59,22 +111,59 @@ const MOVEMENTS = {
...
         super();
         this.camera = view.camera.camera3D;
         this.view = view;
         this.moves = new Set();
-        this.moveSpeed = options.moveSpeed || 10; // backward or forward move speed in m/s
-        this._isUserInteracting = false;
+        if (options.panoramaRatio) {
+            const radius = (options.panoramaRatio * 200) / (2 * Math.PI);
+            options.verticalFOV =
+                options.panoramaRatio == 2 ? 180 : THREE.Math.radToDeg(2 * Math.atan(200 / (2 * radius)));
```
gchoqueux, line 132: `options.panoramaRatio == 2` is a state?

gchoqueux (changes requested):

src/Renderer/ThreeExtended/FirstPersonControls.js
peppsac: Nope. 2 is the image ratio for equirectangular panoramas.

src/Renderer/ThreeExtended/FirstPersonControls.js
peppsac: Indeed, thanks.

gchoqueux (approved):

PR #560, opened 2017-10-27
Add panorama display capabilities

peppsac:
This PR adds all the needed code to display panoramas.
Panoramas can use either:
  - an equirectangular projection: in which case they're drawn using a sphere
  - a cylindrical projection: drawn on a cylinder

The underlying geometry is tiled which allows to display high resolution panoramas,
for instance using the StaticProvider.

The 2nd commit adds an example using the feature.

Note: this PR depends on PRs #556 #555 #554 and #559

### DIFF
--- a/src/Core/Prefab/Panorama/PanoramaTileBuilder.js
+++ b/src/Core/Prefab/Panorama/PanoramaTileBuilder.js
@@ -1,0 +1,153 @@
+import * as THREE from 'three';
+import OBB from '../../../Renderer/ThreeExtended/OBB';
+import Coordinates, { UNIT } from '../../Geographic/Coordinates';
+
+function PanoramaTileBuilder(ratio) {
+    this.tmp = {
+        coords: new Coordinates('EPSG:4326', 0, 0),
+        position: new THREE.Vector3(),
+        normal: new THREE.Vector3(0, 0, 1),
+    };
+
+    if (!ratio) {
+        throw new Error('ratio must be defined');
+    }
+    if (ratio === 2) {
+        this.equirectangular = true;
+        this.radius = 100;
+    } else {
+        this.equirectangular = false; // cylindrical proj
+        this.height = 200;
+        this.radius = (ratio * this.height) / (2 * Math.PI);
+    }
+}
+
+PanoramaTileBuilder.prototype.constructor = PanoramaTileBuilder;
+
+PanoramaTileBuilder.prototype.Center = function Center(params) {
+    this.Prepare(params);
+
+    this.uProjecte(0.5, params);
+    this.vProjecte(0.5, params);
+
+    params.center = this.VertexPosition(params).clone();
+
+    return params.center;
+};
+
+// prepare params
+// init projected object -> params.projected
+PanoramaTileBuilder.prototype.Prepare = function Prepare(params) {
+    if (this.equirectangular) {
+        params.projected = {
+            theta: 0,
+            phi: 0,
+            radius: this.radius,
+        };
+    } else {
+        params.projected = {
+            theta: 0,
+            radius: this.radius,
+            y: 0,
+        };
+    }
+};
+
+// coord u tile to projected
+PanoramaTileBuilder.prototype.uProjecte = function uProjecte(u, params) {
+    // both (theta, phi) and (y, z) are swapped in setFromSpherical
+    params.projected.theta = THREE.Math.lerp(
+        params.extent.west(UNIT.RADIAN),
+        params.extent.east(UNIT.RADIAN),
+        u);
+};
+
+// coord v tile to projected
+PanoramaTileBuilder.prototype.vProjecte = function vProjecte(v, params) {
+    if (this.equirectangular) {
+        params.projected.phi = Math.PI * 0.5 +
+            THREE.Math.lerp(
+                params.extent.south(UNIT.RADIAN),
+                params.extent.north(UNIT.RADIAN),
+                v);
+    } else {
+        params.projected.y =
+            this.height *
+            THREE.Math.lerp(params.extent.south(), params.extent.north(), v) / 180;
+    }
+};
+
+// get position 3D cartesian
+PanoramaTileBuilder.prototype.VertexPosition = function VertexPosition(params) {
+    if (this.equirectangular) {
+        this.tmp.position.setFromSpherical(params.projected);
+    } else {
+        this.tmp.position.setFromCylindrical(params.projected);
+    }
+    const swap = this.tmp.position.y;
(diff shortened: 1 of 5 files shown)


peppsac:
Rebased and ready to review.

zarov:
It doesn't seem to work for me in Firefox (57, Ubuntu), unless I load the tab without having the focus on it (right-click on non focused-tab, "refresh tab"). Does anyone else have the same problem ?

Another thing that bothers me is the fact that tiles take some time to reload again, when displayed again. I understand that it is the same functioning as for a map, but for a mid resolution panorama such as your examples, it is kinda a pity to see this transition. Is it possible to specify a case for low/mid res panorama, where this doesn't happen ?
But keep in mind that for (very) high resolution panorama, I agree that the original solution is a good solution.

And lastly, concerning cube projection, is it planned to be added ?

peppsac:
> It doesn't seem to work for me in Firefox (57, Ubuntu), unless I load the tab without having the focus on it (right-click on non focused-tab, "refresh tab"). Does anyone else have the same problem ?

I just rebased on master and it haved fixed your problem (thanks to #574 ). Can you test again?

> Another thing that bothers me is the fact that tiles take some time to reload again, when displayed again

Not sure what you mean here?

> Is it possible to specify a case for low/mid res panorama, where this doesn't happen ?

It's already possible to use a single image, this way you wouldn't have any transition.
Also if you (or anyone) can share some really high resolution panorama I would happily use them for the example.

> And lastly, concerning cube projection, is it planned to be added ?

Not yet.

zarov:
It works fine, thanks ! And nvm the second comment, all solved now too.

Unfortunately, a lot of high res panorama are partial, not full spherical, and also not free to use. I'm having a look into it, if I can find some.

zarov:
Got one ! There is this [Prague 34Gigapixels](http://360gigapixels.com/petrin-prague-photo/) that's available in CC BY-NC-SA, that can be downloaded [here (last link in the list)](http://www.jeffrey-martin.com/#gigapixel-photography). 

You can also get full equirectangular (you'll have to tile it) by doing a research similar to this https://www.google.fr/search?q=equirectangular&tbs=isz:lt,islt:70mp,sur:fmc&tbm=isch&source=lnt

src/Core/Prefab/Panorama/PanoramaTileBuilder.js
```diff
@@ -0,0 +1,153 @@
...
+    } else {
+        this.equirectangular = false; // cylindrical proj
+        this.height = 200;
+        this.radius = (ratio * this.height) / (2 * Math.PI);
+    }
+}
+
+PanoramaTileBuilder.prototype.constructor = PanoramaTileBuilder;
+
+PanoramaTileBuilder.prototype.Center = function Center(params) {
```
zarov: Maybe invert `Center` and `Prepare` for coherence with other builders.

src/Core/Prefab/Panorama/PanoramaTileBuilder.js
```diff
@@ -0,0 +1,153 @@
...
+        params.projected = {
+            theta: 0,
+            radius: this.radius,
+            y: 0,
+        };
+    }
+};
+
+// coord u tile to projected
+PanoramaTileBuilder.prototype.uProjecte = function uProjecte(u, params) {
```
zarov: Same here, `uProjecte` and `vProjecte` should come after `VertexPosition` and `VertexNormal`.

src/Core/Prefab/Panorama/PanoramaTileBuilder.js
```diff
@@ -0,0 +1,153 @@
...
+        coords: new Coordinates('EPSG:4326', 0, 0),
+        position: new THREE.Vector3(),
+        normal: new THREE.Vector3(0, 0, 1),
+    };
+
+    if (!ratio) {
+        throw new Error('ratio must be defined');
+    }
+    if (ratio === 2) {
+        this.equirectangular = true;
```
zarov, line 16: Couldn't we change this to an enum, if we want to add cube support later ?

src/Core/Prefab/PanoramaView.js
```diff
@@ -0,0 +1,206 @@
...
+    this.camera.camera3D.near = 0.1;
+    this.camera.camera3D.far = 1000;
+    this.camera.camera3D.up = coordinates.geodesicNormal;
+    this.camera.camera3D.quaternion.setFromUnitVectors(
+        new THREE.Vector3(0, 1, 0), coordinates.geodesicNormal);
+    this.camera.camera3D.updateProjectionMatrix();
+    this.camera.camera3D.updateMatrixWorld();
+
+    const tileLayer = createPanoramaLayer('panorama', coordinates, ratio, options);
+    this.scene.add(tileLayer.object3d);
```
zarov: Not necessary, it's added later in `View#addLayer`.

src/Core/Prefab/PanoramaView.js
```diff
@@ -0,0 +1,206 @@
...
+
+    View.prototype.addLayer.call(this, tileLayer);
```
zarov: Why not `this.addLayer` instead ?

zarov (changes requested):
Seems quite good to me, needs some changes though.

peppsac:
> Got one ! There is this Prague 34Gigapixels that's available in CC BY-NC-SA, that can be downloaded here (last link in the list).

Unfortunately you can only download the preprocessed files (split in small tiles) so I can't use it.

Note that I ran my preprocess app again with different settings and the quality should be better now.

src/Core/Prefab/PanoramaView.js
peppsac: Because `PanoramView.addLayer` is overriden currently, and is meant to be used with 'color' layers.

src/Core/Prefab/PanoramaView.js
peppsac: Correct, fixed.

src/Core/Prefab/Panorama/PanoramaTileBuilder.js
peppsac: I changed the order of the functions.

zarov:
Changes are good to me.
Quality is quite good now ! Thanks for that improvement.

zarov (approved):

nosy-b:
Thanks @peppsac, very cool new feature!

I'm a bit late but a few remarks:
- When creating a new example (panorama.html), please add some comments in it
- It should appear in the examples/index.html
- Why do you refine the geometry (that much)? It creates too much triangles

PR #561, opened 2017-10-30
refactor (geometry): simplify TileGeometry code

peppsac:
This commit simplifies the code of TileGeometry and fixes a cache
bug: the invariant used as cache key (nsegments) failed when mixing
geometries with/without skirt.

### DIFF
--- a/src/Core/TileGeometry.js
+++ b/src/Core/TileGeometry.js
@@ -35,24 +35,13 @@ function TileGeometry(params, builder) {
     this.center = builder.Center(params).clone();
     this.OBB = builder.OBB(params);
 
-    // TODO : free array
-
-    var bufferAttribs = this.computeBuffers(params, builder);
+    const bufferAttribs = this.computeBuffers(params, builder);
 
     this.setIndex(bufferAttribs.index);
     this.addAttribute('position', bufferAttribs.position);
     this.addAttribute('normal', bufferAttribs.normal);
     this.addAttribute('uv_wgs84', bufferAttribs.uv.wgs84);
     this.addAttribute('uv_pm', bufferAttribs.uv.pm);
-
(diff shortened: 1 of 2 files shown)


src/Core/TileGeometry.js
```diff
@@ -35,24 +35,13 @@ function TileGeometry(params, builder) {
...
-
-    bufferAttribs.position = null;
-    bufferAttribs.normal = null;
-    bufferAttribs.uv.pm = null;
-
-    // Update cache
-    if (!cache.getRessource(params.segment)) {
-        cache.addRessource(params.segment, bufferAttribs);
-    }
+    this.addAttribute('uv_pm', new THREE.BufferAttribute(bufferAttribs.uv.pm, 1));
```
autra: I would suggest having either everything as BufferAttributes, or everything as TyperArray in this `bufferAttribs`.

src/Core/TileGeometry.js
```diff
@@ -180,113 +167,92 @@ TileGeometry.prototype.computeBuffers = function computeBuffers(params, builder)
...
 
     // TODO: WARNING beware skirt's size influences performance
     // The size of the skirt is now a ratio of the size of the tile.
     // To be perfect it should depend on the real elevation delta but too heavy to compute
     if (!params.disableSkirt) {
+        // We compute the actual size of tile segment to use later for the skirt.
+        const segmentSize = new THREE.Vector3().fromArray(outBuffers.position).distanceTo(
+            new THREE.Vector3().fromArray(outBuffers.position, 3));
+
         const r = segmentSize; // Tile size divided by the number of segments
```
autra: So we can get rid of `r` right?

autra (changes requested):

autra (approved):
Thanks!

PR #563, opened 2017-10-30
fix (geometry): don't make material state depends on layer update state

### LINKED ISSUE #562: Update of layers opacity and visibility is broken
## Your Environment

## Context
With the globe, the layer opacity and visibility is not correctly updated when we change it. 
It seems like this problem was not visible before commit e3db2308f3cca7f6eaa17f68779d48909d92339c.

1. open the main globe itowns example
2. play with the layer manager tool (change opacity/visibility of the layers).

## Expected Behavior
The layers states (opacity/visibility) are immediately changing.

## Actual Behavior
The layers states are not changing, unless we zoom/de-zoom, and tiles are not fully loaded.

peppsac:
This is a quick fix for issue #562 but a better fix would be to split correctly
material state update to another file.

Fixes #562

Note:  I tried to add a unit test to cover this bug but `LayeredMaterialNodeProcessing.js` needs some refactoring to ease unit testing...

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -178,20 +178,24 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
             initNodeImageryTexturesFromParent(node, node.parent, layer);
         }
     }
-    // An update is pending / or impossible -> abort
-    if (!node.layerUpdateState[layer.id].canTryUpdate(ts)) {
-        return;
-    }
 
     // Node is hidden, no need to update it
     if (!node.isDisplayed() || !layer.visible) {
         return;
     }
 
+    // TODO: move this to defineLayerProperty() declaration
+    // to avoid mixing layer's network updates and layer's params
     // Update material parameters
     const layerIndex = material.indexOfColorLayer(layer.id);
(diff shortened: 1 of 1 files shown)


autra (approved):
Thanks!

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -178,21 +178,25 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
-        return;
-    }
 
     // Node is hidden, no need to update it
     if (!node.isDisplayed() || !layer.visible) {
         return;
     }
 
+    // TODO: move this to defineLayerProperty() declaration
+    // to avoid mixing layer's network updates and layer's params
```
autra, line 188: wait, isn't it already the case in View.js:174?

autra: Ok, done for geometryLayer, but not for color layers.

autra: Actually, this code set an implicit dependency on LayeredMaterial or at least part of its api (this probably makes sense for this update function here). So if we move it somewhere else, the dependency will also be transferred. I'm not against it per se, but it's something we want to keep it mind.

PR #564, opened 2017-10-31
Unit test issue 562

peppsac:
## Description
This is a followup of PR #563 : this PR adds a unit test reproducing the issue #562

2 commits:
- the first one refactor the code a bit to ease unit-testing
- the 2nd is the real unit test

The unit test is failing when reverting the fix from #563 .

### DIFF
--- a/test/layeredmaterial_unit_test.js
+++ b/test/layeredmaterial_unit_test.js
@@ -1,0 +1,44 @@
+import LayerUpdateState from '../src/Core/Layer/LayerUpdateState';
+import { updateLayeredMaterialNodeImagery } from '../src/Process/LayeredMaterialNodeProcessing';
+/* global describe, it */
+
+const assert = require('assert');
+
+describe('material state vs layer state', function () {
+    let opacity = 0.0;
+
+    const node = {
+        parent: { },
+        layerUpdateState: {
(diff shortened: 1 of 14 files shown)


test/layeredmaterial_unit_test.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('material state vs layer state', function () {
+    let opacity = 0.0;
+
+    const node = {
+        parent: { },
+        layerUpdateState: {
+            test: new LayerUpdateState(),
+        },
+        material: {
```
autra, line 15: Food for thought, what you did here is actually a cheap version of sinon.js (or equivalent). It's a sort of spy. I'm not saying we should introduce this dependency, but let's keep it in mind if this use case becomes frequent.

autra (approved):
Thinking out loud: I have mixed feelings about that in the long term. Actually maybe a better solution would be to properly mock the dom (with jsdom for instance). We won't be able to shuffle dependencies around all the time and this would also remove the need for sinon.js or an add-hoc mock for this type of test. WDYT?

That being said, in this instance, it's a good thing to have CanceledCommandException in its own file, I'm fine having constants in their own module, and it's good to have a unit test :-) Thanks for that!

PR #565, opened 2017-10-31
fix (geometry): sync material visibility and layer visibility in all case

### LINKED ISSUE #562: Update of layers opacity and visibility is broken

peppsac:
Followup of b524928a15d0dc8b8bcf2862d4f7d8ced5f993b0

Fix #562

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -180,7 +180,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     }
 
     // Node is hidden, no need to update it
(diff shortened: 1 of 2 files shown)


autra (approved):
LGTM. Thanks!

PR #568, opened 2017-11-10
Add unit test test for scheduler

peppsac:
This PR adds a simple unit test for scheduler.

Some 3dtiles code have been slightly modified to allow to instanciate a `Scheduler` without the need to polyfill `TextDecoder`.

This is some preparatory work for a later PR that would improve commands priority handling.

### DIFF
--- a/src/Core/Scheduler/Providers/3dTiles_Provider.js
+++ b/src/Core/Scheduler/Providers/3dTiles_Provider.js
@@ -65,7 +65,6 @@ function $3dTilesIndex(tileset, baseURL) {
 
 function $3dTiles_Provider() {
     Provider.call(this);
-    this.b3dmLoader = new B3dmLoader();
 }
 
 $3dTiles_Provider.prototype = Object.create(Provider.prototype);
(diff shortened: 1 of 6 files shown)


src/Core/Scheduler/Providers/3dTiles_Provider.js
```diff
@@ -220,13 +220,15 @@ function configureTile(tile, layer, metadata, parent) {
...
 
-const textDecoder = new TextDecoder('utf-8');
 $3dTiles_Provider.prototype.executeCommand = function executeCommand(command) {
     const layer = command.layer;
     const metadata = command.metadata;
     const tile = new THREE.Object3D();
     configureTile(tile, layer, metadata, command.requester);
     const path = metadata.content ? metadata.content.url : undefined;
+    this._textDecoder = this._textDecoder || new TextDecoder('utf-8');
+    const textDecoder = this._textDecoder;
```
gchoqueux, line 230: `const textDecoder` is still useful?

gchoqueux (approved):

src/Core/Scheduler/Providers/3dTiles_Provider.js
peppsac: Yes it's used at lines 245 and 247

gchoqueux: you don't use `this._textDecoder` instead because of the context?

peppsac: yep

PR #569, opened 2017-11-10
fix: use babel-polyfill and url-polyfill to support IE11

### LINKED ISSUE #510: IE11 : Promise Polyfill doesn't seem to apply
When trying to load itowns examples (for instance : http://www.itowns-project.org/itowns/examples/globe.html )  with ie11 browser? I've the following result : 

globe-ie11

with that error in the console : 

promise-undef

As itowns bundle contains es6-promise polyfill (as a result of #100 ), it seems that it isn't enough to make iTowns work on ie11 browser.

## Your Environment
OS : Windows 7 Professionnel - SP1 (via Oracle VM VirtualBox)
Browser : IE11 (version : 11.0.9600.18792)

## Context

Integration of iTowns in geoportal web site (live template [here](https://demos.ign.fr/geoportail/refonte-3d/map-viewer.html) : clic on  bottom right 2D|3D button to switch to itowns mode), where ie11 compatibility would be greatly appreciated (IE11 comes in third position among browsers used to acces our web site). 

1. With IE11 browser, open this sample URL : http://www.itowns-project.org/itowns/examples/globe.html
2. See the result :'(

## Expected Behavior

itowns-ff

(Obtained with Firefox. I hope it to be the same with IE11)

## Actual Behavior

globe-ie11

And following error in console : 

```
SCRIPT5009: « Promise » est indéfini
Fichier : itowns.js, ligne : 12, colonne : 12627
```

I'm not very aware of es6-promise and  webpack use in itowns. I don't know if it is has something to do with that problem but we had to patch 4.1.0 version of es6-promise in our geoportal-accesslib project : https://github.com/IGNF/geoportal-access-lib/blob/master/lib/es6-promise/es6-promise-4.1.0.js
(reason here : https://github.com/IGNF/geoportal-access-lib/blob/6ff4fe4d9a3255881d333daa4440b3a3331c8a75/src/Config.js#L8)

peppsac:
This change increases itowns.js bundle size by ~100KB.

Fixes #510

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,14 +32,15 @@
   },
   "homepage": "https://itowns.github.io/",
   "dependencies": {
+    "babel-polyfill": "^6.26.0",
     "babel-runtime": "^6.20.0",
     "earcut": "^2.1.1",
-    "es6-promise": "^4.0.5",
     "js-priority-queue": "^0.1.5",
     "jszip": "^3.1.3",
     "monotone-convex-hull-2d": "^1.0.1",
     "text-encoding-utf-8": "^1.0.1",
     "togeojson": "^0.16.0",
+    "url-polyfill": "^1.0.8",
     "whatwg-fetch": "^2.0.2"
   },
   "peerDependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -226,7 +226,7 @@
       "dev": true,
       "requires": {
         "babel-core": "6.25.0",
-        "babel-polyfill": "6.23.0",
+        "babel-polyfill": "6.26.0",
         "babel-register": "6.24.1",
         "babel-runtime": "6.23.0",
         "chokidar": "1.7.0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
   },
   "homepage": "https://itowns.github.io/",
   "dependencies": {
+    "babel-polyfill": "^6.26.0",
     "babel-runtime": "^6.20.0",
     "earcut": "^2.1.1",
     "es6-promise": "^4.0.5",
```
tbroyer: This needs to be removed then.

tbroyer (dismissed):
> This change increases itowns.js bundle size by ~100KB.

fwiw, I believe that was the reason for not using it initially.

Should we build 2 bundles depending on browser support? (i.e. whether one wants to support IE11 –and earlier?– or not)

Or choose other, smaller, polyfills? E.g. `es6-promise/auto` and [`String.prototype.includes`](https://github.com/mathiasbynens/String.prototype.includes)

autra:
> fwiw, I believe that was the reason for not using it initially.

Are you thinking about https://github.com/iTowns/itowns/pull/417 or another discussion?

peppsac:
@gcebelieu since you reported https://github.com/iTowns/itowns/issues/510, could you confirm that this PR works for you?

@tbroyer I've removed es6-promise.

gcebelieu:
Thanks @peppsac, I will have access to a windows with ie11 machine only tomorrow. I'll test then and tell you.

gcebelieu:
@peppsac I couldn't find a working example on ie11. For instance, index.html or globe.html examples are now blocking on this error : 

> "dat" est indéfini

in this piece of code (line 9) from GuiTools.js

```
dat.GUI.prototype.removeFolder = function removeFolder(name) {
    var folder = this.__folders[name];
    if (!folder) {
        return;
    }
    folder.close();
    this.__ul.removeChild(folder.domElement.parentNode);
    delete this.__folders[name];
    this.onResize();
};
```

Maybe a side effet, not really linked to this PR, but is there an example that doesn't involves gui stuff ? On which example did you make your ie tests ?

autra (changes requested):
Things are not perfect for the globe but it's already a *lot* better (other issues look like webgl ones). Other examples are working good. You just need to amend your commit messages ("fix(compat)" maybe?). Also imho you can squash these 2 commits because babel-polyfill replaces es6-promise.

@gcebelieu weird. Maybe dat.gui hasn't been downloaded by your browser? Can you check the `network` tab of your devtools? That being said, we do have some examples without this js: planar.html, multiglobe.html and a bunch of others (split, planar kml, globe gpx, the cube example... check the html in `examples/`)

autra (approved):

autra:
@gcebelieu I believe your issue is not linked to this PR. Can you retest with one of the example not including dat.gui please?

gcebelieu:
I have other errors with examples non involving dat.gui but this may comes from my environment (I'm not very familiar with windows).
As Promise errors disappeared, I'm OK with that PR.

autra:
Okay, let's merge that. @gcebelieu please don't hesitate to open issues if you still have errors. You can directly test on http://www.itowns-project.org/itowns/examples/index.html, this'll remove some environment issues hopefully.

PR #570, opened 2017-11-10
 fix (core): Cancel command or not allow elevation when tile's elevation level is superior

gchoqueux:
## Description
* Not allow parent's elevation when tile's elevation level is superior
* Cancel command or not allow elevation when tile's elevation level is superior

## Motivation and Context
There's conflict between with two elevation layers
Second elevation layer can be resolved before the first layer
Cancel command if target level is lower or equal to current elevation level

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -40,8 +40,11 @@ function initNodeImageryTexturesFromParent(node, parent, layer) {
 }
 
 function initNodeElevationTextureFromParent(node, parent, layer) {
-    // inherit parent's elevation texture
-    if (parent.material && parent.material.getElevationLayerLevel() > EMPTY_TEXTURE_ZOOM) {
+    // Inherit parent's elevation texture. Note that contrary to color layers the elevation level of the
+    // node might not be EMPTY_TEXTURE_ZOOM in this init function. That's because we can have
+    // multiple elevation layers (thus multiple calls to initNodeElevationTextureFromParent) but a given
+    // node can only use 1 elevation texture
+    if (parent.material && parent.material.getElevationLayerLevel() > node.material.getElevationLayerLevel()) {
         const coords = node.getCoordsForLayer(layer);
 
         const texture = parent.material.textures[l_ELEVATION][0];
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -117,6 +117,14 @@ function refinementCommandCancellationFn(cmd) {
         return false;
     }
 
+    // Cancel the command if the tile already has a better texture.
+    // This is only needed for elevation layers, because we may have several
+    // concurrent layers but we can only use one texture.
+    if (cmd.layer.type == 'elevation' &&
+        cmd.targetLevel <= cmd.requester.material.getElevationLayerLevel()) {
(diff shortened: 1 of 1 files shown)


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -41,7 +41,7 @@ function initNodeImageryTexturesFromParent(node, parent, layer) {
 
 function initNodeElevationTextureFromParent(node, parent, layer) {
     // inherit parent's elevation texture
-    if (parent.material && parent.material.getElevationLayerLevel() > EMPTY_TEXTURE_ZOOM) {
+    if (parent.material && parent.material.getElevationLayerLevel() > node.material.getElevationLayerLevel()) {
```
peppsac: Is this change really needed?

(when would `initNodeElevationTextureFromParent` with node's elevation level not being EMPTY_TEXTURE_ZOOM)

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -117,6 +117,14 @@ function refinementCommandCancellationFn(cmd) {
         return false;
     }
 
+    // There's conflict between with two elevation layers
```
peppsac: You could reword this comment:
```js
// Cancel the command if the tile already has a better texture.
// This is only needed for elevation layers, because we may have several
// concurrent layers but we can only use one texture.
```

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -352,6 +360,15 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, force)
                 return;
             }
 
+            // There's conflict between two elevation layers
```
peppsac:
```js
// Do not apply the new texture if its level is < than the current one.
// This is only needed for elevation layers, because we may have several
// concurrent layers but we can only use one texture.
```

peppsac (changes requested):

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: it's precisely this case that I avoid.
Because with two elevation layers we passe twice in `initNodeElevationTextureFromParent`

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: ok, thanks

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: ok, thanks

src/Process/LayeredMaterialNodeProcessing.js
peppsac: Oh right. Then modify the comment above because it's not obvious.
Seomthing like:
```js
      // Inherit parent's elevation texture. Note that contrary to color layers the elevation level of the 
      // node might not be EMPTY_TEXTURE_ZOOM in this init function. That's because we can have
      // multiple elevation layers (thus multiple calls to initNodeElevationTextureFromParent) but a given
      // node can only use 1 elevation texture
```

peppsac (approved):
(please see https://github.com/iTowns/itowns/pull/570/files#r150302965 before merging)

Thanks.

gchoqueux:
ok

PR #571, opened 2017-11-13
chore(example): Minor fix network options

gliegard:
## Description
fix example : network option for WFS flux

## Motivation and Context
The examples deployed on itowns-project.org doesn't work : we can't see buildings.

http://www.itowns-project.org/itowns/examples/wfs.html
http://www.itowns-project.org/itowns/examples/globe_wfs_extruded.html


peppsac:
Thanks for the fix.
Can you reword your git commit message to match #518 ?

gliegard:
Yep, reworded !

peppsac (approved):

PR #573, opened 2017-11-14
perf(wfs): Avoid creating needless objects

gliegard:
## Description
Fix performances issues introduced in #519

## Motivation and Context
In the example globe_wfs_extruded, we can see that it's slow.

It's because, #519 introduced a lot of objects creation needlessly.

@autra explained that here https://github.com/iTowns/itowns/pull/519#pullrequestreview-72819496

### DIFF
--- a/src/Renderer/ThreeExtended/Feature2Mesh.js
+++ b/src/Renderer/ThreeExtended/Feature2Mesh.js
@@ -40,14 +40,13 @@ function getColor(options, properties) {
     return randomColor();
 }
 
-function createColorArray(length, color, brightness) {
-    const colors = new Array(length * 3);
-    for (let i = 0; i < length; ++i) {
-        colors[3 * i] = color.r * brightness;
-        colors[3 * i + 1] = color.g * brightness;
-        colors[3 * i + 2] = color.b * brightness;
+function fillColorArray(colors, offset, length, r, g, b) {
+    const len = offset + length;
+    for (let i = offset; i < len; ++i) {
+        colors[3 * i] = r;
+        colors[3 * i + 1] = g;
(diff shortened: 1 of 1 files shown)


src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -121,41 +120,49 @@ function addFaces(indices, length, offset) {
...
+    let offset = 0;
 
     /* eslint-disable guard-for-in */
     for (const id in coordinates.featureVertices) {
         const { contour, property } = extractFeature(coordinates, properties, id);
         // get altitude from properties
         const altitude = getAltitude(options, property);
         const newVertices = coordinatesToVertices(contour, altitude);
-        vertices = vertices.concat(newVertices);
+        for (let i = 0; i < newVertices.length; i++) {
```
peppsac: You should instead change `coordinatesToVertices` signature and add an `target` optionnal parameter.

Then use it like this: 
```js
coordinatesToVertices(contour, altitude, new Float32Array(vertices.buffer, offset * 4));
```

peppsac: (Or pass the target and offset as arguments if you prefer)

gliegard:
updated

src/Renderer/ThreeExtended/Feature2Mesh.js
gliegard: thanks for your suggestion, I've updated the code !

gchoqueux:
@gliegard I rebased the branch

gliegard:
@itowns-reviewers

peppsac (approved):
Looks good, thanks.

PR #575, opened 2017-11-15
fix(core): Don't try to add scene to itself

autra:
## Description

User can use the scene as layer.object3d. In this case, we shouldn't try
to add it to itself 

## Motivation and context

This fixes `THREE.Object3D.add: object can't be added as a child of itself. ` errors when layer.object3d is the scene object.


peppsac (approved):
Thanks

PR #576, opened 2017-11-15
feat(wfs): add a function to filter feature

gliegard:
## Description
feat(wfs): add a function to filter feature

## Motivation and Context
The goal of this PR is to filter the features to show, using properties.
In the example 'globe wfs extruded', we can see how it is used to clean bad data.

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -78,6 +78,10 @@ function extrudeBuildings(properties) {
     return properties.hauteur;
 }
 
+function acceptFeature(properties) {
+    return !!properties.hauteur;
+}
+
 globeView.addLayer({
     type: 'geometry',
     update: itowns.FeatureProcessing.update,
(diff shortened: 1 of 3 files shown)


src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -192,6 +192,9 @@ function readFeatureCollection(crsIn, crsOut, json, filteringExtent, options) {
 
     let featureIndex = 0;
     for (const feature of json.features) {
+        if (options.filter && !options.filter(feature.properties)) {
```
autra: Let's make instead an early return undefined in readFeature. You also want this for `feature` object I guess, and it would avoid the `continue` keyword.

examples/globe_wfs_extruded.js
```diff
@@ -78,13 +78,20 @@ function extrudeBuildings(properties) {
     return properties.hauteur;
 }
 
+function acceptFeature(properties) {
+    if (properties.hauteur === 0) {
```
autra: This is actually `return properties.hauteur !== 0`, or maybe even better: `return !!properties.hauteur`: this would also eliminate feature not having this property.

autra (changes requested):
Thanks for this! A useful feature indeed. Already looking good :-)

gliegard:
thanks, updated

autra (approved):
Thanks!

PR #577, opened 2017-11-15
pointcloud object3d fix

autra:
## Description

Really honour layer.object3d when using pointclouds.

## Motivation and Context

One of the interesting things in using a custom object3d is to have it translated/rotationed/shown or hidden. But this isn't really working with the current code because some part of it ignored this local object3d transforms.

### DIFF
--- a/src/Core/Scheduler/Providers/PointCloudProvider.js
+++ b/src/Core/Scheduler/Providers/PointCloudProvider.js
@@ -226,7 +226,6 @@ export default {
             points.material.transparent = layer.opacity < 1.0;
             points.material.uniforms.opacity.value = layer.opacity;
             points.updateMatrix();
-            points.updateMatrixWorld(true);
             points.layers.set(layer.threejsLayer);
             points.layer = layer.id;
             return points;
--- a/src/Core/Scheduler/Providers/PointCloudProvider.js
+++ b/src/Core/Scheduler/Providers/PointCloudProvider.js
@@ -135,6 +135,18 @@ export default {
         if (!layer.file) {
             layer.file = 'cloud.js';
         }
+        if (!layer.group) {
+            layer.group = new THREE.Group();
+            layer.object3d.add(layer.group);
+            layer.group.updateMatrixWorld();
+        }
+
+        if (!layer.bboxes) {
+            layer.bboxes = new THREE.Group();
+            layer.object3d.add(layer.bboxes);
+            layer.bboxes.updateMatrixWorld();
+        }
+
         // default options
         layer.fetchOptions = layer.fetchOptions || {};
         layer.octreeDepthLimit = layer.octreeDepthLimit || -1;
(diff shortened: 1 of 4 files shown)


src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -135,6 +135,18 @@ export default {
         if (!layer.file) {
             layer.file = 'cloud.js';
         }
+        if (!layer.group) {
+            layer.group = new THREE.Group();
+            layer.object3d.add(layer.group);
```
gchoqueux, line 140: you add a new group to apply transform on it?
because it doesn't work on `layer.object3d`?
if so, it should work on  `layer.object3d`

autra: No. Tranformations works fine on layer.object3d, and that's what user should transform. These 2 groups (layer.group and layer.bboxes) are just implementation details of  PointCloudProvider and Processing. I just moved this piece of code around, and fixed it (they were added previously to scene, which is wrong now that we have the object3d mechanism).  For reference, they are useful to hide bboxes and pointclouds independently, for example.

src/Process/PointCloudProcessing.js
```diff
@@ -103,6 +103,7 @@ function initBoundingBox(elt, layer) {
     elt.obj.boxHelper.material.linewidth = 2;
     elt.obj.boxHelper.layers.mask = layer.bboxes.layers.mask;
     layer.bboxes.add(elt.obj.boxHelper);
+    elt.obj.boxHelper.updateMatrixWorld();
```
gchoqueux, line 106: `boxHelper`  should be the child of the element, because it doesn't work if the parent's element have a transformation

autra: That is debatable. elt is an internal object to PointCloudProcessing and this processing won't move it. The important part is that both are in the hierarchy of `object3d`, and that's the goal of this PR.

 I'm not saying I'm against it, but I need to think about it (it's cool to have them each in a different group to hide them, something we already do. Today we don't need to move individual patch of points, and imho it's unlikely we'll need it).

peppsac (approved):
Looks ok to me, thanks.

src/Core/Scheduler/Providers/PointCloudProvider.js
gchoqueux: It's better to tie the obbs as we do for the tiles.
There is in `tileDebug.js` an example of visibility management. it'll be nice to follow this line.
Moreover this approach will not work in the example `LAYERS TRANSFORMATIONS`

autra: Please note that this PR is merely a bugfix PR over existing code and structures. You want a refacto, here, right?

PR #581, opened 2017-11-17
Browserstack testing

peppsac:
http://browserstack.com/ is a service allowing to test a web app on various browser / OS.
They also offer a free plan to open-source project so I subscribed one and started playing with it.

A few things:
- this relies on Selenium and can all the tests can be run locally (you just need to have a selenium server running on your machine and a webdriver installed). For instance, here's how to run the tests on a local Chromium:
   * install chromium-driver
   * download selenium and start it: `java -jar -Dwebdriver.chrome.driver=/usr/bin/chromedriver  selenium-server-standalone-3.5.3.jar`
   * run the test (from itowns2): `./node_modules/.bin/wdio wdio.conf.js`
- travis has a BrowserStack integration (see `addons` section in .travis.yml
- I only wrote a very basic test - but once setup we could write more interesting ones
- timeouts are painful: low timeout values can cause failures, big values can stall the test pipeline for a long time if there's a real bug...
- we can run 5 parallel test and queue 5 more. If we try to execute another test when the queue is full the test will fail...
- the free plan only allows access to 5 users
- browserstack is pretty cool: you can watch the test running, see the screenshot of failed tests, etc
- this test mode should replace the `itowns-testing.js` test mode IMHO. When done, we won't need to keep splitting our examples in separate  `html`  and `js` files.

Open questions:
- I'd like to run browserstack tests only once per branch. I don't think it's useul to run them for 2 versions of node. Any idea on how to achieve this?
- I run `npm start` before launching the tests (see wdio.conf.js). I'm not sure this is the best way...

So, WDYT?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,9 @@ node_js:
 after_success:
   - openssl aes-256-cbc -K $encrypted_822645c864aa_key -iv $encrypted_822645c864aa_iv -in deploy_key.enc -out deploy_key -d
   - bash autodeploy.sh
+
+addons:
+  browserstack:
+    username: "itowns1"
+    access_key:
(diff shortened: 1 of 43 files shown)


autra:
Looks very interesting indeed. What do you plan to test on this plateforme? Is it just to check if all our supported browsers are working or do you have something else in mind (integration test? Screenshots?)?

peppsac:
@autra at first I'd like to use this to validate that all our examples work on the supported platforms.

Then we can do more complex things: simulate user interaction for instance, screenshots comparison, etc

tbroyer:
Have you looked at SauceLabs? You might not need Selenium then.
Also, how about Karma for driving the tests?

zarov:
@tbroyer from what I've seen, SauceLabs doesn't support WebGL context yet (and it's in their backlog since 2014) (see this most recent comment from april 2017 by Chiarng Lin, a SauceLabs member https://saucelabs.ideas.aha.io/ideas/SLIDEA-I-54).

Otherwise, I couldn't try the automate and local testing, but the live app is really nice to have ! A bit slow though sometimes.

peppsac:
>  Also, how about Karma for driving the tests?

We already use `mocha` for other tests so I used it for BrowserStack.

tbroyer:
@zarov Ah yes, I now remember seeing this when we started iTowns 2 ~2 years ago.

@peppsac it's not a "mocha or karma" situation; karma is "only" a runner, and will happily use mocha as a test framework; see https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai for instance.
But, looking more closely at your proposal, it looks like the goal is different. Karma would let us run our existing tests in real browsers (locally, seamlessly without the need for a Selenium server; or remotely, on BrowserStack for instance), whereas you're proposing doing some smoke-testing of the examples.
Actually, we could (technically) have the wdio-based smoke tests in parallel with Karma (rewriting `itowns-testing.js` to actually run in a browser rather than "emulate" it; or even detect the environment so it can run in both: Mocha cli for very-fast feedback, and Karma for heavier/slower but more accurate tests)

peppsac:
I don't have time to work on this so I'll close this for now.

If someone wants to take over this task, it'd be welcomed :)

(also I've written a small command line tool to produce a screenshot from all itowns example: https://github.com/peppsac/itowns-screenshot - it's not robust enough to replace CI, but it's handy to check that all examples work)

PR #583, opened 2017-11-17
Limit the addition of layers

gchoqueux:
## Description
Adjust the ability to add more color layers based on gpu's capabilities
and limit the addition of layers so that the abilities don't exceed

## Motivation and Context
Before the capabilities were clamp to 15 color layers
and if the limit is exceeded it will cause malfunction

**example if you exceed limit**
bug layer

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -166,6 +166,22 @@ export function createGlobeLayer(id, options) {
         return false;
     }
 
+    /**
+     * Counts the number of texture samplers for color layers.
+     * @param      {Array} colorLayers The color layers
+     * @return     {number} Number of texture samplers for color layers.
+     */
+    wgs84TileLayer.countTextureSamplersForColorLayers = (colorLayers) => {
+        let occupancy = 0;
+        for (const layer of colorLayers) {
+            const projection = layer.projection || layer.options.projection;
+            // 'EPSG:3857' occupies the maximum 3 textures on tiles
+            // 'EPSG:4326' occupies 1 textures on tile
+            occupancy += projection == 'EPSG:3857' ? 3 : 1;
+        }
+        return occupancy;
+    };
+
     wgs84TileLayer.update = processTiledGeometryNode(globeCulling(2), subdivision);
     wgs84TileLayer.builder = new BuilderEllipsoidTile();
     wgs84TileLayer.onTileCreated = nodeInitFn;
--- a/src/Core/System/Capabilities.js
+++ b/src/Core/System/Capabilities.js
@@ -1,11 +1,28 @@
+import SampleTestFS from '../../Renderer/Shader/SampleTestFS.glsl';
+import SampleTestVS from '../../Renderer/Shader/SampleTestVS.glsl';
+
 // default values
 let logDepthBufferSupported = false;
 let maxTexturesUnits = 8;
+
+function _WebGLShader(renderer, type, string) {
+    const gl = renderer.context;
+    const shader = gl.createShader(type);
+
+    gl.shaderSource(shader, string);
+    gl.compileShader(shader);
+    return shader;
+}
+
+function isFirefox() {
+    return navigator && navigator.userAgent && navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+}
 
 export default {
     isLogDepthBufferSupported() {
         return logDepthBufferSupported;
     },
+    isFirefox,
     isInternetExplorer() {
         const internetExplorer = false || !!document.documentMode;
         return internetExplorer;
@@ -17,16 +34,33 @@ export default {
         const gl = renderer.context;
         maxTexturesUnits = gl.getParameter(gl.MAX_TEXTURE_IMAGE_UNITS);
 
-        const debugInfo = gl.getExtension('WEBGL_debug_renderer_info');
-        if (debugInfo !== null) {
-            const vendor = gl.getParameter(debugInfo.UNMASKED_VENDOR_WEBGL);
-            if (vendor.indexOf('mesa') > -1 || vendor.indexOf('Mesa') > -1) {
-                maxTexturesUnits = Math.min(16, maxTexturesUnits);
+        const program = gl.createProgram();
+        const glVertexShader = _WebGLShader(renderer, gl.VERTEX_SHADER, SampleTestVS);
+
+        let fragmentShader = `#define SAMPLE ${maxTexturesUnits}\n`;
+        fragmentShader += SampleTestFS;
+
+        const glFragmentShader = _WebGLShader(renderer, gl.FRAGMENT_SHADER, fragmentShader);
+
+        gl.attachShader(program, glVertexShader);
+        gl.attachShader(program, glFragmentShader);
+        gl.linkProgram(program);
+
+        if (gl.getProgramParameter(program, gl.LINK_STATUS) === false) {
+            // eslint-disable-next-line no-console
+            console.warn(`Sampler limit exceeded: it's down from ${maxTexturesUnits} to 16`);
+            if (isFirefox()) {
+                // eslint-disable-next-line no-console
+                console.warn('it can come from a bug mesa/firefox \n\n' +
+                    'Error compile shader when using more than 16 sampler uniforms\n\n' +
+                    'https://bugzilla.mozilla.org/show_bug.cgi?id=777028');
             }
-        } else {
             maxTexturesUnits = Math.min(16, maxTexturesUnits);
         }
 
+        gl.deleteProgram(program);
+        gl.deleteShader(glVertexShader);
+        gl.deleteShader(glFragmentShader);
         logDepthBufferSupported = renderer.capabilities.logarithmicDepthBuffer;
     },
 };
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -3,6 +3,7 @@ import { Scene, EventDispatcher, Vector2 } from 'three';
 import Camera from '../Renderer/Camera';
 import MainLoop from './MainLoop';
 import c3DEngine from '../Renderer/c3DEngine';
+import { getMaxTextureSamplerCount } from '../Renderer/LayeredMaterial';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from './Layer/LayerUpdateStrategy';
 import { GeometryLayer, Layer, defineLayerProperty } from './Layer/Layer';
 import Scheduler from './Scheduler/Scheduler';
@@ -305,7 +306,22 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
     }
     layer = _preprocessLayer(this, layer, provider, parentLayer);
     if (parentLayer) {
-        parentLayer.attach(layer);
+        if (layer.type == 'color') {
+            const colorLayers = this.getLayers(l => l.type === 'color');
+            colorLayers.push(layer);
+
+            const sumColorLayers = parentLayer.countTextureSamplersForColorLayers ?
+                parentLayer.countTextureSamplersForColorLayers(colorLayers) :
+                colorLayers.length;
+
+            if (sumColorLayers <= getMaxTextureSamplerCount()) {
+                parentLayer.attach(layer);
+            } else {
+                throw new Error('Cant add color layer: the maximum layer is reached');
+            }
+        } else {
+            parentLayer.attach(layer);
+        }
     } else {
         if (typeof (layer.update) !== 'function') {
             throw new Error('Cant add GeometryLayer: missing a update function');
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -317,7 +317,7 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
             if (sumColorLayers <= getMaxTextureSamplerCount()) {
                 parentLayer.attach(layer);
             } else {
-                throw new Error('Cant add color layer: the maximum layer is reached');
+                throw new Error(`Cant add color layer ${layer.id}: the maximum layer is reached`);
             }
         } else {
             parentLayer.attach(layer);
--- a/src/Renderer/LayeredMaterial.js
+++ b/src/Renderer/LayeredMaterial.js
@@ -77,17 +77,9 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
 const arrayColorLength = 15;
 const nbSamplersElevation = 1;
 
-function getTextureSamplerCount() {
-    const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
-    const nbSamplersColor = Math.min(maxTexturesUnits - nbSamplersElevation);
-    const sizeColor_01 = nbSamplersColor > arrayColorLength ? arrayColorLength : nbSamplersColor;
-    const sizeColor_02 = nbSamplersColor > arrayColorLength ? arrayColorLength : 0;
-    return { sizeColor_01, sizeColor_02 };
-}
-
 export function getMaxTextureSamplerCount() {
-    const { sizeColor_01, sizeColor_02 } = getTextureSamplerCount();
-    return sizeColor_01 + sizeColor_02;
+    const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
+    return Math.min(maxTexturesUnits - nbSamplersElevation, arrayColorLength);
 }
 
 const LayeredMaterial = function LayeredMaterial(options) {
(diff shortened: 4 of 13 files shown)


src/Renderer/LayeredMaterial.js
```diff
@@ -74,14 +74,46 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
...
+export function getMaxColorLayerCount() {
+    const { sizeColor_01, sizeColor_02 } = getCountColorLayer();
+    return sizeColor_01 + sizeColor_02;
+}
+
 const LayeredMaterial = function LayeredMaterial(options) {
     THREE.RawShaderMaterial.call(this);
 
-    const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
-    const nbSamplers = Math.min(maxTexturesUnits - 1, 16 - 1);
```
autra, line 81: Unfortunately we cannot remove the `16-1` part yet, at least not on linux, not until https://bugzilla.mozilla.org/show_bug.cgi?id=777028 is fixed... :-/

autra: Oh forget this, you still have it on `arrayColorLength` I guess.

peppsac:
I'm not a big fan of the changes proposed :(

IMHO using #309 as a starting point to reduce the number of used samplers would have been a better idea.
Then we could have improved it even further, by not tying canvas allocation and texture downloads. This way we could efficiently use GPU memory without pushing the complexity on the shaders.

src/Renderer/LayeredMaterial.js
gchoqueux: is with Mesa?

gchoqueux:
@peppsac 
Texture atlas is good way to reduce  the number of used samplers
This PR isn't to reduce  the number of used samplers, it's to be able to use all the samplers.
the #309 doesn't resolve this problem, isn't it?
it would take a combination of both, ultimately

src/Renderer/LayeredMaterial.js
autra: Actually, I was right: on max, you can have `arrayColorLength * 2` samplers. So bug above applies.

peppsac:
>  Texture atlas is good way to reduce the number of used samplers

Yes. Or to put it differently: to not waste samplers.

>  This PR isn't to reduce the number of used samplers, it's to be able to use all the samplers.

Sure. But IMHO it's way too convoluted, hence my suggestion to first simplify the code and then extend it.

src/Renderer/LayeredMaterial.js
gchoqueux: no, I don't think so
[because we test if we are in mesa driver](https://github.com/iTowns/itowns/pull/583/files#diff-dd48a35666ea82b037874482f9b61b60R23)

autra: Cool!

gchoqueux: but in fact it doesn't work
I am looking for a way to detect if mesa is present

peppsac:
Another bothering issue with this change: you're adding a new layer of complexity that will only benefits to a portion of our users.
According to http://feedback.wildfiregames.com/report/opengl/feature/GL_MAX_TEXTURE_IMAGE_UNITS_ARB only 35% of users might have more than 16 active textures at a time - and keep in mind that this are statistics from a game so they may be skewed towards high-end graphics card.

And with all these changes, the total layer count will still be limited: even on an expensive GPU you won't be able draw all the Frecnh departments as geojson because that would require 101 layers...

So IMHO, we'd better focus on how to support an unlimited amout of color layers rather than adding code to just push the limit a bit further (and only for a fraction of users).

nosy-b:
I think the first commit was needed: Avoiding the app to crash by not allowing to add too much layers (controling max textures per shader).

In real life we are talking about the number of textures per tile mesh which in the worst case we met  (projection between geometry tiles and textures tiles differs) is 3, meaning a mix of 5 different color layers per tileMesh which is for 99.9% of the users enough.

It is the job of the High level API to reorganize the layers if the user wants to add more than 5 color layer  by for example deactivating the rendering of the oldest one added. -> Not a job of itowns itself.

Clearly in the best world it would be better to handle an unlimited number of color layers together which would pass by texture atlas when we'll be able to make it as fast and light as our actual simple rendering architecture.

peppsac:
@nosy-b sure the first commit implements a useful check. My comment is about the `SECOND_SAMPLER` changes.

nosy-b:
@peppsac I don't feel that it looks much more complicated than before, we still got vec4 getColorAtIdUv(int id, vec2 uv)` and it looks quite transparent for most users.

Sure the architecture is not very elegant but it's mainly due to the openGL ES constraint: constant for loop and max number of _if_ statement. 

I join you on the fact that it might look like a 'lot' of code for a fraction of users so I'm not crazy about it but we can certainly imagine that the number of users with 32 texture samples available will grow in the future.

Another argument toward texture atlas ( or even mega texture :) ) is that we now make the assumption that we always have a single texture sample for the elevation but for some use cases this might not be enough...

peppsac:
Then why not use something simpler that 2 manually unrolled for-loop?
The following code works for TEX_UNITS = 32 and is actually a simplification over the current iTowns code:
```glsl
vec4 colorAtIdUv(sampler2D dTextures[TEX_UNITS],vec3 offsetScale[TEX_UNITS],int id, vec2 uv){
    for (int i=0; i<TEX_UNITS; i++) {
        if (id == i) {
            return texture2D(dTextures[i],  pitUV(uv,offsetScale[i]));
        }
    }
    return vec4(0.0, 0.0, 0.0, 0.0);
}
```

gchoqueux:
Some navigators doesn't support this type of for-loop (like IE)
In my memories, one or two years ago firefox or chrome didn't support it either.
We can't take the risk.

**Warning that appears with the addition of this method**
**_Firefox 52 / ubuntu 16.0_**
   * loop only executes for 1 iteration(s), consider removing [loop]

**_Google Chrome 62/ ubuntu 16.0_**
  * Can't unroll loops marked with loop attribute
  * Sampler array index must be a literal expression, forcing loop to unroll
	
**IE 12/ Windows 7**
 * error: shader can't be linked

zarov:
Unfortunately, none of these versions are working with Mesa on Firefox with Ubuntu. The only working version is with the fallback from 32 to 16 [here](https://github.com/iTowns/itowns/pull/583/files#diff-dd48a35666ea82b037874482f9b61b60R36) **IF** we remove the condition on `/Sampler limit exceeded/`.

gchoqueux:
@zarov I remove `/Sampler limit exceeded`/
But the roll for-loop, doens't work on Mesa/Ubuntu, because the index `i` isn't constant

src/Renderer/LayeredMaterial.js
gchoqueux: I fixe the firefox / mesa bug
unfortunately, the last test commit with roll loop doens't work in mesa

gchoqueux:
branch updated

src/Core/System/Capabilities.js
```diff
@@ -1,11 +1,33 @@
...
+
+    gl.shaderSource(shader, string);
+    gl.compileShader(shader);
+    return shader;
+}
+
+function isFirefox() {
+    if (navigator && navigator.userAgent) {
+        const isFfox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+        return isFfox;
```
zarov: Why not put directly `return navigator.userAgent.toLowerCase().indexOf('firefox') > -1;`, or even `return navigator && navigator.userAgent && navigator.userAgent.toLowerCase().indexOf('firefox') > -1;` for the whole function ?

src/Renderer/LayeredMaterial.js
```diff
@@ -345,14 +398,13 @@ LayeredMaterial.prototype.setTexturesLayer = function setTexturesLayer(textures,
...
 };
 
-LayeredMaterial.prototype.setTexture = function setTexture(texture, layerType, slot, offsetScale) {
-    if (this.textures[layerType][slot] === undefined || this.textures[layerType][slot].image === undefined) {
+LayeredMaterial.prototype.setTexture = function setTexture(texture = emptyTexture, layerType, slot, offsetScale = new THREE.Vector4(0.0, 0.0, 1.0, 1.0)) {
+    if (this.getTextureByIndex(layerType, slot) === undefined || this.getTextureByIndex(layerType, slot).image === undefined) {
         this.loadedTexturesCount[layerType] += 1;
     }
 
     // BEWARE: array [] -> size: 0; array [10]="wao" -> size: 11
```
zarov: Can be removed

src/Renderer/Shader/TileFS.glsl
```diff
@@ -12,13 +12,17 @@ const vec4 COrange = vec4( 1.0, 0.3, 0.0, 1.0);
 const vec4 CRed = vec4( 1.0, 0.0, 0.0, 1.0);
 
 
-uniform sampler2D   dTextures_01[TEX_UNITS];
-uniform vec4        offsetScale_L01[TEX_UNITS];
+uniform sampler2D   texColor_01[TEX_UNITS];
+uniform vec4        offsetScale_color_01[TEX_UNITS];
+
+#if  defined(SECOND_SAMPLER)
```
zarov: Space

zarov (commented):
Looks good to me overall.

peppsac (dismissed):
Still not convinced we want to add so much complexity, just for being able to add more color layers on few machines...
(and this PR will break panorama rendering because of missing `sumOccupancyColorLayers` function)

peppsac:
There are code issues in the PR as well - but it's irrelevant for now: I don't like the solution proposed (for reasons already explained here: https://github.com/iTowns/itowns/pull/583#issuecomment-345948070)

gchoqueux:
@peppsac 
* the addition of a sampler array adds little complexity. In addition there are getter and setter to make the use of textures completely transparent.

* 35% is a lot, it justifies to do it

* Indeed we all like to have the possibility to add an infinity of layer (within the limits of the capacity of the gpu). I will work on it at another time.

This PR doesn't add any bug or regression. So there's no real interest to block the PR

src/Core/Prefab/GlobeView.js
```diff
@@ -163,6 +163,18 @@ export function createGlobeLayer(id, options) {
         return false;
     }
 
+    wgs84TileLayer.sumOccupancyColorLayers = (layerColors, layer) => {
```
peppsac: `wgs84TileLayer.sumOccupancyColorLayers = (colorLayers, layer) => {` is more correct.

src/Core/Prefab/GlobeView.js
```diff
@@ -163,6 +163,18 @@ export function createGlobeLayer(id, options) {
...
+        layerColors.push(layer);
```
peppsac: Better avoid modifying arguments.
Change the signature of the function to `colorLayers` and let the caller passes the correct layer array.

src/Core/Prefab/GlobeView.js
peppsac: And an even better name would be: `countTextureSamplersForColorLayers`.
It's longer but it really describe what this function is doing.

Btw, this function (or its usage) isn't documented anywhere...

src/Core/Prefab/GlobeView.js
```diff
@@ -341,6 +353,8 @@ GlobeView.prototype.removeLayer = function removeImageryLayer(layerId) {
             layerId,
         });
 
+        this.wgs84TileLayer.colorLayersCount -= this.wgs84TileLayer.getIncreColorlayer(layer);
```
peppsac: What is `getIncreColorlayer`?

src/Renderer/LayeredMaterial.js
```diff
@@ -74,14 +74,46 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
 };
 /* eslint-enable */
 
+const arrayColorLength = 15;
+const nbSamplersElevation = 1;
+
+function getCountColorLayer() {
```
peppsac: `getColorLayerCount`

peppsac: Or actually `getTextureSamplerCount`

src/Renderer/LayeredMaterial.js
```diff
@@ -74,14 +74,46 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
...
+    const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
+    const nbSamplersColor = Math.min(maxTexturesUnits - nbSamplersElevation);
+    const sizeColor_01 = nbSamplersColor > arrayColorLength ? arrayColorLength : nbSamplersColor;
+    const sizeColor_02 = nbSamplersColor > arrayColorLength ? arrayColorLength : 0;
+    return { sizeColor_01, sizeColor_02 };
+}
+
+export function getMaxColorLayerCount() {
```
peppsac: `getMaxTextureSamplerCount`?

src/Renderer/LayeredMaterial.js
```diff
@@ -74,14 +74,46 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
...
+}
+
 const LayeredMaterial = function LayeredMaterial(options) {
     THREE.RawShaderMaterial.call(this);
 
-    const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
-    const nbSamplers = Math.min(maxTexturesUnits - 1, 16 - 1);
+    this.defines = {};
+
+    // const maxTexturesUnits = Capabilities.getMaxTextureUnitsCount();
```
peppsac: Remove dead code please.

src/Renderer/LayeredMaterial.js
```diff
@@ -74,14 +74,46 @@ var moveElementsArraySafe = function moveElementsArraySafe(array,index, howMany,
...
+    // Why same size?
+    //      If there was 2 arrays of 15 and 16 samplers
+    //      so it would require two functions getColorAtIdUv with signatures of different parameters
+    const { sizeColor_01, sizeColor_02 } = getCountColorLayer();
+    const maxColorLayerCount = sizeColor_01 + sizeColor_02;
+
     this.vertexShader = TileVS;
 
-    this.fragmentShaderHeader = `${PrecisionQualifier}\nconst int   TEX_UNITS   = ${nbSamplers.toString()};\n`;
+    this.fragmentShaderHeader = `${PrecisionQualifier}\n`;
```
peppsac: You could use multi-line syntax instead.

peppsac (commented):
Removing my `Request Changes` tag as I'm the only one annoyed with those changes.

(I wrote a quick review nonetheless but only about cosmetic changes)

gchoqueux:
fix #600

zarov:
@peppsac did you want to remove your request for change ? It seems it's still here and locking this PR.

gchoqueux:
[review](https://github.com/iTowns/itowns/pull/583#pullrequestreview-82833695) has been taken into account

gchoqueux:
@iTowns/reviewers, the review has been taken into account

gchoqueux:
we'll be able to merge this PR?

gchoqueux:
@iTowns/reviewers this PR will be merged 23 January, if there's no request

peppsac:
> @iTowns/reviewers this PR will be merged 23 January, if there's no request

Don't forget that every PR needs a review + approval.
I dismissed my review because I disagree with the ideas of this PR but that's different from an approval :)

So you still need to wait for one of the reviewers to perform a review.

gchoqueux:
@peppsac thank you for finally answering
Don't worry, I haven't forgotten review + approval
I also asked if you are satisfied with the [cosmetic changes](https://github.com/iTowns/itowns/pull/583#pullrequestreview-82833695)

autra:
> this PR will be merged 23 January, if there's no request

So you meant: "if there is a review", I guess ;-)

src/Core/View.js
```diff
@@ -301,7 +302,22 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+            colorLayers.push(layer);
+
+            const sumColorLayers = parentLayer.countTextureSamplersForColorLayers ?
+                parentLayer.countTextureSamplersForColorLayers(colorLayers) :
+                colorLayers.length;
+
+            if (sumColorLayers <= getMaxTextureSamplerCount()) {
+                parentLayer.attach(layer);
+            } else {
+                throw new Error('Cant add color layer: the maximum layer is reached');
```
zarov: This is causing an error in `index.html` with the `menuGlobe` not being completed. Please adapt the example before merging.

zarov (commented):
I'm not a fan of this solution, but if it means more layers for now until we implement atlases, I'm all for it. Next step is to make up-to-date #309 imo

src/Core/View.js
gchoqueux: Ok PR Updated

gchoqueux:
PR updated remove double array color texture

src/Renderer/LayeredMaterial.js
```diff
@@ -398,6 +417,57 @@ LayeredMaterial.prototype.getTextureCountByLayerIndex = function getTextureCount
     return this.layerTexturesCount[index];
 };
 
+LayeredMaterial.prototype.getTextureByIndex = function getTextureByIndex(layerType, index) {
```
peppsac, line 420: `getTextureByIndex(l_COLOR, ...)` and `getTextureColorByIndex` are redundant from an API point of view.

`getTextureColorByIndex` shouldn't be exposed since its only a utility function.

And the same reasonning applies to all the new functions you're adding to LayeradMaterial: does it make sense to expose them?

src/Renderer/Shader/TileFS.glsl
```diff
@@ -12,13 +12,12 @@ const vec4 COrange = vec4( 1.0, 0.3, 0.0, 1.0);
 const vec4 CRed = vec4( 1.0, 0.0, 0.0, 1.0);
 
 
-uniform sampler2D   dTextures_01[TEX_UNITS];
-uniform vec4        offsetScale_L01[TEX_UNITS];
+uniform sampler2D   texColor_01[TEX_UNITS];
```
peppsac, line 15: Good idea to rename all the variables to better names.

This could be improved though:
- texColor_01 -> colorTextures
- offsetScale_color_01 -> colorOffsetScale or colorTexturesOffsetScale
- paramLayers -> layerParameters or layerParams
- texElevation[1] -> elevationTexture
- offsetScale_elevation -> elevationTextureOffsetScale

And this change alone could be moved to a cleanup PR.

peppsac (changes requested):
I just reviewed the renaming part.

And your last commit removes the initial idea of this PR: add a 2nd array of textures. Can you explain a bit more on this change?

gchoqueux:
splited on 2 PRs : #782 and #784

PR #584, opened 2017-11-17
fix (core): blending with premultiplied color to transparent layer

gchoqueux:
## Description
add full support opacity in texture and in fragment

## Motivation and Context
Remove colour artifact on border 

before fix, we can see white color on border line
png2

after fix
png

### DIFF
--- a/src/Core/Scheduler/Providers/WMTS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMTS_Provider.js
@@ -109,6 +109,9 @@ WMTS_Provider.prototype.getColorTexture = function getColorTexture(coordWMTS, la
         result.texture = texture;
         result.texture.coords = coordWMTS;
         result.pitch = new THREE.Vector4(0, 0, 1, 1);
+        if (layer.transparent) {
+            texture.premultiplyAlpha = true;
+        }
 
         return result;
     });
--- a/src/Renderer/Shader/TileFS.glsl
+++ b/src/Renderer/Shader/TileFS.glsl
@@ -36,6 +36,19 @@ varying vec3        vNormal;
 
 uniform float opacity;
 
+vec4 applyWhiteToInvisibleEffect(vec4 color, float intensity) {
+    float a = (color.r + color.g + color.b) * 0.333333333;
+    color.a *= 1.0 - pow(abs(a), intensity);
+    return color;
+}
+
+vec4 applyLightColorToInvisibleEffect(vec4 color, float intensity) {
+    float a = max(0.05,1.0 - length(color.xyz - CWhite.xyz));
+    color.a *= 1.0 - pow(abs(a), intensity);
+    color.rgb *= color.rgb * color.rgb;
+    return color;
+}
+
 #if defined(DEBUG)
     uniform bool showOutline;
     const float sLine = 0.008;
@@ -82,7 +95,7 @@ void main() {
 
         float fogIntensity = 1.0/(exp(depth/distanceFog));
 
-        vec4 diffuseColor = CWhite;
+        vec4 diffuseColor = vec4(noTextureColor, 1.0);
         bool validTexture = false;
 
         // TODO Optimisation des uv1 peuvent copier pas lignes!!
@@ -122,21 +135,24 @@ void main() {
                             textureIndex,
                             projWGS84 ? vUv_WGS84 : uvPM);
 
-                        if (layerColor.a > 0.0) {
+                        if (layerColor.a > 0.0 && paramsA.w > 0.0) {
                             validTexture = true;
-                            float lum = 1.0;
-
-                            if(paramsA.z > 0.0) {
-                                float a = max(0.05,1.0 - length(layerColor.xyz-CWhite.xyz));
-                                if(paramsA.z > 2.0) {
-                                    a = (layerColor.r + layerColor.g + layerColor.b)*0.333333333;
-                                    layerColor*= layerColor*layerColor;
-                                }
-                                lum = 1.0-pow(abs(a),paramsA.z);
+                            if(paramsA.z > 2.0) {
+                                layerColor.rgb /= layerColor.a;
+                                layerColor = applyLightColorToInvisibleEffect(layerColor, paramsA.z);
+                                layerColor.rgb *= layerColor.a;
+                            } else if(paramsA.z > 0.0) {
+                                layerColor.rgb /= layerColor.a;
+                                layerColor = applyWhiteToInvisibleEffect(layerColor, paramsA.z);
+                                layerColor.rgb *= layerColor.a;
                             }
 
-                            diffuseColor = mix( diffuseColor,layerColor, lum*paramsA.w * layerColor.a);
-
+                            // Use premultiplied-alpha blending formula because source textures are either:
+                            //     - fully opaque (layer.transparent = false)
+                            //     - or use premultiplied alpha (texture.premultiplyAlpha = true)
+                            // Note: using material.premultipliedAlpha doesn't make sense since we're manually blending
+                            // the multiple colors in the shader.
+                            diffuseColor = diffuseColor * (1.0 - layerColor.a * paramsA.w) + layerColor * paramsA.w;
                         }
                     }
                 }
--- a/examples/globe_vector.js
+++ b/examples/globe_vector.js
@@ -29,6 +29,7 @@ promises.push(globeView.addLayer({
     protocol: 'rasterizer',
     id: 'Kml',
     name: 'kml',
+    transparent: true,
 }));
 
 promises.push(globeView.addLayer({
@@ -37,6 +38,7 @@ promises.push(globeView.addLayer({
     protocol: 'rasterizer',
     id: 'Gpx',
     name: 'Ultra 2009',
+    transparent: true,
 }));
 
 promises.push(globeView.addLayer({
@@ -44,6 +46,7 @@ promises.push(globeView.addLayer({
     url: 'https://raw.githubusercontent.com/gregoiredavid/france-geojson/master/departements/09-ariege/departement-09-ariege.geojson',
     protocol: 'rasterizer',
     id: 'ariege',
+    transparent: true,
     style: {
         fill: 'orange',
         fillOpacity: 0.5,
--- a/examples/planar_vector.js
+++ b/examples/planar_vector.js
@@ -32,6 +32,7 @@ view.addLayer({
     id: 'wms_imagery',
     name: 'Ortho2009_vue_ensemble_16cm_CC46',
     projection: 'EPSG:3946',
+    transparent: true,
     options: {
         mimetype: 'image/jpeg',
     },
@@ -59,6 +60,7 @@ view.addLayer({
     protocol: 'rasterizer',
     id: 'Kml',
     extent: extent,
+    transparent: true,
     options: { zoom: { min: 0, max: 6 } },
 });
 
@@ -68,6 +70,7 @@ view.addLayer({
     protocol: 'rasterizer',
     options: { zoom: { min: 0, max: 6 } },
     id: 'gpx',
+    transparent: true,
     style: {
         stroke: 'blue',
     },
@@ -80,6 +83,7 @@ view.addLayer({
     projection: 'EPSG:3946',
     id: 'geo',
     extent: extent,
+    transparent: true,
     options: { zoom: { min: 0, max: 6 } },
     style: {
         fill: 'orange',
--- a/src/Core/Scheduler/Providers/Raster_Provider.js
+++ b/src/Core/Scheduler/Providers/Raster_Provider.js
@@ -57,6 +57,10 @@ function createTextureFromVector(tile, layer) {
         result.texture.extent = tile.extent;
         result.texture.coords = coords;
         result.texture.coords.zoom = tile.level;
+
+        if (layer.transparent) {
+            result.texture.premultiplyAlpha = true;
+        }
         return Promise.resolve(result);
     } else {
         return Promise.resolve();
--- a/src/Core/Scheduler/Providers/StaticProvider.js
+++ b/src/Core/Scheduler/Providers/StaticProvider.js
@@ -59,7 +59,9 @@ function getTexture(tile, layer) {
         }
         // TODO: modify TileFS to handle tiles with ratio != image's ratio
         result.pitch = tile.extent.offsetToParent(selection.extent);
-
+        if (layer.transparent) {
+            texture.premultiplyAlpha = true;
+        }
 
         return result;
     });
--- a/src/Core/Scheduler/Providers/TMS_Provider.js
+++ b/src/Core/Scheduler/Providers/TMS_Provider.js
@@ -46,6 +46,9 @@ TMS_Provider.prototype.executeCommand = function executeCommand(command) {
         result.pitch = coordTMSParent ?
             coordTMS.offsetToParent(coordTMSParent) :
             new THREE.Vector4(0, 0, 1, 1);
+        if (layer.transparent) {
+            texture.premultiplyAlpha = true;
+        }
         return result;
     });
 };
--- a/src/Core/Scheduler/Providers/WMS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMS_Provider.js
@@ -126,6 +126,9 @@ WMS_Provider.prototype.getColorTexture = function getColorTexture(tile, layer, t
     return OGCWebServiceHelper.getColorTextureByUrl(url, layer.networkOptions).then((texture) => {
         result.texture = texture;
         result.texture.extent = extent;
+        if (layer.transparent) {
+            texture.premultiplyAlpha = true;
+        }
         if (tileCoords) {
             result.texture.coords = tileCoords;
         } else {


src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -113,6 +113,9 @@ WMTS_Provider.prototype.getColorTexture = function getColorTexture(coordWMTS, la
     return OGCWebServiceHelper.getColorTextureByUrl(url, layer.networkOptions).then((texture) => {
         const result = {};
         result.texture = texture;
+        if (layer.options.mimetype == 'image/png') {
+            texture.premultiplyAlpha = true;
```
peppsac: This is wrong IMHO: you can't deduce `premultiplyAlpha` solely based on the mimetype.

examples/layers/JSONLayers/Cada.json
```diff
@@ -0,0 +1,165 @@
+{
```
peppsac: Why do you add a new example layer in a `fix` PR?

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,22 @@ void main() {
                             textureIndex,
                             projWGS84 ? vUv_WGS84 : uvPM);
 
-                        if (layerColor.a > 0.0) {
+                        // (!validTexture) test : force to have a white background for mixing color layers
```
peppsac: IMHO, instead of adding more ad-hoc code to TileFS (ie: customized rendering based on a float value in paramsA) you should instead expose a feature allowing the user to insert their custom glsl code here to modify diffuseColor for a given layer.

peppsac (changes requested):

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: what would you expect?

examples/layers/JSONLayers/Cada.json
gchoqueux: why not

src/Renderer/Shader/TileFS.glsl
gchoqueux: it's an existing code fix not an improvement.
it takes another PR

src/Core/Scheduler/Providers/WMTS_Provider.js
peppsac: You can't guess the value, hence why three default to false.
`premultiplyAlpha` means that the data has been prepared in a specific way - which is quite unlikely in GIS data.

examples/layers/JSONLayers/Cada.json
peppsac: Seriously :-/ ?

src/Renderer/Shader/TileFS.glsl
peppsac: Hmmm I'm pretty sure you're adding new code ( `} else if(paramsA.z > 0.0) {` ...)...

gchoqueux: No, I do not think so

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: the easiest way is to add it as an option to the layer

peppsac: Yes.
Is there a service that would serve premultiplied-alpha image though?

src/Renderer/Shader/TileFS.glsl
peppsac: Ok. I realized that you duplicated the code that was shared before.
I don't understand your comments :
- "// separation of states to avoid errors in final statement" ?
- "// (!validTexture) test : force to have a white background for mixing color layers" ?

src/Core/Scheduler/Providers/WMTS_Provider.js
peppsac: So back on this:
- texture.premultiplyAlpha will simply ask WebGL to perform color.rgb *= color.a when loading this texture.
- material.premultipliedAlpha (which you're not using) changes the blending equation to account for premultiplied alpha texture

So if you use only premultiplyAlpha without premultipliedAlpha the rgb channels are going to be multiply twice by alpha:
```
   pixel_color = premultiplied_color * a + pixel_color * (1 - a)
   pixel_color = (original_texture_color * a) * a + pixel_color * (1 - a)
```

gchoqueux: I will remove the `premultiplyAlpha` because the problem doesn't come from that.
white borders come from Linear Filter


png2

peppsac: Filtering and premultiplied-alpha are related topics.
Any filtering applied on a texture with transparent pixels (except NEAREST which introduces ugly aliasing) will cause problem - that's why premultiplied alpha has been created in the first place...

src/Renderer/Shader/TileFS.glsl
gchoqueux: I remove `// separation of states to avoid errors in final statement`

and change test `(!validTexture)` to `(layerColor.a == 0.0 && !validTexture)` because if first texture is transparent, we need to force blending with CWhite

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: I change the blending in : 
```js
diffuseColor.rgb = (layerColor.rgb * layerColor.a) + (diffuseColor.rgb * (1.0 - layerColor.a));
```

peppsac: Changing the blending in the shader is a different issue.

As long as you don't use premultipliedAlpha + premultiplyAlpha you'll get blended pixels color when reading the texture from the shader.

src/Renderer/Shader/TileFS.glsl
peppsac: > I remove // separation of states to avoid errors in final statement

Then please explain why you need to duplicate the code logic.

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: `material.premultipliedAlpha` is to materials blending.
Here, we are talking about  color layers blending.

peppsac: >  Here, we are talking about color layers blending.

I don't think you can really separate the 2 issues.
Anyway, what is your change fixing?

You're replacing:
```glsl
diffuseColor = mix( diffuseColor,layerColor, lum*paramsA.w * layerColor.a);
```
With the equivalent code:
```
layerColor.a *= paramsA.w;
layerColor.a *= lum;
diffuseColor.rgb = (layerColor.rgb * layerColor.a) + (diffuseColor.rgb * (1.0 - layerColor.a));
```

So I don't really expect this to fix anything..

gchoqueux: oh yes I didn't have to push the trick

gchoqueux:
@peppsac @zarov PR updated

zarov (commented):
Looks good to me, I see no problem to add a new layer here, but you need to use it in an example to show the fixed problem here.

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,25 @@ void main() {
                             textureIndex,
                             projWGS84 ? vUv_WGS84 : uvPM);
 
-                        if (layerColor.a > 0.0) {
+                        bool isNotTransparent = layerColor.a > 0.0;
```
peppsac: `isNotTransparent` means `isVisible` right?

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,25 @@ void main() {
...
-
-                            if(paramsA.z > 0.0) {
-                                float a = max(0.05,1.0 - length(layerColor.xyz-CWhite.xyz));
-                                if(paramsA.z > 2.0) {
-                                    a = (layerColor.r + layerColor.g + layerColor.b)*0.333333333;
-                                    layerColor*= layerColor*layerColor;
-                                }
-                                lum = 1.0-pow(abs(a),paramsA.z);
+                            // unassociate the alpha (because id premultiply alpha texture)
+                            if (isNotTransparent) {
```
peppsac: What's the point of hacking the premultiplied alpha flow instead of using it completely (by using material.premultipliedAlpha)?

(at the very least you must add a comment explaining why you don't make this obvious choice)

peppsac (changes requested):
> I see no problem to add a new layer here, but you need to use it in an example to show the fixed problem here.

Right, and this would be a new PR: "feat (examples): add a new example blablabla".

Splitting PR is about easing the reviews and the revert where needed.

zarov:
Ok, I'm making an example with this dataset. @gchoqueux please remove it from your PR.

> Splitting PR is about easing the reviews and the revert where needed.

 I agree splitting makes PR easier to review, but PR have nothing to do with reverting, commits do.

gchoqueux:
@zarov example removed

gchoqueux:
@iTowns/reviewers ?

peppsac:
@gchoqueux you didn't answer the questions from my last review...

src/Renderer/Shader/TileFS.glsl
gchoqueux: We use premultiplied alpha for mixing color layers
We don't use material.premultipliedAlpha which is used to blend the materials

src/Renderer/Shader/TileFS.glsl
gchoqueux: yes i changed to `isVisible`

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,27 @@ void main() {
                             textureIndex,
                             projWGS84 ? vUv_WGS84 : uvPM);
 
-                        if (layerColor.a > 0.0) {
+                        bool isVisible = layerColor.a > 0.0;
+
+                        // if first texture is transparent, we need to force blending with CWhite
+                        if (isVisible || (!isVisible && !validTexture)) {
```
peppsac: Can be simplified to `isVisible || !validTexture`.
But it still doesn't make sense sorry:
- your comment mentions `CWhite` but it's only used if `paramsA.z > 2.0`?
- why do you do anything if layerColor.a is 0?
- if the problem occurs when blending a non-opaque texture with no previous texture, maybe a simple fix would be to do no blending at all? (eg: diffuseColor = layerColor)

peppsac (changes requested):

src/Renderer/Shader/TileFS.glsl
peppsac: You still didn't answer my question: why don't you use `material.premultipliedAlpha`?

gchoqueux: We don't use `material.premultipliedAlpha` which is used to blend materials/mesh in rendering buffer with `blendFunc`.
Here the blending takes place in the shader, and we can't do it with the webgl functions (blendFunc...)

peppsac: `̀// We use premultiplied alpha for mixing color layers` is wrong since you demultiply the color before the blending. 
Premultiply-alpha is only used for texture fetching here and avoid gl_linear filtering artifact.

Regarding `material.premultipliedAlpha` let's not use it for now.

gchoqueux:
We agree that this can be improved. but it's okay for now, it should add a comment and a issue

src/Renderer/Shader/TileFS.glsl
gchoqueux: yes, I change the comment

src/Renderer/Shader/TileFS.glsl
gchoqueux: > your comment mentions CWhite but it's only used if paramsA.z > 2.0?

it's to force the blending with the background first background color (CWhite)
I change comment

> why do you do anything if layerColor.a is 0?

because the `mix` function doesn't make any changes

> if the problem occurs when blending a non-opaque texture with no previous texture, maybe a simple fix would be to do no blending at all? (eg: diffuseColor = layerColor)

but what should be the background color if the opacity (texture or layer) is less than 1.0?

peppsac: > but what should be the background color if the opacity (texture or layer) is less than 1.0?

You have N layers and you want to blend them together, so my guess is the algorithm is supposed to look like this:

The initial color can be either:
  - `noTextureColor`
  - `vec4(0, 0, 0, 0)` (which thanks to premultiplied alpha blending won't affect the final result)

Then for each layer: 
   layer_color = colorAtIdUv(...);
   // premultiplied alpha blending
   diffuseColor = diffuseColor * (1 - layer_color.a * paramsA.w) + layer_color * paramsA.w)

Does this answer your question?

gchoqueux: in your proposal
`noTextureColor` is equal to `vec4(0, 0, 0, 0)`?

gchoqueux: Yes, exactly. 
I separated , unassociated the alpha `(layerColor.rgb /= layerColor.a)`, to use `mix` function.
I use your equivalent formulation.

The background color is `noTextureColor`, the user assume the background colour.

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,22 @@ void main() {
...
+                                layerColor.rgb *= layerColor.rgb * layerColor.rgb;
+                                layerColor.a *= 1.0 - pow(abs(a), paramsA.z);
+                            } else if(paramsA.z > 0.0) {
+                                float a = (layerColor.r + layerColor.g + layerColor.b) * 0.333333333;
+                                layerColor.a *= 1.0 - pow(abs(a), paramsA.z);
                             }
 
-                            diffuseColor = mix( diffuseColor,layerColor, lum*paramsA.w * layerColor.a);
-
+                            // In formula, Premultiply-alpha is used to avoid gl_linear filtering artifact.
```
peppsac: What about using this comment instead:
```js
// Use premultiplied-alpha blending formula because source textures are either:
//     - fully opaque (layer.transparent = false)
//     - or use premultiplied alpha (texture.premultiplyAlpha = true)
// Note: using material.premultipliedAlpha doesn't make sense since we're manually blending
// the multiple colors in the shader.
```

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,22 @@ void main() {
                             textureIndex,
                             projWGS84 ? vUv_WGS84 : uvPM);
 
-                        if (layerColor.a > 0.0) {
+                        bool isVisible = layerColor.a > 0.0;
+                        // if first texture is transparent, we need to force blending with first background color (CWhite)
+                        if ((isVisible && paramsA.w > 0.0) || !validTexture) {
```
peppsac: Are you sure you still need this `!validTexture` test now that `diffuseColor` is properly initialized to `noTextureColor`?

src/Renderer/Shader/TileFS.glsl
```diff
@@ -122,21 +122,22 @@ void main() {
...
-
-                            if(paramsA.z > 0.0) {
-                                float a = max(0.05,1.0 - length(layerColor.xyz-CWhite.xyz));
-                                if(paramsA.z > 2.0) {
-                                    a = (layerColor.r + layerColor.g + layerColor.b)*0.333333333;
-                                    layerColor*= layerColor*layerColor;
-                                }
-                                lum = 1.0-pow(abs(a),paramsA.z);
+                            if(paramsA.z > 2.0) {
+                                float a = max(0.05,1.0 - length(layerColor.xyz - CWhite.xyz));
```
peppsac: I'm not really sure of what is happening when paramsA.z is > 0.0 or > 2.0 but: are you sure that you shouldn't de-premultiply-alpha before tweaking the color?

Actually, you could do something like:
```glsl
if (paramsA.z > 2.0) {
    layerColor = applyXXXEffect(layerColor / layerColor.a);
} else if (paramsA.z > 0.0) {
    layerColor = applyYYYEffect(layerColor / layerColor.a);
}
```
to improve readability. WDYT?

src/Renderer/Shader/TileFS.glsl
gchoqueux: yes, I removed

src/Renderer/Shader/TileFS.glsl
gchoqueux: yes, we need to de-premultiply and re-premultiply

src/Renderer/Shader/TileFS.glsl
gchoqueux: ok

gchoqueux:
This PR is updated

peppsac (changes requested):
Looks good now, thanks!

peppsac (approved):

PR #585, opened 2017-11-17
feat(examples): add point features representing route points, in the …

gliegard:
Add point features in the WFS examples

## Description
feat(examples): add point features representing route points

## Motivation and Context
The goal is to show / verify point representation from WFS in an example.
I add theses points layer in the planar WFS example and in the globe WFS example.

## Screenshots
example_points

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -17,6 +17,8 @@ function addLayerCb(layer) {
 // Define projection that we will use (taken from https://epsg.io/3946, Proj4js section)
 itowns.proj4.defs('EPSG:3946',
     '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+itowns.proj4.defs('EPSG:2154',
+    '+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
 
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
@@ -29,7 +31,13 @@ promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(add
 promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
 
 function setMaterialLineWidth(result) {
-    result.children[0].material.linewidth = 5;
+    var i = 0;
+    var mesh;
+    for (; i < result.children.length; i++) {
+        mesh = result.children[i];
+
+        mesh.material.linewidth = 5;
+    }
 }
 
 function colorLine(properties) {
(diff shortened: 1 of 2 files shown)


examples/globe_wfs_extruded.js
```diff
@@ -104,5 +106,41 @@ globeView.addLayer({
...
 
+
+function setPointMaterial(result) {
+    var mesh = result.children[0];
+    mesh.material.size = 5;
+    mesh.material.sizeAttenuation = false;
+}
+
+function colorPoint(/* properties */) {
+    return new itowns.THREE.Color(0xff0000);
```
autra: what about 0x7F180D?

examples/globe_wfs_extruded.js
```diff
@@ -104,5 +106,41 @@ globeView.addLayer({
     },
 }, globeView.tileLayer);
 
+
+function setPointMaterial(result) {
```
autra: `configPointMaterial`?

examples/globe_wfs_extruded.js
```diff
@@ -104,5 +106,41 @@ globeView.addLayer({
...
+    var mesh = result.children[0];
```
autra: So you use an assumption about the data, but you could make this code works in all cases by doing a loop.

examples/globe_wfs_extruded.js
```diff
@@ -104,5 +106,41 @@ globeView.addLayer({
...
+
+function selectRoad(properties) {
+    return properties.gestion === 'CEREMA';
+}
+
+globeView.addLayer({
+    type: 'geometry',
+    update: itowns.FeatureProcessing.update,
+    convert: itowns.Feature2Mesh.convert({
+        altitude: 400,
```
autra, line 139: We can do better: call `DEMUtils.placeObjectOnGround` (or `getElevationValueAt` if you want to place each point individually) in onMeshCreated instead.

examples/wfs.js
```diff
@@ -144,4 +146,38 @@ view.addLayer({
...
+function setPointMaterial(result) {
+    var mesh = result.children[0];
+
+    mesh.material.size = 15;
+    mesh.material.sizeAttenuation = false;
+}
+
+function colorPoint(properties) {
+    if (properties.gestion === 'CEREMA') {
+        return new itowns.THREE.Color(0xff0000);
```
autra: may I suggest 0x7F180D  and 0xFFB300?

examples/wfs.js
```diff
@@ -144,4 +146,38 @@ view.addLayer({
     },
 }, view.tileLayer);
 
+function setPointMaterial(result) {
```
autra: configPointMaterial?

autra (approved):
Already good like it is. It would be extra nice to use DEMUtils to correctly place the points.

Have you considered to merge the layers with the same url? The callbacks would be a bit more complex, but these examples would be shorter and more readable, WDYT?

examples/globe_wfs_extruded.js
gliegard: DEMUtils.placeObjectOnGround seem to place the object under the ground, for this example.
And DEMUtils.getElevationValueAt use coordinate and I don't have the coordinate at this point.

Your suggestion is good, but I'll do that in another PR. 

In this other PR, maybe I will add and option in the Feature2Mesh convertion to use DEMUtils.getElevationValueAt to place the feature at the good altitude during the conversion.

examples/globe_wfs_extruded.js
gliegard: yep, good idea

examples/globe_wfs_extruded.js
gliegard: ok, I've changed the code for a loop.

examples/globe_wfs_extruded.js
peppsac: > DEMUtils.placeObjectOnGround seem to place the object under the ground, for this example.

Maybe you could take a few minutes to investigate why this happen (it could be a bug and you could open an issue)?

autra: > DEMUtils.placeObjectOnGround seem to place the object under the ground, for this example.

Maybe your other PR could be a bugfix ? ;-) I'm actually not even kidding: they shouldn't be placed under the ground with `placeObjectOnGround`.

Furthermore:

> In this other PR, maybe I will add and option in the Feature2Mesh convertion to use DEMUtils.getElevationValueAt to place the feature at the good altitude during the conversion.

onMeshCreated gives you the mesh, that gives you the geometry, which in turns contains coordinates. I don't think you need another option. 

>  I'll do that in another PR.

I've already approved your PR, so I'm not requesting it here, though I really feel that these floating points are... odd and we should do something about it soon.

autra: Or we can rename it `placeObjectUnderGround`, good naming is important :rofl:

gliegard: Floating points is better than no point for now.
I'll look at that later.
Can you merge it ?

| Or we can rename it placeObjectUnderGround, good naming is important :rofl:
Lol !!!!! Thanks for that joke !!!

autra:
Thanks!

nosy-b:
Ya, it's a new feature, not introducing bugs so that's always good move! But, if it shows a bug from previous functions it's worth creating the adequate issue! :)

gliegard:
Thanks, 
have a nice week-end :)

PR #586, opened 2017-11-17
Re-use identical tile's geometry and oriented the main normal on the Z axis

### LINKED ISSUE #558: Mesh from features should be in local coordinate
Mesh from features should be in local coordinate

## Context
Suggestion of an animation that scale buildings on Z axis
https://github.com/iTowns/itowns/pull/519#pullrequestreview-72184139

The animation works in the WFS planar example : We can see that the builldings is growing.
But it doesn't work in WFS globe example.

It's because feature is in global coordinates, and not in local coordinates.

## Possible Solution
Suggestion of a way to do it 
https://github.com/iTowns/itowns/pull/519#discussion_r147365812

gchoqueux:
## Description
Re-use identical tile's geometry and oriented the main normal on the Z axis
* The normal in the center of the geometry is aligned with the Z axis.
* Bounding boxes can use the Parent's worldMatrix (tile).

The geometry normal is computed with correct geodesic normal.
The bounding boxes is calculated by THREE.js.

## Motivation and Context
* reduce memory
* increase performance
* work tile's space, by example wfs (fix #558 issue)
* simplify obb computation
* reduce/fix obb calculation errors see #87 issues
* fix obb/culling for pano tile

**WARNING**
change subdivision count in PLANAR test, the bounding sphere are bigger, so the subdivision are changed

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -4,6 +4,8 @@
 // Define initial camera position
 var positionOnGlobe = { longitude: 4.818, latitude: 45.7354, altitude: 3000 };
 var promises = [];
+var meshes = [];
+var scaler;
 
 // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
 var viewerDiv = document.getElementById('viewerDiv');
@@ -95,6 +97,24 @@ function acceptFeature(properties) {
     return !!properties.hauteur;
 }
 
+scaler = function update(/* dt */) {
+    var i;
+    var mesh;
+    if (meshes.length) {
+        globeView.notifyChange(true);
+    }
+    for (i = 0; i < meshes.length; i++) {
+        mesh = meshes[i];
+        if (mesh.children.length) {
+            mesh.scale.z = Math.min(
+                1.0, mesh.scale.z + 0.016);
+            mesh.updateMatrixWorld(true);
+        }
+    }
+    meshes = meshes.filter(function filter(m) { return m.scale.z < 1; });
+};
+
+globeView.addFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, scaler);
 globeView.addLayer({
     type: 'geometry',
     update: itowns.FeatureProcessing.update,
@@ -102,6 +122,10 @@ globeView.addLayer({
         color: colorBuildings,
         altitude: altitudeBuildings,
         extrude: extrudeBuildings }),
+    onMeshCreated: function scaleZ(mesh) {
+        mesh.scale.z = 0.01;
+        meshes.push(mesh);
+    },
     filter: acceptFeature,
     url: 'http://wxs.ign.fr/72hpsel8j8nhb5qgdh07gcyp/geoportail/wfs?',
     networkOptions: { crossOrigin: 'anonymous' },
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -136,8 +136,10 @@ function _convert(coordsIn, newCrs, target) {
         }
     } else {
         if (coordsIn.crs === 'EPSG:4326' && newCrs === 'EPSG:4978') {
-            const cartesian = ellipsoid.cartographicToCartesian(coordsIn);
-            return target.set(newCrs, cartesian);
+            const cartesian = ellipsoid.cartographicToCartesian(coordsIn, coordsIn.geodesicNormal);
+            target.set(newCrs, cartesian);
+            target._normal = coordsIn.geodesicNormal;
+            return target;
         }
 
         if (coordsIn.crs === 'EPSG:4978' && newCrs === 'EPSG:4326') {
@@ -269,6 +271,7 @@ Coordinates.prototype.set = function set(crs, ...coordinates) {
             this._values[i] = 0;
         }
     }
+    this._normal = undefined;
     this._internalStorageUnit = crsToUnit(crs);
     return this;
 };
@@ -280,6 +283,9 @@ Coordinates.prototype.clone = function clone(target) {
         r = target;
     } else {
         r = new Coordinates(this.crs, ...this._values);
+    }
+    if (this._normal) {
+        r._normal = this._normal.clone();
     }
     r._internalStorageUnit = this._internalStorageUnit;
     return r;
--- a/src/Core/Math/Ellipsoid.js
+++ b/src/Core/Math/Ellipsoid.js
@@ -55,11 +55,9 @@ Ellipsoid.prototype.setSize = function setSize(size) {
     this._radiiSquared = new THREE.Vector3(size.x * size.x, size.y * size.y, size.z * size.z);
 };
 
-
+const k = new THREE.Vector3();
 Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto) {
-    // var n;
-    var k = new THREE.Vector3();
-    var n = this.geodeticSurfaceNormalCartographic(coordCarto);
+    const n = coordCarto.geodesicNormal.clone();
 
     k.multiplyVectors(this._radiiSquared, n);
 
--- a/src/Core/Prefab/Globe/BuilderEllipsoidTile.js
+++ b/src/Core/Prefab/Globe/BuilderEllipsoidTile.js
@@ -1,7 +1,11 @@
 import * as THREE from 'three';
-import { C } from '../../Geographic/Coordinates';
+import { C, UNIT } from '../../Geographic/Coordinates';
 import Projection from '../../Geographic/Projection';
 import OBB from '../../../Renderer/ThreeExtended/OBB';
+import Extent from '../../Geographic/Extent';
+
+const axisZ = new THREE.Vector3(0, 0, 1);
+const axisY = new THREE.Vector3(0, 1, 0);
 
 function BuilderEllipsoidTile() {
     this.projector = new Projection();
@@ -11,14 +15,16 @@ function BuilderEllipsoidTile() {
             C.EPSG_4326_Radians(0, 0),
             C.EPSG_4326_Radians(0, 0)],
         position: new THREE.Vector3(),
-        normal: new THREE.Vector3(),
     };
+
+    this.type = 'e';
 }
 
 BuilderEllipsoidTile.prototype.constructor = BuilderEllipsoidTile;
 
 // prepare params
 // init projected object -> params.projected
+
 BuilderEllipsoidTile.prototype.Prepare = function Prepare(params) {
     params.nbRow = Math.pow(2.0, params.level + 1.0);
 
@@ -33,30 +39,34 @@ BuilderEllipsoidTile.prototype.Prepare = function Prepare(params) {
 
     params.deltaUV1 = (st1 - start) * params.nbRow;
 
+    // transformation to align tile's normal to z axis
+    params.quatNormalToZ = new THREE.Quaternion().setFromAxisAngle(axisY, -(Math.PI * 0.5 - params.extent.center().latitude()));
+
     // let's avoid building too much temp objects
     params.projected = { longitudeRad: 0, latitudeRad: 0 };
 };
 
 // get center tile in cartesian 3D
-BuilderEllipsoidTile.prototype.Center = function Center(params) {
-    params.center = params.extent.center(this.tmp.coords[0])
+BuilderEllipsoidTile.prototype.Center = function Center(extent) {
+    return extent.center(this.tmp.coords[0])
         .as('EPSG:4978', this.tmp.coords[1]).xyz();
-    return params.center;
 };
 
 // get position 3D cartesian
 BuilderEllipsoidTile.prototype.VertexPosition = function VertexPosition(params) {
-    this.tmp.coords[0]._values[0] = params.projected.longitudeRad;
-    this.tmp.coords[0]._values[1] = params.projected.latitudeRad;
+    this.tmp.coords[0].set(
+        'EPSG:4326',
+        params.projected.longitudeRad,
+        params.projected.latitudeRad);
+    this.tmp.coords[0]._internalStorageUnit = UNIT.RADIAN;
 
     this.tmp.coords[0].as('EPSG:4978', this.tmp.coords[1]).xyz(this.tmp.position);
-    this.tmp.normal.copy(this.tmp.position).normalize();
     return this.tmp.position;
 };
 
 // get normal for last vertex
 BuilderEllipsoidTile.prototype.VertexNormal = function VertexNormal() {
-    return this.tmp.normal;
+    return this.tmp.coords[1].geodesicNormal;
 };
 
 // coord u tile to projected
@@ -79,9 +89,39 @@ BuilderEllipsoidTile.prototype.getUV_PM = function getUV_PM(params) {
     return t - params.deltaUV1;
 };
 
+const quatToAlignLongitude = new THREE.Quaternion();
+const quatToAlignLatitude = new THREE.Quaternion();
+
+BuilderEllipsoidTile.prototype.computeSharableExtent = function fnComputeSharableExtent(extent) {
+    // Compute sharable extent to pool the geometries
+    // the geometry in common extent is identical to the existing input
+    // with a transformation (translation, rotation)
+
+    // TODO: It should be possible to use equatorial plan symetrie,
+    // but we should be reverse UV on tile
+    // Common geometry is looking for only on longitude
+    const sizeLongitude = Math.abs(extent.west() - extent.east()) / 2;
+    const sharableExtent = new Extent(extent.crs(), -sizeLongitude, sizeLongitude, extent.south(), extent.north());
+    sharableExtent._internalStorageUnit = extent._internalStorageUnit;
+
+    // compute rotation to transform tile to position it on ellipsoid
+    // this transformation take into account the transformation of the parents
+    const rotLon = extent.west() - sharableExtent.west();
+    const rotLat = Math.PI * 0.5 - extent.center().latitude();
+    quatToAlignLongitude.setFromAxisAngle(axisZ, rotLon);
+    quatToAlignLatitude.setFromAxisAngle(axisY, rotLat);
+    quatToAlignLongitude.multiply(quatToAlignLatitude);
+
+    return {
+        sharableExtent,
+        quaternion: quatToAlignLongitude,
+        position: this.Center(extent),
+    };
+};
+
 // use for region for adaptation boundingVolume
-BuilderEllipsoidTile.prototype.OBB = function OBBFn(params) {
-    return OBB.extentToOBB(params.extent);
+BuilderEllipsoidTile.prototype.OBB = function OBBFn(boundingBox) {
+    return new OBB(boundingBox.min, boundingBox.max);
 };
 
 export default BuilderEllipsoidTile;
--- a/src/Core/Prefab/Panorama/PanoramaTileBuilder.js
+++ b/src/Core/Prefab/Panorama/PanoramaTileBuilder.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import OBB from '../../../Renderer/ThreeExtended/OBB';
 import Coordinates, { UNIT } from '../../Geographic/Coordinates';
+import Extent from '../../Geographic/Extent';
 
 function PanoramaTileBuilder(ratio) {
     this.tmp = {
@@ -14,9 +15,11 @@ function PanoramaTileBuilder(ratio) {
     }
     if (ratio === 2) {
         this.equirectangular = true;
+        this.type = 's';
         this.radius = 100;
     } else {
         this.equirectangular = false; // cylindrical proj
+        this.type = 'c';
         this.height = 200;
         this.radius = (ratio * this.height) / (2 * Math.PI);
     }
@@ -26,14 +29,19 @@ PanoramaTileBuilder.prototype.constructor = PanoramaTileBuilder;
 
 // prepare params
 // init projected object -> params.projected
+const axisX = new THREE.Vector3(0, 1, 0);
 PanoramaTileBuilder.prototype.Prepare = function Prepare(params) {
+    const angle = (params.extent.north(UNIT.RADIAN) + params.extent.south(UNIT.RADIAN)) * 0.5;
+
     if (this.equirectangular) {
+        params.quatNormalToZ = new THREE.Quaternion().setFromAxisAngle(axisX, (Math.PI * 0.5 - angle));
         params.projected = {
             theta: 0,
             phi: 0,
             radius: this.radius,
         };
     } else {
+        params.quatNormalToZ = new THREE.Quaternion().setFromAxisAngle(axisX, (Math.PI * 0.5));
         params.projected = {
             theta: 0,
             radius: this.radius,
@@ -42,15 +50,13 @@ PanoramaTileBuilder.prototype.Prepare = function Prepare(params) {
     }
 };
 
-PanoramaTileBuilder.prototype.Center = function Center(params) {
+PanoramaTileBuilder.prototype.Center = function Center(extent) {
+    const params = { extent };
     this.Prepare(params);
-
     this.uProjecte(0.5, params);
     this.vProjecte(0.5, params);
 
-    params.center = this.VertexPosition(params).clone();
-
-    return params.center;
+    return this.VertexPosition(params).clone();
 };
 
 // get position 3D cartesian
@@ -60,9 +66,8 @@ PanoramaTileBuilder.prototype.VertexPosition = function VertexPosition(params) {
     } else {
         this.tmp.position.setFromCylindrical(params.projected);
     }
-    const swap = this.tmp.position.y;
-    this.tmp.position.y = this.tmp.position.z;
-    this.tmp.position.z = this.equirectangular ? -swap : swap;
+
+    this.tmp.position.set(this.tmp.position.z, this.tmp.position.x, this.tmp.position.y);
 
     return this.tmp.position;
 };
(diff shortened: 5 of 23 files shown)


peppsac:
3 fixes, 2 feat, 1 refactor => please, split this PR.

gchoqueux:
@peppsac 
normally, these commits should be squashed, because  last 2 commits fix the first: 
* refactor (core): use common geometry in meshTile b5b20f2 
* fix (test): !change subdivision count in PLANAR test! 11d9e71
* fix (protocols): wfs, adapt the mesh transformation with the tile space e2e7c8b

and the 2 debugging features are needed to verify how this PR works :
* feat (examples): add gui debug in planar in debug mode only 70a6cb0 
* feat (core) : add bounding sphere helper in debug mode 65220e3

I remove 
fix (core): improve search for min/max altitude in xbil buffer  663b96d5

index.html
```diff
@@ -45,9 +45,9 @@
 
             var promises = []
             promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/Ortho.json').then(function(result) { return globeView.addLayer(result); }));
-            promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/OrthosCRS.json').then(function(result) { return globeView.addLayer(result); }));
-            promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/ScanEX.json').then(function(result) { return globeView.addLayer(result); }));
-            promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/Region.json').then(function(result) { return globeView.addLayer(result); }));
+            // promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/OrthosCRS.json').then(function(result) { return globeView.addLayer(result); }));
```
peppsac: You forgot to remove this temporary change

src/Core/Geographic/Coordinates.js
```diff
@@ -136,7 +136,8 @@ function _convert(coordsIn, newCrs, target) {
         }
     } else {
         if (coordsIn.crs === 'EPSG:4326' && newCrs === 'EPSG:4978') {
-            const cartesian = ellipsoid.cartographicToCartesian(coordsIn);
+            target._normal = ellipsoid.geodeticSurfaceNormalCartographic(coordsIn);
```
peppsac: Doing:
```js
target._normal = coordsIn.geodesicNormal;
```
Would avoid recomputing the normal if it was already done for `coordsIn`

src/Core/Math/Ellipsoid.js
```diff
@@ -56,10 +56,10 @@ Ellipsoid.prototype.setSize = function setSize(size) {
 };
 
 
-Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto) {
+Ellipsoid.prototype.cartographicToCartesian = function cartographicToCartesian(coordCarto, normal) {
```
peppsac: You must change the code to not modify the `normal` given by the user (modifying parameters makes the API error-prone)

src/Core/Prefab/GlobeView.js
```diff
@@ -163,6 +164,31 @@ export function createGlobeLayer(id, options) {
         return false;
     }
 
+    // return common extent to pool the geometries
+    // the geometry in common extent is identical to the existing input
+    // with a transformation (translation, rotation)
+    wgs84TileLayer.getCommonGeometryExtent = (extent) => {
```
peppsac: These functions don't belong to `GlobeView` because:
- the panorama layers will want to use them
- they are tile geometries implementation details

peppsac: Also the naming is wrong: you're basically doing a modulo operation on the longitude component (why `sizeLatitude`?) of the extent so the name has to express this.
Btw, why aren't you doing the same thing for the other angle? (I guess it has to do with the ellipsoid shape, but you may want to add a comment in the code)

src/Core/Prefab/GlobeView.js
```diff
@@ -163,6 +164,31 @@ export function createGlobeLayer(id, options) {
...
+    };
+
+    const axisZ = new THREE.Vector3(0, 0, 1);
+    const axisY = new THREE.Vector3(0, 1, 0);
+    const quatToAlignLongitude = new THREE.Quaternion();
+    const quatToAlignLatitude = new THREE.Quaternion();
+
+    // return rotation to transform tile to position it on ellipsoid
+    // this transformation take into account the transformation of the parents
+    wgs84TileLayer.getQuaternionTile = (tile) => {
```
peppsac: Wrong naming: the name doesn't express what the function does.
And the argument should be an extent, since you don't use any other properties of the `tile`.

src/Core/Prefab/Planar/PlanarTileBuilder.js
```diff
@@ -50,19 +50,8 @@ PlanarTileBuilder.prototype.vProjecte = function vProjecte(v, params)
...
-        params.extent.north(),
-        0).sub(center);
-    const min = new THREE.Vector3(
-        params.extent.west(),
-        params.extent.south(),
-        0).sub(center);
-    const translate = new THREE.Vector3(0, 0, 0);
-    // normal is up vector
-    return new OBB(min, max, undefined, translate);
+PlanarTileBuilder.prototype.OBB = function OBBFn(params, boundingBox) {
```
peppsac: Why keep the `params` parameter if it's unused?

peppsac: (same question for `BuilderEllipsoidTile`)

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,50 +41,84 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+    if (!this.cacheGeometry[level]) {
+        this.cacheGeometry[level] = new Map();
+    }
 
-    var tile = new TileMesh(geometry, params);
+    const ce = command.layer.getCommonGeometryExtent(extent);
+    const south = ce.south().toFixed(8);
+
+    // build geometry if doesn't exist
+    if (!this.cacheGeometry[level].get(south)) {
```
peppsac: This is wrong: the `south` value alone can't serve as a cache key.
You must at least depend on `command.layer.disableSkirt` (and probably on ellipsoidSize - even if currently this cannot be changed) and the `builder` used.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,50 +41,84 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+        this.cacheGeometry[level].set(south, defer());
```
peppsac: Why is it necessary to use a Promise mechanism here (instead of storing the geometry directly in the cache)?

src/Core/TileMesh.js
```diff
@@ -25,11 +25,10 @@ function TileMesh(geometry, params) {
     this.extent = params.extent;
 
     this.geometry = geometry;
-    this.normal = params.center.clone().normalize();
 
-    this.boundingSphereOffset = new THREE.Vector3();
+    this.obb = this.geometry.OBB.clone();
```
peppsac, line 29: What's the point of `this.obb`?

src/Core/TileMesh.js
```diff
@@ -43,14 +42,17 @@ function TileMesh(geometry, params) {
     this.layerUpdateState = {};
 
     this.material.setUuid(this.id);
+
+    this.invWorldQuaternion = new THREE.Quaternion();
```
peppsac: Why do you need `invWorldQuaternion`?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -219,13 +221,14 @@ function coordinateToPolygonExtruded(coordinates, properties, options) {
...
     let nbVertices = 0;
-
+    let minAltitude = Infinity;
     /* eslint-disable-next-line */
     for (const id in coordinates.featureVertices) {
         // extract contour coodinates and properties of one feature
         const { contour, property } = extractFeature(coordinates, properties, id);
         // get altitude and extrude amount from properties
         const altitudeBottom = getAltitude(options, property);
+        minAltitude = Math.min(minAltitude, altitudeBottom);
```
peppsac: What is this `minAltitude` stuff? And why is that necessary?

examples/globe_wfs_extruded.js
```diff
@@ -90,13 +92,37 @@ function acceptFeature(properties) {
...
+                mesh.scale.z = Math.min(
+                    1.0, mesh.scale.z + 0.016);
+                mesh.updateMatrixWorld(true);
+            }
+        }
+        meshes = meshes.filter(function filter(m) { return m.scale.z < 1; });
+    },
+};
+
+globeView.addFrameRequester(scaler);
```
peppsac: Nice to get this working on the globe!

peppsac (changes requested):
Here's a first round of comments on the changes proposed.

index.html
gchoqueux: ok

src/Core/Geographic/Coordinates.js
gchoqueux: done

src/Core/Math/Ellipsoid.js
gchoqueux: done

src/Core/Prefab/GlobeView.js
gchoqueux: * `getCommonGeometryExtent` moved in tile builder
* `sizeLatitude` is modified
* `getCommonGeometryExtent`, in function of builder, it doesn't do modulo

src/Core/Prefab/GlobeView.js
gchoqueux: changed in `getQuaternionFromExtent`

src/Core/Prefab/Planar/PlanarTileBuilder.js
gchoqueux: cleaned

src/Core/Scheduler/Providers/TileProvider.js
gchoqueux: yes, in another PR, there's not user cases for the moment

src/Core/Scheduler/Providers/TileProvider.js
gchoqueux: As the commands are asynchronous many tiles may ask for geometry during construction

src/Core/TileMesh.js
gchoqueux: obliged to clone because the DEM is different for each tile,

src/Renderer/ThreeExtended/Feature2Mesh.js
gchoqueux: It's the minimum altitude of the vertex of the points. It's used to position the object3D on the surface normal

src/Core/Scheduler/Providers/TileProvider.js
autra: Everything here is synchronous, so it is guaranteed that no other part of itowns (or any other js in this page really) will get executed between this line and line 83. So the deferred will actually never be used here and you can directly set the TileGeometry in the cacheGeometry, without all the promise boilerplate.

gchoqueux: However the commands are executed in asynchronous, many tiles can ask for geometry during construction.
I tested it didn't work

autra: Please be more specific. What does "during construction" mean?

>  I tested it didn't work

What did you test? What didn't work?

> However the commands are executed in asynchronous

The fact that commands execute asynchronously has nothing to do with the matter. You set a promise on line 76 and resolve it on line 83.  No code can execute between these 2 lines, whether synchronous or asynchronous. 

Have you removed the `then` logic on line 87 when you tested?

gchoqueux: Thanks @autra , I started this PR a long time ago, where commands was concurrent.
I remove the promise in provider

src/Core/Math/Ellipsoid.js
```diff
@@ -56,10 +56,10 @@ Ellipsoid.prototype.setSize = function setSize(size) {
...
     // var n;
     var k = new THREE.Vector3();
-    var n = this.geodeticSurfaceNormalCartographic(coordCarto);
+    const n = normal ? normal.clone() : this.geodeticSurfaceNormalCartographic(coordCarto);
```
peppsac: You could replace this with:
```js
const n = normal ? normal.clone() : coordCarto.geodesicNormal.clone();
```
Or even better, you could avoid changing the API and you'll get the same result by doing:
```js
const n = coordCarto.geodesicNormal.clone();
```

src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -79,9 +87,37 @@ BuilderEllipsoidTile.prototype.getUV_PM = function getUV_PM(params) {
...
+// return common extent to pool the geometries
+// the geometry in common extent is identical to the existing input
+// with a transformation (translation, rotation)
+BuilderEllipsoidTile.prototype.getCommonGeometryExtent = function getCommonGeometryExtentFn(extent) {
+    // Common geometry is looking for only on longitude
+    // TODO: It should be possible to use equatorial plan symetrie,
+    // but we should be reverse UV on tile
+    const sizeLongitude = Math.abs(extent.west() - extent.east()) / 2;
+    const communExtent = new Extent(extent.crs(), -sizeLongitude, sizeLongitude, extent.south(), extent.north());
+    communExtent._internalStorageUnit = extent._internalStorageUnit;
```
peppsac: `commun` is not an English word :)

src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -79,9 +87,37 @@ BuilderEllipsoidTile.prototype.getUV_PM = function getUV_PM(params) {
     return t - params.deltaUV1;
 };
 
+// return common extent to pool the geometries
+// the geometry in common extent is identical to the existing input
+// with a transformation (translation, rotation)
+BuilderEllipsoidTile.prototype.getCommonGeometryExtent = function getCommonGeometryExtentFn(extent) {
```
peppsac: I'd prefer to reserve `getXXX` for simple getter method.

I have more comment about this API, see layer.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,47 +42,78 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+    const builder = layer.builder;
+    const parent = command.requester;
+    const level = (command.level === undefined) ? (parent.level + 1) : command.level;
 
-    var parent = command.requester;
+    if (!this.cacheGeometry[level]) {
+        this.cacheGeometry[level] = new Map();
+    }
 
+    const ce = builder.getCommonGeometryExtent(extent);
```
peppsac: I don't think TileProvider needs to be aware of this: this could an internal feature of each tile builder.

But even without this, the API  could be simplified a lot if `getCommonGeometryExtent` was something that would return 2 things: the new extent and the transformation to apply.
And this would give you a better naming as well: something like: `computeSharableExtent`.

WDYT?

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,47 +42,78 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+    const south = ce.south().toFixed(8);
+
+    let geometry = this.cacheGeometry[level].get(south);
```
peppsac: I already commented about this: you can't use `south` as a cache key - and no it won't wait for another PR because it can break today.
Additional things to used for the key: builder type, segment count and probably disableSkirt.

Also I'd prefer a single cache than 1 cache per level.

Last but not least: when are you cleaning this cache?
If I zoom in a lot and zoom out, the geometries for high levels will never get disposed..

peppsac (changes requested):
Looks better, thanks.

You didn't change the cleanup code (see ObjectRemovalHelper.removeChildrenAndCleanup) so geometries are going to be disposed even if they are still in used...

Another question: this change is introducing a behavior change - not an API one but a behavior one (tiles' transformation is different now). So you need to document this somewhere in your commit message with a BREAKING CHANGE tag so that `Release Notes` can grab this change.

src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -33,24 +38,30 @@ BuilderEllipsoidTile.prototype.Prepare = function Prepare(params) {
...
+
     // let's avoid building too much temp objects
     params.projected = { longitudeRad: 0, latitudeRad: 0 };
 };
 
 // get center tile in cartesian 3D
-BuilderEllipsoidTile.prototype.Center = function Center(params) {
-    params.center = params.extent.center(this.tmp.coords[0])
+BuilderEllipsoidTile.prototype.Center = function Center(extent) {
+    this.tmp.coords[0]._normal = undefined;
```
peppsac: Can you please move this to `Extent` (probably in Extent.center() method) or in `Coordinates` (in the Coordinates.set())?

src/Core/Prefab/Panorama/PanoramaTileBuilder.js
```diff
@@ -97,57 +102,55 @@ PanoramaTileBuilder.prototype.vProjecte = function vProjecte(v, params) {
...
-        const max = new THREE.Vector3(length * 0.5, height * 0.5, thickness * 0.5);
-
-        const translate = new THREE.Vector3(0, 0, thickness * -0.5);
-        return new OBB(min, max, direction, translate);
-    }
+PanoramaTileBuilder.prototype.OBB = function _OBB(boundingBox) {
+    return new OBB(boundingBox.min, boundingBox.max);
+};
+
+PanoramaTileBuilder.prototype.getCommonGeometryExtent = function getCommonGeometryExtentFn(extent) {
```
peppsac: Is this still used?

src/Core/Prefab/Panorama/PanoramaTileBuilder.js
```diff
@@ -97,57 +102,55 @@ PanoramaTileBuilder.prototype.vProjecte = function vProjecte(v, params) {
...
+    // with a transformation (translation, rotation)
+
+    // TODO: It should be possible to use equatorial plan symetrie,
+    // but we should be reverse UV on tile
+    // Common geometry is looking for only on longitude
+    const sizeLongitude = Math.abs(extent.west() - extent.east()) / 2;
+    const sharableExtent = new Extent(extent.crs(), -sizeLongitude, sizeLongitude, extent.south(), extent.north());
+    sharableExtent._internalStorageUnit = extent._internalStorageUnit;
+
+    // compute rotation to transform tile to position it on ellipsoid
```
peppsac: The comment refers to an "ellipsoid" but there's no ellipsoid in this context.
Can you please fix the comment?

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,47 +42,73 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+
+    if (!this.cacheGeometry[level]) {
+        this.cacheGeometry[level] = new Map();
+    }
+
+    const { sharableExtent, quaternion, position } = builder.computeSharableExtent(extent);
+    const south = sharableExtent.south().toFixed(8);
 
-    var parent = command.requester;
+    console.log(quaternion, position);
```
peppsac: This should probably be removed :)

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,47 +42,73 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+    if (!geometry) {
+        const paramsGeometry = {
+            extent: sharableExtent,
+            level,
+            segment: layer.segments || 16,
+            disableSkirt: layer.disableSkirt,
+        };
+
+        geometry = new TileGeometry(paramsGeometry, builder);
+        this.cacheGeometry[level].set(south, geometry);
```
peppsac: `level` isn't a good cache key (see earlier comment https://github.com/iTowns/itowns/pull/586#discussion_r157729520)

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -34,6 +26,7 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
 
     layer.level0Nodes = [];
     layer.onTileCreated = layer.onTileCreated || (() => {});
+    layer.object3d.invQuaternion = layer.object3d.quaternion.clone().inverse();
```
peppsac: `invQuaternion` is never updated. And IMHO it's a bad idea to introduce this concept: just call inverse() when you need it.
Or maybe rewrite the code doing this to avoid requiring a matrix inversion (in worldToLocal) + 2 quaternion.inverse() call.

gchoqueux:
@peppsac updated

src/Process/ObjectRemovalHelper.js
```diff
@@ -8,7 +8,12 @@ export default {
             obj.dispose();
         } else {
             if (obj.geometry) {
-                obj.geometry.dispose();
+                if (obj.geometry.count) {
```
peppsac: This isn't a good idea since you're breaking geometry disposal for other types of geometries.

See other comment  in TileProvider.js

src/Process/PanoramaTileProcessing.js
```diff
@@ -1,3 +1,5 @@
+// import * as THREE from 'three';
```
peppsac: Remove dead code please.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,47 +41,80 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+        const paramsGeometry = {
+            extent: sharableExtent,
+            level,
+            segment,
+            disableSkirt: layer.disableSkirt,
+        };
+
+        geometry = new TileGeometry(paramsGeometry, builder);
+        geometry._count = 0;
+        this.cacheGeometry.set(key, geometry);
```
peppsac: You could do this:
```js
geometry.dispose = () => {
    geometry._count--;
    if (geometry._count == 0) {
       THREE.BufferGeometry.prototype.dispose.call(geometry);
       this.cacheGeometry.delete(key);
    }
}
```
And remove the `count` property. This way the ref-counting feature would be internal to TileProvider.

peppsac (commented):
Did you observe some measurable memory or perf gains with this changeset?

gchoqueux:
> Did you observe some measurable memory or perf gains with this changeset?

the processes time (to build tiles) and used memory (geometryTile) are divided by 3

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -49,47 +41,79 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
...
+        level,
+        materialOptions: layer.materialOptions,
     };
 
-    const geometry = new TileGeometry(params, command.layer.builder);
-
-    var tile = new TileMesh(geometry, params);
-
-    tile.layer = command.layer.id;
+    geometry.count++;
```
peppsac: Shouldn't it be `geometry._count++`?

src/Core/Geographic/Extent.js
```diff
@@ -245,6 +245,7 @@ Extent.prototype.center = function center(target) {
     let c;
     if (target) {
         Coordinates.call(target, this._crs, this._values[0], this._values[2]);
+        target._normal = undefined;
```
peppsac: Resetting the `_normal` value should be done inside `Coordinates`.

src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -33,24 +40,27 @@ BuilderEllipsoidTile.prototype.Prepare = function Prepare(params) {
...
-    params.center = params.extent.center(this.tmp.coords[0])
+BuilderEllipsoidTile.prototype.Center = function Center(extent) {
+    return extent.center(this.tmp.coords[0])
         .as('EPSG:4978', this.tmp.coords[1]).xyz();
-    return params.center;
 };
 
 // get position 3D cartesian
 BuilderEllipsoidTile.prototype.VertexPosition = function VertexPosition(params) {
+    this.tmp.coords[0]._normal = undefined;
```
peppsac: Proposal: move this reset in `Coordinates.set()` and use it here:
```js
this.tmp.coords[0].set(
    'EPSG:4326',
    params.projected.longitudeRad,
    params.projected.latitudeRad);
```

peppsac (commented):
> the processes time (to build tiles) and used memory (geometryTile) are divided by 3

This info should be in one of the commits message - and maybe add explain you measured it.

The code is almost good, I just left a few comments to address.

gchoqueux:
> the processes time (to build tiles) and used memory (geometryTile) are divided by 3

I compare master and this PR with different calls from the `GlobeControls` API :
 * Total running time on `TileProvider.executeCommand`
  * Snapshots of memory occupied by all instances of  `TileGeometry`

src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -33,24 +40,30 @@ BuilderEllipsoidTile.prototype.Prepare = function Prepare(params) {
...
+    this.tmp.coords[0].set(
+        'EPSG:4326',
+        params.projected.longitudeRad,
+        params.projected.latitudeRad);
+
+    this.tmp.coords[0]._internalStorageUnit = UNIT.RADIAN;
 
     this.tmp.coords[0].as('EPSG:4978', this.tmp.coords[1]).xyz(this.tmp.position);
-    this.tmp.normal.copy(this.tmp.position).normalize();
+    this.tmp.normal.copy(this.tmp.coords[1].geodesicNormal);
```
peppsac: Btw, you could remove `this.tmp.normal` completely and replace VertexNormal code by `return this.tmp.coords[1].geodesicNormal`.

peppsac (commented):
Thanks for all the edits, the PR looks good now.

I left a small suggestion to simplify the code. And of course you need to rebase / edit commits properly before merging :)

gchoqueux:
@peppsac  thanks for review
commit message updated

gchoqueux:
@peppsac this PR can be merged

peppsac (approved):
Ok this PR looks good now; thanks for the hard work!

peppsac:
@gchoqueux This PR broke the http://www.itowns-project.org/itowns/examples/multiglobe.html example.

Could you take a look at this please? (I can open a new issue if you prefer)

PR #587, opened 2017-11-17
fix(loader): remove two warnings

gliegard:
## Description
Remove warning from GLTFLoader

Prior to this change, we can see these warnings
WARNING in ./src/Renderer/ThreeExtended/B3dmLoader.js
12:26-42 "export 'GLTFLoader' (imported as 'THREE') was not found in 'three'
WARNING in ./src/Renderer/ThreeExtended/GLTFLoader.js
18:0-16 "export 'GLTFLoader' (imported as 'THREE') was not found in 'three'

### DIFF
--- a/src/Renderer/ThreeExtended/GLTFLoader.js
+++ b/src/Renderer/ThreeExtended/GLTFLoader.js
@@ -12,7 +12,7 @@ import * as THREE from 'three';
  * @author Takahiro / https://github.com/takahirox
  */
 
-THREE.GLTFLoader = ( function () {
+export default (function () {
 
 	function GLTFLoader( manager ) {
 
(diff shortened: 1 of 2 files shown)


src/Renderer/ThreeExtended/GLTFLoader.js
```diff
@@ -12,7 +12,7 @@ import * as THREE from 'three';
  * @author Takahiro / https://github.com/takahirox
  */
 
-THREE.GLTFLoader = ( function () {
+GLTFLoader = ( function () {
```
autra: You need to export it. `export default (function() {...}` would be enough I guess

autra (changes requested):
It removes warnings, but break itowns :-O

autra:
To be more precise, the error is `ReferenceError: assignment to undeclared variable GLTFLoader`, because the outer GLTFLoader var is not defined in this scope, directly exposing the result of calling the outer function will get rid of it (not tested though).

gliegard:
tanks for the precisions :)
I've updated the code !

autra (approved):
Perfect, thanks for this clean!

Issue #589, opened 2017-11-22
Allow multiple callback to View#onBeforeRender/onAfterRender


zarov:
### Suggestion
We can improve the `preRender`/`onAfterRender`  mechanism by allowing more than one callback per property.

### Context
There are only two examples for now where this is used (see below). The two of them are overloading in each case the `preRender`/`onAfterRender` properties of a `View`, which is inconvenient in case we have things already set in those properties (like in [`GlobeView` with the `preRender` property](https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/GlobeView.js#L246-L272)). 

- https://github.com/iTowns/itowns/blob/76ae961cbaffb0736ce5babaf27e0f2a34742985/examples/layersColorVisible.html#L93-L103 (note that the `preRender` here needs to be changed in `onBeforeRender`)
- https://github.com/iTowns/itowns/blob/d919316900e3f209555525483f08702127701ec6/examples/globe.js#L39-L48

### Solution
- Modify the `View` to have two arrays, one for `onBeforeRender` and the other for `onAfterRender`.
- Call every these arrays in `MainLoop`, with `view.onBeforeRender.forEach((fn) => fn.call(this))` or something like this

Note that this solved https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/GlobeView.js#L247-L249

WDYT ?

autra:
Even though I agree these callbacks are totally inconvenient, let's not jump too soon to technical discussion. 
You didn't actually explained your use case, and with so little context, it's difficult to know what would fit you best. 

What problem are you trying to solve (in a general, functional fashion)? Why do you think preRender / afterRender are a good technical solution for your problem? Can you provide real-life example (ideally with the functions you want to register in these hook)? 

Maybe we want several hook points for a FrameRequester instead? Or maybe we just want to have some `before-render` and `after-render` events dispatched instead?

But yeah, about these callbacks, you're totally right, we need to do something about these. If I have to make a wild guess, I *think* the former (several hook points for frame requesters) would be a good solution. It's been several times already that we tell ourselves we need this.

In the current state of the code, [frameRequesters are conceptually `preUpdate` callbacks](https://github.com/iTowns/itowns/blob/master/src/Core/MainLoop.js#L65). So it'd be easy to extend this concept to have `preRender`, `postRender` etc... callbacks. We could register them this way:

```js
view.addFrameRequester('preRender', cb1);
view.addFrameRequester('postRender', cb2);
```
[...]

PR #590, opened 2017-11-22
fix (core): allow multiple callback before and after render

zarov:
This commit:
- changes the View#onAfterRender method to an array
- adds the View#onBeforeRender array
- changes the GlobeView#preRender method to onBeforeRender.push
- changes the onAfterRender overloading in globe example
- changes the preRender overloading in layers color example
- modifies MainLoop#_renderView to use onBeforeRender and onAfterRender

Solves #589

### DIFF
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -36,7 +36,7 @@ if (!renderer) {
     miniView.mainLoop.gfxEngine.renderer.setClearColor(0x000000, 0);
 
     // update miniview's camera with the globeView's camera position
-    globeView.onAfterRender = function onAfterRender() {
+    globeView.onAfterRender.push(() => {
         // clamp distance camera from globe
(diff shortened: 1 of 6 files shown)


examples/globe.js
```diff
@@ -36,7 +36,7 @@ if (!renderer) {
     miniView.mainLoop.gfxEngine.renderer.setClearColor(0x000000, 0);
 
     // update miniview's camera with the globeView's camera position
-    globeView.onAfterRender = function onAfterRender() {
+    globeView.onAfterRender.push(() => {
```
autra, line 39: Typically, this use case of `I want to execute something after each render` is a very good fit for a frame requester hooked on a `postRender` (or better in this case: `postUpdate`) phase.

zarov:
Closing it given #591

PR #591, opened 2017-11-23
Feature: rework FrameRequesters to be used from the MainLoop

zarov:
## Description

This PR adds eight moments when a `FrameRequester` can be called in the
MainLoop :
- `UPDATE_START`
- `BEFORE_CAMERA_UPDATE`
- `AFTER_CAMERA_UPDATE`
- `BEFORE_LAYER_UPDATE`
- `AFTER_LAYER_UPDATE`
- `BEFORE_RENDER`
- `AFTER_RENDER`
- `UPDATE_END`

They can be used like this anywhere (at least where a view is reachable):
`view.addFrameRequester(frameRequester, itowns.View.BEFORE_CAMERA_UPDATE);`
where `frameRequester` being either a function or a `FrameRequester`.

They are removed following a similar process:
`view.removeFrameRequester(frameRequester, itowns.View.BEFORE_CAMERA_UPDATE);`.

In the MainLoop, they are triggered each time (according to where they
belong) using `view.execFrameRequesters(itowns.View.BEFORE_CAMERA_UPDATE);`.

## Motivation and Context
This change adds some coherence in naming, and allow users to execute commands depending on the rendering update.

Inspired by https://github.com/iTowns/itowns/compare/master...peppsac:feat_attach_point_frame_req, responds to #589 and replaces #590

### DIFF
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -26,7 +26,7 @@ if (!renderer) {
         // since the mini globe will always be seen from a far point of view (see minDistance above)
         maxSubdivisionLevel: 2,
         // Don't instance default controls since miniview's camera will be synced
-        // on the main view's one (see globeView.onAfterRender)
+        // on the main view's one (see globeView.addFrameRequester)
         noControls: true,
     });
 
@@ -36,7 +36,7 @@ if (!renderer) {
     miniView.mainLoop.gfxEngine.renderer.setClearColor(0x000000, 0);
 
     // update miniview's camera with the globeView's camera position
-    globeView.onAfterRender = function onAfterRender() {
+    globeView.addFrameRequester(itowns.MAIN_LOOP_EVENTS.AFTER_RENDER, function () {
         // clamp distance camera from globe
         var distanceCamera = globeView.camera.camera3D.position.length();
         var distance = Math.min(Math.max(distanceCamera * 1.5, minDistance), maxDistance);
@@ -45,7 +45,7 @@ if (!renderer) {
         camera.position.copy(globeView.controls.moveTarget()).setLength(distance);
         camera.lookAt(globeView.controls.moveTarget());
         miniView.notifyChange(true);
-    };
+    });
 
     // Add one imagery layer to the miniview
     itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) { miniView.addLayer(layer); });
--- a/examples/multiglobe.html
+++ b/examples/multiglobe.html
@@ -93,27 +93,25 @@
             }, globe2);
 
             // Globe animation
-            var animator = {
-                t: 0,
-                update: function(dt, ignore) {
-                    this.t = ignore ? 0 : Math.min(1, dt / 1000 + this.t);
+            var t = 0;
+            function animator (dt, ignore) {
+                t = ignore ? 0 : Math.min(1, dt / 1000 + t);
 
-                    layers[front].object3d.position.y = itowns.THREE.Math.lerp(front ? 10000000 : -10000000, 0, this.t);
-                    layers[front].object3d.scale.setScalar(itowns.THREE.Math.lerp(0.3, 1, this.t));
+                layers[front].object3d.position.y = itowns.THREE.Math.lerp(front ? 10000000 : -10000000, 0, t);
+                layers[front].object3d.scale.setScalar(itowns.THREE.Math.lerp(0.3, 1, t));
 
-                    layers[1 - front].object3d.position.y = itowns.THREE.Math.lerp(front ? -10000000 : 10000000, 0, 1 - this.t);
-                    layers[1 - front].object3d.scale.setScalar(itowns.THREE.Math.lerp(0.3, 1, 1 - this.t));
+                layers[1 - front].object3d.position.y = itowns.THREE.Math.lerp(front ? -10000000 : 10000000, 0, 1 - t);
+                layers[1 - front].object3d.scale.setScalar(itowns.THREE.Math.lerp(0.3, 1, 1 - t));
 
-                    layers[0].object3d.updateMatrixWorld(true);
-                    layers[1].object3d.updateMatrixWorld(true);
+                layers[0].object3d.updateMatrixWorld(true);
+                layers[1].object3d.updateMatrixWorld(true);
 
-                    if (this.t < 1) {
-                        globeView.notifyChange(true);
-                    } else {
-                        globeView.removeFrameRequester(animator);
-                    }
+                if (t < 1) {
+                    globeView.notifyChange(true);
+                } else {
+                    globeView.removeFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, animator);
                 }
-            };
+            }
 
             // Last but not least, add 'ref' object to three.js scene, otherwise our globes
             // won't be displayed
@@ -126,8 +124,8 @@
                 if (evt.keyCode == 32) {
                     front = 1 - front;
 
-                    animator.t = 0;
-                    globeView.addFrameRequester(animator);
+                    t = 0;
+                    globeView.addFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, animator);
                     globeView.notifyChange(true);
                 }
             }
--- a/examples/wfs.js
+++ b/examples/wfs.js
@@ -5,7 +5,6 @@ var extent;
 var viewerDiv;
 var view;
 var meshes;
-var scaler;
 
 // Define projection that we will use (taken from https://epsg.io/3946, Proj4js section)
 itowns.proj4.defs('EPSG:3946',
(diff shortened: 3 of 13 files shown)


zarov:
Note that I'm undecided on `view.addFrameRequester(frameRequester, itowns.View.BEFORE_CAMERA_UPDATE);`. Maybe arguments could be reversed to have `view.addFrameRequester(itowns.View.BEFORE_CAMERA_UPDATE, frameRequester);`, and thus a  quicker understanding when reading the code ?

src/Core/View.js
```diff
@@ -394,18 +396,60 @@ View.prototype.getLayers = function getLayers(filter) {
...
  * MainLoop update with the time delta between last update, or 0 if the
  * MainLoop has just been relaunched.
+ * @param {String} when - decide when the frameRequester.update function should
+ * be called during the update cycle.
  */
-View.prototype.addFrameRequester = function addFrameRequester(frameRequester) {
-    this._frameRequesters.push(frameRequester);
+View.prototype.addFrameRequester = function addFrameRequester(frameRequester, when = View.AFTER_CAMERA_UPDATE) {
+    if (typeof frameRequester.update !== 'function' && typeof frameRequester !== 'function') {
+        throw new Error('frameRequester must be a function or have a function \'update\'');
```
autra: Now that frameRequesters can be notified several time, what about requiring objects to have named method `onBeforeRender`, `onAfterRender` instead? This could simplify client code if they need callbacks on several hook points and want a single FrameRequester to handle them. And maybe require to specify the hook point only if frameRequester is a function?

Or maybe allow all those 3 mechanisms? So these 3 statements would be accepted:

```js
// adding a simple function
view.addFrameRequester(View.UPDATE_START, function(dt) { /* ... */ });
// adding a simple object with update fn
view.addFrameRequester(View.UPDATE_START, { update: () => { /* ... */ } });
// adding an object with some `on...` methods. We don't specify hook points in this case
view.addFrameRequester({
    onUpdateStart: /* ... */,
    onBeforeCameraUpdate: /* ... */,
    /* ... */
});
```
Maybe it's overkill I don't know. WDYT? At least n°1 and either 2 or 3 should stay supported though.

zarov: > This could simplify client code if they need callbacks on several hook points and want a single FrameRequester to handle them.

Well they can always do something like this, which is rather simple IMHO:
```js
function frameRequester(dt) { /* ... */}
view.addFrameRequester(View.UPDATE_START, frameRequester);
view.addFrameRequester(View.UPDATE_END, frameRequester);
```
But I concur that your example can be prettier: I'm not fond of how those constants look. So we would have:
- `UPDATE_START` -> `onUpdateStart`
- `BEFORE_CAMERA_UPDATE` -> `onBeforeCameraUpdate`
- `AFTER_CAMERA_UPDATE` -> `onAfterCameraUpdate`
- `BEFORE_LAYER_UPDATE` -> `onBeforeLayerUpdate`
- `AFTER_LAYER_UPDATE` -> `onAfterLayerUpdate`
- `BEFORE_RENDER` -> `onBeforeRender`
- `AFTER_RENDER` -> `onAfterRender`
- `UPDATE_END` -> `onUpdateEnd`

I'm for one solution though, either 1+2 or 3. Any opinion of someone else on the matter ?

peppsac: Solution 3 is nicer if you want to work with an object, but how is this going to work if you want to register a function?

Other random ideas: 
```js
view.onUpdateStart.addFrameRequester(...)
view.exec(myfunction).on('View.UPDATE_START')
```

zarov: Do you find bothering to only be able to write `view.addFrameRequester({ onUpdateStart: /* ... */ });` for one function ?

From your first idea, we could also have:
```js
view.onUpdateStart.add(...);
view.onUpdateStart.remove(...);
view.onUpdateStart.exec();
```

peppsac: I don't have a strong opinion about this so choose whatever suits you best :)

autra: > Well they can always do something like this, which is rather simple IMHO:
>```js
>function frameRequester(dt) { /* ... */}
>view.addFrameRequester(View.UPDATE_START, frameRequester);
>view.addFrameRequester(View.UPDATE_END, frameRequester);
>```

Not in the current state. If we want this to work, we need to pass the name of the current step as a parameter to the frame requester when we call it: you want to be able to know what the current step is *from inside* your frame requester. It's possible, but that would mean each frame requester that wants to be notified only for one step must start with `if (currentStep === myStep)`. A bit cumbersome if you ask me.

> Solution 3 is nicer if you want to work with an object, but how is this going to work if you want to register a function?

Solution 3 would only be supported for objects. Solution 1 and/or 2 for both objets and functions.  We can have all 3 if we want to. Not sure I understood your question actually :-)

> I'm for one solution though, either 1+2 or 3. Any opinion of someone else on the matter ?

Actually, maybe the safest solution for today would be to do 1/ only (aka `view.addFrameRequester(hookName, fn)`), because it works in all cases, it is the most js idiomatic and the simplest. So I'd say we could even drop 2/ (objects with `update` fn) for now (there would be a couple of controls to amend, but that will be very simple). Adding 2/ or 3/ later would be easy anyway, if we feel the need.

I'm not fond of the other possibilities you guys proposed.

WDYT?

zarov: Good to me to keep only 1, I don't know what the original thought behind `update` was. I think we can all agree on 1 ? I just need to invert `fn` and `hookName` and that'll be good. We can always come back later if someone find a better solution.

autra: >  I don't know what the original thought behind update was.

I did that :-) My original work on FrameRequesters was to move away from having listeners everywhere (like in GlobeControls for instance) OR adding `requestAnimationFrame` in random part of the code. You know that already, but the end goal is to provide a mechanism that offers better performances and strong guarantees (be called at each frame after or before a definite step) for client code than events. 

The first application for it was controls (see every controls except GlobeControls), and as GlobeControls does have an `update` method (that was the only control at the time), I started from that. Later @peppsac added the ability to use simple functions. We still need to retrofit GlobeControls to use or be a frame requester though.

That's the story! It's quite a recent one though, so we can still experiment with this concept.

zarov: I see, I think this may be too specific to Controls then, so better drop it for now. I'm changing all of this then !

peppsac (approved):
LGTM.
(except you forgot to modify the `wfs.js` example)

src/Core/View.js
```diff
@@ -364,48 +362,96 @@ View.prototype.getLayers = function getLayers(filter) {
...
 /**
  * Add a frame requester to this view.
  *
  * FrameRequesters can activate the MainLoop update by calling view.notifyChange.
  *
- * @param {FrameRequester} frameRequester
- * @param {Function} frameRequester.update - update will be called at each
- * MainLoop update with the time delta between last update, or 0 if the
- * MainLoop has just been relaunched.
+ * @param {String} when - decide when the frameRequester should be called during
```
autra, line 403: Could you document the possible values?  You can list them here, or add jsdoc when the constants are defined and reference it here, as you like

zarov: Just updated it ! I moved the constants in `MainLoop`, under `MAIN_LOOP_EVENTS`.

examples/multiglobe.html
```diff
@@ -127,7 +127,7 @@
                     front = 1 - front;
 
                     animator.t = 0;
-                    globeView.addFrameRequester(animator);
+                    globeView.addFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, animator);
```
gchoqueux: `animator` isn't a function.
So space key (switch feature) doesn't work

examples/wfs.js
```diff
@@ -115,7 +115,7 @@ scaler = {
     },
 };
 
-view.addFrameRequester(scaler);
+view.addFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, scaler);
```
gchoqueux: `scaler` isn't a function.

examples/globe.js
```diff
@@ -36,7 +36,7 @@ if (!renderer) {
     miniView.mainLoop.gfxEngine.renderer.setClearColor(0x000000, 0);
```
gchoqueux, line 36: update comment in line 29

gchoqueux (changes requested):

gchoqueux (approved):

src/Core/View.js
gchoqueux: @autra it's ok for you?

examples/multiglobe.html
```diff
@@ -110,7 +110,7 @@
                     if (this.t < 1) {
                         globeView.notifyChange(true);
                     } else {
-                        globeView.removeFrameRequester(animator);
+                        globeView.removeFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, animator.update);
```
autra: If I'm not mistaken, animator uses `this`. You need to bind update to animator for this to work,  or maybe better: make animator a simple function (with t in the outside scope). As you wish :-)

examples/wfs.js
```diff
@@ -115,7 +115,7 @@ scaler = {
     },
 };
 
-view.addFrameRequester(scaler);
+view.addFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, scaler.update);
```
autra: (not very important, but as you'll amend this PR anyway, scaler could be a simple function)

autra (changes requested):
Could you squash your commits? I don't see the added value of having a `feat` commit for examples. Beside, we want examples to work at every commit of the history anyway.

examples/multiglobe.html
zarov: Changed it for a global `t` and simpler `animator`.

examples/wfs.js
zarov: Good call, done !

autra (approved):
LGTM thanks!

One open questions: I'm wondering why part of FrameRequester's code is in view and the other in MainLoop. This split seems artificial to me now. Maybe everything could be in MainLoop? (It makes sense to me to add a frameRequester to a loop, and it would avoid this split)

Anyway, nothing that can prevent this to get merged now :-)

zarov:
Thank you !

Well as I'm a newbie here I didn't want to disrupt your previous work in `View` ;) But you are right, it may be more logical to have all of this in MainLoop. The only downside I can see is the call length will be stretched with `view.mainLoop.addFrameRequester`, but that's a very tiny issue in this case.

Anyway, if we want to change that we may want to do it before the next release to avoid too much API changes.

autra:
We could keep the `add/removeFrameRequester` in View (or just a proxy) for ease of use, while having all the rest in MainLoop. I think this work can be done without breaking the current api.

PR #592, opened 2017-12-05
chore: add out/ to .gitignore

zarov:
Really quick PR: the `out/` folder is generated by running `npm run doc`, and it is not to be commited.

It bothers me to have a dirty index after generating the documentation ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /node_modules/
 /dist/
 /lib/
+/out/


tbroyer (approved):
Fwiw, we may want to fix the `package.json`'s `"files"` which lists `"doc"` and should possibly use `"out"` instead (or maybe call JSDoc with `-d ./doc` and update the `autodeploy.sh` script, and `.gitignore`)

zarov:
Thanks, yes I think that a durable solution would be to move this `out` folder to a clean one like `doc`.

Issue #285, opened 2017-04-12
Implement getCapabilities call to follow WMTS, WMS etc... specs


autra:
Before querying WMTS or WMS servers for textures, we need to query their getCapabilities method.

I have already started implemented this, so better create an issue to follow this.

gcebelieu:
If this this is implemented, user should still have the possibility to disable this feature and provide wms or wmts layers configuration by themself. Indeed : querying capabilities introduces a delay before accessing to textures that may degrade user experience.

autra:
@gcebelieu I don't plan on breaking existing functionalities (yet). 

Some questions: 
- how will user update their conf if the getCapabilities has changed? Will they periodically check manually?
- what is the geoportail planning to use ? GetCap of json conf?
- If the latter, where will you put your json configuration ? External files ? in Index.html ? In code ?

gcebelieu:
> how will user update their conf if the getCapabilities has changed? Will they periodically check manually?

My opinion is that if user (I mean : application developer using itowns) don't use getCapabilities features but json conf, it is its  responsability to check capabilities changes.

> what is the geoportail planning to use ? GetCap of json conf?

This is my point :) For geoportal site we will use json conf with iTowns. 

> If the latter, where will you put your json configuration ? External files ? in Index.html ? In code ?

The process is already in place and running : We grab all layers conf "periodically" from getcapabilites (this is a backoffice web site task), store them in the web site dedicated database. At run time, they are thus immediately available and given as json parameters to the viewer for each layers users display (in 2D or 3D mode either). The libraries currently used (OpenLayers and VirtualGeo) allow this.

autra:
@gcebelieu is this code published somewhere, to have a clear idea of what is happening in your side?

gcebelieu:
> @gcebelieu is this code published somewhere, to have a clear idea of what is happening in your side?

No, it isn't. The mechanism I explained aboved with our backoffice has been set up precisely to avoid latency due to external webservices call (like getCapabilities) when adding layers to the viewer. The webmaster of the site configure (once and for all) layers that has to be proposed on the geoportal site before publishing them. The viewer has then all the parameters needed to simply send GetMap or GetTile requests.

Thus my need is only to have the possibility of explicitly providing layers parameters to iTowns (as it is currently possible) to avoid a getCapabilities request when adding a wmts or wms layer if this feature (which may have also an interest, I don't deny it) is added to iTowns.

autra:
ok thanks for your answers.

PR #594, opened 2017-12-06
feat(WMS): call getCapabilities during layer preprocessing

autra:
## Description

Add GetCapabilities query for WMS layers. This PR implement the fetching of capabilities and set or check the following informations:
- projection
- extent
- format

BREAKING CHANGE:

- WMS layer will now query getCapabilities. If you want to disable this  behaviour (because you already provide all the informations in the json object), you need to set layer.disableGetCap to true.

- previously we had layer.format and layer.options.mimetype coexisting, with no clear difference in role. This commit removes layer.options.mimetype to only keep layer.format. This is needed to reliably check configured format.

## Motivation and Context

Supporting getCapabilities in our providers is a necessary step to present a usable SIG application for the outside world, let's start this.

Moreover, if we want itowns to be able to display a layer from its url only, we need to be able to query capabilities from these layers.

Work that would need to follow this PR:
- support getCapabilities for WFS and WMTS 
- add the possibility to create an entire scene (view + layers) with just a WMS/WMTS/WFS url.

This is a first pass at fixing #285.

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -9,6 +9,9 @@ import proj4 from 'proj4';
 import Ellipsoid from '../Math/Ellipsoid';
 
 proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
+// used in some protocol
+// CRS:84 is equivalent to EPSG:4326 - ie, basic WGS84 degrees.
+proj4.defs('CRS:84', proj4.defs('EPSG:4326'));
 
 const projectionCache = {};
 
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -261,7 +261,7 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
  * @property {string} protocol wmts and wms (wmtsc for custom deprecated)
  * @property {string} url Base URL of the repository or of the file(s) to load
  * @property {string} format Format of this layer. See individual providers to check which formats are supported for a given layer type.
- * @property {boolean} disableGetCap set this to true to skip getCapabilities request (if the layer object has already all the needed infos)
+ * @property {boolean} disableGetCapabilities set this to true to skip getCapabilities request (if the layer object has already all the needed infos)
  * @property {NetworkOptions} networkOptions Options for fetching resources over network
  * @property {Object} updateStrategy strategy to load imagery files
  * @property {OptionsWmts|OptionsWms} options WMTS or WMS options
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -261,6 +261,7 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
  * @property {string} protocol wmts and wms (wmtsc for custom deprecated)
  * @property {string} url Base URL of the repository or of the file(s) to load
  * @property {string} format Format of this layer. See individual providers to check which formats are supported for a given layer type.
+ * @property {boolean} disableGetCap set this to true to skip getCapabilities request (if the layer object has already all the needed infos)
  * @property {NetworkOptions} networkOptions Options for fetching resources over network
  * @property {Object} updateStrategy strategy to load imagery files
  * @property {OptionsWmts|OptionsWms} options WMTS or WMS options
@@ -359,6 +360,11 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         }
 
         layer.whenReady.then((layer) => {
+            // if we still haven't any validExtent after getCapabilities, we assume
+            // it is the same as the configured extent
+            if (!layer.validExtent) {
+                layer.validExtent = layer.extent;
+            }
             this.notifyChange(parentLayer || layer, false);
             if (!this._frameRequesters[MAIN_LOOP_EVENTS.UPDATE_END] ||
                     this._frameRequesters[MAIN_LOOP_EVENTS.UPDATE_END].indexOf(this._allLayersAreReadyCallback) == -1) {
--- a/examples/planar_vector.html
+++ b/examples/planar_vector.html
@@ -55,6 +55,7 @@
                 networkOptions: { crossOrigin: 'anonymous' },
                 type: 'color',
                 protocol: 'wms',
+                disableGetCapabilities: true, // their getCap is too slow
                 version: '1.3.0',
                 id: 'wms_imagery',
                 name: 'Ortho2009_vue_ensemble_16cm_CC46',
@@ -68,6 +69,7 @@
                 url: 'https://download.data.grandlyon.com/wms/grandlyon',
                 type: 'elevation',
                 protocol: 'wms',
+                disableGetCapabilities: true, // their getCap is too slow
                 networkOptions: { crossOrigin: 'anonymous' },
                 version: '1.3.0',
                 id: 'wms_elevation',
--- a/examples/wfs.html
+++ b/examples/wfs.html
@@ -61,6 +61,7 @@
                 networkOptions: { crossOrigin: 'anonymous' },
                 type: 'color',
                 protocol: 'wms',
+                disableGetCapabilities: true, // their getCap call is too slow
                 version: '1.3.0',
                 id: 'wms_imagery',
                 name: 'Ortho2009_vue_ensemble_16cm_CC46',
--- a/package-lock.json
+++ b/package-lock.json
@@ -198,6 +198,12 @@
         }
       }
     },
+    "abab": {
+      "version": "1.0.4",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.4.tgz",
+      "integrity": "sha1-X6rZwsB/YN12dw9xzwJbYqY8/U4=",
+      "dev": true
+    },
     "accepts": {
       "version": "1.3.4",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
@@ -229,6 +235,15 @@
           "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c=",
           "dev": true
         }
+      }
+    },
+    "acorn-globals": {
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/acorn-globals/-/acorn-globals-4.1.0.tgz",
+      "integrity": "sha512-KjZwU26uG3u6eZcfGbTULzFcsoz6pegNKtHPksZPOUsiKo5bUmiBPa38FuHZ/Eun+XYh/JCCkS9AS3Lu4McQOQ==",
+      "dev": true,
+      "requires": {
+        "acorn": "^5.0.0"
       }
     },
     "acorn-jsx": {
@@ -348,6 +363,12 @@
       "integrity": "sha1-45sJrqne+Gao8gbiiK9jkZuuOcQ=",
       "dev": true
     },
+    "array-equal": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "integrity": "sha1-jCpe8kcv2ep0KwTHenUJO6J1fJM=",
+      "dev": true
+    },
     "array-find": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/array-find/-/array-find-1.0.0.tgz",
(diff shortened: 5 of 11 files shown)


autra:
@iTowns/reviewers please r?

gchoqueux (changes requested):
We need to valid the default value of `layer.disableGetCap`.

Why you change  `options.mimetype` to format?
You have to support both if you make this type of change. (and tag to deprecated)

autra:
> Why you change options.mimetype to format?

From the description

> previously we had layer.format and layer.options.mimetype coexisting, with no clear difference in role. This commit removes layer.options.mimetype to only keep layer.format. This is needed to reliably check configured format.

2 option for the same thing is not needed. I picked the better.

>  You have to support both if you make this type of change. (and tag to deprecated)

For such a simple thing? Deprecation process is necessary if the change will notably impact client, preventing them to work before extensive work is done to adapt themselves to a new version. It is not necessary to go into such a process for a simple renaming. As the commit is annotated with a `BREAKING CHANGE`, it'll appear automatically in the release note. I'll make sure it is clearly visible.

> We need to valid the default value of layer.disableGetCap.

It is very surprising that itowns managed to do *without* getCap this far honestly :-) Every SIG app I'm aware of do that by default. Of course we'll keep a way to bypass it (hence the disableGetCap, the name can be changed btw). It is very useful in some case I agree (for instance for Lyon data, their getCap takes 17 sec :-/ )

I believe this will make iTowns a lot simpler to use, especially for WMTS layers.

src/Core/Geographic/Coordinates.js
```diff
@@ -10,6 +10,8 @@ import mE from '../Math/MathExtended';
 import Ellipsoid from '../Math/Ellipsoid';
 
 proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
+// used in some protocol
+proj4.defs('CRS:84', proj4.defs('EPSG:4326'));
```
gchoqueux, line 14: can you add comment?
```js
// CRS:84 is equivalent to EPSG:4326 - ie, basic WGS84 degrees.
```

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -7,6 +7,78 @@
...
+/* For reference, there are also these 2 possible tags that do not translate to mimetype
+   INIMAGE display text in the returned image -->
+   BLANK return an image with all pixels transparent if
+   */
+
+function _parseSupportedCrs(version, xmlLayer) {
+    let supportedCrs = [];
+    const crsPropName = version === '1.3.0' ? 'CRS' : 'SRS';
+
+    for (const childElem of xmlLayer.children) {
```
gchoqueux: Searching for items by tag name is often used. It could be factoring.

Otherwise, it will not be easier to transform the xml object js?
I also found lib js that turns the capes into json.

autra:
> We need to valid the default value of layer.disableGetCap.

Ok let's validate it. What's your thoughts about it?

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -7,6 +7,78 @@
...
+                    childElem.getAttribute('maxy'));
+        }
+    }
+    if (xmlLayer.parentNode && xmlLayer.parentNode.tagName === 'Layer') {
+        return parseLayerBoundingBox(version, targetCrs, xmlLayer.parentNode);
+    }
+}
+
+function parseGeoBounds(version, xmlLayer) {
+    const geoTagName = version === '1.3.0' ? 'EX_GeographicBoundingBox' : 'LatLonBoundingBox';
```
gchoqueux: you could group in private functions 
```js
getCrsPNByVersion(version)...

getGeoPNByVersion(version)...
```

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -45,53 +117,155 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
...
+        }&FORMAT=${layer.format
+        }&TRANSPARENT=${layer.transparent
+        }&BBOX=%bbox` +
+        `&${crsPropName}=${layer.projection
+        }&WIDTH=${layer.width
+        }&HEIGHT=${layer.width}`;
+    });
+};
+
+WMS_Provider.prototype.findXmlLayer = function findXMLLayer(name, xml) {
```
gchoqueux: Could you add documentation on new functions?

src/Core/Scheduler/Providers/WMS_Provider.js
```diff
@@ -45,53 +117,155 @@ WMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer)
...
+            return n.parentNode;
+        }
+    }
+};
+
+WMS_Provider.prototype.parseSupportedCrs = function parseSupportedCrs(version, xmlLayer) {
+    return new Set(_parseSupportedCrs(version, xmlLayer));
+};
+
+WMS_Provider.prototype.parseSupportedFormats = function parseSupportedFormats(version, xmlCapa) {
```
gchoqueux: There's a difference between `xmlLayer` and `xmlCapa`

src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -21,18 +21,18 @@ WMTS_Provider.prototype.customUrl = function customUrl(layer, url, tilematrix, r
...
         const options = layer.options;
         options.version = options.version || '1.0.0';
         options.tileMatrixSet = options.tileMatrixSet || 'WGS84';
-        options.mimetype = options.mimetype || 'image/png';
         options.style = options.style || 'normal';
         options.projection = options.projection || 'EPSG:3857';
         let newBaseUrl = `${layer.url}` +
             `?LAYER=${options.name}` +
-            `&FORMAT=${options.mimetype}` +
+            `&FORMAT=${layer.format}` +
```
gchoqueux: `options.mimetype` to `layer.format` to be in another PR, both must be supported with a deprecated message

src/Core/View.js
```diff
@@ -296,6 +296,13 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         throw new Error(`${layer.protocol} is not a recognized protocol name.`);
     }
     layer = _preprocessLayer(this, layer, provider);
+    layer.whenReady.then(() => {
+        // if we still haven't any validExtent after getCapabilities, we assume
+        // it is the same as the configured extent
+        if (!layer.validExtent) {
+            layer.validExtent = layer.extent;
```
gchoqueux: and if `layer.extent` is undefined?

gchoqueux (changes requested):

src/Core/Geographic/Coordinates.js
autra: I don't see the point of this comment. This line is pretty self-explanatory.

src/Core/Scheduler/Providers/WMS_Provider.js
autra: You want a function for a one-liner?

src/Core/Scheduler/Providers/WMS_Provider.js
autra: Oups I forgot! yes will do.

src/Core/Scheduler/Providers/WMS_Provider.js
autra: Yes

src/Core/Geographic/Coordinates.js
gchoqueux: not for the WGS84

src/Core/Scheduler/Providers/WMS_Provider.js
gchoqueux: it's just to group the constants
you can suggest something else

src/Core/Scheduler/Providers/WMTS_Provider.js
autra: I can put that in another PR if you want, but I still don't see the need for a depreciation message. I did give my reasons, but you didn't give yours.

src/Core/View.js
autra: Normally it shouldn't happen: if disableGetCap is true, we check for it. Otherwise the `parseExtent` method has several fallbacks that cannot reasonably fail.

src/Core/Geographic/Coordinates.js
autra: I see what you mean, but I'm not sure this is the place to explain CRS.

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: I think it breaks the users cases

src/Core/Scheduler/Providers/WMS_Provider.js
autra: > Searching for items by tag name is often used. It could be factoring.

I've hesitated to do so, but each time it's a bit different. So I decided against it because:
- It will only abstract the for loop in the general case (the if part is always a bit different)
- it is less readable than the current code. And it's better to optimize for the reader of our code.

autra: > Otherwise, it will not be easier to transform the xml object js?

In my opinion, not worthy the extra lib. It won't save me the for loop anyway, and it adds a layer of abstraction that doesn't bring a lot of value here.

src/Core/Geographic/Coordinates.js
gchoqueux: I had to watch on the internet for explanations CRS:84, so it's a good place to explain to the developers. Why it makes sense to add this equivalence because we try to support wgs84 in itowns. It's better to be generous in commentary and it'sn't much to add

src/Core/Scheduler/Providers/WMTS_Provider.js
autra: Yes, but what about my suggestion in my previous comment? If it's clearly documented in the release notes, as it is a trivial change to do, user won't have any difficulty right? They would upgrade their itowns lib, and in the same commit rename this property. 

Anyway, I'll see if I can remove this from this PR.

src/Core/Geographic/Coordinates.js
autra: ok

src/Core/Scheduler/Providers/WMS_Provider.js
autra: To elaborate, as we already have everything we need to query and parse xml in js, I think we can do without. It will avoid growing itowns.js.

gchoqueux: ok

src/Core/Scheduler/Providers/WMS_Provider.js
autra: Oh I see. Yes, a function should be better than constants here I guess.

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: As it breaks the user cases, it's better, we have to support both and put a deprecated message

autra: Here it is: https://github.com/iTowns/itowns/pull/597

src/Core/Scheduler/Providers/WMS_Provider.js
autra: Actually I did better: I only expose them through a `_testing` export, which avoid polluting the prototype of the provider with private functions.

autra:
@gchoqueux updated.

This now depends on #597

autra:
@iTowns/reviewers updated and rebased :-)

examples/wfs.js
```diff
@@ -31,6 +31,7 @@ view.addLayer({
     networkOptions: { crossOrigin: 'anonymous' },
     type: 'color',
     protocol: 'wms',
+    disableGetCap: true, // their getCap call is too slow
```
peppsac: Can't this be implicit? If the user supplies all needed paramaters (format, version?) then you assume that the `GetCapabilities` query is not needed.
WDYT?

autra: It's an interesting idea. The advantage of `disableGetCap` is that it *explicitly* ensure getCap won't be called. If it's implicit, it can still be called if you forget something. Maybe we can cover both use cases ? if we have all the parameters, we disable it. If we have disableGetCap, we check the needed parameters and throw if one is missing. WDYT?

Also, WDYT about `disableGetCap` naming? The more I see it, the more I prefer `disableGetCapabilities`.

peppsac: Your suggestion seems fine.

And I prefer `disableGetCapabilities` (or `noGetCapabilities`?)

autra:
@peppsac updated

PR #596, opened 2017-12-08
example: add one with five different layers

zarov:
The goal is to provide a (not so ugly) usecase with sufficient layers to test the behavior of itowns with several layers, while charging and displaying them.

There also is a sixth layer commented, that can be added to push the number of texture further than 16, and thus test some issue such as #583 and #584.

Layers added are:

- Administrative delimitations (France)
- Roads (France)
- Railways (France)
- Cadastral plan (France)
- Denomination of locations (town, lake, mountain etc...)

multiple_layers

### DIFF
--- a/examples/layers/JSONLayers/Cada.json
+++ b/examples/layers/JSONLayers/Cada.json
@@ -1,0 +1,162 @@
+{
+  "type": "color",
+  "url": "https://wxs.ign.fr/an7nvfzojv5wa96dsga5nk8w/geoportail/wmts",
+  "protocol": "wmts",
+  "id": "Cadastre",
+  "title": "Parcelles cadastrales",
+  "visible": false,
+  "opacity": 1,
+  "transparent": true,
+  "updateStrategy": {
+    "type": "0",
+    "options": {}
+  },
+  "networkOptions": {
+    "crossOrigin": "omit"
+  },
+  "options": {
+    "tileMatrixSet": "PM",
+    "tileMatrixSetLimits": {
+      "0": {
+        "minTileRow": "0",
+        "maxTileRow": "1",
+        "minTileCol": "0",
+        "maxTileCol": "1"
+      },
+      "1": {
+        "minTileRow": "0",
+        "maxTileRow": "2",
+        "minTileCol": "0",
+        "maxTileCol": "2"
+      },
+      "2": {
+        "minTileRow": "0",
+        "maxTileRow": "4",
+        "minTileCol": "0",
+        "maxTileCol": "4"
+      },
+      "3": {
+        "minTileRow": "0",
+        "maxTileRow": "8",
+        "minTileCol": "0",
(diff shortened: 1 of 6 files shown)


autra (approved):
I'm fine with it, but I have a slight preference for having all the layers in `index.html`, with `.visible = false` for most of them. This won't add loading time, while still allowing us to test them when needed. I see the `index.html` as the "page for itowns developer".

WDYT?

autra:
To elaborate: this example doesn't add something new to already existing examples, that's why I'm advocating to put them in `index.html`.

zarov:
Yes, it is fine with me to remove this example and put the layers in the global `index.html`. I updated that.

autra:
Thanks! This Pr is already useful: it exhibits a bug in subdivision. I'm opening an issue after merging this.

examples/layers/JSONLayers/Cada.json
```diff
@@ -0,0 +1,164 @@
...
+    "name": "CADASTRALPARCELS.PARCELS",
+    "style": "bdparcellaire",
+    "zoom": {
+      "min": 5,
+      "max": 20
+    }
+  },
+  "version": "1.0.0",
+  "minScaleDenominator": 533.1823959624461,
+  "maxScaleDenominator": 17471320.75089743
```
autra: What are those 2 properties for? (They are not used in itowns)

examples/layers/JSONLayers/Denomination.json
```diff
@@ -0,0 +1,27 @@
+{
+  "type": "color",
+  "url": "https://wxs.ign.fr/an7nvfzojv5wa96dsga5nk8w/geoportail/wmts",
+  "protocol": "wmts",
+  "id": "Denomination",
+  "title": "Geographical denomination in France",
+  "visible": false,
```
autra, line 7: apparently this doesn't work with the current code :-( Let's remove these `visible: false`. Actually, because they were all visible, I discovered a bug, so maybe it's actually better to have them all! (sorry to backtrack on that)

autra (commented):
Some nits though:

examples/layers/JSONLayers/Cada.json
zarov: I honestly don't know, I just copy/paste them :D I'll remove them.

examples/layers/JSONLayers/Denomination.json
zarov: From what I've seen, they are not visible, but the dat.gui panel is not showing the correct status. What is the bug you discovered ? It sometimes doesn't load correctly the tiles for my part, but that's not always the case so I don't really know.

zarov: I pushed a fix for the GUITools in #598

autra:
LGTM :-)

zarov:
Thanks !

PR #597, opened 2017-12-11
refacto: Harmonize the mimetype configuration

autra:
## Description

Previously we had layer.format and layer.options.mimetype coexisting, with no difference in role. This commits removes layer.options.mimetype in favor of the former.

I'm not sure of which component to put in the commit title though.

### DIFF
--- a/src/Core/Scheduler/Providers/WMTS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMTS_Provider.js
@@ -10,6 +10,14 @@ import Extent from '../../Geographic/Extent';
 
 const coordTile = new Extent('WMTS:WGS84', 0, 0, 0);
 
+const supportedFormats = new Map([
+    ['image/png', getColorTextures],
+    ['image/jpg', getColorTextures],
+    ['image/jpeg', getColorTextures],
+    ['image/x-bil;bits=32', getXbilTexture],
+]);
+
+
 function customUrl(layer, url, tilematrix, row, col) {
     let urld = url.replace('%TILEMATRIX', tilematrix.toString());
     urld = urld.replace('%ROW', row.toString());
@@ -21,18 +29,21 @@ function customUrl(layer, url, tilematrix, row, col) {
 function preprocessDataLayer(layer) {
     layer.fx = layer.fx || 0.0;
 
-    layer.options = layer.options || {};
+    layer.format = layer.format || 'image/png';
+    if (!supportedFormats.has(layer.format)) {
+        throw new Error(
+            `Layer ${layer.name}: unsupported layer.format '${layer.format}', must be one of '${Array.from(supportedFormats.keys()).join('\', \'')}'`);
+    }
 
     if (layer.protocol === 'wmts') {
         const options = layer.options;
         options.version = options.version || '1.0.0';
         options.tileMatrixSet = options.tileMatrixSet || 'WGS84';
-        options.mimetype = options.mimetype || 'image/png';
         options.style = options.style || 'normal';
         options.projection = options.projection || 'EPSG:3857';
         let newBaseUrl = `${layer.url}` +
             `?LAYER=${options.name}` +
-            `&FORMAT=${options.mimetype}` +
+            `&FORMAT=${layer.format}` +
             '&SERVICE=WMTS' +
             `&VERSION=${options.version}` +
             '&REQUEST=GetTile' +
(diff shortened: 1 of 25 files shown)


autra:
@iTowns/reviewers please r?

zarov (commented):
I approve the merge of `format` and `mimeType`, but I'm quite confused in some case. For example, you have the format `geojson` in the `globe_wfs_extruded` example, while we are looking to `vector/geojson` later in the `Raster_Provider`. We should agree on a "standard" for the format, or make a list of available formats supported by iTowns. WDYT ?

autra:
Yes it is confusing. I just had a more detailed look, and what `Raster_Provider` does is very odd to me (I'm speaking about current master, anyway this PR does not change that). There is a `format` local  variable constructed from the extension of `layer.url`. This format is used to check a list of supported formats, *but* we check layer.options.mimetype is present (on line 76), only to override it later in *all* cases from the fetched file.... It is then used used to parse the fetched data.

The local variables `format` is actually only ever used to select the right method in Fetcher (which we could extrapolate from layer.format).

So my take: let's use a single source of truth:
- let's only ever use `layer.format` and extrapolate the `const format` (used to select the right fetch method) from it (should be easy enough). In other word, let's stop inferring a format from a url, which could be wrong anyway, or maybe *only* as a fallback if layer.format is not specified.
- let's trust the user and not try to overwrite the format from the received data: if they are wrong, the parsing code (from the `togeojson` lib) will fail, and that will be a valuable information for the user (allowing to detect errors in url for instance). I think it's better to warn the user about their configuration mistake than autocorrect them
- This will be consistent with `layer.format` everywhere else in itowns

This could be done here or in another PR, no strong opinion either way.

Pinging @gchoqueux on this one: WDYT about this proposal? Did I miss something about the current code?

>  make a list of available formats supported by iTowns. WDYT ?

This is actually provider/protocol dependant, so the current state (having a list per provider) is fine by me. We still haven't really solved the underlying problem of how to document our provider though...

gchoqueux:
**For the raster provider**

* if the user doesn't give the format. So `itowns` is trying to find the format by itself. first by url and confirms with the content of the file.

* If the user gives the format. So this format is used and returns an error if it doesn't match.

can be do in another PR

**Harmonize the mimetype**
it would be necessary to harmonize `geojson` (in wfs) and and `vector/geojson` (in raster layer)
May be remove `vector/`

peppsac:
Thanks @autra for working on this.

For `Rasterizer_Provider`: I'd remove completely the `layer.format` property.
It's only used for validation purpose but nothing guarantees that layer.format is really the format of the downloaded file.
Instead, it's easier to reject the command in `executeCommand` if the downloaded file has an invalid format.

For other formats, the `layer.format` content should match what is expected from the underlying protocol (so basically your PR seems to do the right thing).

autra:
@peppsac I did the work on rasterizer_provider in #665

autra:
@iTowns/reviewers updated

src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -20,19 +20,24 @@ function customUrl(layer, url, tilematrix, row, col) {
 
 function preprocessDataLayer(layer) {
     layer.fx = layer.fx || 0.0;
-
     layer.options = layer.options || {};
 
+    if (!layer.format && layer.options.mimetype) {
```
zarov: Couldn't it be moved in `View` in `addLayer` or `preprocessLayer` ? To avoid duplicates.

autra:
@zarov updated (this'll need squashing :-) )

src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -128,11 +128,11 @@ function executeCommand(command) {
         'image/x-bil;bits=32': getXbilTexture,
     };
 
-    const func = supportedFormats[layer.options.mimetype];
+    const func = supportedFormats[layer.format];
     if (func) {
         return func(tile, layer, command.targetLevel);
     } else {
-        return Promise.reject(new Error(`Unsupported mimetype ${layer.options.mimetype}`));
+        return Promise.reject(new Error(`Unsupported mimetype ${layer.format}`));
```
zarov: Last thing I can see, maybe change the mimetype word here too (present in WFS/Static/WMTS/WMS). Otherwise it is good to me.

src/Core/View.js
```diff
@@ -109,6 +109,13 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
         tmp.id = layer.id;
     }
 
+    // TODO remove it after next release
```
zarov: Add a release number or we will forget about it.

autra:
@zarov updated :-)

autra:
@zarov @peppsac updated. I took advantage of this PR to also use mimetypes in WFS_Provider.

zarov (approved):
This seems good to me, thanks for the work ! Before merging, **do not forget** to squash commits !

PR #598, opened 2017-12-12
fix (example): set correct value on init of GUITools

zarov:
When adding a layer to GUITools, values of properties were not respected, while being correct into itowns. Now we read the correct values on init.

Solves a problem met in #596


autra (approved):
Thanks for this!

PR #599, opened 2017-12-12
Example WFS on globe : Place points and bus line on the ground

gliegard:
Example WFS on globe : Place points and bus line on the ground

## Description
- example(wfs): Place route points on the ground, and buses lines on the ground. It allows the user to set specific altitude for each point of the line. So the bus line in Lyon follow the relief.
- refactor(wfs): Add contour coordinates to the callback that compute altitude.

## Screenshot
screenshot-pr-place-on-ground

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -40,6 +40,20 @@ function setMaterialLineWidth(result) {
     }
 }
 
+function altitudeLine(properties, contour) {
+    var altitudes = [];
+    var i = 0;
+    var alt = 0;
+    if (contour.length && contour.length > 0) {
+        for (; i < contour.length; i++) {
+            alt = itowns.DEMUtils.getElevationValueAt(globeView.wgs84TileLayer, contour[i]).z + 2;
+            altitudes.push(alt);
+        }
+        return altitudes;
+    }
+    return 0;
+}
+
 function colorLine(properties) {
     var rgb = properties.couleur.split(' ');
     return new itowns.THREE.Color(rgb[0] / 255, rgb[1] / 255, rgb[2] / 255);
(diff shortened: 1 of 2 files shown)


examples/globe_wfs_extruded.js
```diff
@@ -40,6 +40,20 @@ function setMaterialLineWidth(result) {
     }
 }
 
+function altitudeLine(properties, contour) {
+    var altitudes = [];
+    var i = 0;
```
zarov, line 45: Why put it here and not in the `for` loop ?

zarov (commented):
Good to me, nice to have paths following the terrain ! I think you could fuse `altitudeLine` and `altitudePoint` at some point, as a line is composed of points, but that may not really be pertinent here ;) 
Change the `i` (see comment) before approval though.

Also why put a `+ 5` in the altitude ? I don't see this much to be necessary.

examples/globe_wfs_extruded.js
gliegard: Because of the lint.
[eslint] All 'var' declarations must be at the top of the function scope. (vars-on-top)

zarov: Ah, yeah, the lint, I forgot about it. Thanks for the information.

gliegard:
@zarov, + 5 in the altitude was for the style ; I changed it to +2 :)

gliegard:
@itowns-reviewers

zarov (approved):
Looks good to me

peppsac:
Please change the commit scope from `example` to `examples` before merging.
Looks good otherwise.

gliegard:
ok, I've updated the commit message.

peppsac:
Thanks

PR #601, opened 2017-12-13
fix (examples): Wrong zoom min in layer example

### LINKED ISSUE #600: Some layers prevent subdivision on globe
## Your Environment

* Version used: current master
* Browser Name and version: Chromium and Firefox (latests) on Linux

## Context

The new denomination layer sometimes prevent subdivision on globe. Here is a gif of the process:
peek 2017-12-12 20-13

1.  `npm start` and load the root `index.html`
1.  check Layer globe > Show tiles
1.  zoom on paris 
1.  check and uncheck the Denomination layer

## Expected Behavior
 the subdivision doesn't change (because there are other layers with textures)

## Actual Behavior
When Denomination is active, it de-subdivises tiles.

gchoqueux:
remove error network WMTS

### DIFF
--- a/examples/layers/JSONLayers/Denomination.json
+++ b/examples/layers/JSONLayers/Denomination.json
@@ -20,7 +20,7 @@
     "name": "GEOGRAPHICALNAMES.NAMES",
     "style": "normal",
     "zoom": {
-      "min": 5,
+      "min": 6,
       "max": 20
     }
   }


gchoqueux:
~update:  to fix totally  #600, it needs PR #583~

autra:
>  it needs PR #583

Can you explain why? From the tile of #583, it is not obvious.

gchoqueux:
@autra because we exceed the max count color layers allowed in itowns

autra:
So this issue does not help for #600, right? (I'm not against it, just wondering)

gchoqueux:
@autra yes, it doesn't fix #600

autra (approved):
LGTM, thanks :-)

PR #602, opened 2017-12-13
feat(wfs): Add a debug UI for geometry layer

gliegard:
feat(wfs): Add a debug UI for geometry layer, to change visibility, opacity and toggle wireframe, material size and material linewith.

This change shows a new legend in the globe WFS extruded example, for each WFS layer (building, bus line and route points)

## Screenshots
screenshot-pr-legend

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -30,16 +30,6 @@ promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLaye
 promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
 promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
 
-function setMaterialLineWidth(result) {
-    var i = 0;
-    var mesh;
-    for (; i < result.children.length; i++) {
-        mesh = result.children[i];
-
-        mesh.material.linewidth = 5;
-    }
-}
-
 function altitudeLine(properties, contour) {
     var altitudes = [];
     var i = 0;
@@ -60,15 +50,16 @@ function colorLine(properties) {
 }
 
 globeView.addLayer({
+    type: 'geometry',
     update: itowns.FeatureProcessing.update,
     convert: itowns.Feature2Mesh.convert({
         color: colorLine,
         altitude: altitudeLine }),
-    onMeshCreated: setMaterialLineWidth,
+    linewidth: 5,
     url: 'https://download.data.grandlyon.com/wfs/rdata?',
     protocol: 'wfs',
     version: '2.0.0',
-    id: 'tcl_bus',
+    id: 'WFS Bus lines',
     typeName: 'tcl_sytral.tcllignebus',
     level: 14,
     projection: 'EPSG:3946',
@@ -116,7 +107,7 @@ globeView.addLayer({
     networkOptions: { crossOrigin: 'anonymous' },
     protocol: 'wfs',
     version: '2.0.0',
-    id: 'wfsBuilding',
+    id: 'WFS Buildings',
     typeName: 'BDTOPO_BDD_WLD_WGS84G:bati_remarquable,BDTOPO_BDD_WLD_WGS84G:bati_indifferencie,BDTOPO_BDD_WLD_WGS84G:bati_industriel',
     level: 14,
     projection: 'EPSG:4326',
@@ -126,14 +117,12 @@ globeView.addLayer({
     },
 }, globeView.tileLayer);
 
-
 function configPointMaterial(result) {
     var i = 0;
     var mesh;
     for (; i < result.children.length; i++) {
         mesh = result.children[i];
 
-        mesh.material.size = 5;
         mesh.material.sizeAttenuation = false;
     }
 }
@@ -159,13 +148,14 @@ globeView.addLayer({
     convert: itowns.Feature2Mesh.convert({
         altitude: altitudePoint,
         color: colorPoint }),
+    size: 5,
     onMeshCreated: configPointMaterial,
     filter: selectRoad,
     url: 'http://wxs.ign.fr/72hpsel8j8nhb5qgdh07gcyp/geoportail/wfs?',
     networkOptions: { crossOrigin: 'anonymous' },
     protocol: 'wfs',
     version: '2.0.0',
-    id: 'wfsPoint',
+    id: 'WFS Route points',
     typeName: 'BDPR_BDD_FXX_LAMB93_20170911:pr',
     level: 12,
     projection: 'EPSG:2154',
@@ -174,5 +164,6 @@ globeView.addLayer({
         mimetype: 'json',
     },
 }, globeView.tileLayer);
+
 exports.view = globeView;
 exports.initialPosition = positionOnGlobe;
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -170,16 +170,18 @@ function _preprocessLayer(view, layer, provider) {
             }
         };
         defineLayerProperty(layer, 'opacity', 1.0, () => {
-            layer.object3d.traverse((o) => {
-                if (o.layer !== layer.id) {
-                    return;
-                }
-                changeOpacity(o);
-                // 3dtiles layers store scenes in children's content property
-                if (o.content) {
-                    o.content.traverse(changeOpacity);
-                }
-            });
+            if (layer.object3d) {
+                layer.object3d.traverse((o) => {
+                    if (o.layer !== layer.id) {
+                        return;
+                    }
+                    changeOpacity(o);
+                    // 3dtiles layers store scenes in children's content property
+                    if (o.content) {
+                        o.content.traverse(changeOpacity);
+                    }
+                });
+            }
         });
     }
     return layer;
--- a/src/Process/FeatureProcessing.js
+++ b/src/Process/FeatureProcessing.js
@@ -34,6 +34,21 @@ export default {
         }
 
         const features = node.children.filter(n => n.layer == layer.id);
+        for (const feat of features) {
+            feat.traverse((o) => {
+                if (o.material) {
+                    o.material.transparent = layer.opacity < 1.0;
+                    o.material.opacity = layer.opacity;
+                    o.material.wireframe = layer.wireframe;
+                    if (layer.size) {
+                        o.material.size = layer.size;
+                    }
+                    if (layer.linewidth) {
+                        o.material.linewidth = layer.linewidth;
+                    }
+                }
+            });
+        }
         if (features.length > 0) {
             return features;
         }
@@ -68,7 +83,6 @@ export default {
                 if (layer.onMeshCreated) {
                     layer.onMeshCreated(result);
                 }
-
                 node.layerUpdateState[layer.id].success();
                 if (!node.parent) {
                     ObjectRemovalHelper.removeChildrenAndCleanupRecursively(layer.id, result);
--- a/utils/debug/GeometryDebug.js
+++ b/utils/debug/GeometryDebug.js
@@ -1,0 +1,24 @@
+export default {
+
+    addWireFrameCheckbox(gui, view, layer) {
+        layer.wireframe = layer.wireframe || false;
+        gui.add(layer, 'wireframe').name('Wireframe').onChange(() => view.notifyChange(true));
+    },
+
+    addMaterialSize(gui, view, layer, begin, end) {
+        layer.size = layer.size || 1;
+        gui.add(layer, 'size', begin, end).name('Size').onChange(() => view.notifyChange(true));
+    },
+
+    addMaterialLineWidth(gui, view, layer, begin, end) {
+        layer.linewidth = layer.linewidth || 1;
+        gui.add(layer, 'linewidth', begin, end).name('Line Width').onChange(() => view.notifyChange(true));
+    },
+
+    createGeometryDebugUI(datDebugTool, view, layer) {
+        const gui = datDebugTool.addFolder(`Layer ${layer.id}`);
+        gui.add(layer, 'visible').name('Visible').onChange(() => view.notifyChange(true));
+        gui.add(layer, 'opacity', 0, 1).name('Opacity').onChange(() => view.notifyChange(true));
+        return gui;
+    },
+};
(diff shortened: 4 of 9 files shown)


examples/globe_wfs_extruded.js
```diff
@@ -67,7 +68,10 @@ globeView.addLayer({
     options: {
         mimetype: 'geojson',
     },
-}, globeView.tileLayer);
+}, globeView.tileLayer).then(function _(layer) {
```
autra: The reason why the `GuiTools` was defined in the second `script` tag and why it was not used in the external js is to make the `test-examples` script possible. Ok, this examples doesn't run in our `test-examples` phase for now, but it's more because nobody took the effort to add it than a real will to exclude it. Let's keep that possibility and register the UI outside this file. For instance, something like

```js
for (const layer of globeView.getLayers()) {
    layer.whenReady.then( function _(layer) {
        // ...
    });
}
```

in the html file would have the same effect.

src/Core/View.js
```diff
@@ -170,16 +170,19 @@ function _preprocessLayer(view, layer, provider) {
...
-                if (o.layer !== layer.id) {
-                    return;
-                }
-                changeOpacity(o);
-                // 3dtiles layers store scenes in children's content property
-                if (o.content) {
-                    o.content.traverse(changeOpacity);
-                }
-            });
+            const object3d = layer.object3d || layer.parentLayer.object3d;
```
autra: I don't think it's a good idea. The concept of parentLayer is different than the THREE.js object hierarchy. Being a child layer only mean your update will be called with your parent's node after your parent update. You can choose to modify this node, or you can choose to add other node elsewhere. As we don't default childLayer.object3D to parentLayer.object3D when adding it, we shouldn't default to parentLayer.object3D here neither. If a child layer wants to add stuff to the same object3D, they can just use parentLayer.object3d during construction. 

What brought you to add this change? Let's see if we can find something else.

utils/debug/GeometryDebug.js
```diff
@@ -0,0 +1,56 @@
...
+                cb(object.material);
+            }
+        });
+        view.notifyChange(true);
+    },
+
+    addWireFrameCheckbox(gui, view, layer, value = false) {
+        layer.wireframe = value;
+        gui.add(layer, 'wireframe').name('Wireframe').onChange((newValue) => {
+            layer.wireframe = newValue;
```
autra: not needed, dat.gui does the `layer.wireframe = newValue` for you (see opacity for instance).

utils/debug/TileDebug.js
```diff
@@ -44,13 +36,7 @@ export default function createTileDebugUI(datDebugTool, view, layer, debugInstan
...
 
     // tiles wireframe
-    gui.add(layer, 'wireframe').name('Wireframe').onChange((newValue) => {
-        layer.wireframe = newValue;
-
-        applyToNodeFirstMaterial(view, layer.object3d, layer.id, (material) => {
-            material.wireframe = newValue;
-        });
-    });
+    GeometryDebug.addWireFrameCheckbox(gui, view, layer);
```
autra: maybe just change line 9 into `const gui = createGeometryDebugUI(...)`, and remove common part from this function. After all, a Tile is a Geometry.

utils/debug/GeometryDebug.js
```diff
@@ -0,0 +1,56 @@
...
+
+            this.applyToNodeFirstMaterial(view, layer.object3d || layer.parentLayer.object3d, layer.id, (material) => {
```
autra: as said earlier, not convinced by the default to parentLayer.object3D. What problem are you trying to solve?

utils/debug/GeometryDebug.js
```diff
@@ -0,0 +1,56 @@
...
+            layer.size = newValue;
+
+            this.applyToNodeFirstMaterial(view, layer.object3d || layer.parentLayer.object3d, layer.id, (material) => {
+                material.size = newValue;
+            });
+        });
+    },
+
+    addMaterialLineWidth(gui, view, layer, begin, end, value = 1) {
+        layer.linewidth = value;
```
autra: this could be a config value (useful in itself, not only for debugging), so be in globe_wfs_{...}.js. This would save one parameter (we have 6 params here) for this function. dat.gui will nicely set the right value when you do `gui.add(layer, 'property', ...)`.

utils/debug/GeometryDebug.js
```diff
@@ -0,0 +1,56 @@
...
+            layer.wireframe = newValue;
+
+            this.applyToNodeFirstMaterial(view, layer.object3d || layer.parentLayer.object3d, layer.id, (material) => {
+                material.wireframe = newValue;
+            });
+        });
+    },
+
+    addMaterialSize(gui, view, layer, begin, end, value = 1) {
+        layer.size = value;
```
autra: See my comments for `addMaterialLineWidth`

utils/debug/GeometryDebug.js
autra: Same think for wireframe actually. If people wants it to true, they'll do a layer.wireframe =  true before initializing debug UI. Let's dat.gui pick up whatever value it finds in layer.

autra (changes requested):
Already looking good :-) Debugging tools are always very useful, thanks!

utils/debug/GeometryDebug.js
gliegard: The problem is that my WFS layer doesn't have any object3d, every meshes are added to the globe tile layer.

src/Core/View.js
gliegard: The problem is that my WFS layer doesn't have any object3d, every meshes are added to the globe tile layer.

examples/globe_wfs_extruded.js
gliegard: ok

src/Core/View.js
gchoqueux: Apparently it's changing the opacity of the layer

What could be when attaching a geometric layer:

```js
var layer = new itowns.GeometryLayer('3d-tiles-discrete-lod', globeView.tileLayer.object3d);
globeView.addLayer({
    object3d: globeView.tileLayer.object3d,
}, globeView.tileLayer);
```

or in `View.prototype.addLayer`

```js
if(!layer.object3d) {
   layer.object3d = parentLayer.object3d
}
```

autra: Oh I see. I don't have anything better to propose for now, so let's keep it like that.

gchoqueux:
nice work!
Geometry debug tools can be use in 3d tiles examples?

utils/debug/GeometryDebug.js
gliegard: No, it's needed. Opacity is a in the default property of a layer, but not wireframe.

autra: so line 12 is needed somewhere, but not line 14 right?

src/Core/View.js
autra: But for this reason, that's why I think we should work more with groups, because most of the time, we change opacity/wireframe stuff and not layer.object3d transformations.

utils/debug/GeometryDebug.js
gliegard: The problem is if you forget to add the layer.linewidth , you get an exception.
If I do what you propose, when you add the debug panel, you need to add the layer.linewith default value in another place.

You can see in TileDebug line 19-20, showOutline and wireframe default value are added in the TileDebug object, because it's needed for this object.

WDYT ?

utils/debug/TileDebug.js
gliegard: Yep, good idea

utils/debug/GeometryDebug.js
gliegard: yes line 12 is needed somewhere.

line 14 is needed too. If I remove it, it's doesn't work. You can see in View .js line 172, opacity is defined and managed as a property, but not wireframe.

gliegard:
@gchoqueux Good idea, I have refactored 3DTileDebug to use GeometryDebug

utils/debug/GeometryDebug.js
autra: I don't know how you tested, but just replace line 14 with 
```js
console.log(layer.wireframe)
```
and see :-)

autra: For reference https://github.com/dataarts/dat.gui/blob/05c85ecf7d2ef5150eed53eebad83f499a9d3bf0/src/dat/controllers/BooleanController.js#L35 and https://github.com/dataarts/dat.gui/blob/master/src/dat/controllers/Controller.js#L89

gliegard: ok, you're right ! (I don't know how I have tested.. my bad)

gliegard:
updated

utils/debug/TileDebug.js
gliegard: I have changed line 9 into createGeometryDebugUI ...

But I still need to add GeometryDebug.addWireFrameCheckbox...
Because wireframe is not in the createGeometryDebugUI, since wireframe doesn't make sense for points or lines.

src/Core/View.js
autra: actually: this code works with the assumption that when `layer.object3d` is null, this layer only attach things to parentLayer.object3d. This is not true in the general case. Therefore, we should not default to `parentLayer.object3d` here, but to `view.scene` instead.

AND we should also do, in WFS_Provider.preprocessDataLayer:
```js
layer.object3d = layer.object3d || parentLayer.object3d;
```

because it's WFS_Processing that add the resulted WFS object to the current node. The fact that the layer object3d is the same as its parent object3d is only true for WFS layers, so it makes sense that WFS preprocessing handle that. (you'll have to move `layer.parentLayer = parentLayer` in View.prototype.addLayer before preprocessing for this to work)

With this 2 changes:
1. WFS layers would be correctly configured (preprocessing phase will correctly init layers for the update method of WFS)
1. the opacity property will work in all cases

N°1 is necessary for this PR, 2 is not (but as you're going to touch this line anyway...)

src/Core/View.js
```diff
@@ -170,16 +170,19 @@ function _preprocessLayer(view, layer, provider) {
...
+            if (object3d) {
```
autra: you won't need this test any more

utils/debug/GeometryDebug.js
```diff
@@ -0,0 +1,44 @@
...
+                cb(object.material);
+            }
+        });
+        view.notifyChange(true);
+    },
+
+    addWireFrameCheckbox(gui, view, layer) {
+        layer.wireframe = layer.wireframe || false;
+        gui.add(layer, 'wireframe').name('Wireframe').onChange((newValue) => {
+            this.applyToNodeFirstMaterial(view, layer.object3d || layer.parentLayer.object3d, layer.id, (material) => {
```
autra: So same thing here, we should default to scene.

autra (changes requested):
I finally made up my mind about object3d :-) The rest is looking good

gliegard:
Ok, I remove the parentLayer trick, and I follow the view.scene approach...
Hope it's ok !!

gliegard:
@autra : I've updated the code. Now everything work together.

I put the 
`layer.object = layer.parentLayer.object3d ` 
in the WFS Provider.

And I use the `scene.view` for the general case.

src/Process/FeatureProcessing.js
```diff
@@ -68,7 +68,19 @@ export default {
                 if (layer.onMeshCreated) {
                     layer.onMeshCreated(result);
                 }
-
+                result.traverse((o) => {
+                    if (o.material) {
```
autra: Actually, what about putting this update code above, for example somewhere after line 36 (by looping on `features`). After all, the update function is made for this kind of things:
- all these properties would automatically get updated at each frame
- we won't need to bother about parentLayer in View.js (basically, we won't need any change to View.js in this PR any more)
- we won't need to set layer.object3d to parentLayer.object3d in WFS_Provider.

The reasons why I think it's better:
- assigning `parentLayer.object3d` to `layer.object3d` breaks the assumption that a layer's object3d is its *own* object, with which it can do whatever it wants. I'd like to keep that assumption for layers that do have an object3d. So no object3d actually gives the correct information in this case: wfs layers cannot act on a single object3d to reach all its node.
- itowns already provides a mechanism that allows you to traverse each node of a layer: it's precisely the layer `update` function. This mechanism is already battle-tested, and made to be efficient (we don't do it more than 1 time per frame for example), and you actually launch it by calling `notifyChange` anyway. Here, if you attach the opacity to a slider, it would guarantee that we won't traverse the object tree more than one time per frame (click and input events can trigger a lot faster).

Sorry, I should have started with this, but I didn't realize it at first.

utils/debug/GeometryDebug.js
```diff
@@ -0,0 +1,44 @@
+export default {
+    applyToNodeFirstMaterial(view, root, layerId, cb) {
```
autra: and you won't need this function any more. Or maybe only for wireframe, though I'm fine with having `feature.material.wireframe = layer.wireframe` in WFS_Processing.update. This is a useful feature for client apps too.

autra (changes requested):

gliegard:
Thank for yours suggestions, it's elegant !! I've implemented it.

src/Process/FeatureProcessing.js
```diff
@@ -34,6 +34,21 @@ export default {
         }
 
         const features = node.children.filter(n => n.layer == layer.id);
+        for (const feat of features) {
```
gchoqueux, line 37: I don't agree too much, because we traverse even if no property has changed

gchoqueux: may be add a `layer._propertyChange` to notify change

peppsac: I believe it's good enough for now: it does the job and the perf hit from traversing a few nodes each frame will be unnoticable.

autra: I tend to agree with @peppsac for that: traversal cost for such simple operations are very cheap. Of course, profiling can prove us wrong.

src/Core/View.js
```diff
@@ -170,16 +170,18 @@ function _preprocessLayer(view, layer, provider) {
...
-                if (o.layer !== layer.id) {
-                    return;
-                }
-                changeOpacity(o);
-                // 3dtiles layers store scenes in children's content property
-                if (o.content) {
-                    o.content.traverse(changeOpacity);
-                }
-            });
+            if (layer.object3d) {
```
autra, line 173: note for the future: so I guess we should revisit this part to do it in 3dtiles processing?

autra (approved):
Looks good to me in the current form.

Of course, we may want to revisit some later: one way of avoiding too much traversal would be to work on exploiting better the  changedSources parameter of `View.notifyChange` (currently, there's no way of saying "hey, I just changed the opacity of a child layer, don't update the parent layer or other layers"). I think it would decrease the number of visited elements a lot in this case. @gchoqueux is that ok for you?

gchoqueux:
We agree that this can be improved. but it's okay for now, it should add a comment and a issue

autra:
Thanks @gliegard !

gliegard:
Thanks !

PR #603, opened 2017-12-14
fix (protocols): stop parsing unnecessarily xbil buffer

gchoqueux:
The buffer was parsed (`computeMinMaxElevation`) twice in the `IoDriver_XBIL.prototype.read` and `WMTS_Provider.prototype.getXbilTexture`

### DIFF
--- a/src/Core/Scheduler/Providers/WMTS_Provider.js
+++ b/src/Core/Scheduler/Providers/WMTS_Provider.js
@@ -5,7 +5,7 @@
  */
 
 import * as THREE from 'three';
-import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from './OGCWebServiceHelper';
+import OGCWebServiceHelper from './OGCWebServiceHelper';
 
 function WMTS_Provider() {
 }
@@ -85,18 +85,12 @@ WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, ta
     const url = this.url(coordWMTS, layer);
 
     return OGCWebServiceHelper.getXBilTextureByUrl(url, layer.networkOptions).then((texture) => {
-        const { min, max } = OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
-            texture.image.data,
-            SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
-            pitch);
-
         texture.coords = coordWMTS;
-
         return {
             texture,
             pitch,
-            min: min === undefined ? 0 : min,
-            max: max === undefined ? 0 : max,
+            min: !texture.min ? 0 : texture.min,
+            max: !texture.max ? 0 : texture.max,
         };
     });
 };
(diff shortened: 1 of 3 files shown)


src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -85,18 +85,19 @@ WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, ta
     const url = this.url(coordWMTS, layer);
 
     return OGCWebServiceHelper.getXBilTextureByUrl(url, layer.networkOptions).then((texture) => {
-        const { min, max } = OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
-            texture.image.data,
-            SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
-            pitch);
+        const { min, max } = (texture.min == undefined && texture.max == undefined) ? texture :
+            OGCWebServiceHelper.ioDXBIL.computeMinMaxElevation(
```
autra: if `getXBilTextureByUrl` call `io_driverxbil.read` that calls `parseXbil` that calls `computeMinMaxElevation`, what's the point of this call? It is guaranteed we have already done it. We could simply remove it.

src/Core/Scheduler/Providers/WMTS_Provider.js
```diff
@@ -85,18 +85,19 @@ WMTS_Provider.prototype.getXbilTexture = function getXbilTexture(tile, layer, ta
...
 
         texture.coords = coordWMTS;
 
         return {
             texture,
             pitch,
-            min: min === undefined ? 0 : min,
-            max: max === undefined ? 0 : max,
+            min: !min ? 0 : min,
+            max: !max ? 0 : max,
```
autra: If there is no value at this point, it means there is a problem with the array (there is no buffer basically). Which means that returning 0 in min and max is not a solution. Imo this is an error if it happens, and we should throw somewhere (in parseXBil maybe). WDYT?

autra (changes requested):

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: yes

src/Core/Scheduler/Providers/WMTS_Provider.js
gchoqueux: It should be studied why this case happens and what is the impact itowns. In any case it isn't the scope of the PR

autra (approved):

autra:
LGTM! Sorry for the delay.

 The commit message is now just a "fix(protocols): Remove duplicate `computeMinMaxElevation` call". (notice there is no space between fix and the (, this is the format of conventional-changelog. This would allow us to auto-generate changelogs)

After this little edit, this can be landed, thanks!

gchoqueux:
there's already the space

PR #604, opened 2017-12-14
New jsdoc template and new home page for the API documentation

vcoindet:
## Description

I added a new template for jsdoc
I created a home page for the API documentation

## Motivation and Context

Found the default template of jsdoc a bit sad and wanted to add a more official home page to the API documentation. (see the screenshot)

jsdoc


zarov:
Really nice addition, way easier to read imho. What's the point of the `JSDOC.md` file ? If it supposed to be the homepage, I think the `README` as homepage is quite good (but may require some change in IGN Oslandia and AtolCD log that are too low res ;) )

vcoindet:
The idea of the `JSDOC.md` file was to put something else than the `README` with a globe you can manipulate on the homepage (of course you can't see it on a screenshot, but the globe here is a iframe of JSFiddle). If you prefer i can put the `README` instead.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "lib/Main.js",
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
-    "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js src/Core/Scheduler/Providers/GpxUtils.js src/Core/MainLoop.js",
+    "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js src/Core/Scheduler/Providers/GpxUtils.js src/Core/MainLoop.js -t node_modules/docdash --readme README.md",
```
autra: I thing your intention was `--readme JSDOC.md` right?

autra (commented):
I quite like this theme, it's indeed better than our current one, *except* the purple in each api pages. Can we modify that to a color in better harmony with our website?

autra:
same thing for the links btw

package.json
vcoindet: Yes of course, my bad !

vcoindet:
I'll see what I can do about the color ! ;)

vcoindet:
I made some changes : the color indeed, and some other stuff to be in better harmony with our website.
I'm not sur about : 
`"include": [
   "../../doc.css"
]`
in jsdoc-config.json. @autra : can you confirm ?

Home page : 
jsdoc1

Other pages : 
jsdoc2

zarov:
`"include": [ "../../doc.css" ]` is not good when run from the root of the project. In my case it is the equivalent of doing `$HOME/doc.css`. I can't seem to make it work when changing it though.

vcoindet:
Thank @zarov for your answer, that's what I wanted to know.
I found something else : 
`"source": {
                "include": [
                    "doc.css"
                ]
            }`
This should work fine now.

zarov:
I get a 404 on the css file unfortunately.

vcoindet:
@zarov : I pushed a new modification, tell me if it works better !

zarov:
Yes it works, thanks !

JSDOC.md
```diff
@@ -0,0 +1,18 @@
+<!--p align="center">
+<img src="https://raw.githubusercontent.com/iTowns/itowns.github.io/master/images/itowns_logo_300x134.png" />
+</p-->
+
+<h1 align="center">
+API DOCUMENTATION
+</h1>
+<iframe width="100%" height="500" src="//jsfiddle.net/iTownsIGN/zrdgzz26/embedded/result/" allowpaymentrequest allowfullscreen="allowfullscreen" frameborder="0"></iframe>
```
autra: Can you start the `src` property with http? This would make the jsfiddle work locally.

autra: Do we need the `allowpaymentrequest` parameter? And actually, I think it'd be better to embed `<script async src="//jsfiddle.net/iTownsIGN/zrdgzz26/embed/js,html,result/"></script>` instead. After all, our visitors would be devs here :8)

JSDOC.md
```diff
@@ -0,0 +1,18 @@
...
+
+
+
+<p style="display:flex; justify-content:space-around;" align="center">
```
autra: as we are including our own CSS, this page style could be in it. As you wish.

layout.tmpl
```diff
@@ -0,0 +1,41 @@
...
+
+    <script src="scripts/prettify/prettify.js"> </script>
+    <script src="scripts/prettify/lang-css.js"> </script>
+    <!--[if lt IE 9]>
+      <script src="//html5shiv.googlecode.com/svn/trunk/html5.js"></script>
+    <![endif]-->
+    
+    <link type="text/css" rel="stylesheet" href="styles/jsdoc.css">
+    <link type="text/css" rel="stylesheet" href="styles/prettify.css">
+    <link type="text/css" rel="stylesheet" href="../doc.css">
```
autra: This doesn't work in the current form. `doc.css` never get copied to out/API_Doc. You need 3 things:

1. change jsdoc-config.json to: 
```js
{
    "templates": {
        "default": {
            "layoutFile": "layout.tmpl",
            "staticFiles": {
                "paths": ["static_jsdoc"]
           }
        }
    }
} 
 
```
2. move `./doc.css` into a subfolder of itowns, like `static_jsdoc` for instance (see 1.)

1 and 2 are needed because of  what happens here: https://github.com/clenemt/docdash/blob/master/publish.js#L503. This code fails for single files directly on the root folder of the git repo: on line 518, sourcePath would be `.`, outDir is whatever outDir you gave, and filename would be doc.css, so `filename.replace(...)` yields something like docout/ if outdir is out... (it's a bug in jsdoc/docdash imho)

3. change the href here to `./doc.css`

doc.css
```diff
@@ -0,0 +1,66 @@
+h4.name {
```
autra: I noticed the "home" link on top of the navbar is still purple.

Otherwise, very nice color, a lot better, thanks!

doc.css
```diff
@@ -0,0 +1,66 @@
...
+  background-color: rgba(86, 134, 172, 1);
+}
+
+nav > ul > li > a {
+  
+  color: #09085F;
+}
+
+nav > ul > li > ul > li > a {
+  color: rgba(86, 134, 172, 1)
```
autra: nits: missing a semicolon here, and 2 more later in the files (lines 53, 57).

autra (changes requested):

vcoindet:
Thanks for your review @autra ! WDYT now ?

autra (approved):
That's *a lot* nicer with your template @vcoindet, thanks!

autra:
@vcoindet actually we forgot to amend `autodeploy.sh`!!! Sorry for that, I completely forgot to check during my review. So now the style is broken on www.itowns-project.org :-(

Do you want to open another PR to fix it? Actually, now that we have a config file, I think it's better if every argument we currently give to jsdoc in `npm run doc` is now in this config file.

autra:
Ok, actually everything is fine :-D sorry for the false alarm!

PR #606, opened 2017-12-18
fix (renderer): set DEBUG after defines definition

zarov:
`DEBUG` was set before defines definition (in case of depth buffer supported), so it was always overwritten.


autra (approved):
:-O Thanks!

PR #607, opened 2017-12-18
examples: add anaglyph and parallax effects

zarov:
## Description

Add two effects to display the scene in 3D. The first one is the Anaglyph, heavily copied from https://github.com/mrdoob/three.js/blob/dev/examples/js/effects/AnaglyphEffect.js and the second one is the Parallax, heavily copied from https://github.com/mrdoob/three.js/blob/dev/examples/js/effects/ParallaxBarrierEffect.js

To enable these effects, three new methods have been added to View:
- `enableAnaglyph`, for the first effect
- `enableParallax`, for the second
- `disableEffect`, for turning off an effect

Before merging, there is one thing that (may) need to change: where do we put these functions ? FOr now they are in `View`, but I don't feel it's the right location. WDYT ?

## Motivation and Context

I have a 3D screen that doesn't display 3D enough ;)

### Anaglyph effect
screenshot from 2017-12-18 16-55-31
### Parallax effect
screenshot from 2017-12-18 16-55-53


peppsac:
This demo is a nice addition!

But, IMHO, this PR should only add the example; all the core stuff added by this PR doesn't belong to iTowns.

iTowns must allow user to easily implement this but shouldn't have any effects nor special 3d features built-in.

zarov:
Yeah I thought that after making this PR, viewing the `+524 −2` lines. I'm good to do it as an example.

zarov:
Updated

examples/ParallaxBarrierEffect.js
```diff
@@ -0,0 +1,115 @@
+// This file is copy pasted from itowns.THREE
```
autra: I think you can include it from node_modules/three/... directly in your html. We'll have to check that it'll work in `gh_pages`. Maybe `autodeploy.sh` will need some update.

autra: (I mean, I'm sure we don't need the copy-paste)

zarov: We need the copy-paste, there are some changes below (see  https://github.com/iTowns/itowns/pull/607/files#diff-5715fdf0d63640d34276838bfcde4429R43)

peppsac: I think the far values could be overwritten here: https://github.com/iTowns/itowns/pull/607/files#diff-8f232072bea770740260eca3c875177eR51

zarov: I don't see how we could access the `_stereo` var.  Plus there is the `dispose` method at the end, that also can't be added like that, as we have no direct access to `_renderTargetL/R`.

autra: @peppsac I thought so too (because `render` calls `_stereo.update`), but the far value seem not to be taken fully into account here: https://github.com/mrdoob/three.js/blob/master/src/cameras/StereoCamera.js#L36

(looks like a bug in THREE for me)

zarov:
I added the Stereo effect, usable for 3D ready screen only (and maybe cardboard with a phone, only if a gyroscope controller is added to the example)

screenshot from 2017-12-21 10-32-19

peppsac (approved):
@zarov I can't test since I don't own the necessary hardware :)

As already said before, it's a bit annoying that we can't use three.js effect unmodified but if there's no other way I'm fine with merging as is.

zarov:
Yes I agree. Fortunately, @gchoqueux found a way around this, which I pushed in the last update !

gchoqueux:
@peppsac wdyt about the last update?

peppsac:
Looks better indeed, nice!

PR #608, opened 2017-12-19
fix(wfs): use bigger integer data type for indices array

gliegard:
## Description
fix(wfs): use bigger integer data type for indices array

## Motivation and Context
Prior to this change, there was a bug, visible in the wfs globe example :
if you we set buse line WFS layer Level to 10 or less, you will see bugged lines.
The bug comes from the big size of the created LineSegments.

screenshot-lines-bug

screenshot-lines-ok

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -70,7 +70,7 @@ globeView.addLayer({
     version: '2.0.0',
     id: 'tcl_bus',
     typeName: 'tcl_sytral.tcllignebus',
-    level: 14,
+    level: 9,
     projection: 'EPSG:3946',
     extent: {
         west: 1822174.60,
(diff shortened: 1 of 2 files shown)


peppsac:
Performance wise it's preferable to have a single object than many objects.

Can you elaborate on the original bug? Maybe the vertices count was too big for the indices array (which use 16 bit ushort)?

And btw, if you modify line drawing within WFS it could be a good idea to switch to `MeshLine` (see GpxUtils.js)

gliegard:
You're right ! use 32 bit array instead of 16 bit array for indices is solving the problem.

gliegard:
You're right, changing 16 bit ushort to 32 bit ushort for indices array solve the problem.
I've removed my changes to do this.

I'll look at MeshLine later.

peppsac (approved):
You could have used Uint32Array onnly when needed but that doesn't matter much.

Thanks for the fix.

PR #609, opened 2017-12-21
feat(3dtiles): Add wireframe checkbox to 3dTilesDebugUI

gliegard:
## Description
feat(3dtiles): Add wireframe checkbox to 3dTilesDebugUI

## Screenshots
screenshot-3dtiles-wireframe

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -309,6 +309,14 @@ export function process3dTilesNode(cullingTest, subdivisionTest) {
                     markForDeletion(layer, n);
                 }
             }
+            // toggle wireframe
+            if (node.content && node.content.visible) {
+                node.content.traverse((o) => {
+                    if (o.material) {
+                        o.material.wireframe = layer.wireframe;
+                    }
+                });
+            }
             return returnValue;
         }
 
--- a/utils/debug/3dTilesDebug.js
+++ b/utils/debug/3dTilesDebug.js
@@ -6,6 +6,9 @@ import GeometryDebug from './GeometryDebug';
 export default function create3dTilesDebugUI(datDebugTool, view, layer) {
     const gui = GeometryDebug.createGeometryDebugUI(datDebugTool, view, layer);
 
+    // add wireframe
+    GeometryDebug.addWireFrameCheckbox(gui, view, layer);
+
     // Bounding box control
     const obb_layer_id = `${layer.id}_obb_debug`;
 


autra:
It seems that the higher level are not "wireframed" correctly (you can see that by playing with high sse threshold, or looking at the dragon from far away). Also, possibly related, in some part of the code, we also modify `node.content` (see View.js:180 for instance, or `setDisplayed` in 3dTilesProcessing. Intuitively, I would think you need to modify those too...

gliegard:
I move the code to setDisplayed ; and now the higher level is "wiframed" too.

I don't know if it's the best way to do it...

autra:
I think I get it. If I'm not mistaken, it works now because you traverse node.content for each node, and not `resultValue` (not because you moved it to `setDisplayed`). 

The way 3dTiles is currently implemented in itowns: it is composed of "tiles" (the `node` variable here, mainly for subdivision purpose, if I understand correctly) which will have `.layer` set to `layer.id`. These tiles does not have any actual geometry, because one "tile" can actually be a full THREE.js scene (see the 3dTiles specification). In fact the meshes that composed this tile are in `node.content` (if you debug, you'll see that some (all?) `node.content` are THREE.Scene objects). Of course, to appear in the screen, they are *also* added to `node.children` but won't be contained in `resultValue` (because we filter on `node.layer`). 

So in short : `_process3dTilesNodes` (the `update` function for 3dTiles layer), when called by mainLoop, iterates on a virtual tree hierarchy composed of these "tiles". Each tile contains a full object hierarchy that get displayed or not, depending of the SSE calculation.

Therefore, your second traversal version is correct: for each call of update, we need to set wireframe for one tile, so we traverse `node.content`. The children tiles will be treated the same way when `_process3dTilesNodes` will be called on them. Hope my explanations are clear :-)

That being said, I indeed don't believe `setDisplayed` is a good place for it. It will be better at the place where you put your original attempt, so just before `return returnValue`, surrounded with `if (node.content.visible)` to achieve the same effect as `if (display)` in `setDisplayed`. 

(ping @Jeremy-Gaillard to eventually correct my description of the 3dTiles process)

Jeremy-Gaillard:
Exactly @autra.

I'll just insist on the fact that you shouldn't modify `returnValue`, it contains the tiles that will be updated at the next step of the main loop.

autra:
> I'll just insist on the fact that you shouldn't modify returnValue

you mean we shouldn't touch either the array or the elements inside, that's it?

autra:
Oh sorry, misclick!

Jeremy-Gaillard:
> you mean we shouldn't touch either the array or the elements inside, that's it?

Yes.

gliegard:
I've updated the code.

autra (approved):
Perfect!

autra:
One last thing: the commit message speaks only about the debug checkbox, but most importantly: we added support for `layer.wireframe` for 3dTiles layer. Could you amend this message with something like

> feat(3dtiles): add wireframe support

?

gliegard:
Yep, I've updated the commit message !

autra:
Thanks!

A little reminder for the future: we try to keep the first line of the commit message below 65 chars. This is to have it displayed entirely in `git log --oneline`, on github and various tools (you have plugins for the most common text editors/IDE warning you when this limit is reached). Then a blank line, and the body, which can be arbitrary long. In this case it would have been:

```
feat(3dtiles): add wireframe support for 3dTiles layer

Also add a wireframe checkbox to 3dTilesDebugUI etc.
```

In the long run it really improves the readability of git logs :-)

gliegard:
Thanks, and thanks for 65 chars reminder, I didn't know !

And merry christmas :)

PR #610, opened 2017-12-22
Improve elevation layers handling with StaticProvider

peppsac:
## Description
2 commits fixing elevation layer processing.

The first one reorganize the code a bit to avoid needless work.
The second fix StaticProvider to use the proper download function.

### DIFF
--- a/src/Core/Scheduler/Providers/StaticProvider.js
+++ b/src/Core/Scheduler/Providers/StaticProvider.js
@@ -23,6 +23,11 @@ function selectBestImageForExtent(images, extent) {
     return selection;
 }
 
+function buildUrl(layer, image) {
+    return layer.url.href.substr(0, layer.url.href.lastIndexOf('/') + 1)
+        + image;
+}
+
 function getTexture(tile, layer) {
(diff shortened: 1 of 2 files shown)


zarov:
With the `mimetype` doomed to disappear, it may be a better option to use `layer.format` instead, no ?
Thanks for this cleaning anyway !

peppsac:
Can someone review this short PR please? Thanks!

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -95,17 +102,29 @@ export default {
                     extent,
                 });
             }
+        }).then(() => {
+            if (!layer.options.mimetype) {
+                // fetch the 1st image
```
autra: `// fetch the first image to detect mimetype` ?

autra (approved):
LGTM, one minor comment if you care to fix :-)

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Fixed.

gchoqueux (approved):

PR #611, opened 2018-01-04
fix (controls): add an offset after mouse movement

### LINKED ISSUE #545: Rework GlobeControls

zarov:
When moving the globe using `GlobeControls`, there was some kind of offset, the globe not staying right under the cursor. This is visible if the container of the canvas is not occupying the full page, the container has a CSS position property different of the default `static` and there are elements right before it on the page, like the geoportal case.

The solution to avoid this, is to detect the position of the container and add an offset according to this position, using the `getBoundingClientRect` method of the container.

@elias75015 could you test this fix ?

Note: as I tweaked the `GlobeControls`, I can confirm that we need to resolve issue #545, as there were some other bugs as well.

### DIFF
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -904,8 +904,13 @@ function GlobeControls(view, target, radius, options = {}) {
 
             event.preventDefault();
 
+            const staticPos = window.getComputedStyle(event.target.parentElement).position !== 'static';
+            const bounds = staticPos ? event.target.getBoundingClientRect() : { left: 0, top: 0 };
+            const x = event.clientX - event.target.offsetLeft - bounds.left;
+            const y = event.clientY - event.target.offsetTop - bounds.top;
+
             if (state === this.states.ORBIT || state === this.states.PANORAMIC) {
-                rotateEnd.set(event.clientX - event.target.offsetLeft, event.clientY - event.target.offsetTop);
+                rotateEnd.set(x, y);
                 rotateDelta.subVectors(rotateEnd, rotateStart);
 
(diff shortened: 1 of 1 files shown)


src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -904,8 +904,13 @@ function GlobeControls(view, target, radius, options = {}) {
 
             event.preventDefault();
 
+            const staticPos = window.getComputedStyle(event.target.parentElement).position !== 'static';
```
gchoqueux, line 907: could you factorize this part that appears several times?

gchoqueux (commented):
Nice, thanks

src/Renderer/ThreeExtended/GlobeControls.js
zarov: Nope, it depends on the `event`, and this variable is available in the function scope only. As `GlobeControls` is a sort of singleton (correct me if I'm wrong), we can't compute the value at the init of the class.

gchoqueux: Ok, you could pass the event as a parameter.
In short, it isn't important.

gchoqueux:
@elias75015 can you test if the bug is fixed?

pjjmunier:
The following bugs have been corrected :
- inconsistency of the coords picked in 3D with the mouse
- appearance of an offset between the cursor and the globe by moving the globe with the mouse

gchoqueux (approved):

PR #612, opened 2018-01-08
fix(core): marked dependency is insecure version

gchoqueux:
In jsdoc, marked dependency is insecure version (0.3.6)
A issue propose a temporary solution: to add an explicit dependency to marked ~0.3.9
[https://github.com/jsdoc3/jsdoc/issues/1489#issuecomment-355521471](https://github.com/jsdoc3/jsdoc/issues/1489#issuecomment-355521471)


zarov (approved):
No problem to update it, but FYI we are not impacted by the CVE as we don't run this package in a server.

gchoqueux:
@zarov, yes

Issue #613, opened 2018-01-10
Bug while loading a GPX file with multiple Track Segments


nosy-b:
When using GpxUtils to display a GPX file containing different Track Segments it doesn't give the right trajectory rendering as we can see here it mixes polylines extremities although it s a uniform gps trajectory

## Context
Tested _examples/gpx.html_ with http://stereopolis.ign.fr/Frontiere_fortifiee_alt.gpx file sample

## Actual Behavior
![buggpxutils](https://user-images.githubusercontent.com/158514/34783934-295742bc-f62d-11e7-8b38-9308dfab169d.jpg)

## Expected Behavior
![buggpxutilscorrected](https://user-images.githubusercontent.com/158514/34784017-62fd98ae-f62d-11e7-95c3-7487a3431497.jpg)

## Possible Cause/Fix/Solution
Cause: All the `<trkpt>` are mixed together although they can belong to different `<trkseg>` in which the order of the points composing the trajectory is not ordered continuously

Fix: Add the handling of multiple `<trkseg>`

PR #614, opened 2018-01-10
fix (GpxUtils): add handling of multiple track segments #613

nosy-b:
## Description
Add the handling of Multiple track segments when loading GPX Files.

## Motivation and Context
It solves the bug #613

## Screenshots before/after
correctedbugg

### DIFF
--- a/src/Core/Scheduler/Providers/GpxUtils.js
+++ b/src/Core/Scheduler/Providers/GpxUtils.js
@@ -19,6 +19,10 @@ function _gGpxToWTrackPointsArray(gpxXML) {
     return gpxXML.getElementsByTagName('trkpt');
 }
 
+function _gGpxToWTrackSegmentsArray(gpxXML) {
+    return gpxXML.getElementsByTagName('trkseg');
+}
+
 function _gpxPtToCartesian(pt, crs) {
     var longitude = Number(pt.attributes.lon.nodeValue);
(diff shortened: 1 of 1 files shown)


src/Core/Scheduler/Providers/GpxUtils.js
```diff
@@ -82,44 +86,52 @@ function updatePath(renderer, scene, camera) {
...
+                });
+
+                if (Capabilities.isLogDepthBufferSupported()) {
+                    material.fragmentShader = material.fragmentShader.replace(/.*/, '').substr(1);
+                    patchMaterialForLogDepthSupport(material);
+                    // eslint-disable-next-line no-console
+                    console.warn('MeshLineMaterial shader has been patched to add log depth buffer support');
+                }
+
+                const pathMesh = new THREE.Mesh(line.geometry, material);
```
peppsac, line 125: Any idea why a new Mesh is built instead of using `line`? (I know this isn't a change introduced by your PR but this could simplified while modifing this code)

peppsac (approved):
Looks good, thanks for the fix.

src/Core/Scheduler/Providers/GpxUtils.js
nosy-b: That's the syntax for Meshline (https://github.com/spite/THREE.MeshLine), it doesn't inherit from THREE.Mesh nor Object3D strangely

PR #615, opened 2018-01-15
 Avoid excessive memory allocations that cause GC work

gchoqueux:
## Description
add or use optional target to methods to avoid memory allocations
If target is defined then use this to write the result instead of allocating an objet.

## Motivation and Context
Avoid excessive memory allocations that cause GC work.

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -256,11 +256,11 @@ Extent.prototype.center = function center(target) {
     return c;
 };
 
-Extent.prototype.dimensions = function dimensions(unit) {
-    return {
-        x: Math.abs(this.east(unit) - this.west(unit)),
-        y: Math.abs(this.north(unit) - this.south(unit)),
-    };
+Extent.prototype.dimensions = function dimensions(unit, target) {
+    target = target || { x: 0, y: 0 };
+    target.x = Math.abs(this.east(unit) - this.west(unit));
+    target.y = Math.abs(this.north(unit) - this.south(unit));
+    return target;
 };
 
 /**
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -362,11 +362,18 @@ Extent.prototype.intersect = function intersect(other) {
 };
 
 
-Extent.prototype.set = function set(west, east, south, north) {
-    this._values[CARDINAL.WEST] = west;
-    this._values[CARDINAL.EAST] = east;
-    this._values[CARDINAL.SOUTH] = south;
-    this._values[CARDINAL.NORTH] = north;
+Extent.prototype.set = function set(...values) {
+    if (_isTiledCRS(this.crs())) {
+        this._zoom = values[0];
+        this._row = values[1];
+        this._col = values[2];
+    } else {
+        Object.keys(CARDINAL).forEach((key) => {
+            const cardinal = CARDINAL[key];
+            this._values[cardinal] = values[cardinal];
+        });
+    }
+    return this;
 };
 
 Extent.prototype.union = function union(extent) {
--- a/src/Core/Geographic/Projection.js
+++ b/src/Core/Geographic/Projection.js
@@ -4,7 +4,7 @@
  * Description: Outils de projections cartographiques et de convertion
  */
 import MathExt from '../Math/MathExtended';
-import { UNIT } from './Coordinates';
+import Coordinates, { UNIT } from './Coordinates';
 import Extent from './Extent';
 
 
@@ -26,15 +26,10 @@ Projection.prototype.WGS84ToOneSubY = function WGS84ToOneSubY(latitude) {
     return 0.5 + Math.log(Math.tan(MathExt.PI_OV_FOUR + latitude * 0.5)) * MathExt.INV_TWO_PI;
 };
 
+const min = -86 / 180 * Math.PI;
+const max = 84 / 180 * Math.PI;
 Projection.prototype.WGS84LatitudeClamp = function WGS84LatitudeClamp(latitude) {
-    // var min = -68.1389  / 180 * Math.PI;
-    var min = -86 / 180 * Math.PI;
-    var max = 84 / 180 * Math.PI;
-
-    latitude = Math.max(min, latitude);
-    latitude = Math.min(max, latitude);
-
-    return latitude;
+    return Math.min(max, Math.max(min, latitude));
 };
 
 Projection.prototype.getCoordWMTS_WGS84 = function getCoordWMTS_WGS84(tileCoord, bbox, tileMatrixSet) {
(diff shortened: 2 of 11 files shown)


src/Core/Math/Ellipsoid.js
```diff
@@ -55,11 +56,10 @@ Ellipsoid.prototype.setSize = function setSize(size) {
     this._radiiSquared = new THREE.Vector3(size.x * size.x, size.y * size.y, size.z * size.z);
 };
 
-
+const k = new THREE.Vector3();
```
peppsac: Allocating `k` outside of `cartographicToCartesian` isn't a good idea, since you're returning it at each call.

I'd prefer that you add a `target` param and allocate a return vector if missing.

gchoqueux: updated

gchoqueux:
PR updated

src/Core/Geographic/Projection.js
```diff
@@ -42,7 +42,7 @@ Projection.prototype.getCoordWMTS_WGS84 = function getCoordWMTS_WGS84(tileCoord,
     if (tileMatrixSet === 'PM') {
         return this.WMTS_WGS84ToWMTS_PM(tileCoord, bbox);
     } else if (tileMatrixSet === 'WGS84G') {
-        return [tileCoord];
+        return [tileCoord.clone()];
```
autra, line 40: Why do you need to clone here?

src/Core/Geographic/Projection.js
```diff
@@ -123,20 +125,22 @@ Projection.prototype.WGS84toWMTS = function WGS84toWMTS(bbox) {
     var uX = MathExt.TWO_PI / nX;
     var uY = MathExt.PI / nY;
 
-    const center = bbox.center();
+    bbox.center(center);
     var col = Math.floor((MathExt.PI + center.longitude(UNIT.RADIAN)) / uX);
     var row = Math.floor(nY - (MathExt.PI_OV_TWO + center.latitude(UNIT.RADIAN)) / uY);
+    target = target || new Extent('WMTS:WGS84G', 0, 0, 0);
```
autra: nits: I'd prefer this line at the beginning of the function (it's clearer to read, makes the intent of `target` easier to spot).

src/Core/Geographic/Extent.js
```diff
@@ -362,11 +362,17 @@ Extent.prototype.intersect = function intersect(other) {
...
-    this._values[CARDINAL.EAST] = east;
-    this._values[CARDINAL.SOUTH] = south;
-    this._values[CARDINAL.NORTH] = north;
+Extent.prototype.set = function set(...values) {
+    if (_isTiledCRS(this.crs())) {
+        this._zoom = values[0];
+        this._row = values[1];
+        this._col = values[2];
+    } else {
+        Object.keys(CARDINAL).forEach((key) => {
```
autra, line 371: Note, we have currently 4 ways of constructing an Extent. In the future, I guess we want to support the same ways in this method (so there are still 2 ways missing, with an object like `{west, east, south, north}` and with 2 `Coordinates` objects).

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -22,7 +22,20 @@ OBB.prototype.clone = function clone() {
 OBB.prototype.update = function update() {
     this.updateMatrixWorld(true);
 
-    this.pointsWorld = this._cPointsWorld(this._points());
+    if (!this.pointsWorld) {
```
autra: Any reason why this couldn't be done in the constructor once and for all instead? `update` can potentially be called often.

src/Core/Geographic/Projection.js
autra: Because of https://github.com/iTowns/itowns/pull/615/files#diff-fe4e9c46a7cb968cb268ab300bafff1cR114 ? 

If it's the case, the clone must be done in the calling function, because it is the one which knows it needs a new instance.

autra (changes requested):
Already looking good thanks!

If you want to push this to the far end, there are still 2 instances of `new Extent` without targets in `Projection.js`:
- `Projection.prototype.WMTS_WGS84ToWMTS_PM `, 

I think this one is important because the call chain goes up to `TileMesh.getCoordsForLayer`, which is used by some providers' `tileInsideLimits` methods, often called in `layer.update`. I expect some win here because the `new Extent` is called inside a [loop](https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Providers/WMTS_Provider.js#L149) of [loops](https://github.com/iTowns/itowns/blob/master/src/Core/Geographic/Projection.js#L108).

It would require passing an array of targets of the right size somewhere.

- `Projection.prototype.getCoordsWMTS `: seemed to be only called by `getAllCoordsWMTS`: we could delete both of them, for the sake of completion.

src/Renderer/ThreeExtended/OBB.js
gchoqueux: because for the tile, `update` function is called each time DEM is updated

autra: Ok, but that wouldn't prevent `this.pointsWorld` initialisation and all the `new THREE.Vector3` to go into the constructor, right? I mean: `this._cPointsWorld` call would stay here.

src/Core/Geographic/Projection.js
gchoqueux: This function shouldn't return an object's reference
The user isn't supposed to know that are result can be changed via another object

gchoqueux: maybe rename in `computeCoordWMTS`

gchoqueux:
> `Projection.prototype.WMTS_WGS84ToWMTS_PM`

This function is called only if coords haven't been computed, there's no point in passing a target

> It would require passing an array of targets of the right size somewhere

this size is precisely calculated in `WMTS_WGS84ToWMTS_PM`

autra:
> > Projection.prototype.WMTS_WGS84ToWMTS_PM
> 
> This function is called only if coords haven't been computed, there's no point in passing a target

Indeed it is cached in tile.wmtsCoords anyway, so no need to touch that.

src/Renderer/ThreeExtended/OBB.js
gchoqueux: yes, but these points are only used for the horizon culling I wanted to keep this separation to be able to create them only when necessary

src/Core/Geographic/Projection.js
autra: Ok we only currently use this in cases where we need a new object, so we can leave it like that for now, it makes sense.

src/Core/Geographic/Projection.js
gchoqueux: I updated the PR
I moved in default parameter

src/Renderer/ThreeExtended/OBB.js
autra: Yes, but OBB.update is called in `TileMesh.updateMatrixWorld` which is called in subdivideNode and is also called in the constructor. So anyway, these vector3 will get initialized when you do `new OBB` whether or not you're doing horizontal culling (they also get init in planar views). So initializing them in the constructor would be clearer and more natural.

src/Core/Geographic/Projection.js
autra: I don't think you can do that: if I'm not mistaken, default parameter values are instanced once and for all at parsing time, so you'll share one instance for everyone here :-)

src/Renderer/ThreeExtended/OBB.js

src/Core/Geographic/Projection.js
gchoqueux: I verified, a new instance is returned on each call for default parameter value

gchoqueux:
@autra this PR can be merged

src/Core/Math/Ellipsoid.js
```diff
@@ -55,39 +53,30 @@ Ellipsoid.prototype.setSize = function setSize(size) {
...
- */
 /**
  * Convert cartesian coordinates to geographic according to the current ellipsoid of revolution.
- *
  * @param {Object} position - The coordinate to convert
  * @param {number} position.x
  * @param {number} position.y
  * @param {number} position.z
- * @returns {EllipsoidCoordinate} an object describing the coordinates on the reference ellipsoid, angles are in degree
+ * @param {Coordinate} target coordinate to copy result
```
autra: nits: you can indicate that the parameter is optional (see http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

autra (approved):
LGTM, thanks!

PR #617, opened 2018-01-19
fix(examples): wrong tile matrix limit

gchoqueux:
in scanEx example


zarov (approved):
LGTM

peppsac:
This is the 3rd PR fixing tile matrix limits.. PR #594 will hopefully help with this problem by adding `GetCapabilities` support :)

PR #618, opened 2018-01-19
improve imagery layer processing

gchoqueux:
## Description
* add target level mechanism in imagery
* try inferior level if an error occurs to fetch an image
* stop to fetch image when is outside limit (**+test**)

## Motivation and Context
For tiles, some wmts level wasn't at their best

### DIFF
--- a/src/Core/Layer/LayerUpdateState.js
+++ b/src/Core/Layer/LayerUpdateState.js
@@ -52,15 +52,18 @@ LayerUpdateState.prototype.newTry = function newTry() {
 };
 
 LayerUpdateState.prototype.success = function success() {
+    this.failureParams = undefined;
     this.lastErrorTimestamp = 0;
     this.state = UPDATE_STATE.IDLE;
 };
 
 LayerUpdateState.prototype.noMoreUpdatePossible = function noMoreUpdatePossible() {
+    this.failureParams = undefined;
     this.state = UPDATE_STATE.FINISHED;
 };
 
-LayerUpdateState.prototype.failure = function failure(timestamp, definitive) {
+LayerUpdateState.prototype.failure = function failure(timestamp, definitive, failureParams) {
+    this.failureParams = failureParams;
     this.lastErrorTimestamp = timestamp;
     this.state = definitive ? UPDATE_STATE.DEFINITIVE_ERROR : UPDATE_STATE.ERROR;
     this.errorCount++;
--- a/src/Core/Layer/LayerUpdateStrategy.js
+++ b/src/Core/Layer/LayerUpdateStrategy.js
@@ -46,17 +46,29 @@ function _dichotomy(nodeLevel, currentLevel, options = {}) {
         Math.ceil((currentLevel + nodeLevel) / 2));
 }
 
-export function chooseNextLevelToFetch(strategy, node, nodeLevel, currentLevel, layer) {
-    switch (strategy) {
-        case STRATEGY_GROUP:
-            return _group(nodeLevel, currentLevel, layer.updateStrategy.options);
-        case STRATEGY_PROGRESSIVE:
-            return _progressive(nodeLevel, currentLevel, layer.updateStrategy.options);
-        case STRATEGY_DICHOTOMY:
-            return _dichotomy(nodeLevel, currentLevel, layer.options);
-        // default strategy
-        case STRATEGY_MIN_NETWORK_TRAFFIC:
-        default:
-            return _minimizeNetworkTraffic(node, nodeLevel, currentLevel);
+export function chooseNextLevelToFetch(strategy, node, nodeLevel, currentLevel, layer, failureParams) {
+    let nextLevelToFetch;
+    const maxZoom = layer.options.zoom ? layer.options.zoom.max : Infinity;
+    if (failureParams) {
+        nextLevelToFetch = _dichotomy(failureParams.targetLevel, currentLevel, layer.options);
+    } else {
+        switch (strategy) {
+            case STRATEGY_GROUP:
+                nextLevelToFetch = _group(nodeLevel, currentLevel, layer.updateStrategy.options);
+                break;
+            case STRATEGY_PROGRESSIVE: {
+                nextLevelToFetch = _progressive(nodeLevel, currentLevel, layer.updateStrategy.options);
+                break;
+            }
+            case STRATEGY_DICHOTOMY:
+                nextLevelToFetch = _dichotomy(nodeLevel, currentLevel, layer.options);
+                break;
+            // default strategy
+            case STRATEGY_MIN_NETWORK_TRAFFIC:
+            default:
+                nodeLevel = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : nodeLevel;
+                nextLevelToFetch = _minimizeNetworkTraffic(node, nodeLevel, currentLevel);
+        }
     }
+    return Math.min(nextLevelToFetch, maxZoom);
 }
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -238,10 +238,19 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
         return;
     }
 
+    const failureParams = node.layerUpdateState[layer.id].failureParams;
     const currentLevel = node.material.getColorLayerLevelById(layer.id);
-    const zoom = node.getCoordsForLayer(layer)[0].zoom || node.level;
-    const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, zoom, currentLevel, layer);
+    const nodeLevel = node.getCoordsForLayer(layer)[0].zoom || node.level;
+    const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, nodeLevel, currentLevel, layer, failureParams);
     if (targetLevel <= currentLevel) {
+        return;
+    }
+
+    // Retry tileInsideLimit because you must check with the targetLevel
+    // if the first test layer.tileInsideLimit returns that it is out of limits
+    // and the node inherits from its parent, then it'll still make a command to fetch texture.
+    if (!layer.tileInsideLimit(node, layer, targetLevel)) {
+        node.layerUpdateState[layer.id].noMoreUpdatePossible();
         return;
     }
 
@@ -284,7 +293,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                     console.warn('Imagery texture update error for', node, err);
                 }
                 const definitiveError = node.layerUpdateState[layer.id].errorCount > MAX_RETRY;
-                node.layerUpdateState[layer.id].failure(Date.now(), definitiveError);
+                node.layerUpdateState[layer.id].failure(Date.now(), definitiveError, { targetLevel });
                 if (!definitiveError) {
                     window.setTimeout(() => {
                         context.view.notifyChange(false, node);
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-examples && npm run test-unit",
     "test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --compilers js:babel-core/register test/*unit_test.js",
-    "test-examples": "npm run transpile && mocha test/globe_test.js && mocha test/planar_test.js && mocha test/postprocessing_test.js && mocha test/externalscene_test.js",
+    "test-examples": "npm run transpile && mocha test/globe_test.js && mocha test/planar_test.js && mocha test/postprocessing_test.js && mocha test/externalscene_test.js && mocha test/wmts_test.js",
     "build": "webpack -p",
     "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -155,7 +155,7 @@ Extent.prototype.as = function as(crs) {
     });
 };
 
-Extent.prototype.offsetToParent = function offsetToParent(other) {
+Extent.prototype.offsetToParent = function offsetToParent(other, target = new THREE.Vector4()) {
     if (this.crs() != other.crs()) {
         throw new Error('unsupported mix');
     }
@@ -167,7 +167,7 @@ Extent.prototype.offsetToParent = function offsetToParent(other) {
         const r = (this._row - (this._row % diff)) * invDiff;
         const c = (this._col - (this._col % diff)) * invDiff;
 
-        return new THREE.Vector4(
+        return target.set(
             this._col * invDiff - c,
             this._row * invDiff - r,
             invDiff, invDiff);
@@ -191,7 +191,7 @@ Extent.prototype.offsetToParent = function offsetToParent(other) {
         Math.abs(
             this.north() - this.south()) / dimension.y;
 
-    return new THREE.Vector4(originX, originY, scaleX, scaleY);
+    return target.set(originX, originY, scaleX, scaleY);
 };
 
 Extent.prototype.west = function west() {
--- a/test/itowns-testing.js
+++ b/test/itowns-testing.js
@@ -164,6 +164,21 @@ exports.countVisibleAndDisplayed = (node) => {
     }
 };
 
+exports.countVisibleAndDisplayedLayerImage = (node, layerId) => {
+    function levelMaterial(node) {
+        if (node.material) {
+            const currentLevel = node.material.getColorLayerLevelById(layerId);
+            if (currentLevel > -1) {
+                while (counters.visible_at_level.length <= currentLevel) {
+                    counters.visible_at_level.push(0);
+                }
+                counters.visible_at_level[currentLevel]++;
+            }
+        }
+    }
+    node.traverseVisible(levelMaterial);
+};
+
 exports.counters = counters;
 exports.runTest = () => {
     testStarted = true;
--- a/test/layeredmaterialnodeprocessing_unit_test.js
+++ b/test/layeredmaterialnodeprocessing_unit_test.js
@@ -4,7 +4,7 @@ import TileMesh from '../src/Core/TileMesh';
 import Extent from '../src/Core/Geographic/Extent';
 import OBB from '../src/Renderer/ThreeExtended/OBB';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from '../src/Core/Layer/LayerUpdateStrategy';
-/* global describe, it, xit, beforeEach */
+/* global describe, it, beforeEach */
 
 const assert = require('assert');
 
@@ -88,7 +88,7 @@ describe('updateLayeredMaterialNodeImagery', function () {
         assert.equal(context.scheduler.commands.length, 1);
     });
 
-    xit('tile should not request texture with level > layer.zoom.max', () => {
+    it('tile should not request texture with level > layer.zoom.max', () => {
         const tile = new TileMesh(geom, {
             extent: new Extent('EPSG:4326', 0, 0, 0, 0),
             level: 15,
--- a/test/layeredmaterialnodeprocessing_unit_test.js
+++ b/test/layeredmaterialnodeprocessing_unit_test.js
@@ -102,7 +102,7 @@ describe('updateLayeredMaterialNodeImagery', function () {
         // Since layer is using STRATEGY_MIN_NETWORK_TRAFFIC, we should emit
         // a single command, requesting a texture at layer.options.zoom.max level
         updateLayeredMaterialNodeImagery(context, layer, tile);
-
+        updateLayeredMaterialNodeImagery(context, layer, tile);
         assert.equal(context.scheduler.commands.length, 1);
         assert.equal(
             context.scheduler.commands[0].targetLevel,
--- a/test/wmts_test.js
+++ b/test/wmts_test.js
@@ -1,0 +1,42 @@
+/* global itowns, assert, describe, it */
+var itownsTesting = require('./itowns-testing.js');
+var example = require('../examples/globe.js');
+
+var initialState = 0;
+
+
+describe('wmts', function () {
+    it('should fetch like expected for wmts', function (done) {
+        itownsTesting.counters.visible_at_level = [];
+        let layer;
+        const listener = () => {
+            if (initialState == 0) {
+                initialState++;
+                layer = example.view.getLayers(l => l.id == 'Ortho')[0];
+                layer.options.zoom.max = 8;
+
+                example.view.camera.camera3D.position.copy(
+                    new itowns.Coordinates('EPSG:4326',
+                        example.initialPosition.longitude,
+                        example.initialPosition.latitude,
+                        10000).as('EPSG:4978').xyz());
+                example.view.notifyChange(true);
+            } else if (initialState < 5) {
+                initialState++;
+            } else {
+                for (var obj of example.view.wgs84TileLayer.level0Nodes) {
+                    itownsTesting.countVisibleAndDisplayedLayerImage(obj, 'Ortho');
+                }
+                assert.equal(layer.options.zoom.max, itownsTesting.counters.visible_at_level.length - 1);
+                assert.equal(itownsTesting.counters.visible_at_level[8], 2);
+                assert.equal(itownsTesting.counters.visible_at_level[7], 0);
+                example.view.mainLoop.removeEventListener('command-queue-empty', listener);
+                done();
+                process.exit(0);
+            }
+        };
+        example.view.mainLoop.addEventListener('command-queue-empty', listener);
+        itownsTesting.runTest();
+    });
+});
+


zarov:
You forgot to call `wmts_test` in `package.json`.

gchoqueux:
@zarov  thanks, updated

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -228,11 +228,19 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     }
 
     const currentLevel = node.material.getColorLayerLevelById(layer.id);
-    const zoom = node.getCoordsForLayer(layer)[0].zoom || node.level;
-    const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, zoom, currentLevel, layer);
+    const bestLevel = node.getCoordsForLayer(layer)[0].zoom || node.level;
+    let targetLevel = node.layerUpdateState[layer.id].targetLevel ? node.layerUpdateState[layer.id].targetLevel - 1 :
```
peppsac: Please add a comment explaining your intent.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -228,11 +228,19 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
+        chooseNextLevelToFetch(layer.updateStrategy.type, node, bestLevel, currentLevel, layer);
+
+    targetLevel = Math.min(layer.options.zoom.max, targetLevel);
```
peppsac: This shouldn't be necessary because strategies must use layer.options.zoom.max.

(ie: if you had a case requiring this Math.min, the fix should probably be in the strategy code)

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -228,11 +228,19 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
+
     if (targetLevel <= currentLevel) {
         return;
     }
+    if (!layer.tileInsideLimit(node, layer, targetLevel)) {
```
peppsac: This has already been tested above. Could you add a comment explaining why this 2nd check is needed?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -274,6 +282,10 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                     console.warn(`Imagery texture update error for ${node}: ${err}`);
                 }
                 const definitiveError = node.layerUpdateState[layer.id].errorCount > MAX_RETRY;
+                if (layer.protocol == 'wmts' || layer.protocol == 'wmtsc') {
```
peppsac: No no no :)

Please don't make this code protocol aware.
If remembering the previous `targetLevel` is useful then do it everytime.

peppsac (changes requested):
Thanks for the PR / fixes.

I left a few comments to be fixed before we can proceed with merging.

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: Because if the first test `layer.tileInsideLimit` returns that it is out of limits and the node inherits from its parent, then it'll still make a command to fetch texture.
and you must check with the `targetLevel`

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: moved in strategy code

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: comment added

src/Process/LayeredMaterialNodeProcessing.js
peppsac: This comment should be in the code; otherwise someone may remove this thinking it's useless.

And I believe this shouldn't be done here. But rather around l.156 where we explicitely ignore the `tileInsideLimit` result.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -282,10 +284,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                     console.warn(`Imagery texture update error for ${node}: ${err}`);
                 }
                 const definitiveError = node.layerUpdateState[layer.id].errorCount > MAX_RETRY;
-                if (layer.protocol == 'wmts' || layer.protocol == 'wmtsc') {
-                    // keep the targetLevel that caused an error
-                    node.layerUpdateState[layer.id].targetLevel = targetLevel;
-                }
+                node.layerUpdateState[layer.id].failedFetchLevel = targetLevel;
```
peppsac: What about modifying `LayerUpdateState.failure` function to accept a optional argument,  that would be stored in a `failureParams` member.
Something like:

```js
node.layerUpdateState[layer.id].failure(Date.now(), definitiveError, targetLevel);
```

that could be used as:
```js
const failedFetchLevel = node.layerUpdateState[layer.id].failureParams.level;
```

WDYT?

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: it's more difficult to do it higher because we have to consider the `targetLevel`

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: I updated the PR

src/Process/LayeredMaterialNodeProcessing.js
peppsac: Does it really depends on `targetLevel`?

Couldn't be fixed by always calling `node.layerUpdateState[layer.id].noMoreUpdatePossible();` here:
```js
if (!layer.tileInsideLimit(node, layer)) {
 node.layerUpdateState[layer.id].noMoreUpdatePossible();
 ...
}
```

This would allow to inherit the parent's texture where possible, but would prevent any further fetch. WDYT?

gchoqueux: if the parent has a down-scaled texture, inheritance'll not provide the best texture

by example a layer with limit level 0 to 10, a node with level 12  can inherit a texture with level 0.
and have in this case its status to `UPDATE_STATE.FINISHED` and then never get the level 10 texture

peppsac: OK I get it.

The problem here is that `tileInsideLimit` can be expensive - especially for StaticProvider which uses a naive search algo. Anyway your change makes sense and StaticProvider can be improved.

src/Core/Layer/LayerUpdateStrategy.js
```diff
@@ -47,16 +47,21 @@ function _dichotomy(nodeLevel, currentLevel, options) {
...
-            return _dichotomy(nodeLevel, currentLevel, layer.options);
+            nextLevelToFetch = _dichotomy(nodeLevel, currentLevel, layer.options);
+            break;
         // default strategy
         case STRATEGY_MIN_NETWORK_TRAFFIC:
         default:
-            return _minimizeNetworkTraffic(node, nodeLevel, currentLevel);
+            nextLevelToFetch = _minimizeNetworkTraffic(node, nodeLevel, currentLevel);
     }
+    return Math.min(nextLevelToFetch, layer.options.zoom.max);
```
peppsac: `layer.options.zoom` isn't always defined. This change breaks `panorama.html`

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -273,12 +286,21 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
-                node.layerUpdateState[layer.id].failure(Date.now(), definitiveError);
-                if (!definitiveError) {
-                    window.setTimeout(() => {
-                        context.view.notifyChange(false, node);
-                    }, node.layerUpdateState[layer.id].secondsUntilNextTry() * 1000);
+                const nextLevelToTry = targetLevel - 1;
+                const pFailureParams = node.layerUpdateState[layer.id].failureParams;
+                const pFailureLevel = pFailureParams ? pFailureParams.targetLevel : undefined;
+                // notify a new try specifying failure parameters
+                if (layer.tileInsideLimit(node, layer, nextLevelToTry) && nextLevelToTry != pFailureLevel) {
```
peppsac: I'd prefer not to have this logic here: it's a duplicate of the code you added 10 lines above.

peppsac (changes requested):

gchoqueux:
@peppsac Branch updated

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -151,12 +151,13 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
     if (node.layerUpdateState[layer.id] === undefined) {
         node.layerUpdateState[layer.id] = new LayerUpdateState();
+        const isInsideLimit = layer.tileInsideLimit(node, layer);
 
         if (!layer.tileInsideLimit(node, layer)) {
             // we also need to check that tile's parent doesn't have a texture for this layer,
             // because even if this tile is outside of the layer, it could inherit it's
             // parent texture
-            if (!layer.noTextureParentOutsideLimit &&
+            if (!isInsideLimit &&
```
peppsac: You're removing the code handling `layer.noTextureParentOutsideLimit` that you added a couple months ago.

Can you explain why it's not needed anymore?

(and technically this is an API breakage, so it would require deprecation handling)

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -202,13 +203,13 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
     material.setLayerVisibility(layerIndex, layer.visible);
     material.setLayerOpacity(layerIndex, layer.opacity);
 
+    const failureParams = node.layerUpdateState[layer.id].failureParams;
     const ts = Date.now();
     // An update is pending / or impossible -> abort
-    if (!layer.visible || !node.layerUpdateState[layer.id].canTryUpdate(ts)) {
+    if (!layer.visible || (!node.layerUpdateState[layer.id].canTryUpdate(ts) && !failureParams)) {
```
peppsac: This condition is becoming hard to parse.
If I understand correctly, you want to skip update if:
- layer is not visible
- layer is in error and failureParam is unset?

Why would you want to update if `canTryUpdate` returns false? (remember than `layerUpdateState` spaces retries so I don't think it makes sense to ignore it simply because we have `failureParams` from a previous attempt.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -273,12 +286,21 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
-                node.layerUpdateState[layer.id].failure(Date.now(), definitiveError);
-                if (!definitiveError) {
-                    window.setTimeout(() => {
-                        context.view.notifyChange(false, node);
-                    }, node.layerUpdateState[layer.id].secondsUntilNextTry() * 1000);
+                const nextLevelToTry = targetLevel - 1;
+                const pFailureParams = node.layerUpdateState[layer.id].failureParams;
+                const pFailureLevel = pFailureParams ? pFailureParams.targetLevel : undefined;
+                // notify a new try specifying failure parameters
+                if (nextLevelToTry != pFailureLevel) {
```
peppsac: I'm not sure to understand the code nor the comment. Can you explain a bit more?

peppsac (changes requested):

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: oops, it's a error

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: the reason is if there was an error and we want to re-fetch with a new target level there are two things that go against new try

* the max number of test that can block a try with a new target level
* the pause time between tests which is also blocking

I was thinking of adding a status of the kind : re-try with new Params

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: I'll continue to search because without `layer.tileInsideLimit (node, layer, nextLevelToTry)`, the test no longer has the same meaning, and may be the same behaviour.
I'll also clarify a comment

gchoqueux:
@peppsac I tested and pushed a proposition to retry a fetching

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -227,13 +226,26 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
 
+    const failureParams = node.layerUpdateState[layer.id].failureParams;
     const currentLevel = node.material.getColorLayerLevelById(layer.id);
-    const zoom = node.getCoordsForLayer(layer)[0].zoom || node.level;
-    const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, zoom, currentLevel, layer);
+    const bestLevel = node.getCoordsForLayer(layer)[0].zoom || node.level;
+    const failedFetchLevel = failureParams ? failureParams.targetLevel : undefined;
+    // if there has already been a failed fetch on previous level
+    // then try at the lower level
+    const targetLevel = failedFetchLevel ? failedFetchLevel - 1 :
```
peppsac: Actually I'm not sure that retrying at `level - 1` makes sense.
Can you give me an example scenario where it would be useful/correct ?

gchoqueux: As indicated in the description : `For tiles, some wmts level wasn't at their best`

You can try in asia with scanEx example, if you change camera's Range 100000 to 3000 with no animation, the wmts will not be at its best zoom. the reason is in this [comment](https://github.com/iTowns/itowns/pull/618#discussion_r165647342)

peppsac: Isn't this problem caused by the strategy bug returning a targetLevel = 12 while the layer has a max = 10 ? (and you already fixed this).

gchoqueux: yes that's it.

it also adds the matrix limit case are false.

by example for wmts layer.
it's indicated an interval limit 0 to 15 but the server only returns images from level 0 to 10.
So a level 18 tile can stay at level 0  image because it's blocked by an error at level 15 image

peppsac: Then using `dichotomy` as the default strategy could make more sense than blindly retrying `level - 1` for all failing tiles.

gchoqueux: yes, it was planned. I'll add it in this PR

gchoqueux: PR updated, retry use dichotomy

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -261,7 +272,11 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                 // and stop retrying after X attempts.
             }
 
-            node.layerUpdateState[layer.id].success();
+            if (!failureParams) {
+                node.layerUpdateState[layer.id].success();
+            } else {
+                node.layerUpdateState[layer.id].retry(Date.now(), { targetLevel: failureParams.targetLevel });
```
peppsac: Why do you use `retry()` on success?

src/Core/Layer/LayerUpdateState.js
```diff
@@ -52,20 +52,29 @@ LayerUpdateState.prototype.newTry = function newTry() {
...
 };
 
 LayerUpdateState.prototype.failure = function failure(timestamp, definitive) {
+    this.failureParams = undefined;
     this.lastErrorTimestamp = timestamp;
     this.state = definitive ? UPDATE_STATE.DEFINITIVE_ERROR : UPDATE_STATE.ERROR;
     this.errorCount++;
 };
 
+LayerUpdateState.prototype.retry = function failure(timestamp, failureParams) {
```
peppsac: Here's an alternative proposal: add an optional `params`  argument to the `try()` method and store it in `this.lastTryParams`.
This way you can get rid of the `retry()` calls in `LayeredMaterialNodeProcessing.js` and you call still act differently when needed (ie: if errorCount is > 0 and lastTryParams is defined).

peppsac (changes requested):

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: if there was a previous failure, a loading success doesn't mean that the best texture is loaded
i add a comment

peppsac: `sucess()` doesn't imply that it will not try to improve the texture later.

gchoqueux: ok

src/Core/Layer/LayerUpdateState.js
gchoqueux: what try() method?

peppsac: Sorry, I meant the `newTry()`  method.

gchoqueux: finally i remove retry(), and use failure() method to save the last failure param. because it's interesting data to try new fetch

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -227,13 +226,25 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
...
 
+    const failureParams = node.layerUpdateState[layer.id].failureParams;
     const currentLevel = node.material.getColorLayerLevelById(layer.id);
-    const zoom = node.getCoordsForLayer(layer)[0].zoom || node.level;
-    const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, zoom, currentLevel, layer);
+    const bestLevel = node.getCoordsForLayer(layer)[0].zoom || node.level;
+    // if there's failureParams, compute the next level to try by dichotomic calculation
+    // dichotomy between the current level and last failed level
+    const nextLevelToTry = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : undefined;
+    const targetLevel = nextLevelToTry || chooseNextLevelToFetch(layer.updateStrategy.type, node, bestLevel, currentLevel, layer);
```
peppsac: Now that this code has been simplified, it seems that it would be better in `chooseNextLevelToFetch`...

This way the code would be shared between imagery and elevation and the update process code would be simpler.
You may want to enrich the `LayerUpdateState` object to remember the N previous failure, so that it can decide if it wants to continue (using dichotomy) or stop.

gchoqueux: in first step, I move the code in chooseNextLevelToFetch

gchoqueux: @peppsac for the first step, WDYT?

src/Core/Layer/LayerUpdateStrategy.js
```diff
@@ -46,17 +46,29 @@ function _dichotomy(nodeLevel, currentLevel, options) {
...
-export function chooseNextLevelToFetch(strategy, node, nodeLevel, currentLevel, layer) {
+export function chooseNextLevelToFetch(strategy, node, nodeLevel, currentLevel, layer, failureParams) {
+    let nextLevelToFetch;
+    const maxZoom = layer.options.zoom ? layer.options.zoom.max : Infinity;
     switch (strategy) {
         case STRATEGY_GROUP:
-            return _group(nodeLevel, currentLevel, layer.updateStrategy.options);
-        case STRATEGY_PROGRESSIVE:
-            return _progressive(nodeLevel, currentLevel, layer.updateStrategy.options);
+            nodeLevel = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : nodeLevel;
```
peppsac: Why do you override `nodeLevel` in different ways?
```js
nodeLevel = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : nodeLevel;
// vs
nodeLevel = failureParams ? failureParams.targetLevel : nodeLevel;
```
I fail to see a good reason to not always use the same formula.

src/Core/Layer/LayerUpdateState.js
```diff
@@ -31,6 +31,9 @@ LayerUpdateState.prototype.canTryUpdate = function canTryUpdate(timestamp) {
         }
         case UPDATE_STATE.ERROR:
         default: {
+            if (this.failureParams) {
```
peppsac: Why do you want to retry without a pause here?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -227,13 +226,22 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
         return;
     }
 
+    const failureParams = node.layerUpdateState[layer.id].failureParams;
     const currentLevel = node.material.getColorLayerLevelById(layer.id);
-    const zoom = node.getCoordsForLayer(layer)[0].zoom || node.level;
-    const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, zoom, currentLevel, layer);
+    const bestLevel = node.getCoordsForLayer(layer)[0].zoom || node.level;
```
peppsac: `nodeLevel` is probably a better name.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -273,9 +280,15 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                     // eslint-disable-next-line no-console
                     console.warn(`Imagery texture update error for ${node}: ${err}`);
                 }
-                const definitiveError = node.layerUpdateState[layer.id].errorCount > MAX_RETRY;
-                node.layerUpdateState[layer.id].failure(Date.now(), definitiveError);
-                if (!definitiveError) {
+
+                // try next level if a dichotomic research makes sense
+                if (currentLevel != targetLevel - 1) {
+                    node.layerUpdateState[layer.id].failure(Date.now(), false, { targetLevel });
```
peppsac: These changes aren't necessary now, are they?

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: ok

src/Core/Layer/LayerUpdateStrategy.js
gchoqueux: by example 
* for STRATEGY_DICHOTOMY : there'll be a double dichotomies
* for STRATEGY_GROUP: it would return `failureParams.targetLevel`

src/Core/Layer/LayerUpdateState.js
gchoqueux: I don't think there is a delay in trying to fetch a new node with a different level

peppsac: But `failureParams` will almost always be defined. So you're completely disabling the delay thing.

So IMHO this isn't correct, especially since the measure you're introducing here is a mitigation for a broken data service (advertising an incorrect max zoom level).

src/Core/Layer/LayerUpdateStrategy.js
peppsac: - STRATEGY_DICHOTOMY: I don't see the problem. The node is at level 5, the max level is 15. So it'll try 10. If it fails, it makes sense to try 7 (it also makes sense to try 10 - 1 but I don't see the value of adding a special case for this).

- STRATEGY_GROUP: IMHO dichotomy or -1 are both broken. We should only ever fetch one of the values from the groups defined by the user.

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: yes, I removed this part

src/Core/Layer/LayerUpdateStrategy.js
gchoqueux: * for STRATEGY_DICHOTOMY: 
so for this strategy with `nodeLevel = failureParams ? failureParams.targetLevel : nodeLevel;` you to get your result `7`

* for STRATEGY_GROUP: we use `nodeLevel = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : nodeLevel` to get next level to try and then we use it in `_group` to fetch one of the values from the groups defined by the user.

src/Core/Layer/LayerUpdateState.js
gchoqueux: yes, i updated the PR
if two last Failure params aren't equivalent we can try without delay

peppsac: You added some complicated code to do this, but IMHO you missed my point.

The code https://github.com/iTowns/itowns/pull/618/files#diff-ad71c3a479d08240c3053171dd65d576R285 can't distinguish between a broken data server (in this case retrying with a lower level can make sense), a network error (wifi connection lost, in which case skipping the error delay is causing more harm than good).

So basically, I'd prefer to be on the safe side, and always use the pause delay in case of error.

Btw, did you try to turn off your internet connection while using itowns to see what happens before/after this PR?

gchoqueux: In this PR in offline, itowns retries with delay we find the current mechanism.
because there are several errors on the same fail parameters.

if we respect the pause delay to try again with new parameters
we can wait for more than 10 seconds the best texture

src/Core/Layer/LayerUpdateState.js
```diff
@@ -31,14 +54,23 @@ LayerUpdateState.prototype.canTryUpdate = function canTryUpdate(timestamp) {
         }
         case UPDATE_STATE.ERROR:
         default: {
+            const lastFailParams = this.lastFailureParams() || {};
+            const beforeLastFailParams = this.failureParams[this.failureParams.length - 2] || {};
+            // if two last Failure params aren't equivalent we can try without delay
```
peppsac: You can remove all this code because `secondsUntilNextTry` returns 0 in this case.

src/Core/Layer/LayerUpdateState.js
```diff
@@ -31,14 +54,23 @@ LayerUpdateState.prototype.canTryUpdate = function canTryUpdate(timestamp) {
...
             return errorDuration <= (timestamp - this.lastErrorTimestamp);
         }
     }
 };
 
 LayerUpdateState.prototype.secondsUntilNextTry = function secondsUntilNextTry() {
-    if (this.state !== UPDATE_STATE.ERROR) {
+    const lastFailParams = this.lastFailureParams() || {};
+    const beforeLastFailParams = this.failureParams[this.failureParams.length - 2] || {};
+    // if two last failure params aren't equivalent we can try without delay
```
peppsac: Your comment is describing what your doing (which is clear enough from the var/func names).
What's missing is the *why* you're doing this?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -283,13 +290,12 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
                 if (__DEBUG__) {
                     console.warn('Imagery texture update error for', node, err);
                 }
+
                 const definitiveError = node.layerUpdateState[layer.id].errorCount > MAX_RETRY;
-                node.layerUpdateState[layer.id].failure(Date.now(), definitiveError);
-                if (!definitiveError) {
```
peppsac: Why did you remove this? There's no need to schedule a `notifyChange()` if the `definitiveError` state has been reached.

peppsac (changes requested):

gchoqueux:
@peppsac PR updated

src/Core/Layer/LayerUpdateState.js
```diff
@@ -57,10 +58,16 @@ LayerUpdateState.prototype.success = function success() {
 };
 
 LayerUpdateState.prototype.noMoreUpdatePossible = function noMoreUpdatePossible() {
+    this.failureParams = undefined;
     this.state = UPDATE_STATE.FINISHED;
 };
 
-LayerUpdateState.prototype.failure = function failure(timestamp, definitive) {
+LayerUpdateState.prototype.lastFailureParams = function _lastFailureParams() {
+    return this.failureParams[this.failureParams.length - 1];
```
peppsac: Since you're now only ever using the last failure parameters, there's no need to keep an array anymore.

And maybe you could reset (`this.failureParams = undefined;) it on success as well?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -219,7 +219,6 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
         return;
     }
 
-
```
peppsac: Unneeded whitespace change.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -273,7 +281,6 @@ export function updateLayeredMaterialNodeImagery(context, layer, node) {
             }
 
             node.layerUpdateState[layer.id].success();
-
```
peppsac: Unneeded whitespace change.

gchoqueux:
done

gchoqueux:
@peppsac  we could merge this one?

peppsac (approved):

PR #619, opened 2018-01-22
fix(core): error update OBBHelper

gchoqueux:
## Motivation and Context
it isn't longer possible to display `OBBHelper`


zarov (approved):
Looks good to me, nice catch !

PR #623, opened 2018-01-23
perf(core): reduce to obb's highest points for horizon culling

gchoqueux:
## Description
remove unnecessary points

### DIFF
--- a/src/Process/GlobeTileProcessing.js
+++ b/src/Process/GlobeTileProcessing.js
@@ -72,7 +72,7 @@ function pointHorizonCulling(pt) {
 }
 
 function horizonCulling(node) {
-    const points = node.OBB().pointsWorld;
+    const points = node.OBB().topPointsWorld;
 
     for (const point of points) {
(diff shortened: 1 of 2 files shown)


src/Process/GlobeTileProcessing.js
```diff
@@ -73,9 +73,9 @@ function pointHorizonCulling(pt) {
 
 function horizonCulling(node) {
     const points = node.OBB().pointsWorld;
-
-    for (const point of points) {
-        if (!pointHorizonCulling(point)) {
+    // loop on obb's highest points
```
peppsac: Is `pointsWorld` used by anyone else?

Because if it's only used here, it might be worth only computing the top 4 in `_cPointsWorld()`

zarov: Same as in #619 then

gchoqueux: yes, PR updated

peppsac (approved):
LGTM, thanks

PR #624, opened 2018-01-23
OPEN QUESTION: eslint no-use-before-define error

gchoqueux:
## Description
eslint no-use-before-define error excepted for function

## Motivation and Context
Avoid to use undefined variable, see commit

WDYT?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,7 +59,7 @@ module.exports = {
 
     // maybe 'no-mixed-operators': ['error', { allowSamePrecedence: true }],
     'no-mixed-operators': 'off',
-    'no-use-before-define': 'off',
+    'no-use-before-define': ['error', { 'functions': false }],
     // should probably be
     // 'no-underscore-dangle': ['error', { allowAfterThis: true, allowAfterSuper: true }],
     'no-underscore-dangle': 'off',
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -99,10 +99,10 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
 };
 
 MainLoop.prototype._step = function _step(view, timestamp) {
+    const dt = timestamp - this._lastTimestamp;
     view.execFrameRequesters(MAIN_LOOP_EVENTS.UPDATE_START, dt, this._updateLoopRestarted);
 
     const willRedraw = this.needsRedraw;
-    const dt = timestamp - this._lastTimestamp;
     this._lastTimestamp = timestamp;
 
     // Reset internal state before calling _update (so future calls to View.notifyChange()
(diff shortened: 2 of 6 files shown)


autra:
Good idea! +1 for activation, but let's make it an error instead of a warning:
- it's bad enough to justify an error
- Warnings don't get fixed. See for example `func-names` eslint rules (we currently have 5 warnings that nobody cares about).

autra:
Something like
```js
    "no-use-before-define": ["error", { "functions": false }]
```
would be good in my opinion: functions are hoisted, so it's not a problem if they are defined after their use.

gchoqueux:
PR change : warning to error, excepted for function
```js
    "no-use-before-define": ["error", { "functions": false }]
```

gliegard:
@autra : when you said "Warnings don't get fixed. See for example func-names eslint rules (we currently have 5 warnings that nobody cares about)."
It's just because we don't see theses warnings every day :  as it's in the examples, it doesn't appear when we do "npm start"

I think we should use the difference between Warning and Example.

Usused variable, style errors should be warnings.

When I am coding, I want an Error when I'have really done a mistake (like a compilation error in other langage), and only a Warning when I left a space after a brace...

autra:
@gliegard even if they appear, warnings don't get fixed, period. Believe me on that, I have seen it several time, on several projects, in several different contexts. The thing is: maybe you'll fix it, maybe I'll fix it, maybe @gchoqueux'll fix it. But a random contributor won't, and the reviewer won't see it on github.

So: either we consider one eslint rule to be important, and we want travis to enforce it and put some red on a PR, and that's an error. Or we don't want travis to block on that, and we drop the rule completely, because it won't be respected consistently. Imho, there is no middle-ground.

Anyway, for this rule in particular, I don't think it should be a warning anyway.

gliegard:
ok, I understand. So there are 2 situations
- When we are coding... when we do "npm start"
- When Travis verify a PR. It does "npm test"

Maybe we can configure Travis to raise Error when it finds a Warning ?
Or another EsLint configuration file, one for the PR, one for dev ? 
WDYT ?

And I agree that for this rule in particular, It should be an error :)

autra:
> Maybe we can configure Travis to raise Error when it finds a Warning ?

What's the point though? If it serious enough to justify blocking a PR, it's `error` right?

Warnings are made for not so important points, where we want people to be careful, but still not block their workflow. My (opinionated) point is: this is useless, as it'll get ignored anyway, and we'll be better off either caring completely or not caring at all, instead of half-caring.

Anyway, maybe this is a discussion we can have during the next itowns code sprint?

gchoqueux:
ok, let's go to merge?

peppsac (approved):
LGTM, thanks :)

PR #625, opened 2018-01-23
Increase performance in feature protocol

gchoqueux:
## Description
increase algo, remove GC

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -35,11 +35,17 @@ promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').th
 function altitudeLine(properties, contour) {
     var altitudes = [];
     var i = 0;
-    var alt = 0;
+    var result;
+    var tile;
+    var layer = globeView.wgs84TileLayer;
     if (contour.length && contour.length > 0) {
         for (; i < contour.length; i++) {
-            alt = itowns.DEMUtils.getElevationValueAt(globeView.wgs84TileLayer, contour[i]).z + 2;
-            altitudes.push(alt);
+            result = itowns.DEMUtils.getElevationValueAt(layer, contour[i], 0, tile);
+            if (!result) {
+                result = itowns.DEMUtils.getElevationValueAt(layer, contour[i], 0);
+            }
+            tile = [result.tile];
+            altitudes.push(result.z + 2);
         }
         return altitudes;
     }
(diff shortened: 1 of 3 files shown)


src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -26,21 +26,25 @@ function readCRS(json) {
...
     for (const pair of coordinates) {
         // TODO: 1 is a default z value, makes this configurable
-        const coords = new Coordinates(crsIn, pair[0], pair[1], 1);
         if (crsIn === crsOut) {
-            out.push(coords);
+            out[i] = new Coordinates(crsIn, pair[0], pair[1], 1);
         } else {
-            out.push(coords.as(crsOut));
+            coords.set(crsIn, pair[0], pair[1], 1);
+            coords._normal = undefined;
```
zarov: This is a no from me, a variable prefixed with an underscore is a private value AFAIK, so we shouldn't access it like that. You should tweak `Coordinates.js` if you need to change this value.

examples/globe_wfs_extruded.js
```diff
@@ -35,11 +36,17 @@ promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').th
...
     var i = 0;
-    var alt = 0;
+    var result;
+    var tile;
+    var layer = globeView.wgs84TileLayer;
     if (contour.length && contour.length > 0) {
         for (; i < contour.length; i++) {
-            alt = itowns.DEMUtils.getElevationValueAt(globeView.wgs84TileLayer, contour[i]).z + 2;
-            altitudes.push(alt);
+            result = itowns.DEMUtils.getElevationValueAt(layer, contour[i], 0, tile);
```
zarov, line 44: It saddens me that you need to reiterate later if no result have been found with cache. Even if it is only an example, `getElevationValueAt` should be handling this on its own.

zarov (changes requested):
Results are incredibly goods ! Any idea why is it still this slow in Chrome though ?
Edit: slowness was caused by non activation of GPU acceleration in Chrome

examples/globe_wfs_extruded.js
gchoqueux: yes, I'll do another PR

src/Renderer/ThreeExtended/GeoJSON2Features.js
gchoqueux: yes you right, now `Coordinates.set` undefines `_normal`

gchoqueux:
PR updated

zarov (approved):

PR #626, opened 2018-01-24
Add Example with THREE  collada loader

gchoqueux:
## Description
Example THREE  v89 collada loader
buildingcollada


nosy-b:
Very cool new example!

Please add some code to allow clicking on the model without freezing the camera.


```
// dirty old way for example:
var colladaID = globeView.mainLoop.gfxEngine.getUniqueThreejsLayer();
globeView.mesh.traverse((obj) => { obj.layers.set(colladaID); }); 
globeView.camera.camera3D.layers.enable(colladaID); 

```

Is there a free collada available online with a building instead of the famous duck to attract more architects & co?

gchoqueux:
@nosy-b 
I would have liked but I haven't yet find a satisfactory model, I'll replace it as soon as I find good one.
PR updated

nosy-b (changes requested):
LGTM
But can you change the title of the example ('THREE loader ') to something more meaningful like 'Collada Loader' or if you want to show the genericity of the example "3D model loader" for example?

gchoqueux:
@nosy-b PR updated

examples/building.dae
```diff
@@ -0,0 +1,451 @@
+<?xml version="1.0" encoding="utf-8"?>
```
autra: We have a repo for the examples data. Can you host this there?

autra (changes requested):

autra:
Also: `features(examples)`

Examples are not features. `examples: add...` is enough. As a rule of thumb, if it doesn't touch `src/`, it cannot be feature imo. Thanks, it'll ease the next release!

examples/building.dae
gchoqueux: yes

gchoqueux:
@autra @nosy-b PR updated

examples/collada.js
```diff
@@ -0,0 +1,87 @@
...
+    meshCoord.setAltitude(cameraTargetPosition.altitude());
+
+    loader.load('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/collada/building.dae', function col(collada) {
+        var colladaID = globeView.mainLoop.gfxEngine.getUniqueThreejsLayer();
+
+        model = collada.scene;
+        model.position.copy(meshCoord.as(globeView.referenceCrs).xyz());
+        model.lookAt(new itowns.THREE.Vector3(0, 0, 0));
+        model.rotateX(Math.PI);
+        model.rotateZ(-Math.PI * 0.2);
```
autra: Where did you get those rotation (and position) data? If they are not empiric, maybe we could add a reference.

examples/collada.js
```diff
@@ -0,0 +1,87 @@
+/* global itowns, document, renderer, GuiTools, Promise */
+// # Simple Globe viewer
+
+// Define initial camera position
+// Coordinate can be found on https://www.geoportail.gouv.fr/carte
+// setting is "coordonnée geographiques en degres decimaux"
+
+// Position near Annecy lake.
+// var positionOnGlobe = { longitude: 6.2230, latitude: 45.8532, altitude: 5000 };
```
autra: Remove dead code and comment.

examples/collada.js
```diff
@@ -0,0 +1,87 @@
...
+        model.traverse(function _(obj) { obj.layers.set(colladaID); });
+        globeView.camera.camera3D.layers.enable(colladaID);
+
+        // update coordinate of the mesh
+        model.updateMatrixWorld();
+    });
+}
+
+// Listen for globe full initialisation event
+globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function init() {
```
autra: I know it's done this way on other examples, but I think it's not very good. I think it's better if we add each layer individually to menuGlobe as soon as it is ready. So essentially, that means calling `menuGlobe.addImageryLayerGUI` in `addLayerCb`. Pro: if one layer fails, the others would still show up in the GUI.

autra (changes requested):
Nice example, thanks!

What would be *very* great is to have debug tools for this. A good possible way to do it would be to just adapt `GeometryDebug` to take an object3d. (ie renaming and testing for the type of the `layer` parameter, and do a traverse if it's a Object3D).

In this case, the actual Mesh containing the house has an array in `.material`, not a single material, so you'll need to deal with this case too.

examples/collada.js
gchoqueux: I do that.
for the other examples it will then be necessary to solve the problem of order of the layers in the interface

examples/collada.html
```diff
@@ -0,0 +1,53 @@
...
+        <div id="viewerDiv"></div>
+        <script src="GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script type="text/javascript">
+            var renderer;
+            var exports = {};
+            var THREE = itowns.THREE;
+        </script>
+        <script src="https://cdn.rawgit.com/mrdoob/three.js/dev/examples/js/loaders/ColladaLoader.js"></script>
```
autra: Oh forgot this: let's point to our current tag version: v0.87 (we'll need to remember changing that when we update three.js version)

autra: And I need to think about a way to use them in our build without this trick...

gchoqueux:
@autra yes good idea for  `GeometryDebug`, I create a new issue  #641 to initiate a new PR

examples/collada.html
gchoqueux: yes we have to find a solution
Here, we point to dev THREE version

autra: >  Here, we point to dev THREE version

yes, that's my point, let's point to the tag here instead to ensure compatibility with our npm version.

gchoqueux: i updated to check npm version

examples/collada.html
```diff
@@ -0,0 +1,56 @@
...
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script type="text/javascript">
+            var renderer;
+            var exports = {};
+            var THREE = itowns.THREE;
+            var script = document.createElement('script');
+            script.type = 'text/javascript';
+            script.src = 'https://cdn.rawgit.com/mrdoob/three.js/r' + THREE.REVISION + '/examples/js/loaders/ColladaLoader.js';
+            document.getElementsByTagName('script')[0].parentNode.appendChild(script);
```
autra: this is equivalent to `document.head.appendChild`. Actually here, I'd advocate for `document.body.appendChild`. It'll still get added before `collada.js`, and the order in the dom would still match the *real* execution order.

autra: Otherwise, it's a good idea to reference THREE.VERSION :-)

examples/collada.js
```diff
@@ -0,0 +1,89 @@
...
+
+var menuGlobe = new GuiTools('menuDiv');
+
+var promiseElevation = [];
+
+menuGlobe.view = globeView;
+
+function addLayerCb(layer) {
+    return globeView.addLayer(layer).then(function addGui() {
+        var la = globeView.getLayers(function cl(l) { return l.id === layer.id; })[0];
```
autra: `whenReady` is resolved with the layer, so you already have it as arguments to your `addGui` function above.

gchoqueux: I tried with whenReady, but I was surprised that it did not work. It'll look for this bug

examples/collada.html
gchoqueux: ok, thanks, I replace by `document.body.appendChild`

examples/collada.js
```diff
@@ -0,0 +1,89 @@
...
+    });
+    // collada
+    var loader = new itowns.THREE.ColladaLoader(loadingManager);
+
+    // get the position on the globe, from the camera
+    var cameraTargetPosition = globeView.controls.getCameraTargetGeoPosition();
+
+    // position of the mesh
+    var meshCoord = cameraTargetPosition;
+    meshCoord.setAltitude(cameraTargetPosition.altitude());
```
autra: Unfortunately this doesn't work in the general case. Especially, if you throttle the network with chrome devtools, you'll see that this gets resolves before the correct elevation texture arrives, causing the house to be under ground. 

To fix it, you need to implement a child layer of the right elevation layer (or of the base layer) to be able to react to any elevation change at this point.

Other problem: if the user move the camera before the house is loaded, it will get loaded at the wrong position.

That makes me think we need an easier mechanism for laying objects on grounds at a specific position, that will basically doing all the boilerplate of adding the mesh to the right tile, put it in local coordinates and update its z when / if this tile get some elevations.

For this PR, I'm not requiring this. This is a problem we already have in other examples anyway.

autra: Actually I also reproduce with firefox, everytime...

examples/collada.js
autra:
```js
function addLayerCb(layer) {
    return globeView.addLayer(layer).then(function addGui(la) {
        if (la.type === 'color') {
            menuGlobe.addImageryLayerGUI(la);
        } else if (la.type === 'elevation') {
            menuGlobe.addElevationLayerGUI(la);
        }
    });
}
```
that's what I meant.

autra: whenReady is returned by `addLayer`

gchoqueux: yes, i know, as I told you I tried but it's buggy. I will create a issue

autra: This snippet works, I've tested.

examples/collada.js
gchoqueux: yes indeed, we know these problems well.
I can remove all the layers in the example to reach the goal of this PR to load a collada

examples/collada.js
gchoqueux: ok, I didn't follow, I updated

examples/collada.js
autra: Or you can do this:

> To fix it, you need to implement a child layer of the right elevation layer (or of the base layer) to be able to react to any elevation change at this point.

This works, for example, added at the end of `addMeshToScene`:

```js
window.position = new itowns.Coordinates('EPSG:4326',  positionOnGlobe.longitude, positionOnGlobe.latitude);
globeView.addLayer({
        update: (context, layer, element) => {
            if (element.extent.isPointInside(position) && element.visible && element.material &&  element.material.visible) {
                itowns.DEMUtils.placeObjectOnGround(globeView.wgs84TileLayer, globeView.referenceCrs, model, {}, [element]);
            }
        }
    },
    globeView.wgs84TileLayer);
```
The beauty of it: this way you don't need to wait for elevationPromise to load your collada. 

There is still the issue of the camera: it's not placed correctly if the elevation layer arrives too slowly (I reproduce it in Firefox and in chrome with network slowed)... Maybe we can update its position the first time we have an element with an elevation texture in the update above ?

examples/collada.html
```diff
@@ -0,0 +1,56 @@
...
+        <script type="text/javascript">
+            var renderer;
+            var exports = {};
+            var THREE = itowns.THREE;
+            var script = document.createElement('script');
+            script.type = 'text/javascript';
+            script.src = 'https://cdn.rawgit.com/mrdoob/three.js/r' + THREE.REVISION + '/examples/js/loaders/ColladaLoader.js';
+            document.body.appendChild(script);
+        </script>
+        <script src="collada.js"></script>
```
autra: Actually, as I'm testing with slow network, I had cases where collada failed because ColladaLoader constructor was not defined. In fact, you also need a `defer` attribute on this `script` tag, because newer versions of firefox apparently treats dynamically inserted script as `async`. `defer` ensures they are executed after, see https://stackoverflow.com/a/8996894

gchoqueux:
PR updated 
Altitude set by the user / architect

examples/collada.js
```diff
@@ -0,0 +1,79 @@
...
+    // loading manager
+    var loadingManager = new itowns.THREE.LoadingManager(function _addModel() {
+        globeView.scene.add(model);
+        globeView.notifyChange(true);
+    });
+    // collada loader
+    var loader = new itowns.THREE.ColladaLoader(loadingManager);
+
+    // building coordinate
+    var coord = new itowns.Coordinates('EPSG:4326', 4.2165, 44.844, 1417);
```
autra, line 55: Placing the camera in a static fashion is good, but it's better if the house is placed directly on the floor like I explained [here](https://github.com/iTowns/itowns/pull/626#discussion_r166372874). It's a good showcase of a use case of a child layer.

autra (changes requested):

examples/collada.js
gchoqueux: in SIG and architecture, it is better for the building to be placed at a fixed altitude.
your proposal makes more sense for a place mark

autra: > in SIG and architecture, it is better for the building to be placed at a fixed altitude.

Why?

In the context of a 3D viewer with an elevation, it makes more sense to place it according to the elevation. What will happen if elevation get updated by new measures? The house will either float or be underground?

Beside, `examples/` are also showcases: placing it in a fixed z like you do is easy: everybody can do it. But using child layer mechanism to do so is a good thing to show to our user, allowing them to better understand how iTowns work and use it at their advantage. I think this feature of itowns should be better advertised and that's the perfect example to it.

gchoqueux: Altitude is fixed and assumed in this example, otherwise the measures of the building would be false

autra: Why would they be false?

gchoqueux: A measurement of a building can't be exact if it moves

autra: What are you calling "measurement of a building"? Please be more specific, otherwise I need to guess what you meant :-)

Anyway, its size won't change. If you're talking about its position, then you're already placing it arbitrarily.

If you're talking about measurement between buildings, it depends on what you want. Sometimes you want them on the DEM, sometimes you will want them placed in a fixed way, that's true, but in this case, my point still stands:

> examples/ are also showcases: placing it in a fixed z like you do is easy: everybody can do it

If architecture guys uses iTowns, they would be able to place it absolutely, or following the elevation depending on their use case. It's better to showcase the latter.

gchoqueux: In this example altitude is fixed in a non arbitrary way and assumed.
Otherwise the altitude of the building would be wrong

autra: >  altitude is fixed

by who?

Jeremy-Gaillard: I think what @gchoqueux means is that 3D objects are generally geolocated, they already have an altitude that has been measured in real life. So, changing it by projecting it on the ground isn't necessarily a good idea.

nosy-b: Let's be clear, We need to handle both ways: 
- A fixed altitude passed with the model or defined once for all
- A clamp to ground option, updating the altitude of the model each time the DTM is updated.

-> This option can come in another PR!

nosy-b (approved):

examples/collada.js
autra: > 3D objects are generally geolocated

Well in this case, this model is not and placed arbitrarily.

> Let's be clear, We need to handle both ways:

yes of course, but way n°1 is just a z placement of a 3D object... No added value in that right? Just basic three.js stuff. The 2nd way is a good opportunity to showcase a powerful feature of iTowns.

> -> This option can come in another PR!

Right, I won't block any more on that, but please note that until this PR comes (if it comes), this house floats in the air during the time we wait for the elevation to be loaded :-/

autra (approved):

autra:
@nosy-b as you are (maybe) releasing, I let you merge that, depending on whether or not you want it after or before the release.

examples/collada.js
nosy-b: @gchoqueux Can you adjust the height to reduce the levitation effect (with DEMUtils for ex) ? Or is it just due to the DTM changing its value depending on level?

autra: @nosy-b you mean the altitude, or the z-scale of the building?

PR #627, opened 2018-01-24
Fix example lint warnings

gliegard:
## Description
- fix eslint warnings in the examples
- consider formatting error as warning during development phase

## Motivation and Context
The goal of this PR is to deal with style warning like 'no trailing space' smarter
- To be strict during 'npm test' phase, so Travis will raise error on warning, so Reviewer will see it.
- And during 'npm start' phase, when we are coding, warnings like 'no trailing space' appears as warning and not as error.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,10 +22,13 @@ module.exports = {
     commonjs: true
   },
   rules: {
+    'no-trailing-spaces': 'warn',
+    'padded-blocks' : 'warn',
+    'no-unused-vars': 'warn',
     'no-plusplus': 'off',
     // this option sets a specific tab width for your code
     // http://eslint.org/docs/rules/indent
-    indent: ['error', 4, {
+    indent: ['warn', 4, {
       SwitchCase: 1,
       VariableDeclarator: 1,
       outerIIFEBody: 1,
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -36,7 +36,7 @@ if (!renderer) {
     miniView.mainLoop.gfxEngine.renderer.setClearColor(0x000000, 0);
 
     // update miniview's camera with the globeView's camera position
-    globeView.addFrameRequester(itowns.MAIN_LOOP_EVENTS.AFTER_RENDER, function () {
+    globeView.addFrameRequester(itowns.MAIN_LOOP_EVENTS.AFTER_RENDER, function updateMiniView() {
         // clamp distance camera from globe
         var distanceCamera = globeView.camera.camera3D.position.length();
         var distance = Math.min(Math.max(distanceCamera * 1.5, minDistance), maxDistance);
--- a/examples/positionGlobe.js
+++ b/examples/positionGlobe.js
@@ -69,12 +69,12 @@ function addMeshToScene() {
 }
 
 // Listen for globe full initialisation event
-globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
+globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function globeInitialized() {
     // eslint-disable-next-line no-console
     console.info('Globe initialized');
-    Promise.all(promises).then(function () {
-        menuGlobe.addImageryLayersGUI(globeView.getLayers(function (l) { return l.type === 'color'; }));
-        menuGlobe.addElevationLayersGUI(globeView.getLayers(function (l) { return l.type === 'elevation'; }));
+    Promise.all(promises).then(function init() {
+        menuGlobe.addImageryLayersGUI(globeView.getLayers(function filterColor(l) { return l.type === 'color'; }));
+        menuGlobe.addElevationLayersGUI(globeView.getLayers(function filterElevation(l) { return l.type === 'elevation'; }));
 
         addMeshToScene();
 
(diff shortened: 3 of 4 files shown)


autra:
Ok, I see. The idea is to get warnings fixed while keeping the 2 colors, that's right?

Imho, it means that warnings actually behave as errors, so I don't see why we shouldn't use error. If I'm constrained to fix them anyway, I don't care if they're all red,

That being said, thanks for these fixes :-)

gliegard:
The goal of this PR is to change Not important Errors (trailing space is just the first example but I think we should add 'usused variable' and all the style errors..) to Warnings.

And the goal is to use the console output colors when we are coding (Red means Errors, Yellow means warning). Now if you leave a trailing space, the console become Red, I think we waste time reading this kind of errors...

And as you said here https://github.com/iTowns/itowns/pull/624#issuecomment-360122442 "warning don't get fixed" ; So this PR make the 'npm run lint' (so indirectly 'npm test', and the Travid, and the PR reviewing) raiser error when warning occurs... So Warning will be fixed :)

autra:
Yes I understood what this Pr is doing :-) 

> The goal of this PR is to change Not important Errors to Warnings.

My opinion stated differently: if these rules are not important, let's drop them. We'll have enough with important ones. 

That being said, I won't block this if another reviewer wants this.

tbroyer:
@autra As I understand it from the comments in the various issues/PRs, the problem might actually be that having eslint as part of the "npm start" is the actual problem: you cannot simply "poke at the code until it works", you also have to make sure it passes all lints (including formatting, which some editors don't automatically do – heck, apparently some use editors that don't know about the `.editorconfig` that already says to trim trailing spaces :wink:); so one alternative could be to disable eslint in webpack when the Node environment is `DEVELOPMENT`…  
The reason it was made this way was that we were almost certain developers would **only** run "npm start" locally, and we'd had a couple merge conflict resolution issues only due to different formatting, including indents and… trailing spaces (we didn't have tests yet at the time, and didn't yet mandate the use of pull requests). Maybe it's time to revise this approach? WDYT? (only run eslint in "production" builds, i.e. "npm run build", "npm test", and not in the webpack-dev-server, aka "npm start", to not "get in the way" and "break the flow")

gliegard:
@tbroyer I think we should keep the EsLint test in the 'npm start' ; because the way it watches the files is powerfull. If you remove it, we'll have to do the 'npm test' manually to see style warnings, and we will waste time doing that.

@autra We shouldn't drop these Not Important Error (Warnings) because, as tbroyer said, 
"we'd had a couple merge conflict resolution issues only due to different formatting, including indents..."


It think it will be better to see Warnings, instead of Errors for this kind of style warning, in the 'npm start', during the development phase... and at the end of the dev phase, fix the warnings and push the code !

tbroyer:
Though as I said, if the problem is "npm start", then maybe we should turn (some of) the lints to warnings instead of errors in this case. E.g. in `.eslintrc.js`:
```js
const formatting = process.env.NODE_ENV === 'development' ? 'warn' : 'error';
// …
    'indent': [ formatting, 4, {
```
(using `NODE_ENV` here as that's what's already used in `webpack.config.js` to turn `__DEBUG__` on, any other mean of turning it on/off would be OK for me)

autra:
> @autra We shouldn't drop these Not Important Error (Warnings) because, as tbroyer said,
"we'd had a couple merge conflict resolution issues only due to different formatting, including indents..."

if they generate merge conflicts, they *are* important :-) My point was: no need to have 2 categories (which is a different question than "should we lint in npm start"). I'm all in favor of having end-of-line linted. Anyway, we can have the 2 categories it's important for you.

@gliegard have you tested `failOnError` and `failOnWarning` eslint-loader options? I would expect that setting them to false in the webpack config would do exactly what we want.

gliegard:
@autra Thank for the suggestion, I've tested failOnError and failOnWarning ... but it's not needed since it already doesn't fail on warning and I think it's a good thing that it fails on errors. 
For me it's important to have 2 categories (error and warning), that's why I did this PR :)

@tbroyer I did your trick, it works !
So I have removed the "max-warning=0" in the 'eslint ..' command.

I've updated and rebase the code.
Also, I have added the 'no-unused-vars' error to be warnings during dev phase.

autra (changes requested):
First of all, sorry for the delay.

Ok @gliegard I understand your use case. I don' feel the need myself to have the 2 categories, especially because it will lead to unimportant debate over whether or not one particular thing is a warning or an error, but I'm fine with merging that. So let's get this merged!

That being said, I preferred your `max-warnings` approach, which conveys better our intention:
- some rules are warnings and some are errors, it doesn't depend on the mode. I strongly disagree with making that environment dependent.
- we want to ensure warnings get fixed before a PR get merged, and the `max-warnings` in the eslint command does just that.

Can you revert to your previous version? Or alternatively, another maybe finer possibility: add this option to the `test` command instead like so: 

```js
"test": "npm run lint -- --max-warning=0 && npm run build && npm run test-examples && npm run test-unit",
```
That way, `npm run lint` will still return 0 on warnings, but `npm run test` (what is run by travis) will fail. Maybe it's more flexible for people who use `npm run lint` in their script or in commit hooks... Pick your favorite approach.

Thanks for caring about the linter btw ;-) Hopefully, we'll manage to tackle all the deactivated rules in `.eslintrc.js` one day (a linter session during the code sprint? Anyone? :-p)

autra:
> npm run lint will still return 0 on warnings

I meant: by default. Individuals could then still add `-- --max-warning=<whatever>` in their scripts.

gliegard:
@autra Thank's for the suggestion, it's a good compromise.
I've updated the code !

autra (approved):
LGTM, thanks!

gliegard:
thanks!

PR #628, opened 2018-01-25
fix(protocols): avoid to set points's count at zero

gchoqueux:
## Description
avoid to set points's count at zero

## Motivation and Context
With count equal zero, there's warning message : RENDER WARNING: Render count or primcount is 0


zarov (approved):
Thanks, LGTM

autra:
@gchoqueux @zarov this is not the correct way to fix it: if `shouldBeLoaded` is 0, it means pts should be hidden, not shown with only one point.

Issue #595, opened 2017-12-07
Clean Repository


gchoqueux:
I start to clean repository,  I remove dead or redundant branches.
Can you move/delete a maximum of your dead or redundant branches.

For my part I move some of my branches on my fork. 
Do you know a way to automatically synchronize upstream repo with a github fork (without cloning the fork locally).

gchoqueux:
I found this [https://upriver.github.io/](https://upriver.github.io/)
But there are so many branches that it doesn't show them all in combo box :)

zarov:
Not automatically, but there are some resources online about simple synchronization:
https://gist.github.com/EvanDotPro/1506822

autra:
Or a simpler solution: don't synchronize it ;-) After all, you don't need an updated master on your fork, just create local branches from itowns/itowns#master, you can still push your wip branches to \<user\>/itowns first, then open a PR from them, or push them to itowns/itowns later.

PR #629, opened 2018-01-29
3d tiles : GLTF 2.0 and fix OBB 

gchoqueux:
## Description
* update to new gltf 2.0 loader
* fix compute error in OBB and OBBHelper for 3DTiles

### DIFF
--- a/src/Renderer/ThreeExtended/B3dmLoader.js
+++ b/src/Renderer/ThreeExtended/B3dmLoader.js
@@ -1,5 +1,6 @@
 import * as THREE from 'three';
 import GLTFLoader from './GLTFLoader';
+import LegacyGLTFLoader from './deprecated/LegacyGLTFLoader';
 import BatchTable from './BatchTable';
 
 const matrixChangeUpVectorZtoY = (new THREE.Matrix4()).makeRotationX(Math.PI / 2);
@@ -8,6 +9,7 @@ const matrixChangeUpVectorZtoX = (new THREE.Matrix4()).makeRotationZ(-Math.PI / 
 
 function B3dmLoader() {
     this.glTFLoader = new GLTFLoader();
+    this.LegacyGLTFLoader = new LegacyGLTFLoader();
 }
 
 function filterUnsupportedSemantics(obj) {
@@ -44,7 +46,7 @@ function applyOptionalCesiumRTC(data, gltf, textDecoder) {
     }
 }
 
-B3dmLoader.prototype.parse = function parse(buffer, gltfUpAxis, textDecoder) {
+B3dmLoader.prototype.parse = function parse(buffer, gltfUpAxis, url, textDecoder) {
     if (!buffer) {
         throw new Error('No array buffer provided.');
     }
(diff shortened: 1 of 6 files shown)


autra:
If I'm not mistaken, it's the second PR that needs THREE v89. Could you submit the THREE update as a separate PR please? It needs a review on its own, and it could be merged quickly this way.

Jeremy-Gaillard:
Pinging @jailln that was also working on the glTF2 upgrade.

gchoqueux:
@Jeremy-Gaillard, yes I see this branch [3Dtiles_glTF2](https://github.com/iTowns/itowns/tree/3Dtiles_glTF2), it's on this one that you worked?

jailln:
Yes it's on this branch but I've added one commit on my fork to support the batch id with glTF 2 : https://github.com/jailln/itowns/commit/028510d7c80635fc93c1a22e11b52764fa16823c 

We basically used the same method as for glTF1 Loader: we duplicated the code in a `GLTF2Loader` file and added a few lines to load the batch id attribute.

gchoqueux:
@Jeremy-Gaillard @jailln 
This PR is ready
I add support for batch id in gltf 1.0 and 2.0
Can you tell me what you think of this PR?

jailln:
It looks good to me, I tested it with two 3d-tiles tilesets respectively based on glTF 1.0 and 2.0 and it works for both; thanks

gchoqueux:
@Jeremy-Gaillard @jailln 
the batch id is used to pick building?
What are the user cases?

Jeremy-Gaillard:
In b3dm files, all the different objects are merged in a single geometry. The batch id is an attribute attached to each triangle, which allows to identify these objects (all the triangles of the same object will have the same batch id).

The standard use cases are picking (and retrieving information linked to the picked object) or [applying a style to the geometries](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/Styling).

jailln:
Another usecase is to attach attributes (e.g. address, name, etc.) in the [batch table](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/BatchTable) of the b3dm files to each object identified by a batch id in the glTF.

gchoqueux:
ok, the buffer makes sense for the style, although we could use the COLOR buffer

But for the picking and get table batch info, we could store the points ranges for each object in the b3dm batch table. Which is less expensive in RAM, VRAM and network Transfert.

jailln:
Yes we could also do that but this won't be the standard way of doing it though; I think that this is a proposition that could be made to the 3d-tiles comunity and that we should stick to the batch id way for now. What do you think ?

gchoqueux:
yes, for the moment, we keep _batchid

gchoqueux:
> a proposition that could be made to the 3d-tiles comunity 

Maybe not, B3dm Batch table contains user's info,  so we can use it as we want.

src/Renderer/ThreeExtended/B3dmLoader.js
```diff
@@ -104,9 +106,18 @@ B3dmLoader.prototype.parse = function parse(buffer, gltfUpAxis, textDecoder) {
...
             };
-            this.glTFLoader.parse(buffer.slice(28 + b3dmHeader.FTJSONLength +
-                b3dmHeader.FTBinaryLength + b3dmHeader.BTJSONLength +
-                b3dmHeader.BTBinaryLength), onload);
+
+            const gltf = buffer.slice(28 + b3dmHeader.FTJSONLength +
+                    b3dmHeader.FTBinaryLength + b3dmHeader.BTJSONLength +
+                    b3dmHeader.BTBinaryLength);
+
+            const onError = (ex) => {
```
Jeremy-Gaillard: You should read the glTF version from the header instead of reacting to the GLTFReader error. You have no guarantee that the error you catch is caused by the glTF version.

Jeremy-Gaillard (changes requested):
Just one little thing to fix, the rest looks good to me.

src/Renderer/ThreeExtended/B3dmLoader.js
gchoqueux: yes, I'll change that

gchoqueux: Ok, PR is updated. 
can I delete the `origin/3Dtiles_glTF2` branch?
I take this opportunity to know you would have no other obsolete branches to delete? (#595)

Jeremy-Gaillard: I don't think you need to print a warning for glTF version 1, otherwise everything seems OK.

You can delete my branch. I'll take a look at what other branches can be deleted soon.

Jeremy-Gaillard (approved):

src/Renderer/ThreeExtended/B3dmLoader.js
gchoqueux: ok thanks

PR #630, opened 2018-01-29
feat(core): update to THREE v89

gchoqueux:
## Description
Update to THREE v89, with validation that nothing is breaking

## Motivation and Context
need in PR #629 , #626, #631, #634

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6361,9 +6361,9 @@
       "dev": true
     },
     "three": {
-      "version": "0.87.1",
-      "resolved": "https://registry.npmjs.org/three/-/three-0.87.1.tgz",
-      "integrity": "sha1-Rmo07cRUNFnO2bnX0na2Uhb+K6g=",
+      "version": "0.89.0",
+      "resolved": "https://registry.npmjs.org/three/-/three-0.89.0.tgz",
+      "integrity": "sha1-RELYGaYWiHG40ss3rRKiQxDBcPU=",
       "dev": true
     },
     "three.meshline": {
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
   },
   "peerDependencies": {
     "proj4": "^2.4.4",
-    "three": "^0.87.0",
+    "three": "^0.89.0",
     "three.meshline": "^1.0.3"
   },
   "devDependencies": {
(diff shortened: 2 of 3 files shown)


peppsac (changes requested):
Nope.

We can't jump 2 versions of three.js without a review of the changes.

Did you review the release notes for v0.88 and v0.89?
Did you check what breaking changes could affect us?
Or if any new features would be helpful?

Quick example with 1 change:
```
    Added decodeText() static method and implemented in all loaders. #12654 (@donmccurdy)
```
Which could allow us to drop the `text-encoding-utf-8` dependency.

gchoqueux:
Yes of course, I read the reviews
In THREE.js's working methods, they passe the break changes in **deprecated**
By example, [AxesHelper](https://github.com/iTowns/itowns/pull/630/files#diff-df61a67eee71572718ef066564a3a0afR258), the change was reflected in itowns and deprecate in THREE ([here](https://github.com/mrdoob/three.js/pull/12321/files#diff-697749014a2009c4ea9c6741f042c804R306)). 
Except this one, I think there's no other break change

There are still efforts to make the most of the THREE's features. (Some are in PR #629 , #626)

I push a commit to replace  `text-encoding-utf-8`, but I'm not sure of their implementation that instantiates a `DecodeText` to use, see [here](https://github.com/mrdoob/three.js/pull/12654/files#diff-75041da7958590b0bd3002b909522af0R7). it goes against the efforts made in this commit dc4c41d5

peppsac:
And what about:
```
    Replaced triangulator code with Earcut. #12661 #12725 (@Mugen87)
```
We're using `earcut` as well so it's probably an opportunity to get rid of this dependency (or to explicitely state in the commit message why we don't want to do this).

gchoqueux:
I add in the commit message : `TODO try the new earcut in three v89` see #631

and 9cc4f57 you really want this change?

peppsac:
> I add in the commit message : TODO try the new earcut in three v89 see #631

I hoped that you could integrate it in this PR since it's probably a 1 line change in Feature2Mesh.js

> and 9cc4f57 you really want this change?

I think so. But I'd love to have some context: for instance did you check the itowns.js bundle size to see if removing the dependency had a measurable effect?

gchoqueux:
This PR was created to facilitate the review and separate the refactos needed in other PRs. (#629 , #626 #634). In this sense I moved the decode text refacto in #634.
This PR is an update of three.js with a check if nothing is breaking.
it need a new PR for the EarCut, for the re-factoring, the test of this new feature and the reviewing (a issue is open #631)

peppsac (approved):
OK, let's split the upgrades on new threejs feature to other PRs.

PR #633, opened 2018-01-30
Fix wrong preSSE computation

Jeremy-Gaillard:
## Description
Based on the formula slide 16 of [this presentation](http://52.4.31.236/presentations/Rendering%20the%20Whole%20Wide%20World%20on%20the%20World%20Wide%20Web.pdf), there was a little mistake in our SSE computation.

This fix should increase the number of tiles loaded for a given view. We may need to adjust our SSE threshold if we want to keep the same behaviour as before.

EDIT: tests are failing, I still need to reflect the change of this PR to the testing scripts

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -199,7 +199,7 @@ export function pre3dTilesUpdate(context, layer) {
      // TODO: not correct -> see new preSSE
     // const HFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) / context.camera.ratio);
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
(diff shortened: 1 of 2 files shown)


gchoqueux (changes requested):
Thanks for the fix!
Actually the formula seems to be wrong.
I started new tests on this PR
It would be necessary to change the value by default `options.sseSubdivisionThreshold` to have a suitable subivision (I think between 2.0 and 3.0).
As you pointed out, there are many more subdivisions. The behaviour isn't longer the same in terms of performance / load (RAM / CPU / NETWORK).

You must also check before you merge the functions:
`computeTileZoomFromDistanceCamera` and `computeDistanceCameraFromTileZoom`

and modify  the number of subdivision expected in test globe example

peppsac:
BTW, I have a work-in-progress PR to unify SSE computing for all layer types instead of having a slightly different formula for each one.

peppsac:
Replaced by #674

PR #634, opened 2018-01-31
feat(protocols): replace text-encoding-utf-8 module by THREE.LoaderUtils.decodeText

gchoqueux:
## Description
Replace `text-encoding-utf-8` module by `THREE.LoaderUtils.decodeText`

## Motivation and Context
Remove module  `text-encoding-utf-8`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7565,11 +7565,6 @@
       "integrity": "sha1-IGvo4YiGC1FEJTdebxrom/sB/Y0=",
       "dev": true
     },
-    "text-encoding-utf-8": {
-      "version": "1.0.1",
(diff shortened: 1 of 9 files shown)


zarov:
@gchoqueux can you rebase this PR ? Thanks

Jeremy-Gaillard:
Updated.

peppsac:
Currently three.js creates a new `TextDecoder` each time you ease the `decodeText` method (see https://github.com/mrdoob/three.js/blob/dev/src/loaders/LoaderUtils.js#L7) so maybe we don't want to merge this PR.

gchoqueux:
replace by #698

Issue #182, opened 2016-09-01
[loader] Vector tiles support


vpicavet:
iTowns should support vector tiles at some point.

Vector tiles are an efficient vector format for rendering of 2D data. Vector tile services start to be available, and server-side components do exist.
- https://www.mapbox.com/vector-tiles/specification/
- https://www.mapbox.com/vector-tiles/
- https://en.wikipedia.org/wiki/Vector_tiles
- http://wiki.openstreetmap.org/wiki/Vector_tiles#Ongoing_projects.2C_people

Implementing vector tiles support suppose having Clamping available first ( #181 ).

We could use the following services for testing purposes : 
- Klokan : https://maps.klokantech.com/
- MapBox
- Mapzen

zarov:
I'm currently working on integrating vector tiles, and I identified a few step ahead:
- load a `pbf`/`mvt` file and parse it
- support `pbf`/`mvt` loading for TMS, WMS, WFS and WMTS
- display the result as an image (similar to RasterProvider)
- display the result as a Mesh

First I'm concentrating on solving one way, and then the others (that should be smooth).

Below are a list of PR related to this work (will be updated):
- #639 feat(core): compute TMS coordinates from north to south

zarov:
Let's lay a solid basis, with the aim to easily iterate later on this subject (for example when expanding style possibilities). Here's a more detailed point of view I got.

### Tile loading

Vector tiles are accessible through TMS, WMTS, WMS and WFS. Currently, no provider can fetch `pbf` files. A method needs to be added for this, that could be used like `OGCWebServiceHelper.getColorTextureByUrl` for example.

Question: do we add all the methods related to vector tiles in a new file, like `VectorTile_Provider.js` ? I don't think it's a good idea to create a new single provider, as it would be a pain to add support for all protocols in it. I was thinking more of adding a `VectorTileHelper.js`, featuring some public methods:

- `getVectorTileRawByUrl`: would return a "raw" vector tile, not the loaded `pbf` but a [Mapbox VectorTile](https://github.com/mapbox/vector-tile-js).
- `getVectorTileGeoJSONByUrl`: would return a vector tile as a GeoJSON (needed for next methods, so why not making it available)
- `getVectorTileTextureByUrl`: if the `layer.type` is `color`, would return a texture, using the `Feature2Texture` class.
- `getVectorTileMeshByUrl`: if the `layer.type` is `geometry`, would return a mesh, using the `Feature2Mesh` class.

[...]

PR #635, opened 2018-02-02
refactor(core): unify mesh picking methods

peppsac:
The idea of this PR is to expose a single entry point for querying the view.

This method belongs to the View.

Here are a few example usage:
  * view.pickObjectsAt({ x, y })
  * view.pickObjectsAt({ x, y }, 'wfsBuilding')
  * view.pickObjectsAt({ x, y }, 'wfsBuilding', myLayer)
  * view.pickObjectsAt({ x, y }, 'wfsBuilding', myObject3d)

It supports either layers, layer's id or THREE.Object3D.
By default it'll use THREE.Raycaster, but some layer can provide their own implementation.

Currently the layer with a custom implementation are:
  - layers using TileMesh (globe, planar, panorama)
  - pointclouds

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -152,15 +152,12 @@
                 }
             }
             function picking(event) {
-                // Pick an object with batch id
-                var mouse = globe.eventToNormalizedCoords(event);
-                var raycaster = new itowns.THREE.Raycaster();
                 var htmlInfo = document.getElementById('info');
                 htmlInfo.innerHTML = ' ';
 
-                raycaster.setFromCamera(mouse, globe.camera.camera3D);
-                // calculate objects intersecting the picking ray
-                var intersects = raycaster.intersectObjects($3dTilesLayerRequestVolume.object3d.children, true);
+                const intersects = view.pickObjectsAt(
+                    event,
+                    $3dTilesLayerRequestVolume);
 
                 for (var i = 0; i < intersects.length; i++) {
                     var interAttributes = intersects[i].object.geometry.attributes;
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -1,0 +1,147 @@
+import * as THREE from 'three';
+import TileMesh from './TileMesh';
+import RendererConstant from '../Renderer/RendererConstant';
+import { unpack1K } from '../Renderer/LayeredMaterial';
+
+function hideEverythingElse(view, object, threejsLayer = 0) {
+    // We want to render only 'object' and its hierarchy.
+    // So if it uses threejsLayer defined -> force it on the camera
+    // (or use the default one: 0)
+    const prev = view.camera.camera3D.layers.mask;
+
+    view.camera.camera3D.layers.mask = 1 << threejsLayer;
+
+    return () => {
+        view.camera.camera3D.layers.mask = prev;
+    };
+}
+
+const depthRGBA = new THREE.Vector4();
+// TileMesh picking support function
+function screenCoordsToNodeId(view, tileLayer, mouse) {
+    const dim = view.mainLoop.gfxEngine.getWindowSize();
+
+    mouse = mouse || new THREE.Vector2(Math.floor(dim.x / 2), Math.floor(dim.y / 2));
+
+    const restore = tileLayer.level0Nodes.map(n => n.pushRenderState(RendererConstant.ID));
+
+    const undoHide = hideEverythingElse(view, tileLayer.object3d, tileLayer.threejsLayer);
+
+    var buffer = view.mainLoop.gfxEngine.renderViewTobuffer(
+        { camera: view.camera, scene: tileLayer.object3d },
+        view.mainLoop.gfxEngine.fullSizeRenderTarget,
+        mouse.x, dim.y - mouse.y,
+        1, 1);
+
+    undoHide();
+
+    restore.forEach(r => r());
+
+    depthRGBA.fromArray(buffer).divideScalar(255.0);
+
+    // unpack RGBA to float
+    var unpack = unpack1K(depthRGBA, Math.pow(256, 3));
+
+    return Math.round(unpack);
+}
+
+function findLayerIdInParent(obj) {
+    if (obj.layer) {
+        return obj.layer;
+    }
+    if (obj.parent) {
+        return findLayerIdInParent(obj.parent);
+    }
+}
+
+const raycaster = new THREE.Raycaster();
+
+export default {
+    pickTilesAt: (_view, mouse, layer) => {
+        const results = [];
+        const _id = screenCoordsToNodeId(_view, layer, mouse);
+
+        const extractResult = (node) => {
+            if (node.id === _id && node instanceof TileMesh) {
+                results.push({
+                    object: node,
+                    layer: layer.id,
+                });
+            }
+        };
+        for (const n of layer.level0Nodes) {
+            n.traverse(extractResult);
+        }
+        return results;
+    },
+
+    pickPointsAt: (view, mouse, layer) => {
+        if (!layer.root) {
+            return;
+        }
+
+        const dim = view.mainLoop.gfxEngine.getWindowSize();
+
+        // enable picking mode for points material
+        layer.object3d.traverse((o) => {
+            if (o.isPoints && o.baseId) {
+                o.material.enablePicking(true);
+            }
+        });
+
+        const undoHide = hideEverythingElse(view, layer.object3d, layer.threejsLayer);
+
+        // render 1 pixel
+        // TODO: support more than 1 pixel selection
+        const buffer = view.mainLoop.gfxEngine.renderViewTobuffer(
+                { camera: view.camera, scene: layer.object3d },
+                view.mainLoop.gfxEngine.fullSizeRenderTarget,
+                mouse.x, dim.y - mouse.y, 1, 1);
+
+        undoHide();
+
+        // see PointCloudProvider and the construction of unique_id
+        const objId = (buffer[0] << 8) | buffer[1];
+        const index = (buffer[2] << 8) | buffer[3];
+
+        let result;
+        layer.object3d.traverse((o) => {
+            if (o.isPoints && o.baseId) {
+                // disable picking mode
+                o.material.enablePicking(false);
+
+                // if baseId matches objId, the clicked point belongs to `o`
+                if (!result && o.baseId === objId) {
+                    result = {
+                        object: o,
+                        index,
+                        layer: layer.id,
+                    };
+                }
+            }
+        });
+
+        if (result) {
+            return [result];
+        } else {
+            return [];
+        }
+    },
+
+    /*
+     * Default picking method. Uses THREE.Raycaster
+     */
+    pickObjectsAt(view, mouse, object, target = []) {
+        // raycaster use NDC coordinate
+        raycaster.setFromCamera(
+            view.viewToNormalizedCoords(mouse),
+            view.camera.camera3D);
+        const intersects = raycaster.intersectObject(object, true);
+        for (const inter of intersects) {
+            inter.layer = findLayerIdInParent(inter.object);
+            target.push(inter);
+        }
+
+        return target;
+    },
+};
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -5,7 +5,6 @@ import { RENDERING_PAUSED, MAIN_LOOP_EVENTS } from '../MainLoop';
 import { COLOR_LAYERS_ORDER_CHANGED } from '../../Renderer/ColorLayersOrdering';
 import RendererConstant from '../../Renderer/RendererConstant';
 import GlobeControls from '../../Renderer/ThreeExtended/GlobeControls';
-import { unpack1K } from '../../Renderer/LayeredMaterial';
 
 import { GeometryLayer } from '../Layer/Layer';
 
@@ -18,6 +17,7 @@ import { updateLayeredMaterialNodeImagery, updateLayeredMaterialNodeElevation } 
 import { globeCulling, preGlobeUpdate, globeSubdivisionControl, globeSchemeTileWMTS, globeSchemeTile1 } from '../../Process/GlobeTileProcessing';
 import BuilderEllipsoidTile from './Globe/BuilderEllipsoidTile';
 import SubdivisionControl from '../../Process/SubdivisionControl';
+import Picking from '../Picking';
 
 /**
  * Fires when the view is completely loaded. Controls and view's functions can be called then.
@@ -71,7 +71,6 @@ export const GLOBE_VIEW_EVENTS = {
     LAYER_REMOVED: 'layer-removed',
     COLOR_LAYERS_ORDER_CHANGED,
 };
-
 
 export function createGlobeLayer(id, options) {
     // Configure tiles
@@ -176,6 +175,9 @@ export function createGlobeLayer(id, options) {
         enable: false,
         position: { x: -0.5, y: 0.0, z: 1.0 },
     };
+    // provide custom pick function
+    wgs84TileLayer.pickObjectsAt = (_view, mouse) => Picking.pickTilesAt(_view, mouse, wgs84TileLayer);
+
     return wgs84TileLayer;
 }
 
@@ -360,8 +362,8 @@ GlobeView.prototype.removeLayer = function removeImageryLayer(layerId) {
 };
 
 GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
-    // update the picking ray with the camera and mouse position
-    const selectedId = this.screenCoordsToNodeId(mouse);
+    const picked = this.wgs84TileLayer.pickObjectsAt(this, mouse);
+    const selectedId = picked.length ? picked[0].object.id : undefined;
 
     for (const n of this.wgs84TileLayer.level0Nodes) {
         n.traverse((node) => {
@@ -380,41 +382,12 @@ GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
     this.notifyChange(true);
 };
 
-GlobeView.prototype.screenCoordsToNodeId = function screenCoordsToNodeId(mouse) {
-    const dim = this.mainLoop.gfxEngine.getWindowSize();
-
-    mouse = mouse || new THREE.Vector2(Math.floor(dim.x / 2), Math.floor(dim.y / 2));
-
-    const previousRenderState = this._renderState;
-    this.changeRenderState(RendererConstant.ID);
-
-    // Prepare state
-    const prev = this.camera.camera3D.layers.mask;
-    this.camera.camera3D.layers.mask = 1 << this.wgs84TileLayer.threejsLayer;
-
-    var buffer = this.mainLoop.gfxEngine.renderViewTobuffer(
-        this,
-        this.mainLoop.gfxEngine.fullSizeRenderTarget,
-        mouse.x, dim.y - mouse.y,
-        1, 1);
-
-    this.changeRenderState(previousRenderState);
-    this.camera.camera3D.layers.mask = prev;
-
-    var depthRGBA = new THREE.Vector4().fromArray(buffer).divideScalar(255.0);
-
-    // unpack RGBA to float
-    var unpack = unpack1K(depthRGBA, Math.pow(256, 3));
-
-    return Math.round(unpack);
-};
-
 GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, height) {
     const g = this.mainLoop.gfxEngine;
-    const previousRenderState = this._renderState;
-    this.changeRenderState(RendererConstant.DEPTH);
+    const restore = this.wgs84TileLayer.level0Nodes.map(n => n.pushRenderState(RendererConstant.DEPTH));
     const buffer = g.renderViewTobuffer(this, g.fullSizeRenderTarget, x, y, width, height);
-    this.changeRenderState(previousRenderState);
+    restore.forEach(r => r());
+
     return buffer;
 };
 
@@ -476,26 +449,6 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
         { return undefined; }
 
     return pickWorldPosition;
-};
-
-GlobeView.prototype.changeRenderState = function changeRenderState(newRenderState) {
-    if (this._renderState == newRenderState || !this.wgs84TileLayer.level0Nodes) {
-        return;
-    }
-
-    // build traverse function
-    var changeStateFunction = (function getChangeStateFunctionFn() {
-        return function changeStateFunction(object3D) {
-            if (object3D.changeState) {
-                object3D.changeState(newRenderState);
-            }
-        };
-    }());
-
-    for (const n of this.wgs84TileLayer.level0Nodes) {
-        n.traverseVisible(changeStateFunction);
-    }
-    this._renderState = newRenderState;
 };
 
 GlobeView.prototype.setRealisticLightingOn = function setRealisticLightingOn(value) {
--- a/src/Core/Scheduler/Providers/PointCloudProvider.js
+++ b/src/Core/Scheduler/Providers/PointCloudProvider.js
@@ -3,6 +3,7 @@ import Fetcher from './Fetcher';
 import PointCloudProcessing from '../../../Process/PointCloudProcessing';
 import PotreeBinLoader from './PotreeBinLoader';
 import PotreeCinLoader from './PotreeCinLoader';
+import Picking from '../../Picking';
 
 // Create an A(xis)A(ligned)B(ounding)B(ox) for the child `childIndex` of one aabb.
 // (PotreeConverter protocol builds implicit octree hierarchy by applying the same
@@ -160,6 +161,8 @@ export default {
         layer.update = PointCloudProcessing.update;
         layer.postUpdate = PointCloudProcessing.postUpdate;
 
+        // this probably needs to be moved to somewhere else
+        layer.pickObjectsAt = (view, mouse) => Picking.pickPointsAt(view, mouse, layer);
 
         return Fetcher.json(`${layer.url}/${layer.file}`, layer.fetchOptions).then((cloud) => {
             layer.metadata = cloud;
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -42,6 +42,8 @@ function TileMesh(geometry, params) {
     this.layerUpdateState = {};
 
     this.material.setUuid(this.id);
+
+    this._state = RendererConstant.FINAL;
 }
 
 TileMesh.prototype = Object.create(THREE.Mesh.prototype);
@@ -70,6 +72,9 @@ TileMesh.prototype.isDisplayed = function isDisplayed() {
 
 // switch material in function of state
 TileMesh.prototype.changeState = function changeState(state) {
+    if (state == this._state) {
+        return;
+    }
     if (state == RendererConstant.DEPTH) {
         this.material.defines.DEPTH_MODE = 1;
         delete this.material.defines.MATTE_ID_MODE;
@@ -81,7 +86,28 @@ TileMesh.prototype.changeState = function changeState(state) {
         delete this.material.defines.DEPTH_MODE;
     }
 
+    this._state = state;
+
     this.material.needsUpdate = true;
+};
+
+function applyChangeState(n, s) {
+    if (n.changeState) {
+        n.changeState(s);
+    }
+}
+
+TileMesh.prototype.pushRenderState = function pushRenderState(state) {
+    if (this._state == state) {
+        return () => { };
+    }
+
+    const oldState = this._state;
+    this.traverse(n => applyChangeState(n, state));
+
+    return () => {
+        this.traverse(n => applyChangeState(n, oldState));
+    };
 };
 
 TileMesh.prototype.setFog = function setFog(fog) {
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -1,11 +1,12 @@
-/* global window, requestAnimationFrame */
-import { Scene, EventDispatcher, Vector2 } from 'three';
+/* global window */
+import { Scene, EventDispatcher, Vector2, Object3D } from 'three';
 import Camera from '../Renderer/Camera';
 import MainLoop from './MainLoop';
 import c3DEngine from '../Renderer/c3DEngine';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from './Layer/LayerUpdateStrategy';
 import { GeometryLayer, Layer, defineLayerProperty } from './Layer/Layer';
 import Scheduler from './Scheduler/Scheduler';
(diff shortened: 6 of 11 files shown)


src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -130,6 +130,51 @@ function loadPointFile(layer, url) {
     });
 }
 
+function pick(view, mouse, layer) {
```
gchoqueux: Good refacto and unification.
However I have an objection on the choice of files for the picking declaration : the providers 

To be more generic, clearer and to reuse,  should be in a picking toolbox.
These tools must have as input the type of object to pick.

Then the layer's picking functionality must be assigned to its construction, but not in the pre-process phase of the provider. Because the provider must take care to supply only what has in title and pre-process works only for this supply.

it's same case in PR #546 
what is your motivation to put this in the providers?

I also ask the opinion to  @mbredif ?

peppsac: > To be more generic, clearer and to reuse, should be in a picking toolbox.
> [...]
>  what is your motivation to put this in the providers?

Yes; we already had this discussion earlier, so I won't copy-paste all my arguments from there.

I can move `pickObjectsAt` to a `Picking.js` file if it you prefer.

But I still believe that `preprocessing` of geometry Providers is a good place to extend layer's object so I'll leave the `layer.pickObjectsAt` there.

peppsac: Done.

src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -9,6 +9,31 @@ import TileGeometry from '../../TileGeometry';
 import TileMesh from '../../TileMesh';
 import CancelledCommandException from '../CancelledCommandException';
 import { requestNewTile } from '../../../Process/TiledNodeProcessing';
+import RendererConstant from '../../../Renderer/RendererConstant';
+import Picking from '../../Picking';
+
+function changeRenderState(tileLayer) {
```
gchoqueux: All references to the layer should disappear from `tileProvider`. Because there isn't reason for a tile provider to know the layers. For example, in the future, remove any notion of layer in `TileProvider.prototype.executeCommand`. You have to be able to use the tile provider without a layer concept. 

`preprocessDataLayer` also has no reason to be here and even less if that add functions that aren't `tile building`

So don't add new reverence to the layer

gchoqueux (changes requested):

src/Core/Scheduler/Providers/TileProvider.js
autra: > Because there isn't reason for a tile provider to know the layers

Every single provider of itowns absolutely *does* know about layers. Providers are here to interface layers with 2 things: 
- a protocol by:
    - initializing the layer with some default values for this protocol and / or metadata from the network, following the protocol specifications (getCapabilities, tileset.json...)
    - abstract the fetching of data from servers implementing this protocol
- The scheduler mechanism, because we're fetching datas from the internet. And it uses the layer because it provides some configurations that `executeCommand` needs to generate new nodes (opacity, colors, overrideMaterials etc...)

And in turns, the way we create node for a given protocol have an influence on how we should pick it on the scene. So it makes a lot of sense that this logic should be at least partly in providers.

> preprocessDataLayer also has no reason to be here

For the aforementioned reason, it has all the reason to be in a provider

Not sure I understand you well. Could you describe what alternative solution you would choose? Where would you move all that code then? It might make things clearer.

peppsac: Moved more code so now TileProvider isn't modified.

Is this what you had in mind @gchoqueux ?

gchoqueux: The best place for picking is the construction of the geometric layer using the three.js raycaster caster picking. And then overload it in more specific cases like the tile layer. In this way, there is no need to add the picking in each provider of geometry.
@autra  It's clearer?

autra: Somehow... 

> And then overload it in more specific cases like the tile layer

Where would you overload it then? I found that providers was a good place for that (because they constructs the objects, so they know how they should be picked). That being said I'm fine with the current version.

src/Core/Picking.js
```diff
@@ -0,0 +1,98 @@
...
+    var depthRGBA = new THREE.Vector4().fromArray(buffer).divideScalar(255.0);
+
+    // unpack RGBA to float
+    var unpack = unpack1K(depthRGBA, Math.pow(256, 3));
+
+    return Math.round(unpack);
+}
+
+export default {
+    pickTilesAt: (_view, mouse, layer) => {
```
gchoqueux: The picking function shouldn't have as parameter the layer but the objects to pick.
Because this function must be usable without the notion of layer

src/Core/View.js
```diff
@@ -458,4 +458,106 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
...
+ * property which is the Object3D under the cursor. Then depending on the queried
+ * layer/source, there may be additionnal properties (coming from THREE.Raycaster
+ * for instance).
+ *
+ * @example
+ * view.pickObjectsAt({ x, y })
+ * view.pickObjectsAt({ x, y }, 'wfsBuilding')
+ * view.pickObjectsAt({ x, y }, 'wfsBuilding', myLayer)
+ */
+View.prototype.pickObjectsAt = function pickObjectsAt(mouse, ...where) {
```
gchoqueux: why `pickObjectsAt` isn't found in picking.js

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -160,6 +161,7 @@ export default {
         layer.update = PointCloudProcessing.update;
         layer.postUpdate = PointCloudProcessing.postUpdate;
 
+        layer.pickObjectsAt = (view, mouse) => Picking.pickPointsAt(view, mouse, layer);
```
gchoqueux: `pickObjectsAt` must be moved in geometry layer

gchoqueux (changes requested):

src/Core/View.js
peppsac: Because Picking.js is the implementation of the various picking methods (I'm also moving `raycastObjectsAt` in Picking.js)

`View.prototype.pickObjectsAt` is the boilerplate needed to use the pickings methods on a view.

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: (small aside: please avoid this wording. "You must do this" is kinda rude)

There's no perfect place for this kind of code actually. PR #546 aimed to fix this but it couldn't be merged yet...

Here you're nitpicking on one line and your argument "Provider is to provide things" can be used to not move the picking code in `Layer.js` : "Layer should only contain generic things but Picking.pickPointsAt is a custom pick code".

So I suggest we move and work on improving #546 (or submit another PR that would solve this problem).

src/Core/View.js
peppsac: See previous answer.

It's boilerplate code to ease picking usage from the View.

peppsac:
BRanch updated.

gchoqueux:
There is also `FeatureUtils` there is also foo as a method of picking

peppsac:
> There is also FeatureUtils there is also foo as a method of picking

`FeatureUtils.filterFeaturesUnderCoordinate` is picking from a `Coordinates` object while this PR is focusing on objects selection from mouse coordinates.

Later, it could be interesting to extend the picking API to any kind of coordinates (pixels  or geographic) though.

gchoqueux:
Without picking feature, so idea to expose a single entry point for querying isn't completed?

peppsac:
> Without picking feature, so idea to expose a single entry point for querying isn't completed?

You can pick features at given window coordinates: `view.pickObjectsAt(mouse, wfsLayer)` will return all features under your mouse (or you can use Picking.pickObjectsAt(mouse, camera, someObject3d)` if you prefer, but for wfs the first form is easier since wfs objects are attached to TileMesh).

src/Core/View.js
gchoqueux: ok

gchoqueux:
and for 2D vector features?

peppsac:
> and for 2D vector features?

This function can pick anything that is rendered as a Mesh. Does that answer your question?

src/Core/Scheduler/Providers/PointCloudProvider.js
gchoqueux: It can have several providers return the same type of object, THREE.Object3D for example, it makes more sense to move it in the construction of the geometric layer than to put picking code in  each provider that returns an object3D

gchoqueux:
yes, the PR title and commit should be changed to : `unify picking mesh methods`

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: > put  picking code in each provider that returns an object3D

- I'm doing this solely for the PointCloudProvider
- this line will go away as soon as #546 is merged
- I disagree that putting this pointcloud specific line is better in GeometryLayer constructor.

Will you block this PR for this?

gchoqueux: Are you ok with this [comment](https://github.com/iTowns/itowns/pull/635#discussion_r166295422)?

Otherwise, I agree, it makes more sense, first of all, to work and merge #546

peppsac: No.
I'd prefer to first merge this one, and then merge the more complicated one.

peppsac: Reminder: this PR doesn't remove nor add feature. It simply groups the picking-from-window-coordinate in the same module. I don't even see why we're arguing so much about it :(

peppsac:
Added a small change : `Picking` module also returns the `layer.id` of the object (if it exists).

peppsac:
So can I rebase this PR (and change the title) or are there any more comments?

peppsac:
Rebased. And added a small fix: make sure we only display objects from the layer when doing the picking (see https://github.com/iTowns/itowns/pull/635/files#diff-11ecfe7e737c10e75920ac54de872437R8)

peppsac:
Rebased since #637 was merged.

peppsac:
@gchoqueux can you update your review of this PR please? (or dismiss it if you don't have time to update your review)

gchoqueux:
there's [comment](https://github.com/iTowns/itowns/pull/635#discussion_r166301777) without reply 
and I add this previous [comment](https://github.com/iTowns/itowns/pull/637#issuecomment-364384095)

peppsac:
> there's comment without reply

I've answered to all comments in this thread 6 days ago so...
If I missed something please be more specific.

>  and I add this previous comment

which I also answered.

gchoqueux:
> I disagree that putting this pointcloud specific line is better in GeometryLayer constructor.

Why?

peppsac:
> Why

Already answered as well in the comment  [above](https://github.com/iTowns/itowns/pull/635#discussion_r165957424): `Layer.js` should only contains generic code that applies to all layers, and `layer.pickObjectsAt = (view, mouse) => Picking.pickPointsAt(....)` is pointcloud generic.
And a more important argument: IMHO this dicussion is ±pointless because PR #546 should reorganize the code properly.
And see also the issue #182 that details the need for refactoring regarding Providers.

So unless you feel this is really important and you intend to block this PR because of this, I consider this a non issue.

examples/3dtiles.html
```diff
@@ -152,15 +152,12 @@
...
-                // Pick an object with batch id
-                var mouse = globe.eventToNormalizedCoords(event);
-                var raycaster = new itowns.THREE.Raycaster();
                 var htmlInfo = document.getElementById('info');
                 htmlInfo.innerHTML = ' ';
 
-                raycaster.setFromCamera(mouse, globe.camera.camera3D);
-                // calculate objects intersecting the picking ray
-                var intersects = raycaster.intersectObjects($3dTilesLayerRequestVolume.object3d.children, true);
+                const intersects = view.pickObjectsAt(
```
gchoqueux, line 158: i'll be better than that goes on like this:
```js
layer.on('mouseMove', (event) => { 
    const intersects = event.intersects;
}
```

gchoqueux (changes requested):

examples/3dtiles.html
peppsac: Feel free to implement `.on()` for layers if you think it's useful, but that's not the scope of this PR.

gchoqueux:
With approach `viewer.on`

* remove methods, `eventToViewCoords,  eventToNormalizedCoords, viewToNormalizedCoords and normalizedToViewCoords` in itowns's core and all users cases. Because their responsibilities are unloaded to the `viewerEvent`
* the same event `viewerEvent` is created and is dispatched to all listeners (which avoids redundant calculations)
* simplifies the code

```js
function picking(event) {
        var intersects = view.pickObjectsAt(view.eventToViewCoords(event), layer);
}
window.addEventListener('mousemove', picking, false);
```
become
```js
 function picking(event) {
        var intersects = event.intersects;
 }
 view.on('mousemove', picking, layer);
```
-------------------------------

**An example of partial implementation**
```js
const viewerListeners = {};
const raycaster = new Raycaster();
function viewerEvent(type, event, viewer) {
    const _event = {
        type,
        viewCoords: _eventCoords.set(event.offsetX, event.offsetY),
        event,
    };
    let normalizedCoords;
    let intersects;
    Object.defineProperty(
        _event,
        'normalizedCoords',
        {
            get: () => {
                if (!normalizedCoords) {
                    _eventCoords.x = 2 * (_event.viewCoords.x / viewer.camera.width) - 1;
                    _eventCoords.y = -2 * (_event.viewCoords.y / viewer.camera.height) + 1;
                    normalizedCoords = _eventCoords;
                }
                return normalizedCoords;
            },
        });
    Object.defineProperty(
    _event,
    'intersects',
        {
            get: () => {
                if (!intersects) {
                    if (_event.any instanceof GeometryLayer) {
                        raycaster.setFromCamera(_event.normalizedCoords, viewer.camera.camera3D);
                        intersects = raycaster.intersectObjects(_event.any.object3d.children, true);
                    } else {
                        intersects = [];
                    }
                }
                return intersects;
            },
        });

    return _event;
}

View.prototype.on = function _on(type, listener, any) {
    if (!viewerListeners[type]) {
        const domElement = this.mainLoop.gfxEngine.renderer.domElement;
        viewerListeners[type] = (event) => {
            this.dispatchEvent(viewerEvent(type, event, this));
        };
        if (type == 'click') {
            domElement.addEventListener('mousedown', viewerListeners[type]);
        } else if (type == 'mousemove') {
            domElement.addEventListener('mousemove', viewerListeners[type]);
        }
    }
    this.addEventListener(type, (event) => {
        event.any = any || { object3d: this.scene };
        listener(event);
    });
};
```

peppsac:
Your proposed implementation won't work properly for pointclouds (raycasting in millions of points doesn't work well).
You want to replace the line:
```js
   intersects = raycaster.intersectObjects(_event.any.object3d.children, true);
```
With:
```js
   intersects  = _event.any.pickOjectsAt(viewer, viewer.eventToViewCoords(_event));
```
To get something that works with all types of GeometryLayers (and of course `pickOjectsAt` is exactly the function added by this PR).

And again, sorry but your `view.on(...)` proposal isn't relevant for this PR (already said so [earlier](https://github.com/iTowns/itowns/pull/635#discussion_r167595186)).

gchoqueux:
The `view.on (...)` approach doesn't prevent appropriate processing for each type of `GeometryLayer`
Just define it in the `intersects` getter property

autra:
Another PR for `view.on` please. We need to measure all the impacts with a working implementation (I'm not against it in itself btw). It's largely orthogonal with this changeset.

peppsac:
@gchoqueux you missed the most important part of my comment : "And again, sorry but your view.on(...) proposal isn't relevant for this PR (already said so earlier).".
Maybe the solution here is not the best but it's an improvment over the current code.

gchoqueux:
I didn't miss it. I will study this PR, so that it can be easily improved towards the proposal `view.on()`.

autra:
@gchoqueux so you're blocking a PR for a proposal that is:
- unrelated to it
- won't even touch the same line numbers

why?

peppsac:
PR rebased and commits squashed.
@gchoqueux there's nothing here that will prevent implementing the `.on(...)` mechanism.

src/Core/Picking.js
```diff
@@ -0,0 +1,159 @@
...
+        view,
+        view.mainLoop.gfxEngine.fullSizeRenderTarget,
+        mouse.x, dim.y - mouse.y,
+        1, 1);
+
+    undoHide();
+
+    restore.forEach(r => r());
+
+    var depthRGBA = new THREE.Vector4().fromArray(buffer).divideScalar(255.0);
```
gchoqueux: declare `depthRGBA` outside the function

src/Core/Picking.js
```diff
@@ -0,0 +1,159 @@
...
+
+    fakeScene.add(object);
+
+    view.scene = fakeScene;
+
+    view.camera.camera3D.layers.mask = 1 << threejsLayer;
+
+    return () => {
+        view.scene = oldScene;
+        oldParent.add(object);
```
gchoqueux: In cloud globe example, `oldParent` could be null on first picking

src/Core/View.js
```diff
@@ -509,4 +510,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+        if (typeof (source[i]) === 'string') {
+            const lookup = this.getLayers(l => l.id == source[i]);
+            if (!lookup.length) {
+                throw new Error(`Invalid layer id used as where argument (value = ${where})`);
+            }
+            source[i] = lookup[0];
+        }
+    }
+
+    for (let i = 0; i < source.length; i++) {
```
gchoqueux: could be replace by `for (const s of source)`?

src/Core/View.js
```diff
@@ -509,4 +510,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+ * view.pickObjectsAt({ x, y }, 'wfsBuilding')
+ * view.pickObjectsAt({ x, y }, 'wfsBuilding', myLayer)
+ */
+View.prototype.pickObjectsAt = function pickObjectsAt(mouse, ...where) {
+    const results = [];
+    const source = where.length == 0 ?
+        this.getLayers(l => l.type == 'geometry') :
+        [...where];
+
+    for (let i = 0; i < source.length; i++) {
```
gchoqueux: you convert all the layer'id string to layer?

src/Core/View.js
```diff
@@ -509,4 +510,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+                //   - it hasn't: this layer is attached to another one
+                let parentLayer;
+                this.getLayers((l, p) => {
+                    if (l.id == source[i].id) {
+                        parentLayer = p;
+                    }
+                });
+
+                const obj = [];
+                // raycast using parent layer object3d
```
gchoqueux: by example, if `source[i]`is point cloud layer the `raycast` method isn't appropriate

src/Core/View.js
```diff
@@ -509,4 +510,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+                Picking.pickObjectsAt(
```
gchoqueux: ~~we could add layers filter parameter in the parent's method `pickObjectsAt`~~

src/Core/Picking.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+    },
+
+    /*
+     * Default picking method. Uses THREE.Raycaster
+     */
+    pickObjectsAt(mouse, camera, object, target = []) {
+        // raycaster use NDC coordinate
+        const ndc = {
+            x: 2 * (mouse.x / camera.width) - 1,
```
gchoqueux: `View.prototype.viewToNormalizedCoords `could be a static method and used here (and avoid redundant code)

src/Core/Picking.js
```diff
@@ -0,0 +1,159 @@
...
+
+    var depthRGBA = new THREE.Vector4().fromArray(buffer).divideScalar(255.0);
+
+    // unpack RGBA to float
+    var unpack = unpack1K(depthRGBA, Math.pow(256, 3));
+
+    return Math.round(unpack);
+}
+
+function findLayerIdInParent(obj) {
```
gchoqueux, line 48: Could you used `Object3D.traverseAncestors` ?

src/Core/Picking.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+    if (obj.parent) {
+        return findLayerIdInParent(obj.parent);
+    }
+}
+
+const raycaster = new THREE.Raycaster();
+
+export default {
+    pickTilesAt: (_view, mouse, layer) => {
```
gchoqueux, line 60: it should also move `getPickingPositionFromDepth` in picking.js

src/Core/Picking.js
```diff
@@ -0,0 +1,159 @@
...
+            if (o.isPoints && o.baseId) {
+                o.material.enablePicking(true);
+            }
+        });
+
+        const undoHide = hideEverythingElse(view, layer.object3d, layer.threejsLayer);
+
+        // render 1 pixel
+        // TODO: support more than 1 pixel selection
+        const buffer = view.mainLoop.gfxEngine.renderViewTobuffer(
```
gchoqueux, line 96: may be change function parameters to `renderViewTobuffer(scene, camera, x, y, width, height)` to clarify the scene switch in the viewer (and avoid `fakeScene`)

src/Core/Prefab/GlobeView.js
```diff
@@ -363,8 +365,8 @@ GlobeView.prototype.removeLayer = function removeImageryLayer(layerId) {
 };
 
 GlobeView.prototype.selectNodeAt = function selectNodeAt(mouse) {
-    // update the picking ray with the camera and mouse position
-    const selectedId = this.screenCoordsToNodeId(mouse);
+    const picked = this.wgs84TileLayer.pickObjectsAt(this, mouse);
+    const selectedId = picked.length ? picked[0].object.id : undefined;
 
     for (const n of this.wgs84TileLayer.level0Nodes) {
```
gchoqueux, line 368: it is necessary to browse the tree so we have the node in `picked [0] .object`?

src/Core/TileMesh.js
```diff
@@ -81,9 +86,30 @@ TileMesh.prototype.changeState = function changeState(state) {
...
+    }
+}
+
+TileMesh.prototype.pushRenderState = function pushRenderState(state) {
+    if (this._state == state) {
+        return () => { };
+    }
+
+    const oldState = this._state;
+    this.traverse(n => applyChangeState(n, state));
```
gchoqueux, line 106: i don't know whether to check if `n` is an instance of `TileMesh`?

gchoqueux (changes requested):

src/Core/Picking.js
peppsac: This shouldn't be possible? Can you be more specific?

src/Core/Picking.js
peppsac: Done.

src/Core/View.js
peppsac: Yes.

src/Core/View.js
peppsac: That's why we don't do raycast. We use instead the specialized `pickObjectsAt` method (see 10 lines above).

src/Core/View.js
peppsac: Sorry I don't understand your comment.

src/Core/Picking.js
peppsac: Instead I'll change the signature, to match the layer pickObjectsAt function:
```js
pickObjectsAt(view, mouse, object, target = []) { ... }
```

src/Core/Picking.js
peppsac: No, because `traverseAncestors` can't be stopped as soon as you find what you were looking for.

src/Core/Picking.js
peppsac: Well no.
For now these functions are only valid for tile based layer - so they'll stay in their respective View implementation until we provide a way to query this for ~all geometry layers.

src/Core/Picking.js
peppsac: Let's discuss the change to this function in the dedicated PR: https://github.com/iTowns/itowns/pull/672

In the meantime this PR can use the fakeScene trick because it's harmless.

src/Core/Prefab/GlobeView.js
peppsac: I'm not sure to understand what you mean.
But we have to browse the tree to select/deselect tiles based on the picking result.

src/Core/TileMesh.js
peppsac: It's not needed since `applyChangeState` checks for `changeState` existence.

src/Core/Picking.js
peppsac: I simplified this code and removed the need for the `fakeScene`

src/Core/Picking.js
gchoqueux: in pointcloud_globe.html example, in the first double click, `oldParent` is null

src/Core/View.js
gchoqueux: ok

src/Core/View.js
gchoqueux: forget the previous comment

Why geometry layer doesn't have default `pickObjectsAt` method?

peppsac: All GeometryLayer have a default picking method (see https://github.com/iTowns/itowns/pull/635/files#diff-61e012c3e00bdc510be33d9eac684b2aR80)

But the method in View support picking for all objects (for instance raw THREE.Object3D) so we need to use `Picking.pickObjectsAt` as a fallback.

src/Core/View.js
```diff
@@ -512,4 +513,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+ * view.pickObjectsAt({ x, y }, 'wfsBuilding')
+ * view.pickObjectsAt({ x, y }, 'wfsBuilding', myLayer)
+ */
+View.prototype.pickObjectsAt = function pickObjectsAt(mouse, ...where) {
+    const results = [];
+    const sources = where.length == 0 ?
+        this.getLayers(l => l.type == 'geometry') :
+        [...where];
+
+    for (let i = 0; i < sources.length; i++) {
```
gchoqueux: can be a single loop is sufficient?

src/Core/Picking.js
gchoqueux: `pickTilesAt` is only valid for tile based layer?

src/Core/Prefab/GlobeView.js
gchoqueux: in other PR: it will later be necessary to keep the selection to avoid traverse the tree to update the nodes

src/Core/TileMesh.js
gchoqueux: yes

src/Core/Picking.js
peppsac: yep

src/Core/View.js
peppsac: It could be done in a single loop but I think it helpes readability to separate the 2 loops.

peppsac:
I think I've answered/fixed to all your comments. Did I miss something?

src/Core/Picking.js
gchoqueux: your proposed picking allows you to get position for which type of geometry layers?

src/Core/View.js
gchoqueux: With a comment will be more readable
and could you replace by `for (const source of sources)`

src/Core/View.js
```diff
@@ -512,4 +513,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+            } else {
+                //   - it hasn't: this layer is attached to another one
+                let parentLayer;
+                this.getLayers((l, p) => {
+                    if (l.id == source.id) {
+                        parentLayer = p;
+                    }
+                });
+
+                const obj = [];
```
gchoqueux: use of a target isn't necessary

```js
const obj = Picking.pickObjectsAt(
                    this,
                    mouse,
                    parentLayer.object3d);
```

src/Core/View.js
gchoqueux: > for instance raw THREE.Object3D 

Should be explain in the title commit because it doesn't unify the picking position/coordinates and features (in raster)

src/Core/View.js
```diff
@@ -512,4 +513,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
     return _eventCoords;
 };
 
+/**
+ * Return objects from some layers/objects3d under the mouse in this view.
```
gchoqueux: returns only object3D?

src/Core/Scheduler/Providers/PointCloudProvider.js
gchoqueux: add a temporary comment : to indicate that it'll not stay in the provider

src/Core/View.js
peppsac: Fixed both comments.

peppsac: And also adapted pickObjectsAt to accept either mouse coords or a (Mouse|Touch)Event

src/Core/View.js
peppsac: See the @return documentation for the returned values.

peppsac:
I'll change the commit title when squashing the commits.

src/Core/View.js
```diff
@@ -512,4 +513,82 @@ View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoord
...
+        if (source.type === 'geometry') {
+            // does this layer have a custom picking function?
+            if (source.pickObjectsAt) {
+                results.splice(
+                    results.length, 0,
+                    ...source.pickObjectsAt(this, mouse));
+            } else {
+                //   - it hasn't: this layer is attached to another one
+                let parentLayer;
+                this.getLayers((l, p) => {
```
gchoqueux: if a geometric layer is attached to another geometric layer.
We could affect it the reference of `parentLayer.object3d` in the `GeometryLayer.prototype.attach` function
`layer.object3d = this.object3d`

Who would avoid the party with `it hasn't: this layer is attached to another one`

WDYT?

peppsac: In other places of itowns, layer.object3d is the object used to add layer's mesh to the scene.

So I don't think it make sense for now to store parent layer's object3D here.

gchoqueux:
OK rebase, squash and re word commit then we merge

peppsac:
Done

gchoqueux (approved):
thanks

PR #636, opened 2018-02-02
feat(core): add a loading screen

peppsac:
This PR adds a simple feature: a loading screen displaying an animated
itowns logo while waiting for the View to be ready to display.

The loading screen is also dismissed after 3 sec as well.

This feature can be disabled, using the disableLoadingScreen View option.

Loading screen logo :

capture d ecran de 2018-02-02 15-20-44


zarov:
This demo is a nice addition!

But, IMHO, this PR should only add this in an example; all the core stuff added by this PR doesn't belong to iTowns.

iTowns must allow user to easily implement this but shouldn't have an iTowns logo built-in, as a user may want to replace the logo by its own.

peppsac:
> iTowns must allow user to easily implement this but shouldn't have an iTowns logo built-in, as a user may want to replace the logo by its own.

I understand your point. On the other hand:
- the loading screen is easy to disable (pass `disableLoadingScreen` to the View constructor)
- it doesn't prevent users to implement their own version

And IMHO having a default loading page is better than not having one :)

src/utils/LoadingScreen.css
```diff
@@ -0,0 +1,80 @@
...
+#itowns-loader > .text {
+  text-align: center;
+  font-size: 5em;
+}
+
+.itowns-invisible {
+  -webkit-transition: opacity 0.5s cubic-bezier(0.55, 0.085, 0.68, 0.53);
+  -moz-transition: opacity 0.5s cubic-bezier(0.55, 0.085, 0.68, 0.53);
+  -ms-transition: opacity 0.5s cubic-bezier(0.55, 0.085, 0.68, 0.53);
+  -o-transition: opacity 0.5s cubic-bezier(0.55, 0.085, 0.68, 0.53);
```
tbroyer: Comparing https://caniuse.com/#feat=webgl and https://caniuse.com/#feat=css-transitions (I could add flexbox and animations to the comparison matrix too), there's no overlap (besides Chrome for Android 12, would iTowns even work there?) so all those prefixes should be replaced with a simple, cross-browser, `transition` property.

gchoqueux:
In the first place it is better to put it in the examples

> And IMHO having a default loading page is better than not having one :)

this is argument isn't valid because with this assumption everything can be added in itowns from the moment that it isn't there

src/Core/View.js
```diff
@@ -79,6 +82,33 @@ function View(crs, viewerDiv, options = {}) {
...
+        // loading screen
+        viewerDiv.insertAdjacentHTML('beforeend', LoadingScreenHTML);
+        const node = document.createElement('style');
+        node.innerHTML = LoadingScreenCSS;
+        document.body.appendChild(node);
+
+        const view = this;
+        // auto-hide in 3 sec or if view is loaded
+        const hideLoader = () => {
+            const parent = document.getElementById('itowns-loader');
```
tbroyer: Doesn't that prevent using more than one instance of iTowns on the same page?

tbroyer (commented):
I'm a bit concerned about the increase in size (albeit probably relatively small as it's all text that should gzip well) of such a thing.

IFF we want to "impose" an iTowns splashscreen to all/most users, then it's fine. If it's only because we thing it _looks better_, then maybe we should rather make it easy to add it (any kind of splashscreen) to any iTowns "application" and properly document it. This could come in the form of a small additional library to make it as easy as possible to have the iTowns splashscreen, and a small code snippet to copy/paste for those who prefer their own splashscreen (and an additional JS file to require/import for those using iTowns through NPM).

Maybe the logic for the splashscreen could live into iTowns proper (it's 20 lines only), and have a simple API to trigger it using any HTML element (e.g. put your element inside your `viewerDir` and pass a reference to it to a `splashscreen` property; when that property is non-null, then the logic is triggered; and a small additional library could have the few lines of code to inject the iTowns splashscreen and configure that property).

gchoqueux:
@tbroyer : your idea is to facilitate the setting up of a loading screen in the viewer and to give the responsibility of the code and the module to the user?

tbroyer:
> @tbroyer : your idea is to facilitate the setting up of a loading screen in the viewer and to give the responsibility of the code and the module to the user?

Yes.

With the handling code in iTowns proper and triggered by a `splashscreen` option, you'd use it as something like:
```html
<script src="dist/itowns.js"></script>
<script src="dist/itownsplash.css"></script>
…
<div id="viewerDiv"></div>
<script>
…
const globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, {
  splashscreen: itownsplash.inject(viewerDir)
});
…
```
or with your own splashscreen:
```html
<script src="dist/itowns.js"></script>
…
<div id="viewerDiv">
  <svg id="mySplashScreen">…</svg>
</div>
<script>
…
const globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, {
  splashscreen: document.getElementById('mySplashScreen')
});
…
```

peppsac:
Modified the code, to support something similar to what you (@tbroyer) proposed (didn't test yet, but it should work).

The user can either:
- use the default loading screen (`if (!options.loadingScreen || options.loadingScreen === 'itowns') {`)
- pass an element id
- pass a HTMLElement

src/Core/View.js
```diff
@@ -79,6 +82,44 @@ function View(crs, viewerDiv, options = {}) {
...
+        if (!options.loadingScreen || options.loadingScreen === 'itowns') {
+            const node = document.createElement('style');
+            node.innerHTML = LoadingScreenCSS;
+            document.body.appendChild(node);
+            // loading screen
+            viewerDiv.insertAdjacentHTML('beforeend', LoadingScreenHTML);
+            loadingScreenContainer = document.getElementById('itowns-loader');
+        } else if (typeof (options.loadingScreen) === 'string') {
+            loadingScreenContainer = document.getElementById(options.loadingScreen);
+        } else if (options.loadingScreen instanceof (HTMLElement)) {
```
tbroyer: That should probably be just `Element` to make it possible to use inline SVGs (unless there are specific reasons not to support those, and require them to be embedded in a `<div>` for example)

src/Core/View.js
```diff
@@ -79,6 +82,44 @@ function View(crs, viewerDiv, options = {}) {
...
+        const view = this;
+        // auto-hide in 3 sec or if view is loaded
+        const hideLoader = () => {
+            if (!loadingScreenContainer) {
+                return;
+            }
+
+            loadingScreenContainer.style.opacity = 0;
+            loadingScreenContainer.style.pointerEvents = 'none';
+            loadingScreenContainer.style.transition = 'opacity 0.5s cubic-bezier(0.55, 0.085, 0.68, 0.53)';
```
tbroyer: Should this be deferred to CSS, to make it easily customizable?

src/Core/View.js
```diff
@@ -79,6 +82,44 @@ function View(crs, viewerDiv, options = {}) {
...
+            loadingScreenContainer = document.getElementById('itowns-loader');
+        } else if (typeof (options.loadingScreen) === 'string') {
+            loadingScreenContainer = document.getElementById(options.loadingScreen);
+        } else if (options.loadingScreen instanceof (HTMLElement)) {
+            loadingScreenContainer = options.loadingScreen;
+        } else {
+            throw new Error('Invalid value for options.loadingScreen. Expected: false, string or HTMLElement');
+        }
+
+        const view = this;
```
tbroyer: You don't need this with a fat-arrow function, just use `this` inside the function (or if you'd prefer this, e.g. for readability –this is subjective–, then maybe switch to a `function` instead?)

src/Core/View.js
```diff
@@ -79,6 +82,44 @@ function View(crs, viewerDiv, options = {}) {
...
+        // auto-hide in 3 sec or if view is loaded
+        const hideLoader = () => {
+            if (!loadingScreenContainer) {
```
tbroyer: This `if` should probably be outside the fat-arrow function, to avoid setting it up when not needed.

The code never sets `loadingScreenContainer` to a falsy value (or any other value than it's initial one actually) so this `if` is useless. The code that calls `removeChild` should `null` the variable for that test to actually be useful (and even then, it wouldn't prevent the `hideLoader` body to be executed twice under some circumstances).

src/Core/View.js
```diff
@@ -79,6 +82,44 @@ function View(crs, viewerDiv, options = {}) {
...
+                return;
+            }
+
+            loadingScreenContainer.style.opacity = 0;
+            loadingScreenContainer.style.pointerEvents = 'none';
+            loadingScreenContainer.style.transition = 'opacity 0.5s cubic-bezier(0.55, 0.085, 0.68, 0.53)';
+
+            setTimeout(() => {
```
tbroyer: Should this use a `transitionend` event instead? (except, well, if we leave it to the CSS, as suggested above, then there might not even be a transition)
```js
loadingScreenContainer.addEventListener('transitionend', e => {
  viewerDiv.removeChild(e.target);
});
// using e.target allows us to eager 'null' the variable to prevent re-running this code
loadingScreenContainer = null;
```

tbroyer (commented):
What about the size increase? Did you quantify it?

Fwiw, the more I look at this code, the more I think it shouldn't be in iTowns proper, but provided as a "companion" library and/or snippets to copy/paste into your own code.
AFAICT, the only missing thing is that `mainLoop` and its `command-queue-empty` event aren't part of the public API (`GlobeView` has an equivalent `initialized` event, but that wouldn't apply to other kinds of `View`)

YMMV

peppsac:
Yep, I looked at the size change, forgot to write this in my previous comment.
itowns.js size: 1104752 ->1107305 (so a 3kB increase for a 1.1MB js file).

I adopted your proposed changes (see 2 latest commits).

Regarding the customisation aspect: we could stick to the simple transition as currently implemented. 

Loading screen would be used like this:
- beginner usage: default logo + default animation (opacity -> 0 + element removal at the end)
- intermediate usage: custom logo + default animation
- advanced usage : custom loading screen (`command-queue-empty` is accessible: view.mainLoop.addEventListener())

tbroyer:
> advanced usage : custom loading screen (command-queue-empty is accessible: view.mainLoop.addEventListener())

Sure it's accessible, but it's not documented, right? (and a specific `initialized` event, generalized to all views, would be a better API; are there downsides to this?)

Code looks fine to me; I'm not going to block this PR on that basis or the size increase, so I'm deferring to others whether we want this in iTowns proper or not.

(just noticed: `transitionend` event bubbles, so maybe using `e.target` isn't such a good idea and your `const elt = …` would be better, adding checks on `e.propertyName === 'opacity'` and `e.target === elt`)

gchoqueux:
I am more for a companion library and remove the loading screen from itowns's core

peppsac:
I won't make a companion library for 20 lines of code.

I pushed a new commit: instead of adding 20 lines in View.js, it adds 3 lines in each example.
Let me know what you think. My opinion: I prefer the version in View.js because it costs us nothing (ok, 20 lines) and makes iTowns nicer by default.

autra:
> iTowns must allow user to easily implement this but shouldn't have an iTowns logo built-in, as a user may want to replace the logo by its own.

As soon as it is now possible to do so, having a default itowns logo when people are using iTowns doesn't seem to be a problem right? Apart from the fact it looks really better with it, I also think it's really good for self-promotion: most of our user will think it's nice and that will give us some credits. Those who don't can disable it quite easily.

So I prefer the previous version ( where a default version is provided in View.js).  I do agree that this should be documented in View.js' jsdoc. Maybe also we can have an example with another splashscreen, to show users how they can replace with their own? This way, everything will be covered. I would also disable it in dev mode (when starting with npm start), as it could be useful to see how the scene loads (and you can't do that if it's not in src/ btw).

gchoqueux:
> I won't make a companion library for 20 lines of code.

Why?

-----------------------

The itowns's core is to visualize SIG, all GUI feature must be outsourced

examples/3dtiles.html
```diff
@@ -70,13 +70,27 @@
...
         </style>
+        <link rel="stylesheet" type="text/css" href="css/loadingscreen.css">
         <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <script src="GUI/dat.gui/dat.gui.min.js"></script>
     </head>
     <body>
-        <div id="viewerDiv"></div>
+        <div id="viewerDiv">
+            <div id="itowns-loader">
```
gchoqueux: it must be possible to factorize this, no? 

```html
<div id="itowns-loader">
       <div class="text">
       <span class="c1">i</span>
               <span class="c2">T</span>
               <div id="spinner" class="c3"></div>
               <span class="c4">w</span>
               <span class="c5">n</span>
               <span class="c6">s</span></div>
</div>
```

autra:
> The itowns's core is to visualize SIG, all GUI feature must be outsourced

Come on, you can hardly call that a GUI feature.

I do agree that the core of itowns should be to visualize SIG data, but we should be pragmatic about the rest. You did add `AnimationPlayer.js` into itowns core last year right? And we also have all the controls, haven't we? If we wanted to be all purity, all these should be in examples/ (and animations should be done with FrameRequesters)... but we didn't put them there, because we do need a good *default* experience at some point (a bit more complete than THREE.js alone at least). Not so long ago, it was impossible to use the Globe without GlobeControls, and nobody made a fuss about it, and btw, it's still the default experience, despite being (very) opinionated and imposing much more constraints on the user than this... But hey, we need a default experience, so it makes sense to have it this way.

Do you guys *really* see a problem with a customizable loading screen comparing to this?

GuiTools is external, because anybody can craft a HTML ui that changes some properties and everyone likes different things concerning these kind of controls. It's not the case with iTowns' control, and I do think it is quite the same with the loading screen. Beside, the code is simple and won't bother us.

If @peppsac reverts to the version with View.js, it's gonna be a r+ for me.

I do agree we should be able to disable and customize it of course.

peppsac:
Can we make a decision here?
Should I close this PR? Or merge as-is? Or revert to the first implementation (`the version with View.js` as @autra called it)?

zarov:
I'm for the version of @tbroyer

gchoqueux:
> a specific initialized event, generalized to all views, would be a better API;

@tbroyer's suggestion is a good approach to the problem.

and this next part should be factorize :

```html
<div id="itowns-loader">
       <div class="text">
       <span class="c1">i</span>
               <span class="c2">T</span>
               <div id="spinner" class="c3"></div>
               <span class="c4">w</span>
               <span class="c5">n</span>
               <span class="c6">s</span></div>
</div>
```

elias75015:
IMHO (from a user point of view), we shouldn't impose a default loading screen logo, it only should be really easy to add it. 
Anyway, the most important is to correctly document this feature.

peppsac:
> IMHO (from a user point of view), we shouldn't impose a default loading screen logo, it only should be really easy to add it.

Good. That's exactly what this PR is doing: see https://github.com/iTowns/itowns/pull/636#issuecomment-363112851.

>  Anyway, the most important is to correctly document this feature.

It is documented already (I'll update the doc when we settle on the final version).


@zarov ("I'm for the version of @tbroyer") and @gchoqueux ("@tbroyer's suggestion is a good approach to the problem") :

Do you realize that the previous solution was doing more or less what was proposed by @tbroyer, except for the `inject` thing - which is unlikely to be useful anyway.
Just give this PR a try (and revert the latest commit), and try to implement your own loading screen and write your feedback here.

gchoqueux:
I'm talking about a specific initialized event, generalized to all views. 
It would be an event looking like `GLOBE_INITIALIZED`, but generalized to all viewers.
This would help the users and will be perfectly illustrated by an example of splashScreen
Note: `command-queue-empty` but it's not documented

zarov:
Ok for me then, you need to do two things though: 
- disable the loading screen of the `miniView` in `globe.js`
- set correctly the `o` of the logo
screenshot from 2018-02-19 16-01-48

I'm not a big fan of the animation, but we can improve it later.

peppsac:
Updated:
- `globe.html` example is now fixed.
- added a `VIEW_EVENTS.INITIALIZED` event. It's fired on the next rendering done after adding (at least) one layer and there's no more command to execute

WDYT?

(@zarov if you have an better animation/logo, feel free to propose it here, I don't mind changing it)

src/Core/View.js
```diff
@@ -79,6 +87,44 @@ function View(crs, viewerDiv, options = {}) {
     if (__DEBUG__) {
         this.isDebugMode = true;
     }
+
+    if (options.loadingScreen !== false) {
```
gchoqueux: loadingScreen has to be outsourced

gchoqueux (changes requested):

src/Core/View.js
peppsac: Can you explain what you mean please?

gchoqueux: the loading screen should gone in the examples

peppsac: I don't think so.
A default loading screen doesn't cost us much and provide a nicer default experience (see https://github.com/iTowns/itowns/pull/636#issuecomment-363471759 for a more elaborate answer).

gchoqueux: Or move it to a GUI library.
it's better to separate the html interface and the core

peppsac: > Or move it to a GUI library.

No. Because the point is to have something built-in so it works by default.

>  it's better to separate the html interface and the core

This is a rule of thumb, not an absolute rule that works in all case. 
In this specific case it doesn't make sense because again, I believe that a nicer experience by default is important and the cost (20 lines of JS) is negligible.

peppsac:
Last update: changed the default loading screen to avoid the "O" position issue.

nosy-b:
I was gonna go directly for a reject reading the title _feat(core): loading screen_ but it's more complicated, as show 38 thread interventions :)

(+) The loading animation is nice to have (Our friends at Cesium got a default one too for example)
(+) The addition of a view initialization event is a good one.
(-) Very weird to have its implementation in Core/View.js
(-) Painful/not very practical when developing

The externalisation from core is a necessity for me too so if you revert your reversion I vote for it.

peppsac:
>  (-) Very weird to have its implementation in Core/View.js

If it makes you feel better I can add a new file `LoadingScreen.js` and move the implementation there.

>  (-) Painful/not very practical when developing

As we have a __DEBUG__ variable, the loading screen could be disabled by default when __DEBUG__ is true.

src/Core/View.js
```diff
@@ -317,8 +363,26 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+    if (this._readyCallback) {
+        this.removeFrameRequester(MAIN_LOOP_EVENTS.AFTER_RENDER, this._readyCallback);
+    }
+
+    return layer.whenReady.then((layer) => {
+        this.notifyChange(false);
+
+        if (!this._readyCallback) {
+            this._readyCallback = () => {
+                if (this.mainLoop.scheduler.commandsWaitingExecutionCount() == 0) {
```
gchoqueux: it may not be enough. Elements may be waiting to be update in `viewer._changeSources`
it might be necessary to add in test `viewer._changeSources.size == 0` to make sure the queue isn't going to fill up again

gchoqueux (changes requested):

src/Core/View.js
peppsac: I fixed this but I used the `RENDERING_SCHEDULED` state of the mainloop instead.

peppsac:
@nosy-b & @gchoqueux PR updated following your comments.

peppsac:
> @nosy-b & @gchoqueux PR updated following your comments.

ping?

peppsac:
PR updated to remove the loading screen mechanism from core.

The only core modification left is the addition of a `'layers-initialized'` event for the view.
The loading screen is implemented for the examples only, in `examples/js/loading_screen.js`.

Please update your reviews. Thanks!

examples/js/loading_screen.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    // loading screen
+    // eslint-disable-next-line no-multi-str
+    viewerDiv.insertAdjacentHTML('beforeend', '\
+        <div id="itowns-loader">\
+        <div class="text">\
+        <span class="c1">i</span><span class="c2">T</span><span class="c3">o</span><span class="c4">w</span><span class="c5">n</span><span class="c6">s</span>\
+        </div></div>');
+    loadingScreenContainer = document.getElementById('itowns-loader');
```
tbroyer: nit: you could also:
```js
loadingScreenContainer = document.createElement('div');
loadingScreenContainer.innerHTML = '\
    <div class="text">\
    <span class="c1">i</span><span class="c2">T</span><span class="c3">o</span><span class="c4">w</span><span class="c5">n</span><span class="c6">s</span>\
    </div>';
viewerDiv.appendChild(loadingScreenContainer);
```

examples/pointcloud.html
```diff
@@ -160,8 +161,9 @@ <h3>Available pointclouds:</h3>
                     }
                 }
             </script>
-            <script src="GUI/dat.gui/dat.gui.min.js"></script>
+            <script src="js/GUI/dat.gui/dat.gui.min.js"></script>
             <script src="../dist/itowns.js"></script>
+        <script src="./js/loading_screen.js"></script>
```
tbroyer: nit: indent

examples/globe.html
```diff
@@ -64,18 +64,20 @@
...
             <span class="divScaleWidget"> Scale </span>
+            <div id="miniDiv">
+            </div>
         </div>
-        <div id="miniDiv">
-        </div>
-        <script src="GUI/GuiTools.js"></script>
+        <script src="js/GUI/GuiTools.js"></script>
         <script src="../dist/itowns.js"></script>
+        <script src="./js/loading_screen.js"></script>
```
tbroyer: nit: maybe `js/loading_screen.js` (without leading `./`), for consistency with dat.gui and the CSS.

webpack-babel.config.js
```diff
@@ -11,6 +11,8 @@ module.exports = {
         rules: [
             commonConfig.glslLoader,
             commonConfig.jsonLoader,
+            commonConfig.cssLoader,
```
tbroyer: No longer needed, right?

tbroyer (commented):
:shipit: Just nits (feel free to ignore), and I believe the Babel configuration is no longer needed.

peppsac:
Thanks for the feedback @tbroyer - I fix all the issues you mentionned.

tbroyer (approved):

src/Core/View.js
```diff
@@ -79,6 +92,10 @@ function View(crs, viewerDiv, options = {}) {
 
     if (__DEBUG__) {
         this.isDebugMode = true;
+        // Disable loading screen by default in dev build
+        if (options.loadingScreen === undefined) {
```
gchoqueux: Disabling `options.loadingScreen` should not be the responsibility of the viewer

src/Core/View.js
```diff
@@ -324,8 +341,29 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+    }
+
+    return layer.whenReady.then((layer) => {
+        this.notifyChange(false);
+
+        if (!this._allLayersAreReadyCallback) {
+            this._allLayersAreReadyCallback = () => {
+                if (this.mainLoop.scheduler.commandsWaitingExecutionCount() == 0 &&
+                    this.mainLoop.renderingState == RENDERING_PAUSED) {
+                    this.dispatchEvent({ type: VIEW_EVENTS.LAYERS_INITIALIZED });
```
gchoqueux, line 354: in view of the dispatch conditions for event `LAYERS_INITIALIZED` also represents the loading of the data of the layers? and not only that the layer is ready

src/Core/View.js
```diff
@@ -439,7 +477,12 @@ View.prototype.addFrameRequester = function addFrameRequester(when, frameRequest
  * @param {FrameRequester} frameRequester
  */
 View.prototype.removeFrameRequester = function removeFrameRequester(when, frameRequester) {
-    this._frameRequesters[when].splice(this._frameRequesters[when].indexOf(frameRequester), 1);
+    const index = this._frameRequesters[when].indexOf(frameRequester);
```
gchoqueux, line 477: it's a bug fix or a enhancement? Is it necessary for PR?
just to understand

src/Core/View.js
peppsac: this code should have been removed (options.loadingScreen is unused)

src/Core/View.js
peppsac: Yes, LAYER_INITIALIZED mean: all currently added layers have loaded enough data to be displayable.

src/Core/View.js
peppsac: It's a bug fix, that'll go in its own commit as soon as the rest of the PR is ok.

src/Core/View.js
gchoqueux: could you complete the documentation on this event?

peppsac: Sure but be more specific. What should I add/remove/edit from [the existing doc](https://github.com/iTowns/itowns/pull/636/files#diff-d297dcf7b9d0e696ece4b1f6e1d7c69dR12)?

gchoqueux: the documentation doesn't explain the ready status and can be confused with `layer.whenReady`. This event indicates status of viewer, mainloop and the layers

peppsac: Done

peppsac:
Rebased and squashed commits.

gchoqueux (approved):
Warning, before the merge :
in `index.html`, the paths aren't updated (`GuiTools.js`, `dat.js`)

PR #637, opened 2018-02-02
feat(core): add event to coordinates helpers

peppsac:
This commits introduces 4 functions to deal with mouse coordinates.

They provide:
  - event (mouse or touch) to window coordinates (in pixels)
  - event (mouse or touch) to normalized coords (eg: as used by THREE.Raycaster)
  - window <-> normalized conversions

@zarov this PR replaces some code you introduced in https://github.com/iTowns/itowns/commit/f3d77724200e651077ff0c9857d5e6b4e3d284f6#diff-df61a67eee71572718ef066564a3a0afR907 Could you test and confirm that this PR doesn't change the behavior?

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -153,7 +153,7 @@
             }
             function picking(event) {
                 // Pick an object with batch id
-                var mouse = new itowns.THREE.Vector2((event.clientX / window.innerWidth) * 2 - 1, -(event.clientY / window.innerHeight) * 2 + 1);
+                var mouse = globe.eventToNormalizedCoords(event);
                 var raycaster = new itowns.THREE.Raycaster();
                 var htmlInfo = document.getElementById('info');
                 htmlInfo.innerHTML = ' ';
@@ -161,6 +161,7 @@
                 raycaster.setFromCamera(mouse, globe.camera.camera3D);
                 // calculate objects intersecting the picking ray
                 var intersects = raycaster.intersectObjects($3dTilesLayerRequestVolume.object3d.children, true);
+
                 for (var i = 0; i < intersects.length; i++) {
                     var interAttributes = intersects[i].object.geometry.attributes;
                     if (interAttributes) {
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -458,4 +458,55 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
     }
 };
 
+const _eventCoords = new Vector2();
+/**
+ * Extract view coordinates from a mouse-event / touch-event
+ * @param {event} event - event can be a MouseEvent or a TouchEvent
+ * @param {number} touchIdx - finger index when using a TouchEvent (default: 0)
+ * @return {THREE.Vector2} - view coordinates (in pixels, 0-0 = top-left of the View)
+ */
+View.prototype.eventToViewCoords = function eventToViewCoords(event, touchIdx = 0) {
+    if (event.touches === undefined || !event.touches.length) {
+        return _eventCoords.set(event.offsetX, event.offsetY);
+    } else {
+        // originalTarget should always be viewerDiv
+        const br = event.originalTarget.getClientBoundingRect();
+        return _eventCoords.set(
+            event.touches[touchIdx].clientX - br.x,
+            event.touches[touchIdx].clientY - br.y);
+    }
+};
+
+/**
+ * Extract normalized coordinates (NDC) from a mouse-event / touch-event
+ * @param {event} event - event can be a MouseEvent or a TouchEvent
+ * @param {number} touchIdx - finger index when using a TouchEvent (default: 0)
+ * @return {THREE.Vector2} - NDC coordinates (x and y are [-1, 1])
+ */
+View.prototype.eventToNormalizedCoords = function eventToNormalizedCoords(event, touchIdx = 0) {
+    return this.viewToNormalizedCoords(this.eventToViewCoords(event, touchIdx));
+};
+
+/**
+ * Convert view coordinates to normalized coordinates (NDC)
+ * @param {Vector2} viewCoords (in pixels, 0-0 = top-left of the View)
+ * @return {THREE.Vector2} - NDC coordinates (x and y are [-1, 1])
+ */
+View.prototype.viewToNormalizedCoords = function viewToNormalizedCoords(viewCoords) {
+    _eventCoords.x = 2 * (viewCoords.x / this.camera.width) - 1;
+    _eventCoords.y = -2 * (viewCoords.y / this.camera.height) + 1;
+    return _eventCoords;
+};
+
+/**
+ * Convert NDC coordinates to view coordinates
+ * @param {Vector2} ndcCoords
+ * @return {THREE.Vector2} - view coordinates (in pixels, 0-0 = top-left of the View)
+ */
+View.prototype.normalizedToViewCoords = function normalizedToViewCoords(ndcCoords) {
+    _eventCoords.x = (ndcCoords.x + 1) * 0.5 * this.camera.width;
+    _eventCoords.y = (ndcCoords.y - 1) * -0.5 * this.camera.height;
+    return _eventCoords;
+};
+
 export default View;
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -907,11 +907,9 @@ function GlobeControls(view, target, radius, options = {}) {
             if (this.enabled === false) return;
 
             event.preventDefault();
-
-            const staticPos = window.getComputedStyle(event.target.parentElement).position !== 'static';
-            const bounds = staticPos ? event.target.getBoundingClientRect() : { left: 0, top: 0 };
-            const x = event.clientX - event.target.offsetLeft - bounds.left;
-            const y = event.clientY - event.target.offsetTop - bounds.top;
+            const coords = this._view.eventToViewCoords(event);
+            const x = coords.x;
+            const y = coords.y;
 
             if (state === this.states.ORBIT || state === this.states.PANORAMIC) {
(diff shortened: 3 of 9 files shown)


zarov:
Unfortunately it breaks my correction. In `eventToWindowCoords`, I think you may need to add the first four lines you linked from the PR.

FYI the test case I have is `globe_vector.html` with  `position: relative; height: 80%;` for the `viewerDiv` and a small div before `viewerDiv`, something like `< div style="height: 100px"></div>`, to have something like this:
screenshot from 2018-02-05 10-12-56
You can see that the tooltip is in the wrong place. Any `position` other than `static` will result in this result without the correction.

peppsac:
The tooltips were positionned using pageX / pageY.
I pushed a fix to use `viewer.eventToWindowCoords` and they seem at the correct position now.

zarov:
Seems ok to me with this fix, but you need @elias75015 or @pjjmunier approval too !

src/Core/View.js
```diff
@@ -458,4 +458,49 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
     }
 };
 
+const _eventCoords = new Vector2();
+/**
+ * Extract window coordinates from a mouse-event
+ * @param {mouseevent} event
+ * @return {THREE.Vector2} - window coordinates (in pixels, 0-0 = top-left)
+ */
+View.prototype.eventToWindowCoords = function eventToWindowCoords(event) {
```
gchoqueux: may be change to `mouseEventToViewerCoords`, replace `event` by `mouseEvent` and `windows` by `viewer`

peppsac: It could be a touchevent as well. So I think `event` is fine.

Window coordinates could be named View coordinates but Window Coordinates is the usual name... I can clarify the documentation and say they are relative to the View (e.g: `0-0 = top-left corner of the View`).

gchoqueux: ok, then function should indicate, in his parameters, which finger is used?
Documentation can be clarify

peppsac: I updated the documentation.

pjjmunier:
Ok, we adapted our API according to those changes and it works well.

peppsac:
Ok, then I'm rebasing this PR.

gchoqueux:
What do you think of this approach?
```js
view.on ('click', function(e) {
    // coordinates in viewer
    e.viewPt;
    // projection Coordinates in viewer crs
    e.geoCoord;
    // normalized device coordinates
    e.normalizedPt;
});
```

peppsac:
> What do you think of this approach?

There's 2 aspects in your proposal:
- `view.on('click', ...)`: we already use the `EventDispatcher` from threejs, so we expose `addEventListener` method elsewhere: I think we should stick to addEventListener. But I don't have any strong opinion on this point.
- e.viewPt, e.geoCoord, e.normalizedPt: most of the time  (maybe everytime in itowns codebase) we only need on the coordinates so I prefer having 1 method per coordinate types. But yes, we could add methods to cover event -> GeographicCoordinates methods and conversion between the 3 types.

peppsac:
Can we merge this PR yet?

src/Core/View.js
```diff
@@ -458,4 +458,51 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
...
+ * Extract window coordinates from a mouse-event / touch-event
+ * @param {event} event - event can be a MouseEvent or a TouchEvent
+ * @param {number} touchIdx - finger index when using a TouchEvent (default: 0)
+ * @return {THREE.Vector2} - window coordinates (in pixels, 0-0 = top-left of the View)
+ */
+View.prototype.eventToWindowCoords = function eventToWindowCoords(event, touchIdx = 0) {
+    if (event.touches === undefined || !event.touches.length) {
+        return _eventCoords.set(event.offsetX, event.offsetY);
+    } else {
+        return _eventCoords.set(event.touches[touchIdx].pageX, event.touches[touchIdx].pageY);
```
autra: If I read correctly the doc on mdn, pageX and pageY are not equivalent to offsetX and Y in the case of touch events...

autra: pageX/Y are relative to the whole document (they are `window` coordinates in the web sense). I guess you need to take into account where viewerDiv is placed.

src/Core/View.js
```diff
@@ -458,4 +458,51 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
     }
 };
 
+const _eventCoords = new Vector2();
+/**
+ * Extract window coordinates from a mouse-event / touch-event
+ * @param {event} event - event can be a MouseEvent or a TouchEvent
+ * @param {number} touchIdx - finger index when using a TouchEvent (default: 0)
+ * @return {THREE.Vector2} - window coordinates (in pixels, 0-0 = top-left of the View)
```
autra: > Window Coordinates is the usual name...

in which context? Because for the web devs, this will be very confusing... I'd prefer `viewport` (a bit better, though there *is* a viewport size in a webpage) or `canvasCoords`? WDYT?

autra (changes requested):
I'm just a little doubtful for touch devices, see my comments ( unfortunately I don't have any way to test that currently)

src/Core/View.js
peppsac: I renamed `WindowCoords` to `ViewCoords`.

src/Core/View.js
peppsac: I reused the existing code from `FirstPersonControls`.

peppsac: I replaced the pageXY by clientXY - it's closer to offsetXY.

src/Core/View.js
```diff
@@ -458,4 +458,51 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
...
+ * Extract view coordinates from a mouse-event / touch-event
+ * @param {event} event - event can be a MouseEvent or a TouchEvent
+ * @param {number} touchIdx - finger index when using a TouchEvent (default: 0)
+ * @return {THREE.Vector2} - view coordinates (in pixels, 0-0 = top-left of the View)
+ */
+View.prototype.eventToViewCoords = function eventToViewCoords(event, touchIdx = 0) {
+    if (event.touches === undefined || !event.touches.length) {
+        return _eventCoords.set(event.offsetX, event.offsetY);
+    } else {
+        return _eventCoords.set(event.touches[touchIdx].clientX, event.touches[touchIdx].clientY);
```
autra: If I'm not mistaken that wouldn't work if the viewerDiv have an offset comparing to the viewport (aka in situation where the page is scrollable for example). We have viewport coordinates here, and we want client coordinates (relative to viewerDiv boundingRect).

The only possible solution I see is:
```js
const br = event.originalTarget.getClientBoundingRect(); // originalTarget should always be viewerDiv
return _eventCoords.set(event.touches[touchIdx].clientX - br.x, /* same for y */); 
```
I'm quite sure it does the right thing with margins and paddings (taking into account the first one, but not the second), but I'm not sure about borders. I still can't test this morning, but I'll have a touch device this afternoon.

gchoqueux:
I'll be more for a specific event, dedicated to the viewer it would be an extent of mouseEvent.

the methods `eventToNormalizedCoords, viewToNormalizedCoords, normalizedToViewCoords`  added in this PR will be static methods and will be used to get in the getters properties (`e.viewPt, e.geoCoord, e.normalizedPt`) of the event. 

the advantage is to make a good part of the calls of `eventToNormalizedCoords, viewToNormalizedCoords, normalizedToViewCoords` disappear in code.

next time, it could be added to the layer event properties as `featuresOnMouse`.
```js
layer.on ('click', function(e) {
    // property with getter which will compute all the features under the mouse
    e.featuresOnMouse
});
```

peppsac:
> the advantage is to make a good part of the calls of eventToNormalizedCoords, viewToNormalizedCoords, normalizedToViewCoords disappear in code.

Before making them disappear by providing a higher-level API I'd suggest we first merge this low-level API.

src/Core/View.js
peppsac: Or I can reuse the code from https://github.com/iTowns/itowns/commit/f3d77724200e651077ff0c9857d5e6b4e3d284f6#diff-df61a67eee71572718ef066564a3a0afR907 for touch events?

autra: Unless I miss something, I think my version is clearer and more correct? I don't understand why we use `offsetLeft` there, which is the offset to the parent element. We don't care about the parent element right? Beside I don't understand this 'static' test (and the name staticPos is actually reversed from its meaning).

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -907,11 +907,9 @@ function GlobeControls(view, target, radius, options = {}) {
             if (this.enabled === false) return;
 
             event.preventDefault();
-
-            const staticPos = window.getComputedStyle(event.target.parentElement).position !== 'static';
-            const bounds = staticPos ? event.target.getBoundingClientRect() : { left: 0, top: 0 };
-            const x = event.clientX - event.target.offsetLeft - bounds.left;
-            const y = event.clientY - event.target.offsetTop - bounds.top;
+            const coords = this._view.eventToViewCoords(event);
```
autra, line 910: this actually fixes a bug that is on master: if the page is scrollable, the previous version didn't pick the correct position when the page was scrolled.

autra (approved):
It makes the code simpler and fixes a bug on master (from #611 I guess), LGTM!

examples/3dtiles.html
```diff
@@ -153,14 +153,15 @@
             }
             function picking(event) {
                 // Pick an object with batch id
-                var mouse = new itowns.THREE.Vector2((event.clientX / window.innerWidth) * 2 - 1, -(event.clientY / window.innerHeight) * 2 + 1);
+                var mouse = globe.eventToNormalizedCoords(event);
```
gchoqueux, line 156: `globe` must be replaced by `view` to fix example

PR #638, opened 2018-02-05
fix(core): wrong building 3dTiles's OBB

gchoqueux:
## Description
fix : 
* wrong ellipsoid's normal
* careless mistake


peppsac (approved):
LGTM, thanks.

PR #639, opened 2018-02-05
feat(core): compute TMS coordinates from north to south

zarov:
## Description
As it can be read in the link at the end, TMS coordinates differ from other "TMS" coordinates from Google Maps, OSM, Bing etc... The first one goes from south to north, while the latter north to south. In this commit, the parameter `northToSouth` in the method `computeTMSCoordinates` can be set to true to obtain the same result as Google Maps and others.

Source: https://alastaira.wordpress.com/2011/07/06/converting-tms-tile-coordinates-to-googlebingosm-tile-coordinates/

## Motivation and Context
No result to see for now, but it is being used in Mapbox Vector Tile integration in iTowns in another branch. While this branch can wait, I also want an input on the name `northToSouth`, as none of the cited provider above gave a name to this convention. WDYT ?

### DIFF
--- a/src/Core/Scheduler/Providers/OGCWebServiceHelper.js
+++ b/src/Core/Scheduler/Providers/OGCWebServiceHelper.js
@@ -114,7 +114,12 @@ export default {
                 projection.getCoordWMTS_WGS84(tileCoord, tile.extent, tileMatrixSet);
         }
     },
-    computeTMSCoordinates(tile, extent) {
+    // The origin parameter is to be set to the correct value, bottom or top
+    // (default being bottom) if the computation of the coordinates needs to be
+    // inverted to match the same scheme as OSM, Google Maps or other system.
+    // See link below for more information
+    // https://alastaira.wordpress.com/2011/07/06/converting-tms-tile-coordinates-to-googlebingosm-tile-coordinates/
+    computeTMSCoordinates(tile, extent, origin = 'bottom') {
         if (tile.extent.crs() != extent.crs()) {
             throw new Error('Unsupported configuration. TMS is only supported when geometry has the same crs than TMS layer');
         }
@@ -129,7 +134,12 @@ export default {
 
         // Now that we have computed zoom, we can deduce x and y (or row / column)
         const x = (c.x() - extent.west()) / layerDimension.x;
-        const y = (extent.north() - c.y()) / layerDimension.y;
+        let y;
+        if (origin == 'top') {
+            y = (extent.north() - c.y()) / layerDimension.y;
+        } else {
+            y = (c.y() - extent.south()) / layerDimension.y;
+        }
 
         return [new Extent('TMS', zoom, Math.floor(y * tileCount), Math.floor(x * tileCount))];
     },
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -188,7 +188,7 @@ TileMesh.prototype.getCoordsForLayer = function getCoordsForLayer(layer) {
             throw new Error('unsupported projection wms for this viewer');
         }
     } else if (layer.protocol == 'tms') {
-        return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent);
+        return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent, layer.origin);
     } else {
         return [this.extent];
     }
--- a/examples/orthographic.js
+++ b/examples/orthographic.js
@@ -53,6 +53,7 @@ view.addLayer({
     networkOptions: { crossOrigin: 'anonymous' },
     extent: [extent.west(), extent.east(), extent.south(), extent.north()],
     projection: 'EPSG:3857',
+    origin: 'bottom',
     options: {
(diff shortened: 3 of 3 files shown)


peppsac:
I'd suggest `origin: top-left` or `origin: bottom-left`.

And if origin is the main difference  (only?) difference between TMS (bottom-left) and XYZ (top-left) protocols, maybe we could add the `xyz` protocol:
```js
this.addProtocolProvider('xyz', new TMS_Provider());
```
And in `TMS_Provider.prototype.preprocessDataLayer` assign the correct origin, depending on the chosen protocol.

(we could possibly rename this provider as well but I don't have any proposal in mind..)

WDYT?

peppsac:
If you're working on VectorTiles support it could be interesting to write a comment in this issue: https://github.com/iTowns/itowns/issues/182  to avoid duplication of effort. Thanks!

zarov:
Very good idea with `xyz`, it is way clearer than adding an option !

Yes I'm completing this issue, thanks.

autra:
> I'd suggest origin: top-left or origin: bottom-left.

FWIW the gdal TMS mini-driver has a `<YOrigin>top/bottow</YOrigin>` parameter. If we don't plan on supporting origin change on the x-axis, maybe `.yorigin` is enough?

peppsac (approved):
LGTM, thanks!

peppsac:
Before merging, can you confirm that orthographic is still working?

zarov:
It works, as long as we don't set the protocol to `xyz`, it isn't changing a thing ;)

peppsac (changes requested):
I believe the logic is reversed :)

See:
https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification
https://www.planet.com/docs/reference/tile-services/

zarov:
Well it seems like it 😄 Updated with 0e31411

src/Core/Scheduler/Providers/OGCWebServiceHelper.js
```diff
@@ -114,7 +114,11 @@ export default {
                 projection.getCoordWMTS_WGS84(tileCoord, tile.extent, tileMatrixSet);
         }
     },
-    computeTMSCoordinates(tile, extent) {
+    // The topLeft parameter is to be set at true if the computation of the
+    // coordinates needs to be inverted to match the same scheme as OSM, Google
+    // Maps or other system. See link below for more information
+    // https://alastaira.wordpress.com/2011/07/06/converting-tms-tile-coordinates-to-googlebingosm-tile-coordinates/
+    computeTMSCoordinates(tile, extent, topLeft = false) {
```
peppsac: I'd prefer the second argument to be `origin` (or `yOrigin`) with a default value : `top-left` than a boolean.

This way at the caller place it's easier to read the code:
```js
computeTMSCoordinates(tile, extent, layer.origin);
computeTMSCoordinates(tile, extent, 'top-left');
```
Are both clearer than:
```js
computeTMSCoordinates(tile, extent, false);
computeTMSCoordinates(tile, extent, true);
```

src/Core/TileMesh.js
```diff
@@ -187,8 +187,8 @@ TileMesh.prototype.getCoordsForLayer = function getCoordsForLayer(layer) {
         } else {
             throw new Error('unsupported projection wms for this viewer');
         }
-    } else if (layer.protocol == 'tms') {
-        return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent);
+    } else if (layer.protocol == 'tms' || layer.protocol == 'xyz') {
+        return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent, layer.protocol == 'xyz');
```
peppsac: Even if it's equivalent, I'd prefer to add a origin (or yOrigin) property to these layer.
You could assign a default value in  `TMS_Provider.prototype.preprocessDataLayer` based on the protocol - but this would allow the user to explicity define it.

peppsac (approved):
2 style comments; feel free to ignore them if you disagree.

LGTM otherwise.

src/Core/TileMesh.js
zarov: So you would have here, if I understand correctly:
```js
} else if (layer.protocol == 'tms') {
    return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent, layer.origin);
```
Seems good to me as we drop two tests doing it like this.

src/Core/Scheduler/Providers/OGCWebServiceHelper.js
zarov: I don't know, wouldn't setting a string like `top-left` induces that we could have value like `bottom-right` or something else ? But I'm maybe being too picky here ;)

autra: `yOrigin: top/bottom`? If we never have a {top,bottom}-right, then it's the clearest option imo.

peppsac:
> Before merging, can you confirm that orthographic is still working?

Well orthographic.html is broken. Can you fix it please?

PR #643, opened 2018-02-06
fix(core): handle near pos. in box3 visibility functions

peppsac:
Projection matrix produces strange results when projecting a point behind
the camera near plane.
To avoid errors this commits clamp z coordinate to camera.near in camera space,
before projecting.

The issue can be tested in planar.html, by positionning the camera with:
```js
    view.camera.camera3D.position.copy({x: 1841430.8707152607, y: 5173717.817231011, z: 229.93251060258268});
    view.camera.camera3D.rotation.set(-0.7231494579332582, -1.1919412510926017, -2.3305711080202998, "XYZ")
    view.notifyChange(true)
```
In this case box3SizeOnScreen used to return an y value between 0.91 and 1.3 while
the projected tile covers all the screen.

### DIFF
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -70,14 +70,55 @@ const tmp = {
     box3: new THREE.Box3(),
 };
 
-Camera.prototype.isBox3Visible = function isBox3Visible(box3, matrixWorld) {
+const points = [
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+];
+
+function projectBox3PointsInCameraSpace(camera, box3, matrixWorld) {
+    // Projects points in camera space
+    // We don't project directly on screen to avoid artifacts when projecting
+    // points behind the near plane.
+    let m = camera.camera3D.matrixWorldInverse;
     if (matrixWorld) {
-        tmp.matrix.multiplyMatrices(this._viewMatrix, matrixWorld);
-        tmp.frustum.setFromMatrix(tmp.matrix);
-    } else {
-        tmp.frustum.setFromMatrix(this._viewMatrix);
+        m = tmp.matrix.multiplyMatrices(camera.camera3D.matrixWorldInverse, matrixWorld);
     }
-    return tmp.frustum.intersectsBox(box3);
+    points[0].set(box3.min.x, box3.min.y, box3.min.z).applyMatrix4(m);
+    points[1].set(box3.min.x, box3.min.y, box3.max.z).applyMatrix4(m);
+    points[2].set(box3.min.x, box3.max.y, box3.min.z).applyMatrix4(m);
+    points[3].set(box3.min.x, box3.max.y, box3.max.z).applyMatrix4(m);
+    points[4].set(box3.max.x, box3.min.y, box3.min.z).applyMatrix4(m);
+    points[5].set(box3.max.x, box3.min.y, box3.max.z).applyMatrix4(m);
+    points[6].set(box3.max.x, box3.max.y, box3.min.z).applyMatrix4(m);
+    points[7].set(box3.max.x, box3.max.y, box3.max.z).applyMatrix4(m);
+
+    // In camera space objects are along the -Z axis
+    // So if min.z is > -near, the object is invisible
+    let atLeastOneInFrontOfNearPlane = false;
+    for (let i = 0; i < 8; i++) {
+        if (points[i].z <= -camera.camera3D.near) {
+            atLeastOneInFrontOfNearPlane = true;
+        } else {
+            // Clamp to near plane
+            points[i].z = -camera.camera3D.near;
+        }
+    }
+
+    return atLeastOneInFrontOfNearPlane ? points : undefined;
+}
+
+const ndcBox3 = new THREE.Box3(
+    new THREE.Vector3(-1, -1, -1),
+    new THREE.Vector3(1, 1, 1));
+
+Camera.prototype.isBox3Visible = function isBox3Visible(box3, matrixWorld) {
+    return this.box3SizeOnScreen(box3, matrixWorld).intersectsBox(ndcBox3);
 };
 
 Camera.prototype.isSphereVisible = function isSphereVisible(sphere, matrixWorld) {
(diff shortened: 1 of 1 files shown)


src/Renderer/Camera.js
```diff
@@ -70,14 +70,55 @@ const tmp = {
...
+        } else {
+            // Clamp to near plane
+            points[i].z = -camera.camera3D.near;
+        }
+    }
+
+    return atLeastOneInFrontOfNearPlane ? points : undefined;
+}
+
+const ndcBox3 = new THREE.Box3(
```
autra, line 116: what does ndc stands for?

src/Renderer/Camera.js
```diff
@@ -70,14 +70,55 @@ const tmp = {
...
     }
-    return tmp.frustum.intersectsBox(box3);
+    points[0].set(box3.min.x, box3.min.y, box3.min.z).applyMatrix4(m);
+    points[1].set(box3.min.x, box3.min.y, box3.max.z).applyMatrix4(m);
+    points[2].set(box3.min.x, box3.max.y, box3.min.z).applyMatrix4(m);
+    points[3].set(box3.min.x, box3.max.y, box3.max.z).applyMatrix4(m);
+    points[4].set(box3.max.x, box3.min.y, box3.min.z).applyMatrix4(m);
+    points[5].set(box3.max.x, box3.min.y, box3.max.z).applyMatrix4(m);
+    points[6].set(box3.max.x, box3.max.y, box3.min.z).applyMatrix4(m);
+    points[7].set(box3.max.x, box3.max.y, box3.max.z).applyMatrix4(m);
```
autra, line 99: Not sure you need all this. Can't we work with box3 all along? By doing here:
```js
box3.applyMatrix4(m)
```
(This should still work at this point)

Computing `atLeastOneInFrontOfNearPlane` is then just testing box3.min.z and box3.max.z, then clamping to near plane is just a matter of changing `box3.{min,max}.z`.

Then you could return an empty box3 here directly, and test for `isEmpty` in `box3SizeOnScreen`, or maybe get rid of this function.

Did I miss something?

autra (changes requested):
Just thinking out loud: `box3SizeOnScreen` should probably be renamed `box3OnScreen` at some point (not requiring it here though), because that's an API change.

src/Renderer/Camera.js
peppsac: Normalized device coordinates (see https://www.ncl.ucar.edu/Document/Graphics/ndc.shtml for instance)

src/Renderer/Camera.js
peppsac: It's not the same thing. Because in one case you're applying matrices to min/max of the box, and in the other case you're applying matrices to individual points and then take min/max.

Here's an example patch doing something similar to what you suggested:
```patch
iff --git a/src/Renderer/Camera.js b/src/Renderer/Camera.js
index a3e5d8cf..25edc994 100644
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -70,18 +70,7 @@ const tmp = {
     box3: new THREE.Box3(),
 };
 
-const points = [
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-    new THREE.Vector3(),
-];
-
-function projectBox3PointsInCameraSpace(camera, box3, matrixWorld) {
+function projectBox3InCameraSpace(camera, box3, matrixWorld) {
     // Projects points in camera space
     // We don't project directly on screen to avoid artifacts when projecting
     // points behind the near plane.
@@ -89,28 +78,18 @@ function projectBox3PointsInCameraSpace(camera, box3, matrixWorld) {
     if (matrixWorld) {
         m = tmp.matrix.multiplyMatrices(camera.camera3D.matrixWorldInverse, matrixWorld);
     }
-    points[0].set(box3.min.x, box3.min.y, box3.min.z).applyMatrix4(m);
-    points[1].set(box3.min.x, box3.min.y, box3.max.z).applyMatrix4(m);
-    points[2].set(box3.min.x, box3.max.y, box3.min.z).applyMatrix4(m);
-    points[3].set(box3.min.x, box3.max.y, box3.max.z).applyMatrix4(m);
-    points[4].set(box3.max.x, box3.min.y, box3.min.z).applyMatrix4(m);
-    points[5].set(box3.max.x, box3.min.y, box3.max.z).applyMatrix4(m);
-    points[6].set(box3.max.x, box3.max.y, box3.min.z).applyMatrix4(m);
-    points[7].set(box3.max.x, box3.max.y, box3.max.z).applyMatrix4(m);
+    box3.applyMatrix4(m);
 
     // In camera space objects are along the -Z axis
     // So if min.z is > -near, the object is invisible
-    let atLeastOneInFrontOfNearPlane = false;
-    for (let i = 0; i < 8; i++) {
-        if (points[i].z <= -camera.camera3D.near) {
-            atLeastOneInFrontOfNearPlane = true;
-        } else {
-            // Clamp to near plane
-            points[i].z = -camera.camera3D.near;
-        }
+    if (box3.min.z > -camera.camera3D.near) {
+        box3.makeEmpty();
+        return false;
+    } else {
+        // Clamp to near plane
+        box3.max.z = Math.min(-camera.camera3D.near, box3.max.z);
+        return true;
     }
-
-    return atLeastOneInFrontOfNearPlane ? points : undefined;
 }
 
 const ndcBox3 = new THREE.Box3(
@@ -132,19 +111,16 @@ Camera.prototype.isSphereVisible = function isSphereVisible(sphere, matrixWorld)
 };
 
 Camera.prototype.box3SizeOnScreen = function box3SizeOnScreen(box3, matrixWorld) {
-    const pts = projectBox3PointsInCameraSpace(this, box3, matrixWorld);
+    tmp.box3.copy(box3);
 
-    // All points are in front of the near plane -> box3 is invisible
-    if (!pts) {
-        return tmp.box3.makeEmpty();
-    }
+    const visible = projectBox3InCameraSpace(this, tmp.box3, matrixWorld);
 
-    // Project points on screen
-    for (let i = 0; i < 8; i++) {
-        pts[i].applyMatrix4(this.camera3D.projectionMatrix);
+    if (!visible) {
+        return tmp.box3;
     }
 
-    return tmp.box3.setFromPoints(pts);
+    tmp.box3.applyMatrix4(this.camera3D.projectionMatrix);
+    return tmp.box3;
 };
```

And a test-case where you can see the difference:
- enable "Show tile outline"
- enable "Display charts" option
- move the camera:
```js
    view.camera.camera3D.position.copy({x: 1838889.1024189065, y: 5168612.526095718, z: 1999.999999999998});
    view.camera.camera3D.rotation.set(1.1252825684050354, -0.8163390815001759, -0.33484909519875167, "XYZ")
    view.notifyChange(true)
```
- and observe the level 1 tiles visibility count. It should be `3` (because the NW tile isn't visible) but when using the box3.applyMatrix4 version it's 4

autra: > Did I miss something?

I did! Thanks for the explanation.

autra (approved):

PR #640, opened 2018-02-06
test(core): add unit tests for updateLayeredMaterialNodeImagery

peppsac:
This code is full of branches and is error-prone.

This PR adds a couple of unit tests, with 1 being ignored (this is the bug fixed by #618)

@gchoqueux we need to merge this one before #618, and to add more tests (features not tested: `noTextureParentOutsideLimit`, texture inheritance, error handling, elevation update)

PR #644, opened 2018-02-07
perf(protocols): decrease the count of fetch image/xbil

gchoqueux:
Return after the initialization of imaging/elevation allows
to delay the request for data fetching. where appropriate,
allows to privilege the subdivision and fetch the necessary datas.

example perf: -30% in the count of fetching for the wmts for `examples/globe `

Proposed new process `updateLayeredMaterialNodeXXX`, in two separate processes:
   * _FIRST PASS_ : `initNodeXXXFromParent` and get out of the function
   * _SECOND PASS_ : Fetch best texture

  The two-step allows you to filter out unnecessary requests
  Indeed in the second pass, their state (not visible or not displayed) can block them to fetch

### DIFF
--- a/test/layeredmaterialnodeprocessing_unit_test.js
+++ b/test/layeredmaterialnodeprocessing_unit_test.js
@@ -15,6 +15,9 @@ describe('updateLayeredMaterialNodeImagery', function () {
 
     // Mock scheduler
     const context = {
+        view: {
+            notifyChange: () => true,
+        },
         scheduler: {
             commands: [],
             execute: (cmd) => {
@@ -75,8 +78,13 @@ describe('updateLayeredMaterialNodeImagery', function () {
         tile.material.indexOfColorLayer = () => 0;
         tile.material.isColorLayerDownscaled = () => true;
         tile.material.getColorLayerLevelById = () => 1;
-        updateLayeredMaterialNodeImagery(context, layer, tile);
 
+        // FIRST PASS: init Node From Parent and get out of the function
+        // without any network fetch
+        updateLayeredMaterialNodeImagery(context, layer, tile);
+        assert.equal(context.scheduler.commands.length, 0);
+        // SECOND PASS: Fetch best texture
+        updateLayeredMaterialNodeImagery(context, layer, tile);
         assert.equal(context.scheduler.commands.length, 1);
     });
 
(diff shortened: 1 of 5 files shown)


peppsac:
Can we first merge the unit test PR #640 before adding more code to this file?

Also please add comments to explain your intent / the reasonning behind the proposed changes. Thanks!

gchoqueux:
PR updated

peppsac:
I don't see any difference in fetch count in the `globe.html` example. Did you measure before/after fetch count?

Also I'm afraid that your approach won't prevent all necessary updates: you should also prevent fetches for a subdividing node.

gchoqueux:
> example perf: -30% in the count of fetching for the wmts for examples/globe

Yes, of course, I had tests 

**in globe.html (render size 1100x1100) after Globe is initialized, try :**
`globeView.controls.setCameraTargetGeoPositionAdvanced({ longitude:  6.886991921801097, latitude: 45.87929727669077, range: 3000})`

**Result the count of fetch on Ortho**
* master :  ~280
* this PR :  ~210

> Also I'm afraid that your approach won't prevent all necessary updates: you should also prevent fetches for a subdividing node

you could specify?

test/layeredmaterialnodeprocessing_unit_test.js
```diff
@@ -75,6 +78,8 @@ describe('updateLayeredMaterialNodeImagery', function () {
         tile.material.indexOfColorLayer = () => 0;
         tile.material.isColorLayerDownscaled = () => true;
         tile.material.getColorLayerLevelById = () => 1;
+
+        updateLayeredMaterialNodeImagery(context, layer, tile);
```
peppsac: If there needs to be a second call, you should add a test to validate the intermediate state.

I suppose that you want to check that the first didn't make any network fetch, since it's the feature added by your PR.

peppsac (commented):
Looks good.
Can you just add an assert in the tests between the 2 successive calls to updateLayeredMaterialNodeImagery please?

test/layeredmaterialnodeprocessing_unit_test.js
gchoqueux: done

peppsac (approved):
Thanks!

PR #646, opened 2018-02-07
package(npm): add 'test-unit-watch' 

gliegard:
Add feature npm run unit-test-watch

## Description
It runs unit test and keep watching files, and re run the test if a file has changed.

## Motivation and Context
It can be useful during test driven development.


zarov:
Or you can do `npm run test-unit -- --watch` ;)

tbroyer:
I agree with @zarov, better document `npm run test-unit -- --watch`.

gliegard:
ok, thanks for this solution.
I close the PR.

PR #648, opened 2018-02-08
fix(examples): bad orthographic origin

zarov:
Forgot to change the origin from `bottom` to `top` in the orthographic example, sorry about that.

See #639

### DIFF
--- a/examples/orthographic.js
+++ b/examples/orthographic.js
@@ -46,14 +46,13 @@ view.tileLayer.disableSkirt = true;
 // Add an TMS imagery layer
 view.addLayer({
     type: 'color',
-    protocol: 'tms',
+    protocol: 'xyz',
     id: 'OPENSM',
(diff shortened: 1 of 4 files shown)


peppsac (commented):
Can you add this in `TMS_Provider.prototype.preprocessDataLayer`  as well:
```js
layer.origin = layer.origin || (layer.protocol == 'xyz' ? 'top' : 'bottom');
```
And then remove the `origin` declaration in `orthographic.js`

zarov:
The protocol `xyz` has been removed in the other PR, do you want to add it back ?

peppsac:
Oh. I misread the code yesterday when I approved... Yes I would prefer to keep the 2 protocol names + the `origin` option. Sorry.

zarov:
OK, fixed

peppsac (approved):
Thanks!

Issue #645, opened 2018-02-07
Partial rendering on mobile browsers (no globe, no WMTS tiles, bad controls)


fnicollet:
Hello,

We are looking at different 3D rendering engines, as we want to integrate 3D in our GIS (2D) software.

iTowns works pretty well on Desktop (the WMTS tiles are rather slow but that might be an IGN problem, or maybe the wrong tiles are loaded first depending on the field of view, I am not sure) but we also tried iTowns on a number of mobile devices (android phones, android tablets, ipad mini)

On those platforms, iTowns doesn't work well (or at all). The globe is not displayed as the WMTS tiles seem to never appear. 
Also, the controls are mostly not working on mobile (you can kind of move, zoom in or out, but you can't rotate). Here is a screenshot of the GPX official example on mobile:
![screenshot_20180207-124410](https://user-images.githubusercontent.com/1240481/35915075-e0595462-0c05-11e8-8397-c0e3e6a8e47a.png)
The other examples don't work either, so it is quite easy to reproduce.

iTowns is based on WebGL, which has sparse support on mobile, but other rendering engines work, not sure if they use a fallback or feature detection but they do work.

Our GIS (2D) software is working on mobile and we would need the 3D to work as where on mobile.

It looks like iTowns was not made to be working on mobile. Is it something that is planned at some point or is it out of the focus of the project?

Thanks for your help,
Fabien

vpicavet:
Hi,

I would say that iTowns is not specifically developed currently to be used on mobile phones. The problem is primarily with testing. Depending on your device, you may have very different outcome.

They are probably multiple causes for that, from hardware limitations to browser versions and JS support, to real itowns bugs.

That said, we would like to have it work on mobile. Not a strong priority right now, but contributions and bugfixes are welcome.

New controls for mobile should probably be written to be adapted for mobile and tablets.

I guess people at IGN like @nosy-b, @gchoqueux or @gcebelieu  will have an opinion, as geoportail.fr is now using itowns for 3D.

fnicollet:
Thanks for the detailed answer, i am also interested to know if the
geoportail team could have something handy to fix that (seems that 3d is
just disabled on mobile)

Fabien

Le 7 févr. 2018 17:01, "Vincent Picavet" <notifications@github.com> a
écrit :

autra:
@fnicollet I know that pointcloud examples works on some android devices. To elaborate on what @vpicavet said, I'd say our goal is to support most - if not all - devices that three.js supports.

[...]

PR #649, opened 2018-02-08
fix(core): disable logdepthbuffer if EXT_frag_depth is missing

peppsac:
Since r88 threejs allows enabling logdepthbuffer if EXT_frag_depth is missing,
see https://github.com/mrdoob/three.js/pull/11995.

iTowns doesn't handle this case correctly: so for now we recreate the renderer
if we realize that EXT_frag_depth isn't supported.

Before this PR:

capture d ecran de 2018-02-08 17-21-13

After:
capture d ecran de 2018-02-08 17-21-47

Note: the rendering is broken, but I have another PR that fixes this by setting correct values to camera.near/far (this PR depends on: #635, #637 and a last one not yet submitted).

### DIFF
--- a/src/Renderer/c3DEngine.js
+++ b/src/Renderer/c3DEngine.js
@@ -89,6 +89,21 @@ function c3DEngine(rendererOrDiv, options = {}) {
         throw new Error('WebGL unsupported');
     }
 
+    if (!renderer && options.logarithmicDepthBuffer) {
+        // We don't support logarithmicDepthBuffer when EXT_frag_depth is missing.
+        // So recreated a renderer if needed.
+        if (!this.renderer.extensions.get('EXT_frag_depth')) {
+            const _canvas = this.renderer.domElement;
+            this.renderer.dispose();
+            this.renderer = new THREE.WebGLRenderer({
+                canvas: _canvas,
+                antialias: options.antialias,
+                alpha: options.alpha,
+                logarithmicDepthBuffer: false,
+            });
+        }
+    }
+
     // Let's allow our canvas to take focus
     // The condition below looks weird, but it's correct: querying tabIndex
     // returns -1 if not set, but we still need to explicitly set it to force


nosy-b:
Good for smartphone :+1: #645

nosy-b:
I don't have the EXT_frag_depth bug anymore on my phone but now it crashes (WebGL had a problem)

peppsac:
> I don't have the EXT_frag_depth bug anymore on my phone but now it crashes (WebGL had a problem)

Don't you have any logs?

peppsac:
And on which device / example did you test? Does it crash on all examples?
Can you try to simplify TileFS.glsl to see if you can isolate the issue?

peppsac:
I limited the number of samplers to 8 on Android, let me know if it helps - except if your mobile phone is not an android :)
Edit: well not really, I only limited it if the vendor string begins with `Google` - which is true on the emulator but probably not on real phones. Can you report the value of UNMASKED_VENDOR_WEBGL on your phone and try to use it in `Capabilities.js` to limit the number of samplers?

autra (approved):
This fixes Midori at least (which does not have logdepthbuffer), so r+ for me. @peppsac for the last commit, it's up to you to keep it or not. If you do, please edit the commit messages. Thanks!

peppsac:
The last commit is a test for @nosy-b. If this commit doesn't fix the crash on his phone => we need to look for the real cause.

If it fixes the crash, either:
- we realize the issue only happen on one mobile GPU config (say Qualcomm). Then we can adapt the last commit to limit max_texture_unit to 8 on such GPUs
- or it happen ons various combination of GPU/iOS/Android/other param => in which case there's nothing to do about it - except implementing a more elaborate fix to limit our texture sampler usage (e.g: #309)

nosy-b:
I tried to limit the number of samplers too. I make a coffee and then I'll test your last commit  with remote debugging :)
(My phone is an android 6., and I remember itowns 2 was working like a year ago)

autra:
@nosy-b So you're not saying that this PR particularly breaks your phone right? Can you test on master? If it is also broken there, then, let's merge the first commit: it's already making things better for some devices / browsers, and open an issue / PR for your specific problem containing the second commit.

nosy-b:
@autra this PR makes the app crash. It was not the case before but the globe was all white.

Here is my log from the last commit on my android 6:

`THREE.WebGLRenderer 89
three.module.js:20895 THREE.WebGLRenderer: WEBGL_depth_texture extension not supported.
three.module.js:20895 THREE.WebGLRenderer: OES_texture_float_linear extension not supported.
three.module.js:20895 THREE.WebGLRenderer: EXT_frag_depth extension not supported.
three.module.js:21214 THREE.WebGLRenderer 89
three.module.js:20895 THREE.WebGLRenderer: WEBGL_depth_texture extension not supported.
three.module.js:20895 THREE.WebGLRenderer: OES_texture_float_linear extension not supported.
three.module.js:18051 THREE.WebGLProgram: gl.getProgramInfoLog() --From Vertex Shader:

three.module.js:18047 THREE.WebGLProgram: shader error:  0 gl.VALIDATE_STATUS false gl.getProgramInfoLog  
THREE.WebGLRenderer: Context Lost.`

I'll check the number of samples

peppsac:
@nosy-b see my comment: https://github.com/iTowns/itowns/pull/649#issuecomment-364266587

Use wegglreport.com or similar to get your vendor_id and then adapt the last commit to limit the number of samplers for your vendor (or hardcode a limit to 8 or less)

peppsac:
And IMHO, this PR makes the app crash because the crashing bug was hidden by the depthlog bug...

nosy-b:
@peppsac Yes I think so too.
(btw my vendor string is qualcomm)

peppsac:
> @peppsac Yes I think so too.

Ok, so we should merge this PR, without the last comit.

>  (btw my vendor string is qualcomm)

Did you test the workaround with the correct vendor string then?

autra:
Merging. We'll open other issues for all the other many problems on mobile, is that right with you?

Also pinging @fnicollet : can you test again now that this is merged? Did you see some improvements?

fnicollet:
Great work, i'll try it out.
I'm quite lazy, is there a way to test the master version without building it from source (rawgit or something)?
if not are there some build instructions?

Fabien

autra:
@fnicollet yes there is ! Please go to http://www.itowns-project.org/itowns/examples/, these examples are generated with current master once the master build on https://travis-ci.org/iTowns/itowns is done.

nosy-b:
@peppsac  ok good for the merge.
(I'll have to check later cause even forcing to 8 tex samplers and handling log depth it crashes.)

fnicollet:
@autra Amazing, exactly what I needed!

So, on the devices that I have here:
iPad mini: Chrome/Safari: no changes
Samsung Galaxy Tab A: Chrome: It works!
Sony Xperia Z3 mini: Chrome: crash (WebGL had a problem)

So, there is probably some tweaking to do here. I'll try to log some github issues, specific to each device problem, with the webgl diagnostic later

autra:
@fnicollet that'll be great! I'm taking a mental note that you have a bunch of different mobile devices ;-) Thanks for your help

fnicollet:
Yes, that's our device bench :) we also have a samsung smartphone, some
iPhone and some Windows touch devices, which could in handy too for testing

We are benchmarking 3D viewers for our products (iTowns, Cesium and Melown
mostly) and device compatibility would be a big plus.
I am going to create a POC next week on iTowns, see how it performs and how
we can integrate it with our software. We have some doubt on the
performance of iTowns regarding how it works. From what I understand, it is
based on WMTS tiling to get the terrain heights, while other engines use
pre-generated (tiled) terrains in binary format, which seem to be more
effective.

Anyway, if iTowns does the job, we will probably add some PR regarding the
"touch" controls, shouldn't be too hard to do

vpicavet:
@fnicollet just a side note : when replying to GH issues via email, be careful not to include your email signature which includes your mobile phone number. I removed it from your previous comment.

fnicollet:
@vpicavet Thanks! I removed my signature, to avoid future errors :)

PR #650, opened 2018-02-09
fix(examples): make examples test friendly

peppsac:
Make sure itowns' view is always accessible to ease testing (for instance
injecting a test script in an example page to query the View state).

(see also #581)

### DIFF
--- a/examples/panorama.html
+++ b/examples/panorama.html
@@ -75,13 +75,14 @@
                         protocol: 'static',
                         id: _id,
                         projection: 'EPSG:4326',
+                    }).then(() => {
+                        view.notifyChange(false);
                     });
 
                     layers.push(view.getLayers(function (l) { return (l.id == _id); })[0]);
                     index ++;
                 }
                 view.camera.camera3D.far = 10000;
-                view.notifyChange(true);
 
                 // Setup debug menu
                 var gui = new dat.GUI();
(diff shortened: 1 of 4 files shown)


examples/panorama.html
```diff
@@ -75,13 +75,14 @@
                         protocol: 'static',
                         id: _id,
                         projection: 'EPSG:4326',
+                    }).then(() => {
+                        view.notifyChange(false);
```
autra, line 79: I'm wondering if this shouldn't be done automatically in `addLayer` (or with `whenReady` in _preprocessDataLayer`)?

src/Renderer/c3DEngine.js
```diff
@@ -89,6 +89,21 @@ function c3DEngine(rendererOrDiv, options = {}) {
         throw new Error('WebGL unsupported');
     }
 
+    if (!renderer && options.logarithmicDepthBuffer) {
```
autra: You'll need a rebase I guess.

examples/panorama.html
peppsac: Indeed it could be done in `whenReady`. Do you want me to do this in this PR?

src/Renderer/c3DEngine.js
peppsac: Oops :)

examples/panorama.html
autra: We can leave it like it is here, and remove it in another PR. Whatever suits you best :-)

autra (approved):
LGTM :-)

PR #659, opened 2018-02-21
feat(core): re-enable three.js object sorting

peppsac:
It was disabled in ee7cfa0088c2d987868ef9660f3056690555ba93 because it wasn't relevant
when using RTC.
Since we removed RTC from iTowns we can reenable object sorting: it's not very costly
(one mat4 multiplication per object per frame) and it allows us to use the renderOrder
feature from three.js.

### DIFF
--- a/src/Renderer/c3DEngine.js
+++ b/src/Renderer/c3DEngine.js
@@ -117,7 +117,7 @@ function c3DEngine(rendererOrDiv, options = {}) {
 
     this.renderer.setClearColor(0x030508);
     this.renderer.autoClear = false;
-    this.renderer.sortObjects = false;
+    this.renderer.sortObjects = true;
 
     if (!renderer) {
         this.renderer.setPixelRatio(viewerDiv.devicePixelRatio);


gchoqueux:
The sort object  is for the management of transparency, isn't it?
The sorting may be wrong because of the vertex moving with the DEM

peppsac:
> The sort object is for the management of transparency, isn't it?

No.
It's to limit overdraw, reduce opengl state changes and to be able to force the drawing of an object after another, within the same draw list (three has 2 draw list: 1 for opaque objects, 1 for transparent objects).

See:
https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L1345
and:
https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLRenderLists.js#L7

gchoqueux:
is it necessary because it can be wrong with displacement shader?

peppsac:
> is it necessary because it can be wrong with displacement shader?

I'm not sure to understand what you mean but I'll try to answer anyway.
The object sorting feature is useful because:
- it allows to use the `renderOrder` feature
- it may improve performance: reduce pixel redraw (opaque objects are sorted from nearest to fursthest), reduce opengl state change (object with the same material are next to each other in the list)
- doesn't affect the visual result: because in the end the depth buffer guarantees correct results.

gchoqueux:
the tiles's vertex are moved by the vertex shader so their sort is wrong

peppsac:
> the tiles's vertex are moved by the vertex shader so their sort is wrong

The sort feature is an heuristic to reduce overdraw and state change, so it can't be correct or wrong.

The current drawing order used is: in the processing order.
This PR enable a smarter ordering mode, that's it.

gchoqueux:
ok

gchoqueux (approved):

PR #660, opened 2018-02-22
Release v2.3.0.

gchoqueux:
Bumping version for release

@iTowns/reviewers

PR #663, opened 2018-02-22
chore(packaging): update to webpack 3

autra:
## Description

This updates webpack to version 3. 

## Motivation and Context

The main motivation was to fix sourcemaps in firefox, they were broken with the version of webpack we are currently using.

A second motivation was to test a new optimization option that landed in webpack3, possibly making our production build faster. It's not included in this PR because I need to test the impact.

As we are currently releasing (#660), I let @iTowns/reviewers judge whether or not we should merge it before or after this release

Pros:
- it corrects a bug
- there is no incompatibility between version 2 and 3 and webpack changelogs says we shouldn't have to change anything (and according to my test, that's the case)

cons:
- well it's a major upgrade the day before the release :-)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,12 +5,12 @@
   "requires": true,
   "dependencies": {
     "accepts": {
-      "version": "1.3.3",
-      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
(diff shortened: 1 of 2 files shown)


zarov:
Yes to this, especially the sourcemaps fix ! I don't think merging it now should break anything, as it is a development tool, and I don't see how it can impact the production release.

zarov (approved):
Good to merge once 2.3.0 has been tagged.

PR #664, opened 2018-02-22
Fix(gis): correctly set extent _internalStorageUnit when intersecting

autra:
## Description

Everthing is in the title, we were setting cardinals in a unit that *could* be different from the default for this crs (as we support it). The new extent must have its internalStorageUnit correctly set.

I need that to finish #597

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -354,11 +354,13 @@ Extent.prototype.intersect = function intersect(other) {
     if (!this.intersectsExtent(other)) {
         return new Extent(this.crs(), 0, 0, 0, 0);
     }
-    return new Extent(this.crs(),
+    const extent = new Extent(this.crs(),
         Math.max(this.west(), other.west(this._internalStorageUnit)),
         Math.min(this.east(), other.east(this._internalStorageUnit)),
         Math.max(this.south(), other.south(this._internalStorageUnit)),
         Math.min(this.north(), other.north(this._internalStorageUnit)));
+    extent._internalStorageUnit = this._internalStorageUnit;
+    return extent;
 };
 
 


peppsac (approved):

src/Core/Geographic/Extent.js
```diff
@@ -354,11 +354,13 @@ Extent.prototype.intersect = function intersect(other) {
     if (!this.intersectsExtent(other)) {
         return new Extent(this.crs(), 0, 0, 0, 0);
     }
-    return new Extent(this.crs(),
+    const extent = new Extent(this.crs(),
         Math.max(this.west(), other.west(this._internalStorageUnit)),
         Math.min(this.east(), other.east(this._internalStorageUnit)),
         Math.max(this.south(), other.south(this._internalStorageUnit)),
         Math.min(this.north(), other.north(this._internalStorageUnit)));
+    extent._internalStorageUnit = this._internalStorageUnit;
```
zarov, line 362: As always, we should avoid accessing a private member like this. As I saw that this is used heavily in other files, what about adding an accessor to it or making it public ?

autra: > As always, we should avoid accessing a private member like this

Not here. Even following the strictest interpretation of this rule, this is fine, because we are accessing it *inside* `Extent.js`.

The reason behind it is that private fields are here to hide implementation details, but that's totally normal that *inside* the class or module, you can exploit this implementation detail.

FWIW, even java (and every languages that allows private members that I know of) allows this. For instance, this compiles and runs just fine :-)

```java
/* Main.java */
public class Main {
  public static void main(String[] args) {
    System.out.println("coucou");

    Foo foo = new Foo(1);
    Foo foo2 = new Foo(5);

    System.out.println(foo.addFoo(foo2));
  }
}

/* Foo.java */
public class Foo {
  private int bar;

  public Foo(int bar) {
    this.bar = bar;
  }

  public Foo addFoo(Foo foo2) {
    /*********************************************************
     * here I'm accessing a private properties of another Foo
     *********************************************************/
    Foo result = new Foo(this.bar + foo2.bar); 
    return result;
  }

  public String toString() {
    return "I'm a foo with bar: " + this.bar;
  }
}
```

autra: Btw, in the team, we have different interpretation of this private `_`, some of us think of it as a project-wide private. I personnally prefer the strict interpretation of class/module-wide private à la Java. We could also find a middle-ground and use it like it was a java `protected` field (so package-private), but that would mean having a directory structure that would look like packages. In my opinion, it's not the case yet in iTowns, and I'm not sure it's worth the trouble.

zarov: Thanks, good to know. However note that `_internalStorageUnit` is used in `OBB` and `Coordinates`, among other things.

autra: yep, that is subject to debate indeed. 

For `OBB`, this should be an additional optional constructor parameter I  guess. 

For coordinates, I'm less disturbed by it, as `Coordinates` and `Extent` are very much linked together in their logic (and they are in the same `geographic` "package"). A solution would be to have accessors of course, or maybe move `offsetInExtent` in `Extent.js` ? Or even make them a single module... Or leave it like that, imo it's not a big problem, no strong opinion there :-)

PR #665, opened 2018-02-22
Refacto rasterizer: guess the file type based on its content

autra:
## Description

We can easily guess the type of these vecto data when we get them. Therefore, `layer.format` is not useful here, and we can avoid guessing the content from the extension, which is not guaranteed to work with http request (and endpoint does not need to have the extension). 

Depends on #664

### DIFF
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -92,6 +92,7 @@ GeometryLayer.prototype.attach = function attach(layer) {
 GeometryLayer.prototype.detach = function detach(layer) {
     const count = this._attachedLayers.length;
     this._attachedLayers = this._attachedLayers.filter(attached => attached.id != layer.id);
+    layer.parentLayer = null;
     return this._attachedLayers.length < count;
 };
 
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -92,7 +92,6 @@ GeometryLayer.prototype.attach = function attach(layer) {
 GeometryLayer.prototype.detach = function detach(layer) {
     const count = this._attachedLayers.length;
     this._attachedLayers = this._attachedLayers.filter(attached => attached.id != layer.id);
-    layer.parentLayer = null;
     return this._attachedLayers.length < count;
 };
 
--- a/src/Core/Scheduler/Providers/Raster_Provider.js
+++ b/src/Core/Scheduler/Providers/Raster_Provider.js
@@ -11,28 +11,6 @@ import Feature2Texture from '../../../Renderer/ThreeExtended/Feature2Texture';
 import GeoJSON2Features from '../../../Renderer/ThreeExtended/GeoJSON2Features';
 import Fetcher from './Fetcher';
 
-const supportedFormats = [
-    'vector/kml',
-    'vector/gpx',
-    'vector/geojson',
-];
-
-const fetcher = {
-    'vector/kml': Fetcher.xml,
-    'vector/gpx': Fetcher.xml,
-    'vector/geojson': Fetcher.json,
-};
-
-function getExtension(path) {
-    path = String(path);
-    const basename = path.split(/[\\/]/).pop();
-    const pos = basename.lastIndexOf('.');
-    if (basename === '' || pos < 1) {
-        return '';
-    }
-    return basename.slice(pos + 1);
-}
-
 function getExtentFromGpxFile(file) {
     const bound = file.getElementsByTagName('bounds')[0];
     if (bound) {
@@ -51,7 +29,7 @@ function createTextureFromVector(tile, layer) {
     }
 
     if (layer.type == 'color') {
-        const coords = tile.extent.as(layer.projection);
+        const coords = tile.extent;
         const result = { pitch: new THREE.Vector4(0, 0, 1, 1) };
         result.texture = Feature2Texture.createTextureFromFeature(layer.feature, tile.extent, 256, layer.style);
         result.texture.extent = tile.extent;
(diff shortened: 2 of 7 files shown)


src/Core/Scheduler/Providers/Raster_Provider.js
```diff
@@ -73,81 +51,64 @@ export default {
...
+
+        // Rasterization of data vector
+        // It shouldn't use parent's texture outside its extent
+        // Otherwise artefacts appear at the outer edge
+        layer.noTextureParentOutsideLimit = true;
+
+        return Fetcher.text(layer.url, layer.networkOptions).then((text) => {
+            let geojson;
+            // if it's an xml file, then it can be kml or gpx
+            if (text.startsWith('<')) {
```
zarov: Bad assumption, for all I know there could be a `\n` (due to bad copy/paste) at the beginning of my file.

src/Core/Scheduler/Providers/Raster_Provider.js
```diff
@@ -73,81 +51,64 @@ export default {
             throw new Error('layer.url is required');
         }
 
-        const extention = getExtension(layer.url).toLowerCase();
-        const format = `vector/${extention}`;
         layer.options = layer.options || {};
+        // KML and GPX specifications all says that they should be in
+        // EPSG:4326. We still support reprojection for them through this
+        // configuration option
+        layer.projection = layer.projection || 'EPSG:4326';
```
zarov, line 58: Isn't it `reprojection` ?

src/Core/Scheduler/Providers/Raster_Provider.js
```diff
@@ -73,81 +51,64 @@ export default {
...
                 }
-                // GeoJSON2Features.parse reprojects in local tile texture space
-                // Rasterizer gives textures in this new reprojection space
-                // layer.projection is now reprojection
-                layer.originalprojection = layer.projection;
-                layer.projection = layer.reprojection;
-            });
-        }
+            } else {
+                // assume it's json
```
zarov: Oh no

zarov (changes requested):
So, correct me if I'm wrong, but it seems to me that `layer.parentLayer` is finally useful ;) Thanks for that !

While I approve those change, file format detection here is a no for me; a more accurate way should be found.

autra:
> So, correct me if I'm wrong, but it seems to me that layer.parentLayer is finally useful ;)

Actually I'm not sure about this part. Is it reasonable to think that raster layer *needs* a parent? I assumed yes, but maybe this is something we should be clearer about in the doc (and for other protocol as well).

src/Core/Scheduler/Providers/Raster_Provider.js
autra: It is the original crs of the layer (see line 61, I use it as crsIn after). I removed the reprojection options, because it is always the parent layer projection, AFAIK.

src/Core/Layer/Layer.js
```diff
@@ -92,6 +92,7 @@ GeometryLayer.prototype.attach = function attach(layer) {
 GeometryLayer.prototype.detach = function detach(layer) {
     const count = this._attachedLayers.length;
     this._attachedLayers = this._attachedLayers.filter(attached => attached.id != layer.id);
+    layer.parentLayer = null;
```
gchoqueux: It seems to me that we have avoided adding a parent layer.
This isn't detailed and motivated in the PR and in the commit.
If you want to do that, you might have to be a separate PR

gchoqueux (dismissed):

autra:
For the format detection part, I answer here because your comments are kind of linked together.

> Bad assumption, for all I know there could be a \n (due to bad copy/paste) at the beginning of my file.

AFAIK, xml files are required to start with the `<` character. I could `trim` here, but I think it's reasonable to expect the user to give a valid xml file, right?

> Oh no

What disturbs you exactly?

Let me elaborate: For these 2 assumptions, I actually rely on DOMParser.parseFromString and JSON.parse to throw an error if there are errors in the file or if they are not in a supported format. If an exception is thrown here, the promise of `view.addLayer` will be rejected, and the layer won't be added to the scene. 

Because of these 2 parsers, I don't *need* to make sure they are valid json or xml before parsing them, I just need a way to know which parser to use. Testing for `<` is perfect for that, because GeoJSON can *never* start with this char.

That being said, maybe I can:
- add a comment basically saying exactly that
- make sure error messages in the console are explicit and understandable when the file is not valid and / or supported. That might involve catching and rethrowing parsing errors

WDYT of this proposal @zarov ?

src/Core/Layer/Layer.js
zarov: No, it is currently present in `View`. So it is ok imho.

zarov:
> AFAIK, xml files are required to start with the < character. I could trim here, but I think it's reasonable to expect the user to give a valid xml file, right?

No, specification allows a `<` or a whitespace character (`\s`, `\n`, etc) (see https://stackoverflow.com/a/47803127). Trimming could be a solution

> What disturbs you exactly?

JSON shouldn't be the default case, the default case should still be an error.

>  For these 2 assumptions, I actually rely on DOMParser.parseFromString and JSON.parse to throw an error

`DOMParser.parseFromString` doesn't throw an error but return an error document.

> make sure error messages in the console are explicit and understandable when the file is not valid and / or supported. That might involve catching and rethrowing parsing errors

Yes please do ! I'm inclined to see this solution.

src/Core/Scheduler/Providers/Raster_Provider.js
zarov: Ok, I was misguided by the comment above

src/Core/Layer/Layer.js
gchoqueux: I meant to keep the link with the parent

autra:
>  specification allows a < or a whitespace character (\s, \n, etc)

Oh you're right, my mistake. Changing this.

> DOMParser.parseFromString doesn't throw an error but return an error document.

You're right again :-) I need to have a better error handling, thanks for pointing this out!

autra:
@zarov I updated this PR with a (hopefully) complete error handling. Please tell me what you think :-)

src/Core/Layer/Layer.js
autra: I preferred to add it where I needed it, than just opening a PR without precise application case (as it is a minor change really). That being said, if you agree with this `layer.parentLayer`, I can open a new PR with it, no problem.

src/Core/Scheduler/Providers/Raster_Provider.js
```diff
@@ -73,81 +51,70 @@ export default {
...
         }
 
-        const extention = getExtension(layer.url).toLowerCase();
-        const format = `vector/${extention}`;
         layer.options = layer.options || {};
+        // KML and GPX specifications all says that they should be in
+        // EPSG:4326. We still support reprojection for them through this
+        // configuration option
+        layer.projection = layer.projection || 'EPSG:4326';
+        const parentCrs = layer.parentLayer.extent.crs();
```
peppsac: An alternative to storing the `parentLayer` would be to pass it as an extra argument `preprocessDataLayer`.
WDYT?

autra: I thought about that too. I didn't because I expect we'll also need parentLayer in other parts (`executeCommand` ?), but I might be mistaken. The advantage of your solution is that attaching `parentLayer` to layer could be decided independently for each provider.

That being said, for me, having a reference to the parentLayer on the child layer makes sense. At the very least, it's not something neither unexpected nor "dangerous" in my opinion. That's why I did it like that.

Thoughts?

peppsac: The problem is when you want to add the same layer twice (e.g: you have 2 GlobeView and want to add the same color layer to each).
But I'm not sure other parts of iTowns handle this edge-case correctly, so we may want to disable it entirely (e.g: prevent adding a `Layer` instance to another `Layer`)

autra: Oh yes, I assumed this was just not supported. Do you think it'll be interesting to support it? If yes, I'm modifying this.

peppsac: Yeah, I'd prefer not adding the `parentLayer` to `Layer` for now, if the other proposal is ok for you (= adding a param to `preprocessLayer`).

autra:
@iTowns/reviewers updated

zarov (approved):
Good for me, as I think it may be changed later if we rewrite most of the Providers, and maybe by now we will find another solution than `startsWith` (but it doesn't seem there is, except by using `format`/`mimetype`).

PR #666, opened 2018-02-23
Fix(examples): avoid displaying the tooltip before the js loads

autra:
## Description

Before this change, the tooltip can be briefly seen at the top left of the screen when the page loads. We should hide it by default and show it only with javascript, to avoid showing it with the hover before the js has a chance to kick in.


peppsac (approved):
Thanks :)

PR #667, opened 2018-02-26
chore(debug): use FrameRequester instead of overriding layer.update fn

peppsac:
This helps reading profiling charts.

Note: I'd suggest adding `debug` to the allowed scopes from #518 - WDYT?


autra:
> I'd suggest adding debug to the allowed scopes from #518 - WDYT?

Good idea!

gchoqueux (approved):

Issue #642, opened 2018-02-06
Multiglobe example is broken


gchoqueux:
The PR #586 broke the [multiglobe example](http://www.itowns-project.org/itowns/examples/multiglobe.html) .

the transformation of the layer must not be taken into account

gchoqueux:
fixed in #669

PR #669, opened 2018-02-26
fix(core): wrong tile transformation

### LINKED ISSUE #668: Cubic planar example produces visual artifacts
Hello,

Trying out the "Cubic planar" example on my laptop:
http://www.itowns-project.org/itowns/examples/cubic_planar.html
When you move the camera around, it generates some artifacts. Some faces become orthogonal to the plane, making them stand out:
image
Also, some tiles never seem to load (wrong LOD or something):
image

There is no error in the console
Fabien

gchoqueux:
stop use root layer transformation to compute tile transformation
especially for the animation of the root

fix #668 and #642

### DIFF
--- a/src/Core/Scheduler/Providers/TileProvider.js
+++ b/src/Core/Scheduler/Providers/TileProvider.js
@@ -41,7 +41,6 @@ TileProvider.prototype.preprocessDataLayer = function preprocessLayer(layer, vie
     });
 };
 
-const worldQuaternion = new THREE.Quaternion();
 TileProvider.prototype.executeCommand = function executeCommand(command) {
     const extent = command.extent;
     if (command.requester &&
@@ -95,11 +94,12 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
     tile.layer = layer.id;
     tile.layers.set(command.threejsLayer);
 
-    if (parent) {
-        position.applyMatrix4(layer.object3d.matrixWorld);
-        parent.worldToLocal(position);
-        worldQuaternion.setFromRotationMatrix(parent.matrixWorld).inverse().multiply(layer.object3d.quaternion);
-        quaternion.premultiply(worldQuaternion);
+    if (parent && parent instanceof TileMesh) {
+        // get parent extent transformation
+        const pTrans = builder.computeSharableExtent(parent.extent);
+        // place relative to his parent
+        position.sub(pTrans.position).applyQuaternion(pTrans.quaternion.inverse());
+        quaternion.premultiply(pTrans.quaternion);
     }
 
     tile.position.copy(position);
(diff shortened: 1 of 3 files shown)


src/Core/Scheduler/Providers/TileProvider.js
```diff
@@ -95,11 +94,11 @@ TileProvider.prototype.executeCommand = function executeCommand(command) {
     tile.layer = layer.id;
     tile.layers.set(command.threejsLayer);
 
-    if (parent) {
-        position.applyMatrix4(layer.object3d.matrixWorld);
-        parent.worldToLocal(position);
-        worldQuaternion.setFromRotationMatrix(parent.matrixWorld).inverse().multiply(layer.object3d.quaternion);
-        quaternion.premultiply(worldQuaternion);
+    tile.extentTransform = { quaternionInv: quaternion.clone().inverse(), position: position.clone() };
```
peppsac: At the very least add a `_` prefix.
But storing 2 extra properties, only to be able to compute the transform of children is not necessary.

Here's a version that also fixes the issue:
```js
    if (parent) {
        // position / quaternion are initially in GIS coordinates
        const p1 = new THREE.Vector3();
        const p2 = p1.clone();

        position.applyMatrix4(layer.object3d.matrixWorld);

        layer.object3d.matrixWorld.decompose(p1, worldQuaternion, p2);
        quaternion.premultiply(worldQuaternion);

        // Now position / quaternion are in World coordinates
        // But we want them in parent coordinates
        const m = new THREE.Matrix4();
        m.getInverse(parent.matrixWorld);
        parent.matrixWorld.decompose(p1, worldQuaternion, p2);

        // First, transform position
        position.applyMatrix4(m);
        // Second, transform quaternion
        worldQuaternion.inverse();
        quaternion.premultiply(worldQuaternion);
    }
```

(we have to use `decompose` and not `setFromRotationMatrix` because the latter is broken when scale is not 1)

gchoqueux: I also wanted to stop the use of `layer.object3d` which does not make much sense here

We can do this

```js
if (parent && parent instanceof TileMesh) {
        // get parent extent transformation
        const pTrans = builder.computeSharableExtent(parent.extent);
        // place relative to his parent
        position.sub(pTrans.position).applyQuaternion(pTrans.quaternion.inverse());
        quaternion.premultiply(pTrans.quaternion);
}
```

peppsac: > layer.object3d which does not make much sense here`

Using `layer.object3d` is mandatory if you use `parent.matrixWorld`, because `parent` is a descendant of `layer.object3d`.

But your alternative solution is looking good too.

gchoqueux: later I will calculate the position and quaternion relative to the parent in `computeSharableExtent`

peppsac:
Looks good. Can you squash the commits and change the commit message?

gchoqueux:
what commit message is not suitable?

peppsac:
` fix(core): wrong tile transformation`: should summarize the content of the commit and must start with a verb.
And you could squash the 2nd commit in the first.

gchoqueux:
done

peppsac:
> done

You forgot to squash the 2nd commit?

gchoqueux:
squashed

peppsac (approved):

PR #670, opened 2018-02-27
fix(core): stop using wmtsc for OSM layers on globe

peppsac:
OSM servers uses the 'xyz' protocol so this commit adds proper support for these URL.
The wmstc support is still present (but not used).

### DIFF
--- a/src/Core/Scheduler/Providers/TMS_Provider.js
+++ b/src/Core/Scheduler/Providers/TMS_Provider.js
@@ -4,12 +4,17 @@ import Extent from '../../Geographic/Extent';
 
 function preprocessDataLayer(layer) {
     if (!layer.extent) {
-        throw new Error(`Missing extent property for layer '${layer.id}'`);
+        // default to the full 3857 extent
+        layer.extent = new Extent('EPSG:3857',
+            -20037508.342789244, 20037508.342789244,
+            -20037508.342789255, 20037508.342789244);
     }
-    if (!layer.projection) {
-        throw new Error(`Missing projection property for layer '${layer.id}'`);
+    if (!(layer.extent instanceof (Extent))) {
+        if (!layer.projection) {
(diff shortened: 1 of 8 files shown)


src/Core/Scheduler/Providers/TMS_Provider.js
```diff
@@ -7,12 +7,17 @@ function TMS_Provider() {
 
 TMS_Provider.prototype.preprocessDataLayer = function preprocessDataLayer(layer) {
     if (!layer.extent) {
-        throw new Error(`Missing extent property for layer '${layer.id}'`);
+        // default to the full 3857 extent
+        layer.extent = new Extent('EPSG:3857',
+            -20037508.342789244, 20037508.342789244,
```
zarov, line 9: How did you choose those bounds ? Here are [other ones](http://epsg.io/3857), and I can find other too for this projection.

zarov (commented):
Is `wmtsc` in use somewhere ? I see it in the `Scheduler`, but do we need to keep it ?

src/Core/Scheduler/Providers/TMS_Provider.js
peppsac: From proj4:
```
itowns.proj4('EPSG:3857').forward([-180, 0])
(2) [-20037508.342789244, -7.081154551613622e-10]
```
y-min/y-max are set to have a square extent (which seems to be correct).
Note that this is only a default extent value: you're free to override it.

peppsac:
Regarding `wmtsc`: I think the some people (@qdnguyen  maybe) were using it so I didn't removed it.

src/Core/Scheduler/Providers/TMS_Provider.js
zarov: Ok for letting it as it is

zarov:
I can see the usage now, that's ok for me.

zarov (approved):

PR #671, opened 2018-02-27
refactor(core): Homogenize all providers as static

zarov:
## Description
There is no reason to keep an instance of each provider, and moving them
to static like some other one (PointCloudProvider for example) do no
harm, but homogenize all providers to one structure.

Note that the function `getPointOrder` in the WFS provider is unused, should I delete it in this PR, or the associated TODO will be done by someone soon ?

Also note that there are two required functions in a provider:
preprocessDataLayer and executeCommand, that are verified when adding a
provider, in the scheduler.

## Motivation and Context

This comes in the work of #182, part of rewriting the Providers/Loader/etc system.


src/Core/Scheduler/Scheduler.js
```diff
@@ -162,8 +161,14 @@ Scheduler.prototype.execute = function execute(command) {
...
 
-
 Scheduler.prototype.addProtocolProvider = function addProtocolProvider(protocol, provider) {
+    if (typeof (provider.executeCommand) !== 'function') {
+        throw new Error(`Can't add provider for ${protocol}: missing a executeCommand function.`);
+    }
+    if (typeof (provider.preprocessDataLayer) !== 'function') {
+        throw new Error(`Can't add provider for ${protocol}: missing a preprocessDataLayer function.`);
+    }
+
```
autra, line 171: Nice! That's the first step to *really* allow people to add their own providers (I mean, it was already possible, but you needed to know the internals of iTowns). Second step would be to document it :-)

autra (approved):
LGTM, thanks!

I didn't test it though, so if we want this merged before the release, you might want to make sure all the affected examples are still working ok.

peppsac:
Could you remove the unused providers as well (`BuildingBox_Provider.js`, `KML_Provider.js`)?

src/Core/Scheduler/Scheduler.js
zarov: What do you have in mind for documentation ? Something like iTowns/itowns.github.io#15 ? I can't see where else we could add it.

zarov: Maybe if we document the `Scheduler` in fact

peppsac: Using http://usejsdoc.org/tags-interface.html#virtual-comments in the Scheduler makes sense.
But maybe you should finish your refactoring commits before adding documentation...

zarov: Yes, will do in another PR.

zarov:
@autra tested it, all good, but let's wait the 2.3.0 for merging.

@peppsac done.

peppsac:
You may want to add a `BREAKING CHANGE: .....` statement at the end of your commit message to ease the changelog writing.

PR #672, opened 2018-02-27
refactor(core): improve render view to buffer / target functions

peppsac:
Improve these 2 functions and document them.

Work-around a oddity in three.js causing setScissor / setViewport functions
behaving oddly when using a RenderTarget.

Changed coordinates to explicit window-coordinate (0,0: top-left) to avoid
requesting the user to do (height - y).

BREAKING CHANGE: c3DEngine.renderViewTobuffer has been renamed renderViewToBuffer
and has a new signature.

This PR depends on #635

### DIFF
--- a/src/Core/Prefab/PlanarView.js
+++ b/src/Core/Prefab/PlanarView.js
@@ -1,7 +1,7 @@
 import * as THREE from 'three';
 
 import View from '../View';
-import { RENDERING_PAUSED } from '../MainLoop';
+import { RENDERING_PAUSED, MAIN_LOOP_EVENTS } from '../MainLoop';
 import RendererConstant from '../../Renderer/RendererConstant';
 
 import { GeometryLayer } from '../Layer/Layer';
@@ -148,6 +148,12 @@ function PlanarView(viewerDiv, extent, options = {}) {
 
     this._renderState = RendererConstant.FINAL;
     this._fullSizeDepthBuffer = null;
+    this.addFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, () => {
+        if (this._fullSizeDepthBuffer != null) {
+            // clean depth buffer
+            this._fullSizeDepthBuffer = null;
+        }
+    });
 
     this.tileLayer = tileLayer;
 }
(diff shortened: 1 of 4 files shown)


gchoqueux:
@peppsac you could rebase this PR?

peppsac:
Done

src/Renderer/c3DEngine.js
```diff
@@ -141,21 +145,68 @@ c3DEngine.prototype.getRenderer = function getRenderer() {
     return this.renderer;
 };
 
-c3DEngine.prototype.renderViewTobuffer = function renderViewTobuffer(view, buffer, x, y, width, height) {
-    // TODO Deallocate render texture
+/**
+ * Render view to a Uint8Array.
+ *
+ * @param {View} view - The view to render
+ * @param {object} [zone] - partial zone to render (zone x/y uses view coordinates)
```
autra: You can describe your object's property with additional `@param {number} zone.x - <desc>` tags

src/Renderer/c3DEngine.js
```diff
@@ -141,21 +145,68 @@ c3DEngine.prototype.getRenderer = function getRenderer() {
...
-    this.renderer.setScissorTest(true);
-    this.renderer.clearTarget(buffer, true, true, false);
-    this.renderer.render(view.scene, view.camera.camera3D, buffer);
-    this.renderer.setScissorTest(false);
-    var pixelBuffer = new Uint8Array(4 * width * height);
-    this.renderer.readRenderTargetPixels(buffer, x, y, width, height, pixelBuffer);
+
+    // Don't use setViewport / setScissor after setting the renderTarget, because
+    // they affect non-rendertarget rendering as well.
+    // Instead set properties before calling setRenderTarget
```
autra: Not sure I understand this well, but I suggest:

> Don't use setViewport / setScissor on renderer because they would affect screen rendering as well. Instead set them on the render target

because the problem is not the call order (as I thought first because of `after setting...`) but the fact they are set on the renderer instead of the renderTarget. But that's assuming I understand this well...

src/Core/Prefab/PlanarView.js
```diff
@@ -148,6 +148,12 @@ function PlanarView(viewerDiv, extent, options = {}) {
 
     this._renderState = RendererConstant.FINAL;
     this._fullSizeDepthBuffer = null;
+    this.addFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, () => {
```
autra, line 151: Can you elaborate why you need that?

autra (changes requested):

src/Core/Prefab/PlanarView.js
peppsac: Because `_fullSizeDepthBuffer` will never get invalidated otherwise.

So we clear it before each render, and `_fullSizeDepthBuffer` only act as a cache (ie: if you request depth values several times without redrawing the scene the depth buffer will only be read once).

src/Renderer/c3DEngine.js
peppsac: Done

src/Renderer/c3DEngine.js
peppsac: Yep, done.

autra (approved):
LGTM thanks!

PR #673, opened 2018-02-28
add picking in wfs examples

gchoqueux:
add picking in wfs examples and show properties

### DIFF
--- a/examples/globe_wfs_extruded.html
+++ b/examples/globe_wfs_extruded.html
@@ -24,6 +24,28 @@
                 top: 0;
                 margin-left: 0;
             }
+            #batchInformationDiv {
+                position: absolute;
+                z-index: 0;
+                top: 0;
+                right: 0;
+                color: white;
+                color: #eee;
+                font: 11px 'Lucida Grande',sans-serif;
+                line-height: normal;
+                text-shadow: 0 -1px 0 #111;
+                padding: 0 1rem;
+                background: #1a1a1a;
+                border: 1px solid #2c2c2c;
+                opacity: 0.8;
+            }
+            #batchInformationDiv > p {
+                margin: 0.5rem 0;
+            }
+
+            #batchInformationDiv > ul {
+                padding: 0 1rem;
+            }
             @media (max-width: 600px) {
                 #menuDiv {
                     display: none;
@@ -44,7 +66,14 @@
             var renderer;
             var exports = {};
         </script>
-        <script src="globe_wfs_extruded.js"></script>
+        <script src="globe_wfs_extruded.js">
+            console.log('label');
+        </script>
+        <div id="batchInformationDiv">
+            <p><b>Information Batiment</b></p>
+            <ul id="info">
+            </ul>
+        </div>
         <script type="text/javascript">
             /* global itowns, document, GuiTools, globeView, promises */
             var menuGlobe = new GuiTools('menuDiv');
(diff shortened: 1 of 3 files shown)


examples/globe_wfs_extruded.html
```diff
@@ -60,10 +89,37 @@
...
-            const d = new debug.Debug(globeView, menuGlobe.gui);
+            var d = new debug.Debug(globeView, menuGlobe.gui);
             debug.createTileDebugUI(menuGlobe.gui, globeView, globeView.wgs84TileLayer, d);
 
-            for (const layer of globeView.getLayers()) {
+            function picking(event) {
+                var htmlInfo = document.getElementById('info');
+                htmlInfo.innerHTML = ' ';
+
+                var intersects = globeView.pickObjectsAt(event, 'WFS Buildings');
```
peppsac, line 99: I think `intersects` could be replaced with `picks` or `picked`

examples/globe_wfs_extruded.html
```diff
@@ -60,10 +89,37 @@
...
+
+                for (var i = 0; i < intersects.length; i++) {
+                    var interAttributes = intersects[i].object.geometry.attributes;
+                    if (interAttributes && intersects[i].object instanceof itowns.THREE.Mesh) {
+                        if (interAttributes.batchid) {
+                            var face = intersects[i].face.a;
+                            var batchid = interAttributes.batchid.array[face];
+                            var batchTable = intersects[i].object.parent.features[batchid].properties;
+                            Object.keys(batchTable).map(function(objectKey) {
```
peppsac, line 108: Do you think you could add a more user friendly interface here?
Something that would allow the user to easily get the properties from the picking result.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -202,10 +202,14 @@ function coordinateToLines(coordinates, properties, options) {
...
 }
 
-function coordinateToPolygon(coordinates, properties, options) {
+function coordinateToPolygon(coordinates, properties, options = {}) {
     const indices = [];
     const vertices = new Float32Array(3 * coordinates.coordinates.length);
     const colors = new Uint8Array(3 * coordinates.coordinates.length);
+    let batchId;
+    if (options.buildBatchId) {
+        batchId = new Uint16Array(2 * coordinates.coordinates.length);
```
peppsac: You only need 1 value per vertices in the non-extruded case.

peppsac: And I wouldn't mind always building the `batchId` property, instead of using using an opt-in option.

examples/globe_wfs_extruded.html
peppsac: Actually it would make sense to provide a custom `pickObjectsAt` for WFS_Provider.
It would use `Picking.pickObjectsAt()` method, and then would enrich the result by adding an array of properties.

That way, as a user you could do:
```js
for (var i = 0; i < intersects.length; i++) {
    var obj = intersects[i].object;
    var batchTable = intersects[i].properties;
}
```

WDYT?

peppsac (changes requested):

examples/globe_wfs_extruded.html
gchoqueux: It's more representative because it returns all the buildings under the mouse, even the hidden buildings, moreover it's the result of a raycaster. Generally picking allows to select a single entity. And the name seems to be consistent with your [last commit](https://github.com/iTowns/itowns/pull/635/commits/b0bbc5e5071c357ae872b45cec731e05df3bbaca#diff-8eb5b5e6b345c723fa6df0ae1c3c7c39R158)

examples/globe_wfs_extruded.html
gchoqueux: Yes completely, as I said, I would do that in the method `on()`

```js
 function picking(event) {
        const batchTable = event.features[i].properties;
 }
```

gchoqueux: Yes completely, as I said, I would do that in the method `on()`

```js
featureLayer.on(event) {
        const batchTable = event.features[i].properties;
 }
```

src/Renderer/ThreeExtended/Feature2Mesh.js
gchoqueux: > You only need 1 value per vertices in the non-extruded case.

oops

> And I wouldn't mind always building the batchId property, instead of using using an opt-in option.

ok

gchoqueux:
@peppsac done

examples/globe_wfs_extruded.html
peppsac: Well my last commit was a bad idea.

Here's why: intersects makes sens because you know `pickObjectsAt()` uses raycasting for wfs features. But a regular itowns will not know this - hence my suggestion.

examples/globe_wfs_extruded.html
peppsac: Don't you want to do this in this PR (not the `on()` thing, I'm only suggestion to implement my proposal above)?
It would add value to itowns and simplify the example code.

examples/globe_wfs_extruded.js
```diff
@@ -138,7 +138,8 @@ globeView.addLayer({
     convert: itowns.Feature2Mesh.convert({
         color: colorBuildings,
         altitude: altitudeBuildings,
-        extrude: extrudeBuildings }),
+        extrude: extrudeBuildings,
+        buildBatchId: true }),
```
peppsac, line 142: You can remove this change then.

gchoqueux:
Not necessary with merged PR #691

PR #674, opened 2018-03-01
Refactor(core): unify SSE computation methods

peppsac:
Until this PR each geometry type had its own SSE computation methods.
The goal of this PR is to provide a single method based on:
  - bounding-box (BB)
  - geometricError (in meters)
  - distance

The computation is simple:
  - transform camera in BB's basis
  - compute distance camera - BB
  - project geometricError on screen at distance

The last step produce a screen-space-error in pixels.

Geometric error values depend on the data type:
  - 3dtiles: the tileset defines a geometricError per node so we can use this value
  - pointcloud: PotreeConverter defines a spacing, which is the average distance
between points in each node. We use: geometricError = spacing / 2^depth
  - tiles: this PR adapted the existing formula to use consistent units

The second commit adds a SSE inspector debug tool:
sse_inspector

red: computed SSE
yellow: texture resolution (only used for tile sse)
green: sse treshold

### DIFF
--- a/src/Core/ScreenSpaceError.js
+++ b/src/Core/ScreenSpaceError.js
@@ -7,6 +7,7 @@ const m3 = new THREE.Matrix4();
 const basis = [
     new THREE.Vector3(1, 0, 0),
     new THREE.Vector3(0, 1, 0),
+    new THREE.Vector3(0, 0, 1),
 ];
 
 function computeVectorSizeAtDistance(vector, matrix, camera, distance, _3d) {
@@ -14,7 +15,7 @@ function computeVectorSizeAtDistance(vector, matrix, camera, distance, _3d) {
     basis[1].set(0, vector.y, 0);
 
     if (_3d) {
-        basis.push(new THREE.Vector3(0, 0, vector.z));
+        basis[2].set(0, 0, vector.z);
     }
 
     m2.identity();
@@ -22,7 +23,9 @@ function computeVectorSizeAtDistance(vector, matrix, camera, distance, _3d) {
     m3.identity();
     m3.extractRotation(camera.camera3D.matrixWorldInverse);
 
-    for (const b of basis) {
+    for (let i = 0; i < (_3d ? 3 : 2); i++) {
+        const b = basis[i];
+
         // Apply rotation
         b.applyMatrix4(m2);
         // Apply inverse camera rotation
@@ -37,6 +40,16 @@ function computeVectorSizeAtDistance(vector, matrix, camera, distance, _3d) {
         b.y = b.y * camera.height * 0.5;
     }
 
+    return basis.map(b => b.length());
+
+    const sse = {
+        x: basis[0].length(),
+        y: basis[1].length(),
+    };
+    if (_3d) {
+        sse.z = basis[2].length();
+    }
+    return sse;
     const lengthsq = basis.map(b => b.lengthSq());
     const min = Math.min.apply(Math, lengthsq);
     return Math.sqrt(min);
@@ -78,7 +91,7 @@ export default {
 
         if (distance <= geometricError) {
             return {
-                sse: Infinity,
+                sse: [Infinity, Infinity, Infinity],
                 distance,
             };
         }
--- a/src/Core/ScreenSpaceError.js
+++ b/src/Core/ScreenSpaceError.js
@@ -1,0 +1,88 @@
+import * as THREE from 'three';
+
+const v = new THREE.Vector3();
+const m = new THREE.Matrix4();
+const m2 = new THREE.Matrix4();
+const m3 = new THREE.Matrix4();
+
+function compute(vector, matrix, camera, distance, _3d) {
+    const basis = [
+        new THREE.Vector3(vector.x, 0, 0),
+        new THREE.Vector3(0, vector.y, 0),
+    ];
+
+    if (_3d) {
+        basis.push(new THREE.Vector3(0, 0, vector.z));
+    }
+
+    m2.identity();
+    m2.extractRotation(matrix);
+    m3.identity();
+    m3.extractRotation(camera.camera3D.matrixWorldInverse);
+
+    for (const b of basis) {
+        // Apply rotation
+        b.applyMatrix4(m2);
+        // Apply inverse camera rotation
+        b.applyMatrix4(m3);
+        // Move at 'distance' from camera
+        b.z += -distance;
+        // Project on screen
+        b.applyMatrix4(camera.camera3D.projectionMatrix);
+        // cancel z component
+        b.z = 0;
+        b.x = b.x * camera.width * 0.5;
+        b.y = b.y * camera.height * 0.5;
+    }
+
+    const lengthsq = basis.map(b => b.lengthSq());
+    const min = Math.min.apply(Math, lengthsq);
+    return Math.sqrt(min);
+}
+
+function findBox3Distance(camera, box3, matrix) {
+    // TODO: can be cached
+    m.getInverse(matrix);
+    // Move camera position in box3 basis
+    // (we don't transform box3 to camera basis because box3 are AABB)
+    const pt = camera.camera3D.position
+        .clone(v).applyMatrix4(m);
+    // Compute distance between the camera / box3
+    return box3.distanceToPoint(pt);
+}
+
+function computeSizeFromGeometricError(box3, geometricError) {
+    const size = box3.getSize();
+    // Build a vector with the same ratio than box3,
+    // and with the biggest component being geometricError
+    size.multiplyScalar(geometricError /
+        Math.max(size.x, Math.max(size.y, size.z)));
+    return size;
+}
+
+export default {
+    MODE_2D: 1,
+
+    MODE_3D: 2,
+
+    computeFromBox3(camera, box3, matrix, geometricError, mode) {
+        const distance = findBox3Distance(camera, box3, matrix);
+
+        if (distance <= geometricError) {
+            return {
+                sse: Infinity,
+                distance,
+            };
+        }
+
+        const size = computeSizeFromGeometricError(box3, geometricError);
+
+        const sse = compute(size, matrix, camera, distance, mode == this.MODE_3D);
+
+        return {
+            sse,
+            distance,
+            size,
+        };
+    },
+};
--- a/src/Core/ScreenSpaceError.js
+++ b/src/Core/ScreenSpaceError.js
@@ -4,12 +4,14 @@ const v = new THREE.Vector3();
 const m = new THREE.Matrix4();
 const m2 = new THREE.Matrix4();
 const m3 = new THREE.Matrix4();
+const basis = [
+    new THREE.Vector3(1, 0, 0),
+    new THREE.Vector3(0, 1, 0),
+];
 
-function compute(vector, matrix, camera, distance, _3d) {
-    const basis = [
-        new THREE.Vector3(vector.x, 0, 0),
-        new THREE.Vector3(0, vector.y, 0),
-    ];
+function computeVectorSizeAtDistance(vector, matrix, camera, distance, _3d) {
+    basis[0].set(vector.x, 0, 0);
+    basis[1].set(0, vector.y, 0);
 
     if (_3d) {
         basis.push(new THREE.Vector3(0, 0, vector.z));
@@ -61,8 +63,14 @@ function computeSizeFromGeometricError(box3, geometricError) {
 }
 
 export default {
+    /*
+     * Compute SSE based on the 2D bounding-box (ignore z size)
+     */
     MODE_2D: 1,
 
+    /*
+     * Compute SSE based on the 3D bounding-box
+     */
     MODE_3D: 2,
 
     computeFromBox3(camera, box3, matrix, geometricError, mode) {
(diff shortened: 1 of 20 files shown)


peppsac:
Also note : this PR provides almost everything needed to compute the camera far / near plane correctly in realtime (near would be minimum distance from all updated elements, far would be the max)

src/Core/Scheduler/Providers/PointCloudProvider.js
```diff
@@ -83,7 +83,15 @@ function parseOctree(layer, hierarchyStepSize, root) {
...
                     }
-                    const item = { numPoints: n, childrenBitField: c, children: [], name: childname, baseurl: url, bbox: bounds };
+                    const item = {
+                        numPoints: n,
+                        childrenBitField: c,
+                        children: [],
+                        name: childname,
+                        baseurl: url,
+                        bbox: bounds,
+                        geometricError: layer.metadata.spacing / Math.pow(2, childname.length),
```
gchoqueux: Evoked in the commit message, there is a reference that explains the calculation?

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,88 @@
+import * as THREE from 'three';
+
+const v = new THREE.Vector3();
+const m = new THREE.Matrix4();
+const m2 = new THREE.Matrix4();
+const m3 = new THREE.Matrix4();
+
+function compute(vector, matrix, camera, distance, _3d) {
```
gchoqueux: `compute` what?

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,88 @@
...
+    const basis = [
+        new THREE.Vector3(vector.x, 0, 0),
```
gchoqueux: `THREE.Vector3` to declare outside the function

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    const lengthsq = basis.map(b => b.lengthSq());
+    const min = Math.min.apply(Math, lengthsq);
+    return Math.sqrt(min);
+}
+
+function findBox3Distance(camera, box3, matrix) {
+    // TODO: can be cached
+    m.getInverse(matrix);
```
gchoqueux, line 45: `matrix` it's OBB worldMatrix? you would cache in `OBB.worldMatrixInv`?

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,88 @@
...
+        Math.max(size.x, Math.max(size.y, size.z)));
+    return size;
+}
+
+export default {
+    MODE_2D: 1,
+
+    MODE_3D: 2,
+
+    computeFromBox3(camera, box3, matrix, geometricError, mode) {
```
gchoqueux: there's no method to compute with sphere?

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,88 @@
...
+        if (distance <= geometricError) {
+            return {
+                sse: Infinity,
+                distance,
+            };
+        }
+
+        const size = computeSizeFromGeometricError(box3, geometricError);
+
+        const sse = compute(size, matrix, camera, distance, mode == this.MODE_3D);
```
gchoqueux: `compute` calculates a pixel projection size?

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,88 @@
...
+    const size = box3.getSize();
+    // Build a vector with the same ratio than box3,
+    // and with the biggest component being geometricError
+    size.multiplyScalar(geometricError /
+        Math.max(size.x, Math.max(size.y, size.z)));
+    return size;
+}
+
+export default {
+    MODE_2D: 1,
```
gchoqueux: could you explain the modes?

src/Process/3dTilesProcessing.js
```diff
@@ -225,25 +226,30 @@ const worldPosition = new THREE.Vector3();
...
+            node.geometricError,
+            ScreenSpaceError.MODE_3D);
+    }
+    if (node.boundingVolume.sphere) {
+        // TODO USE http://iquilezles.org/www/articles/sphereproj/sphereproj.htm
         worldPosition.setFromMatrixPosition(node.matrixWorld);
         cameraLocalPosition.copy(camera.camera3D.position).sub(worldPosition);
-        node.distance = Math.max(0.0, node.boundingVolume.sphere.distanceToPoint(cameraLocalPosition));
+        const distance = Math.max(0.0, node.boundingVolume.sphere.distanceToPoint(cameraLocalPosition));
+        return camera.preSSE * (node.geometricError / distance);
```
gchoqueux: why not place it in `ScreenSpaceError.js`

src/Process/GlobeTileProcessing.js
```diff
@@ -133,9 +96,14 @@ export function globeSubdivisionControl(minLevel, maxLevel, sseThreshold, maxDel
             return false;
         }
 
-        const sse = computeNodeSSE(context.camera, node);
-
-        return SSE_SUBDIVISION_THRESHOLD < sse;
+        node.sse = ScreenSpaceError.computeFromBox3(
```
gchoqueux: why you replace the bounding sphere with OBBs?

src/Process/GlobeTileProcessing.js
```diff
@@ -164,6 +132,7 @@ export function globeSchemeTileWMTS(type) {
 }
 
 export function computeTileZoomFromDistanceCamera(distance, view) {
+    // TODO fixme
```
gchoqueux, line 138: do you have an idea?

gchoqueux (changes requested):
review on first commit: `refactor(core): unify SSE computation methods `

src/Core/Scheduler/Providers/PointCloudProvider.js
peppsac: This is by construction: see https://github.com/peppsac/PotreeConverter/blob/master/PotreeConverter/src/PotreeWriter.cpp#L73

src/Core/ScreenSpaceError.js
peppsac: I'd prefer not to cache anything first (hence the TODO node).

I believe we need to rework the `notifyChange()` mechanism to allow more precise updates / cache invalidation.

src/Core/ScreenSpaceError.js
peppsac: Not yet. As your other comment suggests I could move the sphere -> sse code from 3dtiles to here.

src/Core/ScreenSpaceError.js
peppsac: Yes, I'll change the function name.

src/Core/ScreenSpaceError.js
peppsac: Yes I forgot to add a comment.
The idea here is: do we want to use the 2d bounding-box (x, y) or the 3d one.
For tiles we want to use the 2d one, while for other objects (pointcloud, 3dtiles) the 3d one makes more sense.

src/Process/GlobeTileProcessing.js
peppsac: Because the OBB is closer to the real volume of the `node` then a sphere.

src/Process/GlobeTileProcessing.js
peppsac: I didn't even try to fix this for now.

I'd rather make sure that the sse computation introduced in this PR is ok.

peppsac:
PR updaded.
Note: the computation in `ScreenSpaceError.js` can probably be simplified.
But it's easier to review the non-optimized form and then as a last step simplify where possible.

peppsac:
Screenshots showing subdivision differences between master/PR :

Master:
pr_off
PR :
pr_on

Or for pointclouds:
Master:
pr_off2
PR:
pr_on2

gchoqueux:
The tiles count between `camera's tilt` equals 0 and 85 is significant
But, qualitatively, it's too important for 0 and insufficient for 85.
can you balance the subdivision?

`tilt == 0`
tilt == 0
`tilt == 85`
tilt == 85

peppsac:
@gchoqueux : thanks for the feedback.

The current logic is: we project both the bounding box [x and y axis](https://github.com/iTowns/itowns/pull/674/files#diff-bc30a9362a318408471b5a8f16ce5b76R10) on the [screen](https://github.com/iTowns/itowns/pull/674/files#diff-bc30a9362a318408471b5a8f16ce5b76R31).
The SSE is then the [min of these sizes](https://github.com/iTowns/itowns/pull/674/files#diff-bc30a9362a318408471b5a8f16ce5b76R39).

Subdivision happens [if SSE](https://github.com/iTowns/itowns/pull/674/files#diff-989ef5eaaef422d31895b5f0e7eb2c7aR106) is > 256 (texture size) + layer.sseThreshold (5 by default).

The reason here for using the min is to prevent subdivision when the camera view vector is ~tangent to the surface but I agree that it's problematic in other case, such as your 2nd screenshot.

An alternative would be to keep SSE in x and y directions separately.
Then:
   - subdivide if min(sse.x, sse.y) > 256 or
   - if max(sse.x, sse.y) > 256 and min(sse.x, sse.y) > 128 (or another value that works)

peppsac:
@gchoqueux I pushed an example implementation of https://github.com/iTowns/itowns/pull/674#issuecomment-370891414

It breaks everything but the globe (because I only adapted GlobeTileProcessing). Can you give it a try and let me know what you think?

gchoqueux:
in `computeVectorSizeAtDistance` :
 * The box's diagonal may also be the largest projection?  This approach doesn't take it into account?
 * Could you use [Vector3.Project(camera)](https://github.com/mrdoob/three.js/blob/master/src/math/Vector3.js#L315)?
* the scale transformation isn't in the calculation?

peppsac:
> The box's diagonal may also be the largest projection? This approach doesn't take it into account?

This approach computes the sse for each axis.
For tiles, texture are mapped on the x/y axis, so I don't think it make sense to use the x/y diag.
For points, we could use the diagonal instead because the points are distributed in all directions in the bbox, but having the 3 sse values is enough it seems.

> Could you use Vector3.Project(camera)?

Nope. The `Vector3.Project` projects a world space vector, while I'm computing a [camera space vector](https://github.com/iTowns/itowns/pull/674/files#diff-bc30a9362a318408471b5a8f16ce5b76R32).
This vector is then moved at `distance` from the camera and then  projected on screen.

> the scale transformation isn't in the calculation?

It's used to compute the distance to camera.
For the size to projectd: we only the ratio of the box so the scale isn't relevant (because we scale the bbox to have its largest side being `geometricError`)

gchoqueux:
> so I don't think it make sense to use the x/y diag

I'm talking about the box's diagonal (in x, y and z)

> For tiles, texture are mapped on the x/y axis

DEM is mapped on z axis, not only x and y

>  while I'm computing a camera space vector.

yet you go through the world space when calculating (with `m3`)
the projection would be different with Vector3.Project?

peppsac:
> DEM is mapped on z axis, not only x and y

Can you compute a correct estimation of the DEM error in meters along the z axis for every tiles?
If you have a formula then yes we can use it as well.
Note that the current formula doesn't use the tile vertices count: it could make sense to use too but again, this needs a correct formula that works on all tiles.

> yet you go through the world space when calculating (with m3)

No I'm not. You misread the code: m3 is only the rotation part, not the full world matrix of the camera.

gchoqueux:
What is your formula to estimate error on the x and y axes?

You don't answer this question : the projection would be different with Vector3.Project?

peppsac:
> What is your formula to estimate error on the x and y axes?

Compare size in pixels of the tile on the screen to the texture size.

> You don't answer this question : the projection would be different with Vector3.Project?

Yes it would be diffrent. I detailled in my earlier comment https://github.com/iTowns/itowns/pull/674#issuecomment-371803340 the differences between Vector3.Project and my computations.

gchoqueux:
The texels are also on the box's diagonal (x, y, z), it can more revelant by example with box'size (100, 100, 1000)


> Yes it would be diffrent

I can't understand why there is a difference between the two, I'm going to look

peppsac:
Rebased. I also removed the `MODE_2D` / `MODE_3D` thing since it's not used anymore.

As far as I can tell this PR improves the SSE handling in most situations.

Regarding the diagonal: I still believe it isn't an issue. It's an corner case so I don't think it's worth complexifying the code to cover it. Btw, current master version SSE handling is quite bizarre anyway. E.g: in this screenshot the nearest tile in the middle of the screen is almost 600px wide.
 
sse_bug_example
You can try with:
 ```js
globeView.camera.camera3D.position.copy({x: 4490626.907379815, y: 506551.71291476686, z: 4500105.469154227})
globeView.camera.camera3D.rotation.copy({_x: -1.3092800928909716, _y: 0.4734022844137177, _z: -2.4859644421245957, _order: "XYZ"})
globeView.notifyChange(true);
```

File: src/Core/ScreenSpaceError.js [hunk 11]
```diff
@@ -0,0 +1,96 @@
...
+    return basis.map(b => b.length());
+}
+
+function findBox3Distance(camera, box3, matrix) {
+    // TODO: can be cached
+    m.getInverse(matrix);
+    // Move camera position in box3 basis
+    // (we don't transform box3 to camera basis because box3 are AABB)
+    const pt = camera.camera3D.position
+        .clone(v).applyMatrix4(m);
```
mbredif, line 49: this is wrong if m as some scaling (eigenvalues != 1)

src/Core/ScreenSpaceError.js
```diff
@@ -0,0 +1,96 @@
...
+        b.z += -distance;
+        // Project on screen
+        b.applyMatrix4(camera.camera3D.projectionMatrix);
+        // cancel z component
+        b.z = 0;
+        b.x = b.x * camera.width * 0.5;
+        b.y = b.y * camera.height * 0.5;
+    }
+
+    return basis.map(b => b.length());
```
mbredif, line 40: this looks like an approximation of some distance, which computation is optimized by considering axes independently. could you phrase mathematically the exact distance you are approximating ?

mbredif (commented):
some comments. could you define mathematically the sse?

peppsac:
> some comments. could you define mathematically the sse?

Thanks for the comments.

I'll come back to this PR soon, but for now I'm focused on the more recently opened PRs.

PR #675, opened 2018-03-01
feat(protocols): add spatial hashing to StaticProvider

peppsac:
Using https://github.com/mourner/flatbush speeds up the look up by a factor of 10.

For instance, on my computer and the panorama demo:
  - before the PR selectBestImageForExtent used to take 0.5ms to 3ms
  - after the PR: <= 0.1ms

### DIFF
--- a/src/Core/Scheduler/Providers/StaticProvider.js
+++ b/src/Core/Scheduler/Providers/StaticProvider.js
@@ -1,13 +1,24 @@
+import * as FlatBush from 'flatbush';
 import { Vector4 } from 'three';
 import Extent from '../../Geographic/Extent';
 import OGCWebServiceHelper from './OGCWebServiceHelper';
 import Fetcher from './Fetcher';
 import { l_COLOR, l_ELEVATION } from '../../../Renderer/LayeredMaterialConstants';
 
+function _selectImagesFromSpatialIndex(index, images, extent) {
+    return index.search(
+            extent.west(), extent.south(),
+            extent.east(), extent.north()).map(i => images[i]);
+}
+
 // select the smallest image entirely covering the tile
-function selectBestImageForExtent(images, extent) {
+function selectBestImageForExtent(layer, extent) {
+    const candidates = (layer.extent.crs() == extent.crs()) ?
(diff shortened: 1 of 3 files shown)


src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -104,6 +115,15 @@ export default {
                     extent,
                 });
             }
+            layer._spatialIndex = FlatBush(layer.images.length);
+            for (const image of layer.images) {
+                layer._spatialIndex.add(
```
autra, line 120: You could merge this loop with the one above. Or did I miss something?

src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -128,20 +148,25 @@ export default {
             return false;
         }
 
-        for (const entry of layer.images) {
-            if (tile.extent.isInside(entry.extent)) {
-                return true;
+        if (tile.extent.crs() == layer.extent.crs()) {
```
autra: providing that Extent.isInside starts by doing `other.as(this._crs)`, why can't we use the spatial index in all cases by doing a `tile.extent.as(layer.extent.crs())` first?

autra (changes requested):

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: I need to know the number of images to initialize the index.
And `metadata` is a json object, so I'm not sure how to get the number of images before the loop?

src/Core/Scheduler/Providers/StaticProvider.js
peppsac: Ok, I'll change that.

src/Core/Scheduler/Providers/StaticProvider.js
autra: Oh right, it's a `for..in`, not a `for..of` loop

autra (approved):

autra:
LGTM thanks!

PR #676, opened 2018-03-01
Fix obb 3d tiles

### LINKED ISSUE #656: Wrong calculation 3dtiles's OBB
## Your Environment
all browsers and all os 

## Context
Wrong position/size for 3dtiles's OBB

possible cause : `OBB.extentToOBB` and `OBB.cardinalsXYZToOBB`

gchoqueux:
## Description
fix wrong 3dtiles's OBB using BuilderEllipsoidTile fix #656
fix wrong cachekey value equal to NaN in geometry tile building

obb

### DIFF
--- a/src/Renderer/ThreeExtended/OBB.js
+++ b/src/Renderer/ThreeExtended/OBB.js
@@ -1,5 +1,7 @@
 import * as THREE from 'three';
 import Coordinates, { C, UNIT } from '../../Core/Geographic/Coordinates';
+import TileGeometry from '../../Core/TileGeometry';
+import BuilderEllipsoidTile from '../../Core/Prefab/Globe/BuilderEllipsoidTile';
 
 function OBB(min, max) {
     THREE.Object3D.call(this);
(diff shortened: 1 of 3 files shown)


src/Renderer/ThreeExtended/OBB.js
```diff
@@ -103,119 +105,28 @@ for (let i = 0; i < 9; i++) {
 }
 
 // get oriented bounding box of tile
-OBB.extentToOBB = function extentToOBB(extent, minHeight = 0, maxHeight = 0) {
+const builder = new BuilderEllipsoidTile();
+OBB.extent_R4326_ToOBB_4978 = function _extent_R4326_ToOBB_4978(extent, minHeight = 0, maxHeight = 0) {
```
peppsac: You don't need to rename the function; but you need to add a check at the beginning to validate `extent._internalStorageUnit`

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -103,119 +105,28 @@ for (let i = 0; i < 9; i++) {
...
-    obb.lookAt(tmp.normal);
-    obb.translateX(tmp.translate.x);
-    obb.translateY(tmp.translate.y);
-    obb.translateZ(tmp.translate.z);
-
-    // for 3D
-    if (minHeight !== 0 || maxHeight !== 0) {
-        obb.translateZ(-halfMaxHeight + minHeight + nHalfSize);
-    }
+    const { sharableExtent, quaternion, position } = builder.computeSharableExtent(extent);
```
peppsac: Building a new tile to then build the OBB seems a bit overkill but at least it's a major code simplification, so +1

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -103,119 +105,28 @@ for (let i = 0; i < 9; i++) {
...
+        segment: 2,
+        disableSkirt: true,
+    };
+
+    const geometry = new TileGeometry(paramsGeometry, builder);
+    const obb = geometry.OBB;
+    obb.updateZ(minHeight, maxHeight);
+    obb.position.copy(position);
+    obb.quaternion.copy(quaternion);
+    obb.update();
```
peppsac, line 130: `geometry.dispose()` must be called before leaving the function to allow threejs to release the associated memory.

peppsac (commented):
Looks good overall, just a few minor comments to fix before merging.

src/Renderer/ThreeExtended/OBB.js
gchoqueux: I'm not sure because geometry was not pushed in rendering

src/Renderer/ThreeExtended/OBB.js

gchoqueux: done

src/Renderer/ThreeExtended/OBB.js
peppsac: Yes it's useless for now.
But it doesn't hurt to add it in case threejs change its behavior and begin reserving some memory at creation time.

(or don't add it but add a comment stating something like "Calling geometry.dispose() is not needed since this geometry never gets rendered")

gchoqueux: comment added

peppsac (approved):
LGTM (don't forget to cleanup your commits before merging)

PR #677, opened 2018-03-02
feat(panorama): add projection type constant

peppsac:
Add explicit projection type constants instead of relying on the ratio to deduce
the projection.


zarov (approved):
LGTM, way cleaner than the `equirectangular` and `ratio` properties, thanks !

peppsac:
Thanks for the approval. Can you wait a sec before merging, there's something I need to check before :)

peppsac:
I'll push a commit to fix a breakage from 9241818aadc21cd0ad71b73d6b4013b922c3da40: before this commit the panorama was oriented such as North = center of the image.

PR #678, opened 2018-03-02
fix(protocols): fix wrong orientation in panorama

gchoqueux:
## Description
panorama's Y axis (north) look at the image pano's center
and by default pano's camera look at the same north


peppsac (approved):
LGTM, thanks!

PR #679, opened 2018-03-05
Improve panorama support

peppsac:
PR with small fixes and addition, to improve panorama support in iTowns.

3 commits content:
```
feat(protocols): implement targetLevel for StaticProvider
fix(panorama): improve default settings for Panorama
fix(core): assume min level is 0 if not provided for dichotomy strategy
```

### DIFF
--- a/src/Core/Scheduler/Providers/StaticProvider.js
+++ b/src/Core/Scheduler/Providers/StaticProvider.js
@@ -39,16 +39,23 @@ function buildUrl(layer, image) {
         + image;
 }
 
-function getTexture(tile, layer) {
-    if (!layer.tileInsideLimit(tile, layer)) {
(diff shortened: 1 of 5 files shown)


src/Core/Scheduler/Providers/StaticProvider.js
```diff
@@ -28,10 +28,7 @@ function buildUrl(layer, image) {
         + image;
 }
 
-function getTexture(tile, layer) {
-    if (!layer.tileInsideLimit(tile, layer)) {
```
autra, line 43: Why don't you need that any more?

peppsac: Because it's doing the same work twice: `tileInsideLimit` is implemented by lookup the texture to use. So we can skip `tileInsideLimit` and directly do the texture lookup (a few line below).

autra (approved):
Thanks!

PR #680, opened 2018-03-05
refactor(core): add Provider type constants

zarov:
Move strings used to dissociate providers to constants. Strings are kept, ensuring retrocompatibility.


peppsac:
I'm not sure this PR is an improvement, especially since we still have a confusion bewteen protocols and file format/provider.

eg:
* using Provider.$3DTILES to declare a protocol string seems a bit strange ?
(`$3dTilesLayerDiscreteLOD.protocol = itowns.Provider.$3DTILES`)
* RASTER is not the protocol here but it's the format in which we're going to tranform the vector file
(`protocol: itowns.Provider.RASTER`). Maybe `AS_RASTER` would be better here.

zarov:
Yes I agree, it may be better to address this issue later, when confusion has been lifted.

PR #681, opened 2018-03-05
fix(provider): Allow WFS layer to have any format

autra:
Followup on #597 : it broke the wfs examples. 

Actually I was wrong for WFS: the spec allows any string in their format parameter. Therefore, I can't restrict layer.format to a mimetype. So I reverted the modification in the examples, and as a quick fix I removed the check on WFS layers. 

A more exact check could be implemented (properly catching the Fetcher.json results maybe?) to have better error messages.

### DIFF
--- a/examples/globe_wfs_extruded.js
+++ b/examples/globe_wfs_extruded.js
@@ -68,6 +68,7 @@ function acceptFeatureBus(properties) {
 }
 
 globeView.addLayer({
+    name: 'lyon_tcl_bus',
     type: 'geometry',
     update: itowns.FeatureProcessing.update,
     convert: itowns.Feature2Mesh.convert({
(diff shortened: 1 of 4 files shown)


autra:
@iTowns/reviewers updated: the first commit fixes the examples with lyon data, the second tries to be a bit more friendly with errors. Thanks!

zarov (approved):
Thanks for the fix, sorry to have missed this problem when reviewing ! As for the error handling, that could be changed to a more generic one, reusable by other parts in the code. But priority to fixing the live examples ;)

autra:
> that could be changed to a more generic one, reusable by other parts in the code

Yes possibly, also because the OGC tries to be consistent with the error formats of their protocol. But that could be factorized later

PR #682, opened 2018-03-06
chore: change the no-console property of eslint

zarov:
Change the no-console property to a more permissive one, allowing
console.error and console.warn statement, and removing the annotation in
some file about already present statement.


peppsac (approved):
LGTM, thanks

PR #683, opened 2018-03-06
feat(core): move url function outside providers

zarov:
## Description
url functions in some provider (tms, wfs, wms and wmts) share quite the same content. Following the work in #182, these functions can be put outside the providers without problem.

Full documentation has been added, for easier usage outside the core, and unit testing as well.

## Motivation and Context
This comes in the work of #182, part of rewriting the Providers/Loader/etc system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib/Main.js",
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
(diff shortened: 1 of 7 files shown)


zarov:
Suggestion: if it stays in the `Providers` folder, we can maybe just name it `URL.js`.

peppsac:
Or `OGCProtocolsURLBuilder.js`? It's a bit longer but maybe more descriptive... I'm not sure I like it :)

The rest of the commit looks good.

zarov:
`OGCProtocolsURLBuilder` makes sense, as it is building urls used by OGC protocols. But just `URLBuilder` is ok for me.

peppsac:
Let's go with `URLBuilder` since I'm not sure that `xyz` is really an OGC protocol

zarov:
Done

peppsac (approved):
LGTM thanks!

zarov:
Thanks !

PR #685, opened 2018-03-07
Babel webpack refacto

autra:
## Description
Get rid of babel-plugin-webpack-loader

## Motivation and Context

I didn't like the `babel-plugin-webpack-loader` approach I did in #248. It was slow and feels fragile to me. Also there was too much magic involved.

I believe this new version makes things a bit more streamlined and clear.

I included 2 other commits in the same topic:
- one that updates all babel dependencies to minor, except the 2 minify plugins that go from 0.2 to 0.3 (so considered as major). I've checked the changelogs, nothing to worry about a priori.
- one that removes some unused dependencies (I believe).


package.json
```diff
@@ -8,10 +8,10 @@
     "doc": "jsdoc src/Core/View.js src/Core/Prefab/GlobeView.js src/Core/Layer/Layer.js src/Renderer/ColorLayersOrdering.js src/Renderer/ThreeExtended/GlobeControls.js src/Core/Geographic/Coordinates.js src/Core/Scheduler/Providers/GpxUtils.js src/Core/MainLoop.js -t node_modules/docdash --readme JSDOC.md -c jsdoc-config.json",
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-examples && npm run test-unit",
-    "test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --compilers js:babel-core/register test/*unit_test.js",
```
tbroyer: babel-plugin-inline-import's README says this should still be used; and with webpack that babel-inline-import-loader could be used instead: https://www.npmjs.com/package/babel-plugin-inline-import#caveats

tbroyer (commented):
I'm very supportive of this change, but not sure about the details here wrt babel-plugin-inline-import documented caveats.

package.json
autra: Oh right, thanks for pointing that out. 

> babel-inline-import-loader could be used instead

From the doc of the loader, I understand that this loader actually needs babel-plugin-inline import (so both should be added). Did I understand well?

tbroyer: > From the doc of the loader, I understand that this loader actually needs babel-plugin-inline import (so both should be added). Did I understand well?

Ah yes, you're right. The loader for webpack builds upon the plugin for babel.

zarov (approved):
LGTM, could you rebase before merge ? Thanks !

autra:
Actually, this PR does not work perfectly yet. There's something wrong with __DEBUG__ var in one case. I need to further investigate.

zarov:
Oh ok, what is the use case ?

zarov:
Any update ?

autra:
@zarov not yet... :-/

autra:
@zarov ok, I fixed my bug :-) For some reason `json.stringify` is not needed anymore [here](https://github.com/iTowns/itowns/pull/685/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR5). Actually if I use it, babel complains I gave him a string instead of a boolean.

Could you review again please?

zarov (approved):
Changes are good, if you really really want a review, I would change in the first commit message `execution time` to `launch time` ;) But that's good as it is !

autra:
> Changes are good, if you really really want a review, I would change in the first commit message execution time to launch time ;) But that's good as it is !

I'm speaking about the execution time here though... Ok merging that.

PR #686, opened 2018-03-07
Fix(View): fix event vs mouse detection when picking

autra:
mouseEvent.x actually exists in Firefox and IE, see:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/x

This fixes picking in Firefox


peppsac (approved):
Thanks!

PR #687, opened 2018-03-08
refactor(core) remove CacheRessource

zarov:
## Description
Remove the unecessary `CacheRessource` class, in favour of a `Map`. The `Cache` file has two maps:
- pending is for pending ressources (being fetched)
- cache is for cached ressources

## Motivation and Context
I feel that this `CacheRessource` file is obsolete and that we should use `Map` instead. I'm not sure how it could be implemented though: 
- could `cache` and `pending` be in `Fetcher`, and the cache process integrating in each fetch command ? Or without doing anything, only as `export const cache = new Map()` ?
- could they be in a "higher" part, like `Scheduler` or somewhere else ?
- are they good as it is now ?
I'm undecided here, so please bring your thoughts ! I implemented a quick proposal to this.


autra:
Thanks for digging this. http and javascript have already *plenty* of mechanisms for cache and offline support, that'll be cool if we manage to use these! And  if we do, it'll also work across page refresh...

For example, we could use [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) and [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker). You even have a basic example in the Cache doc linked above. The only problem of this approach is browser support: service workers won't work in IE, Safari, chrome < 40, and Firefox < 44. This only means the cache feature won't work though, the rest of the application will still work ok.WDYT about having partial support for cache for modern browsers only? (I'll be ok with that btw :-) )

If it is not acceptable, then I would advocate to find a polyfill/lib for `Cache` (instead of coding our own) and use that on the main thread instead of inside a Service Workers. 

Of course, the best would be to rely on http server cache configuration all the time, but often they are not configured aggressively enough to avoid even 304 network round trips. In our case (with the high number of requests we do), even those are quite costly because we still get network latency.

zarov:
No problem ! 

Concerning the support, Chrome 40 is 3 years old, and Firefox 44 is 2 years old. I think it is enough, considering IE won't ever support it, and in Safari it should come soon ([see here](https://webkit.org/blog/8042/release-notes-for-safari-technology-preview-46/), and I think Safari is updated quite enough to be ok to move to this). I agree we could add partial support first.

I'm trying it then !

> Of course, the best would be to rely on http server cache configuration all the time

As well as for `format/mimetype`, but in a ideal world :D

zarov:
So I learned more that I thought on the subject, and after trying multiple solutions I think we should stay for now with a simple `Map`. 

- Service Worker:
    - it needs a file beside iTowns
    - it needs to be in `https`
    - it is difficult to do a filter of what we keep in cache: Requests are not that detailed, and there are the only way to transmit information to the worker
- Cache (In case it is used alone, without a Service Worker):
     - it is async, meaning necessary rewrite with `Promise` of some files (OGCWebServiceHelper, etc...)
    - it can only store `Response`, so it needs to be added in `Fetcher`
    - and thus, only `Response` can be obtained from this


Overall I would say that using those things could be interesting if we want to add a support for offline maps. This could be useful when demonstrating some cases without connection to internet. A Service Worker could then be introduced, and the user could opt to include it in its page to activate it or not.

But for now let's rely on the browser cache, that we can force when making request ([see here](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache)). I propose to stay with only a more reliable cache than the current `CacheRessource` (changing it to a `Map`).

autra:
Thanks for these researches!

> it is difficult to do a filter of what we keep in cache: Requests are not that detailed, and there are the only way to transmit information to the worker

You can also use `postMessage` between workers and the main thread though.

> But for now let's rely on the browser cache, that we can force when making request (see here).

How so? request.cache is readonly.

`CacheResource` could be interesting to implement a cache expiration mechanism imho. If we don't, then we're better off with a map, that's right. I'm wondering if this cache logic wouldn't be better at the Scheduler level? It's the same for all providers right?

zarov:
> You can also use `postMessage` between workers and the main thread though.

Oh I missed that :( But this unfortunately doesn't solve the problem, as it is asynchronous from requesting a resource, and then I don't see how we could tell the Worker how to determine if this resource is to keep when it is requested.

> How so? `request.cache` is readonly.

Yes, but you can add options to `fetch` about that. But it can't be done for resources using the `Fetcher.texture` method, as it doesn't use `fetch`, so we would need to drop the `TextureLoader` if we go that way.

> It's the same for all providers right?

Currently, only `WFS_Provider`, `TileProvider` and providers using `OGCWebServiceHelper` are concerned with caching, so not all (like `3dTilesProvider` for example). Moving the cache in/around the `Scheduler` could indeed be a good thing to centralize a bit more things.

autra:
I don't think we'll be able to avoid the asynchronism permanently here. But I don't see the problem, as `executeCommand` (and most of the methods involved here reallly) is already asynchronous. It's necessary anyway, as we need to populate the map with still in progress requests. It doesn't really change the code overall right?

As for the cache option of `fetch`, thanks for the explanation! It is indeed quite limited: basically either you ignore the cache config of the server, or you don't ignore it. Actually, reading [this](https://stackoverflow.com/questions/31601167/fetch-api-cache-mode) I think the default is actually the best, while keeping our own cache mechanism to avoid requesting the same tiles all the time if the server is misconfigured... 

So long story short, your PR is basically equivalent to the current code and I feel it is in the general direction we want to take, while doing things in a cleaner and simpler way (I didn't like the ` this.cacheObjects = [];` and `this.cacheObjects[url]` (meh!) of CacheResource :-)) Also, your code probably performs better, as it rely on Map, a builtin objects. 

So r+ on this, we can merge this now imo.

I guess future work would include:
- Provide a cache expiration mechanism, with a configurable duration on a per layer basis? 0 could mean no cache and -1 could mean "cache infinitely".
- probably move this behaviour in Scheduler, or at least in a common part, to make every layer benefits of this
- Maybe use localstorage or something similar to have part of the cache persists across session?

This is an area where iTowns could greatly improve!

autra (approved):

tbroyer (approved):

PR #689, opened 2018-03-09
fix(panorama): change flatbush version and import line

peppsac:
flatbush applied a [breaking change](https://github.com/mourner/flatbush/commit/7cc362a3e4d26293a559353dbe7e92f53fd47f62) in a minor version.
So this commits forces upgrade to the new version and adapts the import line.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2987,9 +2987,9 @@
       }
     },
     "flatbush": {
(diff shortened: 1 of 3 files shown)


autra:
odd...

autra (approved):

PR #690, opened 2018-03-12
Fix coordinates conversions

gliegard:
Fix coordinate conversion :
- First bug is about the test of internalStorageUnit and refUnit.
- Second bug is that we forget to pass the 'target' parameter to the 'as' method, so target ouput parameter wasn't changed (before this commit).


peppsac:
Thanks for the PR. 
IMHO this internalUnitStorage thing was a bad idea to begin with, so I've started a branch to remove it completely: https://github.com/peppsac/itowns2/tree/remove_internalstorageunit

peppsac (approved):

PR #691, opened 2018-03-13
refactor(vector): rework feature support

### LINKED ISSUE #688: (big) multipolygon does not seem to be displayed correctly
A big multipolygon is not rendered correctly.

It is originally served from a  wfs  from geoserver, but I'm providing a geojson file of the multipolygon.
[multipolygon.zip](https://github.com/iTowns/itowns/files/1794114/multipolygon.zip)

It is correctly displayed in Qgis as shown in the screen captures below
**itowns**
multipolygon_itowns

**Qgis**
multipolygon_qgis

## Environment

* Version used: "itowns": "^2.3.0"
* Browser Name and version: firefox 58.0.2 /chromium 64.0
* Operating System and version (desktop or mobile): Linux Ubuntu 16.04

peppsac:
This commits refactors Feature related code.

It introduces a documented Feature object.
GeoJSON2Feature returns either a Feature or an array of Feature (when parsing
a FeatureCollection object).

Geometry merges now only happens for multi-* geometries to simplify user
handling of individual features.

Geometry holes are now supported:
capture d ecran de 2018-03-13 13-52-32

And last but not least this commit fixes the geometry merge bug from issue 688.

Fixes #688. This is a screenshot of the example geojson, displayed both as a texture (in yellow) and as a mesh (blue wireframe):

capture d ecran de 2018-03-13 13-54-23

BREAKING CHANGE: feature collection are not merged anymore
BREAKING CHANGE: GeoJSON2Feature now returns a Promise

### DIFF
--- a/src/Renderer/ThreeExtended/Feature2Mesh.js
+++ b/src/Renderer/ThreeExtended/Feature2Mesh.js
@@ -40,7 +40,7 @@ function getColor(options, properties) {
     return randomColor();
 }
 
-function fillColorArray(colors, offset, length, r, g, b) {
+function fillColorArray(colors, length, r, g, b, offset) {
     const len = offset + length;
     for (let i = offset; i < len; ++i) {
         colors[3 * i] = r;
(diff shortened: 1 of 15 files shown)


peppsac:
Notes:
  - this PR probably makes PR #673 useless.
  - this PR allows to color features individually (since a feature = a mesh)

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
+     * have an extent property containing the area covered by the geom
+     * @param {function} options.filter - Filter function to remove features
+     * @returns {Promise} - a Promise resolving with a Feature or an array of Feature
+     */
     parse(crsOut, json, filteringExtent, options = {}) {
         options.crsIn = options.crsIn || readCRS(json);
         switch (json.type.toLowerCase()) {
             case 'featurecollection':
-                return readFeatureCollection(options.crsIn, crsOut, json, filteringExtent, options);
+                return Promise.resolve(readFeatureCollection(options.crsIn, crsOut, json, filteringExtent, options));
```
autra, line 274: why do you need a `Promise.resolve`?

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
+        return collec;
     }
 }
 
 export default {
+    /**
+     * @typedef FeatureGeometry
+     * @type {object}
+     * @property {string} type - Geometry type ('(multi)point', '(multi)linestring' or '(multi)polygon')
+     * @property {Coordinate[]} vertices - All the vertices of the geometry.
```
autra: nits `Coordinates`

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
+     * @property {?number[]} contour - If this geometry is a polygon, contour contains
+     * the indices that compose the contour (outer ring)
+     * @property {?Array} holes - If this geometry is a polygon, holes contains
+     * an array of indices reprensting holes in the polygon.
```
autra: `representing`

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
+     * @property {?Extent} extent - The 2D extent containing all the geometry
```
autra: `all the geometries` or `FeatureGeometry`

autra: Or all the `FeatureGeometry`

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
+     * @param {string} crsOut - The CRS to convert the input coordinates to
+     * @param {string} json - The GeoJSON file to parse
+     * @param {?Extent} filteringExtent - Optional filter to reject features
+     * outside of this extent.
+     * @param {object} options - options controlling the parsing
+     * @param {string} options.crsIn - override the data crs
+     * @param {boolean} options.buildExtent - if true the geometry will
+     * have an extent property containing the area covered by the geom
+     * @param {function} options.filter - Filter function to remove features
+     * @returns {Promise} - a Promise resolving with a Feature or an array of Feature
```
autra, line 268: See below, I don't see why you need a promise here. Or you want to prepare the way for workers?

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
-        if (geom.polygons.length) {
-            concatGeometries(result, geom.polygons);
-        }
-        // remember individual features properties
-        // eslint-disable-next-line arrow-body-style
-        result.features = collec.map((c) => { return { properties: c.properties }; });
-        if (result.geometries.length) {
-            return result;
-        }
+        return collec;
```
autra: Why not returning `collec` everytime? At worst it'd be an empty array, maybe nicer to work with from the caller perspective...

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -179,67 +209,71 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
...
-
 function readFeatureCollection(crsIn, crsOut, json, filteringExtent, options) {
     const collec = [];
 
     let featureIndex = 0;
     for (const feature of json.features) {
         const f = readFeature(crsIn, crsOut, feature, filteringExtent, options);
         if (f) {
+            if (options.buildExtent) {
+                if (!collec.extent) {
```
autra: (nits: I prefer the positive style, if possible)

src/Renderer/ThreeExtended/GeoJSON2Features.js
```diff
@@ -60,78 +66,102 @@ const GeometryToCoordinates = {
...
-            } else {
-                // merge coordinates
-                for (const coordinate of geom.coordinates) {
-                    result.coordinates.push(coordinate);
-                }
-                // union extent if present
-                if (geom.extent) {
-                    result.extent.union(geom.extent);
-                }
+            // only test the first line
```
autra: copy-paste leftover :-)

autra: Not sure it's still true for `multilineString` below anyway....

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -350,7 +410,8 @@ export default {
     convert(options = {}) {
         return function _convert(feature) {
             if (!feature) return;
-            if (feature.geometries) {
+
+            if (feature instanceof Array) {
```
autra: Better to use `Array.isArray` (see [this](http://web.mit.edu/jwalden/www/isArray.html) for more details). Here, this *should* work for our internal use case, but user can import Feature2Mesh directly so...
(also, babel/webpack heavily modifies `Array.prototype`, and I'm not sure this'll always work in all situations considering our stack or if we modify babel conf).

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -294,26 +325,57 @@ function coordinateToPolygonExtruded(coordinates, properties, options) {
  * param  {callbacks} callbacks defines functions to read altitude and extrude amout from feature properties
  * return {THREE.Mesh} mesh
  */
-function coordinatesToMesh(coordinates, properties, options) {
-    if (!coordinates) {
+function geometryToMesh(geometry, properties, options) {
```
autra: The jsdoc should be updated (apparently it hasn't been for a while).

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -350,7 +410,8 @@ export default {
     convert(options = {}) {
```
autra, line 431: could you add a doc on this function while you're at it?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -294,26 +325,57 @@ function coordinateToPolygonExtruded(coordinates, properties, options) {
...
+
+
     var mesh;
-    switch (coordinates.type) {
-        case 'point': {
-            mesh = coordinateToPoints(coordinates, properties, options);
+    switch (geometry.type) {
+        case 'point':
+        case 'multipoint': {
+            mesh = geometryToPoint(geometry, properties, options, multiGeometries);
```
autra, line 365: I'm wondering if it won't be simpler to have `geometryToMulti*` functions here. They don't share that much of common code.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,230 +57,261 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
-        coordinatesToVertices(contour, altitude, vertices, offset * 3);
-
-        // assign color to each point
-        const color = getColor(options, property);
-        fillColorArray(colors, offset, contour.length, color.r * 255, color.g * 255, color.b * 255);
-
-        // increment offset
-        offset += contour.length;
+function prepareBufferGeometry(vert, altitude, color) {
+    const multiplyVerticesCount = altitude instanceof Array ? altitude.length : 1;
```
autra: `Array.isArray`

autra: I don't get this. You can have several altitude per Coordinates? Why is the altitude changing the length of vertices? (I missed something somewhere)

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,230 +57,261 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
+        altitude,
+        color);
+
+    if (!multiGeomAttributes) {
+        return new THREE.Line(geom);
+    } else {
+        const indices = [];
+        // Multi line case
+        for (let i = 0; i < geometry.length; i++) {
+            const start = multiGeomAttributes[i].offset + geometry[i].contour.offset;
```
autra: There's no geometry.contour for lines right?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,230 +57,261 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
-        const altitudeTopFace = altitudeBottom;
-        // add vertices of the top face
-        coordinatesToVertices(contour, altitudeTopFace, vertices, offset * 3);
-        const verticesTopFace = vertices.slice(offset * 3, offset * 3 + contour.length * 3);
-        // triangulate the top face
-        const triangles = Earcut(verticesTopFace, null, 3);
+    // Build indices
+    if (!multiGeomAttributes) {
+        // Single polygon case
+        const holes = geometry.holes.map(h => h.offset);
```
autra: `holes` -> `holesOffsets` ?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,230 +57,261 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
+        const triangles = Earcut(geom.attributes.position.array, holes, 3);
         for (const indice of triangles) {
-            indices.push(offset + indice);
+            indices.push(indice);
+        }
+    } else {
+        // Multi polygon case
+        for (let i = 0; i < geometry.length; i++) {
+            const holes = geometry[i].holes.map(h => h.offset);
+            const start = multiGeomAttributes[i].offset + geometry[i].contour.offset;
```
autra: `geometry[i].contour.offset` is always 0 with the current code right? Added here for the sake of consistency/robustness in case the contour is not at the beginning of the array?

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: Oh ok, it's because altitude can contains the extrude information. So actually `altitude.length` is < 2 in all cases. Maybe consider having a `extrude` param instead? If we hardcode the `extrude` logic, we could do it entirely. WDYT?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,230 +57,261 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
+
+    // get altitude / color from properties
+    const altitudes = [
+        getAltitude(options, properties, vertices),
+        getExtrude(options, properties),
+    ];
+    const colors = [
+        getColor(options, properties)];
+    colors.push(colors[0].clone());
+    colors[0].multiplyScalar(155 / 255);
```
autra: If I understand this correctly, you have a slightly different color for the extruded face, is that it? If so why don't we rely on lights and materials for this?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,230 +57,261 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
+                    vertices.length,
+                    geometry.holes[j],
+                    isClockWise);
+            }
+        }
+    } else {
+        // Multi polygon case
+        const isClockWise = THREE.ShapeUtils.isClockWise(
+            vertices.slice(
+                multiGeomAttributes[0].offset + geometry[0].contour.offset,
```
autra: So we assume that other geometries will have the same orientation, right? Is it safe to assume according to you?

test/featureUtils_unit_test.js
```diff
@@ -6,34 +6,49 @@ import Coordinates from '../src/Core/Geographic/Coordinates';
...
 const geojson = require('../examples/geojson/simple.geojson.json');
 
-const feature = GeoJSON2Features.parse('EPSG:4326', geojson, undefined, { buildExtent: true });
+const promise = GeoJSON2Features.parse('EPSG:4326', geojson, undefined, { buildExtent: true });
 
 describe('FeaturesUtils', function () {
-    it('should correctly parse geojson', () => {
-        assert.equal(feature.geometries.length, 3);
+    it('should correctly parse geojson', (done) => {
+        promise.then((feature) => {
```
autra: With newer version of `mocha`, you can directly return a promise, this will have the same effect than the `done` mechanism.

autra: But actually, GeoJSON2Features could return a plain object, see above.

src/Renderer/ThreeExtended/GeoJSON2Features.js
autra: (note: if it's because of the providers, they don't need it, they already wrap stuff in promises)

autra (changes requested):
Nice!! I do have some nits and questions, but overall LGTM already. Thanks for this work :-)

autra:
> this PR allows to color features individually (since a feature = a mesh)

I actually do have this need currently. Also, I feel that a feature = a mesh is more aligned with the spirit of the GeoJSON spec.

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: No. It's simply to expose an async API.
(and then later, it opens the possibility to use WebWorker to parse the geojson, without changing again the API: see https://github.com/peppsac/itowns2/commit/729255b16cdf3d9e47745c3b750b997739c2ceeb#diff-df7974827a71f76f18b31a30dbc160b2R79 for an quick & dirty implementation)

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: Fixed.

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: Fixed.

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: Fixed.

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: Ok, I modified the code to always return `collec`.

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: Fixed.

src/Renderer/ThreeExtended/GeoJSON2Features.js
peppsac: Fixed.

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: Thanks, I didn't know that `instanceof Array` had weird corner cases. Fixed.

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: Fixed.

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: Right, fixed. Thanks

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: Yep

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: This comes from the original extrude code and I didn't want to change it.

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: Yeah that's the assumption and I feel it's safe (ie: I assume that a server will always return geometries with the same orientation).

test/featureUtils_unit_test.js
peppsac: > With newer version of mocha

Does our version of `mocha` support this?

peppsac:
Thanks for the review @autra 
I updated the PR and addressed all your comments.

test/featureUtils_unit_test.js
autra: Yes, I did it for the `getCapabilities` PR

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: Looks like a hack to me. I'd vote to remove this (but that can be in another PR).

autra:
So I have some errors in `globe_wfs_extruded`:
```
TypeError: itowns.DEMUtils.getElevationValueAt(...) is undefined
Stack trace:
altitudePoint@http://localhost:8080/examples/globe_wfs_extruded.js:178:16
getAltitude@webpack-internal:///283:13:20
geometryToPoint@webpack-internal:///283:154:20
geometryToMesh@webpack-internal:///283:423:24
featureToThree@webpack-internal:///283:452:16
featureCollectionToThree@webpack-internal:///283:468:24
_convert@webpack-internal:///283:511:24
getFeatures/<@webpack-internal:///779:126:28
```
and
```
TypeError: result is undefined
Stack trace:
altitudeLine@http://localhost:8080/examples/globe_wfs_extruded.js:48:21
getAltitude@webpack-internal:///283:13:20
geometryToLine@webpack-internal:///283:166:20
geometryToMesh@webpack-internal:///283:429:24
featureToThree@webpack-internal:///283:452:16
featureCollectionToThree@webpack-internal:///283:468:24
_convert@webpack-internal:///283:511:24
getFeatures/<@webpack-internal:///779:126:28
```

autra:
Forget this, seems like I didn't pull (maybe)

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,263 +57,326 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
+        let offset = 0;
+        for (let i = 0; i < geometry.length; i++) {
+            vertices = vertices.concat(geometry[i].vertices);
+            multiGeometries.push({
+                offset,
+                count: geometry[i].vertices.length,
+            });
+            offset += geometry[i].vertices.length;
+        }
+        multiGeometries.vertices = vertices;
```
zarov, line 357: Not a big fan as `multiGeometries` is an array.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,263 +57,326 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
+
+        for (let i = 0; i < geometry.length; i++) {
+            const holes = geometry[i].holes.map(h => h.offset);
+            // triangulate the top face
+            const start = vertices.length + multiGeomAttributes[i].offset + geometry[i].contour.offset;
+            const end = vertices.length + multiGeomAttributes[i].offset + multiGeomAttributes[i].count;
+            const triangles = Earcut(geom.attributes.position.array.slice(start * 3, end * 3),
+                holes,
+                3);
+            for (const indice of triangles) {
```
zarov: If you try the `wfs` example, you'll get some error: `glDrawElements: attempt to access out of range vertices in attribute 0`. This is caused by the `indices` array being false, and from what I've seen it is only when `triangles` is empty at this stage.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -57,263 +57,326 @@ function fillColorArray(colors, offset, length, r, g, b) {
...
-    return new THREE.LineSegments(geometry);
+    return new THREE.Points(geom);
 }
 
-function coordinateToPolygon(coordinates, properties, options) {
+function geometryToLine(geometry, properties, options, multiGeomAttributes) {
+    const vertices = multiGeomAttributes ? multiGeomAttributes.vertices : geometry.vertices;
+
+    // get altitude / color from properties
+    const altitude = getAltitude(options, properties, vertices);
```
zarov, line 163: In `globe_wfs_extruded`, it seems that `altitude` resolves as an array, making everything a `NaN` afterwards.

zarov (changes requested):
Thanks for doing this, I encountered some problems unfortunately.

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: I don't like this either but do you have any other proposal?

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: The `altitude` is an array case is supposedly handled here: https://github.com/iTowns/itowns/pull/691/files#diff-2a78502be7a11a150dcce98113215193R68

So I'm not sure what's going wrong.

zarov: My bad, I was working with a previous version. All good for this !

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: Either make it an object, containing the array and `vertices`, or make `vertices` available outside this condition to make it available farther below.

peppsac: Oh, right :)

I made it an object.

peppsac:
I added a pair of unit test (with and without holes), validating the Mesh built are correct by computing the area.

autra (approved):

autra:
LGTM. I pushed 2 commits, please r? (and squash) them before merging:
- one commit fixing the errors I've seen in console. It was probably already there before this PR btw. I *think* it happens when we load the geojson before an elevation is loaded there. There might be a better way of fixing this though.
- one commit fixing some nits (Array.isArray and inverting some loops to have positive logics).

Thanks!

zarov (approved):

PR #693, opened 2018-03-21
refactor(gis): remove _internalUnitStorage from Coordinates/Extent

peppsac:
_internalUnitStorage was initially introduced in fc6bc1a0f8d75edba03d to be able to store coordinates using EPSG:4326 but with radians angles to avoid too many conversions degrees <-> radians.

While this possibility is useful, it also introduces a tricky issue: 2 Coordinates with the same *crs* might not use the same internal representation.
(see https://github.com/iTowns/itowns/pull/690, or https://github.com/iTowns/itowns/pull/664 for recent fixes related to this).

This PR removes `_internalUnitStorage` completely and instead introduces a custom EPSG code (`EPSG:4326:R`) when using 4326 with Radians.


examples/FeatureToolTip.js
```diff
@@ -62,8 +62,8 @@ function ToolTip(viewer, viewerDiv, tooltip, precisionPx) {
                         symb = '<span id=' + name + ' style=color:' + color + ';>&#9679</span>';
                         label = point.properties.name || point.properties.description || layer.name;
                         tooltip.innerHTML += '<div>' + symb + ' ' + label + '<br></div>';
-                        tooltip.innerHTML += '<span class=coord>long ' + point.coordinates.longitude().toFixed(4) + '<br /></span>';
-                        tooltip.innerHTML += '<span class=coord>lati &nbsp; ' + point.coordinates.latitude().toFixed(4) + '<br /></span>';
```
zarov: Why not removing it everywhere ? Or do not remove it at all.

src/Core/Geographic/Coordinates.js
```diff
@@ -55,13 +57,13 @@ export function crsToUnit(crs) {
...
 }
 
-export function reasonnableEpsilonForUnit(unit) {
-    switch (unit) {
-        case UNIT.RADIAN: return 0.00001;
-        case UNIT.DEGREE: return 0.01;
-        case UNIT.METER: return 0.001;
-        default:
-            return 0;
+export function reasonnableEpsilonForUnit(crs) {
```
zarov: Why not rename it to `reasonnableEpsilonForCRS` instead ?

src/Core/Geographic/Extent.js
```diff
@@ -380,24 +379,29 @@ Extent.prototype.set = function set(...values) {
 
 Extent.prototype.union = function union(extent) {
     if (extent.crs() != this.crs()) {
-        throw new Error('unsupported union between 2 diff crs');
+        // If one of them is using Radians
```
zarov: Bad comment ?

zarov (changes requested):
Thanks for removing this. You missed the use in `obb_unit_test` though.

I feel that the radian parts (like in `latitude`, `longitude` etc...) could be made more generic, for other CRS than `EPGS:4326` that may use radian too, so maybe drop the `is4326Radians` for something more generic, no ? But that would mean to use something similar to what you removed so I'm not really sure.

Concerning the example, there are problems with:
- `multiglobe`: `Uncaught Error: Can't query crs EPSG:4326:R x/y/z`
- `split`: same error
- `globe_vector`: information box is way out of bounds

src/Core/Geographic/Coordinates.js
peppsac: Done

src/Core/Geographic/Extent.js
peppsac: No the comment is correct but maybe unclear: we know (because of the above `if` that the 2 crs are different, and if the condition below the comment is true, then the 2 crs are using 4326 => one of them must be in radians.

peppsac:
I pushed a commit to fix the issues you had.

> I feel that the radian parts (like in latitude, longitude etc...) could be made more generic, for other CRS than EPGS:4326 that may use radian too

This radian thing is an internal optimization (or hack depend of the POV) so I'd rather not have other CRS using it.
Btw, I'm wondering if it shouldn't be completely hidden from the user point of view, eg: always default to Degrees unit when in all functions accepting a unit param (instead of defaulting to returning the raw value).

peppsac:
PR updated: removed completely the `radians` unit support.

zarov (approved):

Issue #695, opened 2018-03-26
Split providers into protocols, formatters and symbolizers


Jeremy-Gaillard:
Follow up to [this](https://github.com/iTowns/itowns/issues/182#issuecomment-364138674) comment.

Providers are presently in charge of handling a protocol, as well as generating and styling geometries.

These tasks should be split in several modules to improve factorisation:

- Providers: they fetch data following a specific protocol
- Formatters: they transform the raw data given by the providers into a generic TypedArray (when possible)
- Symbolizers: they create Three.js objects from the generic TypedArrays and styling rules

TODO list:
- [x] Move and rename JS files in charge of formatting into a *Parser* directory
- [x] Homogeneize the interface of parsers to a single static function : `parse(file, options) -> promise`
- [ ] ~Move formatting and symbolizing code out of providers~ 
- [ ] Move symbolizing code out of formatters
- [ ] Create a generic symbolizer module supporting:
  - [ ] [3D Tiles styling](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/Styling)
  - [ ] [Vector Tile styling](https://www.mapbox.com/vector-tiles/)
  - [ ] [Additional styling from PLU2PLUS](https://github.com/IGNF/PLU2PLUS)

gchoqueux:
Now the itowns architecture is
* Layer to add data on viewer, with process to update data.
* Source to define data access
* the chain fetch/parse/convert to provide data to viewer

PR #696, opened 2018-03-26
refactor(parsers): move and rename parsers to a new parser directory

mbredif:
## Description
Multiple classes are in charge of decoding file chunks. They were scattered into multiple directories and had different names : Loaders, Parsers, IoDrivers...
They are now all moved to a new src/Parser directory.

## Motivation and Context
This commit addresses first item of #695 : "Move and rename JS files in charge of formatting into a Parser directory"


src/Parser/KmzParser.js
```diff
@@ -4,19 +4,19 @@
 
 import JSZip from 'jszip';
 import * as THREE from 'three';
-import Coordinates from '../../Core/Geographic/Coordinates';
+import Coordinates from '../Core/Geographic/Coordinates';
 
-function KMZLoader() {
+function KMZParser() {
```
zarov: If the file is `KmzParser` please rename it here as `KmzParser` too.

zarov (approved):
LGTM thanks !

zarov (changes requested):
Not good in fact, there are still some `ioDXBIL` references.

zarov (approved):
My bad, sorry :(

PR #698, opened 2018-03-27
Replace TextDecoder polyfill with static instance

Jeremy-Gaillard:
## Description

This PR replaces #634 and depends on #696.

Replace TextDecoder polyfill with static utf8 decoder instance.

## Motivation and Context

Remove the need to pass TextDecoder in function parameters or to instantiate it at each use.


autra (approved):
LGTM thanks!

PR #700, opened 2018-03-27
feature(Core): add function view.on to listen for a certain type of event

gchoqueux:
Add function view.on to listen for a certain type of event

The returned viewEvent properties:
   * `type`
   * `normalizedCoords`
   * `viewCoords`
   * `pickeds`
   * `Coordinates`

Use case in `3DTiles.html` example

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -131,57 +131,35 @@
             $3dTilesLayerRequestVolume.type = 'geometry';
             $3dTilesLayerRequestVolume.visible = true;
             $3dTilesLayerRequestVolume.sseThreshold = 1;
+            // Picking example - - - - - - - - - - -- - - - -- - - - -- - - - -- - - - -- - - - -- - - - -
+
+
+            function picking(event) {
+                var htmlInfo = document.getElementById('info');
+                var pickeds = event.pickeds;
+                htmlInfo.innerHTML = ' ';
+                if (pickeds[0]) {
+                    Object.keys(pickeds[0].properties).map(function _(objectKey) {
+                        var value = pickeds[0].properties[objectKey];
+                        // if the value is a integer or not
+                        var info = Number.isInteger(value) ?
+                            value.toString() :
+                            value.toFixed(3).toString();
+                        htmlInfo.innerHTML += '<li><b>' + objectKey.toString() + ': </b>' + info + '</li>';
+                    });
+                }
+            }
+
             // add an event for have information when you move your mouse on a building
-            itowns.View.prototype.addLayer.call(view, $3dTilesLayerRequestVolume).then(function _() { window.addEventListener('mousemove', picking, false); })
+            itowns.View.prototype.addLayer.call(view, $3dTilesLayerRequestVolume).then(function _(layer) {
+                layer.on('mousemove', picking);
+            });
 
             // Add the UI Debug
             var d = new debug.Debug(view, menuGlobe.gui);
             debug.createTileDebugUI(menuGlobe.gui, view, view.wgs84TileLayer, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerDiscreteLOD, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerRequestVolume, d);
-
-
-// Picking example - - - - - - - - - - -- - - - -- - - - -- - - - -- - - - -- - - - -- - - - -
-            function findBatchTableParent(object) {
-                if (object.batchTable) {
-                    return object.batchTable;
-                }
-                else if (object.parent) {
-                    return findBatchTableParent(object.parent);
-                } else {
-                    return undefined;
-                }
-            }
-            function picking(event) {
-                var htmlInfo = document.getElementById('info');
-                htmlInfo.innerHTML = ' ';
-
-                const intersects = view.pickObjectsAt(
-                    event,
-                    $3dTilesLayerRequestVolume);
-
-                for (var i = 0; i < intersects.length; i++) {
-                    var interAttributes = intersects[i].object.geometry.attributes;
-                    if (interAttributes) {
-                        if (interAttributes._BATCHID) {
-                            var face = intersects[i].face.a;
-                            var batchID = interAttributes._BATCHID.array[face];
-                            var batchTable = findBatchTableParent(intersects[i].object);
-
-                            htmlInfo.innerHTML +='<li><b> Batch id: </b>'+ batchID +'</li>';
-                            Object.keys(batchTable).map(function(objectKey) {
-                                var value = batchTable[objectKey][batchID];
-                                // if the value is a integer or not
-                                var info = Number.isInteger(value) ? value.toString() : value.toFixed(3).toString();
-                                htmlInfo.innerHTML +='<li><b>' + objectKey.toString() + ': </b>'+ info +'</li>';
-                                return true;
-                            });
-
-                            return;
-                        }
-                    }
-                }
-            }
 
         </script>
         <div id="batchInformationDiv">
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -2,6 +2,7 @@ import * as THREE from 'three';
 import TileMesh from './TileMesh';
 import RendererConstant from '../Renderer/RendererConstant';
 import { unpack1K } from '../Renderer/LayeredMaterial';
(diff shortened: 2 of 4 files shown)


gchoqueux:
@iTowns/reviewers PR ready
It remains to update examples and controls

examples/3dtiles.html
```diff
@@ -134,58 +126,36 @@
...
-                                // if the value is a integer or not
-                                var info = Number.isInteger(value) ? value.toString() : value.toFixed(3).toString();
-                                htmlInfo.innerHTML +='<li><b>' + objectKey.toString() + ': </b>'+ info +'</li>';
-                                return true;
-                            });
-
-                            return;
-                        }
-                    }
+                if (pickeds[0]) {
```
autra, line 141: So all this code just to remove the `view.pickObjectsAt`call, making things more opaque and slower in the process? (`defineProperty` makes things very slow)

src/Core/Picking.js
```diff
@@ -137,11 +137,20 @@ export default {
             view.viewToNormalizedCoords(mouse),
             view.camera.camera3D);
         const intersects = raycaster.intersectObject(object, true);
+        function layerIdToLayer(id) { return view.getLayers(l => l.id == id); }
         for (const inter of intersects) {
             inter.layer = findLayerIdInParent(inter.object);
+            const protocol = layerIdToLayer(inter.layer)[0].protocol;
+            switch (protocol) {
```
autra, line 144: We already have a mechanism for custom picking logic per protocol, this file shouldn't have any protocol specific code.

src/Core/View.js
```diff
@@ -629,4 +632,96 @@ View.prototype.pickObjectsAt = function pickObjectsAt(mouseOrEvt, ...where) {
...
+    }
+
+    const viewerListener = (event) => {
+        event.where = where;
+        listener(event);
+    };
+
+    viewListeners[listener] = viewerListener;
+    this.addEventListener(type, viewerListener);
+};
```
autra, line 725: This is not the correct approach. Imho the interesting part of this PR is the autocalculation of `view.eventToViewCoords`, but in this case, you don't need a new object and this complicated double-event listeners. Just proxy and call user callback with your augmented event (you can add `viewX` and `viewY` properties to the MouseEvent for examples).

src/Core/View.js
```diff
@@ -629,4 +632,96 @@ View.prototype.pickObjectsAt = function pickObjectsAt(mouseOrEvt, ...where) {
     return results;
 };
 
+function viewEvent(type, event, view) {
+    this.type = type;
+    this.viewCoords = view.eventToViewCoords(event).clone();
+    let normalizedCoords;
+    let pickeds;
+    let coordinate;
+    Object.defineProperty(
```
autra, line 643: For a thing that'll get called every mouse event (so even faster than `requestAnimationFrame` in some case), we shouldn't use defineProperty as it prevents all sort of optimization.

autra (changes requested):
I'm not against `view.on` logic, but I think we could have a better approach, at least for the picking use case. 

The 2 things that are interesting to abstract to the user are:
- the coordinate conversion, because it's dull and always the same. Also it fits well into the `View.on` semantic: you expect a `view` coordinate
- the polling: picking can be quite costly and MouseEvent can fire too often. Instead of intercepting the mouse event to dispatch *another* event (which double the cost and the complexity of the code), View.on could abstract the creation of a frame requester and only call the user callback every frame (or even every n frames, where n is configurable by the user).

This'll give the user flexibility without magic, and control over how often his callback is called when 60fps is important. That'll also gives `view.on` a real added value comparing to just syntaxic sugar over `addEventListener`.

src/Core/View.js
gchoqueux: So, [the second implementation](
https://github.com/iTowns/itowns/pull/700/files/8d4537380ce0ae02401914a7c212dcae07c5a9f0#diff-d297dcf7b9d0e696ece4b1f6e1d7c69dR675) is better

gchoqueux:
> the coordinate conversion...

Are you talking about the view's coordinates or geographical coordinates?

> the polling: picking can be quite costly...

With what event in MAIN_LOOP_EVENTS, do you want to return the mouse movement?
How to get the mouse position without events?

src/Core/Picking.js
gchoqueux: Where will you place it?

examples/3dtiles.html
gchoqueux: > So all this code just to remove the view.pickObjectsAtcall

Thank you for the encouragements :(
But if you didn't understand the interest of the PR, you should not answer like that

> more opaque 

As I said I have not done all the examples yet.

`eventView` unifies the compute of `pickObjectsAt`, but also `coordinate` and `viewCoords`
we can also add `viewEvent.features`

`eventView` also allows you to calculate once all the properties and especially not to recalculate them when there are several event calls. 

For example if the controls calculates a geographic coordinate under the mouse click, this calculation will be shared to the user in his event callback

Yes actually itowns make opaque a lot of code, it's a question of choice. And it's not your only opinion that decides what should be in or out of itowns.

> slower in the process

The 1 first `defineProperty` can be replace by `class` `eventViewer`

autra: Okay let's take it point by points, meta first:

> Yes actually itowns make opaque a lot of code, it's a question of choice. 

I'm not sure we use the word "opaque" in the same sense then. Opaque is never good for code, period. And it is even more the case when one have the ambition to develop a **framework**.

If it's a question of choice, then mine is all made and I'm not gonna r+ PRs that decrease code clarity if there is not a clear interest in doing so. I don't think it's the case here.

> And it's not your only opinion that decides what should be in or out of itowns.

Why the aggressivity? I have never pretended so, but as a reviewer, I do reviews. Something against that?

Now onto this PR:

>  But if you didn't understand the interest of the PR, you should not answer like that

I do understand what you want to do. I disagree with the current implementation of `view.on`. I disagree on creating a custom event class and adding an object creation in an event handler (which is perf problem n°2 btw) just to replace 2 lines of code (coords calculation + picking function calculation calls).

Is it that complicated to do:

```js
domElement.addEventListener('mousemove', () => {
    const vector = view.getPickingPositionFromDepth(eventToViewCoords(mouseEvent));
    // user then can create (or not!) coordinates from this Vector3. 
    // Or use it directly if they want to display an object for instance
});
```
? 
It's simpler, more flexible. Fits more with the fact we don't want to hide THREE.js logic, which is the differentiating factor of iTowns (or should be at least).

> eventView also allows you to calculate once all the properties and especially not to recalculate them when there are several event calls.

You do realize that your code is not doing that, right? Just to be sure :-)

> For example if the controls calculates a geographic coordinate under the mouse click, this calculation will be shared to the user in his event callback

In this case, this calculated value is specific to a control type, so belongs to the control API interface (and should be documented). The user can then register a framerequester to query this cached value at each frame. Your current code only works for the globe, and more precisely, for the globe with GlobeControls.

I do agree that some general helpers can be added (and `eventToViewCoords` was one of them). I do agree that we either need a shortcut to renderer.domElement, or a `view.on` shortcut. But in this case, view.on should just be:

```js
View.prototype.on = function on(event, cb) {
    view.mainLoop.gfxEngine.renderer.domElement.addEventListener(event, cb);
}
```
But I'd prefer a shortcut to the domElement (`view.domElement`? `view.canvas`?), because it's more "webby": people will understand they can use all the standard web stuff on this element and they won't have to rediscover a event api  custom to itowns.

I'll explain my idea about framerequester on the other comment.

src/Core/Picking.js
autra: See #635 (you were the reviewer on this PR). In this precise case, have a `pick3dTilesObjects` that calls this method and add some specific 3dtiles properties on the result can be one possible solution.

autra:
> > the polling: picking can be quite costly...
> 
> With what event in MAIN_LOOP_EVENTS, do you want to return the mouse movement?
> How to get the mouse position without events?

See what is done in `FirstPersonControls` or `FlyControls`: they do add event listeners, but inside of it, they only store current mouse positions to be later processed by their `update` function: you can do more expensive stuff there (like picking), because it's not called as often as mouse event callbacks. Also, their rate is more predictable from device to device.

`view.on/off` could be a shortcut to this logic. A possible approach: they subscribe one and only one event listener getting mouse coordinates. Then they register one frame requester that calls all the user provided callbacks with viewCoords.

PR #701, opened 2018-03-27
Homogenize parser interface

Jeremy-Gaillard:
## Description

Part of #695.

Changed parsers so they share a same interface:
- File (or buffer) and options parameter
- Return a promise. The wrapped type is not homogenized yet

Also contains some little improvements:
- Additional documentation for parsers
- Renamed BatchTable to BatchTableParser
- Removed obsolete KMZLoader
- Removed Xbil export from OGCWebServiceHelper


mbredif:
LGTM

peppsac (approved):
+1 thanks!

PR #704, opened 2018-03-27
Move providers

Jeremy-Gaillard:
Providers are now in the *Provider* directory at the root of the source, like processes and parsers.

Underscores were also removed from the few provider names that still had them.


zarov (approved):
LGTM, thanks

PR #707, opened 2018-03-28
fix(protocols): respect the order of the file features to draw texture

gchoqueux:
## Description
Respect the order of the file features to draw texture


peppsac (changes requested):
Looks more like a hack than a fix to me (it will "fix" the problem on one file and break other files).
If you want a simple fix, you can probably allow to specify `globalCompositeOperation` in the `style` properties instead.

mbredif:
Would you agree on having a `style` property that defaults to `'destination-over'` ?

peppsac:
> Would you agree on having a style property that defaults to 'destination-over' ?

Yep

gchoqueux:
done

peppsac (approved):

PR #708, opened 2018-03-28
chore(parser): remove unneeded gpx/geojson parser options

peppsac:
Since we can detect if a given object has already been parsed or not,
there's no need to require the user to specify this fact using an option.


zarov (approved):
LGTM, thanks

PR #709, opened 2018-03-28
Fix 3D tiles points and clean 3D Tiles Provider

Jeremy-Gaillard:
## Description
- Move mesh initialisation out of 3dTilesProvider into B3dmParser
- Fix 3D tiles pnts loading
- Add a button to 3D tiles example to jump to the point cloud

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -85,6 +85,7 @@
             var viewerDiv = document.getElementById('viewerDiv');
 
             var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            view.camera.camera3D.near = 5;
             setupLoadingScreen(viewerDiv, view);
 
             var menuGlobe = new GuiTools('menuDiv', view, 300);
@@ -139,7 +140,12 @@
             debug.createTileDebugUI(menuGlobe.gui, view, view.wgs84TileLayer, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerDiscreteLOD, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerRequestVolume, d);
-
+            d.zoom = function() {
+                view.camera.camera3D.position.set(1215013.9, -4736315.5, 4081597.5);
+                view.camera.camera3D.quaternion.set(0.9108514448729665, 0.13456816437801225, 0.1107206134840362, 0.3741416847378546);
+                view.notifyChange(true);
+            }
+            menuGlobe.gui.add(d, 'zoom').name('Go to point cloud');
 
 // Picking example - - - - - - - - - - -- - - - -- - - - -- - - - -- - - - -- - - - -- - - - -
             function findBatchTableParent(object) {
--- a/src/Parser/B3dmParser.js
+++ b/src/Parser/B3dmParser.js
@@ -2,6 +2,8 @@ import * as THREE from 'three';
 import GLTFLoader from './GLTFLoader';
 import LegacyGLTFLoader from './LegacyGLTFLoader';
 import BatchTableParser from './BatchTableParser';
+import Capabilities from '../Core/System/Capabilities';
+import shaderUtils from '../Renderer/Shader/ShaderUtils';
 import utf8Decoder from '../utils/Utf8Decoder';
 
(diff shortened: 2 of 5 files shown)


src/Parser/GpxParser.js
```diff
@@ -8,7 +8,7 @@ import * as THREE from 'three';
 import Line from 'three.meshline';
 import Coordinates from '../Core/Geographic/Coordinates';
 import Capabilities from '../Core/System/Capabilities';
-import { patchMaterialForLogDepthSupport } from '../Provider/3dTilesProvider';
+import { patchMaterialForLogDepthSupport } from './B3dmParser';
```
mbredif: While you are at moving patchMaterialForLogDepthSupport, it does not look like it is more specific to B3DM than to 3Dtiles as it is used by GPX. Should not it be placed in a more generic location ?

src/Parser/B3dmParser.js
```diff
@@ -54,6 +86,9 @@ export default {
      * @param {Object} options - additional properties.
      * @param {string=} [options.gltfUpAxis='Y'] - embeded glTF model up axis.
      * @param {string} options.urlBase - the base url of the b3dm file (used to fetch textures for the embeded glTF model).
+     * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patchin material with logarithmic depth buffer support.
+     * @param {float} [options.opacity=1.0] - the b3dm opacity.
+     * @param {boolean|Material=} [options.overrideMaterials='false'] - override b3dm's embeded glTF materials. If overrideMaterials is a three.js material, it will material used to override.
```
mbredif: [options.material] would be a simpler naming.

> it will material used to override.

This could be rephrased to "this material will used to override existing glTF materials."

src/Parser/GpxParser.js
Jeremy-Gaillard: Done. Moved it to a *ShaderUtils* file.

src/Parser/B3dmParser.js
Jeremy-Gaillard: Fixed the doc, but kept the name. I think it is important to insist on the fact that we override the B3dm material and not only define one.

Jeremy-Gaillard:
Updated. Note that the "go to point cloud" button is a bit buggy. The current globe controls seem to have trouble with positions very close to the globe (poke @gchoqueux), so I had to set directly the camera's position, which is pretty unreliable when used with an active controller.

mbredif:
Thanks @Jeremy-Gaillard for addressing my comments.
(as a note, I see typos 'embeded' should be 'embedded')

Jeremy-Gaillard:
Fixed.

examples/3dtiles.html
```diff
@@ -85,6 +85,8 @@
             var viewerDiv = document.getElementById('viewerDiv');
 
             var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            view.camera.camera3D.near = 5;
+            view.camera.camera3D.updateMatrixWorld();
```
peppsac: The `near` value isn't used by the `updateMatrixWorld` method, so there's no need to call it.

examples/3dtiles.html
```diff
@@ -139,7 +141,12 @@
             debug.createTileDebugUI(menuGlobe.gui, view, view.wgs84TileLayer, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerDiscreteLOD, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerRequestVolume, d);
-
+            d.zoom = function() {
+                view.camera.camera3D.position.set(1215013.9, -4736315.5, 4081597.5);
+                view.camera.camera3D.quaternion.set(0.9108514448729665, 0.13456816437801225, 0.1107206134840362, 0.3741416847378546);
+                view.camera.camera3D.updateMatrixWorld();
```
peppsac: You should replace this line by a call to `view.notifyChange(true)` to make sure the rendering is updated (and it will call `camera3D.updateMatrixWorld()` for you)

peppsac (changes requested):
2 things to fix before merging :)

Jeremy-Gaillard:
@peppsac done, thanks!

peppsac (approved):

PR #710, opened 2018-03-28
feat(core): add vector tile loading in TMS

zarov:
Add vector tile loading in TMS protocol, with only Mapbox Vector Tile supported for now. It adds a new Parser (VectorTileParser), and other utility functions for vector tiles are in VectorTileHelper. 
Thanks to @peppsac for helping on this.

This is still in progress, as I have two things I want to clarify:
- should we keep the cache in `VectorTileHelper` ? If we drop it, it can drastically reduce the file.
- some things have been added for the styling in `Feature2Texture`, maybe it can be moved into another PR / commit ?

Also I need to add a correct `package-lock.json` file.


zarov:
I updated the PR, with two working examples. It seems like a big PR, but you can review it as two differents commits, the first one for the core and the other for the examples.

examples/globe_vector_tile.js
```diff
@@ -0,0 +1,132 @@
...
+            zoom: {
+                min: 2,
+                max: 15,
+            },
+            opacity: 0.5,
+        },
+        updateStrategy: {
+            type: itowns.STRATEGY_DICHOTOMY,
+        },
+        style: function (properties, feature) {
```
peppsac: I'd move the mapbox-style converter to itowns core.

examples/globe_vector_tile.js
```diff
@@ -0,0 +1,132 @@
...
+                styles.push(r);
+            });
+
+            if (styles.length === 1) {
+                return styles[0];
+            }
+
+            return styles;
+        },
+        filter: function (properties, geometry) {
```
peppsac: Idem: the implementation of mapbox style filtering should go in core.

examples/globe_vector_tile.js
```diff
@@ -0,0 +1,132 @@
...
+                }
+            });
+            return properties.mapboxLayer.length > 0;
+        },
+    }));
+});
+
+
+exports.view = view;
+exports.initialPosition = positionOnGlobe;
```
peppsac: Is this needed?

examples/planar_vector_tile.js
```diff
@@ -0,0 +1,180 @@
...
+            });
+            return properties.mapboxLayer.length > 0;
+        },
+    });
+});
+
+viewerDiv.addEventListener('DOMMouseScroll', onMouseWheel);
+viewerDiv.addEventListener('mousewheel', onMouseWheel);
+
+viewerDiv.addEventListener('mousedown', function mouseDown(event) {
```
peppsac: I suppose you'll want to use `PlanarControls` instead, like you did in #734 ?

examples/planar_vector_tile.html
```diff
@@ -0,0 +1,54 @@
...
+            body {
+                margin: 0;
+                overflow: hidden;
+                height: 100%;
+            }
+
+            #viewerDiv {
+                margin: auto auto;
+                width: 100%;
+                height: 100%;
```
peppsac: We could add a `example.css` file instead of copy-pasting everytime :)

peppsac (changes requested):

examples/globe_vector_tile.js
zarov: I was thinking of maybe doing another PR introducing Stylizers ?

examples/globe_vector_tile.js
zarov: Nope, bad copy paste

examples/planar_vector_tile.js
zarov: Yep, after merging 734 I'll remove this here.

examples/planar_vector_tile.html
zarov: Yes !

examples/globe_vector_tile.js
peppsac: IMHO it's parsing / translating mapbox-style format, so you could already move this in `VectorTileHelper.js`

examples/planar_vector_tile.html
zarov: Did #736 about that

examples/globe_vector_tile.js
zarov: The problem I encounter here is with the `mapboxLayers` global var: I can't figure how to pass properly this to internal methods in iTowns. Do you have a suggestion ?

peppsac: - add a `style: https://raw.githubusercontent.com/Oslandia/postile-openmaptiles/master/style.json` to the layer
- fetch this json in the `preprocess` method of the provider. Then parse it and store the mapboxLayers in the layer?
WDYT?

zarov: That could be one option yes, I'm trying it out

src/Parser/VectorTileParser.js
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @return {Promise} A Promise resolving with a Feature or an array a
+     * Features.
+     */
+    parse(file, options = {}) {
+        switch (options.format) {
+            case 'application/x-protobuf;type=mapbox-vector':
+            default:
+                return Promise.resolve(readPBF(file, options));
+        }
```
mbredif: Should not you `Promise.reject` the default case ?

src/Parser/VectorTileParser.js
```diff
@@ -0,0 +1,85 @@
...
+            if (layers.length > 1) {
+                feature.properties.vt_layer = layer_id;
+            }
+
+            geojson.features.push(feature);
+        }
+    });
+
+    return GeoJsonParser.parse(geojson, {
+        json: true,
```
mbredif: the `json` option is not used anymore, please remove it ([it is now automatic](https://github.com/iTowns/itowns/blob/master/src/Parser/GeoJsonParser.js#L273-L275))

src/Provider/TMSProvider.js
```diff
@@ -23,32 +24,56 @@ function preprocessDataLayer(layer) {
...
     const promises = [];
+    const supportedFormats = {
+        'application/x-protobuf;type=mapbox-vector': getVectorTile.bind(this),
+    };
+    const func = supportedFormats[layer.format];
+
     for (const coordTMS of tile.getCoordsForLayer(layer)) {
         const coordTMSParent = (command.targetLevel < coordTMS.zoom) ?
             OGCWebServiceHelper.WMTS_WGS84Parent(coordTMS, command.targetLevel) :
             undefined;
```
mbredif: Is it intentionnal that `coordTMSParent` is not used for vector tiles ? if yes, please move it to the OGC case.
My guess is that `const urld = URLBuilder.xyz(coordTMSParent || coordTMS, layer);` could be instead moved out here of the format switch (and of `getVectorTile`).

src/Provider/TMSProvider.js
```diff
@@ -23,32 +24,56 @@ function preprocessDataLayer(layer) {
...
+    const url = URLBuilder.xyz(coords, layer);
+
+    if (layer.type == 'color') {
+        return VectorTileHelper.getVectorTileTextureByUrl(url, tile, layer, coords)
+            .then((texture) => {
+                const result = {
+                    texture,
+                };
+                return result;
+            });
```
mbredif: could not this wrapping be done in `VectorTileHelper.getVectorTileTextureByUrl` and the `URLBuilder.xyz` be done outside of the format switch (see below), so as to remove this function entirely ?

mbredif (commented):
Looks good !
I do not really like the coupling of the `TMSProvider` introduced by the imports of parsers/helpers (`OGCWebServiceHelper` and `VectorTileHelper`) and of this local registry `supportedFormats`, but this is the current practice in itowns. IMHO, a parser/helper object should be instead provided in the layer options.

src/Parser/VectorTileParser.js
zarov: I was on the assumption that if someone uses this parser, the file is always a vector tile, whatever the format, with the default case being being this one. So no rejection, as we trust the user to know what he does.

src/Parser/VectorTileParser.js
zarov: Thanks !

src/Provider/TMSProvider.js
zarov: Yes it is intentional. What do you mean by moving it to the OGC case ?

src/Provider/TMSProvider.js
zarov: This wrapping could, but unfortunately the url no.

zarov:
Thanks !

> I do not really like the coupling of the TMSProvider introduced by the imports of parsers/helpers (OGCWebServiceHelper and VectorTileHelper) and of this local registry supportedFormats, but this is the current practice in itowns. IMHO, a parser/helper object should be instead provided in the layer options.

I agree, this is not ideal, but we can either merge this PR now, or wait for all the modifications we wants to be integrated in iTowns: it doesn't matter to me, I'm ok with waiting.

src/Parser/VectorTileParser.js
mbredif: If you trust the user, then there is no use to switch on the format (maybe you have an extension in mind, but for now the switch is useless, just return the resolve directly).
If you want to advertise the mimetype, I propose in #723 to make it a field of the parsers.

zarov: Yes, that's why I'm eager to see the parsers coming here :grinning:

src/Provider/TMSProvider.js
mbredif: `URLBuilder.xyz` is called at two places with the same arguments : 
- L31 `const url = URLBuilder.xyz(coords, layer); `  (coords being `coordTMSParent || coordTMS`)
- L55 `const urld = URLBuilder.xyz(coordTMSParent || coordTMS, layer); `
You could move it right after L51 and pass the url to `VectorTileHelper.getVectorTileTextureByUrl` directly (removing the need for a `getVectorTile` function)

src/Provider/TMSProvider.js
```diff
@@ -23,32 +24,50 @@ function preprocessDataLayer(layer) {
...
+                    new THREE.Vector4(0, 0, 1, 1);
+                result.texture = texture;
+                result.texture.coords = coordTMSParent || coordTMS;
+                result.pitch = pitch;
+                if (layer.transparent) {
+                    texture.premultiplyAlpha = true;
+                }
+                return result;
+            }));
+        }
```
mbredif: now that you use `coordTMSParent` if `func` is defined, dont you need to introduce this postprocessing (most notably `pitch`) in all cases ?

src/Provider/TMSProvider.js
zarov: Yeah I forgot to remove the `coordTMSParent ||` for the vector tile when I was testing before, if it is added tiles are not showing correctly.

mbredif: Does it make sense to support `command.targetLevel` and `coordTMSParent` for vector tiles as well ? if not, why ? if yes, then I think this feature could be added to Vector tiles as well, that would help to have an equal treatment for OGCWebService, VectorTiles and any future formats... 

It is not a blocker to me, but if you merge it like this please open an issue about this for future reference.

src/Provider/TMSProvider.js
zarov: A quick update: I need to keep this method if I want to support different type of layer (like `geometry`), as it is the only way for now ;)

zarov:
@iTowns/reviewers I rebased this PR, this should be good to validate unless I missed something

examples/globe_vector_tile.html
```diff
@@ -0,0 +1,157 @@
...
+            // These will deform iTowns globe geometry to represent terrain elevation.
+            promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
+            promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+
+            var mapboxLayers = [];
+
+            var count = 0;
+            // Add a vector tile layer
+            itowns.Fetcher.json('https://raw.githubusercontent.com/Oslandia/postile-openmaptiles/master/style.json').then(function (style) {
+                // add one layer per layer in style.json
```
peppsac: This comment is outdated and should be removed.

src/Provider/TMSProvider.js
```diff
@@ -23,32 +24,50 @@ function preprocessDataLayer(layer) {
...
+        return VectorTileHelper.getVectorTileTextureByUrl(url, tile, layer, coords);
+    }
 }
 
 function executeCommand(command) {
     const layer = command.layer;
     const tile = command.requester;
 
     const promises = [];
-    for (const coordTMS of tile.getCoordsForLayer(layer)) {
```
peppsac: I'd prefer to keep this logic, and adapt the vector-tiles processing to fit in.

(it'll probably needs to use `Cache` to be able to re-use parent tile)

src/Provider/VectorTileHelper.js
```diff
@@ -0,0 +1,77 @@
...
+        }));
+};
+
+/**
+ * @module VectorTileHelper
+ */
+export default {
+    /**
+     * Get a vector tile file, parse it and return a [Feature]{@link module:GeoJsonParser.Feature}
+     * or an array of Features. See [VectorTileParser]{@link module:VectorTileParser.parse}
```
peppsac: I think that this comment is wrong since GeoJsonParser always returns a FeatureCollection

src/Provider/VectorTileHelper.js
```diff
@@ -0,0 +1,77 @@
...
+     */
+    getVectorTileByUrl,
+
+    /**
+     * Get a vector tile, parse it and return a [THREE.Texture]{@link https://threejs.org/docs/#api/textures/Texture}.
+     *
+     * @param {string} url - The URL of the tile to fetch, NOT the template: use a
+     * Provider instead if needed.
+     * @param {TileMesh} tile
+     * @param {Layer} layer
```
peppsac, line 52: Maybe the documentation could be made more explicit here.
Do you really need a layer? or do you need an object with a couple of properties (.style, .filter, ... any others?)

examples/planar_vector_tiles.html
```diff
@@ -0,0 +1,94 @@
...
+
+                style.layers.forEach(function(layer) {
+                    if (layer.type === 'fill' || layer.type === 'line') {
+                        supportedLayers.push(layer);
+                    }
+                });
+
+                view.addLayer({
+                    type: 'color',
+                    protocol: 'xyz',
```
peppsac, line 65: Maybe we could introduce a `vector-tiles` protocol and guess `protocol`, `type`, `projection` and `format` automatically since we only support 1 combination of these attributes.

peppsac: I know that vector-tiles isn't a protocol, but 3dtiles isn't a protocol either. Since we currently mix protocol/format, we could take advantage of this to simplify the API.

src/Parser/VectorTileParser.js
```diff
@@ -0,0 +1,80 @@
+import Protobuf from 'pbf';
+import { VectorTile } from '@mapbox/vector-tile';
+import GeoJsonParser from './GeoJsonParser';
+
+function readPBF(file, options) {
```
peppsac, line 5: It would be nice to add at least one test for this function.

Trying to load a pbf file and verify the output features types / count would be a good start.

peppsac (commented):
Thanks for updating this PR!

src/Provider/VectorTileHelper.js
zarov: Yep nice catch

examples/planar_vector_tiles.html
zarov: I dunno, I think it is too much a hassle, since this (declaration of layer type, protocol etc) should be more generic in the future ? And in our example we only have the `xyz` protocol, but here at IGN we have internal example with the `tms` protocol.

src/Parser/VectorTileParser.js
zarov: Yep

src/Provider/TMSProvider.js
zarov: I'll try to see what I can do.

zarov:
What I find incredible here is the offset we got on the first level, see below. Is it because of the source or an error here ? 

image

examples/planar_vector_tiles.html
peppsac: Agreed, but you can still assume reasonnable default values, and allow overriding by the user (in the same way as #811).

src/Provider/TMSProvider.js
```diff
@@ -23,27 +24,34 @@ function preprocessDataLayer(layer) {
             max: 18,
         };
     }
+    layer.fx = layer.fx || 0.0;
 }
 
 function executeCommand(command) {
     const layer = command.layer;
     const tile = command.requester;
+    const isGlobeView = !command.view.wgs84TileLayer;
```
peppsac: Why do you need this?

test/unit/vectortiles.js
```diff
@@ -0,0 +1,34 @@
...
+    const coords = new Extent('TMS', 1, 1, 1);
+    const extent = new Extent(
+        'EPSG:3857',
+        -20037508.342789244, 20037508.342789244,
+        -20037508.342789255, 20037508.342789244);
+    return VectorTileParser.parse(pbf, { coords, extent });
+}
+
+describe('Vector tiles', function() {
+    it('should return two squares', () =>
```
peppsac: :+1:

examples/planar_vector_tiles.html
```diff
@@ -0,0 +1,95 @@
...
+                    id: 'MVT',
+                    // eslint-disable-next-line no-template-curly-in-string
+                    url: 'https://osm.oslandia.io/data/v3/${z}/${x}/${y}.pbf',
+                    extent: [extent.west(), extent.east(), extent.south(), extent.north()],
```
peppsac: You could simplify the example, and just use:
```js
url: 'https://osm.oslandia.io/data/v3/${z}/${x}/${y}.pbf',
extent,
format: 'application/x-protobuf;type=mapbox-vector',
```

(see TMSProvider.preprocessDataLayer)
```

File: src/Provider/TMSProvider.js (hunk 17 above)
Comment (zarov) [reply src/Provider/TMSProvider.js]: Forgot to remove it, sorry, it was my first fix for bad tiling when using a globe view.

File: examples/planar_vector_tiles.html (hunk 19 above)
Comment (zarov) [reply examples/planar_vector_tiles.html]: Done

File: src/Provider/TMSProvider.js [hunk 20]
```diff
@@ -23,27 +24,33 @@ function preprocessDataLayer(layer) {
...
+        const promise = layer.format === 'application/x-protobuf;type=mapbox-vector' ?
+            VectorTileHelper.getVectorTileTextureByUrl(urld, tile, layer, coordTMS) :
+            OGCWebServiceHelper.getColorTextureByUrl(urld, layer.networkOptions);
+
+        promises.push(promise.then((texture) => {
             const result = {};
-            result.texture = texture;
-            result.texture.coords = coordTMSParent || coordTMS;
-            result.pitch = coordTMSParent ?
+            const pitch = coordTMSParent ?
```
peppsac: I think all these changes can be removed, since it's equivalent to the original code.

src/Provider/VectorTileHelper.js
```diff
@@ -0,0 +1,95 @@
...
+    getVectorTileTextureByUrl(url, tile, layer, coords) {
+        if (layer.type !== 'color') return;
+
+        return Cache.get(url) || Cache.set(url, getVectorTileByUrl(url, tile, layer, coords).then((features) => {
+            let texture;
+
+            const backgroundColor = (layer.backgroundLayer && layer.backgroundLayer.paint) ?
+                new THREE.Color(layer.backgroundLayer.paint['background-color']) :
+                undefined;
+            if (features) {
```
peppsac: Suggestion: remove the `if (features)` from here and move it to `Feature2Texture.createTextureFromFeature` to avoid duplicating the backgroundColor handling?

WDYT?

examples/planar_vector_tiles.html
peppsac: You didn't reply here?

src/Provider/TMSProvider.js
zarov: Yep

examples/planar_vector_tiles.html
zarov: Sorry ! Yes we could but that means the introduction of a new provider if I'm not mistaken. And that doesn't make a lot of sense to me. We seem to both agree that the current `3dtiles` semantic is not the most appropriate, and that it should change, and I see this proposition for `vector-tiles` as doing a similar error like what we did for `3dtiles`.

IMHO we already have a lot of ongoing and oncoming PR refactoring the layers, providers, sources, parsers and other to take care of this later, WDYT ?

peppsac: Right, let's forget about the `vector-tiles` protocol for now.

src/Provider/VectorTileHelper.js
zarov: Yep, this could even be moved in another PR adding the `backgroundColor` functionnality ? Or I will at least create a new commit for it.

examples/planar_vector_tiles.html
zarov: Thanks !

src/Provider/VectorTileHelper.js
zarov: Done in 4f2987c

zarov:
@peppsac rebased to the last test changes.

src/Renderer/ThreeExtended/Feature2Texture.js
```diff
@@ -28,7 +39,6 @@ function drawPolygon(ctx, vertices, indices, origin, scale, properties, style =
             _lineTo(ctx, vertices[indice.offset + j], scale, origin);
         }
     }
-    ctx.closePath();
```
peppsac, line 31: Are you sure about this removal?

Maybe it's not needed because the polygons we draw are guaranteed to be closed? (but I'd keep it to be on the safe side anyway)...

peppsac (approved):
There's still a lot of work to be done to fully support vector-tiles but this PR is a great start!

src/Renderer/ThreeExtended/Feature2Texture.js
zarov: Yes this is a fix needed because not closing the path would create visual artifacts for `multipolygon`. I missed that when working on the geojson refactoring last time, sorry about that.

PR #712, opened 2018-03-29
fix(examples): fix mouse move event doesn't work under menu, in 3dtiles examples

gchoqueux:
## Motivation and Context
Renderer.domElement's **mousemove** doesn't work under menu, in 3dtiles examples

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -44,16 +44,13 @@
                     display: none;
                 }
             }
-            #menuDiv {
-                margin : auto auto;
-                width: 100%;
-                padding: 0;
-                height: 100%;
 
+            #menuDiv {
                 position: absolute;
                 top:0px;
                 margin-left: 0px;
             }
+
             @media (max-width: 600px) {
                 #menuDiv {
                     display: none;


autra:
In firefox, these 2 behaves exactly the same. Or maybe I don't see what you're trying to fix? Could you give precise description of the problem (not a technical description, a functional one) and step to reproduce please?

gchoqueux:
@autra, in 3dtiles examples, if you move the globe with the mouse click  from under the menu, it doesn’t work. 
In 3dtiles examples, under the menu, the mouse event work with `window.addEventListener` but not with `renderer.domElement.addEventListener`

gchoqueux:
@autra WDYT?

autra:
I see, but I had to guess. What you meant is that the `menuDiv` takes all the page height even though it occupies only a portion of it, and in this zone, the mouse drag is unresponsive (so *below* the menu, not under), that's it?

Screenshots are often clearer:
itowns - 3d-tiles example - firefox developer edition_082

examples/3dtiles.html
```diff
@@ -44,16 +44,8 @@
                     display: none;
                 }
             }
-            #menuDiv {
-                margin : auto auto;
-                width: 100%;
-                padding: 0;
-                height: 100%;
+            #menuDiv {position: absolute; top:0px; margin-left: 0px;}
```
autra: can you reformat this please? (one rule per line, a space after colon)

gchoqueux: @autra done

autra (approved):
LGTM thanks!

PR #713, opened 2018-03-29
perf(renderer): stop creating new vector for each feature

zarov:
The dimension property was instantiated for every feature read in Feature2Texture. It can be computed one time per texture creation instead.


autra (approved):
LGTM, thanks!

PR #714, opened 2018-03-29
feat(core): view.addLayer(url) support (#697)

### LINKED ISSUE #697: view.addLayer(url) support
The examples that use JSON layers currently follow the following pattern : 

```
function addLayerCb(layer) {
    return view.addLayer(layer);
}
promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
```

Should we support `view.addLayer('./layers/JSONLayers/Ortho.json')` directly which could return a promise to the layer ? That would simplify most of the examples !

zarov:
Layers can be added simplier by adding the string, representing the url leading to the configuration of the layer. For example:

```js
view.addLayer('http://geo.server/layers/ortho.json');
```

This resolves #697

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -92,7 +92,7 @@
 
             var menuGlobe = new GuiTools('menuDiv', view, 300);
 
-            itowns.Fetcher.json('layers/JSONLayers/Ortho.json').then(function (result) { return view.addLayer(result) });
+            view.addLayer('layers/JSONLayers/Ortho.json');
 
             // function use :
             // For preupdate Layer geomtry :
--- a/examples/collada.js
+++ b/examples/collada.js
@@ -16,27 +16,23 @@ var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe);
 
 var menuGlobe = new GuiTools('menuDiv');
 
-var promiseElevation = [];
-
 menuGlobe.view = globeView;
 
 function addLayerCb(layer) {
-    return globeView.addLayer(layer).then(function addGui(la) {
-        if (la.type === 'color') {
-            menuGlobe.addImageryLayerGUI(la);
-        } else if (la.type === 'elevation') {
-            menuGlobe.addElevationLayerGUI(la);
-        }
-    });
+    if (layer.type === 'color') {
+        menuGlobe.addImageryLayerGUI(layer);
+    } else if (layer.type === 'elevation') {
+        menuGlobe.addElevationLayerGUI(layer);
+    }
 }
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
-itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb);
+globeView.addLayer('./layers/JSONLayers/Ortho.json').then(addLayerCb);
 // Add two elevation layers.
 // These will deform iTowns globe geometry to represent terrain elevation.
-promiseElevation.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
-promiseElevation.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+globeView.addLayer('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb);
+globeView.addLayer('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb);
 
 exports.view = globeView;
 exports.initialPosition = positionOnGlobe;
--- a/examples/externalscene.js
+++ b/examples/externalscene.js
@@ -11,12 +11,8 @@ setupLoadingScreen(viewerDiv, globeView);
 
 globeView.mainLoop.name = 'external-ML';
 
-function addLayerCb(layer) {
-    return globeView.addLayer(layer);
-}
-
-itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb);
-itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT.json').then(addLayerCb);
+globeView.addLayer('./layers/JSONLayers/Ortho.json');
+globeView.addLayer('./layers/JSONLayers/IGN_MNT.json');
 
 exports.globeView = globeView;
 exports.scene = scene;
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -15,9 +15,6 @@ var miniDiv = document.getElementById('miniDiv');
 // Instanciate iTowns GlobeView*
 var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, { renderer: renderer });
 setupLoadingScreen(viewerDiv, globeView);
-function addLayerCb(layer) {
-    return globeView.addLayer(layer);
-}
 
 // Dont' instance mini viewer if it's Test env
 if (!renderer) {
@@ -49,17 +46,17 @@ if (!renderer) {
     });
 
     // Add one imagery layer to the miniview
-    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(layer) { miniView.addLayer(layer); });
+    miniView.addLayer('./layers/JSONLayers/Ortho.json');
 }
 
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+promises.push(globeView.addLayer('./layers/JSONLayers/Ortho.json'));
 // Add two elevation layers.
 // These will deform iTowns globe geometry to represent terrain elevation.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
+promises.push(globeView.addLayer('./layers/JSONLayers/WORLD_DTM.json'));
+promises.push(globeView.addLayer('./layers/JSONLayers/IGN_MNT_HIGHRES.json'));
 
 exports.view = globeView;
 exports.initialPosition = positionOnGlobe;
--- a/examples/globe_fly.js
+++ b/examples/globe_fly.js
@@ -18,14 +18,11 @@ setupLoadingScreen(viewerDiv, globeView);
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
-function addLayer(layer) {
-    return globeView.addLayer(layer);
-}
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayer));
+promises.push(globeView.addLayer('./layers/JSONLayers/Ortho.json'));
 // Add two elevation layers.
 // These will deform iTowns globe geometry to represent terrain elevation.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayer));
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayer));
+promises.push(globeView.addLayer('./layers/JSONLayers/WORLD_DTM.json'));
+promises.push(globeView.addLayer('./layers/JSONLayers/IGN_MNT_HIGHRES.json'));
 
 exports.view = globeView;
 exports.initialPosition = positionOnGlobe;
--- a/examples/globe_vector.js
+++ b/examples/globe_vector.js
@@ -11,18 +11,15 @@ var viewerDiv = document.getElementById('viewerDiv');
 // Instanciate iTowns GlobeView*
 var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, { renderer: renderer });
 setupLoadingScreen(viewerDiv, globeView);
-function addLayerCb(layer) {
-    return globeView.addLayer(layer);
-}
 
 // Add one imagery layer to the scene
 // This layer is defined in a json file but it could be defined as a plain js
 // object. See Layer* for more info.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
+promises.push(globeView.addLayer('./layers/JSONLayers/Ortho.json'));
 // Add two elevation layers.
 // These will deform iTowns globe geometry to represent terrain elevation.
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
-promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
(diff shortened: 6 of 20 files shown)


autra:
I'm sorry but I'm pretty much against this. The json thing is a workaround and we might delete it one day. For example, in lyon examples, there is no such external json. This mechanism was mainly done because WMTS options objects were very verbose, due to the fact we currently have to harcode tilematrixset there. Actually, I'm against it even in that case, because it prevents people to learn how to configure itowns

During the code sprint, we've even both seen people download json from www.itowns-project.org, which is a real nonsense!!! I now realize that this is dangerous: people seem to think this json mechanism is some kind of official protocol we support, but it's not! We should not take a config object from the network.

A lot of js libraries takes a `options` objects to a `init` methods or similar. Do they offer you to enter a url to get this json? No because the way you get this options is irrelevant from the library POV. 

pinging @mbredif to let you know what I think about that ;-) To elaborate on what you said on the issue

> What I propose is to have the possibility to do what you can currently do with webpack in projects that are not using webpack

This is not *at all* equivalent: webpack imports at build time, whereas this makes a network request. If you want the feature of a module bundler, you need to use a module bundler.

So: I'm really against committing to an API of this sort.

nosy-b:
My POV is that it makes adding a layer much nicer than our previous verbose way.
Maybe we can make the API clearer with a different function to handle json file such as` view.addJSONLayer() `? If it is well documented I don't see why the API could not have this method which offers generic mapping services access, as long as we have _addImageryProvider(url) and co_.

autra:
> our previous verbose way.

Wait. our previous - and only way - has *always* been

```js
view.addLayer(options);
```
Not sure what is verbose here.

I failed to see where we - as a framework - are making stuff clearer for the user by hiding that in json objects hosted somewhere. We should instead delete those json objects, and put the options objects directly in the code of our examples.

That way we could *also* document and explain them, instead of hiding them far away. That's actually something central and *very* important to document, because we want our user to be able to build their own layer - not just to use IGN ones -  and we failed to do so **because** we hid them from ourselves too. 

For a start, if the json were in our code, we could add this comment near the url:

```js
// don't forget to change your api key in production if you use IGN servers
```

Won't that be a good thing to do? Instead of explaining again and again to people why these don't work on their production servers? 

We need to bring these json back in `examples/` and document them.

nosy-b:
I agree on many points
- Yes, it would be good to change some of our examples to show that they can add layer without the need of a JSON file.
- Yes JSON file is an option when having many parameters or for exchanging them easily

So why not create a 'specific' method for the JSON?

nosy-b:
In fact if we have a stable way to handle short config for mapping services (without tileMatrixSetLimits verbose description) I would also be for the NO JSON way (all provider infos declared in the html)
Is it the case? What about the getCapabilities?

autra:
> So why not create a 'specific' method for the JSON?

Answering this question was the point of my previous  comment :-) Please read them again ? In short: the cons of it overweight the pros.

I try again anyway:

> Yes, it would be good to change some of our examples to show that they can add layer without the need of a JSON file.

The fact you said `to show they can add layer without the need of a JSON file` reinforce my point: our api *never* talks about json files. The `json` variable we give to `addLayer` is *not* a file, it's a plain javascript object. Even you are confused man :-) Fetching them for the network was just a (bad) convenience for us, because we know how to configure our layers. But what about our users?

Btw, from your previous comment

> If it is well documented I don't see why the API could not have this method which offers generic mapping services access

It's not offering a "generic mapping services access". It's a conf object stored in a distant file!

These json files have made things unclear for everybody. A good example of how we should structure our example is actually from OpenLayers, see https://openlayers.org/en/latest/examples/wmts-ign.html, starting from `var matrixIds = [];`. This is how we should construct the options object of our example (and notice that they're doing it without `getCapabilities`. It is an orthogonal matter)

autra (changes requested):
I expressed my disagreement, let's make things official ;-) Sorry for that, but that's not an usage I want to promote for the aforementioned reasons.

nosy-b:
We could follow open layer way (https://openlayers.org/en/latest/examples/wmts.html) and add a method to compute the tileGrid in one line from the projection to save some lines for example.

I let you discuss on the implementation, I agree on the fact that it's better to expose the config object but it's important to offer a very short and simple way (a few lines) to add mapping services.

autra:
> it's important to offer a very short and simple way (a few lines) to add mapping services.

agreed, it's important to offer a simple way to add *any* mapping services ;-)

elemoine:
Placing layers definitions in JSON files is an unusual and strange pattern. So I'd certainly not promote that pattern by making the library support it! Please just don't :)

gchoqueux:
Imagine an application using iTowns, which proposes to save / load its project (added layers, layer states). How would you do it?

autra:
> Imagine an application using iTowns, which proposes to save / load its project (added layers, layer states). How would you do it?

Plenty of options, [JSON.parse and JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON) for instance. Or you can generate xml with some template strings...

Then, where you store the layer definition depends on your business logic (not iTowns' concern): it could be remotely with a web api, or on a file, or on a cookie, localStorage, indexedDB or...

elemoine:
> Imagine an application using iTowns, which proposes to save / load its project (added layers, layer states). How would you do it?

Assuming that you have somehow saved the layer state in a `layer.json` file then, yes, you would use this to load it back:

```js
fetch('mylayer.json').then(resp => {
    var layerOptions = resp.json();
    view.addLayer(layerOptions);
});
```

That's a specific use-case, that can easily be handled at the application level. And the application developer may want to store the layer definition/states using a totally different format, or even using XML instead of JSON.

That's just an application-specific thing, completely out of the scope of iTowns IMO.

gchoqueux:
Adding a layer with a parameter file is used by [mapbox](https://www.mapbox.com/mapbox.js/api/v3.1.1/l-mapbox-tilelayer/)
Note that the layer also has a method for [generating this json file](https://www.mapbox.com/mapbox.js/api/v3.1.1/l-mapbox-tilelayer/#section-tilelayer-gettilejson)

autra:
> Adding a layer with a parameter file is used by mapbox

Yes, it's an open specification called TileJSON, see my comment here: https://github.com/iTowns/itowns/pull/715#issuecomment-378166203. We could support it, I have no strong opinion either way.

>  Note that the layer also has a method for generating this json file

If we support TileJSON, then it is a requirement yes, but I think I prefer to name it toJSON, to allow [standard JSON.stringify to use it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior).

gchoqueux:
I propose that this PR offers a method similar to `L.mapbox.tileLayer` for iTowns with a specific format (with documentation). And then, eventually, add the `tilejson` support in new PR. 
Also add new explicit examples (with documentation) `addLayer` by objects, as in PR 715.

elemoine:
> I propose that this PR offers a method similar to L.mapbox.tileLayer for iTowns with a specific format (with documentation)

I am not an iTowns developer, but as a user, I would personally be against that. To me this goes against having a minimal, simple and consistent API, which I'd love iTowns have.

And if both `addLayer(options)` and `addLayer(url)` were supported would you apply the same rule to all the functions and constructors that receive a config object as an argument? For example would you change the `View` constructor to support both `View(crs, viewDiv, options)` and `View(crs, viewDiv, url)`?

Just my 2 cents.

gchoqueux:
So, we will use a new `addLayerFromXXX` function and we will not change `addLayer`

autra:
> So, we will use a new addLayerFromXXX function and we will not change addLayer

What does it change from the current proposition that I r-'d?

elemoine:
> So, we will use a new addLayerFromXXX function and we will not change addLayer

It sounds like you really want it :)

What if there are users who prefer XML or YAML? Are you going to add `addLayerFromJSON`, `addLayerFromXML`, and `addLayerFromYAML`?

`fetch(url).then(resp => { view.addLayer(resp.json()); })` is so easy for users who want to store their layer definitions in JSON files that I fail to understand why you would add that to the library.

I would really recommend to keep it minimal and simple.

zarov:
> What does it change from the current proposition that I r-'d?

It will keep the current `addLayer` method as is, and it will give a new method. If the method is well documented, I don't how it could failed the user. We could even specify that there are other ways to add a layer in each example using the `addLayerFromJSON` method.

> What if there are users who prefer XML or YAML? Are you going to add `addLayerFromJSON`, `addLayerFromXML`, and `addLayerFromYAML`?

Why not ? If it is asked, I see no problem in adding it.

> `fetch(url).then(resp => { view.addLayer(resp.json()) })` is so easy for users who want to store their layer definitions in JSON files that I fail to understand why you would add that to the library.

And yet so complicated for people who don't know the fetch API and would only want to add a layer to their view (because the target of this library is imho not only full time developers like us here).

> I would really recommend to keep it minimal and simple.

If we are going to take this road, a lot of things needs to be trimmed off iTowns.

elemoine:
> Why not ? If it is asked, I see no problem in adding it.

Because to me it's clobbering the library with things that can easily be done at the application level.

Again if you add `addLayerFromJSON` will you also add `ViewFromJSON` for users who want to store their view options in a JSON file?

> And yet so complicated for people who don't know the fetch API and would only want to add a layer to their view (because the target of this library is imho not only full time developers like us here).

If the examples did not use the "store layer definitions in JSON files" pattern, people who are not "full-time developers like us" would not even think about placing their layer definitions in JSON files. Beginners just use what they see in the official examples.

> If we are going to take this road, a lot of things needs to be trimmed off iTowns.

At least a few things yes :)

Just friendly giving my personnal opinion here, with the experience I have working on OpenLayers and other JS librairies. I'll stop here I guess :)

nosy-b:
I think that there is no perfect solution :)

However the current state is not ideal for beginners/intermediate users. (@elemoine: Mapbox seems to use what @zarov initially proposed https://www.mapbox.com/mapbox.js/api/v3.1.1/l-mapbox-tilelayer/)

iTowns must stay a rather light API but it doesn't mean it can't use some configuration file for animation for example or as we talk here, layer information.
We created initially the JSON to avoid having dirty examples full of tileMatrixSet caused by French Geoportal not covering the whole world. -> We should present more examples without the use of JSON with more simple settings (using world coverage OSM for example) to make it clearer for users.
However I'm not against a helper function that uses preconfigured layer files following a documented format

If we don't agree on this, as it's an API decision we could end up by a vote :)

autra:
> Why not ? If it is asked, I see no problem in adding it.

And then addLayerFromLocalStorage, addLayerFromCookie, addLayerFromPigeonVoyageur ;-) We want to avoid interface pollution here. We provide an API, which takes an object as parameter. We can't make it simpler.

We have to draw clear boundaries on what iTowns is or is not. I'll make the framework clearer and easier to use. Itowns is not a fetch library.

> the target of this library is imho not only full time developers like us here

We certainly do target js-developers ("itowns is a framework..."), whether full-time or as a hobby. I expect any js dev to know (or to be able to find) about a way to get file from the internet from a browser context, whether `Fetch` or `XMLHttpRequest`. It is one of the first thing you learn when you start js dev. If they can't do that, will you expect them to build a full itowns scene? It's just not realistic.

On this particular instance, I see no difficulty even for beginners.

But anyway, *how* putting the layer definition behind the network will make it simpler for beginners??? They'll still have to write it anyway.

> (@elemoine: Mapbox seems to use what @zarov initially proposed https://www.mapbox.com/mapbox.js/api/v3.1.1/l-mapbox-tilelayer/)

@nosy-b It's the third time I answer to that in the last 2 weeks. It's the second time your link appears in this thread and we already talked about it. Please read the thread before commenting, thanks :-)

Mkonini:
"itowns is a framework..." everyone agrees with this definition?

gchoqueux:
> Itowns is not a fetch library

How do you justify the presence in itowns of [Fetcher.js ](https://github.com/iTowns/itowns/blob/master/src/Provider/Fetcher.js)

it seems that itowns offers simplifications for loading and **parsing** different type format files.
Why then not offer a simple fecth and given the responsibility to the user to parse his data?

Jeremy-Gaillard:
> "itowns is a framework..." everyone agrees with this definition?

Well, it's in the readme since the beginning of the project and, if I recall correctly, was agreed upon in one of the first iTowns2 meeting.

vpicavet:
> "itowns is a framework..." everyone agrees with this definition?

This is probably not the place for such a question, but the main itowns-project pages clearly states "What is it? iTowns is a web framework written...". 

Should anyone want to build something else, then start a side-project.

nosy-b:
@autra The Mapbox remark was targeted at Eric as it was referenced to get HIS opinion emphasizing the fact that it's not an unusual way of dealing with layers but thanks for your delicate remark.

Following the remark of @Mkonini, I think we should agree on what are the limits of iTowns as some of you really focus on javascripters and some others on GIS newbies.

autra:
> How do you justify the presence in itowns of Fetcher.js 

It must disappear from the api and be only used internally.

> it seems that itowns offers simplifications for loading and parsing different type format files.

If I recall, all of these format are standards. I recall that I'm not against TileJSON support. We're going in circle here :-)

autra:
> some others on GIS newbies.

"GIS newbies" don't need a framework, they need a configurable app. They need something built on iTowns that will provide all of this functionalities and make their life easier. iTowns has been made to make these kind of projets possible, I suggest you start one. I might even contribute, and there I'll totally agree to add all the fancy features that belongs to an app, including this one.

A framework - by definition - targets developers.

Mkonini:
The question is obviously : what do we want to put inside.
HTH

vpicavet:
> Following the remark of @Mkonini, I think we should agree on what are the limits of iTowns as some of you really focus on javascripters and some others on GIS newbies.

@nosy-b again, probably not the best place to discuss this. But a framework is indeed targeted at Javascript developers, whereas a web application may be targeted at GIS users and newbies. As stated in the definition of iTowns, we build a framework, we work for developers. 

Should we wish to target non-developers, or non-JS proficient users, then let's build "itowns-app" and put every single facilitator or end-users oriented feature there. 

But lowering the quality of the iTowns framework for this reason is a no-go and a mistake we will not make.

nosy-b:
Still, we need to agree on what we want inside the framework. It is **normal** that the borders are hard to define ( ex the loading screen? ;)) so that we discuss it.
A framework helps you to build applications, but how much should it helps is something we need to agree so we don't have too many threads like this one :) (In another place yes, @vpicavet )

elemoine:
Agree with @nosy-b that calling it a "framework" doesn't automatically define its boundaries. That'd be way too easy :)

@nosy-b, L.mapbox.tileLayer can indeed work with a URL. But that URL points to a TileJSON doc. TileJSON is a tile standard so that's a completely different story. Also it's to be noted that Mapbox GL JS is much more focused than OpenLayers or iTowns. TileJSON and Vector Tiles are part of their focus, but things like WMS and WMTS are not. That is why they can take shortcuts and decide that tileLayer(url) works with TileJSON.

And to me addLayer(url) does not help users build their apps. It just clobbers the API and broadens the API surface for no benefit to the user.

Also, when I look at the current examples I feel that there are much more important things to do than adding addLayer(url). To me addLayer(url) is just a false impression of easyness.

zarov:
Well this PR won't be merged, and the discussion is diverging to another topic, so I'm closing this PR. For discussing the initial topic, there is still #697 and #715, otherwise a new issue can be opened.

Note: deleting the branch but keeping the commit reference for posterity: https://github.com/iTowns/itowns/pull/714/commits/505fa2e1f6b86d1e16709dbe0f02c9c18ce438a2

PR #715, opened 2018-03-30
WIP: refacto(examples): stop hiding our config objects and document them

autra:
Followup on the discussion on #714:

## Description

Creating a config object to add a layer is an essential piece to understand when you want to use iTowns and plug your own data in.

Hiding them in json files far away from the code had 2 very bad consequences:
- they weren't documented
- some mistakes and discrepancies between protocols have been forgotten, see the changes in WMTS_Provider

This PR is a proof-of-concept of putting our config back in examples. They are now documented.

To further my idea, we can then autogenerate this kind of webpages from our examples (based on the work @peppsac did on docco): http://37.187.164.233/itowns-docs2/docs/globe.html

### DIFF
--- a/examples/globe.js
+++ b/examples/globe.js
@@ -1,5 +1,5 @@
 /* global itowns, document, renderer, setupLoadingScreen */
-// # Simple Globe viewer
+// Simple Globe viewer
 
 // Define initial camera position
 var positionOnGlobe = { longitude: 2.351323, latitude: 48.856712, altitude: 25000000 };
@@ -19,7 +19,7 @@ function addLayerCb(layer) {
     return globeView.addLayer(layer);
 }
 
-// Dont' instance mini viewer if it's Test env
+// Don't instance mini viewer if it's Test env
 if (!renderer) {
     miniView = new itowns.GlobeView(miniDiv, positionOnGlobe, {
         // `limit globe' subdivision level:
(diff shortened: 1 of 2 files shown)


examples/globe.js
```diff
@@ -55,7 +55,166 @@ if (!renderer) {
...
+        // This defines the tilematrixset of our WMTS layer. Supported types are currently PM and WGS84G.
+        // More to be supported in the futur.
+        "tileMatrixSet": "PM",
+    }
+};
+
+// now we need to set the WMTS limits. This should be compatible from what GetCapabilities returns.
+// (cf http://wxs.ign.fr/va5orxd0pgzvq3jxutqfuy0b/geoportail/wmts?service=wmts&version=1.0.0&request=GetCapabilities)
+orthoLayerOptions.options.tileMatrixSetLimits = {
+    "2": {
```
autra, line 99: For this verbose definition, we can add a auto-hide functionality to the generated webpage. It could only display like 10 lines by default, then the user can unfold it if he wants to see everything.

src/Provider/WMTSProvider.js
```diff
@@ -28,16 +28,16 @@ function preprocessDataLayer(layer) {
     }
 
     if (layer.protocol === 'wmts') {
+        layer.version = layer.version || '1.0.0';
```
autra, line 31: This change is to stay coherent with WMS. It probably deserves its own PR.

mbredif:
- I agree that json layers were not properly documented and that it is causing confusion (eg: geoportail key issues)
- Adding the functionality to load a json by url instead of reading a json object directly is rather simplifying the maintenance (#714 removes most of the fetchers from examples and just adding [a few lines](https://github.com/iTowns/itowns/blob/505fa2e1f6b86d1e16709dbe0f02c9c18ce438a2/src/Core/Prefab/GlobeView.js#L306-L312) in the views ). The functionality exists elsewhere (eg [`L.mapbox.tileLayer(id|*url*|tilejson, options)`
](https://www.mapbox.com/mapbox.js/api/v3.1.1/l-mapbox-tilelayer/)), so it might not be completely nonsensical.
- I am however not sure that copy-pasting the documentation for the json layers in all examples is the simplest fix...

Could we agree on a design where json configs are properly documented in one place and where docs and comments of examples that are focusing on other functionalities are just 1. linking to the relevant documentation and 2. have minimal self-descriptive code for adding layers (one-liners if possible) ?

Just a proposition, but maybe a helper js file could help build those json configs and centralize the documentation of these, with at least one example that does not use raw json layers to show that it is possible.

autra:
> have minimal self-descriptive code for adding layers (one-liners if possible)

Could you give me an explanation why you are so focused on one-liners? oneliners are good for expressivity but they are not the Holy Graal of coding. They are bad when you need to explain something (at least in this context, where examples are for people to learn), because they do make things less understandable.

> Adding the functionality to load a json by url instead of reading a json object directly is rather simplifying the maintenance (#714 removes most of the fetchers from examples and just adding a few lines in the views ). 

no it does not simplify the maintenance, really. It more or less changes nothing. It also causes us to forget about them, see the changes in this PR.

> The functionality exists elsewhere (eg L.mapbox.tileLayer(id|*url*|tilejson, options)
), so it might not be completely nonsensical.

Ok, I see where you come from. You cannot compare [tilejson, which is an open format and has its specification](https://github.com/mapbox/tilejson-spec) to our custom object config we pass to `addLayer` though. The former is more comparable to our WMS support. Btw, we could choose to support this specification, and maybe realize later we don't need our custom format after all and ditch it. I'd be happy to r+ any PR in this direction (providing it's a good idea strategically to support tilejson, I have no opinion yet on the subject).

> I am however not sure that copy-pasting the documentation for the json layers in all examples is the simplest fix...

Examples are too opaque. An examples should be able to be read easily by an outsider. One of their goals is to allow people to learn. Maybe there's too much layers added on most examples. Most of them only need 2 (elevation and color) to demonstrate the concept. So I still think it's a lot better to do the config directly into the configuration than what we do currently. Have you looked at the example of generated doc I've put in the description?

That being said, I do totally agree with this::

>  maybe a helper js file could help build those json configs and centralize the documentation of these

For `tilematrixSetlimits`, at least, I had that in mind too. I'll think about it.

Also what can help going nearer a "oneliner" is to have sensible defaults for most values. For examples, `layer.format` should probably default to `image/png` in most cases. Tilematrixset should probably default to `wgs84` (but we support only `wgs84g` atm, that's a problem, not sure of the differences between the two) and we could simplify the config. For instance, in most cases, we could generate these limits from min and max zoom level (and maybe extent).

elemoine:
I'd like to chime in. I am just getting started with iTowns. And as a beginner I've looked at the examples, attempting to understand them and making sense of the iTowns API.

And I do agree that the examples are too opaque right now. So any effort to make them less opaque would be greatly appreciated in my opinion. Making them self-contained, with all their js code in a single js file, is a step in the right direction.

I also agree that having layer definitions in JSON files fetched using itowns.Fetcher.json is confusing and misleading. Some people could even think that they need to put their layer definitions in JSON files. (I've read tutorials where they repeat exactly that!)

To ease the reading of examples their js file could be organized as follows:

```js
// main
var view = new itowns.View(…);
var orthoLayer = createOrthoLayer();
view.addLayer(orthoLayer);

// layer definitions
function createOrthoLayer() {
    return {
       …
    };
}
```

In this way the layer definitinos would not pollute the example's "main" code. Just an idea.

nosy-b:
@elemoine Thank you to share your opinion it's always good to have a certain number of them!

We should handle 'better' default minimal _WMTS_ description (iTowns should work even if _TileMatrixSet_ are not set for not-all world coverage services) and, yes, ease with methods the description of _TileMatrixSet&co_ so we could work without external _JSON_ file and keep examples tight.

autra:
> We should handle 'better' default minimal WMTS description (iTowns should work even if TileMatrixSet are not set for not-all world coverage services)

I'm not sure about that. Did you see something stating WMTS clients are expected to work with incomplete tilematrixsets in the specifications? If yes, then we need to implement it of course.

examples/globe.js
zarov: You could also let the first one expanded, and write in a more compact version the others:
`"3": { "minTileRow": 0, "maxTileRow": 8, "minTileCol": 0, "maxTileCol": 8 }`

zarov:
Do you still want to push those changes ? I'm okay with updating examples gradually, but I would like to see the generated tutorials first. Could you do that before ? Thanks

autra:
>  I would like to see the generated tutorials first

You have an example here: http://37.187.164.233/itowns-docs2/docs/globe.html

I'll change the style to better match the rest of the website style of course (and maybe add a collapse style for the long code part or other improvement)

autra:
> Do you still want to push those changes ?

Yes I do :-) but this is not ready of course.

PR #716, opened 2018-04-03
fix(core): wrong zoom, row and col definition in extent

zarov:
The zoom, row and col properties, when the CRS is tiled, were
incorrectly defined, thus setting any of these properties later
impossible.

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -24,26 +24,13 @@ function Extent(crs, ...values) {
 
     if (_isTiledCRS(crs)) {
         if (values.length == 3) {
-            this._zoom = values[0];
-            this._row = values[1];
-            this._col = values[2];
+            this.zoom = values[0];
+            this.row = values[1];
+            this.col = values[2];
 
-            if (this._zoom < 0) {
+            if (this.zoom < 0) {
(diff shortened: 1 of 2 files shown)


src/Core/Geographic/Extent.js
```diff
@@ -34,16 +34,22 @@ function Extent(crs, ...values) {
...
+                    set: (r) => { this._row = r; },
+                });
             Object.defineProperty(this,
                 'col',
-                { get: () => this._col },
-                { set: (c) => { this._col = c; } });
+                {
+                    get: () => this._col,
+                    set: (c) => { this._col = c; },
+                });
```
autra: This got me wondering: why don't we directly use property access? I mean these getters and setters don't add logic or checks of any sort...

zarov: I think it is simply because those properties can be available only when the CRS is tiled. But we could indeed write `this.zoom = this._zoom`.

zarov: I'm rewriting all the file in fact, wait for the update

zarov:
I removed the "private" properties in favor of simple "public" ones.

peppsac (approved):
LGTM, thanks!

PR #717, opened 2018-04-09
refactor(core): move commonAncestorLookup to TileMesh

zarov:
Move it from all the views to the prototype of TileMesh.

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -237,4 +237,31 @@ TileMesh.prototype.getZoomForLayer = function getZoomForLayer(layer) {
     }
 };
 
+/**
+ * Search for a common ancestor between this tile and another one. It goes
+ * through parents on each side until one is found.
+ *
+ * @param {TileMesh} tile
+ *
+ * @return {TileMesh} the resulting common ancestor
+ */
+TileMesh.prototype.findCommonAncestor = function findCommonAncestor(tile) {
+    if (!tile) {
+        return undefined;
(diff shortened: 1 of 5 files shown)


peppsac (commented):
Could you add a couple of unit-tests for this please?

Also the newly introduced method should start with a verb ( `findLowestCommunAncestor`?)

zarov:
No problem, doing it.

zarov:
Updated

src/Core/TileMesh.js
```diff
@@ -237,4 +237,31 @@ TileMesh.prototype.getZoomForLayer = function getZoomForLayer(layer) {
...
 
+/**
+ * Search for a common ancestor between this tile and another one. It goes
+ * through parents on each side until one is found.
+ *
+ * @param {TileMesh} tile
+ *
+ * @return {TileMesh} the resulting common ancestor
+ */
+TileMesh.prototype.findCommonAncestor = function findCommonAncestor(tile) {
```
gchoqueux, line 248: is that [this code](https://github.com/iTowns/itowns/pull/719/files#diff-6b60bae2240ab556c955b4eb0a07a12eR189)  is similar in ancestor search with different criteria (id and name)

zarov: It is similar, but I don't see how it could be mutualized: pointcloud is a very specific usecase.

peppsac (approved):
LGTM, thanks!

PR #718, opened 2018-04-09
fix(examples): unhide dat.gui elements when loadingscreen is finished

peppsac:
On at least one example (pointcloud), the dat.gui elements was still
hidden after the loading screen ended.

### DIFF
--- a/examples/css/loading_screen.css
+++ b/examples/css/loading_screen.css
@@ -1,5 +1,5 @@
 #itowns-loader span {
-    display: inline-block;
+  display: inline-block;
   color: #404e73;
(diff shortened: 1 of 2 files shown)


zarov:
Elements in dat.gui are not visible anymore in at least the `globe` example.

peppsac:
@zarov you're right, I missed this issue. I modified the commit to use `z-index` property, to avoid messing with dat.gui `display` property.

zarov (approved):
All good now, good to merge (with squash, I removed the unused `document` eslint reference).

PR #719, opened 2018-04-09
feat(pointcloud): use 'changeSource' mechanism to avoid useless work

peppsac:
Implements a simple logic to avoid updating the full layer if we
already know that the result won't change.

This feature is already implemented for tile based layers
(see 3f340b9ed610009f197f0d43982faf01e899d3ad).

### DIFF
--- a/src/Process/PointCloudProcessing.js
+++ b/src/Process/PointCloudProcessing.js
@@ -163,16 +163,48 @@ function markForDeletion(elt) {
 }
 
 export default {
-    preUpdate(context, layer) {
-        // TODO: use changeSource
-        layer.counters = {
-            pointCount: 0,
-            displayedCount: 0,
-        };
-
+    preUpdate(context, layer, changeSources) {
         // Bail-out if not ready
         if (!layer.root) {
             return [];
+        }
+
+        if (changeSources.has(undefined) || changeSources.size == 0) {
+            return [layer.root];
+        }
+
+        // lookup lowest common ancestor of changeSources
+        let commonAncestorName;
+        for (const source of changeSources.values()) {
+            if (source.isCamera) {
+                // if the change is caused by a camera move, no need to bother
+                // to find common ancestor: we need to update the whole tree:
+                // some invisible tiles may now be visible
+                return [layer.root];
+            }
+            if (source.obj === undefined) {
+                continue;
+            }
+            // filter sources that belong to our layer
+            if (source.obj.isPoints && source.obj.layer == layer.id) {
+                if (!commonAncestorName) {
+                    commonAncestorName = source.name;
+                } else {
+                    let i;
+                    for (i = 0; i < Math.min(source.name.length, commonAncestorName.length); i++) {
+                        if (source.name[i] != commonAncestorName[i]) {
+                            break;
+                        }
+                    }
+                    commonAncestorName = commonAncestorName.substr(0, i);
+                    if (commonAncestorName.length == 0) {
+                        break;
+                    }
+                }
+            }
+        }
+        if (commonAncestorName) {
+            return [layer.root.findChildrenByName(commonAncestorName)];
         }
 
         // Start updating from hierarchy root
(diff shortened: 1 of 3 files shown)


src/Process/PointCloudProcessing.js
```diff
@@ -163,18 +163,49 @@ function markForDeletion(elt) {
...
+                        }
+                    }
+                    commonAncestorName = commonAncestorName.substr(0, i);
+                    if (commonAncestorName.length == 0) {
+                        break;
+                    }
+                }
+            }
+        }
+        if (commonAncestorName && commonAncestorName.length > 0) {
```
elemoine: (neat) I guess the second test (after `&&`) is useless, so this could be simplified to `if (commonAncestorName)`.

peppsac: Indeed: fixed.

src/Process/PointCloudProcessing.js
```diff
@@ -163,18 +163,49 @@ function markForDeletion(elt) {
 }
 
 export default {
-    preUpdate(context, layer) {
-        // TODO: use changeSource
-        layer.counters = {
-            pointCount: 0,
```
elemoine, line 169: I see that the `pointcloud.js` and `pointcloud_globe.js` examples refer to `pointcloud.counters.pointCount`. Wouldn't that be a problem if you removed that count?

elemoine: Also, if you just keep `displayedCount` you may want to go with `layer.displayCount` instead of having an intermediary `layer.counters` object.

peppsac: Fixed.

elemoine (approved):
Looks good to me.

elemoine:
Just a quick unrelated note:

I note that iTowns often dynamically adds properties to objects. For example, `layer.displayedCount` in this patch. For performance it is better to set properties at object construction time, and not dynamically add properties afterwards. See for example the Hidden Classes section in this [HTML5 Rocks article](https://www.html5rocks.com/en/tutorials/speed/v8/). For that reason I think it's a good thing to more strictly type things, using specific object constructors/types. For example I think the Layer constructor should look like this:

```js
function  Layer(options) {
    this.name = options.name;
    this.projection = options.projection;
    …
    this.displayedCount = 0;
}
```

It also makes the code more readable, and easier to document.

Just my 2 cents, and sorry for the unrelated comment here :)

zarov (approved):
LGTM, great improvement. 

~~A question though (that have nothing to do with this PR): why is the node name is stored as a string ? Wouldn't be more efficient to store it with something as an octal ? I think it would simplify operation on it, and it could be simply converted to a string when needed.~~ Nevermind, tried it up, not efficient.

zarov:
@elemoine I agree completely on this, there could be improvement made here.

src/Process/PointCloudProcessing.js
```diff
@@ -163,18 +163,49 @@ function markForDeletion(elt) {
...
+        // lookup lowest common ancestor of changeSources
+        let commonAncestorName;
+        for (const source of changeSources.values()) {
+            if (source.isCamera) {
+                // if the change is caused by a camera move, no need to bother
+                // to find common ancestor: we need to update the whole tree:
+                // some invisible tiles may now be visible
+                return [layer.root];
+            }
+            if (source.obj === undefined) {
```
gchoqueux, line 185: I don't remember where this `obj` came from?
it's an Object3D?

src/Process/PointCloudProcessing.js
```diff
@@ -163,18 +163,49 @@ function markForDeletion(elt) {
...
+                continue;
+            }
+            if (source.obj.isPoints && source.obj.layer == layer.id) {
```
gchoqueux: you could add comment to explain this part with `commonAncestorName`

src/Provider/PointCloudProvider.js
```diff
@@ -206,6 +219,8 @@ export default {
             // eslint-disable-next-line no-console
             console.log('LAYER metadata:', root);
             layer.root = root;
+            root.findChildrenByName = findChildrenByName.bind(root, root);
```
gchoqueux, line 222: why you use `bind`?

elemoine: To do currying and pre-set the first argument to `root`.

src/Process/PointCloudProcessing.js
peppsac: See: https://github.com/iTowns/itowns/blob/master/src/Process/PointCloudProcessing.js#L228

src/Process/PointCloudProcessing.js
peppsac: done

PR #720, opened 2018-04-10
chore(examples): simplify the multiglobe example

elemoine:
## Description

This PR brings a few simplifications to the multiglobe example.

In particular it removes the code that adds the globes' 3D objects to the scene, as this is already done by `View.addLayer`.

## Motivation and Context

Examples need simplifications. This is a tiny step in that direction. (Also me getting more familiar with iTowns…)

### DIFF
--- a/examples/multiglobe.html
+++ b/examples/multiglobe.html
@@ -83,8 +83,11 @@
             });
 
             // Globe animation
+            // Exchange postion and scale of globes in a 1 second animation
             var t = 0;
             function animator (dt, ignore) {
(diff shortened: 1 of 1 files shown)


zarov (approved):
LGTM, simplification well spotted ! If you really want to go further, you could also comment the `animator` function imho.

Thanks for your first PR anyway !

elemoine:
> you could also comment the animator function imho

Agree. Done with an additional commit.

zarov:
Thanks !

PR #722, opened 2018-04-11
chore(examples): no need to pass a parent layer

elemoine:
## Description

When adding a "geometry" layer using view.addLayer there's no need to pass a parent layer.

And in the case of a globe view (globe_wfs_extruded.js) view.tileLayer is undefined anyway, the "tile layer" being view.wgs84TileLayer not view.tileLayer.

## Motivation and Context

Make the examples simpler when possible.


zarov (approved):
LGTM, thanks

PR #723, opened 2018-04-11
feat(parsers) Parser registry

mbredif:
## Description
This PR introduces a registry of format parsers in the scheduler, similar to the protocol registry that is already there.

## Motivation and Context
The goal is to decouple format parsing from the providers, so that protocols, formats and styling may be mixed and matched (instead of current providers that hardcode formats).

(extracted from #705)

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -5,6 +5,8 @@
  */
 
 import PriorityQueue from 'js-priority-queue';
+import CancelledCommandException from './CancelledCommandException';
+
 import WMTSProvider from '../../Provider/WMTSProvider';
 import WMSProvider from '../../Provider/WMSProvider';
 import TileProvider from '../../Provider/TileProvider';
@@ -14,7 +16,16 @@ import PointCloudProvider from '../../Provider/PointCloudProvider';
 import WFSProvider from '../../Provider/WFSProvider';
 import RasterProvider from '../../Provider/RasterProvider';
 import StaticProvider from '../../Provider/StaticProvider';
-import CancelledCommandException from './CancelledCommandException';
+
+import GpxParser from '../../Parser/GpxParser';
+import XbilParser from '../../Parser/XbilParser';
+import GeoJsonParser from '../../Parser/GeoJsonParser';
+import PntsParser from '../../Parser/PntsParser';
+import B3dmParser from '../../Parser/B3dmParser';
+import PotreeBinParser from '../../Parser/PotreeBinParser';
+import PotreeCinParser from '../../Parser/PotreeCinParser';
+// import GLTFLoader from '../../Parser/GLTFLoader';
+// import LegacyGLTFLoader from '../../Parser/LegacyGLTFLoader';
 
 var instanceScheduler = null;
 
@@ -80,12 +91,14 @@ function Scheduler() {
     this.hostQueues = new Map();
 
     this.providers = {};
+    this.parsers = {};
 
     this.maxConcurrentCommands = 16;
     this.maxCommandsPerHost = 6;
 
     // TODO: add an options to not instanciate default providers
     this.initDefaultProviders();
+    this.initDefaultParsers();
 }
 
 Scheduler.prototype.constructor = Scheduler;
@@ -103,6 +116,20 @@ Scheduler.prototype.initDefaultProviders = function initDefaultProviders() {
     this.addProtocolProvider('wfs', WFSProvider);
     this.addProtocolProvider('rasterizer', RasterProvider);
     this.addProtocolProvider('static', StaticProvider);
+};
+
+Scheduler.prototype.initDefaultParsers = function initDefaultParsers() {
+    // Register all parsers
+    this.addFormatParser('gpx', GpxParser);
+    this.addFormatParser('xbil', XbilParser);
+    this.addFormatParser('application/json', GeoJsonParser);
+    this.addFormatParser('geojson', GeoJsonParser);
+    this.addFormatParser('pnts', PntsParser);
+    this.addFormatParser('b3dm', B3dmParser);
+    this.addFormatParser('bin', PotreeBinParser);
+    this.addFormatParser('cin', PotreeCinParser);
+    // this.addFormatParser('gltf', GLTFLoader);
+    // this.addFormatParser('gltf-legacy', LegacyGLTFLoader);
 };
 
 Scheduler.prototype.runCommand = function runCommand(command, queue, executingCounterUpToDate) {
@@ -254,6 +281,18 @@ Scheduler.prototype.addProtocolProvider = function addProtocolProvider(protocol,
  */
 Scheduler.prototype.getProtocolProvider = function getProtocolProvider(protocol) {
     return this.providers[protocol];
+};
+
+
+Scheduler.prototype.addFormatParser = function addParser(format, parser) {
+    if (typeof (parser.parse) !== 'function') {
+        throw new Error(`Can't add parser for ${format}: missing a parse function.`);
+    }
+    this.parsers[format] = parser;
+};
+
+Scheduler.prototype.getFormatParser = function getFormatParser(format) {
+    return this.parsers[format];
 };
 
 Scheduler.prototype.commandsWaitingExecutionCount = function commandsWaitingExecutionCount() {
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -21,6 +21,7 @@ import GpxParser from '../../Parser/GpxParser';
 import GeoJsonParser from '../../Parser/GeoJsonParser';
 import PotreeBinParser from '../../Parser/PotreeBinParser';
 import PotreeCinParser from '../../Parser/PotreeCinParser';
+import PntsParser from '../../Parser/PntsParser';
 
 var instanceScheduler = null;
 
@@ -115,12 +116,11 @@ Scheduler.prototype.initDefaultProviders = function initDefaultProviders() {
 
 Scheduler.prototype.initDefaultParsers = function initDefaultParsers() {
     // Register all parsers
-    this.addFormatParser('gpx', GpxParser);
-    this.addFormatParser('application/json', GeoJsonParser);
-    this.addFormatParser('geojson', GeoJsonParser);
-    this.addFormatParser('json', GeoJsonParser);
-    this.addFormatParser('bin', PotreeBinParser);
-    this.addFormatParser('cin', PotreeCinParser);
+    this.addFormatParser(GpxParser);
+    this.addFormatParser(GeoJsonParser);
+    this.addFormatParser(PotreeBinParser);
+    this.addFormatParser(PotreeCinParser);
+    this.addFormatParser(PntsParser);
 };
 
 Scheduler.prototype.runCommand = function runCommand(command, queue, executingCounterUpToDate) {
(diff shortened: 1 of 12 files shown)


src/Core/Scheduler/Scheduler.js
```diff
@@ -175,6 +202,18 @@ Scheduler.prototype.getProtocolProvider = function getProtocolProvider(protocol)
     return this.providers[protocol];
 };
 
+
+Scheduler.prototype.addFormatParser = function addParser(format, parser) {
```
zarov: Why not take an array as `format`, so we could batch a lot of format into one parser ?

src/Core/Scheduler/Scheduler.js
```diff
@@ -14,7 +16,16 @@ import PointCloudProvider from '../../Provider/PointCloudProvider';
...
-import CancelledCommandException from './CancelledCommandException';
+
+import GpxParser from '../../Parser/GpxParser';
+import XbilParser from '../../Parser/XbilParser';
+import GeoJsonParser from '../../Parser/GeoJsonParser';
+import PntsParser from '../../Parser/PntsParser';
+import B3dmParser from '../../Parser/B3dmParser';
+import PotreeBinParser from '../../Parser/PotreeBinParser';
+import PotreeCinParser from '../../Parser/PotreeCinParser';
+// import GLTFLoader from '../../Parser/GLTFLoader';
```
zarov: Do you plan to support it before merging ? If not, please remove those comments.

zarov (commented):
Even if it doesn't really do nothing alone, as you said, I'm interested in having this merged. Thanks for this !

autra:
What I meant by "parser needs more work" in the other PR: as we create a registry, we should use it in other providers instead of referencing parsers directly. This will allow to be able to use different parsers in a provider according to the needed output format. This work should be done in this PR to avoid an inconsistent state in the codebase (I don't expect it to be too complicated anyway). Thanks!

zarov:
>  we should use it in other providers instead of referencing parsers directly.

Even better: why not moving them out of providers, and have something like this:
```js
return provider.executeCommand(cmd)
    .then(blob => parser.parse(blob, options))
    .then((result) => {
...
```
The problem here could be that we don't have anything to parse for texture, but a "fake" parser for image could be created: `parser = { parse: blob => blob };`

I can also see (later on) having the `Fetcher` getting introduced between the first and second line ;)

zarov:
I quickly added my thoughts here https://github.com/zarov/itowns/commit/cc623b6929e3a2dfaaf00fb8f5dca2257c87166d

elemoine:
This PR uses the scheduler as registry for parsers, but it doesn't use the scheduler for the parsing itself. That doesn't make too much sense to me.

> I quickly added my thoughts here zarov/itowns@cc623b6

That may be just me but I actually don't see the value of doing the parsing in the scheduler. What is the benefit compared to doing it in the provider? I understand that the core thing of the scheduler is the priority queue. Here for the parsing the priority queue is not used, so what's the point of doing the parsing within the scheduler?

zarov:
@elemoine the idea is to separate more the things we currently have in itowns: instead of adding everything in providers, we could have a simpler chain, managed here in the scheduler.
```txt
Provider -> (Fetcher ->) Parser -> Stylizer
```
I think having those blocks called somewhere will be a pain to maintain later, and moving it all outside of providers makes things easier. See the discussion in #182

elemoine:
> I think having those blocks called somewhere will be a pain to maintain later, and moving it all outside of providers makes things easier. See the discussion in #182

Thanks for pointing me to #182. I'll go ahead and read that before anything.

src/Core/Scheduler/Scheduler.js
mbredif: No strong opinion on this... `parser` is provided provided as a class  rather than an instance so the change is only cosmetic.

src/Core/Scheduler/Scheduler.js
mbredif: Ok, I will remove them

autra:
To add on what @zarov said: parser are also good candidates for workers. If we start using workers, we'll need to schedule call to them too. This is a step is this direction.

mbredif:
> What I meant by "parser needs more work" in the other PR: as we create a registry, we should use it in other providers instead of referencing parsers directly. This will allow to be able to use different parsers in a provider according to the needed output format. This work should be done in this PR to avoid an inconsistent state in the codebase (I don't expect it to be too complicated anyway). Thanks!

As I anticipated, that means it turns into a big PR, but you asked for it...
- PointcloudProvider and WFSProvider now use the parser registry
- the parsers I updated (Potree CIN/BIN, geojson, gpx, PNTS) have metadata beyond the parse function : strings for `format`, `mimetypes`, `extensions` (that provides an collection of registry keys as suggested by @zarov) but also a `fetchtype` to select the Fetcher method)

I will probably not have time in the near future to go much further on this PR, so do not ask me to cover all providers and parsers :). In the meantime, I created a branch for using the format registry on 3dtiles : [format_registry_3dtiles](https://github.com/itowns/itowns/compare/format_registry...format_registry_3dtiles). I think it works but I have not tested it much and changes are not trivial so it may deserve its own PR.

peppsac:
`RasterProvider.js` has a lot of parsing code: don't you think it would be a good idea to move all of it out of this provider?
(also RasterProvider implements a gpx parser instead of using `GpxParser`)

mbredif:
> so do not ask me to cover all providers and parsers :)

:)

Yes of course, but I sincerely think raster provider is obsolete as it contains too much hardcoded stuff and that the existing functionnality of rasterizing single files will be superseded by a properly configured FileProvider in the same way that the new functionnality of rasterizing WFS features will be offered by a properly configured WfsProvider . So that will be treated in another PR once this PR and #705 are merged. (this PR is already big enough), similar to the upgrade of the  [3dTilesProvider](https://github.com/itowns/itowns/compare/format_registry...format_registry_3dtiles).

src/Core/Scheduler/Scheduler.js
mbredif: done

src/Core/Scheduler/Scheduler.js
mbredif: done, parsers now have mimetypes, extensions and format strings that are all used as registration keys

mbredif (commented):
(comments of @zarov have been addressed)

peppsac:
Do we even want to introduce a `FileProvider`?

At the last codesprint we exchanged on iTowns' API and the need of moving away from the current layer/provider design and instead implement something similar to OpenLayers or vector-tiles = splitting sources and layers.

So IMHO we should work toward this goal instead of adding new features built on the current broken layer/provider API.

mbredif:
We definitely agree that legacy providers have to be removed/reworked/replaced and that the provider/layer API should be redesigned. 

To stay on topic here, I encourage you to use more relevant PR/issues :
- #695 (or a new one) to discuss the (admittedly needed) reworking of the provider/layer API. Be sure I will join this discussion.
- #705 : FileProvider discussion.

Putting that aside, do you have any thoughts on the scope of this PR, which is introducing a parser registry, normalizing the parser interface and using it in the providers of geometry layers ? (putting [aside](https://github.com/itowns/itowns/compare/format_registry...format_registry_3dtiles) 3dtiles to limit the size of the PR, and the TileProvider which is not reading from any source as far as I understand)

src/Core/Scheduler/Scheduler.js
```diff
@@ -256,6 +274,29 @@ Scheduler.prototype.getProtocolProvider = function getProtocolProvider(protocol)
...
+    if (typeof (parser.parse) !== 'function') {
+        throw new Error(`Can't add parser for ${parser.format}: missing a parse function.`);
+    }
+    var that = this;
+    function register(format) {
+        that.parsers[format] = that.parsers[format] || [];
+        that.parsers[format].push(parser);
+    }
+    register(parser.format);
+    parser.extensions.forEach(register);
```
zarov, line 290: I think it should be simple and only rely on `format`. We ditched `mimetypes` in #597, and I fail to see how we could benefit by having `extensions` and `mimetypes` on top of `format`.

src/Core/Scheduler/Scheduler.js
```diff
@@ -256,6 +274,29 @@ Scheduler.prototype.getProtocolProvider = function getProtocolProvider(protocol)
     return this.providers[protocol];
 };
 
+
+Scheduler.prototype.addFormatParser = function addFormatParser(parser) {
+    // eslint-disable-next-line no-console
```
zarov, line 279: To not forget to remove the log before merging, you should remove this comment imho ;)

zarov (changes requested):
@mbredif if you don't have time to work on this, I'll gladly take over, in another branch and PR (as I'm really interested to see more advance on this as I said). It needs more work here, in particularly in documentation.

src/Core/Scheduler/Scheduler.js
mbredif: I am not so sure either, but these 3 (extensions, mimetypes, format) seem complementary :
- extension may be the only thing you get when you are handed over a file (by url or drag n drop)
- mimetype is needed by wfs (in the query string)
- mimetype may be provided in the response header of fetches
- format is for me the name of the parser given by itowns, with more uniform naming conventions
- ...

What might still be missing is some kind of `accept` function that analyse the file content (header?) to determine whether it is an acceptable file to parse (eg reading the first 4 bytes of tiles in 3dTiles formats) .

mbredif:
The Parser normalization part of this PR is for me compulsory, but I must confess that I am now not entirely sold at a parser registry maintained by the scheduler. In fact I started on the premise that the protocol provider registry was a good thing, but I am not sure of it any more : if we get rid of JSON layers, we could ask the user code to create protocol and parser objects and hand them over in the layer options which would not require any protocol registry. I will open soon a new issue to discuss this.

For now, a middle ground could be to have each provider have its own registry as a simple array of default parsers in the preprocess function, but leave the opportunity to pass in the layer options a `parser` or an array of `parsers` that would override the default array of parsers defined by the provider. This way, we would still have coupling of a provider with its default parsers, but the goal is to make it **strictly** limited to the import of default parsers and their instantiation in the default array of parsers.

WDYT?

>  I'll gladly take over

Please go ahead :grinning: .

src/Core/Scheduler/Scheduler.js
zarov: > extension may be the only thing you get when you are handed over a file (by url or drag n drop)

Extension is not really reliable, and can be confusing: a `xml` file can be a KML, GPX or even something else.

> mimetype is needed by wfs (in the query string)

It isn't really a mimetype, see for example the (usecase in GeoServer)[http://docs.geoserver.org/stable/en/user/services/wfs/outputformats.html]. But I see the necessity here.

> mimetype may be provided in the response header of fetches

Yes, but then we should maybe rework on the whole format/mimetype thing.

> format is for me the name of the parser given by itowns, with more uniform naming conventions

Agreed, that's why I think it should only be registered with the `format` option.

zarov:
> we could ask the user code to create protocol and parser objects and hand them over in the layer options which would not require any protocol registry

It is an option indeed, and then we wouldn't have to maintain a list of relation between format/mimetype and parsers.

> a middle ground could be to have each provider have its own registry as a simple array of default parsers in the preprocess function

I feel that going to this is like the current `supportedFormats` behavior we have in providers: it would be an improvement, but a tiny one.

>  a parser registry maintained by the scheduler

It is to me the best solution, everything assembled in one place, the Scheduler, based on the configuration given by the user. We could drop the list of default providers and let the user decide which one to use with their layer. But I really don't think we will benefit by spliting everything in multiple class.

src/Core/Scheduler/Scheduler.js
peppsac: I'm with @zarov : I'd rather not have 3 different ways of declaring a format.

What we could have, though, is a format detector: if the format isn't explicitely specified, we can try to deduce from: url, filename, magic bytes, etc (something similar to what is done in [3dTileProvider](https://github.com/iTowns/itowns/blob/master/src/Provider/3dTilesProvider.js#L180) and [RasterProvider](https://github.com/iTowns/itowns/blob/master/src/Provider/RasterProvider.js#L79))

mbredif:
Let me rephrase myself to show that I am sure we are already on the same page :smiley: .
I do not care much about the registration key (which you propose to be only `format` instead of what this PR does with multiple registrations). In fact, I am now proposing that this registry may be a simple array of parsers, ordered by priority ? All parsers may then expose relevant metadata so that a generic format detector may be used (mimetype, possible extension, accept function that checks the file header for e.g. magic bytes ... ).

I agree that file extension only is not reliable (or even sometimes against the spec as for [3dtiles](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/TileFormats/PointCloud/README.md#file-extension)), but sometimes it works and some other times it may still allow to filter out easily some parsers.

> It isn't really a mimetype, see for example the (usecase in GeoServer)

In this case, maybe parsers with mentions in the wfs spec may have an optional 'wfsOutputFormat' string. (maybe with some defaults for others)

> I feel that going to this is like the current supportedFormats behavior we have in providers: it would be an improvement, but a tiny one. it would be an improvement, but a tiny one.

I agree, it is a small step to prepare for an upcoming more drastic decoupling that would require more discussion. but if there is already an agreement for a centralized parser registry, then go for it !

zarov:
#966 covers this, we can close this PR

PR #724, opened 2018-04-12
refactor(core): remove unused condition in Scheduler

zarov:
I gave a test in all examples we have, everything is working, and providers seem to be ok too, so we can drop this `|| Promise.resolve`. A provider with a `executeCommand` returning a Promise should be mandatory I think.

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -62,6 +62,14 @@ function _instanciateQueue() {
     };
 }
 
+/**
+ * The Scheduler is in charge of managing the [Providers]{@link Provider} that
+ * are used to gather resources needed to display the layers on a {@link View}.
+ * There is only one instance of a Scheduler per webview, and it is instanciated
+ * with the creation of the first view.
+ *
+ * @constructor
+ */
 function Scheduler() {
     // Constructor
     if (instanceScheduler !== null) {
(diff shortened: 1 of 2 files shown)


elemoine (approved):
It makes sense to me that provider.executeCommand should always return a promise. That should be part of the provider/scheduler contract.

zarov:
Yes, I'm adding more documentation on this then.

autra:
I kinda disagree here. We expect people to be able to write their own provider. That's why we did that: it allows people to return an undefined value if they don't have anything to give.

That's part of what a *framework* should do: be strict in what it returns, be open in what it receives (from user code). So I would prefer to keep this part, I find it more ergonomic. Just my 2 cents.

elemoine:
> I kinda disagree here. We expect people to be able to write their own provider. That's why we did that: it allows people to return an undefined value if they don't have anything to give.

Fair enough. So no strong opinion on my side.

zarov:
> I kinda disagree here. We expect people to be able to write their own provider. That's why we did that: it allows people to return an undefined value if they don't have anything to give.

Well if people can write their own provider, sure they can add `return Promise.resolve();` at the end of it.

>  be open in what it receives

Please read #714 again

elemoine:
> Well if people can write their own provider, sure they can add return Promise.resolve(); at the end of it.

Yep, that is why I have no strong opinion about this.

One minor thing maybe: to avoid creating objects we could allow provider.executeCommand to return undefined and do this in the scheduler:

```js
function _instantiateQueue() {
    var resolvedPromise = Promise.resolve();
    return {
        …,
        execute(cmd, provider, executingCounterUpToDate) {
            …
            var p = provider.executeCommand(cmd) || resolvedPromise;
            p.then(result => {
            });
            …
         },
    };
}
```

But that's pretty minor to me.

autra:
>     be open in what it receives
> 
> Please read #714 again

Come on, this is not the same matter, I'm sure you can admit that. I'm talking about parameters and return values here, whereas #714 speaks about API surface. The goal is to have a clean and minimal API, while easing implementation of individual extensions.

Btw, *technically* this breaks the current API (user can't return `undefined` or `null` any more), so we need a `BREAKING CHANGE` section in the commit to have it in the next changelog.

Anyway, do as you wish here, as @elemoine says, it's pretty minor.

zarov:
> Come on, this is not the same matter

In fact it is, it all converges to the fact of what a user of iTowns is/isn't capable of doing himself.

> The goal is to have a clean and minimal API

I'm curious on who decided that.

> Btw, technically this breaks the current API (user can't return undefined or null any more), so we need a BREAKING CHANGE section in the commit to have it in the next changelog.

Thanks, forgot to add it.

autra (approved):
LGTM, I'm fine either way, and this is simpler. Let's land this.

autra:
(but I might need to wait for the commit message rephrasing)

zarov:
Yes please wait, I'm adding at the same time documentation about this part.

zarov:
So this PR has quite changed as I've added more documentation to the Scheduler, especially about the Provider interface.

zarov:
Note: the description of `command` could be more complete as it also have parameters such as `earlyDropFunction` in it, but I didn't feel to add them as it is not relevant to the user imho.

src/Core/Scheduler/Scheduler.js
```diff
@@ -171,10 +245,27 @@ Scheduler.prototype.addProtocolProvider = function addProtocolProvider(protocol,
...
     return this.providers[protocol];
 };
 
+/**
+ * Get a copy of the full list of [providers]{@link Provider} added to the
+ * Scheduler.
+ *
+ * @return {Array<Provider>}
+ */
+Scheduler.prototype.getProviders = function getProviders() {
```
peppsac: This function is unused and I don't see how it could be useful. Maybe you could remove it completely?

peppsac (approved):
Thanks for the documentation effort!

This PR is a great improvment, but maybe you'll want to split your commit in half: a `refactor` commit and a `docs` one?

src/Core/Scheduler/Scheduler.js
zarov: Yes I agree !

zarov:
Done !

peppsac:
Cool thanks!

PR #725, opened 2018-04-12
fix(core): sphere argument needed for box3D.getBoundingSphere

mbredif:
## Description
THREE now outputs a [deprecation warning](https://github.com/mrdoob/three.js/blob/master/src/math/Box3.js#L524-L529) if `getBoundingSphere` is called with no sphere argument.

## Motivation and Context
 `getBoundingSphere` is now called with a preexisting sphere `target`.

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -29,7 +29,8 @@ function TileMesh(geometry, params) {
 
     this.obb = this.geometry.OBB.clone();
 
-    this.boundingSphere = this.OBB().box3D.getBoundingSphere();
+    this.boundingSphere = new THREE.Sphere();
+    this.OBB().box3D.getBoundingSphere(this.boundingSphere);
 
     this.material = new LayeredMaterial(params.materialOptions);
 


src/Core/TileMesh.js
```diff
@@ -29,7 +29,7 @@ function TileMesh(geometry, params) {
 
     this.obb = this.geometry.OBB.clone();
 
-    this.boundingSphere = this.OBB().box3D.getBoundingSphere();
+    this.OBB().box3D.getBoundingSphere(this.boundingSphere);
```
elemoine: The following should be used imo:

```js
this.boundingSphere = this.OBB().box3D.getBoundingSphere(new THREE.Sphere());
```

elemoine: I realize that ` this.OBB().box3D.getBoundingSphere(this.boundingSphere)` is not even correct (as `this.boundingSphere` is `undefined`). No?

autra: Yep. That's why travis is red currently.

gchoqueux: @mbredif you use THREE v91?

mbredif: > you use THREE v91?

Yes, by mistake :)

mbredif:
Updated.
Feel free to merge that between now and the upgrade to 0.91 :)

autra (approved):
The API already accepts a target in v0.89, so it definitely doesn't hurt to have this now :-) 

Thanks, LGTM!

autra:
Nitpicking the commit message though:
- it's a `chore` not a `fix` (we adapt to api change, we're not correcting a bug)
- let's start the message with a verb in the imperative form (it's a common good practice and we have adopted it in iTowns)

I propose something like:

```
chore: adapt getBoundingSphere call to three v0.91

This version of three adds a depreciation warning when using getBoundingSphere without argument
```

mbredif:
I took your proposition as is, just to be sure you wil not object :)

PR #726, opened 2018-04-13
Feat oriented picture

gliegard:
## Description

Example showing an oriented picture.

Also, a slider allows the user to translate the picture in depth, and to change transparency.

## Motivation and Context
The goal of this PR is to show how a user can simply set up an oriented image by defining the metadatas of the picture, the camera, the axises where rotations occurs (here is the aircraft convention).

oriented_picture


examples/oriented_images.js
```diff
@@ -0,0 +1,107 @@
...
+
+var promises = [];
+
+// `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+var viewerDiv = document.getElementById('viewerDiv');
+
+// Instanciate iTowns GlobeView*
+var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, {
+    renderer: renderer,
+    handleCollision: false,
```
peppsac: This is not needed, since it's only used when `noControls` isn't set.

examples/oriented_images.js
```diff
@@ -0,0 +1,107 @@
+/* global itowns, document, renderer, initPicture  */
+// # Simple Globe viewer
```
peppsac: Could you adapt the title of the example :) ?

examples/oriented_images.js
```diff
@@ -0,0 +1,107 @@
...
+
+// Define initial camera position
+var positionOnGlobe = {
```
peppsac: This isn't really used since you're going to override it anyway.
I'd suggest to either:
- change `GlobeView` API and move `coordCarto` to `options` and if unset: don't modify the camera's position.
- use 0, 0, 0 in your example and add a comment to explain that's useless

I greatly prefer option 1: it should be easy to implement (but should go in its own PR)

examples/oriented_images.js
```diff
@@ -0,0 +1,107 @@
...
+    longitude: 2.423814,
+    latitude: 48.844882,
+    altitude: 100 };
+
+var promises = [];
```
peppsac: I don't think you need to store the `promises`: we don't need to show the layers used in this example since it's not really the point of this example.

examples/oriented_images.js
```diff
@@ -0,0 +1,107 @@
...
+    var euler = new itowns.THREE.Euler(
+        itowns.THREE.Math.degToRad(panoramic.tilt),
+        itowns.THREE.Math.degToRad(panoramic.azimuth),
+        itowns.THREE.Math.degToRad(panoramic.roll),
+        'ZYX');
+
+    return new itowns.THREE.Matrix4().makeRotationFromEuler(euler);
+}
+
+function toCoordGeographic(panoramic) {
```
peppsac: This one line function is only used once: let's get rid of it and use its code directly at the calling site.

examples/oriented_images.html
```diff
@@ -0,0 +1,94 @@
+<html>
+    <head>
+        <title>Itowns - Globe + WFS</title>
```
peppsac: Wrong title :)

peppsac (changes requested):
Thanks for the PR @gliegard :)

I did a quick review, I'll followup when these easy-to-fix comments are fixed :)

peppsac:
Forgot one thing: you should move your example up in the *Examples* page so it's more visible :)

examples/oriented_images.js
gliegard: ok

examples/oriented_images.js
gliegard: yes

examples/oriented_images.js
gliegard: I understand, but I don't think it's a good idea to change GlobeView API.
Basically, I think it's a good idea to have a camera start position in the GlobeView, and it's usefull during dev process.
That being said, I've added a comment to explain that this variable is not very usefull and camera position will be overwritten to be placed in front of the picture later.

examples/oriented_images.html
gliegard: ok

examples/oriented_images.js
gliegard: ok

examples/js/OrientedImageHelper.js
```diff
@@ -0,0 +1,108 @@
...
+// eslint-disable-next-line no-unused-vars
+function initPicture(view, image, coord, up, lookAt, rotMatrix, distance, sizeX, sizeY, focale) {
+    var coordView;
+    var localSpace;
+    var orientedImage;
+    var quaternion;
+    var plane;
+
+    coordView = new itowns.Coordinates(view.referenceCrs, 0, 0, 0);
+    coord.as(view.referenceCrs, coordView);
```
peppsac: `coordView = coord.as(view.referenceCrs);` is clearer IMHO

examples/js/OrientedImageHelper.js
```diff
@@ -0,0 +1,108 @@
...
+    placeObjectFromCoordinate(localSpace, coordView);
+
+    // add second object : 'oriented image'
+    orientedImage = new itowns.THREE.Object3D();
+    // place the 'oriented image' in the 'local space'
+    localSpace.add(orientedImage);
+
+    // setup initial convention orientation.
+    orientedImage.up.copy(up);
+    orientedImage.lookAt(lookAt);
```
peppsac: Why are you using the camera properties (`lookAt` comes from `pictureInfos.camera.lookAt`) to orient the plane?

The plane and the camera should look to each other, not in the same direction (that might explain why you need to use `side: itowns.THREE.BackSide`...)

examples/js/OrientedImageHelper.js
```diff
@@ -0,0 +1,108 @@
...
+    object.lookAt(coord.geodesicNormal.clone().add(object.position));
+}
+
+function createTexturedPlane(textureUrl) {
+    var texture;
+    var geometry;
+    var material;
+
+    texture = new itowns.THREE.TextureLoader().load(textureUrl);
+    texture.flipY = false;
```
peppsac: Why do you need this?

examples/oriented_images.js
```diff
@@ -0,0 +1,100 @@
+/* global itowns, document, renderer, initPicture  */
+// # Oriented picture
+
+// Define initial camera position. Note that camera initial position will be overwritten by picture position later
```
peppsac: Then use some dummy (`0` ?) values instead and collapse on a single line.

examples/oriented_images.js
```diff
@@ -0,0 +1,100 @@
...
+    latitude: 45,
+    altitude: 100000000 };
+
+// `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+var viewerDiv = document.getElementById('viewerDiv');
+
+// Instanciate iTowns GlobeView*
+var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe, {
+    renderer: renderer,
+    sseSubdivisionThreshold: 6,
```
peppsac: Why is this needed?

examples/oriented_images.js
```diff
@@ -0,0 +1,100 @@
...
+        height: 4680.55588294683,
+        longitude: 6.83256920100156,
+        azimuth: 526.3900321920207,
+        roll: 2.1876227239518276,
+        tilt: -11.668910605126001,
+        image: './orientedImages/MontBlanc.jpg',
+    },
+    camera: {
+        size: [6490, 4408],
+        focale: 1879.564256099287 * 4.4,
```
peppsac: Where does the `4.4` value comes from?

examples/oriented_images.js
```diff
@@ -0,0 +1,100 @@
...
+        ppaX: 3245,
```
peppsac: Where does this come from as well? (I suppose that 6490 and 4408 is the image size in pixels?)

utils/debug/MaterialUV.js
```diff
@@ -0,0 +1,14 @@
+import * as THREE from 'three';
+import materialUV_VS from './shader/debugUV_VS.glsl';
+import materialUV_FS from './shader/debugUV_FS.glsl';
+
+const materialUV = new THREE.ShaderMaterial({
```
peppsac: I don't think we need to merge this in iTowns: this is some simple debug code with nothing itowns specific in it.

peppsac (changes requested):

utils/debug/shader/debugUV_FS.glsl
```diff
@@ -0,0 +1,8 @@
+#include <logdepthbuf_pars_fragment>
+
+varying vec2 vUv;
+
+void main()  {
+    #include <logdepthbuf_fragment>
+    gl_FragColor = vec4(vUv.x * vUv.y, vUv.x * vUv.y, vUv.x * vUv.y, 1.0);
```
mbredif: That outputs graylevel fragments. the following would be more colourful and informative, as it would depend on both u and v rather than on their product only :
`gl_FragColor = vec4(vUv, 0.5, 1.0);`

utils/debug/MaterialUV.js
mbredif: I think it is a good place in `debug` as it will be very useful as a debugging tool when we will implement projective texturing. The alternative of placing it in `examples/js` is fine with me, but that would require to put some webpack magic there.

utils/debug/MaterialUV.js
```diff
@@ -0,0 +1,14 @@
...
+    uniforms: {},
+    side: THREE.BackSide,
+    vertexShader: materialUV_VS,
+    fragmentShader: materialUV_FS,
+});
+
+export default {
+    materialUV,
+};
```
mbredif: Maybe you should add `materialUV` to `itowns/utils/debug/Main.js` and use it in your example, as for now it is disconnected.

examples/oriented_images.js
mbredif: not crucial but agreed : replacing the function call by a comment or choose a more specific variable name would better document what is going on.

Maybe you could just assign the result to `pictureInfos.camera.position`

examples/js/OrientedImageHelper.js
mbredif: I guess that it is to make it work unflipped ;). Photogrammetric conventions rely on unflipped images.

examples/oriented_images.js
mbredif: I propose to document this a little more and have ppa be an array
```
size: [6490, 4408], // size of the image in pixels
ppa: [3245, 2204], // coordinates of the principal point in pixels
focal: 1879.564256099287 * 4.4, // focal length in pixels
```

examples/js/OrientedImageHelper.js
```diff
@@ -0,0 +1,108 @@
...
+    // apply rotation
+    quaternion = new itowns.THREE.Quaternion().setFromRotationMatrix(rotMatrix);
+    orientedImage.quaternion.multiply(quaternion);
+    orientedImage.updateMatrixWorld();
+
+    // create a textured plane, representing the picture.
+    plane = createTexturedPlane(image);
+    orientedImage.add(plane);
+
+    transformTexturedPlane(plane, distance, sizeX, sizeY, focale, sizeX / 2, sizeY / 2);
```
mbredif: cheating with the ppa, aren't you ? :)

examples/js/OrientedImageHelper.js
```diff
@@ -0,0 +1,108 @@
+/* global itowns, document */
+function transformTexturedPlane(plane, distance, sizeX, sizeY, focale, ppaX, ppaY) {
+    // compute picture size
+    var Xreel = (sizeX * distance) / focale;
+    var Yreel = (sizeY * distance) / focale;
+
+    // compute ppa
+    var ppaXReel = ((ppaX - (sizeX / 2)) * distance) / focale;
+    var ppaYReel = (-(ppaY - (sizeY / 2)) * distance) / focale;
```
mbredif: this `-` sign here might be the reason why you need backface culling

examples/js/OrientedImageHelper.js
mbredif: agreed, specifying a target is when you optimize for memory allocation, which is not the case here as you initialize a dummy Coordinate positions just above.

examples/oriented_images.js
```diff
@@ -0,0 +1,100 @@
...
+    return new itowns.Coordinates('EPSG:4326', panoramic.longitude, panoramic.latitude, panoramic.height);
+}
+
+coord = toCoordGeographic(pictureInfos.panoramic);
+rotationMatrix = parseAircraftConventionOrientationToMatrix(pictureInfos.panoramic);
+
+// eslint-disable-next-line no-unused-vars
+plane = initPicture(globeView, pictureInfos.panoramic.image, coord, pictureInfos.camera.up,
+    pictureInfos.camera.lookAt, rotationMatrix, pictureInfos.distance,
+    pictureInfos.camera.size[0], pictureInfos.camera.size[1], pictureInfos.camera.focale);
```
mbredif: It is a matter of taste, but I would rather pass the `pictureInfos.camera` here

mbredif (approved):
That is a great demo, thanks @gliegard !
I have added some comments, but no blockers for me.

examples/js/OrientedImageHelper.js
gliegard: pictureInfos.camera.up and lookAt are used to define an axises where rotation will be applied.
Maybe I should change the name to 'pictureInfos.camera.orientationLookAt and orientationUp' ?

utils/debug/MaterialUV.js
gliegard: yep, it's done

examples/oriented_images.js
peppsac: But where does this magic `4.4` value comes from?
And how did you determine the position of the ppa? Is this always the center of the image?

utils/debug/MaterialUV.js
```diff
@@ -0,0 +1,13 @@
...
+import materialUV_FS from './shader/debugUV_FS.glsl';
+
+const materialUV = new THREE.ShaderMaterial({
+    uniforms: {},
+    vertexShader: materialUV_VS,
+    fragmentShader: materialUV_FS,
+});
+
+export default {
+    materialUV,
```
peppsac: The name should be uvMaterial (all materials in threejs are named as XXXXMaterial)

examples/oriented_images.html
```diff
@@ -0,0 +1,89 @@
...
+        <script src="../dist/debug.js"></script>
+        <script type="text/javascript">
+            var renderer;
+            var exports = {};
+        </script>
+        <script src="oriented_images.js"></script>
+        <script type="text/javascript">
+            /* global itowns, document, GuiTools, globeView, promises */
+            var menuGlobe = new GuiTools('menuDiv');
+            menuGlobe.view = globeView;
```
peppsac: This line is not needed: GuiTools stores view  in its constructor.

examples/oriented_images.html
```diff
@@ -0,0 +1,89 @@
...
+            // Listen for globe full initialisation event
```
peppsac: These 4 lines are not useful in this example.

examples/oriented_images.js
```diff
@@ -0,0 +1,117 @@
...
+var rotationMatrix;
+// eslint-disable-next-line no-unused-vars
+var plane;
+var camera;
+
+function addLayerCb(layer) {
+    return globeView.addLayer(layer);
+}
+
+// Define projection that we will use (taken from https://epsg.io/3946, Proj4js section)
```
peppsac: This is not useful in this example.

utils/debug/MaterialUV.js
peppsac: You're always returning the same instance. Maybe you should return `materialUV.clone()` instead

examples/oriented_images.html
```diff
@@ -0,0 +1,89 @@
...
+        <script src="js/OrientedImageHelper.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv">
+            <p id="info">
+                Picture taken by Walter Mittelholzer in 1919, metadata from 
+                <a href="https://smapshot.heig-vd.ch/map/?imageId=2873" target="_blank"> smapshot.heig-vd.ch </a>
+            </p>
+        </div>
+        <!--script type="text/javascript">
```
peppsac: Please remove dead code

autra:
Just passing by: we have a repo for our data: https://github.com/iTowns/itowns-sample-data. I think the jpg (alongside its metadata, so the `pictureInfos` object) should be committed there instead of here, if possible.

examples/oriented_images.html
```diff
@@ -0,0 +1,79 @@
...
+            <p id="info">
+                Picture taken by Walter Mittelholzer in 1919, metadata from 
+                <a href="https://smapshot.heig-vd.ch/map/?imageId=2873" target="_blank"> smapshot.heig-vd.ch </a>
+            </p>
+        </div>
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script type="text/javascript">
+            var renderer;
```
peppsac: This is not needed because you're not testing this example.
You can also remove the `oriented_image.js` file and move its content in this file.

peppsac: You should use the `css` introduced in PR #736

gliegard:
@peppsac I've updated the code

peppsac (changes requested):
You forgot to answer https://github.com/iTowns/itowns/pull/726#discussion_r182046599.

And you silently replaced one magic value by another.

(and don't forget to document the `size` parameter as well, because the image at http://www.itowns-project.org/itowns-sample-data/images/MontBlanc/MontBlanc.jpg size is 1024x1024).

mbredif:
I see you removed the handling of an off-center ppa. It is ok as it simplifies this PR. We will add it back later.

typo : focal length is spelled `focal` rather than `focale`

gliegard:
> You forgot to answer #726 (comment).

The 4.4 was a factor that we found manually because we don't have all metadatas of the camera.

> And you silently replaced one magic value by another.

Yes, it's the focal in pixel. I found it manually because I don't have all the metadatas of the camera.

> (and don't forget to document the size parameter as well, because the image at http://www.itowns-project.org/itowns-sample-data/images/MontBlanc/MontBlanc.jpg size is 1024x1024).

It's the size of the original image. I put a 1024x1024 image on the github to save space.

examples/oriented_images.html
```diff
@@ -0,0 +1,160 @@
...
+                    height: 4680.55588294683,
+                    longitude: 6.83256920100156,
+                    azimuth: 526.3900321920207,
+                    roll: 2.1876227239518276,
+                    tilt: -11.668910605126001,
+                    image: 'http://www.itowns-project.org/itowns-sample-data/images/MontBlanc/MontBlanc.jpg',
+                },
+                camera: {
+                    // size of the picture in pixel
+                    size: [6490, 4408],
```
peppsac, line 80: You should remove this field completely because:
1. it's bogus in your example because it doesn't match the image resolution
2. you can deduce it from the image, so there's no need to ask the user
3. it's not a `camera` parameter (it's an image parameter)

And you should keep the image ratio when reiszing the image (even if threejs complains about non-power-of-2 texture).

examples/oriented_images.html
```diff
@@ -0,0 +1,160 @@
...
+                    // focal length in pixel
+                    focal: 8270,
```
peppsac, line 82: Why do you request a focal in pixels? exif metadata usually contains the focal in `mm`. If you need pixels => perform the conversion in your code.

utils/debug/DebugUvMaterial.js
```diff
@@ -0,0 +1,13 @@
+import * as THREE from 'three';
+import debugUV_VS from './shader/debugUV_VS.glsl';
+import debugUV_FS from './shader/debugUV_FS.glsl';
+
+const debugUvMaterial = new THREE.ShaderMaterial({
```
peppsac: In fact I really don't think we need this: even three doesn't have this. So I fail to see why as a GIS framework we should add a debugging shader.

If this is useful to you: keep it in a branch / stash / patch, but not in itowns repository.

peppsac (changes requested):

utils/debug/DebugUvMaterial.js
gliegard: I think it can be usefull in the for dev that will deals with image/texture in the future, like projective texture. And I think it's not very usable for other dev if it's hidden in a branch...

I can add a checkbox in my example to show the debug shader, so it has more sense to keep it.
What do you think about this compromise ?

examples/oriented_images.html
mbredif: I am with @gliegard here : This `size` here is not the size of the image that is to be rendered but the size of the image that was used for the photogrammetric calibration. The beauty of projective texturing and uv mapping is that you do not care about pixel coordinates but about uv coordinates that are normalized to 0-1, so you can use any resized version of the same image with the same photogrametric calibration (even anisotropic ones that change the aspect ratio with different scaling in x and y). This was used in itowns-legacy where the same calibration was used for multiple resized versions of the same image.

peppsac: `size` is used in 2 place: 
- computing the `fov` but it should become useless as soon as the focal is back to `mm` (I'm not 100% sure on this)
- to determine the `PerspectiveCamera` ratio to use

If the 2 above points are correct the absolute value of `size` doesn't matter => we can drop the `size` attribute completely, because we can read them from the image directly (of course this needs that the image ratio is preserved instead of using a square image).

examples/oriented_images.html
mbredif: In photogrammetry (eg [orientation files in micmac](https://github.com/micmacIGN/Documentation/raw/master/DocMicMac.pdf)), you do not primarily care about the focal length in mm : you want to construct the projection matrix and the simplest route is to get everything expressed in pixels : image size, ppa, focal... So that is the other way around. If you have the focal length in mm, you should use your knowledge of the pixel size or sensor size to convert it into a focal length in pixels. Let us for now defer the estimation of the camera parameters from exif metadata to another PR or delegate this to some photogrammetric preprocessing.

utils/debug/DebugUvMaterial.js
mbredif: +1 for keeping the uvMaterial and the cameraHelper in this PR, but I think we need a gui option (like in a Debug or Advanced subfolder) in the example to activate/deactivate them (they are currently [commented out](https://github.com/iTowns/itowns/blob/14c18b31122c44e150e0a2581b5fb9f7208d1301/examples/oriented_images.html#L138-L142)).

examples/oriented_images.html
mbredif: Looking at the source of [THREE.PerspectiveCamera](https://github.com/mrdoob/three.js/blob/master/src/cameras/PerspectiveCamera.js), the focal in mm [assumes a 35mm sensor](https://github.com/mrdoob/three.js/blob/master/src/cameras/PerspectiveCamera.js#L28) or that the equivalent focal length is computed for such a sensor size. Thus that masks a hidden `filmGauge` parameter. When the PPA will be added back, it will certainly be expressed in pixels and we will then need the size for that as well.

I see your point of easing the usage by an end user that has just to look at the exif, but in any case, some photogrammetric preprocessing will be necessary to get a precise pose estimate (position + orientation). That processing may use the exif as initial guesses but is usually outputting intrinsic parameters in pixels (if not directly as a 3x4 projection matrix).


Why do you want to impose a constraint on the images not being resampled anisotropically ? If you want to make the size optional, then it could default to the actual image size (which would sadly fail silently if the image has been resampled).

I think the issue comes from using the THREE.PerspectiveCamera that is not well suited for photogrammetry but rather for easy configuration. Maybe another PR could use a custom camera that could be fed directly with the projection matrix rather than massaging it into a THREE.PerspectiveCamera.

examples/oriented_images.html
peppsac: You keep refering to photogrammetry but the image for this example isn't a coming from a photogrammetry dataset.
That's why it seems strange to me to rely on photogrammetric conventions.

utils/debug/DebugUvMaterial.js
peppsac: You only have to add one line in any shader to display the uv so I really don't see the added value of this `DebugUvMaterial.js` :-(

examples/oriented_images.js
peppsac: You forgot to reply to this question. Can you elaborate on why this is needed?

peppsac:
side note: http://jmfriedt.free.fr/dji/ has a nice set of pictures with all the required metadata... that could be an alternative example

mbredif:
@gliegard decided on this simple nice example of a single image of the Mont-Blanc mountain to take advantage of itowns's nice view of the Alps.

That may have misguided you, as our primary use case is indeed to support photogrammetric dataset such as the one you just mentioned (which has been produced by IGN's photogrammetric suite [Micmac](https://github.com/micmacIGN) by the way :grinning:, where orientation files give focal lengths in pixels) : image collections from handheld cameras, mobile mapping vehicles, drones, aerial imagery...

Whenever someone has another use case for image orientation (such as relying on exif metadata only), then that could be added in another PR, couldn't it ?

utils/debug/DebugUvMaterial.js
mbredif: No strong opinion, but it may be very handy to have this material already setup for itowns (logdepth support...). Do you feel it is outweighted by the maintenance burden of these few lines of code in debug ?

peppsac:
Thanks for the details, but I'm already aware of all this :)

My concern is simple: you're using photogrammetric conventions (because that's **your** use case) to display a photo that has nothing to do with photogrammetry. I simply find this strange and I think potential users of itowns might find this strange too.

But I won't argue anymore on this point because it can be improved in future PRs (as you wrote multiple times already :grinning: ).

A more important point though: I'd be *very* interested to know how much work it is to display this image (from the linked dataset, reduced): 
small

If it only requires to change the metadata => good. If it needs more than this => then this PR can probably be improved.

Regarding the debug thing: I don't want this merged, because it's trivial and unrelated to itowns.

gliegard:
@mbredif  Thank for your answers :)

@peppsac 
So I've removed the debug shader. If someone want to use the shader, he can find it here 55b097fd952759aded53a7fb232ed4ba348a7dd6 on a branch debug_uv_material.

mbredif:
@peppsac : Next step/PR for me is to make a MicMacOrientationParser that would automatize the reading of this xml file into a camera object.
In the mean time, you could manually lookup the focal in pixels in `CalibrationInternConique/F`, size in `CalibrationInternConique/SzIm`, the position in `Externe/Centre` and the rotation in `Externe/ParamRotation`.

gliegard:
> A more important point though: I'd be very interested to know how much work it is to display this image

I agree, the goal of this code is to be re-used with other datas. But it will be done in future PR.

peppsac (approved):

peppsac:
Cancelling my `Request Changes` (misclicked Approve but it doens't matter anyway, since @mbredif already approved) since my expectations for this demo are not aligned with the ones from @gliegard .

peppsac:
Note: [this question](https://github.com/iTowns/itowns/pull/726#discussion_r184605822) still hasn't got an answer (it looks like this line is there to hide a bug in itowns. If it's the case it would be appropriate to open an issue)

peppsac:
> Note: this question still hasn't got an answer (it looks like this line is there to hide a bug in itowns. If it's the case it would be appropriate to open an issue)

@gliegard: could you answer this question please? thanks!

gliegard:
@peppsac : I have changed the sseSubdivisionThreshold to 6 to have bigger tiles, so this example is faster. If I don't change it, it is very long to wait all the tiles to be loaded.

gliegard:
@peppsac: Also, changing the sseSubdivisionThreshold have also the side effect of hiding a bug in itowns, this kind of hole in the mountain.
bug

peppsac:
IMHO your 2 comments are bugs (itowns shouldn't be slow to load tiles when the camera is close to the ground).

Would you mind opening 2 issues so we don't forget about these?

Issue #727, opened 2018-04-14
3d Tiles example is broken


mbredif:
## Your Environment
Current master : commit c091b0d9532a97fc62c567121983ceee433b26f7

## Context
3d tiles example : http://www.itowns-project.org/itowns/examples/3dtiles.htm

## Steps to Reproduce (for bugs)
1.  visit http://www.itowns-project.org/itowns/examples/3dtiles.html
2. zoom in

## Expected Behavior
It works with no error logging

## Actual Behavior
The examples trigger many errors in the console.
- many Orthoimagery related errors are triggered
```
Access to Image at 'http://wxs.ign.fr/va5orxd0pgzvq3jxutqfuy0b/geoportail/wmts?LAYER=ORTHOIMAGERY.ORTHOPHOTOS&FORMAT=image/jpeg&SERVICE=WMTS&VERSION=1.0.0&REQUEST=GetTile&STYLE=normal&TILEMATRIXSET=PM&TILEMATRIX=13&TILEROW=3099&TILECOL=2373' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8080' is therefore not allowed access. The response had HTTP status code 404.
LayeredMaterialNodeProcessing.js?ad71:293 Imagery texture update error for TileMesh {uuid: "DDE63791-0C25-4505-BAC0-4B8DFC7C7DED", name: "", type: "Mesh", parent: TileMesh, children: Array(1), …} Event {isTrusted: true, type: "error", target: null, currentTarget: null, eventPhase: 0, …}
```
- a blocking error is triggered
```
MainLoop.js?a6e3:76 Uncaught TypeError: Cannot read property 'findCommonAncestor' of null
    at TileMesh.findCommonAncestor (TileMesh.js?57b9:262)
    at GeometryLayer.wgs84TileLayer.preUpdate (GlobeView.js?caf5:119)
    at MainLoop._update (MainLoop.js?a6e3:90)
    at MainLoop._step (MainLoop.js?a6e3:133)
    at eval (MainLoop.js?a6e3:52)
```

## Possible Cause/Fix/Solution
- Imagery related layers could be fixed by switching to a OpenSM layer as a quick fix to prevent the triggering of these errors (but the cause of these errors should be investigated in a second step)
`itowns.Fetcher.json('layers/JSONLayers/OPENSM.json').then(function (result) { return view.addLayer(result) });`
[...]

PR #728, opened 2018-04-15
Two fixes

peppsac:
2 one-liners :
* fix(core): correct typo in event's name
* fix(core): unassign layer of deleted objects
    
    As soon as an object is deleted is doesn't belong to any layer,
    so let's remove its layer attribute value.
    
    Fixes one part of #727 (Uncaught TypeError: Cannot read property
    'findCommonAncestor' of null).


zarov (approved):
Well spotted for the constant ! The other fix is okay too.

PR #730, opened 2018-04-16
fix(debug): cleanup pointcloud debug tools

peppsac:
This commit:
  - removes the custom color usage from debug tool to avoid messing with
custom colors set by the user
  - repairs the 'display children' / 'display parents' feature
  - cleans the debug UI by removing unused settings

### DIFF
--- a/utils/debug/PointCloudDebug.js
+++ b/utils/debug/PointCloudDebug.js
@@ -1,4 +1,3 @@
-/* global itowns */
 function isInHierarchy(elt, hierarchyNode) {
     if (elt.name.length > hierarchyNode.length) {
         return elt.name.startsWith(hierarchyNode);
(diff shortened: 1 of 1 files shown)


zarov:
Could you use the opportunity to reduce the `pointBudget` maximum value ? We'll never have 15M points in our usecases.

peppsac:
I've already modified it in the other pointcloud PR (see https://github.com/iTowns/itowns/pull/732/files#diff-4033c2db9cc0d018b29c8a27df1bf722L166)

elemoine:
For what it's worth, it looks good to me.

zarov (approved):
I'm not well aware on pointclouds, but this PR seems to do no harm from what I tested. LGTM

peppsac:
Thanks for the reviews :)

PR #732, opened 2018-04-16
Pointcloud improvement

peppsac:
2 commits improving pointcloud support for itowns.

The first one fixes the point budget handling for .bin format.

The second reworks SSE: it's similar to #674 but simpler and only for pointclouds.

Full commits message:
   - refactor(pointcloud): rework point budget implementation
    
    Since .bin and .cin have different repartitions of the points inside a node,
    they must use different algorithms.
    For .cin we want to draw a %-age of each node because points are evenly
    distributed.
    For .bin, we sort the nodes by 'importance', and stop drawing nodes when we
    have enoug points.
   - refactor(pointcloud): simplify SSE computation for pointcloud
    
    Project average point spacing on screen and compare to sse threshold.
    It's a simplification of the approach proposed in #674.
    
    This allows to drop a dependency (convexhull) and brings the SSE
    computation for pointcloud closer to what is done for other geometries.

Edit:
The SSE value computed here is the estimation of the maximum distance between 2 points in a given node.
PotreeConverter guarantees that the distance between each pair of points is >= `metadata.spacing / node.level` in all nodes. This is easy to visualize if you display successive children (using the 'sticky' debug tool):
distances

The SSE is the estimation of this distance projected on the screen.


peppsac:
Edited description + added image

src/Process/PointCloudProcessing.js
```diff
@@ -155,7 +78,7 @@ function markForDeletion(elt) {
 
     if (!elt.notVisibleSince) {
         elt.notVisibleSince = Date.now();
-        elt.shouldBeLoaded = -1;
+        elt.sse = -1;
```
mbredif: Maybe you could add a comment here stating what -1 means

src/Process/PointCloudProcessing.js
```diff
@@ -169,6 +92,11 @@ export default {
             return [];
         }
 
+        context.camera.preSSE =
```
mbredif: You could link here to an existing technical paper to explain the derivation

src/Process/PointCloudProcessing.js
```diff
@@ -212,78 +140,85 @@ export default {
...
                         }
+                        elt.obj.boxHelper.visible = true;
+                        elt.obj.boxHelper.material.color.r = 1 - elt.sse;
+                        elt.obj.boxHelper.material.color.g = elt.sse;
+                    }
+                }
+            } else if (!elt.promise) {
+                const distance = Math.max(0.001, bbox.distanceToPoint(context.camera.camera3D.position));
+                const priority = computeScreenSpaceError(context, layer, elt, distance) / distance;
+                elt.promise = context.scheduler.execute({
```
mbredif: Why do you divide by distance ?

src/Process/PointCloudProcessing.js
```diff
@@ -293,22 +228,52 @@ export default {
...
         }
 
         if (layer.displayedCount > layer.pointBudget) {
-            const reduction = layer.pointBudget / layer.displayedCount;
-            for (const pts of layer.group.children) {
-                if (pts.material.visible) {
-                    const count = Math.max(1.0, Math.floor(pts.geometry.drawRange.count * reduction));
-                    pts.geometry.setDrawRange(0, count);
+            // 2 different point count limit implementation, depending on the pointcloud source
+            if (layer.metadata.customBinFormat) {
```
mbredif: I would rather have the processing be agnostic of formats, nut depend rather on format properties or strategies. Could not you have reordered bin files as in cin files, or the other way around?

mbredif (commented):
Looks good and simplifies things, thanks !
I added a few comments to improve it even further :)
(I have not tested it though)

src/Process/PointCloudProcessing.js
mbredif: What about renaming customBinFormat to something like prefixSubsampling ?

src/Process/PointCloudProcessing.js
peppsac: To load closest node first.

src/Process/PointCloudProcessing.js
peppsac: I added a `layer.supportsProgressiveDisplay` property. (Note that the rest of the file relies on the potreeconverter format, there's more work to do to be able to use this processing code on any pointcloud source)

src/Process/PointCloudProcessing.js
peppsac: Done

src/Process/PointCloudProcessing.js
peppsac: Done

src/Process/PointCloudProcessing.js
mbredif: If I understand correctly, minimizing SSE is what you want eventually (as things settle), but you prefer to bias the ordering of the fetches towards closer nodes, even if those have a smaller SSE. Is that correct ? I think a comment should be placed here for future reference (eg, we might want to enable the customization of this strategy)

gchoqueux:
this commit : **refactor(pointcloud): rework point budget implementation**
Does it allow  `PointCloud` to be fully compliant with the data potree (from [potree converter](https://github.com/potree/PotreeConverter))? Especially in the organization of points in the data structure

src/Process/PointCloudProcessing.js
```diff
@@ -106,41 +52,18 @@ function initBoundingBox(elt, layer) {
...
 }
 
-function shouldDisplayNode(context, layer, elt) {
-    let shouldBeLoaded = 0;
-
-    if (layer.octreeDepthLimit >= 0 && layer.octreeDepthLimit < elt.name.length) {
-        return { shouldBeLoaded, surfaceOnScreen: 0 };
+function computeScreenSpaceError(context, layer, elt, distance) {
+    if (distance <= 0) {
+        return layer.sseThreshold;
```
mbredif: is it equivalent to returning Infinity ? (which would be parameterless and thus more explicit)

peppsac:
> this commit : refactor(pointcloud): rework point budget implementation
Does it allow PointCloud to be fully compliant with the data potree (from potree converter)? Especially in the organization of points in the data structure

Yes.

mbredif (commented):
LGTM

gchoqueux:
Ok thanks, it's good to be completely compliant with Potree data

src/Process/PointCloudProcessing.js
peppsac: Thanks for the review. I addressed both comments.

mbredif (approved):
LGTM

peppsac:
Thanks for the review @mbredif - and welcome to the @iTowns/reviewers team !

autra:
> Thanks for the review @mbredif - and welcome to the @iTowns/reviewers team !

actually he just has write access. The convention is : a review from someone from itowns-reviewers group. 

Nevermind, I'll make some comments on this if needed and we'll do a followup, ok for you?

peppsac:
Oops my bad. I'm reverting this PR then, until your approval.

peppsac:
PR reverted in #745 and reopened as #746.

PR #733, opened 2018-04-17
feat(controls): make event listeners opt-out for FirstPersonControls

peppsac:
Allows the user to control precisely when FirstPersonControls reacts to user input.
e.g: if a user might want to have multiple controls in parallel, and switch from one to another based on the context.

This commits allows to easily forward the events to only one of them, depending of the application state.

### DIFF
--- a/src/Renderer/ThreeExtended/FirstPersonControls.js
+++ b/src/Renderer/ThreeExtended/FirstPersonControls.js
@@ -5,44 +5,11 @@ import { MAIN_LOOP_EVENTS } from '../../Core/MainLoop';
 // but including these controls in itowns allows use to integrate them tightly with itowns.
 // Especially the existing controls are expecting a continuous update loop while we have a pausable one (so our controls use .notifyChange when needed)
 
-
-// Mouse movement handling
-function onDocumentMouseDown(event) {
-    event.preventDefault();
-    this._isMouseDown = true;
-
-    const coords = this.view.eventToViewCoords(event);
-    this._onMouseDownMouseX = coords.x;
-    this._onMouseDownMouseY = coords.y;
-
-    this._stateOnMouseDown = this._state.snapshot();
-}
-
 function limitRotation(camera3D, rot, verticalFOV) {
     // Limit vertical rotation (look up/down) to make sure the user cannot see
     // outside of the cone defined by verticalFOV
     const limit = THREE.Math.degToRad(verticalFOV - camera3D.fov) * 0.5;
     return THREE.Math.clamp(rot, -limit, limit);
-}
-
-function onPointerMove(event) {
-    if (this._isMouseDown === true) {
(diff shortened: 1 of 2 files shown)


mbredif:
Interesting use case :) Do you have a simple example to showcase that feature ?

> e.g: if a user might want to have multiple controls in parallel, and switch from one to another based on the context.

For that use case, don't you rather need an activation/deactivation method rather than a constructor option ?

peppsac:
> Do you have a simple example to showcase that feature ?

Here's an example : https://github.com/peppsac/itowns2/blob/measure_tool/examples/pointcloud.js#L15 that uses this feature.

mbredif:
Looking up upstream, `FirstPersonControls` has an `enabled` [flag](https://github.com/mrdoob/three.js/blob/master/examples/js/controls/FirstPersonControls.js#L14) that seems to do kind of what you want...

peppsac:
> Looking up upstream, FirstPersonControls has an enabled flag that seems to do kind of what you want...

Yes, and I implemented it like this initially.
But being able to feed event to the controls is more convenient and you can map several events to the `onMouseMove` functions.
I could add one more feature and replace the `onKeyDown|Up`  with something like `onMoveBegin|End` to allow easy customisation of input keys (eg: `controls.onMoveBegin('translateX')`).

peppsac:
>  I could add one more feature and replace the onKeyDown|Up with something like onMoveBegin|End to allow easy customisation of input keys (eg: controls.onMoveBegin('translateX')).

On a second though, I'd prefer to instead allow the user to specify her own mapping (and replace the default one defined in `MOVEMENTS`).

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -122,6 +42,7 @@ class FirstPersonControls extends THREE.EventDispatcher {
      * @param {number} options.verticalFOV - define the max visible vertical angle of the scene in degrees (default 180)
      * @param {number} options.panoramaRatio - alternative way to specify the max vertical angle when using a panorama.
      * You can specify the panorama width/height ratio and the verticalFOV will be computed automatically
+     * @param {boolean} options.disableEventListeners - if true, the controls will not self listen to mouse/key events
```
autra: maybe we can complete this documentation, at least listing `onMouseDown`, etc... 

It's obvious when you see the source code, but I expect that when one does set this option, they will still need these `onMouse` fns somewhere. Listing them here will save them a trip to the source code and make this option even more explanatory.

peppsac: Done. I also documented `update` and `reset` functions. And also added `FirstPersonControls.js` to the list of files that jsdoc needs to process :)

autra (approved):
Thanks!

PR #734, opened 2018-04-20
example: simplify orthographic controls

zarov:
Removed the kind of hacky controls of the orthographic example that I really didn't like and
replaced them by the PlanarControls provided by iTowns.

### DIFF
--- a/examples/orthographic.js
+++ b/examples/orthographic.js
@@ -1,4 +1,4 @@
-/* global itowns, renderer, setupLoadingScreen */
+/* global itowns, setupLoadingScreen */
 // # Orthographic viewer
 
 // Define geographic extent: CRS, min/max X, min/max Y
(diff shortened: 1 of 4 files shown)


examples/orthographic.js
```diff
@@ -10,34 +10,18 @@ var extent = new itowns.Extent(
...
-    view.camera.camera3D.top = cy + halfNewHeight;
-    view.camera.camera3D.bottom = cy - halfNewHeight;
+var controls = new itowns.PlanarControls(view, {
+    maxAltitude: 40000000,
+});
 
-    view.notifyChange(true);
-};
+// disable the rotation
+controls.handleRotation = function _() {};
```
peppsac: I'd prefer an option in `PlanarControls` constructor to disable rotation, instead of overriding the function.
(and currently, clicking the middle button will still rotate the camera)

peppsac: Also adding a zoom-speed option would be nice: the default zoom speed is slooooow :)

zarov: I was hesitant to change the core, but I'm ok with adding an option (damn middle button, didn't see that one). I'm tweaking a speed too !

zarov:
Updated

examples/orthographic.js
```diff
@@ -10,34 +10,20 @@ var extent = new itowns.Extent(
...
-    view.camera.camera3D.top = cy + halfNewHeight;
-    view.camera.camera3D.bottom = cy - halfNewHeight;
-
-    view.notifyChange(true);
-};
+var view = new itowns.PlanarView(viewerDiv, extent, { maxSubdivisionLevel: 10 });
+
+// eslint-disable-next-line no-new
+new itowns.PlanarControls(view, {
+    maxAltitude: 40000000,
```
peppsac: Would you mind adding the documentation for these parameters?

zarov: Do you mean here or in `PlanarControls` ?

peppsac: In PlanarControls' jsdoc

elemoine: Explaining the option values used here would make sense to me too. For example, while reading the example code I would like to understand why `enableRotation` is set to `false`.

zarov: Doing both !

elemoine: And on the `maxAltitude` value: why not making it bigger? As a user of the example I want to be able to zoom out, and the current `maxAltitude` value prevents it. I tried to change it to 16000000 but the map disappears when I zoom out.

elemoine:
I just added two comments, but overall I really like the idea of removing the custom controls that were used in this example.

zarov:
So I cleaned the documentation of `PlanarControls` and added an explanation of options. Note that only the construction and its options are generated in the final documentation, as I felt that a lambda user will have no interest in the others methods, but a more advanced user could still take a look at it by going into the code.

examples/orthographic.js
zarov: Concerning the `maxAltitude`, I had the same error as you, so I suggest to let it as it is.

peppsac: Probably another issue with `camera.far`...
#674  would help fixing this by providing a min/max distance for everything we need to display => we could dynamically adjust near/far

peppsac (approved):
LGTM, thanks

PR #738, opened 2018-04-24
chore: change the JSDOC configuration

zarov:
I'm getting tired of editing the `package.json` each time I want to add documentation, so I moved the list of files to jsdoc-config.json.

### DIFF
--- a/jsdoc-config.json
+++ b/jsdoc-config.json
@@ -6,5 +6,25 @@
                 "paths": ["static_jsdoc"]
             }
         }
+    },
+    "source": {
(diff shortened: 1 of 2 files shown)


peppsac (approved):
I'd prefer to be able to generate the jsdoc for all files :)

But I don't mind moving the jsdoc source file declaration to `jsdoc-config.json` so +1.

zarov:
I do too, but there is a lot of files to clean, and I prefer to do that for now ;) I'm rebasing and merging.

PR #740, opened 2018-04-25
test: use puppeteer to test all examples

peppsac:
This commit replaces the itowns-testing.js hack with puppeteer (https://github.com/GoogleChrome/puppeteer).

Advantages:
  - a real browser (Chrome) instead of node + mocks for missing features
  - WebGL capable
  - Can take screenshots and save them (not used on Travis yet)
  - Debuggable using Chrome Dev tools (Chrome instance used by puppeteer can be headless or not)

Each example now have at least one test to verify we can open it and that it
reaches the "mainloop has finished running" state.

I also implemented a couple tests for other examples (see 3dtiles and globe).

Adding test is easy, and we could add more to exercise more parts of itowns.

Last but not least: with this change the separation of examples in 2 files (html + js) is
not needed anymore.

Environment variables supported:
  - `USE_DEV_SERVER=1`: do not start a new itowns server (npm start). Useful if you already have a itowns server running
  - `SCREENSHOT_FOLDER=/tmp`: if specified, tests will save a screenshot in this folder (must exist)
 - `DEBUG=1`: launch Chrome in non-headless mode


gchoqueux:
@peppsac  The PR's subject  seems good. But for such a choice, it seems more appropriate to open an issue in advance. (to desist together) 
Thanks

zarov:
Good thing to see the merge of html + js files, we're going to clearer sample like threejs. To reduce this PR, this could be done in another one, or at least in another commit.

Otherwise I couldn't install puppeteer, with the error a lot of people seems to encounter [(`ERROR: Failed to download Chromium r549031! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD"`)](https://github.com/GoogleChrome/puppeteer/issues/2173). I know this might be because I am behing a proxy/firewall, but this is a no for me since I think users outside the iTowns team could be in the same situation, and I don't expect them to spend too much time configuring everything to make puppeteer work.

peppsac:
@gchoqueux :
> @peppsac The PR's subject seems good. But for such a choice, it seems more appropriate to open an issue in advance. (to desist together)

I prefer to first implement something to see if it's worth it. So I'll probably keep doing like this.
This doesn't mean that I force my choices on you: this still has to be approved before being merged.

@zarov :
> this is a no for me since I think users outside the iTowns team could be in the same situation

`puppeteer` is added as a dev dependency, so this issue will only affect iTowns developers, not regular itowns users.

gchoqueux:
As this PR proposes a technological choice, this choice can be revoked. To avoid wasting time in development, reviewing and testing (and not a story of forcing), it's better for the community to make this choice in advance. 
And don't confuse your preferences with those of the community

peppsac:
> As this PR proposes a technological choice, this choice can be revoked.

Yep. Feel free to expose your arguments against this change so we can start a discussion.

> To avoid wasting time in development, reviewing and testing (and not a story of forcing), it's better for the community to make this choice in advance

That's your call. 
But I think I can spend my time like I want :)

>  And don't confuse your preferences with those of the community

You're misunderstanding what I wrote.

I wrote "**I** prefer to write code and then ask a review". It's quite different from "everyone must write code before discussing".

gchoqueux:
For important changes, it is advisable to notify the community in advance. [see](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md#opening-a-pull-request)

It's better to validate in advance to avoid the time lost by the reviewers if finally the technological choice isn't validated

Could you add installation instructions with proxy (or link) for puppeteer

Could you split commit : clean examples and puppeteer

peppsac:
> For important changes, it is advisable to notify the community in advance. see

This document is a guide for new contributors.
Because closing a PR for a newcomer might be rude, this document suggests to open an issue.
But I'm not a newcomer regarding itowns, so I won't be offended if we decide that this PR isn't worth it.

> It's better to validate in advance to avoid the time lost by the reviewers if finally the technological choice isn't validated

Discussing in an issue or in a PR isn't "time lost".
And at least with a PR you can actually see the changes and test them, rather than discussing vague ideas that might not even work.

> Could you add installation instructions with proxy (or link) for puppeteer

I'll amend the README with the link @zarov posted above (https://github.com/GoogleChrome/puppeteer/issues/2173#issuecomment-372727628)

> Could you split commit : clean examples and puppeteer

I'll do this before merging.
Though the first one won't be "clean example" but "test: remove itown- testing" (because it's the removal of itowns-testing.js that allows to merge the .js and .html)

gchoqueux:
This document is a guide for all contributors

I'm talking about the loss of time in reviewing code, however there's no waste of time on the discussion to choice a technological (especially in a dedicated issue).

(https://github.com/GoogleChrome/puppeteer/issues/2173#issuecomment-372727628) it needs to be applied to every chromium update, it's not a viable solution.

The split is to simplify reviewing, so it's best to do it before the merge

peppsac:
>This document is a guide for all contributors
> I'm talking about the loss of time in reviewing code, however there's no waste of time on the discussion to choice a technological (especially in a dedicated issue).

@gchoqueux ok I'm done: I won't answer any of your comments not directly related to the code / design of this PR.

nosy-b:
@peppsac Thanks for the great work, puppeteer looks promising :smiley: 

- But yes you could have started an issue on that subject 
- The PR should show that it is _WIP_, or a _Suggestion for enhancement_
- Too bad that we need some special tweeking to make it work. -> Not yet ready for itowns

peppsac:
> But yes you could have started an issue on that subject

See my previous answers.

> The PR should show that it is WIP, 

This is not a WIP PR since it's fully working.

> or a Suggestion for enhancement

Done

> Too bad that we need some special tweeking to make it work. -> Not yet ready for itowns

So, between having no test at all for itowns and having to do a one time setup on your machine you choose the no-test option?

I pushed an alternative: I removed puppeteer from our dev-dependency and install it manually on travis.

peppsac:
@zarov : did you try with the documented environment variables https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#environment-variables ? 

Or another simpler way to fix it:
- install Chrome (recent enough, Chrome beta works fine: https://www.google.fr/chrome/browser/beta.html)
- install puppeteer without downloading Chrome: `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install puppeteer`
- then use the env var `CHROME` to tell itowns/mocha/puppeteer what Chrome app it should use:
`CHROME=/opt/google/chrome-beta/chrome npm run test-examples`

I don't think it's that much complicated, is it?

And if `puppeteer` is not acceptable, fine but what solution do you propose that you can commit to work on in a near future?

zarov:
Fair enough, it works following these instructions, I didn't try to fix the problem before. Those need to amended to the readme.

Love the screenshot feature ! However, it seems to overwrite a previous screenshot for an example that have multiple tests (see 3dtiles).
I can also see that some tests should be failing, as they are not correctly rendered: it is the case for `3dtiles` (white screenshot), `panorama` (black), `gpx` (no gpx trace visible). I also have a failing test for `positionGlobe`, but I think you are working on it ?

peppsac:
> Love the screenshot feature ! However, it seems to overwrite a previous screenshot for an example that have multiple tests (see 3dtiles).

Yes, I have a not-yet-pushed modification that allows to name screenshots to avoid this problem

> I can also see that some tests should be failing, as they are not correctly rendered: it is the case for 3dtiles (white screenshot), panorama (black),

Not sure whatr's wrong here. Can you run with `DEBUG=1` so you can see what's going on in the debugger?

`positionGlobe` test is expected to fail, until PR #748  is merged.

test/examples/boostrap.js
```diff
@@ -0,0 +1,179 @@
...
+const os = require('os');
+const path = require('path');
+
+// We could run 'npm start' to serve itowns for the tests,
+// but it's slow to start (so tests might fail on timeouts).
+// Since the 'test-examples' target depends on the 'run' target,
+// we instead run the simplest http server.
+// It needs to be forked, so it's written to a temp file before
+// running.
+const simplestHttpServer = `
```
zarov: I see no problem with moving this to its own file, like in `test/examples/server.js` maybe ?

peppsac: Yeah, I did it this way to only have test case in this folder. But this can be changed.

peppsac:
Rebased.

I also squashed the commits and move the 2 `View` fixes to PR #749.
And pushed a fix to make sure the loading screen is hidden when taking the screenshots.

Now waiting for comments on this PR (ping @iTowns/reviewers).

peppsac:
@gchoqueux:
> Could you split commit : clean examples and puppeteer

is done (thanks to #750)

src/Core/View.js
```diff
@@ -609,9 +609,10 @@ View.prototype.pickObjectsAt = function pickObjectsAt(mouseOrEvt, ...where) {
 
             // does this layer have a custom picking function?
             if (layer.pickObjectsAt) {
-                results.splice(
-                    results.length, 0,
-                    ...layer.pickObjectsAt(this, mouse));
+                const sp = layer.pickObjectsAt(this, mouse);
+                for (let i = 0; i < sp.length; i++) {
+                    results.push(sp[i]);
```
tbroyer: I suppose this rewrite was to avoid the spread operator. How about using this one-liner instead then?
```js
Array.prototype.push.apply(result, layer.pickObjectsAt(this, mouse));
```
(see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push#Merging_two_arrays)

test/examples/boostrap.js
```diff
@@ -0,0 +1,184 @@
+/* global process, Promise, before, after */
+// eslint-disable-next-line import/no-extraneous-dependencies
```
tbroyer: Should this be disabled for the whole `test/` folder?

test/examples/boostrap.js
```diff
@@ -0,0 +1,184 @@
...
+const net = require('net');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+// We could run 'npm start' to serve itowns for the tests,
+// but it's slow to start (so tests might fail on timeouts).
+// Since the 'test-examples' target depends on the 'run' target,
+// we instead run the simplest http server.
+// It needs to be forked, so it's written to a temp file before
```
tbroyer: Any reason you need to fork? Couldn't you run the server in the same process? (and btw use the `listening` event for `waitServerReady`)

You could then use:
* the `listening` event for `waitServerReady` (sure you'd have to rework the way `USE_DEV_SERVER` is handled, but that should be pretty easy as first glance)
* port 0 (i.e. let NodeJS find an available port, and retrieve it with `server.address().port`) instead of `findFreeTcpPort`
* https://www.npmjs.com/package/serve-static as the request listener

test/examples/boostrap.js
```diff
@@ -0,0 +1,184 @@
...
+    global.waitNextRender = page =>
+        page.evaluate(() => new Promise((resolve) => {
+            function getView() {
+                if (typeof (view) === 'object') {
+                    return Promise.resolve(view);
+                }
+                if (typeof (globeView) === 'object') {
+                    return Promise.resolve(globeView);
+                }
+                return undefined;
```
tbroyer: call `resolve(false)` and return `Promise.reject()` like below?

test/examples/boostrap.js
```diff
@@ -0,0 +1,184 @@
...
+                    resolve(true);
+                }
+                v.addEventListener(itowns.VIEW_EVENTS.LAYERS_INITIALIZED, resolveWhenReady);
+            });
+        }));
+
+        if (process.env.SCREENSHOT_FOLDER) {
+            await waitNextRender(page);
+
+            const sanitized = screenshotName.replace(/[^\w]/g, '_');
```
tbroyer: nit: authorize `_` in addition to `\w`?

test/examples/boostrap.js
```diff
@@ -0,0 +1,184 @@
...
+            console.log('Wrote ', file);
+        }
+
+        return result;
+    };
+    global.browser = await puppeteer.launch({
+        executablePath: process.env.CHROME,
+        headless: !process.env.DEBUG,
+        devtools: !!process.env.DEBUG,
+        args: ['--no-sandbox', '--disable-setuid-sandbox'] });
```
tbroyer: Could you document why those flags are needed?

test/examples/.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+        'eslint-config-airbnb-base/rules/strict',
+    ],
+    parserOptions: {
+        ecmaVersion: 2017,
+        sourceType: 'script',
+        ecmaFeatures: {
+            impliedStrict: true
+        }
+    },
+    env: {
```
tbroyer, line 13: Add Mocha? https://eslint.org/docs/user-guide/configuring#specifying-environments (IIUC, that'll save you from manually declaring its globals)

tbroyer (commented):
I wholeheartedly support this initiative. I'll let others eventually approve the PR though.

src/Core/View.js
peppsac: The problem was `javascript maximum call stack size exceeded` because the array might be very large.
Your suggestion unfortunately won't work either (`Do not use this method if the second array`) but I'll add a comment to explain why we need to do this.

test/examples/boostrap.js
peppsac: It's only used here, because `puppeteer` must be manually installed (the other tes files don't directly depend on `puppeteer` so they don't need the `eslint-disable` trick).
So I'd say we can keep it here, WDYT?

test/examples/boostrap.js
peppsac: Nope, the fork was the historical reasons, I removed it.

I've reworked this part a bit, but I'd rather not add another dependency to simply serve static files.

test/examples/boostrap.js
peppsac: Yep, done.

test/examples/.eslintrc.js
peppsac: Ah good idea, thanks

peppsac:
Thanks for the review @tbroyer, I pushed a commit addressing your comments.

test/examples/boostrap.js
tbroyer: Ah OK; yeah, keep it that way then.

test/examples/boostrap.js
```diff
@@ -59,38 +52,18 @@ function waitServerReady(port) {
...
-    });
-}
-
 before(async () => {
     let port = 8080;
 
     if (!process.env.USE_DEV_SERVER) {
-        port = await findFreeTcpPort(port);
+        global.itownsServer = startStaticFileServer();
+        port = global.itownsServer.address().port;
```
tbroyer: IIRC, documentation says you should wait for the `listening` event to fire.

How about some more refactoring?
* let `startStaticFileServer()` assign `global.itownsServer`
* also let it listen for the `listening` event and assign `global.itownsPort` there (and also do the `console.log`); and return a promise that resolves when the event is dispatched (meaning the server accepts requests, replacing the need for `_waitServerReady` in this case)
* have `waitServerReady()` only be called when `USE_DEV_SERVER` is used, `await`ing `startStaticFileServer` otherwise

```js
if (!process.env.USE_DEV_SERVER) {
  await startStaticFileServer();
} else {
  global.itownsPort = process.env.USE_DEV_SERVER;
  await waitServerReady(global.itownsPort);
}
```

Alternatively, have `createStaticFileServer`'s returned promise resolve to the port it's listening on, and `waitServerReady` resolve to the passed in port; so you can assign `global.itownsPort` at a single location:
```js
let server;
if (!process.env.USE_DEV_SERVER) {
  server = startStaticFileServer();
} else {
  server = waitServerReady(process.env.USE_DEV_SERVER);
}
// wait for the server to be ready
global.itownsPort = await server;

File: test/examples/boostrap.js [hunk 10]
```diff
@@ -59,38 +52,18 @@ function waitServerReady(port) {
...
-            resolve(port);
-        });
-    }
-    return new Promise((resolve) => {
-        _test(startPort, resolve);
-    });
-}
-
 before(async () => {
     let port = 8080;
```
tbroyer: Remove initialization here, value is always overwritten.

test/examples/boostrap.js
```diff
@@ -170,15 +144,23 @@ before(async () => {
...
 
 // close browser and reset global variables
 after(() => {
     global.browser.close();
-    if (global.itownsProcess) {
-        // stop itowns
-        process.kill(global.itownsProcess.pid);
+    if (global.itownsServer) {
+        // stop server
+        global.itownsServer.close();
```
tbroyer: Make the hook async?

```js
after(done => {
    …
    global.itownsServer.close(done);
```

test/examples/boostrap.js
```diff
@@ -158,7 +132,7 @@ before(async () => {
         if (process.env.SCREENSHOT_FOLDER) {
             await waitNextRender(page);
 
-            const sanitized = screenshotName.replace(/[^\w]/g, '_');
+            const sanitized = screenshotName.replace(/[^\w]|_/g, '_');
```
tbroyer: That will strip `_` from the input.

```js
replace(/[^\w_]/g, '_')
```

tbroyer (dismissed):

peppsac:
@tbroyer thanks for your feedback, should be all fixed now.

tbroyer:
Did I miss something or you forgot to assign `global.itownsServer`?

Otherwise SGTM

peppsac:
> Did I miss something or you forgot to assign global.itownsServer?

Fixed, thanks!

peppsac:
> As said above, I'll let someone else "formally" approve those changes. SGTM otherwise.

Thanks @tbroyer. Any reviewer willing to "formally" review this PR then?

README.md
```diff
@@ -63,6 +63,22 @@ See our [release page](https://github.com/iTowns/itowns/releases).
 
 See the [feature list wiki page](https://github.com/iTowns/itowns/wiki/Supported-Features) for a complet list of features and data formats.
 
+## Tests
+
+If you want to run tests you'll need to install [puppeteer](https://github.com/GoogleChrome/puppeteer).
+
+If puppeteer fails to download Chrome, you can try with the [documented environment variables](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#environment-variables).
+Or you can download it manually, and then:
+- install puppeteer without downloading Chrome: `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install puppeteer`
```
zarov, line 72: It can also be `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install` (without the specified package) to respect the current package-lock and all.

zarov (approved):
Good to me too, I had problems this morning with the third test of `3dtiles`, with timeout, but it seems okay now (test done in 13s).

package.json
```diff
@@ -36,8 +36,10 @@
     "babel-runtime": "^6.20.0",
     "earcut": "^2.1.1",
     "flatbush": "^1.3.0",
+    "http-server": "^0.11.1",
     "js-priority-queue": "^0.1.5",
     "jszip": "^3.1.3",
+    "puppeteer": "^1.3.0",
```
autra: these 2 dependencies should be dev dependencies.

autra (changes requested):

package.json
peppsac: `http-server` isn't needed so I'll remove it.

And `puppeteer` won't be installed by default (even in dev dependencies) to avoid download failures.

Fixing that.

autra (approved):

autra:
Thanks!

peppsac:
Rebased, squashed + added a test for the latest added example (from #726).

I'll merge this PR soon.

zarov:
A quick update: we have the final word on our previous problem with the proxy. It seems that in Ubuntu, when setting the `https_proxy` env variable, it automatically appends the `https` protocol to it, even if our proxy is not available in `https` and we want to access it through `http`. The solution is to export in `.profile`/`.bashrc`/`whatever` the correct variable: `export https_proxy="http://my.pro.xy"`.

So maybe now we could pass puppeteer as a dev dependency ?

PR #742, opened 2018-04-26
refactor: change the return value of Fetcher.texture

zarov:
To harmonize all `Fetcher` methods, I removed the object returned when calling `Fetcher.texture` to a simple promise. If it does impact someone outside of the iTowns core, I'll be happy to discuss it, as we are several to feel that the `Fetcher` module shouldn't be available in the public API.

BREAKING CHANGE: the signature of `Fetcher.texture` has changed, and only return a promise now, instead of an object containing the texture and the promise.

### DIFF
--- a/src/Provider/Fetcher.js
+++ b/src/Provider/Fetcher.js
@@ -58,19 +58,14 @@ export default {
     },
 
     /**
-     * @typedef {Object} TexturePromise
-     * @property {Promise} promise - a promise that resolves when the texture is loaded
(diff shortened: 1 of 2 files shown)


autra (approved):
We weren't using the texture object before the promise was resolved anyway, so for me this code is functionnally equivalent as the previous one, and is more clean, so LGTM :-)

zarov:
Thanks !

Issue #741, opened 2018-04-25
Caching in iTowns


zarov:
I wanted to talk more about caching and how we handle our resources in memory. I know I removed `CacheResource` in #687, but I'm having some regrets about it, I should have gone further than only replacing it by a `Map`.

Let's say I fetch a resource, and I want to parse it before using it: the perfect example is with `x-bil` textures for the elevation. It is logical to add it in the cache after the parsing. But if we are going to make the architecture more malleable, with providers/parsers/others separated, [we have a problem here](https://github.com/iTowns/itowns/blob/master/src/Provider/OGCWebServiceHelper.js#L64-L76). How can we move the parser outside of this part if we still need to do some cache action after the parsing, knowing the cache is specific to `OGCWebServiceHelper` here ?

What we need is a global cache, for all classes in iTowns to use (but not in the public API !), as @autra suggested:
> - Provide a cache expiration mechanism, with a configurable duration on a per layer basis? 0 could mean no cache and -1 could mean "cache infinitely".
> - Probably move this behaviour in Scheduler, or at least in a common part, to make every layer benefits of this
> - Maybe use localstorage or something similar to have part of the cache persists across session?

The last point could be taken into account when writing the first Cache, to ease the integration later also with other methods, like the `Cache` API.

It could currently be used in `OGCWebServiceHelper`, `TileProvider`, `WFSProvider` and `3dTilesProcessing` (removing the not so beautiful `Math.random` mechanism).

Anyone has a thought on this ?

mbredif:
Just a though, but a complementary strategy would be to limit the cache size by a size in MB rather than by a time expiration. I do not know how easy it is to implement though.

have you looked at https://www.npmjs.com/package/lru-cache ?

zarov:
I agree, but the size of a javascript object in MB is rather complicated afaik. And the linked package is only having a size in term of number of object in it. I'd rather not add another dependency, but it is subject to discussion what is the best solution !

mbredif:
about the size, the `max` cache size in `lru-cache` is in not necessarily in MB but in the units your `length` function provides to give an (possibly approximate) size of each object (and defaults to 1, to simply count objects).

That was just a proposition to be considered before we end up recoding it ;). I have no opinion on this maintainance / recoding tradeoff.

zarov:
Closing since we merged #743

PR #743, opened 2018-04-26
feat(core): add a Cache system with expiration policy

zarov:
As explained in #741, the CacheResource was removed in #687, but the cache was locked inside
module like OGCWebServiceHelper. In order to remove the cache outside of this module later, a new Cache has been added, introducing a flush system based on a cache expiration mechanism.

The cache expiration mechanism is simple: when setting something into the cache, a lifetime value is specified, which is then read when the `Cache.flush` method is being called. If the lifetime is too old, the entry is flushed from the cache.
We could also used something more complete like a LFRU ([see common cache replacement policies here](https://en.wikipedia.org/wiki/Cache_replacement_policies)), but I'm not sure how we can determine the size of the cache here.

Note: this is a proposition to #741

### DIFF
--- a/src/Core/Scheduler/Cache.js
+++ b/src/Core/Scheduler/Cache.js
@@ -1,0 +1,157 @@
+const data = new Map();
+const stats = new Map();
+
+/**
+ * This is a copy of the Map object, except that it also store a value for last
+ * time used. This value is used for cache expiration mechanism.
+ * <br><br>
+ * This module can be imported anywhere, its data will be shared, as it is a
+ * single instance.
+ *
+ * @module Cache
+ *
+ * @example
+ * import Cache from './Cache';
+ *
+ * Cache.set('foo', { bar: 1 }, Cache.POLICIES.TEXTURE);
+ * Cache.set('acme', { bar: 32 });
+ *
+ * Cache.get('foo');
+ *
+ * Cache.delete('foo');
+ *
+ * Cache.clear();
+ */
+const Cache = {
+    /**
+     * Cache policies for flushing. Those policies can be used when something is
+     * [set]{@link Cache.set} into the Cache, as the lifetime property.
+     *
+     * @name module:Cache
+     * @typedef {Object} module:Cache.POLICIES
+     *
+     * @property INFINITE - The entry is never flushed, except when the
+     * <code>all</code> flag is set to <code>true</code> when calling {@link
+     * Cache.flush}.
+     * @property TEXTURE - Shortcut for texture resources. Time is 15 minutes.
+     * @property ELEVATION - Shortcut for elevation resources. Time is 15
+     * minutes.
+     */
+    POLICIES: {
+        INFINITE: Infinity,
+        TEXTURE: 900000,
+        ELEVATION: 900000,
+    },
+
+    /**
+     * Returns the entry related to the specified key from the cache. The last
+     * time used property of the entry is updated to extend the longevity of the
+     * entry.
+     *
+     * @name module:Cache.get
+     * @function
+     *
+     * @param {string} key
+     *
+     * @return {Object}
+     */
+    get: (key) => {
+        const entry = data.get(key);
+        const stat = stats.get(key) || stats.set(key, { hit: 0, miss: 0 });
+
+        if (entry) {
+            stat.hit++;
+            entry.lastTimeUsed = Date.now();
+            return entry.value;
+        }
+
+        stat.miss++;
+    },
+
+    /**
+     * Adds or updates an entry with a specified key. A lifetime can be added,
+     * by specifying a numerical value or using the {@link Cache.POLICIES}
+     * values. By default an entry has an infinite lifetime.
+     *
+     * @name module:Cache.set
+     * @function
+     *
+     * @param {string} key
+     * @param {Object} value
+     * @param {number} [lifetime]
+     *
+     * @return {Object} the added value
+     */
+    set: (key, value, lifetime = Infinity) => {
+        const entry = {
+            value,
+            lastTimeUsed: Date.now(),
+            lifetime,
+        };
+        data.set(key, entry);
+
+        return value;
+    },
+
+    /**
+     * Deletes the specified entry from the cache.
+     *
+     * @name module:Cache.delete
+     * @function
+     *
+     * @param {string} key
+     *
+     * @return {boolean} - Confirmation that the entry has been deleted.
+     */
+    delete: key => data.delete(key),
+
+    /**
+     * Removes all entries of the cache.
+     *
+     * @name module:Cache.clear
+     * @function
+     */
+    clear: data.clear(),
+
+    /**
+     * Flush the cache: entries that have been present for too long since the
+     * last time they were used, are removed from the cache. By default, the
+     * time is the current time, but the interval can be reduced by doing
+     * something like <code>Cache.flush(Date.now() - reductionTime)</code>. If
+     * you want to clear the whole cache, use {@link Cache.clear} instead.
+     *
+     * @name module:Cache.flush
+     * @function
+     *
+     * @param {number} [time]
+     *
+     * @return {Object} Statistics about the flush: <code>before</code>
+     * gives the number of entries before flushing, <code>after</code> the
+     * number after flushing, <code>hit</code> the number of total successful
+     * hit on resources in the cache, and </code>miss</code> the number of
+     * failed hit. The hit and miss are based since the last flush, and are
+     * reset on every flush.
+     */
+    flush: (time = Date.now()) => {
+        const before = data.size;
+
+        data.forEach((entry, key) => {
+            if (entry.lifetime < time - entry.lastTimeUsed) {
+                data.delete(key);
+            }
+        });
+
+        let hit = 0;
+        let miss = 0;
+        stats.forEach((stat) => {
+            hit += stat.hit;
+            miss += stat.miss;
+        });
+        stats.clear();
+
+        return { before, after: data.size, hit, miss };
+    },
+};
+
+Object.freeze(Cache);
+export default Cache;
--- a/src/Provider/OGCWebServiceHelper.js
+++ b/src/Provider/OGCWebServiceHelper.js
@@ -1,17 +1,11 @@
 import * as THREE from 'three';
 import Fetcher from './Fetcher';
+import Cache from '../Core/Scheduler/Cache';
 import XbilParser from '../Parser/XbilParser';
 import Projection from '../Core/Geographic/Projection';
 import Extent from '../Core/Geographic/Extent';
 
-
 export const SIZE_TEXTURE_TILE = 256;
-
-// CacheRessource is necessary for neighboring PM textures
-// The PM textures overlap several tiles WGS84, it is to avoid net requests
-// Info : THREE.js have cache image https://github.com/mrdoob/three.js/blob/master/src/loaders/ImageLoader.js#L25
-const cache = new Map();
-const pending = new Map();
 
 const getTextureFloat = function getTextureFloat(buffer) {
     const texture = new THREE.DataTexture(buffer, SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE, THREE.AlphaFormat, THREE.FloatType);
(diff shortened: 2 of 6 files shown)


src/Provider/OGCWebServiceHelper.js
```diff
@@ -23,62 +17,47 @@ const tileCoord = new Extent('WMTS:WGS84G', 0, 0, 0);
...
-            if (!cache.has(url)) {
-                cache.set(url, texture);
-            }
-
-            pending.delete(url);
+            Cache.set(url, texture, Cache.POLICIES.TEXTURE);
             return texture;
         });
+
+        Cache.set(url, promise, Cache.POLICIES.PENDING);
```
peppsac: Can't we remove this `PENDING` thing and instead always store a `Promise` in the cache?
```js
if (Cache.has(url)) {
   return Cache.get(url);
}
const promise = Fetcher.texture(...).then((texture) => {
   ...
   return texture;
});
Cache.set(url, promise);
return promise;
```
?

zarov: Yes we could, it would be better indeed. But note that would impact `TileProvider` and make operation in it async.

peppsac: I don't think it's a problem since commands are already executed in an async context (see https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Scheduler.js#L154).

zarov: So, if I remove those policies, would that mean dropping the time-based lifetime in favor of something size-based ? If you feel that way too, it may be better to go in the suggestion of @mbredif in #741.

peppsac: I'm not sure how the removal of the `PENDING` policy is related to entirely dropping the time-based lifetime?

IMHO if this PR already unifies the caching mechanism + provides a way to remove outdated entries + simplifies the providers code it's already a big win.

Then we can keep working on this to improve it:
- add support for pre-loading/pre-caching (e.g: the application knows that in a near future it will display a currently hidden pointcloud and wants to start loading it now)
- allow itowns (eg LayeredMaterialNodeProcessing.js) to query the cache and adapt its behavior (no need to run 10s of async commands if it knows that everything needed is already available in the cache)

zarov: Oh I may have misunderstand the initial remark, sorry about that. It does not impact `TileProvider` at all, so I just updated for what you suggested.

src/Provider/OGCWebServiceHelper.js
```diff
@@ -23,64 +17,49 @@ const tileCoord = new Extent('WMTS:WGS84G', 0, 0, 0);
...
             texture.magFilter = THREE.LinearFilter;
             texture.minFilter = THREE.LinearFilter;
             texture.anisotropy = 16;
 
-            if (!cache.has(url)) {
-                cache.set(url, texture);
-            }
-
-            pending.delete(url);
+            Cache.set(url, texture, Cache.POLICIES.TEXTURE);
```
peppsac: I think it's not needed anymore because you already saved this request in the cache at line 34.

peppsac: Maybe you could adapt the Cache API to get `Cache.set` return the entry. Then you could use this pattern:
```js
getColorTextureByUrl(url, networkOptions) {
    return Cache.get(url) || Cache.set(url, Fetcher.texture(url, networkOptions).then((texture) => {
             texture.generateMipmaps = false;
             texture.magFilter = THREE.LinearFilter;
             texture.minFilter = THREE.LinearFilter;
             texture.anisotropy = 16;
    }));
}
```
WDYT?

zarov: Oh I didn't know I could do something like this. I'll try it !

zarov:
@peppsac updated !

src/Provider/WFSProvider.js
```diff
@@ -63,8 +62,8 @@ function getFeatures(crs, tile, layer) {
 
     const urld = URLBuilder.bbox(tile.extent.as(layer.crs), layer);
 
-    if (cache.has(urld)) {
-        return Promise.resolve({ feature: cache.get(urld) });
+    if (Cache.has(urld)) {
+        return Promise.resolve({ feature: Cache.get(urld) });
```
peppsac: You could use the same pattern as above.
```js
return Cache.get(urld) || Cache.set(urld, Fetcher.json(...).then(...)
```

peppsac (commented):
Looks better!

I left one minor style issue to fix and then we could merge this PR.

src/Provider/WFSProvider.js
zarov: Done !

gchoqueux: @peppsac Your last comment was taken into account, we can merge?

src/Provider/WFSProvider.js
```diff
@@ -89,7 +84,7 @@ function getFeatures(crs, tile, layer) {
                     throw err;
                 }
             })
-        .then(feature => assignLayer(layer.convert(feature), layer));
+        .then(feature => assignLayer(layer.convert(feature), layer)));
```
peppsac: This is a problem: you're putting into the global cache an entry that depends on a layer configuration.

You should either:
- not use the `url` as the key, but something that also depends on the layer
- or save 2 cache entries: `url -> json` (which can be shared) and then `key-that-depends-on-the-layer-id-and-the-json -> converted json`.

src/Core/Scheduler/Cache.js
```diff
@@ -0,0 +1,146 @@
...
+     * time is the current time, but the interval can be reduced by doing
+     * something like <code>Cache.flush(Date.now() - reductionTime)</code>. If
+     * you want to clear the whole cache, use {@link Cache.clear} instead.
+     *
+     * @name module:Cache.flush
+     * @function
+     *
+     * @param {number} [time]
+     */
+    flush: (time = Date.now()) => {
```
peppsac: Having a cache is good, but having statistics on how much hit/miss it generates is better.

Could you modify `flush` to return something easy to log/plot: `(hit, miss, entry_count)` for instance?
Or even better: add a new chart that draws these so we can see how useful this cache is...

src/Provider/WFSProvider.js
zarov: I don't get what is wrong here. The `url` seems the most appropriate candidate here to know which tile we are storing. And isn't using 2 cache entries an overkill ? If I understand that would mean having:
```js
return Cache.get(urldAndID) || (Cache.get(urld) || Cache.set(urld, Fetcher.json()))
    .then(
        json => GeoJsonParser.parse(),
        err => throw new Error()
    ).then(feature => Cache.set(urldAndID, assignLayer(layer.convert(feature), layer))));
```

I feel that this is going back to the first version of this PR (in the part of `OGCWebServiceHelper`). WDYT ?

src/Core/Scheduler/Cache.js
zarov: Added, but keep in mind that the current time values are not big enough to see rapid changes.

src/Provider/WFSProvider.js
peppsac: > I don't get what is wrong here.

Let's say that I declare 2 layers using the same wfs datasource, but layer1.type = 'color' and layer2.type='geometry'.
With the current implementation, here's what happen: layer1 will fetch the features using one url and store in the cache the result of layer.convert (which is a texture). Later layer2 will fetch the features using the same url, so it will be a cache hit. Except that layer2 expect geometries and that cache.get(url) will return a texture.

Is it clearer?

zarov: Oh yeah that makes sense, and the 2 cache entries even more. Trying it !

src/Core/Scheduler/Cache.js
```diff
@@ -132,13 +132,24 @@ const Cache = {
...
+     * cache.
      */
     flush: (time = Date.now()) => {
+        const entry_count = data.size;
+        let hit = 0;
+
         data.forEach((entry, key, map) => {
             if (entry.lifetime < time - entry.lastTimeUsed) {
                 map.delete(key);
+                hit++;
```
peppsac: You're using hit/miss in a rather unconventionnal way :)
Generally, in the context of a cache, `hit` is the number of times the requested value was in the cache. And `miss` is how many times the value wasn't. 

So `hit` and `miss` should be updated in `get` or `has`.

Actually: you could delete completely the `has` method since testing the result of `get` gives you the same result.
Then updating `miss` or `hit` would be solely in the `get` method.

And `flush` would reset them - by the way you could specify in the documentation that `flush` returns the number of hits/misses since last flush.

src/Provider/WFSProvider.js
autra: As a related note, we should be quite careful about storing promises in cache containing `.then` with side-effects. Here, for example, if I'm not mistaken, user can customize `layer.convert`. We cannot guarantee that the user-provided function will be side-effect free. Maybe its result is time-dependant?

Maybe it'd be better to stick to caching request results for now (the most costly thing in term of user perceived time)?

src/Core/Scheduler/Cache.js
zarov: Yeah sorry, updated !

src/Provider/WFSProvider.js
zarov: What do you mean by side-effects ? As long that it returns something compatible with the specified `update` function I see no problem, and the `update` function is specifiable by the user. I don't think we are taking a risk by storing promise here.

autra: We store promise with `.then` handlers, and these `then` callbacks will only be executed once with the cache mechanism, but in some cases they need to be executed each time.

Btw here, I just realized we already have a problem:

Let's say that I declare 2 layers using the same wfs datasource, but this time both are type 'geometry'.
With the current implementation, here's what happen: layer1 will fetch the features using one url and store in the cache the result of ` assignLayer(layer.convert(feature), layer))`, which is a Object3D with layer assigned to layer1. Later layer2 will fetch the features using the same url, so it will be a cache hit and layer2 will get a Object3D assigned to layer1.

It's because assignLayer has a side-effect. So imho: either we have no side-effect in the cache, or the key must depends on all the inputs, be explicit parameters or implicit context (meaning, all the side-effects of the attached `then` callbacks).

The latter is quite hard and fragile, because you need to adapt your key to potentially every changes of methods that are called in the `then` callback, supposing you got it right the first time (and the current thread shows it's not an easy task :-D), for each cases. That's why I think we should start by caching *only* the results of calls to `Fetcher` methods as a first step. So the pattern must be changed from:

```js
return Cache.get(...) || Cache.set(url, Fetcher.json()).then(...);
```
to
```js
return (Cache.get(...) || Cache.set(url, Fetcher.json(...))).then(...);
```

(notice than in the second case, `.then` operates on the result of `(Cache.get(...) || Cache.set(...))`)

Hope this is clearer :-)

autra: And getting the key right would be provider-dependant btw. I got actually beaten by that while trying this PR with the StaticProvider :-)

zarov: With your suggestion, we could directly move the cache into the `Fetcher`, but the only thing we would gain is the lack of a new http request. Only caching the received file is not useful, as the browser already does that. But caching the processed resource, that is one thing that we can benefit from.

I feel that the struggle here is caching in the `WFSProvider`. I suggest that I revert back to the current `master` version, only replacing the `cache` by the new `Cache`, and do the definition of the key later. The initial goal of the PR is only to replace the `Map` we have in several classes, with a goal in mind to move the cache at a higher level later (maybe in the `Scheduler` ?).

autra: > but the only thing we would gain is the lack of a new http request

which is one of the main cause of latency, it's not a little gain.

> Only caching the received file is not useful, as the browser already does that

*If* the server is correctly configured...

I just mean: it's not completely useless :-)

But yeah, you're right: it would actually be a pity if we stopped at the request level for this cache... But at the very least caching should be done per-layer for WFS, because the results depends too much on the layer config.

No strong opinion on your last suggestion. I'd say: do what feels more comfortable to you :-)

zarov: > If the server is correctly configured...

Yes indeed :sweat_smile: 

I'll let @peppsac or someone else give a last opinion on the subject, and I propose we do not take this PR too much further.

mbredif:
maybe a middleground would be to cache the parsed object. As parsers only offer a static parse function, a key that combines the url and some hash of the parse options should be ok.

mbredif:
> Let's say that I declare 2 layers using the same wfs datasource, but this time both are type 'geometry'. (@autra)

do you have an example of this ? do you mean a shared provider object or 2 layers that happen to configure 2 different providers with identical options ?

autra:
@mbredif None of what you said. Just 2 layers with the same protocol and url (but different options elsewhere). For instance, 2 wfs layers with 2 different colorization or extrusion logic.

zarov:
> maybe a middleground would be to cache the parsed object. As parsers only offer a static parse function, a key that combines the url and some hash of the parse options should be ok.

It still is a problem, as two layers with the same source could have different filter (done in the parser). I suggest to let the Cache only on the Fetcher for now.

zarov:
0ce3edb is my final proposal here, I think it solves all problems raised above, but it could be more optimal later.

src/Provider/WFSProvider.js
```diff
@@ -89,7 +84,7 @@ function getFeatures(crs, tile, layer) {
                     throw err;
                 }
             })
-        .then(feature => assignLayer(layer.convert(feature), layer));
+        .then(feature => Promise.resolve(assignLayer(layer.convert(feature), layer)));
```
peppsac: I don't think this change is necessary. Otherwise the rest looks good.

zarov: It is not indeed, I pushed the change !

peppsac (approved):
LGTM, thanks!

zarov:
Thank you for the review !

PR #744, opened 2018-04-26
feat(core): implement layer priorities

peppsac:
Before this commit, commands issued by layers were stored per-host.
These commands queues are sorted by commands priority.

The problem caused by this design is that commands priority values
are not normalized and if 2 different types of layer (eg: color and pointcloud)
shared the same host, their commands ended up in the same queue.
Their ordering would be determined by comparing "apples and oranges": color
layer priorities are not comparable to pointcloud layer priorities.

To address this problem, this commit introduces 1 queue per layer
within the same host and adds an optional layer priority.

When the scheduler has an empty slot for a given host, it will first
select the layer (based on their relative priorities) and then pick
its top most command.

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -18,18 +18,59 @@ import CancelledCommandException from './CancelledCommandException';
 
 var instanceScheduler = null;
 
+function queueOrdering(a, b) {
+    const cmp = b.priority - a.priority;
+    // Prioritize recent commands
+    if (cmp === 0) {
+        return b.timestamp - a.timestamp;
+    }
+    return cmp;
+}
+
+function drawNextLayer(storages) {
+    // Dithering algorithm to select the next layer
+    // see https://gamedev.stackexchange.com/a/95696 for more details
+    let sum = 0;
+    let selected;
+    let max;
+    for (const item of storages) {
+        const st = item[1];
+        if (st.q.length > 0) {
+            sum += st.priority;
+            st.accumulator += st.priority;
+            // Select the biggest accumulator
+            if (!selected || st.accumulator > max) {
+                selected = st;
+                max = st.accumulator;
+            }
+        }
+    }
+
+    if (selected) {
+        selected.accumulator -= sum;
+        return selected.q;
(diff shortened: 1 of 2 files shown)


peppsac:
This PR has one test failing, only when using node lts version.
@gchoqueux it's a test you added so let me know if you have any idea (especially since I don't understand where the expected values comes from: when running the `globe.html` example in a 800x600 window I don't get the same values at all).

peppsac:
> @gchoqueux it's a test you added so let me know if you have any idea (especially since I don't understand where the expected values comes from: when running the globe.html example in a 800x600 window I don't get the same values at all).

Ping @gchoqueux ? (I'm considering removing this test since it doesn't make sense)

gchoqueux:
test fixed in #742

test/scheduler_unit_test.js
```diff
@@ -31,9 +31,10 @@ function cmd(layerId = 'foo', prio = 0) {
         layer: {
             id: layerId,
             protocol: 'test',
+            priority: prio,
         },
+        priority: 1,
```
zarov: Unecessary now

src/Core/Scheduler/Scheduler.js
```diff
@@ -18,18 +18,58 @@ import CancelledCommandException from './CancelledCommandException';
...
-                return cmp;
-            },
-        }),
+        queue(command) {
+            const layer = command.layer;
+            let st = this.storages.get(layer.id);
+            if (!st) {
+                st = {
+                    q: new PriorityQueue({ comparator: queueOrdering }),
+                    priority: layer.priority || 1,
```
zarov: Duplicate from below

zarov (commented):
LGTM, I just don't understand the necessity to also add the dithering algorithm ? Won't the separation of each layer already doing the wanted job ?

peppsac:
> Won't the separation of each layer already doing the wanted job ?

The dithering is there to answer this question:
 
I have N layer (so N queues) that want to execute 1 command; which command do I execute next ?

zarov (approved):
Ok, LGTM then

PR #746, opened 2018-04-27
Pointcloud improvement

peppsac:
(reopening #732 )

2 commits improving pointcloud support for itowns.

The first one fixes the point budget handling for .bin format.

The second reworks SSE: it's similar to #674 but simpler and only for pointclouds.

Full commits message:
   - refactor(pointcloud): rework point budget implementation
    
    Since .bin and .cin have different repartitions of the points inside a node,
    they must use different algorithms.
    For .cin we want to draw a %-age of each node because points are evenly
    distributed.
    For .bin, we sort the nodes by 'importance', and stop drawing nodes when we
    have enoug points.
   - refactor(pointcloud): simplify SSE computation for pointcloud
    
    Project average point spacing on screen and compare to sse threshold.
    It's a simplification of the approach proposed in #674.
    
    This allows to drop a dependency (convexhull) and brings the SSE
    computation for pointcloud closer to what is done for other geometries.

Edit:
The SSE value computed here is the estimation of the maximum distance between 2 points in a given node.
PotreeConverter guarantees that the distance between each pair of points is >= `metadata.spacing / node.level` in all nodes. This is easy to visualize if you display successive children (using the 'sticky' debug tool):
distances

The SSE is the estimation of this distance projected on the screen.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8=",
       "dev": true,
       "requires": {
-        "mime-types": "2.1.18",
+        "mime-types": "~2.1.16",
         "negotiator": "0.6.1"
       }
     },
@@ -26,7 +26,7 @@
       "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
       "dev": true,
       "requires": {
-        "acorn": "4.0.13"
+        "acorn": "^4.0.3"
       },
       "dependencies": {
         "acorn": {
@@ -43,7 +43,7 @@
       "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
       "dev": true,
       "requires": {
-        "acorn": "3.3.0"
+        "acorn": "^3.0.4"
       },
       "dependencies": {
         "acorn": {
@@ -60,8 +60,8 @@
       "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
       "dev": true,
       "requires": {
-        "co": "4.6.0",
-        "json-stable-stringify": "1.0.1"
+        "co": "^4.6.0",
+        "json-stable-stringify": "^1.0.1"
       }
     },
     "ajv-keywords": {
@@ -76,9 +76,9 @@
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
       "requires": {
-        "kind-of": "3.2.2",
-        "longest": "1.0.1",
-        "repeat-string": "1.6.1"
+        "kind-of": "^3.0.2",
+        "longest": "^1.0.1",
+        "repeat-string": "^1.5.2"
       }
     },
     "ansi-escapes": {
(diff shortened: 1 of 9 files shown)


package-lock.json
```diff
@@ -5954,7 +5034,7 @@
     },
     "onetime": {
       "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
```
autra: huh?

src/Process/PointCloudProcessing.js
```diff
@@ -212,78 +143,86 @@ export default {
...
+                            initBoundingBox(elt, layer);
                         }
+                        elt.obj.boxHelper.visible = true;
+                        elt.obj.boxHelper.material.color.r = 1 - elt.sse;
+                        elt.obj.boxHelper.material.color.g = elt.sse;
+                    }
+                }
+            } else if (!elt.promise) {
+                const distance = Math.max(0.001, bbox.distanceToPoint(context.camera.camera3D.position));
+                // Increase priority of nearest node
```
autra, line 181: I don't understand the "nearest" here.

src/Process/PointCloudProcessing.js
```diff
@@ -212,78 +143,86 @@ export default {
...
+                const priority = computeScreenSpaceError(context, layer, elt, distance) / distance;
```
autra, line 182: Can't you set `elt.sse` here? You'd have it already on line 217 this way.

autra (changes requested):
I just want to understand the change `https` -> `http` in the `package-lock.json`. Seems very strange to me.

package-lock.json
peppsac: No idea where this change comes from.

src/Process/PointCloudProcessing.js
peppsac: `nearest` = closest to the camera.

src/Process/PointCloudProcessing.js
peppsac: I could but that would complexify a bit the code at line 127 (with a `if (elt.sse is already computed)`) for something that only happens once per node: when we decide to fetch them.

autra: Would it be guaranteed to be computed on line 127? (because elt.promise would be undefined the first time update is called on this node). But even if so, I agree this is a bit fragile. We can keep it like that.

src/Process/PointCloudProcessing.js
autra: Oh! right... I find the comment more confusing than the code alone :-)

package-lock.json
autra: I'm investigating

autra: Can you try to run again `npm i` in your branch to see if package-lock.json changes?

peppsac: nope, doesn't change

autra: I've pushed a commit with a package-lock.json generated with npm 6. Can you try if your `npm i` still works ok? if so, we can use this one I guess.

peppsac: It works, but `npm i` reverts all your changes to `package-lock.json`, eg:
```diff
-            "find-up": "^1.0.0",
-            "read-pkg": "^1.0.0"
+            "find-up": "1.1.2",
+            "read-pkg": "1.1.0"
```
I guess it's expected since I don't use npm 6...

autra (approved):

autra:
thanks!

PR #748, opened 2018-04-27
example: fix positionGlobe camera moves and altitude

peppsac:
Two fixes:
  - make sure we only render the globe (or plane) objects when querying the depthbuffer
  - update globecontrols before using getCameraTargetGeoPosition to avoid inconsistent
results.

### DIFF
--- a/examples/positionGlobe.js
+++ b/examples/positionGlobe.js
@@ -44,12 +44,15 @@ function addMeshToScene() {
     var geometry = new THREE.CylinderGeometry(0, 10, 60, 8);
     var material = new THREE.MeshBasicMaterial({ color: 0xff0000 });
     var mesh = new THREE.Mesh(geometry, material);
+    var cameraTargetPosition;
+    var meshCoord;
 
     // get the position on the globe, from the camera
-    var cameraTargetPosition = globeView.controls.getCameraTargetGeoPosition();
+    globeView.controls.updateCameraTransformation();
+    cameraTargetPosition = globeView.controls.getCameraTargetGeoPosition();
 
     // position of the mesh
-    var meshCoord = cameraTargetPosition;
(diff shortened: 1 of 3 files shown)


mbredif:
Thanks for the quick fix.
As I understand, that changes the semantics `view.getPickingPositionFromDepth` : now it only picks the base layer whereas it previously picked all objects, right ?

peppsac:
>  As I understand, that changes the semantics view.getPickingPositionFromDepth : now it only picks the base layer whereas it previously picked all objects, right ?

No it doesn't.
Because the `getPickingPositionFromDepth` only works with meshes that have a special material.

peppsac:
And https://github.com/iTowns/itowns/pull/740/commits/5071f16eada129a9564fd1efc58397f3da6cb776 is an example of a non-regression test that we could add.
This test reliably fails before the fix, and reliably works after the fix from this PR.

zarov:
Aside the goal of this PR, I'm quite undecided what to think of `renderViewToBuffer`. We call it everywhere with those fake `view` object, so imho we could change the method signature to `renderViewToBuffer(scene, camera, zone)` but then it would be necessary to change its name because the view part would make no sense. And then we could also change the `renderViewToRenderTarget` in the same spirit. But isn't it too much ?

peppsac:
@zarov yes, `renderToBuffer` and `renderToRenderTarget` + modified signatures would be better.

zarov:
Ok, if you can do it in this PR that would be great, so we don't forget it.

peppsac:
> Ok, if you can do it in this PR that would be great, so we don't forget it.

It's trivial to do but as it's a refactoring I'd prefer not include it in this `fix example` PR.

zarov (approved):
Ok, let's merge this then.

PR #749, opened 2018-05-03
Fix core view

peppsac:
This PR fixes 2 bugs in `View`, found thanks to #740.

@iTowns/reviewers r?

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -87,6 +87,17 @@ function View(crs, viewerDiv, options = {}) {
     }
 
     this._delayedFrameRequesterRemoval = [];
+
+    this._allLayersAreReadyCallback = () => {
+        // all layers must be ready
+        const allReady = this.getLayers().every(layer => layer.ready);
+        if (allReady &&
+                this.mainLoop.scheduler.commandsWaitingExecutionCount() == 0 &&
+                this.mainLoop.renderingState == RENDERING_PAUSED) {
+            this.dispatchEvent({ type: VIEW_EVENTS.LAYERS_INITIALIZED });
+            this.removeFrameRequester(MAIN_LOOP_EVENTS.UPDATE_END, this._allLayersAreReadyCallback);
+        }
+    };
 }
 
 View.prototype = Object.create(EventDispatcher.prototype);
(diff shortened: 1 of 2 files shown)


src/Core/View.js
```diff
@@ -495,6 +511,10 @@ View.prototype.execFrameRequesters = function execFrameRequesters(when, dt, upda
             frameRequester(dt, updateLoopRestarted, args);
         }
     }
+
+    if (this._delayedFrameRequesterRemoval.length > 0) {
```
autra: I'm pretty sure this should be at the beginning of `execFrameRequesters`, otherwise if you remove a frame requester outside of this `step` context, it will execute once more before being removed.

src/Core/View.js
```diff
@@ -87,6 +87,18 @@ function View(crs, viewerDiv, options = {}) {
     }
 
     this._delayedFrameRequesterRemoval = [];
+
+    this._allLayersAreReadyCallback = () => {
+        // all layers must be ready
+        const reducer = (allReady, layer) => allReady && layer.ready;
+        const allReady = this.getLayers().reduce(reducer, true);
```
autra: We have a shortcut for this: [Array.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every).

(note that we still don't have a `Array.remove(object)`, but okay, javascript ;-) )

autra (changes requested):

src/Core/View.js
peppsac: Done

src/Core/View.js
peppsac: Done

src/Core/View.js
```diff
@@ -85,6 +85,19 @@ function View(crs, viewerDiv, options = {}) {
     if (__DEBUG__) {
         this.isDebugMode = true;
     }
+
+    this._delayedFrameRequesterRemoval = [];
+
+    this._allLayersAreReadyCallback = () => {
```
autra, line 91: oh forgot this, do we need to attach it to `this`? If not, please consider extracting a simple module function.

autra: MMhh actually forget it, you need this inside. We could use `bind` but it's fine this way too.

autra (approved):

PR #750, opened 2018-05-04
refactor(examples): make most examples single file

peppsac:
Examples that doesn't use itowns-testing.js don't need to be splitted in 2
files (js and html).
This commit reunify them and remove the testing stuff from them as well (
renderer, exports, etc).


autra (approved):
Great, thanks!

PR #751, opened 2018-05-04
fix(chore): tag eslint config as root

peppsac:
From eslint doc: "By default, ESLint will look for configuration files in all
 parent folders up to the root directory." (see https://eslint.org/docs/user-guide/configuring)

So if one user cloned itowns in a folder with a .eslintrc.js, running npm run lint
in itowns folder would also use these rules.

This commit prevents this by adding a "root: true" attribute to our config.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+ root: true,
  'extends': [
    'eslint-config-airbnb-base',
    'eslint-config-airbnb-base/rules/strict',


autra (approved):
Thanks!

zarov:
Wait no, why would you add this ? If I have custom rules that I want to work, I can't make them work anymore ? This seems too personal to me to be merged.

autra:
@zarov you mean, if you have some rules in a `.eslintrc` your home folder?

zarov:
Yes exactly

autra:
Oh I see, I haven't thought about that.

> So if one user cloned itowns in a folder with a .eslintrc.js, running npm run lint in itowns folder would also use these rules.

is actually a feature.

I think the use case was to be able to embed itowns in other projects (with git submodules for instance). Is there even a way to support both of these use cases? If not, we have to choose (no opinion on my part). @peppsac WDYT?

peppsac:
This is a short reproducer of the bug I intended to fix. Let's say you have a project using Vue and itowns as a submodule.
In the parent directory you could have an `.eslintrc.js` file:
```js
module.exports = {
  plugins: [
    'vue'
  ]
}
```
Then if you go in itowns subfolder and run `npm run lint`-> it'll fail because of the `vue` plugin rule in the parent folder.
Is there any way to prevent this (= failure to lint itowns because of external rules) than adding `root: true` to our config?

zarov:
I understand the problem, it seems tricky to solve indeed. What about using `--no-eslintrc -c` for you parent directory ? Or use directly `--plugin` maybe ? (see options here https://eslint.org/docs/user-guide/command-line-interface#options)

autra:
@peppsac can you try to add a `plugins: []` in the `.eslintrc.js` of your itowns embedded in your vue project to see what happens?

peppsac:
> @peppsac can you try to add a plugins: [] in the .eslintrc.js of your itowns embedded in your vue project to see what happens?

Seems to work if I understood you correctly. But you can test it for yourself:
```bash
mkdir foo
cd foo
// put the .eslintrc.js from above in this folder
git clone https://github.com/iTowns/itowns.git
cd itowns
npm i
npm run lint
```

@zarov I don't think we can require users of itowns to change their configuration, in order to not break itowns' linter (or did I misunderstood your suggestion?)

zarov:
Seems fair to me.

autra:
> Seems to work if I understood you correctly. But you can test it for yourself:

Nice! So maybe being less implicit about our eslint conf would mitigate this problem enough to get rid of `root: true`? I don't know if it'll be enough though.

peppsac:
> Nice!

Ooops I misunderstood your comment. Now I really did the test you asked and it fails.

PR #752, opened 2018-05-04
test: fix wmts test

gchoqueux:
fix test for #744

### DIFF
--- a/src/Main.js
+++ b/src/Main.js
@@ -7,7 +7,7 @@ export { default as PlanarView, createPlanarLayer } from './Core/Prefab/PlanarVi
 export { default as PanoramaView, createPanoramaLayer } from './Core/Prefab/PanoramaView';
 export { default as Panorama } from './Core/Prefab/Panorama/Constants';
 export { default as Fetcher } from './Provider/Fetcher';
-export { MAIN_LOOP_EVENTS } from './Core/MainLoop';
+export { MAIN_LOOP_EVENTS, RENDERING_PAUSED } from './Core/MainLoop';
 export { default as View } from './Core/View';
 export { VIEW_EVENTS } from './Core/View';
 export { default as GpxParser } from './Parser/GpxParser';
--- a/test/wmts_test.js
+++ b/test/wmts_test.js
@@ -21,14 +21,12 @@ describe('wmts', function () {
                         example.initialPosition.latitude,
                         10000).as('EPSG:4978').xyz());
                 example.view.notifyChange(true);
-            } else if (initialState < 5) {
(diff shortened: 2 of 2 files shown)


peppsac:
Any details on how you get these numbers?

peppsac:
Just tested: it fixes #744 test problem.

gchoqueux:
you noticed that the tests were different?
the `wmts_test` counts the fetched textures (for `Ortho`) of the two states
while the `globe_test`counts the visible tiles by state

peppsac:
> you noticed that the tests were different?

Yes.
My question was simply: where does these numbers come from?

peppsac:
To elaborate:
- where did this `2` came from?
- how did you realized you should replace it with `20`?
- this doens't seem to count the `fetch` but only the number of tiles with a texture from layer 'Ortho' where texture.level == tile.level. If this is true you should test that all tiles at level 8 have a texture from 'Ortho' with level = 8, instead of testing against a hardcoded value (`= 20`)

Edit: actually if my last point is correct you can completely remove this test because this is the normal behavior of itowns (aka: stop updates as soon as texture.level = tile.level)

zarov:
Can we close this PR, as the `wmts_test` is no more ?

gchoqueux:
wmts_test was removed in PR #740

> this doens't seem to count the fetch but only the number of tiles with a texture from layer 'Ortho' where texture.level == tile.level.

This point is wrong, because the test collects all WMTS textures from the `Ortho` layer in visible tiles without testing `texture.level == tile.level`. This gives some indication of the count of loaded and assigned textures. 

This test validates [this optimization](https://github.com/iTowns/itowns/blob/dd4d7722e860bef4903aa9012322ab4c326bb13a/src/Process/LayeredMaterialNodeProcessing.js#L198). The Count of loaded and assigned textures was decreased.

There is no longer this test, it's a regression in the tests.

peppsac:
> This point is wrong, because the test collects all WMTS textures from the Ortho layer in visible tiles without testing texture.level == tile.level. This gives some indication of the count of loaded and assigned textures.

Your test did the following:
1. load the globe
2. then change the altitude of the camera to 10000
3. then wait for itowns to finish all its work (`example.view.mainLoop.renderingState == itowns.RENDERING_PAUSED`)
4. then counted verified that you had 20 tiles with texture level 8 (for layer `Ortho`)

In step 3 itowns is free to perform as much network fetches as it wants. And the optimization you were testing is supposed to limit the number of fetches...
So I stand by my point: this test has nothing to do with this optimization => not a regression.

PR #754, opened 2018-05-15
refactor: align Feature2Mesh and Feature2Texture

zarov:
## Description

The two modules are very similar: they take a feature or an array of feature as arguments, and return either a mesh or a texture. It makes sense to regroup them under the same scheme.

The documentation was completed in the same time.

No breaking change as the old function has been marked as deprecated.

## Question

I'm not quite happy with the name of the common method (`convert`), and I wish to find a better name. In the same vein, I would like to take the opportunity to move those two module in `src/Core/newFolder` to start the gathering of all "stylizing/symbolizing" modules (as in #695 maybe). Any suggestion ?

### DIFF
--- a/examples/globe_wfs_extruded.html
+++ b/examples/globe_wfs_extruded.html
@@ -94,9 +94,10 @@
                 name: 'lyon_tcl_bus',
                 type: 'geometry',
                 update: itowns.FeatureProcessing.update,
-                convert: itowns.Feature2Mesh.convert({
+                style: {
                     color: colorLine,
-                    altitude: altitudeLine }),
+                    altitude: altitudeLine,
+                },
                 linewidth: 5,
                 filter: acceptFeatureBus,
                 url: 'https://download.data.grandlyon.com/wfs/rdata?',
@@ -157,10 +158,11 @@
             globeView.addLayer({
                 type: 'geometry',
                 update: itowns.FeatureProcessing.update,
-                convert: itowns.Feature2Mesh.convert({
+                style: {
                     color: colorBuildings,
                     altitude: altitudeBuildings,
-                    extrude: extrudeBuildings }),
+                    extrude: extrudeBuildings
+                },
                 onMeshCreated: function scaleZ(mesh) {
                     mesh.scale.z = 0.01;
                     meshes.push(mesh);
@@ -212,9 +214,10 @@
             globeView.addLayer({
                 type: 'geometry',
                 update: itowns.FeatureProcessing.update,
-                convert: itowns.Feature2Mesh.convert({
+                style: {
                     altitude: altitudePoint,
-                    color: colorPoint }),
+                    color: colorPoint
+                },
                 size: 5,
                 onMeshCreated: configPointMaterial,
                 filter: selectRoad,
--- a/jsdoc-config.json
+++ b/jsdoc-config.json
@@ -23,10 +23,12 @@
             "src/Provider/URLBuilder.js",
 
             "src/Renderer/ColorLayersOrdering.js",
-            "src/Renderer/ThreeExtended/Feature2Mesh.js",
             "src/Renderer/ThreeExtended/GlobeControls.js",
             "src/Renderer/ThreeExtended/PlanarControls.js",
-            "src/Renderer/ThreeExtended/FirstPersonControls.js"
+            "src/Renderer/ThreeExtended/FirstPersonControls.js",
+
+            "src/Transform/feature2Mesh.js",
+            "src/Transform/feature2Texture.js"
         ]
     }
 }
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -8,6 +8,8 @@ import { GeometryLayer, Layer, defineLayerProperty } from './Layer/Layer';
 import Scheduler from './Scheduler/Scheduler';
 import Picking from './Picking';
 import { updateLayeredMaterialNodeImagery, updateLayeredMaterialNodeElevation } from '../Process/LayeredMaterialNodeProcessing';
+import feature2Mesh from '../Transform/feature2Mesh';
+import feature2Texture from '../Transform/feature2Texture';
 
 export const VIEW_EVENTS = {
     /**
@@ -181,6 +183,8 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
         defineLayerProperty(layer, 'visible', true);
         defineLayerProperty(layer, 'opacity', 1.0);
         defineLayerProperty(layer, 'sequence', 0);
+
+        layer.transform = feature2Texture;
     } else if (layer.type == 'elevation') {
         defineLayerProperty(layer, 'frozen', false);
     } else if (layer.type == 'geometry' || layer.type == 'debug') {
@@ -215,6 +219,8 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
                 });
             }
         });
+
+        layer.transform = feature2Mesh;
     }
     return layer;
 }
--- a/src/Provider/RasterProvider.js
+++ b/src/Provider/RasterProvider.js
@@ -7,7 +7,7 @@
 import * as THREE from 'three';
 import togeojson from 'togeojson';
 import Extent from '../Core/Geographic/Extent';
-import Feature2Texture from '../Renderer/ThreeExtended/Feature2Texture';
+import feature2Texture from '../Transform/feature2Texture';
 import GeoJsonParser from '../Parser/GeoJsonParser';
 import Fetcher from './Fetcher';
 
@@ -31,7 +31,11 @@ function createTextureFromVector(tile, layer) {
     if (layer.type == 'color') {
         const coords = tile.extent;
         const result = { pitch: new THREE.Vector4(0, 0, 1, 1) };
-        result.texture = Feature2Texture.createTextureFromFeature(layer.feature, tile.extent, 256, layer.style);
+        result.texture = feature2Texture(layer.feature, {
+            extent: tile.extent,
+            size: 256,
+            style: layer.style,
+        });
         result.texture.extent = tile.extent;
         result.texture.coords = coords;
         result.texture.coords.zoom = tile.level;
--- a/src/Transform/feature2Mesh.js
+++ b/src/Transform/feature2Mesh.js
@@ -258,16 +258,9 @@ function featureToExtrudedPolygon(feature, properties, options) {
     return new THREE.Mesh(geom);
 }
 
-/**
- * Convert a [Feature]{@link Feature#geometry}'s geometry to a Mesh
- *
- * @param {Object} feature - a Feature's geometry
- * @param {Object} options - options controlling the conversion
- * @param {number|function} options.altitude - define the base altitude of the mesh
- * @param {number|function} options.extrude - if defined, polygons will be extruded by the specified amount
- * @param {object|function} options.color - define per feature color
- * @return {THREE.Mesh} mesh
- */
+// Convert a Feature geometry to a three.js Mesh
+// Supported attributes (as constant or functions) in options are: altitude,
+// extrude and color. See the module exportation below for more details.
 function featureToMesh(feature, options) {
     if (!feature.vertices) {
         return;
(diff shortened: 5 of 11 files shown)


mbredif:
Thanks for this, that definitely goes in the right direction !
I agree '`convert`' might not be the best name ever. '`transform`' might be too generic as well, '`symbolize`' or '`stylize`' might be more explicit... If there is no agreement on a name, maybe we can just export the function rather than an object encapsulating a function.

#695 hints at a `src/Symbolizer` directory.

zarov:
> maybe we can just export the function rather than an object encapsulating a function.

As in `export default function`, and later call `Feature2Mesh/Texture()` ? Sounds good to me. I see no others methods that could be exported from this module, as it is quite long at the moment, so no problem to limit it to a function.

peppsac:
Suggestion: could you use this PR to add a test covering the features you're modifying? For instance add a test to `globe_vector.js` that would verify that picking the various rasterized elements (that use `Feature2Texture`) works as expected?

zarov:
@peppsac is d1805f3 adding a test as you expected ? I'll add more after.

zarov:
I pushed in the last commit the proposition of @mbredif, but the downside is we can't have a deprecation warning on `createTextureFromFeature`. WDYT ?

zarov:
@iTowns/reviewers anybody has an opinion here ?

jsdoc-config.json
```diff
@@ -23,6 +23,7 @@
 
             "src/Renderer/ColorLayersOrdering.js",
             "src/Renderer/ThreeExtended/Feature2Mesh.js",
+            "src/Renderer/ThreeExtended/Feature2Texture.js",
```
peppsac: :+1:  :)

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -331,6 +331,8 @@ function geometryToExtrudedPolygon(geometry, properties, options, multiGeomAttri
  * @param {number|function} options.extrude - if defined, polygons will be extruded by the specified amount
  * @param {object|function} options.color - define per feature color
  * @return {THREE.Mesh} mesh
+ *
+ * @ignore
```
peppsac: It's a bit weird to bother writing docs using jsdoc syntax and then add a `@ignore` tag.

IMHO, I'm fine with not publishing internal code documentation, but in this case I'd suggest to use a lighter (= easier to read by a human) doc format.
In this case it could be:
```js
// Convert a Feature geometry to a three.js mesh.
// Supported attributes (as constant or functions) in options are: altitude, extrude and color
```

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -415,28 +417,33 @@ function featureCollectionToThree(featureCollection, options) {
 }
 
 /**
- * @module Feature2Mesh
+ * Return a function that converts [Features]{@link module:GeoJsonParser} to
```
peppsac: Maybe we could ease the usage of this, and stop returning a function?

If a user needs to attach `options` we could instead support `Feature2Mesh.convert.bind(options)`?

peppsac (commented):
This PR is a good idea.

I left a couple suggestion regarding the API and the `@ignore` thing.

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: The goal was indeed to not publish this internal documentation, and not bloating the PR with a change here with `//`, but I'll add it if we're okay on it.

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: I'm all for it, I was afraid to break too much the current API.

src/Renderer/ThreeExtended/Feature2Mesh.js
autra: Why do you consider this as internal documentation? People can import every individual files in iTowns if they wish to. Imo this is completely public (and imo again, it's good that way).

zarov: This part is kind of a duplicate of the function documentation below, so I don't see the point of exposing it.

peppsac: > Why do you consider this as internal documentation

Becaue this function is not exported - so not accessible outside of Feature2Mesh.

autra: Oh I see, sorry I misunderstood, you're actually documenting a private function. Is the current theme exposing these comments? I intent to hide them in the future theme.

examples/globe_wfs_extruded.html
```diff
@@ -94,7 +94,7 @@
                 name: 'lyon_tcl_bus',
                 type: 'geometry',
                 update: itowns.FeatureProcessing.update,
-                convert: itowns.Feature2Mesh.convert({
+                convert: itowns.Feature2Mesh({
```
autra: Little detail, but can we then change the name of the export to `feature2Mesh` without capitalization?

autra:
> I would like to take the opportunity to move those two module in src/Core/newFolder to start the gathering of all "stylizing/symbolizing" modules (as in #695 maybe).

Good idea! I think you can drop the `Core` part, and put them directly into src/newFolder. We must find a balance between too many top level folders and too many folder level of course, but I think iTowns currently needs a flatter hierarchy (and I generally find them easier for newcomers, ex THREE.js has 15 folders under `src/`, and it' actually quite easy to find what you're looking for imho).

(a little detail, but I'd also like to drop folder capitalization `Core` -> `core` etc... I don't see the point of having them capitalized)

As for `newFolder`, I don't have any really great idea. Are these 2 files already stylizer according to you? Or will the styling process be after this transformation?

zarov:
> As for newFolder, I don't have any really great idea. Are these 2 files already stylizer according to you? Or will the styling process be after this transformation?

Currently I don't really see if the styling would be directly done inside those transformers, and I guess if we don't want to over bloat those files, it should be written outside. What about something like `src/transform` ?

examples/globe_wfs_extruded.html
zarov: Yep

zarov:
So in the last commit I did a few things:
- moved the `Feature2Mesh/Texture` files to `src/Transform/feature2Mesh/Texture`
- removed the option `convert` when adding a WFS layer (see the `wfs.html` for example), in favor of a `style` property of the layer, which is an object containing options for the styling of the layer (may be completed in the future when styling will be more advanced)
- the choice of using either the `feature2Mesh` or `feature2Texture` is done when adding the layer, depending on its type. The chosen method is set in `layer.transform`.
- `feature2Mesh/Texture` are now resolving directly a Promise, without getting the trouble to first return a function to execute
- `feature2Mesh` is not exposed in the public API anymore

As I feel there are a few breaking changes here, I would like at least two reviews before merging ! ping @iTowns/reviewers and @mbredif

autra:
> removed the option convert when adding a WFS layer (see the wfs.html for example), in favor of a style property of the layer, which is an object containing options for the styling of the layer (may be completed in the future when styling will be more advanced)

A couple of thoughts about this:

I still have some trouble drawing the lines between transformers and stylizers, I'm not even sure we should draw it.

Also, transformers could be chained. For instance, here we get geojson, we transform it once to get some features, then a second time to get a mesh, then a third time to extrude it, then a fourth time to set its altitude. Maybe this could be explicit?

But also we have colors and such, which looks like a stylizer, but the more I think about it, the more it looks like yet another transformation to me. 

So instead of "style", maybe in the future we could aim to have

```js
const transformation = chainTransformation(
        toFeature,
        toMesh,
        extrusion,
        altitude.bind(null, altitudeLine),
        color.bind(null, colorLine));
globeView.addLayer({
    name: 'lyon_tcl_bus',
    type: 'geometry',
    transform: transfomation,
    linewidth: 5,
    filter: acceptFeatureBus,
    url: 'https://download.data.grandlyon.com/wfs/rdata?',
    protocol: 'wfs',
    version: '2.0.0',
    id: 'WFS Bus lines',
    typeName: 'tcl_sytral.tcllignebus',
    level: 9,
    projection: 'EPSG:3946',
    extent: {
        west: 1822174.60,
        east: 1868247.07,
        south: 5138876.75,
        north: 5205890.19,
    },
    format: 'geojson',
});

```
Where each transformation as an output format and a list of supported input format (and you could have adapters in case one transformer returns an array from a single input element). WDYT about this kind of pipeline API? 

Imo the main advantages:
- it's more explicit (like a lego, whereas a configuration is always "interpreted" somehow)
- it's more flexible (a config only allows you to do what is supported)
- it's more extensible (you could write your own transformer)
- it's also coherent with the `filter` property, and could also be coherent with a futur `animate` property (that'll serve to animate a layer)?

If you like it, then maybe Feature2Mesh could be seen as our first transformer, before getting splitted in the future.

In short, I think I preferred the previous version (nearer to what I propose here, with some naming to change).

Some questions remains:
- is it feasible?
- is it worthy? :-)

src/Core/View.js
```diff
@@ -180,6 +182,8 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
         defineLayerProperty(layer, 'visible', true);
         defineLayerProperty(layer, 'opacity', 1.0);
         defineLayerProperty(layer, 'sequence', 0);
+
+        layer.transform = feature2Texture;
```
autra, line 187: layer.type color is also used for simple texture coming from WMTS, TMs and WMS, btw. It will still work with that, but it might be a bit confusing.

zarov: Maybe then we could change the type if we want to use `feature2Texture`, something like `raster` (and then drop the `RasterProvider`)

peppsac:
FWIW I'd prefer to stick with the style vocabulary, for one main reason: it's closer to existing APIs.

zarov:
> Also, transformers could be chained. For instance, here we get geojson, we transform it once to get some features, then a second time to get a mesh, then a third time to extrude it, then a fourth time to set its altitude.

The geojson to features is the goal of the GeojsonParser, and as it is/will be specified by the `format`, it is already kinda chained. For the following things, the type `geometry` tells that we want a mesh, and we apply some options to it (`style`). It seems already chained to me.

Flexibility can be seducing, but it can quickly became complicated to implement and follow it, but also not efficient in performance. We already aim for an enough divided cutting in the API imho.

autra:
> It seems already chained to me.

Yes it is chained of course, but users don't have direct control over that chain.

> Flexibility can be seducing, but it can quickly became complicated to implement and follow it

Good point.

To better explain my motivation, I'm a bit disturbed by this `style` object which:
- seems specific to wfs layers in its current form.
- is not extensible by external app devs.

That being said, I don't have other concrete non supported use case yet (is there anything else to do with buildings than extrusion, altitude and colors?), and also, I don't see a type of data other than vectors that would need styling. So maybe I'm worrying for nothing :-) 

Nothing preventing this PR to go on, anyway.

src/Core/View.js
autra: It's a bit confusing because I feel like we're mixing input and output type. Regular color layers are `color` all the way, whereas in our case, the input type is `vector`, but the output type is `color`... 

Maybe just have:
```js

  if (layer.type == 'color') {
    layer.transform = feature2Texture;
  } else if (layer.type == 'geometry') {
    layer.transform = feature2Mesh;
  }
```  
in WFSProvider.preprocess ? 

After all, the input format is determined - or at least constrained - by the protocol, and it is only currently supported for WFS layer. 

Or another idea (that'll take a bit more work maybe), make the `preprocessing` phase marks each layer with the output format of the protocol. So for layers with protocol set to 'raster', 'wfs' (and '3dtiles' in the future?), that would be `layer.protocolOutputFormat = 'feature'` (ok maybe a better name needs to be found), and for WMS that would be `'texture'` for instance. Then in View.js:178, you'd have:

```js
if (layer.protocolOutputFormat  == 'feature') {
   // logic to set the right layer.transform according to the type
}
```

Also the advantage is that it'll be easy to have this property set by other providers like `FileProvider`, which would be one good step (maybe the final one actually) towards the removal of RasterProvider (parsers would return Feature instead of Mesh directly). WDYT?

src/Transform/feature2Texture.js
```diff
@@ -95,37 +102,55 @@ function drawFeature(ctx, feature, origin, scale, extent, style = {}) {
...
+ * @function feature2Texture
+ *
+ * @param {module:GeoJsonParser~Feature} feature - a Feature or an array of
+ * Feature.
+ * @param {Object} [options] - Options controlling the conversion.
+ * @param {Extent} options.extent - The extent containing the feature.
+ * @param {number} options.size - The size of the texture, in pixels. The
+ * resulting texture will be a square, and it is highly recommended to use a
+ * power of 2 for GPU optimization purpose.
+ * @param {Object|function} options.style - The style to apply to the
```
autra, line 94: Oh I didn't realize functions were still supported. Which should actually gives all the possibilities I would ever want :-)

src/Provider/RasterProvider.js
```diff
@@ -31,7 +31,11 @@ function createTextureFromVector(tile, layer) {
     if (layer.type == 'color') {
         const coords = tile.extent;
         const result = { pitch: new THREE.Vector4(0, 0, 1, 1) };
-        result.texture = Feature2Texture.createTextureFromFeature(layer.feature, tile.extent, 256, layer.style);
+        result.texture = feature2Texture(layer.feature, {
```
autra, line 34: so following the idea above, RasterProvider would set `layer.outputFormat = 'feature'` and use layer.transform here. And now I'm wondering if `layer.transform` shouldn't be called in the Scheduler, on line 54:

```js
return provider.executeCommand(cmd).then((result) => {
    if (cmd.layer.transform) {
        cmd.layer.transform(result);
    }
    this.counters.executing--;
    cmd.resolve(result);
    // only count successul commands
    this.counters.executed++;
}, (err) => {
    this.counters.executing--;
    cmd.reject(err);
    this.counters.failed++;
    if (__DEBUG__ && this.counters.failed < 3) {
        console.error(err);
    }
});
```

src/Transform/feature2Mesh.js
```diff
@@ -415,28 +407,31 @@ function featureCollectionToThree(featureCollection, options) {
...
-
-            if (Array.isArray(feature)) {
-                return featureCollectionToThree(feature, options);
-            } else {
-                return featureToThree(feature, options);
-            }
-        };
-    },
-};
+export default function (feature, options = {}) {
```
autra: Let's rename `options` to `style` then.

autra: (and maybe also support functions, though maybe not in this PR)

autra (commented):
I should have started a review sooner :-)

overall LGTM, though I hope we'll find something for the `transform` mechanism.

Don't hesitate to tell if you find out my proposal implies the rewriting of half of itowns :-)

zarov:
> (is there anything else to do with buildings than extrusion, altitude and colors?)

Yes, texturing ! I hope to have a `texture` or even `material` as another option someday.

src/Provider/RasterProvider.js
zarov: > layer.transform shouldn't be called in the Scheduler

It should, but not in this PR. We can have the same principle as the current providers and the future parsers.

src/Transform/feature2Mesh.js
zarov: Yep, that would be better.

zarov: In fact I prefer to let the `options` naming. It makes sense to change it in `feature2Mesh`, but not in `feature2Texture`. I feel more that I should change the `style` options when adding a layer. Any suggestion on a name ?

autra: Yes indeed.

We might need something more generic: `paint` like mapboxgl? Or `draw`? `drawOptions`?

@peppsac, you said you wanted to stay with `style` though?

peppsac: I don't mind `options` here (but I prefer `style` when its declared within the layer)

zarov:
Closing this PR, as this is too old to still work on.

PR #755, opened 2018-05-16
feat(core): add a 'radius' parameter to picking methods

peppsac:
This radius parameter allows to be more tolerant towards user input.

Especially when working wiht point cloud, it's tedious to require the
user to exactly hover a point with the mouse.

BREAKING CHANGE: this modifies the signature of all picking functions.

### DIFF
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -81,7 +81,9 @@ function traversePickingCircle(radius, callback) {
                     continue;
                 }
 
-                callback(x, y);
+                if (callback(x, y) === false) {
+                    return;
+                }
             }
         }
         prevSq = sq;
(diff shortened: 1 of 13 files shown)


src/Core/Picking.js
```diff
@@ -60,12 +100,12 @@ function findLayerIdInParent(obj) {
 const raycaster = new THREE.Raycaster();
 
 export default {
-    pickTilesAt: (_view, mouse, layer) => {
+    pickTilesAt: (_view, mouse, radius, layer) => {
```
zarov: Could you add a quick documentation at least, so we aren't lost later ? (even if this isn't an exposed API and no documentation is generated from it)

test/examples/wfs.js
```diff
@@ -12,6 +12,10 @@ describe('wfs', () => {
         const result = await exampleCanRenderTest(page, this.test.fullTitle());
 
         assert.ok(result);
+
+        // test picking
+        const buildingId = await page.evaluate(() => picking({ x: 342, y: 243 }).properties.id);
```
zarov: This could be in another test, to be on the same page as other examples.

src/Core/Picking.js
```diff
@@ -30,22 +30,62 @@ function screenCoordsToNodeId(view, tileLayer, mouse) {
...
+    // iterate on radius so we get closer to the mouse
+    // results first.
+    // Result traversal order for radius=2
+    // --3--
+    // -323-
+    // 32123
+    // -323
+    // --3--
+    let prevSq;
+    for (let r = 0; r <= radius; r++) {
```
zarov, line 69: You can have instead:
```js
let prevSq = 0;
for (let r = 1; r <= radius; r++)
```

zarov (changes requested):
While this is really nice to have, it is really reducing the performance (at least in my case), even if it is to be expected, as more things are verified. See the two chrome profiles here: 
https://framadrop.org/r/V-k-XsRgOk#PMpo6NrV60XOooZxPx4F9Ia+mH7wbainhC5xgOs+8Lw=
https://framadrop.org/r/yRsZqCVNNr#c7v2a3DpLpAMA6QgT+sVaKHZUISrsevYzcQOB4SySfM=

src/Core/Picking.js
peppsac: That would miss the central point.

zarov: Yep my bad

peppsac:
@zarov can you retest now? Performance should be better. I also added a bit of documentation for the Picking module.

zarov (approved):
Way better, thanks ! This seems good to me.

zarov:
@peppsac please don't forget to squash when merging PR like this one.

peppsac:
@zarov arg, sorry !

PR #753, opened 2018-05-15
docs: clean the GeoJsonParser documentation

zarov:
Quick clean of the `GeoJsonParser` documentation - not sure if the `Feature` typedef could be declared in a higher scope ?

PR #756, opened 2018-05-16
refactor geojson parsing

zarov:
When parsing a multi polygon, line or point, the result was a modified `Array` with custom properties, not respecting the original object. It has been replaced by an `Object` containig the `Array` in question, under the name elements.

Also, when a multi polygon, line or point, is composed of only one element, we simplify the result by returning directly a polygon, line or point.

I would like to wait #753 and #757 to be merged before merging this one.

### DIFF
--- a/src/Parser/GeoJsonParser.js
+++ b/src/Parser/GeoJsonParser.js
@@ -1,6 +1,14 @@
 import Coordinates from '../Core/Geographic/Coordinates';
 import Extent from '../Core/Geographic/Extent';
 
+function applyOffset(indices, offset) {
+    for (const indice of indices) {
+        indice.offset += offset;
+    }
+
+    return indices;
+}
+
 function readCRS(json) {
     if (json.crs) {
         if (json.crs.type.toLowerCase() == 'epsg') {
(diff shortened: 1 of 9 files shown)


src/Parser/GeoJsonParser.js
```diff
@@ -145,21 +157,28 @@ const GeometryToCoordinates = {
         return lines;
     },
     multiPolygon(crsIn, crsOut, coordsIn, filteringExtent, options) {
-        const polygons = [];
-        polygons.type = 'multipolygon';
+        // if (coordsIn.length == 1) {
```
peppsac: Why did you comment this?

peppsac (commented):
Are you sure having the same function returning sometimes a geometry and sometimes an array of geometries a good idea?
IMHO as a user it's easier to use if you know that you're going to always receive the same type of object:
```js
result = {
   type: 'point',
   elements: [
      { vertices: [...], extent: ... }
   ]
};
```

src/Parser/GeoJsonParser.js
zarov: Error on my part, it wasn't supposed to be commented !

zarov:
I agree it can be confusing, but I really want to 1) drop the modified array and 2) return a simple object instead of a multi one. I'll update to follow the model with `elements` as you proposed for simple object too.

Anyway I'm moving the second commit to another PR to keep place for discussion on the first one.

zarov:
I added a proposition to what you suggested.

zarov:
I went further in the last commit by merging all `multi-` methods into one, but it is only a proposition.

zarov:
So I reworked (again...) all the previous things, and now GeoJsonParser can return two things, a `Feature` or a `FeatureCollection`, defined as follow:

```js
// FeatureGeometry (used after)
geometry = {
    vertices: [ Coordinates ],
    contour: [ number ],
    holes: [ [ number ] ],
    extent
}

// Feature
feature = {
    type: "polygon",
    geometry: [ FeatureGeometry ],
    properties: { ... },
    extent
}

// FeatureCollection
collection = {
    collection: [ Feature ],
    extent
}
```

What do you think ? I was also thinking that we should always return a `FeatureCollection`, but that is subject to some discussion.

src/Parser/GeoJsonParser.js
```diff
@@ -204,22 +178,24 @@ function readFeature(crsIn, crsOut, json, filteringExtent, options) {
 }
 
 function readFeatureCollection(crsIn, crsOut, json, filteringExtent, options) {
-    const collec = [];
+    const res = {
+        collection: [],
```
peppsac: I think I'd prefer `features` instead of `collection`

src/Renderer/ThreeExtended/Feature2Texture.js
```diff
@@ -111,9 +104,9 @@ export default {
         const scale = new THREE.Vector2(ctx.canvas.width / dimension.x, ctx.canvas.width / dimension.y);
 
         // Draw the canvas
-        if (Array.isArray(features)) {
-            for (let i = 0; i < features.length; i++) {
-                drawFeature(ctx, features[i], origin, scale, extent, style);
+        if (Array.isArray(features.collection)) {
```
peppsac: Surely this kind of code would be simpler to write if the parser would always return a `FeatureCollection`:
```js
        if (Array.isArray(features.collection)) {
            for (const feature of features.collection) {
                drawFeature(ctx, feature, origin, scale, extent, style);
             }
         } else {
             drawFeature(ctx, features, origin, scale, extent, style);
}
```
So +1 for returning a collection.

zarov:
@iTowns/reviewers any thoughts here ?

src/Parser/GeoJsonParser.js
zarov: Changed !

src/Renderer/ThreeExtended/Feature2Texture.js
zarov: Tried it in the last commit !

peppsac:
@zarov one way to simplify the output and the usage would be to split vertices and indices, as suggested in #763. This way you could store one big vertices array, next to the `type` (https://github.com/iTowns/itowns/pull/756/files#diff-eb8b327101f78c0753970cd17d1bdcb0R154).

WDYT?

zarov:
So merging `contour` and `holes` ? Apart that I don't understand what can be improved here (based on the fact that vertices are still set as `Coordinates`). 

Do you mean storing indices as `[offset, count, offset, count, offset, count...]` ? If so, we would ditch the `geometry` property of `Feature` and have: 
```js
// Feature
feature = {
    type: "polygon",
    vertices: Coordinates[],
    indices: [offset, count, offset, count, offset, count...]
    properties: { ... },
    extent
}

// FeatureCollection
collection = {
    collection: [ Feature ],
    extent
}
```
That would be a solution for "simple" point/line/polygon, but what about "multi" ones ? We wouldn't know where the next one start in indices. But I may have misunderstood what you suggested !

zarov:
Oh you meant keeping the `FeatureGeometry` but only with indexes !

peppsac:
> Oh you meant keeping the FeatureGeometry but only with indexes !

Yep :)

zarov:
So I pushed the modifications in the last commit, note that this makes #757 obsolete, as the fix has been done here in the same time. This commit also really simplify a lot `Feature2Mesh`.

zarov:
To answer a question asked on the other PR, yes the case of a complex multipolygon has been tested, you can use [this patch](https://gist.github.com/zarov/6d72481518e8df43a89c96af5d93178c) to view it.

peppsac:
> To answer a question asked on the other PR, yes the case of a complex multipolygon has been tested, you can use this patch to view it.

Do you know if this geojson is available under a license that would allow use to include it in our `globe_vector` example?

src/Parser/GeoJsonParser.js
```diff
@@ -244,37 +214,51 @@ export default {
...
-     * compose the contour (outer ring).
-     * @property {?Array} holes - If this geometry is a <code>polygon</code>,
-     * <code>holes</code> contains an array of indices representing holes in the
-     * polygon.
-     * @property {?Extent} extent - The 2D extent containing all the geometries.
-    */
+     * @property {Extent} extent - The 2D extent containing all the points
+     * composing the geometry.
+     * @property {?number[]} indices - If this geometry is a
+     * <code>polygon</code>, contains the indices that define the polygon. The
```
peppsac: I believe the indices are also defined for the linestrings (not just for polygons)

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -186,134 +182,79 @@ function geometryToLine(geometry, properties, options, multiGeomAttributes) {
...
 
     geom.setIndex(new THREE.BufferAttribute(new Uint16Array(indices), 1));
     return new THREE.Mesh(geom);
 }
 
 
-function geometryToExtrudedPolygon(geometry, properties, options, multiGeomAttributes) {
-    const vertices = multiGeomAttributes ? multiGeomAttributes.vertices : geometry.vertices;
-
+function geometryToExtrudedPolygon(feature, properties, options) {
```
peppsac: You could rename this method since it now takes a `feature` instead of a `geometry`

src/Parser/GeoJsonParser.js
```diff
@@ -59,108 +59,65 @@ function readCoordinates(crsIn, crsOut, coordinates, extent, target) {
...
         const coordinates = readCoordinates(crsIn, crsOut, coordsIn[0], extent);
         if (filteringExtent && !filteringExtent.isPointInside(coordinates[0])) {
-            return;
+            return [];
         }
-        const contour = {
-            offset: 0,
-            count: coordinates.length,
-        };
+        const indices = [globalOffset, coordinates.length];
```
peppsac: The `globalOffset` bit could be handled at the calling site instead of here. And would allow to remove the global var.

peppsac: Also I think having an array of objects `{ offset: , count: }` would be clearer and less error prone.

src/Parser/GeoJsonParser.js
```diff
@@ -59,108 +59,65 @@ function readCoordinates(crsIn, crsOut, coordinates, extent, target) {
...
+                return [];
             }
-            // only test the first line
             filteringExtent = undefined;
-            lines.push(l);
-            if (options.buildExtent) {
-                lines.extent = lines.extent || l.extent;
-                lines.extent.union(l.extent);
-            }
+            geometries.push(geometry[0]);
```
peppsac: Here is the place to manage the globalOffset thing.

src/Provider/RasterProvider.js
```diff
@@ -114,7 +114,7 @@ export default {
         }).then((feature) => {
             if (feature) {
                 layer.feature = feature;
-                layer.extent = layer.feature.extent || layer.feature.geometry.extent;
+                layer.extent = feature.extent || feature.geometry[0].extent;
```
peppsac: Do you think it's possible to have `feature.extent` that would be undefined but feature.geometry[0].extent that would not?

AFAICT both are computed if `options.buildExtent` is true so you could simplify this code.

zarov:
Unfortunately, afaik we can't use it officially here.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -324,67 +265,43 @@ function geometryToExtrudedPolygon(geometry, properties, options, multiGeomAttri
...
  * @return {THREE.Mesh} mesh
  */
-function geometryToMesh(geometry, properties, options) {
-    if (!geometry) {
+function geometryToMesh(feature, options) {
+    if (!feature.vertices) {
         return;
     }
 
-    // concat vertices of multigeometries in one big array
```
peppsac, line 340: :+1:

peppsac (commented):
This PR has been really improved with the latest changes.

I think this PR is going to be merged, so you can probably close #757.

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: Done

src/Parser/GeoJsonParser.js
zarov: I changed that, and got back to an object containing offset and count

src/Provider/RasterProvider.js
zarov: Yep, good catch

src/Parser/GeoJsonParser.js
zarov: Yes

zarov:
@iTowns/reviewers As we are starting to have a lot of PR, could this one be merged soon ?

peppsac (approved):
LGTM, thanks @zarov

zarov:
Thank you !

PR #757, opened 2018-05-17
fix(renderer): extruded polygon in Feature2Mesh

zarov:
Simple polygons in Feature2Mesh were not extruded properly. This has
been resolved, and in the same time the two very similar part of the
code related to this have been merged.

(moved out of #756)

### DIFF
--- a/src/Renderer/ThreeExtended/Feature2Mesh.js
+++ b/src/Renderer/ThreeExtended/Feature2Mesh.js
@@ -96,24 +96,27 @@ function addExtrudedPolygonSideFaces(indices, length, ring, isClockWise) {
     // add indices to make two triangle, that make the side face
     const mod = ring.offset + ring.count + length;
     for (let i = ring.offset; i < ring.offset + ring.count; ++i) {
+        const i1 = Math.max((i + 1) % mod, ring.offset);
+        const i2 = Math.max((i + length) % mod, ring.offset);
+        const i3 = Math.max((i + length + 1) % mod, ring.offset);
         if (isClockWise) {
             // first triangle indices
             indices.push(i);
-            indices.push((i + length) % mod);
-            indices.push((i + 1) % mod);
+            indices.push(i2);
+            indices.push(i1);
             // second triangle indices
-            indices.push((i + 1) % mod);
-            indices.push((i + length) % mod);
-            indices.push((i + length + 1) % mod);
+            indices.push(i1);
+            indices.push(i2);
+            indices.push(i3);
         } else {
             // first triangle indices
-            indices.push((i + length) % mod);
+            indices.push(i2);
             indices.push(i);
-            indices.push((i + length + 1) % mod);
+            indices.push(i3);
             // second triangle indices
-            indices.push((i + length + 1) % mod);
+            indices.push(i3);
             indices.push(i);
-            indices.push((i + 1) % mod);
+            indices.push(i1);
         }
     }
 }
--- a/src/Renderer/ThreeExtended/Feature2Mesh.js
+++ b/src/Renderer/ThreeExtended/Feature2Mesh.js
@@ -94,25 +94,26 @@ function coordinatesToVertices(contour, altitude, target, offset = 0, extrude = 
 function addExtrudedPolygonSideFaces(indices, length, ring, isClockWise) {
     // loop over contour length, and for each point of the contour,
     // add indices to make two triangle, that make the side face
-    for (let i = ring.offset; i < ring.offset + ring.count - 1; ++i) {
+    const mod = ring.offset + ring.count + length;
+    for (let i = ring.offset; i < ring.offset + ring.count; ++i) {
         if (isClockWise) {
             // first triangle indices
             indices.push(i);
-            indices.push(i + length);
-            indices.push(i + 1);
+            indices.push((i + length) % mod);
+            indices.push((i + 1) % mod);
             // second triangle indices
-            indices.push(i + 1);
-            indices.push(i + length);
-            indices.push(i + length + 1);
+            indices.push((i + 1) % mod);
+            indices.push((i + length) % mod);
+            indices.push((i + length + 1) % mod);
         } else {
             // first triangle indices
-            indices.push(i + length);
+            indices.push((i + length) % mod);
             indices.push(i);
-            indices.push(i + length + 1);
+            indices.push((i + length + 1) % mod);
             // second triangle indices
-            indices.push(i + length + 1);
+            indices.push((i + length + 1) % mod);
             indices.push(i);
-            indices.push(i + 1);
+            indices.push((i + 1) % mod);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


peppsac:
> Simple polygons in Feature2Mesh were not extruded properly

Could you add a unit test that would fail before this PR and gets fixed with this PR?

Maybe something that would:
- parse a simple geojson with a single polygon (a square)
- convert it to Mesh without extrusion and verify the number of vertices (should be 4)
- convert it to Mesh with extrusion and verify the number of vertices (should be 8)

What do you think?

zarov:
Yep, I'll do it !

zarov:
@peppsac is the test I added ok for you ?

test/feature2mesh_unit_test.js
```diff
@@ -55,4 +55,19 @@ describe('Feature2Mesh', function () {
...
         }));
+
+    it('square polygon should have double the vertices for being extruded', () =>
+        parse().then((feature) => {
+            const noExtrude = Feature2Mesh.convert()(feature[0]);
+            const extrude = Feature2Mesh.convert({
+                extrude: 2,
+            })(feature[0]);
+
+            const indexExtrude = new Uint16Array([5, 4, 7, 7, 6, 5, 0, 4, 1, 1,
```
peppsac: You're hardcoding in the test the ordering of the vertices. IMHO this isn't a good idea, because the test might fail if this order change - even if the output is correct.

test/feature2mesh_unit_test.js
```diff
@@ -55,4 +55,19 @@ describe('Feature2Mesh', function () {
...
+                4, 5, 1, 5, 2, 2, 5, 6, 2, 6, 3, 3, 6, 7]);
+
+            assert.equal(noExtrude.geometry.attributes.position.count, 4);
+            assert.equal(extrude.geometry.attributes.position.count, 8);
+            assert.equal(extrude.geometry.index.array.every((e, i) => indexExtrude[i] === e), true);
```
peppsac: You can verify the number of triangles instead.

It's less likely to change: it should be 2 triangles per face of the quad/cube. So 2 triangles for the `noExtrude` case and 6*2 triangles for the `extrude`  one.

peppsac (commented):
One small comment + I'd change the scope use in the commit to `vector`.

test/feature2mesh_unit_test.js
zarov: This makes more sense than verifying vertices, indeed.

zarov:
So I updated the test, and it happened that the extrusion was wrong on paper (but good on visuals ?), so I added a fix to it, but I'm really lost here as we already have a good visual on example like `wfs`. Let me know if I missed something here.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -252,12 +257,16 @@ function geometryToExtrudedPolygon(geometry, properties, options, multiGeomAttri
         const holesOffsets = polygonGeometry.holes.map(h => h.offset);
 
         // triangulate the top face
-        const start = vertices.length + polygonGeometry.contour.offset + offset;
+        let start = vertices.length + polygonGeometry.contour.offset + offset;
```
peppsac: Any idea why `end` doesn't depend on ` polygonGeometry.contour.offset`? (or why end isn't `start + count`)?

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -115,6 +115,11 @@ function addExtrudedPolygonSideFaces(indices, length, ring, isClockWise) {
             indices.push(i + 1);
         }
     }
+
+    // last indexes are too high - wrap them inside boundaries
```
peppsac: This is ugly :)

Maybe use the modulo in the previous loop, for all `i + ...` cases.
Something like: `mod = ring.offset + ring.count + length;`
Then:
```js
...
indices.push((i + length) % mod);
indices.push(i);
indices.push((i + length + 1) % mod);
...
```

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: I concede that I didn't think here, but it could make more sense (and more understandable).

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: Yes, but if we do the modulo operation each time on a big ring, we compute it for nothing. But I guess we won't have complex shapes, so I'm okay with changing it.

peppsac:
Did you test your change on a complex polygon? For instance the one from issue #688?

zarov:
I can't make the linked polygon work as a geometry simply, so I stopped for now on this. However, tests with not so complex polygon but still complex, in the WFS example, seems to give the same result as what is currently on the master. I'm working on the loading of unique geojson file as a geometry in #766.

zarov:
I succeeded to display the multipolygon, unfortunately, as you can see below, this PR is breaking things, so I'll be working a bit more on this.

image

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -247,6 +248,46 @@ function geometryToExtrudedPolygon(geometry, properties, options, multiGeomAttri
...
+        const holesOffsets = polygonGeometry.holes.map(h => h.offset);
+
+        // triangulate the top face
+        let start = vertices.length + polygonGeometry.contour.offset + offset;
+        const end = start + count;
+
+        const triangles = Earcut(geom.attributes.position.array.slice(start * 3,
+            start == end ? undefined : end * 3), holesOffsets, 3);
+
+        if (start == end) {
```
peppsac, line 265: Why would this happen?

zarov: In single polygon case, the count is 0, which is the same as `start == end`. Could be better to change it to `count == 0`, my bad.

src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -96,24 +96,27 @@ function addExtrudedPolygonSideFaces(indices, length, ring, isClockWise) {
     // add indices to make two triangle, that make the side face
     const mod = ring.offset + ring.count + length;
     for (let i = ring.offset; i < ring.offset + ring.count; ++i) {
+        const i1 = Math.max((i + 1) % mod, ring.offset);
+        const i2 = Math.max((i + length) % mod, ring.offset);
+        const i3 = Math.max((i + length + 1) % mod, ring.offset);
```
peppsac: Suggestion: `i` and `i1` could be `floor1`/`floor2` and `i2/i3` could be `ceil1` `ceil2` (or `roof1`..) ?

src/Renderer/ThreeExtended/Feature2Mesh.js
peppsac: I'd prefer `count` to have a meaningful value in all case instead :)

So the calling site should be replaced by `addPolygon(geometry, isClockWise, 0, geometry.contour.count);`

src/Renderer/ThreeExtended/Feature2Mesh.js
zarov: Thanks ! Even maybe `bottom1/2` and `top1/2` as floor/ceil/roof is vocabulary for buildings.

zarov:
Closing this PR as everything should be fixed in #756

PR #758, opened 2018-05-18
feat(core): store Layer instead of layer's id

peppsac:
It's more convenient (no need to use `View.getLayers()` to retrieve the real layer) and won't be costlier than storing strings.

(I have another PR coming that depends on this)

### DIFF
--- a/src/Process/ObjectRemovalHelper.js
+++ b/src/Process/ObjectRemovalHelper.js
@@ -21,29 +21,29 @@ export default {
     },
 
     /**
-     * Remove obj's children belonging to layerId layer.
+     * Remove obj's children belonging to a layer.
      * Neither obj nor its children will be disposed!
-     * @param {String} layerId The id of the layer that objects must belong to. Other object are ignored
+     * @param {Layer} layer The layer that objects must belong to. Other object are ignored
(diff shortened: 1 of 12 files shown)


src/Process/ObjectRemovalHelper.js
```diff
@@ -23,27 +23,27 @@ export default {
     /**
      * Remove obj's children belonging to layerId layer.
```
zarov: Forgot to change some comments in this file

src/Provider/TileProvider.js
```diff
@@ -83,7 +83,7 @@ function executeCommand(command) {
 
     geometry._count++;
     const tile = new TileMesh(geometry, params);
-    tile.layer = layer.id;
+    tile.layer = layer;
```
zarov, line 85: I take this one as an example, but you could remove the `layerId` param above.

zarov (commented):
A few things forgotten, but LGTM overall.

peppsac:
Thanks for the review @zarov. I fixed your comments and rebased the PR.

zarov (approved):
Thanks !

PR #759, opened 2018-05-18
Points shader optimization

mbredif:
## Description
This PR simplifies and tentatively optimizes Points shader code.

## Motivation and Context
this PR:
- prevents an [unnecessary mat4 * mat4 multiplication in the VS](https://github.com/iTowns/itowns/compare/points_shader?expand=1#diff-a6598601c3959d96f711a02a6c151364L27)
- [change the 0.01 magic number for automatic size selection to 0](https://github.com/iTowns/itowns/compare/points_shader?expand=1#diff-a6598601c3959d96f711a02a6c151364R35) (but does not look into the automatic size selection itself)
- simplifies to the minimum the FS, moving all color computation to the VS. (to optimize and simplify further, one could remove the bool uniforms and use instead the a pair of attribute and uniform vec4 colors)

apart for the 0.01->0 autosize threshold, the output should be unchanged.

### DIFF
--- a/src/Renderer/PointsMaterial.js
+++ b/src/Renderer/PointsMaterial.js
@@ -1,4 +1,4 @@
-import { Color, Uniform, Vector2, NoBlending, NormalBlending, RawShaderMaterial } from 'three';
+import { Vector4, Uniform, NoBlending, NormalBlending, RawShaderMaterial } from 'three';
 import PointsVS from './Shader/PointsVS.glsl';
 import PointsFS from './Shader/PointsFS.glsl';
 import Capabilities from '../Core/System/Capabilities';
@@ -8,13 +8,13 @@ class PointsMaterial extends RawShaderMaterial {
         super();
         this.vertexShader = PointsVS;
         this.fragmentShader = PointsFS;
+        this.size = size || 0;
+        this.scale = 0.05 * 0.5 / Math.tan(1.0 / 2.0); // autosizing scale
 
-        this.uniforms.size = new Uniform(size);
-        this.uniforms.resolution = new Uniform(new Vector2(window.innerWidth, window.innerHeight));
+        this.uniforms.size = new Uniform(this.size);
         this.uniforms.pickingMode = new Uniform(false);
         this.uniforms.opacity = new Uniform(1.0);
-        this.uniforms.useCustomColor = new Uniform(false);
-        this.uniforms.customColor = new Uniform(new Color());
+        this.uniforms.overlayColor = new Uniform(new Vector4(0, 0, 0, 0));
 
         if (Capabilities.isLogDepthBufferSupported()) {
             this.defines = {
@@ -26,12 +26,19 @@ class PointsMaterial extends RawShaderMaterial {
         if (__DEBUG__) {
             this.defines.DEBUG = 1;
         }
+
+        this.updateUniforms();
     }
 
-    enablePicking(v) {
+    enablePicking(pickingMode) {
         // we don't want pixels to blend over already drawn pixels
-        this.blending = v ? NoBlending : NormalBlending;
-        this.uniforms.pickingMode.value = v;
+        this.blending = pickingMode ? NoBlending : NormalBlending;
+        this.uniforms.pickingMode.value = pickingMode;
+    }
+
+    updateUniforms() {
+        // if size is null, switch to autosizing using the canvas height
+        this.uniforms.size.value = (this.size > 0) ? this.size : -this.scale * window.innerHeight;
     }
 }
 
--- a/src/Renderer/Shader/PointsVS.glsl
+++ b/src/Renderer/Shader/PointsVS.glsl
@@ -4,16 +4,16 @@ precision highp int;
 #include <logdepthbuf_pars_vertex>
 #define EPSILON 1e-6
 
-uniform float size;
-uniform float scale;
+attribute vec3 position;
 uniform mat4 projectionMatrix;
 uniform mat4 modelViewMatrix;
-uniform vec2 resolution;
-uniform bool pickingMode;
+uniform float size;
 
-attribute vec4 unique_id;
+uniform bool pickingMode;
+uniform float opacity;
+uniform vec4 overlayColor;
 attribute vec3 color;
-attribute vec3 position;
+attribute vec4 unique_id;
 
 varying vec4 vColor;
 
@@ -21,21 +21,15 @@ void main() {
     if (pickingMode) {
         vColor = unique_id;
     } else {
-        vColor = vec4(color, 1.0);
+        vColor = vec4(mix(color, overlayColor.rgb, overlayColor.a), opacity);
     }
-    vec4 mvPosition = modelViewMatrix * vec4( position, 1.0 );
-    mat4 projModelViewMatrix = projectionMatrix * modelViewMatrix;
-    gl_Position = projModelViewMatrix* vec4( position, 1.0);
 
-    if (size > 0.01) {
+    gl_Position = projectionMatrix * (modelViewMatrix * vec4( position, 1.0 ));
+
+    if (size > 0.) {
         gl_PointSize = size;
     } else {
-        float pointSize = 1.0;
-        float slope = tan(1.0 / 2.0);
-        float projFactor =  -0.5 * resolution.y / (slope * mvPosition.z);
-
-        float z = min(0.5 * -gl_Position.z / gl_Position.w, 1.0);
-        gl_PointSize = max(3.0, min(10.0, 0.05 * projFactor));
+        gl_PointSize = clamp(-size / gl_Position.w, 3.0, 10.0);
     }
 
     #include <logdepthbuf_vertex>
--- a/src/Process/PointCloudProcessing.js
+++ b/src/Process/PointCloudProcessing.js
@@ -164,8 +164,10 @@ export default {
         if (elt.numPoints > 0) {
             if (elt.obj) {
                 elt.obj.material.visible = true;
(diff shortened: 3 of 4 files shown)


src/Renderer/Shader/PointsVS.glsl
```diff
@@ -4,32 +4,38 @@ precision highp int;
...
 uniform mat4 projectionMatrix;
 uniform mat4 modelViewMatrix;
+
+uniform float size;
 uniform vec2 resolution;
-uniform bool pickingMode;
 
-attribute vec4 unique_id;
+uniform bool pickingMode;
+uniform bool useCustomColor;
```
peppsac: While you're improving this shader, could you define `useCustomColor` as a `float`, that would serve as:
```glsl
vColor = vec4(mix(color, customColor, useCustomColor), opacity);
```
?

mbredif:
done: the vec3 customColor uniform and the hard-coded 0.5 are now replaced by a vec4 rgba uniform : 
- a=0 is vertex color (as with useCustomColor=false)
- a=0.5 corresponds to the previous color with useCustomColor=true
- a=1 is plain rgb color (overwriting the vertex color)

src/Renderer/PointsMaterial.js
```diff
@@ -13,8 +13,7 @@ class PointsMaterial extends RawShaderMaterial {
         this.uniforms.resolution = new Uniform(new Vector2(window.innerWidth, window.innerHeight));
         this.uniforms.pickingMode = new Uniform(false);
         this.uniforms.opacity = new Uniform(1.0);
-        this.uniforms.useCustomColor = new Uniform(false);
-        this.uniforms.customColor = new Uniform(new Color());
+        this.uniforms.rgba = new Uniform(new Vector4(0, 0, 0, 0));
```
autra: nitpicking the name: `customColor` was fine. Or maybe `overlayColor`? We need something more precise than rgba here.

mbredif: agreed. rgba is as bad as customColor...
overlayColor sounds good to me.

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -4,32 +4,35 @@ precision highp int;
...
     vec4 mvPosition = modelViewMatrix * vec4( position, 1.0 );
-    mat4 projModelViewMatrix = projectionMatrix * modelViewMatrix;
-    gl_Position = projModelViewMatrix* vec4( position, 1.0);
+    gl_Position = projectionMatrix * mvPosition;
 
-    if (size > 0.01) {
+    if (size > 0.) {
         gl_PointSize = size;
     } else {
+        // automatic sizing
```
zarov: This entire block could be simplified imho in a single line:
```glsl
gl_PointSize = clamp(-0.05 * resolution.y / gl_Position.w, 3.0, 10.0);
```
I don't see the real point of having this slope, and even if we really want it, it should be precomputed like the `EPSILON`.

Concerning `gl_Position.w`, we could ditch the `mvPosition` above, and precompute the matrices in javascript (and pass the result as a uniform), and directly have `gl_Position = projectionModelViewMatrix * vec4(position, 1.0);`.

zarov: I may be wrong, but we could even only rely on `resolution.y` in fact, by applying a smaller factor (0.003 instead of -0.05 ?)

mbredif: Looking [at the sse code](https://github.com/iTowns/itowns/blob/6b45cf1d4ef863a8ecbf9d884ed545520507f634/src/Process/PointCloudProcessing.js#L55-L67), I am tempted to propose  (do we really need to clamp?)
```
gl_PointSize = (size <0.) ? (size / gl_Position.w) : size;
```
with size being a float uniform set to :
- positive: the sprite size in pixels  or 
- negative: `- context.camera.preSSE * pointSpacing` (possibly multiplied by a user provided tuning factor with a default to 1)

I agree with your proposal for ditching `mvPosition` and passing in a single matrix
(I do not get your second comment)

zarov: Sorry for the bad explanation. I found that `gl_Position = clamp(0.003 * resolution.y, 3.0, 10.0);` gives a pretty good result. If we are ok with this default value, which does not depend from the point position, we could remove the clamp from the shader, and only compute a default size once in javascript. And then we would only have `gl_Position = size;`.

peppsac: @zarov, @mbredif: did you test with different pointclouds (to avoid picking a solution that only works well for one example)?

autra: Just one little remark while passing by: as the calculus is quite simple from the computer POV, we should really aim for readability here. I'm not saying the current code is good regarding this, but that might be a reason to avoid one-liner or to keep the calculus in shader... I expect neither the cost of calculating here nor uploading an additional uniform to really matter here anyway.

For this kind of work, the best way is to test and make screenshots to show reviewers imo :-)

zarov: @peppsac no, but from what I understand: there is no need to compute a varying default point size. The only thing we want is a good ratio to the size of the canvas. If the user wants to set a specific and varying size, he still can. But I agree that this formula might not be the one we want, and other use-cases need to be tested. 

@autra I agree, and by moving to `gl_Position = size;` I think it gives the most readability.

mbredif: For the moment, I just propose a clean up (as in "removing dead code" : bouh! :) ) and add a slight generalization. That will turn
```
        // automatic sizing
        float pointSize = 1.0;
        float slope = tan(1.0 / 2.0);
        float projFactor =  -0.5 * resolution.y / (slope * mvPosition.z);

        float z = min(0.5 * -gl_Position.z / gl_Position.w, 1.0);
        gl_PointSize = max(3.0, min(10.0, 0.05 * projFactor));
```
into
```
        gl_PointSize = clamp(size * scale / gl_Position.w, 3.0, 10.0);
```
or even
```
        gl_PointSize = size * scale / gl_Position.w;
```
The previous automatic behavior would be matched if  `size * scale = (0.05 * -0.5 / tan(1.0 / 2.0)) * resolution.y` (for instance `size = -0.04576219304`, wherever that magic value comes from, and `scale = resolution.y`). That simplification would be almost a perfect match with the [THREE.PointsMaterial code](https://github.com/mrdoob/three.js/blob/master/src/renderers/shaders/ShaderLib/points_vert.glsl#L17-L21)

(@zarov I agree with @peppsac that we should not completely remove automatic sizing)

zarov: Well if you want to remove dead code only, remove only the `pointSize` line and the `float z` which are unused. I'd rather commit a thorough solution if we want to change things inside.

mbredif:
The first commit removes the dead code only.
The last commit simplifies the shader, and lifts a requirement for the material to have a size uniforms.

WDYT ?

zarov:
IMHO the changes in the shader in the last commit should be reverted. As @peppsac said, we can always rework this part when we are sure it doesn't change other cases outside our samples.

mbredif:
To get exactly the previous behavior (but with a customizable scale factor that defaults to the previous magic value), I can add back the clamping [here](https://github.com/iTowns/itowns/pull/759/files#diff-a6598601c3959d96f711a02a6c151364R32) as follows :

```
gl_PointSize = clamp(-size / gl_Position.w, 3.0, 10.0);
```

zarov:
At least yes please.

mbredif:
clamping added: this PR should have no functionality change.

zarov (approved):
LGTM, thanks ! Don't forget to squash before merging.

mbredif:
squashed!

zarov:
Thanks !

src/Renderer/PointsMaterial.js
```diff
@@ -26,12 +26,19 @@ class PointsMaterial extends RawShaderMaterial {
...
         // we don't want pixels to blend over already drawn pixels
-        this.blending = v ? NoBlending : NormalBlending;
-        this.uniforms.pickingMode.value = v;
+        this.blending = pickingMode ? NoBlending : NormalBlending;
+        this.uniforms.pickingMode.value = pickingMode;
+    }
+
+    updateUniforms() {
+        // if size is null, switch to autosizing using the canvas height
+        this.uniforms.size.value = (this.size > 0) ? this.size : -this.scale * window.innerHeight;
```
autra, line 41: @mbredif this isn't the canvas size. Are you sure it displays correctly when the canvas is not "fullscreen"?

mbredif: I am sure it does not, but that is how it was coded (I just made it slightly better by taking the current window.innerHeight rather than the one at material creation time). 

If we trust these pieces of code:
https://github.com/iTowns/itowns/blob/f18548f4bb907dea96efdc620acb0863da5fca88/src/Process/PointCloudProcessing.js#L96-L100 
https://github.com/iTowns/itowns/blob/f18548f4bb907dea96efdc620acb0863da5fca88/src/Process/PointCloudProcessing.js#L59-L66

then something like this might be better :
```
 this.uniforms.size.value = (this.size > 0) ? this.size : -camera.preSSE * layer.metadata.spacing / Math.pow(2, elt.name.length);
```
but that would require passing the necessary parameters and necessitate some testing :). maybe in another PR ?

autra: >  that is how it was coded  

oh right, it was there before this PR, my bad... I was thinking of using at least the canvas height instead of the windows height at least. But yeah, another PR another day :-) Thanks for pointing this to me!

PR #760, opened 2018-05-18
refactor(pointsmaterial) : shared potree material among point tiles (no cloning)

mbredif:
## Description
Potree parsers used a hardcoded itowns.PointsMaterial  material
This PR tackles some part of #695 to decouple the parsing and the styling of potree pointclouds.

## Change
- Potree parsers now output a THREE.BufferGeometry, rather than a THREE.Points.
- The potree PointCloudProvider (to be renamed more explicitly as well as the pointcloud examples by the way :) ) is now responsible for creating the THREE.Points and adding the layer material (which defaults to itowns.PointsMaterial.
- customBinFormat is now a local variable (private implementation detail)
- realPointCount is removed as it was not used apparently (if needed, it can be accessed using the size of the position attribute)

## Screenshots
No visible change

## Question
I am not sure I understand why I need to [clone](https://github.com/iTowns/itowns/compare/points_material#diff-4033c2db9cc0d018b29c8a27df1bf722R240) the material. Ideally, I would like to use the same material for all pointcloud tiles.
@peppsac : any clue ?

### DIFF
--- a/examples/pointcloud.html
+++ b/examples/pointcloud.html
@@ -234,11 +234,14 @@
                         // update stats window
                         oldPostUpdate = pointcloud.postUpdate;
                         pointcloud.postUpdate = function postUpdate() {
-                            var info = document.getElementById('info');
                             oldPostUpdate.apply(pointcloud, arguments);
+                            pointcloud.statsUpdate();
+                        };
+                        pointcloud.statsUpdate = function statsUpdate() {
+                            var info = document.getElementById('info');
                             info.textContent = 'Nb points: ' +
                                 pointcloud.displayedCount.toLocaleString();
-                        };
+                        }
                     }
                     window.view = view;
 
--- a/src/Process/PointCloudProcessing.js
+++ b/src/Process/PointCloudProcessing.js
@@ -68,10 +68,10 @@ function computeScreenSpaceError(context, layer, elt, distance) {
 
 function markForDeletion(elt) {
     if (elt.obj) {
-        elt.obj.material.visible = false;
+        elt.obj.visible = false;
         if (__DEBUG__) {
             if (elt.obj.boxHelper) {
-                elt.obj.boxHelper.material.visible = false;
+                elt.obj.boxHelper.visible = false;
             }
         }
     }
@@ -93,6 +93,7 @@ export default {
             return [];
         }
 
+
         // See https://cesiumjs.org/hosted-apps/massiveworlds/downloads/Ring/WorldScaleTerrainRendering.pptx
         // slide 17
         context.camera.preSSE =
@@ -104,6 +105,9 @@ export default {
             layer.material.opacity = layer.opacity;
             layer.material.transparent = layer.opacity < 1;
             layer.material.size = layer.pointSize;
+            if (layer.material.updateUniforms) {
+                layer.material.updateUniforms();
+            }
         }
 
         // lookup lowest common ancestor of changeSources
@@ -165,11 +169,7 @@ export default {
         // only load geometry if this elements has points
         if (elt.numPoints > 0) {
             if (elt.obj) {
-                if (elt.obj.material.update) {
-                    elt.obj.material.update(layer.material);
-                } else {
-                    elt.obj.material.copy(layer.material);
-                }
+                elt.obj.visible = true;
                 if (__DEBUG__) {
                     if (layer.bboxes.visible) {
                         if (!elt.obj.boxHelper) {
@@ -236,7 +236,7 @@ export default {
 
         layer.displayedCount = 0;
         for (const pts of layer.group.children) {
-            if (pts.material.visible) {
+            if (pts.visible) {
                 const count = pts.geometry.attributes.position.count;
                 pts.geometry.setDrawRange(0, count);
                 layer.displayedCount += count;
@@ -251,12 +251,12 @@ export default {
                 // representation
                 const reduction = layer.pointBudget / layer.displayedCount;
                 for (const pts of layer.group.children) {
-                    if (pts.material.visible) {
+                    if (pts.visible) {
                         const count = Math.floor(pts.geometry.drawRange.count * reduction);
                         if (count > 0) {
                             pts.geometry.setDrawRange(0, count);
                         } else {
-                            pts.material.visible = false;
+                            pts.visible = false;
                         }
                     }
                 }
@@ -272,7 +272,7 @@ export default {
                 for (const pts of layer.group.children) {
                     const count = pts.geometry.attributes.position.count;
                     if (limitHit || (layer.displayedCount + count) > layer.pointBudget) {
-                        pts.material.visible = false;
+                        pts.visible = false;
                         limitHit = true;
                     } else {
                         layer.displayedCount += count;
@@ -284,11 +284,11 @@ export default {
         const now = Date.now();
         for (let i = layer.group.children.length - 1; i >= 0; i--) {
             const obj = layer.group.children[i];
-            if (!obj.material.visible && (now - obj.owner.notVisibleSince) > 10000) {
+            if (!obj.visible && (now - obj.owner.notVisibleSince) > 10000) {
                 // remove from group
                 layer.group.children.splice(i, 1);
 
-                obj.material.dispose();
+                // no need to dispose obj.material, as it is shared by all objects of this layer
                 obj.geometry.dispose();
                 obj.material = null;
                 obj.geometry = null;
--- a/src/Provider/PointCloudProvider.js
+++ b/src/Provider/PointCloudProvider.js
@@ -246,7 +246,7 @@ export default {
         const url = `${node.baseurl}/r${node.name}.${layer.extension}?isleaf=${command.isLeaf ? 1 : 0}`;
 
         return Fetcher.arrayBuffer(url, layer.fetchOptions).then(layer.parse).then((geometry) => {
-            const points = new THREE.Points(geometry, layer.material.clone());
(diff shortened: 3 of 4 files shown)


peppsac:
You could use this PR to work towards one important aspect needed for #695: unifying parser output.

So instead of adding a new possibility for parser output (we already have meshes and Features), I'd prefer that you use the same output as the geojson parser - improved if needed (maybe using a `Float32Array` for vertices + a crs attributes instead of `Coordinates`?).

mbredif:
Here, I was more targetting the "Move symbolizing code out of formatters" facet of #695.
I would rather have simpler single-purpose PRs : unifying the parser outputs will need some discussions and touch more files... (and I agree that the current Feature is not suitable for pointclouds as we definitely need it to host typedarrays to enable direct uploading to the gpu whenever possible) 


Concerning my question on the need to clone the material, I figured out that the issue was triggered by the pointcloud process that was setting `pts.material.visible` instead of `pts.visible` (see fa5f613).
- What is the rationale to set the material visibility (that could be shared by different objects) rather than the object visibility ?
- Would not it be simpler and more efficient if all the objects created by a layer shared a unique material `layer.material`, as sketched in fa5f613?

It might be that both per-layer and per-tile materials make sense depending on the context and that we might end up supporting both (eg by accepting layer.material as a material instance to be shared or as a material creator function called for each tile)

peppsac:
If you want to use your own material, this PR is not needed, you can already do it:
```js
layer.onPointsCreated = (layer, pts) => {
   pts.material = mymaterial;
};
```

That's why I'm suggesting to do something more useful for the long term. I don't request that you modify all parsers; you can simply improve the ones you're interested in (`PotreeBinParser` and `PotreeCinParser`).

> I agree that the current Feature is not suitable for pointclouds as we definitely need it to host typedarrays to enable direct uploading to the gpu whenever possible

Great, then open another PR to fix this as well :)
Despite the smiley I'm serious: this would be useful for your issue and other parts of itowns.

mbredif:
> If you want to use your own material, this PR is not needed, you can already do it:

unfortunately, it is. Your snippet is likely not working, due to the current handling of the visibility in the pointcloud process (see fa5f613)...

> Great, then open another PR to fix this as well :)

I can open an issue to start the discussion. Whenever someone has time to work on that, he will then have some material to base its work on...

peppsac:
> Your snippet is likely not working, due to the current handling of the visibility in the pointcloud process (see fa5f613)...

Can you elaborate (I don't understand the problem)?
The two visible come from three.js:
   * mesh.visible means the mesh is visible (as in not culled)
   * material.visible means the object will be displayed
As far as I can tell, the `PointCloudProcessing.js` respects this convention.

mbredif:
If you use a single material shared by all objects and the process modifies material.visible, then either all or no objects of the layer will be displayed (depending on the status of the last updated object)

> mesh.visible means the mesh is visible (as in not culled)

indeed, that is lod-culling !

mbredif:
#761 shows the shared material bug in action

peppsac:
IMHO you need to clone the `material`. That's the way `three.js` works: if you use `point.visible = false` it'll also hide all its children.


So, with this error fixed in my snippet, you can do:
```js
layer.onPointsCreated = (layer, pts) => {
   pts.material = mymaterial.clone();
};
```

mbredif:
> IMHO you need to clone the material

what do you mean by "need" ? this PR achieves its goal of a shared material decoupled from the parsers without cloning the material...

>  That's the way three.js works: if you use point.visible = false it'll also hide all its children.

...which does not matter here as all nodes are added to a flat group (no hierarchy), as far as I can tell. :)

peppsac:
> what do you mean by "need"

You need because of the meaning of the 2 `visible` property in three.js.

> ...which does not matter here as all nodes are added to a flat group (no hierarchy), as far as I can tell. :)

Yes, but you're only looking at itowns code. You don't know how it'll be used and you can't assume that users of itowns will never add children to points objects.

mbredif:
I do not know what they are doing, but if user add children to points objects, which is not a documented behavior (and might interfere with pointcloudprocess logic), they might be happy to see that these objects get invisible whenever the points get invisible. That's actually a very good idea to simplify the box helper code, isn't it ? :)

If we cannot agree on the second commit for now, what do you think of the first one ?

peppsac:
> which is not a documented behavior (and might interfere with pointcloudprocess logic)

This is fully documented: it's the core behavior of three.js.

>  they might be happy to see that these objects get invisible whenever the points get invisible.

Sorry but this is wrong as well. We mark invisible points as invisible ([here](https://github.com/iTowns/itowns/blob/master/src/Process/PointCloudProcessing.js#L155)). But we also hide visible points when they are not needed.
This is not the same thing, and I don't think it's reasonnable for iTowns to mix the 2 concepts.

So not cloning the material would be handy for a few things but I don't think it's possible.

peppsac:
> If we cannot agree on the second commit for now, what do you think of the first one ?

I think I already answered to the first one here: https://github.com/iTowns/itowns/pull/760#issuecomment-390911368

mbredif:
I understand that you prefer to close this PR unmerged.
- You recommend using the onPointsCreated hook rather than introducing a `material` creation parameter ?
- You do not want to decouple styling from parsing by leaving the [dependency on PointsMaterial](https://github.com/iTowns/itowns/pull/760/commits/e40680c658b0691b2f287de5778ee4572d0816ef#diff-e5b83e61d7ca18dd4b4346d9b978cb70L2) and the [rendering logic](https://github.com/iTowns/itowns/blob/master/src/Parser/PotreeCinParser.js#L32-L37) in the potree parsers.

mbredif:
The first part of this PR is ready to review : #764 (with 1 material per object).
This PR is now rebased on it (6084200888b96e75897c9f308940c5739f120baf) with a single material per layer (no cloning).

peppsac:
Could you rebase this PR please?

mbredif:
done! I added a second commit to fix some issues in the debug helper. I am not satisfied on the call to updatestats in the helper as this function is defined in the example. do you have a better proposition for fixing the display of the stats ?

peppsac:
> done!

Thanks.

I'm still not convinced that this is good idea because that goes against the logic exposed by threejs, so IMHO this shouldn't be merged.

mbredif:
Not sure about what the 'logic of threejs' is, about sharing materials across objects. Do you have a pointer ?

Both approaches (master and this PR) work and offer different tradeoffs, depending if you want more or less to hide to the user that the layer is managing internally a collection of objects by exposing a single `layer.material`, and if you are more or less concerned by syncing issues for material properties :
- I think the current implementation works on stock threejs materials but the user have extra work to keep cloned materials (cf `updateUniforms()`, but that may not be sufficient if more than the uniforms need to be synced (`defines`, custom properties...), whereas this PR does not clone any material so no sync is required.
- for use cases that need to modify the material on a per-object basis you are likely using `onPointsCreated` so you can clone the material there and setup your custom update logic

let us keep this around for a while, until we settle on a common approach for all itowns materials :)

peppsac:
> Not sure about what the 'logic of threejs' is, about sharing materials across objects. Do you have a pointer ?

See https://github.com/iTowns/itowns/pull/760#issuecomment-390929552, and https://github.com/iTowns/itowns/pull/760#issuecomment-391248407 and https://github.com/iTowns/itowns/pull/760#issuecomment-391276243 

In threejs, a material has per object properties  (at the very least the `visible` property) so you can't use one instance for many objects.

Feel free to prove me wrong, but I don't want to merge something that breaks threejs semantics (again, especially for the .visible property).

mbredif:
I meant pointers to comments not written by you ;). For reference, there are at least 2 `visible` properties : This PR shows that using the `visible` property from the object works as well. Let's not argue too much on for now and wait for use cases that would help us decide to merge or close this PR.

I found the pointcloud debug helper a little buggy (eg: sync issues for the stats and the visibility of boxes) and lacked the option to show only the sticky node. Would you be interested in writing a PR for the second commit only ?

autra:
>  I found the pointcloud debug helper a little buggy (eg: sync issues for the stats and the visibility of boxes) and lacked the option to show only the sticky node.

untick "display parent" and "display children". It'll display only this node.

> I think the current implementation works on stock threejs materials but the user have extra work to keep cloned materials (cf updateUniforms())

But that's a flaw to the particular solution chosen in #764 . In short, we (I included) have complicated our life on current master (properties are in layer.foo, then layer.material.foo, then layer.material.uniforms.foo, leading indeed to painful and difficult synchronization process). This can be a lot simpler, without requiring to keep the same material.

mbredif:
> layer.foo

I am for removing this one, and keep all stylization to `layer.material`, which could be renamed to `layer.style` to give it a more user-friendly name. Even `layer.opacity` could go there. (but that is already the discussion on the material/style normalization)

>layer.material.foo

This one sounds good and should be kept. The good thing is that it does not impose anything on the material (it is free to use or not this new property).

>then layer.material.uniforms.foo

The issue here is that materials may not have a uniforms property. Updating them should be a internal concern of the material

You are missing the syncing of the `object.material(.uniform).foo` of all tiles created by the layer if they are cloned (this PR is all about not having to care about this one !)

> painful and difficult synchronization process

I agree, but it is only the downside of garanteeing the sync of the layer options with the object materials (I consider the previous behavior of only copying layer properties at tile creation time a bug). This PR is an alternative approach with the same sync garantee, there are surely other way to achieve this goal.

gchoqueux:
this PR is rebased in  #1589

PR #761, opened 2018-05-22
[bug] [do not merge] single material bug in point clouds

mbredif:
## Description
@peppsac suggested in [comment](https://github.com/iTowns/itowns/pull/760#issuecomment-390911368) that a shared custom material may be assigned to the potree tiles as follows :

```
layer.onPointsCreated = (layer, pts) => {
   pts.material = mymaterial;
};
```

This PR (DO NOT MERGE!) shows that it does not work (at least not the way I understand the comment).

## Motivation and Context
due to the coupling of the parsers/provider/process for potree tiles, it does not work : 
- the custom material may not have a `uniforms` property (this PR provides a [workaround](https://github.com/iTowns/itowns/pull/761/files#diff-6b60bae2240ab556c955b4eb0a07a12eL167) for this)
- the process sets the visibility of the tile material instead of the visibility of the tile itself, so that the visibility status is then shared by all tiles (either all or none are displayed, depending on the last process update)

## Fix
#760 works towards fixing this bug.

### DIFF
--- a/examples/pointcloud.html
+++ b/examples/pointcloud.html
@@ -204,6 +204,13 @@
                         view.notifyChange(true);
                     }
 
+                    var material = new itowns.THREE.PointsMaterial({
+                        color: 0xff8888,
+                        sizeAttenuation: false,
+                        size: 1,
+                        vertexColors: itowns.THREE.VertexColors
+                    });
+
                     // add pointcloud to scene
                     function onLayerReady() {
                         var ratio;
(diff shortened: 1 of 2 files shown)


zarov:
Apart from the initial subject, you should open an issue instead, where you can reference the branch/commit that concerns this bug. No need to open a PR if it isn't intended to be merged. I'll let you open a new one.

mbredif:
OK, I will do that next time, thanks for educating me :)

PR #764, opened 2018-05-23
refactor(parser) : move Materials from potree parsers to potree provider (+custom layer material option)

mbredif:
## Description
Potree parsers used a hardcoded itowns.PointsMaterial  material
This PR tackles some part of #695 to decouple the parsing and the styling of potree pointclouds.

## Change
- Potree parsers now output a THREE.BufferGeometry, rather than a THREE.Points.
- The potree PointCloudProvider (to be renamed more explicitly as well as the pointcloud examples by the way :) ) is now responsible for creating the THREE.Points and adding the layer material (which defaults to itowns.PointsMaterial : Potree Parser now have no knowledge of materials and rendering logic.
- customBinFormat is now a local variable (private implementation detail)
- realPointCount is removed as it was not used apparently (if needed, it can be accessed using the size of the position attribute)
- the pointcloud.html example shows a sample usage of potree layer with a custom material : [a simple upstream THREE.PointsMaterial](https://github.com/iTowns/itowns/compare/potree_custom_material?expand=1#diff-d619f5f9635467ba864c74d37cfcda7fR185).

## Screenshots
- http://localhost:8080/examples/pointcloud.html : No visible change
- http://localhost:8080/examples/pointcloud.html?material=three
image
(note the red diffuse color and the square sprites)

## Note
This PR is a slight rework of the first commit of #760. The second commit that allows sharing the material with no cloning will be reworked in its own PR.

### DIFF
--- a/examples/pointcloud.html
+++ b/examples/pointcloud.html
@@ -152,6 +152,7 @@
             <script src="js/loading_screen.js"></script>
             <script src="../dist/debug.js"></script>
             <script type="text/javascript">
+                var searchParams = new URL(window.location.href).searchParams;
                 function showPointcloud(serverUrl, fileName, lopocsTable) {
                     var pointcloud;
                     var oldPostUpdate;
@@ -182,6 +183,14 @@
                     pointcloud.protocol = 'potreeconverter';
                     pointcloud.url = serverUrl;
                     pointcloud.table = lopocsTable;
+                    if (searchParams.get('material') === 'three') {
+                        pointcloud.material = new itowns.THREE.PointsMaterial({
+                            color: 0xff8888,
+                            sizeAttenuation: false,
+                            size: 1,
+                            vertexColors: itowns.THREE.VertexColors
+                        });
+                    }
 
                     // point selection on double-click
                     function dblClickHandler(event) {
@@ -236,8 +245,7 @@
                     view.addLayer(pointcloud).then(onLayerReady);
                 }
 
-
-                if (new URL(window.location.href).searchParams.get('selector')) {
+                if (searchParams.get('selector')) {
                     document.getElementsByClassName('centered')[0].style.display = 'block';
                     document.getElementById('submit').addEventListener('click', onConnectClicked);
                 } else {
--- a/src/Process/PointCloudProcessing.js
+++ b/src/Process/PointCloudProcessing.js
@@ -99,6 +99,12 @@ export default {
             context.camera.height /
                 (2 * Math.tan(THREE.Math.degToRad(context.camera.camera3D.fov) * 0.5));
 
+        if (layer.material) {
+            layer.material.opacity = layer.opacity;
+            layer.material.transparent = layer.opacity < 1;
+            layer.material.size = layer.pointSize;
+        }
+
         // lookup lowest common ancestor of changeSources
         let commonAncestorName;
         for (const source of changeSources.values()) {
@@ -158,11 +164,7 @@ export default {
         // only load geometry if this elements has points
         if (elt.numPoints > 0) {
             if (elt.obj) {
-                elt.obj.material.visible = true;
-                elt.obj.material.size = layer.pointSize;
-                if (elt.obj.material.updateUniforms) {
-                    elt.obj.material.updateUniforms();
-                }
+                elt.obj.material.copy(layer.material);
                 if (__DEBUG__) {
                     if (layer.bboxes.visible) {
                         if (!elt.obj.boxHelper) {
--- a/src/Provider/PointCloudProvider.js
+++ b/src/Provider/PointCloudProvider.js
@@ -3,6 +3,7 @@ import Fetcher from './Fetcher';
 import PointCloudProcessing from '../Process/PointCloudProcessing';
 import PotreeBinParser from '../Parser/PotreeBinParser';
 import PotreeCinParser from '../Parser/PotreeCinParser';
+import PointsMaterial from '../Renderer/PointsMaterial';
 import Picking from '../Core/Picking';
 
 // Create an A(xis)A(ligned)B(ounding)B(ox) for the child `childIndex` of one aabb.
@@ -140,17 +141,6 @@ function addPickingAttribute(points) {
     return points;
 }
 
-
-function loadPointFile(layer, url) {
-    return fetch(url, layer.fetchOptions).then(foo => foo.arrayBuffer()).then((ab) => {
-        if (layer.metadata.customBinFormat) {
-            return PotreeCinParser.parse(ab).then(result => addPickingAttribute(result));
-        } else {
-            return PotreeBinParser.parse(ab).then(result => addPickingAttribute(result));
-        }
-    });
-}
-
 export default {
     preprocessDataLayer(layer) {
         if (!layer.file) {
@@ -176,6 +166,8 @@ export default {
         layer.pointSize = layer.pointSize === 0 || !isNaN(layer.pointSize) ? layer.pointSize : 4;
         layer.sseThreshold = layer.sseThreshold || 2;
         layer.type = 'geometry';
+        layer.material = layer.material || {};
+        layer.material = layer.material.isMaterial ? layer.material : new PointsMaterial(layer.material);
 
         // default update methods
         layer.preUpdate = PointCloudProcessing.preUpdate;
@@ -189,6 +181,7 @@ export default {
             layer.metadata = cloud;
 
             let bbox;
+            var customBinFormat = true;
 
             // Lopocs pointcloud server can expose the same file structure as PotreeConverter output.
             // The only difference is the metadata root file (cloud.js vs infos/sources), and we can
@@ -196,7 +189,7 @@ export default {
             // (if `scale` is defined => we're fetching files from PotreeConverter)
             if (layer.metadata.scale != undefined) {
                 // PotreeConverter format
-                layer.metadata.customBinFormat = layer.metadata.pointAttributes === 'CIN';
+                customBinFormat = layer.metadata.pointAttributes === 'CIN';
                 bbox = new THREE.Box3(
                     new THREE.Vector3(cloud.boundingBox.lx, cloud.boundingBox.ly, cloud.boundingBox.lz),
                     new THREE.Vector3(cloud.boundingBox.ux, cloud.boundingBox.uy, cloud.boundingBox.uz));
@@ -205,7 +198,7 @@ export default {
                 layer.metadata.scale = 1;
                 layer.metadata.octreeDir = `itowns/${layer.table}.points`;
                 layer.metadata.hierarchyStepSize = 1000000; // ignore this with lopocs
-                layer.metadata.customBinFormat = true;
+                customBinFormat = true;
 
                 let idx = 0;
                 for (const entry of cloud) {
@@ -219,7 +212,9 @@ export default {
                    new THREE.Vector3(cloud[idx].bbox.xmax, cloud[idx].bbox.ymax, cloud[idx].bbox.zmax));
             }
 
-            layer.supportsProgressiveDisplay = layer.metadata.customBinFormat;
+            layer.parse = customBinFormat ? PotreeCinParser.parse : PotreeBinParser.parse;
+            layer.extension = customBinFormat ? 'cin' : 'bin';
+            layer.supportsProgressiveDisplay = customBinFormat;
 
             return parseOctree(
                     layer,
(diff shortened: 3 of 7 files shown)


mbredif:
Rebased and ready to review ! (after #764)

src/Provider/PointCloudProvider.js
```diff
@@ -255,7 +253,10 @@ export default {
             points.tightbbox.max.z *= layer.metadata.scale;
             points.tightbbox.translate(node.bbox.min);
             points.material.transparent = layer.opacity < 1.0;
-            points.material.uniforms.opacity.value = layer.opacity;
+            points.material.opacity = layer.opacity;
+            if (points.material.refreshUniforms) {
+                points.onBeforeRender = points.material.refreshUniforms.bind(points.material);
```
peppsac: Just a comment here: I'd prefer to keep the old mechanism (update during process).

That way the `onBeforeRender` callbacks provided by `three` are still available for the user.

mbredif:
@autra any comment on this ? (it is complementary with #767 )

src/Provider/PointCloudProvider.js
autra: +1 :-)

src/Provider/PointCloudProvider.js
```diff
@@ -236,15 +230,19 @@ export default {
...
         // when we request .hrc files)
-        const url = `${node.baseurl}/r${node.name}.${extension}?isleaf=${command.isLeaf ? 1 : 0}`;
+        const url = `${node.baseurl}/r${node.name}.${layer.extension}?isleaf=${command.isLeaf ? 1 : 0}`;
+
+        return Fetcher.arrayBuffer(url, layer.fetchOptions).then(layer.parse).then((geometry) => {
+            const points = new THREE.Points(geometry, layer.material.clone());
+            addPickingAttribute(points);
+            points.frustumCulled = false;
+            points.matrixAutoUpdate = false;
+            points.tightbbox = geometry.boundingBox;
```
autra: points.tightbbox is actually never used directly, but copied into elt.tightbbox in processing. As we now use geometry.boundingBox, maybe we can take this opportunity to remove an intermediary step and store geometry.boundingBox directly into elt.tightbbox in processing? Also it'll allow elt.tightbbox be always up-to-date if we always use `geometry.computeBoundingBox()` after that.

src/Renderer/PointsMaterial.js
```diff
@@ -36,9 +35,10 @@ class PointsMaterial extends RawShaderMaterial {
         this.uniforms.pickingMode.value = pickingMode;
     }
 
-    updateUniforms() {
+    refreshUniforms(/* renderer, scene, camera, geometry, material, group */) {
```
autra: not a big deal, but why did you change `update...`  to `refresh...` ?

examples/pointcloud.html
```diff
@@ -182,6 +182,14 @@ <h3>Available pointclouds:</h3>
                     pointcloud.protocol = 'potreeconverter';
                     pointcloud.url = serverUrl;
                     pointcloud.table = lopocsTable;
+                    if (new URL(window.location.href).searchParams.get('material') === 'three') {
```
autra: nits: `new URL(...).searchParams` is used 2 times, let's define a `var searchParams` on top.

autra (changes requested):
> it is complementary with #767

Not really? Or I didn't see the link apart the fact both touches pointclouds...

But anyway, this looks good to me overall, apart the `onBeforeRender` thing.

It makes sense to return a three.js object here (and not to return a Feature in this case), because atm we use feature only to have either mesh or raster, and I don't see a use case for pointclouds in raster form (yet).

src/Provider/PointCloudProvider.js
```diff
@@ -255,7 +253,10 @@ export default {
             points.tightbbox.max.z *= layer.metadata.scale;
             points.tightbbox.translate(node.bbox.min);
             points.material.transparent = layer.opacity < 1.0;
-            points.material.uniforms.opacity.value = layer.opacity;
+            points.material.opacity = layer.opacity;
```
peppsac: `points.material.uniforms.overlayColor` should be set as well. Can you add this?

mbredif: this is now done by the inside the PointsMaterial copy function (I dont want anything material specific such as overlayColor in the provider!)

examples/pointcloud.html
mbredif: done (9529b2c)

src/Renderer/PointsMaterial.js
mbredif: changed back to `updateUniforms` (I was following the naming pattern in three for other onbeforerender callbacks)

src/Provider/PointCloudProvider.js
mbredif: I did not want to change to much in the `PointCloudProcess`, so I only simplified the tightbox computation to a one liner in the  `PointCloudProvider`. https://github.com/iTowns/itowns/blob/ab92777fdaeaaf5a488772701e2156858c0256d0/src/Provider/PointCloudProvider.js#L248

mbredif:
> Not really? Or I didn't see the link apart the fact both touches pointclouds...

 #767 makes all attributes available, this PR lets you provide a custom material to the potree layer, so that client code may design any fancy styling of all these attributes.

> But anyway, this looks good to me overall, apart the onBeforeRender thing.

onBeforeRender is now untouched.

@autra, WDYT ?

src/Process/PointCloudProcessing.js
```diff
@@ -163,11 +166,7 @@ export default {
         // only load geometry if this elements has points
         if (elt.numPoints > 0) {
             if (elt.obj) {
-                elt.obj.material.visible = true;
-                elt.obj.material.size = layer.pointSize;
-                if (elt.obj.material.updateUniforms) {
-                    elt.obj.material.updateUniforms();
-                }
+                elt.obj.material.copy(layer.material);
```
mbredif, line 167: this implies `elt.obj.material.visible = true;` if the `layer.material` is visible

src/Provider/PointCloudProvider.js
```diff
@@ -236,27 +231,21 @@ export default {
...
             points.scale.set(layer.metadata.scale, layer.metadata.scale, layer.metadata.scale);
-            points.tightbbox.min.x *= layer.metadata.scale;
-            points.tightbbox.min.y *= layer.metadata.scale;
-            points.tightbbox.min.z *= layer.metadata.scale;
-            points.tightbbox.max.x *= layer.metadata.scale;
-            points.tightbbox.max.y *= layer.metadata.scale;
-            points.tightbbox.max.z *= layer.metadata.scale;
-            points.tightbbox.translate(node.bbox.min);
-            points.material.transparent = layer.opacity < 1.0;
-            points.material.uniforms.opacity.value = layer.opacity;
```
mbredif, line 266: material update is now triggered in the Process update (previously transparency was not updated !)

mbredif:
updated and ready for review ! (testing is fixed)

utils/debug/PointCloudDebug.js
```diff
@@ -10,39 +10,36 @@ function isInHierarchy(elt, hierarchyNode) {
...
         layer.dbgEnableStickyNode = false;
         layer.dbgStickyNode = '';
         layer.dbgDisplayChildren = true;
         layer.dbgDisplayParents = true;
         layer.dbgDisplaybbox = false;
 
+        var styleUI = layer.debugUI.addFolder('Styling');
+        styleUI.add(layer, 'opacity', 0, 1).name('Layer Opacity').onChange(update);
+        styleUI.add(layer, 'pointSize', 0, 15).name('Point Size').onChange(update);
+        styleUI.add(layer.material, 'visible').name('Display Points').onChange(update);
```
autra: The same checkbox is called `Visible` for other layers. It makes sense because it also hides the bounding box btw. 

Also for me it's not a matter of styling but more something useful to hide temporarily a layer. Can you move this on the parent folder?

autra (commented):
LGTM, just a little cosmetic change on the debug tools and we can merge that

autra:
Also can you resolve the conflicts and squash your commits?

mbredif:
resolved and squashed !

utils/debug/PointCloudDebug.js
```diff
@@ -10,39 +10,36 @@ function isInHierarchy(elt, hierarchyNode) {
...
         layer.dbgEnableStickyNode = false;
         layer.dbgStickyNode = '';
         layer.dbgDisplayChildren = true;
         layer.dbgDisplayParents = true;
         layer.dbgDisplaybbox = false;
 
+        var styleUI = layer.debugUI.addFolder('Styling');
+        styleUI.add(layer, 'opacity', 0, 1).name('Layer Opacity').onChange(update);
+        styleUI.add(layer, 'pointSize', 0, 15).name('Point Size').onChange(update);
+        styleUI.add(layer.material, 'visible').name('Visible').onChange(update);
```
autra: I'd really move it in the parent folder, it's very useful to be able to hide / show layers without having to unfold something, WDYT?

mbredif: no strong opinion. I have just amended the commit to have the following UI layout :

image

autra (approved):
Thanks!

autra:
Actually, it seems that the case where point size is 0 is broken... Could you have a look please?

autra (commented):
Dismissing my review to avoid accidental merge.

autra (changes requested):
(seems I can't dismiss :-( )

mbredif:
There is no automatic sizing with the THREE material so testing http://localhost:8080/examples/pointcloud.html?material=three with size=0 will give you 0-sized points (hence nothing to see). Is that what you are talking about ?

autra (approved):

autra:
Oh right, silly me! Sorry for the noise :-)

LGTM, thanks!

utils/debug/PointCloudDebug.js
```diff
@@ -10,39 +10,36 @@ function isInHierarchy(elt, hierarchyNode) {
...
+        const update = () => view.notifyChange(layer);
         layer.debugUI = datUi.addFolder(`${layer.id}`);
 
-        layer.debugUI.add(layer, 'sseThreshold').name('SSE threshold')
-            .onChange(() => view.notifyChange(layer));
-        layer.debugUI.add(layer, 'octreeDepthLimit', -1, 20).name('Depth limit')
-            .onChange(() => view.notifyChange(layer));
-        layer.debugUI.add(layer, 'pointBudget', 1, 15000000).name('Max point count')
-            .onChange(() => view.notifyChange(layer));
+        layer.debugUI.add(layer.material, 'visible').name('Visible').onChange(update);
```
peppsac, line 16: I missed this before the merge, but this breaks other parts of itowns which expect `layer.visible`:
at the very least the test here https://github.com/iTowns/itowns/blob/master/src/Core/MainLoop.js#L111 is broken :(

It also has a different behavior than `3dtiles` layer which already supported having a layer.material property: in the 3dtiles case the material is cloned and layer.material is never copied in object.material.

Btw, https://github.com/iTowns/itowns/pull/764/files#diff-6b60bae2240ab556c955b4eb0a07a12eR167 is a bad idea as well IMHO: a user can't set one parameter for a given point object material because it will always be overriden with the one from the layer (so the overlay color feature can now only be enabled globally).

(edit: fixed the 2nd link)

src/Renderer/PointsMaterial.js
```diff
@@ -26,19 +28,31 @@ class PointsMaterial extends RawShaderMaterial {
...
+    }
+
+    copy(source) {
+        this.visible = source.visible;
+        this.opacity = source.opacity;
+        this.transparent = source.transparent;
+        this.size = source.size;
+        this.scale = source.scale;
+        this.overlayColor.copy(source.overlayColor);
+        this.updateUniforms();
```
peppsac, line 54: Another too late comment: you're overriding three.js `copy` method but you don't call the `super` method.
This should be fixed:
- either rename your `copy` method and keep its current code
- keep the `copy` name and call RawShaderMaterial.copy as well

autra:
@mbredif Sorry for not catching these problems during my review :-/

I just restored the branch and I'm going to revert this commit for now, if that's ok for you.

zarov:
> I just restored the branch and I'm going to revert this commit for now, if that's ok for you.

Please no, we don't need to revert commits every time we missed something in a review. Another PR will be opened, adding fixes to the remarks that have not been answered. The mistake is human, you validated it, no problem, it happens. If there are bugs, we should open an issue.

autra:
@zarov as you wish :-)

utils/debug/PointCloudDebug.js
mbredif: > I missed this before the merge, but this breaks other parts of itowns which expect layer.visible:

For this one a simple fix would be to treat `visible` as `opacity` is treated, with a copy in the `preUpdate`.

mbredif: For the rest of the comment, I left `points.onBeforeRender` untouched just for that use case :).

That way, I think you can still get your per-tile functionality, while fixing the bug present before the merge : if you change styling properties on the layer, then it will only apply to newly created tiles, and not to previously created tiles (just take an earlier commit, change the point size and move around to see that in action).

That being said, maybe we should open an issue to agree and document properly a common strategy for the style/material of all layers.

src/Renderer/PointsMaterial.js
mbredif: I have just tried proper overriding, but it does not work (unless you manually set `this.uniforms = source.uniforms` after the call to `super`). I do not know if it is a misuse or an upstream bug, so meanwhile, renaming `copy` to `update` could do the trick.

utils/debug/PointCloudDebug.js
peppsac: I agree there was a bug before the merge.

I disagree that `onBeforeRender` is an acceptable substitute tough: having to register a callback for each object you want to customize vs modifying their material once is quite different (looks very much like pushing the complexity on the user of itowns to ease the implementation in itowns core).

That being said, this is a minor feature, so we can discuss this in an issue and try to find a solution that works for all kind of layer.

In the meantime please fix the other issues, especially the `visible`. Thanks!

mbredif:
follow up fixing pr : #780

PR #766, opened 2018-05-25
feat(core): add a Parser registry

zarov:
Following #723 and the model of Providers, a `Parser` registry has been
partially added to the `Scheduler`. You can now register a `Parser` using
`scheduler.addFormatParser(formatName, parser)`.

In the `Scheduler` queue, the `Parser` will be called right after the
`Provider`. It takes the datablob returned by the `Provider`, parses it, and
return it.

The `Parser` selection relies on the format setted in the layer. For
layers that don't have a format, or have a format that is not supported,
a fake parser, returning immediatly the blob, is setted.

This commit introduces three parsers: `GeoJson`, `Gpx` and `Kml`. This allows
to refactor a bit the `RasterProvider`, in hope of letting go of it in the
future.

BREAKING CHANGE: `GpxParser` doesn't return a `THREE.Mesh` anymore.

Note: `RasterProvider` is kind of ugly for now, but I hope to get rid of it in
the near future. When "stylizers" will be added, the last real work of this
provider will be removed.


zarov:
I would like to see #754 #756 and #757 merged before merging this one, but that's not mandatory.

mbredif:
Thanks for taking this over :) !
After a quick look, 2 comments :
- This PR removes the GPX styling code. Does it break the gpx example ? If so, can Feature2Mesh achieve a similar result or should we move the GPX styling code into Feature2Mesh or to a new stylizer ?
- A suggestion : I find it odd to let the [provider decide on the fetchType](https://github.com/iTowns/itowns/pull/766/files#diff-6028308188afca277c93f77d477764c0R55). An alternative would be to let the fetchType be a property of the command, set (not setted ;) ) by the scheduler based on a property of the parser ([like here](https://github.com/iTowns/itowns/blob/9a9f01943229eacdc182d23dda572cb934c4537d/src/Parser/GeoJsonParser.js#L289)).

zarov:
> This PR removes the GPX styling code. Does it break the gpx example ? If so, can Feature2Mesh achieve a similar result or should we move the GPX styling code into Feature2Mesh or to a new stylizer ?

Yes it breaks it, I don't understand how I could have miss this one ! I'm working on putting it back together.

> A suggestion : I find it odd to let the provider decide on the fetchType

Yeah, I know, it is meant to be temporary and changed in the future. I missed this line and I'll try to add it here for parsers in this PR, thanks !

zarov:
@mbredif so I fixed back some of the things we had before, about the GPX. I was undecided where to add the customization of the track, and I felt it would suit more to only do it outside itowns, in the example. Two reasons: this customization is too specific to this usecase, and having it in the example shows people that you can still do 3D manipulation outside itowns.

I'm blocking on two aspects though, address to all in general:
- the drawn line is only working in system not using Angle (so not Windows), and joints are terrible. I was thinking either to implement our own solution about this in `feature2Mesh`, or use [`THREE.MeshLine`](https://github.com/spite/THREE.MeshLine) to draw all lines.
- I can't seem to succeed in adding the waypoint in the correct place, does someone know what I'm missing ?

zarov:
Closing this PR, as this is too old to still work on.

Issue #763, opened 2018-05-23
Improving the Feature type


mbredif:
The pivot [Feature type](https://github.com/iTowns/itowns/blob/master/src/Parser/GeoJsonParser.js#L233-L268), parsed by the GeoJSON parser will not scale well, with positions as an array of [Coordinates](https://github.com/iTowns/itowns/blob/master/src/Core/Geographic/Coordinates.js) objects. For instance, using that type as a pivot for binary pointclouds would mean deserializing typedarrays into Coordinates objects, just to immediately reserialize it into the typedarray of a BufferAttribute.

This issue is to propose and discuss what would be a better fit.

my proposal is to change the property `{Coordinates[]} vertices` to a `{THREE.BufferGeometry} geometry` with a mandatory 'position' attribute (which could be a [BufferAttribute](https://threejs.org/docs/#api/core/BufferAttribute) or a [InterleavedBufferAttribute](https://threejs.org/docs/#api/core/InterleavedBufferAttribute) and have an `crs` property).

autra:
> The pivot Feature type, parsed by the GeoJSON parser will not scale well

Yep indeed :-) We cannot keep a `Coordinates[]`. It's slow to create and slow to use.

The downside of a `BufferGeometry` is that you'll loose the SIG side of Coordinates. I don't know if we care about this for the Feature type though. Thoughts?

An alternative (if we do care about it) would be to support array (or Float32/64Array) of x,y,z triplets in `Coordinates`.

peppsac:
> my proposal is to change the property {Coordinates[]} vertices to a {THREE.BufferGeometry} geometry with a mandatory 'position' attribute (which could be a BufferAttribute or a InterleavedBufferAttribute and have an crs property).

Seems reasonnable, but I have 2 issues with this proposal:
- the `crs` issue raised by @autra in the previous comment
- the geometry -> `BufferGeometry` is "lossy". For instance:
  * linestring will be converted to polygons (if we use meshline, which is needed to support `linewidth` on windows platform)
  * polygons (and their holes) will converted to triangles

Basically this prevents implementing any transformations on the source data (remove a hole from a polygon, or add points to a line). It may not be a big deal though...

mbredif:
> An alternative (if we do care about it) would be to support array (or Float32/64Array) of x,y,z triplets in Coordinates.

[...]

PR #767, opened 2018-05-28
Add support for other data types in potree pointclouds

autra:
## Description

We were only supporting position and rgb informations from potree converted pointclouds. This PR aims at supporting all of them. At least I wanted to have them parsed in attributes, but I also added some default rendering to check that it is working correctly. 

Some remarks:
- I didn't test the oct16 normals, because I don't have any las with normal information or already converted pointclouds with oct16 format for normals. If some of you have one, I'd be very interested :-)
- what to do with classification remains to be decided. I could add some default rendering, but I'm not sure which type :-)

### DIFF
--- a/src/Parser/PotreeBinParser.js
+++ b/src/Parser/PotreeBinParser.js
@@ -48,6 +48,20 @@ const POINT_ATTTRIBUTES = {
     },
 };
 
+for (const potreeName of Object.keys(POINT_ATTTRIBUTES)) {
+    const attr = POINT_ATTTRIBUTES[potreeName];
+    attr.potreeName = potreeName;
+    attr.numByte = attr.numByte || attr.arrayType.BYTES_PER_ELEMENT;
+    attr.byteSize = attr.numElements * attr.numByte;
+    attr.normalized = attr.normalized || false;
+    // chrome is known to perform badly when we call a method without respecting its arity
+    // also, not using ternary because I measured a 25% perf hit on firefox doing so...
+    const fnName = `getUint${attr.numByte * 8}`;
+    attr.value = attr.numByte === 1 ?
+        function value(view, offset) { return view[fnName](offset); } :
+        function value(view, offset) { return view[fnName](offset, true); };
+}
+
 export default {
     /** @module PotreeBinParser */
     /** Parse .bin PotreeConverter format and convert to a THREE.BufferGeometry
@@ -66,50 +80,26 @@ export default {
         // Format: X1,Y1,Z1,R1,G1,B1,A1,[...],XN,YN,ZN,RN,GN,BN,AN
         let pointByteSize = 0;
         for (const potreeName of pointAttributes) {
-            const attr = POINT_ATTTRIBUTES[potreeName];
-            pointByteSize += attr.numElements * (attr.numByte || attr.arrayType.BYTES_PER_ELEMENT);
+            pointByteSize += POINT_ATTTRIBUTES[potreeName].byteSize;
         }
         const numPoints = Math.floor(buffer.byteLength / pointByteSize);
 
-        const attrs = [];
-        // get the variable attributes
+        const geometry = new THREE.BufferGeometry();
+        let elemOffset = 0;
+        let attrOffset = 0;
         for (const potreeName of pointAttributes) {
             const attr = POINT_ATTTRIBUTES[potreeName];
-            const numByte = attr.numByte || attr.arrayType.BYTES_PER_ELEMENT;
-            attrs.push({
-                potreeName,
-                numElements: attr.numElements,
-                attributeName: attr.attributeName,
-                normalized: attr.normalized,
-                array: new attr.arrayType(attr.numElements * numPoints),
-                numByte,
-                // Potree stores everything as int, and uses scale + offset
-                fnName: `getUint${numByte * 8}`,
-            });
-        }
-
-        let offset = 0;
-        for (let pntIdx = 0; pntIdx < numPoints; pntIdx++) {
-            for (const attr of attrs) {
+            const arrayLength = attr.numElements * numPoints;
+            const array = new attr.arrayType(arrayLength);
+            for (let arrayOffset = 0; arrayOffset < arrayLength; arrayOffset += attr.numElements) {
                 for (let elemIdx = 0; elemIdx < attr.numElements; elemIdx++) {
-                    // chrome is known to perform badly when we call a method without respecting its arity
-                    // also, not using ternary because I measured a 25% perf hit on firefox doing so...
-                    let value;
-                    if (attr.numByte === 1) {
-                        value = view[attr.fnName](offset);
-                    } else {
-                        value = view[attr.fnName](offset, true);
-                    }
-                    attr.array[pntIdx * attr.numElements + elemIdx] = value;
-
-                    offset += attr.numByte;
+                    array[arrayOffset + elemIdx] = attr.value(view, attrOffset + elemIdx * attr.numByte);
                 }
+                attrOffset += pointByteSize;
             }
-        }
-
-        const geometry = new THREE.BufferGeometry();
-        for (const attr of attrs) {
-            geometry.addAttribute(attr.attributeName, new THREE.BufferAttribute(attr.array, attr.numElements, attr.normalized));
+            elemOffset += attr.byteSize;
+            attrOffset = elemOffset;
+            geometry.addAttribute(attr.attributeName, new THREE.BufferAttribute(array, attr.numElements, attr.normalized));
         }
 
         geometry.computeBoundingBox();
--- a/src/Parser/PotreeBinParser.js
+++ b/src/Parser/PotreeBinParser.js
@@ -55,11 +55,10 @@ for (const potreeName of Object.keys(POINT_ATTTRIBUTES)) {
     attr.byteSize = attr.numElements * attr.numByte;
     attr.normalized = attr.normalized || false;
     // chrome is known to perform badly when we call a method without respecting its arity
-    // also, not using ternary because I measured a 25% perf hit on firefox doing so...
     const fnName = `getUint${attr.numByte * 8}`;
-    attr.value = attr.numByte === 1 ?
-        function value(view, offset) { return view[fnName](offset); } :
-        function value(view, offset) { return view[fnName](offset, true); };
+    attr.getValue = attr.numByte === 1 ?
+        function getValue(view, offset) { return view[fnName](offset); } :
+        function getValue(view, offset) { return view[fnName](offset, true); };
 }
 
 export default {
@@ -93,7 +92,7 @@ export default {
             const array = new attr.arrayType(arrayLength);
             for (let arrayOffset = 0; arrayOffset < arrayLength; arrayOffset += attr.numElements) {
                 for (let elemIdx = 0; elemIdx < attr.numElements; elemIdx++) {
-                    array[arrayOffset + elemIdx] = attr.value(view, attrOffset + elemIdx * attr.numByte);
+                    array[arrayOffset + elemIdx] = attr.getValue(view, attrOffset + elemIdx * attr.numByte);
                 }
                 attrOffset += pointByteSize;
             }
--- a/src/Parser/PotreeBinParser.js
+++ b/src/Parser/PotreeBinParser.js
@@ -1,52 +1,118 @@
 import * as THREE from 'three';
 
+// See the different constants holding ordinal, name, numElements, byteSize in PointAttributes.cpp in PotreeConverter
+// elementByteSize is byteSize / numElements
+const POINT_ATTTRIBUTES = {
+    POSITION_CARTESIAN: {
+        numElements: 3,
+        arrayType: Float32Array,
+        attributeName: 'position',
+    },
+    COLOR_PACKED: {
+        numElements: 4,
+        arrayType: Uint8Array,
+        attributeName: 'color',
+        normalized: true,
+    },
+    INTENSITY: {
+        numElements: 1,
+        numByte: 2,
+        // using Float32Array because Float16Array doesn't exist
+        arrayType: Float32Array,
+        attributeName: 'intensity',
+        normalized: true,
+    },
+    CLASSIFICATION: {
+        numElements: 1,
+        arrayType: Uint8Array,
+        attributeName: 'classification',
+    },
+    // Note: at the time of writing, PotreeConverter will only generate normals in Oct16 format
+    // see PotreeConverter.cpp:121
+    // we keep all the historical value to still supports old conversion
+    NORMAL_SPHEREMAPPED: {
+        numElements: 2,
+        arrayType: Uint8Array,
+        attributeName: 'sphereMappedNormal',
+    },
+    // see https://web.archive.org/web/20150303053317/http://lgdv.cs.fau.de/get/1602
+    NORMAL_OCT16: {
+        numElements: 2,
+        arrayType: Uint8Array,
+        attributeName: 'oct16Normal',
+    },
+    NORMAL: {
+        numElements: 3,
+        arrayType: Float32Array,
+        attributeName: 'normal',
+    },
+};
+
 export default {
     /** @module PotreeBinParser */
     /** Parse .bin PotreeConverter format and convert to a THREE.BufferGeometry
      * @function parse
      * @param {ArrayBuffer} buffer - the bin buffer.
+     * @param {Object} pointAttributes - the point attributes information contained in layer.metadata coming from cloud.js
      * @return {Promise} - a promise that resolves with a THREE.BufferGeometry.
      *
      */
-    parse: function parse(buffer) {
+    parse: function parse(buffer, pointAttributes) {
         if (!buffer) {
             throw new Error('No array buffer provided.');
         }
 
         const view = new DataView(buffer);
         // Format: X1,Y1,Z1,R1,G1,B1,A1,[...],XN,YN,ZN,RN,GN,BN,AN
-        const numPoints = Math.floor(buffer.byteLength / 16);
+        let pointByteSize = 0;
+        for (const potreeName of pointAttributes) {
+            const attr = POINT_ATTTRIBUTES[potreeName];
+            pointByteSize += attr.numElements * (attr.numByte || attr.arrayType.BYTES_PER_ELEMENT);
+        }
+        const numPoints = Math.floor(buffer.byteLength / pointByteSize);
 
-        const positions = new Float32Array(3 * numPoints);
-        const colors = new Uint8Array(4 * numPoints);
-
-        const box = new THREE.Box3();
-        box.min.set(Infinity, Infinity, Infinity);
-        box.max.set(-Infinity, -Infinity, -Infinity);
-        const tmp = new THREE.Vector3();
+        const attrs = [];
+        // get the variable attributes
+        for (const potreeName of pointAttributes) {
+            const attr = POINT_ATTTRIBUTES[potreeName];
+            const numByte = attr.numByte || attr.arrayType.BYTES_PER_ELEMENT;
+            attrs.push({
+                potreeName,
+                numElements: attr.numElements,
+                attributeName: attr.attributeName,
+                normalized: attr.normalized,
+                array: new attr.arrayType(attr.numElements * numPoints),
+                numByte,
+                // Potree stores everything as int, and uses scale + offset
+                fnName: `getUint${numByte * 8}`,
+            });
+        }
 
         let offset = 0;
-        for (let i = 0; i < numPoints; i++) {
-            positions[3 * i] = view.getUint32(offset + 0, true);
-            positions[3 * i + 1] = view.getUint32(offset + 4, true);
-            positions[3 * i + 2] = view.getUint32(offset + 8, true);
+        for (let pntIdx = 0; pntIdx < numPoints; pntIdx++) {
+            for (const attr of attrs) {
+                for (let elemIdx = 0; elemIdx < attr.numElements; elemIdx++) {
+                    // chrome is known to perform badly when we call a method without respecting its arity
+                    // also, not using ternary because I measured a 25% perf hit on firefox doing so...
+                    let value;
+                    if (attr.numByte === 1) {
+                        value = view[attr.fnName](offset);
+                    } else {
+                        value = view[attr.fnName](offset, true);
+                    }
+                    attr.array[pntIdx * attr.numElements + elemIdx] = value;
 
-            tmp.fromArray(positions, 3 * i);
-            box.min.min(tmp);
-            box.max.max(tmp);
-
-            colors[4 * i] = view.getUint8(offset + 12);
-            colors[4 * i + 1] = view.getUint8(offset + 13);
-            colors[4 * i + 2] = view.getUint8(offset + 14);
-            colors[4 * i + 3] = 255;
-
-            offset += 16;
+                    offset += attr.numByte;
+                }
+            }
         }
 
         const geometry = new THREE.BufferGeometry();
-        geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
-        geometry.addAttribute('color', new THREE.BufferAttribute(colors, 4, true));
-        geometry.boundingBox = box;
+        for (const attr of attrs) {
+            geometry.addAttribute(attr.attributeName, new THREE.BufferAttribute(attr.array, attr.numElements, attr.normalized));
+        }
+
+        geometry.computeBoundingBox();
 
         return Promise.resolve(geometry);
     },
--- a/src/Process/PointCloudProcessing.js
+++ b/src/Process/PointCloudProcessing.js
@@ -104,7 +104,6 @@ export default {
             layer.material.opacity = layer.opacity;
             layer.material.transparent = layer.opacity < 1;
             layer.material.size = layer.pointSize;
-            layer.material.mode = layer.mode;
         }
 
         // lookup lowest common ancestor of changeSources
--- a/src/Process/PointCloudProcessing.js
+++ b/src/Process/PointCloudProcessing.js
@@ -104,6 +104,7 @@ export default {
             layer.material.opacity = layer.opacity;
             layer.material.transparent = layer.opacity < 1;
             layer.material.size = layer.pointSize;
+            layer.material.mode = layer.mode;
         }
 
         // lookup lowest common ancestor of changeSources
--- a/src/Provider/PointCloudProvider.js
+++ b/src/Provider/PointCloudProvider.js
@@ -3,7 +3,7 @@ import Fetcher from './Fetcher';
 import PointCloudProcessing from '../Process/PointCloudProcessing';
 import PotreeBinParser from '../Parser/PotreeBinParser';
 import PotreeCinParser from '../Parser/PotreeCinParser';
-import PointsMaterial from '../Renderer/PointsMaterial';
+import PointsMaterial, { MODE } from '../Renderer/PointsMaterial';
 import Picking from '../Core/Picking';
 import Extent from '../Core/Geographic/Extent';
 
@@ -169,6 +169,7 @@ export default {
         layer.type = 'geometry';
         layer.material = layer.material || {};
         layer.material = layer.material.isMaterial ? layer.material : new PointsMaterial(layer.material);
+        layer.mode = MODE.COLOR;
 
         // default update methods
         layer.preUpdate = PointCloudProcessing.preUpdate;
@@ -194,6 +195,13 @@ export default {
                 bbox = new THREE.Box3(
                     new THREE.Vector3(cloud.boundingBox.lx, cloud.boundingBox.ly, cloud.boundingBox.lz),
                     new THREE.Vector3(cloud.boundingBox.ux, cloud.boundingBox.uy, cloud.boundingBox.uz));
+
+                // do we have normal information
+                const normal = Array.isArray(layer.metadata.pointAttributes) &&
+                    layer.metadata.pointAttributes.find(elem => elem.startsWith('NORMAL'));
+                if (normal) {
+                    layer.material.defines[normal] = 1;
+                }
             } else {
                 // Lopocs
                 layer.metadata.scale = 1;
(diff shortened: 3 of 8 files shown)


src/Parser/PotreeBinParser.js
```diff
@@ -1,53 +1,122 @@
...
+        normalized: true,
+    },
+    CLASSIFICATION: {
+        numElements: 1,
+        arrayType: Uint8Array,
+        attributeName: 'classification',
+    },
+    // Note: at the time of writing, PotreeConverter will only generate normals in Oct16 format
+    // see PotreeConverter.cpp:121
+    // keeping all the values to stay iso to PointAttributes.cpp
```
autra: Actually keeping all the values to be able to read clouds converted with old version. I'm editing that.

src/Parser/PotreeBinParser.js
```diff
@@ -1,53 +1,122 @@
...
 
-            offset += 16;
+                // calculate tightbox
+                if (attr.potreeName === 'POSITION_CARTESIAN') {
+                    tmp.set(attr.array[pntIdx * attr.numElements + 0],
+                            attr.array[pntIdx * attr.numElements + 1],
+                            attr.array[pntIdx * attr.numElements + 2]);
+                    tightbbox.min.min(tmp);
+                    tightbbox.max.max(tmp);
+                }
```
mbredif: would not it be simpler to use `geometry.computeBoundingBox()` ?

src/Process/PointCloudProcessing.js
```diff
@@ -165,6 +165,7 @@ export default {
             if (elt.obj) {
                 elt.obj.material.visible = true;
                 elt.obj.material.size = layer.pointSize;
+                elt.obj.material.uniforms.mode.value = layer.mode;
```
mbredif: that would break using a material with no `mode` uniform.
you could instead set here  `elt.obj.material.mode = layer.mode;` and update the uniform inside the `updateUniforms` function.

mbredif (commented):
I left some comments. Do you get better performance than using an interleavedarraybuffer (assuming it is possible) ?

mbredif:
> what to do with classification remains to be decided. I could add some default rendering, but I'm not sure which type :-)

two ideas:
- simplest thing : you can use an overlaycolor indexed by the classification Id.
- much more interesting and involved approach : [this paper :)](https://www.researchgate.net/publication/269809625_Out-of-Core_Visualization_of_Classified_3D_Point_Clouds)

src/Process/PointCloudProcessing.js
autra: Good idea!

src/Parser/PotreeBinParser.js
autra: Indeed. I've checked the implementation, and it is essentially the same as here. It'll add a for loop, but I don't expect its cost to be perceivable.

autra:
> Do you get better performance than using an interleavedarraybuffer (assuming it is possible) ?

I just didn't think about using it :-) Thanks for the idea, I'm gonna try.

>> what to do with classification remains to be decided. I could add some default rendering, but I'm not sure which type :-)
> 
> two ideas:
> 
> - simplest thing : you can use an overlaycolor indexed by the classification Id.
> - much more interesting and involved approach : this paper :)

Interesting paper! 

For a basic render for testing purpose, I'd prefer to have something simpler (this paper technique necessitates a special pre-processing (so not all converted pointclouds can be displayed with it) and multipass rendering). I was thinking of doing exactly what your first proposal is, but couldn't decide if this was worth it here or not. After all, once you have the classification in the shader, applying some color is very easy and we could let app developers decide how they want to render it, because they'll know their classification, the number of them etc.? As I was not sure, I thought it better to leave it out for now and add it later once our mind is made, WDYT?

Btw, I didn't upload a screenshot of normal rendering!

point cloud viewer - firefox developer edition_113

This pointcloud is lion_takanawa present in the potree repository. I'll look into the licence to see if we can add it to itowns-sample-data (and get the las, that'll be great to test oct16 normal rendering).

autra:
@mbredif concerning InterleavedBuffer, the way they are implemented in THREE.js seems to lack an important feature: you cannot specify the stride in bytes. You can specify the itemSize and the offset in InterleavedBufferAttribute, and you can specify the stride in terms of number of typed-array elements. 

This works fine if one point's bytesize is a multiple of your data type for one attribute, but it's not always the case. For example, if my datas are positions (32 bits) and classification (8 bits), I can specify a stride of 5 for the classification buffer, but there's no way I can specify the right stride for the positions buffer.

Do you reach the same conclusion as I did? Or did I misunderstand something? If you do I'll ask a question/open an issue in the THREE.js repository.

src/Process/PointCloudProcessing.js
autra: Actually, what about passing `layer` to `updateUniforms`? That's more or less the pattern we used in iTowns: update uniforms according to some config values attached to layer. It'll make it explicit, and would avoid yet another value copy (they are getting difficult to follow).

autra:
updated (@iTowns/reviewers and @mbredif )

autra:
I also pushed a proposal about the `updateUniforms` mechanism.

src/Parser/PotreeBinParser.js
```diff
@@ -1,61 +1,117 @@
...
+    },
+    COLOR_PACKED: {
+        numElements: 4,
+        arrayType: Uint8Array,
+        attributeName: 'color',
+        normalized: true,
+    },
+    INTENSITY: {
+        numElements: 1,
+        numByte: 2,
```
peppsac, line 19: Maybe you should add a comment to explain why you need this? (AFAIU it's because `Float16Array` doesn't exist so you emulate it)

src/Parser/PotreeBinParser.js
```diff
@@ -1,61 +1,117 @@
...
 
         let offset = 0;
-        for (let i = 0; i < numPoints; i++) {
-            positions[3 * i] = view.getUint32(offset + 0, true);
-            positions[3 * i + 1] = view.getUint32(offset + 4, true);
-            positions[3 * i + 2] = view.getUint32(offset + 8, true);
+        for (let pntIdx = 0; pntIdx < numPoints; pntIdx++) {
+            for (const attr of attrs) {
+                for (let elemIdx = 0; elemIdx < attr.numElements; elemIdx++) {
+                    const fnName = `getUint${(attr.numByte || attr.array.BYTES_PER_ELEMENT) * 8}`;
```
peppsac: This could be done in the parent loop since it doesn't depend on `elemIdx`.

src/Renderer/PointsMaterial.js
```diff
@@ -26,19 +35,24 @@ class PointsMaterial extends RawShaderMaterial {
...
-
-        this.updateUniforms();
     }
 
     enablePicking(pickingMode) {
         // we don't want pixels to blend over already drawn pixels
         this.blending = pickingMode ? NoBlending : NormalBlending;
-        this.uniforms.pickingMode.value = pickingMode;
+        if (pickingMode) {
+            this.oldMode = this.uniforms.mode.value;
```
peppsac: Shouldn't `this.oldMode` be also assigned in the constructor (like @elemoine mentionned here https://github.com/iTowns/itowns/pull/719#issuecomment-380376078)?

peppsac (commented):
Just left a couple comments, looks good overall.

src/Renderer/PointsMaterial.js
autra: Yes, at least initialized to an empty value, if we want to rely entirely on the update mechanism.

autra:
@peppsac updated and rebased.

src/Parser/PotreeBinParser.js
```diff
@@ -1,61 +1,118 @@
...
+        // get the variable attributes
+        for (const potreeName of pointAttributes) {
+            const attr = POINT_ATTTRIBUTES[potreeName];
+            attrs.push({
+                potreeName,
+                numElements: attr.numElements,
+                attributeName: attr.attributeName,
+                normalized: attr.normalized,
+                array: new attr.arrayType(attr.numElements * numPoints),
+                numByte: attr.numByte,
```
mbredif: you could apply the default here and be done with it

src/Parser/PotreeBinParser.js
```diff
@@ -1,61 +1,118 @@
...
+        }
 
         let offset = 0;
-        for (let i = 0; i < numPoints; i++) {
-            positions[3 * i] = view.getUint32(offset + 0, true);
-            positions[3 * i + 1] = view.getUint32(offset + 4, true);
-            positions[3 * i + 2] = view.getUint32(offset + 8, true);
+        for (let pntIdx = 0; pntIdx < numPoints; pntIdx++) {
+            for (const attr of attrs) {
+                const fnName = `getUint${(attr.numByte || attr.array.BYTES_PER_ELEMENT) * 8}`;
```
mbredif: this could move out of the `pntIdx` loop, by setting it as a function of `attr` in its definition loop.

Maybe a comment on why getUint* works even on all types (float, Int...) (It looks strange to me)

src/Provider/PointCloudProvider.js
```diff
@@ -138,7 +139,7 @@ function loadPointFile(layer, url) {
         if (layer.metadata.customBinFormat) {
             return PotreeCinParser.parse(ab).then(result => addPickingAttribute(result));
         } else {
-            return PotreeBinParser.parse(ab).then(result => addPickingAttribute(result));
+            return PotreeBinParser.parse(ab, layer.metadata.pointAttributes).then(result => addPickingAttribute(result));
```
mbredif: maybe the whole `layer.metadata` should be given for future parsers to consume it.

src/Renderer/PointsMaterial.js
mbredif: this code looks buggy : if you enable picking twice in a row, oldmode will be PICKING. A more sensible invariant could be "oldmode holds the last non-picking mode used" and could be initialized by the initial mode (or whatever if the initial mode is picking)

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -3,24 +3,73 @@ precision highp int;
...
 #include <logdepthbuf_pars_vertex>
 #define EPSILON 1e-6
+// see PointsMaterial.js
+#define MODE_COLOR 0
+#define MODE_PICKING 1
+#define MODE_INTENSITY 2
+#define MODE_CLASSIFICATION 3
+#define MODE_NORMAL 4
+#define MODE_NORMAL_OCT16 5
+#define MODE_NORMAL_SPHEREMAPPED 6
```
mbredif: to prevent the replication of code values, material.defines could be used instead

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -3,24 +3,73 @@ precision highp int;
...
+    if (mode == MODE_PICKING) {
         vColor = unique_id;
+    } else if (mode == MODE_INTENSITY) {
+        vColor = vec4(intensity, intensity, intensity, 1.0);
+    } else if (mode == MODE_NORMAL) {
+        vColor = normal;
+    } else if (mode == MODE_NORMAL_OCT16) {
+        vColor = vec4(decodeOct16Normal(oct16Normal), 1.0);
+    } else if (mode == MODE_NORMAL_SPHEREMAPPED) {
+        vColor = vec4(decodeSphereMappedNormal(sphereMappedNormal), 1.0);
```
mbredif: normals range from -1 to 1. maybe you can take  the absolute value of it.

mbredif (commented):
That works great already ! here are some remarks.

Concerning interleaved buffer attributes, there seems to be issues in webgl when attributes are not type-aligned, hence the restriction. That means that creating an interleaved buffer with no attribute decoding may not always be possible. a possibility would be to have a hint option that tentatively tries it and defaults to non interleaved buffers if decoding is required.

src/Parser/PotreeBinParser.js
autra: > this could move out of the pntIdx loop, by setting it as a function of attr in its definition loop.

What do we gain by that?

src/Parser/PotreeBinParser.js
autra: I don't understand what you mean by that.

src/Parser/PotreeBinParser.js
autra: > Maybe a comment on why getUint* works even on all types (float, Int...) (It looks strange to me)

it doesn't work on all types, it's just that potree converter stores everything as int, then putting an int into a Float32Array will convert it into a float.

src/Parser/PotreeBinParser.js
mbredif: here : ` numByte: attr.numByte || attr.array.BYTES_PER_ELEMENT, `
elsewhere: `attr.numByte` instead of `attr.numByte || attr.array.BYTES_PER_ELEMENT`

src/Parser/PotreeBinParser.js
mbredif: > What do we gain by that?

`numPoints` times less evaluations of ```view[`getUint${(attr.numByte || attr.array.BYTES_PER_ELEMENT) * 8}`]``` (and cleaner code IMHO :))

autra: > by setting it as a function of attr in its definition loop.

you mean as a property of attr then?

mbredif: > it doesn't work on all types, it's just that potree converter stores everything as int, then putting an int into a Float32Array will convert it into a float.

Indeed, by looking into it, potree is relying on a [custom view object](https://github.com/potree/potree/blob/develop/src/workers/BinaryDecoderWorker.js) rather than a DataView. I am not sure the 2 approaches are strictly equivalent (It is not clear to me who is manipulating bytes and who is doing type casting), but if your approach works, then let us go with it !

mbredif: > you mean as a property of attr then?

yes

src/Provider/PointCloudProvider.js
autra: I'm not sure yet, that's why I did an adhoc call here. I think this question should be tackled when/if we have a parser registry mechanism for pointclouds, WDYT? Not yet sure how this will be done though (there are probably things that can be unified with 3dTiles pnts format I guess).

autra:
@iTowns/reviewers updated.

Also pinging @mbredif: I think I have adressed all your remarks.

src/Renderer/PointsMaterial.js
```diff
@@ -12,17 +20,24 @@ class PointsMaterial extends RawShaderMaterial {
...
+
+        for (const key in MODE) {
+            if (Object.prototype.hasOwnProperty.call(MODE, key)) {
+                this.defines[`MODE_${key}`] = MODE[key];
+            }
+        }
 
         this.uniforms.size = new Uniform(this.size);
+        this.uniforms.mode = new Uniform(this.mode);
         this.uniforms.pickingMode = new Uniform(false);
```
mbredif: unused, to be removed

src/Provider/PointCloudProvider.js
```diff
@@ -193,6 +194,13 @@ export default {
                 bbox = new THREE.Box3(
                     new THREE.Vector3(cloud.boundingBox.lx, cloud.boundingBox.ly, cloud.boundingBox.lz),
                     new THREE.Vector3(cloud.boundingBox.ux, cloud.boundingBox.uy, cloud.boundingBox.uz));
+
+                // do we have normal information
+                const normal = Array.isArray(layer.metadata.pointAttributes) &&
+                    layer.metadata.pointAttributes.find(elem => elem.startsWith('NORMAL'));
+                if (normal) {
+                    layer.material.defines[normal] = 1;
+                }
```
mbredif: what about an opt-in `isNormal` property for pointAttributes ? that would prevent having a dependency on this specific material...

src/Parser/PotreeBinParser.js
```diff
@@ -1,52 +1,118 @@
...
+                potreeName,
+                numElements: attr.numElements,
+                attributeName: attr.attributeName,
+                normalized: attr.normalized,
+                array: new attr.arrayType(attr.numElements * numPoints),
+                numByte,
+                // Potree stores everything as int, and uses scale + offset
+                fnName: `getUint${numByte * 8}`,
+            });
+        }
```
mbredif: Apart from `array`, this could be done statically once and for all in the global scope

src/Parser/PotreeBinParser.js
```diff
@@ -1,52 +1,118 @@
...
 
         let offset = 0;
-        for (let i = 0; i < numPoints; i++) {
-            positions[3 * i] = view.getUint32(offset + 0, true);
-            positions[3 * i + 1] = view.getUint32(offset + 4, true);
-            positions[3 * i + 2] = view.getUint32(offset + 8, true);
+        for (let pntIdx = 0; pntIdx < numPoints; pntIdx++) {
+            for (const attr of attrs) {
+                for (let elemIdx = 0; elemIdx < attr.numElements; elemIdx++) {
```
mbredif: if the previous attr loop goes in global scope, the array could be instanced here (just after `for (const attr of attrs)`) and added to the geometry at the end of the iteration instead of in another loop.

mbredif (commented):
Apart from the inline comments, I have 2 concerns :

1. how well does it work with the stock `THREE.PointsMaterial` ? you can browse here : `examples/pointcloud.html?material=three`. Ensuring it works well would prevent too much coupling with `itowns.PointsMaterial`...

2. I find the `oldMode` logic a bit convoluted.
Maybe you could have `this.picking=false` and `this.mode=MODE.COLOR` in the constructor, remove the `oldMode` property and the `enablePicking` function (just set `material.picking` instead), and do this in `updateUniforms` : 

```
this.blending = pickingMode ? NoBlending : NormalBlending;
 this.uniforms.mode.value = this.picking ? MODE.PICKING : this.mode;
```

src/Parser/PotreeBinParser.js
autra: I think I don't see what you mean, because I feel I can't. I'm constructing an array of attr that depends on pointAttributes. The point of this code is to copy the `POINT_ATTTRIBUTES` static array.

src/Parser/PotreeBinParser.js
autra: Put some code please, that will be faster and clearer.

src/Provider/PointCloudProvider.js
autra: The more general question is: do we expect users to be able subclass `Material` and not `ShaderMaterial`?

src/Renderer/PointsMaterial.js
autra: Indeed, thanks for the catch :-)

autra:
> how well does it work with the stock THREE.PointsMaterial ? you can browse here : examples/pointcloud.html?material=three. Ensuring it works well would prevent too much coupling with itowns.PointsMaterial...

To understand a bit more what you're after, can you tell me a bit more about your use case for other Materials?

Otherwise, please note that the debug tools won't completely work with `material=three`. These debug tools are all pretty much material dependant.

>  I find the oldMode logic a bit convoluted. Maybe you could have this.picking=false and this.mode=MODE.COLOR in the constructor, remove the oldMode property and the enablePicking function (just set material.picking instead), and do this in updateUniforms :

It's a bit convoluted, but for the sake of being clearer for the user: picking is mutually exclusive with other display modes. Having a separate boolean to set it will complicate things from a user POV.

mbredif:
- For the `attr` loop (https://github.com/iTowns/itowns/pull/767#discussion_r195420612 and https://github.com/iTowns/itowns/pull/767#discussion_r195419937), see https://github.com/mbredif/itowns/commit/fe850acba60e28065eb9c5c3d8dedcf49caa4349

> The more general question is: do we expect users to be able subclass Material and not ShaderMaterial?
To understand a bit more what you're after, can you tell me a bit more about your use case for other Materials?

Ideally, I would rather not have any coupling between individual parsers, providers and materials, meaning (the includes in the potree provider would just be there to provide default parsers and materials, that could be overriden) : 
- we should be able to use any material, including the ones from THREE (eg THREE.PointsMaterial) or other THREE-based libraries, for most itowns objects
- extra features (such as modes) should be opt-in features of the materials that need specification
- parsed attributes should have semantics (isNormal, isPosition, crs...) to decouple parsers from materials : ideally parsed pointclouds would just be `Feature`s (with improvements discussed in #763).
- the same itowns.PointsMaterial could be used on any pointcloud (static PLY files, Potree layers and 3Dtiles/PNTS layers...) with its improved features, but THREE.PointsMaterial would work as well.

> Otherwise, please note that the debug tools won't completely work with material=three. These debug tools are all pretty much material dependant.

... unless we specify the opt-in `mode` feature for itowns materials. (there is already a similar concept, with a completely different implementation for the LayeredMaterial, that is weirdly managed by the [TileMesh](https://github.com/iTowns/itowns/blob/4bdb62ac705e2a7bf334fb423eb1f3de1a739911/src/Core/TileMesh.js#L76-L94) )

> It's a bit convoluted, but for the sake of being clearer for the user: picking is mutually exclusive with other display modes. Having a separate boolean to set it will complicate things from a user POV.

I agree from a shader perspective that these are exclusive modes, but from a user perspective (and that is how we use it!), picking is more like an overriding mode which deserves (IMHO) an overriding property. For symmetry, I suggest to go from the boolean `this.picking` to a `this.overrideMode` property (much like THREE's `overrideMaterial`)

mbredif:
Note that, even I had the rights, I would not block this PR, I am just expressing my opinion on the coupling it reintroduces... However that can be addressed in a follow up PR (eg: for the equivalent PR on PNTS pointclouds)

peppsac:
> picking is more like an overriding mode which deserves (IMHO) an overriding property. For symmetry, I suggest to go from the boolean this.picking to a this.overrideMode property

FWIW, I agree with both of you:
- picking is an overriding mode so an `this.overrideMode` would make sense
- but the only usage we have (and that I can see) of `this.overrideMode` is for picking, so it makes sense to use a plain boolean instead

mbredif:
Now that I have reviewed most of the itowns codebase, I suggest to : 
- use the RenderMode enum in the core for modes of all material (possibly materials could add new render modes as well)
- let materials have an opt-in  `material.mode` property to select the mode (which could be implemented with defines, uniforms or a combination at the discretion of each material using `Object.defineProperty` and setting `needsUpdate` or `uniformsNeedUpdate` as required). Materials with no `mode` property are simply not selectable. (maybe mode support could be advertised with a function)
- it is the caller's responsibility (picking, styling, depth estimation...) to backup previous state and restore it (something like pushRenderMode that would just set or restore the mode property or maybe just retraversing the tree)

autra:
So I kept your first commit @mbredif  (thanks for the simplification). As for the second, I suggest to open a PR with it (I'd be glad to r+ it). I have pushed it in its own branch (https://github.com/iTowns/itowns/tree/points_shader_refacto).

I reintroduced a picking boolean as you and @peppsac suggested, and restored the compatibility with THREE material (I even adapted PointCloudDebug to support it by making picking and mode opt-in).

@iTowns/reviewers please r? :-)

src/Provider/PointCloudProvider.js
peppsac: > that would prevent having a dependency on this specific material

IMHO, we don't need to impose ourselves  too much constraints. In the same way that TileMesh/LayeredMaterial is not replaceable by another mesh/material, we should totally rely on PointsMaterial to implement our features.

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -10,17 +10,71 @@ uniform mat4 modelViewMatrix;
...
+}
+#endif
+
 void main() {
     if (pickingMode) {
         vColor = unique_id;
+    } else if (mode == MODE_INTENSITY) {
+        vColor = vec4(intensity, intensity, intensity, opacity);
+    } else if (mode == MODE_NORMAL) {
+#if defined(NORMAL_OCT16)
```
peppsac: I'd like to separate the `normal` parsing from colorisation mode.
Something like:
```glsl
#if defined(NORMAL_OCT16)
  vec3  normal = decodeOct16Normal(...);
#elif defined(NORMAL_SPHEREMAPPED)
   vec3 normal = decodeSphereMappedNormal(sphereMappedNormal);
#elif defined(NORMAL)
  // nothing to do
#else
   vec3 normal;
#endif


if (pickingMode) {
    vColor = unique_id;
    vColor = unique_id;
} else if (mode == MODE_INTENSITY) {
    vColor = vec4(intensity, intensity, intensity, opacity);
} else if (mode == MODE_NORMAL) {
    vColor = vec4(abs(normal), opacity);
} else {
...
}
```

test/PotreeBinParser_unit_test.js
```diff
@@ -0,0 +1,72 @@
+/* global describe, it */
+import PotreeBinParser from '../src/Parser/PotreeBinParser';
+
+const assert = require('assert');
+
+
+describe('PotreeBinParser', function () {
```
peppsac, line 7: :+1:

peppsac (approved):
Looks good (I left a suggestion for the shader code)

src/Provider/PointCloudProvider.js
mbredif: I disagree, the less coupling, the more itowns is untangled, usable and extensible... I would rather set opt-in hint properties to materials so that materials could honor the hint if possible.

src/Renderer/Shader/PointsVS.glsl
mbredif: I agree, I wrote a commit about it with shader chunks : 58c6c21e3b17f78afe680db80b9ad5fb65207779

src/Provider/PointCloudProvider.js
autra: On this instance, we don't really care. The only "coupling" was the fact layer.material.defines was undefined (!). The current code works with a basic three material.

gchoqueux (dismissed):
we should add tests on :
PointCloudProvider.js
PointsMaterial.js

peppsac:
> gchoqueux requested changes
> we should add tests on :
> PointCloudProvider.js
> PointsMaterial.js

:tired_face: 

@gchoqueux you should take a step back and think if you're attitude is really helping the itowns project (hint: it is not).

autra:
@gchoqueux what test I can add for these 2 modules is unclear to me. Can you push a test example in a branch to see what you have in mind?

src/Provider/PointCloudProvider.js
mbredif: I completely agree here. (We'll talk again about it when we will want to have the same functionality with the same material applied to pointclouds parsed from PNTS, PLY, LAS, LAZ...)

autra:
@gchoqueux I pushed a test for `PointsCloudProvider.js`. On the other hand, I don't think `PointsMaterial` really needs any testing.

autra:
@iTowns/reviewers please r? again :-)

src/Provider/PointCloudProvider.js
```diff
@@ -272,3 +274,7 @@ export default {
         });
     },
 };
+
+export const _testing = {
+    parseMetadata,
```
zarov, line 287: I'm not okay with this, please move it to the default export, even if I understand that this is for testing purpose only.

zarov (commented):
Seems good to me, I don't see either which test could be added to `PointsMaterial`.

src/Provider/PointCloudProvider.js
autra: Thanks for your review. Can you elaborate on the reasons you don't like this?

autra:
Btw, I'm not quite happy with the name and the return value of `parseMetadata` (they don't really match). Do you guys have a better suggestion?

src/Provider/PointCloudProvider.js
autra: To elaborate on my position, I'm pretty against exposing this to the default export, because it's a private method that has no reason to be exposed. Once you expose such a method, it becomes part of the public api, which means we have an implicit commitment to it. I know that this way of thinking is not very common on this project, but I'd like not to introduce api breakage when we can avoid it. For me `_testing` is a good middle-ground as it conveys the fact it is a private method.

zarov: I don't really know, maybe in fact it is a good idea. We could agree on a special exported namespace we could use to test internal methods. So in fact maybe keep it and find a more suitable name ? `_testing` feels muddled to me.

zarov: > Once you expose such a method, it becomes part of the public api

Providers are not in the public API so no problems here, but see my change of pov above ;)

autra: > Providers are not in the public API so no problems here

Just a quick note on this: I'm afraid they *are*. Everything importable from a ES6 configuration is part of the public API and we should treat it that way. Every transpiled file in `lib` is importable. Actually, in my opinion this should be the preferred way of using iTowns. The 2 `Main` files are just convenience.

> So in fact maybe keep it and find a more suitable name ? _testing feels muddled to me.

I don't know, `testing` seemed a suitable name for an export for testing purpose... I don't see a better name (yet).

mbredif:
as parseMetadata and parseOctree are parse commands, should not they be part of a 'parsers/PotreeCloudParser.js'  file (or something similar) ?

autra:
> as parseMetadata and parseOctree are parse commands, should not they be part of a 'parsers/PotreeCloudParser.js' file (or something similar) ?

Not sure. We use `parse` here, but it's not really the case, we are just reading protocol-specific metadata (so their place is in a Provider). Maybe they should be renamed though. I'm not happy with this `parseMetadata` name anyway.

src/Provider/PointCloudProvider.js
autra: Ok I see one possible solution: surround testing exports with `__DEBUG__` tests. But I need to add support for this in our `.babelrc`. So as not to enter into a rabbit hole for this PR, I suggest to keep it that way here and revisit it later. Is it acceptable for you @zarov?

autra: (and then rename `__DEBUG__` with `__DEV__`)

zarov: (sorry for being so undecided) yeah `testing` is good :D But if you feel that surrounding testing with those is better, let's go for it.

autra:
Ok, I pushed another version with `parseMetadata` renamed into `setLayerMetadata`. @peppsac @zarov WDYT?

autra:
actually, `setXXX` should not be that complex (should be a one liner), reverting to `parseXXXX`. I'm sure that this won't threaten the future of iTowns to have a `parseXXX` (private) method in a Provider ;-)

PR #769, opened 2018-05-30
refactor(core): change view.notifyChange semantics

peppsac:
This commit changes the semantics of view.notifyChange function to something more
useful: `changeSource` now drives layer update.

This implies that no `changeSource` -> no update.

`changeSource` can be either a layer, an object from a layer or the camera.

The signature also changed slightly:
```js
   notifyChange(needsRedraw = false, changeSource = undefined)
```
becomes:
```js
   notifyChange(changeSource = undefined, needsRedraw = true)
```

Some examples usage:
  - `notifyChange()`: the scene will be redraw, but no update will be performed. This
is useful if a user adds some objetcs to the scene and wants to request a redraw but
do no need a full update of all layer (eg: a measure tool drawn with threejs lines)
  - `notifyChange(XXX)`: only the layer XXX (or the layer to which XXX belongs) will
be updated. So if you have 100 layers, and layer #4 has a new texture, it will not
cause a full update of other layers.
  - `notifyChange(camera)`: this is equivalent to the previous notifyChange(true): all
layers will be updated.

MainLoop also handles `changeSource` filtering and could later even handle ancestor
lookup (if we build a common ancestor lookup interface).

### DIFF
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -1,4 +1,5 @@
 import { EventDispatcher } from 'three';
+import { GeometryLayer, Layer } from './Layer/Layer';
 
 export const RENDERING_PAUSED = 0;
 export const RENDERING_SCHEDULED = 1;
@@ -73,6 +74,19 @@ function updateElements(context, geometryLayer, elements) {
     }
 }
 
+function filterChangeSources(updateSources, geometryLayer) {
+    let fullUpdate = false;
+    const filtered = new Set();
+    updateSources.forEach((src) => {
+        if (src === geometryLayer || src.isCamera) {
+            fullUpdate = true;
+        } else if (src.layer === geometryLayer) {
+            filtered.add(src);
+        }
+    });
+    return fullUpdate ? new Set([geometryLayer]) : filtered;
+}
+
 MainLoop.prototype._update = function _update(view, updateSources, dt) {
     const context = {
         camera: view.camera,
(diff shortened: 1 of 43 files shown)


src/Core/MainLoop.js
```diff
@@ -81,17 +95,31 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
...
         if (geometryLayer.ready && geometryLayer.visible && !geometryLayer.frozen) {
             view.execFrameRequesters(MAIN_LOOP_EVENTS.BEFORE_LAYER_UPDATE, dt, this._updateLoopRestarted, geometryLayer);
 
-            // `preUpdate` returns an array of elements to update
-            const elementsToUpdate = geometryLayer.preUpdate(context, geometryLayer, updateSources);
-            // `update` is called in `updateElements`.
-            updateElements(context, geometryLayer, elementsToUpdate);
-            // `postUpdate` is called when this geom layer update process is finished
-            geometryLayer.postUpdate(context, geometryLayer, updateSources);
+            const srcs = filterChangeSources(updateSources, geometryLayer);
```
zarov: Could you add a quick comment on what is filtered ?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -156,8 +156,10 @@ class FirstPersonControls extends THREE.EventDispatcher {
             applyRotation(this.view, this.camera, this._state);
         }
 
-        if (this.moves.size || this._isMouseDown) {
-            this.view.notifyChange(true);
+        if (this.moves.size) {
+            this.view.notifyChange(this.view.camera.camera3D);
+        } else if (this._isMouseDown) {
+            // this.view.notifyChange();
```
zarov: What is the purpose here ?

utils/debug/PointCloudDebug.js
```diff
@@ -13,19 +13,19 @@ export default {
         layer.debugUI = datUi.addFolder(`${layer.id}`);
 
         layer.debugUI.add(layer, 'sseThreshold').name('SSE threshold')
-            .onChange(() => view.notifyChange(true));
+            .onChange(() => view.notifyChange(true, layer));
```
zarov: It should be the other way

zarov (commented):
You left some old `notifyChange` call in `planar.html`, `3dtiles.js` and `positionGlobe.js`.

Otherwise it makes more sense to always force the redraw by default but not the update.

zarov (approved):
LGTM, thanks !

PR #772, opened 2018-06-01
refacto(core): add or use optional target to getPickingPositionFromDepth

gchoqueux:
## Description
add or use optional target to getPickingPositionFromDepth

## Motivation and Context
 - Avoid excessive memory allocations that cause GC work.
If target is defined then use this to write the result instead of
allocating an objet
- it's also to avoid the returned result being modified by a new function call

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -375,10 +375,19 @@ GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, heig
 
 const matrix = new THREE.Matrix4();
 const screen = new THREE.Vector2();
-const pickWorldPosition = new THREE.Vector3();
 const ray = new THREE.Ray();
 const direction = new THREE.Vector3();
-GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse) {
+
+/**
+ * Returns the world position (view's crs: referenceCrs) under view coordinates.
+ * This position is computed with depth buffer.
+ *
+ * @param      {THREE.Vector2}  mouse  position in view coordinates (in pixel), if it's null so it's view's center.
+ * @param      {THREE.Vector3}  [target=THREE.Vector3()] target the result will be copied into this Vector3. If not present a new one will be created.
+ * @return     {THREE.Vector3}  the world position in view's crs: referenceCrs.
+ */
+
+GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse, target = new THREE.Vector3()) {
     const l = this.mainLoop;
     const viewPaused = l.scheduler.commandsWaitingExecutionCount() == 0 && l.renderingState == RENDERING_PAUSED;
     const g = l.gfxEngine;
@@ -423,14 +432,14 @@ GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFro
     const orthoZ = g.depthBufferRGBAValueToOrthoZ(buffer, camera);
     const length = orthoZ / Math.cos(angle);
 
-    pickWorldPosition.addVectors(camera.position, ray.direction.setLength(length));
+    target.addVectors(camera.position, ray.direction.setLength(length));
 
     camera.layers.mask = prev;
 
-    if (pickWorldPosition.length() > 10000000)
+    if (target.length() > 10000000)
         { return undefined; }
 
-    return pickWorldPosition;
+    return target;
 };
 
 GlobeView.prototype.setRealisticLightingOn = function setRealisticLightingOn(value) {
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -243,6 +243,7 @@ var initialPosition;
 var initialZoom;
 
 // picking
+const pickedPosition = new THREE.Vector3();
 const ptScreenClick = new THREE.Vector2();
 const sizeRendering = new THREE.Vector2();
 
@@ -789,11 +790,9 @@ function GlobeControls(view, target, radius, options = {}) {
 
         direction.subVectors(movingCameraTargetOnGlobe, this.camera.position);
 
-        const pickingPosition = view.getPickingPositionFromDepth();
-
         // Position movingCameraTargetOnGlobe on DME
-        if (pickingPosition) {
-            const distanceTarget = pickingPosition.distanceTo(this.camera.position);
+        if (view.getPickingPositionFromDepth(null, pickedPosition)) {
+            const distanceTarget = pickedPosition.distanceTo(this.camera.position);
             direction.setLength(distanceTarget);
             movingCameraTargetOnGlobe.addVectors(this.camera.position, direction);
         }
@@ -1082,13 +1081,11 @@ function GlobeControls(view, target, radius, options = {}) {
                     snapShotCamera.shot(this.camera);
                     ptScreenClick.x = x;
                     ptScreenClick.y = y;
-
-                    const point = view.getPickingPositionFromDepth(ptScreenClick);
                     lastRotation = [];
                     // update tangent sphere which passes through the point
-                    if (point) {
+                    if (view.getPickingPositionFromDepth(ptScreenClick, pickedPosition)) {
                         ctrl.range = this.getRange();
-                        updateSpherePicking.bind(this)(point, ptScreenClick);
+                        updateSpherePicking.bind(this)(pickedPosition, ptScreenClick);
                     } else {
                         state = this.states.NONE;
                     }
@@ -1121,11 +1118,9 @@ function GlobeControls(view, target, radius, options = {}) {
 
         // Double click throws move camera's target with animation
         if (!currentKey) {
-            const point = view.getPickingPositionFromDepth(this._view.eventToViewCoords(event));
-
-            if (point) {
+            if (view.getPickingPositionFromDepth(this._view.eventToViewCoords(event), pickedPosition)) {
                 animatedScale = 0.6;
-                this.setCameraTargetPosition(point);
+                this.setCameraTargetPosition(pickedPosition);
             }
         }
     };
@@ -1922,9 +1917,7 @@ GlobeControls.prototype.setCameraTargetGeoPositionAdvanced = function setCameraT
  * @return {Coordinates} position
  */
 GlobeControls.prototype.pickGeoPosition = function pickGeoPosition(windowCoords) {
-    var pickedPosition = this._view.getPickingPositionFromDepth(windowCoords);
-
-    if (!pickedPosition) {
+    if (!this._view.getPickingPositionFromDepth(windowCoords, pickedPosition)) {
         return;
     }
 
--- a/src/Renderer/ThreeExtended/PlanarControls.js
+++ b/src/Renderer/ThreeExtended/PlanarControls.js
@@ -438,16 +438,7 @@ function PlanarControls(view, options = {}) {
      * @ignore
      */
     this.initiateSmartZoom = function initiateSmartZoom() {
-        // point under mouse cursor
-        const pointUnderCursor = new THREE.Vector3();
-
-        // check if there is valid geometry under cursor
-        if (typeof this.view.getPickingPositionFromDepth(mousePosition) !== 'undefined') {
-            pointUnderCursor.copy(this.view.getPickingPositionFromDepth(mousePosition));
-        }
-        else {
-            return;
-        }
+        const pointUnderCursor = this.getWorldPointAtScreenXY(mousePosition);
 
         // direction of the movement, projected on xy plane and normalized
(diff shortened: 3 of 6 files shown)


autra:
> If target is defined then use this to write the result instead of
allocating an objet

I'm confused. The number of created object is bigger after this patch.

Unless I missed something, `view.getPickingPositionFromDepth` was already creating one and only one instance of Vector3 once and for all. So each module that uses `view.getPickingPositionFromDepth` will need to allocate a temp Vector3 somewhere now, raising the total number of created Vector3 at startup time.

Worst, after this change, getPickingPosition creates a Vector3 at each invocation by default: user *have to remember* to pass a Vector3 to this function to avoid an allocation. So any external client code that already uses `getPickingPositionFromDepth` will now create one Vector3 per call if they get this patch.

What are you trying to achieve exactly?

gchoqueux:
it's also to avoid the returned result being modified by a new function call

autra:
What about keeping the current code but document the fact returned value should be cloned if modification has to be made inplace? That way we keep the no-cloning by default and avoid breaking existing code.

gchoqueux:
it's a classic behavior function, like `Ray.intersectBox (), Ray.intersectPlane (), Coordinates.as(), Coordinates.offsetInExtent()`.. And it does not break the code.

peppsac (commented):
I don't mind the change but when modifying functions that are part of the API it would be nice to:
- document them (and insist that the function returns a **new** Vector3 each time, unless `target` is provided)
- add a test. For example in `planar.js` and `globe.js` add a call to `getPickingPositionFromDepth` and verify the value.

gchoqueux:
@peppsac  
* I document the functions
* Add test is nice, but out PR scope

peppsac:
> Add test is nice, but out PR scope

My review rule is: no test => no approve, and I strongly encourage other reviewers to do the same.

Now that we have unit testing and functional testing in place, there's no excuse to not add tests.

gchoqueux:
> My review rule is: no test => no approve

you didn't forget to apply your rule on PR #772 ?

*  as EXT_frag_depth extension not supported in headless mode, it isn't possible to test this function in headless mode

peppsac:
> you didn't forget to apply your rule on PR #772 ?

PR 772 is this PR :)

>     as EXT_frag_depth extension not supported in headless mode, it isn't possible to test this function in headless mode

What's the link between `EXT_frag_depth` and these functions?

gchoqueux:
Sorry, I was talking about the #767 

This function uses depth buffer, as `EXT_frag_depth` extension not supported in headless mode, the buffer values aren't correct.

peppsac:
> Sorry, I was talking about the #767

You must have been mistaken, because I didn't approve this PR yet.

> This function uses depth buffer, as EXT_frag_depth extension not supported in headless mode, the buffer values aren't correct.

Reading from the depth buffer doesn't require EXT_frag_depth (which allows you to write to the depth buffer from the fragment shader. See https://developer.mozilla.org/en-US/docs/Web/API/EXT_frag_depth).

So nice try, but you'll have to write a test for this function.

Oh, and if your next argument is "I can't write a test because logdepth buffer is broken when ext_fragdepth is missing" I can already answer: you can still write a test. Yes the value will be incorrect but at least the test will verify that the rest of the function work as expected.

gchoqueux:
tests added

examples/globe_depth.html
```diff
@@ -0,0 +1,46 @@
+<html>
```
peppsac: This file isn't useful and can be removed.

src/Core/Prefab/GlobeView.js
```diff
@@ -371,10 +371,17 @@ GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, heig
 
 const matrix = new THREE.Matrix4();
 const screen = new THREE.Vector2();
-const pickWorldPosition = new THREE.Vector3();
 const ray = new THREE.Ray();
 const direction = new THREE.Vector3();
-GlobeView.prototype.getPickingPositionFromDepth = function getPickingPositionFromDepth(mouse) {
+
+/**
+ * returns the pick position under the mouse
```
peppsac: Sorry, but this isn't a documentation.

What is a "pick position"?
When should I use this function?
Why does the function name ends with `FromDepth`?

If you don't answer this kind of questions, the documentation is useless.

src/Core/Prefab/GlobeView.js
```diff
@@ -371,10 +371,17 @@ GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, heig
...
+ *
+ * @param      {THREE.Vector2}  mouse   mouse coordinates to pick position
```
peppsac: What coordinates systems? Is this geographic coordinates? Or normalized? Or pixels?

src/Core/Prefab/GlobeView.js
```diff
@@ -371,10 +371,17 @@ GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, heig
...
+ * @param      {THREE.Vector3}  target optional Vector3 target. the result will be copied into this Vector3. If not present a new one will be created
```
peppsac: optional Vector3 target => use the jsdoc syntax to mark the parameter as optional instead.

src/Core/Prefab/GlobeView.js
```diff
@@ -371,10 +371,17 @@ GlobeView.prototype.readDepthBuffer = function readDepthBuffer(x, y, width, heig
...
+ * @return     {THREE.Vector3}  The picking position
```
peppsac: Obviously it will return "The picking position", because it's already written in the name of the function.

The documentation must specify what's missing from the function name (eg: units)

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -789,11 +790,9 @@ function GlobeControls(view, target, radius, options = {}) {
 
         direction.subVectors(movingCameraTargetOnGlobe, this.camera.position);
 
-        const pickingPosition = view.getPickingPositionFromDepth();
-
         // Position movingCameraTargetOnGlobe on DME
-        if (pickingPosition) {
-            const distanceTarget = pickingPosition.distanceTo(this.camera.position);
+        if (view.getPickingPositionFromDepth(null, pickedPosition)) {
```
peppsac, line 794: The `mouse = null` usecase isn't documented :-(

test/examples/planar.js
```diff
@@ -14,4 +14,36 @@ describe('planar', () => {
         assert.ok(result);
         await page.close();
     });
+    it('should get picking position from depth', async function _() {
+        const page = await browser.newPage();
+
+        await page.setViewport({ width: 400, height: 300 });
+        await page.goto(`http://localhost:${itownsPort}/examples/planar_vector.html`);
+        await page.waitFor('#viewerDiv > canvas');
+        await exampleCanRenderTest(page, this.test.fullTitle());
```
peppsac: You could speed up the test by removing this. No need to wait for all textures to be loaded since you're going to move the camera anyway.

test/examples/planar.js
```diff
@@ -14,4 +14,36 @@ describe('planar', () => {
...
+
+        // get range with depth buffer and altitude
+        const result = await page.evaluate((l) => {
+            const lookat = extent.center().xyz();
+
+            view.camera.camera3D.position.copy(lookat);
+            view.camera.camera3D.position.z = l;
+            view.camera.camera3D.lookAt(lookat);
+            view.notifyChange(view.camera.camera3D, true);
+            const depthMethod = view
```
peppsac: But here you should wait for all textures to be loaded - because you're using them in DEMUtils below.

test/examples/planar.js
```diff
@@ -14,4 +14,36 @@ describe('planar', () => {
...
     });
+    it('should get picking position from depth', async function _() {
+        const page = await browser.newPage();
+
+        await page.setViewport({ width: 400, height: 300 });
+        await page.goto(`http://localhost:${itownsPort}/examples/planar_vector.html`);
+        await page.waitFor('#viewerDiv > canvas');
+        await exampleCanRenderTest(page, this.test.fullTitle());
+
+        const length = 1500;
```
peppsac: height? altitude?

test/examples/planar.js
```diff
@@ -14,4 +14,36 @@ describe('planar', () => {
...
+            const depthMethod = view
+                .getPickingPositionFromDepth().distanceTo(view.camera.camera3D.position);
+
+            const altitude = itowns.DEMUtils
+                .getElevationValueAt(view.tileLayer, extent.center().clone()).z;
+
+            return { depthMethod, altitude };
+        }, length);
+        // threoric range between ground and camera
+        const theoricRange = length - result.altitude;
```
peppsac: theoricRange = theoricalRange (or theoricalDistance)

test/examples/planar.js
peppsac: and `depthMethod` is `distanceUsingDepthMethod`

peppsac (dismissed):
Thanks for the change and moving this PR in the good direction.

There's still work to do though, hopefully my comments will help improve the documentation and the tests.

nosy-b:
So much talk for a very small modification, can we move forward on this one? :)

peppsac:
> So much talk for a very small modification

:-1: 

> can we move forward on this one

I made extensive comments on what could be improved and how it can be improved, 28 days ago. 
I believe my requests are reasonnable, especially since the lack of documentation is a regular complaint.

But feel free to dismiss my review, if you think it's not relevant.

nosy-b:
@peppsac Yes, thank you for your comments and possible improvement suggestions. But what I want to bring here is that as a small team it's better to make deep reviews on important PR and lighter reviews for fix or minimal improvement.
More globally in this specific period I would suggest to let a bit of ease in the PR reviews to smoother the development processes.

peppsac:
> More globally in this specific period I would suggest to let a bit of ease in the PR reviews to smoother the development processes.

What specific period? When did it start? or will it end?

> to smoother the development processes.

One alternate way to smooth the development process without sacrifying too much on the quality is to review open PR quickly and follow up on comments quickly.

There are a lot of open PR and you're a member of itowns reviewer group so you can actually help.

nosy-b:
This period -> where few active reviewers are available.
And I think bringing some positivism could help attracting more iTowns dev/reviewers :)

src/Renderer/ThreeExtended/PlanarControls.js
```diff
@@ -442,10 +442,7 @@ function PlanarControls(view, options = {}) {
         const pointUnderCursor = new THREE.Vector3();
 
         // check if there is valid geometry under cursor
-        if (typeof this.view.getPickingPositionFromDepth(mousePosition) !== 'undefined') {
-            pointUnderCursor.copy(this.view.getPickingPositionFromDepth(mousePosition));
-        }
-        else {
+        if (typeof this.view.getPickingPositionFromDepth(mousePosition, pointUnderCursor) === 'undefined') {
```
zarov: You can also modify `getWorldPointAtScreenXY` to be something like `return getPickingPosition || getWorldPoint`.

test/examples/positionGlobe.js
```diff
@@ -55,6 +55,52 @@ describe('positionGlobe', () => {
...
+            }));
+
+        // Hide cone the cone and set range
+        const destRange = 1500;
+        await page.evaluate((range) => {
+            globeView.mesh.material.visible = false;
+            globeView.controls.setRange(range);
+        }, destRange);
+
+        // wait animation ending and get range value with globeControls method
```
zarov: Which animation ?

test/examples/positionGlobe.js
```diff
@@ -55,6 +55,52 @@ describe('positionGlobe', () => {
...
+        const controlsMethod = await page.evaluate(() =>
+            new Promise((resolve) => {
+                const endAni = () => {
+                    globeView.controls.removeEventListener('animation-ended', endAni);
+                    resolve(globeView.controls.getRange());
```
zarov, line 92: Sorry, I don't get the point of doing all of that ?

test/examples/positionGlobe.js
```diff
@@ -55,6 +55,52 @@ describe('positionGlobe', () => {
...
+        await page.close();
+    });
+    it('should get picking position from depth', async function _() {
+        const page = await browser.newPage();
+
+        await loadExample(page,
+            `http://localhost:${itownsPort}/examples/positionGlobe.html`,
+            this.test.fullTitle());
+
+        // wait cone creation
```
zarov: Same here, what is this cone ?

test/examples/planar.js
zarov: :+1:

zarov (commented):
LGTM in terms of functionality, but please update the documentation given the current comments and we'll be good.

gchoqueux:
updated

zarov (approved):
LGTM, thanks for this !

nosy-b (approved):

PR #774, opened 2018-06-01
Extended THREE.Quaternion with precise slerp

gchoqueux:
the calculation isn't precise enough with Quaternion.slerp

I'm going to need in a next PR to use `Quaternion.preciseSlerp`.
I make this function accessible with the `MainBundle.js`, but it may not be the best place.

### DIFF
--- a/src/Renderer/ThreeExtended/GlobeControls.js
+++ b/src/Renderer/ThreeExtended/GlobeControls.js
@@ -34,60 +34,6 @@ const CONTROL_KEYS = {
     S: 83,
 };
 
-// TODO: can be optimize for some uses
-var presiceSlerp = function presiceSlerp(qb, t) {
-    if (t === 0) {
-        return this;
-    }
-
-    if (t === 1) {
-        return this.copy(qb);
-    }
-
-    const x = this._x;
-    const y = this._y;
-    const z = this._z;
-    const w = this._w;
-
-    // http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/
-
-    var cosHalfTheta = w * qb._w + x * qb._x + y * qb._y + z * qb._z;
-
-    if (cosHalfTheta < 0) {
-        this._w = -qb._w;
-        this._x = -qb._x;
-        this._y = -qb._y;
-        this._z = -qb._z;
-
-        cosHalfTheta = -cosHalfTheta;
-    } else {
-        this.copy(qb);
-    }
-
-    if (cosHalfTheta >= 1.0) {
-        this._w = w;
-        this._x = x;
-        this._y = y;
-        this._z = z;
-
-        return this;
-    }
-
-    const sinHalfTheta = Math.sqrt(1.0 - cosHalfTheta * cosHalfTheta);
-    const halfTheta = Math.atan2(sinHalfTheta, cosHalfTheta);
-    const ratioA = Math.sin((1 - t) * halfTheta) / sinHalfTheta;
-    const ratioB = Math.sin(t * halfTheta) / sinHalfTheta;
-
-    this._w = (w * ratioA + this._w * ratioB);
-    this._x = (x * ratioA + this._x * ratioB);
-    this._y = (y * ratioA + this._y * ratioB);
-    this._z = (z * ratioA + this._z * ratioB);
-
-    this.onChangeCallback();
-
-    return this;
-};
-
 // private members
 const EPS = 0.000001;
 
@@ -156,10 +102,6 @@ const ctrl = {
     lengthTarget: 0,
     lengthCamera: 0,
 };
-
-ctrl.qDelta.presiceSlerp = presiceSlerp;
-quatGlobe.presiceSlerp = presiceSlerp;
-
 // Animation
 
 let enableAnimation = true;
@@ -173,9 +115,11 @@ var lastTimeMouseMove = 0;
 
 // Expression used to damp camera's moves
 var dampingMoveAnimatedExpression = (function getDampMoveAniExprFn() {
-    const damp = new THREE.Quaternion(0, 0, 0, 1);
+    const damp = [0, 0, 0, 1];
+    const result = [];
     return function dampingMoveAnimatedExpression(root) {
-        root.qDelta.presiceSlerp(damp, root.dampingFactor * 0.2);
+        THREE.Quaternion.slerpFlat(result, 0, root.qDelta.toArray(), 0, damp, 0, root.dampingFactor * 0.2);
+        root.qDelta.fromArray(result);
         root.quatGlobe.multiply(root.qDelta);
     };
 }());
@@ -206,7 +150,9 @@ function zoomCenterAnimatedExpression(root, progress) {
     // Rotation
     root.quatGlobe.set(0, 0, 0, 1);
     root.progress = 1 - Math.pow((1 - (Math.sin((progress - 0.5) * Math.PI) * 0.5 + 0.5)), 2);
-    root.quatGlobe.presiceSlerp(root.qDelta, root.progress);
+    const result = [];
+    THREE.Quaternion.slerpFlat(result, 0, root.quatGlobe.toArray(), 0, root.qDelta.toArray(), 0, root.progress);
+    root.quatGlobe.fromArray(result);
     // clamp
     clampToGround(root);
 }
--- a/test/quaternion_unit_test.js
+++ b/test/quaternion_unit_test.js
@@ -1,0 +1,43 @@
+import * as THREE from 'three';
+import assert from 'assert';
+/* global describe, it */
+
+describe('Quaternion', function () {
+    it('should compare imprecise Quaternion.slerp with Quaternion.slerpFlat with close values', () => {
+        const angleA = Math.PI * 0.3330;
+        const angleB = Math.PI * 0.3329;
+        const A = new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(0, 0, 1), angleA);
+        const B = new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(0, 0, 1), angleB);
+
+        const slerp = 0.25;
+        const slerpResult = [];
+        THREE.Quaternion.slerpFlat(slerpResult, 0, A.toArray(), 0, B.toArray(), 0, slerp);
+        const R = new THREE.Quaternion().fromArray(slerpResult);
+        const O = A.clone().slerp(B, slerp);
+
+        const F = new THREE.Euler().setFromQuaternion(R);
+        const S = new THREE.Euler().setFromQuaternion(O);
+
+        const result = (angleA * (1 - slerp)) + (angleB * (slerp));
+        assert.ok(result - F.z < 1e-15);
+        assert.ok(result - S.z > 1e-5);
+        if (result - S.z < 1e-15) {
+            // eslint-disable-next-line no-console
+            console.log(`If this failed, so itowns uses a version greater than or equal to 94. You could replace, itowns code, the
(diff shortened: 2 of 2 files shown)


peppsac:
Please add some unit tests.
At the very least a test case where the normal `Quaternion.slerp` isn't precise enough would be appreciated.

gchoqueux:
@peppsac  unit test added

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -34,60 +34,6 @@ const CONTROL_KEYS = {
     S: 83,
 };
 
-// TODO: can be optimize for some uses
-var presiceSlerp = function presiceSlerp(qb, t) {
```
nosy-b, line 38: Typo (all over) :It's 'precise' not presice

gchoqueux: done

src/Renderer/ThreeExtended/Quaternion.js
```diff
@@ -0,0 +1,56 @@
+import * as THREE from 'three';
+
+if (!THREE.Quaternion.prototype.preciseSlerp) {
+    THREE.Quaternion.prototype.preciseSlerp = function _preciseSlerp(qb, t) {
```
peppsac: This function looks very similar to [the one from three](https://github.com/mrdoob/three.js/blob/dev/src/math/Quaternion.js#L499).

Could you explain what is the difference?

gchoqueux: https://github.com/mrdoob/three.js/blob/dev/src/math/Quaternion.js#L538 : 

If the angle is small, it averages the values of the quaternion

peppsac: Ok.
Then please add some comments to your implementation stating that it's a copy-paste from three with this modification.
And a justification of why this modification is needed - because from an outside perspective I don't see any harm in averaging the quaternion.

Especially since the original algorithm explicitely says:
`When theta is 180 degrees the result is undefined because there is no shortest direction to rotate`.
So if I'm not wrong your solution isn't more "precise", it's simply a different implementation choice for the case where theta is ~180°.

WDYT?

gchoqueux: yes indeed, I commented and I took the condition when theta = 180 °

peppsac: Better thanks.

So, threejs rounds when the angle is `180° ± 0.114°`. And your solution theshold is `180° ± 0.00000114°`.
`0.114°` seems reasonnable, so can you explain why did you need `0.00000114°`?

Also, did you consider opening a PR to fix this issue directly in threejs (either by changing the constant or making it a parameter)?

gchoqueux: Because the equivalent of 0.001 ° on the surface of the globe is greater than 110 meters, so it's not precise enough

peppsac: But the issue only happen when `theta` (ie: the rotation needed to transform quaternion1 in quaternion2) is almost 180° - that would mean rotating completely around the globe, right? In which case the camera will travel ~20km, so the 110m error is less impressive.

Do you have a reproducable example so I can understand the problem you're solving?

gchoqueux: You can see this [example](https://github.com/iTowns/itowns/pull/774/files#diff-0936d27ecea5501d163f8546d2615538)

peppsac: Yeah. That's not what I meant :)

I meant: a real example in iTowns where the `slerp` "precision" issue is visible or has a visible effect.

gchoqueux: By example, if slerp is used in globeControls damping, the imprecisions cause error in this feature.

peppsac: I know it's in GlobeControls, since it's the only user of this function. I really need a step-by-step reproducer to understand the necessity of this PR.

You also forgot to answer a couple questions:
- did you consider opening a PR to fix this issue directly in threejs (either by changing the constant or making it a parameter)? (from https://github.com/iTowns/itowns/pull/774#discussion_r192720807)
- the issue only happen when theta (ie: the rotation needed to transform quaternion1 in quaternion2) is almost 180° - that would mean rotating completely around the globe, right?

gchoqueux: - replace preciseSlerp by slerp in globeControls and try move globe with only left button (being close to the ground). The damping doesn't work properly

- I have not considered for the moment to open a THREE PR 

- that would mean rotating completely around the globe, right?  it's possible, but the unit test also show that small rotations don't give the good result

peppsac: > replace preciseSlerp by slerp in globeControls and try move globe with only left button (being close to the ground). The damping doesn't work properly

AFAICT it's a logic issue in the GlobeControls code. It's doing lerp between 2 nearly identical quaternions.
e.g:
```js
        root.qDelta.preciseSlerp(damp, root.dampingFactor * 0.2);
```
You shouldn't accumulate the change in root.qDelta, but instead make the `t` parameter for `preciseSlerp` vary from 0 to 1 (otherwise you'll end up lerping identical quaternions)

> the unit test also show that small rotations don't give the good result

The first unittest shows that the result is different when testing nearly identical quaternion. But it misses the most useful part: comparing to a known good value (`assert.equal(R.w, known_value_computed_using_another_method`).

Anyway, IMHO this PR isn't necessary: the calling site (GlobeControls) should be fixed instead.

gchoqueux: I have new PR where I use `slerp` with `t` vary from 0 to 1, and there's the same problem. 
I push a unit test with another method, which shows that the projected error on the globe can reach 500 m

peppsac: > I have new PR where I use slerp with t vary from 0 to 1, and there's the same problem.

Did you push this branch somewhere so I can test it?

>  I push a unit test with another method, which shows that the projected error on the globe can reach 500 m

Did you push it?
Because it's the same test, lerping between 2 ~identical quaternions...

gchoqueux: yes, I [pushed](https://github.com/iTowns/itowns/pull/774/commits/a62d8214c59dc48d02f5124e8da353dc55887c59#diff-0936d27ecea5501d163f8546d2615538R9).
As announced from the beginning this PR, this test highlights the imprecision of the original method, which is also highlighted in `globeControl`. I don't think that  vary `t` from 0 to 1 fix the problem

peppsac: Actually the original paper says the result is undefined if theta is 180° but threejs code tests for sin(theta) ~= 0 which happens when theta is 0° or 180°.

So IMHO: the bug should be fixed upstream (for instance by replacing the if test with `cosHalfTheta <= 0 && Math.abs(sinHalfTheta) < 0.001`).

mbredif:
IMHO, this is an upstream issue (see the related issue : mrdoob/three.js#14258, see my [comment](mrdoob/three.js/issues/14258#issuecomment-397279052) there) : slerp should default to lerp instead of averaging.

@gchoqueux :  Have you tried the alternate slerp implementation using [Quaternion.slerpFlat](https://github.com/mrdoob/three.js/blob/013b604d20ffc6f0bcd1b0ac36f2c755d9b851ba/src/math/Quaternion.js#L43-L77) which is doing lerp for small angles (maybe with a different threshold) ?

gchoqueux:
@mbredif It works with Quaternion.slerpFlat, thank
I replaced in the PR

mbredif:
> It works with Quaternion.slerpFlat

good to know !
FYI, the upstream PR (mrdoob/three.js#14349)  is tagged for r95.
- @gchoqueux Could you provide the wxyz values of a pair of quaternions that is relevant for our use case (and currently failing), so that we can add it in the tests upstream ?
- @gchoqueux I think you can remove your `preciseSlerp` implementation (which is no longer used I think) and provide instead a reimplementation of slerp in terms of slerpFlat here (until we upgrade iTowns' THREE dependency)

gchoqueux:
* the tests in this Pr, provide somes wxyz values that is relevant for our use case
* yes of course, I want to delete preciseSlerp

gchoqueux:
@mbredif, I change test unit to compare slerpFlat and slerp with relevant  values

test/quaternion_unit_test.js
```diff
@@ -0,0 +1,38 @@
...
+        const slerpResult = [];
+        THREE.Quaternion.slerpFlat(slerpResult, 0, A.toArray(), 0, B.toArray(), 0, slerp);
+        const R = new THREE.Quaternion().fromArray(slerpResult);
+        const O = A.clone().slerp(B, slerp);
+
+        const F = new THREE.Euler().setFromQuaternion(R);
+        const S = new THREE.Euler().setFromQuaternion(O);
+
+        const result = (angleA * (1 - slerp)) + (angleB * (slerp));
+        assert.ok(result - F.z < 1e-15);
```
peppsac, line 22: These 2 tests are only testing three.js features, so I don't think they should be in iTowns repository.

gchoqueux: yes exactly, this allows us to test upstream changes and have a signal to make the change to `Quaternion.slerp`, I can add a message to the test

peppsac: I don't understand what signal this will send?

This test will pass, regardless of when upstream change its `Quaternion.slerp` method.
If you plan to use the "fixed" version of `Quaternion.slerp`, you could just wait for the next threejs release - instead of pushing this temp fix.

gchoqueux:
fix in THREE 94 [PR](https://github.com/mrdoob/three.js/pull/14349/files)
thanks @mbredif

test/quaternion_unit_test.js
gchoqueux: If someone upgrades THREE  to v94, in itowns, it will be indicated that it can also replace `Quaternion.preciseSlerp` by `Quaternion.slerp`,  I can add a message in the test to be more explicit

gchoqueux: I'll add a comparison with preciseSlerp

peppsac:
> fix in THREE 94 PR

Great. Then I'd suggest transforming this PR in a "update three.js PR" (+ `preciseSlerp` removal of course)

zarov:
> If someone upgrades THREE to v94, in itowns, it will be indicated that it can also replace Quaternion.preciseSlerp by Quaternion.slerp, I can add a message in the test to be more explicit

Could you do that please ? Then I suggest to move forward and merge this PR, as the upgrade to THREE 94 is a more complex task and not the goal of this PR. I'll open an issue about it.

peppsac:
> Could you do that please ? Then I suggest to move forward and merge this PR, as the upgrade to THREE 94 is a more complex task and not the goal of this PR. I'll open an issue about it.

Sorry but merging this PR as is doesn't make sense now, and that's a good thing (= I'm happy that the fix has been made upstream instead of in iTowns).

The "bug" fixed in three 94 is already fixed in iTowns (via GlobeControls presiceSlerp method), so there's no point in commiting a temporary test.

So there's no need to update the test message or similar.

peppsac (dismissed):
See https://github.com/iTowns/itowns/pull/774#issuecomment-407800687 for the details.

zarov:
I opened #824 for the update of THREE to v94, so I hope we will merge it quickly, but not in a month.

zarov (approved):
LGTM, the update to THREE.js is on the way but is going to take some time, so let's merge this PR in the meantime.

nosy-b (approved):

PR #775, opened 2018-06-04
feat(parser): add partial z support for geojson parsing

zarov:
According to RFC 7946 [1], we can have coordinates set as [lon, lat] but
also as [lon, lat, alt] in a geojson file. When parsing a geojson to
provide features, this wasn't respected and the z coordinates was always
set to 1. While this default value is not customizable yet, it is a
small improvement.

Unit tests for geojson have been added to reflect this change.

[1]: https://tools.ietf.org/html/rfc7946#section-9


gchoqueux (approved):
LGTM

PR #777, opened 2018-06-06
fix(core): remove correctly event listener after globe is initialized

gchoqueux:
## Description
remove correctly event listener after globe is initialized


zarov (approved):
LGTM, thanks for catching this !

Issue #776, opened 2018-06-05
Tiles not loading in PlanarView


zarov:
Since 881013dd, the `wfs.html` example is broken, and I suspect more too.

## Steps to reproduce

1- Go to an example using `PlanarView`, for instance http://www.itowns-project.org/itowns/examples/wfs.html
2- Wait for the page to load everything
3- Move the camera to see the bottom of the map

## Expected behavior

All tiles should be loading

![screenshot from 2018-06-05 15-07-15](https://user-images.githubusercontent.com/741255/40978282-7a65e97e-68d3-11e8-9ed7-2e6106297375.png)

## Actual behavior

Not all tiles are loading

![screenshot from 2018-06-05 15-06-55](https://user-images.githubusercontent.com/741255/40978216-4e892320-68d3-11e8-87e8-ff82d73af6f8.png)

## Possible Cause/Fix/Solution

It may be a problem with the last PR on `notifyChange`. Ping @peppsac as you may have an answer here

zarov:
After verification, it seems that it only occurs in example using `PlanarView`.

PR #778, opened 2018-06-06
Micro fixes

peppsac:
2 fix commits:

* fix(core): assign the correct layer property
This fixes a few places in the code missed during PR #758.

* fix(core): make sure PlanarControls call notifyChange when needed
This one fixes issue #776

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -110,7 +110,7 @@ export function createGlobeLayer(id, options) {
                 // some invisible tiles may now be visible
                 return layer.level0Nodes;
             }
-            if (source.layer === layer.id) {
+            if (source.layer === layer) {
                 if (!commonAncestor) {
                     commonAncestor = source;
                 } else {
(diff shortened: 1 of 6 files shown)


zarov (commented):
Thanks for those fixes, I'm sorry for not catching them before approbation ! About the `package-lock`, couldn't we update in another PR or at least commit ? Something like a `chore`

package-lock.json
```diff
@@ -26,7 +26,7 @@
       "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
       "dev": true,
       "requires": {
-        "acorn": "^4.0.3"
+        "acorn": "4.0.13"
```
gchoqueux: you could remove all changes on this file

peppsac:
The package-lock is a mistake: reverting this :)

package-lock.json
peppsac: Yep, fixed.

zarov:
There are some missing, if I'm not mistaken:
- https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/PlanarView.js#L57
- https://github.com/iTowns/itowns/blob/master/src/Core/View.js#L655
- https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/GlobeView.js#L113

I'll let you confirm and fix those !

peppsac:
> I'll let you confirm and fix those !

Fixed, thanks.

zarov (approved):

PR #779, opened 2018-06-06
fix(debug): apply notifyChange breaking change on debug Tools

gchoqueux:
apply notifyChange breaking change on debug Tools


zarov (approved):
LGTM, thanks and again sorry for missing that in the review !

PR #780, opened 2018-06-07
fix(pointcloud) : layer.visible and PointsMaterial.copy

mbredif:
## Description
post review fixes of #764 

## Motivation and Context
see #764

### DIFF
--- a/src/Renderer/PointsMaterial.js
+++ b/src/Renderer/PointsMaterial.js
@@ -44,7 +44,7 @@ class PointsMaterial extends RawShaderMaterial {
         this.uniforms.overlayColor.value = this.overlayColor;
     }
 
-    copy(source) {
+    update(source) {
         this.visible = source.visible;
         this.opacity = source.opacity;
         this.transparent = source.transparent;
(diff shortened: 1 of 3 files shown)


peppsac (approved):
Thanks for the quick fixes.
One side note: the git commit messages should contain the rationale of the changes, not just a link to another issue/PR (to avoid depending to much on the code hosting solution).

peppsac:
(and fix the commit title which doesn't respect #518 as well)

mbredif:
okay, okay... :)

peppsac:
> okay, okay... :)

Believe it or not but commit messages are useful :)

src/Renderer/PointsMaterial.js
```diff
@@ -44,7 +44,7 @@ class PointsMaterial extends RawShaderMaterial {
         this.uniforms.overlayColor.value = this.overlayColor;
     }
 
-    copy(source) {
+    update(source) {
```
autra, line 47: I remember you mentioned calling the super method didn't work... Did you have the chance to investigate why @mbredif?

PR #782, opened 2018-06-22
fix (core): get max texture sampler unit with mesa/firefox

gchoqueux:
## Description

fix mesa/firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=777028)

### DIFF
--- a/src/Core/System/Capabilities.js
+++ b/src/Core/System/Capabilities.js
@@ -1,11 +1,28 @@
+import SampleTestFS from '../../Renderer/Shader/SampleTestFS.glsl';
+import SampleTestVS from '../../Renderer/Shader/SampleTestVS.glsl';
+
 // default values
 let logDepthBufferSupported = false;
 let maxTexturesUnits = 8;
+
+function _WebGLShader(renderer, type, string) {
+    const gl = renderer.context;
+    const shader = gl.createShader(type);
+
+    gl.shaderSource(shader, string);
+    gl.compileShader(shader);
+    return shader;
+}
+
+function isFirefox() {
+    return navigator && navigator.userAgent && navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+}
 
 export default {
     isLogDepthBufferSupported() {
         return logDepthBufferSupported;
     },
+    isFirefox,
     isInternetExplorer() {
         const internetExplorer = false || !!document.documentMode;
         return internetExplorer;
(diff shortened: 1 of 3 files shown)


src/Core/System/Capabilities.js
```diff
@@ -17,16 +34,33 @@ export default {
...
+
+        const glFragmentShader = _WebGLShader(renderer, gl.FRAGMENT_SHADER, fragmentShader);
+
+        gl.attachShader(program, glVertexShader);
+        gl.attachShader(program, glFragmentShader);
+        gl.linkProgram(program);
+
+        if (gl.getProgramParameter(program, gl.LINK_STATUS) === false) {
+            // eslint-disable-next-line no-console
+            console.warn(`Sampler limit exceeded: it's down from ${maxTexturesUnits} to 16`);
```
peppsac: What about:
- printing the actual error message from WebGL instead of `Sampler limit exceeded`? 
- and replacing `it's down from ${maxTexturesUnits} to 16` by `using a maximum of 16 texture units instead of the reported value (${maxTexturesUnits})`

You should also make sure that this error essage only occurs when `maxTexturesUnits`is > 16 (and add another case when it's <= 16 but the `gl.linkProgram()`fails).

src/Core/System/Capabilities.js
```diff
@@ -17,16 +34,33 @@ export default {
...
+            if (isFirefox()) {
+                // eslint-disable-next-line no-console
+                console.warn('it can come from a bug mesa/firefox \n\n' +
```
peppsac: You could do a multiline string like this:
```js
console.warn(`It can come from a bug mesa/firefox.
Error compile shader when using more than 16 sampler uniforms,
see https://bugzilla.mozilla.org/show_bug.cgi?id=777028`);
```

src/Core/System/Capabilities.js
gchoqueux: done

src/Core/System/Capabilities.js
gchoqueux: done

gchoqueux:
@iTowns/reviewers can be merged?

src/Core/System/Capabilities.js
```diff
@@ -17,16 +34,39 @@ export default {
...
+                const info = gl.getProgramInfoLog(program);
+                // eslint-disable-next-line no-console
+                console.warn(`${info}: using a maximum of 16 texture units instead of the reported value (${maxTexturesUnits})`);
+                if (isFirefox()) {
+                    // eslint-disable-next-line no-console
+                    console.warn(`it can come from a bug mesa/firefox.
+                        Error compile shader when using more than 16 sampler uniforms,
+                        see https://bugzilla.mozilla.org/show_bug.cgi?id=777028`);
+                }
                 maxTexturesUnits = Math.min(16, maxTexturesUnits);
```
zarov: Shouldn't be necessary as we checked above that `maxTexturesUnits > 16`.

src/Core/System/Capabilities.js
```diff
@@ -17,16 +34,39 @@ export default {
...
+        gl.linkProgram(program);
+
+        if (gl.getProgramParameter(program, gl.LINK_STATUS) === false) {
+            if (maxTexturesUnits > 16) {
+                const info = gl.getProgramInfoLog(program);
+                // eslint-disable-next-line no-console
+                console.warn(`${info}: using a maximum of 16 texture units instead of the reported value (${maxTexturesUnits})`);
+                if (isFirefox()) {
+                    // eslint-disable-next-line no-console
+                    console.warn(`it can come from a bug mesa/firefox.
```
zarov: Don't forget capitals ! Also I would write a slightly more understandable message:
> It can come from a Mesa/Firefox bug; the shader compiles to an error when using more than 16 sampler uniforms, see https://bugzilla.mozilla.org/show_bug.cgi?id=777028

src/Core/System/Capabilities.js
```diff
@@ -17,16 +34,39 @@ export default {
...
+                        Error compile shader when using more than 16 sampler uniforms,
+                        see https://bugzilla.mozilla.org/show_bug.cgi?id=777028`);
+                }
                 maxTexturesUnits = Math.min(16, maxTexturesUnits);
+            } else {
+                throw (new Error(`the GPU capabilities could not be determined accurately.
```
zarov: Capital too ;)

zarov (approved):
LGTM, some things to change. Good to have a more accurate test about this problem.

PR #783, opened 2018-06-25
fix(core): stop to add layer if there's a error

gchoqueux:
## Description
Stop to add layer if there's a function error

## Motivation and Context
The layers work is disrupted
By example, if a layer is added with the same id, it's still add.
if we delete a layer then both are deleted

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -280,6 +280,9 @@ GlobeView.prototype = Object.create(View.prototype);
 GlobeView.prototype.constructor = GlobeView;
 
 GlobeView.prototype.addLayer = function addLayer(layer) {
+    if (!layer) {
+        return new Promise((resolve, reject) => reject(new Error('layer is undefined')));
+    }
     if (layer.type == 'color') {
         const colorLayerCount = this.getLayers(l => l.type === 'color').length;
         layer.sequence = colorLayerCount;
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -319,9 +319,14 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         layer.update = layer.update || updateLayeredMaterialNodeElevation;
     }
     return new Promise((resolve, reject) => {
+        if (!layer) {
+            reject(new Error('layer is undefined'));
+            return;
+        }
         const duplicate = this.getLayers((l => l.id == layer.id));
         if (duplicate.length > 0) {
             reject(new Error(`Invalid id '${layer.id}': id already used`));
+            return;
         }
 
         if (parentLayer && !layer.extent) {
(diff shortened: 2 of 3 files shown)


zarov (approved):
LGTM, good catch thanks

src/Core/Prefab/GlobeView.js
```diff
@@ -281,6 +281,9 @@ GlobeView.prototype = Object.create(View.prototype);
 GlobeView.prototype.constructor = GlobeView;
 
 GlobeView.prototype.addLayer = function addLayer(layer) {
+    if (!layer) {
```
peppsac, line 283: Are these check really necessary ?
(if they are, you should add them to View.addLayer and PanoramaView.addLayer as well)

autra: Also, this breaks current code: now this function can return undefined (it couldn't before). Please consider returning a rejected promise with a suitable `Error` object.

src/Core/View.js
```diff
@@ -325,16 +326,19 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
         if (parentLayer) {
             parentLayer.attach(layer);
         } else {
             if (typeof (layer.update) !== 'function') {
                 reject(new Error('Cant add GeometryLayer: missing a update function'));
+                return;
             }
             if (typeof (layer.preUpdate) !== 'function') {
                 reject(new Error('Cant add GeometryLayer: missing a preUpdate function'));
+                return;
```
autra, line 351: :+1:

src/Core/Prefab/GlobeView.js
gchoqueux: yes it is necessary because this error is not handle
I added error handling in all `addLayer`

autra: At the same time, is it really necessary to do so? If we let the code like it is before, users would have a rejected promise with "Error cannot read "type" of undefined" or something like that, which is imho clear enough. Anyway, why not having a better error message.

src/Core/Prefab/GlobeView.js
```diff
@@ -280,6 +280,9 @@ GlobeView.prototype = Object.create(View.prototype);
...
+        return new Promise((resolve, reject) => reject(new Error('layer is undefined')));
```
autra, line 284: (you can do 

```js
return Promise.reject(new Error(...))
```
btw )

src/Core/View.js
```diff
@@ -319,9 +319,14 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         layer.update = layer.update || updateLayeredMaterialNodeElevation;
     }
     return new Promise((resolve, reject) => {
+        if (!layer) {
```
autra, line 322: Oh we missed that: this is dead code, if layer is undefined, it'd throw on line 316 before reaching this.

autra (commented):
2 nits, if someone wants to make a followup.

PR #784, opened 2018-06-25
feat (core): restriction to maximum count of color layers

gchoqueux:
## Description
Add restriction to maximum count of color layers

## Motivation and Context
Adding color layers may overflow graphics card capabilities

## Screenshots
**example if you exceed limit**
bug layer

### DIFF
--- a/examples/js/GUI/GuiTools.js
+++ b/examples/js/GUI/GuiTools.js
@@ -47,6 +47,14 @@ function GuiTools(domId, view, w) {
     }
 }
 
+GuiTools.prototype.addLayerGUI = function fnAddLayerGUI(layer) {
+    if (layer.type === 'color') {
+        this.addImageryLayerGUI(layer);
+    } else if (layer.type === 'elevation') {
+        this.addElevationLayerGUI(layer);
+    }
+};
+
 GuiTools.prototype.addLayersGUI = function fnAddLayersGUI() {
     function filterColor(l) { return l.type === 'color'; }
     function filterElevation(l) { return l.type === 'elevation'; }
--- a/examples/layersColorVisible.html
+++ b/examples/layersColorVisible.html
@@ -26,13 +26,12 @@
             var menuGlobe = new GuiTools('menuDiv', globeView);
 
             function addLayerCb(layer) {
-                return globeView.addLayer(layer);
+                globeView.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
             }
 
-            var promises = [];
-            promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb));
-            promises.push(itowns.Fetcher.json('./layers/JSONLayers/Region.json').then(addLayerCb));
-            promises.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT.json').then(addLayerCb));
+            itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(addLayerCb);
+            itowns.Fetcher.json('./layers/JSONLayers/Region.json').then(addLayerCb);
+            itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT.json').then(addLayerCb);
 
             function removeDups(a) {
                 var temp = {};
@@ -69,10 +68,6 @@
             globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 // eslint-disable-next-line no-console
                 console.info('Globe initialized');
-                Promise.all(promises).then(function () {
-                    menuGlobe.addImageryLayersGUI(globeView.getLayers(function(l) { return l.type === 'color'; }));
-                    menuGlobe.addElevationLayersGUI(globeView.getLayers(function (l) { return l.type === 'elevation'; }));
-                }).catch(console.error);
             });
 
             // Use frame requester to update menu before rendering
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -4,6 +4,8 @@ import Camera from '../Renderer/Camera';
 import MainLoop, { MAIN_LOOP_EVENTS, RENDERING_PAUSED } from './MainLoop';
 import c3DEngine from '../Renderer/c3DEngine';
 
+import { getMaxColorSamplerUnitsCount } from '../Renderer/LayeredMaterial';
+
 import Layer from '../Layer/Layer';
 import ColorLayer from '../Layer/ColorLayer';
 import ElevationLayer from '../Layer/ElevationLayer';
@@ -293,7 +295,20 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         }
         layer = _preprocessLayer(this, layer, provider, parentLayer);
         if (parentLayer) {
-            parentLayer.attach(layer);
+            if (layer.type == 'color') {
+                const layerColors = this.getLayers(l => l.type === 'color');
+
+                const sumColorLayers = parentLayer.countColorLayersTextures(...layerColors, layer);
+
+                if (sumColorLayers <= getMaxColorSamplerUnitsCount()) {
+                    parentLayer.attach(layer);
+                } else {
+                    reject(new Error(`Cant add color layer ${layer.id}: the maximum layer is reached`));
+                    return;
+                }
+            } else {
+                parentLayer.attach(layer);
+            }
         } else {
             if (typeof (layer.update) !== 'function') {
                 reject(new Error('Cant add GeometryLayer: missing a update function'));
--- a/src/Provider/WMTSProvider.js
+++ b/src/Provider/WMTSProvider.js
@@ -33,7 +33,16 @@ function preprocessDataLayer(layer) {
         options.version = options.version || '1.0.0';
         options.tileMatrixSet = options.tileMatrixSet || 'WGS84';
         options.style = options.style || 'normal';
-        options.projection = options.projection || 'EPSG:3857';
+        // If the projection is undefined,
+        // It is deduced from the tileMatrixSet,
+        // The projection is coherent with the projection
+        if (!options.projection) {
+            if (options.tileMatrixSet === 'WGS84' || options.tileMatrixSet === 'WGS84G') {
+                options.projection = 'EPSG:4326';
+            } else {
+                options.projection = 'EPSG:3857';
+            }
+        }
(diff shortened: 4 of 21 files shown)


examples/js/GUI/GuiTools.js
```diff
@@ -47,6 +47,14 @@ function GuiTools(domId, view, w) {
     }
 }
 
+GuiTools.prototype.addLayerGUI = function fnAddLayerGUI(layer) {
```
zarov, line 50: :+1:

test/examples/globe.js
```diff
@@ -2,6 +2,45 @@
...
+        error = await page.evaluate(() => itowns.Fetcher.json('./layers/JSONLayers/Transport.json').then(addLayerCb)).catch(() => true);
+        error = await page.evaluate(() => itowns.Fetcher.json('./layers/JSONLayers/Railways.json').then(addLayerCb)).catch(() => true);
+        error = await page.evaluate(() => itowns.Fetcher.json('./layers/JSONLayers/Denomination.json').then(addLayerCb)).catch(() => true);
+        const colorLayersCount = await page.evaluate(() => globeView.getLayers(l => l.type === 'color').length);
+
+        assert.ok(error && colorLayersCount < 8);
+        page.close();
+        await page.close();
+    });
+    it('shouldn t add layer with id already used', async function _() {
```
zarov: You can escape the quote character with `\`

src/Core/View.js
```diff
@@ -325,16 +327,34 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+                const layerColors = this.getLayers(l => l.type === 'color');
+
+                const sumColorLayers = parentLayer.sumOccupancyColorLayers ?
+                parentLayer.sumOccupancyColorLayers(layerColors, layer) :
+                layerColors.length + 1;
+
+                if (sumColorLayers <= getMaxColorSamplerUnitsCount()) {
+                    parentLayer.attach(layer);
+                } else {
+                    reject(new Error(`Cant add color layer ${layer.id}: the maximum layer is reached`));
```
zarov, line 306: Couldn't we just emit a warning here ?

zarov (commented):
Good call on adding the `addLayerGUI` method in `GUITools`. This commit (4f633f7) could be put out of this PR for a quick merge if you want.

Otherwise this seems good to me, I would prefer a warning alerting the user that too much layer will result in graphical errors. 

A sidenote, not really in the scope of this PR: couldn't we move the `createGlobeLayer` part in a `GlobeLayer` file, inheriting from `GeometryLayer` ? I'd be to discuss it outside.

src/Core/Prefab/GlobeView.js
```diff
@@ -144,6 +144,18 @@ export function createGlobeLayer(id, options) {
         return false;
     }
 
+    wgs84TileLayer.sumOccupancyColorLayers = (layerColors, layer) => {
```
peppsac: `sumOccupancyColorLayers` is a weird name.

This functions counts how many textures (samplers) are required in the shader for color layers, so maybe `countColorLayersTextures`?

examples/layersColorVisible.html
```diff
@@ -26,13 +26,12 @@
             var menuGlobe = new GuiTools('menuDiv', globeView);
 
             function addLayerCb(layer) {
-                return globeView.addLayer(layer);
+                return globeView.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
```
peppsac: addLayerGui returns nothing, so you can remove the `return` in all `addLayerCb` functions.

test/examples/globe.js
```diff
@@ -2,6 +2,45 @@
...
+        const page = await browser.newPage();
+
+        await page.setViewport({ width: 400, height: 300 });
+        await page.goto(`http://localhost:${itownsPort}/examples/globe.html`);
+        await page.waitFor('#viewerDiv > canvas');
+
+        await exampleCanRenderTest(page, this.test.fullTitle());
+
+        let error;
+        error = await page.evaluate(() => itowns.Fetcher.json('./layers/JSONLayers/ScanEX.json').then(addLayerCb)).catch(() => true);
```
peppsac: Why do you assign error each time? (because only the last one will be effective)
How do you know for sure that the limit is `8`?

If what this test does is adding layer until there's an error then I'd suggest to do it this way:
- execute a first call to page.evaluate() to get the max number of textures (N)
- then execute N - 1 times somethign that adds a new layer and verify that there's no error
- execute one more time and verify there's an error.

peppsac (dismissed):

src/Core/Prefab/GlobeView.js
gchoqueux: yes

examples/layersColorVisible.html
gchoqueux: removed

test/examples/globe.js
gchoqueux: done

gchoqueux:
PR Updated

src/Core/View.js
```diff
@@ -340,7 +341,22 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
         }
         layer = _preprocessLayer(this, layer, provider, parentLayer);
         if (parentLayer) {
-            parentLayer.attach(layer);
+            if (layer.type == 'color') {
+                const layerColors = this.getLayers(l => l.type === 'color');
+
+                const sumColorLayers = parentLayer.countColorLayersTextures ?
+                parentLayer.countColorLayersTextures(layerColors, layer) :
+                layerColors.length + 1;
```
peppsac: Why `+1`?

src/Provider/WMTSProvider.js
```diff
@@ -32,7 +32,13 @@ function preprocessDataLayer(layer) {
         options.version = options.version || '1.0.0';
         options.tileMatrixSet = options.tileMatrixSet || 'WGS84';
         options.style = options.style || 'normal';
-        options.projection = options.projection || 'EPSG:3857';
+        if (!options.projection) {
```
peppsac: Could you add a comment to explain this change?

test/examples/globe.js
```diff
@@ -33,4 +34,61 @@ describe('globe', () => {
...
+        // add one layer just over the capacity limit
+        // verify if the error is handled
+        const errorOverLimit = await page.evaluate(() =>
+            itowns.Fetcher.json('./layers/JSONLayers/OrthosCRS.json').then((params) => {
+                const layerParams = Object.assign({}, params);
+                layerParams.id = 'max';
+                return globeView.addLayer(layerParams).then(() => false).catch(() => true);
+            }));
+
+        assert.ok(underLimit);
```
peppsac, line 62: The `assert` could be moved above (just after `underLimit` is assigned).

peppsac (commented):
Looks almost ok.

src/Core/View.js
gchoqueux: by default the number of layers is the same as the number of textures used. (+1 is for added layer)

peppsac: Ah ok.

Some parentheses are needed (`const sumColorLayers = (... ? ... : ...) + 1`) to ease readability and to fix the priority bug in your code: the `+1` only applies in the `layerColors.length` case.

gchoqueux: there's no bug
by default the number of layers is the same as the number of textures used :  `layerColors.length + 1`

if not, we use `countColorLayersTextures` which returns the number of textures used after adding layer

src/Provider/WMTSProvider.js
gchoqueux: If the projection is undefined, it can be deduced from the `tileMatrixSet`, previously the projection was incoherent with the projection

test/examples/globe.js
gchoqueux: why?

src/Core/View.js
peppsac: Oh, right. That's because `countColorLayersTextures` counts the not-added layer.

I'd suggest a less confusing API:
```js
wgs84TileLayer.countColorLayersTextures = (...layers) => {
   for (const layer of layers) {
      ...
   }
}
```
And at the calling site use:
```js
parentLayer.countColorLayersTextures(...layerColors, layer)
```

---

As a side note: you're defining this function in GlobeView but it clearly depends on TileMesh and LayeredMaterial so could also live in TileProvider, or in LayeredMaterial... and this is not documented either.

But I don't mind it (= I'm not clicking *Request Changes*) because there's no perfect solution, and this is mainly visible for itowns dev, not itowns user.

(the last paragraph is more a "how to (try to) be a useful reviewer" note, but I felt it was relevant, given the amount of blocked PR we have)

gchoqueux:
PR rebased and updated

src/Renderer/LayeredMaterial.js
```diff
@@ -57,6 +57,14 @@ var moveElementArray = function moveElementArray(array, oldIndex, newIndex) {
     array.splice(newIndex, 0, array.splice(oldIndex, 1)[0]);
 };
 
+const maxSamplersColorCount = 15;
+const samplersElevationCount = 1;
+
+export function getMaxColorSamplerUnitsCount() {
```
zarov: Maybe it could move to `Capabilities` ?

test/examples/globe.js
```diff
@@ -25,4 +25,56 @@ describe('globe', () => {
...
+        assert.ok(errorOverLimit);
+        await page.close();
+    });
+    it('shouldn t add layer with id already used', async function _() {
```
zarov: You can escape the `'`: `\'`.

zarov (commented):
Looks good ! Could you rebase one more time to add modifications to the `globe_vector_tiles` example ?

gchoqueux:
new raise,  before merge
@peppsac, the changes suit you?

src/Renderer/LayeredMaterial.js
nosy-b: Looks ok: Already in cap for the test `Capabilities.getMaxTextureUnitsCount() `
` maxSamplersColorCount = 15` is specific to layeredMaterial

gchoqueux:
PR Updated

gchoqueux:
@nosy-b ?

nosy-b (approved):
LGTM

nosy-b:
@gchoqueux please rebase

gchoqueux:
@nosy-b rebased

PR #785, opened 2018-06-25
refactor(tilemesh): move getCoordsForLayer+getZoomForLayer to Providers

mbredif:
## Description
This refactor moves the `getCoordsForLayer` and `getZoomForLayer` from `TileMesh` to new methods in layers (`getCoords` and `getZoom`) that are configured in the Providers.

## Motivation and Context
IMHO, code in Core should not depend explicitly (via `import`) or `implicitly` (via switching on properties defined outside the core, such as the layer `protocol` here). This will help untangle the dependencies and improve the modularity of the code. 

not relevant (no functionnal change)

## Note
This is the first commit of a series (cf see https://github.com/iTowns/itowns/compare/master...mbredif:refactor_coords to see where I am heading)

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -8,8 +8,7 @@ import * as THREE from 'three';
 import LayeredMaterial from '../Renderer/LayeredMaterial';
 import { l_ELEVATION } from '../Renderer/LayeredMaterialConstants';
 import RendererConstant from '../Renderer/RendererConstant';
-import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from '../Provider/OGCWebServiceHelper';
-import { is4326 } from './Geographic/Coordinates';
+import { SIZE_TEXTURE_TILE } from '../Provider/OGCWebServiceHelper';
 
 function TileMesh(geometry, params) {
     // Constructor
@@ -173,8 +172,7 @@ TileMesh.prototype.isElevationLayerLoaded = function isElevationLayerLoaded() {
 };
 
 TileMesh.prototype.isColorLayerDownscaled = function isColorLayerDownscaled(layer) {
-    const mat = this.material;
-    return mat.isColorLayerDownscaled(layer.id, this.getZoomForLayer(layer));
+    return this.material.isColorLayerDownscaled(layer.id, layer.getZoom(this));
 };
 
 TileMesh.prototype.OBB = function OBB() {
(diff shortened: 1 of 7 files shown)


src/Core/View.js
```diff
@@ -141,6 +141,9 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
         };
     }
 
+    layer.getCoords = layer.getCoords || (node => [node.extent]);
```
gchoqueux, line 144: it's necessary that any layer is these functions?

mbredif: I have no clue. It is only a refactor that does not change any functionality (and `getCoordsForLayer` was returning `[node.extent]` as a general case for all layers)

src/Core/TileMesh.js
```diff
@@ -204,38 +202,13 @@ TileMesh.prototype.changeSequenceLayers = function changeSequenceLayers(sequence
 };
 
 TileMesh.prototype.getCoordsForLayer = function getCoordsForLayer(layer) {
-    if (layer.protocol.indexOf('wmts') == 0) {
```
gchoqueux, line 207: for now, I'm writing two static functions

`getCoordsForLayer (layer, node)` and `getZoomForLayer (layer, node)`

this will avoid contaminating the `preprocessDataLayer` which should only affect the properties of the source (WMTS, WMS...)

src/Core/View.js
gchoqueux: with the static functions as seen above you could move this part in this one

peppsac, line 144: I support the effort of moving things out of TileMesh but at the same time I don't think that adding a `getZoom` API for layers is a good idea.
The `zoom` (or level) concept only applies to parts of iTowns (tiled geometries and protocols) and making a global API out of a local concept should be avoided.

peppsac: FWIW, I have a wip branch that removes the notion of level from everyplaces where it doesn't make sense.

You can see the diff the branch here: https://github.com/iTowns/itowns/compare/master...peppsac:colorisation

The changeset is quite large so I'll have to split it in several PR / commits, but here's an overview:
  - splits LayeredMaterialNodeProcessing in 2 (color and texture)
  - moves code that depends on the protocol in the provider (see the new function: `canTextureBeImproved`) and simplifies a lot the texture processing code
  - allows coloring pointclouds with WMS textures (WMTS could work later, but it needs a bit more work) - see the pointcloud_lyon example)

mbredif: Do you agree on `getCoords` ?

I don't like `getZoom` altogether as well :). it is only used by the LayeredMaterial AFAIK and I have a huge PR on a refacto of this material that does not need getZoom anymore. So the long-term plan is to discard it altogether.

mbredif: > The zoom (or level) 

Is it the same thing ? why do we have both terminologies ?

peppsac: I think the `getCoords` function would be unnecessary in my branch as well because the only thing that prevents the removal in your branch is that LayeredMaterialNodeProcessing uses it.

gchoqueux:
`TileMesh.getExtentsByProjection(crs)` method returns extents in crs projection system.
The extents are computed with the method `Extent.tiledCovering(crs)`.
This part has evolved and now much simpler and clearer.
Thank you for this proposal, but it is unfortunately obsolete.

PR #786, opened 2018-06-25
refacto(core): auto-assign texture layer update function

peppsac:
Currently itowns only support one update function for textures so
let's assign it by default in View.


zarov (approved):
LGTM thanks !

PR #787, opened 2018-06-25
refacto(3dtiles): adapt PntsParser to return a Geometry

peppsac:
This PR applies the idea from PR #764 and adapt it for 3dtiles pointclouds.

So now the PntsParser returns a BufferGeometry and the Provider create
a renderable object out of it.

### DIFF
--- a/src/Parser/PntsParser.js
+++ b/src/Parser/PntsParser.js
@@ -68,7 +68,6 @@ export default {
 function parseFeatureBinary(array, byteOffset, FTJSONLength) {
     // Init geometry
     const geometry = new THREE.BufferGeometry();
-    const material = new THREE.PointsMaterial({ size: 0.05, vertexColors: THREE.VertexColors, sizeAttenuation: true });
 
     // init Array feature binary
     const subArrayJson = utf8Decoder.decode(new Uint8Array(array, byteOffset, FTJSONLength));
@@ -105,14 +104,13 @@ function parseFeatureBinary(array, byteOffset, FTJSONLength) {
     if (parseJSON.BATCH_ID) {
         throw new Error('For pnts loader, BATCH_ID: not yet managed');
     }
-    // creation points with geometry and material
-    const points = new THREE.Points(geometry, material);
-    points.realPointCount = lengthFeature;
 
     // Add RTC feature
-    if (parseJSON.RTC_CENTER) {
-        points.position.fromArray(parseJSON.RTC_CENTER);
-    }
+    const offset = parseJSON.RTC_CENTER ?
+        new THREE.Vector3().fromArray(parseJSON.RTC_CENTER) : undefined;
 
-    return points;
+    return {
+        geometry,
+        offset,
+    };
 }
--- a/src/Provider/3dTilesProvider.js
+++ b/src/Provider/3dTilesProvider.js
@@ -131,8 +131,21 @@ function b3dmToMesh(data, layer, url) {
     });
 }
 
-function pntsParse(data) {
-    return PntsParser.parse(data).then(result => ({ object3d: result.point }));
+function pntsParse(data, layer) {
+    return PntsParser.parse(data).then((result) => {
+        const material = layer.material ?
+            layer.material.clone() :
+            new THREE.PointsMaterial({ size: 0.05, vertexColors: THREE.VertexColors });
+
+        // creation points with geometry and material
+        const points = new THREE.Points(result.point.geometry, material);
+
+        if (result.point.offset) {
+            points.position.copy(result.point.offset);
+        }
+
+        return { object3d: points };
+    });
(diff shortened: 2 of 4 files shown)


src/Parser/PntsParser.js
```diff
@@ -105,14 +104,13 @@ function parseFeatureBinary(array, byteOffset, FTJSONLength) {
...
-        points.position.fromArray(parseJSON.RTC_CENTER);
-    }
+    const offset = parseJSON.RTC_CENTER ?
+        new THREE.Vector3().fromArray(parseJSON.RTC_CENTER) : undefined;
 
-    return points;
+    return {
+        geometry,
+        offset,
+    };
```
gchoqueux, line 115: it seems there is no test on `PntsParser`

src/Provider/3dTilesProvider.js
```diff
@@ -130,8 +130,21 @@ function b3dmToMesh(data, layer, url) {
     });
 }
 
-function pntsParse(data) {
-    return PntsParser.parse(data).then(result => ({ object3d: result.point }));
+function pntsParse(data, layer) {
+    return PntsParser.parse(data).then((result) => {
+        const material = layer.material ?
```
gchoqueux, line 136: the overload of the material by the layer, is it documented?

src/Parser/PntsParser.js
peppsac: > it seems there is no test on PntsParser

Yes.

src/Provider/3dTilesProvider.js
peppsac: No it's not documented.

IMHO this part of iTowns is still moving, and should be properly documented when the refactoring regarding styling is done. WDYT?

src/Parser/PntsParser.js
gchoqueux: you could add it?

gchoqueux (changes requested):

src/Parser/PntsParser.js
peppsac: No, because that's not the point of this PR.

gchoqueux: if there's no test for this parser, it would take one
and more precisely for `pntsParse`

src/Provider/3dTilesProvider.js
gchoqueux: we must add a minimum of documentation
and why not just pass the material?

gchoqueux (dismissed):

src/Parser/PntsParser.js
peppsac: I added a test so we can hopefully move forward...

src/Provider/3dTilesProvider.js
peppsac: > we must add a minimum of documentation

Sure. But we currently we don't have any place to add this documentation. I might start working on something like this, but this PR is definitely not the place to start documenting layers.

> we must add a minimum of documentation

What do you mean?

test/.eslintrc.js
```diff
@@ -1,6 +1,9 @@
...
-  rules: {
-      'func-names': 'off',
-      'prefer-arrow-callback': 'off',
-  }
+    rules: {
+        'func-names': 'off',
+        'prefer-arrow-callback': 'off',
+    },
+    env: {
+        mocha: true,
```
zarov, line 7: :+1:

test/pnts_parser_unit_test.js
```diff
@@ -0,0 +1,46 @@
...
+                typed[next++] = parseInt(word, 16);
+            }
+        }
+    }
+    assert.equal(next, size);
+    return buffer;
+}
+
+describe('pnts parser', function () {
+    const pnts = `
```
zarov, line 18: Not a big fan of this, I saw some other methods (like base64 string) but all seem to not look good either, so leave it like this.

test/pnts_parser_unit_test.js
```diff
@@ -0,0 +1,46 @@
+import assert from 'assert';
+import PntsParser from '../src/Parser/PntsParser';
+
+function bufferFromString(pnts, size) {
+    const buffer = new ArrayBuffer(size);
+    const typed = new Uint8Array(buffer);
+    let next = 0;
+    for (const line of pnts.split('\n')) {
```
zarov: You could do one loop and remove the `if`: `for (const word of pnts.split(/\s+/))`

zarov (approved):
This seems good to me, do the changes in the test if you have time. Don't forget to squash too.

test/pnts_parser_unit_test.js
peppsac: Done thanks.

PR #788, opened 2018-06-25
fix(3dtiles): apply matrixWorld to boundingVolume.box

peppsac:
Before this commit we only used the translation part of the transform.
We're now using the full 4x4 matrix to match the wording of the 3dtiles
specification:
> The transform property applies to [...] tile.boundingVolume , except
when tile.boundingVolume.region is defined

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -250,22 +250,26 @@ export function pre3dTilesUpdate(context, layer) {
     return [layer.root];
 }
 
-const cameraLocalPosition = new THREE.Vector3();
-const worldPosition = new THREE.Vector3();
-function computeNodeSSE(camera, node) {
+const boundingVolumeBox = new THREE.Box3();
+const boundingVolumeSphere = new THREE.Sphere();
+export function computeNodeSSE(camera, node) {
     node.distance = 0;
     if (node.boundingVolume.region) {
-        worldPosition.setFromMatrixPosition(node.boundingVolume.region.matrixWorld);
-        cameraLocalPosition.copy(camera.camera3D.position).sub(worldPosition);
-        node.distance = node.boundingVolume.region.box3D.distanceToPoint(cameraLocalPosition);
+        boundingVolumeBox.copy(node.boundingVolume.region.box3D);
+        boundingVolumeBox.applyMatrix4(node.boundingVolume.region.matrixWorld);
+        node.distance = boundingVolumeBox.distanceToPoint(camera.camera3D.position);
     } else if (node.boundingVolume.box) {
-        worldPosition.setFromMatrixPosition(node.matrixWorld);
-        cameraLocalPosition.copy(camera.camera3D.position).sub(worldPosition);
-        node.distance = node.boundingVolume.box.distanceToPoint(cameraLocalPosition);
+        // boundingVolume.box is affected by matrixWorld
+        boundingVolumeBox.copy(node.boundingVolume.box);
+        boundingVolumeBox.applyMatrix4(node.matrixWorld);
+        node.distance = boundingVolumeBox.distanceToPoint(camera.camera3D.position);
     } else if (node.boundingVolume.sphere) {
-        worldPosition.setFromMatrixPosition(node.matrixWorld);
-        cameraLocalPosition.copy(camera.camera3D.position).sub(worldPosition);
-        node.distance = Math.max(0.0, node.boundingVolume.sphere.distanceToPoint(cameraLocalPosition));
+        // boundingVolume.sphere is affected by matrixWorld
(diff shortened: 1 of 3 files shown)


peppsac:
Updated with some unittest. Note that something is broken with the distance computation when using region but I couldn't fix it.

gchoqueux:
> Updated with some unittest. Note that something is broken with the distance computation when using region but I couldn't fix it.

an issue should be opened to specify the problem

peppsac:
> an issue should be opened to specify the problem

I've added 2 unit tests that need to be fixed. That's visible everytime a developer runs the tests so it's good enough:

>  50 passing (93ms)
>  2 pending

gchoqueux (changes requested):
this bug statement is clearly not visible, the best location for declaring and looking for bugs is the issues

peppsac:
> this bug statement is clearly not visible, the best location for declaring and looking for bugs is the issues

Done: https://github.com/iTowns/itowns/issues/796

peppsac:
@gchoqueux anything else bothering you on this PR or can we move forward and merge it?

src/Process/3dTilesProcessing.js
```diff
@@ -222,20 +222,25 @@ export function pre3dTilesUpdate(context, layer) {
 
 const cameraLocalPosition = new THREE.Vector3();
 const worldPosition = new THREE.Vector3();
-function computeNodeSSE(camera, node) {
+const boundingVolumeBox = new THREE.Box3();
+const boundingVolumeSphere = new THREE.Sphere();
+export function computeNodeSSE(camera, node) {
     node.distance = 0;
     if (node.boundingVolume.region) {
         worldPosition.setFromMatrixPosition(node.boundingVolume.region.matrixWorld);
```
gchoqueux: the test works with 
```js
        boundingVolumeBox.copy(node.boundingVolume.region.box3D);
        boundingVolumeBox.applyMatrix4(node.boundingVolume.region.matrixWorld);
        node.distance = boundingVolumeBox.distanceToPoint(camera.camera3D.position);
  }
```

gchoqueux (changes requested):

src/Process/3dTilesProcessing.js
```diff
@@ -222,20 +222,25 @@ export function pre3dTilesUpdate(context, layer) {
...
+        node.distance = boundingVolumeBox.distanceToPoint(camera.camera3D.position);
     } else if (node.boundingVolume.sphere) {
-        worldPosition.setFromMatrixPosition(node.matrixWorld);
-        cameraLocalPosition.copy(camera.camera3D.position).sub(worldPosition);
-        node.distance = Math.max(0.0, node.boundingVolume.sphere.distanceToPoint(cameraLocalPosition));
+        // boundingVolume.sphere is affected by matrixWorld
+        boundingVolumeSphere.copy(node.boundingVolume.sphere);
+        boundingVolumeSphere.applyMatrix4(node.matrixWorld);
+        node.distance = Math.max(0.0,
+            boundingVolumeSphere.distanceToPoint(camera.camera3D.position));
```
mbredif: Relying on `camera.camera3D.position` assumes that the camera's `matrix` is equal to its `matrixWorld` (as when the camera is a direct child of the scene). I would rather not have this assumption but if we have, let's write it somewhere.

To be conservative, but correct, I think you should apply `node.matrixWorld * camera.matrixWorldInverse` instead and measure the distance to the origin.

src/Process/3dTilesProcessing.js
peppsac: Fixed and pushed. Thanks

src/Process/3dTilesProcessing.js
peppsac: Agreed. As there are other places that make the same assumption I suggest to not fix  this in this PR.

I can add a // TODO to not forget about this?

gchoqueux: it would seem wise:
* to create an issue, to specify the problem and indicate where are the other places that make the same assumption (because apparently, this can be a source of bugs)
* apply the solution here and propose it also in the same issue

peppsac: > apply the solution here and propose it also in the same issue

It's not free performance-wise so I won't sneak this small change in an unrelated PR.

I've opened the following issue to not forget about it: https://github.com/iTowns/itowns/issues/800

peppsac:
Squashed and ready to merge.

src/Process/3dTilesProcessing.js
gchoqueux: > I can add a // TODO to not forget about this?

just add the TODO

gchoqueux (approved):

PR #789, opened 2018-06-25
feat(core): allow layer to specify which elements should be updatable

peppsac:
Some layers use separated objects for metadata and drawable objects.
For instance PointCloudProvider or 3dTilesProvider.

Attached layers (e.g texture layers) expect to receive a drawable object,
in their update function.

So this commit adds an optional metadataToElements function to layers,
allowing them to returns the to-be-updated object from metadata, and
its parent.

### DIFF
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -65,10 +65,37 @@ function updateElements(context, geometryLayer, elements) {
         // and then update Debug.js:addGeometryLayerDebugFeatures
         const newElementsToUpdate = geometryLayer.update(context, geometryLayer, element);
 
-        // update attached layers
-        for (const attachedLayer of geometryLayer._attachedLayers) {
-            if (attachedLayer.ready) {
-                attachedLayer.update(context, attachedLayer, element);
+        const sub = geometryLayer.getObjectToUpdateForAttachedLayers(element);
+
+        if (sub) {
+            if (sub.element) {
+                if (__DEBUG__) {
+                    if (!(sub.element.isObject3D)) {
+                        throw new Error(`
+                            Invalid object for attached layer to update.
+                            Must be a THREE.Object and have a THREE.Material`);
+                    }
+                }
+                // update attached layers
+                for (const attachedLayer of geometryLayer._attachedLayers) {
+                    if (attachedLayer.ready) {
+                        attachedLayer.update(context, attachedLayer, sub.element, sub.parent);
+                    }
+                }
+            } else if (sub.elements) {
+                for (let i = 0; i < sub.elements.length; i++) {
+                    if (!(sub.elements[i].isObject3D)) {
+                        throw new Error(`
+                            Invalid object for attached layer to update.
+                            Must be a THREE.Object and have a THREE.Material`);
+                    }
+                    // update attached layers
+                    for (const attachedLayer of geometryLayer._attachedLayers) {
+                        if (attachedLayer.ready) {
+                            attachedLayer.update(context, attachedLayer, sub.elements[i], sub.parent);
+                        }
+                    }
+                }
             }
         }
         updateElements(context, geometryLayer, newElementsToUpdate);
--- a/src/Provider/3dTilesProvider.js
+++ b/src/Provider/3dTilesProvider.js
@@ -8,7 +8,7 @@ import { init3dTilesLayer } from '../Process/3dTilesProcessing';
 import utf8Decoder from '../utils/Utf8Decoder';
 
 export function $3dTilesIndex(tileset, baseURL) {
-    let counter = 0;
+    let counter = 1;
     this.index = {};
     const inverseTileTransform = new THREE.Matrix4();
     const recurse = function recurse_f(node, baseURL, parent) {
@@ -60,9 +60,33 @@ export function $3dTilesIndex(tileset, baseURL) {
     };
 }
 
+export function getObjectToUpdateForAttachedLayers(meta) {
+    if (meta.content) {
+        const result = [];
+        meta.content.traverse((obj) => {
+            if (obj.isObject3D && obj.material && obj.layer == meta.layer) {
+                result.push(obj);
+            }
+        });
+        const p = meta.parent;
+        if (p && p.content) {
+            return {
+                elements: result,
+                parent: p.content,
+            };
+        } else {
+            return {
+                elements: result,
+            };
+        }
+    }
+}
+
 function preprocessDataLayer(layer, view, scheduler) {
     layer.sseThreshold = layer.sseThreshold || 16;
     layer.cleanupDelay = layer.cleanupDelay || 1000;
+    // override the default method, since updated objects are metadata in this case
+    layer.getObjectToUpdateForAttachedLayers = getObjectToUpdateForAttachedLayers;
 
     layer._cleanableTiles = [];
     return Fetcher.json(layer.url, layer.networkOptions).then((tileset) => {
@@ -180,6 +204,8 @@ function executeCommand(command) {
 
     const setLayer = (obj) => {
         obj.layers.set(layer.threejsLayer);
+        obj.userData.metadata = metadata;
+        obj.layer = layer;
     };
     if (path) {
         // Check if we have relative or absolute url (with tileset's lopocs for example)
--- a/src/Provider/PointCloudProvider.js
+++ b/src/Provider/PointCloudProvider.js
@@ -93,6 +93,7 @@ function parseOctree(layer, hierarchyStepSize, root) {
                         baseurl: url,
                         bbox: bounds,
                         layer,
+                        parent: snode,
                     };
                     snode.children.push(item);
                     stack.push(item);
@@ -196,6 +197,22 @@ function parseMetadata(metadata, layer) {
     layer.supportsProgressiveDisplay = customBinFormat;
 }
 
+export function getObjectToUpdateForAttachedLayers(meta) {
+    if (meta.obj) {
+        const p = meta.parent;
+        if (p && p.obj) {
+            return {
+                element: meta.obj,
+                parent: p.obj,
+            };
+        } else {
+            return {
+                element: meta.obj,
+            };
+        }
+    }
+}
+
 export default {
     preprocessDataLayer(layer, view) {
         if (!layer.file) {
@@ -231,6 +248,9 @@ export default {
         layer.update = PointCloudProcessing.update;
         layer.postUpdate = PointCloudProcessing.postUpdate;
 
+        // override the default method, since updated objects are metadata in this case
+        layer.getObjectToUpdateForAttachedLayers = getObjectToUpdateForAttachedLayers;
+
         // this probably needs to be moved to somewhere else
         layer.pickObjectsAt = (view, mouse, radius) => Picking.pickPointsAt(view, mouse, radius, layer);
 
@@ -253,31 +273,32 @@ export default {
 
     executeCommand(command) {
         const layer = command.layer;
-        const node = command.requester;
+        const metadata = command.requester;
 
         // Query HRC if we don't have children metadata yet.
-        if (node.childrenBitField && node.children.length === 0) {
-            parseOctree(layer, layer.metadata.hierarchyStepSize, node).then(() => command.view.notifyChange(layer, false));
+        if (metadata.childrenBitField && metadata.children.length === 0) {
+            parseOctree(layer, layer.metadata.hierarchyStepSize, metadata).then(() => command.view.notifyChange(layer, false));
         }
 
         // `isLeaf` is for lopocs and allows the pointcloud server to consider that the current
         // node is the last one, even if we could subdivide even further.
         // It's necessary because lopocs doens't know about the hierarchy (it generates it on the fly
         // when we request .hrc files)
-        const url = `${node.baseurl}/r${node.name}.${layer.extension}?isleaf=${command.isLeaf ? 1 : 0}`;
+        const url = `${metadata.baseurl}/r${metadata.name}.${layer.extension}?isleaf=${command.isLeaf ? 1 : 0}`;
 
         return Fetcher.arrayBuffer(url, layer.fetchOptions).then(buffer => layer.parse(buffer, layer.metadata.pointAttributes)).then((geometry) => {
             const points = new THREE.Points(geometry, layer.material.clone());
             addPickingAttribute(points);
             points.frustumCulled = false;
             points.matrixAutoUpdate = false;
-            points.position.copy(node.bbox.min);
+            points.position.copy(metadata.bbox.min);
             points.scale.set(layer.metadata.scale, layer.metadata.scale, layer.metadata.scale);
             points.updateMatrix();
             points.tightbbox = geometry.boundingBox.applyMatrix4(points.matrix);
             points.layers.set(layer.threejsLayer);
             points.layer = layer;
-            points.extent = Extent.fromBox3(command.view.referenceCrs, node.bbox);
+            points.extent = Extent.fromBox3(command.view.referenceCrs, metadata.bbox);
+            points.userData.metadata = metadata;
             return points;
         });
     },
--- a/src/Core/Layer/Layer.js
+++ b/src/Core/Layer/Layer.js
@@ -80,6 +80,21 @@ function GeometryLayer(id, object3d) {
     // Setup default picking method
     this.pickObjectsAt = (view, mouse, radius) => Picking.pickObjectsAt(view, mouse, radius, this.object3d);
 
+    // Attached layers expect to receive the visual representation of a layer (= THREE object with a material).
+    // So if a layer's update function don't process this kind of object, the layer must provide
+    // a getObjectToUpdateForAttachedLayers function that returns the correct object to update for attached
+    // layer.
+    // See 3dtilesProvider or PointCloudProvider for examples.
+    // eslint-disable-next-line arrow-body-style
+    this.getObjectToUpdateForAttachedLayers = (obj) => {
+        if (obj.parent && obj.material) {
+            return {
+                element: obj,
+                parent: obj.parent,
+            };
+        }
+    };
+
     this.postUpdate = () => {};
 }
 
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -16,6 +16,11 @@ function requestNewTile(view, scheduler, geometryLayer, metadata, parent, redraw
     return scheduler.execute(command);
 }
 
+function getChildTiles(tile) {
+    // only keep children that have the same layer and a valid tileId
+    return tile.children.filter(n => n.layer == tile.layer && n.tileId);
+}
+
 function subdivideNode(context, layer, node, cullingTest) {
     if (node.additiveRefinement) {
         // Additive refinement can only fetch visible children.
@@ -88,7 +93,7 @@ function _subdivideNodeAdditive(context, layer, node, cullingTest) {
 }
 
 function _subdivideNodeSubstractive(context, layer, node) {
-    if (!node.pendingSubdivision && node.children.filter(n => n.layer == layer).length == 0) {
+    if (!node.pendingSubdivision && getChildTiles(node).length == 0) {
         const childrenTiles = layer.tileIndex.index[node.tileId].children;
         if (childrenTiles === undefined || childrenTiles.length === 0) {
             return;
@@ -195,7 +200,7 @@ function cleanup3dTileset(layer, n, depth = 0) {
             n.parent.remove(n);
         }
     } else {
-        const tiles = n.children.filter(n => n.tileId != undefined);
+        const tiles = getChildTiles(n);
         n.remove(...tiles);
     }
 }
@@ -330,11 +335,11 @@ export function process3dTilesNode(cullingTest, subdivisionTest) {
                 subdivideNode(context, layer, node, cullingTest);
                 // display iff children aren't ready
                 setDisplayed(node, node.pendingSubdivision || node.additiveRefinement);
-                returnValue = node.children.filter(n => n.layer == layer);
+                returnValue = getChildTiles(node);
             } else {
                 setDisplayed(node, true);
 
-                for (const n of node.children.filter(n => n.layer == layer)) {
+                for (const n of getChildTiles(node)) {
                     n.visible = false;
                     markForDeletion(layer, n);
                 }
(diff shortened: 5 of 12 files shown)


zarov:
This needs a little more documentation, on what is expected to be returned from `metadataToElements`, in case someone writes this function as well. What about putting it in the documentation of `Layer`, or the `Provider` interface ?

src/Core/MainLoop.js
```diff
@@ -68,7 +68,19 @@ function updateElements(context, geometryLayer, elements) {
         // update attached layers
         for (const attachedLayer of geometryLayer._attachedLayers) {
             if (attachedLayer.ready) {
-                attachedLayer.update(context, attachedLayer, element);
+                // Attached layers expect to receive the visual representation
```
gchoqueux: An example would be needed to understand the use case.
it's necessary to add a new update?

src/Provider/3dTilesProvider.js
```diff
@@ -62,6 +62,21 @@ function $3dTilesIndex(tileset, baseURL) {
 function preprocessDataLayer(layer, view, scheduler) {
     layer.sseThreshold = layer.sseThreshold || 16;
     layer.cleanupDelay = layer.cleanupDelay || 1000;
+    layer.metadataToElements = (meta) => {
```
gchoqueux: `metadataToElements`  shouldn't be exposed in layer. We must avoid putting processing in the provider (especially that the providers should disappear)

why not do it in the  `attachedLayer.update`?

src/Core/MainLoop.js
peppsac: You can find an example in the https://github.com/peppsac/itowns2/tree/colorisation branch.

It uses the ColorTextureProcessing to colorize a 3dtiles pointcloud.

But 3dTilesProcessing update function process metadata objects (= the content of the tileset.json file) while ColorTextureProcessing expects objects with a material (because it was initially built for TileMesh).

src/Provider/3dTilesProvider.js
peppsac: > metadataToElements shouldn't be exposed in layer
> We must avoid putting processing in the provider

Why?

>  (especially that the providers should disappear)

I don't think so. While currently their scope/responsability is unclear, I don't think they'll ever disappear. For instance, in the 3dtiles case you'll always need "something" that understands the tileset.json structure and parse/prepare it for the processing code to operate on.

The purpose of this `metadataToElements` function is to allow Processing code to be independant of the providers.

peppsac: We already had the same discussion for the picking feature where no good conclusion was reached. 

This confirms that the Provider/Processing boundary isn't good enough, but IMHO this shouldn't prevent merging this.

peppsac:
@zarov I shuffled the code a bit and moved the comment to `Layer.js` + added an error in MainLoop. WDYT?

src/Core/MainLoop.js
```diff
@@ -68,17 +68,16 @@ function updateElements(context, geometryLayer, elements) {
...
-                // transmit to its attached layer.
-                const sub =
-                    geometryLayer.metadataToElements ?
-                        geometryLayer.metadataToElements(element) :
-                        { element, parent: element.parent };
+                const sub = geometryLayer.getObjectToUpdateForAttachedLayers(element);
+
                 if (sub) {
+                    if (__DEBUG__) {
+                        if (!(sub.isObject && sub.material)) {
```
zarov: Are you sure about this ? It will always throw an error here.

peppsac: No the correct code is `if (!(sub.element.isObject3D && sub.element.material)) {`.

I'll push it, but first I need to resolve a bug that this assert shows.

peppsac: Fixed.

src/Core/MainLoop.js
```diff
@@ -68,7 +68,18 @@ function updateElements(context, geometryLayer, elements) {
         // update attached layers
         for (const attachedLayer of geometryLayer._attachedLayers) {
             if (attachedLayer.ready) {
-                attachedLayer.update(context, attachedLayer, element);
+                const sub = geometryLayer.getObjectToUpdateForAttachedLayers(element);
+
+                if (sub) {
+                    if (__DEBUG__) {
+                        if (!(sub.element.isObject3D && sub.element.material)) {
```
zarov: Some objects still go through this condition, for example the `Scene` of iTowns. Thus on the `3dtiles` example I can't navigate and move at all.

peppsac: It may actually not really a bug.

Do we really want to pass all the root of 3dtiles scene to attached layer? (in which case the attached layer needs to be aware that it may receive a hierarchy, rather than a single object). Or do we want to pass the drawable elements? (in this case the code in getObjectToUpdateForAttachedLayers should collect all objects in `content` that have a material).

I think I prefer the second option. WDYT?

peppsac: Since I'm not sure about this, I removed the `sub.element.material` check for now.

src/Core/MainLoop.js
```diff
@@ -68,7 +68,18 @@ function updateElements(context, geometryLayer, elements) {
...
+                    if (__DEBUG__) {
+                        if (!(sub.element.isObject3D)) {
+                            throw new Error(`
+                                Invalid object for attached layer to update.
+                                Must be a THREE.Object and have a THREE.Material`);
+                        }
+                    }
+                    attachedLayer.update(context, attachedLayer, sub.element, sub.parent);
+                }
             }
```
mbredif: the above does not depend on attachedLayer and could be moved outside the loop

src/Core/MainLoop.js
```diff
@@ -68,7 +68,18 @@ function updateElements(context, geometryLayer, elements) {
...
+                const sub = geometryLayer.getObjectToUpdateForAttachedLayers(element);
+
+                if (sub) {
+                    if (__DEBUG__) {
+                        if (!(sub.element.isObject3D)) {
+                            throw new Error(`
+                                Invalid object for attached layer to update.
+                                Must be a THREE.Object and have a THREE.Material`);
+                        }
+                    }
```
mbredif: the above does not depend on attachedLayer and could be moved outside the loop

mbredif (commented):
When do you need the parent ?

src/Core/MainLoop.js
peppsac: I missed this, thanks.

peppsac:
> When do you need the parent ?

e.g: for initializing a texture layer from the parent's texture = every usage of node.parent in `LayeredMaterialNodeProcessing.js`.

src/Core/MainLoop.js
peppsac: Fixed.

src/Core/MainLoop.js
zarov: I'll go with the second option as well.

peppsac: I implemented it. Note that I had to adapt 3dTilesDebug to reflect this change - this part is a bit tedious due to the way we handle gltfUpAxis (= we rotate the matrix of the model but the boundingVolume must not be affected by this rotation).

gchoqueux (changes requested):
I'm currently reviewing this PR
but first you have to add some tests

src/Core/MainLoop.js
```diff
@@ -65,10 +65,37 @@ function updateElements(context, geometryLayer, elements) {
...
+                            Must be a THREE.Object and have a THREE.Material`);
+                    }
+                }
+                // update attached layers
+                for (const attachedLayer of geometryLayer._attachedLayers) {
+                    if (attachedLayer.ready) {
+                        attachedLayer.update(context, attachedLayer, sub.element, sub.parent);
+                    }
+                }
+            } else if (sub.elements) {
```
gchoqueux, line 85: `getObjectToUpdateForAttachedLayers` can always return an Array?
it would simplify this part

src/Provider/PointCloudProvider.js
```diff
@@ -174,6 +175,23 @@ export default {
         layer.update = PointCloudProcessing.update;
         layer.postUpdate = PointCloudProcessing.postUpdate;
 
+        // override the default method, since updated objects are metadata in this case
+        layer.getObjectToUpdateForAttachedLayers = (meta) => {
```
gchoqueux: It should move this function in the processing, and assign it as above (same think  for 3dtiles)
```js
layer.getObjectToUpdateForAttachedLayers = PointCloudProcessing.getObjectToUpdateForAttachedLayers;
```

the `provider.preprocess` it looks like a layer's constructor. `3dTiles` and `pointCloud` should inherit `geometryLayer` and overload its processing functions (if that is necessary)

gchoqueux (changes requested):

src/Provider/PointCloudProvider.js
peppsac: > It should move this function in the processing, and assign it as above

Where? Processing have no code that run once for a layer, while Provider have a preprocess step.
And processing code should try to be as independant as possible from providers so moving these functions that are highly dependent on the protocol/provider makes very little sense.

One of my mid-term goal is to have a single processing module for pointclouds, that would work with all kind of pointcloud sources.
To be able to do this, the provider of the said sources has to format/prepare the layer to meet the expectations of this processing module.

Basically: the processing code defines a contract, and will work with every layer that implements it.

> the provider.preprocess it looks like a layer's constructor 

Can you explain what you mean?

> 3dTiles and pointCloud should inherit geometryLayer and overload its processing functions (if that is necessary)

I assume that you meant 3dtiles *layer* and pointcloud *layer*?
What would be the benefits of inheriting + overriding `GeometryLayer` over the current situation?

(as there are no processing functions in GeometryLayer so I don't understand this either)

src/Core/MainLoop.js
peppsac: This part is internal only (= no user of iTowns will have to write the same code in its application), so I don't mind having a bit more complicated code if it performs better.

I'll check, and simplify if possible.

peppsac:
> I'm currently reviewing this PR

Ok, thanks.

>  but first you have to add some tests

This part is central in iTowns, so it's exercised by all functionnal tests.
I'll see if I can add some useful tests anyway - don't hesitate to suggest what kind of test you would like to have.

src/Core/MainLoop.js
gchoqueux: I feel that it is more of a problem of structure of our elements. We should not complicate this part of the process to be able to adapt to all structures. I'll go deeper and I'll come back on the subject

src/Provider/3dTilesProvider.js
```diff
@@ -62,6 +62,28 @@ function $3dTilesIndex(tileset, baseURL) {
 function preprocessDataLayer(layer, view, scheduler) {
     layer.sseThreshold = layer.sseThreshold || 16;
     layer.cleanupDelay = layer.cleanupDelay || 1000;
+    // override the default method, since updated objects are metadata in this case
+    layer.getObjectToUpdateForAttachedLayers = (meta) => {
```
gchoqueux: you should add a test  will verify the function work as expected
and others clarify / document the data structure used for 3dTiles (as for example, the `object`, the `userData` and the `content` are structured)

src/Provider/PointCloudProvider.js
gchoqueux: you should add a test will verify the function work as expected
and others clarify / document the data structure used for cloudPoint (as for example, what is the difference this structure `geometryLayer`)

gchoqueux: > as there are no processing functions in GeometryLayer so I don't understand this either

`geometryLayer.getObjectToUpdateForAttachedLayers` is processing function

gchoqueux (dismissed):

src/Provider/PointCloudProvider.js
peppsac: Ok I don't want to change this because I simply don't agree.

The problem currently is not this code, but the fact that Provider and Processing have strong coupling.

I and others are working on improving this and I would appreciate that you don't undermine this effort by bringing this topic on each review.

src/Provider/3dTilesProvider.js
peppsac: I reverted part of the userData change.

The point was to store our objects in this dict created by three.js instead of patching the constructed objects (mainly because it's a good pratice and it might improve performance).

See [this comment](https://github.com/iTowns/itowns/pull/719#issuecomment-380376078) for a more info.

peppsac:
Pushed a commit to address the reviews + added some tests.

peppsac:
@gchoqueux can you update your review please?

src/Provider/3dTilesProvider.js
gchoqueux: As presented in #798, this PR proposes, not to couple the provider and the processing, as it makes the architecture more rigid and less modular. We must try to place this code in the processing not to increase this coupling

src/Provider/3dTilesProvider.js
gchoqueux: OK fine,
but what is the structure of the data?

src/Provider/PointCloudProvider.js
gchoqueux: Test `getObjectToUpdateForAttachedLayers` is added, is it testing for `geometryLayer`?
What is data structure used for cloudPoint (as for example, what is the difference this structure `geometryLayer`)

src/Provider/3dTilesProvider.js
peppsac: You missed the point again and again.

I won't repeat myself please read my previous comments.

src/Provider/3dTilesProvider.js
peppsac: >  what is the structure of the data?

I don't understand your question sorry.

src/Provider/PointCloudProvider.js
peppsac: `getObjectToUpdateForAttachedLayers` always return the same thing, as it's a contract between MainLoop and Layer.

The description is here https://github.com/iTowns/itowns/pull/789/files#diff-61e012c3e00bdc510be33d9eac684b2aR83, and the comment / ref implementation explains the returned types (it's not using jsdoc since it's not exposed anyway).

src/Provider/3dTilesProvider.js
gchoqueux: Can you clarify the structure of objects used for 3dTiles, in itowns

src/Provider/PointCloudProvider.js
gchoqueux: I can not find an explanation of data structure used for cloudPoind,  in itowns

src/Provider/3dTilesProvider.js
peppsac: I didn't change any the structure or its behavior in this PR. Why are you requesting this?

(also the structure is exactly the one from the 3dtiles spec)

src/Provider/PointCloudProvider.js
peppsac: See https://github.com/iTowns/itowns/pull/789#discussion_r200361337

gchoqueux:
for this PR:

* What is the structure of instanced objects (THREE.js and itowns) and not the specification. 
for example in node tile

```js
tileA  // (THREE.Mesh->itowns.tileMesh)
     {
          children: [tileB, tileC] // (Array) tile children
          parent: tile0  // (THREE.Mesh->itowns.tileMesh) tile parent 
          obb: // (itowns.OBB)  oriented bouding box
     }
```
in `3dTiles`, how are the objects linked: `content, parent, bounding Volume, region, transform ...`
in `pointCloud`, how are the objects linked: `parent, obj, tightbbox  ...`

This information is important to understand the `MainLoop.update` and `getObjectToUpdateForAttachedLayers` (and for the maintenance of the code)

----------------

 * for getObjectToUpdateForAttachedLayers

As seen in the PR #798 , it is an error to place processing functions in the provider, to go in this direction we have to strive to place in the processing like `process3dTilesNode` or `pointCloudProcessing.update`

peppsac:
> As seen in the PR #798 , it is an error to place processing functions in the provider, to go in this direction we have to strive to place in the processing like process3dTilesNode or pointCloudProcessing.update

I disagree for the 1000th times. You got the logic backward: processing code should only depend on the data type (points or tiles or whatever). And 3dTilesProcessing should disappear, not 3dTilesProvider.

To achieve this, the Provider must hide away the specific of each protocols. And  this is exactly  the purpose of getObjectToUpdateForAttachedLayers.

With this logic, I'm able to implement pointcloud coloring using a WMS server (aka = I have a branch where it works). And the pointcloud texture selection uses the *same* code than the tile texturing code.

Do you have something similar with your branch? If not, then maybe you could let this PR in, so I can submit the next PR that depends on it. And then you might realize that you're wrong.

src/Core/MainLoop.js
```diff
@@ -65,10 +65,37 @@ function updateElements(context, geometryLayer, elements) {
...
+                for (let i = 0; i < sub.elements.length; i++) {
```
zarov, line 86: Use `for (let element of sub.elements)` instead.

src/Core/MainLoop.js
zarov: I agree with @gchoqueux, even if it is hidden, it should always return `elements` as an array.

zarov (commented):
> I disagree for the 1000th times. You got the logic backward: processing code should only depend on the data type (points or tiles or whatever). And 3dTilesProcessing should disappear, not 3dTilesProvider.

And I disagree too. Both should disappear, data type has nothing to do with the provider. It makes more sense to me to add this processing code in `3dTilesProcessing`, which is more relevant to the 3dtiles format and data type.
Anyway, what @gchoqueux is saying (if I understand correctly) is that we should avoid adding methods like this if both are meant to disappear in the future. 

>  gchoqueux 7 days ago Member
Can you clarify the structure of objects used for 3dTiles, in itowns
@peppsac
peppsac 5 days ago Member
I didn't change any the structure or its behavior in this PR. Why are you requesting this?
(also the structure is exactly the one from the 3dtiles spec)

Concerning this, I don't think the documentation of data structure is relevant to this PR, even if it could be good to have it.

src/Core/MainLoop.js
peppsac: Keep in mind that this piece of code is executed multiple times during each update loop.

So creating an array by default, even if it's for a single element might pressure the GC or cause a perf hit.

peppsac:
> And I disagree too. Both should disappear, data type has nothing to do with the provider. It makes more sense to me to add this processing code in 3dTilesProcessing, which is more relevant to the 3dtiles format and data type.

The function here has no expectation on the data type. It only depends on the 3dtiles protocol (metadata would be more accurate).

To be clear: **3dtiles is not a data type, it's a protocol** (in itowns current semantics at least which mix protocol and metadata). 

So the provider is aware of the metadata, and build a data structure understood by itowns / the processing functions. This structure will evolve, but the point of this PR (and the next one) is to iterate on its definition (at the very least: `getObjectToUpdateForAttachedLayers` and `.extent` are part of it).

In the long term: the 3dtilesprocessing module should be gone, and be replaced either by a single processing.js module (see #674  for something that goes in this direction) or several datatypeprocessing.js (pointcloudprocessing.js, tileprocessing.js ...).

But the 3dtilesprovider must remain (even if its renamed 3dtilessource): because we need a place in itowns that is aware of the protocol/metadata of each supported sources.

Is it clearer?

src/Core/MainLoop.js
peppsac: I dislike more and more the `for of ` syntax because it's harder to debug (lots of indirection). I'm also wondering on the perf cost of this (for the same reason: lots of indirection)...

Anyway I'll change it because for now we settled on using it everywhere.

zarov: Yeah it seems that performance wise it is the slowest (as tested quickly here http://jsben.ch/iEYTg)

src/Core/MainLoop.js
zarov: Ok, that's a good point. We still have duplicated code though, could you move both part into another function ?

zarov:
It is clearer thanks, I was confused with the fact that 3dtiles is a protocol indeed. Could we have a glimpse of the next PR ? 

Otherwise it seems good now to me, maybe @mbredif can give his opinion too ? I'm not the most aware contributor on 3dtiles.

src/Core/MainLoop.js
peppsac: And this test misses one of the cost: in itowns we never run native `for of` loops, but only transpiled `for of`.
For instance a simple:
```js
for (const word of words) {
  console.log(word);
}
```
becomes:
```js
"use strict";

var _iteratorNormalCompletion = true;
var _didIteratorError = false;
var _iteratorError = undefined;

try {
  for (var _iterator = words[Symbol.iterator](), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
    var word = _step.value;

    console.log(word);
  }
} catch (err) {
  _didIteratorError = true;
  _iteratorError = err;
} finally {
  try {
    if (!_iteratorNormalCompletion && _iterator.return) {
      _iterator.return();
    }
  } finally {
    if (_didIteratorError) {
      throw _iteratorError;
    }
  }
}
```
([see it live](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFAxiB2BnALhA7uAJhEx1cQEoIBvAKAgjiRABsBTAOlpAHMoMxNCBuMgXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=))

peppsac:
> Could we have a glimpse of the next PR ?

Already posted elsewhere: https://github.com/peppsac/itowns2/commits/colorisation

My recent PRs are a split of the work done there. At the end the point is to be able to reuse the 'texture update processing' code to colorize pointclouds.

(and I have another feature built on top of this branch that leverage the cache to reduce the latency in itowns... but this will have to wait until all the previous PRs are merged)

zarov:
Thanks ! I only took a quick look but I think there are some things I might disagree on later ;) 

Anyway, could you rebase this PR ?

peppsac:
> Thanks ! I only took a quick look but I think there are some things I might disagree on later ;)

Sure :)

> Anyway, could you rebase this PR ?

Done

zarov (approved):
Okay let's merge this, rebase and test before ! Thanks

PR #790, opened 2018-06-26
chore(travis): remove the manual installation of puppeteer for travis

zarov:
Since the puppeteer dependency is present in `package.json`, we don't have to install it manually in travis. I had to relaunch a build three times before it got to pass because it hanged while installing puppeteer.

The presence of `package-lock.json` in this PR is because Travis needs it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: node_js
 node_js:
   - "lts/*"
   - "node"
-before_install:
-  - npm install puppeteer
 after_success:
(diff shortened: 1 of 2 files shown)


peppsac:
> The presence of package-lock.json in this PR is because Travis needs it.

I don't understand this comment, can you explain?

zarov:
It seems that Travis looks after the package-lock to determine what to install, and we didn't add the puppeteer dependency in it, so I updated it here. Otherwise it was causing an error.

peppsac (approved):
Ok, looks good then.

PR #791, opened 2018-06-26
fix(core): force the use of decimal notation in protocol requests

peppsac:
OGC services don't expect scientific notation in request so explicitely use .toFixed when building the URL.

### DIFF
--- a/src/Provider/URLBuilder.js
+++ b/src/Provider/URLBuilder.js
@@ -70,11 +70,12 @@ export default {
      * @return {string} the formed url
      */
     bbox: function bbox(bbox, layer) {
+        const precision = layer.projection == 'EPSG:4326' ? 9 : 2;
         const box = bbox.as(layer.projection);
-        const w = box.west();
-        const s = box.south();
-        const e = box.east();
-        const n = box.north();
+        const w = box.west().toFixed(precision);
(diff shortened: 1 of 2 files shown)


zarov (approved):
LGTM

src/Provider/URLBuilder.js
```diff
@@ -71,10 +71,10 @@ export default {
      */
     bbox: function bbox(bbox, layer) {
         const box = bbox.as(layer.projection);
-        const w = box.west();
-        const s = box.south();
-        const e = box.east();
-        const n = box.north();
+        const w = box.west().toFixed(5);
```
gchoqueux: I'm not sure there's enough number after the comma

peppsac: For layers using metres this won't be a problem (and we could reduce to `2`).

For 4326, `5` might be a bit low, so I can increase it to 9.

peppsac: I've updated the branch.

gchoqueux: > For 4326, 5 might be a bit low, so I can increase it to 9.

yes, `1e-9` is enough because the projection on the globe is less than 1 cm

PR #792, opened 2018-06-26
Refacto limit dependency on layeredmaterial

peppsac:
2  commits, related to the same context: stop leaking internal attributes (WMTS level, LayeredMateiral texture indexing schema) in all iTowns.

First commit content:

    fix: stop using 'level' property where it's not mandatory
    
    Currently the 'level' concept from WMTS has leaked everywhere in iTowns
    codebase.
    
    This commits replace its usage in one part of iTowns, by something
    equivalent but that doesn't force texture Providers to mock WMTS level
    mechanism.

2nd one is:
    refacto(core): avoid manipulating LayeredMaterial internal state
    
    This commits is a step forward stopping to rely on LayeredMaterial
    internal representation of textures (using an index and such) in other
    parts of iTowns.
    
    This will allow to reuse all this code for objects that don't use a
    LayerMaterial instance as their material.

### DIFF
--- a/examples/split.html
+++ b/examples/split.html
@@ -86,17 +86,12 @@
 
             function changeLayerVisibility(ortho, osm) {
                 var material;
-                var orthoIndex;
-                var opensmIndex;
 
                 globeView.scene.traverse(function _(obj) {
                     if (obj.material && obj.material.setLayerVisibility && obj.material.visible) {
                         material = obj.material;
-                        orthoIndex = material.indexOfColorLayer(orthoLayer.id);
-                        opensmIndex = material.indexOfColorLayer(osmLayer.id);
-
-                        material.setLayerVisibility(orthoIndex, ortho);
-                        material.setLayerVisibility(opensmIndex, osm);
+                        material.setLayerVisibility(orthoLayer, ortho);
+                        material.setLayerVisibility(osmLayer, osm);
                     }
                 });
             }
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -5,25 +5,21 @@
  */
 
 import * as THREE from 'three';
-import LayeredMaterial from '../Renderer/LayeredMaterial';
-import { l_ELEVATION } from '../Renderer/LayeredMaterialConstants';
 import RendererConstant from '../Renderer/RendererConstant';
 import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from '../Provider/OGCWebServiceHelper';
 import { is4326 } from './Geographic/Coordinates';
 
-function TileMesh(geometry, params) {
+function TileMesh(layer, geometry, material, extent, level) {
     // Constructor
-    THREE.Mesh.call(this);
+    THREE.Mesh.call(this, geometry, material);
+
+    this.layer = layer;
 
     this.matrixAutoUpdate = false;
     this.rotationAutoUpdate = false;
 
-    if (!params.extent) {
-        throw new Error('params.extent is mandatory to build a TileMesh');
-    }
-
-    this.level = params.level;
-    this.extent = params.extent;
+    this.level = level;
+    this.extent = extent;
 
     this.geometry = geometry;
 
@@ -32,11 +28,10 @@ function TileMesh(geometry, params) {
     this.boundingSphere = new THREE.Sphere();
     this.OBB().box3D.getBoundingSphere(this.boundingSphere);
 
-    this.material = new LayeredMaterial(params.materialOptions);
-
     this.frustumCulled = false;
 
     this.updateGeometricError();
+    this.wmtsCoords = {};
 
     // Layer
     this.setDisplayed(false);
@@ -120,15 +115,12 @@ TileMesh.prototype.setSelected = function setSelected(select) {
     this.material.setSelected(select);
 };
 
-TileMesh.prototype.setTextureElevation = function setTextureElevation(elevation) {
+TileMesh.prototype.setTextureElevation = function setTextureElevation(layer, elevation) {
     if (this.material === null) {
         return;
     }
-
-    const offsetScale = elevation.pitch || new THREE.Vector4(0, 0, 1, 1);
     this.setBBoxZ(elevation.min, elevation.max);
-
-    this.material.setTexture(elevation.texture, l_ELEVATION, 0, offsetScale);
+    this.material.setLayerTextures(layer, elevation);
 };
 
 
@@ -158,31 +150,8 @@ TileMesh.prototype.setTexturesLayer = function setTexturesLayer(textures, layerT
     }
 };
 
-TileMesh.prototype.getLayerTextures = function getLayerTextures(layerType, layerId) {
-    const mat = this.material;
-    return mat.getLayerTextures(layerType, layerId);
-};
-
-TileMesh.prototype.isColorLayerLoaded = function isColorLayerLoaded(layerId) {
-    const mat = this.material;
-    return mat.getColorLayerLevelById(layerId) > -1;
-};
-
-TileMesh.prototype.isElevationLayerLoaded = function isElevationLayerLoaded() {
-    return this.material.loadedTexturesCount[l_ELEVATION] > 0;
-};
-
-TileMesh.prototype.isColorLayerDownscaled = function isColorLayerDownscaled(layer) {
-    const mat = this.material;
-    return mat.isColorLayerDownscaled(layer.id, this.getZoomForLayer(layer));
-};
-
 TileMesh.prototype.OBB = function OBB() {
     return this.obb;
-};
-
-TileMesh.prototype.getIndexLayerColor = function getIndexLayerColor(idLayer) {
-    return this.material.indexOfColorLayer(idLayer);
 };
 
 TileMesh.prototype.removeColorLayer = function removeColorLayer(idLayer) {
@@ -224,8 +193,13 @@ TileMesh.prototype.getCoordsForLayer = function getCoordsForLayer(layer) {
         } else {
             return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent, layer.origin);
         }
-    } else {
+    } else if (layer.extent.crs() == this.extent.crs()) {
+        // Currently extent.as() always clone the extent, even if the output
+        // crs is the same.
+        // So we avoid using it if both crs are the same.
         return [this.extent];
+    } else {
+        return [this.extent.as(layer.extent.crs())];
     }
 };
 
--- a/src/Process/GlobeTileProcessing.js
+++ b/src/Process/GlobeTileProcessing.js
@@ -2,6 +2,7 @@ import * as THREE from 'three';
 import { ellipsoidSizes } from '../Core/Geographic/Coordinates';
 import { SIZE_TEXTURE_TILE } from '../Provider/OGCWebServiceHelper';
 import Extent from '../Core/Geographic/Extent';
+import { l_ELEVATION } from '../Renderer/LayeredMaterialConstants';
 
 const cV = new THREE.Vector3();
 let vhMagnitudeSquared;
@@ -125,11 +126,14 @@ export function globeSubdivisionControl(minLevel, maxLevel, sseThreshold, maxDel
         // Prevent to subdivise the node if the current elevation level
         // we must avoid a tile, with level 20, inherits a level 3 elevation texture.
         // The induced geometric error is much too large and distorts the SSE
-        const currentElevationLevel = node.material.getElevationLayerLevel();
-        if (node.level < context.maxElevationLevel + maxDeltaElevationLevel &&
-            currentElevationLevel >= 0 &&
-            (node.level - currentElevationLevel) >= maxDeltaElevationLevel) {
-            return false;
+        const currentTexture = node.material.textures[l_ELEVATION][0];
+        if (currentTexture.extent) {
+            const offsetScale = node.material.offsetScale[l_ELEVATION][0];
+            const ratio = offsetScale.z;
+            // ratio is node size / texture size
+            if (ratio < 1 / Math.pow(2, maxDeltaElevationLevel)) {
+                return false;
+            }
         }
 
         const sse = computeNodeSSE(context.camera, node);
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -16,7 +16,7 @@ function initNodeImageryTexturesFromParent(node, parent, layer) {
 
         let textureIndex = offsetTextures;
         for (const c of coords) {
-            for (const texture of parent.material.getLayerTextures(l_COLOR, layer.id)) {
+            for (const texture of parent.material.getLayerTextures(layer)) {
                 if (c.isInside(texture.coords)) {
                     const result = c.offsetToParent(texture.coords);
                     node.material.textures[l_COLOR][textureIndex] = texture;
@@ -66,9 +66,7 @@ function initNodeElevationTextureFromParent(node, parent, layer) {
             elevation.max = max;
         }
 
-        node.setTextureElevation(elevation);
-        node.material.elevationLayersId =
-            parent.material.elevationLayersId;
+        node.setTextureElevation(layer, elevation);
     }
 }
 
@@ -164,19 +162,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
         }
 
         if (material.indexOfColorLayer(layer.id) === -1) {
-            const texturesCount =
-                node.getCoordsForLayer(layer).length;
-
-            const paramMaterial = {
-                tileMT: layer.options.tileMatrixSet || node.getCoordsForLayer(layer)[0].crs(),
-                texturesCount,
-                visible: layer.visible,
-                opacity: layer.opacity,
-                fx: layer.fx,
-                idLayer: layer.id,
-            };
-
-            material.pushLayer(paramMaterial);
+            material.pushLayer(layer, node.getCoordsForLayer(layer));
             const imageryLayers = context.view.getLayers(l => l.type === 'color');
             const sequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
             material.setSequence(sequence);
@@ -205,9 +191,8 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
     // TODO: move this to defineLayerProperty() declaration
     // to avoid mixing layer's network updates and layer's params
     // Update material parameters
-    const layerIndex = material.indexOfColorLayer(layer.id);
-    material.setLayerVisibility(layerIndex, layer.visible);
-    material.setLayerOpacity(layerIndex, layer.opacity);
+    material.setLayerVisibility(layer, layer.visible);
+    material.setLayerOpacity(layer, layer.opacity);
 
     const ts = Date.now();
     // An update is pending / or impossible -> abort
@@ -223,7 +208,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
             node.layerUpdateState[layer.id].noMoreUpdatePossible();
             return;
         }
-    } else if (!node.isColorLayerDownscaled(layer)) {
+    } else if (!material.isColorLayerDownscaled(layer, node.getZoomForLayer(layer))) {
         // default decision method
         node.layerUpdateState[layer.id].noMoreUpdatePossible();
         return;
@@ -268,9 +253,9 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
             }
 
             if (Array.isArray(result)) {
-                node.setTexturesLayer(result, l_COLOR, layer.id);
+                node.material.setLayerTextures(layer, result);
             } else if (result.texture) {
-                node.setTexturesLayer([result], l_COLOR, layer.id);
+                node.material.setLayerTextures(layer, [result]);
             } else {
                 // TODO: null texture is probably an error
                 // Maybe add an error counter for the node/layer,
@@ -350,10 +335,6 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
         return Promise.resolve();
     }
 
-    // TODO
-    if (material.elevationLayersId.length === 0) {
-        material.elevationLayersId.push(layer.id);
-    }
     node.layerUpdateState[layer.id].newTry();
 
     const command = {
@@ -397,7 +378,7 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
                 insertSignificantValuesFromParent(terrain.texture, node, parent, layer);
             }
 
-            node.setTextureElevation(terrain);
+            node.setTextureElevation(layer, terrain);
         },
         (err) => {
             if (err instanceof CancelledCommandException) {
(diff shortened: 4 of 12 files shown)


src/Core/TileMesh.js
```diff
@@ -224,8 +219,10 @@ TileMesh.prototype.getCoordsForLayer = function getCoordsForLayer(layer) {
         } else {
             return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent, layer.origin);
         }
-    } else {
+    } else if (layer.extent.crs() == this.extent.crs()) {
         return [this.extent];
```
mbredif: Shouldn't this test go into `Extent.as` ?

src/utils/DEMUtils.js
```diff
@@ -169,7 +169,7 @@ function tileAt(pt, tile) {
                 return t;
             }
         }
-        if (tile.getLayerTextures(l_ELEVATION)[0].coords.zoom > -1) {
+        if (tile.material.textures[l_ELEVATION][0].coords.zoom > -1) {
```
mbredif: Is it equivalent to `isElevationLayerLoaded` ? (which would be clearer)

src/Core/TileMesh.js
```diff
@@ -11,10 +11,12 @@ import RendererConstant from '../Renderer/RendererConstant';
 import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from '../Provider/OGCWebServiceHelper';
 import { is4326 } from './Geographic/Coordinates';
 
-function TileMesh(geometry, params) {
+function TileMesh(layer, geometry, params) {
     // Constructor
     THREE.Mesh.call(this);
```
mbredif: the call to super could pass directly the geometry and material : `super(geometry, material)` (with a material constructed at the call site) and the constructor could be more explicit with : `TileMesh(geometry, material, extent, level)`, as in 6dd1ba54153a9123dfd7c8d8a0c28503a69056e2

src/Core/TileMesh.js
```diff
@@ -11,10 +11,12 @@ import RendererConstant from '../Renderer/RendererConstant';
...
 
+    this.layer = layer;
+
```
mbredif: What is the semantics of this layer ? which one is it ? I would rather keep it outside of TileMesh as with other meshes in itowns.

src/Core/TileMesh.js
```diff
@@ -37,6 +39,7 @@ function TileMesh(geometry, params) {
     this.frustumCulled = false;
 
     this.updateGeometricError();
+    this.wmtsCoords = {};
```
mbredif, line 34: I think it makes more sense to move this to `Extent`.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -66,7 +66,7 @@ function initNodeElevationTextureFromParent(node, parent, layer) {
             elevation.max = max;
         }
 
-        node.setTextureElevation(elevation);
+        node.setTextureElevation(layer, elevation);
         node.material.elevationLayersId =
             parent.material.elevationLayersId;
```
mbredif: you removed it...

src/Renderer/LayeredMaterial.js
```diff
@@ -363,7 +379,15 @@ LayeredMaterial.prototype.setColorLayerParameters = function setColorLayerParame
...
-LayeredMaterial.prototype.pushLayer = function pushLayer(param) {
+LayeredMaterial.prototype.pushLayer = function pushLayer(layer, extents) {
+    const param = {
+        tileMT: layer.options.tileMatrixSet || extents[0].crs(),
+        texturesCount: extents.length,
+        visible: layer.visible,
+        opacity: layer.opacity,
+        fx: layer.fx,
+        idLayer: layer.id,
+    };
```
mbredif: why do you create this intermediate object instead of using `layer` directly ?

src/Renderer/LayeredMaterial.js
```diff
@@ -423,12 +447,15 @@ LayeredMaterial.prototype.getLayerUV = function setLayerUV(index) {
...
+    if (this.uniforms.paramLayers.value[index]) {
+        this.uniforms.paramLayers.value[index].w = opacity;
+    }
 };
 
-LayeredMaterial.prototype.setLayerVisibility = function setLayerVisibility(index, visible) {
+LayeredMaterial.prototype.setLayerVisibility = function setLayerVisibility(layer, visible) {
+    const index = Number.isInteger(layer) ? layer : this.indexOfColorLayer(layer.id);
     this.uniforms.visibility.value[index] = visible;
 };
```
mbredif, line 442: could we deprecate the index-based api (with a console.warn) ? is it still used elsewhere ?

src/Core/TileMesh.js
```diff
@@ -158,14 +158,13 @@ TileMesh.prototype.setTexturesLayer = function setTexturesLayer(textures, layerT
...
-    return mat.getLayerTextures(layerType, layerId);
-};
-
-TileMesh.prototype.isColorLayerLoaded = function isColorLayerLoaded(layerId) {
+TileMesh.prototype.isColorLayerLoaded = function isColorLayerLoaded(layer) {
     const mat = this.material;
-    return mat.getColorLayerLevelById(layerId) > -1;
+    const textures = mat.getLayerTextures(layer);
+    if (textures.length) {
+        return textures[0].coords.zoom > -1;
```
mbredif: should not it be `textures.every(texture => texture.coords.zoom == this.level)` or `textures.every(texture => somefun(layer, this.extent, texture.coords)` for a suitable function ?

src/utils/DEMUtils.js
```diff
@@ -371,7 +371,7 @@ function _readZ(layer, method, coord, nodes, cache) {
     }
 
     const tile = tileWithValidElevationTexture;
-    const src = tileWithValidElevationTexture.getLayerTextures(l_ELEVATION)[0];
+    const src = tileWithValidElevationTexture.material.textures[l_ELEVATION][0];
```
mbredif, line 374: `tileWithValidElevationTexture.getLayerTextures(layer)[0]` ?

mbredif (commented):
This definitely goes in the right direction !
I made a few comments.

src/Core/TileMesh.js
peppsac: It could but actually the `as()` function should be improved to support a 'no-clone' argument (currently we always clone the extent, even if the requested crs is the same).

src/utils/DEMUtils.js
peppsac: Yes.

I'll use this *but* I'll remove it from TileMesh and move it where it belongs = LayeredMaterial.

src/Core/TileMesh.js
peppsac: It's the GeometryLayer, and as stated above it was already stored in TileMesh (and I think it's useful).

src/Process/LayeredMaterialNodeProcessing.js
peppsac: What do you mean?

src/Core/TileMesh.js
peppsac: Yeah... I moved it here, again because it was assigned somewhere else (https://github.com/iTowns/itowns/pull/792/files#diff-5e0253bd7132719c7fe2322c297efe16L48).

I propose to keep it that way for this PR, and use your PR #785  as an opportunity to fix this since it's more focused on the extent side. WDYT?

src/Renderer/LayeredMaterial.js
peppsac: I moved this intermediate object from elsewhere, but you're right it's useless. I'll remove it.

src/Renderer/LayeredMaterial.js
peppsac: It's only used internally in LayeredMaterial.

There are a bunch of other index based method in LayeredMaterial and I think they should all be removed. I'll do a separate PR for this stuff.

src/Core/TileMesh.js
peppsac: Actually it doesn't matter, since we always set textures in one shot (this is what getColorLayerLevelById was doing anyway)

src/utils/DEMUtils.js
peppsac: Yep, thanks.

peppsac: Ah no, because `layer` here is the geometry layer. And the calling site is not aware of which elevation layer we can use.

So for now I'll leave it like that.

peppsac:
Pushed a commit which should fix the above issues. Thanks for the review @mbredif :)

src/Core/TileMesh.js
mbredif: ok

src/Core/TileMesh.js
gchoqueux: where `tile.layer` is used?

src/Core/TileMesh.js
```diff
@@ -158,18 +157,13 @@ TileMesh.prototype.setTexturesLayer = function setTexturesLayer(textures, layerT
...
-TileMesh.prototype.isColorLayerLoaded = function isColorLayerLoaded(layerId) {
-    const mat = this.material;
-    return mat.getColorLayerLevelById(layerId) > -1;
-};
-
-TileMesh.prototype.isElevationLayerLoaded = function isElevationLayerLoaded() {
-    return this.material.loadedTexturesCount[l_ELEVATION] > 0;
+    const textures = mat.getLayerTextures(layer);
+    if (textures.length) {
+        return textures[0].coords.zoom > -1;
```
gchoqueux: use `EMPTY_TEXTURE_ZOOM` instead of -1

src/Core/TileMesh.js
gchoqueux: add some comment to explain your choice

src/Process/GlobeTileProcessing.js
```diff
@@ -125,11 +125,14 @@ export function globeSubdivisionControl(minLevel, maxLevel, sseThreshold, maxDel
         // Prevent to subdivise the node if the current elevation level
         // we must avoid a tile, with level 20, inherits a level 3 elevation texture.
         // The induced geometric error is much too large and distorts the SSE
-        const currentElevationLevel = node.material.getElevationLayerLevel();
-        if (node.level < context.maxElevationLevel + maxDeltaElevationLevel &&
-            currentElevationLevel >= 0 &&
-            (node.level - currentElevationLevel) >= maxDeltaElevationLevel) {
-            return false;
+        const currentTexture = node.material.textures[0][0];
```
gchoqueux: use `l_color` instead of 0

src/Process/GlobeTileProcessing.js
```diff
@@ -125,11 +125,14 @@ export function globeSubdivisionControl(minLevel, maxLevel, sseThreshold, maxDel
...
+        if (currentTexture.extent) {
+            const offsetScale = node.material.offsetScale[0][0];
+            const ratio = offsetScale.z;
+            // ratio is node size / texture size
+            if (ratio < 1 / Math.pow(2, maxDeltaElevationLevel)) {
```
gchoqueux: it's a ratio, between the tile's extent and texture's extent?

src/Process/PlanarTileProcessing.js
```diff
@@ -42,11 +42,14 @@ export function planarSubdivisionControl(maxLevel, maxDeltaElevationLevel) {
         // Prevent to subdivise the node if the current elevation level
         // we must avoid a tile, with level 20, inherits a level 3 elevation texture.
         // The induced geometric error is much too large and distorts the SSE
-        const currentElevationLevel = node.material.getElevationLayerLevel();
-        if (node.level < context.maxElevationLevel + maxDeltaElevationLevel &&
-            currentElevationLevel >= 0 &&
-            (node.level - currentElevationLevel) >= maxDeltaElevationLevel) {
-            return false;
+        const currentTexture = node.material.textures[0][0];
```
gchoqueux: this part of this process is identical to the globe processing, it should be encapsulated in one place. Duplicate code is harder and longer to maintain

src/Process/SubdivisionControl.js
```diff
@@ -28,7 +28,7 @@ export default {
             if (node.layerUpdateState[c.id] && node.layerUpdateState[c.id].inError()) {
                 continue;
             }
-            if (c.tileInsideLimit(node, c) && !node.isColorLayerLoaded(c.id)) {
+            if (c.tileInsideLimit(node, c) && !node.isColorLayerLoaded(c)) {
```
gchoqueux: if `isElevationLayerLoaded` is material function, why `isColorLayerLoaded` is `tileMesh` function?

src/Provider/TileProvider.js
```diff
@@ -80,8 +80,7 @@ function executeCommand(command) {
     };
 
     geometry._count++;
-    const tile = new TileMesh(geometry, params);
-    tile.layer = layer;
+    const tile = new TileMesh(layer, geometry, params);
```
gchoqueux: same question, you remember where `tile.layer` is used?

src/Renderer/LayeredMaterial.js
```diff
@@ -330,22 +329,39 @@ LayeredMaterial.prototype.removeColorLayer = function removeColorLayer(layer) {
...
-    const index = this.indexOfColorLayer(layer);
-    const slotOffset = this.getTextureOffsetByLayerIndex(index);
-    for (let i = 0, max = textures.length; i < max; i++) {
-        if (textures[i]) {
-            if (textures[i].texture !== null) {
-                this.setTexture(textures[i].texture, layerType, i + (slotOffset || 0), textures[i].pitch);
-            } else {
-                this.setLayerVisibility(index, false);
-                break;
+LayeredMaterial.prototype.setLayerTextures = function setLayerTextures(layer, textures) {
```
gchoqueux, line 321: you saw   @mbredif's proposal for the [`LayeredMaterial`](https://github.com/mbredif/itowns/blob/layered_material_rebased/src/Renderer/LayeredMaterial.js)

src/utils/DEMUtils.js
gchoqueux: and with the famous `tile.layer`?
you could write:
`tileWithValidElevationTexture.getLayerTextures(tile.layer)[0]`

gchoqueux (dismissed):
we should also add the tests on the layeredMaterial

src/Core/TileMesh.js
peppsac: For instance [here](https://github.com/iTowns/itowns/blob/master/src/Core/MainLoop.js#L101).

And I'll repeat myself: this isn't a new thing. I simply moved the `layer` assignement inside the constructor.

src/utils/DEMUtils.js
peppsac: > and with the famous tile.layer?

Are you sure the `famous` bit is needed?

> you could write: tileWithValidElevationTexture.getLayerTextures(tile.layer)[0]

And it would be incorrect. `tile.layer` is the GeometryLayer, not the Layer used for the elevation.

src/Core/TileMesh.js
peppsac: Done

src/Core/TileMesh.js
peppsac: Done

src/Process/GlobeTileProcessing.js
peppsac: Yes.

src/Process/PlanarTileProcessing.js
peppsac: > Duplicate code is harder and longer to maintain

This is not an absolute truth, and here I disagree. This is a 5 line copy-paste so I don't think it hurts to keep them separate.

If you really want to improve code sharing between part of itowns I opened a PR sometimes ago: #674

src/Process/SubdivisionControl.js
peppsac: Mainly because this PR is a part of a bigger PR. I tried to extract a meaningful part of it, but there are left over like this one that doesn't make sense.

I'm moving isColorLayerLoaded to LayeredMaterial.

src/Provider/TileProvider.js
peppsac: Answered above.

peppsac:
Updated.

peppsac:
@gchoqueux could you update your review please?

src/Process/PlanarTileProcessing.js
gchoqueux: the duplicate code is technical debt

autra: > Duplicate code is harder and longer to maintain

Abusive factorization too, by introducing another layer of abstraction. 

> the duplicate code is technical debt

The program without technical debt is the one with 0 codes. I agree duplicate code has a cost, but introducing another function has one too. I call it more or less even here. But as it is not the scope of this PR and as this snippet is quite short, maybe you can let this one go?

src/utils/DEMUtils.js
gchoqueux: Ok,
So we don't know in elevation layer the data is searched for?

gchoqueux: we should add tests to DEMUtils, if it doesn't exist

peppsac: Not the scope of this PR. I'm only changing this file to remove one useless abstraction (and it's a 2 line diff)

peppsac:
Ok, so IMHO this PR improves greatly itowns and will allow to reuse the texture update process for other types of data.

@gchoqueux: I don't think your Request Changes on this PR is justified.

src/Process/LayeredMaterialNodeProcessing.js
mbredif: do you still support material.elevationLayersId ? I tought you proposed to remove it.

peppsac: Ah ok. I pushed a commit removing this line.

src/Process/PlanarTileProcessing.js
gchoqueux: This function can be factorized:
  * as an OGC conversion tool. there is WMTS_WGS84Parent. Why not have the opposite function?
  * it's an identified conversion tool, so no code duplication

peppsac: Actually this code is a hack to workaround a bug in the SSE logic (= computing the geometric error using a sphere - except that a sphere is a poor approximation for a square tile).

So I'd rather get rid of it, instead of transforming it in a API that would be used in other places of iTowns.

gchoqueux:
> Ok, so IMHO this PR improves greatly itowns and will allow to reuse the texture update process for other types of data

yes, but you also need the missing tests of impacted objects

peppsac:
> yes, but you also need the missing tests of impacted objects

Most of them already have some tests or are pretty well exercised during with the `test-examples` target. So unless you have a specific test case in mind, I won't add any.

Stop blocking PRs for random reasons: being a reviewer is a responsability, you can't just click Request Changes without valid concerns about the code being merged.

peppsac:
Rebased.

examples/split.html
```diff
@@ -92,11 +92,8 @@
                 globeView.scene.traverse(function _(obj) {
                     if (obj.material && obj.material.setLayerVisibility && obj.material.visible) {
                         material = obj.material;
-                        orthoIndex = material.indexOfColorLayer(orthoLayer.id);
```
zarov, line 95: You can remove their declaration above.

src/Core/TileMesh.js
zarov: Why not rename it to a more appropriate name in the same time ?

src/Renderer/LayeredMaterial.js
```diff
@@ -330,22 +329,39 @@ LayeredMaterial.prototype.removeColorLayer = function removeColorLayer(layer) {
...
+                            i + (slotOffset || 0), textures[i].pitch);
+                    } else {
+                        this.setLayerVisibility(index, false);
+                        break;
+                    }
+                }
             }
+        } else if (textures.texture !== null) {
+            this._setTexture(textures.texture,
+                layer.type == 'color' ? l_COLOR : l_ELEVATION, (slotOffset || 0), textures.pitch);
```
zarov: Condition unnecessary, result is always `l_COLOR`.

src/Core/TileMesh.js
```diff
@@ -32,11 +28,10 @@ function TileMesh(geometry, params) {
     this.boundingSphere = new THREE.Sphere();
     this.OBB().box3D.getBoundingSphere(this.boundingSphere);
 
-    this.material = new LayeredMaterial(params.materialOptions);
```
zarov, line 35: :+1:

src/Provider/TileProvider.js
```diff
@@ -73,15 +74,9 @@ function executeCommand(command) {
...
-    const params = {
-        extent,
-        level,
-        materialOptions: layer.materialOptions,
-    };
-
     geometry._count++;
-    const tile = new TileMesh(geometry, params);
-    tile.layer = layer;
+    const material = new LayeredMaterial(layer.materialOptions);
```
zarov, line 78: Could you add somewhere a documentation to what is `materialOptions` ? Even if it is a quick comment, so we do no forget when we'll generate a better documentation on layers.

src/Renderer/LayeredMaterial.js
```diff
@@ -456,23 +477,31 @@ LayeredMaterial.prototype.getColorLayerLevelById = function getColorLayerLevelBy
...
+};
+
 LayeredMaterial.prototype.getElevationLayerLevel = function getElevationLayerLevel() {
     return this.textures[l_ELEVATION][0].coords.zoom;
 };
 
-LayeredMaterial.prototype.getLayerTextures = function getLayerTextures(layerType, layerId) {
-    if (layerType === l_ELEVATION) {
+LayeredMaterial.prototype.getLayerTextures = function getLayerTextures(layer) {
+    if (layer.type === 'elevation') {
```
zarov, line 481: Side note: shouldn't it be more practical to change the constants in `LayeredMaterialConstants` to `elevation`, `color` and all ?

zarov (commented):
I see a lot of good things here, thanks for this !

src/Provider/TileProvider.js
peppsac: Ok. Note that these options are a workaround to limitations in the current code.

Hopefully these limitations will be removed soon and so will the options :)

src/Renderer/LayeredMaterial.js
peppsac: IMHO, `LayeredMaterialConstants` should be removed soon because that should be an implementation details.

But we'll also need to have valid layer types as constants somewhere (`elevation`, `color`, etc).

src/Renderer/LayeredMaterial.js
peppsac: Done

peppsac:
> Why not rename it to a more appropriate name in the same time ?

I can't answer inline for this one so I'll reply here: the PR https://github.com/iTowns/itowns/pull/785 tackles this part, so I limited myself to moving code around.

examples/split.html
peppsac: Done

peppsac:
@zarov PR updated.

zarov:
Ok this is looking good to me. One last thing: do you have a suggestion for removing the `tile.layer` that I don't feel too is really useful (apart from the case you quote) ?

peppsac:
> do you have a suggestion for removing the tile.layer that I don't feel too is really useful (apart from the case you quote) ?

Not really, and I feel it's useful (another use case: in the picking code where we need to filter if an mesh belongs to the requested layer).

zarov (approved):
Ok good to go then, squash your last commit please.

PR #793, opened 2018-06-26
feat(pointclouds): compute extent for pointclouds elements

peppsac:
In order to be able to unify processing for elements of heterogeneous sources we must have some shared properties.
`layer.extent` and `element.extent` is one of these, and will allow to build colorization for pointcloud.

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -1,4 +1,5 @@
 import * as THREE from 'three';
+import Extent from '../Core/Geographic/Extent';
 
 function requestNewTile(view, scheduler, geometryLayer, metadata, parent, redraw) {
     const command = {
@@ -26,6 +27,32 @@ function subdivideNode(context, layer, node, cullingTest) {
     }
 }
 
+function boundingVolumeToExtent(crs, layer, volume, transform) {
+    if (volume.region) {
+        return new Extent('EPSG:4326',
+            THREE.Math.radToDeg(volume.region[0]),
+            THREE.Math.radToDeg(volume.region[2]),
+            THREE.Math.radToDeg(volume.region[1]),
+            THREE.Math.radToDeg(volume.region[3]));
+    } else if (volume.box) {
+        const box = volume.box.clone().applyMatrix4(transform);
+        return new Extent(crs, {
+            west: box.min.x,
+            east: box.max.x,
+            south: box.min.y,
+            north: box.max.y,
+        });
+    } else {
+        const sphere = volume.sphere.clone().applyMatrix4(transform);
+        return new Extent(crs, {
+            west: sphere.center.x - sphere.radius,
+            east: sphere.center.x + sphere.radius,
+            south: sphere.center.y - sphere.radius,
+            north: sphere.center.y + sphere.radius,
+        });
+    }
+}
+
 const tmpMatrix = new THREE.Matrix4();
 function _subdivideNodeAdditive(context, layer, node, cullingTest) {
     for (const child of layer.tileIndex.index[node.tileId].children) {
@@ -50,6 +77,12 @@ function _subdivideNodeAdditive(context, layer, node, cullingTest) {
         child.promise = requestNewTile(context.view, context.scheduler, layer, child, node, true).then((tile) => {
             node.add(tile);
             tile.updateMatrixWorld();
+
+            const extent = boundingVolumeToExtent(layer.extent.crs(), tile.layer, tile.boundingVolume, tile.matrixWorld);
+            tile.traverse((obj) => {
+                obj.extent = extent;
+            });
+
             context.view.notifyChange(child);
             child.loaded = true;
             delete child.promise;
@@ -254,6 +287,8 @@ export function init3dTilesLayer(view, scheduler, layer) {
                 tile.updateMatrixWorld();
                 layer.tileIndex.index[tile.tileId].loaded = true;
                 layer.root = tile;
+                layer.extent = boundingVolumeToExtent(layer.projection || view.referenceCrs,
+                    tile, tile.boundingVolume, tile.matrixWorld);
             });
 }
 
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -27,6 +27,8 @@ function subdivideNode(context, layer, node, cullingTest) {
     }
 }
 
+const tmpBox3 = new THREE.Box3();
+const tmpSphere = new THREE.Sphere();
 function boundingVolumeToExtent(crs, volume, transform) {
     if (volume.region) {
         return new Extent('EPSG:4326',
@@ -35,10 +37,10 @@ function boundingVolumeToExtent(crs, volume, transform) {
             THREE.Math.radToDeg(volume.region[1]),
             THREE.Math.radToDeg(volume.region[3]));
     } else if (volume.box) {
-        const box = volume.box.clone().applyMatrix4(transform);
+        const box = tmpBox3.copy(volume.box).applyMatrix4(transform);
         return Extent.fromBox3(crs, box);
     } else {
-        const sphere = volume.sphere.clone().applyMatrix4(transform);
+        const sphere = tmpSphere.copy(volume.sphere).applyMatrix4(transform);
         return new Extent(crs, {
             west: sphere.center.x - sphere.radius,
             east: sphere.center.x + sphere.radius,
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -36,12 +36,7 @@ function boundingVolumeToExtent(crs, volume, transform) {
             THREE.Math.radToDeg(volume.region[3]));
     } else if (volume.box) {
         const box = volume.box.clone().applyMatrix4(transform);
-        return new Extent(crs, {
-            west: box.min.x,
-            east: box.max.x,
-            south: box.min.y,
-            north: box.max.y,
-        });
+        return Extent.fromBox3(crs, box);
     } else {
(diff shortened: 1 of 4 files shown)


src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,32 @@ function subdivideNode(context, layer, node, cullingTest) {
     }
 }
 
+function boundingVolumeToExtent(crs, layer, volume, transform) {
+    if (volume.region) {
+        return new Extent('EPSG:4326',
+            THREE.Math.radToDeg(volume.region[0]),
+            THREE.Math.radToDeg(volume.region[2]),
+            THREE.Math.radToDeg(volume.region[1]),
+            THREE.Math.radToDeg(volume.region[3]));
```
mbredif, line 38: what if crs is not 'EPSG:4326' ?

src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,32 @@ function subdivideNode(context, layer, node, cullingTest) {
     }
 }
 
+function boundingVolumeToExtent(crs, layer, volume, transform) {
```
mbredif: layer is unused

src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,32 @@ function subdivideNode(context, layer, node, cullingTest) {
...
+            south: box.min.y,
+            north: box.max.y,
+        });
+    } else {
+        const sphere = volume.sphere.clone().applyMatrix4(transform);
+        return new Extent(crs, {
+            west: sphere.center.x - sphere.radius,
+            east: sphere.center.x + sphere.radius,
+            south: sphere.center.y - sphere.radius,
+            north: sphere.center.y + sphere.radius,
```
mbredif, line 48: is transform restricted to affine transforms with isotropic scaling or is applyMatrix4 ensuring a bounding sphere of the resulting quadric ?

src/Process/3dTilesProcessing.js
```diff
@@ -254,6 +287,8 @@ export function init3dTilesLayer(view, scheduler, layer) {
                 tile.updateMatrixWorld();
                 layer.tileIndex.index[tile.tileId].loaded = true;
                 layer.root = tile;
+                layer.extent = boundingVolumeToExtent(layer.projection || view.referenceCrs,
+                    tile, tile.boundingVolume, tile.matrixWorld);
```
mbredif: `tile` should have been `tile.layer` (but this arguments is unused and should go away)

src/Provider/PointCloudProvider.js
```diff
@@ -141,8 +143,24 @@ function addPickingAttribute(points) {
     return points;
 }
 
+function bboxToExtent(view, bbox) {
```
mbredif: this only depends on the crs -> `bboxToExtent(crs, bbox)`. I would prefer something like `extent.fromBox3(box3, crs)`.
Should not it share the same implementation as in the bbox case for 3dtiles (except with no transform) ?

src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,32 @@ function subdivideNode(context, layer, node, cullingTest) {
...
+        const sphere = volume.sphere.clone().applyMatrix4(transform);
+        return new Extent(crs, {
+            west: sphere.center.x - sphere.radius,
+            east: sphere.center.x + sphere.radius,
+            south: sphere.center.y - sphere.radius,
+            north: sphere.center.y + sphere.radius,
+        });
+    }
+}
+
```
mbredif, line 52: Should not it be moved to Extent as well ?

mbredif (commented):
Great!
Going further in this direction (in follow up PRs), I would discard the TileMesh class altogether and just have Layered Material colorMap and displacementMap all nodes with an extent property.

(Note for later : relying on extents is heavily influenced by orthographic projections of coloring with axis-aligned (lat-lon) orthophotos and displacing with DSMs, projective texturing from arbitrary coverages will likely have to rely on other types bounding volumes directly in `layer.getCoord` without an intermediate conversion to an extent)

src/Process/3dTilesProcessing.js
peppsac: crs is guaranteed to be 4326 by the specification (see https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/schema/boundingVolume.schema.json#L19)

src/Process/3dTilesProcessing.js
peppsac: The scaling is not necessarly isotropic but, yes, three.js scales the radius by the largest scale factor (see https://github.com/mrdoob/three.js/blob/master/src/math/Sphere.js#L157) so we're good.

src/Process/3dTilesProcessing.js
peppsac: Removed, thanks.

src/Provider/PointCloudProvider.js
peppsac: I'm reluctant to abstract / share code too early. So I prefer to keep the 2 separate implementations.

Maybe in some time we may want to merge them (or move them up in Extent) - or maybe not for reasons that we're not aware of now.

src/Process/3dTilesProcessing.js
mbredif: ok for the input region crs, but that does noy say anything about the target crs. I think you need a reprojection using ,`as(crs)`. Or a comment stating why you do not need it !

peppsac: > but that does noy say anything about the target crs. 

The 4326 extent is the correct one to use here by definition. Why should we need to reproject?

src/Process/3dTilesProcessing.js
mbredif: good to know!
Please leave a comment here that THREE does what we expect (conservative sphere bounding the transformed quadric)

src/Process/3dTilesProcessing.js
mbredif: i see : I thought it was the ouput crs , but it is indeed the input crs for unspecified cases (all but regions)

src/Provider/PointCloudProvider.js
mbredif: I see your point, but the downside of your approach is that you end up coding twice the box3 to extent conversion with 2 different implementations and no tests...

Why is the hard-coded crs test required in one case and not the other?

peppsac: Ok, I'll make a shared method in Extent (with a unit test !) :)

peppsac: Done - and removed the 4978 -> 4326 conversion, since it loses one information (z)

src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,27 @@ function subdivideNode(context, layer, node, cullingTest) {
...
 
+function boundingVolumeToExtent(crs, volume, transform) {
+    if (volume.region) {
+        return new Extent('EPSG:4326',
+            THREE.Math.radToDeg(volume.region[0]),
+            THREE.Math.radToDeg(volume.region[2]),
+            THREE.Math.radToDeg(volume.region[1]),
+            THREE.Math.radToDeg(volume.region[3]));
+    } else if (volume.box) {
+        const box = volume.box.clone().applyMatrix4(transform);
```
autra: Let's have a temp value here.

src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,27 @@ function subdivideNode(context, layer, node, cullingTest) {
...
+        return Extent.fromBox3(crs, box);
+    } else {
+        const sphere = volume.sphere.clone().applyMatrix4(transform);
```
autra: also a tmp value here.

src/Process/3dTilesProcessing.js
```diff
@@ -26,6 +27,27 @@ function subdivideNode(context, layer, node, cullingTest) {
     }
 }
 
+function boundingVolumeToExtent(crs, volume, transform) {
```
autra: What about having a `target` param as well, to avoid creating a new `Extent` if possible ?

src/Process/3dTilesProcessing.js
autra: Do we really need a comment saying "Works as intended" ?

src/Process/3dTilesProcessing.js
autra: The volume part is specific to 3dtiles AFAIK. Maybe the sphere thing could be moved in a `.fromSphere`, although it'll still be possible to do it later if/when it is used elsewhere. @peppsac WDYT?

autra (commented):
Already looking good to me. Left a few comments, mostly to avoid useless instanciation.

src/Process/3dTilesProcessing.js
peppsac: Yes, let's keep it here for now.

src/Process/3dTilesProcessing.js
peppsac: For now it's always used to create a `Extent` so a `target` param wouldn't be used.

src/Process/3dTilesProcessing.js
peppsac: Done (both tmp values)

autra (approved):
Looks good to me. Can you squash and mention that this is for *3Dtiles* pointclouds in your commit message please?

peppsac:
> Can you squash and mention that this is for 3Dtiles pointclouds in your commit message please?

It's for all pointclouds, not just 3dtiles ones.

autra:
Oh indeed! Allez klar then.

PR #794, opened 2018-06-27
refactor: move geometric layers off views

zarov:
Move all contents of `createXXXLayer` in XXXView to XXXLayer in Prefab/XXX/. This declutters the XXXView files. This change applies for Globe, Planar and Panorama.

In the same time, GeometryLayer has been extract from Layer and given some new methods which were repeated through each View.

The idea here is to simplifies some files, and later move the related processing of each layer (GlobeTileProcessing, etc) inside either the Layer itself, or in the same folder, to group each "type" of view and its workings (this, but also Controls too) in the same spot. Note that this is a proposition.


src/Core/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,104 @@
...
+    if (!id) {
+        throw new Error('Missing id parameter (GeometryLayer must have a unique id defined)');
+    }
+    if (!object3d || !object3d.isObject3D) {
+        throw new Error('Missing/Invalid object3d parameter (must be a three.js Object3D instance)');
+    }
+
+    this._attachedLayers = [];
+    this.type = 'geometry';
+    this.protocol = 'tile';
```
mbredif: Is GeometryLayer restricted to be used with the TileProtocol?

src/Core/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,104 @@
...
+    this._attachedLayers.push(layer);
+};
+
+GeometryLayer.prototype.detach = function detach(layer) {
+    const count = this._attachedLayers.length;
+    this._attachedLayers = this._attachedLayers.filter(attached => attached.id != layer.id);
+    return this._attachedLayers.length < count;
+};
+
+GeometryLayer.prototype.onTileCreated = function onTileCreated(layer, parent, node) {
```
mbredif: This does not use `this`. which `layer` is it ? I suggest you either use `this` instead or make it a static function.

src/Core/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,104 @@
...
+        node.material.uniforms.noTextureColor.value.copy(layer.noTextureColor);
+    }
+
+    if (__DEBUG__) {
+        node.material.uniforms.showOutline = { value: layer.showOutline || false };
+        node.material.wireframe = layer.wireframe || false;
+    }
+};
+
+GeometryLayer.prototype.preUpdate = function preUpdate(context, layer, changeSources) {
```
mbredif: + same thing here (no use of `this`)

+ it relies on `layer.level0Nodes` which is only relevant for tiled GeometryLayer. Maybe there is room to create a TiledGeometryLayer subclass.

mbredif (commented):
I support the approach and left some comments to go further.

src/Core/Layer/GeometryLayer.js
zarov: Good catch, I also removed `parent` as it is not used.

src/Core/Layer/GeometryLayer.js
zarov: Yes

gchoqueux: Actually, it shouldn't have a tile protocol in `geometryLayer`. you would need another `tiledGeometryLayer` or `tiledMeshLayer` inherited of `geometryLayer`.

Ideally it would be necessary to remove the layer protocol, and to replace it a source (where source is object which makes it possible to obtain the data)

src/Core/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,104 @@
+import { EventDispatcher } from 'three';
+import Picking from '../Picking';
+
+function GeometryLayer(id, object3d) {
```
gchoqueux: you could use `class`?

src/Core/Layer/GeometryLayer.js
zarov: - removed
- yeah @gchoqueux said that as well, what do you suggest ?

zarov:
So I pushed some changes, but I'm not satisfied with moving from `layer` to `this` in PointCloudProcessing and 3dtilesProcessing. Maybe we could move to PointCloudLayer and 3dtilesLayer in the same time ?

src/Core/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,104 @@
...
+        throw new Error('Missing id parameter (GeometryLayer must have a unique id defined)');
+    }
+    if (!object3d || !object3d.isObject3D) {
+        throw new Error('Missing/Invalid object3d parameter (must be a three.js Object3D instance)');
+    }
+
+    this._attachedLayers = [];
+    this.type = 'geometry';
+    this.protocol = 'tile';
+    this.visible = true;
```
gchoqueux: would visibility and opacity not be rendering/style options?

src/Core/Layer/GeometryLayer.js
gchoqueux: logically, it should be called meshLayer

src/Core/Layer/GeometryLayer.js
gchoqueux: rename on `onMeshCreated`

src/Core/Layer/GeometryLayer.js
gchoqueux: yes, i agree

you would need a `meshLayerProcessing` class consisting of static functions that would then be attached to the layer.

actually, this `preUpdate` should be specific to the `quadtree / tiledMesh` and should be in `tiledMeshLayerProcessing`

src/Core/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,104 @@
...
+        if (__DEBUG__) {
+            this._latestUpdateStartingLevel = commonAncestor.level;
+        }
+        return [commonAncestor];
+    } else {
+        return this.level0Nodes;
+    }
+};
+
+export default GeometryLayer;
```
gchoqueux: finally it might look like this
```js
class meshLayer {
   source: {
       // how get data
   },
   processing: { 
       // processing attached to layer
   },
   style: {
        // how render this layer
   },
}
```

src/Core/Layer/GeometryLayer.js
zarov: > you could use `class`?

Wouldn't that be losing coherence with the actual code or do you wish to push the use of `class` elsewhere ?

> logically, it should be called meshLayer

But then the `type` would be `mesh`, and I find it too specific.

src/Core/Layer/GeometryLayer.js
zarov: Yes for `opacity`, no for `visibility` imho.

zarov:
@mbredif @gchoqueux I understand the concern on introducing a `TiledGeometryLayer` here, but I feel this is out of scope of this PR which is only a reorganization of code. I'm open to work on it right after this one, and maybe take the opportunity to clear some definition on layers. WDYT ?

peppsac:
> I understand the concern on introducing a TiledGeometryLayer here, but I feel this is out of scope of this PR which is only a reorganization of code.

As GeometryLayer is currently used by 3dtiles and pointcloud, you can't ignore them.
If you want a simple "reorganize" code PR, then I'd suggest to not modify `GeometryLayer` at all, and instead create a `TiledGeometryLayer` (that would be more or less the GeometryLayer from your PR).

WDYT?

zarov:
Closing this PR, I will do another one soon from another branch (see this more complete commit https://github.com/zarov/itowns/commit/3af73b0ed9f39bceaac2625e417a7625901e349c)

PR #795, opened 2018-06-28
Utilities to place a camera with geographic settings

gchoqueux:
## Description

Utilities to place a camera with geographic settings.
This utility works in globe and planar, and doesn't yet integrate the panoramic or immersive mode.
all settings are all optional 
```js
settings = {
   coord,  // target coordinate to camera's lookAt
   range, // distance between target coordinate and camera
   heading, // camera's heading
   pitch, // camera's pitch
}
```

## Motivation and Context
it is a tool with a geomatic approach the visualization of geographical data.
Simplify enormously `globeControls` (it's a first step in its refacto), the method is more robust and removes a lot of bugs in `globeControls`


examples/syncCameras.js
```diff
@@ -0,0 +1,86 @@
+/* global itowns, document, GuiTools, Promise */
```
zarov: Don't forget that now we have examples in a single html file

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,340 @@
+import * as THREE from 'three';
+import TWEEN from '@tweenjs/tween.js';
+import DEMUtils from './DEMUtils';
+import { MAIN_LOOP_EVENTS } from '../Core/MainLoop';
+import Coordinates from '../Core/Geographic/Coordinates';
+
+THREE.Object3D.DefaultUp = new THREE.Vector3(0, 0, 1);
```
zarov: What is that ? You are modifying the global prototype used everywhere here so I'm a little concerned.

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,340 @@
...
+    this.target.rotation.order = 'ZXY';
+    // camera look at target
+    this.camera = new THREE.Camera();
+    this.add(this.seaLevel);
+    this.seaLevel.add(this.target);
+    this.target.add(this.camera);
+    this.coord = new Coordinates('EPSG:4978', 0, 0);
+}
+
+RigCamera.prototype = Object.assign(Object.create(THREE.Object3D.prototype), {
```
zarov: Why not inherit like we do it in other classes ?

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,340 @@
...
+ */
+export default {
+    /**
+     * Stop camera's animation
+     *
+     * @param      {View}  view    The camera view
+     * @param      {Camera}  camera  The camera to stop animation
+     */
+    stop(view, camera) {
+        getRig(view, camera).stop(view);
```
zarov: This seems sloppy, I would prefer to see this going in `View`, as it is something really specific to the view. Why not instantiating a `RigCamera` when creating a `View` ? Is it to avoid dependency between those two class ? Maybe putting it inside a `Controls` abstract class would be better ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -770,210 +460,84 @@ function GlobeControls(view, target, radius, options = {}) {
...
-                update();
-            }
-        };
-    }());
+        if (state !== this.states.NONE) {
+            update();
+        }
+    }
 
     this.states = {
```
zarov: Should be moved outside this class.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -770,210 +460,84 @@ function GlobeControls(view, target, radius, options = {}) {
...
+        event.preventDefault();
+        const coords = view.eventToViewCoords(event);
 
-            snapShotCamera.updateRay(ray, mouse);
-            // pick position on tSphere
-            const its = tSphere.intersectWithRay(ray);
-            if (its != undefined) {
-                tSphere.picking.position.copy(its);
-                tSphere.picking.normal = tSphere.picking.position.clone().normalize();
+        if (state === this.states.ORBIT || state === this.states.PANORAMIC) {
```
zarov: You can use a `switch` structure to stay on the same page as other methods below.

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,340 @@
...
+    Object.defineProperty(camera.position, key, {
+        get: () => rig.proxy.position[key],
+        set: (newValue) => {
+            rig.removeProxy(view, camera);
+            camera.position[key] = newValue;
+        },
+    });
+}
+
+function RigCamera() {
```
zarov: This needs a thorougher description.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -332,58 +161,6 @@ export const CONTROL_EVENTS = {
...
-    return deferedPromise;
-}
-
-let initPromise;
-
-/* globals document,window */
-
 /**
  * @class
  * @param {GlobeView} view
```
zarov: Please add a description of the options, in the same idea as in `PlanarControls`.

zarov (commented):
A first very quick review, more to come as it is quite long.

And congrats on reducing and simplifying the `GlobeControls` :+1: !

src/utils/CameraUtils.js
mbredif: ... and a rename as it is not a `Camera` as this name implies. maybe `CameraRig` ?

peppsac:
>  +2,950 −2,650 

That's a huge PR.

I'm a bit disappointed that you didn't [open](https://github.com/iTowns/itowns/pull/740#issuecomment-384212508) an [issue](https://github.com/iTowns/itowns/pull/740#issuecomment-384241399) to discuss it [first](https://github.com/iTowns/itowns/pull/740#issuecomment-384583079) :stuck_out_tongue: 

(to be clear: I'm kidding. I already stated multiple times that I prefer to review PR than endlessly discussing)

src/utils/CameraUtils.js
gchoqueux: this is already the case in the viewers

examples/syncCameras.js
gchoqueux: done

src/utils/CameraUtils.js
gchoqueux: i changed and i used `class`

src/Renderer/ThreeExtended/StateControl.js
```diff
@@ -0,0 +1,74 @@
...
+    UP: 38,
+    RIGHT: 39,
+    BOTTOM: 40,
+    SPACE: 32,
+    SHIFT: 16,
+    CTRL: 17,
+    S: 83,
+};
+
+class StateControl {
```
zarov: Could you add some documentation here on how to read a state ? That would be useful if we add some in the future and if we use it with other control than `GlobeControls`.

test/unit/CameraUtils_unit_test.js
```diff
@@ -0,0 +1,135 @@
+/* global describe, it */
```
zarov: This can be removed

test/unit/CameraUtils_unit_test.js
```diff
@@ -0,0 +1,135 @@
...
+function pickEllipsoid(camera) {
+    raycaster.setFromCamera(center, camera);
+    return ellipsoid.intersection(raycaster.ray);
+}
+
+const view = {};
+const camera = new THREE.PerspectiveCamera();
+
+view.getPickingPositionFromDepth = function getPickingPositionFromDepth() {
+    return pickEllipsoid(camera);
```
zarov: You could write the `pickEllipsoid` function directly here.

test/unit/CameraUtils_unit_test.js
```diff
@@ -0,0 +1,135 @@
...
+import * as THREE from 'three';
+import assert from 'assert';
+import Coordinates, { ellipsoidSizes } from '../../src/Core/Geographic/Coordinates';
+import Ellipsoid from '../../src/Core//Math/Ellipsoid';
+import CameraUtils from '../../src/utils/CameraUtils';
+import DEMUtils from '../../src/utils/DEMUtils';
+
+THREE.Object3D.DefaultUp = new THREE.Vector3(0, 0, 1);
+
+DEMUtils.getElevationValueAt = () => {
```
zarov: Return directly the `result` object: `DEMUtils.getElevationValueAt = () => ({ z: 0 });`

test/unit/CameraUtils_unit_test.js
```diff
@@ -0,0 +1,135 @@
...
+}];
+view.addFrameRequester = () => {};
+view.removeFrameRequester = () => {};
+view.notifyChange = () => { camera.updateMatrixWorld(true); };
+
+const range = 25000000;
+const coord = new Coordinates('EPSG:4326', 2.35, 48.85, 0);
+
+/*
+// FIX ME convert
```
zarov: Will it be fixed ? If not remove it

test/unit/CameraUtils_unit_test.js
```diff
@@ -0,0 +1,135 @@
...
+*/
+
+
+function equalToFixed(value1, value2, toFixed) {
+    assert.equal(value1.toFixed(toFixed), value2.toFixed(toFixed));
+}
+
+describe('Camera utils unit test', function () {
+    it('init like expected', function () {
+        // FIX ME doesn't work without tilt and heading
```
zarov: If it needs a fix me, the initialization should be without the tilt and heading, and the test not working: that's the point of testing.

test/unit/CameraUtils_unit_test.js
```diff
@@ -0,0 +1,135 @@
...
+console.log(coord.longitude());
+const a = coord.as(view.referenceCrs);
+
+const b = a.as('EPSG:4326');
+console.log(b.longitude());
+*/
+
+
+function equalToFixed(value1, value2, toFixed) {
+    assert.equal(value1.toFixed(toFixed), value2.toFixed(toFixed));
```
zarov: This is not working, as the `assert` condition is outside the `describe` part: this method should return the comparison between the two values, and then you could do below `assert.ok(equalToFixed(value1, value2, toFixed));`.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1,92 +1,10 @@
...
-    this._x = (x * ratioA + this._x * ratioB);
-    this._y = (y * ratioA + this._y * ratioB);
-    this._z = (z * ratioA + this._z * ratioB);
-
-    this.onChangeCallback();
-
-    return this;
-};
+import StateControl from './StateControl';
+import CameraUtils from './../../utils/CameraUtils';
```
zarov: No need for the `./`

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -456,118 +207,65 @@ function GlobeControls(view, target, radius, options = {}) {
...
+        const te = this.camera.matrix.elements;
         // get Y column of matrix
-        panOffset.set(te[4], te[5], te[6]);
+        panOffset.fromArray(te, 4);
         panOffset.multiplyScalar(distance);
-
         panVector.add(panOffset);
     };
 
+    const states = new StateControl();
```
zarov: Couldn't you do this where you declare `state` above ? Is there really a point in having it bind to `this` ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -456,118 +207,65 @@ function GlobeControls(view, target, radius, options = {}) {
...
+    state = states.NONE;
+    this.states = states;
+
     // pass in x,y of change desired in pixel space,
     // right and down are positive
     this.mouseToPan = function mouseToPan(deltaX, deltaY) {
-        var element = this.domElement === document ? this.domElement.body : this.domElement;
-
+        const gfx = view.mainLoop.gfxEngine;
```
zarov: You use a lot the `gfxEngine` for height and width purpose: what about having those properties available directly like `this.size.height` ? You could emit an event in `c3DEngine/onWindowResize` that tell all subscribed elements the new values of `height` and `width`.
But that could may be done in another PR.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -456,118 +207,65 @@ function GlobeControls(view, target, radius, options = {}) {
...
             // half of the fov is center to top of screen
-            targetDistance *= Math.tan((this.camera.fov / 2) * Math.PI / 180.0);
+            targetDistance *= 2 * Math.tan(THREE.Math.degToRad(this.camera.fov * 0.5));
 
             // we actually don't use screenWidth, since perspective camera is fixed to screen height
-            this.panLeft(2 * deltaX * targetDistance * this.camera.aspect / element.clientWidth);
-            this.panUp(2 * deltaY * targetDistance / element.clientHeight);
+            this.panLeft(deltaX * targetDistance / gfx.width * this.camera.aspect);
+            this.panUp(deltaY * targetDistance / gfx.height);
+            state = states.PAN;
```
zarov: This should be done at the top of the method.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -456,118 +207,65 @@ function GlobeControls(view, target, radius, options = {}) {
...
         } else if (this.camera instanceof THREE.OrthographicCamera) {
             // orthographic
-            this.panLeft(deltaX * (this.camera.right - this.camera.left) / element.clientWidth);
-            this.panUp(deltaY * (this.camera.top - this.camera.bottom) / element.clientHeight);
-        } else {
-
-            // camera neither orthographic or perspective
```
zarov, line 515: What about removing it too in the next `dolly` methods ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -622,515 +319,361 @@ function GlobeControls(view, target, radius, options = {}) {
...
+    const resultDamping = [];
+
+    const update = () => {
         // We compute distance between camera's bounding sphere and geometry's obb up face
         if (this.handleCollision) { // We check distance to the ground/surface geometry
             // add minDistanceZ between camera's bounding and tiles's oriented bounding box (up face only)
             // Depending on the distance of the camera with obbs, we add a slowdown or constrain to the movement.
             // this constraint or deceleration is suitable for two types of movement MOVE_GLOBE and ORBIT.
             // This constraint or deceleration inversely proportional to the camera/obb distance
-            if (this._view.wgs84TileLayer) {
```
zarov, line 579: There is still some `this._view` that could be changed in `view`.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -622,515 +319,361 @@ function GlobeControls(view, target, radius, options = {}) {
...
-        return sphericalDelta;
     };
 
-    const direction = new THREE.Vector3();
-    const coordTarget = new Coordinates(this._view.referenceCrs, 0, 0, 0);
-    const coordTile = new Coordinates(this._view.referenceCrs, 0, 0, 0);
-    const reposition = new THREE.Vector3();
-    const delta = 0.001;
-
-    const updateCameraTargetOnGlobe = function updateCameraTargetOnGlobe() {
```
zarov, line 734: :+1:

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -96,136 +14,47 @@ const rotateStart = new THREE.Vector2();
...
 
 // Globe move
-const quatGlobe = new THREE.Quaternion();
-const cameraTargetOnGlobe = new THREE.Object3D();
-const movingCameraTargetOnGlobe = new THREE.Vector3();
-var animatedScale = 0.0;
+const moveAroundGlobe = new THREE.Quaternion();
+const cameraTarget = new THREE.Object3D();
+cameraTarget.matrixWorldInverse = new THREE.Matrix4();
+cameraTarget.needUpdate = true;
```
zarov: If you want to be similar to THREE, it is `needsUpdate` with an `s` ;)

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -622,515 +319,361 @@ function GlobeControls(view, target, radius, options = {}) {
...
         }
     };
 
-    function onMouseUp(/* event */) {
+    this._onEndingMove = (current) => {
+        state = states.NONE;
+        this._handlingEvent(current);
+    };
+
+    // test event remove it
```
zarov: Don't forget to remove this !

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -622,515 +319,361 @@ function GlobeControls(view, target, radius, options = {}) {
...
-            return;
-        }
+    this._handlingEvent = (current) => {
+        current = current || CameraUtils.getTransformCameraLookingAtTarget(view, this.camera);
+        const diff = CameraUtils.getDiffParams(previous, current);
+        if (diff) {
+            if (diff.range) {
+                this.dispatchEvent({
+                    type: CONTROL_EVENTS.RANGE_CHANGED,
+                    previous: { range: diff.range.previous },
```
zarov: Why not making it more generic and set directly the `previous` and `next` values ? I didn't see a case where there is more than one property set.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1186,129 +727,130 @@ function GlobeControls(view, target, radius, options = {}) {
...
-                    var dx = event.touches[0].pageX - event.touches[1].pageX;
-                    var dy = event.touches[0].pageY - event.touches[1].pageY;
-                    var distance = Math.sqrt(dx * dx + dy * dy);
-                    dollyStart.set(0, distance);
-                    break;
+            state = states.touchToState(event.touches.length);
+
+            updateTarget(true);
+
+            if (state !== states.NONE) {
```
zarov: You can remove this condition and remove an indentation next ;)

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1186,129 +727,130 @@ function GlobeControls(view, target, radius, options = {}) {
...
-                    panStart.set(event.touches[0].pageX, event.touches[0].pageY);
-                    break;
-                default:
+                this.dispatchEvent(this.startEvent);
             }
+        });
+    }
 
-            this.dispatchEvent(this.startEvent);
+    function onTouchMove(event) {
```
zarov, line 790: What is the real difference between this function and `onMouseMove` ? Couldn't you merge both ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1517,76 +994,7 @@ GlobeControls.prototype.setHeading = function setHeading(heading, isAnimated) {
...
-};
-
-/**
- * Sets orientation angles of the current camera, in degrees.
- * <iframe width="100%" height="400" src="http://jsfiddle.net/iTownsIGN/9qr2mogh/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
- * @param {{tilt:number,heading:number,range:number}} position
- * @param {boolean}  isAnimated
- * @return {Promise<void>}
- */
-GlobeControls.prototype.setOrbitalPosition = function setOrbitalPosition(position, isAnimated) {
```
zarov, line 1471: Please add a deprecated notice for following functions, and bind them to their replacement.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1714,10 +1042,8 @@ GlobeControls.prototype.moveTarget = function moveTarget() {
  */
 GlobeControls.prototype.pan = function pan(pVector) {
     this.mouseToPan(pVector.x, pVector.y);
-    this.updateCameraTransformation(this.states.PAN);
-    return this.waitSceneLoaded().then(() => {
-        this.updateCameraTransformation();
-    });
+    this.update();
+    return true;
```
zarov: Don't forget to update the documentation

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1856,12 +1189,11 @@ GlobeControls.prototype.metersToPixels = function metersToPixels(value, pixelPit
  */
  // TODO pas de scale supérieur à 0.05....
```
zarov: What about this TODO, can you resolve it here ?

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1874,23 +1206,77 @@ GlobeControls.prototype.setScale = function setScale(scale, pitch, isAnimated) {
...
+
+/**
+ * Changes the center of the scene on screen to the specified in lat, lon. See {@linkcode Coordinates} for conversion.
+ * This function allows to change the central position, the zoom, the range, the scale and the camera orientation at the same time.
+ * The zoom has to be between the [getMinZoom(), getMaxZoom()].
+ * Zoom parameter is ignored if range is set
+ * The tilt's interval is between 4 and 89.5 degree
+ *
+ * @param      {cameraTransformOptions}   params camera transformation to apply
+ * @param      {number}   [params.zoom]   zoom
```
zarov, line 1179: You forgot other parameters.

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1874,23 +1206,77 @@ GlobeControls.prototype.setScale = function setScale(scale, pitch, isAnimated) {
...
+};
+
+/**
+ * Changes the center of the scene on screen to the specified coordinates
+ *
+ * @param      {Coordinates} coord specified coordinates
+ * @param      {boolean}    isAnimated  Indicates if change is animated
+ * @return     {Promise}    A promise that resolves when center is on specified coordinates
+ */
+GlobeControls.prototype.lookAtCoordinate = function _lookAtCoordinate(coord, isAnimated) {
```
zarov: I would delete this function, rename the next one to `lookAtCoordinate`, and assume that if the first parameter is an instance of `Coordinates`,  you change it to `{ coord: params }`. I don't really like the `lookAtCoordinateAdvanced` name ;)

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,448 @@
...
+const slerp = [];
+
+const deferred = () => {
+    let resolve;
+    return { promise: new Promise((r) => { resolve = r; }), resolve };
+};
+
+// Wrap angle in degrees to [-180 180]
+function wrapTo180(angle) {
+    return angle - Math.floor((angle + 179.0) / 360) * 360;
```
zarov: What about `(angle % 360) - 180` ?

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,448 @@
...
+}
+
+function tileLayer(view) {
+    return view.getLayers(l => l.protocol == 'tile')[0];
+}
+
+// fix coord.altitude < 0 and !pickedPosition (with plane and ellipsoid intersection)
+function getGroundTargetFromCamera(view, camera, target) {
+    camera.updateMatrixWorld(true);
+    // FIXME view.wgs84TileLayer
```
zarov: What is the problem here ?

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,448 @@
...
+        if (this.proxy) {
+            camera.quaternion.onChange(() => {});
+            this.camera.matrixWorld.decompose(this.proxy.position, camera.quaternion, camera.scale);
+            camera.quaternion.onChange(() => this.removeProxy(view, camera));
+        } else {
+            this.camera.matrixWorld.decompose(camera.position, camera.quaternion, camera.scale);
+        }
+    }
+
+    setProxy(view, camera) {
```
zarov, line 108: Could you elaborate more on the role of `proxy` ?

src/utils/CameraUtils.js
```diff
@@ -0,0 +1,448 @@
...
+            .to(this.end.camera.position, time)
+            .easing(params.easing));
+
+        // update animations, transformation and view
+        const animationFrameRequester = () => {
+            tweenGroup.update();
+            this.updateMatrixWorld(true);
+            this.applyTransformToCamera(view, camera);
+            this.targetWorldPosition.setFromMatrixPosition(this.seaLevel.matrixWorld);
+            if (params.callBack) {
```
zarov: Change to `callback`

zarov (commented):
Looking good, have some comments though. I may have missed some things !

Concerning the `syncCameras` example, there seems to be some problems with it. First, you can see latency between the synchronization of the two views. Second, if you move the left view, and move the right after, it will revert back quickly to the initial state of the right view.

For testing files, you can drop the `_test` and `_unit_test` suffix.

test/unit/CameraUtils_unit_test.js
gchoqueux: I must to overload the function `getPickingPositionFromDepth` because it doesn't work properly. Because Chrome in headless mode doesn't support the WebGL EXT_frag_depth extension.

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: I move it and i changed to function to get the states.

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: yes I would see that in another PR

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: because they are used

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: but there's `this.dispatchEvent(this.startEvent)`

gchoqueux: yes surely, but I would do it later

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: yes surely, but I would do it later

src/Renderer/ThreeExtended/GlobeControls.js
gchoqueux: `cameraTransformOptions` is defined in CameraUtils

src/utils/CameraUtils.js
gchoqueux: if angle is -181, it doesn't work
we are getting
 * `(angle % 360) - 180 // => -361`
 * `angle - Math.floor((angle + 179.0) / 360) * 360 // => 179`

src/utils/CameraUtils.js
gchoqueux: see  in cameraTransformOptions's documentation

test/unit/CameraUtils_unit_test.js
zarov: Weird, but ok

src/Renderer/ThreeExtended/GlobeControls.js
zarov: Ok

src/Renderer/ThreeExtended/GlobeControls.js
zarov: I'm not sure about that, there is nothing in the `else` below.

src/Renderer/ThreeExtended/GlobeControls.js
zarov: You forgot to remove it L711

src/Renderer/ThreeExtended/GlobeControls.js
zarov: Yep sorry !

gchoqueux:
@zarov  Add globeControls's documentation and new example: globe travel

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -279,83 +132,43 @@ export const CONTROL_EVENTS = {
...
- * @param {*} target
- * @param {number} radius
- * @param {options} options
+ * GlobeControls is camera controller
+ *
+ * @class      GlobeControls
+ * @param      {GlobeView}  view the view where the control will be used
+ * @param      {Coordinates}  targetCoordinate the target looked by camera, at initialization
+ * @param      {number}  range distance between the target looked and camera, at initialization
+ * @param      {number}  globeRadius The globe'sradius
```
zarov: Space

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -279,83 +132,43 @@ export const CONTROL_EVENTS = {
...
-/* globals document,window */
+let previous;
 
 /**
- * @class
- * @param {GlobeView} view
- * @param {*} target
- * @param {number} radius
- * @param {options} options
+ * GlobeControls is camera controller
```
zarov: Missing a `a`

src/Renderer/ThreeExtended/GlobeControls.js
```diff
@@ -1678,18 +1036,28 @@ GlobeControls.prototype.getCameraOrientation = function getCameraOrientation() {
  * @return {Coordinates} position
  */
 
-GlobeControls.prototype.getCameraLocation = function getCameraLocation() {
+GlobeControls.prototype.getCameraCoordinate = function _getCameraCoordinate() {
```
zarov, line 1029: The attached fiddle is no longer relevant, delete it or update it. Same for other fiddle in this file.

zarov (commented):
Cool to see an animated example ! Few typo in the documentation and it's good to go

gchoqueux:
rebased and updated PR

zarov (approved):
I think we can finally mark it as good to merge, good work ! Please consider removing the jsfiddle link from the documentation of `GlobeControls` before merging

src/utils/CameraUtils.js
mbredif: `angle - Math.floor((angle + 179.0) / 360) * 360` is not good either (as expected with a magic number 179 degrees !):  angle =180.5 returns 180.5 instead of -179.5
A correct formula would be `angle - Math.floor((angle + 180.0) / 360) * 360`

gchoqueux: thanks

Issue #731, opened 2018-04-16
Performance, very slow (5 FPS), in globe WFS extruded example, since hole support


gliegard:
The globe WFS extruded example is very slow since hole support.

## Steps to Reproduce (for bugs)
1. Start FPS Meter on Chrome : console -> Rendering -> FPS meter
2. Try the code before hole support
git checkout 2de6ba7a66b5332dec0bc35cb067eeae50818f8d
3. Check the WFS Extruded example.
It shows between 40 and 60 FPS
4. Try the code after the hole support
git checkout a212dde70542b2a4480eb064df45820cad847363
5. WFS Extruded example shows between 3 and 8 FPS

## Expected Behavior
The examples should give a good user experience.

## Actual Behavior
The globe WFS extruded example is very hard to use, the computer seems to freeze.

gliegard:
ping @peppsac

peppsac:
Could you use the profiler to find out the culprit?

gchoqueux:
@peppsac 
The @gliegard's issue isn't a reason to assign him the work to fix it. 
or maybe you just wanted to know if the profiler has been used ?

peppsac:
> maybe you just wanted to know if the profiler has been used

@gchoqueux yes, I only suggested ("could you?") that he took a quick look at the profiler to see if the issue is obvious.

gchoqueux:
Since the PR #691, each feature have one mesh, right?
suggestions of culprit :
- too many drawcalls
- resolve ray casting

gliegard:
I think the problem is the number of meshes :
- Before PR #691 , the globe WFS extruded example had to create 9 meshes for all the buildings.
- Now, it creates 8000 meshes for the same example.

Another thing, your commit a212dde70542b2a4480eb064df45820cad847363 
- do the hole support, and thanks for this work, it's great !
- remove the geometry merging (I think the problem is here)

It's shame that there is only one commit for these 2 features.
If there was one commit for the 'geometry merge', I would revert it in a branch to test performance.

@peppsac What do you think about that ? Can we go back to have less meshes to draw easily ?

peppsac:
Thanks for your suggestions of what the problem might be.

A profiling capture would be more helpful though because I don't reproduce the problem here (maybe my machine is faster? maybe this only happens on Windows? maybe I'm not doing the same test as you?).

gliegard:
> Thanks for your suggestions of what the problem might be.

[...]

PR #798, opened 2018-06-29
Refacto provider and add Sources

gchoqueux:
## Description
add `Source` in layer : all the parameters to get the data from a source
add `convert` in layer : fonction to convert data from source to layer's elements

The  layer does
```js
const layer = {
  id, 
  type,
  source,
  convert,
}
```
The  provider does
``` js
function executeCommand() {
      const extent = command.extent; //  the datas's extent  to fetch
      const dataFetchedFromSource = await fetcher(source.urlFromExtent(extent));
      const dataParsed = await parse(dataFetchedFromSource);
      const layerElement = convert(dataParsed);
      return layerElement;
}
```
## Motivation
 * Isolate **`sources`** and **`converters/parsers`** from providers and use them for other layers 
 * simplify providers and merge certain providers (WMTS, WMS, Rasterizer), see delete them


src/Core/TileMesh.js
```diff
@@ -205,8 +204,9 @@ TileMesh.prototype.changeSequenceLayers = function changeSequenceLayers(sequence
 
 TileMesh.prototype.getCoordsForLayer = function getCoordsForLayer(layer) {
     if (layer.protocol.indexOf('wmts') == 0) {
```
mbredif: layer.source.protocol ?

src/Core/View.js
```diff
@@ -177,12 +184,15 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
 
     // probably not the best place to do this
     if (layer.type == 'color') {
+        layer.fx = layer.fx || 0.0;
```
mbredif: why not a defineLayerProperty as for other properties?

mbredif (commented):
It looks better than the current state IMHO.
Some remarks after a quick review : 
- `layer.source || layer` looks odd, is it temporary ? can't we get rid of it ?
- converts and parsers are 2 different things (eg: WMTS source -> Geojson parser -> Mesh converter)

I however feel that tiledCRS version of extents should be an implementation detail of quadtree-based sources and that other parts of itowns should only handle extents as a (BBox3 + CRS).

Likewise instead of relative pitches / offsetscales, we could use absolute coordinates (BBox2 + CRS) for textures so that downscaling a texture would be a noop and uvs in the shader would be well defined coordinates with a proper crs instead of normalized coordinates.

src/Parser/textureConverter.js
```diff
@@ -0,0 +1,20 @@
...
+const textureLayer = (texture, extent) => {
+    texture.coords = extent;
+    texture.generateMipmaps = false;
+    texture.magFilter = THREE.LinearFilter;
+    texture.minFilter = THREE.LinearFilter;
+    return texture;
+};
+
+export default {
+    textureColorLayer(texture, extent, layer) {
```
zarov: I wouldn't call this textureColorLayer. Why not have two methods, `color` and `elevation`, which gives `textureConverter.color` and `textureConverter.elevation` calls.

Also by introducing this you could drop the `getColor/XBilTextureByUrl` in `OGCWebServiceHelper` which have nothing to do with OGC.

src/Provider/Fetcher.js
```diff
@@ -1,14 +1,24 @@
-import { TextureLoader } from 'three';
+import { TextureLoader, DataTexture, AlphaFormat, FloatType, LinearFilter } from 'three';
```
zarov: My opinion: we should always import `* as THREE`, I don't get the appeal behind selecting what we are importing.

src/Provider/Fetcher.js
```diff
@@ -90,10 +99,12 @@ export default {
      *
      * @return {Promise}
      */
-    arrayBuffer(url, options = {}) {
-        return fetch(url, options).then((response) => {
-            checkResponse(response);
-            return response.arrayBuffer();
+    arrayBuffer,
+    textureFloat(url, options = {}) {
```
zarov, line 103: This part should be in the `elevationTexture` part imho. Let's keep the `Fetcher` simple.

src/Core/Scheduler/Scheduler.js
```diff
@@ -282,9 +282,9 @@ Scheduler.prototype.addProtocolProvider = function addProtocolProvider(protocol,
     if (typeof (provider.executeCommand) !== 'function') {
         throw new Error(`Can't add provider for ${protocol}: missing a executeCommand function.`);
     }
-    if (typeof (provider.preprocessDataLayer) !== 'function') {
-        throw new Error(`Can't add provider for ${protocol}: missing a preprocessDataLayer function.`);
-    }
+    // if (typeof (provider.preprocessDataLayer) !== 'function') {
```
zarov: :+1:

src/Core/View.js
```diff
@@ -139,17 +142,22 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
...
+            options: {},
         };
     }
 
-    if (provider) {
+    layer.style = layer.style || {};
+
+    // temp
+    // devrait etre supprime et placer dans source
+    if (provider && !layer.source) {
```
zarov: Yep, a provider shouldn't be the one defining the limit.

src/Core/View.js
```diff
@@ -166,6 +174,15 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
             if (!(providerPreprocessing && providerPreprocessing.then)) {
                 providerPreprocessing = Promise.resolve();
             }
+        } else if (layer.source) {
```
zarov, line 193: I don't like having this here, as much as I don't like having all other things below (conditions on type of layer). I know though that you hope to move it to the layer itself, so I'm ok with this here for now.

src/Core/View.js
zarov: `defineLayerProperty` is a bad practice imho, and achieves nothing besides reducing a little the duplicated code, executing a callback in really rare usecases and emitting events on properties change. But I wouldn't change it here, and in the meantime keep this property declaration as it is.

zarov (commented):
I think this is a good idea. The introduction of a `Source` object (kind of like OpenLayers ?) is a good step towards decluttering providers. Could you add a more accurate description (maybe documentation like the current `Provider` interface ?) to this object ?

However I don't really like a few things:
- the fetching in `geoFileSource`: no operation like should happen in a `Source`
- the `convert` method in `layer`: what about your initials thoughts with `source, processing, style` only ?

peppsac:
I'm all for refactoring this part, and the chosen solution looks similar to what was previously discussed.
(the only downside is that we already have almost 30 open PR and that updating/reviewing/merging others would be more relevant)

Random notes:
- IMHO, processing can depend on the data type (points or textures) but should not depend on the source type
- provider (or source) implements the protocol specific knowledge and don't necesarly need to do network accesses themselves
- a big refactoring must like this one:
  * not lose features (obviously)
  * bring new features (not necessarly in the same PR) to show its usefulness.

The last point is important. For instance, if this PR doesn't help implement vector-tiles support easily, then something is wrong.

nosy-b:
Sounds like the way to go!

index.html
```diff
@@ -49,14 +49,15 @@
             var promises = [];
             promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/Ortho.json').then(addLayerCb));
             promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/OrthosCRS.json').then(addLayerCb));
-            promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/ScanEX.json').then(addLayerCb));
+            // FIXME
+            // promises.push(itowns.Fetcher.json('examples/layers/JSONLayers/ScanEX.json').then(addLayerCb));
```
zarov: What is the problem here ?

src/Core/Geographic/Extent.js
```diff
@@ -68,11 +68,15 @@ Extent.prototype.clone = function clone() {
     }
 };
 
+Extent.prototype.isTiledCrs = function fnIsTiledCrs() {
+    return _isTiledCRS(this._crs);
```
zarov: You could move the code of `_isTiledCRS` directly here and call everywhere else in the file `this.isTiledCRS` only.

src/Core/Scheduler/Scheduler.js
```diff
@@ -132,21 +127,13 @@ Scheduler.prototype.constructor = Scheduler;
...
-    this.addProtocolProvider('xyz', TMSProvider);
     this.addProtocolProvider('potreeconverter', PointCloudProvider);
-    this.addProtocolProvider('wfs', WFSProvider);
-    this.addProtocolProvider('rasterizer', RasterProvider);
-    this.addProtocolProvider('static', StaticProvider);
 };
 
 Scheduler.prototype.runCommand = function runCommand(command, queue, executingCounterUpToDate) {
-    var provider = this.providers[command.layer.protocol];
+    var provider = this.getProtocolProvider(command.layer.protocol || command.layer.source.protocol);
```
zarov: Why move to `getProtocolProvider` ?

src/Core/TileMesh.js
```diff
@@ -141,12 +141,12 @@ TileMesh.prototype.updateGeometricError = function updateGeometricError() {
...
 };
 
-TileMesh.prototype.setTexturesLayer = function setTexturesLayer(textures, layerType, layerId) {
+TileMesh.prototype.setTexturesLayer = function setTexturesLayer(textures, layerType, layerId, pitchs) {
     if (this.material === null) {
         return;
     }
     if (textures) {
-        this.material.setTexturesLayer(textures, layerType, layerId);
+        this.material.setTexturesLayer(textures, layerType, layerId, pitchs);
```
zarov: What is the usage of `setTexturesLayer` ? Isn't it meant to be `setLayerTextures` ?

src/Core/TileMesh.js
```diff
@@ -172,41 +172,42 @@ TileMesh.prototype.changeSequenceLayers = function changeSequenceLayers(sequence
...
-                (layer.extent.crs() == 'EPSG:3857' || is4326(layer.extent.crs()))) {
+                (source.extent.crs() == 'EPSG:3857' || is4326(source.extent.crs()))) {
             OGCWebServiceHelper.computeTileMatrixSetCoordinates(this, 'PM');
             return this.wmtsCoords.PM;
         } else {
-            return OGCWebServiceHelper.computeTMSCoordinates(this, layer.extent, layer.origin);
+            return OGCWebServiceHelper.computeTMSCoordinates(this, source.extent, source.origin);
         }
-    } else if (layer.extent.crs() == this.extent.crs()) {
+        // FIXME: source.extent == undefined
```
zarov: What is the fix you have in mind ?

src/Parser/textureConverter.js
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            texture = Feature2Texture.createTextureFromFeature(data, extentTexture, 256, layer.style, backgroundColor);
+            texture.parsedData = data;
+            texture.coords = extentDestination;
+        } else if (data.isTexture) {
+            texture = data;
+        }
+
+        if (layer.type === 'color') {
```
zarov: What about putting this part in `ColorLayer` and `ElevationTexture` ?

src/Parser/GeoJsonParser.js
```diff
@@ -211,6 +211,7 @@ function readFeatures(crsIn, crsOut, features, filteringExtent, options) {
             res.features.push(f);
         }
     }
+    res.isFeature = true;
```
zarov, line 214: :+1:

src/Layer/TiledGeometryLayer.js
```diff
@@ -123,6 +124,31 @@ class TiledGeometryLayer extends GeometryLayer {
         }
     }
 
+    convert(requester, extent) {
+        return convertToTile.convert(requester, extent, this);
+    }
+
+    buildRootTile(view, scheduler) {
```
zarov: Should be in the constructor imho.

src/Parser/VectorTileParser.js
```diff
@@ -4,17 +4,18 @@ import GeoJsonParser from './GeoJsonParser';
 
 function readPBF(file, options) {
     const vectorTile = new VectorTile(new Protobuf(file));
-
+    const extentSource = options.extentSource || file.coords;
     const layers = Object.keys(vectorTile.layers);
 
     if (layers.length < 1) return;
 
+    const crsInId = Number(options.crsIn.slice(5));
```
zarov, line 12: I don't understand the point of this. How can you be sure you got the correct value ?

src/Process/FeatureProcessing.js
```diff
@@ -44,12 +55,15 @@ export default {
         }
 
         const features = node.children.filter(n => n.layer == layer);
+        // FIXME: traverse is do for each frame in each object3D
```
zarov, line 58: It is indeed a big problem, and could be a partial response to #731 thanks for discovering this.

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -326,29 +349,42 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
...
         priority: nodeCommandQueuePriorityFunction(node),
         earlyDropFunction: refinementCommandCancellationFn,
     };
 
     return context.scheduler.execute(command).then(
-        (terrain) => {
+        (textures) => {
+            const terrain = { texture: textures[0] };
+
+            // deja verfier dans le provider
```
zarov: So does it mean you are going to remove things here ?

src/Provider/DataSourceProvider.js
```diff
@@ -0,0 +1,135 @@
+import GeoJsonParser from '../Parser/GeoJsonParser';
+import VectorTileParser from '../Parser/VectorTileParser';
+import Fetcher from './Fetcher';
+import Cache from '../Core/Scheduler/Cache';
+import CancelledCommandException from '../Core/Scheduler/CancelledCommandException';
+
+const supportedFetchers = new Map([
```
zarov: Maybe we could agree on a default fetcher ? The default parser being none, I suggest that `Fetcher.texture` to be the default.

src/Provider/DataSourceProvider.js
```diff
@@ -0,0 +1,135 @@
...
+}
+
+const supportedParsers = new Map([
+    ['geojson', GeoJsonParser.parse],
+    ['application/json', GeoJsonParser.parse],
+    ['application/x-protobuf;type=mapbox-vector', VectorTileParser.parse],
+    [true, noParsingNeeded],
+]);
+
+// Todo: move this method to layer
```
zarov: Why ?

src/Provider/DataSourceProvider.js
```diff
@@ -0,0 +1,135 @@
...
+        .then(parsedData =>
+    // Convert parsed data, it converts itowns's object to THREE's object
+            layer.convert(parsedData, extentDestination),
+            (err) => {
+                source.handlingError(err, url);
+                throw err;
+            });
+}
+
+function extentToString(extent) {
```
zarov: Should be a `toString` method of `Extent`.

src/Source/WMSSource.js
```diff
@@ -0,0 +1,76 @@
+import Source from './Source';
+import URLBuilder from '../Provider/URLBuilder';
+
+const supportedFormats = [
```
zarov: Is this still relevant ?

src/Source/fileSource.js
```diff
@@ -0,0 +1,88 @@
...
+        const west = bound.getAttribute('minlon');
+        const east = bound.getAttribute('maxlon');
+        const south = bound.getAttribute('minlat');
+        const north = bound.getAttribute('maxlat');
+        return new Extent('EPSG:4326', west, east, south, north);
+    }
+    return new Extent('EPSG:4326', -180, 180, -90, 90);
+}
+
+// TODO move and refacto
```
zarov: I'm OK to let it as it is for now, I have plan for this part.

test/unit/layeredmaterialnodeprocessing.js
```diff
@@ -28,18 +28,23 @@ describe('updateLayeredMaterialNodeImagery', function () {
...
     };
 
     beforeEach('reset state', function () {
         // clear commands array
         context.scheduler.commands = [];
         // reset default layer state
-        layer.tileInsideLimit = () => true;
-        layer.visible = true;
+
         layer.updateStrategy = STRATEGY_MIN_NETWORK_TRAFFIC;
```
zarov: Side note, but this should be `layer.updateStrategy = { type: STRATEGY_MIN_NETWORK_TRAFFIC }`.

zarov (changes requested):
I couldn't leave a comment on the file: the screenshot `wfs_color` is not used, please remove it.

Otherwise it looks good, we are heading a path I like. I saw a lot of `TODO`, what are your plans here ? Could you list the one you would do later (ie in another PR) ?

index.html
gchoqueux: We can add all the layers without breaking the globe. I will fix it

src/Core/Geographic/Extent.js

src/Core/Scheduler/Scheduler.js
gchoqueux: Encapsulation was already existing, it is better to use it. 
We can return a default value with this fonction

src/Core/TileMesh.js
gchoqueux: `Mesh.setTexturesLayer` and `this.material.setTexturesLayer` doesn't exist anymore

src/Core/TileMesh.js
gchoqueux: I revert this part, it's unnecessary

src/Parser/textureConverter.js
gchoqueux: I don't like the actual textureConverter. It do too much things. I wan't add THREE.js code in layer. I could delete the layer parameter and have a more generic fonction.

src/Parser/VectorTileParser.js
gchoqueux: the crs is always `EPSG:XXXX`, and there's 5 characters in `EPSG:`

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: this check is maybe useless, see in `dataSourceProvider`

src/Provider/DataSourceProvider.js
gchoqueux: ok, how would you do this?

src/Provider/DataSourceProvider.js
gchoqueux: this function should be layer's callback in function of data

src/Provider/DataSourceProvider.js
gchoqueux: ok

src/Source/WMSSource.js
gchoqueux: I moved this part in `Source`. But, I think it's not respond to the question.
could you precise?

test/unit/layeredmaterialnodeprocessing.js
gchoqueux: yes

src/Source/fileSource.js
gchoqueux: what is your plan?

src/Process/LayeredMaterialNodeProcessing.js
zarov: It seems indeed. Maybe let it as it is for now and we'll see to remove that when refactoring this processing.

src/Provider/DataSourceProvider.js
zarov: Something simple like adding a `['default', Fetcher.texture]` entry to this map, and a simple check in the `fetchData` below. Even if you don't add a default fetcher, you need to catch the exception for a not supported format, as there seems to be no safe guard for now.

src/Source/WMSSource.js
zarov: No indeed, I was thinking that we could get rid of this array ? Even if the WMS spec says that files should be images, I think we can get rid of it.

src/Source/fileSource.js
zarov: Drop this function and add some parser, kinda like in #766

src/Layer/TiledGeometryLayer.js
gchoqueux: yes, I move it in constructor

zarov:
I tested the examples, here's what I saw:
- __cubic_planar__: there seems to have artifact on the sides, see those two comparisons screenshots (first actual version, second this PR) 
image
image
- __globe_vector__: picking is not working anymore
- __syncCameras__: not working, I have a `Cannot read property 'protocol' of undefined`

gchoqueux:
@zarov all examples are fixed

New change:
 - root tile are build in constructor
 - `Layer.projection` are determined in `addLayer`

src/Source/Source.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+/**
+ * Source are parameters to fetch source
+ *
+ * @class Source
+ */
+class Source {
+    /**
+     * @param  {sourceParams}  source  object to set source
```
zarov: You should remove the `@class` declaration above, and write a `@constructor` here, to avoid duplication in the generated page of the documentation. 

You should also write an example of declaring a random Source. Best would be to have it for all Sources files.

src/Source/FileSource.js
```diff
@@ -0,0 +1,100 @@
...
+     * File source to use file in {@link GeometryLayer}, {@link ColorLayer} or {@link ElevationLayer}.
+     * @constructor
+     * @extends Source
+     *
+     * @param      {sourceParams}  source  The source
+     * @param      {string}  crsOut  crd output data
+     */
+    constructor(source, crsOut) {
+        super(source);
+        if (!source.url) {
```
zarov: No need to repeat this as it is already been checked in `Source`.

src/Source/FileSource.js
```diff
@@ -0,0 +1,100 @@
...
+            throw new Error('source.url is required in FileSource');
+        }
+
+        if (!source.projection) {
```
zarov: You should specify in the documentation above that the projection is mandatory for this type of Source. This check can also be moved before calling `super(source)`.

src/Source/Source.js
```diff
@@ -0,0 +1,104 @@
...
+    handlingError(err) {
+        console.warn(`err ${this}`, err);
+    }
+    /**
+     * Generate url from extent. This url is link to fetch data inside the extent.
+     *
+     * @param      {Extent}  extent  extent to convert in url
+     * @return     {string}  url from extent
+     */
+    // eslint-disable-next-line
```
zarov, line 66: This is a problem I encountered as well, there is three solutions:
- `eslint-disable-next-line`
- declaring this method as `static`: `static urlFromExtent(extent)`, that can still be called with `this.urlFromExtent(extent)` or `source.urlFromExtent(extent)`
- removing the `class-method-use-this` eslint check from the configuration

src/Source/StaticSource.js
```diff
@@ -0,0 +1,131 @@
...
+     * Images source to panorama layer {@link PanoramaLayer}
+     *
+     * @constructor
+     * @extends Source
+     * @param {sourceParams}  source  The source
+    */
+    constructor(source) {
+        super(source);
+
+        if (!source.extent) {
```
zarov: Same as above, you can move this check before calling the parent constructor and specify in the documentation that the `extent` is required here.

src/Source/WFSSource.js
```diff
@@ -0,0 +1,70 @@
...
+     * @constructor
+     * @extends Source
+     *
+     * @param {sourceParams}  source
+     * @param {string} source.typeName typeName of wfs
+     * @param {string} [source.version='2.0.2'] wfs protocol version
+     * @param {Object} [source.zoom]
+     * @param {number} [source.zoom.min] layer's zoom minimum
+     * @param {number} [source.zoom.max] layer's zoom maximum
+     * @param {string} [source.zoom.max] layer's zoom maximum
```
zarov: Did you mean something else ?

src/Source/WFSSource.js
```diff
@@ -0,0 +1,70 @@
...
+    */
+    constructor(source) {
+        super(source);
+        if (!source.typeName) {
```
zarov: Same, move the `super` below

src/Source/WMSSource.js
```diff
@@ -0,0 +1,86 @@
...
+            // EPSG 4326 x = lat, long = y
+            // version 1.1.0 long/lat while version 1.3.0 mandates xy (so lat,long)
+                this.axisOrder = (this.version === '1.1.0' ? 'wsen' : 'swne');
+            } else {
+            // xy,xy order
+                this.axisOrder = 'wsen';
+            }
+        }
+
+        let crsPropName = 'SRS';
```
zarov: Why not directly `const crsPropName = (this.version === '1.3.0') ? 'CRS' : 'SRS';` ?

src/Source/WMTSSource.js
```diff
@@ -0,0 +1,100 @@
+import Source, { supportedFormats } from './Source';
+import URLBuilder from '../Provider/URLBuilder';
+
+
+/**
+ * Options to wtms protocol
+ * @typedef {Object} OptionsWmts
```
zarov: Will it be detailed ?

zarov (commented):
Added some comments on the Sources files !

src/Provider/DataSourceProvider.js
gchoqueux: I do something for not supported format

src/Source/WMSSource.js
gchoqueux: removed

src/Source/FileSource.js
gchoqueux: yes

src/Source/Source.js
gchoqueux: yes, there are errors lint too for jsdoc

gchoqueux:
@zarov PR updated

zarov (approved):
OK let's validate this, I think I didn't see any problem.

For fixing your failing test, either remove it, or install a polyfill for `URL` (like this one https://github.com/webcomponents/URL) as it is not a component supported in nodejs.

zarov:
It is even occurring because the `URL` component has only been added in node 10.

PR #799, opened 2018-07-02
fix(3dtiles): dont stop subdivision at tileset boundaries

peppsac:
In this scenario:
  - a tileset references an external tileset at node N
  - the subdivision algorithm decides to stop at N and display N
We must go one step further and display the content of the root
node R from the external tileset.

The reason is that R and N are the same node regarding content,
but are stored in different tilesets (for technical reasons)  and
thus considered as different node in iTowns (with N having R as
its only child).

Here's an example of the bug.

This is the same pointcloud as we use for the pointcloud.html example, but converted as 3dtiles. I forced the first nodes to be external tilesets to easily see the problem.
The top part is master, the bottom part is with this PR applied:

3dtiles_external_tileset_bug

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -324,6 +324,9 @@ export function $3dTilesSubdivisionControl(context, layer, node) {
     if (layer.tileIndex.index[node.tileId].children === undefined) {
         return false;
     }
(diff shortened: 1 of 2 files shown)


autra (approved):
LGTM thanks !

PR #801, opened 2018-07-06
ci(test): add unit test coverage report.

gliegard:
Add unit test coverage report
## Description
- The command 'npm run test-with-coverage' create a report in ./coverage/index.html
- Travis is set up to send coverage report to Coveralls.io.
- The coverage evolution is visible on Github for each PR.

## Motivation and Context
It's not so easy to write tests and to be sure that the code is really tested.
Coverage report show us if all part of the code is tested (each branch of the IF ..)

During dev, we can see coverage report locally

coverage0

And the file details, for example Extent.js

coverage0-1

During code review, we can also see coverage evolution metric.
Here is the example of a PR where I remove Coordine unit test.

coverage1

Clicking on detail, we can see in Coveralls where the code coverage decrease

coverage2

Then we can look at the file

coverage3


peppsac (commented):
Looks like a good idea.

It's a bit depressing that it requires to add 2300 new lines to the package-lock.json, but I guess that's how js dev works.
(e.g: the `right-align` package added to the dependency, is a 2 lines wrapper to the `text-align` package...)

zarov:
Yay for this ! Shouldn't we take this opportunity to move the unit tests in a `unit_test` folder ?

> It's a bit depressing that it requires to add 2300 new lines to the package-lock.json, but I guess that's how js dev works.

I agree, and I think we could chose to remove it from the repo (there are lots of hot discussion about this on the internet ;) )

zarov:
@gliegard could you rebase (you need to remove `node-fetch` I think) ? I think we could merge this soon.

zarov (approved):
LGTM, I'd love to see this added, thus nobody can say "I don't want to add a test in this PR" ;)

Wait for at least another approval before merging !

package.json
```diff
@@ -7,9 +7,11 @@
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\"",
     "doc": "jsdoc -t node_modules/docdash --readme JSDOC.md -c jsdoc-config.json",
     "doclint": "npm run doc -- -t templates/silent",
-    "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-examples && npm run test-unit",
+    "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-examples && npm run test-with-coverage",
     "test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --require babel-core/register test/*unit_test.js",
     "test-examples": "mocha -t 300000 --recursive test/examples",
+    "test-with-coverage": "nyc -n src -r html cross-env BABEL_DISABLE_CACHE=1 mocha --require babel-core/register test/*unit_test.js",
```
autra: `nyc -n src -r html cross-env npm run test-unit` does not work?

.travis.yml
```diff
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - "lts/*"
   - "node"
+after_script:
+- "npm install coveralls && npm run deploy-coverage"
```
autra: If we have a script in `package.json` using coveralls, we need to add it to `devDependencies` instead of here.

package.json
tbroyer: Could this just call `npm run test-unit`? (I suppose not, otherwise you'd done it; just checking that it's not an oversight)

.travis.yml
```diff
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - "lts/*"
   - "node"
+after_script:
```
autra: I'd say `after_success`? WDYT?

package.json
```diff
@@ -7,9 +7,11 @@
...
+    "deploy-coverage": "nyc report --reporter=text-lcov | coveralls",
```
tbroyer: I'm a bit concerned that this uses a tool that's not listed as a dependency (too bad there's no "optional dev dependencies").

Could maybe the Travis build be rewritten to use `npx nyc report --reporter=txt-lcov | npx coveralls` instead of `npm run deploy-coverage`?

autra: Do we need to configure our build process for that? ([the doc](https://docs.coveralls.io/supported-ci-services) is a bit unclear on that). Let's check this before merging (just to be sure not to break travis in the meantime). Maybe the best way is to test in your fork?

autra (commented):
This is a nice addition and I'd be happy to have a coverage tool in our repo... providing we don't make the coverage percentage an end goal :-) It's just another (very useful) tool for a better stability.

Thanks @gliegard for this work

.travis.yml
gliegard: ok, I've moved it to after success

package.json
gliegard: ok, I've remove the line from package.json, and move it to .travis.yml

package.json
gliegard: yes, it works, I've changed it

.travis.yml
gliegard: Since this dependency is only for Travis to publish the report to Coveralls, it's not needed to add it in devDependencies. And I've remove the line from package.json.

autra:
> Do we need to configure our build process for that? (the doc is a bit unclear on that). Let's check this before merging (just to be sure not to break travis in the meantime). Maybe the best way is to test in your fork?

So I tested in mine. Apparently it works out of the box for open-source project. I've asked admin of the itowns organization to grant access to coveralls.

autra:
> I've asked admin of the itowns organization to grant access to coveralls.

And it's done. I'm not sure it's done correctly though. I feel it still goes through my account to access it. Not sure about that.

@gliegard if you feel like it, you can add the badge in the README.md

```md
[![Coverage Status](https://coveralls.io/repos/github/iTowns/itowns/badge.svg?branch=test_coverage)](https://coveralls.io/github/iTowns/itowns?branch=test_coverage)
```
that looks like [Coverage Status](https://coveralls.io/github/iTowns/itowns?branch=test_coverage)

.travis.yml
autra: Yep, I've seen your other comment. It's also fine by me :+1:

gliegard:
already done ;) https://github.com/iTowns/itowns/pull/801/files#diff-04c6e90faac2675aa89e2176d2eec7d8R6

tbroyer (approved):

package.json
gliegard: As it works for open source projects, It already works for this PR :
https://coveralls.io/builds/17969819

autra:
Ok, just a last nitpick for the commit message: we don't use `ci`, so I guess it should be `ĉhore` instead? Thanks!

gliegard:
ok, i've updated the commit msg

autra:
Let's rebase and get rid of the merge commit, and I think it can be landed :-)

autra:
Thanks!

PR #802, opened 2018-07-10
docs(3dtiles): add 3dtiles layer documentation

peppsac:
This is a proposal to document the type of layers and their protocol specific options.

I think it would work well for all *Provider files.

WDYT?

### DIFF
--- a/src/Provider/3dTilesProvider.js
+++ b/src/Provider/3dTilesProvider.js
@@ -7,6 +7,29 @@ import Extent from '../Core/Geographic/Extent';
 import { pre3dTilesUpdate, process3dTilesNode, init3dTilesLayer } from '../Process/3dTilesProcessing';
 import utf8Decoder from '../utils/Utf8Decoder';
 
+/**
+ * iTowns supports loading and displaying 3d-tiles tilesets.
+ *
+ *
+ * @module 3d-tiles
+ *
+ * @example
+ * var layer = new itowns.GeometryLayer('3dtiles-example');
+ * layer.protocol = '3d-tiles'
+ * layer.url = 'http://example/tileset.json';
+ *
+ * @property {string} url - tileset.json URL
+ * @property {boolean|THREE.Material} [overrideMaterial=false] - override meshes material
+ * embedded in the binary files and use a default one instead.
+ * @property {THREE.Material} [material=THREE.PointsMaterial] - material cloned
+ * and assigned each time a points mesh is created.
+ * @property {number} [sseThreshold=16] - s(creen) s(pace) e(rror) threshold in pixels.
+ * Define how many pixels the geometricError from a tile must cover to kick the
+ * subdivision mechanism in.
+ * @property {number} [cleanupDelay=1000] - delay in milliseconds after which an
+ * undisplayed tiles will be removed from memory.
+ *
+ */
 
 export function $3dTilesIndex(tileset, baseURL) {
     let counter = 1;
--- a/jsdoc-config.json
+++ b/jsdoc-config.json
@@ -21,6 +21,7 @@
             "src/Parser/GpxParser.js",
 
             "src/Provider/URLBuilder.js",
+            "src/Provider/3dTilesProvider.js",
 
             "src/Renderer/ColorLayersOrdering.js",
             "src/Renderer/ThreeExtended/Feature2Mesh.js",


src/Provider/3dTilesProvider.js
```diff
@@ -7,6 +7,37 @@ import Extent from '../Core/Geographic/Extent';
...
+ * @module 3d-tiles
+ *
+ * @example
+ * var layer = new itowns.GeometryLayer('3dtiles-example');
+ * layer.preUpdate = preUpdateGeo;
+ * layer.update = itowns.process3dTilesNode(
+ *     itowns.$3dTilesCulling,
+ *     itowns.$3dTilesSubdivisionControl
+ * );
+ * layer.url = 'http://example/tileset.json;
```
autra: :-O

peppsac:
I also simplified the 3dtiles usage by defining sane default value for layer.

Now creating a 3dtiles layer only requires to define the `url` and the `protocol`.

src/Provider/3dTilesProvider.js
```diff
@@ -4,9 +4,35 @@ import PntsParser from '../Parser/PntsParser';
...
-import { init3dTilesLayer } from '../Process/3dTilesProcessing';
+import { pre3dTilesUpdate, process3dTilesNode, init3dTilesLayer } from '../Process/3dTilesProcessing';
 import utf8Decoder from '../utils/Utf8Decoder';
 
+
+/**
+ * iTowns supports loading and displaying 3d-tiles tilesets.
+ *
+ *
+ * @module 3d-tiles
```
gchoqueux, line 14: * What is the structure of instanced objects for 3dTiles (THREE.js and itowns) and not the specification. 

for example in node tile :

```js
tileA  // (THREE.Mesh->itowns.tileMesh)
     {
          children: [tileB, tileC] // (Array) tile children
          parent: tile0  // (THREE.Mesh->itowns.tileMesh) tile parent 
          obb: // (itowns.OBB)  oriented bouding box
     }
```
in `3dTiles`, how are the objects linked: `content, parent, bounding Volume, region, transform ...`

peppsac:
Rebased to remove the commit merged in #811.

peppsac:
Here's the result in the documentation side-bar:

capture d ecran de 2018-07-19 14-44-30

I think @autra has concern about using `@module` for something that's not really a module so I'd be happy to hear (and see!) other proposals.

The objective here is to provide an easy to find documentation for each protocol-specific options.

src/Provider/3dTilesProvider.js
```diff
@@ -7,6 +7,29 @@ import Extent from '../Core/Geographic/Extent';
...
+ *
+ * @example
+ * var layer = new itowns.GeometryLayer('3dtiles-example');
+ * layer.protocol = '3d-tiles'
+ * layer.url = 'http://example/tileset.json';
```
gchoqueux, line 19: add `layer.type = 'geometry';`

src/Provider/3dTilesProvider.js
```diff
@@ -7,6 +7,29 @@ import Extent from '../Core/Geographic/Extent';
...
+ *
+ * @property {string} url - tileset.json URL
+ * @property {boolean|THREE.Material} [overrideMaterial=false] - override meshes material
+ * embedded in the binary files and use a default one instead.
+ * @property {THREE.Material} [material=THREE.PointsMaterial] - material cloned
+ * and assigned each time a points mesh is created.
+ * @property {number} [sseThreshold=16] - s(creen) s(pace) e(rror) threshold in pixels.
```
gchoqueux, line 26: would you add which sse method is used?

src/Provider/3dTilesProvider.js
```diff
@@ -7,6 +7,29 @@ import Extent from '../Core/Geographic/Extent';
...
+ * Define how many pixels the geometricError from a tile must cover to kick the
+ * subdivision mechanism in.
+ * @property {number} [cleanupDelay=1000] - delay in milliseconds after which an
```
gchoqueux, line 29: there's also not a random draw for node deletion?

gchoqueux (changes requested):

zarov:
I'll join @autra here. This may be better to not document each protocol using jsdoc, but maybe in a markdown file, and include it during the generation of the documentation, like we include the readme in the first page. Thus we could have a section in the menu like `Protocols`.

mbredif:
I think that is a sign that our protocol implementation is odd : with static functions modifying an external layer object  instead of constructing an object (a source?).

peppsac:
> This may be better to not document each protocol using jsdoc, but maybe in a markdown file, and include it during the generation of the documentation, like we include the readme in the first page.

I agree with this. But until then happens, the protocols options aren't documented.

> I think that is a sign that our protocol implementation is odd : with static functions modifying an external layer object instead of constructing an object (a source?).

Nope. That's only a sign that `jsdoc` can only handle class/object based API.


Anyway closing this PR, since the consensus seems to indicate that this is not the correct way to do it.

jailln:
Too bad that this job was left over; as a user it would have been great to have documentation on the layers :)

PR #803, opened 2018-07-12
Fix security vulnerabilities found in our current dependency tree.

autra:
## Description

npm has a new "audit" scripts. For the first 2 commits, I merely used `npm audit fix` twice.

For the second, by default `npm audit fix` won't upgrade the MAJOR number of a version, so I did it by hand. I didn't upgrade to mocha 5, because according to their changelog, they are struggling quite a bit with regression on their `--watch` option. v4 is enough to fix the vulnerabilities.

Note that this is imo a good reason to keep a package-lock.json in our repo.

@iTowns/reviewers please r?


tbroyer:
Given that NPM has changed the way it stores `requires` in the package-lock.json, I wonder if we shouldn't have a first commit with just the package-lock after an `npm install`, and then the changes due dependency updates.

tbroyer (approved):

autra:
> Given that NPM has changed the way it stores requires in the package-lock.json

oh yes. again.

> I wonder if we shouldn't have a first commit with just the package-lock after an npm install

doing that.

PR #804, opened 2018-07-12
Remove package-lock.json

gliegard:
## Description
- remove package-lock.json from version control
- ignore it from version control
- don't generate it with npm install

## Motivation and Context
I think we waste time managing modification in this file.

What do you think ?


autra (changes requested):
ok so cons:
- we're wasting time

pros:
- build reproducibility (everyone gets the same versions of dependencies of dependencies)
- time-machine (if you checkout an old version of the code, you're guaranteed to get the same environment people had at this time)
- security (we can force people to get secure dependencies, see #803 )

I know that npm has been changing the format a lot, but after all, `npm i` with the last npm version is all we need to do. And you're going to do it if you add dependencies anyway.

So I'm still in favor of keeping it :-)

autra:
(The red is my way of voting for the `question` tag)

tbroyer (changes requested):
I tend to agree with @autra 

On the other hand, if there's literature about why it's a bad idea to commit the lock file; please share it so we can possibly change opinion in face of well-argued reasoning.

autra:
> On the other hand, if there's literature about why it's a bad idea to commit the lock file; please share it so we can possibly change opinion in face of well-argued reasoning.

Totally! I'm just restating what I have heard on the Internet... (and the reason I committed this file in the first place)

tbroyer:
For example, without a package lock, you'd never know whether you'd use eslint-scope 3.7.1, or the 3.7.2 that contains a malware: https://github.com/eslint/eslint-scope/issues/39 (our package-lock says we apparently don't depend on `eslint-scope`, so we're safe)

gliegard:
Ok, I understand.

I'm not an expert of NPM, but for example, in this commit https://github.com/iTowns/itowns/pull/801/commits/569bc41f299ceddef79f96be01ff6ab02b5d73e2 , I add 'nyc' to the dependencies, then I run npm install with npm 5.6.0, and all the versions of all the dependencies are changed. As package-lock.json is in the version control, I commit it.

My "problem" is to commit modifications from a generated file (modifications that I don't really understand).

Is it the normal way to use npm and the package-lock.json ?
If you do 'npm install' with another version of npm, all the version will change again ?

autra:
> I'm not an expert of NPM, but for example, in this commit 569bc41 , I add 'nyc' to the dependencies, then I run npm install with npm 5.6.0, and all the versions of all the dependencies are changed. As package-lock.json is in the version control, I commit it.

That's because you were the first one to generate it with the latest version of npm, and npm *did* change the way it stores the versions in the `package-lock.json`. A good explanation for the rationale ~~https://github.com/npm/npm/issues/20891#issuecomment-395591772~~ CORRECT LINK: https://github.com/npm/npm/issues/20434#issuecomment-402988364.

But #803 did the same (see the first commit). Now that it is merged, if you rebase #801, you should see that the diff on `package-lock.json` is reduced. (for the conflict on `package-lock.json` you *will* have when rebasing, `git checkout --theirs package-lock.json` will take your version)

>  If you do 'npm install' with another version of npm, all the version will change again ?

If it is an old version, yes. But PR that touches `package-lock.json` should always be generated with the latest version imho.

> Is it the normal way to use npm and the package-lock.json ?

It's the recommended way of using it (at least by the npm team), yes.

autra:
Sorry didn't mean to close this.

autra:
And my link explaining the rationale of the change is wrong, it should be https://github.com/npm/npm/issues/20434#issuecomment-402988364

Time to call it a day I guess :-)

gliegard:
Ok, thank for the explanation.

#801 was rebased, and it reverts the modification of #803 because I use an old version of npm (5.6.0).

@autra Which version of npm are you using ? if I understand, you are using npm v6 ...
So we should all use the same version of NPM ?

autra:
> @autra Which version of npm are you using ? if I understand, you are using npm v6 ...
So we should all use the same version of NPM ?

Yes you're right, I use npm v6. I think it's better if we all use the more recent stable version to generate these `package-lock.json` (because it's a clear and easy-to-follow rule, mainly). I hope this restriction will be unnecessary in the future, when the format of this file will be more stable.

gliegard:
ok, and it's easy to have the latest version of npm with

sudo npm install npm -g

so meta :)

I close so PR

PR #805, opened 2018-07-13
test: move unit test to specific folder

zarov:
Something I wanted to do for some time, nothing special here, only reorganization.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "doc": "jsdoc -t node_modules/docdash --readme JSDOC.md -c jsdoc-config.json",
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-examples && npm run test-unit",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "doc": "jsdoc -t node_modules/docdash --readme JSDOC.md -c jsdoc-config.json",
     "doclint": "npm run doc -- -t templates/silent",
     "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-examples && npm run test-unit",
-    "test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --require babel-core/register test/*unit_test.js",
+    "test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --require babel-core/register test/unit",
```
peppsac, line 11: Would you mind changing this as well to `unit-test` or `unit-tests`?

zarov: I guess why not but as it is a subtask of `test`, it makes sense to keep the `test-` prefix.

peppsac (approved):
Thanks for removing all the /* global descibe, it */ :)

PR #806, opened 2018-07-13
refacto(renderer): get mesh properties generically

zarov:
If we want to add further properties to a mesh layer, it is simpler to
have a generic method to read those property than to add them all by
hand and clutter Feature2Mesh.

Note: this is more of a utility method that could be move in a `Transform/Utils` file later.

### DIFF
--- a/src/Renderer/ThreeExtended/Feature2Mesh.js
+++ b/src/Renderer/ThreeExtended/Feature2Mesh.js
@@ -1,43 +1,28 @@
 import * as THREE from 'three';
 import Earcut from 'earcut';
 
-function getAltitude(options, properties, contour) {
-    if (options.altitude) {
-        if (typeof options.altitude === 'function') {
-            return options.altitude(properties, contour);
-        } else {
-            return options.altitude;
-        }
-    }
-    return 0;
-}
+function getProperty(name, options, defaultValue, ...args) {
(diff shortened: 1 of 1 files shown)


src/Renderer/ThreeExtended/Feature2Mesh.js
```diff
@@ -143,8 +128,8 @@ function prepareBufferGeometry(vert, color, altitude, extrude) {
 
 function featureToPoint(feature, properties, options) {
     // get altitude / color from properties
-    const altitude = getAltitude(options, properties, feature.vertices);
-    const color = getColor(options, properties);
+    const altitude = getProperty('altitude', options, 0, properties, feature.vertices);
+    const color = getProperty('color', options, randomColor, properties);
```
peppsac, line 132: You're passing a function (`randomColor`) instead of a value. Did you meant `randomColor()`?

zarov: No, because we don't want to execute it if the options already have a value for it. Tiny optimization ;)

peppsac: Oh right, I missed the `if (typeof defaultValue === 'function') {` above. All good then :)

peppsac (approved):

PR #807, opened 2018-07-13
refactor(core): rework texture update process

peppsac:
The main goal of this PR is to ease the usage of the texture update features of
itowns with objects that aren't TileMesh/LayeredMaterial.

One example of such situation is: being able to colorise a pointcloud using
a WMS server.

First the color and texture update mechanism have been splitted, since they don't
share that much code (yet).

Second, part of the update code has been moved back to the Provider and it helps
clarify the roles. Basically, the update loop has now no knowledge at all of
the internal logic of the behavior.

On the other hand the provider now directly implement the various update strategies,
since they're the only one capable of applying them to their protocol.

So the update process is now this;
   1. update code asks the provider if it wants to download new textures
   2. the provider returns the extents/url of the said textures
   3. the update code forward these commands to the scheduler
   4. when the command resolve, the update code process the result
(until now, the responsability of step 1/2 was shared between update/providers)

This also opens the way for a nice pattern: if at step 2 the provider returns the
full url the scheduler can directly check if the cache already has this resource
available and skip the whole queue/priority mechanism and improve the latency.

(but this'll wait for the next PR...)


peppsac:
Pushed another commit that completely remove `LayeredMaterialConstants` - the stuff exposed in this file was an implementation details of `LayeredMaterials`.

gchoqueux (changes requested):
the color and elevation processes should not be different because they are both rasterized and tiled data. The direction to take is to factorize the two processes into one process.

peppsac:
> the color and elevation processes should not be different because they are both rasterized and tiled data. The direction to take is to factorize the two processes into one process.

Currently they are splitted, even if both processes are in the same file.
This PR is a step toward unifying them by improving the whole texture update process.

We can't unite them right now, because as you can see, they still differ too much, but I agree that in the long term the split should be removed.

gchoqueux:
> Currently they are splitted, even if both processes are in the same file.

I prefer factorize the code duplicate rather than separate into two files

peppsac:
> I prefer factorize the code duplicate rather than separate into two files

I moved back the 2 process code in the same file.

I disagree that this is a good move so I left this in a separate commit, waiting for other reviewers opinion.

zarov:
I started a code review, but I'd like to start with two points here:

- for the sake of reviewing, you should keep the changes of the new `TextureProcessing` inside `LayeredMaterialNodeProcessing`, even if none of us likes this file. It is hard to read the diff between the two files. If we come to an agreement on this PR, it would be easy to then move each refactored part to their specific file in a future PR. But even, there is no need to do a temporary separation if we are going to go to a single `TextureProcessing` file.
- `canTextureBeImproved` starting to be more present is really nice, nobody can say the opposite, but I feel that this is the wrong way to take. Here we have moving to a more separate way of handling tiled layer, while we should aim for a more unified way.

peppsac:
> for the sake of reviewing, you should keep the changes of the new TextureProcessing inside LayeredMaterialNodeProcessing, even if none of us likes this file. It is hard to read the diff between the two files.

Done. The last commit can easily be removed at the end of the review (since the goal of this PR is to have a texture update process independant of LayeredMaterial, I'd prefer to change the name here as well)

> canTextureBeImproved starting to be more present is really nice, nobody can say the opposite, but I feel that this is the wrong way to take. Here we have moving to a more separate way of handling tiled layer, while we should aim for a more unified way.

My point here is : each protocol have its own specifics details, and to be able to exploit them we need to have a separate implementation.
This doesn't prevent code sharing (via Helper or shared methods) but I think that it's more efficient than trying to fit all protocols in the same mold.

If you try to use WMS for a pointcloud for instance, you don't have a "tiled" geometry, so if your protocol implementation depends on this concept (and currently it does) you're stuck.

gchoqueux:
> If you try to use WMS for a pointcloud for instance, you don't have a "tiled" geometry, so if your protocol implementation depends on this concept (and currently it does) you're stuck.

There's a design error in itowns. Don't request protocols with a tile, but with extents see #798.
The PR's direction is to remove the duplicate code and unnecessary specifics of the protocols. With this PR you can fetch source with any process. So adding specific code in the providers doesn't seem the right direction at the moment.

On the other hand, texture update process must actually be factored and no longer be dependent on tile.

peppsac:
> Don't request protocols with a tile, but with extents see #798.

Sure. What about download strategies then? Currently they depend on levels,so how are they going to work?

Or what about: https://github.com/iTowns/itowns/pull/798/files#diff-ee787fdb656a91b748143cc97d5e6bc1R82 : what does this `zoom`  parameter means for a pointcloud?

Your PR is nice and all, but it's quite orthogonal with this PR, which tries to improve/clarify the role of Provider (or Source if you prefer) and the processing code. 

> So adding specific code in the providers doesn't seem the right direction at the moment.

Well, I disagree. Unifying / abstracting everything is not a goal in itself - the majority of the technical debt in iTowns comes from this philosophy (remember: 2 years ago, itowns2 was an over-engineered sig 3d engine, that was only capable to display a textured globe).

And last but not least: this PR doesn't add that much code to the Provider. It mostly move some duplicate work (canTextureBeImproved from Provider + select a new extent from Strategies) to a single place (Provider) where we have more info to select the correct extent.

gchoqueux:
> Sure. What about download strategies then? Currently they depend on levels,so how are they going to work?

yes, it is poorly designed. It should be more generic, not be related to zoom, but to resolution and SSE

> what does this zoom parameter means for a pointcloud?

same thing, for all raster data, the zoom must replace the projected resolution. I tested on some part
it will be added to this PR

> Well, I disagree. Unifying / abstracting everything is not a goal in itself - the majority of the technical debt in iTowns comes from this philosophy (remember: 2 years ago, itowns2 was an over-engineered sig 3d engine, that was only capable to display a textured globe).

I can't understand the technical contribution of this argument

the unification and deletion of unnecessary specific codes PR #798, allowed these rapid evolutions like WFS to color layer, or geojson to extrudedMesh, wms to elevation layer... there is a big profit with this abstraction

By example, I can see the interest relatively similar code  `canTextureBeImproved`  in WMTSProvider.js and TMSProviders.js

peppsac:
> I can't understand the technical contribution of this argument

That's pretty easy to understand. You're putting abstraction / genericity above as priority goals in your designs. And I was pointing out that this resulted in massive technical debt and a framework with very few features.

 > allowed these rapid evolutions like WFS to color layer, or geojson to extrudedMesh, wms to elevation layer

All of these are already possible in itowns.

Anyway I give up here. I'll close this PR and maybe submit layer.

PR #810, opened 2018-07-16
feat(core): compute camera near/far automatically

peppsac:
Adds a new feature to compute automatically min/max distance of displayed elements.
The near/far values are then deduced.

The advantage of this solution is we can disable the logarithmicDepthBuffer by default which improve itowns compatibility with lower end / mobile devices.

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -21,9 +21,11 @@
             // iTowns namespace defined here
             var viewerDiv = document.getElementById('viewerDiv');
 
-            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe, { noControls: true });
             view.camera.camera3D.near = 5;
             setupLoadingScreen(viewerDiv, view);
+
+            var controls = new itowns.FirstPersonControls(view, { focusOnClick: true });
 
             var menuGlobe = new GuiTools('menuDiv', view, 300);
 
@@ -32,6 +34,7 @@
             // function use :
             // For preupdate Layer geomtry :
             var preUpdateGeo = itowns.pre3dTilesUpdate;
+            var pickingEnabled = true;
 
             // Create a new Layer 3d-tiles For DiscreteLOD
             // -------------------------------------------
@@ -78,8 +81,12 @@
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerDiscreteLOD, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerRequestVolume, d);
             d.zoom = function() {
-                view.camera.camera3D.position.set(1215013.9, -4736315.5, 4081597.5);
-                view.camera.camera3D.quaternion.set(0.9108514448729665, 0.13456816437801225, 0.1107206134840362, 0.3741416847378546);
+                pickingEnabled = false;
+                const pos = new itowns.Coordinates('EPSG:4978', 1215020.1908037297, -4736310.743783287, 4081602.6452904404);
+                pos.xyz(view.camera.camera3D.position);
+                view.camera.camera3D.up.copy(pos.geodesicNormal);
+                view.camera.camera3D.lookAt(new itowns.THREE.Vector3(1215012, -4736310, 4081602));
+                controls.reset();
                 view.notifyChange(view.camera.camera3D);
             }
             menuGlobe.gui.add(d, 'zoom').name('Go to point cloud');
@@ -96,6 +103,9 @@
                 }
             }
             function picking(event) {
+                if (!pickingEnabled) {
+                    return;
+                }
                 var htmlInfo = document.getElementById('info');
                 htmlInfo.innerHTML = ' ';
 
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -21,11 +21,9 @@
             // iTowns namespace defined here
             var viewerDiv = document.getElementById('viewerDiv');
 
-            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe, { noControls: true });
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
             view.camera.camera3D.near = 5;
             setupLoadingScreen(viewerDiv, view);
-
-            var controls = new itowns.FirstPersonControls(view, { focusOnClick: true });
 
             var menuGlobe = new GuiTools('menuDiv', view, 300);
 
@@ -34,7 +32,6 @@
             // function use :
             // For preupdate Layer geomtry :
             var preUpdateGeo = itowns.pre3dTilesUpdate;
-            var pickingEnabled = true;
 
             // Create a new Layer 3d-tiles For DiscreteLOD
             // -------------------------------------------
@@ -81,12 +78,8 @@
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerDiscreteLOD, d);
             debug.create3dTilesDebugUI(menuGlobe.gui, view, $3dTilesLayerRequestVolume, d);
             d.zoom = function() {
-                pickingEnabled = false;
-                const pos = new itowns.Coordinates('EPSG:4978', 1215020.1908037297, -4736310.743783287, 4081602.6452904404);
-                pos.xyz(view.camera.camera3D.position);
-                view.camera.camera3D.up.copy(pos.geodesicNormal);
+                view.camera.camera3D.position.set(1215020.1908037297, -4736310.743783287, 4081602.6452904404);
                 view.camera.camera3D.lookAt(new itowns.THREE.Vector3(1215012, -4736310, 4081602));
-                controls.reset();
                 view.notifyChange(view.camera.camera3D);
             }
             menuGlobe.gui.add(d, 'zoom').name('Go to point cloud');
@@ -103,9 +96,6 @@
                 }
             }
             function picking(event) {
-                if (!pickingEnabled) {
-                    return;
-                }
                 var htmlInfo = document.getElementById('info');
                 htmlInfo.innerHTML = ' ';
 
--- a/src/Core/MainLoop.js
+++ b/src/Core/MainLoop.js
@@ -121,7 +121,18 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
         engine: this.gfxEngine,
         scheduler: this.scheduler,
         view,
+        distance: {
+            min: Infinity,
+            max: 0,
+        },
+        fastUpdateHint: undefined,
     };
+
+    const previousNear = view.camera.camera3D.near;
+    const previousFar = view.camera.camera3D.far;
+    view.camera.camera3D.near = 0.1;
+    view.camera.camera3D.far = 2000000000;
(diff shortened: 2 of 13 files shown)


examples/3dtiles.html
```diff
@@ -21,17 +21,20 @@
             // iTowns namespace defined here
             var viewerDiv = document.getElementById('viewerDiv');
 
-            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe, { noControls: true });
             view.camera.camera3D.near = 5;
             setupLoadingScreen(viewerDiv, view);
 
+            var controls = new itowns.FirstPersonControls(view, { focusOnClick: true });
```
gchoqueux: why you change controls?

src/Core/MainLoop.js
```diff
@@ -121,8 +121,22 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
         engine: this.gfxEngine,
         scheduler: this.scheduler,
         view,
+        distance: {
```
gchoqueux: you could comment on these new parameters, `distance` and `fastUpdateHint`

src/Core/MainLoop.js
```diff
@@ -121,8 +121,22 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
...
+            min: Infinity,
+            max: 0,
+        },
+        fastUpdateHint: undefined,
     };
 
+    const previousNear = view.camera.camera3D.near;
+    const previousFar = view.camera.camera3D.far;
+    view.camera.camera3D.near = 0.1;
```
gchoqueux: the near and the far are reset here for processing requiring the camera?

and they are restored at the end of the process?

src/Process/3dTilesProcessing.js
```diff
@@ -257,24 +257,36 @@ export function pre3dTilesUpdate(context, layer) {
...
+export function computeNodeSSE(context, camera, node) {
     node.distance = 0;
     if (node.boundingVolume.region) {
         boundingVolumeBox.copy(node.boundingVolume.region.box3D);
         boundingVolumeBox.applyMatrix4(node.boundingVolume.region.matrixWorld);
         node.distance = boundingVolumeBox.distanceToPoint(camera.camera3D.position);
+        // update distance min/max
+        const s = node.boundingVolume.region.box3D.getSize();
+        const maxComponent = Math.max(s.x, Math.max(s.y, s.z));
+        context.distance.min = Math.min(context.distance.min, node.distance - maxComponent);
```
gchoqueux: the calculations of min and max can be done, only once, after the calculation of the `boudingVolume`

The name of the function must also take into account this new calculation

src/Process/PointCloudProcessing.js
```diff
@@ -86,6 +86,15 @@ function markForDeletion(elt) {
     }
 }
 
+function updateMinMaxDistance(context, bbox) {
```
gchoqueux, line 89: this isn't a calculation equivalent to those performed in 3Dtiles?

src/Core/MainLoop.js
```diff
@@ -153,6 +168,16 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
             view.execFrameRequesters(MAIN_LOOP_EVENTS.AFTER_LAYER_UPDATE, dt, this._updateLoopRestarted, geometryLayer);
         }
     }
+
+    if (context.distance.min != Infinity) {
```
gchoqueux, line 189: could be a camera function,` camera.adjustNearFar(min, max)`

src/Process/3dTilesProcessing.js
```diff
@@ -257,24 +257,36 @@ export function pre3dTilesUpdate(context, layer) {
 
 const boundingVolumeBox = new THREE.Box3();
 const boundingVolumeSphere = new THREE.Sphere();
-export function computeNodeSSE(camera, node) {
+export function computeNodeSSE(context, camera, node) {
```
gchoqueux: context used only for distance, so the params could be with the `distance`.

src/Core/MainLoop.js
```diff
@@ -153,6 +168,16 @@ MainLoop.prototype._update = function _update(view, updateSources, dt) {
...
+        // Multiply min/max by (0.9, 1.1) to be on the safe side
+        view.camera.camera3D.near = Math.max(0.1, 0.9 * context.distance.min);
```
gchoqueux, line 191: As the `min` distance is calculated relative to the camera's position, close objects may be truncated by the near.
This min distance must be projected on the z-axis of the camera

src/Process/PointCloudProcessing.js
```diff
@@ -137,87 +146,95 @@ export default {
...
-                        }
-                        elt.obj.boxHelper.visible = true;
-                        elt.obj.boxHelper.material.color.r = 1 - elt.sse;
-                        elt.obj.boxHelper.material.color.g = elt.sse;
+            if (!elt.visible) {
+                markForDeletion(elt);
+                return;
+            }
+
+            const distance = updateMinMaxDistance(context, bbox);
```
gchoqueux, line 175: I would do a function that calculates the distance, and then with the distance, I'll refresh the context

gchoqueux (changes requested):
by aggregating the displayable bboxes, to obtain the bbox of the scene, (with box3.union), the calculation of the distance min and max would be made only once

src/Core/MainLoop.js
peppsac: Yes.
Because the update functions may use the view frustum to determine visibility.
The near/far affects the result, so we reset them to ~0 and ~Infinity before the processing.
After the visibility process is over, we can define them to the correct values (correct = the minimum z-range that contains all visible objects).

src/Process/3dTilesProcessing.js
peppsac: > the calculations of min and max can be done, only once, after the calculation of the boudingVolume

I guess you meant "after the calculation of `boundingBoxVolume` for the boundingVolume.region and boundingVolume.box cases" ?

src/Process/PointCloudProcessing.js
peppsac: Yep

src/Core/MainLoop.js
peppsac: Yeah, but in this case it would be accessible to anyone. Since it's an internal itowns thing, I prefer to keep this private.

src/Process/3dTilesProcessing.js
peppsac: Yeah.. or I could remove camera (since it's part of context as well).

src/Core/MainLoop.js
peppsac: I don't understand, can you rephrase?

src/Process/PointCloudProcessing.js
peppsac: Err.. I didn't get this comment either. Can you reformulate?

src/Core/MainLoop.js
gchoqueux: it's possible to move it in Camera and it's not accessible

src/Core/MainLoop.js
gchoqueux: ok, comment on your approach

src/Core/MainLoop.js
gchoqueux: `view.camera.camera3D.near ` it 's the distance between the near plane and the camera's position, on camera's  z Axis.

`context.distance.min` is the distance between camera's position and bbox. This distance should be reprojected on camera's  z Axis.

src/Process/3dTilesProcessing.js
gchoqueux: it's preferable to pass only the parameters used in the function

src/Process/3dTilesProcessing.js
```diff
@@ -257,24 +257,36 @@ export function pre3dTilesUpdate(context, layer) {
...
     } else if (node.boundingVolume.box) {
         // boundingVolume.box is affected by matrixWorld
         boundingVolumeBox.copy(node.boundingVolume.box);
         boundingVolumeBox.applyMatrix4(node.matrixWorld);
         node.distance = boundingVolumeBox.distanceToPoint(camera.camera3D.position);
+        // update distance min/max
+        const s = node.boundingVolume.box.getSize();
+        const maxComponent = Math.max(s.x, Math.max(s.y, s.z));
+        context.distance.min = Math.min(context.distance.min, node.distance);
+        context.distance.max = Math.max(context.distance.max, node.distance + maxComponent);
```
gchoqueux: `maxComponent` is the diagonal

src/Process/3dTilesProcessing.js
gchoqueux: in line 293

src/Process/PointCloudProcessing.js
gchoqueux: it should be factorized then or use the [aggregation bboxs](https://github.com/iTowns/itowns/pull/810#pullrequestreview-139481142) to make the generic method

examples/3dtiles.html
peppsac: I'll restore `GlobeControls`.

src/Process/PointCloudProcessing.js
gchoqueux: ~this is important for now I will come back to it later~

src/Core/MainLoop.js
peppsac: Done

src/Process/3dTilesProcessing.js
peppsac: Nope, because in the third case ther's a bounding sphere, not box.
I've simplified this part anyway.

src/Process/3dTilesProcessing.js
peppsac: Makes sense, I'll change this.

src/Process/PointCloudProcessing.js
peppsac: Done.

src/Process/PointCloudProcessing.js
peppsac: > this is important for now I will come back to it later

:confused:

peppsac:
@gchoqueux: I pushed an update that should fix all your comments (except [the mysterious one](https://github.com/iTowns/itowns/pull/810#discussion_r204467818))

src/Process/PointCloudProcessing.js
gchoqueux: this is not important for now I will come back to it later

src/Core/MainLoop.js
peppsac: Then give me an example please :)

src/Core/MainLoop.js
peppsac: `min` is the minimum distance between the camera's center and any (visible) objects in the scene.
So I don't see why / what should be reprojected on camera's z axis.

Can you make a drawing of a case where using the `min` distance as the `near` is incorrect?

nosy-b:
@gchoqueux What do you think now?

gchoqueux:
Bouding volumes should be aggregated to obtain the scene's bbox (only with the displayed objects). And at the end, calculate the distance from the camera to this box

PR #811, opened 2018-07-17
fix(3dtiles): simplify 3dtiles layer creation

peppsac:
Add default parameters for the main layer properties to avoid forcing the user to define them each time.

(commit extracted from PR #802)

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -314,7 +314,7 @@ function markForDeletion(layer, elt) {
     }
 }
 
-export function process3dTilesNode(cullingTest, subdivisionTest) {
+export function process3dTilesNode(cullingTest = $3dTilesCulling, subdivisionTest = $3dTilesSubdivisionControl) {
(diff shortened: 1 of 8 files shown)


src/Process/3dTilesProcessing.js
```diff
@@ -314,7 +314,7 @@ function markForDeletion(layer, elt) {
     }
 }
 
-export function process3dTilesNode(cullingTest, subdivisionTest) {
+export function process3dTilesNode(cullingTest = $3dTilesCulling, subdivisionTest = $3dTilesSubdivisionControl) {
```
zarov, line 317: :+1:

zarov (approved):
LGTM, thanks for this, more things should be moved and regrouped like this.

FYI I have a soon-to-come PR with updated Layer, similar to #794 (that I will certainly close), see https://github.com/zarov/itowns/commit/3af73b0ed9f39bceaac2625e417a7625901e349c

peppsac:
Thanks!

PR #812, opened 2018-07-17
fix(core): compute globe fog distance after camera update

peppsac:
The fog distance is used during the update process of each tile of the
globe so the computation must happen between the camera update and the
tile update.

This commit moves this update to MAIN_LOOP_EVENTS.AFTER_CAMERA_UPDATE.


zarov (commented):
You can remove the `TODO` note below your modification.

peppsac:
> You can remove the TODO note below your modification.

Done

zarov (approved):

PR #814, opened 2018-07-23
fix(packaging): fix package_lock.json

gliegard:
Fix package-lock.json

Prior to this change, since commit 6411908e, package_lock was generated with npm 5.6
Now, package_lock is generated with npm 6.1.0


zarov:
Thanks for fixing this, what about specifying something in the `package.json` too, in order to be consistent all-together in the future ? See https://docs.npmjs.com/files/package.json#engines

peppsac:
Thanks for the fix.

Just one small nit: can you replace `packaging` with `chore` in the commit title please?

gliegard:
@peppsac  ok, done
@zarov If I do that, It won't work any more on npm 5.6, which is tested in Travis... Maybe I'll propose that on another PR : Force every body to use npm 6 to avoid wasting time......

zarov (approved):
LGTM thanks

PR #815, opened 2018-07-23
refactor: move Layer folder, extract geometric layers from views

zarov:
This is a PR similar to #794 but more advanced. It is still missing documentation and test at the moment, but we can discuss the main goal of the PR while I add it.

## Description

Several things have been done here:
- move the content of `src/Core/Layer` to `src/Layer`
- create several class for layers in this folder: `GeometryLayer`, `TiledGeometryLayer`,  `ColorLayer` and `ElevationLayer`
- from each type of view in `src/Core/Prefab`, remove the `createXXXLayer` in profit of a `XXXLayer` file

## Motivation and Context

I feel that layers are one of the main aspect in itowns, and refactoring them like this can help us see clearer through it and help refactoring things around them (like #798 and #807).


peppsac:
If we move away from layers being data-container + attached (external) functions to layers being classes, then we should also reconsider the existence of the process code.

For instance, all the code from `GlobeTileProcessing` is now only meant to be used from `GlobeLayer` so IMHO, it would make more sense to remove `GlobeTileProcessing` and move all of its code in `GlobeLayer`.

Other questions:
- why `GlobeLayer` exists but not `$3dtilesLayer` or `PointCloudLayer`?
- why not removing completely the `view.addLayer({ type: 'color' })` API, and only allow the  `view.addLayer(new ColorLayer(...))` form?

zarov:
> For instance, all the code from GlobeTileProcessing is now only meant to be used from GlobeLayer so IMHO, it would make more sense to remove GlobeTileProcessing and move all of its code in GlobeLayer.

Completely agree, all layers should have their own `update` method. I didn't do it here to not clutter the PR.

> why GlobeLayer exists but not $3dtilesLayer or PointCloudLayer?

It could, and it should, but those layers are for now more present in examples. But we could take the opportunity to create them here.

> why not removing completely the view.addLayer({ type: 'color' }) API, and only allow the view.addLayer(new ColorLayer(...)) form?

I don't have an opinion here, I'm okay to leave it or remove it. It could be at least deprecated in this PR.

peppsac:
I don't have any opinion on the proposed changes, so I'll keep asking questions :)

Is there a point to keep `PlanarView`, `GlobeView`, `PanoramaView`? Can't you move all the view type specific code to each Layer class?

> It could, and it should, but those layers are for now more present in examples. But we could take the opportunity to create them here.

I'm not sure that theses layers would make sense actually.
Can you show the code to display a 3dtiles point cloud using itowns using both this PR and a `source` concept (the one from #798 or similar)? what about a pointcloud from PotreeConverter?

zarov:
> Is there a point to keep PlanarView, GlobeView, PanoramaView? Can't you move all the view type specific code to each Layer class?

There is some code that we could move from views to layers, but I don't really see the point of dropping views. To me, views are the basis of any project using itowns. Or do you meant we could only keep `View` and only use that ?

> I'm not sure that theses layers would make sense actually.

OK, so maybe a 3dtiles layer would be added as a GeometryLayer/TiledGeometryLayer, and then the provider(/source ?) would change the specific update stuff ?

src/Core/Layer/Layer.js
```diff
@@ -1,199 +0,0 @@
-import { EventDispatcher } from 'three';
```
gchoqueux, line 1: move Layer.js in other PR to see history change

src/Core/Prefab/Globe/GlobeLayer.js
```diff
@@ -0,0 +1,61 @@
...
+                    config.maxDeltaElevationLevel || 4)(context, layer, node);
+            }
+            return false;
+        }
+
+        this.update = processTiledGeometryNode(globeCulling(2), subdivision);
+        this.builder = new BuilderEllipsoidTile();
+    }
+
+    pickObjectsAt(view, mouse, radius = 5) {
```
gchoqueux: why he doesn't inherit from geometryLayer or TiledGeometryLayer.js?

src/Core/Prefab/Globe/GlobeLayer.js
```diff
@@ -0,0 +1,61 @@
...
+        return Picking.pickTilesAt(view, mouse, radius, this);
+    }
+
+    preUpdate(context, changeSources) {
+        if (__DEBUG__) {
+            this._latestUpdateStartingLevel = 0;
+        }
+
+        preGlobeUpdate(context, this);
```
gchoqueux: Why `globeLayer.preUpdate` should be identical to `planarLayer.preUpdate` or `panoramaLayer.preUpdate`, exepted for `preGlobeUpdate`, `prePlanarUpdate`

src/Layer/Layer.js
```diff
@@ -0,0 +1,167 @@
...
+     * @constructor
+     * @protected
+     * @param {Object} config
+     */
+    constructor(config) {
+        super();
+
+        Object.assign(this, config);
+
+        Object.defineProperty(this, 'id', {
```
gchoqueux, line 56: we could you use class's getter and setter?

src/Layer/Layer.js
```diff
@@ -0,0 +1,167 @@
...
+
+export default Layer;
+
+export const ImageryLayers = {
+    // move this to new index
+    // After the modification :
+    //      * the minimum sequence will always be 0
+    //      * the maximum sequence will always be layers.lenght - 1
+    // the ordering of all layers (Except that specified) doesn't change
+    moveLayerToIndex: function moveLayerToIndex(layer, newIndex, imageryLayers) {
```
gchoqueux, line 150: maybe this should be in `geometryLayer` because the color and elevation layers are only attached to geometric layers : `geometryLayer.moveColorLayerToIndex(layer, index)`?

gchoqueux (commented):
the specific globe's part in `GlobeTileProcessing` is the horizonCulling, this part could be moved in `globeLayer`. We must strive to unify the other parts of the GlobeTileProcessing (if possible and justified)

> Do you meant we could only keep `View` and only use that ?

Yes, it would be necessary to have only `View`, the different things are the `controls / Camera` and the `fog / atmosphere` (atmosphere should be a layer)

src/Core/Layer/Layer.js
zarov: Yep !

src/Core/Prefab/Globe/GlobeLayer.js
zarov: Because of the default `radius` value which isn't the same, and I don't know a way of specifying this differently.

src/Layer/Layer.js
zarov: Only using a `getter` would do the same job indeed, but does not emit an error: what is your take here ?

src/Layer/Layer.js
zarov: This kind of makes sense, but this method is used in `ColorLayersOrdering` too; how do you see the integration with this one, very specific to color layers ?

src/Core/Prefab/Globe/GlobeLayer.js
zarov: Done

zarov:
PR updated, I moved the `pickObjectsAt` and added documentation. 

Note that it is getting quite unbrowseable in the jsdoc page, so it may be interesting to regroup classes by theme (here we could have `Layers` for example) and only display their name and not the method names. But this is the matter of another PR.

src/Core/Layer/Layer.js
```diff
@@ -1,158 +1,151 @@
...
+        Object.defineProperty(this, 'type', {
+            value: type,
+            writable: false,
+        });
+
+        // Default properties
+        this.options = config.options || {};
+
+        // Does this layer is attached to another one
+        this.isAttached = false;
```
gchoqueux: it's necessary, because if `(this.attachedLayers.length === 0)` so `this.isAttached === false`

src/Core/Layer/Layer.js
```diff
@@ -1,158 +1,151 @@
...
+
+        if (!this.updateStrategy) {
```
gchoqueux: **Note**
_It's necessary to think where to place the strategy of loading.
The layer is a good place because it must independent of the source._

src/Core/Layer/Layer.js
```diff
@@ -1,158 +1,151 @@
...
+            writable: false,
+        });
+
+        Object.defineProperty(this, 'type', {
+            value: type,
+            writable: false,
+        });
+
+        // Default properties
+        this.options = config.options || {};
```
gchoqueux: It would be necessary, to write `_` before private property name.
Could you tell me the properties that make sense to be accessible? 

by example` this.isAttached` is not supposed to be modifiable by the user

src/Core/Prefab/Globe/GlobeLayer.js
```diff
@@ -0,0 +1,72 @@
...
+        // Configure tiles
+        this.schemeTile = globeSchemeTileWMTS(globeSchemeTile1);
+        this.extent = this.schemeTile[0].clone();
+        for (let i = 1; i < this.schemeTile.length; i++) {
+            this.extent.union(this.schemeTile[i]);
+        }
+
+        function subdivision(context, layer, node) {
+            if (TiledGeometryLayer.hasEnoughTexturesToSubdivide(context, node)) {
+                return globeSubdivisionControl(2,
```
gchoqueux: * Could you standardize the signature of  `globeSubdivisionControl`, `panoramaSubdivisionControl` and `planarSubdivisionControl`, because these functions are pretty much the same

* Could you move `subdivision` and `this.update` declaration to `TiledGeometryLayer`

src/Core/Prefab/Panorama/PanoramaLayer.js
```diff
@@ -0,0 +1,117 @@
...
+            if (TiledGeometryLayer.hasEnoughTexturesToSubdivide(context, node)) {
+                return panoramaSubdivisionControl(
+                    config.maxSubdivisionLevel || 10, new THREE.Vector2(512, 256))(context, layer, node);
+            }
+            return false;
+        }
+
+        this.update = processTiledGeometryNode(panoramaCulling, subdivision);
+        this.builder = new PanoramaTileBuilder(type, config.ratio || 1);
+        this.segments = 8;
```
gchoqueux: move to `this.options`

src/Core/View.js
```diff
@@ -76,7 +79,7 @@ function View(crs, viewerDiv, options = {}) {
     window.addEventListener('resize', () => {
         // If the user gave us a container (<div>) then itowns' size is
         // the container's size. Otherwise we use window' size.
-        const newSize = new Vector2(viewerDiv.clientWidth, viewerDiv.clientHeight);
+        const newSize = new THREE.Vector2(viewerDiv.clientWidth, viewerDiv.clientHeight);
```
gchoqueux, line 82: you don't like call without namespace?

src/Core/View.js
```diff
@@ -175,47 +169,11 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
...
-        defineLayerProperty(layer, 'visible', true);
-        defineLayerProperty(layer, 'opacity', 1.0);
-        defineLayerProperty(layer, 'sequence', 0);
-    } else if (layer.type == 'elevation') {
-        defineLayerProperty(layer, 'frozen', false);
-    } else if (layer.type == 'geometry' || layer.type == 'debug') {
-        defineLayerProperty(layer, 'visible', true, () => _syncGeometryLayerVisibility(layer, view));
-        defineLayerProperty(layer, 'frozen', false);
+    if (layer.type == 'geometry' || layer.type == 'debug') {
+        layer.defineLayerProperty('visible', true, () => _syncGeometryLayerVisibility(layer, view));
```
gchoqueux, line 173: It's a shame that (layer.defineLayerProperty) this remains
It should use the notifyChange mechanism

src/Layer/GeometryLayer.js
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @param {View} view - The view instance.
+     * @param {Object} coordinates - The coordinates to pick in the view. It
+     * should have at least <code>x</code> and <code>y</code> properties.
+     * @param {number} radius - Radius of the picking circle.
+     *
+     * @return {Array} An array containing all targets picked under the
+     * specified coordinates.
+     */
+    pickObjectsAt(view, coordinates, radius) {
```
gchoqueux: why radius parameter don't use default value?
`pickObjectsAt(view, coordinates, radius = this.config.defaultRadius)`

src/Layer/TiledGeometryLayer.js
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @param {View} view - The view instance.
+     * @param {Object} coordinates - The coordinates to pick in the view. It
+     * should have at least <code>x</code> and <code>y</code> properties.
+     * @param {number} radius - Radius of the picking circle.
+     *
+     * @return {Array} An array containing all targets picked under the
+     * specified coordinates.
+     */
+    pickObjectsAt(view, coordinates, radius) {
```
gchoqueux: you can declare in `class geometryLayer`
```js
 pickObjectsAt(view, coordinates, radius = this.defaultPickingRadius) {
        return Picking.pickTilesAt(view, coordinates, radius, this);
 }
```
and remove `pickObjectsAt` in `TiledGeometryLayer`

src/Process/GlobeTileProcessing.js
```diff
@@ -42,15 +42,6 @@ export function preGlobeUpdate(context, layer) {
 
     // pre-sse
     context.camera.preSSE = _preSSE(context.view);
-
```
gchoqueux, line 45: what is this deleted or moved part for?

src/Process/PlanarTileProcessing.js
```diff
@@ -24,17 +24,6 @@ function _isTileBigOnScreen(camera, node) {
     return (dim.x >= 256 && dim.y >= 256);
 }
 
-export function prePlanarUpdate(context, layer) {
```
gchoqueux, line 27: same part that [here](https://github.com/iTowns/itowns/pull/815/files#diff-989ef5eaaef422d31895b5f0e7eb2c7aL46)
There's the same `preUpdate` in `globePreUpdate` ?
It's used for elevation with color data , can use color data for elevation in the `globeLayer`?

gchoqueux (changes requested):

src/Core/Layer/Layer.js
zarov: So this line comes from `View`, so I kinda let it as it is. For private property name, I'm okay to write it with `_`, but I don't know a way to keep a property modifiable by us but not by the user.

src/Core/Prefab/Globe/GlobeLayer.js
zarov: - pretty much yes, but the panorama one is kinda annoying... I'll see what I can do here
- I don't agree with this, I'd prefer to transfer the called methods (from `GlobeTiledProcessing` for example) in layers in the PR following this one.

src/Core/View.js
zarov: No, I feel that `THREE` parts should be visible from the itowns code. But I can revert back to what it was, no problem.

src/Core/View.js
zarov: How could I use it ?

src/Layer/GeometryLayer.js
zarov: Thanks, didn't think of it this way !

src/Layer/TiledGeometryLayer.js
zarov: But `TiledGeometryLayer` uses `pickTilesAt` and `GeometryLayer` uses `pickObjectsAt`.

src/Process/PlanarTileProcessing.js
zarov: It is the same indeed, so I moved it to TiledGeometryLayer.

src/Core/Prefab/Panorama/PanoramaLayer.js
zarov: Done

gchoqueux (approved):

gchoqueux:
good job

PR #816, opened 2018-07-23
feat(chore): verify that we are using npm 6

gliegard:
Prior to this change, doing 'npm install' with npm version 5 generate a packagoe_lock.json in the old format.
The problem is that we waste time managing package_lock.json in different formats.

A solution is that we all use the last version of npm
This change, verify npm version during 'nmp install', log an error if npm is not in the version 6 or older.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -29,3 +29,5 @@ iTowns uses NPM as the build tool. If you haven't already, install Node.js: http
 ## Contribute back
 
 See [the contributor's guide](CONTRIBUTING.md)
+
+Note : You should not commit changes to `package-lock.json` if you're not using NPM >= 6 (particularly if you didn't make any change to the `package.json` either).
--- a/package-lock.json
+++ b/package-lock.json
@@ -1727,6 +1727,79 @@
         "color-name": "^1.0.0"
       }
     },
+    "check-node-version": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/check-node-version/-/check-node-version-3.2.0.tgz",
+      "integrity": "sha512-mJu4dADRf+NUeOyGgFTXaLtjyyffD3Eej2RA9IEk1CdHmoVurErLD++e/Ps6uKfsB273ky+0Z9NlOiuplxuNdw==",
+      "dev": true,
+      "requires": {
+        "chalk": "^2.3.0",
+        "map-values": "^1.0.1",
+        "minimist": "^1.2.0",
+        "object-filter": "^1.0.2",
+        "object.assign": "^4.0.4",
+        "run-parallel": "^1.1.4",
+        "semver": "^5.0.3"
+      },
+      "dependencies": {
+        "ansi-styles": {
+          "version": "3.2.1",
+          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
+          "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
+          "dev": true,
+          "requires": {
+            "color-convert": "^1.9.0"
+          }
+        },
+        "chalk": {
+          "version": "2.4.1",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.1.tgz",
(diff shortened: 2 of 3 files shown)


tbroyer (dismissed):
If I'm not mistaken, this would also force anyone **using** iTowns (as published on NPM) to 1) use NPM 6, and 2) run that `check-node-version` check (which, btw, won't work, as `check-node-version` is declared as a `devDependency`).

`"engines":{"npm": …}` is bad, and `install` is the wrong script to put that check in; as we only want to target iTowns contributors, not "consumers".

gliegard:
You can try it, it works : when you do 'npm install' , it first download the dependencies and THEN it run automatically 'npm run install' that checks the npm version.

The documentation https://docs.npmjs.com/files/package.json#engines
_Unless the user has set the engine-strict config flag, this field is advisory only and will only produce warnings when your package is installed as a dependency._

So It will only produces warning for npm 5 users.

tbroyer:
> You can try it, it works : when you do 'npm install' , it first download the dependencies and THEN it run automatically 'npm run install' that checks the npm version.

You're running `npm install` inside the iTowns checkout, so yes it'll download the `devDependencies`. Now, try to use iTowns in another project:
```
$ npm install --save git+https://github.com/iTowns/itowns.git#verify_npm_6
npm WARN deprecated togeojson@0.16.0: This module has moved: please install @mapbox/togeojson instead

> itowns@2.3.0 install /tmp/itowns-test/node_modules/itowns
> check-node-version --package || true

sh: check-node-version : commande introuvable
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN itowns@2.3.0 requires a peer of proj4@^2.4.4 but none is installed. You must install peer dependencies yourself.
npm WARN itowns@2.3.0 requires a peer of three@^0.89.0 but none is installed. You must install peer dependencies yourself.
npm WARN itowns@2.3.0 requires a peer of three.meshline@^1.0.3 but none is installed. You must install peer dependencies yourself.
npm WARN itowns-test@1.0.0 No description
npm WARN itowns-test@1.0.0 No repository field.

+ itowns@2.3.0
added 30 packages from 23 contributors and audited 35 packages in 5.896s
found 0 vulnerabilities
```
And of course, same when running `npm install` without argument afterwards, once the `package.json` has been modified by the above command:
```
$ npm install

> itowns@git+https://github.com/iTowns/itowns.git#a83561edb0b5c34303fb1e7e74ec0f1a54db9422 install /tmp/itowns-test/node_modules/itowns
> check-node-version --package || true

sh: check-node-version : commande introuvable
npm WARN itowns-test@1.0.0 No description
npm WARN itowns-test@1.0.0 No repository field.

added 30 packages from 23 contributors and audited 35 packages in 1.169s
found 0 vulnerabilities
```

This is why I said `install` is not the appropriate script, `prepare` would be more appropriate.

> The documentation https://docs.npmjs.com/files/package.json#engines
> _Unless the user has set the engine-strict config flag, this field is advisory only and will only produce warnings when your package is installed as a dependency._
> 
> So It will only produces warning for npm 5 users.

But it will do so not only for people working on iTowns, but also anyone using iTowns in their own apps.

Replace that with a `check-node-version --npm '>= 6'` in a `prepare` script (and no `engine` at all) and I'll be OK.

gliegard:
ok, thanks for the explanation :)
updated

tbroyer (commented):
How about adding something to our `BUILDING.md`? Maybe a note saying that you should not commit changes to `package-lock.json` if you're not using NPM >= 6 (particularly if you didn't make any change to the `package.json` either)

gliegard:
Yes, good idea !

tbroyer (approved):

autra:
To explain a bit more what's going on, I propose:

```
"prepare": "check-node-version --npm '>= 6' || echo \"We detected that you are using npm < 6, please use npm >= 6 if you want to make modification to package-lock.json.\n\""
```

Maybe even a `feel free to ignore the above error if not.` at the end? WDYT?

gliegard:
Yes, good idea.
updated

autra:
> feat(chore)

Nope. it's either feat or chore. Feat means feature and chore means chore. They are mutually exclusive, see #518. So it should be something like (with slightly amended commit message and typo corrected):

> chore(build): verify we are using  npm >=6
>
> Doing 'npm install' with npm version 5 generate a package_lock.json in the old format. So we were wasting time managing package_lock.json in different formats.
> Let's check we all use npm >= 6 when generating it.

Also can you squash? Thanks!

gliegard:
ok

PR #817, opened 2018-07-24
chore(features): remove unused minAltitude property

zarov:
The minAltitude property when creating a Mesh from a Feature was not
assigned and always equal to a NaN, so it has been removed.

Not sure if we could remove it from FeatureProcessing too.

### DIFF
--- a/src/Renderer/ThreeExtended/Feature2Mesh.js
+++ b/src/Renderer/ThreeExtended/Feature2Mesh.js
@@ -258,7 +258,7 @@ function featureToMesh(feature, options) {
         return;
     }
 
-    var mesh;
+    let mesh;
     switch (feature.type) {
         case 'point':
         case 'multipoint': {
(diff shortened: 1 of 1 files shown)


gchoqueux:
Should be fixed, i was removed in pr #691 [here ](https://github.com/iTowns/itowns/pull/691/files#diff-2a78502be7a11a150dcce98113215193L211)

peppsac:
> Should be fixed, i was removed in pr #691 here 

This PR was merged 3 months ago, and no one complained about this `minAltitude` thing being broken.

So instead of assessing "should be fixed" in a rather unhelpful way, you should explain why it should be "fixed" (I assume you meant restored?).

zarov:
> in a rather unhelpful way

Your comment is not helpful either, a simple "could you explain more on the issue ?" would have been enough. But let's not change the subject here.

The problem with the removal of `minAltitude` is kind of hidden. If you load the `globe_wfs_extruded` example, and select a wireframe view before the apparition of the WFS layer, you can see that buildings are not extruding from their real altitude but rather from "far" below the ground. The `minAltitude` parameter was introduced to prevent this problem in the past. So I need to either add it back or find a replacement to it.

peppsac:
> Your comment is not helpful either, a simple "could you explain more on the issue ?" would have been enough.

Noted.

gchoqueux (changes requested):
doesn't remove minAltitude, it is needed to position features. on the other hand, it is necessary to fix the calculation that was deleted

gchoqueux:
fix in #870

PR #818, opened 2018-07-25
fix(core): implement preSSE once and fix the formula

peppsac:
3dTilesProcessing and GlobeTileProcessing use the same wrong SSE formula.

PR #633 proposed to fix this, but wasn't merged.

This commit fixes the computation, and comment it so the details should
be clearer and less dependant on an external reference that might go away.

The fixed bug can be seen easily in the oriented image example: if you
zoom in, the moutains behind the photos loses precision instead of increase it
(ping @gliegard since you noticed this issue while working on this example)

This should be fixed, and I added a unit test for this problem.

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -227,15 +227,6 @@ export function pre3dTilesUpdate(context, layer) {
         return [];
     }
 
-    // pre-sse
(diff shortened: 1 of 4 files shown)


peppsac:
Note: the previous SSE intended to compute a diagonal FOV and I replaced with something simpler (= the original formula using the vertical FOV).
Since this whole SSE thing is already an approximation (because it doesn't consider the orientation of the objects) this simplification doesn't hurt.

gchoqueux:
move in #947

PR #820, opened 2018-07-25
fix(protocols): wrong xbil/DEM texture's min and max

gchoqueux:
Wrong xbil/DEM texture's min and max was computed during fetching.

xbil/DEM texture's min and max was only calculated on an entire texture, during fetching.
it didn't consider the smaller pitchs

### DIFF
--- a/src/Provider/WMTSProvider.js
+++ b/src/Provider/WMTSProvider.js
@@ -5,9 +5,10 @@
  */
 
 import * as THREE from 'three';
-import OGCWebServiceHelper from './OGCWebServiceHelper';
+import OGCWebServiceHelper, { SIZE_TEXTURE_TILE } from './OGCWebServiceHelper';
 import URLBuilder from './URLBuilder';
 import Extent from '../Core/Geographic/Extent';
+import { computeMinMaxElevation } from '../Parser/XbilParser';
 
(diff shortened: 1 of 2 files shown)


src/Provider/WMTSProvider.js
```diff
@@ -79,11 +80,26 @@ function getXbilTexture(tile, layer, targetZoom) {
 
     return OGCWebServiceHelper.getXBilTextureByUrl(urld, layer.networkOptions).then((texture) => {
         texture.coords = coordWMTS;
+        let min;
+        let max;
+
+        // if pitch is full texture, the min and max are already known
+        if (pitch.z == 1.0 && pitch.w == 1.0) {
+            min = !texture.min ? 0 : texture.min;
```
peppsac: Is this test `!texture.min` to handle the:
```js
    if (max === -1000000 || min === 1000000) {
        return { min: null, max: null };
    }
```
case from XbilParser?

If yes, then would it make more sense to adapt XbilParser to return `{ max: 0, min: 0 }` in this case instead of fixing here?

gchoqueux: yes, it's changed

peppsac (approved):
LGTM, thanks.

Issue #797, opened 2018-06-29
3dtiles cache clean memory failed.


Oncledjoe:
The 3dtiles cache is never clean, see picture. This test  textured cache has a size of 264.7 Mo and we can see the memory overhead and freeze my computer. This problem is also true without textured cache but the size is very lower so the problem doesn't occur.  
## Your Environment

* Version used: master
* Browser Name and version: Chrome last version
* Operating System and version (desktop or mobile): desktop ubuntu 16.4
* Link to your project: Nop

## Context
I need to visualize 3dtiles cache. 

## Steps to Reproduce (for bugs)
You need a textured  cache. A simple cache works also but doesn't crash you computer

## Expected Behavior
Clean memory and video memory. 

## Actual Behavior
Overflow memory

## Possible Cause/Fix/Solution
3dTilesProcessing.js the cleanup function : cleanup3dTileset doesn't work at all! Think to clean webgl memory too. 

Looking forward.

Best Regards,
![issue](https://user-images.githubusercontent.com/36001626/42095912-405ae880-7bb4-11e8-8f76-448f78e45647.png)

Oncledjoe

zarov:
The problem may be coming from here https://github.com/iTowns/itowns/blob/master/src/Process/3dTilesProcessing.js#L204

Maybe we could change it to use the recently added `Cache` ?

autra:
AFAIK this is orthogonal to a cache system: this just cleans tiles not used any more, but don't do it every frame to impact the user less (only 2% of frame, so roughly every 1,5-5 sec depending on the framerate I guess?).

It's very possible this bug comes from this part though :-)

jailln:
Fixed in #1988

PR #821, opened 2018-07-25
refactor(3dtiles): improve 3dtiles tiles cleanup mechanism

peppsac:
Simplify 'garbage collection' and remove the randomness from it.

It may help / fix the  issue https://github.com/iTowns/itowns/issues/797 but I don't have any 3dtiles textured tileset to test / reproduce...

### DIFF
--- a/src/Process/3dTilesProcessing.js
+++ b/src/Process/3dTilesProcessing.js
@@ -236,20 +236,27 @@ export function pre3dTilesUpdate(context, layer) {
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
     context.camera.preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
 
-    // once in a while, garbage collect
-    if (Math.random() > 0.98) {
+    // Elements removed are added in the layer._cleanableTiles list.
+    // Since we simply push in this array, the first item is always
+    // the oldest one.
+    const now = Date.now();
+    if (layer._cleanableTiles.length
+        && (now - layer._cleanableTiles[0].cleanableSince) > layer.cleanupDelay) {
         // Make sure we don't clean root tile
         layer.root.cleanableSince = undefined;
 
-        // Browse
-        const now = Date.now();
-
-        for (const elt of layer._cleanableTiles) {
+        let i = 0;
+        for (; i < layer._cleanableTiles.length; i++) {
+            const elt = layer._cleanableTiles[i];
             if ((now - elt.cleanableSince) > layer.cleanupDelay) {
                 cleanup3dTileset(layer, elt);
+            } else {
+                // later entries are younger
+                break;
             }
         }
-        layer._cleanableTiles = layer._cleanableTiles.filter(n => (layer.tileIndex.index[n.tileId].loaded && n.cleanableSince));
+        // remove deleted elements from _cleanableTiles
+        layer._cleanableTiles.splice(0, i);
     }
 
     return [layer.root];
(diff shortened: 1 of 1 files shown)


src/Process/3dTilesProcessing.js
```diff
@@ -236,20 +236,27 @@ export function pre3dTilesUpdate(context, layer) {
     const HYFOV = 2.0 * Math.atan(Math.tan(radAngle * 0.5) * hypotenuse / context.camera.width);
     context.camera.preSSE = hypotenuse * (2.0 * Math.tan(HYFOV * 0.5));
 
-    // once in a while, garbage collect
-    if (Math.random() > 0.98) {
+    // Elements removed are added in the layer._cleanableTiles list.
+    // Since we simply push in this array, the first item is always
+    // the oldest one.
+    const now = Date.now();
+    if (layer._cleanableTiles.length
```
zarov, line 243: In `3dTilesProvider`, maybe `_cleanableTiles` could be declared before the `preUpdate` method and this test could be avoided ?

src/Process/3dTilesProcessing.js
```diff
@@ -236,20 +236,27 @@ export function pre3dTilesUpdate(context, layer) {
...
         // Make sure we don't clean root tile
         layer.root.cleanableSince = undefined;
 
-        // Browse
-        const now = Date.now();
-
-        for (const elt of layer._cleanableTiles) {
+        let i = 0;
+        for (; i < layer._cleanableTiles.length; i++) {
+            const elt = layer._cleanableTiles[i];
```
zarov, line 250: Since I think you are going for optimization here, maybe move the `const` declaration outside the loop ?

src/Process/3dTilesProcessing.js
```diff
@@ -236,20 +236,27 @@ export function pre3dTilesUpdate(context, layer) {
...
             if ((now - elt.cleanableSince) > layer.cleanupDelay) {
                 cleanup3dTileset(layer, elt);
+            } else {
+                // later entries are younger
+                break;
             }
         }
-        layer._cleanableTiles = layer._cleanableTiles.filter(n => (layer.tileIndex.index[n.tileId].loaded && n.cleanableSince));
+        // remove deleted elements from _cleanableTiles
+        layer._cleanableTiles.splice(0, i);
```
zarov, line 259: :+1:

src/Process/3dTilesProcessing.js
```diff
@@ -328,7 +335,10 @@ export function process3dTilesNode(cullingTest = $3dTilesCulling, subdivisionTes
 
 
         if (isVisible) {
-            node.cleanableSince = undefined;
+            if (node.cleanableSince) {
+                layer._cleanableTiles.splice(layer._cleanableTiles.indexOf(node), 1);
```
zarov, line 339: Why remove this one ?

zarov (commented):
Looks good, I'll test with our data set and keep you updated. Thanks !

src/Process/3dTilesProcessing.js
peppsac: Because it's not cleanable anymore - and I want the layer._cleanableTiles` array to only cleanable elements.

src/Process/3dTilesProcessing.js
peppsac: - this PR is not really about optimizing - I moved away from the `for .. of` loop because I need the index anyway (cf `layer._cleanableTiles.splice(0, i);` after)
- the `const elet = ...` depends on the loop indice, so I cannot move it outside

src/Process/3dTilesProcessing.js
peppsac: This test is to check if the `_cleanableTiles` is empty - if it's empty we skip the garbage collection mechanism.
So I don't really understand your suggestion?

zarov: Yep I was mistaken, please dismiss my comment.

src/Process/3dTilesProcessing.js
zarov: Ok !

src/Process/3dTilesProcessing.js
zarov: Ok !

zarov (approved):
So it seems to have improved a little, but I'll keep #797 still open. Thanks for this !

PR #825, opened 2018-07-31
Fix vector tiles

gchoqueux:
## Description

in TMS/VectorTile, wrong projection convert for globe
Because, vector tile's projection is 'EPSG:4326' and texture's projection is 'EPSG:3857'

## Screenshots
**after fix**
vector tile
**before fix**
image

### DIFF
--- a/examples/globe_vector_tiles.html
+++ b/examples/globe_vector_tiles.html
@@ -28,7 +28,11 @@
             var viewerDiv = document.getElementById('viewerDiv');
 
             // Instanciate iTowns GlobeView*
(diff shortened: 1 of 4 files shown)


zarov:
Thanks for this ! The failed CI comes from the change of the return value, which impacts other parts of itowns. That change is necessary, as there seems to be a bug with `babel` (but I can't find what is it really), so I'll propose another structure of the returning value in `GeojsonParser` tomorrow.

zarov (approved):
Thanks for the fix, it looks really better this time !

PR #827, opened 2018-08-01
fix xBilParser

### LINKED ISSUE #823: Tiles not loading in PlanarView
Since #820, there are some problems to display bottom screen tiles in planar mode (see screenshot below). The issue is similar to #776 in what we get and what we should get.

image

I'm assigning @gchoqueux since he is the author of the PR, but anyone feel free to fix it.

gchoqueux:
Failing to determine the min and max elevation doesn't mean that it is zero
fix #823


zarov (approved):
That does it, thanks !

PR #829, opened 2018-08-08
Helper to use ThreeJS Loader in the examples

gliegard:
Helper to use ThreeJS loaders in the examples.

## Description
- Load script asynchronously
- Load ThreeJs example loaders script
- Create Promise to load ressources asynchronously with these loaders
- Use this helper in Collada example

## Motivation and Context
I want the ThreeJS loaders to be easy to use in the examples.

### DIFF
--- a/examples/collada.html
+++ b/examples/collada.html
@@ -13,6 +13,7 @@
         <div id="viewerDiv"></div>
         <script src="js/GUI/GuiTools.js"></script>
         <script src="../dist/itowns.js"></script>
+        <script src="js/ThreeLoader.js"></script>
         <script src="js/loading_screen.js"></script>
         <script src="../dist/debug.js"></script>
         <script type="text/javascript">
@@ -54,52 +55,38 @@
             // These will deform iTowns globe geometry to represent terrain elevation.
             promiseElevation.push(itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addLayerCb));
             promiseElevation.push(itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addLayerCb));
-
-            function loadCollada(url) {
-                var model;
-                // loading manager
-                var loadingManager = new itowns.THREE.LoadingManager(function _addModel() {
-                    globeView.scene.add(model);
-                    globeView.notifyChange();
-                });
-                // collada loader
-                var loader = new itowns.THREE.ColladaLoader(loadingManager);
+            
+            // ThreeLoader can load each format proposed in ThreeJs examples loaders : https://github.com/mrdoob/three.js/tree/dev/examples/js/loaders
+            var promiseCollada = ThreeLoader.load('Collada', 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/collada/building.dae')
+            .then(collada => {
+                var model = collada.scene;
 
                 // building coordinate
                 var coord = new itowns.Coordinates('EPSG:4326', 4.2165, 44.844, 1417);
+                var colladaID = globeView.mainLoop.gfxEngine.getUniqueThreejsLayer();
 
-                loader.load(url, function col(collada) {
-                    var colladaID = globeView.mainLoop.gfxEngine.getUniqueThreejsLayer();
-                    model = collada.scene;
-                    model.position.copy(coord.as(globeView.referenceCrs).xyz());
-                    // align up vector with geodesic normal
-                    model.lookAt(model.position.clone().add(coord.geodesicNormal));
-                    // user rotate building to align with ortho image
-                    model.rotateZ(-Math.PI * 0.2);
-                    model.scale.set(1.2, 1.2, 1.2);
+                model.position.copy(coord.as(globeView.referenceCrs).xyz());
+                // align up vector with geodesic normal
+                model.lookAt(model.position.clone().add(coord.geodesicNormal));
+                // user rotate building to align with ortho image
+                model.rotateZ(-Math.PI * 0.2);
+                model.scale.set(1.2, 1.2, 1.2);
 
-                    // set camera's layer to do not disturb the picking
-                    model.traverse(function _(obj) { obj.layers.set(colladaID); });
-                    globeView.camera.camera3D.layers.enable(colladaID);
+                // set camera's layer to do not disturb the picking
+                model.traverse(function _(obj) { obj.layers.set(colladaID); });
+                globeView.camera.camera3D.layers.enable(colladaID);
 
-                    // update coordinate of the mesh
-                    model.updateMatrixWorld();
-                });
-            };
+                // update coordinate of the mesh
+                model.updateMatrixWorld();
+
+                globeView.scene.add(model);
+                globeView.notifyChange();
+            });
 
             // Listen for globe full initialisation event
             globeView.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function init() {
                 globeView.controls.setOrbitalPosition({ heading: 180, tilt: 60 });
             });
-
-            var script = document.createElement('script');
-            var THREE = itowns.THREE;
-            script.type = 'text/javascript';
-            script.src = 'https://cdn.rawgit.com/mrdoob/three.js/r' + THREE.REVISION + '/examples/js/loaders/ColladaLoader.js';
-            script.onload = function l() {
-                loadCollada('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/collada/building.dae');
-            };
-            document.body.appendChild(script);
         </script>
     </body>
 </html>
--- a/examples/js/ThreeLoader.js
+++ b/examples/js/ThreeLoader.js
@@ -1,0 +1,77 @@
+/* global itowns, Promise */
+var manager = new itowns.THREE.LoadingManager();
+
+// namespace ThreeLoader.
+var ThreeLoader = {};
+
+// Utility method to use Promise.
+function defer() {
+    var deferredPromise = {};
+    deferredPromise.promise = new Promise(function _(resolve, reject) {
+        deferredPromise.resolve = resolve;
+        deferredPromise.reject = reject;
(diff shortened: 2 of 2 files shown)


examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
...
+    var res;
+    var promise = new Promise(function p(resolve) {
+        res = resolve;
+    });
+    // eslint-disable-next-line no-undef
+    THREE = itowns.THREE;
+    loadScriptAsync('https://cdn.rawgit.com/mrdoob/three.js/r' + itowns.THREE.REVISION + '/examples/js/loaders/' + format + 'Loader.js')
+    .then(function createLoader() {
+        res(new itowns.THREE[format + 'Loader'](manager));
+    });
```
zarov: You should add a catch clause with a warning, in case the loader is not available.

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Load a ressource asynchronously using ThreeJs example loader.
+ * @param {String} format Format to load
+ * @param {String} url URL of the ressource
+ * @returns {Promise} A promise resolved when the ressource is loaded.
+ */
+// eslint-disable-next-line no-unused-vars
+function load(format, url) {
+    // return getThreeJsLoader(format).then(loader => useThreeJsLoader(loader, url));
```
zarov: To remove

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Load a script asynchronously
+ * @param {String} uri URI of the script to load.
+ * @returns {Promise} A promise resolved when the script is loaded.
+ */
+function loadScriptAsync(uri) {
+    return new Promise(function p(resolve) {
+        var firstScriptTag = document.getElementsByTagName('script')[0];
+        var tag = document.createElement('script');
+        tag.src = uri;
```
zarov: Should be put after `tag.onload`

examples/collada.html
```diff
@@ -54,52 +55,37 @@
...
-                    // set camera's layer to do not disturb the picking
-                    model.traverse(function _(obj) { obj.layers.set(colladaID); });
-                    globeView.camera.camera3D.layers.enable(colladaID);
-
-                    // update coordinate of the mesh
-                    model.updateMatrixWorld();
-                });
-            };
+                var colladaID = globeView.mainLoop.gfxEngine.getUniqueThreejsLayer();
+                // model =
```
zarov: To remove

zarov (changes requested):
Thanks for adding this, this will be useful ! You could have even broaden the usage by letting the user to specify a more complete path (not only a Loader). But otherwise it is good, thanks !

examples/collada.html
```diff
@@ -54,52 +55,37 @@
...
-                var model;
-                // loading manager
-                var loadingManager = new itowns.THREE.LoadingManager(function _addModel() {
-                    globeView.scene.add(model);
-                    globeView.notifyChange();
-                });
-                // collada loader
-                var loader = new itowns.THREE.ColladaLoader(loadingManager);
+            
+            var promiseCollada = load('Collada', 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/collada/building.dae')
```
gchoqueux: - You should explain how to write the `format`. Because you can not write 'collada' or 'dae', because it's according to the file name of the loader. You could also link with [this page](https://github.com/mrdoob/three.js/tree/dev/examples/js/loaders).

- a another method :

```js
const supportedLoaders = new Map([
    ['collada', ColladaLoader],
    ['dae', ColladaLoader],
    ['fbx', FBXLoader],
    //  ...
]);
 const format = 'Dae';
 const getLoaderNameFile = supportedLoaders.get(format.toLowerCase());
if (!getLoaderNameFile) {
 // unsupported format 
}
```

- You could add the space name `load` to clarify where the loader comes from :
`ThreeLoader.load(format, url)` because the loader return THREE object

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
+/* global itowns, Promise */
```
gchoqueux, line 1: would not it be better to make a package like Debug ,?

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
...
+var manager = new itowns.THREE.LoadingManager();
+
+/**
+ * Load a script asynchronously
+ * @param {String} uri URI of the script to load.
```
gchoqueux: the documentation is generated?

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
...
+ * @returns {Promise} A promise with the ThreeJS Loader
+ */
+function getThreeJsLoader(format) {
+    var res;
+    var promise = new Promise(function p(resolve) {
+        res = resolve;
+    });
+    // eslint-disable-next-line no-undef
+    THREE = itowns.THREE;
+    loadScriptAsync('https://cdn.rawgit.com/mrdoob/three.js/r' + itowns.THREE.REVISION + '/examples/js/loaders/' + format + 'Loader.js')
```
gchoqueux: you could add cache to get loader if it is already loaded

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * Load a ressource asynchronously using an already created loader
+ * @param {Loader} loader Loader from ThreeJs example loaders
+ * @param {String} url URL of the ressource
+ * @returns {Promise} A promise resolved when the ressource is loaded.
+ */
+function useThreeJsLoader(loader, url) {
+    var res;
```
gchoqueux: As you often use this structure you can declare it only once

```js
function defer() {
    const deferedPromise = {};
    deferedPromise.promise = new Promise((resolve, reject) => {
        deferedPromise.resolve = resolve;
        deferedPromise.reject = reject;
    });
    return deferedPromise;
}
```

gchoqueux (changes requested):
very nice tool ;)

examples/js/ThreeLoader.js
gchoqueux: because `Promise` it's not supported in old browser

examples/js/ThreeLoader.js
gliegard: The developer can use ThreeLoader.getThreeJsLoader if he wants to re use the loader.
So I think a cache is not needed.

examples/collada.html
gliegard: ok for the namespace

examples/js/ThreeLoader.js
gliegard: since it's in a namespace, it's not in the output of the generated doc when I add it in jsdoc-config.json

examples/js/ThreeLoader.js
gliegard: Thank's, nice refactoring idea !!
It removes duplicate code ;)

gliegard:
updated

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+ThreeLoader.getThreeJsLoader = function getThreeJsLoader(format) {
+    var deferedPromise = defer();
+    // eslint-disable-next-line no-undef
+    THREE = itowns.THREE;
+    loadScriptAsync('https://cdn.rawgit.com/mrdoob/three.js/r' + itowns.THREE.REVISION + '/examples/js/loaders/' + format + 'Loader.js')
+    .then(function createLoader() {
+        deferedPromise.resolve(new itowns.THREE[format + 'Loader'](manager));
+    }).catch(function error(e) {
+        console.error('Error creating', format, 'loader : ', e);
```
zarov: The message should be that the script is not/no longer available and could not be loaded.

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,78 @@
...
+function loadScriptAsync(uri) {
+    var deferedPromise = defer();
+    var firstScriptTag = document.getElementsByTagName('script')[0];
+    var tag = document.createElement('script');
+    tag.async = true;
+    tag.onload = function r() {
+        deferedPromise.resolve();
+    };
+    tag.src = uri;
+    firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
```
zarov: No need for that imho, use something like `document.body.append(tag)` instead.

examples/js/ThreeLoader.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+ThreeLoader.useThreeJsLoader = function useThreeJsLoader(loader, url) {
+    var deferedPromise = defer();
+    loader.load(url, deferedPromise.resolve, function _() {}, deferedPromise.reject);
+    return deferedPromise.promise;
+};
+
+/**
+ * Load a ressource asynchronously using ThreeJs example loader.
+ * @param {String} format Format to load. Format can be one of these : https://github.com/mrdoob/three.js/tree/dev/examples/js/loaders
```
zarov: This needs to be more precise, like the user needs to take the name of the loader file minus `Loader.js` or something like that.

zarov (commented):
Almost good !

examples/js/ThreeLoader.js
gliegard: The catch is not about script available, but is about the createLoader function

examples/js/ThreeLoader.js
gliegard: ok, good idea !

examples/js/ThreeLoader.js
gliegard: yep

examples/js/ThreeLoader.js
zarov: Ok you can dismiss the `catch` then, I didn't see that

zarov (approved):

gchoqueux (approved):

gliegard:
Thanks

PR #830, opened 2018-08-09
Orientation Utils

gliegard:
## Description
OrientationUtils create quaternion from rotation properties : roll, pitch, heading or omega, phi, kappa.
It can also compute ENU (East North Up) quaternion from a coordination if needed (it's needed in globe view).
Two examples, for globe and planar, show a car oriented at many points of a path, in Paris.

## Motivation and Context
Long-term goal : Immersive view
Next step : Use this parser to add camera oriented

## Screenshots
orientationparser


examples/index.html
```diff
@@ -137,6 +137,17 @@ <h3><a href='./planar_vector_tiles.html' target='_blank'>Plane mode
             <img src='screenshots/planar_vector_tiles.jpg'>
             <iframe src="" style="display: none;"></iframe>
         </div>
+        <div class="exampleContainer">
+            <h3><a href='./orientation_parser.html' target='_blank'>Orientation parser</a></h3>
+            <img src='screenshots/orientation_parser.jpg'>
+            <iframe src="" style="display: none;"></iframe>
+        </div>
+        </div>
+        <div class="exampleContainer">
```
zarov: I don't feel that we should always add a globe and a planar example everytime we do a feature. What about keeping only the globe one ?

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,111 @@
+import * as THREE from 'three';
+
+const DEG2RAD = THREE.Math.DEG2RAD;
+
+// The transform from world to local is  RotationZ(heading).RotationX(pitch).RotationY(roll)
+// The transform from local to world is (RotationZ(heading).RotationX(pitch).RotationY(roll)).transpose()
+THREE.Quaternion.prototype.setFromRollPitchHeading = function setFromRollPitchHeading(roll = 0, pitch = 0, heading = 0) {
```
zarov: I'm sorry for being always like this, but I don't like changing the prototype of external things of itowns. Why not make those methods static and add a `target` parameter to them to replace the `this` ?

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,111 @@
...
+// RotationX(omega).RotationY(phi).RotationZ(kappa).RotationX(PI)
+// RotationX(PI) = Scale(1, -1, -1) converts between the 2 conventions for the camera local frame:
+//  X right, Y bottom, Z front : convention in webGL, threejs and computer vision
+//  X right, Y top,    Z back  : convention in photogrammetry
+THREE.Quaternion.prototype.setFromOmegaPhiKappa = function setFromOmegaPhiKappa(omega = 0, phi = 0, kappa = 0) {
+    omega *= DEG2RAD;
+    phi *= DEG2RAD;
+    kappa *= DEG2RAD;
+    this.setFromEuler(new THREE.Euler(omega, phi, kappa, 'XYZ'));
+    // this.setFromRotationMatrix(new THREE.Matrix4().makeRotationFromQuaternion(this).scale(new THREE.Vector3(1, -1, -1)));
```
zarov: To remove

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {Vector3} up - the normalized geodetic normal to the ellipsoid (given by Coordinates.geodeticNormal)
+ * @returns {Quaternion} this
+ */
+THREE.Quaternion.prototype.setENUFromGeodesicNormal = (() => {
+    const matrix = new THREE.Matrix4();
+    const elements = matrix.elements;
+    const north = new THREE.Vector3();
+    const east = new THREE.Vector3();
+    return function setENUFromGeodesicNormal(up) {
+        // this is an optimized version of matrix.lookAt(up, new THREE.Vector3(0, 0, 0), new THREE.Vector3(0, 0, 1));
```
zarov: Could you elaborate on this optimization ?

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,111 @@
...
+// The transform from local to world is (RotationZ(heading).RotationX(pitch).RotationY(roll)).transpose()
+THREE.Quaternion.prototype.setFromRollPitchHeading = function setFromRollPitchHeading(roll = 0, pitch = 0, heading = 0) {
+    roll *= DEG2RAD;
+    pitch *= DEG2RAD;
+    heading *= DEG2RAD;
+    // return this.setFromEuler(new THREE.Euler(pitch, roll, heading , 'ZXY')).conjugate();
+    return this.setFromEuler(new THREE.Euler(-pitch, -roll, -heading, 'YXZ')); // optimized version of above
+};
+
+// From DocMicMac, the transform from local to world is:
```
zarov: What is DocMicMac ? Can you put a link to it ?

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,111 @@
...
+ * @function setFromAttitude
+ * @param {Attitude} attitude - [Attitude]{@link module:OrientedImageParser~Attitude}
+ * with properties: (omega, phi, kappa), (roll, pitch, heading) or none.
+ * @returns {THREE.Quaternion} this
+ */
+THREE.Quaternion.prototype.setFromAttitude = function setFromAttitude(attitude) {
+    if ((attitude.roll !== undefined) || (attitude.pitch !== undefined) || (attitude.heading !== undefined)) {
+        return this.setFromRollPitchHeading(attitude.roll, attitude.pitch, attitude.heading);
+    }
+    if ((attitude.omega !== undefined) || (attitude.phi !== undefined) || (attitude.kappa !== undefined)) {
```
zarov: Specify that this is a specific case for data coming from MicMac.

zarov (commented):
It's really nice to have an example illustrating this partial utilisation of the parser ! Also the example is quite detailed so thanks a lot for that.

Could you also maybe move the geojson to the iTowns2-sample-data repo ?

examples/index.html
mbredif: The globe example needs the local frame rotation (using `setENUFromGeodesicNormal`), and the planar does not. 2 examples show that it supports both cases. With one example, we need tests for the 2 cases.

zarov: Ok my bad I didn't read the planar example at all, I assumed it was the same one as the globe !

src/Parser/OrientationParser.js
mbredif: Another option is to have a CRS-aware subclass of THREE.Quaternion (itowns.Quaternion ?) with these added methods (kinda like itowns.Coordinate is a CRS-aware version of THREE.Vector3).

src/Parser/OrientationParser.js
mbredif: not really, eg : https://support.pcigeomatics.com/hc/en-us/articles/203483349-Heading-Pitch-Roll-vs-Omega-Phi-Kappa

src/Parser/OrientationParser.js
mbredif: the (0,0,0) and (0,0,1) values just make a few operations of Matrix4.lookAt trivial
(I wrote this code :) )

src/Parser/OrientationParser.js
mbredif: It is not really deadcode, it is the unoptimized version of the next line. If removed, the next line comment should state that it is an optimized version of a rotation by Pi around X.

src/Parser/OrientationParser.js
zarov: I'd prefer that yes, maybe we could give it a try.

src/Parser/OrientationParser.js
zarov: I read too fast once again ;)

src/Parser/OrientationParser.js
zarov: Ok thanks for the info

src/Parser/OrientationParser.js
zarov: I read too fast once again ;)

src/Parser/OrientationParser.js
zarov: How do you handle the case for `up == Vector3(0, 0, 0)` ? Does the result (being a matrix filled with 0 if I'm correct) works afterwards ?

src/Parser/OrientationParser.js
gliegard: ok, I've changed the code to use target parameter.
@mbredif No need to add another itownsQuaternion object since this code is the parser responsability.

gliegard:
@zarov updated

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,118 @@
...
+                    var model = collada.scene.clone();
+                    
+                    // get point coordinate
+                    var coord = feature.vertices[0]; 
+                    // set the ground altitude
+                    coord = coord.as('EPSG:4326');
+                    coord.setAltitude(0);
+        
+                    // set model position
+                    model.position.copy(coord.as(view.referenceCrs).xyz());
```
mbredif: could you use `feature.position` instead ?

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,118 @@
...
+                    model.rotateZ(-Math.PI / 2 );
+                    
+                    // set model scale
+                    model.scale.set(0.01, 0.01, 0.01);
+                                     
+                    // update matrix world and add the model
+                    model.updateMatrixWorld();
+                    view.scene.add(model);
+                }
+                view.notifyChange(undefined);
```
mbredif: why not `view.notifyChange()` ?

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,118 @@
...
+                    
+                    // apply another orientation (depending on model orientation)
+                    // in this example, we want the car to go straigth.
+                    model.rotateZ(-Math.PI / 2 );
+                    
+                    // set model scale
+                    model.scale.set(0.01, 0.01, 0.01);
+                                     
+                    // update matrix world and add the model
+                    model.updateMatrixWorld();
```
mbredif: Could you put 105-111 (which are relative to the particular collada model) before the clone in a `then` clause of the Collada loading promise ?

src/Parser/OrientationParser.js
mbredif: I think we could use crs-aware quaternion functions in other places sooner than later (eg in #795)

examples/orientation_parser_planar.html
gliegard: yep

examples/orientation_parser_planar.html
gliegard: It's needed to see the model on the scene

examples/orientation_parser_planar.html
gliegard: Yes good idea, only for the scale. So I move the scale call up.
But
RotateZ must be called after the quaternion changes.
and updateMatrixWorld must be called after all theses changes.

src/Parser/OrientationParser.js
mbredif: up is assumed to be a unit vector (@gliegard to be added to the docs!)

examples/orientation_parser_planar.html
mbredif: I thought `view.notifyChange()` and `view.notifyChange(undefined)` were equivalent ...

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,116 @@
...
+                for (const feature of features) {                  
+                
+                    // clone the collada model
+                    var model = colladaModel.clone();
+                    
+                    // get point coordinate
+                    var coord = feature.vertices[0]; 
+                    // set the ground altitude
+                    coord = coord.as('EPSG:4326');
+                    coord.setAltitude(0);
```
mbredif: I think you can remove these 3 lines : `coord` is unused

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,116 @@
...
+        
+                    // set model position
+                    feature.position.z = 0;
+                    model.position.copy(feature.position);
+                    
+                    // set model orientation
+                    model.quaternion.premultiply(feature.quaternion);
+
+                    model.updateMatrixWorld();
```
mbredif: Are you sure it is needed ?

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,116 @@
...
+                    // set the ground altitude
+                    coord = coord.as('EPSG:4326');
+                    coord.setAltitude(0);
+        
+                    // set model position
+                    feature.position.z = 0;
+                    model.position.copy(feature.position);
+                    
+                    // set model orientation
+                    model.quaternion.premultiply(feature.quaternion);
```
mbredif: (for both examples) I understand you need to premultiply since it already had a rotation... would it be clearer to add the parsed collada model to a new group and clone that group instead, so that a simple copy here would be sufficient ?

examples/orientation_parser_planar.html
gliegard: yes, thanks

examples/orientation_parser_planar.html
gliegard: yes

examples/orientation_parser_planar.html
gliegard: ok, I've updated the code with your idea.

examples/orientation_parser.html
```diff
@@ -0,0 +1,119 @@
...
+            // load collada model of a renault trafic
+            promises.push(ThreeLoader.load('Collada',
+             'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/collada/renault_trafic_no_texture.dae')
+             .then(collada => {
+                // prepare collada model base scale and rotation. in this example, we want the car to go straigth.
+                var colladaModel = collada.scene;
+                colladaModel.scale.set(0.5, 0.5, 0.5);
+                colladaModel.rotateZ(-Math.PI / 2 );
+                // put the model in a group, to allow user to do other transformation (rotation..)
+                var group = new itowns.THREE.Group();
```
gchoqueux: you can return directy `colladaModel`
and replace line 109 by
```js 
model.quaternion.premultiply(feature.quaternion);
```

examples/orientation_parser.html
```diff
@@ -0,0 +1,119 @@
...
+                    // clone the collada model
+                    var model = colladaModel.clone();
+                    
+                    // get point coordinate
+                    var coord = feature.vertices[0]; 
+                    
+                    // change the coordinate to have the object at the altitude of the ground
+                    result = itowns.DEMUtils.getElevationValueAt(globeView.wgs84TileLayer, coord, 1, undefined);
+                    coord = coord.as('EPSG:4326');
+                    coord.setAltitude(result.z);
```
gchoqueux: you should explain why you don't use the altitude from the geojson.
The altitude in geojson comes from Inertial measurement unit?

examples/orientation_parser.html
```diff
@@ -0,0 +1,119 @@
...
+            itowns.proj4.defs('EPSG:2154', '+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+            // Add a promise to load oriented point datas
+            promises.push(
+                // first parse the JSON format
+                itowns.Fetcher.json(panoramics, {})
+                // then get geographic informations (coordinates) using GeoJson parser
+                .then(orientations => itowns.GeoJsonParser.parse(orientations, layer))
+                // then get orientation informations 
+                .then(collection => itowns.OrientationParser.parse(collection.features, layer))
```
gchoqueux: It's not really a parser anymore. To be more generic and reusable without the features, it would take a function that takes directly attitude and crs.

```js
.then(collection => {
    const attitude = collection.features.properties;
    const quaternion = itowns.quaternionFromAttitude(attitude, globeView.referenceCrs));
    collection.features.quaternion = quaternion;
    collection.features.position = feature.vertices[0].xyz();
    return features;
}
```

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,110 @@
+import * as THREE from 'three';
```
gchoqueux: Global comment on `OrientationParser` : **very good optimization of calculations**

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,110 @@
...
+    const matrix = new THREE.Matrix4();
+    const elements = matrix.elements;
+    const north = new THREE.Vector3();
+    const east = new THREE.Vector3();
+    return function setENUFromGeodesicNormal(target = new THREE.Quaternion(), up) {
+        // this is an optimized version of matrix.lookAt(up, new THREE.Vector3(0, 0, 0), new THREE.Vector3(0, 0, 1));
+        east.set(-up.y, up.x, 0);
+        east.normalize();
+        north.crossVectors(up, east);
+        north.normalize();
```
gchoqueux: as `up` and `east` are normalized, `north.normalize()` isn't necessary

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,110 @@
...
+     * @param {string} options.crsOut - the CRS of View (must be specified (as 'EPSG:4978') for GlobeView)
+     * @return {Object} a promise that resolves with the features given in parameter
+     */
+    parse(features, options = {}) {
+        const needsENUFromGeodesicNormal = options.crsOut == 'EPSG:4978';
+
+        const ENUQuat = new THREE.Quaternion();
+        for (const feature of features) {
+            const coordinates = feature.vertices[0];
+            feature.position = coordinates.xyz();
```
gchoqueux: it is not necessary to copy coordinates in `feature.position`

gchoqueux (changes requested):

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,110 @@
...
+
+            // compute orientation
+            feature.quaternion = setFromAttitude(new THREE.Quaternion(), feature.properties);
```
gchoqueux: does this calculation of orientation, take into account the convergence of meridians ?

examples/orientation_parser.html
gchoqueux: if you want, I didn't see previous [comment ](https://github.com/iTowns/itowns/pull/830#discussion_r209332056)

examples/orientation_parser.html
gchoqueux: normally in itowns :
`parser` : transform `file` to `itowns.feature`
`convert` : `itowns.feature` to `THREE.Object`

and the `OrientationParser` computes `THREE.Quaternion` and `THREE.Vector3`

src/Parser/OrientationParser.js
gchoqueux: proj4 c++ computes [convergence of merdidian](https://proj4.org/apps/proj.html#id7), but not proj4js.  

It would be necessary for the user to be able to indicate the convergence so that it is introduced in the quaternion (or that itowns can be calculated by it).

It should at least be indicated that the quaternion in planar doesn't take into account the convergence of the meridians

gliegard: ok, I've added a comment saying convergence of meridian is not taken into account.
User need to knwo the convergence of meridian value and multiply quaternion himself.

src/Parser/OrientationParser.js
gliegard: ok, let's remove the position and move it in the example code.

src/Parser/OrientationParser.js
gliegard: ok

examples/orientation_parser.html
gliegard: ok i've exposed re usable methods

mbredif:
@gliegard : I gave a try to implement properly the meridian convergence in #832. Feel free to rebase my work in your PR.

edit: I seems you force pushed while I worked. only 77f01bea6a67f6f7c68e19c2b713127a5dd472a3 is relevant in #832.

gliegard:
@mbredif  Thank's for the work. Yes I rebase on each review/changes. So we should wait for this PR to be merged, then make the convergence in your PR.

@gchoqueux I've updated the code :)

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,124 @@
...
+     * @param {Object} options - additional properties.
+     * @param {string} options.crsOut - the CRS of View (must be specified (as 'EPSG:4978') for GlobeView)
+     * @return {Object} a promise that resolves with the features given in parameter
+     */
+    parse(features, options = {}) {
+        const needsENUFromGeodesicNormal = options.crsOut == 'EPSG:4978';
+
+        for (const feature of features) {
+            const coordinates = feature.vertices[0];
+            // feature.position = coordinates.xyz();
```
gchoqueux: remove dead code

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,124 @@
...
+
+    /** @module OrientationParser */
+    /** Parse features properties, and add position and quaternion attributes.
+     * @function parse
+     * @param {Object} features - a JSON array of Point Features
+     * @param {Object} options - additional properties.
+     * @param {string} options.crsOut - the CRS of View (must be specified (as 'EPSG:4978') for GlobeView)
+     * @return {Object} a promise that resolves with the features given in parameter
+     */
+    parse(features, options = {}) {
```
gchoqueux: use directly `crsOut` in parameter:
`parse(features, crsOut)`

examples/orientation_parser.html
```diff
@@ -0,0 +1,150 @@
...
+            };
+            // Input data are given in Lambert93 projection
+            itowns.proj4.defs('EPSG:2154', '+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+            // Add a promise to load oriented point datas
+            promises.push(
+                // first parse the JSON format
+                itowns.Fetcher.json(panoramics, {})
+                // then get geographic informations (coordinates) using GeoJson parser
+                .then(orientations => itowns.GeoJsonParser.parse(orientations, layer))
```
gchoqueux: don't use => in example, because is ES6

examples/orientation_parser.html
```diff
@@ -0,0 +1,150 @@
...
+                    
+                    // clone the collada model
+                    var model = colladaModel.clone();
+                    
+                    // set model orientation
+                    model.quaternion.copy(feature.quaternion);
+                  
+                    // copy feature position
+                    var position = feature.vertices[0].xyz();
+                    model.position.copy(position);
```
gchoqueux: you cloud write :

```js
// store base position
model.positionBase = feature.vertices[0].xyz();

model.position.copy(model.positionBase);
```

examples/orientation_parser.html
```diff
@@ -0,0 +1,150 @@
...
+                    model.positionBase = new itowns.THREE.Vector3();
+                    model.positionBase.copy(model.position);
+
+                    // compute position on the ground
+                    var coord = feature.vertices[0]; 
+                    result = itowns.DEMUtils.getElevationValueAt(globeView.wgs84TileLayer, coord, 1, undefined);
+                    coord = coord.as('EPSG:4326');
+                    coord.setAltitude(result.z);
+                    model.positionOnGround = new itowns.THREE.Vector3();
+                    model.positionOnGround.copy(coord.as(globeView.referenceCrs).xyz());
```
gchoqueux: you cloud write :

```js
model.positionOnGround  = coord.as(globeView.referenceCrs).xyz();
```

gchoqueux (commented):
- There's no test on parser?
- You can add example with `setFromAttitudeGlobe`?

src/Parser/OrientationParser.js
gliegard: Maybe we'll need other options in the future (ex meridian convergence..)

examples/orientation_parser.html
gliegard: ok

examples/orientation_parser.html
gliegard: ok

src/Parser/OrientationParser.js
gliegard: ok

gliegard:
> There's no test on parser?

Ok, I'm working on theses tests

> You can add example with setFromAttitudeGlobe?

This method is used indirectly in the orientatation parser example on the globe.
There are two examples (planar and globe) for this parser, I think it's enought.

examples/orientation_parser.html
gliegard: yes, thanks

examples/orientation_parser.html
```diff
@@ -0,0 +1,146 @@
...
+                    
+                    // clone the collada model
+                    var model = colladaModel.clone();
+                    
+                    // set model orientation
+                    model.quaternion.copy(feature.quaternion);
+                  
+                    // store base position
+                    model.positionBase = feature.vertices[0].xyz();
+                    model.position.copy(model.positionBase);
```
mbredif: what is the point of the positionBase property? could not you just use 

`feature.vertices[0].xyz(model.position)`

test/unit/orientationParser.js
```diff
@@ -0,0 +1,83 @@
+import * as THREE from 'three';
+import assert from 'assert';
+import OrientationParser from '../../src/Parser/OrientationParser';
+import Coordinates from '../../src/Core/Geographic/Coordinates';
+
+// Assert two quaternion obects are equals.
```
mbredif: typo: objects

mbredif (commented):
#832 improves both the docs and the applicability of this PR, but I am fine merging this one if the other follows quickly.

examples/orientation_parser.html
gliegard: In this example, we store 2 positions, positionBase (given by the input file) and positionOnGround (to have the car on the road again again)

test/unit/orientationParser.js
gliegard: oh thanks ;)

examples/orientation_parser.html
gliegard: positionBase is the position from the file, positionOnGround is the position on the ground, and position is the real current position used by ThreeJs.

src/Parser/OrientationParser.js
gchoqueux: you could move `OrientationParser.parse` in `GeoJsonParser`, and compute quaternion if you have attitude in properties and if type of feature is `point`. So can delete `parse` function from  `OrientationParser` and rename it `OrientationTools`

gliegard: Good idea ! 
So it will detect automatically if there is a quaternion information in geoJson properties.
I try that, thanks.

src/Parser/OrientationParser.js
```diff
@@ -0,0 +1,119 @@
...
+    /** Parse rotation information (roll pitch heading or omega phi kappa),
+     * and create a ThreeJs quaternion representing a rotation.
+     *
+     * @param {Attitude} attitude - [Attitude]{@link module:OrientationParser~Attitude}
+     * @param {Coordinates} coordinate position the oject (used to apply another rotation on Globe CRS)
+     * @param {Boolean} needsENUFromGeodesicNormal should be true on globe CRS.
+     * If true, we will apply another rotation : The rotation use to create ENU local space at coordinate parameter position.
+     * @returns {THREE.Quaternion} rotation
+     */
+    parse(attitude, coordinate, needsENUFromGeodesicNormal) {
```
gchoqueux: The geojson parsing  is done in the `GeoJsonParser`. All the responsibility for parsing should be in this file.
It's easier to parse the file once and it will give simpler examples

examples/orientation_parser_planar.html
```diff
@@ -0,0 +1,118 @@
...
+                    // get coordinate from point feature
+                    var coord = feature.vertices[0];
+
+                    // set model position to the ground
+                    var position = coord.xyz();
+                    position.z = 0;
+                    model.position.copy(position);
+                                        
+                    // set model orientation
+                    var quat = itowns.OrientationParser.parse(feature.properties);
```
gchoqueux: why not use directly `this.quaternionFromAttitude(attitude)` ?

gliegard: Yes I could, but I want to show examples that use of the top level function parse, 
same function called in planar and in globe, to show that the same function can handle planar and globe.

gchoqueux: You use the same function, but you don't use the same count of parameters (in planar or globe examples). So, if the user have this knowledge, he could directly use the right function.

gliegard:
@gchoqueux updated

gchoqueux (approved):

gchoqueux:
Thanks @gliegard

PR #832, opened 2018-08-14
Feat orientation parser conv meridian

mbredif:
## Description
- docs for OrientationParser
- systematic proj4 based parsing of the crs and support for quaternion computation for geocent and lcc projections.

## Motivation and Context
geocent was supported, but not lcc, which caused inaccuracies (meridian convergence was not taken into account)
I have setup a function cache so that adding new projection should be easy (except for the maths :) )


mbredif:
superseded by #845

PR #834, opened 2018-08-20
fix(core): chain animations stops after the first animation

gchoqueux:
chain animations stops after the first animation


zarov (approved):
LGTM thanks

PR #836, opened 2018-08-22
test: retry cubic_planar test if necessary

zarov:
This test fails at least one times out of three, and we always have to rerun it. Using `retries()` (see https://mochajs.org/#retry-tests) we can stop having to relaunch it everytime.

Note that this could be extended to every test, but I feel that this is the only one that really needs it.


zarov:
Updated to reduce the timeout of test from 5min to 1min (all tests are executed under 20s, but let's keep a margin here).

gchoqueux (approved):

PR #838, opened 2018-08-24
 feature(GeometryLayer): wireframe and opacity are updated with defineProperty

gchoqueux:
- feature(GeometryLayer): wireframe and opacity are updated with defineProperty

- fix(core): wrong picking with depth buffer
with wireframe == true, opacity < 0.0 and visibility == false

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -55,7 +55,7 @@
 
             // Add the UI Debug
             var d = new debug.Debug(view, menuGlobe.gui);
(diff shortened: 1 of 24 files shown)


zarov (commented):
This could be the opportunity to unify all base layers of views: `wgs84TileLayer`, `tileLayer` and `baseLayer`. And then we could move all the things you added to `View.js` only (and other things later).

gchoqueux:
@zarov updated

zarov (approved):
You still have a `wgs84TileLayer` call in `orientation_utils`, otherwise it is good, thanks for doing this, we should have done it a long time ago !

PR #840, opened 2018-08-27
refactor: move PlanarTileProcessing internals to PlanarLayer

zarov:
To continue the removal of some Processing files in Process, all
contents of PlanarTileProcessing have been moved to PlanarLayer.


src/Layer/TiledGeometryLayer.js
```diff
@@ -192,7 +192,7 @@ class TiledGeometryLayer extends GeometryLayer {
         }
 
         // do proper culling
-        node.visible = (!this.culling(node, context.camera));
+        node.visible = this.culling(node, context.camera);
```
gchoqueux: why you inverse?

zarov: There was an error, but it was because of the `this.culling = () => true` in `GeometryLayer`, that I removed. It shouldn't have been a placeholder.

gchoqueux (approved):

PR #847, opened 2018-09-21
refactor(gis): change ellipsoidSizes from function to object

zarov:
The `ellipsoidSizes` export in `Coordinates` was a function, which was
creating an object each time it was called. It has now be changed to an
object, meaning that we don't have to execute the function each time.

BREAKING CHANGE: `ellipsoidSizes` from `Coordinates.js` is no longer a
function but an object.


gchoqueux (approved):

mbredif:
should not we rely on the proj4 ellipsoid instead?

gchoqueux:
yes, you know how get ellipsoid size with Proj4js?

gchoqueux:
ok it's `proj4.WGS84.a` and `proj4.WGS84.b`

PR #849, opened 2018-10-03
refactor(renderer): LayeredMaterial splitted and simplified

zarov:
This a WIP from some time ago from @mbredif, rebased on the current master and adapted to reflect the last big changes we made in iTowns. I still want to do a few things on this, but I'm creating this PR to allow us to see where we are going.

Work has been done to simplify LayeredMaterial and its surrounding parts, especially with the management of textures in shaders.

LayeredMaterial has been splitted, and has now a LayeredMaterialLayer
property: this object will contain the textures linked to the material,
and uniforms related to each material.

Uniforms from LayeredMaterial are no longer accessible as is, and should
be accessible with directly using the uniform name. The material does
not do a lot of things, apart attaching a layer to it and managing
uniforms.

TileMesh has been simplified, removing all accessors in favor of direct
access to its properties. Material operations that were executed in it
have also been removed.

Shaders have been rewritten to reflect change quoted above. In an
attempt to simplify things and make them more comprehensible, TileFS and
TileVS have been splitted in multiple chunks, which are installed
through ShaderChunk.install.

BREAKING CHANGES: the API of TileMesh and LayeredMaterial have changed.
BREAKING CHANGES: LayeredMaterialConstants and RendererConstant have
been removed.

@mbredif I'm counting on you to tell me what I've missed when rebasing, and provide a better commit message ;)


examples/multiglobe.html
```diff
@@ -58,7 +58,7 @@
             object3ds.push(object3d);
 
             // define pole texture
-            globe2.noTextureColor = new itowns.THREE.Color(0xd0d5d8);
+            globe2.diffuse = new itowns.THREE.Color(0xd0d5d8);
```
gchoqueux, line 61: is breaking change

examples/split.html
```diff
@@ -84,14 +84,13 @@
...
-            function changeLayerVisibility(ortho, osm) {
-                var material;
-
+            function changeLayerVisibility(orthoVisible, osmVisible) {
                 globeView.scene.traverse(function _(obj) {
-                    if (obj.material && obj.material.setLayerVisibility && obj.material.visible) {
-                        material = obj.material;
-                        material.setLayerVisibility(orthoLayer, ortho);
-                        material.setLayerVisibility(osmLayer, osm);
+                    if (obj.material && obj.material.visible && obj.material.getLayer) {
```
gchoqueux, line 89: you could use `layer.visible` directly?

src/Core/Prefab/Globe/GlobeLayer.js
```diff
@@ -170,9 +169,10 @@ class GlobeLayer extends TiledGeometryLayer {
         // Prevent to subdivise the node if the current elevation level
         // we must avoid a tile, with level 20, inherits a level 3 elevation texture.
         // The induced geometric error is much too large and distorts the SSE
-        const currentTexture = node.material.textures[l_ELEVATION][0];
-        if (currentTexture.extent) {
-            const offsetScale = node.material.offsetScale[l_ELEVATION][0];
+        const nodeLayer = node.material.getElevationLayer();
```
gchoqueux, line 172: could you rename in `elevationLayer`

src/Layer/TiledGeometryLayer.js
```diff
@@ -250,11 +250,13 @@ class TiledGeometryLayer extends GeometryLayer {
      * otherwise.
      */
     static hasEnoughTexturesToSubdivide(context, node) {
+        const layerUpdateState = node.layerUpdateState || {};
         for (const e of context.elevationLayers) {
             const extents = node.getCoordsForSource(e.source);
-            if (!e.frozen && e.ready && e.source.extentsInsideLimit(extents) && !node.material.isElevationLayerLoaded()) {
+            const nodeLayer = node.material.getElevationLayer();
+            if (!e.frozen && e.ready && e.source.extentsInsideLimit(extents) && (!nodeLayer || nodeLayer.level < 0)) {
```
gchoqueux, line 258: may be use a new function:
`material.isLayerLoaded(layer)`
```js
if (!e.frozen && e.ready && e.source.extentsInsideLimit(extents) && !material.isLayerLoaded(e))) {
// ...
}
```

src/Layer/TiledGeometryLayer.js
```diff
@@ -266,11 +268,12 @@ class TiledGeometryLayer extends GeometryLayer {
...
             }
             // no stop subdivision in the case of a loading error
-            if (node.layerUpdateState[c.id] && node.layerUpdateState[c.id].inError()) {
+            if (layerUpdateState[c.id] && layerUpdateState[c.id].inError()) {
                 continue;
             }
             const extents = node.getCoordsForSource(c.source);
-            if (c.source.extentsInsideLimit(extents) && !node.material.isColorLayerLoaded(c)) {
+            const nodeLayer = node.material.getLayer(c.id);
+            if (c.source.extentsInsideLimit(extents) && (!nodeLayer || nodeLayer.level < 0)) {
```
gchoqueux, line 277: see previous comment

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -27,31 +26,20 @@ function getSourceExtent(node, extent, targetLevel, source) {
 }
 
 function initNodeImageryTexturesFromParent(node, parent, layer) {
-    if (parent.material && parent.material.isColorLayerLoaded(layer)) {
+    const parentLayer = parent.material && parent.material.getLayer(layer.id);
+    const nodeLayer = node.material && node.material.getLayer(layer.id);
```
gchoqueux, line 30: it's necessary to test if `node.material` is `defined`?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -27,31 +26,20 @@ function getSourceExtent(node, extent, targetLevel, source) {
...
-                    node.material.offsetScale[l_COLOR][textureIndex] = result;
-                    textureIndex++;
+                    nodeLayer.setTexture(index, texture, c.offsetToParent(texture.coords));
                     break;
                 }
             }
+            index++;
         }
-
-        if (__DEBUG__) {
```
gchoqueux: why your remove this test ?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -98,28 +87,30 @@ function getIndiceWithPitch(i, pitch, w) {
 }
 
 function insertSignificantValuesFromParent(texture, node, parent, layer) {
```
gchoqueux, line 95: what does this function do?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -157,9 +148,16 @@ function checkNodeElevationTextureValidity(texture, noDataValue) {
...
+    if (!parent || !material) {
+        return;
+    }
+
     const extentsDestination = node.getCoordsForSource(layer.source);
 
+    let nodeLayer = material.getLayer(layer.id);
+
     // Initialisation
+    node.layerUpdateState = node.layerUpdateState || {};
```
gchoqueux: it's better to declare in `node.constructor`

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -168,23 +166,30 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
...
+            const colorLayer = {
+                tileMT: layer.options.tileMatrixSet || extentsDestination[0].crs(),
+                texturesCount: extentsDestination.length,
+                visible: layer.visible,
+                opacity: layer.opacity,
+                effect: layer.fx,
+                id: layer.id,
+            };
+
+            nodeLayer = material.addLayer(colorLayer);
```
gchoqueux, line 194: you could write  : 
```js
nodeLayer = material.addLayer(layer, countTextures);
```

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -194,23 +199,31 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
...
     // to avoid mixing layer's network updates and layer's params
     // Update material parameters
-    material.setLayerVisibility(layer, layer.visible);
-    material.setLayerOpacity(layer, layer.opacity);
+    if (nodeLayer) {
+        const values = {
+            visible: layer.visible,
+            opacity: layer.opacity,
+        };
+        if (nodeLayer.setValues) {
```
gchoqueux, line 226: why `setValues` or `assign`?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -264,21 +276,23 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
         layer,
         extentsSource,
         extentsDestination,
-        parsedData: node.material.getLayerTextures(layer)[0].parsedData,
+        parsedData: layer.source.parsedData,
```
gchoqueux, line 284: no, it's `parsedData` attached to texture

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -264,21 +276,23 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
...
+            if (material === null) {
                 return;
             }
 
-            if (result) {
-                const pitchs = extentsDestination.map((ext, i) => ext.offsetToParent(extentsSource[i]));
-                node.material.setLayerTextures(layer, result, pitchs);
+            const pitchs = extentsDestination.map((ext, i) => ext.offsetToParent(extentsSource[i]));
+            if (Array.isArray(result)) {
+                nodeLayer.setTextures(result, pitchs);
```
gchoqueux, line 298: `result` is always a `Array`

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -376,43 +393,56 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
         extentsSource,
         extentsDestination,
         requester: node,
-        priority: nodeCommandQueuePriorityFunction(node),
+        targetLevel,
```
gchoqueux: `targetLevel` can be removed from `command`

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -376,43 +393,56 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
...
             node.layerUpdateState[layer.id].success();
 
-            if (terrain.texture && terrain.texture.image.data && !checkNodeElevationTextureValidity(terrain.texture, layer.noDataValue)) {
-                // Quick check to avoid using elevation texture with no data value
-                // If we have no data values, we use value from the parent tile
-                // We should later implement multi elevation layer to choose the one to use at each level
-                insertSignificantValuesFromParent(terrain.texture, node, parent, layer);
-            }
+            if (elevation.texture) {
+                if (elevation.texture.flipY) {
```
gchoqueux, line 424: where comes from this part?

src/Renderer/LayeredMaterial.js
```diff
@@ -31,31 +22,106 @@ export function unpack1K(color, factor) {
...
+    let _value = initValue !== undefined ? initValue : defaultValue;
+    layerNeedsUpdate = layerNeedsUpdate || (() => false);
+    Object.defineProperty(layer, property, {
+        get: () => _value,
+        set: (value) => {
+            if (_value !== value) {
+                layer.needsUpdate |= layerNeedsUpdate(_value, value);
+                layer.material.uniformsNeedUpdate = true;
+                _value = value;
+                layer.updateUniforms();
```
gchoqueux: This update shouldn't be here, because it's call for each change.
`updateUniforms` should be one time just before rendering

src/Renderer/LayeredMaterial.js
```diff
@@ -31,31 +22,106 @@ export function unpack1K(color, factor) {
...
+        for (const texture of this.textures) {
+            if (texture instanceof THREE.Texture) {
+                texture.dispose();
+            }
+        }
+        this.level = EMPTY_TEXTURE_ZOOM;
+        this.textures = [];
+        this.offsetScales = [];
+        this.needsUpdate = needsUpdate;
+        this.updateUniforms();
```
gchoqueux: why `updateUniforms` ? `dispose` is call when the tile is removed

src/Renderer/LayeredMaterial.js
```diff
@@ -31,31 +22,106 @@ export function unpack1K(color, factor) {
...
+    setTexture(index, texture, offsetScale) {
+        this.level = (texture && (index == 0)) ? texture.coords.zoom : this.level;
+        this.textures[index] = texture || null;
+        this.offsetScales[index] = offsetScale || identityOffsetScale;
+        this.needsUpdate = true;
+        this.updateUniforms();
+    }
+
+    setTextures(textures, pitchs) {
+        this.dispose(false);
```
gchoqueux: I don't think we should dispose these textures because they must belong to the parents.

src/Renderer/LayeredMaterial.js
```diff
@@ -31,31 +22,106 @@ export function unpack1K(color, factor) {
...
+        const autoUpdate = this.autoUpdate;
+        this.autoUpdate = false;
```
gchoqueux: `this.autoUpdate` is complex, you could remove it with one update before rendering

gchoqueux (changes requested):

examples/multiglobe.html
mbredif: I agree that this is a breaking change, my point here is to reuse THREE.js concepts/names wherever they exist

examples/split.html
mbredif: I am not sure which layer you are referring to. Could you propose a specific change here ?

src/Core/Prefab/Globe/GlobeLayer.js
mbredif: That could be done, but the naming nodeLayer tries to prepare for a common handling of color and elevation layers. (maybe it is too early :) )

examples/multiglobe.html
gchoqueux: yes, i agree to use `THREE.Material` concepts. 
We should also use other concepts like textures, lights ...

src/Core/Prefab/GlobeView.js
```diff
@@ -280,26 +272,23 @@ GlobeView.prototype.setRealisticLightingOn = function setRealisticLightingOn(val
...
 GlobeView.prototype.setLightingPos = function setLightingPos(pos) {
     const lightingPos = pos || CoordStars.getSunPositionInScene(this.ellipsoid, new Date().getTime(), 48.85, 2.35);
 
-    this.updateMaterialUniform('lightPosition', lightingPos.clone().normalize());
+    this.updateMaterialProperty('lightPosition', lightingPos.clone().normalize());
     this.notifyChange(this.tileLayer);
 };
 
-GlobeView.prototype.updateMaterialUniform = function updateMaterialUniform(uniformName, value) {
+GlobeView.prototype.updateMaterialProperty = function updateMaterialProperty(property, value) {
```
mbredif, line 287: The rationale here is that a material property is a uniform or a definition is an implementation detail of the material. Code base outside materials should not be aware of uniforms and defines.

src/Renderer/LayeredMaterial.js
mbredif: who says so ? It is used here to reset the textures : https://github.com/iTowns/itowns/pull/849/files#diff-f5c5a4b625fe789476515783a6e21153R113

src/Renderer/LayeredMaterial.js
mbredif: What happens if a image tile spans 2 geometric tiles ? is the imagery texture duplicated (which wastes memory) ? if not, which of the 2 tiles is responsible for disposing the texture ? I would argue to use a reference counting approach (smart pointer) so that no one has the responsibility in particular.

gchoqueux: For the moment, these data are handled by the `Cache` object. Which limits the duplication of objects (like textures by examples)

src/Core/Prefab/Globe/GlobeLayer.js
gchoqueux: ok

src/Process/LayeredMaterialNodeProcessing.js
zarov: No, I removed it

src/Layer/TiledGeometryLayer.js
zarov: Maybe more something like `nodeLayer.isLoaded()` ?

src/Process/LayeredMaterialNodeProcessing.js
zarov: Done

src/Process/LayeredMaterialNodeProcessing.js
zarov: What would we benefit from knowing this ?

src/Process/LayeredMaterialNodeProcessing.js
zarov: What should it be then ?

src/Renderer/LayeredMaterial.js
zarov: Yes, `updateUniforms` needs to be changed anyway, as it is called too much

src/Renderer/LayeredMaterial.js
mbredif: I understand you say that textures are shared (because thanks to the Cache). Then what prevents multiple disposal of the a texture spanning multiple geometric tiles ?

src/Process/LayeredMaterialNodeProcessing.js
mbredif: that adds a constraint on `node`... IMHO, the requirement should only be `THREE.Object3D` with our `LayeredMaterial`.

src/Process/LayeredMaterialNodeProcessing.js
mbredif: `Object.assign` works in all cases, but `setValues` is better if present (error checking, optimized auto updating...)

src/Renderer/LayeredMaterial.js
```diff
@@ -67,459 +132,233 @@ export function getMaxColorSamplerUnitsCount() {
...
-    const texturesCount = this.getTextureCountByLayerIndex(layerIndex);
-
-    // remove layer
-    this.colorLayersId.splice(layerIndex, 1);
-    this.uniforms.colorLayersCount.value = this.getColorLayersCount();
-
-    // remove nb textures
-    this.layerTexturesCount.splice(layerIndex, 1);
-    this.layerTexturesCount.push(0);
+        const nbSamplers = [samplersElevationCount, getMaxColorSamplerUnitsCount()];
```
zarov, line 117: I don't really like this part: the higher `getMaxColorSamplerUnitsCount` gets, the higher the number of initialized uniform there is. And if uniforms are initialized, they will be called even if their value is `undefined`. So that means that we will have a high number of draw calls for nothing (here I get from ~1000 DC to ~2800DC for a single frame, with most DC contributing nothing). I get that this is supposed to ease things and how we use them, but can't we make a change here ?

mbredif: (for future reference, here is a summary of the offline discussion)
To reduce the number of `gl.uniform*` calls then you have to tightly adjust   `this.defines.NUM_VS_TEXTURES` (resp. `this.defines.NUM_FS_TEXTURES`)  and resize the `this.uniforms.elevation*` (resp. `this.uniforms.color*`) arrays in `updateUniforms` instead of relying on the uniform value `this.uniforms.elevationTextureCount` (resp. `this.uniforms.colorTextureCount`). That will save you a lot of uniform updates on tiles that only use a small number of textures, maybe at the expense of more shader compilations (as you modify `define` values)

src/Process/LayeredMaterialNodeProcessing.js
zarov: @mbredif I let you answer

mbredif: you can reintroduce it if you like.
```
if (index != coords.length) { console.error(`non-coherent result ${index} vs ${coords.length}.`, coords); }
```

examples/split.html
zarov: No, we are not really setting it to a `layer` but to all `LayeredMaterialLayer` here.

src/Main.js
```diff
@@ -30,3 +30,4 @@ export { default as DEMUtils } from './utils/DEMUtils';
 export { default as CameraUtils } from './utils/CameraUtils';
 export { default as OrientationUtils } from './utils/OrientationUtils';
 export { default as ShaderChunk } from './Renderer/Shader/ShaderChunk';
+export { getMaxColorSamplerUnitsCount } from './Renderer/LayeredMaterial';
```
gchoqueux, line 33: Why export `getMaxColorSamplerUnitsCount`?

src/Core/Scheduler/Scheduler.js
```diff
@@ -160,7 +160,7 @@ Scheduler.prototype.execute = function execute(command) {
 
     // parse host
     const layer = command.layer;
-    const host = layer.url ? new URL(layer.url, document.location).host : undefined;
+    const host = layer.source && layer.source.url ? new URL(layer.source.url, document.location).host : undefined;
```
gchoqueux, line 163: can be move out from this PR?

src/Layer/LayerUpdateStrategy.js
```diff
@@ -38,7 +37,7 @@ function _progressive(nodeLevel, currentLevel, options) {
 // This produces smoother transitions and a single fetch updates multiple
 // tiles thanks to caching.
 function _dichotomy(nodeLevel, currentLevel, options = {}) {
-    if (currentLevel == EMPTY_TEXTURE_ZOOM) {
+    if (currentLevel < 0) {
```
gchoqueux, line 40: why remove this constant?

src/Core/Scheduler/Scheduler.js
zarov: Already in #850

src/Main.js
zarov: I don't know, we can remove it

src/Layer/LayerUpdateStrategy.js
zarov: It can be restored

src/Main.js
gchoqueux: sorry, It's used in test

gchoqueux:
replaced by #912

PR #850, opened 2018-10-03
fix(core): missed layer to layer.source 

zarov:
Original message: Few things I did in #849 that can be done outside #849.

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -160,7 +160,7 @@ Scheduler.prototype.execute = function execute(command) {
 
     // parse host
     const layer = command.layer;
-    const host = layer.url ? new URL(layer.url, document.location).host : undefined;
+    const host = layer.source && layer.source.url ? new URL(layer.source.url, document.location).host : undefined;
 
     command.promise = new Promise((resolve, reject) => {
         command.resolve = resolve;


src/Core/Scheduler/Scheduler.js
```diff
@@ -160,7 +160,7 @@ Scheduler.prototype.execute = function execute(command) {
 
     // parse host
     const layer = command.layer;
-    const host = layer.url ? new URL(layer.url, document.location).host : undefined;
+    const host = layer.source && layer.source.url ? new URL(layer.source.url, document.location).host : undefined;
```
gchoqueux, line 163: it doesn't work for 3Dtiles by examples?

gchoqueux:
@zarov Could you split in two PRs to merge OBB

zarov:
@gchoqueux done, with #871

zarov:
Rebased !

src/Core/Scheduler/Scheduler.js
gchoqueux: What is the behavior with `3dTiles` and `point cloud`?

zarov: I see no problem in the 3dtiles example, is this what you meant ?

gchoqueux: `3dtiles` haven't `Source`, so `host` is `undefined`

zarov: `host` being `undefined` doesn't seem to make any problem, other than it is not in the `hostQueues`. The URL here for `3dtiles` corresponds to the root file of 3dtiles, which is fetch inside the provider in `preprocessDataLayer`. So no need to add it in the `hostQueues` as far as I know. We can change this behavior when rewriting the 3dtiles system.

gchoqueux:
Could you rebase this branch?

zarov:
Done

gchoqueux (approved):

PR #851, opened 2018-10-03
Optimisation in Coordinates and Extent

gchoqueux:
## Description

Avoid new instance or convert in `Coordinates` and `Extent`


zarov (approved):
LGTM, thanks

PR #852, opened 2018-10-03
upgrade(core): upgrade Earcut to 2.1.3

gchoqueux:
## Description
Upgrade Earcut to 2.1.3

Earcut in THREE isn't accessible


zarov (approved):
LGTM, so we are dropping the use of the THREE js version then.

PR #853, opened 2018-10-03
fix(protocol): handling error source

gchoqueux:
## Description
best handling error in data source provider

## motivation
the error's message was poorly handled


zarov (approved):
LGTM thanks

PR #854, opened 2018-10-03
fix(debug): stop to update chart when is hidden

gchoqueux:
## Description
Avoid to update chart when is hidden

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1704,29 +1704,29 @@
       }
     },
     "chart.js": {
-      "version": "2.6.0",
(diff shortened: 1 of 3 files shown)


zarov (commented):
Maybe add a mark to first two charts where the chart ended before being redrawn again ? Like this 

screenshot from 2018-10-03 17-24-10

gchoqueux:
To draw line, it would be necessary to use plugin and there's no simple to use chart plugin with es6

zarov (approved):
Ok no problem, let it as it is then.

PR #861, opened 2018-10-10
refacto(Core): Avoid intermediate projection to convert a Extent

gchoqueux:

### DIFF
--- a/src/Core/Geographic/Extent.js
+++ b/src/Core/Geographic/Extent.js
@@ -72,6 +72,16 @@ Extent.prototype.isTiledCrs = function fnIsTiledCrs() {
     return this._crs.indexOf('WMTS:') == 0 || this._crs == 'TMS';
 };
 
+// EPSG:3857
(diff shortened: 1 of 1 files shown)


src/Core/Geographic/Extent.js
```diff
@@ -72,27 +72,44 @@ Extent.prototype.isTiledCrs = function fnIsTiledCrs() {
     return this._crs.indexOf('WMTS:') == 0 || this._crs == 'TMS';
 };
 
+const s = 20037508.34 * 2;
```
zarov: What is this ? Please add a comment so we don't forget

zarov (commented):
Thanks for noticing this ! Apart the comment to add, nothing to add

src/Core/Geographic/Extent.js
gchoqueux: added

zarov (approved):
LGTM, thanks for the comment

PR #862, opened 2018-10-10
Reduce examples test time

zarov:
The current way of testing the example is relatively slow: for each test (`it`), we are creating and loading a new page. For tests on the same example, it doesn't make sense to reload it each time.

So what I'm proposing here is keeping the same page for a same test file. Some states are restored after each test, like the position on the view and the position of the mouse.

On my computer, I'm going from 9'14" to 4'31" in test time, so it is a clear benefit, and we won't have to wait a lot of time when doing a pull request and waiting for CI.

The first commit of this PR also changes all `globeView` variables in examples to `view`.

### DIFF
--- a/examples/collada.html
+++ b/examples/collada.html
@@ -30,14 +30,14 @@
             var viewerDiv = document.getElementById('viewerDiv');
 
             // Instanciate iTowns GlobeView*
-            var globeView = new itowns.GlobeView(viewerDiv, positionOnGlobe);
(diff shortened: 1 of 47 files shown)


zarov:
I removed the screenshots in the process, but I'm working on adding them back, please wait before merging.

zarov:
I'm good, screenshots are back ! Also notice that we gain 10min on a Travis CI, and 3-4min in AppVeyor.

gchoqueux (approved):
\o/ very nice

PR #865, opened 2018-10-10
Feat camera calibration

gliegard:
## Description
- OrientedImageCamera : Extension of a ThreeJs Camera adapted to computer vision.
- CameraCalibrationParser : Create a set of  OrientedImageCamera from a configuration file
- an example show a car with 5 cameras

## Motivation and Context
Long-term goal : Immersive view

## Screenshots
camera_parser


zarov:
I feel this is the same example as the current `orientation_utils` examples, why not update those instead ?

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,80 @@
...
+import OrientedImageCamera from '../Renderer/OrientedImageCamera';
+
+/**
+ * The CameraCalibrationParser module provide a [parse]{@link module:CameraCalibrationParser.parse}
+ * method that takes a JSON array of camera calibrations in and yields an array of THREE.Camera
+ *
+ * @module CameraCalibrationParser
+ */
+
+THREE.Matrix4.prototype.setMatrix3 = function setMatrix3(m) {
```
zarov, line 11: What about renaming it to `setFromMatrix3` ? Did you also try to open discussion to add this method in THREE.js ? I didn't find anything like that on their repository, so I don't know if it's relevant to suggest it.

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,80 @@
...
+    this.elements[4] = m.elements[3];
+    this.elements[5] = m.elements[4];
+    this.elements[6] = m.elements[5];
+    this.elements[8] = m.elements[6];
+    this.elements[9] = m.elements[7];
+    this.elements[10] = m.elements[8];
+    return this;
+};
+
+// the json format encodes the following transformation:
```
zarov, line 25: This description should be below in the `parse` documentation, so we can read it in the generated documentation, or above in the `module` section.

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,80 @@
...
+    const focal = new THREE.Vector2(proj[0], proj[4]);
+    const center = new THREE.Vector2(proj[2], proj[5]);
+    const skew = proj[1];
+    var camera = new OrientedImageCamera(size, focal, center, options.near, options.far, skew);
+
+    // parse extrinsics: Object3d.matrix is from local to world
+    // p_world = position + transpose(rotation) * p_local
+    camera.position.fromArray(calibration.position);
+    // calibration.rotation is row-major but fromArray expects a column-major array, yielding the transposed matrix
+    var rotationInverse = new THREE.Matrix3().fromArray(calibration.rotation);
```
zarov: Why not alter the `calibration.rotation` array, to be a 16-length array, and then create directly a `Matrix4` array ? Or at least you should declare the `rotationInverse` Matrix3 outside this function.

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,80 @@
...
+// intrinsics: p_pixel = projection * p_local
+// distortion: p_raw = distortion(p_pixel)
+function parseCalibration(calibration, options) {
+    // parse intrinsics
+    const proj = calibration.projection;
+    const size = new THREE.Vector2().fromArray(calibration.size);
+    const focal = new THREE.Vector2(proj[0], proj[4]);
+    const center = new THREE.Vector2(proj[2], proj[5]);
+    const skew = proj[1];
+    var camera = new OrientedImageCamera(size, focal, center, options.near, options.far, skew);
```
zarov: Don't use `var`.

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,80 @@
...
+    var rotationInverse = new THREE.Matrix3().fromArray(calibration.rotation);
+    camera.matrix.setMatrix3(rotationInverse);
+    // local axes for cameras is (X right, Y up, Z back) rather than (X right, Y down, Z front)
+    camera.scale.set(1, -1, -1);
+    camera.quaternion.setFromRotationMatrix(camera.matrix);
+    camera.matrix.compose(camera.position, camera.quaternion, camera.scale);
+
+    // parse distortion
+    if (calibration.distortion) {
+        camera.distortion = {
```
zarov, line 51: What is the purpose of the distortion property ? I find no other mention of it. Also it should be at least initialized in `OrientedImageCamera`.

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,80 @@
...
+            pps: new THREE.Vector2().fromArray(calibration.distortion.pps),
```
zarov, line 52: I don't know the usage you will have of this, but is it necessary to be Vectors ?

src/Renderer/OrientedImageCamera.js
```diff
@@ -0,0 +1,60 @@
+import * as THREE from 'three';
+
+class OrientedImageCamera extends THREE.PerspectiveCamera {
```
zarov, line 7: We need more information on the purpose of this new camera, please add some documentation.

src/Renderer/OrientedImageCamera.js
```diff
@@ -0,0 +1,60 @@
...
+    constructor(size = 1024, focal = 1024, center, near = 0.1, far = 1000, skew, aspect) {
+        size = size.isVector2 ? size : new THREE.Vector2(size);
+        aspect = aspect || size.x / size.y;
+        super(undefined, aspect, near, far);
+        this.size = size;
+        this.focal = focal.isVector2 ? focal : new THREE.Vector2(focal, focal);
+        this.center = center || size.clone().multiplyScalar(0.5);
+        this.skew = skew || 0;
+        Object.defineProperty(this, 'fov', {
+            get: () => Math.atan2(this.size.y, 2 * this.focal.y) * 360 / Math.PI,
```
zarov, line 27: This value should be computed somewhere else, not every time we ask for the fov.

zarov (changes requested):
A few changes to make. Also you should add a unit test for the parser with a few cameras, and the `setMatrix3` method also.

src/Parser/CameraCalibrationParser.js
mbredif: `setFromMatrix3` would imply that the whole Matrix4 is set, whereas this only updates the upper left Matrix3.
`setMatrix3` is somewhat more in line with `setPosition` which only updates the upper right Vector3.
I definitely support an upstream PR.

src/Parser/CameraCalibrationParser.js
mbredif: A 4x4 rotation would have 7 constant entries (0s on the last lines and columns except a 1 on the lower right) which seems a bit redundant...
Ok for the memory optimization.

src/Parser/CameraCalibrationParser.js
mbredif: I think so : it will be passed as a `vec2` uniform.

src/Renderer/OrientedImageCamera.js
mbredif: This class makes `fov` a computed property depending on size.y and focal. An alternative would be to use setters for size and focal that update a cached fov value.

src/Renderer/OrientedImageCamera.js
mbredif: I continued working on this camera after handing it to @gliegard and my version has a better documentation that you can adapt/reuse : https://github.com/mbredif/three-photogrammetric-camera/blob/master/src/cameras/PhotogrammetricCamera.js

mbredif (commented):
Looks good !
+1 for extending the existing example instead of creating a new one.

src/Parser/CameraCalibrationParser.js
zarov: Ok for `setMatrix3`

src/Parser/CameraCalibrationParser.js
zarov: Yeah I was kind of conflicted on this one, just move the `new` outside the method.

src/Parser/CameraCalibrationParser.js
zarov: Oh yeah makes sense, thanks

src/Parser/CameraCalibrationParser.js
gliegard: It will be used in #866 https://github.com/iTowns/itowns/blob/6293c10d33328a3068ee23fb6dab35cee4aebf5b/src/Renderer/OrientedImageMaterial.js#L61 
It can be defined and used, or it can be undefined and not used.

src/Renderer/OrientedImageCamera.js
gliegard: ok, I've add documentation.
Thanks Mathieu.

src/Renderer/OrientedImageCamera.js
zarov: Sounds good to leave as it is then, no problem

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,82 @@
...
+    return camera;
+}
+
+export default {
+    /**
+     * @function parse
+     * @param {string|JSON} json - the json content of the calibration file.
+     * @param {Object} options - Options controlling the parsing.
+     * @param {string} options.near - Near of the created cameras.
+     * @param {string} options.near - Far of the created cameras.
```
zarov: `options.far`

src/Parser/CameraCalibrationParser.js
```diff
@@ -0,0 +1,82 @@
...
+            camera.distortion.l1l2.z = calibration.distortion.etats;
+        }
+    }
+
+    camera.name = calibration.id;
+    return camera;
+}
+
+export default {
+    /**
```
zarov, line 68: Could you align the documentation for something similar to "clean" parser ? Like `GeojsonParser` or `VectorTileParser` for example.

src/Renderer/OrientedImageCamera.js
```diff
@@ -0,0 +1,74 @@
+import * as THREE from 'three';
+
+/**
+ * @classdesc OrientedImageCamera is a ThreeJs camera adapted to photogrammetric description.
+ * So we can build a ThreeJs perspective camera from size and focal information.
+ */
+class OrientedImageCamera extends THREE.PerspectiveCamera {
+    /**
+     * @Constructor
```
zarov: `@constructor`

src/Renderer/OrientedImageCamera.js
```diff
@@ -0,0 +1,74 @@
...
+            set: (fov) => {
+                var focal = 0.5 * this.size.y / Math.tan(fov * Math.PI / 360);
+                this.focal.x = focal;
+                this.focal.y = focal;
+            },
+        });
+        this.updateProjectionMatrix();
+    }
+
+    updateProjectionMatrix() {
```
zarov, line 41: Please comment why you are overriding this method.

zarov (commented):
A few things to change, only documentation. Thanks for adding unit tests !

zarov (approved):
Thanks for these changes !

PR #866, opened 2018-10-10
material to make projective texture from cameras

gliegard:
## Description
feat(OrientedImageMaterial): Material used to compute projective texture
The material is created with a set of cameras.
It supports distortion.

## Motivation and Context
Long-term goal : Immersive view

PR #867, opened 2018-10-11
refactor(shader): manage chunk directly into shaders

zarov:
Before, shader chunks were added manually in JS code for each material.
Now they are directly added in shaders. To use it, `ShaderChunk` has been
added with the method `install`, that is call in `LayeredMaterial`.

In the same time, the `colorAtIdUv` unrolling has been moved to the
shader, and unrolled using the new `ShaderUtils.unrollLoop` method and the
pragma `unroll_loop`.

Note that this is adapted from #849 and that some of this code can be used in #866 if we merged this PR before #866.


src/Renderer/LayeredMaterial.js
```diff
@@ -77,10 +62,8 @@ const LayeredMaterial = function LayeredMaterial(options) {
     THREE.RawShaderMaterial.call(this);
 
     const nbSamplers = getMaxColorSamplerUnitsCount();
-    this.vertexShader = TileVS;
 
-    this.fragmentShaderHeader = `${PrecisionQualifier}\nconst int   TEX_UNITS   = ${nbSamplers.toString()};\n`;
-    this.fragmentShaderHeader += pitUV;
+    this.fragmentShaderHeader = `#define TEX_UNITS ${nbSamplers.toString()}\n`;
```
mbredif: Could you use `this.defines.TEX_UNITS = nbSamplers;` instead of creating 
`this.fragmentShaderHeader` ?

src/Renderer/LayeredMaterial.js
```diff
@@ -99,11 +82,8 @@ const LayeredMaterial = function LayeredMaterial(options) {
         vsOptions = '\n#define DATA_TEXTURE_ELEVATION\n';
     }
 
-    // see GLOBE FS
-    this.fragmentShaderHeader += getColorAtIdUv(nbSamplers);
-
-    this.fragmentShader = this.fragmentShaderHeader + TileFS;
-    this.vertexShader = PrecisionQualifier + vsOptions + TileVS;
+    this.fragmentShader = ShaderUtils.unrollLoops(this.fragmentShaderHeader + TileFS, { TEX_UNITS: nbSamplers });
```
mbredif: Using `this.defines`, this simplifies to `this.fragmentShader = ShaderUtils.unrollLoops(TileFS, this.defines);`

src/Renderer/LayeredMaterial.js
```diff
@@ -77,10 +62,8 @@ const LayeredMaterial = function LayeredMaterial(options) {
...
 
     const nbSamplers = getMaxColorSamplerUnitsCount();
-    this.vertexShader = TileVS;
 
-    this.fragmentShaderHeader = `${PrecisionQualifier}\nconst int   TEX_UNITS   = ${nbSamplers.toString()};\n`;
-    this.fragmentShaderHeader += pitUV;
+    this.fragmentShaderHeader = `#define TEX_UNITS ${nbSamplers.toString()}\n`;
 
     if (__DEBUG__) {
         this.fragmentShaderHeader += '#define DEBUG\n';
```
mbredif: `this.defines` again

src/Renderer/LayeredMaterial.js
```diff
@@ -99,11 +82,8 @@ const LayeredMaterial = function LayeredMaterial(options) {
         vsOptions = '\n#define DATA_TEXTURE_ELEVATION\n';
```
mbredif: `this.defines` again

mbredif (commented):
Thanks a lot for taking the time to marshall my oversized PR into manageable bits !!!
LGTM, I am just suggesting some simple simplifcations for the FS defines. For the VS defines, you could do the same thing with `_minElevation` and  `_maxElevation` as defined quantities to get rid of `vsOptions` as well.

src/Renderer/LayeredMaterial.js
zarov: I couldn't make it work :sweat_smile: if I write as you said, `TEX_UNITS` doesn't seem to be appended to the shader

zarov:
No problem, thanks for doing all that in the first place ! Yes I agree, if I can make it work I'll definitely replace all `#define`.

mbredif:
Is it because of the overwriting here ? https://github.com/iTowns/itowns/blob/f19f8f82112001bdc8066538f2282466832e51bb/src/Renderer/LayeredMaterial.js#L152-L159

zarov:
Ah yes that must be it, I didn't thought of that ! Rewriting all of this now

zarov:
@mbredif updated !

src/Renderer/Shader/ShaderChunk.js
```diff
@@ -0,0 +1,17 @@
+import pitUV from './Chunk/pitUV.glsl';
```
gchoqueux, line 3: Could you load directly load  itowns shader chunk in main?
So, you could rename this file in `itownsShaderChunk.js`

mbredif: I guess we can keep the name `ShaderChunk.js` and have client code use `itowns.ShaderChunk`.

ok to expose `ShaderChunk.js` in main, but is there a way to tag this and other lowlevel API as such, to segment the THREE aware and unaware APIs ?

src/Renderer/LayeredMaterial.js
```diff
@@ -8,17 +8,18 @@
...
 import TileVS from './Shader/TileVS.glsl';
 import TileFS from './Shader/TileFS.glsl';
-import pitUV from './Shader/Chunk/pitUV.glsl';
-import PrecisionQualifier from './Shader/Chunk/PrecisionQualifier.glsl';
+import ShaderUtils from './Shader/ShaderUtils';
+import ShaderChunk from './Shader/ShaderChunk';
 import Capabilities from '../Core/System/Capabilities';
 import { l_COLOR, l_ELEVATION, EMPTY_TEXTURE_ZOOM } from './LayeredMaterialConstants';
 
+ShaderChunk.install(THREE.ShaderChunk, 'itowns.');
```
mbredif: this and all the includes could be change to `'itowns/'` once we THREE r95+ is supported

mbredif:
LGTM

src/Renderer/Shader/ShaderChunk.js
gchoqueux: it isn't to expose `ShaderChunk.js` but to load directly all `itowns Chunk`.
With `itownsShaderChunk.js`, it's clearer to differentiate the chunks from itowns and those from THREE

mbredif: > load directly all itowns Chunk.

I do not get what your are proposing, sorry

zarov:
So I updated in what I think @gchoqueux had in mind, let me know if it is okay for both of you !

gchoqueux (approved):

PR #871, opened 2018-10-12
refactor(core): remove OBB() method of TileMesh 

zarov:
Remove OBB() method of TileMesh in favor of direct access to obb property


gchoqueux (approved):
thanks

PR #872, opened 2018-10-12
Fix lookAt in oriented image

gchoqueux:
In example, the object look at local position.
We should transform all in the world coordinate.
Or found an easier solution


gchoqueux:
I pushed a new fix

zarov (approved):
Thanks for this !

PR #878, opened 2018-10-17
fix(shaderUtils): use boundary delimiter in the regex

gliegard:
## Description
This PR change the regex of the unroll loop method.
It uses the boundary delimiter, that allow a more flexible usage of this method in the shader.


zarov (approved):
LGTM, thanks

PR #881, opened 2018-10-19
Feat first person controls globe

gliegard:
## Description
FirstPersonControls adapted to Globe view

### DIFF
--- a/src/Renderer/ThreeExtended/FirstPersonControls.js
+++ b/src/Renderer/ThreeExtended/FirstPersonControls.js
@@ -23,13 +23,25 @@ function applyRotation(view, camera3D, state) {
 }
 
 const MOVEMENTS = {
-    38: { method: 'translateZ', sign: -1 }, // FORWARD: up key
-    40: { method: 'translateZ', sign: 1 }, // BACKWARD: down key
+    38: { method: 'translateY', sign: -1 }, // FORWARD: up key
+    40: { method: 'translateY', sign: 1 }, // BACKWARD: down key
     37: { method: 'translateX', sign: -1 }, // STRAFE_LEFT: left key
     39: { method: 'translateX', sign: 1 }, // STRAFE_RIGHT: right key
-    33: { method: 'translateY', sign: 1 }, // UP: PageUp key
-    34: { method: 'translateY', sign: -1 }, // DOWN: PageDown key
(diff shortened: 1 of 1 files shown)


zarov:
I'm okay with this change, even if you can, as you told me, add a simpler check in `translateY`, like `view instanceof GlobeView`. I'll let someone else take a look at that too.

gliegard:
@zarov we don't want to test if view is the GlobeView all the time, whereas the view doesn't change. That's why I've created another object !!

gliegard:
@zarov I don't want to test if view is the GlobeView all every time I press the key, whereas the view is not supposed to change.

gchoqueux:
Could you do this in `FirstPersonControls` to make it transparent to the user?

gliegard:
ok, I did the check in the constructor

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -161,6 +165,16 @@ class FirstPersonControls extends THREE.EventDispatcher {
         }
     }
 
+    translateY(dt, sign) {
+        this.camera.position.z += sign * this.options.moveSpeed * dt / 1000;
```
gchoqueux: Strange name `translateY` to add on `z`.
But it is coherent to keep this name?

src/Renderer/ThreeExtended/FirstPersonControls.js
```diff
@@ -161,6 +165,16 @@ class FirstPersonControls extends THREE.EventDispatcher {
...
+    }
+
+    translateYGlobe(dt, sign) {
+        // compute geodesic normale
+        const normal = this.camera.position.clone().normalize();
```
gchoqueux: You could declare normal, outside the function and use `copy`, instead of `clone`

gchoqueux (changes requested):

src/Renderer/ThreeExtended/FirstPersonControls.js
gliegard: yes, you're right, so I changed method name to 'translateZ'

src/Renderer/ThreeExtended/FirstPersonControls.js
gliegard: yes, good idea !

gchoqueux (approved):

PR #882, opened 2018-10-19
chore: update the Geoportal API key

zarov:
The current key is about to expire soon

### DIFF
--- a/examples/globe_wfs_extruded.html
+++ b/examples/globe_wfs_extruded.html
@@ -169,7 +169,7 @@
                 },
                 filter: acceptFeature,
                 source: {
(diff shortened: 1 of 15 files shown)


examples/globe_wfs_extruded.html
```diff
@@ -186,67 +186,6 @@
                 }
             });
 
-            function configPointMaterial(result) {
```
gchoqueux, line 189: Why you remove the point layer ?

gchoqueux, line 189: Why you remove the point layer ?

zarov: It is no longer available in the current API key, and it doesn't really add something to this example.

gchoqueux (approved):

PR #884, opened 2018-10-23
OBB to class ES6 + Doc

gchoqueux:
And fix little typo

### DIFF
--- a/src/Renderer/ThreeExtended/OBB.js
+++ b/src/Renderer/ThreeExtended/OBB.js
@@ -1,133 +1,168 @@
 import * as THREE from 'three';
-import Coordinates from '../../Core/Geographic/Coordinates';
 import TileGeometry from '../../Core/TileGeometry';
 import BuilderEllipsoidTile from '../../Core/Prefab/Globe/BuilderEllipsoidTile';
 
-function OBB(min, max) {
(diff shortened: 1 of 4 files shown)


gchoqueux:
updated

src/Renderer/ThreeExtended/OBB.js
```diff
@@ -1,133 +1,168 @@
...
+        // this is the same as isPointInsideSphere.position
+        const distance = Math.sqrt((x - localSpherePosition.x) * (x - localSpherePosition.x) +
+                               (y - localSpherePosition.y) * (y - localSpherePosition.y));
+
+        return distance < sphere.radius;
+    }
+
+    /**
+     * Compute OBB from extent.
+     * The OBB resulted can be only in the system 'EPSG:3946'.
```
zarov, line 130: Do you mean `4326` ?

zarov (approved):
LGTM, thanks for reformatting it in `class` and adding some documentation.

src/Renderer/ThreeExtended/OBB.js
gchoqueux: No, the extent in entry is in `4326` but the result is `3946` because it is meant to viewer

PR #885, opened 2018-10-24
Npm fix and upgrade

### LINKED ISSUE #868: Check-node-version doesn't run for Windows
Working environment Windows 7
Node.js version 10.11.0
During npm install you get the following:
"prepare": "check-node-version --npm '>= 6' || echo \"We detected that you are using npm < 6, please use npm >= 6 if you want to make modification to package-lock.json, feel free to ignore the above error if not.\n\""
 The installation process is stopped.
 
Do a npm install in Windows environment. 

Indication: package added in #816

gchoqueux:
## Description
Fix fail test

Fix wrong dependencies (Earcut)

Upgrade to `babel` 7 and `webpack` 4

Remove check node version fix #868


src/Core/Math/Ellipsoid.js
```diff
@@ -128,7 +128,7 @@ Ellipsoid.prototype.intersection = function intersection(ray) {
 
     var d = ((b * b) - (4 * a * c));
     if (d < 0 || a === 0 || b === 0 || c === 0)
-        { return false; }
+    { return false; }
```
zarov: Could you maybe change things like that to 
```js
if (d < 0 || a === 0 || b === 0 || c === 0) { return false; }
```
or
```js
if (d < 0 || a === 0 || b === 0 || c === 0) {
    return false;
}
```

.eslintrc.js
```diff
@@ -77,7 +77,20 @@ module.exports = {
     'no-shadow': 'off',
     'no-restricted-properties': 'off',
     'prefer-spread': 'off',
+    'prefer-destructuring': 'off',
```
zarov, line 80: Why removing all these rules ?

zarov (commented):
Thanks for updating all the packages ! You missed to modify examples and tests however.

.eslintrc.js
gchoqueux: it's new eslint rules I have disabled.

src/Core/Math/Ellipsoid.js
gchoqueux: fixed

package.json
```diff
@@ -34,51 +33,47 @@
...
-    "text-encoding-utf-8": "^1.0.1",
-    "togeojson": "^0.16.0",
-    "url-polyfill": "^1.0.8",
-    "whatwg-fetch": "^2.0.2"
-  },
-  "peerDependencies": {
-    "proj4": "^2.4.4",
+    "proj4": "^2.5.0",
+    "text-encoding-utf-8": "^1.0.2",
     "three": "^0.97.0",
```
tbroyer, line 49: Why did it (along with three.meshline and proj4) move from a peer dependency to a "normal" dependency?

Fwiw, url-polyfill and whatwg-fetch should probably be dev dependencies (they're polyfills, and only used by the webpack build)

gchoqueux: >Why did it (along with three.meshline and proj4) move from a peer dependency to a "normal" dependency?

It seems that the peer dependencies are used in the development of plugin package. These dependencies (`Three, proj4`...) are necessary for webpack building when using `itowns` package in other project.
By example, `Itowns` isn't a `Three package plugin,` it's dependent on `Three` so it must be installed automatically.  Otherwise why not move all the itowns dependencies in the peerDependencies.

>Fwiw, url-polyfill and whatwg-fetch should probably be dev dependencies (they're polyfills, and only used by the webpack build)

yes indeed there will be no reason to leave in dependencies

tbroyer: > It seems that the peer dependencies are used in the development of plugin package. These dependencies (`Three, proj4`...) are necessary for webpack building when using `itowns` package in other project.

See https://github.com/iTowns/itowns/commit/400d3d060940b4b63049678e2523c49df0a1b0eb (and the 2.2.0 release notes)
The problem is that as soon as you use a different version of Three (or proj4), NPM will no longer _hoist_ it, you'll use the two separate versions of Three at the same time, and of course any `instanceof` check will fail.

gchoqueux: Ok, but it's the same with the packages `tween, earcut, @mapbox/vector-tile, js-priority-queue, pbf`...?

tbroyer: AFAICT, js-priority-queue (for example) is entirely internal to iTowns, so it does not really matter whether it appears twice. Three is exposed through the API of iTowns, so it's another matter (and iTowns uses `instanceof` on Three objects, so it's important that the user of iTowns and iTowns itself use the exact same version)

gchoqueux: ok, I see, thanks for the explanation

gchoqueux: I restored the peerDependencies

gchoqueux: I think we should not expose `THREE` anymore. Because he weighed down the package and it's not too much in the logic of the packages. WDYT?

zarov: I'm even for removing three and proj4 from the exported package of itowns: it is up to the user to include three in its webpage. That may cause more problem on our side, being careful when upgrading a version of three, because a user could want a less/more recent version of three. I'm not pushing anything of course, as this opinion is subject to some lengthy discussions.

gchoqueux: It should be noted that `PeerDependencies` don't prohibit the use of different versions of THREE. They don't raise the problem of conflict with `instanceof`.

gchoqueux:
@zarov 
Tests are fixed

gchoqueux:
@zarov 
PR updated

zarov (approved):
Thanks for the fixes, I hope we'll be good with everything now !

PR #887, opened 2018-10-26
fix(core): remove useless console.info in animation

gchoqueux:


zarov (commented):
`setPlayerState` was only here for the debug message, if you remove it you might as well remove this function and set directly the state instead of calling this function.

gchoqueux:
@zarov `setPlayerState` removed

zarov (approved):
Thanks LGTM

PR #893, opened 2018-11-07
fix(core): problem in runtime with bundle itowns and debug

gchoqueux:

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -43,26 +43,8 @@ module.exports = {
         umdNamedDefine: true,
     },
     optimization: {
-        splitChunks: {
-            chunks: 'async',
-            minSize: 30000,
-            maxSize: 0,
-            minChunks: 1,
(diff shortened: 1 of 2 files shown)


gchoqueux:
I remove useless change in `index.html`

zarov (approved):
Seems to solve the bbox problem

webpack.config.js
```diff
@@ -1,7 +1,12 @@
 var fs = require('fs');
 var path = require('path');
 
+var webpack = require('webpack');
+
 const debugBuild = process.env.NODE_ENV === 'development';
+var definePlugin = new webpack.DefinePlugin({
+    __DEBUG__: debugBuild,
```
tbroyer: Isn't this already done in the minify-replace Babel plugin below? Should it replace Babel's minify-replace?

gchoqueux: yes, thanks, it's a duplicate, I remove it

PR #897, opened 2018-11-07
Add info object in layer TiledGeometryLayer

gchoqueux:
## Description
Add property **`info`** object in `TiledGeometryLayer`.
`Info` is `InfoTiledGeometryLayer` that provides some states layer informations.
These informations are:
- displayed tiles;
- displayed `colorLayer` and `elevationLayer`;
- `extent` of displayed tiles;

## Motivation and Context
This prevents the user from re-browsing the tiles while `mainLoop` is already doing it.

- Displayed `colorLayer` and `elevationLayer` information can used to show **Attributions**
- `Extent` of displayed tiles information can used to compute a automatic near and far

### Extent of displayed tiles 
global obb

### Displayed colorLayer
info vibility layer

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -29,6 +29,7 @@
             "src/Layer/ElevationLayer.js",
             "src/Layer/GeometryLayer.js",
             "src/Layer/TiledGeometryLayer.js",
+            "src/Layer/InfoLayer.js",
 
             "src/Source/Source.js",
             "src/Source/WMTSSource.js",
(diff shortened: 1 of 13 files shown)


gchoqueux:
@zarov new file `InfoLayer.js`

zarov (approved):
LGTM

Note that I'm not a fan of this, as it looks like it is kind of a debug thing; but I think we may be able to remove it once we finish our API refactoring and some other bits.

PR #902, opened 2018-11-08
chore: upgrade THREE to v98

gchoqueux:
## r97 → r98
- [x] Renamed `ObjectLoader.setTexturePath()` to `ObjectLoader.setResourcePath()`. Added `ObjectLoader.setPath()`.
- [x] `CanvasRenderer` has been removed.
- [x] The order of `LoadingManager`'s callbacks has changed. `onError()` is now called before `onLoad()`.
- [x]  LegacyGLTFLoader.js
- [x]  Reorganize files and folders
- [x] Remove relative path to import module in `src`
- [x] Remove relative path to import module in `test`

### Automatic patch three examples to import directly in `itowns`.

`three/examples` files are patched to support ES6 `import`.
**Warning only files listed,** in `threeExamples.js`, are patched.

By example, three/examples can be imported in itowns by : 
```js
import GLTFLoader from 'ThreeExtended/loaders/GLTFLoader';
```

**Resolve import ES6**

This PR add `./src` path to resolve importing. You can write:
```js
// relative path
import xyz from '../../src/Utils/xyz';
// or 
import xyz from 'Utils/xyz';
```


.babelrc
```diff
@@ -1,6 +1,11 @@
 {
-    "presets": ["@babel/env"],
+    "presets": ["@babel/preset-env"],
     "plugins": [
+        ["module-resolver",  { "root": ["./"],
+            "alias": {
+                "threeExamples": "./src/Renderer/ThreeExtended/jsPatched/"
```
zarov: Instead of creating yet another folder, couldn't we take the opportunity to move all files that _really_ extend THREE.js to a proper folder, under `src/` ? Something like `src/ThreeExtended`, not under `Renderer`. The current folder looks like something where we put things we don't know where to put.

mbredif: An even simpler proposal would be to put the file in itowns at the position they would have been in the THREE project.

gchoqueux: I reorganized the files and folders.

It's difficult to put the patched files in different folders because it would be necessary to specify, at **Git**, all the files to be ignored. And these can't be in `node_modules/three` because `babel-register` doesn't work.

zarov:
Good, we have almost get rid of this `ThreeExtended` folder ! A few remarks though: 
- move `FeaturesUtils` in `utils`
- what about the `Parser` folder ? It contains a few converters (textureConvert, Feature2Mesh...) that could be in another folder, because their role is different than a simple parser, parsing data to our custom `Features` format. WDYT ?
- I don't know about the last commit: this seems a clean way to do import, but is it necessary ?

gchoqueux:
PR updated

config/replace.config.js
```diff
@@ -0,0 +1,114 @@
...
+
+patchs.push(() => replace({
+    files: fileGLTFLoader,
+    from: patchBatchID,
+    to: patchGLTFLoader,
+}));
+
+// Patch _BATCHID in LegacyGLTFLoader
+const fileLegacyGLTFLoader = files.filter(f => path.basename(f, '.js') === 'LegacyGLTFLoader');
+const fromBatchIDLegacy = /case +'JOINT':[\n\t ]+geometry\.addAttribute\( 'skinIndex', bufferAttribute \);[\n\t ]+break;/g;
```
zarov: You could use `\s` instead of `[\n\t\ ]`.

config/replace.config.js
```diff
@@ -0,0 +1,114 @@
...
+    console.log(`\t${t.map(f => path.basename(f)).join('\n\t')}`);
+});
+
+const patchs = [];
+// New header file:
+// Message Header
+// import * as THREE from 'three';
+// Declare New SpaceName
+
+const modifiedFile = '// This file has modified after NPM install ITOWNS, by NPM script Prepare\n';
```
zarov: Please rephrase this, it is not understandable.

zarov (commented):
Almost good for me ! @mbredif what do you think of this system ? It seems to answer our needs imho.

config/replace.config.js
gchoqueux: ok

config/replace.config.js
gchoqueux: rephrased...

mbredif:
>  @mbredif what do you think of this system ?

I have not got into the details of this PR but the general file layout seems better (shallower, better organized) and the semi automated way of using three extras seems to go in the right direction.

Is there a more specific point where you want my feedback ?

zarov:
Ok thanks for the input, if you are ok with the file layout then it's a go !

zarov (approved):

PR #904, opened 2018-11-09
feat(view): add support for a coordinate when creating a GlobeView

zarov:
Passing a `Coordinates` when creating a `GlobeView` is now supported.

### DIFF
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -71,16 +71,26 @@ export function createGlobeLayer(id, options = {}) {
 }
 
 /**
- * Creates the viewer Globe (the globe of iTowns).
- * The first parameter is the coordinates on wich the globe will be centered at the initialization.
- * The second one is the HTML div in wich the scene will be created.
+ * Creates a view of a globe.
+ *
  * @constructor
- * @example view = new GlobeView(viewer, positionOnGlobe);
- * // positionOnGlobe in latitude, longitude and altitude
- * @augments View
- * @param {HTMLDivElement} viewerDiv - Where to instanciate the Three.js scene in the DOM
- * @param {object} coordCarto
- * @param {object=} options - see {@link View}
+ *
+ * @example
+ * var viewerDiv = document.getElementById('viewerDiv');
+ * var position = new itowns.Coordinates('WGS84', 2.35, 48.8, 25e6);
+ * var view = new itowns.GlobeView(viewerDiv, position);
+ *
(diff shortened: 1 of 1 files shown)


src/Core/Prefab/GlobeView.js
```diff
@@ -71,27 +71,42 @@ export function createGlobeLayer(id, options = {}) {
...
+ * var view = new itowns.GlobeView(viewerDiv, position);
+ *
+ * @example
+ * var viewerDiv = document.getElementById('viewerDiv');
+ * var position = { longitude: 2.35, latitude: 48.8, altitude: 25e6 };
+ * var view = new itowns.GlobeView(viewerDiv, position);
+ *
+ * @param {HTMLDivElement} viewerDiv - Where to attach the view and display it
+ * in the DOM.
+ * @param {object|Coordinates} coordCarto - An object containing three
```
gchoqueux, line 90: This could also be [`cameraTransformOptions`](https://github.com/iTowns/itowns/blob/master/src/utils/CameraUtils.js#L322). 
it would also be necessary to remove the ambiguity between the camera position and the center of the view, because they can be different.

zarov: Updated, WDYT

src/Core/Prefab/GlobeView.js
```diff
@@ -101,6 +102,8 @@ function GlobeView(viewerDiv, coordCarto, options = {}) {
     let positionCamera;
     if (coordCarto instanceof Coordinates) {
         positionCamera = coordCarto.as('EPSG:4326');
+    } else if (coordCarto.coord instanceof Coordinates) {
+        positionCamera = coordCarto.coord.as('EPSG:4326');
```
gchoqueux: I meant to be able to use the all properties `cameraTransformOptions`  (tilt, heading...). This request may be a modification in  `GlobeControls`

zarov: This needs more changes than the initial intent of this PR, as there isn't always a `GlobeControls` (if `options.noControls === true`). What about doing it later ?

gchoqueux: Yes, you can remove this change

if `options.noControls === true` you can use `CameraUtils` instead of controls

zarov: Changed ! Ok I see, we may want to change the current behavior for simplification with `CameraUtils` indeed.

gchoqueux (approved):

PR #905, opened 2018-11-09
Add babel-inline-import-loader in webpack

gchoqueux:
## Motivation
A webpack loader enabling files imported by babel-plugin-inline-import to trigger rebuilds when content changes.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -29,10 +29,17 @@ babelConf.babelrc = false; // disabel babelrc reading, as we've just done it
 const replacementPluginConf = babelConf.plugins.find(plugin => Array.isArray(plugin) && plugin[0] === 'minify-replace');
 replacementPluginConf[1].replacements.find(decl => decl.identifierName === '__DEBUG__').replacement.value = debugBuild;
 
+const include = [
+    path.resolve(__dirname, 'src'),
+    path.resolve(__dirname, 'test'),
+    path.resolve(__dirname, 'utils'),
+];
+
 module.exports = {
+    context: path.resolve(__dirname),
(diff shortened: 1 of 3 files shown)


tbroyer (commented):
6 commits, 1 published version, last published 3 years ago :thinking: Not the kind of dependency I'd add in my projects…

BTW, does it play well with `babel-plugin-inline-import` that we already have in `.babelrc`? That one points to https://github.com/elliottsj/babel-inline-import-loader when using webpack.

gchoqueux:
I had seen that it was an old fossil, I hesitated ;)
Thank you, your proposal seems more logical.

gchoqueux:
@tbroyer
I replaced by `babel-inline-import-loader`

webpack.config.js
```diff
@@ -29,10 +29,17 @@ babelConf.babelrc = false; // disabel babelrc reading, as we've just done it
...
 replacementPluginConf[1].replacements.find(decl => decl.identifierName === '__DEBUG__').replacement.value = debugBuild;
 
+const include = [
+    path.resolve(__dirname, 'src'),
+    path.resolve(__dirname, 'test'),
+    path.resolve(__dirname, 'utils'),
+];
+
 module.exports = {
+    context: path.resolve(__dirname),
```
zarov, line 39: Weird from webpack that we can't use this context everywhere :thinking:

zarov (approved):
This is welcome !

webpack.config.js
gchoqueux: some options need absolute path...

PR #908, opened 2018-11-13
fix(FirstPersonControl): Fix a bug introduced in #881.

gliegard:
## Description
 Fix a bug introduced in #881 in FirstPersonControl

## Motivation and Context
Before this change, this controls was going up when we pressed key up instead of going forward as expected. Now it goes forward again when we press key up.


zarov (approved):
Good call on moving the time computing ! LGTM

gliegard:
thanks

PR #912, opened 2018-11-20
refactor(renderer): LayeredMaterial splitted and simplified

gchoqueux:
Replace PR  #849 

Work has been done to simplify `LayeredMaterial` and its surrounding
parts, especially with the management of textures in shaders.

`LayeredMaterial` has been splitted, and has now a `LayeredMaterialLayer`
property: this object will contain the textures linked to the material,
and uniforms related to each material.

Uniforms from `LayeredMaterial` are no longer accessible as is, and should
be accessible with directly using the uniform name. The material does
not do a lot of things, apart attaching a layer to it and managing
uniforms.

TileMesh has been simplified, removing all accessors in favor of direct
access to its properties. Material operations that were executed in it
have also been removed.

Shaders have been rewritten to reflect change quoted above. In an
attempt to simplify things and make them more comprehensible, TileFS and
TileVS have been splitted in multiple chunks, which are installed
through ShaderChunk.install.

BREAKING CHANGES: the API of TileMesh and LayeredMaterial have changed.
BREAKING CHANGES: LayeredMaterialConstants and RendererConstant have
been removed.


src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
```diff
@@ -0,0 +1,70 @@
...
+uniform sampler2D   colorTextures[NUM_FS_TEXTURES];
+uniform vec4        colorOffsetScales[NUM_FS_TEXTURES];
+uniform Layer       colorLayers[NUM_FS_TEXTURES];
+uniform int         colorTextureCount;
+
+vec3 uvs[NUM_CRS];
+
+float getBorderDistance(vec2 uv) {
+    vec4 p4 = vec4(uv, 1. - uv);
+    vec2 p2 = min(p4.xy, p4.zw);
```
zarov: You can do directly `vec2 p2 = min(uv, 1. -uv);`

src/Renderer/MaterialLayer.js
```diff
@@ -0,0 +1,145 @@
...
+                }
+            }
+
+            if (__DEBUG__) {
+                if (index != extents.length) {
+                    console.error(`non-coherent result ${index} vs ${extents.length}.`, extents);
+                }
+            }
+        }
+        return {};
```
zarov: No need for this.

src/Renderer/MaterialLayer.js
```diff
@@ -0,0 +1,145 @@
...
+        this.level = (texture && (index == 0)) ? texture.coords.zoom : this.level;
+        this.textures[index] = texture || null;
+        this.offsetScales[index] = offsetScale;
+        this.material.layersNeedUpdate = true;
+    }
+
+    setTextures(textures, pitchs) {
+        this.dispose(false);
+        for (let i = 0, il = textures.length; i < il; ++i) {
+            // TEST if (textures[i]) is needed ?
```
zarov: It seems unnecessary indeed, I don't see a case where `textures[i]` is not set, but one way to confirm this would be to write some tests

src/Renderer/MaterialLayer.js
```diff
@@ -0,0 +1,145 @@
...
+            console.error('Unknown crs:', tileMT);
+        }
+
+        // Define color properties
+        let _valueOpacity = layer.opacity !== undefined ? layer.opacity : true;
+        Object.defineProperty(this, 'opacity', {
+            get: () => _valueOpacity,
+            set: (value) => {
+                if (_valueOpacity !== value) {
+                    if (value === 0) {
```
zarov: This may cause an error: when setting the opacity to 0 and then setting it back to a higher value, the layer is still invisible on screen.

src/Renderer/Shader/Chunk/elevation_pars_vertex.glsl
```diff
@@ -0,0 +1,39 @@
...
+        float zmin;
+        float zmax;
+    };
+
+    uniform Layer       elevationLayers[NUM_VS_TEXTURES];
+    uniform sampler2D   elevationTextures[NUM_VS_TEXTURES];
+    uniform vec4        elevationOffsetScales[NUM_VS_TEXTURES];
+    uniform int         elevationTextureCount;
+
+    highp float decode32(highp vec4 rgba) {
```
zarov, line 15: No need to specify `highp` ? The only time this shader is used is with `precision_qualifier`.

src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
```diff
@@ -0,0 +1,70 @@
...
+    // #pragma unroll_loop
+    for ( int i = 0; i < NUM_CRS; i ++ ) {
+        if ( i == layer.crs ) uv = uvs[ i ];
+    }
+
+    float borderDistance = getBorderDistance(uv.xy);
+    if (textureOffset != layer.textureOffset + int(uv.z) || borderDistance < minBorderDistance ) return vec4(0);
+    vec4 color = texture2D(texture, pitUV(uv.xy, offsetScale));
+    if(color.a > 0.0) {
+        if(layer.effect > 2.0) {
```
zarov, line 58: This part needs a better thinking, those conditions seem quite ugly imho.

zarov (changes requested):
Quick review, I'll take a better look at it next week

src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
mbredif: agreed, but as it is the old code untouched, I propose to improve it in another PR to keep this one reasonnable (!).

src/Renderer/Shader/Chunk/elevation_pars_vertex.glsl
mbredif: this function is plainly incorrect if vec4 is not highp, so I d rather keep the modifiers here to document it, even if it is redundant.

src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
mbredif: sure !

src/Renderer/Shader/Chunk/elevation_pars_vertex.glsl
zarov: That's weird but ok

src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
zarov: Yes no problem

src/Layer/Layer.js
```diff
@@ -53,6 +53,7 @@ class Layer extends THREE.EventDispatcher {
     constructor(id, type, config = {}) {
         super();
 
+        // ??
```
zarov: What is that ?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -26,67 +25,7 @@ function getSourceExtent(node, extent, targetLevel, source) {
...
-                pitch);
-            elevation.min = min;
-            elevation.max = max;
-        }
-
-        node.setTextureElevation(layer, elevation, pitch);
-    }
-}
-
+// TODO Move to XBilParser
```
zarov, line 28: Todo when ?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -168,24 +115,29 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
...
-        if (material.indexOfColorLayer(layer.id) === -1) {
-            material.pushLayer(layer, node.getCoordsForSource(layer.source));
-            const imageryLayers = context.view.getLayers(l => l.type === 'color');
-            const sequence = ImageryLayers.getColorLayersIdOrderedBySequence(imageryLayers);
+        if (!nodeLayer) {
+            // Create new MaterialLayer
+            const tileMT = layer.options.tileMatrixSet || extentsDestination[0].crs();
+            nodeLayer = material.addLayer(layer, tileMT);
+
+            // TODO: Sequence must be returned by parent geometry layer
```
zarov, line 133: Why not do it now ?

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: I have a new PR just after this PR

src/Layer/Layer.js
gchoqueux: I create a issue to explain my comment

gchoqueux: see #921

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: This PR is already big, it would take a new PR. I add a new issue.

gchoqueux: I open #922

src/Layer/Layer.js
zarov: Thanks, but please remove these comments, or replace them by linking to the issue

gchoqueux: removed

zarov (approved):
Ok everything seems good to me, I think we shouldn't have missed anything since it's the third time on this code ;)

gchoqueux:
@mbredif and @zarov, thank you for all your efforts!

PR #907, opened 2018-11-12
refacto(layer): deprecate adding a layer with only a configuration 

zarov:
As suggested in #880 I'm starting to lock the "finale API" of iTowns, with retrocompatibility of course.

First commit is the locking, with all examples changed as needed.

Second commit is a new tutorial. This could be done outside this PR, but I wrote it alongside these modifications, so I would like to add it together.
This tutorial will be modified further along the locking of the API.

Issue #880, opened 2018-10-19
Proposition for locking a final design in itowns


zarov:
We are currently (a lot offline) experiencing some disagreements on the design of iTowns and its modularity. So in order to share it with everyone, here's my shot at a proposition of where I see the design going, in hope to see it locked and stop refactoring it.

### TL;DR

- stop to rely on big configuration objects for initialization
- define a specified scope for the available API
- allow for a user to easily add a custom source, parser, etc...
- specify an internal format for handling data (the famous Feature)

### Final Example

Here's a complete example (of a globe view with a simple ortho layer, and a wfs one) illustrating my proposition:

```js
// Specify the position, with a real coordinate
var position = new itowns.Coordinate('WGS84', 2.35, 48.8, 25e6);

var viewerDiv = document.getElementById('viewerDiv');

// Create a view
var view = new itowns.GlobeView(viewerDiv, position);

// Specify the ortho source
var orthoSource = new itowns.WMTSSource({
    url: 'http://wxs.ign.fr/<key>/geoportail/wmts',
    format: 'image/jpeg',
    attribution : {
        name: 'IGN',
        url: 'http://www.ign.fr/'
    },
    name: 'ORTHOIMAGERY.ORTHOPHOTOS',
    tileMatrixSet: 'PM',
    tileMatrixSetLimits: {
        '2': { minTileRow: 0, maxTileRow: 4, minTileCol: 0, maxTileCol: 4 },
        '3': { minTileRow: 0, maxTileRow: 8, minTileCol: 0, maxTileCol: 8 },
        '4': { minTileRow: 0, maxTileRow: 16, minTileCol: 0, maxTileCol: 16 },
        '5': { minTileRow: 0, maxTileRow: 32, minTileCol: 0, maxTileCol: 32 },
        '6': { minTileRow: 1, maxTileRow: 64, minTileCol: 0, maxTileCol: 64 },
        '7': { minTileRow: 3, maxTileRow: 128, minTileCol: 0, maxTileCol: 128 },
        '8': { minTileRow: 7, maxTileRow: 256, minTileCol: 0, maxTileCol: 256 },
        '9': { minTileRow: 15, maxTileRow: 512, minTileCol: 0, maxTileCol: 512 },
        '10': { minTileRow: 31, maxTileRow: 1024, minTileCol: 0, maxTileCol: 1024 },
        '11': { minTileRow: 62, maxTileRow: 2048, minTileCol: 0, maxTileCol: 2048 },
[...]

PR #917, opened 2018-11-22
Refacto sources as instance

zarov:
Same vein as #907, to answer the #880 discussion

First commits makes all the changes in Sources and affiliated classes. Documentation has also been updated.

Second commit is the update of the Creating a globe view tutorial.

Third commit is the update of the examples.

### DIFF
--- a/examples/globe_vector_tiles.html
+++ b/examples/globe_vector_tiles.html
@@ -38,7 +38,8 @@
             setupLoadingScreen(viewerDiv, view);
 
             promises.push(itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(config) {
-                var layer = new itowns.ColorLayer(config.id, config);
+                config.source = new itowns.WMTSSource(config.source);
+                var layer = new itowns.ColorLayer('Ortho', config);
                 return view.addLayer(layer);
             }));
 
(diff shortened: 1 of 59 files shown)


examples/globe_vector_tiles.html
```diff
@@ -65,25 +66,26 @@
...
+                        name: 'OpenStreetMap',
+                        url: 'http://www.openstreetmap.org/',
+                    },
+                    zoom: {
+                        min: 2,
+                        max: 14,
+                    },
+                    tileMatrixSet: 'PM',
+                    projection: 'EPSG:3857',
+                    isInverted: true,
```
gchoqueux, line 83: `origin` to `isInverted`, it's deprecated or a breaking change?

src/Source/Source.js
```diff
@@ -1,45 +1,39 @@
...
     constructor(source) {
-        if (!source.protocol) {
-            throw new Error('New Source: protocol is required');
-        }
-
         if (!source.url) {
             throw new Error('New Source: url is required');
         }
 
+        this.url = source.url;
```
gchoqueux, line 40: A generic source doesn't necessarily need a `url` because it could be an object instance.

src/Source/Source.js
```diff
@@ -1,45 +1,39 @@
...
     /**
-     * Source are parameters to fetch source
-     *
-     * To extend {@link Source}, it is necessary to implement 2 functions:
-     * {@link Source#urlFromExtent} and {@link Source#extentInsideLimit}
-     * @param  {sourceParams}  source  object to set source
+     * @param {Object} source - An object that can contain all properties of a
+     * Source. Only the <code>url</code> property is mandatory.
      *
+     * @constructor
```
gchoqueux, line 31: Could you add `isSource` as new Source property?

gchoqueux (changes requested):
Thanks

examples/globe_vector_tiles.html
zarov: Deprecation, I forgot to add retrocompatibility, will do !

src/Source/Source.js
zarov: Yes, I agree, so maybe we need to remove the check above ?

src/Source/Source.js
zarov: Ok

src/Source/Source.js
gchoqueux: ok, we could leave that for now and change when can add an object instance

gchoqueux (approved):
it's good to squash and rebase

PR #918, opened 2018-11-22
docs: add a simple tutorial on geometry layer

zarov:
Another tutorial on geometry layers this time, really simple and not going too much into details. It is not as good as the first one I think, so please tell me what I can change !


Mkonini:
Besides small changes in legends, I like it as is!

gchoqueux (approved):
thanks

PR #920, opened 2018-11-22
Proj def refacto

gliegard:
## Description
refactoring : Add a file to describe projections.

## Motivation and Context
The goal is to avoid re write projection description.

### DIFF
--- a/examples/cubic_planar.html
+++ b/examples/cubic_planar.html
@@ -11,6 +11,7 @@
     <body>
         <div id="viewerDiv"></div>
         <script src="../dist/itowns.js"></script>
+        <script src="js/proj4defs/3946.js"></script>
         <script src="js/loading_screen.js"></script>
         <script type="text/javascript">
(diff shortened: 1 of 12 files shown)


zarov (commented):
Good idea, you are missing however four examples: pointcloud, orientation_utils, wfs and cubic_planar.

examples/cubic_planar.html
```diff
@@ -11,6 +11,7 @@
     <body>
         <div id="viewerDiv"></div>
         <script src="../dist/itowns.js"></script>
+        <script src="js/ProjDef.js"></script>
```
gchoqueux: may be use specific file by def, like
```html
<script src="js/Proj4Def/3946.js"></script>
```
or use link to [spatialreference](http://spatialreference.org/)
```html
http://spatialreference.org/ref/epsg/XXXX/proj4js/
```

gchoqueux (approved):

PR #928, opened 2018-11-27
Elevation processing

gchoqueux:
## Description
* move responsibility to replace `nodatavalue` in `MaterialLayer`
* move elevation data process in `xbilparser`

## Motivation and Context
this PR is part of the simplification of `LayeredMaterialNodeProcessing`.
It clarifies the distribution of jobs after fetching elevation data.

### DIFF
--- a/src/Parser/XbilParser.js
+++ b/src/Parser/XbilParser.js
@@ -7,7 +7,7 @@
   * @param      {THREE.Vector4}  pitch  The pitch,  restrict zone to parse
   * @return     {Object}  The minimum maximum elevation.
   */
-function computeMinMaxElevation(buffer, width, height, pitch) {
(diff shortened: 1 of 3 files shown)


zarov (commented):
It seems to me that we put things in `XbilParser` that have nothing to do (directly) with parsing a `xbil` file. What about moving it to `ElevationLayer` or something like `ElevationUtils` or something else ?

gchoqueux:
I would move his content in `DEMUtils`, which could be renamed `ElevationUtils`?

zarov:
In fact after looking at `DEMUtils`, this file is quite a mess too. Maybe let's keep what you did and view how to organize those things later ?

gchoqueux:
ok

zarov (approved):

PR #932, opened 2018-11-27
Ecma script 2016

gchoqueux:
up to Ecma script 2016

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,7 @@ module.exports = {
    'eslint-config-airbnb-base/rules/strict',
  ],
   parserOptions: {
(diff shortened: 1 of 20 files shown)


zarov:
Beside `**` and `includes`, could you provide a list of things we could update ?

gchoqueux:
@zarov see #943

zarov (approved):
You forgot some places:
- `Array.prototype.indexOf` in `View`
- `Math.pow` in `PlanarLayer`

Otherwise it's good

gchoqueux:
@zarov  thanks.
I don't remove `indexOf` in view because index  is used to `splice`.

PR #933, opened 2018-11-27
Fix(examples):Center help div to avoid overlapping with debug menu

jailln:
## Description
Really small PR that centers the help div which displays the key bindings in the planar examples. It allows to avoid superposition with the debug menu, e.g. in planar example (see following screenshots).

Planar example before:

image

After:

image

### DIFF
--- a/examples/3dtiles.html
+++ b/examples/3dtiles.html
@@ -11,6 +11,10 @@
     </head>
     <body>
         <div id="viewerDiv"></div>
+        <div class="help">
+            <p><b>Batch Table Information:</b></p>
+            <ul id="info"></ul>
+        </div>
         <script src="js/GUI/GuiTools.js"></script>
         <script src="../dist/itowns.js"></script>
         <script src="js/loading_screen.js"></script>
@@ -113,9 +117,5 @@
             }
 
         </script>
-        <div class="help" style="right:0;left:auto">
-            <p><b>Information Batch Table: </b></p>
-            <ul id="info"></ul>
-        </div>
     </body>
 </html>
--- a/examples/css/example.css
+++ b/examples/css/example.css
@@ -35,7 +35,7 @@ body {
     position: absolute;
     z-index: 0;
     top: 0;
-    left: 0;
+    right: 0;
     color: #eee;
     font: 11px 'Lucida Grande',sans-serif;
     line-height: normal;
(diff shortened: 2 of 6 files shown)


gchoqueux:
We should rather modify the examples, because each use case can be different. (see examples `wfs.html`, `globe_wfs_extruded.html`...)

jailln:
Modify the example in which way ? Do you mean by putting the css related to the help div in each example (like in multiglobe) ?

gchoqueux:
My bad, I misspoke.
You are right, `menuDiv` and `help` div are by default positioned on the left, and so they overlap. 

As in most examples, `help` div is located on the right (look at the `style` changes in the examples : [wfs](https://github.com/iTowns/itowns/blob/326722f83b79477a53718c78aa4b19d127f1f15c/examples/globe_wfs_extruded.html#L19) or [3dtiles](https://github.com/iTowns/itowns/blob/master/examples/3dtiles.html#L116)). 

The best thing is that you put it right by default. After that you impact the examples with this new value.

if a user wants to position it differently he will be able to use `style` in declaration of the div
```html
<div class="help" style="....">
```

jailln:
In the two examples you mention ([wfs](https://github.com/iTowns/itowns/blob/326722f83b79477a53718c78aa4b19d127f1f15c/examples/globe_wfs_extruded.html#L19) and [3dtiles](https://github.com/iTowns/itowns/blob/master/examples/3dtiles.html#L116)), the div class 'help' is actually not an helper presenting the key bindings but a div allowing to display metadata attributes of the geometry (i.e. the help div has been "hijacked" for this purpose). 

My proposal would be (to be coherent in all the examples an to facilitates user/devel understanding):
  * We choose a position for:
     * the controls menu (e.g. top left)
     * the help div presenting the key bindings (e.g. top center or right)
     * the metadata ouput div (e.g. top center or right)
  * We update the examples you mention to display the metadata attributes in the 'metadata output div' instead of the help div.

gchoqueux:
Ok, for your proposals. By cons, for an example, I'm not for a display of `div` in the center, if there isn't another `div` to right or left.

jailln:
I ended up putting the help div on the top right corner by default and updating the wfs example which already had a div on the right side of the screen.

gchoqueux:
ok.
Could you, in the end, remove the style overload for the `div help` in these examples (because it's no longer useful):

- `./examples/3dtiles.html`
- `./examples/globe_wfs_color.html`
- `./examples/globe_wfs_extruded.html`
- `./examples/stereo.html`

jailln:
Yes, you're right. Done.
I also translated remaining french text to english.

gchoqueux (approved):
thanks

PR #939, opened 2018-11-29
Immersive view

gliegard:
## Description
This PR create an immersive view.
An immersive view is made of a set of panoramic view, each panoramic view is a set of texture projected on the scene.

- OrientedImageLayer : Layer used to load oriented images, and create projective texture mapping (using existing OrientedImageMaterial). It uses existing DataSourceProvider to load texture.

- OrientedImageSource : Build Url of the textures.

- StreetControls : camera controller to naviguate threw these panoramics

- An example with this immersive view in Paris.

immersif_paris


src/Controls/FirstPersonControls.js
```diff
@@ -182,7 +182,7 @@ class FirstPersonControls extends THREE.EventDispatcher {
     // Event callback functions
     // Mouse movement handling
     onMouseDown(event) {
-        event.preventDefault();
+        // event.preventDefault();
```
gchoqueux, line 186: Why are you doing this? to fix what?

I prefer remove and not comment code.

src/Controls/ImmersiveControls.js
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Adds a layer with datas.
+     *
+     * @param {OrientedImageLayer} layer - Add a layer
+     */
+    addLayer(layer) {
+        this.layers.push(layer);
+        this.currentLayer = layer;
+        this.currentLayerIndex = this.layers.length - 1;
```
gchoqueux: Can you explain the role of `this.currentLayerIndex`?

src/Controls/ImmersiveControls.js
```diff
@@ -0,0 +1,109 @@
...
+ * <li><b> left + right keys: </b> strafing movements </li>
+ * <li><b> pageUp + pageDown: </b> vertical movements </li>
+ * <li><b> mouse click+drag: </b> pitch and yaw movements (as looking at a panorama) </li>
+ * </ul>
+ * <ul> Bindings added
+ * <li><b> keys Z : </b> Move camera to the next panoramic </li>
+ * </ul>
+ * @extends FirstPersonControls
+ */
+class ImmersiveControls extends FirstPersonControls {
```
gchoqueux: It isn't necessary that this camera is specific to immersive. It should be a camera that follows a path. This path will consist of ordered positions. We could then use this camera for other user cases.

package.json
```diff
@@ -43,6 +43,7 @@
     "flatbush": "^3.1.0",
     "js-priority-queue": "^0.1.5",
     "pbf": "^3.1.0",
+    "string-format": "^0.5.0",
```
gchoqueux: If you use this, you could extend the use `Provider/URLBuilder`.
You could upgrade to 2.0.0 version.

src/Core/Scheduler/Scheduler.js
```diff
@@ -130,6 +131,7 @@ Scheduler.prototype.initDefaultProviders = function initDefaultProviders() {
     this.addProtocolProvider('tile', TileProvider);
     this.addProtocolProvider('3d-tiles', $3dTilesProvider);
     this.addProtocolProvider('potreeconverter', PointCloudProvider);
+    this.addProtocolProvider('orientedImage', OrientedImageProvider);
```
gchoqueux: You have to add a new provider, you can not use the `dataSourceProvider`?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,118 @@
...
+     */
+    constructor(id, config = {}) {
+        super(id, new THREE.Group(), config);
+
+        this.background = config.background || createSphere(config.sphereRadius);
+        this.background.layer = this;
+
+        this.object3d.add(this.background);
+
+        this.poses = null;
```
gchoqueux: could you explain the difference between `pose`, `pano` and `camera`?

src/Provider/OrientedImageProvider.js
```diff
@@ -0,0 +1,35 @@
...
+        return Promise.resolve();
+    }
+    // prepare a promise for each texture to load.
+    var promises = [];
+    // panoramic id
+    const imageId = pano.geometry[0].properties.id;
+    // request one texture for each camera of the panoramic
+    for (const camera of layer.cameras) {
+        var sensorId = camera.name;
+        var url = format(layer.source.url.images, { imageId, sensorId });
```
gchoqueux: In the logic of itowns, `url` should be built and provided by the `source`.

src/Source/OrientedImageSource.js
```diff
@@ -0,0 +1,68 @@
...
+     * @param { string } source.url.calibrations - Url of a configuration file for the set of cameras. see [CameraCalibrationParser]{@link module:CameraCalibrationParser.parse}
+     * @param { string } source.url.images - Url for all the textures.
+     * This Url must contains {sensorId} and {cameraId}, and these pattern will be replaced be the {@link OrientedImageProvider} o find the good texture for each camera for each panoramic.
+     */
+    constructor(source) {
+        source.format = source.format || 'json';
+        super(source);
+        const promises = [];
+
+        this.url = source.url;
```
gchoqueux: Already do in `Source`.

src/Source/OrientedImageSource.js
```diff
@@ -0,0 +1,68 @@
...
+        // option for GeoJsonParser
+        this.mergeFeatures = false;
+        this.crsOut = source.crsOut;
```
gchoqueux: ` this.crsOut` should not exist it should use `layer.projection` instead. `layer.projection` is the destination projection.

gchoqueux (changes requested):
Great work.
There are adjustments to get closer to the architecture of _itowns_.

package.json
zarov: IMHO we don't need this dependency and can rely on a simple RegExp here (and even without a RegExp, a `replace` would be sufficient).

src/Controls/FirstPersonControls.js
gliegard: It's fixes the key events, even if the code is in mouse down event. It's strange.
When I let the event.preventDefault() line in onMouseDown function , I don't receive key events any more, and the key binding to go forward/backward/... doesn't work.

I've removed the code.

package.json
gliegard: Ok, I've used a string.replace instead.

src/Controls/ImmersiveControls.js
gliegard: I've removed Layer from ImmersiveControls to simplify the control

src/Controls/ImmersiveControls.js
gliegard: I simplied this control, now it only store currentPosition and nextPosition , and it's the layer that is updating these twos current and next position.

src/Core/Scheduler/Scheduler.js
gliegard: No, I can't use dataSourceProvider, because the urls I need to build doesn't depend on extent.

src/Provider/OrientedImageProvider.js
gliegard: ok

src/Source/OrientedImageSource.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Build the url of the image, for a given panoramic id, and a given camera id.
+     *
+     * @param      {string}  cameraId  Id of the camera
+     * @param      {string}  panoId   Id of the panoramic
+     * @return     {string}  Url of the image
+     */
+    imageUrl(cameraId, panoId) {
```
zarov, line 40: Is there a point to have this method ? If not, please build the entire url in `urlFromExtent`.

gliegard: Yes, it's the most simple string replacing method, the heart of this source, and it doesn't change when we change the way the source is called.. So I keep it.

src/Controls/FirstPersonControls.js
```diff
@@ -182,7 +182,6 @@ class FirstPersonControls extends THREE.EventDispatcher {
     // Event callback functions
     // Mouse movement handling
     onMouseDown(event) {
-        event.preventDefault();
```
gchoqueux, line 185: could you add comment in code too

src/Controls/ImmersiveControls.js
gchoqueux: Could you rename and clean up class and properties. Could you remove `pano reference, this.layers`...

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+        const position = layer.panos[i].position;
+        const D = camera.position.distanceTo(position);
+        if (D < minD) {
+            minD = D;
+            minI = i;
+        }
+    }
+    // detection of oriented image change
+    const ori = layer.panos[minI];
+    if (ori && layer.currentPano != ori) {
```
gchoqueux: Could you uniformize the name `ori` and `pano`?
To avoid having multiple denomination for the same concept.

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+        layer.currentPano = ori;
+        ori.index = minI;
```
gchoqueux: if you save index to get next pano, save the next pano directly.

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+
+        if (context.view.controls && context.view.controls instanceof ImmersiveControls) {
```
gchoqueux: Could you use frame requester to update the control?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+            context.view.controls.setCurrentPanoPosition(layer.getCurrentPano().position);
+            context.view.controls.setNextPanoPosition(layer.getNextPano().position);
+        }
+
+        // create an array with informations for each texture needed.
+        const imagesInfo = [];
```
gchoqueux: could you use `Array.prototype.map` instead of loop?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+            requester: ori,
+        };
+
+        // async call to scheduler to get textures
+        context.scheduler.execute(command)
+            .then(textures => layer.material.setTextures(textures, ori));
+    }
+}
+
+function updateSphere(layer) {
```
gchoqueux: could you rename to `updateBackground`?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+    if (layer.background && layer.currentPano) {
+        layer.background.position.copy(layer.currentPano.position);
+        layer.background.updateMatrixWorld();
+        layer.background.material = layer.material || layer.background.material;
+    }
+}
+
+function createSphere(radius) {
```
gchoqueux: could you rename to `createBackground`?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+     * @constructor
+     * @param { string } id - Configuration object
+     * @param { Object } config - configuration of the layer
+     * @param { OrientedImageSource } config.source - Source of the datas
+     */
+    constructor(id, config = {}) {
+        super(id, new THREE.Group(), config);
+
+        this.background = config.background || createSphere(config.sphereRadius);
+        this.background.layer = this;
```
gchoqueux: could you explain double reference?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,166 @@
...
+
+        // wait for the twos promises to tell layer is ready.
+        this.whenReady = Promise.all([p1, p2]).then(() => {
+            this.ready = true;
+            return this;
+        });
+    }
+
+    update(context) {
+        this.material.updateUniforms(context.camera.camera3D);
```
gchoqueux, line 161: the material should be attached to a object 3d and not to the layer

examples/immersive_paris.html
```diff
@@ -0,0 +1,172 @@
...
+                if (properties.id.indexOf('bati_remarquable') === 0) {
+                    return new itowns.THREE.Color(0x5555ff);
+                } else if (properties.id.indexOf('bati_industriel') === 0) {
+                    return new itowns.THREE.Color(0xff5555);
+                }
+                return new itowns.THREE.Color(0xeeeeee);
+            }
+
+            function altitudeBuildings(properties) {
+                return properties.z_min - properties.hauteur - 3;
```
gchoqueux, line 76: Why this value: 3?

examples/immersive_paris.html
```diff
@@ -0,0 +1,172 @@
...
+            promises.push(itowns.Fetcher.json(orientationsUrl, networkOptions) );
+            promises.push(itowns.Fetcher.json(calibrationUrl, networkOptions) ) ;
+            Promise.all(promises).then((res) => {
+
+                var orientation = res[0];
+                var calibration = res[1];
+
+                // Create oriented image layer
+                var olayer = new itowns.OrientedImageLayer('demo_orientedImage', {
+                    sphereRadius: 50,
```
gchoqueux: `sphereRadius` it's to backgroung? Could you add default value? and remove in example?

examples/immersive_paris.html
```diff
@@ -0,0 +1,172 @@
...
+                    source: orientedImageSource,
+                    orientation: orientation,
+                    calibration: calibration,
+                    projection: view.referenceCrs,
+                    networkOptions: networkOptions
```
gchoqueux: Could you use default value?

gchoqueux (changes requested):
We arrive at the goal

src/Layer/OrientedImageLayer.js
gliegard: No, because I don't want the control to be updated at each frame.

src/Layer/OrientedImageLayer.js
gliegard: The only goal of this layer is to create and update this material.

examples/immersive_paris.html
gliegard: It's specific to this example. It add a delta of 3 meters, to be sure that the building start under the ground (define by the MNT).

src/Layer/OrientedImageLayer.js
gliegard: ok, I've removed it

examples/immersive_paris.html
gliegard: Yes it's for the background sphere. User can change the radius.
I prefer that the user see the option instead of hiding it.

src/Layer/OrientedImageLayer.js
gchoqueux: It's better to decouple the `controls` and the `layers` (as it is the case in the others). A layer must only manage its data and update it.

src/Layer/OrientedImageLayer.js
gchoqueux: Then this layer must inherit from `ColorLayer`

examples/immersive_paris.html
gchoqueux: a comment will be welcome

examples/immersive_paris.html
gchoqueux: if it's exposed, it should be documented

src/Layer/OrientedImageLayer.js
gliegard: ok, I add a callback to the layer, so it doesn't have to know the controls

examples/immersive_paris.html
```diff
@@ -0,0 +1,178 @@
...
+            // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+            var viewerDiv = document.getElementById('viewerDiv');
+
+            var orientedImageLayer;
+
+            // Instanciate iTowns GlobeView
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe, {
+                noControls: true,
+                handleCollision: false,
+                sseSubdivisionThreshold: 10,
```
gchoqueux, line 38: Why you set `sseSubdivisionThreshold` to 10?

examples/immersive_paris.html
```diff
@@ -0,0 +1,178 @@
...
+            // These will deform iTowns globe geometry to represent terrain elevation.
+            function addElevationLayerFromConfig(config) {
+                config.source = new itowns.WMTSSource(config.source);
+                var layer = new itowns.ElevationLayer(config.id, config);
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            }
+            itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
+
+
+            function colorBuildings(properties) {
```
gchoqueux: `colorBuildings` isn't necessary in this example. You should remove it to simplify this example.

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * @constructor
+     * @param { string } id - Configuration object
+     * @param { Object } config - configuration of the layer
+     * @param { OrientedImageSource } config.source - Source of the datas
+     */
+    constructor(id, config = {}) {
+        super(id, new THREE.Group(), config);
+
+        this.background = config.background || createBackground(config.sphereRadius);
```
gchoqueux: could you rename `sphereRadius` by `distanceOfBackground`?

src/Layer/OrientedImageLayer.js
```diff
@@ -0,0 +1,169 @@
...
+            // add position and quaternion attributes from point feature
+            let i = 0;
+            for (const pano of this.panos) {
+                // set position
+                coord.set(pano.crs, pano.vertices[0], pano.vertices[1], pano.vertices[2]);
+                pano.position = coord.xyz();
+                // set quaternion
+                pano.quaternion = OrientationUtils.quaternionFromAttitude(pano.geometry[0].properties, coord, isGlobe);
+
+                // TODO clean DataSourceProvider, so that we don't have this hack to do
```
gchoqueux, line 134: the hack is only  `pano.material = {};` ?
Could you add issue to not forget and motivate to work on subject?

src/Source/OrientedImageSource.js
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @constructor
+     * @param { Object } source - Configuration object
+     * @param { string } source.url - Url for all the textures.
+     * This Url must contains {sensorId} and {cameraId}, and these pattern will be replaced be the {@link OrientedImageProvider} o find the good texture for each camera for each panoramic.
+     */
+    constructor(source) {
+        source.format = source.format || 'json';
+        super(source);
+        this.protocol = 'orientedImage';
```
gchoqueux: `this.protocol` is from source could you use `this.isOrientedImageSource = true`

src/Controls/StreetControls.js
```diff
@@ -0,0 +1,111 @@
+import FirstPersonControls from 'Controls/FirstPersonControls';
+import AnimationPlayer, { AnimatedExpression } from 'Core/AnimationPlayer';
```
gchoqueux, line 3: `AnimationPlayer` and `AnimatedExpression` will be removed soon because it's better to use TWEEN.js. This to be done later.

src/Controls/StreetControls.js
```diff
@@ -0,0 +1,111 @@
...
+ * <li><b> left + right keys: </b> strafing movements </li>
+ * <li><b> pageUp + pageDown: </b> vertical movements </li>
+ * <li><b> mouse click+drag: </b> pitch and yaw movements (as looking at a panorama) </li>
+ * </ul>
+ * <ul> Bindings added
+ * <li><b> keys Z : </b> Move camera to the next position </li>
+ * </ul>
+ * @extends FirstPersonControls
+ */
+class ImmersiveControls extends FirstPersonControls {
```
gchoqueux: Could you rename `ImmersiveControls` in `StreetControls`?

src/Controls/StreetControls.js
```diff
@@ -0,0 +1,111 @@
...
+        this.setCameraOnPosition(this.currentPosition, this.nextPosition);
+    }
+
+    /**
+     * Set the camera on a position, looking at another position.
+     *
+     * @param      { THREE.Vector3 }  currentPosition   The position to set the camera
+     * @param      { THREE.Vector3 }  nextPosition      The position where the camera look at.
+     */
+    setCameraOnPosition(currentPosition, nextPosition) {
```
gchoqueux: `currentPosition` could you rename in `position` it's more generic and `nextPosition` to `lookAt` it's more adapted.

src/Controls/StreetControls.js
```diff
@@ -0,0 +1,111 @@
...
+        if (!currentPosition || !this.nextPosition) {
+            return;
+        }
+        this.camera.position.copy(currentPosition);
+        this.camera.up.copy(currentPosition.clone().multiplyScalar(1.1).normalize());
+        this.camera.lookAt(nextPosition.clone());
```
gchoqueux: it'sn't necessary to `clone` and `multiplyScalar`.
```js
this.camera.up.copy(currentPosition).normalize();
this.camera.lookAt(nextPosition);
```

`this.camera.up.copy(currentPosition).normalize();` will not work for `PlanarView`.

src/Controls/StreetControls.js
```diff
@@ -0,0 +1,111 @@
...
+        this.reset();
+    }
+
+    /**
+     * Move the camera
+     *
+     * @param { THREE.Vector3 }  positionTo  Destination of the movement.
+     */
+    moveCameraTo(positionTo) {
+        this.positionFrom = this.camera.position.clone();
```
gchoqueux: could you use `copy` instead of `clone`?

examples/immersive_paris.html
gliegard: It was a way to control subdivision, to avoid requesting many unnecessary tiles ; so it's to get better performances.

gliegard: It's not needed since I've change the far of the camera to get better performances, but I think it's important to keep it, for other uses.

examples/immersive_paris.html
gliegard: yes

src/Layer/OrientedImageLayer.js
gliegard: No, I think sphereRadius is clearer

src/Layer/OrientedImageLayer.js
gliegard: yes, it's done here #990

src/Source/OrientedImageSource.js
gliegard: It's seem to be unused, so I've remove this protocol line.

src/Controls/StreetControls.js
gliegard: yes

src/Controls/StreetControls.js
gliegard: yep, good idea

src/Controls/StreetControls.js
gliegard: ok thanks.

src/Controls/StreetControls.js
gliegard: ok

src/Controls/StreetControls.js
gliegard: I have an old example of ImmersiveView in planarView, I can add it on this PR to verify that it works well on planar , and eventually debug it. Should I add it on this PR ? WDYT ?

src/Layer/OrientedImageLayer.js
gchoqueux: good

src/Layer/OrientedImageLayer.js
gchoqueux: Indeed we can see it's the radius of a sphere but it's the sphere of what?
`distanceOfBackground`, this parameter name is closer to the immersive scope.

src/Source/OrientedImageSource.js
gchoqueux: but we must add `this.isOrientedImageSource = true`

src/Controls/StreetControls.js
gchoqueux: If you want but it's not necessary at the moment.
Later you could create an example in planar with points cloud.
An example consisting only of the background (without globe and building) would be interesting because it could replace **Panoramas example**.

gchoqueux:
### Notes on use

- [ ] The camera is initialized down, it would be better should place it in better direction immediately.
- [x] A help of control commands.
- [x] The camera doesn't follow the next panoramic.
- [ ] Interface debug layer is empty.
- [x] There's not panoramic image on ground?
- [x] In end of path, if you click during the return to the beginning you come back to the end of the street
- [x] To correct of positioning defects it would be necessary to use an offset matrix. In particular to correct the problem on the ground.

gliegard:
> The camera is initialized down, it would be better should place it in better direction immediately.

Yes.. This bug don't appear everytime, so I let it for the moment

> A help of control commands.

Thanks, it've added it

> The camera doesn't follow the next panoramic.

Yes it's the expected behavior , camera keep it's orientation (changed by the user with the mouse)

> Interface debug layer is empty.

It's a bug on all other examples too

> There's not panoramic image on ground?

Not yet, I'll plan to do it later in another PR

> In end of path, if you click during the return to the beginning you come back to the end of the street

Yes, It's because it's the more simple StreetControls for the moment.

> To correct of positioning defects it would be necessary to use an offset matrix. In particular to correct the problem on the ground.

Why not, I'm not really sure if it's a problem on the MNT or a problem on the panoramic position.
Thanks, an offset is a good idea.

src/Source/OrientedImageSource.js
gliegard: ok

src/Layer/OrientedImageLayer.js
gliegard: Ok, I've changed it to backgroundDistance !

gchoqueux:
> The camera is initialized down, it would be better should place it in better direction immediately.
Yes.. This bug don't appear everytime, so I let it for the moment

Ok, could you open issue?

> Interface debug layer is empty.
It's a bug on all other examples too

Ok, could you open issue?

gchoqueux (approved):
Thanks, great work!!

PR #950, opened 2018-12-04
fix(geometryLayer): wireFrame property change 

### LINKED ISSUE #940: 3D Tiles debug tools wireframe mode also affects the bounding boxes
## Your Environment
* Version used: Current master, SHA1 810a756
* Browser Name and version: Firefox 62.0.3
* Operating System and version (desktop or mobile): Ubuntu 18.04.1 LTS

## Context
The wireframe mode of the debug tools of 3D Tiles layers also affects the representation of the bounding boxes:

image

1. Open [3D tiles example](https://github.com/iTowns/itowns/blob/master/examples/3dtiles.html)
2. Zoom in 
3. Expand 'Layer 3d-tiles-discrete-lod' in the menu
4. Check 'Bounding boxes'
5. Check 'Wireframe'
6. Uncheck 'Wireframe' --> the bounding box is represented as a volume

## Expected Behavior
Only the geometry of the 3D tiles layer is set (and unset) in wireframe.  

## Actual Behavior
The bounding box is also set and unset in wireframe.

gliegard:
## Description
fix(geometryLayer): wireFrame property change only affect This layer objects. 

Before this commit, it could also affect objects on attached layers (for example, bounding box from attached debug layers). 

## Motivation and Context
fix #940

### DIFF
--- a/src/Layer/GeometryLayer.js
+++ b/src/Layer/GeometryLayer.js
@@ -80,7 +80,7 @@ class GeometryLayer extends Layer {
                     object.material.wireframe = this.wireframe;
                 } else if (object.content && object.content.layer == this) {
                     object.content.traverse((o) => {
-                        if (o.material) {
+                        if (o.material && o.layer == this) {
                             o.material.wireframe = this.wireframe;
                         }
                     });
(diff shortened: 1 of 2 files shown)


zarov (commented):
Could you fix also the update of the view when checking the `boundingBox` options in debug ? Thanks

zarov (approved):
Thanks !

PR #951, opened 2018-12-04
feat(layer): add isXXXLayer properties to layers objects

zarov:
In the same vein than Sources, Layers have a isXXXLayer property,
helping identifying the type of the layer.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -155,7 +155,7 @@ const _syncGeometryLayerVisibility = function _syncGeometryLayerVisibility(layer
 };
 
 function _preprocessLayer(view, layer, provider, parentLayer) {
-    if (!(layer instanceof Layer)) {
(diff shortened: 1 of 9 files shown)


src/Core/View.js
```diff
@@ -585,7 +585,7 @@ View.prototype.pickObjectsAt = function pickObjectsAt(mouseOrEvt, radius, ...whe
     radius = radius || 0;
 
     for (const source of sources) {
-        if (source instanceof Layer ||
+        if (source.isLayer ||
```
gchoqueux, line 588: it's an oversight of the PR #917 ?

gchoqueux (approved):
thanks

src/Core/View.js
zarov: Yes !

PR #952, opened 2018-12-05
Fix panoramic

gchoqueux:
Update `StaticSource` with the new `LayeredMaterial` and `layerMaterialProcessing`


zarov (approved):
LGTM, thanks

PR #953, opened 2018-12-05
chore: update THREE r98 to r99

gchoqueux:
### Migration Guide

- [ ] WebGLRenderTarget.texture.generateMipmaps is now set to false by default.
- [ ] There is a new (not backwards compatible) implementation for SSAOShader and SSAOPass.
- [ ] JSONLoader has been removed from core. It is now located in examples/js/loaders/deprecated/LegacyJSONLoader.js.
- [ ] Removed Geometry support from ObjectLoader. You have to include LegacyJSONLoader if you still want to load geometry data of type Geometry.
- [ ] Removed Geometry support from SkinnedMesh. Use BufferGeometry instead.
- [ ] Removed SkinnedMesh.initBones(). The SkinnedMesh constructor does not build the bone hierarchy anymore. You have to do this by yourself and then call SkinnedMesh.bind() in order to bind the prepared skeleton.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3974,8 +3974,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


gchoqueux:
I close because travis build on a non-existent branch.

PR #958, opened 2018-12-07
fix(debug): disable fog with camera debug rendering

gchoqueux:


zarov:
Where can we reproduce this bug ? I can't get to it.

zarov (approved):
Ok I understand now ! LGTM

PR #960, opened 2018-12-10
Best place to layer WFS update state 

gchoqueux:
Decreases the wfs process


zarov (commented):
Couldn't it even be moved before `const features` ?

gchoqueux:
> Couldn't it even be moved before const features ?

Yes, thanks, PR updated

zarov (approved):
Thanks, LGTM

Issue #911, opened 2018-11-15
New documentation layout proposal


zarov:
I have never been a big fan of the layout of the current iTowns documentation. 

So I have started working on a new one, as can be seen here https://zarov.github.io/itowns/docs/out/. It still looks like the THREE.js documentation (as seen here https://threejs.org/docs/), which I'm quite fond of. 

What I propose is this:
- a simple navigation, with (for now) two sections: general documentation like __Tutorials__, and the __API__
- the __API__ documentation is sorted not by classes/interfaces/modules as usual in jsdoc, but by packages, following a defined list (https://github.com/zarov/itowns/blob/gh-pages/docs/navigation.json)
- there will be a field to search the documentation, like in THREE.js
- a light colorscheme and visual, compared to the big one we currently have
- BONUS: automatically build a jsfiddle for each tutorial 

Once again, this is totally my point of view, so I would like to maybe get more inputs from others before going further. Thanks !

Mkonini:
Good job!
Suggestion: I think it would be better to rename the first module of API documentation "socle" instead of "core" which encompasses the all

zarov:
Yes, maybe `Base` instead: `Socle` is not very used as far as I know.

gliegard:
Great doc !!

gliegard:
Looking at the doc, for example  [Coordinate page](https://zarov.github.io/itowns/docs/out/#api/Geographic/Coordinates), I think it's too white : all the page is styled with white background.
- Example boxes should be like old doc color : dark blue background, or grey but not white
- Same remark for Method boxes, I prefer old doc color : blue background

zarov:
Thanks @gliegard, I updated the general colors

gliegard:
Nice ! And can you add something to separate two methods, like an horizontal line ?

zarov:
Yes sorry missed that, I pushed it, it should be visible in a few minutes

gliegard:
Also, I've seen a 404 error on [InfoLayer](https://zarov.github.io/itowns/docs/out/#api/Layer/InfoLayer) and [ColorLayersOrdering](https://zarov.github.io/itowns/docs/out/#api/Layer/ColorLayersOrdering)

zarov:
I'm writing the missing documentation for them ! Thanks

zarov:
I'm writing the missing documentation for them ! Thanks

zarov:
I'm letting this issue opened for two months, if someone wants to talk about change on the layout.

[...]

PR #962, opened 2018-12-10
New JSDOC layout

zarov:
As explained in #911 I wrote a new jsdoc layout. The result is visible here https://zarov.github.io/itowns/docs/out/

Three commits in this PR:
- a first one to move the documentation from `API_Doc` to `docs`, giving a more explicit URL
- a second with the implementation of the new layout
- a third with some update in the current documentation

There is still things to do, but I feel that this implementation is sufficiently advanced to start pushing it now.


jailln:
Thanks for this work, the documentation is much easier to read this way.

Just one remark: There is remaining references to Providers ([at least in the Scheduler](https://zarov.github.io/itowns/docs/out/#api/Base/Scheduler)) in the doc (and in the code). They should be replaced by Sources ?

zarov:
Thanks ! Yes `Providers` are still present, as they are going deprecated but still used for some usecases (3dtiles, pointcloud...). Concerning the documentation, I would let them but I could at least generate correctly the documentation of the `Provider` interface.

mbredif:
That looks very good !
some quick remarks
- the tutorials could have a disclaimer about the geoportail key usage restriction or use a key-less ogc source (maybe as an alternative).
- the following keywords yield nothing : pointcloud, potree, 3dtiles (or variations), mapbox

zarov:
> the tutorials could have a disclaimer about the geoportail key usage restriction or use a key-less ogc source (maybe as an alternative).

If you have some alternative I'm for it, otherwise I'll put a disclaimer.

> the following keywords yield nothing : pointcloud, potree, 3dtiles (or variations), mapbox

That's normal, the search is not a search, but a filter in the current menu: it doesn't search inside the documentation. This feature is for later, because more complex to handle in a static website.

nosy-b (approved):
Nice!
Agreed with @mbredif remarks but that's already an improvement

zarov:
Thanks ! I know there is still some documentation to update, but I suggest to open other PRs to do it and quickly close this one to see those updates in the new layout.

Issue #954, opened 2018-12-05
Support for PNG8 format


SesticM:
Provide support for `image/png8` texture format via `DataSourceProvider.js`.

There are two PNG formats commonly in use: PNG-8 and PNG-24. The numbers represent bit depth. Current "image/png" format actually points to 24-bit PNG nomenclature. It would be of much help to support 8-bit PNG format, as resulting raster sizes do differ significanly without observable loss - please see examples below:

24-bit PNG image (195 KB):
![image-png-24](https://user-images.githubusercontent.com/11457703/49547703-c15a8600-f8e3-11e8-859b-366c34aa3231.png)

8-bit PNG image (65 KB):
![image-png-8](https://user-images.githubusercontent.com/11457703/49547740-db946400-f8e3-11e8-923c-29c7468cb752.png)

zarov:
If I'm not mistaken, the fix is pretty simple: adding the entry `'image/png8': Fetcher.texture` [here](https://github.com/iTowns/itowns/blob/master/src/Provider/DataSourceProvider.js#L7-L16) should do the trick.

Note that this is something we are currently working on: it is quite a pain to have strings defining a lot of things in iTowns, and we are pushing changes to ease the usage of simple format (like png8) that we won't think of initially.

So please test this fix meanwhile, if it is working we can push it in `master`, so it could be used while we are changing things.

SesticM:
Yes, you're correct about this exact implementation related to `Fetcher.texture` part. On the other hand - why do you need to validate texture format at all?

zarov:
It is an error in the design in iTowns: it shouldn't be validated, and we're working on fixing that. In my opinion, by default we should always use `Fetcher.texture`, as an other format always comes with a parser (`geojson` for example). I hope to fix this before the end of the year. 

If you have suggestions for the API, don't hesitate to discuss in #880 or to keep opening issues ! Thanks

zarov:
Fixed in #965, by default `Fetcher.texture` is used and you can even specify your own fetching method.

PR #965, opened 2018-12-12
feat(fetcher): add the possibility to specify fetcher

zarov:
Before this commit, the fetching in DataSourceProvider (default
providers for most resources) was limited to a list of format linked to
a fetching method. Now, it is possible to declare the fetching method in
the Source, and by default it is using the texture fetching.

Note that this doesn't break and change anything in the current way of
doing things. The default value being `Fetcher.texture` is of course
subject to some debate.

Solves #954


gchoqueux (approved):
LGTM, thanks

PR #966, opened 2018-12-12
feat(parser): add the possibility to specify parser

zarov:
(to review once #965 is merged)

Before this commit, the parsing in DataSourceProvider (default provider
for most resources) was limited to a list of format linked to a parsing
method. Now, it is possible to declare the parsing method in the Layer,
and by default it does no parsing at all.

A few remarks: I intend to write some documentation on parsers, as a 
tutorial or as an interface (same as we have with `Provider` for
example). I'd like to add this documentation in this PR, so it is still a 
WIP for now.


src/Layer/Layer.js
```diff
@@ -77,6 +112,7 @@ class Layer extends THREE.EventDispatcher {
         this.defineLayerProperty('frozen', false);
 
         this.info = new InfoLayer(this);
+        this.parser = config.parser;
```
gchoqueux: It would not be better placed in `Source`?
Because `Layer` doesn't have a data format. It should no longer contain the convert/style options?

zarov: Yes indeed, I was misguided when thinking about something else here. Moving it to `Source`.

gchoqueux (approved):
Thanks

zarov:
Thanks !

PR #967, opened 2018-12-13
chore(examples): Simplify example page

gliegard:
Before this commit, the example was running in a tiny frame when we clicked on it.
Now, when we click on an example, it loads a new tab.


zarov (approved):
Thanks !

Issue #968, opened 2018-12-13
Cleaning and updating the documentation


zarov:
Now that the layout has been updated, it would be good to re-read, clean and correct the current documentation. It looks like a big task but I think it could be done quickly if everyone gives a hand 😉 So if you are updating/cleaning a file in this list, please let us know in this issue and open your PR when you are ready. The commit message should something like `docs: cleaning and updating __filename__ (part of #968)`

Here is the list (some files may already be okay):
- [ ] Base
  - [ ] Cache
  - [ ] Scheduler
  - [ ] TileMesh
- [ ] View
  - [ ] View
  - [ ] GlobeView
- [ ] Geographic
  - [ ] Coordinates
  - [ ] Extent
  - [ ] OrientationUtils
  - [ ] OBB
- [ ] Layer
  - [ ] Layer
  - [ ] ColorLayer
  - [ ] ElevationLayer
  - [ ] GeometryLayer
  - [ ] TiledGeometryLayer
  - [ ] InfoTiledGeometryLayer
  - [ ] ColorLayersOrdering
  - [ ] GlobeLayer
  - [ ] PanoramaLayer
  - [ ] PlanarLayer
- [x] Source
  - [x] Source
  - [x] WMTSSource
  - [x] WMSSource
  - [x] WFSSource
  - [x] TMSSource
  - [x] StaticSource
  - [x] FileSource
  - [x] URLBuilder
  - [x] Fetcher
- [ ] Parser
  - [ ] GeoJsonParser
  - [ ] GpxParser
  - [ ] VectorTileParser
  - [ ] CameraCalibrationParser
- [ ] Converter
  - [ ] Feature2Mesh
- [ ] Controls
  - [ ] GlobeControls
  - [ ] StateControl
  - [ ] PlanarControls
  - [ ] FirstPersonControls
  - [ ] OrientedImageCamera
  - [ ] CameraUtils
- [ ] Renderer
  - [ ] OrientedImageMaterial

Thanks for the help !

jailln:
Done while migrating to typescript, see #2396

PR #969, opened 2018-12-14
docs: clean and update of the Source/ folder

zarov:
Part of #968 

I also deprecated `heightMapWidth` in profit of `width` which is the real name of the parameter in the creation of as `WMS` url. I don't know if I could also add the `height` property ? So we could have something like:

```js
this.width = source.width || source.height || 256;
this.height = source.height || source.width || 256;
```


gchoqueux (approved):
thanks

PR #970, opened 2018-12-14
refactor(layer): replace all layer.type by layer.isLayer

zarov:
In 36c6d03 (PR #951), the isXXXLayer has been added, but wasn't used
everywhere. The change has been reflected in this commit to remove all
occurence of layer.type, leaving only the use for the debug layer (to be
removed).

DEPRECATION NOTICE: type is no longer a property of a Layer

### DIFF
--- a/test/examples/.eslintrc.js
+++ b/test/examples/.eslintrc.js
@@ -54,7 +54,8 @@ module.exports = {
 
         // turned off to use the this object in describe
         'prefer-arrow-callback': 'off',
-        
+
         'prefer-destructuring': 'off',
+        'max-len': 'off',
     }
 }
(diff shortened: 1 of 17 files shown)


test/examples/.eslintrc.js
```diff
@@ -54,7 +54,8 @@ module.exports = {
 
         // turned off to use the this object in describe
         'prefer-arrow-callback': 'off',
-        
+
         'prefer-destructuring': 'off',
+        'max-len': 'off',
```
gchoqueux, line 59: could you move it in new commit?

gchoqueux (approved):
thanks

test/examples/.eslintrc.js
zarov: Done !

PR #971, opened 2018-12-14
make DataSourceProvider more generic

gliegard:
## Description
make DataSourceProvider more generic:
- Add default convert function to Layer object
- In DataSourceProvider, give the requester in parameter to urlFromExtend method

## Motivation and Context
The goal of this PR is to simplify #939 , removing OrientedImageProvider  to use existing DataSourceProvider.

### DIFF
--- a/src/Layer/Layer.js
+++ b/src/Layer/Layer.js
@@ -141,6 +141,12 @@ class Layer extends THREE.EventDispatcher {
                 });
         }
     }
+
+    // Placeholder
+    // eslint-disable-next-line
+    convert(data) {
+        return data;
+    }
 }
 
 export default Layer;


src/Provider/DataSourceProvider.js
```diff
@@ -85,7 +85,7 @@ export default {
 
             // If source, we must fetch and convert data
             // URL of the resource you want to fetch
-            const url = source.urlFromExtent(extSource);
+            const url = source.urlFromExtent(extSource, requester);
```
zarov: Why adding this parameter ?

gliegard: In #939 , First I coded an OrientedImageProvider to load my textures. Then, I've managed to replace it using existing DataSourceProvider , to load my textures. As I load one texture for each camera, I used this parameter to give the list on camera , so my OrientedImageSource can genrate URLs , one for each camera. Doing that, I can load many textures using DataSourceProvider, without addind a new Provider in PR #939

zarov: Ok I think I understand. However this is too specific to #939 to be in this PR. I suggest to remove the second commit (containing this change) from this PR, and only add the `convert` bit.

Also please rename your commit with `fix(layer)` or something like that.

gliegard: ok, I've remove the commit

gchoqueux (approved):

Issue #959, opened 2018-12-07
Enable vendor-specific parameter(s) for WMS source


SesticM:
Many map servers provide their own, vendor-specific URL parameters when handling WMS calls. Such parameters are out of standard [Web Map Service](https://www.opengeospatial.org/standards/wms) specification, but contribute large amount during actual implementation. It would be of much help if `WMSSource` supported such capability via `vendorSpecific` property containing an array of name/value pairs (if any) being added to resulting GetMap URL call. Default value for `vendorSpecific` property would be, naturally, empty array. So, instead of writing a custom, JavaScript-based vendor-specific WMS source, an additional property to existing class would do the job.

Good example for this is [GeoServer](http://geoserver.org/)'s WMS service and related [vendor-specific parameters](https://docs.geoserver.org/latest/en/user/services/wms/vendor.html). GeoServer comes bundled with GeoWebCache, providing WMTS capability and caching out of the box. To be able to utilize GWC's GridSet functionality via "classic" WMS GetMap call, additional parameters (TILED, TILESORIGIN) are required. There are many other vendor-specific parameters available for GeoServer's WMS service.

zarov:
Good remark ! This seems simple to patch in the current `WMSSource`, if you want to do a PR about it. Meanwhile you can also write your own `WMSVendorSource` inheriting from `WMSSource` if you want to add only specific parameters.

I quickly read the geoserver documentation, I didn't see any dynamic parameter, can you confirm that ?

SesticM:
Please find below possible use-cases of utilizing GeoServer-specific WMS vendor parameters:

1) Single vendor-specific parameter:

```
var layer1 = new itowns.ColorLayer('dof_2012_02500_rs', {
	source: {
        	protocol: 'wms',
               	url: 'http://igeo.geoinova.com/geoserver/inova/wms',
               	name: 'inova:dof_2012_02500_rs',
                projection: 'EPSG:4326',
		transparent: true,
		format: 'image/png,
		vendorSpecific: [{'buffer':2}]
      	}
});
view.addLayer(layer1);
```

Produces:

```
http://igeo.geoinova.com/geoserver/inova/wms?SERVICE=WMS&REQUEST=GetMap&LAYERS=inova:dof_2012_02500_rs&VERSION=1.3.0&STYLES=&FORMAT=image/png&TRANSPARENT=true&BBOX=45.000000000,-135.000000000,90.000000000,-
[...]

PR #973, opened 2018-12-17
feat(wms): enable vendor specific parameters

zarov:
Many map servers provide their own, vendor-specific URL parameters when
handling WMS calls. Such parameters are out of standard Web Map Service
specification, but contribute large amount during actual implementation.
A good example of such parameters is in the GeoServer documentation:
https://docs.geoserver.org/latest/en/user/services/wms/vendor.html

Solves #959

PS: @SesticM I used your message for the commit message :wink:


gchoqueux (approved):
thanks

PR #976, opened 2018-12-19
chore: remove unless require

gchoqueux:


zarov (approved):
Please change the commit message from `unless` to `useless`.

PR #986, opened 2019-01-07
fix: windows install no more backslash

nosy-b:
## Description
Correct the bug while installing on windows due to inserted backslash on Patching source files

## Motivation and Context
It is necessary for win install in dev mode: npm install using itowns source code.


gchoqueux (approved):
Thanks

PR #988, opened 2019-01-07
chore: upgrade to THREE r100

gchoqueux:
### Migrate Guide r99 → r100

- [x] `Octree` has been removed.
- [x] Removed Geometry support from `Mesh.updateMorphTargets()`. Use `BufferGeometry` instead.
- [x] The default orientation of `RectAreaLight` has changed. It now looks along the negative z-axis.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9496,9 +9496,9 @@
       "dev": true
     },
     "three": {
-      "version": "0.99.0",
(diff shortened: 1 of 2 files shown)


gchoqueux:
I checked all THREE r100 change log and there's no change needed in itowns.
PR is ready.

nosy-b (approved):
No migration problem, let's go for r100!

PR #991, opened 2019-01-10
test: trap console errors when an example page is loaded

### LINKED ISSUE #929: Improve tests
## Context

Current issues with tests:

  * It can happen that a test is successful even if the tested example is not working (e.g. #926 ). 
  * @zarov [spotted some weird behaviour when running the tests](https://github.com/iTowns/itowns/issues/926#issuecomment-442048728). 
  * Currently, performing an operation (e.g. `console.log()`) before checking that layers are ready helps executing the test (see [this comment](https://github.com/iTowns/itowns/issues/926#issuecomment-442056291)).

To answer the first issue, we could improve the verification process in the tests:
  * In most case, we only verify that the added layers are ready. We may want to check something else to make sure everything is running smoothly.
  * We should assert in the test that there are no console errors in the tested example. Maybe by trapping the console errors in the test ? (if possible with puppeteer)

jailln:
## Description
Adds a verification to the example tests asserting that there are no errors when example pages are loaded.  
This PR Fixes #929 .


test/examples/bootstrap.js
```diff
@@ -5,6 +5,15 @@ const puppeteer = require('puppeteer');
 const net = require('net');
 const fs = require('fs');
 const http = require('http');
+// this line allows to disable a warning generated by node when more than 10
```
zarov, line 8: 10 or 100 ?

zarov (commented):
So if I understand correctly, any error will make the test to fail ?

Could you also move the fix for node 10+ in another commit ? Thanks !

jailln:
Yes any error should make the test fail.

test/examples/bootstrap.js
jailln: The default value set by node is 10, I set it to 100 to avoid the warning. I can precise the comment if you think its necessary.

zarov: Óh OK I understand, don't worry no problem !

jailln:
Commits splitted and branch rebased

zarov (approved):
Thanks for this !

PR #909, opened 2018-11-13
batchId option in Feature2Mesh

mbredif:
## Description
adds a batchId option to Feature2Mesh that creates a batchId attribute to be consumed by shaders.

## Motivation and Context
restore the functionnality. may be used for  feature selection, selective shading with the selecred batchId passed as a uniform...

PR #992, opened 2019-01-11
feat(converter): add batchId attribute in mesh from convert of feature2Mesh

gchoqueux:
Resumption of the work of @mbredif in #909

## Description
adds a `batchId`  attribute in `mesh.material` that creates a `batchId` attribute to be consumed by shaders.

## Motivation and Context
restore the functionnality. may be used for  feature selection, selective shading with the selected `batchId` passed as a uniform...

### DIFF
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -37,6 +37,12 @@ function fillColorArray(colors, length, color, offset = 0) {
     }
 }
 
+function fillBatchIdArray(batchId, batchIdArray, start, end) {
+    for (let i = start; i < end; i++) {
+        batchIdArray[i] = batchId;
+    }
+}
+
 /**
  * Convert coordinates to vertices positionned at a given altitude
  *
@@ -127,6 +133,9 @@ function featureToPoint(feature, options) {
     const vertices = new Float32Array(ptsIn.length);
     const colors = new Uint8Array(ptsIn.length);
 
+    const batchIds = options.batchId ?  new Uint32Array(ptsIn.length / 3) : undefined;
+    let featureId = 0;
+
     coordinatesToVertices(ptsIn, normals, vertices, options.altitude);
 
     for (const geometry of feature.geometry) {
@@ -134,11 +143,18 @@ function featureToPoint(feature, options) {
         const start = geometry.indices[0].offset;
         const count = geometry.indices[0].count;
         fillColorArray(colors, count, color, start);
+
+        if (batchIds) {
+            const id = options.batchId(geometry.properties, featureId);
+            fillBatchIdArray(id, batchIds, start, start + count);
+            featureId++;
+        }
     }
 
     const geom = new THREE.BufferGeometry();
     geom.addAttribute('position', new THREE.BufferAttribute(vertices, 3));
     geom.addAttribute('color', new THREE.BufferAttribute(colors, 3, true));
+    if (batchIds) { geom.addAttribute('batchId', new THREE.BufferAttribute(batchIds, 1)); }
 
     return new THREE.Points(geom, pointMaterial);
 }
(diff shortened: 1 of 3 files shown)


src/Converter/Feature2Mesh.js
```diff
@@ -399,7 +453,43 @@ export default {
      * @param {number|function} options.altitude - define the base altitude of the mesh
      * @param {number|function} options.extrude - if defined, polygons will be extruded by the specified amount
      * @param {object|function} options.color - define per feature color
+     * @param {function} [options.batchId] - optional function to create batchId attribute. It is passed the feature property and the feature index
```
zarov: Could you give an example of use of the generated batchId ?

src/Converter/Feature2Mesh.js
```diff
@@ -308,11 +353,20 @@ function featureToExtrudedPolygon(feature, options) {
...
+                batchIds[i] = id;
+            }
+            featureId++;
+        }
     }
 
     const geom = new THREE.BufferGeometry();
     geom.addAttribute('position', new THREE.BufferAttribute(vertices, 3));
     geom.addAttribute('color', new THREE.BufferAttribute(colors, 3, true));
+    if (batchIds) { geom.addAttribute('batchId', new THREE.BufferAttribute(batchIds, 1)); }
```
zarov, line 366: What is the point of adding them as an attribute ? AFAIK it is not used in a shader, so it could be stored in an array.

src/Converter/Feature2Mesh.js
```diff
@@ -268,6 +310,9 @@ function featureToExtrudedPolygon(feature, options) {
 
     vertices.minAltitude = Infinity;
 
+    const batchIds = options.batchId ?  new Uint32Array(vertices.length / 3) : undefined;
```
zarov, line 311: If we only allow `uint32` to be generated, it needs to be specified in the documentation.

zarov (changes requested):
Implementation is good, however I sense a lot of duplicate things: could you mutualize somehow in an external method this part:
```js
 if (batchIds) {
     const id = options.batchId(geometry.properties, featureId);
     for (let i = start; i < end; i++) {
         batchIds[i] = id;
     }
     featureId++;
 }
```

src/Converter/Feature2Mesh.js
gchoqueux: I added a example to pick building in wfs example. In another PR, I'll add a example to use `batchId` attribute with shader material.

src/Converter/Feature2Mesh.js
gchoqueux: see my [comment](https://github.com/iTowns/itowns/pull/992#discussion_r251505124)

src/Converter/Feature2Mesh.js
```diff
@@ -399,7 +450,43 @@ export default {
      * @param {number|function} options.altitude - define the base altitude of the mesh
      * @param {number|function} options.extrude - if defined, polygons will be extruded by the specified amount
      * @param {object|function} options.color - define per feature color
+     * @param {function} [options.batchId] - optional function to create batchId attribute. It is passed the feature property and the feature index. As the batchId is using an unsigned int structure on 32 bits, le batchId could be between 0 and 4,294,967,295.
```
zarov: "le batchId" -> "the batchId"

zarov (approved):
All good, please correct the french typo before merging

src/Converter/Feature2Mesh.js
gchoqueux: added

src/Converter/Feature2Mesh.js
gchoqueux: changed

PR #993, opened 2019-01-14
fix(gui): Fix GUI debug color layer and elevation layer

gliegard:
## Description
Fix GUI debug 
- color layer
- elevation layer


zarov (approved):
LGTM thanks

Issue #994, opened 2019-01-14
problem on start, the paths of GLTFLoader and GLTFLoaderLegacy  are unfound


Oncledjoe:
## Your Environment
* Version used: last (master)
* Browser Name and version: FireFox
* Operating System and version (desktop or mobile): Ubuntu 18.04LTS
* Link to your project: /

npm install works. 
I can't run itowns, errors occur after the command line npm start : 
ERROR in ./src/Parser/B3dmParser.js
Module not found: Error: Can't resolve 'ThreeExtended/loaders/GLTFLoader' in '/home/myusername/itowns/src/Parser'

ERROR in ./src/Parser/B3dmParser.js
Module not found: Error: Can't resolve 'ThreeExtended/loaders/deprecated/LegacyGLTFLoader' in '/home/myusername/itowns/src/Parser'

zarov:
It may be because you are using an older version of nodejs. Try with something at least 10+. 

The fact is that we don't store `GLTFLoader` anymore in our repository but we get it directly from THREE. The module that we use for this works only after node 10+ I think.

I'm not entirely sure about that, but @gchoqueux could confirm that.

Oncledjoe:
It work! thanks a lot.

aboutqx:
node 12+ the same problem， can't get threeExtended.

zarov:
Do you have any error message ? Also, which version of itowns/npm/node are you using precisely ?

aboutqx:
> ERROR in ./src/Parser/B3dmParser.js
Module Error (from ./node_modules/_eslint-loader@3.0.3@eslint-loader/dist/cjs.js):

> 
> D:\Github\itowns\src\Parser\B3dmParser.js
>   3:24  error  Unable to resolve path to module 'ThreeExtended/loaders/GLTFLoader'   import/no-unresolved
>   4:25  error  Unable to resolve path to module 'ThreeExtended/loaders/DRACOLoader'  import/no-unresolved
> 
> ✖ 2 problems (2 errors, 0 warnings)
> 
>  @ ./src/Main.js 70:0-86 70:0-86 70:0-86 70:0-86
>  @ ./src/MainBundle.js
>  @ multi @babel/polyfill url-polyfill whatwg-fetch ./src/MainBundle.js
> 
> ERROR in ./src/Renderer/c3DEngine.js
> Module Error (from ./node_modules/_eslint-loader@3.0.3@eslint-loader/dist/cjs.js):
> 
[...]

PR #996, opened 2019-01-15
Improved messages when installing packages

gchoqueux:
* Check PUPPETEER_SKIP_CHROMIUM_DOWNLOAD to warn user.
* Check npm version for information.
* Check node version to prevent error in THREE examples installation. (#994)

### DIFF
--- a/CODING.md
+++ b/CODING.md
@@ -5,6 +5,7 @@
 ### Node.js
 
 iTowns uses NPM as the build tool. If you haven't already, install Node.js: https://nodejs.org/
+You must have Node.js 10 or superior to develop in iTowns core.
 
 ## Building
 
--- a/config/prepare.js
+++ b/config/prepare.js
@@ -1,3 +1,25 @@
+const chalk = require('chalk');
+
+// Check node version to prevent error in THREE examples installation
+const nodeMajorVersion = process.versions.node.split('.')[0];
+const minNodeMajorVersion = 10;
+
+if (nodeMajorVersion < minNodeMajorVersion) {
+    console.log(chalk.red('Node.js version :', process.versions.node));
+    console.error(chalk.red(`Error:\tYour installed Node version is inferior to ${minNodeMajorVersion},
+\tyou must have Node.js ${minNodeMajorVersion} or superior to develop in iTowns core`));
+} else {
+    console.log(chalk.green('Node.js version :', process.versions.node));
+}
+
+const regex = new RegExp(/npm\/([^ ]+)/g);
+const npmVersion = regex.exec(process.env.npm_config_user_agent)[1];
+
+if (npmVersion) {
+    console.log(chalk.green('Npm version :', npmVersion), '\n');
+}
+
+// Copy and patch THREE Examples Files
 const copyfiles = require('copyfiles');
 const { filesExamples, patchedPath } = require('./threeExamples.js');
 
--- a/config/prepare.js
+++ b/config/prepare.js
@@ -1,5 +1,14 @@
 const chalk = require('chalk');
 
+// Verify Puppeteer configuration
+const pupSkip = process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD;
+
+if (!pupSkip) {
+    console.log(chalk.yellow(`Warning PUPPETEER_SKIP_CHROMIUM_DOWNLOAD is undefined,
+the installation'll be longer because Puppeteer'll download Chromium,
+only needed for testing. Read CODING.md for more information.\n`));
+}
+
 // Check node version to prevent error in THREE examples installation
 const nodeMajorVersion = process.versions.node.split('.')[0];
 const minNodeMajorVersion = 10;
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
     "start": "cross-env NODE_ENV=development webpack-dev-server -d --inline --hot",
     "prepublishOnly": "npm run build && npm run transpile",
+    "preinstall": "node ./config/preinstall.js",
```
tbroyer: Why is this done in `preinstall`? Shouldn't it be in `prepare`? (IIUC, `preinstall` will also run when installing iTowns as a dependency of the project you're currently developing in, not only when hacking on iTowns proper).

config/preinstall.js
```diff
@@ -0,0 +1,29 @@
...
+
+if (npmVersion) {
+    console.log(greenColor, 'Npm version :', npmVersion, '\n');
+}
+
+const pupSkip = process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD;
+
+if (!pupSkip) {
+    console.log(yellowColor, `Warning PUPPETEER_SKIP_CHROMIUM_DOWNLOAD is undefined,
+the installation'll be longer because Pupperter'll download Chromium,
```
tbroyer: typo: Puppeteer

config/preinstall.js
```diff
@@ -0,0 +1,29 @@
+const nodeMajorVersion = process.versions.node.split('.')[0];
+const minNodeMajorVersion = 10;
+const redColor = '\x1b[31m';
+const greenColor = '\x1b[32m';
+const yellowColor = '\x1b[33m';
+
+if (nodeMajorVersion < minNodeMajorVersion) {
+    console.log(redColor);
+    console.log('Node.js version :', process.versions.node);
+    console.error(`Error:\tYour installed Node version is inferior at ${minNodeMajorVersion},
```
tbroyer: nit: inferior *to*, or “less than” (sounds better IMO)

config/preinstall.js
```diff
@@ -0,0 +1,29 @@
...
+\tAs a minimum you must have Node.js ${minNodeMajorVersion} or superior to develop in iTowns core`);
```
tbroyer: nit: “*At* a minimum” (but maybe remove this, as this somehow duplicates the “or superior”)

package.json
gchoqueux: Yes, thanks, I didn't choose the correct script. By cons I would like to warn before the **`npm install`** (the download of Chromium by example) I should use **`prepublish`**.

tbroyer: From https://docs.npmjs.com/misc/scripts#prepublish-and-prepare, `prepare` is the same as `prepublish`, and should probably be preferred here. It will _also_ run before publishing, but :man_shrugging:

config/preinstall.js
gchoqueux: ok

config/preinstall.js
gchoqueux: ok

config/preinstall.js
gchoqueux: yes

package.json
gchoqueux: Ok, I moved the code in **`prepare`** script.

config/prepare.js
```diff
@@ -1,3 +1,39 @@
...
+}
+
+const regex = new RegExp(/npm\/([^ ]+)/g);
+const npmVersion = regex.exec(process.env.npm_config_user_agent)[1];
+
+if (npmVersion) {
+    console.log(greenColor, 'Npm version :', npmVersion, '\n');
+}
+
+console.log(whiteColor);
```
tbroyer: Did you actually meant to reset to the default? (what about someone using a white background?) ⇒ `\033[0m;`?

Fwiw, using chalk would remove that need, as it automatically resets after each message.

config/prepare.js
```diff
@@ -1,3 +1,39 @@
+const redColor = '\x1b[31m';
```
tbroyer: How about using [`chalk`](https://github.com/chalk/chalk)? (it's already a dependency of a bunch of tools, so it might not even add disk space) Added benefit: it automatically detects if colors are supported.

nit: I've seen `\033` used much more times than `\x1b`. YMMV.

CODING.md
```diff
@@ -5,6 +5,7 @@
 ### Node.js
 
 iTowns uses NPM as the build tool. If you haven't already, install Node.js: https://nodejs.org/
+As a minimum you must have Node.js 10 or superior to develop in iTowns.
```
tbroyer: Missed that one earlier: same "At a minimum" vs "or superior" (and "At" a minimum rather than "As")

gchoqueux: yes

config/prepare.js
gchoqueux: I spotted `chalk`, but I couldn't use it in the **`presinstall`** script because it was not installed yet. But it's different now. I'm going to change that.

config/prepare.js
gchoqueux: done, chalk used

tbroyer (approved):
Not tried it locally, but SGTM.

PR #998, opened 2019-01-17
Fix picking feature

gchoqueux:
## Description

1. Example used the removed properties : `Source.Protocol`. It should be verified that this property isn't longer used. see #1001
2. `filterFeaturesUnderCoordinate` didn't work with 3D coordinates

**TODO: verify why 3D coordinates are computed for features used in textures.
The expected behaviour is a feature with a 2D coordinates array**

### DIFF
--- a/test/examples/globe_vector.js
+++ b/test/examples/globe_vector.js
@@ -9,4 +9,24 @@ describe('globe_vector', function _() {
     it('should run', async () => {
         assert.ok(result);
     });
+
+    it('should pick feature from Layer with SourceFile', async () => {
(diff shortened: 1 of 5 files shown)


test/examples/globe_vector.js
```diff
@@ -9,4 +9,24 @@ describe('globe_vector', function _() {
     it('should run', async () => {
         assert.ok(result);
     });
+
+    it('Should pick feature from Layer with SourceFile', async () => {
```
zarov: It should be a lower case `s`, only thing I can say here

zarov (approved):
LGTM, correct the typo and you can merge

Issue #989, opened 2019-01-09
3DTiles bounding boxes debug tool erroneous display


EricBoix:
## Your Environment
 * Version used: V2.6.0 (git tag), [98abec9d](https://github.com/iTowns/itowns/commit/98abec9da9e7e5809c70b49ad2ffc35875c83989) (technically we use [this fork](https://github.com/jailln/itowns/tree/planar_3dtiles) as WIP) 
 * Browser Name and version: Firefox 64.0
 * Operating System and version (desktop or mobile): macOS 10.13.6
 * Link to your project: [this WIP fork](https://github.com/jailln/itowns/tree/planar_3dtiles) of the official iTowns

## Context
Gltf and 3D-Tiles made different choices for their referential systems ([gltf is Y-Up while 3DTiles is Z-up](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/280)). 
Previous versions of 3D-Tiles specifications allowed to deal with this discrepancy by enabling the usage of an [`asset.gltfUpAxis` attribute](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/175).
But the current version of 3D-Tiles specification (that [uses gltf version 2 as base for its B3dm's](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/TileFormats/Batched3DModel/README.md#binary-gltf)) dropped this `asset.gltfUpAxis` feature as mentionned by [ lilleyse comment dated May 10, 2018](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/280).
When working with source data that is inherently z-up, the current [3D-Tiles manner of dealing with the gltf/3DTiles axis discrepancy](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification#gltf-transforms) recommends the following common workflow:
```
   - Mesh data, including positions and normals, are not modified - they remain z-up.
   - The root node matrix specifies a column-major z-up to y-up transform. This transforms the source data into a y-up coordinate system as required by glTF.
   - At runtime the glTF is transformed back from y-up to z-up with the matrix above. Effectively the transforms cancel out.
```
where the considered `root node matrix` is the gltf one (as opposed to the [tile-set's root](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification#tileset) tile [transform matrix](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification#tile-transforms)).

[...]

PR #999, opened 2019-01-18
fix(debug):don't rotate bounding boxes for 3D Tiles v1.0 tilesets

jailln:
This PR fixes issue #989

### DIFF
--- a/utils/debug/3dTilesDebug.js
+++ b/utils/debug/3dTilesDebug.js
@@ -68,15 +68,19 @@ export default function create3dTilesDebugUI(datDebugTool, view, _3dTileslayer) 
                 }
 
                 if (helper && !metadata.boundingVolume.region) {
-                    // compensate B3dm orientation correction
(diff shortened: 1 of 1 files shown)


zarov:
Unfortunately, it doesn't fix the problem, as you can see:
image

Or maybe this reveals another bug ?

jailln:
My bad I forgot to test it with this tileset.

This tileset is in version 1.0 (where gltfUpAxis has been deprecated) and [3D Tiles is z-up](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification#coordinate-reference-system-crs) so my understanding is that its bounding boxes should also be z-up and hence should not be rotated (even if gltfUpAxis is undefined). However, I don't have time to investigate this more in depth currently so I'll close this PR and wait for 3D Tiles to be re-factored for 2.9.0 :)

zarov:
Ok, thanks anyway !

PR #1003, opened 2019-01-21
fix(core): wrong Z min and max  to compute tile's obb

### LINKED ISSUE #844: Wrong Tile OBB  with DTM
The size of tile OBB are wrong with DTM

bbox

## Expected Behavior
the box should be the size of tile

The min and max elevation aren't computed (with xbilParser). The calculation must be do for each elevation texture assigned to a tile

gchoqueux:
fix(core): wrong Z min and max  to compute tile's obb
because the offset/pitch was it was not taken into account

fix #844

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -339,7 +339,10 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
                     elevation.min = layer.colorTextureElevationMinZ;
                     elevation.max = layer.colorTextureElevationMaxZ;
                 } else {
-                    const { min, max } = computeMinMaxElevation(elevation.texture.image.data);
+                    const { min, max } = computeMinMaxElevation(elevation.texture.image.data,
+                        SIZE_TEXTURE_TILE,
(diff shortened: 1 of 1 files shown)


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -254,12 +254,16 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
...
-            node.setBBoxZ(min, max);
+        if (nodeLayer.textures[0]) {
+            if (!useMinMaxFromParent) {
+                const { min, max } =  computeMinMaxElevation(
+                    nodeLayer.textures[0].image.data,
+                    SIZE_TEXTURE_TILE, SIZE_TEXTURE_TILE,
+                    nodeLayer.offsetScales[0]);
+                node.setBBoxZ(min, max);
+            } else {
+                // TODO: to verify we don't pass here
```
zarov: TODO when ? Specify a release if its meant to be done later, and open an issue about it.

gchoqueux: i open #1011

zarov (approved):
Thanks !

PR #1004, opened 2019-01-22
Refacto camera: Update proporties only when it's needed

gchoqueux:
## Description
refactor(core): update view Matrix view only when it's needed
refactor(core): update preSSE view only when it's needed
refactor(core): update size view only when it's needed

### DIFF
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -14,33 +14,32 @@ function Camera(crs, width, height, options = {}) {
     this._viewMatrix = new THREE.Matrix4();
     this.width = width;
     this.height = height;
+    this.resize(width, height);
 
     this._preSSE = Infinity;
 }
 
-function resize(camera, width, height) {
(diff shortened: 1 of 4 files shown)


src/Renderer/Camera.js
```diff
@@ -87,9 +47,60 @@ Camera.prototype.update = function update(width, height) {
...
+            set: (newFov) => {
+                fov = newFov;
+                updatePreSse(this, this.height, fov);
+            },
+        });
+    }
+}
+
+Camera.prototype.resize = function resize(width, height) {
+    this.width = width;
```
zarov, line 79: This would duplicate things, but maybe you could also declare those variables in the constructor too ?

zarov (approved):
Seems good to me, I don't see any breaking changes here.

src/Renderer/Camera.js
gchoqueux: ok

PR #1005, opened 2019-01-22
fix(core): avoid to access to DOM properties

### LINKED ISSUE #140: Store DOM properties instead of access them every time in mouse events
For performance reason, and [avoid reflow/repaint](https://gist.github.com/paulirish/5d52fb081b3570c81e3a), it's better to store DOM properties instead of accessing them every time.

For example, in `Renderer/ThreeExtended/GlobeControls.js`, many access to `elem.offsetTop`/`elem.offsetLeft`/`elem.clientWidth`/`elem.clientHeight`, where `elem` is the canvas element (to confirm).

gchoqueux:
fix #140


zarov (approved):
Thanks !

PR #1006, opened 2019-01-23
Add function removeLayer for color, elevation and geometry layers

### LINKED ISSUE #906: Proposition : function removeLayer() for geometric layer
For now it is only possible to remove a 'color' layer. Could be nice to do the same with a geometric (geojson) layer.

gchoqueux:
## Description
New function `removeLayer` to remove color, elevation and geometry layers from viewer

Add tests on `removeLayer`.

fix #906

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -43,6 +43,7 @@ export const VIEW_EVENTS = {
      * @property type {string} layers-initialized
      */
     LAYERS_INITIALIZED: 'layers-initialized',
+    LAYER_REMOVED: 'layer-removed',
 };
 
 /**
@@ -290,6 +291,51 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
             resolve(layer);
         });
     });
+};
+
+/**
+ * Removes a specific imagery layer from the current layer list. This removes layers inserted with attach().
+ * @example
+ * view.removeLayer('layerId');
+ * @param      {string}   layerId      The identifier
+ * @return     {boolean}
+ */
+View.prototype.removeLayer = function removeLayer(layerId) {
+    const layer = this.getLayerById(layerId);
+    if (layer) {
+        const parentLayer = layer.parent;
+
+        // Remove and dispose all nodes
+        layer.delete();
+
+        // Detach layer if it's attached
+        if (parentLayer && !parentLayer.detach(layer)) {
+            throw new Error(`Error to detach ${layerId} from ${parentLayer.id}`);
+        } else if (parentLayer == undefined) {
+            // Remove layer from viewer
+            this._layers.splice(this._layers.findIndex(l => l.id == layerId), 1);
+        }
+        if (layer.isColorLayer) {
+            // Update color layers sequence
+            const imageryLayers = this.getLayers(l => l.isColorLayer);
+            for (const color of imageryLayers) {
+                if (color.sequence > layer.sequence) {
+                    color.sequence--;
+                }
+            }
+        }
+
+        this.notifyChange(parentLayer || this.camera.camera3D, true);
+
+        this.dispatchEvent({
+            type: VIEW_EVENTS.LAYER_REMOVED,
+            layerId,
+        });
+
+        return true;
+    } else {
+        throw new Error(`${layerId} doesn't exist`);
+    }
 };
 
 /**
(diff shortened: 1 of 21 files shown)


src/Core/View.js
```diff
@@ -292,6 +295,82 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
...
+
+/**
+ * Removes a specific imagery layer from the current layer list. This removes layers inserted with attach().
+ * @example
+ * view.removeLayer('layerId');
+ * @param      {string}   layerId      The identifier
+ * @return     {boolean}
+ */
+View.prototype.removeLayer = function removeLayer(layerId) {
+    const layer = this.getLayers(l => l.id === layerId)[0];
```
zarov: General comment: we need a method to access directly a layer from a view, I can't look anymore at this line

src/Core/View.js
```diff
@@ -292,6 +295,82 @@ View.prototype.addLayer = function addLayer(layer, parentLayer) {
     });
 };
 
+function disposeMesh(obj) {
```
zarov: This should be in `GeometryLayer`, WDYT ?

src/Core/View.js
gchoqueux: yes, completely agree

src/Core/View.js
gchoqueux: yes, I will move it.

gchoqueux: In thinking about it, I think that we should do the removing/disposing nodes in the `detach` method of the layer. I'll soon make a new proposal.

src/Core/View.js
gchoqueux: function `View.getLayerById` added

src/Core/View.js
gchoqueux: A new proposal is pushed.

src/Core/View.js
```diff
@@ -340,6 +386,21 @@ View.prototype.getLayers = function getLayers(filter) {
...
+/**
+ * Gets the layer by identifier.
+ *
+ * @param {String}  layerId  The layer identifier
+ * @return {Layer}  The layer by identifier.
+ */
+
+View.prototype.getLayerById = function getLayers(layerId) {
+    const layers = this.getLayers(l => l.id === layerId);
+    if (layers.length === 1) {
```
zarov: Unnecessary ? It should be impossible to have two layers with the same id.

src/Layer/ColorLayer.js
```diff
@@ -73,6 +73,15 @@ class ColorLayer extends Layer {
     convert(data, extentDestination) {
         return textureConverter.convert(data, extentDestination, this);
     }
+
+    /**
+    * All layer's textures are removed from scene and disposed from video device.
+    */
+    removeDisposeNodes() {
```
zarov: Maybe add a "fake" method in `Layer` as well: either doing nothing, or printing a warning that this method needs to be implemented (like in `Source`).

src/Core/View.js
```diff
@@ -340,6 +386,21 @@ View.prototype.getLayers = function getLayers(filter) {
...
 };
 
+/**
+ * Gets the layer by identifier.
+ *
+ * @param {String}  layerId  The layer identifier
+ * @return {Layer}  The layer by identifier.
+ */
+
+View.prototype.getLayerById = function getLayers(layerId) {
```
zarov: `function getLayerById` instead.

Also we need to change (in next PR) everywhere this method could be used.

src/Layer/ColorLayer.js
zarov: Also this seems like a weird name to understand

src/Core/View.js
gchoqueux: done

src/Core/View.js
gchoqueux: done

src/Layer/ColorLayer.js
gchoqueux: done

gchoqueux: I rename to `Layer.delete`

gchoqueux:
@zarov  you could review update.
 I need to implement delete function for `OrientedImageLayer`.

zarov (approved):
LGTM

PR #1007, opened 2019-01-23
fix(example): Improve performance on immersive example.

gliegard:
### Description
fix(example): Improve performance on immersive example.

### Context
Before this commit, when user was doing mouse drag, there was lag.
Now, it's very smooth.
To do that, I've reconfigured Wfs building source to request it on a the good zoom level.


gchoqueux (approved):
Nice

PR #1013, opened 2019-01-25
Controls add key to previous pano

gliegard:
## Description
Immersive: Add the possibility to go backward with key S

- Street controls : Add key binding to previsous pano
- Oriented Image Layer : 
  - Add the previsous pano position in the callback
  - Also, manage case wise no background object
- Immersive example updated

### DIFF
--- a/src/Controls/StreetControls.js
+++ b/src/Controls/StreetControls.js
@@ -20,7 +20,12 @@ function moveCameraExp(root, progress) {
  * </ul>
  * <ul> Bindings added
  * <li><b> keys Z : </b> Move camera to the next position </li>
+ * <li><b> keys S : </b> Move camera to the previous position </li>
+ * <li><b> keys A : </b> Set camera to current position </li>
  * </ul>
+ * @property {number} keyGoToNextPosition key code to go to next position, default to 90 for key Z
+ * @property {number} keyGoToPreviousPosition key code to go to previous position, default to 83 for key S
+ * @property {number} keySetCameraToCurrentPosition key code set camera to current position, default to 65 for key  A
  * @extends FirstPersonControls
  */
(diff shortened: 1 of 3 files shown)


zarov (approved):
Nice thanks, however I would like to discuss one thing before merging: the choice of the keys. In your case, you have a AZERTY keyboard, so it makes sense to have `Z`, `A` etc... as keys to move inside the view. But with another keyboard, it is quite weird to use those keys. That's why I'm proposing more logical keys to use:
- `P` for previous
- `N` for next
- `R` (for reset) or `C` (for current) for the current position

WDYT ?

gliegard:
I can add a setter with key code to let the user remap the keys ?
like
setGoForwardKeyCode(keyCode)
set GoBackwardKeyCode(keyCode)

WDYT ?

zarov:
Sounds good to me, thanks.

gliegard:
ok, updated :)

src/Controls/StreetControls.js
```diff
@@ -51,8 +51,16 @@ class StreetControls extends FirstPersonControls {
         this.player.addEventListener('animation-frame', this.updateView.bind(this));
 
         // twos positions used by this control : current and next
+        this.previousPosition = undefined;
         this.currentPosition = undefined;
         this.nextPosition = undefined;
+
+        // store key binding to go to next position, default to Z
```
zarov: Please move those comments before the class declaration, as `@property` of the class.

gliegard: ok, done

zarov: Thanks, merged

gliegard: Yeahh , thanks :) :)

PR #1016, opened 2019-01-28
chore: update and rewrite contents in README and other .md files.

zarov:
Following some incredible motivation after #1014, I updated almost all .md files.


gchoqueux (approved):
thanks for this update

PR #1017, opened 2019-01-29
Improve Street controls 1

gliegard:
## Description
improve StreetControls : add click and go, surface helpers, click and look at, tweens animations :
- Mouse click on ground detection, and associated callback, that go to the clicked position
- Mouse click on other object (not ground), and associated callback, that makes camera look at the position
- Surfaces as helpers for the end user :
  - Circle displayed when user moves the mouse on the ground
  - Rectangle displayed when user moves the mouse on other objects (like walls of a building)
- Tweens animations :
  - camera can move to another position smoothly
  - camera can look at a position smoothly in an animation

## Motivation and Context
User should have a good experience using immersive example

helper_wall
helper_ground

### DIFF
--- a/examples/immersive_paris.html
+++ b/examples/immersive_paris.html
@@ -45,7 +45,7 @@
             });
 
             // limit camera far, to increase performance
-            view.camera.camera3D.far = 100;
+            view.camera.camera3D.far = 10000;
             view.camera.camera3D.near = 0.1;
 
             // open camera fov
@@ -107,7 +107,7 @@
                 // Create oriented image layer
                 var olayer = new itowns.OrientedImageLayer('demo_orientedImage', {
                     // Radius in meter of the sphere used as a background.
-                    backgroundDistance: 80,
+                    backgroundDistance: 1200,
                     source: orientedImageSource,
                     orientation: orientation,
                     calibration: calibration,
@@ -140,7 +140,7 @@
                     });
 
                     // create geometry layer for the buildings
-                    var wfsBuildingLayer = new itowns.GeometryLayer('WFS Building', new itowns.THREE.Group(), {
+                    var wfsBuildingLayer = new itowns.GeometryLayer('Buildings', new itowns.THREE.Group(), {
                         update: itowns.FeatureProcessing.update,
                         convert: itowns.Feature2Mesh.convert({
                             altitude: altitudeBuildings,
@@ -152,8 +152,11 @@
                         overrideAltitudeInToZero: true,
                     });
 
-                    // add the created building layer
-                    view.addLayer(wfsBuildingLayer);
+                    // add the created building layer, and debug UI
+                    view.addLayer(wfsBuildingLayer).then(function addDebugUI(buildingLayer) {
+                        var gui = debug.GeometryDebug.createGeometryDebugUI(menuGlobe.gui, view, buildingLayer);
+                        debug.GeometryDebug.addWireFrameCheckbox(gui, view, buildingLayer);
+                    });
                 });
 
             });
@@ -161,6 +164,9 @@
             /* global itowns, document, GuiTools, view, promises */
             var menuGlobe = new GuiTools('menuDiv');
             menuGlobe.view = view;
+
+            var d = new debug.Debug(view, menuGlobe.gui);
+            debug.createTileDebugUI(menuGlobe.gui, view, view.tileLayer, d);
 
             // Listen for globe full initialisation event
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
@@ -171,6 +177,7 @@
                 // set camera to current panoramic
                 view.controls.setCameraToCurrentPosition();
                 view.notifyChange(view.camera.camera3D);
+
             });
 
         </script>
--- a/src/Controls/StreetControls.js
+++ b/src/Controls/StreetControls.js
@@ -1,17 +1,84 @@
+import TWEEN from '@tweenjs/tween.js';
 import * as THREE from 'three';
+import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
 import FirstPersonControls from 'Controls/FirstPersonControls';
-import AnimationPlayer, { AnimatedExpression } from 'Core/AnimationPlayer';
 
-// Expression used to damp camera's moves
-function moveCameraExp(root, progress) {
-    const dampingProgress = 1 - Math.pow((1 - (Math.sin((progress - 0.5) * Math.PI) * 0.5 + 0.5)), 2);
-    root.camera.position.lerpVectors(root.positionFrom, root.positionTo, dampingProgress);
+const material = new THREE.MeshBasicMaterial({
+    color: 0xffffff,
+    depthTest: false,
+    transparent: true,
+    opacity: 0.5 });
+
+function createCircle() {
+    const geomCircle = new THREE.CircleGeometry(1, 32);
+    return new THREE.Mesh(geomCircle, material);
 }
+
+function createRectangle() {
+    const geomPlane = new THREE.PlaneGeometry(4, 2, 1);
+    const rectangle = new THREE.Mesh(geomPlane, material);
+    rectangle.rotateX(-Math.PI / 2);
+    return rectangle;
+}
+
+// update a surfaces node
+function updateSurfaces(surfaces, position, norm) {
+    surfaces.position.copy(position);
+    surfaces.up.copy(position).normalize();
+    surfaces.lookAt(norm);
+    surfaces.updateMatrixWorld(true);
+}
+
+// vector use in the pick method
+const target = new THREE.Vector3();
+
+function pick(event, view, buildingsLayer, pickGround = () => {}, pickObject = () => {}, pickNothing = () => {}) {
+    // get real distance to ground, with a specific method to pick on the elevation layer
+    view.getPickingPositionFromDepth(view.eventToViewCoords(event), target);
+    const distanceToGround = view.camera.camera3D.position.distanceTo(target);
+
+    // pick on building layer
+    const buildings = view.pickObjectsAt(event, 0, buildingsLayer);
+
+    // to detect pick on building, compare first picked building distance to ground distance
+    if (buildings.length && buildings[0].distance < distanceToGround) { // pick buildings
+        // callback
+        pickObject(buildings[0].point, buildings[0].face.normal);
+    } else {
+        // pick on ground layer (tile layer)
+        const ground = view.pickObjectsAt(event, 0, view.tileLayer);
+        if (ground.length) {
+            // compute normal
+            const normal = target.clone().multiplyScalar(1.1);
+
+            // callback
+            pickGround(target, normal);
+        } else {
+            // callback
+            pickNothing();
+        }
+    }
+}
+
+// default function to compute time (in millis), used for the animation to move to a distance (in meter)
+function computeTime(distance) {
+    return 500 + Math.sqrt(distance) * 300;
+}
+
 /**
  * @classdesc Camera controls that can follow a path.
  * It is used to simulate a street view.
  * It stores a currentPosition and nextPosition, and do a camera traveling to go to next position.
- *
+ * It also manages picking on the ground and on other object, like building.
+ * <ul> It manages 2 surfaces, used as helpers for the end user :
+ * <li> a circle is shown when mouse is moving on the ground </li>
+ * <li> a rectangle is shown when mouse is moving on other 3d object </li>
+ * </ul>
+ * <ul>
+ * This controls is designed
+ * <li> to move forward when user click on the ground (click and go) </li>
+ * <li> to rotate the camera when user click on other object (click to look at) </li>
+ * </ul>
  * <ul> Bindings inherited from FirstPersonControls
  * <li><b> up + down keys : </b> forward/backward </li>
  * <li><b> left + right keys: </b> strafing movements </li>
@@ -23,6 +90,7 @@ function moveCameraExp(root, progress) {
  * <li><b> keys S : </b> Move camera to the previous position </li>
  * <li><b> keys A : </b> Set camera to current position </li>
  * </ul>
+ * Note that it only works in globe view.
  * @property {number} keyGoToNextPosition key code to go to next position, default to 90 for key Z
  * @property {number} keyGoToPreviousPosition key code to go to previous position, default to 83 for key S
  * @property {number} keySetCameraToCurrentPosition key code set camera to current position, default to 65 for key  A
@@ -33,27 +101,21 @@ class StreetControls extends FirstPersonControls {
      * @constructor
      * @param { View } view - View where this control will be used
      * @param { Object } options - Configuration of this controls
-     * @param { number } options.animationDuration - Duration of the animation to another panoramic.
+     * @param { number } [options.wallMaxDistance=1000] - Maximum distance to click on a wall, in meter.
+     * @param { number } [options.animationDurationWall=1000] - Time in millis for the animation when clicking on a wall.
+     * @param { THREE.Mesh } [options.surfaceGround] - Surface helper to see when mouse is on the ground, default is a transparent circle.
+     * @param { THREE.Mesh } [options.surfaceWall] - Surface helper to see when mouse is on a wall, default is a transparent rectangle.
+     * @param { string } [options.buildingsLayer='Buildings'] - Name of the building layer (used to pick on wall).
+     * @param { function } [options.computeTime] - Function to compute time (in millis), used for the animation to move to a distance (in meter)
+     * @param { number } [options.offset=4] - Altitude in meter up to the ground to move to when click on a target on the ground.
      */
     constructor(view, options = {}) {
         super(view, options);
 
         this.isStreetControls = true;
 
-        // store layers
-        this.layers = [];
-        this.currentLayerIndex = 0;
-
-        // manage camera movements
-        this.player = new AnimationPlayer();
-        this.positionFrom = new THREE.Vector3();
-        this.animationMoveCamera = new AnimatedExpression({
-            duration: options.animationDuration || 50,
-            root: this,
-            expression: moveCameraExp,
-            name: 'Move camera',
-        });
-        this.player.addEventListener('animation-frame', this.updateView.bind(this));
+        const domElement = view.mainLoop.gfxEngine.renderer.domElement;
+        domElement.addEventListener('mouseout', super.onMouseUp.bind(this));
 
         // twos positions used by this control : current and next
         this.previousPosition = undefined;
@@ -63,6 +125,25 @@ class StreetControls extends FirstPersonControls {
         this.keyGoToNextPosition = 90;
         this.keyGoToPreviousPosition = 83;
         this.keySetCameraToCurrentPosition = 65;
+
+        // Tween is used to make smooth animations
+        this.tweenGroup = new TWEEN.Group();
+
+        // init surfaces used as helper for end user
+        this.surfaceGround = options.surfaceGround || createCircle();
+        this.surfaceWall = options.surfaceWall || createRectangle();
+
+        // surfaces is an object3D containing the two surfaces
+        this.surfaces = new THREE.Object3D();
+        this.surfaces.add(this.surfaceGround);
+        this.surfaces.add(this.surfaceWall);
+        this.view.scene.add(this.surfaces);
+
+        this.wallMaxDistance = options.animationDurationWall || 1000;
+        this.animationDurationWall = options.animationDurationWall || 1000;
+        this.buildingsLayer = options.buildingsLayer || 'Buildings';
+        this.computeTime = options.computeTime || computeTime;
+        this.offset = options.offset || 4;
     }
 
     setCurrentPosition(newCurrentPosition) {
@@ -77,8 +158,39 @@ class StreetControls extends FirstPersonControls {
         this.previousPosition = newPreviousPosition;
     }
 
+    onMouseUp(event) {
+        super.onMouseUp();
+        if (this._stateOnMouseDrag) {
+            this._stateOnMouseDrag = false;
+        } else {
+            pick(event, this.view, this.buildingsLayer, this.onClickOnGround.bind(this), this.onClickOnWall.bind(this));
+        }
+    }
+
+    onMouseMove(event) {
+        super.onMouseMove(event);
+
+        if (this._isMouseDown) {
+            // state mouse drag (move + mouse click)
+            this._stateOnMouseDrag = true;
+            this.stopAnimations();
+        }
+        // mouse pick and manage surfaces
+        pick(event, this.view, this.buildingsLayer,
+            (groundTarget, normal) => {
+                updateSurfaces(this.surfaces, groundTarget, normal);
+                this.surfaceGround.visible = true;
+                this.surfaceWall.visible = false;
+            }, (wallTarget, normal) => {
+                updateSurfaces(this.surfaces, wallTarget, normal);
+                this.surfaceWall.visible = true;
+                this.surfaceGround.visible = false;
+            });
+        this.view.notifyChange(this.surfaces);
+    }
+
     /**
-     * Sets the camera to the current position, looking at the next position.
+     * Sets the camera to the current position (stored in this controls), looking at the next position (also stored in this controls).
      */
     setCameraToCurrentPosition() {
         this.setCameraOnPosition(this.currentPosition, this.nextPosition);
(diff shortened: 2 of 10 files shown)


src/Core/TileMesh.js
```diff
@@ -34,6 +34,7 @@ class TileMesh extends THREE.Mesh {
         this.rotationAutoUpdate = false;
 
         this.layerUpdateState = {};
+        this.isTileMesh = true;
```
gchoqueux, line 37: Could you replace the condition `(node instanceof TileMesh)` by (this.isTileMesh)

src/Controls/StreetControls.js
```diff
@@ -1,17 +1,23 @@
+/* globals window */
```
zarov: You can remove it

src/Controls/StreetControls.js
```diff
@@ -1,17 +1,23 @@
...
-function moveCameraExp(root, progress) {
-    const dampingProgress = 1 - Math.pow((1 - (Math.sin((progress - 0.5) * Math.PI) * 0.5 + 0.5)), 2);
-    root.camera.position.lerpVectors(root.positionFrom, root.positionTo, dampingProgress);
-}
 /**
  * @classdesc Camera controls that can follow a path.
  * It is used to simulate a street view.
  * It stores a currentPosition and nextPosition, and do a camera traveling to go to next position.
- *
+ * It also manages picking on the ground and on other object, like building.
```
zarov, line 72: :+1:

src/Controls/StreetControls.js
```diff
@@ -1,17 +1,23 @@
+/* globals window */
+import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
```
zarov: Can you move this import after three ?

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+            i++;
+        }
+
+        // return first interesting picked object, or undefined
+        if (i < intersects.length) {
+            return intersects[i];
+        }
+    }
+
+    pick(event, pickGround = () => {}, pickObject =  () => {}) {
```
zarov: Mind your spaces

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+            depthTest: false,
+            transparent: true,
+            opacity: 0.5 });
+
+        const geomPlane = new THREE.PlaneGeometry(4, 2, 1);
+        this._rectangleMesh = new THREE.Mesh(geomPlane, material);
+        const geomCircle = new THREE.CircleGeometry(1, 32);
+        this._circleMesh = new THREE.Mesh(geomCircle, material);
+
+        this.view.scene.add(this._rectangleMesh);
```
zarov: Why adding it if you add it below in `this.surfacesNode`, which is added to the scene ?

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
         this.previousPosition = newPreviousPosition;
     }
 
+    firstPickedObject(event) {
+        // basic threeJs pick with raycaster
+        const normalized = this.view.viewToNormalizedCoords(event);
+        this.raycaster.setFromCamera(normalized.clone(), this.camera);
```
gchoqueux: Could you use picking from itowns?

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+        const distance = this.camera.position.distanceTo(position);
+        // 500 millis constant, plus an amount of time depending of the distance (but not linearly)
+        const time = 500 + Math.sqrt(distance) * 300;
+
+        // move the camera
+        this.moveCameraTo(position, time);
+    }
+
+    /**
+     * Method called when user click on oject that is not the ground.</br>
```
zarov, line 237: Pro-tip: you can avoid using `<br/>` by adding another breakline, thus also making the code more readable.

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+     * Note that this function contains default values that can be overrided, by overriding this function.
+     *
+     * @param {THREE.Vector3} position - The position
+     */
+    onClickOnWall(position) {
+        const distance = this.camera.position.distanceTo(position);
+
+        // can't click on a wall that is at 1km distance.
+        if (distance < 1000) {
```
zarov: This value is very arbitrary, could you set it in a variable in the constructor ?

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+     * @param { number } time - Time in millisecond
+     */
+    animateCameraLookAt(position, time) {
+        // stop existing animation
+        this.stopTweensAnimations();
+        // prepare start point and end point
+        this.start = this.camera.clone();
+        this.end = this.camera.clone();
+        this.end.lookAt(position);
+        const factor = { t: 0 };
```
zarov: No need for a `const`

src/Controls/StreetControls.js
```diff
@@ -63,6 +57,14 @@ class StreetControls extends FirstPersonControls {
...
         this.keyGoToPreviousPosition = 83;
         this.keySetCameraToCurrentPosition = 65;
+
+        // Raycaster used to pick objects
+        this.raycaster = new THREE.Raycaster();
+        // Tween is used to make smooth animations
+        this.tweenGroup = new TWEEN.Group();
+
+        // init surfaces used as helper for end user
+        this.initSurfaces();
```
gchoqueux: if `initSurfaces` is used to instantiate the helpers, you could make a private fonction, like this:
```js
initSurfacesHelper(this);
```

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+        // basic threeJs pick with raycaster
+        const normalized = this.view.viewToNormalizedCoords(event);
+        this.raycaster.setFromCamera(normalized.clone(), this.camera);
+        const intersects = this.raycaster.intersectObjects(this.view.scene.children, true);
+
+        // skipping specific objects:
+        // - skip the two surfaces (rectangle and circle)
+        // - skip all the object without face, because we use the face to manage surface angle, so it's needed.
+        let i = 0;
+        while (i < intersects.length &&
```
gchoqueux: With picking itowns, you can filter by layer. this should avoid this loop.

src/Controls/StreetControls.js
```diff
@@ -63,6 +57,14 @@ class StreetControls extends FirstPersonControls {
         this.keyGoToNextPosition = 90;
         this.keyGoToPreviousPosition = 83;
         this.keySetCameraToCurrentPosition = 65;
+
+        // Raycaster used to pick objects
+        this.raycaster = new THREE.Raycaster();
+        // Tween is used to make smooth animations
+        this.tweenGroup = new TWEEN.Group();
```
zarov, line 130: Declare `this.tweens` (which should be `this.tween`) here. Side note: why do you need a group and a tween ? (I know nothing about this library)

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+        // return first interesting picked object, or undefined
+        if (i < intersects.length) {
+            return intersects[i];
+        }
+    }
+
+    pick(event, pickGround = () => {}, pickObject =  () => {}) {
+        // specific method to pick on the elevation layer at the good elevation, even if elevation layer is done in the shader
+        const mouse = this.view.eventToViewCoords(event);
+        const target = new THREE.Vector3();
```
gchoqueux: could you move out `target` from this function .

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+                // callback
+                pickGround(target);
+            }
+        }
+    }
+
+    onClick() {
+        // here we detect mouse click when mouse is not moving,
+        // to avoid interfering with mouse drag movement, managed by FirstPersonControls
+        this.maybeClickWithoutMoving = true;
```
gchoqueux: may be use state machine would be more clear. WDYT?

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
         this.previousPosition = newPreviousPosition;
     }
 
+    firstPickedObject(event) {
+        // basic threeJs pick with raycaster
+        const normalized = this.view.viewToNormalizedCoords(event);
+        this.raycaster.setFromCamera(normalized.clone(), this.camera);
+        const intersects = this.raycaster.intersectObjects(this.view.scene.children, true);
```
zarov: This will be very very heavy if you are inside a building with a complex mesh: we need a better solution. Now, I'm not going to block this PR for this, but I would appreciate if you already have some insights on how to fix this. 

Anyway, please add a warning concerning this problem for now.

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+        this.surfacesNode = new THREE.Object3D();
+        this.view.scene.add(this.surfacesNode);
+
+        this.surfacesNode.add(this._rectangleMesh);
+        this.surfacesNode.add(this._circleMesh);
+    }
+
+    // hide both rectangle and circle helpers
+    hideSurfaces() {
+        this._circleMesh.visible = false;
```
gchoqueux: Could you use a `THREE.Group` to manage the `surfaceHelpers`. You could write only `surfaceHelpers.visible = false`.

src/Controls/StreetControls.js
zarov: This seems fishy: can you explain the interference you got ?

src/Controls/StreetControls.js
```diff
@@ -77,8 +79,143 @@ class StreetControls extends FirstPersonControls {
...
+    initSurfaces() {
+        const material = new THREE.MeshBasicMaterial({
+            side: THREE.DoubleSide,
+            color: 0xffffff,
+            depthTest: false,
+            transparent: true,
+            opacity: 0.5 });
+
+        const geomPlane = new THREE.PlaneGeometry(4, 2, 1);
+        this._rectangleMesh = new THREE.Mesh(geomPlane, material);
```
zarov: It would be nice to be able to customise those meshes

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
     }
 
     /**
-     * Move the camera
+     * Method called when user click on the ground.</br>
+     * Note that this funtion contains default values that can be overrided, by overriding this function.
+     *
+     * @param {THREE.Vector3} position - The position
+     */
+    onClickOnGround(position) {
```
gchoqueux, line 222: these functions don't need to be exposed.

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+        // compute a new position to go : 4 meter up from the picked position on the ground
+        const up = position.clone().normalize();
```
gchoqueux, line 224: does it work with the planar?

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+        position.add(up.multiplyScalar(4));
```
gchoqueux: it's necessary to multiply by 4?

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+        }
+    }
+
+    /**
+     * Animate the camera to make it look at a position, in a given time
+     *
+     * @param { THREE.Vector3 }  position - Position to look at
+     * @param { number } time - Time in millisecond
+     */
+    animateCameraLookAt(position, time) {
```
gchoqueux, line 257: it would be interesting to have `animateCameraLookAt` and `moveCameraTo` in `CameraUtils`.

src/Controls/StreetControls.js
```diff
@@ -102,19 +239,123 @@ class StreetControls extends FirstPersonControls {
...
+            this.tweens.stop();
+            this.tweens = undefined;
+        }
+    }
+
+    // stop all animations
+    onMouseOut() {
+        this.maybeClickWithoutMoving = false;
+        this.onMouseUp();
+        this.stopTweensAnimations();
```
gchoqueux: why you stop the animation.

gchoqueux (changes requested):
Nice improvement

src/Core/TileMesh.js
zarov, line 37: Do the appropriate change in `convertToTile` and `Picking` please.

zarov (changes requested):
Could you add an explanation in the message of your `fix` commit ?

Anyway, very happy to see something like this, though it needs a bit more work. Thanks a lot !

src/Core/TileMesh.js
gliegard: ok

src/Controls/StreetControls.js
gliegard: No, because existing picking methods are picking in a specific layer, and I want this StreetControls to be generic : it doesn't know the other layers, it picks all the scene, and will work with every layers (3dTiles, WFS buildings...).

src/Controls/StreetControls.js
gliegard: yes good idea

src/Controls/StreetControls.js
gliegard: I don't manage to make it work with existing picking method from iTowns, because I need to pick the TileMesh object, to get the face and the normal. So I prefer using threeJs picking.

src/Controls/StreetControls.js
gliegard: ok

src/Controls/StreetControls.js
gliegard: yep, I replace it with a boolean '_stateOnMouseDrag' , it's clearer.

src/Controls/StreetControls.js
gliegard: No because, on the other hand, when I will want to make only one surface visible, I'll have to make the Group visible again, so this idea remove on line of code somewhere, but make me add another line at other place. So I think it's not a good idea.

src/Controls/StreetControls.js
gliegard: As I write it on the doc 

"Note that this funtion contains default values that can be overrided, by overriding this function."

That means these methods are the default behavior , and it can be overrided to do specific behavior. So it's designed to be exposed.

src/Controls/StreetControls.js
gliegard: Not yet, but I plan to add another example in planar. At this moment I will make all this work in planar view.

src/Controls/StreetControls.js
gliegard: It's a default value for the given example. The goal is to move 4 meter up to the ground. It can be overrided, by overriding the method.

src/Controls/StreetControls.js
gliegard: Finally I don't stop animations anymore.

src/Controls/StreetControls.js
gliegard: ok

src/Controls/StreetControls.js
gliegard: ok

src/Controls/StreetControls.js
gliegard: oki

src/Controls/StreetControls.js
gliegard: yep it's changed

src/Controls/StreetControls.js
gliegard: Adding a newline in the comment don't work in the doc.

src/Controls/StreetControls.js
gliegard: ok

src/Controls/StreetControls.js
gliegard: I don't understand what you mean. The object 'factor' is constant, even if it's content 'factor.t' is varying.

src/Controls/StreetControls.js
gliegard: ok. side answer: First I try without a Group but I didn't manage to make it work.

src/Controls/StreetControls.js
gliegard: I've already changed it.

src/Core/TileMesh.js
gliegard: yep

src/Controls/StreetControls.js
gliegard: You don't like these ones ?
;)
Yes good idea, I change that

src/Controls/StreetControls.js
zarov: Weird, I thought it would. Thanks !

src/Controls/StreetControls.js
```diff
@@ -1,17 +1,58 @@
...
+function initSurfaces(circle = createCircle(), rectangle = createRectangle()) {
+    rectangle.visible = false;
+    circle.visible = false;
+
+    // surface object is used to store the surfaces
+    const surfaces = new THREE.Object3D();
+    surfaces.add(rectangle);
+    surfaces.add(circle);
+
+    // to acces specific surface, we also add it to the surfaces object
```
zarov: I'd prefer to use [some functions of Object3D](https://threejs.org/docs/#api/en/core/Object3D.getObjectById) instead of adding properties.

src/Controls/StreetControls.js
zarov: Sorry, I meant that you could directly write `new TWEEN.Tween({ t: 0 }, this.tweenGroup)` below.

src/Controls/StreetControls.js
zarov: Ok, I'll also take a look at it to see if we can make it work

src/Controls/StreetControls.js
zarov: I like them, but we never know, someone may want a heart and a flower ;) More seriously, we could imagine a sort of game where you are moving a spider for example, and then replace the two meshes by a spider.

src/Controls/StreetControls.js
gliegard: The problem is I don't know the Id given by ThreeJs, and since we can add a custom Mesh, I can't sure of the name of the Mesh.
I change the code to add these meshes as properties of this StreetControls.

zarov: Good point, nvm then

src/Controls/StreetControls.js
gliegard: Finally, I use existing picking method !! :)

src/Controls/StreetControls.js
gliegard: Finally, I use existing picking method !! :)

src/Controls/StreetControls.js
gliegard: Finally, I use existing picking method, and pick on specific layers !!

examples/immersive_paris.html
```diff
@@ -162,6 +162,9 @@
             var menuGlobe = new GuiTools('menuDiv');
             menuGlobe.view = view;
 
+            var d = new debug.Debug(view, menuGlobe.gui);
```
gchoqueux, line 168: Change opacity building doesn't work, may be projection material hasn't opacity propertie, no?

examples/immersive_paris.html
```diff
@@ -171,6 +174,15 @@
                 // set camera to current panoramic
                 view.controls.setCameraToCurrentPosition();
                 view.notifyChange(view.camera.camera3D);
+
```
gchoqueux, line 180: Could you use `view.getLayerById('Building')` instead of `for loop`?

src/Controls/StreetControls.js
```diff
@@ -1,17 +1,41 @@
...
+import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
 import FirstPersonControls from 'Controls/FirstPersonControls';
-import AnimationPlayer, { AnimatedExpression } from 'Core/AnimationPlayer';
 
-// Expression used to damp camera's moves
-function moveCameraExp(root, progress) {
-    const dampingProgress = 1 - Math.pow((1 - (Math.sin((progress - 0.5) * Math.PI) * 0.5 + 0.5)), 2);
-    root.camera.position.lerpVectors(root.positionFrom, root.positionTo, dampingProgress);
+const material = new THREE.MeshBasicMaterial({
+    side: THREE.DoubleSide,
```
gchoqueux: Why is double sided?

src/Controls/StreetControls.js
```diff
@@ -77,8 +113,74 @@ class StreetControls extends FirstPersonControls {
         this.previousPosition = newPreviousPosition;
     }
 
+    pick(event, pickGround = () => {}, pickObject = () => {}, target = new THREE.Vector3()) {
+        // get real distance to ground, with a specific method to pick on the elevation layer
+        this.view.getPickingPositionFromDepth(this.view.eventToViewCoords(event), target);
```
gchoqueux: It would be necessary to open an issue to add the distance calculation in `Picking.pickTilesAt`.

src/Controls/StreetControls.js
```diff
@@ -77,8 +113,74 @@ class StreetControls extends FirstPersonControls {
...
+        const distanceToGround = this.camera.position.distanceTo(target);
+
+        const intersectsBuildings = this.view.pickObjectsAt(event, 0, this.buildingsLayer);
+        const intersectsGround = this.view.pickObjectsAt(event, 0, this.view.tileLayer);
+
+        // to detect pick on building, compare first picked building distance to ground distance
+        if (intersectsBuildings.length && intersectsBuildings[0].distance < distanceToGround) { // pick buildings
+            this.updateSurface(intersectsBuildings[0].point, intersectsBuildings[0].face.normal, this.surfaceWall, this.surfaceGround);
```
gchoqueux: There are certain functions that do not need to be exposed (`updateSurface, onMouseUp`...)

src/Controls/StreetControls.js
gchoqueux: I can't overload them. How are you doing?

src/Controls/StreetControls.js
gchoqueux: Could you create a issue and add comment in the documentation?

src/Controls/StreetControls.js
gchoqueux: could you open a issue? to move this part in `cameraUtils` in the futur?

src/Controls/StreetControls.js
```diff
@@ -102,19 +204,114 @@ class StreetControls extends FirstPersonControls {
...
+        };
+
+        this.view.addFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, this.animationFrameRequester);
+    }
+
+    onComplete() {
+        this.view.removeFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, this.animationFrameRequester);
+    }
+
+    stopTweensAnimations() {
```
gchoqueux: could you rename in `stopAnimation`? the user doesn't know it's `tween`.
could add `this.view.removeFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, this.animationFrameRequester);` in `stopTweensAnimations`? 

or merge `onComplete` and `stopTweensAnimations`...

src/Controls/StreetControls.js
```diff
@@ -63,6 +79,26 @@ class StreetControls extends FirstPersonControls {
         this.keyGoToNextPosition = 90;
         this.keyGoToPreviousPosition = 83;
         this.keySetCameraToCurrentPosition = 65;
+
+        // Raycaster used to pick objects
+        this.raycaster = new THREE.Raycaster();
```
gchoqueux: could you remove `raycaster`, it's useless

src/Controls/StreetControls.js
```diff
@@ -77,8 +113,74 @@ class StreetControls extends FirstPersonControls {
         this.previousPosition = newPreviousPosition;
     }
 
+    pick(event, pickGround = () => {}, pickObject = () => {}, target = new THREE.Vector3()) {
+        // get real distance to ground, with a specific method to pick on the elevation layer
+        this.view.getPickingPositionFromDepth(this.view.eventToViewCoords(event), target);
+        const distanceToGround = this.camera.position.distanceTo(target);
+
+        const intersectsBuildings = this.view.pickObjectsAt(event, 0, this.buildingsLayer);
+        const intersectsGround = this.view.pickObjectsAt(event, 0, this.view.tileLayer);
```
gchoqueux: you could do the ground picking only if there's not picking with the building

src/Controls/StreetControls.js
gliegard:
```
            class myControl extends itowns.StreetControls {
                onClickOnGround(position) {
                    console.log('user click on ground !');
                }
            }
```

src/Controls/StreetControls.js
gliegard: yes #1036

src/Controls/StreetControls.js
gliegard: yes, i've updated #1036 !

src/Controls/StreetControls.js
gchoqueux:
```js
 pick(event, pickGround = () => {}, pickObject = () => {}, target = new THREE.Vector3())
```
 you don't use `target` apart from this function, you could remove it

gliegard: ok, I remove it (this method has been refactored during the review)

examples/immersive_paris.html
gliegard: Yes, thanks, I've updated OrientedImageMaterial and shaders. Now it supports opacity.

examples/immersive_paris.html
gliegard: yes, I've removed the loop. Finally I add the debug UI before,  when I add the layer.

src/Controls/StreetControls.js
gliegard: Ok, I fix one surface orientation, so I don't need DoubleSide anymore. Thanks !

src/Controls/StreetControls.js
gliegard: ok #1037

src/Controls/StreetControls.js
gliegard: yes, I've renamed to stopAnimation. And I 've remove onComplete because it's not need to be exposed.

src/Controls/StreetControls.js
gliegard: yep

src/Controls/StreetControls.js
gliegard: Yes

src/Controls/StreetControls.js
```diff
@@ -77,8 +160,39 @@ class StreetControls extends FirstPersonControls {
...
+    onMouseUp(event) {
+        super.onMouseUp();
+        if (this._stateOnMouseDrag) {
+            this._stateOnMouseDrag = false;
+        } else {
+            pick(event, this.view, this.buildingsLayer, this.onClickOnGround.bind(this), this.onClickOnWall.bind(this));
+        }
+    }
+
+    onMouseMove(event) {
```
zarov, line 170: I have an error when moving the mouse before the layer is ready. Could you add a condition like `if (!layer.ready)` ? Or something that fixes the error.

zarov (commented):
Overall it seems the quality downgraded. When i click on the street to move, it takes one second to start, and then it is not smooth at all.

src/Controls/StreetControls.js
gliegard: Yes I moved some of them

src/Controls/StreetControls.js
gliegard: Yes, thanks, I 've added a bugfix in View

zarov (approved):

src/Controls/StreetControls.js
gchoqueux: `this.view.removeFrameRequester` isn't removed in all cases

gliegard: yes, thanks, updated

gchoqueux (approved):
thanks

PR #1022, opened 2019-01-31
Add 3D Tiles extension support and implement 3DTILES_batch_table_hierarchy extension

jailln:
## Short Description
This PR:
  * Introduces proposes a support of [3D Tiles extensions](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/extensions)
  * Exemplify this support with the implementation of [3DTILES_batch_table_hierarchy extension](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/extensions/3DTILES_batch_table_hierarchy)
  * Adds an example `3dtiles_hierarchy.html` which displays a 3D Tiles tileset with 3DTILES_batch_table_hierarchy extension
  * (side effect): update 3d tiles batch table support 

## Detailed description:

Proposed architecture for managing 3D Tiles extensions:
  * For implementing an extension in iTowns, one needs to write:
       * a class (e.g. `BatchTableHierarchyExtension`) extending `$3dTilesAbstractExtension` (abstract base classes for extensions).
       * a parser returning a promise resolving with an object of this class (e.g. parse method in `src/Parser/BatchTableHierarchyExtensionParser.js`)
  * iTowns has a global object `$3dTilesExtensions` of type `$3DTilesExtensions` which is exposed through the API. This object holds the registered extensions by mapping their names to their parser. To use an extension (whether defined in iTowns or user-defined), one needs to register this extension to `$3dTilesExtensions` (e.g. [here](https://github.com/iTowns/itowns/compare/master...jailln:batchtable_hierarchy#diff-70bc25ebdc51e081755f49f65b6c9ae4R68))
   * When parsing a 3D Tiles class (e.g. `BatchTable`), if an extension is found in the json, call the parser associated to this extension using `$3dTilesExtensions` (e.g. [here](https://github.com/iTowns/itowns/compare/master...jailln:batchtable_hierarchy#diff-d712526466c4b0c8ba9411cdceab250eR132)). The returned parsed object (from the extension) is then stored to an extensions object of the 3D Tiles class (e.g. `BatchTable`) 
   * When calling a method from a 3DTiles class (e.g. [getPickingInfo](https://github.com/iTowns/itowns/compare/master...jailln:batchtable_hierarchy#diff-d712526466c4b0c8ba9411cdceab250eR45) of BatchTable), this method calls the equivalent methods of the extensions objects (e.g. [getPickingInfo](https://github.com/iTowns/itowns/compare/master...jailln:batchtable_hierarchy#diff-da89552b85f2977d5d6eea94dd0240b9R71) of `BatchTableHierarchyExtension`)

This has several advantages:
  * A user can easily define an extension and append it to iTowns
  * There is no references to specific extensions in 3D Tiles classes 

This is only a proposal and I would appreciate some feedback :)


examples/3dtiles.html
```diff
@@ -73,49 +73,64 @@
...
+                    return findBatchTable(object.parent);
                 } else {
                     return undefined;
                 }
             }
+
             function picking(event) {
-                var htmlInfo = document.getElementById('info');
-                htmlInfo.innerHTML = ' ';
+                const htmlInfo = document.getElementById('featureInfo');
```
zarov: Avoid using ES6 components in our html examples. While iTowns is transpiled to ES5, examples are not.

examples/3dtiles.html
```diff
@@ -73,49 +73,64 @@
...
-                else if (object.parent) {
-                    return findBatchTableParent(object.parent);
+                } else if (object.parent) {
+                    return findBatchTable(object.parent);
                 } else {
                     return undefined;
                 }
             }
+
             function picking(event) {
```
zarov: Could you move this into a helper in `examples/js` maybe ?

examples/js/GUI/GuiTools.js
```diff
@@ -122,3 +122,31 @@ GuiTools.prototype.addGUI = function addGUI(name, value, callback) {
 GuiTools.prototype.colorLayerFolder = function colorLayerFolder(nameLayer, value) {
     this.colorGui.colorLayerFolder(nameLayer, value);
 };
+
+// Recursive function that creates an HTML List from a javascript
```
zarov, line 126: As it is recursive, we could get quite big lists displayed: can you reduce the `padding` of `ul` for it to be more readable ?

src/Parser/BatchTableHierarchyExtensionParser.js
```diff
@@ -0,0 +1,109 @@
+import { $3dTilesAbstractExtension } from '../Provider/3dTilesProvider';
+
+/** Class for storing and accessing information relative to the
+ *  3DTILES_batch_table_hierarchy extension of 3D Tiles */
```
zarov: Could you add a link to [the specification](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/extensions/3DTILES_batch_table_hierarchy) ?

src/Provider/3dTilesProvider.js
```diff
@@ -7,6 +7,83 @@ import Extent from 'Core/Geographic/Extent';
 import { pre3dTilesUpdate, process3dTilesNode, init3dTilesLayer } from 'Process/3dTilesProcessing';
 import utf8Decoder from 'Utils/Utf8Decoder';
 
+/**
+ * Class mapping 3D Tiles extensions names to their associated parsing methods
+ */
+export class $3DTilesExtensions {
```
zarov: Do we really want to export it if we export its instance below ?

zarov (commented):
Well this is a really good PR, thanks ! Not a lot to say, as I am not an expert in 3dtiles, so I'm going to have another look at it next week again to be sure I understood everything.

examples/3dtiles.html
jailln: Ok, done

examples/3dtiles.html
jailln: Yes, done

examples/js/GUI/GuiTools.js
jailln: You mean in the css then ? (Or is there a way to do that in JS ?) 

If in the css, maybe directly [here](https://github.com/iTowns/itowns/blob/master/examples/css/example.css#L53) or do you prefer that I do it in the examples that use this recursive function (3dtiles.html and 3dtiles_hierarchy.html) ?

src/Provider/3dTilesProvider.js
jailln: Yep you're right, I removed the export.

src/Parser/BatchTableHierarchyExtensionParser.js
jailln: Ok, done

examples/js/GUI/GuiTools.js
zarov: Yes sorry, in `example.css`

jailln:
Thanks for your review !

Apart from your comments, I noticed that I forgot to add BatchTableParser.js and BatchTableHierarchyExtensionParser.js to jsdoc and that 3DTilesProvider.js wasn't generated.

A few questions about the doc:
 * What is the criteria to document portions of code with jsdoc (and hence be generated in itowns doc in the end) ? Is it only for exported classes, methods, etc. ? (I used jsdoc for almost every class and function in this PR but maybe it wasn't necessary all the time ?)
 * I cannot manage to output 3dTilesProvider doc in the last commit I pushed; do you have an idea of why ?

docs/navigation.json
```diff
@@ -50,7 +50,15 @@
...
         "GpxParser",
         "VectorTileParser",
-        "CameraCalibrationParser"
+        "CameraCalibrationParser",
+        "BatchTableParser",
+        "BatchTableHierarchyExtensionParser"
+    ],
+
+    "Provider": [
+        "URLBuilder",
```
zarov: Don't forget to remove them from above !

zarov:
> What is the criteria to document portions of code with jsdoc (and hence be generated in itowns doc in the end) ? Is it only for exported classes, methods, etc. ? (I used jsdoc for almost every class and function in this PR but maybe it wasn't necessary all the time ?)

Let's say that it would be the minimum to have all the exported classes outside iTowns (so declared in `Main.js`) with generated documentation. In your case, it could be interesting to have your "sub" classes generated as well, but the current template doesn't support it yet. But it is welcome and thanks for writing it !

> I cannot manage to output 3dTilesProvider doc in the last commit I pushed; do you have an idea of why ?

I can't check it right now, but I would assume it is because the name in iTowns is `$3dTilesProvider` with a `$`, and the template is poorly designed to handle a difference between the filename and the exported class (thus the first problem). My bad, I hope to work on this quickly !

examples/js/GUI/GuiTools.js
jailln: Done, it turns out that there was a way to do it in js

docs/navigation.json
jailln: My bad I didn't see that they were in Source, I'll let them there then.

jailln:
Thanks for the clarification :+1: 

Regarding the doc in 3dTilesProvider: from what you said I think it cannot be generated with the current template so I removed it from the config json files of jsdoc for now and maybe we can open an issue about it; WDYT ?

zarov:
> Regarding the doc in 3dTilesProvider: from what you said I think it cannot be generated with the current template so I removed it from the config json files of jsdoc for now and maybe we can open an issue about it; WDYT ?

So I could generate some sub-classes documentation like `BatchTable`, by simply adding it to `navigation.json`, but `$3DTilesExtensions` couldn't work. So yes we can open an issue to discuss that. Generally speaking, you can open an issue about anything not working/missing in the generation of the documentation ;)

examples/3dtiles.html
zarov: Can't the whole `picking` method be moved in the helper ?

docs/navigation.json
zarov: Oh no, I would prefer to move them in `Provider`, as it is the name of the folder they are in.

jailln: Ok, done

examples/3dtiles.html
jailln: Yep you're right, done

zarov (approved):
LGTM, what a quick and clean PR, even considering the length ! Thank you very much. I'll wait a little to merge it, I think @gchoqueux wanted to take a quick look at it.

jailln:
Thanks!

I just realized that I misunderstood your first comment about the doc and that I can generate the documentation for the classes BatchTable, BatchTableHierarchy, etc. I'll push this modification this afternoon.

jailln:
Documented 3D Tiles related classes are now exposed in the documentation. The only thing I couldn't manage to expose is `$3dTilesExtensions`; maybe because it is not a class but a variable ?

Anyway, the PR is ready for my part.

examples/3dtiles.html
```diff
@@ -73,49 +73,67 @@
...
+                    var interAttributes =
+                        intersects[i].object.geometry.attributes;
+                    if (interAttributes && interAttributes._BATCHID) {
+                        // face is a Face3 object of THREE which is a triangular
+                        // face. face.a is its first vertex
+                        var vertex = intersects[i].face.a;
+                        // get batch id of the face
+                        var batchID = interAttributes._BATCHID.array[vertex];
+                        var batchTable = findBatchTable(
+                            intersects[i].object, 0);
```
gchoqueux: `0`  is useless in `findBatchTable`?

src/Parser/BatchTableParser.js
```diff
@@ -1,16 +1,89 @@
...
+     * Creates and returns a javascript object holding the displayable
+     * information from the batch table
+     * @param {integer} featureId - id of the feature
+     * @returns {Object} - displayable information relative to the batch table
+     * for the feature with id=featureId. Object is
+     * formatted as follow: {BatchTable: {BatchTableFeatureProperties}
+     */
+    getPickingInfo(featureId) {
+        const featureDisplayableInfo = {};
+        if (Object.keys(this.content).length === 0) {
```
gchoqueux: Could you use `this.batchLength === 0` instead of `if (Object.keys(this.content).length === 0)`?
if yes,[ the warning has already done](https://github.com/iTowns/itowns/pull/1022/commits/039c3813dd31f9d40a1da43b580c1acad6a3738d#diff-d712526466c4b0c8ba9411cdceab250eR20)

src/Parser/BatchTableParser.js
```diff
@@ -84,6 +121,30 @@ export default {
...
+                            .then((extObject) => {
+                                batchTable.extensions[extName] = extObject;
+                                batchTable.removeExtensionFromContent(extName);
+                            }));
+                    }
+                });
+        }
+        // values[0] is the results of the first promise pushed in promises,
+        // i.e. batchTable
+        return Promise.all(promises).then(values => values[0]);
```
gchoqueux: you could return directly `batchTable`:
```js
return Promise.all(promises).then(() => batchTable);
```
and remove ` promises.push(Promise.resolve(batchTable));`

examples/3dtiles_hierarchy.html
```diff
@@ -0,0 +1,149 @@
...
+                longitude: -75.61349109370566,
+                latitude: 40.04425929988183,
+                altitude: 45.52006505522877
+            };
+
+            var viewerDiv = document.getElementById('viewerDiv');
+
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+
+            var cameraLookAt = new itowns.THREE.Vector3(1215018.0820894851,
```
gchoqueux: Could you use `view.controls.lookAtCoordinate` to set camera position .
You could use `view.controls.getLookAtCoordinate ( )`  to get the parameters to use.

gchoqueux (commented):
Thank you for this contribution, I have some small comments.

examples/3dtiles.html
jailln: Yep you're right. It has been removed in the following commits.

src/Parser/BatchTableParser.js
jailln: Yep you're right, done.

src/Parser/BatchTableParser.js
jailln: Yes, it's cleaner this way; done

examples/3dtiles_hierarchy.html
jailln: I'm sorry I don't understand your comment, could you be more specific please ?

examples/3dtiles.html
gchoqueux: Ok fine. 
In the same PR, it would be necessary to squash, the modifications of the same code part, in the same commit. It's more simple for the reviewing.

examples/3dtiles_hierarchy.html
gchoqueux: Could you use `view.controls.lookAtCoordinate({coord: new itowns.Coordinates('EPSG:4326', longitude, latitude), range: 500, tilt: 10, heading}, false)` to place camera.

examples/3dtiles.html
jailln: That's what I tried to do while still keeping a "logical" commit division based on new features added but I might have missed some details on the road like this one. Anyway, I edited the commit ;)

examples/3dtiles_hierarchy.html
jailln: Ok, thanks, done

gchoqueux: I had not seen the variable `positionOnGlobe`, you could simply write:
```js
var positionOnGlobe = {
    longitude: -75.61127,
    latitude: 40.04190,
    altitude: 0,
};
view.controls.lookAtCoordinate({ range: 400, tilt: 10, heading: -145 }, false);
```

jailln: Done, thanks it's cleaner this way

gchoqueux:
Thanks for your work.

nosy-b:
Ya, Thanks @jailln for this nice work :)

PR #1026, opened 2019-02-04
chore: update THREE.js to r101.1

gchoqueux:
## Description

- [x] `FirstPersonControls.lookAt()`. lat, lon, phi, theta and target were removed from the public API. FirstPersonControls also respects the initial camera orientation now. **=> nothing use FirstPersonControls, it is rewrited in iTowns**
- [x] MeshStandardMaterial and MeshPhysicalMaterial now preserve energy for IBL lighting, resulting in brighter, more accurate colors for metallic materials with high roughness values when lit via a map generated by PMREMGenerator. **=> No impact**
- [x] Scene Introduce `.dispose()`.  **=> we can use this function with the architecture in layer**
- [x] Added support for `Multisampled Renderbuffers`. **=> I open a specific issue on this subject #1028, because it can be very interesting.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2568,7 +2568,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
-          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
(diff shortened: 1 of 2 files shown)


zarov:
> nothing use FirstPersonControls, it is rewrited in iTowns

@gliegard please have a look on how it has evolved, as it is you who worked recently on this

> we can use this function with the architecture in layer

Ok, let's open an issue about it.

zarov (approved):

PR #1029, opened 2019-02-07
chore(test): rename test/examples to test/functional

gchoqueux:
## Motivation and Context
This renaming is to avoid a ambiguity with the examples folder.


zarov (approved):
OK, seems logical.

PR #1032, opened 2019-02-12
feat(parser): add a kml parser and modify gpx parser

zarov:
## Description

- I added a very tiny KML parser (almost the same as the one we used to have some time ago)
- I modified the GPX parser to be very simple like the KML one
- Both parsers use `togeojson`
- The `gpx.html` example has been rewritten to generate some meshes, as the new GPX parser doesn't generate it now

## Motivation and Context

I'm doing this now, because I'm making `FileSource` simpler, and making `DataSourceProvider` malleable. This will be cover in my next PR (coming very soon).

### DIFF
--- a/examples/gpx.html
+++ b/examples/gpx.html
@@ -25,9 +25,6 @@
 
             setupLoadingScreen(viewerDiv, view);
 
-            function addLayerCb(layer) {
-                view.addLayer(layer);
-            }
             // Add one imagery layer to the scene
             // This layer is defined in a json file but it could be defined as a plain js
             // object. See Layer* for more info.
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -79,7 +79,6 @@
     "puppeteer": "^1.12.1",
     "replace-in-file": "^3.4.3",
     "three": "^0.101.1",
-    "three.meshline": "^1.2.0",
```
gchoqueux, line 82: You don't  remove `three.meshline` from `peerDependencies`?

src/Parser/KMLParser.js
```diff
@@ -0,0 +1,26 @@
...
+ * @module KMLParser
+ */
+export default {
+    /**
+     * Parse a KML file content and return a [FeatureCollection]{@link
+     * module:GeoJsonParser~FeatureCollection}.
+     *
+     * @param {XMLDocument} kml - The KML file content to parse.
+     * @param {Object} options - Options controlling the parsing, see the parse
+     * method of GeoJsonParser for a list of possible options.
```
gchoqueux: could you define type `jsdoc`, to document this option, (like `cameraTransformOptions`)

examples/gpx.html
```diff
@@ -46,15 +43,64 @@
             itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
 
+            // update the waypoint
+            const positionCamera = new itowns.THREE.Vector3();
+            function getDistance(object, camera) {
+                const point = object.geometry.boundingSphere.center.clone().applyMatrix4(object.matrixWorld);
```
gchoqueux: could you declare `point` before function  and use copy to avoid a cloning?

examples/gpx.html
```diff
@@ -46,15 +43,64 @@
...
+                positionCamera.setFromMatrixPosition(camera.matrixWorld);
```
gchoqueux: you could use `camera.position` directly.

gchoqueux (commented):
thanks, it's good way.

package.json
zarov: Forgot it, thanks

src/Parser/KMLParser.js
zarov: On it

examples/gpx.html
zarov: Yes

examples/gpx.html
zarov: Done

src/Parser/KMLParser.js
zarov: Note: typedef like `cameraTransformOptions` are not yet generated in the documentation, I'm working on it in a separate PR. Meanwhile, I don't think this should stop the PR to be merged ;)

examples/gpx.html
```diff
@@ -46,15 +43,65 @@
             itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
 
+            // update the waypoint
+            var point = new itowns.THREE.Vector3();
+            function getDistance(object, camera) {
+                point.copy(object.geometry.boundingSphere.center);
```
gchoqueux: if there's not matrix transform, so point is useless.
You could write : 
```js
return camera.position.distanceTo(object.geometry.boundingSphere.center);
```

zarov: Done

gchoqueux (approved):
thanks

Issue #892, opened 2018-11-06
Proposition : creation of a layer from an object


hcaillau:
Instead of :

buildings.source = {
            	url: 'https://example.com/geosjon.json',
            	protocol: 'file',
            	projection: 'EPSG:4326',
            	format: 'application/json',
            	zoom: { min: 2, max: 2 }
            };

could be great to be able to do this :

buildings.source = {
            	source :  geojson,
            	protocol : 'object',
            	projection: 'EPSG:4326',
            	zoom: { min: 2, max: 2 }
            };

where geojson is an object.

zarov:
Yes, thanks for the proposition: we should indeed be able to load data without fetching it, but by providing it directly. That's basically a new type of source to add, if you want you could do it, I don't see too much work on this.

zarov:
If you need something to work right now, you can patch iTowns using this commit https://github.com/zarov/itowns/commit/2dd057a3d7ee9e9f4a06d977d15de5cc38cd424f

It is a quick hack, but I don't want to have something like this to do until I'm done with the quick reorganization explained in #880

gchoqueux:
@zarov it's added in #1033 ?

zarov:
Yes, and we don't even have a new protocol or whatsoever

gchoqueux:
resolved with #1033

PR #1033, opened 2019-02-12
refacto/feat: create a layer/source from data loaded by the user

zarov:
## Description

Main goal of the PR: create a Layer/Source directly from data loaded 
by the user. To achieve this, it is mainly `DataSourceProvider` and
`FileSource` that have been rewritten. So now we have three way of 
receiving data:
- by fetching it, parsing it and converting it in the provider
- by giving fetched data to the layer and parsing it and converting it
  in the provider
- by giving parsed data to the layer and converting it in the provider

And thus FileSource can be simplified drastically and extended to
receive ANY type of data, not only geojson, kml and gpx (see #1032)

Documentation and examples have been updated, and I tried to not 
introduce breaking changes, but I may have miss some.

## Motivation and Context

People have asked about it in issue (#892), offline and I need this 
working for another thing. It is also the occasion to clean `FileSource` 
and `DataSourceProvider` that were too messy imho.

## Problems

I still got two problems, that maybe someone can help with:
- the first one is visible in example `globe_vector`, the GPX path and 
the geojson texture seems to overlap, but I don't understand why
- in example `globe_geojson_to3D`, I can't seem to make it work too

NOTE: this PR needs #1032 to be merged first

### DIFF
--- a/examples/globe_geojson_to3D.html
+++ b/examples/globe_geojson_to3D.html
@@ -68,7 +68,7 @@
                 projection: 'EPSG:4326',
                 format: 'application/json',
                 zoom: { min: 7, max: 7 },
-            }, view.referenceCrs);
+            });
 
             view.addLayer(ariege).then(function menue(layer) {
                 var gui = debug.GeometryDebug
--- a/examples/globe_vector.html
+++ b/examples/globe_vector.html
@@ -13,8 +13,7 @@
                 position: absolute;
                 z-index: 1000;
                 color: #CECECE;
-                font-family: 'Open Sans',
-                sans-serif;
+                font-family: 'Open Sans', sans-serif;
                 font-size: 14px;
                 line-height: 18px;
                 text-align: left;
@@ -63,7 +62,10 @@
             itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(config) {
                 config.source = new itowns.WMTSSource(config.source);
                 var layer = new itowns.ColorLayer('Ortho', config);
-                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+                view.addLayer(layer).then(function _() {
+                    menuGlobe.addLayerGUI.bind(menuGlobe);
+                    itowns.ColorLayersOrdering.moveLayerToIndex(view, 'Ortho', 0);
+                });
             });
             // Add two elevation layers.
             // These will deform iTowns globe geometry to represent terrain elevation.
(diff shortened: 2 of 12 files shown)


zarov:
Something else I would change: as you can see, you can use `source.fetchedData` and `source.parsedData` to add new data that you have already fetched or parsed. To avoid complexity for the user, I was thinking to maybe also add `data` when declaring the source, so you could have:

```js
Fetcher.json('foo.geojson').then((fetchedData => {
    var source = new FileSource('source', {
        data: fetchedData,
    });
...
``` 
or

```js
Fetcher.json('foo.geojson').then(GeoJsonParser.parse)
    .then((parsedData => {
        var source = new FileSource('source', {
            data: parsedData,
        });
...
``` 

WDYT ?

zarov:
A quick update: `globe_geojson_to3D` seems to work partially, it seems there is a problem in `FeatureProcessing`. Needs more investigation unfortunately.

zarov:
Update: the first problem (`globe_vector`) has been explained in #1040, so I'm considering it as not blocking this PR anymore. The second problem (`globe_geojson_to3d`) seems fixed.

examples/globe_geojson_to3D.html
```diff
@@ -62,13 +62,15 @@
             ariege.convert = itowns.Feature2Mesh.convert({
                 color: color,
                 extrude: extrude });
+            ariege.projection = view.tileLayer.extent.crs();
```
gchoqueux: it is not up to the user to specify the target projection. it must be `view.referenceCrs`?

examples/globe_geojson_to3D.html
```diff
@@ -62,13 +62,15 @@
...
 
             ariege.source = new itowns.FileSource({
                 url: 'https://raw.githubusercontent.com/gregoiredavid/france-geojson/master/departements/09-ariege/departement-09-ariege.geojson',
                 projection: 'EPSG:4326',
                 format: 'application/json',
                 zoom: { min: 7, max: 7 },
-            }, view.referenceCrs);
+                parser: itowns.GeoJsonParser.parse,
```
gchoqueux: it will be necessary to make more friendly: add specofic `Source` by format. This is **Breaking change**? it's necessary to add parser?

examples/globe_vector.html
```diff
@@ -75,53 +77,79 @@
...
+                url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/croquis.kml',
+                projection: 'EPSG:4326',
+                fetcher: itowns.Fetcher.xml,
+                parser: itowns.KMLParser.parse,
+            });
 
             var kmlLayer = new itowns.ColorLayer('Kml', {
                 name: 'kml',
                 transparent: true,
+                projection: view.tileLayer.extent.crs(),
```
gchoqueux: it's not up to the user to specify the target projection, this should be done by itowns, when adding the layer.

examples/globe_vector.html
```diff
@@ -75,53 +77,79 @@
...
+                        buildExtent: true,
+                        crsIn: 'EPSG:4326',
+                        crsOut: view.tileLayer.extent.crs(),
+                        mergeFeatures: true,
+                        withNormal: false,
+                        withAltitude: false,
+                    });
+                }).then(function _(parsedData) {
+                    var ariegeSource = new itowns.FileSource({
+                        projection: 'EPSG:4326',
```
gchoqueux: in this case, the projection is necessary?

src/Core/View.js
```diff
@@ -193,11 +193,6 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
                 providerPreprocessing = Promise.resolve();
             }
         } else if (layer.source) {
-            // TODO: move to dataSourceProvider
```
gchoqueux, line 194: nice

src/Process/FeatureProcessing.js
```diff
@@ -104,6 +104,17 @@ export default {
             // if request return empty json, WFSProvider.getFeatures return undefined
             result = result[0];
             if (result) {
+                // special case for FileSource, as it is not tiled and we need
+                // to attach it to the correct node
+                if (layer.source && layer.source.isFileSource) {
+                    for (const extentSrc of extentsSource) {
```
gchoqueux, line 116: could we use the global extent of `parsedData`?

src/Provider/DataSourceProvider.js
```diff
@@ -27,36 +27,48 @@ const error = (err, url, source) => {
...
         crsOut: layer.projection,
         // TODO FIXME: error in filtering vector tile
         // filteringExtent: extentDestination.as(layer.projection),
-        filteringExtent: layer.isGeometryLayer ? extentDestination : undefined,
+        filteringExtent: !source.isFileSource && layer.isGeometryLayer ? extDest : undefined,
         overrideAltitudeInToZero: layer.overrideAltitudeInToZero,
         filter: layer.filter,
         isInverted: source.isInverted,
-        mergeFeatures: layer.mergeFeatures === undefined ? true : layer.mergeFeatures,
+        mergeFeatures: source.isFileSource || layer.mergeFeatures === undefined ? true : layer.mergeFeatures,
```
gchoqueux: the source shouldn't decide in the options parsing, it's rather the layer that influences the options.

src/Source/FileSource.js
```diff
@@ -1,71 +1,70 @@
-import togeojson from '@mapbox/togeojson';
 import Source from 'Source/Source';
-import Fetcher from 'Provider/Fetcher';
-import Extent from 'Core/Geographic/Extent';
-import GeoJsonParser from 'Parser/GeoJsonParser';
-
-function getExtentFromGpxFile(file) {
```
gchoqueux, line 7: a good suppression

src/Source/FileSource.js
```diff
@@ -76,74 +75,55 @@ function fileParser(text) {
...
      * @constructor
      */
-    constructor(source, crsOut) {
+    constructor(source) {
         if (!source.projection) {
             throw new Error('source.projection is required in FileSource');
         }
+
+        // the fake url is for when we use the fetchedData or parsedData mode
+        source.url = source.url || 'fake-file-url';
```
gchoqueux, line 122: could you verify if `source.url` is undefined so `fetchedData` or `parsedData` are defined?

gchoqueux (changes requested):
thanks, it's a good enhancement

examples/globe_geojson_to3D.html
zarov: Yes I agree, we can add some new Sources. This is not a breaking change, as `format` is still read. In fact, we don't really need to have the parser specified here.

examples/globe_geojson_to3D.html
zarov: Ok I'm looking into changing that

examples/globe_vector.html
zarov: No, but the previous version of `FileSource` has `projection` mandatory, so I kept it like this. But we can change it if you want.

src/Process/FeatureProcessing.js
zarov: It is the case below ? I don't understand sorry

src/Provider/DataSourceProvider.js
zarov: Ok

src/Source/FileSource.js
zarov: Yes good idea

examples/globe_geojson_to3D.html
zarov: Removed, because unnecessary

examples/globe_geojson_to3D.html
zarov: Removed to avoid the confusion

examples/globe_vector.html
zarov: Removed

examples/globe_vector.html
zarov: Removed because the `FileSource` now reads `parsedData` if present.

src/Process/FeatureProcessing.js
zarov: Factorized

src/Provider/DataSourceProvider.js
zarov: Removed

src/Source/FileSource.js
zarov: Done

gchoqueux:
in `/examples/globe_vector.html`, there are 2 problems:
* no line with gpx;
* **ariege** json can be displayed halfway;

zarov:
* I'm working on the gpx problem
* the ariege json is the same problem as in #1040

zarov:
Lines are now visibles on the gpx path

gchoqueux:
There's `non-coherent result 0 vs 1.` error  in example `globe_vector.html`.

zarov:
So the last problem has been solved ! It  was due to the fact that `FileSource#urlFromExtent` was returning the same string without some unicity. This is now fixed with the generation of a random string.

gchoqueux:
there is always the problem with `non-coherent result`.

zarov:
The caching in `DataSourceProvider` has been reworked, and it fixes all the problem we got in a very simpler way.

src/Provider/DataSourceProvider.js
```diff
@@ -47,16 +51,26 @@ function FetchAndConvertSourceData(url, layer, extentSource, extentDestination)
         withAltitude: layer.isGeometryLayer,
     };
 
+    data.coords = extSrc;
+
+    return parser(data, options).then((parsedFile) => {
+        if (source.isFileSource) {
+            source.parsedData = parsedFile;
+            if (parsedFile.extent._crs != 'EPSG:4978') {
```
zarov, line 59: We currently have some problems where the `parsedFile.extent` is in `4978`: this makes no sense (in itowns at least) to have a geocentric extent. This problem is not fixed in this PR, but an issue will be opened about it.

gchoqueux:
there's a problem with immersive example...

zarov:
Fixed in 5d38ecc but not very satisfied with it... WDYT ?

src/Provider/DataSourceProvider.js
gchoqueux: could you open it?

gchoqueux (approved):
LGTM, Thanks for your effort.

gchoqueux:
Rebase and edit commit 5d38ecc message before merge the PR.

src/Provider/DataSourceProvider.js
zarov: Done in #1063

zarov:
Thanks !

PR #1034, opened 2019-02-13
feat(examples): add an example with 3D Tiles on plane

jailln:
## Description
This PR adds a new iTowns example displaying 3D Tiles layers on a planar view. 

The layers displayed in this example respectively represent the full city of Lyon in 2009, 2012 and 2015 and the user can switch between layers with a "temporal bar". 

## Motivation and Context
This example demonstrate the capability of iTowns to scale-up with 3D Tiles layers.


examples/planar_3dtiles.html
```diff
@@ -0,0 +1,205 @@
+<html>
+    <head>
+        <title>Itowns - 3D Tiles planar example</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/loading_screen.css">
+        <style type="text/css">
```
gchoqueux: could you move style in `example.css`?
and could you add a `padding: 10;`?

examples/planar_3dtiles.html
```diff
@@ -0,0 +1,205 @@
...
+                useColorTextureElevation: true,
+                colorTextureElevationMinZ: 37,
+                colorTextureElevationMaxZ: 240,
+                source: wmsElevationSource,
+            });
+
+            view.addLayer(wmsElevationLayer);
+
+            // Camera
+            view.camera.setPosition(new itowns.Coordinates('EPSG:3946',
```
gchoqueux: could you use `itowns.CameraUtils.transformCameraToLookAtTarget` [like this](https://github.com/iTowns/itowns/blob/master/examples/planar.html#L96)?

gchoqueux (commented):
Nice example. I have just 2 comments.
The 3D tiles are very slow to load that comes from the server (`threedtilesets.alpha.grandlyon`) ?

test/functional/planar_3dtiles.js
```diff
@@ -0,0 +1,57 @@
...
+                x: 194,
+                y: 100,
+            })
+                .map(p => p.layer.id),
+        );
+
+        assert.ok(layers.indexOf('planar') >= 0);
+    });
+
+    it('should pick the correct building for 2012', async () => {
```
gchoqueux, line 28: The test failed because` timeout of 60000ms` exceeded.  So, it would have to position the camera differently to load less data.

examples/planar_3dtiles.html
jailln: Yes, done

examples/planar_3dtiles.html
jailln: Ok, done

test/functional/planar_3dtiles.js
jailln: Yep good idea, done; let's see if it succeeds now

jailln:
> Nice example. I have just 2 comments.
> The 3D tiles are very slow to load that comes from the server (`threedtilesets.alpha.grandlyon`) ?

Thanks for your review. 
Yes that comes from the server but this is currently our best option (in terms of stability and durability) to host data and demos.

test/functional/planar_3dtiles.js
gchoqueux: it's work

gchoqueux:
> Yes that comes from the server but this is currently our best option (in terms of stability and durability) to host data and demos.

The b3dm files are big (~1.5MB by b3dm), this is slow down the data fetching. Theses files have metadata or textures?

jailln:
> > Yes that comes from the server but this is currently our best option (in terms of stability and durability) to host data and demos.
> 
> The b3dm files are big (~1.5MB by b3dm), this is slow down the data fetching. Theses files have metadata or textures?

They don't have metadata nor textures but the triangulation is not always optimum (see wireframe mode).

The spatial indexing method used for creating these tilesets is a 2d-tree with a predefined maximum number of features (e.g. buildings) in the leaves of the tree. I guess there is a trade-off to find between having more features in the leaves (and hence bigger tiles but less fecth calls) and having less features in the leaves (and hence smaller tiles but a deeper tree and more fecth calls). 

The year 2015 seems slower to load than 2009 and 2012 and it has a deeper tree with smaller tiles so maybe we should increase the maximum number of features per leaves for this year.

gchoqueux:
If there is no texture or metadata, the size of the b3dms seems too big compared to the complexity of Meshs. I will look exactly why. For the moment, the loading of the data is much too slow to propose this example.

gchoqueux:
I have several ways to decrease the size of `b3dm`:
* decrease the number of vertex using the indexes for the faces;
* don't export normals and calculate them with `THREE` when importing;
* use Draco compression;

by applying these 3 processes to the file [https://threedtilesets.alpha.grandlyon.com/TileSet_Lyon_2012_full/tiles/108.b3dm](https://threedtilesets.alpha.grandlyon.com/TileSet_Lyon_2012_full/tiles/108.b3dm), the `b3dm` size goes from `1.1Mo` to `28.5Ko`.

jailln:
Thanks for looking into this, the size benefit seems substantial! I have a few questions:

* Which software component(s) did you use to process the b3dm file ? 
* Do you implement the indexes of the faces with [sparse accessors](https://raw.githubusercontent.com/KhronosGroup/glTF/master/specification/2.0/figures/gltfOverview-2.0.0b.png) of glTF ? Some other way ?
* Is there a mechanism implemented in iTowns which looks for the normals in the loaded glTF file and computes them if they are not ?

gchoqueux:
>  Which software component(s) did you use to process the b3dm file ? 

I don't have real process, I tested Blender and [insimo](https://gltf.insimo.com/)

> Do you implement the indexes of the faces with [sparse accessors](https://raw.githubusercontent.com/KhronosGroup/glTF/master/specification/2.0/figures/gltfOverview-2.0.0b.png) of glTF? Some other way ?

I don't know what `insimo` does.

> Is there a mechanism implemented in **iTowns** which looks for the normals in the loaded **glTF** file and computes them if they are not ?

Yes, the normals are computed if they are not.

I write a wiki page [advices for generating small glTF files](https://github.com/iTowns/itowns/wiki/3D-Tiles#advices-for-generating-small-gltf-files).

jailln:
Thank you for the details, I will look into this in more details as soon as I can.

jailln:
See #1034

PR #1035, opened 2019-02-18
chore: fix 'npm ci' when it prints npm version.

gliegard:
chore: fix 'npm ci' when it prints npm version.

## Description
Before this commit, there was an error on Travis with node 11.10 (because in prepare.js, process.env.npm_config_user_agent was not defined),
Now, it get the npm version executing 'npm --version'.

## Motivation and Context
New PR doesn't work on Travis with Node 11.10

### DIFF
--- a/config/prepare.js
+++ b/config/prepare.js
@@ -1,4 +1,5 @@
 const chalk = require('chalk');
+const { execSync } = require('child_process');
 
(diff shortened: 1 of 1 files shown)


gchoqueux (approved):
thanks

PR #1038, opened 2019-02-21
fix(process): fix and unify calls to window.setTimeOut

zarov:
Noticed in `FeatureProcessing` that a `window.setTimeout` call was not correctly written. Fixed that and took the opportunity to unify all calls to this.


gchoqueux (commented):
thanks

gchoqueux (approved):

PR #1054, opened 2019-03-04
fix(core): event target change doesn't trigged with little change

gchoqueux:
In CameraUtils, event target is changed doesn't trigged with little change.


zarov (approved):
Maybe change it to something simpler to edit like `10e-8` ? Anyway do as you want, both are good for me.

PR #1056, opened 2019-03-04
feat(parser): add a shapefile parser

zarov:
The parser can take into account four files: shp, shx, dbf and prj. It
uses https://www.npmjs.com/package/shpjs

An example and some data has been added.

This PR should not be merged before #1033


zarov:
This PR can be reviewed now that #1033 has been merged.

gchoqueux (commented):
could you move data files in itowns sample data?

examples/shapefile.html
```diff
@@ -0,0 +1,79 @@
+<html>
```
gchoqueux, line 1: add in index.html + add image

examples/shapefile.html
```diff
@@ -0,0 +1,79 @@
...
+            // These will deform iTowns globe geometry to represent terrain elevation.
+            function addElevationLayerFromConfig(config) {
+                config.source = new itowns.WMTSSource(config.source);
+                var layer = new itowns.ElevationLayer(config.id, config);
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            }
+
+            // Load all the necessary files for a shapefile, parse them and
+            // display them.
+            Promise.all([
```
gchoqueux: could you add `Fetcher.Shape(baseUrl, extentionfiles = {shp: 'shp',.... }, networkOptions)`?

examples/shapefile.html
```diff
@@ -0,0 +1,79 @@
...
+            ]).then(function _(res) {
+                return itowns.ShapefileParser.parse({
+                    shp: res[0],
+                    dbf: res[1],
+                    shx: res[2],
+                    prj: res[3],
+                }, {
+                    buildExtent: true,
+                    crsIn: 'EPSG:4326',
+                    crsOut: view.tileLayer.extent.crs(),
```
gchoqueux, line 50: could you remove `crsOut`?

src/Parser/ShapefileParser.js
```diff
@@ -0,0 +1,78 @@
...
+        if (data.zip) {
+            result = shp.parseZip(data.zip);
+        } else if (data.shp && data.shx && data.dbf) {
+            result = Promise.all([
+                shp.parseShp(data.shp, data.prj),
+                shp.parseDbf(data.dbf),
+            ]).then(shp.combine);
+        }
+
+        options.crsIn = data.prj ? proj4(data.prj).oProj.datumName : undefined;
```
gchoqueux, line 74: why define `crsIn`  in example?

examples/shapefile.html
zarov: I can't ? It is needed by `GeoJsonParser`.

src/Parser/ShapefileParser.js
zarov: Removed it !

examples/shapefile.html
zarov: Done

examples/shapefile.html
zarov: So I pushed a quick draft, let me know what you think

examples/shapefile.html
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            // Load all the necessary files for a shapefile, parse them and
+            // display them.
+            itowns.Fetcher.multiple('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/shapefile/velib-disponibilite-en-temps-reel', {
+                arrayBuffer: ['shp', 'dbf', 'shx'],
+                text: ['prj'],
+            }).then(function _(res) {
+                return itowns.ShapefileParser.parse(res, {
+                    buildExtent: true,
```
gchoqueux, line 49: Sorry, effectively, `crsout` is necessary but  i think it's not user  friendly to write `view.extent.crs()`. 

In future, may be we'll add `view.avaibleProjectionColor()`.

zarov: Yes sorry I added it back. Yeah I agree, we may find something simpler.

zarov:
I completed the documentation and rebased everything in 2 commits

gchoqueux:
I have some comments:

- The example doesn't load, it needs to move to refresh globe. could you fix this issue?
- I remarked the points circle are deformed in the direction of latitude because the texture have a size squared. could you open a issue? a solution to fix it, is to use PM projection.
- The drawing of points is very slow because there is not test on extent. I have start a optimisation could add it?  
[0001-optimization-draw-point.patch](https://github.com/iTowns/itowns/files/2956137/0001-optimization-draw-point.txt)

zarov:
- This has been adressed in #1073 
- I can't find how to apply the correct projection, sorry. I'm opening an issue on this.
- Done

examples/shapefile.html
```diff
@@ -0,0 +1,71 @@
...
+            // This layer is defined in a json file but it could be defined as a plain js
+            // object. See Layer* for more info.
+            itowns.Fetcher.json('./layers/JSONLayers/OPENSM.json').then(function _(config) {
+                config.source = new itowns.TMSSource(config.source);
+                var layer = new itowns.ColorLayer('OPENSM', config);
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            });
+            // Add two elevation layers.
+            // These will deform iTowns globe geometry to represent terrain elevation.
+            function addElevationLayerFromConfig(config) {
```
gchoqueux: `addElevationLayerFromConfig` is useless.

src/Converter/Feature2Texture.js
```diff
@@ -65,27 +66,37 @@ function drawPoint(ctx, x, y, origin, scale, style = {}) {
...
     ctx.strokeStyle = style.stroke || 'red';
     ctx.stroke();
 }
 
+const coord = new Coordinates('EPSG:4326', 0, 0, 0);
+
 function drawFeature(ctx, feature, origin, scale, extent, style = {}) {
     let gStyle = style;
+    // Determinate with radius point
+    const px = 0.01;
```
gchoqueux: `px` needs to be determinate with radius, because if radius increases then the condition is wrong.

examples/shapefile.html
```diff
@@ -0,0 +1,71 @@
...
+            // display them.
+            itowns.Fetcher.multiple('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/shapefile/velib-disponibilite-en-temps-reel', {
+                arrayBuffer: ['shp', 'dbf', 'shx'],
+                text: ['prj'],
+            }).then(function _(res) {
+                return itowns.ShapefileParser.parse(res, {
+                    buildExtent: true,
+                    crsOut: view.tileLayer.extent.crs(),
+                });
+            }).then(function _(geojson) {
```
gchoqueux: As the parser returns a `Feature`, could you rename,  `geojson`, in `feature`?

gchoqueux:
> The example doesn't load, it needs to move to refresh globe. could you fix this issue?

fixed with #1068

examples/shapefile.html
zarov: removed

src/Converter/Feature2Texture.js
zarov: Damn I copied/pasted it without noticing the comment, my bad ! Fixing it.

examples/shapefile.html
zarov: Done

src/Converter/Feature2Texture.js
zarov: So this seems complicated to do: it depends a lot on the current projection/view, and I don't think this will simply things. What about allowing only a maximum radius size ? I don't see a case where a point is larger than 30 or 50px, and unless someone displays this point on a globe with a zoom at 0, there will be no artefact. What do you say ?

zarov: Nervemind, it is done simply in fact, my bad

src/Converter/Feature2Texture.js
```diff
@@ -65,27 +66,45 @@ function drawPoint(ctx, x, y, origin, scale, style = {}) {
...
 
         if (typeof (style) == 'function') {
             gStyle = style(properties, feature);
         }
+
+        gStyle.radius = Math.round(gStyle.radius) || 3;
+
+        // cross multiplication to know in the extent system the real size of
+        // the point
+        px = gStyle.radius * (extentDim.x / 256);
```
gchoqueux, line 95: you could use `ctx.canvas.width` instead of `256`

gchoqueux (approved):

PR #1058, opened 2019-03-04
chore: update to Three r102.1

### LINKED ISSUE #1044: Stereo example is broken
## Context
In example `/examples/stereo.html`, when you apply effect the screen becomes black.

gchoqueux:
## Description
update to Three r102.1
## migration
- [x] Removed `renderTarget` and `forceClear` parameters from `WebGLRenderer.render()`. Please use `.setRenderTarget() `and `.clear()` instead before you perform the rendering. Be aware that it's now necessary to execute `renderer.setRenderTarget( null )` in order to unset an active render target.

- [x]  Removed .`activeCubeFace` and .`activeMipMapLevel` from `WebGLRenderTargetCube`. They are now parameters of `WebGLRenderer.setRenderTarget().`

- [x] In` WebGLRenderer.setViewport() `and` WebGLRenderer.setScissor()`, (x, y) is the coordinate of the lower left corner of the rectangular region.

- [x] `WebGLRenderer.getSize() `now requires a Vector2 argument.

- [x] `WebGLRenderer.getCurrentViewport()` now requires a Vector4 argument.

Fix #1044

### DIFF
--- a/examples/gpx.html
+++ b/examples/gpx.html
@@ -45,10 +45,12 @@
 
             // update the waypoint
             var distance, scale, point = new itowns.THREE.Vector3();
+            var size = new itowns.THREE.Vector2();
             function updatePointScale(renderer, scene, camera) {
                 point.copy(this.geometry.boundingSphere.center).applyMatrix4(this.matrixWorld);;
                 distance = camera.position.distanceTo(point);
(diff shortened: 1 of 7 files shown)


zarov:
There is a problem in the `stereo` example: when enabling the Stereo effect and disabling it, the viewport is wrong.

zarov (approved):

PR #1066, opened 2019-03-11
chore(docs): add generation of each property type in typedef

zarov:
There is no way to know the property type until now.


gchoqueux (approved):
thanks

PR #1068, opened 2019-03-11
Process and Layer imagery/elevation

gchoqueux:
## Description
Work on `process` and `Layer` imagery/elevation:
* Move (see delete) parts of the code into more appropriate classes or clearer functions.
* Extract common methods.
* Fix


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -54,6 +43,36 @@ function refinementCommandCancellationFn(cmd) {
     return !cmd.requester.material.visible;
 }
 
+function handlingError(err, node, layer, targetLevel, view) {
```
zarov: Couldn't this be moved to `TileMesh` or in a `ProcessHelper` file along with `buildCommand` below ?

src/Converter/convertToTile.js
```diff
@@ -45,12 +72,13 @@ export default {
...
-        // build tile
+        // build tile mesh
         geometry._count++;
         const material = new LayeredMaterial(layer.materialOptions);
         const tile = new TileMesh(geometry, material, layer, extent, level);
-        // TODO semble ne pas etre necessaire
-        tile.layers.set(layer.threejsLayer);
+
+        // TODO unecessary because is undefined...????
+        // tile.layers.set(layer.threejsLayer);
```
zarov, line 82: `threejsLayer` is used in `Picking` and for debug isn't it ? Do you have clue on how to maybe get rid of this property ?

zarov (commented):
Thanks for this first cleaning !

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: I move `handlingError` but not `buildCommand` is specify for each process

src/Converter/convertToTile.js
gchoqueux: It's used with visibility. This part needs to be fixed, because there's conflict with `object3d` added in `view.scene`;

src/Process/LayeredMaterialNodeProcessing.js
zarov: Thanks

zarov (approved):
LGTM

PR #1073, opened 2019-03-12
fix(process): add condition when getting extentsSource

zarov:
Fix a problem encountered in #1056 but already present in master.


gchoqueux:
fixed with #1068

PR #1074, opened 2019-03-12
Fix draco decoder for b3dm

gchoqueux:
Enable draco loader for b3dm


zarov (approved):
I don't like the copy/paste of the library, even if it is in the example and it is like that it is done in THREE. However the official package seems out of update, so I'm okay for now with validating it, but if it ever updates, we update too.

PR #1080, opened 2019-03-15
fix(controls): fix of removeEventListener in GlobeControls.dispose.

dorian-moffroid-prolexia:
## Description

The method dispose of the GlobeControls class does not remove mouse listeners.

## Your Environment
* Version used: 2.8.0 (but the current code seems have same issue)
* Browser Name and version: Chrome 73
* Operating System and version (desktop or mobile): Desktop Windows 10

## Context

For some needs of my project, I tryed to dispose the GlobeControls after some actions, so I cannot use the 'noControls' options. After the call of the dispose method, the GlobeControls continues to listen the mouse events and move the camera.

1. Adds the following line at the end of the sample script 'globe.html':
```
view.controls.dispose();
```
2. Runs the sample and tries to move the camera.
3. The GlobeControls still manipulates the camera.

After some code analysis, I found that many listeners are registered with a '.bind(this)' and unregistered by passing the function instance. Also, it seems that some handler are registered but not unregistered in the dispose method (fixed in my pull request).

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -77,9 +77,6 @@ if (enableTargetHelper) {
     helpers.target = new THREE.AxesHelper(500000);
 }
 
-// Handle function
-let _handlerMouseMove;
-let _handlerMouseUp;
 // current downed key
 let currentKey;
 
@@ -541,6 +538,8 @@ function GlobeControls(view, targetCoordinate, range, globeRadius, options = {})
         spherical.setFromVector3(targetPosition);
     };
 
+    const _onMouseMoveListener = onMouseMove.bind(this);
+    const _onMouseUpListener = onMouseUp.bind(this);
     function onMouseDown(event) {
         CameraUtils.stop(view, this.camera);
         player.stop().then(() => {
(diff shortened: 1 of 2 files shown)


src/Controls/GlobeControls.js
```diff
@@ -911,8 +932,8 @@ function GlobeControls(view, targetCoordinate, range, globeRadius, options = {})
         this.camera.layers.enable(layerTHREEjs);
     }
 
-    _handlerMouseMove = onMouseMove.bind(this);
-    _handlerMouseUp = onMouseUp.bind(this);
+    _handlerMouseMove = _onMouseMoveListener;
```
gchoqueux: Could you remove `_handlerMouseMove` and `_handlerMouseUp` and could you use `let _onMouseMoveListener` and `let _onMouseUpListener`  instead?

gchoqueux (commented):
Thanks, for your contribution. I have just one comment.

src/Controls/GlobeControls.js
dorian-moffroid-prolexia: I pushed a new commit one the same branch (its the first time I use PR so I hope that is the right way to do). I hope the modifications are what you expect.

Just one precision: I moved up the two `_onMouseMoveListener` and `_onMouseUpListener` (stayed as `const`) variable definitions just before the `onMouseDown` in order to respect the ESLint rule `no-use-before-define`.

gchoqueux (approved):
LGTM
Thanks for your contribution

src/Controls/GlobeControls.js
gchoqueux: This is good, that's what I wanted.

PR #1081, opened 2019-03-18
fix(controls): fix ORBIT touch event making North down.

dorian-moffroid-prolexia:
## Description

The globe manipulation with two fingers on a touch screen, for zooming, causes an inversion making the North down and the South up on the screen.

## Your Environment
* Version used: 2.8.0 (but the current code seems have same issue)
* Browser Name and version: Chrome 73
* Operating System and version (desktop or mobile): Desktop Windows 10 with a Touch Screen

After some code analysis, I found that the `states.ORBIT` state is not referenced in the `onTouchStart` function, and the rotateStart vector is not initialized in the same function but used in the `onTouchMove` function.


gchoqueux (approved):
Thanks for your contribution.
LGTM.

(I haven't the device to test your fix.)

PR #1084, opened 2019-03-26
Fix geoportail stream

gliegard:
## Description
- fix geoportail stream for french administrative boundaries
- add geoportail stream helper files, for old maps and Top25 map

### DIFF
--- a/examples/layers/JSONLayers/Administrative.json
+++ b/examples/layers/JSONLayers/Administrative.json
@@ -1,14 +1,14 @@
 {
   "id": "Administrative",
   "title": "Administrative delimitations",
-  "visible": false,
+  "visible": true,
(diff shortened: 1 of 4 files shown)


examples/layers/JSONLayers/Administrative.json
```diff
@@ -1,14 +1,14 @@
 {
   "id": "Administrative",
   "title": "Administrative delimitations",
-  "visible": false,
+  "visible": true,
```
gchoqueux, line 4: could you keep false?

gliegard: No, I think we can loose time with visible = false...

gchoqueux (approved):

PR #1085, opened 2019-03-26
fix(3dtiles): fix batch table support, for null object.

gliegard:
fix(3dtiles): fix batch table support, for null object.

Before this commit, null object in the input data was raising an error.
To fix it, I've changed null (object) to 'null' (string). So the UI shows, for example: property: null.


gchoqueux (approved):

gchoqueux:
Thanks, LGTM

PR #1091, opened 2019-04-02
refactor(parser): Class Feature

gchoqueux:
## Description
Add `class Feature` and `MultiFeature`.

This class improves and simplifies the construction and conversion of geographic data structures. It's an intermediary structure between **geomatic formats** and **THREE objects**.

The parsers transform the data into `MultiFeatures` and the converters convert them into `THREE.Mesh` or `THREE.Texture`.

For the moment, theses class are used in `GeoJsonParser`, in a future Pr, they'll used in `VectorTileParse`.

### DIFF
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -1,0 +1,264 @@
+import Extent from 'Core/Geographic/Extent';
+
+function defaultExtent(crs) {
+    return new Extent(crs, Infinity, -Infinity, Infinity, -Infinity);
+}
+
+function _extendBuffer(feature, size) {
+    feature.vertices.length += size * feature.size;
+    if (feature.normals) {
+        feature.normals.length = feature.vertices.length;
+    }
+}
+
+
+/**
+ * @property {Extent} extent - The 2D extent containing all the points
+ * composing the geometry.
+ * @property {Object[]} indices - Contains the indices
+ * that define the geometry. Objects stored in this array have two
+ * properties, an <code>offset</code> and a <code>count</code>. The offset
+ * is related to the overall number of vertices in the Feature.
+ *
+ * @property {Object} properties - Properties of the geometry. It can be
+ * anything specified in the GeoJSON under the <code>properties</code>
+ * property.
+ */
+class FeatureGeometry {
+    /**
+     * @param {Feature} feature geometry
+     */
+    constructor(feature) {
+        this.extent = feature.extent ? defaultExtent(feature.crs) : undefined;
+        this.indices = [];
+        this._feature = feature;
+        this.properties = {};
+    }
+    /**
+     * Add a new marker to indicate the starting of sub geometry and extends the vertices buffer.
+     * Then you have to push new the coordinates of sub geometry.
+     * The sub geometry stored in indices, see constructor for more information.
+     * @param {number} count count of vertices
+     */
+    startSubGeometry(count) {
+        const last = this.indices.length - 1;
+        const offset = last > -1 ?
+            this.indices[last].offset + this.indices[last].count :
+            this._feature.vertices.length / this.size;
+        this.indices.push({ offset, count });
+        _extendBuffer(this._feature, count);
+    }
+
+    /**
+     * After you have pushed new the coordinates of sub geometry without <code>startSubGeometry</code>,
+     * this function close sub geometry. The sub geometry stored in indices,
+     * see constructor for more information.
+     * @param {number} count count of vertices
+     */
+    closeSubGeometry(count) {
+        const last = this.indices.length - 1;
+        const offset = last > -1 ?
+            this.indices[last].offset + this.indices[last].count :
+            this._feature.vertices.length / this.size - count;
+        this.indices.push({ offset, count });
+    }
+    /**
+     * Push new coordinates in vertices buffer.
+     * @param {Coordinates} coord The coordinates to push.
+     */
+    pushCoordinates(coord) {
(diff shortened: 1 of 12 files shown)


src/Core/Feature.js
```diff
@@ -0,0 +1,150 @@
...
+
+function push3DValues(feature, value0, value1, value2) {
+    feature.vertices[feature._pos++] = value0;
+    feature.vertices[feature._pos++] = value1;
+    feature.vertices[feature._pos++] = value2;
+}
+
+export default class Feature {
+    constructor(type, crs, options) {
+        this.type = type;
```
zarov: Maybe you can use an enum for this ?

src/Core/Feature.js
```diff
@@ -0,0 +1,150 @@
...
+        this.geometry = [];
+        this.vertices = [];
+        this.vertices.crs = crs;
```
zarov: Where is this used ? We shouldn't add properties on a native type.

src/Core/Feature.js
```diff
@@ -0,0 +1,154 @@
+import Extent from 'Core/Geographic/Extent';
+
+function defaultExtent(crs) {
+    return new Extent(crs, Infinity, -Infinity, Infinity, -Infinity);
+}
+
+function unionExtent(extent1, extent2) {
```
zarov: Maybe this should be move in `Extent`, no ?

src/Core/Feature.js
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        this._feature._pushValues(this._feature, coord._values[0], coord._values[1], coord._values[2]);
+        // expand extent if present
+        if (this.extent) {
+            this.extent.expandByCoordinates(coord);
+        }
+    }
+
+    pushCoordinatesValues(value0, value1, value2) {
```
zarov: So I get that this is the method using `Extent#expandByValuesCoordinates`, but now how do you see the usage of this ? Can't we only allow the push of complete coordinates ?

src/Core/Feature.js
```diff
@@ -0,0 +1,154 @@
...
+        this._feature._pushValues(this._feature, value0, value1, value2);
+        // expand extent if present
+        if (this.extent) {
+            this.extent.expandByValuesCoordinates(value0, value1, value2);
+        }
+    }
+
+    get size() {
```
zarov, line 103: Use it above, or I don't see the point of this.

src/Core/Feature.js
```diff
@@ -0,0 +1,154 @@
...
+            unionExtent(this.extent, geometry.extent);
+        }
+    }
+
+    get geometryCount() {
+        return this.geometry.length;
+    }
+}
+
+export class MultiFeatures {
```
zarov: What about using `FeatureCollection` like it is now ?

zarov (commented):
Thanks for doing this ! Aside the comments, there are two things missing: the documentation and some tests. Could you at least write the documentation ? I can help if you want. You can also update the documentation of `GeoJsonParser`.

src/Core/Feature.js
gchoqueux: removed

src/Core/Feature.js

src/Core/Feature.js
gchoqueux: This function will be used when the projection is identical and there's no conversion. This avoids having a useless `Coordinates` object in the parser. This will be used in `VectorTileParser`.

src/Core/Feature.js
gchoqueux: yes

src/Core/Feature.js

src/Core/Feature.js
gchoqueux: added

gchoqueux:
@zarov I have added some tests and documentation.

src/Core/Feature.js
```diff
@@ -0,0 +1,260 @@
...
+        if (this.extent) {
+            this.extent.expandByCoordinates(coord);
+        }
+    }
+
+    /**
+    * Push new values coordinates in vertices buffer.
+    * No geographical conversion is made or the normal doesn't stored.
+    *
+    * @param {number} value0 The longitude coordinate.
```
zarov: Why not rename `lat, long, alt` then ?

src/Core/Feature.js
```diff
@@ -0,0 +1,260 @@
...
+    feature.vertices[feature._pos++] = value1;
+}
+
+function push3DValues(feature, value0, value1, value2) {
+    feature.vertices[feature._pos++] = value0;
+    feature.vertices[feature._pos++] = value1;
+    feature.vertices[feature._pos++] = value2;
+}
+
+const FEATURE_TYPES = ['point', 'line', 'polygon'];
```
zarov: I was thinking something more like

```js
export const FEATURE_TYPES = {
    POINT: 'point',
    MULTIPOINT: 'multipoint',
    LINESTRING: 'linestring',
    ....
}   
```

But if you are happy with this, it is ok for me.

zarov (commented):
You need to change the type `line` to `linestring`.

src/Core/Feature.js
gchoqueux: done

src/Core/Feature.js
gchoqueux: done

zarov (approved):
Thanks, all seems good !

PR #1098, opened 2019-04-09
Use Feature class in vector tile parser.

gchoqueux:
feature(parser): parse vector tile in Feature without coordinates conversion.

* Use class Feature in `VectorTileParser` to get directly
  `CollectionFeature` and avoid `geojson` step.

* Add transformation in `CollectionFeature` to avoid coordinates conversion.

### DIFF
--- a/src/Converter/Feature2Texture.js
+++ b/src/Converter/Feature2Texture.js
@@ -1,90 +1,95 @@
 import * as THREE from 'three';
+import { FEATURE_TYPES } from 'Core/Feature';
 import Extent from 'Core/Geographic/Extent';
-import Coordinates from '../Core/Geographic/Coordinates';
+import Coordinates from 'Core/Geographic/Coordinates';
 
 const _extent = new Extent('EPSG:4326', [0, 0, 0, 0]);
 
-const pt = new THREE.Vector2();
-
-function _moveTo(ctx, x, y, scale, origin) {
-    pt.x = x - origin.x;
-    pt.y = y - origin.y;
-    pt.multiply(scale);
-    ctx.moveTo(Math.round(pt.x), Math.round(pt.y));
-}
-
-function _lineTo(ctx, x, y, scale, origin) {
-    pt.x = x - origin.x;
-
-    pt.y = y - origin.y;
-    pt.multiply(scale);
-    ctx.lineTo(Math.round(pt.x), Math.round(pt.y));
-}
-
-function drawPolygon(ctx, vertices, indices = [{ offset: 0, count: 1 }], origin, scale, properties, style = {}, size) {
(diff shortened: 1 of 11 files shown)


src/Converter/Feature2Texture.js
```diff
@@ -143,12 +151,23 @@ export default {
...
+            const t = collection.translation;
+            const s = collection.scale;
+            extentTransformed.transformedCopy(t, s, ex);
+
+            scale.set(ctx.canvas.width, ctx.canvas.width).divide(dimension);
+            origin.set(extent.west, extent.south).add(t).multiply(scale).negate();
+            ctx.setTransform(scale.x / s.x, 0, 0, scale.y / s.y, origin.x, origin.y);
+
+            // to scale line width and radius circle
+            ctx._mscale = s.x / scale.x;
```
zarov: Isn't there another way to store this ?

src/Core/Feature.js
```diff
@@ -93,7 +98,7 @@ class FeatureGeometry {
         this._feature._pushValues(this._feature, long, lat, alt);
         // expand extent if present
         if (this.extent) {
```
zarov: Don't forget to change this as well

src/Parser/VectorTileParser.js
```diff
@@ -1,172 +1,125 @@
...
-            type,
-            coordinates: coords,
-        },
-        properties: this.properties,
-    };
-
-    if ('id' in this) {
-        result.id = this.id;
-    }
+const mergesType = [FEATURE_TYPES.POINT, FEATURE_TYPES.LINE, FEATURE_TYPES.POLYGON];
```
zarov: This can be avoided by changing `FEATURE_TYPES` to numbers instead of strings.

src/Parser/VectorTileParser.js
```diff
@@ -1,172 +1,125 @@
+import * as THREE from 'three';
 import Protobuf from 'pbf';
 import { VectorTile } from '@mapbox/vector-tile';
-import GeoJsonParser from 'Parser/GeoJsonParser';
```
zarov, line 3: Good deletion !

zarov (commented):
A few comments, overall it is good.

src/Converter/Feature2Texture.js
gchoqueux: done

src/Core/Feature.js
gchoqueux: done

src/Parser/VectorTileParser.js
gchoqueux: done

zarov (approved):

PR #1100, opened 2019-04-09
refactor(core): use class Es6 for Builder tile and add factory of Tile.

gchoqueux:
## Description
* add a new class `TileBuilder`, it extends ellipsoid and planar tile builders.
* use Class es6 for geometry `TileBuilder`.
* move buffer and geometry tile cache in `TileBuilder`.
* Use `TileBuilder` like a factory of tile with `builder.newTileGeometry`.

## Motivation
Prepare to introduce **worker** in **iTowns**.

### DIFF
--- a/src/Core/Prefab/Globe/BuilderEllipsoidTile.js
+++ b/src/Core/Prefab/Globe/BuilderEllipsoidTile.js
@@ -6,122 +6,118 @@ import Extent from 'Core/Geographic/Extent';
 
 const axisZ = new THREE.Vector3(0, 0, 1);
 const axisY = new THREE.Vector3(0, 1, 0);
+const quatToAlignLongitude = new THREE.Quaternion();
+const quatToAlignLatitude = new THREE.Quaternion();
 
 function WGS84ToOneSubY(latitude) {
     return 1.0 - Projection.WGS84ToY(latitude);
 }
 
-function BuilderEllipsoidTile() {
-    this.tmp = {
-        coords: [
-            new Coordinates('EPSG:4326', 0, 0),
-            new Coordinates('EPSG:4326', 0, 0)],
-        position: new THREE.Vector3(),
-    };
+class BuilderEllipsoidTile {
+    constructor() {
+        this.type = 'e';
+        this.tmp = {
+            coords: [
+                new Coordinates('EPSG:4326', 0, 0),
+                new Coordinates('EPSG:4326', 0, 0)],
+            position: new THREE.Vector3(),
+        };
+    }
+    // prepare params
+    // init projected object -> params.projected
+    prepare(params) {
+        params.nbRow = 2 ** (params.level + 1.0);
 
-    this.type = 'e';
+        var st1 = WGS84ToOneSubY(params.extent.south);
+
+        if (!isFinite(st1)) { st1 = 0; }
+
+        var sizeTexture = 1.0 / params.nbRow;
+
+        var start = (st1 % (sizeTexture));
+
+        params.deltaUV1 = (st1 - start) * params.nbRow;
+
+        // transformation to align tile's normal to z axis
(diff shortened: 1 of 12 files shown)


src/Core/Prefab/computeBufferTileGeometry.js
```diff
@@ -0,0 +1,218 @@
...
+    //        * v = textureid + v in builder texture
+    this.uv = {
+        wgs84: null,
+        pm: null,
+    };
+}
+
+export default function computeBuffers(params) {
+    // Create output buffers.
+    const outBuffers = new Buffers();
```
zarov: What about using an object instead ? You could have:
```js
const outBuffers = {
    index: null,
    position: null,
    normal: null,
    uv: {
        wgs84: null,
        pm: null,
    }
};
```

src/Core/Prefab/Planar/PlanarTileBuilder.js
```diff
@@ -1,73 +1,72 @@
...
+        this.type = 'p';
+        this.tmp = {
+            coords: new Coordinates('EPSG:4326', 0, 0),
+            position: new THREE.Vector3(),
+            normal: new THREE.Vector3(0, 0, 1),
+        };
+    }
+    // prepare params
+    // init projected object -> params.projected
+    Prepare(params) {
```
zarov: Can you take this opportunity to also rename methods of this class to lowercase as first letter ?

src/Core/Prefab/Globe/BuilderEllipsoidTile.js
```diff
@@ -6,122 +7,118 @@ import Extent from 'Core/Geographic/Extent';
...
+        this.tmp = {
+            coords: [
+                new Coordinates('EPSG:4326', 0, 0),
+                new Coordinates('EPSG:4326', 0, 0)],
+            position: new THREE.Vector3(),
+        };
+    }
+    // prepare params
+    // init projected object -> params.projected
+    Prepare(params) {
```
zarov: Same as for PlanarTileBuilder please.

src/Core/Prefab/Planar/PlanarTileBuilder.js
```diff
@@ -1,73 +1,72 @@
+/* eslint-disable class-methods-use-this */
```
zarov: This may be outside the scope of this PR, but I think we should drop this rule: I don't see the value in having always `this` used in class methods.

src/Core/Prefab/TileBuilder.js
```diff
@@ -0,0 +1,60 @@
+import * as THREE from 'three';
+import TileGeometry from 'Core/TileGeometry';
+import Cache from 'Core/Scheduler/Cache';
+import computeBuffers from 'Core/Prefab/computeBufferTileGeometry';
+
+const _computeBuffers = computeBuffers;
```
zarov: Why do you need to do this ?

zarov (commented):
Thanks for keeping doing the transition to ES6 !

src/Core/Prefab/computeBufferTileGeometry.js
gchoqueux: yes

src/Core/Prefab/Planar/PlanarTileBuilder.js
gchoqueux: ok

src/Core/Prefab/Planar/PlanarTileBuilder.js
gchoqueux: rule dropped

src/Core/Prefab/TileBuilder.js
gchoqueux: removed

zarov (approved):

PR #1102, opened 2019-04-11
test: add tests for sources

zarov:
Add more tests for sources, and add the sinon dependency to test the console output (to cover more branch, but I'm not convinced by the utility of this).

### DIFF
--- a/test/unit/source.js
+++ b/test/unit/source.js
@@ -2,11 +2,11 @@ import assert from 'assert';
 import fs from 'fs';
 import Path from 'path';
 import Source from 'Source/Source';
+import WFSSource from 'Source/WFSSource';
 import WMTSSource from 'Source/WMTSSource';
 import WMSSource from 'Source/WMSSource';
 import TMSSource from 'Source/TMSSource';
 import FileSource from 'Source/FileSource';
-import Fetcher from 'Provider/Fetcher';
 import Extent from 'Core/Geographic/Extent';
 
 function defer() {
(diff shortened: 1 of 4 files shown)


test/unit/source.js
```diff
@@ -30,104 +31,260 @@ global.URL = function URL() {
...
-
-describe('Source', function () {
-    it('Should instance and throw error for Source', function () {
-        const source = new Source({ url: 'http://'  });
-        assert.throws(source.urlFromExtent, Error);
-        assert.throws(source.extentInsideLimit, Error);
-        assert.throws(source.extentsInsideLimit, Error);
+describe('Sources', function () {
+    beforeEach(function () {
+        sinon.spy(console, 'warn');
```
gchoqueux: The test fails if `sinon` finds a warn console?

package.json
```diff
@@ -80,6 +80,7 @@
     "proj4": "^2.5.0",
     "puppeteer": "^1.14.0",
     "replace-in-file": "^3.4.4",
+    "sinon": "^7.3.1",
```
gchoqueux: * You use Sinon to spy Console, What do you test exactly?

* I don't know if [startStaticFileServer](https://github.com/iTowns/itowns/blob/master/test/functional/bootstrap.js#L22) is fake server? if yes we could use [sinon fake server](https://sinonjs.org/releases/v7.3.1/fake-xhr-and-server/#fake-server).

* Sinon could be used to make mocks, too.

gchoqueux (commented):
It's good to add tests.

test/unit/source.js
zarov: Before each test, `sinon` is spying on what we tell it: here it is `console.warn`. It modifies the `console.warn` method, and allow us to do some tests later on (see below `console.warn.calledOnce` for example).

package.json
zarov: Yes, `sinon` seems to be quite complete. I'm looking more into it then.

zarov: I'm dropping `sinon` and letting the PR as it is without it.

gchoqueux (approved):

PR #1109, opened 2019-04-16
doc(tuto): update geometry layer using sources object instead of configuration

zarov:
Quick update of the tutorial


gchoqueux (approved):
thanks

PR #1111, opened 2019-04-16
chore: fix useless or missing conditions.

gchoqueux:
## Description
finds with Deepscan.

### DIFF
--- a/src/Layer/LayerUpdateStrategy.js
+++ b/src/Layer/LayerUpdateStrategy.js
@@ -66,7 +66,6 @@ export function chooseNextLevelToFetch(strategy, node, nodeLevel = node.level, c
             // default strategy
             case STRATEGY_MIN_NETWORK_TRAFFIC:
             default:
-                nodeLevel = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : nodeLevel;
                 nextLevelToFetch = _minimizeNetworkTraffic(node, nodeLevel, currentLevel);
         }
     }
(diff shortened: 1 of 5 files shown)


src/Layer/LayerUpdateStrategy.js
```diff
@@ -66,7 +66,6 @@ export function chooseNextLevelToFetch(strategy, node, nodeLevel = node.level, c
             // default strategy
             case STRATEGY_MIN_NETWORK_TRAFFIC:
             default:
-                nodeLevel = failureParams ? Math.ceil((currentLevel + failureParams.targetLevel) / 2) : nodeLevel;
```
zarov, line 69: Are you sure this is never used ?

src/Provider/3dTilesProvider.js
```diff
@@ -101,8 +101,9 @@ export function $3dTilesIndex(tileset, baseURL) {
 
         // getBox only use inverseTileTransform for volume.region so let's not
         // compute the inverse matrix each time
+        // Assumes that node.boundingVolume is defined if node.viewerRequestVolume is undefined
```
zarov, line 104: What is the confidence here ?

zarov (commented):
Thanks for noticing those !

src/Layer/LayerUpdateStrategy.js
gchoqueux: before we test if `failureParams` is `undefined`.

zarov: Didn't see it, my bad !

src/Provider/3dTilesProvider.js
gchoqueux: I think is the specification, and after in the function `getBox` ([here](https://github.com/iTowns/itowns/blob/7d3644e9d9288523d06fbb7fc347c43974bed97a/src/Provider/3dTilesProvider.js#L115))  `node.boundingVolume` is considered defined.

zarov: Ok

zarov (approved):

PR #1114, opened 2019-04-17
refacto(crs): create a CRS module

zarov:
All methods exported from Coordinates have been moved to a single CRS
module, to simplify readability of Coordinates file.

### DIFF
--- a/src/Renderer/OBB.js
+++ b/src/Renderer/OBB.js
@@ -1,7 +1,8 @@
 import * as THREE from 'three';
 import TileGeometry from 'Core/TileGeometry';
 import BuilderEllipsoidTile from 'Core/Prefab/Globe/BuilderEllipsoidTile';
-import Coordinates, { crsIsGeocentric } from 'Core/Geographic/Coordinates';
+import Coordinates from 'Core/Geographic/Coordinates';
+import CRS from 'Core/Geographic/Crs';
 
 // get oriented bounding box of tile
 const builder = new BuilderEllipsoidTile();
@@ -138,7 +139,7 @@ class OBB extends THREE.Object3D {
             this.position.copy(position);
             this.quaternion.copy(quaternion);
             this.updateMatrixWorld(true);
-        } else if (!extent.isTiledCrs() && crsIsGeocentric(extent.crs)) {
+        } else if (!extent.isTiledCrs() && CRS.isMetricUnit(extent.crs)) {
             extent.center(coord).xyz(this.position);
             extent.dimensions(dimension);
             size.set(dimension.x, dimension.y, Math.abs(maxHeight - minHeight));
--- a/docs/config.json
+++ b/docs/config.json
@@ -21,6 +21,7 @@
 
             "Geographic": [
                 "Coordinates",
+                "CRS",
                 "Extent",
                 "OrientationUtils",
                 "OBB"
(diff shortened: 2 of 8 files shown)


src/Core/Geographic/CRS.js
```diff
@@ -0,0 +1,122 @@
+import proj4 from 'proj4';
```
gchoqueux: I prefer `Crs.js`

src/Core/Geographic/CRS.js
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Assert that the CRS is geographic.
+     *
+     * @param {string} crs - The CRS to validate.
+     * @return {boolean}
+     * @throws {Error} if the CRS is not valid.
+     */
+    isGeographic(crs) {
+        return (toUnitWithError(crs) != UNIT.METER);
```
gchoqueux: could replace by `toUnitWithError(crs) != UNIT.DEGREE` ?

src/Core/Geographic/CRS.js
```diff
@@ -0,0 +1,122 @@
...
+    },
+
+    /**
+     * Assert that the CRS is geocentric.
+     *
+     * @param {string} crs - The CRS to validate.
+     * @return {boolean}
+     * @throws {Error} if the CRS is not valid.
+     */
+    isGeocentric(crs) {
```
gchoqueux: Geocentric doesn't mean that the unit is metric. Where is used this function?
By example 'EPSG:4978' is geocentric, but 'EPSG:3857' isn't geocentric while they are in metric unit.

src/Core/Geographic/CRS.js
```diff
@@ -0,0 +1,122 @@
...
+     */
+    toUnit,
+
+    /**
+     * Is the CRS EPSG:4326 ?
+     *
+     * @param {string} crs - The CRS to test.
+     * @return {boolean}
+     */
+    is4326,
```
gchoqueux: if `is4326` isn't used in another private functions, you could move the code function here.

gchoqueux: same comment for the other functions.

gchoqueux (changes requested):

src/Core/Geographic/CRS.js
zarov: Changed

src/Core/Geographic/CRS.js
zarov: It is used in `Coordinates` and `OBB`; I renamed it from `crsIsGeocentric`.

src/Core/Geographic/CRS.js
zarov: `is4326` is used in `reasonnableEpsilon` and `toUnit` is used in `toUnitWithError`, so I can't change anything here.

src/Core/Geographic/CRS.js
zarov: Done.

zarov:
Updated, I have removed `isGeocentric` whenever it was called.

src/Renderer/OBB.js
```diff
@@ -138,7 +139,7 @@ class OBB extends THREE.Object3D {
             this.position.copy(position);
             this.quaternion.copy(quaternion);
             this.updateMatrixWorld(true);
-        } else if (!extent.isTiledCrs() && crsIsGeocentric(extent.crs)) {
+        } else if (!extent.isTiledCrs() && CRS.toUnit(extent.crs) == CRS.UNIT.METER) {
```
gchoqueux: could you add `CRS.isMetricUnit(crs)`?

zarov: Done

gchoqueux (approved):

PR #1121, opened 2019-04-19
Style

gchoqueux:
## Description
refactor(core): add style class to define `Feature` style.

* Improvement of the vector tile style parsing.

## Motivation and Context
* Begin to unify style management in itowns.

Master:
Capture du 2019-04-19 16-56-15

PR:
Capture du 2019-04-19 16-55-08


src/Converter/Feature2Texture.js
```diff
@@ -34,94 +34,86 @@ function _drawPolygon(ctx, vertices, indices, properties = {}, style, size, exte
...
-    ctx.strokeStyle = style.stroke || 'red';
-    ctx.stroke();
+    ctx.arc(x, y, (style.point.radius || 3.0) * invCtxScale, 0, 2 * Math.PI, false);
+    if (style.point.color) {
+        ctx.fillStyle = style.point.color;
+        ctx.fill();
+    }
+    if (style.point.line) {
+        ctx.lineWidth = (style.point.width || 1.0) * invCtxScale;
+        ctx.strokeStyle = style.point.line;
```
zarov, line 86: No more default color ?

src/Core/Style.js
```diff
@@ -0,0 +1,113 @@
...
+     * @param {string} [params.stroke.color] stroke color string css.
+     * @param {number} [params.stroke.opacity] stroke opacity.
+     * @param {number} [params.stroke.width] stroke line width.
+     * @param {object} [params.point] point style.
+     * @param {THREE.Vector3} [params.point.color] point
+     * @param {string} [params.point.color] point color string css.
+     * @param {number} [params.point.opacity] point opacity.
+     * @param {number} [params.point.radius] point line radius
+     */
+    constructor(params = {}) {
```
zarov, line 41: Some parameters are in double here. Also you could simply write that `params` is an object that can contain all properties of `Style`, and move the list of parameters above, something like the documentation of `Source`.

src/Converter/Feature2Texture.js
```diff
@@ -34,94 +34,86 @@ function _drawPolygon(ctx, vertices, indices, properties = {}, style, size, exte
...
+    const alpha = style.stroke.opacity || 1.0;
     if (alpha !== ctx.globalAlpha && typeof alpha == 'number') {
         ctx.globalAlpha = alpha;
     }
 }
 
 function drawPoint(ctx, x, y, style = {}, invCtxScale) {
     ctx.beginPath();
-    const opacity = style.pointOpacity || 1.0;
+    const opacity = style.point.opacity || 1.0;
```
zarov: Let's say for some reason I have a full transparent point: this condition is not working for me here. This needs to move to something like `style.point.opacity == undefined ? 1.0 : style.point.opacity`. Or maybe you could do something about it directly in `Style`, I dunno.

src/Core/Style.js
```diff
@@ -0,0 +1,113 @@
+import { FEATURE_TYPES } from 'Core/Feature';
+
+function rgba2rgb(orig) {
+    if (orig && orig.includes('rgba')) {
+        const rgb = orig.replace(/\s/g, '').match(/^rgba?\((\d+),(\d+),(\d+),(\d\.\d)/i);
```
zarov: You can drop the `rgba` match, as well as `hsla` below. Also opacity should be `[\d.]+`.

src/Core/Style.js
```diff
@@ -0,0 +1,113 @@
...
+        return (rgb && rgb.length === 5) ? { color: `rgb(${rgb[1]}, ${rgb[2]}, ${rgb[3]})`, opacity: rgb[4] } : { style: orig, opacity: 1.0 };
+    } else if (orig && orig.includes('hsla')) {
+        const hsl = orig.replace(/\s/g, '').match(/^hsla?\((\d+),(\d+)%,(\d+)%,(\d\.\d)/i);
+        return (hsl && hsl.length === 5) ? { color: `rgb(${hsl[1]}, ${hsl[2]}, ${hsl[3]})`, opacity: hsl[4] } : { style: orig, opacity: 1.0 };
```
zarov: You can't say that `h=r` `s=g` and `l=b`, some conversion needs to be done !

src/Parser/GeoJsonParser.js
```diff
@@ -154,6 +157,7 @@ function jsonFeaturesToFeatures(crsIn, crsOut, jsonFeatures, filteringExtent, op
         features.updateExtent();
     }
 
+    // console.log('features', features);
```
zarov: To remove

src/Parser/VectorTileParser.js
```diff
@@ -3,12 +3,29 @@ import Protobuf from 'pbf';
...
+import Style from 'Core/Style';
 
 const VectorTileFeature = { types: ['Unknown', 'Point', 'LineString', 'Polygon'] };
 const globalExtent = new Vector3(worldDimension3857.x, worldDimension3857.y, 1);
 const lastPoint = new Vector2();
 const firstPoint = new Vector2();
-
+const styleCache = new Map();
+
+function parseStyle(vtFeature, layer) {
```
zarov: Quick link as a comment to the specification of a VectorTile style ? Thanks

zarov (changes requested):
Looks good and simple, thanks !

src/Converter/Feature2Texture.js
gchoqueux: For the moment, no, because if `style.point.line == undefined` then the line will not drawn.

src/Converter/Feature2Texture.js
gchoqueux: yes, thanks

src/Core/Style.js
gchoqueux: oopps

src/Parser/GeoJsonParser.js
gchoqueux: removed

src/Parser/VectorTileParser.js
gchoqueux: done

src/Core/Style.js
zarov: You still need to change the documentation here.

src/Core/Style.js
```diff
@@ -0,0 +1,113 @@
+import { FEATURE_TYPES } from 'Core/Feature';
+
+function rgba2rgb(orig) {
+    if (orig && orig.includes('rgba')) {
+        const rgb = orig.replace(/\s/g, '').match(/(\d+),(\d+),(\d+),(\d\.\d)/i);
+        return (rgb && rgb.length === 5) ? { color: `rgb(${rgb[1]}, ${rgb[2]}, ${rgb[3]})`, opacity: rgb[4] } : { style: orig, opacity: 1.0 };
+    } else if (orig && orig.includes('hsla')) {
+        const hsl = orig.replace(/\s/g, '').match(/(\d+),(\d+)%,(\d+)%,(\d\.\d)/i);
```
zarov: You need to add a wider range for the regex, see here https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#HSL_colors

src/Core/Style.js
```diff
@@ -0,0 +1,113 @@
+import { FEATURE_TYPES } from 'Core/Feature';
+
+function rgba2rgb(orig) {
+    if (orig && orig.includes('rgba')) {
+        const rgb = orig.replace(/\s/g, '').match(/(\d+),(\d+),(\d+),(\d\.\d)/i);
```
zarov: You need to add a wider range for the regex, see here https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#RGB_colors (you're missing the `%` for example.

zarov (commented):
Almost good

gchoqueux:
Pr updated

zarov (approved):
Looks good for a first version, let's go further in this direction ! Thanks for the hard work.

PR #1123, opened 2019-04-25
chore(md): add some testing tips for development

zarov:
Added some commands I'm using for development, I find it useful, maybe it can interest someone ?


CODING.md
```diff
@@ -64,6 +64,20 @@ Note: Chrome in headless mode doesn't support the WebGL `EXT_frag_depth`
...
+### Useful commands for continuous testing
+
+If you wish to have unit tests to continuously run, you can append to your
+running command `-- --watch`. You will be getting in this case `npm run
+test-unit -- --watch`.
+
+If you want to work on a single test and debug it, without having all the extra
+output, you can use this command:
+```sh
+./node_modules/.bin/cross-env BABEL_DISABLE_CACHE=1 \
```
tbroyer: How about `npx cross-env` rather than `./node_modules/.bin/cross-env` (and similarly for `mocha`)

CODING.md
```diff
@@ -64,6 +64,20 @@ Note: Chrome in headless mode doesn't support the WebGL `EXT_frag_depth`
...
+  ./node_modules/.bin/mocha --require @babel/register \
+  test/unit/3dtiles.js -- --watch # replace 3dtiles.js with the correct filename
```
tbroyer: Would it then be a good idea to add a `base-test-unit` NPM script that would call mocha without any "target", and rewrite `test-unit` as: `npm run base-test-unit -- test/unit`? then this command could become `npm run base-test-unit -- --watch test/unit/3dtiles.js`.

CODING.md
zarov: Oh I finally understand what `npx` is, thanks :smile: Will do

CODING.md
zarov: Yeah why not, this seems like a good idea

gchoqueux:
This Pr could be merged?

zarov:
I need to get back to it first, I think I wanted to change things.

zarov:
@gchoqueux this is good to merge

gchoqueux (approved):

PR #1125, opened 2019-04-29
doc: add the support of quotes to write code strings in documentation

zarov:
We can use `` instead of `<code></code>` to encapsulate some code when
writing documentation.

No changes done in the content, I only realigned all impacted documentation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -11,8 +11,10 @@ tutorials.
 ### API reference
 
 The API reference is written in the source of itowns, using
-[JSDoc](http://usejsdoc.org/). This is a classic documentation, explaining all
-exposed methods and classes in itowns, and some internal others.
+[JSDoc](http://usejsdoc.org/), along with support for the
+[Markdown](https://commonmark.org/help/) syntax. This is a
+classic documentation, explaining all exposed methods and classes in itowns, and
+some internal others.
 
 When documenting something, don't forget to check the presence of your file
 inside `docs/config.json`.
--- a/docs/config.json
+++ b/docs/config.json
@@ -103,5 +103,6 @@
     "source": {
         "include": [ "src" ],
         "exclude": [ "src/ThreeExtended" ]
-    }
+    },
+    "plugins": ["plugins/markdown"]
 }
--- a/docs/static/styles/itowns.css
+++ b/docs/static/styles/itowns.css
@@ -163,3 +163,7 @@ p, ul, ol {
     max-width: 780px;
     text-align: justify;
 }
+
+.params p {
+    display: inline;
+}
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -31,8 +31,8 @@ const vectorZero = new THREE.Vector3();
  *
  * @typedef {Object} PlanarControls~PlanarControlsOptions
  *
- * @property {boolean} [enableRotation=true] Enable the rotation with the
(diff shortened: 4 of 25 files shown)


docs/publish.js
```diff
@@ -80,6 +80,7 @@ function generate(title, content, filename, template) {
     const outpath = path.join(outDir, filename);
     let html = view.render(template || 'page.tmpl', { env, title, content });
     html = helper.resolveLinks(html);
+    html = html.replace(/`([\S ]*)`/g, '<code>$1</code>');
```
tbroyer: This should use [_non-greedy_ matching](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers) (i.e. `*?` rather than `*`), otherwise something like
```
Some `text` with `two` code snippets
```
will be replaced by
```
Some <code>text` with `two</code> code snippets
```

tbroyer (changes requested):
What's the impact when looking at the docs from an editor? Will it also process the backticks as if it were markdown? (might depend on the editor)

How about going all in and use the JSDoc Markdown plugin? http://usejsdoc.org/plugins-markdown.html

Btw, use _backticks_ rather than _quotes_ in the commit message.

Also, shouldn't docs/README be updated to reflect that change?

docs/publish.js
zarov: Ah yes, in fact I used `/<\/code>/` when replacing them by a ` ! Changing that

zarov: Yep, thanks for noticing this

zarov:
I don't know about the part from an editor.

I removed what I've done, in favor of the markdown plugin. I thought it was a plugin that needed to be installed, that's why I didn't want it to be added. The change has been reflected in the readme as well.

docs/README.md
```diff
@@ -11,8 +11,10 @@ tutorials.
 ### API reference
 
 The API reference is written in the source of itowns, using
-[JSDoc](http://usejsdoc.org/). This is a classic documentation, explaining all
-exposed methods and classes in itowns, and some internal others.
+[JSDoc](http://usejsdoc.org/), along with support for the
+[Markdown](https://daringfireball.net/projects/markdown/) syntax. This is a
```
tbroyer: Given that marked.js (used by JSDoc) supports CommonMark and (more importantly I'd say) GitHub-Flavored Markdown, I'm not sure about this link…

tbroyer (approved):

tbroyer:
> I don't know about the part from an editor.

I know that at least VS Code supports Markdown in JSDoc: https://code.visualstudio.com/updates/v1_30#_highlighting-of-markdown-elements-in-jsdoc-comments (note the part _“This inline highlighting better matches VS Code's Markdown rendering of JSDocs for hovers and IntelliSense documentation.”_)

docs/README.md
zarov: I don't think it changes a lot in this context, but I concur that the CommonMark cheatsheet is friendlier (https://commonmark.org/help/)

tbroyer: I think the main difference with "original markdown.pl" is that CommonMark and GFM both support fenced code blocks (using three backticks as the fence; CommonMark also supports three tildes), with language tag for syntax highlighting.

tbroyer (approved):

docs/README.md
zarov: Yes, but in our case it may be better to use `@example` I think, but we never know ;)

PR #1126, opened 2019-04-30
Es6 coordinates

zarov:
The long awaited refactor of Coordinates to ES6 class, along with some 
optimisations in it. Now the CRS is no longer checked everywhere when 
converting a coordinate, or accessing properties fro example. It is only
checked when the Coordinates is created. This is a debatable choice, but
I can assure you that we used to check the CRS way too much for nothing.
I also got rid of the `Float64Array` containing the values, in profit of more
direct properties (see this benchmark https://jsbench.me/40jumfag6g/1).
Otherwise it is quite the same as before.

BREAKING CHANGE:
* `x()`, `y()`, `z()`, `latitude()`, `longitude()` and `altitude()` have
been replaced by getters of the same names
* `setAltitude()` has been replaced by a setter on `altitude`
* the validity of a CRS is no longer check when doing conversion using
`as()`, so it needs to be checked beforehand if needed

### DIFF
--- a/examples/collada.html
+++ b/examples/collada.html
@@ -57,7 +57,7 @@
             }
             itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
-            
+
             // ThreeLoader can load each format proposed in ThreeJs examples loaders : https://github.com/mrdoob/three.js/tree/dev/examples/js/loaders
             var promiseCollada = ThreeLoader.load('Collada', 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/collada/building.dae')
             .then(collada => {
@@ -67,7 +67,7 @@
                 var coord = new itowns.Coordinates('EPSG:4326', 4.2165, 44.844, 1417);
                 var colladaID = view.mainLoop.gfxEngine.getUniqueThreejsLayer();
 
-                model.position.copy(coord.as(view.referenceCrs).xyz());
+                model.position.copy(coord.as(view.referenceCrs));
                 // align up vector with geodesic normal
                 model.lookAt(model.position.clone().add(coord.geodesicNormal));
                 // user rotate building to align with ortho image
--- a/examples/js/OrientedImageHelper.js
+++ b/examples/js/OrientedImageHelper.js
@@ -4,7 +4,7 @@
 // set object ENH orientation: X to the east, Y (green) to the north, Z (blue) look to the sky.
 function placeObjectFromCoordinate(object, coord) {
     // set object position to the coordinate
-    coord.xyz(object.position);
+    coord.toVector3(object.position);
     // set ENH orientation, looking at the sky (Z axis), so Y axis look to the north..
     object.lookAt(coord.geodesicNormal.clone().add(object.position));
 }
@@ -94,7 +94,7 @@ function setupPictureFromCamera(camera, imageUrl, opacity, distance) {
 // BUT keep the geodesic normal as Up vector
 // eslint-disable-next-line no-unused-vars
 function setupViewCameraLookingAtObject(camera, coord, objectToLookAt) {
-    camera.position.copy(coord.xyz());
+    camera.position.copy(coord);
     camera.up.copy(coord.geodesicNormal);
     camera.lookAt(objectToLookAt.getWorldPosition());
 }
--- a/examples/multiglobe.html
+++ b/examples/multiglobe.html
@@ -120,11 +120,11 @@
                 }
 
                 if (delta > 0) {
-                    geo.setAltitude(geo.altitude() * 0.9);
+                    geo.altitude *= 0.9;
                 } else {
-                    geo.setAltitude(geo.altitude() * 1.1);
+                    geo.altitude *= 1.1;
                 }
-                view.camera.camera3D.position.copy(geo.as('EPSG:4978').xyz());
+                view.camera.camera3D.position.copy(geo.as('EPSG:4978'));
                 view.notifyChange(view.camera.camera3D);
             };
             viewerDiv.addEventListener('DOMMouseScroll', onMouseWheel);
--- a/examples/orientation_utils.html
+++ b/examples/orientation_utils.html
@@ -86,7 +86,7 @@
 
             // Load camera calibrations
             var calibrations = 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/immersive/exampleParis1/cameraCalibration.json';
-            
+
             // Use the cameraCalibrationParser to create OrientedImageCamera from camera calibration description file.
             promises.push(
                 itowns.Fetcher.json(calibrations, {})
@@ -116,7 +116,7 @@
                     var coord = new itowns.Coordinates(feature.crs, vertices[0], vertices[1], vertices[2]);
 
                     // set model position
-                    model.position.copy(coord.xyz());
+                    model.position.copy(coord);
 
                     // set model orientation (using target parameter)
                     itowns.OrientationUtils.quaternionFromAttitude(feature.geometry[0].properties, coord, true, model.quaternion);
@@ -127,8 +127,8 @@
                     // compute position on the ground
                     coord = coord.as('EPSG:4326');
                     result = itowns.DEMUtils.getElevationValueAt(view.tileLayer, coord, 1, undefined);
-                    coord.setAltitude(result.z);
-                    model.positionOnGround = coord.as(view.referenceCrs).xyz();
+                    coord.altitude = result.z;
+                    model.positionOnGround = coord.as(view.referenceCrs).toVector3();
 
                     // update matrix world and add the model
                     model.updateMatrixWorld();
--- a/examples/orientation_utils_planar.html
+++ b/examples/orientation_utils_planar.html
@@ -100,9 +100,8 @@
                     var coord = new itowns.Coordinates(feature.crs, vertices[0], vertices[1], vertices[2]);
 
                     // set model position to the ground
-                    var position = coord.xyz();
-                    position.z = 0;
-                    model.position.copy(position);
+                    model.position.copy(coord);
+                    model.position.z = 0;
 
                     // set model orientation (quaternion is the target parameter)
                     itowns.OrientationUtils.quaternionFromAttitude(feature.geometry[0].properties, undefined, false, model.quaternion);
--- a/examples/positionGlobe.html
+++ b/examples/positionGlobe.html
@@ -64,10 +64,10 @@
 
                 // position of the mesh
                 var meshCoord = cameraTargetPosition;
-                meshCoord.setAltitude(cameraTargetPosition.altitude() + 30);
+                meshCoord.altitude += 30;
 
                 // position and orientation of the mesh
-                mesh.position.copy(meshCoord.as(view.referenceCrs).xyz());
+                mesh.position.copy(meshCoord.as(view.referenceCrs));
                 mesh.lookAt(new THREE.Vector3(0, 0, 0));
                 mesh.rotateX(Math.PI / 2);
 
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -37,7 +37,7 @@ const xyz = new Coordinates('EPSG:4978', 0, 0, 0);
 const c = new Coordinates('EPSG:4326', 0, 0, 0);
 // Position object on globe
 function positionObject(newPosition, object) {
-    xyz.set('EPSG:4978', newPosition).as('EPSG:4326', c);
+    xyz.setFromVector3(newPosition).as('EPSG:4326', c);
     object.position.copy(newPosition);
     object.lookAt(c.geodesicNormal.add(newPosition));
     object.rotateX(Math.PI * 0.5);
@@ -934,7 +934,7 @@ function GlobeControls(view, targetCoordinate, range, globeRadius, options = {})
         this.camera.layers.enable(layerTHREEjs);
     }
 
-    positionObject(targetCoordinate.as('EPSG:4978').xyz(), cameraTarget);
+    positionObject(targetCoordinate.as('EPSG:4978', xyz), cameraTarget);
 
     this.lookAtCoordinate({
         coord: targetCoordinate,
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -68,7 +68,7 @@ function coordinatesToVertices(ptsIn, normals, target, altitude = 0, extrude = 0
     } else if (Array.isArray(altitude)) {
         fnAltitude = id => altitude[(id - startIn) / 3];
     } else {
-        fnAltitude = id => altitude({}, coord.set(coord.crs, ptsIn[id], ptsIn[id + 1], ptsIn[id + 2]));
+        fnAltitude = id => altitude({}, coord.setFromArray(ptsIn, id));
     }
 
     for (let i = startIn, j = offsetOut; i < endIn; i += 3, j += 3) {
@@ -426,7 +426,8 @@ function featuresToThree(features, options) {
     if (!features || features.length == 0) { return; }
 
     if (features.length == 1) {
-        coord.set(features[0].crs, 0, 0, 0);
+        coord.crs = features[0].crs;
+        coord.setFromValues(0, 0, 0);
         return featureToMesh(features[0], options);
     }
 
@@ -434,7 +435,8 @@ function featuresToThree(features, options) {
     group.minAltitude = Infinity;
 
     for (const feature of features) {
-        coord.set(feature.crs, 0, 0, 0);
+        coord.crs = feature.crs;
+        coord.setFromValues(0, 0, 0);
         const mesh = featureToMesh(feature, options);
         group.add(mesh);
         group.minAltitude = Math.min(mesh.minAltitude, group.minAltitude);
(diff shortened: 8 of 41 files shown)


src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,270 @@
...
+    /**
+     * Returns a new Coordinates with the same values as this one. A target can
+     * be passed to avoid the creation of a new one. Otherwise it will
+     * instantiate a new Coordinates with the same CRS as this one.
+     *
+     * @param {Coordinates} [target]
+     *
+     * @return {Coordinates} The target with its new coordinates.
+     */
+    clone(target = new Coordinates(this.crs)) {
```
gchoqueux: Target isn't necessary because copy do the same thing.

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,270 @@
...
- */
+    /**
+     * Return this Coordinates values into a `THREE.Vector3`.
+     *
+     * @param {THREE.Vector3} [target] - The target to put the values in. If not
+     * specified, a new vector will be created.
+     *
+     * @return {THREE.Vector3}
+     */
+    xyz(target = new THREE.Vector3()) {
```
gchoqueux: May be rename by `toVector3` and to do mandatory the target, because `Coordinates` have `x`, `y` and `z` yet.
It should be seen in the code, when `Coordinates.xyz()` is used, if `Vector3` is needed.

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,270 @@
...
+     * Coordinates e.g: `extent.center().offsetInExtent(extent)` would return
+     * `(0.5, 0.5)`.
+     *
+     * @param {Extent} extent
+     * @param {Vector2} [target] - Optional Vector2 target. If not specified a
+     * new one will be created.
+     *
+     * @return {Vector2} Normalized offset in extent.
+     */
+    offsetInExtent(extent, target = new THREE.Vector2()) {
```
gchoqueux: May be move in `Extent Class`? what do you think?

src/Core/Geographic/Extent.js
```diff
@@ -244,15 +251,15 @@ class Extent {
         const c = (this.crs == coord.crs) ? coord : coord.as(this.crs, _c);
         // TODO this ignores altitude
         if (CRS.isGeographic(this.crs)) {
-            return c.longitude() <= this.east + epsilon &&
-                   c.longitude() >= this.west - epsilon &&
-                   c.latitude() <= this.north + epsilon &&
-                   c.latitude() >= this.south - epsilon;
+            return c.longitude <= this.east + epsilon &&
```
gchoqueux: you could return directly  the second part (the `else` part)  with x and z.

src/Utils/CameraUtils.js
```diff
@@ -252,7 +252,8 @@ class CameraRig extends THREE.Object3D {
             if (params.callback) {
                 params.callback(this);
             }
-            targetCoord.set(view.referenceCrs, this.targetWorldPosition).as(tileLayer(view).extent.crs, this.coord);
+            targetCoord.crs = view.referenceCrs;
+            targetCoord.set(this.targetWorldPosition).as(tileLayer(view).extent.crs, this.coord);
```
gchoqueux: I think that now, may be rename `set` to `setValues`? 
it may be necessary to remove the ambiguity, on the crs which is not modified.

gchoqueux (commented):
Good move!

src/Core/Geographic/Extent.js
zarov: Yes, didn't see this at all...

src/Core/Geographic/Coordinates.js
zarov: Yeah I agree, but I think it doesn't hurt to leave both of them, no ?

src/Core/Geographic/Coordinates.js
zarov: I moved it directly to `DEMUtils`, where it is the only place it used.

src/Utils/CameraUtils.js
zarov: I would keep it like this, because in `Extent#set` we only set the values too, not the CRS.

src/Core/Geographic/Coordinates.js
zarov: Done

zarov:
Updated

src/Core/Geographic/Coordinates.js
gchoqueux: the function must itself instantiate `Coordinates` and not have target.

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,250 @@
...
+    /**
+     * Return this Coordinates values into a `THREE.Vector3`.
+     *
+     * @param {THREE.Vector3} [target] - The target to put the values in. If not
+     * specified, a new vector will be created.
+     *
+     * @return {THREE.Vector3}
+     */
+    toVector3(target = new THREE.Vector3()) {
+        return target.set(this.x, this.y, this.z);
```
gchoqueux: you could write `return target.copy(this);`?

src/Core/Geographic/Coordinates.js
gchoqueux: ok

src/Utils/CameraUtils.js
gchoqueux: ok

src/Core/Geographic/Coordinates.js
gchoqueux: you could write  
```
return target.set(this.x, this.y, this.z);
```
there are other functions where is possible to simplify as well.

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,250 @@
...
+        } else if (CRS.is4326(this.crs) && crs == 'EPSG:3857') {
+            this.y = THREE.Math.clamp(this.y, -89.999999, 89.999999);
+            target.set(proj4cache(this.crs, crs).forward([this.x, this.y]), this.z);
+        } else {
+            target.set(proj4cache(this.crs, crs).forward([this.x, this.y]), this.z);
+        }
 
-    const x = CRS.isGeographic(this.crs) ? this.longitude() : this.x();
-    const y = CRS.isGeographic(this.crs) ? this.latitude() : this.y();
+        target.crs = crs;
```
gchoqueux, line 272: it will simplify the code if you write `setCrs()` function.

```js
// this will replace
coord.crs = crs;
coord.set(...);

// by 
coord.set(...).setCrs(crs);
```

src/Layer/OrientedImageLayer.js
```diff
@@ -130,8 +130,9 @@ class OrientedImageLayer extends GeometryLayer {
             let i = 0;
             for (const pano of this.panos) {
                 // set position
-                coord.set(pano.crs, pano.vertices[0], pano.vertices[1], pano.vertices[2]);
-                pano.position = coord.xyz();
+                coord.crs = pano.crs;
+                coord.set(pano.vertices[0], pano.vertices[1], pano.vertices[2]);
```
gchoqueux: `set` doesn't work with Array? 

it may be well to code the function `setFromArray(array, offset = 0)`, WDYT?

src/Renderer/Camera.js
```diff
@@ -119,7 +119,7 @@ Camera.prototype.position = function position(crs) {
  * @param {Coordinates} position the new position of the camera
  */
 Camera.prototype.setPosition = function setPosition(position) {
-    this.camera3D.position.copy(position.as(this.crs).xyz());
+    this.camera3D.position.copy(position.as(this.crs).toVector3());
```
gchoqueux: `toVector` isn't necessary.

src/Renderer/Camera.js
```diff
@@ -224,11 +224,11 @@ Camera.prototype.adjustAltitudeToAvoidCollisionWithLayer = function adjustAltitu
...
         const elevationUnderCamera = DEMUtils.getElevationValueAt(elevationLayer, camLocation);
         if (elevationUnderCamera != undefined) {
-            const difElevation = camLocation.altitude() - (elevationUnderCamera.z + minDistanceCollision);
+            const difElevation = camLocation.altitude - (elevationUnderCamera.z + minDistanceCollision);
             // We move the camera to avoid collisions if too close to terrain
             if (difElevation < 0) {
-                camLocation.setAltitude(elevationUnderCamera.z + minDistanceCollision);
-                view.camera.camera3D.position.copy(camLocation.as(view.referenceCrs).xyz());
+                camLocation.altitude = elevationUnderCamera.z + minDistanceCollision;
+                view.camera.camera3D.position.copy(camLocation.as(view.referenceCrs).toVector3());
```
gchoqueux: `toVector` isn't necessary.

examples/collada.html
```diff
@@ -67,7 +67,7 @@
                 var coord = new itowns.Coordinates('EPSG:4326', 4.2165, 44.844, 1417);
                 var colladaID = view.mainLoop.gfxEngine.getUniqueThreejsLayer();
 
-                model.position.copy(coord.as(view.referenceCrs).xyz());
+                model.position.copy(coord.as(view.referenceCrs).toVector3());
```
gchoqueux: `toVector` isn't necessary.

examples/js/OrientedImageHelper.js
```diff
@@ -94,7 +94,7 @@ function setupPictureFromCamera(camera, imageUrl, opacity, distance) {
 // BUT keep the geodesic normal as Up vector
 // eslint-disable-next-line no-unused-vars
 function setupViewCameraLookingAtObject(camera, coord, objectToLookAt) {
-    camera.position.copy(coord.xyz());
+    camera.position.copy(coord.toVector3());
```
gchoqueux: `toVector` isn't necessary.

examples/multiglobe.html
```diff
@@ -120,11 +120,11 @@
...
 
                 if (delta > 0) {
-                    geo.setAltitude(geo.altitude() * 0.9);
+                    geo.altitude *= 0.9;
                 } else {
-                    geo.setAltitude(geo.altitude() * 1.1);
+                    geo.altitude *= 1.1;
                 }
-                view.camera.camera3D.position.copy(geo.as('EPSG:4978').xyz());
+                view.camera.camera3D.position.copy(geo.as('EPSG:4978').toVector3());
```
gchoqueux: `toVector` isn't necessary.

examples/positionGlobe.html
```diff
@@ -64,10 +64,10 @@
 
                 // position of the mesh
                 var meshCoord = cameraTargetPosition;
-                meshCoord.setAltitude(cameraTargetPosition.altitude() + 30);
+                meshCoord.altitude += 30;
 
                 // position and orientation of the mesh
-                mesh.position.copy(meshCoord.as(view.referenceCrs).xyz());
+                mesh.position.copy(meshCoord.as(view.referenceCrs).toVector3());
```
gchoqueux: `toVector` isn't necessary.

gchoqueux:
there isn't BREACKING CHANGE on renaming `xzy()` to `toVector3()`

src/Core/Geographic/Coordinates.js
zarov: Yes, well seen !

src/Core/Geographic/Coordinates.js
zarov: Ok, but do you see `CRS.isValid` called in `setCrs` ?

src/Layer/OrientedImageLayer.js
zarov: It does work with an array, I just didn't change it correctly here. But I'm all for a `setFromArray` as well as `setFromVector3`, it would avoid having conditions in `set`. WDYT ?

zarov:
Removed all you said, thanks for noticing this !

> there isn't BREACKING CHANGE on renaming xzy() to toVector3()

It isn't really one but it's still a change in the function signature, no ?

zarov:
Rebased and updated

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,276 @@
...
+    /**
+     * Copies the values of the passed Coordinates to this one. The CRS is
+     * however not copied.
+     *
+     * @param {Coordinates} src - The source to copy from.
+     *
+     * @return {Coordinates} This Coordinates.
+     */
+    copy(src) {
+        this.setFromVector3(src);
```
gchoqueux: you could write
```js
copy(src) {
   return this.setFromVector3(src);
}
```

src/Core/Geographic/Coordinates.js
gchoqueux: not necessarily, to validate crs in `setCrs()`

src/Layer/OrientedImageLayer.js
gchoqueux: ok to write `setFromArray`

zarov:
@gchoqueux updated

examples/orientation_utils.html
```diff
@@ -116,7 +116,7 @@
                     var coord = new itowns.Coordinates(feature.crs, vertices[0], vertices[1], vertices[2]);
 
                     // set model position
-                    model.position.copy(coord.xyz());
+                    model.position.copy(coord.toVector3());
```
gchoqueux: not necessary `object.position.copy(coord)`

examples/orientation_utils_planar.html
```diff
@@ -100,7 +100,7 @@
                     var coord = new itowns.Coordinates(feature.crs, vertices[0], vertices[1], vertices[2]);
 
                     // set model position to the ground
-                    var position = coord.xyz();
+                    var position = coord.toVector3();
```
gchoqueux: simplify by
```js
model.position.copy(coord);
model.position.z = 0;
```

src/Controls/GlobeControls.js
```diff
@@ -934,7 +934,7 @@ function GlobeControls(view, targetCoordinate, range, globeRadius, options = {})
         this.camera.layers.enable(layerTHREEjs);
     }
 
-    positionObject(targetCoordinate.as('EPSG:4978').xyz(), cameraTarget);
+    positionObject(targetCoordinate.as('EPSG:4978'), cameraTarget);
```
gchoqueux: try
```suggestion
    positionObject(targetCoordinate.as('EPSG:4978', c), cameraTarget);
```

src/Converter/Feature2Mesh.js
```diff
@@ -68,7 +68,7 @@ function coordinatesToVertices(ptsIn, normals, target, altitude = 0, extrude = 0
     } else if (Array.isArray(altitude)) {
         fnAltitude = id => altitude[(id - startIn) / 3];
     } else {
-        fnAltitude = id => altitude({}, coord.set(coord.crs, ptsIn[id], ptsIn[id + 1], ptsIn[id + 2]));
+        fnAltitude = id => altitude({}, coord.set(ptsIn[id], ptsIn[id + 1], ptsIn[id + 2]));
```
gchoqueux: there isn't `coord.setFromArray(array, offset)`?

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,280 @@
...
+     * Set the values of this Coordinates from an array.
+     *
+     * @param {Array<number>} array - An array of number to assign to the
+     * Coordinates.
+     * @param {number} [offset] - Optional offset into the array.
+     *
+     * @return {Coordinates} This Coordinates.
+     */
+    setFromArray(array, offset = 0) {
+        this.x = array[offset];
```
gchoqueux, line 125: could you use `setFromValues`?

src/Core/Geographic/Coordinates.js
```diff
@@ -1,441 +1,280 @@
...
+     * Set the values of this Coordinates from a `THREE.Vector3` or an `Object`
+     * having `x/y/z` properties, like a `Coordinates`.
+     *
+     * @param {THREE.Vector3|Coordinates} v0 - The object to read the values
+     * from.
+     *
+     * @return {Coordinates} This Coordinates.
+     */
+    setFromVector3(v0) {
+        this.x = v0.x;
```
gchoqueux, line 145: could you use `setFromValues`?

src/Converter/Feature2Texture.js
```diff
@@ -106,7 +106,7 @@ function drawFeature(ctx, feature, extent, style, invCtxScale) {
                     const offset = indice.offset * feature.size;
                     const count = offset + indice.count * feature.size;
                     for (let j = offset; j < count; j += feature.size) {
-                        coord.set(extent.crs, feature.vertices[j], feature.vertices[j + 1]);
+                        coord.setFromValues(feature.vertices[j], feature.vertices[j + 1]);
```
gchoqueux: could you use `setFromArray`?

src/Controls/GlobeControls.js
zarov: Changed `c` to `xyz` to stay on the same CRS.

src/Core/Geographic/Coordinates.js
zarov: No, because there is no point to have it separate across multiple method if you always need to call one specific method. At least that's what I think in term of performance, but I could be wrong.

zarov:
Updated !

src/Core/Geographic/Coordinates.js
gchoqueux: to check

gchoqueux (approved):
Very good PR that goes in the right direction.

src/Core/Geographic/Coordinates.js
zarov: It is approximately the same in fact https://jsbench.me/0ajvz7nexn/1

gchoqueux: ok, keep the actual code.

PR #1131, opened 2019-05-09
feat(parser): add a TIFF parser using utif

zarov:
It automatically rejects TIFFs that are bigger than 4096x4096, because
[it is the minimum that is supported on 99.99% of the devices][0].

[0]: https://webglstats.com/webgl/parameter/MAX_TEXTURE_SIZE


src/Parser/TIFFParser.js
```diff
@@ -0,0 +1,67 @@
...
+        UTIF.decodeImage(data, IFD);
+
+        // A 4096 limitation sounds the most probable here
+        // https://webglstats.com/webgl/parameter/MAX_TEXTURE_SIZE
+        if (IFD.width > 4096 || IFD.height > 4069) {
+            throw new Error('The loaded TIFF is too big: please reduce it to a maximum size of 4096 by 4096');
+        }
+
+        // Round to next power of two
+        const width = 2 ** (~~Math.log2(IFD.width - 1) + 1);
```
gchoqueux: there is utils in THREE to compute `next power of two`.

src/Parser/TIFFParser.js
```diff
@@ -0,0 +1,67 @@
...
+        const it = IFD.data.values();
+        const rowSize = IFD.height * alpha;
+        let rowOffset = 0;
+        for (let i = IFD.width; i > 0; i--) {
+            rowOffset = i * width * alpha;
+            resizedData.fill(255, rowOffset, rowOffset + rowSize);
+            for (let j = 0; j < rowSize; j += alpha) {
+                resizedData[rowOffset + j] = it.next().value;
+                resizedData[rowOffset + j + 1] = it.next().value;
+                resizedData[rowOffset + j + 2] = it.next().value;
```
gchoqueux: the alpha isn't take in account?

src/Parser/TIFFParser.js
gchoqueux: there is utils in THREE to compute `next power of two`.

src/Parser/TIFFParser.js
```diff
@@ -0,0 +1,67 @@
...
+        const alpha = (IFD.width == width && IFD.height == height) ? 3 : 4;
+
+        const resizedData = new IFD.data.constructor(width * height * alpha);
+
+        const it = IFD.data.values();
+        const rowSize = IFD.height * alpha;
+        let rowOffset = 0;
+        for (let i = IFD.width; i > 0; i--) {
+            rowOffset = i * width * alpha;
+            resizedData.fill(255, rowOffset, rowOffset + rowSize);
```
gchoqueux: why you fill the `resizedData`.

src/Parser/TIFFParser.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        // Round to next power of two
+        const width = 2 ** (~~Math.log2(IFD.width - 1) + 1);
+        const height = 2 ** (~~Math.log2(IFD.height - 1) + 1);
+
+        // Avoid using alpha layer if the TIFF is already at the right size
+        const alpha = (IFD.width == width && IFD.height == height) ? 3 : 4;
+
+        const resizedData = new IFD.data.constructor(width * height * alpha);
```
gchoqueux: `THREE` must be able to handle image resizing.

examples/tiff.html
```diff
@@ -0,0 +1,74 @@
...
+
+            // Instanciate iTowns GlobeView*
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            setupLoadingScreen(viewerDiv, view);
+
+            function addTIFF() {
+                itowns.Fetcher.arrayBuffer('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/tiff/FLAG_T24.TIF')
+                    .then(itowns.TIFFParser.parse)
+                    .then(function _(texture) {
+                        texture.coords = new itowns.Coordinates('EPSG:4326', 1.4116667, 43.6074422);
```
gchoqueux: The `texture.coords` have bad name, because it's actually a `Extent`. 
this Extent allows you to place the texture on the tiles. In particular with initialization from parent texture. 

There are little changes to enhance the image display. 
* Use `Extent` insteaf of `Coordinates`, and add it a `zoom` property to set image scale.
* Add a layer property : use `imageParent` after zoom max to always display the ColorLayer after the Source.zoom.max.

src/Parser/TIFFParser.js
zarov: It is set to `255` before this loop, I didn't add it here because it needs a `if`.

src/Parser/TIFFParser.js
zarov: Nice, changed thanks !

src/Parser/TIFFParser.js
zarov: For the opacity, see other comment

src/Parser/TIFFParser.js
zarov: Yes, but I don't want to stretch the image from its initial size, and I want to avoid having `THREE` warnings in the console.

examples/tiff.html
zarov: This needs another change in [`MaterialLayer` as well](https://github.com/iTowns/itowns/blob/master/src/Renderer/MaterialLayer.js#L135).

> Add a layer property : use imageParent after zoom max to always display the ColorLayer after the Source.zoom.max.

I don't understand what you mean here

gchoqueux: A ColorLayer isn't displayed when Source.zoom.max is exceeded, it isn't the expected behaviour because the ColorLayer displaying shouldn't be controlled by the source.

zarov:
Updated, with a tiled example (the same as https://www.itowns-project.org/demo-geoid/) but needs #1126 and #1142 to be merged first.

zarov:
Now that all others things have been merged, this PR is now ready !

src/Parser/TIFFParser.js
gchoqueux: Verify if the tiff supports opacity. if not, could you use RGB texture?

examples/tiff.html
```diff
@@ -0,0 +1,53 @@
...
+            var viewerDiv = document.getElementById('viewerDiv');
+
+            // Instanciate iTowns GlobeView*
+            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var menuGlobe = new GuiTools('menuDiv', view);
+            setupLoadingScreen(viewerDiv, view);
+
+            // Add one imagery layer to the scene, read from TIFFs.
+            var tiffSource = new itowns.TMSSource({
+                url: 'http://localhost/iTowns2-sample-data/geoid/localcolors/tiff/${z}/localcolors_${x}_${y}.tif',
```
gchoqueux: could add the tiff on iTowns2-sample-data?

src/Parser/TIFFParser.js
```diff
@@ -0,0 +1,74 @@
...
+     * @return {Promise} A promise resolving with a `THREE.DataTexture`.
+     */
+    parse(data) {
+        const IFD = UTIF.decode(data)[0];
+        UTIF.decodeImage(data, IFD);
+        IFD.data = UTIF.toRGBA8(IFD);
+
+        // A 4096 limitation sounds the most probable here
+        // https://webglstats.com/webgl/parameter/MAX_TEXTURE_SIZE
+        if (IFD.width > 4096 || IFD.height > 4069) {
```
gchoqueux: use gpu capabilities.

src/Parser/TIFFParser.js
```diff
@@ -0,0 +1,74 @@
...
+
+        // Round to next power of two
+        const width = THREE.Math.ceilPowerOfTwo(IFD.width);
+        const height = THREE.Math.ceilPowerOfTwo(IFD.height);
+
+        let resizedData;
+        if (width == IFD.width && height == IFD.height) {
+            resizedData = IFD.data;
+        } else {
+            resizedData = new IFD.data.constructor(width * height * 4);
```
gchoqueux: THREE use canvas to resize image see : https://github.com/mrdoob/three.js/blob/master/src/renderers/webgl/WebGLTextures.js#L63

gchoqueux (commented):
It should be interesting to create a itowns plugin to support Tiff.

src/Parser/TIFFParser.js
zarov: Yes but then we got a warning: I want to avoid that.

gchoqueux: yes, but may be is better, to resizing, to use `canva.context.drawImage( image, 0, 0, width, height )`.

src/Parser/TIFFParser.js
zarov: Done

examples/tiff.html
zarov: Done

zarov:
Updated and rebased, I moved the parser to be a plugin. I also exposed the `Capabilities` class in the same time. In a next PR I'll start cleaning the `examples/` folder, because the plugins are hard to see.

src/Parser/TIFFParser.js
zarov: I tried it: it is too complicated IMO, because `drawImage` can't take a `TypedArray` as an argument, so I first need to convert the `TypedArray` to an image, then draw it again to resized it. I prefer to read it manually and put it into a `THREE.DataTexture`.

gchoqueux (approved):

PR #1132, opened 2019-05-09
chore(view): remove support of passing a source as an object

zarov:
BREAKING CHANGE: Adding a layer to a view, with a source not
instantiated (only as an object) is no longer supported, even in
deprecation mode.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -9,20 +9,6 @@ import { getMaxColorSamplerUnitsCount } from 'Renderer/LayeredMaterial';
 
 import Scheduler from 'Core/Scheduler/Scheduler';
 import Picking from 'Core/Picking';
-import WMTSSource from 'Source/WMTSSource';
-import WMSSource from 'Source/WMSSource';
-import WFSSource from 'Source/WFSSource';
-import TMSSource from 'Source/TMSSource';
-import FileSource from 'Source/FileSource';
-
-const supportedSource = new Map([
-    ['wmts', WMTSSource],
-    ['file', FileSource],
-    ['wfs', WFSSource],
-    ['wms', WMSSource],
-    ['tms', TMSSource],
-    ['xyz', TMSSource],
-]);
 
 export const VIEW_EVENTS = {
     /**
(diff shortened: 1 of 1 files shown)


src/Core/View.js
```diff
@@ -165,12 +151,6 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
                 providerPreprocessing = Promise.resolve();
             }
         } else if (layer.source) {
-            if (!layer.source.isSource) {
```
gchoqueux, line 168: `else if (layer.source)` is necessary? may be replace by simple `else`.

src/Core/View.js
```diff
@@ -165,12 +151,6 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
                 providerPreprocessing = Promise.resolve();
```
gchoqueux: could you rename `providerPreprocessing` in `whenReady`.

gchoqueux: `provider.preprocessDataLayer` should be return Promise to simplify this part.

src/Core/View.js
zarov: I let it as it is, but removed the `|| providerPreprocessing`

src/Core/View.js
zarov: Done

zarov: Done, I simplified the whole section of `whenReady`.

src/Core/View.js
```diff
@@ -158,29 +144,19 @@ function _preprocessLayer(view, layer, provider, parentLayer) {
...
-        } else if (layer.source) {
-            if (!layer.source.isSource) {
-                console.warn('Deprecation warning: passing a source as an object is deprecated. Instantiate the source before adding it to the layer instead.');
-                const protocol = layer.source.protocol;
-                layer.source = new (supportedSource.get(protocol))(layer.source, layer.projection);
-            }
-
-            providerPreprocessing = layer.source.whenReady || providerPreprocessing;
+            layer.whenReady = provider.preprocessDataLayer(layer, view, view.mainLoop.scheduler, parentLayer);
+        } else if (layer.source && layer.source.whenReady) {
```
gchoqueux, line 149: All sources should be have `whenReady`.
we could have:
```suggestion
        } else { 
            layer.whenReady = layer.source.whenReady;
        }
```

zarov: Ok so we would have in `Source.js` constructor something like `this.whenReady = Promise.resolve();` ?

gchoqueux: yes

zarov: In fact, we can't do that, because a layer doesn't always have a source.

gchoqueux: ok, you have a example?

zarov: Yes, see the `3dtiles.html` example

gchoqueux: ok

gchoqueux (approved):

PR #1138, opened 2019-05-13
chore: update THREE to r104.0

gchoqueux:
## Description
update THREE to r104.0


zarov (approved):
No breaking changes - that's the first release of THREE that I see like this !

PR #1139, opened 2019-05-13
feature(3dtile): support rtc center from feature.

gchoqueux:
## Description
Add support of rtc center from feature table.

### DIFF
--- a/src/Parser/B3dmParser.js
+++ b/src/Parser/B3dmParser.js
@@ -37,17 +37,6 @@ function filterUnsupportedSemantics(obj) {
         }
     }
 }
-// parse for RTC values
-function applyOptionalCesiumRTC(data, gltf) {
-    const headerView = new DataView(data, 0, 20);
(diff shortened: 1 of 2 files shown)


zarov (approved):
LGTM. 

@jailln I don't remember, but do you plan to add support to feature table ? I'm asking since you (and I'm very grateful for that) did the recent 3dtiles cleaning.

jailln:
@zarov I don't plan to add it for now; however this PR might be a good opportunity to introduce it as the feature table for b3dm is only composed of two attributes: RTC_CENTER and BATCH_LENGTH (see [this json schema](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/schema/b3dm.featureTable.schema.json)) ?

zarov:
Oh I say why not then. @gchoqueux can you do that in this PR if it doesn't take too much time ?

gchoqueux:
`BATCH_LENGTH` support added.

zarov (approved):
LGTM, thanks for this !

PR #1142, opened 2019-05-17
fix(tms): return the correct coordinates for a TMS source

zarov:
When using a GlobeView and a TMSSource, the coordinates returned for a
tile were only Pseudo-Mercator; it can be WGS84G now.

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -1,6 +1,6 @@
 import * as THREE from 'three';
 import OGCWebServiceHelper from 'Provider/OGCWebServiceHelper';
-import CRS from 'Core/Geographic/Crs';
+import Projection from 'Core/Geographic/Projection';
 
 /**
  * A TileMesh is a THREE.Mesh with a geometricError and an OBB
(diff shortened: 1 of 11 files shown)


src/Core/TileMesh.js
```diff
@@ -76,11 +76,11 @@ class TileMesh extends THREE.Mesh {
                 throw new Error('unsupported projection wms for this viewer');
             }
         } else if (source.isTMSSource) {
-            // Special globe case: use the P(seudo)M(ercator) coordinates
+            // Special globe case
             if (CRS.is4326(this.extent.crs) &&
                     (source.extent.crs == 'EPSG:3857' || CRS.is4326(source.extent.crs))) {
-                OGCWebServiceHelper.computeTileMatrixSetCoordinates(this, 'PM');
-                return this.wmtsCoords.PM;
+                OGCWebServiceHelper.computeTileMatrixSetCoordinates(this, source.tileMatrixSet);
```
gchoqueux: if source is TMS, it must have `tileMatrixSet`, it should not have any difference with `wmtsSource`.
we should have
```js
else if (source.isTMSSource || source.isWTMSSource) {
  OGCWebServiceHelper.computeTileMatrixSetCoordinates(this, source.tileMatrixSet);
  return this.wmtsCoords[source.tileMatrixSet];
}
```

zarov: This is not possible, as this breaks planar cases.

gchoqueux: `computeTileMatrixSetCoordinates` should return `computeTMSCoordinates`  when `tilematrixset` isn't `PM` or `WGS84`

zarov: I tried it all, I unfortunately can't do any better :/ There's too much different cases to be supported.

gchoqueux: You could try a refactoring of `getCoordsForSource`.
It should be replaced by `getExtent(crs)`. The `crs` parameter will be a tiled crs or not.
`getExtent(crs)` could be cached the computed extent in `Set` object.

zarov:
PR updated, I simplified a lot of things, like computing the extents at the creation of the `TileMesh`. Only exception where I can't find a solution for now is when using a `TMSSource`. I wrote a `TODO` if we can find an alternative in the future, but for now I suggest to merge and close this PR, if it's good with you of course.

gchoqueux (approved):

PR #1143, opened 2019-05-17
fix(tooltip): display all properties of a feature

zarov:
Before, only a few selected properties of a feature were displayed in
FeatureToolTip. Now, everything is displayed.
In the same time, a quick cleaning has been done, and the FeatureToolTip
has been added to the shapefile example.

### DIFF
--- a/examples/css/example.css
+++ b/examples/css/example.css
@@ -99,3 +99,26 @@ body {
     border: 2px solid #775555;
     vertical-align: middle;
 }
+
+/* tooltip */
+.tooltip {
+    display: none;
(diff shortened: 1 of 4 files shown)


gchoqueux (commented):
You should add a filter callback, there are too much displayed informations.
it would be necessary that the example globe vecto should be similar to the previous version.

zarov:
I updated `FeatureToolTip`, added some documentation and simplified its usage.

gchoqueux (commented):
Just a comment, the line symbol on tool type have two colors.
By example, on gpx, the Line symbol is red and white, while it should be only red.

zarov:
Well seen, updated

gchoqueux (approved):

PR #1148, opened 2019-06-04
chore: update THREE to r105.2

gchoqueux:
### update THREE to r105.2

- [x] For usability reasons, the default of WebGLRenderer.debug.checkShaderErrors is now false.
- [x] EffectComposer.setSize() now respects the pixel ratio. An instance of EffectComposer can now be resized with the same width and height values like WebGLRenderer.
- [x] Renamed QuickHull to ConvexHull. The file is now located in examples/js/math.
- [x] SimplexNoise and ImprovedNoise are now in the THREE namespace and located in examples/js/math.
- [x] AnimationClipCreator and TimelinerController are now located in examples/js/animation.
- [x] ParametricGeometries is now located in examples/js/geometries.
- [x] hilbert2d and hilbert3D were removed. Please use GeometryUtils.hilbert2D() and GeometryUtils.hilbert3D() instead.

### Interesting features for iTowns:
- Matrix4 
  - Support `.setPosition( x, y, z )`
- Object3D
  - Added `.attach()`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9040,9 +9040,9 @@
       "dev": true
     },
     "three": {
-      "version": "0.104.0",
-      "resolved": "https://registry.npmjs.org/three/-/three-0.104.0.tgz",
-      "integrity": "sha512-q617IMBC5k40U2E9UC4/LtmhzTOOLB1jGMIooUL+QrhZ7abiGCSDrKrpCDt9V8RTl6xw+0FYfA1PYsIPKbQOgg==",
+      "version": "0.105.2",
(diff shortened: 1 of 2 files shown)


zarov (approved):
LGTM.

Concerning `setPosition`, I quickly checked but didn't see where it could be used. On the other side, `attach` could be used I think, but it seems like a weird method: maybe let's avoid it for now,  WDYT?

gchoqueux:
Thank you for checking the `setPosition`.
For `attach` we will see in the future.

PR #1151, opened 2019-06-05
fix(geographic): return the correct center of an extent

### LINKED ISSUE #1150: MULTIPOLYGON AS A GEOMETRY example is broken
The example:
http://www.itowns-project.org/itowns/examples/multipolygon.html

is broken -> The geometry disappears and never reappears

## Context
When zooming on the geometry it can disappear.

## Solution/ IDEAS
It seems that it can be related to the other regression on building geometry:
http://www.itowns-project.org/itowns/examples/globe_wfs_extruded.html
 The buildings nearly never show anymore eventhough WFS requests are ok

zarov:
The altitude is not taken in account here, as it was before the
refactoring to ES6 Coordinates. `undefined` cases have also been handled
better, and all `set` methods of `Coordinates` redirect to
`setFromValues`.

Fixes #1150

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -104,9 +104,9 @@ class Coordinates {
      * @return {Coordinates} This Coordinates.
      */
     setFromValues(v0 = 0, v1 = 0, v2 = 0) {
-        this.x = v0;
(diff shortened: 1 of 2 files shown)


nosy-b (approved):
Nice

PR #1160, opened 2019-06-20
fixes and optimization for StreetControls

gchoqueux:
### fixes
* error when moveTo is undefined.
* Support planarView and View.

### optimizations
* streetcontrols picking.
* optimize computing rotation state.

### Internal feature
* doesn't clean depth buffer when notify change for helpers.

### DIFF
--- a/src/Controls/StreetControls.js
+++ b/src/Controls/StreetControls.js
@@ -17,7 +17,7 @@ function createCircle() {
 function createRectangle() {
     const geomPlane = new THREE.PlaneGeometry(4, 2, 1);
     const rectangle = new THREE.Mesh(geomPlane, material);
-    rectangle.rotateX(-Math.PI / 2);
+    rectangle.rotateX(-Math.PI * 0.5);
     return rectangle;
 }
 
@@ -31,6 +31,8 @@ function updateSurfaces(surfaces, position, norm) {
 
 // vector use in the pick method
 const target = new THREE.Vector3();
+const up = new THREE.Vector3();
+const startQuaternion = new THREE.Quaternion();
 
 function pick(event, view, buildingsLayer, pickGround = () => {}, pickObject = () => {}, pickNothing = () => {}) {
     // get real distance to ground, with a specific method to pick on the elevation layer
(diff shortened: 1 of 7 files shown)


src/Controls/StreetControls.js
```diff
@@ -178,26 +193,33 @@ class StreetControls extends FirstPersonControls {
             // state mouse drag (move + mouse click)
             this._stateOnMouseDrag = true;
             this.stopAnimations();
+        } else if (!this.tween) {
```
zarov, line 194: I don't understand why you are moving it like this ?

src/Controls/StreetControls.js
```diff
@@ -211,7 +233,13 @@ class StreetControls extends FirstPersonControls {
             return;
         }
         this.camera.position.copy(position);
-        this.camera.up.copy(position).normalize();
+
+        if (this.view.tileLayer && this.view.tileLayer.isGlobeLayer) {
+            this.camera.up.copy(position).normalize();
+        } else {
+            this.camera.up.set(0, 0, 1);
```
zarov, line 238: Maybe add a warning here that something else other than a globe view is not fully supported yet ?

src/Core/MainLoop.js
```diff
@@ -176,6 +176,14 @@ MainLoop.prototype._step = function _step(view, timestamp) {
     this.needsRedraw = false;
     this.renderingState = RENDERING_PAUSED;
     const updateSources = new Set(view._changeSources);
+    let onlyHelpers = true;
+    for (const source of view._changeSources) {
```
zarov: I don't understand the point of this ? Is it really needed ?

src/Core/Prefab/GlobeView.js
```diff
@@ -128,8 +128,9 @@ class GlobeView extends View {
 
         this._fullSizeDepthBuffer = null;
 
-        this.addFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, () => {
-            if (this._fullSizeDepthBuffer != null) {
+        this.addFrameRequester(MAIN_LOOP_EVENTS.BEFORE_RENDER, (dt, updateLoopRestarted, args) => {
```
zarov: Can't this be moved to `View` ?

src/Controls/StreetControls.js
gchoqueux: Because I would like stop picking when there is a animation

src/Controls/StreetControls.js
gchoqueux: there's no need because both are supported

src/Controls/StreetControls.js
zarov: Fair enough

src/Controls/StreetControls.js
zarov: Yes but why are you moving it to `0, 0, 1` ?

src/Core/MainLoop.js
gchoqueux: I remove that, I added a simpler mechanism.

src/Core/Prefab/GlobeView.js
gchoqueux: yes

src/Controls/StreetControls.js
gchoqueux: for planar the up camera is on Z (altitude)

zarov: Oh my bad I was reading `this.camera.set(0, 0, 1); ` :sweat_smile:

zarov (approved):

PR #1162, opened 2019-06-21
fix(core): add a missing case for WMS in TileMesh extent

zarov:
I couldn't get a better fix unfortunately, so we are almost back to the last version.


gchoqueux:
I work on other solution to clarify the handling projection.

gchoqueux:
fixed in #1183

PR #1163, opened 2019-06-24
fix: FT_RTC have to be a local variable

gmaillet:
## Description
Replace a global variable in B3dmParser with a local one

## Motivation and Context
Conflicts with two 3DTiles sources with different RTC


zarov (approved):
Big thanks for finding this !

PR #1174, opened 2019-07-03
feat(layer): add full support to scaling for elevation layers

zarov:
Changing the `scale` property of an `ElevationLayer` can changed the
aspect of the visual, allowing exageration of the elevation of the
globe, as in the tiff.html example.

This is an up-to-date version of #511

### DIFF
--- a/src/Parser/XbilParser.js
+++ b/src/Parser/XbilParser.js
@@ -29,7 +29,7 @@ export function computeMinMaxElevation(buffer, width, height, pitch) {
         const pit = y * (width || 0);
         for (let x = xs; x < xs + sizeX; x += inc) {
             const val = buffer[pit + x];
-            if (val > -10.0) {
+            if (val > -10) {
                 max = Math.max(max, val);
                 min = Math.min(min, val);
             }
--- a/src/Renderer/LayeredMaterial.js
+++ b/src/Renderer/LayeredMaterial.js
@@ -244,6 +244,12 @@ class LayeredMaterial extends THREE.RawShaderMaterial {
     getElevationLayer() {
         return this.layers.find(l => l.id === this.elevationLayerIds[0]);
     }
+
+    setElevationScale(scale) {
+        if (this.elevationLayerIds.length) {
+            this.getElevationLayer().scale = scale;
+        }
+    }
 }
 
 export default LayeredMaterial;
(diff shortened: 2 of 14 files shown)


src/Parser/XbilParser.js
```diff
@@ -29,10 +29,8 @@ export function computeMinMaxElevation(buffer, width, height, pitch) {
         const pit = y * (width || 0);
         for (let x = xs; x < xs + sizeX; x += inc) {
             const val = buffer[pit + x];
-            if (val > -10.0) {
-                max = Math.max(max, val);
-                min = Math.min(min, val);
-            }
+            max = Math.max(max, val);
```
gchoqueux: I think this clamp was to handle the elevation error (with huge negative elevation)

gchoqueux (commented):
Thanks, we will finally be able to integrate this functionality!
I have just on comment on the code and a bug on OBB visibility :

There are culling errors with obb tile, see the bbox culling in camera and horizon culling.

bbox

src/Parser/XbilParser.js
zarov: I understand, but then the example won't work, as there is huge negative elevation, for example Marina trench at -11.000m, times the scale is way beyond the -10 authorized here.

gchoqueux: you could add layer option to clamp the elevation?

zarov:
Up to date and ready, I couldn't get to perfect bbox culling, so I'm living like it is now, as it is only for special case that this is visible (with great exageration of altitude).

src/Renderer/LayeredMaterial.js
```diff
@@ -253,6 +253,12 @@ class LayeredMaterial extends THREE.RawShaderMaterial {
     getElevationLayer() {
         return this.layers.find(l => l.id === this.elevationLayerIds[0]);
     }
+
+    setElevationScale(scale) {
+        if (this.elevationLayerIds.length > 0) {
```
gchoqueux: could you replace by  `if (this.elevationLayerIds.length)`

zarov: done

gchoqueux:
I push a fix on @gchoqueux / fix_geoid. Could you comment and test?

zarov:
It is better, thanks, but it is not perfect yet: I think it will be very difficult to have a perfect result here anyway.

zarov:
So I added your suggestion, and rebase it, now it should be good to go !

gchoqueux:
could you open folder gui, to see directly the scale slider?

zarov:
Done

gchoqueux (approved):
since we've been waiting for this feature: Thanks!

PR #1176, opened 2019-07-08
fix(core): handle the case where the Mesh.material is an array

Crejak:
When disposing the material of a Mesh, the type of the attribute
was not checked and was assumed to be an instance of `Material`.
The case where `material` is an array has now been handled.

## Description

As specified in the [THREE.js documentation](https://threejs.org/docs/index.html#api/en/objects/Mesh) for the Mesh object, the `material` attribute can either be "a single or an array of Material". In iTowns, only the first case was handled. This is easily fixed by adding an condition on the object type (by calling `Array.isArray`) to check wether the `material` attribute is an array.

## Motivation and Context

We found out the problem when trying to color a 3DTiles object using the `groups` attribute of the mesh. When using groups, THREE.js require that the mesh uses an array of materials instead of a single one. This causes a bug where the application crashes at a `material.dispose()` call in `3DTilesProcessing` when a tile unloads.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -25,6 +25,7 @@ The following people have contributed to iTowns 2.
   * [Emmanuel Schmück](https://github.com/EmmanuelSchmuck/)
   * [Marie Lamure](https://github.com/mlamure)
   * [Vincent Jaillot](https://github.com/jailln)
+  * [Valentin Rigolle](https://github.com/Crejak)
 
 * [virtualcitySYSTEMS](https://www.virtualcitysystems.de/)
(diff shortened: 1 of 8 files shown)


zarov (approved):
Thanks for the fix ! Could you also do this modification in `OBBHelper` ?

I never noticed that the material could be an array, I suppose it can break things elsewhere as well, this needs to be investigated then.

Crejak:
> Thanks for the fix ! Could you also do this modification in `OBBHelper` ?

Thank you for the review, I've added the modification in `OBBHelper` and done another commit.

zarov:
Thanks, merged !

PR #1177, opened 2019-07-10
feat(parser): add a plugin to parse OGR VRT and CSV files

zarov:
A very limited plugin, as OGR VRT accepts a wide range of format. Here, only CSV is accepted, and with limited feature (display of points, lines and polygon with no style or other customisation).

### DIFF
--- a/examples/js/CSVnVRTParser.js
+++ b/examples/js/CSVnVRTParser.js
@@ -1,0 +1,201 @@
+/* global itowns, Promise */
+
+/**
+ * A module to parse OGR Virtual Format files.
+ *
+ * See the [GDAL documentation](https://gdal.org/drivers/vector/vrt.html) and
+ * the [xsd
+ * schema](https://github.com/OSGeo/gdal/blob/master/gdal/data/ogrvrt.xsd) of
+ * the OGR VRT file.
+ *
+ * It is highly recommended to use the [Papa Parse](https://www.papaparse.com/)
+ * parser to parse the CSV content before feeding it to this parser.
+ *
+ * @example
+ * Fetcher.multi('data', {
+ *     xml: ['vrt'],
+ *     text: ['csv']
+ * }).then(function _(res) {
+ *     res.csv = Papa.parse(res.csv.trim()).data;
+ *     return CSVnVRTParser.parse(res, {
+ *         buildExtent: true,
+ *         crsOut: 'EPSG:4326'
+ *     });
+ * }).then(function _(feature) {
+ *     var source = new itowns.FileSource({ parsedData: feature });
+ *     var layer = new itowns.ColorLayer('CSVnVRT', { source });
+ *     view.addLayer(layer);
+ * });
+ *
+ * @module CSVnVRTParser
+ */
+var CSVnVRTParser = (function _() {
+    var coord = new itowns.Coordinates('EPSG:4326');
+    var header;
+
+    function xml2json(xml, json) {
+        var res = {};
+
+        var attributes = xml.getAttributeNames();
+        if (attributes.length > 0) {
+            res['@attributes'] = {};
+            for (var i = 0; i < attributes.length; i++) {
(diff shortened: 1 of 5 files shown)


examples/js/CSVnVRTParser.js
```diff
@@ -0,0 +1,207 @@
...
+ *     var layer = new itowns.ColorLayer('CSVnVRT', { source });
+ *     view.addLayer(layer);
+ * });
+ *
+ * @module CSVnVRTParser
+ */
+var CSVnVRTParser = (function _() {
+    var coord = new itowns.Coordinates('EPSG:4326');
+
+    function csv2json(csv) {
```
gchoqueux: may be you could use existing packages to convert `csv2json` and `xml2json`, to lighten the code. WDYT?

examples/js/CSVnVRTParser.js
```diff
@@ -0,0 +1,207 @@
...
+        var crs = (layer.LayerSRS && layer.LayerSRS.value) || crs;
+
+        var type = itowns.FEATURE_TYPES.POINT;
+        if (layer.GeometryType) {
+            type = getGeometryType(layer.GeometryType.value);
+        }
+
+        var feature = new itowns.Feature(type, options.crsOut, options);
+
+        if (layer.Field) {
```
gchoqueux: Could you comment this part?

examples/js/CSVnVRTParser.js
```diff
@@ -0,0 +1,207 @@
...
+                        line = data.content[i];
+                        var geometry = feature.bindNewGeometry();
+
+                        if (layer.Field) {
+                            for (var f = 0; f < layer.Field.length; f++) {
+                                geometry.properties[layer.Field[f].attributes.name] = line[layer.Field[f].attributes.pos];
+                            }
+                        }
+
+                        geometry.startSubGeometry(1);
```
gchoqueux: It's points cloud?

examples/js/CSVnVRTParser.js
zarov: I agree that this may be better, as there is a lot of things that can be taking into account to parse a CSV (more than I thought of). However I can't find something working correctly for now, I'll try to do my best about that.

examples/js/CSVnVRTParser.js
zarov: What do you mean ? If I'm not mistaken, in this combination (csv + vrt), there are only points, so I'm adding each in a geometry (or maybe I'm confused with the way `Feature` works ?)

examples/js/CSVnVRTParser.js
zarov: Do you want to have more precision on the properties read here ? I think the linked documentation (https://gdal.org/drivers/vector/vrt.html) should be ok, no ?

gchoqueux: yes

gchoqueux:
could you rebase your branch?

examples/vrt.html
```diff
@@ -0,0 +1,71 @@
...
+            // This layer is defined in a json file but it could be defined as a plain js
+            // object. See Layer* for more info.
+            itowns.Fetcher.json('./layers/JSONLayers/OPENSM.json').then(function _(config) {
+                config.source = new itowns.TMSSource(config.source);
+                var layer = new itowns.ColorLayer('OPENSM', config);
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            });
+
+            // In this case, a VRT is associated to a CSV
+            itowns.Fetcher.multiple('http://localhost/iTowns2-sample-data/vrt/velib-disponibilite-en-temps-reel', {
```
gchoqueux: where are the files?

gchoqueux:
this example  isn't added to examples page?

examples/vrt.html
zarov: Sorry, I updated everything

gchoqueux (approved):
LGTM, thanks
don't forget to squash

PR #1183, opened 2019-07-17
Refactor crs, tms and projection handling

gchoqueux:
## Description
- Extent refactoring to convert `Extent` ⇔ `Tiled Extent`.
- Add handling `Tile matrix set` in Extent.js :
   - `globalExtentTMS` contains all extents by `Tile matrix set` and `EPSG`.
   - `schemeTiles` contains all scheme by `Tile matrix set`.
- Add function to convert crs string `'EPSG:XXXX'`  ⇔ `'WMTS:XXXX'`. (`formatToEpsg` and `formatToTMS`).
- Refactoring `LayeredMaterial` and `tileBuilder` to use generic UVs : `uv_0` and `uv_1`.

## Motivation and Context
The conversions of extent was confusing and scattered. Some extents conversions didn't use Coordinates, but used `OGCWebServiceHelper.js` and `Projection.js` (they have been deleted). 
Obtaining Extents in `tileMesh.getExtentsForSource` was a set of special cases, it's now generic.

~~possibility to register new `tileMatrixSet` for WMTS.~~ 
        => **BREAKING CHANGES, it's mandatory to add projection in option to instance a `WMTSSource`   or `TMSSource`.**

## fixes:
- Wrong elevation on some tiles.
- Wrong global extent for `WMTS:PM`.
- `ColorLayer` with `WMSSource` doesn't work.
- Remove glitch lines on map.

## Screenshot 
Remove line on rasterized vector data:
**MASTER**
chrome-capture (1)
**This PR**
chrome-capture (3)


src/Core/Geographic/Crs.js
```diff
@@ -2,6 +2,50 @@ import proj4 from 'proj4';
...
+}
+
+function formatToEPSG(crs) {
+    if (crs) {
+        if (crs.includes('EPSG')) {
+            return crs;
+        } else if (EPSG[TMS.indexOf(crs)]) {
+            return EPSG[TMS.indexOf(crs)];
+        } else {
+            return `EPSG:${crs.replace('WMTS:TMS:', '')}`;
```
zarov: What about using `crs.search(/\d*/)` instead ?

src/Core/Geographic/Extent.js
```diff
@@ -530,22 +542,29 @@ class Extent {
...
-            _c.x, this.east,
-            this.south, _c.y);
-
-        return [northWest, northEast, southWest, southEast];
+        return this.subdivisionByScheme();
+    }
+    /**
+     * subdivise extent by scheme.x on west-east and scheme.y on south-north.
+     *
+     * @param      {Vector2}  [scheme=defaultScheme]  The scheme to subdivise.
```
zarov: Specify more what is a scheme here.

src/Core/Geographic/Extent.js
```diff
@@ -574,8 +593,57 @@ class Extent {
...
     }
+    /**
+     * clamp south and north values
+     *
+     * @param      {number}  [south=this.south]  The min south
+     * @param      {number}  [north=this.north]  The max north
+     * @return     {Extent}  this extent
+     */
+    clampSouthNorth(south = this.south, north = this.north) {
+        this.south = this.south > south ? this.south : south;
```
zarov: Why not use `Math.min`/`Math.max` ?

zarov (commented):
Good job on removing some parts ! It is quite a difficult PR to read, so I'll just let my few comments and do a review with you offline, to ease things.

gchoqueux:
PR updated, new review commit added.

src/Core/Geographic/Extent.js
```diff
@@ -350,6 +355,7 @@ class Extent {
      * @returns {Boolean}
      */
     intersectsExtent(extent) {
+        // TODO don't work when is on limit
```
zarov, line 358: If you wish to continue this later, please open an issue ! You can also write unit tests so we can see the use case where it fails, using `xit` instead of `it` so the test is not ran.

zarov (approved):
This seems like a quick review to outsiders, but we already did a big one offline, hence this quick validation. 

Good to me, thanks for the great work !

src/Core/Geographic/Extent.js
gchoqueux: ok, I will open issue and actually I write some new test.

PR #1188, opened 2019-07-24
chore(build): add the windows integration in travis

### LINKED ISSUE #888: Use the travis Windows integration instead of Appveyor
Windows is now supported on Travis, it would be simpler to ditch Appveyor and have only one configuration. It also seems the case that Appveyor isn't building since some times ago.

https://blog.travis-ci.com/2018-10-11-windows-early-release

zarov:
Seems like it works correctly now.

Fixes #888 

Before merging though: do we really need the Windows integration ? It greatly increases the waiting time when modifying things in a PR.


gchoqueux:
May be use test node Current and add window test.

zarov:
Done !

gchoqueux:
could you remove the `LTS linux `node?

zarov:
Done

.travis.yml
```diff
@@ -1,7 +1,12 @@
...
-  - "node"
+
+matrix:
+  include:
+    - os: linux
+      node_js: "node"
+    - os: windows
+      node_js: "node"
+
 after_success:
```
gchoqueux, line 10: you need to add test in` autodeploy.sh` to deploy only on linux os.

zarov: Well seen

gchoqueux:
Unfortunately, windows tests takes double time.

gchoqueux:
integration migrated in github action

Issue #942, opened 2018-12-02
Enable multiple sub-domains/domains template for ULRBuilder


SesticM:
Currently it's possible to set following URL for e.g. OpenStreetMap layer source: `https://a.tile.openstreetmap.org/${z}/${x}/${y}.png`

It would be beneficial (in terms of speed and efficiency) if we could set alternating sub-domains/domains on URL based on specified rule, for example:
`https://${u:a|b|c}.tile.openstreetmap.org/${z}/${x}/${y}.png`

...or, for example:
`https://${u:xyz.org|yzx.org|zxy.org}/${z}/${x}/${y}.png`

SesticM:
Enchancement related to enabling multiple sub-domains/domains template for ULRBuilder

tbroyer:
With HTTP/2, you'd better use the same server to benefit from multiplexing over the same connection. Granted, for those few, older, browsers that don't support HTTP/2 (really not that much: https://caniuse.com/http2, tl;dr: IE11 on Win7, IE mobile, and UC Browser), "sharding" that way could be beneficial (particularly in this case as the setup of those OSM servers allows for HTTP/2 connection coalescing: names resolving to the same IP and covered by the same TLS certificate).

Are IE11 on Win7, IE mobile, and UC Browser the reason you ask for this feature? Or was there any other reason?

SesticM:
Yes, exactly - IE11 on Win7 was a major reason for this request. It's a common corporate "suite" and will be hanging around for a long time...

zarov:
> `https://${u:a|b|c}.tile.openstreetmap.org/${z}/${x}/${y}.png`

I'm not familiar with this format, what are the resulting urls ?

SesticM:
It was just a staged example, no actual standards were taken into consideration (if any).

For example, resulting URLs based on following template:
`https://${u:a|b|c}.tile.openstreetmap.org/${z}/${x}/${y}.png`

...would produce following (continuously) alternating calls:
`https://a.tile.openstreetmap.org/${z}/${x}/${y}.png`
`https://b.tile.openstreetmap.org/${z}/${x}/${y}.png`
`https://c.tile.openstreetmap.org/${z}/${x}/${y}.png`

Or, for example, resulting URLs based on following template:
`https://${u:xyz.org|yzx.org}/${z}/${x}/${y}.png`

...would produce following (continuously) alternating calls:
`https://xyz.org/${z}/${x}/${y}.png`
[...]

PR #1189, opened 2019-07-24
feat(url): add sub-domains handling in url construction

zarov:
Solves #942

### DIFF
--- a/src/Provider/URLBuilder.js
+++ b/src/Provider/URLBuilder.js
@@ -1,14 +1,37 @@
 import Extent from 'Core/Geographic/Extent';
 
 const extent = new Extent('EPSG:4326', [0, 0, 0, 0]);
+
+let subDomainsCount = 0;
+function subDomains(url) {
+    const subDomainsPtrn = /\$\{u:([\w-_.|]+)\}/.exec(url);
+
+    if (!subDomainsPtrn) {
+        return url;
+    }
+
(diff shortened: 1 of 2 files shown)


src/Provider/URLBuilder.js
```diff
@@ -1,14 +1,37 @@
 import Extent from 'Core/Geographic/Extent';
 
 const extent = new Extent('EPSG:4326', [0, 0, 0, 0]);
+
+let subDomainsCount = 0;
+function subDomains(url) {
+    var subDomainsPtrn = /\$\{u:([\w-_.|]+)\}/.exec(url);
```
gchoqueux: could use const?

src/Provider/URLBuilder.js
```diff
@@ -1,14 +1,37 @@
...
+
+    if (!subDomainsPtrn) {
+        return url;
+    }
+
+    var subDomainsList = subDomainsPtrn[1].split('|');
```
gchoqueux: could use const?

gchoqueux (approved):

src/Provider/URLBuilder.js
zarov: done

src/Provider/URLBuilder.js
zarov: done

PR #1190, opened 2019-07-30
new unit tests

gchoqueux:
## Description
Add unit tests


zarov (approved):
LGTM, thanks for adding those. Not the best solution I think, as it overlaps with functional testing. However, we can't do anything else at the moment, so I'm okay with this.

PR #1194, opened 2019-08-01
fix(example): wrong move atmosphere in multi globe.

### LINKED ISSUE #1113: Broken multiglobe example
You can see on the multiglobe example that the atmosphere stays in place when switching the globe.

http://www.itowns-project.org/itowns/examples/multiglobe.html

gchoqueux:
fix(example): wrong move atmosphere in multi globe.

fix #1113

### DIFF
--- a/examples/multiglobe.html
+++ b/examples/multiglobe.html
@@ -44,6 +44,8 @@
             setupLoadingScreen(viewerDiv, view);
             object3ds.push(object3d);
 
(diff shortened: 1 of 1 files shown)


zarov:
There is still some problem with the `y ` component, which seems slightly off:

image

gchoqueux:
it's the normal behavior of the Atmosphere shader.

zarov (approved):
Ok, weird, but I won't block this PR for this.

PR #1195, opened 2019-08-01
chore: update npm packages.

gchoqueux:
## Description
THREE checking:

- [x] In the Texture Filter Constants, MipMap is now Mipmap. For example, THREE.LinearMipMapLinearFilter is now THREE.LinearMipmapLinearFilter.
- [x] Renamed WebGLRenderer.getActiveMipMapLevel() to WebGLRenderer.getActiveMipmapLevel().
- [x] WEBGL (the namespace from examples/js/WebGL.js) is now in the THREE namespace.
- [x] WEBVR (the namespace from examples/js/vr/WebVR.js) is now in the THREE namespace.
- [x] The module MapControls is now part of OrbitControls. Check out the official example for more information.
- [x] OrbitControls and MapControls now have a new default value for dampingFactor.
- [x] WebGLRenderer.context has been removed. Please use WebGLRenderer.getContext() instead.
- [x] FBXLoader now correctly sets the texture encoding. When using FBX assets in your scene, you have to set renderer.gammaOutput = true; (unless you need post-processing in linear colorspace).
- [x] When loading an FBX asset with TGA textures, FBXLoader requires now the following setup: THREE.Loader.Handlers.add( /\.tga$/i, new TGALoader() );.


zarov:
You are missing a `getContext` in `test/unit/mock.js`.

zarov (approved):

Issue #81, opened 2016-06-09
Error using "Crtl + left click" when using the Globe API in JSFiddle


vcoindet:
When we're loading the Globe API in JSFiddle, the "crtl+left click" control is not working.
It's only working after clicking on the controls panel.

Ex : http://jsfiddle.net/iTownsIGN/pxv1Lw16/

zarov:
I don't see the problem anymore, maybe it was something coming from jsFIddle.

gchoqueux:
it doesn't work on http://www.itowns-project.org/itowns/docs/#home

zarov:
Well seen. But I think it is more of a JsFiddle issue than an iTowns one, as it may something to do with iframe focusing ?

PR #1197, opened 2019-08-05
fix(controls): focus on the viewer div after view creation

zarov:
So I got in a use case similar to #81, so here's a fix !

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -185,6 +185,9 @@ class View extends THREE.EventDispatcher {
                 this._fullSizeDepthBuffer = null;
             }
         });
+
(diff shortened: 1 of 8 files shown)


zarov:
I fixed the error in CI by pushing a new commit on bad `View` and `GlobeView` calling.

gchoqueux (approved):

gchoqueux:
and if viewer loses focus?

zarov:
Unfortunately we can't cover all situations, so I don't know. I trust that people using iTowns handle this situation themselves.

gchoqueux:
ok

PR #1198, opened 2019-08-06
feature(point): option to link OrientedImageMaterial to PointsMaterial.

gchoqueux:
## Description
Add option to link `OrientedImageMaterial` to `PointsMaterial` when creating Points Layer.
I'll add an example when i'll have the datas

### DIFF
--- a/src/Renderer/PointsMaterial.js
+++ b/src/Renderer/PointsMaterial.js
@@ -2,6 +2,7 @@ import { Vector4, Uniform, NoBlending, NormalBlending, RawShaderMaterial } from 
 import PointsVS from 'Renderer/Shader/PointsVS.glsl';
 import PointsFS from 'Renderer/Shader/PointsFS.glsl';
 import Capabilities from 'Core/System/Capabilities';
+import ShaderUtils from 'Renderer/Shader/ShaderUtils';
 
 export const MODE = {
     COLOR: 0,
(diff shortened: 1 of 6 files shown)


src/Renderer/Shader/PointsFS.glsl
```diff
@@ -1,17 +1,37 @@
...
-    gl_FragColor = vColor;
+#if defined(USE_TEXTURES_PROJECTIVE)
+    if (!pickingMode) {
+        vec4 color = vColor;
+        vec4 aColor;
+        float alpha = 0.0;
 
-    #include <logdepthbuf_fragment>
+        #pragma unroll_loop
+        for (int i = 0; i < ORIENTED_IMAGES_COUNT; i++) {
```
zarov, line 20: Maybe factorize this part with `ProjectiveTextureFS.glsl` ? It is almost the same, beside the `#ifdef USE_BASE_MATERIAL`.

src/Renderer/PointsMaterial.js
```diff
@@ -42,9 +60,23 @@ class PointsMaterial extends RawShaderMaterial {
         if (__DEBUG__) {
             this.defines.DEBUG = 1;
         }
+
         this.updateUniforms();
     }
 
+    copy(source) {
+        super.copy(source);
+        if (source.uniforms.projectiveTextureAlphaBorder) {
```
zarov, line 70: Why would you need to do that ? This should be handled by the implementation of `ShaderMaterial.copy`.

src/Renderer/Shader/PointsFS.glsl
gchoqueux: Factorized

src/Renderer/PointsMaterial.js
gchoqueux: Because, it's a link material to oriented image material, it's necessary to reference these uniforms.

zarov: My bad, I understand what you are doing here !

zarov (approved):

PR #1204, opened 2019-08-19
chore(example): correct the view source link

zarov:


examples/index.html
```diff
@@ -51,10 +51,10 @@
         iframe.focus();
 
         var viewSrc = '<a href="https://github.com/iTowns/itowns/blob/master/examples/';
-        viewSrc += window.location.hash.substr(1);
+        viewSrc += window.location.hash.substr(1) || 'view_3d_map';
```
gchoqueux, line 54: The `code source` history is fixed?

gchoqueux:
I also have two comments on the examples (which can be fixed later):

- It's possible to loose the focus, so the `ctrl` key doesn't handled.
- Could you export `view` instance to debug iTowns in console?

examples/index.html
zarov: Unfortunately I couldn't make it work. Three.js, which has inspired this template, haven't find a solution too. It doesn't seems possible from what I've searched.

zarov:
> Could you export view instance to debug iTowns in console?

If you want to debug, you can still open the examples outside this view, by remove the `#` from the url and adding `.html` at its end.

> It's possible to loose the focus, so the ctrl key doesn't handled.

I can't find a solution for the moment unfortunately.

gchoqueux (approved):

PR #1205, opened 2019-08-20
feature(core): support fill pattern image for features to raster.

gchoqueux:
## Description
Support fill pattern image for features (geojson) to raster.

## Screenshots

cross


cboucheIGN:
This look great !

examples/source_file_geojson_raster.html
```diff
@@ -52,39 +52,70 @@
...
-                        name: 'ariege',
-                        transparent: true,
-                        style: {
-                            fill: {
-                              color: 'orange',
-                              opacity: 0.5,
-                            },
-                            stroke: {
-                                color:'white',
+                    return itowns.Fetcher.texture('images/cross.png').then(texture => {
```
zarov: Where is `texture` used here ?

src/Core/Style.js
```diff
@@ -26,6 +26,7 @@ class Style {
      * @param {object} [params]
      * @param {object} [params.fill] fill style.
      * @param {string} [params.fill.color] fill color string css.
+     * @param {Image|Canvas} [params.fill.pattern] fill with pattern image.
```
zarov: I would prefer allowing a `THREE.Texture`, then you can test it using `isTexture`.

src/Converter/Feature2Texture.js
```diff
@@ -40,15 +42,19 @@ function _drawPolygon(ctx, vertices, indices, style, size, extent, invCtxScale,
...
         ctx.fill();
     }
 }
 
-function fillStyle(style, ctx) {
-    if (ctx.fillStyle !== style.fill.color) {
+function fillStyle(style, ctx, invCtxScale) {
+    if (style.fill.color && ctx.fillStyle !== style.fill.color) {
         ctx.fillStyle = style.fill.color;
+    } else if (style.fill.pattern && ctx.fillStyle.src !== style.fill.pattern.src) {
```
zarov: Shouldn't this be the first option to check ?

examples/source_file_geojson_raster.html
```diff
@@ -52,39 +52,70 @@
...
+                        var ariegeLayer = new itowns.ColorLayer('pyrenees-orientales', {
+                            name: 'pyrenees-orientales',
+                            transparent: true,
+                            style: {
+                                fill: {
+                                  pattern: texture.image,
+                                  opacity: 0.8,
+                                },
+                                stroke: {
+                                    color:'IndianRed',
```
zarov, line 109: How is this affecting the visual ? Is it a fallback in case the image isn't available ?

zarov (commented):
Really nice feature to have, way less complicated than I thought it could be ! You should consider expanding a little more the `Style` documentation with examples, or even consider writing a tutorial about customisation.

examples/source_file_geojson_raster.html
gchoqueux: I use `texture fetcher` to load image, by using `texture.image`.

zarov: Yes, but you don't use it for this layer, so you can remove it.

src/Converter/Feature2Texture.js
gchoqueux: yes

src/Core/Style.js
gchoqueux: Canvas context couldn't use `THREE.Texture`.

examples/source_file_geojson_raster.html
gchoqueux: No, it's to color the stroke.

src/Core/Style.js
zarov: Yes, but below you can then use `params.fill.pattern.image`. This way the texture could be updated without having to change the style ?

examples/source_file_geojson_raster.html
zarov: I can see that in the screenshot yes, sorry.

examples/source_file_geojson_raster.html
gchoqueux: ok

zarov (approved):
LGTM, thanks

PR #1208, opened 2019-08-21
feature(3dtile): add callback called when new content is added.

### LINKED ISSUE #1165: Add an event for 3DTiles tile loading
In a project using iTowns and 3DTiles data, we need to retrieve information specific to tiles (especially the Batch Table and the Batch IDs). Ideally, this would only be done once when the 3DTiles layer is added. However, as the tiles are loaded only when needed (ie. when they need to be rendered), information such as Batch Tables & Batch IDs are mostly unavailable at this moment.

A solution would be to trigger an event when a 3DTiles tile is loaded, so that we could retrieve and store the information we need.

Do you think it would be possible to add such an event ?

gchoqueux:
## Description
add callback called when new content is added in 3dtile layer.

Warning, the documentation isn't added properly.

Fix #1165

### DIFF
--- a/src/Provider/3dTilesProvider.js
+++ b/src/Provider/3dTilesProvider.js
@@ -7,7 +7,6 @@ import Extent from 'Core/Geographic/Extent';
 import { pre3dTilesUpdate, process3dTilesNode, init3dTilesLayer } from 'Process/3dTilesProcessing';
 import utf8Decoder from 'Utils/Utf8Decoder';
 
-
 /** @classdesc
  * Class mapping 3D Tiles extensions names to their associated parsing methods.
  */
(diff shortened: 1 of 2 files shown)


src/Provider/3dTilesProvider.js
```diff
@@ -7,6 +7,29 @@ import Extent from 'Core/Geographic/Extent';
 import { pre3dTilesUpdate, process3dTilesNode, init3dTilesLayer } from 'Process/3dTilesProcessing';
 import utf8Decoder from 'Utils/Utf8Decoder';
 
+/**
+ * iTowns supports loading and displaying 3d-tiles tilesets.
+ *
+ *
+ * @typedef {Object} 3dtiles
```
zarov: What is the purpose of this section ?

src/Provider/3dTilesProvider.js
```diff
@@ -155,11 +178,35 @@ export function getObjectToUpdateForAttachedLayers(meta) {
...
+ * @property {THREE.Material} [material=THREE.PointsMaterial] - material cloned
+ * and assigned each time a points mesh is created.
+ * @property {number} [sseThreshold=16] - s(creen) s(pace) e(rror) threshold in pixels.
+ * Define how many pixels the geometricError from a tile must cover to kick the
+ * subdivision mechanism in.
+ * @property {number} [cleanupDelay=1000] - delay in milliseconds after which an
+ * undisplayed tiles will be removed from memory.
+ * @property {Function} onAddContent - callback called when new content is added.
+ */
+export class $3dtilesLayer {
```
zarov, line 177: I don't understand that either.

src/Provider/3dTilesProvider.js
```diff
@@ -155,11 +178,35 @@ export function getObjectToUpdateForAttachedLayers(meta) {
...
+}
+
 function preprocessDataLayer(layer, view, scheduler) {
     layer.preUpdate = layer.preUpdate || pre3dTilesUpdate;
     layer.update = layer.update || process3dTilesNode();
     layer.sseThreshold = layer.sseThreshold || 16;
     layer.cleanupDelay = layer.cleanupDelay || 1000;
+    layer.onAddContent = layer.onAddContent || (() => {});
```
zarov: I wouldn't use this name: it seems like a callback to be trigger after a function named `addContent`. What about something like `onContentAdded` or `onContentLoaded` ?

gchoqueux: ok to `onContentAdded`

jailln: Maybe `onTileContentLoaded` to be consistent with the standard names ?

src/Provider/3dTilesProvider.js
gchoqueux: to generate documentation

src/Provider/3dTilesProvider.js
gchoqueux: Removed

src/Provider/3dTilesProvider.js
gchoqueux: ok for `onTileContentLoaded`

zarov (approved):
LGTM, we need a cleaner way of declaring 3dtiles layer though.

PR #1211, opened 2019-08-28
fix(picking): NaN value when getting the tile under the picking position

zarov:
When `idx` was given the value `-4`, we got `buffer.slice(-4, 0)` which is not a "valid" call: it results in an empty array. Adding ` || undefined` fix this problem.

### DIFF
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -43,7 +43,7 @@ function screenCoordsToNodeId(view, tileLayer, viewCoords, radius = 0) {
 
     traversePickingCircle(radius, (x, y) => {
         const idx = (y * 2 * radius + x) * 4;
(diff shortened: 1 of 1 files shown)


gchoqueux:
In which case `idx` is -4? when x or y is negative?
if `idx is` -4 so we have `buffer.slice(-4);`, the slice returns which value?
The test fails, I'm relaunching it.

gchoqueux:
> `buffer.slice(-4);`

Ok, for more information [visiting doc](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/TypedArray/slice).

> when x or y is negative?

Yes, could be negative.

gchoqueux (approved):

PR #1218, opened 2019-09-04
Fix #1215 (Meridian convergence looks good but seems to be applied in reverse)

### LINKED ISSUE #1215: Meridian convergence looks good but seems to be applied in reverse
Meridian convergence looks good but seems to be applied in reverse ( [Pull #1212  ](https://github.com/iTowns/itowns/pull/1212))

Weird effect when using:
`var quat_crs2crs = itowns.OrientationUtils.quaternionFromCRSToCRS("EPSG:2154", "EPSG:4978")(coords);`

What worked to convert a quaternion from local (L93) to WGS 84:
```
var quaternion = new THREE.Quaternion(x,y,z,w)  // Local
var coords = new itowns.Coordinates('EPSG:4978', pos.x, pos.y, pos.z);
quaternion.premultiply(itowns.OrientationUtils.quaternionToCRS('EPSG:2154')(coords));
quaternion.premultiply(itowns.OrientationUtils.quaternionToCRS('EPSG:4978')(coords));

```
@mbredif Could you check? :)

mbredif:
this PR :
- fixes #1215
- generalizes the use of  ```quaternionFrom[*]``` and ```quaternionTo[*]``` functions. By being explicit, I expect less bugs in client code.
- reintroduces currying : similar to [proj4](http://proj4js.org/) where ```proj4(fromProjection[, toProjection, coordinates])``` returns a function when `coordinates` is undefined. This way, ```crs2crs=OrientationUtils.quaternionFromCRSToCRS('EPSG:4978', 'EPSG:4978')``` may do some precomputing and be stored to be applied later to many coordinates. These two lines are equivalent : 
```
OrientationUtils.quaternionFromCRSToCRS('EPSG:4978', 'EPSG:4978')(coord)
OrientationUtils.quaternionFromCRSToCRS('EPSG:4978', 'EPSG:4978', coord)
```

### DIFF
--- a/src/Utils/OrientationUtils.js
+++ b/src/Utils/OrientationUtils.js
@@ -3,13 +3,27 @@ import proj4 from 'proj4';
 import Coordinates from '../Core/Geographic/Coordinates';
 
 const DEG2RAD = THREE.Math.DEG2RAD;
+const matrix = new THREE.Matrix4();
+const north = new THREE.Vector3();
+const east = new THREE.Vector3();
+const axis = new THREE.Vector3().set(0, 0, 1);
+const coord = new Coordinates('EPSG:4326', 0, 0, 0);
+const euler = new THREE.Euler();
+const quat = new THREE.Quaternion();
+
+function quaternionIdentity(coordinates, target = new THREE.Quaternion()) {
+    return coordinates ? target.set(0, 0, 0, 1) : quaternionIdentity;
+}
 
 /**
  * The OrientationUtils module provides methods to compute the quaternion that
- * model a rotation provided in various conventions, including between different
- * CRS.<br/>
- * CRS-related methods return an OrientationFunction, so that you can
- * create it once and apply it efficiently to many points of origin.
+ * models a rotation defined with various conventions, including between different
+ * CRS.
+ * The local <a href="https://en.wikipedia.org/wiki/Local_tangent_plane_coordinates#Local_east,_north,_up_(ENU)_coordinates">
+ * East/North/Up frame (ENU)</a> is used as a pivot frame when computing the rotation between two distinct CRS.
+ * If the origin of the frame is undefined, CRS-related methods precompute and return a function
+ * that can be applied efficiently to many points of origin.
+ * Otherwise, the target quaternion is returned at the provided origin coordinates.
  *
  * @example
  * // Compute the rotation around the point of origin from a frame aligned with Lambert93 axes (epsg:2154),
@@ -45,14 +59,14 @@ export default {
      * @param {number} [heading=0] - angle in degrees
      * @param {THREE.Quaternion} [target=new THREE.Quaternion()] - output Quaternion
      *
-     * @return {THREE.Quaternion} target
+     * @return {THREE.Quaternion} target quaternion
      */
     quaternionFromRollPitchHeading(roll = 0, pitch = 0, heading = 0, target = new THREE.Quaternion()) {
         roll *= DEG2RAD;
         pitch *= DEG2RAD;
         heading *= DEG2RAD;
-        // return this.setFromEuler(new THREE.Euler(pitch, roll, heading , 'ZXY')).conjugate();
-        return target.setFromEuler(new THREE.Euler(-pitch, -roll, -heading, 'YXZ')); // optimized version of above
+        // return this.setFromEuler(euler.set(pitch, roll, heading , 'ZXY')).conjugate();
+        return target.setFromEuler(euler.set(-pitch, -roll, -heading, 'YXZ')); // optimized version of above
     },
 
     /**
@@ -73,13 +87,13 @@ export default {
      * @param {number} [kappa=0] - angle in degrees
      * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
      *
-     * @return {THREE.Quaternion} target
+     * @return {THREE.Quaternion} target quaternion
      */
     quaternionFromOmegaPhiKappa(omega = 0, phi = 0, kappa = 0, target = new THREE.Quaternion()) {
         omega *= DEG2RAD;
         phi *= DEG2RAD;
         kappa *= DEG2RAD;
-        target.setFromEuler(new THREE.Euler(omega, phi, kappa, 'XYZ'));
+        target.setFromEuler(euler.set(omega, phi, kappa, 'XYZ'));
         target.set(target.w, target.z, -target.y, -target.x); // <=> target.multiply(new THREE.Quaternion(1, 0, 0, 0));
         return target;
     },
@@ -91,7 +105,7 @@ export default {
      * @param {Attitude} attitude - Attitude
      * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
      *
-     * @return {THREE.Quaternion} target
+     * @return {THREE.Quaternion} target quaternion
      */
     quaternionFromAttitude(attitude, target = new THREE.Quaternion()) {
         if ((attitude.roll !== undefined) || (attitude.pitch !== undefined) || (attitude.heading !== undefined)) {
@@ -104,17 +118,16 @@ export default {
     },
 
     /**
-     * Functions of type OrientationFunction accept arguments `(origin, target)`
-     * and return the quaternion that models a rotation around the point of
-     * origin. If target is not provided, a new quaternion is created and
-     * returned instead.
+     * FunctionOrQuaternion is either a THREE.Quaternion or a function that accepts
+     * arguments `(coordinates, target)` and returns the quaternion that models a rotation
+     * around the point of origin. If target is not provided, a new quaternion is
+     * created and returned instead.
      *
-     * @typedef {function} OrientationFunction
+     * @typedef {function|THREE.Quaternion} FunctionOrQuaternion
      *
-     * @property {Coordinates} origin the origin of the local East North Up
+     * @property {Coordinates} coordinates the origin of the local East North Up
      * (ENU) frame
-     * @property {THREE.Quaternion} target output Quaternion (default: `new
-     * THREE.Quaternion()`)
+     * @property {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion.
     */
 
     /**
@@ -127,17 +140,18 @@ export default {
      */
 
     /**
-     * Set the quaternion according to the rotation from the East North Up (ENU)
+     * Set the quaternion according to the rotation from the local East North Up (ENU)
      * frame to the geocentric frame. The up direction of the ENU frame is
      * provided by the normalized geodetic normal of the provided coordinates
      * (geodeticNormal property).
      *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates] the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
      */
-    quaternionToGeocent() {
-        const matrix = new THREE.Matrix4();
-        const north = new THREE.Vector3();
-        const east = new THREE.Vector3();
+    quaternionFromEnuToGeocent(coordinates, target = new THREE.Quaternion()) {
+        if (coordinates) { return this.quaternionFromEnuToGeocent()(coordinates, target); }
         return (coordinates, target = new THREE.Quaternion()) => {
             const up = coordinates.geodesicNormal;
             if (up.x == 0 && up.y == 0) {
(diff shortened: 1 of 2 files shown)


nosy-b:
All good now!

frejus

src/Utils/OrientationUtils.js
```diff
@@ -4,12 +4,17 @@ import Coordinates from '../Core/Geographic/Coordinates';
 
 const DEG2RAD = THREE.Math.DEG2RAD;
 
+function quaternionIdentity(coordinates, target = new THREE.Quaternion()) {
+    return coordinates ? target.set(0, 0, 0, 1) : quaternionIdentity;
```
gchoqueux, line 15: you could add `qIdentity = new THREE.Quaternion()` and replace all `target.set(0, 0, 0, 1)` by `target.copy(qIdentity)`

src/Utils/OrientationUtils.js
```diff
@@ -132,9 +136,13 @@ export default {
...
      *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates] the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
      */
-    quaternionToGeocent() {
+    quaternionToGeocent(coordinates, target = new THREE.Quaternion()) {
+        if (coordinates) { return this.quaternionToGeocent()(coordinates, target); }
```
gchoqueux: could you take these variables (matrix, north, east) out of the function, to avoid multi instances.

src/Utils/OrientationUtils.js
```diff
@@ -151,6 +159,24 @@ export default {
         };
     },
 
+    /**
+     * Set the quaternion according to the rotation from the East North Up (ENU)
+     * frame to the geocentric frame. The up direction of the ENU frame is
```
gchoqueux: > the geocentric frame

you're talking about the `Coordinates system EPSG:4978`?

src/Utils/OrientationUtils.js
```diff
@@ -160,10 +186,13 @@ export default {
...
-     *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
     */
-    quaternionToLCC(proj) {
+    quaternionFromLCC(proj, coordinates, target = new THREE.Quaternion()) {
+        if (coordinates) { return this.quaternionFromLCC(proj)(coordinates, target); }
```
gchoqueux: could you declare outside the function `axis` and `coord`?

src/Utils/OrientationUtils.js
```diff
@@ -173,6 +202,26 @@ export default {
...
+     * @param {number} proj.long0 - the longitude of the central meridian
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
+    */
+    quaternionToLCC(proj, coordinates, target = new THREE.Quaternion()) {
+        if (coordinates) { return this.quaternionToLCC(proj)(coordinates, target); }
+        const fromLCC = this.quaternionFromLCC(proj);
+        return (coordinates, target = new THREE.Quaternion()) => fromLCC(coordinates, target).conjugate();
```
gchoqueux, line 228: could remove new THREE.Quaternion() is already handling in `quaternionFromLCC`?

src/Utils/OrientationUtils.js
```diff
@@ -160,10 +186,13 @@ export default {
...
      * @param {number} proj.long0 - the longitude of the central meridian
-     *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
     */
-    quaternionToLCC(proj) {
+    quaternionFromLCC(proj, coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: The parameter `proj` may not be useless if we use the `crs coordinates` to find it.

src/Utils/OrientationUtils.js
```diff
@@ -173,6 +202,26 @@ export default {
...
+     *
+     * @param {Projection} proj the lcc projection (may be parsed using proj4)
+     * @param {number} proj.lat0 - the latitude of origin
+     * @param {number} proj.long0 - the longitude of the central meridian
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
+    */
+    quaternionToLCC(proj, coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: The parameter `proj` may not be useless if we use the `crs coordinates` to find it.

src/Utils/OrientationUtils.js
```diff
@@ -186,9 +235,13 @@ export default {
...
      * @param {number} proj.long0 - the longitude of the central meridian
      *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
     */
-    quaternionToTMerc(proj) {
+    quaternionFromTMerc(proj, coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: The parameter `proj` may not be useless if we use the `crs coordinates` to find it.

src/Utils/OrientationUtils.js
```diff
@@ -211,6 +264,54 @@ export default {
...
+     * @param {number} proj.a - the semimajor radius of the ellipsoid axis
+     * @param {number} proj.b - the semiminor radius of the ellipsoid axis
+     * @param {number} proj.long0 - the longitude of the central meridian
+     *
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
+    */
+    quaternionToTMerc(proj, coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: same comment with `proj`

src/Utils/OrientationUtils.js
```diff
@@ -233,15 +337,20 @@ export default {
...
      * proj4-compatible object.
      *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
      */
-    quaternionToCRS(crsOrProj) {
+    quaternionToCRS(crsOrProj, coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: May be replace by  `quaternionToCoordinates` and so, and use Coordinates.crs to find `crsOrProj`.

src/Utils/OrientationUtils.js
```diff
@@ -252,11 +361,21 @@ export default {
...
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates]  coordinates the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
      */
-    quaternionFromCRS(crsOrProj) {
-        const toCrs = this.quaternionToCRS(crsOrProj);
-        return (coordinates, target = new THREE.Quaternion()) => toCrs(coordinates, target).conjugate();
+    quaternionFromCRS(crsOrProj, coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: May be replace by  `quaternionFromCoordinates` and so, and use Coordinates.crs to found `crsOrProj`.

gchoqueux (commented):
Thanks for these features. 
After merge this PR, I would apply them on immersive example.

src/Utils/OrientationUtils.js
mbredif: what's the point ? is it for code self-documentation only or is there any efficiency speedups implied ?

src/Utils/OrientationUtils.js
mbredif: possibly, but not necessarily, I am talking about any geocentric CRS (with the 'geocent' type in proj4)

src/Utils/OrientationUtils.js
mbredif: I prefer my way, to make the code self-documenting (otherwise you have to look at fromLCC defaults to know what's the defaults of the returned function).

src/Utils/OrientationUtils.js
mbredif: do you mean "may be useless"?
In general, I do not want to require the coordinates to be in the proj CRS.

src/Utils/OrientationUtils.js
mbredif: I do not like ` quaternionToCoordinates ` as a function name, because it computes the quaternion modeling the rotation from the local ENU frame **to a CRS**, rather than to the coordinates of a 3D point (which is only the origin of the local frame).

mbredif:
I chose to make the temp variables in the currying functions to limit the number of global variables to hold temp values. This is a tradeoff of code simplicity/readability against the cost of constructing these objects once per construction of a curried function rather than once per application.

src/Utils/OrientationUtils.js
gchoqueux: just to avoid hard-coded values

src/Utils/OrientationUtils.js
gchoqueux: Ok, just to be sure, we agree that other geocentric projections cannot be displayed in itowns?

src/Utils/OrientationUtils.js
gchoqueux: ~~you could write~~
~~`quaternionFromCoordinate(coord)` and use `coords.crs` to map with `Proj4.projection`~~
~~and if  the coordinates isn't in the proj CRS:~~
~~`quaternionFromCoordinate(coord.as('EPSG:XXXX'))`~~

I just read the comment below

gchoqueux:
> I chose to make the temp variables in the currying functions to limit the number of global variables to hold temp values. This is a tradeoff of code simplicity/readability against the cost of constructing these objects once per construction of a curried function rather than once per application.

We avoid as much as possible because the garbage collector is slowness. In addition, you can use the same variables because the functions aren't in concurrency.

src/Utils/OrientationUtils.js
gchoqueux: ok, my bad, I will come back to the reviews because I didn't understand the role of the functions

src/Utils/OrientationUtils.js
mbredif: You know better ;).

src/Utils/OrientationUtils.js
mbredif: However, the doc of [THREE.Quaternion()](https://threejs.org/docs/#api/en/math/Quaternion) does not mention a default to be the identity quaternion, so, for correctness, you end up creating `qIdentity = new THREE.Quaternion(0,0,0,1)`, which is an extra useless variable with hard coded values :). Are you sure ?

gchoqueux: Actually, it's not very important

src/Utils/OrientationUtils.js
gchoqueux: Ok, we could see that in a future re-factorization.

src/Utils/OrientationUtils.js
```diff
@@ -104,17 +109,16 @@ export default {
...
+     * arguments `(coordinates, target)` and returns the quaternion that models a rotation
+     * around the point of origin. If target is not provided, a new quaternion is
+     * created and returned instead.
      *
-     * @typedef {function} OrientationFunction
+     * @typedef {function|THREE.Quaternion} FunctionOrQuaternion
      *
-     * @property {Coordinates} origin the origin of the local East North Up
+     * @property {Coordinates} coordinates the origin of the local East North Up
      * (ENU) frame
```
gchoqueux, line 129: In introduction, could add this link to ENU documentation  https://en.wikipedia.org/wiki/Local_tangent_plane_coordinates#Local_east,_north,_up_(ENU)_coordinates

src/Utils/OrientationUtils.js
```diff
@@ -132,9 +136,13 @@ export default {
...
      * (geodeticNormal property).
      *
-     * @return {OrientationFunction} the orientation function
+     * @param {Coordinates} [coordinates] the origin of the local East North Up
+     * (ENU) frame
+     * @param {THREE.Quaternion} [target=new THREE.Quaternion()] output Quaternion
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
      */
-    quaternionToGeocent() {
+    quaternionToGeocent(coordinates, target = new THREE.Quaternion()) {
```
gchoqueux: To clarify the use of `From` and `to`
What do you think to rely on `quaternionFromCRSToCRS` name:

`quaternionFromEnuToGeocent` (or `quaternionFromEnuToGeocent`)
`quaternionFromGeocentToEnu`...

src/Utils/OrientationUtils.js
```diff
@@ -151,6 +159,24 @@ export default {
         };
     },
 
+    /**
+     * Set the quaternion according to the rotation from the East North Up (ENU)
```
gchoqueux: it's the same definition that `quaternionToGeocent`

src/Utils/OrientationUtils.js
```diff
@@ -173,6 +202,26 @@ export default {
         };
     },
 
+    /**
+     * Set the quaternion to correct for the
+     * <a href="https://geodesie.ign.fr/contenu/fichiers/documentation/algorithmes/alg0060.pdf">meridian convergence</a>
```
gchoqueux, line 213: same that doc `quaternionFromLCC`

src/Utils/OrientationUtils.js
```diff
@@ -211,6 +264,54 @@ export default {
         };
     },
 
+    /**
+     * Set the quaternion to correct for the
```
gchoqueux: same documentation that `quaternionFromTMerc`

src/Utils/OrientationUtils.js
```diff
@@ -211,6 +264,54 @@ export default {
...
+     * @return {FunctionOrQuaternion} The target quaternion if coordinates is defined, otherwise, a function to compute it from coordinates.
+    */
+    quaternionToTMerc(proj, coordinates, target = new THREE.Quaternion()) {
+        if (coordinates) { return this.quaternionToTMerc(proj)(coordinates, target); }
+        const fromTMerc = this.quaternionFromTMerc(proj);
+        return (coordinates, target = new THREE.Quaternion()) => fromTMerc(coordinates, target).conjugate();
+    },
+
+    /**
+     * Returns the identity quaternion (0,0,0,1), as longlat and ENU frame are aligned (?).
```
gchoqueux: `(?)` it's not sure?

src/Utils/OrientationUtils.js
mbredif: done

src/Utils/OrientationUtils.js
mbredif: done

src/Utils/OrientationUtils.js
mbredif: I am not expert enough in geodesy to answer that.
I put a more explicit doc : "The identity quaternion (0,0,0,1) is returned, as longlat and ENU frame are assumed to be aligned."

src/Utils/OrientationUtils.js
mbredif: done

src/Utils/OrientationUtils.js
mbredif: done

src/Utils/OrientationUtils.js
mbredif: done

src/Utils/OrientationUtils.js
mbredif: done

src/Utils/OrientationUtils.js
mbredif: done

mbredif:
Thanks for the review. I updated the PR accordingly

src/Utils/OrientationUtils.js
gchoqueux: ok, we'll assume, we'll find out.

gchoqueux (approved):
Thanks,
Don't forget to squash or rename commit before merging.

gchoqueux:
@mbredif just a simply reminder to not forget the last step.

mbredif:
I did not understand it was a task for me :)

mbredif:
ok now?

gchoqueux:
yes

PR #1219, opened 2019-09-04
feat(view): add a pickFeaturesAt method in View

zarov:
This new method comes in addition to `pickObjectsAt`: it looks for all
features under the given screen coordinate, and return it sorted by
layer.

In the same time, `FeatureToolTip` has been greatly improved.

### DIFF
--- a/examples/js/plugins/FeatureToolTip.js
+++ b/examples/js/plugins/FeatureToolTip.js
@@ -1,150 +1,229 @@
 /* global itowns */
 /**
  * A tooltip that can display some useful information about a feature when
- * hovering it. Only works for layers using FileSource.
+ * hovering it.
  *
- * @param {View} viewer - The view to bind the tooltip to.
- * @param {Object} options - Options to have more custom content displayed.
- * @param {number} [options.precisionPx=5] - The precision of the picking.
- * @param {function} [options.format] - A function that takes the name of the
- * property currently being processed and its value, and gives the appropriate
- * HTML output to it. If this method is specified, no others properties other
- * than the ones handled in it will be displayed.
- * @param {Array<string>} [options.filter] - An array of properties to filter.
- * @param {boolean} [options.filterAll=true] - Filter all the properties.
- * Default to true.
- *
- * @class FeatureToolTip
+ * @module FeatureToolTip
  *
  * @example
- * view.addEventListener(itowns.VIEW_EVENTS.LAYERS_INITIALIZED, function() {
- *      new FeatureToolTip(view);
- * });
+ * // Initialize the FeatureToolTip
+ * FeatureToolTip.init(viewerDiv, view);
+ *
+ * // Add layers
+ * var wfsSource = new itowns.WFSSource(...);
+ * var wfsLayer = new itowns.ColorLayer(wfsSource...);
+ * view.addLayer(wfsLayer);
+ *
+ * var fileSource = new itowns.FileSource(...);
+ * var fileLayer = new itowns.GeometryLayer(fileSource...);
+ * view.addLayer(fileLayer);
+ *
+ * FeatureToolTip.addLayer(wfsLayer);
+ * FeatureToolTip.addLayer(fileLayer);
  */
-function FeatureToolTip(viewer, options) {
-    var opts = options || { filterAll: true };
-    opts.filter = opts.filter == undefined ? [] : opts.filter;
-
-    var tooltip = document.createElement('div');
-    tooltip.className = 'tooltip';
-    viewer.mainLoop.gfxEngine.renderer.domElement.parentElement.appendChild(tooltip);
+var FeatureToolTip = (function _() {
+    var tooltip;
+    var view;
+    var layers = [];
+    var layersId = [];
 
     var mouseDown = 0;
-    document.body.onmousedown = function onmousedown() {
+    document.body.addEventListener('mousedown', function _() {
         ++mouseDown;
-    };
-    document.body.onmouseup = function onmouseup() {
+    }, false);
+    document.body.addEventListener('mouseup', function _() {
         --mouseDown;
-    };
+    }, false);
 
-    var layer;
-    var result;
-    var feature;
-    var symb;
-    var style;
-    var fill;
-    var stroke;
-    var content;
-    var prop;
-    var layers = viewer.getLayers(function _(l) { return l.source && l.source.isFileSource; });
-    function buildToolTip(geoCoord, e) {
-        var visible = false;
-        var i = 0;
-        var p = 0;
-        var precision = viewer.controls.pixelsToDegrees(opts.precisionPx || 5);
-
+    function moveToolTip(event) {
         tooltip.innerHTML = '';
         tooltip.style.display = 'none';
-        if (geoCoord) {
-            visible = false;
 
-            // Pick on each layer, and display them all in the tooltip
-            for (i = 0; i < layers.length; i++) {
-                layer = layers[i];
+        var features = view.pickFeaturesAt(event, 3, layersId);
 
-                if (!layer.source.parsedData) { continue; }
+        var layer;
+        for (var layerId in features) {
+            if (features[layerId].length == 0) {
+                continue;
+            }
 
-                result = itowns.FeaturesUtils.filterFeaturesUnderCoordinate(geoCoord, layer.source.parsedData, precision);
+            layer = layers[layersId.indexOf(layerId)];
+            if (typeof layer.options.filterGeometries == 'function') {
+                features[layerId] = layer.options.filterGeometries(features[layerId], layer.layer) || [];
+            }
+            tooltip.innerHTML += fillToolTip(features[layerId], layer.layer, layer.options);
+        }
 
-                for (p = 0; p < result.length; p++) {
-                    content = '';
-                    visible = true;
+        if (tooltip.innerHTML != '') {
+            tooltip.style.display = 'block';
+            tooltip.style.left = view.eventToViewCoords(event).x + 'px';
+            tooltip.style.top = view.eventToViewCoords(event).y + 'px';
+        }
+    }
 
-                    feature = result[p].geometry;
-                    style = layer.style.isStyle ? layer.style : feature.properties.style;
-                    fill = style.fill.color;
-                    stroke = '1.25px ' + style.stroke.color;
+    function fillToolTip(features, layer, options) {
+        var content = '';
+        var feature;
+        var geometry;
+        var style;
+        var fill;
+        var stroke;
+        var symb = '';
+        var prop;
 
-                    if (result[p].type === itowns.FEATURE_TYPES.POLYGON) {
-                        symb = '&#9724';
-                    } else if (result[p].type === itowns.FEATURE_TYPES.LINE) {
-                        symb = '&#9473';
-                        fill = style.stroke.color;
-                        stroke = '0px';
-                    } else if (result[p].type === itowns.FEATURE_TYPES.POINT) {
-                        symb = '&#9679';
+        for (var p = 0; p < features.length; p++) {
+            feature = features[p];
+
+            geometry = feature.geometry;
+            style = (layer.style && layer.style.isStyle) ? layer.style : geometry.properties.style;
+            fill = style.fill.color;
+            stroke = '1.25px ' + style.stroke.color;
+
+            if (feature.type === itowns.FEATURE_TYPES.POLYGON) {
+                symb = '&#9724';
+            } else if (feature.type === itowns.FEATURE_TYPES.LINE) {
+                symb = '&#9473';
+                fill = style.stroke.color;
+                stroke = '0px';
+            } else if (feature.type === itowns.FEATURE_TYPES.POINT) {
+                symb = '&#9679';
+            }
+
+            content += '<div>';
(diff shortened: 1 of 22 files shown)


examples/js/plugins/FeatureToolTip.js
```diff
@@ -1,150 +1,224 @@
...
-                    }
+        var layer;
+        for (var layerId in features) {
+            if (features[layerId].length == 0) {
+                continue;
+            }
+
+            layer = layers[layersId.indexOf(layerId)];
+            if (typeof layer.options.filterGeometries == 'function') {
+                features[layerId] = layer.options.filterGeometries(features[layerId], layer.layer);
```
gchoqueux: I prefer that the user gives a function `getIdFromProperties(properties)` and that we develop with a filtering function.
May be throw a error if this function isn't exist.

examples/js/plugins/FeatureToolTip.js
```diff
@@ -1,150 +1,224 @@
...
+            content += symb + ' ';
+            content += '</span>';
+            content += (geometry.properties.name || geometry.properties.nom || geometry.properties.description || layer.name || '');
+
+            if (feature.type === itowns.FEATURE_TYPES.POINT) {
+                content += '<br/><span class="coord">long ' + feature.coordinates[0].toFixed(4) + '</span>';
+                content += '<br/><span class="coord">lat ' + feature.coordinates[1].toFixed(4) + '</span>';
+            }
+
+            if (geometry.properties && !options.filterAllProperties) {
```
gchoqueux, line 103: `filterAllProperties` could explain this option?

src/Core/View.js
```diff
@@ -636,6 +637,75 @@ class View extends THREE.EventDispatcher {
...
+
+        let precision;
+        const precisions = {
+            PX: radius,
+            M: this.isGlobeView && this.controls ? this.controls.pixelsToMeters(radius) : radius,
+            D: this.isGlobeView && this.controls ? this.controls.pixelsToDegrees(radius) : 0,
+        };
+
+        // Get the mouse coordinates to the correct system
+        const positionVector = new THREE.Vector3();
```
gchoqueux: could you instance the `positionVector` and `position` before function?

src/Core/View.js
```diff
@@ -636,6 +637,75 @@ class View extends THREE.EventDispatcher {
...
+        layers = layers.map((l) => {
+            const id = l.isLayer ? l.id : l;
+            result[id] = [];
+            return id;
+        });
+
+        let precision;
+        const precisions = {
+            PX: radius,
+            M: this.isGlobeView && this.controls ? this.controls.pixelsToMeters(radius) : radius,
```
gchoqueux: the precision in meter in wrong for planar view.

src/Core/View.js
```diff
@@ -636,6 +637,75 @@ class View extends THREE.EventDispatcher {
...
+        const positionVector = new THREE.Vector3();
+        const mouse = (mouseOrEvt instanceof Event) ? this.eventToViewCoords(mouseOrEvt) : mouseOrEvt;
+        this.getPickingPositionFromDepth(mouse, positionVector);
+        const position = new Coordinates(this.referenceCrs, positionVector);
+
+        // Get the tile corresponding to where the cursor is
+        const tiles = Picking.pickTilesAt(this, mouse, precisions.PX, this.tileLayer);
+
+        for (const t in tiles) {
+            if (!tiles[t].object) {
```
gchoqueux: `!tiles[t].object` What is this case?

src/Core/View.js
```diff
@@ -636,6 +637,75 @@ class View extends THREE.EventDispatcher {
...
+                continue;
+            }
+
+            for (const materialLayer of tiles[t].object.material.layers) {
```
gchoqueux: could use loop on `layers` and use `material.getLayer(id)` to get `parsedData`

src/Core/View.js
```diff
@@ -636,6 +637,75 @@ class View extends THREE.EventDispatcher {
...
+                }
+
+                for (const texture of materialLayer.textures) {
+                    if (!texture.parsedData) {
+                        continue;
+                    }
+
+                    precision = CRS.isMetricUnit(texture.parsedData.crs) ? precisions.M : precisions.D;
+                    FeaturesUtils.filterFeaturesUnderCoordinate(position, texture.parsedData, precision)
+                        .forEach(f => result[materialLayer.id].push(f));
```
gchoqueux: could use `Array.concat`?

gchoqueux (commented):
Thanks, the performance is good.
After this PR, could you do a PR to apply this feature on vector tile?

examples/js/plugins/FeatureToolTip.js
zarov: The idea here is that the user could filter the features as he wishes, not necessary by using the id.

examples/js/plugins/FeatureToolTip.js
zarov: This is the previous `filterAll` option (see the documentation below)

src/Core/View.js
zarov: Yes !

src/Core/View.js
zarov: Yes, I forgot to look at this

src/Core/View.js
zarov: None, I'm removing it, it something I forgot

src/Core/View.js
zarov: Ok

src/Core/View.js
zarov: Done

zarov:
Updated

examples/js/plugins/FeatureToolTip.js
gchoqueux: ok

examples/js/plugins/FeatureToolTip.js
gchoqueux: yes, the documentation doesn't give a real explanation.

src/Core/View.js
gchoqueux: `for (const material of tile.object.material.getLayers(layers)) {`

just rename  `material` to `materialLayer`

examples/js/plugins/FeatureToolTip.js
zarov: Yes, updated

src/Core/View.js
zarov: Done

zarov:
I rebased, and added a commit that refactor the whole meter/degree/pixel system to a more simple and generic one. I also marked some methods in `GlobeControls` as deprecated, as I think we can remove them in the next release (or the one after).

I will add some functional testing to really validate those conversions, but for now it seems to give correct values.

zarov:
Bad news, the `getMetersToPixels` and `getPixelsToMeters` don't return the same value, I'll get back to it.

src/Utils/FeaturesUtils.js
```diff
@@ -144,8 +145,23 @@ export default {
 
         // We can take this shortcut because either Feature and
         // FeatureCollection have an extent property
-        if (features.extent && !features.extent.isPointInside(coordinate, epsilon)) {
-            return result;
+        if (features.extent) {
+            // Special case, because of the way tiles in VectorTileParser are
+            // handled (see Feature2Texture for a similar solution)
+            if (features.extent.crs == 'WMTS:PM') {
+                const ex = new Extent(coordinate.crs, 0, 0, 0, 0);
```
gchoqueux: could you instance ex before function?

src/Utils/FeaturesUtils.js
```diff
@@ -144,8 +145,23 @@ export default {
 
         // We can take this shortcut because either Feature and
         // FeatureCollection have an extent property
-        if (features.extent && !features.extent.isPointInside(coordinate, epsilon)) {
-            return result;
+        if (features.extent) {
+            // Special case, because of the way tiles in VectorTileParser are
+            // handled (see Feature2Texture for a similar solution)
+            if (features.extent.crs == 'WMTS:PM') {
```
gchoqueux: it isn't the PM case, it's the case when features have scale or/and translation transformation.

src/Core/View.js
```diff
@@ -636,6 +654,120 @@ class View extends THREE.EventDispatcher {
...
+     * @param {number} [pixels=1] - The size, in pixels, to get in meters.
+     * @param {THREE.Vector3} [screenCoord] - The screen coordinate to get the
+     * projected distance at. By default, this is the middle of the screen.
+     *
+     * @return {number} The projected distance in meters.
+     */
+    getPixelsToMeters(pixels = 1, screenCoord) {
+        const br = this.mainLoop.gfxEngine.renderer.domElement.getBoundingClientRect();
+
+        let _screenCoord = (screenCoord instanceof Event) ? this.eventToViewCoords(screenCoord) : screenCoord;
```
gchoqueux: Could you use `const meter = pixel * D / this.camera._preSSE;` instead of?

src/Core/Prefab/GlobeView.js
```diff
@@ -151,6 +152,14 @@ class GlobeView extends View {
 
         return super.addLayer(layer, this.tileLayer);
     }
+
+    getPixelsToDegrees(pixels, screenCoord) {
```
gchoqueux: There's no need a specific function for `GlobeView`

gchoqueux (changes requested):
It's work perfectly.

src/Core/Prefab/GlobeView.js
zarov: Why ?

src/Core/View.js
zarov: Done

src/Utils/FeaturesUtils.js
zarov: done

src/Utils/FeaturesUtils.js
zarov: done

zarov:
Rebased and updated !

src/Controls/GlobeControls.js
```diff
@@ -1105,6 +1105,9 @@ GlobeControls.prototype.setZoom = function setZoom(zoom, isAnimated) {
  * This function compute the scale of a map
  * @param      {number}  pitch   Screen pitch, in millimeters ; 0.28 by default
  * @return     {number}  The zoom scale.
+ *
+ * @deprecated To remove with pixelsToMeters, pixelsToDegrees and
```
gchoqueux: `getScale` will have to be moved to view

src/Core/View.js
```diff
@@ -636,6 +640,123 @@ class View extends THREE.EventDispatcher {
...
+     * @return {number} The distance in meters.
+     */
+    getDistanceFromCamera(screenCoord) {
+        const br = this.mainLoop.gfxEngine.renderer.domElement.getBoundingClientRect();
+
+        let _screenCoord = (screenCoord instanceof Event) ? this.eventToViewCoords(screenCoord) : screenCoord;
+        if (!_screenCoord) {
+            // TODO: This is totally arbitrary, there can be nothing here. Maybe
+            // find a better solution for this.
+            _screenCoord = new THREE.Vector2(br.width / 2, br.height / 2);
```
gchoqueux: could you use `screen` instead of new instance.

src/Core/View.js
```diff
@@ -636,6 +640,123 @@ class View extends THREE.EventDispatcher {
...
+        }
+
+        this.getPickingPositionFromDepth(_screenCoord, positionVector);
+        position.setFromVector3(positionVector);
```
gchoqueux: why you copy `position` from `positionVector`?

src/Utils/FeaturesUtils.js
```diff
@@ -144,8 +147,33 @@ export default {
...
+            // handled (see Feature2Texture for a similar solution)
+            if (features.translation || features.scale) {
+                ex.crs = coordinate.crs;
+                features.extent.as(coordinate.crs, ex);
+                if (!ex.isPointInside(coordinate, epsilon)) {
+                    return result;
+                }
+
+                if (!features.translation) {
+                    features.translation = new THREE.Vector3();
```
gchoqueux: could you avoid new instance?

src/Core/View.js
zarov: done

src/Controls/GlobeControls.js
zarov: done

src/Core/View.js
zarov: to be used in `pickFeaturesAt` but I modified that

src/Utils/FeaturesUtils.js
zarov: done

src/Controls/GlobeControls.js
gchoqueux: use `view.getScale()` inside `GlobeControls.getScale()`

src/Utils/FeaturesUtils.js
gchoqueux: this is useless `features.translation` ans `features.scale` have default values.

zarov:
Rebased and cleaned !

gchoqueux (approved):

PR #1226, opened 2019-09-10
fix(style): add circle style support.

gchoqueux:
## Description
add support to circle style

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -112,6 +112,11 @@ class Style {
             this.point.color = color;
             this.point.opacity = opacity;
             this.point.radius = 1.5;
+        } else if (layer.type === 'circle') {
+            const { color, opacity } = rgba2rgb(layer.paint['circle-color']);
+            this.point.color = color;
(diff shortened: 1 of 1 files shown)


src/Core/Style.js
```diff
@@ -112,6 +112,11 @@ class Style {
             this.point.color = color;
             this.point.opacity = opacity;
             this.point.radius = 1.5;
+        } else if (layer.type === 'circle') {
+            const { color, opacity } = rgba2rgb(layer.paint['circle-color']);
+            this.point.color = color;
+            this.point.opacity = opacity;
+            this.point.radius = layer.paint['circle-radius'];
```
zarov, line 119: No default value ?

gchoqueux: if type is circle `color` and `radius` are defined.
`rgba2rgb` returns 1 to opacity by default.

zarov: Ok !

zarov (approved):

PR #1228, opened 2019-09-10
chore(example): add debug ui in some examples.

gchoqueux:
## Description
add debug ui in some examples.


zarov:
Suggestion: why not put it and harmonize it in all examples ? Or we could also have to only include `debug.js` and it will handle the `if (view.isDebugMode...` part, with no need to write it ourselves. WDYT ?

gchoqueux:
yes good idea

gchoqueux:
Pr updated

zarov (approved):

Issue #1063, opened 2019-03-07
Inconsistency in DataSourceProvider if wrong extent


zarov:
## Context

As said in [#1033](https://github.com/iTowns/itowns/pull/1033#discussion_r262529859), there is a problem in the parsing of data to features.

As we can see in [`DataSourceProvider`](), there is no utility in having `EPSG:4978` as an extent, as it is a geocentric extent and makes no sense to use it here. This case can occur if `buildExtent = true` and `crsOut = 'EPSG:4978'`, which is fortunately for now not the case for any of our samples.

## Possible fixes

In parsers, if `crsOut = 'EPSG:4978'`, we may need to force `crsOut` to switch to `crsIn`, or something appropriate. However, we don't control all parsers for iTowns, so something else may need to be found, something else than the simple check there is for now. In the meantime, checks and warning messages can be added.

jailln:
DataSourceProvider no longer has this check.

PR #1230, opened 2019-09-11
Refactor source and dataSourceProvider

gchoqueux:
## Description
- handling fetcher/parser in class `Source.`
- clean up `DataSourceProvider`.


src/Source/Source.js
```diff
@@ -1,4 +1,26 @@
 import Extent from 'Core/Geographic/Extent';
+import GeoJsonParser from 'Parser/GeoJsonParser';
+import KMLParser from 'Parser/KMLParser';
+import GpxParser from 'Parser/GpxParser';
+import VectorTileParser from 'Parser/VectorTileParser';
+import Fetcher from 'Provider/Fetcher';
+
+export const supportedFetchers = new Map([
```
zarov, line 8: Note: in the future, we should have these two lists in a separate file, where we can put the format itowns supports by default.

src/Source/FileSource.js
```diff
@@ -136,6 +136,14 @@ class FileSource extends Source {
         return this.url;
     }
 
+    onParsedFile(parsedFile) {
+        this.parsedData = parsedFile;
+        if (parsedFile.extent && parsedFile.extent.crs != 'EPSG:4978') {
```
zarov, line 141: I know it's me that put this filter here, but I don't really remember the case where it is necessary. Do you know if it is still necessary ?

src/Provider/DataSourceProvider.js
```diff
@@ -51,55 +29,33 @@ function parseSourceData(data, extSrc, extDest, layer) {
...
-            }
-        }
-
-        return parsedFile;
-    });
+    return source.parser(data, options).then(parsedFile => source.onParsedFile(parsedFile));
 }
 
-function fetchSourceData(url, layer) {
+function fetchSourceData(extSrc, layer) {
```
zarov, line 35: :+1:

src/Provider/DataSourceProvider.js
```diff
@@ -1,40 +1,18 @@
...
 
 function isValidData(data, extentDestination, validFn) {
     if (data && (!validFn || validFn(data, extentDestination))) {
         return data;
     }
 }
 
-const error = (err, url, source) => {
-    source.handlingError(err, url);
+const error = (err, source) => {
```
zarov, line 9: Why not replace `url` by `extent` ?

src/Provider/DataSourceProvider.js
```diff
@@ -1,40 +1,18 @@
...
+    source.handlingError(err);
```
zarov, line 10: There is a modification to do in `WFSSource` and `test/unit/source.js`

zarov (approved):
Looks good, if you want to modify some things on the `handlingError` please do, otherwise let's move on.

src/Source/FileSource.js
gchoqueux: because the extent `EPSG:4978` doesn't make sense

src/Source/Source.js
gchoqueux: yes

src/Provider/DataSourceProvider.js
gchoqueux: thanks

src/Source/FileSource.js
zarov: Ah yes that's #1063 my bad

PR #1231, opened 2019-09-13
chore: build/transpile itowns to specific targets.

gchoqueux:
## Description
With babel it's possible to transpile itowns to specific targets. ([visiting](https://babeljs.io/docs/en/babel-preset-env)).
I would like target to specific browser using [browserslist](https://github.com/browserslist/browserslist).

## Questions
* For example, the possible targets could be browsers supporting ES6.
* In `webpack.config.js` there is this comment (see bellow), I think it's obsolete,  then we could remove target option  `{ modules: false }`, WDYT?
```
/*
   configuring babel:
   - when babel runs alone (for `test-unit` for instance), we let him deal with
   ES6 modules, because node doesn't support them yet (planned for v10 lts).
   - however, webpack also has ES6 module support and these 2 don't play well
   together. When running webpack (either `build` or `start` script), we prefer
   to rely on webpack loaders (much more powerful and gives more possibilities),
   so let's disable modules for babel here.
*/
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,9 @@
 {
-    "presets": ["@babel/preset-env"],
+    "presets": [["@babel/preset-env",
+        {
+            "targets": "> 5%"
+        }
+    ]],
     "plugins": [
         ["module-resolver",  { "root": ["./src"] } ],
         ["babel-plugin-inline-import", {
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,7 @@
 {
     "presets": [["@babel/preset-env",
         {
-            "targets": "> 5%"
+            "targets": "> 0.5%, last 2 versions, Firefox ESR, not dead"
         }
     ]],
     "plugins": [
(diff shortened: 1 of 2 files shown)


.babelrc
```diff
@@ -1,5 +1,9 @@
 {
-    "presets": ["@babel/preset-env"],
+    "presets": [["@babel/preset-env",
+        {
+            "targets": "> 0.5%, last 2 versions, Firefox ESR, not dead"
```
tbroyer, line 4: This query (which is equivalent to `"defaults"`) includes Opera Mini which, among others things, doesn't support WebGL.

tbroyer (changes requested):
How about putting the browserslist configuration in the package.json or a `.browserslistrc` file, and then (possibly) use the `eslint-plugin-compat` plugin to also detect usage of features that are not supported by the target browsers? (you'd probably have to "annotate" non-client files with `/* eslint-env node */` to disable the plugin where it doesn't make sense, such as the `webpack.config.js`)

Desplandis:
Done in #2296 and #2252.
We can close this PR.

PR #1232, opened 2019-09-23
chore(es6): use ES6 type class for c3DEngine

zarov:
I am doing some things on c3DEngine in another branch and took the opportunity to quickly update the syntax of `c3DEngine`. No changes here apart from this.

### DIFF
--- a/src/Renderer/c3DEngine.js
+++ b/src/Renderer/c3DEngine.js
@@ -7,259 +7,242 @@
 import * as THREE from 'three';
 import Capabilities from 'Core/System/Capabilities';
 import { unpack1K } from 'Renderer/LayeredMaterial';
+import { WEBGL } from 'ThreeExtended/WebGL';
 
-function c3DEngine(rendererOrDiv, options = {}) {
-    const NOIE = !Capabilities.isInternetExplorer();
-    // pick sensible default options
-    if (options.antialias === undefined) {
-        options.antialias = true;
-    }
-    if (options.alpha === undefined) {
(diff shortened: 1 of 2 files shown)


src/Renderer/c3DEngine.js
```diff
@@ -8,258 +8,256 @@ import * as THREE from 'three';
...
-    // The condition below looks weird, but it's correct: querying tabIndex
-    // returns -1 if not set, but we still need to explicitly set it to force
-    // the tabindex focus flag to true (see
-    // https://www.w3.org/TR/html5/editing.html#specially-focusable)
-    if (this.renderer.domElement.tabIndex === -1) {
-        this.renderer.domElement.tabIndex = -1;
-    }
+        if (!this.renderer) {
+            // from THREE/WebGL.js
+            const element = document.createElement('div');
```
gchoqueux: Could you move this html message error in specific file (with message in parameter)?

There's may be a redundancy with console error in line 63.

gchoqueux (commented):
Thanks
It's difficult to see if you have only pass `c3Engine` on Class.

src/Renderer/c3DEngine.js
zarov: I replaced it by calling the `WEBGL.getErrorMessage()` from `three` directly.

zarov:
Updated. You can use the whitespace flags to hide the indentation changes.

gchoqueux (approved):

PR #1235, opened 2019-09-26
fix(3dtile): avoid to clone overrideMaterials to each each assignment.

gchoqueux:
## Description
Avoid to clone overrideMaterials to each each assignment.


zarov (approved):
LGTM

PR #1240, opened 2019-10-07
Fix vector tile picking

zarov:
Fixes some problem encountered with the recent picking feature.


src/Utils/FeaturesUtils.js
```diff
@@ -145,39 +147,37 @@ export default {
     filterFeaturesUnderCoordinate(coordinate, features, epsilon = 0.1) {
         const result = [];
 
+        coord.crs = coordinate.crs;
```
gchoqueux: `Coordinates.copy` should be copy already the crs, no? WDYM?

zarov: Yes, that's something I saw and wanted to do ! Updating now

gchoqueux (approved):

PR #1245, opened 2019-10-10
refactor(Raster): use extent to handle texture instead of index integer.

gchoqueux:
## Description
Use extent to handle texture instead of index integer.

## Motivation and Context
it's more appropriate and intuitive to apply a texture on extent.
In addition, one texture could be applied individually.

### DIFF
--- a/src/Layer/TiledGeometryLayer.js
+++ b/src/Layer/TiledGeometryLayer.js
@@ -10,6 +10,7 @@ import { ImageryLayers } from 'Layer/Layer';
 
 const subdivisionVector = new THREE.Vector3();
 const boundingSphereCenter = new THREE.Vector3();
+const offsetScale = new THREE.Vector4();
 
 /**
  * @property {InfoTiledGeometryLayer} info - Status information of layer
@@ -371,14 +372,16 @@ class TiledGeometryLayer extends GeometryLayer {
         // The induced geometric error is much too large and distorts the SSE
         const nodeLayer = node.material.getElevationLayer();
         if (nodeLayer) {
-            const currentTexture = nodeLayer.firstTexture();
-            if (currentTexture && currentTexture.extent) {
-                const offsetScale = nodeLayer.firstOffsetScale();
-                const ratio = offsetScale.z;
-                // ratio is node size / texture size
-                if (ratio < 1 / Math.pow(2, this.maxDeltaElevationLevel)) {
-                    return false;
+            for (const [extent, texture] of nodeLayer.textures.entries()) {
+                if (texture) {
+                    extent.offsetToParent(texture.coords, offsetScale);
+                    const ratio = offsetScale.z;
+                    // ratio is node size / texture size
+                    if (ratio < 1 / Math.pow(2, this.maxDeltaElevationLevel)) {
+                        return false;
+                    }
                 }
+                break;
             }
         }
 
--- a/src/Layer/TiledGeometryLayer.js
+++ b/src/Layer/TiledGeometryLayer.js
@@ -371,9 +371,9 @@ class TiledGeometryLayer extends GeometryLayer {
         // The induced geometric error is much too large and distorts the SSE
         const nodeLayer = node.material.getElevationLayer();
         if (nodeLayer) {
-            const currentTexture = nodeLayer.textures[0];
+            const currentTexture = nodeLayer.firstTexture();
             if (currentTexture && currentTexture.extent) {
-                const offsetScale = nodeLayer.offsetScales[0];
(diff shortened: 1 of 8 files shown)


zarov:
I got some bad display when getting closer to the ground:

Screenshot from 2019-10-11 11-02-58

While I should get:

Screenshot from 2019-10-11 11-03-15

src/Renderer/MaterialLayer.js
```diff
@@ -82,8 +82,9 @@ class MaterialLayer {
         defineLayerProperty(this, 'zmin', layer.zmin, defaultEle.zmin);
         defineLayerProperty(this, 'zmax', layer.zmax, defaultEle.zmax);
 
-        this.textures = [];
-        this.offsetScales = [];
+        this.textures = new Map();
+        extents.forEach(ex => this.textures.set(ex, null));
+        this.offsetScales = new Map();
```
zarov: You don't set `this.offsetScales` ?

src/Renderer/MaterialLayer.js
```diff
@@ -126,22 +130,30 @@ class MaterialLayer {
             }
         }
         this.level = EMPTY_TEXTURE_ZOOM;
-        this.textures = [];
-        this.offsetScales = [];
+        this.textures.forEach((v, k) => this.textures.set(k, null));
```
zarov, line 132: No `this.textures.clear()` instead ?

src/Renderer/MaterialLayer.js
```diff
@@ -126,22 +130,30 @@ class MaterialLayer {
...
         this.material.layersNeedUpdate = true;
     }
 
-    setTexture(index, texture, offsetScale) {
-        this.level = (texture && (index == 0)) ? texture.coords.zoom : this.level;
-        this.textures[index] = texture || null;
-        this.offsetScales[index] = offsetScale;
+    setTexture(texture, extent) {
+        this.level = texture ? texture.coords.zoom : this.level;
+        this.textures.set(extent, texture || null);
```
zarov, line 138: I'm not a fan of using `extent` as a key, maybe it's me but using such a "complex" key when getting something in a `Map` seems to be slow and overkill. Why not write a method that generate a unique key from the content of the extent ?

However, I see two drawbacks in this suggestion (that needs to be verified):
- the conversion to a string can mean a lot of process if it is called a lot
- I'm not sure if the key is used as it is sometimes (it seems in `updateLayeredMaterialNodeImagery`, so...)

zarov (commented):
Looking good, can you do a quick benchmark to see the impact on performance ? I'm quite sceptical about this, sorry. I'm afraid it will have too much of an impact.

src/Renderer/MaterialLayer.js
gchoqueux: No, the previous `offsetScales` wasn't initializing.

src/Renderer/MaterialLayer.js
gchoqueux: I would like keep this `Extents` to assign the texture by `Extent`

src/Renderer/MaterialLayer.js
gchoqueux: Me too, I'm not fan. We can optimize later.

src/Renderer/MaterialLayer.js
gchoqueux: I thinks remove this property.

src/Renderer/MaterialLayer.js
zarov: Good to me

src/Renderer/MaterialLayer.js
gchoqueux: done

src/Layer/TiledGeometryLayer.js
```diff
@@ -371,14 +372,16 @@ class TiledGeometryLayer extends GeometryLayer {
...
-            const currentTexture = nodeLayer.textures[0];
-            if (currentTexture && currentTexture.extent) {
-                const offsetScale = nodeLayer.offsetScales[0];
-                const ratio = offsetScale.z;
-                // ratio is node size / texture size
-                if (ratio < 1 / Math.pow(2, this.maxDeltaElevationLevel)) {
-                    return false;
+            for (const [extent, texture] of nodeLayer.textures.entries()) {
+                if (texture) {
+                    extent.offsetToParent(texture.coords, offsetScale);
```
zarov, line 377: You can avoid calling `offsetToParent` by maybe adding a new method on `Extent`, something that only compute the `z` part, as it is the only thing used here.

Desplandis:
@gchoqueux Are you still working on it?

PR #1250, opened 2019-10-30
Best support vector tiles

gchoqueux:
## Description
Add supports to
 * layers ordering
 * pattern
 * fill stroke
 * line cap and join
 * dash line
 * multi styles
 * styles by zoom level

and some fixes.

vt

### DIFF
--- a/examples/js/plugins/FeatureToolTip.js
+++ b/examples/js/plugins/FeatureToolTip.js
@@ -75,7 +75,7 @@ var FeatureToolTip = (function _() {
             feature = features[p];
 
             geometry = feature.geometry;
-            style = (layer.style && layer.style.isStyle) ? layer.style : geometry.properties.style;
+            style = (layer.style && layer.style.isStyle) ? layer.style : (geometry._feature.style || geometry.properties.style);
             fill = style.fill.color;
             stroke = '1.25px ' + style.stroke.color;
 
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -274,15 +274,8 @@ export class FeatureCollection {
         this.updateExtent(feature.extent);
     }
 
-    /**
-     * Returns the Feature by type if `mergeFeatures` is `true` or returns the
-     * new instance of typed Feature.
-     *
-     * @param {string} type the type requested
-     * @returns {Feature}
-     */
-    getFeatureByType(type) {
-        const feature = this.features.find(feature => feature.type === type);
+    requestFeature(type, callback) {
+        const feature = this.features.find(callback);
         if (feature && this.optionsFeature.mergeFeatures) {
             return feature;
         } else {
@@ -290,5 +283,46 @@ export class FeatureCollection {
             this.features.push(newFeature);
             return newFeature;
         }
+    }
+
+    /**
+     * Returns the Feature by type if `mergeFeatures` is `true` or returns the
+     * new instance of typed Feature.
+     *
+     * @param {string} type the type requested
+     * @returns {Feature}
+     */
+    requestFeatureByType(type) {
+        return this.requestFeature(type, feature => feature.type === type);
+    }
+
+    /**
+     * Returns the Feature by type if `mergeFeatures` is `true` or returns the
+     * new instance of typed Feature.
+     *
+     * @param {string} id the id requested
+     * @param {string} type the type requested
+     * @returns {Feature}
+     */
+    requestFeatureById(id, type) {
+        return this.requestFeature(type, feature => feature.id === id);
+    }
+    /**
+     * Add a new feature with references to all properties.
+     * It allows to have features with different styles
+     * without having to duplicate the geometry.
+     * @param      {Feature}   feature  The feature to reference.
+     * @return     {Feature}  The new referenced feature
+     */
+    newFeatureByReference(feature) {
+        const ref = new Feature(feature.type, this.crs, this.optionsFeature);
+        ref.extent = feature.extent;
+        ref.geometry = feature.geometry;
+        ref.normals = feature.normals;
+        ref.size = feature.size;
+        ref.vertices = feature.vertices;
+        ref._pos = feature._pos;
+        this.features.push(ref);
+        return ref;
     }
 }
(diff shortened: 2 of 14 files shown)


src/Source/Source.js
```diff
@@ -116,6 +116,10 @@ class Source {
         }
     }
 
+    isVectorSource() {
+        return supportedParsers.get(this.format) !== undefined;
```
zarov: This could be a value set in the constructor, as it doesn't seem to change.

src/Core/Feature.js
```diff
@@ -291,4 +291,26 @@ export class FeatureCollection {
             return newFeature;
         }
     }
+
+    getFeatureById(id, type) {
+        const feature = this.features.find(feature => feature.id === id);
+        if (feature && this.optionsFeature.mergeFeatures) {
```
zarov: If I understand correctly, simply having `feature` is not sufficient and will return a new one ? Why ?

src/Core/Feature.js
```diff
@@ -291,4 +291,26 @@ export class FeatureCollection {
...
+            return feature;
+        } else {
+            const newFeature = new Feature(type, this.crs, this.optionsFeature);
+            this.features.push(newFeature);
+            return newFeature;
+        }
+    }
+
+    getReference(feature) {
```
zarov: Could you add a little comment/documentation on what is this method supposed to return ? It seems that it is a `clone`, so maybe do a `Feature#clone` method instead ?

src/Core/Style.js
```diff
@@ -88,35 +101,58 @@ class Style {
...
+            const { color, opacity } = rgba2rgb(readVectorProperty(layer.paint['fill-color'] || layer.paint['fill-pattern']));
             this.fill.color = color;
-            this.fill.opacity = layer.paint['fill-opacity'] || opacity || 1.0;
+            this.fill.opacity = readVectorProperty(layer.paint['fill-opacity'], zoom) || opacity || 1.0;
+            if (layer.paint['fill-pattern']) {
+                const sprite = sprites[layer.paint['fill-pattern']];
+                const c = document.createElement('canvas');
+                c.width = sprite.width;
+                c.height = sprite.height;
+                const ctx = c.getContext('2d');
```
zarov: Could you instance a canvas globally, so you don't have to get the context everytime ?

src/Parser/VectorTileParser.js
```diff
@@ -138,25 +139,44 @@ function readPBF(file, options) {
...
-            const layerStyle = layersStyle.filter(l => l.filterExpression({ zoom: extentSource.zoom }, vtFeature))[0];
-            if (layerStyle) {
-                const properties = vtFeature.properties;
-                properties.style = styleCache.get(layerStyle.id);
-                if (!properties.style) {
-                    properties.style = new Style();
-                    properties.style.setFromVectorTileLayer(layerStyle);
-                    styleCache.set(layerStyle.id, properties.style);
+            const layerStyles = layersStyle.filter(l => l.filterExpression({ zoom: z }, vtFeature));
+            const layerStyle = layerStyles.filter(l => (!l.minzoom || l.minzoom <= z) && (!l.maxzoom || l.maxzoom >= z));
```
zarov: Why not apply all conditions to the filter the first time ?

zarov (changes requested):
Nice to have those improvements !

src/Core/Feature.js

src/Source/Source.js
gchoqueux: yes

src/Core/Feature.js
gchoqueux: For rendering, Vector tiles layers need to be ordered. If layers are grouped by type, it's more complicated to order. So, for vector tiles I add a feature/reference by layer.

src/Core/Feature.js
gchoqueux: I added a documentation

src/Core/Style.js
gchoqueux: I'll try

src/Parser/VectorTileParser.js
gchoqueux: yes

gchoqueux:
@zarov PR updated

src/Source/Source.js
```diff
@@ -106,6 +106,7 @@ class Source {
         this.format = source.format;
         this.fetcher = source.fetcher || supportedFetchers.get(source.format) || Fetcher.texture;
         this.parser = source.parser || supportedParsers.get(source.format) || (d => Promise.resolve(d));
+        this.isVectorSource = source.parser || supportedParsers.get(source.format);
```
zarov: This isn't a boolean. It might works for some case, but it's better to set a boolean.

src/Core/Feature.js
```diff
@@ -273,6 +273,16 @@ export class FeatureCollection {
         this.features.push(feature);
         this.updateExtent(feature.extent);
     }
+    getFeature(type, callback) {
+        const feature = this.features.find(callback);
+        if (feature && this.optionsFeature.mergeFeatures) {
```
zarov, line 279: It should return the `feature` if it is found, if `mergeFeatures` is not defined, so you get something like this:
```js
if (feature && (this.optionsFeature.mergeFeatures == undefined || this.optionsFeature.mergeFeatures)) {
```
or else you should rename the function as it does not really get the Feature but only the Feature if the Features are merged.

examples/js/plugins/FeatureToolTip.js
```diff
@@ -75,7 +75,7 @@ var FeatureToolTip = (function _() {
             feature = features[p];
 
             geometry = feature.geometry;
-            style = (layer.style && layer.style.isStyle) ? layer.style : geometry.properties.style;
+            style = (layer.style && layer.style.isStyle) ? layer.style : (feature.geometry._feature.style || geometry.properties.style);
```
zarov: Can be `geometry._feature.style`

zarov (approved):
LGTM, thanks for these improvements !

PR #1258, opened 2019-11-21
Update proj4 geocent

zarov:
By updating the proj4 version to 2.6.0, we get the support for
geocentric projection, like EPSG:4978. Now the complicated behavior in
Coordinates#as can be ditched.


zarov:
This PR has been rebased and is ready to be merged !

gchoqueux (approved):
good cleaning of `Coordinate.as()`

PR #1264, opened 2019-11-28
refacto(cache): ditch Map() in Cache for simple tree

zarov:
The previous Cache was working on a single Map, which can become really
slow and big after a few minutes of navigation in iTowns.

Now, the Cache is composed of a single tree, that can contain node up
to 3 levels. It simplifies things attached to a layer, a source or
others things like a zoom value. Instead of having a complex string
composed with all those information, it now uses several key accross all
the node of the Cache.

In term of performance, the proposed implementation is 20 times faster
than the previous one.

### DIFF
--- a/src/Core/Scheduler/Cache.js
+++ b/src/Core/Scheduler/Cache.js
@@ -1,5 +1,5 @@
-const data = new Map();
-const stats = new Map();
+let data = {};
+let entry;
 
 /**
  * This is a copy of the Map object, except that it also store a value for last
@@ -13,14 +13,16 @@ const stats = new Map();
  * @example
  * import Cache from './Cache';
  *
- * Cache.set('foo', { bar: 1 }, Cache.POLICIES.TEXTURE);
- * Cache.set('acme', { bar: 32 });
+ * Cache.set({ bar: 1 }, Cache.POLICIES.TEXTURE, 'foo');
+ * Cache.set({ bar: 32 }, Cache.POLICIES.INFINITE, 'foo', 'toto');
  *
  * Cache.get('foo');
  *
  * Cache.delete('foo');
  *
  * Cache.clear();
+ *
+ * Cache.flush();
  */
 const Cache = {
     /**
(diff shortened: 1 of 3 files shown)


src/Core/Scheduler/Cache.js
```diff
@@ -49,44 +51,68 @@ const Cache = {
...
+     * @param {number} [lifetime=Infinity]
+     * @param {string} key1
+     * @param {string} [key2]
+     * @param {string} [key3]
      *
      * @return {Object} the added value
      */
-    set: (key, value, lifetime = Infinity) => {
-        const entry = {
+    set: (value, lifetime = Infinity, key1, key2, key3) => {
```
gchoqueux: could you remove default value for `lifetime`?

src/Core/Scheduler/Cache.js
```diff
@@ -49,44 +51,68 @@ const Cache = {
...
+        if (key3 == undefined) {
+            data[key1][key2] = entry;
+            return value;
+        }
+
+        if (!data[key1][key2]) {
+            data[key1][key2] = {};
+        }
+
+        data[key1][key2][key3] = entry;
```
gchoqueux, line 120: could you improve this part. May be in using `get` function?

src/Core/Scheduler/Cache.js
```diff
@@ -97,19 +123,31 @@ const Cache = {
...
+        if (data[key1] == undefined) {
+            throw Error('Please specify at least a key of something to delete');
+        } else if (data[key1][key2] == undefined) {
+            delete data[key1];
+        } else if (data[key1][key2][key3] == undefined) {
+            delete data[key1][key2];
+        } else {
+            delete data[key1][key2][key3];
+        }
+    },
```
gchoqueux, line 145: Destroying only the `key1` could be useful to clean the cache of one `source/Layer`?

src/Core/Scheduler/Cache.js
```diff
@@ -121,32 +159,26 @@ const Cache = {
      * @name module:Cache.flush
      * @function
      *
-     * @param {number} [time]
-     *
-     * @return {Object} Statistics about the flush: `before` gives the number of
```
gchoqueux, line 126: `Statistics` isn't used in `Debug`?

src/Core/Scheduler/Cache.js
zarov: Done

src/Core/Scheduler/Cache.js
zarov: It is the case: if `key2 == undefined`, you are in the case `data[key1][key2] == undefined`, as `data[key1][undefined] == undefined`. Deleting then `data[key1]` deletes all the children of `delete[key1]`.

src/Core/Scheduler/Cache.js
zarov: No, that's why I removed it.

src/Core/Scheduler/Cache.js
zarov: Not really, I still need to create nodes if they are not present.

gchoqueux: Ok for now, we can simplify later.

gchoqueux (approved):

PR #1265, opened 2019-12-03
feat(tile): add a refreshMaterial method

zarov:
In a case where you have a server, with the data updated from what is
displayed, you want to update it without destroying the whole layer and
recreating it. If you know the tile you want to update, you can use the
new `TileMesh#refreshMaterial` method.

In the same time, a new `FORCE` state has been added to
`LayerUpdateState`, that is used when doing this refresh.

I also wanted to refactor a little bit `LayerUpdateState`, but I'm leaving this
for a later time.

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -1,5 +1,6 @@
 import * as THREE from 'three';
 import CRS from 'Core/Geographic/Crs';
+import Cache from 'Core/Scheduler/Cache';
 
 /**
  * A TileMesh is a THREE.Mesh with a geometricError and an OBB
@@ -99,6 +100,39 @@ class TileMesh extends THREE.Mesh {
     onBeforeRender() {
         if (this.material.layersNeedUpdate) {
             this.material.updateLayersUniforms();
+        }
+    }
+
+    /**
+     * Refresh a material of this tile, associated with a specific layer. Only
+     * works with ColorLayer and ElevationLayer for now. It also can't work with
+     * a layer containing images like `png` or `jpeg`, as there is no way to
+     * specify the necessary flag to retrieve images again from a server. Once
+     * this method called, you need to call `view.notifyChange(node)` if you
+     * want the update to take place.
+     *
+     * @param {ColorLayer|ElevationLayer} layer - The layer associated to the
+     * material.
+     * @param {boolean} [traverse=false] - Refresh the children of the tile - if
+     * any.
+     */
+    refreshMaterial(layer, traverse = false) {
+        if (layer.isColorLayer || layer.isElevationLayer) {
+            const materialLayer = this.material.getLayer(layer.id);
+            if (materialLayer) {
+                materialLayer.textures.forEach(t => Cache.delete(layer.source.uid, layer.id, t.extent.toString('-')));
+            }
+
+            this.material.removeLayer(layer.id);
+            this.layerUpdateState[layer.id].forceUpdate();
+
+            if (traverse) {
+                this.traverse((c) => {
+                    if (c.isTileMesh) {
+                        c.refreshMaterial(layer);
+                    }
+                });
+            }
         }
     }
 }
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -154,7 +154,7 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
     }
 
     node.layerUpdateState[layer.id].newTry();
-    const parsedData = nodeLayer.textures.map(t => t.parsedData);
+    const parsedData = node.layerUpdateState[layer.id].isForcingUpdate() ? nodeLayer.textures.map(t => t.parsedData) : [];
     const command = buildCommand(context.view, layer, extentsSource, extentsDestination, node, forceUpdate, parsedData);
 
     return context.scheduler.execute(command).then(
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -29,7 +29,7 @@ function refinementCommandCancellationFn(cmd) {
     return !cmd.requester.material.visible;
 }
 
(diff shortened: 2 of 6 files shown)


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -90,6 +90,14 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
...
     }
 
+    if (node.layerUpdateState[layer.id].isForcingUpdate()) {
+        // Create new MaterialLayer
+        nodeLayer = material.addLayer(layer);
+        // Init the new MaterialLayer by parent
+        const parentLayer = parent.material && parent.material.getLayer(layer.id);
+        nodeLayer.initFromParent(parentLayer, extentsDestination);
+    }
+
```
gchoqueux: Could you merge code with the `Initialisation`?
There is no forcing in elevation process?

src/Core/TileMesh.js
```diff
@@ -101,6 +102,26 @@ class TileMesh extends THREE.Mesh {
...
+     * Refresh a material of this tile, associated with a specific layer. Only
+     * works with ColorLayer for now.
+     *
+     * @param {ColorLayer} layer - The layer associated to the material.
+     * @param {View} [view] - The view to update, don't forget to call
+     * `view.notifyChange` otherwise.
+     */
+    refreshMaterial(layer, view) {
+        if (layer.isColorLayer) {
+            this.material.getLayer(layer.id).textures.forEach(t => Cache.delete(`${layer.source.uid}-${t.extent.toString('-')}`));
```
gchoqueux: The key cache could be factorized with the one done in the `dataSourceProvider`, because it's better for the maintainability of the code

zarov: Yes, but if we merged #1264 there is no point to this, so let's wait and rebase on #1264 when it's done.

src/Process/LayeredMaterialNodeProcessing.js
zarov: Added a `createMaterialLayer` function, and add the support for an `ElevationLayer`.

zarov:
I updated the refresh mechanism with a temporary change of the `cache` flag in `DataSourceProvider`.

src/Core/TileMesh.js
```diff
@@ -41,6 +42,7 @@ class TileMesh extends THREE.Mesh {
 
         this.layerUpdateState = {};
         this.isTileMesh = true;
+        this.isRefreshing = false;
```
gchoqueux: I prefer use only layer update state for refreshing statement. You could pass force statement in the command.

src/Core/TileMesh.js
```diff
@@ -101,6 +103,43 @@ class TileMesh extends THREE.Mesh {
...
+        this.isRefreshing = true;
+
+        if (layer.isColorLayer || layer.isElevationLayer) {
+            const materialLayer = this.material.getLayer(layer.id);
+            if (materialLayer) {
+                materialLayer.textures.forEach(t => Cache.delete(layer.source.uid, layer.id, t.extent.toString('-')));
+            }
+
+            this.material.removeLayer(layer.id);
+            this.layerUpdateState[layer.id].forceUpdate();
```
gchoqueux, line 127: May be reset `layerUpdateState` to return to the initial state then there may be no need to modify the processing.

src/Core/TileMesh.js
```diff
@@ -101,6 +103,43 @@ class TileMesh extends THREE.Mesh {
...
+
+            if (view) {
+                view.notifyChange(this);
```
gchoqueux: I don't think this is the best place to refresh the view.

src/Renderer/MaterialLayer.js
```diff
@@ -139,7 +139,6 @@ class MaterialLayer {
     }
 
     setTextures(textures, pitchs) {
-        this.dispose(false);
```
gchoqueux, line 142: Why remove the dispose?

gchoqueux (changes requested):

src/Renderer/MaterialLayer.js
zarov: Some error, I'll change it back.

src/Core/TileMesh.js
zarov: Done

src/Core/TileMesh.js
zarov: Can't do, if I change it, I can't tell if it is refreshing in `LayeredMaterialNodeProcessing`, unless we keep `isRefreshing`, which isn't better indeed.

src/Core/TileMesh.js
zarov: Moved to a better place

zarov:
Updated and rebased

src/Core/TileMesh.js
```diff
@@ -101,6 +102,43 @@ class TileMesh extends THREE.Mesh {
...
+            if (traverse) {
+                this.traverse((c) => {
+                    if (c.isTileMesh) {
+                        c.refreshMaterial(layer);
+                    }
+                });
+            }
+        }
+
+        if (view) {
```
gchoqueux: could move `notifyChange` from this file?

src/Provider/DataSourceProvider.js
```diff
@@ -69,7 +76,7 @@ export default {
                 const extDest = extentsDestination[i];
 
                 // Already fetched and parsed data that can be used
-                const validedParsedData = isValidData(parsedData[i], extDest, layer.isValidData) || source.parsedData;
+                const validedParsedData = !command.forceUpdate && (isValidData(parsedData[i], extDest, layer.isValidData) || source.parsedData);
```
gchoqueux: Maybe there's a duplicate in the verification, because `parsedData` could be deleted in `refreshMaterial`.

src/Core/TileMesh.js
zarov: Removed it and specified it in the documentation that it needs to be called.

Desplandis:
Obsolete.

PR #1266, opened 2019-12-03
refactor(controls): simplify animation and globeControls damping.

gchoqueux:
## Description
simplify `Animation` and `globeControls` damping.


zarov (approved):
LGTM, thanks for the simplification !

PR #1270, opened 2019-12-04
refactor(controls): use Class for GlobeControls.

gchoqueux:
## Description
Use Class for `GlobeControls`.

Waiting #1266

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -130,7 +130,27 @@ export const CONTROL_EVENTS = {
     CAMERA_TARGET_CHANGED: 'camera-target-changed',
 };
 
+const quaterPano = new THREE.Quaternion();
+const quaterAxis = new THREE.Quaternion();
+const axisX = new THREE.Vector3(1, 0, 0);
+let minDistanceZ = Infinity;
+const lastNormalizedIntersection = new THREE.Vector3();
+const normalizedIntersection = new THREE.Vector3();
+const raycaster = new THREE.Raycaster();
+const targetPosition = new THREE.Vector3();
+const pickedPosition = new THREE.Vector3();
+
 let previous;
+
+function getRangeFromScale(scale, pitch, fov, height) {
+    // Screen pitch, in millimeters
+    pitch = (pitch || 0.28) / 1000;
+    fov = THREE.Math.degToRad(fov);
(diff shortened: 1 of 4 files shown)


src/Controls/GlobeControls.js
```diff
@@ -131,8 +130,28 @@ export const CONTROL_EVENTS = {
...
+function getRangeFromScale(scale, pitch, fov, height) {
+    // Screen pitch, in millimeters
+    pitch = (pitch || 0.28) / 1000;
+    fov = THREE.Math.degToRad(fov);
+    // Invert one unit projection (see getDollyScale)
+    const range = pitch * height / (scale * 2 * Math.tan(fov * 0.5));
+
+    return range;
+}
+
```
zarov, line 154: Maybe you could replace this by `View#getScaleFromDistance`, or at least use it.

src/Controls/GlobeControls.js
```diff
@@ -156,104 +175,172 @@ let previous;
...
+        this.player = new AnimationPlayer();
+        this._view = view;
+        this.camera = view.camera.camera3D;
+        this.domElement = view.mainLoop.gfxEngine.renderer.domElement;
+
+        // State control
+        this.states = new StateControl();
+        this.state = this.states.NONE;
+
+        this.getStates = () => this.states;
```
zarov, line 190: Is it necessary ?

src/Controls/GlobeControls.js
zarov: It seems you can also replace things like `getRange` by `View#getDistanceFromCamera`. There seems to be quite a lot of method that are giving the same result while doing different things, so maybe take a look into that if you have time (between `View`, `CameraUtils` and `GlobeControls` for instance).

zarov (commented):
Nothing much to say, thanks for continuing this cleaning ! Waiting for the test to approve.

src/Controls/GlobeControls.js
gchoqueux: no, I will remove in next clean PR.

src/Controls/GlobeControls.js
gchoqueux: Yes, I'll create a new PR to re-factorize and I only use the class in this PR.

zarov (approved):

PR #1271, opened 2019-12-06
chore: update packages.

gchoqueux:
LegacyGLTFLoader was removed from THREE v111.0.0.


zarov (approved):
You can add `LegacyGLTFLoader` to the list of exclusion of `nyc` in `package.json`, as we won't do any maintenance on it, and remove it as soon 3dtiles examples are in GLTF2.0.

PR #1272, opened 2019-12-11
chore: add proxy argument to functional tests.

gchoqueux:
## Description
add proxy argument to functional tests.

### DIFF
--- a/test/functional/bootstrap.js
+++ b/test/functional/bootstrap.js
@@ -182,6 +182,10 @@ before(async () => {
     // using --no-sandbox.
     const args = [];
 
+    if (process.env.HTTPS_PROXY) {
+        args.push(`--proxy-server=${process.env.HTTPS_PROXY}`);
+    }
+
     if (process.env.REMOTE_DEBUGGING) {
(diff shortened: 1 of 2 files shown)


test/functional/bootstrap.js
```diff
@@ -182,6 +182,10 @@ before(async () => {
     // using --no-sandbox.
     const args = [];
 
+    if (process.env.HTTPS_PROXY) {
+        args.push(`--proxy-server=${process.env.HTTPS_PROXY}`);
+    }
```
zarov, line 187: This breaks everything for me now. What I propose is this instead:

```js
const isProxy = process.argv.pop().match('--proxy');
if (proxy && process.env.HTTPS_PROXY) {
    args.push(`--proxy-server=${process.env.HTTPS_PROXY}`);
}
```

And now you can use `npm run test-functional -- --proxy` if you need a proxy.

zarov: nevermind, I got a bad `HTTPS_PROXY` variable.

zarov (approved):

PR #1273, opened 2019-12-12
Feature position camera with View constructor

gchoqueux:
## Description
* Feature position camera with View constructor 
* Clean `GLobeControls`

**BREAKING CHANGE**
View constructor's signature has changed.

### DIFF
--- a/examples/source_stream_wfs_3d.html
+++ b/examples/source_stream_wfs_3d.html
@@ -27,6 +27,11 @@
 
             // Define initial camera position
             var positionOnGlobe = { longitude: 4.818, latitude: 45.7354, altitude: 3000 };
+            var placement = {
+                coord: new itowns.Coordinates('EPSG:4326', 4.818, 45.7354),
+                range: 3000,
+                tilt: 45,
+            }
             var meshes = [];
             var linesBus = [];
             var scaler;
@@ -35,7 +40,7 @@
             var viewerDiv = document.getElementById('viewerDiv');
 
             // Instanciate iTowns GlobeView*
-            var view = new itowns.GlobeView(viewerDiv, positionOnGlobe);
+            var view = new itowns.GlobeView(viewerDiv, placement);
             setupLoadingScreen(viewerDiv, view);
 
             // Add one imagery layer to the scene
@@ -197,7 +202,6 @@
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 // eslint-disable-next-line no-console
                 console.info('Globe initialized');
-                view.controls.setTilt(45);
             });
 
             debug.createTileDebugUI(menuGlobe.gui, view);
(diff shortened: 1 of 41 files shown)


nosy-b:
Hi guys,
Is their a reason to stop propagation on the mouseWheel event? It creates some difficulties when using the mouswheel for specific tasks.

src/Core/Prefab/GlobeView.js
```diff
@@ -81,28 +81,15 @@ class GlobeView extends View {
...
      * @param {HTMLDivElement} viewerDiv - Where to attach the view and display it
      * in the DOM.
-     * @param {object|Coordinates} coordCarto - An object containing three
-     * properties: longitude, latitude and altitude. It will help placing the camera
-     * on the globe at the creation.
+     * @param {CameraTransformOptions} placement - An object to place view
      * @param {object=} options - See options of {@link View}.
      */
-    constructor(viewerDiv, coordCarto, options = {}) {
+    constructor(viewerDiv, placement, options = {}) {
```
zarov: You need to have `placement = {}` or it will break if it is `undefined` and `options.noControls = true`.

examples/source_stream_wfs_3d.html
```diff
@@ -197,7 +202,6 @@
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 // eslint-disable-next-line no-console
                 console.info('Globe initialized');
-                view.controls.setTilt(45);
             });
 
             debug.createTileDebugUI(menuGlobe.gui, view);
```
zarov, line 207: You need to do something for the `view.controls.isPaused()` below.

zarov (commented):
It is way better, thanks ! A few things to do and it's good to go !

zarov:
@nosy-b it was already the case before it seems, what is your usecase ?

nosy-b:
@zarov yes it was the case before but I was wondering as you work on GlobeControl if we could remove  it if there is no use. 
My use case is using the mousewheel in an app and after globecontrol is loaded I can't have control on it.

gchoqueux:
@nosy-b I removed the `stopProgation`

src/Core/Prefab/GlobeView.js
gchoqueux: done

examples/source_stream_wfs_3d.html
gchoqueux: done

zarov (approved):
LGTM thanks

examples/source_stream_wfs_3d.html
```diff
@@ -197,13 +202,12 @@
...
 
             function picking(event) {
-                if(view.controls.isPaused()) {
```
zarov, line 206: You deleted an extra parenthesis

PR #1290, opened 2020-01-13
refactor: remove the Feature reference from FeatureGeometry

zarov:
Even if it was a simple pointer, scenes with a lot of FeatureGeometry
were having a lot of memory taken because of those pointers.

### DIFF
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -31,25 +31,26 @@ export class FeatureGeometry {
     constructor(feature) {
         this.extent = feature.extent ? defaultExtent(feature.crs) : undefined;
         this.indices = [];
-        this._feature = feature;
         this.properties = {};
         this._currentExtent = feature.extent ? defaultExtent(feature.crs) : undefined;
+        this.size = feature.size;
     }
     /**
      * Add a new marker to indicate the starting of sub geometry and extends the vertices buffer.
      * Then you have to push new the coordinates of sub geometry.
      * The sub geometry stored in indices, see constructor for more information.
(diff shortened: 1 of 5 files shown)


src/Core/Feature.js
```diff
@@ -77,49 +79,44 @@ export class FeatureGeometry {
...
+            coord.as(feature.crs, coord);
         }
-        if (this._feature.normals) {
-            coord.geodesicNormal.toArray(this._feature.normals, this._feature._pos);
+        if (feature.normals) {
+            coord.geodesicNormal.toArray(feature.normals, feature._pos);
         }
 
-        this._feature._pushValues(this._feature, coord.x, coord.y, coord.z);
+        feature._pushValues(feature, coord.x, coord.y, coord.z);
```
gchoqueux: it's possible to remove `feature` from `_pushValues`  params?

gchoqueux (commented):
add properties like: vertices, normal, crs 
and remove feature from function params.

src/Core/Feature.js
zarov: Done

zarov:
> add properties like: vertices, normal, crs and remove feature from function params.

I'm not convinced by this, because in some case that would be a lot of parameters. [See this function](https://github.com/iTowns/itowns/blob/master/src/Core/Feature.js#L44-L52): we need the `crs`, `vertices` and then `normals` and `size` in the `_extendBuffer` call. WDYT ?

gchoqueux:
see #1295

gchoqueux (approved):

PR #1298, opened 2020-01-15
Clean oriented image layer

### LINKED ISSUE #990: Design of DataSourceProvider (the requester need a material)
## Description

In the DataSourceProvider :

Now, if there is a requester, and if it doesn't have a material, it throw an error.
Here is the code :

https://github.com/iTowns/itowns/blob/daba997cea841a282a82403d3ae9f941ac5bb3a8/src/Provider/DataSourceProvider.js#L76-L80

## Context

To use the DataSourceProvider in #939 , I must add a fake material to avoid the error, and it's not very clean.

gchoqueux:
refactor(orientedImage): remove useless material property from oriented image feature.

fix #990


zarov (approved):
LGTM, but do you know why this fix was added ? I didn't find anything about this apart from the comment.

PR #1299, opened 2020-01-17
WIP fix(controls): jittering move globe with chrome devtools is open.

### LINKED ISSUE #1277: Bug on Globe Control when Console opened
Globe control is buggy when console is opened (Chrome linux & win tested)

When the camera is far (we see the whole globe) and try to turn it it 's buggy. It only happens when the Console is opened

Tested on Chrome on Linux and Win

gchoqueux:
## Description
Try to fix.
When devtools is opened there are 2 raycasting (2 successive mouse move events) before the update camera, so the camera matrix world needs to be update before to cast a new ray.
When devtools is closed there are 1 raycasting before the update camera, so the camera matrix world is update by `mainloop.step`.

This fix isn't the best solution. May be a best synchronizing with the frame requester is a good way.

fix #1277

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -559,6 +559,7 @@ class GlobeControls extends THREE.EventDispatcher {
                 break;
             case this.states.MOVE_GLOBE: {
                 const normalized = this.view.viewToNormalizedCoords(coords);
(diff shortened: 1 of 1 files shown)


gchoqueux:
fixed in #1704

PR #1303, opened 2020-01-21
feat(renderer): adding a first implementation of Labels

zarov:
This is a first PR in managing labels (either text, icon or both) in iTowns. There is still a lot of room for improvement, but I would like to see this first bit going into the codebase before it is too big to review. Below is the result for the published `vector_tile_raster_2d` example.

image

Here is the result for another unpublished example using vector tile as well:

image

#### feat(renderer): add a renderer for Labels

This new renderer (Label2DRenderer) is launched in c3DEngine, alongside
WebGLRenderer. It manages Labels present on the scene.

#### feat(label): add a first implementation of Labels

This commit has two main aspects:
- `Label` object, which represents a text and/or an icon. It has a 3D
position, and it is displayed through DOM.
- `LabelLayer`, which is managing the present labels on a scene. It can
be created on its own, or it can get data from another layer.

There has also been some update in the `vector_tile_raster_2d` example,
as well as in the parsing of vector tiles and style.

### DIFF
--- a/examples/source_file_gpx_raster.html
+++ b/examples/source_file_gpx_raster.html
@@ -22,7 +22,9 @@
             // Define initial camera position
             var placement = {
                 coord: new itowns.Coordinates('EPSG:4326', 0.089, 42.8989),
-                range: 80000,
+                range: 30000,
+                tilt: 20,
+                heading: 70,
             }
 
             // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
@@ -33,7 +35,6 @@
             var menuGlobe = new GuiTools('menuDiv', view);
 
             setupLoadingScreen(viewerDiv, view);
-            FeatureToolTip.init(viewerDiv, view);
 
             // Add one imagery layer to the scene
             // This layer is defined in a json file but it could be defined as a plain js
@@ -59,6 +60,28 @@
             // Parse and Convert by iTowns
             itowns.Fetcher.xml('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/ULTRA2009.gpx')
                 .then(function _(gpx) {
+                    var gpxStyle = new itowns.Style({
+                        zoom: {
+                            min: 9,
+                        },
+                        stroke: {
+                            color: 'red',
+                        },
+                        point: {
+                            color: 'white',
+                            line: 'red',
+                        },
+                        text: {
+                            field: '{name}',
+                            transform: 'uppercase',
+                            font: ['Arial', 'sans-serif'],
+                            halo: {
+                                color: 'white',
+                                width: 1,
+                            },
+                        },
+                    });
+
                     var gpxSource = new itowns.FileSource({
                         fetchedData: gpx,
                         projection: 'EPSG:4326',
@@ -69,19 +92,12 @@
                         name: 'Ultra 2009',
                         transparent: true,
                         source: gpxSource,
-                        style: {
-                            stroke: {
-                                color: 'red',
-                            },
-                            point: {
-                                color: 'white',
-                                line: 'red',
-                            }
-                        },
+                        style: gpxStyle,
+                        labelEnabled: true,
                     });
 
                     return view.addLayer(gpxLayer);
-                }).then(FeatureToolTip.addLayer);
+                });
 
             debug.createTileDebugUI(menuGlobe.gui, view);
         </script>
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -332,4 +332,19 @@ export class FeatureCollection {
         this.features.push(ref);
         return ref;
     }
+
+    /**
+     * Transforms a given {@link Coordinates}, using the translation and the
+     * scale of this collection.
+     *
+     * @param {Coordinates} coordinates - The coordinates to transform
+     *
+     * @return {Coordinates} The same coordinates, with transformation applied.
+     */
+    transformCoordinates(coordinates) {
+        coordinates.x = (coordinates.x / this.scale.x) - this.translation.x;
+        coordinates.y = (coordinates.y / this.scale.y) - this.translation.y;
+        coordinates.z = (coordinates.z / this.scale.z) - this.translation.z;
+        return coordinates;
+    }
 }
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -1,0 +1,154 @@
+import * as THREE from 'three';
+import DEMUtils from 'Utils/DEMUtils';
+import Coordinates from 'Core/Geographic/Coordinates';
+
+const coord = new Coordinates('EPSG:4326');
+let rect;
+
+// set it once
+let STYLE_TRANSFORM = '';
+if (document.documentElement.style.transform !== undefined) {
+    STYLE_TRANSFORM = 'transform';
+} else if (document.documentElement.style.webkitTransform !== undefined) {
+    STYLE_TRANSFORM = 'webkitTransform';
+} else if (document.documentElement.style.mozTransform !== undefined) {
+    STYLE_TRANSFORM = 'mozTransform';
+} else if (document.documentElement.style.oTransform !== undefined) {
+    STYLE_TRANSFORM = 'oTransform';
+} else {
+    STYLE_TRANSFORM = 'transform';
+}
+
+/**
+ * An object that handles the display of a text and/or an icon, linked to a 3D
+ * position. The content of the `Label` is managed through a DOM object, in a
+ * `<div>` handled by the `Label2DRenderer`.
+ *
+ * @property {boolean} isLabel - Used to checkout whether this object is a
+ * Label. Default is true. You should not change this, as it is used internally
+ * for optimisation.
+ * @property {Element} content - The DOM object that contains the content of the
+ * label. The style and the position are applied on this object.
+ * @property {THREE.Vector3} position - The position in the 3D world of the
+ * label.
+ * @property {Coordinates} coordinates - The coordinates of the label.
+ */
+class Label extends THREE.Object3D {
+    /**
+     * @param {Element|string} content - The content; can be a
+     * string, with or without HTML tags in it, or it can be an Element.
+     * @param {Coordinates} coordinates - The world coordinates, where to place
+     * the Label.
+     * @param {Style} style - The style to apply to the content. Once the style
+     * is applied, it cannot be changed directly. However, if it really needed,
+     * it can be accessed through `label.content.style`, but it is highly
+     * discouraged to do so.
+     */
+    constructor(content, coordinates, style = {}) {
+        if (content == undefined || coordinates == undefined) {
+            throw new Error('content and coordinates are mandatory to add a Label');
+        }
+
+        super();
+
+        let _visible = this.visible;
+        // can't do an ES6 setter/getter here
+        Object.defineProperty(this, 'visible', {
+            set(v) {
+                if (v != _visible) { // avoid changing the style
+                    _visible = v;
+                    this.content.style.display = v ? 'block' : 'none';
+                    // TODO: add smooth transition for fade in/out
+                }
+            },
+            get() {
+                return _visible;
+            },
+        });
+
+        this.isLabel = true;
+        this.coordinates = coordinates;
+
+        this.projectedPosition = { x: 0, y: 0 };
+        this.boundaries = { left: 0, right: 0, top: 0, bottom: 0 };
+
+        this.content = document.createElement('div');
+        this.content.style.userSelect = 'none';
+        this.content.style.position = 'absolute';
+        if (typeof content == 'string') {
+            this.content.textContent = content || '';
+        } else {
+            this.content.appendChild(content);
+        }
+        this.baseContent = content;
+
+        if (style.isStyle) {
+            this.anchor = style.getTextAnchorPosition();
+            style.applyToHTML(this.content);
+        }
+
+        this.zoom = {
+            min: style.zoom && style.zoom.min != undefined ? style.zoom.min : 2,
+            max: style.zoom && style.zoom.max != undefined ? style.zoom.max : 24,
+        };
+    }
+
+    /**
+     * Moves a label on the screen, using screen coordinates. It updates the
+     * boundaries as it moves it.
+     *
+     * @param {number} x - X coordinates in pixels, from left.
+     * @param {number} y - Y coordinates in pixels, from top.
+     */
+    updateProjectedPosition(x, y) {
+        if (x != this.projectedPosition.x || y != this.projectedPosition.y) {
+            this.projectedPosition.x = x;
+            this.projectedPosition.y = y;
+
+            this.boundaries.left = x + this.offset.left;
+            this.boundaries.right = x + this.offset.right;
+            this.boundaries.top = y + this.offset.top;
+            this.boundaries.bottom = y + this.offset.bottom;
+        }
+    }
+
+    updateCSSPosition() {
+        this.content.style[STYLE_TRANSFORM] = `translate(${this.boundaries.left}px, ${this.boundaries.top}px)`;
+    }
+
+    /**
+     * Updates the screen dimensions of the label, using
+     * `getBoundingClientRect`.  It updates `width` and `height` property of the
+     * label, and the boundaries.
+     */
+    initDimensions() {
+        if (!this.offset) {
+            rect = this.content.getBoundingClientRect();
+            const width = Math.round(rect.width);
+            const height = Math.round(rect.height);
+            this.offset = {
+                left: width * this.anchor[0],
+                top: height * this.anchor[1],
+            };
+            this.offset.right = this.offset.left + width;
+            this.offset.bottom = this.offset.top + height;
+        }
+    }
+
+    update3dPosition(crs) {
+        this.coordinates.as(crs, coord);
+        coord.toVector3(this.position);
+        this.parent.worldToLocal(this.position);
+        this.updateMatrixWorld();
+    }
+
+    updateElevationFromLayer(layer) {
+        const elevation = DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z);
+        if (elevation && elevation != this.coordinates.z) {
+            this.coordinates.z = elevation;
+            return true;
+        }
+    }
+}
+
+export default Label;
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -7,11 +7,10 @@ import RenderMode from 'Renderer/RenderMode';
 import CRS from 'Core/Geographic/Crs';
 import Coordinates from 'Core/Geographic/Coordinates';
 import FeaturesUtils from 'Utils/FeaturesUtils';
-
 import { getMaxColorSamplerUnitsCount } from 'Renderer/LayeredMaterial';
-
 import Scheduler from 'Core/Scheduler/Scheduler';
 import Picking from 'Core/Picking';
+import LabelLayer from 'Layer/LabelLayer';
 
 export const VIEW_EVENTS = {
     /**
@@ -64,10 +63,21 @@ function _preprocessLayer(view, layer, parentLayer) {
         _syncGeometryLayerVisibility(layer, view);
         // Find projection layer, this is projection destination
         layer.projection = view.referenceCrs;
-    } else if (parentLayer.tileMatrixSets.includes(CRS.formatToTms(source.projection))) {
+    } else if (parentLayer && parentLayer.tileMatrixSets && parentLayer.tileMatrixSets.includes(CRS.formatToTms(source.projection))) {
         layer.projection = source.projection;
     } else {
-        layer.projection = parentLayer.extent.crs;
+        layer.projection = parentLayer && parentLayer.extent.crs;
+    }
+
+    if (layer.labelEnabled) {
+        const labelLayer = new LabelLayer(`${layer.id}-label`, view.referenceCrs);
+        labelLayer.source = source;
+        labelLayer.style = layer.style;
+
+        layer.whenReady = layer.whenReady.then(() => {
+            view.addLayer(labelLayer);
+            return layer;
+        });
     }
 
     return layer;
@@ -333,9 +343,11 @@ class View extends THREE.EventDispatcher {
             if (!filter || filter(layer)) {
                 result.push(layer);
             }
-            for (const attached of layer.attachedLayers) {
-                if (!filter || filter(attached, layer)) {
-                    result.push(attached);
+            if (layer.attachedLayers) {
+                for (const attached of layer.attachedLayers) {
+                    if (!filter || filter(attached, layer)) {
+                        result.push(attached);
+                    }
                 }
             }
         }
--- a/src/Layer/LabelLayer.js
+++ b/src/Layer/LabelLayer.js
@@ -1,0 +1,221 @@
+import LayerUpdateState from 'Layer/LayerUpdateState';
+import ObjectRemovalHelper from 'Process/ObjectRemovalHelper';
+import Layer from 'Layer/Layer';
+import Coordinates from 'Core/Geographic/Coordinates';
+import Extent from 'Core/Geographic/Extent';
+import Label from 'Core/Label';
+import { FEATURE_TYPES } from 'Core/Feature';
+
+const coord = new Coordinates('EPSG:4326', 0, 0, 0);
+
+let content;
+const _extent = new Extent('EPSG:4326', 0, 0, 0, 0);
+
+/**
+ * A layer to handle a bunch of `Label`. This layer can be created on its own,
+ * but it is better to use the option `labelEnabled` on another `Layer` to let
+ * it work with it (see the `vector_tile_raster_2d` example).
+ *
+ * @property {boolean} isLabelLayer - Used to checkout whether this layer is a
+ * LabelLayer.  Default is true. You should not change this, as it is used
+ * internally for optimisation.
+ * @property {string} crs - The crs of this layer.
+ */
+class LabelLayer extends Layer {
+    constructor(id, crs) {
+        super(id);
+
+        this.isLabelLayer = true;
+        this.crs = crs;
+    }
+
+    /**
+     * Reads each {@link FeatureGeometry} that contains label configuration, and
+     * creates the corresponding {@link Label}. To create a `Label`, a geometry
+     * needs to have a `label` object with at least a few properties:
+     * - `content`, which refers to `Label#content`
+     * - `position`, which refers to `Label#position`
+     * - (optional) `config`, containing miscellaneous configuration for the
+     *   label
+     *
+     * The geometry (or its parent Feature) needs to have a Style set.
+     *
+     * @param {FeatureCollection} data - The FeatureCollection to read the
+     * labels from.
+     * @param {Extent} extent
+     *
+     * @return {Label[]} An array containing all the created labels.
+     */
+    convert(data, extent) {
+        const labels = [];
+
+        const layerField = this.style && this.style.text && this.style.text.field;
+
+        // Converting the extent now is faster for further operation
+        extent.as(data.crs, _extent);
+        coord.crs = data.crs;
+
+        data.features.forEach((f) => {
+            // TODO: add support for LINE and POLYGON
+            if (f.type !== FEATURE_TYPES.POINT) {
+                return;
+            }
+
+            const featureField = f.style && f.style.text.field;
+
+            f.geometry.forEach((g) => {
+                const minzoom = (g.properties.style && g.properties.style.zoom.min)
+                    || (f.style && f.style.zoom.min)
+                    || (this.style && this.style.zoom && this.style.zoom.min);
+
+                // Don't create a label if it is in-between two steps of zoom
+                if (!this.source.isFileSource) {
+                    if (data.extent.zoom != minzoom) { return; }
+                } else if (extent.zoom != minzoom) { return; }
+
+                // NOTE: this only works because only POINT is supported, it
+                // needs more work for LINE and POLYGON
+                coord.setFromArray(f.vertices, g.size * g.indices[0].offset);
+                data.transformCoordinates(coord);
+                if (f.size == 2) { coord.z = 0; }
+                if (!_extent.isPointInside(coord)) { return; }
+
+                const geometryField = g.properties.style && g.properties.style.text.field;
+                if (!geometryField && !featureField && !layerField) {
+                    return;
+                } else if (geometryField) {
+                    content = g.properties.style.getTextFromProperties(g.properties);
+                } else if (featureField) {
+                    content = f.style.getTextFromProperties(g.properties);
+                } else if (layerField) {
+                    content = this.style.getTextFromProperties(g.properties);
+                }
+
+                // FIXME: Style management needs improvement, see #1318.
+                const label = new Label(content,
+                    coord.clone(),
+                    g.properties.style || f.style || this.style);
+
+                if (f.size == 2) {
+                    label.needsAltitude = true;
+                }
+
+                labels.push(label);
+            });
+        });
+
+        return labels;
+    }
+
+    // placeholder
+    preUpdate() {}
+
+    update(context, layer, node, parent) {
+        if (!parent && node.children.length) {
+            // if node has been removed dispose three.js resource
+            ObjectRemovalHelper.removeChildrenAndCleanupRecursively(this, node);
+            return;
+        }
+
+        if (this.frozen || !node.visible) {
+            return;
+        }
+
+        if (node.layerUpdateState[this.id] === undefined) {
+            node.layerUpdateState[this.id] = new LayerUpdateState();
+        }
+
+        const elevationLayer = node.material.getElevationLayer();
+        if (elevationLayer && node.layerUpdateState[elevationLayer.id].canTryUpdate()) {
+            node.children.forEach((c) => {
+                if (c.isLabel && c.needsAltitude && c.updateElevationFromLayer(this.parent)) {
+                    c.update3dPosition(this.crs);
+                }
+            });
+        }
+
+        if (!node.layerUpdateState[this.id].canTryUpdate()) {
+            return;
+        }
+
+        const extentsDestination = node.getExtentsByProjection(this.source.projection);
+
+        const extentsSource = [];
+        for (const extentDest of extentsDestination) {
+            const ext = this.source.projection == extentDest.crs ? extentDest : extentDest.as(this.source.projection);
+            if (!this.source.extentInsideLimit(ext)) {
+                node.layerUpdateState[this.id].noMoreUpdatePossible();
+                return;
+            }
+            extentsSource.push(extentDest);
+        }
+        node.layerUpdateState[this.id].newTry();
+
+        const command = {
+            layer: this,
+            extentsSource,
+            view: context.view,
+            threejsLayer: this.threejsLayer,
+            requester: node,
+        };
+
+        return context.scheduler.execute(command).then((result) => {
+            if (!result) { return; }
+
+            const renderer = context.view.mainLoop.gfxEngine.label2dRenderer;
+            const labelsDiv = [];
+
+            result.forEach((labels) => {
+                if (!node.parent) {
+                    labels.forEach((l) => {
+                        ObjectRemovalHelper.removeChildrenAndCleanupRecursively(this, l);
+                        renderer.removeLabelDOM(l);
+                    });
+                    return;
+                }
+
+                labels.forEach((label) => {
+                    if (label.needsAltitude) {
+                        label.updateElevationFromLayer(this.parent);
+                    }
+
+                    node.add(label);
+                    label.update3dPosition(this.crs);
+
+                    labelsDiv.push(label.content);
+                });
+            });
+
+            if (labelsDiv.length > 0) {
+                // Add all labels for this tile at once to batch it
+                node.domElement = document.createElement('div');
+                node.domElement.append(...labelsDiv);
+                renderer.domElement.appendChild(node.domElement);
+                node.domElementVisible = true;
+
+                // Batch update the dimensions of labels all at once to avoid
+                // redraw for at least this tile.
+                result.forEach(labels => labels.forEach(label => label.initDimensions()));
+                result.forEach(labels => labels.forEach((label) => { label.visible = false; }));
+
+                // Sort labels so they can be the first in the renderer. That
+                // way, we cull labels on parent tile first, and then on
+                // children tile. This allows a z-order priority, and reduce
+                // flickering.
+                node.children.sort(c => (c.isLabel ? -1 : 1));
+
+                // Necessary event listener, to remove any Label attached to
+                // this tile
+                node.addEventListener('removed', () => {
+                    renderer.hideNodeDOM(node);
+                    result.forEach(labels => labels.forEach(label => node.remove(label)));
+                    node.domElement.parentElement.removeChild(node.domElement);
+                });
+            }
+
+            node.layerUpdateState[this.id].noMoreUpdatePossible();
+        });
+    }
+}
+
+export default LabelLayer;
--- a/src/Renderer/Label2DRenderer.js
+++ b/src/Renderer/Label2DRenderer.js
@@ -83,6 +83,11 @@ class Label2DRenderer {
         this.domElement.style.top = 0;
         this.domElement.style.zIndex = 1;
 
+        // Used to destroy labels that are not added to the DOM
+        this.garbage = document.createElement('div');
+        this.garbage.style.display = 'none';
+        this.domElement.appendChild(this.garbage);
+
         this.halfWidth = 0;
         this.halfHeight = 0;
 
@@ -117,7 +122,7 @@ class Label2DRenderer {
         this.grid.visible.forEach((l, i) => {
             if (this.grid.insert(l, i)) {
                 l.visible = true;
-                l.move();
+                l.updateCSSPosition();
             }
         });
 
@@ -149,7 +154,7 @@ class Label2DRenderer {
             vector.setFromMatrixPosition(object.matrixWorld);
             vector.applyMatrix4(viewProjectionMatrix);
 
-            object.preMove(Math.round(vector.x * this.halfWidth + this.halfWidth), Math.round(-vector.y * this.halfHeight + this.halfHeight));
+            object.updateProjectedPosition(Math.round(vector.x * this.halfWidth + this.halfWidth), Math.round(-vector.y * this.halfHeight + this.halfHeight));
 
             // Are considered duplicates, labels that have the same screen
             // coordinates and the same base content.
--- a/src/Renderer/Label2DRenderer.js
+++ b/src/Renderer/Label2DRenderer.js
@@ -1,0 +1,190 @@
+import * as THREE from 'three';
+
+// A grid to manage labels on the screen.
+class ScreenGrid {
+    constructor(x = 12, y = 10, width, height) {
+        this.x = x;
+        this.y = y;
+
+        this.grid = [];
+        this.hidden = [];
+        this.visible = [];
+
+        this.resize();
+        this.reset();
+
+        this.width = width;
+        this.height = height;
+    }
+
+    // Reset each cell and hidden and visible.
+    reset() {
+        for (let i = 0; i < this.x; i++) {
+            for (let j = 0; j < this.y; j++) {
+                this.grid[i][j] = false;
+            }
+        }
+
+        this.hidden = [];
+        this.visible = [];
+    }
+
+    // Add rows if needed — but don't delete anything else. Columns are taken
+    // care in reset().
+    resize() {
+        for (let i = 0; i < this.x; i++) {
+            if (!this.grid[i]) {
+                this.grid[i] = [];
+            }
+        }
+    }
+
+    // Insert a label using its boundaries. It is either added to hidden or
+    // visible, given the result. The grid is populated with true for every
+    // filled cell.
+    insert(obj) {
+        const minx = Math.max(0, Math.floor(obj.boundaries.left / this.width * this.x));
+        const maxx = Math.min(this.x - 1, Math.floor(obj.boundaries.right / this.width * this.x));
+        const miny = Math.max(0, Math.floor(obj.boundaries.top / this.height * this.y));
+        const maxy = Math.min(this.y - 1, Math.floor(obj.boundaries.bottom / this.height * this.y));
+
+        for (let i = minx; i <= maxx; i++) {
+            for (let j = miny; j <= maxy; j++) {
+                if (this.grid[i][j]) {
+                    this.hidden.push(obj);
+                    return false;
+                }
+            }
+        }
+
+        for (let i = minx; i <= maxx; i++) {
+            for (let j = miny; j <= maxy; j++) {
+                this.grid[i][j] = true;
+            }
+        }
+
+        return true;
+    }
+}
+
+const viewProjectionMatrix = new THREE.Matrix4();
+const vector = new THREE.Vector3();
+
+/**
+ * This renderer is inspired by the
+ * [`THREE.CSS2DRenderer`](https://threejs.org/docs/#examples/en/renderers/CSS2DRenderer).
+ * It is instanciated in `c3DEngine`, as another renderer to handles Labels.
+ */
+class Label2DRenderer {
+    constructor() {
+        this.domElement = document.createElement('div');
+        this.domElement.style.overflow = 'hidden';
+        this.domElement.style.position = 'absolute';
+        this.domElement.style.top = 0;
+        this.domElement.style.zIndex = 1;
+
+        this.halfWidth = 0;
+        this.halfHeight = 0;
+
+        this.grid = new ScreenGrid();
+
+        this.infoTileLayer = undefined;
+    }
+
+    setSize(width, height) {
+        this.domElement.style.width = `${width}`;
+        this.domElement.style.height = `${height}`;
+
+        this.halfWidth = width / 2;
+        this.halfHeight = height / 2;
+
+        this.grid.width = width;
+        this.grid.height = height;
+        this.grid.x = Math.ceil(width / 30);
+        this.grid.y = Math.ceil(height / 30);
+
+        this.grid.resize();
+    }
+
+    render(scene, camera) {
+        if (!this.infoTileLayer) { return; }
+        this.grid.reset();
+
+        viewProjectionMatrix.multiplyMatrices(camera.projectionMatrix, camera.matrixWorldInverse);
+
+        this.culling(scene, this.infoTileLayer.currentMaxTileZoom, this.infoTileLayer.displayed.extent);
+
+        this.grid.visible.forEach((l, i) => {
+            if (this.grid.insert(l, i)) {
+                l.visible = true;
+                l.move();
+            }
+        });
+
+        this.grid.hidden.forEach((label) => { label.visible = false; });
+    }
+
+    culling(object, currentMaxZoom, extent) {
+        if (!object.isLabel) {
+            if (!object.visible) {
+                this.hideNodeDOM(object);
+                return;
+            }
+
+            // Don't go further if the node can't be in the screen space (it
+            // does not need to be rendered to continue the culling)
+            if (object.extent && !object.extent.intersectsExtent(extent)) {
+                this.hideNodeDOM(object);
+                return;
+            }
+
+            this.showNodeDOM(object);
+
+            object.children.forEach(c => this.culling(c, currentMaxZoom, extent));
+        // By verifying the maxzoom and the presence of the label inside the
+        // visible extent, we can filter more labels.
+        } else if (object.zoom.max <= currentMaxZoom || !extent.isPointInside(object.coordinates)) {
+            this.grid.hidden.push(object);
+        } else {
+            vector.setFromMatrixPosition(object.matrixWorld);
+            vector.applyMatrix4(viewProjectionMatrix);
+
+            object.preMove(Math.round(vector.x * this.halfWidth + this.halfWidth), Math.round(-vector.y * this.halfHeight + this.halfHeight));
+
+            // Are considered duplicates, labels that have the same screen
+            // coordinates and the same base content.
+            if (this.grid.visible.some(l => l.projectedPosition.x == object.projectedPosition.x
+                && l.projectedPosition.y == object.projectedPosition.y
+                && l.baseContent == object.baseContent)) {
+                object.parent.remove(object);
+                this.grid.hidden.push(object);
+            } else if (object.visible) {
+                // Give priority to already visible label, to reduce jittering
+                this.grid.visible.unshift(object);
+            } else {
+                this.grid.visible.push(object);
+            }
+        }
+    }
+
+    removeLabelDOM(label) {
+        this.garbage.appendChild(label.content);
+        this.garbage.innerHTML = '';
+    }
+
+    hideNodeDOM(node) {
+        if (node.domElementVisible == true) {
+            node.domElement.style.display = 'none';
+            node.domElementVisible = false;
+        }
+    }
+
+    showNodeDOM(node) {
+        if (node.domElementVisible == false) {
+            node.domElement.style.display = 'block';
+            node.domElementVisible = true;
+        }
+    }
+}
+
+export default Label2DRenderer;
--- a/docs/config.json
+++ b/docs/config.json
@@ -12,7 +12,8 @@
                 "Feature",
                 "FeatureGeometry",
                 "FeatureCollection",
-                "Style"
+                "Style",
+                "Label"
             ],
 
             "View": [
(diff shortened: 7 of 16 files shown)


src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,141 @@
...
+    setSize(width, height) {
+        this.domElement.style.width = `${width}`;
+        this.domElement.style.height = `${height}`;
+
+        this.halfWidth = width / 2;
+        this.halfHeight = height / 2;
+    }
+
+    render(scene, camera) {
+        viewMatrix.copy(camera.matrixWorldInverse);
```
gchoqueux: why you copy `camera.matrixWorldInverse`? `viewMatrix` is useless?

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    renderLabels(object, camera) {
+        // if the object isn't a Label, there are a few options:
+        // - it is visible
+        //      - it has a material but not visible
+        //          - hide its children Label
+        //          - iterate through other children (essentially TileMesh)
+        //      - otherwise iterate through its children
+        // - it not visible, hide all its potential Label children
```
gchoqueux: May be, the culling could be move in update `LabelLayer`.

It's possible to group Labels in one `div`, if it's in same `children` array?

Where is the case where `style.display = ''` ?

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,141 @@
...
+            return;
+        }
+
+        object.x = vector.x;
+        object.y = vector.y;
+
+        // each Label is stored in labels to be sorted in zOrder later
+        labels.push(object);
+
+        if (object.content.parentNode == null) {
```
gchoqueux: what is this part?

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * The final ordering of labels is done here:
+     * - order the labels by their x/y position on screen
+     * - check the intersection with close neighboors and remove some if they
+     *   are overlapping
+     * - move the remaining labels to their correct positions
+     */
+    zOrder() {
```
gchoqueux: There are 3 loops (4 with `labelLayer.update()`), how to minimize tree paths?

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,141 @@
...
+            } else {
+                i++;
+            }
+        }
+
+        for (let i = 0, l = labels.length; i < l; i++) {
+            label = labels[i];
+            label.content.style.zIndex = l - i;
+            label.content.style.display = '';
+            label.content.style[this.transform] = `${label.anchor} translate(${label.x * this.halfWidth + this.halfWidth}px,${-label.y * this.halfHeight + this.halfHeight}px)`;
```
gchoqueux: from what you've told me, this transformation is the most consuming...?

gchoqueux (commented):
I comment only the first commit

src/Renderer/Label2DRenderer.js
zarov: It is true, I'll fix it

src/Renderer/Label2DRenderer.js
zarov: The label is attached to the tile, but it hasn't been added to the DOM yet.

src/Renderer/Label2DRenderer.js
zarov: Yes, but by correctly culling in the process before, we should be able to keep it quite fast (I think like a maximum of 100~ labels updated each time)

gchoqueux:
Could move label example to gpx raster ? as this's not yet sufficiently optimized for the vector tiles.

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+     * @param {number} [config.minlevel=2] - Minimum level to display the label
+     * at.
+     * @param {number} [config.maxlevel=20] - Maximum level to display the
+     * label at.
+     * @param {Style} style - The style to apply to the content. Once the style
+     * is applied, it cannot be changed directly. However, if it really needed,
+     * it can be accessed through `label.content.style`, but it is highly
+     * discouraged to do so.
+     */
+    constructor(content, position, config = {}, style) {
```
gchoqueux: It's better to support only `Coordinates` to place label.

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+        if (content == undefined || position == undefined) {
+            throw new Error('content and position are mandatory to add a Label');
+        }
+
+        super();
+
+        this.isLabel = true;
+
+        if (position.isVector3) {
```
gchoqueux: Which interest to support `Vector3`?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+            this.coordinates = new Coordinates(config.crs, position);
+            this.position.copy(position);
+        } else if (position.isCoordinates) {
+            this.coordinates = position;
+            position.toVector3(this.position);
+        } else {
+            throw new Error('position is not valid; must be a THREE.Vector3 or Coordinates');
+        }
+
+        this.boundaries = { left: 0, right: 0, top: 0, bottom: 0 };
```
gchoqueux, line 73: May be use `Box2`?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+        this.content.style.userSelect = 'none';
+        this.content.style.position = 'absolute';
+        if (typeof content == 'string') {
+            content = content.replace('\n', '<br/>');
+            this.content.innerHTML = content || '';
+        } else {
+            this.content.appendChild(content);
+        }
+
+        if (config.id) {
```
gchoqueux: What is this `id`?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+            this.content.id = config.id;
+        }
+        if (config.class) {
```
gchoqueux: What is this `class`?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+            this.content.className = Array.isArray(config.class) ? config.class.join(' ') : config.class;
+        }
+
+        if (style && style.text) {
+            this.anchor = this.getAnchorPosition(style.text.anchor);
+            this.zOrder = style.text.zOrder;
+        }
+
+        if (style && style.icon) {
```
gchoqueux: Could you write a function `Style.toCss()`?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Checks if the label is in a specified node (usually a {@link TileMesh}.
+     *
+     * @param {TileMesh} node - The node to check with.
+     *
+     * @return {boolean} `true` if in, otherwise `false`.
+     */
+    isInNode(node) {
```
gchoqueux: It's too specific function, no?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+        return node.extent.isPointInside(this.coordinates);
+    }
+
+    /**
+     * Hides the label. All the necessary things are done here, to ensure a
+     * correct hiding of the label. The style of the content of the label should
+     * not be modified directly to hide it.
+     */
+    hide() {
```
gchoqueux: Could you write a `Label.visible` setter to encapsulate the display `content`.

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+     * Moves a label on the screen, using screen coordinates. It updates the
+     * boundaries as it moves it.
+     *
+     * @param {number} x - X coordinates in pixels, from left.
+     * @param {number} y - Y coordinates in pixels, from top.
+     * @param {string} transform - The name of the `transform` property, usually
+     * `transfom`, but can `-webkit-transform` for example, depending on the
+     * browser.
+     */
+    move(x, y, transform) {
```
gchoqueux: may be use a global constant for `transform` property ?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+        if (x != this.x || y != this.y) {
+            this.x = x;
+            this.y = y;
+            this.content.style[transform] = `${this.anchor} translate(${x}px, ${y}px)`;
+            this.updateBoundaries();
```
gchoqueux: it's necessary to `updateBoundaries` only if the label is tested in the Grid?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Tells if this label is intersecting with another.
+     *
+     * @param {Label} label - The other label to test against.
+     *
+     * @return {boolean} `true` if intersecting, `false` otherwise.
+     */
+    intersection(label) {
```
gchoqueux: intersection isn't used?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+
+    /**
+     * Updates the screen dimensions of the label, using
+     * `getBoundingClientRect`.  It updates `width` and `height` property of the
+     * label, and the boundaries.
+     */
+    updateDimensions() {
+        if (!this.width && !this.height) {
+            rect = this.content.getBoundingClientRect();
+            this.width = rect.width;
```
gchoqueux: may be write getter to link this.width and rect.width? or maybe it's too costly?

src/Core/Label.js
```diff
@@ -0,0 +1,267 @@
...
+            this.height = rect.height;
+            this.updateBoundaries();
+        }
+    }
+
+    getAnchorPosition(anchor) {
```
gchoqueux: could you move to `Style.toCss()`

src/Core/TileMesh.js
```diff
@@ -41,6 +41,18 @@ class TileMesh extends THREE.Mesh {
 
         this.layerUpdateState = {};
         this.isTileMesh = true;
+
+        // Necessary event listener, to remove any child (beside OBB) attached
+        // to this tile
+        this._removedHandler = this.addEventListener('removed', () => {
```
gchoqueux: Could you explain this part? there's `ObjectRemovalHelper`, it is similar ?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,166 @@
...
+        const labels = [];
+
+        data.features.forEach((f) => {
+            f.geometry.forEach((g) => {
+                if (g.label && g.label.content && g.label.position) {
+                    g.label.config = g.label.config || {};
+                    g.label.config.crs = g.label.config.crs || this.crs;
+
+                    if (g.label.position.z == Infinity) {
+                        g.label.position.z = DEMUtils.getElevationValueAt(this.parent, g.label.position) + 100;
```
gchoqueux: the z position must be updated with the parent tile, in `LabelLayer.update` function.

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,166 @@
...
+     * @return {Label[]} An array containing all the created labels.
+     */
+    convert(data) {
+        const labels = [];
+
+        data.features.forEach((f) => {
+            f.geometry.forEach((g) => {
+                if (g.label && g.label.content && g.label.position) {
+                    g.label.config = g.label.config || {};
+                    g.label.config.crs = g.label.config.crs || this.crs;
```
gchoqueux: why save `config` and `crs` in geometry?
`crs` is already in Feature.
Pass directly `params` in Label constructor?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,166 @@
...
+
+                    const label = new Label(g.label.content, g.label.position, g.label.config, f.style || (g.properties && g.properties.style));
+                    label.layer = label;
+
+                    if (this.source.isVectorSource) {
+                        // Shift from 3857 to referenceCrs to local
+                        label.coordinates.x = (label.coordinates.x / data.scale.x) - data.translation.x;
+                        label.coordinates.y = (label.coordinates.y / data.scale.y) - data.translation.y;
+
+                        label.coordinates.as(this.crs, label.coordinates);
```
gchoqueux: it would be best to keep the coordinates in their original projection. Especially when updating the altitude?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,166 @@
...
+                        label.coordinates.toVector3(label.position);
+
+                        // The label needs a translation to a local reference
+                        label.needsMoveToLocal = true;
```
gchoqueux: `needsMoveToLocal` isn't necessary? we can know when we add the label to local object

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,166 @@
...
+
+        if (node.layerUpdateState[this.id] === undefined) {
+            node.layerUpdateState[this.id] = new LayerUpdateState();
+        }
+
+        if (!node.layerUpdateState[this.id].canTryUpdate()) {
+            return;
+        }
+
+        const extentsDestination = node.getExtentsByProjection(this.source.projection) || [node.extent];
```
gchoqueux: `[node.extent]` is necessary?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,166 @@
...
+        };
+
+        return context.scheduler.execute(command).then((result) => {
+            if (!result) { return; }
+
+            result.forEach((labels) => {
+                const labelsToAdd = [];
+
+                labels.forEach((label) => {
+                    if (label.parent || !label.isInNode(node)) { return; }
```
gchoqueux: * `!label.isInNode(node)` : why labels aren't in node extent  ?
    you could add filtering to remove label  that are out of extents tile.

If label have parent, his `layerUpdateState` should keep him from returning here.

src/Parser/GeoJsonParser.js
```diff
@@ -51,6 +51,15 @@ const toFeature = {
         geometry.properties = properties;
         geometry.properties.style = new Style().setFromGeojsonProperties(properties, feature.type);
         this.populateGeometry(crsIn, coordsIn, geometry, setAltitude, feature);
+
+        // TODO: take a decision for labels with other FEATURE_TYPES
```
gchoqueux: We should keep `FEATURE_TYPES.POINT` and have a convert `FeaturePoint2Label`

src/Parser/GeoJsonParser.js
```diff
@@ -42,7 +42,7 @@ const toFeature = {
         }
         geometry.updateExtent();
     },
-    default(feature, crsIn, coordsIn, filteringExtent, setAltitude, properties) {
+    default(feature, crsIn, coordsIn, filteringExtent, setAltitude, properties, style) {
```
gchoqueux: why inherit the style?

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,163 @@
...
+
+    render(scene, camera) {
+        this.grid.reset();
+        viewProjectionMatrix.multiplyMatrices(camera.projectionMatrix, camera.matrixWorldInverse);
+        cameraSquaredPosition.setFromMatrixPosition(camera.matrixWorld);
+
+        // culling
+        this.culling(scene);
+
+        this.grid.forEach(c => c && c.show());
```
gchoqueux: Could move `Label.show()` when it inserts in the grid?

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,163 @@
...
+        //      - otherwise iterate through its children
+        // - it not visible, hide all its potential Label children
+        if (!object.isLabel) {
+            if (!object.visible) {
+                return;
+            }
+
+            if (object.material && !object.material.visible) {
+                object.children.forEach((c) => {
+                    if (c.isLabel) {
```
gchoqueux: may be group label in `THREE.Group` to avoid loop on all children.

src/Core/Label.js
zarov: Yes, this is a residue of my first work on this, when I did not know any coordinate to place a label. I'll remove it.

src/Core/Label.js
zarov: See other comment

src/Core/Label.js
zarov: I'll try it. I'm not sure about the gain here, as it introduces a heavier structure than this simple object.

src/Core/Label.js
zarov: As labels are HTML node, this corresponds to the `id` attribute of a node.

src/Core/Label.js
zarov: Same as `id`.

src/Core/Label.js
zarov: Good idea !

src/Core/Label.js
zarov: Maybe, it was to make things more understandable in `LabelLayer`. I'll change it to see how it looks.

src/Core/Label.js
zarov: Ok

src/Core/Label.js
zarov: It makes sense yes, I'll change it.

src/Core/Label.js
zarov: A label is only moved if added to the Grid, and then we need boundaries to test it, so it won't change much.

src/Core/Label.js
zarov: Not anymore, I'll remove it

src/Core/Label.js
zarov: Yes

src/Core/Label.js
zarov: I removed it

src/Core/Label.js
zarov: I removed it

src/Core/Label.js
zarov: It is too costly, as we access `width/height` a multiple times, and we don't want to check or recompute them everytime. Also, they shouldn't change much, once they are set they should be good.

src/Layer/LabelLayer.js
zarov: Done, thanks

src/Layer/LabelLayer.js
zarov: This is for `FileSource`, as each Label is attached to a tile. If this check is not done,  there is a duplication of label. But maybe it can (and should) be done before. I'm looking into it.

src/Parser/GeoJsonParser.js
zarov: I removed this part

src/Parser/GeoJsonParser.js
zarov: I moved the parts in parsers to `LabelLayer#convert`.

src/Layer/LabelLayer.js
zarov: I removed the complicated `config` object for now. I need to add back `minlevel` and `maxlevel`.

src/Core/Label.js
```diff
@@ -0,0 +1,140 @@
...
+        if (typeof content == 'string') {
+            content = content.replace('\n', '<br/>');
+            this.content.innerHTML = content || '';
+        } else {
+            this.content.appendChild(content);
+        }
+
+        if (style) {
+            this.anchor = style.getTextAnchorPositionInCSS();
+            this.zOrder = style.text.zOrder;
```
gchoqueux: could you clean and comment this part?

src/Core/Label.js
```diff
@@ -0,0 +1,140 @@
...
+
+            style.addIconToElement(this.content);
+            style.applyToHTML(this.content);
```
gchoqueux: could you merge the functions?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+            if (f.type !== FEATURE_TYPES.POINT) {
+                return;
+            }
+
+            const featureField = f.style && f.style.text.field;
+
+            f.geometry.forEach((g) => {
+                const geometryField = g.properties.style && g.properties.style.text.field;
+
+                if (!geometryField && !featureField && !layerField) {
```
gchoqueux, line 84: `Style` handling needs architecture refactoring.

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                    return;
+                } else {
+                    content = g.properties[geometryField || featureField || layerField];
+                }
+
+                // NOTE: this only works because only POINT is supported, it
+                // needs more work for LINE and POLYGON
+                vertices = f.getVerticesFromGeometry(g);
```
gchoqueux: Could you strike directly the point?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                    label.coordinates.as(this.crs, label.coordinates);
+                    label.coordinates.toVector3(label.position);
+
+                    // The label needs a translation to a local reference
+                    label.needsMoveToLocal = true;
+                }
+
+                if (f.size == 2) {
+                    label.needsAltitude = true;
+                }
```
gchoqueux, line 101: Why ?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                labels.push(label);
+            });
+        });
+
+        return labels;
+    }
+
+    // TODO: see how to factorize this behavior with FeatureProcessing which is
+    // similar.
+    update(context, layer, node, parent) {
```
gchoqueux, line 113: Very similar to `featureProcessing#update`?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+        };
+
+        return context.scheduler.execute(command).then((result) => {
+            if (!result) { return; }
+
+            result.forEach((labels) => {
+                const labelsToAdd = [];
+
+                labels.forEach((label) => {
+                    if (label.parent || !node.extent.isPointInside(label.coordinates)) { return; }
```
gchoqueux: it is necessary to see the best place to filter, there is already a mechanism in the geojson parser

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+
+                    label.minlevel = node.extent.zoom;
+                    if (this.source.isVectorSource && label.needsMoveToLocal) {
+                        node.worldToLocal(label.position);
```
gchoqueux: it's not necessary to convert position.

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                }
+
+                if (labelsToAdd.length > 0) {
+                    labelsToAdd.forEach((l) => {
+                        node.add(l);
+                        context.view.mainLoop.gfxEngine.label2dRenderer.domElement.appendChild(l.content);
+
+                        if (l.needsAltitude) {
+                            // TODO: add a Style property for the offset
+                            l.position.z = DEMUtils.getElevationValueAt(this.parent, l.coordinates) + 2;
```
gchoqueux: Update altitude label when the DEM elevation is updated.

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                    label.minlevel = node.extent.zoom;
+                    if (this.source.isVectorSource && label.needsMoveToLocal) {
+                        node.worldToLocal(label.position);
+                        label.needsMoveToLocal = false;
+                    }
+
+                    labelsToAdd.push(label);
+                });
+
+                if (!node.parent) {
```
gchoqueux, line 169: move before `labelsToAdd` building? WDYT?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                        node.add(l);
+                        context.view.mainLoop.gfxEngine.label2dRenderer.domElement.appendChild(l.content);
+
+                        if (l.needsAltitude) {
+                            // TODO: add a Style property for the offset
+                            l.position.z = DEMUtils.getElevationValueAt(this.parent, l.coordinates) + 2;
+                        }
+
+                        l.visible = true;
+                        l.updateDimensions();
```
gchoqueux: not necessary?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,197 @@
...
+                    });
+                }
+            });
+
+            node.updateMatrixWorld();
```
gchoqueux: not necessary, update only label.

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,156 @@
...
+
+    render(scene, camera) {
+        this.grid.reset();
+        viewProjectionMatrix.multiplyMatrices(camera.projectionMatrix, camera.matrixWorldInverse);
+        cameraSquaredPosition.setFromMatrixPosition(camera.matrixWorld);
+
+        // culling
+        this.culling(scene);
+
+        this.grid.forEach((c) => {
```
gchoqueux: move in `grid#insert`

src/Core/Label.js
zarov: Done

src/Core/Label.js
zarov: Done in #1345

src/Layer/LabelLayer.js
zarov: Done

src/Layer/LabelLayer.js
zarov: I don't see another way to do it ? Keep in mind that not all labels need have a computed altitude.

src/Layer/LabelLayer.js
zarov: The part before the `scheduler.execute`, but I can discard the comment as it was more of something I thought at when I began to work on it.

src/Layer/LabelLayer.js
zarov: If I don't do this, I have a problem with `FileSource`. As it call `update` for each tile, and for a file such as the example with GPX, it always read __all__ points, and thus it creates a `Label` each time. So we have at the end a lot of duplicated labels. If you have another solution, I'm happy to implement it !

src/Layer/LabelLayer.js
zarov: Yes, done

src/Layer/LabelLayer.js
gchoqueux: ok

src/Renderer/Label2DRenderer.js
zarov: Forgot to do it, done !

src/Layer/LabelLayer.js
zarov: I have to do it after adding it to the DOM, I can't really see another place to do it, because I can get the dimensions only when the content of the Label is visible.

src/Layer/LabelLayer.js
zarov: Yes, done

src/Core/Label.js
```diff
@@ -0,0 +1,138 @@
...
+            },
+        });
+
+        this.isLabel = true;
+        this.coordinates = coordinates;
+        coordinates.toVector3(this.position);
+
+        this.boundaries = { left: 0, right: 0, top: 0, bottom: 0 };
+
+        this.content = document.createElement('div');
```
gchoqueux, line 75: It's possible to create css file and import properties to instance the div?

src/Core/Label.js
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        if (style) {
+            this.anchor = style.getTextAnchorPositionInCSS();
+            style.applyToHTML(this.content);
+        }
+
+        this.addEventListener('removed', () => { this.visible = false; });
+
+        // FIXME: this needs to be moved outside, to be executed only on start.
```
gchoqueux: Ok.
There's the same `document` in `FeatureToTexture.js` there is no problem.

src/Core/TileMesh.js
gchoqueux: Could you move this handler, when the labels are added to the parent?
If you handle when Labels are added, you could remove the test `this.children.length > 1` because there will labels in children.

src/Core/TileMesh.js
```diff
@@ -41,6 +41,18 @@ class TileMesh extends THREE.Mesh {
...
+            this.removeEventListener(this._removedHandler);
+            // > 1 permits to filter tiles with only OBB
+            if (this.children.length > 1) {
+                const toRemove = [];
+                this.children.forEach((c) => { if (c.isLabel) { toRemove.push(c); } });
```
gchoqueux: This code could work?
`this.children.forEach((c) => { if (c.isLabel) { this.remove(c); } });`

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,189 @@
...
+                coord.setFromArray(f.vertices, g.size * g.indices[0].offset);
+                if (f.size == 2) { coord.z = 0; }
+
+                if (this.source.isFileSource && !_extent.isPointInside(coord)) { return; }
+
+                // FIXME: Style management needs improvement, see #1318.
+                // NOTE: the config param is empty, but will be filled in the
+                // next version.
+                const label = new Label(content, coord.clone(), g.properties.style || f.style || this.style);
+                label.layer = label;
```
gchoqueux: it isn't `label.layer = layer;`?

If it's the case, could add layer in  `constructor parameter `?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,189 @@
...
+
+                // coord.x = (coord.x / data.scale.x) - data.translation.x;
+                // coord.y = (coord.y / data.scale.y) - data.translation.y;
+                coord.as(this.crs, coord);
+                coord.toVector3(label.position);
+                label.needsToBeMoved = true;
```
gchoqueux: why don't do this in constructor?

```suggestion
      // need to add layer parameter in Label constructor
      this.coord.as(this.layer.crs, this.position);
      this.needsToBeMoved = true;
```

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,189 @@
...
+
+                labels.forEach((label) => {
+                    if (label.parent) { return; }
+
+                    label.minlevel = node.extent.zoom;
+
+                    if (label.needsToBeMoved) {
+                        node.worldToLocal(label.position);
+                        label.needsToBeMoved = false;
+                    }
```
gchoqueux: to remove `needsToBeMoved` 

```suggestion
if (label.needsAltitude) {
    // TODO: add a Style property for the offset
    label.coordinates.z = DEMUtils.getElevationValueAt(this.parent, label.coordinates);
}
label.coord.as(layer.crs, label.position);
node.worldToLocal(label.position);
```

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,189 @@
...
+
+                    node.add(label);
+                    context.view.mainLoop.gfxEngine.label2dRenderer.domElement.appendChild(label.content);
+
+                    if (label.needsAltitude) {
+                        // TODO: add a Style property for the offset
+                        label.position.z = DEMUtils.getElevationValueAt(this.parent, label.coordinates) + 2;
```
gchoqueux: you add `+2` to world coordinate and not to altitude.

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,189 @@
...
+                    }
+
+                    label.visible = true;
+                    label.updateDimensions();
+                    label.updateMatrixWorld();
```
gchoqueux: May be override `Label#updateMatrixWorld` to encapsulate `updateDimensions`?

src/Core/TileMesh.js
zarov: Good idea, will do !

src/Core/TileMesh.js
zarov: No, it breaks the `forEach` unfortunately.

src/Layer/LabelLayer.js
zarov: This is a mistake indeed, thanks. I copied this from `FeatureProcessing`, is it really necessary ? I can't remember the use for it.

src/Layer/LabelLayer.js
zarov: This won't work,  `Coordinates#as()` can't take a `THREE.Vector3` as second argument. But the operation could be done in the constructor indeed, I'm trying it.

src/Layer/LabelLayer.js
zarov: It won't work, as `worldToLocal` will be called everytime we add back a `Label` to the scene. Also, it applies a conversion everytime too.

src/Layer/LabelLayer.js
zarov: Won't `updateDimensions` be called too many times if I do this ?

src/Layer/LabelLayer.js
zarov: At this point, the label has been moved to a local position (so its parent is a node). The orientation of `z` is following the altitude axis.

src/Core/Label.js
zarov: Indeed, that's weird. I'm looking into it.

src/Layer/LabelLayer.js
gchoqueux: `isInNode `is removed

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,151 @@
...
+        // - it not visible, hide all its potential Label children
+        if (!object.isLabel) {
+            if (!object.visible) {
+                return;
+            }
+
+            if (object.material && !object.material.visible) {
+                object.children.forEach((c) => {
+                    if (c.isLabel) {
+                        c.visible = false;
```
gchoqueux: could you remove `c.visible = false;`? because is already hidden in `computeLabelPosition`

src/Layer/LabelLayer.js
gchoqueux: WDYT?

src/Core/TileMesh.js
gchoqueux: I tested without problem, what's breacking message ?

src/Renderer/Label2DRenderer.js
zarov: But `computeLabelPosition` isn't called for every label, so it isn't always hidden.

src/Core/Label.js
zarov: Yes, I'll do it in another PR, as it can be done entirely in `Style`.

src/Layer/LabelLayer.js
gchoqueux: I'm not sure because label is added to node, only one time, isn't it?

src/Layer/LabelLayer.js
gchoqueux: yes, sorry!

src/Layer/LabelLayer.js
zarov: Not if we unzoom and the node is deleted, and we zoom again, the node is added back and we go again here.

src/Core/TileMesh.js
zarov: It is because of this line https://github.com/mrdoob/three.js/blob/master/src/core/Object3D.js#L371

By calling `splice` in a `forEach`, the array is completely rearranged, but the `forEach` doesn't know it, so you miss a test on one element. If you try this, you can see what I mean:
```js
const b = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
// this should remove 3 and 4 from the array
b.forEach((e, i, a) => { if (e == 3 || e == 4) { a.splice(i, 1); } });
// but if you print b, you'll see that 4 is still present
console.log(b);
```

src/Layer/LabelLayer.js
zarov: No idea, I shamelessly copied this from `FeatureProcessing`.

src/Layer/LabelLayer.js
zarov: This is done in the constructor now

src/Core/TileMesh.js
gchoqueux: ok

src/Layer/LabelLayer.js
gchoqueux: but if the node is deleted a new label is instanced, no?

src/Layer/LabelLayer.js
gchoqueux: to verify

src/Layer/LabelLayer.js
zarov: No, it is only hidden. In my first thoughts, it was costing less to keep a hidden label instead of creating one everytime. But if there is a lot of them, it can be costly indeed. I'll think of a solution, but it seems that adding a `removed` handler will do the trick.

zarov: In fact, as it is parsed data, it automatically goes into cache (see `DataSourceProvider`). So I'm not sure on what to do.

src/Renderer/Label2DRenderer.js
gchoqueux: add event in `material.visible` to hidden/show the label.

src/Layer/LabelLayer.js
gchoqueux: Brainstorm a better solution

gchoqueux: we're putting aside for the moment

src/Renderer/Label2DRenderer.js
gchoqueux: Add event in `material.visible` to hidden/show the label.

zarov: Added it in #1351

src/Core/TileMesh.js
zarov: As I moved it in `LabelLayer`, I was able to remove the necessary selection of labels (see here https://github.com/iTowns/itowns/pull/1303/files#diff-9229f99ccb02f6c4aa8d3d42de700a92R179-R182)

zarov:
I rebased and cleaned everything, it should be good to go now !

zarov:
Rebased again, on but needs to wait for #1353 to be merged before.

src/Core/Label.js
zarov: It seems that fixing other things (having nothing to do with this) made this works outside the constructor. I'll check a few things before pushing it.

gmaillet:
I think that there is confusion in updateElevationFromLayer between z in the geocentric coordinates and z above the ground. 
It can produce large errors in the placement of labels.

gmaillet:
In addition to the altitude issue (see above), there are errors related to updating the position in local coordinates to the tile (worldToLocal). For exemple, when a tile is removed, label survive and could be added in a new tile. In these cases,  position need to be updated.
I propose to separate the coordinated part (in the Label class) and the position part in the LabelLayer class.

```diff
diff --git a/src/Core/Label.js b/src/Core/Label.js
index 6f5848ba..0ea8e826 100644
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -75,15 +75,8 @@ class Label extends THREE.Object3D {
         if (options.translation && options.scale) {
             this.coordinates.x = (this.coordinates.x / options.scale.x) - options.translation.x;
             this.coordinates.y = (this.coordinates.y / options.scale.y) - options.translation.y;
-            this.coordinates.as(options.crs, coord);
-        } else {
-            this.coordinates.as(options.crs || this.coordinates.crs, coord);
         }
 
-        coord.toVector3(this.position);
-
-        this.needsToBeMoved = true;
-
         this.boundaries = { left: 0, right: 0, top: 0, bottom: 0 };
 
         this.content = document.createElement('div');
@@ -139,7 +132,11 @@ class Label extends THREE.Object3D {
     }
 
     updateElevationFromLayer(layer) {
-        this.position.z = DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z) || 0;
+        const alti = DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z);
+        if (alti && alti > -99999 && alti != this.coordinates.z) {
+            this.coordinates.z = alti;
+            return true;
+        }
     }
 }
 
diff --git a/src/Layer/LabelLayer.js b/src/Layer/LabelLayer.js
index fcd7a83d..6aed63ab 100644
--- a/src/Layer/LabelLayer.js
+++ b/src/Layer/LabelLayer.js
@@ -133,7 +133,12 @@ class LabelLayer extends Layer {
         if (!node.layerUpdateState[node.material.getElevationLayer().id].canTryUpdate()) {
             node.children.forEach((c) => {
                 if (c.isLabel && c.needsAltitude) {
-                    c.updateElevationFromLayer(this.parent);
+                    if (c.updateElevationFromLayer(this.parent)) {
+                        c.coordinates.as(this.crs, coord);
+                        coord.toVector3(c.position);
+                        node.worldToLocal(c.position);
+                        c.updateMatrixWorld();
+                    }
                 }
             });
         }
@@ -183,14 +188,13 @@ class LabelLayer extends Layer {
 
                     label.minzoom = node.extent.zoom;
 
-                    if (label.needsToBeMoved) {
-                        node.worldToLocal(label.position);
-                        label.needsToBeMoved = false;
-                    }
 
                     if (label.needsAltitude) {
                         label.updateElevationFromLayer(this.parent);
                     }
+                    label.coordinates.as(this.crs, coord);
+                    coord.toVector3(label.position);
+                    node.worldToLocal(label.position);
 
                     node.add(label);
                     label.updateMatrixWorld();

```

zarov:
Thanks a lot, it seems to solve all my problems ! I'm implementing it right now

src/Renderer/Label2DRenderer.js
gchoqueux: do you group the labels in a div?

src/Layer/LabelLayer.js
gchoqueux: it's not necessary in `featureProcessing`, could you try it?

src/Core/Label.js
gchoqueux: Fixed with `--file bootstrap.js`

examples/source_file_gpx_raster.html
```diff
@@ -59,29 +60,42 @@
...
+                        point: {
+                            color: 'white',
+                            line: 'red',
+                        },
+                        text: {
+                            field: '{name}',
+                            transform: 'uppercase',
+                            font: ['Arial', 'sans-serif'],
+                            halo: {
+                                color: 'white',
```
gchoqueux, line 79: There's not halo in the gpx example?

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+ * label. The style and the position are applied on this object.
+ * @property {THREE.Vector3} position - The position in the 3D world of the
+ * label.
+ * @property {Coordinates} coordinates - The coordinates of the label.
+ */
+class Label extends THREE.Object3D {
+    /**
+     * @param {Element|string} content - The content; can be a
+     * string, with or without HTML tags in it, or it can be an Element.
+     * @param {Coordinates} coordinates - The world coordinates, where to place
```
gchoqueux, line 40: What is the coordinates projection? it's the same of `view.referenceCrs`?

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+            },
+            get() {
+                return _visible;
+            },
+        });
+
+        this.isLabel = true;
+        this.coordinates = coordinates;
+
+        if (options.translation && options.scale) {
```
gchoqueux: Could you move the transformation out of the constructor?
You could add a method in `Feature` to get the good coordinates.

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+            this.coordinates.x = (this.coordinates.x / options.scale.x) - options.translation.x;
+            this.coordinates.y = (this.coordinates.y / options.scale.y) - options.translation.y;
+            this.coordinates.as(options.crs, coord);
+        } else {
+            this.coordinates.as(options.crs || this.coordinates.crs, coord);
+        }
+
+        coord.toVector3(this.position);
+
+        this.needsToBeMoved = true;
```
gchoqueux: boolean to move on local object?
Could you add a comment?

examples/source_file_gpx_raster.html
zarov: There is not anymore, because I can't merge all the styles unfortunately, so I have to choose a style among three: https://github.com/iTowns/itowns/pull/1303/files/9322d13d0f4a2fdf548bcc7704f684eec88d151e#diff-9229f99ccb02f6c4aa8d3d42de700a92R102
I am not sure how I can fix this problem yet.

src/Core/Label.js
zarov: Not necessary, it can be anything, as `label.position` is not affected by the coordinates projection.

src/Core/Label.js
zarov: This has been removed, I'm waiting on #1377 to be merged to rebase this PR.

src/Core/Label.js
zarov: Ok, will do

src/Renderer/Label2DRenderer.js
zarov: Yes, this is done in `LabelLayer`

src/Layer/LabelLayer.js
zarov: I removed it

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Moves a label on the screen, using screen coordinates. It updates the
+     * boundaries as it moves it.
+     *
+     * @param {number} x - X coordinates in pixels, from left.
+     * @param {number} y - Y coordinates in pixels, from top.
+     */
+    preMove(x, y) {
+        if (x != this.x || y != this.y) {
+            this.x = x;
```
gchoqueux: could you initialize `x` and `y` in constructor in object `projectedPosition`, so as not to confuse with `label` 3d position?

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+
+            this.boundaries.left = this.x;
+            this.boundaries.right = this.x + this.width;
+            this.boundaries.top = this.y;
+            this.boundaries.bottom = this.y + this.height;
+        }
+    }
+
+    move() {
+        this.visible = true;
```
gchoqueux: I'm not sure to display the label when the method is `move`. 
Even if it's the rendering loop they're done in sequence.

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,224 @@
...
+                if (!node.parent) {
+                    labels.forEach((l) => {
+                        ObjectRemovalHelper.removeChildrenAndCleanupRecursively(this, l);
+                        renderer.removeLabelDOM(l);
+                    });
+                    return;
+                }
+
+                labels.forEach((label) => {
+                    if (label.parent) { return; }
```
gchoqueux: this can't happen?
I kind of tested label has no parent here.

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+     */
+    updateDimensions() {
+        if (!this.width && !this.height) {
+            rect = this.content.getBoundingClientRect();
+            this.width = rect.width;
+            this.height = rect.height;
+        }
+    }
+
+    updatePositionFromNode(node, crs) {
```
gchoqueux: Could you add property `layer` in label and remove crs param

```js
update3DPosition(crs) {
      this.coordinates.as(crs, coord);
      coord.toVector3(this.position);
      if (this.parent) {
            this.parent.worldToLocal(this.position);
      }
      this.updateMatrixWorld();
}
```

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,224 @@
...
+                    }
+
+                    node.add(label);
+                    label.updatePositionFromNode(node, this.crs);
+
+                    labelsDiv.push(label.content);
+                });
+            });
+
+            if (labelsDiv.length > 0) {
```
gchoqueux, line 189: because  `if (label.parent) { return; }` could be removed so `labelsDiv` could be remove?

src/Core/Label.js
```diff
@@ -0,0 +1,146 @@
...
+        this.visible = true;
+        this.content.style[STYLE_TRANSFORM] = `${this.anchor} translate(${this.x}px, ${this.y}px)`;
+    }
+
+    /**
+     * Updates the screen dimensions of the label, using
+     * `getBoundingClientRect`.  It updates `width` and `height` property of the
+     * label, and the boundaries.
+     */
+    updateDimensions() {
```
gchoqueux: could you rename `initDimensions `because is called only one time?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,224 @@
...
+                // Add all labels for this tile at once to batch it
+                node.domElement = document.createElement('div');
+                node.domElement.append(...labelsDiv);
+                renderer.domElement.appendChild(node.domElement);
+                node.domElementVisible = true;
+
+                // Batch update the dimensions of labels all at once to avoid
+                // redraw for at least this tile.
+                result.forEach(labels => labels.forEach(label => label.updateDimensions()));
+                result.forEach(labels => labels.forEach((label) => { label.visible = false; }));
```
gchoqueux, line 199: There are 2 loops for performance?

Could you try :
```js
result.forEach(labels => labels.forEach((label) => {
    label.updateDimensions();
    label.visible = false;
```
or 
```js
result.forEach(labels => labels.forEach((label) => {
    label.content.style.display = 'hidden';
    label.updateDimensions();
    label.content.style.display = 'none';
}));
```

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,224 @@
...
+
+                // Reorganize all Label to be the first in the list of children,
+                // to have an horizontal rendering instead of a vertical one
+                node.children.sort(c => (c.isLabel ? -1 : 1));
```
gchoqueux, line 205: I don't know `horizontal rendering`

src/Renderer/Label2DRenderer.js
```diff
@@ -0,0 +1,183 @@
...
+                this.hideNodeDOM(object);
+                return;
+            }
+
+            this.showNodeDOM(object);
+
+            object.children.forEach(c => this.culling(c, currentMaxZoom, extent));
+        // By verifying the maxzoom and the presence of the label inside the
+        // visible extent, we can filter more labels.
+        } else if (object.maxzoom <= currentMaxZoom || !extent.isPointInside(object.coordinates)) {
```
gchoqueux: the object visibility and material visibility aren't enough to filter out the labels?
may be check `tile.material.visible` and if the children tile have label.

src/Core/Label.js
zarov: Done !

src/Layer/LabelLayer.js
zarov: Not anymore indeed, I'll remove it.

src/Layer/LabelLayer.js
zarov: No, because `result` can contain empty arrays.

src/Core/Label.js
zarov: Yes done

src/Layer/LabelLayer.js
zarov: No, this needs to be separated for performance: it is faster to batch the reads (here `updateDimensions`) and then batch the `writes` (here `display = 'none'`).

src/Layer/LabelLayer.js
zarov: I removed that part, it was useful before but it doesn't seem the case anymore.

src/Renderer/Label2DRenderer.js
zarov: No, it isn't: a material not visible does not mean a label is not visible anymore.

src/Core/Feature.js
```diff
@@ -332,4 +332,18 @@ export class FeatureCollection {
...
+    /**
+     * Transforms a given {@link Coordinates}, using the translation and the
+     * scale of this collection.
+     *
+     * @param {Coordinates} coordinates - The coordinates to transform
+     *
+     * @return {Coordinates} The same coordinates, with transformation applied.
+     */
+    transformCoordinates(coordinates) {
+        coordinates.x = (coordinates.x / this.scale.x) - this.translation.x;
```
gchoqueux, line 345: Could you add z/altitude transformation ?

src/Core/Label.js
```diff
@@ -0,0 +1,154 @@
...
+            this.projectedPosition.y = y;
+
+            this.boundaries.left = x + this.offset.left;
+            this.boundaries.right = x + this.offset.right;
+            this.boundaries.top = y + this.offset.top;
+            this.boundaries.bottom = y + this.offset.bottom;
+        }
+    }
+
+    move() {
```
gchoqueux: Could you rename to `updateCssPosition` ?

src/Core/Label.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Moves a label on the screen, using screen coordinates. It updates the
+     * boundaries as it moves it.
+     *
+     * @param {number} x - X coordinates in pixels, from left.
+     * @param {number} y - Y coordinates in pixels, from top.
+     */
+    preMove(x, y) {
```
gchoqueux: Could you rename to `updateProjectedPosition`?

src/Core/View.js
```diff
@@ -64,10 +63,21 @@ function _preprocessLayer(view, layer, parentLayer) {
...
         layer.projection = view.referenceCrs;
-    } else if (parentLayer.tileMatrixSets.includes(CRS.formatToTms(source.projection))) {
+    } else if (parentLayer && parentLayer.tileMatrixSets && parentLayer.tileMatrixSets.includes(CRS.formatToTms(source.projection))) {
         layer.projection = source.projection;
     } else {
-        layer.projection = parentLayer.extent.crs;
+        layer.projection = parentLayer && parentLayer.extent.crs;
+    }
+
+    if (layer.labelEnabled && source) {
```
gchoqueux: I'm not sure, there are Layers without source?

src/Core/View.js
```diff
@@ -64,10 +63,21 @@ function _preprocessLayer(view, layer, parentLayer) {
...
+        const labelLayer = new LabelLayer(`${layer.id}-label`, view.referenceCrs);
+        labelLayer.source = source;
+        labelLayer.style = layer.style;
+
+        layer.whenReady = layer.whenReady.then(() => {
+            view.addLayer(labelLayer);
+            return layer;
+        });
```
gchoqueux, line 80: could you try `view.addLayer(labelLayer);` only?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,221 @@
...
+        data.features.forEach((f) => {
+            // TODO: add support for LINE and POLYGON
+            if (f.type !== FEATURE_TYPES.POINT) {
+                return;
+            }
+
+            const featureField = f.style && f.style.text.field;
+
+            f.geometry.forEach((g) => {
+                const minzoom = (g.properties.style && g.properties.style.zoom && g.properties.style.zoom.min)
```
gchoqueux: `Style.zoom == undefined` is possible?

src/Layer/LabelLayer.js
```diff
@@ -0,0 +1,221 @@
...
+                    content = g.properties.style.getTextFromProperties(g.properties);
+                } else if (featureField) {
+                    content = f.style.getTextFromProperties(g.properties);
+                } else if (layerField) {
+                    content = this.style.getTextFromProperties(g.properties);
+                }
+
+                // FIXME: Style management needs improvement, see #1318.
+                const label = new Label(content,
+                    coord.clone(),
```
gchoqueux, line 96: If is possible, could you remove `clone()`?

src/Core/Label.js
zarov: Done

src/Core/Label.js
zarov: Done

src/Core/View.js
zarov: Not anymore, thanks !

src/Core/Feature.js
zarov: Done, I thought they were `Vector2`

src/Layer/LabelLayer.js
zarov: Well `coord` is global, so I can't really avoid this.

gchoqueux: ok

src/Core/View.js
zarov: It works, thanks !

src/Layer/LabelLayer.js
zarov: It can, but only for `this.style`. If you take a look at `View`, you have the following line:
```js
labelLayer.style = layer.style;
```
which default to `{}` if you check `ColorLayer`.

gchoqueux (approved):
**Thank you very much for this big job that brings an important feature!**

Mkonini:
Good job, thnx a lot!

PR #1312, opened 2020-02-04
fix(parser): add an option to parse symbols as circles in VT

### LINKED ISSUE #1310: The points features are not displayed on the vector tiles layers
iTowns do not handle the display of points feature of the vector tile Layer.

Ex. The sub stations are not displayed with iTowns, but they are with openLayers (see attached screenshots ol_vt_points.png / it_vt_points.png)
it_vt_points
ol_vt_points

zarov:
Some modifications have been done on the Style in prevision of the
support of symbol in Vector Tiles, but previously the symbols were
represented as circle.

To get back to this behavior, a flag (`symbolToCircle`) can be specified
in the options of the ColorLayer containing a VectorTileSource.

Fix #1310

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -126,9 +126,10 @@ class Style {
      * @param {object} layer vector tile layer.
      * @param {Number} zoom vector tile layer.
      * @param {Object} sprites vector tile layer.
+     * @param {boolean} [symbolToCircle=false]
      * @returns {Style}
      */
-    setFromVectorTileLayer(layer, zoom, sprites) {
+    setFromVectorTileLayer(layer, zoom, sprites, symbolToCircle = false) {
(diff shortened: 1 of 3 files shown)


examples/vector_tile_raster_2d.html
```diff
@@ -70,7 +70,6 @@
                     name: 'OpenStreetMap',
                     url: 'http://www.openstreetmap.org/',
                 },
-                filter: function (layer) { return ['fill', 'line'].includes(layer.type) },
```
gchoqueux: could remove the `symbolToCircle` from example ?

zarov: done

gchoqueux (approved):

PR #1321, opened 2020-02-14
chore(test): move variables declaration in describe()

zarov:
In some unit tests, variables were declared before the first
`describe()` section of the file. When calling `npm run test-unit`, all
files are loaded and executed at the same time, with all `describe`
content being moved to a queue. So this causes a problem with things
declared before `describe`: a lot of variables (noticeably `context`)
were being rewritten and shared between all tests. While all tests were
passing, some of them would be false positive.
Moving all those declarations in `describe` fixes this.

### DIFF
--- a/CODING.md
+++ b/CODING.md
@@ -25,9 +25,12 @@ See the [node website](https://nodejs.org) to install node and npm.
 * Test and lint changes: see test, lint and test-examples npm script
 
 ## Debugging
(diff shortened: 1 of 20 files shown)


gchoqueux (approved):
good point, maybe it should be indicated in `coding.md`

gchoqueux (approved):

PR #1323, opened 2020-02-17
Refactor Helper

gchoqueux:
## Description
* `OBBHelper` extended by `Three.BoxHelper`;
* fix Sse Helper;


zarov (approved):
LGTM, but I don't understand why the CSS part is necessary.

PR #1324, opened 2020-02-20
doc(style): add default values and proper documentation of Style properties

zarov:
Needs #1325 merged before being valid

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -319,7 +319,7 @@ class Style {
             this.text.rotation = readVectorProperty(layer.layout['text-rotation-alignment'], zoom) || this.text.rotation;
 
             // content
-            this.text.field = readVectorProperty(layer.layout['text-field'], zoom) || this.text.field
+            this.text.field = readVectorProperty(layer.layout['text-field'], zoom) || this.text.field;
             this.text.wrap = readVectorProperty(layer.layout['text-max-width'], zoom) || this.text.wrap;
             this.text.spacing = readVectorProperty(layer.layout['text-letter-spacing'], zoom) || this.text.spacing;
             this.text.transform = readVectorProperty(layer.layout['text-transform'], zoom) || this.text.transform;
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -83,7 +83,6 @@ function getImageFromSprite(sprites, key) {
  * @property {string} point.line - The color of the border of the point. Can be
  * any [valid color
  * string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
- * Default is `red`.
  * @property {number} point.width - The width of the border, in pixel. Default
  * is `0.0` (no border).
  * @property {number} point.opacity - The opacity of the point. Can be between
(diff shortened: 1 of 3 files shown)


src/Core/Style.js
```diff
@@ -43,59 +43,180 @@ function getImageFromSprite(sprites, key) {
...
+            font: params.text.font || ['Open Sans Regular', 'Arial Unicode MS Regular', 'sans-serif'],
+            halo: {
+                color: (params.text.halo && params.text.halo.color) || 'black',
+                width: (params.text.halo && params.text.halo.width) || 0,
+                blur: (params.text.halo && params.text.halo.blur) || 0,
+            },
         };
     }
+
+    clone() {
```
gchoqueux, line 232: could add `copy`  and use it in `clone` function?

src/Core/Style.js
```diff
@@ -160,39 +281,39 @@ class Style {
             this.point.radius = readVectorProperty(layer.paint['circle-radius'], zoom) || 1.5;
         } else if (layer.type === 'symbol') {
             // overlapping order
-            this.text.zOrder = readVectorProperty(layer.layout && layer.layout['symbol-z-order'] || 'auto', zoom);
+            this.text.zOrder = readVectorProperty(layer.layout && layer.layout['symbol-z-order'], zoom) || 'auto';
```
gchoqueux: Why there's not `this.text.zOrder` default value in constructor?

src/Core/Style.js
```diff
@@ -160,39 +281,39 @@ class Style {
...
             }
 
             // position
-            this.text.anchor = readVectorProperty(layer.layout && layer.layout['text-anchor'] || 'center', zoom);
-            this.text.offset = readVectorProperty(layer.layout && layer.layout['text-offset'] || [0, 0], zoom);
-            this.text.padding = readVectorProperty(layer.layout && layer.layout['text-padding'] || 2, zoom);
-            this.text.size = readVectorProperty(layer.layout && layer.layout['text-size'] || 16, zoom);
-            this.text.placement = readVectorProperty(layer.layout && layer.layout['symbol-placement'] || 'point', zoom);
-            this.text.rotation = readVectorProperty(layer.layout && layer.layout['text-rotation-alignment'] || 'auto', zoom);
+            this.text.anchor = readVectorProperty(layer.layout && layer.layout['text-anchor'], zoom) || 'center';
```
gchoqueux: Could you use the default value from constructor, instead of `'center'`.
Same comment for all properties.

May be write a new function:
```js
setPropertyFromLayout(this.text, 'offset', layer.layout, zoom);
```

src/Core/Style.js
zarov: Forgotten, thanks !

src/Core/Style.js
zarov: Done

src/Core/Style.js
zarov: I changed them all to use the default value, otherwise no new function as it would be overkill: it checks `layer.layout` and `layer.paint` one time only

gchoqueux (approved):

PR #1328, opened 2020-02-20
parser: add style as a parameter when parsing

zarov:
It doesn't change much, but make my life easier for others things.


zarov:
I found a workaround, closing this PR

PR #1335, opened 2020-03-02
fix(VectorTile): Differentiated style caches with the VectorTileSources.

### LINKED ISSUE #1334: Tiles of vectorTile Layers not correctly reloaded
# Your Environment
Ubuntu 18
Chrome 80

## Context
On our map interface, buttons make user able to change the style of an already added vectorTileLayer.

Unfortunately, for the moment, the only way to do it is to remove the layer A with style X from the map, and to add again this layer A, but with the style Y.

The problem is that the already loaded tiles of the layer A with the original style X are not cleaned and are still displayed. 
Only tiles not loaded yet will be displayed with the selected Y style.

Cache problem ? 

NB : the layerID is the same, only the style changes.

1. add a vectorTileLayer (layerId = A, style = X)
2. remove layer with layerId = A
3. add a vectorTileLayer (layerId = A, style = Y)

## Expected Behavior
The layer A is displayed with style Y only.

## Actual Behavior
The layer A is displayed with a mix of X style for the tiles already loaded before, and Y style for the new requested tiles.

Mix of "color" and "grey" styles : 
Capture d’écran de 2020-02-28 15-15-25

gchoqueux:
## Description
New `VectorTileSource` aren't correctly loaded after change vector tile style file.  Differentiated style caches with the `VectorTileSources`.

Fix #1334

### DIFF
--- a/src/Parser/VectorTileParser.js
+++ b/src/Parser/VectorTileParser.js
@@ -144,7 +144,7 @@ function readPBF(file, options) {
             const layers = layersSource.filter(l => l.filterExpression({ zoom: z }, vtFeature) && (!l.minzoom || l.minzoom <= z) && (!l.maxzoom || l.maxzoom >= z));
             let feature;
             for (const layer of layers) {
-                const tag = `${layer.id}_zoom_${z}`;
+                const tag = `${layer.sourceUid}_${layer.id}_zoom_${z}`;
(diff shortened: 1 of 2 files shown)


zarov (approved):
LGTM, it works if you create a new `VectorTileSource` each time, but we still need a better solution in the future.

PR #1348, opened 2020-03-13
refactor(geographic): unify and encapsulate crs.

gchoqueux:
## Description
* unify and encapsulate crs.
* crs string aren't hard coded.
* there are two types of crs : `EPSG:XXXX` and `TMS:XXXX`

### DIFF
--- a/src/Core/Geographic/Crs.js
+++ b/src/Core/Geographic/Crs.js
@@ -2,40 +2,20 @@ import proj4 from 'proj4';
 
 proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
 
-const TMS = [
-    'WMTS:WGS84',
-    'WMTS:PM',
-];
+function isTms(crs) {
+    return crs.startsWith('TMS');
+}
 
-const EPSG = [
-    'EPSG:4326',
-    'EPSG:3857',
-];
+function isEpsg(crs) {
+    return crs.startsWith('EPSG');
+}
 
 function formatToTms(crs) {
-    if (crs) {
-        if (crs.includes('WMTS')) {
-            return crs;
-        }
-        const i = EPSG.indexOf(crs);
-        if (i > -1) {
-            return TMS[i];
-        } else if (crs.includes('EPSG')) {
-            return `WMTS:TMS:${crs.replace('EPSG:', '')}`;
-        }
-    }
+    return isTms(crs) ? crs : `TMS:${crs.match(/\d+/)[0]}`;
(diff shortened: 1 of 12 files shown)


src/Core/Geographic/Crs.js
```diff
@@ -2,38 +2,43 @@ import proj4 from 'proj4';
...
+const root_tms = 'TMS';
+const root_epsg = 'EPSG';
+const nCrs = ['4326', '3857'];
 
-const EPSG = [
-    'EPSG:4326',
-    'EPSG:3857',
-];
+const TMS = nCrs.map(n => `${root_tms}:${n}`);
+const EPSG = nCrs.map(n => `${root_epsg}:${n}`);
```
zarov: Too complicated imho, I understand what you are aiming for, but I see no problem in using already completed array. Same goes for `root_`, use directly the string: it won't change any time soon, so no worry about maintainability.

src/Core/Geographic/Crs.js
```diff
@@ -2,38 +2,43 @@ import proj4 from 'proj4';
...
+
+function isTms(crs) {
+    return crs.includes(root_tms);
+}
+
+function isEpsg(crs) {
+    return crs.includes('EPSG');
```
zarov: See, this is not coherent at all, following your initial idea, you should use `root_epsg`.

src/Core/Geographic/Crs.js
```diff
@@ -162,10 +167,14 @@ export default {
...
-     * format crs to tile matrix set notation : WMTS:XXXX.
+     * format crs to tile matrix set notation : TMS:XXXX.
      *
      * @param      {string}  crs     The crs to format
      * @return     {string}  formated crs
      */
     formatToTms,
+    isTms,
+    isEpsg,
+    tms_3857: formatToTms('EPSG:3857'),
```
zarov: Same here, why don't you use `EPSG[1]` ? Even more, you should use `TMS[1]`.

zarov (changes requested):
I feel like the modification done in `Crs.js` adds a lot of confusion on the different operations. Please fix that by simply using strings like `EPSG:4326` or `TMS:3857`, not complicated structure like you are proposing.

src/Core/Geographic/Crs.js
gchoqueux: ok, I removed `root_`

gchoqueux:
I simplified the code too

zarov (approved):
It is way better, thanks for this !

PR #1351, opened 2020-03-27
feat(renderer): emit a (in)visible event for LayeredMaterial

zarov:
Knowing the visibility of a `LayeredMaterial` is useful. For example in a
`GlobeView`, if you zoom in, "parent" tiles seems hidden; in fact, there
are not, it is only their material (so `LayeredMaterial`) that is set to
not visible.

Adding an event when changing this property can be useful to hide others
things, like in `TileDebug`, or in later PR to come (#1303 for example).

### DIFF
--- a/src/Renderer/LayeredMaterial.js
+++ b/src/Renderer/LayeredMaterial.js
@@ -166,6 +166,18 @@ class LayeredMaterial extends THREE.RawShaderMaterial {
         this.uniforms.colorTextures = new THREE.Uniform(new Array(nbSamplers[1]).fill(null));
         this.uniforms.colorOffsetScales = new THREE.Uniform(new Array(nbSamplers[1]).fill(identityOffsetScale));
         this.uniforms.colorTextureCount = new THREE.Uniform(0);
+
+        let _visible = this.visible;
+        // can't do an ES6 setter/getter here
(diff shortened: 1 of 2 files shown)


src/Renderer/LayeredMaterial.js
```diff
@@ -166,6 +166,20 @@ class LayeredMaterial extends THREE.RawShaderMaterial {
...
+            get() { return _visible; },
+            set(v) {
+                _visible = v;
+                if (!v) {
+                    this.dispatchEvent({ type: 'invisible' });
+                } else {
+                    this.dispatchEvent({ type: 'visible' });
+                }
+            },
+        });
```
gchoqueux: To handle the changing
```suggestion
        let _visible = this.visible;
        // can't do an ES6 setter/getter here
        Object.defineProperty(this, 'visible', {
            get() { return _visible; },
            set(v) {
                if (_visible != v) {
                   _visible = v;
                    this.dispatchEvent({ type: v ? 'showed' : 'hidden' });
                }
            },
        });
```

zarov: Yes, done, thanks for the suggestion !

gchoqueux (approved):

PR #1358, opened 2020-04-07
fix(VectorTileParser): wrong layers filtering by zoom.

gchoqueux:
## Description
fix wrong layers filtering by maxZoom in vector tile parsing.


zarov (approved):
Thanks for noticing this

PR #1360, opened 2020-04-09
refactor(source): add C3DtilesSource.

gchoqueux:
Add C3DtilesSource.

BREAKING CHANGES:
To instance C3DtilesLayer, it's needed to instance C3DtilesSource too.

### DIFF
--- a/test/unit/1__start.js
+++ b/test/unit/1__start.js
@@ -1,0 +1,5 @@
+import Renderer from './mock';
+
+// to add global.document
+const renderer = new Renderer();
+renderer.render();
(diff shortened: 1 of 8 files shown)


test/unit/1__start.js
```diff
@@ -0,0 +1,5 @@
+import Renderer from './mock';
+
+// to add global.document
```
zarov, line 3: Will this make my problem in #1303 away ?

zarov (approved):

test/unit/1__start.js
gchoqueux: yes

PR #1362, opened 2020-04-09
refactor(Layer): Robust layer#whenReady mechanisms.

gchoqueux:
## Description
Move All when ready in source

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -203,6 +203,7 @@ class View extends THREE.EventDispatcher {
         if (parentLayer) {
             if (layer.isColorLayer) {
                 const layerColors = this.getLayers(l => l.isColorLayer);
+                layer.sequence = layerColors.length;
 
                 const sumColorLayers = parentLayer.countColorLayersTextures(...layerColors, layer);
 
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -70,11 +70,6 @@ function _preprocessLayer(view, layer, parentLayer) {
         layer.projection = parentLayer.extent.crs;
     }
 
-    layer.whenReady = layer.whenReady.then(() => {
-        layer.ready = true;
-        return layer;
-    });
-
     return layer;
 }
 const _eventCoords = new THREE.Vector2();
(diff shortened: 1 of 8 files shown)


src/Layer/Layer.js
```diff
@@ -70,9 +71,32 @@ class Layer extends THREE.EventDispatcher {
 
         this.info = new InfoLayer(this);
 
-        this.whenReady = config.source && config.source.whenReady ? config.source.whenReady : Promise.resolve();
+        this.source = this.source || new Source({ url: 'none' });
```
zarov, line 84: Could you add documentation for `this.source`, `this.whenReady` and `this.ready` ?

src/Core/View.js
```diff
@@ -195,65 +190,60 @@ class View extends THREE.EventDispatcher {
...
-            });
+            this._layers.push(layer);
+        }
+
+        if (layer.object3d && !layer.object3d.parent && layer.object3d !== this.scene) {
+            this.scene.add(layer.object3d);
+        }
 
+        Promise.all(layer._promises).then(() => {
+            layer._resolve(layer);
```
zarov: You don't need to pass `layer` as an argument, `_resolve` don't use any.

src/Layer/PotreeLayer.js
```diff
@@ -136,8 +138,7 @@ class PotreeLayer extends GeometryLayer {
             this.root.bbox.max.set(cloud.boundingBox.ux, cloud.boundingBox.uy, cloud.boundingBox.uz);
 
             this.extent = Extent.fromBox3(view.referenceCrs, this.root.bbox);
-
-            return this.root.loadOctree().then(() => this);
+            return this.root.loadOctree().then(resolve, reject);
```
zarov: I am not sure that `reject` is handled by `loadOctree` (in fact, `Fetcher` doesn't handle `reject` correctly). So unless `Fetcher` handles rejection, you can remove the `reject` bit here.

src/Layer/OrientedImageLayer.js
```diff
@@ -147,14 +149,15 @@ class OrientedImageLayer extends GeometryLayer {
                 pano.id = pano.geometry[0].properties.id;
                 pano.index = i++;
             }
-        }).then(() => {
+        }, () => reject(new Error('oriented image layer: failed parse orientation'))).then(() => {
```
zarov: The rejection doesn't occur on the failed parsing of the orientation configuration, but rather on the `source.whenReady` giving a rejection. This rejection should be the next one, rather than a "failed parse calibration", which should be removed as well.

src/Layer/TiledGeometryLayer.js
```diff
@@ -68,13 +68,12 @@ class TiledGeometryLayer extends GeometryLayer {
...
+
+        this._promises.push(Promise.all(promises).then((level0s) => {
             this.level0Nodes = level0s;
-            for (const level0 of level0s) {
-                this.object3d.add(level0);
-                level0.updateMatrixWorld();
-            }
-        });
+            this.object3d.add(...level0s);
+            this.object3d.updateMatrixWorld();
```
zarov, line 75: 👍

src/Layer/C3DTilesLayer.js
```diff
@@ -212,12 +212,15 @@ class C3DTilesLayer extends GeometryLayer {
...
+
+        const { resolve, reject } = this._addPromise();
+
+        this.source.whenReady.then((tileset) => {
             this.tileset = tileset;
             this.tileIndex = new $3dTilesIndex(tileset, this.source.baseUrl);
             this.asset = tileset.asset;
             // TODO: Move all init3dTilesLayer code to constructor
-            return init3dTilesLayer(view, view.mainLoop.scheduler, this, tileset.root);
+            init3dTilesLayer(view, view.mainLoop.scheduler, this, tileset.root).then(resolve, reject);
```
zarov: As you can guess, `reject` can never occur here too. See other comments.

zarov (commented):
I understand where you are going, and this work is appreciated. Thanks ! I am just not a fan of the weird `_promises`, `_addPromise`, `_resolve` and `_reject` part. It seems complicated to follow where promises are resolved and all.

src/Core/View.js
gchoqueux: the whenReady promise needs to return the layer.

zarov: Yes, but it is already done in `Layer`, here the `layer` argument is only given to the function declared in `Layer`.

src/Layer/PotreeLayer.js
gchoqueux: I check, the error is handle to whenReady promise.

src/Core/View.js
gchoqueux: Yes, good catch

src/Layer/PotreeLayer.js
zarov: Disregard what I said, as `fetch()` returns only a resolving `Promise`. But it still confirms what I'm saying, if you go back through `loadOctree` and then `PotreeSource#whenReady`, you'll see no place with a rejecting `Promise`. So no point of having a rejection handling here I think.

src/Layer/OrientedImageLayer.js
gchoqueux: You are right. We'll have to do a PR to handle the errors properly.

src/Layer/C3DTilesLayer.js
gchoqueux: I remove it

src/Layer/PotreeLayer.js
gchoqueux: yes, i remove it.

gchoqueux:
> I understand where you are going, and this work is appreciated. Thanks ! I am just not a fan of the weird _promises, _addPromise, _resolve and _reject part. It seems complicated to follow where promises are resolved and all.

I go to think to another more simple way.

zarov (approved):

PR #1364, opened 2020-04-15
fix(view): allow only `ColorLayer` to be picked by `pickFeaturesAt`

zarov:
`pickObjectsAt` should be used to pick `GeometryLayer`.

### DIFF
--- a/examples/js/plugins/FeatureToolTip.js
+++ b/examples/js/plugins/FeatureToolTip.js
@@ -39,7 +39,7 @@ var FeatureToolTip = (function _() {
         tooltip.innerHTML = '';
         tooltip.style.display = 'none';
 
-        var features = view.pickFeaturesAt(event, 3, layersId);
+        var features = view.pickFeaturesAt.apply(view, [event, 3].concat(layersId));
 
         var layer;
         for (var layerId in features) {
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -537,19 +537,23 @@ class View extends THREE.EventDispatcher {
     }
 
     /**
-     * Return objects from some layers/objects3d under the mouse in this view.
+     * Searches for objects in {@link GeometryLayer} and specified
+     * `THREE.Object3D`, under the mouse or at a specified coordinates, in this
+     * view.
      *
-     * @param {Object} mouseOrEvt - mouse position in window coordinates (0, 0 = top-left)
-     * or MouseEvent or TouchEvent
-     * @param {number} radius - picking will happen in a circle centered on mouseOrEvt. Radius
-     * is the radius of this circle, in pixels
-     * @param {...*} where - where to look for objects. Can be either: empty (= look
-     * in all layers with type == 'geometry'), layer ids or layers or a mix of all
-     * the above.
-     * @return {Array} - an array of objects. Each element contains at least an object
-     * property which is the Object3D under the cursor. Then depending on the queried
-     * layer/source, there may be additionnal properties (coming from THREE.Raycaster
-     * for instance).
+     * @param {Object} mouseOrEvt - Mouse position in window coordinates (from
+     * the top left corner of the window) or `MouseEvent` or `TouchEvent`.
+     * @param {number} [radius=0] - The picking will happen in a circle centered
+     * on mouseOrEvt. This is the radius of this circle, in pixels.
+     * @param {...GeometryLayer|string|Object3D} [where] - Where to look for
+     * objects. It can be anything of {@link GeometryLayer}, IDs of layers, or
+     * `THREE.Object3D`. If no location is specified, it will query on all
+     * {@link GeometryLayer} present in this `View`.
+     *
+     * @return {Object[]} - An array of objects. Each element contains at least
+     * an object property which is the `THREE.Object3D` under the cursor. Then
+     * depending on the queried layer/source, there may be additionnal
+     * properties (coming from `THREE.Raycaster` for instance).
      *
      * @example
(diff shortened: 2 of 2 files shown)


gchoqueux:
So, Could you rename `pickFeaturesAt` to `pick2dFeatureAt` and add `pickFeaturesAt` to switch to `pick2dFeatureAt` or `pickObjectAt` in function layer type?

zarov:
I updated it quite a bit, now `View#pickFeaturesAt` calls `View#pickObjectsAt` to get the missing objects, and the signature of it has changed from `pickFeaturesAt(Object, number, (string|Layer)[])` to `pickFeaturesAt(Object, number, ...(string|Layer))`.

src/Core/View.js
```diff
@@ -547,53 +547,60 @@ class View extends THREE.EventDispatcher {
...
      * @example
      * view.pickObjectsAt({ x, y })
      * view.pickObjectsAt({ x, y }, 1, 'wfsBuilding')
      * view.pickObjectsAt({ x, y }, 3, 'wfsBuilding', myLayer)
      */
     pickObjectsAt(mouseOrEvt, radius = 0, ...where) {
+        const sources = [];
+
+        where = where.length == 0 ? this.getLayers(l => l.isGeometryLayer) : where;
+        where.forEach((l) => {
```
gchoqueux, line 567: `where` is always a `Array`?

src/Core/View.js
```diff
@@ -670,35 +677,65 @@ class View extends THREE.EventDispatcher {
...
      * @return {Object} - An object, with a property per layer. For example,
      * looking for features on layers `wfsBuilding` and `wfsRoads` will give an
      * object like `{ wfsBuilding: [...], wfsRoads: [] }`. Each property is made
      * of an array, that can be empty or filled with found features.
      *
      * @example
-     * view.pickFeaturesAt({ x, y });
-     * view.pickFeaturesAt({ x, y }, 1, ['wfsBuilding']);
-     * view.pickFeaturesAt({ x, y }, 3, ['wfsBuilding', myLayer]);
+     * view.pickFeaturesOnColorLayersAt({ x, y });
```
gchoqueux: `pickFeaturesOnColorLayersAt` => update the documentation.

src/Core/View.js
```diff
@@ -670,35 +677,65 @@ class View extends THREE.EventDispatcher {
...
+    pickFeaturesAt(mouseOrEvt, radius = 3, ...where) {
+        if (Array.isArray(where[0])) {
+            console.warn('Deprecated: the ...where argument of View#pickFeaturesAt should not be an array anymore, but a list: use the spread operator if needed.');
+            where = where[0];
+        }
+
+        const layers = [];
+
+        where = where.length == 0 ? this.getLayers(l => l.isColorLayer) : where;
+        where.forEach((l) => {
```
gchoqueux, line 701: `where` is Array ? (same comment)

src/Core/View.js
```diff
@@ -670,35 +677,65 @@ class View extends THREE.EventDispatcher {
...
+            }
         });
 
         // Get the mouse coordinates to the correct system
         const mouse = (mouseOrEvt instanceof Event) ? this.eventToViewCoords(mouseOrEvt, _eventCoords) : mouseOrEvt;
+        const objects = this.pickObjectsAt(mouse, radius, ...where);
+
+        const result = {};
+
+        if (objects.length > 0) {
```
gchoqueux, line 716: We could be return the data structure to `pickObjectsAt` to avoid this re-structuration ?
And unify the returns

src/Core/View.js
zarov: Yes, it was an error, thanks !

src/Core/View.js
zarov: With the `...` operator, yes !

src/Core/View.js
zarov: Yes, and I'm updating the examples I missed !

src/Core/View.js
zarov: Yes we could, but `pickObjectsAt` doesn't pick only `GeometryLayers`, but `Object3D` too, so we can't really have the same data structure, because we lack layer ids. If it's okay to use the uuid from the objects though, it's okay with me.

src/Core/View.js
gchoqueux: ok

src/Core/View.js
gchoqueux: or `withoutLayer`...?

examples/js/plugins/FeatureToolTip.js
```diff
@@ -39,7 +39,7 @@ var FeatureToolTip = (function _() {
         tooltip.innerHTML = '';
         tooltip.style.display = 'none';
 
-        var features = view.pickFeaturesAt(event, 3, layersId);
+        var features = view.pickFeaturesAt.apply(view, [event, 3].concat(layersId));
```
gchoqueux, line 42: Why you use `apply` to call `pickFeaturesAt`?

zarov: Because of the `...` operator, that we don't allow in plugins (we allow only ES5 syntax). See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Replace_apply

gchoqueux: ah, yes!

gchoqueux (approved):

PR #1365, opened 2020-04-16
Choose the smart crs to build feature extent.

gchoqueux:
## Description
Some Features are building with 'EPSG:4978'  and the extent in the same.
Extent  in 'EPSG:4978'  doesn't really make sense.
So the feature chooses better crs for his extent.

### DIFF
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -1,5 +1,6 @@
 import * as THREE from 'three';
 import Extent from 'Core/Geographic/Extent';
+import Coordinates from 'Core/Geographic/Coordinates';
 
 function defaultExtent(crs) {
     return new Extent(crs, Infinity, -Infinity, Infinity, -Infinity);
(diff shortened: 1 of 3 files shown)


src/Core/Feature.js
```diff
@@ -185,7 +188,10 @@ class Feature {
         this.normals = options.withNormal ? [] : undefined;
         this.crs = crs;
         this.size = options.withAltitude ? 3 : 2;
-        this.extent = options.buildExtent ? defaultExtent(crs) : undefined;
+        if (options.buildExtent) {
+            this.extent = options.buildExtent ? defaultExtent(crs != 'EPSG:4978' ?  crs : options.crsIn) : undefined;
```
zarov: You missed something here: `this.extent` will never be `undefined` in this case. 

I also don't understand the `crs` condition: could you explain a bit ?

gchoqueux: I change the code and add comment.

zarov (approved):
Thanks !

PR #1369, opened 2020-04-17
refactor(FileSource): pre-Parse file in constructor Layer.

gchoqueux:
## Description
* Fetch/Parse `FileSource` data when is added.
  * Fetch in `FileSource`
  * Parse fetched data in `Layer`: inline or by callBack? 
       ➡️ **I choose inline parsing.**

## Motivation and Context
* Build `FileSource#extent` on Layer adding.
* Remove specific test for `FileSource` in ProcessingFeature.
* Avoid multi fetch/parse/convert in ProcessingFeature for `FileSource`.

### DIFF
--- a/src/Layer/Layer.js
+++ b/src/Layer/Layer.js
@@ -2,6 +2,7 @@ import * as THREE from 'three';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from 'Layer/LayerUpdateStrategy';
 import InfoLayer from 'Layer/InfoLayer';
 import Source from 'Source/Source';
+import { parseSourceData } from 'Provider/DataSourceProvider';
 
 /**
  * @property {boolean} isLayer - Used to checkout whether this layer is a Layer.
@@ -90,6 +91,10 @@ class Layer extends THREE.EventDispatcher {
         this.whenReady = new Promise((re, rj) => {
             this._resolve = re;
             this._reject = rj;
+        }).then(() => {
+            if (this.source.fetchedData && !this.source.parsedData) {
+                return parseSourceData(this.source.fetchedData, this.source.extent, this);
+            }
         }).then(() => {
             this.ready = true;
             return this;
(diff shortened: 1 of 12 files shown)


src/Layer/Layer.js
```diff
@@ -2,6 +2,7 @@ import * as THREE from 'three';
 import { STRATEGY_MIN_NETWORK_TRAFFIC } from 'Layer/LayerUpdateStrategy';
 import InfoLayer from 'Layer/InfoLayer';
 import Source from 'Source/Source';
+import { parseSourceData } from 'Provider/DataSourceProvider';
```
zarov, line 5: Do you plan on moving it in this class ?

test/unit/dataSourceProvider.js
```diff
@@ -26,9 +25,13 @@ const holes = require('../data/geojson/holesPoints.geojson.json');
 describe('Provide in Sources', function () {
     const renderer = new Renderer();
     renderer.setClearColor();
-
-    supportedFetchers.set('image/png', () => Promise.resolve(new THREE.Texture()));
-    supportedFetchers.set('application/json', () => Promise.resolve(holes));
+    // /!\ Avoid to overload fetcher because could troubleshoot the other unit tests?
+    // formatTag to avoid it
```
zarov, line 29: Good idea !

test/unit/source.js
```diff
@@ -198,17 +198,18 @@ describe('Sources', function () {
         it('should instance and use FileSource with url', function () {
             const source = new FileSource({
                 url: '..',
+                parsedData: {},
```
zarov: If you add this, you are doing the "should instance and use FileSource with parsedData" test below.

zarov (commented):
Looks good, a few things to change

src/Layer/Layer.js
gchoqueux: I don't think so because maybe I move the responsibility to the future NodeLayer.

test/unit/source.js
gchoqueux: I will change it to fetch a reel file.

src/Layer/Layer.js
zarov: Ok good to me

test/unit/source.js
zarov: Ok

zarov: Ok

gchoqueux: `done()`

zarov (approved):
Looks good, thanks !

PR #1371, opened 2020-04-21
fix(Core): unnecessary testing.

gchoqueux:
## Description
From DeepScan report


zarov (approved):
Thanks ! Could it be an option to run deepscan support when creating PR ?

gchoqueux:
> Could it be an option to run deepscan support when creating PR ?

I will check if there is option.

PR #1373, opened 2020-04-22
refactor(test): add required file to unit tests

gchoqueux:
* Use required option mocha with `bootstrap.js` file to initialize unit tests instead of use first file list.
* add unit tests.
* deprecated `placeObjectOnGround`.
* move mock to bootstrap


package.json
```diff
@@ -11,7 +11,7 @@
     "test-unit": "npm run base-test-unit test/unit",
     "test-functional": "mocha -t 60000 --recursive test/functional",
     "test-with-coverage": "nyc -n src -r html cross-env npm run test-unit",
-    "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --require @babel/register",
+    "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --require @babel/register test/unit/bootstrap.js",
```
zarov: Are you sure it is working correctly ? From what I recall, the `--require` part tries to load iTowns before the next file. So in case iTowns loads first, it will emit an error.

package.json
```diff
@@ -23,6 +23,7 @@
   "nyc": {
     "exclude": [
       "**/*ThreeExtended",
+      "**/*src/Utils/placeObjectOnGround.js",
```
zarov, line 26: Why would you exclude it ?

src/Utils/DEMUtils.js
```diff
@@ -416,7 +283,7 @@ function offsetInExtent(point, extent, target = new THREE.Vector2()) {
     return target.set(originX, originY);
 }
 
-function _readZ(layer, method, coord, nodes, cache) {
+export function _readZ(layer, method, coord, nodes, cache) {
```
zarov: If it is being exported, you could even merge it with `getTerrainObjectAt`, because it is almost the same return value (`coord` instead of `coord.z`).

zarov (commented):
Don't forget to add a `BREAKING CHANGE` for the deprecation of `placeObjectOnGround`. Maybe the documentation of `readZ/getTerrainObjectAt` could be updated to provide a simple `@example` with how to get the coordinate and the tile to place an object on the ground ?

package.json
gchoqueux: yes, it's better to use `--file`  : Explicitly include a test file to be loaded before other test files
https://mochajs.org/#-file-filedirectoryglob

package.json
gchoqueux: Because I deprecated `placeObjectOnGround` until I will test it or remove it.

src/Utils/DEMUtils.js
gchoqueux: yes

gchoqueux:
> Don't forget to add a `BREAKING CHANGE` for the deprecation of `placeObjectOnGround`.

added

> Maybe the documentation of `readZ/getTerrainObjectAt` could be updated to provide a simple 
`@example` with how to get the coordinate and the tile to place an object on the ground ?

done

package.json
zarov: Yes, but what I meant is the `--require` is loaded before any other things: "Require a module before loading the user interface or **test files**."
https://mochajs.org/#-require-module-r-module

package.json
zarov: Ok

test/unit/style.js
```diff
@@ -0,0 +1,36 @@
...
+        assert.equal(style1.stroke.color, style2.stroke.color);
+    });
+    it('Clone style', () => {
+        const style2 = style1.clone(style1);
+        assert.equal(style1.point.color, style2.point.color);
+        assert.equal(style1.fill.color, style2.fill.color);
+        assert.equal(style1.stroke.color, style2.stroke.color);
+    });
+    it('applyToHTML', () => {
+        const dom = { style: {} };
```
zarov: You can use `document.createElement` which is present in `mock`.

zarov (commented):
What about moving `mock` in to bootstrap ? It would simplify everything.

gchoqueux:
> What about moving mock in to bootstrap ? It would simplify everything.

Yes I was thinking of doing that, too.

test/unit/style.js
gchoqueux: yes, thanks!

gchoqueux:
> Yes I was thinking of doing that, too.

done

gchoqueux:
@zarov 
i added more unit tests.

package.json
gchoqueux: `@babel/register` is required to compile unit tests.

test/unit/3dtileslayerprocessbatchtable.js
```diff
@@ -0,0 +1,44 @@
...
+        heading: -145,
+    };
+    const viewer = new GlobeView(renderer.domElement, p, { renderer, noControls: true });
+
+
+    const threedTilesLayerBTHierarchy = new C3DTilesLayer(
+        '3d-tiles-bt-hierarchy', {
+            name: 'BTHierarchy',
+            source: new C3DTilesSource({
+                url: 'https://raw.githubusercontent.com/AnalyticalGraphicsInc/cesium/master/Apps/SampleData/Cesium3DTiles/Hierarchy/BatchTableHierarchy/tileset.json',
```
zarov, line 27: Couldn't we store file like this on the repository ? That way we are sure we always have access to it. I know this isn't the only place we are doing this, so I'm more asking as a general question rather than for this PR in particular.

test/unit/demutils.js
```diff
@@ -0,0 +1,93 @@
...
+
+describe('DemUtils', function () {
+    const renderer = new Renderer();
+    const placement = { coord: new Coordinates('EPSG:4326', 1.5, 43), zoom: 10 };
+    const viewer = new GlobeView(renderer.domElement, placement, { renderer });
+
+    const source = new WMTSSource({
+        format: 'image/x-bil;bits=32',
+        projection: 'EPSG:4326',
+        url: 'https://wxs.ign.fr/3ht7xcw6f7nciopo16etuqp2/geoportail/wmts',
```
zarov, line 25: Same, couldn't we have a local file to load ? That way we don't have to search for hours new values if there is any change.

zarov (approved):
Thanks for all those tests, and the +6.2% bump in coverage !

PR #1377, opened 2020-04-28
feat(style): add partial support for stops in vector tiles

zarov:
If there are `stops` in any property of the Style, they will be taken
into account to override `minzoom` and `maxzoom` to reflect those
`stops`.

Note that this is a partial support, as the Mapbox Style specs have evolved:
https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -167,6 +167,9 @@ class Style {
     constructor(params = {}) {
         this.isStyle = true;
 
+        this.minzoom = 0;
+        this.maxzoom = 24;
+
         params.fill = params.fill || {};
         params.stroke = params.stroke || {};
         params.point = params.point || {};
@@ -271,14 +274,15 @@ class Style {
     /**
      * set Style from vector tile layer properties.
      * @param {object} layer vector tile layer.
-     * @param {Number} zoom vector tile layer.
      * @param {Object} sprites vector tile layer.
      * @param {boolean} [symbolToCircle=false]
      * @returns {Style}
      */
-    setFromVectorTileLayer(layer, zoom, sprites, symbolToCircle = false) {
+    setFromVectorTileLayer(layer, sprites, symbolToCircle = false) {
         layer.layout = layer.layout || {};
         layer.paint = layer.paint || {};
+
+        const zoom = this.minzoom;
 
         if (layer.type === 'fill' && !this.fill.color) {
             const { color, opacity } = rgba2rgb(readVectorProperty(layer.paint['fill-color'] || layer.paint['fill-pattern']));
--- a/src/Parser/VectorTileParser.js
+++ b/src/Parser/VectorTileParser.js
@@ -3,15 +3,11 @@ import Protobuf from 'pbf';
 import { VectorTile } from '@mapbox/vector-tile';
 import { globalExtentTMS } from 'Core/Geographic/Extent';
 import { FeatureCollection, FEATURE_TYPES } from 'Core/Feature';
-import { featureFilter } from '@mapbox/mapbox-gl-style-spec';
-import Style from 'Core/Style';
 
 const worldDimension3857 = globalExtentTMS.get('EPSG:3857').dimensions();
 const globalExtent = new Vector3(worldDimension3857.x, worldDimension3857.y, 1);
 const lastPoint = new Vector2();
 const firstPoint = new Vector2();
-const styleCache = new Map();
-
 
 // Classify option, it allows to classify a full polygon and its holes.
 // Each polygon with its holes are in one FeatureGeometry.
@@ -94,7 +90,6 @@ function vtFeatureToFeatureGeometry(vtFeature, feature, classify = false) {
     feature.updateExtent(geometry);
 }
 
-const defaultFilter = () => true;
 function readPBF(file, options) {
     const vectorTile = new VectorTile(new Protobuf(file));
     const sourceLayers = Object.keys(vectorTile.layers);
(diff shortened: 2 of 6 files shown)


gchoqueux:
There's yet a little `mapbox Expression` support in [VectorTileParser]( https://github.com/iTowns/itowns/blob/master/src/Parser/VectorTileParser.js#L133) and [`expression/maxZoom/minZoom`](https://github.com/iTowns/itowns/blob/master/src/Parser/VectorTileParser.js#L143)

This already existing support isn't the best, I'm trying to find out if there is any overlap with what you're adding?

zarov:
This isn't the same thing, as my changes create more specific `Style`, without any `stops`. This is to simplify things for object that use those `Style`, without having to read everytime the `stops` array. No filtering is done here.

gchoqueux:
I don't have a clear view yet, but here are some questions/proposals :

* In `readVectorProperty` method in `Style`, there is some  `stops` support. There are no duplicates? May be move this in Style? and add best method instead of `readVectorProperty`?

* In your user case, there's mapbox expression in the `layer.layout` or `layer.paint`? Could you use mapbox `Expression` with `@mapbox/mapbox-gl-style-spec` package?

* comment outside of this PR, this expression filtering needs to move on `VectorTileSource`:
 https://github.com/iTowns/itowns/blob/master/src/Parser/VectorTileParser.js#L131

zarov:
> In `readVectorProperty` method in `Style`, there is some  `stops` support. There are no duplicates? May be move this in Style? and add best method instead of `readVectorProperty`?

This isn't really the same things: what I'm adding is only to determine `style.text.maxzoom`. However, it could be an idea to add global properties `minzoom` and `maxzoom` to a `Style`, to create a `Style` for each `stops` value (if any). For example, if `layout.color` has three stops, it doesn't create a single `Style`, but rather 3. Then, we could reference those style without having to create them in `VectorTileParser`.

> In your user case, there's mapbox expression in the `layer.layout` or `layer.paint`? Could you use mapbox `Expression` with `@mapbox/mapbox-gl-style-spec` package?

There is no Mapbox `Expression` in my user case. I agree it would be better to implement the support for `Expression`, but that seems too much for this PR. But we can add it to the roadmap !

Anyway, I'm looking into `VectorTileSource` and `Style` to clean it a little bit and try what I described above.

src/Core/Style.js
```diff
@@ -167,6 +167,9 @@ class Style {
     constructor(params = {}) {
         this.isStyle = true;
 
+        this.minzoom = 0;
+        this.minzoom = 24;
```
gchoqueux: ... `this.maxZoom`, no?

src/Parser/VectorTileParser.js
```diff
@@ -3,15 +3,11 @@ import Protobuf from 'pbf';
 import { VectorTile } from '@mapbox/vector-tile';
 import { globalExtentTMS } from 'Core/Geographic/Extent';
 import { FeatureCollection, FEATURE_TYPES } from 'Core/Feature';
-import { featureFilter } from '@mapbox/mapbox-gl-style-spec';
```
gchoqueux, line 6: Very good  `VectorTileParser.js` cleaning!!

src/Parser/VectorTileParser.js
```diff
@@ -117,66 +112,46 @@ function readPBF(file, options) {
...
-        allLayers.forEach((l) => {
-            l.filterExpression = featureFilter(l.filter);
-        });
-        features.filter.loaded = true;
-    }
+    collection.scale.set(size, -size, 1).divide(globalExtent);
+    collection.translation.set(-(vFeature.extent * x + center), -(vFeature.extent * y + center), 0).divide(collection.scale);
 
     sourceLayers.forEach((layer_id) => {
+        if (!options.filter[layer_id]) { return; }
```
gchoqueux: `filter`, in `options.filter`, the name was never very clear, it's time to change it?

src/Provider/DataSourceProvider.js
```diff
@@ -18,7 +18,6 @@ export function parseSourceData(data, extDest, layer) {
         buildExtent: source.isFileSource || !layer.isGeometryLayer,
         crsIn: source.projection,
         crsOut: layer.projection,
-        sprites: layer.sprites || source.sprites,
```
gchoqueux, line 21: very good deletions!

src/Source/VectorTilesSource.js
```diff
@@ -88,7 +57,74 @@ class VectorTilesSource extends TMSSource {
...
+                        this.styles[layer.id].push(style);
+                    }
+
+                    if (!this.filter[layer['source-layer']]) {
+                        this.filter[layer['source-layer']] = [];
+                    }
+
+                    this.filter[layer['source-layer']].push({
+                        id: layer.id,
+                        order: order++,
```
gchoqueux: Maybe sort by order `this.filter` to avoid the collections sorting [here](https://github.com/iTowns/itowns/pull/1377/files/fec9444760e63fae9230f1fbb8029e9f5082aad1..dac4cb2201384c48409b670d95bb366f7f52c937#diff-cc5b474688a4526a987e5193926b28a3R152)

src/Source/VectorTilesSource.js
```diff
@@ -88,7 +57,74 @@ class VectorTilesSource extends TMSSource {
...
+                        style.maxzoom = stops[i + 1];
+                        style.setFromVectorTileLayer(layer, this.sprites, this.symbolToCircle);
+                        this.styles[layer.id].push(style);
+                    }
+
+                    if (!this.filter[layer['source-layer']]) {
+                        this.filter[layer['source-layer']] = [];
+                    }
+
+                    this.filter[layer['source-layer']].push({
```
gchoqueux: Rename `this.filter` to `this.layers`?

src/Source/VectorTilesSource.js
```diff
@@ -88,7 +57,74 @@ class VectorTilesSource extends TMSSource {
...
+                    stops.sort((a, b) => (a - b));
+
+                    this.styles[layer.id] = [];
+                    for (let i = 0; i < stops.length - 1; i++) {
+                        if (stops[i] == stops[i + 1]) { continue; }
+                        const style = new Style();
+                        style.minzoom = stops[i];
+                        style.maxzoom = stops[i + 1];
+                        style.setFromVectorTileLayer(layer, this.sprites, this.symbolToCircle);
+                        this.styles[layer.id].push(style);
```
gchoqueux, line 103: I'm not convinced by the multi-styles by zoom level because is possible to have zoom and  properties to change the style, there would be a lot of combinations.

May be is better to have properties in function of zoom or properties. 
we should think about it.

src/Source/VectorTilesSource.js
gchoqueux: may be sort `the.filter` by `order` here to avoid sort Features in each vector tile parsing?

src/Source/VectorTilesSource.js
gchoqueux: In vector tile example, the layer `landcover_wood` 3 styles, two of which are identical.
There's an extra style?

gchoqueux (commented):
Good feature and refactoring
This seems like the right way but I'm not sure to instance a Style by `minzoom `or `maxzoom`. ([see comment](https://github.com/iTowns/itowns/pull/1377#discussion_r418185823))

src/Core/Style.js
zarov: Yes, changed thanks !

src/Source/VectorTilesSource.js
zarov: Good catch, I'll see what I can do

src/Source/VectorTilesSource.js
zarov: I could, but it impacts all the way to the `DataSourceProvider`: how could I pass it through `VectorTileParser` then ?

src/Source/VectorTilesSource.js
zarov: You mean if we want to update a layer style and that's impacting all three styles ?

Indeed, I see the duplicate style, but they are not following each other: I don't see what could be done here. If we are going with `one layer = one style`, regardless of zoom, we'll have to read the zoom value each time we are in `Feature2Texture` or something else. I think this is slowing things, and complicating them.

src/Source/VectorTilesSource.js
zarov: It seems that we can't escape a `sort`, as the loop is built around the received features, and looping around the whole `filter` wouldn't be a good idea too.

src/Source/VectorTilesSource.js
gchoqueux: Let's actually follow a simpler path for the moment. But configurable style is the future way.
(For `Feature2Texture`, the style can be set before and passed to the `createTextureFromFeature` as before).

On the other hand, unnecessary styles should be removed. For example, if there are 2 stops, there should only be 2 styles.

src/Source/VectorTilesSource.js
gchoqueux: Yes, the impact isn't very important.

src/Source/VectorTilesSource.js
gchoqueux: Could you sort `this.filter` after the end of its construction?

zarov: Impossible, it is an object, and I made it that way because it is much faster than an array to read (`obj[key]` vs `arr.indexOf(key)` for example). But as I said, we can't escape the `sort` as we don't know the order features in the `.pbf/.mvt` are.

gchoqueux: Ok, I didn't realize it was an object. we'll see later to avoid being sorted .

src/Source/VectorTilesSource.js
zarov: > On the other hand, unnecessary styles should be removed. For example, if there are 2 stops, there should only be 2 styles.

Yes, it shouldn't be the case afaik: the case you cited is as expected. If no `minzoom` is set, it defaults to `2`, thus creating a new `stop`. That's why you have the same style twice. But as they are not "touching", it is necessary to create the same style twice.

```js
fill-opacity: {
    base: 1, // used for zooms [2; 8[
    stops: [
        [8, 0.6], // 0.6 is used for zooms [8; 22[
        [22, 1] // 1 is used for zooms [22; 24[
    ]}
```

src/Source/VectorTilesSource.js
zarov: Ok done

src/Parser/VectorTileParser.js
zarov: Done !

zarov:
I updated it all, documentation and tests have been added, it only needs a clean on the commits if it is good !

zarov:
All clean and ready to be validated ;)

src/Provider/DataSourceProvider.js
```diff
@@ -28,7 +27,7 @@ export function parseSourceData(data, extDest, layer) {
         mergeFeatures: layer.mergeFeatures === undefined ? true : layer.mergeFeatures,
         withNormal: layer.isGeometryLayer !== undefined,
         withAltitude: layer.isGeometryLayer !== undefined,
-        symbolToCircle: layer.symbolToCircle || false,
+        layers: source.layers,
```
gchoqueux, line 30: You could remove option `filter: layer.filter || source.filter` is the previous option.

test/unit/bootstrap.js
```diff
@@ -71,15 +79,15 @@ global.document = {
             const img = new DOMElement();
             img.width = 10;
             img.height = 10;
+            Object.defineProperty(img, 'src', {
+                set: () => img.emitEvent('load'),
```
gchoqueux, line 83: why you add this emit?

src/Provider/DataSourceProvider.js
gchoqueux: you can verify but I thinks is ok

test/unit/bootstrap.js
zarov: To test the loading of an image, using `THREE.TextureLoader` in `Fetcher`.

src/Provider/DataSourceProvider.js
zarov: No, for example it can be use like this https://github.com/iTowns/itowns/blob/master/examples/source_stream_wfs_25d.html#L245

test/unit/bootstrap.js
gchoqueux: To test sprite loading?
I try to know because I remove all `emitEvent` in last PR on unit test ?
it complicates to handle

gchoqueux: Good, so it's remove the ambiguity.

zarov: Yes that's it ! Unfortunately I didn't see it in the last PR as `mock` was moved.

src/Provider/DataSourceProvider.js
gchoqueux: Good, so it's remove the ambiguity.

gchoqueux (approved):

PR #1378, opened 2020-04-28
feat(layer): add `getCurrentMaxZoom` property to InfoLayer

zarov:
This new method returns the value of maximum zoom of all tiles presents
on the screen.

### DIFF
--- a/src/Layer/InfoLayer.js
+++ b/src/Layer/InfoLayer.js
@@ -60,6 +60,16 @@ export class InfoTiledGeometryLayer extends InfoLayer {
             });
     }
 
+    get currentMaxTileZoom() {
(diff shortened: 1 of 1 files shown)


src/Layer/InfoLayer.js
```diff
@@ -60,6 +60,16 @@ export class InfoTiledGeometryLayer extends InfoLayer {
             });
     }
 
+    get currentMaxZoom() {
```
gchoqueux: could you rename `currentMaxTileZoom` ?

zarov: done

gchoqueux (approved):

Issue #1381, opened 2020-04-29
Circular dependency / global variable causing bug ?


jailln:
## Your Environment
* Version used: a branch of my fork based on the current master
* Browser Name and version: Firefox 75.0 and Chromium 81.0.4044.122
* Operating System and version (desktop or mobile): Ubunut 18.04
* Link to your project: [the branch in question](https://github.com/jailln/itowns/tree/3dtiles_refactoring-rebase)

## Context

While [rebasing one of my branch](https://github.com/jailln/itowns/tree/3dtiles_refactoring-rebase) I came across a bug that does not seem directly related to the code I introduced/modified. 

On this branch, all the examples of iTowns produce the following error (while I have not modified `Renderer/OBB.js` nor `Core/Prefab/Globe/BuilderEllipsoidTile.js`):

![image](https://user-images.githubusercontent.com/16967916/80577834-d881a280-8a07-11ea-984d-50c2ec26f828.png)

## Steps to Reproduce (for bugs)
1. git clone https://github.com/jailln/itowns.git
2. cd itowns
3. git checkout 3dtiles_refactoring-rebase
4. npm install
5. npm start
6. open any example in your browser

## Expected Behavior
Examples not related to 3DTiles should work since I only modified 3DTiles related code.

## Actual Behavior
All examples are broken.

## Possible Cause

By looking into the code I think this error is caused by a combination of:
- the circular dependency between [Renderer/OBB.js](https://github.com/iTowns/itowns/blob/master/src/Renderer/OBB.js) and [Core/Prefab/Globe/BuilderEllipsoidTile.js](https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/Globe/BuilderEllipsoidTile.js)
- the initialization of the global variable `builder` which uses BuilderEllipsoidTile's constructor.

However, I don't know why it works on the master branch and why it does not on my branch...

## Possible Fix/Solution

Initializing `builder` to `undefined` or moving its declaration into the `setFromExtent()` method (where it is used) solves this issue. However, since I am not familiar with this part of the code, I will let you decide what is the best solution to handle this issue. In addition, is there any reason for this circular dependency and for using global variables in `Renderer/OBB.js` ?

zarov:
A better solution, that seems to work on at least a few example: removing the circular dependency. I removed the `OBB` import in `BuilderEllipsoidTile`, as the only method on it serves no purpose imho. 

I submitted the fix in #1382

jailln:
[...]

PR #1382, opened 2020-04-29
refacto(core): remove BuilderEllipsoidTile#OBB

zarov:
This method serves no real purpose, and having it here creates a
circular dependency, which is not good.

Solves #1381

### DIFF
--- a/src/Renderer/OBB.js
+++ b/src/Renderer/OBB.js
@@ -10,6 +10,7 @@ const size = new THREE.Vector3();
 const dimension = new THREE.Vector2();
 const center = new THREE.Vector3();
 const coord = new Coordinates('EPSG:4326', 0, 0, 0);
+let obb;
 
 class OBB extends THREE.Object3D {
     /**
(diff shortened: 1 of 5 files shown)


gchoqueux (commented):
could modify `PlanarTileBuilder` too?

zarov:
Yes done !

src/Renderer/OBB.js
```diff
@@ -141,7 +141,7 @@ class OBB extends THREE.Object3D {
             };
 
             const geometry = new TileGeometry(paramsGeometry);
-            this.copy(builder.OBB(geometry.boundingBox));
+            this.copy(new OBB(geometry.boundingBox.min, geometry.boundingBox.max));
```
gchoqueux: I just noticed It's possible to remove the `OBB` instance:

```js
let obb;

//...
// this.copy(new OBB(geometry.boundingBox.min, geometry.boundingBox.max));
obb.natBox.copy(geometry.boundingBox);
this.copy(obb);
//...

obb = new OBB();
```

zarov: I added a `copyBox3`, it is even simpler imho. See my latest commit.

gchoqueux: you need to `this.copy(obb);` because the transformation is copied too.

zarov: I disagree, because at this point, `obb` should have the default value of `z`, so I don't understand how that is different from what I added ?

zarov: In fact no transformation at all has been done at this point, so there is nothing to copy except the box3 ?

gchoqueux: if this obb instance  have transformation, so it needs to copy identity transformation.

gchoqueux (approved):

Issue #1171, opened 2019-07-01
3D Tiles refactoring


jailln:
I started working on the refactoring of 3D Tiles [here](https://github.com/jailln/itowns/tree/3dtiles_refactoring). Current proposed modifications are: (some of these points have been discussed with @gchoqueux offline).

  * [ ] Create a class for each object [of the spec](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification/schema) and put them in src/Core/3DTiles/ . I started working on some of these classes [here](https://github.com/jailln/itowns/tree/3dtiles_refactoring/src/Core/3DTiles).
  * [x] Create a 3DTilesLayer which extend GeometryLayer.
  * [ ] Harmonize the tileIndex and object3d hierarchies (see [this issue](https://github.com/iTowns/itowns/issues/1039)) to have only one hierarchy of tiles; probably with a boolean attribute `loaded` and the content of the tile in a `content` attribute when it is loaded.
 * [x] Refactor the extensions management:
    * [x] Don't use a global object to register extension (namely [$3dTilesExtensions](https://github.com/iTowns/itowns/blob/master/src/Provider/3dTilesProvider.js#L53)). Use a property of the layer instead (currently with layer.defineLayerProperty, e.g. [here](https://github.com/jailln/itowns/blob/3dtiles_refactoring/examples/3dtiles_hierarchy.html#L69) but then it must become a property of the 3DTilesLayer class). 
    * [x] Register extensions as objects and not parsers (so they can have other methods than parsing).
 * [ ] Implement cache at iTowns level (see issue #1170)

Open questions:
  * Regarding the parsers, we can either:
   * Keep the BatchTableParser and create a FeatureTableParser to be consistent (and maybe regroup these files in a folder).
   * Remove BatchTableParser, create a TileContentParser which groups the common parsing code of the different 3D Tiles format (b3dm, pnts, etc.), such as the parsing of the headers, of the batch table and features tables, etc.

Do you have any inputs on these propositions ?

gchoqueux:
Hello, thank you for your contribution.

>  Create a class for each object of the spec and put them in src/Core/3DTiles/ . I started working on some of these classes here.

I think we need to get to the heart of the matter and open a PR on the new classes.

jailln:
Yep, I started this work [here](https://github.com/jailln/itowns/tree/3dtiles_refactoring). I'll open a PR next week.

gchoqueux:
[...]

Issue #1039, opened 2019-02-22
[3D Tiles] What is tileIndex ?


jailln:
## Context

I'm currently working on improving 3D Tiles extension management introduced in #1022 to give the possibility to have extensions defined in multiple 3D Tiles objects (i.e. not only in batch table as it is currently the case). 

To this end, I need to start refactoring the 3D Tiles Provider ([which I think is planned later this year in your roadmap](https://github.com/iTowns/itowns/issues/989#issuecomment-453536509)). I'll come up with propositions in the next days maybe as an 'in progress' PR. I still have questions about the provider and the structure resulting from the parsing of the tileset and its tiles though.

## Questions

 * What is [tileIndex](https://github.com/iTowns/itowns/blob/master/src/Provider/3dTilesProvider.js#L186) ? From what I understand, it seems to be a list of [tiles](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/schema/tile.schema.json) (i.e. a flattened [tileset](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/schema/tileset.schema.json)). However, the tiles still have children attributes (and hence the hierarchy between tiles is still present but not stored as originally in the tileset tree) 
 * What is the purpose of the [extendTileset](https://github.com/iTowns/itowns/blob/master/src/Provider/3dTilesProvider.js#L128) method ?

gchoqueux:
Thank you for this refactoring proposal.

> What is tileIndex ? From what I understand, it seems to be a list of tiles (i.e. a flattened tileset). However, the tiles still have children attributes (and hence the hierarchy between tiles is still present but not stored as originally in the tileset tree)

I don't know the code of 3DTiles yet. I still don't clearly understand `tileIndex`. It seems to be the structure of tiles that allows the process to perform culling, replacement and fecthing of tiles.

There is also the meshs structure of the tiles in the `LayerGeometry.object3d`. It must be possible to unify these two structures.

> What is the purpose of the extendTileset method ?

This method is used when loading a new tile. It just adds this new tile to the `tileIndex`.

Otherwise, an important thing in the refactoring will be to do `3DTilesLayer` and `3DTilesSource` and remove `3DTileProvider` and `3DTileProcessing`.

jailln:
Thank you for the clarifications! With these information and by having a deeper look in the code, it seems that:

[...]

PR #1387, opened 2020-05-11
3D Tiles refactoring

jailln:
## Description
Some refactoring of 3D Tiles management to bring it closer to the specification. Main changes are (see commits for a detailed list of changes):

* Update naming convention to match the one that seemed to have been chosen for the 3D Tiles layer (prefixing with C3DTiles, that I simplified to C3DT for core classes).
* Start creating classes matching the specification (C3DTileset, C3DTBatchTable, C3DTBoundingVolume)
* Various attributes renaming (e.g. tileIndex -> tileset, index -> tiles)
* Update 3D Tiles extensions management: they are now registered in the layer with their name and an object holding a parsing method and other methods such as `getPickingInfo`. 
* Update 3D Tiles related tests and documentation.

Some of the doc doesn't seem to be generated (for classes: C3DTileset, C3DTBatchTable, C3DTBoundingVolume and C3DTBatchTableHierarchyExtension) while it does for C3DTExtensions and C3DTAbstractExtension; any ideas why?

## Motivation and Context

This PR starts fixing issue #1171


gchoqueux:
Thank you for this contribution it looks very promising!

#### Before we start the review, I have a few checks and precisions.

* prefixing class with `C3DTile`  ✔️ 
* start creating classes matching the specification ✔️ 
    * To facilitate the reviewing, could you specify if there are **algorithms rewrites** or **algorithms moves** to their corresponding class?
* Various attributes renaming ✔️ 
* Update 3D Tiles extensions management: they are now registered in the layer with their name and an object holding a parsing method and other methods such as `getPickingInfo`.
   * I need to take a closer look, but generally `picking `is usually decorrelated from the data 
* Update 3D Tiles related tests and documentation. ✔️ 


#### Could you squash some commits to simplify the history ?
and move all messages commit in squashed commit.
I can participate in this work
by example:
  - chore(3dtiles): add check on bounding volume attribute.
  - split this commit : chore(3dtiles): clean C3DTilesSource and C3DTilesLayer.
  - chore(3dtiles): clean and document C3DTileset class.

#### Generate documentation

> Some of the doc doesn't seem to be generated (for classes: C3DTileset, C3DTBatchTable, C3DTBoundingVolume and C3DTBatchTableHierarchyExtension) while it does for C3DTExtensions and C3DTAbstractExtension; any ideas why?

@zarov , have you any ideas?

zarov:
> Some of the doc doesn't seem to be generated (for classes: C3DTileset, C3DTBatchTable, C3DTBoundingVolume and C3DTBatchTableHierarchyExtension) while it does for C3DTExtensions and C3DTAbstractExtension; any ideas why?

You need to declare your classes like this:
```js
class C3DTBatchTable {
    // ...
}

export default C3DTBatchTable;
```

I'm not sure why your version does not work, but this way all classes are the same.

jailln:
@gchoqueux :

**Clarifications:**

> * start creating classes matching the specification ✔️
>   * To facilitate the reviewing, could you specify if there are algorithms rewrites or algorithms moves to their corresponding class?

It is mainly code moving and sometimes simplification and renaming but without changes to the core algorithm.

> * Update 3D Tiles extensions management: they are now registered in the layer with their name and an object holding a parsing method and other methods such as getPickingInfo.
>     * I need to take a closer look, but generally picking is usually decorrelated from the data

The idea behind this choice is to allow one to register an object taking care of everything that is related with this extension, including parsing or getting the information of relative to this extension that can be displayed to a user; so that the code relative to an extension is autonomous and can be "plugged" to itowns from an external project.

**Commits organization:**

> Could you squash some commits to simplify the history ?
> and move all messages commit in squashed commit.
> I can participate in this work
> by example:
> 
>     chore(3dtiles): add check on bounding volume attribute.
>     split this commit : chore(3dtiles): clean C3DTilesSource and C3DTilesLayer.
>     chore(3dtiles): clean and document C3DTileset class.

Ok for the commit to split. Regarding the other ones: should I squash them with their respective previous commits ?

**Documentation:**

@zarov, thanks it worked :+1:

gchoqueux:
> It is mainly code moving and sometimes simplification and renaming but without changes to the core algorithm.

Very good

> The idea behind this choice is to allow one to register an object taking care of everything that is related with this extension, including parsing or getting the information of relative to this extension that can be displayed to a user; so that the code relative to an extension is autonomous and can be "plugged" to itowns from an external project.

ok

> Ok for the commit to split. Regarding the other ones: should I squash them with their respective previous commits ?

If possible, the idea is to minimize the number of commits for one file and describe all changes in the commit message.

jailln:
@gchoqueux done; let me know if more squashing are needed.

gchoqueux:
Yeah, thanks, that's fine. Just squashes those last few commits.
Because they are 4 commits with minor modifications (`console`, missing properties, docs...)

* Keep **refactor(3dtiles):Add spec classes and modify extensions management.**  squashes with :
   * chore(3dtiles): clean C3DTilesSource
   * chore(3dtiles): Remove unnecessary and temporary C3DTilesLayer attribute
   * chore(3dtiles): add warning for unsupported bounding volumes types.
   * refactor(3dtiles): Add property to 3D Tiles core classes indicating their type

jailln:
Done.

gchoqueux:
Ok, I'm gonna start reviewing

src/Core/3DTiles/C3DTBatchTable.js
```diff
@@ -0,0 +1,74 @@
...
+        this.batchLength = batchLength;
+
+        // Separate the content and the possible extensions
+        // When an extension is found, we call its parser and append the
+        // returned object to batchTable.extensions
+        // Extensions must be registered in the layer (see an example of this in
+        // 3dtiles_hierarchy.html)
+        if (json.extensions) {
+            this.extensions =
+                registeredExtensions.parseExtensions(json.extensions, this);
```
gchoqueux: If extensions were filtered, there would be no need to pass context. 
We could type the extensions.

By example in `C3DTBatchTableHierarchyExtension` and `C3DTBatchTable` :

Add property `C3DTBatchTableHierarchyExtension.type = 'batchTableExtension'`
then in `BatchTableParser` we could filter `registeredExtensions` by this type,
before passing it in to the `C3DTBatchTable` constructor.

We could also merge the `BatchTableParser.parse` and `C3DTBatchTable.constructor` because
this parser instances `C3DTBatchTable`.

src/Core/3DTiles/C3DTBatchTable.js
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Creates and returns a javascript object holding the displayable
+     * information from the batch table and from extensions of the batch table
+     * for a given feature
+     * @param {integer} batchID - id of the feature in the batch
+     * @returns {Object} - displayable information relative to the batch
+     * table. Object is formatted as follow:
+     * {BatchTable: {BatchTableFeatureProperties} }.
+     */
+    getPickingInfo(batchID) {
```
gchoqueux: could you rename to `getInfoById` ?

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Creates and returns a javascript object holding the displayable
+     * information relative to this extension and for a given feature and
+     * its parents
+     * @param {integer} featureId - id of the feature
+     * @returns {Object} - displayable information relative to this extension
+     * and for the feature with id=featureId and for its parents
+     */
+    getPickingInfo(featureId) {
```
gchoqueux: could you rename `getInfoById` ?

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,149 @@
...
+ * instance (i.e. georgraphic feature e.g. building, roof, etc.) an array of the
+ * indexes of its parents. For example, the parents of the instance 0 can be
+ * found using inverseHierarchy[0].
+ * @property {number[]} instancesIdxs - For each instance of the extension,
+ * contains a javascript object with classId and instanceIdx. classId is the id
+ * of the class (from this.classes) of the instance. instanceIdx is the index of
+ * the instance in this class. Goal: Ease the retrieval of the properties of an
+ * instance.
+ */
+class C3DT_BTHExtension_BatchTable {
```
gchoqueux: In class name `C3DT_BTHExtension_BatchTable`, the 2 letters **`BT`** also stands for **B**atch **T**able?

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,149 @@
...
+        return pickingInfo;
+    }
+}
+
+/**
+ * @classdesc
+ * 3D Tiles
+ * [Batch Table Hierarchy Extension](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/extensions/3DTILES_batch_table_hierarchy)
+ */
+class C3DTBatchTableHierarchyExtension extends C3DTAbstractExtension {
```
gchoqueux: `C3DTBatchTableHierarchyExtension` isn't a really extension?
because is parse/instance `C3DT_BTHExtension_BatchTable` and handle picking Info?

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -0,0 +1,127 @@
...
+            new THREE.Vector3(e, n, t));
+    }
+
+    initBoundingSphere(sphere) {
+        this.sphere = new THREE.Sphere(
+            new THREE.Vector3(sphere[0], sphere[1],
+                sphere[2]), sphere[3]);
+    }
+
+    static boundingVolumeCulling(layer, boundingVolume, camera, tileMatrixWorld) {
```
gchoqueux: If `boundingVolume` is instance of  `C3DTBoundingVolume`, the static function isn't necessary?

same comment for `viewerRequestVolumeCulling` function.

src/Core/3DTiles/C3DTExtensions.js
```diff
@@ -0,0 +1,118 @@
...
+ * This class is an interface for 3D Tiles extensions. Extensions implemented by
+ * the user must inherit from this class. Example of extension extending this
+ * class:
+ *  [$3DTBatchTableHierarchyExtension]{@link $3DTBatchTableHierarchyExtension}
+ */
+export class C3DTAbstractExtension {
+    /**
+     * This class must not be instanciated directly.
+     */
+    constructor() {
```
gchoqueux: `C3DTAbstractExtension` is more parser and picking handler?

This abstraction may not be necessary if it has no common properties and implemented functions?

`C3DTExtensions` could check the integrity `Extension` in `registerExtension`.
by example : handle  the methods  `parse` or `getPickingInfo` if doesn't exist.

src/Core/3DTiles/C3DTileset.js
```diff
@@ -0,0 +1,112 @@
...
+     * (nodes of the tree).
+     * @param {object} tile - current tile
+     * @param {string} baseURL - url of the source tileset and tiles
+     * @param {object} parent - parent tile (used for computing the transform
+     * matrix from local to global coordinates)
+     * @param {object} registeredExtensions - 3D Tiles extensions registered
+     * in the C3DTilesLayer (see {@link C3DTExtensions}
+     */
+    parseTiles(tile, baseURL, parent, registeredExtensions) {
+        const inverseTileTransform = new THREE.Matrix4();
```
gchoqueux: `inverseTileTransform ` could be instanced before  `class C3DTileset`

src/Core/3DTiles/C3DTileset.js
```diff
@@ -0,0 +1,112 @@
...
+
+        // compute transform (will become Object3D.matrix when the object is
+        // downloaded)
+        tile.transform =
+            tile.transform ? (new THREE.Matrix4()).fromArray(tile.transform) :
+                undefined;
+
+        // The only reason to store _worldFromLocalTransform is because of
+        // extendTileset where we need the transform chain for one tile.
+        tile._worldFromLocalTransform = tile.transform;
```
gchoqueux, line 66: `tile._worldFromLocalTransform = tile.transform;` could be moved in `else` because is crushed if there is `parent._worldFromLocalTransform`

src/Core/3DTiles/C3DTileset.js
```diff
@@ -0,0 +1,112 @@
...
+        tile.boundingVolume = tile.boundingVolume ?
+            new C3DTBoundingVolume(tile.boundingVolume,
+                inverseTileTransform, registeredExtensions) : undefined;
+
+        this.tiles.push(tile);
+        tile.tileId = this.tiles.length - 1;
+        tile.baseURL = baseURL;
+        if (tile.children) {
+            for (const child of tile.children) {
+                this.parseTiles(child, baseURL, tile, registeredExtensions);
```
gchoqueux, line 100: I notice that the tree structure is flattened and that this is also the case in the master branch.

here:
```js
this.parseTiles(child, baseURL, tile, registeredExtensions);
```

All `child` and `tile`  will finally be added in `this.tiles`, is that normal?

src/Layer/C3DTilesLayer.js
```diff
@@ -210,15 +42,31 @@ class C3DTilesLayer extends GeometryLayer {
         // custom cesium shaders are not functional;
         this.overrideMaterials = config.overrideMaterials !== undefined ? config.overrideMaterials : true;
         this.name = config.name;
+        this.registeredExtensions = config.registeredExtensions !== undefined ? config.registeredExtensions : new C3DTExtensions();
```
gchoqueux: you could write:
```js
```

src/Parser/BatchTableParser.js
```diff
@@ -1,116 +1,22 @@
...
-        if (this.content.extensions[extensionName]) {
-            delete this.content.extensions[extensionName];
-        }
-        // Delete extensions from content if empty
-        if (Object.keys(this.content.extensions).length === 0) {
-            delete this.content.extensions;
-        }
-    }
-}
+import C3DTBatchTable from '../Core/3DTiles/C3DTBatchTable';
```
gchoqueux: you could write 
```js
import C3DTBatchTable from 'Core/3DTiles/C3DTBatchTable';
```

gchoqueux (commented):
Thanks, this PR better structures the 3d tiles codes.

I'm doing a first global review that focuses on the structure of the classes that manage the `Extensions`.

src/Core/3DTiles/C3DTBatchTable.js
jailln: I am not sure to understand the beginning of your comment regarding adding a `type` attribute to the Extensions classes (e.g. `C3DTBatchTableHierarchyExtension`) to avoid passing the context; could you elaborate a bit please?

I agree that we should remove the `BatchTableParser` and move its content to `C3DTBatchTable`; I initially planned to do it in another PR but I will do it here.

src/Core/3DTiles/C3DTBatchTable.js
jailln: Ok

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: Ok

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: Yes and the **H** for **H**ierarchy, as a shorthand for the name of the extension (https://github.com/CesiumGS/3d-tiles/tree/master/extensions/3DTILES_batch_table_hierarchy)

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: The idea behind this structure is that one object is in charge of the extension (e.g. `C3DTBatchTableHierarchyExtension` is in charge of the 3DTILES_batch_table_hierarchy extension). Let's name it the extension manager. However, an extension can have data scattered in multiple classes of 3D Tiles (in this case it is only in the batch table but other extensions also have information in tileset, bounding volume, etc.). 
The idea is to have one class per 'piece' of extension (e.g. one class for the piece in the batch table, one class for the piece in the tileset) that is responsible for parsing and possibly doing other thing with this 'piece' of extension. The instance of this class corresponds to the parsed 'piece of extension' and is stored in the corresponding 3D Tiles class and instantiated by the extension manager depending on the calling context (i.e. the 3D Tiles class).

You can see an example implementation [here](https://github.com/jailln/UDV/tree/UDV-temporal/UDV-Core/src/Modules/Temporal/3DTILES_temporal) where we implement an extension of 3D Tiles (not yet in the standard spec) having data in multiple 3D Tiles classes (the entry point is [this file](https://github.com/jailln/UDV/blob/UDV-temporal/UDV-Core/src/Modules/Temporal/3DTILES_temporal/3DTemporalExtension.js)). 

Let me know if my explanation is not clear, I can work on a schema / a sequence diagram.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Yep you're right; I removed the `static`

src/Core/3DTiles/C3DTExtensions.js
jailln: The idea behind `C3DTAbstractExtension` was to expose a clearer interface for extensions to users (and the associated documentation). But if you prefer I can remove the `C3DTAbstractExtension` and check the integrity of registered extensions in the `registerExtension` method.

src/Core/3DTiles/C3DTileset.js
jailln: You mean before the class declaration ? What would be the benefit ?

src/Core/3DTiles/C3DTileset.js
jailln: Ok

src/Core/3DTiles/C3DTileset.js
jailln: Yes it is already the case on the master branch. In addition to that, there is a tree structure with the nodes representing the tiles contents stored as threejs objects in the `object3d` attribute of the layer. (it is in fact a bit more complicated than this, see issue #1039 )

I don't know why this choice has been made in the first place but I think we may be able to simplify this and keep only one tree structure with the tiles and their content loaded as threejs objects (as discussed in the third item of issue #1171 ). I haven't looked into it yet though and I think it might be better to do it in a future PR ?

src/Layer/C3DTilesLayer.js
jailln: Ok

src/Parser/BatchTableParser.js
jailln: Ok

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
gchoqueux: What bothers me is that it's a two-time `batch table` in the name.

src/Core/3DTiles/C3DTExtensions.js
gchoqueux: Yes, I prefer a explanations in the documentation, unless it is possible to add to this class properties/methods common to all `Extensions`.

src/Core/3DTiles/C3DTileset.js
gchoqueux: yes, before the class declaration because `inverseTileTransform` is instanced each call function and increases the memory so decreases performance to clean memory with the Garbage Collector.

src/Core/3DTiles/C3DTileset.js
gchoqueux: Yes, indeed a major refactor is necessary.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
gchoqueux: **Common response with this other [comment](https://github.com/iTowns/itowns/pull/1387#discussion_r427359755)**

`C3DTBatchTableHierarchyExtension` is in charge of `C3DT_BTHExtension_BatchTable` ?
Yes, I'd like a diagram to make it clear.

May be, it's possible to simplify, by example :
```js
// Extension class to register in C3DTExtensions
class fooExtension {
    constructor(json) {
        // Parsing json
    }
    // type extension to filter in context
    static get type() { return 'foo' };
}

// in  C3DTExtensions class
C3DTExtensions.parseExtensions(extensionsJSON, type) {
    const parsedExtensions = {};
    for (const extName in extensionsJSON) {
        if (this.isExtensionRegistered(extName)) {
            const extension = this.getExtension(extName);
            if (extension.type == type) {
                parsedExtensions[extName] = new extension(extensionsJSON[extName]);
            }
        }
    }
    return parsedExtensions;
}
```
wdym?

src/Core/3DTiles/C3DTBatchTable.js
gchoqueux: [see comment](https://github.com/iTowns/itowns/pull/1387#discussion_r427881753)

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: Yes, in my opinion the choice of the extension name is confusing but it is what they have chosen in the standard specification.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: In case of the `3DTILES_batch_table_hierarchy` extension, it may work but for extensions in general I think it won't. 3D Tiles extensions can indeed be divided in several parts; each part being stored in a 3D Tiles class. For instance, the [draco point compression extension](https://github.com/CesiumGS/3d-tiles/tree/master/extensions/3DTILES_draco_point_compression) is composed of two parts: one in the `featureTable` class of 3D Tiles and one in the `batchTable` class (see [its JSON schema](https://github.com/CesiumGS/3d-tiles/tree/master/extensions/3DTILES_draco_point_compression/schema)). The implementation therefore follows this idea: the extension is implemented by one class and each extension class is also implemented by one class. When a 3D Tiles class is parsed (e.g. batch table), if an extension part is found it is passed to the corresponding extension which calls the extension part class for the parsing. 

In this PR, `C3DTBatchTableHierarchyExtension` corresponds to the extension and `C3DT_BTHExtension_BatchTable` corresponds to the extension part stored in the batch table.

src/Core/3DTiles/C3DTileset.js
jailln: Ok, I'll move it. Just for my personal curiosity, do you know if this gain is quantifiable or if it has been quantified on simple cases?

src/Core/3DTiles/C3DTBatchTable.js
jailln: `BatchTableParser` removed.

src/Core/3DTiles/C3DTExtensions.js
jailln: Actually, I think the parse method may be shared by all extension; I'll push a proposition.

src/Core/3DTiles/C3DTileset.js
gchoqueux: With tools performances in Chrome you can track the garbage collector time (Major GC).
Instance Object only used in scope function increase major GC time.

jailln: Ok, thanks.

jailln:
I updated the extensions management as we discussed; let me know if it's ok (except from some documentation that I still need to update). I also took this opportunity to simplify `3DTilesHelper` and to move some code to `C3DTilesLayer`.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
```diff
@@ -69,46 +73,36 @@ class BatchTableHierarchyExtension extends $3dTilesAbstractExtension {
...
-     * object.
-     * @param {Object} json - json holding the extension
-     * @return {Promise} - a promise that resolves with a
-     *     BatchTableHierarchyExtension object.
-     */
-    parse(json) {
-        return Promise.resolve(new BatchTableHierarchyExtension(json));
-    },
-};
+function C3DT_BTHExtension_BatchTable_parser(json) {
```
gchoqueux: This function shouldn't be necessary , could you directly register the class instead of `C3DT_BTHExtension_BatchTable_parser`?

src/Core/3DTiles/C3DTExtensions.js
```diff
@@ -0,0 +1,88 @@
...
+        const result = {};
+        // Call getInfoById for each extension of type batch table registered
+        // in the layer
+        for (const extName in batchTableExtensions) {
+            // if it is registered it has been parsed and it has a
+            // getInfoByID method
+            if (this.isExtensionRegistered(extName)) {
+                const extInfo = {};
+                extInfo[extName] = batchTableExtensions[extName].getInfoById(batchID);
+                Object.assign(result, extInfo);
```
gchoqueux: You could directly write `result[extName] = batchTableExtensions[extName].getInfoById(batchID);` ?

src/Core/3DTiles/C3DTExtensions.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    getFeaturesInfoById(batchTableExtensions, batchID) {
+        const result = {};
+        // Call getInfoById for each extension of type batch table registered
+        // in the layer
+        for (const extName in batchTableExtensions) {
+            // if it is registered it has been parsed and it has a
+            // getInfoByID method
+            if (this.isExtensionRegistered(extName)) {
```
gchoqueux: This test `this.isExtensionRegistered(extName)`  isn't necessary?
The classes in `batchTableExtensions ` are instanced only if  extension is already registered?

src/Layer/C3DTilesLayer.js
```diff
@@ -253,6 +102,63 @@ class C3DTilesLayer extends GeometryLayer {
...
+    getInfoFromIntersectObject(intersects) {
+        const resultInfo = {};
+
+        // Finds the batch table of an object in a 3D Tiles layer. This is
+        // for instance needed when picking because we pick the geometric
+        // object which is not at the same level in the layer structure as
+        // the  batch table. More details here on itowns internal
+        // organization of 3DTiles:
+        // https://github.com/MEPP-team/RICT/blob/master/Doc/iTowns/Doc.md#itowns-internal-organisation-of-3d-tiles-data
+        const findBatchTable = (object) => {
```
gchoqueux: Could you move `findBatchTable` out of `getInfoFromIntersectObject`?

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
```diff
@@ -69,46 +73,36 @@ class BatchTableHierarchyExtension extends $3dTilesAbstractExtension {
...
-                instanceProperties[property] =
-                    this.classes[instanceClassId].instances[property][instanceIdx];
-            });
+
+        const properties = Object.keys(this.classes[instanceClassId].instances);
+        for (let i = 0; i < properties.length; i++) {
+            const property = properties[i];
+            instanceProperties[property] =
+                this.classes[instanceClassId].instances[property][instanceIdx];
+        }
```
gchoqueux: Could you replace by :
```js
const instances = this.classes[instanceClassId].instances;
for (const key in instances) {
    if (Object.prototype.hasOwnProperty.call(instances, key)) {
        instanceProperties[key] = instances[key][instanceIdx];
    }
}    
```

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
```diff
@@ -69,46 +73,36 @@ class BatchTableHierarchyExtension extends $3dTilesAbstractExtension {
...
         pickingInfo[featureClass] = instanceProperties;
         // If this feature has parent(s), recurse on them
         if (this.inverseHierarchy && this.inverseHierarchy[featureId]) {
-            this.inverseHierarchy[featureId].forEach(
-                parentId => Object.assign(pickingInfo,
-                    this.getPickingInfo(parentId)));
+            for (let i = 0; i < this.inverseHierarchy[featureId].length; i++) {
+                const parentID = this.inverseHierarchy[featureId][i];
+                Object.assign(pickingInfo, this.getInfoById(parentID));
+            }
```
gchoqueux: could you replace by : 
```js
for (const parentID of  this.inverseHierarchy[featureId]) {
  Object.assign(pickingInfo, this.getInfoById(parentID));
}
```

src/Core/3DTiles/C3DTExtensions.js
```diff
@@ -0,0 +1,88 @@
...
+        const parsedExtensions = {};
+        for (const extName in extensionsJSON) {
+            if (this.isExtensionRegistered(extName)) {
+                parsedExtensions[extName] = this.getExtension(extName, type)(extensionsJSON[extName]);
+            }
+        }
+        return parsedExtensions;
+    }
+
+    getFeaturesInfoById(batchTableExtensions, batchID) {
```
gchoqueux: I'm not sure that the responsibility for acquiring information should be placed in `3DTExtensions`.
It should only handle the registration of extensions and the provisioning of them.

src/Core/3DTiles/C3DTileset.js
```diff
@@ -0,0 +1,113 @@
...
+ *  @propertiy {object[]} tiles - an array holding all the
+ *  [tiles](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/schema/tile.schema.json)
+ *  (not their content which is stored in the attribute object3d of the
+ *  layer). This list represents a flattened tileset.
+ *  @property {object} extensions - Extensions of the tileset in the form:
+ * {extensioName1: extensionObject1, extensioName2: extensionObject2, ...}
+ */
+class C3DTileset {
+    constructor(json, baseURL, registeredExtensions) {
+        this.type = C3DTilesTypes.tileset;
```
gchoqueux, line 31: It's more simple to write `'tileset'` directly?

src/Core/3DTiles/C3DTileset.js
```diff
@@ -0,0 +1,113 @@
...
+        this.properties = json.properties;
+        this.geometricError = json.geometricError;
+        this.extensionsUsed = json.extensionsUsed;
+        this.extensionsRequired = json.extensionsRequired;
+        this.tiles = [];
+        this.parseTiles(json.root, baseURL, undefined, registeredExtensions);
+
+        if (json.extensions) {
+            this.extensions =
+                registeredExtensions.parseExtensions(json.extensions, this);
```
gchoqueux: is `this.type` to pass?

gchoqueux (commented):
Yes, that's good, the management of extensions is simpler.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: I can register an instance of the class and move the content of the constructor to a parse method of the class; would you prefer that ?

src/Core/3DTiles/C3DTExtensions.js
jailln: Yes, indeed. Done

src/Core/3DTiles/C3DTExtensions.js
jailln: Removed.

src/Layer/C3DTilesLayer.js
jailln: Done.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: Done.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: Done.

src/Core/3DTiles/C3DTExtensions.js
jailln: Moved to `getInfoFromIntersectObject()` of `C3DTilesLayer`.

src/Core/3DTiles/C3DTileset.js
jailln: The idea was to use `C3DTilesTypes` when registering an extension to avoid having to write the same string at different places and relating on the fact they are written in the same way to correctly parse extensions; except that I forgot to use it in `3dtiles_batch_table` example... I modified that.

src/Core/3DTiles/C3DTileset.js
jailln: Yes, thanks for spotting this; updated.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
gchoqueux: I meant : could register `C3DT_BTHExtension_BatchTable` instead of parser:

```js
extensions.registerExtension('3DTILES_batch_table_hierarchy',
        { batchtable: C3DT_BTHExtension_BatchTable });
```

src/Core/3DTiles/C3DTileset.js
gchoqueux: `C3DTilesTypes.tileset` it also needs to be written correctly each time and it takes longer to write than `tileset`.

src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js
jailln: No this is not possible, I can only register a class instance or a function. Maybe registering a class instance is clearer than registering `C3DT_BTHExtension_BatchTable_parser`.

src/Core/3DTiles/C3DTileset.js
jailln: Well relying on hard-coded strings instead of an enum is generally a bad idea in my opinion... First, `C3DTilesTypes.tileset` being a variable, you are less likely to make a mistake when writting it (e.g. thanks to you IDE). Then, maintenance is easier in case a value must be changed (you only change it in the enum). Finally, you can also expose to the user what values are available by exporting this enum with associated documentation.

On another note, I can write C3DTilesTypes this way: (which is a bit simpler)

````
const C3DTilesTypes = {
    tileset: {},
    batchtable: {},
    boundingVolume: {},
};
````
Anyway, that is your call, let me know.

Comment (gchoqueux) [reply src/Core/3DTiles/C3DTileset.js:31]: You're right. Just keep it like that.

File: src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js (hunk 13 above)
Comment (gchoqueux) [reply src/Core/3DTiles/C3DTBatchTableHierarchyExtension.js]: I've tried this and it works.

The registering:
```js
extensions.registerExtension("3DTILES_batch_table_hierarchy",
    { [itowns.C3DTilesTypes.batchtable]:  itowns.C3DT_BTHExtension_BatchTable });
```

The instancing in `C3DTExtensions#parseExtensions`
```js
parsedExtensions[extName] = new (this.getExtension(extName, type))(extensionsJSON[extName]);
```

jailln: Nice, thanks ! I must have made a mistake when I tried it. It is simpler this way.

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -0,0 +1,128 @@
...
+        this.region = box;
+    }
+
+    initBoundingBox(box) {
+        // TODO: only works for axis aligned boxes
+        // box[0], box[1], box[2] = center of the box
+        // box[3], box[4], box[5] = x axis direction and half-length
+        // box[6], box[7], box[8] = y axis direction and half-length
+        // box[9], box[10], box[11] = z axis direction and half-length
+        const center = new THREE.Vector3(box[0], box[1], box[2]);
```
gchoqueux: Could you move out this class the `center` instance (to avoid Garbage collector).

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -0,0 +1,128 @@
...
+        const w = center.x - box[3];
+        const e = center.x + box[3];
+        const s = center.y - box[7];
+        const n = center.y + box[7];
+        const b = center.z - box[11];
+        const t = center.z + box[11];
+
+        this.box = new THREE.Box3(new THREE.Vector3(w, s, b),
```
gchoqueux: Could you move out this class the `Vector3` instance (to avoid Garbage collector).

You can use some `Vector3` instances in all methods.

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -0,0 +1,128 @@
...
+    initBoundingSphere(sphere) {
+        this.sphere = new THREE.Sphere(
+            new THREE.Vector3(sphere[0], sphere[1],
+                sphere[2]), sphere[3]);
+    }
+
+    boundingVolumeCulling(camera, tileMatrixWorld) {
+        if (this.region &&
+            !camera.isBox3Visible(this.region.box3D,
+                tileMatrixWorld.clone()
```
gchoqueux: Could you instance (before the class) a `Matrix4` and use  `Matrix4.multiplyMatrices` to avoid a clone (and Garbage collector)
```js
matrix.multiplyMatrices(tileMatrixWorld,this.region.matrix);
```

src/Core/3DTiles/C3DTExtensions.js
```diff
@@ -0,0 +1,78 @@
...
+     */
+    parseExtensions(extensionsJSON, type) {
+        const parsedExtensions = {};
+        for (const extName in extensionsJSON) {
+            if (this.isExtensionRegistered(extName)) {
+                const extManager = this.getExtension(extName, type);
+                if (extManager) {
+                    parsedExtensions[extName] = new extManager(extensionsJSON[extName]);
+                } else {
+                    console.error(`${'No extension manager registered for' +
```
gchoqueux: It's not possible to have the error because you check with `isExtensionRegistered`?

src/Layer/C3DTilesLayer.js
```diff
@@ -253,6 +102,68 @@ class C3DTilesLayer extends GeometryLayer {
...
+    // organization of 3DTiles:
+    // https://github.com/MEPP-team/RICT/blob/master/Doc/iTowns/Doc.md#itowns-internal-organisation-of-3d-tiles-data
+    findBatchTable(object) {
+        if (object.batchTable) {
+            return object.batchTable;
+        }
+        if (object.parent) {
+            return this.findBatchTable(object.parent);
+        }
+        return undefined;
```
gchoqueux: `return undefined;` isn't necessary.

src/Layer/C3DTilesLayer.js
```diff
@@ -253,6 +102,68 @@ class C3DTilesLayer extends GeometryLayer {
...
+        let batchTable = {};
+        // First, we get the ID and the batch table of the intersected object.
+        // (the semantic information about a feature is located in its batch
+        // table (see 3D Tiles specification).
+        for (let i = 0; i < intersects.length; i++) {
+            // interAttributes are glTF attributes of b3dm tiles (i.e.
+            // position, normal, batch id)
+            const interAttributes = intersects[i].object.geometry.attributes;
+            if (interAttributes) {
+                const batchIDs = interAttributes._BATCHID || interAttributes.batchId;
```
gchoqueux: there is 2 attributes names (`_BATCHID` and `batchId`)  for batch id in the specification ?

gchoqueux (commented):
I've added a few more comments on the form

src/Layer/C3DTilesLayer.js
jailln: To my knowledge there is only `_BATCHID` but you added `batchId` in [this commit](https://github.com/iTowns/itowns/commit/7b067c7deb21614b731c90e8486ac2fffbcb5e82#diff-276f2190b0230ea4d24aeca3427e133eR25), so I thought there was a good reason to that ? If no, I will remove it since I tried it without `batchId` and it works.

src/Layer/C3DTilesLayer.js
jailln: Ok, removed.

src/Core/3DTiles/C3DTExtensions.js
jailln: Actually `isExtensionRegistered` only checks for the extension name when it is encountered in the tileset while this test also checks that the extension manager is registered for the `type`. I moved this test into `getExtension`.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Ok, done.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Ok, done.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Ok, done.

src/Layer/C3DTilesLayer.js
gchoqueux: ok, I'll look into it. We'll leave it like this for now.

gchoqueux:
All right, thank you for your efforts.
We'll now test this PR with different use cases.

@gmaillet , @Mkonini could try this PR with with yours 3d tiles applications?

src/Layer/C3DTilesLayer.js
```diff
@@ -132,7 +131,7 @@ class C3DTilesLayer extends GeometryLayer {
             // position, normal, batch id)
             const interAttributes = intersects[i].object.geometry.attributes;
             if (interAttributes) {
-                const batchIDs = interAttributes._BATCHID || interAttributes.batchId;
+                const batchIDs = interAttributes._BATCHID;
```
gchoqueux: could you roll back, i don't remember why i added the `.batchId`

jailln: Done.

src/Core/3DTiles/C3DTileset.js
```diff
@@ -0,0 +1,113 @@
...
+ *  (e.g. {@link C3DTBatchTableHierarchyExtension}) to know in which context
+ *  (i.e. for which 3D Tiles class) the parsing of the extension should be done.
+ *  This value is set to true by default.
+ *  @property {object} asset - Generic information about the tileset, see
+ *  [asset specification]https://github.com/CesiumGS/3d-tiles/blob/master/specification/schema/asset.schema.json
+ *  @property {object} properties - Properties associated with the tileset, see
+ *  [tileset specification](https://github.com/CesiumGS/3d-tiles/blob/master/specification/schema/tileset.schema.json#L11)
+ *  @property {number} geometricError - see [tileset
+ *  specification](https://github.com/CesiumGS/3d-tiles/blob/master/specification/schema/tileset.schema.json#L18)
+ *  @propertu {array} extensionsUsed - see [tileset specification](https://github.com/CesiumGS/3d-tiles/blob/master/specification/schema/tileset.schema.json#L27)
```
zarov: Typo `@property`, and could you use `string[]` instead ?

jailln: Done, thanks for spotting that!

jailln:
I added some documentation, squashed the commits and rebased on master.

The documentation of `C3DTilesTypes` doesn't get generated, co you have any ideas why ? @zarov maybe ?

gchoqueux:
We revealed a issue (may be on screen space error).  We had try the same data  with this PR, the buildings wasn't displayed. Have you try with yours datas?

zarov:
> The documentation of `C3DTilesTypes` doesn't get generated, co you have any ideas why ? @zarov maybe ?

You can do something similar to [`CACHE_POLICIES` here](https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Cache.js#L3-L19).

gchoqueux:
I find a problem, the bounding boxes are wrong:

#### Master
bbox master

#### PR
bbox PR

jailln:
> You can do something similar to [`CACHE_POLICIES` here](https://github.com/iTowns/itowns/blob/master/src/Core/Scheduler/Cache.js#L3-L19).

It doesn't seem to work (see my last commit). I also tried to change the default export to a named export but it work either.

@gchoqueux, thanks for finding that! I will look into it.

zarov:
> It doesn't seem to work (see my last commit). I also tried to change the default export to a named export but it work either.

It does not indeed, in fact it should use `@enum` instead of `@typedef`, my bad. However, enums are not correctly generated in the current documentation. What you can do is makes this change, but remove it from `docs/config.json` for now. I'll try to find some times to add it in the documentation, sorry about that.

jailln:
@gchoqueux Removing global variables from C3DTBoundingVolume fixes this bug. I guess we should avoid them...

@zarov, ok thanks!

gchoqueux:
It's not necessary to remove all global variables:
* in `initBoundingRegion`:
  just keep `const regionBox = new OBB();` because it needs to instance a new `OBB` for each `this.region`.
* in `initBoundingBox` replace all code by:
```js
center.set(box[0], box[1], box[2]);
size.set(box[3], box[7], box[11]).multiplyScalar(2);
this.box = new THREE.Box3();
this.box.setFromCenterAndSize(center, size);
```

jailln:
> It's not necessary to remove all global variables:
> 
>     * in `initBoundingRegion`:
>       just keep `const regionBox = new OBB();` because it needs to instance a new `OBB` for each `this.region`.
> 
>     * in `initBoundingBox` replace all code by:
> 
> 
> ```js
> center.set(box[0], box[1], box[2]);
> size.set(box[3], box[7], box[11]).multiplyScalar(2);
> this.box = new THREE.Box3();
> this.box.setFromCenterAndSize(center, size);
> ```

Ok, done.

gchoqueux:
Ok, thanks
Could you resolve resolve the conflict?
I re-check with other datas and it'll be over soon.

jailln:
> Ok, thanks
> Could you resolve resolve the conflict?
> I re-check with other datas and it'll be over soon.

Squashed and rebased.

gchoqueux (approved):
Thank you for your work!

zarov:
Well done @jailln :partying_face:

jailln:
Nice, thanks! :partying_face:

PR #1388, opened 2020-05-12
refacto(controls): reuse picked position

zarov:
The picked position on the ground was computed 3 times for a mouse wheel
handling, while it was the same everytime. This has been fixed by adding
a parameter to `CameraUtils#getTransformCameraLookingAtTarget`.

This impacts only `GlobeControls` as far as I know. One thing I'm not sure on is the impact on the `getLookAtFromMath` part. I don't see when this is used, beside for functional testing ?

### Before
Capture d’écran 2020-05-12 à 13 15 42

### After
Capture d’écran 2020-05-12 à 13 17 25

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -639,7 +639,7 @@ class GlobeControls extends THREE.EventDispatcher {
         event.preventDefault();
 
         this.updateTarget();
-        previous = CameraUtils.getTransformCameraLookingAtTarget(this.view, this.camera);
+        previous = CameraUtils.getTransformCameraLookingAtTarget(this.view, this.camera, pickedPosition);
         this.state = this.states.inputToState(event.button, currentKey);
 
         const coords = this.view.eventToViewCoords(event);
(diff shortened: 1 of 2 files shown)


gchoqueux:
I think it's good idea, in try it on other points, see :
https://github.com/gchoqueux/itowns/tree/REVIEW_speedup_globecontrols

But I doesn't confirm the robustness.

zarov:
It could be added to other parts indeed, I'll do that.

zarov:
Your proposition doesn't seem to work correctly, maybe because of the asynchronous tasks using animation. I'll let it as it is in my original version. However I saw potential improvements in `GlobeControls`, like [this part](https://github.com/iTowns/itowns/blob/master/src/Controls/GlobeControls.js#L583-L588), that could potentially be replaced by
```js
cameraTarget.position.copy(pickedPosition);
targetPosition.copy(pickedPosition);
```

gchoqueux:
> Your proposition doesn't seem to work correctly, maybe because of the asynchronous tasks using animation. I'll let it as it is in my original version. However I saw potential improvements in GlobeControls, like this part, that could potentially be replaced by

I project picket position on z axis camera to improve accuracy.

src/Controls/GlobeControls.js
```diff
@@ -974,26 +974,29 @@ class GlobeControls extends THREE.EventDispatcher {
 
     /**
      * Returns the "range": the distance in meters between the camera and the current central point on the screen.
+     * @param {THREE.Vector3} [position] - The position to consider as picked.
```
gchoqueux: `The position to consider as picked on the ground`

gchoqueux (approved):

PR #1389, opened 2020-05-13
fix: sort features by order in VectorTileParser

zarov:
This was removed in 0bca7c4, but it is necessary because features in a
mvt/pbf isn't always in the right order when reading it.

#### What we currently have (no fix)

Capture d’écran 2020-05-13 à 09 09 28

#### What we should have (fix)

Capture d’écran 2020-05-13 à 09 09 06

### DIFF
--- a/src/Source/VectorTilesSource.js
+++ b/src/Source/VectorTilesSource.js
@@ -79,7 +79,7 @@ class VectorTilesSource extends TMSSource {
             const s = Object.keys(style.sources)[0];
             const os = style.sources[s];
 
-            style.layers.forEach((layer) => {
+            style.layers.forEach((layer, order) => {
                 layer.sourceUid = this.uid;
                 if (layer.type === 'background') {
                     this.backgroundLayer = layer;
@@ -109,6 +109,7 @@ class VectorTilesSource extends TMSSource {
 
                     this.layers[layer['source-layer']].push({
                         id: layer.id,
+                        order,
                         filterExpression: featureFilter(layer.filter),
                         minzoom: stops[0],
                         maxzoom: stops[stops.length - 1],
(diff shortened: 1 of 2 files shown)


src/Source/VectorTilesSource.js
```diff
@@ -79,7 +79,7 @@ class VectorTilesSource extends TMSSource {
             const s = Object.keys(style.sources)[0];
             const os = style.sources[s];
 
-            style.layers.forEach((layer) => {
+            style.layers.forEach((layer, order) => {
```
gchoqueux, line 82: `style.layers` is already sort, later we will try to use it.

src/Source/VectorTilesSource.js
```diff
@@ -79,7 +79,7 @@ class VectorTilesSource extends TMSSource {
             const s = Object.keys(style.sources)[0];
```
gchoqueux, line 79: I hadn't see in previous PR, but the 'then' ([line 78](https://github.com/iTowns/itowns/pull/1389/files#diff-2d9ab3dfc05c71bd9e11450089c13392L78)) isn't neccessary

zarov: It is, as we need to wait for the sprites resources to load before doing any further processing (it is needed to be used when creating new `Style`).

src/Source/VectorTilesSource.js
zarov: Yes, but it is quicker to use an index instead of doing `indexOf` everytime in the parser.

src/Source/VectorTilesSource.js
gchoqueux: ok

gchoqueux (approved):

PR #1392, opened 2020-05-19
Refactor Cache

gchoqueux:
## Description
 * add Cache Class.
 * instance one cache by Layer.
 * smart flush cache.

## Motivation
There only was one Cache for all datas. This cache was cleaned every frame.

### DIFF
--- a/src/Core/Scheduler/Cache.js
+++ b/src/Core/Scheduler/Cache.js
@@ -1,55 +1,56 @@
-let data = {};
 let entry;
+
+/**
+ * Cache policies for flushing. Those policies can be used when something is
+ * [set]{@link Cache.set} into the Cache, as the lifetime property.
+ *
+ * @typedef {Object} CACHE_POLICIES
+ *
+ * @property {number} INFINITE - The entry is never flushed, except when the
+ * `all` flag is set to `true` when calling {@link Cache.flush}.
+ * @property {number} TEXTURE - Shortcut for texture resources. Time is 15 minutes.
+ * @property {number} GEOMETRY - Shortcut for geometry resources. Time is 15 minutes.
+ * minutes.
+ */
+export const CACHE_POLICIES = {
+    INFINITE: Infinity,
+    TEXTURE: 900000,
+    GEOMETRY: 900000,
+};
 
 /**
  * This is a copy of the Map object, except that it also store a value for last
  * time used. This value is used for cache expiration mechanism.
- * <br><br>
- * This module can be imported anywhere, its data will be shared, as it is a
- * single instance.
- *
- * @module Cache
  *
  * @example
- * import Cache from './Cache';
+ * import Cache, { CACHE_POLICIES } from 'Core/Scheduler/Cache';
  *
- * Cache.set({ bar: 1 }, Cache.POLICIES.TEXTURE, 'foo');
- * Cache.set({ bar: 32 }, Cache.POLICIES.INFINITE, 'foo', 'toto');
+ * const cache = new Cache(CACHE_POLICIES.TEXTURE)
+ * cache.set({ bar: 1 }, 'foo');
+ * cache.set({ bar: 32 }, 'foo', 'toto');
  *
- * Cache.get('foo');
+ * cache.get('foo');
  *
- * Cache.delete('foo');
+ * cache.delete('foo');
  *
- * Cache.clear();
+ * cache.clear();
  *
- * Cache.flush();
+ * cache.flush();
  */
-const Cache = {
+class Cache {
     /**
-     * Cache policies for flushing. Those policies can be used when something is
-     * [set]{@link Cache.set} into the Cache, as the lifetime property.
-     *
-     * @typedef {Object} POLICIES
-     *
-     * @property {number} INFINITE - The entry is never flushed, except when the
-     * `all` flag is set to `true` when calling {@link Cache.flush}.
-     * @property {number} TEXTURE - Shortcut for texture resources. Time is 15 minutes.
-     * @property {number} ELEVATION - Shortcut for elevation resources. Time is 15
-     * minutes.
+     * @param      {number}  [lifetime=CACHE_POLICIES.INFINITE]  The cache expiration time for all values.
      */
-    POLICIES: {
-        INFINITE: Infinity,
-        TEXTURE: 900000,
-        ELEVATION: 900000,
-    },
+    constructor(lifetime = CACHE_POLICIES.INFINITE) {
+        this.lifeTime = lifetime;
+        this.lastTimeFlush = Date.now();
+        this.data = new Map();
+    }
 
     /**
      * Returns the entry related to the specified key from the cache. The last
      * time used property of the entry is updated to extend the longevity of the
      * entry.
-     *
-     * @name module:Cache.get
-     * @function
      *
      * @param {string} key1
      * @param {string} [key2]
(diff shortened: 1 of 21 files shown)


gchoqueux:
I fix coverage code and I keep `Map()` to build cache. It must be more efficient but difficult to see with measuring in this context.

src/Core/Scheduler/Cache.js
```diff
@@ -24,24 +23,31 @@ let entry;
...
+ * @property {number} GEOMETRY - Shortcut for geometry resources. Time is 15 minutes.
+ * minutes.
+ */
+export const CACHE_POLICIES = {
+    INFINITE: Infinity,
+    TEXTURE: 900000,
+    GEOMETRY: 900000,
+};
+
+class Cache {
```
zarov, line 40: You need to change the documentation: remove all `@name/@function` from methods, and remove the `@module` as well. There are also some typo to fix.

src/Core/Scheduler/Cache.js
```diff
@@ -166,26 +183,51 @@ const Cache = {
...
+                v1.forEach((v2, j) => {
+                    if (this.lifeTime < time - v2.lastTimeUsed) {
+                        delete v1.get(j);
+                        v1.delete(j);
                     } else {
-                        for (const k in data[i][j]) {
-                            if (data[i][j][k].lifetime < time - data[i][j][k].lastTimeUsed) {
-                                delete data[i][j][k];
+                        v2.forEach((v3, k) => {
+                            const lastTimeUsed = v3.lastTimeUsed;
```
zarov: I don't think you need to do this, we only read it twice.

src/Core/Scheduler/Cache.js
```diff
@@ -24,24 +23,31 @@ let entry;
...
+ */
+export const CACHE_POLICIES = {
+    INFINITE: Infinity,
+    TEXTURE: 900000,
+    GEOMETRY: 900000,
+};
+
+class Cache {
+    constructor(lifetime = CACHE_POLICIES.INFINITE) {
+        this.lifeTime = lifetime;
```
zarov, line 45: 👍

src/Core/Scheduler/Cache.js
```diff
@@ -166,26 +183,51 @@ const Cache = {
...
-                            if (data[i][j][k].lifetime < time - data[i][j][k].lastTimeUsed) {
-                                delete data[i][j][k];
+                        v2.forEach((v3, k) => {
+                            const lastTimeUsed = v3.lastTimeUsed;
+                            if (this.lifeTime < time - v3.lastTimeUsed) {
+                                delete v2.get(k);
+                                v2.delete(k);
+                            } else {
+                                // Work for the moment because all flushed caches have 3 key!
+                                this.lastTimeFlush = Math.min(this.lastTimeFlush, lastTimeUsed);
```
zarov: Not true for `cacheStyle` in `Core/Style`. Also why not moved this somewhere else to avoid doing repetitive `Math.min` ? You could add a flag and check it at the end of the flush.

src/Layer/ColorLayer.js
```diff
@@ -56,6 +57,7 @@ class ColorLayer extends Layer {
      * view.addLayer(color);
      */
     constructor(id, config = {}) {
+        config.cacheLifeTime = config.cacheLifeTime == undefined ? CACHE_POLICIES.TEXTURE : config.cacheLifeTime;
```
zarov, line 60: Add it to the documentation, or maybe to the documentation of `Layer/Layer` ?

src/Core/Geographic/Extent.js
```diff
@@ -545,6 +545,14 @@ class Extent {
         }
     }
 
+    toKeyTag() {
+        if (CRS.isTms(this.crs)) {
+            return [this.zoom, this.row, this.col];
```
zarov: I don't really like arrays being created like this, but at the time I don't really see a better solution. Did you have other ideas while doing this ?

src/Core/Scheduler/Cache.js
```diff
@@ -166,26 +183,51 @@ const Cache = {
      *
      * @param {number} [time=Date.now()]
      */
-    flush: (time = Date.now()) => {
-        for (const i in data) {
-            if (data[i].lifetime < time - data[i].lastTimeUsed) {
-                delete data[i];
+    flush(time = Date.now()) {
+        if (this.lifeTime == CACHE_POLICIES.INFINITE ||
+            this.lifeTime > time - this.lastTimeFlush ||
```
zarov, line 171: Really good addition !

test/unit/dataSourceProvider.js
```diff
@@ -224,6 +224,7 @@ describe('Provide in Sources', function () {
         tile.parent = { pendingSubdivision: false };
         featureLayer.source.uid = 8;
         featureLayer.mergeFeatures = true;
+        featureLayer.cache.data.clear();
```
zarov, line 227: Curious: zhy is it needed ?

zarov, line 227: Curious: why is it needed ?

test/unit/extent.js
```diff
@@ -278,6 +278,31 @@ describe('Extent', function () {
...
     });
 
+    it('should convert TMS extent values to string', function () {
+        const withValues = new Extent('TMS:4326', 0, 1, 2);
+        const tostring = withValues.toString(',');
+        const toValues = tostring.split(',').map(s => Number(s));
+        assert.equal(toValues[0], withValues.zoom);
+        assert.equal(toValues[1], withValues.row);
+        assert.equal(toValues[2], withValues.col);
+    });
```
zarov, line 288: Why not compare it to `0,1,2` ?

zarov (commented):
What you can test is the memory footprint, by trying to put the keys in different orders. For example, I am not sure that `cacheTile.set(promiseGeometry, bufferKey, params.level, south);` is the same as `cacheTile.set(promiseGeometry, south, params.level, bufferKey);`. I know it was better for the previous cache, but it may be good to give another run here.

src/Core/Scheduler/Cache.js
gchoqueux: done

src/Core/Scheduler/Cache.js
gchoqueux: For `cacheStyle`, there's not `flush` because there is `if statement` in start function.

src/Layer/ColorLayer.js
gchoqueux: added

test/unit/extent.js
gchoqueux: To test `toString`  with inverse function.

test/unit/dataSourceProvider.js
gchoqueux: Because is the same layer used in all the tests, we need to clear cache to avoid return the same value.

gchoqueux: Because is the same layer used in all the tests, we need to clear cache to avoid return the same value.

src/Core/Scheduler/Cache.js
gchoqueux: removed

src/Core/Geographic/Extent.js
gchoqueux: I'll think about it.

src/Layer/ColorLayer.js
zarov: This is not present in your last commit

zarov (approved):
Looks good, please add documentation for `cacheLifeTime` and you can merge it.

gchoqueux:
I change keying  system with Source method

src/Layer/Layer.js
```diff
@@ -37,6 +37,8 @@ class Layer extends THREE.EventDispatcher {
      * contains three elements `name, protocol, extent`, these elements will be
      * available using `layer.name` or something else depending on the property
      * name.
+     * @param {Object} [config.cacheLifeTime=Infinity] - set life time value in cache.
```
zarov: This should be a `number`.

test/unit/extent.js
```diff
@@ -287,22 +287,6 @@ describe('Extent', function () {
         assert.equal(toValues[2], withValues.col);
     });
 
-    it('should convert TMS extent values to key tag', function () {
```
zarov: You could still have some unit test ;)

src/Source/WMSSource.js
```diff
@@ -135,6 +135,10 @@ class WMSSource extends Source {
         return URLBuilder.bbox(extent, this);
     }
 
+    requestToKey(extent) {
```
zarov: What about moving this as the default `requestToKey` in `Source` ? If not, we should have a mock method in `Source` to tell that the developer needs to implement this method, like the other methods.

zarov (commented):
Good idea, it makes more sense that the `Source` helps to store the data.

src/Layer/Layer.js
gchoqueux: done

src/Source/WMSSource.js
gchoqueux: moved

test/unit/extent.js
gchoqueux: added

zarov:
Looks good, you can merge it, and don't forget to squash !

PR #1395, opened 2020-05-20
fix(source): set the Style.minzoom correctly in VectorTilesSource

zarov:
The previous behavior was as follow:
- check the presence of `minzoom` for a specific layer
- if it is not present, use `2` has a default value
- insert this value to the `stops` array
The resulting problem is that if `minzoom` is not defined and the first
`stop` value was greater than 2, it was still adding a `2 to first stop`
step.

A solution has been added here, to consider that the `minzoom` is either
the specified minzoom, or whichever is the highest between 2 and the
littlest `stop`.

### DIFF
--- a/src/Source/VectorTilesSource.js
+++ b/src/Source/VectorTilesSource.js
@@ -91,8 +91,8 @@ class VectorTilesSource extends TMSSource {
                     checkStopsValues(layer.paint, stops);
 
                     let minStop = Math.min(...stops);
-                    // if none is found, default to 2
-                    minStop = (minStop == Infinity) ? 2 : minStop;
+                    // if none is found, default to 0
+                    minStop = (minStop == Infinity) ? 0 : minStop;
                     // compare to layer.minzoom and take the highest
                     minStop = (layer.minzoom == undefined) ? minStop : Math.max(layer.minzoom, minStop);
 
--- a/src/Source/VectorTilesSource.js
+++ b/src/Source/VectorTilesSource.js
@@ -86,12 +86,22 @@ class VectorTilesSource extends TMSSource {
                 } else if (ffilter(layer)) {
                     // TODO: add support for expressions
                     // https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions
-                    const stops = [];
-                    stops.push(layer.minzoom == undefined ? 2 : layer.minzoom);
+                    let stops = [];
                     checkStopsValues(layer.layout, stops);
                     checkStopsValues(layer.paint, stops);
+
+                    let minStop = Math.min(...stops);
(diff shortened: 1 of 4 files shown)


src/Source/VectorTilesSource.js
```diff
@@ -86,13 +86,23 @@ class VectorTilesSource extends TMSSource {
...
                     // https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions
-                    const stops = [];
-                    stops.push(layer.minzoom == undefined ? 2 : layer.minzoom);
+                    let stops = [];
                     checkStopsValues(layer.layout, stops);
                     checkStopsValues(layer.paint, stops);
+
+                    let minStop = Math.min(...stops);
+                    // if none is found, default to 2
+                    minStop = (minStop == Infinity) ? 2 : minStop;
```
gchoqueux: may be use 0 to min stop?

src/Source/VectorTilesSource.js
```diff
@@ -86,13 +86,23 @@ class VectorTilesSource extends TMSSource {
...
+                    // compare to layer.minzoom and take the highest
+                    minStop = (layer.minzoom == undefined) ? minStop : Math.max(layer.minzoom, minStop);
```
gchoqueux, line 97: `layer.minzoom` is never defined?

src/Source/VectorTilesSource.js
zarov: Yes, I don't know why I kept that

src/Source/VectorTilesSource.js
zarov: It can be defined, and undefined too https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#minzoom

gchoqueux: Actually, the minimum display zoom option is `Source.zoom.min`,  but it should not represent the layer display. We should do a review and add this option also in `Layer`.

zarov: Ok, I changed all `minzoom`/`maxzoom` to `zoom.min/zoom.max`.

gchoqueux: could you change `Layer#minzoom`/`maxzoom` to `Layer#zoom.min/zoom.max`?

zarov: This has nothing to do with `Layer/Layer`, as it is a mapbox thing here (see link above). So I can't change that.

gchoqueux: sorry, I complete muddle itowns layer and mapbox layer.

gchoqueux (approved):

PR #1396, opened 2020-05-20
refacto(style): get anchor position values instead of CSS

zarov:
Before, the `Style#getTextAnchorPositionInCSS()` method was returning a
string, to be used directly in CSS.

Now, it has been modified to `Style#getTextAnchorPosition()`, and
returns an array containing the `x` and `y` proportions to applied to a
`Label` for its anchor.

BREAKING CHANGE: `Style#getTextAnchorPositionInCSS` has been renamed to
`Style#getTextAnchorPosition`.

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -42,6 +42,18 @@ function getImageFromSprite(sprites, key) {
     return image;
 }
 
+const textAnchorPosition = {
+    left: [0, -0.5],
+    right: [-1, -0.5],
+    top: [-0.5, 0],
+    bottom: [-0.5, -1],
+    topRight: [-1, 0],
+    bottomLeft: [0, -1],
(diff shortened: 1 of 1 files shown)


src/Core/Style.js
```diff
@@ -42,6 +42,18 @@ function getImageFromSprite(sprites, key) {
     return image;
 }
 
+const textAnchorPosition = {
```
gchoqueux, line 45: could you use anchor string to define properties and remove the switch case.
```js
    //...
    'bottom-left': [0, -1],
    'bottom-right': [-1, -1],
    //...
```

zarov: Yes indeed, changing that

gchoqueux (approved):

PR #1402, opened 2020-06-02
fix(Cache): wrong key handling when the keys count is inferior at 3.

gchoqueux:
## Description
Wrong key handling when the keys count is inferior at 3.

## Motivation and Context
Use Cache with one or two keys.

### DIFF
--- a/src/Core/Scheduler/Cache.js
+++ b/src/Core/Scheduler/Cache.js
@@ -60,20 +60,20 @@ class Cache {
      */
     get(key1, key2, key3) {
         const entry_1 = this.data.get(key1);
-        if (entry_1 == undefined) {
-            // eslint-disable-next-line
-            return;
+        if (entry_1 == undefined) { return; }
+
+        if (entry_1.lastTimeUsed != undefined) {
+            entry = entry_1;
         } else {
             const entry_2 = entry_1.get(key2);
-            if (entry_2 == undefined) {
-                entry = entry_1;
+            if (entry_2 == undefined) { return; }
+
+            if (entry_2.lastTimeUsed != undefined) {
+                entry = entry_2;
             } else {
                 const entry_3 = entry_2.get(key3);
-                if (entry_3 == undefined) {
-                    entry = entry_2;
-                } else {
-                    entry = entry_3;
-                }
+                if (entry_3 == undefined) { return; }
+                entry = entry_3;
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Core/Scheduler/Cache.js
```diff
@@ -134,18 +137,28 @@ class Cache {
...
+        if (entry_1 !== undefined && !(entry_1 instanceof Map)) {
             delete this.data.get(key1);
             this.data.delete(key1);
-        } else if (entry.get(key2).get(key3) == undefined) {
-            delete entry.get(key2);
-            entry.delete(key2);
         } else {
-            delete entry.get(key2).get(key3);
-            entry.get(key2).delete(key3);
+            const entry_2 = entry_1.get(key2);
```
zarov, line 144: This will break if `entry_1` is `undefined`.

src/Core/Scheduler/Cache.js
```diff
@@ -61,16 +61,19 @@ class Cache {
     get(key1, key2, key3) {
         const entry_1 = this.data.get(key1);
         if (entry_1 == undefined) {
-            // eslint-disable-next-line
             return;
+        } else if (!(entry_1 instanceof Map)) {
```
zarov: Not a big fan of `instanceof`, won't it slow too much things ? But I don't really see an alternative, besides a ugly `entry_1.constructor.name == 'Map'`.

zarov (commented):
Could you fix the double key in `Style` in the same PR ? Thanks ! https://github.com/iTowns/itowns/blob/master/src/Core/Style.js#L372

src/Core/Scheduler/Cache.js
```diff
@@ -134,18 +137,28 @@ class Cache {
      * @param {string} [key3]
      */
     delete(key1, key2, key3) {
-        const entry = this.data.get(key1);
-        if (entry == undefined) {
-            throw Error('Please specify at least a key of something to delete');
-        } else if (entry.get(key2) == undefined) {
+        const entry_1 = this.data.get(key1);
+        if (entry_1 !== undefined && !(entry_1 instanceof Map)) {
```
zarov: You can ditch the `entry_1 !== undefined` because `undefined instanceof Map` results in `false`.

src/Core/Scheduler/Cache.js
gchoqueux: fixed

src/Core/Scheduler/Cache.js
gchoqueux: I replace by `entry_1.lastTimeUsed != undefined`

src/Core/Scheduler/Cache.js
gchoqueux: There's logical `not` with `instanceof`, we need to check if `entry_1 `is defined.

src/Core/Scheduler/Cache.js
```diff
@@ -134,18 +137,35 @@ class Cache {
...
-            delete this.data.get(key1);
-            this.data.delete(key1);
-        } else if (entry.get(key2).get(key3) == undefined) {
-            delete entry.get(key2);
-            entry.delete(key2);
-        } else {
-            delete entry.get(key2).get(key3);
-            entry.get(key2).delete(key3);
+        const entry_1 = this.data.get(key1);
+        if (entry_1 !== undefined) {
```
zarov: Could you do `entry_n === undefined` instead to avoid too much nesting ?

zarov (commented):
Looking good

zarov (approved):

PR #1408, opened 2020-06-05
fix(layer): hide labels when hiding LabelLayers

zarov:
When hiding a `LabelLayer`, hide its associated `Label`s using
`forceHidden`. A check is then added in `Label2DRenderer` to hide all
`Label`s is needed.

Also, when hiding the layer that `LabelLayer` is associated with (if
any), it hides the `LabelLayer` as well.

### DIFF
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -67,6 +67,7 @@ class Label extends THREE.Object3D {
         });
 
         this.isLabel = true;
+        this.forceHidden = false;
         this.coordinates = coordinates;
 
         this.projectedPosition = { x: 0, y: 0 };
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -67,7 +67,6 @@ class Label extends THREE.Object3D {
         });
 
         this.isLabel = true;
-        this.forceHidden = false;
         this.coordinates = coordinates;
 
         this.projectedPosition = { x: 0, y: 0 };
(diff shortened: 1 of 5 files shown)


src/Core/Label.js
```diff
@@ -67,6 +67,7 @@ class Label extends THREE.Object3D {
         });
 
         this.isLabel = true;
+        this.forceHidden = false;
```
gchoqueux: if layer.visible == false
then in `LabelLayer`, traverse all `labels/node.dom` to set hidden. 
or there is a main div to hide all labels?

zarov: I wanted to do this at first, but we don't want to hide all labels, if there are multiple layers with labels in it. So each label needs to be treated individually.

gchoqueux: Could add a div for each label layer?

zarov: Yes of course that is the perfect solution, sorry for missing this ! Should we still keep something like `forceHidden` though, as it will prevent from doing too much computation in the render loop ? Or maybe add a `layer` property to the label, so we can perform a `label.layer.visible` check in the renderer ?

gchoqueux: Or think about browsing the div tree for label rendering?

zarov: I tried to find a solution using the `div` tree, but I can't see how it could be possible, besides having a list of all displayed labels, or something like that as reference. I updated the PR with the removal of `forceHidden` and a `div` for each `LabelLayer`.

gchoqueux (approved):

PR #1409, opened 2020-06-05
example: add a vector tile example with style drag and drop

zarov:
This new example allows to drag and drop a valid style file, with the
sources of the stream specified in it, in order to test it. It is a
relatively simple tester, as nothing too complex has been tested with
it.

The DragNDrop is not used here, because it is only adapted for files
sources, not streams.

You can check this example here for the moment https://zarov.github.io/itowns/examples/vector_tile_dragndrop.html

### DIFF
--- a/examples/vector_tile_dragndrop.html
+++ b/examples/vector_tile_dragndrop.html
@@ -1,0 +1,82 @@
+<html>
+    <head>
+        <title>Itowns - Globe + vector-tiles</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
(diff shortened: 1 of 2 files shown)


examples/vector_tile_dragndrop.html
```diff
@@ -0,0 +1,82 @@
...
+
+            setupLoadingScreen(viewerDiv, view);
+
+            var currentLayer = {};
+
+            // Drag and drop handler
+            var fileReader = new FileReader();
+
+            var id = 0;
+            fileReader.onload = function onload(e) {
```
gchoqueux, line 42: it's possible to use `itowns.fetch.json()`?

zarov: Not when doing this, because the file doesn't have an URL to fetch from.

gchoqueux (approved):

Issue #1318, opened 2020-02-11
Feature and layer Style


gchoqueux:
There isn't simple method to get `feature/featureCollection/layer Style`. 
It's necessary to improve the architecture and expose methods to obtain the `style`.

zarov:
My suggestion:
- `FeatureCollection` has the master style
- `Feature` has its own style, with default values read from the `FeatureCollection` style
- same goes for `FeatureGeometry` which inherits from the style of `Feature` (and so `FeatureCollection`)

With this, only two things need to be done:
- add a `Style#clone()` method
- in the constructor of each of these types, add an option to initiate the style, `true` by default

Generally speaking, regarding these types, what we are also missing either a great tutorial concerning the creation of `FeatureCollection/Feature/FeatureGeometry`, or either a better API (it is too complicated to understand `bindNewGeometry/startSubGeometry/...` imho.

am2222:
Can this approach is able to apply to the diffrent layers? such as vector tiles?

zarov:
@am2222 what do you mean ?

jailln:
Created #1924 and #1923 as followups of this issue. Closing

PR #1411, opened 2020-06-08
Style inheritance

zarov:
Two commits, that try to add `Style` inheritance. Still need to fill in some proper documentation, but we can discuss it in the meantime.

#### feat(style): add Style inheritance

It is now possible to specify a `Style` parent for another `Style`. That
way, it is possible to define a global `Style` for a `ColorLayer`, while
defining one with property specific to a `FeatureGeometry` for example.
The `Style` of the `FeatureGeometry` will then only have the minimum
requirement, and it will read missing properties from its parent. It is
a dynamic reading, so it the parent values are changing, it is changing
as well for the child style.

This is a proposition for improvement in #1318.

BREAKING CHANGE: `Style#text.halo` is not available anymore, and each
property of it as been place in the `text` object, with the `halo`
prefix: `haloWidth`, `haloColor` and `haloBlur`.

### DIFF
--- a/examples/source_file_geojson_raster.html
+++ b/examples/source_file_geojson_raster.html
@@ -57,12 +57,12 @@
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
 
             var optionsGeoJsonParser = {
-                        buildExtent: true,
-                        crsIn: 'EPSG:4326',
-                        crsOut: view.tileLayer.extent.crs,
-                        mergeFeatures: true,
-                        withNormal: false,
-                        withAltitude: false,
+                buildExtent: true,
+                crsIn: 'EPSG:4326',
+                crsOut: view.tileLayer.extent.crs,
+                mergeFeatures: true,
+                withNormal: false,
+                withAltitude: false,
             };
 
             // Convert by iTowns
@@ -74,19 +74,21 @@
                         parsedData,
                     });
 
+                    var ariegeStyle = new itowns.Style({
+                        fill: {
+                            color: 'orange',
+                            opacity: 0.5,
+                        },
+                        stroke: {
+                            color:'white',
+                        },
+                    });
+
                     var ariegeLayer = new itowns.ColorLayer('ariege', {
                         name: 'ariege',
                         transparent: true,
-                        style: {
-                            fill: {
-                              color: 'orange',
-                              opacity: 0.5,
-                            },
-                            stroke: {
-                                color:'white',
-                            },
-                        },
                         source: ariegeSource,
+                        style: ariegeStyle,
                     });
 
                     return view.addLayer(ariegeLayer);
@@ -100,25 +102,25 @@
                         parsedData,
                     });
 
-                    return itowns.Fetcher.texture('images/cross.png').then(texture => {
-                        var pyoLayer = new itowns.ColorLayer('pyrenees-orientales', {
-                            name: 'pyrenees-orientales',
-                            transparent: true,
-                            style: {
-                                fill: {
-                                  pattern: texture.image,
-                                  opacity: 0.8,
-                                },
-                                stroke: {
-                                    color:'IndianRed',
-                                },
-                            },
-                            source: pyoSource,
-                        });
+                    var pyoStyle = new itowns.Style({
+                        fill: {
+                            pattern: 'images/cross.png',
+                            opacity: 0.8,
+                        },
+                        stroke: {
+                            color:'IndianRed',
+                        },
+                    });
 
-                        return view.addLayer(pyoLayer);
-                    }).then(FeatureToolTip.addLayer);
-                });
+                    var pyoLayer = new itowns.ColorLayer('pyrenees-orientales', {
+                        name: 'pyrenees-orientales',
+                        transparent: true,
+                        source: pyoSource,
+                        style: pyoStyle,
+                    });
+
+                    return view.addLayer(pyoLayer);
+                }).then(FeatureToolTip.addLayer);
 
             debug.createTileDebugUI(menuGlobe.gui, view);
         </script>
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import Extent from 'Core/Geographic/Extent';
 import Coordinates from 'Core/Geographic/Coordinates';
+import Style from 'Core/Style';
 
 function defaultExtent(crs) {
     return new Extent(crs, Infinity, -Infinity, Infinity, -Infinity);
@@ -167,6 +168,7 @@ export const FEATURE_TYPES = {
  * FeatureGeometry}.
  * @property {Extent?} extent - The extent containing all the geometries
  * composing the feature.
+ * @property {Style} style - The style of the Feature.
  */
 class Feature {
     /**
@@ -177,6 +179,8 @@ class Feature {
      * @param {boolean} [options.buildExtent] Build extent and update when adding new vertice.
      * @param {boolean} [options.withAltitude] Set vertice altitude when adding new vertice.
      * @param {boolean} [options.withNormal] Set vertice normal when adding new vertice.
+     * @param {Style} [options.style] The style to inherit when creating a new
+     * style for this feature.
      */
     constructor(type, crs, options = {}) {
         if (Object.keys(FEATURE_TYPES).find(t => FEATURE_TYPES[t] === type)) {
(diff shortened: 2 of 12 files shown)


examples/source_file_geojson_raster.html
```diff
@@ -68,6 +68,15 @@
             // Convert by iTowns
             itowns.Fetcher.json('https://raw.githubusercontent.com/gregoiredavid/france-geojson/master/departements/09-ariege/departement-09-ariege.geojson')
                 .then(function _(geojson) {
+                    optionsGeoJsonParser.style = new itowns.Style({
```
gchoqueux: If we add a style in layer instance, the inheritance will work?
I would prefer add the global style in Layer because is more intuitive

examples/source_file_geojson_raster.html
```diff
@@ -77,48 +86,44 @@
...
-                    return itowns.GeoJsonParser.parse(geojson, optionsGeoJsonParser);
-                }).then(function _(parsedData) {
-                    var pyoSource = new itowns.FileSource({
-                        parsedData,
+            var pyoStyle;
+            itowns.Fetcher.texture('images/cross.png')
+                .then(texture => {
+                    pyoStyle = new itowns.Style({
+                        fill: {
+                            pattern: texture.image,
```
gchoqueux: We could think to fetch image in Style

examples/source_file_geojson_raster.html
zarov: I agree, and this should be the case most of the times. 

However, here we have a special case: the data is parsed before being passed to its `FileSource`, using `GeoJsonParser`. In it, the `Style` assignment is done, but at this time it still don't know to which layer it will be attached, so it doesn't have a `parentStyle` passed in it. Normally, it is done in `DataSourceProvider`, when we don't parse the content ourselves. But here, we have to specify the `style` option.

gchoqueux: This is only for `FileSource` ?
It's possible to set `parent` style after parse data?

examples/source_file_geojson_raster.html
zarov: Done in 3d10626

examples/source_file_geojson_raster.html
zarov: This is only for `FileSource` yes. Not currently, but I'll try to think of a solution for it.

zarov: I added a proposition in 63ec609: it modifies `DataSourceProvider`, but to me it seems correct to do this.

src/Provider/DataSourceProvider.js
```diff
@@ -9,6 +9,33 @@ const error = (err, source) => {
     throw err;
 };
 
+function convertSourceData(data, extDest, layer) {
+    // Add missing parent style
+    if (data.isFeatureCollection) {
```
gchoqueux: Could you move this in new `FeatureCollection` function?

src/Provider/DataSourceProvider.js
```diff
@@ -9,6 +9,33 @@ const error = (err, source) => {
...
+            if (f.style && f.style.isStyle) {
+                parentStyle = f.style;
+
+                if (layer.style && !f.style.parent.isStyle) {
+                    f.style.parent = layer.style;
+                }
+            }
+
+            if (parentStyle) {
+                f.geometries.forEach((g) => {
```
gchoqueux: Could you move this in new `Feature` function?

src/Provider/DataSourceProvider.js
```diff
@@ -9,6 +9,33 @@ const error = (err, source) => {
...
+function convertSourceData(data, extDest, layer) {
+    // Add missing parent style
+    if (data.isFeatureCollection) {
+        data.features.forEach((f) => {
+            let parentStyle = layer.style;
+
+            if (f.style && f.style.isStyle) {
+                parentStyle = f.style;
+
+                if (layer.style && !f.style.parent.isStyle) {
```
gchoqueux: you check if `style.parent` is defined to avoid reassigning?

gchoqueux (commented):
May be add `inheritance` property to Style instead of `parent`.
because it avoids to traverse all child styles.

```js
const parentStyle = new Style(c);
const inheritance = { parent: parentStyle };
const style_a = new Style(a, inheritance);
const style_b = new Style(b, inheritance);

inheritance.parent = new Style(v);
```

> Add missing parent style

is only for `FileSource` case?
if so then you could do in `layer.whenReady` property.

src/Provider/DataSourceProvider.js
zarov: Yes, it was to set it only once if missing.

zarov:
> May be add `inheritance` property to Style instead of `parent`.
> because it avoids to traverse all child styles.
> 
> ```js
> const parentStyle = new Style(c);
> const inheritance = { parent: parentStyle };
> const style_a = new Style(a, inheritance);
> const style_b = new Style(b, inheritance);
> 
> inheritance.parent = new Style(v);
> ```

Why not, but I don't see where this could be used ? The `inheritance` object isn't kept, no ?

> > Add missing parent style
> 
> is only for `FileSource` case?
> if so then you could do in `layer.whenReady` property.

In this case yes, but I don't see how I could use `layer.whenReady` ?

gchoqueux:
> Why not, but I don't see where this could be used ? The inheritance object isn't kept, no ?

May be like this:
```js
style_b.inheritance.parent = new Style(v);
```

> In this case yes, but I don't see how I could use layer.whenReady ?

May be here:

https://github.com/iTowns/itowns/blob/master/src/Layer/Layer.js#L100-L103

zarov:
I updated from your comments, but I can't find a correct way to use the `inheritance` property, sorry.

src/Core/Feature.js
```diff
@@ -132,6 +132,12 @@ export class FeatureGeometry {
             }
         }
     }
+
+    setParentStyle(parentStyle) {
+        if (this.properties.style && this.properties.style.isStyle && !this.properties.style.parent.isStyle) {
```
gchoqueux: is necessary to check `isStyle`?

src/Core/Feature.js
```diff
@@ -223,6 +229,15 @@ class Feature {
     get geometryCount() {
         return this.geometries.length;
     }
+
+    setParentStyle(parentStyle) {
+        if (this.style && this.style.isStyle) {
+            this.style.parent = parentStyle;
```
gchoqueux: may be add inheritance when you instance  Geometry?

src/Layer/Layer.js
```diff
@@ -99,7 +99,13 @@ class Layer extends THREE.EventDispatcher {
             this._reject = rj;
         }).then(() => {
             if (this.source.fetchedData && !this.source.parsedData) {
-                return parseSourceData(this.source.fetchedData, this.source.extent, this);
+                return parseSourceData(this.source.fetchedData, this.source.extent, this).then((parsedData) => {
+                    // Set the missing style if it is a FileSource and a FeatureCollection
+                    if (parsedData.isFeatureCollection) {
```
gchoqueux: is necessary?

src/Parser/GeoJsonParser.js
```diff
@@ -42,54 +42,54 @@ const toFeature = {
...
+        geometry.properties.style = new Style({}, parentStyle).setFromGeojsonProperties(properties, feature.type);
 
         // Then read contour and holes
         for (let i = 0; i < coordsIn.length; i++) {
             this.populateGeometry(crsIn, coordsIn[i], geometry, setAltitude, feature);
         }
         feature.updateExtent(geometry);
     },
-    multi(type, feature, crsIn, coordsIn, filteringExtent, setAltitude, properties) {
+    multi(type, feature, crsIn, coordsIn, filteringExtent, setAltitude, properties, parentStyle) {
```
gchoqueux: could you pass parent style when you instance feature?

src/Core/Feature.js
zarov: It is not, I removed it, before we needed to do that but indeed it is not the case anymore.

src/Layer/Layer.js
zarov: No, removed it !

src/Parser/GeoJsonParser.js
zarov: Done

src/Core/Feature.js
zarov: I didn't, because `Feature.style` is always present, and we don't always want a style on a geometry (see vector tiles). However, I set the inheritance right after instancing it in `GeoJsonParser`.

gchoqueux (approved):

PR #1412, opened 2020-06-09
chore(test): use jsdom to mock the dom

zarov:
Instead of adding a wobbly mock system for unit testing, using jsdom
improves stability and adds more coverage.

Also added some tests for `GeoJsonParser`, `GpxParser` and `KMLParser`.


test/unit/bootstrap.js
```diff
@@ -1,104 +1,36 @@
...
+// hack to set the width and height of the image, to be used for the icon of labels
+Object.defineProperty(window.HTMLImageElement.prototype, 'width', { get: () => 10 });
+Object.defineProperty(window.HTMLImageElement.prototype, 'height', { get: () => 10 });
 
 class Renderer {
     constructor() {
-        this.domElement = new DOMElement();
-        this.domElement.parentElement = new DOMElement();
-        this.domElement.parentElement.appendChild(this.domElement);
+        this.domElement = document.createElement('div');
```
gchoqueux, line 29: It's possible to instance `Three#WebGLRenderer` with this new `document.createElement('div')`?

gchoqueux (commented):
Great improvement for unit tests! thanks
I just have one question, on `WebGLRenderer` that I'm gonna try on my own.

test/unit/bootstrap.js
zarov: I tried, but I couldn't find a library that mocks a `canvas.getContext('webgl')`, that's why I had to keep the previous `Renderer`.

gchoqueux: it doesn't seem to be working : 
```
Error: Error creating WebGL context
```
jsdom is similar to Puppetter, but doesn't support WebGL context.
I don't if we could use, simply, Puppetter instead if jsdom?

zarov: Puppeteer launch a whole browser each time and seems too big, especially with the whole async system. Also, I'm afraid we'll have to write our tests again. But I'll test it anyway !

zarov:
So I gave a try to doing it with Puppeteer, and it should be ok. I'm leaving some guidelines here, because I will not do it soon:
- [This guide explains how to run unit tests in the browser](https://medium.com/dailyjs/running-mocha-tests-as-native-es6-modules-in-a-browser-882373f2ecb0): all tests files should be called in a unique `test.js` file. 
- Minimal change should be done in test files, the only one that comes to mind is replacing the `assert` import by `import { assert } from 'chai';`.
- Launch the resulting `test.html` page using Puppeteer (see the `test/functional/bootstrap.js` file).
- To keep track of coverage, tests needs to still be launching in command line: it will be called on `test.js` and not on the html page.

gchoqueux (approved):
Thanks for the analysis. We'll get back to you.

Desplandis:
Obsolete.

PR #1416, opened 2020-06-12
chore: update packages.

### LINKED ISSUE #1241: Change the ThreeExtended source in config/prepare.js
See this comment: https://github.com/mrdoob/three.js/pull/16920#issuecomment-534257165

The `examples/js` folder will be removed at the beginning of 2020, so `config/prepare.js` needs to be adapted to use `examples/jsm` instead.

gchoqueux:
## Description

- [x] Three 117
    - [x] Files in the examples/js directoy are now deprecated and will be deleted in December 2020 (r124). It's recommended to start new projects with ES6 modules (examples/jsm) or migrate existing code to ES6 modules over the next months.
- [x] Mocha
   - [x] parallel tests https://mochajs.org/#parallel-tests
          **iTowns doesn't support parallel tests because it uses unique global variable.
          it's not really necessary because travis seems to be running the tests on a single process.**

Resolves #1241


.eslintignore
```diff
@@ -1,3 +1,14 @@
...
+examples/layers/
+src/ThreeExtended/
+test/hooks.js
+dist/
+node_modules/
+docs/out/
+docs/static/
+docs/tutorials/
+docs/tmpl/
+test/data/
```
zarov, line 11: What's with the weird things ? Doesn't eslint also look at `.gitignore` ?

config/replace.config.js
```diff
@@ -1,147 +1,31 @@
...
     to: patchGLTFLoader,
-}));
-
-// Patch _BATCHID in LegacyGLTFLoader
-const fileLegacyGLTFLoader = files.filter(f => path.basename(f, '.js') === 'LegacyGLTFLoader');
-const fromBatchIDLegacy = /case +'JOINT':\s+geometry\.addAttribute\( 'skinIndex', bufferAttribute \);\s+break;/g;
-const toLegacyGLTFLoader = 'case \'JOINT\':\ngeometry.addAttribute( \'skinIndex\', bufferAttribute );\nbreak;\ncase \'_BATCHID\':\ngeometry.addAttribute( \'_BATCHID\', bufferAttribute );\nbreak;';
+});
 
 patchs.push(() => replace({
```
zarov: No need to use `patchs` anymore if it is the only one in it.

zarov (commented):
You should keep `hooks.js` in `test/functional`, we could keep the history this way, no ? Or at least rename it to `hooks_functional.js` to avoid potential confusion in the future.

Also good to see that it simplifies things with the three modules.

.eslintignore
gchoqueux: > Doesn't eslint also look at .gitignore

I don't think so

zarov: It seems to have changed a bit (see https://eslint.org/docs/user-guide/migrating-to-7.0.0#default-ignore-patterns-have-changed), but ok with leaving this like it is

config/replace.config.js
gchoqueux: done

zarov (approved):

PR #1420, opened 2020-06-22
refactor(Provider): remove unnecessary fetchedData source case.

gchoqueux:
## Description
`FileSource#fetchedData`  case in `DataSourceProvider` could be removed because is handle by Layer Class.


zarov (approved):
Good catch !

PR #1422, opened 2020-06-24
fix wrong feature property calling in oriented image layer.

gchoqueux:
## Description
* **fix(oriented image):** wrong feature property calling.
* **test(Layer):** add `OrientedImageLayer` unit test.
* **refactor(oriented image):** remove unnecessary key to cache.

### DIFF
--- a/src/Source/OrientedImageSource.js
+++ b/src/Source/OrientedImageSource.js
@@ -44,7 +44,7 @@ class OrientedImageSource extends Source {
     }
 
     requestToKey(image) {
-        return [0, image.cameraId, image.panoId];
+        return [image.cameraId, image.panoId];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Source/OrientedImageSource.js
```diff
@@ -44,7 +44,7 @@ class OrientedImageSource extends Source {
     }
 
     requestToKey(image) {
-        return [0, image.cameraId, image.panoId];
+        return [image.cameraId, image.panoId];
```
zarov, line 47: Won't this break in `DataSourceProvider` ?

zarov (approved):

src/Source/OrientedImageSource.js
zarov: It doesn't seem to so ok with me

PR #1424, opened 2020-06-25
fix(gui): prevent other events when the GUI is used.

### LINKED ISSUE #1423: menu sliders moves the view
For example, opacity slider on Ortho Layer for 3d map example move the view

## Environment
* Version used: master 
* Browser Name and version: Chrome 83.0.4103.116
* Operating System and version (desktop or mobile): macOS 10.13.6

## Description
The bug is not visible in the initial placement for the [3D map example](http://www.itowns-project.org/itowns/examples/index.html#view_3d_map) (because there is nothing under the menu so the picking is not active), but if you zoom enough using the opacity slider moves the globe.

gchoqueux:
## Description
Prevent other events when the GUI is used.

fix #1423

### DIFF
--- a/examples/css/example.css
+++ b/examples/css/example.css
@@ -141,17 +141,20 @@ canvas {
     left: 0;
 }
 
-/* PLUGINS/GUI */
-.dg.main {
+#menuDiv {
     position: absolute;
-    top: 0;
-    z-index: 10;
+    left: 10px;
+    top: 0px;
+    z-index: 0;
 }
 
-@media (max-width: 600px) {
-    #menuDiv {
-        display: none;
-    }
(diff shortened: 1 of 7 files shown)


gmaillet:
In Potree examples, menu is now in the top right corner and hides the nb of points.
There are also some differences in the Stereo effects example, but I'm not sure that the master version is better.

gchoqueux:
> In Potree examples, menu is now in the top right corner and hides the nb of points.
There are also some differences in the Stereo effects example, but I'm not sure that the master version is better.

Thanks!
it's fixed

gmaillet (approved):

Issue #1170, opened 2019-07-01
3D Tiles cache ?


jailln:
Cache does not seem enabled by default for 3D Tiles.

For instance, if you zoom in/out in [this example](http://rict.liris.cnrs.fr/iTownsPlanar3DTiles/itowns/examples/planar_3dtiles.html) (warning: temporary link) you can see from the network debugging tool that the tiles are re-requested.

Is there a way to enable cache for 3D Tiles ?

zarov:
While it is not checked by iTowns, it should be checked in the browser cache. In your debugging tool, is the `Disable cache` option unchecked ?

Here is what I got

![image](https://user-images.githubusercontent.com/741255/60443952-85d72e80-9c1c-11e9-906a-48a515dd7496.png)

jailln:
Yes you're right, my question was not well expressed.

While it is cached, it seems to be unloaded from iTowns' memory and hence requested again (from cache) and then parsed again ? If this is correct; is there an option to avoid unloading these tiles and to just not render them ?

zarov:
I don't really know the 3dtiles part of iTowns, so I have no clue here. Maybe @jailln can help here ?

jailln:
Haha that's actually me asking the question ;)

How does it work for other types of data ?

zarov:
Damn sorry I got confused with another thread !

For most data in iTowns, apart from 3dtiles and pointclouds, it is set in the `Cache` in the [`DataSourceProvider`](https://github.com/iTowns/itowns/blob/master/src/Provider/DataSourceProvider.js#L126). For 3dtiles, I think tweaking things [here](https://github.com/iTowns/itowns/blob/master/src/Provider/3dTilesProvider.js#L316-L346) could be a solution.

jailln:
Ok, thanks. I guess we should use the Cache object as well in 3D Tiles then. 

A solution might also be to turn the loaded tiles contents visible or not using Threejs' [Object3D.visible](https://threejs.org/docs/index.html#api/en/core/Object3D.visible) instead of unloading them after a fixed delay that is defined [here](https://github.com/iTowns/itowns/blob/master/src/Provider/3dTilesProvider.js#L162).

gchoqueux:
> A solution might also be to turn the loaded tiles contents visible or not using Threejs' Object3D.visible instead of unloading them after a fixed delay that is defined here.

yes, and unloading children when theirs parent aren't visible.

mgermerie:
[...]

PR #1425, opened 2020-06-26
3D Tiles on 2.5D map

jailln:
## Description
This is a followup of PR #1034. 
Since I won't have time to apply the process you suggested to reduce the size of the tiles, I use a smaller dataset in this PR (only the 1st borough of Lyon). 

## Motivation and Context
I think it would be useful to have an example of 3D Tiles on plane in the codebase of iTowns. It demonstrates this capability and it is useful for us (and other people working with 3D Tiles on a planar view of itowns in their projects) for debugging purposes.

image

### DIFF
--- a/examples/3dtiles_25d.html
+++ b/examples/3dtiles_25d.html
@@ -1,0 +1,124 @@
+<html>
+    <head>
+        <title>Itowns - 3D Tiles on 2.5D map example</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv" class="viewer"></div>
+        <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="js/3dTilesHelper.js"></script>
+        <script type="text/javascript">
+            // Define projection that we will use (taken from https://epsg.io/3946, Proj4js section)
+            itowns.proj4.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+            // Define geographic extent: CRS, min/max X, min/max Y
+            var extent = new itowns.Extent( 'EPSG:3946',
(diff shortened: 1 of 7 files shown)


gchoqueux:
Thanks, for this example.
The b3dm file size are too heavy (`~3.6/5.5 MB`)

jailln:
Why is it a problem since the dataset is not from itowns but an external source ?

gchoqueux:
Yes indeed, but when I tested the demo, the 3d tiles took a long time to load and I thought there was a performance issue in itowns.

jailln:
Ok, unfortunatly I won't have time to work on reducing their size. Would you have time to apply the process you mentionned in #1034 to this dataset ? (or to give more specific instruction so I can apply it?). Otherwise I think we can close this PR.

gchoqueux:
Yes, I understand. I'll look into ways to quickly convert this data.

jailln:
Nice, thanks! Let me know if I can help.

gchoqueux:
### Options to lighten the b3dm file size
#### During mesh construction
* avoid to create a vertex for a face if any are already in the same position (use face index list).
* avoid to have more than 2 vertices to define a segment.
* avoid to build faces under the building because they will hidden by the ground.
* avoid to compute face normals.
#### The above options can also be applied with a optimization mesh tools 
* merge all vertices with the same position.
* merge all adjacent faces with a angle of 0.
* delete face normals.
#### Convert mesh to `glTF`
* doesn't export normals face.
* use Draco compression.
#### If you have already generated the `glTF`
* converting a `glTF` to `Draco glTF` [glTF-pipeline](https://github.com/CesiumGS/gltf-pipeline#converting-a-gltf-to-draco-gltf)
#### If you have already generated the `b3dm`
* apply draco with [optimizeB3dm](https://github.com/CesiumGS/3d-tiles-validator/tree/master/tools#optimizeb3dm) unfortunately it doesn't work if the embedded `glTF` is in version 2.0.
* if `glTF` is in version 2.0, creates a `glb` from a `b3dm` with [b3dmToGlb](https://github.com/CesiumGS/3d-tiles-validator/tree/master/tools#b3dmtoglb) and apply Draco with [glTF-pipeline](https://github.com/CesiumGS/gltf-pipeline#converting-a-gltf-to-draco-gltf)

**This last option could the short way to optimize yours `b3dm` files.**

I had written a wiki page [advices for generating small glTF files](https://github.com/iTowns/itowns/wiki/3D-Tiles#advices-for-generating-small-gltf-files).

jailln:
Thanks for these propositions! I applied draco compression (from the second process you suggested since our b3dm files are in gltf 2.0 version) and it greatly reduces the size of the b3dm files. However, the batch table (and therefore the semantic information) is lost during this process, so I removed the picking from the example.
Let me know if it's ok on your side; then I'll squash some commits and rebase on master.

gchoqueux (approved):
Nice!
I've noticed that b3dms are loaded several times as soon as they switch from invisible to visible. We should adjust the cache management.

gchoqueux:
@jailln Could you squash the commits?

jailln:
Squashed and rebased on master!

> I've noticed that b3dms are loaded several times as soon as they switch from invisible to visible. We should adjust the cache management.

Yes, I already opened issue #1170 on this subject

PR #1429, opened 2020-07-01
refactor(Vector Tiles): set style in vector tiles parser.

gchoqueux:
## Description

Move vector style parsing in `VectorTileParser`.

## Motivation
* Remove `onParsedFile` in Itowns.
* do all the geometry and style job in  `VectorStyleParse` and avoid a next step to add style.
* Remove Loop on all features to add Style.

### DIFF
--- a/src/Provider/DataSourceProvider.js
+++ b/src/Provider/DataSourceProvider.js
@@ -26,6 +26,7 @@ export function parseSourceData(data, extDest, layer) {
         withNormal: layer.isGeometryLayer !== undefined,
         withAltitude: layer.isGeometryLayer !== undefined,
         layers: source.layers,
+        styles: source.styles,
         style: layer.style,
     };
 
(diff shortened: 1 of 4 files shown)


src/Provider/DataSourceProvider.js
```diff
@@ -26,6 +26,7 @@ export function parseSourceData(data, extDest, layer) {
         withNormal: layer.isGeometryLayer !== undefined,
         withAltitude: layer.isGeometryLayer !== undefined,
         layers: source.layers,
+        styles: source.styles,
```
gmaillet, line 29: update the documentation for this styles option

gchoqueux: doc added

gmaillet (approved):

src/Provider/DataSourceProvider.js
```diff
@@ -26,6 +26,7 @@ export function parseSourceData(data, extDest, layer) {
...
         style: layer.style,
```
zarov, line 30: I am not a fan of having`styles` and `style` on the same object. Any idea on an alternative here ?

gchoqueux: yes, it will move in future PR.

PR #1431, opened 2020-07-02
refactor(GeoJsonParser): change filteringExtent parsing option 

gchoqueux:
## Description

Change `filteringExtent` parsing option  to `boolean` or `Extent`, instead of only `Extent`.

## Motivation and Context
This simplifies the functions in the providers.

### DIFF
--- a/src/Parser/GeoJsonParser.js
+++ b/src/Parser/GeoJsonParser.js
@@ -179,8 +179,8 @@ export default {
      * @property {string} crsOut - The CRS to convert the input coordinates
      * to.
      * @property {string} crsIn - Override the data CRS.
-     * @property {Extent} [filteringExtent] - Optional filter to reject
-     * features outside of this extent.
+     * @property {Extent|boolean} [filteringExtent=undefined] - Optional filter to reject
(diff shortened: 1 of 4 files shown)


src/Parser/GeoJsonParser.js
```diff
@@ -212,6 +211,15 @@ export default {
         options.withNormal = options.withNormal == undefined ? true : options.withNormal;
         options.withAltitude = options.withAltitude == undefined ? true : options.withAltitude;
 
+        let filteringExtent;
+        if (options.filteringExtent) {
+            if (typeof options.filteringExtent == 'boolean') {
```
gmaillet, line 216: what happens if filteringExtent is False?

gchoqueux: `options.filteringExtent` is checked [line 215 ](https://github.com/iTowns/itowns/pull/1431/files/0c89b3ec050694a59ccd785f4c18c22fe7fa8f52#diff-eb8b327101f78c0753970cd17d1bdcb0R215)then `filteringExtent` will be undefined, so there will be no filtering.

gmaillet (approved):

PR #1432, opened 2020-07-02
Up packages

gchoqueux:
### THREE.js v118
-  **WebGl 2.0 not yet support** follow #1443 
### Puppetter 5.1
- `Mouse#wheel `=> Fix me : `Mouse#wheel` doesn't work with functional test.


src/Renderer/c3DEngine.js
```diff
@@ -128,7 +129,7 @@ class c3DEngine {
 
     /**
      * return renderer THREE.js
-     * @returns {undefined|c3DEngine_L7.THREE.WebGLRenderer}
+     * @returns {undefined|c3DEngine_L7.THREE.WebGL1Renderer}
```
zarov: Nothing to do with this PR, but what is this `c3DEngine_L7` ?

zarov (approved):

PR #1434, opened 2020-07-07
feature(Label): text stroke width label supports value superior at 1.0.

gchoqueux:
## Description
Text stroke width label supports value superior at 1.0.
This support uses pseudo element `:before`.

### DIFF
--- a/examples/css/example.css
+++ b/examples/css/example.css
@@ -234,3 +234,32 @@ canvas {
     vertical-align: middle;
     text-shadow: 0px 0px 3px black;
 }
+
+:root {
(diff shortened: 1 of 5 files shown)


zarov (commented):
Looks good! In Chrome on some texts, I have a weird artifact, but I have no idea where it is coming from: 
Screenshot from 2020-07-07 16-27-57

I also have some problems in Firefox:
Screenshot from 2020-07-07 16-32-44

zarov (approved):
It works if I add the missing fonts in my use case (first screenshot), and if I remove a core modification I forgot I did ;) (second screenshot)

PR #1438, opened 2020-07-17
Enhancement handling of no data source.

gchoqueux:
## Description
* Fix error cache with vector tile source.

* Enhancement handling of no data source.
   There was no handling when the tile extent is outside source limit.
   The strategy is to find the highest level between the current level and level error.

### DIFF
--- a/src/Layer/LayerUpdateState.js
+++ b/src/Layer/LayerUpdateState.js
@@ -18,6 +18,16 @@ class LayerUpdateState {
         this.state = UPDATE_STATE.IDLE;
         this.lastErrorTimestamp = 0;
         this.errorCount = 0;
+        // lowestLevelError is lowest level with error.
+        //
+        // if lowestLevelError is Infinity, so there has been no error.
+        //
+        // if lowestLevelError isn't Infinity, so the strategy is to find the
+        // highest level between the current level and lowestLevelError.
+        // the dichotomy method is used to find it.
+        this.failureParams = {
+            lowestLevelError: Infinity,
+        };
     }
 
     canTryUpdate(timestamp = Date.now()) {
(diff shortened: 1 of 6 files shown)


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -124,6 +124,10 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
     const targetLevel = chooseNextLevelToFetch(layer.updateStrategy.type, node, destinationLevel, nodeLayer.level, layer, failureParams);
 
     if (targetLevel <= nodeLayer.level || targetLevel > destinationLevel) {
+        if (failureParams.lowestLevelError != Infinity) {
+            // this is the highest level found in case of error.
+            node.layerUpdateState[layer.id].noMoreUpdatePossible();
+        }
```
zarov, line 130: Shouldn't this be `if (failureParams.lowestLevelError < targetLevel)` or something like that ?

src/Layer/LayerUpdateState.js
```diff
@@ -53,18 +63,21 @@ class LayerUpdateState {
...
     }
 
+    noData(failureParams) {
+        this.state = UPDATE_STATE.IDLE;
+        this.failureParams.lowestLevelError = Math.min(failureParams.targetLevel, this.failureParams.lowestLevelError);
+    }
+
     failure(timestamp, definitive, failureParams) {
-        this.failureParams = failureParams;
+        this.failureParams.lowestLevelError = Math.min(failureParams.targetLevel, this.failureParams.lowestLevelError);
```
zarov: This will break if `failureParams.targetLevel` is not set (giving a `NaN`)

test/unit/layerupdatestrategy.js
```diff
@@ -0,0 +1,134 @@
...
+    const node = new TileMesh(geom, material, layer, extent, 19);
+    node.layerUpdateState[layer.id] = new LayerUpdateState();
+    const state = node.layerUpdateState[layer.id];
+    state.lowestLevelError = layer.source.zoom.max;
+    const failureParams = state.failureParams;
+    const history = [];
+    let LOADED = false;
+    let maxLevel = 13;
+
+    function isExist(level, currentLevel) {
```
zarov: Could you rename this method ? To understand easily the test you are doing ? Thanks

zarov (changes requested):

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: `failureParams.lowestLevelError` can't be **inferior**  to `targetLevel` because the **next level to fetch** (`targetLevel`) returns the level between **current level** and `lowestLevelError`

src/Layer/LayerUpdateState.js
gchoqueux: yes,i add test.

test/unit/layerupdatestrategy.js
gchoqueux: renamed

zarov (approved):

PR #1442, opened 2020-07-22
feature(layer): add max and min zoom properties Layer.

gchoqueux:
## Description
Add max and min zoom properties `Layer,` they define visibility interval.

### Motivation
* Add properties to handling visibility layer with zoom level.
* Avoid to use zoom Source to set starting display Feature Layer.


src/Layer/GeometryLayer.js
```diff
@@ -37,6 +37,10 @@ function traverse(obj, callback) {
  * @property {boolean} isGeometryLayer - Used to checkout whether this layer is
  * a GeometryLayer. Default is true. You should not change this, as it is used
  * internally for optimisation.
+ * @property {number} [zoom.max=Infinity] - this is the maximum zoom beyond which it'll be hidden.
```
zarov, line 40: Why can't someone set the maximum value ?

src/Process/FeatureProcessing.js
```diff
@@ -69,6 +69,13 @@ export default {
 
         const extentsDestination = node.getExtentsByProjection(layer.source.projection) || [node.extent];
 
+        const zoomDest = extentsDestination[0].zoom;
+
+        if (zoomDest != layer.zoom.min) {
```
zarov, line 74: Why not having a `zoomDest > layer.zoom.max || zoomDest < layer.zoom.max` too ?

src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -49,6 +49,11 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
     }
     const extentsDestination = node.getExtentsByProjection(layer.projection);
 
+    const zoom = extentsDestination[0].zoom;
+    if (zoom > layer.zoom.max || zoom < layer.zoom.min) {
+        return;
```
zarov, line 54: Is it relevant to add a `noMoreUpdatePossible()` call ?

src/Layer/TiledGeometryLayer.js
```diff
@@ -276,6 +276,10 @@ class TiledGeometryLayer extends GeometryLayer {
 
         for (const e of context.elevationLayers) {
             const extents = node.getExtentsByProjection(e.projection);
+            const zoom = extents[0].zoom;
```
zarov, line 297: Are all `extents` always the same zoom ?

src/Layer/Layer.js
```diff
@@ -86,6 +89,20 @@ class Layer extends THREE.EventDispatcher {
...
+        if (config.zoom != undefined) {
+            if (config.zoom.max != undefined && config.zoom.min != undefined) {
+                this.zoom = { max: config.zoom.max, min: config.zoom.min };
+            } else if (config.zoom.min != undefined) {
+                this.zoom = { max: Infinity, min: config.zoom.min };
+            } else if (config.zoom.max != undefined) {
+                this.zoom = { max: config.zoom.max, min: config.zoom.min };
+            } else {
+                this.zoom = { max: Infinity, min: 0 };
+            }
```
zarov: What about instead:

```suggestion
            this.zoom = { max: config.zoom.max, min: config.zoom.min || 0 };
            if (this.zoom.max == undefined) {
                this.zoom.max = Infinity;
            }
```

zarov (commented):
Thanks for removing this atrocious thing, especially for `GeometryLayer`!

src/Layer/GeometryLayer.js
gchoqueux: Yes!
one could imagine that the layer becomes invisible when this zoom level is reached.

src/Layer/TiledGeometryLayer.js
gchoqueux: yes

src/Layer/Layer.js
gchoqueux: yes, just check if `config.zoom` is defined!

jailln:
I have to admit I never understood how to use this property :) Could you precise in the doc to what this zoom is relative please ? (e.g. to the zoom of the camera ?) And also how to use it and maybe why you use certain values in the examples ?

gchoqueux:
@jailln Yes, I will add some documentations.

src/Layer/GeometryLayer.js
gchoqueux: For **`GeometryLayer`**, the properties **`zoom.max`** and **`zoom.min`** are used if the layer is attached to **`TileLayer`**. This is tile level that determines the **`GeometryLayer`** displaying. That is, if **`GeometryLayer.zoom.min == 6`**  and a **tile with level of 6** is displayed, then the features geometry, inside of this tile extent are displayed.

This could make sense if the features were fetched/parsed for each level but for the moment the data would be the same.

Finally,  I keep **`zoom.max`** at **`Infinity`**, because it's complex to stop displaying when all intermediates levels aren't loaded, (like raster).

src/Process/FeatureProcessing.js
gchoqueux: because the features are fetched for one tile level (see this [comment](https://github.com/iTowns/itowns/pull/1442#discussion_r459319482))

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: in this step, `layerUpdateState` isn't instanced

src/Layer/GeometryLayer.js
zarov: Sounds good to me. Maybe specify in this documentation that there is not a maximum level after it is hidden.

src/Process/LayeredMaterialNodeProcessing.js
zarov: Yes, but what about moving it after the initialization ?

src/Layer/Layer.js
gchoqueux: updated

src/Process/LayeredMaterialNodeProcessing.js
gchoqueux: because `MaterialLayer` must not be instanced because this node won't be never visible.

src/Layer/GeometryLayer.js
gchoqueux: doc updated

gchoqueux:
@jailln docs added

src/Layer/Layer.js
```diff
@@ -22,6 +22,14 @@ import Cache from 'Core/Scheduler/Cache';
  * @property {boolean} [labelEnabled=false] - Used to tell if this layer has
  * labels to display from its data. For example, it needs to be set to `true`
  * for a layer with vector tiles.
+ * @property {object} [zoom] - This property is used only the layer is attached to [TiledGeometryLayer]{@link TiledGeometryLayer}.
+ * The layer checks the tile zoom level to determine if the layer is visible in this tile.
+ *
+ * ![tiled geometry](/docs/static/images/wfszoommaxmin.jpeg)
```
zarov, line 29: Great addition ! Can you add a legend/description to tell that the tiles in question are red or orange, and the different levels between the top ones and the bottom one ? Then it is good with me, but please wait for an opinion of @jailln before merging

zarov (approved):

jailln:
> @jailln docs added

Thanks, that's a lot clearer! However, I still have two remaining questions: 

- How do you determine how much zoom levels a `TiledGeometryLayer` has and to what do they correspond visually ? Is it empiric ?

- In the documentation you say that this zoom level is used only if the layer is attached to a `TiledGeometryLayer`. However, I can't find out where it's (attaching the layer to a `TiledGeometryLayer`) done.  Could you point me to where it's done please ?

gchoqueux:
@jailln 

> How do you determine how much zoom levels a TiledGeometryLayer has and to what do they correspond visually ? Is it empiric ?

I added docs, to answer at your question.

> In the documentation you say that this zoom level is used only if the layer is attached to a TiledGeometryLayer. However, I can't find out where it's (attaching the layer to a TiledGeometryLayer) done. Could you point me to where it's done please ?

In [View#addLayer](http://www.itowns-project.org/itowns/docs/#api/View/View), the parameter `parentLayer` is the layer to which the layer will be attached. I will add doc.

jailln:
Great doc additions, it's much clearer for me now, thanks!

PR #1445, opened 2020-07-27
feat(parser): add LAS/LAZ files parser

zarov:
The parser uses the `@loaders.gl/las` package. It produces a
`THREE.BufferGeometry` that can be used to construct a `THREE.Points`.


gchoqueux (approved):

zarov:
@gchoqueux updated, and I added an example that allows drag and drop of a las or laz file.

gchoqueux:
super, you can merge

PR #1449, opened 2020-08-18
refacto: add an abstract PointCloud system

zarov:
## Description
Now, the Potree system is based on this new PointCloud system. It can now be used as a basis for other type of point clouds, like Entwine Point Tile.

**BREAKING CHANGE**: The signature of the `PotreeLayer` constructor has
changed.

## Motivation and Context
I am in the process of the integration of Entwine Point Tile format, and it is very similar to the current potree system we have. Instead of inheriting from it, I prefer to have a common point cloud basis to inherit from.

### DIFF
--- a/examples/potree_25d_map.html
+++ b/examples/potree_25d_map.html
@@ -61,8 +61,9 @@
                     source: new itowns.PotreeSource({
                         file: 'eglise_saint_blaise_arles.js',
                         url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/eglise_saint_blaise_arles',
-                    })
-                }, view);
+                    }),
+                    crs: view.referenceCrs,
+                });
 
                 // point selection on double-click
                 function dblClickHandler(event) {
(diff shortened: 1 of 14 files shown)


examples/potree_25d_map.html
```diff
@@ -61,8 +61,9 @@
                     source: new itowns.PotreeSource({
                         file: 'eglise_saint_blaise_arles.js',
                         url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/eglise_saint_blaise_arles',
-                    })
-                }, view);
+                    }),
+                    crs: view.referenceCrs,
```
gchoqueux: Need to use same property `Source/Layer` : `crs` or `projection`?

src/Layer/PotreeLayer.js
```diff
@@ -77,45 +27,25 @@ class PotreeLayer extends GeometryLayer {
...
+     * contains three elements `name, protocol, extent`, these elements will be
+     * available using `layer.name` or something else depending on the property
+     * name. See the list of properties to know which one can be specified.
+     * @param {string} [config.crs=ESPG:4326] - The CRS of the {@link View} this
+     * layer will be attached to. This is used to determine the extent of this
+     * layer.  Default to `EPSG:4326`.
      */
-    constructor(id, config, view) {
-        super(id, new THREE.Group(), config);
+    constructor(id, config) {
```
gchoqueux, line 46: good `view` remove!

examples/potree_25d_map.html
zarov: I used `Source#projection` in 2431a08

gchoqueux (approved):

PR #1450, opened 2020-08-24
feat(crs): add the support of WKT with proj4

zarov:


gchoqueux:
Proj4 is exported in itowns ?
If proj4 is exported, It's already possible to add proj4 definition ?

zarov:
It is indeed, I thought it was easier to have this wrapper instead of doing `itowns.proj4` ?

gchoqueux (commented):
Between
`itowns.CRS.registerWKT(name, wkt)`
and
`itowns.proj4.defs(name, wtk)`
I prefer the last because is more explicit

zarov:
Ok, no problem !

PR #1451, opened 2020-08-25
release 2.24.1

zarov:
A new release had to be done, because the previous one got the wrong version of THREE.

This could be avoided if we didn't included THREE inside iTowns.


gmaillet (approved):

gchoqueux:
the version is not changed in `Main.js` file.

zarov:
Yes my bad 😅

PR #1452, opened 2020-08-28
fix: label style

zarov:
Two commits:
- fix(style): move label halo management from CSS example to Style.js
The CSS is created once, and can be used outside the iTowns examples.

- fix(label): add compatibility with <!DOCTYPE html>
By setting the height and width of the label container, it should
display the labels when using <!DOCTYPE html>

@elias75015 this should solve your problems


elias75015:
I tried this PR locally and my "DOCTYPE html" problem is now solved :+1:  

Thanks for your work

src/Core/Style.js
```diff
@@ -7,6 +7,33 @@ const cacheStyle = new Cache();
 const inv255 = 1 / 255;
 const canvas = document.createElement('canvas');
 
+// add label halo management once
+const labelStyleSheet = document.createElement('style');
+labelStyleSheet.type = 'text/css';
+labelStyleSheet.innerHTML = `.itowns-stroke-single:before {
```
gchoqueux: Could you add inline file loading to handling this css configuration (with a adding `specific itowns_stroke_single_before.ccs` file)?
See `glsl/json` file loading and `babel-plugin-inline-import` (https://github.com/iTowns/itowns/blob/master/.babelrc#L5)

zarov: Good idea, done

gchoqueux (approved):

PR #1466, opened 2020-09-14
Fixes on labels

zarov:
A few fixes on labels:

- **fix(label): makes LabelLayer able to run alone**
Before, no configuration could be passed to it, like a source. This
fixes it.

- **feat(label): add a new class to Labels**
The class corresponds to `itowns-label-[layer-id]`. Documentation has
been added as well.

- **fix(label): allow label to only have an icon**

- **fix(label): labels don't jitter anymore**
There was a rounding that was done when comparing positions in the
renderer. This is still used for comparison, but the assigned value is
not the rounded one anymore.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -72,10 +72,11 @@ function _preprocessLayer(view, layer, parentLayer) {
     if (layer.isLabelLayer) {
         view.mainLoop.gfxEngine.label2dRenderer.registerLayer(layer);
     } else if (layer.labelEnabled) {
-        const labelLayer = new LabelLayer(`${layer.id}-label`, view.referenceCrs);
-        labelLayer.source = source;
-        labelLayer.style = layer.style;
-        labelLayer.zoom = layer.zoom;
+        const labelLayer = new LabelLayer(`${layer.id}-label`, {
+            source,
+            style: layer.style,
+            zoom: layer.zoom,
+        });
 
         layer.addEventListener('visible-property-changed', () => {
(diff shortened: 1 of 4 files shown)


src/Core/View.js
```diff
@@ -72,7 +72,7 @@ function _preprocessLayer(view, layer, parentLayer) {
     if (layer.isLabelLayer) {
         view.mainLoop.gfxEngine.label2dRenderer.registerLayer(layer);
     } else if (layer.labelEnabled) {
-        const labelLayer = new LabelLayer(`${layer.id}-label`, view.referenceCrs);
+        const labelLayer = new LabelLayer(`${layer.id}-label`, { crs: view.referenceCrs });
```
gchoqueux: `crs` will be always `view.referenceCrs`?

src/Layer/LabelLayer.js
```diff
@@ -86,7 +101,13 @@ class LabelLayer extends Layer {
 
                 const geometryField = g.properties.style && g.properties.style.text.field;
                 if (!geometryField && !featureField && !layerField) {
-                    return;
+                    // Check if there is an icon, with no text
```
gchoqueux, line 102: could you remove the checking in `Label`?

zarov: Done

src/Core/View.js
zarov: Yes, I change things in `LabelLayer` and removed that

gchoqueux (approved):

PR #1471, opened 2020-09-16
feat(point): add an intensity range for point clouds

zarov:
It is possible to specify the min and the max of the intensity range,
while updating the material in the same time to reflect those changes.

### DIFF
--- a/src/Layer/PointCloudLayer.js
+++ b/src/Layer/PointCloudLayer.js
@@ -61,6 +61,12 @@ function markForDeletion(elt) {
     }
 }
 
+function changeIntensityRange(layer) {
+    if (layer.material.intensityRange) {
+        layer.material.intensityRange.set(layer.minIntensityRange, layer.maxIntensityRange);
+    }
+}
+
 /**
  * The basis for all point clouds related layers.
  *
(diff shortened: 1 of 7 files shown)


src/Layer/PointCloudLayer.js
```diff
@@ -86,6 +93,12 @@ function markForDeletion(elt) {
  * `PointsMaterial`.
  * @property {number} [mode=MODE.COLOR] - The displaying mode of the points.
  * Values are specified in `PointsMaterial`.
+ * @property {number} [minIntensityRange=0] - The minimal intensity of the
+ * layer. Changing this value will affect the material, if it has the
+ * corresponding uniform.
+ * @property {number} [maxIntensityRange=65536] - The maximal intensity of the
```
gchoqueux: It isn't intuitive the maximal intensity value.
`65536` is the `256x256` result , so the `intensity` is coded with a **`uint16`**. 
What it's the `intensity` specification in `Potree`, `las/laz` and `ept`?

zarov: There are all in 16 bits, so that's why I choose this value.

gchoqueux: Could you normalize the intensity to have values between 0.0 and 1.0?
It's more intuitive.

gchoqueux: just normalize max and min

zarov: Done

gchoqueux: The material property could be normalized too, like Three js properties material.

gchoqueux: and use [BufferAttribute.normalized](https://threejs.org/docs/index.html#api/en/core/BufferAttribute.normalized) for intensity Attribute.

zarov: Done, I normalized it in LASParser, because this was the problem

gchoqueux: but `bufferAttribute` isn't normalized.

zarov: Sorry, I forgot to push the changes :sweat_smile:

zarov:
PR updated @gchoqueux

gchoqueux (approved):

PR #1477, opened 2020-09-22
Some labels fixes (see 1473)

### LINKED ISSUE #1473: Main report on Label bugs
Label feature is be introduce recently. It's work in progress to consolidate it.

 :warning: **Don't hesitate to add bugs in this issue if you identify one.**

# This issue lists identified bugs

## 1 | Breaking word without interest

- [x] **Fixed in  #1477**

<p align="center">
  
</p>

**Possible cause** on `Style.js` : **`domElement.style.overflowWrap = 'break-word';`**

## 2 | No horizon culling on label
- [x] **Fixed in  #1477**
<p align="center">
  
</p>

**Brussel** label is visible because the parent tile is visible. we should add a horizon culling on label with little level (see `GlobeLayer.js`)

## 3 | Label flickering
- [x] **Fixed in  #1477**
<p align="center">
  
</p>

Possible fix: Start culling by ordering labels with visibility and level.

## 4 | Bug culling label
- [x] **Fixed in  #1477**
<p align="center">
  
</p>

Possible cause: label is displayed while its parent is not visible.

zarov:
Three commits, to solve three problems listed in #1473 

- **fix(label): breaks word correctly**
It does not break in the middle of it anymore.

- **fix(label): correctly hide dom**
There was a problem for a specific situation. Let's consider that there
is a level 2 node, with four level 3 nodes. The L2 node contains no
DOM elements, while there are some in the L3 nodes. So the DOM in L3
have no parents, they are the first in the hierarchy.  So if we find L2
hidden, we don't go further, and thus inside L3 nodes, and don't hide
the DOM.
This commit fixes this behaviour, by allowing to go further in the
hierarchy until DOM elements are found.

- **fix(label): horizon culling**
Add a proper horizon culling mechanism, by splitting the GlobeLayer
culling to use it.

- **fix(label): reduce flickering by ordering label**
    It is not perfect yet, but the flickering has been reduced by taking into
    account the order the labels have.

- **test(label): add test for `Label2DRenderer`**

### DIFF
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -155,7 +155,14 @@ class Label extends THREE.Object3D {
         const elevation = DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z);
         if (elevation && elevation != this.coordinates.z) {
             this.coordinates.z = elevation;
+            this.updateHorizonCullingPoint();
             return true;
+        }
+    }
+
+    updateHorizonCullingPoint() {
+        if (this.horizonCullingPoint) {
+            this.getWorldPosition(this.horizonCullingPoint);
         }
     }
 }
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -92,6 +92,8 @@ class Label extends THREE.Object3D {
                 this.content.classList.add('itowns-stroke-single');
             }
             style.applyToHTML(this.content);
+        } else {
+            this.anchor = [0, 0];
         }
 
         this.zoom = {
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -35,6 +35,8 @@ if (document.documentElement.style.transform !== undefined) {
  * @property {THREE.Vector3} position - The position in the 3D world of the
  * label.
  * @property {Coordinates} coordinates - The coordinates of the label.
+ * @property {number} order - Order of the label that will be read from the
+ * style. It helps sorting and prioritizing a Label during render.
  */
 class Label extends THREE.Object3D {
     /**
@@ -100,6 +102,8 @@ class Label extends THREE.Object3D {
             min: style.zoom && style.zoom.min != undefined ? style.zoom.min : 2,
             max: style.zoom && style.zoom.max != undefined ? style.zoom.max : 24,
         };
+
+        this.order = style.order || 0;
     }
 
     /**
@@ -116,10 +120,12 @@ class Label extends THREE.Object3D {
             this.projectedPosition.x = X;
             this.projectedPosition.y = Y;
 
-            this.boundaries.left = x + this.offset.left;
-            this.boundaries.right = x + this.offset.right;
-            this.boundaries.top = y + this.offset.top;
-            this.boundaries.bottom = y + this.offset.bottom;
+            // 2 is a padding value, to avoid labels being too close to each
+            // other. This value has been choosen arbitrarily.
+            this.boundaries.left = x + this.offset.left - 2;
+            this.boundaries.right = x + this.offset.right + 2;
+            this.boundaries.top = y + this.offset.top - 2;
+            this.boundaries.bottom = y + this.offset.bottom + 2;
         }
     }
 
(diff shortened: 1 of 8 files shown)


zarov:
I added two commits:
- **test(label): add test for Label2DRenderer**
- **fix(label): reduce jittering by ordering label**
It is not perfect yet, but the jittering has been reduced by taking into
account the order the labels have.

gchoqueux:
>     * **fix(label): reduce jittering by ordering label**
>       It is not perfect yet, but the jittering has been reduced by taking into
>       account the order the labels have.

It reduces **flickering**.

src/Core/Label.js
```diff
@@ -114,10 +120,10 @@ class Label extends THREE.Object3D {
             this.projectedPosition.x = X;
             this.projectedPosition.y = Y;
 
-            this.boundaries.left = x + this.offset.left;
-            this.boundaries.right = x + this.offset.right;
-            this.boundaries.top = y + this.offset.top;
-            this.boundaries.bottom = y + this.offset.bottom;
+            this.boundaries.left = x + this.offset.left - 2;
```
gchoqueux, line 125: `2` is padding value?

src/Core/Label.js
```diff
@@ -155,9 +161,18 @@ class Label extends THREE.Object3D {
...
         if (elevation && elevation != this.coordinates.z) {
             this.coordinates.z = elevation;
+            if (this.horizonCullingPoint) {
+                this.updateHorizonCullingPoint();
+            }
             return true;
         }
     }
+
+    updateHorizonCullingPoint() {
```
gchoqueux, line 171: You could use the world position with  `getWorldPosition` instead of `horizonCullingPoint`?

src/Renderer/Label2DRenderer.js
```diff
@@ -21,7 +21,7 @@ class ScreenGrid {
     reset() {
         for (let i = 0; i < this.x; i++) {
             for (let j = 0; j < this.y; j++) {
-                this.grid[i][j] = false;
+                this.grid[i][j].splice(0, this.grid[i][j].length);
```
gchoqueux, line 30: add comment to explain splice.

src/Renderer/Label2DRenderer.js
```diff
@@ -50,16 +56,35 @@ class ScreenGrid {
 
         for (let i = minx; i <= maxx; i++) {
             for (let j = miny; j <= maxy; j++) {
-                if (this.grid[i][j]) {
-                    this.hidden.push(obj);
-                    return false;
+                if (this.grid[i][j].length > 0) {
+                    if (this.grid[i][j].some(l =>
+                        ((obj.boundaries.right > l.boundaries.left && obj.boundaries.right < l.boundaries.right)
```
gchoqueux: could you extract `containsBox` and `intersectsBox` methods

src/Renderer/Label2DRenderer.js
```diff
@@ -155,6 +190,8 @@ class Label2DRenderer {
         // visible extent, we can filter more labels.
         } else if (object.zoom.max <= currentMaxZoom || !extent.isPointInside(object.coordinates)) {
             this.grid.hidden.push(object);
+        } else if (object.parent.level < 4 && object.parent.layer.horizonCulling && object.parent.layer.horizonCulling(object.horizonCullingPoint)) {
```
gchoqueux, line 184: could you explain `object.parent.level < 4`?

gchoqueux (commented):
Nice Work!!

src/Core/Label.js
zarov: Yes, I added a comment about it.

src/Core/Label.js
zarov: Done, thanks for noticing this

src/Renderer/Label2DRenderer.js
zarov: Done

src/Renderer/Label2DRenderer.js
zarov: Done

src/Renderer/Label2DRenderer.js
zarov: Done

gchoqueux (approved):

PR #1487, opened 2020-09-30
Rename parsedData to features

gchoqueux:
## Description
Rename **`parsedData`** to **`features`**.

### DIFF
--- a/examples/js/plugins/CSVnVRTParser.js
+++ b/examples/js/plugins/CSVnVRTParser.js
@@ -21,8 +21,8 @@
  *         buildExtent: true,
  *         crsOut: 'EPSG:4326'
  *     });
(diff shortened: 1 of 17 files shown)


zarov (commented):
Looking good, although I am not sure about the `feature` name. Singular for `parsedData` made sense, but less for `feature`. Parsers return a `FeatureCollection` or an array of `Features` most of the time. What do you thing about `features`, plural ?

gchoqueux:
> Looking good, although I am not sure about the `feature` name. Singular for `parsedData` made sense, but less for `feature`. Parsers return a `FeatureCollection` or an array of `Features` most of the time. What do you thing about `features`, plural ?

Yes, it's done

zarov (approved):

PR #1490, opened 2020-10-01
More labels fixes

zarov:
- **chore(label):** remove unnecessary registerLayer
- **fix(label):** correctly remove labels when LabelLayer is deleted

### DIFF
--- a/src/Layer/LabelLayer.js
+++ b/src/Layer/LabelLayer.js
@@ -120,6 +120,7 @@ class LabelLayer extends Layer {
                 const label = new Label(content,
                     coord.clone(),
                     g.properties.style || f.style || this.style);
+                label.layerId = this.id;
 
                 if (f.size == 2) {
                     label.needsAltitude = true;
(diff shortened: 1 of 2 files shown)


src/Layer/LabelLayer.js
```diff
@@ -248,19 +249,32 @@ class LabelLayer extends Layer {
...
+        }
+    }
+
     delete() {
         this.domElement.parentElement.removeChild(this.domElement);
+
+        for (const root of this.parent.level0Nodes) {
+            root.traverse((obj) => {
+                if (obj.domElements && obj.domElements[this.id]) {
+                    this.removeLabelsFromNode(obj, [obj.children.filter(c => c.isLabel && c.layerId === this.id)]);
```
gchoqueux: the **`traverse`** method runs through the entire tree structure (all tiles and children). So you could remove label in **`traverse`**;

zarov: Yes indeed: but better, I ditched the traverse because I can limit the children I visit. See my last update !

gchoqueux (approved):

Issue #164, opened 2016-07-20
Edge/poor video card/Mobil Support


gchoqueux:
iTowns must  support maximum devices and navigator.
Some WebGL extensions is not supported :

- float texture (use in elevation)
- log depth

**WebGL 2.0 is now supported on mobile and Edge.**
As webgl 2.0 is also supported by itowns, 
Float texture and log depth are no longer a issue for these platforms.

vpicavet:
Please be more specific with the intended target : 
- we will definitely not support all version of IE
- some versions do not even have webgl
- the situation is quickly evolving with Edge replacing IE ( and other browsers rising)

Reminder of Web browsers shares : 
- IE 11 : 6%
- IE 9 : 2%

Source : W3 Counter

It would be interesting to have statistics from Geoportail over the last 3 months to be able to focus on the important things.

One option would be to have a very degraded mode for IE compatibility.

tbroyer:
Only IE11 continues to be supported by Microsoft (announced on January this year), and is the only one to natively support WebGL anyway. This is WebGL 0.94 though, when all other browsers support WebGL 1.

gchoqueux:
Only IE11, I think
IE11 support will open the support of certain cards and mobile

gchoqueux:
It's work on IE11 058517f6da10bf763a81ab3f5bd6ac272a096bc6

![ie](https://cloud.githubusercontent.com/assets/11291849/17019434/83c68eb8-4f3c-11e6-912f-91caa3405734.PNG)

gchoqueux:
PR merged #174 Best support Internet Explorer

lazd:
There is discussion of solutions to the lack of `EXT_frag_depth` on this three.js issue: https://github.com/mrdoob/three.js/issues/13047

jailln:
@LoicUTC @PhilippeMarc-IGN, what web browsers, in which versions and what platforms do we aim to support with itowns? Depending on this answer, we can close this issue or not.

Currently, itowns is known to be supported on last versions Chrome, Firefox, Safari and on these browsers for mobile. Others needs to be tested.

PhilippeMarc-IGN:
Browser market share : 

- Chrome 65 %
- Safari 19 %
- Edge 4 %
- Firefox 3 %

Edge is now Chromium based, I think it could be added.
For geoportail we support the last year versions, it seems to be a good option.

Desplandis:
[...]

PR #1492, opened 2020-10-05
[DRAFT] fix(WebGL): support Float elevation when OES_texture_float is unsupported.

gchoqueux:
Encode float buffer to RGBA Texture (UInt8Array).

Itowns support display mobile, tablet... #164


Desplandis:
We removed webgl1 support, OES_texture_float is no longer optional.
Can close this PR.

PR #1497, opened 2020-10-08
Fix remove layer with multi view

gchoqueux:
### Add `View.dispose()` method.

Method dispose all viewer objects
 * remove control
 * remove all layers
 * remove all frame requester
 * remove all events

I add this method to manage of multi-views, which is a consequence of the fix.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -350,9 +350,16 @@ class View extends THREE.EventDispatcher {
                 }
             }
 
-            // remove unused cache
-            const sameSource = this.getLayers(l => l.source.uid == layer.source.uid && l.crs == layer.crs);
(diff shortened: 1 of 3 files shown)


src/Core/View.js
```diff
@@ -201,8 +202,49 @@ class View extends THREE.EventDispatcher {
 
         // Focus needed to capture some key events.
         this.domElement.focus();
+
+        // push all viewer to keep source.cache
+        viewers.push(this);
     }
 
+    /**
+     * Dispose viewer before delete before you delete it.
```
gmaillet: ?

gmaillet (approved):
fix the dispose documentation

PR #1500, opened 2020-10-12
feature(Point): add classification support.

gchoqueux:
## Description
feature(Point): add classification support.

### DIFF
--- a/src/Renderer/PointsMaterial.js
+++ b/src/Renderer/PointsMaterial.js
@@ -12,11 +12,67 @@ export const MODE = {
     NORMAL: 3,
 };
 
+const white = new THREE.Color(1.0,  1.0,  1.0);
+
+/**
+ * Every lidar point can have a classification assigned to it that defines
+ * the type of object that has reflected the laser pulse. Lidar points can be
+ * classified into a number of categories including bare earth or ground,
+ * top of canopy, and water. The different classes are defined using numeric
+ * integer codes in the files.
+ *
+ * @property {object} category - category classification,
+ * @property {boolean} category.visible - category visibility,
+ * @property {string} category.name - category name,
+ * @property {THREE.Color} category.color - category color,
+ * @property {number} category.opacity - category opacity,
+ */
+// eslint-disable-next-line
+class /* istanbul ignore next */ Classification {}
+
+export const ClassificationScheme = {
+    DEFAULT: {
+        0: { visible: true, name: 'never classified', color: new THREE.Color(0.5,  0.5,  0.5), opacity: 1.0 },
+        1: { visible: true, name: 'unclassified', color: new THREE.Color(0.5,  0.5,  0.5), opacity: 1.0 },
+        2: { visible: true, name: 'ground', color: new THREE.Color(0.63, 0.32, 0.18), opacity: 1.0 },
+        3: { visible: true, name: 'low vegetation', color: new THREE.Color(0.0,  1.0,  0.0), opacity: 1.0 },
+        4: { visible: true, name: 'medium vegetation', color: new THREE.Color(0.0,  0.8,  0.0), opacity: 1.0 },
+        5: { visible: true, name: 'high vegetation', color: new THREE.Color(0.0,  0.6,  0.0), opacity: 1.0 },
+        6: { visible: true, name: 'building', color: new THREE.Color(1.0,  0.66, 0.0), opacity: 1.0 },
+        7: { visible: true, name: 'low point(noise)', color: new THREE.Color(1.0,  0.0,  1.0), opacity: 1.0 },
+        8: { visible: true, name: 'key-point', color: new THREE.Color(1.0,  0.0,  0.0), opacity: 1.0 },
+        9: { visible: true, name: 'water', color: new THREE.Color(0.0,  0.0,  1.0), opacity: 1.0 },
+        10: { visible: true, name: 'rail', color: new THREE.Color(0.8,  0.8,  1.0), opacity: 1.0 },
+        11: { visible: true, name: 'road Surface', color: new THREE.Color(0.4,  0.4,  0.7), opacity: 1.0 },
+        12: { visible: true, name: 'overlap', color: new THREE.Color(1.0,  1.0,  0.0), opacity: 1.0 },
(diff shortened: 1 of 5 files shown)


src/Renderer/PointsMaterial.js
```diff
@@ -12,6 +12,23 @@ export const MODE = {
...
+        4: { visible: true, name: 'medium vegetation', color: [0.0,  0.8,  0.0,  1.0] },
+        5: { visible: true, name: 'high vegetation', color: [0.0,  0.6,  0.0,  1.0] },
+        6: { visible: true, name: 'building', color: [1.0,  0.66, 0.0,  1.0] },
+        7: { visible: true, name: 'low point(noise)', color: [1.0,  0.0,  1.0,  1.0] },
+        8: { visible: true, name: 'key-point', color: [1.0,  0.0,  0.0,  1.0] },
+        9: { visible: true, name: 'water', color: [0.0,  0.0,  1.0,  1.0] },
+        12: { visible: true, name: 'overlap', color: [1.0,  1.0,  0.0,  1.0] },
+        DEFAULT: { visible: true, name: 'default', color: [0.3,  0.6,  0.6,  0.5] },
+    },
+};
```
zarov: You could add some more colors, as there is a few more classes defined as of now https://desktop.arcgis.com/en/arcmap/10.3/manage-data/las-dataset/lidar-point-classification.htm#ESRI_SECTION1_FDEB62EB5C2B463F86507C3EE0A7F441

src/Renderer/PointsMaterial.js
```diff
@@ -28,7 +76,22 @@ class PointsMaterial extends THREE.RawShaderMaterial {
         CommonMaterial.setUniformProperty(this, 'picking', false);
         CommonMaterial.setUniformProperty(this, 'opacity', this.opacity);
         CommonMaterial.setUniformProperty(this, 'overlayColor', options.overlayColor || new THREE.Vector4(0, 0, 0, 0));
-        CommonMaterial.setUniformProperty(this, 'intensityRange', options.intensityRange || new THREE.Vector2(0, 1));
+        CommonMaterial.setUniformProperty(this, 'intensityRange', intensityRange);
+
+        // add classification texture to apply classification lut.
+        const [width, height] = [256, 1];
```
zarov: What is the point of doing this, if you don't use any array ?

src/Renderer/Shader/PointsVS.glsl
```diff
@@ -79,6 +81,9 @@ void main() {
         vColor = vec4(i, i, i, opacity);
     } else if (mode == MODE_NORMAL) {
         vColor = vec4(abs(normal), opacity);
+    } else if (mode == MODE_CLASSIFICATION) {
+        vec2 uv = vec2(classification / 255.0, 0.5);
```
zarov: Could you add normalization of classification attributes, to avoid having to do that the division ?

src/Renderer/PointsMaterial.js
```diff
@@ -12,11 +12,59 @@ export const MODE = {
...
+     * @param      {THREE.Vector4}  [options.overlayColor=new THREE.Vector4(0, 0, 0, 0)]  overlay color.
+     * @param      {THREE.Vector2}  [options.overlayColor=new THREE.Vector2(0, 1)]  intensity range.
+     *
+     * @property {object}  classification -  points can be classified into a number of categories.
+     *
+     * @example
+     * // change color category classification
+     * const pointMaterial = new PointsMaterial();
+     * pointMaterial.classification[3].color.setStyle('red');
+     * pointMaterial.recomputeClassification();
```
zarov, line 67: This could be moved to the `recomputeClassification` method below

src/Renderer/PointsMaterial.js
```diff
@@ -28,7 +76,22 @@ class PointsMaterial extends THREE.RawShaderMaterial {
...
-        CommonMaterial.setUniformProperty(this, 'intensityRange', options.intensityRange || new THREE.Vector2(0, 1));
+        CommonMaterial.setUniformProperty(this, 'intensityRange', intensityRange);
+
+        // add classification texture to apply classification lut.
+        const [width, height] = [256, 1];
+        const data = new Uint8Array(width * 4);
+        const texture = new THREE.DataTexture(data, width, height, THREE.RGBAFormat);
+        texture.magFilter = THREE.NearestFilter;
+
+        this.classificationTexture = texture;
```
zarov: This can be optional, no ? You can use `texture` directly, and use `this.classificationLUT` instead of this.

src/Renderer/PointsMaterial.js
```diff
@@ -57,6 +120,73 @@ class PointsMaterial extends THREE.RawShaderMaterial {
...
+            const r = parseInt(255 * color.r, 10);
+            const g = parseInt(255 * color.g, 10);
+            const b = parseInt(255 * color.b, 10);
+            const a = visible ? parseInt(255 * opacity, 10) : 0;
+            const j = 4 * i;
+
+            if (data[j + 0] !== r) {
+                data[j + 0] = r;
+                valuesChanged = true;
+            }
```
zarov: Is it really necessary to do all of this ? As far as I know, this shouldn't be done a lot of time, so maybe you could simplify the code here ?

zarov (changes requested):
Looking good

src/Renderer/PointsMaterial.js
gchoqueux: removed

src/Renderer/PointsMaterial.js
gchoqueux: yes

src/Renderer/Shader/PointsVS.glsl
gchoqueux: yes

src/Renderer/PointsMaterial.js
```diff
@@ -12,11 +12,67 @@ export const MODE = {
     NORMAL: 3,
 };
 
+const white = new THREE.Color(1.0,  1.0,  1.0);
```
zarov, line 15: You could use `THREE.Color.NAMES` for predefined colors - just a suggestion.

zarov (approved):

src/Renderer/PointsMaterial.js
gchoqueux: simplified

PR #1505, opened 2020-10-20
WIP update doc

### LINKED ISSUE #1504: Error in documentation
## Your Environment

* Version used: release 2.26.0
* Browser Name and version: Mozilla Firefox for Ubuntu Version 81.0
* Operating System and version (desktop or mobile): Ubuntu (desktop)

## Context
The issue does not concern a bug. It concerns an possible error in the documentation.
On the webpage http://www.itowns-project.org/itowns/docs/#tutorials/Create-a-simple-globe, it is said to create a Coordinates object with a CRS argument value set to 'WGS84'. However, when I try this, I get the following error : _Uncaught Error: Invalid crs parameter value 'WGS84'_.
Reading the documentation for the Coordinates class, I find out that the CRS parameter can be placed as an EPSG number. Doing so - with 'EPSG:4326' instead of 'WGS84' - I do not get any error.

It might be possible that the documentation I mentioned earlier is out of date regarding the CRS attribute of Coordinates isntances.

mgermerie:
## Description

Documentation update :
- update of the page http://www.itowns-project.org/itowns/docs/#tutorials/Create-a-simple-globe : some parts of the code where outdated
- update of the documentation of WMTSSource class (`src/Source/WMTSSource.js`) : the crs property was not described as mandatory.

## Motivation and Context

Fixes the issues : https://github.com/iTowns/itowns/issues/1504#issue-724728861

### DIFF
--- a/docs/tutorials/Create-a-simple-globe.md
+++ b/docs/tutorials/Create-a-simple-globe.md
@@ -34,7 +34,10 @@ element of the page in order to be displayed.
 
 ```js
 var viewerDiv = document.getElementById('viewerDiv');
-var position = new itowns.Coordinates('WGS84', 2.35, 48.8, 25e6);
+var placement = {
+    coord: new itowns.Coordinates('EPSG:4326', 2.35, 48.8),
+    range: 25e6
+};
 var view = new itowns.GlobeView(viewerDiv, position);
 ```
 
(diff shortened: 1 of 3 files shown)


zarov (commented):
Thanks for this !

Could you rename your commits, following the Angular convention ? (see the [CONTRIBUTING](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md#pull-request-guidelines) guide). You can check the previous commits to get an idea of it. Here you should have something like `doc(tutorial)` and `doc(source)`.

gchoqueux (approved):

PR #1506, opened 2020-10-20
fix(source): move default format before calling parent constructor

zarov:
The default value was set, but it was making everything not working in the `Source` constructor.


gchoqueux (approved):

gchoqueux:
may be change test with Node 14 before merge this PR

gchoqueux:
thank you for all your contributions, which have been very important to the development of itowns!!!!

PR #1533, opened 2020-12-03
fix(PlanarControls): remove breaking change

mgermerie:
## Description

Modification of `PlanarControls` to prevent a breaking change after refactoring.

## Motivation and Context

After refactoring `PlanarView` in PR #1522, some now deprecated way of implementing `PlanarControls` with a `PlanarView` could cause errors with the controls. It is the case when `PlanarControls` are instantiated outside of `PlanarView`.
This PR adds a checking step in `PlanarControls` constructor to check if the associated view already has controls. If so, these controls are disposed.

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -83,7 +83,7 @@ const defaultOptions = {
     minZenithAngle: 0,
     maxZenithAngle: 82.5,
     focusOnMouseOver: true,
-    focusOnMouseClick: true,
+    focusOnMouseClick: false,
     handleCollision: true,
     minDistanceCollision: 30,
     enableSmartTravel: true,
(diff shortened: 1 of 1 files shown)


src/Controls/PlanarControls.js
```diff
@@ -940,6 +946,28 @@ class PlanarControls extends THREE.EventDispatcher {
         }
     }
 
+    /**
+     * Set the focus on view's domElement according to focus policy regarding MouseOver
+     *
+     * @ignore
+     */
+    onMouseOver() {
+        if (this.focusOnMouseOver) {
```
gchoqueux: move test in `addInputListeners`, because the test is checked  every mouse move

gchoqueux (approved):

PR #1536, opened 2020-12-07
fix(examples): fix key events in examples

mgermerie:
## Description

Fix key event in the following examples : 
- [2.5d map](http://www.itowns-project.org/itowns/examples/#view_25d_map)
- [Multiple globes](http://www.itowns-project.org/itowns/examples/#view_multiglobe)

## Motivation and Context

Key event were not detected because of a focus issue on the view.

### DIFF
--- a/src/Core/Prefab/PlanarView.js
+++ b/src/Core/Prefab/PlanarView.js
@@ -55,11 +55,11 @@ class PlanarView extends View {
         placement.heading = placement.heading || 0;
         placement.range = placement.range || max;
 
+        CameraUtils.transformCameraToLookAtTarget(this, camera3D, placement);
+
         if (!options.noControls) {
             this.controls = new PlanarControls(this, options.controls);
         }
-
-        CameraUtils.transformCameraToLookAtTarget(this, camera3D, placement);
 
         this.tileLayer = tileLayer;
     }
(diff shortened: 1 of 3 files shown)


src/Core/Prefab/PlanarView.js
```diff
@@ -55,12 +55,12 @@ class PlanarView extends View {
         placement.heading = placement.heading || 0;
         placement.range = placement.range || max;
 
+        CameraUtils.transformCameraToLookAtTarget(this, camera3D, placement);
```
gchoqueux, line 58: it's out of PR scope?

src/Core/View.js
```diff
@@ -201,6 +201,7 @@ class View extends THREE.EventDispatcher {
         this._pixelDepthBuffer = new Uint8Array(4);
 
         // Focus needed to capture some key events.
+        this.domElement.tabIndex = -1;
```
gchoqueux, line 205: Could you add comment (`tabIndex=-1`) inspired from https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

src/Core/Prefab/PlanarView.js
mgermerie: The placement of this line caused an issue with the `Y` key control in 2.5D planar example (the key is supposed to move the camera to its initial position, and without this change, the initial position was coordinates `(0, 0, 0)`).
Since it is related to key controls, I figured perhaps mistakenly this change could relate to the PR subject.

src/Core/View.js
gchoqueux: Could you add comment to merge the PR?

gchoqueux (approved):

PR #1544, opened 2021-01-06
example(label): add an example of label creation from mouse click

mgermerie:
## Description

Add an example of label creation from mouse event.
When user clicks some position on the map while holding `p` key pressed, a popup created from a custom `div` will appear at the clicked position.

### DIFF
--- a/examples/popup.html
+++ b/examples/popup.html
@@ -1,0 +1,134 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <title>Custom popup</title>
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="description">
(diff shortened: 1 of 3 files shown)


examples/label.html
```diff
@@ -0,0 +1,94 @@
+<!DOCTYPE html>
```
gchoqueux: Could you rename the file to popup?

examples/label.html
```diff
@@ -0,0 +1,94 @@
...
+                const geometry = feature.bindNewGeometry();
+                // fill geometry with point
+                geometry.startSubGeometry(1, feature);
+                geometry.pushCoordinates(coordinates, feature);
+                geometry.updateExtent();
+
+                feature.updateExtent(geometry);
+                collection.pushFeature(feature);
+
+                return Promise.resolve(collection);
```
gchoqueux: Promise isn't necessary

examples/label.html
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            function addFeatureFromMouseEvent(event) {
+                featureCoord.setFromVector3(view.getPickingPositionFromDepth(view.eventToViewCoords(event)));
+                console.log(featureCoord);
+                createFeatureAt(featureCoord).then(function _(features) {
+                    // the source of the feature layer
+                    let source = new itowns.FileSource({ features });
```
gchoqueux: replace let by const

gchoqueux (commented):
Could you change Label to have custom `content` parameter.
For user could be create custom div for the label.
So, change the example to use custom div.

mgermerie:
This PR is rebased on PR #1570.

PR #1546, opened 2021-01-07
Fix rendering issues 

gchoqueux:
## Description
* use opacity classification to filter points in all display modes.
* add tolerance constant to remove white color in ColorLayer.


src/Renderer/PointsMaterial.js
```diff
@@ -69,8 +70,12 @@ class PointsMaterial extends THREE.RawShaderMaterial {
     constructor(options = {}) {
         const intensityRange = options.intensityRange || new THREE.Vector2(0, 1);
         const oiMaterial = options.orientedImageMaterial;
+        const classification = options.classification || ClassificationScheme.DEFAULT;
```
mgermerie, line 73: For my own culture : what is the interest in creating a const for `classification` instead of setting `this.classification` value from `options.classification` directly ?

mgermerie (approved):

src/Renderer/PointsMaterial.js
gchoqueux: it's spesific case with the extending `THREE.RawShaderMaterial` because its options constructor  don't support a unknown parameter. It's the raison we delete unknown parameters before constructor. 
After constructor, we set `this.classification`.

PR #1555, opened 2021-01-17
Fix TMS projection/ crs

AugmentedGeoBae:
## Description
Changed TMSSource.js to accept both "source.projection" and "source.crs" as input key value

## Motivation and Context

The example source layers JSON had keys named "crs",
but TMSSource.js (and the other sources on which it's based) expects "projection"

TMSSource now accepts both "source.projection" and "source.crs",
if both "source.projection" and "source.crs" are null,
it still throws the error that a crs/ projection is required.
If "source.projection" is available it is used, otherwise "source.crs" is used.

### DIFF
--- a/src/Provider/URLBuilder.js
+++ b/src/Provider/URLBuilder.js
@@ -61,9 +61,15 @@ export default {
      * @return {string} the formed url
      */
     xyz: function xyz(coords, source) {
-        return subDomains(source.url.replace(/(\$\{z\}|%TILEMATRIX)/, coords.zoom)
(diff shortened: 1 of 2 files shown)


gchoqueux:
Since the version [v2.26.0](https://github.com/iTowns/itowns/releases/tag/v2.26.0), `projection` is deprecated.
`projection` can always be used, but it returns warning in console.
`projection` will be removed in a future.

You have a case where `projection` isn't used?

gchoqueux:
@murphysurveys re open if you need

PR #1567, opened 2021-01-25
fix(PlanarControls): fix zoom movement with an orthographic camera

mgermerie:
## Description

Correct an issue with `PlanarControl` in an orthographic camera context.

## Motivation and Context

Before PR, zooming in and out with an orthographic camera in a `PlanarView` would not target cursor position. With this PR, the zoom targets cursor position.

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -31,6 +31,7 @@ export const STATE = {
     PAN: 1,
     ROTATE: 2,
     TRAVEL: 3,
+    ORTHO_ZOOM: 4,
 };
 
 const vectorZero = new THREE.Vector3();
@@ -82,7 +83,7 @@ const defaultOptions = {
     rotateSpeed: 2.0,
     minPanSpeed: 0.05,
     maxPanSpeed: 15,
-    zoomTravelTime: 0.2,
+    zoomTravelTime: 0.2,  // must be a number
     zoomInFactor: 0.5,
     zoomOutFactor: 0.45,
     maxAltitude: 50000000,
@@ -176,6 +177,10 @@ class PlanarControls extends THREE.EventDispatcher {
             // Therefore, neutralizing by default the maxAltitude limit allows zooming out with an orthographic camera,
             // no matter its initial position.
             this.maxAltitude = Infinity;
+
+            // the zoom travel time (stored in `this.zoomTravelTime`) can't be `auto` with an orthographic camera
+            this.zoomTravelTime = typeof options.zoomTravelTime === 'number' ?
+                options.zoomTravelTime : defaultOptions.zoomTravelTime;
         } else {
             // enable rotation movements
             this.enableRotation = options.enableRotation === undefined ?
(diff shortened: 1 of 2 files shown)


src/Controls/PlanarControls.js
```diff
@@ -82,7 +83,7 @@ const defaultOptions = {
     rotateSpeed: 2.0,
     minPanSpeed: 0.05,
     maxPanSpeed: 15,
-    zoomTravelTime: 0.2,
+    zoomTravelTime: 0.2,  // must be a number
```
gchoqueux, line 86: why you this comment?

src/Controls/PlanarControls.js
```diff
@@ -501,51 +510,81 @@ class PlanarControls extends THREE.EventDispatcher {
...
+        }
+    }
+
+    /**
+     * Handle the animated zoom change for an orthographic camera, when state is `ZOOM`.
+     *
+     * @param   {number}    dt  the delta time between two updates in milliseconds
+     * @ignore
+     */
+    handleZoom(dt) {
```
gchoqueux: could you rename to `handleZoomOrtho`

src/Controls/PlanarControls.js
```diff
@@ -501,51 +510,81 @@ class PlanarControls extends THREE.EventDispatcher {
...
-        // Zoom OUT
-        } else if (delta < 0 && this.maxAltitude > this.camera.position.z) {
-            // increase the camera field of view if the camera is orthographic
+        if (delta > 0 || (delta < 0 && this.maxAltitude > this.camera.position.z)) {
+            const zoomFactor = delta > 0 ? this.zoomInFactor : -1 * this.zoomOutFactor;
+
+            // change the camera field of view if the camera is orthographic
             if (this.camera.isOrthographicCamera) {
+                // switch state to STATE.ZOOM
+                this.state = STATE.ZOOM;
```
gchoqueux: if this is just for camera ortho, could you rename  to STATE.ORTHO_ZOOM

src/Controls/PlanarControls.js
```diff
@@ -501,51 +510,81 @@ class PlanarControls extends THREE.EventDispatcher {
...
+
+            // new camera position
+            this.camera.position.x += pointUnderCursor.x - vect.x;
+            this.camera.position.y += pointUnderCursor.y - vect.y;
+
+            this.camera.updateMatrixWorld(true);
+        }
+
+        // completion test
+        if (travelAlpha === 1) {
```
gchoqueux: this test already exists in other function?

src/Controls/PlanarControls.js
```diff
@@ -890,6 +932,9 @@ class PlanarControls extends THREE.EventDispatcher {
             case STATE.TRAVEL:
                 this.view.domElement.style.cursor = 'wait';
                 break;
+            case STATE.ZOOM:
+                this.view.domElement.style.cursor = 'wait';
+                break;
```
gchoqueux: could you replace by 
```js
            case STATE.TRAVEL:
            case STATE.ZOOM:
                this.view.domElement.style.cursor = 'wait';
                break;
```

src/Controls/PlanarControls.js
mgermerie: I added this comment so that the default value can't be changed to `auto` (which is a relevant value for `zoomTravelTime` with a perspective camera).
Was this value to be switched to `'auto'`, the call of `initiateZoom()` shall trigger an error : it would set const `travelDuration` to `'auto'`, then try to divide a number by `travelDuration`.
This comment is a warning so that the default value is not changed to `'auto'` in the future.

src/Controls/PlanarControls.js
mgermerie: Indeed, it is just for orthographic camera. I shall rename it.

src/Controls/PlanarControls.js
mgermerie: This test is also computed in `handleTravel`. Do you prefer it to be factorized as a method (something like `testAnimationEnd()`) ?

gchoqueux (approved):

PR #1570, opened 2021-01-29
feature(Feature): add usefull methods to Feature

mgermerie:
## Description

Add a method to `Features` to add `FeatureGeometries` of type `point` from a list of `Coordinates`.

### DIFF
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -290,6 +290,7 @@ export class FeatureCollection {
      *
      * @param      {string}  crs      The crs projection.
      * @param      {FeatureBuildingOptions|Layer}  options  The building options .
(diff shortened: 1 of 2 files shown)


src/Core/Feature.js
```diff
@@ -249,6 +257,42 @@ class Feature {
...
+
+            pGeometry.startSubGeometry(1, this);
+            pGeometry.pushCoordinates(coordinates[i], this);
+            pGeometry.updateExtent();
+
+            pGeometry.properties.style = this.style;
+
+            this.updateExtent(pGeometry);
+        }
+    }
```
gchoqueux: Could you remove properties parameter?

Why you add the point twice?

PR #1571, opened 2021-02-01
fix(PlanarControls): fix pan movement

mgermerie:
## Description
Fix pan movement in a `PlanarView` context.

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -297,15 +297,6 @@ class PlanarControls extends THREE.EventDispatcher {
      * @ignore
      */
     update(dt, updateLoopRestarted) {
-        // We test if camera collides to the geometry layer or is too close to the ground, and adjust its altitude in
-        // case
-        if (this.handleCollision) { // check distance to the ground/surface geometry (could be another geometry layer)
-            this.view.camera.adjustAltitudeToAvoidCollisionWithLayer(
-                this.view,
-                this.view.tileLayer,
-                this.minDistanceCollision,
-            );
-        }
         // dt will not be relevant when we just started rendering. We consider a 1-frame move in this case
         if (updateLoopRestarted) {
             dt = 16;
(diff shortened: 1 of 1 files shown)


gchoqueux:
there's a [failed test](https://github.com/iTowns/itowns/pull/1571/checks#step:8:493)

src/Controls/PlanarControls.js
```diff
@@ -396,30 +400,22 @@ class PlanarControls extends THREE.EventDispatcher {
...
+        this.camera.position.x = newPosition.x;
+        this.camera.position.y = newPosition.y;
 
-        // lateral movement (local x axis)
-        vect.set(panSpeed * -1 * deltaMousePosition.x, 0, 0);
-        this.camera.position.copy(this.camera.localToWorld(vect));
-
-        // vertical movement (world z axis)
-        const newAltitude = this.camera.position.z + panSpeed * deltaMousePosition.y;
+        // compute ground level under the camera
```
mgermerie: This part (checking if destination position does not collide with layer) might be redundant with the check which is done in `update` method. However, if we just test it in `update` method, the camera is able to go under a layer, then move back above it when `update` is called (which is at the next image refreshment).

gchoqueux: Could you try to remove altitude checking in `handlePanMovement` and move handling Collision to the `udapte` end?

mgermerie: Indeed, moving collision handling to the end of `update` method works as expected.

src/Controls/PlanarControls.js
```diff
@@ -396,31 +396,9 @@ class PlanarControls extends THREE.EventDispatcher {
...
-        // vertical movement (world z axis)
-        const newAltitude = this.camera.position.z + panSpeed * deltaMousePosition.y;
-
-        // check if altitude is valid
-        if (this.groundLevel < newAltitude && newAltitude < this.maxAltitude) {
-            this.camera.position.z = newAltitude;
-        }
+        vect.set(-deltaMousePosition.x, deltaMousePosition.y, 0);
+        const newPosition = new THREE.Vector3().copy(this.camera.localToWorld(vect));
+        this.camera.position.copy(newPosition);
```
gchoqueux: could you try?
```js
vect.set(-deltaMousePosition.x, deltaMousePosition.y, 0);
this.camera.localToWorld(vect);
this.camera.position.copy(vect);
```

gchoqueux (approved):

PR #1577, opened 2021-02-02
Introduce Raster Tile

gchoqueux:
## Description
* `RasterNode` is `MaterialLayer` refactoring.
* 2 new Nodes:  `RasterColorNode` and `RasterElevationNode` extends  `RasterNode`.
* `RasterElevationNode` handles min and max computation and emits elevation updating to TileMesh, it is then applied to his bbox.
* `RasterElevationNode` handles  `noDataValue` replacing with init `Textures`.

## Motivation
This PR is a refactoring to move specificities processes in layer node.
The new layer node are `RasterNode`, `RasterColorNode` and `RasterElevationNode`.
The  max/min update and replace no data  process are moved from `updateLayeredMaterialNodeElevation` in `RasterElevationNode`.

The `updateLayeredMaterialNodeElevation` refactoring makes it more generic and is closer to `updateLayeredMaterialNodeImagery`.  

The next step is to move the last color and elevation process to have only one raster layer process instead of `updateLayeredMaterialNodeElevation` and `updateLayeredMaterialNodeImagery`.


mgermerie (commented):
For my knowledge (and perhaps documentation too), could you explain how node are used in iTowns ?

gchoqueux:
> For my knowledge (and perhaps documentation too), could you explain how node are used in iTowns ?

Yes, I added a documentation.

src/Renderer/MaterialLayer.js
```diff
@@ -142,4 +100,85 @@ class MaterialLayer {
...
+    }
+
+    replaceNoDataValueFromTexture(texture) {
+        const nodatavalue = this.layer.noDataValue;
+        if (nodatavalue == undefined) {
+            return;
+        }
+        // replace no datat value with parent texture value.
+        const parentTexture = this.textures[0];
+        const parentDataElevation =  parentTexture && parentTexture.image && parentTexture.image.data;
```
mgermerie: You left two spaces after the `=`.

src/Renderer/MaterialLayer.js
```diff
@@ -142,4 +100,85 @@ class MaterialLayer {
     }
 }
 
-export default MaterialLayer;
+export default RasterNode;
+
+export class RasterColorNode extends RasterNode {
+    constructor(material, layer) {
+        super(material, layer);
+        this.fx = layer.fx || 0;
```
mgermerie: I couldn't find information about `layer.fx` property. Maybe it could be added to the doc, since it is used as a parameter in `ColorLayer` constructor in example _vector_tiles_raster_3D.html_.

src/Renderer/MaterialLayer.js
```diff
@@ -142,4 +100,85 @@ class MaterialLayer {
...
+        };
+
+        let scaleFactor = 1.0;
+
+        // Define elevation properties
+        if (layer.useRgbaTextureElevation) {
+            defaultEle.mode = ELEVATION_MODES.RGBA;
+            defaultEle.zmax = 5000;
+            throw new Error('Restore this feature');
+        } else if (layer.useColorTextureElevation) {
```
mgermerie, line 131: Same as `layer.fx` : the parameter `useColorTextureElevation` is used in example _3dtiles_25d.html_ but it is not described in documentation.

src/Renderer/MaterialLayer.js
```diff
@@ -142,4 +100,85 @@ class MaterialLayer {
...
+    replaceNoDataValueFromTexture(texture) {
+        const nodatavalue = this.layer.noDataValue;
+        if (nodatavalue == undefined) {
+            return;
+        }
+        // replace no datat value with parent texture value.
+        const parentTexture = this.textures[0];
+        const parentDataElevation =  parentTexture && parentTexture.image && parentTexture.image.data;
+        const dataElevation = texture.image && texture.image.data;
+        if (dataElevation && parentDataElevation && !checkNodeElevationTextureValidity(dataElevation, nodatavalue)) {
```
mgermerie, line 180: If I got it right, `checkNodeElevationTextureValidity` only checks if values at the corners of the elevation texture are significant (superior to `nodatavalue`).

Is the reason why this check is only performed on corners because the most extreme elevation values will be at those corners for the majority of cases ?

src/Renderer/MaterialLayer.js
```diff
@@ -1,62 +1,33 @@
...
-        get: () => _value,
-        set: (value) => {
-            if (_value !== value) {
-                _value = value;
-            }
-        },
-    });
-}
 
-class MaterialLayer {
```
mgermerie, line 22: `MaterialLayer` is still called in some tests :

- _test/unit/demutils.js_ at line 62
- _test/unit/layeredmaterialnodeprocessing.js_ at line 54
- _test/unit/layerupdatestrategy.js_ at line 57

src/Renderer/MaterialLayer.js
gchoqueux: removed

src/Renderer/MaterialLayer.js
gchoqueux: yes,I add documentation.

I move default value in ColorLayer to avoid checking in RasterColorNode.

src/Renderer/MaterialLayer.js
gchoqueux: yes,I add documentation.

src/Renderer/MaterialLayer.js
gchoqueux: We never see yet a case where there are `noDataValue` without be in corners, too.

src/Renderer/MaterialLayer.js
gchoqueux: yes, thanks

src/Layer/ElevationLayer.js
```diff
@@ -61,6 +72,28 @@ class ElevationLayer extends RasterLayer {
...
+    setupRasterNode(node) {
+        const rasterElevationNode = new RasterElevationNode(node.material, this);
+
+        node.material.addLayer(rasterElevationNode);
+        node.material.setSequenceElevation(this.id);
+        // bounding box initialisation
+        node.setBBoxZ(rasterElevationNode.min, rasterElevationNode.max, this.scale);
+
+        // listen elevation updating
+        rasterElevationNode.addEventListener('updatedElevation', () =>
```
gchoqueux: don't forget dispose the event

gchoqueux: done

mgermerie (approved):
Could you rename file `MaterialLayer.js` to `RasterTile.js` ?

gchoqueux:
> Could you rename file `MaterialLayer.js` to `RasterTile.js` ?

yes

PR #1590, opened 2021-02-16
fix: standardize zoom factor in PlanarControls and GlobeControls

### LINKED ISSUE #1586: zoomFactor parameters inconsistance in planarView
## Context

`zoomInFactor` and `zoomOutFactor` are two optional parameters when instantiating `PlanarControls`. They are supposed to quantify the 'amount' of zoom which is performed when a user roll the mouse wheel.

If these parameter are set to the same values, the effect they will have on the zoom will differ according to the camera type (the example of `zoomInFactor` is given here) :
- When a user wheels the mouse with a `PerspectiveCamera`, the camera performs a fraction of a travel between its position before wheeling and the position on the ground pointed by the mouse. The `zoomInFactor` describes this fraction. Therefore, if a user wants to double the size of the world on a single mouse wheel movement, the `zoomInFactor` shall be set to `0.5`.
- If the mouse is wheeled with an `OrthographicCamera`, the camera is not moved, but its `zoom` attribute is multiplied by `(1 + zoomInFactor)`. The `zoom` attribute of a camera scales the world seen by the camera : if the `zoom` is multiplied by `n`, the size of the objects visible will be multiplied by `n` too. In this context, having a zoom which doubles the size of the world on a single mouse wheel movement requires `zoomInFactor` value to be set to `1`.

## Expected Behavior
The zoom for an `OrthographicCamera` and a `PerspectiveCamera` shall be standardized so that the same value of `zoomInFactor` and `zoomOutFactor` produce the same visual effect on zoom.

mgermerie:
## Description

- Fix issue #1586 regarding inconsistent values of zoom factors for an orthographic camera and a perspective camera in a planar context.
- Refacor the zoom factor parameters both in `PlanarControls` and `GlobeControls` : the user shall now only give a single `zoomFactor` parameter instead of the previous two factors (`zoomInFactor` and `zoomOutFactor`) in planar and `zoomSpeed` in globe.

## Motivation

- Standardize code and usage between `GlobeControls` and `PlanarControls`.
- Give more relevance to the values of zoomFactor.

### DIFF
--- a/examples/vector_tile_raster_2d.html
+++ b/examples/vector_tile_raster_2d.html
@@ -38,9 +38,6 @@
                     maxAltitude: 40000000,
                     // We want to keep the rotation disabled, to only have a view from the top
                     enableRotation: false,
-                    // Faster zoom in/out speed
-                    zoomInFactor: 0.5,
-                    zoomOutFactor: 0.5,
                     // Don't zoom too much on smart zoom
                     smartTravelHeightMax: 100000,
                 },
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -26,6 +26,7 @@ const panOffset = new THREE.Vector3();
 const dollyStart = new THREE.Vector2();
 const dollyEnd = new THREE.Vector2();
 const dollyDelta = new THREE.Vector2();
+let dollyScale;
 
 // Globe move
 const moveAroundGlobe = new THREE.Quaternion();
@@ -150,7 +151,8 @@ let previous;
  * @param      {CameraTransformOptions|Extent} placement   the {@link CameraTransformOptions} to apply to view's camera
  * or the extent it must display at initialisation, see {@link CameraTransformOptions} in {@link CameraUtils}.
  * @param      {object}  options
- * @param      {number}  options.zoomSpeed Speed zoom with mouse
+ * @param      {number}  [options.zoomFactor=2] The factor the scale is multiplied by when dollying (zooming) in or
+ * divided by when dollying out.
  * @param      {number}  options.rotateSpeed Speed camera rotation in orbit and panoramic mode
  * @param      {number}  options.minDistance Minimum distance between ground and camera
  * @param      {number}  options.maxDistance Maximum distance between ground and camera
@@ -178,9 +180,13 @@ class GlobeControls extends THREE.EventDispatcher {
         // Set to false to disable this control
         this.enabled = true;
 
-        // This option actually enables dollying in and out; left as "zoom" for
+        // These options actually enables dollying in and out; left as "zoom" for
         // backwards compatibility
-        this.zoomSpeed = options.zoomSpeed || 2.0;
+        if (options.zoomSpeed) {
+            console.warn('Controls zoomSpeed parameter is deprecated. Use zoomInFactor and zoomOutFactor instead.');
+            options.zoomFactor = options.zoomFactor || options.zoomSpeed;
+        }
+        this.zoomFactor = options.zoomFactor || 1.25;
 
         // Limits to how far you can dolly in and out ( PerspectiveCamera only )
         this.minDistance = options.minDistance || 250;
@@ -285,8 +291,11 @@ class GlobeControls extends THREE.EventDispatcher {
         this.lookAtCoordinate(placement, false);
     }
 
-    get dollyScale() {
-        return 0.95 ** this.zoomSpeed;
+    get dollyInScale() {
+        return this.zoomFactor;
+    }
+    get dollyOutScale() {
+        return 1 / this.zoomFactor;
     }
 
     get isPaused() {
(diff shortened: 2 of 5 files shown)


gchoqueux:
Thanks, good feature!

Could you add in `globeControls` ([see zoom method](https://github.com/iTowns/itowns/blob/master/src/Controls/GlobeControls.js#L733-L768))?

examples/vector_tile_raster_2d.html
```diff
@@ -39,8 +39,8 @@
                     // We want to keep the rotation disabled, to only have a view from the top
                     enableRotation: false,
                     // Faster zoom in/out speed
-                    zoomInFactor: 0.5,
-                    zoomOutFactor: 0.5,
+                    zoomInFactor: 3,
```
gchoqueux: may be remove this setting in examples and use default value?

src/Controls/GlobeControls.js
```diff
@@ -178,9 +179,15 @@ class GlobeControls extends THREE.EventDispatcher {
         // Set to false to disable this control
         this.enabled = true;
 
-        // This option actually enables dollying in and out; left as "zoom" for
+        // These options actually enables dollying in and out; left as "zoom" for
         // backwards compatibility
-        this.zoomSpeed = options.zoomSpeed || 2.0;
+        if (options.zoomSpeed) {
```
gchoqueux, line 185: it's good to depreciate it

src/Controls/GlobeControls.js
```diff
@@ -178,9 +179,15 @@ class GlobeControls extends THREE.EventDispatcher {
...
+            console.warn('Controls zoomSpeed parameter is deprecated. Use zoomInFactor and zoomOutFactor instead.');
+            options.zoomInFactor = options.zoomInFactor || options.zoomSpeed;
+            options.zoomOutFactor = options.zoomOutFactor || (1 / options.zoomSpeed);
+        }
+        this.zoomInFactor = options.zoomInFactor || 2.0;
```
gchoqueux: I ask whether it is necessary to have two parameters (in and out) when one is the inverse of the other.
WDYT?

src/Controls/GlobeControls.js
```diff
@@ -348,11 +358,7 @@ class GlobeControls extends THREE.EventDispatcher {
         }
     }
 
-    dollyIn(dollyScale) {
-        if (dollyScale === undefined) {
-            dollyScale = this.dollyScale;
-        }
-
+    dolly(dollyScale = 1) {
```
gchoqueux: What is the case where is undefined?

src/Controls/GlobeControls.js
```diff
@@ -552,10 +544,8 @@ class GlobeControls extends THREE.EventDispatcher {
                 dollyEnd.copy(coords);
                 dollyDelta.subVectors(dollyEnd, dollyStart);
 
-                if (dollyDelta.y > 0) {
-                    this.dollyIn();
-                } else if (dollyDelta.y < 0) {
-                    this.dollyOut();
+                if (dollyDelta.y !== 0) {
+                    this.dolly(dollyDelta.y > 0 ? this.dollyInScale : this.dollyOutScale);
```
gchoqueux: move all tests in `dolly` method?

test/unit/globecontrol.js
```diff
@@ -152,13 +152,13 @@ describe('GlobeControls', function () {
...
     it('dolly', function () {
-        controls.dollyIn(10);
+        controls.dolly(10);
         controls.state = controls.states.ORBIT;
         controls.update();
-        controls.dollyOut(10);
+        controls.dolly(0.1);
         controls.state = controls.states.NONE;
-        controls.dollyOut();
-        controls.dollyIn();
```
mgermerie, line 161: I removed these two lines which were just here to test the condition `if (dollyScale === undefined)` in `dollyIn` and `dollyOut` methods.

examples/vector_tile_raster_2d.html
mgermerie: I removed them here but left them in the other example so that there is a track on how to use these parameters.

src/Controls/GlobeControls.js
mgermerie: Thanks !

src/Controls/GlobeControls.js
mgermerie: I changed `GlobeControl` as well as `PlanarControls` so that user passes a single `zoomFactor` parameter (which is equal to the previous `zoomInFactor` parameter). The `zoomOutFactor` is now computed as the inverse of `zoomFactor`.

src/Controls/GlobeControls.js
mgermerie: Doesn't happen in iTowns code. I changed the method signature which now take a `delta` in argument. The method will dolly in or out according to the sign of `delta`.

src/Controls/GlobeControls.js
mgermerie: I moved them in `dolly` method, now only the delta value is sent to the method.

gchoqueux (approved):

PR #1601, opened 2021-03-02
fix(PlanarControls): fix drag when moving out of view's domElement

### LINKED ISSUE #1598: Pan and out of window moves doesn't release the 'pan view mode'
## Context
In a PlanarView, if you click on the map (without release) you enter in a pan area mode, which you quit when you release the button.
But in the case you release the button out of the window view, the pan mode isn't quit and thus as soon as you re-enter the view you are back in the pan mode even without any mouse button pushed down.

mgermerie:
## Description
When the mouse exits the view's domElement area, switches controls to state `NONE`.

## Motivation and Context
Fixes issue #1598.

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -871,6 +871,7 @@ class PlanarControls extends THREE.EventDispatcher {
         this.view.domElement.addEventListener('keydown', this._handlerOnKeyDown, false);
         this.view.domElement.addEventListener('mousedown', this._handlerOnMouseDown, false);
         this.view.domElement.addEventListener('mouseup', this._handlerOnMouseUp, false);
+        this.view.domElement.addEventListener('mouseleave', this._handlerOnMouseUp, false);
         this.view.domElement.addEventListener('mousemove', this._handlerOnMouseMove, false);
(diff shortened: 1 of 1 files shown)


gchoqueux (commented):
it's work, if I up the button on a menu GUI?

mgermerie:
> it's work, if I up the button on a menu GUI?

When hovering the menu while in drag mode, the drag stops. Nothing happens when releasing the mouse button on a menu option (like 'Layer planar' for instance). I assume this is the behavior we expect, since it would not be relevant to trigger menu actions when dragging on the menu (instead of deliberately clicking on the menu).

I'm working on an alternative solution which would add the following behaviors :
- If the cursor exits the view's domElement area while mouse button is down, the drag/pan movement shall be paused.
- If the mouse button is released outside of the area, the controls should switch to state `NONE`.
- If the cursor exits the area while mouse button is down and then return on to the area with mouse button still down, the drag/pan movement shall resume.

Do you think this solution would be relevant ?

gchoqueux:
I would like the same `GlobeControls` Behaviors.

mgermerie:
Alright, then the current version of the PR offers the same behavior than `GlobeControls`.

gchoqueux (approved):
Thanks!

PR #1602, opened 2021-03-03
feat(PlanarControls): Dynamique cursor shape modifications

### LINKED ISSUE #1597: Default mouse pointer not always the one wanted after zoom/unzoom pointer change
## Context

During zooming and unzooming the pointer goes to the 'waiting shape' pointer before going back to the 'default pointer'.
Or in our case, the pointer just before the time of the zoom or dezoom is not always the default pointer.
(for instance when the server is busy by a calculation we set the pointer to 'waiting' and thus want it to stay at waiting at the end of the zoom/unzoom.)

ftoromanoff:
## Description
Add a class 'Cursor' in PlanarControl for the gestion of cursor shape 

## Motivation and Context
It's helpfull when the client wants to change the default curseur (Work in progress for instance) without the state of itowns being differente than NONE

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -34,6 +34,16 @@ export const STATE = {
     ORTHO_ZOOM: 4,
 };
 
+// cursor shape linked to control state
+const cursor = {
+    default: 'auto',
+    drag: 'move',
+    pan: 'cell',
+    travel: 'wait',
+    rotate: 'move',
+    ortho_zoom: 'wait',
+};
+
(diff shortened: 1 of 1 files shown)


mgermerie (approved):

gchoqueux (changes requested):
could you move `Cursor` method and `shape` property in `PlanarControls`?

gchoqueux:
Could you add documentation on method (to explain all states)?
Could you rename `setShape` to `setCursor`?

gchoqueux (approved):
Thanks!

PR #1604, opened 2021-03-03
fix(PlanarControls): prevent triggering new movement when already moving

### LINKED ISSUE #1603: Overriden movements in a planar context
## Your Environment

* Version used: 2.30

## Context

In a planar context, when performing a movement which is not travel (drag / pan / rotation), the movement is overridden by another one if the user activates another movement. 

## Steps to Reproduce
1. Go to http://www.itowns-project.org/itowns/examples/index.html#view_25d_map.
2. Press the left mouse button and start moving the mouse. This will start a drag movement.
3. While holding left mouse button pressed, press the right mouse button. The movement switches to a pan movement.

## Expected Behavior
Several behavior could occur, at developers choice :

- Either all movement should be frozen instead of switching to the latest requested one,
- Or the latest requested movement should be ignored if another movement is already running.

mgermerie:
This PR fixes issue #1603, implementing the following behavior in a planar context :

- if a user triggers a second movement while already in movement, the second movement is ignored.

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -14,6 +14,7 @@ const mouseButtons = {
     MIDDLECLICK: THREE.MOUSE.MIDDLE,
     RIGHTCLICK: THREE.MOUSE.RIGHT,
 };
+let currentPressedButton;
 
 // starting camera position and orientation target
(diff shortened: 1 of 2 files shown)


gchoqueux (approved):
Thanks!

PR #1605, opened 2021-03-03
examples: update irrelevant titles in some examples

mgermerie:
## Description
Update irrelevant titles in some examples.


gchoqueux (approved):
thanks

PR #1610, opened 2021-03-11
Mars Atmosphere example

gchoqueux:
## Description
Add Mars Atmosphere example
Add Atmosphere documentation
Fix realistic Atmosphere

image

### DIFF
--- a/examples/mars.html
+++ b/examples/mars.html
@@ -1,0 +1,94 @@
+<html>
+    <head>
+        <title>Itowns - Mars</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <span id="divScaleWidget"> Scale </span>
+        <div id="miniDiv"></div>
+
+        <script src="js/GUI/GuiTools.js"></script>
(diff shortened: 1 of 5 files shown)


examples/mars.html
```diff
@@ -0,0 +1,90 @@
...
+                    outerRadius: 6600000,
+                    wavelength: [0.350, 0.470, 0.575],
+                    scaleDepth: 0.38,
+                }
+            });
+            setupLoadingScreen(viewerDiv, view);
+
+            // Add a tms imagery source
+            var tmsMarsSource = new itowns.TMSSource({
+                    // crs: 'EPSG:104905',
```
mgermerie, line 54: Could this comment be removed as well as the same on line 69 ? If so, I believe the definition of `EPSG:104905` on line 24 could also be removed, since I don't see it used anywhere else.

mgermerie (approved):

examples/mars.html
gchoqueux: I will add a comment the mars projection is `EPSG:104905` but i use earth to add mars color layer...

zarov:
Really nice ! You could also have an example like `view_multipleglobe` which display planets of the Solar system in the same example.

However the color of the atmosphere is not on point as far as I can tell. The current [SkyShader](https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/Globe/SkyShader.js) has some constants specified on the composition of the Earth atmosphere but the Mars atmosphere has quite a different composition. It could be interesting to be able to specify those constants based on the composition of a planet.

gchoqueux:
Thanks!

> You could also have an example like view_multipleglobe which display planets of the Solar system in the same example.

Yes, indeed, @mgermerie suggested the same example to me

> The current SkyShader has some constants specified on the composition of the Earth atmosphere but the Mars atmosphere has quite a different composition.

I look at this [article](https://elib.dlr.de/86477/1/Collienne_GI_VRAR_2013.pdf)

PR #1612, opened 2021-03-17
feature(PlanarControls): add resolution interval parameters to limit zoom

### LINKED ISSUE #1611: limitation of zoom max and zoom min in ortho view
How to set a max and min level of zoom in an orthographic view ?

## Context

In a orthographic view we can define the zoomFactor. We can then zoom in and out to infinite.
In the Packo project we would like to be able to limite the zoom in and out to some specified zoom level.

by adding a control parameter zoomLimit (or zoomMax + zoom Min) that we aren't able to overpass

mgermerie:
## Description
Add maximum and minimum resolution optional parameters to `PlanarControls`.

- The `maxResolution` parameter defines the maximum resolution the view can reach -- i.e. the smallest size in meters a pixel at the center of the view can represent.
- The `minResolution` parameter defines the minimum resolution the view can reach -- i.e. the largest size in meters a pixel at the center of the view can represent.

Usage of these parameters is added in examples :

- http://www.itowns-project.org/itowns/examples/#view_2d_map where a maximum resolution limit is set to prevent from zooming in too much ;
- http://www.itowns-project.org/itowns/examples/#misc_orthographic_camera where a maximum and a minimum resolution limits are set to prevent from zooming outside a given scale interval.

## Motivation and Context
Bring a solution to issue #1611.

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -96,6 +96,8 @@ const defaultOptions = {
     maxPanSpeed: 15,
     zoomTravelTime: 0.2,  // must be a number
     zoomFactor: 2,
+    maxResolution: 1 / Infinity,
+    minResolution: Infinity,
     maxAltitude: 50000000,
     groundLevel: 200,
     autoTravelTimeMin: 1.5,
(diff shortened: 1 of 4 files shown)


src/Controls/PlanarControls.js
```diff
@@ -515,6 +526,12 @@ class PlanarControls extends THREE.EventDispatcher {
         if (delta > 0 || (delta < 0 && this.maxAltitude > this.camera.position.z)) {
             const zoomFactor = delta > 0 ? this.zoomInFactor : this.zoomOutFactor;
 
+            // do not zoom if the resolution after the zoom is outside resolution limits
+            endResolution = this.view.getPixelsToMeters() / zoomFactor;
```
gchoqueux: could you declare `endResolution` variable in this scope?

gchoqueux (approved):

PR #1616, opened 2021-03-25
chore: up to webpack 5.

gchoqueux:
## Description
Up to `webpack 5.0.0` and `webpack dev server 4.0.0`

`Debug` bundle size is reduced because it depends on `itowns` bundle.

Motivation
use `dependOn` option to facilitate plugin creating


mgermerie (commented):
Now that the examples are included in the watched files (the examples webpages are reloaded when changing their `html` code), the following error can occur (at least on Linux) when running two parallel `npm start` : 
```
Error: ENOSPC: System limit for number of file watchers reached
```

On Linux, this is caused by the `inotify` package. This package watches specified files and detects when changes occur in these files. There is a system setting which defines what I understood to be the maximum number of files `inotify` can simultaneously watch. It can be read with : 
```
cat /proc/sys/fs/inotify/max_user_watches
```
This number can be increased with the following command, replacing `{number}` by the value of your choice :
```
echo fs.inotify.max_user_watches={number} | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```

When tested, increasing this value seamed  to fix the issue for me and I were able to run two parallel `npm start` again.

**Perhaps there could be a way to reduce the number of files watched when running `npm start` ?**

gchoqueux:
> Perhaps there could be a way to reduce the number of files watched when running npm start ?

For  the moment I prefer keep all watched files, to debug npm packages.

mgermerie (approved):

PR #1619, opened 2021-04-01
fix(Label): wrong  label div translate.

gchoqueux:
## Description
Fix  label div translate and add Label padding property.


mgermerie (changes requested):
The tests must be updated so that they don't expect anymore the additional padding value in translation.

mgermerie (approved):

PR #1625, opened 2021-04-06
fix(GeojsonParser): prevent properties from being overriden

### LINKED ISSUE #1406: Overwritten properties in GeoJsonParser
Some properties may be overwritten by [this part](https://github.com/iTowns/itowns/blob/0d400730e274d0840102adfb6092b4324658deb0/src/Parser/GeoJsonParser.js#L131-L136) in the `GeoJsonParser`. 

If you take the `source_stream_wfs_3d` for example and you read the properties before and after this loop, you got two different `id` properties. I don't think this should be the case, and the first one should be kept. However, where can the second one be stored ? I am not really sure of this.

mgermerie:
## Description

As mentioned by #1406, some properties present in geojson features may also be present in those features `properties` property. In that case, the properties present in geojson features would override those appearing under the `properties` key in geojson features.

With this PR, properties that are defined in a feature and are also present in this feature `properties` property are now stored under the `geojson` property. They can thus still be stored without overriding other properties anymore.

## Motivation and Context

Closes #1406.

### DIFF
--- a/examples/source_stream_wfs_3d.html
+++ b/examples/source_stream_wfs_3d.html
@@ -124,9 +124,9 @@
             view.addLayer(lyonTclBusLayer);
 
             function colorBuildings(properties) {
-                if (properties.id.indexOf('bati_remarquable') === 0) {
+                if (properties.geojson.id.indexOf('bati_remarquable') === 0) {
                     return color.set(0x5555ff);
-                } else if (properties.id.indexOf('bati_industriel') === 0) {
+                } else if (properties.geojson.id.indexOf('bati_industriel') === 0) {
                     return color.set(0xff5555);
                 }
                 return color.set(0xeeeeee);
--- a/examples/source_stream_wfs_25d.html
+++ b/examples/source_stream_wfs_25d.html
@@ -173,10 +173,10 @@
             view.addLayer(lyonTclBusLayer);
 
             function colorBuildings(properties) {
-                if (properties.id.indexOf('bati_remarquable') === 0) {
+                if (properties.geojson.id.indexOf('bati_remarquable') === 0) {
                     return color.set(0x5555ff);
                 }
-                if (properties.id.indexOf('bati_industriel') === 0) {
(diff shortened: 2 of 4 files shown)


examples/source_stream_wfs_3d.html
```diff
@@ -124,9 +124,9 @@
             view.addLayer(lyonTclBusLayer);
 
             function colorBuildings(properties) {
-                if (properties.id.indexOf('bati_remarquable') === 0) {
+                if (properties.id_json.indexOf('bati_remarquable') === 0) {
```
gchoqueux: Why there's already a `id` key?

gchoqueux (commented):
There's source_stream_wfs_2d.html example to change?

examples/source_stream_wfs_3d.html
mgermerie: In geojson features, there is an `id` property with a value like "_bati_remarquable.107720_". There is also a `properties` property which contains an `id` like "_BATIMENT0000000240851490_". I don't know if there is an interest in keeping the "_BATIMENT0000000240851490_" id in this specific example, but I assume the good standard procedure would be to keep any duplicate keys.
In the example this result in the fact that the `id` determining the building type (i.e. "_bati_remarquable.107720_") would override the other `id` ("_BATIMENT0000000240851490_"), so it is suffixed.

mgermerie: This made me wonder : which `id` ("_BATIMENT0000000240851490_" or "_bati_remarquable.107720_") do we want to display in the top-right information (see screenshot below) ?
Screenshot from 2021-04-06 16-15-34

gchoqueux: In `geojson` [specification](https://tools.ietf.org/html/rfc7946) `id` is the only member that we don't process.
If there are other members, they are `foreign members`.

we could formalise, as follows:

```js
properties: {
    id: 'a1'
    type: 'other'
    geojson: {
          id: 'bat1',
          title: 'geom',
    }
}

```

gchoqueux (approved):

PR #1628, opened 2021-04-12
New fx option (-1 -> contour)

gmaillet:
## Description

- Add a new option in the color layer shader to view contours (fx = -1)
- Add a new layer (PanIGNV2) in the JSONLayers folder (a PNG wmts source is needed for the example)
- Add a new example (contour_render.html)

The shader compare a pixel with its 4 neighbors (up/down/left/right), it renders red (1, 0, 0, 1) if there is differences and transparent (r, g, b, 0) otherwise. It's important to check the image limits so as not to create false contour. It's also important to only used un-interpolated color to obtain clear contours.

## Motivation and Context

This feature is necessary to display the outlines of a mosaic graph from an image containing one color per area.

Capture-1
Capture-2

### DIFF
--- a/examples/config.json
+++ b/examples/config.json
@@ -47,7 +47,8 @@
         "effects_postprocessing": "Post processing",
         "effects_split": "Splitted rendering",
         "effects_stereo": "Stereo effects",
-        "mars": "Mars atmosphere"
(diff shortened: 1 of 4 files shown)


gchoqueux:
see  #1656

PR #1646, opened 2021-04-29
fix(CameraUtils): the camera rotation animation take the shortest angle

### LINKED ISSUE #1517: cameraUtils.setHeading doesn't take the short
There is a bug in cameraUtils, if the animation between to heading isn't the short way.

You could try in console

```js
view.controls.setTilt(10)
view.controls.setHeading(-36)
view.controls.setHeading(10)
```

Could you fix this bug in other PR?

_Originally posted by @gchoqueux in https://github.com/iTowns/itowns/pull/1511#issuecomment-723922620_

mgermerie:
## Description

- Change the direction with which headings are counted : from anti-clockwise to clockwise (as navigation headings).
- Fix `CameraUtils` method  `animateCameraToLookAtTarget`. The animation when rotating the camera (by changing its heading) is now such as the rotation angle is the smallest possible.

## Motivation and Context

Before this PR, if one wanted to rotate the camera from heading 10° to heading 350°, the animation would rotate clockwise by almost a complete turn around (340°). With this PR, the animation shall rotate of only 20° counter-clockwise.

This PR closes #1517.

## BREAKING CHANGE
This PR introduces a breaking change : the given headings must now be counted clockwise. For instance, setting a heading to `90°` will now move the camera toward East instead of previously moving it toward West.

### DIFF
--- a/examples/3dtiles_25d.html
+++ b/examples/3dtiles_25d.html
@@ -31,7 +31,7 @@
             var cameraCoord = new itowns.Coordinates('EPSG:3946', 1841980,
                 5175682, 3000)
             var view = new itowns.PlanarView(viewerDiv, extent, { placement: {
-                coord: cameraCoord, heading: -30, range: 4000, tilt: 30 } });
+                coord: cameraCoord, heading: 30, range: 4000, tilt: 30 } });
 
             setupLoadingScreen(viewerDiv, view);
 
--- a/examples/3dtiles_batch_table.html
+++ b/examples/3dtiles_batch_table.html
@@ -29,7 +29,7 @@
                 coord: new itowns.Coordinates('EPSG:4326', -75.61349, 40.044259),
                 range: 200,
                 tilt: 10,
-                heading: -145
+                heading: 145,
             };
 
             var viewerDiv = document.getElementById('viewerDiv');
--- a/examples/misc_camera_animation.html
+++ b/examples/misc_camera_animation.html
@@ -47,14 +47,14 @@
             atmosphere.setRealisticOn(true);
 
             pathTravel.push({ coord: new itowns.Coordinates('EPSG:4326', 2.0889, 42.809), range: 100000, time: time * 0.2 });
-            pathTravel.push({ range: 13932, time: time * 0.2, tilt: 7.59, heading: -110.9 });
+            pathTravel.push({ range: 13932, time: time * 0.2, tilt: 7.59, heading: 110.9 });
             pathTravel.push({ tilt: 8, time: time * 0.2 });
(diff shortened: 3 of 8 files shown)


gchoqueux:
There's a issue with azimuth, the 90 angle is pointed to west.

gchoqueux:
could you fix the unit tests?

mgermerie:
@gchoqueux I'm on fixing the tests, but I found out another bug with `CameraUtils` method `animateCameraToLookAtTarget` : 

If you go to [this example](http://www.itowns-project.org/itowns/examples/view_25d_map.html) and you type in console : 
```js
itowns.CameraUtils.animateCameraToLookAtTarget(view, view.camera.camera3D, { heading: 0 });
```
then : 
```js
itowns.CameraUtils.animateCameraToLookAtTarget(view, view.camera.camera3D, { heading: 90 });
```

you shall see several things :
- the bug that this PR aims to fix, which are the fact that the camera rotation is not in the correct clockwise direction and the fact that it does not takes the shortest path
- Another bug which is that when camera is pointing north (heading of absolute 0), the animation moves the starting point of the camera to the south. This only happens when the camera starting heading is 0. I tried with 0.000001 and the issue didn't occur. Therefore, i believe it has something to do with a modulo 180° when getting the initial camera heading. I shall try to figure this out.

mgermerie:
> @gchoqueux I'm on fixing the tests, but I found out another bug with `CameraUtils` method `animateCameraToLookAtTarget` :
> 
> If you go to [this example](http://www.itowns-project.org/itowns/examples/view_25d_map.html) and you type in console :
> 
> ```js
> itowns.CameraUtils.animateCameraToLookAtTarget(view, view.camera.camera3D, { heading: 0 });
> ```
> 
> then :
> 
> ```js
> itowns.CameraUtils.animateCameraToLookAtTarget(view, view.camera.camera3D, { heading: 90 });
> ```
> 
> you shall see several things :
> 
>     * the bug that this PR aims to fix, which are the fact that the camera rotation is not in the correct clockwise direction and the fact that it does not takes the shortest path
> 
>     * Another bug which is that when camera is pointing north (heading of absolute 0), the animation moves the starting point of the camera to the south. This only happens when the camera starting heading is 0. I tried with 0.000001 and the issue didn't occur. Therefore, i believe it has something to do with a modulo 180° when getting the initial camera heading. I shall try to figure this out.

This is now fixed in this PR.

gchoqueux (commented):
Thanks, LGTM

Could you add BREAKING CHANGE in PR description and message commit for
the reverse `heading` behaviour ?

gchoqueux:
@mgermerie could change your commit message

gchoqueux (approved):
Thanks!

PR #1650, opened 2021-05-04
refactor(Feature): FeatureCollection extends by Object3D and use local Matrix

gchoqueux:
## Description
`FeatureCollection` uses local transform matrix. All data needs to be transformed to be  in `crs` projection.

The data (`extent` or `Coordinates`) are stored in a local system.
To use `Feature` vertices or `FeatureCollection/Feature` extent in crs projection,
it's necessary to apply `FeatureCollection.matrixWorld` on feature `Coordinates` or `Extent`.

## BREACKING CHANGES:
FeatureCollection/Feature/FeatureGeometry use local transform matrix.

## Motivation and Context
This PR makes the `FeatureCollection` transformations consistent. They are compatible with THREE js transformations.

In `Feature2Texture` the transformation of feature vertices to coordinates texture uses these transformation matrices.  It will be easier to switch from 2d canvas to 3d canvas.

### DIFF
--- a/src/Converter/Feature2Texture.js
+++ b/src/Converter/Feature2Texture.js
@@ -3,7 +3,6 @@ import { FEATURE_TYPES } from 'Core/Feature';
 import Extent from 'Core/Geographic/Extent';
 import Coordinates from 'Core/Geographic/Coordinates';
 
-const _extent = new Extent('EPSG:4326', [0, 0, 0, 0]);
 const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
 const matrix = svg.createSVGMatrix();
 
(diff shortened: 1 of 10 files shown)


src/Converter/Feature2Texture.js
```diff
@@ -166,25 +171,38 @@ export default {
...
+            // compute matrix transformation `world2texture` to convert coordinates to texture coordinates
+            if (collection.isInverted) {
+                worldTextureOrigin.set(extent.west, extent.north, 0);
+                canvasSize.set(ctx.canvas.width, -ctx.canvas.height, 1.0);
+            } else {
+                worldTextureOrigin.set(extent.west, extent.south, 0);
+                canvasSize.set(ctx.canvas.width, ctx.canvas.height, 1.0);
+            }
+
+            world2texture.compose(worldTextureOrigin, quaternion, dimension.divide(canvasSize)).invert();
```
mgermerie: Should the `invert()` be removed, you could try :
```js
const scale = canvasSize.divide(dimension);
world2texture.compose(
    worldTextureOrigin.multiply(scale).negate(),
    quaternion,
    scale,
);
```
Assumed the quaternion is allways neutral (i.e. no rotation), it gives the same result than : 
```js
world2texture.compose(worldTextureOrigin, quaternion, dimension.divide(canvasSize)).invert()
```

mgermerie (approved):

src/Converter/Feature2Texture.js
gchoqueux: yes, i changed

PR #1654, opened 2021-05-10
feature: add support to reassign all controls bindings in `GlobeView`

### LINKED ISSUE #978: Provide mouse wheel's click-down alternative to existing "Control" button's click-down event
Please provide mouse wheel's click-down alternative to existing left "Control" button click-down event, as it enables single-handed scene tilt manipulation, similar to Google Earth's.

mgermerie:
## Description

- Add possibility to modify the `states` of a `GlobeControls` instance with a `StateControl.setFromOpions` method. This grant an easy reassigning of controls key/mouse bindings after creating a `GlobeView`.

- Refactorize double click event in `GlobeControls`. A travel animation which zoomed in on the mouse position was previously triggered with a `dblclick` event. Now, this travel animation is linked to a state (from `StateControls`). Therefore, it is now possible to modify the event which triggers the travel animation.

  For example, a user could have the travel animation triggered by pressing the `p` key instead of double-click by doing : 
```js
// Assuming `view` is an instance of `GlobeView` :
view.controls.states.setFromOptions({
    TRAVEL_IN: {
        enable: true,
        keyboard: 80,  // 80 is the keyCode for the `p` key
    },
});
```

- Add support in `GlobeControls` for a travel animation which zooms the camera out, linked to the state `StateControl.TRAVEL_OUT`. The event triggering this animation can be changed the same way as for the zoom-in travel. It is deactivated by default, but it can be activated with :
```js
view.controls.states.setFromOptions({
    TRAVEL_OUT: {
        enable: true,
        mouseButton: itowns.THREE.MOUSE.LEFT,  // for instance
        double: true,  // For state to be bound to double click
    },
});
```

- Add an example of a globe view where custom controls are defined.

## Important note

`TRAVEL_IN` and `TRAVEL_OUT` states can't yet be bound to single mouse click (right, left or middle) as it would require a complete refactoring of `GlobeControls`. Should a user try to do so, the travel animations would never occur.

## Left to do on this PR

- [x] Add unit tests

## Motivation and Context

Closes #978.

### DIFF
--- a/examples/misc_custom_controls.html
+++ b/examples/misc_custom_controls.html
@@ -1,0 +1,96 @@
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Itowns - Custom controls</title>
+
+    <link rel="stylesheet" type="text/css" href="css/example.css">
+    <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+</head>
+<body>
+<div id="description">
+    Key bindings
+    <ul>
+        <li>Left-click + drag : rotate the globe</li>
+        <li>Wheel-click + drag : rotate the camera around its target</li>
+        <li>Ctrl + left-click + drag : rotate the camera around its position</li>
+        <li>Mouse wheel up/down : zoom in/out</li>
+        <li>Double left-click : zoom in</li>
+        <li>Double right-click : zoom out</li>
+        <li>Right-click + drag up/down : zoom in/out</li>
+    </ul>
+</div>
+<div id="viewerDiv"></div>
+
+<script src="../dist/itowns.js"></script>
+<script src="../dist/debug.js"></script>
+<script src="js/GUI/LoadingScreen.js"></script>
+<script src="js/GUI/GuiTools.js"></script>
+<script type="text/javascript">
+    /* global itowns, setupLoadingScreen, GuiTools, debug */
+
+    const viewerDiv = document.getElementById('viewerDiv');
+    const placement = {
+        coord: new itowns.Coordinates('EPSG:4326', 2.351323, 48.856712),
+        range: 25000000,
+    }
+    const view = new itowns.GlobeView(viewerDiv, placement);
+    setupLoadingScreen(viewerDiv, view);
+
+    // Add imagery and elevation layers
+    itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function (config) {
+        config.source = new itowns.WMTSSource(config.source);
+        view.addLayer(
+            new itowns.ColorLayer('Ortho', config),
+        );
+    })
+    function addElevationLayerFromConfig(config) {
+        config.source = new itowns.WMTSSource(config.source);
+        view.addLayer(
+            new itowns.ElevationLayer(config.id, config),
+        );
+    }
+    itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
+    itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
+
+    const customControls = {
+        // Disable pan movement
+        PAN: {
+            enable: false,
+        },
+        // Change the key bindings for globe rotation
+        MOVE_GLOBE: {
+            enable: true,
+            mouseButton: itowns.THREE.MOUSE.LEFT,
+        },
+        // Change the key bindings for orbit movement (rotation around the camera target)
+        ORBIT: {
+            enable: true,
+            mouseButton: itowns.THREE.MOUSE.MIDDLE,
+        },
+        // Change the key bindings for dolly movement
+        DOLLY: {
+            enable: true,
+            mouseButton: itowns.THREE.MOUSE.RIGHT,
+        },
+        // Change the key bindings for panoramic movement (rotation around the camera position)
+        PANORAMIC: {
+            enable: true,
+            mouseButton: itowns.THREE.MOUSE.LEFT,
+            keyboard: 17,  // keyCode for the ctrl key
+        },
+        // Allow travel out movement when double right-clicking
+        TRAVEL_OUT: {
+            enable: true,
+            mouseButton: itowns.THREE.MOUSE.RIGHT,
+            double: true,
+        },
+    }
+
+    // Modify view's control to be set as the custom controls we just defined
+    view.controls.states.setFromOptions(customControls);
+</script>
+</body>
+</html>
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -174,7 +174,7 @@ class GlobeControls extends THREE.EventDispatcher {
         this.camera = view.camera.camera3D;
 
         // State control
-        this.states = new StateControl();
+        this.states = new StateControl(this.view);
         this.state = this.states.NONE;
 
         // Set to false to disable this control
@@ -242,7 +242,7 @@ class GlobeControls extends THREE.EventDispatcher {
         this._onMouseDown = this.onMouseDown.bind(this);
         this._onMouseWheel = this.onMouseWheel.bind(this);
         this._onContextMenuListener = this.onContextMenuListener.bind(this);
-        this._ondblclick = this.ondblclick.bind(this);
+        this._onTravel = this.travel.bind(this);
         this._onTouchStart = this.onTouchStart.bind(this);
         this._update = this.update.bind(this);
         this._onTouchMove = this.onTouchMove.bind(this);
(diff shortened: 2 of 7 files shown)


examples/misc_custom_controls.html
```diff
@@ -0,0 +1,112 @@
...
+        },
+        // Allow travel out movement when double right-clicking
+        TRAVEL_OUT: {
+            enable: true,
+            trigger: 'dblclick-right',
+        },
+    }
+
+    // Modify view's control to be set as the custom controls we just defined
+    view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
```
gchoqueux: could you use a new instance of `StateControls` or directly modify the current instance...

src/Controls/GlobeControls.js
```diff
@@ -242,23 +245,31 @@ class GlobeControls extends THREE.EventDispatcher {
...
         this.view.domElement.addEventListener('mousedown', this._onMouseDown, false);
         this.view.domElement.addEventListener('mousewheel', this._onMouseWheel, false);
-        this.view.domElement.addEventListener('dblclick', this._ondblclick, false);
         this.view.domElement.addEventListener('DOMMouseScroll', this._onMouseWheel, false); // firefox
         this.view.domElement.addEventListener('touchstart', this._onTouchStart, false);
         this.view.domElement.addEventListener('touchend', this._onMouseUp, false);
         this.view.domElement.addEventListener('touchmove', this._onTouchMove, false);
 
+        // Add event listener to update the mouse position when moved
+        this.view.domElement.addEventListener('mousemove', this._onMouseUpdate, false);
```
gchoqueux: could remove this part and replace by travel on center instead of mouse position.

src/Controls/GlobeControls.js
```diff
@@ -675,17 +686,40 @@ class GlobeControls extends THREE.EventDispatcher {
...
-        if (point && range > this.minDistance) {
-            return this.lookAtCoordinate({
-                coord: new Coordinates('EPSG:4978', point),
-                range: range * 0.6,
-                time: 1500,
-            });
+    updateMousePosition(event) {
+        coords.copy(this.view.eventToViewCoords(event));
+    }
+
```
gchoqueux: could you remove it?

src/Controls/GlobeControls.js
```diff
@@ -675,17 +686,40 @@ class GlobeControls extends THREE.EventDispatcher {
...
+        if (this.states.TRAVEL_IN.enable) {
+            if (event && event.previous) {
+                // Remove previous travel in event listener
+                this.view.domElement.removeEventListener(event.previous, this._onTravelIn, false);
+            }
+            // Add new updated travel in event listener
+            this.view.domElement.addEventListener(this.states.TRAVEL_IN.trigger, this._onTravelIn, false);
+        } else {
+            this.view.domElement.removeEventListener(this.states.TRAVEL_IN.trigger, this._onTravelIn, false);
+        }
```
gchoqueux: instance `StateControls` with `view.domElement` parameter to handling mouse and keyboard events.

src/Controls/StateControl.js
```diff
@@ -71,6 +86,44 @@ class StateControl {
...
+                previousTravelInTrigger = travelInTrigger;
+                // Update trigger property value
+                travelInTrigger = value;
+                // Emits event to warn listeners that the trigger property has changed value
+                this.dispatchEvent({
+                    type: 'travel-in-changed',
+                    previous: previousTravelInTrigger,
+                });
+            },
+        });
```
gchoqueux: add `dispose` method to remove all `StateControls` events.

src/Controls/GlobeControls.js
mgermerie: I removed this part. Now, the travel targets the mouse position if the triggering event is a `MouseEvent`, and the view center otherwise.

src/Controls/GlobeControls.js
mgermerie: Removed.

src/Controls/GlobeControls.js
mgermerie: I did so.

src/Controls/StateControl.js
mgermerie: I added a `dispose` method in `StateControl`.

examples/misc_custom_controls.html
mgermerie: I added a method in `StateControls` : `setFromOptions` which allows passing all the custom options to `controls.states`.

gchoqueux: why you wait `GLOBE_INITIALIZED`

examples/misc_custom_controls.html
```diff
@@ -0,0 +1,109 @@
...
+    // Create a custom `dblclick-right` event that is triggered when double right-clicking
+    let time;
+    window.addEventListener('mouseup', function (event) {
+        if (event.which === 3) {
+            if (time && event.timeStamp - time < 500) {
+                view.domElement.dispatchEvent(new MouseEvent('dblclick-right', event));
+            }
+            time = event.timeStamp;
+        }
+    });
```
gchoqueux: move in itowns

src/Controls/GlobeControls.js
```diff
@@ -675,15 +676,20 @@ class GlobeControls extends THREE.EventDispatcher {
         }
     }
 
-    ondblclick(event) {
-        if (this.enabled === false || currentKey) { return; }
+    travel(event) {
+        if (this.enabled === false) { return; }
```
gchoqueux, line 681: may be move `this.enable` in `StateControls`? WDYT?

src/Controls/GlobeControls.js
```diff
@@ -675,15 +676,20 @@ class GlobeControls extends THREE.EventDispatcher {
...
         this.player.stop();
-        const point = this.view.getPickingPositionFromDepth(this.view.eventToViewCoords(event));
+        const viewCoords = this.view.eventToViewCoords(event.origin);
```
gchoqueux: `this.view.eventToViewCoords` could be return `undefined` and this avoids computing screen center.

src/Controls/StateControl.js
```diff
@@ -27,13 +41,178 @@ const CONTROL_KEYS = {
  * @class StateControl
  *
  */
-class StateControl {
-    constructor() {
+class StateControl extends THREE.EventDispatcher {
+    constructor(eventTarget, options = {}) {
```
gchoqueux: rename by dom element

src/Controls/StateControl.js
```diff
@@ -27,13 +41,178 @@ const CONTROL_KEYS = {
...
+
+                // Do nothing if TRAVEL_IN.trigger is a keyboard event
+                // and the pressed key doesn't match the TRAVEL_IN key
+                if (event.keyCode && event.keyCode !== this.TRAVEL_IN.keyboard) { return; }
+
+                // If none of the above condition failed, dispatch an event to warn a travel in must be performed
+                this.dispatchEvent({
+                    type: 'travel',
+                    inOrOut: 'in',
+                    origin: event,
```
gchoqueux: return only `event` with adding `type` and  `inOrOut`

src/Controls/GlobeControls.js
mgermerie: Doing so would require to pass the `Controls` instance to `StateControl` (which is not the case now). Therefore, I believe this check is better here than in `StateControl`. Is there a particular reason why you want to move it ?

examples/misc_custom_controls.html
mgermerie: I created a `dblclick-right` event in `View`.

src/Controls/GlobeControls.js
mgermerie: In deed, I changed the method to return `undefined` by default. The previous default behavior was to consider a touch pad event was passed (which did not help the robustness of the method).

src/Controls/StateControl.js
mgermerie: I renamed the property as well as the parameter to `domElement`.

src/Controls/StateControl.js
```diff
@@ -12,65 +12,90 @@ const CONTROL_KEYS = {
 };
 
 
+let travelIn;
+let travelOut;
+
```
gchoqueux: move inside constructor

src/Controls/StateControl.js
```diff
@@ -104,6 +129,101 @@ class StateControl {
...
+        const newTrigger = stateToTrigger(newState);
+        if (!newState.enable && previousState) {
+            this._domElement.removeEventListener(previousTrigger, previousState._emitEvent, false);
+        } else if (!previousState) {
+            this._domElement.addEventListener(newTrigger, newState._emitEvent, false);
+        } else if (newTrigger !== previousTrigger) {
+            this._domElement.removeEventListener(previousTrigger, previousState._emitEvent, false);
+            this._domElement.addEventListener(newTrigger, newState._emitEvent, false);
+        }
+    }
```
gchoqueux: move out this function and remove all tests

src/Controls/StateControl.js
```diff
@@ -12,65 +12,90 @@ const CONTROL_KEYS = {
...
+            },
+        });
+        Object.defineProperty(this, 'TRAVEL_OUT', {
+            get: () => travelOut,
+            set: (value) => {
+                value._emitEvent = this._handleTravelOutEvent;
+                this.updateEventListeners(value, travelOut);
+                travelOut = value;
+            },
+        });
```
gchoqueux: remove  all `Object.defineProperty` and move all in `setFromOption`

src/Controls/StateControl.js
```diff
@@ -104,6 +129,101 @@ class StateControl {
...
+            // Do nothing if state trigger is a keyboard event and the pressed key does not match the state key
+            if (event.keyCode && event.keyCode !== state.keyboard) {
+                return;
+            }
+
+            this.dispatchEvent({
+                type: 'travel',
+                viewCoords: this._view.eventToViewCoords(event),
+                inOrOut,
+            });
```
gchoqueux: move `dispatchEvent`  in `inputToState`

gchoqueux (approved):

PR #1655, opened 2021-05-10
Passing an non-null empty string into source.style gets overridden to 'normal'

Sam-Scolari:
## Motivation and Context

I am working with WMTS data from NASA and I ran into an issue where I needed to either pass &STYLE= into the api request or just leave it out as the tile set I am working with does not have a style. 

## Description

However, WMTSSource only checks whether source.style length is > 0.  If source.style is null or "" then it passes &STYLE=normal into the api request. Furthermore, simply not setting source.style a value passes &STYLE="normal" into the api request, because ("" || "normal") and (null || "normal") both return "normal". WMTSSource must check whether source.style is null or an empty string or else setting source.style = "" will just be overridden to "normal" and therefore pass &STYLE=normal into the api request.

This solution should ensure that existing code that does not set source.style will continue to pass "normal" as the default assuming source.style is initialized as null. 

*This is my first ever pull request to any github project so forgive me if I'm missing something


gchoqueux (changes requested):
Thanks for first pull request.

With this change, it's not possible to set custom wmts `Style` (like `default`, `dark`...).
I propose to just write : 

```js
`&STYLE=${source.style}` +
```

Sam-Scolari:
> Thanks for first pull request.
> 
> With this change, it's not possible to set custom wmts `Style` (like `default`, `dark`...).
> I propose to just write :
> 
> ```js
> `&STYLE=${source.style}` +
> ```

Woops just realized my mistake, recommitted it. I made it return source.style instead of an empty string that way it would allow anyone to pass style: "" or any non-null string into WMTSSource. If you feel just passing source.style would work better that sounds fine as well.

src/Source/WMTSSource.js
```diff
@@ -76,7 +76,7 @@ class WMTSSource extends TMSSource {
             '&SERVICE=WMTS' +
             `&VERSION=${source.version || '1.0.0'}` +
             '&REQUEST=GetTile' +
-            `&STYLE=${source.style || 'normal'}` +
+            `&STYLE=${source.style == null ? "normal" : source.style}` +
```
gchoqueux: Could you replace `null` by `undefined`, because it's the value if the user doesn't define `style`.

Could replace `"` by `'`, because it's syntax rules in itowns code.

gchoqueux (commented):
Could you squash yours commits in one ?

Could you change message commit with angular convention?

`fix(WMTSSource): Passing an non-null empty string into source.style gets overridden to "normal"`

the body message it's not necessaty

Thanks

Sam-Scolari:
I recommitted the new fix, but I am unsure how to squash the 3 commits. Once everything is validated and ready to be merged isn't there an option to "Merge and Squash pull request"?

gchoqueux:
There is squash and merge, but we use rebase and merge.
but there are also commits messages that are not good

You could use command `git rebase` interactive [see examples](https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history#reword-other-commit-messages-take-2) to squash and rename commits.
```
git rebase -i origin/master
```

PR #1656, opened 2021-05-11
Add WebGL ColorLayer Features.

gchoqueux:
## Description
* Set texture `ColorLayer` filtering.
* Add custom shader effect to `ColorLayer`.
* Cleaning of effect setting
* Refactoring chunk shader management
* add custom shader example 

image

### DIFF
--- a/examples/customColorLayerEffect.html
+++ b/examples/customColorLayerEffect.html
@@ -1,0 +1,141 @@
+<html>
+    <head>
+        <title>Itowns - Custom ColorLayer effect</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <div id="miniDiv"></div>
+        <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script src="js/GUI/GuiTools.js"></script>
+        <script type="text/javascript">
+            // # Orthographic viewer
+
+            // Define geographic extent: CRS, min/max X, min/max Y
+            var extent = new itowns.Extent(
+                'EPSG:3857',
+                -20026376.39, 20026376.39,
+                -20048966.10, 20048966.10);
+
+            // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+            var viewerDiv = document.getElementById('viewerDiv');
+
+            // Instanciate PlanarView
+            // By default itowns' tiles geometry have a "skirt" (ie they have a height),
+            // but in case of orthographic we don't need this feature, so disable it
+            var view = new itowns.PlanarView(viewerDiv, extent, { disableSkirt: true, maxSubdivisionLevel: 10,
+                camera: { type: itowns.CAMERA_TYPE.ORTHOGRAPHIC },
+                placement: new itowns.Extent('EPSG:3857', -20000000, 20000000, -8000000, 20000000),
+                diffuse: new itowns.THREE.Color(0x444444),
+                controls: {
+                    // Faster zoom in/out speed
+                    zoomFactor: 3,
+                    // prevent from zooming in too much
+                    maxResolution: 0.005  // a pixel shall not represent a metric size smaller than 5 mm
+                },
+            });
+
+            setupLoadingScreen(viewerDiv, view);
+
+            itowns.ShaderChunk.customHeaderColorLayer(`
+                const mat3 sobelKernelX = mat3(1.0, 2.0, 1.0,
+                                0.0, 0.0, 0.0,
+                                -1.0, -2.0, -1.0);
+
+                const mat3 sobelKernelY = mat3(1.0, 0.0, -1.0,
+                                                2.0, 0.0, -2.0,
+                                                1.0, 0.0, -1.0);
+
+                //performs a convolution on an image with the given kernel
+                float convolve(mat3 kernel, mat3 image) {
+                    float result = 0.0;
+                    for (int i = 0; i < 3; i++) {
+                        for (int j = 0; j < 3; j++) {
+                            result += kernel[i][j]*image[i][j];
+                        }
+                    }
+                    return result;
+                }
+
+                float edge(sampler2D textu, float stepx, float stepy, vec2 center){
+                    // get samples around pixel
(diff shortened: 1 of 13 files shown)


examples/customColorLayerEffect.html
```diff
@@ -0,0 +1,140 @@
...
+
+            // Add a TMS imagery layer
+            var opensmLayer = new itowns.ColorLayer('OPENSM', {
+                updateStrategy: {
+                    type: itowns.STRATEGY_DICHOTOMY,
+                },
+                source: opensmSource,
+                type_effect: itowns.colorLayerEffects.customEffect,
+                amount_effect: 1.0,
+            });
```
gmaillet: this source it not very suitable for this example, is it possible to use a more classical cartographic layer with solid colors (like layers/JSONLayers/OPENSM.json

examples/customColorLayerEffect.html
```diff
@@ -0,0 +1,140 @@
...
+                ivec2 textureSize2d = textureSize(tex,0);
+                vec2 resolution = vec2(float(textureSize2d.x), float(textureSize2d.y));
+
+                float step = layer.amount_effect;
+
+                vec2 cuv = pitUV(uv.xy, offsetScale);
+
+                float value = edge(tex, step/resolution.x, step/resolution.y, cuv);
+
+                color = vec4(vec3(0.27, 0.4, 0.42) *  value * 0.8, value);
```
gmaillet: why these strange values rather than 
color = vec4(vec3(1.0, 0.0, 0.0) * value, value);

examples/customColorLayerEffect.html
```diff
@@ -0,0 +1,140 @@
...
+            var opensmLayer = new itowns.ColorLayer('OPENSM', {
+                updateStrategy: {
+                    type: itowns.STRATEGY_DICHOTOMY,
+                },
+                source: opensmSource,
+                type_effect: itowns.colorLayerEffects.customEffect,
+                amount_effect: 1.0,
+            });
+
+            view.addLayer(opensmLayer);
```
gmaillet: it could be a good idea to add two layers in the view with the same source: one normal and one with the special effect.

examples/customColorLayerEffect.html
```diff
@@ -0,0 +1,140 @@
...
+            });
+
+            // Add a TMS imagery layer
+            var opensmLayer = new itowns.ColorLayer('OPENSM', {
+                updateStrategy: {
+                    type: itowns.STRATEGY_DICHOTOMY,
+                },
+                source: opensmSource,
+                type_effect: itowns.colorLayerEffects.customEffect,
+                amount_effect: 1.0,
```
gmaillet: it could be a good idea to add magFilter and minFilter in these example

src/Layer/ColorLayer.js
```diff
@@ -26,11 +27,13 @@ import Style from 'Core/Style';
...
- * if `fx` equals:
+ * @property {number} type_effect - type effect to apply on raster color.
+ * if `type_effect` equals:
  * * `0`: no special effect.
- * * `> 0. to < 2.0`: light color to invisible effect.
- * * `>= 2`: white color to invisible effect.
+ * * `1`: light color to invisible effect.
+ * * `2`: white color to invisible effect.
+ * * `3`: custom shader effect (defined `ShaderChunk.customBodyColorLayer` and `ShaderChunk.customHeaderColorLayer`).
+ * @property {number} amount_effect - amount value used with effect applied on raster color.
```
gmaillet: "amount_effect": not very precise or explicit?
maybe some thing more generic like "effect_parameter" because it's not necessary a question of amount

gchoqueux:
@gmaillet thanks, I have taken into account all your comments

gmaillet (approved):

src/Layer/ColorLayer.js
```diff
@@ -50,6 +53,8 @@ class ColorLayer extends RasterLayer {
      * available using `layer.name` or something else depending on the property
      * name.
      * @param {Source} [config.source] - Description and options of the source.
+     * @param {Source} [config.magFilter] - How the texture is sampled when a texel covers more than one pixel. [see](https://threejs.org/docs/?q=texture#api/en/textures/Texture.magFilter)
+     * @param {Source} [config.minFilter] - How the texture is sampled when a texel covers less than one pixel. [see](https://threejs.org/docs/?q=texture#api/en/textures/Texture.minFilter)
```
mgermerie: The `MagFilter` and `minFilter` values are supposed to be of type `number`, according to threejs doc.

src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
```diff
@@ -1,10 +1,13 @@
 struct Layer {
     int textureOffset;
     int crs;
-    float effect;
+    int effect_type;
+    float effect_parameter;
     float opacity;
 };
 
+#include <itowns/custom_header_colorLayer>
```
mgermerie, line 9: Could the custom header be included only if `Layer.effect_type` is worth `3` ?

examples/vector_tile_raster_3d.html
```diff
@@ -68,7 +68,8 @@
 
             var mvtLayer = new itowns.ColorLayer('MVT', {
                 source: mvtSource,
-                fx: 2.5,
+                effect_type: itowns.colorLayerEffects.removeLightColor,
+                effect_parameter: 2.5,
```
mgermerie, line 72: I may be missing some point, but isn't the `fx: 2.5` previous value corresponding to a 'white color to invisible effect' ? If so, shouldn't the `effect_type` be `itowns.colorLayerEffects.removeWhiteColor` ?

examples/customColorLayerEffect.html
```diff
@@ -0,0 +1,141 @@
...
+                -20026376.39, 20026376.39,
+                -20048966.10, 20048966.10);
+
+            // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+            var viewerDiv = document.getElementById('viewerDiv');
+
+            // Instanciate PlanarView
+            // By default itowns' tiles geometry have a "skirt" (ie they have a height),
+            // but in case of orthographic we don't need this feature, so disable it
+            var view = new itowns.PlanarView(viewerDiv, extent, { disableSkirt: true, maxSubdivisionLevel: 10,
```
mgermerie, line 34: Unless I'm wrong : since the camera is orthographic, setting `disableSkirt` to `true` or `false` does not have a visual effect.
Is there another effect I am missing with this parameter, like a performance effect or something ?

src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl
gchoqueux: the custom header always includes in preprocess

examples/vector_tile_raster_3d.html
gchoqueux: yes,  it was an error in previous documentation

examples/customColorLayerEffect.html
gchoqueux: Skirt removes terrain visual artefact when there's a DEM.

mgermerie (approved):

PR #1657, opened 2021-05-20
Fix Feature issues

gchoqueux:
## Description
This PR fixes some little issues:
* Wrong condition to choose `Feature` extent crs projection.
* Remove unnecessary `Feature` parameters in `VectorTileParser`
* Transform extent `FileSource` if is needed.

### DIFF
--- a/src/Source/FileSource.js
+++ b/src/Source/FileSource.js
@@ -169,7 +169,9 @@ class FileSource extends Source {
             if (data.extent) {
                 this.extent = data.extent.clone();
                 // Transform local extent to data.crs projection.
-                this.extent.applyMatrix4(data.matrixWorld);
+                if (this.extent.crs == data.crs) {
+                    this.extent.applyMatrix4(data.matrixWorld);
+                }
             }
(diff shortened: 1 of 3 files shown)


src/Source/FileSource.js
```diff
@@ -169,7 +169,9 @@ class FileSource extends Source {
             if (data.extent) {
                 this.extent = data.extent.clone();
                 // Transform local extent to data.crs projection.
-                this.extent.applyMatrix4(data.matrixWorld);
+                if (this.extent.crs == data.crs) {
```
mgermerie, line 172: As you relevantly suggested, maybe you could add a comment on the fact that both crs are different only when `data.crs` is 4978, and that if this is the case, applying the transformation to the extent would result in an irrelevant extent (since 4978 is a 3d crs).

mgermerie (approved):

src/Source/FileSource.js
gchoqueux: added

PR #1662, opened 2021-05-25
FeatureCollection local system and Style GeometryLayer.

### LINKED ISSUE #1330: Feature2Mesh.convert and Altitute function option
In `Feature2Mesh.convert`, it's possible to add callback to get altitude when mesh is built.

But this callback doesn't work the same way if we're building point, line or polygon.

It's necessary to unify the behaviours and in particular to be able to acquire the altitudes in the `properties`.

### LINKED ISSUE #1596: Precision loss when adding a new color layer from a geojson
## Your Environment
* Version used: 2.30
* Browser Name and version: Windows
* Operating System and version (desktop or mobile): desktop
* Link to your project: PackO

## Context

I'm trying to add a new Layer from a GeoJson in Lambert83 and the resulting layer has a wrong positioning for the y axes.

image

le polygone vert et le fond rouge devraient être superposable. 
On peut aussi remarquer sur le polygone orange qui devrait être continu que seuls des segments verticaux sont représentés (j'ai utilisé une width: 3)

gchoqueux:
## Description
* Use local system to store `FeatureCollection` Coordinates. 
* Use `GeometryLayer` Style instead of `layer.convert` options.
* The altitude coordinates is determined in `FeatureCollection` during parsing phase (instead of Converting phase).
* the altitude maximum and minimum are computed  during parsing phase.

## Motivation
* the local system avoids precision issue (fix  #1596)
* The mesh vertices are directly in good system, and there is not a new transformation.
* the `FeatureCollection` local system axes are **East, North, Up (ENU)**  [Axes conventions](https://en.wikipedia.org/wiki/Axes_conventions).
* Unify altitude coordinates determining  (fix #1330).
* Using `Function` to define properties Style (in `ColorLayer` and `GeometricLayer`)

## WIP
- [x] fix raster picking color.

## BREAKING CHANGES
* `GeometryLayer.convert` options are moved in Style properties. Use  
  * `Style.xxx.color`
  * `Style.xxx.base_altitude`
  * `Style.fill.extrusion_height`
  * `Style.stroke.width`
  * `Style.point.radius`
* `overrideAltitudeInToZero` layer options is removed use `Style.xxx.base_altitude` instead.

#### Local system 
image

#### Use `function` to style `ColorLayer`
image

### DIFF
--- a/examples/source_file_gpx_3d.html
+++ b/examples/source_file_gpx_3d.html
@@ -62,7 +62,6 @@
             var waypointGeometry = new itowns.THREE.BoxGeometry(1, 1, 80);
             var waypointMaterial = new itowns.THREE.MeshBasicMaterial({ color: 0xffffff });
             // Listen for globe full initialisation event
-            var waypoints = [];
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 console.info('Globe initialized');
                 itowns.Fetcher.xml('https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/ULTRA2009.gpx')
@@ -72,38 +71,36 @@
                         },
                         out: {
                             crs: view.referenceCrs,
-                            buildExtent: true,
-                            mergeFeatures: true,
                             structure: '3d',
+                            style: new itowns.Style({
+                                stroke : {
+                                    color: 'red',
+                                    width: 2,
+                                },
+                                point : {
+                                    color: 'white',
+                                }
+                            })
                         }
                     }))
-                    .then(collection => itowns.Feature2Mesh.convert({
-                        color: new itowns.THREE.Color(0xff0000),
-                    })(collection))
+                    .then(itowns.Feature2Mesh.convert())
                     .then(function (mesh) {
                         if (mesh) {
+                            mesh.updateMatrixWorld();
                             mesh.traverse((m) => {
-                                if (m.type == 'Line') {
-                                    m.material.linewidth = 5;
-                                }
-
                                 if (m.type == 'Points') {
                                     var vertices = m.feature.vertices;
-                                    for (var i = 0; i < vertices.length; i++) {
+                                    for (var i = 0; i < vertices.length; i += 3) {
                                         var waypoint = new itowns.THREE.Mesh(waypointGeometry, waypointMaterial);
-                                        waypoint.position.set(
-                                            vertices[i * 3],
-                                            vertices[i * 3 + 1],
-                                            vertices[i * 3 + 2]);
-                                        waypoint.lookAt(0, 0, 0);
+                                        waypoint.position.fromArray(vertices, i);
+                                        waypoint.lookAt(mesh.worldToLocal(new itowns.THREE.Vector3()));
                                         waypoint.onBeforeRender = updatePointScale;
+                                        waypoint.updateMatrix();
+                                        mesh.add(waypoint);
                                         waypoint.updateMatrixWorld();
-                                        waypoints.push(waypoint);
                                     }
                                 }
                             });
-                            mesh.add(...waypoints);
-                            mesh.updateMatrixWorld();
                             view.scene.add(mesh);
                             view.notifyChange();
                         }
--- a/examples/source_stream_wfs_25d.html
+++ b/examples/source_stream_wfs_25d.html
@@ -235,8 +235,10 @@
                     altitude: altitudeBuildings,
                     extrude: extrudeBuildings }),
                 onMeshCreated: function scaleZ(mesh) {
-                    mesh.scale.z = 0.01;
-                    meshes.push(mesh);
+                    mesh.children.forEach(c => {
+                        c.scale.z = 0.01;
+                        meshes.push(c);
+                    })
                 },
                 filter: acceptFeature,
                 overrideAltitudeInToZero: true,
--- a/examples/source_stream_wfs_25d.html
+++ b/examples/source_stream_wfs_25d.html
@@ -101,18 +101,6 @@
 
             view.addLayer(wmsElevationLayer);
 
-            function setMaterialLineWidth(result) {
-                result.traverse(function _setLineWidth(mesh) {
-                    if (mesh.material) {
-                        mesh.material.linewidth = 5;
-                    }
-                });
-            }
-
-            function colorLine(properties) {
-                return color.set(Math.round(Math.random() * 0xffffff));
-            }
-
             var tile;
             var linesBus = [];
 
@@ -161,13 +149,17 @@
 
             var lyonTclBusLayer = new itowns.GeometryLayer('lyon_tcl_bus', new itowns.THREE.Group(), {
                 update: itowns.FeatureProcessing.update,
-                convert: itowns.Feature2Mesh.convert({
-                    color: colorLine,
-                    altitude: altitudeLine }),
-                onMeshCreated: setMaterialLineWidth,
+                convert: itowns.Feature2Mesh.convert(),
                 filter: acceptFeatureBus,
                 source: lyonTclBusSource,
                 zoom: { min: 2 },
+
+                style: new itowns.Style({
+                    stroke: {
+                        base_altitude: altitudeLine,
+                        width: 5,
+                    }
+                })
             });
 
             view.addLayer(lyonTclBusLayer);
@@ -230,10 +222,8 @@
             var wfsBuildingLayer = new itowns.GeometryLayer('wfsBuilding', new itowns.THREE.Group(), {
                 update: itowns.FeatureProcessing.update,
                 convert: itowns.Feature2Mesh.convert({
-                    color: colorBuildings,
-                    batchId: function (property, featureId) { return featureId; },
-                    altitude: altitudeBuildings,
-                    extrude: extrudeBuildings }),
+                    batchId: function (property, featureId) { return featureId; }
+                }),
                 onMeshCreated: function scaleZ(mesh) {
                     mesh.children.forEach(c => {
                         c.scale.z = 0.01;
@@ -241,10 +231,17 @@
                     })
                 },
                 filter: acceptFeature,
-                overrideAltitudeInToZero: true,
                 crs: 'EPSG:3946',
                 source: wfsBuildingSource,
                 zoom: { min: 4 },
+
+                style: new itowns.Style({
+                    fill: {
+                        color: colorBuildings,
+                        base_altitude: altitudeBuildings,
+                        extrusion_height: extrudeBuildings,
+                    }
+                })
             });
 
             view.addLayer(wfsBuildingLayer);
--- a/examples/view_immersive.html
+++ b/examples/view_immersive.html
@@ -135,14 +135,16 @@
                 // create geometry layer for the buildings
                 var wfsBuildingLayer = new itowns.GeometryLayer('Buildings', new itowns.THREE.Group(), {
                     update: itowns.FeatureProcessing.update,
-                    convert: itowns.Feature2Mesh.convert({
-                        altitude: altitudeBuildings,
-                        extrude: extrudeBuildings }),
-
+                    convert: itowns.Feature2Mesh.convert(),
+                    style: new itowns.Style({
+                        fill: {
+                            base_altitude: altitudeBuildings,
+                            extrusion_height: extrudeBuildings,
+                        }
+                    }),
                     // when a building is created, it get the projective texture mapping, from oriented image layer.
                     onMeshCreated: (mesh) => mesh.traverse(object => object.material = orientedImageLayer.material),
                     source: wfsBuildingSource,
-                    overrideAltitudeInToZero: true,
                     zoom: { min: 15 },
                 });
 
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -1,31 +1,20 @@
 import * as THREE from 'three';
 import Earcut from 'earcut';
-import Coordinates from 'Core/Geographic/Coordinates';
 import { FEATURE_TYPES } from 'Core/Feature';
+import { deprecatedFeature2MeshOptions } from 'Core/Deprecated/Undeprecator';
 
-function getProperty(name, options, defaultValue, ...args) {
-    const property = options[name];
+const _color = new THREE.Color();
 
-    if (property) {
-        if (typeof property === 'function') {
-            const p = property(...args);
-            if (p) {
-                return p;
-            }
+function toColor(color) {
+    if (color) {
+        if (color.type == 'Color') {
+            return color;
         } else {
-            return property;
+            return _color.set(color);
         }
+    } else {
+        return _color.set(Math.random() * 0xffffff);
     }
-
-    if (typeof defaultValue === 'function') {
-        return defaultValue(...args);
-    }
-
-    return defaultValue;
-}
-
-function randomColor() {
-    return new THREE.Color(Math.random() * 0xffffff);
 }
 
 function fillColorArray(colors, length, color, offset = 0) {
@@ -50,37 +39,23 @@ function fillBatchIdArray(batchId, batchIdArray, start, end) {
  * @param      {number[]} ptsIn - Coordinates of a feature.
  * @param      {number[]} normals - Coordinates of a feature.
  * @param      {number[]} target - Target to copy result.
- * @param      {(Function|number)}  altitude - Altitude of feature or function to get altitude.
- * @param      {number} extrude - The extrude amount to apply at each point
+ * @param      {number}  zTranslation - Translation on Z axe.
  * @param      {number} offsetOut - The offset array value to copy on target
  * @param      {number} countIn - The count of coordinates to read in ptsIn
  * @param      {number} startIn - The offser array to strat reading in ptsIn
  */
-const coord = new Coordinates('EPSG:4326', 0, 0);
-function coordinatesToVertices(ptsIn, normals, target, altitude = 0, extrude = 0, offsetOut = 0, countIn = ptsIn.length / 3, startIn = offsetOut) {
+function coordinatesToVertices(ptsIn, normals, target, zTranslation, offsetOut = 0, countIn = ptsIn.length / 3, startIn = offsetOut) {
     startIn *= 3;
     countIn *= 3;
     offsetOut *= 3;
     const endIn = startIn + countIn;
-    let fnAltitude;
-    if (!isNaN(altitude)) {
-        fnAltitude = () => altitude;
-    } else if (Array.isArray(altitude)) {
-        fnAltitude = id => altitude[(id - startIn) / 3];
-    } else {
-        fnAltitude = id => altitude({}, coord.setFromArray(ptsIn, id));
-    }
 
     for (let i = startIn, j = offsetOut; i < endIn; i += 3, j += 3) {
         // move the vertex following the normal, to put the point on the good altitude
-        const t = fnAltitude(i) + (Array.isArray(extrude) ? extrude[(i - startIn) / 3] : extrude);
-        if (target.minAltitude) {
-            target.minAltitude = Math.min(t, target.minAltitude);
-        }
         // fill the vertices array at the offset position
-        target[j] = ptsIn[i] + normals[i] * t;
-        target[j + 1] = ptsIn[i + 1] + normals[i + 1] * t;
-        target[j + 2] = ptsIn[i + 2] + normals[i + 2] * t;
+        target[j] = ptsIn[i] + normals[i] * zTranslation;
+        target[j + 1] = ptsIn[i + 1] + normals[i + 1] * zTranslation;
+        target[j + 2] = ptsIn[i + 2] + normals[i + 2] * zTranslation;
     }
 }
 
@@ -136,18 +111,21 @@ function featureToPoint(feature, options) {
     let featureId = 0;
 
     let vertices;
-    if (options.altitude !== 0) {
+    const zTranslation = options.GlobalZTrans - feature.altitude.min;
+    if (zTranslation !== 0) {
         vertices = new Float32Array(ptsIn.length);
-        coordinatesToVertices(ptsIn, normals, vertices, options.altitude);
+        coordinatesToVertices(ptsIn, normals, vertices, zTranslation);
     } else {
         vertices = new Float32Array(ptsIn);
     }
-
+    const globals = { point: true };
     for (const geometry of feature.geometries) {
-        const color = getProperty('color', options, randomColor, geometry.properties);
+        const context = { globals, properties: () => geometry.properties };
+        const style = feature.style.drawingStylefromContext(context);
+
         const start = geometry.indices[0].offset;
         const count = geometry.indices[0].count;
-        fillColorArray(colors, count, color, start);
+        fillColorArray(colors, count, toColor(style.point.color), start);
 
         if (batchIds) {
             const id = options.batchId(geometry.properties, featureId);
@@ -161,9 +139,9 @@ function featureToPoint(feature, options) {
     geom.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
     if (batchIds) { geom.setAttribute('batchId', new THREE.BufferAttribute(batchIds, 1)); }
 
-    const points = new THREE.Points(geom, pointMaterial);
-    points.minAltitude = 0;
-    return points;
+    pointMaterial.size = feature.style.point.radius;
+
+    return new THREE.Points(geom, pointMaterial);
 }
 
 var lineMaterial = new THREE.LineBasicMaterial();
@@ -177,9 +155,10 @@ function featureToLine(feature, options) {
     let featureId = 0;
 
     let vertices;
-    if (options.altitude !== 0) {
+    const zTranslation =  options.GlobalZTrans - feature.altitude.min;
+    if (zTranslation != 0) {
         vertices = new Float32Array(ptsIn.length);
-        coordinatesToVertices(ptsIn, normals, vertices, options.altitude);
+        coordinatesToVertices(ptsIn, normals, vertices, zTranslation);
     } else {
         vertices = new Float32Array(ptsIn);
     }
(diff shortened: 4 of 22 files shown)


src/Converter/Feature2Mesh.js
```diff
@@ -502,10 +477,23 @@ export default {
...
+        deprecatedFeature2MeshOptions(options);
         return function _convert(collection) {
             if (!collection) { return; }
 
-            return featuresToThree(collection.features, options);
+            const features = collection.features;
+
+            if (!features || features.length == 0) { return; }
+
+            const group = new THREE.Group();
```
mgermerie, line 467: Here you create a `Group` without 'cleaning' the other previous groups.

mgermerie:
Could you specify in doc that if no altitude is in the data, the objects altitude is set to `0` ?

examples/source_file_gpx_3d.html
```diff
@@ -72,38 +71,37 @@
...
-                    .then(function (mesh) {
+                    .then(collection => {
+                        collection.setParentStyle(new itowns.Style({
+                            stroke : {
+                                color: 'red',
+                                width: 2,
+                            },
+                            point : {
+                                color: 'white',
+                            }
```
mgermerie: Maybe the style definition for `points` could be removed, since the points are hidden under the meshes.

examples/source_stream_wfs_25d.html
```diff
@@ -161,13 +161,18 @@
 
             var lyonTclBusLayer = new itowns.GeometryLayer('lyon_tcl_bus', new itowns.THREE.Group(), {
                 update: itowns.FeatureProcessing.update,
-                convert: itowns.Feature2Mesh.convert({
-                    color: colorLine,
-                    altitude: altitudeLine }),
-                onMeshCreated: setMaterialLineWidth,
```
mgermerie, line 167: Could you remove the method `setMaterialLineWidth` (from line 104 to line 110) ?  I believe it is no longer used in the example.

examples/source_stream_wfs_25d.html
```diff
@@ -230,19 +235,26 @@
...
-                    extrude: extrudeBuildings }),
+                    batchId: function (property, featureId) { return featureId; }
+                }),
                 onMeshCreated: function scaleZ(mesh) {
-                    mesh.scale.z = 0.01;
-                    meshes.push(mesh);
+                    mesh.children.forEach(c => {
+                        c.scale.z = 0.01;
+                        meshes.push(c);
+                    })
```
mgermerie, line 231: For my information : is this part combined with the `scaler` method defined from lines 199 to 212 responsible for the buildings 'rising from the ground' effect ?

mgermerie: Also : what is the difference between applying a scale to the whole meshes group instead of applying it individualy to each mesh within the group ?

examples/view_immersive.html
```diff
@@ -135,10 +135,14 @@
...
                 var wfsBuildingLayer = new itowns.GeometryLayer('Buildings', new itowns.THREE.Group(), {
                     update: itowns.FeatureProcessing.update,
-                    convert: itowns.Feature2Mesh.convert({
-                        altitude: altitudeBuildings,
-                        extrude: extrudeBuildings }),
-
+                    convert: itowns.Feature2Mesh.convert(),
+                    style: new itowns.Style({
+                        fill: {
+                            color: () => new itowns.THREE.Color(0xbbffbb),
```
mgermerie: Why is a method passed for the color ? I believe it could simply be `color: new itowns.THREE.Color(0xbbffbb)`, or even no color at all.

src/Core/Deprecated/Undeprecator.js
```diff
@@ -36,20 +36,41 @@ export const deprecatedParsingOptionsToNewOne = (options) => {
         }
         newOptions.out.filteringExtent = options.filteringExtent;
         newOptions.out.style = options.style;
-        newOptions.out.overrideAltitudeInToZero = options.overrideAltitudeInToZero;
+        if (options.overrideAltitudeInToZero) {
+            console.error('Parsing options out.overrideAltitudeInToZero is removed, use Style.xxx.base_altitude instead');
+        }
```
mgermerie: I believe this could be removed, since it is already implemented at lines 55 to 57.

src/Converter/Feature2Mesh.js
```diff
@@ -136,18 +111,21 @@ function featureToPoint(feature, options) {
...
     let vertices;
-    if (options.altitude !== 0) {
+    const zTranslation = feature.altitude.min - options.GlobalZTrans;
+    if (zTranslation !== 0) {
         vertices = new Float32Array(ptsIn.length);
-        coordinatesToVertices(ptsIn, normals, vertices, options.altitude);
+        coordinatesToVertices(ptsIn, normals, vertices, zTranslation);
     } else {
         vertices = new Float32Array(ptsIn);
     }
```
mgermerie: Could this part be replaced by the following ?
```js
vetices = new Float32Array(ptsIn);
if (zTranslation !== 0) {
    coordinatesToVertices(ptsIn, normals, vertices, zTranslation);
}
```

I believe it wouldn't change anything since the values within `vertices` are changed in `coordinatesToVertices` anyway.

src/Converter/Feature2Mesh.js
gchoqueux: done

gchoqueux:
> Could you specify in doc that if no altitude is in the data, the objects altitude is set to `0` ?

documentation added

src/Core/Style.js
```diff
@@ -137,42 +143,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
  * @property {number} order - Order of the features that will be associated to
  * the style. It can helps sorting and prioritizing features if needed.
  * @property {Object} fill - Polygons and fillings style.
- * @property {string} fill.color - Defines the main color of the filling. Can be
+ * @property {string|function} fill.color - Defines the main color of the filling. Can be
  * any [valid color
  * string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
  * Default is no value, indicating that no filling needs to be done.
- * @property {Image|Canvas|string} fill.pattern - Defines a pattern to fill the
+ * If the `Layer` is a `GeometricLayer` you could use `THREE.Color`.
```
mgermerie: Good addition to the doc. Could you replace `GeometricLayer` by `GeometryLayer` ?

src/Core/Style.js
```diff
@@ -137,42 +143,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * If `base_altitude` is `undefined`, the original altitude is keeped and if it doesn't exist so
+ * the altitude value is setted to 0.
  *
  * @property {Object} point - Point style.
- * @property {string} point.color - The color of the point. Can be any [valid
+ * @property {string|function} point.color - The color of the point. Can be any [valid
  * color string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
  * Default is no value, indicating that no point will be shown.
- * @property {number} point.radius - The radius of the point, in pixel. Default
+ * For a `GeometricLayer`, this opacity property isn't used.
```
mgermerie: I think this line should be removed

src/Parser/GeoJsonParser.js
```diff
@@ -31,17 +31,12 @@ const firstPtIsOut = (extent, aCoords, crs) => {
...
+    populateGeometry(crsIn, coordinates, geometry, feature) {
         geometry.startSubGeometry(coordinates.length, feature);
-
-        const useAlti = !collection.overrideAltitudeInToZero &&
-                        collection.size == 3 &&
-                        typeof coordinates[0][2] == 'number';
-
+        coord.crs = crsIn;
         // coordinates is a list of pair [[x1, y1], [x2, y2], ..., [xn, yn]]
         for (const pair of coordinates) {
```
mgermerie: Maybe the name `pair` is no longer adapted, since in a 3d context we have 3 coordinates

test/unit/featureprocess.js
```diff
@@ -18,13 +19,13 @@ describe('Layer with Feature process', function () {
...
+    // function extrude() {
+    //     return 5000;
+    // }
 
-    function color() {
-        return new THREE.Color(0xffcc00);
-    }
+    // function color() {
+    //     return new THREE.Color(0xffcc00);
+    // }
```
mgermerie: Could you remove these commented lines ?

src/Core/Style.js
```diff
@@ -137,42 +143,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
  * for how to use.
- * @property {number} fill.opacity - The opacity of the color or the
+ * @property {number|function} fill.opacity - The opacity of the color or the
  * pattern. Can be between `0.0` and `1.0`. Default is `1.0`.
- *
+ * For a `GeometricLayer`, this opacity property isn't used.
+ * @property {number|function} fill.base_altitude - `GeometryLayer` Style, defines altitude
+ * for each Coordinates.
+ * If `base_altitude` is `undefined`, the original altitude is keeped and if it doesn't exist so
+ * the altitude value is setted to 0.
```
mgermerie: I would rather write it as such : 
```
If `base_altitude` is `undefined`, the original altitude is kept, and if it doesn't exist then the altitude value is set to 0.
```

src/Core/Style.js
```diff
@@ -137,42 +143,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * If the `Layer` is a `GeometricLayer` you could use `THREE.Color`.
+ * @property {number|function} stroke.opacity - The opacity of the line. Can be between
  * `0.0` and `1.0`. Default is `1.0`.
- * @property {number} stroke.width - The width of the line. Default is `1.0`.
+ * For a `GeometricLayer`, this opacity property isn't used.
+ * @property {number|function} stroke.width - The width of the line. Default is `1.0`.
+ * @property {number|function} stroke.base_altitude - `GeometryLayer` Style, defines altitude
+ * for each Coordinates.
+ * If `base_altitude` is `undefined`, the original altitude is keeped and if it doesn't exist so
+ * the altitude value is setted to 0.
```
mgermerie: I would rather write it as such : 
```
If `base_altitude` is `undefined`, the original altitude is kept, and if it doesn't exist then the altitude value is set to 0.
```

src/Core/Style.js
```diff
@@ -137,42 +143,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * With a `GeometricLayer` you could use `THREE.Color`.
+ * @property {number|function} point.width - The width of the border, in pixel. Default
  * is `0.0` (no border).
- * @property {number} point.opacity - The opacity of the point. Can be between
+ * @property {number|function} point.opacity - The opacity of the point. Can be between
  * `0.0` and `1.0`. Default is `1.0`.
+ * @property {number|function} point.base_altitude - `GeometryLayer` Style, defines altitude
+ * for each Coordinates.
+ * If `base_altitude` is `undefined`, the original altitude is keeped and if it doesn't exist so
+ * the altitude value is setted to 0.
```
mgermerie: I would rather write it as such : 
```
If `base_altitude` is `undefined`, the original altitude is kept, and if it doesn't exist then the altitude value is set to 0.
```

src/Core/Style.js
```diff
@@ -137,42 +143,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
- * @property {number} point.radius - The radius of the point, in pixel. Default
+ * For a `GeometricLayer`, this opacity property isn't used.
+ * @property {number|function} point.radius - The radius of the point, in pixel. Default
  * is `2.0`.
- * @property {string} point.line - The color of the border of the point. Can be
+ * @property {string|function} point.line - The color of the border of the point. Can be
  * any [valid color
  * string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
- * @property {number} point.width - The width of the border, in pixel. Default
+ * With a `GeometricLayer` you could use `THREE.Color`.
```
mgermerie: You should instead specify that `point.line` is not supported with a `GeometryLayer`.

examples/source_stream_wfs_25d.html
gchoqueux: removed

examples/source_stream_wfs_25d.html
gchoqueux: > For my information : is this part combined with the scaler method defined from lines 199 to 212 responsible for the buildings 'rising from the ground' effect ?

yes

gchoqueux: > Also : what is the difference between applying a scale to the whole meshes group instead of applying it individually to each mesh within the group ?

because the mesh `Feature` is added to new `Group`.

but yes you're right, it's not very intuitive
I will look for a way to avoid this

examples/view_immersive.html
gchoqueux: yes, i removed it

src/Core/Deprecated/Undeprecator.js
gchoqueux: I changed it but i'm not sure of this part

src/Converter/Feature2Mesh.js
gchoqueux: I'm not sure, because  `new Float32Array(ptsIn.length)` is memory allocation and  `new Float32Array(ptsIn)` is memory copy.

src/Core/Style.js
gchoqueux: yes

src/Core/Style.js
gchoqueux: removed

src/Parser/GeoJsonParser.js
gchoqueux: yes

test/unit/featureprocess.js
gchoqueux: yes

src/Core/Style.js
gchoqueux: yes

src/Core/Style.js
gchoqueux: Thanks!

src/Core/Style.js
gchoqueux: Yes, i changed it

src/Core/Style.js
gchoqueux: yes

src/Core/Deprecated/Undeprecator.js
mgermerie: Now that you added the `return newOptions` again, it shall not pass into the second verification.
Maybe you could do the following ?

```js
if (options.crsOut || options.crsIn) {
   // add all the newOptions creation but remove the 'overrideAltitudeInToZero'
   options = newOptions;
}

// add verification for 'options.out.withAltitude' and 'options.out.withNormal' as they already are implemented

if (options.overrideAltitudeInToZero) {
    console.error('error message');
}

return options;
```

src/Converter/Feature2Mesh.js
mgermerie: Ok, then its a good implementation

src/Core/Style.js
```diff
@@ -137,42 +143,62 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * If `base_altitude` is `undefined`, the original altitude is kept, and if it doesn't exist
+ * then the altitude value is set to 0.
  *
  * @property {Object} point - Point style.
- * @property {string} point.color - The color of the point. Can be any [valid
+ * @property {string|function} point.color - The color of the point. Can be any [valid
  * color string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
  * Default is no value, indicating that no point will be shown.
- * @property {number} point.radius - The radius of the point, in pixel. Default
+ * For a `GeometryLayer`, this opacity property isn't used.
```
mgermerie: This line could be removed too, as it mentions `opacity` property within the `color` property description.

examples/source_stream_wfs_25d.html
mgermerie: Ok, thanks !

src/Core/Style.js
gchoqueux: removed

mgermerie: Thanks

src/Converter/Feature2Mesh.js
```diff
@@ -1,31 +1,20 @@
 import * as THREE from 'three';
 import Earcut from 'earcut';
-import Coordinates from 'Core/Geographic/Coordinates';
 import { FEATURE_TYPES } from 'Core/Feature';
+import { deprecatedFeature2MeshOptions } from 'Core/Deprecated/Undeprecator';
 
-function getProperty(name, options, defaultValue, ...args) {
-    const property = options[name];
+const rColor = new THREE.Color();
```
mgermerie: It's a small detail but maybe `rColor` is not the most suited name for this constant, since it is not always set to a random color.

src/Converter/Feature2Mesh.js
```diff
@@ -188,13 +168,18 @@ function featureToLine(feature, options) {
 
     let lines;
 
+    // TODO CREATE material for each feature
```
mgermerie, line 170: By this `TODO`, do you mean that the `var lineMaterial = new THREE.LineBasicMaterial();` on line 148 should later be here instead ?

src/Converter/Feature2Mesh.js
```diff
@@ -257,14 +252,15 @@ function featureToPolygon(feature, options) {
             console.warn('Feature to Polygon: integer overflow, too many points in polygons');
             break;
         }
-        const color = getProperty('color', options, randomColor, geometry.properties);
+        const context = { globals, properties: () => geometry.properties };
+        const style = feature.style.drawingStylefromContext(context);
+
+        const color = toColor(style.fill.color);
```
mgermerie: Maybe you don't need to create a new constant, since `color` is only used once at line 264 in `fillColorArray(colors, count, color, start);`.

mgermerie: Also : a constant named `color` allready exists (at line 226).

src/Converter/Feature2Mesh.js
```diff
@@ -321,31 +316,33 @@ function featureToExtrudedPolygon(feature, options) {
...
-        const extrude = getProperty('extrude', options, 0, geometry.properties);
+        const context = { globals, properties: () => geometry.properties };
+        const style = feature.style.drawingStylefromContext(context);
 
-        const colorTop = getProperty('color', options, randomColor, geometry.properties);
-        color.copy(colorTop);
-        color.multiplyScalar(0.6);
+        fillColor = toColor(style.fill.color);
+        color.copy(fillColor);
+        color.multiplyScalar(0.5);
```
mgermerie: Maybe you could add a comment to specify that this `color` is applied to the bottom surface of extruded polygons, and that `fillColor` is applied to the top surface.

src/Converter/Feature2Mesh.js
```diff
@@ -161,8 +139,9 @@ function featureToPoint(feature, options) {
     geom.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
     if (batchIds) { geom.setAttribute('batchId', new THREE.BufferAttribute(batchIds, 1)); }
 
+    pointMaterial.size = feature.style.point.radius;
+
     const points = new THREE.Points(geom, pointMaterial);
```
mgermerie: You don't need to declare a constant here. You could simply put : 
```js
return new THREE.Points(geom, pointMaterial);
```

src/Converter/Feature2Mesh.js
```diff
@@ -293,7 +289,6 @@ function featureToPolygon(feature, options) {
     geom.setIndex(new THREE.BufferAttribute(new Uint16Array(indices), 1));
 
     const mesh = new THREE.Mesh(geom, material);
```
mgermerie: Same as in `featureToPoint` method : you don't need a constant declaration here and you could simply put : 
```js
return new THREE.Mesh(geom, material);
```

src/Core/Feature.js
```diff
@@ -148,6 +168,11 @@ export class FeatureGeometry {
         if (this._currentExtent) {
             this._currentExtent.expandByValuesCoordinates(long, lat, alt);
```
mgermerie: The method `Extent.expandByValuesCoordinates` does not take an altitude parameter.

src/Core/Feature.js
```diff
@@ -321,6 +370,59 @@ export class FeatureCollection extends THREE.Object3D {
...
+        };
+    }
+
+    /**
+     * Apply the matrix World inverse on the coordinates.
+     * This method is used when the coordinates is pushed
+     * to transform it in local system.
+     *
+     * @param   {Coordinates}  coordinates  The coordinates
+     * @returns {Coordinates} The coordinates in local sytem
```
mgermerie: Could you replace `sytem` by `system` ?

test/unit/dataSourceProvider.js
```diff
@@ -96,7 +97,7 @@ describe('Provide in Sources', function () {
         featureCountByCb = fc.features.length;
     };
 
-    featureLayer.convert = Feature2Mesh.convert({ color, extrude });
+    featureLayer.convert = Feature2Mesh.convert();
```
mgermerie: The `convert` property could be declared inside of `GeometryLayer` constructor call (at line 75 for example).

test/unit/featureUtils.js
```diff
@@ -12,13 +12,15 @@ describe('FeaturesUtils', function () {
     it('should correctly parse geojson', () =>
         promise.then((collection) => {
             assert.equal(collection.features.length, 3);
+            collection.updateMatrix();
```
mgermerie: Is it mandatory to test the `updateMatrix` method ? If the object returned by the `GeoJsonParser.parse` method has the property tested above (`collection.features.lenght`), that means it must be an instance of `FeatureCollection`. Therefore, it also is an instance of `Three.Object3d`. Testing the `updateMatrix` method appears to be testing Threejs.

Moreover, this `collection` object is not used later in the scope of the `promise.then`, so I don't get the point of this line.

Am I missing something ?

test/unit/geojson.js
```diff
@@ -17,12 +17,12 @@ function parse(geojson) {
 describe('GeoJsonParser', function () {
     it('should set all z coordinates to 1', () =>
         parse(holes).then((collection) => {
-            assert.ok(collection.features[0].vertices.every((v, i) => i == 0 || ((i + 1) % 3) != 0 || v == 0));
+            assert.ok(collection.features[0].vertices.every((v, i) => i == 0 || ((i + 1) % 3) != 0 || (v + collection.position.z) == 0));
```
mgermerie: The condition `i == 0` could be removed, since when this condition is true, the second condition (`(i + 1) % 3 != 0`) is always true as well.

The same goes for the line 25.

test/unit/geojson.js
```diff
@@ -17,12 +17,12 @@ function parse(geojson) {
 describe('GeoJsonParser', function () {
     it('should set all z coordinates to 1', () =>
```
mgermerie: Unless I'm missing something, shouldn't this description be `'should set all z coordinates to 0'` ?

src/Converter/Feature2Mesh.js
gchoqueux: yes

src/Converter/Feature2Mesh.js
gchoqueux: it needs of new PR

src/Converter/Feature2Mesh.js
gchoqueux: yes, I change it

src/Converter/Feature2Mesh.js
gchoqueux: added

gchoqueux:
@mgermerie  thanks for the review !
I push a new version

src/Converter/Feature2Mesh.js
```diff
@@ -411,7 +404,7 @@ function featureToMesh(feature, options) {
             mesh = featureToLine(feature, options);
             break;
         case FEATURE_TYPES.POLYGON:
-            if (options.extrude) {
+            if (feature.style.fill.extrusion_height) {
```
mgermerie, line 404: Could you change the doc for the `featureToMesh` method ?  It still mentions `options.altitude`, `options.extrude` and `options.color`.

src/Core/Deprecated/Undeprecator.js
```diff
@@ -29,27 +29,48 @@ export const deprecatedParsingOptionsToNewOne = (options) => {
         newOptions.out.mergeFeatures = options.mergeFeatures;
         newOptions.out.withNormal = options.withNormal;
         newOptions.out.withAltitude = options.withAltitude;
```
mgermerie: Is there a point in copying `options.withNormal` and `options.withAltitude` parameters in `newOptions` since both these parameters are deprecated ?
Maybe you could instead remove lines 30 and 31, and replace line 32 by : 
```js
if (options.withAltitude && options.withNormal)
```

src/Core/Deprecated/Undeprecator.js
```diff
@@ -29,27 +29,48 @@ export const deprecatedParsingOptionsToNewOne = (options) => {
...
+            console.error('Parsing options out.overrideAltitudeInToZero is removed, use Style.xxx.base_altitude instead');
+        }
         newOptions.out.filter = options.filter;
         return newOptions;
-    } else if (options.out && (options.out.withAltitude !== undefined || options.out.withNormal !== undefined)) {
-        console.warn('Parsing options out.withAltitude and out.withNormal are deprecates, use out.structure: 2d or 3d.');
+    }
+
+    if (options.out && (options.out.withAltitude !== undefined || options.out.withNormal !== undefined)) {
+        console.warn('Parsing options out.withAltitude and out.withNormal are deprecated, use out.structure: 2d or 3d.');
```
mgermerie: Could you replace 'are' by 'is' ?

examples/source_stream_wfs_25d.html
gchoqueux: We keep it for the moment.

gchoqueux:
I refactor the `Undeprecator.js`, there were some issues.

src/Converter/Feature2Mesh.js
mgermerie: Thanks !

mgermerie (approved):

PR #1665, opened 2021-05-28
fix(PlanarControls): fix rounding issues with max and min resolution

### LINKED ISSUE #1659: ortho cam: bug in level Max/Min window
In a orthographic camera view, we can set a maxResolution and a minResolution.
But it appears that the min resolution cannot be reached in my case.

>   const zoomFactor = 2;// customizable
  const view = new itowns.PlanarView(viewerDiv, extent, {
    camera: {
      type: itowns.CAMERA_TYPE.ORTHOGRAPHIC,
    },
    placement,
    maxSubdivisionLevel: 30,
    disableSkirt: true,
    controls: {
      enableSmartTravel: false,
      zoomFactor,
      maxResolution: resolLvMax * zoomFactor ** -1,
      minResolution: resolLvMin * zoomFactor,
    },
  });

In addition it might be a good idea for the 'placement' to be able to define an unique coordinate and a resolution, instead of (or in addition) to an extent.

mgermerie:
## Description

Fix some rounding issues concerning `minResolution` and `maxResolution` parameters.
due to some incorrect rounding, the user could be unable to zoom out to reach a resolution equals to the `minResolution` parameter (or to zoom in to reach a resolution equals to the `maxResolution` parameter).

To fix this, the following is done : 

- The resolution values are all rounded up to 10 to the minus sixth meters.
- The `maxResolution` (which is the minimal value the resolution must take) is rounded bellow.
- The `minResolution` (which is the maximal value the resolution must take) is rounded above.
- The current view resolution is rounded bellow or above depending if it is compared to `maxResolution` or minResolution.

## Motivation and Context

Closes #1659.

## Left to do

- [ ] When zooming in or out in a `PlanarView` with a `PerspectiveCamera`, the distance traveled does not strictly matches the value implied by the `zoomFactor` parameter. This imprecision causes the same issue as mentioned in #1659.

### DIFF
--- a/examples/misc_orthographic_camera.html
+++ b/examples/misc_orthographic_camera.html
@@ -46,6 +46,8 @@
             // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
             const viewerDiv = document.getElementById('viewerDiv');
 
+            const startResolution = 3.5429583702391496;
+
             // instantiate PlanarView with an orthographic camera.
             // The extent passed in `placement` is the area that will be displayed by camera.
             const view = new itowns.PlanarView(viewerDiv, extent, {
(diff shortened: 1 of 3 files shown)


examples/misc_orthographic_camera.html
```diff
@@ -46,15 +46,19 @@
             // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
             const viewerDiv = document.getElementById('viewerDiv');
 
+            const startResolution = 3.5429583702391496;
```
gchoqueux, line 49: why this is value?

examples/misc_orthographic_camera.html
```diff
@@ -74,6 +78,25 @@
             });
             view.addLayer(wmsImageryLayer);
 
+            var wmsElevationSource = new itowns.WMSSource({
```
gchoqueux, line 81: why add `ElevationLayer`?

gchoqueux:
@mgermerie could you refresh this PR?

jailln:
@mgermerie are you still working on it ? is the issue and the PR still relevant ? If yes, can you refresh it ? Otherwise we will close it.

PR #1672, opened 2021-06-10
fix(FeatureToolTip): fix tooltip legend icon in examples

mgermerie:
## Description

Fix `FeatureToolTip` `fillToolTip` method in examples to correct issues with tool tips legend.

## Motivation and Context

In examples where the tool tip is used (such as [this](http://www.itowns-project.org/itowns/examples/#source_file_shapefile) or [this](http://www.itowns-project.org/itowns/examples/#source_file_kml_raster) examples), the style of the legend within the tooltip could be wrongly computed. It could sometimes cause `undefined` errors, and also have a legend whose style wouldn't match the style of the associated feature.

### DIFF
--- a/examples/js/plugins/FeatureToolTip.js
+++ b/examples/js/plugins/FeatureToolTip.js
@@ -84,17 +84,23 @@ var FeatureToolTip = (function _() {
             style = (geometry.properties && geometry.properties.style) || feature.style || layer.style;
             var context = { globals: {}, properties: getGeometryProperties(geometry) };
             style = style.drawingStylefromContext(context);
-            fill = style.fill.color;
-            stroke = '1.25px ' + style.stroke.color;
 
             if (feature.type === itowns.FEATURE_TYPES.POLYGON) {
                 symb = '&#9724';
+                if (style) {
+                    fill = style.fill && style.fill.color;
+                    stroke = style.stroke && ('1.25px ' + style.stroke.color);
+                }
             } else if (feature.type === itowns.FEATURE_TYPES.LINE) {
                 symb = '&#9473';
-                fill = style.stroke.color;
+                fill = style && style.stroke && style.stroke.color;
                 stroke = '0px';
             } else if (feature.type === itowns.FEATURE_TYPES.POINT) {
                 symb = '&#9679';
+                if (style && style.point) {  // Style and style.point can be undefined if no style options were passed
+                    fill = style.point.color;
+                    stroke = '1.25px ' + style.point.line;
+                }
             }
 
             content += '<div>';


examples/js/plugins/FeatureToolTip.js
```diff
@@ -84,17 +84,21 @@ var FeatureToolTip = (function _() {
             style = (geometry.properties && geometry.properties.style) || feature.style || layer.style;
             var context = { globals: {}, properties: getGeometryProperties(geometry) };
             style = style.drawingStylefromContext(context);
-            fill = style.fill.color;
-            stroke = '1.25px ' + style.stroke.color;
 
             if (feature.type === itowns.FEATURE_TYPES.POLYGON) {
                 symb = '&#9724';
+                fill = style.fill.color;
```
gchoqueux: could you add check if `style` and `style.fill` exits?

examples/js/plugins/FeatureToolTip.js
```diff
@@ -84,17 +84,21 @@ var FeatureToolTip = (function _() {
...
+                stroke = style.stroke && ('1.25px ' + style.stroke.color);
             } else if (feature.type === itowns.FEATURE_TYPES.LINE) {
                 symb = '&#9473';
                 fill = style.stroke.color;
```
gchoqueux: could you add check if `style` and `style.stroke` exits?

examples/js/plugins/FeatureToolTip.js
mgermerie: I don't know if this is necessary in curent itowns version.
If I'm not mistaking, `FeatureToolTip` doesn't work on polygons displayed within a `GeometryLayer`. It thus only work in a `ColorLayer`.
If you display some polygons within a `ColorLayer` without specifying a `fill` property for the `Style` of the layer, the polygons won't be displayed. Therefore, I believe that there is no case where you try to display a tool tip on a polygon and the `Style.fill` property is undefined.
Should I add a check for `style` and `style.fill` anyways ?

gchoqueux: The polygon isn't displayed but polygon exists in `FeatureCollection`?
Could you try, in example, if you remove `fill.color` in style?

mgermerie: Actually, after checking acording to your latest indications, you are right : 
- On the [kml example](http://www.itowns-project.org/itowns/examples/index.html#source_file_kml_raster), the polygons appear even if they is no definition for `fill`.
- On the [geojson example](http://www.itowns-project.org/itowns/examples/index.html#source_file_geojson_raster), the polygons don't appear when there is no `fill`, `stroke`, or even no `Style` definition in the layer. Yet, they exists in `FeatureCollection`. Therefore a tooltip is displayed when hovering their position with cursor (event if the polygons are not visible).

I shall add the checks for `style`, `style.fill` and `style.stroke`

mgermerie: Furthermore, if you add a polygon with no `fill` but a `stroke`, the tool tip cannot display the legend correctly if `style.fill` is not tested, hence the need of a check as you suggested.

examples/js/plugins/FeatureToolTip.js
mgermerie: added

gchoqueux (approved):

PR #1674, opened 2021-06-15
example(vectorTile): replace expired source in vector tile examples

mgermerie:
## Description

Replace the source of the data displayed in vector tile [2d](http://www.itowns-project.org/itowns/examples/index.html#vector_tile_raster_2d) and [3d](http://www.itowns-project.org/itowns/examples/index.html#vector_tile_raster_3d) examples.

## Motivation and Context

The source that was used is no longer available.

### DIFF
--- a/examples/vector_tile_raster_2d.html
+++ b/examples/vector_tile_raster_2d.html
@@ -46,18 +46,11 @@
 
             setupLoadingScreen(viewerDiv, view);
 
+            // Defines a VectorTilesSource to load Vector Tiles data from the geoportail
             var mvtSource = new itowns.VectorTilesSource({
-                style: 'https://raw.githubusercontent.com/Oslandia/postile-openmaptiles/master/style.json',
-                // eslint-disable-next-line no-template-curly-in-string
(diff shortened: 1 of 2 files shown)


examples/vector_tile_raster_2d.html
```diff
@@ -46,17 +46,15 @@
...
-                zoom: {
-                    min: 2,
-                    max: 14,
+                style: 'https://wxs.ign.fr/static/vectorTiles/styles/PLAN.IGN/standard.json',
+                filter: (layer) => {
+                    // We don't display mountains related data to ease visualisation
+                    if (layer['source-layer'].includes('oro_')) {
+                        return false;
+                    }
+                    return true;
```
gchoqueux: Could you simplify method filter?
```js
// We don't display mountains related data to ease visualisation
filter: (layer) => !layer['source-layer'].includes('oro_'),
```

mgermerie: I simplified it, thanks !

gchoqueux (approved):

PR #1680, opened 2021-06-30
feature(Source): add encoding parameter for x-bil data

mgermerie:
## Description

Add `encoding` property to `Source`. This property defines the number of bits the source data are encoded on. It only works with `image/x-bil` source data.

### DIFF
--- a/examples/mars.html
+++ b/examples/mars.html
@@ -65,7 +65,6 @@
             view.addLayer(tmsMarsLayer);
 
             const demMars = new itowns.ElevationLayer('demMars', {
-                useColorTextureElevation: true,
                 colorTextureElevationMinZ: 0,
                 colorTextureElevationMaxZ: 20000,
                 crs: 'EPSG:4326',
@@ -74,6 +73,7 @@
                     crs: 'EPSG:4326',
                     url : 'https://api.nasa.gov/mars-wmts/catalog/Mars_MGS_MOLA_DEM_mosaic_global_463m_8/1.0.0//default/default028mm/${z}/${y}/${x}.png',
                     zoom: { min: 0, max: 5 },
+                    encoding: 'RGB',
                 }),
             });
             // view.addLayer(demMars);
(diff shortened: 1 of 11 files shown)


gchoqueux (commented):
* handling deprecated  `useColorTextureElevation`
* add documentation

examples/mars.html
```diff
@@ -74,6 +73,7 @@
                     crs: 'EPSG:4326',
                     url : 'https://api.nasa.gov/mars-wmts/catalog/Mars_MGS_MOLA_DEM_mosaic_global_463m_8/1.0.0//default/default028mm/${z}/${y}/${x}.png',
                     zoom: { min: 0, max: 5 },
+                    encoding: 'RGB',
```
gchoqueux, line 76: It's not `RGB` but grey rgb

src/Core/View.js
```diff
@@ -311,7 +311,8 @@ class View extends THREE.EventDispatcher {
                 } else {
                     return layer._reject(new Error(`Cant add color layer ${layer.id}: the maximum layer is reached`));
                 }
-            } else if (layer.isElevationLayer && layer.source.format == 'image/x-bil;bits=32') {
+            } else if (layer.isElevationLayer) {
+                layer.source.networkOptions.encoding = layer.source.encoding;
```
gchoqueux: I think it will better to add a new `fileOptions`.

src/Layer/ElevationLayer.js
```diff
@@ -53,12 +51,20 @@ class ElevationLayer extends RasterLayer {
      * view.addLayer(elevation);
      */
     constructor(id, config = {}) {
+        if (config.useColorTextureElevation !== undefined) {
```
gchoqueux, line 54: use undeprecator

Desplandis:
@mgermerie Have we encounter some use-cases where elevation data is provided as 8bits or 16bits BIL? I think we'll depreciate the BIL format in future versions in favor of more "standard" formats (replacing it with COG and Mapbox RGB). So this PR should be closed? WDYT?

mgermerie:
I must admit I don't remember the use-case. Maybe @gchoqueux remembers ?

Otherwise, should we drop BIL support, I agree to close the PR.

PR #1684, opened 2021-07-07
Feature(Style): add support for custom icon in labels

mgermerie:
## Description

Add an `icon` parameter in `Style`, allowing user to define the `source`, the `key`, the `anchor` and the `size` of the icons displayed within labels.

The anchoring of icons within labels is now independent from the label text anchoring. This allows passing different `anchor` parameters for `Style.text` and `Style.icon`.

The boundaries of a label now also includes the icon it may contain. This results in icons no more superposing other label text.

## Screenshot
The picture bellow shows the [shapefile example](https://github.com/iTowns/itowns/blob/master/examples/source_file_shapefile.html) where the `Style` for the _velib_ `ColorLayer` has been modified as such : 

```js
var velibStyle = new itowns.Style({
    zoom: { min: 10, max: 20 },
    text: {
        field: '{name}\n(id: {station_id})',
        size: 14,
        haloColor: 'white',
        haloWidth: 1,
        font: ['monospace'],
        justify: left,
        anchor: 'top-left',
    },
    icon: {
        source: '{link to the position marker image}',
        anchor: 'bottom',
    },
});     
```

Capture d’écran du 2021-07-07 14-30-17

## Note
This PR is rebased on #1683, as it changes labels boundaries, and #1683 make label position computing independent from its boundaries.

### DIFF
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -102,6 +102,9 @@ class Label extends THREE.Object3D {
             this.styleOffset = [0, 0];
         }
 
+        this.icon = this.content.getElementsByClassName('itowns-icon')[0];
+        this.iconOffset = { left: 0, right: 0, top: 0, bottom: 0 };
+
         this.zoom = {
             min: style.zoom && style.zoom.min != undefined ? style.zoom.min : 2,
             max: style.zoom && style.zoom.max != undefined ? style.zoom.max : 24,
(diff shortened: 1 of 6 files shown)


src/Core/Label.js
```diff
@@ -130,15 +133,31 @@ class Label extends THREE.Object3D {
...
             this.boundaries.bottom = y + this.offset.bottom + this.padding;
+
+            // The boundaries of the label are the union of the boundaries of the text
+            // and the boundaries of the icon, if it exists
+            // TODO : Maybe these four line could be within a `if (this.icon)` condition to
+            //  avoid computing when no icon is defined ???
+            this.boundaries.left = Math.min(this.boundaries.left, this.boundaries.left + this.iconOffset.left);
+            this.boundaries.right = Math.min(this.boundaries.right, this.boundaries.right + this.iconOffset.right);
+            this.boundaries.top = Math.min(this.boundaries.top, this.boundaries.top + this.iconOffset.top);
+            this.boundaries.bottom = Math.min(this.boundaries.bottom, this.boundaries.bottom + this.iconOffset.bottom);
```
gchoqueux: and `Math.max` ?

src/Core/Label.js
```diff
@@ -157,6 +176,20 @@ class Label extends THREE.Object3D {
             };
             this.offset.right = this.offset.left + width;
             this.offset.bottom = this.offset.top + height;
+
+            if (this.icon) {
+                rect = this.icon.getBoundingClientRect();
+                const x = Math.round(rect.x);
+                const y = Math.round(rect.y);
+                const width = Math.round(rect.width);
+                const height = Math.round(rect.height);
```
gchoqueux: don't use `const`, compute  directly with `round` and `floor` in `this.iconOffset` properties.

src/Core/Label.js
mgermerie: I changed this part

src/Core/Label.js
mgermerie: I removed the `const`

src/Core/Label.js
```diff
@@ -130,15 +133,33 @@ class Label extends THREE.Object3D {
             this.boundaries.right = x + this.offset.right + this.padding;
             this.boundaries.top = y + this.offset.top - this.padding;
             this.boundaries.bottom = y + this.offset.bottom + this.padding;
+
+            // The boundaries of the label are the union of the boundaries of the text
+            // and the boundaries of the icon, if it exists.
+            // Checking if this.icon is not only zeros is mandatory, to prevent case
+            // when a boundary is set to x or y coordinate
+            if (JSON.stringify(this.iconOffset) !== JSON.stringify({ left: 0, right: 0, top: 0, bottom: 0 })) {
```
gchoqueux: could you check properties?

mgermerie: I now check for all `left`, `right`, `top` and `bottom` properties.

gchoqueux (approved):

PR #1688, opened 2021-07-08
feature(Label): add support to pass custom domElements to labels 

mgermerie:
## Description

Add `domElement` parameter to `Style.text`. It can be set to a `DomElement` (like a `div` for instance) which will be displayed as a `Label`.

Add an example of label creation from mouse clicking. The created label is set with the `Style.text.domElement` parameter.

## Screenshots

The image bellow shows a custom `Label` displayed in the example.

Capture d’écran du 2021-07-08 17-34-31

## Note

This PR is rebased on #1686, as it needs to pass a custom `anchor` to the custom label.

### DIFF
--- a/examples/misc_custom_label.html
+++ b/examples/misc_custom_label.html
@@ -1,0 +1,148 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Custom popup</title>
+    <meta charset="UTF-8">
+    <link rel="stylesheet" type="text/css" href="css/example.css">
+    <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+    <style>
+        .bubble {
+            background-color: #fff;
+            border-radius: 10px;
+            text-align: center;
+            padding: 10px 15px;
+            box-shadow: 0 0.125rem 0.5rem rgba(0, 0, 0, .3), 0 0.0625rem 0.125rem rgba(0, 0, 0, .2);
+        }
+        .pointer {
+            height: 12px;
+            width: 12px;
+            background-color: #fff;
+            transform: translate(-50%, -50%) rotate(45deg);
+            position: relative;
+            left: 80%;
+        }
+    </style>
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+</head>
+<body>
+<div id="description">
+    Clicking a position while holding `p` key down will pop a custom popup.
+</div>
+<div id="viewerDiv" class="viewer"></div>
+
+<script src="js/GUI/GuiTools.js"></script>
+<script src="../dist/itowns.js"></script>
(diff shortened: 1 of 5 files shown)


examples/misc_custom_label.html
```diff
@@ -0,0 +1,150 @@
...
+    pointer.classList.add('pointer');
+    customDiv.appendChild(pointer);
+
+    // Define a method to set the content of custom label
+    function setLabelContent(properties) {
+        const coord = properties.position.as('EPSG:4326');
+        bubble.textContent = `
+            lon : ${Math.round(coord.x * 100) / 100}° |
+            lat : ${Math.round(coord.y * 100) / 100}° |
+            alt : ${Math.round(coord.z)}m `;
```
gchoqueux, line 79: could you move `textContent` in feature properties?
and use `style.text.field` instead?

examples/misc_custom_label.html
```diff
@@ -0,0 +1,150 @@
...
+        // the source of the feature layer
+        const source = new itowns.FileSource({ features });
+
+        // create labelLayer
+        const layer = new itowns.LabelLayer('idLabelLayer', {
+            source: source,
+            style: new itowns.Style({
+                text: {
+                    anchor: [-0.8, -1],
+                    domElement: setLabelContent,
```
gchoqueux: could you move `domElement` to `LabelLayer` parameter?

src/Core/Label.js
```diff
@@ -79,24 +79,32 @@ class Label extends THREE.Object3D {
         this.projectedPosition = { x: 0, y: 0 };
         this.boundaries = { left: 0, right: 0, top: 0, bottom: 0 };
 
-        this.content = document.createElement('div');
+        if (style.text && style.text.domElement) {
+            this.content = style.text.domElement.cloneNode(true);
```
gchoqueux: use `content` parameter to pass `domElement` instead of `style.text`

src/Core/Style.js
```diff
@@ -656,9 +674,11 @@ class Style {
      *
      * @param {Object} ctx - An object containing the feature context.
      *
-     * @return {string} The formatted string.
+     * @return {string|undefined} The formatted string if `style.text.field` is defined, nothing otherwise.
      */
     getTextFromProperties(ctx) {
+        if (!this.text.field) { return; }
```
gchoqueux, line 698: why you return `undefined` if there is no `this.text.field` ?

mgermerie: I changed it so the method doesn't trigger an error when user set `Style.text` without setting `Style.text.field`.
For example, one could declare : 
```js
const layer = new LabelLayer('id', {
    source,
    domElement,
    style: new itowns.Style({
        text: { anchor : "top" },
    }),
};
```

Before this modification, setting `Style.text.field` was mandatory if you wanted to set some other properties of `Style.text`.

As I did not yet find a way of moving the anchor in the domElement css style, this part is unchanged. I will try to find a way of doing so.

gchoqueux (approved):

PR #1700, opened 2021-08-04
Use the same wheel event across all browsers

aakaydin:
## Description
iTowns uses 'mouseWheel' event which is deprecated and not supported by FireFox and uses 'MozMousePixelScroll' event for Firefox. I replaced this event with the 'wheel' event which is supported in all browsers. On the event we check the value of deltaY property to determine zooming also changed the unit tests to work with this attribute.

old event: https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event
new one: https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event

## Motivation and Context
I encountered this problem while trying to add listeners for iTowns on a canvas element which didn't register 'mouseWheel' event. By changing this we use the same code across all browsers.

### DIFF
--- a/test/unit/globecontrol.js
+++ b/test/unit/globecontrol.js
@@ -180,13 +180,13 @@ describe('GlobeControls', function () {
 
     it('mouse wheel', function () {
         const startRange = controls.getRange();
-        event.wheelDelta = -10;
-        controls.onMouseWheel(event);
-        assert.ok(controls.getRange() > startRange);
-        event.wheelDelta = 10;
-        controls.onMouseWheel(event);
+        event.deltaY = -10;
         controls.onMouseWheel(event);
         assert.ok(controls.getRange() < startRange);
+        event.deltaY = 10;
+        controls.onMouseWheel(event);
+        controls.onMouseWheel(event);
+        assert.ok(controls.getRange() > startRange);
     });
 
     it('travel in', function (done) {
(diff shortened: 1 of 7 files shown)


test/unit/globecontrol.js
```diff
@@ -180,10 +180,10 @@ describe('GlobeControls', function () {
 
     it('mouse wheel', function () {
         const startRange = controls.getRange();
-        event.wheelDelta = -10;
+        event.deltaY = -10;
         controls.onMouseWheel(event);
         assert.ok(controls.getRange() > startRange);
```
mgermerie: Could you please switch the `>` for a `<` ? 
The sign of `deltaY` for a `wheel` event is the opposite of the `wheelDelta` for a `mouseWheel` event. Having 
```js
event.deltaY = -10
```
on line 183 means we are now zooming in, thus the range after zooming shall be smaller than before.

test/unit/globecontrol.js
```diff
@@ -180,10 +180,10 @@ describe('GlobeControls', function () {
...
-        event.wheelDelta = 10;
+        event.deltaY = 10;
         controls.onMouseWheel(event);
         controls.onMouseWheel(event);
         assert.ok(controls.getRange() < startRange);
```
mgermerie: Could you please switch `<` for `>`, because of the same reasons mentioned in the previous comment ?

mgermerie (commented):
Hi ! Thanks for this useful PR ! 
Could you please fix the unit tests following the comments I posted ?

Could you also squash your commits into one and name this single commit following the Angular convention ? I this case, you could have something close to your first commit message but with the relevant prefix : `refacto(Controls): change deprecated mouseWheel event to wheel event`.

aakaydin:
I just changed the unit test and squashed the commits together 
let me know if anything needs to be done

mgermerie:
Thanks for the update, it's all good for me !

mgermerie (approved):

PR #1701, opened 2021-08-09
examples: define z-index for minimap

mgermerie:
## Description

Enforce minimap `z-index` css property to a high value to make sure that the minimap will always appear on top of other objects, however the structure of the html page.


mgermerie:
There is actually no point in defining a hard-coded value of `z-index` property for the minimap `div` in iTowns source code. Users will need to redefine this `z-index` at their convenience.

PR #1703, opened 2021-08-16
fix(Label): fix size of labels div

mgermerie:
## Description

Reduce labels div height to 0, and let Labels overflow this container. This fixes an issue where user could not have `mouseover` nor `mouseclick` events triggered on the canvas. This issue mainly occurred using Potree tools when integrating Potree in iTowns.


mgermerie:
Not a modification proper to iTowns, but rather specific to the integration of Potree within iTowns. As this integration is not fully supported yet, this PR is irrelevant.

PR #1709, opened 2021-08-26
refacto(StateControl): simplify `setFromOptions` method

mgermerie:
## Description

Simplify the `setFromOptions` method of `StateControl`.
Now, the `enable` property of each `State` does not need to be redefined when using the method. For example, one could now use :
```js
view.controls.states.setFromOptions({
    PAN: { mouseButton: itowns.THREE.MOUSE.LEFT },
    MOVE_GLOBE: { mouseButton: itowns.THREE.MOUSE.RIGHT },
});
```
to switch pan and drag movement controls, where it was previously needed to call :
```js
view.controls.states.setFromOptions({
    PAN: {
        enable: true, 
        mouseButton: itowns.THREE.MOUSE.LEFT,
    },
    MOVE_GLOBE: {
        enable: true, 
        mouseButton: itowns.THREE.MOUSE.RIGHT,
    },
});
```
---
Fix minor glitches on documentation.

### DIFF
--- a/src/Controls/StateControl.js
+++ b/src/Controls/StateControl.js
@@ -12,6 +12,54 @@ const CONTROL_KEYS = {
 };
 
 
+const DEFAULT_STATES = {
+    ORBIT: {
+        enable: true,
+        mouseButton: THREE.MOUSE.LEFT,
+        double: false,
+        keyboard: CONTROL_KEYS.CTRL,
+        finger: 2,
+    },
+    MOVE_GLOBE: {
+        enable: true,
+        mouseButton: THREE.MOUSE.LEFT,
+        double: false,
+        finger: 1,
+    },
+    DOLLY: {
+        enable: true,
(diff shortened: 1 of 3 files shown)


src/Controls/StateControl.js
```diff
@@ -12,6 +12,73 @@ const CONTROL_KEYS = {
...
+            && options[state] && options[state][key] === undefined
+        ) {
+            newState[key] = stateControls[state][key];
+        }
+    }
+
+    stateControls[state] = newState;
+}
+
+
```
gchoqueux: could move the method in `setFromOptions`?
Could you keep only `enabled` from previous trigger?

mgermerie: I added the modifications.

gchoqueux (approved):

gchoqueux (approved):

PR #1713, opened 2021-09-07
Refacto(Controls): switch keyboard management to `StateControl`

mgermerie:
## Description

Switch keyoard input management from `GlobeControls` to `StateControl`.

A `StateControl` instance now listens for keyboard events. When a keyboard events occurs, `StateControl` determines which state the event is associated to. It then dispatches a custom event, whose type relates to the determined state. `GlobeControls` listens for every `StateControl` custom events, and moves camera accordingly.

## Motivation and Context

1. Simplify GlobeControls by removing input management from it.
2. Factorize input management in StateControl so that other controls (like PanarControls) can later be refactored to implement it.

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -154,7 +154,6 @@ let previous;
  * @param      {number}  options.minDistance Minimum distance between ground and camera
  * @param      {number}  options.maxDistance Maximum distance between ground and camera
  * @param      {bool}  options.handleCollision enable collision camera with ground
- * @property   {bool} enabled enable control
  * @property   {number} minDistance Minimum distance between ground and camera
  * @property   {number} maxDistance Maximum distance between ground and camera
  * @property   {number} zoomSpeed Speed zoom with mouse
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -8,6 +8,13 @@ import StateControl from 'Controls/StateControl';
 // private members
 const EPS = 0.000001;
 
+const direction = {
+    up: new THREE.Vector2(0, 1),
+    bottom: new THREE.Vector2(0, -1),
+    left: new THREE.Vector2(1, 0),
+    right: new THREE.Vector2(-1, 0),
+};
+
 // Orbit
 const rotateStart = new THREE.Vector2();
 const rotateEnd = new THREE.Vector2();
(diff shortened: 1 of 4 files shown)


src/Controls/GlobeControls.js
```diff
@@ -604,10 +615,15 @@ class GlobeControls extends THREE.EventDispatcher {
...
 
     handlePan(event) {
-        panEnd.copy(event.viewCoords);
-        panDelta.subVectors(panEnd, panStart);
+        if (event.viewCoords) {
+            panEnd.copy(event.viewCoords);
+            panDelta.subVectors(panEnd, panStart);
+            panStart.copy(panEnd);
+        } else if (event.direction) {
+            panDelta.copy(direction[event.direction].clone().multiplyScalar(this.keyPanSpeed));
```
gchoqueux: could you remove `clone` calling?

src/Controls/StateControl.js
```diff
@@ -41,17 +41,19 @@ const DEFAULT_STATES = {
...
-        up: CONTROL_KEYS.UP,
-        bottom: CONTROL_KEYS.BOTTOM,
-        left: CONTROL_KEYS.LEFT,
-        right: CONTROL_KEYS.RIGHT,
+        keyboard: {
+            up: CONTROL_KEYS.UP,
+            bottom: CONTROL_KEYS.BOTTOM,
+            left: CONTROL_KEYS.LEFT,
+            right: CONTROL_KEYS.RIGHT,
+        },
```
gchoqueux: could you define 4 triggers to each pan direction?

src/Controls/StateControl.js
```diff
@@ -310,6 +317,10 @@ class StateControl extends THREE.EventDispatcher {
             // TODO : add touch event management
             default:
         }
+
+        window.removeEventListener('keydown', this._onKeyDown, false);
+        window.removeEventListener('keyup', this._onKeyUp, false);
+
```
gchoqueux: could you figure out if there's a simple way?

gchoqueux (commented):
could you move `handleMouseDown` in `onPointerDown`?

src/Controls/GlobeControls.js
mgermerie: I removed it.

src/Controls/StateControl.js
mgermerie: I split Pan in  4 triggers.

src/Controls/StateControl.js
mgermerie: I removed those lines and added a check in `onKeyUp` : if the mouse is pressed, the `currentState` is not switched to `NONE`.

gchoqueux (approved):
thx

PR #1715, opened 2021-09-16
feature(Style): points can now be displayed as icons

mgermerie:
## Description

Add support to display points as label icons in a `ColorLayer` or a `LabelLayer`.

If user defines `Style.point` of a `ColorLayer`, and set the `ColorLayer` `addLabelLayer` parameter to true, points will be displayed as label icons. If `addLabelLayer` is unset or set to false, point will be kept rasterized.
User can also define `Style.point` for a `LabelLayer`, in which case points will obviously be displayed as label icons.

### DIFF
--- a/src/Layer/LabelLayer.js
+++ b/src/Layer/LabelLayer.js
@@ -12,6 +12,9 @@ const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 
 const _extent = new Extent('EPSG:4326', 0, 0, 0, 0);
 
+const divDomElement = document.createElement('div');
+
+
 /**
  * A layer to handle a bunch of `Label`. This layer can be created on its own,
  * but it is better to use the option `addLabelLayer` on another `Layer` to let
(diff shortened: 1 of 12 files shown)


src/Layer/LabelLayer.js
```diff
@@ -57,6 +57,49 @@ class LabelLayer extends Layer {
...
+        ) {
+            // Create a canvas
+            const canvas = document.createElement('canvas');
+
+            // Set canvas dimensions
+            const width = this.style.point.radius + this.style.point.width;
+            canvas.width = 2 * width;
+            canvas.height = 2 * width;
+
+            if (canvas.getContext) {
```
gchoqueux: could you use `feature2Texture` ?

src/Layer/LabelLayer.js
```diff
@@ -57,6 +57,49 @@ class LabelLayer extends Layer {
...
+                if (this.style.point.line) {
+                    ctx.lineWidth = this.style.point.width;
+                    ctx.strokeStyle = this.style.point.line;
+                    ctx.stroke();
+                }
+
+                this.style.icon = this.style.icon || {};
+                this.style.icon.source = canvas.toDataURL();
+                this.style.point.color = undefined;
+                this.style.point.line = undefined;
```
gchoqueux: I'm not sure to keep this part....

src/Layer/LabelLayer.js
mgermerie: I now use it.

src/Layer/LabelLayer.js
mgermerie: Changing `this.style.icon.source` is necessary so that the canvas drawing is displayed as icon. The following two lines prevent drawing points projected on ground, as can be seen in the attached screenshot. Moreover, they do not actually change the `Style` object defined by the user. Do you still think I should remove them ?
Capture d’écran du 2022-02-15 12-00-18

PR #1728, opened 2021-10-18
fix: resolve deprecated usage of function Three.Quaternion.slerp()

gnominomnom:
## Description
Use of Three.Quaternion.slerpQuaternions() in src/Controls/GlobeControls.js instead of deprecated function slerp()

## Motivation and Context

A function was deprecated in src/Controls/GlobeControls.js and had to be replaced.

Browser : Mozilla Firefox 93.0

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -519,7 +519,7 @@ class GlobeControls extends THREE.EventDispatcher {
         } else {
             sphericalDelta.theta *= (1 - dampingFactorDefault);
             sphericalDelta.phi *= (1 - dampingFactorDefault);
-            moveAroundGlobe.slerp(dampingMove, this.dampingMoveFactor * 0.2);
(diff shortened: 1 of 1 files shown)


mgermerie:
Hi,
Thanks for your contribution on improving iTowns.

However, I believe you might have gotten confused with three.js `Quaternion.slerp` methods.

Indeed, three used to provide users with a static `slerp` method that was depreciated in [this three PR](https://github.com/mrdoob/three.js/pull/21532). In `GlobeControls`, we use the non static `slerp` method, whose documentation can be read [here](https://threejs.org/docs/?q=Quaternion#api/en/math/Quaternion.slerp). 

This non static method is still in three, and not deprecated. It is actually called in the `Quaternion.slerpQuaternions` method. You can see it in the following code which quotes three [Quaternion.js file](https://github.com/mrdoob/three.js/blob/master/src/math/Quaternion.js) :
```js
slerpQuaternions( qa, qb, t ) {

	this.copy( qa ).slerp( qb, t );

}
```

PR #1729, opened 2021-10-21
fix vertical problems in polygons triangulation

gmaillet:
## Description

Using an average plane for polygons triangulation.

## Motivation and Context

In featureToPolygon (Feature2Mesh.js) polygons are triangulate with earcut to create a BufferGeometry. This is a 2D triangulation method and it doesn't work on vertical faces.


src/Parser/GeoJsonParser.js
```diff
@@ -67,7 +67,15 @@ const toFeature = {
 
         // Then read contour and holes
         for (let i = 0; i < coordsIn.length; i++) {
-            this.populateGeometry(crsIn, coordsIn[i], geometry, feature);
+            // GeoJson standard: The first and last positions are equivalent,
+            // and they MUST contain identical values; their representation SHOULD also be identical.
+            const ring = coordsIn[i];
+            if (ring.length > 1) {
+                if (JSON.stringify(ring[0]) === JSON.stringify(ring[ring.length - 1])) {
```
gchoqueux: it's recommanded to use `ring[0].every((v, i) => v == ring[ring.length - 1][i]`

src/Converter/Feature2Mesh.js
```diff
@@ -274,8 +274,29 @@ function featureToPolygon(feature, options) {
...
+        // const pt0 = new THREE.Vector3(geomVertices[0], geomVertices[1], geomVertices[2]);
+        for (let i = 0; i < count; i++) {
+            const pt = new THREE.Vector3(geomVertices[3 * i],
+                geomVertices[3 * i + 1],
+                geomVertices[3 * i +  2]);
+            pt.applyQuaternion(quaternion);
+            geomVertices[3 * i] = pt.x;
+            geomVertices[3 * i + 1] = pt.y;
+            geomVertices[3 * i + 2] = pt.z;
+        }
```
gchoqueux: Avoid to instance object in method or loop, to reduce the garbage collector.

I simplified the algorithm with THREE.js methods

```js
// before the method
const quaternion = new THREE.Quaternion();
const normal = new THREE.Vector3(0, 0, 0);
const v0 = new THREE.Vector3(0, 0, 0);
const v1 = new THREE.Vector3(0, 0, 0);
const vZ = new THREE.Vector3(0, 0, 1);

// ######################

function featureToPolygon(feature, options) {
///....
// transform all vertices to XY plane
// calculate average plane from points
// http://www.les-mathematiques.net/phorum/read.php?13,728203,728209#msg-728209
const size = count * 3;
for (let i = 3; i < size; i += 3) {
    v0.fromArray(geomVertices, i - 3);
    v1.fromArray(geomVertices, i);
    normal.x += (v0.y - v1.y) * (v0.z + v1.z);
    normal.y += (v0.z - v1.z) * (v0.x + v1.x);
    normal.z += (v0.x - v1.x) * (v0.y + v1.y);
}

// calculate normal plane
normal.normalize();
// calculate quaternion to transform the normal plane to z axis
quaternion.setFromUnitVectors(normal, vZ);

// apply transformation to vertices
for (let i = 0; i < size; i += 3) {
    v1.fromArray(geomVertices, i).applyQuaternion(quaternion);
    v1.toArray(geomVertices, i);
}
//...
}
```

gchoqueux (commented):
May add option if the user knows that vertices are in plane XY
to avoid a transformation

gchoqueux:
may be invert default value for option.
How pass option in layer?

gchoqueux:
could you add example?

PR #1731, opened 2021-10-26
Fix scale elevation updating

gchoqueux:
## Description
The scale elevation updating isn't really correctly impacted on the terrain tiles.

This update has re-factorized to use events instead of `Object3D.traverse`.

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -57,8 +57,9 @@ class TileMesh extends THREE.Mesh {
         if (min == null && max == null) {
             return;
         }
-        // FIXME: Why the floors ? This is not conservative : the obb may be too short by almost 1m !
-        if (Math.floor(min) !== Math.floor(this.obb.z.min) || Math.floor(max) !== Math.floor(this.obb.z.max)) {
+        // update bbox if min or max have changed by at least one decimal
+        // or if scale changed
+        if (min.toFixed(1) !== this.obb.z.min.toFixed(1) || max.toFixed(1) !== this.obb.z.max.toFixed(1) || scale != this.obb.z.scale) {
             this.obb.updateZ(min, max, scale);
             if (this.horizonCullingPointElevationScaled) {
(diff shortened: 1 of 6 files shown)


src/Core/TileMesh.js
```diff
@@ -58,7 +58,7 @@ class TileMesh extends THREE.Mesh {
             return;
         }
         // FIXME: Why the floors ? This is not conservative : the obb may be too short by almost 1m !
```
jailln: Could you take this opportunity to fix this FIXME ?

src/Layer/ElevationLayer.js
```diff
@@ -88,9 +74,16 @@ class ElevationLayer extends RasterLayer {
         const updateBBox = () => node.setBBoxZ(rasterElevationNode.min, rasterElevationNode.max, this.scale);
         updateBBox();
 
-        // listen elevation updating
+        // listen scaling updating
```
jailln: this is still an elevation update isn't it ?

src/Process/ObjectRemovalHelper.js
```diff
@@ -27,6 +27,7 @@ export default {
                 }
                 // obj.material = null;
```
jailln: to remove ?

src/Core/TileMesh.js
gchoqueux: I'm going to look at this

src/Layer/ElevationLayer.js
gchoqueux: yes, I change it

src/Process/ObjectRemovalHelper.js
gchoqueux: removed

src/Core/TileMesh.js
gchoqueux: I removed the `Math.floor`, it's useless

jailln: shouldn't we use an epsilon instead ?

gchoqueux: yes, I will add a `Number.tofixed`

gmaillet (approved):

PR #1733, opened 2021-10-27
Fix label renderer

gchoqueux:
## Description
Some labels can appear in sky.
Effectively, labels are displayed even though they aren't in frustum camera.

## Motivation and Context
* remove labels glitch 
* improve performance because it reduce the labels count to be processed.

## Screenshot
chrome-capture

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -95,6 +95,7 @@ function _preprocessLayer(view, layer, parentLayer) {
             zoom: layer.zoom,
             crs: source.crs,
             visible: layer.visible,
+            margin: 15,
         });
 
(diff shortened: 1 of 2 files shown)


jailln:
Nice! Do you know how much does it improves performance ? Does it only have an impact when looking at the horizon ?

gchoqueux:
> Does it only have an impact when looking at the horizon ?

yes, labels that aren't projected on the screen, are also removed from the process.

> Nice! Do you know how much does it improves performance ?

I didn't do a benchmark, but severals labels are removed from process, so the grid computing and style calculations are reduced.

gmaillet (approved):

PR #1734, opened 2021-10-28
Add new distance and extent dimensions methods.

gchoqueux:
## Description
feature(Coordinates): add methods to calculate distance between coordinates.
* `planarDistanceTo`
* `geodesicDistanceTo`
* `earthEuclideanDistanceTo`

feature(Extent): add methods to calculate extent dimensions.

* `planarDimensions`
* `geodesicDimensions`
* `earthEuclideanDimensions`

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -8,6 +8,12 @@ proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
 const ellipsoid = new Ellipsoid();
 const projectionCache = {};
 
+const v0 = new THREE.Vector3();
+const v1 = new THREE.Vector3();
+
+let coord0;
+let coord1;
+
 function proj4cache(crsIn, crsOut) {
     if (!projectionCache[crsIn]) {
         projectionCache[crsIn] = {};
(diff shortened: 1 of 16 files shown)


src/Core/Geographic/Coordinates.js
```diff
@@ -8,6 +8,12 @@ proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
 const ellipsoid = new Ellipsoid();
 const projectionCache = {};
 
+const v0 = new THREE.Vector3();
+const v1 = new THREE.Vector3();
+
+let coord0;
```
mgermerie, line 14: Is there a point in not instantiating `Coordinates` here (and instantiate it at line 302 as you do) ?

src/Core/Geographic/Extent.js
```diff
@@ -256,11 +258,67 @@ class Extent {
...
+     * @return     {THREE.Vector2}  Planar dimensions
+     */
+    planarDimensions(target = new THREE.Vector2()) {
+        // Calculte the dimensions for x and y
+        return target.set(Math.abs(this.east - this.west), Math.abs(this.north - this.south));
+    }
+
+    /**
+     *  Geodesic dimensions are two planar distances west/east and south/north.
+     *  Geodesic distance is calculated in a ellispoid space as the distance
```
mgermerie: Could you replace 
> calculated in a ellipsoid

by : 
> calculated in **an** ellipsoid

src/Core/Geographic/Coordinates.js
mgermerie: Indeed, you could not instantiate `Coordinates` before it is defined, so you have to do it the way you do it.

mgermerie (approved):

PR #1736, opened 2021-11-02
fix(VectorTileSource): set style parent with style Layer

gchoqueux:
## Description
fix(VectorTileSource): set style parent with style Layer.

Possibility to define a non existing style for the layer `VectorTileLayer`.

### DIFF
--- a/src/Source/VectorTilesSource.js
+++ b/src/Source/VectorTilesSource.js
@@ -126,6 +126,13 @@ class VectorTilesSource extends TMSSource {
             }
         });
     }
+
+    onLayerAdded(options) {
+        super.onLayerAdded(options);
+        const keys = Object.keys(this.styles);
+
+        keys.forEach((k) => { this.styles[k].parent = options.out.style; });
+    }
 }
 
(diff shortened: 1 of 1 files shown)


src/Source/VectorTilesSource.js
```diff
@@ -126,6 +126,13 @@ class VectorTilesSource extends TMSSource {
             }
         });
     }
+
+    onLayerAdded(options) {
+        super.onLayerAdded(options);
+        const keys = Object.keys(this.styles);
+
+        keys.forEach((k) => { this.styles[k].parent = options.out.style; });
+    }
```
gmaillet, line 135: What happens if a source is used by several layers?

gchoqueux: Effectively, there's a issue. I'll look into it. But there's a problem with parent style property.

gchoqueux: I created the issue, the Style architecture needs refactoring.

gmaillet (approved):

PR #1737, opened 2021-11-02
Fix scale and altitude features

gchoqueux:
## Description
fix(Feature): wrong altitude and altitude limits.
fix(Feature2Mesh): set scale transformation from `FeatureCollection`.

### DIFF
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -484,6 +484,7 @@ export default {
 
             group.quaternion.copy(collection.quaternion);
             group.position.copy(collection.position);
+            group.scale.copy(collection.scale);
 
             return group;
         };
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -15,6 +15,19 @@ function _extendBuffer(feature, size) {
     }
 }
 
+function _setGeometryValues(geom, feature, long, lat, alt, normal) {
+    if (feature.normals) {
+        normal.toArray(feature.normals, feature._pos);
+    }
+
+    feature._pushValues(long, lat, alt);
+
+    if (geom.size == 3) {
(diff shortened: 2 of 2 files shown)


src/Converter/Feature2Mesh.js
```diff
@@ -484,6 +484,8 @@ export default {
 
             group.quaternion.copy(collection.quaternion);
             group.position.copy(collection.position);
+            group.scale.copy(collection.scale);
+            group.feature = collection;
```
gmaillet: why the collection is referenced in group.feature?

src/Core/Feature.js
```diff
@@ -157,6 +157,11 @@ export class FeatureGeometry {
      * @param {THREE.Vector3} [normal=THREE.Vector3(0,0,1)] the normal on coordinates.
      */
     pushCoordinatesValues(feature, long, lat, alt = 0, normal = defaultNormal) {
+        if (this.size == 3) {
+            // set altitude from context
+            const base_altitude = feature.style[typeToStyleProperty[feature.type]].base_altitude;
+            alt = isNaN(base_altitude) ? base_altitude(this.properties) : base_altitude;
+        }
```
gmaillet: is it possible to factorize this with pushCoordinates?

src/Converter/Feature2Mesh.js
gchoqueux: removed

src/Core/Feature.js
gchoqueux: refactorized

ign-packo (approved):
test with a FeatureGeometryLayer based on VectorTilesSource: scale and altitude.min/max are actually updated

gmaillet (approved):
test with a FeatureGeometryLayer based on VectorTilesSource: scale and altitude.min/max are actually updated

PR #1742, opened 2021-11-05
fix(PointCloud): fix memory crash when loading point cloud.

gchoqueux:
## Description
Fix memory OOM crash with `LASLoader`.

Fix for the multi-files LAS but i don't test with large file

see #1721


jailln (approved):

Desplandis:
Since our LAS parser doesn't use `loaders.gl` anymore, this PR is no longer relevant.

PR #1746, opened 2021-11-09
Fix picking point position in potree example

gchoqueux:
fix(examples): return picking point position instead of bounding box position.

### DIFF
--- a/examples/potree_25d_map.html
+++ b/examples/potree_25d_map.html
@@ -20,6 +20,23 @@
                     display: none;
                 }
             }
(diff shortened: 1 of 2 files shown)


gmaillet (commented):
It's definitely better, but:

- for each double click there are 10 or 20 points displayed in the console: those are maybe very close points but is there a way to choose the closest?
- This console-based functionality is hidden, shouldn't the selected point be displayed with a sphere and the coordinates displayed in the view?

PR #1752, opened 2021-11-17
refacto(Feature): defaults buildExtent parameter to true for 2d structure

mgermerie:
## Description

Defaults `buildExtent` parameters for `FeatureCollection` to true if `structure` parameter is set to `2d`.

## Motivation and Context

Simplify `FeatureCollection` implementation process by avoiding having to specify parameter that is often true with 2d features.


src/Source/FileSource.js
```diff
@@ -80,12 +80,7 @@ import CRS from 'Core/Geographic/Crs';
  *     .then(function _(geojson) {
  *         return itowns.GeoJsonParser.parse(geojson, {
  *             in: { in: 'EPSG:4326' },
```
gchoqueux: could you replace `in` by crs?

mgermerie: done

gchoqueux (approved):

PR #1769, opened 2021-12-13
feature(ElevationLayer): add visible property.

gchoqueux:
## Description
Add visible property to `ElevationLayer`.
If visible is false the elevation terrain is removed.
The bbox tile is updated too. 

## Screenshots
<details><summary>Capture</summary>
<p>

chrome-capture

</p>
</details>


mgermerie (approved):

mgermerie:
It is necessary to re-position the camera when switching `visible` property, both in `GlobeControls` and `PlanarControls`. This should prevent having the camera under the ground in some cases, when switching `visible` to `true`.

mgermerie:
Maybe you could call the `Controls` `update` when visible property is changed, not only in the GUI tools ?

PR #1798, opened 2022-02-25
fix(parser): GeoJsonParser add legacy identifier to fct readCRS()

ftoromanoff:
## Description
The function readCRS() in GeoJsonParser is currently not able to read a crs named under the 'legacy identifier' format (i.e. EPSG:2154), which is used in various application. (as PostGIS)

## Motivation and Context
This fix allows the recognition of the crs used in a geojson if this one is defined under the 'legacy identifier' format.


gchoqueux (approved):

gchoqueux:
Thanks!

PR #1803, opened 2022-03-04
refactor(GeometryLayer): reference to material properties to Layer properties.

gchoqueux:
## Description
Add referencing to material properties to Layer properties.
Material properties (opacity, wire-frame) point directly to Layer properties.

## Motivation and Context
It avoids, when a layer property is modified, to browse each 3D object to modify their property.

### DIFF
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import Earcut from 'earcut';
 import { FEATURE_TYPES } from 'Core/Feature';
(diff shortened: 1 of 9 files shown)


src/Converter/Feature2Mesh.js
```diff
@@ -470,6 +480,10 @@ export default {
      */
     convert(options = {}) {
```
mgermerie: Maybe you could add doc to specify that user can specify `layer` as an option ?

gchoqueux: Finally, I remove this parameter

mgermerie (approved):

PR #1807, opened 2022-03-09
chore: add support ecma 2022

gchoqueux:
## Description
add support ecma 2016 to 2022
add nullish coalescing operator support.

Ecma es6 to 2022 used on a part of the code 

## Motivation and Context
* private field class 
* new operator to simplify 
      `a !== undefined ? a : false;` to `a ?? false`

### DIFF
--- a/test/functional/misc_clamp_ground.js
+++ b/test/functional/misc_clamp_ground.js
@@ -30,7 +30,11 @@ describe('misc_clamp_ground', function _() {
             // compute the on screen cone position
             const coneCenter = new itowns.THREE.Vector3(0, 0, 0)
                 .applyMatrix4(view.mesh.matrixWorld);
-            coneCenter.applyMatrix4(view.camera._viewMatrix);
+
+            const camera3D = view.camera.camera3D;
+            const viewMatrix = new itowns.THREE.Matrix4().multiplyMatrices(camera3D.projectionMatrix, camera3D.matrixWorldInverse);
+
+            coneCenter.applyMatrix4(viewMatrix);
             const mouse = view.normalizedToViewCoords(coneCenter);
 
             // So read the depth buffer at cone's position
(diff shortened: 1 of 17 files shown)


src/Renderer/Label2DRenderer.js
```diff
@@ -8,7 +8,7 @@ function isIntersectedOrOverlaped(a, b) {
 // find label in children
 function hasLabelChildren(object) {
     const parent = object.parent;
-    return parent.material && !parent.material.visible &&
+    return !(parent.material?.visible) &&
```
mgermerie: This line will not have the same value as before if `parent.material` is `undefined` : previously, it would have returned `undefined` and it now returns `true`.

test/functional/misc_clamp_ground.js
```diff
@@ -30,7 +30,11 @@ describe('misc_clamp_ground', function _() {
             // compute the on screen cone position
             const coneCenter = new itowns.THREE.Vector3(0, 0, 0)
                 .applyMatrix4(view.mesh.matrixWorld);
-            coneCenter.applyMatrix4(view.camera._viewMatrix);
+
+            const camera3D = view.camera.camera3D;
+            const viewMatrix = new itowns.THREE.Matrix4().multiplyMatrices(camera3D.projectionMatrix, camera3D.matrixWorldInverse);
+
+            coneCenter.applyMatrix4(viewMatrix);
```
mgermerie, line 37: Why not doing a getter for `Camera.#_viewMatrix` property ?

gchoqueux: it's not necessary for one use case in functional test

src/Renderer/Label2DRenderer.js
gchoqueux: done

mgermerie (approved):

PR #1816, opened 2022-03-15
examples: add widgets and change appearance

mgermerie:
## Description

Add possibility to setup placeholder in search bar widget.
Add navigation and search bar widgets on view_3d_map example.
Change the style of `view source` button.


gchoqueux (approved):
Thanks!

PR #1822, opened 2022-03-25
Introduce `FeatureMesh`

gchoqueux:
## Description
Introduce `FeatureMesh` to handle the feature converted to mesh.
`FeatureMesh` could be it can dynamically change its geographic projection system.
This method avoids to use `proj4`, which is more accurate less efficient.
It's still possible to use `proj4` with an `FeatureGeometryLayer` option.

 `FeatureMesh`  is used with `FeatureGeometryLayer`. 
Now, the 3d objects of `FeatureGeometryLayer` are added to `FeatureGeometryLayer.object3d`.

## Motivation and Context
Avoid `proj4` increases performance.
The `feature processing` is more simple and efficient.
`FeatureMesh` allows to display the tiled vector in 3d without any geographical transformation (directely from Protobufs raw data )

## Screenshots
                                  
image


examples/vector_tile_3d_mesh.html
```diff
@@ -0,0 +1,104 @@
...
+                     && !layer['source-layer'].includes('parcellaire')
+                }
+            });
+
+            var mvtLayer = new itowns.ColorLayer('MVT', {
+                source: mvtSourceColor,
+                addLabelLayer: true,
+            });
+
+            var mvtBuildingLayer = new itowns.FeatureGeometryLayer('WFS Building',{
```
jailln: MVT Buildings instead of WFS buildings ?

src/Layer/FeatureGeometryLayer.js
```diff
@@ -22,22 +22,37 @@ class FeatureGeometryLayer extends GeometryLayer {
...
+    constructor(id, options = {}) {
+        options.update = FeatureProcessing.update;
+        options.convert = Feature2Mesh.convert({
+            batchId: options.batchId,
         },
         );
-        super(id, config.object3d || new Group(), config);
+        super(id, options.object3d || new Group(), options);
         this.isFeatureGeometryLayer = true;
+        this.useProj4 = options.useProj4 ?? false;
```
jailln: Shouldn't it be true by default ? So it is by default more precise geographically but you can use an option to increase performances ?

jailln:
Nice feature! Could you quantify how faster it is to not use proj4 on an example dataset ? And to estimate what is the geometric loss ?

examples/vector_tile_3d_mesh.html
gchoqueux: yes

gchoqueux:
> Nice feature! Could you quantify how faster it is to not use proj4 on an example dataset ? And to estimate what is the geometric loss ?

the `parse` is twice as fast. In wfs example the error is inferior to 0.01 mm.

gchoqueux:
I will add a test to calculate and verify the geometric loss

jailln:
> > Nice feature! Could you quantify how faster it is to not use proj4 on an example dataset ? And to estimate what is the geometric loss ?
> 
> the `parse` is twice as fast. In wfs example the error is inferior to 0.01 mm.

Cool! Could you add it in the doc please ? So one knows what it implies to use one option or the other.

src/Process/FeatureProcessing.js
```diff
@@ -20,15 +19,12 @@ export default {
         if (node.layerUpdateState[layer.id] === undefined) {
             node.layerUpdateState[layer.id] = new LayerUpdateState();
         } else if (!node.layerUpdateState[layer.id].canTryUpdate()) {
+            // toggle visibility features
+            const features = node.link.filter(n => n.layer && (n.layer.id == layer.id));
```
mgermerie: Would it be possible to replace callback with `n => n.layer?.id === layer.id ` ?

src/Parser/VectorTileParser.js
```diff
@@ -62,17 +62,15 @@ function vtFeatureToFeatureGeometry(vtFeature, feature, classify = false) {
             if (count == 1) {
                 firstPoint.set(x, y);
                 lastPoint.set(x, y);
-            } else if (classify && count > 1) {
+            } else if (count > 1) {
```
mgermerie: Maybe it could be in a separate PR/commit ?

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
-                    layer.onMeshCreated(result);
-                }
-                node.layerUpdateState[layer.id].success();
-                if (!node.parent) {
-                    ObjectRemovalHelper.removeChildrenAndCleanupRecursively(layer, result);
-                    return;
+        return context.scheduler.execute(command).then((meshes) => {
+            node.layerUpdateState[layer.id].noMoreUpdatePossible();
+
+            meshes.forEach((featureNode) => {
```
mgermerie: Could you rename `featureNode` to `featureMesh` ?

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
+                if (featureNode) {
```
mgermerie: Is it necessary to do this check ? I mean : can the commend return an array containing an `undefined` value if the matching extent has no associated data ?

src/Core/Feature.js
```diff
@@ -145,7 +145,9 @@ export class FeatureGeometry {
 
         feature.transformToLocalSystem(coordOut);
 
-        _setGeometryValues(this, feature, coordOut.x, coordOut.y, coordOut.z, coordOut.geodesicNormal);
+        const normal = coordIn.crs == feature.crs ? defaultNormal : coordOut.geodesicNormal;
+
+        _setGeometryValues(this, feature, coordOut.x, coordOut.y, coordOut.z, normal);
```
mgermerie: The `normal` array could be simplified if all normal are worth `defaultNormal`.

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
+                    featureNode.as(context.view.referenceCrs);
+
+                    if (layer.onMeshCreated) {
```
mgermerie, line 71: On another subject than this PR, the `onMeshCreated` property has no documentation to be found anywhere.

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
+                        layer.onMeshCreated(featureNode, context);
```
mgermerie: The `context` parameter is not used in any examples or source code. It can be a useful addition, but would maybe need an example and docs.

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
 
-        return context.scheduler.execute(command).then((result) => {
-            // if request return empty json, WFSProvider.getFeatures return undefined
-            result = result[0];
-            if (result) {
-                // call onMeshCreated callback if needed
-                if (layer.onMeshCreated) {
-                    layer.onMeshCreated(result);
-                }
-                node.layerUpdateState[layer.id].success();
```
mgermerie, line 65: Don't we need to specify it anymore ?

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
-                node.remove(...node.children.filter(c => c.layer && c.layer.id == layer.id));
-                const group = new THREE.Group();
-                group.layer = layer;
-                group.matrixWorld.copy(node.matrixWorld).invert();
-                group.matrixWorld.decompose(group.position, group.quaternion, group.scale);
-                group.position.z += node.geoidHeight;
-                node.add(group.add(result));
-                group.updateMatrixWorld(true);
-            } else {
-                node.layerUpdateState[layer.id].failure(1, true);
```
mgermerie, line 80: Same question as for the `success` : we don't manage update state in case fetched data is empty.

src/Converter/Feature2Mesh.js
```diff
@@ -3,12 +3,72 @@ import Earcut from 'earcut';
...
+
+            // Position and orientation
+            // remove original position
+            this.place.position.copy(this.meshesCollection.position).negate();
+
+            // get mesh coordinate
+            coord.setFromVector3(this.meshesCollection.position);
+
+            // get method to calculate orientation
+            const crsWGS84 = 'EPSG:4326';
```
mgermerie: This could be defined outside the method.

src/Process/FeatureProcessing.js
```diff
@@ -54,31 +50,27 @@ export default {
...
+                    }
+                    featureNode.layer = layer;
                 }
-                // remove old group layer
-                node.remove(...node.children.filter(c => c.layer && c.layer.id == layer.id));
-                const group = new THREE.Group();
-                group.layer = layer;
-                group.matrixWorld.copy(node.matrixWorld).invert();
-                group.matrixWorld.decompose(group.position, group.quaternion, group.scale);
-                group.position.z += node.geoidHeight;
```
mgermerie, line 76: The geoid height is no longer managed. You could add 
```js
featureNode.place.position.z += node.geoidHeight;
featureNode.updateMatrixWorld();
```
just after `featureNode.as(context.view.referenceCrs);` on line 58.

There must be a cleaner way of doing this (to avoid calling `updateMatrixWorld` both in `FeatureMesh.as` method and just after).

src/Process/FeatureProcessing.js
gchoqueux: yes

src/Parser/VectorTileParser.js
gchoqueux: yes

src/Process/FeatureProcessing.js
gchoqueux: yes

src/Process/FeatureProcessing.js
gchoqueux: indeed, there seems to be no case where the `featureMesh` is undefined

src/Core/Feature.js
gchoqueux: I add a new PR to resolve this issue

src/Process/FeatureProcessing.js
gchoqueux: added

src/Process/FeatureProcessing.js
gchoqueux: in another PR

src/Process/FeatureProcessing.js
gchoqueux: is replaced by `noMoreUpdatePossible()`

src/Process/FeatureProcessing.js
gchoqueux: but it's necessary to add PR to solve this issue

src/Converter/Feature2Mesh.js
```diff
@@ -29,14 +30,16 @@ class FeatureMesh extends THREE.Group {
         this.meshesCollection.updateMatrix();
 
         this.crs = collection.crs;
+        this.#currentCrs = this.crs;
```
mgermerie: In my opinion, it would make more sense to rename `this.crs` to something like `this.collectionCrs` and `this.#currentCrs` to `this.#crs`. WDYT ?

Also, I don't see the the current `this.crs` property being used outside of `FeatureMesh` class. Could it be set private ?

examples/vector_tile_3d_mesh.html
```diff
@@ -0,0 +1,157 @@
...
+        <!-- Import iTowns source code -->
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        <!-- Import iTowns LoadingScreen and GuiTools plugins -->
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+
+        <script type="text/javascript">
+
+             var promises = [];
```
mgermerie: I forgot this unused line. Could you remove it ?

mgermerie:
There is another issue with geoid height : 
If i display a `GeoidLayer` along with a `GeometryLayer`, toggling the `GeoidLayer` visibility won't translate `GeometryLayer` meshes, as can be seen in both screenshots bellow.

Geoid visibility ON :
geoid-on

Geoid visibility OFF : 
geoid-off

This behavior does not appears on master branch.

examples/vector_tile_3d_mesh.html
gchoqueux: solved

gchoqueux:
The geoid issue is resolved
But there is an geoid issue on master, i will fix it after the PR merging

src/Converter/Feature2Mesh.js
gchoqueux: I simplified the code

src/Process/FeatureProcessing.js
```diff
@@ -20,15 +19,12 @@ export default {
         if (node.layerUpdateState[layer.id] === undefined) {
             node.layerUpdateState[layer.id] = new LayerUpdateState();
         } else if (!node.layerUpdateState[layer.id].canTryUpdate()) {
+            // toggle visibility features
+            const features = node.link.filter(n => n.layer?.id == layer.id);
```
mgermerie: It is a detail, but it might be more relevant to rename `n` parameter to `feature` or `f`, since `node.link` contains `FeatureMeshes`.

src/Layer/FeatureGeometryLayer.js
```diff
@@ -22,22 +22,45 @@ class FeatureGeometryLayer extends GeometryLayer {
...
+     * @param {Object} [options] - Optional configuration, all elements in it
      * will be merged as is in the layer.
      * @param {function} [options.batchId] - optional function to create batchId attribute.
      * It is passed the feature property and the feature index.
      * As the batchId is using an unsigned int structure on 32 bits, the batchId could be between 0 and 4,294,967,295.
-     * @param {THREE.Object3D} [config.object3d=new THREE.Group()] root object3d layer.
+     * @param {THREE.Object3D} [options.object3d=new THREE.Group()] root object3d layer.
+     * @param {function} [options.onMeshCreated] this callback is called when the mesh is created. The callback parameters are the
+     * `mesh` and the `context`.
+     * @param {boolean} [options.accurate=TRUE] If `accurate` is `true` the data are reprojected with maximum geographical accurate.
```
mgermerie: Could you replace by :
```
* @param {boolean} [options.accurate=TRUE] If `accurate` is `true`, data are re-projected with maximum geographical accuracy.
* With `true`, `proj4` is used to transform data source.
*
* If `accurate` is `false`, re-projecting is faster but less accurate.
* With `false`, an affine transformation is used to transform data source.
* This method is an approximation. The error increases with the extent
* dimension of the object or queries.

* For example : 
...
```

mgermerie (approved):

PR #1828, opened 2022-03-30
refactor(Feature): simplify normals data.

gchoqueux:
## Description
Simplify using `Feature.normals` data. Finally, they used only with the geocentric projection.

### DIFF
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -69,6 +69,7 @@ export class FeatureBuildingOptions {}
  * anything specified in the GeoJSON under the `properties` property.
  */
 export class FeatureGeometry {
+    #currentExtent;
     /**
      * @param {Feature} feature geometry
      */
@@ -78,7 +79,7 @@ export class FeatureGeometry {
         this.size = feature.size;
         if (feature.extent) {
             this.extent = defaultExtent(feature.extent.crs);
-            this._currentExtent = defaultExtent(feature.extent.crs);
+            this.#currentExtent = defaultExtent(feature.extent.crs);
         }
         this.altitude = {
             min: Infinity,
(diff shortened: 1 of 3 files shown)


jailln:
Why only using them for the geocentric position ? They are not used in Planar View ? If yes, why ?

gchoqueux:
> Why only using them for the geocentric position ? 

The normal is used to change altitude vertex or extrude polygon.
For each geocentric position the normal is different, this is normal to ellipsoid.

> They are not used in Planar View ? If yes, why ?

For planar view the normal is always the same `(0, 0, 1)`

jailln:
Oh ok it makes sense, thanks

src/Core/Feature.js
```diff
@@ -163,10 +162,10 @@ export class FeatureGeometry {
      * @param {number} lat The latitude coordinate.
      * @param {THREE.Vector3} [normal=THREE.Vector3(0,0,1)] the normal on coordinates.
      */
-    pushCoordinatesValues(feature, long, lat, normal = defaultNormal) {
+    pushCoordinatesValues(feature, long, lat) {
```
mgermerie: Could you remove `normal` parameter from the documentation, and also change `'oesn't'` on line 157 by `'is not'` ?

src/Core/Feature.js
```diff
@@ -163,10 +162,10 @@ export class FeatureGeometry {
...
         const altitude = this.baseAltitude(feature);
 
-        _setGeometryValues(this, feature, long, lat, altitude, normal);
+        _setGeometryValues(this, feature, long, lat, altitude);
```
mgermerie: Shouldn't we check if `feature.crs` is different than `EPSG:4978` to prevent error if user calls this method with an `EPSG:4978` feature ?

src/Core/Feature.js

src/Core/Feature.js
gchoqueux: I finally revert this part, to support the `EPSG:4978` projection.

mgermerie (approved):

Issue #1830, opened 2022-04-03
 Problem adding 3dtile


GGBond-GIS:
![Q`C4AFC1(3RV3@%_K6KIY`3](https://user-images.githubusercontent.com/92559206/161434768-7d7e8922-3677-4402-8bc7-66061b6a0eb8.png)

jailln:
Hi,
Could you try with [this branch](https://github.com/iTowns/itowns/tree/feat/binary-batch-table) please and let me know if it works ? It implements the support for binary batch table which seems to be the problem with your dataset.

One way to test it is to replace [this URL](https://github.com/iTowns/itowns/blob/feat/binary-batch-table/examples/3dtiles_basic.html#L64) by your tileset URL.

jailln:
I suppose it worked since you opened #1835 ? Can you confirm please so we can close this issue if this is the case ?
Thanks.

Issue #919, opened 2018-11-22
Error loading b3dm build by FME 


gchoqueux:
Error loading/parsing b3dm build by FME

## Possible Cause/Fix/Solution
Error when BatchTable is parsing because start index buffer and length buffer are wrong.

mgermerie:
The complete 3d-tiles support is planned in the short term road map. We are currently studying possibilities in terms of libraries to enhance support (as mentioned in https://github.com/iTowns/itowns/discussions/1917).

jailln:
Done in #1834

PR #1834, opened 2022-04-05
feat(3dtiles): add support for binary batch table

### LINKED ISSUE #1826: How to load such tiles? No error is reported but tilt photography is not displayed. Can you see where the error is
image

I can't see my tilt photography

jailln:
## Description

* Add support for [binary batch table attributes](https://github.com/CesiumGS/3d-tiles/blob/main/specification/TileFormats/BatchTable/README.md#binary-body)
* Add tests for 3d tiles binary property parsing

## Motivation and Context

* Improve 3D Tiles support
* Fixes #1826 , #1830 and #919 

*Note: `BinaryPropertyAccessor.js` may also be used for parsing the binary content of the 3D Tiles feature table (not supported by itowns yet).*

### DIFF
--- a/src/Core/3DTiles/C3DTBatchTable.js
+++ b/src/Core/3DTiles/C3DTBatchTable.js
@@ -1,4 +1,5 @@
 import utf8Decoder from 'Utils/Utf8Decoder';
+import binaryPropertyAccessor from './utils/BinaryPropertyAccessor';
 import C3DTilesTypes from './C3DTilesTypes';
 
 /** @classdesc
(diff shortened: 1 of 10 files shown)


gchoqueux:
Hello Vincent,
Thanks for your contribution.
Your PR is ready for review?

jailln:
Yes it is :slightly_smiling_face: 
I will just add more tests to pass the coverage check but it won't change the code already pushed.

jailln:
Hello @gchoqueux 
It's ready to be reviewed.

gchoqueux:
I am currently busy 
I'll look at it as soon as I have some time

src/Core/3DTiles/C3DTBatchTable.js
```diff
@@ -17,41 +18,57 @@ import C3DTilesTypes from './C3DTilesTypes';
...
+            const binaryBuffer = buffer.slice(jsonLength, jsonLength + binaryLength);
 
-        // Parse JSON content
-        const content = utf8Decoder.decode(new Uint8Array(jsonBuffer));
-        const json = JSON.parse(content);
+            for (const propKey in jsonContent) {
+                if (!Object.prototype.hasOwnProperty.call(jsonContent, propKey)) {
+                    continue;
+                }
+                const propVal = jsonContent[propKey];
```
mgermerie, line 52: Do you know if there could be a case where `propVal` would be `undefined` - jsonContent would be like `{ foo: undefined }` ? If so, lines 46 to 48 could cause an error : reading `byteOffset` from `undefined`, and so on with the two other properties.
If this case is possible, it could be solved replacing lines 46 to 48 with the following :
```js
if (typeof propVal?.byteOffset !== 'undefined' &&
    typeof propVal?.componentType !== 'undefined &&
    typeof propVal?.type !== 'undefined') {
```

src/Core/3DTiles/C3DTBatchTable.js
```diff
@@ -17,41 +18,57 @@ import C3DTilesTypes from './C3DTilesTypes';
...
             this.extensions =
-                registeredExtensions.parseExtensions(json.extensions, this.type);
-            delete json.extensions;
+                registeredExtensions.parseExtensions(jsonContent.extensions, this.type);
+            delete jsonContent.extensions;
         }
 
         // Store batch table json content
-        this.content = json;
+        // this.content = json;
```
mgermerie: Could you remove this commented line ?

src/Core/3DTiles/C3DTBatchTable.js
```diff
@@ -17,41 +18,57 @@ import C3DTilesTypes from './C3DTilesTypes';
...
+     * @param {ArrayBuffer} buffer - batch table buffer to parse
+     * @param {number} jsonLength - batch table json part length
+    * @param {number} binaryLength - batch table binary part length
      * @param {number} batchLength - the length of the batch.
      * @param {Object} registeredExtensions - extensions registered to the layer
      */
-    constructor(buffer, binaryLength, batchLength, registeredExtensions) {
+    constructor(buffer, jsonLength, binaryLength, batchLength, registeredExtensions) {
         this.type = C3DTilesTypes.batchtable;
         this.batchLength = batchLength;
```
mgermerie: The constructor signature change is a breaking change. I think we could add a depreciation warning like so : 
```js
if (jsonLength + binaryLength !== buffer.byteLength) {
    // The constructor was called with its deprecated signature
    console.warn('some deprecation error message');
    registeredExtensions = batchLength;
    batchLength = binaryLength;
    binaryLength = jsonLength;
    jsonLength = buffer.byteLength - binaryLength;
}
```
WDYT ?

src/Core/3DTiles/utils/BinaryPropertyAccessor.js
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Parses a 3D Tiles binary property. Used for batch table and feature table parsing. See the 3D Tiles spec for more
+ * information on how these values are encoded:
+ * [3D Tiles spec](https://github.com/CesiumGS/3d-tiles/blob/main/specification/TileFormats/BatchTable/README.md#binary-body))
+ * @param {ArrayBuffer} buffer The buffer to parse values from.
+ * @param {Number} batchLength number of objects in the batch (= number of elements to parse).
+ * @param {Number} byteOffset the offset in bytes into the buffer.
+ * @param {String} componentType the type of component to parse (one of componentTypeBytesSize keys)
+ * @param {String} type the type of element to parse (one of typeComponentsNumber keys)
+ * @returns {Array} an array of values parsed from the buffer. An array of componentType if type if SCALAR. An array
```
mgermerie: There is a small typo : `if type is SCALAR` instead of `if type if SCALAR`

src/Core/3DTiles/C3DTBatchTable.js
jailln: I think it's a good idea to add this test to check if the provided arguments seem correct but not to check if the constructor was called with the former signature since the test can pass if the constructor is called with the former signature and if the batch length equals the binary length.

However, since there is no optional argument in this function, we can test if the number of arguments is correc and add a deprecation warning and compute the values as suggested if not. WDYT?

src/Core/3DTiles/C3DTBatchTable.js
jailln: I don't know if this can happen but better be careful :) I will add it.

src/Core/3DTiles/C3DTBatchTable.js
jailln: yep done

src/Core/3DTiles/utils/BinaryPropertyAccessor.js
jailln: well spotted! done

src/Core/3DTiles/C3DTBatchTable.js
mgermerie: Indeed, I didn't think of simply counting the number of arguments but it is a safer test :+1:

jailln: After some more thinking, I think we should not add a deprecation test based on arguments length either since it may happen for other reasons than because the previous constructor was used. In addition, batch tables are not meant to be constructed directly by users of itowns but by our 3D tiles parsers so there is little chance that a user calls this constructor anyway. Therefore, I think we should keep it this way and only add BREAKING CHANGE in the commit title (and therefore in the release). WDYT ?

jailln:
I added the deprecation warning, added the breaking change message in the commit and rebased.

mgermerie (approved):

PR #1857, opened 2022-05-11
fix(Minimap): remove focus on minimap, and focus back on view

gliegard:


src/Utils/gui/Minimap.js
```diff
@@ -119,12 +119,10 @@ class Minimap extends Widget {
             placement: layer.source.extent,  // TODO : the default placement should be the view extent for ortho camera
             noControls: true,
             maxSubdivisionLevel: view.tileLayer.maxSubdivisionLevel,
+            disableFocusOnStart: true,
         });
         this.view.addLayer(layer);  // TODO : should this promise be returned by constructor so that user can use it ?
 
-        // Give the focus back to the main view. Indeed, `View` constructor takes the focus, and we don't want the focus
-        // on the latest created `View`, which is the minimap view.
-        view.domElement.focus();
```
gchoqueux, line 127: Could you add this check

```js
if(!view.disableFocusOnStart) {
view.domElement.focus();
}
```

gchoqueux (changes requested):

src/Utils/gui/Minimap.js
gliegard: the minimap don't take the focus anymore, so we dont need to give the focus back to the view

gchoqueux (approved):

PR #1867, opened 2022-06-21
Improve 3D Tiles support

jailln:
## Description

* Some improvements of 3D Tiles support, especially for old tilesets generated in early versions of the spec and with old cesium tools: better support of gltf 1.0 binary extension and avoid a bug when tilesets contains cesium specific uniforms and/or functions in gltf shaders.
* Correction of `overrideMaterials` for 3D Tiles layers: it was true by default and it did not work: material where not disposed since they were still referenced by the mesh and disposal of textures were not handled.

### DIFF
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -31,6 +31,16 @@ class C3DTilesLayer extends GeometryLayer {
      * @param {C3TilesSource} config.source The source of 3d Tiles.
      *
      * name.
+     * @param {Number} [config.sseThreshold=16] The [Screen Space Error](https://github.com/CesiumGS/3d-tiles/blob/main/specification/README.md#geometric-error)
+     * threshold at which child nodes of the current node will be loaded and added to the scene.
+     * @param {Number} [config.cleanupDelay=1000] The time (in ms) after which a tile content (and its children) are
+     * removed from the scene.
+     * @param {Function} [config.onTileContentLoaded] Callback executed when the content of a tile is loaded.
+     * @param {Boolean|Material} [config.overrideMaterials='false'] option to override the materials of all the layer's
+     * objects. If true, a threejs [MeshBasicMaterial](https://threejs.org/docs/index.html?q=meshbasic#api/en/materials/MeshBasicMaterial)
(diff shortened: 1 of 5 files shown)


src/Layer/C3DTilesLayer.js
```diff
@@ -31,6 +31,16 @@ class C3DTilesLayer extends GeometryLayer {
      * @param {C3TilesSource} config.source The source of 3d Tiles.
      *
      * name.
+     * @param {Number} [config.sseThreshold=16] The [Screen Space Error](https://github.com/CesiumGS/3d-tiles/blob/main/specification/README.md#geometric-error)
+     * threshold at which child nodes of the current node will be loaded and added to the scene. Deaul
+     * @param {Number} [config.cleanupDelay=1000] The time after which a tile content (and its children) are removed
+     * from the scene.
```
mgermerie: It's a detail but maybe we could specify the `ms` unit ?
Also, does the 'Deaul' on line 35 mean something or is it a typo ? :)

src/Layer/C3DTilesLayer.js
```diff
@@ -41,7 +51,7 @@ class C3DTilesLayer extends GeometryLayer {
         this.onTileContentLoaded = config.onTileContentLoaded || (() => {});
         this.protocol = '3d-tiles';
         // custom cesium shaders are not functional;
-        this.overrideMaterials = config.overrideMaterials !== undefined ? config.overrideMaterials : true;
+        this.overrideMaterials = config.overrideMaterials !== undefined ? config.overrideMaterials : false;
```
mgermerie: Small detail here again, but the line could be simplified with 
```js
this.overrideMaterials = config.overrideMaterials ?? false;
```
By the way the `??` operator also checks for `null` values.

src/Layer/C3DTilesLayer.js
jailln: Yep I added the unit in the comment and removed the Deaul :)

src/Layer/C3DTilesLayer.js
jailln: Yep you're right it's better, done.

mgermerie (approved):

PR #1870, opened 2022-07-08
WebXR support

mbredif:
## Description
This proof of concept show some initial support for [WebXR](https://immersiveweb.dev/). This adds 3 examples which feature the Three's VRButton for entering the immersive mode.

## Motivation and Context
Three.js supports WebXR, which is the web standard for virtual/mixed/augmented reality. Having iTowns support WebXR would enable an immersive iTowns experience on any WebXR compatible device (hololens, vive, android...)

## Screenshots
image
image

## Tested Platforms
- [x] Hololens 2 on Edge
- [x] [Web XR API Emulator](https://github.com/MozillaReality/WebXR-emulator-extension) on firefox 102

## Known Issues (to be completed...)
- Enabling WebXR is hacky
- Camera poses between immersive and non immersive mode are not synchronized
- Camera orientation in immersive mode does not align the vertical direction of the virtual world with the one of the real world.
- LOD computation is broken (low res raster tiles and point patches, no 3D features)

### DIFF
--- a/examples/config.json
+++ b/examples/config.json
@@ -38,6 +38,13 @@
         "source_stream_wfs_raster": "WFS to raster"
     },
 
+
+    "WebXR (experimental)": {
+        "webxr_3dtiles_25d": "3DTiles on 2.5D map (WebXR)",
+        "webxr_potree_25d_map": "Potree on 2.5D map (WebXR)",
+        "webxr_source_stream_wfs_25d": "WFS to 2.5D objects (WebXR)"
+    },
+    
     "Specific source options": {
         "source_file_from_format": "format",
         "source_file_from_methods": "fetcher and parser"
--- a/examples/js/VRButton.js
+++ b/examples/js/VRButton.js
@@ -1,0 +1,203 @@
+class VRButton {
+
+	static createButton( renderer, options ) {
+
+		if ( options ) {
+
+			console.error( 'THREE.VRButton: The "options" parameter has been removed. Please set the reference space type via renderer.xr.setReferenceSpaceType() instead.' );
+
+		}
+
+		const button = document.createElement( 'button' );
+
+		function showEnterVR( /*device*/ ) {
+
+			let currentSession = null;
+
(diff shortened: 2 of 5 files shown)


jailln:
I've also tested VR (and AR) with itowns [on this branch](https://github.com/jailln/itowns/tree/poc/xr)

I found the same conclusions and difficulties:
-  the rendering loop needs to be modified
- difficulties to sync the VR camera with the 3D viewer camera

Regarding the rendering loop modifications, there is at least a need to replace `requestAnimationFrame` by `setAnimationLoop`, but I haven't taken enough time to check what are the implications of this modifications and how we should manage it in itowns.

Regarding camera synchronization, I think that [setReferenceSpaceType](https://threejs.org/docs/?q=xrrefe#api/en/renderers/webxr/WebXRManager.setReferenceSpaceType) might be the way to go but it didn't work for me and I saw afterwards that according to [mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/API/XRReferenceSpace) it only works in https.

mittermichal:
hi,
in my project (not public atm) that uses https://github.com/NASA-AMMOS/3DTilesRendererJS I have tried to use XRReferenceSpace for teleporting in VR and ran into issue that when user position coordinates were too big, then controllers did not move smoothly in the scene. I suspect [this](https://github.com/immersive-web/webxr/issues/1236) is the issue causes that.

So I ended up moving center of tileset group to center of its bounding box and moving group containing camera and controllers when teleporting. Like they do in the NASA-AMMOS renderer: https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/9cead4f1dceb28138ee6b0cda4df7db427fc0384/example/vr.js#L304

> setReferenceSpaceType only works in https

secure context is available in localhost same as WebXR - https://developer.mozilla.org/en-US/docs/Web/API/WebXR_Device_API

gchoqueux:
new webXR support in #2092

PR #1875, opened 2022-07-21
Add a script to start an https local session with webpack

jailln:
## Description
Add an npm script to start a local server in https

## Motivation and Context
Can be useful to test some features in https (e.g. for VR, see #1870 ) and will be useful in a context where https is progressively becoming mandatory.


mgermerie (approved):

mgermerie:
Thanks for this addition !

Issue #1877, opened 2022-08-02
2.5D 3D tiles example is down


gchoqueux:
The 3D tiles isn't loaded in `2.5D 3D tiles example`.

because the `Liris server` [link](http://rict.liris.cnrs.fr/DataStore/Lyon1er-2015/tileset.json) responded with a status of 503 

@ericboix and @agrignard have you an idea?

jailln:
I don't know about the server but I can compute another tileset if this can help (similar to the previous one, but maybe on a larger zone and with textures). However where would we store it ? Is there a server for itowns demo data ?

vladimir-rybalko:
@jailln, share your experience. With what software do you prepare tilesets?

jailln:
I use [py3dtilers](https://github.com/VCityTeam/py3dtilers) which allows to transform citygml, obj, geojson and ifc files to 3d-tiles.

jailln:
BTW the dataset seems to also be stored [here](https://github.com/iTowns/iTowns2-sample-data/tree/master/3DTiles/dataset-dl.liris.cnrs.fr/three-d-tiles-lyon-metropolis/Lyon_2015_TileSet)

vladimir-rybalko:
> I use [py3dtilers](https://github.com/VCityTeam/py3dtilers) which allows to transform citygml, obj, geojson and ifc files to 3d-tiles.

@jailln, maybe you can suggest.
I'm trying to get 3D tiles to display from an obj file with a texture using py3dtilers tools. But as a result I get only one tile.
I created a [issue](https://github.com/VCityTeam/py3dtilers/issues/106) but i'm not sure if it's not on my side.

jailln:
> > I use [py3dtilers](https://github.com/VCityTeam/py3dtilers) which allows to transform citygml, obj, geojson and ifc files to 3d-tiles.
> 
> @jailln, maybe you can suggest. I'm trying to get 3D tiles to display from an obj file with a texture using py3dtilers tools. But as a result I get only one tile. I created a [issue](https://github.com/VCityTeam/py3dtilers/issues/106) but i'm not sure if it's not on my side.

[...]

PR #1878, opened 2022-08-03
Fix picking bug and don't pick atmosphere layer

jailln:
## Description
Two improvements on picking:
* Fix a bug that occurred when picking on multiple sources:
`View.pickObjectsAt()` calls `View.eventToViewCoords()` without a target result to get mouse view coordinates (`mouse`), so the result is stored in `_eventCoords`. `View.pickObjectsAt()` then calls `Picking.pickObjectsAt` on each sources which sometimes calls `View.viewToNormalizedCoords()` also with no result target, so the result is also stored in `_eventCoords`, which modifies the value of `mouse` which is then used for picking on other sources but with a wrong value.
* Don't pick atmosphere layer (I cannot see any use case)

### DIFF
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -99,6 +99,7 @@ function findLayerInParent(obj) {
 }
 
 const raycaster = new THREE.Raycaster();
+const normalized = new THREE.Vector2();
 
 /**
  * @module Picking
@@ -208,8 +209,9 @@ export default {
         } else {
             raycaster.layers.enableAll();
         }
+        // Raycaster use NDC coordinate
+        view.viewToNormalizedCoords(viewCoords, normalized);
         if (radius < 0) {
-            const normalized = view.viewToNormalizedCoords(viewCoords);
             raycaster.setFromCamera(normalized, view.camera.camera3D);
 
             const intersects = raycaster.intersectObject(object, true);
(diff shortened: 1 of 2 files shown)


src/Core/Picking.js
```diff
@@ -208,8 +208,10 @@ export default {
         } else {
             raycaster.layers.enableAll();
         }
+        // Raycaster use NDC coordinate
+        const normalized = new THREE.Vector2();
```
mgermerie: Could you instantiate this `THREE.Vector2()` outside of the method (the same way as `raycaster` constant is implemented in line 100 for instance) ? If I got it wright, implementing a new constant at each call of a method implies that the garbage collector is used to clear memory space allocated to each constant, which takes a bit of time.

mgermerie (commented):
Hi,

Thanks for this PR ! Except for a minor detail I left in a review comment, it seams all good for me.

src/Core/Picking.js
jailln: Done :)

mgermerie (approved):
Perfect, thanks :+1:  ! I shall merge it as soon as functional tests will pass (they currently fail as explained in issue #1877)

mgermerie:
Could you please rebase your branch on master ? For a reason I can't figure out, I am not able to re-run tests manually, so rebasing the PR branch should do the trick.

jailln:
Rebased

Issue #1868, opened 2022-06-30
3D measurement tools


jailln:
We are interested in having 3D measurement tools and in particular:

* a tool to mesure the elevation (terrain or 3D models) in the scene:

![image](https://user-images.githubusercontent.com/16967916/176690961-17a98f6f-1d53-42e7-87f0-0c48cb85a5ad.png)

* a tool to mesure the distance and the height between two points:

![image](https://user-images.githubusercontent.com/16967916/176691507-7dc978e1-f49b-46c7-92dd-f5d9b945070c.png)

I was thinking to add these tools as widgets. Would this be the good way to add such tools ?

mgermerie:
Hi !
You're wright, the widgets would be a relevant place for this, as they theoretically should gather all user graphical tools and interface. 
Moreover, the addition of a tool to pinpoint coordinates -including elevation- on the terrain would indeed be interesting. An extension of this tool to measure height and distance between two points would then be pretty straight forward -the most difficult part being displaying the rectangular triangle as the one shown in your second screenshot.

jailln:
Hi,
Thanks for your feedback. I'm currently working on it on [this branch](https://github.com/iTowns/itowns/tree/feat/elevation-measure). I still have a few things to improve but the main features are working.

PR #1881, opened 2022-08-09
Elevation measure widget

jailln:
## Description
Add a widget to measure elevation anywhere in the 3D scene. Also contains a few modifications of the API (`Picking` and `View`).

I will write tests and squash commits once we agree on the implementation :) (i.e. after a first review)

Note that I opened a [PR on itowns2-sample-data](https://github.com/iTowns/iTowns2-sample-data/pull/8) for the assets related to this widget (sprite, logo and dataset for the example). 

## Motivation and Context
Implementation of the first tool described in #1868

Mesure elevation on the terrain:

image

Mesure elevation on any 3D object:
image

### DIFF
--- a/src/Core/Geographic/Coordinates.js
+++ b/src/Core/Geographic/Coordinates.js
@@ -58,7 +58,11 @@ class Coordinates {
     /**
      * @constructor
      *
-     * @param {string} crs - A supported crs (see the `crs` property below).
+     * @param {string} crs - A supported Coordinate Reference System. 'EPSG:4978' and 'EPSG:4326' are
+     * supported by default. To use another CRS, you have to declare it with proj4. For instance:
+     * @example
+     * itowns.proj4.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+     * You can find most projections and their proj4 code at [epsg.io]{@link https://epsg.io/}
      * @param {number|Array<number>|Coordinates|THREE.Vector3} [v0=0] -
      * x or longitude value, or a more complex one: it can be an array of three
      * numbers, being x/lon, x/lat, z/alt, or it can be `THREE.Vector3`. It can
@@ -90,6 +94,15 @@ class Coordinates {
         }
 
         this._normalNeedsUpdate = true;
+    }
+
+    /**
+     * Sets the Coordinate Reference System.
+     * @param {String} crs Coordinate Reference System (e.g. 'EPSG:4978')
+     */
+    setCrs(crs) {
+        CRS.isValid(crs);
+        this.crs = crs;
     }
 
     /**
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -215,6 +215,8 @@ export default {
         const clearG = Math.round(255 * clearColor.g);
         const clearB = Math.round(255 * clearColor.b);
 
+        // Raycaster use NDC coordinate
+        const normalized = view.viewToNormalizedCoords(viewCoords);
         const tmp = normalized.clone();
         traversePickingCircle(radius, (x, y) => {
             // x, y are offset from the center of the picking circle,
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import RenderMode from 'Renderer/RenderMode';
 import { unpack1K } from 'Renderer/LayeredMaterial';
+import Coordinates from 'Core/Geographic/Coordinates';
 
 const depthRGBA = new THREE.Vector4();
 // TileMesh picking support function
@@ -84,6 +85,11 @@ function findLayerInParent(obj) {
 const raycaster = new THREE.Raycaster();
 const normalized = new THREE.Vector2();
 
+const pointPos = new THREE.Vector3();
+const pointPosCoord = new Coordinates('EPSG:4978'); // default crs, will be set to view crs when used
+const cameraPos = new THREE.Vector3();
+const cameraPosCoord = new Coordinates('EPSG:4978'); // default crs, will be set to view crs when used
+
 /**
  * @module Picking
  *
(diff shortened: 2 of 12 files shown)


mgermerie:
There is an issue with the Elevation widget : if you use it while zoomed out, the elevation is not red on the most detailed tile (since it is not necessarily loaded yet). The value will therefore differ on the same point, whether you use the widget while zoomed out or zoomed in at maximum.

src/Core/View.js
```diff
@@ -1057,9 +1057,9 @@ class View extends THREE.EventDispatcher {
      * @param   {Coordinates}           [target]    The result will be copied into this {@link Coordinates}. If not
                                                     * specified, a new {@link Coordinates} instance will be created.
      *
-     * @returns {Coordinates}   The world {@link Coordinates} at the given view coordinates.
+     * @returns {Coordinates}   The world {@link Coordinates} of the terrain at the given view coordinates.
      */
-    pickCoordinates(mouse, target = new Coordinates(this.tileLayer.extent.crs)) {
+    pickTerrainCoordinates(mouse, target = new Coordinates(this.referenceCrs)) {
```
mgermerie, line 1042: Renaming this method introduces a breaking change. Could you still define `pickCoordinates` with a deprecation warning and a calling to the new method ?

src/Core/View.js
```diff
@@ -1072,9 +1072,7 @@ class View extends THREE.EventDispatcher {
         }
 
         this.getPickingPositionFromDepth(_eventCoords, positionVector);
-        coordinates.crs = this.referenceCrs;
-        coordinates.setFromVector3(positionVector);
-        coordinates.as(target.crs, target);
+        target.setFromVector3(positionVector);
```
mgermerie: We need to leave the transformation here, since use should be able to call the method with any `target` coordinate. We can't assume the CRS of `target`. For example if I type : 
```js
const coordinates = new itowns.Coordinates('EPSG:4326');
view.pickTerrainCoordinates({ /* some mouse position */ }, coordinates);
```
it will return some coordinates with an `EPSG:4326` CRS but whose `x`, `y` and `z` values are intended as `EPSG:4978`, so most probably not the coordinates under the mouse position.

src/Core/View.js
mgermerie, line 1042: We can indeed change the default target CRS with the view's reference CRS, as it seems more logical.
However we should keep these two following points in mind : 
- this introduces a breaking change for some users (who for example feed the default output of this method to other processes without transforming it).
- In a `PlanarView` context, I believe that `view.referenceCrs` and `view.tileLayer.extent.crs` are always the same. Having the default output CRS to `view.tileLayer.extent.crs` allows to output coordinates in longitude, latitude and altitude, whose representation is more easy to understand than coordinates in `EPSG:4978`.

Therefore, do you still think we should change it ? In any case, could you please add the previously missing documentation on the default output CRS ?

src/Core/Geographic/Coordinates.js
```diff
@@ -58,15 +58,19 @@ class Coordinates {
...
+     * You can find most projections and their proj4 code at [epsg.io]{@link https://epsg.io/}
      * @param {number|Array<number>|Coordinates|THREE.Vector3} [v0=0] -
      * x or longitude value, or a more complex one: it can be an array of three
      * numbers, being x/lon, x/lat, z/alt, or it can be `THREE.Vector3`. It can
      * also simply be a Coordinates.
      * @param {number} [v1=0] - y or latitude value.
      * @param {number} [v2=0] - z or altitude value.
      */
-    constructor(crs, v0 = 0, v1 = 0, v2 = 0) {
+    constructor(crs = 'EPSG:4978', v0 = 0, v1 = 0, v2 = 0) {
```
mgermerie: The documentation is a good addition, thanks !
On another hand, I don't think we should assume a default CRS for coordinates. No coordinates make sense without a CRS and in my opinion, user should always have to specify the CRS associated with the coordinates values he is dealing with.

src/Core/Geographic/Coordinates.js
```diff
@@ -92,6 +96,15 @@ class Coordinates {
         this._normalNeedsUpdate = true;
     }
 
+    /**
+     * Sets the Coordinate Reference System.
+     * @param {String} crs Coordinate Reference System (e.g. 'EPSG:4978')
+     */
+    setCrs(crs) {
```
mgermerie, line 103: Good idea to add this check of parameter validity !

src/Core/Picking.js
```diff
@@ -187,9 +188,16 @@ export default {
                 // if baseId matches objId, the clicked point belongs to `o`
                 for (let i = 0; i < candidates.length; i++) {
                     if (candidates[i].objId == o.baseId) {
+                        // Compute distance to the camera
+                        const pointCoordinate = new Coordinates(view.referenceCrs, o.position.x, o.position.y, o.position.z);
```
mgermerie: I think this method for computing distance could cause issues : the `position` property for threejs `Object3D` is expressed in the object's parent coordinates system.
Here, the point could be centered within a `Group`, and the `Group` could be positioned at the correct world coordinates. the `position` of the point would therefore be `(0, 0, 0)`. Using world position for the point, as you did for the camera would prevent this.

mgermerie: Also, do you know of an usage of this picking method other than with `PointCloudLayer` ?

src/Core/View.js
jailln: I think yes because it is clearer for users since in my opinion users don't really know the crs of the extent of the tileLayer which is quite internal to itowns, they should expect to get coordinates in the view CRS which they actually know of. You're right that it is a breaking change, I will add one.

src/Core/Geographic/Coordinates.js
jailln: Yes I agree with you, I will remove the default CRS.

src/Core/View.js
jailln: Yep and I will also add a breaking change in the commit message.

src/Core/View.js
jailln: yep you're right, I got mixed up :)

src/Core/Picking.js
jailln: Yes, you're right, I added the world transformation and updated a bit this part to fix a bug that I had not seen for point cloud picking.

jailln:
@mgermerie I made the modifications you recommended, squashed and rebased.

jailln:
Some functional tests don't pass (timeout), while they pass locally. I tried to update a few things but it didn't change anything. Any ideas @mgermerie @gchoqueux ?

jailln:
I won't have time to finish this feature in the short term, closing for now.

PR #1884, opened 2022-08-17
fix(THREE): change THREE.Math to THREE.MathUtils

lafflan:
threejs has no such class of Math

### DIFF
--- a/src/Renderer/Camera.js
+++ b/src/Renderer/Camera.js
@@ -153,8 +153,8 @@ class Camera {
                 this.camera3D.bottom = -halfH;
                 this.camera3D.top = halfH;
             } else if (this.camera3D.isPerspectiveCamera) {
-                this.camera3D.fov = 2 * THREE.Math.radToDeg(Math.atan(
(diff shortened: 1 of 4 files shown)


mgermerie:
Hi ! Thanks for this PR.

You are wright, `THREE.Math` does not exists in threejs library. Yet, for a reason I need to figure out, there is actually a `Math` property within `itowns.THREE` which seams to be the same as `MathUtils`.

Could you squash your two commits into a single one, and name in agreement with the Angular convention (see the [CONTRIBUTING](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md#pull-request-guidelines) guide) ? Here, you should have something like `fix(THREE): change THREE.Math to THREE.MathUtils`.

Also, I found two more usage of `THREE.Math` in unit test (`CameraUtils` and `globeview`). I shall add their replacement to your PR before merging.

lafflan:
@mgermerie done

mgermerie:
Thanks ! I added the two changes in unit tests I mentioned earlier.

mgermerie (approved):

PR #1885, opened 2022-08-22
fix(Scheduler): replace url subdomains alternatives

### LINKED ISSUE #1883: Sub-domains URL templating does not work
Either using sub-domains URL templating (such as described [here](https://github.com/iTowns/itowns/blob/master/src/Provider/URLBuilder.js#L25)) does not work or I don't use it properly.

The following error is thrown:

```
ypeError: Failed to construct 'URL': Invalid URL
    at Scheduler.execute (Scheduler.js:181:1)
    at updateLayeredMaterialNodeImagery (LayeredMaterialNodeProcessing.js:142:1)
    at ColorLayer.update (ColorLayer.js:145:46)
    at updateElements (MainLoop.js:88:1)
    at MainLoop._update2 (MainLoop.js:274:1)
    at MainLoop._step5 (MainLoop.js:326:1)
    at MainLoop.js:212:1
```

## Your Environment
* Version used: 2.38.2
* Browser Name and version: Google Chrome Version 104.0.5112.79 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): Ubuntu 20.04

## Context
Url sub domains templating has been implemented in #1189

1. Create a GlobeView
2. Add a TMSSource with the following config:
```
{
        "crs": "EPSG:3857",
        "isInverted": true,
        "format": "image/png",
        "url": "https://${u:a|b|c}.tile.openstreetmap.org/${z}/${x}/${y}.png",
        "attribution": {
            "name":"OpenStreetMap",
            "url": "http://www.openstreetmap.org/"
        },
        "tileMatrixSet": "PM"
}
```
4. Add a ColorLayer with this TMSSource
5. Add the layer to the globe
6. Launch the app in your browser

mgermerie:
## Description

When using subdomains alternatives in a `Source` `url` property, the Scheduler tries to generate a new `URL` object passing an incorrect URL. The subdomains alternatives are not replaced, it will be something like `https://${u:xyz.org|yzx.org|zxy.org}/${z}/${x}/${y}.png` instead of `https://yzx.org/${z}/${x}/${y}.png` for instance. 

This PR adds the use of [this method](https://github.com/iTowns/itowns/blob/master/src/Provider/URLBuilder.js#L6) to replace subdomains alternative by the relevant one before generating an `URL` object in `Scheduler`.

## Motivation and Context

fixes #1883

### DIFF
--- a/src/Core/Scheduler/Scheduler.js
+++ b/src/Core/Scheduler/Scheduler.js
@@ -9,6 +9,7 @@ import DataSourceProvider from 'Provider/DataSourceProvider';
 import TileProvider from 'Provider/TileProvider';
 import $3dTilesProvider from 'Provider/3dTilesProvider';
 import PointCloudProvider from 'Provider/PointCloudProvider';
+import URLBuilder from 'Provider/URLBuilder';
 import CancelledCommandException from './CancelledCommandException';
 
 function queueOrdering(a, b) {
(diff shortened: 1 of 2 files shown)


jailln:
Thanks! I've tested it and it fixes #1883

gchoqueux (approved):

PR #1886, opened 2022-08-23
Improve resources disposal

jailln:
## Description

Improve resource disposal by:
* using `ObjectRemovalHelper` to remove 3D objects when a `GeometryLayer` is deleted
* disposing textures in `ObjectRemovalHelper`
* allowing to clear the cache or not when disposing the view and layers

## Motivation and Context

Some resources where not disposed after disposing itowns view. To make sure of it, you can add `setTimeout(function() {view.dispose();}, 40000);` at the end of any itowns example. Some tests I made with the chrome Memory tool, on two itowns examples:

### Before the MR

**3dtiles_basic.html:**

*Before disposing the view*
3dtiles_basic_before

*After disposing the view*
3dtiles_basic_after

Notice that nearly any memory is freed and that there is still `ArrayBuffers` which contains geometries

**vector_tile_3d_mesh.html:**

*Before disposing the view*
vector_tile_3d_mesh_before

*After disposing the view*
vector_tile_3d_mesh_after

Same here, not much is deleted

### After the MR

*Before disposing the view*
image

*After disposing the view*
image

It's not perfect but more resources are disposed.

**vector_tile_3d_mesh.html:**

I have less success with this one :grin: 

*Before disposing the view*
image

*After disposing the view*
image

There is still many `Feature` and `FeatureGeometry` after the view disposal, but I haven't found out where they are stored and how to delete them. Note that the cache and the threejs object3Ds are deleted the same way than for `c3DTileslayer` since both `c3DTilesLayer` and `FeatureGeometryLayer` inherit from `GeometryLayer`.  Help wanted for this one :slightly_smiling_face: 

## Open discussion

I think we could still improve some other things. Let me know what you think, if you agree I will implement them:
* Harmonize functions names for resource disposal (some are named dispose, other delete, other clear, other remove, etc.): rename them all to `dispose` (the name used in threejs)
* Some caches are declared as constants outside classes and are not exported: `cacheStyle` in `Style.js`, `cacheTile` and `cacheBuffer` in `TileBuilder.js`. Therefore they cannot be cleared properly when disposing the view. I think they should be stored in the view so we can clear them when needed.
* Fetched data are cached twice:
  * Once after being parsed, in `Source._featuresCaches`
  * Once after being converted, in `Layer.cache`
I think this greatly affects itowns memory footprint, especially when loading heavy dataset. In addition, I don't think there is a case where only one cache is deleted: they are both deleted when the layer is deleted (or when the cache lifetime has expired). WDYT about caching them only once ? (in `Layer.cache` to avoid transforming them again)

### DIFF
--- a/src/Layer/OrientedImageLayer.js
+++ b/src/Layer/OrientedImageLayer.js
@@ -108,10 +108,14 @@ class OrientedImageLayer extends GeometryLayer {
         }
         super(id, new THREE.Group(), config);
 
-        this.background = config.background || createBackground(config.backgroundDistance);
         this.isOrientedImageLayer = true;
 
+        this.background = config.background || createBackground(config.backgroundDistance);
+
         if (this.background) {
+            // Add layer id to easily identify the objects later on (e.g. to delete the geometries when deleting the layer)
+            this.background.layer = this.background.layer ?? {};
+            this.background.layer.id = this.background.layer.id ?? id;
             this.object3d.add(this.background);
         }
 
@@ -205,9 +209,16 @@ class OrientedImageLayer extends GeometryLayer {
      * Delete background, but doesn't delete OrientedImageLayer.material. For the moment, this material visibility is set to false.
      * You need to replace OrientedImageLayer.material applied on each object, if you want to continue displaying them.
      * This issue (see #1018 {@link https://github.com/iTowns/itowns/issues/1018}) will be fixed when OrientedImageLayer will be a ColorLayer.
-     */
-    delete() {
-        super.delete();
+    * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
+    */
+    delete(clearCache) {
+        if (this.background) {
+            // only delete geometries if it has some
+            super.delete();
+        }
+        if (clearCache) {
+            this.cache.clear();
+        }
         this.material.visible = false;
         console.warn('You need to replace OrientedImageLayer.material applied on each object. This issue will be fixed when OrientedImageLayer will be a ColorLayer. the material visibility is set to false. To follow issue see https://github.com/iTowns/itowns/issues/1018');
     }
--- a/src/Process/ObjectRemovalHelper.js
+++ b/src/Process/ObjectRemovalHelper.js
@@ -1,3 +1,14 @@
+function disposeSingleMaterialAndTextures(material) {
+    material.dispose();
+    // dispose textures
+    for (const key of Object.keys(material)) {
+        const val = material[key];
+        if (val && val.isTexture) {
+            val.dispose();
+        }
+    }
+}
+
 export default {
     /**
      * Cleanup obj to release three.js allocated resources
@@ -6,7 +17,8 @@ export default {
     cleanup(obj) {
         obj.layer = null;
 
-        if (typeof obj.dispose === 'function') {
+        // THREE.Scene dispose function displays a deprecation message
+        if (!obj.isScene && typeof obj.dispose === 'function') {
             obj.dispose();
         } else {
             if (obj.geometry) {
@@ -20,10 +32,10 @@ export default {
             if (obj.material) {
                 if (Array.isArray(obj.material)) {
                     for (const material of obj.material) {
-                        material.dispose();
+                        disposeSingleMaterialAndTextures(material);
                     }
                 } else {
-                    obj.material.dispose();
+                    disposeSingleMaterialAndTextures(obj.material);
                 }
             }
             obj.dispatchEvent({ type: 'dispose' });
@@ -44,7 +56,7 @@ export default {
     },
 
     /**
-     * Remove obj's children belonging to a layer and cleanup objexts.
+     * Remove an obj and all its children belonging to a layer and only cleanup the obj (and not its children).
      * obj will be disposed but its children **won't**!
      * @param {Layer} layer The layer that objects must belong to. Other object are ignored
      * @param {Object3D} obj The Object3D we want to clean
@@ -61,13 +73,15 @@ export default {
     },
 
     /**
-     * Recursively remove obj's children belonging to a layer.
+     * Recursively remove an obj and all its children belonging to a layer.
      * All removed obj will have their geometry/material disposed.
      * @param {Layer} layer The layer that objects must belong to. Other object are ignored
      * @param {Object3D} obj The Object3D we want to clean
      * @return {Array} an array of removed Object3D from obj (not including the recursive removals)
      */
     removeChildrenAndCleanupRecursively(layer, obj) {
+        // Objects are filtered by id because the obj hierarchy may also contain labels that have been added as childs
+        // of the objects which have their own removal logic
         let toRemove = obj.children.filter(c => (c.layer && c.layer.id) === layer.id);
         if (obj.link) {
             toRemove = toRemove.concat(obj.link);
--- a/docs/tutorials/Fundamentals.md
+++ b/docs/tutorials/Fundamentals.md
@@ -76,6 +76,8 @@ Some point clouds formats such as Potree, Las and Entwine already have parsers d
 | :---: | :---: | :---: |
 | A simple `ColorLayer` is displayed | An `ElevationLayer` is added to represent terrain elevation | A `GeometryLayer` is added to model the buildings |
 
+Note that `{@link ColorLayer}` and `{@link ElevationLayer}` don't have their own geometry and are always *attached* to a `{@link GeometryLayer}` (generally the layer representing the geometry of the globe or of the plane). `{@link ColorLayer}` are projected onto this `{@link GeometryLayer}` and `{@link ElevationLayer}` are used to read elevation and to apply it to a `{@link GeometryLayer}`.
+
 ***
 
 ## Style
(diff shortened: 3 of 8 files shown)


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -237,6 +237,9 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
 
 export function removeLayeredMaterialNodeLayer(layerId) {
     return function removeLayeredMaterialNodeLayer(node) {
+        if (node.geometry) {
+            node.geometry.dispose();
+        }
```
mgermerie: I don't think the geometry should be disposed when removing a `RasterLayer`. The geometry of each tile should not depend on layers attached to the tile. This allows us to remove a `ColorLayer` without causing issue for an `ElevationLayer` visualization for instance.

On the screenshots bellow, you can se the effect caused by disposing each tile geometry when removing a ColorLayer, and what should happen (without disposing geometry).

Before removing `ColorLayer` :


After removing `ColorLayer` and disposing tiles geometry : 


What should normaly happen (if you don't dispose geometry) : 


For this third one, this is not exactly what happens, I just ran into some glitch and I shall make an issue about it (some of the tiles do not display elevation data until the camera moves and the view is updated. However, the tiles are still displayed, unlike in the second screenshot).

src/Layer/OrientedImageLayer.js
```diff
@@ -205,9 +205,12 @@ class OrientedImageLayer extends GeometryLayer {
      * Delete background, but doesn't delete OrientedImageLayer.material. For the moment, this material visibility is set to false.
      * You need to replace OrientedImageLayer.material applied on each object, if you want to continue displaying them.
      * This issue (see #1018 {@link https://github.com/iTowns/itowns/issues/1018}) will be fixed when OrientedImageLayer will be a ColorLayer.
-     */
-    delete() {
-        super.delete();
+    * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
+    */
+    delete(clearCache) {
+        if (clearCache) {
```
mgermerie, line 219: The call to `super.delete()` is indeed not needed if `background` property for `OrientedImageLayer` has not been set. In this case, `OrientedImageLayer.object3d` is just an empty `THREE.Group()`, with nothing to dispose.

However, if `background` property is set (as it is the case in [immersive view example](http://www.itowns-project.org/itowns/examples/#view_immersive)), `OrientedImageLayer.object3d` can contain meshes. Therefore, these meshes need to be disposed by calling `super.delete()`.

In the `GeometryLayer.delete` method modified with this PR, the objects that are disposed are filtered (only those with the correct `layer.id` property are disposed, which was not the case before). The issue with `OrientedImageLayer` objects is that they lack the `layer.id` property, which should therefore be added.

To summarise : I think we should add `super.delete()` back, and add something like the two following lines in `OrientedImageLayer` constructor, just above line 115.

```js
this.background.layer = this.background.layer ?? {};
this.background.layer.id = id;
```

This will allow `OrientedImageLayer.background` meshes to be correctly disposed if they exist.

What do you think about this ?

src/Process/LayeredMaterialNodeProcessing.js
jailln: I agree with you that `The geometry of each tile should not depend on layers attached to the tile.` 
However, I thought that `node.geometry` represented the own geometry of the `RasterLayer` if it was not attached, but now I'm not sure of it. Can a`RasterLayer` have its own geometry or is it always attached to a GeometryLayer? Is `node.geometry` always the geometry of the layer it is attached too?

src/Layer/OrientedImageLayer.js
jailln: Yep I think you're right, I will make the modifications :)

Side question: Regarding the filtering of the geometries by layer id in ObjectRemovalHelper, which usecase does it serve? I cannot think of a case where a GeometryLayer would have geometries that don't have the id of this geometry layer, is there one ?

src/Process/LayeredMaterialNodeProcessing.js
mgermerie: `RasterLayers` are always attached to another `GeometryLayer` - which most of the times is a [`TiledGeometryLayer`](https://github.com/iTowns/itowns/blob/master/src/Layer/TiledGeometryLayer.js). This `TiledGeometryLayer` displays the globe -or planar- tiles with a default material (the blue color we see if no `ColorLayer` is attached). When we add a `RasterLayer` : 
- if it is a `ColorLayer`, the image is used as a texture for the tile material,
- if it is an `ElevationLayer`, the `z` coordinate of each tile vertex is translated by the correct elevation.

When removing a `RasterLayer`, we take the tiles from the minimum zoom level, and we traverse all their children which are passed to `removeLayeredMaterialNodeLayer` parameter. Therefore, the `node` in this case is always a tile from the `TiledGeometryLayer` to which the `RasterLayer` is attached.

To conclude and answer your question : 
- a `RasterLayer` does not have its own geometry and is always attached to a `GeometryLayer`.
- `node.geometry` is always the geometry of the layer the `RasterLayer` is attached to.

I hope this is clear enough, don't hesitate if it needs clarification.

_PS : each tile from a `TiledGeometryLayer` actually has five children : the four tiles matching the subsequent zoom level subdivision, and the tile's OBB. Perhap this could be improved, having the OBB as a property and not a children._

jailln:
My remarks regarding the cache have been moved to #1928

src/Layer/OrientedImageLayer.js
mgermerie: In case a `LabelLayer` is added, the labels are added as children of a tile. Sometimes, the method is called with the tile as object parameter. The filtering prevents removing labels.

src/Process/LayeredMaterialNodeProcessing.js
jailln: Ok thanks for the explaination, it is clearer now! I removed this line and I added your explanation in the documentation.

Can you open issues about the problem you mention and about your proposition for OBB management please?

src/Layer/OrientedImageLayer.js
jailln: Done

jailln:
I also created issues #1937 and #1936 for the remaining remarks of my PR description.

src/Layer/OrientedImageLayer.js
```diff
@@ -205,9 +209,16 @@ class OrientedImageLayer extends GeometryLayer {
...
      * This issue (see #1018 {@link https://github.com/iTowns/itowns/issues/1018}) will be fixed when OrientedImageLayer will be a ColorLayer.
-     */
-    delete() {
-        super.delete();
+    * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
+    */
+    delete(clearCache) {
+        if (this.background) {
+            // only delete geometries if it has some
+            super.delete();
```
mgermerie, line 217: I noticed an inconsistency regarding this part : 

An `OrientedImageLayer` is never attached to a `GeometryLayer`. It has an empty `update` method which does not depend on a parent layer (i.e. a layer it could be attached to). The projection is handled in the shader.  Therefore, the `OrientedImageLayer.parent` property should always be `undefined`. 

However, when adding an `OrientedImageLayer` in a `GlobeView`, the `GlobeView.addLayer` method called enforce each layer's parent to its `TiledGeometryLayer` (i.e. the globe terrain). It leaves us with `OrientedImageLayers` whose `parent` property is the `GlobeLayer`.

This does not causes issue I am aware of. I just noticed it while looking if testing `this.background` was redundant with `GeometryLayer.delete` tests (and you're wright, it is not :smiley: )

jailln: Ahah ok, maybe we should open an issue about this ? And maybe `OrientedImageLayer` should be added `View`'s `addLayer` then ? : `itowns.View.prototype.addLayer.call(view, orientedImgLayer);`

jailln:
@mgermerie could you create the issues of the remaining points you noted and add your review please? I need this MR to solve #1947 :slightly_smiling_face:

src/Process/LayeredMaterialNodeProcessing.js
mgermerie: #1983 opened on this subject.

src/Layer/OrientedImageLayer.js
mgermerie: #1984 opened on the subject.

src/Process/ObjectRemovalHelper.js
```diff
@@ -1,3 +1,14 @@
...
+    material.dispose();
+    // dispose textures
+    for (const key of Object.keys(material)) {
+        const val = material[key];
+        if (val && val.isTexture) {
+            val.dispose();
+        }
+    }
+}
+
```
gchoqueux, line 11: what type of materials do you want to dispose?

jailln: I made it generic so all type of materials and their textures can be disposed with this method

jailln:
commits squashed and branch rebased

jailln:
@gchoqueux @mgermerie is it ok for you? can we merge it ?

gchoqueux (approved):
it's ok for me

PR #1887, opened 2022-09-05
chore: add submodule for examples.

kgtnyshhd:
## Description
Add submodule for examples.

## Motivation and Context
After #1817, it seems that "exmaples/js/*.js" cannot be imported.
It would be great if it allowed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "main": "lib/Main.js",
   "exports": {
       ".": "./lib/Main.js",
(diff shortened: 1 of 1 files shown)


mgermerie:
Hi, thanks for your contribution ! 
I made an issue regarding the export of plugins (#1930) so that it can be planed. We need first to specify what we want to export, and where we want to export it (as widgets, as an utils package, in sources...).

PR #1902, opened 2022-09-21
Add a boundingBox attribute to C3DTilesLayer and improve C3DTilesLayer doc

jailln:
Add a boundingBox attribute to C3DTilesLayer to ease processes based on 3D Tiles layers for itowns users (e.g. to help writting zoomTo methods, to say if a layer is visible or not, etc.).

I think a good improvement would be to add a bounding box attributes to all layers, see #1903 .

Also, I haven't written a test for the region case, see #1905 for an explaination.

### DIFF
--- a/src/Core/3DTiles/C3DTBoundingVolume.js
+++ b/src/Core/3DTiles/C3DTBoundingVolume.js
@@ -117,6 +117,26 @@ class C3DTBoundingVolume {
         }
         return false;
     }
+
+    /**
+     * Gets the bounding box of the bounding volume.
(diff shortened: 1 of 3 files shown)


jailln:
Still some improvements to make:
* Apply the transform matrix to the bounding box since 3D Tiles box and sphere (but not region) bounding boxes can be defined in local space
* Somehow the bounding box computation does not seem to work for point clouds (even after applying the transform matrix, and even after applying the transform matrix without the scale component)
* Find a way to allow the user to recompute the bounding box of the layer in case the embedded tileset bounding boxes are wrong 
* Repair the broken test

PR #1909, opened 2022-09-26
chore(integration): differentiate deployment for LTS and current version

mgermerie:
## Description

Add a deployment step to deploy on different repo if it is a release or if it is a merge on master branch.
This allow to have a "LTS" documentation and examples version (matching the latest release) and a "current" version (matching master branch).

TODO next : 

- [x] We could add another npm repository ('itowns-next', as @gchoqueux suggested) published each time a merge occur on master. This would grant an npm package for itowns latest 'unstable' version
- [x] I could add a button in itowns documentation and examples, allowing to switch between LTS and current version.
- [x] The changelog generation should be updated to prevent including bump commits for next versions

You can check the new versioning [here](https://www.npmjs.com/package/itowns-test-deployment-5?activeTab=versions).
You can also see how documentation and example would look [here for the doc](https://mgermerie.github.io/itowns.github.io/itowns/docs/#home) and [here for the examples](https://mgermerie.github.io/itowns.github.io/itowns/examples/index.html).


jailln:
I totally agree with the proposed "TODO next" :)

.github/workflows/integration.yml
```diff
@@ -7,180 +9,243 @@ on:
...
-
-  # Create github release
+
+      # Download artifact from build
+      - name: Download itowns bundle
+        uses: actions/download-artifact@v2
+        with:
+          name: dist-itowns
+
+      # Deploy potree repo artifact
```
jailln: Shouldn't we remove this part since if I remember correctly it won't be used in the examples but in a different repository that shows itowns and potree integration?

.github/workflows/integration.yml
```diff
@@ -7,180 +9,243 @@ on:
...
+        with:
+            name: dist-itowns
+            path: |
+                dist/**/*.js
+                examples
+                buildDocs
+
+
+  # Publish NPM package
+  # TODO: publish another npm package matching master branch
```
jailln: remove this todo?

.github/workflows/integration.yml
```diff
@@ -7,180 +9,243 @@ on:
...
+      # Copy files for deployment
+      - name: build site
+        run: |
+          mkdir -p itowns/dist
+          mkdir -p itowns/potree/build
+          mkdir -p itowns/potree/libs
+          cp -R dist/*.js itowns/dist/
+          cp -R examples itowns/
+          cp -R buildDocs itowns/docs
+          # cp -R build itowns/potree/
```
jailln: same remark for potree related configuration

.github/workflows/integration.yml
```diff
@@ -7,180 +9,243 @@ on:
...
+          deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
+          external_repository: iTowns/itowns.github.io
+          publish_dir: ./itowns
+          destination_dir: ./itowns
+          publish_branch: master
+          enable_jekyll: true
+
+      # Deploy to itowns.github.io Dev
+      - name: Deploy Dev to itowns.github.io
+        if: ${{ !startsWith(github.event.head_commit.message, 'release v' ) }}
```
jailln, line 164: I think we don't need this if here to get the latest commit version?

.github/workflows/integration.yml
```diff
@@ -7,180 +9,243 @@ on:
       - master
   workflow_dispatch: {}
 
-name: integration
-env:
-  CI: true
```
jailln, line 12: We don't need this env var anymore ?

docs/static/styles/itowns.css
```diff
@@ -76,10 +76,29 @@ nav h2 {
...
+    margin-bottom: 10px;
+}
+#switch-button:hover {
+    cursor: pointer;
+    background-color: #E5E5E5;
+}
+#switch-button:active {
+    background-color: #404E76;
+    color: white;
+}
```
jailln, line 99: This is more a detail but I think  we should work a bit more on the Switch button style to make it a bit more modern (e.g. colored background, more rounded borders, another font, etc.). I can send some inspirations if you want, let me know.

docs/tmpl/main.tmpl
```diff
@@ -128,9 +135,32 @@
         }
     }
 
+    // Add version switch
+    function initVersionSwitch() {
+        document.getElementById('version-switch-text').innerHTML = `
+            You are currently viewing iTowns documentation for the ${
```
jailln: I would update the text a bit to make the difference between the two versions clearer to users. I propose something like:

"iTowns documentation/example of the latest stable release version (vXX.XX.XX)." 
and
"iTowns documentation/example of itowns next version (current master branch)."

I would also update the button texts to:
"Switch to latest release (vXX.XX.XX)"
"Switch to next version"

examples/index.html
```diff
@@ -16,13 +16,17 @@
 
 <nav class="scroll-section">
     <a href="/">
-        <img src="images/itowns_logo.png" width="245px"/>
+        <img src="images/itowns_logo.png" width="245px" id="itowns-logo">
```
jailln, line 19: missing `/` at the end?

README.md
```diff
@@ -78,6 +78,21 @@ See our [release page](https://github.com/iTowns/itowns/releases). Note that
 there isn't a lot of support for older version of iTowns, we highly recommand to
 use the last release everytime.
 
+### Try modifications before they are released
+
+Should you want to try some features or bug fixes that are planned for the next release, we provide
```
jailln: Should -> If ?

jailln (changes requested):
Thanks for this MR! :) 

A more general remark: Is it possible to create a job to automatically remove next versions from the npm registry at some point? For instance, delete @next versions after two new stable releases (two is arbitrary here, we can discuss it). If we do so, we also need to document it in the main README.

.github/workflows/integration.yml
mgermerie: I took inspiration from [this thread](https://stackoverflow.com/questions/60006265/whats-the-purpose-of-the-environment-variable-ci-set-to-true-within-a-github-wo), and since it is not used anywhere in tests or other scripts, I took the liberty to remove it.

.github/workflows/integration.yml
mgermerie: Yep, I forgot to remove it.

.github/workflows/integration.yml
mgermerie: Indeed, I removed it.

.github/workflows/integration.yml
mgermerie: Removed as well.

.github/workflows/integration.yml
mgermerie: I had to do this to prevent having two pushes on iTowns/itowns.github.io when making a release. This would cause an error on the second push, since apparently the same ssh connection can't be used simultaneously for multiple transfers. I tried to setup two separate ssh connections with iTowns/itowns.github.io : one for pushing releases and the other one for pushing next build bundles, but this also failed.
What I do then is preventing the deployment of @next version when deploying a release (thanks to this `if` statement), and copying the release bundles in the dev folder on iTowns/itowns.github.io (on lines 153 to 157 of this script).

README.md
mgermerie: Indeed, I just learned that the use of "should" in conditional sentences suggests that the clause it describes is unlikely to happen, which is not the case here :slightly_smiling_face:

examples/index.html
mgermerie: `/` at the end of self-closing tags (such as `img`) is not needed in html5, as I red [here](https://stackoverflow.com/a/15149657)

docs/static/styles/itowns.css
mgermerie: I admit I lacked inspiration on this one. Any suggestion would indeed be welcome.

docs/tmpl/main.tmpl
mgermerie: Yep, I shall make the modification. However, I don't think getting the version number in the docs template will be easy. I'll see what I can find about this.

.github/workflows/integration.yml
jailln: Ah ok! thanks for the explaination! Maybe we should add this explanation as a comment ?

docs/tmpl/main.tmpl
jailln: Ok, if it's too complicated don't do it, it's not so important

mgermerie:
> Thanks for this MR! :)
> 
> A more general remark: Is it possible to create a job to automatically remove next versions from the npm registry at some point? For instance, delete @next versions after two new stable releases (two is arbitrary here, we can discuss it). If we do so, we also need to document it in the main README.

Thanks for the review ! :slightly_smiling_face: 

Unpublishing a package from the npm registry blocks any publication of a package with the same name for 24 hours. This means that regularly unpublishing @next versions would regularly block us from merging PRs on master branch (we could actually merge, but changes would not be published on a specific @next version).

Moreover, I think any version of a package (intentionally published, not one published by accident or for testing) should remain always available for download. The use case would be for instance a user who wants to test the differences brought by a specific commit in a specific PR, even if it's an old commit.

In any generic case, I believe (but I might be wrong about that) users will download @next version with `npm install itowns@next` without specifying the @next version number. Even if they would, I can't think of an issue with having all @next versions still published. Did you have a specific one in mind ?

.github/workflows/integration.yml
mgermerie: Yep, I added comments here and before line 153.

jailln:
> Even if they would, I can't think of an issue with having all @next versions still published. Did you have a specific one in mind ?

Not really, it's just to avoid storing too many releases that may not be used anymore at some point but let's leave it like this for now (especially because of the 24h rule) and if we need to remove versions at some point we'll figure something out.

jailln (approved):

gchoqueux (approved):

Issue #1937, opened 2022-10-24
Harmonize dispose function names


jailln:
Harmonize functions names for resource disposal (some are named dispose, other delete, other clear, other remove, etc.): rename them all to dispose (the name used in threejs).

jamesmyose:
I can do this one. Just to be clear, just renaming the functions for resource disposal in src?

jamesmyose:
I'm currently working on this issue.

PR #1938, opened 2022-10-25
Harmonize Dispose Function Names

jamesmyose:
## Description
- Renamed delete function names to dispose

## Motivation and Context
This is work on #1937. I found some delete functions that I could change, but I was unsure about changing anything else.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,152 @@
+<a name="2.40.0"></a>
+# [2.40.0](https://github.com/iTowns/itowns/compare/v2.38.2...v2.40.0) (2023-07-28)
+
+
+### Features
+
+* **3dtiles:** add classification for rendering pointscloud ([b924cd7](https://github.com/iTowns/itowns/commit/b924cd7))
+* **3DTiles:** add style API + C3DTFeature ([864268a](https://github.com/iTowns/itowns/commit/864268a))
+* **3dtiles:** add support for 3d tiles from cesium ion server ([e9793a3](https://github.com/iTowns/itowns/commit/e9793a3))
+* **3DTiles:** add C3DTILES_LAYER_EVENTS.ON_TILE_REQUESTED ([ec837c7](https://github.com/iTowns/itowns/commit/ec837c7))
+* **Coordinates:** Add a setCrs method ([606b6b8](https://github.com/iTowns/itowns/commit/606b6b8))
+* **elevationLayers:** Add config.clampValues for clamping values of the elevation dataset ([1985078](https://github.com/iTowns/itowns/commit/1985078))
+* **instancing:** use instancing for large number of 3d objects ([619a611](https://github.com/iTowns/itowns/commit/619a611))
+* **LabelLayer:** add option to clamp labels to terrain. ([f46ca97](https://github.com/iTowns/itowns/commit/f46ca97))
+* **Parser:** detect if original source have elevation data. ([17aaa8b](https://github.com/iTowns/itowns/commit/17aaa8b))
+* **picking:** Add distance and point position to returned object by points picking ([97a5e6a](https://github.com/iTowns/itowns/commit/97a5e6a))
+* **planarcontrols:** add enabled attribute ([f13a060](https://github.com/iTowns/itowns/commit/f13a060))
+* **Style:** add icon.color and icon.opacity ([4fd5dc4](https://github.com/iTowns/itowns/commit/4fd5dc4))
+* **typescript:** Add devDeps on three's definitions ([2fbbcc3](https://github.com/iTowns/itowns/commit/2fbbcc3))
+* **typescript:** Add tsconfig.json configuration ([3c8c5f5](https://github.com/iTowns/itowns/commit/3c8c5f5))
+* **typescript:** Add typescript dependency ([71a2370](https://github.com/iTowns/itowns/commit/71a2370))
+* **view:** allow to pass an array of layers for picking ([9b6d59f](https://github.com/iTowns/itowns/commit/9b6d59f))
+* **Widget:** add C3DTilesStyle widget ([7862bc1](https://github.com/iTowns/itowns/commit/7862bc1))
+* **WMTS:** Support vendor specific parameters ([cff042c](https://github.com/iTowns/itowns/commit/cff042c))
+* **XbilParser:** gestion nodata in elevation layer (elevation set to 0) ([20075b8](https://github.com/iTowns/itowns/commit/20075b8))
+
+
+### Bug Fixes
+
+* **3dTiles:** addEventListener onTileContentLoaded constructor ([6f9a9d2](https://github.com/iTowns/itowns/commit/6f9a9d2))
+* **3dtiles:** fix batchtable reading ([b245301](https://github.com/iTowns/itowns/commit/b245301))
+* **3dtiles:** improve 3D Tiles cache cleaning ([284be24](https://github.com/iTowns/itowns/commit/284be24))
+* **3DTiles:** tileId == 0 can update style ([11582c9](https://github.com/iTowns/itowns/commit/11582c9))
+* **controls:** keyboard events are now on the domElement on the view instead of window ([d5c80f4](https://github.com/iTowns/itowns/commit/d5c80f4))
+* **CRS:** more robust parameter tests ([a2e0f5c](https://github.com/iTowns/itowns/commit/a2e0f5c))
+* **Debug:** fix 3dTiles bbox visibility ([cd8d106](https://github.com/iTowns/itowns/commit/cd8d106))
+* **example:** change config file linked with the clampValues config ([2a4e911](https://github.com/iTowns/itowns/commit/2a4e911))
+* **example:** invert order of the ElevationLayer addition when using 2 ([c6800c9](https://github.com/iTowns/itowns/commit/c6800c9))
+* **examples:** Fix  cesium ion token ([7f86d26](https://github.com/iTowns/itowns/commit/7f86d26))
+* **examples:** Replace Lyon's deprecated MNT sources ([7490590](https://github.com/iTowns/itowns/commit/7490590))
+* **examples:** Replace Lyon's deprecated Ortho2009 sources ([8dbb3d6](https://github.com/iTowns/itowns/commit/8dbb3d6))
+* **examples:** Update Lyon bus source options ([0881b2d](https://github.com/iTowns/itowns/commit/0881b2d))
+* **Feature2Texture:** prevent empty style ([ab5713a](https://github.com/iTowns/itowns/commit/ab5713a))
+* **GeoJsonParser:** fix firstCoordinates when empty ([e54f352](https://github.com/iTowns/itowns/commit/e54f352))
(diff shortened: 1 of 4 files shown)


jailln:
Thanks for this contribution! 

Could you modify the commit message so it follows the [angular convention](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md) please ? :) Also, it introduces breaking changes in the API so it must be added to the commit message. 

Currently (before this PR), the convention for 'removing' functions was:
* cache: `clear()` to empty it and `delete(smthg)` to delete smthg
* layer: `delete()` to remove a layer and dispose its content
* view: `dispose()` to dispose a view and its content
* controls: `dispose()` to dispose controls and their contents

The proposal is to move everything to `dispose()` to simplify the API. I think we should also change the cache `clear()` to `dispose()` and `delete()` to `remove()` (this last one to match threejs naming conventions for removal). 

@mgermerie WDYT ?

mgermerie:
Hi, and thanks for contributing !

I agree with the simplification proposition, as well as with what @jailln suggested for `cache` methods.

To prevent breaking changes we can simply add a console warning when using old methods name, and then call the method with its new name. With `Layers` for example, it would be something like this : 
```js
dispose() {
    // The body of the 'dispose' method, ex-content of the 'delete' method
}

delete() {
    console.warn('`Layer.delete` method is deprecated. Please use `Layer.dispose` instead.');
    this.dispose();
}
```

This warning could be remove later, in an eventual update of iTowns major version. WDYT @jailln ?

jailln:
Yep I agree with you @mgermerie . @jamesmyose can you make these modifications please ? :) And can you also modify the tests so they call the new functions please?

jamesmyose:
Yes, I'll get right on that.

jamesmyose:
I changed the tests for cache, but it looks like the other tests call the new methods or don't exist.

jamesmyose:
Alright, I got those changes in cache and those warning methods implemented.

jailln:
Thanks! I will look into it as soon as possible

jailln:
You can add your name in the [Contributors](https://github.com/iTowns/itowns/blob/master/CONTRIBUTORS.md) :)

jamesmyose:
Thank you! Just added myself :)

src/Core/Scheduler/Cache.js
```diff
@@ -192,20 +192,39 @@ class Cache {
         }
     }
 
+    /**
+     * Deprecated function. Use remove(key1, key2, key3)
+     * @param {string|number} key1
+     * @param {string|number} key2
+     * @param {string|number} key3
+     */
```
jailln: I think we shouldn't document the deprecated functions. Could you remove the comments on those please?

jailln:
Hi @jamesmyose,
Sorry for the delay, I was quite busy, I will be more reactive next time! 
Except from my small comment, the code is good for me. 

Could you squash your first three commits into one please? And could you remove the package-lock.json from the commited files please? (we update it at each release in general).

Then we will be ready to merge :)

src/Layer/GeometryLayer.js
```diff
@@ -193,6 +193,14 @@ class GeometryLayer extends Layer {
...
     }
 
+    /**
+     * Deprecated function. Use dispose()
+     */
+    delete() {
+        console.warn('`GeometryLayer.delete` method is deprecated. Please use `GeometryLayer.dispose` instead.');
+        this.dispose();
+    }
+
```
gchoqueux: Deprecated function `delete` could be only in `Layer` to avoid it being in every subclass  of `Layer`?

jamesmyose:
Sorry for the late reply. I didn't notice that there were new comments on the pull request. I'm not sure if I was able to do that squash correctly, but I got those other changes made.

src/Layer/GeometryLayer.js
```diff
@@ -171,7 +171,7 @@ class GeometryLayer extends Layer {
      * All layer's 3D objects are removed from the scene and disposed from the video device.
      * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
      */
-    delete(clearCache) {
+    dispose(clearCache) {
         if (clearCache) {
             this.cache.clear();
```
jailln: All Layers implementing this method start by:
```
if (clearCache) {
            this.cache.clear();
}
``` 

Could you factorize it in `Layer.dispose` please ? 

Also, you need to replace `this.cache.clear()`  by `this.cache.dispose()`.

jailln (changes requested):
Thanks for the modifications :) I just added one more comment.

Also, the squash didn't work as expected, could you squash all the commits into one please ? (we are a bit rigorous with the commit history sorry, it's because releases' changelogs are automatically generated from this commit history).
To do so, you can reset on the first commit before your commits with:
```
git reset --soft 8f7ae1ac3c2778c8bce9fe871b3bc31357bed27a
```
This will stage all your changes.
Then you can commit them all with `git commit` with the following commit message:
```
refacto(Cache, Layer): Rename Cache.clear() to Cache.dispose(), Cache.delete() to Cache.remove() and Layer.delete() to Layer.dispose()

BREAKING CHANGE: Cache.clear() has been renamed to Cache.dispose()
BREAKING CHANGE: Cache.delete() has been renamed to Cache.remove() 
BREAKING CHANGE: Layer.delete() has been renamed to Layer.dispose()
```

Thanks :)

jamesmyose:
Thank you for the feedback! Those changes should be in now.

src/Layer/Layer.js
```diff
@@ -239,8 +239,16 @@ class Layer extends THREE.EventDispatcher {
      * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
      */
     // eslint-disable-next-line
-    delete(clearCache) {
-        console.warn('Function delete doesn\'t exist for this layer');
+    dispose(clearCache) {
+        if (clearCache) {
```
jailln: Thanks! Now that you have factorized it in the Layer, you can remove this part from the other layers `dispose` methods and call `super.dispose();` instead.

jamesmyose: Alright, gotcha. Hopefully that fixes it.

gchoqueux:
@jamesmyose have you time to finish this PR?

jamesmyose:
> @jamesmyose have you time to finish this PR?

Sorry, it's been a while since I've worked on this. I don't remember what issues I had left. Would you be able to help me figure out?

ftoromanoff:
Hello @jamesmyose ,
You don't have much to do left. You just need to clean up the number of commit in this PR.
For instance squashing your commits saying 'last commit didn't work' with the previous ones
as well ad the commits that seems to have the same name.

If you don't really understand we can try to discuss about that or we can also do it for you if you are too far away from that

src/Core/Scheduler/Cache.js
```diff
@@ -192,20 +192,30 @@ class Cache {
         }
     }
 
+    delete(key1, key2, key3) {
+        console.warn('`Cached.delete` method is deprecated. Please use `Cache.remove` instead.');
```
jailln: small typo: `Cached` -> `Cache`

src/Layer/Layer.js
```diff
@@ -243,8 +243,16 @@ class Layer extends THREE.EventDispatcher {
...
+            super.dispose();
```
jailln: don't call `super.dispose()` here but actually dispose the cache by calling `this.cache.dispose()`

src/Layer/GeometryLayer.js
```diff
@@ -171,9 +171,9 @@ class GeometryLayer extends Layer {
      * All layer's 3D objects are removed from the scene and disposed from the video device.
      * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
      */
-    delete(clearCache) {
+    dispose(clearCache) {
         if (clearCache) {
```
jailln: you can remove the `if (clearCache)` since it is done in the `Layer` now (same for the other layers, except `Layer.js`)

src/Layer/OrientedImageLayer.js
```diff
@@ -210,13 +210,13 @@ class OrientedImageLayer extends GeometryLayer {
...
     * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
     */
-    delete(clearCache) {
+    dispose(clearCache) {
         if (this.background) {
             // only delete geometries if it has some
-            super.delete();
+            super.dispose();
         }
         if (clearCache) {
```
jailln: you can remove this if since the clearCache test is done in `Layer.js` and since `super.dispose()` is done in if `this.background` above.

jamesmyose:
Hey @gchoqueux, @jailln, I just pushed some changes to get those needs in. I'm also calling super.dispose with clearCache. Is there anything else needed or any adaptations that I need to make in response to recent changes? I'm sorry about all this, I feel a bit too inexperienced at the moment.

ftoromanoff:
Hello jamesmyose !
Thank you to take the time to come back to this PR. We had a look at it and the history of the commit is pretty messed up and it will be hard to be able to merge your changes from this PR for several reasons : you are directly working on your master branch and becasue of the delay there is also some change on our master branch.
As your work and your implication is still important for us, can i advice you to create a new branch on your repository (by typing ```git branch <newBranchName>``` (or  ```git checkout -b <newBranchName>``` doing the same thing)
then you will have a branch with your current change, thus you will be able to update your master branch to the current state of the itowns master branch. And then you will have to do a new branch form that new master on wich you willhave to do all the final changes you did on your old branch before asking for a new merge.
Thanks in advance and you can come to me for another/better explaination if you need it !

jamesmyose:
> Hello jamesmyose ! Thank you to take the time to come back to this PR. We had a look at it and the history of the commit is pretty messed up and it will be hard to be able to merge your changes from this PR for several reasons : you are directly working on your master branch and becasue of the delay there is also some change on our master branch. As your work and your implication is still important for us, can i advice you to create a new branch on your repository (by typing `git branch <newBranchName>` (or `git checkout -b <newBranchName>` doing the same thing) then you will have a branch with your current change, thus you will be able to update your master branch to the current state of the itowns master branch. And then you will have to do a new branch form that new master on wich you willhave to do all the final changes you did on your old branch before asking for a new merge. Thanks in advance and you can come to me for another/better explaination if you need it !

Hey @ftoromanoff, thank you for helping me out. Is the intention to create a new pull request with different commits than what is here? I've created a new branch off of my master branch and I synced my master based on itowns master, but I didn't encounter any merge conflicts so my master still has my commits. I can rewrite the commit history to be in line with itowns master, I've been told that it's generally better practice to create a new commit that reverts the older ones.

ftoromanoff:
Currently your master might not have conflict with the itowns master, but there is a lot of commits that you did, at different moment in your master time line, what makes it extremly hard to put all your commit together for a clear vision of what your modifications are. To be able to merge what you did correctly, we need to be able to group your commits all together.
Your master must be the exact same as the itowns master and not a itowns.master + extra commits spread around.

In a master branch it's indeed adviced to create a new commits that reverts the older ones, and it's the best practice. But wha tyou want ot do here is not to recreate a itowns master branch but to create a branch that you want to merge in the itowns master. Thus our master, is in way a branch that duplicate the itowns master, for you to be able to add branches as if they were directly added in the itowns repository even if they are not.
That the reason why you should try to have your master as an exact copy of the itowns master branch and another branch with your modifications on which you can do a git rebase -i origin/master (to mimic a git rebase -i itowns/master)

PR #1943, opened 2022-11-03
fix (wfs, wms, wmts): support urls ending with or without ? character

jailln:
Support WMTS urls that already end with '?'.


mgermerie:
Maybe we could add the modification in `WMSSource` and `WFSSource` as well ?

jailln:
Yep good idea! Done :)

mgermerie (approved):

PR #1958, opened 2022-11-25
3D Tiles Cesion ion support

jailln:
## Description
* Add support for 3D Tiles from Cesium ion server (`C3DTilesIonSource`) (+ tests)
* Add an example displaying OSM extruded buildings from a Cesium ion server (+ tests)
* Rename and reorganize 3D tiles examples
* Add doc and tests on `C3DTilesSource`

### DIFF
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -22,7 +22,7 @@ class C3DTilesLayer extends GeometryLayer {
      * }, view);
      * View.prototype.addLayer.call(view, l3dt);
      *
-     * // Create a new 3d-tiles layer from a Cesion ion server
+     * // Create a new 3d-tiles layer from a Cesium ion server
      * const l3dt = new C3DTilesLayer('3dtiles', {
      *      name: '3dtl',
      *      source: new C3DTilesIonSource({
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -13,11 +13,21 @@ class C3DTilesLayer extends GeometryLayer {
      * @extends GeometryLayer
      *
      * @example
-     * // Create a new Layer 3d-tiles For DiscreteLOD
+     * // Create a new 3d-tiles layer from a web server
      * const l3dt = new C3DTilesLayer('3dtiles', {
      *      name: '3dtl',
      *      source: new C3DTilesSource({
      *           url: 'https://tileset.json'
+     *      })
+     * }, view);
+     * View.prototype.addLayer.call(view, l3dt);
+     *
+     * // Create a new 3d-tiles layer from a Cesion ion server
+     * const l3dt = new C3DTilesLayer('3dtiles', {
+     *      name: '3dtl',
+     *      source: new C3DTilesIonSource({
+     *              accessToken: 'myAccessToken',
+                    assetId: 12
      *      })
      * }, view);
      * View.prototype.addLayer.call(view, l3dt);
--- a/src/Source/C3DTilesIonSource.js
+++ b/src/Source/C3DTilesIonSource.js
@@ -1,0 +1,60 @@
+import Fetcher from 'Provider/Fetcher';
+import C3DTilesSource from './C3DTilesSource';
+
+/**
+ * @classdesc
+ * An object defining the source connection to a 3DTiles asset of a [Cesium ion server](https://cesium.com/learn/ion/).
+ *
+ * @extends Source
+ *
+ * @property {boolean} isC3DTilesIonSource - Used to checkout whether this source is a C3DTilesIonSource. Default is
+ * true. You should not change this, as it is used internally for optimisation.
+ * @property {string} url - The URL of the tileset json.
+ * @property {string} baseUrl - The base URL to access tiles.
+ * @property {string} accessToken - The Cesium ion access token used to retrieve the resource.
+ * @property {string} assetId - The id of the asset on Cesium ion.
+ */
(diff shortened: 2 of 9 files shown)


src/Layer/C3DTilesLayer.js
```diff
@@ -22,6 +22,16 @@ class C3DTilesLayer extends GeometryLayer {
      * }, view);
      * View.prototype.addLayer.call(view, l3dt);
      *
+     * // Create a new 3d-tiles layer from a Cesion ion server
```
mgermerie: Little typo on Cesium here

src/Source/C3DTilesIonSource.js
```diff
@@ -0,0 +1,51 @@
+import Source from 'Source/Source';
+import Fetcher from 'Provider/Fetcher';
+
+/**
+ * @classdesc
+ * An object defining the source connection to a 3DTiles asset of a [Cesium ion server](https://cesium.com/learn/ion/).
+ *
+ * @extends Source
+ *
+ * @property {boolean} isC3DTilesIonSource - Used to checkout whether this source is a C3DTilesIonSource. Default is
```
mgermerie, line 10: I have a remark which is not really in the scope of this PR : maybe we could refactor all `isSomethingLayer` with something like the following.

```js
let isSomethingLayer = true;
Object.defineProperty(this, 'isSomethingLayer', {
    get: () => isSomethingLayer,
    set: () => { console.error('Some error log'); },
});
```
What do you think of it ?

src/Source/C3DTilesIonSource.js
```diff
@@ -0,0 +1,51 @@
...
+ * true. You should not change this, as it is used internally for optimisation.
+ * @property {string} url - The URL of the tileset json.
+ * @property {string} baseUrl - The base URL to access tiles.
+ * @property {string} accessToken - The Cesium ion access token used to retrieve the resource.
+ * @property {string} assetId - The id of the asset on Cesium ion.
+ */
+class C3DTilesIonSource extends Source {
+    constructor(source) {
```
mgermerie, line 27: Reading the properties lists does not tell user that only `accessToken` and `assetId` are to be passed as parameters of the `constructor`. Maybe we could add doc on which parameters the `constructor` takes ?

src/Source/C3DTilesSource.js
```diff
@@ -1,9 +1,21 @@
...
+ *
+ * @extends Source
+ *
+ * @property {boolean} isC3DTilesSource - Used to checkout whether this source is a isC3DTilesSource. Default is
+ * true. You should not change this, as it is used internally for optimisation.
+ * @property {string} url - The URL of the tileset json.
+ * @property {string} baseUrl - The base URL to access tiles.
+ */
 class C3DTilesSource extends Source {
     constructor(source) {
```
mgermerie, line 25: Here again, could we add doc on the `constructor`'s parameters ?

src/Source/C3DTilesIonSource.js
jailln: Yep good idea!
I think we should indeed do it for all layers in another issue/PR

src/Layer/C3DTilesLayer.js
jailln: done

src/Source/C3DTilesIonSource.js
jailln: yep you are right, I added more doc.

src/Source/C3DTilesIonSource.js
mgermerie: I opened #1980 on the matter.

src/Source/C3DTilesSource.js
jailln: yes, done too.

src/Source/C3DTilesIonSource.js
mgermerie: If I understood correctly, `url` seems not to be a mandatory parameter for this constructor, since it is overwritten on line 36 before being passed to `Source` constructor.

jailln: my bad I went to fast on updating the doc. 

Access to a tileset on Cesium ion is done in two passes:
* one query to the server with the asset id and the access token that returns among other formation the actual url of the tileset
* a second query to actually get the tileset

Therefore the URL of the first query is added as source url at line 36 (since it is mandatory for `Source` constructor) and then replaced by the URL of the actual resource at line 50.

gchoqueux (commented):
I wonder if a new `source` is needed, what you think @mgermerie a,d @jailln ?
If you keep `C3DTilesIonSource` could you extend from  `C3DTilesSource` ?

mgermerie:
> I wonder if a new `source` is needed, what you think @mgermerie a,d @jailln ? If you keep `C3DTilesIonSource` could you extend from `C3DTilesSource` ?

Do you have something in mind other than a new `source` ? Adding options to `C3DTilesSource` to specify whether data come from a Cesium Ion server or from another server ?

jailln:
> I wonder if a new `source` is needed, what you think @mgermerie a,d @jailln ? 

In my opinion it makes the API and the code clearer (since the way to access the data is different between a simple HTTP server and a Cesium ion Server), so I would say yes.

> If you keep `C3DTilesIonSource` could you extend from `C3DTilesSource` ?

Yes, I hesitated between extending from `Source` and `C3DTilesSource` but I think you're right, it's clearer from `C3DTilesSource`.

jailln:
Updated and rebased :slightly_smiling_face:

gchoqueux (approved):
Thanks!

PR #1965, opened 2022-12-08
fix(view): fix view resize when width or height is 0

jailln:
## Description
Don't resize the camera when the div containing the view has a width or height of 0 and improve parameters verification in `View.resize` and `Camera.resize`.

## Motivation and Context

In a context where itowns was in a div that was hidden (by setting the css property of the div to `display:none`) and then displayed again, the rendering was broken (black screen) if we resized the scene when the div was hidden. This was due to division by 0 in `Camera.resize`. I removed the Camera resize in this case since nothing is rendered when the div holding the view is hidden anyway.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -1088,6 +1088,11 @@ class View extends THREE.EventDispatcher {
      * the viewer with. By default it is the `clientHeight` of the `viewerDiv`.
      */
     resize(width, height) {
+        if (width < 0 || height < 0) {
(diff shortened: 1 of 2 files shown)


src/Core/View.js
```diff
@@ -1088,6 +1088,11 @@ class View extends THREE.EventDispatcher {
      * the viewer with. By default it is the `clientHeight` of the `viewerDiv`.
      */
     resize(width, height) {
+        if (width < 0 || height < 0) {
```
mgermerie, line 1091: Good idea to check for negative values !

mgermerie (approved):

PR #1971, opened 2022-12-15
fix(controls): keyboard events are now on the domElement

### LINKED ISSUE #1931: iTowns view traps keyboard events even if the view div is not focused
## Your Environment
* Version used: v2.38.2
* Browser Name and version: firefox and chrome
* Operating System and version (desktop or mobile): ubuntu, windows
* Link to your project: x

## Context

When using itowns in a window with other elements (a form in this case), I noticed that using some keyboard keys such as arrows, ctrl and shift moves itowns view even if the div containing the view is not focused. 

1. Create a page with an input and a div containing itowns view
2. use ctrl + arrows in the input

## Expected Behavior
The view doesn't move since the user is focusing the input.

## Actual Behavior
The view actually moves

This happens because itowns traps events from the window and not from the viewer div, originally set like that for a specific usecase where it was more comfortable to always have itowns view focused. In my opinion itowns shouldn't watch keyboards events on the window because it reduces its possible integration in other softwares. I propose to change this behaviour and to watch events on the viewer div. @mgermerie what's your opinion on that?

jailln:
Stop listening for keyboard events on `window` (fixes #1931 )

### DIFF
--- a/src/Controls/StateControl.js
+++ b/src/Controls/StateControl.js
@@ -214,15 +214,11 @@ class StateControl extends THREE.EventDispatcher {
 
         this._domElement.addEventListener('pointerdown', this._onPointerDown, false);
         this._domElement.addEventListener('wheel', this._onMouseWheel, false);
-
-        // The event listener is added on `window` so that key input can be accounted event if the view does not have
-        // the focus. This can occur at page loading, when a mini map is displayed : the minimap initially has the focus
-        // and key input would not be considered on the view's domElement.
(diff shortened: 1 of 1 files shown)


mgermerie (approved):
Thanks ! Having keyboard events triggered on `window` was, amongst other things, a workaround for an issue with `GuiTools` : using the menu from the plugin looses the focus on the view's `domElement`. However, this should be fixed differently from listening to keyboard events on `window`.
I opened a new [issue](https://github.com/iTowns/itowns/issues/1978) on that matter.

PR #1974, opened 2023-01-03
Remove Three.js layers using

### LINKED ISSUE #1863: FeatureGeometryLayer doesn't seem to be remove form the list of layer after few iteration
I am actually working with this team https://github.com/VCityTeam where some of our features are based on itowns. In my research projects, I am working on ABM simulation and I am now trying to implement a visualizator in VCITY/Itowns (see this issue https://github.com/VCityTeam/UD_ReAgent_ABM/issues/11).

The idea is to get a geojson file from a server ( in this case a simulation run by [Gama Platform](https://gama-platform.org/)) updating the geojson at each step of the iteration. I got some first results using a ```FeatureGeometryLayer```that I'd like to update each frame.

To do so I create a ```FeatureGeometryLayer``` first and then to update it I erase it and the recreate it and readd it in the itowns view. 

However I get the same error described in this issue after 31 iteration I get this error message

`````
Too much three.js layers. Starting from now all of them will use layerMask = 31
`````
This is what I have in my SetInterval function
````
if (added) {
    app.view.removeLayer("GAMA");
    view.notifyChange();
}  
added = 1;

_source = new itowns.FileSource({
  fetchedData: geojson,
  crs: 'EPSG:3946',
  format: 'application/json',
});

gama_layer = new itowns.FeatureGeometryLayer('GAMA', {
  source: _source,
  transparent: true,
  opacity: 1,
  style: new itowns.Style({
    fill: {
         color: 'red' ,
    }
  })
});

  
app.view.addLayer(gama_layer);
````

This seems somehow related to this issue https://github.com/iTowns/itowns/issues/1283 however the fix proposed seems not to work in my case. 

FYI we are defining this in our ```package.json```   ```"itowns": "2.37.0"```

### PR DESCRIPTION
**Pull request by gchoqueux**
## Description
- Remove every usage of [`Three.js layers`](https://threejs.org/docs/?q=layer#api/en/core/Layers).
- Replace the control `GeometryLayer` visibility, handled by `Three.js layers`, by `GeometryLayer.object3D` visibility.
- Use [GeometryLayer](http://www.itowns-project.org/itowns/docs/#api/Layer/GeometryLayer) to display tile debug options ;

## Motivation and Context
- Code simplification, in particular `picking`.
- ThreeJS Layers usage can be problematic in some case as explained in [this issue](https://github.com/iTowns/itowns/issues/1863) #1863
- Remove conflict with `Three.WebXRManager`.

Closes #1863 #1869

### DIFF
--- a/src/Layer/GeometryLayer.js
+++ b/src/Layer/GeometryLayer.js
@@ -56,6 +56,11 @@ class GeometryLayer extends Layer {
      */
     constructor(id, object3d, config = {}) {
         config.cacheLifeTime = config.cacheLifeTime ?? CACHE_POLICIES.GEOMETRY;
+
+        // Remove this part when Object.assign(this, config) will be removed from Layer Constructor
+        const visible = config.visible;
+        delete config.visible;
+
         super(id, config);
 
         this.isGeometryLayer = true;
@@ -79,7 +84,7 @@ class GeometryLayer extends Layer {
         this.wireframe = false;
 
         this.attachedLayers = [];
-        this.visible = config.visible == undefined ? true : config.visible;
+        this.visible = visible ?? true;
         Object.defineProperty(this.zoom, 'max', {
             value: Infinity,
             writable: false,
@@ -88,6 +93,20 @@ class GeometryLayer extends Layer {
         // Feature options
         this.filteringExtent = !this.source.isFileSource;
         this.structure = '3d';
+    }
+
+    get visible() {
+        return this.object3d.visible;
+    }
+
+    set visible(value) {
+        if (this.object3d.visible !== value) {
+            const event = { type: 'visible-property-changed', previous: {}, new: {} };
+            event.previous.visible = this.object3d.visible;
+            event.new.visible = value;
+            this.dispatchEvent(event);
+            this.object3d.visible = value;
+        }
     }
 
     // Attached layers expect to receive the visual representation of a
@@ -182,7 +201,7 @@ class GeometryLayer extends Layer {
      * specified coordinates.
      */
     pickObjectsAt(view, coordinates, radius = this.options.defaultPickingRadius, target = []) {
-        return Picking.pickObjectsAt(view, coordinates, radius, this.object3d, target, this.threejsLayer);
+        return Picking.pickObjectsAt(view, coordinates, radius, this.object3d, target);
     }
 }
 
--- a/src/Process/ObjectRemovalHelper.js
+++ b/src/Process/ObjectRemovalHelper.js
@@ -82,9 +82,13 @@ export default {
     removeChildrenAndCleanupRecursively(layer, obj) {
         // Objects are filtered by id because the obj hierarchy may also contain labels that have been added as childs
         // of the objects which have their own removal logic
-        let toRemove = obj.children.filter(c => (c.layer && c.layer.id) === layer.id);
+        let toRemove = obj.children.filter(c => c.layer && c.layer.id === layer.id);
         if (obj.link) {
-            toRemove = toRemove.concat(obj.link);
+            const linkedObjects = obj.link.filter(c => c.layer && c.layer.id === layer.id);
+            if (linkedObjects.length) {
+                toRemove = toRemove.concat(linkedObjects);
+                obj.link = obj.link.filter(c => c.layer && c.layer.id !== layer.id);
+            }
         }
         for (const c of toRemove) {
             this.removeChildrenAndCleanupRecursively(layer, c);
--- a/utils/debug/TileDebug.js
+++ b/utils/debug/TileDebug.js
@@ -1,6 +1,5 @@
 import * as THREE from 'three';
 import TWEEN from '@tweenjs/tween.js';
-import View from 'Core/View';
 import GeometryLayer from 'Layer/GeometryLayer';
 import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
 import ObjectRemovalHelper from 'Process/ObjectRemovalHelper';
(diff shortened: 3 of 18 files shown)

### CODE REVIEW

File: src/Layer/GeometryLayer.js [hunk 1]
```diff
@@ -104,7 +109,7 @@ class GeometryLayer extends Layer {
         this.wireframe = false;
 
         this.attachedLayers = [];
-        this.visible = config.visible == undefined ? true : config.visible;
+        this.visible = visible == undefined ? true : visible;
```
mgermerie: I wonder if the following line would allow to get rid of lines 86 and 87 ?
```js
this.visible = this.visible ?? true;
```

mgermerie: Actually, is there a point in setting this property since the getter and setter of `visible` property both use `this.object3d.visible` ?

mgermerie (commented):
I spotted some remaining usage of ThreeJS layers : 

- in [`GlobeControls`](https://github.com/iTowns/itowns/blob/d5c80f4796599ada37b7e6275fae529bd880b9b3/src/Controls/GlobeControls.js#L308-L311) for the camera target helpers. They should not be needed, since I don't believe (correct me if I'm wrong) there is a option to switch camera helper visibility (other than hard coding [this property](https://github.com/iTowns/itowns/blob/d5c80f4796599ada37b7e6275fae529bd880b9b3/src/Controls/GlobeControls.js#L81) to `true`).
- In `c3DEngine`, [this method](https://github.com/iTowns/itowns/blob/d5c80f4796599ada37b7e6275fae529bd880b9b3/src/Renderer/c3DEngine.js#L245) should no longer be required, other from the [collada example](https://github.com/iTowns/itowns/blob/d5c80f4796599ada37b7e6275fae529bd880b9b3/examples/misc_collada.html#L73).

src/Process/ObjectRemovalHelper.js
```diff
@@ -68,9 +68,13 @@ export default {
      * @return {Array} an array of removed Object3D from obj (not including the recursive removals)
      */
     removeChildrenAndCleanupRecursively(layer, obj) {
-        let toRemove = obj.children.filter(c => (c.layer && c.layer.id) === layer.id);
+        let toRemove = obj.children.filter(c => c.layer && c.layer.id === layer.id);
```
jailln, line 85: Shouldn't all this filtering logic on layers be removed since we don't set layers anymore?

utils/debug/TileDebug.js
```diff
@@ -117,101 +116,81 @@ export default function createTileDebugUI(datDebugTool, view, layer, debugInstan
...
-        if (!node.parent || !enabled) {
+        if (!node.parent || !layer.visible) {
             ObjectRemovalHelper.removeChildrenAndCleanupRecursively(layer, node);
             return;
         }
 
-        let helper = node.children.filter(n => n.layer && (n.layer.id == layer.id))[0];
-
-        if (node.material && node.material.visible) {
+        let helper = node.link.filter(n => n.layer && (n.layer.id == layer.id))[0];
```
jailln, line 125: The layer logic is kept for debug objects only then? If yes, could you add a small comment explaining that we use threejs layers and why please?

src/Process/ObjectRemovalHelper.js
jailln: I just saw the modifications in TileDebug.js so I understand why you added this logic. Could you move it to a dedicated function like `getLinkedDebugObjects` for instance so it is clearer why we do all this logic here?

utils/debug/TileDebug.js
```diff
@@ -117,101 +116,81 @@ export default function createTileDebugUI(datDebugTool, view, layer, debugInstan
...
+            layer.object3d.add(helper);
             helper.updateMatrixWorld(true);
-            helper.visible = true;
         } else if (helper) {
-            helper.visible = false;
+            layer.object3d.remove(helper);
+        }
+    }
+
+    class DebugLayer extends GeometryLayer {
```
jailln, line 157: I think we could create it in a separate file and use it in 3DTilesDebug.js too, WDYT?

src/Layer/GeometryLayer.js
gchoqueux: It's more intuitive to use `Layer.visible` and more similar to other APIs.
`this.object3d.visible` is like a private property and `setter` is used for emit event.

src/Process/ObjectRemovalHelper.js
gchoqueux: > Shouldn't all this filtering logic on layers be removed since we don't set layers anymore?

yes, it seems there's cleaning to do
i'll try to remove some parts of this file

jailln: Yes, but mind [this PR](https://github.com/iTowns/itowns/pull/1886) that modifies this file too. Maybe we should merge it and you should rebase before cleaning?

gchoqueux: > I just saw the modifications in TileDebug.js so I understand why you added this logic. Could you move it to a dedicated function like `getLinkedDebugObjects` for instance so it is clearer why we do all this logic here?

what part of the code would you like to move into this function?

gchoqueux: > Yes, but mind [this PR](https://github.com/iTowns/itowns/pull/1886) that modifies this file too. Maybe we should merge it and you should rebase before cleaning?

yes, I'll rebase this PR

utils/debug/TileDebug.js
gchoqueux: this logic is used for all attached geometry layers, by example `FeatureGeometryLayer`

utils/debug/TileDebug.js
gchoqueux: I will do it in a function to refactor the debug

src/Layer/GeometryLayer.js
gchoqueux: > ```js
> this.visible = this.visible ?? true;
> ```

small typo 
```js
this.visible = visible ?? true;
```

src/Process/ObjectRemovalHelper.js
jailln: > what part of the code would you like to move into this function?

My comment is misplaced sorry, I meant the part just below (from line 72 to 77) could be moved to a separate function.

utils/debug/TileDebug.js
jailln: Ah ok! Could you add a comment please so it is easier to remember why we need this logic?

utils/debug/TileDebug.js
gchoqueux: Sorry,  I'll open a new PR to refactor the debug

jailln: ok :+1:

src/Layer/GeometryLayer.js
mgermerie: > It's more intuitive to use `Layer.visible` and more similar to other APIs. `this.object3d.visible` is like a private property and `setter` is used for emit event.

Yep, I get that, but my point was that line 112 could be removed, since the getter and setter are set for `visible` property, and there is no usage of `this.visible`.




> > ```js
> > this.visible = this.visible ?? true;
> > ```
> 
> small typo
> 
> ```js
> this.visible = visible ?? true;
> ```

Indeed sorry, well spotted

src/Process/ObjectRemovalHelper.js
gchoqueux: I thought about this `ObjectRemovalHelper.js` and I would like move all nodes in their layer. So, it would be  possible to remove `node.layer.id === layer.id` in this file. (In same case, if `links` are grouped by `layer` it wouldn't have this test)

However, `LabelLayer` needs to be extend by `FeatureGeometryLayer`. In bonus we would simplify  label rendering!

`OBB` box must also be removed in tile's children.

All these changes need a new PR

utils/debug/TileDebug.js
gchoqueux: yes

src/Process/ObjectRemovalHelper.js
jailln: Yes good idea! Do you think you will be able to it soon and remember it or should we open an issue about it ?

gchoqueux:
> in `GlobeControls` for the camera target helpers. 
> They should not be needed, since I don't believe (correct me if I'm wrong) 
> there is a option to switch camera helper visibility

Fixed

>  In `c3DEngine`, this method should no longer be required, other from the collada example.

This method is removed

src/Process/ObjectRemovalHelper.js
gchoqueux: I opened #1991

jailln (approved):

PR #1989, opened 2023-01-25
refactor(OBB): remove OBB from node's children.

gchoqueux:
## Description
Remove tile's `OBB` (oriented bounding box)

## Motivation and Context
Simplify the data structure and in future simplify the clean processing

see #1991

### DIFF
--- a/src/Converter/convertToTile.js
+++ b/src/Converter/convertToTile.js
@@ -71,8 +71,6 @@ export default {
             tile.visible = false;
             tile.updateMatrix();
 
-            tile.add(tile.obb);
-
             setTileFromTiledLayer(tile, layer);
 
             if (parent) {
--- a/src/Layer/TiledGeometryLayer.js
+++ b/src/Layer/TiledGeometryLayer.js
@@ -272,7 +272,7 @@ class TiledGeometryLayer extends GeometryLayer {
 
     // eslint-disable-next-line
     culling(node, camera) {
-        return !camera.isBox3Visible(node.obb.box3D, node.obb.matrixWorld);
+        return !camera.isBox3Visible(node.obb.box3D, node.matrixWorld);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


jailln:
Thanks, that will simplify the object hierarchy! What was it used for ?

We may be able to do the same for 3D Tiles, WDYT?

gchoqueux:
> Thanks, that will simplify the object hierarchy! What was it used for ?
> 
> We may be able to do the same for 3D Tiles, WDYT?

OBBs are used for culling

The scope of this PR is to simplify the node structure and remove the presence of nodes from other layers in `Object3d.children` see  #1991.

In 3d tiles structure, the bounding volume is `tile.boudingVolume`, no?

jailln:
> OBBs are used for culling
> 
> The scope of this PR is to simplify the node structure and remove the presence of nodes from other layers in `Object3d.children` see #1991.

Yes I understood that but it seems that OBBs where stored twice: in the node (in `node.obb`) and also [as a child of the node](https://github.com/iTowns/itowns/pull/1989/files#diff-64f79d232bad66d53493f8dd4dca87745216ea36a76e6e323f664e4aec29cc72L78). And from what I see, it was `node.obb` that was used for [culling](https://github.com/iTowns/itowns/pull/1989/files#diff-47ce599a07f1b72478616419353cba9fb27a617390bfc90d469822070d439006L275) and not the child OBB. So my question is more: what was the use of the children obb? (I think the answer is that is wasn't use anywhere anymore, but I just want to be sure)

> In 3d tiles structure, the bounding volume is `tile.boudingVolume`, no?

Yes but we also add helpers in `3dTilesDebug.js` that are used to display the bounding boxes. I think we should follow the same logic and move these helpers to another layer too, but it can be done in another PR.

jailln (approved):

mgermerie (approved):

Issue #1981, opened 2023-01-09
Bathymetry not showing 


AdrienHoule:
I'm trying to show the world bathymetry on the globe using gebco wms :

   const wmsSource = new itowns.WMSSource({
                url: 'https://www.gebco.net/data_and_products/gebco_web_services/web_map_service/mapserv',
						version: '1.3.0',
						name: 'GEBCO_LATEST_3',
						style: '',
						crs: 'EPSG:4326',
						extent: {
							west: '-180',
							east: '360',
							south: '-90',
							north: '90',
						},
						transparent: true,
					});
	
But everything appears flat. 
![cas1](https://user-images.githubusercontent.com/120030087/211355112-dcca15a9-e321-4928-a3a4-098fe21ad471.png)

The only way I can view the data is to up the scale.
![cas2](https://user-images.githubusercontent.com/120030087/211355135-bb17e043-df3b-4579-a403-f1b8fce6e8f0.png)

So my question was do you know why this is happening and is there any solution to show my bathymetry without touching the scale ?

mgermerie:
Hi ! We currently have an issue on negative elevation display : all negative elevation values are clamped above -10 meters. This is explained in #1945. I will provide a fix for that in a short time.
However, since the clamp value is constant, I would intuitively say that increasing the scale should not make elevation data bellow -10 meters visible. I will take a closer look at scale implementation for `ElevationLayer` and let you know of my findings.

AdrienHoule:
Hi, I would like to know when you say short time, do you have any idea of when a fix for https://github.com/iTowns/itowns/issues/1945 will be available ?
Also I would like to offer my help to fix this issues, is there any branch I could begin with and anything to know ?
If you need to contact me my email is adrien.houle@geomatys.com.
Thanks in advance for your return.

mgermerie:
Hi ! You are most welcome to contribute to solve #1945, thanks for your proposition. There is no branch at the moment, so you could start from iTowns' master. I will send you an email to try to set up a meeting if you agree.
Otherwise, I should be able to provide a fix by the end of this month (probably sooner though).

PR #1993, opened 2023-01-27
Bounding box: calculation of MinMax values and add clamping options

### LINKED ISSUE #1945: Wrong bounding box computing for negative elevation data
When displaying some negative elevation data in an `ElevationLayer`, the bounding boxes of tiles are clamped above `-10` meters.
This is caused by [this line](https://github.com/iTowns/itowns/blob/master/src/Parser/XbilParser.js#L53) when computing the minimal and maximal value of elevation data within a given tile.

This value has been hard coded to prevent an issue with geoportail elevation data. Along the coast line, these data contain isolated points with extremely low altitude (sometimes lower than the earth radius). The bounding sphere of the tiles is therefore huge, exceeding the `sse` threshold above which tiles are subdivided. It causes tiles along the coast line to be subdivided too much. This can be seen in the screenshot bellow.

Capture d’écran du 2022-11-08 17-33-27

Clamping the minimal elevation value of a tile prevents this. However, relevant negative elevation values should not be clamped.

The `-10` hard coded value should be parametrized in the `ElevationLayer` creation, with a default value of `-Infinity`.

ftoromanoff:
## Description
I changed the way the NodataValue is taken into account for the computation of the MinMax for the bounding box and also added an option to add clamping min and max value when creating a new layer.

## Motivation and Context
There was a probleme with the segmentation of tile in relation with abnormal values in particular on see area.
issue #1945 
It should also solve the problem seen on the issue #1981

### DIFF
--- a/examples/3dtiles_ion.html
+++ b/examples/3dtiles_ion.html
@@ -77,8 +77,8 @@
                     new itowns.ElevationLayer(config.id, config),
                 );
             }
-            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
+            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
 
             // ---------- ADD 3D TILES MODEL FROM CESIUM ION SERVER : ----------
 
--- a/examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
+++ b/examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
@@ -1,6 +1,9 @@
 {
 	"id":         "IGN_MNT_HIGHRES",
 	"noDataValue" : -99999,
+	"clampValues": {
+        "min": 0
+    },
 	"updateStrategy": {
 		"type": 1,
 		"options": {
--- a/examples/view_3d_map.html
+++ b/examples/view_3d_map.html
@@ -71,9 +71,8 @@
                     new itowns.ElevationLayer(config.id, config),
                 ).then(debugMenu.addLayerGUI.bind(debugMenu));
             }
-            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
-
+            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
 
 
             // ---------- ADD SOME WIDGETS : ----------
--- a/examples/view_3d_mns_map.html
+++ b/examples/view_3d_mns_map.html
@@ -72,8 +72,6 @@
                     new itowns.ElevationLayer(config.id, config),
                 ).then(debugMenu.addLayerGUI.bind(debugMenu));
             }
-
-            itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNS_HIGHRES.json').then(addElevationLayerFromConfig);
 
             // ---------- ADD SOME WIDGETS : ----------
--- a/src/Layer/ElevationLayer.js
+++ b/src/Layer/ElevationLayer.js
@@ -7,6 +7,8 @@ import { RasterElevationTile } from 'Renderer/RasterTile';
  * an ElevationLayer. Default is true. You should not change this, as it is used
  * internally for optimisation.
  * @property {number} noDataValue - Used to specify a **null** or **no data value** in the elevation terrain.
+ * @property {number} [zmin] - Used to specify a minimum value for the elevation terrain (if the data goes lower, it will be clamped).
+ * @property {number} [zmax] - Used to specify a maximum value for the elevation terrain (if the data goes higher, it will be clamped)
  * @property {number} scale - Used to apply a scale on the elevation value. It
  * can be used for exageration of the elevation, like in [this
  * example](https://www.itowns-project.org/itowns/examples/#plugins_pyramidal_tiff).
@@ -37,6 +39,11 @@ class ElevationLayer extends RasterLayer {
      * contains three elements `name, protocol, extent`, these elements will be
      * available using `layer.name` or something else depending on the property
      * name.
+     * @param {number} [config.noDataValue]   The value coding the noData in the data set
+     * @param {Object} [config.clampValues] - Optional information for clamping
+     * the elevation between a minimum and a maximum value
+     * @param {number} [config.clampValues.min]   The minimum value to clamp the elevation
+     * @param {number} [config.clampValues.max]   The maximum value to clamp the elevation
      *
      * @example
      * // Create an ElevationLayer
@@ -54,6 +61,11 @@ class ElevationLayer extends RasterLayer {
      */
     constructor(id, config = {}) {
         super(id, config);
+        if (config.zmin || config.zmax) {
+            console.warn('Config using zmin and zmax are deprecated, use {clampValues: {min, max}} structure.');
+        }
+        this.zmin = config.clampValues?.min ?? config.zmin;
+        this.zmax = config.clampValues?.max ?? config.zmax;
         this.isElevationLayer = true;
         this.defineLayerProperty('scale', this.scale || 1.0);
     }
--- a/src/Parser/XbilParser.js
+++ b/src/Parser/XbilParser.js
@@ -1,30 +1,38 @@
 import { readTextureValueWithBilinearFiltering } from 'Utils/DEMUtils';
 
-function minMax4Corners(texture, pitch, noDataValue) {
+function minMax4Corners(texture, pitch, options) {
     const u = pitch.x;
     const v = pitch.y;
     const w = pitch.z;
     const z = [
-        readTextureValueWithBilinearFiltering({ noDataValue }, texture, u, v),
-        readTextureValueWithBilinearFiltering({ noDataValue }, texture, u + w, v),
-        readTextureValueWithBilinearFiltering({ noDataValue }, texture, u + w, v + w),
-        readTextureValueWithBilinearFiltering({ noDataValue }, texture, u, v + w),
-    ].filter(v => v != undefined && v > -10);
+        readTextureValueWithBilinearFiltering(options, texture, u, v),
+        readTextureValueWithBilinearFiltering(options, texture, u + w, v),
+        readTextureValueWithBilinearFiltering(options, texture, u + w, v + w),
+        readTextureValueWithBilinearFiltering(options, texture, u, v + w),
+    ].filter(val => val != undefined);
 
     if (z.length) {
         return { min: Math.min(...z), max: Math.max(...z) };
+    } else {
+        return {
+            min: Infinity,
+            max: -Infinity,
+        };
     }
 }
 
 /**
-  * Calculates the minimum maximum texture elevation with xbil data
-  *
-  * @param      {THREE.Texture}  texture       The texture to parse
-  * @param      {THREE.Vector4}  pitch  The pitch,  restrict zone to parse
- * @param      {number}  noDataValue  No data value
-  * @return     {Object}  The minimum maximum elevation.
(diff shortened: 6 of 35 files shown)


examples/view_3d_map.html
```diff
@@ -71,9 +71,8 @@
                     new itowns.ElevationLayer(config.id, config),
                 ).then(debugMenu.addLayerGUI.bind(debugMenu));
             }
-            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
-
+            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
```
mgermerie, line 75: This inversion is due to #2000. It should be done also on other examples with multiple DEM until #2000 is solved.

examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
```diff
@@ -1,6 +1,9 @@
 {
 	"id":         "IGN_MNT_HIGHRES",
 	"noDataValue" : -99999,
+	"clampValues": {
```
jailln, line 4: Is there negative values in this MNT that we need to clamp ? Or is it just to show how it works?

examples/view_3d_map.html
jailln, line 75: Does the order changes anything? Is it related to #2000 ?

src/Layer/ElevationLayer.js
```diff
@@ -7,6 +7,8 @@ import { RasterElevationTile } from 'Renderer/RasterTile';
  * an ElevationLayer. Default is true. You should not change this, as it is used
  * internally for optimisation.
  * @property {number} noDataValue - Used to specify a **null** or **no data value** in the elevation terrain.
+ * @property {number} zmin - Used to specify a minimum value under which the elevation terrain will be clamped.
+ * @property {number} zmax - Used to specify a maximum value over which the elevation terrain will be clamped.
```
jailln: I'm wondering if we shouldn't rename these properties to something like `minClampValue` and `maxClampValue` to avoid confusions with actual `zmin` or `zmax` of the layer and to be closer to the `config` object passed to the constructor . WDYT @ftoromanoff @mgermerie ?

src/Layer/ElevationLayer.js
```diff
@@ -54,6 +61,8 @@ class ElevationLayer extends RasterLayer {
      */
     constructor(id, config = {}) {
         super(id, config);
+        this.zmin = config.clampValues?.min ?? config.zmin;
```
jailln, line 67: What if `config.zmin` and `config.zmax` are not set either? Also, `config.zmin` and `config.zmax` are not documented and the same as the `clampValues` ?
Shouldn't it be something like: 
```
if (config.clampValues && config.clampValue.min) {
  this.zmin = config.clampValues.min;
}
```
or like
```
this.zmin = config.clampValues?.min ?? -Infinity;
```
?

jailln (changes requested):
Thanks for this PR! Except from my comments, don't forget to rename your commit with the angular convention :)

examples/view_3d_map.html
jailln: +1 and this answer one of my questions :)

ftoromanoff: I changed it in all the other examples where the 2 elevation layers were added.

examples/layers/JSONLayers/IGN_MNT_HIGHRES.json
ftoromanoff: There is plenty of absurd data (as -99998.0235 or -21562) in the data flux (probably linked wuth a bad interpolation in the creation of these flux. That's the raison why there were a hard coded -10 in the code. So the 0 in the MNT HIGHRES is needed.

jailln: Ok, thanks for the explaination

src/Layer/ElevationLayer.js
ftoromanoff: I kept config.zmin and config.zmax in discussion with @mgermerie to limit break change if someone used them before (even if there were not really any documentation of these).
And when both are undefined, it's normally not a problem as the default value (-Infinity and Infinity) will be taken. see RasterTile.js line 150
```
        this.zmin = layer.zmin ?? defaultEle.zmin;
        this.zmax = layer.zmax ?? defaultEle.zmax;
```

ftoromanoff: About the documentation i added some line for it :
```
 * @property {number} noDataValue - Used to specify a **null** or **no data value** in the elevation terrain.
 * @property {number} zmin - Used to specify a minimum value under which the elevation terrain will be clamped.
 * @property {number} zmax - Used to specify a maximum value over which the elevation terrain will be clamped.
```
and
```
     * @param {Object} [config] - Optional configuration, all elements in it
     * will be merged as is in the layer. For example, if the configuration
     * name.
     * @param {number} [config.noDataValue]   The value coding the noData in the data set
     * @param {Object} [config.clampValues] - Optional information for clamping
     * the elevation between a minimum and a maximum value
     * @param {number} [config.clampValues.min]   The minimum value to clamp the elevation
     * @param {number} [config.clampValues.max]   The maximum value to clamp the elevation
```

src/Layer/ElevationLayer.js
ftoromanoff: For me zmin and zmax are the min and max elevation for a layer after clamping, to avoid absurd values in the data.
So it's indeed the zmin and zmax of a layer, but these can be played with throught the clampValues parameters.

src/Layer/ElevationLayer.js
jailln: Ok. In this case my opinion is that we should:
* not document them because we don't want users to use these but to use `config.clampValues.min` and `config.clampValues.max` (and we don't want to have two possibilities to configure the same way I think, it complicates the API)
* Add a deprecation warning for these config values

src/Layer/ElevationLayer.js
jailln: It is the case when the values are actually clamped but if the min clamp value is lower than the min value of the layer it is not the case for instance but this might be a detail. My remark was also to be more consistent with the configuration value. But no strong opinions on this, I will let you decide.

ftoromanoff: the clamp parameters values are only used if there is a data value lower (or higher) that the clamp value on a tile. It will not be the default value taken. (Zmin and zmax are a 'global' value for the layer, but bounding boxes will be calculated on each tile.

src/Layer/ElevationLayer.js
ftoromanoff: I removed the documentation of the layer properties zmin, zmax and noDataValue.
And i added a deprecated warning if zmin and zmax are directly used.
```
        if (config.zmin || config.zmax) {
            console.warn('Config using zmin and zmax are deprecated, use {clampValues: {min, max}} structure.');
        }
```

src/Layer/ElevationLayer.js
jailln: Yes that's what I understood too :slightly_smiling_face: let's keep these names

src/Layer/ElevationLayer.js
jailln: Ok sorry I think there was a mixup between the Layer properties documentation and the constructor documentation :sweat_smile: Could you please:
* In the layer: leave the doc about zmin, zmax and noDataValue since they are public properties
* In the constructor: leave it like it is currently (with clampValues and noDataValue documented)

ftoromanoff: Done ^^
I changed a bit the description hoping it will be clearer.

examples/view_3d_mns_map.html
```diff
@@ -72,9 +72,8 @@
                     new itowns.ElevationLayer(config.id, config),
                 ).then(debugMenu.addLayerGUI.bind(debugMenu));
             }
-
-            itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNS_HIGHRES.json').then(addElevationLayerFromConfig);
+            // itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT.json').then(addElevationLayerFromConfig);
```
jailln: Shouldn't be commented out ?

src/Layer/ElevationLayer.js
jailln: Thanks :slightly_smiling_face:

examples/view_3d_mns_map.html
ftoromanoff: There is a probleme with the dataset linked with this configuration file or with the configuration file (IGN_MNT.json), that appear with the new changes. As it is only used in this example and the extra Elevation layer doesn't really seems to be important. I prefered to removed it and it can be eluded when the issue #2000 will be studied.

jailln: Ok, could you describe this problem in #2000 please so we don't forget about it ?

jailln (approved):

examples/3dtiles_ion.html
```diff
@@ -77,8 +77,8 @@
                     new itowns.ElevationLayer(config.id, config),
                 );
             }
-            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
+            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
```
gchoqueux, line 81: it's issue #2000 ?

src/Layer/ElevationLayer.js
```diff
@@ -37,6 +39,11 @@ class ElevationLayer extends RasterLayer {
      * contains three elements `name, protocol, extent`, these elements will be
      * available using `layer.name` or something else depending on the property
      * name.
+     * @param {number} [config.noDataValue]   The value coding the noData in the data set
+     * @param {Object} [config.clampValues] - Optional information for clamping
+     * the elevation between a minimum and a maximum value
+     * @param {number} [config.clampValues.min]   The minimum value to clamp the elevation
+     * @param {number} [config.clampValues.max]   The maximum value to clamp the elevation
```
gchoqueux, line 46: you rename to make the parameters more understandable?
you keep `zmin` and `zmax` name for the layer properties?

src/Parser/XbilParser.js
```diff
@@ -1,30 +1,38 @@
...
-  */
-export function computeMinMaxElevation(texture, pitch, noDataValue) {
+ * Calculates the minimum maximum texture elevation with xbil data
+ *
+ * @param      {THREE.Texture}   texture                     The texture to parse
+ * @param      {THREE.Vector4}   pitch                       The pitch,  restrict zone to parse
+ * @param      {Object}          options                     No data value and clamp values
+ * @param      {number}          options.noDataValue         No data value
+ * @param      {number}          [options.zmin]   The minimum elevation value after which it will be clamped
+ * @param      {number}          [options.zmax]   The maximum elevation value after which it will be clamped
```
gchoqueux, line 32: could you homogenize with `clampValues` name?

src/Renderer/Shader/Chunk/elevation_pars_vertex.glsl
```diff
@@ -36,7 +37,9 @@
     float getElevation(vec2 uv, sampler2D tex, vec4 offsetScale, Layer layer) {
         uv = uv * offsetScale.zw + offsetScale.xy;
         float d = getElevationMode(uv, tex, layer.mode);
-        if (d < layer.zmin || d > layer.zmax) d = 0.;
+        if (d < layer.zmin) d = layer.zmin;
+        if (d > layer.zmax) d = layer.zmax;
+        if (d == layer.noDataValue) d = 0.;
```
gchoqueux: there's a `clamp` method in glsl

it seems to me that `d` is an interpolated value,  the filter `nodatavalue` will not work here

ftoromanoff: I changed it with the clamp method.

ftoromanoff: I still want the nodata value to be interpreted as an elevation of zero in the texture. Can we check that together ?

mgermerie (approved):

PR #1997, opened 2023-02-06
Chore/integration

mgermerie:
## Description

Fix an issue with next version publication on npm registry.
Update version for some github actions.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -85,6 +85,9 @@ jobs:
     steps:
 
       - uses: actions/checkout@v2
+        with:
+          # fetch all branches
+          fetch-depth: 0
 
       # Configure git user for later command induced commits
       - uses: fregante/setup-git-user@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/integration.yml
```diff
@@ -180,18 +196,19 @@ jobs:
...
         with:
-          node-version: 15.x
+          node-version: 16.x
 
       - name: Message commit
         run: echo "is RELEASE => ${{ github.event.head_commit.message }} !!"
 
       - name: Get release version
-        run: echo ::set-output name=VERSION::$(node -e "console.log(require('./package.json').version);")
+        # run: echo ::set-output name=VERSION::$(node -e "console.log(require('./package.json').version);")
```
jailln: Remove commented out?

config/conventionalChangelog/config.js
```diff
@@ -49,10 +49,6 @@ var writerOpts = {
             discard = false;
         });
 
-        if (commit.header && commit.header.includes('-next.')) {
```
jailln, line 52: Don't we need to keep this to avoid publishing too many changelogs?

.github/workflows/integration.yml
mgermerie: Yep my bad, I forgot it

config/conventionalChangelog/config.js
mgermerie: Not anymore, since next version commits are now pushed to the `next` branch. The changelog only looks for commits on `master` branch whose commit tree is untouched by the "next" versioning.

.github/workflows/integration.yml
mgermerie: I just found out there is a problem with this change : the tag created on [this line](https://github.com/iTowns/itowns/blob/7f86d26e184d3b64981d6c8767c195e8c2ab6a79/.github/workflows/integration.yml#L205) lacks the version number. It create a `v` tag instead of `v2.xx.xx`. I'll add that to #1998, so that this PR can be merged to have a working CI.

jailln (approved):

PR #2006, opened 2023-02-14
Rework style

ftoromanoff:
## Description
Change how itowns.Style() is used in relation with itowns.Feature.
- remove propertie 'style' at some level to keep to the miminum:
   - 2 were kept: layer and feature
   - 2 were removed: collection and geometry.properties
- make prioritarization of style work : layer (user) > feature (data) and this for all sub properties of style.
   - currently, the layer style is not always applied over the data style. (see example 1)
   -  when one properties of point, stroke or fill is given, it all needed to be given or the default value will be used even if the child properties is existing. 

Additional changes :
- The instanciation of itowns.style is now only done on converter and not anymore on Source nor during the Feature generation:
For that we currently create an object style, to store all properties used later on for the instanciation of itowns.Style().

- A slight change in base_altitude has been made, to be able to use a function for feature.style.

- All references to parent Style had been removed.

- Merge Style.symbolStylefromContext() with Style.drawingStylefromContext()

Examples:

1) source_file_kml_raster -> adding text.color: 'red'

in master branch:
image   
with that PR:
image

### DIFF
--- a/examples/js/plugins/FeatureToolTip.js
+++ b/examples/js/plugins/FeatureToolTip.js
@@ -64,42 +64,38 @@ const FeatureToolTip = (function _() {
         }
     }
 
-    function getGeometryProperties(geometry) {
-        return function properties() { return geometry.properties; };
-    }
-
     function fillToolTip(features, layer, options) {
         let content = '';
         let feature;
         let geometry;
-        let style;
+        const style = layer.style;
         let fill;
         let stroke;
         let symb = '';
         let prop;
 
+        const context = style.context;
+
         for (let p = 0; p < features.length; p++) {
             feature = features[p];
             geometry = feature.geometry;
-            style = (geometry.properties && geometry.properties.style) || feature.style || layer.style;
-            const context = { globals: {}, properties: getGeometryProperties(geometry) };
-            style = style.applyContext(context);
+
+            context.setFeature(feature);
+            context.setGeometry(geometry);
 
             if (feature.type === itowns.FEATURE_TYPES.POLYGON) {
                 symb = '&#9724';
-                if (style) {
-                    fill = style.fill && style.fill.color;
-                    stroke = style.stroke && ('1.25px ' + style.stroke.color);
-                }
+                fill = style.fill && style.fill.color;
+                stroke = style.stroke && ('1.25px ' + style.stroke.color);
             } else if (feature.type === itowns.FEATURE_TYPES.LINE) {
                 symb = '&#9473';
-                fill = style && style.stroke && style.stroke.color;
+                fill = style.stroke && style.stroke.color;
                 stroke = '0px';
             } else if (feature.type === itowns.FEATURE_TYPES.POINT) {
                 symb = '&#9679';
-                if (style && style.point) {  // Style and style.point can be undefined if no style options were passed
-                    fill = style.point.color;
-                    stroke = '1.25px ' + style.point.line;
+                if (style.point || style.icon) {  // Style and style.point can be undefined if no style options were passed
+                    fill = (style.point && style.point.color) || (style.icon && style.icon.color);
+                    stroke = '1.25px ' + ((style.point && style.point.line) || 'black');
                 }
             }
 
@@ -109,10 +105,10 @@ const FeatureToolTip = (function _() {
             content += '</span>';
 
             if (geometry.properties) {
-                content += (geometry.properties.description || geometry.properties.name || geometry.properties.nom || layer.name || '');
+                content += (geometry.properties.description || geometry.properties.name || geometry.properties.nom || geometry.properties.title || layer.name || '');
             }
 
-            if (feature.type === itowns.FEATURE_TYPES.POINT) {
+            if (feature.type === itowns.FEATURE_TYPES.POINT && options.writeLatLong) {
                 content += '<br/><span class="coord">long ' + feature.coordinates[0].toFixed(4) + '</span>';
                 content += '<br/><span class="coord">lat ' + feature.coordinates[1].toFixed(4) + '</span>';
             }
@@ -231,8 +227,8 @@ const FeatureToolTip = (function _() {
             }
 
             const opts = options || { filterAllProperties: true };
-            opts.filterProperties = opts.filterProperties == undefined ? [] : opts.filterProperties;
-            opts.filterProperties.concat(['name', 'nom', 'style', 'description']);
+            opts.filterProperties = opts.filterProperties === undefined ? [] : opts.filterProperties;
+            opts.writeLatLong = opts.writeLatLong || false;
 
             layers.push({ layer: layer, options: opts });
             layersId.push(layer.id);
--- a/examples/source_file_gpx_3d.html
+++ b/examples/source_file_gpx_3d.html
@@ -61,6 +61,15 @@
 
             var waypointGeometry = new itowns.THREE.BoxGeometry(1, 1, 80);
             var waypointMaterial = new itowns.THREE.MeshBasicMaterial({ color: 0xffffff });
+            const style = {
+                stroke: {
+                    color: 'red',
+                    width: 2,
+                },
+                point: {
+                    color: 'white',
+                }
+            };
             // Listen for globe full initialisation event
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 console.info('Globe initialized');
@@ -72,18 +81,9 @@
                         out: {
                             crs: view.referenceCrs,
                             structure: '3d',
-                            style: new itowns.Style({
-                                stroke: {
-                                    color: 'red',
-                                    width: 2,
-                                },
-                                point: {
-                                    color: 'white',
-                                }
-                            }),
                         }
                     }))
-                    .then(itowns.Feature2Mesh.convert())
+                    .then(itowns.Feature2Mesh.convert({style}))
                     .then(function (mesh) {
                         if (mesh) {
                             mesh.updateMatrixWorld();
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -191,6 +191,8 @@ function featureToPoint(feature, options) {
     const vertices = new Float32Array(ptsIn);
     inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
     normal.set(0, 0, 1).multiply(inverseScale);
+
+    const pointMaterialSize = [];
     context.globals = { point: true };
 
     for (const geometry of feature.geometries) {
@@ -207,8 +209,12 @@ function featureToPoint(feature, options) {
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, v));
             const style = feature.style.applyContext(context);
-            const { base_altitude, color } = style.point;
+            const { base_altitude, color, radius } = style.point;
             coord.z = 0;
+
+            if (!pointMaterialSize.includes(radius)) {
+                pointMaterialSize.push(radius);
+            }
 
             // populate vertices
             base.copy(normal).multiplyScalar(base_altitude).add(coord).toArray(vertices, v);
@@ -223,7 +229,11 @@ function featureToPoint(feature, options) {
     geom.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
     geom.setAttribute('batchId', new THREE.BufferAttribute(batchIds, 1));
 
-    options.pointMaterial.size = feature.style.point.radius;
+    options.pointMaterial.size = pointMaterialSize[0];
+    if (pointMaterialSize.length > 1) {
+        // TODO CREATE material for each feature
+        console.warn('Too many differents point.radius, only the first one will be used');
+    }
 
     return new THREE.Points(geom, options.pointMaterial);
 }
@@ -241,8 +251,7 @@ function featureToLine(feature, options) {
     const geom = new THREE.BufferGeometry();
     geom.setAttribute('position', new THREE.BufferAttribute(vertices, 3));
 
-    // TODO CREATE material for each feature
-    options.lineMaterial.linewidth = feature.style.stroke.width;
+    const lineMaterialWidth = [];
     context.globals = { stroke: true };
 
     const countIndices = (count - feature.geometries.length) * 2;
@@ -280,16 +289,24 @@ function featureToLine(feature, options) {
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, v));
             const style = feature.style.applyContext(context);
-            const { base_altitude, color } = style.stroke;
+            const { base_altitude, color, width } = style.stroke;
             coord.z = 0;
+
+            if (!lineMaterialWidth.includes(width)) {
+                lineMaterialWidth.push(width);
+            }
 
             // populate geometry buffers
             base.copy(normal).multiplyScalar(base_altitude).add(coord).toArray(vertices, v);
             toColor(color).multiplyScalar(255).toArray(colors, v);
             batchIds[j] = id;
         }
-
         featureId++;
+    }
+    options.lineMaterial.linewidth = lineMaterialWidth[0];
+    if (lineMaterialWidth.length > 1) {
+        // TODO CREATE material for each feature
+        console.warn('Too many differents stroke.width, only the first one will be used');
     }
     geom.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
     geom.setAttribute('batchId', new THREE.BufferAttribute(batchIds, 1));
@@ -634,7 +651,6 @@ export default {
             context.setCollection(collection);
 
             const features = collection.features;
-
             if (!features || features.length == 0) { return; }
 
             const meshes = features.map(feature => featureToMesh(feature, options));
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -7,7 +7,7 @@ import Extent from 'Core/Geographic/Extent';
 import Crs from 'Core/Geographic/Crs';
 import OrientationUtils from 'Utils/OrientationUtils';
 import Coordinates from 'Core/Geographic/Coordinates';
-import { StyleContext } from 'Core/Style';
+import Style, { StyleContext } from 'Core/Style';
 
 const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 const context = new StyleContext();
@@ -194,6 +194,7 @@ function featureToPoint(feature, options) {
 
     const pointMaterialSize = [];
     context.globals = { point: true };
+    context.setFeature(feature);
 
     for (const geometry of feature.geometries) {
         const start = geometry.indices[0].offset;
@@ -208,7 +209,7 @@ function featureToPoint(feature, options) {
             }
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, v));
-            const style = feature.style.applyContext(context);
+            const style = Style.applyContext(context);
             const { base_altitude, color, radius } = style.point;
             coord.z = 0;
 
@@ -253,6 +254,7 @@ function featureToLine(feature, options) {
 
     const lineMaterialWidth = [];
     context.globals = { stroke: true };
+    context.setFeature(feature);
 
     const countIndices = (count - feature.geometries.length) * 2;
     const indices = getIntArrayFromSize(countIndices, count);
@@ -288,7 +290,7 @@ function featureToLine(feature, options) {
             }
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, v));
-            const style = feature.style.applyContext(context);
+            const style = Style.applyContext(context);
             const { base_altitude, color, width } = style.stroke;
             coord.z = 0;
 
@@ -322,6 +324,7 @@ function featureToPolygon(feature, options) {
     const batchIds = new Uint32Array(vertices.length / 3);
     const batchId = options.batchId || ((p, id) => id);
     context.globals = { fill: true };
+    context.setFeature(feature);
 
     inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
     normal.set(0, 0, 1).multiply(inverseScale);
@@ -349,7 +352,7 @@ function featureToPolygon(feature, options) {
             }
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, i));
-            const style = feature.style.applyContext(context);
+            const style = Style.applyContext(context);
             const { base_altitude, color } = style.fill;
             coord.z = 0;
 
@@ -410,6 +413,7 @@ function featureToExtrudedPolygon(feature, options) {
     let featureId = 0;
 
     context.globals = { fill: true };
+    context.setFeature(feature);
     inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
     normal.set(0, 0, 1).multiply(inverseScale);
     coord.setCrs(context.collection.crs);
@@ -435,7 +439,7 @@ function featureToExtrudedPolygon(feature, options) {
 
             coord.copy(context.setLocalCoordinatesFromArray(ptsIn, i));
 
-            const style = feature.style.applyContext(context);
+            const style = Style.applyContext(context);
             const { base_altitude, extrusion_height, color } = style.fill;
             coord.z = 0;
 
@@ -653,6 +657,7 @@ export default {
                 // as in examples/source_file_gpx_3d.html.
                 options.layer = this || { style: options.style };
             }
+            context.layerStyle = options.layer.style;
 
             context.setCollection(collection);
 
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -525,12 +525,13 @@ function createInstancedMesh(mesh, count, ptsIn) {
  * Convert a [Feature]{@link Feature} of type POINT to a Instanced meshes
  *
  * @param {Object} feature
+ * @param {Object} options - options controlling the conversion
  * @returns {THREE.Mesh} mesh or GROUP of THREE.InstancedMesh
  */
-function pointsToInstancedMeshes(feature) {
+function pointsToInstancedMeshes(feature, options) {
     const ptsIn = feature.vertices;
     const count = feature.geometries.length;
-    const modelObject = feature.style.point.model.object;
+    const modelObject = options.layer.style.point.model.object;
 
     if (modelObject instanceof THREE.Mesh) {
         return createInstancedMesh(modelObject, count, ptsIn);
@@ -541,7 +542,7 @@ function pointsToInstancedMeshes(feature) {
         meshes.forEach(mesh => group.add(createInstancedMesh(mesh, count, ptsIn)));
         return group;
     } else {
-        throw new Error('The format of the model object provided in the feature style (feature.style.point.model.object) is not supported. Only THREE.Mesh or THREE.Object3D are supported.');
+        throw new Error('The format of the model object provided in the style (layer.style.point.model.object) is not supported. Only THREE.Mesh or THREE.Object3D are supported.');
     }
 }
 
@@ -560,9 +561,9 @@ function featureToMesh(feature, options) {
     let mesh;
     switch (feature.type) {
         case FEATURE_TYPES.POINT:
-            if (feature.style.point?.model?.object) {
+            if (options.layer?.style?.point?.model?.object) {
                 try {
-                    mesh = pointsToInstancedMeshes(feature);
+                    mesh = pointsToInstancedMeshes(feature, options);
                     mesh.isInstancedMesh = true;
                 } catch (e) {
                     mesh = featureToPoint(feature, options);
@@ -575,7 +576,7 @@ function featureToMesh(feature, options) {
             mesh = featureToLine(feature, options);
             break;
         case FEATURE_TYPES.POLYGON:
-            if (feature.style.fill.extrusion_height) {
+            if (options.layer?.style?.fill.extrusion_height) {
                 mesh = featureToExtrudedPolygon(feature, options);
             } else {
                 mesh = featureToPolygon(feature, options);
@@ -645,7 +646,12 @@ export default {
                 options.pointMaterial = ReferLayerProperties(new THREE.PointsMaterial(), this);
                 options.lineMaterial = ReferLayerProperties(new THREE.LineBasicMaterial(), this);
                 options.polygonMaterial = ReferLayerProperties(new THREE.MeshBasicMaterial(), this);
-                options.layer = this;
+                // options.layer.style will be used later on to define the final style.
+                // In the case we didn't instanciate the layer before the convert, we can directly
+                // pass a style using options.style.
+                // This is usually done in some tests and if you want to use Feature2Mesh.convert()
+                // as in examples/source_file_gpx_3d.html.
+                options.layer = this || { style: options.style };
             }
 
             context.setCollection(collection);
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -11,6 +11,8 @@ import Style, { StyleContext } from 'Core/Style';
 
 const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 const context = new StyleContext();
+const defaultStyle = new Style();
+let style;
 
 const dim_ref = new THREE.Vector2();
 const dim = new THREE.Vector2();
@@ -193,7 +195,6 @@ function featureToPoint(feature, options) {
     normal.set(0, 0, 1).multiply(inverseScale);
 
     const pointMaterialSize = [];
-    context.globals = { point: true };
     context.setFeature(feature);
 
     for (const geometry of feature.geometries) {
@@ -209,7 +210,7 @@ function featureToPoint(feature, options) {
             }
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, v));
-            const style = Style.applyContext(context);
+            style.setContext(context);
             const { base_altitude, color, radius } = style.point;
             coord.z = 0;
 
@@ -253,7 +254,6 @@ function featureToLine(feature, options) {
     geom.setAttribute('position', new THREE.BufferAttribute(vertices, 3));
 
     const lineMaterialWidth = [];
-    context.globals = { stroke: true };
     context.setFeature(feature);
 
     const countIndices = (count - feature.geometries.length) * 2;
@@ -290,7 +290,7 @@ function featureToLine(feature, options) {
             }
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, v));
-            const style = Style.applyContext(context);
+            style.setContext(context);
             const { base_altitude, color, width } = style.stroke;
             coord.z = 0;
 
@@ -323,7 +323,6 @@ function featureToPolygon(feature, options) {
 
     const batchIds = new Uint32Array(vertices.length / 3);
     const batchId = options.batchId || ((p, id) => id);
-    context.globals = { fill: true };
     context.setFeature(feature);
 
     inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
@@ -352,7 +351,7 @@ function featureToPolygon(feature, options) {
             }
 
             coord.copy(context.setLocalCoordinatesFromArray(feature.vertices, i));
-            const style = Style.applyContext(context);
+            style.setContext(context);
             const { base_altitude, color } = style.fill;
             coord.z = 0;
 
@@ -412,7 +411,6 @@ function featureToExtrudedPolygon(feature, options) {
 
     let featureId = 0;
 
-    context.globals = { fill: true };
     context.setFeature(feature);
     inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
     normal.set(0, 0, 1).multiply(inverseScale);
@@ -439,7 +437,7 @@ function featureToExtrudedPolygon(feature, options) {
 
             coord.copy(context.setLocalCoordinatesFromArray(ptsIn, i));
 
-            const style = Style.applyContext(context);
+            style.setContext(context);
             const { base_altitude, extrusion_height, color } = style.fill;
             coord.z = 0;
 
@@ -529,13 +527,12 @@ function createInstancedMesh(mesh, count, ptsIn) {
  * Convert a [Feature]{@link Feature} of type POINT to a Instanced meshes
  *
  * @param {Object} feature
- * @param {Object} options - options controlling the conversion
  * @returns {THREE.Mesh} mesh or GROUP of THREE.InstancedMesh
  */
-function pointsToInstancedMeshes(feature, options) {
+function pointsToInstancedMeshes(feature) {
     const ptsIn = feature.vertices;
     const count = feature.geometries.length;
-    const modelObject = options.layer.style.point.model.object;
+    const modelObject = style.point.model.object;
 
     if (modelObject instanceof THREE.Mesh) {
         return createInstancedMesh(modelObject, count, ptsIn);
@@ -552,9 +549,9 @@ function pointsToInstancedMeshes(feature, options) {
 
 /**
  * Convert a [Feature]{@link Feature} to a Mesh
- *
  * @param {Feature} feature - the feature to convert
  * @param {Object} options - options controlling the conversion
+ *
  * @return {THREE.Mesh} mesh or GROUP of THREE.InstancedMesh
  */
 function featureToMesh(feature, options) {
@@ -565,9 +562,9 @@ function featureToMesh(feature, options) {
     let mesh;
     switch (feature.type) {
         case FEATURE_TYPES.POINT:
-            if (options.layer?.style?.point?.model?.object) {
+            if (style.point?.model?.object) {
                 try {
-                    mesh = pointsToInstancedMeshes(feature, options);
+                    mesh = pointsToInstancedMeshes(feature);
                     mesh.isInstancedMesh = true;
                 } catch (e) {
                     mesh = featureToPoint(feature, options);
@@ -580,7 +577,7 @@ function featureToMesh(feature, options) {
             mesh = featureToLine(feature, options);
             break;
         case FEATURE_TYPES.POLYGON:
-            if (options.layer?.style?.fill.extrusion_height) {
+            if (style.fill && Object.keys(style.fill).includes('extrusion_height')) {
                 mesh = featureToExtrudedPolygon(feature, options);
             } else {
                 mesh = featureToPolygon(feature, options);
@@ -594,10 +591,6 @@ function featureToMesh(feature, options) {
         mesh.material.color = new THREE.Color(0xffffff);
     }
     mesh.feature = feature;
-
-    if (options.layer) {
-        mesh.layer = options.layer;
-    }
 
     return mesh;
 }
@@ -614,6 +607,8 @@ export default {
      * @param {function} [options.batchId] - optional function to create batchId attribute.
      * It is passed the feature property and the feature index. As the batchId is using an unsigned int structure on 32 bits,
      * the batchId could be between 0 and 4,294,967,295.
+     * @param {StyleOptions} [options.style] - optional style properties. Only needed if the convert is used without instancing
+     * a layer beforehand.
      * @return {function}
      * @example <caption>Example usage of batchId with featureId.</caption>
      * view.addLayer({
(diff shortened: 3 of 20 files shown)


jailln:
Can you rebase to resolve conflicts please?

examples/js/plugins/FeatureToolTip.js
```diff
@@ -81,9 +81,51 @@ var FeatureToolTip = (function _() {
...
+            //         ...featStyle.point,
+            //         ...layer.style.point,
+            //     },
+            //     icon: {
+            //         ...featStyle.icon,
+            //         ...layer.style.icon,
+            //     },
+            // };
+
+            // TO DO solve the problem with {...}
```
jailln: Do this todo :)

examples/js/plugins/FeatureToolTip.js
```diff
@@ -81,9 +81,51 @@ var FeatureToolTip = (function _() {
...
+            /* eslint-disable prefer-object-spread */
+            var styleConc = {
```
jailln: I think it should not be the responsability of the user (in this case `FeatureToolTip` but it is also done in the converters e.g. `Feature2Mesh`) to deal with this concatenation. Also, this code is duplicated many times. I think it should be factorized somewhere so one just have to do feat.style to get the final style and not to concatenate the styles. Maybe in `Feature` or in `Style` or in `Layer` or in a dedicated controller. WDYT?

examples/js/plugins/FeatureToolTip.js
```diff
@@ -231,8 +273,9 @@ var FeatureToolTip = (function _() {
             }
 
             var opts = options || { filterAllProperties: true };
-            opts.filterProperties = opts.filterProperties == undefined ? [] : opts.filterProperties;
-            opts.filterProperties.concat(['name', 'nom', 'style', 'description']);
+            opts.filterProperties = opts.filterProperties === undefined ? [] : opts.filterProperties;
+            opts.writeLatLong = opts.writeLatLong || false;
+            // opts.filterProperties.concat(['name', 'nom', 'style', 'description']);
```
jailln: remove commented out

src/Converter/Feature2Mesh.js
```diff
@@ -200,6 +201,7 @@ function addExtrudedPolygonSideFaces(indices, length, offset, count, isClockWise
 }
 
 function featureToPoint(feature, options) {
+    // console.log('featureToPoint');
```
jailln: remove

src/Converter/Feature2Mesh.js
```diff
@@ -235,12 +253,13 @@ function featureToPoint(feature, options) {
...
     if (batchIds) { geom.setAttribute('batchId', new THREE.BufferAttribute(batchIds, 1)); }
 
-    options.pointMaterial.size = feature.style.point.radius;
+    options.pointMaterial.size = styleConc.point.radius;
 
     return new THREE.Points(geom, options.pointMaterial);
 }
 
 function featureToLine(feature, options) {
+    // console.log('featureToLine');
```
jailln: remove

src/Converter/Feature2Mesh.js
```diff
@@ -318,6 +352,7 @@ function featureToLine(feature, options) {
 }
 
 function featureToPolygon(feature, options) {
+    // console.log('featureToPolygon');
```
jailln: remove

src/Converter/Feature2Mesh.js
```diff
@@ -395,6 +449,7 @@ function area(contour, offset, count) {
 
 const bottomColor = new THREE.Color();
 function featureToExtrudedPolygon(feature, options) {
+    // console.log('featureToExtrudedPolygon');
```
jailln: remove

src/Converter/Feature2Mesh.js
```diff
@@ -502,7 +572,8 @@ function featureToMesh(feature, options) {
             mesh = featureToLine(feature, options);
             break;
         case FEATURE_TYPES.POLYGON:
-            if (feature.style.fill.extrusion_height) {
+            // if (feature.style.fill.extrusion_height) {
```
jailln: remove

src/Converter/Feature2Mesh.js
```diff
@@ -581,6 +652,7 @@ export default {
             if (!features || features.length == 0) { return; }
 
             options.GlobalZTrans = collection.center.z;
+            options.collection = { style: collection.style };
```
jailln: unused?

src/Converter/Feature2Texture.js
```diff
@@ -200,7 +222,8 @@ export default {
 
             // Draw the canvas
             for (const feature of collection.features) {
-                drawFeature(ctx, feature, featureExtent, feature.style || style, invCtxScale);
+                // drawFeature(ctx, feature, featureExtent, feature.style || style, invCtxScale);
```
jailln: remove

src/Core/Feature.js
```diff
@@ -198,6 +198,10 @@ function push3DValues(value0, value1, value2 = 0) {
     this.vertices[this._pos++] = value2;
 }
 
+function base_altitudeDefault(properties, coordinates = { z: 0 }) {
```
jailln: `baseAltitudeDefault`

src/Core/Feature.js
```diff
@@ -259,7 +263,19 @@ class Feature {
         }
         this._pos = 0;
         this._pushValues = (this.size === 3 ? push3DValues : push2DValues).bind(this);
-        this.style = new Style({}, collection.style);
+        // this.style = new Style({}, collection.style);
```
jailln: remove

src/Core/Feature.js
```diff
@@ -259,7 +263,19 @@ class Feature {
...
+        this.style = {
+            fill: {},
+            stroke: {},
+            point: {},
+        };
+
+        this.style = Style.setFromGeojsonProperties;
```
jailln: what about when it is not a geojson?

src/Core/Feature.js
```diff
@@ -259,7 +263,19 @@ class Feature {
         }
         this._pos = 0;
         this._pushValues = (this.size === 3 ? push3DValues : push2DValues).bind(this);
-        this.style = new Style({}, collection.style);
+        // this.style = new Style({}, collection.style);
+        this.style = {
```
jailln: why setting it to empty objects and setting it to something else the following line?

src/Core/Feature.js
```diff
@@ -532,11 +548,11 @@ export class FeatureCollection extends THREE.Object3D {
...
     }
 
-    setParentStyle(style) {
-        if (style) {
-            this.features.forEach((f) => {
-                f.style.parent = style;
-            });
-        }
-    }
+    // setParentStyle(style) {
```
jailln: remove

src/Core/Style.js
```diff
@@ -139,19 +139,13 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
-                }
-            },
+            get: () => property ?? defaultValue,
             set: (v) => {
                 property = v;
             },
         });
 
-    style[category][name] = value;
+    if (value !== undefined) { style[category][name] = value; }
```
jailln: we should set something if it is undefined? + what about null?

src/Core/Style.js
```diff
@@ -324,19 +318,20 @@ class Style {
      *
      * @constructor
      */
-    constructor(params = {}, parent) {
+    // constructor(params = {}, parent) {
```
jailln: remove

src/Core/Style.js
```diff
@@ -324,19 +318,20 @@ class Style {
...
 
-        this.parent = parent || {
-            zoom: {},
-            fill: {},
-            stroke: {},
-            point: {},
-            text: {},
-            icon: {},
-        };
+        // this.parent = parent || {
```
jailln: remove

src/Core/Style.js
```diff
@@ -437,8 +434,11 @@ class Style {
      * @return     {Object}  mapped style depending on context.
      */
     symbolStylefromContext(context) {
-        const style = new Style();
-        mapPropertiesFromContext('text', this, style, context);
+        // const style = new Style();
```
jailln: remove

src/Core/Style.js
```diff
@@ -437,8 +434,11 @@ class Style {
...
+        const style = {};
+        if (this.text) {
+            mapPropertiesFromContext('text', this, style, context);
```
jailln: this method is always called with `style = {}` now so you can remove it from the parameters

src/Core/Style.js
```diff
@@ -476,39 +476,57 @@ class Style {
      * set Style from geojson properties.
      * @param {object} properties geojson properties.
      * @param {number} type
-     * @returns {Style}
+     * @returns {object} contening all properties for itowns.Style
```
jailln: typo: containing

src/Core/Style.js
```diff
@@ -517,92 +535,100 @@ class Style {
      * @param {Object} sprites vector tile layer.
      * @param {number} [order=0]
      * @param {boolean} [symbolToCircle=false]
-     * @returns {Style}
+     * @returns {object} contening all properties for itowns.Style
```
jailln: containing

src/Layer/ColorLayer.js
```diff
@@ -84,7 +83,8 @@ class ColorLayer extends RasterLayer {
         deprecatedColorLayerOptions(config);
         super(id, config);
         this.isColorLayer = true;
-        this.style = new Style(config.style);
+        // this.style = new Style(config.style);
```
jailln: remove

test/unit/style.js
```diff
@@ -107,4 +108,74 @@ describe('Style', function () {
...
+            assert.equal(style.point.radius, 2);
+            assert.equal(style.text.color, '#eba55f');
+            assert.equal(style.icon.color, '#eba55f');
+        });
+        it('FEATURE_TYPES.POLYGONE', () => {
+            const geoJsonProperties = {
+                fill: '#eba55f',
+                stroke: '#eba55f',
+            };
+            const style = Style.setFromGeojsonProperties(geoJsonProperties, FEATURE_TYPES.POLYGONE);
```
jailln: POLYGONE -> POLYGON

jailln (changes requested):
Thanks for the corrections :)

Some cleaning is required (many comments are about removing commented out code).

Some questions / remarks:
* How was the collection and geometry.properties used and what features do we loose by removing them (if any)
* Regarding: 
> when one properties of point, stroke or fill is given, it all needed to be given or the default value will be used even if the child properties is existing.
Just to be sure to understand correctly, it means that if someones specifies the fill color in the layer it will be used for the features but the fill opacity (and other fill properties) will be set to default values instead of the source style (from the original file)?
*  In many places, the style is manipulated through json objects instead of `Style` objects, why? Whats is the purpose of having a class then? I would prefer that you use a `Style` instance each time like it was done before instead (you can write a constructor that instantiates an empty style if that's why you are not using it?)
* Why taking base altitude out from the style and putting it in the feature? I think it should stay in the `Style` to be consistent
* How was the parent style used and what feature do we loose by removing it (if any)?
* Once the implementation is done we need to think about the breaking changes introduced by this PR

gchoqueux:
Thanks for your contribution,

There are conventions for PRs, you could find them in these documents:
* [CODING.md](https://github.com/iTowns/itowns/blob/master/CODING.md) with part **`Code rules`**
* [CONTRIBUTING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) with part **`Pull Request Guidelines`**

Could you verify if this PR verifies these conventions?

examples/js/plugins/FeatureToolTip.js
ftoromanoff: Done

examples/js/plugins/FeatureToolTip.js
ftoromanoff: I did add a mergeStyles function in Style for doign that concatenation of style properties with hyerarchisation of value.

examples/js/plugins/FeatureToolTip.js
ftoromanoff: Cleaned up

src/Converter/Feature2Mesh.js
ftoromanoff: Cleaned up

src/Converter/Feature2Mesh.js
ftoromanoff: Cleaned up

src/Converter/Feature2Mesh.js
ftoromanoff: Cleaned up

src/Converter/Feature2Mesh.js
ftoromanoff: Cleaned up

src/Converter/Feature2Mesh.js
ftoromanoff: Cleaned up

src/Converter/Feature2Mesh.js
ftoromanoff: It was indeed not really used anymore and the purpose was to remove it, so I did.

src/Converter/Feature2Texture.js
ftoromanoff: Cleaned up

src/Core/Feature.js
ftoromanoff: Renamed

src/Core/Feature.js
ftoromanoff: Cleaned up

src/Core/Feature.js
ftoromanoff: It was not a good name indeed (It was at first used only in Geojson so the original name). I renamed it as now it's just pick the information for the itowns.feature.geometry.properties.

src/Core/Feature.js
ftoromanoff: Cleaned up

src/Core/Feature.js
ftoromanoff: Cleaned up

src/Core/Style.js
ftoromanoff: I tought it was neccessary but it's not, so i put it back to what it was before.

src/Core/Style.js
ftoromanoff: Cleaned up

src/Core/Style.js
ftoromanoff: Cleaned up

src/Core/Style.js
ftoromanoff: Cleaned up

src/Core/Style.js
ftoromanoff: It seems like indeed style is always empty, but it's not as this function change the style object in parameter, so the next line style is not empty anymore.

src/Core/Style.js
ftoromanoff: Corrected

src/Core/Style.js
ftoromanoff: idem

src/Layer/ColorLayer.js
ftoromanoff: Cleaned up

test/unit/style.js
ftoromanoff: Corrected

ftoromanoff:
> Thanks for the corrections :)
> 
> Some cleaning is required (many comments are about removing commented out code).
> 
> Some questions / remarks:
> 
> * How was the collection and geometry.properties used and what features do we loose by removing them (if any)

It was not really used as it was just copy of other styles, removing them doesn't change anything and simplify the understanding of the real styles needed.

> * Regarding:
> 
> > when one properties of point, stroke or fill is given, it all needed to be given or the default value will be used even if the child properties is existing.
> > Just to be sure to understand correctly, it means that if someones specifies the fill color in the layer it will be used for the features but the fill opacity (and other fill properties) will be set to default values instead of the source style (from the original file)?

> 
> * In many places, the style is manipulated through json objects instead of `Style` objects, why? Whats is the purpose of having a class then? I would prefer that you use a `Style` instance each time like it was done before instead (you can write a constructor that instantiates an empty style if that's why you are not using it?)

Using a Style instance will set the default value or we want to be able to now if a value of a child propertie is define at the layer (user) or in the data before using the default value. I fell like the Style;object is more an intrinsect itowns object than really an object the user should use. As the user should just give parameters to be used for the style, as any other options.


> * Why taking base altitude out from the style and putting it in the feature? I think it should stay in the `Style` to be consistent
The base altitude, in opposition with the other style properties, is used during the creation of the itowns.feature object (to get the 3 coordinates: x, y, z) but it still need to be fill out by the user in the style options.
As the other style properties will be used during the set up of the mesh, the texture or the label, putting it on the side allow us to use function to get the differentes style properties.

> * How was the parent style used and what feature do we loose by removing it (if any)?
The previous style systeme using style at all level (with many duplication) was using the parent systeme to be alble to get throught all that duplication, as i choose to only keep the two style level needed: the user level (layer.style) and the data level (itowns.feature.geometry.style that i moved to feature with the help of function) there is no purpose of the old systeme. Removing it simplify the understanding of the general styles, but doesn't change anything as it was only duplication to support the different parsing systeme.

> * Once the implementation is done we need to think about the breaking changes introduced by this PR
This PR doesnt bring breaking change as it's mainly simplify and homogenize the general style syteme used in itowns.feature.
the only change for the user is that if it used itowns.style instead of just an object style, it will not use the data style inforamtion if any.

src/Core/Style.js
```diff
@@ -407,7 +382,7 @@ class Style {
     }
 
     /**
-     * Map drawing properties style (fill, stroke and point) from context to object.
+     * Map drawing properties style (fill, stroke, point, symbol and icon) from context to object.
      * Only the necessary properties are mapped to object.
      * if a property is expression, the mapped value will be the expression result depending on context.
      * @param      {Object}  context  The context
```
jailln: Could you document what is context (whats is it structure and purpose) please if you know what it is since you've worked on it? (even if it's not the purpose of this PR)

src/Core/Style.js
```diff
@@ -473,42 +438,94 @@ class Style {
     }
 
     /**
-     * set Style from geojson properties.
-     * @param {object} properties geojson properties.
+     * merge 2 style object into 1. Prioritazing for each properties the first style
+     * in the list
+     * @param {object} styleHigh first object style with higher priority.
+     * @param {object} styleLow object style with lower priority.
+     * @returns {object} containing all properties for itowns.Style
```
jailln: the type of the return is `Style` and not `object`, could you update it and the description please?

jailln (commented):
Thanks it's clearer this way! 

There is still one thing that is bothering me:
I think we should continue to pass a `Style` object in layers options and not a plain javascript object because:
* Style options can be verified in `Style` constructor
* It provides the user with a clear style API and associated documentation of what style options can be passed
* We always manipulate a `Style` object and not sometimes a plain javascript object and other times a `Style` object

This may imply to remove static from `mergeStyles` and to put back `base_altitude` in `Style`.

WDYT? @mgermerie and @gchoqueux do you have an opinion about this?

src/Converter/Feature2Mesh.js
```diff
@@ -249,7 +251,8 @@ function featureToPoint(feature, options) {
     geom.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
     if (batchIds) { geom.setAttribute('batchId', new THREE.BufferAttribute(batchIds, 1)); }
 
-    options.pointMaterial.size = feature.style.point.radius;
+    options.pointMaterial.size = options.layer?.style.point.radius ||
```
jailln: I think doing `options.pointMaterial.size = style.point.radius` after `const style = Style.merge(options.layer?.style, feature.style).drawingStylefromContext(context);` would be clearer. In this case we should however check that the radius is always the same for all geometries, display a warning and keep the first one if it's not.

src/Converter/Feature2Mesh.js
```diff
@@ -277,7 +280,9 @@ function featureToLine(feature, options) {
     let lines;
 
     // TODO CREATE material for each feature
-    options.lineMaterial.linewidth = feature.style.stroke.width;
+    options.lineMaterial.linewidth = options.layer?.style.stroke.width ||
```
jailln: same comment than for points

src/Converter/Feature2Mesh.js
```diff
@@ -522,12 +533,13 @@ function createInstancedMesh(mesh, count, ptsIn) {
  * Convert a [Feature]{@link Feature} of type POINT to a Instanced meshes
  *
  * @param {Object} feature
+ * @param {Object} options - options controlling the conversion
```
jailln: Update doc to something more specific explaining which options can be passed and how they "control" the conversion

src/Converter/Feature2Mesh.js
```diff
@@ -643,7 +655,7 @@ export default {
                 options.pointMaterial = ReferLayerProperties(new THREE.PointsMaterial(), this);
                 options.lineMaterial = ReferLayerProperties(new THREE.LineBasicMaterial(), this);
                 options.polygonMaterial = ReferLayerProperties(new THREE.MeshBasicMaterial(), this);
-                options.layer = this;
+                options.layer = this || options.layer;
```
jailln: why this change?

src/Core/Style.js
jailln: context contains feature properties

src/Converter/Feature2Mesh.js
```diff
@@ -643,7 +686,7 @@ export default {
                 options.pointMaterial = ReferLayerProperties(new THREE.PointsMaterial(), this);
                 options.lineMaterial = ReferLayerProperties(new THREE.LineBasicMaterial(), this);
                 options.polygonMaterial = ReferLayerProperties(new THREE.MeshBasicMaterial(), this);
-                options.layer = this;
+                options.layer = this || { style: options.style };
```
jailln: why this change?

src/Converter/Feature2Texture.js
```diff
@@ -105,17 +105,27 @@ const coord = new Coordinates('EPSG:4326', 0, 0, 0);
...
 
     for (const geometry of feature.geometries) {
         if (Extent.intersectsExtent(geometry.extent, extent)) {
             const context = { globals, properties: () => geometry.properties };
-            const contextStyle = (geometry.properties.style || style).drawingStylefromContext(context);
+
+            let featStyle = feature.style;
+            if (featStyle instanceof Function) {
+                featStyle = feature.style(geometry.properties, feature.type);
+            }
```
jailln: Factorize these lines somewhere since we always call them before `Style.merge(style, featStyle)`

src/Core/Feature.js
```diff
@@ -129,7 +129,7 @@ export class FeatureGeometry {
     }
 
     baseAltitude(feature, coordinates) {
-        const base_altitude = feature.style[typeToStyleProperty[feature.type]].base_altitude || 0;
+        const base_altitude = feature.base_altitude[typeToStyleProperty[feature.type]] || 0;
```
jailln: `|| 0` not needed because set to `baseAltitudeDefault` in `Feature` constructor?

src/Core/Feature.js
```diff
@@ -259,7 +263,9 @@ class Feature {
         }
         this._pos = 0;
         this._pushValues = (this.size === 3 ? push3DValues : push2DValues).bind(this);
-        this.style = new Style({}, collection.style);
+
+        this.style = Style.setFromProperties;
+        this.base_altitude = collection.base_altitude;
```
jailln: rename to `baseAltitude`

src/Core/Style.js
```diff
@@ -155,16 +145,158 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
 }
 
 /**
- * A Style is an object that defines the visual appearance of {@link
+ * @description An object that can contain any properties
+ * (order, zoom, fill, stroke, point, text or/and icon)
+ * and sub properties of a Style.
+ * @typedef {Object} Style_Interface
```
jailln: I think it should be called `StyleOptions` instead to be more consistent (e.g. with CameraTransformOptions)

src/Converter/Feature2Mesh.js
ftoromanoff: To be able to manage the case where there is no layer instanced yet: in soem tests or if you use your own fetcher/parser.

src/Core/Feature.js
ftoromanoff: I think it's for the case or we use baseAltitudeDefault with coordinates defined but with no z key in it..

src/Converter/Feature2Mesh.js
jailln: can you add a comment in the code to explain it please ?

jailln (approved):

src/Core/Style.js
```diff
@@ -27,27 +23,6 @@ function mapPropertiesFromContext(mainKey, from, to, context) {
     }
 }
 
-function rgba2rgb(orig) {
```
gchoqueux: Could you avoid to move `rgba2rgb` method if there's no reason

src/Core/Style.js
```diff
@@ -355,29 +481,20 @@ class Style {
         defineStyleProperty(this, 'fill', 'color', params.fill.color);
         defineStyleProperty(this, 'fill', 'opacity', params.fill.opacity, 1.0);
         defineStyleProperty(this, 'fill', 'pattern', params.fill.pattern);
-        defineStyleProperty(this, 'fill', 'base_altitude', params.fill.base_altitude, base_altitudeDefault);
```
gchoqueux, line 584: you move `base_altitude`?

examples/source_file_from_fetched_data.html
```diff
@@ -70,9 +70,9 @@
 
         view.addLayer(new itowns.ColorLayer('Roads', {
             source: sourceFromFetchedData,
-            style: new itowns.Style({
+            style: {
                 stroke: { color: 'yellow' },
-            }),
```
gchoqueux: Remove instancing, it's good move. 
This change alone can be in another PR. 
This would simplify the review and see if there are any side effects.

It's still possible to put an instance of Style as an option?

examples/js/plugins/FeatureToolTip.js
```diff
@@ -81,9 +85,16 @@ const FeatureToolTip = (function _() {
...
+
+            const globals = {
+                fill: true,
+                stroke: true,
+                point: true,
+                icon: true,
+            };
+            const context = { globals, specifics: { properties: getGeometryProperties(geometry), type: getFeatureType(feature) } };
+
+            style = itowns.Style.merge(layer.style, feature.style, context).drawingStylefromContext(context);
```
gchoqueux: `Style.Merge` is always used with `drawingStylefromContext`?

src/Core/Feature.js
```diff
@@ -370,7 +376,11 @@ export class FeatureCollection extends THREE.Object3D {
         this.mergeFeatures = options.mergeFeatures === undefined ? true : options.mergeFeatures;
         this.size = options.structure == '3d' ? 3 : 2;
         this.filterExtent = options.filterExtent;
-        this.style = options.style;
+        this.baseAltitude = {
+            fill: (options.style?.fill?.base_altitude) || baseAltitudeDefault,
+            stroke: (options.style?.stroke?.base_altitude) || baseAltitudeDefault,
+            point: (options.style?.point?.base_altitude) || baseAltitudeDefault,
+        };
```
gchoqueux: This move is debatable, it's out PR scope, could you move in other PR?

src/Core/Style.js
```diff
@@ -155,16 +145,163 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
 }
 
 /**
- * A Style is an object that defines the visual appearance of {@link
+ * @typedef {Object} StyleOptions
```
gchoqueux, line 302: To lighten this PR, this documentation could be moved to another PR?

src/Core/Style.js
ftoromanoff: The reason is readability.
mapPropertiesFromContext() use readExpression() so i don't see why the rgba2rgb() wichich is only used in Style.setFromVectorTileLayer() is put in between them...
I don't see why to do just a PR for readability so as i changed a lot of Style.js in this PR, in my opinion, it's a good time to do the readabilty change as well...

src/Core/Style.js
ftoromanoff: Yes, as you noticed on a following comment, I did made a little change on the gestion of base_altitude. Base altitude is used on the instanciation of itowns.Feature objects and not as other style on the ThreeJs objects. Thus i moved it to Feature.js.

examples/source_file_from_fetched_data.html
ftoromanoff: It's still possible to put an instance of Style as an option but in this case (as it was before (in the cases where it was working)) you won't be able to get the missing style properties from the data metadata.
Removing the instancing can't really be done in another PR as you still need to instance it in the convert, for steing the default values)

examples/js/plugins/FeatureToolTip.js
ftoromanoff: Yes, This is something that indeed can be improved by merging this 2 functions but It will be better be done when we move all Style to Layer. as for the moment we still need 2 styles interface for the layer/user style and the data style.

src/Core/Feature.js
ftoromanoff: It's not out of scope as it is need if we want to be able to use functions to set the feature.style which is exactly what is been done in this PR.
This changes is invisible for the user, it is just internal.

src/Core/Style.js
ftoromanoff: The documentation is mostly the documentation of what is been changed in this PR, so i don't really see the use of doing a pre PR with minime change before added all the other changed in this PR...

LoicUTC:
Hi, could you please give us a status on this PR. Thanks

ftoromanoff:
Hello,
As this PR is pretty deep into the code, @gchoqueux wanted to have a look at it before the merge and he had a few remarks.
We also found a case where the style was not well handle in that PR (using image sprite in a VectorTileSource).
This last point is solved and the PR is waiting for @gchoqueux to have a final look.

jailln:
> Thanks it's clearer this way!
> 
> There is still one thing that is bothering me: I think we should continue to pass a `Style` object in layers options and not a plain javascript object because:
> 
> * Style options can be verified in `Style` constructor
> * It provides the user with a clear style API and associated documentation of what style options can be passed
> * We always manipulate a `Style` object and not sometimes a plain javascript object and other times a `Style` object
> 
> This may imply to remove static from `mergeStyles` and to put back `base_altitude` in `Style`.
> 
> WDYT? @mgermerie and @gchoqueux do you have an opinion about this?

My opinion after coming back to it a little while after:
* use a json object for the layer configuration (for users)
* instantiate a `Style` object from the style config in `Layer.js` and always manipulate `Style` objects internally. Therefore, keep the following in `Layer.js`:
```
        if (config.style) {
            config.style = new Style(config.style);
        }
```
WDYT @ftoromanoff ?

examples/source_file_gpx_3d.html
```diff
@@ -62,6 +62,15 @@
             var waypointGeometry = new itowns.THREE.BoxGeometry(1, 1, 80);
             var waypointMaterial = new itowns.THREE.MeshBasicMaterial({ color: 0xffffff });
             // Listen for globe full initialisation event
```
jailln: move comment below style declaration

examples/source_file_gpx_3d.html
```diff
@@ -72,18 +81,9 @@
...
-                                    width: 2,
-                                },
-                                point: {
-                                    color: 'white',
-                                }
-                            }),
                         }
                     }))
-                    .then(itowns.Feature2Mesh.convert())
+                    .then(itowns.Feature2Mesh.convert({style: new itowns.Style(style)}))
```
jailln: I thought we should pass a json object containing style options and `Style` instanciation should be done in itowns core? (and not by the user)

src/Converter/Feature2Mesh.js
```diff
@@ -628,16 +643,24 @@ export default {
...
             context.setCollection(collection);
 
             const features = collection.features;
-
             if (!features || features.length == 0) { return; }
 
-            const meshes = features.map(feature => featureToMesh(feature, options));
+            const meshes = features.map((feature) => {
+                const mesh = featureToMesh(feature, options);
+                mesh.layer = this;
```
jailln, line 663: why do we need to store the layer in the mesh now ?

src/Converter/Feature2Texture.js
```diff
@@ -121,8 +121,9 @@ const featureExtent = new Extent('EPSG:4326', 0, 0, 0, 0);
 export default {
     // backgroundColor is a THREE.Color to specify a color to fill the texture
     // with, given there is no feature passed in parameter
-    createTextureFromFeature(collection, extent, sizeTexture, style = {}, backgroundColor) {
+    createTextureFromFeature(collection, extent, sizeTexture, layerStyle = userStyle, backgroundColor) {
```
jailln: weird to do `layerStyle = userStyle` and then `userStyle = layerStyle` 2 lines below, what do you want to achieve?

src/Core/Style.js
```diff
@@ -39,19 +28,23 @@ export function readExpression(property, ctx) {
...
+                if (ctx.zoom >= stop[0]) {
                     return stop[1];
                 }
             }
             return property.stops[0][1];
-        } else if (property instanceof Function) {
+        }
+        if (typeof property === 'string' || property instanceof String) {
+            property = property.replace(/\{(.+?)\}/g, (a, b) => (ctx.properties[b] || '')).trim();
+            // return property.replace(/\{(.+?)\}/g, (a, b) => (ctx.properties[b] || '')).trim();
```
jailln: rm commented out

src/Core/Style.js
```diff
@@ -138,27 +131,27 @@ const textAnchorPosition = {
...
-        name,
+        parameter,
         {
             enumerable: true,
             get: () => {
-                if (property === undefined) {
-                    return style.parent[category][name] || defaultValue;
-                } else {
-                    return property;
+                if (property) { return property; }
```
jailln: Can you explain this getter please ? It seems overcomplicated but I might miss some cases

src/Core/Style.js
```diff
@@ -209,6 +208,17 @@ export class StyleContext {
         return this.#geometry.properties;
     }
 
+    get type() {
+        return this.#feature.type;
+    }
+    get fStyle() {
```
jailln: I think `featureStyle` would be clearer

src/Core/Style.js
```diff
@@ -221,6 +231,62 @@ export class StyleContext {
...
 
+function _addIcon(icon, domElement, opt) {
+    const cIcon = icon.cloneNode();
+
+    cIcon.setAttribute('class', 'itowns-icon');
+
+    // cIcon.width = icon.width * this.icon.size;
+    // cIcon.height = icon.height * this.icon.size;
+    // cIcon.style.color = this.icon.color;
+    // cIcon.style.opacity = this.icon.opacity;
```
jailln: rm commented out

src/Core/Style.js
```diff
@@ -649,83 +709,96 @@ class Style {
         return this;
     }
 
+    copy2(...style) {
```
jailln: Unused

src/Core/Style.js
```diff
@@ -848,16 +929,19 @@ class Style {
      * @param {Boolean} canBeFilled - true if feature.type == FEATURE_TYPES.POLYGON.
      */
     applyToCanvasPolygon(txtrCtx, polygon, invCtxScale, canBeFilled) {
+        const context = this.context;
         // draw line or edge of polygon
         if (this.stroke) {
             // TO DO add possibility of using a pattern (https://github.com/iTowns/itowns/issues/2210)
-            Style.prototype._applyStrokeToPolygon.call(this, txtrCtx, invCtxScale, polygon);
+            // Style.prototype._applyStrokeToPolygon.call(this, txtrCtx, invCtxScale, polygon);
```
jailln: rm

src/Core/Style.js
```diff
@@ -848,16 +929,19 @@ class Style {
...
+            this._applyStrokeToPolygon(txtrCtx, invCtxScale, polygon, context);
         }
 
         // fill inside of polygon
         if (canBeFilled && this.fill) {
             // canBeFilled can be move to StyleContext in the later PR
-            Style.prototype._applyFillToPolygon.call(this, txtrCtx, invCtxScale, polygon);
+            // Style.prototype._applyFillToPolygon.call(this, txtrCtx, invCtxScale, polygon);
```
jailln: rm

src/Layer/ColorLayer.js
```diff
@@ -84,7 +84,7 @@ class ColorLayer extends RasterLayer {
         deprecatedColorLayerOptions(config);
         super(id, config);
         this.isColorLayer = true;
-        this.style = config.style;
+        // this.style = config.style;
```
jailln: rm

src/Layer/LabelLayer.js
```diff
@@ -181,6 +181,7 @@ class LabelLayer extends GeometryLayer {
         delete config.domElement;
         super(id, config.object3d || new THREE.Group(), config);
 
+        // this.style = config.style || {};
```
jailln: rm

src/Layer/Layer.js
```diff
@@ -98,15 +100,16 @@ class Layer extends THREE.EventDispatcher {
...
         if (config.style && !(config.style instanceof Style)) {
             if (typeof config.style.fill?.pattern === 'string') {
                 console.warn('Using style.fill.pattern = { source: Img|url } is adviced');
                 config.style.fill.pattern = { source: config.style.fill.pattern };
             }
             config.style = new Style(config.style);
         }
-        this.isLayer = true;
-
+        this.style = config.style || userStyle;
```
jailln: I think we should avoid using the same instance here. There might indeed be a bug if a user does not define a style for two layers (that will therefore share the same Style instance) and if one then decides to change the style of one of the layer, this style will be applied to both layers.

src/Converter/Feature2Mesh.js
ftoromanoff: It was already done before. I just move the place where it's done.
At line 576 - 579, I removed :
```
    if (options.layer) {
        mesh.layer = options.layer;
    }
```

jailln: oh ok sorry, missed that

src/Converter/Feature2Texture.js
ftoromanoff: My idea was to limit the number of instanciation of a new Style as well as avoiding passing option.style as parameter in all sub functions of Feature2Texture.
So I decided to store userStyle at file scope and use it in all sub function.

I did some llittle change to make it clearer (I hope) and avoid potential side effects (I added a defautStyle = new Style() as file scope level.

src/Core/Style.js
ftoromanoff: Done.

examples/source_file_gpx_3d.html
ftoromanoff: Done.

src/Core/Style.js
ftoromanoff: Done.
In some way feature.Style fucnitons are style parsers. Maybe later on it can be rewriten differently.

src/Core/Style.js
ftoromanoff: Done.

src/Core/Style.js
ftoromanoff: Indeed, At some point I wanted to use a way to copy desciptor for a style and not the values (That can be a problem to get when the context in not yet defined).

src/Core/Style.js
ftoromanoff: Done.

src/Core/Style.js
ftoromanoff: Done.

src/Layer/ColorLayer.js
ftoromanoff: Done.

src/Layer/LabelLayer.js
ftoromanoff: Done.

src/Layer/Layer.js
ftoromanoff: Indeed, at Layer level we might want all Styles from all layers to be independents. I removed userStyle.

src/Core/Style.js
```diff
@@ -138,10 +138,10 @@ function defineStyleProperty(style, category, parameter, userValue, defaultValue
         {
             enumerable: true,
             get: () => {
-                if (property) { return property; }
-                if (userValue) { return readExpression(userValue, style.context); }
+                if (property !== undefined) { return property; }
```
jailln: you have to check for `null` too

src/Core/Style.js
```diff
@@ -895,6 +895,13 @@ class Style {
                 }
             }
         }
+        // VectorTileSet: by default minZoom = 0 and maxZoom = 24
+        // https://docs.mapbox.com/style-spec/reference/layers/#maxzoom and #minzoom
+        // Should be move to layer properties, when (if) VectorTile will be considered as multiLayer.
```
jailln: what do you mean by "multilayer" ?

src/Core/Style.js
ftoromanoff: done. Using value != undefined.

src/Core/Style.js
ftoromanoff: A VectorTileSource send for each tile data on several 'layer' (building, road etc for instance), thus the use of the propertiy filter (that should be move to layer in my opinion). So one Layer with a VectorTileSource will have data on several vectorTile 'layer' that can have diferent 'layer' properties (style, zoom max and min etc). That why i fell like this kind of source is more a multiLayer that a real Layer.

src/Core/Style.js
jailln: Seems to work for `null` and `undefined` but are you sure it won't be `true` in any other case? I haven't found many documentation about it and it is more generally done with `property !== undefined && property !== null` (even if it is more verbose)

src/Core/Style.js
```diff
@@ -130,6 +130,18 @@ const textAnchorPosition = {
     'top-left': [0, 0],
 };
 
+/**
+ * Define Getter/Setter for a specifique parameter of one of the five Style categories (fill, stroke, point, text, icon)
+ * and zoom of a Style instance.
+ * When the getter of a categorie.parameter is called, the purpose is to send, in that order, the value set by the user 'userValue'
+ * (if available), or the value read from the data 'dataValue' or the default value 'defaultValue' (if set). In case we
+ * need to overpass all we can set a value using the setter.
```
jailln: I propose to update the comment a bit:

```
Defines a new style property and its getter and setter for a given parameter in a given category (one of fill, stroke, point, text, icon or zoom).
The getter is in charge of returning the right style value from the following ones if they are defined (in that specific order):
* the value set by the user (`userValue`)
* the value read from the data source (`dataValue`)
* the default fallback value (`defaultValue`)
```

src/Core/Style.js
jailln: Ok I understand what you mean, there is itowns layer concept and mapbox layer concept; can you precise your comment then or open an issue please? Otherwise I'm afraid that we won't understand this comment in a few month...

src/Main.js
```diff
@@ -39,7 +39,7 @@ export { CAMERA_TYPE } from 'Renderer/Camera';
 
 // Internal itowns format
 export { default as Feature, FeatureCollection, FeatureGeometry, FEATURE_TYPES } from 'Core/Feature';
-export { default as Style } from 'Core/Style';
+// export { default as Style } from 'Core/Style';
```
jailln: why do you need this change ? I think it is out of scope of this PR and we should uncomment it for now.

src/Core/Style.js
ftoromanoff: I got the information on internet https://codedamn.com/news/javascript/check-if-undefined-null and checked it with https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Operators/Inequality.

src/Core/Style.js
ftoromanoff: I did add some documentation. Hopefully It will make it more clear.

src/Main.js
ftoromanoff: Indeed. It's not in the scope of the PR and was just a test to see if it can be removed safely in the futur (with the actual modification).

src/Core/Style.js
ftoromanoff: Thanks for that proposal that is indeed clearer that mine. I used it to update the documentation.

jailln (approved):

Issue #1936, opened 2022-10-24
Improve resources disposal when deleting a view


jailln:
Resource disposal might be improved when deleting a view (issue created form #1886)

## Context

When removing a view, not all resources of layers are correctly disposed. This has been improved in #1886 but there is still work to do. In particular, when removing a `FeatureGeometryLayer`, a lot of `Feature` and `FeatureGeometry` objects are not disposed correctly. For instance, here are Memory snapshots in the Chrome dev tools before and after removing the view of the `vector_tile_3d_mesh.html` example:

*Before removing the view:*

![image](https://user-images.githubusercontent.com/16967916/197491291-5f6ce083-80e2-46d0-88f5-38a0626ffaa7.png)

*After removing the view:*

![image](https://user-images.githubusercontent.com/16967916/197491448-0e091e7a-4255-4d73-b966-927f6fa684c4.png)

jailln:
* For the `vector_tile_3d_mesh.html` examples:
  * look at the retainers in the memory snapshot
  * it may come from the `_featuresCaches` in the Source. Therefore this issue may be related to  #1928.
*  Improvements are also still needed for 3d tiles examples (see #1886 )
* Other examples should be tested and dispose should be improved if needed

PR #2016, opened 2023-03-03
Resize listener is referenced to remove it at dispose

valentinMachado:
## Description
In View class the resize listener is referenced, to remove it at dispose

## Motivation and Context
This change allow to well dispose a View, and to allow GC to collect View disposed.

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -185,7 +185,8 @@ class View extends THREE.EventDispatcher {
 
         this._frameRequesters = { };
 
-        window.addEventListener('resize', () => this.resize(), false);
+        this._resizeListener = () => this.resize();
+        window.addEventListener('resize', this._resizeListener, false);
 
         this._changeSources = new Set();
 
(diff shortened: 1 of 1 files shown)


jailln:
Thanks! Does this solves #1936 ? Have you compared memory snapshots before and after removing the view before and after your changes?

jailln (approved):

valentinMachado:
> Thanks! Does this solves #1936 ? Have you compared memory snapshots before and after removing the view before and after your changes?

I don't know but before this change when a view was disposed the resize listener was still called, meaning the view was still referenced and thus not garbage collected (meaning everything referenced by the view such as layers I guess was not garbage collected, it could solves #1936 or be part of the fix).

In our use case https://github.com/VCityTeam/UD-Imuv we are creating and disposing many PlanarView, what point me this issue was that after some create/dispose a warning was pointing that there were too many webgl context created. 

I am still going to investigate this and will check memory snapshot after many create/dispose of a view (https://github.com/VCityTeam/UD-Imuv/issues/410).

jailln:
It doesn't fixes #1936 but it is still a good fix. Can you rename your commit to follow the [angular convention](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md) please ? Then we can merge

valentinMachado:
> It doesn't fixes #1936 but it is still a good fix. Can you rename your commit to follow the [angular convention](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md) please ? Then we can merge

it would have been too simple ^^
done :)

PR #2019, opened 2023-03-06
doc(tutorials): fix tutorials source API keys

AnthonyGlt:
## Description
Use IGN new webservices in the tutorials instead of API keys


gchoqueux (approved):
Thanks!

PR #2032, opened 2023-03-13
feature(Parser): detect if original source have elevation data.

gchoqueux:
## Description
Parser detect if original source have elevation data.

## Motivation and Context
After parsering, there is no way to have this information.
We will need it for the label refactoring

### DIFF
--- a/src/Core/Feature.js
+++ b/src/Core/Feature.js
@@ -223,6 +223,8 @@ function push3DValues(value0, value1, value2 = 0) {
  * @property {number[]} normals - All the normals of the Feature.
  * @property {number} size - the number of values of the array that should be associated with a coordinates.
  * The size is 3 with altitude and 2 without altitude.
+ * @property {boolean} hasRawElevationData - indicates if the geographic coordinates, from original source, has an elevation,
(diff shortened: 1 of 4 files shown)


src/Core/Feature.js
```diff
@@ -223,6 +223,7 @@ function push3DValues(value0, value1, value2 = 0) {
  * @property {number[]} normals - All the normals of the Feature.
  * @property {number} size - the number of values of the array that should be associated with a coordinates.
  * The size is 3 with altitude and 2 without altitude.
+ * @property {boolean} hasRawElevationData - indicates if the raw data,from original source, has elevation.
```
jailln: I would add `(i.e. is 3D)` at the end of the comment to be even clearer

gchoqueux: or a formulation more sig
`the geographic coordinates, from original source, have an elevation`

jailln: yes why not but I would still add explicitly something like "is 3D" or "has a third coordinate"

gchoqueux: ok 👍🏽

jailln (approved):

PR #2035, opened 2023-03-13
refactor(TileMesh): add change visibility event.

gchoqueux:
## Description
Add change visibility event in `TileMesh`.
For future use, in nodal architecture, in particular for linked/attached layer.

### DIFF
--- a/src/Core/TileMesh.js
+++ b/src/Core/TileMesh.js
@@ -14,6 +14,7 @@ import { geoidLayerIsVisible } from 'Layer/GeoidLayer';
  */
 class TileMesh extends THREE.Mesh {
     #_tms = new Map();
+    #visible = true;
     constructor(geometry, material, layer, extent, level = 0) {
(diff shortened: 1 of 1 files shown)


src/Core/TileMesh.js
```diff
@@ -14,6 +14,7 @@ import { geoidLayerIsVisible } from 'Layer/GeoidLayer';
  */
 class TileMesh extends THREE.Mesh {
     #_tms = new Map();
+    #visible = true;
```
jailln, line 17: why make it private if it can be get and set ?

gchoqueux: to avoid `this._visible and this.visible`

jailln: oh ok I thought we could write getters on `this.visible` directly but apparently not

jailln (approved):

jailln (approved):

PR #2036, opened 2023-03-13
refactor(RasterTile): emit event when raster elevation changed.

gchoqueux:
## Description
Emit event when raster elevation changed and not only when the max/min have changed

## Motivation and Context
More meaning and genericity  to the event.

### DIFF
--- a/src/Renderer/RasterTile.js
+++ b/src/Renderer/RasterTile.js
@@ -166,14 +166,22 @@ export class RasterElevationTile extends RasterTile {
     }
 
     initFromParent(parent, extents) {
+        const currentLevel = this.level;
         super.initFromParent(parent, extents);
         this.updateMinMaxElevation();
+        if (currentLevel !== this.level) {
+            this.dispatchEvent({ type: 'rasterElevationLevelChanged', node: this });
+        }
     }
 
     setTextures(textures, offsetScales) {
+        const currentLevel = this.level;
         this.replaceNoDataValueFromTexture(textures[0]);
         super.setTextures(textures, offsetScales);
         this.updateMinMaxElevation();
+        if (currentLevel !== this.level) {
+            this.dispatchEvent({ type: 'rasterElevationLevelChanged', node: this });
+        }
     }
 
     updateMinMaxElevation() {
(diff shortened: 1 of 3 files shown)


jailln:
It's clearer this way. Shouldn't we export it while we're at it ? It may be useful to some users

src/Renderer/RasterTile.js
```diff
@@ -166,14 +166,22 @@ export class RasterElevationTile extends RasterTile {
...
+        }
     }
 
     setTextures(textures, offsetScales) {
+        const currentLevel = this.level;
         this.replaceNoDataValueFromTexture(textures[0]);
         super.setTextures(textures, offsetScales);
         this.updateMinMaxElevation();
+        if (currentLevel !== this.level) {
+            this.dispatchEvent({ type: 'rasterElevationChanged', node: this });
```
jailln: Rename to `rasterElevationLevelChanged` to be more specific?

src/Renderer/RasterTile.js
```diff
@@ -182,7 +190,6 @@ export class RasterElevationTile extends RasterTile {
             if (this.min != min || this.max != max) {
                 this.min = min;
                 this.max = max;
-                this.dispatchEvent({ type: 'updatedElevation', node: this });
```
jailln, line 185: Maybe we should leave it, rename it to `minMaxElevationChanged`, send the min and max with it and export it? WDYT?

src/Renderer/RasterTile.js
gchoqueux: yes

src/Renderer/RasterTile.js
gchoqueux: I find it duplicates the other event (`rasterElevationChanged`) because the only time where the min/max changes is when the texture level changes

test/unit/tilemesh.js
```diff
@@ -177,20 +177,20 @@ describe('TileMesh', function () {
     material.addLayer = () => {};
     material.setSequenceElevation = () => {};
 
-    it('event updatedElevation RasterElevationTile sets TileMesh bounding box ', () => {
+    it('event rasterElevationChanged RasterElevationTile sets TileMesh bounding box ', () => {
```
jailln: don't forget to change the tests with `rasterElevationLevelChanged`

src/Renderer/RasterTile.js
jailln: ok

jailln (approved):

Issue #2011, opened 2023-02-28
Zoom in and out at the cursor position instead of the center of the canvas


LoicUTC:
I propose to develop the ability to zoom in and out from the cursor position because is more common (OpenLayersn Cesium) and seems more natural to me than from the center of the screen.

alexLuky:
The same for the camera rotation.... Current ORBIT rotation always around the screen center. 

Another not usefull behavior it's when you're near the ground and trying to pan a little forward you can catch the point that is near the horizont and camera flys away to the horizont. It happens because of raycast to the Globe mesh.

PR #2041, opened 2023-03-20
feat(globeControls): zoom on mouse position while using wheel

AnthonyGlt:
## Description
In globe view, zoom to the cursor position while using mousewheel.

## Motivation and Context
Linked issue: https://github.com/iTowns/itowns/issues/2011

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -159,7 +159,7 @@ let previous;
  * @param      {object}  [options] An object with one or more configuration properties. Any property of GlobeControls
  * can be passed in this object.
  * @property      {number}  zoomFactor The factor the scale is multiplied by when dollying (zooming) in or
- * divided by when dollying out. Default is 2.
+ * divided by when dollying out. Default is 1.1.
  * @property      {number}  rotateSpeed Speed camera rotation in orbit and panoramic mode. Default is 0.25.
  * @property      {number}  minDistance Minimum distance between ground and camera in meters (Perspective Camera only).
  * Default is 250.
@@ -211,7 +211,7 @@ class GlobeControls extends THREE.EventDispatcher {
             console.warn('Controls zoomSpeed parameter is deprecated. Use zoomFactor instead.');
             options.zoomFactor = options.zoomFactor || options.zoomSpeed;
         }
-        this.zoomFactor = options.zoomFactor || 1.25;
+        this.zoomFactor = options.zoomFactor || 1.1;
 
         // Limits to how far you can dolly in and out ( PerspectiveCamera only )
         this.minDistance = options.minDistance || 250;
@@ -268,7 +268,7 @@ class GlobeControls extends THREE.EventDispatcher {
         this.updateHelper = enableTargetHelper ? (position, helper) => {
             positionObject(position, helper);
             view.notifyChange(this.camera);
-        } : function empty() {};
+        } : function empty() { };
 
         this._onEndingMove = null;
         this._onTravel = this.travel.bind(this);
@@ -323,10 +323,10 @@ class GlobeControls extends THREE.EventDispatcher {
         coordCameraTarget.crs = this.view.referenceCrs;
     }
 
-    get dollyInScale() {
+    get zoomInScale() {
         return this.zoomFactor;
     }
-    get dollyOutScale() {
+    get zoomOutScale() {
         return 1 / this.zoomFactor;
     }
 
@@ -389,9 +389,10 @@ class GlobeControls extends THREE.EventDispatcher {
         }
     }
 
+    // For Mobile
     dolly(delta) {
         if (delta === 0) { return; }
-        dollyScale = delta > 0 ? this.dollyInScale : this.dollyOutScale;
+        dollyScale = delta > 0 ? this.zoomInScale : this.zoomOutScale;
 
         if (this.camera.isPerspectiveCamera) {
             orbitScale /= dollyScale;
@@ -454,7 +455,8 @@ class GlobeControls extends THREE.EventDispatcher {
                 quaterPano.setFromAxisAngle(normal, sphericalDelta.theta).multiply(quaterAxis.setFromAxisAngle(axisX, sphericalDelta.phi));
                 cameraTarget.position.applyQuaternion(quaterPano);
                 this.camera.localToWorld(cameraTarget.position);
-                break; }
+                break;
+            }
             // ZOOM/ORBIT Move Camera around the target camera
             default: {
                 // get camera position in local space of target
@@ -594,6 +596,7 @@ class GlobeControls extends THREE.EventDispatcher {
 
         // Initialize dolly movement.
         dollyStart.copy(event.viewCoords);
+        this.view.getPickingPositionFromDepth(event.viewCoords, pickedPosition);        // mouse position
 
         // Initialize pan movement.
         panStart.copy(event.viewCoords);
@@ -629,11 +632,9 @@ class GlobeControls extends THREE.EventDispatcher {
     handleDolly(event) {
         dollyEnd.copy(event.viewCoords);
         dollyDelta.subVectors(dollyEnd, dollyStart);
-
-        this.dolly(-dollyDelta.y);
         dollyStart.copy(dollyEnd);
-
-        this.update();
+        event.delta = dollyDelta.y;
+        if (event.delta != 0) { this.handleZoom(event); }
     }
 
     handlePan(event) {
@@ -764,23 +765,31 @@ class GlobeControls extends THREE.EventDispatcher {
     handleZoom(event) {
         this.player.stop();
         CameraUtils.stop(this.view, this.camera);
+        const zoomScale = event.delta > 0 ? this.zoomInScale : this.zoomOutScale;
+        let point = event.type === 'dolly' ? pickedPosition : this.view.getPickingPositionFromDepth(event.viewCoords);        // get cursor position
+        let range = this.getRange();
+        range *= zoomScale;
 
-        this.updateTarget();
-        const delta = -event.delta;
-        this.dolly(delta);
+        if (point && (range > this.minDistance && range < this.maxDistance)) {  // check if the zoom is in the allowed interval
+            const camPos = xyz.setFromVector3(cameraTarget.position).as('EPSG:4326', c).toVector3();
+            point = xyz.setFromVector3(point).as('EPSG:4326', c).toVector3();
 
-        const previousRange = this.getRange(pickedPosition);
-        this.update();
-        const newRange = this.getRange(pickedPosition);
-        if (Math.abs(newRange - previousRange) / previousRange > 0.001) {
-            this.dispatchEvent({
-                type: CONTROL_EVENTS.RANGE_CHANGED,
-                previous: previousRange,
-                new: newRange,
-            });
+            if (camPos.x * point.x < 0) {     // Correct rotation at 180th meridian by using 0 <= longitude <=360 for interpolation purpose
+                if (camPos.x - point.x > 180) { point.x += 360; } else if (point.x - camPos.x > 180) { camPos.x += 360; }
+            }
+            point.lerp(  // point interpol between mouse cursor and cam pos
+                camPos,
+                zoomScale, // interpol factor
+            );
+
+            point = c.setFromVector3(point).as('EPSG:4978', xyz);
+
+            return this.lookAtCoordinate({       // update view to the interpolate point
+                coord: point,
+                range,
+            },
+            false);
         }
-        this.dispatchEvent(this.startEvent);
-        this.dispatchEvent(this.endEvent);
     }
 
     onTouchStart(event) {
@@ -804,7 +813,8 @@ class GlobeControls extends THREE.EventDispatcher {
                     } else {
                         this.state = this.states.NONE;
                     }
-                    break; }
+                    break;
+                }
                 case this.states.ORBIT:
                 case this.states.DOLLY: {
                     const x = event.touches[0].pageX;
@@ -814,7 +824,8 @@ class GlobeControls extends THREE.EventDispatcher {
                     const distance = Math.sqrt(dx * dx + dy * dy);
                     dollyStart.set(0, distance);
                     rotateStart.set(x, y);
-                    break; }
+                    break;
+                }
                 case this.states.PAN:
                     panStart.set(event.touches[0].pageX, event.touches[0].pageY);
                     break;
@@ -851,7 +862,8 @@ class GlobeControls extends THREE.EventDispatcher {
                 } else {
                     this.onTouchEnd();
                 }
-                break; }
+                break;
+            }
             case this.states.ORBIT.finger:
             case this.states.DOLLY.finger: {
                 const gfx = this.view.mainLoop.gfxEngine;
@@ -875,7 +887,8 @@ class GlobeControls extends THREE.EventDispatcher {
 
                 dollyStart.copy(dollyEnd);
 
-                break; }
+                break;
+            }
             case this.states.PAN.finger:
                 panEnd.set(event.touches[0].pageX, event.touches[0].pageY);
                 panDelta.subVectors(panEnd, panStart);
--- a/src/Controls/StateControl.js
+++ b/src/Controls/StateControl.js
@@ -396,6 +396,7 @@ class StateControl extends THREE.EventDispatcher {
 
         if (this.enabled && this.ZOOM.enable) {
             viewCoords.copy(this._view.eventToViewCoords(event));
+            this.currentState = this.NONE;
             this.dispatchEvent({ type: this.ZOOM._event, delta: event.deltaY, viewCoords });
         }
     }
--- a/src/Controls/StateControl.js
+++ b/src/Controls/StateControl.js
@@ -395,7 +395,8 @@ class StateControl extends THREE.EventDispatcher {
         event.preventDefault();
 
         if (this.enabled && this.ZOOM.enable) {
-            this.dispatchEvent({ type: this.ZOOM._event, delta: event.deltaY });
+            viewCoords.copy(this._view.eventToViewCoords(event));
+            this.dispatchEvent({ type: this.ZOOM._event, delta: event.deltaY, viewCoords });
         }
     }
 


ftoromanoff:
From what i understand of this PR, the purpose is that when you zoom in or out, the terrain under the mouse should always stay the same.

I tried this PR on a globe view. I noticed that zooming in on the mouse position seems to work when you are at close range but not at all when you are far enought and the zoom out never stay at the mouse position.
I havn't really gone into the code yet, and from my first test, the aim is not achived.

AnthonyGlt:
Thanks for the feedback.

I use zoom coordinates interpolated between the mouse position and the camera position that's why we don't go directly to the mouse position. This allows us to have a smoother transition. The behaviour is similar to the globe view of Google Maps. This _**alpha**_ parameter (interpolation factor) is fixed but depending on feedback, could be edited before release. Concerning the zoom out, the factor is different (**_dezoom_interp_**) to avoid brutal rotation while zooming out. Same as alpha, this could be edited before release. I could even make these values as parameters of globecontrols, then the user could decide of the zoom behaviour.

alexLuky:
@ftoromanoff @AnthonyGlt @jailln 
By the way, take a look at this [control](https://github.com/yomotsu/camera-controls). It has all useful functions. 
I've mentioned it here in another [discussion](https://github.com/iTowns/itowns/discussions/2075#discussioncomment-5678605). 

I think it would be great to integrate it in iTowns. Also for navigation under the ground or in hidden globe mode.

gchoqueux:
With your PR, the target position doesn't stay under the mouse cursor.

https://user-images.githubusercontent.com/11291849/233647952-99628964-9631-4e0f-8397-aad7f396ccaa.mp4

In the same condition with google earth, the target position stay under the cursor and doesn't come out of the screen.

https://user-images.githubusercontent.com/11291849/233649762-fdc19375-c9dc-4198-8ba5-758b611694a8.mp4

AnthonyGlt:
I was referring to Google Maps (cf attached record). If this isn't an appropriate solution, an alternative could be implemented to have a similar zoom as Google Earth instead. Currenctly, I used a bit of a mix of the zoom function from the _planarControls_ (interpolation) and the double click zoom from the _globeControls_ (**travel** -> **lookAtCoordinate**).


https://user-images.githubusercontent.com/126568810/234286574-3025cbcd-d436-4035-98c6-a20d39f43a27.mp4

gchoqueux:
it's good feature that improves the user experience.
I'm asking if it would be possible to factor your code for use in the `handleDolly(event)` method (call by pressing the middle button)

AnthonyGlt:
> it's good feature that improves the user experience. I'm asking if it would be possible to factor your code for use in the `handleDolly(event)` method (call by pressing the middle button)

It's now implemented. I store the coordinates when the wheel button is clicked to zoom at the precise mouse location. But it's not really a "dolly" anymore.
In a future improvement, we could  also store the first coordinates when the mouse wheel is used, to zoom to the exact mouse location. I did it on another branch of my repo. I did a draft PR on it: https://github.com/iTowns/itowns/pull/2091




https://user-images.githubusercontent.com/126568810/236457349-444ab904-15cd-4567-8309-7e53c7891212.mp4

src/Controls/GlobeControls.js
```diff
@@ -624,14 +626,19 @@ class GlobeControls extends THREE.EventDispatcher {
         }
     }
 
+
     handleDolly(event) {
+        if (!dollyOrigin) {
+            dollyOrigin = this.view.getPickingPositionFromDepth(event.viewCoords);        // mouse position
```
gchoqueux: could move to `onStateChange` method?

src/Controls/GlobeControls.js
```diff
@@ -759,26 +768,41 @@ class GlobeControls extends THREE.EventDispatcher {
...
+    handleZoom(event, dollyOrigin) {
         this.player.stop();
         CameraUtils.stop(this.view, this.camera);
 
-        this.updateTarget();
-        const delta = -event.delta;
-        this.dolly(delta);
+        var point = dollyOrigin ?? this.view.getPickingPositionFromDepth(event.viewCoords);        // mouse position
+        this.view.getPickingPositionFromDepth(null, pickedPosition);
+        var range = this.getRange(pickedPosition);
```
gchoqueux: Could you use `cameraTarget.position` instead of `getPickingPositionFromDepth` calling 
`cameraTarget.position` is camera target position on the globe.

if you calculate the length between camera and the target, the result is directly the range

src/Controls/GlobeControls.js
```diff
@@ -759,26 +768,41 @@ class GlobeControls extends THREE.EventDispatcher {
...
-        if (Math.abs(newRange - previousRange) / previousRange > 0.001) {
-            this.dispatchEvent({
-                type: CONTROL_EVENTS.RANGE_CHANGED,
-                previous: previousRange,
-                new: newRange,
-            });
+        if (point && (range > this.minDistance && range < this.maxDistance)) {  // check if the zoom is in the allowed interval
+            const camPos = this.getLookAtCoordinate().toVector3();
+
+            point = new Coordinates('EPSG:4978', point).as('EPSG:4326').toVector3();
```
gchoqueux: you could use `c` or `xyz` to avoid `Coordinates` instance

src/Controls/GlobeControls.js
```diff
@@ -759,26 +768,41 @@ class GlobeControls extends THREE.EventDispatcher {
...
+            if (camPos.x * point.x < 0) {     // Correct rotation at 180th meridian by using 0 <= longitude <=360
+                if (camPos.x - point.x > 180) { point.x += 360; } else if (point.x - camPos.x > 180) { camPos.x += 360; }
+            }
```
gchoqueux: It's a `lookAtCoordinate` issue, it's not support a correct rotation at 180th meridian by using `0 <= longitude <=360`?

src/Controls/GlobeControls.js
```diff
@@ -759,26 +768,41 @@ class GlobeControls extends THREE.EventDispatcher {
...
+
+            point.lerp(  // point interpol between mouse cursor and cam pos
+                camPos,
+                (event.delta > 0 ? this.zoomInScale : this.zoomOutScale), // interpol factor
+
+            );
+            if (point.x > 180) {  // Correct rotation at 180th meridian
+                point.x -= 360;
+            }
```
gchoqueux: It's a `lookAtCoordinate` issue?

src/Controls/GlobeControls.js
```diff
@@ -759,26 +768,41 @@ class GlobeControls extends THREE.EventDispatcher {
...
+            point = new Coordinates('EPSG:4326', point).as('EPSG:4978').toVector3();
+
+            pickedPosition.copy(point);
+
+            return this.lookAtCoordinate({       // update view to the interpolate point
+                coord: new Coordinates('EPSG:4978', point),
```
gchoqueux: `point` is already in  EPSG:4978?

gchoqueux (commented):
Your PR is good way to add this feature
In the future we will have to think about how to solve the issue of the pointer shift when zooming.
This issue is accentuated when the `zoomFactor` is increased

src/Controls/GlobeControls.js
AnthonyGlt: No it's for the lerp method (interpolation). At the 180th meridian, we have values going from -179 to 179 (-179, -180/180, 179). So when interpolating a value between, for example, -170 and 172, we will obtain a value of **1** instead of the expecting **-179**. 
To avoid that, I add 360 (to keep the same value modulo 360) and interpolate on positive values only. Then I substract the 360 to get the value back between -180 and 180.

src/Controls/GlobeControls.js
AnthonyGlt: Looks like it's not necessary to substract 360. I'll remove this part of the code then

src/Controls/GlobeControls.js
```diff
@@ -593,6 +595,11 @@ class GlobeControls extends THREE.EventDispatcher {
         // Initialize dolly movement.
         dollyStart.copy(event.viewCoords);
 
+        // Store first position of mouse when dolly
+        if (!dollyOrigin) {
+            dollyOrigin = this.view.getPickingPositionFromDepth(event.viewCoords);        // mouse position
+        }
+
```
gchoqueux: could you use `pickedPosition` instead of `dollyOrigin`, it seems to have a duplicate use

src/Controls/GlobeControls.js
```diff
@@ -759,26 +770,36 @@ class GlobeControls extends THREE.EventDispatcher {
...
 
-    handleZoom(event) {
+    handleZoom(event, dollyOrigin) {
         this.player.stop();
         CameraUtils.stop(this.view, this.camera);
 
-        this.updateTarget();
-        const delta = -event.delta;
-        this.dolly(delta);
+        var point = dollyOrigin ?? this.view.getPickingPositionFromDepth(event.viewCoords);        // mouse position
```
gchoqueux: There's a case where the `dollyOrigin` is `null`

src/Controls/GlobeControls.js
```diff
@@ -759,26 +770,36 @@ class GlobeControls extends THREE.EventDispatcher {
         }
     }
 
-    handleZoom(event) {
+    handleZoom(event, dollyOrigin) {
```
gchoqueux: I can't remember if there's a change of state for the zoom?
to handle the state zoom

src/Controls/GlobeControls.js
```diff
@@ -759,26 +770,36 @@ class GlobeControls extends THREE.EventDispatcher {
...
-        const previousRange = this.getRange(pickedPosition);
-        this.update();
-        const newRange = this.getRange(pickedPosition);
-        if (Math.abs(newRange - previousRange) / previousRange > 0.001) {
-            this.dispatchEvent({
-                type: CONTROL_EVENTS.RANGE_CHANGED,
-                previous: previousRange,
-                new: newRange,
-            });
+        range *= (event.delta > 0 ?  this.zoomInScale :  this.zoomOutScale);
```
gchoqueux: could you encapsulate `(event.delta > 0 ? this.zoomInScale : this.zoomOutScale);` to re-use it?

src/Controls/GlobeControls.js
gchoqueux: could you rename `point` with explicit name, `pickedPoint`?

src/Controls/GlobeControls.js
AnthonyGlt: no state change for the zoom. I need to disguish between a regular zoom and a dolly one. Maybe I could 
```suggestion
    handleZoom(event, dollyFlag) {
```
with `dollyFlag` used to know if I need to update `pickedPosition` value

src/Controls/GlobeControls.js
AnthonyGlt: Yes you're right, I could use `pickedPosition`. 

`        this.view.getPickingPositionFromDepth(event.viewCoords, pickedPosition);        // mouse position
`

But I need something to know if I'm doing a dolly (cf next comment)

src/Controls/GlobeControls.js
AnthonyGlt: If dolly, then I don't update `pickedPoint` value. So I can use a boolean instead of `dollyOrigin`. What do you think ?
```suggestion
        var pickedPoint = dollyFlag ? pickedPosition :  this.view.getPickingPositionFromDepth(event.viewCoords);        // mouse position
```

src/Controls/GlobeControls.js
AnthonyGlt: You mean in a function ? Like: 
`

    #zoomScale(delta) {
        return delta > 0 ?  this.zoomInScale :  this.zoomOutScale;
    }`

and then : 
 ```suggestion
         range *= this.#zoomScale(event.delta);
```

File: src/Controls/GlobeControls.js (hunk 7 above)
Comment (gchoqueux) [reply src/Controls/GlobeControls.js]: `StateControl` should handle dolly state?

File: src/Controls/GlobeControls.js (hunk 9 above)
Comment (gchoqueux) [reply src/Controls/GlobeControls.js]: > no state change for the zoom. I need to disguish between a regular zoom and a dolly one. Maybe I could

This would seem to confirm that `StateControl` should manage its two states?

File: src/Controls/GlobeControls.js (hunk 10 above)
Comment (gchoqueux) [reply src/Controls/GlobeControls.js]: or in a constant

File: src/Controls/GlobeControls.js (hunk 9 above)
Comment (AnthonyGlt) [reply src/Controls/GlobeControls.js]: in my last commit, i've used directly the event type to distinguish between the 2 zoom

**Comment by AnthonyGlt** [comment]
TODO edit tests

**Comment by AnthonyGlt** [comment]
@Desplandis rebased :+1:

**Comment by AnthonyGlt** [comment]
@Desplandis rebased :+1:

File: src/Controls/GlobeControls.js [hunk 11]
```diff
@@ -762,23 +765,32 @@ class GlobeControls extends THREE.EventDispatcher {
...
-                type: CONTROL_EVENTS.RANGE_CHANGED,
-                previous: previousRange,
-                new: newRange,
-            });
+            if (camPos.x * point.x < 0) {     // Correct rotation at 180th meridian by using 0 <= longitude <=360 for interpolation purpose
+                if (camPos.x - point.x > 180) { point.x += 360; } else if (point.x - camPos.x > 180) { camPos.x += 360; }
+            }
+            point.lerp(  // point interpol between mouse cursor and cam pos
+                camPos,
+                (event.delta > 0 ? this.zoomInScale : this.zoomOutScale), // interpol factor
```
mgermerie: You could directly use `zoomScale` defined on line 768 instead of having to check `event.delta` again.

src/Controls/GlobeControls.js
```diff
@@ -762,23 +765,32 @@ class GlobeControls extends THREE.EventDispatcher {
...
         CameraUtils.stop(this.view, this.camera);
+        const zoomScale = event.delta > 0 ? this.zoomInScale : this.zoomOutScale;
+        let point = event.type === 'dolly' ? pickedPosition : this.view.getPickingPositionFromDepth(event.viewCoords);        // get cursor position
+        let range = this.camera.position.distanceTo(cameraTarget.position);
+        range *= zoomScale;
 
-        this.updateTarget();
-        const delta = -event.delta;
-        this.dolly(delta);
+        if (point && (range > this.minDistance && range < this.maxDistance)) {  // check if the zoom is in the allowed interval
```
mgermerie, line 773: There is a problem with the distance check : if I log some message inside this `if` block, it will always log event if I am visually at the minimum distance (the camera doesn't seam to move toward the floor).
This causes the camera to move laterally (but no longer vertically) when continuing to zoom in with scroll, which is a weird behavior in my opinion.

mgermerie: Perhaps this is caused by an issue in `range` calculation since when logging `range` in this situation, I get values that are way higher than `minDistance` (`680m` for a minimum distance of `250m`) and the camera still doesn't get closer to ground.

src/Controls/GlobeControls.js
```diff
@@ -624,14 +627,13 @@ class GlobeControls extends THREE.EventDispatcher {
         }
     }
 
+
```
mgermerie: Can you remove this empty line addition ?

src/Controls/GlobeControls.js
```diff
@@ -665,6 +667,7 @@ class GlobeControls extends THREE.EventDispatcher {
     handleEndMovement(event = {}) {
         this.dispatchEvent(this.endEvent);
 
+
```
mgermerie: Can you remove this empty line addition as well ?

mgermerie (commented):
Hi ! Thanks for the improvement on zoom ! There is just one functional issue and a few remarks on form listed in my review.
Also, when you'll rebase your commits, could you add a breaking change note in the commit message ?

src/Controls/GlobeControls.js
AnthonyGlt: Yes you are right, it seems there is issue at high zoom.
I've just did some change and it looks better.
I'm now using the `getRange` function.
It's been so long, I don't know why I didn't used it at first :thinking: but it looks better like this.

mgermerie (commented):
Thanks for your commit that indeed fixes the issue when zooming over `minDistance` value.

Trying out the example again, I came upon a glitch that occurs when zooming with the mouse wheel while orbiting (with `ctrl+drag`). On `master` branch, one can do so as you can test on [`view_3d_map` example](https://www.itowns-project.org/itowns/examples/view_3d_map.html)

When doing the same on your branch, I noticed that the camera moves strangely. Moreover, if I keep orbiting after having rolled the mouse wheel, the camera zooms out. The zoom should not be affected by an orbit movement.

I found out this is due to [this line](https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Controls/GlobeControls.js#L1186) in `lookAtCoordinates` method which you call in your new `handleZoom`. It modifies `cameraTarget` which is used as a reference in `handlePanoramic`, thus for orbit movement. Removing this line fixes the issue I mention but brings out some more issues, notably with "dolly" movement (`wheel-click + drag`)

AnthonyGlt:
@mgermerie we could just cancel the Panoramic once the zoom is triggered (looks like how it's handled in Google Earth), WDYT ?
By adding
 `this.currentState = this.NONE;` 
before
 `this.dispatchEvent({ type: this.ZOOM._event, delta: event.deltaY, viewCoords });`

in

https://github.com/iTowns/itowns/blob/8de37c94ff839baf8756a336954b16137def8d60/src/Controls/StateControl.js#L394-L401

Desplandis:
@AnthonyGlt I agree with you on that point, we should disable this behavior (is this really used BTW?) to mimic the one in Earth and to decrease the complexity of state handling.

@mgermerie Any thoughts?

AnthonyGlt:
@Desplandis squashed :tennis:

mgermerie:
@AnthonyGlt Your proposition is ok for me in terms of how to code it.
Now, I personally would not be so found of deleting it. The behavior already exists in master branch, and is implemented in Google Maps as well. However, this is a personal preference and I don't know if this behavior is used in iTowns so if you decide to remove it, it is ok for me as long as you document a breaking change.

AnthonyGlt:
> @AnthonyGlt Your proposition is ok for me in terms of how to code it. Now, I personally would not be so found of deleting it. The behavior already exists in master branch, and is implemented in Google Maps as well. However, this is a personal preference and I don't know if this behavior is used in iTowns so if you decide to remove it, it is ok for me as long as you document a breaking change.

@mgermerie I've added the breaking change to the commit

Thank you for your review

mgermerie (approved):

PR #2042, opened 2023-03-20
add classification for 3D tiles point cloud

AnthonyGlt:
## Description
Use classification for attributing colors to pointscloud
Fix batchtable reading using the correct batch size

Example using data from https://github.com/iTowns/iTowns2-sample-data/pull/13 

## Screenshots
Capture d'écran 2023-03-15 123524

### DIFF
--- a/examples/3dtiles_pointcloud.html
+++ b/examples/3dtiles_pointcloud.html
@@ -1,0 +1,102 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Itowns - Pointcloud classification</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="description">
+            Pointcloud classification
+            <ul>
+              <li>Display your pointcloud</li>
+              <li>Use classification to assign colour to each points</li>
+              <li>Switch between mode on the left panel</li>             
+            </ul>
+        </div>
+        <div id="viewerDiv"></div>
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script src="js/3dTilesHelper.js"></script>
+        <script type="text/javascript">
+            /* global itowns,document,GuiTools*/
+
+            
+            var placement = {
+                coord: new itowns.Coordinates('EPSG:4326', 3.695885, 43.397379, 0),
+
+                range: 3000,
+                tilt: 55,
+                heading: 180
+            }
+            // iTowns namespace defined here
+            var viewerDiv = document.getElementById('viewerDiv');
+
+            var view = new itowns.GlobeView(viewerDiv, placement);
+            view.camera.camera3D.near = 5;
+            setupLoadingScreen(viewerDiv, view);
+
+            var menuGlobe = new GuiTools('menuDiv', view, 300);
+
+            itowns.Fetcher.json('./layers/JSONLayers/OPENSM.json').then(function _(config) {
+                config.source = new itowns.TMSSource(config.source);
+                var layer = new itowns.ColorLayer('Ortho', config);
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            });
+            function updatePointCloudSize({tileContent}) {
+                tileContent.traverse(function (obj) {
+                    if (obj.isPoints) {
+                        obj.material.size = 2.0;
+                    }
+                });
+            }
+            // Create a new Layer 3d-tiles For Pointcloud
+            // -------------------------------------------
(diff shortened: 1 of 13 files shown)


src/Parser/PntsParser.js
```diff
@@ -66,8 +66,11 @@ export default {
             if (pntsHeader.BTJSONLength > 0) {
                 // parse batch table
                 const sizeBegin = byteOffset + pntsHeader.FTJSONLength + pntsHeader.FTBinaryLength;
-                const BTBuffer = buffer.slice(sizeBegin, pntsHeader.BTJSONLength + sizeBegin);
-                batchTable = new C3DTBatchTable(BTBuffer, pntsHeader.BTJSONLength, pntsHeader.BTBinaryLength, FTJSON.BATCH_LENGTH, registeredExtensions);
+                const BTBuffer = buffer.slice(sizeBegin, pntsHeader.BTJSONLength + pntsHeader.BTBinaryLength + sizeBegin);
+
+                // If the BATCH_ID semantic is not defined, then the Batch Table stores per-point metadata, and the length of the Batch Table arrays will equal POINTS_LENGTH.
+                batchTable = new C3DTBatchTable(BTBuffer, pntsHeader.BTJSONLength, pntsHeader.BTBinaryLength, FTJSON.BATCH_LENGTH ?? FTJSON.POINTS_LENGTH, registeredExtensions);
```
mgermerie: If I understood well the spec regarding this `BATCH_ID` matter, wouldn't it be more rigourous to replace 
```js
FTJSON.BATCH_LENGTH ?? FTJSON.POINTS_LENGTH
```
by
```js
(FTJSON.BATCH_ID && FTJSON.BATCH_LENGHT) || FTJSON.POINTS_LENGTH
```
?

I get that the first solution is quicker in terms of coding, but the second is perhaps more robust to errors in the data - such as a dataset with undefined `BATH_ID`, but a wrong non null value for `BATCH_LENGTH`. 
WDYT ?

src/Layer/C3DTilesLayer.js
```diff
@@ -64,6 +66,16 @@ class C3DTilesLayer extends GeometryLayer {
         this.name = config.name;
         this.registeredExtensions = config.registeredExtensions || new C3DTExtensions();
 
+        this.pntsMode = PNTS_MODE.COLOR;
+        this.classification =  config.classification;
```
mgermerie: It's nothing but you have an extra space after the `=`. I thought the linter would track this but it seems not

mgermerie:
Thanks for this PR ! Could you please add an example with the dataset in your screenshot (if you can make it public, otherwise we would have to find another classified dataset) ?

AnthonyGlt:
@mgermerie  Do you have any other requirements about this feature ?

examples/3dtiles_pointcloud.html
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                });
+            }
+            // Create a new Layer 3d-tiles For Pointcloud
+            // -------------------------------------------
+            var $3dTilesLayerSetePC = new itowns.C3DTilesLayer('3d-tiles-sete', {
+                name: 'SetePC',
+                sseThreshold: 5,
+                pntsMode: itowns.PNTS_MODE.CLASSIFICATION,
+                onTileContentLoaded: updatePointCloudSize,
```
mgermerie: The use of this parameter has been deprecated in #2060. Because of what is described in #2113, the callback would not work anymore. Could you replace it with : 
```js
$3dTilesLayerSetePC.addEventListener(
    itowns.C3DTILES_LAYER_EVENTS.ON_TILE_CONTENT_LOADED,
    updatePointCloudSize,
);
```

examples/3dtiles_pointcloud.html
```diff
@@ -0,0 +1,103 @@
...
+            setupLoadingScreen(viewerDiv, view);
+
+            var menuGlobe = new GuiTools('menuDiv', view, 300);
+
+            itowns.Fetcher.json('./layers/JSONLayers/OPENSM.json').then(function _(config) {
+                config.source = new itowns.TMSSource(config.source);
+                var layer = new itowns.ColorLayer('Ortho', config);
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            });
+            function updatePointCloudSize(tileContent) {
```
mgermerie: Due to changes in #2060, the signature of this method should be changed to `updatePointCloudSize({ tileContent })`

mgermerie:
It's a good thing that `3DTilesLayers` now use the same material for points as `PointCloudLayer`, that will simplify development point clouds visualization, however their format.
I believe @jailln did some benchmark at some time and found out that 3dtiles point cloud visualization was quite performant. Would you be able to test with these benchmarks if there is an impact on performance to switch from `THREE.PointMaterial` to our custom one ?

AnthonyGlt:
> It's a good thing that `3DTilesLayers` now use the same material for points as `PointCloudLayer`, that will simplify development point clouds visualization, however their format. I believe @jailln did some benchmark at some time and found out that 3dtiles point cloud visualization was quite performant. Would you be able to test with these benchmarks if there is an impact on performance to switch from `THREE.PointMaterial` to our custom one ?

I didn't see bad impact on the performances. By doing some tests and time measurements, it looked like it was faster to display the points.

AnthonyGlt:
Before Merging, I need to edit the data url in the example, see this related PR : https://github.com/iTowns/iTowns2-sample-data/pull/13

mgermerie (approved):

mgermerie:
Ok thanks !
Could you just either squash your second and third commit before merging ? Or else remove the example from the second commit to keep it only in the third ?

mgermerie (approved):

PR #2044, opened 2023-03-21
refact(test): add message when test are failing (and no timeout)

### LINKED ISSUE #2057: Unit test and wrong message for failing tests
Currently it's not always easy to understand why a specifique test is failing as the message recived is not directly link with the failure.
And this for 2 reasons:

1) In the unit test files we use 'assert.equal()' that send an error if it's not true but the error is not catched thus we got, if if it tested inside a promise:
> Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

2) itowns handles itself several errors to avoid crashing with 'Process/handlerNodeError.js'. In this file we can see that we get the error message only if we are in develpment mode.
```
         if (__DEBUG__) {
            if (layer.isColorLayer) {
                console.warn('Error in process color on layer', layer.id, ', node', node, err);
            } else if (layer.isElevationLayer) {
                console.warn('Error in process elevation on layer', layer.id, ', node', node, err);
            } else {
                console.warn('Error in process feature on layer', layer.id, ', node', node, err);
            }
        }
```
or during the execution of the test we are in production mode.

In the PR #2044 I propose some ways to improve that :
For 1) I had a catch catchback to send the good error message (and not waste time waiting for the timeout), i did it only for 1 file: test/unit/featuregeometrylayererror.js but it can be done in several other.

For 2) we don't really want to change the way that github action lauch the tests as we want them to be the closest to the production setup. But we can changed it during the previous step on the local setup, That's the reason why i added a new line in package.json to be able to lauch the test in development mode.

ftoromanoff:
Currently when some tests are failing, the process wait for time out to get to the next one, thus the failing message is not precis (error timeout) and time is wasted.

This PR aims at catching the real error to save time and get the correct error message.

Currently done for all unit tests.
But there is 1 test in cameraUtils.js which need to be rewriten:
```
    it('should heading, tilt, range and coordinate like expected with animation (200ms)', function (done) {
        const params = { heading: 17, tilt: 80, range: 20000, coord: coord.clone(), time: 200 };
        params.coord.setFromValues(params.coord.longitude + 3, params.coord.latitude + 4, 0);
        CameraUtils.animateCameraToLookAtTarget(view, camera, params)
            .then((result) => {
                // ne passe jamais ici...
                assert.ok(equalToFixed(result.heading, params.heading, 4));
                assert.ok(equalToFixed(result.tilt, params.tilt, 4));
                assert.ok(equalToFixed(result.range, params.range, 1));
                assert.ok(equalToFixed(result.coord.longitude, params.coord.longitude, 4));
                assert.ok(equalToFixed(result.coord.latitude, params.coord.latitude, 4));
                done();
            }, done);
    });
```

### DIFF
--- a/test/unit/cameraUtils.js
+++ b/test/unit/cameraUtils.js
@@ -114,19 +114,24 @@ describe('Camera utils unit test', function () {
                 done();
             }, done);
     });
+
+    // TODO: to verify and recode
+    /* the Promise animateCameraToLookAtTarget is never resolving...
     it('should heading, tilt, range and coordinate like expected with animation (200ms)', function (done) {
         const params = { heading: 17, tilt: 80, range: 20000, coord: coord.clone(), time: 200 };
         params.coord.setFromValues(params.coord.longitude + 3, params.coord.latitude + 4, 0);
         CameraUtils.animateCameraToLookAtTarget(view, camera, params)
             .then((result) => {
+                // we never pass here
                 assert.ok(equalToFixed(result.heading, params.heading, 4));
                 assert.ok(equalToFixed(result.tilt, params.tilt, 4));
                 assert.ok(equalToFixed(result.range, params.range, 1));
                 assert.ok(equalToFixed(result.coord.longitude, params.coord.longitude, 4));
                 assert.ok(equalToFixed(result.coord.latitude, params.coord.latitude, 4));
                 done();
-            }, done);
+            }, done);// neither here
     });
+    */
 
     it('should transform camera from given extent', function () {
         view.isPlanarView = true;
--- a/test/unit/cameraUtils.js
+++ b/test/unit/cameraUtils.js
@@ -56,62 +56,76 @@ describe('Camera utils unit test', function () {
     const range = 25000000;
     const coord = new Coordinates('EPSG:4326', 2.35, 48.85, 0);
 
-    it('init like expected', function () {
+    it('init like expected', function (done) {
         const params = { range, coord };
-        CameraUtils.transformCameraToLookAtTarget(view, camera, params).then((result) => {
-            assert.ok(equalToFixed(result.range, params.range, 1));
-            assert.ok(equalToFixed(result.coord.longitude, params.coord.longitude, 4));
-            assert.ok(equalToFixed(result.coord.latitude, params.coord.latitude, 4));
-        });
+        CameraUtils.transformCameraToLookAtTarget(view, camera, params)
+            .then((result) => {
+                assert.ok(equalToFixed(result.range, params.range, 1));
+                assert.ok(equalToFixed(result.coord.longitude, params.coord.longitude, 4));
+                assert.ok(equalToFixed(result.coord.latitude, params.coord.latitude, 4));
+                done();
+            }, done);
     });
-    it('should set range like expected', function () {
+    it('should set range like expected', function (done) {
         const params = { range: 10000 };
-        CameraUtils.transformCameraToLookAtTarget(view, camera, params).then((result) => {
-            const range = result.range;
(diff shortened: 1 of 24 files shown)


jailln (approved):

jailln:
Do you think you could do it for all tests as you said in your issue please?

Could you also add documentation about this new npm script in the [testing guide](https://github.com/iTowns/itowns/blob/master/CODING.md#Testing)?

Thanks :)

ftoromanoff:
@jailln 
I did what you adviced:
- issue open (#2057 )
- add the npm documentation
- and rewrite all test

There is one test that wasn't passing and that i couldn't resolve

jailln:
> @jailln I did what you adviced:
> 
> * issue open ([Unit test and wrong message for failing tests #2057](https://github.com/iTowns/itowns/issues/2057) )
> * add the npm documentation
> * and rewrite all test
> 
> There is one test that wasn't passing and that i couldn't resolve

Thanks!! This really improves itowns testing suite :) 
Everything seems ok for me except the last commit that should be renamed to something like `refactor(test): remove failing CameraUtils test` or squashed with the previous commit. Also, could you change the comments added in this commit from french to english please?

test/unit/cameraUtils.js
```diff
@@ -56,63 +56,82 @@ describe('Camera utils unit test', function () {
...
+                // ne passe jamais ici...
+                assert.ok(equalToFixed(result.heading, params.heading, 4));
+                assert.ok(equalToFixed(result.tilt, params.tilt, 4));
+                assert.ok(equalToFixed(result.range, params.range, 1));
+                assert.ok(equalToFixed(result.coord.longitude, params.coord.longitude, 4));
+                assert.ok(equalToFixed(result.coord.latitude, params.coord.latitude, 4));
+                done();
+            }, done);// ni ici
     });
+    */
```
gchoqueux: could fix it or open a issue?

test/unit/potreelayerparsing.js
```diff
@@ -90,18 +90,21 @@ describe('Potree Provider', function () {
...
+                assert.ok(l.material.defines.NORMAL_OCT16);
+            });
 
-        ps.forEach(p => View.prototype.addLayer.call(view, p));
+        layers.forEach(p => View.prototype.addLayer.call(view, p));
 
-        Promise.all(ps.map(p => p.whenReady)).then(() => done());
+        Promise.all([p1, p2, p3, p4])
+            .then(() => done())
+            .catch(done);
```
gchoqueux, line 107: you could write

`Promise.all([p1, p2, p3, p4]).then(done, done)`

gchoqueux (commented):
thanks for this enhancement

test/unit/cameraUtils.js
ftoromanoff: Moved to a new issue #2080

test/unit/potreelayerparsing.js
ftoromanoff: No it can't as Promise.all return something when all promises are resolved and done need to be send with empty param.
Maybe there is another way to simplify it ?

gchoqueux: `Promise.all([p1, p2, p3, p4]).then(done).catch(done)` ?

ftoromanoff: neither as then(done) is short for then ((res) => done(res)) or done should be call with no parameters and Promise.all return an array. (In this case the Array is [,,] as the result of all Promise is empty.)

gchoqueux: ok

mgermerie (approved):

gchoqueux (approved):

PR #2046, opened 2023-03-21
feat(planarcontrols) : add enabled attribute

valentinMachado:
## Description
Add an enabled attribute in PlanarControls, the same way that [THREE.OrbitControls](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/OrbitControls.js).

## Motivation and Context
Motivation have been discussed [here](https://github.com/iTowns/itowns/discussions/2045#discussioncomment-5381274)

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -90,6 +90,7 @@ const plane = new THREE.Plane(
 
 // default parameters :
 const defaultOptions = {
+    enabled: true,
     enableRotation: true,
     rotateSpeed: 2.0,
     minPanSpeed: 0.05,
@@ -134,6 +135,7 @@ export const PLANAR_CONTROL_EVENT = {
  * @class   PlanarControls
  * @param   {PlanarView}    view                                the view where the controls will be used
  * @param   {object}        options
+ * @param   {boolean}       [options.enabled=true]              Set to false to disable this control
  * @param   {boolean}       [options.enableRotation=true]       Enable the rotation with the `CTRL + Left mouse button`
  * and in animations, like the smart zoom.
  * @param   {boolean}       [options.enableSmartTravel=true]    Enable smart travel with the `wheel-click / space-bar`.
@@ -172,6 +174,9 @@ class PlanarControls extends THREE.EventDispatcher {
         super();
         this.view = view;
         this.camera = view.camera.camera3D;
+
+        // Set to false to disable this control
+        this.enabled = typeof options.enabled == 'boolean' ? options.enabled : defaultOptions.enabled;
 
         if (this.camera.isOrthographicCamera) {
             cameraInitialZoom = this.camera.zoom;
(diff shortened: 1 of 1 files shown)


jailln:
I'll let @mgermerie give its opinion since he implemented the `State Controls` but I just found out that doing `planarControls.state.enabled = false` should do what you need. That's how it seems to be implemented for the `GlobeControls` anyway. Sorry to finding it out after you implemented this ^^ Let me know if that works.

valentinMachado:
I've tried to set `planarControls.state.enabled = false` but this is not working, `planarControls.state` values are not a [StateControl](https://github.com/iTowns/itowns/blob/master/src/Controls/StateControl.js) as `globeControls.states` but an enum define [here](https://github.com/iTowns/itowns/blob/9b6d59f19ec5ee3e65b6a0f731365d51e5644060/src/Controls/PlanarControls.js#L29) like the THREE.OrbitControls.

jailln:
Ah ok sorry I missed that! Then the implementation you propose is ok for me. I'll let @mgermerie review too.

jailln (approved):

src/Controls/PlanarControls.js
```diff
@@ -1099,6 +1116,10 @@ class PlanarControls extends THREE.EventDispatcher {
      * @ignore
      */
     onContextMenu(event) {
+        if (!this.enabled) {
+            return;
+        }
+
```
mgermerie: Doing this will allow the right-click menu to pop when right-clicking on a PlanarView with disabled controls. Currently, it is prevented and I don't think this behavior should change, but I might miss some use case. Perhaps you know of such an use case @valentinMachado and @jailln ?

mgermerie (commented):
Hi ! Thanks for this PR. 
In deed, `PlanarView` does not implements `StateControls` yet. Therefore we cannot disable `PlanarControls` the same way we do with `GlobeControls`. The `PlanarControls` and `GlobeControls` should be aligned to both use `StateControls`, so we can have a single entry point to enable/disable input events and change movements keybindings.
Except for the comment I posted, it should be good to me.

src/Controls/PlanarControls.js
valentinMachado: Right, to be honest I jus copy the [THREE.OrbitControl behavior](https://github.com/mrdoob/three.js/blob/c0b03d674b82fb27ad163070c84816f58df77f8e/examples/jsm/controls/OrbitControls.js#L1192) but if the event.preventDefault() should be the behavior in anycase these lines should be removed.

@mgermerie @jailln Do I remove them ?

jailln: I don't really have an opinion on this, I think both can be done. I'll let you decide @mgermerie

mgermerie: I also thing we should remove them, so the preventDefault is the behavior in any case as you mentioned.

valentinMachado: done

mgermerie (approved):

PR #2056, opened 2023-03-27
fix(Label): catch no data elevation.

gchoqueux:
## Description
fix label elevation when there isn't no data texture or elevation layer

### DIFF
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -206,8 +206,17 @@ class Label extends THREE.Object3D {
     }
 
     updateElevationFromLayer(layer, nodes) {
-        const elevation = Math.max(0, DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z, nodes));
-        if (elevation && elevation != this.coordinates.z) {
+        if (layer.attachedLayers.filter(l => l.isElevationLayer).length == 0) {
+            return;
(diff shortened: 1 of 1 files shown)


src/Core/Label.js
```diff
@@ -206,8 +206,17 @@ class Label extends THREE.Object3D {
...
-        const elevation = Math.max(0, DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z, nodes));
-        if (elevation && elevation != this.coordinates.z) {
+        if (layer.attachedLayers.filter(l => l.isElevationLayer).length == 0) {
+            return;
+        }
+
+        let elevation = Math.max(0, DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z, nodes));
+
+        if (isNaN(elevation)) {
+            elevation = Math.max(0, DEMUtils.getElevationValueAt(layer, this.coordinates, DEMUtils.FAST_READ_Z));
```
jailln, line 216: Is it globally faster than always doing one terrain elevation without the `nodes` hint?

gchoqueux: yes, the tiles hint avoid the traverse all tiles tree to get tile with valid elevation texture

jailln (approved):

PR #2059, opened 2023-03-30
refactoring `LabelLayer` to architecture node

gchoqueux:
### Apply architecture node, simplify process and the rendering.

* `LabelLayer` extends `GeometryLayer`;
* introduce `DomNode` and `LabelsNode` to structure data in **architecture node**;
* simplify and optimize the process post-convert and the label culling;

### Performance

It added an **option perfomance** in `LabelLayer`.
This option allows to **reduce the number of labels** to be processed and rendered **without altering the final rendering**.
This reduction is done by filtering the labels that can never be displayed because they are too numerous.
The results are interesting when the data have too many labels for a given extent.
By example, the label process and rendering are **ten times faster** on IGN vector tiles.

### Others

* add **option to force clamping on terrain**, in case the original elevation data are wrong;
* add unit tests with all the context of the itowns processes;
* fix label culling on `GlobeView`;

https://user-images.githubusercontent.com/11291849/234934004-5ac0752f-f4f4-4b75-82da-19180f8aa462.mp4

## Motivation and Context

* remove last children in `TileMesh` other that `TileMesh`: will simplify removal `ObjectRemovalHelper`;

### DIFF
--- a/src/Layer/LabelLayer.js
+++ b/src/Layer/LabelLayer.js
@@ -1,7 +1,7 @@
 import * as THREE from 'three';
 import LayerUpdateState from 'Layer/LayerUpdateState';
 import ObjectRemovalHelper from 'Process/ObjectRemovalHelper';
-import Layer from 'Layer/Layer';
+import GeometryLayer from 'Layer/GeometryLayer';
 import Coordinates from 'Core/Geographic/Coordinates';
 import Extent from 'Core/Geographic/Extent';
 import Label from 'Core/Label';
@@ -13,6 +13,122 @@ const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 const _extent = new Extent('EPSG:4326', 0, 0, 0, 0);
 
 /**
+ * DomNode is a node in the tree data structure of labels divs.
+ *
+ * @class DomNode
+ */
+class DomNode {
+    #domVisibility = false;
+
+    constructor() {
+        this.dom = document.createElement('div');
+
+        this.dom.style.display = 'none';
+
+        this.visible = true;
+    }
+
+    get visible() { return this.#domVisibility; }
+
+    set visible(v) {
+        if (v !== this.#domVisibility) {
+            this.#domVisibility = v;
+            this.dom.style.display = v ? 'block' : 'none';
+        }
+    }
+
+    hide() { this.visible = false; }
+
+    show() { this.visible = true; }
+
+    add(node) {
+        this.dom.append(node.dom);
+    }
+}
+
+/**
+ * LabelsNode is node of tree data structure for LabelLayer.
+ * the node is made of dom elements and 3D labels.
+ *
+ * @class      LabelsNode
+ */
+class LabelsNode extends THREE.Group {
+    constructor(node) {
+        super();
+        // attached node parent
+        this.nodeParent = node;
+        // When this is set, it calculates the position in that frame and resets this property to false.
+        this.needsUpdate = true;
+    }
+
+    // instanciate dom elements
+    initializeDom() {
+        // create root dom
+        this.domElements = new DomNode();
+        // create labels container dom
+        this.domElements.labels = new DomNode();
+
+        this.domElements.add(this.domElements.labels);
+    }
+
+    // add node label
+    // add label 3d and dom label
+    addLabel(label) {
+        // add 3d object
+        this.add(label);
+
+        // add dom label
+        this.domElements.labels.dom.append(label.content);
+
+        // Batch update the dimensions of labels all at once to avoid
+        // redraw for at least this tile.
+        label.initDimensions();
+
+        // add horizon culling point if it's necessary
+        // the horizon culling is applied to nodes that trace the horizon which
+        // corresponds to the low zoom node, that's why the culling is done for a zoom lower than 4.
+        if (this.nodeParent.layer.isGlobeLayer && this.nodeParent.level < 4) {
+            label.horizonCullingPoint = new THREE.Vector3();
+        }
+    }
+
+    // remove node label
+    // remove label 3d and dom label
+    removeLabel(label) {
+        // remove 3d object
+        this.remove(label);
+
+        // remove dom label
+        this.domElements.labels.dom.removeChild(label.content);
+    }
+
+    // update position if it's necessary
+    updatePosition(label) {
+        if (this.needsUpdate) {
+            // update elevation from elevation layer.
+            if (this.needsAltitude) {
+                label.updateElevationFromLayer(this.nodeParent.layer, [this.nodeParent]);
+            }
+
+            // update elevation label
+            label.update3dPosition(this.nodeParent.layer.crs);
+
+            // update horizon culling
+            label.updateHorizonCullingPoint();
+        }
+    }
+
+    // return labels count
+    count() {
+        return this.children.length;
+    }
+
+    get labels() {
+        return this.children;
+    }
+}
+
+/**
  * A layer to handle a bunch of `Label`. This layer can be created on its own,
  * but it is better to use the option `addLabelLayer` on another `Layer` to let
  * it work with it (see the `vector_tile_raster_2d` example).
@@ -21,7 +137,7 @@ const _extent = new Extent('EPSG:4326', 0, 0, 0, 0);
  * LabelLayer.  Default is true. You should not change this, as it is used
  * internally for optimisation.
  */
-class LabelLayer extends Layer {
+class LabelLayer extends GeometryLayer {
     /**
      * @constructor
      * @extends Layer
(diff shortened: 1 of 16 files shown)


src/Layer/Layer.js
```diff
@@ -56,6 +53,13 @@ class Layer extends THREE.EventDispatcher {
      * the layer doesn't need Source (like debug Layer or procedural layer).
      * @param {number} [config.cacheLifeTime=Infinity] - set life time value in cache.
      * This value is used for [Cache]{@link Cache} expiration mechanism.
+     * @param {(boolean|Object)} [config.addLabelLayer=false] - Used to tell if this layer has
+     * labels to display from its data. For example, it needs to be set to `true`
+     * for a layer with vector tiles. If it's `true` a new `LabelLayer` is added and attached to this `Layer`.
+     * You could pass [LabelLayer]{@link LabelLayer} options in `addLabelLayer` (`addLabelLayer: { performance: true }`).
```
jailln: I would say "you can also configure it with options described below such as: `addLabelLayer: { performance: true }`" instead, WDYT?

src/Renderer/Label2DRenderer.js
```diff
@@ -136,16 +128,30 @@ class Label2DRenderer {
...
-        if (!this.infoTileLayer || !this.infoTileLayer.layer.attachedLayers.find(l => l.isLabelLayer && l.visible)) { return; }
+        const labelLayers = this.infoTileLayer && this.infoTileLayer.layer.attachedLayers.filter(l => l.isLabelLayer && l.visible);
+        if (labelLayers.length == 0) { return; }
         this.grid.reset();
 
         // set camera frustum
         frustum.setFromProjectionMatrix(camera.projectionMatrix);
-        this.culling(scene, this.infoTileLayer.displayed.extent, camera);
+
+        labelLayers.forEach((labelLayer) => {
```
jailln, line 142: This maked the `Label2DRenderer` aware of the internal structure of `LabelLayer` and we will have to modify here if we change it. I think it would be better to implement a method in `LabelLayer` to get all the `LabelsNode` instead. WDYT?

test/unit/label.js
```diff
@@ -173,63 +179,105 @@ describe('Label', function () {
...
-                    isTileMesh: true,
-                }],
-            }],
-        };
+    const renderer = new Renderer();
+    const placement = { coord: new Coordinates('EPSG:4326', 4.631512, 43.675626), range: 100000 };
+    const view = new GlobeView(renderer.domElement, placement, { renderer });
+
+    const gpxSource = new FileSource({
+        // url: urlGeojson,
```
jailln: remove?

src/Layer/LabelLayer.js
```diff
@@ -1,17 +1,134 @@
...
 const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 
 const _extent = new Extent('EPSG:4326', 0, 0, 0, 0);
 
+const nodeDimensions = new THREE.Vector2();
+const westNorthNode = new THREE.Vector2();
+const labelPosition = new THREE.Vector2();
+
+/**
+ * DomNode is node of tree data structure for labels div.
```
jailln: I would say `DomNode is a node in the tree data structure of labels divs`

src/Renderer/Label2DRenderer.js
```diff
@@ -163,87 +169,48 @@ class Label2DRenderer {
...
-        // the presence of the label inside frustum camera.
-        } else if (!frustum.containsPoint(object.getWorldPosition(worldPosition).applyMatrix4(camera.matrixWorldInverse))) {
-            this.grid.hidden.push(object);
+    culling(label, camera) {
+        label.getWorldPosition(worldPosition);
+        // Check if the frustum contains tle label
+        if (!frustum.containsPoint(worldPosition.applyMatrix4(camera.matrixWorldInverse)) ||
+        // Check if globe horizon culls the label
         // Do some horizon culling (if possible) if the tiles level is small
         // enough. The chosen value of 4 seems to provide a good result.
```
jailln: remove the part of the comment about the 4 value

src/Layer/LabelLayer.js
```diff
@@ -1,17 +1,134 @@
...
+
+        // add dom label
+        this.domElements.labels.dom.append(label.content);
+
+        // Batch update the dimensions of labels all at once to avoid
+        // redraw for at least this tile.
+        label.initDimensions();
+
+        // add horizon culling point if it's necessary
+        if (this.nodeParent.layer.isGlobeLayer && this.nodeParent.level < 4) {
```
jailln, line 97: could you document this magic number and declare it as a constant? why only do horizon culling in low levels?

src/Layer/LabelLayer.js
```diff
@@ -34,6 +152,9 @@ class LabelLayer extends Layer {
      * contains three elements `name, protocol, extent`, these elements will be
      * available using `layer.name` or something else depending on the property
      * name.
+     * @param {boolean} [config.performance=false] - remove labels that have no chance of being visible.
```
jailln: If I understand correctly, if set to true, we remove culled out labels. Why don't we do that in all cases?

src/Layer/LabelLayer.js
```diff
@@ -34,6 +152,9 @@ class LabelLayer extends Layer {
...
+     * Indeed, even in the best case, labels will never be displayed. By example, if there's many labels.
+     * @param {boolean} [config.forceClampToTerrain=false] - use elevation layer to clamp label on terrain.
```
jailln: Does this have an impact on performances? If yes, we should precise it in the doc. What happens if we don't set it to true? The labels may be under the terrain?

src/Layer/Layer.js
gchoqueux: yes

src/Renderer/Label2DRenderer.js
gchoqueux: I add:
* `LabelLayer.submittedLabelNodes` to get all submitted nodes for rendering;
* `LabelsNode.labels` to get all labels's node;

test/unit/label.js
gchoqueux: yes

src/Layer/LabelLayer.js
gchoqueux: yes

src/Renderer/Label2DRenderer.js
gchoqueux: yes

src/Renderer/Label2DRenderer.js
jailln: What is `submittedLabelNodes` ? Is it the nodes that `needsUpdate` ? Or visible labels ? Something else?

src/Layer/LabelLayer.js
gchoqueux: i add
```js
// the horizon culling is applied to nodes that trace the horizon which
// corresponds to the low zoom node, that's why the culling is done for a zoom lower than 4.
```

src/Renderer/Label2DRenderer.js
gchoqueux: it's all label nodes submitted to the render processing

src/Layer/LabelLayer.js
gchoqueux: yes, it is true that you can set the default to true

src/Layer/LabelLayer.js
gchoqueux: i add comment on documentation.

> The labels may be under the terrain?

For the moment, these labels have not yet and never been managed by itowns

jailln (approved):

src/Layer/LabelLayer.js
mgermerie: Would there be a point for user not to use `config.performance=true` ? What would be the use case of having labels inside a `LabelsNode` that will never be displayed ? Maybe `performance = true` should be the only behavior (without letting user parametise it) ?

src/Renderer/Label2DRenderer.js
```diff
@@ -163,87 +169,47 @@ class Label2DRenderer {
...
         } else {
-            // projecting world position object
+            // projecting world position label
             worldPosition.applyMatrix4(camera.projectionMatrix);
 
-            object.updateProjectedPosition(worldPosition.x * this.halfWidth + this.halfWidth, -worldPosition.y * this.halfHeight + this.halfHeight);
+            label.updateProjectedPosition(worldPosition.x * this.halfWidth + this.halfWidth, -worldPosition.y * this.halfHeight + this.halfHeight);
 
             // Are considered duplicates, labels that have the same screen
             // coordinates and the same base content.
```
mgermerie: Could you remove this comment ?
Also, why couldn't we use the comparison of `projectedPosition` to filter labels that have no text based on their position ?

mgermerie (commented):
Other than the two question and the minor change requested, it should be good to me. Thanks for the PR !

src/Layer/LabelLayer.js
gchoqueux: ~~Generally, the data is optimized and there is no need for additional processing.~~

Sorry, the default is true...

I will add a comment

src/Renderer/Label2DRenderer.js
gchoqueux: > Could you remove this comment ?

yes, i will remove

> Also, why couldn't we use the comparison of projectedPosition to filter labels that have no text based on their position ?

because this process is performed during culling

PR #2060, opened 2023-03-30
C3DTilesLayer Style

valentinMachado:
## Description
Add the possiblity to set a Style for a C3DTilesLayer with an api like so:

```
const layer = new C3DTIlesLayer(...., new Style({ fill: { color: function( tileBatchElement) => {
if(tileBatchElement.userData.selected === true) return "blue";
if(tileBatchElement["nameBatchTableAttributes"] === "something") return "red";
return "yellow";
});
```
tileBatchTableElement / tileBatchElement / tileBatchElementController (not sure about the name readability vs semantic) Element could be replace by Object ....
Anyway this object wrap the batchTable properties + some runtime attributes such as box3 or indexStart + count  (to retrieve it easily in geometry.attributes.position.array) + batchId + some userData (so itowns user can implement dynamic style based on its own code logic).

I am going to implement an example + a widget to demonstrate it.

## Motivation and Context
This change is required to simply add dynamic style to a C3DTilesLayer.

### DIFF
--- a/examples/widgets_3dtiles_style.html
+++ b/examples/widgets_3dtiles_style.html
@@ -1,0 +1,130 @@
+<html>
+    <head>
+        <title>Itowns - 3D Tiles Style Widget </title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/widgets.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv" class="viewer"></div>
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/itowns_widgets.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script src="js/GUI/GuiTools.js"></script>
+        <script type="text/javascript">
+            // Define crs projection that we will use (taken from https://epsg.io/3946, Proj4js section)
+            itowns.proj4.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+            // Define geographic extent: CRS, min/max X, min/max Y
+            const extent = new itowns.Extent( 'EPSG:3946',
+                1840816.94334, 1843692.32501,
+                5175036.4587, 5177412.82698);
+
+            // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+            const viewerDiv = document.getElementById('viewerDiv');
+
+            // Instanciate PlanarView*
+            const cameraCoord = new itowns.Coordinates('EPSG:3946', 1841980,
+                5175682, 3000)
+            const view = new itowns.PlanarView(viewerDiv, extent, { placement: {
+                coord: cameraCoord, heading: 30, range: 4000, tilt: 30 } });
+
+            setupLoadingScreen(viewerDiv, view);
+
+            // Add a WMS imagery source
+            const wmsImagerySource = new itowns.WMSSource({
+                extent: extent,
+                name: 'Ortho2009_vue_ensemble_16cm_CC46',
+                url: 'https://download.data.grandlyon.com/wms/grandlyon',
+                version: '1.3.0',
+                crs: 'EPSG:3946',
+                format: 'image/jpeg',
+            });
+
+            // Add a WMS imagery layer
+            const wmsImageryLayer = new itowns.ColorLayer('wms_imagery', {
+                updateStrategy: {
+                    type: itowns.STRATEGY_DICHOTOMY,
+                    options: {},
+                },
+                source: wmsImagerySource,
+            });
+
+            view.addLayer(wmsImageryLayer);
+
+            // Add a WMS elevation source
+            const wmsElevationSource = new itowns.WMSSource({
+                extent: extent,
+                url: 'https://download.data.grandlyon.com/wms/grandlyon',
+                name: 'MNT2012_Altitude_10m_CC46',
+                crs: 'EPSG:3946',
+                width: 256,
+                format: 'image/jpeg',
+            });
+
+            // Add a WMS elevation layer
+            const wmsElevationLayer = new itowns.ElevationLayer('wms_elevation', {
+                useColorTextureElevation: true,
+                colorTextureElevationMinZ: 144,
+                colorTextureElevationMaxZ: 622,
+                source: wmsElevationSource,
+            });
+
+            view.addLayer(wmsElevationLayer);
+
+            const $3dTilesLayer = new itowns.C3DTilesLayer(
+                0, {
+                    name: 'First Lyon Rounding 2018',
+                    source: new itowns.C3DTilesSource({
+                        url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/' +
+                            '3DTiles/lyon1_with_surface_type_2018/tileset.json',
+                    })
+                }, view);
+            itowns.View.prototype.addLayer.call(view, $3dTilesLayer);
+
+            // Lights
+            const dirLight = new itowns.THREE.DirectionalLight(0xffffff, 1);
+            dirLight.position.set(-0.9, 0.3, 1);
+            dirLight.updateMatrixWorld();
+            view.scene.add( dirLight );
+
+            const ambLight = new itowns.THREE.AmbientLight(0xffffff, 0.2);
+            view.scene.add( ambLight );
+
+            // Request redraw
+            view.notifyChange();
+
+            // Add 3D Tiles Style Widget
+            const widget = new itowns_widgets.C3DTilesStyle(view);
+            widget.domElement.setAttribute('id', 'widgets-3dtilesstyle');
+
+            // Add a box3 to C3DTileFeature onclick
+            const geometry = new itowns.THREE.BoxGeometry();
+            const material = new itowns.THREE.MeshBasicMaterial( {color: 0x00ff00, opacity: 0.5, transparent: true} );
+            const cube = new itowns.THREE.Mesh( geometry, material );
+            view.domElement.onclick = (event) => {
+                const intersects = view.pickObjectsAt(event, 0, [$3dTilesLayer]);
+                const c3DTileFeatureClicked = $3dTilesLayer.getC3DTileFeatureFromIntersectsArray(intersects);
+                
+                if (c3DTileFeatureClicked) {
+                    const worldBox3 = $3dTilesLayer.computeWorldBox3(c3DTileFeatureClicked);
+                    cube.scale.copy(worldBox3.max.clone().sub(worldBox3.min));
+                    worldBox3.getCenter(cube.position);
+                    cube.updateMatrixWorld();
+                    view.scene.add(cube);
+                    view.notifyChange();
+                }
+            }
+
+            // Add a debug UI
+            const menu = new GuiTools('menuDiv', view);
+            const d = new debug.Debug(view, menu.gui);
+            debug.createTileDebugUI(menu.gui, view, view.tileLayer, d);
+        </script>
+    </body>
+</html>
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -3,10 +3,24 @@ import GeometryLayer from 'Layer/GeometryLayer';
 import { init3dTilesLayer, pre3dTilesUpdate, process3dTilesNode } from 'Process/3dTilesProcessing';
 import C3DTileset from 'Core/3DTiles/C3DTileset';
 import C3DTExtensions from 'Core/3DTiles/C3DTExtensions';
+// eslint-disable-next-line no-unused-vars
+import Style from 'Core/Style';
+import C3DTFeature from 'Core/3DTiles/C3DTFeature';
+import { optimizeGeometryGroups } from 'Utils/ThreeUtils';
+
+export const C3DTILES_LAYER_EVENTS = {
+    /**
+     * Fires when a tile content has been loaded
+     * @event C3DTilesLayer#on-tile-content-loaded
+     * @property type {string} on-tile-content-loaded
+     */
+    ON_TILE_CONTENT_LOADED: 'on-tile-content-loaded',
+};
 
 const update = process3dTilesNode();
 
 class C3DTilesLayer extends GeometryLayer {
+    #fillColorMaterialsBuffer;
     /**
      * Constructs a new instance of 3d tiles layer.
      * @constructor
@@ -45,12 +59,8 @@ class C3DTilesLayer extends GeometryLayer {
      * threshold at which child nodes of the current node will be loaded and added to the scene.
      * @param {Number} [config.cleanupDelay=1000] The time (in ms) after which a tile content (and its children) are
      * removed from the scene.
-     * @param {Function} [config.onTileContentLoaded] Callback executed when the content of a tile is loaded.
-     * @param {Boolean|Material} [config.overrideMaterials='false'] option to override the materials of all the layer's
-     * objects. If true, a threejs [MeshBasicMaterial](https://threejs.org/docs/index.html?q=meshbasic#api/en/materials/MeshBasicMaterial)
-     * is set up. config.overrideMaterials can also be a threejs [Material](https://threejs.org/docs/index.html?q=material#api/en/materials/Material)
-     * in which case it will be used as the material for all objects of the layer.
      * @param {C3DTExtensions} [config.registeredExtensions] 3D Tiles extensions managers registered for this tileset.
+     * @param {Style} [config.style=null] - style used for this layer
      * @param  {View}  view  The view
      */
     constructor(id, config, view) {
@@ -58,11 +68,33 @@ class C3DTilesLayer extends GeometryLayer {
         this.isC3DTilesLayer = true;
         this.sseThreshold = config.sseThreshold || 16;
         this.cleanupDelay = config.cleanupDelay || 1000;
-        this.onTileContentLoaded = config.onTileContentLoaded || (() => {});
         this.protocol = '3d-tiles';
-        this.overrideMaterials = config.overrideMaterials ?? false;
         this.name = config.name;
         this.registeredExtensions = config.registeredExtensions || new C3DTExtensions();
+
+        /** @type {Style} */
+        this._style = config.style || null;
+
+        /** @type {Map<string, THREE.MeshStandardMaterial>} */
+        this.#fillColorMaterialsBuffer = new Map();
+
+        /**
+         * Map all C3DTFeature of the layer according their tileId and their batchId
+         * Map< tileId, Map< batchId, C3DTFeature>>
+         *
+         * @type {Map<number, Map<number,C3DTFeature>>}
+         */
+        this.tilesC3DTileFeatures = new Map();
+
+        if (config.onTileContentLoaded) {
+            console.warn('DEPRECATED onTileContentLoaded should not be passed at the contruction, use C3DTILES_LAYER_EVENTS.ON_TILE_CONTENT_LOADED event instead');
+            this.addEventListener(C3DTilesLayer.EVENT_TILE_CONTENT_LOADED, config.onTileContentLoaded);
+        }
+
+        if (config.overrideMaterials) {
+            console.warn('overrideMaterials is deprecated, use style API instead');
+            this.overrideMaterials = config.overrideMaterials;
+        }
 
         this._cleanableTiles = [];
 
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -2,9 +2,9 @@
   "name": "itowns",
   "version": "2.39.0",
   "description": "A JS/WebGL framework for 3D geospatial data visualization",
-  "main": "lib/Main.js",
+  "main": "lib/MainBundle.js",
```
valentinMachado: use build-dev

src/Process/3dTilesProcessing.js
```diff
@@ -278,7 +279,7 @@ function setDisplayed(node, display) {
     }
 }
 
-function markForDeletion(layer, elt) {
+export function markForDeletion(layer, elt) {
```
valentinMachado: remove this export

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,171 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+                this.batchID === otherID.batchID;
+    }
+
+    toString() {
+        return `C3DTilesLayerTileBatchID-${this.layerID}-${this.tileID}-${this.batchID}`;
+    }
+}
+
+// could be a three util function
+const computeInfoFromAttributes = (object3D, nameAttribute, valueAttribute) => {
```
valentinMachado: where to add three js util ?

src/Layer/C3DTilesLayer.js
```diff
@@ -58,12 +222,18 @@ class C3DTilesLayer extends GeometryLayer {
         this.isC3DTilesLayer = true;
         this.sseThreshold = config.sseThreshold || 16;
         this.cleanupDelay = config.cleanupDelay || 1000;
-        this.onTileContentLoaded = config.onTileContentLoaded || (() => {});
         this.protocol = '3d-tiles';
         this.overrideMaterials = config.overrideMaterials ?? false;
```
valentinMachado: remove override materials

valentinMachado:
I've done an example at http://localhost:8080/examples/widgets_3dtiles_style.html, normally all our needs in [ud-viz](https://github.com/VCityTeam/UD-Viz) should be covered. It could be enhance but it should do the job for now.


Capture d’écran du 2023-04-03 17-23-29

EDIT: Just realized that I did not add the example on the main page of the example + some functional test are broken since I modified the api of C3DTilesLayer. Let me know if the changes maded are okay for you if yes we could then fix this.

examples/layers/Lyon1_With_SurfaceTypes_2018/tiles/0.b3dm
jailln, line 1: Can you add this dataset to https://github.com/iTowns/iTowns2-sample-data please ? (and add it like [this](https://github.com/iTowns/itowns/blob/master/examples/3dtiles_25d.html#L88))

examples/widgets_3dtiles_style.html
```diff
@@ -0,0 +1,111 @@
...
+                colorTextureElevationMinZ: 144,
+                colorTextureElevationMaxZ: 622,
+                source: wmsElevationSource,
+            });
+
+            view.addLayer(wmsElevationLayer);
+
+            const $3dTilesLayer = new itowns.C3DTilesLayer(
+                0, {
+                    name: '1er Arrondissement Lyon 2018',
```
jailln: Can you translate to english please?

examples/widgets_3dtiles_style.html
```diff
@@ -0,0 +1,111 @@
...
+                    source: new itowns.C3DTilesSource({
+                        url: './layers/Lyon1_With_SurfaceTypes_2018/tileset.json',
+                    })
+                }, view);
+            itowns.View.prototype.addLayer.call(view, $3dTilesLayer);// why view.addLayer($3dTilesLayer) is not working ?
```
jailln: Because `view.addLayer()` "attaches" the layer to the `tileLayer` of the view by default. This attaching notion is not crystal clear for me but from what I've understood it is used to bind layers that are not tiled to those that are tiled to make them benefit from this tiling. @mgermerie if you have more information about this I would be happy too :) 
Anyway, I don't like that we have two ways to add layers and that one needs to call the View method through its prototype (and that the user needs to know which one to use while the `attach` notion is not well documented). In my opinion we should add all layers with `view.addLayers`, and attach layer or not automatically based on their type for instance (with a boolean parameter to force attaching or not if needed).

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,252 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+            // remove
+            object3D.material.splice(index, 1);
+        }
+    }
+};
+
+export class BatchElement {
+    constructor(layer, batchId, start, count, worldBox3, info, userData = {}) {
+        /** @type {C3DTilesLayer} */
+        this._layer = layer;// need this ref to update style when userData are modify
```
jailln: you can use [js private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) instead

jailln (changes requested):
The API exposed to the user is good and very handy :) However it adds quite an overhead when loading a tile, tile loading seems a lot slower (event when commenting out the `testGeometryGroups` method that is meant for debug). A quick look at profiling indicates that it seems to be due to the initialization of the batchElements array which is also very handy but seems costly to initialize. Could you try to reduce the time overhead at time loading please?
Maybe we should only compute the minimum information needed in the batchElement to apply the style later? Or maybe we should not store this `batchElements` map and access the values directly each time we need them? 

(I also added some comments about coding style but I will continue that after it has been optimized).

Thanks :)

valentinMachado:
> Maybe we should only compute the minimum information needed in the batchElement to apply the style later? Or maybe we should not store this `batchElements` map and access the values directly each time we need them?

Actually to avoid to parse all geometry of the tile content each time (which is quite an expensive computation) it should be done one time at the init time. I optimize the initBatchElement method by removing the computation of the world box3 which is now a dynamic method of a BatchElement class + add an allowTileIdList to only updateStyle of batch element of a certain tile which avoid to recompute each style of each tile when a tileContent is loaded.

This PR depends on this [one](https://github.com/iTowns/iTowns2-sample-data/pull/11)

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,240 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+        }
+        const tileIds = [];
+        batchElements.forEach((b) => {
+            if (!tileIds.includes(b._tileId)) {
+                tileIds.push(b._tileId);
+            }
+
+            b.setUserData(key, value, false);
+        });
+        batchElements[0].#layer._styleController.updateStyle(batchElements[0].#layer, tileIds);
```
valentinMachado: batchElements in array could be from != layer

valentinMachado:
After a discussion with Clement COLIN of the VCity team, he thinks that BatchElement (which is okay in a 3D context) should be name Feature or Feature3D (which is better in a geospatial/3DTiles context) do you agree with that @jailln ?

EDIT: After reading [3DTiles specs](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Batched3DModel) I think we should name this class C3DTileFeature.
As we can read overthere :
"..., each model is a feature." and
"The batchId indicates the model to which the vertex belongs."

our current batchelement in a 3DTile context is representing a batch id <==> a model <==> feature

Also read that [batchid have not to be sorted](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Batched3DModel#binary-gltf) which is not handle by current code... I am going to do it

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,256 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+     * @param {Array<number>|null} [allowTileIdList] - tile ids to allow in updateStyle computation if null all tile id are computed
+     * @returns {boolean} true if style updated false otherwise
+     */
+    updateStyle(c3DTilesLayer, allowTileIdList) {
+        if (!c3DTilesLayer._style) {
+            return false;
+        }
+
+        // dispose buffered materials
+        for (const [id, fillMaterial] of this._fillColorMaterialsBuffer) {
```
valentinMachado: materials of not allow tile id could be dispose and not recreate then

utils/debug/3dTilesDebug.js
```diff
@@ -108,3 +108,60 @@ export default function create3dTilesDebugUI(datDebugTool, view, _3dTileslayer)
...
+    if (groupsCount != tileMesh.geometry.attributes.position.count) {
+        throw new Error('groups count error');
+    }
+
+    // Check if group index + count are consistent
+    tileMesh.geometry.groups.sort((a, b) => a.start - b.start);
+    if (tileMesh.geometry.groups[0].start != 0) {
+        throw new Error('groups minimal index is not zero'); // not starting at zero
+    }
+    // One batch id cant be in != group
```
valentinMachado: yes indeed

jailln:
> After a discussion with Clement COLIN of the VCity team, he thinks that BatchElement (which is okay in a 3D context) should be name Feature or Feature3D (which is better in a geospatial/3DTiles context) do you agree with that @jailln ?
> 
> EDIT: After reading [3DTiles specs](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Batched3DModel) I think we should name this class C3DTileFeature. As we can read overthere : "..., each model is a feature." and "The batchId indicates the model to which the vertex belongs."
> 
> our current batchelement in a 3DTile context is representing a batch id <==> a model <==> feature
> 
> Also read that [batchid have not to be sorted](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Batched3DModel#binary-gltf) which is not handle by current code... I am going to do it

Yes I agree with the name `C3DTilesFeature` :+1:

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
 import { init3dTilesLayer, pre3dTilesUpdate, process3dTilesNode } from 'Process/3dTilesProcessing';
 import C3DTileset from 'Core/3DTiles/C3DTileset';
 import C3DTExtensions from 'Core/3DTiles/C3DTExtensions';
+// eslint-disable-next-line no-unused-vars
+import Style from 'Core/Style';
+import { optimizeGeometryGroups } from 'Utils/ThreeUtils';
+
+export class C3DTileFeature {
```
jailln: Can you put it in a separate file please? (in `src/Core/3DTiles/`) And add a jsdoc for the class (even if it's a bit redundant with the constructor but move `represent a chunk of the tileContent geometry` in this comment instead of the one of the constructor for instance)

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+
+                target.applyMatrix4(child.matrixWorld);
+            }
+        });
+
+        return target;
+    }
+
+    /**
+     * Use this method to modify userData
```
jailln: can you describe a bit in the comment what is userData and in which cases it updates the style please?

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+    /**
+     * get info (no setter since info should not change at runtime)
+     * @returns {object} - batchTable info
+     */
+    getInfo() {
+        return this.#info;
+    }
+
+    /**
+     * Use this method to modify user data of an array of batchelements (updateStyle only one time)
```
jailln: `batchelements` -> `C3DTileFeature`

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+     * @param {Array<{start:number, count:number}>} groups - groups in geometry.attributes matching batchId
+     * @param {object} info - info in the batchTable
+     * @param {object} [userData={}] - some userData
+     */
+    constructor(layer, tileId, batchId, groups, info, userData = {}) {
+        /** @type {C3DTilesLayer} */
+        this.#layer = layer;
+
+        /** @type {number} */
+        this._tileId = tileId;
```
jailln: why `_` ? same for groups

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+        });
+
+        // update style
+        updateStyleInfos.forEach((usi) => {
+            usi.layer._styleController.updateStyle(usi.layer, usi.tileIds);
+        });
+    }
+}
+
+class StyleController {
```
jailln: Can you rename it to `C3DTilesStyleController` and add doc or jsdoc on it please? (even if it's not exported, only for the devs). Also, can you move it to a separate file in `src/Core/C3DTiles/` please?

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+    constructor() {
+        /** @type {Map<string,THREE.MeshStandardMaterial} */
+        this._fillColorMaterialsBuffer = new Map();
+    }
+
+    /**
+     *
```
jailln: Add comment: updates the style of the tiles with alllowTileIdList and clear others tiles style?

src/Layer/C3DTilesLayer.js
```diff
@@ -45,25 +309,28 @@ class C3DTilesLayer extends GeometryLayer {
...
         this.registeredExtensions = config.registeredExtensions || new C3DTExtensions();
 
+        /** @type {Style} */
+        this._style = config.style || null;
+
+        /** @type {StyleController} */
+        this._styleController = new StyleController();
+
+        /** @type {Map<number, Map<number,C3DTileFeature>>} */
+        this.tilesC3DTileFeatures = new Map();
```
jailln, line 87: can you document what it maps please?

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+     * @param {Array<number>|null} [allowTileIdList] - tile ids to allow in updateStyle computation if null all tile id are computed
+     * @returns {boolean} true if style updated false otherwise
+     */
+    updateStyle(c3DTilesLayer, allowTileIdList = null) {
+        if (!c3DTilesLayer._style) {
+            return false;
+        }
+
+        const currentMaterials = [];// list materials used for this update
+        for (const [tileId, c3DTileFeatures] of c3DTilesLayer.tilesC3DTileFeatures) {
```
jailln: wouldn't it be more efficient to traverse once `C3DTilesLayer.object3d` and update the style of each object depending on `allowTileIdList` ?

src/Layer/C3DTilesLayer.js
```diff
@@ -3,6 +3,270 @@ import GeometryLayer from 'Layer/GeometryLayer';
...
+                if (child.geometry && child.geometry.attributes._BATCHID) {
+                    // check if tile feature should be updated
+                    if (!(allowTileIdList && !allowTileIdList.includes(tileId))) {
+                        // clear
+                        child.geometry.clearGroups();
+                        child.material = [];
+
+                        // eslint-disable-next-line no-unused-vars
+                        for (const [batchId, c3DTileFeature] of c3DTileFeatures) {
+                            /** @type {THREE.Color} */
```
jailln: could we extract the following logic to one or more function to simplify this function please? e.g. something like `styleColorToThreeColor` or `getThreeColorFromStyle` ; same for opacity, and same for the material and materialIndex manipulations

src/Layer/C3DTilesLayer.js
```diff
@@ -45,25 +309,28 @@ class C3DTilesLayer extends GeometryLayer {
...
      * @param {C3DTExtensions} [config.registeredExtensions] 3D Tiles extensions managers registered for this tileset.
+     * @param {Style} [config.style=null] - style used for this layer.
      * @param  {View}  view  The view
      */
     constructor(id, config, view) {
         super(id, new THREE.Group(), { source: config.source });
         this.isC3DTilesLayer = true;
         this.sseThreshold = config.sseThreshold || 16;
         this.cleanupDelay = config.cleanupDelay || 1000;
-        this.onTileContentLoaded = config.onTileContentLoaded || (() => {});
```
jailln, line 61: Can you make it backward compatible please? (i.e. add a deprecation warning and register it on the `EVENT_TILE_CONTENT_LOADED` please?

src/Layer/C3DTilesLayer.js
```diff
@@ -45,25 +309,28 @@ class C3DTilesLayer extends GeometryLayer {
...
         this.protocol = '3d-tiles';
-        this.overrideMaterials = config.overrideMaterials ?? false;
```
jailln, line 63: Can you leave it and add a deprecation warning that explains to use the style API please?

src/Layer/C3DTilesLayer.js
```diff
@@ -45,25 +309,28 @@ class C3DTilesLayer extends GeometryLayer {
...
         this.name = config.name;
         this.registeredExtensions = config.registeredExtensions || new C3DTExtensions();
 
+        /** @type {Style} */
+        this._style = config.style || null;
```
jailln, line 76: why `_` ? (same below)

src/Layer/C3DTilesLayer.js
```diff
@@ -142,48 +409,152 @@ class C3DTilesLayer extends GeometryLayer {
...
+     *
+     * @param {THREE.Object3D} tileContent - tile as THREE.Object3D
+     */
+    initC3DTileFeatures(tileContent) {
+        tileContent.traverse((child) => {
+            if (child.geometry && child.geometry.attributes._BATCHID) {
+                this.tilesC3DTileFeatures.set(tileContent.tileId, new Map());// initialize
+                const batchTable = this.findBatchTable(child);
+                if (!batchTable) {
+                    throw new Error('no batchTable');
```
jailln, line 246: This should not throw an error since batch table is not mandatory for b3dm according to the 3d tiles spec

src/Layer/C3DTilesLayer.js
```diff
@@ -142,48 +409,152 @@ class C3DTilesLayer extends GeometryLayer {
...
+                }
+
+                const geoAttributes = child.geometry.attributes;
```
jailln: `geomAttributs` to avoid confusion with `geo` as `geography` ? ^^

src/Layer/C3DTilesLayer.js
```diff
@@ -142,48 +409,152 @@ class C3DTilesLayer extends GeometryLayer {
...
+                    }
+                }
+            }
+        });
+
+        this._styleController.updateStyle(this, [tileContent.tileId]);// only update tile batchelement
+    }
+
+    /**
+     * Explicit tell the style controller to update style (could be call when Style function have changed)
```
jailln: `Could be call when Style function have changed` -> to remove ?

src/Layer/C3DTilesLayer.js
```diff
@@ -142,48 +409,152 @@ class C3DTilesLayer extends GeometryLayer {
...
+        this._styleController.updateStyle(this);
+    }
+
+    get style() {
+        return this._style;
+    }
 
-        return resultInfo;
+    // EVENT
+    static get EVENT_TILE_CONTENT_LOADED() {
```
jailln: Could you document this event and export it (so it is in the doc); e.g. like [GlobeView events](https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/GlobeView.js)

test/unit/3dtileslayerstyle.js
```diff
@@ -0,0 +1,111 @@
...
+        result.tileId = tileId;
+
+        return result;
+    };
+
+    $3dTilesLayer.style = new Style({
+        fill: {
+            color: (c3DTileFeature) => {
+                if (c3DTileFeature.batchId > 1) {
+                    return 'red';// batchId 0 is red
```
jailln: comments are reversed I think ^^

jailln (changes requested):
Thanks for the modifications and commenting and testing effort! Still some comments but it should only require minor modifications

src/Layer/C3DTilesLayer.js
valentinMachado: Iam using _ to indicate this attribute should only be accessed in this file

src/Layer/C3DTilesLayer.js
valentinMachado: According the implementation of the b3dm parser there should be a batchtable here (empty or not).

src/Layer/C3DTilesLayer.js
valentinMachado: Actually this is the only case where you have to call this method since C3DTileFeature modification handle style update

src/Layer/C3DTilesLayer.js
valentinMachado: Actually C3DTileFeature have a reference to their layer (which allow auto style update + dynamic method of the computeWorldBox3)  doesnt sound right to move to Core since it depends of the Layer

src/Layer/C3DTilesLayer.js
valentinMachado: Same as C3DTileFeature

src/Layer/C3DTilesLayer.js
valentinMachado: In the end I removed the layer reference, so it's up to the user to call updateStyle of the layer

src/Layer/C3DTilesLayer.js
valentinMachado: I remove this function since userData is no longer a private field (no auto update of the layer style)

src/Layer/C3DTilesLayer.js
valentinMachado: same this function has been removed

src/Layer/C3DTilesLayer.js
valentinMachado: I move C3DTFeature in another file so I removed it

src/Layer/C3DTilesLayer.js
valentinMachado: I remove this class, since it's related to the layer I put this code in the Layer class

src/Layer/C3DTilesLayer.js
valentinMachado: it doesn't clear other tiles style but leave them with their current style.
I comment with that => Update style of the C3DTFeatures, an allowList of tile id can be passed to only update certain tile

src/Layer/C3DTilesLayer.js
valentinMachado: done

src/Layer/C3DTilesLayer.js
valentinMachado: I don't understand what you mean

src/Layer/C3DTilesLayer.js
valentinMachado: You want to write a function `styleColorToThreeColor` for these lines ?
```
/** @type {THREE.Color} */
let color = null;
if (typeof this._style.fill.color === 'function') {
    color = new THREE.Color(this._style.fill.color(c3DTileFeature));
} else {
    color = new THREE.Color(this._style.fill.color);
}
```
Since this function should be only used in that function ie `updateStyle` it should then be declared in this function scope in order to not pollute C3DTilesLayer API so it will looks like something like this:

```
const styleColorToThreeColor = (feature) => {
 let color = null;
if (typeof this._style.fill.color === 'function') {
    color = new THREE.Color(this._style.fill.color(feature));
} else {
    color = new THREE.Color(this._style.fill.color);
}
return color;
}

...

const color = styleColorToThreeColor(c3DTFeature);
```

Same goes for opacity + material. This is what you mean ?

Moreover it should be name styleFillColorToThreeColor

src/Layer/C3DTilesLayer.js
valentinMachado: done

src/Layer/C3DTilesLayer.js
valentinMachado: not sure about the deprecated warning message but I've done it

src/Layer/C3DTilesLayer.js
valentinMachado: here the _ allow to use [javascript  property accessors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors) if it was named style layer.style will end with a maximum call stack exceed

src/Layer/C3DTilesLayer.js
valentinMachado: done

src/Layer/C3DTilesLayer.js
valentinMachado: with the new behavior ie no auto update for the feature, this should be call everytime style should be updated ie (style function have changed or feature have changed)

src/Layer/C3DTilesLayer.js
valentinMachado: done

test/unit/3dtileslayerstyle.js
valentinMachado: I remove comments ^^

src/Layer/C3DTilesLayer.js
jailln: ah yes ok, forgot that an empty batch table is instanciated when there is not one in the b3dm tile :)

src/Layer/C3DTilesLayer.js
```diff
@@ -311,6 +61,10 @@ class C3DTilesLayer extends GeometryLayer {
      * removed from the scene.
      * @param {C3DTExtensions} [config.registeredExtensions] 3D Tiles extensions managers registered for this tileset.
      * @param {Style} [config.style=null] - style used for this layer.
+     * @param {Boolean|Material} [config.overrideMaterials='false'] option to override the materials of all the layer's
```
jailln: Can you remove the doc about this one so we don't encourage user to use it since it will be removed in the next major version? (deprecation message)

src/Layer/C3DTilesLayer.js
```diff
@@ -325,12 +79,27 @@ class C3DTilesLayer extends GeometryLayer {
...
+        /**
+         * Map all C3DTFeature of the layer according their tileId and their batchId
+         * Map< tileId, Map< batchId, C3DTFeature>>
+         *
+         * @type {Map<number, Map<number,C3DTFeature>>}
+         */
         this.tilesC3DTileFeatures = new Map();
 
+        if (config.onTileContentLoaded) {
+            console.warn('DEPRECATED onTileContentLoaded should not be passed at the contruction, use addEventListener instead');
```
jailln: Could you modify the end by `use EVENT_TILE_CONTENT_LOADED event instead` please?

src/Layer/C3DTilesLayer.js
jailln: For each `tileId` you get the `Object3D` of `layer.object3d` with the `tileId` property, meaning that for each step of the for loop (for each `tileId`) you traverse the tree once (maybe not fully) to get the right `Object3D`. Another way would be to traverse `layer.object3d` once and apply the style modifications only to the right tiles. So instead of:

```
for (const [tileId, c3DTileFeatures] of this.tilesC3DTileFeatures) {
            const tileContent = this.object3d.getObjectByProperty('tileId', tileId);
            // Do stuff on tileContent
}
```

something like:

```
this.object3d.traverse((obj) => {
     if (obj.tileId && this.tilesC3DTileFeatures.has(obj.tileId)) { 
          // Do stuff on tileContent
     }
});
```

I think it should be more efficient in many cases but I may be wrong, WDYT?

src/Layer/C3DTilesLayer.js
jailln: Yes that's what I meant, but I think you can leave the way it is for now. When we will manage more style options I think we should move this style conversion logic into a dedicated `StyleConverter` file but for now it's ok like this

src/Layer/C3DTilesLayer.js
valentinMachado: Indeed you right, I did the modification

jailln:
Ok the code is good for me, thanks for all the modifications! 
Can you squash and rename the commits to follow the angular convention so we get a nice changelog for the new features added by this PR please? :)

jailln (approved):

Issue #2066, opened 2023-04-06
GeoJsonParser: error when firstCoordinates for an empty array


AnthonyGlt:
I'm getting an error when my coordinates are an empty array while calling firstCoordinates  in the file GeoJsonParser.js (new function from commit 9d509da)

##Resulting error

```
GeoJsonParser.js:188 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at firstCoordinates (GeoJsonParser.js:188:1)
```

A possible solution could be to check if `a` is either _empty_ or _undefined_:

`const firstCoordinates = a => (Array.isArray(a) && (!a.length || !isNaN(a[0]))  ? a : firstCoordinates(a[0]));`

OR

` const firstCoordinates = a =>  a === undefined || (Array.isArray(a) && !isNaN(a[0])  ? a : firstCoordinates(a[0]));`

PR #2067, opened 2023-04-06
fix(GeoJsonParser): fix firstCoordinates when empty

AnthonyGlt:
Fix issue on GeoJsonParser.js

Related issue:
https://github.com/iTowns/itowns/issues/2066

### DIFF
--- a/src/Parser/GeoJsonParser.js
+++ b/src/Parser/GeoJsonParser.js
@@ -141,7 +141,7 @@ function toFeatureType(jsonType) {
 
 const keyProperties = ['type', 'geometry', 'properties'];
 
-const firstCoordinates = a => (Array.isArray(a) && !isNaN(a[0])  ? a : firstCoordinates(a[0]));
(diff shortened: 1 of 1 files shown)


src/Parser/GeoJsonParser.js
```diff
@@ -141,7 +141,7 @@ function toFeatureType(jsonType) {
 
 const keyProperties = ['type', 'geometry', 'properties'];
 
-const firstCoordinates = a => (Array.isArray(a) && !isNaN(a[0])  ? a : firstCoordinates(a[0]));
+const firstCoordinates = a =>  a === undefined || (Array.isArray(a) && !isNaN(a[0])  ? a : firstCoordinates(a[0]));
```
mgermerie: I think you need to add parenthesis as such : 
```js
const firstCoordinates = a => (a === undefined || (Array.isArray(a) && !isNaN(a[0])) ? a : firstCoordinates(a[0]));
```
Otherwise, `firstCoordinates(undefined)` would return `true`, instead of `undefined`.

mgermerie (changes requested):

mgermerie (approved):

PR #2077, opened 2023-04-17
feat(underground): dynamic opacity to visualize underground and add hideSkirt parameter

AnthonyGlt:
## Description
Visualizing underground data using dynamic opacity. Hide the layers' skirt without removing it while loading new ones. This is done using the new parameter _hideSkirt_ of TiledGeometryLayer.
The gound's opacity is dynamically changed while zooming.

## Motivation and Context
Feature useful for underground data visualization. It's working for planar and globe view.
**Added exemple and functional test**

## Screenshots

Capture d'écran 2023-04-17 121734
Capture d'écran 2023-04-17 121749
Capture d'écran 2023-04-17 121808

https://user-images.githubusercontent.com/126568810/232477757-ebf04efa-a062-4997-b3c5-b7d1f6edc8b1.mp4

### DIFF
--- a/examples/config.json
+++ b/examples/config.json
@@ -78,7 +78,9 @@
         "misc_orthographic_camera": "Orthographic camera",
         "misc_custom_controls": "Define custom controls",
         "misc_custom_label": "Custom label popup",
-        "misc_camera_traveling": "Camera traveling"
+        "misc_camera_traveling": "Camera traveling",
+        "underground_25d_map": "Underground visualization"
+
     },
 
     "Widgets": {
--- a/examples/underground_25d_map.html
+++ b/examples/underground_25d_map.html
@@ -1,0 +1,153 @@
+<!DOCTYPE html>
+<!--
+To change this license header, choose License Headers in Project Properties.
+To change this template file, choose Tools | Templates
+and open the template in the editor.
+-->
+<html>
+    <head>
+        <title>Planar</title>
+        <link rel="stylesheet" type="text/css" href="./css/example.css">
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="description">
+            Underground visualization
+            <ul>
+              <li>Option available in "Layer Planar" controls</li>
+              <li>Enable "Dynamic opacity"</li>
+              <li>Zoom-in to look at underground data</li>
+              <li>Dynamic opactiy triggered by the zoom</li>              
+              <li>Can be enabled in globe view</li>              
+            </ul>
+        </div>
+        <div id="viewerDiv" class="viewer"></div>
+        <script src="./js/GUI/GuiTools.js"></script>
+        <script src="./js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        </script>
+        
+        <script>
+            var THREE=itowns.THREE;
+            var linesBus = [];
+
+            Math.Clamp = function clamp(n, min, max) {
+                if(n > max){
+                    return max;
+                } else if(n < min){
+                    return min
+                } else{
+                    return n;
+                }
+                }
+            
+                 // Retrieve the view container
+            const viewerDiv = document.getElementById('viewerDiv');
+    
+            // Define the view geographic extent
+            itowns.proj4.defs('EPSG:2154', '+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'); 
+            itowns.proj4.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+
+            const viewExtent = new itowns.Extent(
+                'EPSG:3946',
+                1837816.94334, 1847692.32501,
+                5170036.4587, 5178412.82698);
+            
+            var placement = {
+                    coord: new itowns.Coordinates('EPSG:3946', 1840839, 5172718, 0),
+                    heading: 45,
+                    range: 1800,
+                    tilt: 30,
+                };
+            var options = {placement: placement,            
+                controls : {minDistance : -50, handleCollision : false},    // To handle underground visualization, disable collision and allow camera below the ground
+            };
+            // Create the planar view
+            const view = new itowns.PlanarView(viewerDiv, viewExtent, options);
+            
+            const wmsImagerySource = new itowns.WMSSource({
+                url: 'https://download.data.grandlyon.com/wms/grandlyon',
+                networkOptions: { crossOrigin: 'anonymous' },
+                version: '1.3.0',
+                name: 'Ortho2009_vue_ensemble_16cm_CC46',
+                crs: 'EPSG:3946',
+                extent: viewExtent,
+                format: 'image/jpeg',
+            });
+
+            // Add a WMS imagery layer
+            const wmsImageryLayer = new itowns.ColorLayer('wms_imagery', {
+                transparent: false,
+                source: wmsImagerySource,
+            });
+
+            view.addLayer(wmsImageryLayer);
+            
+            var wmsElevationSource = new itowns.WMSSource({
+                extent: viewExtent,
+                url: 'https://download.data.grandlyon.com/wms/grandlyon',
+                name: 'MNT2012_Altitude_10m_CC46',
+                crs: 'EPSG:3946',
+                width: 256,
+                format: 'image/jpeg',
+            });
+
+            // Add a WMS elevation layer
+            var wmsElevationLayer = new itowns.ElevationLayer('wms_elevation', {
+                useColorTextureElevation: true,
+                colorTextureElevationMinZ: 144,
+                colorTextureElevationMaxZ: 622,
+                source: wmsElevationSource,
+            });
+
+            view.addLayer(wmsElevationLayer);
+
+
+            function acceptFeatureBus(properties) {
+                if (properties.sens == 'Aller') {
+                    var line = properties.ligne;
+                    if (linesBus.indexOf(line) === -1) {
+                        linesBus.push(line);
+                        return true;
+                    }
+                }
+                return false;
+            }
+            
+            var baseAltitude = 130;
+            var lyonTclBusSource = new itowns.WFSSource({
+                url: 'https://download.data.grandlyon.com/wfs/rdata?',
+                protocol: 'wfs',
+                version: '2.0.0',
+                id: 'tcl_bus',
+                typeName: 'tcl_sytral.tcllignebus',
+                crs: 'EPSG:3946',
+                fetcher: itowns.Fetcher.json,
+                parser: itowns.GeoJsonParser.parse,
+                format: 'geojson',
+            });
+
+            var lyonTclBusLayer = new itowns.FeatureGeometryLayer('WFS Bus lines', {
+                name: 'lyon_tcl_bus',
+                filter: acceptFeatureBus,
+                source: lyonTclBusSource,
+
+                style: new itowns.Style({
+                    stroke: {
+                        base_altitude: baseAltitude,
+                        width: 2,
+
+                    },
+                }),
+            });
+            view.addLayer(lyonTclBusLayer);
+
+            var menuPlanar = new GuiTools('menuDiv', view);
+            debug.createTileDebugUI(menuPlanar.gui, view, undefined, undefined, true);
+
+        </script>
+    </body>
+</html>
--- a/examples/underground_25d_map.html
+++ b/examples/underground_25d_map.html
@@ -26,23 +26,12 @@ and open the template in the editor.
         </div>
         <div id="viewerDiv" class="viewer"></div>
         <script src="./js/GUI/GuiTools.js"></script>
-        <script src="./js/GUI/LoadingScreen.js"></script>
         <script src="../dist/itowns.js"></script>
         <script src="../dist/debug.js"></script>
 
         <script>
             var THREE = itowns.THREE;
             var linesBus = [];
-
-            Math.Clamp = function clamp(n, min, max) {
-                if (n > max) {
-                    return max;
-                } else if (n < min) {
-                    return min;
-                } else {
-                    return n;
-                }
-            };
 
             // Retrieve the view container
             const viewerDiv = document.getElementById("viewerDiv");
--- a/examples/underground_25d_map.html
+++ b/examples/underground_25d_map.html
@@ -7,20 +7,21 @@ and open the template in the editor.
 <html>
     <head>
         <title>Planar</title>
-        <link rel="stylesheet" type="text/css" href="./css/example.css">
-        <meta charset="UTF-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <link rel="stylesheet" type="text/css" href="./css/example.css" />
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
     </head>
     <body>
         <div id="description">
             Underground visualization
+
             <ul>
-              <li>Option available in "Layer Planar" controls</li>
-              <li>Enable "Dynamic opacity"</li>
-              <li>Zoom-in to look at underground data</li>
-              <li>Dynamic opactiy triggered by the zoom</li>              
-              <li>Can be enabled in globe view</li>              
+                <li>Option available in "Layer Planar" controls</li>
+                <li>Enable "Dynamic opacity"</li>
+                <li>Zoom-in to look at underground data</li>
+                <li>Dynamic opactiy triggered by the zoom</li>
+                <li>Can be enabled in globe view</li>
             </ul>
         </div>
         <div id="viewerDiv" class="viewer"></div>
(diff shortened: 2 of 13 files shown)


gchoqueux:
In terms of software architecture, it's preferable to add visualization specificity outside the core. The use of `MAIN_LOOP_EVENTS` would allow to add this feature while keeping an architecture by bricks (by example `BEFORE_LAYER_UPDATE` main loop event)

jailln:
> In terms of software architecture, it's preferable to add visualization specificity outside the core. The use of `MAIN_LOOP_EVENTS` would allow to add this feature while keeping an architecture by bricks (by example `BEFORE_LAYER_UPDATE` main loop event)

It could indeed be added directly in external projects but we identified it as something that could benefit enough other users to add it in the core

src/Core/View.js
```diff
@@ -291,6 +310,29 @@ class View extends THREE.EventDispatcher {
...
     }
 
+
+
+    /**
+     * Update layer opacity depending on camera position.
+     *
+     * @param {Event} event Event used to get camera target position.
+     */
+    #updateTiledLayerOpacity(event) {
```
gchoqueux: This method would be better in `TiledGeometryLayer`

src/Core/View.js
```diff
@@ -238,6 +240,23 @@ class View extends THREE.EventDispatcher {
...
+            this.altitudeForZeroOpacity = 420;
+        } else {
+            this.altitudeForZeroOpacity = options.altitudeForZeroOpacity;
+        }
+
+        if (options.altitudeForFullOpacity === undefined || options.altitudeForFullOpacity === null) {
+            this.altitudeForFullOpacity = 2400;
+        } else {
+            this.altitudeForFullOpacity = options.altitudeForFullOpacity;
+        }
```
gchoqueux: it's `TiledGeometryLayer` options?

src/Core/View.js
```diff
@@ -291,6 +310,29 @@ class View extends THREE.EventDispatcher {
...
+     *
+     * @param {Event} event Event used to get camera target position.
+     */
+    #updateTiledLayerOpacity(event) {
+        var cameraTargetPosition;
+        if (event && event.coord) {
+            cameraTargetPosition = event.coord;
+        } else {
+            cameraTargetPosition = this.controls.getLookAtCoordinate();
+        }
```
gchoqueux:
```js
cameraTargetPosition = event?.coord || this.controls.getLookAtCoordinate();
```

src/Core/View.js
```diff
@@ -291,6 +310,29 @@ class View extends THREE.EventDispatcher {
...
+        var cameraTargetPosition2 = new Coordinates(cameraTargetPosition.crs, cameraTargetPosition).as('EPSG:4978');
+        var cameraPosition = this.camera.position('EPSG:4978');
+
+        const distance = Math.sqrt(((cameraTargetPosition2.x - cameraPosition.x) * (cameraTargetPosition2.x - cameraPosition.x)  + (cameraTargetPosition2.y - cameraPosition.y) * (cameraTargetPosition2.y - cameraPosition.y) + (cameraTargetPosition2.z - cameraPosition.z) * (cameraTargetPosition2.z - cameraPosition.z)));
```
gchoqueux: could you use `Coordinates.spatialEuclideanDistanceTo()` method ?

src/Layer/TiledGeometryLayer.js
```diff
@@ -266,6 +276,20 @@ class TiledGeometryLayer extends GeometryLayer {
...
+                    if (value) {
+                        obj.geometry.setDrawRange(0, this.segments * this.segments * 2 * 3);  //  bufferIndex = (nSeg) * (nSeg) * 2 * 3 (computeBufferTileGeometry.js)
+                    } else {
+                        obj.geometry.setDrawRange(0, Infinity);
+                    }
+                }
+            });
+        }
+    }
+
```
gchoqueux: you could use `tileMesh.onBeforeRender()` to handle the draw range setting.

src/Core/View.js
AnthonyGlt: Won't it be unusual to call this method from an eventlistener which is defined in `View` ?
I implemented it in `TiledGeometryLayer` but I don't find it optimized like that. 
What do you think ?

src/Controls/PlanarControls.js
gchoqueux, line 1: Could you move this part in other PR, this feature is out the scope

examples/underground_25d_map.html
```diff
@@ -0,0 +1,169 @@
...
+
+            Math.Clamp = function clamp(n, min, max) {
+                if (n > max) {
+                    return max;
+                } else if (n < min) {
+                    return min;
+                } else {
+                    return n;
+                }
+            };
```
gchoqueux: `clamp` is already in three.js

examples/underground_25d_map.html
gchoqueux, line 1: `hideSkirt` isn't option for the viewer?

src/Core/Prefab/GlobeView.js
```diff
@@ -124,9 +124,7 @@ class GlobeView extends View {
             CameraUtils.transformCameraToLookAtTarget(this, this.camera.camera3D, placement);
         } else {
             this.controls = new GlobeControls(this, placement, options.controls);
-            this.controls.handleCollision = typeof (options.handleCollision) !== 'undefined' ? options.handleCollision : true;
```
gchoqueux: it's already in `GlobeControls`?

src/Core/TileMesh.js
```diff
@@ -110,6 +110,12 @@ class TileMesh extends THREE.Mesh {
     }
 
     onBeforeRender() {
+        if (this.layer.hideSkirt) {
+            this.geometry.setDrawRange(0, this.layer.segments * this.layer.segments * 2 * 3);  //  bufferIndex = (nSeg) * (nSeg) * 2 * 3 (computeBufferTileGeometry.js)
+        } else if (this.layer.hideSkirt == false && this.geometry.drawRange.count != Infinity) {
+            this.geometry.setDrawRange(0, Infinity);
+        }
+
```
gchoqueux, line 118: it's my bad, it isn't the good idea, because the draw change is set every draw call.
could use event when `this.layer.hideSkirt` is changed

src/Core/View.js
gchoqueux: could you move in `TiledGeometryLayer`?

examples/underground_25d_map.html
gchoqueux: it's not used?

src/Layer/TiledGeometryLayer.js
```diff
@@ -116,6 +121,23 @@ class TiledGeometryLayer extends GeometryLayer {
...
+     */
+    _updateTiledLayerOpacity(event) {
+        if (event == null) {
+            return;
+        }
+        const view = event.target;
+        var cameraTargetPosition = event?.coord || view.controls.getLookAtCoordinate();
+        var cameraTargetPosition2 = new Coordinates(cameraTargetPosition.crs, cameraTargetPosition);
+        var cameraPosition = view.camera.position('EPSG:4978');
+        const distance = cameraTargetPosition2.spatialEuclideanDistanceTo(cameraPosition);
```
gchoqueux: could use range with the `CameraUtils` or `Controls` ?

src/Core/View.js
```diff
@@ -1112,6 +1132,35 @@ class View extends THREE.EventDispatcher {
...
+
+    /**
+     * Trigger underground visualization
+     * @param {boolean} [trigger] - Enable/Disable the underground visualization.
+     */
+    setUndergroundVisualization(trigger) {
+        const atmo = this.getLayerById('atmosphere');
+        if (trigger) {
+            this.tileLayer.updateTiledLayerOpacity({ target: this });
+            this.addEventListener(VIEW_EVENTS.CAMERA_MOVED,  this.tileLayer.updateTiledLayerOpacity);
```
gchoqueux, line 1132: as now the code is the core, you can put it in the `tileLayer.preUpdate` and not use `addEventListener`

src/Core/Prefab/GlobeView.js
AnthonyGlt: I shouldn't have deleted this line because some examples are putting `handleCollision` parameters directly in view options instead of options.controls. I revert this change

AnthonyGlt:
# IMPORTANT
This issue https://github.com/iTowns/itowns/issues/2101 is affecting this PR. The option `undergroundVisualisation` can't be set at the initialization of the PlanarView.

src/Core/View.js
AnthonyGlt: Good idea ! 
By testing, it causes flickering of opacity though:


https://github.com/iTowns/itowns/assets/126568810/c9239661-c982-4dd7-9bcb-a81e01157ea8

src/Controls/PlanarControls.js
AnthonyGlt: This change could be unnecessary if this is implemented : https://github.com/iTowns/itowns/pull/2077#discussion_r1202624562

examples/underground_25d_map.html
AnthonyGlt: It's a layer option (only used to hide skirt but keep it in "memory"). The viewer option to enable the underground visualisation is the boolean : `undergroundVisualisation`. This will be  passed to the layer as an option too

src/Controls/PlanarControls.js
gchoqueux, line 1: this feature could be on other PR?

src/Utils/CameraUtils.js
gchoqueux, line 1: could be move in other PR?

gchoqueux (commented):
I think this functionality should be in the `TiledGeometryLayer.preUpdate` and that the display specification should be handled  by the `Style` object.

A first step could be to move `layer.opacity` to `layer.Style.opacity` 
the user could manage the opacity according to use case or use prefabricated behaviours

```js
view.tileLayer.style.opacity = (context) => context.range < 1000 ? 0 : 1;
// or with itowns behaviours
view.tileLayer.style.opacity = itowns.VIZ_TYPE.UNDERGROUND;
```

src/Utils/CameraUtils.js
AnthonyGlt: already merged
https://github.com/iTowns/itowns/pull/2107

examples/config.json
```diff
@@ -78,7 +78,9 @@
         "misc_orthographic_camera": "Orthographic camera",
         "misc_custom_controls": "Define custom controls",
         "misc_custom_label": "Custom label popup",
-        "misc_camera_traveling": "Camera traveling"
+        "misc_camera_traveling": "Camera traveling",
+        "underground_25d_map": "Underground visualization"
```
Desplandis, line 82:
```suggestion
        "misc_underground_25d_map": "Underground visualization"
```

Could you please change the example filename to respect the naming convention of the miscellaneous examples?

src/Core/Prefab/TileBuilder.js
```diff
@@ -51,7 +51,9 @@ export default function newTileGeometry(builder, params) {
 
         const geometry = new TileGeometry(params, buffers);
         geometry.OBB = new OBB(geometry.boundingBox.min, geometry.boundingBox.max);
-
+        if (params.hideSkirt) {
+            geometry.setDrawRange(0, params.segment * params.segment * 2 * 3);//  bufferIndex = (nSeg) * (nSeg) * 2 * 3 (computeBufferTileGeometry.js)
+        }
```
Desplandis, line 56: This code is fragile. If there is any change to the geometry computation in `computeBuffers`, this code would then be invalid. Moreover, `params.segment` is an optional parameter of `computeBuffers` and it could potentially be `undefined` (and set to a different default value).

I suggest to use and export a function in `computeBufferTileGeometry.js` to compute the draw range of the tile or the tile plus its skirt. You could then use it in `computeBuffers` and here.

src/Core/TileMesh.js
Desplandis: I agree with @gchoqueux, we should restrict code that is called every frame to the strict necessary.
The draw range shall be set only when `this.layer.hideSkirt` has changed.

I suggest adding a `hideSkirt` accessor in `TileMesh` and change the draw range/offset accordingly when set from the `TiledGeometryLayer` instance.

src/Core/View.js
```diff
@@ -146,7 +146,7 @@ class View extends THREE.EventDispatcher {
      * @param {?Color} options.diffuse - [THREE.Color](https://threejs.org/docs/?q=color#api/en/math/Color) Diffuse color terrain material.
      * This color is applied to terrain if there isn't color layer on terrain extent (by example on pole).
      * @param {boolean} [options.enableFocusOnStart=true] - enable focus on dom element on start.
-     *
```
Desplandis, line 149: This feature could be implemented as an itowns plugin with the proposed Core API changes (`hideSkirt` parameter added in `TileGeometryLayer`). Such plugin could register itself to a camera event and update the opacity of the `TileGeometryLayer` accordingly.
This limits the Core API surface area.

src/Core/View.js
```diff
@@ -291,6 +298,7 @@ class View extends THREE.EventDispatcher {
         this.scene.traverse(ObjectRemovalHelper.cleanup);
     }
 
+
```
Desplandis, line 301:
```suggestion
```

src/Utils/CameraUtils.js
```diff
@@ -337,7 +337,7 @@ class CameraRig extends THREE.Object3D {
 }
 
 export function getRig(camera) {
-    rigs[camera.uuid] = rigs[camera.uuid] || new CameraRig(camera);
+    rigs[camera.uuid] = rigs[camera.uuid] || new CameraRig();
```
Desplandis, line 340: Already changed by #2107, rebase from master.
```suggestion
    rigs[camera.uuid] = rigs[camera.uuid] || new CameraRig(camera);
```

src/Converter/convertToTile.js
```diff
@@ -41,12 +41,14 @@ export default {
         const builder = layer.builder;
         const parent = requester;
         const level = (parent !== undefined) ? (parent.level + 1) : 0;
+        layer.segments = layer.segments || 16;
```
Desplandis, line 44: This code has a side-effect on the caller. Suggestion:

```suggestion
        const segments = layer.segments || 16;
```

src/Converter/convertToTile.js
```diff
@@ -41,12 +41,14 @@ export default {
...
 
         const paramsGeometry = {
             extent,
             level,
-            segment: layer.segments || 16,
+            segment: layer.segments,
```
Desplandis, line 49: By the previous comment:

```suggestion
            segment: segments,
```

src/Core/Prefab/TileBuilder.js
Desplandis: After further review, I suggest to move this code in a `hideSkirt` setter in `TileMesh` to change the draw count and offset. The previous comment holds.

src/Core/TileMesh.js
Desplandis: Moreover, the previous comment on `TileBuilder.js` holds, any change `computeBuffers` could break this code.

src/Converter/convertToTile.js
```diff
@@ -41,12 +41,14 @@ export default {
...
             disableSkirt: layer.disableSkirt,
+            hideSkirt: layer.hideSkirt,
```
Desplandis, line 51: See comment below.
If you move the code of `TileBuilder.js` to `TileMesh.js`, you could remove this line.

src/Layer/TiledGeometryLayer.js
```diff
@@ -97,6 +101,21 @@ class TiledGeometryLayer extends GeometryLayer {
             this.object3d.add(...level0s);
             this.object3d.updateMatrixWorld();
         }));
```
Desplandis, line 103: Missing a default value for `this.hideSkirt`.

src/Controls/PlanarControls.js
```diff
@@ -367,6 +369,7 @@ class PlanarControls extends THREE.EventDispatcher {
         }
         if (onMovement) {
             this.view.dispatchEvent({ type: PLANAR_CONTROL_EVENT.MOVED });
+            this.view.dispatchEvent({ type: VIEW_EVENTS.CAMERA_MOVED });
```
Desplandis, line 372: So this feature will only work with `PlanarControls` ? It would be better to have a generic method that will work with all controls.

Seems that pre-PR, this event was only dispatched when calling `CameraRig#applyTransformToCamera`. I don't know if and when this is triggered with e.g. `GlobeControls`.

@mgermerie @gchoqueux: Any inputs?

src/Core/View.js
```diff
@@ -1112,6 +1120,35 @@ class View extends THREE.EventDispatcher {
             this.notifyChange(this.camera.camera3D);
         }
     }
+
+    /**
+     * Trigger underground visualization
+     * @param {boolean} [trigger] - Enable/Disable the underground visualization.
+     */
+    setUndergroundVisualization(trigger) {
```
Desplandis, line 1128: I think this should be implemented as a plugin (e.g in `examples/plugin`) as this code calls only functions available from the public Core API:
- `View#addEventListener(VIEW_EVENTS, function)`
- `View#getLayerById(String): Layer`
- `View#scene: Scene`
- `View#notifyChange()`
- `GlobeView#tileLayer: TileGeometryLayer` or `PlanarView#tileLayer: TileGeometryLayer` and their `opacity` and `hideSkirt` properties.

src/Layer/TiledGeometryLayer.js
```diff
@@ -7,6 +7,8 @@ import ObjectRemovalHelper from 'Process/ObjectRemovalHelper';
 import { SIZE_DIAGONAL_TEXTURE } from 'Process/LayeredMaterialNodeProcessing';
 import { ImageryLayers } from 'Layer/Layer';
 import { CACHE_POLICIES } from 'Core/Scheduler/Cache';
+import CameraUtils from 'Utils/CameraUtils';
+
```
Desplandis, line 11:
```suggestion
```

src/Layer/TiledGeometryLayer.js
```diff
@@ -116,6 +135,20 @@ class TiledGeometryLayer extends GeometryLayer {
         return Picking.pickTilesAt(view, coordinates, radius, this, target);
     }
 
+    /**
+     * Update layer opacity depending on camera position.
+     *
+     * @param {Event} event Event used to get camera target position.
+     */
+    _updateTiledLayerOpacity(event) {
```
Desplandis, line 143: I think this should be implemented as a plugin (e.g in examples/plugin) as this code calls only functions available from the public Core API:
- The current `View` instance
- `CameraUtils.getTransformCameraLookingAtTarget(View, Camera)`
- The current `Camera` instance
- `TileGeometry#opacity`

src/Controls/PlanarControls.js
Desplandis: I agree.

src/Layer/TiledGeometryLayer.js
Desplandis: Or maybe we could drop the `hideSkirt` property altogether and hide the TileMesh's skirt when `TileGeometryLayer`'s `opacity < 1.0`.

I don't know if this a desirable property. Tagging @mgermerie @gchoqueux for suggestions.

examples/underground_25d_map.html
```diff
@@ -0,0 +1,158 @@
...
+                transparent: false,
+                source: wmsImagerySource,
+            });
+
+            view.addLayer(wmsImageryLayer);
+
+            var wmsElevationSource = new itowns.WMSSource({
+                extent: viewExtent,
+                url: "https://download.data.grandlyon.com/wms/grandlyon",
+                name: "MNT2012_Altitude_10m_CC46",
```
Desplandis, line 91:
```suggestion
                url: "https://imagerie.data.grandlyon.com/wms/grandlyon",
                name: "MNT2018_Altitude_2m",
```

examples/underground_25d_map.html
```diff
@@ -0,0 +1,158 @@
...
+                controls: { minDistance: -50, handleCollision: false }, // To handle underground visualization, disable collision and allow camera below the ground
+            };
+            // Create the planar view
+            const view = new itowns.PlanarView(viewerDiv, viewExtent, options);
+
+            const wmsImagerySource = new itowns.WMSSource({
+                url: "https://download.data.grandlyon.com/wms/grandlyon",
+                networkOptions: { crossOrigin: "anonymous" },
+                version: "1.3.0",
+                name: "Ortho2009_vue_ensemble_16cm_CC46",
```
Desplandis, line 74:
```suggestion
                url: "https://imagerie.data.grandlyon.com/wms/grandlyon",
                networkOptions: { crossOrigin: "anonymous" },
                version: "1.3.0",
                name: "ortho_latest",
```

examples/underground_25d_map.html
```diff
@@ -0,0 +1,158 @@
...
+                    source: lyonTclBusSource,
+
+                    style: new itowns.Style({
+                        stroke: {
+                            base_altitude: baseAltitude,
+                            width: 2,
+                        },
+                    }),
+                }
+            );
```
Desplandis, line 145: Update to new tcl source.

examples/underground_25d_map.html
```diff
@@ -0,0 +1,158 @@
+<!DOCTYPE html>
```
Desplandis, line 1: Could you change the filepath from `examples/underground_25d_map.html` to `examples/misc_underground_25d_map.html` to respect the naming convention of the miscellaneous examples?

Desplandis (changes requested):
Hi and thanks for this PR ! =)

I welcome the proposed addition to the Core API:
- `hideSkirt` property in `TiledGeometryLayer`. It allows to provide good visual results when decreasing the opacity of such layers.

The main issue with this PR is that the dynamic opacity feature could be implemented externally as an `itowns` plugin à la `three.js` `three/addons/*` (see [Addons section](https://threejs.org/docs/#manual/en/introduction/Installation) of the manual). Since this functionality does not use any internal API, this limits the Core API surface area.

I understand that this feature could be useful for some people, so I suggest to add it to the plugins available (in `examples/js/plugins`). We could then export them in the `package.json` (see `three.js` `package.json` for an example). I think there is some WIP on this subject (@mgermerie @jailln) ? 

I added other comments on the `hideSkirt` implementation.

jailln:
> The main issue with this PR is that the dynamic opacity feature could be implemented externally as an `itowns` plugin à la `three.js` `three/addons/*` (see [Addons section](https://threejs.org/docs/#manual/en/introduction/Installation) of the manual). Since this functionality does not use any internal API, this limits the Core API surface area.
> 
> I understand that this feature could be useful for some people, so I suggest to add it to the plugins available (in `examples/js/plugins`). We could then export them in the `package.json` (see `three.js` `package.json` for an example). I think there is some WIP on this subject (@mgermerie @jailln) ?

It could be implemented as an addon but it has been discussed upfront and validateted that it should be implemented in the core since it has been identified as a core rendering functionality and not as an addon (I think you were not yet in the loop at this time). One more argument to track these decisions in issues, we have to get better on that and we will :) Anyway, I think this decision should not be reversed at this point of the implementation / review process.

jailln:
Closing since it has been too long and there has not really been a conscensus on it yet. I'ts a good starting point if this feature is needed in the future though

PR #2079, opened 2023-04-17
doc(WMTSSource): Change the link and the extend to TMSSource

a-dutremble:
## Description

It is not possible to specify the TileMatrix identifier. In WMTSSource, the identifier corresponds to the zoom level (`%TILEMATRIX`). The changes in this PR make it possible to specify a different identifier using a callback function.

## Motivation and Context
Using other specific sources can be useful. For example, using a WMTS flow from a GeoServer. This allows people who know how to use GetCapabilities to specify a different identifier. This is only a temporary solution. The goal is to automatically implement the retrieval of this information in the GetCapabilities as mentioned in #285.

### DIFF
--- a/src/Source/WMTSSource.js
+++ b/src/Source/WMTSSource.js
@@ -4,9 +4,9 @@ import TMSSource from 'Source/TMSSource';
  * @classdesc
  * An object defining the source of resources to get from a
  * [WMTS]{@link http://www.opengeospatial.org/standards/wmts} server. It inherits
- * from {@link Source}.
+ * from {@link TMSSource}.
  *
- * @extends Source
+ * @extends TMSSource
  *
  * @property {boolean} isWMTSSource - Used to checkout whether this source is a
  * WMTSSource. Default is true. You should not change this, as it is used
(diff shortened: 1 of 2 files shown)


src/Source/WMTSSource.js
```diff
@@ -87,7 +89,8 @@ class WMTSSource extends TMSSource {
             '&REQUEST=GetTile' +
             `&STYLE=${source.style || 'normal'}` +
             `&TILEMATRIXSET=${source.tileMatrixSet}` +
-            '&TILEMATRIX=%TILEMATRIX&TILEROW=%ROW&TILECOL=%COL';
+            `&TILEMATRIX=${source.tileMatrix('%TILEMATRIX')}` +
```
gchoqueux: Why you use a function to set tile matrix identifier?

a-dutremble: According to WMTS standard, the TileMatrix parameter is not always the zoom level. For exemple, for GeoServer, it is the crs concatenate with the zoom level (ex: EPSG:4326:14)

gchoqueux: ok, but the callback doesn't fix it because it calls only one time, when the source is instanced.

a-dutremble: Actually, when i wanted to do. It was possible in TMSSource. I updated the doc, so that WMTSSource extends TMSSource and not Source.

mgermerie (approved):

jailln (approved):

PR #2084, opened 2023-04-21
feat(instancing): use instancing for large number of 3d objects

AnthonyGlt:
## Description
Using InstancedMesh, we can display a large number of 3D objetcs (could be from 3d model file)

## Test
Functional and unit tests included
Example using files from https://github.com/iTowns/iTowns2-sample-data/pull/14

## Screenshots
image

### DIFF
--- a/examples/misc_instancing.html
+++ b/examples/misc_instancing.html
@@ -1,0 +1,222 @@
+<html>
+    <head>
+        <title>Itowns - Instancing</title>
+        <meta charset="UTF-8" />
+        <link rel="stylesheet" type="text/css" href="css/example.css" />
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css" />
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="description">
+            <p>
+                <b>
+                    Render a large number of objects with the same geometry<br/>
+                    by importing your 3D models
+                </b>
+            </p>
+        </div>
+        <div id="viewerDiv" class="viewer"></div>
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script type="text/javascript">
+            var THREE = itowns.THREE;
+
+            // Define initial camera position
+            var placement = {
+                coord: new itowns.Coordinates("EPSG:4326", -0.57918, 44.837789),
+                range: 1000,
+                tilt: 45,
+            };
+
+            // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+            var viewerDiv = document.getElementById("viewerDiv");
+
+            // Instanciate iTowns GlobeView*
+            var view = new itowns.GlobeView(viewerDiv, placement);
+
+            // Setup loading screen and debug menu
+            setupLoadingScreen(viewerDiv, view);
+            const debugMenu = new GuiTools("menuDiv", view);
+
+            var ambLight = new itowns.THREE.AmbientLight(0xffffff, 0.2);
+            view.scene.add(ambLight);
+
+            // Add one imagery layer to the scene
+            itowns.Fetcher.json("./layers/JSONLayers/Ortho.json").then(
+                function _(config) {
+                    config.source = new itowns.WMTSSource(config.source);
+                    var layer = new itowns.ColorLayer("Ortho", config);
+                    view.addLayer(layer).then(
+                        debugMenu.addLayerGUI.bind(debugMenu)
+                    );
+                }
+            );
+
+            //Tree
+            const trunkRadius = 5;
+            const trunkHeight = 20;
+            const topHeight = 10;
+
+            function makeTree() {
+                const root = new THREE.Object3D();
+
+                // Trunk
+                const geometry = new THREE.CylinderGeometry(
+                    trunkRadius,
+                    trunkRadius,
+                    trunkHeight,
+                    32
+                );
+                const material = new THREE.MeshPhongMaterial({
+                    color: 0x8b4513,
+                });
+                const trunk = new THREE.Mesh(geometry, material);
+                trunk.rotateX(Math.PI / 2);
+                trunk.position.z = 10;
+                trunk.updateMatrix();
+                root.add(trunk);
+
+                // Canopy
+                const geometryCanop = new THREE.SphereGeometry(
+                    topHeight,
+                    topHeight,
+                    10
+                );
+                const materialCanop = new THREE.MeshPhongMaterial({
+                    color: 0x00aa00,
+                });
+                const top = new THREE.Mesh(geometryCanop, materialCanop);
+                top.position.z = trunkHeight - topHeight / 3 + 10;
+                top.updateMatrix();
+                root.add(top);
+
+                return root;
+            }
+
+
+
+            // ---------- DISPLAY VECTOR TILED BUILDING DATA AS 3D MESHES : ----------
+
+            // Define the source of the building data : those are vector tiled data from the geoportail.
+            const buildingsSource = new itowns.VectorTilesSource({
+                style: "https://wxs.ign.fr/essentiels/static/vectorTiles/styles/PLAN.IGN/standard.json",
+                // We only want to display buildings related data.
+                filter: (layer) => {
+                    return (
+                        layer["source-layer"].includes("bati_surf") &&
+                        layer.paint["fill-color"]
+                    );
+                },
+            });
+
+            // Create a FeatureGeometryLayer to support building data.
+            var buildingsLayer = new itowns.FeatureGeometryLayer("VTBuilding", {
+                source: buildingsSource,
+                zoom: { min: 15 },
+                accurate: false,
+                style: new itowns.Style({
+                    fill: {
+                        base_altitude: () => 0,
+                        extrusion_height: (p) => p.hauteur || 0,
+                    },
+                }),
+            });
+
+            // Add the FeatureGeometryLayer to the scene and to the debug menu.
+            view.addLayer(buildingsLayer).then((layer) => {
+                const gui = debug.GeometryDebug.createGeometryDebugUI(
+                    debugMenu.gui,
+                    view,
+                    layer
+                );
+                debug.GeometryDebug.addWireFrameCheckbox(gui, view, layer);
+            });
+
+            // Lights
+            var lightsSource = new itowns.FileSource({
+                url: "https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/geojson/points_lumineux_bordeaux.geojson",
+                crs: "EPSG:4326",
+                fetcher: itowns.Fetcher.json,
+                parser: itowns.GeoJsonParser.parse,
+            });
+
+            // Load a glTF resource
+            itowns.glTFLoader.load(
+                // resource URL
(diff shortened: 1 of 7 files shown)


examples/misc_instancing.html
```diff
@@ -0,0 +1,210 @@
...
+                (gltf) => {
+                    var model = gltf.scene;
+
+                    model.rotateX(Math.PI/2.0);
+                    gltf.scene.position.z = 5 ;
+                    model.scale.set(6, 6, 6);
+                            
+                var styleModel3D = new itowns.Style({
+                    model: { object: model  }
+                });
```
gchoqueux: we could add `model` in style point? if the lights are placed with Coordinates?

src/Converter/Feature2Mesh.js
```diff
@@ -481,12 +495,59 @@ function featureToExtrudedPolygon(feature, options) {
...
+    return instancedMesh;
+}
+
+/**
+ * Convert a [Feature]{@link Feature} to a Instanced 3d Model
+ *
+ * @param {Object} feature
+ * @returns {THREE.Mesh} mesh or GROUP of THREE.InstancedMesh
+ */
+function featureTo3DModel(feature) {
```
gchoqueux: it's more points/coordinates to  instanced meshes?

because we could imagine the same feature with a polygon

test/data/geojson/trees.geojson.json
```diff
@@ -0,0 +1,12 @@
+{
```
gchoqueux: could you move https://github.com/iTowns/itowns-sample-data ?

AnthonyGlt: > could you move https://github.com/iTowns/itowns-sample-data ?

I can use the existing test/data/geojson/points.geojson.json so I won't have to fetch data.
Since the other unit tests use geojson from the iTowns repo (not from the sample repo), I could also reduce the size of my current geojson.

src/Converter/Feature2Mesh.js
AnthonyGlt: Correct, we are previously checking if feature.type == FEATURE_TYPES.POINT.
It will be more accurate to name it pointsToInstancedMeshes

src/Converter/Feature2Mesh.js
```diff
@@ -481,12 +495,59 @@ function featureToExtrudedPolygon(feature, options) {
...
+    if (modelObject instanceof THREE.Mesh) {
+        return createInstancedMesh(modelObject, count, ptsIn);
+    } else if (modelObject instanceof THREE.Object3D) {
+        const group = new THREE.Group();
+        // Get independent meshes from more complexe object
+        var meshes = separateMeshes(modelObject);
+        meshes.forEach(mesh => group.add(createInstancedMesh(mesh, count, ptsIn)));
+        return group;
+    } else {
+        console.error('Format not supported');
```
mgermerie: Could you specify a bit more context in the error message so that users know the unsupported format relates the `style.point.model.object` they set ?

src/Converter/Feature2Mesh.js
```diff
@@ -496,25 +557,33 @@ function featureToMesh(feature, options) {
     var mesh;
     switch (feature.type) {
         case FEATURE_TYPES.POINT:
-            mesh = featureToPoint(feature, options);
+            if (feature.style.point && feature.style.point.model && feature.style.point.model.object) {
```
mgermerie: You could simplify this with : 
```js
if (feature.style.point?.model?.object)
```

src/Converter/Feature2Mesh.js
```diff
@@ -496,25 +557,33 @@ function featureToMesh(feature, options) {
...
+                mesh = pointsToInstancedMeshes(feature);
+            } else {
+                mesh = featureToPoint(feature, options);
+                mesh.material.vertexColors = true;
+                mesh.material.color = new THREE.Color(0xffffff);
```
mgermerie: Maybe instead of repeating these for every case of `feature.type`, we could add 
```js
if (!feature.style.point?.model?.object) {
    mesh.material.vertexColors = true;
    mesh.material.color = new THREE.Color(0xffffff);
}
```
on line 586 just after the switch cases.

Another option would be to add 
```js
mesh.isInstancedMesh = true  // or some name like that
```
just after line 561 and then add 
```js
if (!mesh.isInstancedMesh) {
    mesh.material.vertexColors = true;
    mesh.material.color = new THREE.Color(0xffffff);
}
```
in line 586.

I know these solution are not perfect, but I wonder what's the best between adding one of these checks on line 586 or setting the same options with the same hard coded values multiple times. WDYT ?

src/Core/Style.js
```diff
@@ -345,6 +350,10 @@ class Style {
         defineStyleProperty(this, 'point', 'radius', params.point.radius, 2.0);
         defineStyleProperty(this, 'point', 'width', params.point.width, 0.0);
         defineStyleProperty(this, 'point', 'base_altitude', params.point.base_altitude, base_altitudeDefault);
+        defineStyleProperty(this, 'point', 'model', params.point.model);
+
+        this.light = {};
+        defineStyleProperty(this, 'light', 'object', params.light.object);
```
mgermerie: Why would we add light to itowns Style ? I might be missing something, but I think this should be removed from the commit of this PR as it is not within its scope.

src/Parser/GeoJsonParser.js
```diff
@@ -144,6 +144,12 @@ const keyProperties = ['type', 'geometry', 'properties'];
 const firstCoordinates = a => (a === undefined || (Array.isArray(a) && !isNaN(a[0])) ? a : firstCoordinates(a[0]));
 
 function jsonFeatureToFeature(crsIn, json, collection) {
+    if (!json.geometry || !json.geometry.type) {
+        console.warn('No geometry provided');
+        // throw new Error(`No geometry provided ${json}`);
```
mgermerie: Could you simplify with `if (!json.geometry?.type)` and remove comment on line 149 ?

examples/misc_instancing.html
```diff
@@ -0,0 +1,231 @@
...
+            var THREE = itowns.THREE;
+        </script>
+
+        <script type="text/javascript">
+            // Define initial camera position
+            var positionOnGlobe = {
+                longitude: -0.57918,
+                latitude: 44.837789,
+                altitude: 3000,
+            };
```
mgermerie: Can you remove this unused variable ?

examples/misc_instancing.html
```diff
@@ -0,0 +1,231 @@
...
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+
+    <body style="background-color: black">
+        <div id="description">
+            <p>
+                <b
+                    >Render a large number of objects with the same geometry<br />
```
mgermerie: Can you remove the `<br />` markup which has no effect here ?

examples/misc_instancing.html
```diff
@@ -0,0 +1,231 @@
...
+        <title>Itowns - Instancing</title>
+        <meta charset="UTF-8" />
+        <link rel="stylesheet" type="text/css" href="css/example.css" />
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css" />
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+
+    <body style="background-color: black">
```
mgermerie: Why setting this style ? If this is needed, could you move it in css file ?

examples/misc_instancing.html
```diff
@@ -0,0 +1,231 @@
...
+                top.position.z = trunkHeight - topHeight / 3 + 10;
+                top.updateMatrix();
+                root.add(top);
+
+                return root;
+            }
+
+            // Load a gltf model
+            // Instantiate a loader
+            const loader = itowns.glTFLoader;
```
mgermerie: Could you move this on line 157 ? Moreover, as it is only used once we may not need to create a `const` for that.

mgermerie (commented):
Thanks for this useful feature !
Except from a few comments I made (mostly on format and a few on substance) it should be good for me.

examples/misc_instancing.html
AnthonyGlt: I think it looks better with a line break:
Screenshot from 2023-06-16 09-59-21
Screenshot from 2023-06-16 09-59-33

examples/misc_instancing.html
AnthonyGlt: Not needed you're right

examples/misc_instancing.html
mgermerie: Indeed, let's keep it that way. As I checked it with the navigator console opened, the width of the description was reduced and the line breaking was forced (by the `max-width: 40%;` in css) so I did not see a difference.

src/Converter/Feature2Mesh.js
AnthonyGlt:
```suggestion
            console.error(`The format of the model object provided in the feature style is not supported: ${feature.style.point.model.object.type}`);

```

AnthonyGlt: Like that
```suggestion
        console.error('The format of the model object provided in the feature style (feature.style.point.model.object) is not supported. Only THREE.Mesh or THREE.Object3D are supported.');

```

AnthonyGlt: But it should be good to handle the error. Leaving it like that will make the following `mesh.feature = feature;` crashed.
What could be done in this case in your opinion ?

mgermerie: I was thinking of setting a default mesh in this case, but perhaps it would be better to just throw an error (with `throw new Error('...')` in order to stop code execution and still get the error message in the console (yet preventing the `mesh.feature = feature;` crash). WDYT ?

jailln: If it's not too hard to implement: what about logging a warning (or an error?) and displaying the geometries as simple points without style? WDYT?

src/Converter/Feature2Mesh.js
AnthonyGlt: I think I'll implement the 2nd option. The flag could be reused or adapted for another case, if needed.

src/Converter/Feature2Mesh.js
mgermerie: Yep that could be solution. About implementation, perhaps you could just modify `feature.style.point` to a simple point and then call `featureToPoint` to get a simple point mesh, but you would need to pass the `options` parameter of `featureToMesh` to `pointsToInstancedMesh` method. Moreover this could cause side effects on `feature.style` that I am not aware of. @ftoromanoff as you worked on `Style` ans `feature2Mesh` recently, do you have an idea on this ?

AnthonyGlt:
```suggestion
        throw new Error('The format of the model object provided in the feature style (feature.style.point.model.object) is not supported. Only THREE.Mesh or THREE.Object3D are supported.');

```
Then:

```

 try {
    mesh = pointsToInstancedMeshes(feature);
    mesh.isInstancedMesh = true;
} catch (e) {
    console.error(e);
    mesh = featureToPoint(feature, options);
}
```

src/Converter/Feature2Mesh.js
```diff
@@ -496,7 +557,17 @@ function featureToMesh(feature, options) {
     var mesh;
     switch (feature.type) {
         case FEATURE_TYPES.POINT:
-            mesh = featureToPoint(feature, options);
+            if (feature.style.point?.model?.object) {
+                try {
+                    mesh = pointsToInstancedMeshes(feature);
+                    mesh.isInstancedMesh = true;
+                } catch (e) {
+                    console.error(e);
```
mgermerie: Just thought about something : do we need to log the error since the error message is thrown above ?

mgermerie (approved):

PR #2089, opened 2023-05-02
refactor(test): handling assert fail messages in tests

AnthonyGlt:
## Description
Catch error when assert is false and avoid timeout.

## Motivation and Context
I was getting a timeout when the assert was wrong for unit tests.

Example of timeout: demutils.js

```
it('add elevation layer', (done) => {
        viewer.addLayer(elevationlayer)
            .then((l) => {
                assert.equal('worldelevation1', l.id);
                done();
            }, done);
    });
 

``` 
I've replaced _worldelevation_ with _worldelevation1_ and it will timeout instead of returning the error.

Following the commit from:
https://github.com/iTowns/itowns/pull/2044

### DIFF
--- a/test/unit/featuregeometrylayererror.js
+++ b/test/unit/featuregeometrylayererror.js
@@ -68,12 +68,12 @@ files.forEach((geojson, i) => {
             layerProj4.whenReady
                 .then(() => {
                     tile.visible = true;
-                    layerProj4.update(context, layerProj4, tile)
+                    return layerProj4.update(context, layerProj4, tile)
                         .then(() => {
                             assert.equal(layerProj4.object3d.children.length, 1);
                             done();
-                        }, done);
-                }, done);
+                        });
+                }).catch(done);
         });
 
         it('update without proj4', function (done) {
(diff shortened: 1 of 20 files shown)


test/unit/featuregeometrylayererror.js
```diff
@@ -72,8 +72,8 @@ files.forEach((geojson, i) => {
                         .then(() => {
                             assert.equal(layerProj4.object3d.children.length, 1);
                             done();
-                        }, done);
-                }, done);
+                        }).catch(done);
```
ftoromanoff: Is this catch() really neccessary as there is another catch() just after ?

test/unit/potree.js
```diff
@@ -99,8 +99,8 @@ describe('Potree', function () {
                         .then(() => {
                             assert.equal(2, root.children[0].children.length);
                             done();
-                        });
-                }, done);
+                        }).catch(done);
+                }).catch(done);
```
ftoromanoff: same

test/unit/featuregeometrylayererror.js
AnthonyGlt: We need to catch each _then_ error or else the parents' _then_ will timeout. Like this, the error is catched by each _then_ and returned.

AnthonyGlt: I will do a _return_ when I can

ftoromanoff (approved):

PR #2094, opened 2023-05-10
view.pickFeaturesAt() -> avoid picking twice the same featureGeometry

### LINKED ISSUE #2093: PickFeatureAt: Duplication of Feature
A bug in view.pickFeatureAt() make it, in some special case, send twice the same featureGeometry.
That appear when when request for a point geometry which is situated close to the tile border as each tile is review separately and the  radius can make if find a point which isn't on this specific tile.

Example:
image

Here we can see that the plugin FeatureToolBox using the pickFeatureAt() show the duplication of the properties as if 2 distinct features got picked.

ftoromanoff:
link with the issue #2093

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -836,8 +836,9 @@ class View extends THREE.EventDispatcher {
     }
 
     /**
-     * Searches for {@link Feature} in {@link ColorLayer}, under the mouse of at
-     * a specified coordinates, in this view.
+     * Searches for {@link FeatureGeometry} in {@link ColorLayer}, under the mouse or at
+     * the specified coordinates, in this view. Combining them per layer and in a Feature
+     * like format.
      *
      * @param {Object} mouseOrEvt - Mouse position in window coordinates (from
      * the top left corner of the window) or `MouseEvent` or `TouchEvent`.
@@ -847,10 +848,15 @@ class View extends THREE.EventDispatcher {
      * into. If not specified, all {@link ColorLayer} and {@link GeometryLayer}
      * layers of this view will be looked in.
      *
-     * @return {Object} - An object, with a property per layer. For example,
-     * looking for features on layers `wfsBuilding` and `wfsRoads` will give an
-     * object like `{ wfsBuilding: [...], wfsRoads: [] }`. Each property is made
-     * of an array, that can be empty or filled with found features.
+     * @return {Object} - An object, having one property per layer.
+     * For example, looking for features on layers `wfsBuilding` and `wfsRoads`
+     * will give an object like `{ wfsBuilding: [...], wfsRoads: [] }`.
+     * Each property is made of an array, that can be empty or filled with
+     * Feature like objects composed of:
+     * - the FeatureGeometry
+     * - the feature type
+     * - the style
+     * - the coordinate if the FeatureGeometry is a point
      *
      * @example
      * view.pickFeaturesAt({ x, y });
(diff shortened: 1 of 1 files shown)


gchoqueux (commented):
may be modify `FeaturesUtils.isFeatureUnderCoordinate` to avoid same feature adding?

ftoromanoff:
> may be modify `FeaturesUtils.isFeatureUnderCoordinate` to avoid same feature adding?

At first I was thinking the same, but there is no probleme with it. The duplicatation comes from:
```for (const tile of tiles)``` at View.js line 914.
Or if we went this it would be necessary to add the tile extent to truncate the reserch circle at the tile extent, but i tought it would be more complicated..

**Comment by gchoqueux** [comment]
may be the issue come from `pickTilesAt` that returns too much tiles?

File: src/Core/View.js [hunk 1]
```diff
@@ -920,8 +926,12 @@ class View extends THREE.EventDispatcher {
 
                     precision = CRS.isMetricUnit(texture.features.crs) ? precisions.M : precisions.D;
 
-                    result[materialLayer.id] = result[materialLayer.id].concat(
-                        FeaturesUtils.filterFeaturesUnderCoordinate(coordinates, texture.features, precision));
+                    const featuresUnderCoor = FeaturesUtils.filterFeaturesUnderCoordinate(coordinates, texture.features, precision);
+                    featuresUnderCoor.forEach((feature) => {
+                        if (!result[materialLayer.id].map(f => f.geometry).includes(feature.geometry)) {
```
gchoqueux: could you use `Array.find`?

ftoromanoff: Done

gchoqueux (approved):

src/Core/View.js
```diff
@@ -920,8 +926,12 @@ class View extends THREE.EventDispatcher {
 
                     precision = CRS.isMetricUnit(texture.features.crs) ? precisions.M : precisions.D;
 
-                    result[materialLayer.id] = result[materialLayer.id].concat(
-                        FeaturesUtils.filterFeaturesUnderCoordinate(coordinates, texture.features, precision));
+                    const featuresUnderCoor = FeaturesUtils.filterFeaturesUnderCoordinate(coordinates, texture.features, precision);
+                    featuresUnderCoor.forEach((feature) => {
+                        if (!result[materialLayer.id].find(f => f.geometry === feature.geometry)) {
```
jailln, line 935: Shouldn't we add an id to `Feature` (that  would be generated) to simplify such comparisons ? (can be done in another PR)

src/Core/View.js
```diff
@@ -832,8 +832,9 @@ class View extends THREE.EventDispatcher {
     }
 
     /**
-     * Searches for {@link Feature} in {@link ColorLayer}, under the mouse of at
-     * a specified coordinates, in this view.
+     * Searches for {@link FeatureGeometry} in {@link ColorLayer}, under the mouse of at
+     * a specified coordinates, in this view. Combining them per layer and in a Feature
+     * like format.
```
jailln, line 841: small typos: 
`of` -> `or`
`a specified coordinates` -> `a specified coordinate`

ftoromanoff: corrected

src/Core/View.js
ftoromanoff: I thought about it as well, but not sure it is will easy to set up ans usefull for anything else.

jailln (approved):

PR #2096, opened 2023-05-10
feat(3DTiles) add C3DTILES_LAYER_EVENTS.ON_TILE_REQUESTED

valentinMachado:
## Description
When 3DTilesProcessing is requesting new tile, the C3DTilesLayer dispatch an event ON_TILE_REQUESTED.

## Motivation and Context
The motivation here is to allow itowns user to know when a 3DTiles is loading (has been requested) allowing to display feedbacks.

SideBar Widget example

### DIFF
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -12,9 +12,17 @@ export const C3DTILES_LAYER_EVENTS = {
     /**
      * Fires when a tile content has been loaded
      * @event C3DTilesLayer#on-tile-content-loaded
-     * @property type {string} on-tile-content-loaded
+     * @type {object}
+     * @property {THREE.Object3D} tileContent - object3D of the tile
      */
     ON_TILE_CONTENT_LOADED: 'on-tile-content-loaded',
+    /**
+     * Fires when a tile is requested
+     * @event C3DTilesLayer#on-tile-requested
+     * @type {object}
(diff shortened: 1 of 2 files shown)


src/Layer/C3DTilesLayer.js
```diff
@@ -15,6 +15,12 @@ export const C3DTILES_LAYER_EVENTS = {
      * @property type {string} on-tile-content-loaded
      */
     ON_TILE_CONTENT_LOADED: 'on-tile-content-loaded',
+    /**
+     * Fires when a tile is requested
+     * @event C3DTilesLayer#on-tile-requested
+     * @property type {string} on-tile-requested
+     */
```
jailln, line 24: Could you specify what data is dispatched with the event ? (and could you also specify it for `ON_TILE_CONTENT_LOADED` please?)

valentinMachado: Actually in both case the data dispatched are not itowns class. For `ON_TILE_CONTENT_LOADED` this is the tile content something like that would be good ? `@param {THREE.Object3D} tileContent - object3D of the tile`, and for `ON_TILE_REQUESTED` this is a tile `@param {object} metadata - tile requested` ?

jailln: yes perfect thanks :+1:

jailln (approved):

PR #2097, opened 2023-05-12
fix: 'maxDeltaElevationLevel' parameter name problem in `GlobeLayer` and `PlanarLayer`, resulting in `ElevationLayer` subdivise the node too much;

mini-mimikyu:


ftoromanoff:
Hello @VeinKowal,
Thanks for this contributing to Itowns.
Would it be possible before we fully review your PR, to add a description explaining what is the purpose of this PR and the changes/evolutions added as well as squashing the irrelevant commits ? 
Or in the case you are still working on it, can you change its status to draft ?
Thanks in advance !

jailln:
Thanks for your contribution, I rebased and squashed your commits and pushed it in #2119 .

PR #2107, opened 2023-06-01
remove camera argument from CameraRig

AnthonyGlt:
Useless `camera` argument because the constructor doesn't take argument and a new camera is assigned.

Will it be useful to keep the camera instead of creating a new one ?

### DIFF
--- a/src/Utils/CameraUtils.js
+++ b/src/Utils/CameraUtils.js
@@ -337,7 +337,7 @@ class CameraRig extends THREE.Object3D {
 }
 
 export function getRig(camera) {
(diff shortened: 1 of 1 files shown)


mgermerie (approved):
Thanks for the cleaning !

About keeping the camera instance instead of creating a new one, I thing form reading `CameraRig` code that we should create a new one (as it is done now). It seems it is more a calculation tool on which transformations are applied, and wee don't necessarily want these transformations on the actual `view.camera.camera3D`.
However I admit I am not fully familiar with this part of iTowns, so I might be missing some points. Perhaps this thought should be investigated.

PR #2108, opened 2023-06-01
feat(underground#WIP): dynamic opacity to visualize underground and add hideSkirt parameter

AnthonyGlt:
2nd Version of https://github.com/iTowns/itowns/pull/2077
Not completely optimized.
The underground option can't be enabled by default but it can be once the view is loaded (using a button)

### DIFF
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -374,12 +374,9 @@ class PlanarControls extends THREE.EventDispatcher {
         }
         if (onMovement) {
             this.view.dispatchEvent({ type: PLANAR_CONTROL_EVENT.MOVED });
+            this.view.dispatchEvent({ type: VIEW_EVENTS.CAMERA_MOVED });
         }
         deltaMousePosition.set(0, 0);
-
-        this.view.dispatchEvent({
-            type: VIEW_EVENTS.CAMERA_MOVED,
-        });
     }
 
     /**
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -1,5 +1,7 @@
 import * as THREE from 'three';
 import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
+import { VIEW_EVENTS } from 'Core/View';
+import CameraUtils from 'Utils/CameraUtils';
 
 // event keycode
 export const keys = {
@@ -374,6 +376,10 @@ class PlanarControls extends THREE.EventDispatcher {
             this.view.dispatchEvent({ type: PLANAR_CONTROL_EVENT.MOVED });
         }
         deltaMousePosition.set(0, 0);
+
+        this.view.dispatchEvent({
+            type: VIEW_EVENTS.CAMERA_MOVED,
+        });
     }
 
     /**
@@ -420,6 +426,15 @@ class PlanarControls extends THREE.EventDispatcher {
      */
     initiatePan() {
         this.state = STATE.PAN;
+    }
+
+
+    /**
+     * Returns the {@linkcode Coordinates} of the central point on screen in the crs of the view. See {@linkcode Coordinates} for conversion.
+     * @return {Coordinates} coordinate
+     */
+    getLookAtCoordinate() {
+        return CameraUtils.getTransformCameraLookingAtTarget(this.view, this.camera).coord;
     }
 
     /**
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -425,15 +425,6 @@ class PlanarControls extends THREE.EventDispatcher {
         this.state = STATE.PAN;
     }
 
-
-    /**
-     * Returns the {@linkcode Coordinates} of the central point on screen in the crs of the view. See {@linkcode Coordinates} for conversion.
-     * @return {Coordinates} coordinate
-     */
-    getLookAtCoordinate() {
-        return CameraUtils.getTransformCameraLookingAtTarget(this.view, this.camera).coord;
-    }
-
     /**
      * Handle the pan movement (translation on local x / world z plane) when user moves the mouse while
      * STATE.PAN. The drag movement is previously initiated by [initiatePan]{@link PlanarControls#initiatePan}.
--- a/src/Controls/PlanarControls.js
+++ b/src/Controls/PlanarControls.js
@@ -1,7 +1,6 @@
 import * as THREE from 'three';
 import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
 import { VIEW_EVENTS } from 'Core/View';
-import CameraUtils from 'Utils/CameraUtils';
 
 // event keycode
 export const keys = {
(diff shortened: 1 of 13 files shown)


src/Controls/PlanarControls.js
```diff
@@ -372,6 +373,7 @@ class PlanarControls extends THREE.EventDispatcher {
         }
         if (onMovement) {
             this.view.dispatchEvent({ type: PLANAR_CONTROL_EVENT.MOVED });
+            this.view.dispatchEvent({ type: VIEW_EVENTS.CAMERA_MOVED });
```
AnthonyGlt, line 376: this could be in another PR. This event is dispatch in GlobeControls when there is movement, it should be dispatch from PlanarControls too

src/Utils/CameraUtils.js
```diff
@@ -101,13 +100,6 @@ class CameraRig extends THREE.Object3D {
         } else {
             this.camera.matrixWorld.decompose(camera.position, camera.quaternion, camera.scale);
         }
-        view.dispatchEvent({
```
AnthonyGlt, line 104: this is removed because it's dispatched in the controls (Planar/GlobeControls) directly
Like this, we don't dispatch the event when there is tiles loading or other updates without camera movements

src/Controls/PlanarControls.js
mgermerie: Indeed it is a good point to dispatch the same event here. However, what could be done in another PR is to dispatch the same properties as with `GlobeControls` here (talking about `range`, `heading` etc).
This should however be done when working on controls harmonization.

src/Core/View.js
```diff
@@ -1130,6 +1137,35 @@ class View extends THREE.EventDispatcher {
...
     }
+
+    /**
+     * Trigger underground visualization
+     * @param {boolean} [trigger] - Enable/Disable the underground visualization.
+     */
+    setUndergroundVisualization(trigger) {
+        const atmo = this.getLayerById('atmosphere');
+        if (trigger) {
+            this.tileLayer.updateTiledLayerOpacity({ target: this });
```
mgermerie, line 1148: Perhaps we should add a return point if `this.tileLayer` is not defined (when user is not using a prefab `View`) ? As it is done in `View.getPickingPositionFromDepth`. WDYT ?

src/Layer/TiledGeometryLayer.js
```diff
@@ -97,8 +101,34 @@ class TiledGeometryLayer extends GeometryLayer {
...
+                _hideSkirt = value;
+                this.#hideExistingSkirt(value);
+            },
+        });
+
+        if (config.altitudeForZeroOpacity === undefined || config.altitudeForZeroOpacity === null) {
+            this.altitudeForZeroOpacity = 420;
+        } else {
+            this.altitudeForZeroOpacity = config.altitudeForZeroOpacity;
+        }
```
mgermerie, line 121: Could you simplify with `this.altitudeForZeroOpacity = config.altitudeForZeroOpacity ?? 420;` ?

src/Layer/TiledGeometryLayer.js
```diff
@@ -118,6 +148,20 @@ class TiledGeometryLayer extends GeometryLayer {
...
 
+    /**
+     * Update layer opacity depending on camera position.
+     *
+     * @param {Event} event Event used to get camera target position.
+     */
+    _updateTiledLayerOpacity(event) {
+        if (event == null) {
+            return;
+        }
```
mgermerie, line 159: As the method is called through an event listener, is it possible that `event` might actually be `null` ?

src/Layer/TiledGeometryLayer.js
```diff
@@ -118,6 +148,20 @@ class TiledGeometryLayer extends GeometryLayer {
...
+        const view = event.target;
+        const distance = CameraUtils.getTransformCameraLookingAtTarget(view, view.controls.camera).range;
```
mgermerie, line 161: I just thought about orthographic camera, for which the `range` has no meaning (since only the fov of the camera is changed). 
@AnthonyGlt, @jailln, do you think we should work (in another PR) on adapting the dynamic opacity feature to orthographic cameras or else adapting CameraUtils so that range is computed from camera fov (which would rigorously be a wrong value but would ease development for features that must support both perspective and orthographic cameras) ?

src/Utils/CameraUtils.js
mgermerie: This was dispatched here because, if I remember well, `CameraUtils` can move the camera without the use of `Controls`. Therefore, dispatching it here allows user to be aware of a camera movement if it originates from `CameraUtils`.
The typical use case is for the Navigation widget, which has to update its compass and 2D/3D toggle if the camera is moved through `CameraUtils`.

src/Core/View.js
```diff
@@ -1130,6 +1137,35 @@ class View extends THREE.EventDispatcher {
...
+            this.addEventListener(VIEW_EVENTS.CAMERA_MOVED,  this.tileLayer.updateTiledLayerOpacity);
+            this.tileLayer.hideSkirt = true;
+
+
+            if (atmo) {
+                this.scene.background = new THREE.Color(0x000000);
```
mgermerie, line 1154: Why do we need to set this ? The default value is `null` which already results in a black background if I'm wright. Also, if setting this here, we should set it back to null on line 1162.
Moreover, by testing on `view_3d_map` example, switching `atmo.visible` does not seem to have an effect.

src/Layer/TiledGeometryLayer.js
```diff
@@ -97,8 +101,34 @@ class TiledGeometryLayer extends GeometryLayer {
...
+        // configure dynamic opacity
+
+        let _hideSkirt = this.hideSkirt;
+
+
+        Object.defineProperty(this, 'hideSkirt', {
+            get: () => _hideSkirt,
+            set: (value) => {
+                _hideSkirt = value;
+                this.#hideExistingSkirt(value);
```
mgermerie, line 113: Perhaps we should directly add the method's content here instead of defining a method if it is only called here. WDYT ?

mgermerie (commented):
Thanks for the PR.
The `TileLayer` options are only visible while in debug mode if I'm correct. Could you therefore add a button on your example (a bit like what you did for the 3dTiles classification example) to toggle dynamic opacity so that it appears on the website (which is not in debug mode) ?

src/Utils/CameraUtils.js
AnthonyGlt: Oh... You're right, but this dispatch is trigger too often (like I said, when tiles are loading or event when there is no movement). I'll see how I can do

src/Core/View.js
AnthonyGlt: In `view_3d_map` for example, when I don't change the background we get:
noBACK
When I don't change the atmo:
noAtmo
When I don't change anything:
noThing

With my edits:
Screenshot from 2023-07-26 17-26-45

It will be better to see the data I think.

Although, you are right about switching back to the previous value at line 1162

jailln:
Closing since #2077 has been closed too. Also a good starting point if this feature is needed in the future

PR #2109, opened 2023-06-02
More tests on crs parameters

jailln:
Add function parameters tests to avoid errors such as `crs.startsWith is not a function`

### DIFF
--- a/src/Core/Geographic/Crs.js
+++ b/src/Core/Geographic/Crs.js
@@ -2,19 +2,31 @@ import proj4 from 'proj4';
 
 proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
 
+function isString(s) {
+    return typeof s === 'string' || s instanceof String;
+}
+
+function mustBeString(crs) {
+    if (!isString(crs)) {
+        throw new Error(`Crs parameter value must be a string: '${crs}'`);
+    }
+}
+
 function isTms(crs) {
-    return crs.startsWith('TMS');
+    return isString(crs) && crs.startsWith('TMS');
 }
 
 function isEpsg(crs) {
-    return crs.startsWith('EPSG');
+    return isString(crs) && crs.startsWith('EPSG');
 }
 
 function formatToTms(crs) {
+    mustBeString(crs);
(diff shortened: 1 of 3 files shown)


src/Core/Geographic/Crs.js
```diff
@@ -2,19 +2,29 @@ import proj4 from 'proj4';
...
 
 function isEpsg(crs) {
-    return crs.startsWith('EPSG');
+    return isString(crs) && crs.startsWith('EPSG');
 }
 
 function formatToTms(crs) {
+    if (!isString(crs)) {
+        throw new Error(`Crs parameter value must be a string: '${crs}'`);
+    }
```
gchoqueux: may be encapsulate in method `mustBeString`?

jailln: yes, done

src/Source/TMSSource.js
```diff
@@ -14,7 +14,7 @@ const extent = new Extent(CRS.tms_4326, 0, 0, 0);
  * @extends Source
  *
  * @property {boolean} isTMSSource - Used to checkout whether this source is a
- * TMSSource. Default is true. You should not change this, as it is used
+ * TMSSource. Default is tue. You should not change this, as it is used
```
ftoromanoff: typo

src/Core/Geographic/Crs.js
ftoromanoff, line 1: I saw that there is already a function called isValidCRS() why won't it be possible to add the check about being a string on it ?
And if it's really needed to check if the crs entered is a string maybe that test should be added to the source constructor instead of testing it on every sub function ?

jailln: Since all functions of `CRS.js` are static I think it's useful to add the test in each sub function to avoid errors if one calls `isEPSG()` without calling  `isValid()` before. 

Adding a `CRS.isValid` test to `Source` is a good idea (done) but I think we should also leave tests in each function of `CRS` since they may be called elsewhere in itowns code or by itowns users.

src/Source/TMSSource.js
jailln: ok

gchoqueux (approved):

PR #2110, opened 2023-06-08
Cloud Optimised Point Cloud (COPC) support

Desplandis:
## Description

This PR adds support of [COPC format](https://copc.io/):
- Add [copc.js](https://github.com/connormanning/copc.js) dependency for COPC parsing using [laz-perf](https://github.com/hobuinc/laz-perf) under the hood.
- Expose the following module to users of itowns: `CopcSource` and `CopcLayer` which extends existing `PointCloudLayer`.
- Add an example to load remote COPC file (with some examples from [copc.io](https://copc.io/#example-data)).
- Load the `laz-perf.wasm` by default through a CDN but this could be override by the user.

Note that this PR does not:
- use web worker for parsing (this will be the object of another PR).
- itowns/itowns#2157.
- fix issues in the point cloud engine (SSE computation, point budget, ...).

### Autzen Stadium
image
### SoFI Stadium
image
image

### DIFF
--- a/examples/copc_simple_loader.html
+++ b/examples/copc_simple_loader.html
@@ -1,0 +1,128 @@
+<html>
+    <head>
+        <meta charset="UTF-8">
+
+        <title>Itowns - COPC loader</title>
+
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <style>
+            #description {
+                z-index: 2;
+                left: 10px;
+            }
+        </style>
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="description">Specify the URL of a COPC file to load:
+            <input type="text" id="url" />
+            <button onclick="readURL()">Load</button>
+            <div>
+                <button onClick="loadAutzen()">Load Autzen Stadium (80mb)</button>
+                <button onClick="loadSofi()">Load SoFI Stadium (2.3gb)</button>
+                <button onClick="loadMillsite()">Load Millsite (1.9gb)</button>
+            <div id="share"></div>
+            </div>
+        </div>
+        <div id="viewerDiv"></div>
+
+        <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+        <script type="text/javascript">
+            let layer; // COPCLayer
+
+            const uri = new URL(location);
+
+            const gui = new dat.GUI();
+
+            const viewerDiv = document.getElementById('viewerDiv');
+            const view = new itowns.View('EPSG:4326', viewerDiv);
+            const controls = new itowns.PlanarControls(view);
+            view.mainLoop.gfxEngine.renderer.setClearColor(0xdddddd);
+
+            setUrl(uri.searchParams.get('copc'));
+
+            function onLayerReady(layer) {
+                const camera = view.camera.camera3D;
+
+                const lookAt = new itowns.THREE.Vector3();
+                const size = new itowns.THREE.Vector3();
+                layer.root.bbox.getSize(size);
+                layer.root.bbox.getCenter(lookAt);
+
+                camera.far = 2.0 * size.length();
+
+                controls.groundLevel = layer.root.bbox.min.z;
+                const position = layer.root.bbox.min.clone().add(
+                    size.multiply({ x: 1, y: 1, z: size.x / size.z }),
+                );
+
+                camera.position.copy(position);
+                camera.lookAt(lookAt);
+                camera.updateProjectionMatrix();
+
+                view.notifyChange(camera);
+            }
+
+
+            function readURL() {
+                const url = document.getElementById('url').value;
+
+                if (url) {
+                    setUrl(url);
+                }
+            }
+
+            function setUrl(url) {
+                if (!url) return;
+
+                const input_url = document.getElementById('url');
+                if (!input_url) return;
+
+                uri.searchParams.set('copc', url);
+                history.replaceState(null, null, `?${uri.searchParams.toString()}`);
+
+                input_url.value = url;
+                load(url);
+            }
+
+
+            function load(url) {
+                const source = new itowns.CopcSource({ url });
+
+                if (layer) {
+                    gui.removeFolder(layer.debugUI);
+                    view.removeLayer('COPC');
+                    view.notifyChange();
+                    layer.delete();
+                }
+
+                layer = new itowns.CopcLayer('COPC', {
+                    source,
+                    crs: view.referenceCrs,
+                    sseThreshold: 2,
+                    pointBudget: 3000000,
+                });
+                view.addLayer(layer).then(onLayerReady);
+                debug.PotreeDebug.initTools(view, layer, gui);
+            }
+
+            function loadAutzen() {
+                setUrl("https://s3.amazonaws.com/hobu-lidar/autzen-classified.copc.laz");
+            }
+
+            function loadSofi() {
+                setUrl("https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz");
+            }
+
+            function loadMillsite() {
+                setUrl("https://s3.amazonaws.com/data.entwine.io/millsite.copc.laz");
+            }
+        </script>
+    </body>
+</html>
--- a/src/Core/CopcNode.js
+++ b/src/Core/CopcNode.js
@@ -1,0 +1,194 @@
+import * as THREE from 'three';
+import { Hierarchy } from 'copc';
+import PointCloudNode from 'Core/PointCloudNode';
+
+const size = new THREE.Vector3();
+const position = new THREE.Vector3();
+const translation = new THREE.Vector3();
+
+function buildId(depth, x, y, z) {
+    return `${depth}-${x}-${y}-${z}`;
+}
+
+class CopcNode extends PointCloudNode {
+    /**
+     * Constructs a new instance of a COPC Octree node
+     *
+     * @param {number} depth - Depth within the octree
+     * @param {number} x - X position within the octree
+     * @param {number} y - Y position within the octree
+     * @param {number} z - Z position with the octree
+     * @param {number} entryOffset - Offset from the beginning of the file of
+     * the node entry
+     * @param {number} entryLength - Size of the node entry
+     * @param {CopcLayer} layer - Parent COPC layer
+     * @param {number} [numPoints=0] - Number of points given by this entry
+     */
+    constructor(depth, x, y, z, entryOffset, entryLength, layer, numPoints = 0) {
+        super(numPoints, layer);
+        this.isCopcNode = true;
+
+        this.entryOffset = entryOffset;
+        this.entryLength = entryLength;
+        this.layer = layer;
+        this.depth = depth;
+        this.x = x;
+        this.y = y;
+        this.z = z;
+    }
+
+    get id() {
+        return buildId(this.depth, this.x, this.y, this.z);
+    }
+
+    get octreeIsLoaded() {
+        return this.numPoints >= 0;
+    }
+
+    /**
+     * @param {number} offset
+     * @param {number} size
+     */
+    async _fetch(offset, size) {
+        return this.layer.source.fetcher(this.layer.source.url, {
+            ...this.layer.source.networkOptions,
+            headers: {
+                ...this.layer.source.networkOptions.headers,
+                range: `bytes=${offset}-${offset + size - 1}`,
+            },
+        });
+    }
+
+    /**
+     * Create an (A)xis (A)ligned (B)ounding (B)ox for the given node given
+     * `this` is its parent.
+     * @param {CopcNode} node - The child node
+     */
+    createChildAABB(node) {
+        // factor to apply, based on the depth difference (can be > 1)
+        const f = 2 ** (node.depth - this.depth);
+
+        // size of the child node bbox (Vector3), based on the size of the
+        // parent node, and divided by the factor
+        this.bbox.getSize(size).divideScalar(f);
+
+        // initialize the child node bbox at the location of the parent node bbox
+        node.bbox.min.copy(this.bbox.min);
+
+        // position of the parent node, if it was at the same depth as the
+        // child, found by multiplying the tree position by the factor
+        position.copy(this).multiplyScalar(f);
+
+        // difference in position between the two nodes, at child depth, and
+        // scale it using the size
+        translation.subVectors(node, position).multiply(size);
+
+        // apply the translation to the child node bbox
+        node.bbox.min.add(translation);
+
+        // use the size computed above to set the max
+        node.bbox.max.copy(node.bbox.min).add(size);
+    }
+
+    /**
+     * Create a CopcNode from the provided subtree and add it as child
+     * of the current node.
+     * @param {number} depth - Child node depth in the octree
+     * @param {number} x - Child node x position in the octree
+     * @param {number} y - Child node y position in the octree
+     * @param {number} z - Child node z position in the octree
+     * @param {Hierarchy.Subtree} hierarchy - Octree's subtree
+     * @param {CopcNode[]} stack - Stack of node candidates for traversal
+     */
+    findAndCreateChild(depth, x, y, z, hierarchy, stack) {
+        const id = buildId(depth, x, y, z);
+
+        let pointCount;
+        let offset;
+        let byteSize;
+
+        const node = hierarchy.nodes[id];
+        if (node) {
+            pointCount = node.pointCount;
+            offset = node.pointDataOffset;
+            byteSize = node.pointDataLength;
+        } else {
+            const page = hierarchy.pages[id];
+            if (!page) { return; }
+            pointCount = -1;
+            offset = page.pageOffset;
+            byteSize = page.pageLength;
+        }
+
+        const child = new CopcNode(
+            depth,
+            x,
+            y,
+            z,
+            offset,
+            byteSize,
+            this.layer,
+            pointCount,
+        );
+        this.add(child);
+        stack.push(child);
+    }
+
+    async loadOctree() {
+        // Load hierarchy
+        const buffer = await this._fetch(this.entryOffset, this.entryLength);
+        const hierarchy = await Hierarchy.parse(new Uint8Array(buffer));
+
+        // Update current node entry from loaded subtree
+        const node = hierarchy.nodes[this.id];
+        if (!node) {
+            return Promise.reject('[CopcNode]: Ill-formed data, entry not found in hierarchy.');
+        }
+        this.numPoints = node.pointCount;
+        this.entryOffset = node.pointDataOffset;
+        this.entryLength = node.pointDataLength;
+
+        // Load subtree entries
+        const stack = [];
+        stack.push(this);
+        while (stack.length) {
+            const node = stack.shift();
+            const depth = node.depth + 1;
+            const x = node.x * 2;
+            const y = node.y * 2;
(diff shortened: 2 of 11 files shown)


Desplandis:
This PR is still missing some unit tests but is ready for review.

examples/copc_simple_loader.html
```diff
@@ -0,0 +1,130 @@
...
+                    view.notifyChange();
+                    layer.delete();
+                }
+
+                layer = new itowns.CopcLayer('COPC', {
+                    source,
+                    crs: view.referenceCrs,
+                    sseThreshold: 2,
+                    pointBudget: 3000000,
+                    // supportsProgressiveDisplay: true,
```
jogarnier: Commented line to remove ?

src/Parser/LAZParser.js
```diff
@@ -0,0 +1,199 @@
...
+    for (let i = 0; i < view.pointCount; i++) {
+        // `getPosition` apply scale and offset transform to the X, Y, Z values.
+        // See https://github.com/connormanning/copc.js/blob/master/src/las/extractor.ts.
+        const [x, y, z] = getPosition.map(f => f(i));
+        positions[i * 3] = x;
+        positions[i * 3 + 1] = y;
+        positions[i * 3 + 2] = z;
+
+        const intensity = getIntensity(i);
+        intensities[i] = intensity;
```
mgermerie: Maybe you could avoid creating a `const` that is only used on line 34, and replace line 34 with 
```js
intensities[i] = getIntensity(i);
```
I don't know a lot about memory management but do you know if there is a difference in creating constants at each loop iteration ?

mgermerie: The same question goes for lines 36 to 39.

mgermerie: And 59 to 63 ;)

jogarnier:
I had several eslint errors while doing `npm start` on my local code, please have a look over the following : 

```
itowns\src\Core\Style.js

  4:1  error  `@mapbox/mapbox-gl-style-spec` import should occur before import of `Core/Feature`  import/order
  5:1  error  `three` import should occur before import of `Core/Feature`                         import/order

itowns\src\Parser\B3dmParser.js
  10:1  error  `three` import should occur before import of `Core/System/Capabilities`  import/order

itowns\utils\debug\Debug.js
  5:1  error  `three` import should occur before import of `Core/Geographic/Coordinates`  import/order


itowns\src\Source\CopcSource.js
  6:1  error  `three` import should occur before import of `Core/Geographic/Extent`  import/order
```

src/Source/CopcSource.js
```diff
@@ -0,0 +1,121 @@
+import { Binary, Info, Las } from 'copc';
+import Extent from 'Core/Geographic/Extent';
+import Fetcher from 'Provider/Fetcher';
+import LAZParser from 'Parser/LAZParser';
+import Source from 'Source/Source';
+import * as THREE from 'three';
```
jogarnier, line 6: I had several eslint errors while doing `npm start` on my local code, please have a look over the following : 

```
itowns\src\Core\Style.js

  4:1  error  `@mapbox/mapbox-gl-style-spec` import should occur before import of `Core/Feature`  import/order
  5:1  error  `three` import should occur before import of `Core/Feature`                         import/order

itowns\src\Parser\B3dmParser.js
  10:1  error  `three` import should occur before import of `Core/System/Capabilities`  import/order

itowns\utils\debug\Debug.js
  5:1  error  `three` import should occur before import of `Core/Geographic/Coordinates`  import/order


itowns\src\Source\CopcSource.js
  6:1  error  `three` import should occur before import of `Core/Geographic/Extent`  import/order
```

This a bit out of your scope but if you can handle this, I think it would be nice, 
it refers to this commit a8413433091882be7c461977ccdb343c75d89d27, where some npm packages have been updated (eslint / eslint-plugin-import / eslint-webpack-plugin which has taken one major version update)

It's a bit weird that the tests have passed, by the way

examples/copc_simple_loader.html
```diff
@@ -0,0 +1,130 @@
...
+                });
+                view.addLayer(layer).then(onLayerReady);
+                debug.PotreeDebug.initTools(view, layer, gui);
+            }
+
+            function loadAutzen() {
+                setUrl("https://s3.amazonaws.com/hobu-lidar/autzen-classified.copc.laz");
```
jogarnier, line 116: Can I suggest you to upload the files onto https://github.com/iTowns/iTowns2-sample-data, instead ?

jogarnier (approved):

examples/copc_simple_loader.html
Desplandis: It seems we can't HTTP range request on `raw.githubusercontent.com` (due to CORS preflight). It is an [issue](https://github.com/orgs/community/discussions/14214) on github side... This is why I use the official examples of the [copc.io](https://copc.io/#example-data) site.

examples/copc_simple_loader.html
Desplandis: I'll remove it

src/Parser/LAZParser.js
```diff
@@ -0,0 +1,199 @@
...
+        classification: classifications,
+        pointSourceID: pointSourceIDs,
+        color: colors,
+    };
+}
+
+function parseBuffer(pointData, options) {
+    const { header, eb } = options.in;
+    const view = copc.Las.View.create(pointData, header, eb);
+    const attrs = parseView(view);
```
mgermerie: It seems that this is the only call of `parseView` method. It does not pass an `options` parameter. Maybe you could add it so that the `options.colorDepth` specified by the user when instantiating its `CopcSource` is taken into account ?

src/Layer/CopcLayer.js
```diff
@@ -0,0 +1,62 @@
...
+        this.whenReady = this.source.whenReady.then((/** @type {CopcSource} */ source) => {
+            const { cube, rootHierarchyPage } = source.info;
+            const { pageOffset, pageLength } = rootHierarchyPage;
+
+            this.root = new CopcNode(0, 0, 0, 0, pageOffset, pageLength, this, -1);
+            this.root.bbox.min.fromArray(cube, 0);
+            this.root.bbox.max.fromArray(cube, 3);
+
+            const minElevationRange = source.header.min[2];
+            const maxElevationRange = source.header.max[2];
```
mgermerie: Why storing these in constants ?

src/Core/CopcNode.js
```diff
@@ -0,0 +1,190 @@
...
+     * @param {number} size
+     */
+    async _fetch(offset, size) {
+        return this.layer.source.fetcher(this.layer.source.url, {
+            ...this.layer.source.networkOptions,
+            headers: {
+                ...this.layer.source.networkOptions.headers,
+                range: `bytes=${offset}-${offset + size - 1}`,
+            },
+        });
```
mgermerie, line 59: It might be too tedious for the benefit but do you think this part could be factorized with the first promise of the `get` method in `CopcSource`

src/Core/CopcNode.js
```diff
@@ -0,0 +1,190 @@
...
+            const z = node.z * 2;
+
+            node.findAndCreateChild(depth, x,     y,     z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y,     z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y + 1, z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y + 1, z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y,     z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y,     z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y + 1, z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y + 1, z + 1, hierarchy, stack);
```
mgermerie, line 168: As we discussed, could you return the child in `findAndCreateChild` and push it into `stack` instead of passing `stack` as a parameter ?

src/Core/CopcNode.js
```diff
@@ -0,0 +1,190 @@
...
+
+        let pointCount;
+        let offset;
+        let byteSize;
+
+        const node = hierarchy.nodes[id];
+        if (node) {
+            pointCount = node.pointCount;
+            offset = node.pointDataOffset;
+            byteSize = node.pointDataLength;
```
mgermerie, line 114: I might have missed something but I figure that doing this results in all 8 nodes from depth 1 (assuming the root node is at depth 0) will have the same number of points, entry offset and size (with equal `numPoints`, `entryOffset` and `entryLength` properties values). These common values are those of the parent node (root in my example). Is this intended ?

src/Core/CopcNode.js
mgermerie: This mechanism is also present in `EntwinePointTileNode`.

src/Provider/PointCloudProvider.js
```diff
@@ -36,7 +36,12 @@ export default {
             addPickingAttribute(points);
             points.frustumCulled = false;
             points.matrixAutoUpdate = false;
-            if (!layer.isEntwinePointTileLayer) {
+            if (!layer.isEntwinePointTileLayer && !layer.isCopcLayer) {
```
mgermerie: If I understand well, this test would be removed should we just add `this.scale = new THREE.Vector3(1.0, 1.0, 1.0)` in `EntwinePointCloudLayer.whenReady` ?

src/Layer/CopcLayer.js
mgermerie: Also, I could not find an usage of `minElevationRange` and `maxElevationRange` properties, nor some documentation. What are they for ?

examples/copc_simple_loader.html
```diff
@@ -0,0 +1,130 @@
...
+            let layer; // COPCLayer
+
+            const uri = new URL(location);
+
+            const gui = new dat.GUI();
+
+            const viewerDiv = document.getElementById('viewerDiv');
+            const view = new itowns.View('EPSG:4326', viewerDiv);
+            const controls = new itowns.PlanarControls(view);
+            viewerDiv.style.display = 'block';
```
mgermerie: Why is this needed ?

mgermerie (commented):
Thanks for this PR ! 
I only have a few comments and two general remarks : 
- Could you add yourself to CONTRIBUTORS.md ?
- I looked at coverage reports, and strangely some parts are not covered even if you call them in your new unit tests (the constructor of `CopcLayer` for instance). Could you take a look at it ? and add tests if needed ?

Other than that it should be good for me !

src/Layer/CopcLayer.js
Desplandis: Leftovers of when this PR also added support of other LAS Attributes. I will remove them for now.

src/Parser/LAZParser.js
Desplandis: I suppose the javascript engine is smart enough to optimize such construct. However, they are leftovers of debugging, I will directly set those values.

src/Parser/LAZParser.js
Desplandis: Whoops... Forget to pass an the `options` parameter.
The `parseView` function is only called once because I may, in another PR, add a `parseFile` function to parse an entire LAS/LAZ file and replace entirely the `LASParser` file.

src/Provider/PointCloudProvider.js
Desplandis: You'll also need to check that `node.bbox.min` shall not create unwanted offsets...

src/Source/CopcSource.js
Desplandis: This is caused by recent versions of `eslint` and will be fixed in another PR.

src/Core/CopcNode.js
Desplandis: I agree. Should I modify `EntwinePointTileNode` too ?

src/Core/CopcNode.js
Desplandis: I used `this.layer.source.fetcher` directly since I supposed that a user could use a custom `fetcher` or `Source`. Maybe this is too far-fetched!

examples/copc_simple_loader.html
Desplandis: Removed it.

src/Provider/PointCloudProvider.js
Desplandis: I opened an issue on this topic. See #2189 !

src/Core/CopcNode.js
Desplandis: Fixed it!

Desplandis:
@mgermerie @jogarnier Since your last review, I moved part of `LAZParser.js` to  `LAZLoader.js`. This is done to split the parser in two parts: one that could be "workerisable" (i.e. `LAZLoader`) in a future PR (I already have some POC) and one that creates Geometry Buffers/instantiate workers.

I also took into account all your comments.

This is ready for a second (quick) review!

Desplandis:
@mgermerie @jogarnier I also opened an issue on the copc.js library (see connormanning/copc.js#6)

TLDR: `copc.js` lazy-load the fs node module which is not platform agnostic. This causes bundling issues (with at least webpack and vite) in a similar fashion as our issue #2200.

If this issue is not fixed upstream, we may have to "fork" their package to remove the dozen lines causing this problem...

connormanning:
I've left a comment on the referenced ticket, but regardless I don't think you'd need a fork to accomplish this.  You can accomplish this downstream of copc.js with a simple library whose code simply re-exports everything from copc.js, and whose bundler nullfies any reference to `fs`, for example with webpack's `null-loader`, or using a Rollup IIFE bundle that sets the `fs` global to `null` since the `fs` code-paths would not be executed in a web environment anyway.  You could also do this in the itowns config rather than a separate library.

I think the capability to work around this exists - if you find yourself needing to make a fork over this let me know and I can lend a hand.

examples/copc_simple_loader.html
```diff
@@ -0,0 +1,128 @@
...
+            <input type="text" id="url" />
+            <button onclick="readURL()">Load</button>
+            <div>
+                <button onClick="loadAutzen()">Load Autzen Stadium (80mb)</button>
+                <button onClick="loadSofi()">Load SoFI Stadium (2.3gb)</button>
+                <button onClick="loadMillsite()">Load Millsite (1.9gb)</button>
+            <div id="share"></div>
+            </div>
+        </div>
+        <div id="viewerDiv">
```
AnthonyGlt: missing closing div

examples/copc_simple_loader.html
```diff
@@ -0,0 +1,128 @@
+<html>
+    <head>
+        <meta charset="UTF-8">
+
+        <title>Itowns - COPC loader</title>
+
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <style type="text/css">
```
AnthonyGlt: type is deprecated/unnecessary https://developer.mozilla.org/en-US/docs/web/html/element/style#type

src/Core/CopcNode.js
```diff
@@ -0,0 +1,194 @@
...
+        const f = 2 ** (node.depth - this.depth);
+
+        // size of the child node bbox (Vector3), based on the size of the
+        // parent node, and divided by the factor
+        this.bbox.getSize(size).divideScalar(f);
+
+        // initialize the child node bbox at the location of the parent node bbox
+        node.bbox.min.copy(this.bbox.min);
+
+        // position of the parent node, if it was at the same depth than the
```
AnthonyGlt:
```suggestion
        // position of the parent node, if it was at the same depth as the
```

src/Parser/LASParser.js
```diff
@@ -22,6 +51,39 @@ export default {
...
+
+
+    /**
+     * Parses a chunk of a LAS or LAZ (LASZip) and returns the corresponding
+     * `THREE.BufferGeometry`.
+     *
+     * @param {ArrayBuffer} data - The file content to parse.
+     * @param {Object} options
+     * @param {Object} options.in - Options to give to the parser.
+     * @param {number} option.in.pointCount - Number of points encoded in this
```
AnthonyGlt:
```suggestion
     * @param {number} options.in.pointCount - Number of points encoded in this
```

src/Provider/PointCloudProvider.js
```diff
@@ -36,7 +36,12 @@ export default {
...
+                // TODO: Harmonize properties of all PointCloudLayers
+                // Since both EPT and COPC Layer apply scaling at parsing, they
+                // should set a `layer.scale` at `THREE.Vector(1.0, 1.0, 1.0)`.
+                // For the `offset`, see if possible to move this property
+                // upward in Potree layer.
```
AnthonyGlt: maybe add the link to the issue https://github.com/iTowns/itowns/issues/2189 , easier to come back to it

src/Source/CopcSource.js
```diff
@@ -0,0 +1,121 @@
...
+        [];
+
+    return { header, info, wkt, eb };
+}
+
+/**
+ * @classdesc
+ * A source for [Cloud Optimised Point Cloud](https://copc.io/) (COPC) data.
+ * Such data consists of a [LAZ 1.4](https://www.ogc.org/standard/las/) file
+ * that stores compressed points data organized in a clustered octre.
```
AnthonyGlt:
```suggestion
 * that stores compressed points data organized in a clustered octree.
```

src/Source/CopcSource.js
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @extends {Source}
+ *
+ * @property {boolean} isCopcSource - Read-only flag to check that a given
+ * object is of type CopcSource.
+ * @property {Object} header - LAS header of the source.
+ * @property {Object[]} eb - List of headers of each Variable Length Records
+ * (VLRs).
+ * @property {Object} info - COPC `info` VLR.
+ * @property {number[]} info.cube - Bouding box of the octree as a 6-elements.
```
AnthonyGlt:
```suggestion
 * @property {number[]} info.cube - Bounding box of the octree as a 6-elements.
```

src/Source/CopcSource.js
```diff
@@ -0,0 +1,121 @@
...
+ * root node data chunk.
+ * @property {number} info.rootHierarchyPage.pageOffset - Size (in bytes) of the
+ * root node data chunk.
+ * @property {number[]} gpsTimeRange - A 2-element tuple denoting the minimum
+ * and maximum values of attribute `gpsTime`.
+ */
+class CopcSource extends Source {
+    /**
+     * @param {Object} config - Source configuration
+     * @param {string} config.url - URL of the COPC ressource.
```
AnthonyGlt:
```suggestion
     * @param {string} config.url - URL of the COPC resource.
```
Little mistake made multiple times in the comments

src/Source/CopcSource.js
```diff
@@ -0,0 +1,121 @@
...
+                headers: {
+                    ...this.networkOptions.headers,
+                    range: `bytes=${begin}-${end - 1}`,
+                },
+            }).then(buffer => new Uint8Array(buffer));
+        this.whenReady = getHeaders(get).then((metadata) => {
+            this.header = metadata.header;
+            this.info = metadata.info;
+            this.eb = metadata.eb;
+            // TODO: use wkt definition in `metadata.wkt` to infer/define crs
```
AnthonyGlt, line 108: Is it planned (issue/proposal/PR link) ?

LoicUTC:
That would be a great feature :)

src/Provider/PointCloudProvider.js
Desplandis: Fixed by #2284

src/Source/CopcSource.js
Desplandis: Not for now, we let the user set the `crs` manually. We would need to have `proj4` expose its `wkt`-parsing ability.

examples/copc_simple_loader.html
Desplandis: Done

mgermerie (approved):
:tada:

Issue #2099, opened 2023-05-22
[Example] Google Photorealistic 3D Tiles


jailln:
Add an example displaying Google Photorealistic 3D Tiles:

https://cloud.google.com/blog/products/maps-platform/create-immersive-3d-map-experiences-photorealistic-3d-tiles?hl=en
https://developers.google.com/maps/documentation/tile/create-renderer?hl=en

AnthonyGlt:
I've tried to display these tiles in iTowns, I had to do some workaround but the result is nice and probably worth the integration:
![Screenshot from 2023-05-31 09-53-03](https://github.com/iTowns/itowns/assets/126568810/0c643302-9b92-4b76-9336-d8c7ac28522e)
We need to;
- Implement a new `C3DTilesSource` (like what was done with `C3DTilesIonSource`). It's necessary to request the Google tiles by giving the API key and sessionId for each call.
- Implement a `GLTFParser` and a `C3DTGLTFExtension` to handle `3DTILES_content_gltf`
- Don't forget to `enableDracoLoader`

I did a quick test of the implementation but my work is not good enough for a PR, the tiles are not correclty loaded, some faulty behaviors. But I believe that it could be implementable without too much effort.
Faulty loading of tiles:
![Screenshot from 2023-05-31 10-16-13](https://github.com/iTowns/itowns/assets/126568810/ed9c4cf8-d158-4295-83ef-2a59385e3ba6)
![Screenshot from 2023-05-31 10-15-54](https://github.com/iTowns/itowns/assets/126568810/69ba0282-e70e-4c0e-b53d-495c85aaf00e)

jailln:
Cool thanks! Regarding "faulty loading of tiles", it looks like some tiles may be under the ground, we can try to offset each tile a little bit to verify if it's not the cause of the wrong display

AnthonyGlt:
You were right, by hiding the globe layer, we can see that the tiles are loaded . Still unstable though.


https://github.com/iTowns/itowns/assets/126568810/07281757-757e-474e-950f-7ee141a8562d

jailln:
that may come from the bounding boxes that are considered axis aligned while they should not. I'm currently working on that

alexLuky:
@jailln 
perhaps fixing this issue you'll find the reason why Cesium renders sphere boxes more stable than itowns does.
[...]

PR #2124, opened 2023-06-30
Feat/google3d tiles

jogarnier:
## Description
* added C3DTilesGoogleSource type to C3DTilesSource.
* C3DTilesGoogleSource requires non-standard parsing metadata to provides valid urls => identified this code clearly
* Refactored b3dmParser to extract pure gltfparser logic
* added url parameters to Fetcher methods to allow passing dynamic parameters to url request, (required by C3DTilesGoogleSource authentification logic. https://developers.google.com/maps/documentation/tile/create-renderer?hl=en)

## Motivation and Context
https://github.com/iTowns/itowns/issues/2099


docs/tutorials/Fundamentals.md
```diff
@@ -68,7 +68,7 @@ Several specific types of `Layers` exist, the use of which depends on the data t
 - `{@link PointCloudLayer}` can be used to display 3D point clouds. Any point cloud formats are supported as long as the corresponding `Source` is provided.
 Some point clouds formats such as Potree, Las and Entwine already have parsers defined in itowns that you can use. For 3D Tiles point clouds (pnts), use
 `C3DTilesLayer`.
-- `{@link C3DTilesLayer}` can be used to display 3D Tiles layer (only b3dm and pnts).
+- `{@link C3DTilesLayer}` can be used to display 3D Tiles layer (only b3dm and pnts and Google Photorealistic 3D Tiles).
```
jailln: I would say: 
``{@link C3DTilesLayer}` can be used to display 3D Tiles datasets in version 1.0 (b3dm, pnts and gltf tiles are supported).`

and add `C3DTilesGoogleSource` in the description of sources (line 50)

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+import utf8Decoder from 'Utils/Utf8Decoder';
+import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
+import { MeshBasicMaterial } from 'three';
+import disposeThreeMaterial from 'Utils/ThreeUtils';
+
+const matrixChangeUpVectorZtoY = (new THREE.Matrix4()).makeRotationX(Math.PI / 2);
+
+export const glTFLoader = new GLTFLoader();
+const dracoLoader = new DRACOLoader();
+dracoLoader.setDecoderPath('./libs/draco/');
```
jailln: Draco loader should not be set here but by itowns user (with `enableDracoLoader` like in the 3D Tiles examples) since draco is not provided by itowns and since it may not be in `./libs` for the end user.

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+            const semantic = obj.gltfShader.boundUniforms[name].semantic;
+            if (!supported.includes(semantic)) {
+                delete obj.gltfShader.boundUniforms[name];
+            }
+        }
+    }
+}
+
+/**
+ * @module GLTFParser
```
jailln, line 13: Can you add a description please? Important elements are:
* it supports GLTF 1.0 and 2.0 loading
* GLTF 2.0 loading is done with THREE.GltfLoader()
* it filters out non standard cesium-specific semantic that may have been added to gltf embedded in 3D Tiles

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+    const dracoLoader = new DRACOLoader();
+    dracoLoader.setDecoderPath(path);
+    if (config) {
+        dracoLoader.setDecoderConfig(config);
+    }
+    glTFLoader.setDRACOLoader(dracoLoader);
+}
+
+export default {
+    /** Parse gltf buffer and extract THREE.Scene and batch table
```
jailln: Remove the batch table part from the comment since it is specific to b3dm

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+import { GLTFLoader } from 'ThreeExtended/loaders/GLTFLoader';
+import { DRACOLoader } from 'ThreeExtended/loaders/DRACOLoader';
+import LegacyGLTFLoader from 'Parser/deprecated/LegacyGLTFLoader';
+import shaderUtils from 'Renderer/Shader/ShaderUtils';
+import utf8Decoder from 'Utils/Utf8Decoder';
+import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
+import { MeshBasicMaterial } from 'three';
+import disposeThreeMaterial from 'Utils/ThreeUtils';
+
+const matrixChangeUpVectorZtoY = (new THREE.Matrix4()).makeRotationX(Math.PI / 2);
```
jailln: It's the opposite: Y -> Z (gltf is y-up and itowns is z-up)

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+        dracoLoader.setDecoderConfig(config);
+    }
+    glTFLoader.setDRACOLoader(dracoLoader);
+}
+
+export default {
+    /** Parse gltf buffer and extract THREE.Scene and batch table
+     * @param {ArrayBuffer} buffer - the gltf buffer.
+     * @param {Object} options - additional properties.
+     * @param {Matrix4=} [options.gltfUpAxisMatrix={}] - Matrix4f up axis transformation, by default Z->Y is applied
```
jailln: opposite here also (y->z)

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+     * @param {string} options.urlBase - the base url of the glTF file (used to fetch textures for the embedded glTF model).
+     * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patching material with logarithmic depth buffer support.
+     * @param {float} [options.opacity=1.0] - the b3dm opacity. // unused here for now
```
jailln: Can you remove it then? The user can do it after the parsing if needed

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+        const magicNumberByteLength = 4;
+        gltfHeader.magic = utf8Decoder.decode(new Uint8Array(buffer, 0, magicNumberByteLength));
+        if (gltfHeader.magic) {
+            const promises = [];
+            promises.push(new Promise((resolve/* , reject */) => {
+                const onload = (gltf) => {
+                    for (const scene of gltf.scenes) {
+                        scene.traverse(filterUnsupportedSemantics);
+                    }
+                    // for gltf Z->Y is applied
```
jailln: opposite here too

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+                    if (!gltfUpAxisMatrix) {
+                        gltf.scene.applyMatrix4(matrixChangeUpVectorZtoY);
+                    } else {
+                        gltf.scene.applyMatrix4(gltfUpAxisMatrix);
+                    }
+
+                    const init_mesh = function f_init(mesh) {
+                        mesh.frustumCulled = false;
```
jailln: This should be an option given to the `GLTFParser`. It was always done in the case of gltf embedded in b3dm since frustum culling is done at the tile level (with bounding volumes) for 3D Tiles. However, users loading a GLTF file may want to leave the frustum culling activated

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+                            } else if (Capabilities.isLogDepthBufferSupported()
+                                        && mesh.material.isRawShaderMaterial
+                                        && !options.doNotPatchMaterial) {
+                                shaderUtils.patchMaterialForLogDepthSupport(mesh.material);
+                                console.warn('glTF shader has been patched to add log depth buffer support');
+                            }
+                            ReferLayerProperties(mesh.material, options.layer);
+                        }
+                    };
+                    gltf.scene.traverse(init_mesh);
```
jailln: I think we should avoid traversing gltf scene when we know it's not needed beforehand:
```
var shouldBePatchedForLogDepthSupport = Capabilities.isLogDepthBufferSupported() && !options.doNotPatchMaterial;

if (options.frustumCulling === false || options.overrideMaterials || shouldBePatchedForLogDepthSupport || options.layer) {
     gltf.scene.traverse(init_mesh);
}
```

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+
+        const gltfHeader = {};
+
+        // Magic type is unsigned char [4]
+        const magicNumberByteLength = 4;
+        gltfHeader.magic = utf8Decoder.decode(new Uint8Array(buffer, 0, magicNumberByteLength));
+        if (gltfHeader.magic) {
+            const promises = [];
+            promises.push(new Promise((resolve/* , reject */) => {
+                const onload = (gltf) => {
```
jailln: You can get the onLoad method out of here and just pass it as an argument

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+                        scene.traverse(filterUnsupportedSemantics);
+                    }
+                    // for gltf Z->Y is applied
+                    if (!gltfUpAxisMatrix) {
+                        gltf.scene.applyMatrix4(matrixChangeUpVectorZtoY);
+                    } else {
+                        gltf.scene.applyMatrix4(gltfUpAxisMatrix);
+                    }
+
+                    const init_mesh = function f_init(mesh) {
```
jailln: You can also take this method out

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,158 @@
...
+            throw new Error('No array buffer provided.');
+        }
+
+        const gltfHeader = {};
+
+        // Magic type is unsigned char [4]
+        const magicNumberByteLength = 4;
+        gltfHeader.magic = utf8Decoder.decode(new Uint8Array(buffer, 0, magicNumberByteLength));
+        if (gltfHeader.magic) {
+            const promises = [];
```
jailln: You don't need an array of promises since there is only one

src/Process/3dTilesProcessing.js
```diff
@@ -18,6 +18,11 @@ function requestNewTile(view, scheduler, geometryLayer, metadata, parent, redraw
         redraw,
     };
 
+    if (geometryLayer.source.isC3DTilesGoogleSource) {
+        // special case to handle non-standard 3dTilesGoogle behavior
```
jailln: can you link the issue on the 3D Tiles repo about this special behaviour please?

src/Provider/3dTilesProvider.js
```diff
@@ -25,6 +26,20 @@ function b3dmToMesh(data, layer, url) {
     });
 }
 
+function gltfoMesh(data, layer, url) {
```
jailln: gltfToMesh (T missing)

src/Source/C3DTilesGoogleSource.js
```diff
@@ -0,0 +1,59 @@
+import Fetcher from 'Provider/Fetcher';
+import C3DTilesSource from './C3DTilesSource';
+
+/**
+ * @classdesc
+ * An object defining the source connection to a 3DTiles asset of a [Google api](https://tile.googleapis.com).
```
jailln: `of a [Google..` -> `from a [Google..` ?

src/Source/C3DTilesGoogleSource.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @extends Source
```
jailln: extends `C3DTilesSource`

docs/tutorials/Fundamentals.md
jogarnier: done

src/Parser/GLTFParser.js
jogarnier: removed

src/Parser/GLTFParser.js
jogarnier: done

src/Parser/GLTFParser.js
jogarnier: ok

src/Parser/GLTFParser.js
jogarnier: done

src/Parser/GLTFParser.js
jogarnier: ok

src/Process/3dTilesProcessing.js
jogarnier: done

src/Provider/3dTilesProvider.js
jogarnier: done

src/Source/C3DTilesGoogleSource.js
jogarnier: ok

src/Source/C3DTilesGoogleSource.js
jogarnier: done

src/Parser/GLTFParser.js
jogarnier: ok

src/Parser/GLTFParser.js
jogarnier: added it to options parameters

src/Parser/GLTFParser.js
jogarnier: ok

src/Parser/GLTFParser.js
jogarnier: yes

src/Parser/GLTFParser.js
jogarnier: ok

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,161 @@
...
+}
+
+export default {
+    /** Parse gltf buffer and extract THREE.Scene
+     * @param {ArrayBuffer} buffer - the gltf buffer.
+     * @param {Object} options - additional properties.
+     * @param {Matrix4=} [options.gltfUpAxisMatrix={}] - Matrix4f up axis transformation, by default Y->Z is applied
+     * @param {string} options.urlBase - the base url of the glTF file (used to fetch textures for the embedded glTF model).
+     * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patching material with logarithmic depth buffer support.
+     * @param {boolean=} [options.frustumCulled='false'] - enable frustum culling.
```
jailln: Should be defaulted to true since it is the default behaviour in threejs and in general (when not in the case of 3D Tiles)

src/Parser/B3dmParser.js
```diff
@@ -104,6 +26,7 @@ export default {
      * @param {string} options.urlBase - the base url of the b3dm file (used to fetch textures for the embedded glTF model).
      * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patching material with logarithmic depth buffer support.
      * @param {float} [options.opacity=1.0] - the b3dm opacity.
+     * @param {boolean=} [options.frustumCulled='false'] - enable frustum culling.
```
Desplandis:
```suggestion
     * @param {boolean=} [options.frustumCulled=false] - enable frustum culling.
```

src/Parser/B3dmParser.js
```diff
@@ -19,82 +14,9 @@ export const glTFLoader = new GLTFLoader();
 
 export const legacyGLTFLoader = new LegacyGLTFLoader();
```
Desplandis: Since you moved both `enableDracoLoader` and `enableKtx2Loader` to `GLTFParser`, you should remove both `glTFLoader` and `legacyGLTFLoader` as they are not used elsewhere and not exported to the end-user.

N.B. This comment holds for lines 13-16.

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+
+        if (!buffer) {
+            throw new Error('No array buffer provided.');
+        }
+
+        const gltfHeader = {};
+
+        // Magic type is unsigned char [4]
+        const magicNumberByteLength = 4;
+        gltfHeader.magic = utf8Decoder.decode(new Uint8Array(buffer, 0, magicNumberByteLength));
```
Desplandis: Either set it as a field directly in `gltfHeader` (along with property `version` defined below) or remove this record.

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+        if (gltfHeader.magic) {
```
Desplandis: You should compare `gltfHeader.magic` with the string `'gltf'` (or a constant defined at the module top-level).

Why not doing an early return if the condition does not hold? This would eliminate a scope level.

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+                const onload = (gltf) => {
+                    for (const scene of gltf.scenes) {
+                        scene.traverse(filterUnsupportedSemantics);
+                    }
+                    // for gltf Y->Z is applied
+                    if (!gltfUpAxisMatrix) {
+                        gltf.scene.applyMatrix4(matrixChangeUpVectorYtoZ);
+                    } else {
+                        gltf.scene.applyMatrix4(gltfUpAxisMatrix);
+                    }
```
Desplandis: The `gltfUpAxis` field has been removed before the 1.0 version release of the 3D Tiles spec. Since `3DTILES_content_gltf` is an extension of 3D tiles 1.0 (and mandatory for 3D tiles 1.1) and as such `gltfUpAxis` is only found on `.b3dm` pre-1.0 datasets, we could move the change of basis in the corresponding `B3DMParser` (after the gltf parsing).

However, this comment is valid if this condition holds for all pre-1.0 datasets. Asking @jailln and @mgermerie if they know of some pre-1.0 datasets using `3DTILES_content_gltf`?

src/Parser/GLTFParser.js
Desplandis: I agree with the previous comment

src/Provider/3dTilesProvider.js
```diff
@@ -25,6 +26,20 @@ function b3dmToMesh(data, layer, url) {
     });
 }
 
+function gltfToMesh(data, layer, url) {
+    const urlBase = THREE.LoaderUtils.extractUrlBase(url);
+    const options = {
+        gltfUpAxis: layer.tileset.asset.gltfUpAxis,
```
Desplandis: `gltfUpAxis` is not a valid option of `GLTFParser`.

See previous comment on the validity of this option in the context of datasets using the `3DTILES_content_gltf` extension.

src/Provider/3dTilesProvider.js
```diff
@@ -25,6 +26,20 @@ function b3dmToMesh(data, layer, url) {
...
+        urlBase,
+        overrideMaterials: layer.overrideMaterials,
+        doNotPatchMaterial: layer.doNotPatchMaterial,
+        opacity: layer.opacity,
```
Desplandis: Does not seem to be used.

src/Provider/3dTilesProvider.js
```diff
@@ -111,6 +128,8 @@ function executeCommand(command) {
                     func = supportedFormats.b3dm;
                 } else if (magic == 'pnts') {
                     func = supportedFormats.pnts;
+                } else if (magic == 'glTF') {
+                    func =  supportedFormats.gltf;
```
Desplandis:
```suggestion
                    func = supportedFormats.gltf;
```

src/Parser/B3dmParser.js
```diff
@@ -185,64 +109,26 @@ export default {
...
-                if (version === 1) {
-                    legacyGLTFLoader.parse(gltfBuffer, urlBase, onload);
-                } else {
-                    glTFLoader.parse(gltfBuffer, urlBase, onload);
+            // Rotation managed
+            if (gltfUpAxis === undefined || gltfUpAxis === 'Y') {
+                options.gltfUpAxisMatrix = matrixChangeUpVectorZtoY;
+            } else if (gltfUpAxis === 'X') {
+                options.gltfUpAxisMatrix = matrixChangeUpVectorZtoX;
+            }
```
Desplandis: This create a side-effect on the argument `options`, adding a `gltfUpAxisMatrix` property. Prefer a local variable.

src/Parser/B3dmParser.js
```diff
@@ -104,6 +26,7 @@ export default {
      * @param {string} options.urlBase - the base url of the b3dm file (used to fetch textures for the embedded glTF model).
      * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patching material with logarithmic depth buffer support.
      * @param {float} [options.opacity=1.0] - the b3dm opacity.
+     * @param {boolean=} [options.frustumCulled='false'] - enable frustum culling.
      * @param {boolean|Material=} [options.overrideMaterials='false'] - override b3dm's embedded glTF materials. If
```
Desplandis:
```suggestion
     * @param {boolean|Material=} [options.overrideMaterials=false] - override b3dm's embedded glTF materials. If
```

src/Parser/B3dmParser.js
```diff
@@ -104,6 +26,7 @@ export default {
      * @param {string} options.urlBase - the base url of the b3dm file (used to fetch textures for the embedded glTF model).
      * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patching material with logarithmic depth buffer support.
```
Desplandis:
```suggestion
     * @param {boolean=} [options.doNotPatchMaterial=false] - disable patching material with logarithmic depth buffer support.
```

src/Parser/B3dmParser.js
```diff
@@ -1,15 +1,10 @@
...
-import { DRACOLoader } from 'ThreeExtended/loaders/DRACOLoader';
-import { KTX2Loader } from 'ThreeExtended/loaders/KTX2Loader';
 import LegacyGLTFLoader from 'Parser/deprecated/LegacyGLTFLoader';
-import shaderUtils from 'Renderer/Shader/ShaderUtils';
 import utf8Decoder from 'Utils/Utf8Decoder';
 import C3DTBatchTable from 'Core/3DTiles/C3DTBatchTable';
-import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
-import { MeshBasicMaterial } from 'three';
-import disposeThreeMaterial from 'Utils/ThreeUtils';
+
```
Desplandis:
```suggestion
```

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+    glTFLoader.setKTX2Loader(ktx2Loader);
+}
+
+export default {
+    /** Parse gltf buffer and extract THREE.Scene
+     * @param {ArrayBuffer} buffer - the gltf buffer.
+     * @param {Object} options - additional properties.
+     * @param {Matrix4=} [options.gltfUpAxisMatrix={}] - Matrix4f up axis transformation, by default Y->Z is applied
+     * @param {string} options.urlBase - the base url of the glTF file (used to fetch textures for the embedded glTF model).
+     * @param {boolean=} [options.doNotPatchMaterial='false'] - disable patching material with logarithmic depth buffer support.
```
Desplandis:
```suggestion
     * @param {boolean=} [options.doNotPatchMaterial=false] - disable patching material with logarithmic depth buffer support.
```

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+     * @param {boolean=} [options.frustumCulled='true'] - enable frustum culling.
```
Desplandis:
```suggestion
     * @param {boolean=} [options.frustumCulled=true] - enable frustum culling.
```

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+     * @param {boolean|Material=} [options.overrideMaterials='false'] - override b3dm's embedded glTF materials. If
```
Desplandis:
```suggestion
     * @param {boolean|Material=} [options.overrideMaterials=false] - override b3dm's embedded glTF materials. If
```

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+     * true, a threejs [MeshBasicMaterial](https://threejs.org/docs/index.html?q=meshbasic#api/en/materials/MeshBasicMaterial)
+     * is set up. config.overrideMaterials can also be a threejs [Material](https://threejs.org/docs/index.html?q=material#api/en/materials/Material)
+     * in which case it will be the material used to override.
```
Desplandis: There is still some comments referencing b3dm left. Could you please update the documentation?

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+                    const shouldBePatchedForLogDepthSupport = Capabilities.isLogDepthBufferSupported() && !options.doNotPatchMaterial;
+                    if (options.frustumCulling === false || options.overrideMaterials || shouldBePatchedForLogDepthSupport || options.layer) {
+                        gltf.scene.traverse(init_mesh);
+                    }
+
+                    resolve(gltf);
+                };
+                const headerView = new DataView(buffer, 0, 20);
+                const version = headerView.getUint32(4, true);
+
```
Desplandis: Maybe move this in the `gltfHeader` object above?

src/Parser/GLTFParser.js
```diff
@@ -0,0 +1,185 @@
...
+            names.push(name);
+        }
+        for (const name of names) {
+            const semantic = obj.gltfShader.boundUniforms[name].semantic;
+            if (!supported.includes(semantic)) {
+                delete obj.gltfShader.boundUniforms[name];
+            }
+        }
+    }
+}
```
Desplandis: Both `GLTFLShader` and property `gltfShader` does not seem to exist anymore in `three.js`.
This comment is out-of-scope of this PR (I'll validate if this is not fixed) but this code may be useless and removed in another PR (tagging @jailln).

src/Process/3dTilesProcessing.js
```diff
@@ -18,6 +18,12 @@ function requestNewTile(view, scheduler, geometryLayer, metadata, parent, redraw
...
+        // see also https://github.com/CesiumGS/3d-tiles/issues/746
+        geometryLayer.source.completeMetadata(metadata);
+    }
```
Desplandis: @mgermerie @jailln I know this behavior is non-standard... However, Cesium is the "de facto" 3D Tiles standard. Shall we stick to the standard (adding this kind of duck taping for non-conforming datasets) or follow Cesium's behavior?

src/Provider/Fetcher.js
```diff
@@ -10,10 +10,21 @@ function checkResponse(response) {
...
+}
+
+function formatUrlWithParameters(url, urlParameters = {}) {
+    if (Object.keys(urlParameters).length === 0) {
+        return url;
+    } else {
+        const extraParameters = new URLSearchParams(urlParameters).toString();
+        return /\?/.test(url) ? `${url}&${extraParameters}` : `${url}?${extraParameters}`;
+    }
+}
```
Desplandis: Is there any other use-cases (except the google one) where the use of `urlParameters` could be necessary.
Is there is none, we could otherwise move those changes in `C3DTilesGoogleSource` and `GLTFParser`.

src/Provider/Fetcher.js
```diff
@@ -10,10 +10,21 @@ function checkResponse(response) {
...
-    return response.arrayBuffer();
-});
+function arrayBuffer(url, options = {}, urlParameters = {}) {
+    return fetch(formatUrlWithParameters(url, urlParameters), options).then((response) => {
+        checkResponse(response);
+        return response.arrayBuffer();
+    });
+}
+
+function formatUrlWithParameters(url, urlParameters = {}) {
```
Desplandis: I think `urlParameters` shall be of type `object | undefined`. The check below should be faster.

src/Source/C3DTilesGoogleSource.js
```diff
@@ -0,0 +1,59 @@
+import Fetcher from 'Provider/Fetcher';
+import C3DTilesSource from './C3DTilesSource';
+
+/**
+ * @classdesc
+ * An object defining the source connection to a 3DTiles asset from a [Google api](https://tile.googleapis.com).
```
Desplandis, line 24: Since google 3D tiles support is "experimental", maybe we should precise it here?

src/Source/C3DTilesGoogleSource.js
```diff
@@ -0,0 +1,59 @@
...
+                    this.urlParameters.session = sessionIds[1] !== undefined ? sessionIds[1] : '';
+                }
+            } else {
+                throw new Error('C3DTilesGoogleSource: uri is required');
+            }
+        }
+        metadata.baseURL = this.baseUrl;
+        return metadata;
+    }
+}
```
Desplandis: This function (called by the provider) creates a side-effect on the source.
We could complete the metadata on the provider if we build the url in it (i.e. no change to `Fetcher.js`). Depends on the implementation choice here.

Desplandis (changes requested):
Nice job on this PR! =)
I was able to test it using my google cloud keys and your example branch. This is still "experimental" as they are critical issues (blinking, no loading of some tiles, ...) but these come from our 3D tiles implementation.

As for my comments:
- I think you should move some parts specific to b3dm outside of the gltf parser
- I question if we should follow the standard or cesium behavior
- I'm not convinced by the multiple changes in `Fetcher.js` due to this use-case being specific but I could change my mind with a little explanation =) 
- Some minor issues

P.S.: Before merging, could you please rewrite your commit messages + your history (notably isolate changes in `Fetcher.js`, changes in b3dm/gltf parsers in their own commits)?

src/Parser/GLTFParser.js
jailln: We would still need to change from y-up to z-up here when `gltfUpAxis` is not defined. Or do you mean we could always change from y to z-up here and in the `B3dmParser`, if `gltfUpAxis`, change back from z-up to y-up and then apply the `gltfUpAxisMatrix`?

> However, this comment is valid if this condition holds for all pre-1.0 datasets. Asking @jailln and @mgermerie if they know of some pre-1.0 datasets using 3DTILES_content_gltf?

From what I know, there was no 3D tiles extensions before 3D Tiles 1.0.

jailln:
I just pushed some modifications that should answer most of your comments:
  * Make GLTFParser a wrapper for GltfParser and LeacyGltfParser and move everything 3D tiles related from the GLTFParser to the B3DMParser
* Only export `GLTFParser` and not `GltfLoader` and `LegacyGltfLoader` anymore (BREAKING CHANGE) -> should we keep that or still also export the  `GltfLoader` and `LegacyGltfLoader` until the next major version ? WDYT?

src/Parser/GLTFParser.js
jailln: I removed this test on `magic` since it is only present for binary gltf [according to the spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#binary-header)

src/Parser/GLTFParser.js
jailln: I moved it to the `B3dmParser`; I haven't found any 3D tiles dataset with this property to test it though.

> However, this comment is valid if this condition holds for all pre-1.0 datasets. Asking @jailln and @mgermerie if they know of some pre-1.0 datasets using 3DTILES_content_gltf?

No I don't and I think 3D tiles extensions were added after 3D tiles 1.0.

jailln:
I just pushed some modifications that should answer all of your comments:

* `GLTFParser` is now only a wrapper for `gltfLoader` and `legacyGltfLoader` and everything 3D tiles related as been moved to `B3DMParser`
* Only export `GLTFParser` and not `gltfLoader` and `legacyGltfLoader` anymore, which is a breaking change but avoids exporting two things having the same purpose. WDYT ? Should we keep it like that or also export `gltfLoader` and `legacyGltfLoader` to avoid a breaking change until the next major version?
* Some 3D Tiles related parts could be refactored / simplified (especially the `B3dmParser`) but I didn't go into much trouble here since we are seriously considering to handle 3D tiles parsing with an external lib very soon.
* I wanted to add more tests on the `GLTFParser` but I decided to wait until #2183 is merged since I needed to load local files.

@Desplandis let me know what you think :)

src/Parser/B3dmParser.js
```diff
@@ -43,77 +40,45 @@ function filterUnsupportedSemantics(obj) {
...
      * true, a threejs [MeshBasicMaterial](https://threejs.org/docs/index.html?q=meshbasic#api/en/materials/MeshBasicMaterial)
      * is set up. config.overrideMaterials can also be a threejs [Material](https://threejs.org/docs/index.html?q=material#api/en/materials/Material)
      * in which case it will be the material used to override.
      * @return {Promise} - a promise that resolves with an object containig a THREE.Scene (gltf) and a batch table (batchTable).
      *
      */
     parse(buffer, options) {
-        const gltfUpAxis = options.gltfUpAxis;
-        const urlBase = options.urlBase;
+        const frustumCulled = options.frustumCulled === undefined || options.frustumCulled === null ? true : !!(options.frustumCulled);
```
Desplandis: A simpler way could be:
```suggestion
        const frustumCulled = options.frustumCulled ?? true;
```

Moreover, the default parameter is documented as `false`.

src/Provider/3dTilesProvider.js
```diff
@@ -91,23 +96,36 @@ function executeCommand(command) {
...
+            gltf: gltfToMesh,
         };
         return Fetcher.arrayBuffer(url, layer.source.networkOptions).then((result) => {
             if (result !== undefined) {
                 let func;
                 const magic = utf8Decoder.decode(new Uint8Array(result, 0, 4));
                 if (magic[0] === '{') {
                     result = JSON.parse(utf8Decoder.decode(new Uint8Array(result)));
-                    const newPrefix = url.slice(0, url.lastIndexOf('/') + 1);
+                    let newPrefix = '';
```
Desplandis: No need to assign `newPrefix` since it is assigned in both branch of the following condition (or do a ternary condition).
```suggestion
                    let newPrefix;
```

src/Source/C3DTilesGoogleSource.js
```diff
@@ -0,0 +1,79 @@
...
+import C3DTilesSource from './C3DTilesSource';
+
+function findSessionId(tile) {
+    if (!tile) {
+        return null;
+    }
+    if (tile.content && tile.content.uri) {
+        const uri = tile.content.uri.substr(tile.content.uri.indexOf('?') + 1);
+        const sessionIds = uri.split('=');
+        return sessionIds[1];
```
Desplandis: Is there a non-hacky way to do this with `URLSearchParams` and get this value with the corresponding key (i.e. `sessionId`)?

src/Source/C3DTilesGoogleSource.js
```diff
@@ -0,0 +1,79 @@
...
+                    throw new Error('[C3DTilesGoogleSource]: Cannot find sessionId from the tileset while it is mandatory to request tiles.');
+                }
+            }
+            return json;
+        });
+    }
+
+    /**
+     * Adds the key and session to the tile url (non-standard behaviour, that is specific to Google 3D tiles),
+     * see https://github.com/CesiumGS/3d-tiles/issues/746
```
Desplandis, line 68: Thanks for the notice!

src/Parser/B3dmParser.js
```diff
@@ -185,66 +150,56 @@ export default {
...
-                    for (const scene of gltf.scenes) {
-                        scene.traverse(filterUnsupportedSemantics);
-                    }
-                    // Rotation managed
-                    if (gltfUpAxis === undefined || gltfUpAxis === 'Y') {
-                        gltf.scene.applyMatrix4(matrixChangeUpVectorZtoY);
-                    } else if (gltfUpAxis === 'X') {
-                        gltf.scene.applyMatrix4(matrixChangeUpVectorZtoX);
+            const init_mesh = function f_init(mesh) {
+                mesh.frustumCulled = frustumCulled;
```
Desplandis, line 155: It was set to false initially, any reason for that change?

src/Main.js
```diff
@@ -91,7 +92,7 @@ export { default as LASParser } from 'Parser/LASParser';
 export { default as ISGParser } from 'Parser/ISGParser';
 export { default as GDFParser } from 'Parser/GDFParser';
 export { default as GTXParser } from 'Parser/GTXParser';
-export { enableDracoLoader, enableKtx2Loader, glTFLoader, legacyGLTFLoader } from 'Parser/B3dmParser';
+export { default as GLTFParser, enableDracoLoader, enableKtx2Loader } from 'Parser/GLTFParser';
```
Desplandis: Okay with the change!

Desplandis (commented):
Nice changes, I did have some additional comments nonetheless but should be easy to answer. Since those are mostly aesthetics + some questions, I'll approve this PR but please answer them. =)

> Only export GLTFParser and not gltfLoader and legacyGltfLoader anymore, which is a breaking change but avoids exporting two things having the same purpose. WDYT ? Should we keep it like that or also export gltfLoader and legacyGltfLoader to avoid a breaking change until the next major version?

I'm usually cautious about breaking changes, however I don't think they are used directly by anyone...

> Some 3D Tiles related parts could be refactored / simplified (especially the B3dmParser) but I didn't go into much trouble here since we are seriously considering to handle 3D tiles parsing with an external lib very soon.

I held comments on that topic. x)

> I wanted to add more tests on the GLTFParser but I decided to wait until https://github.com/iTowns/itowns/pull/2183 is merged since I needed to load local files.

This will the object of a future PR I think!

src/Source/C3DTilesGoogleSource.js
jailln: yep, done

src/Parser/B3dmParser.js
jailln: set it back to false, it was a mistake

src/Parser/B3dmParser.js
jailln: Set it back to false and used a more syntetic syntax

jailln:
I went back on a previous change: I keep exporting `gltfLoader` and `legacyGLTFLoader`  for now, for several reasons:
* `GLTFParser` only has a `parse` method and not the other methods of `gltfLoader` and `legacyGLTFLoader`  (e.g. `load`, `loadAsync`, etc.) that itowns users might use (including me :grin: )
* don't introduce an unnecessary breaking change now
* this is a bit out of scope of this PR

Therefore I think it's ready to merge @Desplandis 

Note that I started a branch to migrate to 3d-tiles-renderer-js in which I started implementing an "itowns GltfLoader" that inherits from `THREE.Loader`, shares the same interface than GLTFLoader but can load gltf 1.0 and 2.0 files. It makes more sense to do it in the 3d-tiles-renderer-js migration since we really need it in this context.

Desplandis (approved):
Sounds good to me, all comments have been resolved! =)

Desplandis:
@jailln Since #2183 was merged before this PR, you have a conflict in the `GLTFLoader` unit test but it should be easy to resolve!

PR #2130, opened 2023-07-04
 Update style by clearing cache

a-dutremble:
## Description
Add some features to update the style of a layer.

We've added a function for redrawing a layer. It selects the layer, invalidates the cache and notifies changes.

With this function, you can change the style of the entire layer or only a specific feature in your layer.
However, it clears and updates all the layer even if you change only one feature in this layer.

We've also added an example to illustrate the function.

PS: The refresh one tile that were already refreshing before the style changement doesn't seem to refresh to the new style

### DIFF
--- a/examples/misc_redraw_layer.html
+++ b/examples/misc_redraw_layer.html
@@ -1,0 +1,154 @@
+<html>
+    <head>
+        <title>Itowns - Globe + WFS</title>
+
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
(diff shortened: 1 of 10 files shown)


jailln:
Before going into code details, some general remarks:
* There is a blinking effect when zooming out after changing the style (if the tiles where already loaded):

https://github.com/user-attachments/assets/f28fe60e-c8f6-43ac-85ff-08a3759784fd

We should fix that.

* I'm not sure we should expose a method on the view. I think a method on the layer is enough since it is only related to the layer and not the view. Then, it should be left to the user to call `view.notifyChange` as it is the current philosophy of itowns.

* I would rather have a `Layer.updateStyle(newStyle)` method to make it more explicit (or use the `style` property setter?). A bit like what has been implemented for [C3DTilesLayer](https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Layer/C3DTilesLayer.js#L351). This method should only reapply the style conversion to this layer. Without looking into too much yet (so there might be incompatibilities in doing this), this could mean:
    * taking the styling step out of `convert` - into a new `style` step for instance (this may have more sense for `Feature2Mesh` than for `Feature2Texture`)
    *  reapply this step to the layer when we call `updateStyle()` (or the `style` setter)

PR #2131, opened 2023-07-06
Remove instanciation of Style at user level

ftoromanoff:
Purpose:
Move the instanciation of new itowns.Style to Layer instead of the user level to prepare for a rework on Style.

Modifications:
- remove new itowns.Style() on example and tutorial files
- change JSdoc for Style.js

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -27,27 +27,6 @@ function mapPropertiesFromContext(mainKey, from, to, context) {
     }
 }
 
-function rgba2rgb(orig) {
-    if (!orig) {
-        return {};
-    } else if (orig.stops || orig.expression) {
-        return { color: orig };
-    } else if (typeof orig == 'string') {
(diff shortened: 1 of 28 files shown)


src/Core/Style.js
```diff
@@ -410,11 +557,15 @@ class Style {
     }
 
     /**
+<<<<<<< HEAD
      * Map drawing properties style (fill, stroke and point) from context to object.
+=======
+     * Map drawing properties style (fill, stroke, point, text and icon) from context to object.
+>>>>>>> 8effd638 (doc(style): JSDoc for StyleOptions)
```
gchoqueux: could you resolve?

src/Layer/ColorLayer.js
```diff
@@ -84,7 +83,7 @@ class ColorLayer extends RasterLayer {
         deprecatedColorLayerOptions(config);
         super(id, config);
         this.isColorLayer = true;
-        this.style = new Style(config.style);
+        this.style = config.style;
```
gchoqueux, line 87: the line could be removed?

src/Core/Style.js

src/Layer/ColorLayer.js
ftoromanoff: Layer doesn't have a style properties.
In this PR we added the fact that if there is a style properties in options at Layer we change it to a new Style(config.style).

gchoqueux (approved):

PR #2135, opened 2023-07-10
Fetching of Style.pattern images

ftoromanoff:
Modification of style.pattern image fetching:

Based on the icon image fetching, we propose a similar approch for the style.pattern image fetching
This allow Style constructor to be cleaner.

### DIFF
--- a/src/Converter/Feature2Texture.js
+++ b/src/Converter/Feature2Texture.js
@@ -2,10 +2,23 @@ import * as THREE from 'three';
 import { FEATURE_TYPES } from 'Core/Feature';
 import Extent from 'Core/Geographic/Extent';
 import Coordinates from 'Core/Geographic/Coordinates';
+import Style from 'Core/Style';
 
-const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
-const matrix = svg.createSVGMatrix();
-
+/**
+ * Draw polygon (contour, line edge and fill) based on feature vertices into canvas
+ * using the given style(s). Several styles will re-draws the polygon each one with
+ * a different style.
+ * @param      {CanvasRenderingContext2D} ctx - canvas' 2D rendering context.
+ * @param      {Number[]} vertices - All the vertices of the Feature.
+ * @param      {Object[]} indices - Contains the indices that define the geometry.
+ * Objects stored in this array have two properties, an `offset` and a `count`.
+* The offset is related to the overall number of vertices in the Feature.
+ * @param      {Object}  style - object defining the style of the polygon.
+ * @param      {Number} size - The size of the feature.
+ * @param      {Number} extent - The extent.
+ * @param      {Number} invCtxScale - The ration to scale line width and radius circle.
+ * @param      {Boolean} canBeFilled - true if feature.type == FEATURE_TYPES.POLYGON
+ */
 function drawPolygon(ctx, vertices, indices = [{ offset: 0, count: 1 }], style = {}, size, extent, invCtxScale, canBeFilled) {
     if (vertices.length === 0) {
         return;
@@ -22,63 +35,19 @@ function drawPolygon(ctx, vertices, indices = [{ offset: 0, count: 1 }], style =
 
 function _drawPolygon(ctx, vertices, indices, style, size, extent, invCtxScale, canBeFilled) {
     // build contour
-    ctx.beginPath();
+    const path = new Path2D();
+
     for (const indice of indices) {
         if (indice.extent && Extent.intersectsExtent(indice.extent, extent)) {
             const offset = indice.offset * size;
             const count = offset + indice.count * size;
-            ctx.moveTo(vertices[offset], vertices[offset + 1]);
+            path.moveTo(vertices[offset], vertices[offset + 1]);
             for (let j = offset + size; j < count; j += size) {
-                ctx.lineTo(vertices[j], vertices[j + 1]);
+                path.lineTo(vertices[j], vertices[j + 1]);
             }
         }
     }
-
-    // draw line or edge of polygon
-    if (style.stroke) {
-        strokeStyle(style, ctx, invCtxScale);
-        ctx.stroke();
-    }
-
-    // fill polygon only
-    if (canBeFilled && style.fill) {
-        fillStyle(style, ctx, invCtxScale);
-        ctx.fill();
-    }
-}
-
-function fillStyle(style, ctx, invCtxScale) {
-    if (style.fill.pattern && ctx.fillStyle.src !== style.fill.pattern.src) {
-        ctx.fillStyle = ctx.createPattern(style.fill.pattern, 'repeat');
-        if (ctx.fillStyle.setTransform) {
-            ctx.fillStyle.setTransform(matrix.scale(invCtxScale));
-        } else {
-            console.warn('Raster pattern isn\'t completely supported on Ie and edge');
-        }
-    } else if (ctx.fillStyle !== style.fill.color) {
-        ctx.fillStyle = style.fill.color;
-    }
-    if (style.fill.opacity !== ctx.globalAlpha) {
-        ctx.globalAlpha = style.fill.opacity;
-    }
-}
-
-function strokeStyle(style, ctx, invCtxScale) {
-    if (ctx.strokeStyle !== style.stroke.color) {
-        ctx.strokeStyle = style.stroke.color;
-    }
-    const width = (style.stroke.width || 2.0) * invCtxScale;
-    if (ctx.lineWidth !== width) {
-        ctx.lineWidth = width;
-    }
-    const alpha = style.stroke.opacity == undefined ? 1.0 : style.stroke.opacity;
-    if (alpha !== ctx.globalAlpha && typeof alpha == 'number') {
-        ctx.globalAlpha = alpha;
-    }
-    if (ctx.lineCap !== style.stroke.lineCap) {
-        ctx.lineCap = style.stroke.lineCap;
-    }
-    ctx.setLineDash(style.stroke.dasharray.map(a => a * invCtxScale * 2));
+    Style.prototype.applyToCanvasPolygon.call(style, ctx, path, invCtxScale, canBeFilled);
 }
 
 function drawPoint(ctx, x, y, style = {}, invCtxScale) {
--- a/src/Core/Label.js
+++ b/src/Core/Label.js
@@ -49,11 +49,13 @@ class Label extends THREE.Object3D {
      * is applied, it cannot be changed directly. However, if it really needed,
      * it can be accessed through `label.content.style`, but it is highly
      * discouraged to do so.
-     * @param {Object} [sprites] the sprites.
      */
-    constructor(content = '', coordinates, style = {}, sprites) {
+    constructor(content = '', coordinates, style = {}) {
         if (coordinates == undefined) {
             throw new Error('coordinates are mandatory to add a Label');
+        }
+        if (arguments.length > 3) {
+            console.warn('Deprecated argument sprites in Label constructor. Sprites must be configured in style argument.');
         }
 
         super();
@@ -97,14 +99,18 @@ class Label extends THREE.Object3D {
                 if (style.text.haloWidth > 0) {
                     this.content.classList.add('itowns-stroke-single');
                 }
-                style.applyToHTML(this.content, sprites);
+                style.applyToHTML(this.content)
+                    .then((icon) => {
+                        if (icon) { // Not sure if that test is needed...
+                            this.icon = icon;
+                        }
+                    });
             }
         } else {
             this.anchor = [0, 0];
             this.styleOffset = [0, 0];
         }
 
-        this.icon = this.content.getElementsByClassName('itowns-icon')[0];
         this.iconOffset = { left: 0, right: 0, top: 0, bottom: 0 };
 
         this.zoom = {
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -9,6 +9,9 @@ import itowns_stroke_single_before from './StyleChunk/itowns_stroke_single_befor
 
 export const cacheStyle = new Cache();
 
+const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+const matrix = svg.createSVGMatrix();
+
 const inv255 = 1 / 255;
 const canvas = (typeof document !== 'undefined') ? document.createElement('canvas') : {};
 const style_properties = {};
@@ -79,44 +82,44 @@ function readVectorProperty(property, options) {
     }
 }
 
-function getImage(source, value) {
-    const target = document.createElement('img');
-
-    if (typeof source == 'string') {
-        if (value) {
-            const color = new Color(value);
-            Fetcher.texture(source, { crossOrigin: 'anonymous' })
-                .then((texture) => {
-                    const img = texture.image;
-                    canvas.width = img.naturalWidth;
-                    canvas.height = img.naturalHeight;
-                    const ctx = canvas.getContext('2d', { willReadFrequently: true });
-                    ctx.drawImage(img, 0, 0);
-                    const imgd = ctx.getImageData(0, 0, img.naturalWidth, img.naturalHeight);
-                    const pix = imgd.data;
-
-                    const colorToChange = new Color('white');
-                    for (let i = 0, n = pix.length; i < n; i += 4) {
-                        const d = deltaE(pix.slice(i, i + 3), colorToChange) / 100;
-                        pix[i] = (pix[i] * d +  color.r * 255 * (1 - d));
-                        pix[i + 1] = (pix[i + 1] * d +  color.g * 255 * (1 - d));
-                        pix[i + 2] = (pix[i + 2] * d +  color.b * 255 * (1 - d));
-                    }
-                    ctx.putImageData(imgd, 0, 0);
-                    target.src = canvas.toDataURL('image/png');
-                });
-        } else {
-            target.src = source;
-        }
-    } else if (source && source[value]) {
-        const sprite = source[value];
-        canvas.width = sprite.width;
-        canvas.height = sprite.height;
-        canvas.getContext('2d').drawImage(source.img, sprite.x, sprite.y, sprite.width, sprite.height, 0, 0, sprite.width, sprite.height);
-        target.src = canvas.toDataURL('image/png');
+async function loadImage(source) {
+    let promise = cacheStyle.get(source, 'null');
+    if (!promise) {
+        promise = Fetcher.texture(source, { crossOrigin: 'anonymous' });
+        cacheStyle.set(promise, source, 'null');
     }
+    return (await promise).image;
+}
 
-    return target;
+function cropImage(img, cropValues = { width: img.naturalWidth, height: img.naturalHeight }) {
+    canvas.width = cropValues.width;
+    canvas.height = cropValues.height;
+    const ctx = canvas.getContext('2d', { willReadFrequently: true });
+    ctx.drawImage(img,
+        cropValues.x || 0, cropValues.y || 0, cropValues.width, cropValues.height,
+        0, 0, cropValues.width, cropValues.height);
+    return ctx.getImageData(0, 0, cropValues.width, cropValues.height);
+}
+
+function replaceWhitePxl(imgd, color, id) {
+    if (!color) {
+        return imgd;
+    }
+    const imgdColored = cacheStyle.get(id, color);
+    if (!imgdColored) {
+        const pix = imgd.data;
+        const newColor = new Color(color);
+        const colorToChange = new Color('white');
+        for (let i = 0, n = pix.length; i < n; i += 4) {
+            const d = deltaE(pix.slice(i, i + 3), colorToChange) / 100;
+            pix[i] = (pix[i] * d +  newColor.r * 255 * (1 - d));
+            pix[i + 1] = (pix[i + 1] * d +  newColor.g * 255 * (1 - d));
+            pix[i + 2] = (pix[i + 2] * d +  newColor.b * 255 * (1 - d));
+        }
+        cacheStyle.set(imgd, id, color);
+        return imgd;
+    }
+    return imgdColored;
 }
 
 const textAnchorPosition = {
@@ -170,10 +173,16 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
  * any [valid color string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
  * Default is no value, which means no fill.
  * If the `Layer` is a `GeometryLayer` you can use `THREE.Color`.
- * @property {Image|Canvas|string|function} [fill.pattern] - Defines a pattern to fill the
- * surface with. It can be an `Image` to use directly, or an url to fetch the pattern
+ * @property {Image|Canvas|string|object|function} [fill.pattern] - Defines a pattern to fill the
+ * surface with. It can be an `Image` to use directly, an url to fetch the pattern or an object containing
+ * the url of the image to fetch and the transformation to apply.
  * from. See [this example] (http://www.itowns-project.org/itowns/examples/#source_file_geojson_raster)
  * for how to use.
+ * @property {string} [fill.pattern.source] the url to fetch the pattern image
+ * @property {object} [fill.pattern.cropValues] the x, y, width and height (in pixel) of the sub image to use.
+ * @property {THREE.Color} [fill.pattern.color] Can be any [valid color string]
+ * (https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
+ * It will change the color of the white pixels of the source image.
  * @property {number|function} [fill.opacity] - The opacity of the color or of the
  * pattern. Can be between `0.0` and `1.0`. Default is `1.0`.
  * For a `GeometryLayer`, this opacity property isn't used.
@@ -277,12 +286,13 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
  *
  * @property {Object} [icon] - Defines the appearance of icons attached to label.
  * @property {string} [icon.source] - The url of the icons' image file.
- * @property {string} [icon.key] - The key of the icons' image in a vector tile data set.
+ * @property {string} [icon.id] - The id of the icons' sub-image in a vector tile data set.
+ * @property {string} [icon.cropValues] - the x, y, width and height (in pixel) of the sub image to use.
  * @property {string} [icon.anchor] - The anchor of the icon compared to the label position.
  * Can be `left`, `bottom`, `right`, `center`, `top-left`, `top-right`, `bottom-left`
  * or `bottom-right`. Default is `center`.
- * @property {number} [icon.size] - If the icon's image is passed with `icon.source` or
- * `icon.key`, its size when displayed on screen is multiplied by `icon.size`. Default is `1`.
+ * @property {number} [icon.size] - If the icon's image is passed with `icon.source` and/or
+ * `icon.id`, its size when displayed on screen is multiplied by `icon.size`. Default is `1`.
  * @property {string|function} [icon.color] - The color of the icon. Can be any [valid
  * color string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
  * It will change the color of the white pixels of the icon source image.
@@ -436,12 +446,13 @@ export class StyleOptions {}
  *
  * @property {Object} icon - Defines the appearance of icons attached to label.
  * @property {string} icon.source - The url of the icons' image file.
- * @property {string} icon.key - The key of the icons' image in a vector tile data set.
+ * @property {string} icon.id - The id of the icons' sub-image in a vector tile data set.
+ * @property {string} icon.cropValues - the x, y, width and height (in pixel) of the sub image to use.
  * @property {string} icon.anchor - The anchor of the icon compared to the label position.
  * Can be `left`, `bottom`, `right`, `center`, `top-left`, `top-right`, `bottom-left`
  * or `bottom-right`. Default is `center`.
- * @property {number} icon.size - If the icon's image is passed with `icon.source` or
- * `icon.key`, its size when displayed on screen is multiplied by `icon.size`. Default is `1`.
+ * @property {number} icon.size - If the icon's image is passed with `icon.source` and/or
+ * `icon.id`, its size when displayed on screen is multiplied by `icon.size`. Default is `1`.
  * @property {string|function} icon.color - The color of the icon. Can be any [valid
  * color string](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value).
  * It will change the color of the white pixels of the icon source image.
@@ -505,12 +516,6 @@ class Style {
         defineStyleProperty(this, 'fill', 'base_altitude', params.fill.base_altitude, base_altitudeDefault);
         defineStyleProperty(this, 'fill', 'extrusion_height', params.fill.extrusion_height);
 
-        if (typeof this.fill.pattern == 'string') {
-            Fetcher.texture(this.fill.pattern).then((pattern) => {
-                this.fill.pattern = pattern.image;
-            });
-        }
-
         this.stroke = {};
         defineStyleProperty(this, 'stroke', 'color', params.stroke.color);
         defineStyleProperty(this, 'stroke', 'opacity', params.stroke.opacity, 1.0);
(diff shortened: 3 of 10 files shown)


src/Core/Style.js
```diff
@@ -549,7 +558,8 @@ class Style {
 
         this.icon = {};
         defineStyleProperty(this, 'icon', 'source', params.icon.source);
-        defineStyleProperty(this, 'icon', 'key', params.icon.key);
+        defineStyleProperty(this, 'icon', 'id', params.icon.id);
+        defineStyleProperty(this, 'icon', 'cropValues', params.icon.id);
```
ftoromanoff: params.icon.cropValues

src/Core/Label.js
```diff
@@ -97,14 +96,17 @@ class Label extends THREE.Object3D {
                 if (style.text.haloWidth > 0) {
                     this.content.classList.add('itowns-stroke-single');
                 }
-                style.applyToHTML(this.content, sprites);
+                style.applyToHTML(this.content)
+                    .then(() => {
+                        this.icon = this.content.getElementsByClassName('itowns-icon')[0];
+                    });
```
gchoqueux: could you use `await`?

src/Core/Label.js
```diff
@@ -97,14 +96,17 @@ class Label extends THREE.Object3D {
...
             }
         } else {
             this.anchor = [0, 0];
             this.styleOffset = [0, 0];
+            this.icon = this.content.getElementsByClassName('itowns-icon')[0];
```
gchoqueux: why you move `this.icon = this.content.getElementsByClassName('itowns-icon')[0];` ?

src/Core/Style.js
```diff
@@ -680,7 +693,11 @@ class Style {
             this.fill.color = color;
             this.fill.opacity = readVectorProperty(layer.paint['fill-opacity']) || opacity;
             if (layer.paint['fill-pattern'] && sprites) {
-                this.fill.pattern = getImage(sprites, layer.paint['fill-pattern']);
+                this.fill.pattern = {
+                    id: layer.paint['fill-pattern'],
+                    source: sprites.source,
+                    cropValues: sprites[layer.paint['fill-pattern']],
+                };
```
gchoqueux: Generic functions should be encapsulated to handle `pattern`

By example with a class `Pattern`

src/Layer/LayerUpdateStrategy.js
```diff
@@ -11,8 +11,9 @@ export const STRATEGY_GROUP = 1;
 export const STRATEGY_PROGRESSIVE = 2;
 export const STRATEGY_DICHOTOMY = 3;
 
-function _minimizeNetworkTraffic(node, nodeLevel, currentLevel) {
-    if (node.pendingSubdivision) {
+function _minimizeNetworkTraffic(node, nodeLevel, currentLevel, source) {
+    // TO DO source.isVectorTileSource is a temp fix for pendingSubdivision
+    if (node.pendingSubdivision && !source.isVectorTileSource) {
```
gchoqueux, line 17: could you use `isVectorSource` ?

src/Core/Style.js
```diff
@@ -79,44 +79,51 @@ function readVectorProperty(property, options) {
...
+
+function drawImage(img, cropValues = { width: img.naturalWidth, height: img.naturalHeight }) {
+    canvas.width = cropValues.width;
+    canvas.height = cropValues.height;
+    const ctx = canvas.getContext('2d', { willReadFrequently: true });
+    ctx.drawImage(img,
+        cropValues.x || 0, cropValues.y || 0, cropValues.width, cropValues.height,
+        0, 0, cropValues.width, cropValues.height);
+    return ctx.getImageData(0, 0, cropValues.width, cropValues.height);
+}
```
gchoqueux: it's crop image method?

src/Core/Style.js
```diff
@@ -79,44 +79,51 @@ function readVectorProperty(property, options) {
...
+
+async function getImage(source, options) {
```
gchoqueux: Would it be possible to have just one method, instead of two `getImage` and `loadImage` methods?

src/Core/Style.js
```diff
@@ -79,44 +79,51 @@ function readVectorProperty(property, options) {
...
+        if (!imgdColored) {
+            const pix = imgd.data;
+            const color = new Color(options.color);
+            const colorToChange = new Color('white');
+            for (let i = 0, n = pix.length; i < n; i += 4) {
+                const d = deltaE(pix.slice(i, i + 3), colorToChange) / 100;
+                pix[i] = (pix[i] * d +  color.r * 255 * (1 - d));
+                pix[i + 1] = (pix[i + 1] * d +  color.g * 255 * (1 - d));
+                pix[i + 2] = (pix[i + 2] * d +  color.b * 255 * (1 - d));
+            }
```
gchoqueux: Would it be possible to encapsulate my coloring method?

src/Core/Label.js
ftoromanoff: It's in the constructor, so no, it's not possible.

src/Core/Label.js
ftoromanoff: cause in one case it's asynchrone (case style.isStyle is true)  and when it's false it's synchrone.
When style.isStyle we need to wait for the creation of these element (with style.applyToHTML()) to be able to get them.

src/Core/Style.js
ftoromanoff: I don't really understand what you mean here...  These lines are in setFromVectorTileLayer() which is in some way a parser of VectorTile data to style.

src/Layer/LayerUpdateStrategy.js
ftoromanoff: all VectorTileSource are VectorSource but all VectorSource aren't VectorTileSource. I'm not sure we want to change the behaviour of all VectorSources here...

src/Core/Style.js
ftoromanoff: loadImage(), will fetch the source image and getImage() will apply to tha source image modifications (crop and coloring), if needed, and send it to the canva.s
It's 2 differentes methods. I do'nt really understand how i can merge the 2 but i can change the name if it's not explicite enought ?

src/Core/Style.js
ftoromanoff: I change the name with other modifications. Is it better ?

src/Core/Style.js
ftoromanoff: I did modify it. Is it better ?

src/Core/Style.js
ftoromanoff: I moved it all in a function. Is it better ?

src/Core/Style.js
ftoromanoff: corrected

src/Core/Style.js
```diff
@@ -755,14 +769,45 @@ class Style {
         return this;
     }
 
+    /**
+     * Applies the style.fill to a polygon of the texture canvas
+     * @param {CanvasRenderingContext2D} txtCtx The Context 2D of the texture canvas
+     * @param {DOMMatrix} matrix The DOM matrix scaled to generate pattenr (if any)
```
jailln: typo pattern (same below)

src/Converter/Feature2Texture.js
```diff
@@ -22,44 +23,29 @@ function drawPolygon(ctx, vertices, indices = [{ offset: 0, count: 1 }], style =
 
 function _drawPolygon(ctx, vertices, indices, style, size, extent, invCtxScale, canBeFilled) {
```
jailln, line 36: Could you take this opportunity to add jsdoc to this method?

src/Converter/Feature2Texture.js
```diff
@@ -22,44 +23,29 @@ function drawPolygon(ctx, vertices, indices = [{ offset: 0, count: 1 }], style =
...
             for (let j = offset + size; j < count; j += size) {
-                ctx.lineTo(vertices[j], vertices[j + 1]);
+                path.lineTo(vertices[j], vertices[j + 1]);
             }
         }
     }
 
     // draw line or edge of polygon
     if (style.stroke) {
+        // TO DO add posibility of using a pattern and move to Style as fillPolygon()
```
jailln: Can you:
* Do part of this todo and move `strokeStyle` to `Style` but only keep one method in `Style` (your method `applyToCanvasPolygon`) that manages the stroke and the fill (and maybe separate stroke and fill in two methods local to `Style`), WDYT ? This should be quick to implement and would harmonize the stroke and fill management.
* Open an issue for the other part of the todo: "use a pattern for stroke style"

src/Core/Label.js
```diff
@@ -49,9 +49,8 @@ class Label extends THREE.Object3D {
      * is applied, it cannot be changed directly. However, if it really needed,
      * it can be accessed through `label.content.style`, but it is highly
      * discouraged to do so.
-     * @param {Object} [sprites] the sprites.
      */
-    constructor(content = '', coordinates, style = {}, sprites) {
+    constructor(content = '', coordinates, style = {}) {
```
jailln, line 53: `Label` is exposed in the public API, could you add a deprecation warning if there is 4 parameters please ?

src/Core/Label.js
```diff
@@ -97,14 +96,16 @@ class Label extends THREE.Object3D {
                 if (style.text.haloWidth > 0) {
                     this.content.classList.add('itowns-stroke-single');
                 }
-                style.applyToHTML(this.content, sprites);
+                style.applyToHTML(this.content)
+                    .then(() => {
+                        this.icon = this.content.getElementsByClassName('itowns-icon')[0];
```
jailln: I might be wrong but from my understanding we are not sure the icon element has been created in the dom when `applyToHTML` is resolved (because the domelement is created in a callback [called when the icon is loaded](https://github.com/iTowns/itowns/blob/master/src/Core/Style.js#L859-L863). We might need to use another `await` in `applyToHTML` or return a promise that resolves with the `domElement` icon instead, WDYT?

src/Converter/Feature2Texture.js
```diff
@@ -22,44 +23,29 @@ function drawPolygon(ctx, vertices, indices = [{ offset: 0, count: 1 }], style =
...
-            ctx.fillStyle.setTransform(matrix.scale(invCtxScale));
-        } else {
-            console.warn('Raster pattern isn\'t completely supported on Ie and edge');
-        }
-    } else if (ctx.fillStyle !== style.fill.color) {
-        ctx.fillStyle = style.fill.color;
-    }
-    if (style.fill.opacity !== ctx.globalAlpha) {
-        ctx.globalAlpha = style.fill.opacity;
+        Style.prototype.applyToCanvasPolygon.call(style, ctx, matrix, invCtxScale, path);
```
jailln: Why not use `style.applyToCanvasPolygon(ctx, matrix, invCtxScale, path)` ?

src/Core/Style.js
```diff
@@ -277,12 +283,13 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
  *
  * @property {Object} [icon] - Defines the appearance of icons attached to label.
  * @property {string} [icon.source] - The url of the icons' image file.
- * @property {string} [icon.key] - The key of the icons' image in a vector tile data set.
+ * @property {string} [icon.id] - The id of the icons' sub-image in a vector tile data set.
+ * @property {string} [icon.cropValues] - the x, y, width and height (in pixel) of the sub image to use..
```
jailln: typo: `..` -> `.`

src/Core/Style.js
```diff
@@ -436,12 +443,13 @@ export class StyleOptions {}
  *
  * @property {Object} icon - Defines the appearance of icons attached to label.
  * @property {string} icon.source - The url of the icons' image file.
- * @property {string} icon.key - The key of the icons' image in a vector tile data set.
+ * @property {string} icon.id - The id of the icons' sub-image in a vector tile data set.
```
jailln, line 449: Can you add a deprecation warning for `icon.key`, manage the case where it is still used to avoid breaking the API and add a deprecation message in your commit

src/Core/Style.js
```diff
@@ -566,6 +569,9 @@ class Style {
     drawingStylefromContext(context) {
         const style = {};
         if (this.fill.color || this.fill.pattern || context.globals.fill) {
+            if (typeof this.fill.pattern === 'string') {
```
jailln: I'd rather this to be handled in the constructor

src/Core/Style.js
```diff
@@ -755,14 +769,45 @@ class Style {
         return this;
     }
 
+    /**
+     * Applies the style.fill to a polygon of the texture canvas
+     * @param {CanvasRenderingContext2D} txtCtx The Context 2D of the texture canvas
+     * @param {DOMMatrix} matrix The DOM matrix scaled to generate pattenr (if any)
+     * @param {DOMMatrix} invCtxScale The DOM matrix scaled to generate pattenr (if any)
```
jailln: the inverse DOM matrix

src/Core/Style.js
```diff
@@ -743,9 +753,13 @@ class Style {
             }
 
             // additional icon
-            const key = readVectorProperty(layer.layout['icon-image']);
-            if (key) {
-                this.icon.key = key;
+            const icon = readVectorProperty(layer.layout['icon-image']);
+            if (icon) {
+                this.icon.id = icon;
+                this.icon.source = sprites?.source || icon;
```
jailln: According [to the mapbox style spec](https://docs.mapbox.com/style-spec/reference/layers/#layout-symbol-icon-image), `icon-image` is the "name of image in sprite" so I don't think we should set `icon.source` to `icon` when sprite is not defined and log a warning instead, WDYT?

src/Core/Style.js
```diff
@@ -755,14 +769,45 @@ class Style {
...
+
     /**
      * Applies this style to a DOM element. Limited to the `text` and `icon`
      * properties of this style.
      *
      * @param {Element} domElement - The element to set the style to.
-     * @param {Object} sprites - the sprites.
      */
-    applyToHTML(domElement, sprites) {
+    async applyToHTML(domElement) {
```
jailln, line 862: API modifications should be handled with a deprecation message

src/Core/Style.js
```diff
@@ -784,24 +829,20 @@ class Style {
...
+        const icon = document.createElement('img');
 
-        if (!icon) {
-            if (key && sprites) {
-                icon = getImage(sprites, key);
-            } else {
-                icon = getImage(image, color);
-            }
-            cacheStyle.set(icon, image || key, size, color);
+        if (!this.icon.cropValues && !this.icon.color) {
```
jailln, line 898: is this a way to check if it is a vector tile source or not ?

src/Layer/LayerUpdateStrategy.js
```diff
@@ -11,8 +11,9 @@ export const STRATEGY_GROUP = 1;
 export const STRATEGY_PROGRESSIVE = 2;
 export const STRATEGY_DICHOTOMY = 3;
 
-function _minimizeNetworkTraffic(node, nodeLevel, currentLevel) {
-    if (node.pendingSubdivision) {
+function _minimizeNetworkTraffic(node, nodeLevel, currentLevel, source) {
+    // TO DO source.isVectorTileSource is a temp fix for pendingSubdivision
```
jailln: Can you explain why you need this fix?

jailln (changes requested):

src/Core/Style.js
ftoromanoff: Corrected

src/Converter/Feature2Texture.js
ftoromanoff: I did add some JSdoc. For Feature2Texture.drawPolygon() and I did some modification for Style.applyToCanvasPolygon().

src/Converter/Feature2Texture.js
ftoromanoff: I did move strokeStyle() to Style as suggested. 
The issue #2210 had been opened to suggest the support of pattern image for polygon edges.

src/Core/Label.js
ftoromanoff: I did add a deprecation message, but in my opinion this is linked with a modification on VectorTileParser and I would say that it is more an internal change.

src/Core/Style.js
ftoromanoff: Corrected

src/Core/Style.js
ftoromanoff: Done. (Again In my opinion, this is more an internal modification that should be put privately later on)

src/Core/Style.js
ftoromanoff: I corrected the JSdoc as invCtxScale is a number (a ratio) and not a Matrix...

src/Core/Style.js
ftoromanoff: Indeed, if there is 'icon-image' there should be 'sprites'. I changed the conditional test.

src/Core/Style.js
ftoromanoff: Added. And same thought as before.

src/Core/Style.js
ftoromanoff: No, it's to test if we want the icon directly as the image described (by url) or if we need to modify that image (split it with 'cropValues' or change color with 'color'). In the case there is no modification, the canvas handle directly an url to create the canavas. so we don't need to load the image a priori.

src/Layer/LayerUpdateStrategy.js
ftoromanoff: With this PR I noticed a bug with the refreshing of VectorTileLayer. The modification done on this PR makes that bug much more visible. As some level in some circonstance, we don't load all tile of a fix zoom and thus we got a view with tile at differents zoom level.


As it is not really the purpose of that PR to correct that bug, I just added this fix, that seems to solve it but probably not in the best way.

src/Converter/Feature2Texture.js
ftoromanoff: Because currently style is an object and not an instance of Style.
Style.drawingStylefromContext() return a style object and not a Style instance. This will be changed in the following PR.

src/Core/Style.js
ftoromanoff: Currently drawingStylefromContext() is called before the instanciation of a Style, thus we should do that check here to be able to use the mapPropertiesFromContext().
It's also to prevent a breaking change as in my opinon fill.pattern should always be an object describing the source image to use and the modification (crop and or color change) we want (or not) to apply.

src/Core/Label.js
ftoromanoff: What if i add a new event 'loaded' that will be send when the icon is loaded ?
 and then catch the event in Label.js
 ```
                style.applyToHTML(this.content)
                    .then((icon) => {
                        if (icon) {
                            icon.addEventListener('loaded', ()  => {
                                this.icon = this.content.getElementsByClassName('itowns-icon')[0];
                            });
                        }
                    });
```
See last commit f145fb6734c14fbbf6291d0c82aee42141f31b07

File: src/Core/Style.js [hunk 23]
```diff
@@ -755,14 +777,83 @@ class Style {
...
+     * @param {CanvasRenderingContext2D} txtrCtx The Context 2D of the texture canvas.
+     * @param {Path2D} polygon The current texture canvas polygon.
+     * @param {Number} invCtxScale The ratio to scale line width and radius circle.
+     * @param {Boolean} canBeFilled - true if feature.type == FEATURE_TYPES.POLYGON.
+     */
+    applyToCanvasPolygon(txtrCtx, polygon, invCtxScale, canBeFilled) {
+        // draw line or edge of polygon
+        if (this.stroke) {
+            // TO DO add posibility of using a pattern
+            Style.prototype._drawPolygon.call(this, txtrCtx, invCtxScale);
```
jailln: I think `_applyStrokeToPolygon` would be a better name to avoid confusion with `Feature2Texture._drawPolygon`

src/Core/Style.js
```diff
@@ -755,14 +777,83 @@ class Style {
...
+            txtrCtx.stroke(polygon);
```
jailln: move this in `_applyStrokeToPolygon` to be consistent with fill

src/Core/Style.js
```diff
@@ -755,14 +777,83 @@ class Style {
...
+        }
+
+        // fill inside of polygon
+        if (canBeFilled && this.fill) {
+            // canBeFilled can be move to StyleContext in the later PR
+            Style.prototype._fillPolygon.call(this, txtrCtx, invCtxScale, polygon);
```
jailln: rename to `_applyFillToPolygon` to be consistent with the change suggested above ?

src/Core/Style.js
```diff
@@ -755,14 +777,83 @@ class Style {
...
+     * Applies the style.fill to a polygon of the texture canvas.
+     * @param {CanvasRenderingContext2D} txtrCtx The Context 2D of the texture canvas.
+     * @param {Path2D} polygon The current texture canvas polygon.
+     * @param {Number} invCtxScale The ratio to scale line width and radius circle.
+     * @param {Boolean} canBeFilled - true if feature.type == FEATURE_TYPES.POLYGON.
+     */
+    applyToCanvasPolygon(txtrCtx, polygon, invCtxScale, canBeFilled) {
+        // draw line or edge of polygon
+        if (this.stroke) {
+            // TO DO add posibility of using a pattern
```
jailln: you can remove this todo since you created [#2210](https://github.com/iTowns/itowns/issues/2210) (or add the ticket number in the comment)

src/Core/Label.js
jailln: I prefer using await or a promise that resolves with `cIcon` in `applyToHMTL` because here you rely on an event that you manually add and on its hardcoded name in two different places. Similarly, I think we should not rely on the `itowns-icon` class name to retrieve the icon. Resolving / returning `cIcon` in `applyToHTML` would resolve both these issues.

src/Converter/Feature2Texture.js
jailln: Ok, that would indeed be better :) In a PR that is already created or that is to be created? If already created, which one ?

src/Core/Label.js
```diff
@@ -49,12 +49,14 @@ class Label extends THREE.Object3D {
...
      * it can be accessed through `label.content.style`, but it is highly
      * discouraged to do so.
-     * @param {Object} [sprites] the sprites.
      */
-    constructor(content = '', coordinates, style = {}, sprites) {
+    constructor(content = '', coordinates, style = {}, ...args) {
         if (coordinates == undefined) {
             throw new Error('coordinates are mandatory to add a Label');
         }
+        if (args.length > 0) {
```
jailln: I suggest the following instead to avoid polluting the constructor signature:
```
if (arguments.length > 3) {
    console.log('Deprecated argument sprites in Label constructor. Sprites must be configured in style argument.');
}
```

src/Core/Style.js
```diff
@@ -549,7 +554,12 @@ class Style {
 
         this.icon = {};
         defineStyleProperty(this, 'icon', 'source', params.icon.source);
-        defineStyleProperty(this, 'icon', 'key', params.icon.key);
+        if (params.icon.key) {
+            console.warn('Deprecated: use icon.id instead');
```
jailln: Can you add what is deprecated ? e.g. : `icon.key is deprecated, use icon.id instead`

src/Core/Style.js
jailln: > Currently drawingStylefromContext() is called before the instanciation of a Style, thus we should do that check here to be able to use the mapPropertiesFromContext(). 

Ok, so it would be doable in the following PR where `drawingStylefromContext` will instantiate a `Style` ?

> It's also to prevent a breaking change as in my opinon fill.pattern should always be an object describing the source image to use and the modification (crop and or color change) we want (or not) to apply.

I agree, we could choose this solution (with a deprecation message), also in the following PR, WDYT ?

src/Core/Style.js
jailln: Thanks. Can you add a warning when there is an `iconImg` and no `sprites` please to help debugging ?

src/Core/Style.js
```diff
@@ -755,14 +777,83 @@ class Style {
...
      * properties of this style.
      *
      * @param {Element} domElement - The element to set the style to.
-     * @param {Object} sprites - the sprites.
+     *
+     * @returns {HTMLImageElement} - The icon
      */
-    applyToHTML(domElement, sprites) {
+    async applyToHTML(domElement, ...args) {
+        if (args.length > 0) {
```
jailln: same suggestion as above

src/Layer/LayerUpdateStrategy.js
jailln: Ok thanks, can you open a ticket explaining the bug and with the precisions you gave me and add the ticket number in this comment please ?

src/Core/Style.js
ftoromanoff: Done. I also did the same change for layer.paint['fill-pattern'] which follow the same rule.
There will be a warn before the error.

src/Core/Style.js
ftoromanoff: Indeed, I was thinking also about a name but didn't find a better one. I used your suggestion.

src/Core/Style.js

src/Core/Style.js

src/Core/Style.js
ftoromanoff: done.
I did add the issue link in the code.

src/Core/Label.js
ftoromanoff: done. I was using the same methode as i was doing elsewhere. But here we can use arguments as we don't use any of the argument objects.
Thanks for the proposal

src/Core/Style.js

src/Core/Style.js

src/Converter/Feature2Texture.js
ftoromanoff: In the 'parent' PR #2006 I merge drawingStyleFromContext() and symbolStylefromContext().
Both use styleOptions object and context to generate a Style combining all style informations. The idea is that this merged function, which can be rename getStyleFromContext(), should based on styleOptions objects and context instanciate a clear itowns Style Object (whith the default values and all).
I can take this part out in a new PR, for smaller PR.

src/Core/Label.js
ftoromanoff: I think, I did find a way to do what you propose: see 502ad620fce5a76b219dc2678b8398a806dc835b.
What do you think @jailln ?

src/Layer/LayerUpdateStrategy.js
ftoromanoff: issue opened #2214

src/Converter/Feature2Texture.js
jailln: Ok great :+1: I resolve this conversation then

src/Core/Style.js
jailln: resolving too, related to another comment, will be handled later

src/Core/Style.js
```diff
@@ -679,8 +694,15 @@ class Style {
...
-            if (layer.paint['fill-pattern'] && sprites) {
-                this.fill.pattern = getImage(sprites, layer.paint['fill-pattern']);
+            if (layer.paint['fill-pattern']) {
+                if (!sprites) {
+                    console.warn("argument sprites must not be null when using layer.paint['fill-pattern']");
+                }
+                this.fill.pattern = {
+                    id: layer.paint['fill-pattern'],
+                    source: sprites.source,
+                    cropValues: sprites[layer.paint['fill-pattern']],
```
jailln: same here, if sprites is not defined, we should not set these

src/Core/Label.js
jailln: That's the idea but I think there is a simpler implementation (sorry I paste all the function but I think it's easier to understand this way). I've quickly tested it and it seems to work, can you test also + WDYT ?

```
    async applyToHTML(domElement) {
        if (arguments.length > 1) {
            console.warn('Deprecated argument sprites. Sprites must be configured in style.');
        }
        domElement.style.padding = `${this.text.padding}px`;
        domElement.style.maxWidth = `${this.text.wrap}em`;

        domElement.style.color = this.text.color;
        if (this.text.size > 0) {
            domElement.style.fontSize = `${this.text.size}px`;
        }

        domElement.style.fontFamily = this.text.font.join(',');
        domElement.style.textTransform = this.text.transform;
        domElement.style.letterSpacing = `${this.text.spacing}em`;
        domElement.style.textAlign = this.text.justify;
        domElement.style['white-space'] = 'pre-line';

        if (this.text.haloWidth > 0) {
            domElement.style.setProperty('--text_stroke_display', 'block');
            domElement.style.setProperty('--text_stroke_width', `${this.text.haloWidth}px`);
            domElement.style.setProperty('--text_stroke_color', this.text.haloColor);
            domElement.setAttribute('data-before', domElement.textContent);
        }

        if (!this.icon.source) {
            return;
        }

        const icon = document.createElement('img');

        const addIcon = () => {
            const cIcon = icon.cloneNode();

            cIcon.setAttribute('class', 'itowns-icon');

            cIcon.width = icon.width * this.icon.size;
            cIcon.height = icon.height * this.icon.size;
            cIcon.style.color = this.icon.color;
            cIcon.style.opacity = this.icon.opacity;
            cIcon.style.position = 'absolute';
            cIcon.style.top = '0';
            cIcon.style.left = '0';

            switch (this.icon.anchor) { // center by default
                case 'left':
                    cIcon.style.top = `${-0.5 * cIcon.height}px`;
                    break;
                case 'right':
                    cIcon.style.top = `${-0.5 * cIcon.height}px`;
                    cIcon.style.left = `${-cIcon.width}px`;
                    break;
                case 'top':
                    cIcon.style.left = `${-0.5 * cIcon.width}px`;
                    break;
                case 'bottom':
                    cIcon.style.top = `${-cIcon.height}px`;
                    cIcon.style.left = `${-0.5 * cIcon.width}px`;
                    break;
                case 'bottom-left':
                    cIcon.style.top = `${-cIcon.height}px`;
                    break;
                case 'bottom-right':
                    cIcon.style.top = `${-cIcon.height}px`;
                    cIcon.style.left = `${-cIcon.width}px`;
                    break;
                case 'top-left':
                    break;
                case 'top-right':
                    cIcon.style.left = `${-cIcon.width}px`;
                    break;
                case 'center':
                default:
                    cIcon.style.top = `${-0.5 * cIcon.height}px`;
                    cIcon.style.left = `${-0.5 * cIcon.width}px`;
                    break;
            }

            cIcon.style['z-index'] = -1;
            domElement.appendChild(cIcon);
            return cIcon;
        };

        const iconPromise = new Promise((resolve, reject) => {
            icon.onload = () => resolve(addIcon());
            icon.onerror = err => reject(err);
        });

        if (!this.icon.cropValues && !this.icon.color) {
            icon.src = this.icon.source;
        } else {
            const img = await loadImage(this.icon.source);
            const imgd = cropImage(img, this.icon.cropValues);
            const imgdColored = replaceWhitePxl(imgd, this.icon.color, this.icon.id || this.icon.source);
            canvas.getContext('2d').putImageData(imgdColored, 0, 0);
            icon.src = canvas.toDataURL('image/png');
        }

        return iconPromise;
    }
```

src/Core/Style.js
ftoromanoff: If sprite are not defined but layer.paint['fill-pattern'] or layer.layout['icon-image'] are we should then get an error and not just a warning.
So from what i understand we should modifying the error message that is currently send with the warning.
That's what i propose in the last commit

src/Core/Label.js
ftoromanoff: I did so test and it seems that the test icon.complete is indeed not necessary.
I propose in the last commit some modification inspired by your code. I put out the addicon function from applyToHTML to prevent it being created many time and I fell like it makes the code a bit more readable (and from my point of view a bit more similare to applyToCanvasPolygon), WDYT ?

src/Core/Style.js
```diff
@@ -692,14 +692,16 @@ class Style {
...
-                if (!sprites) {
-                    console.warn("argument sprites must not be null when using layer.paint['fill-pattern']");
+                try {
+                    this.fill.pattern = {
+                        id: layer.paint['fill-pattern'],
+                        source: sprites.source,
+                        cropValues: sprites[layer.paint['fill-pattern']],
+                    };
+                } catch (err) {
+                    err.message = "argument sprites must not be null when using layer.paint['fill-pattern']";
```
jailln: ok :+1: can you add the layer id in the message please so it is easier to find the culprit? (same for the other try catch block)

src/Core/Label.js
jailln: ok :+1: can you just update the doc of `applyToHTML` (especially the @returns please?)

ftoromanoff: done

src/Core/Style.js
ftoromanoff: I add the Vector Tile layer id information on the message. We do'tn have acces to the itowns.layer as this fucntion is called by the VectorTileSource.

jailln (approved):

Issue #1853, opened 2022-05-05
Cesium has more detailed tiles texture render on zoom


alexLuky:
@gchoqueux @mgermerie 

I faced with issue that itowns renders tiles textures much less detailed than cesium does from the same source.

## Your Environment
* itowns version used: 2.38.1
* b3dm tiles
* link to json: hided

## Context
Here is cesium:
![image](https://user-images.githubusercontent.com/43720642/166926025-3b3beb3b-f489-49b4-b456-29944ee566d0.png)

Here is itowns:
![image](https://user-images.githubusercontent.com/43720642/166926703-5c6f7a33-54c9-4279-a593-0a3fe8476e11.png)

Also itowns has darker light wich makes the scene more nasty. How can I make it lighter? Should I use three,js light instances for this?

## Steps to Reproduce (for bugs)
Here is the itowns code:
```
const tiles3dLayer = new itowns.C3DTilesLayer(datasetTilesSet3dObject.id, {
    name: datasetTilesSet3dObject.name,
    source: new itowns.C3DTilesSource({
      url: datasetTilesSet3dObject.url,
    }),
    sseThreshold: 1,
}, this.itownsViewer)
itowns.View.prototype.addLayer.call(this.itownsViewer, tiles3dLayer)
```

## Expected Behavior
I expect that LOD will be increased to maxim while zooming. Now it looks like itowns just stops to update the LOD on zoom.

Any thoughts how can I achieve the same render quality as cesium does? I've attached the link to the source 3dTiles set for experiments.

jailln:
see https://github.com/iTowns/itowns/issues/1859#issuecomment-1239206538

mgermerie:
The complete 3d-tiles support is planned in the short term road map. We are currently studying possibilities in terms of libraries to enhance support (as mentioned in https://github.com/iTowns/itowns/discussions/1917).

jailln:
@alexLuky this should be resolved with #2143 , can you test it and let me know please?

alexLuky:
@jailln it's fantastic news. I surely would. Give me some time!

jailln:
Closing this it has been fixed (and still works with the new OGC3DTilesLayer based on 3DTilesRendererJS). Feel free to reopen if needed.

Issue #2126, opened 2023-06-30
3D tiles -wrong bounding volume box orientation


jogarnier:
When I zoom or pan some tiles just not render or whole tileset destructs until the view angle or zoom will not be changed. (really similar to https://github.com/iTowns/itowns/issues/1859) but with bouding boxes

## Your Environment

https://github.com/iTowns/itowns/assets/111570217/4d987951-2f27-4566-beca-4a92e55e256f

* Version used: 2.39.0
* Browser Name and version: Chrome 114.0.5735.199 / Firefox 114.0.2
* Operating System and version (desktop or mobile): Windows 10
* Link to your project:  https://github.com/jogarnier/itowns/tree/feat/google3dTiles (you'll need an instanciation of a C3DTilesGoogleSource with your own Google api key, example is not provided yet)

## Steps to Reproduce (for bugs)

1. https://github.com/jogarnier/itowns/tree/feat/google3dTiles
2. add a C3DTilesGoogleSource with your own Google api key
3. deactivate globe layer view
4. zoom on globe

## Expected Behavior
Tiles rendered should stay visible until replaced by a lower / higher level tile

## Possible Cause/Fix/Solution
Bouding boxes seems to be non oriented as expected ?

jailln:
It's because bounding volumes of type 'box' are interpreted as AABB while there are OBB in [this method](https://github.com/iTowns/itowns/blob/master/src/Core/3DTiles/C3DTBoundingVolume.js#L70).

jailln:
Fixed by the move to 3DTilesRendererJS

Issue #1926, opened 2022-10-14
Clarify EPSG:4326 and EPSG:4979 mix


jailln:
A small shortcut is taken in the current implementation: EPSG:4326 is assumed to be 3D (with an altitude value) while it is actually 2D (only lat / long); its 3D counterpart is [EPSG:4979](https://epsg.io/4979) (lat / long / alt). Moreover, instantiating a Coordinates object with EPSG:4979 with long/lat/altitude throws an error (probably because EPSG:4979 is not defined with proj4.defs()). Even if the confusion between EPSG:4326 and EPSG:4979 is common, I think we should clarify this in the code by at least defining EPSG:4979 by default and make sure that it works when defining coordinates with this code.

PR #2143, opened 2023-07-17
3D Tiles fixes and perf improvements

### LINKED ISSUE #1859: Unstable 3d tiles rendering in itowns
When I zoom or pan some tiles just not render or whole tileset destructs until the view angle or zoom will not be changed.

## Your Environment
- itowns version used: 2.38.1
- b3dm tiles
- link to json: hided

## Context
Here is videos to see the behavior:
1. https://youtu.be/ei3bzoLoTrs
2. https://youtu.be/EzqhBOsuaBs
3. https://youtu.be/XywzT3f29ps

I've tried:
- to set the sseThreshold = 0.3 and 100
- tileset.geometricError = 1 and 100 and other values
But these options don't solve the issue.

```
const tiles3dLayer = new itowns.C3DTilesLayer(datasetTilesSet3dObject.id, {
    name: datasetTilesSet3dObject.name,
    source: new itowns.C3DTilesSource({
      url: datasetTilesSet3dObject.url,
    }),
    sseThreshold: 1,
}, this.itownsViewer)
itowns.View.prototype.addLayer.call(this.itownsViewer, tiles3dLayer)
```

## Expected Behavior
I expect that all tiles in camera focus will be rendered no matter the zoom or view angle.
This issue and the [previous](https://github.com/iTowns/itowns/issues/1853) are critical for us and if there is no solution for them we'll have to avoid the itowns in our projects.

### LINKED ISSUE #1888: 3D Tiles debug tools: bounding spheres are not displayed
## Your Environment
* Version used: 2.38.2
* Browser Name and version: Firefox and Chrome
* Operating System and version (desktop or mobile): Linux

## Context

3D Tiles debug tools "bounding box" checkbox does not work for 3D Tiles tilesets with bounding spheres. There is no error in the console. It is limiting for debugging such tilesets.

1. Create an example with a 3D Tiles tileset with bounding spheres (e.g. https://storage.yandexcloud.net/gis-3d-models/data/tiles3d/Valaam_Aerial_Survey/Scene/Cesium.json from #1859) and with a 3D Tiles debug UI (e.g. `debug.create3dTilesDebugUI(menuGlobe.gui, view, tileset, d);`)
2. Zoom to the tileset. Camera position for the above-mentioned tileset:
```
        var placement = {
            coord: new itowns.Coordinates('EPSG:4326', 30.950135, 61.395949, 0),
            range: 3000,
            tilt: 22,
        }
```
4. Check the "Bounding box" checkbox

## Expected Behavior
Bounding spheres should be displayed (the code exists but I don't know if it ever worked ?).

## Actual Behavior
Nothing is displayed.

jailln:
Some 3D tiles fixes:
* Fix a bug that impacted display of 3D tiles with sphere bounding volumes, closes #1859
* Fix display of 3D Tiles bounding volumes of type sphere, closes #1888 and #1853

Perf improvement:
* Improve perfs for 3D tiles with bounding region by converting bounding regions to bounding spheres. Previously, bounding regions where converted to [OBB](https://github.com/iTowns/itowns/blob/master/src/Renderer/OBB.js) and the AABB englobing the OBB was used for culling and SSE computation. Now, they are converted to bounding spheres and the sphere is used for culling and SSE computation. Depending on the cases it can be a bigger approximation than before but it is faster and I didn't see any visual discrepencies. I tested the perf differences on a 3D Tiles photogrammetry tileset (~500Go, 13732 tiles with only bounding regions) that I cannot share unfortunately. On the master, we spend most of our time initializing the bounding region:

before-PR-0

before-PR-1

After the PR, we spend most of the time in parsing and preparing for rendering:

after-PR-0
after-PR-1

I also tried to implement a conversion to [threejs' OBB](https://threejs.org/docs/index.html?q=obb#examples/en/math/OBB) and to implement a culling algorithm on OBB (SSE computation for OBB is easy) but didn't succeed completely (can be seen on [the following branch](https://github.com/jailln/itowns/tree/fix/GEO-18291-3dtiles-mel-history-save)). Another way would have been to [improve OBB.setFromExtent](https://github.com/jailln/itowns/blob/fix/GEO-18291-3dtiles-mel-history-save/src/Renderer/OBB.js#L106) which is the culprit but I haven't found another way of implementing it yet. Propositions are welcome :)

BREAKING CHANGE: Remove region, box and sphere properties of C3DTBoundingVolume. They have been replaced by volume property which contains a THREE.Box3 (for box) or a THREE.Sphere (for sphere or region). Initial bounding volume type can be retrieved with the initialVolumeType property.

### DIFF
--- a/src/Core/3DTiles/C3DTBoundingVolume.js
+++ b/src/Core/3DTiles/C3DTBoundingVolume.js
@@ -1,44 +1,122 @@
 import * as THREE from 'three';
-import Extent from '../Geographic/Extent';
-import OBB from '../../Renderer/OBB';
-import C3DTilesTypes from './C3DTilesTypes';
+import Ellipsoid from 'Core/Math/Ellipsoid';
+import Coordinates from '../Geographic/Coordinates';
+import { C3DTilesTypes, C3DTilesBoundingVolumeTypes } from './C3DTilesEnums';
 
-const matrix = new THREE.Matrix4();
-const center = new THREE.Vector3();
-const size = new THREE.Vector3();
-const extent = new Extent('EPSG:4326', 0, 0, 0, 0);
-const sphereCenter = new THREE.Vector3();
+const ellipsoid = new Ellipsoid();
+
+// bounding box scratch variables
+const boxSize = new THREE.Vector3();
+const boxCenter = new THREE.Vector3();
+
+// Bounding region scratch variables
+const southEastUpCarto = new Coordinates('EPSG:4326');
+const southEastUpVec3 = new THREE.Vector3();
+const northWestBottomCarto = new Coordinates('EPSG:4326');
+const northWestBottomVec3 = new THREE.Vector3();
+const radiusScratch = new THREE.Vector3();
+
+// Culling scratch value
 const worldCoordinateCenter = new THREE.Vector3();
+
+/**
+ * Bounding region is converted to a bounding sphere to simplify and speed computation and culling. This function
+ * computes a sphere enclosing the bounding region.
+ * @param {Object} region - the parsed json from the tile representing the region
+ * @param {THREE.Matrix4} tileMatrixInverse - the inverse transformation matrix of the tile to transform the produced
+ * sphere from a global to a reference local to the tile
+ * @return {THREE.Sphere} a sphere enclosing the given region
+ */
+function initFromRegion(region, tileMatrixInverse) {
+    const east = region[2];
+    const west = region[0];
+    const south = region[1];
+    const north = region[3];
+    const minHeight = region[4];
+    const maxHeight = region[5];
+
+    const eastDeg = THREE.MathUtils.radToDeg(east);
+    const westDeg = THREE.MathUtils.radToDeg(west);
+    const southDeg = THREE.MathUtils.radToDeg(south);
+    const northDeg = THREE.MathUtils.radToDeg(north);
+
+    northWestBottomCarto.setFromValues(westDeg, northDeg, minHeight);
+    ellipsoid.cartographicToCartesian(northWestBottomCarto, northWestBottomVec3);
+
+    southEastUpCarto.setFromValues(eastDeg, southDeg, maxHeight);
+    ellipsoid.cartographicToCartesian(southEastUpCarto, southEastUpVec3);
+
+    const regionCenter = new THREE.Vector3();
+    regionCenter.lerpVectors(northWestBottomVec3, southEastUpVec3, 0.5);
+    const radius = radiusScratch.subVectors(northWestBottomVec3, southEastUpVec3).length() / 2;
+
+    const sphere = new THREE.Sphere(regionCenter, radius);
+    sphere.applyMatrix4(tileMatrixInverse);
+
+    return sphere;
+}
+
+/**
+ * Create a bounding box from a json describing a box in a 3D Tiles tile.
+ * @param {Object} box - the parsed json from the tile representing the box
+ * @return {THREE.Box3} the bounding box of the tile
+ */
+function initFromBox(box) {
+    // box[0], box[1], box[2] = center of the box
+    // box[3], box[4], box[5] = x axis direction and half-length
+    // box[6], box[7], box[8] = y axis direction and half-length
+    // box[9], box[10], box[11] = z axis direction and half-length
+    boxCenter.set(box[0], box[1], box[2]);
+    boxSize.set(box[3], box[7], box[11]).multiplyScalar(2);
+    const box3 = new THREE.Box3();
+    box3.setFromCenterAndSize(boxCenter, boxSize);
+    return box3;
+}
+
+/**
+ * Creats a bounding sphere from a json describing a sphere in a 3D Tiles tile.
+ * @param {Object} sphere - the parsed json from the tile representing the sphere
+ * @returns {THREE.Sphere} the bounding sphere of the tile
+ */
+function initFromSphere(sphere) {
+    const sphereCenter = new THREE.Vector3();
+    sphereCenter.set(sphere[0], sphere[1], sphere[2]);
+    return new THREE.Sphere(sphereCenter, sphere[3]);
+}
 
 /**
  * @classdesc 3D Tiles
  * [bounding volume](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/schema/boundingVolume.schema.json)
+ * Used to represent bounding volumes and viewer request volumes. The input bounding volume (from the dataset) can be a
+ * box, a sphere or a region. Regions are transformed to spheres internally for simplification of parsing and to speed
+ * up computations such as culling.
  * @property {C3DTilesTypes} type - Used by 3D Tiles extensions
  * (e.g. {@link C3DTBatchTableHierarchyExtension}) to know in which context
  * (i.e. for which 3D Tiles class) the parsing of the extension should be done.
- * @property {THREE.Box3} box - Bounding box, defined only if the Bounding Volume
- * is a box.
- * @property {OBB} region - Bounding region, defined only if the Bounding
- * Volume is a region.
- * @property {THREE.Sphere} sphere - Bounding sphere, defined only if the
- * Bounding Volume is a sphere.
+ * @property {String} initialVolumeType - the initial volume type to be able to dissociate spheres
+ * and regions if needed since both are converted to spheres (one of {@link C3DTilesBoundingVolumeTypes})
+ * @property {THREE.Box3|THREE.Sphere} volume - The 3D bounding volume created. Can be a THREE.Box3 for bounding volumes
+ * of types box or a THREE.Sphere for bounding volumes of type sphere or region.
  * @property {object} extensions - 3D Tiles extensions of the bounding volume
  * stored in the following format:
  * {extensioName1: extensionObject1, extensioName2: extensionObject2, ...}
  */
 class C3DTBoundingVolume {
-    constructor(json, inverseTileTransform, registeredExtensions) {
+    constructor(json, tileMatrixInverse, registeredExtensions) {
         this.type = C3DTilesTypes.boundingVolume;
 
-        // Init bounding volume
         if (json.region) {
-            this.initBoundingRegion(json.region, inverseTileTransform);
+            this.initialVolumeType = C3DTilesBoundingVolumeTypes.region;
+            this.volume = initFromRegion(json.region, tileMatrixInverse);
         } else if (json.box) {
-            this.initBoundingBox(json.box);
+            this.initialVolumeType = C3DTilesBoundingVolumeTypes.box;
+            this.volume = initFromBox(json.box);
         } else if (json.sphere) {
-            this.initBoundingSphere(json.sphere);
+            this.initialVolumeType = C3DTilesBoundingVolumeTypes.sphere;
+            this.volume = initFromSphere(json.sphere);
         } else {
-            throw new Error('3D Tiles nodes must have a bounding volume');
+            throw new Error(`Unknown bounding volume type: ${json}. 3D Tiles nodes must have a bounding volume of type
+            region, box or sphere.`);
         }
 
         if (json.extensions) {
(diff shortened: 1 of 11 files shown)


src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,27 +1,103 @@
...
+
+// Culling scratch value
 const worldCoordinateCenter = new THREE.Vector3();
 
+//
+/**
+ * Bounding region is converted to a bounding sphere to simplify and speed computation and culling. This function
+ * computes a sphere enclosing the bounding region.
+ * @param {Object} region - the parsed json from the tile representing the region
+ * @param {THREE.Matrix4} inverseTileTransform - the inverse transformation matrix of the tile to transform the produced
```
valentinMachado: If the parameter passed is a matrix4, IMO tileMatrixWorldInverse/tileMatrixInverse is a better name (and fit more THREE way of naming variables)

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,27 +1,103 @@
...
+import { C3DTilesTypes, C3DTilesBoundingVolumeTypes } from './C3DTilesEnums';
+
+const ellipsoid = new Ellipsoid();
+
+// bounding box scratch variables
+const boxSize = new THREE.Vector3();
+const boxCenter = new THREE.Vector3();
+
+// Bounding region scratch variables
+const southEastUpCarto = new Coordinates('EPSG:4326');
```
valentinMachado, line 13: I am not sure but according this [specification](https://docs.ogc.org/cs/18-053r2/18-053r2.html#32) crs of the bouding region is [EPSG 4979](http://spatialreference.org/ref/epsg/4979/) and not EPSG:4326

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,27 +1,103 @@
...
+    const northDeg = THREE.MathUtils.radToDeg(north);
+
+    northWestBottomCarto.setFromValues(westDeg, northDeg, minHeight);
+    ellipsoid.cartographicToCartesian(northWestBottomCarto, northWestBottomVec3);
+
+    southEastUpCarto.setFromValues(eastDeg, southDeg, maxHeight);
+    ellipsoid.cartographicToCartesian(southEastUpCarto, southEastUpVec3);
+
+    const regionCenter = new THREE.Vector3();
+    regionCenter.addVectors(northWestBottomVec3, southEastUpVec3).divideScalar(2);
```
valentinMachado: prefer a [lerp](https://threejs.org/docs/?q=vector#api/en/math/Vector3.lerp)

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,27 +1,103 @@
...
+ * @param {Object} box - the parsed json from the tile representing the box
+ * @return {THREE.Box3} the bounding box of the tile
+ */
+function initFromBox(box) {
+    // box[0], box[1], box[2] = center of the box
+    // box[3], box[4], box[5] = x axis direction and half-length
+    // box[6], box[7], box[8] = y axis direction and half-length
+    // box[9], box[10], box[11] = z axis direction and half-length
+    boxCenter.set(box[0], box[1], box[2]);
+    boxSize.set(box[3], box[7], box[11]).multiplyScalar(2);
```
valentinMachado, line 70: you don't take into acount box[4/5/6/8/9/10] assuming that the box is XY aligned, is that the behavior expected ?

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,27 +1,103 @@
...
  * (i.e. for which 3D Tiles class) the parsing of the extension should be done.
- * @property {THREE.Box3} box - Bounding box, defined only if the Bounding Volume
- * is a box.
- * @property {OBB} region - Bounding region, defined only if the Bounding
- * Volume is a region.
- * @property {THREE.Sphere} sphere - Bounding sphere, defined only if the
- * Bounding Volume is a sphere.
+ * @property {String} initialVolumeType - the initial volume type to be able to dissociate spheres
+ * and regions if needed since both are converted to spheres (one of {@link C3DTilesBoundingVolumeTypes})
+ * @property {THREE.Box3|THREE.Sphere} volume - The 3D bounding volume created. Can be a THREE.Box3 for bounding volumes
```
valentinMachado, line 98: I think it's a poor pattern to store in a single variable different type of object (`volume`) then to use another variable (here `initialVolumeType`) to know how to manipulate `volume`.. (btw you could use [instanceof](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Operators/instanceof) instead of initialVolumeType)

you could create a wrapper abstract class called `Volume` implementing a method `isVisible(camera, tileMatrixWorld)` then specifying `SphereVolume` and `BoxVolume` (class extending `Volume`) the specific behavior.

you could also put two attributes in `C3DTBoundingVolume` instead of `volume` one called `box` and the other one `sphere` which is the solution I rather, since a `C3DTBoundingVolume` could have the both and it avoids to create an abstract class

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -47,71 +126,44 @@ class C3DTBoundingVolume {
...
+     * @returns {boolean} true if the camera is outside the viewer request volume, false otherwise.
+     */
     viewerRequestVolumeCulling(camera, tileMatrixWorld) {
-        if (this.region) {
+        if (this.initialVolumeType === C3DTilesBoundingVolumeTypes.region) {
             console.warn('Region viewerRequestVolume not yet supported');
             return true;
         }
-        if (this.box) {
+        if (this.initialVolumeType === C3DTilesBoundingVolumeTypes.box) {
```
valentinMachado, line 156: why box is not supported here and is above (constructor and `boudingVolumeCulling`) ?

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -47,71 +126,44 @@ class C3DTBoundingVolume {
...
         }
-        if (this.sphere) {
-            worldCoordinateCenter.copy(this.sphere.center);
+        if (this.initialVolumeType === C3DTilesBoundingVolumeTypes.sphere) {
+            worldCoordinateCenter.copy(this.volume.center);
             worldCoordinateCenter.applyMatrix4(tileMatrixWorld);
             // To check the distance between the center sphere and the camera
             if (!(camera.camera3D.position.distanceTo(worldCoordinateCenter) <=
-                this.sphere.radius)) {
+                this.volume.radius)) {
```
valentinMachado: you could write return the condition of the if

src/Core/3DTiles/C3DTileset.js
```diff
@@ -1,7 +1,11 @@
 import * as THREE from 'three';
 import C3DTBoundingVolume from './C3DTBoundingVolume';
-import C3DTilesTypes from './C3DTilesTypes';
+import { C3DTilesTypes } from './C3DTilesEnums';
 
+// Inverse transform of a tile, computed from the tile transform and used when parsing the bounding volume of a tile
```
valentinMachado, line 5: Same as above, sometime you use matrix and sometime transform.. I guess transform is the term used in 3DTiles specs and the other in THREE. I think one should be picked in order to harmonize code, I rather matrix because I think is better to be aligned with THREE

src/Core/3DTiles/C3DTileset.js
```diff
@@ -85,12 +89,9 @@ class C3DTileset {
         }
 
         tile.viewerRequestVolume = tile.viewerRequestVolume ?
-            new C3DTBoundingVolume(tile.viewerRequestVolume,
-                inverseTileTransform,
-                registeredExtensions) : undefined;
+            new C3DTBoundingVolume(tile.viewerRequestVolume, inverseTileTransform, registeredExtensions) : undefined;
```
valentinMachado: Prefer setting the variable to `null` instead of `undefined` to make the difference between something which is meant to be read with something just not defined

[check also](https://stackoverflow.com/questions/6604749/what-reason-is-there-to-use-null-instead-of-undefined-in-javascript)

valentinMachado (commented):
This review does not treat the solution implemented (replacing C3DTBoundingVolume by sphere) and if this is actually working.. but is more about the code in itself.

What could I do to check if the solution is working ?

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,27 +1,103 @@
...
+const southEastUpCarto = new Coordinates('EPSG:4326');
+const southEastUpVec3 = new THREE.Vector3();
+const northWestBottomCarto = new Coordinates('EPSG:4326');
+const northWestBottomVec3 = new THREE.Vector3();
+const radiusScratch = new THREE.Vector3();
+
+// Culling scratch value
 const worldCoordinateCenter = new THREE.Vector3();
 
+//
```
mgermerie: Could you remove the empty comment ?

jailln:
Hi @valentinMachado , thanks for your review and sorry for the late response. 

> What could I do to check if the solution is working ?

You can use a 3D Tiles dataset with bounding volumes of region type to test if it works. The only open source one and available online that I know is [this one](https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.0/TilesetWithRequestVolume).

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Yes. However, EPSG:4979 is the 3D extension of EPSG:4326 (same x,y axes and with a z altitude in meters). Using EPSG:4979 would be clearer but we confuse these two crs in itowns (see #1926)

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: It is a bug, 3D tiles bounding boxes are OBB and not AABB but they are interpreted as AABB. I decided to not treat it in this PR. See #2126

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: The last pattern was the previously used pattern. However, if we choose this one I would keep the `initialVolumeType` to be able to know if the sphere comes from a bounding sphere or from a bounding region, in case it may be useful for itowns' users in some way. The second pattern is also interesting, I actually thought about it when implementing this. I don't have time to implement it right now but I'll come back to it soon.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Just because it has never been implemented but it could indeed easily be supported. Same for regions. I just decided not to do it in this PR

src/Core/3DTiles/C3DTileset.js
jailln: absolutely

src/Core/3DTiles/C3DTileset.js
jailln: ok, I picked matrix

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: yes, done

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: ok, done

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: Yep you're right, done

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,44 +1,123 @@
...
+    return box3;
+}
+
+/**
+ * Creats a bounding sphere from a json describing a sphere in a 3D Tiles tile.
+ * @param {Object} sphere - the parsed json from the tile representing the sphere
+ * @returns {THREE.Sphere} the bounding sphere of the tile
+ */
+function initFromSphere(sphere) {
+    const sphereCenter = new THREE.Vector3();
```
mgermerie, line 82: Why not keeping a "global" constant for `sphereCenter` (like `boxCenter`) ?

src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -1,44 +1,123 @@
...
+    const southDeg = THREE.MathUtils.radToDeg(south);
+    const northDeg = THREE.MathUtils.radToDeg(north);
+
+    northWestBottomCarto.setFromValues(westDeg, northDeg, minHeight);
+    ellipsoid.cartographicToCartesian(northWestBottomCarto, northWestBottomVec3);
+
+    southEastUpCarto.setFromValues(eastDeg, southDeg, maxHeight);
+    ellipsoid.cartographicToCartesian(southEastUpCarto, southEastUpVec3);
+
+    const regionCenter = new THREE.Vector3();
```
mgermerie, line 49: Could you instantiate `regionCenter` outside of the method ? I believe instancing new objects like `THREE.Vector3()` repeatedly can cause performance drops with garbage collector.  I think it is better to avoid it if possible.

src/Core/3DTiles/C3DTBoundingVolume.js
mgermerie: Could you please open an issue about it ?

src/Process/3dTilesProcessing.js
```diff
@@ -265,9 +234,6 @@ export function init3dTilesLayer(view, scheduler, layer, rootTile) {
             tile.updateMatrixWorld();
             layer.tileset.tiles[tile.tileId].loaded = true;
             layer.root = tile;
-            // The extent is calculated but it's never used in 3D tiles process
-            layer.extent = boundingVolumeToExtent(layer.crs || view.referenceCrs,
-                tile.boundingVolume, tile.matrixWorld);
```
mgermerie, line 270: Do you have an idea why the extent was calculated in the first place ?

utils/debug/3dTilesDebug.js
mgermerie, line 1: Perhaps it would be interesting to display bounding region here. We could compute the OBB for bounding regions the way it was done in `C3DTilesBoundingVolume` but only in debug tools ?

I wonder if we rather display debug information that represents the original 3dtiles data (bounding regions) or information relative to what is seen and used for computing (bounding spheres). WDYT is more interesting for users @jailln and @Desplandis ?

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: ok, done

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: this was the cause of #1853 and #1888 : the same `Vector3` instance was used for all spehres bounding boxes and the center is not cloned in [Sphere constructor](https://github.com/mrdoob/three.js/blob/master/src/math/Sphere.js#L12).
For box it's ok to use the same `boxCenter` and `boxSize` instance since they are used in `setFromCenterAndSize` which are not stored directly in the box.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: No for the same reason than for the sphere bounding volumes center

src/Process/3dTilesProcessing.js
jailln: no idea

utils/debug/3dTilesDebug.js
jailln: In my opinion it is better to display what we use to perform the culling (i.e. bounding spheres now)

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: done [2216](https://github.com/iTowns/itowns/issues/2216)

utils/debug/3dTilesDebug.js
Desplandis: @jailln @mgermerie I also think this is better to display what shape the engine actually uses to perform bounds checking. At least, it may not create confusion to the end-user about the culling behavior.

src/Core/3DTiles/C3DTBoundingVolume.js
mgermerie: Ok I see. Thanks for the explanation.

src/Process/3dTilesProcessing.js
mgermerie: @gchoqueux maybe you know some history behind this ?

mgermerie (approved):

mgermerie:
Except from the conflict that should be easy to fix it's all good for me. Thanks !

PR #2149, opened 2023-07-19
fix(StateControl): use uncaught key event

AnthonyGlt:
## Description
When we click outside of the DOM view, the keyboards events aren't caught.
We can see that in the first video, I click on the Location input, and I do "Ctrl" + left click to rotate the globe, but the "ctrl" event is not caught so the rotation doesn't happen. 

https://github.com/iTowns/itowns/assets/126568810/ebea1210-86a2-4b02-bf6d-ed65b111f651

The second video is with the correction.

https://github.com/iTowns/itowns/assets/126568810/24ce4d7e-93f6-4e7d-a45b-1ba5b08485c6

### DIFF
--- a/src/Controls/StateControl.js
+++ b/src/Controls/StateControl.js
@@ -324,17 +324,24 @@ class StateControl extends THREE.EventDispatcher {
         viewCoords.copy(this._view.eventToViewCoords(event));
 
         switch (event.pointerType) {
-            case 'mouse':
+            case 'mouse': {
                 this._currentMousePressed = event.button;
 
+                if (this._currentKeyPressed === undefined) {
+                    if (event.ctrlKey) {
+                        this._currentKeyPressed = CONTROL_KEYS.CTRL;
+                    } else if (event.shiftKey) {
+                        this._currentKeyPressed = CONTROL_KEYS.SHIFT;
+                    }
+                }
                 this.currentState = this.inputToState(
                     this._currentMousePressed,
                     this._currentKeyPressed,
                     // Detect if the mouse button was pressed less than 500 ms before, and if the cursor has not moved two much
                     // since previous click. If so, set dblclick to true.
                     event.timeStamp - this._clickTimeStamp < 500
-                        && this._lastMousePressed.button === this._currentMousePressed
-                        && this._lastMousePressed.viewCoords.distanceTo(viewCoords) < 5,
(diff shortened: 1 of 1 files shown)


src/Controls/StateControl.js
```diff
@@ -324,24 +324,33 @@ class StateControl extends THREE.EventDispatcher {
...
-            case 'mouse':
+            case 'mouse': {
                 this._currentMousePressed = event.button;
-
+                let clickKeyPressed = this._currentKeyPressed;
+
+                if (clickKeyPressed === undefined) {
+                    if (event.ctrlKey) {
+                        clickKeyPressed = CONTROL_KEYS.CTRL;
+                    } else if (event.shiftKey) {
```
Desplandis: Could you please add support to the third mouse modifier, i.e. the `meta` key (and add a `META` property in `CONTROL_KEYS` mapping to the code of one of the meta key ?

```suggestion
                        clickKeyPressed = CONTROL_KEYS.CTRL;
                    } else if (event.metaKey) {
                        clickKeyPressed = CONTROL_KEYS.META;
                    } else if (event.shiftKey) {
```

src/Controls/StateControl.js
```diff
@@ -324,24 +324,33 @@ class StateControl extends THREE.EventDispatcher {
         viewCoords.copy(this._view.eventToViewCoords(event));
 
         switch (event.pointerType) {
-            case 'mouse':
+            case 'mouse': {
                 this._currentMousePressed = event.button;
-
+                let clickKeyPressed = this._currentKeyPressed;
```
Desplandis: I think it would be better to directly assign `this._currentKeyPressed` (for the 3 cases below), given that the corresponding `KeyboardEvent` will be raised later on `onKeyUp` (even when going out of focus).

Desplandis (changes requested):
Hi and thanks for your PR!
I added some comments before merging your PR!

src/Controls/StateControl.js
jailln: yep, done

src/Controls/StateControl.js
jailln: done

Desplandis (approved):
Tested and fixes issues noted in the description of this PR.
Ok with the changes!

Issue #2129, opened 2023-07-04
How to load a ThreeJS Object (GLB) ?


Exareyn:
I'm trying to load a glb on my scene and zoom on it when it's done.

I'm currently facing difficulties in displaying a glb file in Itowns. I've tried several approaches, but I haven't been able to solve the problem. I'm hoping you can help me resolve it or provide guidance on what I might be doing wrong.

I've consulted the Itowns documentation to find information on displaying glb files, but I haven't found a clear solution.
I've searched through the examples provided with Itowns, but I couldn't find a specific example for displaying a glb file. I even searched through MapLibre and Giro3D in case they have something similar.

- im' working on ubuntu on "itowns-2.39.0"  

I suppose I need to create a custom Layer but even like that I'm still lost. I'm kinda new to the Web-SIG

PR #2155, opened 2023-07-25
examples(Glb): load a glb on 3D map

Exareyn:
**Description**

Add an example demonstrating how to load a glb (glTF binary) model on a 3D Map

**Motivation and Context**
I needed a way to load any glb into a 3D Map and the only existing example merged recently that i found didn't really worked with any glb. So I made this example that should work with every glb. Hope it helps other people who need something like this too.
https://github.com/iTowns/itowns/issues/2129#issue-1788051659

This is what I got using the example I found after using it with different glb :
bim2

**Changes Made**
    Created a new example file named "itowns_glb.html".

The example should load a glb like in the next screen :
bim4

If something is wrong or need to be updated let me know. I'm also working on a glTFLayer to make it easier to create and load a glb anywhere but it is not ready yet.

And I don't know where I should push my glb asset used on the screens for testing if it's needed.

This was made during an intership at [Futurmap](https://www.futurmap.com/)

### DIFF
--- a/examples/itowns_GLB.html
+++ b/examples/itowns_GLB.html
@@ -64,8 +64,8 @@
                         "EPSG:4326",
                         4.829501189714537,
                         45.78510982365657
-                    ).as("EPSG:4978");
-                    model.position.set(coord.x, coord.y, coord.z);
+                    ).as("EPSG:4978").toVector3();
+                    model.position.add(coord);
 
                     // Set his own position to work with Itowns referencement
                     object = model;
--- a/examples/itowns_GLB.html
+++ b/examples/itowns_GLB.html
@@ -1,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Display a Glb/Gltf object with iTowns</title>
+
+        <meta charset="UTF-8" />
+        <link rel="stylesheet" type="text/css" href="css/example.css" />
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+
+    <body>
+        <div id="viewerDiv"></div>
+        <script src="../dist/itowns.js"></script>
+        <script type="text/javascript">
+            var THREE = itowns.THREE
+            // iTowns namespace defined here
+            var viewerDiv = document.getElementById("viewerDiv");
+
+            var placement = {
+                coord: new itowns.Coordinates(
+                    "EPSG:4326",
+                    4.827361189714537,
+                    45.78537982365657,
+                    10
+                ),
+                tilt: 22,
+                heading: 0,
+                range: 2840,
+            };
+
+            var view = new itowns.GlobeView(viewerDiv, placement);
+
+            var orthoSource = new itowns.TMSSource({
+                crs: "EPSG:3857",
+                isInverted: true,
+                format: "image/png",
+                url: "http://osm.oslandia.io/styles/klokantech-basic/${z}/${x}/${y}.png",
+                attribution: {
+                    name: "OpenStreetMap",
+                    url: "http://www.openstreetmap.org/",
+                },
+                tileMatrixSet: "PM",
+            });
+
+            var orthoLayer = new itowns.ColorLayer("Ortho", {
+                source: orthoSource,
+            });
+
+            view.addLayer(orthoLayer);
+
+            // Load a new gltf object
+            itowns.glTFLoader.load(
+                "https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/lampadaire/scene.gltf ",
+                (gltf) => {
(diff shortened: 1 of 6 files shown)


examples/itowns_GLB.html
```diff
@@ -0,0 +1,92 @@
...
+            view.addLayer(orthoLayer);
+
+            // Load a new gltf object
+            itowns.glTFLoader.load(
+                "https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/models/lampadaire/scene.gltf ",
+                (gltf) => {
+                    var model = gltf.scene;
+                    model.scale.set(10,10,10)
+                    model.updateMatrixWorld();
+                    model.rotation.set(Math.PI / 2.1, 0, -Math.PI / 4.1);
```
jailln: Why `Math.PI / 2.1` and not `Math.PI / 2.0`? (same for `4.1`)

examples/itowns_GLB.html
```diff
@@ -0,0 +1,92 @@
...
+
+                    // Set coordinates of the GLB
+                    coord = new itowns.Coordinates(
+                        "EPSG:4326",
+                        4.829501189714537,
+                        45.78510982365657
+                    ).as("EPSG:4978");
+                    model.position.set(coord.x, coord.y, coord.z);
+
+                    // Set his own position to work with Itowns referencement
```
jailln: could be simplified by adding `coord` to `model.position` ? : 

```
coord = new itowns.Coordinates(
                        "EPSG:4326",
                        4.829501189714537,
                        45.78510982365657
                    ).as("EPSG:4978").toVector3();
model.position.add(coord);
```

jailln (commented):
Thanks for your contribution :)

You can upload your glb asset to the following repository by opening a new PR: https://github.com/iTowns/iTowns2-sample-data (along with a license file and if its license allows it).

Can you add a functional test in [test/functional](https://github.com/iTowns/itowns/tree/master/test/functional) please? (all itowns examples are tested in functional tests so we easily detect issues). A minimal test is to check that it runs (like [here](https://github.com/iTowns/itowns/blob/master/test/functional/3dtiles_25d.js#L13)). Other ideas of tests are to check if the model has been correctly added in the scene and to pick into the scene to check if the model is displayed at the correct coordinates.

examples/itowns_GLB.html
Exareyn: Well when I write 2.0 and 4.0 it's not perfect for some reason that i don't know yet
Capture d’écran du 2023-07-26 11-37-46

And with 2.1 and 4.1 it's like this :
Capture d’écran du 2023-07-26 11-38-55

And it's like that for every glb I tested (only 5)

jailln: I think you need to apply the following two rotations to correctly align yout model with the ground:
* `Math.PI / 2.0` to X axis to go from gltf y-up convention to itowns z-up convention
* One rotation to align the z-up axis of your model with the geodesic normal of the globe where your model is positioned. You can use [Coordinates.geodesicNormal](https://github.com/iTowns/itowns/blob/master/src/Core/Geographic/Coordinates.js#L191) to get it. We do something similar in the [collada example](https://github.com/iTowns/itowns/blob/master/examples/misc_collada.html#L100)

mgermerie:
Hi, thanks for this new example.
Could you update the initial `placement` so that the initial viewer is zoomed a bit more on your model ?

Also, if that's ok for @LoicUTC and @PhilippeMarc-IGN, could you rename your example to something like `misc_gltf_model.html` and add it in `examples/config.json` under `Miscellaneous` section ?

examples/itowns_GLB.html
Exareyn: I didn't really understand the whole thing but I made something better I hope.

jailln: yes, thanks :+1:

examples/misc_gltf_model.html
```diff
@@ -0,0 +1,96 @@
...
+                    // Set coordinates of the GLB
+                    coord = new itowns.Coordinates(
+                        "EPSG:4326",
+                        4.829501189714537,
+                        45.78510982365657,
+                        12
+                    );
+
+                    // Set position and angle of the glb
+                    model.position.copy(coord.as(view.referenceCrs));
```
jailln, line 68: I think you can replace this line by:
```suggestion
                    model.position.add(coord.as(view.referenceCrs));
```
And remove lines 73-80

examples/misc_gltf_model.html
```diff
@@ -0,0 +1,96 @@
...
+
+                    // // Set his own position to work with Itowns referencement
+                    object = model;
+                    const boundingBox = new THREE.Box3();
+                    boundingBox.setFromObject(object);
+                    const center = boundingBox.getCenter(new THREE.Vector3());
+
+                    object.position.x += object.position.x - center.x;
+                    object.position.y += object.position.y - center.y;
+                    object.position.z += object.position.z - center.z;
```
jailln, line 80: I think you can remove these lines.

jailln:
Hi @Exareyn, are you still active on this subject? Are you able to implement the last modifications asked in the review? Can we help you in any way for doing them?

PR #2156, opened 2023-07-25
MapBox: get access to feature coordinates for parametrable style.base_alti

ftoromanoff:
Currently Vector Tile style is defined using the MapBox Stylesheet enter in the VectorTileSource and there is no way to modify it outside of the itowns code.

This PR aims to be able to add dynamic function allowing the use of the feature altitude as an argument.

Other modifications:
- reorganization of code lines in Feature2Mesh.js to be similar between all featureTo[Point/Line/Polygon/ExtrudedPolygon] functions
- - add a new example using the MapBox Streets v8 tileSet flux
- change order of arguments in feature.pushCoordinates() 
-> to homogenize arg order of feature.pushCoordinates() and feature.pushCoordinatesValues() 

Breaking changes: 
- feature.pushCordinates(): change order of arguments.
- feature.pushCordinatesValues(): change in signature.


examples/vector_tile_3d_mesh_Mapbox.html
```diff
@@ -0,0 +1,131 @@
...
+            // ---------- DISPLAY CONTEXTUAL DATA : ----------
+
+            // Add two elevation layers, each with a different level of detail. Each layer's properties are defined in
+            // a json file.
+            function addElevationLayerFromConfig(config) {
+                config.source = new itowns.WMTSSource(config.source);
+                view.addLayer(new itowns.ElevationLayer(config.id, config))
+                    .then(debugMenu.addLayerGUI.bind(debugMenu));
+            }
+            // itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
```
jailln: remove

examples/vector_tile_3d_mesh_Mapbox.html
```diff
@@ -0,0 +1,131 @@
...
+                    config.source = new itowns.WMTSSource(config.source);
+                    var layer = new itowns.ColorLayer('Ortho', config);
+                    return view.addLayer(layer);
+                });
+
+            // ---------- DISPLAY VECTOR TILED BUILDING DATA AS 3D MESHES : ----------
+
+            // // Define the source of the building data : those are vector tiled data from the geoportail.
+            const buildingsSource = new itowns.VectorTilesSource({
+                style: 'http://localhost:8081/testMvt.json',
```
jailln: Can you publish this service or use a published one ?

examples/vector_tile_3d_mesh_Mapbox.html
```diff
@@ -0,0 +1,131 @@
...
+                }
+            }
+
+            // Create a FeatureGeometryLayer to support building data.
+            var buildingsLayer = new itowns.FeatureGeometryLayer('VTBuilding',{
+                batchId: function (property, featureId) { return featureId; },
+                source: buildingsSource,
+                zoom: {
+                    min: 15,
+                    // max: 13,
```
jailln: remove

examples/vector_tile_3d_mesh_Mapbox.html
```diff
@@ -0,0 +1,131 @@
...
+                },
+                style: {
+                    fill: {
+                        base_altitude: (p, c) => {
+                            return (altitudeBuildings(p, c) || 0) + p.min_height;
```
jailln: this computation can be done in `altitudeBuildings` directly

src/Parser/VectorTileParser.js
```diff
@@ -4,12 +4,25 @@ import { VectorTile } from '@mapbox/vector-tile';
...
 import { FeatureCollection, FEATURE_TYPES } from 'Core/Feature';
 import { deprecatedParsingOptionsToNewOne } from 'Core/Deprecated/Undeprecator';
+import Coordinates from 'Core/Geographic/Coordinates';
 
 const worldDimension3857 = globalExtentTMS.get('EPSG:3857').planarDimensions();
 const globalExtent = new Vector3(worldDimension3857.x, worldDimension3857.y, 1);
 const lastPoint = new Vector2();
 const firstPoint = new Vector2();
 
+function project(x, y, tileNumbers, tileExtent) {
```
jailln, line 16: Can you add some comments to avoid having to get into the maths to understand what is projected on what each time we come accross this function plase ?:)

src/Process/FeatureProcessing.js
```diff
@@ -34,6 +34,8 @@ export default {
         const zoomDest = extentsDestination[0].zoom;
 
         // check if it's tile level is equal to display level layer.
+        // TO DO updata at all level asked
```
jailln, line 37: can you create an issue please? (or add it to the issue on vector data you already created)

src/Core/Feature.js
```diff
@@ -135,14 +135,13 @@ export class FeatureGeometry {
 
     /**
      * Push new coordinates in vertices buffer.
-     * @param {Coordinates} coordIn The coordinates to push.
      * @param {Feature} feature - the feature containing the geometry
+     * @param {Coordinates} coordIn The coordinates to push.
      */
-    pushCoordinates(coordIn, feature) {
+    pushCoordinates(feature, coordIn) {
```
jailln, line 140: this change seems not necessary and it breaks the API, I would prefer that we dont do it.

src/Core/Feature.js
```diff
@@ -154,23 +153,25 @@ export class FeatureGeometry {
...
+     * @param {number} coordIn.x the x coordinate (in a local systeme).
+     * @param {number} coordIn.y the y coordinate (in a local systeme).
+     * @param {THREE.Vector3} [coordIn.normal] the normal on coordinates (only for `EPSG:4978` projection).
+     * @param {Coordinates} [coordProj] An optional argument containing the geodesic coordinates in EPSG:4326
+     * It allows the user to get access to the feature coordinates to set style.base_altitude.
+    */
+    pushCoordinatesValues(feature, coordIn, coordProj) {
+        const altitude = this.baseAltitude(feature, coordProj);
+
+        _setGeometryValues(this, feature, coordIn.x, coordIn.y, altitude, coordIn.normal);
```
jailln: shouldn't we also modify `_setGeometryValues` to take the following arguments:
`_setGeometryValues(this, feature, coordIn, altitude)` and let it get `coordIn.x`, `coordIn.y` and `coordIn.normal` when it needs it?

jailln (commented):
* rename example to `vector_tile_3d_mesh_Mapbox` to `vector_tiles_3d_mesh_mapbox`
* add the example to eamples config (`examples/config.json`)
* add breaking changes to you commit message (the modifications you made to the `Feature` API)

examples/vector_tile_3d_mesh_Mapbox.html
ftoromanoff: I separate both to show more easily the part where we get the ground elevation data for a given building. The p.min_height is a specificité of the mapbow streets V8 building tile set (for building parts that are not directly on the ground).

src/Process/FeatureProcessing.js
ftoromanoff: It's the first item in my issue on vector tile. I tried to clarified it on my list.

src/Core/Feature.js
ftoromanoff: I did this change for the 2 methods: pushCoordinates() and pushCoordinatesValues() to be more similaire as both do the same thing, the second one had been added for the unique case of vectorTile support. The purpose at he end is to have only 1 methode to do both.

examples/vector_tile_3d_mesh_Mapbox.html
ftoromanoff: done

examples/vector_tile_3d_mesh_Mapbox.html
ftoromanoff: done

examples/vector_tile_3d_mesh_Mapbox.html
ftoromanoff: done

src/Parser/VectorTileParser.js
ftoromanoff: I added a little comment to say from where i got the function.

src/Core/Feature.js
ftoromanoff: It might be an idea, But i don't know if it's needed here.

examples/vector_tile_3d_mesh_Mapbox.html
jailln: ok, I found it clearer to put everything in one function but it's not critical so I'll let you decide

src/Core/Feature.js
jailln: Ok, can't we go all the way and keep only one without modifying its argument list ? and deprecate the other (by calling the new one) ; so we don't have a breaking change

src/Core/Feature.js
jailln: It's not mandatory but it would make sense to do it now in my opinion since you started simplifying `pushCoordinatesValues`

src/Core/Feature.js
ftoromanoff: It would be the topic of a new PR aiming at clearing all the process of creating new itowns.Features I think.
It's in the idea mostly link with the way how is handle base_altitude.

jailln: I think it would be interresting to do at least what I propose in my previous comment in this PR so we don't break the API.

test/functional/vector_tile_3d_mesh_mapbox.js
```diff
@@ -0,0 +1,26 @@
...
+    let result;
+    before(async () => {
+        result = await loadExample('examples/vector_tile_3d_mesh_mapbox.html', this.fullTitle());
+    });
+
+    it('should run', async () => {
+        assert.ok(result);
+    });
+
+    it('load features data', async function _2() {
```
jailln: rename to `should correctly load features at given coordinates`

ftoromanoff: Corrected

src/Core/Feature.js
ftoromanoff: I added a deprecation message. It should avoid Breaking the API.

src/Core/Feature.js
ftoromanoff: Done

src/Core/Feature.js
jailln: Thanks! Can you do the same for `pushCoordinatesValues` please?

jailln: Also, can you add `BREAKING CHANGE: <description of the changes>` in the commits that modify these methods please?

ftoromanoff: 1) I didn't change the way pushCoordinatesValues() is called, there is no deprecation there.
2) I did add the BREAKING CHANGE message but indeed not in the good commit. I will correct that mistake now.

jailln: 1. From what I see you moved its signature from `pushCoordinatesValues(feature, long, lat, normal)` to `pushCoordinatesValues(feature, coordIn, coordProj)`

src/Core/Feature.js
```diff
@@ -154,23 +157,25 @@ export class FeatureGeometry {
...
-     * @param {number} long The longitude coordinate.
-     * @param {number} lat The latitude coordinate.
-     * @param {THREE.Vector3} [normal] the normal on coordinates (only for `EPSG:4978` projection).
-     */
-    pushCoordinatesValues(feature, long, lat, normal) {
-        const altitude = this.baseAltitude(feature);
-
-        _setGeometryValues(this, feature, long, lat, altitude, normal);
+     * @param {Object} coordIn An object containing the coordinates values to push.
+     * @param {number} coordIn.x the x coordinate (in a local systeme).
```
jailln: `system` without "e" at the end (same for `coordIn.y`)

jailln (approved):

mgermerie (approved):
I made a small update on examples comments. Thanks for the PR !

PR #2165, opened 2023-08-02
Harmonize Dispose Function Names

jamesmyose:
## Description
- Renamed delete function names to dispose

## Motivation and Context
This is work on #1937. I found some delete functions that I could change, but I was unsure about changing anything else.
This PR should have the same code as https://github.com/iTowns/itowns/pull/1938. Because the commit history was flawed, I rebased this branch based on the current master


jamesmyose:
@ftoromanoff Hey, so I was able to get a rebase working, It looks like it did rebase successfully, but the commits still show the original PR's commits and the new ones on top. I also can't guarantee that all the changes were transferred correctly. I do have a backup branch so I try it again if necessary. Alternatively, though obviously I don't prefer this, I can create a new branch and manually make all the changes into a new commit.

ftoromanoff:
Hello and than you !
Next step will now be to sqaush all (or almost all) commits to be able to have only few commits with clear modifications. Cause 40 commits is really a lot.
Squashing commits means fusing them together and for that, when you do a rebase -i origin/master you can use "s" to sqaush one specifique commit with the previous one.

jamesmyose:
@ftoromanoff Hey, so I merged the most recent changes into my master branch and then I ran `git rebase -i origin/master`. I squashed all the commits into one and I ran through the rebase. I'm noticing that there are still 45 commits in this pull request though.

PR #2171, opened 2023-08-23
[PointMaterial] Custom attenuated points parameters

ketourneau:
## Description
Allow custom attenuatedpoints parameters

## Motivation and Context
We need to adjust attenuated points parameters for a better rendering in our case.

### DIFF
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -70,10 +70,9 @@ class C3DTilesLayer extends GeometryLayer {
      * removed from the scene.
      * @param {C3DTExtensions} [config.registeredExtensions] 3D Tiles extensions managers registered for this tileset.
      * @param {String} [config.pntsMode= PNTS_MODE.COLOR] {@link PointsMaterials} Point cloud coloring mode. Only 'COLOR' or 'CLASSIFICATION' are possible. COLOR uses RGB colors of the points, CLASSIFICATION uses a classification property of the batch table to color points.
-     * @param {String} [config.pntsSize= PNTS_SIZE.VALUE] {@link PointsMaterials} Point cloud size mode. Only 'VALUE' or 'ADAPTIVE' are possible. VALUE use constant size, ADAPTIVE compute size depending on distance from point to camera.
-     * @param {Number} [config.pntsAdaptiveScale=1] Scale factor used by 'ADAPTIVE' size mode
-     * @param {Number} [config.pntsMinAdaptiveSize=3] Minimum scale used by 'ADAPTIVE' size mode
-     * @param {Number} [config.pntsMaxAdaptiveSize=10] Maximum scale used by 'ADAPTIVE' size mode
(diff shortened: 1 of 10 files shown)


AnthonyGlt (commented):
Thank you for the MR.
Could you clean your commits to only keep the last one. The "package" files shouldn't be edited

AnthonyGlt:
This is a nice feature, could you make these parameters (minAdaptiveSize/maxAdaptiveSize/adaptiveScale) editable by the user when he's loading a pointcloud using a C3DTilesLayer?
Like what was done with the pntsMode parameter. There is an example in examples/3dtiles_pointcloud.html.
By searching where pntsMode is called, it could be possible to implements the same process with pntsMinAdaptiveSize/pntsMaxAdaptiveSize & pntsAdaptiveScale .
(could be helpful https://github.com/iTowns/itowns/pull/2042)
Don't hesitate to ask if help is needed

AnthonyGlt:
Maybe could be a good idea to introduce a parameter  `size_mode: ...., // PNTS_SIZE.ADAPTIVE || PNTS_SIZE.VALUE ` that would automatically set the size at 0 (when PNTS_SIZE.ADAPTIVE is set) and use minAdaptiveSize/maxAdaptiveSize/adaptiveScale defined by the user.
WDYT ?

ketourneau:
@AnthonyGlt Thx for your code review.
I applied the change.
I add `size_mode` parameter, I think it's a more intuitive way to set point size behaviour.
Is everything good ?

AnthonyGlt:
The name ADAPTIVE should be changed since the implementation isn't really the adaptive one, it could be called ATTENUATED.
I think it's a good idea to be able to switch between modes like you did.

src/Layer/C3DTilesLayer.js
```diff
@@ -84,13 +87,19 @@ class C3DTilesLayer extends GeometryLayer {
 
         this.pntsMode = PNTS_MODE.COLOR;
         this.classification = config.classification;
-
+        this.pntsSize = PNTS_SIZE.VALUE;
```
AnthonyGlt:
```suggestion
        this.pntsSizeMode = PNTS_SIZE_MODE.VALUE;
```
Maybe rename like this to avoid confusion with the size. WDYT ?

AnthonyGlt:
Looks good to me. After tests, does it answer your needs ?

src/Layer/PointCloudLayer.js
```diff
@@ -160,6 +160,7 @@ class PointCloudLayer extends GeometryLayer {
             this.material.opacity = this.opacity;
             this.material.transparent = this.opacity < 1;
             this.material.size = this.pointSize;
+            this.material.preSSE = context.camera.preSSE;
```
AnthonyGlt, line 163: @Desplandis we should implement something similar in `C3DTilesLayer`, for now only the default `preSSE` value (1.0) will be used in the material when using a C3DTilesLayer, right ?

Concerning this MR, I think your implementation is enough.

src/Layer/C3DTilesLayer.js
ketourneau: In fact it's better, I renamed it

AnthonyGlt (approved):

Issue #2170, opened 2023-08-22
error THREE.GLTFLoader: setKTX2Loader 


vladimir-rybalko:
### Your Environment
**Version used**: current master
**Browser Name and version**: Chrome and FF
**Operating System and version (desktop or mobile)**: Windows
**Link to your project**: https://codesandbox.io/s/itowns-3d-tiles-kt6r4x
### Context
When loading 3D tiles, an error occurs - _THREE.GLTFLoader: setKTX2Loader must be called before loading KTX2 textures._
Data prepared in Agisoft Metashape.
![image](https://github.com/iTowns/itowns/assets/3933179/68df9eca-d6d9-4ca6-b921-e7b6b6b7d833)

jailln:
Some 3D Tiles contains GLTFs with textures compressed with KTX2. To load them, you need to specify to the `GLTFLoader` a [KTXLoader](https://threejs.org/docs/index.html?q=ktx2loade#examples/en/loaders/KTX2Loader) that points to the ktx2 decoder library. I implemented an helper function `enableKtx2Loader` (similar to `enableDracoLoader`) to do so in #2172. I tested it with your model and it works

vladimir-rybalko:
> Some 3D Tiles contains GLTFs with textures compressed with KTX2. To load them, you need to specify to the `GLTFLoader` a [KTXLoader](https://threejs.org/docs/index.html?q=ktx2loade#examples/en/loaders/KTX2Loader) that points to the ktx2 decoder library. I implemented an helper function `enableKtx2Loader` (similar to `enableDracoLoader`) to do so in #2172. I tested it with your model and it works

Thank you for your quick response.
I compiled the code from your [repository](https://github.com/jailln/itowns/tree/feat/add-ktx2-loader) and fixed some of the code in the my [sandbox](https://codesandbox.io/s/itowns-3d-tiles-kt6r4x). The error has fixed, but I don't see the model in the scene.
@jailln, сan you attach a working example of the code here so that I can understand what my problem?

jailln:
I pushed it on [this branch](https://github.com/jailln/itowns/tree/vladimir-example). You can clone it, do `npm install`, `npm start` and go to `http://localhost:8080/examples/vladimir-example.html` :

![image](https://github.com/iTowns/itowns/assets/16967916/3a2152aa-83d0-4e4d-b44f-c0766644fac6)

It has no roof though :grin: but this may come from the model?

vladimir-rybalko:
[...]

PR #2172, opened 2023-08-23
feat(3dtiles): add method to enable ktx2 loader for gltf

jailln:
## Description
Add a method to enable loading of gltf (and 3D tiles) with KTX2 compression. (similar to what was done for draco).

Note: should be moved to `GLTFParser` once #2124 is merged.

## Motivation and Context

Motivated by #2170


src/Parser/B3dmParser.js
```diff
@@ -72,6 +72,31 @@ export function enableDracoLoader(path, config) {
...
+ *
+ * The KTX2 library files are in folder itowns/examples/libs/basis/.
+ * You must indicate this path when you want to enable KTX2 decoding.
+ *
+ * For more information about KTX2, read /itowns/examples/libs/basis/README.md.
+ *
+ * @example <caption>Enable ktx2 decoder</caption>
+ * // if you copied /itowns/examples/libs/draco/ to the root folder of your project,you can set the path to './'.
+ * itowns.enableKtx2Loader('./', view.mainLoop.gfxEngine.renderer);
+ */
```
mgermerie: Could you move this part before `@params` documentation here and for `enableDracoLoader` ? I just figured that this part would appear in `renderer` param documentation, as it appears in`config` param doc in `enableDracoLoader`.

src/Parser/B3dmParser.js
```diff
@@ -72,6 +72,31 @@ export function enableDracoLoader(path, config) {
...
+ * This library is mandatory to load b3dm and gltf with KTX2 compression.
+ *
+ * The KTX2 library files are in folder itowns/examples/libs/basis/.
+ * You must indicate this path when you want to enable KTX2 decoding.
+ *
+ * For more information about KTX2, read /itowns/examples/libs/basis/README.md.
+ *
+ * @example <caption>Enable ktx2 decoder</caption>
+ * // if you copied /itowns/examples/libs/draco/ to the root folder of your project,you can set the path to './'.
+ * itowns.enableKtx2Loader('./', view.mainLoop.gfxEngine.renderer);
```
mgermerie, line 84: Maybe we could ease access to the renderer to the user (scope for another PR). WDYT ? Other than that I understand from `KTX2Loader` the need of passing a `renderer`, but would it be possible to access it in the parser in some way ?

src/Parser/B3dmParser.js
jailln: yep, done

src/Parser/B3dmParser.js
jailln: Regarding the method to access the renderer, I thought the same and implemented it (among other similar methods) in #2174 :slightly_smiling_face: 

Regarding accessing the renderer in the parser, I think we should not access it in the parser and the parser should not know the renderer because it only needs to parse and not to render. However we could avoid passing it as an argument and either store it as a global variable and retrieve it in `enableKtx2Loader` or implement a service that stores it and provides a getter (e.g. `RendererService` and do a `RendererService.getRenderer()` to get it). This pattern is not used in itowns and I'm not sure we should introduce it just for this case but I like it more than storing a global variable. Anyway I think it should be done in another PR. 

Moreover, I think these two methods (`enableKtx2Loader` and `enableDracoLoader`) should be defined in `C3DTilesLayer` for more clarity (especially in the documentation). However they need to know the `gltfParser` and the `legacyGltfParser` which is currently defined in the `B3dmParser`. Same here, we could store them as global variables and use them in the `B3dmParser` or implement a service to store and retrieve them. This should also be done later I think, especially after #2124 that introduces a single class for gltf parsing.

Sorry for the long answer :sweat_smile:

mgermerie: I can see how moving `enableKtx2Loader` and `enableDracoLoader` in `C3DTilesLayer` would ease their use, but I'm not sure I agree in terms of architecture. Those methods are "global" in a sense that they are not related to a specific 3d-tiles layer. Perhaps using them from a specific layer could be misleading.

mgermerie (approved):

src/Parser/B3dmParser.js
jailln: yes you're right

PR #2173, opened 2023-08-23
Custom points shapes

Pourfex:
## Description
Add an option to render point in shapes of Square or Circle.

## Motivation and Context
We've made this change so we can have the same behaviour as Potree's examples. Square point shape provides a better rendering in our use cases.

## Screenshots
Circle point shape : 
image

Square point shape :
image

### DIFF
--- a/examples/3dtiles_pointcloud.html
+++ b/examples/3dtiles_pointcloud.html
@@ -62,6 +62,7 @@
                 name: 'SetePC',
                 sseThreshold: 5,
                 pntsMode: itowns.PNTS_MODE.CLASSIFICATION,
+                pntsShape : itowns.PNTS_SHAPE.CIRCLE,
                 source: new itowns.C3DTilesSource({
                     url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/pnts-sete-2021-0756_6256/tileset.json',
                 }),
@@ -79,6 +80,7 @@
                 if(pntsLayer){
                     pntsLayer = pntsLayer;
                     pntsLayer.pntsMode = pntsLayer.pntsMode == itowns.PNTS_MODE.COLOR ? itowns.PNTS_MODE.CLASSIFICATION : itowns.PNTS_MODE.COLOR;
+                    pntsLayer.pntsShape = pntsLayer.pntsShape == itowns.PNTS_SHAPE.CIRCLE ? itowns.PNTS_SHAPE.CIRCLE : itowns.PNTS_SHAPE.SQUARE;
                     view.notifyChange(view.camera.camera3D);
                 }
             }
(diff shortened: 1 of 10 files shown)


AnthonyGlt (commented):
Thank you for the MR.
Could you clean your commits to only keep the last one. The "package" files shouldn't be edited.
This is a nice feature, could you make this parameter (PNTS_SHAPE/pntsShape) editable by the user when he's loading a pointcloud using a `C3DTilesLayer`?
Like what was done with the `pntsMode` parameter. There is an example in `examples/3dtiles_pointcloud.html`.
By searching where `pntsMode` is called, it could be possible to implements the same process with `pntsShape`.
(could be helpful https://github.com/iTowns/itowns/pull/2042)
Don't hesitate to ask if help is needed

examples/3dtiles_pointcloud.html
```diff
@@ -79,6 +80,7 @@
                 if(pntsLayer){
                     pntsLayer = pntsLayer;
                     pntsLayer.pntsMode = pntsLayer.pntsMode == itowns.PNTS_MODE.COLOR ? itowns.PNTS_MODE.CLASSIFICATION : itowns.PNTS_MODE.COLOR;
+                    pntsLayer.pntsShape = pntsLayer.pntsShape == itowns.PNTS_SHAPE.CIRCLE ? itowns.PNTS_SHAPE.CIRCLE : itowns.PNTS_SHAPE.SQUARE;
```
AnthonyGlt: No need to change here since the option is already present in the menu:
Screenshot from 2023-08-31 14-40-47
so this line can be deleted

src/Renderer/PointsMaterial.js
```diff
@@ -214,6 +224,7 @@ class PointsMaterial extends THREE.RawShaderMaterial {
         this.transparent = source.transparent;
         this.size = source.size;
         this.mode = source.mode;
+        this.pntsShape = source.pntsShape;
```
AnthonyGlt:
```suggestion
        this.shape = source.shape;
```

examples/3dtiles_pointcloud.html
ketourneau: Indeed, sorry for the mistake ! I removed it.

AnthonyGlt:
Is this MR finished ? You can notify the reviewer when it's the case: 
image

FYI, we usually keep only 1 commit per feature/bug fix. I'll squash your commits into one when I'll merge. Next time, when you consider your MR as completed, feel free to rebase your commit into 1 and keep 1 commit message. It means that the intermediary commit messages don't have to follow the commit message syntax because they will get squashed

Pourfex:
@AnthonyGlt 
MR is ready ! Feel frew to review.
Got it for next times, hope we can do more contributions.

AnthonyGlt (approved):

PR #2174, opened 2023-08-24
Add View helpers

jailln:
Add some helper functions in `View` to get threejs renderer, renderer size and camera to avoid having to know itowns internal structure to access them.

### DIFF
--- a/examples/laz_dragndrop.html
+++ b/examples/laz_dragndrop.html
@@ -45,23 +45,21 @@
                         view.scene.add(points);
                         points.updateMatrixWorld(true);
 
-                        var camera = view.camera.camera3D;
-
                         var lookAt = new itowns.THREE.Vector3();
                         var size = new itowns.THREE.Vector3();
                         geometry.boundingBox.getSize(size);
                         geometry.boundingBox.getCenter(lookAt);
 
-                        camera.far = Math.max(2.0 * size.length(), camera.far);
+                        view.camera3D.far = Math.max(2.0 * size.length(), camera.far);
 
                         var position = geometry.boundingBox.min.clone().add(size.multiply({ x: 0, y: 0, z: size.x / size.z }));
 
-                        camera.position.copy(position);
-                        camera.lookAt(lookAt);
+                        view.camera3D.position.copy(position);
+                        view.camera3D.lookAt(lookAt);
 
                         controls.options.moveSpeed = size.length();
 
-                        view.notifyChange(camera);
+                        view.notifyChange(view.camera3D);
                         controls.reset();
                     });
                 };
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -99,8 +99,8 @@ class GlobeView extends View {
         super('EPSG:4978', viewerDiv, options);
         this.isGlobeView = true;
 
-        this.camera.camera3D.near = Math.max(15.0, 0.000002352 * ellipsoidSizes.x);
-        this.camera.camera3D.far = ellipsoidSizes.x * 10;
+        this.camera3D.near = Math.max(15.0, 0.000002352 * ellipsoidSizes.x);
+        this.camera3D.far = ellipsoidSizes.x * 10;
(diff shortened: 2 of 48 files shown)


examples/laz_dragndrop.html
```diff
@@ -45,23 +45,21 @@
...
 
-                        var camera = view.camera.camera3D;
-
                         var lookAt = new itowns.THREE.Vector3();
                         var size = new itowns.THREE.Vector3();
                         geometry.boundingBox.getSize(size);
                         geometry.boundingBox.getCenter(lookAt);
 
-                        camera.far = Math.max(2.0 * size.length(), camera.far);
+                        cameraThree.far = Math.max(2.0 * size.length(), camera.far);
```
ftoromanoff: var cameraThree= view.getThreeJsCamera();
This line seems to be  missing ?

src/Core/Prefab/GlobeView.js
```diff
@@ -99,8 +99,8 @@ class GlobeView extends View {
         super('EPSG:4978', viewerDiv, options);
         this.isGlobeView = true;
 
-        this.camera.camera3D.near = Math.max(15.0, 0.000002352 * ellipsoidSizes.x);
-        this.camera.camera3D.far = ellipsoidSizes.x * 10;
+        this.getThreeJsCamera().near = Math.max(15.0, 0.000002352 * ellipsoidSizes.x);
```
ftoromanoff: Maybe add a line ``` const camera3D = this.getThreeJsCamera(); ``` ?

examples/laz_dragndrop.html
jailln: well spoted! done

src/Core/Prefab/GlobeView.js
jailln: ok, done

ftoromanoff:
By adding helper functions to get specific variable, should we not modify these variables to set them as private ?
We can also instead of implementing function to get variables, use setter and getter ?
I don't really know what is the best way to work around these from a architectural/api point of view...

ftoromanoff (commented):
I get the point that private variable in JS might not be well handled.

For the second point: using methode getVariable() or using setter/getter I can't say what is the best usage, we might need to discuss that internaly.

jailln:
I think setting them as private could be done but it is not related to this PR. In addition, I'm not fond of private fields implementation in JS ([side effects](https://lea.verou.me/blog/2023/04/private-fields-considered-harmful/), badly interpreted by dev tools). But this is another subject :)

Regarding getter vs the methods I implemented, in my opinion both are valid and it's only a matter of coding style and API. I don't have a strong opinion on which one we should used, as long as it's decided quickly. Please let me know, so we can move on quickly.

ftoromanoff:
> I think setting them as private could be done but it is not related to this PR. In addition, I'm not fond of private fields implementation in JS ([side effects](https://lea.verou.me/blog/2023/04/private-fields-considered-harmful/), badly interpreted by dev tools). But this is another subject :)
> 
> Regarding getter vs the methods I implemented, in my opinion both are valid and it's only a matter of coding style and API. I don't have a strong opinion on which one we should used, as long as it's decided quickly. Please let me know, so we can move on quickly.

After discussion, I think we should opt for the setter/getter methods. 
In the case of camera being already in use (by the itowns.camera) we might use ```get camera3D``` or any other name and add a deprecation warning to get camera (wich should be internaly rename _camera.

jailln:
Sorry I don't understand what you mean by:
> In the case of camera being already in use (by the itowns.camera) we might use `get camera3D` or any other name and add a deprecation warning to get camera (wich should be internaly rename _camera.

Can you give more details please ?

jailln:
I changed to implemented to getters, removed the `getRendererSize` because one can use `View.renderer` and then gets the size if one needs it. I also added a commit to document `View` properties (including `camera3D` and `renderer`).

Let me know if that's what you meant by your last comment.

ftoromanoff (approved):

PR #2179, opened 2023-08-31
Rework test functional

ftoromanoff:
The purpose of this PR is do rewrite the functionnal test to limit the failure during githubaction linked to unclear reasons.

Changes:
- fix some issue on hook_functional.js:
   - few minor corrections
   - after initialization save position only once and not after each test
   - gestion of initialization layer timing out
      - better message when encoutering an error
      - and real gestion of timeouterror (initialization of view > 50 000ms => stop camera movement)
- reorganization of the functional tests for a better read (moved to #2324 )
  - creation of folder per functionalities tested

### DIFF
--- a/test/hooks_functional.js
+++ b/test/hooks_functional.js
@@ -117,7 +117,7 @@ const loadExample = async (url, screenshotName) => {
 
     pageErrors.forEach((e) => { throw e; });
 
-    await page.waitForFunction(() => typeof (view) === 'object');
+    await page.waitForFunction(() => typeof view === 'object' && view instanceof itowns.View);
 
     await page.evaluate(() => {
         itowns.CameraUtils.defaultStopPlaceOnGroundAtEnd = true;
@@ -131,6 +131,8 @@ const loadExample = async (url, screenshotName) => {
                 itowns.CameraUtils.stop(view, view.camera3D);
             });
             await layersAreInitialized();
+        } else {
+            throw e;
         }
     }
 
@@ -206,7 +208,7 @@ export const mochaHooks = {
         });
 
         // the page all tests will be tested in
-        return browser.newPage().then((p) => { global.page = p; });
+        global.page = await browser.newPage();
     },
     // store initial position for restoration after the test
     afterAll(done) {
--- a/test/hooks_functional.js
+++ b/test/hooks_functional.js
@@ -16,6 +16,8 @@ import events from 'events';
 
 events.EventEmitter.prototype._maxListeners = 100;
 
+const TIMEOUT = 50000;
+
 let itownsServer;
 let itownsPort;
 let browser;
@@ -23,6 +25,7 @@ let browser;
 // but it's slow to start (so tests might fail on timeouts).
 // Since the 'test-examples' target depends on the 'run' target,
 // we instead run the simplest http server.
+
 function startStaticFileServer() {
     return new Promise((resolve) => {
         const ext2mime = new Map();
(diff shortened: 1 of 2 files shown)


test/functional/hooks_functional.js
```diff
@@ -77,10 +78,11 @@ function waitServerReady(port) {
...
 
-const layersAreInitialized = async () => {
-    await page.waitForFunction(() => view.mainLoop.scheduler.commandsWaitingExecutionCount() === 0
-        && view.mainLoop.renderingState === 0
-        && view.getLayers().every(layer => layer.ready), { timeout: 60000 });
+const layersInitialization = async (forceTimeOut = false) => {
+    await page.waitForFunction(timingOut => (view.mainLoop.scheduler.commandsWaitingExecutionCount() === 0
+            && view.mainLoop.renderingState === 0
+            && view.getLayers().every(layer => layer.ready)
+            && !timingOut), { timeout: 50000 }, forceTimeOut);
```
jailln: 50000 could be defined in a constant at the begining of the file

test/functional/hooks_functional.js
```diff
@@ -103,39 +105,72 @@ const waitNextRender = async page => page.evaluate(() => new Promise((resolve) =
     view.notifyChange();
 }));
 
+async function saveInitialPosition() {
+    global.initialPosition = await page.evaluate(() => {
+        let promise;
+        if (view.isGlobeView && view.controls) {
+            promise = Promise.resolve(itowns.CameraUtils.getTransformCameraLookingAtTarget(view, view.controls.camera));
+        } else if (view.isPlanarView) {
+            // TODO: make the controls accessible from PlanarView before doing
```
jailln: You can use `view.controls` also

test/functional/hooks_functional.js
```diff
@@ -239,7 +263,7 @@ exports.mochaHooks = {
                 );
                 view.controls.lookAtCoordinate(init, false);
                 view.notifyChange();
-            } else if (view.isPlanarView) {
+            } else if (view && view.isPlanarView) {
                 // TODO: make the controls accessible from PlanarView before doing
```
jailln: same remark here regarding the controls

test/functional/examples.js
```diff
@@ -0,0 +1,87 @@
+/* View */
```
jailln: What is the purpose of this file?

test/functional/hooks_functional.js
ftoromanoff: I didn't write the TODO. But if this had been done. I can try to check this indeed.

test/functional/examples.js
ftoromanoff: I removed this file from this PR. It was a proposition of reorganisation of the functionnal test files.
Move to #2324

test/hooks_functional.js
```diff
@@ -16,13 +16,16 @@ import events from 'events';
 
 events.EventEmitter.prototype._maxListeners = 100;
 
+const timeout = 50000;
```
Desplandis: Since this a global constant, could you please capitalize it?
```suggestion
const TIMEOUT = 50000;
```

test/hooks_functional.js
```diff
@@ -79,10 +82,11 @@ function waitServerReady(port) {
     });
 }
 
-const layersAreInitialized = async () => {
-    await page.waitForFunction(() => view.mainLoop.scheduler.commandsWaitingExecutionCount() === 0
-        && view.mainLoop.renderingState === 0
-        && view.getLayers().every(layer => layer.ready), { timeout: 60000 });
+const layersInitialization = async (forceTimeOut = false) => {
```
Desplandis:
```suggestion
const layersInitialize = async (forceTimeOut = false) => {
```

test/hooks_functional.js
```diff
@@ -79,10 +82,11 @@ function waitServerReady(port) {
...
+    await page.waitForFunction(forceTimeOut => (view.mainLoop.scheduler.commandsWaitingExecutionCount() === 0
+            && view.mainLoop.renderingState === 0
+            && view.getLayers().every(layer => layer.ready)
+            && !forceTimeOut), { timeout }, forceTimeOut);
```
Desplandis: It seems that `forceTimeout` is never called with a `true` value, what is the purpose of this parameter? Moreover, you could use the [`signal`](https://pptr.dev/api/puppeteer.framewaitforfunctionoptions) parameter of `waitForFunction` (of type [`AbortSignal`](https://developer.mozilla.org/fr/docs/Web/API/AbortSignal)) to abort waiting for this function.

Desplandis (changes requested):

Desplandis (approved):
Validated.

test/functional/hooks_functional.js
ftoromanoff: View.controls is not yet refactored between all the different type of Views. 
Thes part will need to be done before we can really cahnge the TODOs in here.

PR #2180, opened 2023-09-01
Feature:  Stylize points mesh.

gchoqueux:
## Description
Refactoring `Feature` and `Feature2Mesh` to improve software architecture 
and add styling points feature.

## Software architecture
 * Move `Style` responsibility from `Feature` to `Feature2Mesh`, because no stylisation must be applied when parsing.
 * Add `FeatureContext` class to encapsulate all properties and methods necessary for stylization.

## Motivation and Context
* Start delete `Style` from `Feature`
* Move `Style` in convert give points stylization feature
* Remove `altitude.min` and `max` because they were added purely for aesthetic reasons 

## Screenshots
image

### DIFF
--- a/examples/source_stream_wfs_25d.html
+++ b/examples/source_stream_wfs_25d.html
@@ -52,10 +52,12 @@
             // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
             viewerDiv = document.getElementById('viewerDiv');
 
+            const centerMap = new itowns.Coordinates('EPSG:3946', 1840839, 5172718, 0);
+
             // Instanciate PlanarView*
             view = new itowns.PlanarView(viewerDiv, extent, {
                 placement: {
-                    coord: new itowns.Coordinates('EPSG:3946', 1840839, 5172718, 0),
+                    coord: centerMap,
                     heading: 45,
                     range: 1800,
                     tilt: 30,
@@ -104,13 +106,13 @@
             var tile;
             var linesBus = [];
 
-            function altitudeLine(properties, contour) {
+            function altitudeLine(properties, ctx) {
                 var result;
                 var z = 0;
-                if (contour) {
-                    result = itowns.DEMUtils.getTerrainObjectAt(view.tileLayer, contour, 0, tile);
+                if (ctx.coordinates) {
+                    result = itowns.DEMUtils.getTerrainObjectAt(view.tileLayer, ctx.coordinates, 0, tile);
                     if (!result) {
-                        result = itowns.DEMUtils.getTerrainObjectAt(view.tileLayer, contour, 0);
+                        result = itowns.DEMUtils.getTerrainObjectAt(view.tileLayer, ctx.coordinates, 0);
                     }
                     if (result) {
                         tile = [result.tile];
@@ -121,14 +123,7 @@
             }
 
             function acceptFeatureBus(properties) {
-                if (properties.sens == 'Aller') {
-                    var line = properties.ligne;
-                    if (linesBus.indexOf(line) === -1) {
-                        linesBus.push(line);
-                        return true;
-                    }
-                }
-                return false;
+                return properties.sens === "Aller";
             }
 
             lyonTclBusSource = new itowns.WFSSource({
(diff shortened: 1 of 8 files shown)


src/Converter/Feature2Mesh.js
```diff
@@ -9,8 +9,59 @@ import OrientationUtils from 'Utils/OrientationUtils';
 import Coordinates from 'Core/Geographic/Coordinates';
 
 const coord = new Coordinates('EPSG:4326', 0, 0, 0);
+
+class FeatureContext {
```
ftoromanoff, line 13: Should the class be part of this file ? As it should also be used in Feature2Texture and in LabelLayer.
And is the name well choose as it's not the context of the feature (in itowns definition) but more of the featureGeometry with information of its feature and its featureCollection .
As it's only use for the implementation of the Style, maybe it should be put there ?

ftoromanoff (commented):
As you changed the context from:
```const context = { globals, properties: () => geometry.properties };```
to
```
context.setCollection(options.collection);
...
context.globals = { point: true }
...
context.setGeometry(geometry);
```

It should also be done in file Feature2Texture and LabelLayer that also use 'context'.

File: src/Converter/Feature2Mesh.js [hunk 2]
```diff
@@ -650,6 +669,8 @@ export default {
                 options.layer = this;
             }
 
+            options.collection = collection;
```
ftoromanoff: Why not use ```context.setCollection(options.collection)``` here instead of each sub functions ?
and use ``` options.matrixWorldInverse```  and ``` options.crs```  or context.get matrixWordInverse and crs

gchoqueux:
> It should also be done in file Feature2Texture and LabelLayer that also use 'context'.

yes in other PR, because there's many changes in this current PR

src/Converter/Feature2Mesh.js
gchoqueux: I don't think style should be responsible for context.

ftoromanoff: So should we create a new complet new file for it ?

src/Converter/Feature2Mesh.js
gchoqueux: Yes, thanks, good catch
I pushed a proposal

gchoqueux:
@ftoromanoff with your comment on new `THREE.Group` in `FeatureMesh class`,
I've noticed that it's actually unnecessarily complex 
I push a new proposal but I'm not sure if it covers all cases of use, in particular when `base_altitude` is `undefined`

examples/source_stream_wfs_25d.html
```diff
@@ -120,15 +122,16 @@
...
             function acceptFeatureBus(properties) {
-                if (properties.sens == 'Aller') {
-                    var line = properties.ligne;
-                    if (linesBus.indexOf(line) === -1) {
-                        linesBus.push(line);
-                        return true;
-                    }
+                const line = properties.ligne;
+                if (colorsLine.get(line) == undefined && properties.sens == 'Aller') {
+                    colorsLine.set(line, new itowns.THREE.Color(0xffffff * Math.random()));
```
jailln: I think it would be clearer to move all the logic related to coloring in `colorLine`

examples/source_stream_wfs_25d.html
```diff
@@ -147,37 +150,84 @@
...
+                id: 'pool',
+                typeName: "sytral:tcl_sytral.tclarret",
+                crs: 'EPSG:3946',
+                extent,
+                format: 'application/json',
+            });
 
-            function colorBuildings(properties) {
+            var lyonBusStopLayer = new itowns.FeatureGeometryLayer('lyon_tcl_bus_stop', {
+                // filter: acceptFeatureBus,
```
jailln: to remove

examples/source_stream_wfs_25d.html
```diff
@@ -300,6 +346,11 @@
                 }
             }
 
+            view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function m() {
```
jailln, line 349: I guess we wait for the globe to be initialized because we use the terrain in `altitudeLine` but can you comment it for users that will look at the example please?

src/Converter/Feature2Mesh.js
```diff
@@ -256,101 +269,86 @@ function featureToPoint(feature, options) {
...
+    const countIndices = (count - feature.geometries.length) * 2;
+    const indices = getIntArrayFromSize(countIndices, count);
+
+    let i = 0;
+    inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
+    normal.set(0, 0, 1).multiply(inverseScale);
+    // Multi line case
+    for (const geometry of feature.geometries) {
+        context.setGeometry(geometry);
+        // const context = { globals, properties: () => geometry.properties, collection: options.collection, coordinates };
```
jailln: remove?

src/Converter/Feature2Mesh.js
```diff
@@ -256,101 +269,86 @@ function featureToPoint(feature, options) {
 
 function featureToLine(feature, options) {
```
jailln, line 271: There is many similarities between `featureToPoint`, `featureToLine`, `featureToPolygon` and `featureToExtrudedPolygons`, can you factorize shared code between these methods?

src/Converter/Feature2Mesh.js
jailln: I agree, it should be in a separate file, and commented because it if is an argument of style properties functions (e.g. `base_altitude`) so its documentation would be useful to users

src/Converter/Feature2Mesh.js
```diff
@@ -9,8 +9,54 @@ import OrientationUtils from 'Utils/OrientationUtils';
...
+    }
+
+    properties() {
+        return this.#geometry.properties;
+    }
+
+    get coordinates() {
+        if (this.#worldCoord.isLocal) {
+            this.#worldCoord.isLocal = false;
+            this.#worldCoord.copy(this.#localCoordinates).applyMatrix4(this.collection.matrixWorld);
```
jailln, line 44: you can use a private boolean property (`#worldCoordAvailable` for instance) instead of worldCoord.isLocal

src/Converter/Feature2Mesh.js
```diff
@@ -9,8 +9,54 @@ import OrientationUtils from 'Utils/OrientationUtils';
...
 
 const coord = new Coordinates('EPSG:4326', 0, 0, 0);
+
+class FeatureContext {
+    #worldCoord = new Coordinates('EPSG:4326', 0, 0, 0);
+    #localCoordinates = new Coordinates('EPSG:4326', 0, 0, 0);
+    #geometry = {};
+
+    constructor() {
+        this.globals = {};
```
jailln, line 20: I'm not fond of this `globals` property; I haven't found a better way to do it yet but we can at least document it

src/Converter/Feature2Mesh.js
```diff
@@ -9,8 +9,54 @@ import OrientationUtils from 'Utils/OrientationUtils';
...
+        this.collection = c;
+        this.#localCoordinates.setCrs(c.crs);
+    }
+
+    setLocalCoordinatesFromArray(vertices, offset) {
+        this.#worldCoord.isLocal = true;
+        return this.#localCoordinates.setFromArray(vertices, offset);
+    }
+
+    properties() {
```
jailln: transform to `get properties` ?

src/Core/Style.js
```diff
@@ -41,7 +41,7 @@ export function readExpression(property, ctx) {
             }
             return property.stops[0][1];
         } else if (property instanceof Function) {
-            return property(ctx.properties());
+            return property(ctx.properties(), ctx);
```
jailln: Since properties are in the context, we could replace by `property(ctx)`; That would break the API though but  we could at least write a comment or open an issue to remember to do it at the next major itowns version.

Also, I think it should be documented (maybe in `StyleOptions`) that style options that are functions are called with these two parameters.

jailln (changes requested):

src/Converter/Feature2Mesh.js
Desplandis: Will be done #2196 which is already based on this PR.

src/Core/Style.js
Desplandis: I'm not so fond with breaking this API. We should do it with care since I'm sure this is used by multiple projects (maybe for a 3.x version).

Since there is still some developments on-going on `Style`, maybe we should wait for the stabilization of the API before documenting it?

examples/source_stream_wfs_25d.html
Desplandis: Done

examples/source_stream_wfs_25d.html
Desplandis: Done.

src/Converter/Feature2Mesh.js
Desplandis: Will be done in #2217.

src/Converter/Feature2Mesh.js
Desplandis: Done.

src/Core/Style.js
jailln: I'm ok with not breaking the API too. For this reason I feel like we should document it now since the API won't change (event if there is still work going on with `Style`). It may be done in #2196 or #2006, I will link this comment in one of these PRs so we don't forget about it.

I also think we should leave a comment in the code with the API breaking proposal to remember it when we'll prepare the 3.x version. Something starting with a keyword that we could re-use for other similar propositionns to easily find them all when we will need them. WDYT?

Desplandis: We may document it in #2196 since we are moving and using the context for all converters. For the breaking proposal, we could either add a custom jsdoc tag or just a `TOBREAK` comment!

jailln (approved):

gchoqueux:
the example source_stream_wfs_25d.html is broken ?
there isn't styling by distance

Issue #2152, opened 2023-07-21
[Three.js] Deprecation of WebGL 1 support in WebGLRenderer


Desplandis:
## Context
Since `r153`, three.js [WebGL 1 support in WebGLRenderer is deprecated][three:migration-v153] and will be removed in `r163`.

## To-Do
- [x] #2194
- [x] Remove js code assuming a WebGL1 renderer (`View`, `c3DEngine`, `Fetcher`, different materials, ...) 
- [x] Switch our shaders to glsl `300 es`

[three:migration-v153]: https://github.com/mrdoob/three.js/wiki/Migration-Guide#152--153

PR #2186, opened 2023-09-11
Deprecation of WebGL 1.0 support in c3DEngine

Desplandis:
## Description
This PR deprecates the WebGL 1.0 support option (through `isWebGL2=false`) in iTowns and removes the linked documentation and tutorial.

## Motivation and Context
Since `r153`, three.js [WebGL 1 support in WebGLRenderer is deprecated](https://github.com/mrdoob/three.js/wiki/Migration-Guide#152--153) and will be removed in `r163`.  This is part of our effort to remove WebGL 1.0 support before three.js `r163` release (circa march 2024).

See issue #2152.


src/Core/Deprecated/Undeprecator.js
```diff
@@ -74,4 +74,12 @@ export const deprecatedFeature2MeshOptions = (options) => {
     }
 };
 
+export const deprecatedC3DEngineWebGLOptions = (options) => {
+    if (options.isWebGL2 !== undefined) {
+        if (!options.isWebGL2) {
```
jailln: replace these two if by `options.isWebGL2 === false` (simpler and handles null case)

jailln (approved):

src/Core/Deprecated/Undeprecator.js
Desplandis: Changed !
I first wanted to deprecate `isWebGL2: true` as well. But we may want to use this parameter later when three will add WebGPU support !

mgermerie (approved):

PR #2187, opened 2023-09-13
fix example Gpx 3d and functional test

### LINKED ISSUE #2188: Bug on master with the example source_file_gpx_3d
On master there is a bug on the example source_file_gpx_3d 
- the gpx waypoints doesn't display
- console errors appear
image

Moreover as the functional test wasn't well design (we only test that the view is initialized even if the example add extra layer after that), it creates error in the github action from time to time.

ftoromanoff:
There is a bug on the source_file_gpx_3d.html example:
- the gpx waypoints doesn't display
- console errors appear
image

As the functional test wasn't well design (we only test that the view is initialized even if the example add extra layer after that), it create error in the github action from time to time.
I change the functionnal test to see if the example is fully loaded and corrected in the example the source of the error.

The error is linked to the fact that a style.fill is needed (at least in parent.style) when we instanciate a new Style:
https://github.com/iTowns/itowns/blob/e56bf65c29189b86b87f8b1a9b1bce2f451ff517/src/Core/Style.js#L508
and the way we set the getter:
https://github.com/iTowns/itowns/blob/e56bf65c29189b86b87f8b1a9b1bce2f451ff517/src/Core/Style.js#L144
Or in this example the style is defined during the parsing. 

This issue is solve in the PR #2135 as we remove the need for a fill.patern and in PR #2006 as we remove the definition of parent Style.

### DIFF
--- a/examples/source_file_gpx_3d.html
+++ b/examples/source_file_gpx_3d.html
@@ -72,18 +72,18 @@
                         out: {
                             crs: view.referenceCrs,
                             structure: '3d',
-                            style: {
-                                stroke : {
+                            style: new itowns.Style({
+                                stroke: {
                                     color: 'red',
                                     width: 2,
                                 },
-                                point : {
+                                point: {
                                     color: 'white',
                                 }
-                            }
+                            }),
                         }
                     }))
-                    .then(itowns.Feature2Mesh.convert({style}))
+                    .then(itowns.Feature2Mesh.convert())
                     .then(function (mesh) {
                         if (mesh) {
                             mesh.updateMatrixWorld();
--- a/test/functional/source_file_gpx_3d.js
+++ b/test/functional/source_file_gpx_3d.js
@@ -6,7 +6,11 @@ describe('source_file_gpx_3d', function _() {
         result = await loadExample('examples/source_file_gpx_3d.html', this.fullTitle());
     });
 
-    it('should run', async () => {
+    it('view initialized', async () => {
         assert.ok(result);
+    });
+
+    it('should wait for the mesh to be added to the scene', async function _it() {
+        await page.waitForFunction(() => view.scene.children.length === 5, { timeout: 10000 });
     });
 });
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-with-coverage && npm run test-functional",
     "test-dev": "npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage && npm run test-functional",
     "test-unit": "npm run base-test-unit test/unit",
-    "test-functional": "mocha -t 60000 --require test/hooks_functional.js --recursive test/functional",
+    "test-functional": "mocha -t 100000 --require test/hooks_functional.js --recursive test/functional",
     "test-with-coverage": "nyc -n src -r html cross-env npm run test-unit",
     "test-with-coverage_lcov": "nyc -n src --reporter=lcov cross-env npm run test-unit",
     "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --require @babel/register --file test/unit/bootstrap.js",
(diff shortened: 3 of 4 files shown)


test/functional/source_file_gpx_3d.js
```diff
@@ -6,7 +6,17 @@ describe('source_file_gpx_3d', function _() {
...
-    it('should run', async () => {
+    it('view initialized', async () => {
         assert.ok(result);
     });
+
+    it('should wait for the mesh to be added to the scene', async () => {
+        await page.waitForFunction(() => view.scene.children.length === 5);
+
+        const nbChildren = await page.evaluate(() => {
+            console.log();
```
AnthonyGlt: log could be removed

test/functional/source_file_gpx_3d.js
```diff
@@ -6,7 +6,17 @@ describe('source_file_gpx_3d', function _() {
...
+            return view.scene.children.length;
+        });
+        assert.equal(nbChildren, 5);
```
AnthonyGlt: Is all that necessary since `await page.waitForFunction(() => view.scene.children.length === 5);` is going to timeout if `view.scene.children.length != 5`.
Is it to check if `nbChildren` is not superior to 5 ? Can this value change after the `await page.waitForFunction(...)` ?

WDYT

examples/source_file_gpx_3d.html
```diff
@@ -72,18 +72,18 @@
                         out: {
                             crs: view.referenceCrs,
                             structure: '3d',
-                            style: {
+                            style: new itowns.Style({
                                 stroke : {
```
AnthonyGlt:
```suggestion
                                stroke: {
```

examples/source_file_gpx_3d.html
```diff
@@ -72,18 +72,18 @@
...
                                     color: 'red',
                                     width: 2,
                                 },
                                 point : {
```
AnthonyGlt:
```suggestion
                                point: {
```

test/functional/source_file_gpx_3d.js
ftoromanoff: indeed: removed.

examples/source_file_gpx_3d.html
ftoromanoff: corrected

examples/source_file_gpx_3d.html
ftoromanoff: corrected

examples/source_file_gpx_3d.html
```diff
@@ -72,18 +72,18 @@
...
-                                point : {
+                                point: {
                                     color: 'white',
                                 }
-                            }
+                            }),
                         }
                     }))
-                    .then(itowns.Feature2Mesh.convert({style}))
+                    .then(itowns.Feature2Mesh.convert())
```
AnthonyGlt, line 86: this modification couldn't have been done directly in https://github.com/iTowns/itowns/pull/2006 ?
(because you need this https://github.com/iTowns/itowns/pull/2135 maybe ?)

test/functional/source_file_gpx_3d.js
ftoromanoff: In the example, it's only after the event itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED is received that we add the mesh to the scene.
Thus we need after the view initialization to wait for the mesh to be added: that's what the waitForFunction do.

As i wanted to add an assert.equal I added the page.evaluate to get a result, but indeed you are right it's useless.

So I did remove the page.evaluate()

examples/source_file_gpx_3d.html
ftoromanoff: It's a bug that seems to have appeared in a sous PR of the #2006 and as the funcitonal test is currently not well done, it got invisible.

AnthonyGlt (approved):
Thanks for this MR :)

mgermerie (approved):

PR #2192, opened 2023-09-20
fix(picking): Fix point reference of picking

ketourneau:
All points have same reference in the return array of picking method.


AnthonyGlt:
Thanks for the PR, looks like you are right !

AnthonyGlt (approved):

PR #2196, opened 2023-09-26
Use context for all converters (and not only Feature2Mesh)

ftoromanoff:
The PR #2180 propose a new approach of Context for the Style, that is used in Feature2Mesh.
2 others converters also use the initial context;
- Feature2Texture.convert()
- LabelLayer and Label

This PR propose (based on PR #2180 ) to generalize the use of Context to the other converter.

Changes:
- export the FeatureContext 
- move it to Style
- use it for the others converters
- modify FeatureToolTips plugin  to adapt to that change

### DIFF
--- a/src/Converter/Feature2Texture.js
+++ b/src/Converter/Feature2Texture.js
@@ -3,6 +3,9 @@ import { FEATURE_TYPES } from 'Core/Feature';
 import Extent from 'Core/Geographic/Extent';
 import Coordinates from 'Core/Geographic/Coordinates';
 import Style from 'Core/Style';
+import { FeatureContext } from 'Converter/Feature2Mesh';
+
+const context = new FeatureContext();
 
 /**
  * Draw polygon (contour, line edge and fill) based on feature vertices into canvas
@@ -74,7 +77,7 @@ const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 function drawFeature(ctx, feature, extent, style, invCtxScale) {
     const extentDim = extent.planarDimensions();
     const scaleRadius = extentDim.x / ctx.canvas.width;
-    const globals = {
+    context.globals = {
         fill: true,
         stroke: true,
         point: true,
@@ -83,7 +86,7 @@ function drawFeature(ctx, feature, extent, style, invCtxScale) {
 
     for (const geometry of feature.geometries) {
         if (Extent.intersectsExtent(geometry.extent, extent)) {
-            const context = { globals, properties: () => geometry.properties };
+            context.setGeometry(geometry);
             const contextStyle = (geometry.properties.style || style).applyContext(context);
 
             if (contextStyle) {
--- a/src/Converter/Feature2Texture.js
+++ b/src/Converter/Feature2Texture.js
@@ -2,10 +2,9 @@ import * as THREE from 'three';
 import { FEATURE_TYPES } from 'Core/Feature';
 import Extent from 'Core/Geographic/Extent';
 import Coordinates from 'Core/Geographic/Coordinates';
-import Style from 'Core/Style';
-import { FeatureContext } from 'Converter/Feature2Mesh';
+import Style, { StyleContext } from 'Core/Style';
 
-const context = new FeatureContext();
+const context = new StyleContext();
 
 /**
  * Draw polygon (contour, line edge and fill) based on feature vertices into canvas
@@ -77,12 +76,6 @@ const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 function drawFeature(ctx, feature, extent, style, invCtxScale) {
     const extentDim = extent.planarDimensions();
     const scaleRadius = extentDim.x / ctx.canvas.width;
-    context.globals = {
-        fill: true,
-        stroke: true,
-        point: true,
-        zoom: extent.zoom,
-    };
 
     for (const geometry of feature.geometries) {
         if (Extent.intersectsExtent(geometry.extent, extent)) {
@@ -174,6 +167,13 @@ export default {
 
             // to scale line width and radius circle
             const invCtxScale = Math.abs(1 / scale.x);
+
+            context.globals = {
+                fill: true,
+                stroke: true,
+                point: true,
+                zoom: extent.zoom,
+            };
 
             // Draw the canvas
             for (const feature of collection.features) {
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -986,7 +986,7 @@ class Style {
         if (this.text.field.expression) {
             return readExpression(this.text.field, ctx);
         } else {
-            return this.text.field.replace(/\{(.+?)\}/g, (a, b) => (ctx.properties()[b] || '')).trim();
+            return this.text.field.replace(/\{(.+?)\}/g, (a, b) => (ctx.properties[b] || '')).trim();
         }
     }
 }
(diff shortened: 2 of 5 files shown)


jailln:
To remember when reviewing: addess [the following comment](https://github.com/iTowns/itowns/pull/2180#discussion_r1375833525)

src/Converter/Feature2Texture.js
```diff
@@ -74,7 +76,7 @@ const coord = new Coordinates('EPSG:4326', 0, 0, 0);
 function drawFeature(ctx, feature, extent, style, invCtxScale) {
     const extentDim = extent.planarDimensions();
     const scaleRadius = extentDim.x / ctx.canvas.width;
-    const globals = {
+    context.globals = {
```
jailln: could be done outside this method (right after `context` is initialized)

src/Core/Style.js
```diff
@@ -160,6 +161,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
     style[category][name] = value;
 }
 
+/**
+ * @class
+ * @classdesc StyleContext is a class to store all informations
```
jailln: `StyleContext Stores a feature metadata needed for its style computation. (e.g. its coordinates, if it should fill, stroke or draw a point, etc.)`

src/Core/Style.js
```diff
@@ -160,6 +161,61 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * about context to generate the style of each FeatureGeometry.
+ *
+ * @property {Coordinates}      worldCoord @private Coordinates of the FeatureGeometry in world system.
+ * @property {Coordinates}      localCoordinates @private Are the coordinates systeme origin local or global.
+ * @property {boolean}          isProjected @private Are the coordinates already been projected.
+ * @property {FeatureGeometry}  geometry  @private
+ * @property {Object}           globals
+ * @property {Object}           collection
+ * @property {Coordinates}      coordinates
```
jailln: Can you document them please?

src/Layer/LabelLayer.js
```diff
@@ -240,7 +242,7 @@ class LabelLayer extends GeometryLayer {
         // Converting the extent now is faster for further operation
         extent.as(data.crs, _extent);
         coord.crs = data.crs;
-        const globals = {
+        context.globals = {
```
jailln, line 245: same here, could be done outside of this method

src/Main.js
```diff
@@ -39,7 +39,7 @@ export { CAMERA_TYPE } from 'Renderer/Camera';
 
 // Internal itowns format
 export { default as Feature, FeatureCollection, FeatureGeometry, FEATURE_TYPES } from 'Core/Feature';
-export { default as Style } from 'Core/Style';
+export { default as Style, StyleContext } from 'Core/Style';
```
jailln: I think we don't need to export it; what is the usecase in exposing it in the API? @Desplandis also, what's your opinion about this?

src/Converter/Feature2Texture.js
ftoromanoff: It can indeed be moved before calling drawFeature, but becasue of globals.zoom = extent.zoom (that was already set here before) it can't be moved where context is initialized.
In a futur PR we can change the way globals works. With a co,ntext.setZoom(extent.zoom) and maybe initialized the fill, stroke and point atinitialization.

jailln: ah sorry I didn't see the last line that was hidden by github diff :grin:

src/Layer/LabelLayer.js
ftoromanoff: becasue of zoom.extent it can't really be moved.

src/Core/Style.js

src/Core/Style.js
```diff
@@ -160,6 +161,64 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
     style[category][name] = value;
 }
 
+/**
+ * @class
+ * @classdesc StyleContext stores metadata of one FeatureGeometry that are needed for its style computation:
+ * type of feature and what is needed (fill, stroke or draw a point, etc.) as well as where to get its
+ * properties and its coordinates (for base_altitude).
+ *
+ * @property {Coordinates}      worldCoord @private Coordinates of the FeatureGeometry in world system.
```
jailln: wording: `Feature coordinates in world space`

src/Core/Style.js
```diff
@@ -160,6 +161,64 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * @property {Coordinates}      localCoordinates @private Are the coordinates systeme origin local or global.
```
jailln: wrong doc. Should be `Feature coordinates in local space`

src/Core/Style.js
```diff
@@ -160,6 +161,64 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * @property {boolean}          isProjected @private Are the coordinates already been projected.
```
jailln: would be clearer if named `needToComputeWorldCoords` or `worldCoordsComputed`. Also, update the doc to `Have the world coordinates already been computed from the local coordinates`

src/Core/Style.js
```diff
@@ -160,6 +161,64 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * @property {FeatureGeometry}  geometry  @private
+ * @property {Object}           globals Style type (fill, stroke, point, text and or icon) to consider. As
```
jailln: `as well as` -> `also contains`

src/Core/Style.js
```diff
@@ -160,6 +161,64 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ *                                  well as the zoom.
+ * @property {Object}           collection The FeatureCollection on which the FeatureGeomatry is attached
```
jailln: `on which` -> `to which`

src/Core/Style.js
```diff
@@ -160,6 +161,64 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
...
+ * @property {Coordinates}      coordinates The coordinates (in world crs) of the last vertice set with
```
jailln: * `in world crs` -> `in world space` because the code suggests that the crs is the same as the crs of `localCoordinates`.
* `vertice` -> `vertices`

jailln (commented):
thanks for the update. I added some wording suggestions.

src/Core/Style.js
ftoromanoff: A vertice is the combination of 3 informations: x, y and z, if i'm not mistaken. get coordinates() get us information of the one vertice that we did set.

src/Core/Style.js
ftoromanoff: I changed the doc to move all privates properties at the end as these properties are not really the coordinates of the point but intermediate object to get local or world coordinates.

src/Core/Style.js
ftoromanoff: indeed, i did changed it (I was thinking about it too, but as it was the last modificaiton done by @Desplandis, i thiught i should not changed it)

src/Core/Style.js
ftoromanoff: corrected. And i'm thinking that we should really add a setzoom properties instead of putting in in global.

src/Core/Style.js
ftoromanoff: corrected

src/Core/Style.js
jailln: you're right there is only one, then it's `vertex`

src/Core/Style.js
jailln: yeys it's better thanks

src/Core/Style.js
jailln: I agree, maybe in another PR so we can move on with the one already open ?

jailln (approved):

PR #2205, opened 2023-10-11
Add npm provenance to npm package

jailln:
## Description

Add [npm provenance](https://github.blog/changelog/2023-09-26-npm-provenance-general-availability/) to npm published packages. npm provenance is [only available as of npm cli 9.5.0+](https://docs.npmjs.com/generating-provenance-statements#prerequisites) so I also updated the deployment workflow node version from 16 to 18. 

Tested on my fork (with publication on my npm account).

## Motivation and Context

Use a newly introduced npm feature and explicitely link our npm releases to this repository and to the commit it was created from.

Closes #1963

## Screenshots (if appropriate)

image

PR #2208, opened 2023-10-12
chore(workflow): fix publish rights with npm provenance

jailln:
## Description
Add `content: write` permission to publish step

## Motivation and Context

Fixes publish task in integration workflow that was partly broken since adding  `id-token: write` for #2205


Desplandis (approved):
Tested. Fix previous PR issue.

PR #2220, opened 2023-11-03
Missing label support in doc

a-dutremble:
`LabelLayer` does not work when the features are Polygons and Lines, and this is not added in the doc.

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -231,7 +231,8 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
  * then the altitude value is set to 0.
  * @property {Object} [point.model] - 3D model to instantiate at each point position.
  *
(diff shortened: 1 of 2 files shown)


src/Core/Style.js
```diff
@@ -231,7 +231,8 @@ function defineStyleProperty(style, category, name, value, defaultValue) {
  * then the altitude value is set to 0.
  * @property {Object} [point.model] - 3D model to instantiate at each point position.
  *
- * @property {Object} [text] - All things {@link Label} related.
+ * @property {Object} [text] - All things {@link Label} related. (Isn't wet supported
+ * for Line and Polygone `Layer`.)
```
jailln: wet -> yet
I would say: Labels are currently only supported for points (not yet for line and polygons) 

(same modifications in LabelLayer)

jailln (approved):

PR #2222, opened 2023-11-07
Fix `C3DTilesLayer` and `C3DTFeature`

valentinMachado:
## Description
`C3DTFeature` reference the `THREE.Object3D` where they belong.
`computeWorldBox3` is now a method of `C3DTFeature` possible due to the latter modification
Some code clean.
`updateStyle` is call after dispacthing event `ON_TILE_CONTENT_LOADED`

## Motivation and Context
`C3DTFeature` reference the `THREE.Object3D` where they belong.
=> tileContent can have different child with feature in it which was not handle by the previous code
`updateStyle` is call after dispacthing event `ON_TILE_CONTENT_LOADED`
=> when listener of this event influences `C3DTilesLayer` style the user does not have to call an extra `updateStyle` of the layer

waiting this to be approved to rename commits name

### DIFF
--- a/src/Core/3DTiles/C3DTFeature.js
+++ b/src/Core/3DTiles/C3DTFeature.js
@@ -1,23 +1,32 @@
+// eslint-disable-next-line no-unused-vars
+import { Object3D, Box3 } from 'three';
+
 /**
  * C3DTFeature is a feature of a 3DTiles
  *
  * @class      C3DTFeature
- * @param {number} tileId - tile id
+ * @param {number} tileId - tileId
  * @param {number} batchId - batch id
  * @param {Array<{start:number,count:number}>} groups - groups in geometry.attributes matching batchId
  * @param {object} info - info in the batchTable
- * @param {object} [userData={}] - some userData
+ * @param {object} [userData] - some userData
+ * @param {Object3D} object3d - object3d in which feature is present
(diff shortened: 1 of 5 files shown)


jailln (approved):
Ok for me 👍 thanks !

valentinMachado:
Great ! I've made a single commit, and CI is working.

src/Core/3DTiles/C3DTFeature.js
```diff
@@ -1,23 +1,28 @@
...
  * @property {Array<{start:number,count:number}>} groups - groups in geometry.attributes matching batchId
  * @property {object} info - info in the batchTable
  * @property {object} [userData={}] - some userData
  */
 class C3DTFeature {
     #info;
-    constructor(tileId, batchId, groups, info, userData = {}) {
-        /** @type {number} */
-        this.tileId = tileId;
+    constructor(tileId, object3d, batchId, groups, info, userData = {}) {
```
jailln: I just realized that this class is exposed in the API. Can you handle the case when users use the old API, log a deprecation message and add a BREAKING CHANGE message in the body of your commit please ?

valentinMachado: Done

jailln (approved):

Issue #2219, opened 2023-10-30
[Bug - Proj Mesh3D] 


ftoromanoff:
Linked to the PR #2180 that changed the way base_alti is handled, we noticed a difference in the representation of object in the 3D environnement. 
Take the examples source_file_gpx_raster and source_file_gpx_3d that both use a gpx file to draw it on a itowns scene. The first one project it on the raster layer when the second one draw it as a mesh. We exept the 2 way to show very similar view.
Or (already before) that's not the case moreover the newly merge PR make it even more different (see example following)

On master before merge:
tilt: 45
![image](https://github.com/iTowns/itowns/assets/47628509/76e9cacd-1e78-4c3a-bf27-204cbe50ce3f)
tilt: 0 (Nadir)
![image](https://github.com/iTowns/itowns/assets/47628509/4bac496f-38bc-4863-967e-7401032f91fb)

On master after merge:
tilt: 45
![image](https://github.com/iTowns/itowns/assets/47628509/da07202b-93bf-402c-8c04-28433c6b5ed2)
tilt: 0 (Nadir)
![image](https://github.com/iTowns/itowns/assets/47628509/6ec647b7-9d14-433b-963a-2e4defae1466)

ftoromanoff:
I propose a fix in PR #2223. Might not be the best in regard of efficiency but we get back the same result as we had in master.

PR #2223, opened 2023-11-07
Fix base alti for mesh 3d

ftoromanoff:
Fix a regression appeared on #2180  (issue #2219)

### DIFF
--- a/src/Converter/Feature2Mesh.js
+++ b/src/Converter/Feature2Mesh.js
@@ -17,8 +17,8 @@ let style;
 const dim_ref = new THREE.Vector2();
 const dim = new THREE.Vector2();
 const normal = new THREE.Vector3();
-const base = new THREE.Vector3();
-const extrusion = new THREE.Vector3();
+const baseCoord = new THREE.Vector3();
+const topCoord = new THREE.Vector3();
 const inverseScale = new THREE.Vector3();
 const extent = new Extent('EPSG:4326', 0, 0, 0, 0);
 
@@ -41,15 +41,6 @@ class FeatureMesh extends THREE.Group {
         this.#collection = new THREE.Group().add(this.meshes);
         this.#collection.quaternion.copy(collection.quaternion);
         this.#collection.position.copy(collection.position);
-
-        if (collection.crs == 'EPSG:4978') {
-            normal.copy(collection.center.geodesicNormal);
-        } else {
-            normal.set(0, 0, 1);
-        }
-
-        normal.multiplyScalar(collection.center.z);
-        this.#collection.position.sub(normal);
 
         this.#collection.scale.copy(collection.scale);
         this.#collection.updateMatrix();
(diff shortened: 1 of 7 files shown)


src/Converter/Feature2Mesh.js
```diff
@@ -209,18 +200,24 @@ function featureToPoint(feature, options) {
...
             style.setContext(context);
             const { base_altitude, color, radius } = style.point;
-            coord.z = 0;
 
-            if (!pointMaterialSize.includes(radius)) {
-                pointMaterialSize.push(radius);
-            }
+            coord.copy(localCoord)
+                .applyMatrix4(context.collection.matrixWorld)
+                .as('EPSG:4326', coord);
```
jailln: Why do you always convert it to 4326, the crs should depend on the view, isn't it? (same for line and polygons)

src/Converter/Feature2Mesh.js
```diff
@@ -209,18 +200,24 @@ function featureToPoint(feature, options) {
...
 
             // populate vertices
-            base.copy(normal).multiplyScalar(base_altitude).add(coord).toArray(vertices, v);
+            base.copy(normal)
```
jailln: shouldn't it be coord.z - base_altitude? (same for line and polygons)

Also a comment on what is computed here and about what is `base` would ease code reading

src/Core/Style.js
```diff
@@ -17,7 +17,7 @@ const inv255 = 1 / 255;
 const canvas = (typeof document !== 'undefined') ? document.createElement('canvas') : {};
 
 function baseAltitudeDefault(properties, ctx) {
-    return ctx?.coordinates?.z || ctx?.collection?.center?.z || 0;
+    return ctx?.coordinates?.z || 0;
```
jailln, line 20: should we default it to the terrain elevation using demutils?

src/Converter/Feature2Mesh.js
ftoromanoff: Indeed this conversion to 4326 must only be done when we are in 4978 to convert a distance to the center of the Earth (z in 4978) to an altitude (z in 4326).

src/Converter/Feature2Mesh.js
ftoromanoff: 1) I think there is a ambiguity of the meaning of base here. base is not the initial altitude but the altitude we want the base of the geometry to be set.'
Coord being the original altitude (before the altitude calculation), the shift we want to apply is indeed 'base_altitude - coord.z'.
Maybe we can rename base_altitude fot a less ambiguous name as ground_altitude (even if it's not really the ground)

2) I change the comment, hopefully it's more clear.

src/Core/Style.js
ftoromanoff: What you suggest is, in case there is not elevation value available, we should set the altitude to the terrain level and then in not to zero ? like :
return ctx?.coordinates?.z || itowns.DEMUtils.getTerrainObjectAt(view.tileLayer, ctx.coordinates, 0) || 0;

In this case we need to have acces at some information we don't have currently. It might be possible to get it but I haven't thought baout about to do it.

src/Converter/Feature2Mesh.js
jailln: Thanks! Can you add this justification in a comment in the code so we remember why please ? :)

src/Converter/Feature2Mesh.js
jailln: ok thansk for the clarification

src/Core/Style.js
jailln: Yep that's the idea but the code would be `return ctx?.coordinates?.z || itowns.DEMUtils.getElevationValueAt(view.tileLayer, ctx.coordinates, 0) || 0;` 

We would need to have knowledge of the `view` here though... Moreover it can be costly (to evaluate?) and not what the user wants (even if I don't see any usecase where the user would not like to have its data projected on its terrain by default if no other base altitude is provided). It should not block this PR and we can continue the discussion elsewhere when it's ready to be merged, but meanwhile, @Desplandis any thoughts on this?

Desplandis: @jailln Since the goal of the previous PR was to have a proper separation of concerns in the `Style` module, I don't think this is a good idea to re-introduce a dependency to the `View`. I thought that the elevation was relative to the ellipsoid?

jailln (approved):

PR #2231, opened 2023-11-21
Fix `View.pickObjectAt()`

mathieuLivebardon:
## Description
`pickObjectAt` returns an empty array if parameter `where` is an empty array.

## Motivation and Context
`where` is a parameter of `pickObjectAt` to specify the layers to perform a pick.
The problem:  If `where` is an empty array, the function picks on all layers by default.

```js
const intersects = view.pickObjectsAt(
  event,
  0,
  view.getLayers().filter((el) => el.isC3DTilesLayer)
);
```
Here, if the application had dynamic layers¹ and no C3DTilesLayers in the `View`, `intersects` should be an empty array.

¹ *remove/add layers at runtime*

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -741,7 +741,7 @@ class View extends THREE.EventDispatcher {
     pickObjectsAt(mouseOrEvt, radius = 0, where) {
         const sources = [];
 
-        if (!where || where.length === 0) {
+        if (!where) {
             where = this.getLayers(l => l.isGeometryLayer);
         }
         if (!Array.isArray(where)) {


ketourneau:
I think your PR can cause some trouble for people who use the `pickObjectAt` and want method to use all layers if `where` condition is an empty array.

In your case, why not simply check empty array before call `pickObjectsAt` ?

```js
const layers = view.getLayers().filter((el) => el.isC3DTilesLayer);
let intersects = [];
if (layers.length > 0) {
  intersects = view.pickObjectsAt(event, 0, ayers);
}
```

jailln:
I agree with @ketourneau, and how it works currently seems more intuitive to me. Would the solution proposed by @ketourneau work in your context?

mathieuLivebardon:
Fine, I'll close this PR and use @ketourneau's solution. 
Thanks for your answers.

PR #2234, opened 2023-12-05
chore(deps): December 2023 bump

Desplandis:
## Description
Update all dependencies (when possible). I listed one **major** issue due to license change as well as *minor* issues questioning the necessity of certain dependencies or packaging problems upstream.

## Peer dependencies
@mgermerie @jailln Since peer dependencies tie users of the library to those versions, shall we tag the commits as features?

- [x] `three` (`0.154.0` to `0.159.0`): Breaking changes given by the [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide#r158--r159). Possible issues:
  + [x] [Updates to lighting in three.js r155](https://discourse.threejs.org/t/updates-to-lighting-in-three-js-r155/53733).
  + [x] `AnimationUtils.arraySlice()` has been removed in `r157`. Used by `LegacyGLTFLoader`.
  + [x] `Quaternions` are now expected to be normalized in `r158`.
- [x] `proj4` (`2.9.0` to `2.9.2`): No breaking changes.

## Dependencies
Those are the packages with a new version available:

- [x] `@loaders.gl/las` (`3.4.4` to `4.0.4`): No API changes.
- [x] `@mapbox/mapbox-gl-style-spec` (`13.28.0`): Will not update due to the fact this is the latest version with a FOSS license (`ISC`). This is a huge **red-flag**, shall we consider moving to `@maplibre/maplibre-gl-style-spec`? What are the differences between those two?
- [x] `@tmcw/togeojson` (`5.6.2` to `5.8.1`): Follows semantic versioning, no breaking changes.
- [x] `@tweenjs/tween.js` (`18.6.4`): Cannot update to `21.0.0` due to error `ERR_REQUIRE_ESM`. They distribute a commonJS module but due to its `.js` extension and their `package.json` treating js files as ES modules causing this issue... This could be fixed on their side by changing the extension of their commonJS module to `.cjs`. Does not seems to be addressed upstream yet (see tweenjs/tween.js#649 and tweenjs/tween.js#659).
- [x] `regenerator-runtime` (`0.13.11` to `0.14.0`): No breaking changes. However all browsers support [async functions][async]/[await statements][await] since April 2017 (no IE11), do we still need this dependency (would need to update our babel target)?

Those packages may not be needed anymore:

- `text-encoding-utf-8`: Used to polyfill [TextDecoder][TextDecoder] in our code-base and is natively supported by all major browsers and Node.js since 2017 (with the exception of Edge which added support in 2020). Do we still need this dependency? Note that this can break our test-case since Node.js imports this differently (in `util` module).

## Developer dependencies
Those packages upgrades introduce no breaking change:
- [x] `chalk`.
- [x] `chart.js`.
- [x] `core-js`.
- [x] `https-proxy-agent`.
- [x] `marked`
- [x] `node-fetch`: Update to latest `2.x.x` version. Cannot update to `3.x.x` since it's now a ESM only package but the 2.x.x branch is still maintained.
- [x] `replace-in-file`.
- [x] `typescript`.
- [x] `webpack` and the following loader: `babel-loader`.
- [x] `whatwg-fetch`: Used to polyfill the `fetch` API in web environment (supported by all browsers > 2016) in our webpack bundle. This may not be necessary anymore.

Those packages upgrades introduce a breaking change for our use-cases:
- [x] `conventional-changelog-cli`: Now requires a `node >= 16` environment. Fix issue when the changelog of all versions was inserted in the changelog.md (only the last release is now added).
- [x] `eslint` and the following plug-ins: `eslint-import-resolver-webpack` and `eslint-plugin-import`. Note that the upgrade introduce a change in the [`if-else` indent rule][eslint:fixifelse], causing a linting issue in `Ellipsoid`.
- [x] `puppeteer`: Use new headless mode using chrome (instead of chromium).

This package introduces coverage regression and will not be updated:
- `@babel/cli`, `@babel/register` and the following plug-ins: `@babel/plugin-transform-runtime` and `@babel/preset-env`. Cause regression in coverage.

## Notes to reviewer
I didn't had access to pre-1.0 3D Tiles files to test lighting issues and edits to `LegacyGLTFLoader`. Maybe @jailln you could test it on your proprietary datasets ?

[async]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function
[await]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await
[eslint:releases]: https://github.com/eslint/eslint/releases
[eslint:fixifelse]: https://github.com/eslint/eslint/pull/17318
[TextDecoder]: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder


mgermerie:
> @mgermerie @jailln Since peer dependencies tie users of the library to those versions, shall we tag the commits as features?

I think this is a good idea. Bumping them sometimes grants users with new features introduced in their new versions. On another point, I often pay more attention to the "feature" section of a changelog than the "workflows and chores" (or equivalent name). So based on my own experience, it would make peer dependencies update more visible. However this has yet to be confirmed by other user experiences.

Desplandis:
@jailln @mgermerie Ready for review, I added a "Notes to reviewer" to my initial message. Unfortunately, I dont't have access to 3D Tiles with old gltfs (version = 1.0).

mgermerie:
I also think we could remove `regenerator-runtime` and `text-encoding-utf-8` - perhaps in a future PR ?

jailln (approved):
Thanks for the bump and the digest :)

* Regarding `@mapbox/mapbox-gl-style-spec`, can you open a dedicated issue please?
* I agree that we should remove `regenerator-runtime`, `text-encoding-utf-8` and `whatwg-fetch` especially since we will update our babel target soon.
* I tested with 3D tiles pre-1.0 and with gltf 1.0, everything works fine

Desplandis:
Opened an issue for mapbox and a PR for removing unecessary dependencies.
Will now merge this PR.

PR #2242, opened 2023-12-18
Fix pointcloud node subdivision for orthographic projections

### LINKED ISSUE #2232: Orthographic Camera & Point Cloud - Point cloud not updating
Orthographic camera usage on itowns makes any point cloud added in the scene not updating.

## Your Environment
* Version used: itowns 2.41.1.9
* Browser Name and version: chrome, firefox
* Operating System and version (desktop or mobile): desktop
* Link to your project:
Exemple here : 
https://65674e4ca6a4df257c21c636--demo-hackathon.netlify.app/examples/#misc_orthographic_camera
Code here : 
https://github.com/bloc-in-bloc/itowns/blob/master/examples/misc_orthographic_camera.html

## Context
Point cloud layer is not updating if we use orthographic camera (perspective camera work). Don't know if anyone has this use case before.
We think it's because of the orthographic camera position which never moves but only "zoom" so point cloud layer has the same position of camera whatever the orthographic camera zoom has.

Check exemples, zoom on point cloud with orthographic camera view

## Expected Behavior
Point cloud layer should update based on zoom.

## Actual Behavior
Point cloud layer does not update based on zoom.

We think it's because of the orthographic camera position which never moves but only "zoom" so point cloud layer has the same position of camera whatever the orthographic camera zoom has.

Desplandis:
## Description
For now, point cloud layer's screen-space error (SSE) calculation is only valid for perspective cameras. This causes issues when using an orthographic camera (e.g. #2232: point cloud not updating).

This PR fixes two issues:
- do not compute point attenuation when using an orthographic projection
- add a new way to compute SSE when using an orthographic projection

## Related issues

closes #2232

### DIFF
--- a/src/Layer/PointCloudLayer.js
+++ b/src/Layer/PointCloudLayer.js
@@ -26,7 +26,7 @@ function initBoundingBox(elt, layer) {
     elt.obj.boxHelper.updateMatrixWorld();
 }
 
-function computeScreenSpaceError(context, pointSize, spacing, elt, distance) {
+function computeSSEPerspective(context, pointSize, spacing, elt, distance) {
     if (distance <= 0) {
         return Infinity;
     }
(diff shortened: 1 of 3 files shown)


jailln (approved):

Desplandis:
@Pourfex Does rebasing/cherry-picking on this PR fix your issue?

Pourfex:
> @Pourfex Does rebasing/cherry-picking on this PR fix your issue?

Perfect ! Work like a charm. Thanks for your quick resolve

PR #2243, opened 2023-12-19
Feature/add bbox url precision

ketourneau:
## Description
When we try to use a custom WMS source (omniscale provider) we got some mistake if the coordinate in BBOX url is rounded.

## Motivation and Context
We would like to set custom precision of BBOX passed in url to solve our problem.
In our case if bbox precision is `9` it's work.

```
sourceMap = new itowns.WMSSource({
    url: `https://maps.omniscale.net/v2/${omniscaleKey}/style.default/map`,
    name: "osm",
    format: "image/png",
    crs: "EPSG:2154",
    extent: extent,
    version: '1.1.0',
    bboxUrlPrecision: 9
});
```

## Screenshots
`bboxUrlPrecision: 2`
Capture d’écran 2023-12-19 à 16 34 09

`bboxUrlPrecision: 9`
Capture d’écran 2023-12-19 à 16 33 44

### DIFF
--- a/src/Provider/URLBuilder.js
+++ b/src/Provider/URLBuilder.js
@@ -89,12 +89,19 @@ export default {
      * // http://server.geo/wms/BBOX=12,35,14,46&FORMAT=jpg&SERVICE=WMS
      *
      * @param {Extent} bbox - the bounding box
-     * @param {Source} source
+     * @param {Object} source
+     * @param {string} source.crs
+     * @param {number} source.bboxDigits
+     * @param {string} source.url
+     * @param {string} source.axisOrder
      *
      * @return {string} the formed url
      */
     bbox: function bbox(bbox, source) {
-        const precision = ((source.bboxUrlPrecision === undefined) && (source.crs == 'EPSG:4326')) ? 9 : source.bboxUrlPrecision;
+        let precision = source.crs == 'EPSG:4326' ? 9 : 2;
+        if (source.bboxDigits !== undefined) {
+            precision = source.bboxDigits;
+        }
         bbox.as(source.crs, extent);
         const w = extent.west.toFixed(precision);
         const s = extent.south.toFixed(precision);
--- a/src/Provider/URLBuilder.js
+++ b/src/Provider/URLBuilder.js
@@ -94,7 +94,7 @@ export default {
      * @return {string} the formed url
      */
     bbox: function bbox(bbox, source) {
-        const precision = source.crs == 'EPSG:4326' ? 9 : 2;
+        const precision = ((source.bboxUrlPrecision === undefined) && (source.crs == 'EPSG:4326')) ? 9 : source.bboxUrlPrecision;
         bbox.as(source.crs, extent);
         const w = extent.west.toFixed(precision);
         const s = extent.south.toFixed(precision);
--- a/src/Source/Source.js
+++ b/src/Source/Source.js
@@ -104,6 +104,7 @@ let uid = 0;
  * @property {string} crs - The crs projection of the resources.
  * @property {string} attribution - The intellectual property rights for the
  * resources.
+ * @property {string} bboxUrlPrecision - The bbox decimal precision used in URL
  * @property {Extent} extent - The extent of the resources.
  * @property {function} parser - The method used to parse the resources attached
  * to the layer. iTowns provides some parsers, visible in the `Parser/` folder.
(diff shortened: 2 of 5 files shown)


src/Provider/URLBuilder.js
```diff
@@ -94,7 +94,7 @@ export default {
      * @return {string} the formed url
      */
     bbox: function bbox(bbox, source) {
-        const precision = source.crs == 'EPSG:4326' ? 9 : 2;
+        const precision = ((source.bboxUrlPrecision === undefined) && (source.crs == 'EPSG:4326')) ? 9 : source.bboxUrlPrecision;
```
Desplandis: In this line, precision could be undefined if bboxUrlPrecision == undefined && source.crs !== 'EPSG:4326'.
Could you treat this case?

src/Source/Source.js
```diff
@@ -104,6 +104,7 @@ let uid = 0;
  * @property {string} crs - The crs projection of the resources.
  * @property {string} attribution - The intellectual property rights for the
  * resources.
+ * @property {string} bboxUrlPrecision - The bbox decimal precision used in URL
```
Desplandis: Since this property is only used on `WMSSource` and `WFSSource`, I think we should not pollute their parent class `Source` and only add this property to those two classes.

src/Provider/URLBuilder.js
```diff
@@ -94,7 +94,7 @@ export default {
      * @return {string} the formed url
```
Desplandis, line 98: This is a bit out of scope of this PR, but the type parameters documentation are not accurate...
You could (optionally) replace it by:
```js
     * @param {Object} source
     * @param {string} source.crs
     * @param {number} source.bboxUrlPrecision
     * @param {string} source.url
     * @param {string} source.axisOrdre
```

Desplandis, line 98: I would prefer naming the new parameter as `bboxPrecision` or `bboxDigits` to conform to the naming of the parameter of function [`toFixed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed).

src/Source/Source.js
```diff
@@ -145,6 +146,7 @@ class Source extends InformationsData {
         this.isVectorSource = (source.parser || supportedParsers.get(source.format)) != undefined;
         this.networkOptions = source.networkOptions || { crossOrigin: 'anonymous' };
         this.attribution = source.attribution;
+        this.bboxUrlPrecision = source.bboxUrlPrecision === undefined ? 2 : source.bboxUrlPrecision;
```
Desplandis: Following previous comments, you should move this assignment to `WMSSource` and `WFSSource` as follow:
```js
this.bboxUrlPrecision = source.bboxUrlPrecision;
```

Desplandis (changes requested):
Hi and thanks for your PR! =)
It seems that our defaults are overfitting on IGN datasets, we may have to change them later (don't hesitate to open an issue if the defaults are causing issues with other datasets).

src/Provider/URLBuilder.js
ketourneau: Indeed, I hadn't thought of that.
I think the code is more readable this way : 
```
let precision = source.crs == 'EPSG:4326' ? 9 : 2;
if (source.bboxDigits !== undefined) {
    precision = source.bboxDigits;
}
```

src/Source/Source.js
ketourneau: you're right
I moved code to WMSource and WFSource.

src/Provider/URLBuilder.js
ketourneau: Ok I added it

ketourneau: Ok I rename to `bboxDigits`

src/Source/Source.js
ketourneau: Done, I move code to WMSource and WFSSource

ketourneau:
For the time we don't have any overfitting with IGN ortho data, only with omniscale Open Street Map.
I applied your code review, thanks for the review.

ketourneau (commented):
I made the changes

Desplandis:
Sounds good to me!
I saw the implementation of both mapbox and openlayers and they use numbers directly without truncating them, it may have been a fix for some IGN datasets, I will look into it!

Desplandis (approved):

PR #2246, opened 2023-12-20
fix(examples): change watercolor tile url in view_2d_map

### LINKED ISSUE #2227: [BUG - API Tiles] Examples DOWN because of unreachable hosts
Some current example on iTowns website are down because the tile host failed to answer.

## Context
For the example `view_2d_map`:
It seems like c.tile.stamen.com is unreachable

For other examples (`3dtiles_basic`):
It seems like osm.oslandia.io is unreachable

Try the examples on https://www.itowns-project.org/itowns/examples/index.html#3dtiles_basic

## Actual Behavior
Screenshot from 2023-11-10 10-50-55
Screenshot from 2023-11-10 10-51-23

Change our examples to request different hosts
To replace osm.oslandia.io, we could use the official OSM api: https://a.tile.openstreetmap.org.
It's already used in source_file_geojson_raster_2.html
But should be careful that the following rules are followed:
https://operations.osmfoundation.org/policies/tiles/

AnthonyGlt:
https://github.com/iTowns/itowns/issues/2227

[Related comment](https://github.com/iTowns/itowns/issues/2227#issuecomment-1864711230)

### DIFF
--- a/examples/view_2d_map.html
+++ b/examples/view_2d_map.html
@@ -47,7 +47,7 @@
             var opensmSource = new itowns.TMSSource({
                 isInverted: true,
                 // eslint-disable-next-line no-template-curly-in-string
-                url: 'http://c.tile.stamen.com/watercolor/${z}/${x}/${y}.jpg',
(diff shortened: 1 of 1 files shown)


Desplandis (approved):
Tested, it fixes #2227.

PR #2248, opened 2023-12-22
[Deps] Remove unecessary dependencies

Desplandis:
## Description
Follow-up of #2252 (and proposal #2256) which updates and formalizes the web browser requirements of iTowns.

## Motivation and Context
Following the target update in #2252, the following polyfills are not necessary anymore:
- `TextEncoder` polyfill
- `async/await` polyfill
- babel's null coalescing support plugin

Moreover, the following polyfills on webpack's generated bundle are not needed anymore:
- `fetch` polyfill
- `URL` polyfill

This list may not be exhaustive and I may have forgotten other polyfills.

### DIFF
--- a/src/Utils/Utf8Decoder.js
+++ b/src/Utils/Utf8Decoder.js
@@ -1,7 +1,0 @@
-import {
-    TextDecoder as TextDecoderPolyfill,
-} from 'text-encoding-utf-8';
-
-export const TextDecoder = typeof global.TextDecoder === 'function' ? global.TextDecoder : TextDecoderPolyfill;
-
-export default new TextDecoder('utf-8');
--- a/.babelrc
+++ b/.babelrc
@@ -18,7 +18,9 @@
             ]
         }],
         ["@babel/plugin-proposal-nullish-coalescing-operator"],
-        ["@babel/plugin-transform-runtime"],
+        ["@babel/plugin-transform-runtime", {
+            "regenerator": false
+        }],
         ["minify-replace", {
             "replacements": [{
                 "identifierName": "__DEBUG__",
--- a/.babelrc
+++ b/.babelrc
@@ -17,7 +17,6 @@
                 ".css"
             ]
         }],
-        ["@babel/plugin-proposal-nullish-coalescing-operator"],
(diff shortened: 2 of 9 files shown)


src/Utils/Utf8Decoder.js
```diff
@@ -1,7 +1 @@
-import {
-    TextDecoder as TextDecoderPolyfill,
-} from 'text-encoding-utf-8';
-
-export const TextDecoder = typeof global.TextDecoder === 'function' ? global.TextDecoder : TextDecoderPolyfill;
-
-export default new TextDecoder('utf-8');
+export default new TextDecoder();
```
jailln: Why not removing `Utf8Decoder` and replacing it by `TextDecoder` directly where it is used?

test/unit/bootstrap.js
```diff
@@ -17,7 +17,10 @@ global.URL = function URL(url) {
     };
 };
 
-global.Event = () => {};
+// https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-instanceofoperator
+// ES standard requires that left-hand side shall have a prototype method, this
+// is not the case for lambdas.
```
jailln: so you should update `global.requestAnimationFrame` too?

Desplandis: This issue arises only for check that some function (here `Event`) is an `instanceof` of some types (here `Event`). There is no such use of this operator for `requestAnimationFrame`.

src/Utils/Utf8Decoder.js
Desplandis: I did it initially to have a single shared instance (as previously done) of `TextDecoder` but I could instantiate it on `B3DMParser` & co. WDYT?

jailln: Yes that's what I figured but I think it would be simpler to instantiate it where it's needed (without having a big impact on perfs nor memory)

test/unit/bootstrap.js
jailln: ok :+1:

jailln (approved):

PR #2255, opened 2024-01-09
fix(tests): prevent overwriting `navigator.userAgent`

mgermerie:
## Before contributing

- [x] I have read [CONTRIBUTING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) and [CODING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) to apply `iTowns` conventions on PRs, Git history and code.

## Description

As discussed in #2254, the `navigator.userAgent` has been made read-only in [node 21.1.0](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V21.md#2023-10-24-version-2110-current-targos) (see [this commit](https://github.com/nodejs/node/commit/5a52c518ef)).

In this PR, we check if navigator is already set (allegedly by node). If that's the case, we don't overwrite it.

### DIFF
--- a/test/unit/bootstrap.js
+++ b/test/unit/bootstrap.js
@@ -23,9 +23,13 @@ global.fetch = fetch;
 global.fetch.Promise = Promise;
 
 // this could be replaced by jsdom.Navigator in https://github.com/iTowns/itowns/pull/1412
-global.navigator = {
-    userAgent: 'firefox',
-};
+// Checking if global.navigator exists targets node versions <21.1.0. Since node
+// <21.1.0, global.navigator is read-only.
+if (!global.navigator) {
+    global.navigator = {
+        userAgent: 'firefox',
+    };
+}
 
 class DOMElement {
     constructor() {


test/unit/bootstrap.js
```diff
@@ -23,9 +23,11 @@ global.fetch = fetch;
 global.fetch.Promise = Promise;
 
 // this could be replaced by jsdom.Navigator in https://github.com/iTowns/itowns/pull/1412
-global.navigator = {
-    userAgent: 'firefox',
-};
+if (!global.navigator) {
```
Desplandis, line 28: Could you add a notice that such code targets `node < 21.0.0`? This could ease its removal once our minimum node targets will be `>= 21.0.0`. Something like:
```suggestion
// Partial support for global.navigator was added in node 21.0.0
if (!global.navigator) {
```

Desplandis (changes requested):
Tested on node `21.0.0` and latest LTS (`20.11.0`), works like a charm!
Could you please add a notice (see comment)?

test/unit/bootstrap.js
mgermerie: Done

Desplandis (approved):

PR #2260, opened 2024-01-15
Change implementation of the .las parser

ftoromanoff:
With the implementation of the gestion of copc, we need to integretate a parsing methode for las v1.4 or curretnly the las Parser used (loaders.gl/las) only follow the specification of las v1.2, and is not going to evolve in a near future.
The copc.js module (that will be really handy for the copc gestion) on the other hand manage the las v1.4 and bellow.

This PR aims to change the las parser from the loader.gl library to the copc library.


src/Parser/LASLoader.js
```diff
@@ -0,0 +1,144 @@
...
+ * "extra bytes". Those are described by an Extra Bytes VLR.
+ * @property {number[]} header.scale - Scale factors (an array `[xScale,
+ * yScale, zScale]`) multiplied to the X, Y, Z point record values.
+ * @property {number[]} header.offset - Offsets (an array `[xOffset,
+ * xOffset, zOffset]`) added to the scaled X, Y, Z point record values.
+ */
+
+class LASLoader {
+    constructor() {
+        this._wasmPath = 'https://unpkg.com/laz-perf@0.0.6/lib/';
```
jailln, line 29: We should maybe indicate in the class documentation that by default it gets the wasm loader from a cdn (with the cdn url) and that it can be set to a local path with lazPerf setter

src/Parser/LASLoader.js
```diff
@@ -0,0 +1,144 @@
...
+ * point data records. If the specified size is larger than implied by the
+ * point data record format (see above) the remaining bytes are user-specfic
+ * "extra bytes". Those are described by an Extra Bytes VLR.
+ * @property {number[]} header.scale - Scale factors (an array `[xScale,
+ * yScale, zScale]`) multiplied to the X, Y, Z point record values.
+ * @property {number[]} header.offset - Offsets (an array `[xOffset,
+ * xOffset, zOffset]`) added to the scaled X, Y, Z point record values.
+ */
+
+class LASLoader {
```
jailln, line 27: Does it works for las and laz seamlessly ? If yes, indicate it in the class doc and modify comment of `parse` method

src/Parser/LASParser.js
```diff
@@ -1,66 +1,76 @@
...
  * @module LASParser
  */
 export default {
+    /*
+     * Set the laz-perf decoder path.
+     * @param {string} path - path to `laz-perf.wasm` folder.
+     */
+    enableLazPerf(path) {
+        if (!path) {
+            throw new Error('Path to laz-perf is mandatory');
```
jailln, line 21: Is it? There is a default in LasLoader. Should we remove the default and indicate to the user that one can use a cdn ?

Desplandis: We followed the naming convention of the `enable*` family of functions of 3DTiles parsers. Since removing the default cause a breaking changes for users of `EntwinePointTileLayer`, I advocate for keeping a default CDN.

src/Parser/LASLoader.js
Desplandis: Done.

src/Parser/LASLoader.js
Desplandis: Done.

src/Parser/LASParser.js
Desplandis: + the `Error` is a sort of a type guard against `undefined` values.

jailln: fine by me

jailln (approved):

PR #2264, opened 2024-01-24
Webpack: Stop watching node modules in dev mode

jailln:
## Description
Prevent webpack from watching for node modules changes in dev mode (inspired from [the official doc](https://webpack.js.org/configuration/watch/#watchoptionsignored))

## Motivation and Context
I sometime get the following error when running `npm start`: `Error: ENOSPC: System limit for number of file watchers reached, especially when I have other projects running with webpack at the same time. I found at that webpack is by default also looking for changes in `node_modules/` which is in 99.9% of cases not useful and increases greatly the number of files to watch.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -99,7 +99,7 @@ module.exports = () => {
             static: {
                 directory: path.resolve(__dirname, './'),
                 watch: {
-                    ignored: path.resolve(__dirname, '.git'),
+                    ignored: [path.resolve(__dirname, '.git'), path.resolve(__dirname, 'node_modules')],
                 },
             },
             client: {


Desplandis:
I stumbled upon a similar issue when experimenting with the generation of an ES bundle with webpack for our examples. I'm interested in how you did encounter such problem. x)

I agree with the fix, however when linking to a local dependency (e.g. when using `yarn link`), a rare use-case I concur, it's nice for webpack to watch for edits on this dependency. @mgermerie and I are currently doing this for our `loaders.gl` experimentation.

jailln:
I stumbled upon it when running another app with webpack at the same time. 

Ok I understand the usecase, but as you said it's quite rare. Should we still merge it and one can remove this ignored path when needed for specific use cases? WDYT? We could also think of something more elaborated (like configure it with options in the `npm start` command or such but I don't think it's worth it)

Desplandis:
Erratum, it seems to not affect @mgermerie use-case. Nonetheless, I think we should merge it, this is a niche use-case and this PR could benefit to us all.

Desplandis (approved):
Tested and reviewed!

PR #2266, opened 2024-02-08
3D tiles: fix a bug and improve perfs

### LINKED ISSUE #2074: BatchID in 3DTiles are float and negative
Some batch id in 3DTiles of:

* http://localhost:8080/examples/#3dtiles_25d
* http://localhost:8080/examples/#3dtiles_ion

are float and even negative which does not match [3DTiles specs](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Batched3DModel)

## Your Environment
Google chrome Version 112.0.5615.49 (Build officiel) (64 bits), Ubuntu

Have spot the bug on a fork branch [here](https://github.com/VCityTeam/Itowns/tree/bug-float-batchid). Pull then npm start to the examples I said above.

I have spot this bug in the onTileContentLoaded of the C3DTilesLayer, don't know if this is code before doing this or if this is the 3DTiles used.

jailln:
## Description
This PR introduces two changes:

* Fixes #2074 (that also caused tiles to be removed and never loaded again when removed from the cache): the issue came from accessing threejs buffer attributes array directly instead of using the recommended `getX()` method (which is not the same in the case of [InterleavedBufferAttribute](https://threejs.org/docs/index.html?q=interleaved#api/en/core/InterleavedBufferAttribute).

* Greatly improves 3D tiles perfs (see screenshots below) by reducing loading time overhead due to internal structures pre-filling (mainly for 3D tiles picking and style) -> part of these structures are now filled on first demand.

## Screenshots

Performance profiling on `3dtiles_ion.html` example before the second commit:

before-PR-getInfoById

Performance profiling on `3dtiles_ion.html` example after the second commit:

afterPR-getInfoById

### DIFF
--- a/src/Core/3DTiles/C3DTFeature.js
+++ b/src/Core/3DTiles/C3DTFeature.js
@@ -2,29 +2,41 @@
 import { Object3D, Box3 } from 'three';
 
 /**
+ * Finds the batch table of an object in a 3D Tiles layer. This is
+ * for instance needed when picking because we pick the geometric
+ * object which is not at the same level in the layer structure as
+ * the batch table.
+ * @param {THREE.Object3D} object - a 3D geometric object
+ * @returns {C3DTBatchTable|null} - the batch table of the object
+ */
+function findBatchTable(object) {
+    if (object.batchTable) {
+        return object.batchTable;
+    }
+    if (object.parent) {
+        return findBatchTable(object.parent);
+    }
+    return null;
+}
+
+/**
  * C3DTFeature is a feature of a 3DTiles
  *
  * @class      C3DTFeature
  * @param {number} tileId - tileId
  * @param {number} batchId - batch id
  * @param {Array<{start:number,count:number}>} groups - groups in geometry.attributes matching batchId
- * @param {object} info - info in the batchTable
  * @param {object} [userData] - some userData
  * @param {Object3D} object3d - object3d in which feature is present
  * @property {number} tileId - tile id
  * @property {Object3D} object3d - object3d in which feature is present
  * @property {number} batchId - batch id
  * @property {Array<{start:number,count:number}>} groups - groups in geometry.attributes matching batchId
- * @property {object} info - info in the batchTable
  * @property {object} [userData] - some userData
  */
 class C3DTFeature {
     #info;
-    constructor(tileId, batchId, groups, info, userData, object3d) {
-        if (!object3d) {
-            console.error('BREAKING CHANGE: C3DTFeature constructor changed from (tileId, batchId, groups, info, userData) to (tileId, batchId, groups, info, userData, object3d)');
-        }
-
+    constructor(tileId, batchId, groups, userData, object3d) {
         /** @type {Object3D} */
         this.object3d = object3d;
 
@@ -37,11 +49,11 @@ class C3DTFeature {
         /** @type {object} */
         this.userData = userData;
 
-        /** @type {object} */
-        this.#info = info;
-
         /** @type {number} */
         this.tileId = tileId;
+
+        // Lazy-loaded batch table information for this.batchId.
+        this.#info = null;
     }
 
     /**
@@ -84,10 +96,19 @@ class C3DTFeature {
     }
 
     /**
-     *
+     * Gets the information from the tile batch table for this C3DTFeature batch id.
      * @returns {object} - batchTable info
      */
     getInfo() {
+        if (this.#info) {
+            return this.#info;
+        }
+        const batchTable = findBatchTable(this.object3d);
+        if (!batchTable) {
+            console.warn(`[C3DTFeature]: No batch table found for tile ${this.tileId}.`);
+            return null; // or return undefined;
+        }
+        this.#info = batchTable.getInfoById(this.batchId);
         return this.#info;
     }
 }
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -308,8 +308,8 @@ class C3DTilesLayer extends GeometryLayer {
                     throw new Error('no batchTable');
                 }
 
-                const geometryAttributes = child.geometry.attributes;
-                let currentBatchId = geometryAttributes._BATCHID.array[0];
+                const batchIdAttribute = child.geometry.getAttribute('_BATCHID');
+                let currentBatchId = batchIdAttribute.getX(0);
                 let start = 0;
                 let count = 0;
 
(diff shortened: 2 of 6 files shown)


src/Core/3DTiles/utils/3DTilesUtils.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Finds the batch table of an object in a 3D Tiles layer. This is
+ * for instance needed when picking because we pick the geometric
+ * object which is not at the same level in the layer structure as
+ * the batch table. More details here on itowns internal
+ * organization of 3DTiles:
+ *  https://github.com/MEPP-team/RICT/blob/master/Doc/iTowns/Doc.md#itowns-internal-organisation-of-3d-tiles-data
+ * @param {THREE.Object3D} object - a 3D geometric object
+ * @returns {C3DTBatchTable} - the batch table of the object
```
Desplandis:
```suggestion
 * @returns {C3DTBatchTable|null} - the batch table of the object
```

src/Core/3DTiles/C3DTFeature.js
```diff
@@ -1,5 +1,6 @@
 // eslint-disable-next-line no-unused-vars
 import { Object3D, Box3 } from 'three';
+import findBatchTable from './utils/3DTilesUtils';
```
Desplandis: `findBatchTable` was moved from `C3DTilesLayer` to its own file. However, it is only used once and not exported to the end-user. Shall we move it here instead?

src/Core/3DTiles/C3DTFeature.js
```diff
@@ -37,11 +36,12 @@ class C3DTFeature {
         /** @type {object} */
         this.userData = userData;
 
-        /** @type {object} */
-        this.#info = info;
-
         /** @type {number} */
         this.tileId = tileId;
+
+        // Holds the information from the batch table for this.batchId. Will be filled the first time this information
```
Desplandis: I usually use the term "lazy-loaded" x)

src/Core/3DTiles/C3DTFeature.js
```diff
@@ -8,19 +9,17 @@ import { Object3D, Box3 } from 'three';
...
  * @property {Object3D} object3d - object3d in which feature is present
  * @property {number} batchId - batch id
  * @property {Array<{start:number,count:number}>} groups - groups in geometry.attributes matching batchId
- * @property {object} info - info in the batchTable
  * @property {object} [userData] - some userData
  */
 class C3DTFeature {
     #info;
-    constructor(tileId, batchId, groups, info, userData, object3d) {
+    constructor(tileId, batchId, groups, userData, object3d) {
```
Desplandis, line 39: This is some breaking change since `C3DTFeature` is exposed to the end-user.
However, is there anyone which instantiates `C3DTFeature` directly? Why it is exposed then?

src/Layer/C3DTilesLayer.js
```diff
@@ -225,25 +225,6 @@ class C3DTilesLayer extends GeometryLayer {
...
-     * Finds the batch table of an object in a 3D Tiles layer. This is
-     * for instance needed when picking because we pick the geometric
-     * object which is not at the same level in the layer structure as
-     * the batch table. More details here on itowns internal
-     * organization of 3DTiles:
-     *  https://github.com/MEPP-team/RICT/blob/master/Doc/iTowns/Doc.md#itowns-internal-organisation-of-3d-tiles-data
-     * @param {THREE.Object3D} object - a 3D geometric object
-     * @returns {C3DTBatchTable} - the batch table of the object
-     */
-    findBatchTable(object) {
```
Desplandis, line 238: Reason to not expose this method anymore (i.e. remove it)?

src/Layer/C3DTilesLayer.js
```diff
@@ -328,17 +304,18 @@ class C3DTilesLayer extends GeometryLayer {
...
                     }
                 };
 
-                for (let index = 0; index < geometryAttributes.position.array.length; index += geometryAttributes.position.itemSize) {
-                    const batchIndex = index / geometryAttributes.position.itemSize;
-                    const batchId = geometryAttributes._BATCHID.array[batchIndex];
+                const positionAttribute = child.geometry.getAttribute('position');
+                const positionAttributeSize = positionAttribute.count * positionAttribute.itemSize;
+                for (let index = 0; index < positionAttributeSize; index += positionAttribute.itemSize) {
+                    const batchIndex = index / positionAttribute.itemSize;
```
Desplandis, line 319: Why are you incrementing by `posititonAttribute.itemSize` only to divide the `index` by `positionAttribute.item` in the following line?

Could you not increment it by 1 and stop the iteration at `positionAttributeSize.count`?

Desplandis (changes requested):
Nice PR! I added some comments about a few odd things but they are minor. =)

Nice perf gain BTW.

src/Core/3DTiles/C3DTFeature.js
jailln: Yes it is but I had the same interrogations and I think the answers are nobody instantiates it directly and it shouldn't be exposed (I looked into LIRIS's code which are the main users of this and haven't seen any instantiation of `C3DTFeature`). We should remove it from the public API but I guess this is the purpose of another PR. Meanwhile I think we should accept having this breaking change and document it in the commit (and therefore the release), but no strong opinions on this, we can also do it differently. WDYT ?

src/Layer/C3DTilesLayer.js
jailln: Mmmh I went a bit too fast I think :grin: However I think it should not have been exposed in the first place (no real use outside of itowns). I can either:
* leave it like this and add a BREAKING CHANGE message
* reintroduce `3DTilesUtils` (with `findBatchTable` in it) and export it (still breaking change but with a fallback elsewhere in the API)
* add it back here and make it call `3DTilesUtils.findBatchTable()`

I'd rather do the first one, but WDYT?

src/Layer/C3DTilesLayer.js
jailln: I kept the existing logic but you're right it could be simplified (also `count` becomes useluss in this case). However, as the `C3DTilesStyle` widget is broken (#2267) I'd rather not do this in this PR because I cannot test my modifications.

src/Core/3DTiles/C3DTFeature.js
Desplandis: Agreed with the breaking change! We should keep C3DTFeature exposed for now, it may be the object of another PR.

src/Layer/C3DTilesLayer.js
Desplandis: I think we'll go for the first option.

src/Layer/C3DTilesLayer.js
Desplandis: We are missing test cases here... Could you add a notice here and fix this code when #2267 is fixed?

jailln: Yep agreed... Done, I also added a notice in #2267

Desplandis (approved):
Previous comments were all resolved. Ready to merge!

PR #2269, opened 2024-02-13
fix(examples): migrate ign and grandlyon urls

jailln:
## Description
Migrate examples urls:
* IGN urls from geoportail to geoplateforme
* some deprecated grand lyon urls.

## Motivation and Context
Fixes examples and tests.


Desplandis (approved):
Thanks for your fixes!

PR #2278, opened 2024-02-23
Fix cog

ketourneau:
## Motivation and Context

By testing many different Geotiff we see some memory & performance problem.
Only the highest level of detail from Geotiff was used no matter the zoom level.
No resampling was used (a tile must respect the tileWidth and tileHeight size from GeoTiff).


Desplandis (approved):
Ready to merge!

Issue #2244, opened 2023-12-20
[BUG] Dispose and create new view (non-coherent result errors)


ketourneau:
## Your Environment
* Version used: 2.4.1.1-next.14
* Browser Name and version: Firefox
* Operating System and version (desktop or mobile): Macos (sonoma)
* Link to your project: [source code](https://github.com/bloc-in-bloc/itowns/blob/hackathon2/examples/resize_extent_planarView.html)

## Context
We try to increase the extent size when user add files (view is already created).
So we try dispose the view and create new one.
Maybe the have another way to modify extent size ?

## Steps to Reproduce (for bugs)
[live example](https://demo-hackathon.netlify.app/examples/resize_extent_planarview)
1. Open console
2. Click on Resize view on left panel

## Expected Behavior
We expect view appear correctly without errors.

## Actual Behavior
View doesn't appear correcty and we got some errors.
![Capture d’écran 2023-12-20 à 10 24 19](https://github.com/iTowns/itowns/assets/39763266/aba97e98-fd23-4f58-b38c-613fa20e26ba)

PR #2279, opened 2024-02-23
fix(PlanarLayer): Fix delete new extent from globalExtentTMS const

ketourneau:
Proposed fix for #2244 

## Motivation and Context

When PlanarLayer is deleted, it doesn't remove extent from globalExtentTMS const.
It's why we got some trouble when we try to reload itowns with new extent. It keep previous extent added to globalExtentTMS.

I don't know if it's better to reset globalExtentTMS when view is disposed or remove extent from globalExtentTMS if PlanarLayer is deleted.

I think my fix may cause a problem if two PlanarLayer use the same extent and one was deleted.

What do you think ?

### DIFF
--- a/src/Core/Prefab/Planar/PlanarLayer.js
+++ b/src/Core/Prefab/Planar/PlanarLayer.js
@@ -11,6 +11,7 @@ import PlanarTileBuilder from './PlanarTileBuilder';
  * internally for optimisation.
  */
 class PlanarLayer extends TiledGeometryLayer {
+    #hasNewExtent;
     /**
      * A {@link TiledGeometryLayer} to use with a {@link PlanarView}. It has
      * specific method for updating and subdivising its grid.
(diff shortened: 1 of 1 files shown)


jailln:
Thanks for finding and correcting this bug :slightly_smiling_face:

It actually corrects the issue but I think that we need to re-think the `globalExtentTMS` implementation/usage as it doesn't allow to have multiple views with the same CRS and different extents at the same time. 

It seems that `globalExtentTMS` is used for two things:
* actually storing the "global extent" (or default extent) or a CRS (like EPSG:3857 global extent)
* storing the view extent in a global variable and use it to update layers (e.g. in [LayeredMaterialNodeProcessing](https://github.com/iTowns/itowns/blob/master/src/Process/LayeredMaterialNodeProcessing.js#L49), [LabelLayer](https://github.com/iTowns/itowns/blob/master/src/Layer/LabelLayer.js#L386)); in contexts where we already have access to the view and therefore to its extent (with `context.view`).
 
I propose that we limit the usage of `globalExtentTMS` to the first usecase and therefore:
* don't set new values in `PlanarLayer` constructor.
* Either force to provide an extent for `TMSSource` or set `TMSSource.extent` in `_preprocessLayer` instead than in `TMSSource` constructor. (I like the second option more)

WDYT @ketourneau @Desplandis ?

Disclaimer: I tried to explain my thoughts as clear as possible but we may discuss it in a voice call if needed :grin:

src/Core/Prefab/Planar/PlanarLayer.js
```diff
@@ -11,6 +11,7 @@ import PlanarTileBuilder from './PlanarTileBuilder';
  * internally for optimisation.
  */
 class PlanarLayer extends TiledGeometryLayer {
+    #isNewExtent;
```
jailln: if we decide to merge it without modifying `globalExtentTMS` usage (see my other comment), can you rename to `hasNewExtent` instead please?

jailln (dismissed):

src/Core/Prefab/Planar/PlanarLayer.js
ketourneau: Ok done !

ketourneau:
I agree with you.
I also wondered whether `globalExtentTMS` should be used for global extent (like EPSG:3857).

jailln:
> I also wondered whether globalExtentTMS should be used for global extent (like EPSG:3857).

You mean we should remove it totally?

ketourneau:
No, I just wanted to say that I agree with your suggestion.
I'm waiting to see what you decide.

Desplandis:
@jailln @ketourneau : According to @mgermerie, the root cause of this issue seems to be that tiles of the terrain (as created by the view) does not match that of an actual TMS. He detailed different solutions for this issue in the https://github.com/iTowns/itowns/issues/2290 proposal, which could fix other related issues in iTowns.

ketourneau:
@Desplandis @mgermerie We also have the same error with cog example when we try to dispose view and load another cog.

If you `Load 1 band sample` and then `Load RGB sample` with this commit : https://github.com/bloc-in-bloc/itowns/commit/833fc686e2cc6227fa5262b86deeab5c6ad39382 you will got this result :
Capture d’écran 2024-03-15 à 08 32 33

It's the same error, when we call `view.dispose(true)` it doesn't remove extent from globalExtentTMS const.
So if we reload view with another extent we got problem.

Desplandis:
@ketourneau Yeah I think this again a mix between the issue described in #2290 and cache coherency...

PR #2280, opened 2024-02-26
feat(TileLayer): add hide skirt parameter

AnthonyGlt:
## Description
Add a new parameter "hideSkirt" to the layer to hide the skirt (borders of tiles).

To test, the parameter can be updated like this ->  view.tileLayer.hideSkirt = true;

## Motivation and Context
Useful when the opacity < 1

With hideSkirt = false (current default behavior)
Screenshot from 2024-02-26 11-31-24

with hideSkirt = true
Screenshot from 2024-02-26 11-31-27

# Important
If merged, it's necessary to update the doc to describe this parameter and how to use it

### DIFF
--- a/src/Core/Prefab/TileBuilder.js
+++ b/src/Core/Prefab/TileBuilder.js
@@ -10,7 +10,7 @@ const cacheTile = new Cache();
 export default function newTileGeometry(builder, params) {
     const { sharableExtent, quaternion, position } = builder.computeSharableExtent(params.extent);
     const south = sharableExtent.south.toFixed(6);
-    const bufferKey = `${builder.crs}_${params.disableSkirt ? 0 : 1}_${params.segment}`;
+    const bufferKey = `${builder.crs}_${params.disableSkirt ? 0 : 1}_${params.segments}`;
     let promiseGeometry = cacheTile.get(south, params.level, bufferKey);
 
     // build geometry if doesn't exist
--- a/src/Core/Prefab/computeBufferTileGeometry.js
+++ b/src/Core/Prefab/computeBufferTileGeometry.js
@@ -1,5 +1,10 @@
 import * as THREE from 'three';
 
+export function getBufferIndexSize(segments, noSkirt) {
+    const triangles = (segments) * (segments) * 2 + (noSkirt ? 0 : 4 * segments * 2);
+    return triangles * 3;
+}
+
 export default function computeBuffers(params) {
     // Create output buffers.
     const outBuffers = {
@@ -26,7 +31,7 @@ export default function computeBuffers(params) {
     const computeUvs = [];
 
     const builder = params.builder;
-    const nSeg = params.segment || 8;
+    const nSeg = params.segments;
     // segments count :
     // Tile : (nSeg + 1) * (nSeg + 1)
     // Skirt : 8 * (nSeg - 1)
@@ -34,8 +39,6 @@ export default function computeBuffers(params) {
     if (nVertex > 2 ** 32) {
         throw new Error('Tile segments count is too big');
     }
-
-    const triangles = (nSeg) * (nSeg) * 2 + (params.disableSkirt ? 0 : 4 * nSeg * 2);
 
     outBuffers.position = new Float32Array(nVertex * 3);
     outBuffers.normal = new Float32Array(nVertex * 3);
@@ -46,13 +49,14 @@ export default function computeBuffers(params) {
     }
 
     computeUvs[0] = () => {};
+    const bufferIndexSize = getBufferIndexSize(nSeg, params.disableSkirt);
     if (params.buildIndexAndUv_0) {
         if (nVertex < 2 ** 8) {
-            outBuffers.index = new Uint8Array(triangles * 3);
+            outBuffers.index = new Uint8Array(bufferIndexSize);
         } else if (nVertex < 2 ** 16) {
-            outBuffers.index = new Uint16Array(triangles * 3);
+            outBuffers.index = new Uint16Array(bufferIndexSize);
         } else if (nVertex < 2 ** 32) {
-            outBuffers.index = new Uint32Array(triangles * 3);
+            outBuffers.index = new Uint32Array(bufferIndexSize);
         }
         outBuffers.uvs[0] = new Float32Array(nVertex * 2);
         computeUvs[0] = (id, u, v) => {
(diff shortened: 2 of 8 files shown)


jailln:
I think we should hide it by default when changing the opacity of the layer to always avoid the effect you show in the first screenshot; WDYT @AnthonyGlt @Desplandis ?

AnthonyGlt:
if we hide it, we see through the terrain , cf screen shot below.

To correct this, it's necessary to switch the scene background color to black and disable the atmosphere

Screenshot from 2024-02-26 14-08-03

jailln:
I meant hide it only when the opacity is < 1

AnthonyGlt:
the result will be nice when we don't zoom but when we zoom, we'll see the background color (like the screenshot above), can be a trade off.

jailln:
> the result will be nice when we don't zoom but when we zoom, we'll see the background color (like the screenshot above), can be a trade off.

Yes you're right, I'm not sure what's best so let's leave it like this for now

AnthonyGlt:
Linked proposal for future development of underground visualization
https://github.com/iTowns/itowns/issues/2281

src/Core/Prefab/TileBuilder.js
```diff
@@ -51,7 +51,9 @@ export default function newTileGeometry(builder, params) {
 
         const geometry = new TileGeometry(params, buffers);
         geometry.OBB = new OBB(geometry.boundingBox.min, geometry.boundingBox.max);
-
+        if (params.hideSkirt) {
+            geometry.setDrawRange(0, geometry.segments * geometry.segments * 2 * 3);//  bufferIndex = (nSeg) * (nSeg) * 2 * 3 (computeBufferTileGeometry.js)
+        }
```
Desplandis: This code is fragile. If there is any future change to the geometry computation in `computeBufferTileGeometry#computeBuffers`, this code would then be invalid.

I have two suggestions:
- Create a function in `computeBufferTileGeometry` to compute the draw range of the tile with and without its skirt. This avoid duplication and could be used at vertex creation and when limiting the draw range.
- Since this a code is specific to the tile geometry,  you could maybe move this code in a `TileGeometry#hideSkirt` setter.

src/Core/Prefab/computeBufferTileGeometry.js
```diff
@@ -22,11 +22,12 @@ export default function computeBuffers(params) {
         //        * u = wgs84.u
         //        * v = textureid + v in builder texture
         uvs: [],
+        segments: params.segments || 8,
```
Desplandis: This is not used in the function caller, is this change necessary?

src/Layer/TiledGeometryLayer.js
```diff
@@ -264,6 +275,20 @@ class TiledGeometryLayer extends GeometryLayer {
...
 
+    #hideExistingSkirt(value) {
+        for (const node of this.level0Nodes) {
+            node.traverse((obj) => {
+                if (obj.isTileMesh) {
+                    if (value) {
+                        obj.geometry.setDrawRange(0, obj.geometry.segments * obj.geometry.segments * 2 * 3);  //  bufferIndex = (nSeg) * (nSeg) * 2 * 3 (computeBufferTileGeometry.js)
+                    } else {
+                        obj.geometry.setDrawRange(0, Infinity);
+                    }
```
Desplandis: See previous comment in `TileBuilder.js`, this could be a setter in `TileGeometry`.

src/Layer/TiledGeometryLayer.js
```diff
@@ -71,6 +71,8 @@ class TiledGeometryLayer extends GeometryLayer {
 
         this.protocol = 'tile';
 
+        this.hideSkirt = false;
+
```
Desplandis: By the getter/setter below, this code is not used.

```suggestion
```

src/Layer/TiledGeometryLayer.js
```diff
@@ -98,6 +100,15 @@ class TiledGeometryLayer extends GeometryLayer {
             this.object3d.updateMatrixWorld();
         }));
 
+        let _hideSkirt = this.hideSkirt;
```
Desplandis:
```suggestion
        let _hideSkirt = false;
```

src/Layer/TiledGeometryLayer.js
```diff
@@ -264,6 +275,20 @@ class TiledGeometryLayer extends GeometryLayer {
         return ObjectRemovalHelper.removeChildren(this, node);
     }
 
+    #hideExistingSkirt(value) {
```
Desplandis: Given a recent proposal that I agree with, I think we should use an underscore to mark a function as private. Moreover, since this function is only used once, I think this function could be moved in the `hideSkirt` setter above.

Desplandis (changes requested):
Hi and thanks for your PR ! =)

I approve the proposed addition to the Core API, since the current render is kinda ugly when we need a transparent terrain! We could maybe enforce it (in another PR) when the transparency < 1.0? I dunno.

I have a few architecture comments if you can take a look. =)

src/Core/TileGeometry.js
```diff
@@ -31,6 +31,10 @@ class TileGeometry extends THREE.BufferGeometry {
 
         this.computeBoundingBox();
         this.OBB = {};
+        this.hideSkirt = params.hideSkirt;
```
AnthonyGlt: Not really satisfied about that. The setter will be called everytime, maybe : 
```suggestion
        if(params.hideSkirt){
           this.hideSkirt = params.hideSkirt;
           }
```
OR

```suggestion
        if(params.hideSkirt){
           this.setDrawRange(0, getBufferIndexSize(this.segments, params.hideSkirt));
           }
```

WDYT @Desplandis  ?

Desplandis: I agree, we should never have a property and a setter of the same name. I prefer the first solution since we do not duplicate code.

src/Layer/TiledGeometryLayer.js
```diff
@@ -101,6 +102,19 @@ class TiledGeometryLayer extends GeometryLayer {
         this.maxScreenSizeNode = this.sseSubdivisionThreshold * (SIZE_DIAGONAL_TEXTURE * 2);
     }
 
+    get hideSkirt() {
+        return !!this._hideSkirt;
```
Desplandis: Could you please initialize `this._hideSkirt` in the constructor. This eases type checking and avoids the double negation here?

Desplandis (approved):
Sound good to me, let's merge it! =)

Issue #2282, opened 2024-02-27
[Bug] TileMesh min and max elevation with multiple elevation layers


jailln:
Min and max elevation values of a `TileMesh` node are wrong when there multiple elevation layers are added to the view

## Context
[TileMesh](https://github.com/iTowns/itowns/blob/master/src/Core/TileMesh.js) has an OBB which z min and max values are [reset each time a new ElevationLayer is added to the view](https://github.com/iTowns/itowns/blob/master/src/Layer/ElevationLayer.js#L80) with the new ElevationLayer values (while it should consider all the elevation layers min and max values). Also, when removing one ElevationLayer from the view, [the bbox min and max z values are set to 0,0](https://github.com/iTowns/itowns/blob/master/src/Process/LayeredMaterialNodeProcessing.js#L238) while they should be computed from the remaining ElevationLayers. 

## Possible Cause/Fix/Solution

We need to:
* Assess where this obb z min and max are used
* Properly set its values when there is multiple elevation layers (take the min and max of all elevation layers)
* Re-compute the min and max z values when there is multiple elevation layers instead of resetting them to [0,0]

PR #2283, opened 2024-02-27
Fix ElevationLayer removal

### LINKED ISSUE #2221: [Bug - ElevationLayer]  Issues caused by the removal of an ElevationLayer
When we remove an ElevationLayer, the map bugs, tiles has faulty loading and may crash the browser

## Your Environment
Master Branch

## Context
On this video, I've implemented 2 buttons to use on the view_25d_map example
"Terrain IN" does  `view.addLayer(wmsElevationLayer);`
"Terrain OUT" does  `view.removeLayer('wms_elevation');`

[Screencast from 06-11-2023 16:21:03.webm](https://github.com/iTowns/itowns/assets/126568810/324f6d1e-a27c-4b2a-b059-adf363b06f56)

You can notice that the behavior is not the same before adding the terrain and after removing it. It should be the same

I tried it on the globe using view_3d_map example. The terrain is already added at the beginning, I juste remove it using the button, which does 
```
//ids => list of the elevation layers ids
for (id of ids) {
 view.removeLayer(id);
}
```
[Screencast from 06-11-2023 16:36:26.webm](https://github.com/iTowns/itowns/assets/126568810/ad55751f-8853-4869-b869-b61b42a4f8d6)

Same weird behavior and a huge decrease in performances.
I've tried with the cache boolean `view.removeLayer(id, cacheBoolHere);`, the behavior differs but no improvement.

## Expected Behavior
I expect when deleting the elevationlayer to have the same behavior as if the elevationlayer was never added

jailln:
## Description
`node.material.elevationLayerIds[layerId]` is removed by `node.material.removeLayer(layerId);` so the bbox min and max elevation values were never reset which caused visual artifacts when removing an `ElevationLayer` from the view.

You can use the following example to test the fix: https://gist.github.com/jailln/90c67187a0c8e4e94eb8ea5ac3cc92a0
 
Note that bbox min and max z values should not be reset to 0 but to the min and max of the remaining `ElevationLayers` attached to this `TileMesh`. This is another bug that also implies other changes in this bbox min and max elevation values management so I opened #2282 instead of fixing it now. (also note that the proposed implementation somehow doesn't seem to break anything if there are multiple `ElevationLayer` in the view and one is removed).

## Motivation and Context

Closes #2221

### DIFF
--- a/src/Process/LayeredMaterialNodeProcessing.js
+++ b/src/Process/LayeredMaterialNodeProcessing.js
@@ -236,12 +236,15 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
 }
 
 export function removeLayeredMaterialNodeLayer(layerId) {
+    /**
+     * @param {TileMesh} node - The node to udpate.
+     */
     return function removeLayeredMaterialNodeLayer(node) {
         if (node.material?.removeLayer) {
-            node.material.removeLayer(layerId);
-            if (node.material.elevationLayerIds[0] == layerId) {
+            if (node.material.elevationLayerIds.indexOf(layerId) > -1) {
                 node.setBBoxZ({ min: 0, max: 0 });
             }
+            node.material.removeLayer(layerId);
         }
         if (node.layerUpdateState && node.layerUpdateState[layerId]) {
             delete node.layerUpdateState[layerId];


src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -238,10 +238,10 @@ export function updateLayeredMaterialNodeElevation(context, layer, node, parent)
 export function removeLayeredMaterialNodeLayer(layerId) {
     return function removeLayeredMaterialNodeLayer(node) {
```
Desplandis, line 242:
```suggestion
    /**
     * @param {TileMesh} node - The node to udpate.
     */
    return function removeLayeredMaterialNodeLayer(node) {
```

Desplandis (approved):
Okay for me!
Fixes the announced bug!

PR #2284, opened 2024-03-01
fix(PointCloud): fix precision error for entwinePointTileLayer

ftoromanoff:
A precision error in EntWinePointLayer position has been noted in the current version of itowns.
This PR aim a solving that issue.

### DIFF
--- a/src/Parser/LASLoader.js
+++ b/src/Parser/LASLoader.js
@@ -71,13 +71,16 @@ class LASLoader {
         */
         const scanAngles = new Float32Array(view.pointCount);
 
+        // For precision we take the first point that will be use as origin for a local referentiel.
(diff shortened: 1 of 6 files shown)


src/Parser/LASLoader.js
```diff
@@ -71,13 +71,16 @@ class LASLoader {
         */
         const scanAngles = new Float32Array(view.pointCount);
 
+        // For precision we take the first point that will be use as origin for a local referentiel.
+        const origin = getPosition.map(f => f(0)).map(val => Math.floor(val));
```
AnthonyGlt, line 75: Why is it necessary to floor the value ?

ftoromanoff: To have integer instead of float. With float you can have an extra precision issue with such high numbers.

AnthonyGlt (approved):

PR #2288, opened 2024-03-11
feat(PointCloud): increase range of picking to u32

HoloTheDrunk:
## Description
Change the picking IDs from u16 to u32.

## Motivation and Context
Effectively eliminates the possibility of not having enough IDs to perform picking until GPUs start having 32GB+ of VRAM.
Tested on Firefox 123.0 (64-bit) on Ubuntu Linux 22.04.4 LTS x86_64

### DIFF
--- a/src/Core/Picking.js
+++ b/src/Core/Picking.js
@@ -148,11 +148,11 @@ export default {
 
         traversePickingCircle(radius, (x, y) => {
             const idx = (y * 2 * radius + x) * 4;
-            const data = buffer.slice(idx, idx + 4);
+            const data = buffer.slice(idx, idx + 3);
 
             // see PotreeProvider and the construction of unique_id
-            const objId = (data[0] << 8) | data[1];
-            const index = (data[2] << 8) | data[3];
+            const objId = data[0];
+            const index = (data[1] << 16) | data[2];
 
             const r = { objId, index };
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,6 +13,7 @@ The following people have contributed to iTowns.
   * [Madec Germerie-Guizouarn](https://github.com/mgermerie)
   * [Mathieu Brédif](https://github.com/mbredif)
   * [Aymeric Dutremble](https://github.com/a-dutremble)
+  * [Raphaël Duhen](https://github.com/HoloTheDrunk)
 
 * [CIRIL Group](https://www.cirilgroup.com/en/):
   * [Vincent Jaillot](https://github.com/jailln)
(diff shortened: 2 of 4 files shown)


HoloTheDrunk:
I don't know where the functional test errors are coming from, I can't replicate them locally so I'm guessing (hoping) it's a toolchain problem and not on my end

src/Core/Picking.js
```diff
@@ -148,11 +148,11 @@ export default {
...
         traversePickingCircle(radius, (x, y) => {
             const idx = (y * 2 * radius + x) * 4;
-            const data = buffer.slice(idx, idx + 4);
+            const data = buffer.slice(idx, idx + 3);
 
             // see PotreeProvider and the construction of unique_id
-            const objId = (data[0] << 8) | data[1];
-            const index = (data[2] << 8) | data[3];
+            const objId = data[0];
+            const index = (data[1] << 16) | data[2];
```
Desplandis, line 155: You encode each 48-bits `unique_id` value as 3 `uint16`. However, you read the resulting texture with a `uint8` buffer (see [`c3dEngine#renderViewToBuffer`](https://github.com/iTowns/itowns/blob/ee56ec70626213390fa200573552ddb5a54a3b94/src/Renderer/c3DEngine.js#L147))...

I don't think values are preserved here.

Desplandis (changes requested):

src/Core/Picking.js
HoloTheDrunk: We can provide a u16 buffer ourselves
```diff
- const buffer = view.mainLoop.gfxEngine.renderViewToBuffer(
-     { camera: view.camera, scene: tileLayer.object3d },
-     {
-         x: viewCoords.x - radius,
-         y: viewCoords.y - radius,
-         width: 1 + radius * 2,
-         height: 1 + radius * 2,
-     });
+ const zone = {
+     x: viewCoords.x - radius,
+     y: viewCoords.y - radius,
+     width: 1 + radius * 2,
+     height: 1 + radius * 2,
+ };
+ zone.buffer = new Uint16Array(4 * zone.width * zone.height);
+ const buffer = view.mainLoop.gfxEngine.renderViewToBuffer(
+     { camera: view.camera, scene: tileLayer.object3d },
+     zone,
+ );
```
But looking into things a bit more I'm not sure how this is gonna interact with the default `fullSizeRenderTarget` used by the `c3DEngine` that gets created with u8 rgba formatting.
I currently don't have an example I could use to check for this issue, as the results don't change for the example I originally used for this PR ([potree 3d map](https://github.com/iTowns/itowns/blob/master/examples/potree_3d_map.html)).

Issue #2225, opened 2023-11-10
[Proposal - 3D Tiles] Use an external library for 3D Tiles support


jailln:
This issue is a feature proposal. Feel free to upvote (with 👍 ), comment and provide your use-cases if you're interested by this feature.

## Context

(Originally discussed in https://github.com/iTowns/itowns/discussions/2146)

3D Tiles support in itowns is good but could be improved. 3D Tiles format is quite complex and maintaining its support is costly. In addition, 3D Tiles 1.1 has been released recently.

## Proposal

**Analyze the pros and cons and the different possibilities for using an external library to support 3D Tiles in itowns and decide if we should use one or keep the 3D Tiles support in itowns.**

We have identified three possible libraries:

* [loaders.gl](https://loaders.gl/): contains everything to parse, and handle 3D tiles for rendering (culling for instance)
* [three-loader-3dtiles](https://github.com/nytimes/three-loader-3dtiles): Uses loaders.gl to load 3D tiles and displays them with threejs
* [3DTilesRendererJS](https://github.com/NASA-AMMOS/3DTilesRendererJS): parses and renders 3D Tiles, based on threejs

The (unordered) elements to bear in mind when analyzing external libraries are:
* License 
* 3D Tiles support (1.0, 1.1 and also previous versions official or unofficial in which there are still tilesets available) and GLTF support (1.0 and 2.0)
* Performance
* Community, activity, etc. 
* Possibility to contribute
* Dependencies / bundle size

## Potential Problems

* Losing support of some 3D tiles features
* Losing cache control
* Losing style control
* Losing gltf 1.0 support

## Documentation

Given the wide variety of 3D Tiles datasets available and of the variety of possible representations (meshes with or without textures, point clouds, gltfs, instanced meshes, multiple possible bounding volumes, draco and ktx2 compression, etc., etc.), we need to test against multiple datasets. Here are some public datasets that I have gathered. I also have datasets that I cannot share but I can run tests on them and post the results:

**3D Tiles 1.0**

**b3dm without textures:**
* https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.0/TilesetWithDiscreteLOD: A dragon with multiple LOD in the USA, EPSG:4978 
* https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.0/TilesetWithRequestVolume: Tileset with request volume, EPSG:4978
[...]

PR #2295, opened 2024-03-15
3d tiles renderer js migration bootstrap

jailln:
## Description

Starts the migration towards using 3dtilesRendererJS for 3D tiles. I created the `3d-tiles-migration` branch in which we can merge our work related to this migration. This is a first implementation with some TODOs left in the code 

Already implemented:

* [x] Create a new 3D tiles layer using 3dtilesrendererjs and deprecate `C3DTilesLayer`
* [x] Draco and KTX2 support
* [x] Cesium ion support
* [x] Google 3D tiles support
* [x] Gltf 1.0 support
* [x] CESIUM_RTC GLTF 1.0 extension support
* [x] itowns PointsMaterial support (ongoing) 

Some TODOs left in the code (don't hesite to give your opinion), commits are still messy and still some stuff to do to be iso with the `C3DTilesLayer` implementation (see below).

## Motivation and Context

#2225

## Notes on the implementation

* Used the name `ThreeDTilesLayer`, WDYT?
* Even if 3dtilesRendererJS `TilesRenderer` is conceptually not at the same level than an itowns `Layer`, I decided to instance one `TilesRenderer` per `ThreeDTilesLayer`. Also I kept the `C3DTilesIonSource` and `C3DTilesGoogleSource` for `C3DTilesLayer` to keep working but also to keep itowns Source / Layer instanciation logic in the API (even if these sources are not really useful anymore now that we use `3dtilesRendererJS` because all the logic related to querying cesium ion an google sources are handled by `3dtilesRendererJS`).
* I implemented a wrapper around threejs `GLTFLoader` and `LegacyGLTFLoader` that loads gltf regardless of their version (1.0 or 2.0) and that shares the same interface than three `GLTFLoader` to allow loading gltf 1.0 files with 3dtilesrendererjs (named `iGLTFLoader` for "itownsGltfLoader"). This class is exported so users that need to load gltf 1.0 or 2.0 files in their applications can benefit from it.
* I implemented CESIUM_RTC gltf 1.0 extension support in the `LegacyGLTFLoader`
* Regarding draco and ktx2, I decided to keep the static `enableDracoLoader` and `enableKTX2Loader` that sets the draco and ktx2 path in itowns internal instance of `iGLTFLoader`.

Don't hesitate to give your opinion on these choices :)

## First tests

I started running some tests, it seems that:
* Google 3D tiles display is lightning fast
* `3dtiles_25d.html` example seems slower
* Point clouds:
  * slower
  * there seems to be a SSE issue (some visible tiles are not loaded)
  * the batch table is not loaded

I will take some time today to do some more performance tests and try to assess where these drawbacks may come from.

## Remaining TODO

Still a lot to do, see [this list](https://github.com/iTowns/itowns/issues/2225#issuecomment-1999238765) :)


jailln:
@Desplandis I rebased my branch and updated `3d-tiles-renderer-js` to `0.3.33` (last version).

Except from the remaining points listed [here](https://github.com/iTowns/itowns/issues/2225#issuecomment-1999238765), I'm not really satisfied of the modifications I made to the `PointVS` shader (there should be a better way than having a `#if` inside the if/else conditions for the `PNTS_MODE`) and the classification mode does not work in the `3dtiles_pointcloud.html` yet, I need to check why. I still wanted to push my rebase so you can review before I come back at it next thursday :slightly_smiling_face:  

Note that because of [this](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/577) that has not been released yet you need to make this correction in the 3d-tiles-renderer-js source files in itowns `node_modules/` for it to work (or simply downgrade to 0.3.32 should work too)

mgermerie:
[The PR on 3DTilesRendererJS](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/577) you mentioned has been merged and in their latest `0.3.34` release. I bumped the version and pushed it on your branch so the manual correction is no longer necessary.

PR #2296, opened 2024-03-15
ESM distribution

### LINKED ISSUE #2256: [Proposal - Requirements] Formalize web browsers requirements for iTowns
## Context

There is currently no documented web browsers requirements for iTowns in the whole repository.

The closest to a documentation is our configuration file of the `babel` compiler which states that we are *theoretically* targeting the oldest browsers possible to be ES5 compatible. That means that we are supporting browsers a decade old (e.g. IE11)!

This causes a number of issues:
- we have no guarantee that iTowns indeed works on those old platforms
- the distributed code is unreadable due to code transformations by `babel`
- the distributed code size is large due to polyfills for unsupported constructs on those platforms (even for constructs well-supported for years by mainstream browsers).

Those issues will be exacerbated if we do not take some actions.

## Discussion

@iTowns/product-owners @iTowns/core-devs I propose that we discuss in this issue the minimum targets that we should support with iTowns.

Note that `three` will drop support of `webgl1` at the end of the month, which means that all our targets should at least support `webgl2`, Safari being the last mainstream browser to add support for `webgl2` circa 2021.

A proposed solution is too only support the `defaults` target given by [browserlist](https://github.com/browserslist/browserslist), that is the last two versions of all mainstream browsers (> 0.5% usage) + Firefox ESR at the date of release.

## Actions

- Documents the requirements (`README.md`, website, ...)
- Implements the proposal (see #2252)

Desplandis:
## Description

Follow-up of #2256, this PR drops the `CJS` (CommonJS modules) distribution is favor of a standard `ESM` (ECMAScript) distribution.

Fixes #2256

## Motivation and Context
#2252 bumped the web browser requirements. This guarantees that we now only support ESM-compatible browsers. Moreover, `three.js` is currently migrating to a pure `ESM` package.

## TODOs
- [x] add the `type: module` field to the `package.json`
- [x] `webpack` config file.
- [x] `eslint` config files (loaded with `require()`, shall be .cjs)
- [x] unit tests.
- [x] functional tests (CJS to ESM syntax).
- [x] coverage tests (incompatibility with `nyc`, change to `c8)`. #2299 
- [x] transpilation with `babel`. 
- [x] documentation scripts (JSDoc `publish.js` is loaded with `require()` and shall be commonJS).
- [x] `prepare` script (already `.mjs`)
- [x] `replace` script (already `.mjs`)
- [x] `changelog` script (loaded with `require()`, shall be `.cjs`)
- [x] `bump` script (loaded with `require()`, shall be `.cjs`)


jailln (approved):
Thanks!! :rocket:

Issue #2292, opened 2024-03-14
[BUG] Difference in representation of MVT between iTowns and OpenLayer


AnthonyGlt:
It seems that iTowns is requesting one zoom level above openlayer.

## Context
When we synchronize an OpenLayer view with an Itowns view, we can see that, for the same extent, we have 2 differents representations. 

## Steps to Reproduce (for bugs)
Open a view in openLayer with a MVT feed such as https://data.geopf.fr/annexes/ressources/vectorTiles/styles/PLAN.IGN/standard.json

## Expected Behavior
Having the same details/representation in both engine

## Actual Behavior
![Screenshot from 2024-03-14 15-47-51](https://github.com/iTowns/itowns/assets/126568810/14d6931d-1fe6-404c-b767-da7375e8e7c5)

## Possible Cause/Fix/Solution
From https://wiki.openstreetmap.org/wiki/Zoom_levels, at the end:
```Mapbox GL based libraries uses 512×512-pixel tiles by default, so Mapbox GL zoom levels are one fewer than the zoom levels described above that are used by other tools.```

it sounds a lot like the issue described here

Desplandis:
The vector tiles from IGN's `PLAN.IGN` is distributed as a TMS, so both `openlayers` and `itowns` should have the same tiles for a given zoom level. I think your issue is caused by the fact that `itowns` subdivides the terrain before `openlayers`.

The subdivision threshold should be configurable in both `GlobeView` (resp `PlanarView`) or directly in `GlobeLayer` (resp. `PlanarLayer`). It could maybe be not exposed or documented for the end-user...

However if you decide to delay the subvision of the tile, your data may appear blurry at the default resolution of 256x256 pixels. It seems however that this resolution is hardcoded in `itowns` but IHMO it should set as a parameter of the `Layer`.

N.B. It seems that `FeatureToTexture` has a parameter to control the texture size however it is called with the value `256` : https://github.com/iTowns/itowns/blob/2fb3298f07a07cc1b7fb98025479da29f6f51557/src/Converter/textureConverter.js#L31

AnthonyGlt:
It looks like I need to change in 3 places :
https://github.com/iTowns/itowns/blob/2fb3298f07a07cc1b7fb98025479da29f6f51557/src/Converter/textureConverter.js#L31
https://github.com/iTowns/itowns/blob/2fb3298f07a07cc1b7fb98025479da29f6f51557/src/Provider/Fetcher.js#L4
[...]

PR #2297, opened 2024-03-18
feat(TextureSize): Add parameter to define the textureSize

AnthonyGlt:
## Related issue:

https://github.com/iTowns/itowns/issues/2292
https://github.com/iTowns/itowns/issues/2312
https://github.com/iTowns/itowns/issues/2309

Replace the const use for subdivision and texture conversion with a customizable parameter `sizeTextureTile`

Example:
For MVT layer
```
 const mvtSource = new itowns.VectorTilesSource({
              style: 'https://data.geopf.fr/annexes/ressources/vectorTiles/styles/PLAN.IGN/standard.json',
              crs: 'EPSG:3857',

          });

  const mvtLayer = new itowns.ColorLayer('mvtLayer',{
      source: mvtSource,
      sizeTextureTile:512,
      addLabelLayer: true,

  });
```

For the globe:

` const view = new itowns.GlobeView(viewerDiv, placement, {sizeTextureTile:512});`

### DIFF
--- a/src/Layer/Layer.js
+++ b/src/Layer/Layer.js
@@ -66,6 +66,7 @@ class Layer extends THREE.EventDispatcher {
      * @param {boolean} [config.addLabelLayer.performance=false] - In case label layer adding, so remove labels that have no chance of being visible.
      * Indeed, even in the best case, labels will never be displayed. By example, if there's many labels.
      * @param {boolean} [config.addLabelLayer.forceClampToTerrain=false] - use elevation layer to clamp label on terrain.
+     * @param {number} [config.subdivisionThreshold=256] - set the texture size and, if applied to the globe, affects the tile subdivision.
      *
      * @example
      * // Add and create a new Layer
@@ -108,6 +109,8 @@ class Layer extends THREE.EventDispatcher {
             config.style = new Style(config.style);
         }
         this.style = config.style || new Style();
+        this.subdivisionThreshold = config.subdivisionThreshold || 256;
+        this.sizeDiagonalTexture =  (2 * (this.subdivisionThreshold * this.subdivisionThreshold)) ** 0.5;
         Object.assign(this, config);
 
         Object.defineProperty(this, 'id', {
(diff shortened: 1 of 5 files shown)


src/Layer/Layer.js
```diff
@@ -108,6 +108,8 @@ class Layer extends THREE.EventDispatcher {
             config.style = new Style(config.style);
         }
         this.style = config.style || new Style();
+        this.sizeTextureTile = config.sizeTextureTile || 256;
```
jailln: This seems more "source-dependent" than "layer-dependent" because it is related to the source data. I would rather define this in `Source`. WDYT @Desplandis ? 

Then, we should default it to `512` for `VectorTileSource`.

Also, either here or in `Source`, can you document it in the public API please?

Desplandis: @jailln This seem to be what openlayers does (see https://openlayers.org/en/latest/apidoc/module-ol_source_VectorTile-VectorTile.html).

However, we rasterize the vector data in itowns in the layer (i.e. `convert`), so it could be a parameter of the layer.

AnthonyGlt:
We'll keep it in the layer because we have access to the globe layer from there and then we can affect the tile size for the subdivision.
(TODO put a comment in the code)

mgermerie (commented):
This PR is good to merge in my opinion except for one thing : the name of the parameter.
It think `sizeTextureTile` (which would sound more correct as `tileTextureSize`) refers only to the size of a texture. This is not what it describes in one of its two usecases : the tile size above which a globe tile is subdivided.
Perhaps we could benefit from this PR to change the name inherited from the old constant `SIZE_TEXTURE_TILE` to something like `subdivisionThreshold` or any other suggestion ?

AnthonyGlt:
I agree, the name should be changed. `subdivisionThreshold` is more appropriate but there is `sseSubdivisionThreshold` already existing.

mgermerie (approved):
It's all good for me. You could you just clean your commit tree and it will be good to merge. Thanks !

PR #2299, opened 2024-03-21
Change coverage tool from nyc to c8

Desplandis:
## Description

This PR replace our coverage tool by [`c8`][c8], a functional equivalent and CLI-compatible alternative to `nyc` .

## Motivation and Context
Our coverage tool `nyc` does not support ES modules which is a no-go. An [issue](https://github.com/istanbuljs/nyc/issues/1287) is opened since 2020. Reading this thread, it seems that most projects have migrated from `nyc` to `c8`.

[c8]: https://github.com/bcoe/c8

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "babel-plugin-minify-dead-code-elimination": "^0.5.2",
         "babel-plugin-minify-replace": "^0.5.0",
         "babel-plugin-module-resolver": "^5.0.0",
+        "c8": "^9.1.0",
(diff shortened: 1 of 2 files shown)


jailln:
image
This tool is nicer than nyc :grinning:

jailln (approved):

Desplandis:
> image This tool is nicer than nyc 😀

Seems so xD
I will double-check if it tracks the same file as `nyc` before merging. x)

P.S.: Everything seems ok!

Issue #2197, opened 2023-09-28
[Packaging] export addons (widgets and debug tools)


mgermerie:
## Context

Currently, [widgets](https://github.com/iTowns/itowns/tree/master/src/Utils/gui) are exported within itowns bundle - in `itowns/widgets`. [`GuiTools`](https://github.com/iTowns/itowns/blob/master/examples/js/GUI/GuiTools.js) are available as a single js file download within `examples/js/GUI` folder. [Debug tools](https://github.com/iTowns/itowns/tree/master/utils/debug) are not published and not suitable for use within an external project, as discussed in #2061.
There are also a few addons - some of which could get an update - available as a js file download in [`examples/js/plugins`](https://github.com/iTowns/itowns/tree/master/examples/js/plugins) folder.

The packaging of these addons could be harmonized, making them available and simple to use in an external project. Meanwhile, the following addons would need to be rewritten/updated : 
- [ ] #2201
- [ ] [`GuiTools`](https://github.com/iTowns/itowns/blob/master/examples/js/GUI/GuiTools.js), which could upgrade from `dat.gui` to [`lil-gui`](https://lil-gui.georgealways.com/) (used by three.js).
- [ ] [debug tools](https://github.com/iTowns/itowns/tree/master/utils/debug) whose code need some cleanup and which could also use [`lil-gui`](https://lil-gui.georgealways.com/).

There is also the question of GitHub repository. We can either extract those addons to a separate repository in `iTowns` organization or keep them in `iTowns/itowns` repo. On this matter, three.js opted for the single repository option. It has the advantage of a simpler setup and maintenance.

These subjects are open to discussion.

## Suggested solutions for packaging

Provided we opt for the single GitHub repository option, we could do as three.js addons :
- Gather them within an `addons` folder which could be at root level or in `./examples`. The equivalent in three.js is the [`examples/jsm`](https://github.com/mrdoob/three.js/tree/dev/examples/jsm) folder.
- Export them under an `itowns/addons` sub-package - as done [here](https://github.com/mrdoob/three.js/blob/181e04eea8b569dc09048f9dc644310ed6b745a6/package.json#L15) by three.js - or as a package of an `@itowns` npm organization we could create.

If we choose the separate GitHub repository approach, I don't know if we could still publish addons as an itowns sub-package, but we could still publish them as a package of `@itowns` npm organization.

valentinMachado:
[...]

Issue #2201, opened 2023-10-05
[Refacor] rework plugins within `examples/js/plugins` folder


jailln:
## Context

Plugins within `examples/js/plugins` should refactored and updated. 

- [ ] [`CSVnVRTParser.js`](https://github.com/iTowns/itowns/blob/master/examples/js/plugins/CSVnVRTParser.js) and [`TIFFParser`](https://github.com/iTowns/itowns/blob/master/examples/js/plugins/TIFFParser.js) should at least be rewritten as classes. Also, perhaps the `TIFFParser` could be added to `src/Parser` but this is a discussion for [another issue](https://github.com/iTowns/itowns/issues/2202).
- [ ] [`FeatureToolTip.js`](https://github.com/iTowns/itowns/blob/master/examples/js/plugins/FeatureToolTip.js) could also at least be rewritten as a class.
- [ ] [`DragNDrop.js`](https://github.com/iTowns/itowns/blob/master/examples/js/plugins/DragNDrop.js) could also be rewritten as a class and be added the support of some other format. I think @ftoromanoff previously worked on shapefile support in this plugin.

## Motivation

This work is needed to provide users with a harmonized, clean and easy to use plugin package, as discussed in #2197.

Issue #1930, opened 2022-10-20
Update and export example plugins


mgermerie:
The plugins available [here](https://github.com/iTowns/itowns/tree/master/examples/js) would need an update and refactoring. Moreover, we could export them the same way we do with widgets.

PR #2300, opened 2024-03-21
Monorepo structure

gchoqueux:
## Description
Refactoring iTowns to **monorepo** structure.

The goal is to split itowns down into feature packages.
As an example, I've started with the geographic functionalities. (**@itowns/geographic**)

- [x] move unit test
- [x] improve dev server
- [x] babel resolver plugin is broken
- [x] documentation
- [x] coverage
- [x] bump version
- [x] move all src in packages
- [x] watch src packages
- [x] publishing, auto publish with github actions
- [x] ~~create organization NPM to support  @itowns~~
- [x] move dependencies to packages
- [x] webpack : use alias to resolve dependencies instead of `exportsFields`
- [x] use babel monorepo configuration
- [x] move export from private root package.json
- [x] write all readme.md files

For the moment `Debug` and `Widget` modules are private.

## Motivation and Context
Split code in packages for clearly structured code.
Simplifies development and facilitates contributions.
Increases the scope of users who only want to use a few functions.
This structure makes it necessary to make functions independent 

### Code movement
The classes `Coordinates, Crs, Ellipsoid, Extent, OrientationUtils` files are moved to `./packages/geographic/src`.
The `proj4` dependency moves to geographic package.

The unit tests moves to `packages/xxx/test`.
the code `utils/debug` moves `packages/Debug`
the code `src/Utils/Gui` moves  `packages/Widget`

### dependencies
This proposal from monorepo doesn't use monorepo package (by example [Lerna](https://lerna.js.org/)).
New dev dependency is `Concurrently`, it's used to watch script in parallel 
`Itowns dependencies` (three, proj4...) are moved to the packages that use them

### node scripts
The scripts uses the options `--workspaces` or `-ws` , to call the corresponding script in each sub packages.


gchoqueux:
@Desplandis I resolve some issues 

* resolve module with **exportsFields** option
* remove **lerna** with **--workspace** cli option

Desplandis:
@gchoqueux Nice! I will take a quick look of your changes and write a todo list of things that should absolutely be tested before reviewing more thoroughly this PR. Expect it for tomorrow or next tuesday!

@jailln @mgermerie @ftoromanoff @AnthonyGlt I think we'll need your inputs on this change!  ;)

This should be a good start to fix #2197, #1930, #2201 (list not exhaustive, feel free to complete). Shall we create a meta-issue to aggregate all those issues?

package.json
```diff
@@ -8,33 +8,39 @@
     ".": "./lib/Main.js",
     "./widgets": "./lib/Utils/gui/Main.js"
   },
+  "workspaces": [
+    "packages/*"
+  ],
   "scripts": {
```
Desplandis, line 12: Shall the `lint` script call the `lint` script of each sub-packages?

package.json
```diff
@@ -8,33 +8,39 @@
...
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\" \"docs/*.js\"",
     "doc": "jsdoc --readme docs/HOMEPAGE.md -c docs/config.json",
     "doclint": "npm run doc -- -t templates/silent",
-    "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-with-coverage && npm run test-functional",
-    "test-dev": "npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage && npm run test-functional",
+    "test": "npm run transpile --workspaces && npm run test --workspaces && npm run lint -- --max-warnings=0 && npm run build && npm run test-with-coverage && npm run test-functional",
```
Desplandis: Shall we `transpile` before launching each scripts? It think it is tedious to follow all workspaces call (e.g. `transpile` is also called in `test-with-coverage`...).

package.json
```diff
@@ -8,33 +8,39 @@
...
+    "test-dev": "npm run test --workspaces && npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage && npm run test-functional",
```
Desplandis:
```suggestion
    "test-dev": "npm run test-dev --workspaces && npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage && npm run test-functional",
```

package.json
```diff
@@ -8,33 +8,39 @@
...
     "test-unit": "npm run base-test-unit test/unit",
     "test-functional": "mocha -t 100000 --require test/hooks_functional.js --recursive test/functional",
-    "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
-    "test-with-coverage_lcov": "c8 -n src --reporter=lcov cross-env npm run test-unit",
+    "test-with-coverage": "npm run transpile --workspaces && c8 -n src -r html cross-env npm run test-unit",
```
Desplandis: Shall we transpile before launching each scripts? Can we call test-with-coverage recursively on each sub-package?

package.json
```diff
@@ -8,33 +8,39 @@
...
+    "test-with-coverage_lcov": "npm run transpile --workspaces && c8 --temp-directory ./coverage/tmp --report-dir ./coverage/itowns -n src --reporter=lcov cross-env npm run test-unit && npm run test-with-coverage_lcov --workspaces && c8 report --reporter=lcov",
```
Desplandis: TODO: A global coverage or coverage per packages. What are the best practices ?

package.json
```diff
@@ -8,33 +8,39 @@
...
     "build": "cross-env NODE_ENV=production webpack",
     "build-dev": "cross-env NODE_ENV=development webpack",
-    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
+    "transpile": "npm run transpile --workspaces && cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
     "start": "cross-env NODE_ENV=development webpack serve",
     "start-https": "cross-env NODE_ENV=development webpack serve --https",
     "debug": "cross-env noInline=true npm start",
     "prepublishOnly": "npm run build && npm run transpile",
     "prepare": "cross-env NO_UPDATE_NOTIFIER=true node ./config/prepare.mjs && node ./config/replace.config.mjs",
     "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib",
```
Desplandis: Unfortunately, we do not watch the sub-packages.

package.json
```diff
@@ -8,33 +8,39 @@
...
     "changelog": "conventional-changelog -n ./config/conventionalChangelog/config.cjs -i changelog.md -s",
-    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm install && git add -A && git commit --amend --no-edit",
+    "bump": "npm run bump $npm_config_level --workspaces && if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm run update-packages && npm install && git add -A && git commit --amend --no-edit",
```
Desplandis: If we are to align the version number between the top-level package and all sub-packages, why not have a single script here and not call `bump` recursively on all sub-packages? Shall we not bump the dependency version too?

package.json
```diff
@@ -8,33 +8,39 @@
...
     "publish-next": "npm version prerelease --preid next && npm publish --access public --tag=next --provenance",
-    "publish-latest": "npm publish --access public --tag=latest --provenance"
+    "publish-latest": "npm publish --access public --tag=latest --provenance",
```
Desplandis: Okay and how we publish the sub-package?

package.json
```diff
@@ -8,33 +8,39 @@
...
     "start": "cross-env NODE_ENV=development webpack serve",
     "start-https": "cross-env NODE_ENV=development webpack serve --https",
     "debug": "cross-env noInline=true npm start",
     "prepublishOnly": "npm run build && npm run transpile",
     "prepare": "cross-env NO_UPDATE_NOTIFIER=true node ./config/prepare.mjs && node ./config/replace.config.mjs",
     "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib",
     "changelog": "conventional-changelog -n ./config/conventionalChangelog/config.cjs -i changelog.md -s",
-    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm install && git add -A && git commit --amend --no-edit",
+    "bump": "npm run bump $npm_config_level --workspaces && if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm run update-packages && npm install && git add -A && git commit --amend --no-edit",
     "publish-next": "npm version prerelease --preid next && npm publish --access public --tag=next --provenance",
```
Desplandis: Okay and how we publish the sub-package?

packages/Geodesy/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@itowns/geodesy",
+  "version": "2.43.0",
+  "description": "Geodesy",
+  "type": "module",
+  "main": "lib/Main.js",
+  "itowns_exports": {
+    ".": "./src/Main.js"
+  },
```
Desplandis: Semantic of this option?

package.json
```diff
@@ -8,33 +8,39 @@
     ".": "./lib/Main.js",
     "./widgets": "./lib/Utils/gui/Main.js"
   },
+  "workspaces": [
+    "packages/*"
+  ],
   "scripts": {
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\" \"examples/**/*.js\" \"docs/*.js\"",
     "doc": "jsdoc --readme docs/HOMEPAGE.md -c docs/config.json",
```
Desplandis, line 14: It seems that the changes in this PR causes duplicates in the documentation of modules in `@itowns/geodesy`.

.eslintrc.cjs
```diff
@@ -1,3 +1,7 @@
+const path = require('path');
+
+
+
```
Desplandis:
```suggestion
```

docs/config.json
```diff
@@ -185,7 +185,7 @@
         }
     },
     "source": {
-        "include": [ "src", "examples/js/plugins" ],
+        "include": [ "src", "examples/js/plugins", "packages" ],
```
Desplandis: Seems that the generation of the documentation of modules from `packages` does not work correctly: methods are generated two times.

package.json
```diff
@@ -8,33 +8,39 @@
     ".": "./lib/Main.js",
     "./widgets": "./lib/Utils/gui/Main.js"
   },
+  "workspaces": [
+    "packages/*"
```
Desplandis: It does not currently works with `yarn`, seems that `yarn` is stricter than `npm` and [requires that the top-level package is `private`][yarn:workspaces].

[yarn:workspaces]: https://classic.yarnpkg.com/lang/en/docs/workspaces/#toc-how-to-use-it

Desplandis: I think we should explicitely list `worskpaces`, since a dependant package shall be declared *after* its dependency.

packages/Geodesy/.babelrc
```diff
@@ -0,0 +1,36 @@
+{
```
Desplandis: Can [`overrides`](https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages#selective-compilation-with-overrides) be used to not duplicate the babel config file?

packages/Geodesy/Gruntfile.cjs
```diff
@@ -0,0 +1,27 @@
+module.exports = (grunt) => {
```
Desplandis: Again is there a way to not duplicate this configuration?

packages/Geodesy/README.md
```diff
@@ -0,0 +1,31 @@
+## iTowns Geodesy
+
+# install
+
+`npm install @itowns/geodesy`
+
+# getting started
```
Desplandis:
```suggestion
# Install

`npm install --save @itowns/geodesy`

# Getting started
```

packages/Geodesy/README.md
```diff
@@ -0,0 +1,31 @@
...
+Visiting [epsg.io](https://epsg.io/) to more coordinate system worldwide.
+
+Visiting iTowns [documentation](http://www.itowns-project.org/itowns/docs/#home) for more information.
+
+
+
+
+
+
+
```
Desplandis:
```suggestion
```

packages/Geodesy/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "scripts": {
+    "build": "",
+    "lint": "eslint \"src/**/*.js\" \"test/**/*.js\"",
+    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
+    "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --loader=babel-register-esm",
+    "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-with-coverage",
+    "test-dev": "npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage",
+    "test-unit": "npm run base-test-unit test/unit",
+    "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
+    "test-with-coverage_lcov": "c8 --clean false --temp-directory ../../coverage/tmp  --report-dir ../../coverage/geodesy -n src --reporter=lcov cross-env npm run test-unit",
```
Desplandis: Would this be easier to have all those scripts at top-level?

packages/Geodesy/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi"
+  },
+  "files": [
+    "src"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/itowns"
+  },
+  "license": "(LGPL-2.1)",
```
Desplandis: The code of this package is MIT or CECILL-B.

packages/Geodesy/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "bugs": {
+    "url": "https://github.com/itowns/itowns/issues"
+  },
+  "homepage": "https://itowns project"
```
Desplandis:
```suggestion
  "homepage": "https://itowns.github.io/"
```

packages/Geodesy/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-with-coverage",
+    "test-dev": "npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage",
+    "test-unit": "npm run base-test-unit test/unit",
+    "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
+    "test-with-coverage_lcov": "c8 --clean false --temp-directory ../../coverage/tmp  --report-dir ../../coverage/geodesy -n src --reporter=lcov cross-env npm run test-unit",
+    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi"
+  },
+  "files": [
+    "src"
+  ],
```
Desplandis:
```suggestion
  "files": [
    "*.md",
    "src",
    "lib"
  ],
```

packages/Geodesy/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/itowns"
```
Desplandis:
```suggestion
    "url": "git+https://github.com/iTowns/itowns.git"
```

packages/Geodesy/test/unit/coordinate.js
```diff
@@ -126,3 +126,4 @@ describe('Coordinates', function () {
         assert.equal(distance, coord1.spatialEuclideanDistanceTo(coord0));
     });
 });
+
```
Desplandis:
```suggestion
```

webpack.config.cjs
```diff
@@ -44,7 +44,8 @@ module.exports = () => {
         mode,
         context: path.resolve(__dirname),
         resolve: {
-            modules: [path.resolve(__dirname, 'src'), 'node_modules'],
+            exportsFields: ['itowns_exports', 'exports'],
```
Desplandis: I'm not so fan of using a non-standard field in `package.json`. Is this possible to use `alias` to achieve the same behavior?

webpack.config.cjs
```diff
@@ -44,7 +44,8 @@ module.exports = () => {
...
+            modules: [path.resolve(__dirname, 'src'), path.resolve(__dirname, './packages/Geodesy/src'), 'node_modules'],
```
Desplandis: Shall this not be `babel` which should resolve the files dependencies (i.e. in `include`)?

Desplandis (commented):
Those are my comments for now.
However, I have a few more remarks given the CI, have you tested with a new release?

package.json
gchoqueux: no

package.json
gchoqueux: it should be decided who is responsible for launching workspace tests and transpile

package.json
gchoqueux: The entry package is `lib/main.js` so the package must be transpiled first

package.json
gchoqueux: > If we are to align the version number between the top-level package and all sub-packages, why not have a single script here and not call bump recursively on all sub-packages?

Finally, the methods would be equivalent?

Shall we not bump the dependency version too?

I don't understand

package.json
gchoqueux: it seems so, I'll check it out

packages/Geodesy/package.json
gchoqueux: this option isn't yet validated

package.json
gchoqueux: it's resolved

docs/config.json
gchoqueux: yes, i add exclude Pattern to exclude `lib` folder

packages/Geodesy/.babelrc
gchoqueux: the babel config file seems unnecessary, It keeps looking up the directory tree if it doesn't find it in the current folder.
I remove the file

packages/Geodesy/Gruntfile.cjs
gchoqueux: I simplify the bumping and i remove config file packages

packages/Geodesy/package.json
gchoqueux: changed

webpack.config.cjs
gchoqueux: yes, or specify other standard exports

webpack.config.cjs
gchoqueux: I think that eslint seems to need the paths to resolve

package.json
gchoqueux: in top level package.json

package.json
gchoqueux: no :/

package.json
gchoqueux: I'll try it

gchoqueux: yes

packages/Geodesy/.babelrc
gchoqueux: We keep all config files

packages/Geodesy/package.json
Desplandis: I'll try to propose a solution using only the `webpack` config =)

AnthonyGlt:
I'll we check it out once the PR is reopen, thanks for the work

Desplandis:
@gchoqueux I managed to correctly configure `babel` without having to resort to non-standard configurations in `webpack`. With my fixes, we just need to add an `alias` in the webpack configuration.

Shall I wait for further changes to push those commits?

P.S.: I even managed to convert one of the subpackages to `typescript` by changing its local `babel` configuration.

jailln (commented):
Thanks for this PR that lays the ground for dividing itowns in modules! I like that it can easily be implemented with npm workspaces and without any external library :+1: 

I have small remarks about the code but I will come back at it later. Now that we have a way to do it and a first separation, I think we should discuss alltogether (in an issue or in a voice call) the projected final separation in modules (all the modules we would like to create and what goes where); not in details but to get a general idea to avoid changing too much along the way.

gchoqueux:
note that the Debug package is in private

gchoqueux:
I propose a live presentation to explain this PR

gchoqueux:
Branch rebase on master
Webpack serve was fixed

jailln:
> I propose a live presentation to explain this PR

In my opinion the first thing to do is to agree together on the modules separation (i.e. on a final expected list of sub modules) and on the way / order in which it should be done to avoid going back and forth and so that anyone can contribute to this subject. We can discuss it all together in a call first if you want but I think it will need to end up in an issue eventually. WDYT?

Desplandis:
Yeah, I agree with @jailln and think we should start an issue/proposal where we could centralize all discussions on package splitting (use-cases, explanation of architectural/tooling changes, a sort of informal roadmap, ...). As long as we discuss those topics, I'm all in for a presentation.

@ftoromanoff, @mgermerie and @AnthonyGlt, I think you should participate too. =)

gchoqueux:
in this PR the modules follow the itowns structure
* **itowns** : already existing module
* **widget** : already in separate structure in `src/Utils/gui `and a specific bundle **itowns_widgets**, in this PR **this module is private** and it isn't published
* **debug** : already in separate structure in `utils/debug` and a specific bundle **debug**, in this PR **this module is private** and it isn't published 
* **geodesy** : new module  to measure and represente the geometry, gravity, and spatial orientation of the Earth in temporally varying 3D.

jailln:
> in this PR the modules follow the itowns structure
> 
> * **itowns** : already existing module
> * **widget** : already in separate structure in `src/Utils/gui `and a specific bundle **itowns_widgets**, in this PR **this module is private** and it isn't published
> * **debug** : already in separate structure in `utils/debug` and a specific bundle **debug**, in this PR **this module is private** and it isn't published
> * **geodesy** : new module  to measure and represente the geometry, gravity, and spatial orientation of the Earth in temporally varying 3D.

Ok thanks, this is fine by me. I would just rename `geodesy` to `geographic`, WDYT?
What we should discuss/list in an issue is what are the next steps (next modules) and in which order we should split them

gchoqueux:
`geodesy` includes gravity, orientation, coordinates. Even if the name is less well-known,  it's the right term used by professionals. I think it's worth communicating about these rare features.

gchoqueux:
I rebased this PR who could review it ?

Desplandis:
> Yeah, I agree with @jailln and think we should start an issue/proposal where we could centralize all discussions on package splitting (use-cases, explanation of architectural/tooling changes, a sort of informal roadmap, ...). As long as we discuss those topics, I'm all in for a presentation.

Once again, I reiterate the needs for a proper roadmap for package splitting before doing any kind of review (note that I can't do it nonetheless since I participated initially in the code of this PR). This is in our actual governance document to first open a proposal before validating any new functionality.

I know personally that there is some needs from IGN research team to split geographic package from the rest of itowns but we can't have an half-backed package splitting.

Finally, we have some recent PRs (refactoring and migration from js to ts) that are conflicting with the 300+ files changed (quite a huge PR in my opinion), so if you can't commit to open a discussion soon enough we'll surely merge them before this PR.

Pinging once again @jailln @AnthonyGlt @mgermerie @ftoromanoff.

jailln:
> o if you can't commit to open a discussion soon enough we'll surely merge them before this

This is a valuable PR that we would like to integrate as soon as possible but I agree with @Desplandis that, as said before, we should follow our [new contribution workflow](https://github.com/iTowns/itowns-governance/blob/master/proposal.md) because it is an impacting change. It boils down to opening a proposal issue describing the changes and proposing a plan (or a draft) to complete this migration (this PR being one part of it) and discussing it together so we all agree beforehand. This won't necessary take long and we can be reactive on this discussion so we can finish this first step quickly.

gchoqueux:
I open #2414

Monorepo_Roadmap.md
```diff
@@ -0,0 +1,53 @@
+# Monorepo Roadmap
```
Desplandis: We already have #2414 for tracking the monorepo roadmap. I'm not sure this is needed.

.gitignore
```diff
@@ -3,12 +3,12 @@
 # See https://help.github.com/articles/ignoring-files/ for how to configure your environment.
 # See https://github.com/github/gitignore/tree/master/Global for a list of global ignore rules.
 
-/node_modules/
-/dist/
+**/node_modules/
```
Desplandis: The `node_modules` is supposed to be at top-level.
```suggestion
node_modules/
```

Gruntfile.cjs
```diff
@@ -4,11 +4,21 @@ module.exports = (grunt) => {
         pkg: grunt.file.readJSON('package.json'),
         bump: {
             options: {
-                files: ['package.json', 'package-lock.json', 'src/Main.js'],
+                files: ['package.json', 'package-lock.json', 'packages/Main/src/Main.js',
+                    // 'packages/**/package.json'
```
Desplandis:
```suggestion
```

Gruntfile.cjs
```diff
@@ -4,11 +4,21 @@ module.exports = (grunt) => {
...
+                    'packages/Main/package.json',
+                    'packages/Geodesy/package.json',
+                    'packages/Widget/package.json',
+                ],
                 updateConfigs: [],
                 commit: true,
                 commitMessage: 'release v%VERSION%',
-                commitFiles: ['package.json', 'package-lock.json', 'src/Main.js'],
+                commitFiles: ['package.json', 'package-lock.json',
+                    // 'packages/**/package.json'
```
Desplandis:
```suggestion
```

Desplandis:
Hi, I'm currently reviewing your PR. I didn't had time to answer to the proposal. I'll submit both review and answer to the proposal tomorrow morning (TLDR; I agree with the package-splitting of this PR and have a few comments for future package-splitting).

package-lock.json
```diff
@@ -5,23 +5,14 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "itowns",
```
Desplandis, line 8: Is there a reason for such a huge diff in the `package-lock.json`? It seems a lot of added packages are not dependencies of `concurrently`.

package.json
```diff
@@ -1,43 +1,49 @@
...
-  },
+  "workspaces": [
+    "packages/Geodesy",
+    "packages/Main",
+    "packages/Debug",
+    "packages/Widget"
+  ],
   "scripts": {
-    "lint": "eslint \"src/**/*.{js,ts,tsx}\" \"test/**/*.js\" \"examples/**/*.js\" \"docs/*.js\"",
+    "test-unit": "npm run test-unit --workspaces --if-present",
```
Desplandis: Could you please use either `--workspaces` or `--ws`? This ease the understanding of the scripts.

package.json
```diff
@@ -1,43 +1,49 @@
...
     "doc": "jsdoc --readme docs/HOMEPAGE.md -c docs/config.json",
     "doclint": "npm run doc -- -t templates/silent",
-    "test": "npm run lint -- --max-warnings=0 && npm run build && npm run test-with-coverage && npm run test-functional",
-    "test-dev": "npm run lint -- --max-warnings=0 && npm run build-dev && npm run test-with-coverage && npm run test-functional",
-    "test-unit": "npm run base-test-unit test/unit",
-    "test-functional": "mocha -t 100000 --require test/hooks_functional.js --recursive test/functional",
-    "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
-    "test-with-coverage_lcov": "c8 -n src --reporter=lcov cross-env npm run test-unit",
-    "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --file test/unit/bootstrap.js --import=./config/babel-register/register.mjs",
+    "test": "npm run test-unit && npm run build && npm run test-functional",
```
Desplandis: Shall this be `npm run test-with-coverage` instead? The linting is missing too.

package.json
```diff
@@ -1,43 +1,49 @@
...
+    "test-dev": "npm run test-unit && npm run build-dev && npm run test-functional",
+    "test-functional": "npm run build && mocha -t 100000 --require test/hooks_functional.js --recursive test/functional",
+    "test-with-coverage": "npm run test-with-coverage --workspaces --if-present",
+    "clean-coverage": "node -e \"require('fs-extra').emptyDir('coverage');\"",
```
Desplandis: Not `coverage` directory are created in those scripts. I suppose this is some leftovers of previous attempts to merge coverages?
```suggestion
```

config/prepare.mjs
```diff
@@ -1,8 +1,5 @@
 import chalk from 'chalk';
```
Desplandis, line 1: Notes to all:
- This script is clearly outdated (some dev dependencies now require node > 18.0) and we could simply specify the requirement for both `node` and `npm` directly in the package.json with
```json
  "engines" : { 
    "node" : ">=18.0.0"
  }
```
- Since we now have an ESM distribution, I think we could remove the `ThreeExtended` script and directly imports from `three/addons/*`.

package.json
```diff
@@ -1,43 +1,49 @@
...
     "prepublishOnly": "npm run build && npm run transpile",
-    "prepare": "cross-env NO_UPDATE_NOTIFIER=true node ./config/prepare.mjs && node ./config/replace.config.mjs",
-    "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib",
+    "prepare": "cross-env NO_UPDATE_NOTIFIER=true node ./config/prepare.mjs && npm run prepare --ws --if-present",
+    "watch": "node ./config/concurrently.mjs watch",
     "changelog": "conventional-changelog -n ./config/conventionalChangelog/config.cjs -i changelog.md -s",
-    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm install && git add -A && git commit --amend --no-edit",
-    "publish-next": "npm version prerelease --preid next && npm publish --access public --tag=next --provenance",
-    "publish-latest": "npm publish --access public --tag=latest --provenance"
+    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm run update-packages && npm install && git add -A && git commit --amend --no-edit",
```
Desplandis, line 32: Why are you preferring a global script instead of a local one for each sub-package?

package.json
```diff
@@ -1,43 +1,49 @@
...
+    "publish-next": "npm run publish-next -ws --if-present && npm publish --access public --tag=next --ws --provenance",
+    "publish-latest": "npm publish --access public --tag=latest --provenance --ws",
```
Desplandis: Similar question, why `npm publish` is at toplevel?

packages/Debug/README.md
```diff
@@ -0,0 +1,7 @@
+## iTowns Debug
+
+# Install
+
+`npm install --save @itowns/debug`
```
Desplandis: Since this is a private package, I don't think we should provide install instructions.

packages/Debug/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+It's module for itowns debugging
```
Desplandis:
```suggestion
iTowns debugging utilities.
```

gchoqueux:
for the moment I haven't completely correlated the proposal with the PR.  

For example:
in PR : `@itowns/main` -> `@itowns/main` in proposal 
in PR : `@itowns/debug` -> `@itowns/plugins/debug` in proposal

packages/Debug/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "main": "lib/Main.js",
+  "exports": {
+    ".": "./lib/Main.js"
+  },
+  "scripts": {
+    "lint": "eslint \"src/**/*.{js,ts,tsx}\"",
+    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
+    "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib"
+  },
+  "files": [
```
Desplandis, line 16: A lot of dependencies are missing. A (non-exhaustive) list:
- `itowns`
- `chart.js`
Are you also sure that there is no global referenced?

packages/Debug/src/Debug.js
```diff
@@ -1,7 +1,6 @@
 import { CameraHelper, Color, Vector3 } from 'three';
-import Coordinates from 'Core/Geographic/Coordinates';
-import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
-import OBB from 'Renderer/OBB';
+import { Coordinates } from '@itowns/geodesy';
+import { MAIN_LOOP_EVENTS, OBB } from 'itowns';
```
Desplandis, line 3: The `OBB` module was not exposed as public API until now. Are we not exposing an implementation detail only for this package?

packages/Debug/src/TileDebug.js
```diff
@@ -1,8 +1,6 @@
 import * as THREE from 'three';
 import TWEEN from '@tweenjs/tween.js';
-import GeometryLayer from 'Layer/GeometryLayer';
-import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
-import ObjectRemovalHelper from 'Process/ObjectRemovalHelper';
+import { MAIN_LOOP_EVENTS, GeometryLayer, ObjectRemovalHelper } from 'itowns';
```
Desplandis, line 3: The `ObjectRemovalHelper` module was not exposed as public API until now. Are we not exposing an implementation detail only for this package?

packages/Geodesy/.babelrc
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "plugins": [
+        ["module-resolver", {
+            "cwd": "packagejson",
+            "root": ["./src"],
+            "extensions": [".js", ".ts", ".tsx"]
+        }],
+        ["module-extension-resolver", {
+            "srcExtensions": [".ts", ".js"]
+        }],
```
Desplandis: I think we could have removed one (or both) of those plugins if you had used relative imports on the `geodesy` sub-module.

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
+## iTowns Geodesy
```
Desplandis: My preference would be to name it `Geographic`, keeping the old naming.

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Geodesy is the science of measuring and representing the geometry, gravity, and spatial orientation of the Earth in temporally varying 3D.
+
+* [Coordinates](http://www.itowns-project.org/itowns/docs/#api/Geographic/Coordinates) : A Coordinates object (geodetic datum), defined by a [crs] and three values.
+* [Extent](http://www.itowns-project.org/itowns/docs/#api/Geographic/Extent) : Extent is geographical bounding rectangle defined by 4 limits: west, east, south and north. If crs is tiled projection (WMTS or TMS), the extent is defined by zoom, row and column.
+* [Crs](http://www.itowns-project.org/itowns/docs/#api/Geographic/CRS) : This module provides basic methods to manipulate a CRS (as a string). Visiting [epsg.io](https://epsg.io/) to more coordinate system worldwide. Use **Proj4js** definition in epsg.io.
+* [Geogrid](http://www.itowns-project.org/itowns/docs/#api/Geographic/GeoidGrid) : An instance of GeoidGrid allows accessing some geoid height grid data from Coordinates.
```
Desplandis: Since this is kind of a patch due to the limitation of our current elevation implementation, I don't think there is a lot of value to expose it in the `geodesy` module. We will remove it once this limitation is removed.

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
+## iTowns Geodesy
+
+Geodesy is the science of measuring and representing the geometry, gravity, and spatial orientation of the Earth in temporally varying 3D.
```
Desplandis: A short description of the package (what it does) would be better.

An example:
```suggestion
The geodesy package provides utilities for handling coordinates, ellipsoids, extents and rotations across different coordinate systems.
```

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+// Access a value of geoid height at some geographic coordinates.
+// The value is interpolated from the gridded data.
+const value = geoidGrid.getHeightAtCoordinates(
+    new Coordinates('EPSG:4326', 1.5, 40.25)
+);
+// This should return 2.0, which is the result from the bi-linear
+// interpolation at the center of the `[[1, 2], [2, 3]]` subsection
+// of the grid data.
+```
```
Desplandis: Could be removed (see comment above).

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## OrientationUtils example 
+```js
+// Compute the rotation around the point of origin from a frame aligned with Lambert93 axes (epsg:2154),
+// to the geocentric frame (epsg:4978)
+quat_crs2crs = OrientationUtils.quaternionFromCRSToCRS("EPSG:2154", "EPSG:4978")(origin);
+// Compute the rotation of a sensor platform defined by its attitude
+quat_attitude = OrientationUtils.quaternionFromAttitude(attitude);
```
Desplandis: I don't understand with the current documentation what is an "attitude", could you please explicit it?

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
...
+// Compute the rotation from the sensor platform frame to the geocentric frame
+quat = quat_crs2crs.multiply(quat_attitude);
+```
+
+Visiting iTowns [documentation](http://www.itowns-project.org/itowns/docs/#home) for more information.
```
Desplandis:
```suggestion
Visit the iTowns [documentation](http://www.itowns-project.org/itowns/docs/#home) for more information.
```

packages/Geodesy/src/Ellipsoid.js
```diff
@@ -1,6 +1,6 @@
 import * as THREE from 'three';
 import proj4 from 'proj4';
-import Coordinates from 'Core/Geographic/Coordinates';
+import Coordinates from 'Coordinates';
```
Desplandis: All modules are on the same directory, why not using relative imports?

packages/Debug/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }],
+        ["babel-plugin-inline-import", {
+            "extensions": [
+                ".json",
+                ".geojson",
+                ".glsl",
+                ".css"
+            ]
+        }],
```
Desplandis: We don't import any files with those extensions, this plugin is more useful.

packages/Debug/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+        ["minify-replace", {
+            "replacements": [{
+                "identifierName": "__DEBUG__",
+                "replacement": {
+                    "type": "booleanLiteral",
+                    "value": false
+                }
+            }]
+        }],
```
Desplandis: No use of the symbol `__DEBUG__`, it could be safely deleted.

packages/Debug/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+        ["minify-dead-code-elimination"]
```
Desplandis, line 23: We usually use this plugin to prune branches behind a `__DEBUG__` predicate, is this still useful?

packages/Debug/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    "plugins": [
+        ["module-resolver", {
+            "cwd": "packagejson",
+            "root": ["./src"],
+            "extensions": [".js", ".ts", ".tsx"],
+            "alias": {
+                "@itowns/geodesy": "../Geodesy/src/Main.js",
+                "itowns": "../Main/src/Main.js"
+            }
```
Desplandis: Those alias lead to incorrect compilation of import paths.

For example, the code
```js
import { ... } from 'itowns';
```
is transpiled into:
```js
import { ... } from "../../Main/src/Main.js";
```

You should consider removing them.

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
+## iTowns Geodesy
+
+Geodesy is the science of measuring and representing the geometry, gravity, and spatial orientation of the Earth in temporally varying 3D.
+
+* [Coordinates](http://www.itowns-project.org/itowns/docs/#api/Geographic/Coordinates) : A Coordinates object (geodetic datum), defined by a [crs] and three values.
+* [Extent](http://www.itowns-project.org/itowns/docs/#api/Geographic/Extent) : Extent is geographical bounding rectangle defined by 4 limits: west, east, south and north. If crs is tiled projection (WMTS or TMS), the extent is defined by zoom, row and column.
```
Desplandis: Since #2412 this is no longer correct.

```suggestion
* [Extent](http://www.itowns-project.org/itowns/docs/#api/Geographic/Extent) : Extent is geographical bounding rectangle defined by 4 limits: west, east, south and north.
```

packages/Geodesy/README.md
```diff
@@ -0,0 +1,84 @@
...
+* [Crs](http://www.itowns-project.org/itowns/docs/#api/Geographic/CRS) : This module provides basic methods to manipulate a CRS (as a string). Visiting [epsg.io](https://epsg.io/) to more coordinate system worldwide. Use **Proj4js** definition in epsg.io.
+* [Geogrid](http://www.itowns-project.org/itowns/docs/#api/Geographic/GeoidGrid) : An instance of GeoidGrid allows accessing some geoid height grid data from Coordinates.
+* [OrientationUtils](http://www.itowns-project.org/itowns/docs/#api/Geographic/OrientationUtils) : it provides methods to compute the quaternion that models a rotation defined with various conventions, including between different CRS.
+* Ellipsoid : In geodesy, a reference ellipsoid is a mathematically defined surface that approximates the geoid.
```
Desplandis: No reference to documentation?

packages/Geodesy/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "scripts": {
+    "build": "",
+    "lint": "eslint \"src/**/*.{js,ts,tsx}\" \"test/**/*.js\"",
+    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
+    "test-unit": "npm run base-test-unit test/unit",
+    "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --import=../../config/babel-register/register.mjs",
+    "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
+    "test-with-coverage_lcov": "c8 -n src --reporter=lcov cross-env npm run test-unit",
+    "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib",
+    "echo": "echo $npm_package_name@$npm_package_version",
```
Desplandis: Is there a reason to add this new script?

packages/Geodesy/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "publish-next": "npm version prerelease --preid next"
```
Desplandis: I'm kinda confused, there is both a `publish` script at toplevel and a `publish-next` in a sub-package, shall they not be on the same level?

packages/Main/.babelrc.cjs
```diff
@@ -0,0 +1,52 @@
+const BABEL_ENV = process.env.BABEL_ENV;
+
+const alias_itowns_geodesy = `../Geodesy/${BABEL_ENV == 'test' ? 'lib' : 'src'}/Main.js`;
```
Desplandis: Big nope. We can't distribute package where imports are transpiled to relative path resolving to outside the package...

This
```js
import { Coordinates } from '@itowns/geodesy';
```
is compiled into
```js
import { Coordinates } from "../../../Geodesy/src/Main.js";
```

Moreover, our pipeline is already a mess, adding conditional compilation with `BABEL_ENV` will not ease this...

packages/Main/.babelrc.cjs
```diff
@@ -0,0 +1,52 @@
...
+            }]
+        ],
+        plugins: [
+            ['module-resolver', {
+                cwd: 'packagejson',
+                root: ['./src'],
+                extensions: [".js", ".ts", ".tsx"],
+                alias: {
+                    '@itowns/geodesy': alias_itowns_geodesy
+                }
```
Desplandis: See comment above.

packages/Main/clean.cjs
```diff
@@ -0,0 +1,5 @@
+const fs = require('node:fs');
```
Desplandis, line 1: Why do we have a cleaning script since `examples`, `docs` are at top-level. This is out-of-scope of this PR.

packages/Main/src/Core/Prefab/GlobeView.js
```diff
@@ -2,15 +2,13 @@ import * as THREE from 'three';
 
 import View, { VIEW_EVENTS } from 'Core/View';
 import GlobeControls from 'Controls/GlobeControls';
-import Coordinates from 'Core/Geographic/Coordinates';
+import { Coordinates, CRS, ellipsoidSizes } from '@itowns/geodesy';
 
-import GlobeLayer from 'Core/Prefab/Globe/GlobeLayer';
+// eslint-disable-next-line import/extensions
+import GlobeLayer from 'Core/Prefab/Globe/GlobeLayer.js';
```
Desplandis:
```suggestion
import GlobeLayer from 'Core/Prefab/Globe/GlobeLayer';
```

packages/Main/src/Main.js
```diff
@@ -1,22 +1,20 @@
...
 
-export { default as Ellipsoid, ellipsoidSizes } from 'Core/Math/Ellipsoid';
 export { default as GlobeView, GLOBE_VIEW_EVENTS } from 'Core/Prefab/GlobeView';
 export { default as PlanarView } from 'Core/Prefab/PlanarView';
 export { default as Fetcher } from 'Provider/Fetcher';
 export { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
 export { default as View } from 'Core/View';
 export { VIEW_EVENTS } from 'Core/View';
 export { default as FeatureProcessing } from 'Process/FeatureProcessing';
+export { default as ObjectRemovalHelper } from 'Process/ObjectRemovalHelper';
```
Desplandis, line 18: Is there a reason to expose `ObjectRemovalHelper` which is an implementation detail. I think this is out of scope of this PR.

packages/Main/src/Main.js
```diff
@@ -31,11 +29,11 @@ export { default as Feature2Mesh } from 'Converter/Feature2Mesh';
 export { default as FeaturesUtils } from 'Utils/FeaturesUtils';
 export { default as DEMUtils } from 'Utils/DEMUtils';
 export { default as CameraUtils } from 'Utils/CameraUtils';
-export { default as OrientationUtils } from 'Utils/OrientationUtils';
 export { default as ShaderChunk } from 'Renderer/Shader/ShaderChunk';
 export { getMaxColorSamplerUnitsCount, colorLayerEffects } from 'Renderer/LayeredMaterial';
 export { default as Capabilities } from 'Core/System/Capabilities';
 export { CAMERA_TYPE } from 'Renderer/Camera';
+export { default as OBB } from 'Renderer/OBB';
```
Desplandis, line 37: Is there a reason to expose `OBB` which is an implementation detail. I think this is out of scope of this PR.

packages/Main/src/Main.js
```diff
@@ -114,3 +112,4 @@ export { default as C3DTExtensions } from './Core/3DTiles/C3DTExtensions';
 export { C3DTilesTypes, C3DTilesBoundingVolumeTypes } from './Core/3DTiles/C3DTilesEnums';
 export { default as C3DTBatchTableHierarchyExtension } from './Core/3DTiles/C3DTBatchTableHierarchyExtension';
 export { process3dTilesNode, $3dTilesCulling, $3dTilesSubdivisionControl } from 'Process/3dTilesProcessing';
+
```
Desplandis:
```suggestion
```

packages/Main/src/Renderer/Label2DRenderer.js
```diff
@@ -1,5 +1,6 @@
 import * as THREE from 'three';
-import GlobeLayer from 'Core/Prefab/Globe/GlobeLayer';
+// eslint-disable-next-line import/extensions
+import GlobeLayer from 'Core/Prefab/Globe/GlobeLayer.js';
```
Desplandis:
```suggestion
import GlobeLayer from 'Core/Prefab/Globe/GlobeLayer';
```

packages/Main/test/unit/entwine.js
```diff
@@ -13,6 +13,11 @@ import Renderer from './bootstrap';
 import ept from '../data/entwine/ept.json';
 import eptHierarchy from '../data/entwine/ept-hierarchy/0-0-0-0.json';
 
+// const path = require('path');
+
+// const las_path = path.resolve(__dirname, 'examples/libs/laz-perf');
+
+
```
Desplandis:
```suggestion
```

packages/Widget/package.json
```diff
@@ -0,0 +1,30 @@
+{
```
Desplandis: You are introducing a breaking change in our npm distribution for users of `itowns/widgets`. Moreover, since this package is private, there is no alternative for those users in the current proposal.

packages/Widget/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "private": true,
+  "version": "2.44.2",
+  "description": "Widget",
+  "type": "module",
+  "main": "lib/Main.js",
+  "exports": {
+    ".": "./lib/Main.js"
+  },
+  "scripts": {
```
Desplandis: This package is missing the `itowns` dependency.

tsconfig.json
```diff
@@ -1,13 +1,13 @@
 {
```
Desplandis, line 1: There should be one `tsconfig.json` per sub-package. You could use the `extends` keyword (similar to eslint) to share common configuration.

webpack.config.cjs
```diff
@@ -23,8 +23,10 @@ const babelConf = {
 };
 
 const include = [
-    path.resolve(__dirname, 'src'),
-    path.resolve(__dirname, 'utils'),
+    path.resolve(__dirname, 'packages/Geodesy/src'),
+    path.resolve(__dirname, 'packages/Main/src'),
+    path.resolve(__dirname, 'packages/Debug/src'),
+    path.resolve(__dirname, 'packages/Widget/src'),
 ];
```
Desplandis, line 31: Why not using webpack alias instead of hacking your way with the babel configuration? [This](https://github.com/Desplandis/itowns/blob/subpackages-geography/webpack.config.cjs) was working when I worked on this PR last june.

.babelrc
```diff
@@ -10,7 +10,8 @@
     ],
```
Desplandis, line 10: Is there a reason to keep this `.babelrc` since all code and tests are on the sub-packages?

packages/Main/.eslintignore
```diff
@@ -0,0 +1,14 @@
+examples/libs/
+examples/layers/
```
Desplandis: Examples are at top-level.

```suggestion
```

packages/Main/.eslintignore
```diff
@@ -0,0 +1,14 @@
...
+src/ThreeExtended/
+test/hooks.js
+dist/
+node_modules/
+docs/out/
+docs/static/
+docs/tutorials/
+docs/tmpl/
```
Desplandis: Docs are at top-level.

```suggestion
```

packages/Main/.eslintignore
```diff
@@ -0,0 +1,14 @@
...
+test/data/
+
+
+
```
Desplandis:
```suggestion
```

packages/Main/package.json
```diff
@@ -0,0 +1,72 @@
...
+  "scripts": {
+    "lint": "eslint \"src/**/*.{js,ts,tsx}\" \"test/**/*.js\"",
+    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
+    "test-unit": "cross-env BABEL_ENV=test npm run base-test-unit test/unit",
+    "base-test-unit": "cross-env BABEL_DISABLE_CACHE=1 mocha --timeout 5000 --file test/unit/bootstrap.js --import=../../config/babel-register/register.mjs",
+    "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
+    "test-with-coverage_lcov": "c8 -n src --reporter=lcov cross-env npm run test-unit",
+    "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib",
+    "postinstall": "cross-env NO_UPDATE_NOTIFIER=true node ./scripts/prepare.mjs && node ./scripts/replace.config.mjs",
+    "prepublishOnly": "npx copyfiles -u 1 \"../../examples/**/*\" ./examples/ && npx copyfiles -u 1 \"../../docs/**/*\" ./docs/ && npx copyfiles -u 1 \"../../dist/**/*\" ./dist/ ",
```
Desplandis, line 19: Or just not distribute examples and docs like most other packages.

packages/Widget/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    "plugins": [
+        ["module-resolver", {
+            "cwd": "packagejson",
+            "root": ["./src"],
+            "extensions": [".js", ".ts", ".tsx"],
+            "alias": {
+                "@itowns/geodesy": "../Geodesy/src/Main.js",
+                "itowns": "../Main/src/Main.js"
+            }
```
Desplandis: Once again this leads to incorrect compilation of path.

packages/Widget/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+        }],
+        ["babel-plugin-inline-import", {
+            "extensions": [
+                ".json",
+                ".geojson",
+                ".glsl",
+                ".css"
+            ]
+        }],
```
Desplandis: Not necessary since we do not imports from file with those extensions.

packages/Widget/.babelrc
```diff
@@ -0,0 +1,46 @@
...
+        ["minify-replace", {
+            "replacements": [{
+                "identifierName": "__DEBUG__",
+                "replacement": {
+                    "type": "booleanLiteral",
+                    "value": false
+                }
+            }]
+        }],
+        ["minify-dead-code-elimination"]
```
Desplandis: Same here, unused.

Desplandis (changes requested):
Hi! This is first batch of my review of this PR. Unfortunately I'm kinda blocked on testing the distribution since the compiled code for the `itowns` package is incorrect (especially with imports to relative modules).

Moreover, there is a lot of points that I would like you to address. I'll resume the review once those points have been resolved.

Since this PR is massive, @jailln there may be some things that eluded me, do not hesitate to complement this review. =)

package.json
gchoqueux: yes indeed

package.json
gchoqueux: ok I use only `workspaces` option

package.json
gchoqueux: Removed

config/prepare.mjs
gchoqueux: i open  #2462

package.json
gchoqueux: Because Grunt updates all sub-packages version then `update-packages` script updates all packages with `workspaces`  option.

package.json
gchoqueux: `publish-next` and  `publish-latest` use local script 
I'll try to move options on local scripts

packages/Debug/README.md
gchoqueux: Removed

.babelrc
gchoqueux: removed

packages/Debug/package.json
gchoqueux: I added dependency but webpack isn't explicit in missing packages

packages/Debug/src/Debug.js
gchoqueux: yes

packages/Debug/src/TileDebug.js
gchoqueux: yes

packages/Geodesy/.babelrc
gchoqueux: What is the chosen rule in iTowns for import paths?
It should be added to the linter if it isn’t already there.

packages/Geodesy/README.md
gchoqueux: Could you explain what limitation you're talking about?

packages/Geodesy/README.md
gchoqueux: it doesn't exist yet

packages/Geodesy/README.md
gchoqueux: removed

packages/Geodesy/README.md
gchoqueux: In geodesy, attitude refers to the orientation of a geodetic instrument or platform in three-dimensional space. It is defined by the angles that describe how an instrument or a vehicle (like a satellite or aircraft) is positioned relative to a reference coordinate system, typically the Earth's surface or a local tangent plane.

packages/Geodesy/package.json
gchoqueux: removed

packages/Geodesy/package.json
gchoqueux: `publish-latest` toplevel calls default `publish` in sub-packages, that's why there's no `publish-latest` in sub-packages

packages/Geodesy/src/Ellipsoid.js
gchoqueux: it would be necessary to define an eslint rule

packages/Main/clean.cjs
gchoqueux: this script is necessary to publish itowns package with documentation and examples from toplevel

packages/Main/package.json
gchoqueux: to be decided

packages/Main/src/Core/Prefab/GlobeView.js
gchoqueux: applied

packages/Main/src/Main.js
gchoqueux: Debug package import ObjectRemovalHelper from itowns package

packages/Main/src/Main.js
gchoqueux: same reason

packages/Widget/.babelrc
gchoqueux: removed

packages/Widget/.babelrc
gchoqueux: removed

packages/Widget/package.json
gchoqueux: I fixed the breaking change with a temporary copy of transpiled files

packages/Widget/package.json
gchoqueux: Added

tsconfig.json
gchoqueux: I added a tsconfig.json per sub-package but I don't know how to test them

webpack.config.cjs
gchoqueux: include constant is to set the source code for Babel ans Eslint and not for resolve the dependencies

packages/Geodesy/README.md
gchoqueux: removed

packages/Debug/.babelrc
gchoqueux: removed

packages/Debug/.babelrc
gchoqueux: removed

packages/Debug/.babelrc
gchoqueux: i don't know

packages/Debug/.babelrc
gchoqueux: removed

packages/Main/.babelrc.cjs
gchoqueux: removed

packages/Main/.babelrc.cjs
gchoqueux: removed

packages/Main/.eslintignore
gchoqueux: applied

gchoqueux:
> Hi! This is first batch of my review of this PR. Unfortunately I'm kinda blocked on testing the distribution since the compiled code for the `itowns` package is incorrect (especially with imports to relative modules).
> 
> Moreover, there is a lot of points that I would like you to address. I'll resume the review once those points have been resolved.
> 
> Since this PR is massive, @jailln there may be some things that eluded me, do not hesitate to complement this review. =)

@Desplandis 

I've finished answering/fixing all your comments.
I've also corrected the transpiling for the tests.

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
+## iTowns Geographic
```
jailln, line 1: Just a thought: we may update the doc template to be structured per subpackage and export this page as the welcome page of the subpackage. WDYT @gchoqueux @Desplandis?

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The geographic package provides utilities for handling coordinates, ellipsoids, extents and rotations across different coordinate systems.
+
+* [Coordinates](http://www.itowns-project.org/itowns/docs/#api/Geographic/Coordinates) : A Coordinates object (geodetic datum), defined by a [crs] and three values.
+* [Extent](http://www.itowns-project.org/itowns/docs/#api/Geographic/Extent) : Extent is geographical bounding rectangle defined by 4 limits: west, east, south and north.
+* [Crs](http://www.itowns-project.org/itowns/docs/#api/Geographic/CRS) : This module provides basic methods to manipulate a CRS (as a string). Visiting [epsg.io](https://epsg.io/) to more coordinate system worldwide. Use **Proj4js** definition in epsg.io.
+* [OrientationUtils](http://www.itowns-project.org/itowns/docs/#api/Geographic/OrientationUtils) : it provides methods to compute the quaternion that models a rotation defined with various conventions, including between different CRS.
+* Ellipsoid : In geodesy, a reference ellipsoid is a mathematically defined surface that approximates the geoid.
```
jailln: I'm not sure we should add all classes here as they already each have their own doc? WDYT @Desplandis? If we keep this list, we should add all classes though

package.json
```diff
@@ -1,43 +1,48 @@
...
-  },
+  "workspaces": [
+    "packages/Geographic",
+    "packages/Main",
+    "packages/Debug",
+    "packages/Widget"
+  ],
   "scripts": {
-    "lint": "eslint \"src/**/*.{js,ts,tsx}\" \"test/**/*.js\" \"examples/**/*.js\" \"docs/*.js\"",
+    "test-unit": "npm run test-unit --workspaces --if-present",
```
jailln: can you preserve alphabetical order please ? It was easier to read

jailln (commented):
Not much to add to @Desplandis review. 

I just think we should explain this new structure in the README and add sub package installation possibilities.

Also, how does it work with releasing and versioning? does subpackage all have the same version or can we (should we?) release them independently ?

packages/Widget/.babelrc
gchoqueux: removed

tsconfig.json
gchoqueux: I rebase on typescript migrating and it seems to works

gchoqueux:
> I just think we should explain this new structure in the README and add sub package installation possibilities.

Yes, I'll document

> Also, how does it work with releasing and versioning? does subpackage all have the same version or can we (should we?) release them independently ?

In a monorepo, it's more complicated to manage a different version for each module, but if we can find a simple and robust solution we can implement it later.

packages/Geographic/README.md
gchoqueux: yes, I think it would be recommended for future implementations to structure documentation by sub-module.

packages/Geographic/README.md
gchoqueux: I think it's a good idea to introduce the module's features and point to the detailed documentation. 

 > If we keep this list, we should add all classes though

Except for the Main sub-module, which has yet to be divided.

package.json
gchoqueux: I rearranged it as on the master

packages/Geodesy/.babelrc
Desplandis: For now, this is kind of a mixed bag... A lot of imports are absolute and later transformed to relative with babel. But there exists some modules which uses relative imports. IMO I think this is kinda "overkill" to do it for small packages like geodesy.

packages/Geodesy/src/Ellipsoid.js
Desplandis: Agree.

packages/Geographic/README.md
Desplandis: It seems that typedoc support such configuration (see https://github.com/Gerrit0/typedoc-packages-example). I will replace tsdoc with typedoc in a following PR.

packages/Geographic/README.md
Desplandis: For such a small package, we can explicit the different kind of modules. But I don't think this could scale-up to other big packages.

Desplandis:
@gchoqueux Did you try to publish it in your own `npm` packages? I tried to do it with a custom alias, but didn't manage to make it working...

This is the a log of my last attempt:
https://github.com/Desplandis/itowns/actions/runs/12121853265/job/33793853058

The CI works correctly but I would like to ensure that both npm and website publishing are working.

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
+## iTowns Geographic
+
+The geographic package provides utilities for handling coordinates, ellipsoids, extents and rotations across different coordinate systems.
+
+* [Coordinates](http://www.itowns-project.org/itowns/docs/#api/Geographic/Coordinates) : A Coordinates object (geodetic datum), defined by a [crs] and three values.
+* [Extent](http://www.itowns-project.org/itowns/docs/#api/Geographic/Extent) : Extent is geographical bounding rectangle defined by 4 limits: west, east, south and north.
```
jailln: can you add: `and a crs` ?

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
+## iTowns Geographic
+
+The geographic package provides utilities for handling coordinates, ellipsoids, extents and rotations across different coordinate systems.
+
+* [Coordinates](http://www.itowns-project.org/itowns/docs/#api/Geographic/Coordinates) : A Coordinates object (geodetic datum), defined by a [crs] and three values.
```
jailln, line 5: can you replace `[crs]` by `Coordinate Reference System (crs)`

jailln, line 5: Can you organize this list alphabetically please? and add missing classes and variables (CoordStars and ellipsoidSizes). BTW I would for renaming ellipsoidSizes to WGS84EllipsoidSize while we're at it.

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
...
+* [Extent](http://www.itowns-project.org/itowns/docs/#api/Geographic/Extent) : Extent is geographical bounding rectangle defined by 4 limits: west, east, south and north.
+* [Crs](http://www.itowns-project.org/itowns/docs/#api/Geographic/CRS) : This module provides basic methods to manipulate a CRS (as a string). Visiting [epsg.io](https://epsg.io/) to more coordinate system worldwide. Use **Proj4js** definition in epsg.io.
```
jailln: `Visiting [epsg.io](https://epsg.io/) to more coordinate system worldwide. Use **Proj4js** definition in epsg.io.` -> Visit [epsg.io](https://epsg.io/) to explore available coordinates systems and use their proj4js definition to define them in itowns with `proj4.defs(crs, proj4def)`

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
...
+* [OrientationUtils](http://www.itowns-project.org/itowns/docs/#api/Geographic/OrientationUtils) : it provides methods to compute the quaternion that models a rotation defined with various conventions, including between different CRS.
```
jailln: `utilities to compute a rotation quaternion from various rotation conventions, including between different crs.`

packages/Geographic/README.md
jailln: proposition: `A representation of an [ellipsoid](https://en.wikipedia.org/wiki/Ellipsoid) and useful computation methods (geodetic normal, etc.)`

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
...
+const extent = new Extent('EPSG:4326', 88.002445, 50.336522, 22.021, 50.302548);
+
+// change projection system to pseudo mercator
+
+coordinates.as('EPSG:3857');
+extent.as('EPSG:3857');
+
+// change projection system to EPSG:2154
+
+// defs EPSG:2154 crs (visiting epsg.io to get new definition and export to Proj4js)
```
jailln: - visiting -> visit
- to get new definition and export to Proj4js : to get the proj4js definition of a crs

packages/Geographic/README.md
```diff
@@ -0,0 +1,51 @@
...
+## OrientationUtils example 
+
+In geodesy, attitude refers to the orientation of a geodetic instrument or platform in three-dimensional space. It is defined by the angles that describe how an instrument or a vehicle (like a satellite or aircraft) is positioned relative to a reference coordinate system, typically the Earth's surface or a local tangent plane.
+
+```js
+// Compute the rotation around the point of origin from a frame aligned with Lambert93 axes (epsg:2154),
+// to the geocentric frame (epsg:4978)
+quat_crs2crs = OrientationUtils.quaternionFromCRSToCRS("EPSG:2154", "EPSG:4978")(origin);
+// Compute the rotation of a sensor platform defined by its attitude
+quat_attitude = OrientationUtils.quaternionFromAttitude(attitude);
```
jailln, line 60: thanks for these examples :) can you define a mock attitude above to ease understanding for newbies please?

README.md
```diff
@@ -58,10 +58,48 @@ github release page.
 
 In your project:
 
+To use all iTowns features, install `itowns` package :
+
 ```bash
 npm install --save itowns
 ```
 
+```js
```
jailln, line 67: I propose to replace by something more generic and that will be easier to update when new packages are added and to remove examples since we already have examples in each sub-package documentation:

```
iTowns is currently moving to a monorepo organization and to a segmentation in sub-modules, allowing to import only some of itowns functionalities. Current itowns sub-modules are:
- [@itowns/geographic](packages/Geographic/README.md): `npm install --save @itowns/geographic`
- [@itowns/widgets](packages/Widgets/README.md): `npm install --save @itowns/widgets`
- [@itowns/core](packages/Main/README.md); `npm install --save @itowns/core`
```

Note the s at the end of `@itowns/widgets` since there is more: can you update this package name and also the folder name: packages/Widget -> packages/Widgets

packages/Main/README.md
```diff
@@ -0,0 +1,131 @@
+![iTowns](https://raw.githubusercontent.com/iTowns/itowns.github.io/master/images/itowns_logo_300x134.png)
```
jailln: I think we should not copy a part of the parent README here but just explains that is contains all other itowns functionalities and link to the parent README of the repo and to the doc instead.

jailln (commented):
Last batch of comments for me, only documentation remarks, I let you see with @Desplandis for package publishing. Thanks for the changes you already made

packages/Geographic/README.md
gchoqueux: yes

packages/Geographic/README.md
gchoqueux: ok

gchoqueux: I organize and doc
yes, `ellipsoidSizes` could be renamed in other PR

packages/Geographic/README.md
gchoqueux: ok

packages/Geographic/README.md
gchoqueux: ok

packages/Geographic/README.md
gchoqueux: ok

packages/Geographic/README.md
gchoqueux: ok

packages/Geographic/README.md
gchoqueux: added

README.md
gchoqueux: I rename  `packages/Widget` to `packages/Widgets`

@itowns/widgets is private and  @itowns/core doesn't exist for the moment. I could remove the sub-modules examples ?

packages/Main/README.md
gchoqueux: cleaned

README.md
jailln: thanks! Ah yes ok sorry I thought we chose to make it public in this PR and didn't look if it was the case. Making it public now means also improving the doc, so I guess we can do that in a subsequent PR so we can merge this PR quicker

Ok for `@itowns/core`, my bad. So yes, only that is sufficient in my opinion:
```
iTowns is currently moving to a monorepo organization and to a segmentation in sub-modules, allowing to import only some of itowns functionalities. Current itowns sub-modules are:
- [@itowns/geographic](packages/Geographic/README.md): `npm install --save @itowns/geographic`
```

gchoqueux:
> This is the a log of my last attempt: https://github.com/Desplandis/itowns/actions/runs/12121853265/job/33793853058

You need to register the packages, it seems to manage dependencies as a single package. I think I used a publish-latest in local to register all subpackages.


The deploying an releasing work [see jobs](https://github.com/gchoqueux/itowns/actions/runs/12200671017)
but there's a error for the publishing [`npm error Invalid tag name "@latest" of package "@latest": Tags may not have any characters that encodeURIComponent encodes.`](https://github.com/gchoqueux/itowns/actions/runs/12200671017/job/34037720652#step:7:210)

I'm trying to solve this problem, and I think I've got a lead.

README.md
gchoqueux: ok

Desplandis:
@gchoqueux Did you fix it?

gchoqueux:
@Desplandis 
I fix integration github actions,
we're getting closer to our goal 
@iTowns/core-devs : It's the moment to try hard this PR before merging

Desplandis:
Okay, so you managed to test the publishing of all packages?
If so, I'll start re-re-reviewing next week. x)

gchoqueux:
> Okay, so you managed to test the publishing of all packages?

Yes, the PR manages all packages, next packages

Desplandis:
Thanks @gchoqueux. I managed to successfully publish both npm packages (new release and next) as well as testing the release on github/itowns.github.io. So it's all good to me, ready to merge ! We'll just need to change the coveralls target to the new one (which is a repo settings).

Desplandis (approved):
Okay, let's go!

PR #2303, opened 2024-03-22
feat: introducing workers for LAS parser

Desplandis:
## Description

This PR introduces web workers in iTowns, at first in the LAS parser which greatly benefits from background parsing.

## Motivation and Context
TODO

TODO: Performance screenshots

### DIFF
--- a/src/Parser/LASParser.js
+++ b/src/Parser/LASParser.js
@@ -1,7 +1,23 @@
 import * as THREE from 'three';
-import LASLoader from 'Loader/LASLoader';
+import { spawn, Thread, Transfer } from 'threads';
 
-const lasLoader = new LASLoader();
+let _lazPerf;
+let _thread;
+
+function workerInstance() {
+    return new Worker(
+        /* webpackChunkName: "itowns_lasparser" */
+        new URL('../Worker/LASLoaderWorker.js', import.meta.url),
+        { type: 'module' },
+    );
+}
+
+async function loader() {
+    if (_thread) { return _thread; }
+    _thread = await spawn(workerInstance());
+    if (_lazPerf) { _thread.lazPerf(_lazPerf); }
+    return _thread;
+}
 
 function buildBufferGeometry(attributes) {
     const geometry = new THREE.BufferGeometry();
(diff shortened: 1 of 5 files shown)


src/Parser/LASParser.js
```diff
@@ -1,7 +1,24 @@
...
+const lasWorkerLoader = {
+    async parseFile(data, options) {
+        if (!this.pool) {
+            const initWorker = () =>
+                new Worker(
+                    /* webpackChunkName: "itowns_lasparser" */
+                    new URL('../Worker/LASWorker.js', import.meta.url),
+                );
+            this.pool = new Pool(async () => spawn(initWorker()), {
+                size: 1,
```
jailln: Do we really need a pool if size is 1 ? Or should we make size configurable ?

src/Worker/LASWorker.js
```diff
@@ -0,0 +1,14 @@
+import LASLoader from 'Parser/LASLoader';
+import { expose } from 'threads/worker';
+
+const loader = new LASLoader();
+
+expose({
+    set lazPerf(path) {
```
jailln: Why do you need to exposer `set lazPerf`?

src/Parser/LASParser.js
```diff
@@ -40,7 +60,11 @@ export default {
         if (options.out?.skip) {
             console.warn("Warning: options 'skip' not supported anymore");
         }
-        return lasLoader.parseFile(data, {
+
+        const useWorker = options.in.worker ?? true;
+        const loader = useWorker ? lasWorkerLoader : lasLoader;
+
+        return loader.parseFile(data, {
```
jailln: Since data is an `ArrayBuffer` we should use `Transfer(data)` which [according to the doc](https://threads.js.org/usage-advanced#transferable-objects) speeds up perfs even more.

jailln (commented):
Thanks and congrats on finding an elegant and working way of having web workers in itowns :)

src/Parser/LASParser.js
Desplandis: Damn, forgot to push it!
I'll do it too in the response of `LasParser.js`!

src/Parser/LASParser.js
Desplandis: I used a pool to experiment with different number of workers but failed to notice a difference in performance when using a size $\geq 2$...

Those results could be explained by the fact that I have a low numbers of CPU-cores and those are surely hardware-dependent.
I don't have enough data (performance tests or analytics) to deduce what would be the perfect size but this could be fined-tune later! =)

At least with a single worker, the main thread is not clogged!

P.S.: Maplibre did some A/B testing on their workers and limit the worker pool size to $1$ (see https://github.com/maplibre/maplibre-gl-js/pull/2354). However this could also be task-dependent.

src/Worker/LASWorker.js
Desplandis: Wanted to expose it to set the `laz-perf.wasm` in the same way that with the non-worker loader!

src/Parser/LASParser.js
jailln: Interresting! Thanks

ftoromanoff (approved):

PR #2313, opened 2024-04-02
feat: Remove WebGL1 support

### LINKED ISSUE #2152: [Three.js] Deprecation of WebGL 1 support in WebGLRenderer

Desplandis:
## Description
This PR removes `WebGL1` support. This is a consequence of the removal of `WebGL1` support in the upcoming release `r163` of `three.js` (see [Migrating Guide r162-r163](https://github.com/mrdoob/three.js/wiki/Migration-Guide#162--r163)).

### Changes in this PR:
- Replace uses of `RawShaderMaterial` by `ShaderMaterial` to automatically prepend defines in shaders (given the renderer and material options), as well as built-in uniforms and attributes (see [`WebGLProgram`][WebGLProgram])
- Remove usage of `USE_LOGDEPTHBUF_EXT` as `EXT_frag_depth` is part of the core WebGL2 specification.
- Remove WebGL1-related code paths.

Closes #2152.

[WebGLProgram]: https://threejs.org/docs/index.html#api/en/renderers/webgl/WebGLProgram


src/Renderer/Shader/Chunk/elevation_pars_vertex.glsl
```diff
@@ -25,15 +25,13 @@
...
             return texture2D( tex, uv ).r;
-        #else
-            return texture2D( tex, uv ).w;
-        #endif
         return 0.;
     }
 
     float getElevation(vec2 uv, sampler2D tex, vec4 offsetScale, Layer layer) {
+        // Elevation textures are inverted along the y-axis
+        uv = vec2(uv.x, 1.0 - uv.y);
```
jailln, line 34: Elevation textures were misread? :hushed: this fixes another issue than webgl 1.0 removal isn't it ?

Desplandis: @jailln I now use the `uv` attribute directly added by `WebGLProgram` for a `ShaderMaterial`. Since it has the same semantics as `uv_0` (pre-PR), I decided to replace `uv_0` by `uv`. However, in `TileVS.glsl`, they were already a local variable `vec2 uv = vec2(uv_0.x, 1.0 - uv_0.y);`, which was only used as a parameter of `getElevation`.

I decided to move this code in `getElevation` to reduce the scope of this "hack" as it is only needed because `x-bil` elevation textures are inverted along the y-axis.

jailln: Ah ok thanks for the clarification :)

jailln (approved):

PR #2325, opened 2024-05-03
fix(PointCloud): use preSSE for C3DTilesLayer

AnthonyGlt:
## Description
Use camera preSSE to compute the point size of a C3DTilesLayer in the ATTENUATED mode

## Motivation and Context
I've noticed a difference between the calculus of the size of the points between Potree & 3Dtiles.
In Potree we use the preSSE for the size computation.
In this previous comment https://github.com/iTowns/itowns/pull/2171#discussion_r1311800499, I had highlighted this difference.
Since then, @ftoromanoff change the default `scale` value in the pointMaterial. 
Right now, the attenuated mode doesn't seem that effective.

### DIFF
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -116,8 +116,8 @@ class C3DTilesLayer extends GeometryLayer {
         this.pntsShape = PNTS_SHAPE.CIRCLE;
         this.classification = config.classification;
         this.pntsSizeMode = PNTS_SIZE_MODE.VALUE;
-        this.pntsMinAttenuatedSize = config.pntsMinAttenuatedSize || 3;
-        this.pntsMaxAttenuatedSize = config.pntsMaxAttenuatedSize || 10;
+        this.pntsMinAttenuatedSize = config.pntsMinAttenuatedSize || 1;
+        this.pntsMaxAttenuatedSize = config.pntsMaxAttenuatedSize || 7;
         if (config.pntsMode) {
             const exists = Object.values(PNTS_MODE).includes(config.pntsMode);
             if (!exists) {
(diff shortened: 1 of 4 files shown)


airnez:
There are indeed some differences between 3DTiles and potree point size computation. Here are two videos to compare attenuated size mode behavior. (3DTiles VS Potree)

https://github.com/iTowns/itowns/assets/74316660/4aa5785e-4a77-4f94-b9a3-61b786538070


https://github.com/iTowns/itowns/assets/74316660/fada8738-c8aa-42f4-af53-25ed88e2c18b



Potree implementation seems more correct than 3DTiles one, as max attenuated value is almost not taken into account for 3DTiles. Definitely related to the preSSE scaling :

https://github.com/iTowns/itowns/blob/ee56ec70626213390fa200573552ddb5a54a3b94/src/Renderer/Shader/PointsVS.glsl#L161-L164

A small scaling factor could keep point size under minAttenuated value. This would explain why 3DTiles point size is sticking to minAttenuatedSize.

Desplandis:
Seems that the default clamping values (as well as default size) are not that good, maybe try replacing it with (default values in LiDAR HD demo):
- `size`: 1
- `minSize`: 1
- `maxSize`: 7

src/Layer/C3DTilesLayer.js
```diff
@@ -194,7 +193,8 @@ class C3DTilesLayer extends GeometryLayer {
         });
     }
 
-    preUpdate() {
+    preUpdate(context) {
+        this.scale = context.camera._preSSE;
```
Desplandis: This code shall be in `pre3dTilesUpdate`.

Desplandis (approved):
Sounds good to me. If you have issues with default values, I'll let you modify them and then notify me?

AnthonyGlt:
> Seems that the default clamping values (as well as default size) are not that good, maybe try replacing it with (default values in LiDAR HD demo):
> 
>     * `size`: 1
> 
>     * `minSize`: 1
> 
>     * `maxSize`: 7

I've changed the default values

src/Layer/C3DTilesLayer.js
AnthonyGlt: I did it, I had to carry around the context value though

PR #2326, opened 2024-05-03
fix(C3DTFeature): use correct interleaved buffer getter

AnthonyGlt:
## Description
We should use the getter of threejs to access position in buffer, especially when we have interleaved buffer

## Motivation and Context
Similar to what was done in https://github.com/iTowns/itowns/pull/2266

To note:
I encountered issues related to the batchtable, I wonder if there is some misconception in the treatment of it. I will try to investigate more

### DIFF
--- a/src/Core/3DTiles/C3DTFeature.js
+++ b/src/Core/3DTiles/C3DTFeature.js
@@ -72,13 +72,13 @@ class C3DTFeature {
         target.min.z = Infinity;
 
         this.groups.forEach((group) => {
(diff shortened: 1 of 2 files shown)


AnthonyGlt:
@Desplandis I assign it to you but maybe someone else could be more fitted to this MR.
To whoever review this MR, could you please test it, because I could be the one making a misconception here, thank you :)

Desplandis (approved):
LGTM. I did not find other occurences of using directly the `position` attribute.

PR #2330, opened 2024-05-31
fix(VectorTileLayer): fix icon.id with {xx} or/and .stops

### LINKED ISSUE #2328: Problem adding vector tiles map
Hello!

I want to display a MapBox vector tile map just like the OpenLayers 2D map does.
An example of my code is in the sandbox, here is the [link](https://nrgygm.csb.app/).
1. Why does iTowns draw a sprite of pictures, and not just one icon?
2. How can I add vector tiles from maptiler.com?

@AnthonyGlt, can you help?

ftoromanoff:
Fix Issue #2328 

There is an issue with icons and sprites linked with a bad interpretation of properties containing {data} and/or stops properties.

This PR aims at solving the issue

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -20,32 +20,31 @@ function baseAltitudeDefault(properties, ctx) {
 }
 
 export function readExpression(property, ctx) {
-    if (property != undefined) {
-        if (property.expression) {
-            return property.expression.evaluate(ctx);
-        } else if (property.stops) {
-            const stops = property.stops;
-            property = property.stops[0][1];
-            for (let i = stops.length - 1; i >= 0; i--) {
-                const stop = stops[i];
+    if (property.expression) {
+        return property.expression.evaluate(ctx);
+    }
+    if (property.stops) {
+        const stops = property.stops;
+        property = property.stops[0][1];
+        for (let i = stops.length - 1; i >= 0; i--) {
+            const stop = stops[i];
 
-                if (ctx.zoom >= stop[0]) {
-                    property = stop[1];
-                    break;
-                }
+            if (ctx.zoom >= stop[0]) {
+                property = stop[1];
+                break;
             }
         }
-        if (typeof property === 'string' || property instanceof String) {
-            property = property.replace(/\{(.+?)\}/g, (a, b) => (ctx.properties[b] || '')).trim();
-        }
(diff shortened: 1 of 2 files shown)


ftoromanoff:
@AnthonyGlt could you have a look at this PR?

src/Core/Style.js
```diff
@@ -20,30 +20,31 @@ function baseAltitudeDefault(properties, ctx) {
 }
 
 export function readExpression(property, ctx) {
-    if (property != undefined) {
```
AnthonyGlt, line 23: Is it ok to not check the validity of property anymore ?

src/Core/Style.js
```diff
@@ -20,30 +20,31 @@ function baseAltitudeDefault(properties, ctx) {
...
+    if (property.stops) {
+        const stops = property.stops;
+        property = property.stops[0][1];
+        for (let i = stops.length - 1; i >= 0; i--) {
+            const stop = stops[i];
+
+            if (ctx.zoom >= stop[0]) {
+                property = stop[1];
+                break;
             }
```
AnthonyGlt, line 35: Can `stops` be empty ?
We could do:
```suggestion
        const stops = property.stops;
       
        for (let i = stops.length - 1; i >= 0; i--) {
            const stop = stops[i];
                property = stop[1];
            if (ctx.zoom >= stop[0]) {

                break;
            }
```
WDYT ?

AnthonyGlt (commented):
Thank you for the MR and for the unit test !

src/Core/Style.js
ftoromanoff: Everytime we call this function we always have a check before:
```
if (userValue != undefined) { return readExpression(userValue, style.context); }
```
or
```
if (this.labelDomelement) {
                    content = readExpression(this.labelDomelement, context);
                }
```
Thus i don't see the use to redo the same test afterward.

src/Core/Style.js
ftoromanoff: I don't think stop can be empty but we never know.

I do'tn know if waht you propose really doesn't change anything. I fell lik ein cas we have a lot of stops your might be a bit slower as we will re-affect property at each step instea of only at start and at end. But i might be wrong

AnthonyGlt (approved):

Issue #2159, opened 2023-07-27
Terrain tiles not dividing if not covered by elevation data


mgermerie:
## Context

When displaying elevation data on a `TiledGeometryLayer`, the later stops subdividing its tiles when the attached `ElevationLayer` zoom range is overtaken.

## Steps to Reproduce (for bugs)

This is a minimal example to reproduce the issue : 
```js
const placement = {
    coord: new itowns.Coordinates('EPSG:4326', 2.351323, 48.856712),
    range: 25000000,
}
const viewerDiv = document.getElementById('viewerDiv');
const view = new itowns.GlobeView(viewerDiv, placement);

itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(config) {
    config.source = new itowns.WMTSSource(config.source);
    view.addLayer(new itowns.ColorLayer('Ortho', config));
});

itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then((config) => {
    config.source = new itowns.WMTSSource(config.source);
    view.addLayer(new itowns.ElevationLayer(config.id, config));
});
```
While running it, enter `view.controls.setZoom(17);` in console.

## Expected Behavior

Terrain tiles should subdivide to display ortho images with enough resolution.

## Actual Behavior

Terrain tiles stop subdividing at some point, since no elevation data is available after zoom level 10.

## Possible Cause/Fix/Solution

This is due to [this part](https://github.com/iTowns/itowns/blob/cba214626dc485318e3e3315aa442eecaaff99c2/src/Layer/TiledGeometryLayer.js#L402C1-L416C10), called when updating the terrain layer.

AnthonyGlt:
Do you have an idea about the real impact of deleting this code ?
 https://github.com/iTowns/itowns/blob/cba214626dc485318e3e3315aa442eecaaff99c2/src/Layer/TiledGeometryLayer.js#L402-L416?

Looks to me that it will solve the issue. The original code comes from https://github.com/iTowns/itowns/pull/538
https://github.com/iTowns/itowns/blob/179fa6b176f12aaa49bf1bc35ca75b0bd5e96c9d/src/Process/GlobeTileProcessing.js#L126-L131
and https://github.com/iTowns/itowns/pull/792
https://github.com/iTowns/itowns/blob/182ba92f3d854ac7305c840faea3d1e1c44a7e29/src/Process/GlobeTileProcessing.js#L129-L137

AnthonyGlt:
[...]

PR #2344, opened 2024-06-24
fix(TiledGeometryLayer): remove subdivision checking code

AnthonyGlt:
## Description
I removed the subdivision checking code.
This code was blocking the subdivision of the tiles when no terrain was available at a certain zoom level

## Motivation and Context
This MR is the correction of this issue https://github.com/iTowns/itowns/issues/2159

### DIFF
--- a/src/Core/Prefab/Globe/GlobeLayer.js
+++ b/src/Core/Prefab/Globe/GlobeLayer.js
@@ -44,8 +44,6 @@ class GlobeLayer extends TiledGeometryLayer {
      * level for this tiled layer.
      * @param {number} [config.sseSubdivisionThreshold=1] - Threshold level for
      * the SSE.
-     * @param {number} [config.maxDeltaElevationLevel=4] - Maximum delta between
-     * two elevations tile.
      *
      * @throws {Error} `object3d` must be a valid `THREE.Object3d`.
      */
@@ -68,8 +66,6 @@ class GlobeLayer extends TiledGeometryLayer {
         this.options.defaultPickingRadius = 5;
         this.minSubdivisionLevel = this.minSubdivisionLevel == undefined ? 2 : this.minSubdivisionLevel;
         this.maxSubdivisionLevel = this.maxSubdivisionLevel == undefined ? 19 : this.maxSubdivisionLevel;
-        this.maxDeltaElevationLevel = this.maxDeltaElevationLevel || 4.0;
-
         this.extent = this.schemeTile[0].clone();
 
         for (let i = 1; i < this.schemeTile.length; i++) {
--- a/src/Core/Prefab/Planar/PlanarLayer.js
+++ b/src/Core/Prefab/Planar/PlanarLayer.js
@@ -32,8 +32,6 @@ class PlanarLayer extends TiledGeometryLayer {
      * name.
      * @param {number} [config.maxSubdivisionLevel=5] - Maximum subdivision
      * level for this tiled layer.
-     * @param {number} [config.maxDeltaElevationLevel=4] - Maximum delta between
-     * two elevations tile.
      *
      * @throws {Error} `object3d` must be a valid `THREE.Object3d`.
      */
@@ -50,7 +48,6 @@ class PlanarLayer extends TiledGeometryLayer {
         this.extent = extent;
         this.minSubdivisionLevel = this.minSubdivisionLevel == undefined ? 0 : this.minSubdivisionLevel;
         this.maxSubdivisionLevel = this.maxSubdivisionLevel == undefined ? 5 : this.maxSubdivisionLevel;
-        this.maxDeltaElevationLevel = this.maxDeltaElevationLevel || 4.0;
     }
 }
 
--- a/src/Layer/TiledGeometryLayer.js
+++ b/src/Layer/TiledGeometryLayer.js
@@ -86,6 +86,10 @@ class TiledGeometryLayer extends GeometryLayer {
             throw new Error(`Cannot init tiled layer without builder for layer ${this.id}`);
         }
 
+        if (config.maxDeltaElevationLevel) {
+            console.warn('Config using maxDeltaElevationLevel is deprecated. The parameter maxDeltaElevationLevel is not longer used');
+        }
+
         this.level0Nodes = [];
         const promises = [];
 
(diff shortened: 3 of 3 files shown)


src/Core/Prefab/Globe/GlobeLayer.js
```diff
@@ -68,8 +66,6 @@ class GlobeLayer extends TiledGeometryLayer {
         this.options.defaultPickingRadius = 5;
         this.minSubdivisionLevel = this.minSubdivisionLevel == undefined ? 2 : this.minSubdivisionLevel;
         this.maxSubdivisionLevel = this.maxSubdivisionLevel == undefined ? 19 : this.maxSubdivisionLevel;
-        this.maxDeltaElevationLevel = this.maxDeltaElevationLevel || 4.0;
```
Desplandis, line 71: Could we add a depreciation notice for this parameter?

src/Core/Prefab/Planar/PlanarLayer.js
```diff
@@ -50,7 +48,6 @@ class PlanarLayer extends TiledGeometryLayer {
         this.extent = extent;
         this.minSubdivisionLevel = this.minSubdivisionLevel == undefined ? 0 : this.minSubdivisionLevel;
         this.maxSubdivisionLevel = this.maxSubdivisionLevel == undefined ? 5 : this.maxSubdivisionLevel;
-        this.maxDeltaElevationLevel = this.maxDeltaElevationLevel || 4.0;
```
Desplandis, line 53: Same here or in the parent `TiledGeometryLayer`?

Desplandis (changes requested):
@mgermerie, @jailln, @AnthonyGlt and myself decided that this functionality was no longer relevant, a user is responsible for the resolution of the elevation data he provides. We shall not monkey-patch and cause unexpected behaviors.

src/Layer/TiledGeometryLayer.js
```diff
@@ -412,23 +412,6 @@ class TiledGeometryLayer extends GeometryLayer {
...
-            const currentTexture = nodeLayer.textures[0];
-            if (currentTexture && currentTexture.extent) {
-                const offsetScale = nodeLayer.offsetScales[0];
-                const ratio = offsetScale.z;
-                // ratio is node size / texture size
-                if (ratio < 1 / 2 ** this.maxDeltaElevationLevel) {
-                    return false;
-                }
-            }
-        }
```
Desplandis, line 433: @AnthonyGlt @jailln For info, `offetScale.z` encodes, for a given tile, how many times we should "zoom" an ancestor texture data along its `u`-axis (see [elevation_pars_vertex.glsl](https://github.com/iTowns/itowns/blob/58fc8bb38d4f47fc10e949e92a200c0e8576d8a8/src/Renderer/Shader/Chunk/elevation_pars_vertex.glsl#L35C19-L35C33)).

Given the levels $\mathsf{n}$ of the ancestor elevation data and $\mathsf{m}$ of the current tile, $\mathsf{offsetScale.z} = \frac{1}{2^{\mathsf{m}-\mathsf{n}}}$ (zoom by a power-of-two factor).

As such, this code cause a tile to not subdivide if $\mathsf{m} - \mathsf{n} > \mathsf{maxDeltaElevationLevel}$

src/Core/Prefab/Globe/GlobeLayer.js
AnthonyGlt: Should i leave the parameter and add "deprecated" instead of removing it ?

AnthonyGlt:
It looks like it works because with `STRATEGY_DICHOTOMY` ALL the level are loaded (1 to 10). I don't think this is normal behavior. In `STRATEGY_MIN_TRAFFIC`, we have avoid this thanks to this condition:
https://github.com/iTowns/itowns/blob/d091207146600f80dfd755c5616a16d9333718b0/src/Layer/LayerUpdateStrategy.js#L17-L18
(from eb237cc78cad20dac81fd2723462400d1c751ba5)

If we add this condition in `STRATEGY_DICHOTOMY` (I think we should), we don't load all the levels anymore but we have the same behavior as `STRATEGY_MIN_TRAFFIC`  but the bug is back.


I think the issue could be corrected by adding:
```js
    if (nodeLevel > maxZoom ) {
        return maxZoom;
    }
```

After https://github.com/iTowns/itowns/blob/d091207146600f80dfd755c5616a16d9333718b0/src/Layer/LayerUpdateStrategy.js#L53
So we don't try to reach unreachable level and fallback on the zoom max.
After some tests, it looks like it resolves the issue.

WDYT @Desplandis ?

https://github.com/iTowns/itowns/issues/2159#issuecomment-2309797092

src/Core/Prefab/Globe/GlobeLayer.js
Desplandis: Nope, you should indeed remove it.

Desplandis:
I'm afraid we have a much bigger issue with the strategies...
Your condition is similar to the existing one in `chooseNextLeveltoFetch`, so I'm not sure why this is fixing the problem.

https://github.com/iTowns/itowns/blob/c1409c509328c02aa378ec2749a6e3d9b0885ba5/src/Layer/LayerUpdateStrategy.js#L79

Diving into the code it seems that there is already a code path to fallback to another strategy if the first request failed (e.g. dichotomy). However in practice, it does not seems to be executed since `lowestLevelError` is never assigned in the `LayerUpdateState`... (`failure()` is never called...)

https://github.com/iTowns/itowns/blob/c1409c509328c02aa378ec2749a6e3d9b0885ba5/src/Layer/LayerUpdateStrategy.js#L54-L61


I think we should first document how the elevation layer update state machine works, propose a fix for reaching the fallback code, and surely propose a `highestLevelError` mechanism in the update state for a two-way dichotomy.

Desplandis:
Moreover, I'm all for merging this PR (since it fixes the aforementioned issue, i.e. remove tile subdivision blocking mechanism) and open another one fixing the strategy/state machine issues.

Desplandis (approved):

AnthonyGlt:
@Desplandis Thank you for the review.
I agree, we should understand how to trigger the fallback correctly.

>Your condition is similar to the existing one in chooseNextLeveltoFetch, so I'm not sure why this is fixing the problem.

You are right, it's useful in the case where we have 
```js
 if (node.pendingSubdivision) { 
     return currentLevel; 
 ```
But it means that we force the subdivision, which go against the condition on `pendingSubdivision`.

I also think that merging as it is, is the best way to go BUT this subdivision process must become a priority to understand/refactor because it has some questionable code

PR #2347, opened 2024-06-27
[3D tiles migration] Setup layer whenReady

jailln:
Setup `whenReady` of the new `OGC3DTiles` layer


Desplandis (approved):
LGTM. We should however as noted in this PR move the logic of `View#_preprocessLayer` in each layer! Maybe in a `Layer#preprocess`. I'll open an issue on this subject to clarify, formalize the `Layer` interface as well as the whole lifecycle of a `Layer`.

PR #2348, opened 2024-06-28
PointsMaterial as superset of three's PointsMaterial

Desplandis:
## Motivation and Context
`itowns` pointclouds currently uses a custom points material that greatly diverges from the default material provided by `three`. This divergence leads to complications, especially when using our material in places where the default material is expected (e.g. our on-going work on `3d-tiles-renderer`). To address this, this PR propose refactoring our custom material into a superset of the `three` default points material.

## Description
The exhaustive list of changes in this PR:
- [x] Add support of all relevant `Material` properties  (`alphaTest`, `clippingPlanes`, `toneMapped`, `vertexColors`)
- [x] Add support of `diffuse` property (replacing the old `overlayColor`)
- [x] Add support of `fog` properties (enabled by default)
- [x] Add support of `map` texture properties
- [x] Add support of `alphaMap` and `alphaTest` properties
- [x] Add support of `clippingPlanes` properties
- [x] Cleanup of some unused code (`update`)
- [x] BREAKING CHANGE: Remove support of non-tested and non-documented `orientedImageMaterial`
- [x] BREAKING CHANGE: Remove support of `overlayColor` (replaced by the standard `diffuse` property)
- [x] Use a single `normal` uniform (as `NORMAL_OCT16P` and `NORMAL_SPHEREMAPPED` are unused or at best hidden behind undocumented flags)

Changes in a future PR?
- Add support in the parser (e.g. `PotreeParser`) to decode encoded normals
- Add define guards against different kind of color mode (but will introduce a breaking change) since we'll need to manually trigger recompile (or get from cache) using `material.needsUpdate = true`.
- Dedupe similar uniforms (multiple ranges, multiple textures)

### DIFF
--- a/src/Renderer/PointsMaterial.js
+++ b/src/Renderer/PointsMaterial.js
@@ -186,57 +186,56 @@ class PointsMaterial extends THREE.ShaderMaterial {
      * pointMaterial.recomputeClassification();
      */
     constructor(options = {}) {
-        const intensityRange = options.intensityRange || new THREE.Vector2(1, 65536);
-        const elevationRange = options.elevationRange || new THREE.Vector2(0, 1000);
-        const angleRange = options.angleRange || new THREE.Vector2(-90, 90);
-        const classificationScheme = options.classification || ClassificationScheme.DEFAULT;
-        const discreteScheme = options.discreteScheme || DiscreteScheme.DEFAULT;
-        const size = options.size || 0;
-        const mode = options.mode || PNTS_MODE.COLOR;
-        const shape = options.shape || PNTS_SHAPE.CIRCLE;
-        const sizeMode = size === 0 ? PNTS_SIZE_MODE.ATTENUATED : (options.sizeMode || PNTS_SIZE_MODE.VALUE);
-        const minAttenuatedSize = options.minAttenuatedSize || 3;
-        const maxAttenuatedSize = options.maxAttenuatedSize || 10;
-        let gradients = Gradients;
-        if (options.gradient) {
-            gradients = {
-                ...options.gradient,
-                ...Gradients,
-            };
-        }
+        const gradients = {
+            ...options.gradient,
+            ...Gradients,
+        };
+        options.gradient = Object.values(gradients)[0];
 
-        delete options.intensityRange;
-        delete options.elevationRange;
-        delete options.angleRange;
-        delete options.classification;
-        delete options.discreteScheme;
-        delete options.size;
-        delete options.mode;
-        delete options.shape;
-        delete options.sizeMode;
-        delete options.minAttenuatedSize;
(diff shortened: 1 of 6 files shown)


src/Renderer/PointsMaterial.js
```diff
@@ -187,59 +186,50 @@ class PointsMaterial extends THREE.ShaderMaterial {
...
-        const minAttenuatedSize = options.minAttenuatedSize || 3;
-        const maxAttenuatedSize = options.maxAttenuatedSize || 10;
-        let gradients = Gradients;
-        if (options.gradient) {
-            gradients = {
-                ...options.gradient,
-                ...Gradients,
-            };
-        }
+        const {
```
ftoromanoff, line 195: You can add here:
```
        const gradients = {
            ...options.gradient,
            ...Gradients,
        };
        options.gradient = Object.values(gradients)[0];
```
If we want to add a user defined gradient

src/Renderer/PointsMaterial.js
```diff
@@ -187,59 +186,50 @@ class PointsMaterial extends THREE.ShaderMaterial {
...
+            angleRange = new THREE.Vector2(-90, 90),
+            classificationScheme = ClassificationScheme.DEFAULT,
+            discreteScheme = DiscreteScheme.DEFAULT,
+            size = 1,
+            mode = PNTS_MODE.COLOR,
+            shape = PNTS_SHAPE.CIRCLE,
+            sizeMode = PNTS_SIZE_MODE.ATTENUATED,
+            minAttenuatedSize = 3,
+            maxAttenuatedSize = 10,
+            gradient = Gradients.SPECTRAL,
```
ftoromanoff:
```
gradient,
```

src/Renderer/PointsMaterial.js
```diff
@@ -187,59 +186,50 @@ class PointsMaterial extends THREE.ShaderMaterial {
...
-        delete options.shape;
-        delete options.sizeMode;
-        delete options.minAttenuatedSize;
-        delete options.maxAttenuatedSize;
-        delete options.gradient;
-
-        super(options);
         this.userData.needTransparency = {};
-        this.gradients = gradients;
+        this.gradients = Gradients;
```
ftoromanoff: As well as:
```
this.gradients = gradients;
```

src/Renderer/PointsMaterial.js
Desplandis: Okay, I'll not change this in this PR. But I have some use-case where we could want to access to the 2D gradient to draw it in a legend.

ftoromanoff (approved):

gchoqueux:
ohh my bad, I hadn't noticed that the orientedImageMaterial support is removed without being deprecated. We had a summer school based on this feature

PR #2349, opened 2024-06-28
fix(test): import HttpsProxyAgent

gchoqueux:
## Description
fix import HttpsProxyAgent in unit test


Desplandis (approved):
Thanks for the fix!
Yikes, neither our linter nor our CI (since process.env.HTTPS_PROXY is not defined) could catch this error...

PR #2351, opened 2024-07-01
fix(test): fix potree unit test

gchoqueux:
## Description
Resolve all loading states, to avoid timeouts in unit tests


gchoqueux:
it seems that the timeout on github action is caused by requests that are a little longer than local test unit.
A fix will be increase the timeout with : `mocha --timeout 5000`

PR #2352, opened 2024-07-02
refator(PointCloudDebug): fix SizeMode

ftoromanoff:
Allow dynamic parameters display on PointCloud GUI linked with the Point Size mode choosed

### DIFF
--- a/utils/debug/PointCloudDebug.js
+++ b/utils/debug/PointCloudDebug.js
@@ -14,39 +14,55 @@ function getController(gui, name) {
 }
 
 function hideController(gui, name) {
-    getController(gui, name).__li.style.display = 'none';
+    const controller = getController(gui, name);
+    if (controller) {
+        controller.__li.style.display = 'none';
+    }
 }
 
 function showController(gui, name) {
-    getController(gui, name).__li.style.display = '';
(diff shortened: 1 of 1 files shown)


utils/debug/PointCloudDebug.js
```diff
@@ -21,41 +21,53 @@ function showController(gui, name) {
...
         hideController(gui, 'maxAngleRange');
     }
+
+    sizeMode =  parseInt(sizeMode, 10);
+    if (sizeMode === PNTS_SIZE_MODE.VALUE) {
+        hideController(gui, 'minAttenuatedSize');
+        hideController(gui, 'maxAttenuatedSize');
+        showController(gui, 'pointSize');
+    } else {
+        hideController(gui, 'pointSize');
```
Desplandis: I think we should not hide `pointSize` parameter as it used in the attenuation calculation of `gl_Pointsize` in the shader.

Desplandis (changes requested):
Sounds good to me, except my comment. We should migrate to `lil-gui` someday, this should simplify some code. x)

Desplandis (approved):
Thanks for the change, it should be clearer to the end-user. =)

Issue #2353, opened 2024-07-04
Timeout in unit test potree 2 


gchoqueux:
There's time out in unit test potree 2 

https://github.com/iTowns/itowns/actions/runs/9791693583/job/27036065634

## proposal

Increase the timeout 
`mocha --timeout 5000`

jailln:
What is the timeout related to? Is it related to files taking too long to load/loading too much files? Is it related to the server? I'm wondering if we should increase the timeout or consider changing the test file.

By the way, if we keep this test file, we should consider storing the test dataset on [itowns-sample-data-2](https://github.com/iTowns/iTowns2-sample-data), would that be feasible? 

@Desplandis @ketourneau

ketourneau:
Hi the test file is already in [itowns-sample-data-2](https://github.com/iTowns/iTowns2-sample-data/tree/master/pointclouds/potree2.0/lion).

I created a PR to update sample data url : https://github.com/iTowns/itowns/pull/2354.

I hope it will solve the test timeout

ketourneau:
I launched 20x potree2 test and I only got one time the timeout.
The average test time is 540 ms so I don't think increasing the timeout will solve the problem.
I'll keep investigating.

gchoqueux:
time out in unit test 
https://github.com/iTowns/itowns/actions/runs/10092010021/job/27904819707#step:5:765

PR #2354, opened 2024-07-04
fix(potree2): Fix sample data url

ketourneau:
## Motivation and Context

As mentioned in https://github.com/iTowns/itowns/issues/2353 it's best to use itowns sample data.


jailln (commented):
Thanks :)

jailln (approved):

PR #2361, opened 2024-07-12
fix(PointCloud): Allow using custom object3d on PointCloudLayer

ketourneau:
## Description
I think it's a missing behaviour, we need to pass custom object3d to place pointCloud at a specific position.

You can see live demo here : https://demo-hackathon.netlify.app/examples/demo_hackathon_orvault_planarview

And how we use custom object for point cloud placement here : https://github.com/bloc-in-bloc/itowns/blob/58636e3319a5ef0385e2b031e61d7d071126a169/examples/demo_hackathon_orvault_planarView.html#L303


Desplandis:
Yup, a missing behavior on our part!

Desplandis (approved):

PR #2362, opened 2024-07-15
 fix(shader): Discard any point shape with zero alpha color instead of circles only

airnez:
## Description
Points shader discards shapes if they have a zero alpha color, which corresponds to an invisible layer or opacity set to zero.
Until now it was only applied to circle shapes. However, we are also using square shapes, and could implement others in the future. For instance, using a classification scheme for a 3DTiles layer with one class visibility set to false, the points will get white instead of invisible if shape is set to square. I hid water in this example :

image

Corresponding classification (water corresponds to class id 9) :
```{
    "0": {
        "visible": true,
        "color": 9671571,
        "opacity": 1,
        "layerIndex": 0
    },
    "1": {
        "visible": true,
        "color": 11776947,
        "opacity": 1,
        "layerIndex": 1
    },
    "2": {
        "visible": true,
        "color": 11375223,
        "opacity": 1,
        "layerIndex": 2
    },
    "3": {
        "visible": true,
        "color": 2524966,
        "opacity": 1,
        "layerIndex": 3
    },
    "4": {
        "visible": true,
        "color": 7252534,
        "opacity": 1,
        "layerIndex": 4
    },
    "5": {
        "visible": true,
        "color": 7322917,
        "opacity": 1,
        "layerIndex": 5
    },
    "6": {
        "visible": true,
        "color": 15328998,
        "opacity": 1,
        "layerIndex": 6
    },
    "7": {
        "visible": true,
        "color": 16711680,
        "opacity": 1,
        "layerIndex": 7
    },
    "8": {
        "visible": true,
        "color": 9055202,
        "opacity": 1,
        "layerIndex": 8
    },
    "9": {
        "visible": false,
        "color": 8158421,
        "opacity": 1,
        "layerIndex": 9
    },
    "10": {
        "visible": true,
        "color": 4473924,
        "opacity": 1,
        "layerIndex": 10
    },
    "11": {
        "visible": true,
        "color": 9732464,
        "opacity": 1,
        "layerIndex": 11
    },
    "12": {
        "visible": true,
        "color": 5116551,
        "opacity": 1,
        "layerIndex": 12
    },
    "13": {
        "visible": true,
        "color": 16776960,
        "opacity": 1,
        "layerIndex": 13
    },
    "14": {
        "visible": true,
        "color": 16776960,
        "opacity": 1,
        "layerIndex": 14
    },
    "15": {
        "visible": true,
        "color": 16776960,
        "opacity": 1,
        "layerIndex": 15
    },
    "16": {
        "visible": true,
        "color": 16776960,
        "opacity": 1,
        "layerIndex": 16
    },
    "17": {
        "visible": true,
        "color": 6776736,
        "opacity": 1,
        "layerIndex": 17
    },
    "18": {
        "visible": true,
        "color": 10027008,
        "opacity": 1,
        "layerIndex": 18
    },
    "DEFAULT": {
        "visible": true,
        "color": 9671571,
        "opacity": 1,
        "layerIndex": 19
    }
}
```

### DIFF
--- a/src/Renderer/Shader/PointsFS.glsl
+++ b/src/Renderer/Shader/PointsFS.glsl
@@ -20,7 +20,7 @@ void main() {
 #include <clipping_planes_pars_fragment>
     if (shape == PNTS_SHAPE_CIRCLE) {
         //circular rendering in glsl
-        if ((length(gl_PointCoord - 0.5) > 0.5) || (vColor.a == 0.0)) {
+        if ((length(gl_PointCoord - 0.5) > 0.5)) {
             discard;
         }
     }
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -18,6 +18,7 @@ The following people have contributed to iTowns.
   * [Vincent Jaillot](https://github.com/jailln)
   * [Anthony Gullient](https://github.com/AnthonyGlt)
   * [Jonathan Garnier](https://github.com/jogarnier)
+  * [Irénée Dubourg](https://github.com/airnez)
 
 * [Oslandia](http://www.oslandia.com)
   * [Vincent Picavet](https://github.com/vpicavet)
--- a/src/Renderer/PointsMaterial.js
+++ b/src/Renderer/PointsMaterial.js
@@ -148,7 +148,7 @@ function recomputeTexture(scheme, texture, nbClass) {
         data[j + 2] = parseInt(255 * color.b, 10);
         data[j + 3] = visible ? parseInt(255 * opacity, 10) : 0;
 
-        needTransparency = needTransparency || opacity < 1;
+        needTransparency = needTransparency || opacity < 1 || !visible;
     }
     texture.needsUpdate = true;
     return needTransparency;


Desplandis:
Hi @airnez!

I manage to reproduce your issue on `master` on all point cloud implementations. However, this bug is not triggered when setting the `opacity` property of the classification to `0.0`.

I suppose this could be a problem with the material not setting its `transparent` property when using a classification texture with transparency and thus rendering as opaque. @ftoromanoff you encountered a similar issue, isn't it?

If this is indeed a problem with the `transparent` property, we could remove the early-discard with `vColor.a == 0.0`.

src/Renderer/Shader/PointsFS.glsl
```diff
@@ -18,9 +18,12 @@ void main() {
 
 // Early discard (clipping planes and shape)
 #include <clipping_planes_pars_fragment>
+    if (vColor.a == 0.0) {
+        discard;
+    }
     if (shape == PNTS_SHAPE_CIRCLE) {
         //circular rendering in glsl
-        if ((length(gl_PointCoord - 0.5) > 0.5) || (vColor.a == 0.0)) {
```
ftoromanoff, line 23: this ```vColor.a == 0.0``` has no reason to be here indeed (as there is no reason to be apply only on shape == Circle)
But should it be check on all point to be discarded or  not ?

As  @Desplandis said, it's the property transparent of the material whould should be set to true when one class or more has an opacity that is not 1.
https://github.com/iTowns/itowns/blob/e6bc46212188f7e932f47d1ad00f38ab5e3b39ef/src/Layer/PointCloudLayer.js#L197

Thus we should add ```|| !visible``` in https://github.com/iTowns/itowns/blob/e6bc46212188f7e932f47d1ad00f38ab5e3b39ef/src/Renderer/PointsMaterial.js#L151

src/Renderer/Shader/PointsFS.glsl
```diff
@@ -18,9 +18,12 @@ void main() {
 
 // Early discard (clipping planes and shape)
 #include <clipping_planes_pars_fragment>
+    if (vColor.a == 0.0) {
+        discard;
```
ftoromanoff: Should we keep the discard when ```opacity === 0``` or ```visible === false```.
This i can't tell, i don't have enought knowledge about shader. WDYT @Desplandis @jailln

Desplandis: Yeah, we should drop this condition and respect material semantics which specifies that a material with `transparent = false` shall be opaque.

src/Renderer/Shader/PointsFS.glsl
airnez: I agree. I removed the early discard and applied suggested fix. It solves the issue.

src/Renderer/Shader/PointsFS.glsl
airnez: Removed

Desplandis (approved):
This seems to fix the issue, I tested it with the different classes set to non visible and the points are indeed transparent. OK to merge.

PR #2364, opened 2024-07-18
[OGC3DTiles] Add support for picking

Desplandis:
## Description

This PR add support for picking on new `OGC3DTilesLayer`. It is compatible with the old interface of `C3DTilesLayer` except that `getC3DTileFeatureFromIntersectsArrays`  (please rename this 40 chars method name...) now returns an object containing features for the closest intersection instead of an old `C3DTilesFeature` instance.

This PR also add picking for points data, which was not supported on the old implementation (you could try it live with `3dtiles_pointcloud` example).

Note that we are waiting for NASA-AMMOS/3DTilesRendererJS#627 to be merged and new version of `3d-tiles-renderer` before merging this PR. =)

### DIFF
--- a/src/Layer/OGC3DTilesLayer.js
+++ b/src/Layer/OGC3DTilesLayer.js
@@ -14,7 +14,6 @@ import { KTX2Loader } from 'ThreeExtended/loaders/KTX2Loader';
 import PointsMaterial from 'Renderer/PointsMaterial';
 import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
 
-
 /*
  * A callback to execute for a given tile of a tileset associated with an
  * {@link OGC3DTilesLayer}.
@@ -29,6 +28,7 @@ import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
  * renderer (if it does not meet conditions to be visible for instance).
 */
 
+const _raycaster = new THREE.Raycaster();
 
 // Internal instance of GLTFLoader, passed to 3d-tiles-renderer-js to support GLTF 1.0 and 2.0
 // Temporary exported to be used in deprecated B3dmParser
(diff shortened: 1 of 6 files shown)


src/Layer/OGC3DTilesLayer.js
```diff
@@ -167,7 +169,58 @@ class OGC3DTilesLayer extends GeometryLayer {
...
+     * @param {THREE.Intersection<THREE.Mesh | THREE.Points>[]} [target=[]]
+     * @returns {THREE.Intersection<THREE.Mesh | THREE.Points>[]}
+     */
+    pickObjectsAt(view, coords, radius, target = []) {
+        const camera = view.camera.camera3D;
+        _raycaster.setFromCamera(view.viewToNormalizedCoords(coords), camera);
+        _raycaster.near = camera.near;
+        _raycaster.far = camera.far;
+
+        _raycaster.firstHitOnly = true;
```
jailln, line 337: Are we sure we only want the first object only? Can you make it configurable instead?

Desplandis: This was for efficiency purpose. Since the intersects are already sorted by distance (since this is a ray) and that we only use the closest intersects in `getC3DTileFeatureFromIntersectsArrays` (old and new implementation), I thought that we could stop at the first intersection.

jailln: I think there is use cases where `pickObjectsAt` is used without `getC3DTileFeatureFromIntersectsArray` (either directly on the layer or on the `View` with `View.pickObjectsAt`) and where one would like to have all the intersections returned. We can make it configurable and default `firstHitOnly` to `true`, WDYT?

Desplandis: Agree, but we'll have to modify the signature of `picksObjectAt`.

Desplandis: I propose to remove it for now and test on larger datasets if there is a huge overhead.

Desplandis:
Rebased on #2365

src/Layer/OGC3DTilesLayer.js
jailln: > Agree, but we'll have to modify the signature of `picksObjectAt`.

Yes you're right, that's another subject. Let's keep the optim for now and see if we have a usecase for making it configurable later.

jailln (approved):

Desplandis:
Rebased on #2371

PR #2370, opened 2024-07-25
fix(parser): wrong shpjs 6.0.1 using

gchoqueux:
## Description
fix(parser): wrong shpjs 6.0.1 using 

see https://www.npmjs.com/package/shpjs


mgermerie (approved):
Thanks for seeing this !

PR #2372, opened 2024-07-29
[OGC3DTiles] Missing properties

Desplandis:
## Description

This PR adds missing properties of new `OGC3DTiles` Layer as well as updating debug tools.

### DIFF
--- a/utils/debug/3dTilesDebug.js
+++ b/utils/debug/3dTilesDebug.js
@@ -1,11 +1,7 @@
-import * as THREE from 'three';
-import View from 'Core/View';
-import GeometryLayer from 'Layer/GeometryLayer';
-import { C3DTilesBoundingVolumeTypes } from 'Core/3DTiles/C3DTilesEnums';
+import { DebugTilesPlugin } from '3d-tiles-renderer';
+
 import { PNTS_MODE, PNTS_SHAPE, PNTS_SIZE_MODE } from 'Renderer/PointsMaterial';
 import GeometryDebug from './GeometryDebug';
-
-const bboxMesh = new THREE.Mesh();
 
 export default function create3dTilesDebugUI(datDebugTool, view, _3dTileslayer) {
     const gui = GeometryDebug.createGeometryDebugUI(datDebugTool, view, _3dTileslayer);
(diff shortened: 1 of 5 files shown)


utils/debug/3dTilesDebug.js
```diff
@@ -1,88 +1,39 @@
...
-            } else if (tile.boundingVolume.initialVolumeType === C3DTilesBoundingVolumeTypes.sphere ||
-                       tile.boundingVolume.initialVolumeType === C3DTilesBoundingVolumeTypes.region) {
-                const geometry = new THREE.SphereGeometry(tile.boundingVolume.volume.radius, 32, 32);
-                const material = new THREE.MeshBasicMaterial({ wireframe: true, color: Math.random() * 0xffffff });
-                helper = new THREE.Mesh(geometry, material);
-            } else {
-                console.warn(`[3D Tiles Debug]: Unknown bounding volume: ${tile.boundingVolume}`);
-                return;
-            }
+    gui.add({ frozen: _3dTileslayer.frozen }, 'frozen').onChange(((value) => {
```
jailln, line 12: Maybe we should remove this option ?

utils/scripts.cjs
```diff
@@ -0,0 +1,39 @@
+const babel = require('@babel/core');
```
jailln: Not related to this PR ?

utils/Worker.js
```diff
@@ -0,0 +1,114 @@
+/**
```
jailln: Not related to this PR ?

Desplandis: Yup, I'll remove it

utils/debug/3dTilesDebug.js
Desplandis: I image it was used to test a single subdivision state (LOD of tiles, which tiles are loaded, etc...). We have the same property on all GeometryLayers.

utils/scripts.cjs
Desplandis: Removed it.

utils/debug/3dTilesDebug.js
jailln: Ok, sorry, I thought it wouldn't work with `3d-tiles-renderer` but it will since it is handled in the`MainLoop` and it will prevent to call `OGC3DTilesLayer.preUpdate`

jailln (approved):

PR #2374, opened 2024-07-30
(PointCloudLayer) add points filtre based on classification

ftoromanoff:
Any classification scheme used for PointCloudLayer havea property 'visible' for each class.
This PR aims at using this value to filter point whichever the visualization mode choosen.

In other word, if a class has its property 'visible' set to false, we want to discard all points having the value in the rendering, (even if we choose to use the 'intensity' rendering)

(In this PR, I used migrate all PointCloud examples to lilGUI instead of datGUI, move to #2546 )

### DIFF
--- a/packages/Main/src/Renderer/PointsMaterial.js
+++ b/packages/Main/src/Renderer/PointsMaterial.js
@@ -113,7 +113,6 @@ function generateGradientTexture(gradient) {
 }
 
 function recomputeTexture(scheme, texture, nbClass) {
-    let needTransparency;
     const data = texture.image.data;
     const width = texture.image.width;
     if (!nbClass) { nbClass = Object.keys(scheme).length; }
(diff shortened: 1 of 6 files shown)


HoloTheDrunk:
I see we're using a conditional discard in the fragment shader; it might be worth benchmarking the performance of this against using a transparent color or performing the filtering earlier on CPU (which, given the right optimization work, might turn out to be the best solution).
See this (old) SO post for an explanation: https://stackoverflow.com/questions/8509051/is-discard-bad-for-program-performance-in-opengl
This is always going to depend on each user's hardware anyways, but checking wouldn't hurt if we can gain a bit of performance.

Desplandis (approved):

Desplandis (dismissed):
Need another review.

packages/Main/src/Renderer/PointsMaterial.js
```diff
@@ -372,20 +366,43 @@ class PointsMaterial extends THREE.ShaderMaterial {
     }
 
     recomputeClassification() {
-        const needTransparency = recomputeTexture(this.classificationScheme, this.classificationTexture, 256);
-        this.userData.needTransparency[PNTS_MODE.CLASSIFICATION] = needTransparency;
```
AnthonyGlt, line 376: It's used in https://github.com/iTowns/itowns/blob/2fae44bf1feb769fedcc52b6ae9045d01f278b4c/packages/Main/src/Layer/ReferencingLayerProperties.js#L61

By doing this, we can't hide specific classes of the 3dtiles

examples/copc_simple_loader.html
```diff
@@ -15,7 +15,7 @@
         </style>
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+        <script src="https://cdn.jsdelivr.net/npm/lil-gui@0.19"></script>
```
AnthonyGlt: If we change it here, could we not change it in every examples to keep some coherence ?

AnthonyGlt (commented):
Thank you for the mr :+1:

examples/copc_simple_loader.html
ftoromanoff: I will put the 1st commit in a separate PR.

AnthonyGlt (approved):

PR #2375, opened 2024-08-02
[Ogc3dtiles] Fix tests

Desplandis:
This PR fixes test errors on the `3d-tiles-migration` branch :
- Revert original 3D Tiles examples to those on master
- Add new 3D Tiles example for the new 3D Tiles implementation with picking support and zoom to tileset feature
- Fix a linting issue in a unit test.

### DIFF
--- a/examples/3dtiles_loader.html
+++ b/examples/3dtiles_loader.html
@@ -1,0 +1,150 @@
+<html>
+    <head>
+        <title>Itowns - 3D Tiles loader</title>
+
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <style type="text/css">
+            #description {
+                z-index: 2;
+                right: 10px;
+            }
+        </style>
+
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <div id="description">Specify the URL of a tileset to load:
+            <input type="text" id="url" />
+            <button onclick="setURL(document.getElementById('url').value)">
+                Load
+            </button>
+            <hr />
+            <p><b>Feature Information:</b></p>
+            <div id="featureInfo"></div>
+        </div>
+
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+
+        <script type="importmap">
+            {
+                "imports": {
+                    "three": "https://cdn.jsdelivr.net/npm/three@0.165.0/build/three.module.js",
+                    "three/addons/": "https://cdn.jsdelivr.net/npm/three@0.165.0/examples/jsm/"
+                }
+            }
+		</script>
+
+        <script type="module">
+            import { AmbientLight } from 'three';
+            import {
+                zoomToLayer,
+                fillHTMLWithPickingInfo,
+            } from './jsm/OGC3DTilesHelper.js';
+
+            const {
+                TMSSource, WMTSSource, OGC3DTilesSource,
+                ColorLayer, ElevationLayer, OGC3DTilesLayer,
+                GlobeView, Coordinates, Fetcher,
+            } = itowns;
+
+            const uri = new URL(location);
+            const state = {
+                // URL to tileset JSON
+                tileset: uri.searchParams.get('tileset'),
+                // Cesium ION /
+                assetId: uri.searchParams.get('assetId'),
+            };
+
+            function setURL(url) {
+                if (!url) return;
+
+                uri.searchParams.set('tileset', url);
+                history.pushState(null, '', `?${uri.searchParams.toString()}`);
+
+                location.reload();
+            }
+
+            // ---- CREATE A GlobeView FOR SUPPORTING DATA VISUALIZATION ----
+
+            // Define camera initial position
+            const placement = {
+                coord: new Coordinates('EPSG:4326', 2.351323, 48.856712),
+                range: 12500000,
+            };
+
+            // `viewerDiv` will contain iTowns' rendering area (`<canvas>`)
+            const viewerDiv = document.getElementById('viewerDiv');
+
+            // Create a GlobeView
+            const view = new GlobeView(viewerDiv, placement, {});
+
+            // Add ambient light to globally illuminates all objects
+            const light = new AmbientLight(0x404040, 15);
+            view.scene.add(light);
+
+            // Setup loading screen
+            setupLoadingScreen(viewerDiv, view);
+
+            // Setup debug menu
+            const menuGlobe = new GuiTools('menuDiv', view, 300);
+            debug.createTileDebugUI(menuGlobe.gui, view, view.tileLayer);
+
+
+            // ---- ADD A BASEMAP ----
+
+            // Add one imagery layer to the scene. This layer's properties are
+            // defined in a json file, but it cou   ld be defined as a plain js
+            // object. See `Layer` documentation for more info.
+            Fetcher.json('./layers/JSONLayers/OPENSM.json').then((config) => {
+                const layer = new ColorLayer('Ortho', {
+                    ...config,
+                    source: new TMSSource(config.source),
+                });
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            });
+
+            // ---- ADD 3D TILES TILESET ----
+
+            // Enable various compression support for 3D Tiles tileset:
+            // - `KHR_draco_mesh_compression` mesh compression extension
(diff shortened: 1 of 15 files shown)


Desplandis:
Rebased on #2376. Shall pass all tests.

examples/3dtiles_loader.html
```diff
@@ -0,0 +1,144 @@
...
+                    source: new TMSSource(config.source),
+                });
+                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+            });
+
+            // ---- ADD 3D TILES TILESET ----
+
+            // Enable draco compression support for 3D Tiles tileset using
+            // the KHR_draco_mesh_compression extension
+            itowns.enableDracoLoader('./libs/draco/');
```
jailln, line 120: We could also enable ktx2Loader

examples/3dtiles_loader.html
```diff
@@ -0,0 +1,144 @@
+<html>
```
jailln, line 1: Cool, thanks for adding this example! Nice that we finally have an example where we can specify the url of a tileset and that zooms to it :)

Why reloading the page each time we change the url ? We could remove the tileset and add a new one instead to avoid reloading the page.

examples/3dtiles_loader.html
```diff
@@ -0,0 +1,144 @@
...
+                const pickingArgs = {
+                    htmlDiv: document.getElementById('featureInfo'),
+                    view,
+                    layer,
+                };
+
+                // Add the layer to our view
+                view.addLayer(layer).then((layer) => {
+                    zoomToLayer(view, layer);
+                    window.addEventListener('mousemove',
```
jailln: I'm not sure we should keep this event on mousemove since it can result in very bad performances for big tilesets. I'd rather define it on mouseclick (or mouse + button) and add a little help displaying the action to do to display feature information.

examples/3dtiles_loader.html
```diff
@@ -0,0 +1,144 @@
...
+                        (event) => fillHTMLWithPickingInfo(event, pickingArgs), false);
+                });
+
+                debug.createOGC3DTilesDebugUI(menuGlobe.gui, view, layer);
+            }
+
+            window.setURL = setURL;
```
jailln, line 147: Some tilesets are kind of dark, we should add direct and ambient lightning.

examples/3dtiles_loader.html
```diff
@@ -0,0 +1,144 @@
...
+                    htmlDiv: document.getElementById('featureInfo'),
+                    view,
+                    layer,
+                };
+
+                // Add the layer to our view
+                view.addLayer(layer).then((layer) => {
+                    zoomToLayer(view, layer);
+                    window.addEventListener('mousemove',
+                        (event) => fillHTMLWithPickingInfo(event, pickingArgs), false);
```
jailln, line 141: I get the following error when hovering the geometry with [this tileset](https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/3DTiles/lyon_1_4978/tileset.json)

```
BatchTable.js:36 Uncaught Error: BatchTable: id value "122.00147247314453" out of bounds for "0" features number.
    at BatchTable.getDataFromId (BatchTable.js:36:1)
    at OGC3DTilesLayer.getC3DTileFeatureFromIntersectsArray (OGC3DTilesLayer.js:325:1)
    at fillHTMLWithPickingInfo (OGC3DTilesHelper.js:30:15)
    at 3dtiles_loader.html?tileset=https%3A%2F%2Fraw.githubusercontent.com%2FiTowns%2FiTowns2-sample-data%2Fmaster%2F3DTiles%2Flyon_1_4978%2Ftileset.json:136:36
getDataFromId @ BatchTable.js:36
getC3DTileFeatureFromIntersectsArray @ OGC3DTilesLayer.js:325
fillHTMLWithPickingInfo @ OGC3DTilesHelper.js:30
(anonymous) @ 3dtiles_loader.html?tileset=https%3A%2F%2Fraw.githubusercontent.com%2FiTowns%2FiTowns2-sample-data%2Fmaster%2F3DTiles%2Flyon_1_4978%2Ftileset.json:136
```

This reminds me of the errors fixed in #2266 and #2326

examples/3dtiles_loader.html
```diff
@@ -0,0 +1,144 @@
...
+
+        <script src="js/GUI/GuiTools.js"></script>
+        <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script>
+
+        <script type="importmap">
+            {
+                "imports": {
+                    "three": "https://cdn.jsdelivr.net/npm/three@0.165.0/build/three.module.js",
```
jailln, line 40: I get `WARNING: Multiple instances of Three.js being imported.
(anonymous) @ three.module.js:53469` when launching the example.

examples/jsm/OGC3DTilesHelper.js
```diff
@@ -0,0 +1,105 @@
...
+    let cameraTransform;
+    if (region) {
+        cameraTransform = zoomToRegion(view, region);
+    } else if (box) {
+        cameraTransform = zoomToBox(view, tile, box);
+    } else {
+        cameraTransform = zoomToSphere(view, tile, sphere);
+    }
+
+    view.controls.lookAtCoordinate({
```
jailln, line 94: We could add a little tilt, e.g. 45°

examples/jsm/OGC3DTilesHelper.js
```diff
@@ -0,0 +1,105 @@
...
+        coord: cameraTransform.coord,
+        range: 1.25 * cameraTransform.range, // zoom out a little bit
+    });
+}
+
+export function zoomToLayer(view, layer) {
```
jailln, line 101: Cool thanks! We could ultimately (i.e. later) add a `zoomToLayer` method in `View` or distribute it as an util.

examples/3dtiles_loader.html
Desplandis: Yup, forgot about that one!

examples/3dtiles_loader.html
Desplandis: To simplify the code a little bit. Especially if it used as a base for new itowns users. :)

examples/3dtiles_loader.html
Desplandis: Agree, those are some leftovers of the initial example that I modified. IMO, we shall go for mouseclick for simplicity.

examples/3dtiles_loader.html
Desplandis: Yup, I'll add some.

examples/3dtiles_loader.html
Desplandis: Yeah, this is because I use the ESM import for three instead of the one in the bundle, I could use the later if you wish it.

examples/jsm/OGC3DTilesHelper.js
Desplandis: Agree!

examples/3dtiles_loader.html
jailln: Ok I understand. From a user / tester perspective (which is the first thing a person does with itowns, before coding or using it in its project), I think it can feel weird that the page reloads. IMO we should not reload it and opt for the layer addition / removal (which also shows how to remove a layer :grin: ). WDYT? (this can be done for v1 if we agree on this and you don't have time now)

examples/jsm/OGC3DTilesHelper.js
Desplandis: When we'll have multiple packages ;)

examples/3dtiles_loader.html
Desplandis: Let's be more ambitious (not for now I agree) and provide utilities to maintain a list of layers + gui. :P
This could factorize some redundant code on multiple examples.

examples/3dtiles_loader.html
Desplandis: Done.

examples/3dtiles_loader.html
Desplandis: Done.

examples/jsm/OGC3DTilesHelper.js
Desplandis: I finally use 60°.

examples/3dtiles_loader.html
Desplandis: It seems that values in the `batchId` attribute of your tileset are not integers. I thought that the standard mandates that each of those values are integers?

>  In the binary glTF section, each vertex has an numeric `batchId` attribute in the integer range `[0, number of models in the batch - 1]`.

It does not seem to be an interleaved issue too.

examples/3dtiles_loader.html
jailln: I guess that would be better to not have this warning yes, but let's take care of it later

examples/3dtiles_loader.html
jailln: Let's merge for now but I don't think it comes from the tileset because we didn't have this issue before moving to `3d-tiles-renderer`. I added it to the list of stuff to check for V1.

examples/3dtiles_loader.html
jailln: Thanks, I updated the values to get a better lightning (based on empirical tests done for our software using itowns)

PR #2378, opened 2024-08-08
[OGC3DTiles] Fix old 3d tiles layer gltf parsing

jailln:
Fixes `C3DTilesLayer` gltf tiles conversion by correctly using the new `iGLTFLoader`.

### DIFF
--- a/src/Provider/3dTilesProvider.js
+++ b/src/Provider/3dTilesProvider.js
@@ -1,10 +1,12 @@
 import * as THREE from 'three';
 import B3dmParser from 'Parser/B3dmParser';
 import PntsParser from 'Parser/PntsParser';
-import iGLTFLoader from 'Parser/iGLTFLoader';
 import Fetcher from 'Provider/Fetcher';
 import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
 import PointsMaterial from 'Renderer/PointsMaterial';
+// A bit weird but temporary until we remove this deprecated provider. Mainly to benefit from the enableDracoLoader and enableKtx2Loader
+// methods.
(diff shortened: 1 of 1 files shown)


src/Provider/3dTilesProvider.js
```diff
@@ -1,10 +1,12 @@
 import * as THREE from 'three';
 import B3dmParser from 'Parser/B3dmParser';
 import PntsParser from 'Parser/PntsParser';
-import iGLTFLoader from 'Parser/iGLTFLoader';
 import Fetcher from 'Provider/Fetcher';
 import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
 import PointsMaterial from 'Renderer/PointsMaterial';
+// A bit weird but temporary until we remove this deprecated provider. Mainly to benefit from the enableDracoLoader and enableKtx2Loader
+// methods.
+import { itownsGLTFLoader } from 'Layer/OGC3DTilesLayer';
```
Desplandis, line 9: I think we should in a future version use a global [`LoadingManager`](https://threejs.org/docs/#api/en/loaders/managers/LoadingManager) (à la [`DefaultLoadingManager`](https://threejs.org/docs/index.html#api/en/loaders/managers/DefaultLoadingManager)) to register both `DracoLoader` and `KTXLoader`. This should avoid such code. x)

Desplandis (approved):
Looks good to me!

src/Provider/3dTilesProvider.js
jailln: Yes totally agree; and we should also transform our parsers to loaders

Issue #2175, opened 2023-08-28
How to place and rotate potree point cloud ?


ketourneau:
We try to load many potree point cloud in same view but we got placement problem.
We try to place the point clouds after loading them but it seems that bounding box is not updated so point cloud is not rendering correctly.
Does anyone know how we can rotate and move point cloud after loading them (or before) with bounding box updated ?

Point clouds are not necessarily georeferenced, we'd like to place them as in the [collada](https://github.com/iTowns/itowns/blob/master/examples/misc_collada.html) example : 

```
                // building coordinate
                const coord = new itowns.Coordinates(
                    'EPSG:4326', 4.2165, 44.844, 1417,
                );

                model.position.copy(coord.as(view.referenceCrs));
                // align up vector with geodesic normal
                model.lookAt(model.position.clone().add(coord.geodesicNormal));
                // user rotate building to align with ortho image
                model.rotateZ(-Math.PI * 0.2);
                model.scale.set(1.2, 1.2, 1.2);
```

ketourneau:
Since this [PR](https://github.com/iTowns/itowns/pull/2361) is accepted, we now place potree point cloud with parent object3d but we still got some rendering issue.

Point cloud is not correctly rendering (maybe bouding box still not updated ?) : https://demo-hackathon.netlify.app/examples/demo_hackathon_orvault_planarview

Source code here : https://github.com/bloc-in-bloc/itowns/blob/19f14c00e09280cc2e7065919a1a83c768df3f4a/examples/demo_hackathon_orvault_planarView.html#L306

PR #2381, opened 2024-08-22
fix(pointcloud): Fix point cloud distance to camera whatever point cloud inital placement is

Pourfex:
Fix point cloud distance to camera whatever point cloud inital placement is

## Description
Quick fix to compute camera distance to point (used for refresh the point cloud) whatever the initial placement of the point cloud is. 
This is used when the point cloud is inside a custom object with orientation and position.
Fix issue -> See : https://github.com/iTowns/itowns/issues/2175 (point cloud placement)
Related PR -> https://github.com/iTowns/itowns/pull/2361 (point cloud placement)
 
## Motivation and Context

In our scene, we want the point cloud to be placed wherever with any rotation. The previous code assumed an identity rotation and a local placement ; and the refresh didn't happened whith custom orientation and position because the point compute was wrong, leading to wrong distance to camera thus leading to no refresh.

So now, the point cloud object can be placed wherever with any rotations and still be refresh according to camera's position.

MacOS - Chrome - latest itowns version (2.43.2)

Before : 
image

Fixed : 
image

### DIFF
--- a/src/Layer/PointCloudLayer.js
+++ b/src/Layer/PointCloudLayer.js
@@ -253,7 +253,8 @@ class PointCloudLayer extends GeometryLayer {
         }
 
         elt.notVisibleSince = undefined;
-        point.copy(context.camera.camera3D.position).sub(this.object3d.position);
+        point.copy(context.camera.camera3D.position).sub(this.object3d.getWorldPosition(new THREE.Vector3()));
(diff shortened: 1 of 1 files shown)


Desplandis (approved):
This seems to correctly fix the aforementioned issue. Let's merge it!

PR #2382, opened 2024-08-23
fix(PointCloud): Fix clipping point cloud

ketourneau:
## Description
Actually ClippingPlanes on PointMaterial doesn't clip the point cloud (point cloud totally disappear).
It seems it was a mistake on PointsFS.glsl.
As shown in [this discussion ](https://stackoverflow.com/questions/42532545/add-clipping-to-three-shadermaterial) it need to use ```#include <clipping_planes_fragment>``` in main function and ```#include clipping_planes_pars_fragment.glsl``` only on top.

## Motivation and Context
We need to clip point cloud with ClippingPlanes.

## Screenshots
Capture d’écran 2024-08-23 à 15 24 35
Capture d’écran 2024-08-23 à 15 24 42

### DIFF
--- a/src/Renderer/Shader/PointsFS.glsl
+++ b/src/Renderer/Shader/PointsFS.glsl
@@ -17,7 +17,7 @@ uniform int shape;
 void main() {
 
 // Early discard (clipping planes and shape)
-#include <clipping_planes_pars_fragment>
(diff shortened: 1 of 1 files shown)


Desplandis (approved):
A mistake on my part... Thanks for the fix!

PR #2385, opened 2024-09-02
release v2.44.0

mgermerie:
## Description

New release including 3d tiles migration

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,110 @@
+<a name="2.44.0"></a>
+# [2.44.0](https://github.com/iTowns/itowns/compare/v2.43.1...v2.44.0) (2024-09-02)
+
+
+### Features
+
+* **3dtiles:** add new OGC3DTilesLayer using 3d-tiles-renderer-js ([a2fedd8](https://github.com/iTowns/itowns/commit/a2fedd8))
+* Add support for Cloud Optimized Point Clouds (COPC) ([f1e014f](https://github.com/iTowns/itowns/commit/f1e014f))
+* **COG:** Allow tileWidth, tileHeight & resampleMethod parameter ([24ab82f](https://github.com/iTowns/itowns/commit/24ab82f))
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -1,3 +1,110 @@
+<a name="2.44.0"></a>
+# [2.44.0](https://github.com/iTowns/itowns/compare/v2.43.1...v2.44.0) (2024-09-02)
+
+
+### Features
+
+* **3dtiles:** add new OGC3DTilesLayer using 3d-tiles-renderer-js ([a2fedd8](https://github.com/iTowns/itowns/commit/a2fedd8))
```
jailln, line 7: A part of the commit is missing, could you add it manually please ? (maybe in new lines ? WDYT?):
```
Deprecate C3DTilesLayer (replaced by OGC3DTilesLayer).
Add new iGLTFLoader that loads gltf 1.0 and 2.0 files.
```

mgermerie: Multiple lines commit message never fully appear in changelog. As also done for [this commit for instance](https://github.com/iTowns/itowns/commit/d4779cc), the [matching changelog part](https://github.com/iTowns/itowns/blob/master/changelog.md#2430-2024-04-02) only displays the first line of the commit - except for breaking changes tagged commits where the message appears at the end of the changelog.

I can add them manually but I fear it could break the changelog generation for future releases, as it happened in the past when manually sorting incorrectly named commits in their relevant category. We'll have to be careful next time we release.

jailln: Mmmh ok I see thanks. In this case I'm all for not doing it because it's not worth breaking changelog generation for all future releases. I'll be more careful in my commits message next time.

Desplandis (approved):
Tested and approved!

changelog.md
mgermerie: Okay, I removed the added lines then :)

jailln (approved):

PR #2386, opened 2024-09-03
fix(3dtiles): support point cloud attenuated mode in OGC3DTiles layer

jailln:
## Description

Fix point size attenuated mode for `OGC3DTilesLayer` - same changes that what was done in #2325 for `C3DTilesLayer`.

I think the `scale` uniform could be renamed to `cameraPreSse` since it is always set to this value and because it doesn't really represent the scale of the point cloud which can be misleading for someone using that material. WDYT ? If we agree on renaming it I can push this in this PR.

cc @AnthonyGlt @airnez


Desplandis:
I used the `scale` uniform name for uniformity with the default [`points` shader](https://github.com/mrdoob/three.js/blob/master/src/renderers/shaders/ShaderLib/points.glsl.js) of `three`. In our context, we are using the `sse` as a scaling value but this will not always be the case.

Desplandis (approved):

PR #2272, opened 2024-02-14
PointCloudData reprojection into view CRS ('a la volee')

ftoromanoff:
The purpose of the PR is to reproject 'a la volee' PopintCloudData (entwine, COPC , Potree1 and 2) data into the view crs.

PR #2392, opened 2024-09-05
PoincLoudLayer and Multisource

ftoromanoff:
based on etwineAlaVolee #2272 

Testing 2 way on COPC data:
First commit :
Adding a MultiSource.js file calling source on each url

Second commit
Adding a property isMulitple for sources

-----
done for entwine
fixed for Potree and Potree2


jailln:
@ftoromanoff that would be great to have :)

PR #2407, opened 2024-09-16
[Docs] Fix non-strict jsdoc comments

Desplandis:
## Description

When testing migration from jsdoc to tsdoc, I found that a number of jsdoc comments are actually not valid (and jsdoc silently ignore them...).

This PR includes:
- [x] Support of `typedef` in the jsdoc template
- [x] Removal of empty class used for `typedef` documentation and used as a sort of "fix" of the point above.
- [x] Removal of `@constructor`, `@class` and `@classdec` which are leftovers of pre-class javascript and not valid in post-ES5.
- [x] Various typecheck issues.

### DIFF
--- a/src/Core/3DTiles/C3DTBoundingVolume.js
+++ b/src/Core/3DTiles/C3DTBoundingVolume.js
@@ -85,7 +85,6 @@ function initFromSphere(sphere) {
 }
 
 /**
- * @classdesc 3D Tiles
  * [bounding volume](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/schema/boundingVolume.schema.json)
  * Used to represent bounding volumes and viewer request volumes. The input bounding volume (from the dataset) can be a
  * box, a sphere or a region. Regions are transformed to spheres internally for simplification of parsing and to speed
@@ -143,7 +142,7 @@ class C3DTBoundingVolume {
     }
 
     /**
-     * Checks if the camera is inside the [viewer request volumes](@link https://github.com/CesiumGS/3d-tiles/tree/main/specification#viewer-request-volume).
+     * Checks if the camera is inside the [viewer request volumes](https://github.com/CesiumGS/3d-tiles/tree/main/specification#viewer-request-volume).
      * @param {Camera} camera - the camera to perform culling for
      * @param {THREE.Matrix4} tileMatrixWorld - the world matrix of the tile
      * @returns {boolean} true if the camera is outside the viewer request volume, false otherwise.
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -8,7 +8,7 @@ import Coordinates from 'Core/Geographic/Coordinates';
 
 import itowns_stroke_single_before from './StyleChunk/itowns_stroke_single_before.css';
 
-export const cacheStyle = new Cache();
+const cacheStyle = new Cache();
 
 const matrix = document.createElementNS('http://www.w3.org/2000/svg', 'svg').createSVGMatrix();
 const canvas = document.createElement('canvas');
@@ -169,8 +169,7 @@ function defineStyleProperty(style, category, parameter, userValue, defaultValue
 }
 
 /**
- * @class
- * @classdesc StyleContext stores metadata of one FeatureGeometry that are needed for its style computation:
+ * StyleContext stores metadata of one FeatureGeometry that are needed for its style computation:
  * type of feature and what is needed (fill, stroke or draw a point, etc.) as well as where to get its
  * properties and its coordinates (for base_altitude).
  *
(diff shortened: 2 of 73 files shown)


src/Core/3DTiles/C3DTBoundingVolume.js
```diff
@@ -143,7 +142,7 @@ class C3DTBoundingVolume {
     }
 
     /**
-     * Checks if the camera is inside the [viewer request volumes](@link https://github.com/CesiumGS/3d-tiles/tree/main/specification#viewer-request-volume).
+     * Checks if the camera is inside the [viewer request volumes](https://github.com/CesiumGS/3d-tiles/tree/main/specification#viewer-request-volume).
```
mgermerie, line 145: External links as this one do not work. I believe this is due to the fact that doc pages are embedded into the template page, and the links are opened into the embed container, and not in a new page.
I just checked and this is also the case on master so this was not introduced in your PR.

src/Core/Style.js
```diff
@@ -335,7 +337,7 @@ function _addIcon(icon, domElement, opt) {
  * polygons will be extruded by the specified amount
  *
  * @property {Object} [stroke] - Lines and polygons edges.
- * @property {String|Function|THREE.Color} [stroke.color] The color of the line. Can be any [valid
+ * @property {String|Function|THREE.Color} [stroke.color] The color of the line. Can be any AAA [valid
```
mgermerie: Maybe some residuals of a vim shortcut ?

src/Layer/Layer.js
```diff
@@ -52,17 +52,17 @@ class Layer extends THREE.EventDispatcher {
...
+     * @param {StyleOptions} [config.style] - an object that contain any properties
      * (order, zoom, fill, stroke, point, text or/and icon)
      * and sub properties of a Style (@see {@link StyleOptions}). Or directly a {@link Style} .<br/>
      * When entering a StyleOptions the missing style properties will be look for in the data (if any)
      * what won't be done when you use a Style.
      * @param {number} [config.cacheLifeTime=Infinity] - set life time value in cache.
-     * This value is used for [Cache]{@link Cache} expiration mechanism.
-     * @param {(boolean|Object)} [config.addLabelLayer=false] - Used to tell if this layer has
+     * This value is used for cache expiration mechanism.
+     * @param {Object} [config.addLabelLayer] - Used to tell if this layer has
```
mgermerie: Doesn't this needs to still be typed as `boolean|Object` since it still can be a boolean ?

src/Parser/VectorTileParser.js
```diff
@@ -187,13 +187,13 @@ export default {
      * Parse a vector tile file and return a [Feature]{@link module:GeoJsonParser.Feature}
      * or an array of Features. While multiple formats of vector tile are
      * available, the only one supported for the moment is the
-     * [Mapbox Vector Tile]{@link https://www.mapbox.com/vector-tiles/specification/}.
+     * [Mapbox Vector Tile](https://www.mapbox.com/vector-tiles/specification/).
      *
      * @param {ArrayBuffer} file - The vector tile file to parse.
      *
-     * @param {ParsingOptions} options - Options controlling the parsing {@link ParsingOptions}.
```
mgermerie, line 194: Why not keeping the specific type of `options` ?

src/Core/Style.js
Desplandis: Yeah, you're right!

src/Layer/Layer.js
Desplandis: Does the `true` value is actually used in the code?
It seems the only usage of `addLabelLayer` if it is not `undefined` is in the code below and it is used as an object (a `true` value creates a type error) https://github.com/iTowns/itowns/blob/ff9a86adf64ef3626b5d3542ef2145b3fd8eda32/src/Core/View.js#L66-L84

src/Parser/VectorTileParser.js
Desplandis: Because we declare `options.in` in both `ParsingOptions` and in the lines below, this creates incompatible types. In fact if we declare a field to a paramater, it shall have an `{Object}` type.

src/Core/3DTiles/C3DTBoundingVolume.js
jailln: The syntax that is working is [text](url)

jailln (approved):
Thanks!

src/Layer/Layer.js
mgermerie: It is used in [this example](https://github.com/iTowns/itowns/blob/ff9a86adf64ef3626b5d3542ef2145b3fd8eda32/examples/source_file_geojson_raster_2.html#L107). It should not create a type error in the place you mentioned since `true.someProperty` will be evaluated as `undefined` ([here](https://github.com/iTowns/itowns/blob/ff9a86adf64ef3626b5d3542ef2145b3fd8eda32/src/Core/View.js#L79) and [here](https://github.com/iTowns/itowns/blob/ff9a86adf64ef3626b5d3542ef2145b3fd8eda32/src/Core/View.js#L83)). This is not a clean way of coding though, I admit it.

Desplandis: Forgot that literal values could have method... Fixed it!

mgermerie (approved):

PR #2409, opened 2024-09-17
feat(controls): add state controls at view init

AnthonyGlt:
## Description
Add possibility to define the controls at the view initialization

## Motivation and Context
The controls can be redefined afterwards already but I wondered why it couldn't be at the init

### DIFF
--- a/src/Controls/GlobeControls.js
+++ b/src/Controls/GlobeControls.js
@@ -180,6 +180,8 @@ let previous;
  * @property      {boolean} enableDamping Enable damping or not (simulates the lag that a real camera
  * operator introduces while operating a heavy physical camera). Default is true.
  * @property      {boolean} dampingMoveFactor the damping move factor. Default is 0.25.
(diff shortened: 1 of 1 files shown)


Desplandis:
There is already a lot of options in `GlobeView` but why not?
@mgermerie What do you think?

@AnthonyGlt Could please document it too in `GlobeView`?

AnthonyGlt:
There is already:
```js

@param {Object} [options.controls] - See options of {@link GlobeControls}

``` 
in GlobeView

Should I also add :
```js
@param {StateControl~State} [options.controls.stateControl] redefining which controls state is triggered by the keyboard/mouse
event (For example, rewrite the PAN movement to be triggered with the 'left' mouseButton instead of 'right'

```
 ?

Desplandis:
Yeah you're right, this is not correctly typed for now... At least there is a link to `GlobeControls`, so you can ignore my previous comment. =)

Desplandis (approved):

mgermerie (approved):
:+1:

Issue #2369, opened 2024-07-24
[Proposal] use an alternative to @mapbox/mapbox-gl-style-spec


Desplandis:
This issue is a **feature proposal**. Feel free to upvote (with :+1: ), comment and provide your use-cases if you're interested by this feature.

## Context
As described in #2234 (and more recently in #2342), recent versions of `@mapbox/mapbox-gl-style-spec` (>13.28.0) use a non-FOSS license (the standard Mapbox license) and is thus incompatible with our current license. This proposal aims to discuss potential replacements of this dependency.

## Description of the proposal
An alternative seems to be `@maplibre/maplibre-gl-style-spec`, but we may want to use another library?

jailln:
Done in #2466

PR #2342, opened 2024-06-21
chore(deps): June 2024 bump

Desplandis:
## Description

Update all dependencies (when possible). A lot of dependencies now mandates node >= 18.0.0, shall we mark this as breaking changes?

Previous major bump in #2234

## Peer dependencies
- [x] `three` (`0.159.0` to `0.165.0`): Removed WebGL1 (no issue for us).
- [x] `proj4` (`2.9.2` to `2.11.0`): No breaking changes

## Dependencies
Those are the packages with a new version available:
- [x] `earcut`: Distributes as ESM now.
- [x] `@mapbox/mapbox-gl-style-spec` (`13.28.0`): Still the same problems as #2234. Will not update due to the fact this is the latest version with a FOSS license (`ISC`). This is a huge **red-flag**, shall we consider moving to `@maplibre/maplibre-gl-style-spec` @iTowns/reviewers?
- [x] `@mapbox/vector-tile`: Distributes as ESM now.
- [x] `pbf`: Distributes as ESM now.
- [x] `shpjs` (`4.0.4` to `5.0.1`): Distributes as ESM now. No issue since we already migrated to ESM.
- [x] `@tweenjs/tween.js` (`18.6.4` to `23.1.2`): Distributes as ESM now. A few breaking changes which does not impact us.

## Developer dependencies
Those packages upgrades introduce no breaking change:
- [x] `@babel/cli`, `@babel/core` and the following plug-ins: `@babel/plugin-transform-runtime` and `@babel/preset-env`.
- [x] `c8`.
- [x] `chart.js`.
- [ ] `conventional-changelog-cli`: Pure ESM package, issues with our CJS script.
- [x] `core-js`.
- [ ] `eslint` and the following plug-in: `eslint-plugin-import`: cannot update to ESLint 9 since the plugin requires ESLint8.
- [x] `https-proxy-agent`.
- [x] `jsdoc`.
- [x] `marked`: unused, remove (added in https://github.com/iTowns/itowns/pull/612)
- [ ] `mocha`: update cause deprecation notice of `glob` dependency.
- [ ] `node-fetch`: Can update to `3.x.x` since we are now a ESM only package. TypeError with fetcher.js test. Note that node 18 support natively the `fetch` API.
- [x] `puppeteer`.
- [x] `replace-in-file`.
- [x] `sinon`.
- [x] `typescript`.
- [x] `webpack` and the following plugin: `webpack-dev-server` and `eslint-webpack-plugin`.

Issue #2247, opened 2023-12-22
License issue with @mapbox/mapbox-gl-style-spec


Desplandis:
## Context

Since version [`14.0.0`](https://www.npmjs.com/package/@mapbox/mapbox-gl-style-spec/v/14.0.0), `@mapbox/mapbox-gl-style-spec` does not use a FOSS licence (ISC) anymore:

> This project uses the standard Mapbox license, which is designed to provide flexibility for our customers and a sustained foundation for the development of our technology. Please consult LICENSE.txt for its specific details.

For now, we are using the latest version (`13.28.0`) released with an ISC licence (see #2234) but we'll not be able to update this package in the future.

## Possible Solution
A possible solution would be to move to `@maplibre/maplibre-gl-style-spec`. However, we'll have to assert the differences between those two.

jailln:
Done in https://github.com/iTowns/itowns/pull/2466

PR #2410, opened 2024-09-18
chore(deps): September 2024 bump

jailln:
## Description

Update all dependencies (when possible).

Previous major bump in https://github.com/iTowns/itowns/pull/2342

## Peer dependencies

- [x] `three` (`0.165.0` to `0.168.0`)

~~A bug appears in `0.167.1`: Zooming in a view with a perspective camera does not work anymore. I've narrowed the problem down to the `buffer` computed in [View.getPickingPositionFromDepth](https://github.com/iTowns/itowns/blob/c1409c509328c02aa378ec2749a6e3d9b0885ba5/src/Core/View.js#L1030) which is different with `three` `0.166.1` and `0.167.1`. 
I also found the responsible commit in `threejs` using `yarn link` and `git bisect`. It comes from the [following PR](https://github.com/mrdoob/three.js/pull/28789) (note that the commit hash is `a38bbdd6fbe971049f2083d79bbca2cda7972a44` and not the one in the PR). I haven't found out the buug yet, it's quite low level :sweat_smile: cc @Desplandis if you have any ideas~~ => Solved

- [x] `proj4` (`2.11.0` to `2.12.1`): No breaking changes

## Dependencies

- [ ]  `@mapbox/mapbox-gl-style-spec`  (`13.28.0`): Still the same problems as #2342 . Won't update here, see #2247 and #2369.
- [x] `@tweenjs/tween.js` (`23.1.2` to `25.0.0`): Updates to `StreetControls` and `CameraUtils` to use the new way to add tweens to the tween Group.
- [x] `shpjs` (`6.0.1` to `6.1.0`): No breaking changes

## Developer dependencies
- [x] `@babel/cli`, `@babel/core` and the following plug-ins: `@babel/plugin-transform-runtime` and `@babel/preset-env`.
- [x] `babel-loader`
- [x] `chart.js`
- [ ] conventional-changelog-cli: same as #2342 - Pure ESM package, issues with our CJS script.
- [x] `core-js`
 - [ ] `eslint`: same as #2342 - cannot update to ESLint 9 because `eslint-plugin-import` requires ESLint8.
- [x] `eslint-plugin-import`
- [x] `eslint-webpack-plugin`
- [x] `mocha`
- [x] `node-fetch`: removed since node 18 natively supports the fetch API.
- [x] `puppeteer`
- [x] `replace-in-file`
- [x] `sinon`
- [x] `typescript`
- [x] `webpack`
- [x] `webpack-dev-server`
- [x] `@xmldom/xmldom`

~~I've also noticed that the [Fetcher tests](https://github.com/iTowns/itowns/blob/c1409c509328c02aa378ec2749a6e3d9b0885ba5/test/unit/fetcher.js) are not run by the CI and fail locally (same on master). I haven't look at why they don't run on the CI yet, if someone knows that would be greatly appreciated :) The local failure is quite weird: it only fails un run mode but succeeds in debug mode :sweat_smile:~~ => Solved


Desplandis:
I'm all (in a near future) for removing scripts that are maybe not necessary anymore. That includes:
- The `_BATCHID` substitution script for `GLTFLoader`
- The copying of `three/examples/jsm` scripts into the `ThreeExtended` directory. We could relax the linting rules on `.js` extension and let the bundle bundle all ESM modules together.

This could remove 3/4 dev dependencies in my opinion.

Desplandis:
For the three.js issue (thanks to your bisect), we can assume that since the packing in GPU from depth to RGBA has changed, our unpack in CPU (see the code below) is no longer valid...

https://github.com/iTowns/itowns/blob/e38663701ff8f7ddefedb0721bb61fa7c28078db/src/Renderer/LayeredMaterial.js#L13-L23

Desplandis:
Weird for the `test/unit/fetcher.js` test, it works locally for me.

jailln:
> I'm all (in a near future) for removing scripts that are maybe not necessary anymore. That includes:
> 
> * The `_BATCHID` substitution script for `GLTFLoader`
> * The copying of `three/examples/jsm` scripts into the `ThreeExtended` directory. We could relax the linting rules on `.js` extension and let the bundle bundle all ESM modules together.
> 
> This could remove 3/4 dev dependencies in my opinion.

Totally agree with you! 

> For the three.js issue (thanks to your bisect), we can assume that since the packing in GPU from depth to RGBA has changed, our unpack in CPU (see the code below) is no longer valid...

Yes, that was it! Thanks to you and @AnthonyGlt for finding that

> Weird for the test/unit/fetcher.js test, it works locally for me.

It was related to the node version. The test works locally for me when I switch to node 18. However it is still not run by the ci but this is not related to this PR.

src/Controls/StreetControls.js
```diff
@@ -310,6 +310,8 @@ class StreetControls extends FirstPersonControls {
             })
             .start();
 
+        this.tweenGroup.add(this.tween);
+
```
Desplandis, line 314: From `tween.js` [changelog](https://github.com/tweenjs/tween.js/releases/tag/v24.0.0):

> Tweens are no longer automatically added or removed from groups by default when you call any Tween methods such as start(), stop(), or pause(), and the preserve parameter to Group.update() now defaults to true and is deprecated to be removed in a future major version.

As such, tweens are not removed in the group and updated continuously (they were automatically removed before the update).

Desplandis (changes requested):

src/Controls/StreetControls.js
jailln: Bien vu; done :)

Desplandis (approved):
Thanks for the fixes =)

PR #2412, opened 2024-09-24
[Refactoring] Split the Extent module into a geographic and a tiled ones

Desplandis:
## Before contributing

Read [CONTRIBUTING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) and [CODING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) to apply `iTowns` conventions on PRs, Git history and code.

## Description
This PR refactor and split the `Extent` module into two sub-modules—one for geographic extents (`Extent`) and one for tiled extents (`Tile`).

## Motivation and Context
The current methods of `Extent` are all guarded by assertions checking if the current instance is of geographic or tiled type, with a different behavior given the type. Moreover certain methods are only valid for one type. This does not help the tracking of the current type of `Extent`.
Finally this module currently contributes to at least 25% of all type errors and need a cleaner structure for proper typing.


src/Converter/Feature2Mesh.js
```diff
@@ -65,8 +65,12 @@ class FeatureMesh extends THREE.Group {
                 // calculate the scale transformation to transform the feature.extent
                 // to feature.extent.as(crs)
                 coord.crs = Crs.formatToEPSG(this.#originalCrs);
-                extent.copy(this.extent).applyMatrix4(this.#collection.matrix);
-                extent.as(coord.crs, extent);
+                if (this.extent.isExtent) {
+                    extent.copy(this.extent).applyMatrix4(this.#collection.matrix);
+                    extent.as(coord.crs, extent);
+                } else {
+                    this.extent.toExtent(coord.crs, extent);
```
jailln, line 75: Shouldn't it always be an Extent now? In which case is it a Tile?

src/Converter/textureConverter.js
```diff
@@ -27,7 +27,7 @@ export default {
                 new THREE.Color(backgroundLayer.paint['background-color']) :
                 undefined;
 
-            extentDestination.as(CRS.formatToEPSG(layer.crs), extentTexture);
+            extentDestination.toExtent(CRS.formatToEPSG(layer.crs), extentTexture);
```
jailln: shouldn't we name it destinationTile instead for more clarity?

src/Core/Geographic/Extent.js
```diff
@@ -23,89 +19,50 @@ const cNorthEast =  new Coordinates('EPSG:4326', 0, 0, 0);
...
+     * @param {number} [v2] south value
      * @param {number} [v3] north value
      */
     constructor(crs, v0, v1, v2, v3) {
         if (CRS.isGeocentric(crs)) {
             throw new Error(`${crs} is a geocentric projection, it doesn't make sense with a geographical extent`);
         }
 
+        if (CRS.isTms(crs)) {
+            throw new Error(`${crs} is a tiled projection, use a geographic one`);
```
jailln: Use Tile instead?

src/Core/Geographic/Tile.js
```diff
@@ -0,0 +1,256 @@
...
+    return r;
+}
+
+const _extent = new Extent('EPSG:4326', [0, 0, 0, 0]);
+const _extent2 = new Extent('EPSG:4326', [0, 0, 0, 0]);
+
+const _c = new Coordinates('EPSG:4326', 0, 0);
+
+export const globalExtentTMS = new Map();
+export const schemeTiles = new Map();
```
jailln: I think we should take these globals out of the Tile file, but this may be out of scope of this PR. WDYT?

src/Layer/LabelLayer.js
```diff
@@ -238,7 +238,11 @@ class LabelLayer extends GeometryLayer {
         const labels = [];
 
         // Converting the extent now is faster for further operation
-        extent.as(data.crs, _extent);
+        if (extent.isExtent) {
```
jailln: you need to change the type of the extent parameter to Extent|Tile

src/Provider/URLBuilder.js
```diff
@@ -102,11 +112,10 @@ export default {
         if (source.bboxDigits !== undefined) {
             precision = source.bboxDigits;
         }
-        bbox.as(source.crs, extent);
-        const w = extent.west.toFixed(precision);
-        const s = extent.south.toFixed(precision);
-        const e = extent.east.toFixed(precision);
-        const n = extent.north.toFixed(precision);
+        const w = bbox.west.toFixed(precision);
```
jailln, line 115: there is no case where the bbox is not in the source crs already?

src/Source/VectorTilesSource.js
```diff
@@ -142,8 +142,8 @@ class VectorTilesSource extends TMSSource {
         });
     }
 
-    urlFromExtent(extent, url) {
-        return URLBuilder.xyz(extent, { tileMatrixCallback: this.tileMatrixCallback, url });
+    urlFromExtent(tile, url) {
```
jailln, line 145: rename to urlFromTile ?

src/Source/WFSSource.js
```diff
@@ -166,7 +169,10 @@ class WFSSource extends Source {
         }
     }
 
-    urlFromExtent(extent) {
+    urlFromExtent(extentOrTile) {
+        const extent = extentOrTile.isExtent ?
```
jailln, line 173: Should we follow this pattern in every similar place ? (extentOrTile as parameter and create an extent variable?)

Also, I'm not fond on having to perform this check in the calling code each time. Some thoughts, let me know what you think:
* Rename the method to `as` or `toExtent` in both classes so we don't have to perform this check
* Complementary: Should we have a common interface for these two classes ? and a base class that defines this interface? If yes, should we name the base class `Extent` and the childs `GeographicExtent` (instead of Extent) and `TileExtent` (instead of Tile)?

test/unit/extent.js
```diff
@@ -144,37 +145,37 @@ describe('Extent', function () {
     });
 
     it('should clone tiled extent like expected', function () {
-        const withValues = new Extent('TMS:4326', zoom, row, col);
-        const clonedExtent = withValues.clone();
-        assert.equal(clonedExtent.zoom, withValues.zoom);
-        assert.equal(clonedExtent.row, withValues.row);
-        assert.equal(clonedExtent.col, withValues.col);
+        const withValues = new Tile('TMS:4326', zoom, row, col);
```
jailln: Can you make a new file for Tile tests please?

test/unit/featuregeometrylayererror.js
```diff
@@ -75,6 +75,7 @@ files.forEach((geojson, i) => {
         });
 
         it('update without proj4', function (done) {
+            console.error('TOP KEKE');
```
jailln: oups :)

jailln (commented):
Thanks for this much needed refactoring :)

test/unit/featuregeometrylayererror.js
Desplandis: Yikes... Forgot about this one. xD

test/unit/extent.js
Desplandis: Agree, I was a bit lazy on this one. x)

src/Source/WFSSource.js
Desplandis: I think the method is currently ill-named... From what I inferred from the usage, it seems to be building an url from a compatible "requester" (cf `OrientedImageSource` below).

https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Source/OrientedImageSource.js#L41-L43

As for your next remark, the check was implicitly done in the `Extent#as` method pre-PR. And I choosed to rename `as` to `toExtent` in `Tile` as I think they do not have the same semantics (the former is used to convert an instance to an instance of the same class, the latter to return the corresponding extent).

src/Source/VectorTilesSource.js
Desplandis: Or urlFromRequester

src/Provider/URLBuilder.js
Desplandis: I prefered to move the conversion before calling this function (cf first commit) to make it pure. As such `bbox` does one and one thing only and it is easier to track changes of crs.

src/Core/Geographic/Tile.js
Desplandis: Yup agree!
I considered them as a kind of proto-`Tilegrid`. x)

src/Converter/textureConverter.js
Desplandis: Yup agree.

src/Converter/Feature2Mesh.js
Desplandis: This is one of the pain point with this PR, it seems that pre-PR the `Extent` is tiled one fore Mapbox Vector Tiles and a geographic one for WFS...

I choose not to fix it for now, hoping that migrating to typescript will help us here (at least tracking the types and incompatible one).

But if you think we should fix it here, I will need some more time and help. x)

jailln: I'm ok with fixing it later but I think we should write it somewhere to not forget about it (either in the ts migration proposal or in a new issue, or in a todo)

src/Provider/URLBuilder.js
jailln: Ok, sorry I missed that. In this case, I think we should verify this assumption and either throw an error if it's not met or a log a warning and do the crs conversion; just to avoid future bug.

src/Source/VectorTilesSource.js
jailln: Yes why not but I just remembered that this method is shared amongst all sources so we won't change it now

src/Core/Geographic/Extent.js
```diff
@@ -23,89 +19,50 @@ const cNorthEast =  new Coordinates('EPSG:4326', 0, 0, 0);
...
+        }
+
         this.isExtent = true;
         this.crs = crs;
+
         // Scale/zoom
         this.zoom = 0;
```
jailln: Why can't we remove the zoom property now?

src/Source/WFSSource.js
jailln: Ok, I've given it a some more thoughts and haven't found a way to simplify this without keeping clode clarity so let's keep it like this

jailln (approved):
Thanks for the changes :)

PR #2413, opened 2024-09-24
[Chore] Update babel and ES-lint config to support TypeScript

mgermerie:
## Before contributing

Read [CONTRIBUTING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) and [CODING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) to apply `iTowns` conventions on PRs, Git history and code.

## Description

Update configuration from webpack, babel (thanks to @Desplandis on these) and eslint to support TypeScript.
Babel and Webpack are updated to resolve .ts files at import. Eslint uses `@typescript-eslint/parser` plugin to lint TypeScript files, applying the same rules as for JavaScript files with eventual specific rules added.

### DIFF
--- a/config/babel-register/babel-hooks.mjs
+++ b/config/babel-register/babel-hooks.mjs
@@ -71,7 +71,18 @@ async function transpile(source, context) {
  * the Node.js default resolve hook after the last user-supplied resolve hook
  */
 export async function resolve(specifier, context, nextResolve) {
-    return nextResolve(specifier, context);
+    // Try to resolve the path of an imported module.
+    // If the resolver failed, retry substituting the extension with '.ts'.
+    try {
+        return await nextResolve(specifier, context);
+    } catch (err) {
+        if (err.code === 'ERR_MODULE_NOT_FOUND') {
+            const url = new URL(specifier, context.parentURL);
+            url.pathname = url.pathname.replace(/\.[^/.]+$/, '.ts');
+            return nextResolve(url.href, context);
+        }
+        throw err;
+    }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


config/babel-register/babel-hooks.mjs
```diff
@@ -71,7 +71,16 @@ async function transpile(source, context) {
  * the Node.js default resolve hook after the last user-supplied resolve hook
  */
 export async function resolve(specifier, context, nextResolve) {
-    return nextResolve(specifier, context);
+    try {
+        return await nextResolve(specifier, context);
+    } catch (err) {
+        if (err.code === 'ERR_MODULE_NOT_FOUND') {
```
jailln, line 79: Can you comment why you need this specific case?

config/babel-register/babel-hooks.mjs
```diff
@@ -87,7 +96,14 @@ export async function resolve(specifier, context, nextResolve) {
  * the last user-supplied load hook
  */
 export async function load(url, context, nextLoad) {
-    const { format, shortCircuit, source } = await nextLoad(url, context);
+    const { format, shortCircuit, source } = await nextLoad(url, context)
+        .catch(async (error) => {
+            if (error.code === 'ERR_UNKNOWN_FILE_EXTENSION') {
```
jailln, line 106: Can you comment why you need this specific case?

config/babel-register/babel-hooks.mjs
Desplandis: We first try to resolve the path of the imported module as a `.js` file. However, if such module is a `.ts` file, we retry the resolver substituing the extension from `.js` to `.ts`.

jailln: that's what I inferred, I think it's worth adding your response as a comment in the code

config/babel-register/babel-hooks.mjs
Desplandis: The default loader can load ESM and CJS '.js' files, however '.ts' files shall be loaded with `format: module`. I think we could directly load with the module format if all modules are guaranteed to be ESM.

Desplandis: For more information, `mocha` is loaded as an `CJS` module so we can't directly load with the `format: module`.

jailln (approved):
Thanks for adding the comments :+1:

PR #2268, opened 2024-02-13
chore(workflow): update GitHub actions to use Node.js 20

mgermerie:
## Before contributing

- [x] I read [CONTRIBUTING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) and [CODING.md](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md) to apply `iTowns` conventions on PRs, Git history and code.

## Description

Update GitHub actions used in [integration workflow](https://github.com/iTowns/itowns/blob/master/.github/workflows/integration.yml) to use Node.js 20.x.

## Motivation and Context

The currently used Node.js 16.x GitHub actions are deprecated.

## TODO before merging

- [ ] Test on a local deployment environment
- [ ] Find replacement or workaround for non maintained or updated actions : 
  - [`gsactions/commit-message-checker@v2`](https://github.com/GsActions/commit-message-checker) : two PR ([here](https://github.com/GsActions/commit-message-checker/pull/101) and [here](https://github.com/GsActions/commit-message-checker/pull/102)) exist on the matter, but the repository seems not maintained.
  - [`coverallsapp/github-action@master`](https://github.com/coverallsapp/github-action), 
  - [`peaceiris/actions-gh-pages@v3`](https://github.com/peaceiris/actions-gh-pages), 
  - [`actions/create-release@v1`](https://github.com/actions/create-release), 
  - [`actions/upload-release-asset@v1`](https://github.com/actions/upload-release-asset)

PR #2415, opened 2024-09-27
Migration to 3D Tiles renderer followup

jailln:
Followup on 3D tiles renderer migration - see v1 of #2225:

* Update 3dtiles_loader example:
     * don't reload the page each time we change the url of the tileset
     * Add a terrain (world)  
     * Remove the other 3D Tiles examples
     * Add buttons to load pre-defined 3D tiles datasets
 * Add a deprecation warning in C3DTilesLayer constructor: will be removed in 3.0
 * Add more doc on `OGC3DTilesLayer`

**Notes:** 
I started implementing functional tests on [this branch](https://github.com/jailln/itowns/tree/3d-tiles-renderer-functional-test) but haven't fully succeeded yet and don't have much more time to put into it now. Basically I wanted to click on each button of the example, wait until the camera has moved and the tileset is loaded and then pick the scene to verify the 3D tiles layer is there. I tried several solutions but didn't manage to click on the buttons (or at least it didn't trigger the corresponding load methods) but directly calling the methods bound to onclick works ok. The main pain point is that `waitUntilItownsIsIdle` does not work in this case. I believe it's because it's based on the `command-queue-empty` which is triggered before the tileset has finished loading since tileset loading is now managed by 3DTilesRenderer. We need to find a way to know when tileset loading is over in 3DTilesRendererJS and we may also need to account for that before dispatching `command-queue-empty`. @Desplandis @AnthonyGlt any thoughts on this?


AnthonyGlt:
Thank you for this !

I think that we should wait for the zoom to end and then wait for the tiles to load.
For the zoom, there are events triggered from `lookAtCoordinate` which is called when we zoom, like `animation-ended`.

Concerning the loaded tiles, I didn't find any event worthwhile to use from the lib. Could be interesting to open an issue for this.
We could try to access to these values  `tiles.stats.downloading == 0`, `tiles.stats.parsing == 0`, `tiles.visibleTiles.size > 0`, but it means checking them at each rendering loop :/

EDIT:
There is this event `tiles-load-end `that looks like the one we need

Desplandis (approved):

jailln:
@AnthonyGlt Thanks, using both these events does the job :)

@Desplandis I added functional tests and also updated the object returned by the picking to conform better to what's returned for other layers (types will be more than welcome for such cases :) )

jailln:
Sooo, the tests pass locally but not on the CI :sweat_smile: I tried a few things to fix them on the ci but didn't succeed. If you guys have any idea let me know, otherwise I'll just remove them.

Desplandis:
Maybe a different version of `node` on the CI vs yours locally?

We have an open PR (https://github.com/iTowns/itowns/pull/2268) which should update the version of node on the CI (any news @mgermerie?)

jailln:
> Maybe a different version of `node` on the CI vs yours locally?
> 
> We have an open PR (#2268) which should update the version of node on the CI (any news @mgermerie?)

That was I though too but I already verified I used the right version. What's kind of weird is that one of the three tests is passing while all the test use pretty much the same logic.

jailln:
Sooo, I removed the functional tests for now :sweat_smile: I kept them [here](https://github.com/jailln/itowns/tree/3d-tiles-renderer-functional-test) if we want to finish the work one day

PR #2416, opened 2024-09-28
fix(C3DTilesLayer): updateStyle works with new style API

valentinMachado:
What is wrong:

[widgets_3dtiles_style]( https://www.itowns-project.org/itowns/examples/#widgets_3dtiles_style) is not working, after a look I realised that ````style.[category].property```` is now checking what is the type of the property when getting it.

https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Core/Style.js#L157
https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Core/Style.js#L22

In this case style.fill.color (resp. style.fill.opacity) are functions and that's the layer checking the type

https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Layer/C3DTilesLayer.js#L391

So the problem is C3DTilesLayer should not handling if fill.color is a function or something else but just calling fill.color and get what this is supposed to return.

Solution :

Since the layer is not handling the call of the style function (it's Style now) and that's the layer knowing what feature is being updated, the layer is now setting feature in parameters binded by Style and more precisely in the stylecontext parameter (which looks appropiate since style context has a feature property)

https://github.com/valentinMachado/itowns/blob/45203e1295d952c4a86bf89f6680711a9f865ca6/src/Layer/C3DTilesLayer.js#L390
https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Core/Style.js#L45

so style function can access to what feature is being updated (note that I had to create a getter of feature property to make this possible)

The example is now working with that solution but the style api change had break the C3DTilesLayer style api and this breaking change should be notify to itowns users.

I hope this post is clear and the solution proposed is appropriate

### DIFF
--- a/src/Layer/C3DTilesLayer.js
+++ b/src/Layer/C3DTilesLayer.js
@@ -143,7 +143,7 @@ class C3DTilesLayer extends GeometryLayer {
         }
 
         /** @type {Style} */
-        this._style = config.style || null;
+        this.style = config.style || null;
 
         /** @type {Map<string, THREE.MeshStandardMaterial>} */
         this.#fillColorMaterialsBuffer = new Map();
@@ -386,21 +386,16 @@ class C3DTilesLayer extends GeometryLayer {
                     if (c3DTileFeature.object3d != object3d) {
                         continue;// this feature do not belong to object3d
                     }
+
+                    this._style.context.setGeometry({
+                        properties: c3DTileFeature,
+                    });
+
                     /** @type {THREE.Color} */
-                    let color = null;
-                    if (typeof this._style.fill.color === 'function') {
-                        color = new THREE.Color(this._style.fill.color(c3DTileFeature));
-                    } else {
-                        color = new THREE.Color(this._style.fill.color);
-                    }
+                    const color = new THREE.Color(this._style.fill.color);
 
                     /** @type {number} */
-                    let opacity = null;
-                    if (typeof this._style.fill.opacity === 'function') {
-                        opacity = this._style.fill.opacity(c3DTileFeature);
-                    } else {
-                        opacity = this._style.fill.opacity;
-                    }
+                    const opacity = this._style.fill.opacity;
 
                     const materialId = color.getHexString() + opacity;
 
@@ -464,7 +459,13 @@ class C3DTilesLayer extends GeometryLayer {
     }
 
     set style(value) {
-        this._style = value;
+        if (value instanceof Style) {
+            this._style = value;
+        } else if (!value) {
+            this._style = null;
+        } else {
+            this._style = new Style(value);
+        }
         this.updateStyle();
     }
 
--- a/src/Utils/gui/C3DTilesStyle.js
+++ b/src/Utils/gui/C3DTilesStyle.js
@@ -1,4 +1,3 @@
-import Style from 'Core/Style';
 import { C3DTILES_LAYER_EVENTS } from 'Layer/C3DTilesLayer';
 import Widget from './Widget';
 
@@ -221,12 +220,12 @@ class C3DTilesStyle extends Widget {
                 }
 
                 // set style
-                c3DTilesLayer.style = new Style({
+                c3DTilesLayer.style = {
                     fill: {
                         color: fillColorFunction,
                         opacity: fillOpacityFunction,
                     },
-                });
+                };
             });
         });
 


test/unit/3dtileslayerstyle.js
```diff
@@ -75,26 +76,26 @@ describe('3DTilesLayer Style', () => {
...
+                    opacity: (styleContextProperties, styleContext) => {
+                        if (styleContext.feature.getInfo().something) {
+                            return 0.1;
+                        } else if (styleContext.feature.userData.something === 'random') {
+                            return 1;
+                        } else {
+                            return 0.5;
+                        }
+                    } },
+            });
```
Desplandis: I would prefer to keep the declarative way (i.e. an object) to declare a style, instead of an instance of `Style`. We have plans to keep the style class as internal to iTowns and not expose it anymore. Instead the style parameter shall respect the `Style` interface (or at least a subset of it in the case of `C3DTilesLayer`). You could initialize the `Style` class at `C3DTilesLayer` initialization.

src/Layer/C3DTilesLayer.js
```diff
@@ -386,21 +386,14 @@ class C3DTilesLayer extends GeometryLayer {
                     if (c3DTileFeature.object3d != object3d) {
                         continue;// this feature do not belong to object3d
                     }
+
+                    this._style.context.setFeature(c3DTileFeature);
+
```
Desplandis: The `Style` class provides a `setContext` method to pass a custom `Context`. You could then provide your custom context with a single parameter `properties` to not break the old style function declaration. You could update here this property for each `feature`.

src/Utils/gui/C3DTilesStyle.js
```diff
@@ -87,21 +87,21 @@ class C3DTilesStyle extends Widget {
                 /** @type {Map<HTMLElement, Function>} */
                 const colorFunctions = new Map();
 
-                const fillColorFunction = (c3DTileFeature) => {
+                const fillColorFunction = (styleContextProperties, styleContext) => {
```
Desplandis: See above, you could have a way to not break the API style declaration. =)

src/Core/Style.js
```diff
@@ -222,6 +222,10 @@ export class StyleContext {
         return this.#geometry.properties;
     }
 
+    get feature() {
+        return this.#feature;
+    }
+
```
Desplandis: This could not be necessary if you provide a custom context.

Desplandis (changes requested):
Hi and thanks for fixing the issue! =)

I think there should be a way to not break the old Style declaration API (see comments below). Moreover, I would like to keep the declarative way of styling a layer (see comments for the motivations).

test/unit/3dtileslayerstyle.js
valentinMachado: Ok, I've done a setter checking the type of style since ````C3DTilesLayer.style = value```` could be called in an itowns context ( ie  value is a Style instance) or by an itowns user (ie value is a literal object) if that's not the case and I'll remove the check and just treat the literal object. I also remove Style import from the widget 3dstyle since I've consider that widget code should be seen as itowns user code.

src/Layer/C3DTilesLayer.js
valentinMachado: I did not set a custom context but just the geometry of the style context because the setter was already there.

src/Utils/gui/C3DTilesStyle.js
valentinMachado: Yes I did not get that comment actually 
https://github.com/iTowns/itowns/blob/c862ca7376c195596c1111064f8189ccaad04348/src/Core/Style.js#L42

valentinMachado:
unit test of potree2 is failing but looks like because this resources is not online anymore https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/potree2.0/lion

src/Layer/C3DTilesLayer.js
Desplandis: Okay for me, but we'll surely rework this with a custom context on the new 3D Tiles implementation. =)

Desplandis (approved):
Okay for me! I'll surely re-implement this functionality on the new 3D Tiles implementation. Could I have some of your inputs on this future PR?

valentinMachado:
> Okay for me! I'll surely re-implement this functionality on the new 3D Tiles implementation. Could I have some of your inputs on this future PR?

Nice, I've worked on the style functionality of C3DTilesLayer so if I am around here I can help with that.

PR #2420, opened 2024-10-02
Fix elevation values clamping blocking terrain subdivision

jailln:
Fix elevation values clamping: also clamp the `max` elevation value of  a tile to `options.zmin` and the `min` value to `options.zmax`. 

In some cases an elevation texture tile only has values below `options.zmin`. In this case, only `min` was clamped to `options.zmin` which could for instance result in `min` being superior to `max` (e.g. `min=0` and `max=-943566`). This was causing the bounding sphere of the node to be wrong and then the [subdivision](https://github.com/iTowns/itowns/blob/master/src/Layer/TiledGeometryLayer.js#L413) check to fail. Resulting in rendering like thtis:

image-2024-03-11-17-10-16-562

### DIFF
--- a/src/Parser/XbilParser.js
+++ b/src/Parser/XbilParser.js
@@ -64,8 +64,16 @@ export function computeMinMaxElevation(texture, pitch, options) {
                 }
             }
         }
-        if (options.zmin > min) { min = options.zmin; }
-        if (options.zmax < max) { max = options.zmax; }
+        // Clamp values to zmin and zmax values configured in ElevationLayer
+        if (options.zmin != null) {
+            if (min < options.zmin) { min = options.zmin; }
+            if (max < options.zmin) { max = options.zmin; }
+        }
+
+        if (options.zmax != null) {
+            if (min > options.zmax) { min = options.zmax; }
+            if (max > options.zmax) { max = options.zmax; }
+        }
     }
 
     if (max === -Infinity || min === Infinity) {


src/Parser/XbilParser.js
```diff
@@ -64,8 +64,16 @@ export function computeMinMaxElevation(texture, pitch, options) {
                 }
             }
         }
-        if (options.zmin > min) { min = options.zmin; }
-        if (options.zmax < max) { max = options.zmax; }
+        // Clamp values to zmin and zmax values configured in ElevationLayer
+        if (options.zmin != null) {
```
ftoromanoff, line 68: Why null and not undefined ?

src/Parser/XbilParser.js
```diff
@@ -64,8 +64,16 @@ export function computeMinMaxElevation(texture, pitch, options) {
...
+            if (min < options.zmin) { min = options.zmin; }
+            if (max < options.zmin) { max = options.zmin; }
+        }
+
+        if (options.zmax != null) {
+            if (min > options.zmax) { min = options.zmax; }
+            if (max > options.zmax) { max = options.zmax; }
+        }
```
ftoromanoff, line 76: Can be replaced by
```
const zmin = options.zmin ?? min;
const zmax = options.zmax ?? max;

min = Math.max(zmin, Math.min(min, zmax));
max = Math.min(zmax, Math.max(max, zmin));
```
(No idea if it's quicker)

jailln: I thought about it but I found it harder to read and understand, wdyt?

src/Parser/XbilParser.js
jailln: `!= null` is the same as checking for `null` or `undefined`

ftoromanoff: Ok

src/Parser/XbilParser.js
ftoromanoff: In my opinion I find it better as it suppreses the need to check for null and 6 ifs. But I also not reallypartial as it is my proposal ^^
It can also if needed, keep the initial value of min and max and use zmin/zmax for the following traitement (with a minor change).

Let see what the others prefer, as soon as it does the job ;-)

jailln: I've tested it and it does not work:
options.zmin = 0
options.zmax = undefined = max = -98509.849609375
max = -98509.849609375
Math.max(max, zmin) = 0
max = Math.min(zmax, Math.max(max, zmin)); = -98509.849609375 (instead of 0)

For it to work this way we may need to apply another Math.min and Math.max to zmax and zmin definition at the beginning. 

Shall we go with the good old ifs ? (easier to read and to debug in this case in my opinion)

ftoromanoff: If it works, and you find it more readable, go for it.

ftoromanoff (approved):

PR #2431, opened 2024-10-10
Make globe sun light optional

jailln:
## Description
* Add a configuration option to the `GlobeView` to make sun lightning simulation optional
* Update 3D tiles loader example lightning

## Motivation and Context
This light can cause weird lightning like the folllowing:
 
sunlight-opt

And shadows are not casted on the terrain anyway

### DIFF
--- a/examples/3dtiles_loader.html
+++ b/examples/3dtiles_loader.html
@@ -100,7 +100,7 @@
             itowns.enableKtx2Loader('./lib/basis/', view.renderer);
 
             // Add ambient light to globally illuminates all objects
-            const light = new AmbientLight(0x404040, 15);
+            const light = new AmbientLight(0x404040, 40);
             view.scene.add(light);
 
             // Setup loading screen
--- a/src/Core/Prefab/GlobeView.js
+++ b/src/Core/Prefab/GlobeView.js
@@ -96,11 +96,6 @@ class GlobeView extends View {
         const tileLayer = new GlobeLayer('globe', options.object3d, options);
         this.mainLoop.gfxEngine.label2dRenderer.infoTileLayer = tileLayer.info;
 
-        const sun = new THREE.DirectionalLight();
-        sun.position.set(-0.5, 0, 1);
-        sun.updateMatrixWorld(true);
-        this.scene.add(sun);
-
         this.addLayer(tileLayer);
         this.tileLayer = tileLayer;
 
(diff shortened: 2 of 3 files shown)


examples/3dtiles_loader.html
```diff
@@ -90,7 +90,8 @@
             const view = new GlobeView(viewerDiv, placement, {
                 controls: {
                     minDistance: 100,
-                }
+                },
+		activateSunlight: false,
             });
```
Desplandis: Tabulations instead of spaces

```suggestion
                },
                activateSunlight: false,
            });
```

src/Core/Prefab/GlobeView.js
```diff
@@ -83,6 +83,7 @@ class GlobeView extends View {
      * @param {CameraTransformOptions|Extent} placement - An object to place view
      * @param {object} [options] - See options of {@link View}.
      * @param {Object} [options.controls] - See options of {@link GlobeControls}
+     * @param {Boolean} [options.activateSunlight=true] - whether to add a THREE.DirectionalLight simulating sun lightning
```
Desplandis: I know this is a breaking change but was there a reason to put a `DirectionalLight` in an arbitrary position?
I'm also for renaming `activateSunlight` to `sunlight` or pass a `light` parameter of type `Light` and defaulting to the current sun light.

Desplandis (dismissed):
Agree with the need to disable the default `DirectionalLight`. IMO we could even remove this default light, but let's not break things. x)

jailln:
> Agree with the need to disable the default `DirectionalLight`. IMO we could even remove this default light, but let's not break things. x)

I would also prefer removing it since it does not really simulate the sun. Since we both agree, shall we ? :grinning: I'm not sure anyone has any use of this light anyway.

Desplandis:
Yeah, let's remove it. We'll maybe reintroduce a kinda similar object (using `CoordStars`) later on but this should not be by default in the view (same for the Atmosphere IMO).

jailln:
@Desplandis done

jailln:
@HoloTheDrunk could you take a look please as @Desplandis is off ?  I removed the light as discussed in the PR.

examples/3dtiles_loader.html
```diff
@@ -100,7 +100,7 @@
             itowns.enableKtx2Loader('./lib/basis/', view.renderer);
 
             // Add ambient light to globally illuminates all objects
-            const light = new AmbientLight(0x404040, 15);
+            const light = new AmbientLight(0x404040, 40);
```
HoloTheDrunk, line 103: Maybe remove it entirely?

HoloTheDrunk: Oops, definitely misread that.

HoloTheDrunk (approved):

PR #2432, opened 2024-10-10
Fix wfs wms wmts url building when url already contains query params

tebben:
## Description

Fix building url for WFSSource, WMSSource and WMTSSource when the input url already contains query parameters.

## Motivation and Context

When an url for source WFS, WMS or WMTS contains a `?` the code currently adds another `?` resulting in a bad request. The current code only leaves out the question mark when the url ends with `?` but does not actualy handle urls already containing query parameters.

A more elegant way would be to use `new URL()` that handles query parameters automatically but I didn't want to change the code too much. Example:

```js
const url = new URL(this.url);
url.searchParams.append('LAYER', source.name);
url.searchParams.append('FORMAT', this.format);
url.searchParams.append('SERVICE', 'WMTS');
....

this.url = url.toString();
```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -49,6 +49,7 @@ The following people have contributed to iTowns.
 * [Sogelink](https://www.sogelink.com/)
   * [Kévin ETOURNEAU](https://github.com/ketourneau)
   * [Alexis DELFORGES](https://github.com/pourfex)
+  * [Tim Ebben](https://github.com/tebben)
 
 The following organizations are the current maintainers of iTowns:
 * IGN (http://www.ign.fr)
--- a/src/Source/WFSSource.js
+++ b/src/Source/WFSSource.js
@@ -125,26 +125,25 @@ class WFSSource extends Source {
         this.typeName = source.typeName;
         this.version = source.version || '2.0.2';
         this.bboxDigits = source.bboxDigits;
-
-        // Add ? at the end of the url if it is not already in the given URL
-        if (!this.url.endsWith('?')) {
-            this.url = `${this.url}?`;
-        }
-        this.url = `${source.url
-        }SERVICE=WFS&REQUEST=GetFeature&typeName=${this.typeName
-        }&VERSION=${this.version
-        }&SRSNAME=${this.crs
-        }&outputFormat=${this.format
-        }&BBOX=%bbox,${this.crs}`;
-
         this.zoom = { min: 0, max: Infinity };
+
+        const urlObj = new URL(source.url);
+        urlObj.searchParams.set('SERVICE', 'WFS');
+        urlObj.searchParams.set('REQUEST', 'GetFeature');
+        urlObj.searchParams.set('typeName', this.typeName);
+        urlObj.searchParams.set('VERSION', this.version);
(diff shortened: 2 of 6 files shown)


jailln:
Hi, thanks for your contribution :) 

Do you have a specific use case for which you need to provide an url already containing parameters or is it just to improve robustness ?

tebben:
Yes, I want to add a WMTS layer which only works by providing a servicekey param `https://wmts-server/something?servicekey=abc`

jailln:
Ok thanks for the precision.

I agree with you that it would be better to use `new URL()` instead, would you be willing to do it?
If not, I can review and we can merge the PR with the current way of doing it since your proposal still fixes a bug.

tebben:
Updated to use URL and URLSearchParams API for robustness, better readability and maintainability.

There are dynamic parameters in `this.url` like `%bbox`, `%TILEMATRIX`, ... which are going to be replaced with an actual value in URLBuilder. URL automatically encodes the url changing things like `%` into `%25` which results in URLBuilder not able to replace values. To handle this the url is set using `decodeURIComponent`.

The unit tests contained invalid urls `http://` and therefore fail, these urls are updated to something valid: `http://domain.com`

Manually tested the following sources and urls.

WMS:
https://service.pdok.nl/rws/ahn/wms/v1_0
https://service.pdok.nl/rws/ahn/wms/v1_0?
https://service.pdok.nl/rws/ahn/wms/v1_0?request=GetCapabilities&service=WMS

WMTS:
https://service.pdok.nl/brt/achtergrondkaart/wmts/v2_0
https://service.pdok.nl/brt/achtergrondkaart/wmts/v2_0?
https://service.pdok.nl/brt/achtergrondkaart/wmts/v2_0?foo=bar
https://services.geodan.nl/data/geodan/gws/world/streets/wmts?servicekey=redacted

WFS
https://data.geopf.fr/wfs/ows
https://data.geopf.fr/wfs/ows?
https://data.geopf.fr/wfs/ows?foo=bar

jailln:
Thanks for the modifications, it's clearer and more robust this way.

One last thing, we follow [angular commit conventions](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md) - see [here](https://github.com/iTowns/itowns/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details on the commit messages we support. Also, our release changelog is automatically generated  from the commits between two releases. Could you update the commits of this branch to keep only 2 so the release changelog is clearer please ? : one for adding your name to the contributors (e.g. doc(contributors): add Tim Ebben) and one for the url params handling. Thanks :)

jailln (approved):
Thanks for your contribution and for bearing with me for the requested changes :)

jailln:
Oh it seems that having multiple scopes (wms, wmts, wfs) doesn't pass our commit message checking script run by the ci (maybe something to improve). I took the liberty to update your last commit message to make it pass the commit message check.

tebben:
No problem asking for changes, I should have read the contribution guidelines better. I used this as example for the commit message which had multiple scopes: https://github.com/iTowns/itowns/commit/f44dfb260275858d1f84337eea57aff872ada107

jailln:
The commit message checking script must have been wrongly changed since

Issue #2426, opened 2024-10-04
3Dtiles issue when 2 view are created


AnthonyGlt:
I have 2 itowns view loaded on the same page.
On each view, I'm loading a 3dtiles dataset.
I have to wait for the 3dtiles from the first view to finish loading before seeing the tiles from the second view.
Also, when I switch from one view to the other, the 3dtiles disappear and starts requesting all the tiles again.

## Context
Manage 2 view with 3dtiles in each

## Steps to Reproduce
1. Create 2 views with 2 3dtile dataset
2. Interact with each view
3. Observe that the whole dataset disappear before reloading completely

[Screencast from 04-10-2024 15:11:49.webm](https://github.com/user-attachments/assets/e4c6997b-d8bc-4185-8f2a-8193d90af88b)

## Possible Cause/Fix/Solution
Looks like it's caused by:
https://github.com/iTowns/itowns/blob/c862ca7376c195596c1111064f8189ccaad04348/src/Layer/OGC3DTilesLayer.js#L197

jailln:
We indeed share the 3D tiles cache, download and parse queues among all tilesets. 

With this new use case in mind, one solution would be to share them **per view**. To do so, we could:
* Add an `id` property to `View`
* Declare a `Map` of 3DTilesRendererJS with the view id as key and `lruCache`, `downloadQueue` and `parseQueue` as value. 
* Take care of filling this map and setting up the cache and queues in `OGC3DTilesLayer._setup` (that is called by `View.addLayer(3DTilesLayer)`)

That would have the advantages to manage 3DTiles independently for each view and to take care of this setup for our users (which in many cases won't want to deal with this kind of stuff I think). I also can't think of any usecase where one would want to share the same cache and queues amongst views.

WDYT @AnthonyGlt @Desplandis ?

jailln:
Done in https://github.com/iTowns/itowns/pull/2435

PR #2435, opened 2024-10-11
fix(OGC3DTilesLayer): handle multiple views

AnthonyGlt:
## Description
Handle the 3dtiles cache in each view when multiple views are created

## Motivation and Context
https://github.com/iTowns/itowns/issues/2426

## Modifications
Adding an id to the view. Incremented at each new view.
Adding a list of cache and queues in OGC3DTilesLayer.js, to share them if in in the same view

## Improvements
@jailln @Desplandis 

How to handle a View disposing ?
Also, a "viewers" constant already exist in View.js, maybe this one could be used ?

https://github.com/iTowns/itowns/blob/a4f0d2284fabbe4a5ecf0cb512c9c35d6b684ee3/src/Core/View.js#L118

### DIFF
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -32,6 +32,7 @@ export const VIEW_EVENTS = {
     INITIALIZED: 'initialized',
     COLOR_LAYERS_ORDER_CHANGED,
     CAMERA_MOVED: 'camera-moved',
+    REMOVED: 'removed',
 };
 
 /**
@@ -297,6 +298,7 @@ class View extends THREE.EventDispatcher {
         }
 
         window.removeEventListener('resize', this._resizeListener);
+        this.dispatchEvent({ type: VIEW_EVENTS.REMOVED });
 
         // controls dispose
         if (this.controls) {
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -122,7 +122,8 @@ let screenMeters;
 let id = 0;
 
 /**
- * @property {HTMLElement} domElement - Thhe domElement holding the canvas where the view is displayed
+ * @property {number} id - The id of the view. It's incremented at each new view instance, starting at 0.
+ * @property {HTMLElement} domElement - The domElement holding the canvas where the view is displayed
  * @property {String} referenceCrs - The coordinate reference system of the view
  * @property {MainLoop} mainLoop - itowns mainloop scheduling the operations
  * @property {THREE.Scene} scene - threejs scene of the view
@@ -147,10 +148,10 @@ class View extends THREE.EventDispatcher {
      * var view = itowns.View('EPSG:4326', viewerDiv, { camera: { type: itowns.CAMERA_TYPE.ORTHOGRAPHIC } });
      * var customControls = itowns.THREE.OrbitControls(view.camera3D, viewerDiv);
      *
-     * @param {string} crs - The default CRS of Three.js coordinates. Should be a cartesian CRS.
+     * @param {String} crs - The default CRS of Three.js coordinates. Should be a cartesian CRS.
      * @param {HTMLElement} viewerDiv - Where to instanciate the Three.js scene in the DOM
      * @param {Object} [options] - Optional properties.
-     * @param {object} [options.camera] - Options for the camera associated to the view. See {@link Camera} options.
+     * @param {Object} [options.camera] - Options for the camera associated to the view. See {@link Camera} options.
      * @param {MainLoop} [options.mainLoop] - {@link MainLoop} instance to use, otherwise a default one will be constructed
      * @param {WebGLRenderer|Object} [options.renderer] - {@link WebGLRenderer} instance to use, otherwise
      * a default one will be constructed. In this case, if options.renderer is an object, it will be used to
--- a/src/Core/View.js
+++ b/src/Core/View.js
@@ -119,6 +119,8 @@ const viewers = [];
 // Size of the camera frustrum, in meters
 let screenMeters;
 
+let id = 0;
+
 /**
  * @property {HTMLElement} domElement - Thhe domElement holding the canvas where the view is displayed
  * @property {String} referenceCrs - The coordinate reference system of the view
@@ -169,6 +171,7 @@ class View extends THREE.EventDispatcher {
         super();
 
         this.domElement = viewerDiv;
+        this.id = id++;
 
         this.referenceCrs = crs;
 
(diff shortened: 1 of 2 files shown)


src/Core/View.js
```diff
@@ -169,6 +171,7 @@ class View extends THREE.EventDispatcher {
         super();
 
         this.domElement = viewerDiv;
+        this.id = id++;
```
jailln, line 176: I think it should be exposed in the documentation, WDYT @AnthonyGlt @Desplandis ?

src/Layer/OGC3DTilesLayer.js
```diff
@@ -22,19 +22,14 @@ import PointsMaterial, {
 } from 'Renderer/PointsMaterial';
 
 const _raycaster = new THREE.Raycaster();
-
+const viewers = [];
```
jailln: Can you add a comment saying that it stores ids of viewers where tilesets have been added and explaining why and how it is related to the cache please ? You can also link the issue I think

src/Layer/OGC3DTilesLayer.js
```diff
@@ -190,25 +182,23 @@ class OGC3DTilesLayer extends GeometryLayer {
         this.pntsMaxAttenuatedSize = config.pntsMaxAttenuatedSize || 7;
     }
 
+
     /**
      * Sets the lruCache and download and parse queues so they are shared amongst all tilesets.
```
jailln: Can you update this comment please ? It's shared per view now

jailln:
@AnthonyGlt when a view is disposed, all layers added to this view are removed and their object3Ds are removed from the scene so it should be handled automatically; but it's worth checking that everything works fine :)

AnthonyGlt:
> @AnthonyGlt when a view is disposed, all layers added to this view are removed and their object3Ds are removed from the scene so it should be handled automatically; but it's worth checking that everything works fine :)

I meant that in the viewers of OGC3DTilesLayer here

https://github.com/iTowns/itowns/pull/2435/files#diff-05568d5784ac6e47fd8c2b288edaf3965abce4a92c99c0b8bec7de4ec68a5f77R29

We will store cache & queues for each view.

Once a view is disposed, the corresponding cache & queue will still be stored in the `viewers`, so we never remove anything from `viewers`

@jailln

jailln:
If you want to remove it from viewers I think you should do it [here](https://github.com/iTowns/itowns/pull/2435/files#diff-05568d5784ac6e47fd8c2b288edaf3965abce4a92c99c0b8bec7de4ec68a5f77L320) (called when a view holding this layer is disposed)

AnthonyGlt:
> If you want to remove it from viewers I think you should do it [here](https://github.com/iTowns/itowns/pull/2435/files#diff-05568d5784ac6e47fd8c2b288edaf3965abce4a92c99c0b8bec7de4ec68a5f77L320) (called when a view holding this layer is disposed)

@jailln it should be disposed when the view is disposed, not the layer, because we share it between all layers of a same view. 
It could be linked to the disposal of the view, right ?

https://github.com/iTowns/itowns/blob/d73f0a185c0b41dd791357ed48a540440b92ed28/src/Core/View.js#L283-L292

jailln:
> @jailln it should be disposed when the view is disposed, not the layer, because we share it between all layers of a same view. It could be linked to the disposal of the view, right ?

Yes you're right it should be disposed when the view is disposed. Normally, when a layer is disposed, the tiles of this layer (only) should be removed from the cache (this should be handled by `3DTilesRenderer` with [tilesRenderer.dispose()](https://github.com/iTowns/itowns/pull/2435/files#diff-05568d5784ac6e47fd8c2b288edaf3965abce4a92c99c0b8bec7de4ec68a5f77L320)). When a view is disposed, all layers from this view are disposed, so the lruCache cache corresponding to this view should be emptied (also because [tilesRenderer.dispose()](https://github.com/iTowns/itowns/pull/2435/files#diff-05568d5784ac6e47fd8c2b288edaf3965abce4a92c99c0b8bec7de4ec68a5f77L320)) will be called for each layer). Therefore, in my opinion nothing more should be done for emptying the cache. WDYT? It's worth checking this out in practice though :) Can you validate it's working correctly?

What you may want to do in addition is to remove the view and its associated `lruCache`, `downloadQueue` and `parseQueue` from the `viewers` when a view is disposed? You can setup an event when a view is disposed to do that for instance?

AnthonyGlt:
@jailln 
>What you may want to do in addition is to remove the view and its associated lruCache, downloadQueue and parseQueue from the viewers when a view is disposed? You can setup an event when a view is disposed to do that for instance?

That was what I was talking about :)

I've pushed some edits, maybe a bit overkill to pass the view as parameter but it works.
WDYT ?

src/Core/View.js
```diff
@@ -297,6 +298,7 @@ class View extends THREE.EventDispatcher {
         }
 
         window.removeEventListener('resize', this._resizeListener);
+        this.dispatchEvent({ type: VIEW_EVENTS.REMOVED });
```
jailln: Can you dispatch it at the end of this method instead please?

src/Core/View.js
```diff
@@ -32,6 +32,7 @@ export const VIEW_EVENTS = {
     INITIALIZED: 'initialized',
     COLOR_LAYERS_ORDER_CHANGED,
     CAMERA_MOVED: 'camera-moved',
+    REMOVED: 'removed',
```
jailln: Can you rename to `DISPOSED` please ? It's more commonly used in threejs and it will be the same name than the method

jailln (commented):
thanks :)

src/Core/View.js
AnthonyGlt: I wanted to but we have an event cleaner, so we need to do it before it

```js
 // remove alls events
this.removeAllEvents();

```

jailln: Oh ok, I missed that. This is not ideal since if a user registers to this event, he would expect the view and everything related to the view to be disposed which won't be the case. What about moving `this.removeAllEvents()` at the end of the method too ?

jailln (approved):

jailln:
Ok for me, thanks for the changes :) @Desplandis @mgermerie do you want to take a look at it or should we merge?

Desplandis:
Okay for me, but we'll need to have a better way to handle shared ressources between layers of a given view!

Issue #2396, opened 2024-09-11
[Proposal] Migration from JavaScript to TypeScript


Desplandis:
This issue is a **feature proposal**. Feel free to upvote (with :+1: ), comment and provide your use-cases if you're interested by this feature.

## Context
The `iTowns` codebase is currently written in `Javascript`, with `JSDoc` comments providing a loose form of type-checking that isn't enforced (but in practice used by the LSP of our code editors). This lack of type safety has led to side-effects creeping into the project, making maintenance and debugging more challenging and causing unpredictable behavior in certain areas of the code.

This proposal outlines a migration path to `TypeScript`, which will enforce type safety, simplify the refactoring of complex or difficult-to-maintain parts of the code, and enable developers to take full advantage of modern tooling and best practices.

## Description of the proposal
This proposal details the steps for migrating `iTowns` from `JavaScript` to `TypeScript`. The migration will follow a systematic, bottom-to-top approach, beginning with simpler, independent modules before tackling more complex and interdependent ones.

Along the way, we will take the opportunity to refactor key components to improve clarity and maintainability. These refactorings will be addressed through separate proposals and pull requests.

This issue will primarily serve as a tracker for the migration process and a space to discuss each phase of the migration (see tracking section below).

### Tooling
The migration will necessitate some changes on our tooling:
- Modify the `ESLint` configuration to support `TypeScript`, integrating the `@typescript-eslint` plugin. We should ensure that both `JavaScript` and `TypeScript` code follow a unified set of rules, with `TypeScript` as a superset of `JavaScript`.
- Modifiy both `babel` and `webpack` config to support `TypeScript` as input (this should be relatively easy).
- Transition from `JSDoc` to `Typedoc` for documentation, as `JSDoc` does not support `TypeScript`, and `Typedoc` will provide stricter type-driven documentation.

### Code migration
The migration process will be gradual, but the following steps could serve as a good start:
- **Migrate geographic modules**: migrate geographic modules first, since they have been identified as simple and easy to convert, and could provide a good first case to ensure that our tooling is correct.
- **Refactor and split the `Extent` module**: split `Extent` into two sub-modules—one for geographic extents and one for tiled extents. This refactor is long due, as these modules currently contribute to at least 25% of all type issues and need a cleaner structure for proper typing.
- **Refactor Tile Geometry Creation**: refactor and migrate the tile geometry creation mechanism, which has accumulated several layers of patches and requires simplification to improve maintainability.
[...]

Issue #2414, opened 2024-09-25
Monorepo


gchoqueux:
Adopt a monorepo structure for itowns. The goal is to split itowns down into **reusable  and standalone feature packages**.
As an example, I've started with the geographic features. (@itowns/geographic)

## Context
For the moment, itowns provides an api in a single package.
this architecture poses several problems :
- Users must install the full api even if they want to use a specific feature. 
- The majority of features aren't independent, as users are obliged to use a `Viewer` instance.
- It's more difficult for developers to add feature, re-use, test and dependency management.

## Description of the proposal
Split code in packages for clearly structured code.
Simplifies development and facilitates contributions.
Increases the scope of users who only want to use a few features.
This structure makes it necessary to make features independent and improves the architecture

For the moment only one new package is proposed: Geographic

## Identified use-cases
Users can use only the packages they are interested in.
Three.js developers can use itowns features  more easily.
Developers more easily merge their codes  (I received feedsbacks for this architecture)

## Implementation
see #2300 

## Comments
This monorepo structure supports Yarn
There is not breaking change for itowns users

## Potential Problems
Rebasing but generally Git does the job, Except when there are bumps and new files or move files
The new publishing, only tested on my npm account

# iTowns Monorepo Structure Proposal

## 1. `@itowns/core` : :warning: 
- **Description**: Manages processing, scheduling, rendering and 3D scene management

## 2. `@itowns/geographic` :warning:
- **Description**: Contains the Coordinates, Extent, Crs, Ellipsoid and OrientationUtils

## 4. `@itowns/sources`
- **Description**: Handles data sources
- **Sub-packages**:
  - `@itowns/sources/wms`
  - `@itowns/sources/wmts`
  - `@itowns/sources/3dtiles`
  - `@itowns/sources/gpx`
  - `@itowns/sources/kml`

## 4. `@itowns/layers`
- **Description**: Manages different types of layers and Style
- **Sub-packages**:
  - `@itowns/layers/raster`
	- `@itowns/layers/colorLayer`
	- `@itowns/layers/colorElevation`
  - `@itowns/layers/vector`
  	- `@itowns/sources/3dtiles`
  	- `@itowns/sources/featureGeometryLayer`
  - `@itowns/layers/terrain`

## 5. `@itowns/mouseControls`
- **Description**: Manages user mouse controls for navigating the 3D scene (panning, zooming, rotating, etc.).
- **Sub-packages**:
  - `@itowns/controls/Globe`
[...]

PR #2436, opened 2024-10-11
refacto: migrate Crs to typescript

Desplandis:
## Description

This PR includes:
- Migration of `Crs` from javascript to typescript
- Fix of type issues (mainly undefined case not treated)
- Add a `ProjectionLike` type alias for named CRS which could be later extended as either a string or a `proj4` definition object (à la [OpenLayer](https://openlayers.org/en/latest/apidoc/module-ol_proj.html#~ProjectionLike))
- TSDoc compliant documentation
- Add more unit tests

## Motivation and Context

As described in proposal #2396, we aim to gradually migrate our entire codebase (with the exception of deprecated modules) from javascript to typescript. We choose to start with modules with no-dependency and move up in the dependency tree.

This PR is the first step of **Migrate geographic modules**, migrating the base module `Crs`. Follow-up steps includes migration of dependent modules `Coordinates/Ellipsoid` and `Extent`.


src/Core/Geographic/Crs.ts
```diff
@@ -80,100 +85,92 @@ function toUnitWithError(crs) {
...
 
     /**
-     * Assert that the CRS is valid one.
+     * Assert that the CRS is a valid one.
      *
-     * @param {string} crs - The CRS to validate.
+     * @param crs - The CRS to validate.
      *
-     * @throws {Error} if the CRS is not valid.
+     * @throws {@link Error} if the CRS is not valid.
```
mgermerie: Are you sure about the `@link` ? Will it link to [this doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) should `throws` be shown on doc template ?

src/Core/Geographic/Crs.ts
```diff
@@ -80,100 +85,92 @@ function toUnitWithError(crs) {
...
      */
-    isValid(crs) {
+    isValid(crs: ProjectionLike) {
         toUnitWithError(crs);
```
mgermerie: This throws `No unit found for crs: '${crs}'` when this function should instead tell the CRS is not valid.

src/Core/Geographic/Crs.ts
```diff
@@ -80,100 +85,92 @@ function toUnitWithError(crs) {
...
      *
-     * @param {string} crs - The CRS to validate.
-     * @return {boolean}
-     * @throws {Error} if the CRS is not valid.
+     * @param crs - The CRS to validate.
+     * @throws {@link Error} if the CRS is not valid.
      */
-    isGeographic(crs) {
+    isGeographic(crs: ProjectionLike) {
         return (toUnitWithError(crs) == UNIT.DEGREE);
```
mgermerie: Same here, the message thrown by `toUnitWithError` is not relevant.

src/Core/Geographic/Crs.ts
```diff
@@ -80,100 +85,92 @@ function toUnitWithError(crs) {
...
     /**
-     * Is the CRS geocentric ?
+     * Is the CRS geocentric?
      * if crs isn't defined the method returns false.
      *
-     * @param {string} crs - The CRS to test.
-     * @return {boolean}
+     * @param crs - The CRS to test.
      */
     isGeocentric,
```
mgermerie: Not related to pure TS migration but on a code cleanliness consideration, we could define this the same way we export `isGeographic` or `reasonnableEpsilon` : we could either define them all outside of the `export DEFAULT` object and reference them here, or define them all directly in the `export DEFAULT` object.

test/unit/crs.js
```diff
@@ -34,13 +34,21 @@ describe('CRS assertions', function () {
...
     it('should check if the CRS is EPSG:4326', function () {
         assert.ok(CRS.is4326('EPSG:4326'));
         assert.ok(!CRS.is4326('EPSG:3857'));
     });
 
+    it('should assert that the CRS is geocentric', function () {
+        assert.ok(!CRS.isGeocentric('EPSG:4326'));
+        assert.ok(!CRS.isGeocentric('EPSG:7133'));
+        assert.ok(CRS.isGeocentric('EPSG:4978'));
+        assert.ok(!CRS.isGeocentric('EPSG:3857'));
```
mgermerie, line 49: For unit tests, is it really necessary to test three non geocentric CRS ?

mgermerie: (And the same goes for all other tests, there is no benefits in testing the same result multiple times)

src/Core/Geographic/Crs.ts
```diff
@@ -55,22 +60,22 @@ function _unitFromProj4Unit(projunit) {
     }
 }
 
-function toUnit(crs) {
+function toUnit(crs: ProjectionLike) {
     mustBeString(crs);
     switch (crs) {
         case 'EPSG:4326' : return UNIT.DEGREE;
         case 'EPSG:4978' : return UNIT.METER;
```
mgermerie: We could delete these lines as the EPSG are defined (either in proj4 or in this file) and we could directly call the generic method as in `default` case to not create a specificity.

src/Core/Geographic/Crs.ts
```diff
@@ -55,22 +60,22 @@ function _unitFromProj4Unit(projunit) {
...
+            if (!p?.units) {
                 return undefined;
             }
             return _unitFromProj4Unit(p.units);
         }
     }
 }
 
-function toUnitWithError(crs) {
+function toUnitWithError(crs: ProjectionLike) {
```
mgermerie: Maybe we could rename it to a more explicit name, like `checkUnit` ? I admit I don't understand the basic role of the method without reading its code.

src/Core/Geographic/Crs.ts
Desplandis: In `tsdoc` the [`@throws`](https://tsdoc.org/pages/tags/throws/) tag is purely informational so only contain text. It seems this usually start with an `{@link X}` block. We could use [`typedoc-plugin-mdn-links`](https://github.com/Gerrit0/typedoc-plugin-mdn-links`) to resolve those links.

src/Core/Geographic/Crs.ts
Desplandis: Agree, we should maybe add a function that throws if a call to `proj4.defs` returns `undefined`.

src/Core/Geographic/Crs.ts
Desplandis: Better, in my opinion, we should just not use `export default` anymore in the codebase. There are not compatible with tree-shaking. Imagine we build a bundle importing `Crs.js` and just using the `isGeocentric`, with `export default` the whole code of `Crs.js` is bundled, with multiple `export` only this function is bundled.

Desplandis: However, this may be outside the scope of this PR?

test/unit/crs.js
Desplandis: Don't know, I just tried to have a test symmetric to the `isGeometric` one. I think we should test all four since two here are defined by the test (i.e. the user), one defined by `iTowns` in `Crs` and one by default by `proj4`.

src/Core/Geographic/Crs.ts
Desplandis: Yeah agree.

src/Core/Geographic/Crs.ts
Desplandis: The method instead of returning `unit | undefined` returns `unit` and throws if `undefined`. In my opinion, we should add a method to check whether a `proj4.defs` is defined.

package.json
```diff
@@ -21,13 +21,13 @@
...
     "build-dev": "cross-env NODE_ENV=development webpack",
-    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib",
+    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib --extensions .js,.ts",
     "start": "cross-env NODE_ENV=development webpack serve",
     "start-https": "cross-env NODE_ENV=development webpack serve --https",
     "debug": "cross-env noInline=true npm start",
     "prepublishOnly": "npm run build && npm run transpile",
     "prepare": "cross-env NO_UPDATE_NOTIFIER=true node ./config/prepare.mjs && node ./config/replace.config.mjs",
-    "watch": "cross-env BABEL_DISABLE_CACHE=1 babel --watch src --out-dir lib",
+    "watch": "npm run transpile -- --watch",
```
jailln, line 30: one `--` too many?

src/Core/Geographic/Crs.ts
```diff
@@ -0,0 +1,185 @@
+import proj4 from 'proj4';
+
+import type { ProjectionDefinition } from 'proj4';
+
+proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
```
jailln, line 5: Note that it is also defined in `Coordinates`. Can we remove it in `Coordinates`?

src/Core/Geographic/Crs.ts
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * A projection as a CRS identifier string.
+ */
+export type ProjectionLike = string;
```
jailln, line 12: Should we export it in the doc ? Also, the comment is not complete as it can either be a crs identifier or a wkt string describing the projection and passed directly to proj4 (in isGeocentric for instance)

src/Core/Geographic/Crs.ts
```diff
@@ -0,0 +1,185 @@
...
+    return crs === 'EPSG:4326';
+}
+
+function isGeocentric(crs: ProjectionLike) {
+    mustBeString(crs);
+    const projection = proj4.defs(crs);
+    return !projection ? false : projection.projName == 'geocent';
+}
+
+function _unitFromProj4Unit(proj: ProjectionDefinition) {
```
jailln: I think we should remove the `_` since it is not exported anyway

src/Core/Geographic/Crs.ts
```diff
@@ -0,0 +1,185 @@
...
+    return _unitFromProj4Unit(p);
+}
+
+/**
+ * This module provides basic methods to manipulate a CRS.
+ */
+export default {
+    /**
+     * Units that can be used for a CRS.
+     */
```
jailln: Could you add possible values in the documentation please?

src/Core/Geographic/Crs.ts
```diff
@@ -0,0 +1,185 @@
...
+        return (toUnit(crs) == UNIT.METER);
+    },
+
+    /**
+     * Get the unit to use with the CRS.
+     *
+     * @param crs - The CRS to get the unit from.
+     * @returns Either `UNIT.METER`, `UNIT.DEGREE` or `undefined`.
+     */
+    toUnit,
```
jailln: I'd rather have all the exported method defined here for readability, wdyt?

src/Core/Geographic/Crs.ts
jailln: I would be down for doing it in this PR and in general taking the opportunity of this ts migration to remove default exports.

src/Core/Geographic/Crs.ts
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Is the CRS geocentric?
+     * if crs isn't defined the method returns false.
+     *
+     * @param crs - The CRS to test.
+     */
+    isGeocentric,
+
+    /**
+     * Give a reasonnable epsilon to use with this CRS.
```
jailln: reasonable with one n

jailln (commented):
Thanks :) Nice to see that we are moving forward on this migration :+1: 

While we're at it, I would be down for renaming some functions to make their name clearer:
* `toUnit` -> `getUnit`
* `reasonnableEpsilon` -> `reasonableEpsilon`
* `formatToEPSG` -> `formatToEpsg` (to be consistent with capital letters in other methods). Or use capitals in all methods with EPSG and TMS.

I would also make private `is4326`.

Also, we could add a `isDegreeUnit` since we expose a `isMetricUnit`.

package.json
Desplandis: The added `--` is for avoiding to pass the `--watch` option to the `npm` command and instead pass it to the `babel` command in the `transpile` script.

src/Core/Geographic/Crs.ts
Desplandis: I thought I removed it in Coordinates... Will do it!

P.S.: It was done in the subsequent PR https://github.com/iTowns/itowns/pull/2437

src/Core/Geographic/Crs.ts
Desplandis: Yeah, we should export it in the documentation since this is the type of the argument of some functions of the `CRS` module.

I think the documentation is complete, this is an identifier referencing an already defined projection (I should maybe add this to the documentation). In `isGeocentric`, the call to `proj4.defs` is only for retrieving an already defined projection as an object of type `proj4.ProjectionDefinition`.

Desplandis: Moreover, I added such type alias since we'll surely want in a near future to have a `ProjectionLike` either as an identifier or an `proj4.ProjectionDefinition`.

src/Core/Geographic/Crs.ts
Desplandis: Removed.

src/Core/Geographic/Crs.ts
Desplandis: Done.

src/Core/Geographic/Crs.ts
Desplandis: Agree. Fixed it.

src/Core/Geographic/Crs.ts
Desplandis: I named-exported all functions instead of using the default export.

Desplandis:
@jailln I fixed all issues listed in your review.

I added an [`@internal`](https://tsdoc.org/pages/tags/internal/) tag for all functions that I deemed to be only used internally by itowns. This mainly includes all related epsg and tms functions, that may be removed soon since we split Extent and Tile. I only kept the TMS naming scheme for Tile to identify where Tile was used but we could use the EPSG naming scheme (removing the need for the format and test functions).

>     * `toUnit` -> `getUnit`
Done, shall I mark it as breaking change though?

>     * `formatToEPSG` -> `formatToEpsg` (to be consistent with capital letters in other methods). Or use capitals in all methods with EPSG and TMS.

I chose to kept the old name since we'll surely remove it due to the reasons above.
 
> Also, we could add a `isDegreeUnit` since we expose a `isMetricUnit`.
I think we could add it later if the need arise.

src/Core/Geographic/Crs.ts
jailln: Oh ok thanks, `proj4.defs` is quite misleading :sweat_smile: 

I just checked and it is not exported in the doc though, you may need to add a @typedef tag like for CameraTransformOptions? : `@typedef {Object} CameraTransformOptions`

Also, I'm wandering if it should appear in the Crs doc page or have its own page or be in a `Types` page, WDYT?

src/Core/Geographic/Crs.ts
jailln: thanks :)

jailln:
Thanks for the changes :)

> @jailln I fixed all issues listed in your review.
> 
> I added an [`@internal`](https://tsdoc.org/pages/tags/internal/) tag for all functions that I deemed to be only used internally by itowns. This mainly includes all related epsg and tms functions, that may be removed soon since we split Extent and Tile. I only kept the TMS naming scheme for Tile to identify where Tile was used but we could use the EPSG naming scheme (removing the need for the format and test functions).

I would be clearer to only use the EPSG naming scheme yes.

> > ```
> > * `toUnit` -> `getUnit`
> > ```
> 
> Done, shall I mark it as breaking change though?

I'm not sure it is used but we might add a breaking change to the commit  message just in case.

Desplandis:
Okay, I went a little bit further and removed the two naming schemes (which were not needed since #2412). This means that all `espg/tms` functions are not needed anymore, so I removed them too. However, shall I keep the `espg/tms` functions for backward compatibility (and deprecated them) but remove reference in the codebase?

I don't think they are used outside of the codebase IMO.

As for the documentation, `ProjectionLike` documentation is not generated with JSDoc but is generated with TypeDoc. I however did change our documentation engine for now, since it will introduce a lot of conflicts with #2414. Waiting for it to be merged!

src/Core/Geographic/Crs.ts
Desplandis: The equivalent of typedef in typescript or either an `interface` or a `type`. Instead of being in the documentation, we can write the types directly.

jailln: Yes but except if I'm missing something it is still not exported in the doc (see http://localhost:8080/docs/out/#api/Geographic/CRS after running `npm run doc`).

jailln: Just saw your other comment :)

jailln:
> Okay, I went a little bit further and removed the two naming schemes (which were not needed since #2412). This means that all `espg/tms` functions are not needed anymore, so I removed them too. However, shall I keep the `espg/tms` functions for backward compatibility (and deprecated them) but remove reference in the codebase?
> 
> I don't think they are used outside of the codebase IMO.

Thanks! that was a much needed cleaning :) Removing them like you did works for me.

> As for the documentation, `ProjectionLike` documentation is not generated with JSDoc but is generated with TypeDoc. I however did change our documentation engine for now, since it will introduce a lot of conflicts with #2414. Waiting for it to be merged!

Ok, thanks.

jailln (approved):

Desplandis:
Added some breaking change notes in the commit messages.

PR #2437, opened 2024-10-11
refactor: migrate Coordinates and Ellipsoid to typescript

Desplandis:
## Description

This PR includes:
- Migration of `Coordinates` from javascript to typescript
- Migration of `Ellipsoid` from javascript to typescript
- Add missing types
- Code formatting
- TSDoc compliant documentation

Proposed breaking type changes:
- For simplification and alignment with `Three.Vector3`, I propose that the constructor of `Coordinates` only accepts numbers as parameters. We could set from array, coordinates-like objects from corresponding methods `setFromArray()` or `setFromVector3`. Note that for js users, we would not for now break compatibility (maybe add a depreciation notice).

## Motivation and Context

As described in proposal #2396, we aim to gradually migrate our entire codebase (with the exception of deprecated modules) from javascript to typescript. We choose to start with modules with no-dependency and move up in the dependency tree.

This PR is the second step of **Migrate geographic modules**, migrating the both co-dependent modules `Coordinates` and `Ellipsoid`. This PR is a follow-up of #2436.

### DIFF
--- a/src/Core/Geographic/Coordinates.ts
+++ b/src/Core/Geographic/Coordinates.ts
@@ -34,50 +34,59 @@ function proj4cache(crsIn: string, crsOut: string): proj4.Converter {
 }
 
 /**
- * A Coordinates object, defined by a [crs](http://inspire.ec.europa.eu/theme/rs)
- * and three values. These values are accessible through `x`, `y` and `z`,
- * although it can also be accessible through `latitude`, `longitude` and
- * `altitude`. To change a value, prefer the `set()` method below.
+ * A class representing a geographic or geocentric coordinate.
  *
- * `EPSG:4978` and `EPSG:4326` are supported by default. To use another CRS,
- * you have to declare it with `proj4`. You can find most projections and their
- * proj4 code at [epsg.io](https://epsg.io/).
+ * A coordinate is defined by a [CRS](http://inspire.ec.europa.eu/theme/rs)
+ * (Coordinate Reference System) and a 3-dimensional vector `(x, y, z)`.
+ * For geocentric projections, it is recommended to use the `latitude`,
+ * `longitude` and `altitude` aliases to refer to vector components.
  *
- * @example
- * new Coordinates('EPSG:4978', 20885167, 849862, 23385912); //Geocentric coordinates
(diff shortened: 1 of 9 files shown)


src/Core/Geographic/Coordinates.ts
```diff
@@ -0,0 +1,364 @@
...
+        // slower (see https://jsbench.me/40jumfag6g/1)
+        this.x = 0;
+        this.y = 0;
+        this.z = 0;
+
+        // Normal
+        this._normal = new THREE.Vector3();
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        if ((v0 as any).length > 0) { // deepscan-disable-line
```
AnthonyGlt:
```suggestion
        if ((x as any).length > 0) { // deepscan-disable-line
```
?

AnthonyGlt (commented):
Thanks for the MR, well done :D

Should we not specified the return type of the different function of these new TS files ?
For example, for the function cartesianToCartographic of Ellipsoid
In the JS doc, we were specifying the return type, it was helping while reading the code.
I get that TypeScript is capable of inferring the return type but could be nice to still write it, like you did in Coordinates.ts. WDYT ?

src/Core/Geographic/Coordinates.ts
Desplandis: I would like to remove the instanciation of `Coordinates` with an array or a vector3  for typescript users but not break the API for javascript users until a major version bump. However, it seems the deepscan static checker can infer that `x` (even if cast as `any`) has no property `length` and the CI failed. This is will be removed with the next major version bump.

Desplandis:
@AnthonyGlt Return types in `Ellipsoid` are now explicit!

src/Core/Geographic/Coordinates.ts
AnthonyGlt: I don't understand why we do the condition on `v0` though ? It's not given by the user , isn't ?

Desplandis: Holy crap, I shall read more carefully, this is a suggested change, thought it was my code. xD
I indeed intended to check the value of `x`, will fix it!

Desplandis:
@AnthonyGlt All good!

AnthonyGlt (approved):

PR #2438, opened 2024-10-15
fix(Zoom): use zoom state

AnthonyGlt:
## Description
Switch state to `ZOOM` when zoom by mouse wheel is triggered.

## Motivation and Context
In a use case of mine, I'm using a similar function than `zoomToTile` of  OGC3DTilesHelper.js but instead of `view.controls.lookAtCoordinate`, I'm calling `CameraUtils.transformCameraToLookAtTarget(view, view.camera3D, cameraTransform);`

By doing this, zooming after calling this method was translating me in random direction (cf video)

[Screencast from 15-10-2024 16:10:24.webm](https://github.com/user-attachments/assets/83e4fbc3-4fbb-453e-8eb7-4a50451ebc2e)

(I'm using 3dtiles_loader.html to try it out)

By replacing the `currentState` by `ZOOM`, I don't encounter this issue anymore, because I need this to be called:

https://github.com/iTowns/itowns/blob/cfb9d0f5184f6d569a3f81101778dd767df9a875/src/Controls/GlobeControls.js#L584

We could also see if this part is really necessary:

https://github.com/iTowns/itowns/blob/cfb9d0f5184f6d569a3f81101778dd767df9a875/src/Controls/GlobeControls.js#L568-L570

### DIFF
--- a/src/Controls/StateControl.js
+++ b/src/Controls/StateControl.js
@@ -265,7 +265,7 @@ class StateControl extends THREE.EventDispatcher {
     touchToState(finger) {
         for (const key of Object.keys(DEFAULT_STATES)) {
             const state = this[key];
-            if (state.enable && finger == state.finger) {
+            if (state.enable && finger === state.finger) {
                 return state;
             }
         }
@@ -396,7 +396,7 @@ class StateControl extends THREE.EventDispatcher {
 
         if (this.enabled && this.ZOOM.enable) {
             viewCoords.copy(this._view.eventToViewCoords(event));
-            this.currentState = this.NONE;
+            this.currentState = this.ZOOM;
             this.dispatchEvent({ type: this.ZOOM._event, delta: event.deltaY, viewCoords });
         }
     }


AnthonyGlt:
@jailln

src/Controls/StateControl.js
```diff
@@ -396,7 +396,7 @@ class StateControl extends THREE.EventDispatcher {
 
         if (this.enabled && this.ZOOM.enable) {
             viewCoords.copy(this._view.eventToViewCoords(event));
-            this.currentState = this.NONE;
+            this.currentState = this.ZOOM;
```
jailln, line 399: Why not making it the first instruction of the if?

jailln (approved):

jailln:
Fine by me, this seems legit!

mgermerie:
Switching the state in a `'mousewheel'` event causes a code inconsistency.

When pressing a key or a mouse button, `'keydown'` and `'pointerdown'` events allow us to change the `currentState` to a relevant one. For these cases, events triggered when the key or button is no longer pressed (`'keyup'` and `'pointerup'`) allow us to reset the `currentState` to `NONE` (i.e. `{}`) when needed.

As far as I know, there is no event for mouse-wheel that triggers when mouse-wheel stops being rolled. That is the reason why there was no `currentState` change within `'mousewhell'` event callback.

With this PR's modifications, the `currentState` is changed when rolling the mouse-wheel, but never set back to `NONE`. You can test it by typing `view.controls.states.currentState` in your borwser's console on the [globe view example](https://www.itowns-project.org/itowns/dev/examples/view_3d_map.html).

jailln:
> Switching the state in a `'mousewheel'` event causes a code inconsistency.
> 
> When pressing a key or a mouse button, `'keydown'` and `'pointerdown'` events allow us to change the `currentState` to a relevant one. For these cases, events triggered when the key or button is no longer pressed (`'keyup'` and `'pointerup'`) allow us to reset the `currentState` to `NONE` (i.e. `{}`) when needed.
> 
> As far as I know, there is no event for mouse-wheel that triggers when mouse-wheel stops being rolled. That is the reason why there was no `currentState` change within `'mousewhell'` event callback.
> 
> With this PR's modifications, the `currentState` is changed when rolling the mouse-wheel, but never set back to `NONE`. You can test it by typing `view.controls.states.currentState` in your borwser's console on the [globe view example](https://www.itowns-project.org/itowns/dev/examples/view_3d_map.html).

@mgermerie we could set it to none once the zoom handler has finished his job maybe ?

mgermerie:
> @mgermerie we could set it to none once the zoom handler has finished his job maybe ?

That would be a workaround, but in terms of architecture, `currentState` was not imagined to be edited by controls. It should just be set in `StateControls`, according to user input. Controls should only read its value to know how to move the camera.

A much better architecture in my opinion would be having a `StateControls` that listens to user input, and dispatches corresponding custom events (such as `'rotate'` or `'drag'` for instance). Controls would then just listen to these events and apply movement to the camera. In this approach, we could probably get rid of these consideration with `currentState`. However, the first implementation of `StateControls` did not go that far and it would need a bit of time to implement it.

**On this PR more specific matter :**

> We could also see if this part is really necessary:
> 
> https://github.com/iTowns/itowns/blob/cfb9d0f5184f6d569a3f81101778dd767df9a875/src/Controls/GlobeControls.js#L568-L570

This is necessary to trigger damping movement. When user stops pressing mouse button, the `currentState` switches to `NONE`, `GlobeControls.onStateChange` matching event is triggered, this part gets executed and starts the damping animation if needed.

Finally, I tested the actions described in globe view example and did not get the error described, as can be seen in this video : 

https://github.com/user-attachments/assets/0a908068-e524-41e4-8efd-1105c219ad94

jailln:
> > @mgermerie we could set it to none once the zoom handler has finished his job maybe ?
> 
> That would be a workaround, but in terms of architecture, `currentState` was not imagined to be edited by controls. It should just be set in `StateControls`, according to user input. Controls should only read its value to know how to move the camera.

> A much better architecture in my opinion would be having a `StateControls` that listens to user input, and dispatches corresponding custom events (such as `'rotate'` or `'drag'` for instance). Controls would then just listen to these events and apply movement to the camera. In this approach, we could probably get rid of these consideration with `currentState`. However, the first implementation of `StateControls` did not go that far and it would need a bit of time to implement it.


I meant using an event dispatched from the `GlobeControls` (e.g. zoom-end) that would be cactched in stateControls to reset the `currentState`. I agree that the solution you provide should be the way to go in the long term when we have time to rework it.

> On this PR more specific matter :

Did you use `CameraUtils.transformCameraToLookAtTarget` before?

PR #2440, opened 2024-10-16
refactor(TileBuilder): migrate to TypeScript

HoloTheDrunk:
## Description
Create types for the tile builder section of iTowns.
Improve performance of tile creation.
Improve code readability of tile creation.

~~Depends on #2447.~~ Merged.

## Summary of changes
- Lots of renaming, most notably the `BuilderEllipsoidTile` being changed to `GlobalTileBuilder` to match the naming of the `PlanarTileBuilder` (especially considering its position inside a `Globe` folder). Hesitated between "Globe" and "Global" so feel free to criticize the choice, it's already a breaking change so we might as well pick the best one.
- Resolve circular type dependency between builders and their parameters by making the new builder interface generic over type parameters.
- Significantly refactor `computeBuffers.js` into `computeBuffers.ts`, changing algorithm to take advantage of simple math to avoid using slow JS constructs in favor of `ArrayBuffer`s and moving some checks out of loops.

## Motivation and Context
This fits into the larger motion towards converting iTowns to TypeScript.
Code had to be changed and moved around to allow static type checking.
Some of the functions around buffer generation were cryptic and had low readability, which required additional changes.

Tested on Ubuntu 22.04.5 LTS x86_64 using Firefox 126.0.1 (64-bit)


HoloTheDrunk:
It'd be great if I could please get a go-ahead on the general structure of this refactor so I can start working on the next one—which will inevitably be based off of this one, while the specific details of the PR get reviewed :)

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+
+const PI_OV_FOUR = Math.PI / 4;
+const INV_TWO_PI = 1.0 / (Math.PI * 2);
+const axisZ = new THREE.Vector3(0, 0, 1);
+const axisY = new THREE.Vector3(0, 1, 0);
+const quatToAlignLongitude = new THREE.Quaternion();
+const quatToAlignLatitude = new THREE.Quaternion();
+const quatNormalToZ = new THREE.Quaternion();
+
+function WGS84ToOneSubY(latitude: number) {
```
jailln: Can you document this method please?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+    return 1.0 - (0.5 - Math.log(Math.tan(
+        PI_OV_FOUR + THREE.MathUtils.degToRad(latitude) * 0.5,
+    )) * INV_TWO_PI);
+}
+
+export interface GlobeTileBuilderParams extends TileBuilderParams {
+    nbRow: number;
+    deltaUV1: number;
```
jailln, line 37: Can you document these params (or at least deltaUV1) please?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+    quatNormalToZ: THREE.Quaternion;
+}
+
+export class GlobeTileBuilder
```
jailln, line 46: Can you document the purpose of this class and its properties please?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+implements TileBuilder<GlobeTileBuilderParams> {
+    private _crs: string;
+    private _transform: {
+        coords: Coordinates[];
+        position: THREE.Vector3;
+        dimension: THREE.Vector2;
+    };
+
+    public computeExtraOffset?: (params: GlobeTileBuilderParams) => number;
```
jailln, line 60: Can you document this method please and why it is optional?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+        uvCount: number,
+    }) {
+        this._transform = {
+            coords: [
+                new Coordinates('EPSG:4326', 0, 0),
+                new Coordinates('EPSG:4326', 0, 0),
+            ],
+            position: new THREE.Vector3(),
+            dimension: new THREE.Vector2(),
+        };
```
jailln, line 77: These could be initialized directly in properties declaration above, wdyt?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export interface GlobeTileBuilderParams extends TileBuilderParams {
+    nbRow: number;
+    deltaUV1: number;
+    quatNormalToZ: THREE.Quaternion;
+}
+
+export class GlobeTileBuilder
+implements TileBuilder<GlobeTileBuilderParams> {
+    private _crs: string;
```
jailln: The globe is always in `EPSG:4978`. I would be down for setting it here and not pass it in the constructor. Especially since this class is not exposed outside of itowns. WDYT?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    public computeExtraOffset?: (params: GlobeTileBuilderParams) => number;
+
+    public get crs(): string {
+        return this._crs;
+    }
+
+    public constructor(options: {
+        crs: string,
+        uvCount: number,
```
jailln, line 68: Can you document uvCount please?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+        // UV: Normalized coordinates (from degree) on the entire tile
+        // EPSG:4326
+        // Offset: Float row coordinate from Pseudo mercator coordinates
+        // EPSG:3857
+        if (options.uvCount > 1) {
+            this.computeExtraOffset = GlobeTileBuilder._computeExtraOffset;
+        }
+    }
+
+    private static _computeExtraOffset(params: GlobeTileBuilderParams): number {
```
jailln: Can you document this method and replace t with a more explicit variable name please?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+        const t = WGS84ToOneSubY(params.projected.latitude) * params.nbRow;
+        return (!isFinite(t) ? 0 : t) - params.deltaUV1;
+    }
+
+    // prepare params
+    // init projected object -> params.projected
```
jailln: can you improve documentation of this method please?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+            // let's avoid building too much temp objects
+            projected: new Projected(),
+        };
+
+        params.extent.planarDimensions(this._transform.dimension);
+
+        return { ...params, ...newParams };
+    }
+
+    // get center tile in cartesian 3D
```
jailln: this is purely aesthetics but I think we should follow jsdoc comment style for methods. WDYT @Desplandis?

src/Core/Prefab/Globe/GlobeTileBuilder.ts
```diff
@@ -0,0 +1,170 @@
...
+        return extent.west + u * this._transform.dimension.x;
+    }
+
+    // coord v tile to projected
+    public vProject(v: number, extent: Extent): number {
+        return extent.south + v * this._transform.dimension.y;
+    }
+
+    public computeShareableExtent(extent: Extent): ShareableExtent {
+        // Compute shareable extent to pool the geometries
```
jailln: should be moved above the method for readability

src/Core/Prefab/Planar/PlanarTileBuilder.ts
```diff
@@ -0,0 +1,112 @@
...
+    Projected,
+    ShareableExtent,
+    TileBuilder,
+    TileBuilderParams,
+} from '../TileBuilder';
+
+const quaternion = new THREE.Quaternion();
+const center = new THREE.Vector3();
+
+type Transform = {
```
jailln, line 13: Can you also define it as a type in GlobeTileBuilder for symmetry? Also, we should comment why it's different I think

src/Core/Prefab/Planar/PlanarTileBuilder.ts
```diff
@@ -0,0 +1,112 @@
...
+    coords: Coordinates,
+    position: THREE.Vector3,
+    normal: THREE.Vector3,
+};
+
+export interface PlanarTileBuilderParams extends TileBuilderParams {
```
jailln, line 20: I have the same documentation propositions than for GlobeTileBuilder here.

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+type IndexArray = Option<Uint8Array | Uint16Array | Uint32Array>;
+
+export type Buffers = {
+    index: IndexArray,
+    position: Float32Array,
+    normal: Float32Array,
+    uvs: [Option<Float32Array>, Option<Float32Array>],
+};
+
+type TmpBuffers = Buffers & {
```
jailln: BufferAndSkirt instead?

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+    skirt: IndexArray,
+};
+
+function pickUintArraySize(
```
jailln: I think we should rename to getUintArrayConstructor or getUintArraySize to avoid any confusion with mouse picking that is used elsewhere in itowns

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+): Option<{ index: IndexArray, skirt: IndexArray }> {
+    if (!params.buildIndexAndUv_0) {
+        return undefined;
+    }
+
+    const indexBufferSize = getBufferIndexSize(nSeg, params.disableSkirt);
+    const indexConstructor = pickUintArraySize(nVertex);
+
+    const tileLen = indexBufferSize;
+    const skirtLen = 4 * nSeg;
```
jailln, line 56: Not the purpose of this PR but 4 should be configurable. WDYT?

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+
+function initComputeUv1(value: number): (uv: Float32Array, id: number) => void {
+    return (uv: Float32Array, id: number): void => { uv[id] = value; };
+}
+
+type ComputeUvs =
+    [typeof computeUv0 | (() => void), ReturnType<typeof initComputeUv1>?];
+
+// TODO: Split this even further into subfunctions
+export function computeBuffers(
```
jailln, line 132: At least this one should be documented here since it is exported

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+
+            const vertex = builder.vertexPosition(params.projected);
+            const normal = builder.vertexNormal();
+
+            // move geometry to center world
+            vertex.sub(params.center);
+
+            // align normal to z axis
+            // HACK: this check style is not great
+            if ('quatNormalToZ' in params) {
```
jailln, line 193: Can you check params type instead?

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+            // | | / | / |   o = stored vertices
+            // | o---+---+   s = already stored
+            // | | / | / |
+            // | s---+---+
+            outBuffers.skirt![3 * nVertex - 3 + y] =
+                nVertex * (nVertex - 1 - y);
+        }
+    }
+
+    function bufferizeTri(id: number, va: number, vb: number, vc: number) {
```
jailln, line 248: Can you document this one too please ?

src/Core/Prefab/computeBufferTileGeometry.ts
```diff
@@ -0,0 +1,338 @@
...
+    if (params.buildIndexAndUv_0) {
+        for (let y = 0; y < nSeg; y++) {
+            for (let x = 0; x < nSeg; x++) {
+                const v1 = y * nVertex + (x + 1);
+                const v2 = y * nVertex + x;
+                const v3 = (y + 1) * nVertex + x;
+                const v4 = (y + 1) * nVertex + (x + 1);
+
+                const id = (y * nSeg + x) * 6;
+                bufferizeTri(id, /**/v4, v2, v1);
```
jailln, line 263: unnecessary /**/

src/Core/Prefab/TileBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+const cacheTile = new Cache();
+
+export type GpuBufferAttributes = {
+    index: THREE.BufferAttribute | null;
+    position: THREE.BufferAttribute;
+    normal: THREE.BufferAttribute;
+    uvs: THREE.BufferAttribute[];
+};
+
+export type ShareableExtent = {
```
jailln, line 23: can you describe what a ShareableExtent is ?

src/Core/Prefab/TileBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+    shareableExtent: Extent;
+    quaternion: THREE.Quaternion;
+    position: THREE.Vector3;
+};
+
+// TODO: Check if this order is right
+// Ideally we split this into Vec2 and a simpler LatLon type
+// Somewhat equivalent to a light Coordinates class
+export class Projected extends THREE.Vector2 {
```
jailln: Do we really need this class ? To reduce complexity, I think we should either use Vector2 directly and make it clear in variable names that it represents LatLon or user Coordinates if more properties and method are needed.

src/Core/Prefab/TileBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+        this.y = latitude;
+    }
+}
+
+export interface TileBuilderParams {
+    /** Whether to build the skirt. */
+    disableSkirt: boolean;
+    /** Whether to render the skirt. */
+    hideSkirt: boolean;
+    buildIndexAndUv_0: boolean;
```
jailln, line 38: can you comment this one please?

src/Core/Prefab/TileBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+    /** Number of segments (edge loops) inside tiles. */
+    segments: number;
+    /** Buffer for projected points. */
+    projected: Projected;
+    extent: Extent;
+    level: number;
```
jailln, line 45: what is the difference between level and zoom ?

src/Core/Prefab/TileBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+    /** Convert builder-agnostic params to specialized ones. */
+    prepare(params: TileBuilderParams): SpecializedParams;
+    computeExtraOffset?: (params: SpecializedParams) => number;
+    /** Get the center of the tile in 3D cartesian coordinates. */
+    center(extent: Extent): THREE.Vector3;
+    vertexPosition(position: THREE.Vector2): THREE.Vector3;
+    vertexNormal(): THREE.Vector3;
+    uProject(u: number, extent: Extent): number;
+    vProject(v: number, extent: Extent): number;
+    computeShareableExtent(extent: Extent): ShareableExtent;
```
jailln, line 74: Can you document all these methods please?

src/Core/Prefab/TileBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+    const { shareableExtent, quaternion, position } =
+        builder.computeShareableExtent(params.extent);
+
+    const south = shareableExtent.south.toFixed(6);
+
+    const bufferKey =
+        `${builder.crs}_${params.disableSkirt ? 0 : 1}_${params.segments}`;
+
+    let promiseGeometry = cacheTile.get(south, params.level, bufferKey);
+    // let promiseGeometry;
```
jailln: to remove

src/Core/TileGeometry.ts
```diff
@@ -0,0 +1,157 @@
...
+    public segments: number;
+    public tileCenter: THREE.Vector3;
+
+    private _refCount: {
+        count: number,
+        fn: () => void,
+    } | null;
+
+    public constructor(
+        builder: TileBuilder<TileBuilderParams>,
```
jailln, line 74: The design pattern is not really clear here... There are some circular dependencies and I think that separation of concerns is not really effective. I think we should aim at: (and we are not far from it)
* interface TileBuilder
* implemented by PlanarTileBuilder and GlobeTileBuilder
* each having a method returning a new TileGeometry based on relevant TileBuilderParams (the current newTileGeometry)
* TileGeometry being only a data model that does not know about the TileBuilder.

WDYT @Desplandis @HoloTheDrunk?

(Note that we don't necessarily have to implement it in this PR)

src/Core/TileGeometry.ts
```diff
@@ -0,0 +1,157 @@
...
+    return bufferAttributes;
+}
+
+export class TileGeometry extends THREE.BufferGeometry {
+    public OBB: OBB | null;
+    public extent: Extent;
+    public segments: number;
+    public tileCenter: THREE.Vector3;
+
+    private _refCount: {
```
jailln, line 68: I don't really understand the purpose of _refCount. From what I understand it is always initialized and increased to 1 when we create a new TileGeometry. Am I missing something ?

src/Core/TileGeometry.ts
```diff
@@ -0,0 +1,157 @@
...
+        cacheTile: Cache,
+        keys: [string, number, string],
+    ): void {
+        if (this._refCount !== null) {
+            return;
+        }
+
+        this._refCount = {
+            count: 0,
+            fn: () => {
```
jailln, line 124: Why not directly put this in TileGeometry.dispose instead ?

jailln (commented):
Thanks for all these changes, it's already better and easier to understand with typings. I think we should take this opportunity to improve the architecture and the doc of this part of the code a little bit more, that's what stands out from my comments. Let me know what you think :)

src/Core/Prefab/Globe/GlobeTileBuilder.ts
HoloTheDrunk: If a value isn't final, I'm more inclined to set it in the constructor to keep both the definition of constant / variable values separate and the cleanliness of the field declarations. Might just be the systems programmer in me speaking, but ideally class structure declaration shouldn't be mixed with initialization if it doesn't provide additional benefits.

src/Core/Prefab/Globe/GlobeTileBuilder.ts
HoloTheDrunk: Not entirely sure what this represents, might just be a temp storage for an unremarkable step of the computation? Matter of fact, I'm not exactly sure why nbRow is defined the way it is so that probably doesn't help...

src/Core/Prefab/Globe/GlobeTileBuilder.ts
HoloTheDrunk: The documentation for this method is available on its actual declaration in the `TileBuilder` interface, should it be duplicated or?

src/Core/TileGeometry.ts
HoloTheDrunk: Letting the builders have their own implementation is the most natural architecture to me as well, since it would also allow us to bind the tile cache currently stored in a global variable inside builder instances, enabling the use of multiple builders in the same program without the issue of having a mixed tile cache that might get a bit unwieldy when debugging.
I would rather leave this for a later PR however, as the goal is to transition to TypeScript while changing only the logic that must be changed to enable static typing.

src/Core/TileGeometry.ts
HoloTheDrunk: This comes into play with the caching used in `TileBuilder.ts:newTileGeometry`. You can see in `convertToTile.js:convert` that the call to increase the reference count is made on the result of `newTileGeometry`, which can be either a new `TileGeometry` or a `TileGeometry` that was retrieved from the global tile cache.
We need the `_refCount` to be able to dispose of the cache entry and GPU device memory only when all references to that shared `TileGeometry` are dropped.

src/Core/TileGeometry.ts
HoloTheDrunk: The disposal logic needs references to values that would be cumbersome to pass when calling `dispose`. On top of that, `dispose` is an overriden method from the `THREE.BufferGeometry` base class so we don't have the luxury of changing its prototype.

src/Core/Prefab/Globe/GlobeTileBuilder.ts
HoloTheDrunk: I'd argue this isn't even aesthetics, jsdoc has become the _de facto_ standard for JS/TS comments. Those comments were just left behind from because I forgot to clean them up, oops.

src/Core/Prefab/Globe/GlobeTileBuilder.ts
HoloTheDrunk: What do you reckon would be a description of the class that provides more information than can be gleaned from the name and implemented interface?

src/Core/Prefab/computeBufferTileGeometry.ts
HoloTheDrunk: The `4` here is the number of vertex indices per segment of the edge since we're creating squares. The rest of the generation wouldn't work with anything different (either failing out of bounds checks or leaving trailing uninitialized memory). I also struggle to see a use case for doing something other than squares to be quite honest, do you have any in mind?

src/Core/Prefab/computeBufferTileGeometry.ts
HoloTheDrunk: Type information is erased when transpiling to JS and therefore unavailable at runtime. We could make the function generic over `builder` and `params` types and jump through a few hoops but I don't think the added complexity is worth it.

src/Core/Prefab/computeBufferTileGeometry.ts
HoloTheDrunk: This is a bit of a hack to avoid disabling formatting while visually aligning the passed indices with the line right below this one.

src/Core/Prefab/TileBuilder.ts
HoloTheDrunk: Using Vector2 would decrease readability as the fields are often accessed without intermediary variables. On the other hand, Coordinates is overly complex for the task at hand and I was worried that might affect memory usage while refactoring the code. With hindsight, it probably wouldn't and I will change it.

src/Core/Prefab/TileBuilder.ts
HoloTheDrunk: Jolly good question. I'm pretty sure there's none as one was used exclusively for globe and the other one for planar, seems I forgot to merge them.
After looking around iTowns a bit, I'm opting to name the merged field `level`.

src/Core/Prefab/Globe/GlobeTileBuilder.ts
jailln: Yes I'm ok with that

src/Core/Prefab/Globe/GlobeTileBuilder.ts
jailln: What you added is fine :)

src/Core/Prefab/Globe/GlobeTileBuilder.ts
jailln: Ok, let's keep it like this then, I was just trying to get the knowledge you gathered while working on this in comments :)

src/Core/Prefab/Globe/GlobeTileBuilder.ts
jailln: Yep I saw this afterwards sorry, no need to duplicate it

src/Core/Prefab/computeBufferTileGeometry.ts
jailln: Oh yes sorry, I thought it was the actual lengh of the skirt in the view, I must have been tired at this point of the review :sweat_smile:

src/Core/Prefab/computeBufferTileGeometry.ts
jailln: Ok too bad :(

src/Core/TileGeometry.ts
jailln: Ok, I created #2470 so we don't forget about it

src/Core/TileGeometry.ts
jailln: Ok thanks for the explaination, I missed the cached ones. It makes me think that `newTileGeometry` could be renamed to `getTileGeometry`, it would be less misleading.
Regarding `_refCount`, what do you think about leaving it as a number and putting the logic of `fn` in `TileGeometry:dispose` directly? That may be easier to follow.

HoloTheDrunk: As mentioned in my reply to your other comment about this topic, `dispose` is an overridden method from `THREE.BufferGeometry` so we don't have the leeway to change its parameters, which would be needed considering that the custom disposal logic makes use of references to values that ideally shouldn't be kept as additional fields within the class. There might be a better way to do this, but I currently don't see it.

jailln: The two issues are to tell the cache that it something must be deleted and to identify the tile to delete. In the current architecture, one way would be to add an `id` to `TileGeometry` and dispatch an event in `TileGeometry.dispose` trapped by the cache.

HoloTheDrunk: `TileGeometry` already inherits from `THREE.BufferGeometry`, we could use a mixin to get multiple inheritance or have an `EventDispatcher` field but that's arguably even less readable.
In general, I'd rather avoid relying on events as much as possible: they're an easy solution in a lot of cases, but it's more of a bypass of architecture than anything and they make code flow harder to follow by creating indirection in what should otherwise be simple single-threaded logic, which is at least in part what we're trying to avoid by moving to TS. In an application this would be fine (taking for example the fact that the Godot game engine exposes a similar "signals" system and that works out fine), but for the core internals of a library I believe this is detrimental to maintainability.
Currently, despite the initial surprise of the reference counting code, its locality makes it easy to read and reason about: everything related to the handling of the reference count is in one file and easily fits into a single screen, and changes can be freely made to the cache with only minor changes required in `TileGeometry`.

jailln: I agree that events can hide an architectural flaw, which would indeed be the case here but I think this is also the purpose of this proposed implementation . The question is which one we'd rather have. I think that events would make it easier to read and I proposed that to avoid a too big refactoring but let's give it some thought and see what we do after.

In my opinion, the clean way to go would probably be that the object that takes care of `TileGeometry` disposal also takes care of removing it from the cache if the tile is not referenced anymore. By looking at the code, I think that tile cache should be stored in `TileGeometryLayer` instead of `TileBuilder` and that it should be the responsability of the `TileGeometryLayer` to remove obsolete tiles from its cache. What do you think ? That would remove the need to store a reference to the whole tile cache in each `TileGeometry`.

HoloTheDrunk: Storing the tile cache in `TileGeometryLayer` prevents us from reusing tiles between layers, which could lead to a pretty significant increase in memory usage in complex use cases.
The cache can't be stored in `TileBuilder` either since it's an interface.
After discussing it with @Desplandis we came to the conclusion that the best option would be to create a static alternative constructor for `TileGeometry` that would return a `Proxy` over `TileGeometry` with the caching logic built-in.
As the entire tile generation pipeline is likely to change in the medium term (say, within the next year or so if we don't run into too many issues), we're waiting for that to really kick the current pipeline to the curb and build anew.

jailln: Ok, let's discuss it further at that time. Meanwhile, can you add a comment above `refCount` to explain why it is implemented like this please? (essentially kind of a summary of what we said here, or maybe even a link to this discussion?)

jailln (approved):
Thanks for the commenting and refactoring efforts :pray:

HoloTheDrunk:
Seems like the GitHub Actions are stuck. I don't have the option to force restart them, do you @jailln @Desplandis?

PR #2441, opened 2024-10-17
chore(eslint): Add no-use-before-define and change max-len rules

Desplandis:
## Motivation and Context

This PR relax some rules introduced by #2413 (100 characters per line of code, still 80 chars per line of comment except line including URLs). It also replace the `no-use-before-define` by `@typescript-estlint/no-use-before-define` as per [typescript eslint documentation](https://typescript-eslint.io/rules/no-use-before-define/) since the former was also throwing errors on types used before definition.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -136,7 +136,14 @@ module.exports = {
                 'plugin:@typescript-eslint/recommended',
             ],
             rules: {
-                '@stylistic/max-len': ['warn', 80],
+                '@stylistic/max-len': ['warn', {
+                    code: 100,
+                    comments: 80,
+                    ignoreUrls: true,
+                }],
+                // see https://typescript-eslint.io/rules/no-use-before-define/
+                'no-use-before-define': 'off',
+                '@typescript-eslint/no-use-before-define': 'error',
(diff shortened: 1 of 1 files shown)


mgermerie (approved):

.eslintrc.cjs
```diff
@@ -136,7 +136,14 @@ module.exports = {
                 'plugin:@typescript-eslint/recommended',
             ],
             rules: {
-                '@stylistic/max-len': ['warn', 80],
+                '@stylistic/max-len': ['warn', {
+                    code: 100,
+                    comments: 80,
```
jailln, line 141: Why not setting comments to 100 too ?

Desplandis: I would have likely have 80 characters limit everywhere (which is pretty standard in C-like languages). However we have long functions and methods name in long chain which in practice eat the 80 characters limit... So I decided to change it only for code, I could put comment too in 100 characters if you wish !

jailln: I agree that 80 is more common. Generally it's either 80 or 120 but 100 is fine by me too. However I think it would be easier to have the same limit for both comments and code.

PR #2447, opened 2024-10-22
feat(eslint): remove preference for default export

HoloTheDrunk:
## Description
Change ESlint config to remove the preference for default exports.
Also clear a warning caused by unnecessarily quoted object fields in the config.

## Motivation and Context
Default exports mess with tree shaking and are generally better avoided.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -60,6 +60,7 @@ module.exports = {
       devDependencies: ['test/**', 'tests/**', 'examples/**'],
     }], */
         'import/no-extraneous-dependencies': 'off',
+        'import/prefer-default-export': 'off',
 
         // TODO reactivate all the following rules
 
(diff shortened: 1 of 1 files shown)


jailln:
Thanks for this PR :) Should we modify all our exports then ? WDYT?

HoloTheDrunk:
That's the end goal, but I believe we're better off making these changes gradually as we progress through the TS rewrite, essentially grouping edits of the library into isolated chunks per PR instead of one massive sweep. Easier to keep track of and nicer for the `git diff`s.

jailln:
> That's the end goal, but I believe we're better off making these changes gradually as we progress through the TS rewrite, essentially grouping edits of the library into isolated chunks per PR instead of one massive sweep. Easier to keep track of and nicer for the `git diff`s.

Ok, fine by me

jailln (approved):

Issue #869, opened 2018-10-12
Object Removal Helper and Cache


gchoqueux:
the` Object Removal Helper` causes inconsistencies when it assigns null to a geometry present in the `Cache`. Actually, the cache can provide a mesh whose geometry is null. 
Think of a more intelligent cleaning system.

PR #2451, opened 2024-10-24
replace Cache by Lru cache

ftoromanoff:
replace our own itowns.cache per the lru-cache package.

Currently we use 1 cache for each Layer + 1 for Style and another one for tile. Should we change that ?
I also arbitrary fixed the max size of each cache to 500. Maybe this should be discuse, depending of the type of cache ?

Other question what should I do with export const CACHE_POLICIES, I currently kept the 'Cache.js' file just for this.


jailln:
@ftoromanoff does it solve #869 ?

ftoromanoff:
> @ftoromanoff does it solve #869 ?

I can't say. Is there a way to test it ?

src/Core/TileGeometry.ts
```diff
@@ -131,7 +131,7 @@ export class TileGeometry extends THREE.BufferGeometry {
                     //  (in THREE.WebGLBindingStates code).
                     this.index = null;
                     delete this.attributes.uv;
-                    cacheTile.delete(...keys);
+                    cacheTile.delete(`s${keys[0]}l${keys[1]}bK${keys[2]}`);
```
AnthonyGlt: I don't know we are giving an array `keys` instead of directly the string corresponding to the `key`

src/Source/VectorTilesSource.js
```diff
@@ -171,26 +171,18 @@ class VectorTilesSource extends TMSSource {
...
-                    .then(collections => mergeCollections(collections))
-                    .catch(err => this.handlingError(err)),
-                key);
+            features = Promise.all(this.urls.map(url =>
+                this.fetcher(this.urlFromExtent(extent, url), this.networkOptions)
+                    .then(file => this.parser(file, { out, in: this, extent }))))
+                .then(collections => mergeCollections(collections))
+                .catch(err => this.handlingError(err));
 
-            if (this.onParsedFile) {
```
AnthonyGlt, line 187: In the changelog, needs to specify that it's not supported anymore

src/Core/Prefab/TileBuilder.ts
```diff
@@ -86,13 +86,14 @@ export function newTileGeometry(
     const bufferKey =
         `${builder.crs}_${params.disableSkirt ? 0 : 1}_${params.segments}`;
 
-    let promiseGeometry = cacheTile.get(south, params.level, bufferKey);
+    const key = `s${south}l${params.level}bK${bufferKey}`;
```
AnthonyGlt, line 87: Just a thought:
It would have been nice to have something similar to   `getDataKey` but, in this case, it could be too verbose.

test/unit/source/filesource.js
```diff
@@ -0,0 +1,108 @@
...
+                done();
+            }).catch(done);
+    });
+
+    it('should instance FileSource with source.fetchedData and parse data with a layer', function (done) {
+        // TO DO need cleareance: what is this test for ?
+        //  - testing instanceation Filesource when fetchedData and source.feature is already available ?
+        //  - testing instanciate Layer ?
+        //  - testing source.onLayerAdded ?
+        //  - testing souce.loadData ?
```
AnthonyGlt, line 53: This todo hasn't been solved during this refacto ?

AnthonyGlt (commented):
Thank you for this integration, it looks nice :+1: 

The refactoring looks good to me, well done

AnthonyGlt:
Is there some improvements from these changes ? Some data that could explain this choice ?
Thank you

src/Core/TileGeometry.ts
ftoromanoff: Done

src/Source/VectorTilesSource.js
ftoromanoff: done

src/Core/Prefab/TileBuilder.ts
ftoromanoff: We only call it once so I don't think it's really needed...

test/unit/source/filesource.js
ftoromanoff: I just did a reorganisation of the Source unit tests

ftoromanoff:
> Is there some improvements from these changes ? Some data that could explain this choice ? Thank you

1) better performance when using an LRU cache.
2) No more maintenance of this module
3) We now have a memory bound for the cache

AnthonyGlt (approved):

PR #2456, opened 2024-10-29
fix(COG): Fix extent in COG parser

ketourneau:
We see some error with Cloud Optimized Geotiff Parser (`extent.as` is not a function).

I think it's an oversight of https://github.com/iTowns/itowns/commit/4b57498d3f7a8744002347031822c4bee8eaf56a.


Desplandis (approved):
Yeah... Forgot to test the plugins...
I think I will add some automatic test on the plugins to not break it again.

We have some plans to use your code in the main codebase to add official support of COG and tiff for both terrain and imagery layers. =)

PR #2465, opened 2024-11-18
fix(wms) : assign axis order param from source

airnez:
## Description
WMSSource was'nt taking the axisOrder parameter into account when provided.

## Motivation and Context
Consistency: This parameter was advertised in documentation while not being used afterwards.

### DIFF
--- a/src/Core/Geographic/Crs.ts
+++ b/src/Core/Geographic/Crs.ts
@@ -4,6 +4,12 @@ import type { ProjectionDefinition } from 'proj4';
 
 proj4.defs('EPSG:4978', '+proj=geocent +datum=WGS84 +units=m +no_defs');
 
+// Redefining proj4 global projections to match epsg.org database axis order.
+// See https://github.com/iTowns/itowns/pull/2465#issuecomment-2517024859
+proj4.defs('EPSG:4326').axis = 'neu';
+proj4.defs('EPSG:4269').axis = 'neu';
+proj4.defs('WGS84').axis = 'neu';
+
 /**
  * A projection as a CRS identifier string. This identifier references a
  * projection definition previously defined with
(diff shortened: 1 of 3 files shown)


Desplandis (approved):
Thanks for the fix! It was indeed documented and missing...

I have however a few comments with our current implementation in iTowns. Reading [geoserver's notes](https://docs.geoserver.org/main/en/user/services/wms/basics.html#axis-ordering) on the WMS specification, it seems that
- > The WMS 1.1 always uses east/north axis ordering.
   
    Shall we for WMS $<1.3.0$ queries default to `wsen` for all CRS (instead of just `EPSG:4326`)?
 - > The WMS 1.3.0 specification mandates using the axis ordering as defined in the EPSG database.
 
   It would be nice again to not just handle the `EPSG:4326` projection, and maybe use the projection definition from `proj4` to determine the axis ordering [`+axis`](https://proj.org/en/stable/usage/projections.html#axis-orientation). This could be a nice addition to the `CRS` module.

airnez:
@Desplandis `EPSG:4326`, `EPSG:4269` and other CRS are hardcoded in `proj4` as long/lat, not lat/long as described in epsg.org database. This means that we should redefine them by default in itowns in Crs module ? See https://github.com/proj4js/proj4js/blob/1f1494d14a0efcf078c6a9c27f82186ef82cfc53/lib/global.js#L2-L3

Moreover, what should we do if the itowns user does not define the `+axis` param in `proj4` ? I guess that defaulting to `wsen` would be the option. We should then also implement other orientations like `eswn` or `nesw`.

If you are ok with this analysis I am going to push it here.

airnez:
@Desplandis added suggested features and rebased on master since Crs module was migrated to Ts

Desplandis:
Sounds good to me!

PR #2466, opened 2024-11-20
feat(MVT): change mapbox package to maplib

ftoromanoff:
The mapbox package currently used is no more upadted due to change of the licencing (few years ago).

This PR aim at switching from the mapbox-gl-style-spec package to the mapblib-gl-style-spec package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "2.44.2",
       "license": "(CECILL-B OR MIT)",
       "dependencies": {
-        "@mapbox/mapbox-gl-style-spec": "^13.28.0",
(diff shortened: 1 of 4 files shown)


jailln:
Are they following the same style specification (i.e. the mapbox one) or did maplibre also started its own style spec ?

ftoromanoff:
> Are they following the same style specification (i.e. the mapbox one) or did maplibre also started its own style spec ?

Yes, it's a fork of the of [mapbox-gl-js](https://github.com/mapbox/mapbox-gl-js), before their switch to a non-OSS license in December 2020. From what I understand, the mapib spec may evolve but will still be based on and support the MapBox one.

jailln (approved):

PR #2469, opened 2024-11-26
MapBox VectorTile: bug fixes using an official MapBox stream

ftoromanoff:
Using the mapbox style file "mapbox://styles/mapbox/streets-v8" on mapLib and on itowns, I noticed the result as not at all comparable.
image
image

Thus I looked layers by layers to understand what were we doing wrong in itowns and tried to correct a maximum of bugs. The example using the mapbox stream has been added. (examples/vector_tile_2d_mapbox.html)

What this PR does:
- new feature:
   - support relative urls in style file (ArcGIS style file)
   - add a warn properties in source to add warnings and send them only ones 
- itowns bug fix:
   - Mapbox stream
      - enter url in {z}/{y}/{x} instead of ${z}/${y}/${x}
      - wrong selection of metadata when "{id}" in sprite
      - gestion of MVT layer using the 'ref' properties
      - gestion of line and polygon Label (simplified -> we use the first point of the geometry and sending a warning)
      - good gestion of the layer.order properties
   - Style
      - take into account the zoom in style when displaying layer
      - no display of polygon when fill.color is undefined
      - no display of line when stroke.width is 0
      - no recoloring of icon when icon.color is white (to avoid a lightning of the icon)
      - allow multiple label with same text (for contour lines)
      - no display of icon if size is 0

### DIFF
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -87,14 +87,18 @@ async function loadImage(source) {
     return (await promise).image;
 }
 
-function cropImage(img, cropValues = { width: img.naturalWidth, height: img.naturalHeight }) {
-    canvas.width = cropValues.width;
-    canvas.height = cropValues.height;
+function cropImage(img, cropValues) {
+    const x = cropValues.x || 0;
+    const y = cropValues.y || 0;
+    const width = cropValues.width || img.naturalWidth;
+    const height = cropValues.height || img.naturalHeight;
+    canvas.width = width;
+    canvas.height = height;
     const ctx = canvas.getContext('2d', { willReadFrequently: true });
     ctx.drawImage(img,
-        cropValues.x || 0, cropValues.y || 0, cropValues.width, cropValues.height,
-        0, 0, cropValues.width, cropValues.height);
-    return ctx.getImageData(0, 0, cropValues.width, cropValues.height);
+        x, y, width, height,
+        0, 0, width, height);
+    return ctx.getImageData(0, 0, width, height);
 }
 
 function replaceWhitePxl(imgd, color, id) {
@@ -861,6 +865,12 @@ class Style {
             // additional icon
             const iconImg = readVectorProperty(layer.layout['icon-image']);
             if (iconImg) {
+                const cropValueDefault = {
+                    x: 0,
+                    y: 0,
+                    width: 1,
+                    height: 1,
+                };
                 try {
                     style.icon.id = iconImg;
                     if (iconImg.stops) {
@@ -871,8 +881,15 @@ class Style {
                                 if (stop[1].includes('{')) {
                                     cropValues = function _(p) {
                                         const id = stop[1].replace(/\{(.+?)\}/g, (a, b) => (p[b] || '')).trim();
+                                        if (cropValues === undefined) {
+                                            // const warning = `WARNING: "${id}" not found in sprite file`;
+                                            sprites[id] = cropValueDefault;// or return cropValueDefault;
+                                        }
                                         return sprites[id];
                                     };
+                                } else if (cropValues === undefined) {
+                                    // const warning = `WARNING: "${stop[1]}" not found in sprite file`;
+                                    cropValues = cropValueDefault;
                                 }
                                 return [stop[0], cropValues];
                             }),
@@ -880,11 +897,18 @@ class Style {
                         style.icon.cropValues = iconCropValue;
                     } else {
                         style.icon.cropValues = sprites[iconImg];
-                        if (iconImg[0].includes('{')) {
+                        if (iconImg.includes('{')) {
                             style.icon.cropValues = function _(p) {
                                 const id = iconImg.replace(/\{(.+?)\}/g, (a, b) => (p[b] || '')).trim();
+                                if (sprites[id] === undefined) {
+                                    // const warning = `WARNING: "${id}" not found in sprite file`;
+                                    sprites[id] = cropValueDefault;// or return cropValueDefault;
+                                }
                                 return sprites[id];
                             };
+                        } else if (sprites[iconImg] === undefined) {
+                            // const warning = `WARNING: "${iconImg}" not found in sprite file`;
+                            style.icon.cropValues = cropValueDefault;
                         }
                     }
                     style.icon.source = sprites.source;
@@ -955,7 +979,7 @@ class Style {
         // need doc for the txtrCtx.fillStyle.src that seems to always be undefined
         if (this.fill.pattern) {
             let img = this.fill.pattern;
-            const cropValues = this.fill.pattern.cropValues;
+            const cropValues = { ...this.fill.pattern.cropValues };
             if (this.fill.pattern.source) {
                 img = await loadImage(this.fill.pattern.source);
             }
@@ -1030,7 +1054,7 @@ class Style {
         if (!this.icon.cropValues && !this.icon.color) {
             icon.src = this.icon.source;
         } else {
-            const cropValues = this.icon.cropValues;
+            const cropValues = { ...this.icon.cropValues };
             const color = this.icon.color;
             const id = this.icon.id || this.icon.source;
             const img = await loadImage(this.icon.source);
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -941,17 +941,16 @@ class Style {
      * @param {Boolean} canBeFilled - true if feature.type == FEATURE_TYPES.POLYGON.
      */
     applyToCanvasPolygon(txtrCtx, polygon, invCtxScale, canBeFilled) {
-        const context = this.context;
         // draw line or edge of polygon
         if (this.stroke) {
             // TO DO add possibility of using a pattern (https://github.com/iTowns/itowns/issues/2210)
-            this._applyStrokeToPolygon(txtrCtx, invCtxScale, polygon, context);
+            this._applyStrokeToPolygon(txtrCtx, invCtxScale, polygon);
         }
 
         // fill inside of polygon
-        if (canBeFilled && this.fill) {
+        if (canBeFilled && (this.fill.pattern || this.fill.color)) {
             // canBeFilled can be move to StyleContext in the later PR
-            this._applyFillToPolygon(txtrCtx, invCtxScale, polygon, context);
+            this._applyFillToPolygon(txtrCtx, invCtxScale, polygon);
         }
     }
 
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -162,7 +162,7 @@ function defineStyleProperty(style, category, parameter, userValue, defaultValue
                 const dataValue = style.context.featureStyle?.[category]?.[parameter];
                 if (dataValue != undefined) { return readExpression(dataValue, style.context); }
                 if (defaultValue instanceof Function) {
-                    return defaultValue(style.context.properties, style.context);
+                    return defaultValue(style.context.properties, style.context) ?? defaultValue;
                 }
                 return defaultValue;
             },
@@ -808,7 +808,8 @@ class Style {
                 style.stroke.color = color;
                 style.stroke.opacity = opacity;
                 style.stroke.width = 1.0;
-                style.stroke.dasharray = [];
+            } else {
+                style.stroke.width  = 0.0;
             }
         } else if (layer.type === 'line') {
             const prepare = readVectorProperty(layer.paint['line-color'], { type: 'color' });
@@ -942,7 +943,7 @@ class Style {
      */
     applyToCanvasPolygon(txtrCtx, polygon, invCtxScale, canBeFilled) {
         // draw line or edge of polygon
-        if (this.stroke) {
+        if (this.stroke.width > 0) {
             // TO DO add possibility of using a pattern (https://github.com/iTowns/itowns/issues/2210)
             this._applyStrokeToPolygon(txtrCtx, invCtxScale, polygon);
         }
@@ -958,11 +959,11 @@ class Style {
         if (txtrCtx.strokeStyle !== this.stroke.color) {
             txtrCtx.strokeStyle = this.stroke.color;
         }
-        const width = (this.stroke.width || 2.0) * invCtxScale;
+        const width = this.stroke.width * invCtxScale;
         if (txtrCtx.lineWidth !== width) {
             txtrCtx.lineWidth = width;
         }
-        const alpha = this.stroke.opacity == undefined ? 1.0 : this.stroke.opacity;
+        const alpha = this.stroke.opacity;
         if (alpha !== txtrCtx.globalAlpha && typeof alpha == 'number') {
             txtrCtx.globalAlpha = alpha;
         }
--- a/src/Core/Style.js
+++ b/src/Core/Style.js
@@ -905,13 +905,13 @@ class Style {
                         }
                     }
                     style.icon.source = sprites.source;
-                    style.icon.size = readVectorProperty(layer.layout['icon-size']) || 1;
+                    style.icon.size = readVectorProperty(layer.layout['icon-size']) ?? 1;
                     const { color, opacity } = rgba2rgb(readVectorProperty(layer.paint['icon-color'], { type: 'color' }));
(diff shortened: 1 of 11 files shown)


examples/vector_tile_raster_2d_mapbox.html
```diff
@@ -0,0 +1,98 @@
+<html>
+  <head>
+    <title>Itowns - Globe</title>
```
jailln: Can you update the title please?

examples/vector_tile_raster_2d_mapbox.html
```diff
@@ -0,0 +1,98 @@
...
+        });
+      }
+ 
+      const debugMenu = new GuiTools("menuDiv", view);
+      setupLoadingScreen(viewerDiv, view);
+
+      const source = new itowns.VectorTilesSource({
+        style: "mapbox://styles/mapbox/streets-v8",
+        accessToken:
+          "pk.eyJ1IjoiZ2VvdXJmbyIsImEiOiJjbHdyZWVrd2owOW1rMmlxdmx2Z3Fwb2JhIn0.sg5en26yZ6aSEg6CsfUByA",
```
jailln: is that your personal access token?

examples/vector_tile_raster_2d_mapbox.html
```diff
@@ -0,0 +1,98 @@
...
+    <link
+      rel="stylesheet"
+      type="text/css"
+      href="css/LoadingScreen.css"
+    />
+    <link
+      rel="stylesheet"
+      type="text/css"
+      href="https://www.itowns-project.org/itowns/examples/css/widgets.css"
+    />
```
jailln: we don't need widgets in this example

examples/vector_tile_raster_2d_mapbox.html
```diff
@@ -0,0 +1,98 @@
...
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
+  </head>
+  <body>
+    <div id="viewerDiv"></div>
+
+    <!-- Import iTowns source code -->
+    <script src="../dist/itowns.js"></script>
+    <script src="../dist/debug.js"></script>
+    <!-- Import iTowns Widgets plugin -->
+    <script src="../dist/itowns_widgets.js"></script>
```
jailln: we don't need widgets in this example

src/Source/Source.js
```diff
@@ -108,6 +108,7 @@ class Source extends InformationsData {
     constructor(source) {
         super(source);
         this.isSource = true;
+        this.warn = new Set();
```
jailln: Aggregating logs is a good idea and needed in itowns. I think we should  not store aggregated logs in objects and pass them around as function parameters. An alternative implementation would be to implement a LogAggregator that stores logs, aggregates them and can log info, warn and errors. A global instance could be used by all itowns components then. If we do so, we need to be careful with memory usage and empty the log cache frequently. And if we don't want to implement it ourselves, I'm sure there are libraries available. However I think that's out of scope of this PR, so I'd be for removing this log aggregation for now and implement it differently later. @Desplandis what is your opinion on this?

examples/vector_tile_raster_2d_mapbox.html
```diff
@@ -0,0 +1,98 @@
+<html>
+  <head>
```
jailln: can you add it to the examples list please?

src/Source/VectorTilesSource.js
```diff
@@ -71,38 +71,47 @@ class VectorTilesSource extends TMSSource {
...
+            return mvtStyle;
+        }).then((mvtStyle) => {
+            mvtStyle.layers.forEach((layer, order) => {
                 layer.sourceUid = this.uid;
                 if (layer.type === 'background') {
                     this.backgroundLayer = layer;
                 } else if (ffilter(layer)) {
-                    const style = Style.setFromVectorTileLayer(layer, this.sprites, order, this.symbolToCircle);
+                    if (layer['source-layer'] === undefined) {
+                        const refProperties = ['type', 'source', 'source-layer', 'minzoom', 'maxzoom', 'filter', 'layout'];
```
jailln: I think we should start reducing the cognitive complexity of this constructor by cutting it in small logical pieces. Can you put this logic in a dedicated private method that could be named something like `parseRefLayer` please ?

src/Source/VectorTilesSource.js
```diff
@@ -71,38 +71,47 @@ class VectorTilesSource extends TMSSource {
...
+                        const refLayer = mvtStyle.layers.filter(l => l.id === layer.ref)[0];
```
jailln: Is this list final ? Where does it come from ? Could we look for all properties that are in the ref layer and not in the current layer instead to be more generic?

src/Source/VectorTilesSource.js
```diff
@@ -112,20 +121,18 @@ class VectorTilesSource extends TMSSource {
...
-                            max: layer.maxzoom || 24,
-                        },
                     });
                 }
             });
 
             if (this.url == '.') {
-                const TMSUrlList = Object.values(style.sources).map((sourceVT) => {
+                const TMSUrlList = Object.values(mvtStyle.sources).map((sourceVT) => {
                     if (sourceVT.url) {
```
jailln, line 136: these default values are not needed anymore ?

src/Core/Style.js
```diff
@@ -919,17 +958,16 @@ class Style {
...
         if (this.stroke) {
             // TO DO add possibility of using a pattern (https://github.com/iTowns/itowns/issues/2210)
-            this._applyStrokeToPolygon(txtrCtx, invCtxScale, polygon, context);
+            this._applyStrokeToPolygon(txtrCtx, invCtxScale, polygon);
         }
 
         // fill inside of polygon
-        if (canBeFilled && this.fill) {
+        if (canBeFilled && (this.fill.pattern || this.fill.color)) {
             // canBeFilled can be move to StyleContext in the later PR
```
jailln, line 948: Could we remove canBeFilled ? Its value is feature.type == FEATURE_TYPES.POLYGON . We are in a method named `applyToCanvasPolygon` so aren't already expecting to handle only polygons ?

src/Core/Style.js
```diff
@@ -919,17 +958,16 @@ class Style {
      * @param {Boolean} canBeFilled - true if feature.type == FEATURE_TYPES.POLYGON.
      */
     applyToCanvasPolygon(txtrCtx, polygon, invCtxScale, canBeFilled) {
-        const context = this.context;
         // draw line or edge of polygon
         if (this.stroke) {
             // TO DO add possibility of using a pattern (https://github.com/iTowns/itowns/issues/2210)
```
jailln, line 942: should we check for stroke properties too, like for fill ?

src/Core/Style.js
```diff
@@ -820,6 +826,8 @@ class Style {
             style.point.opacity = opacity;
             style.point.radius = readVectorProperty(layer.paint['circle-radius']);
         } else if (layer.type === 'symbol') {
+            // if symbol we shouldn't draw stroke but defaut value is 1.
+            style.stroke.width = 0.0;
```
jailln, line 819: no need for this comment I think, the code is clear enough

src/Renderer/Label2DRenderer.js
```diff
@@ -186,13 +186,15 @@ class Label2DRenderer {
...
-            label.horizonCullingPoint && GlobeLayer.horizonCulling(label.horizonCullingPoint) ||
-        // Check if content isn't present in visible labels
-            this.grid.visible.some((l) => {
-                // TODO for icon without text filter by position
-                const textContent = label.content.textContent;
-                return textContent !== '' && l.content.textContent.toLowerCase() == textContent.toLowerCase();
-            })) {
+            label.horizonCullingPoint && GlobeLayer.horizonCulling(label.horizonCullingPoint)
+            // Why do we might need this part ?
+            // || // Check if content isn't present in visible labels
```
jailln, line 191: I think it was to avoid overlapping but I'm not sure we need it too so let's remove it completely in this case. We still have git if we need it later. @Desplandis WDYT?

src/Parser/VectorTileParser.js
```diff
@@ -130,41 +131,33 @@ function readPBF(file, options) {
...
-                if (!feature) {
-                    feature = collection.requestFeatureById(layer.id, vtFeature.type - 1);
-                    feature.id = layer.id;
-                    feature.order = layer.order;
-                    feature.style = options.in.styles[feature.id];
-                    vtFeatureToFeatureGeometry(vtFeature, feature);
-                } else if (!collection.features.find(f => f.id === layer.id)) {
-                    feature = collection.newFeatureByReference(feature);
-                    feature.id = layer.id;
-                    feature.order = layer.order;
```
jailln: This seems like some kind of optimization, can you explain in more details why you remove it please?

src/Core/Style.js
```diff
@@ -464,8 +465,6 @@ function _addIcon(icon, domElement, opt) {
  * The first parameter of functions used to set `Style` properties is always an object containing
  * the properties of the features displayed with the current `Style` instance.
  *
- * @property {Number} order - Order of the features that will be associated to
- * the style. It can helps sorting and prioritizing features if needed.
  * @property {Object} fill - Polygons and fillings style.
  * @property {String|Function|THREE.Color} fill.color - Defines the main color of the filling. Can be
  * any [valid color
```
jailln, line 470: can you provide more details about why you remove it please?

jailln (changes requested):
Thanks for the debug and the fixes :)

src/Source/Source.js
Desplandis: Yeah, we could have some kind of heap with a max size to aggregate those warnings. I think there should be some kind of library to do it. However, I agree that this is out of scope of this PR.

src/Renderer/Label2DRenderer.js
Desplandis: If we can't reproduce the overlapping and it seems to be the case here. I think we could totally remove this.

examples/vector_tile_raster_2d_mapbox.html
ftoromanoff: Done. (I titled it 'vector_tile_mapbox_raster')

examples/vector_tile_raster_2d_mapbox.html
ftoromanoff: I changed it to the one I created for itowns example (already used in 'vector_tile_3d_mesh_mapbox')

examples/vector_tile_raster_2d_mapbox.html
ftoromanoff: Removed.

examples/vector_tile_raster_2d_mapbox.html
ftoromanoff: Removed.

examples/vector_tile_raster_2d_mapbox.html
ftoromanoff: Done.

src/Source/VectorTilesSource.js
ftoromanoff: I moved the code in a function 'getPropertiesFromRefLayer()' and added an explaination to it.

src/Source/VectorTilesSource.js
ftoromanoff: From what I managed to find the list is final. In fact using ref for a style layer is deprecated and MapBox removed all trace of how it was used. But there is still data using it (as their own stream that we use in the new example).

src/Source/VectorTilesSource.js
ftoromanoff: I don't really understand your question. I only renamed the variable to avoid confusion with itowns.style and itowns.layer.

src/Core/Style.js
ftoromanoff: In fact `applyToCanvasPolygon` is called in the function `drawPolygon()` in Feature2Texture.js and that function is called on any feature.type that isn't a Point. Thus is include polygon and line, the `canBeFilled` is then the way to distinguish both.
It's the name that is maybe to rethink at some point..

src/Core/Style.js
ftoromanoff: `this.stroke` and `this.fill` are never undefined as the both have properties with default value. The previously done check were then useless.
In the end state of this PR both thes were removed for more specific tests.

src/Core/Style.js
ftoromanoff: A bit to much commentary is too bad ?
I can removed it, I wanted to be sure as up to now the stroke (of a polygon or line) were draw even with a width = 0.

src/Renderer/Label2DRenderer.js
ftoromanoff: As for now, it seems that it is not needed on what I tested and I asked @mgermerie why he used it in the first place.

src/Parser/VectorTileParser.js
ftoromanoff: I removed it to simplified. It is also linked to a trouble to link a specifique feature.style to it's properties.

With this optimization I got these error:
image
linked with https://github.com/iTowns/itowns/blob/18cc3eeda23d557317a8756e4fedef099d9dc95a/src/Core/Style.js#L903-L914
The properties of the current featreu doesn't seems to be the right one when the function is called...

That I don't get when each feature as it's own existence. It might probalby be solved other wise to avoid the probable border effect induce by `newFeatureByReference`...

src/Core/Style.js
ftoromanoff: Currently the style.order was not at all taking into account partialy because it was not called thought getter and setter as all the other properties.
And I also noticed that this properties is only used in VectorTileSource. That's the reason I removed it from style to leave it to the internal use of VectorTile and ordering the different layers inside a same VectorTile. If we would like an ordering of feature inside a same 'layer' we will think at that moment how to put it back into Style.

src/Source/Source.js
jailln: @ftoromanoff can you remove this part from this PR then please ? You can open an issue on this matter if you want to keep track of it

src/Source/VectorTilesSource.js
jailln: thanks for moving it to a dedicated method and commenting it :+1:

src/Source/VectorTilesSource.js
jailln: Sorry, it was meant to be above on lines 115-118 (regarding the default zoom values)

src/Core/Style.js
jailln: Ah ok thanks, we should rename applyToCanvasPolygon then, I agree. Let's do it in another PR

src/Parser/VectorTileParser.js
jailln: It's a bit weird that you get this warning since we look for features with the same id that would already have been parsed.
If we remove this code I'm not sure we need to features collection at all since from my understanding it is to avoid creating multiple times the same feature in memory. It would be interresting to check if it is common that same features are in different pbf and therefore what if this optimization is really needed. It seems a bit unrelated to this PR though so what I propose is to:
* keep this code but comment it for now
* add a comment above explaining what we've said here: this optimization is not fully working and needs to be re-assessed. It has been deactivated for now, see: https://github.com/iTowns/itowns/pull/2469/files#r1861802136

src/Source/VectorTilesSource.js
ftoromanoff: I removed these 'default' values here because the min/max value should be a style properties and not at all a parsing.
So i removed these from here and added the use of the style.zoom values in the relative files as it should have been from the start.

https://github.com/iTowns/itowns/blob/c41e18a26eca71651f6f0d46d62a23c23ce0b6c0/src/Converter/Feature2Texture.js#L73-L75

src/Source/Source.js
ftoromanoff: Done

src/Parser/VectorTileParser.js
ftoromanoff: Done

jailln (approved):
Ready to merge! Thanks @ftoromanoff

PR #2477, opened 2024-12-17
refactor(LayeredMaterial): migrate to TypeScript

HoloTheDrunk:
## Description

Converted the `LayeredMaterial` class and eponymous file to TypeScript.
Took this opportunity to do some internal cleanup and refactoring of things like the way the layers (or more precisely `RasterTile`s) were stored, the uniforms and defines, etc... This means the internal API for the material has changed since we now keep `RasterColorTile`s and the one `RasterElevationTile` separate internally. A convenience `getLayer` method has  been provided matching the previous API for cases where the exact type of `RasterTile` is not important.
Whether we should rename the "layer" fields to "tile" to match the actual data type is up for debate. I'm all for it, but I'm not aware of how dependent people might be on those names.

Also planning on making the new uniforms/defines convenience TS-specific functions available to the `PointsMaterial` class and nuking `src/Renderer/CommonMaterial.js` out of existence in a coming PR.

## Motivation and Context
This PR fits into the larger motion to migrate iTowns to TypeScript.

Tested on Ubuntu 22.04.5 LTS x86_64 using Firefox 126.0.1 (64-bit)


jailln:
Thanks for this PR :) It really makes this material clearer!

I would be down for renaming layer to tile since using layer is really confusing with itowns layers and I think no one is using this API outside of itowns.

I noticed a few bugs though:
* terrain is not displayed anymore: see view_3d_map.html for instance. Also, it seems that now you only let the possibility to have one elevation layer. Even if multiple elevation layers support was not totally implemented (especially when having tiles at the same zoom level or because sometimes we used the first elevation layer by default), it was possible to have two elevation layers.
* debug tiles are always displayed:
image
* zoom is broken:

https://github.com/user-attachments/assets/93ac44e8-7aef-4c5b-a423-38d115fd82b3

HoloTheDrunk:
Hi, yes this is still a WIP and I'm tracking down the source of these issues.
I'll be on holiday until early January though so this isn't gonna move much until then.
Happy holidays!
Update: back from holidays, working on this again

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        maxSamplersColorCount,
+    );
+}
+
+export const colorLayerEffects: Record<string, number> = {
+    noEffect: 0,
+    removeLightColor: 1,
+    removeWhiteColor: 2,
+    customEffect: 3,
+};
```
Desplandis: This is an enum for effects (i.e. for adding custom shader chunks in `Tile.vs`/`Tile.fs`) in `ColorLayer`, I think this should be `as const`.

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        transparent: false,
+    },
+    elevation: {
+        scale: 0,
+        bias: 0,
+        mode: 0,
+        zmin: 0,
+        zmax: 0,
+    },
+};
```
Desplandis: So the `noDataValue` property of the previous default structure was not used?

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        );
+    }
+    uTextureCount.value = count;
+}
+
+export const ELEVATION_MODES = {
+    RGBA: 0,
+    COLOR: 1,
+    DATA: 2,
+};
```
Desplandis: This is an enum and it shall be `as const`.

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        fillInProp(defines, 'NUM_FS_TEXTURES', nbSamplers[1]);
+        // TODO: We do not use the fog from the scene, is this a desired
+        // behavior?
+        fillInProp(defines, 'USE_FOG', 1);
+        fillInProp(defines, 'NUM_CRS', crsCount);
+
+        initModeDefines(defines);
+        fillInProp(defines, 'MODE', RenderMode.MODES.FINAL);
+
+        // @ts-expect-error: global constexpr
```
Desplandis: When you'll rebase on `master`, I suggest adding a `global` declaration in `Main.ts` of the `itowns` package:
```ts
declare global {
  const __DEBUG__: boolean;
}
```

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+    colorTextureCount: number,
+}
+
+let nbSamplers: [number, number] | undefined;
+const fragmentShader: string[] = [];
+
+/** Replacing the default uniforms dynamic type with our own static map. */
+export type LayeredMaterialParameters =
+    Omit<THREE.ShaderMaterialParameters, 'uniforms'>
+    & { uniforms?: MappedUniforms<LayeredMaterialRawUniforms> };
```
Desplandis: I think it would be easier to read as an extended interface:

```suggestion
export interface LayeredMaterialParameters extends THREE.ShaderMaterialParameters {
    uniforms?: MappedUniforms<LayeredMaterialRawUniforms>;
}
```

Also, do we authorize user of the `LayeredMaterial` class to set its own defines/uniforms?

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+    const uTextureCount = uniforms.textureCount;
+
+    // Flatten the 2d array: [i, j] -> layers[_layerIds[i]].textures[j]
+    let count = 0;
+    for (const tile of tiles) {
+        // FIXME: RasterElevationTile are always passed to this function alone
+        // so this works, but it's really not great even ignoring the dynamic
+        // addition of a field.
+        // @ts-expect-error: adding field to passed layer
+        tile.textureOffset = count;
```
Desplandis: Yeah, kinda ugly... I don't see where it is passed to the GPU though.

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+
+    // Debug
+    showOutline: boolean,
+    outlineWidth: number,
+    outlineColors: THREE.Color[]
+
+    // Elevation layers
+    elevationLayers: Array<StructElevationLayer>,
+    elevationTextures: Array<THREE.Texture>,
+    elevationOffsetScales: Array<THREE.Vector4>,
```
Desplandis: I have a coding-style question (which is kinda orthogonal to this PR), shall we go for `T[]` or `Array<T>` for consistency?

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        // TODO: We do not use the fog from the scene, is this a desired
+        // behavior?
+        fillInProp(defines, 'USE_FOG', 1);
+        fillInProp(defines, 'NUM_CRS', crsCount);
+
+        initModeDefines(defines);
+        fillInProp(defines, 'MODE', RenderMode.MODES.FINAL);
+
+        // @ts-expect-error: global constexpr
+        fillInProp(defines, 'DEBUG', +__DEBUG__);
```
Desplandis: I think it would it be better to just assign the properties directly into `defines`. This would remove the need for type assertions and ease the readability of the code. Something like
```typescript
        const defines: typeof this.defines = {
            NUM_VS_TEXTURES: nbSamplers[0],
            NUM_FS_TEXTURES: nbSamplers[1],
            USE_FOG: 1,
            NUM_CRS: crsCount,
            initModeDefines(),
            MODE: RenderMode.MODES.FINAL,
            DEBUG: __DEBUG__,
        };
        this.defines = defines;
```


You would need to change the `initNodeDefines` to output the key-mapped object for both `ELEVATION_MODES` and `RenderModes.MODES`. Such mapping could be done with
```ts
Object.fromEntries(
        Object.entries(ELEVATION_MODES).map(([key, value]) => [`ELEVATION_${key}`, value])
    ) as { [K in keyof typeof ELEVATION_MODES as `ELEVATION_${K}`]: typeof ELEVATION_MODES[K] }
```

P.S.: I didn't find a way to do such mapping without explicit casts
P.P.S: Forgot to use `DefineMapping` in my suggestion

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+
+    public colorTiles: RasterColorTile[];
+    public elevationTile: RasterElevationTile | undefined;
+
+    public colorTileIds: string[];
+    public elevationTileId: string | undefined;
+
+    public layersNeedUpdate: boolean;
+
+    public override defines: LayeredMaterialDefines;
```
Desplandis: Why not adding an override for `uniforms` too?

Something like `MappedUniforms<LayeredMaterialRawUniforms>`? `uniforms` would be correctly typed in the whole module.

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+
+    constructor(options: LayeredMaterialParameters = {}, crsCount: number) {
+        super(options);
```
Desplandis: If we wish to let the user provides its own uniforms, it would be better to assign them directly in this class.
```suggestion
        const { uniforms, ...materialOptions } = options;
        super(materialOptions);
```

This would remove the need (in other part of the code) to check if the uniform is already assigned.

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        fillInProp(defines, 'DEBUG', +__DEBUG__);
+
+        // @ts-expect-error: global constexpr
+        if (__DEBUG__) {
+            const outlineColors = [new THREE.Color(1.0, 0.0, 0.0)];
+            if (crsCount > 1) {
+                outlineColors.push(new THREE.Color(1.0, 0.5, 0.0));
+            }
+
+            this.initUniforms({
```
Desplandis: In the same vein as my suggestion for `defines`, why not have a single call to `initUniforms` which could return all uniforms and assign the `uniforms` property directly?

src/Renderer/LayeredMaterial.ts
```diff
@@ -0,0 +1,542 @@
...
+        for (const [name, value] of Object.entries(uniforms)) {
+            const uniform = this.uniforms[name];
+            if (uniform === undefined) {
+                continue;
+            }
+            if (uniform.value !== value) {
+                uniform.value = value;
+            }
+        }
+    }
```
Desplandis: Any use?

Desplandis (changes requested):
Good work! I made some suggestions while debugging/reviewing the code.

Desplandis:
@HoloTheDrunk You could fix the zoom and picking bug by correctly assigning the `objectId` uniform in the constructor of `TileMesh`.

```js
material.setUniform('objectId', this.id);
// instead of
material.objectId = this.id;
```

src/Renderer/LayeredMaterial.ts
HoloTheDrunk: I don't see a reason to allow custom defines/uniforms given that the pipeline used is fixed (unless the user goes and modifies the shaders, which would be as unintended as can be).

src/Renderer/LayeredMaterial.ts
HoloTheDrunk: It's part of the expected fields for a `StructColorLayer`. Seen in `src/Renderer/Shader/Chunk/color_layers_pars_fragment.glsl`.

src/Renderer/LayeredMaterial.ts
HoloTheDrunk: I personally prefer the `Array<T>` notation because it's standard with the rest of the type system, but given the ubiquity of arrays it might be better to use `T[]`? I don't really have strong feelings on the matter.

src/Renderer/LayeredMaterial.ts
HoloTheDrunk: This is provided for users to be able to set uniform values. Should I remove `setUniform` and only keep `setUniforms` or keep both (and use `setUniform` in `setUniforms`)?

src/Renderer/LayeredMaterial.ts
HoloTheDrunk: Unused in the shaders and therefore unnecessary in the uniform structs.

src/Renderer/LayeredMaterial.ts
Desplandis: Yeah, but were is it copied?

src/Renderer/LayeredMaterial.ts
Desplandis: In that case, you should maybe omit both `uniforms` and `defines` properties.

src/Renderer/LayeredMaterial.ts
Desplandis: I usually use `T[]` when T is a single word (or for primitive types) and reserve Array for more complex types.

src/Renderer/LayeredMaterial.ts
Desplandis: Okay, we'll resolve it later when we'll document the shaders.

Desplandis (approved):

PR #2482, opened 2025-01-08
fix(test): fix local unit tests behind proxy

mgermerie:
## Description

Use undici to prevent unit tests failure when fetching behind proxy. See https://github.com/nodejs/undici/discussions/2167.


Desplandis (approved):
Tested on IGN's proxy, it works!

PR #2491, opened 2025-01-23
fix(TileBuilder): use cached buffers correctly

### LINKED ISSUE #2486: Wrong terrain tiles
## Context
Terrain tiles have been refactored in #2440. Since then, they have a point with altitude 0.
Also, it seems that the skirt is no longer working.

## Steps to Reproduce
Open any globe example and toggle wireframe mode in the debug menu

## Expected Behavior
Before the PR, without terrain:

Image

and with terrain:

Image

## Actual Behavior

Currently on master, without terrain:

Image

and with terrain:

Image

The broken skirt:

Image

HoloTheDrunk:
## Description

Replace `params.buildIndexAndUv_0` by optionally passing the cached buffers to the `computeBuffers` function.

## Motivation and Context

This cleans up the tile building code a bit further and allows us to easily fix the rendering of cached tiles' skirts.

Resolves #2486

### DIFF
--- a/src/Core/Prefab/TileBuilder.ts
+++ b/src/Core/Prefab/TileBuilder.ts
@@ -103,12 +103,10 @@ export function newTileGeometry(
             buffers = computeBuffers(
                 builder,
                 params,
-                // @ts-expect-error: can't sync index type right now
                 cachedBuffers !== undefined
                     ? {
-                        // We lost type information along the way with the
-                        // storage as THREE.BufferAttribute.
-                        index: cachedBuffers.index.array,
+                        index: cachedBuffers.index.array as
+                            Uint8Array | Uint16Array | Uint32Array,
                         uv: cachedBuffers.uv.array as Float32Array,
                     }
                     : undefined,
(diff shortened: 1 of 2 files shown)


src/Core/Prefab/TileBuilder.ts
```diff
@@ -99,21 +97,35 @@ export function newTileGeometry(
...
-            buffers = computeBuffers(builder, params);
+            buffers = computeBuffers(
+                builder,
+                params,
+                // @ts-expect-error: can't sync index type right now
+                cachedBuffers !== undefined
+                    ? {
+                        // We lost type information along the way with the
+                        // storage as THREE.BufferAttribute.
+                        index: cachedBuffers.index.array,
```
Desplandis: Why not just cast `index` to `Uint8Array | Uint16Array | Uint32Array`?

jailln (approved):
well done :clap:

src/Core/Prefab/TileBuilder.ts
HoloTheDrunk: Hadn't thought of it in the moment for some reason, ty

PR #2493, opened 2025-01-28
feat(3d-tiles): support picking of metadata

Desplandis:
## Description
This PR add support for picking metadata from the `EXT_mesh_structural_metadata` GLTF extension (optionally using the `EXT_mesh_features` for getting the feature).

## Motivation and Context
We are testing the production of complex 3D buildings datasets at IGN from [BD Topo/Uni](https://espacecollaboratif.ign.fr/data/about). The pipeline is currently BD Topo -> CityJSON -> 3D Tiles. We are using [tyler](https://github.com/3DGI/tyler) to output 3D Tiles from CityJSON which uses both extensions to store feature data.

When the pipeline will be completed , I will publish some datasets in our test base (in another PR).

### DIFF
--- a/src/Layer/OGC3DTilesLayer.js
+++ b/src/Layer/OGC3DTilesLayer.js
@@ -144,6 +144,90 @@ export function enableMeshoptDecoder(MeshOptDecoder) {
     itownsGLTFLoader.setMeshoptDecoder(MeshOptDecoder);
 }
 
+async function getMeshFeatures(meshFeatures, options) {
+    const { faceIndex, barycoord } = options;
+
+    const features = await meshFeatures.getFeaturesAsync(faceIndex, barycoord);
+    return {
+        features,
+        featureIds: meshFeatures.getFeatureInfo(),
+    };
+}
+
+function getStructuralMetadata(structuralMetadata, options) {
+    const { index, faceIndex, barycoord, tableIndices, features } = options;
+
+    const tableData = [];
+    if (tableIndices !== undefined && features !== undefined) {
+        structuralMetadata.getPropertyTableData(
+            tableIndices,
+            features,
+            tableData,
+        );
+    }
+
+    const attributeData = [];
+    if (index !== undefined) {
+        structuralMetadata.getPropertyAttributeData(index, attributeData);
+    }
+
+    const textureData = [];
+    if (faceIndex !== undefined) {
+        structuralMetadata.getPropertyTextureData(
+            faceIndex,
+            barycoord,
+            textureData,
+        );
+    }
+
+    const metadata = [
+        ...tableData,
+        ...textureData,
+        ...attributeData,
+    ];
+
+    return metadata;
+}
+
+async function getMetadataFromIntersection(intersection) {
+    const { point, object, face, faceIndex } = intersection;
+    const { meshFeatures, structuralMetadata } = object.userData;
+
+    const barycoord = new THREE.Vector3();
+    if (face) {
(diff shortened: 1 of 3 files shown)


examples/3dtiles_loader.html
```diff
@@ -129,7 +129,7 @@
             function addElevationLayerFromConfig(config) {
                 config.source = new itowns.WMTSSource(config.source);
                 var elevationLayer = new itowns.ElevationLayer(config.id, config);
-                view.addLayer(elevationLayer);
+                // view.addLayer(elevationLayer);
```
jailln: to uncomment

src/Layer/OGC3DTilesLayer.js
```diff
@@ -194,6 +307,11 @@ class OGC3DTilesLayer extends GeometryLayer {
         this.pntsMaxAttenuatedSize = pntsMaxAttenuatedSize;
 
         this.tilesRenderer = new TilesRenderer(this.source.url);
+        /** @type {RequestInit} */
+        // const fetchOptions = {
+        //     mode: 'no-cors',
+        // };
+        // this.tilesRenderer.fetchOptions = fetchOptions;
```
jailln: to remove

src/Layer/OGC3DTilesLayer.js
```diff
@@ -368,36 +487,31 @@ class OGC3DTilesLayer extends GeometryLayer {
         this.tilesRenderer.dispose();
     }
 
+    /**
+     * Get the [metadata](https://github.com/CesiumGS/3d-tiles/tree/main/specification/Metadata)
+     * of the closest intersected object from a list of intersections. Note that
+     * it only supports metadata from the [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata)
+     * gltf extension. It also optionnaly uses features from the [](`https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features).
```
jailln: The phrasing make it look like you have to set an option to get the `EXT_mesh_features` metadata which is not the case. May be we should rephrase or, better, explain when the `EXT_mesh_features` is retrieved or not? Also, the alt text is missing in the link to EXT_mesh_features

jailln (commented):
Thanks for this PR :) 

Regarding the public API, I'd rather expose only one method to get all metadata of a given feature (from the batch table and from the extensions) and therefore merge `getMetadataFromIntersections` and `getC3DTileFeatureFromIntersectsArray` into `getMetadataFromIntersections`.
I'm also wondering if we should expose a method to directly retrieve metadata from a mouse event (that will call `layer.pickObjectsAt` and `layer.getMetadataFromIntersections`). WDYT ?

examples/3dtiles_loader.html
Desplandis: Done

src/Layer/OGC3DTilesLayer.js
Desplandis: Done

src/Layer/OGC3DTilesLayer.js
Desplandis: Yeah, rereading now it it is not clear at all. xD Will rewrite it!

Desplandis:
@jailln All good!

jailln (approved):

Desplandis:
@jailln I added some unit tests for this new feature. =)

PR #2494, opened 2025-01-30
Fix: PointCloud Layer use  user elevation range value if present

ftoromanoff:


AnthonyGlt (approved):
Thank you for this MR

Issue #2473, opened 2024-12-04
The webxr example is broken when enabling VR


AnthonyGlt:
Since the bump of the three.js version f7303dec2bdb2a6bb9d35d1fdeb4d220eb42d4f6
The [view_3d_map_webxr example](https://www.itowns-project.org/itowns/dev/examples/index.html#view_3d_map_webxr) doesn't work anymore when entering the VR mode

[Screencast from 04-12-2024 17:30:28.webm](https://github.com/user-attachments/assets/2c1fd16c-a45e-4d5f-ad88-5fc0a2c80865)

## Steps to Reproduce
1. Run the view_3d_map_webxr example
2. Click on Enter VR
3. Blank screen

## Possible Cause/Fix/Solution
Since [this merge](https://github.com/mrdoob/three.js/pull/28533) in three.js, there is a condition on [matrixWorldAutoUpdate](https://github.com/mrdoob/three.js/pull/28533/files#diff-3f051235c898fd468fd06c80b0bc8f8a9f1258561313247a51bbb0cae81f93feR592) .

To fix it, we should somehow set 
`view.scene.matrixWorldAutoUpdate = true;`
either in the file view_3d_map_webxr.html 
or in WebXR.js, using this kind of [workaround](https://github.com/mrdoob/three.js/pull/28533#issuecomment-2451950811)
So in [WebXR.js ](https://github.com/iTowns/itowns/blob/master/src/Renderer/WebXR.js), above these lines

https://github.com/iTowns/itowns/blob/683e55d51e610e106749b78d8a3ef83158532d2a/src/Renderer/WebXR.js#L55-L57

we could add

```js 
                const prevMatrixWorldAutoUpdate = view.scene.matrixWorldAutoUpdate;
                view.scene.matrixWorldAutoUpdate = true;
                view.scene.updateMatrix();
                view.scene.updateMatrixWorld();
                view.scene.matrixWorldAutoUpdate = prevMatrixWorldAutoUpdate;

```

jailln:
I wonder how there aren't more things broken as I read the comments from https://github.com/mrdoob/three.js/pull/28533.

In my opinion, we should harmonize the xr and non-xr behaviour: either set autoupdate to false in both cases and manage matrices updates ourselves, or set autoupdate to true for both context.

I haven't looked into it yet but first we need to remember why we set the autoupdate to false and if there's really a performance gain or if we'd be better off setting it to true and avoiding the hassle and bugs.

@Desplandis wdyt?

[...]

PR #2496, opened 2025-02-06
Bump 3d tiles renderer to 0.4.4 and three.js to 0.170.0

Desplandis:
## Description

This PR bumps `3d-tiles-renderer` from 0.3.39 to 0.4.4 and `three.js` from 0.168.0 to 0.170.0.

### r169
See [changelog](https://github.com/mrdoob/three.js/releases/tag/r169) and [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide#168--169).

There is no changes impacting iTowns.

A list of interesting changes (feel free to extend it):
- Raycaster
  + Add barycoord to intersection result, make attribution interpolation more convenient. https://github.com/mrdoob/three.js/pull/29340 (@gkjohnson). In #2493, we manually calculate the barycoord from the intersection result, we may not need it anymore.
- WebGLRenderer
  + Merge update ranges before issuing updates to the GPU. https://github.com/mrdoob/three.js/pull/29189 (@HunterLarco)
  + Fix feedback loop with RTT read pixels async. https://github.com/mrdoob/three.js/pull/29320 (@gkjohnson). For picking metadata in iTowns.
  + Reduce readRenderTargetPixelsAsync() latency. https://github.com/mrdoob/three.js/pull/29339 (@aardgoose). For picking metadata in iTowns.

### r170
See [changelog](https://github.com/mrdoob/three.js/releases/tag/r170) and [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide#169--170).

A list of interesting changes (feel free to extend it):
- BatchedMesh: different changes used internally by `3d-tiles-renderer` (see BatchedMeshPlugin below)
- WebGLTextures
  + Ensure depth texture properties are deleted. https://github.com/mrdoob/three.js/pull/29697 (@Mugen87) 

## Changes in `3d-tiles-renderer`
There is some changes impacting iTowns, especially for plugins. I disabled some eslint check on imports since it does not support package.json imports... (see https://github.com/import-js/eslint-plugin-import/issues/1868). It seems `eslint-import-resolver-typescript` works but decided to go for disabling this warning for now.

### Added
- `load-error` events when model, tile set, and API token requests fail to fetch or parse. To use internally in iTowns and provide a way to the user to access this event.
- PNTSLoader: Add support for normals, quantized normals.
- TilesRenderer: Support for 3DTILES_ELLIPSOID extension.
- TilesRenderer: added "loadProgress" field.
- BatchedMeshPlugin plugin: https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/628. I think we should examine the performance gains in using BatchedMesh for 3D Tiles rendering.

### DIFF
--- a/config/threeExamples.mjs
+++ b/config/threeExamples.mjs
@@ -5,10 +5,11 @@ export default {
         './loaders/DRACOLoader.js',
         './loaders/KTX2Loader.js',
         './loaders/DDSLoader.js',
+        './math/ColorSpaces.js',
         './utils/BufferGeometryUtils.js',
         './utils/WorkerPool.js',
         './capabilities/WebGL.js',
         './libs/ktx-parse.module.js',
-        './libs/zstddec.module.js'
+        './libs/zstddec.module.js',
     ],
 };
--- a/examples/3dtiles_loader.html
+++ b/examples/3dtiles_loader.html
@@ -50,8 +50,8 @@
         <script type="importmap">
             {
                 "imports": {
-                    "three": "https://cdn.jsdelivr.net/npm/three@0.165.0/build/three.module.js",
-                    "three/addons/": "https://cdn.jsdelivr.net/npm/three@0.165.0/examples/jsm/"
+                    "three": "https://cdn.jsdelivr.net/npm/three@0.170.0/build/three.module.js",
+                    "three/addons/": "https://cdn.jsdelivr.net/npm/three@0.170.0/examples/jsm/"
                 }
             }
         </script>
--- a/examples/effects_stereo.html
+++ b/examples/effects_stereo.html
@@ -23,8 +23,8 @@
         <script type="importmap">
             {
                 "imports": {
-                    "three": "https://cdn.jsdelivr.net/npm/three@0.159.0/build/three.module.js",
-                    "three/addons/": "https://cdn.jsdelivr.net/npm/three@0.159.0/examples/jsm/"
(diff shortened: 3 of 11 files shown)


AnthonyGlt:
There is a breaking change for the WEBXR example, probably this one
https://github.com/iTowns/itowns/issues/2473

BUT I'm currently working on a refactoring of the webxr.

Should we put it in deprecated for now ?

AnthonyGlt (approved):

Desplandis:
@AnthonyGlt Depends on whether or not we considered the WebXR example as ready to production or not.

AnthonyGlt:
There is already a webxr example available (view_3d_map_webxr) , I'm just saying that with this bump, it wouldn't work anymore.
For now, we should, at least, remove the access to this example from the [config.json](https://github.com/iTowns/itowns/blob/387d64f0cf796734afd0d99d88b2bb110c76a419/examples/config.json#L6), until incoming refactoring. I don't think that we should hot fix it because of this refactoring.
IF we rather hot fix it, I've gave a solution in the issue mentioned above.
WDYT ?

Either way, some changes should be made in [this file](https://github.com/iTowns/itowns/blob/master/src/Renderer/WebXR.js)
to fix it or to link to the issue, so users will know that it's temporarily broken and how to bypass the bug

Desplandis:
Okay, I added the quickfix in the example as suggested in your issue, that should be good enough. =)

PR #2498, opened 2025-02-10
PointCloudLayer: geographic reprojection for point clouds on load

ftoromanoff:
The aim of the PR is to allow projection of dataPointCloud in a view crs different than the source crs.
For the projection we use proj4.

There is two main parts in this PR:
- the projection of the bounding box read from metadonne in the layer.
- the projection of the data during the parsing.

Instead of doing pure projection we add after the projection, an transformation of data in a local system orienté in the direction of the data and z orthogonal.

Other minor change:
- ~~the 'spacing' is move from source to the layer.~~
- ~~A 'elevation' propertie was also added to replace the z data (that is not an elevation)~~
- update examples

### DIFF
--- a/examples/copc_3d_loader.html
+++ b/examples/copc_3d_loader.html
@@ -1,6 +1,6 @@
 <html>
     <head>
-        <title>Itowns - Entwine 3D loader</title>
+        <title>Itowns - COPC 3D loader</title>
 
         <script type="importmap">
         {
@@ -30,19 +30,23 @@
         </style>
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <!-- <script src="https://cdn.jsdelivr.net/npm/lil-gui@0.19"></script> -->
     </head>
     <body>
-        <div id="description">Specify the URL of a Entwine Point Tree to load:
+        <div id="description">Specify the URL of a COPC file to load:
             <input type="text" id="loadUrl" />
-            <button id='loadUrlBtn'>Load</button>
-            <button id="loadLyonBtn">Load "Grand Lyon" dataset</button>
+            <button id="loadUrlBtn">Load</button>
+            <button id='loadLidarHDBtn'>Load LiDAR HD dataset</button>
             <div id="share"></div>
         </div>
         <div id="viewerDiv"></div>
 
+        <!-- <script src="../dist/itowns.js"></script>
+        <script src="js/GUI/LoadingScreen.js"></script>
+        <script src="../dist/debug.js"></script> -->
+        <!-- <script type="text/javascript"> -->
         <script type="module">
 
-            // TODO rempa click
             import lil from 'lil';
             import GuiTools from 'GuiTools';
             import setupLoadingScreen from 'LoadingScreen';
@@ -64,7 +68,6 @@
                 var layer = new itowns.ColorLayer('Ortho', config);
                 view.addLayer(layer);
             });
-
             // Add two elevation layers.
             // These will deform iTowns globe geometry to represent terrain elevation.
             function addElevationLayerFromConfig(config) {
@@ -75,70 +78,82 @@
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
 
+            var layer;
+
             // Check if an url had been given as an argument
             // and parse to get url and options.
-            const urlParams = new URL(location).searchParams;
-            let url = urlParams.get('ept');
-            urlParams.delete('ept');
+            const urlParams = new URL(location.href).searchParams
+            let url = urlParams.get('copc');
+            urlParams.delete('copc');
 
             const options = {};
             for (const [k, v] of urlParams) {
                 options[k] = isNaN(parseFloat(v, 10)) ? v : parseFloat(v, 10);
             }
 
-            loadEPT(url);
+            loadCopc(url, options);
 
-            var eptLayer;
             function onLayerReady() {
-                var lookAt = eptLayer.root.origin;
+                var lookAt = layer.root.clampOBB.position;
+                var coordLookAt = new itowns.Coordinates(view.referenceCrs).setFromVector3(lookAt);
 
                 var size = new THREE.Vector3();
-                eptLayer.root.voxelOBB.box3D.getSize(size);
+                layer.root.voxelOBB.box3D.getSize(size);
 
                 view.controls.lookAtCoordinate({
-                    coord: lookAt,
+                    coord: coordLookAt,
                     range: 2 * size.length(),
                 }, false);
             }
 
-            function loadEPT(url) {
+            function loadCopc(url, options) {
                 if(!url) return;
+
+                // Share the view
                 document.getElementById('share').innerHTML = '<a href="' +
-                    location.href.replace(location.search, '?ept=' + url) +
+                    location.href.replace(location.search, '?copc=' + url) +
                     '" target="_blank">Link to share this view</a>';
 
-                const eptSource = new itowns.EntwinePointTileSource({ url });
+                const source = new itowns.CopcSource({ url });
 
-                if (eptLayer) {
-                    eptLayer.debugUI.destroy();
-                    view.removeLayer('Entwine Point Tile');
+                if (layer) {
+                    layer.debugUI.destroy();
+                    view.removeLayer('COPC');
                     view.notifyChange();
-                    eptLayer.delete();
+                    layer.delete();
                 }
 
-                eptLayer = new itowns.EntwinePointTileLayer('Entwine Point Tile', {
-                    source: eptSource,
+                const config = {
+                    source,
                     crs: view.referenceCrs,
-                });
+                    sseThreshold: 2,
+                    pointBudget: 3000000,
+                    ...options,
+                }
 
-                itowns.View.prototype.addLayer.call(view, eptLayer).then(onLayerReady);
+                layer = new itowns.CopcLayer('COPC', config);
 
-                eptLayer.whenReady
-                    .then(() => debug.PointCloudDebug.initTools(view, eptLayer, debugGui));
+                itowns.View.prototype.addLayer.call(view, layer).then(onLayerReady);
+
+                layer.whenReady
+                    .then(() => debug.PointCloudDebug.initTools(view, layer, debugGui));
             }
 
             const loadUrlBtn = document.getElementById('loadUrlBtn');
             loadUrlBtn.addEventListener('click', () => {
-                url = document.getElementById('loadUrl').value;
-                loadEPT(url);
+                const url = document.getElementById('loadUrl').value;
+                if (url) { loadCopc(url); }
             });
 
-            const loadLyonBtn = document.getElementById('loadLyonBtn');
-            loadLyonBtn.addEventListener('click', () => {
-                url = 'https://download.data.grandlyon.com/files/grandlyon/imagerie/mnt2018/lidar/ept/';
-                loadEPT(url);
+            const loadLidarHDBtn = document.getElementById('loadLidarHDBtn');
+            loadLidarHDBtn.addEventListener('click', () => {
+                const url ="https://dl-lidar.ign.fr/ppk-lidar/ML/LHD_FXX_0746_6509_PTS_C_LAMB93_IGN69.copc.laz" // fixed with proxy
+                const options = {
+                    mode: 2,
+                    opacity: 0.5,
+                }
+                loadCopc(url, options);
             });
-
         </script>
     </body>
 </html>
--- a/examples/copc_simple_loader.html
+++ b/examples/copc_simple_loader.html
@@ -2,7 +2,7 @@
     <head>
         <meta charset="UTF-8">
 
-        <title>Itowns - COPC loader</title>
+        <title>Itowns - COPC simple loader</title>
 
         <script type="importmap">
         {
@@ -32,12 +32,12 @@
     </head>
     <body>
         <div id="description">Specify the URL of a COPC file to load:
-            <input type="text" id="copc_url" />
-            <button onclick="readURL()">Load</button>
+            <input type="text" id="loadUrl" />
+            <button id="loadUrlBtn">Load</button>
             <div>
-                <button id="loadAutzenButton">Load Autzen Stadium (80mb)</button>
-                <button id="loadSofiButton">Load SoFI Stadium (2.3gb)</button>
-                <button id="loadMillsiteButton">Load Millsite (1.9gb)</button>
+                <button id="loadAutzenBtn">Load Autzen Stadium (80mb)</button>
+                <button id="loadSofiBtn">Load SoFI Stadium (2.3gb)</button>
+                <button id="loadMillsiteBtn">Load Millsite (1.9gb)</button>
             <div id="share"></div>
             </div>
         </div>
@@ -53,32 +53,45 @@
 
             let layer; // COPCLayer
 
-            const uri = new URL(location);
-
+            const viewerDiv = document.getElementById('viewerDiv');
             const gui = new lil();
 
-            const viewerDiv = document.getElementById('viewerDiv');
             const view = new itowns.View('EPSG:4326', viewerDiv);
             const controls = new itowns.PlanarControls(view);
             view.mainLoop.gfxEngine.renderer.setClearColor(0xdddddd);
 
-            setUrl(uri.searchParams.get('copc'));
+            // Check if an url had been given as an argument
+            // and parse to get url and options.
+            const urlParams = new URL(location).searchParams;
+            let url = urlParams.get('copc');
+            urlParams.delete('copc');
+
+            const options = {};
+            for (const [k, v] of urlParams) {
+                options[k] = isNaN(parseFloat(v, 10)) ? v : parseFloat(v, 10);
+            }
+
+            loadCopc(url, options)
+
+            function changeViewCrs(crs) {
+                view.referenceCrs = crs;
+                view.camera.crs = crs;
+            }
 
             function onLayerReady() {
-                const camera = view.camera.camera3D;
+                var lookAt = layer.root.origin.toVector3();
 
-                const lookAt = new THREE.Vector3();
                 const size = new THREE.Vector3();
-                layer.root.bbox.getSize(size);
-                layer.root.bbox.getCenter(lookAt);
+                layer.root.voxelOBB.box3D.getSize(size);
 
-                camera.far = 2.0 * size.length();
-
-                controls.groundLevel = layer.root.bbox.min.z;
-                const position = layer.root.bbox.min.clone().add(
-                    size.multiply({ x: 1, y: 1, z: size.x / size.z }),
+                controls.groundLevel = layer.minElevationRange;
+                var corner = new THREE.Vector3(...layer.source.header.min);
+                var position = corner.clone().add(
+                    size.multiply({ x: 0, y: 0, z: (size.x / size.z) })
                 );
 
+                const camera = view.camera.camera3D;
+                camera.far = 10.0 * size.length();
                 camera.position.copy(position);
                 camera.lookAt(lookAt);
                 camera.updateProjectionMatrix();
@@ -86,83 +99,67 @@
                 view.notifyChange(camera);
             }
 
+            function loadCopc(url, options) {
+                if(!url) return;
 
-            function readURL() {
-                const url = document.getElementById('copc_url').value;
+                // Share the view
+                document.getElementById('share').innerHTML = '<a href="' +
+                    location.href.replace(location.search, '') +
+                    '?copc=' + url
+                    + '" target="_blank">Link to share this view</a>';
 
-                if (url) {
-                    setUrl(url);
-                }
-            }
-
-            function setUrl(url) {
-                if (!url) return;
-
-                const input_url = document.getElementById('copc_url');
-                if (!input_url) return;
-
-                uri.searchParams.set('copc', url);
-                history.replaceState(null, null, `?${uri.searchParams.toString()}`);
-
-                input_url.value = url;
-                load(url);
-            }
-
-
-            function load(url) {
-                const options = {};
-                const urlParams = uri.searchParams
-                urlParams.keys().forEach(key => {
-                    if (key !== 'copc') {
-                        options[key] = parseInt(urlParams.get(key), 10);
-                    }
-                });
                 const source = new itowns.CopcSource({ url });
-
-                if (layer) {
-                    layer.debugUI.destroy()
-                    view.removeLayer('COPC');
-                    view.notifyChange();
-                    layer.delete();
-                }
 
                 const config = {
                     source,
-                    crs: view.referenceCrs,
                     sseThreshold: 2,
                     pointBudget: 3000000,
                     ...options,
                 };
 
-                layer = new itowns.CopcLayer('COPC', config);
-                view.addLayer(layer).then(onLayerReady);
-                layer.whenReady
-                    .then(() => debug.PointCloudDebug.initTools(view, layer, gui));
+                source.whenReady.then(() => {
+                    changeViewCrs(source.crs);
+                    if (layer) {
+                        layer.debugUI.destroy();
+                        view.removeLayer('COPC', true);
+                    }
+                    layer = new itowns.CopcLayer('COPC', config);
+                    view.addLayer(layer).then(onLayerReady);
+                    layer.whenReady
+                        .then(() => debug.PointCloudDebug.initTools(view, layer, gui));
+                })
             }
 
             function loadAutzen() {
-                setUrl("https://s3.amazonaws.com/hobu-lidar/autzen-classified.copc.laz");
+                loadCopc("https://s3.amazonaws.com/hobu-lidar/autzen-classified.copc.laz");
             }
 
             function loadSofi() {
-                setUrl("https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz");
+                loadCopc("https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz");
             }
 
             function loadMillsite() {
-                setUrl("https://s3.amazonaws.com/data.entwine.io/millsite.copc.laz");
+                loadCopc("https://s3.amazonaws.com/data.entwine.io/millsite.copc.laz");
             }
 
-            const loadAutzenButton = document.getElementById('loadAutzenButton');
-            const loadSofiButton = document.getElementById('loadSofiButton');
-            const loadMillsiteButton = document.getElementById('loadMillsiteButton');
+            const loadAutzenBtn = document.getElementById('loadAutzenBtn');
+            const loadSofiBtn = document.getElementById('loadSofiBtn');
+            const loadMillsiteBtn = document.getElementById('loadMillsiteBtn');
 
-            loadAutzenButton.addEventListener('click', loadAutzen);
-            loadSofiButton.addEventListener('click', loadSofi);
-            loadMillsiteButton.addEventListener('click', loadMillsite);
+            loadAutzenBtn.addEventListener('click', loadAutzen);
+            loadSofiBtn.addEventListener('click', loadSofi);
+            loadMillsiteBtn.addEventListener('click', loadMillsite);
 
-            window.view = view;
+            // window.view = view;
             window.itowns = itowns;
             window.THREE = THREE;
+
+            function load() {
+                const url = document.getElementById('loadUrl').value;
+                if (url) { loadCopc(url); }
+            }
+            const loadUrlBtn = document.getElementById('loadUrlBtn');
+            loadUrlBtn.addEventListener('click', load);
 
         </script>
     </body>
--- a/examples/entwine_simple_loader.html
+++ b/examples/entwine_simple_loader.html
@@ -27,17 +27,15 @@
         </style>
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <script src="https://cdn.jsdelivr.net/npm/lil-gui@0.19"></script>
     </head>
     <body>
         <div id="description">Specify the URL of a Entwine Point Tree to load:
-            <input type="text" id="ept_url" />
-            <button id="readUrlButton">Load</button>
-            <button id="loadLyonButton">Load the Grand Lyon dataset</button>
+            <input type="text" id="loadUrl" />
+            <button id="loadUrlBtn">Load</button>
+            <button id="loadLyonBtn">Load the Grand Lyon dataset</button>
             <div id="share"></div>
         </div>
-        <div id="viewerDiv">
-        </div>
+        <div id="viewerDiv"></div>
 
         <script type="module">
 
@@ -53,24 +51,57 @@
             var debugGui = new lil();
             var viewerDiv = document.getElementById('viewerDiv');
             viewerDiv.style.display = 'block';
-            var view = new itowns.View('EPSG:3946', viewerDiv);
-            view.mainLoop.gfxEngine.renderer.setClearColor(0xcccccc);
 
+            var eptLayer;
+
+            const view = new itowns.View('EPSG:4326', viewerDiv);
             const controls = new itowns.PlanarControls(view);
+            view.mainLoop.gfxEngine.renderer.setClearColor(0xdddddd);
 
-            var eptLayer, eptSource;
+            view.domElement.addEventListener('dblclick', dblClickHandler);
+
+            // Check if an url had been given as an argument
+            // and parse to get url and options.
+            const uri = new URL(location.href);
+            const urlParams = uri.searchParams;
+            let url = urlParams.get('ept');
+            urlParams.delete('ept');
+
+            const options = {};
+            for (const [k, v] of urlParams) {
+                options[k] = isNaN(parseFloat(v, 10)) ? v : parseFloat(v, 10);
+            }
+
+            loadEPT(url, options);
+
+            function dblClickHandler(event) {
+                var pick = view.pickObjectsAt(event, 5, eptLayer);
+
+                for (const p of pick) {
+                    console.info('Selected point #' + p.index + ' in position (' +
+                        p.object.position.x + ', ' +
+                        p.object.position.y + ', ' +
+                        p.object.position.z +
+                        ') - node ' + p.object.userData.node.id);
+                }
+            }
+
+            function changeViewCrs(crs) {
+                view.referenceCrs = crs;
+                view.camera.crs = crs;
+            }
 
             function onLayerReady() {
-                var lookAt = new THREE.Vector3();
-                var size = new THREE.Vector3();
-                eptLayer.root.bbox.getSize(size);
-                eptLayer.root.bbox.getCenter(lookAt);
+                var lookAt = eptLayer.root.origin.toVector3();
 
+                var size = new THREE.Vector3();
+                eptLayer.root.voxelOBB.box3D.getSize(size);
                 view.camera3D.far = 2.0 * size.length();
 
-                controls.groundLevel = eptLayer.root.bbox.min.z;
-                var position = eptLayer.root.bbox.min.clone().add(
-                    size.multiply({ x: 0, y: 0, z: size.x / size.z })
+                controls.groundLevel = eptLayer.minElevationRange;
+                var corner = new THREE.Vector3(...eptLayer.source.boundsConforming.slice(0, 3));
+                var position = corner.clone().add(
+                    size.multiply({ x: 0, y: 0, z: (size.x / size.z) })
                 );
 
                 view.camera3D.position.copy(position);
@@ -78,85 +109,60 @@
                 view.camera3D.updateProjectionMatrix();
 
                 view.notifyChange(view.camera3D);
+
+                // add GUI
+                debug.PointCloudDebug.initTools(view, eptLayer, debugGui)
             }
 
-            function readEPTURL() {
-                const urlParams = new URL(location.href).searchParams
-                var url = document.getElementById('ept_url').value || urlParams.get('ept');
+            let eptName
+            function loadEPT(url, options) {
+                if(!url) return;
 
-                if (url) {
-                    const options = {};
-                    urlParams.keys().forEach(key => {
-                        if (key !== 'ept') {
-                            options[key] = parseInt(urlParams.get(key), 10);
-                        }
-                    });
-                    loadEPT(url, options);
-
-                    document.getElementById('share').innerHTML = '<a href="' +
+                // Share the view
+                document.getElementById('share').innerHTML = '<a href="' +
                         location.href.replace(location.search, '') +
                         '?ept=' + url
                         + '" target="_blank">Link to share this view</a>';
-                    document.getElementById('ept_url').value = url;
-                }
-            }
 
-            let eptName;
-            function loadEPT(url, options) {
-                eptSource = new itowns.EntwinePointTileSource({ url });
-
-                if (eptLayer) {
-                    eptLayer.debugUI.destroy();
-                    view.removeLayer(eptName);
-                    view.notifyChange();
-                    eptLayer.delete();
-                }
+                const eptSource = new itowns.EntwinePointTileSource({ url });
 
                 const config = {
                     source: eptSource,
-                    crs: view.referenceCrs,
                     ...options,
-                }
+                };
 
-                eptName = url.split('/').pop()
-                eptName = eptName[0].toUpperCase() + eptName.slice(1);
-                eptLayer = new itowns.EntwinePointTileLayer(eptName, config);
-
-                view.addLayer(eptLayer).then(onLayerReady);
-
-                eptLayer.whenReady
-                    .then(() => debug.PointCloudDebug.initTools(view, eptLayer, debugGui));
-
-                function dblClickHandler(event) {
-                    var pick = view.pickObjectsAt(event, 5, eptLayer);
-
-                    for (const p of pick) {
-                        console.info('Selected point #' + p.index + ' in position (' +
-                            p.object.position.x + ', ' +
-                            p.object.position.y + ', ' +
-                            p.object.position.z +
-                         ') - node ' + p.object.userData.node.id);
+                eptSource.whenReady.then(() => {
+                    changeViewCrs(eptSource.crs);
+                    if (eptLayer) {
+                        eptLayer.debugUI.destroy();
+                        console.log(eptName)
+                        view.removeLayer(eptName, true);
                     }
-                }
-                view.domElement.addEventListener('dblclick', dblClickHandler);
+
+                    if(url.slice(-1) === '/') url = url.slice(0, -1);
+
+                    eptName = url.split('/').pop()
+                    eptName = eptName[0].toUpperCase() + eptName.slice(1);
+                    eptLayer = new itowns.EntwinePointTileLayer(eptName, config);
+                    view.addLayer(eptLayer).then(onLayerReady);
+
+                })
             }
 
-            function onloadGrandLyon() {
-                document.getElementById('ept_url').value = 'https://download.data.grandlyon.com/files/grandlyon/imagerie/mnt2018/lidar/ept';
-                readEPTURL();
-            }
+            const loadUrlBtn = document.getElementById('loadUrlBtn');
+            loadUrlBtn.addEventListener('click', () => {
+                url = document.getElementById('loadUrl').value;
+                if (url) { loadEPT(url); }
+            });
 
-            const loadLyonButton = document.getElementById('loadLyonButton');
-            const readUrlButton = document.getElementById('readUrlButton');
+            const loadLyonBtn = document.getElementById('loadLyonBtn');
+            loadLyonBtn.addEventListener('click', () => {
+                url = 'https://download.data.grandlyon.com/files/grandlyon/imagerie/mnt2018/lidar/ept/';
+                loadEPT(url);
+            });
 
-            loadLyonButton.addEventListener('click', onloadGrandLyon);
-            readUrlButton.addEventListener('click', readEPTURL);
-
-            readEPTURL();
-            window.view = view;
             window.itowns = itowns;
             window.THREE = THREE;
-
         </script>
     </body>
 </html>
--- a/examples/potree2_25d_map.html
+++ b/examples/potree2_25d_map.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!-- <!DOCTYPE html> -->
 <html>
     <head>
         <title>Point Cloud Viewer</title>
@@ -47,99 +47,101 @@
             <div id="info"></div>
         </div>
 
-        <div>
         <script type="module">
 
-                import lil from 'lil';
-                import setupLoadingScreen from 'LoadingScreen';
-                import * as THREE from 'three';
-                import * as itowns from 'itowns';
-                import * as debug from 'debug';
-                import * as itowns_widgets from 'itowns_widgets';
+            import lil from 'lil';
+            import setupLoadingScreen from 'LoadingScreen';
+            import * as THREE from 'three';
+            import * as itowns from 'itowns';
+            import * as debug from 'debug';
+            import * as itowns_widgets from 'itowns_widgets';
 
-                var potreeLayer;
-                var oldPostUpdate;
-                var viewerDiv;
-                var debugGui;
-                var view;
-                var controls;
+            var potreeLayer;
+            var oldPostUpdate;
+            var viewerDiv;
+            var debugGui;
+            var view;
+            var controls;
 
-                // Define crs projection that we will use (taken from https://epsg.io/3946, Proj4js section)
-                itowns.CRS.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+            // Define crs projection that we will use (taken from https://epsg.io/3946, Proj4js section)
+            itowns.CRS.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
 
-                viewerDiv = document.getElementById('viewerDiv');
-                viewerDiv.style.display = 'block';
+            viewerDiv = document.getElementById('viewerDiv');
+            viewerDiv.style.display = 'block';
 
-                debugGui = new lil();
+            debugGui = new lil();
 
-                // TODO: do we really need to disable logarithmicDepthBuffer ?
-                view = new itowns.View('EPSG:3946', viewerDiv);
-                setupLoadingScreen(viewerDiv, view);
-                view.mainLoop.gfxEngine.renderer.setClearColor(0xcccccc);
+            const crs = 'EPSG:4326';
+            view = new itowns.View(crs, viewerDiv);
+            setupLoadingScreen(viewerDiv, view);
+            view.mainLoop.gfxEngine.renderer.setClearColor(0xcccccc);
 
-                // Configure Point Cloud layer
-                potreeLayer = new itowns.Potree2Layer('Lion', {
-                    source: new itowns.Potree2Source({
-                        file: 'metadata.json',
-                        url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/potree2.0/lion',
-                        crs: view.referenceCrs,
-                    }),
+            // Configure Point Cloud layer
+            const source = new itowns.Potree2Source({
+                file: 'metadata.json',
+                url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/potree2.0/lion',
+                crs,
+            });
+
+            potreeLayer = new itowns.Potree2Layer('Lion', {
+                source,
+            });
+
+            // point selection on double-click
+            function dblClickHandler(event) {
+                var pick = view.pickObjectsAt(event, 5, potreeLayer);
+
+                for (const p of pick) {
+                    console.info('Selected point #' + p.index + ' in position (' +
+                        p.object.position.x + ', ' +
+                        p.object.position.y + ', ' +
+                        p.object.position.z +
+                        ') in Points ' + p.object.layer.id);
+                }
+            }
+            view.domElement.addEventListener('dblclick', dblClickHandler);
+
+            function placeCamera(position, lookAt) {
+                view.camera.camera3D.position.copy(position);
+                view.camera.camera3D.lookAt(lookAt);
+                // create controls
+                controls = new itowns.FirstPersonControls(view, { focusOnClick: true });
+                debugGui.add(controls.options, 'moveSpeed', 1, 100).name('Movement speed');
+
+                view.notifyChange(view.camera.camera3D);
+            }
+
+            // add potreeLayer to scene
+            function onLayerReady() {
+                var position;
+                var lookAt = new THREE.Vector3();
+                var size = new THREE.Vector3();
+
+                potreeLayer.root.voxelOBB.box3D.getSize(size);
+                potreeLayer.root.voxelOBB.box3D.getCenter(lookAt);
+                lookAt.z = potreeLayer.root.voxelOBB.box3D.min.z;
+
+                debug.PointCloudDebug.initTools(view, potreeLayer, debugGui);
+
+                view.camera.camera3D.far = 2.0 * size.length();
+
+                var corner = new THREE.Vector3(...potreeLayer.source.metadata.boundingBox.min);
+                var position = corner.clone().add(
+                    size.multiply({ x: 0, y: 0, z: (size.x / size.z) })
+                );
+
+                placeCamera(position, lookAt);
+                controls.moveSpeed = size.length() / 3;
+
+                // update stats window
+                var info = document.getElementById('info');
+                view.addFrameRequester(itowns.MAIN_LOOP_EVENTS.AFTER_RENDER, () => {
+                    info.textContent = potreeLayer.displayedCount.toLocaleString() + ' points';
                 });
-
-                // point selection on double-click
-                function dblClickHandler(event) {
-                    var pick = view.pickObjectsAt(event, 5, potreeLayer);
-
-                    for (const p of pick) {
-                        console.info('Selected point #' + p.index + ' in position (' +
-                            p.object.position.x + ', ' +
-                            p.object.position.y + ', ' +
-                            p.object.position.z +
-                         ') in Points ' + p.object.layer.id);
-                    }
-                }
-                view.domElement.addEventListener('dblclick', dblClickHandler);
-
-                function placeCamera(position, lookAt) {
-                    view.camera.camera3D.position.copy(position);
-                    view.camera.camera3D.lookAt(lookAt);
-                    // create controls
-                    controls = new itowns.FirstPersonControls(view, { focusOnClick: true });
-                    debugGui.add(controls.options, 'moveSpeed', 1, 100).name('Movement speed');
-
-                    view.notifyChange(view.camera.camera3D);
-                }
-
-                // add potreeLayer to scene
-                function onLayerReady() {
-                    var ratio;
-                    var position;
-                    var lookAt = new THREE.Vector3();
-                    var size = new THREE.Vector3();
-
-                    potreeLayer.root.bbox.getSize(size);
-                    potreeLayer.root.bbox.getCenter(lookAt);
-
-                    debug.PointCloudDebug.initTools(view, potreeLayer, debugGui);
-
-                    view.camera.camera3D.far = 2.0 * size.length();
-
-                    ratio = size.x / size.z;
-                    position = potreeLayer.root.bbox.min.clone().add(
-                        size.multiply({ x: 0, y: 0, z: ratio * 0.5 }));
-                    lookAt.z = potreeLayer.root.bbox.min.z;
-                    placeCamera(position, lookAt);
-                    controls.moveSpeed = size.length() / 3;
-
-                    // update stats window
-                    var info = document.getElementById('info');
-                    view.addFrameRequester(itowns.MAIN_LOOP_EVENTS.AFTER_RENDER, () => {
-                        info.textContent = potreeLayer.displayedCount.toLocaleString() + ' points';
-                    });
-                }
-                window.view = view;
-                view.addLayer(potreeLayer).then(onLayerReady);
-                window.view = view;
+            }
+            window.view = view;
+            view.addLayer(potreeLayer).then(onLayerReady);
+            window.view = view;
             window.itowns = itowns;
             window.THREE = THREE;
 
--- a/examples/potree_25d_map.html
+++ b/examples/potree_25d_map.html
@@ -65,8 +65,24 @@
                 var view;
                 var controls;
 
-                // Define crs projection that we will use (taken from https://epsg.io/3946, Proj4js section)
-                itowns.CRS.defs('EPSG:3946', '+proj=lcc +lat_1=45.25 +lat_2=46.75 +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs');
+                /* OTHER available dataset
+                const name = 'eglise_saint_blaise_arles';
+                const url = 'https://raw.githubusercontent.com/gmaillet/dataset/master/';
+                const file = 'eglise_saint_blaise_arles.js';
+                const crs = 'EPSG:4978';
+                => Geocentric
+
+                Probleme de CORS if 'Allow CORS' plugin off
+                const url = 'https://cors-anywhere.herokuapp.com/https://viz.lib.vt.edu/thessaloniki/resources/pointclouds/thessaloniki';
+                const file = 'cloud.js';
+                const crs = 'EPSG:3857';
+                => crs unknown. Most probaly a projection centered on the center of acquisition... 
+                */
+
+                itowns.CRS.defs("EPSG:25831","+proj=utm +zone=31 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +type=crs");
+                const url = 'https://betaserver.icgc.cat/potree12/resources/pointclouds/templesagradafamilia';
+                const file = 'cloud.js';
+                const crs = 'EPSG:25831';
 
                 viewerDiv = document.getElementById('viewerDiv');
                 viewerDiv.style.display = 'block';
@@ -74,17 +90,19 @@
                 debugGui = new lil();
 
                 // TODO: do we really need to disable logarithmicDepthBuffer ?
-                view = new itowns.View('EPSG:3946', viewerDiv);
+                view = new itowns.View(crs, viewerDiv);
                 setupLoadingScreen(viewerDiv, view);
                 view.mainLoop.gfxEngine.renderer.setClearColor(0xcccccc);
 
                 // Configure Point Cloud layer
+                const potreeSource = new itowns.PotreeSource({
+                    file,
+                    url,
+                    crs,
+                });
+
                 potreeLayer = new itowns.PotreeLayer('eglise_saint_blaise_arles', {
-                    source: new itowns.PotreeSource({
-                        file: 'eglise_saint_blaise_arles.js',
-                        url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/eglise_saint_blaise_arles',
-                        crs: view.referenceCrs,
-                    }),
+                    source: potreeSource,
                 });
 
                 // point selection on double-click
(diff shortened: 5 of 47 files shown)


src/Loader/LASLoader.js
```diff
@@ -75,17 +82,23 @@ class LASLoader {
...
+            // we thus apply the projection to get values in the Crs of the view.
+            const point = getPosition.map(f => f(i));
+            const [x, y, z] = forward(point);
             positions[i * 3] = x - origin[0];
             positions[i * 3 + 1] = y - origin[1];
             positions[i * 3 + 2] = z - origin[2];
 
+            elevations[i] = z;
+            // geocentric height to elevation
+            if (isGeocentric) { elevations[i] = point[2]; }
```
gchoqueux: What your motivation to add elevation attribute ?
I see it's to avoid a transformation, in return  you add a buffer and transfer to GPU

src/Layer/CopcLayer.js
```diff
@@ -45,15 +46,29 @@ class CopcLayer extends PointCloudLayer {
...
-            this.root.bbox.max.fromArray(cube, 3);
+
+            let forward = (x => x);
+            if (this.source.crs !== this.crs) {
+                try {
+                    forward = proj4(this.source.crs, this.crs).forward;
+                } catch (err) {
+                    throw new Error(`${err} is not defined in proj4`);
+                }
+            }
```
gchoqueux: Could be factorized in parent method

src/Layer/CopcLayer.js
```diff
@@ -45,15 +46,29 @@ class CopcLayer extends PointCloudLayer {
...
             this.scale = new THREE.Vector3(1.0, 1.0, 1.0);
             this.offset = new THREE.Vector3(0.0, 0.0, 0.0);
 
+            const bounds = [
+                ...forward(cube.slice(0, 3)),
+                ...forward(cube.slice(3, 6)),
+            ];
+
+            this.root.bbox.setFromArray(bounds);
+
```
gchoqueux: Could be factorized in parent method

examples/copc_simple_loader.html
```diff
@@ -2,7 +2,7 @@
     <head>
```
gchoqueux, line 2: Could you explain more the adding ?

src/Loader/LASLoader.js
ftoromanoff: I added the elevation for when we want the data to be projected in a geocentric crs in a globe view.
without:
image
with:
image

It might be more linked with the followng of my modification (see PR #2272 ) maybe I should then not add this modification here...

examples/copc_simple_loader.html
ftoromanoff: In this example we can switch between 3 differents copc with differents crs. As the view need to be instanciate before any layer, we need to redefine the view each time we want to switch and for that we need to know the crs before hand (we can wait the layer to be loaded to get the crs to instanciate the view)

src/Layer/PotreeLayer.js
```diff
@@ -67,8 +68,25 @@ class PotreeLayer extends PointCloudLayer {
...
+            if (this.source.crs !== this.crs) {
+                try {
+                    forward = proj4(this.source.crs, this.crs).forward;
+                } catch (err) {
+                    throw new Error(`${err} is not defined in proj4`);
+                }
+            }
+
+            this.minElevationRange = cloud.tightBoundingBox.lz;
+            this.maxElevationRange = cloud.tightBoundingBox.uz;
```
ftoromanoff: Doublon linked to rebase.

src/Loader/LASLoader.js
gchoqueux: Yes, I know
It's possible to transform points in local space with the z axis is perpendicular to the ellipsoid.
Like, the class Feature computes local transforming, this operation have advantages : bounding box in good orientation and Z axe in elevation direction. This transformation can also be used to handle the display of elevations

src/Layer/CopcLayer.js
ftoromanoff: done.
I added a new methode in PointCloudLayer.js to compute the bbox and one to compute the ElevationRange.

src/Layer/CopcLayer.js
ftoromanoff: see previous reply.

src/Layer/PotreeLayer.js
ftoromanoff: Cleaned

src/Loader/LASLoader.js
ftoromanoff: I think I will remove this modificaiton from here, and keep it in the other PR, and we will discuss it better at that moment.

gchoqueux: 1. ok

examples/copc_simple_loader.html
gchoqueux: ok

src/Layer/CopcLayer.js
gchoqueux: ok

src/Loader/LASLoader.js
```diff
@@ -47,7 +48,11 @@ class LASLoader {
     }
 
     _parseView(view, options) {
-        const colorDepth = options.colorDepth ?? 16;
+        const colorDepth = options.colorDepth ?? defaultColorEncoding(options.header);
+
+        const forward = (options.in.crs !== options.out.crs) ?
+            proj4(options.in.projDefs, options.out.projDefs).forward :
+            (x => x);
```
gchoqueux, line 86: this is same part with foward, could you factorize ?

src/Loader/LASLoader.js
```diff
@@ -75,13 +80,16 @@ class LASLoader {
...
         for (let i = 0; i < view.pointCount; i++) {
             // `getPosition` apply scale and offset transform to the X, Y, Z
             // values. See https://github.com/connormanning/copc.js/blob/master/src/las/extractor.ts.
-            const [x, y, z] = getPosition.map(f => f(i));
+            // we thus apply the projection to get values in the Crs of the view.
+            const point = getPosition.map(f => f(i));
+            const [x, y, z] = forward(point);
             positions[i * 3] = x - origin[0];
             positions[i * 3 + 1] = y - origin[1];
             positions[i * 3 + 2] = z - origin[2];
```
gchoqueux: it's fist transformation to move on local space
There isn't the rotation

src/Parser/LASParser.js
```diff
@@ -109,8 +110,16 @@ export default {
...
+            eb: options.in.eb,
             colorDepth: options.in.colorDepth,
+            in: {
+                crs: options.in.crs,
+                projDefs: proj4.defs(options.in.crs),
+            },
+            out: {
+                crs: options.out.crs,
+                projDefs: proj4.defs(options.out.crs),
+            },
```
gchoqueux, line 125: Factorize with the line 154

src/Layer/PointCloudLayer.js
```diff
@@ -231,6 +232,29 @@ class PointCloudLayer extends GeometryLayer {
         this.root = undefined;
     }
 
+    setRootBbox(min, max) {
```
gchoqueux: could you the same signature and this
`setRootBboxFromArray(bounds)`

src/Loader/LASLoader.js
ftoromanoff: LASLoader will be called by the worker, so we need to limite the dependances, thus it would be counter-productive.

src/Layer/PointCloudLayer.js
ftoromanoff: If i understand you, you want to add another methode, with a different signature. But currently this function is call 4 times, if we add a 2nd one, both will be called 2 times, won't it be unfactorizing some code ?

src/Parser/LASParser.js
ftoromanoff: I don't really see the benefit of doing so, moreover when we have the idea (in the futur) to allows proj4.defs as crs. In this cas it will simplifed the options as passing only the crs.

jailln:
Nice improvement thanks :)
What is the time overhead of reprojecting ? Can we measure it ? Maybe we should add a comment in the doc explaining that it will add an overhead (and if possible a percentage)?

examples/copc_3d_loader.html
```diff
@@ -0,0 +1,143 @@
...
+                layer = new itowns.CopcLayer('COPC', config);
+
+                itowns.View.prototype.addLayer.call(view, layer).then(onLayerReady);
+
+                layer.whenReady
+                    .then(() => debug.PointCloudDebug.initTools(view, layer, debugGui));
+            }
+
+            function loadLidarHD() {
+                // const url ="https://storage.sbg.cloud.ovh.net/v1/AUTH_63234f509d6048bca3c9fd7928720ca1/ppk-lidar/ML/LHD_FXX_0746_6509_PTS_C_LAMB93_IGN69.copc.laz"
```
airnez: Is it relevant to keep this comment here in example code?

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -22,30 +25,44 @@ class PointCloudNode extends THREE.EventDispatcher {
...
-        node.bbox.min.add(translation);
+        childNode._bbox.min.add(translation);
 
         // use the size computed above to set the max
-        node.bbox.max.copy(node.bbox.min).add(size);
+        childNode._bbox.max.copy(childNode._bbox.min).add(size);
+    }
+
+    getCenter() {
+        // get center of the bbox in the world referentiel
```
airnez:
```suggestion
        // get center of the bbox in the world referential
```

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -22,30 +25,44 @@ class PointCloudNode extends THREE.EventDispatcher {
...
+        const centerBbox = new THREE.Vector3();
+        this._bbox.getCenter(centerBbox);
+        const rotInv = this._quaternion.clone().invert();
+        const origVector = this._position;
+        this.center = centerBbox.clone().applyQuaternion(rotInv).add(origVector);
```
airnez: I always prefer non-abbreviated variables when possible. What do you think?
```suggestion
        const inverseRotation = this._quaternion.clone().invert();
        const originVector = this._position;
        this.center = centerBbox.clone().applyQuaternion(inverseRotation).add(originVector);
```

packages/Main/src/Core/Potree2Node.js
```diff
@@ -70,44 +70,50 @@ class Potree2Node extends PointCloudNode {
...
-            node.bbox.max.x += dHalfLength.x;
+            childNode._bbox.min.y += dHalfLength.y;
+            childNode._bbox.max.y += dHalfLength.y;
+            childNode._bbox.min.x += dHalfLength.x;
+            childNode._bbox.max.x += dHalfLength.x;
         }
     }
 
+    getCenter() {
+        // for potree we send the point use as origin for the data
```
airnez: Phrasing seemed a bit odd
```suggestion
        // for potree we use the origin as center for the data
```

packages/Main/src/Core/PotreeNode.js
```diff
@@ -34,44 +34,50 @@ class PotreeNode extends PointCloudNode {
...
-            node.bbox.max.x += dHalfLength.x;
+            childNode._bbox.min.y += dHalfLength.y;
+            childNode._bbox.max.y += dHalfLength.y;
+            childNode._bbox.min.x += dHalfLength.x;
+            childNode._bbox.max.x += dHalfLength.x;
         }
     }
 
+    getCenter() {
+        // for potree we send the point use as origin for the data
```
airnez:
```suggestion
        // for potree we use the origin as center for the data
```

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -2,28 +2,59 @@ import * as THREE from 'three';
...
 const box3 = new THREE.Box3();
 bboxMesh.geometry.boundingBox = box3;
 
+function positionFromBbox(bbox) {
+    const array =  new Float32Array(8 * 3);
+
+    const min = bbox.min;
+    const max = bbox.max;
+
+    array[0] = max.x; array[1] = max.y; array[2] = max.z;
```
airnez: It could be nice to explain here with a comment that this array corresponds to corner positions.
What do you think?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -314,12 +429,14 @@ class PointCloudLayer extends GeometryLayer {
                             initBoundingBox(elt, layer);
                         }
                         elt.obj.boxHelper.visible = true;
-                        elt.obj.boxHelper.material.color.r = 1 - elt.sse;
-                        elt.obj.boxHelper.material.color.g = elt.sse;
+                        // elt.obj.boxHelper.material.color.r = 1 - elt.sse;
+                        // elt.obj.boxHelper.material.color.g = elt.sse;
```
airnez: Are this lines intentionally commented ?

packages/Main/src/Source/Potree2Source.js
```diff
@@ -151,6 +151,10 @@ class Potree2Source extends Source {
         if (!source.file) {
             throw new Error('New Potree2Source: file is required');
         }
+        if (!source.crs) {
+            // with better data and the spec this might be removed
+            throw new Error('New PotreeSource: crs is required');
```
airnez, line 222: Is this a breaking change?

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -22,30 +25,44 @@ class PointCloudNode extends THREE.EventDispatcher {
...
         // apply the translation to the child node bbox
-        node.bbox.min.add(translation);
+        childNode._bbox.min.add(translation);
 
         // use the size computed above to set the max
-        node.bbox.max.copy(node.bbox.min).add(size);
+        childNode._bbox.max.copy(childNode._bbox.min).add(size);
+    }
+
+    getCenter() {
```
airnez: I would rename this method `computeCenter()`
It does not return anything, and but updates the center attribute depending on implementation

packages/Main/src/Loader/LASLoader.js
```diff
@@ -75,16 +114,33 @@ class LASLoader {
...
         for (let i = 0; i < view.pointCount; i++) {
             // `getPosition` apply scale and offset transform to the X, Y, Z
             // values. See https://github.com/connormanning/copc.js/blob/master/src/las/extractor.ts.
-            const [x, y, z] = getPosition.map(f => f(i));
-            positions[i * 3] = x - origin[0];
-            positions[i * 3 + 1] = y - origin[1];
-            positions[i * 3 + 2] = z - origin[2];
+            // we thus apply the projection to get values in the Crs of the view.
+            const point = getPosition.map(f => f(i));
+            const [x, y, z] = forward(point);
```
airnez, line 126: Is there any performance issue there? It is obvious that there a little overload to compute transformations. But what about memory bloating?
Is there some kind of "data copy / dupplication" added ?
It looks good to me but I would appreciate to have your opinion.

airnez (commented):
Thank you for this work. That's a nice improvement :)
General architecture seems fine and I have few to say about reprojection in itself. Except a remark on performances, already mentionned by @jailln 
I made few comments around naming, phrasing and documentation.

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: fixed

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: changed

packages/Main/src/Core/Potree2Node.js
ftoromanoff: I removed this method as it's useless for Potree2Node

packages/Main/src/Core/PotreeNode.js
ftoromanoff: I changed the commentary, to try to explain better why we don't use the center of the bbox for potree.

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: I did add some documentation on the function and soem commentary. It's is clearer ?

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: No it's a miss ^^. I don't really find these lines really helpful, so I do comment them when I do my tests...

packages/Main/src/Source/Potree2Source.js
ftoromanoff: It's a homogeneisation with all the other PointCloudLayer...

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: I did change it the way this function works (see previous commentary). I think it's clearer that way. But you opinion still matter.

examples/copc_3d_loader.html
ftoromanoff: I hope that at some point we won't need to send the data through a server to avoid the 'blocked by CORS policy'.

packages/Main/src/Loader/LASLoader.js
ftoromanoff: There is indeed an extra calculation but there is no more data than before, no copy of duplication are made. As before we read the data from the 'view' and write them in the 'position' array.

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -4,6 +4,16 @@ const size = new THREE.Vector3();
 const position = new THREE.Vector3();
 const translation = new THREE.Vector3();
 
+/**
+ * @property {nimber} numPoints
```
ftoromanoff: typo

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -4,6 +4,16 @@ const size = new THREE.Vector3();
...
+ * @property {Layer} layer
+ * @property {array} children
+ * @property {THREE.bbox} bbox - bbox in local referential
```
ftoromanoff: THREE.box3

examples/copc_3d_loader.html
airnez: Ok makes sense ! Thank you

packages/Main/src/Layer/PointCloudLayer.js
airnez: Yes it is great thanks !

packages/Main/src/Source/Potree2Source.js
airnez: Ok thank you :)

packages/Main/src/Core/PointCloudNode.js
airnez: That's perfect

packages/Main/src/Loader/LASLoader.js
airnez: Ok looks good to me then

airnez (approved):
Looks good to me✌️

examples/copc_simple_loader.html
```diff
@@ -35,92 +35,82 @@
...
+                controls.groundLevel = layer.minElevationRange;
+                var corner = new itowns.THREE.Vector3(...layer.source.header.min);
+                var position = corner.clone().add(
+                    size.multiply({ x: 0, y: 0, z: (size.x / size.z) })
                 );
 
+                camera.far = 5.0 * size.length();
                 camera.position.copy(position);
                 camera.lookAt(lookAt);
                 camera.updateProjectionMatrix();
```
gchoqueux, line 97: ` CameraUtils.getCameraTransformOptionsFromExtent` could simplify this part

examples/copc_simple_loader.html
```diff
@@ -35,92 +35,82 @@
...
                     pointBudget: 3000000,
                     ...options,
                 };
 
-                layer = new itowns.CopcLayer('COPC', config);
-                view.addLayer(layer).then(onLayerReady);
-                layer.whenReady
-                    .then(() => debug.PointCloudDebug.initTools(view, layer, gui));
+                source.whenReady.then(() => {
+                    setView(source.crs);
```
gchoqueux: a view is instanced for each loading ?

packages/Main/src/Core/CopcNode.js
```diff
@@ -144,6 +144,8 @@ class CopcNode extends PointCloudNode {
             this.layer,
             pointCount,
         );
+        child._quaternion = this._quaternion;
+        child._position = this._position;
```
gchoqueux: By adding these properties, your object looks like an Object3D. Shouldn't it inherit from that class instead

packages/Main/src/Core/CopcNode.js
```diff
@@ -157,13 +159,18 @@ class CopcNode extends PointCloudNode {
...
         const geometry = await this.layer.source.parser(buffer, {
             in: {
                 ...this.layer.source,
                 pointCount: this.numPoints,
             },
-            out: this.layer,
+            out: {
+                ...this.layer,
+                origin,
+            },
```
gchoqueux: `...this.layer,` you copy the layer and the source are copied for each parsing ?
`pointCount` and `origin` don't change?

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -22,30 +35,47 @@ class PointCloudNode extends THREE.EventDispatcher {
...
+     * Compute the center of the bounding box in the local referential
+     * @returns {THREE.Vector3}
+     */
+    getCenter() {
+        const centerBbox = new THREE.Vector3();
+        this._bbox.getCenter(centerBbox);
+        const inverseRotation = this._quaternion.clone().invert();
+        const originVector = this._position;
+        return centerBbox.clone().applyQuaternion(inverseRotation).add(originVector);
     }
```
gchoqueux: it's local position, likes [`Object3D.position`]( https://threejs.org/docs/#api/en/core/Object3D.position) ?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -2,28 +2,83 @@ import * as THREE from 'three';
...
+    boxHelper.updateMatrixWorld(true);
+
+    // tightbbox in local ref -> blue
+    const tightboxHelper = new THREE.BoxHelper(undefined, 0x0000ff);
+    tightboxHelper.geometry.attributes.position.array = getCornerPosition(elt.obj.geometry.boundingBox);
+    tightboxHelper.applyMatrix4(elt.obj.matrixWorld);
+    elt.obj.tightboxHelper = tightboxHelper;
+    layer.bboxes.add(tightboxHelper);
+    tightboxHelper.updateMatrixWorld(true);
+}
```
gchoqueux: I didn't remember that we were creating helpers; it should be moved to debug.

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -291,17 +410,37 @@ class PointCloudLayer extends GeometryLayer {
...
             return;
         }
 
         elt.notVisibleSince = undefined;
-        point.copy(context.camera.camera3D.position).sub(this.object3d.getWorldPosition(new THREE.Vector3()));
-        point.applyQuaternion(this.object3d.getWorldQuaternion(new THREE.Quaternion()).invert());
+
+        point.copy(context.camera.camera3D.position)
+            .sub(obj.getWorldPosition(new THREE.Vector3()))
+            .applyQuaternion(obj.getWorldQuaternion(new THREE.Quaternion()).invert());
```
gchoqueux: could avoid instancing Vector3 and Quaternion

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -434,6 +572,17 @@ class PointCloudLayer extends GeometryLayer {
...
+                        obj.tightboxHelper.removeMe = true;
+                        if (Array.isArray(obj.tightboxHelper.material)) {
+                            for (const material of obj.tightboxHelper.material) {
+                                material.dispose();
+                            }
+                        } else {
+                            obj.tightboxHelper.material.dispose();
+                        }
+                        obj.tightboxHelper.geometry.dispose();
+                    }
```
gchoqueux: The place to handle the helper is debug

packages/Main/src/Layer/Potree2Layer.js
```diff
@@ -167,32 +160,31 @@ class Potree2Layer extends PointCloudLayer {
...
             root.nodeType = 2;
             root.hierarchyByteOffset = 0n;
             root.hierarchyByteSize = BigInt(metadata.hierarchy.firstChunkSize);
-
             root.byteOffset = 0;
-
             this.root = root;
 
-            this.extent = Extent.fromBox3(this.source.crs || 'EPSG:4326', boundingBox);
+            this.setRootBbox(metadata.boundingBox.min, metadata.boundingBox.max);
```
gchoqueux: Remove the extent?

packages/Main/src/Loader/LASLoader.js
```diff
@@ -19,6 +20,37 @@ function defaultColorEncoding(header) {
...
+
+    const res = [];
+    // v + q.w * t + cross( q.xyz, t );
+    res[0] = vx + qw * tx + qy * tz - qz * ty;
+    res[1] = vy + qw * ty + qz * tx - qx * tz;
+    res[2] = vz + qw * tz + qx * ty - qy * tx;
+
+    return res;
+}
+
```
gchoqueux, line 53: You copy the three.js code ?

packages/Main/src/Loader/LASLoader.js
```diff
@@ -75,16 +114,33 @@ class LASLoader {
...
+            positions[i * 3] = position[0];
+            positions[i * 3 + 1] = position[1];
+            positions[i * 3 + 2] = position[2];
+
+            minX = Math.min(minX, position[0]);
+            minY = Math.min(minY, position[1]);
+            minZ = Math.min(minZ, position[2]);
+            maxX = Math.max(maxX, position[0]);
+            maxY = Math.max(maxY, position[1]);
+            maxZ = Math.max(maxZ, position[2]);
```
gchoqueux, line 143: there's Three.Box method ?

packages/Main/src/Parser/LASParser.js
```diff
@@ -106,15 +122,29 @@ export default {
...
+                crs: options.in.crs,
+                projDefs: proj4.defs(options.in.crs),
+            },
+            out: {
+                crs: options.out.crs,
+                projDefs: proj4.defs(options.out.crs),
+                origin,
+                rotation: quaternion.toArray(),
+            },
         });
```
gchoqueux, line 126: could you explain the projDefs option?

packages/Main/src/Parser/Potree2BinParser.js
```diff
@@ -95,6 +97,9 @@ export default {
             }
         });
 
+        geometry.userData.origin = options.out.origin;
+        geometry.userData.rotation = new THREE.Quaternion();
+
```
gchoqueux: Transformation properties in geometry  : This doesn't seem to be the right place.

packages/Main/src/Loader/LASLoader.js
```diff
@@ -182,15 +185,15 @@ class LASLoader {
         const pointData = await Las.PointData.decompressFile(bytes, this._initDecoder());
 
         const header = Las.Header.parse(bytes);
-        const colorDepth = options.colorDepth ?? defaultColorEncoding(header);
+        options.header = header;
```
gchoqueux, line 239: where the header is used in this commit ?

examples/copc_simple_loader.html
ftoromanoff: Is there any other ways to reset the view (to get a different crs) ?

examples/copc_3d_loader.html
```diff
@@ -0,0 +1,169 @@
...
+                                object[subKey] = parseFloat(urlParams.get(key));
+                            }
+                            object = object[subKey];
+                        })
+                    } else {
+                        options[key] = parseFloat(urlParams.get(key));
+                    }
+                    
+                }
+            });
```
gchoqueux: this part complex  to extract : use URLSearchParams instead

examples/copc_simple_loader.html
```diff
@@ -53,91 +53,80 @@
...
 
-            setUrl(uri.searchParams.get('copc'));
+            const options = {};
+            urlParams.keys().forEach(key => {
+                if (key !== 'copc') {
+                    options[key] = parseInt(urlParams.get(key), 10);
+                }
+            });
+            
+            loadCopc(url, options)
```
gchoqueux, line 74: use `URLSearchParams`

examples/copc_simple_loader.html
```diff
@@ -53,91 +53,80 @@
...
+
+            let view;
+            let controls;
+            function setView(crs) {
+                // viewerDiv.innerHTML = '';
+                view = new itowns.View(crs, viewerDiv);
+                controls = new itowns.PlanarControls(view);
+                view.mainLoop.gfxEngine.renderer.setClearColor(0xdddddd);
```
gchoqueux: ~~use specific option clear color~~

examples/copc_simple_loader.html
gchoqueux: could use getCameraTransformOptionsFromExtent ?

EDIT already discussed.

examples/copc_simple_loader.html
gchoqueux: why change view crs if you convert cloud crs ?
EDIT you could use:
```js
                view.referenceCrs = crs;
                view.camera.crs = crs;
```

examples/copc_simple_loader.html
```diff
@@ -53,91 +53,80 @@
...
+                    view.addLayer(layer).then(onLayerReady);
+                    layer.whenReady
+                        .then(() => debug.PointCloudDebug.initTools(view, layer, gui));
+                })
+            }
+
+            function setUrl(urlCopc) {
+                const url = location.href.replace(location.search, '') +
+                    '?copc=' + urlCopc;
+                location.assign(url);
```
gchoqueux: URLSearchParams ?

examples/entwine_3d_loader.html
```diff
@@ -75,40 +75,34 @@
...
+            const urlParams = new URL(location.href).searchParams;
+            let url = urlParams.get('ept');
+            loadEPT(url);
 
+            var eptLayer;
             function onLayerReady() {
-                var lookAt = new THREE.Vector3();
-                eptLayer.root.bbox.getCenter(lookAt);
-                var coordLookAt = new itowns.Coordinates(view.referenceCrs, lookAt);
+                var lookAt = eptLayer.root.origin;
```
gchoqueux: URLSearchParams ?

examples/entwine_3d_loader.html
```diff
@@ -75,40 +75,34 @@
...
 
                 var size = new THREE.Vector3();
-                eptLayer.root.bbox.getSize(size);
+                eptLayer.root.voxelOBB.box3D.getSize(size);
 
                 view.controls.lookAtCoordinate({
-                    coord: coordLookAt,
+                    coord: lookAt,
                     range: 2 * size.length(),
                 }, false);
```
gchoqueux, line 101: Could use extent and camera utils ?

examples/entwine_3d_loader.html
```diff
@@ -75,40 +75,34 @@
...
-                }
-            }
-
             function loadEPT(url) {
-                eptSource = new itowns.EntwinePointTileSource({ url });
+                if(!url) return;
+                document.getElementById('share').innerHTML = '<a href="' +
+                    location.href.replace(location.search, '?ept=' + url) +
+                    '" target="_blank">Link to share this view</a>';
+
```
gchoqueux, line 109: URLSearchParams

examples/entwine_simple_loader.html
```diff
@@ -53,110 +51,125 @@
...
+            const uri = new URL(location.href);
+            const urlParams = uri.searchParams;
+            let url = urlParams.get('ept');
+
+            const options = {};
+            urlParams.keys().forEach(key => {
+                if (key !== 'ept') {
+                    options[key] = parseInt(urlParams.get(key), 10);
+                }
+            });
```
gchoqueux: URLSearchParams

examples/entwine_simple_loader.html
```diff
@@ -53,110 +51,125 @@
...
+                eptLayer.root.voxelOBB.box3D.getSize(size);
                 view.camera3D.far = 2.0 * size.length();
 
-                controls.groundLevel = eptLayer.root.bbox.min.z;
-                var position = eptLayer.root.bbox.min.clone().add(
-                    size.multiply({ x: 0, y: 0, z: size.x / size.z })
+                controls.groundLevel = eptLayer.minElevationRange;
+                var corner = new THREE.Vector3(...eptLayer.source.boundsConforming.slice(0, 3));
+                var position = corner.clone().add(
+                    size.multiply({ x: 0, y: 0, z: (size.x / size.z) })
```
gchoqueux, line 104: could use camera utils or refacto in libs examples

examples/entwine_simple_loader.html
```diff
@@ -53,110 +51,125 @@
...
+                debug.PointCloudDebug.initTools(view, eptLayer, debugGui)
+            }
 
-                    document.getElementById('share').innerHTML = '<a href="' +
+            function loadEPT(url, options) {
+                if(!url) return;
+                document.getElementById('share').innerHTML = '<a href="' +
                         location.href.replace(location.search, '') +
                         '?ept=' + url
                         + '" target="_blank">Link to share this view</a>';
```
gchoqueux, line 125: URLSearchParams

examples/entwine_simple_loader.html
```diff
@@ -53,110 +51,125 @@
...
+                    }
 
-                eptLayer.whenReady
-                    .then(() => debug.PointCloudDebug.initTools(view, eptLayer, debugGui));
+                    if(url.slice(-1) === '/') url = url.slice(0, -1);
 
-                function dblClickHandler(event) {
-                    var pick = view.pickObjectsAt(event, 5, eptLayer);
+                    let eptName = url.split('/').pop()
+                    eptName = eptName[0].toUpperCase() + eptName.slice(1);
```
gchoqueux, line 145: URLSearchParams

examples/entwine_simple_loader.html
```diff
@@ -53,110 +51,125 @@
...
-                }
-
-                eptName = url.split('/').pop()
-                eptName = eptName[0].toUpperCase() + eptName.slice(1);
-                eptLayer = new itowns.EntwinePointTileLayer(eptName, config);
+                };
 
-                view.addLayer(eptLayer).then(onLayerReady);
+                eptSource.whenReady.then(() => {
+                    setView(eptSource.crs);
```
gchoqueux: I don’t really see the point of adapting the CRS to the data.
It would be more interesting to be able to load multiple point clouds with different CRSs into a view of one’s choice.

examples/potree2_25d_map.html
gchoqueux, line 1: See previous comments

examples/potree2_25d_map_Arene.html
gchoqueux, line 1: See previous comments

examples/potree2_3d_map_Arene.html
gchoqueux, line 1: See previous comments

examples/potree_25d_map.html
gchoqueux, line 1: See previous comments

examples/potree_25d_map.html
```diff
@@ -65,26 +65,41 @@
...
+                // const name = 'eglise_saint_blaise_arles';
+                // const url = 'https://raw.githubusercontent.com/gmaillet/dataset/master/';
+                // const file = 'eglise_saint_blaise_arles.js';
+                // const crs = 'EPSG:4978';
+
+                // Most probaly a projection centered on the center of acquisition... 
+                // Probleme de CORS if 'Allow CORS' plugin off
+                // const url = 'https://cors-anywhere.herokuapp.com/https://viz.lib.vt.edu/thessaloniki/resources/pointclouds/thessaloniki';
+                // const file = 'cloud.js';
+                // const crs = 'EPSG:3857';// unknown
```
gchoqueux: to remove

examples/potree_3d_map.html
gchoqueux, line 1: See previous comments

examples/copc_simple_loader.html
```diff
@@ -152,18 +141,25 @@
...
             window.itowns = itowns;
             window.THREE = THREE;
 
+            function load() {
+                const url = document.getElementById('copc_url').value;
+                if (url) {
+                    setUrl(url);
+                }
+            }
+
```
gchoqueux, line 163: URLSearchParams ?

examples/copc_simple_loader.html
gchoqueux: I don’t really see the point of adapting the CRS to the data.
It would be more interesting to be able to load multiple point clouds with different CRSs into a view of one’s choice.

examples/entwine_simple_loader.html
gchoqueux, line 1: see previous comment for all examples

examples/potree_3d_map.html
```diff
@@ -65,29 +65,62 @@
...
+                var layer = new itowns.ColorLayer(config.id, config);
+                view.addLayer(layer);
+            });
+
+            // Donnees geocentrique -> probleme pour le placement camera
+            // name = 'eglise_saint_blaise_arles';
+            // const url = 'https://raw.githubusercontent.com/gmaillet/dataset/master/';
+            // const file = 'eglise_saint_blaise_arles.js';
+            // const crs = 'EPSG:4978';
+
```
gchoqueux: to remove

packages/Debug/src/PointCloudDebug.js
```diff
@@ -47,6 +49,122 @@ function setupControllerVisibily(gui, displayMode, sizeMode) {
...
 
+/**
+ * Generate the position array of the bbox corner form the bbox
+ * Adapted from THREE.BoxHelper.js
+ * https://github.com/mrdoob/three.js/blob/master/src/helpers/BoxHelper.js
+ *
+ * @param {THREE.box3} bbox - Box3 of the node
+ * @returns {array}
+ */
+function getCornerPosition(bbox) {
```
gchoqueux, line 60: could try re use three.js code

packages/Debug/src/PointCloudDebug.js
```diff
@@ -47,6 +49,122 @@ function setupControllerVisibily(gui, displayMode, sizeMode) {
...
+                node.promise.then(() => {
+                    if (node.obj) {
+                        tightboxHelper.geometry.attributes.position.array = getCornerPosition(node.obj.geometry.boundingBox);
+                        tightboxHelper.applyMatrix4(node.obj.matrixWorld);
+                        node.obj.tightboxHelper = tightboxHelper;
+                        helper.add(tightboxHelper);
+                        tightboxHelper.updateMatrixWorld(true);
+                    }
+                });
+            }
```
gchoqueux, line 127: in THREE, there's box helper for mesh ?

packages/Debug/src/PointCloudDebug.js
```diff
@@ -47,6 +49,122 @@ function setupControllerVisibily(gui, displayMode, sizeMode) {
...
+
+            node.link[layer.id] = helper;
+        } else {
+            node.link[layer.id].visible = true;
```
gchoqueux, line 131: could you use visible layer event to toggle helper visibility ?

packages/Main/src/Core/EntwinePointTileNode.js
```diff
@@ -104,6 +104,7 @@ class EntwinePointTileNode extends PointCloudNode {
 
         if (typeof numPoints == 'number') {
             const child = new EntwinePointTileNode(depth, x, y, z, this.source, numPoints);
+            child.crs = this.crs;
```
gchoqueux: why add crs in all children ?

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -31,42 +37,95 @@ class PointCloudNode extends THREE.EventDispatcher {
         throw new Error('In extended PointCloudNode, you have to implement the getter id!');
     }
 
+    // get the center of the node i.e. the center of the bounding box.
+    get center() {
+        if (this._center != undefined) { return this._center; }
+        const centerBbox = new THREE.Vector3();
+        this.voxelOBB.box3D.getCenter(centerBbox);
+        this._center =  new Coordinates(this.crs).setFromVector3(centerBbox.applyMatrix4(this.clampOBB.matrixWorld));
+        return this._center;
```
gchoqueux, line 47: could you remove on return ?

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -31,42 +37,95 @@ class PointCloudNode extends THREE.EventDispatcher {
...
+    }
+
+    // the origin is the center of the bounding box projected on the z=O local plan, in the world referential.
+    get origin() {
+        if (this._origin != undefined) { return this._origin; }
+        const centerCrsIn = proj4(this.crs, this.source.crs).forward(this.center);
+        this._origin =  new Coordinates(this.crs).setFromArray(proj4(this.source.crs, this.crs).forward([centerCrsIn.x, centerCrsIn.y, 0]));
+        return this._origin;
+    }
```
gchoqueux, line 56: * there's one return to many  
* use `Coordinates.as()`

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -31,42 +37,95 @@ class PointCloudNode extends THREE.EventDispatcher {
...
+     */
+    get rotation() {
+        if (this._rotation != undefined) { return this._rotation; }
+        this._rotation = new THREE.Quaternion();
+        if (proj4.defs(this.crs).projName === 'geocent') {
+            this._rotation = OrientationUtils.quaternionFromCRSToCRS(this.crs, this.source.crs)(this.origin);
+        }
+        return this._rotation;
+    }
+
```
gchoqueux, line 71: why add in `center`, `origin `and `rotation` in node?
I don't think it's the responsibility of the node to have transformations.
What are these properties used for?
Depending on their use, you can choose or assign them.

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -31,42 +37,95 @@ class PointCloudNode extends THREE.EventDispatcher {
...
         // use the size computed above to set the max
-        node.bbox.max.copy(node.bbox.min).add(size);
+        childVoxelBBox.max.copy(childVoxelBBox.min).add(size);
+
+        // get a clamped bbox from the voxel bbox
+        childNode.clampOBB.copy(childNode.voxelOBB);
+
+        const childClampBBox = childNode.clampOBB.box3D;
+
+        if (childClampBBox.min.z < this.source.zmax) {
```
gchoqueux, line 190: there's a clamp method in Box3 ? (or add in OBB)
if yes could use it or add it

packages/Main/src/Core/PotreeNode.js
```diff
@@ -6,6 +6,46 @@ import PointCloudNode from 'Core/PointCloudNode';
...
+    } else if (childIndex === 6) {
+        childVoxelBBox.min.y += dHalfLength.y;
+        childVoxelBBox.max.y += dHalfLength.y;
+        childVoxelBBox.min.x += dHalfLength.x;
+        childVoxelBBox.max.x += dHalfLength.x;
+    }
+
+    return childVoxelBBox;
+}
+
```
gchoqueux, line 48: may be add class `voxel`, with all methods

packages/Main/src/Core/PotreeNode.js
```diff
@@ -36,53 +76,39 @@ class PotreeNode extends PointCloudNode {
...
         if (!this.octreeIsLoaded) {
             this.loadOctree();
         }
-        return super.load();
+        // to refacto : can we use node instead of layer in options.out ?
+        return this.source.fetcher(this.url, this.source.networkOptions)
+            .then(file => this.source.parse(file, {
+                in: this,
+            }));
     }
```
gchoqueux: Why change this method?

packages/Main/src/Core/PotreeNode.js
```diff
@@ -36,53 +76,39 @@ class PotreeNode extends PointCloudNode {
...
 
     loadOctree() {
+        // To refacto methode only used for the root node...
+        this.offsetBBox = new THREE.Box3().setFromArray(this.source.boundsConforming);// Only for Potree1
```
gchoqueux: to refacto where?

packages/Main/src/Core/PotreeNode.js
```diff
@@ -103,7 +129,10 @@ class PotreeNode extends PointCloudNode {
                         const childrenBitField = view.getUint8(offset); offset += 1;
                         const numPoints = view.getUint32(offset, true) || this.numPoints; offset += 4;
                         const child = new PotreeNode(numPoints, childrenBitField, this.source);
+                        child.crs = this.crs;
```
gchoqueux: same question why child.crs ?

packages/Main/src/Layer/EntwinePointTileLayer.js
```diff
@@ -53,18 +44,15 @@ class EntwinePointTileLayer extends PointCloudLayer {
...
 
-            this.root.bbox.min.fromArray(this.source.boundsConforming, 0);
-            this.root.bbox.max.fromArray(this.source.boundsConforming, 3);
-            this.minElevationRange = this.minElevationRange ?? this.source.boundsConforming[2];
-            this.maxElevationRange = this.maxElevationRange ?? this.source.boundsConforming[5];
+            this.root = new EntwinePointTileNode(0, 0, 0, 0, this.source, -1);
+            this.root.crs = this.crs;
 
-            this.extent = Extent.fromBox3(config.crs || 'EPSG:4326', this.root.bbox);
+            this.setRootBbox(this.source.bounds.slice(0, 3), this.source.bounds.slice(3, 6));
```
gchoqueux: Why remove `extent `?
with extent you use camera utils to place camera

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -175,10 +156,14 @@ class PointCloudLayer extends GeometryLayer {
         this.protocol = 'pointcloud';
 
         this.group = group;
+        this.group.name = 'points';
         this.object3d.add(this.group);
-        this.bboxes = bboxes || new THREE.Group();
+
+        this.bboxes = bboxes;
+        this.bboxes.name = 'bboxes';
```
gchoqueux, line 161: why add name ?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -228,6 +213,78 @@ class PointCloudLayer extends GeometryLayer {
         this.root = undefined;
     }
 
+    setRootBbox(min, max) {
+        let forward = (x => x);
+        if (this.source.crs !== this.crs) {
+            try {
+                forward = proj4(this.source.crs, this.crs).forward;
```
gchoqueux: why add use proj4 ?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -228,6 +213,78 @@ class PointCloudLayer extends GeometryLayer {
...
+        // get center of box at altitude Z=0 and project it in view crs;
+        const origin = forward([(min[0] + max[0]) * 0.5, (min[1] + max[1]) * 0.5, 0]);
+
+        // get LocalRotation
+        const isGeocentric = proj4.defs(this.crs).projName === 'geocent';
+        let rotation = new THREE.Quaternion();
+        if (isGeocentric) {
+            const coordOrigin = new Coordinates(this.crs).setFromArray(origin);
+            rotation = OrientationUtils.quaternionFromCRSToCRS(this.crs, this.source.crs)(coordOrigin);
+        }
```
gchoqueux: I think this part can be refactored into the geographic package.

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -228,6 +213,78 @@ class PointCloudLayer extends GeometryLayer {
...
+
+        // get the bbox containing all cornersLocal => the bboxLocal
+        this.root.voxelOBB.box3D.setFromArray(cornersLocal);
+        this.root.voxelOBB.position.set(...origin);
+        this.root.voxelOBB.quaternion.copy(rotation).invert();
+
+        this.root.voxelOBB.updateMatrix();
+        this.root.voxelOBB.updateMatrixWorld(true);
+
+        this.root.voxelOBB.matrixWorldInverse = this.root.voxelOBB.matrixWorld.clone().invert();
```
gchoqueux: mayebe move to OBB.Method
For example, by adding a `.as(crs)` method that transforms an OBB to another CRS?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -228,6 +213,78 @@ class PointCloudLayer extends GeometryLayer {
...
+
+        this.root.clampOBB.copy(this.root.voxelOBB);
+
+        const clampBBox = this.root.clampOBB.box3D;
+        if (clampBBox.min.z < this.source.zmax) {
+            clampBBox.max.z = Math.min(clampBBox.max.z, this.source.zmax);
+        }
+        if (clampBBox.max.z > this.source.zmin) {
+            clampBBox.min.z = Math.max(clampBBox.min.z, this.source.zmin);
+        }
```
gchoqueux: add clamp method ?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -287,35 +344,39 @@ class PointCloudLayer extends GeometryLayer {
...
+        if (elt.obj) {
+            const obj = elt.obj;
+            bbox = obj.geometry.boundingBox;
+            matrixWorld = obj.matrixWorld;
+            matrixWorldInverse = obj.matrixWorldInverse;
+        } else {
+            bbox = elt.clampOBB.box3D;
+            matrixWorld = elt.clampOBB.matrixWorld;
+            matrixWorldInverse = elt.clampOBB.matrixWorldInverse;
+        }
```
gchoqueux: could be simplify ?

packages/Main/src/Layer/Potree2Layer.js
```diff
@@ -84,12 +78,10 @@ class Potree2Layer extends PointCloudLayer {
          * @type {boolean}
          * @readonly
          */
-        this.isPotreeLayer = true;
+        this.isPotree2Layer = true;
```
gchoqueux, line 81: may use `version `property ?

packages/Main/src/Layer/Potree2Layer.js
```diff
@@ -98,27 +90,18 @@ class Potree2Layer extends PointCloudLayer {
                 this.material.defines[normal.name] = 1;
             }
 
-            const min = new THREE.Vector3(...metadata.boundingBox.min);
-            const max = new THREE.Vector3(...metadata.boundingBox.max);
-            const boundingBox = new THREE.Box3(min, max);
+            this.setElevationRange();
```
gchoqueux, line 93: it's a new method ?

packages/Main/src/Layer/Potree2Layer.js
gchoqueux: why ?
why not `this.obb.set(min, max);`

packages/Main/src/Loader/LASLoader.js
gchoqueux: same question ?

packages/Main/src/Loader/LASLoader.js
```diff
@@ -47,7 +79,13 @@ class LASLoader {
     }
 
     _parseView(view, options) {
-        const colorDepth = options.colorDepth ?? 16;
+        const colorDepth = options.colorDepth ?? defaultColorEncoding(options.header);
+
+        const forward = (options.in.crs !== options.out.crs) ?
+            proj4(options.in.projDefs, options.out.projDefs).forward :
```
gchoqueux, line 85: could you refacto in geographic package ?

packages/Main/src/Loader/LASLoader.js
gchoqueux: same question ?

packages/Main/src/Loader/LASLoader.js
gchoqueux: same question ?

packages/Main/src/Loader/Potree2Loader.js
```diff
@@ -48,11 +49,51 @@ const typedArrayMapping = {
...
+
+    const res = [];
+    // v + q.w * t + cross( q.xyz, t );
+    res[0] = vx + qw * tx + qy * tz - qz * ty;
+    res[1] = vy + qw * ty + qz * tx - qx * tz;
+    res[2] = vz + qw * tz + qx * ty - qy * tx;
+
+    return res;
+}
+
```
gchoqueux, line 82: same method in LASLoader.js ?

packages/Main/src/Loader/Potree2Loader.js
```diff
@@ -87,19 +128,31 @@ export default function load(buffer, options) {
             for (let j = 0; j < numPoints; j++) {
                 const pointOffset = j * bytesPerPoint;
 
-                const x = (view.getInt32(pointOffset + attributeOffset + 0, true) * scale[0]) + offset[0] - min.x;
-                const y = (view.getInt32(pointOffset + attributeOffset + 4, true) * scale[1]) + offset[1] - min.y;
-                const z = (view.getInt32(pointOffset + attributeOffset + 8, true) * scale[2]) + offset[2] - min.z;
+                const point = [
```
gchoqueux, line 131: could you use Vector3 ?

packages/Main/src/Parser/PotreeBinParser.js
```diff
@@ -79,30 +106,72 @@ export default {
...
+
+        const forward = (source.crs !== options.in.crs) ?
+            proj4(source.crs, options.in.crs).forward :
+            (x => x);
+        const applyQuaternion = (source.crs !== options.in.crs) ?
+            _applyQuaternion : (x => x);
+
+        const origin = options.in.origin.toArray();
+        const quaternion = options.in.rotation.toArray();
+
```
gchoqueux, line 124: could you refacto the common part in all parsers?

packages/Main/src/Parser/PotreeBinParser.js
```diff
@@ -79,30 +106,72 @@ export default {
...
+                    const position2 = applyQuaternion([
+                        x - origin[0],
+                        y - origin[1],
+                        z - origin[2],
+                    ], quaternion);
+
+                    array[arrayOffset + 0] = position2[0];
+                    array[arrayOffset + 1] = position2[1];
+                    array[arrayOffset + 2] = position2[2];
```
gchoqueux, line 161: could you refacto the common part in all parsers?

packages/Main/src/Parser/PotreeBinParser.js
```diff
@@ -79,30 +106,72 @@ export default {
...
                 attrOffset += pointByteSize;
             }
             elemOffset += attr.byteSize;
             attrOffset = elemOffset;
+
             geometry.setAttribute(attr.attributeName, new THREE.BufferAttribute(array, attr.numElements, attr.normalized));
         }
 
+        geometry.userData.origin = options.in.origin;
+        geometry.userData.rotation = options.in.rotation;
```
gchoqueux, line 173: it's mesh transformation ?

gchoqueux (changes requested):
~~The title don't explain the feature  : `geographic reprojection for point clouds on load`~~

examples/copc_simple_loader.html
gchoqueux: this option doesn't exist, sorry

examples/copc_simple_loader.html
gchoqueux: you could use 
```js
view.referenceCrs = crs;
view.camera.crs = crs;
```

gchoqueux: don't forget to dispose layer

packages/Main/src/Loader/LASLoader.js
gchoqueux: Proposal to re-use three.js code 

```js
        const origin = new Vector3().fromArray(options.out.origin);
        const vector = new Vector3();
        const box = new Box3();

        const quaternion = new Quaternion().fromArray(options.out.rotation);
        for (let i = 0; i < view.pointCount; i++) {
            const point = getPosition.map(f => f(i));

            vector.fromArray(forward(point));

            vector.sub(origin);

            vector.applyQuaternion(quaternion);

            vector.toArray(positions, i);

            box.expandByPoint(vector);

            // ...

        return {
           // ...
           // serialization with Box3
            bbox: box.tojson(),
        };

```

packages/Main/src/Loader/LASLoader.js
gchoqueux: ok, header contains some options for parsing

examples/potree_25d_map.html
gchoqueux: use `/* ... */` and explains the comment

packages/Debug/src/PointCloudDebug.js
gchoqueux: or move method on specific `Class`

packages/Debug/src/PointCloudDebug.js
gchoqueux: or add method in `OBBHelper`

packages/Debug/src/PointCloudDebug.js
gchoqueux: or add getter property to return node visibility

packages/Main/src/Core/EntwinePointTileNode.js
gchoqueux: assumes new crs property in extented class from  `PointCloudNode`

packages/Main/src/Core/PointCloudNode.js
gchoqueux:
```js
if (this._center === undefined) {
      const centerBbox = new THREE.Vector3();
            this.voxelOBB.box3D.getCenter(centerBbox);
            this._center =  new Coordinates(this.crs).setFromVector3(centerBbox.applyMatrix4(this.clampOBB.matrixWorld));
        }
return this._center;
```

packages/Main/src/Core/PointCloudNode.js
gchoqueux: you could move in itowns/geographic

packages/Main/src/Layer/PointCloudLayer.js
gchoqueux: for debugging ?

packages/Main/src/Layer/PointCloudLayer.js
gchoqueux: see other comment

packages/Main/src/Layer/PointCloudLayer.js
gchoqueux: verify the transformation between elt, elt.obj, and and clampOBB?

examples/copc_simple_loader.html
ftoromanoff: We will need to create a Extent.fromOBB() method.
Let's do that in another PR !

examples/copc_simple_loader.html
ftoromanoff: Done

packages/Main/src/Layer/Potree2Layer.js
ftoromanoff: To be develop. But the question is deeper than just that..
Currently the layer create a root node and use one of it's own method to set the obbes of this node. (which can indeed be done by a OBB.set(min, max, crsIn, crsOut) method.
We then call the node method loadOctree on this root to create the children nodes.
Should we create an other RootNode class or let the layer call theses 2 differents methods ?

packages/Main/src/Loader/LASLoader.js
ftoromanoff: Using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517

packages/Main/src/Loader/LASLoader.js
ftoromanoff: Using THREE in LASLoader ?

examples/copc_3d_loader.html
ftoromanoff: Done

examples/copc_simple_loader.html
ftoromanoff: done

examples/copc_simple_loader.html
ftoromanoff: It's not a url search param here, but the creation of a new url that will be used for a new window.

examples/entwine_3d_loader.html
ftoromanoff: done

examples/entwine_3d_loader.html
ftoromanoff: Already discussed.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478106293

examples/entwine_3d_loader.html
ftoromanoff: Already discussed.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478363302

examples/entwine_simple_loader.html
ftoromanoff: done

examples/entwine_simple_loader.html
ftoromanoff: Already discussed.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478106293

examples/entwine_simple_loader.html
ftoromanoff: Already discussed.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478363302

examples/entwine_simple_loader.html
ftoromanoff: The url here is not the windows url but the ept url (that will be used in the source).
Maybe I should rename the variable to avoid confusion between the 2....

examples/entwine_simple_loader.html
ftoromanoff: I changed it as discussed in 
https://github.com/iTowns/itowns/pull/2498#discussion_r2468881917

examples/potree_25d_map.html
ftoromanoff: done

examples/copc_simple_loader.html
ftoromanoff: Already discussed.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478387408

examples/potree_3d_map.html
ftoromanoff: done using /*   */

packages/Debug/src/PointCloudDebug.js
ftoromanoff: I move all the debug from PointCloudLayer to PointCloudLayer.
I propose that the rework on that to be done in another PR.

packages/Debug/src/PointCloudDebug.js
ftoromanoff: See previous.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478421061

packages/Debug/src/PointCloudDebug.js
ftoromanoff: See previous.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478421061

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: Done. But it can still be discussed linked with the discussion: https://github.com/iTowns/itowns/pull/2498#discussion_r2478340355
Should the crs only be know by the layer (if it has all methods using it, or by the node)

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: It's a question about break point or nested code. It's all about code lisibility and preferences I would say ?.

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: 1st point :https://github.com/iTowns/itowns/pull/2498#discussion_r2478542261
2nd point: I suggest to open a new PR to move that indeed

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: Let's try to do that in another PR !

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: There is no clamp method form Box3 as indeed it will be hard to define what exactly you should clamp (maybe aclamp by Plane might be okey but even then the plan should be parralle to one of the side of the Box3.

In our case we just need to clamp regarding z, thus we could develop our own method. Moreover for an OBB

packages/Main/src/Core/PotreeNode.js
ftoromanoff: This method should be factorized with computeChildBBox from pointCloudNode, but it's a hard work ^^
I don't think a class for voxel is really needed..

packages/Main/src/Core/PotreeNode.js
ftoromanoff: Indeed, I was doing some refacto that i had to reimplement the method here but finally it was not needed. But it allowed me to rethink it and refactor it even more, and only use PointCloudNode.load() in Potree2Node and PotreeNode.

packages/Main/src/Core/PotreeNode.js
ftoromanoff: The idea would be to have special class for rootNode (or any other way) as currently we have some classes Node with methode that are only use once for the root node.
Should we move the loadOctree() method to layer (as the setRootBBox) ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2478340355

packages/Main/src/Core/PotreeNode.js
ftoromanoff: Corrected.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478487534

packages/Main/src/Layer/EntwinePointTileLayer.js
ftoromanoff: Already discussed
https://github.com/iTowns/itowns/pull/2498#discussion_r2478106293

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: The property name is by default created with a value ''. So I decided to use it with the real name, Make it easier to follow for the debugging indeed.

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: We should set up a new PR to create and move dedicated method in the geographic package, that we will use here and at other part of the code.
I propose to do that in another PR.

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: https://github.com/iTowns/itowns/pull/2498#discussion_r2478973192

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: This should be amix of OBB.method and CRS ? Let's see during the new PR moving dedicated function to geographic.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478973192

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: PR on OBB methods.
https://github.com/iTowns/itowns/pull/2498#discussion_r2478106293

packages/Main/src/Layer/Potree2Layer.js
ftoromanoff: Yes, that is used in all PoinCloudLayer for factorization.

packages/Main/src/Loader/Potree2Loader.js
ftoromanoff: PR using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517
We should be able to factorize that methode or as you suggest use directly THREE in here, that I choose not to do, maybe not for the best reasons...

packages/Main/src/Loader/Potree2Loader.js
ftoromanoff: PR using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517

packages/Main/src/Parser/PotreeBinParser.js
ftoromanoff: PR using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517

packages/Main/src/Parser/PotreeBinParser.js
ftoromanoff: PR using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517

packages/Main/src/Parser/PotreeBinParser.js
ftoromanoff: PR using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517

This is commun at all parser.

packages/Main/src/Loader/LASLoader.js
ftoromanoff: PR using THREE in LASLoader ?
https://github.com/iTowns/itowns/pull/2498#discussion_r2470179517

examples/copc_simple_loader.html
gchoqueux: it's removed ?

examples/copc_simple_loader.html
gchoqueux: the removed PointLayer.extent is a potential breaking change

packages/Main/src/Layer/Potree2Layer.js
gchoqueux: it's not very complicated to calculate the extent and this will simplify the examples and avoid the breaking change.

packages/Main/src/Loader/LASLoader.js
gchoqueux: yes

packages/Main/src/Loader/LASLoader.js
gchoqueux: yes

examples/copc_simple_loader.html
gchoqueux: it's URLSearchParams

examples/entwine_simple_loader.html
gchoqueux: it's URLSearchParams

examples/copc_simple_loader.html
gchoqueux: ok, it's not URLSearchParams

packages/Debug/src/PointCloudDebug.js
gchoqueux: Create a issue to use OBBHelper

packages/Debug/src/PointCloudDebug.js
gchoqueux: open a issue

packages/Main/src/Core/PointCloudNode.js
gchoqueux: no use Coordinates.as()

packages/Main/src/Layer/EntwinePointTileLayer.js
gchoqueux: it's simple to compute the extent to avoid a breaking change

packages/Main/src/Layer/PointCloudLayer.js
gchoqueux: I need to review the code another time

packages/Main/src/Layer/PointCloudLayer.js
gchoqueux:
```js
let bbox;
let object3d;
if (elt.obj) {
    object3d = elt.obj;
    bbox = obj.geometry.boundingBox;
} else {
    object3d = elt.clampOBB;
    bbox = object3d.box3D;
}

const { matrixWorld, matrixWorldInverse } = object3d;
```

ftoromanoff: Done

gchoqueux (approved):

Issue #2229, opened 2023-11-20
[Proposal - WebXR controllers] Add WebXR controllers interaction


jogarnier:
This issue is a **feature proposal**. Feel free to upvote (with :+1: ), comment and provide your use-cases if you're interested by this feature.

## Context
(Following the TODOs of https://github.com/iTowns/itowns/pull/2092)

- add VR controllers
- add VR visual tool for debugging

![image](https://github.com/iTowns/itowns/assets/111570217/6c180952-6ac7-4197-aa8a-cf51a01cf3c5)

## Expected behavior
- Add WebXR interaction with a GIS scene / 3D object / layer.
- The proposal is to cover the public API of WebXR (see details below) :
  - [x] Move via the right joystick
  - [x] Rotate via the left joystick
  - [x] teleportation via the right button
  - [x] Height adjustement via the left button

## Proposal
Add a demo implementation of WebXR controllers. The perimeter of this example is exclusive to VR and its bounds are yet to define (see problems part).

### Defining Itowns controllers API
Registering and using WebXR controllers can be factored to let Itown create and bind technically the instances.
There are 2 types of controllers :
- Hands detection (using ThreeJs XRHandModelFactory)
- controllers (using ThreeJs XRControllerModelFactory)

Tests and PoC has been done on the latter one.
Controller binding is not guaranteed to be consistant from a model to another. In this context, it has been mainly tested on a Pico4.

### Add more WebXR parameters at instance creation 
- The webXR scene is defined using the behavior of [local](https://developer.mozilla.org/en-US/docs/Web/API/XRReferenceSpace#local) (Refering to https://developer.mozilla.org/en-US/docs/Web/API/XRReferenceSpace#reference_space_types). In this setup we can move freely into the scene as we can move physically at a scale 1:1.
-- **TODO** test [bounded-floor](https://developer.mozilla.org/en-US/docs/Web/API/XRReferenceSpace#bounded-floor) could also fit well for small local scene, like around a building.

- Added a callback to the webXR render loop for user rendering needs.

### Listening controllers modification
The exhaustive list of events that can be listened from controller :  
- https://www.w3.org/TR/webxr/#event-types
This covers one trigger button.

For this demo purpose, I added a few more that are fired by itowns WebXR logic and listened by the user example : 
- itowns-xr-axes-changed (controller) 
- itowns-xr-axes-stop (controller) 
- itowns-xr-button-pressed (controller, buttonIndex, button) 
[...]

Issue #2428, opened 2024-10-07
Instanced 3D Tiles - Start Animation?


bertt:
Hi, how can I start an animation on Instanced 3D Tiles containing a windmill (blades are animated), 
sample   https://www.itowns-project.org/itowns/examples/3dtiles_loader.html?tileset=https://bertt.github.io/cesium_issues/i3dm_animating/issue/tileset.json

![image](https://github.com/user-attachments/assets/168c8d66-f450-46d1-984b-a06fed54e2be)

On Cesium it works see https://bertt.github.io/cesium_issues/i3dm_animating/issue/

![animated_windmill](https://github.com/user-attachments/assets/f0f29275-cc1a-4c20-a431-26f9b3db2ee2)

AnthonyGlt:
Apparently, we don't have any code to run the animations. They are accessible though. It looks like we are losing the name of the mesh when we are creating the instances from 3dtilesrenderer. So the animations can't hook to the mesh.
Also, I've tried to make [this animation](https://bertt.github.io/cesium_issues/i3dm_animating/issue//windmill.glb) works on iTowns, without success ([from your issue](https://community.cesium.com/t/animating-instanced-3d-tiles/34036))
So some modif should be done in the [external lib](https://github.com/NASA-AMMOS/3DTilesRendererJS) but also in iTowns


I've tried to use this to update the mixer:

```

  view.render = function render() {
                if(mixer){
                    mixer.update(clock.getDelta())
                
                }
                var g = view.mainLoop.gfxEngine;

                g.renderView(view);

};



```

Desplandis:
I had some doubts with the fact that `AnimationMixer` indeed supported instanced meshes, but it seems I was wrong! I managed to make it working in `three` with a simplified loader for the windmill glb file (i.e. `GLTFLoader` + traversal of the scene to convert each mesh to instanced meshes).

We have some work to do to make it working in `itowns`:
- As @AnthonyGlt inferred, we lost some information when converting a mesh to an instanced mesh in `3d-tiles-renderer-js`, leading to invalid references to meshes in animations. I will push a fix.
- Rework our render loop to continuously draw instead of stopping drawing when there is no update scheduled. We could however force update with the current API to test the fixe in `3d-tiles-renderer-js` and propose a first quick & dirty solution. I'll open a proposal.
- Provide out-of-the-box support for 3D-Tiles animations (à la Cesium), we would need to have an update/animate function on 3D layers?

[...]

PR #2500, opened 2025-02-13
Refactor WebXr with support  of vr controllers

AnthonyGlt:
## Description
Refactoring of the VR. Following the work done [previously](https://github.com/iTowns/itowns/pull/2230) and [here](https://github.com/iTowns/itowns/issues/2229).

Handling the controllers correctly. To do so, I've created `vrHeadSet` which is containing a clone of the camera (`_camXR`). The `vrHeadSet` will be used to translate the webxr scene around the globe. This is to bypass [this bug](https://discourse.threejs.org/t/staircase-effect-in-the-xr-controller-when-using-float34-coordinates/76584) which cause imprecision.
Then at each animation, I use the `_camXR` to update the position of our cam.
I can't directly use our default cam, because the camera needs a parent and it's position will also be affected by the position of the VR scene, which is usually set using this way:

```
const transform = new XRRigidTransform(offsetPosition, offsetRotation);
const teleportSpaceOffset = this.baseReferenceSpace.getOffsetReferenceSpace(transform);
this.parentRenderer.xr.setReferenceSpace(teleportSpaceOffset);
```

I also had to create callbacks in src/Layer/OGC3DTilesLayer.js to update the 3dtiles while being in VR.
[Related issue](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/213)

To activate the VR, we have to pass the `webXR` argument in the view option. To activate the controllers, we have to pass the argument `controllers` to true as follows:
```
{ webXR: { controllers: true } }
```

## Motivation and Context
Update the VR and get a better controllers support

image

[Screencast from 13-02-2025 14:25:53.webm](https://github.com/user-attachments/assets/6aed13a5-7dfc-4a77-a60a-bd82bd0a8750)

##TODO

- [ ] handle xr session end
- [ ] define webxr only in globelayer
- [ ] add more functionnalities (teleportation)
- [ ] better handling of the 3dtiles update/ view update

### DIFF
--- a/packages/Main/src/Controls/VRControls.js
+++ b/packages/Main/src/Controls/VRControls.js
@@ -1,0 +1,457 @@
+import * as THREE from 'three';
+import { Coordinates } from '@itowns/geographic';
+import DEMUtils from 'Utils/DEMUtils';
+import { XRControllerModelFactory } from 'ThreeExtended/webxr/XRControllerModelFactory';
+
+/**
+ * @property {Array} controllers - WebXR controllers list
+ * */
+class VRControls {
+    static MIN_DELTA_ALTITUDE = 1.8;
+    static MAX_NUMBER_CONTROLLERS = 2;  // For now, we are fully supporting a maximum of 2 controllers.
+    /**
+     * Requires a contextXR variable.
+     * @param {*} _view itowns view object
+     * @param {*} _groupXR XR 3D object group
+     */
+    constructor(_view, _groupXR = {}) {
+    // Store instance references.
+        this.view = _view;
+        this.groupXR = _groupXR;
+        this.webXRManager = _view.mainLoop.gfxEngine.renderer.xr;
+
+        this.rightButtonPressed = false;
+        this.controllers = [];
+        this.initControllers();
+    }
+
+    // Static factory method:
+    static init(view, vrHeadSet) {
+        return new VRControls(view, vrHeadSet);
+    }
+
+
+    initControllers() {
+        //  Add a light for the controllers
+        this.groupXR.add(new THREE.HemisphereLight(0xa5a5a5, 0x898989, 3));
+
+        const controllerModelFactory = new XRControllerModelFactory();
+
+        for (let i = 0; i < VRControls.MAX_NUMBER_CONTROLLERS; i++) {
+            const controller = this.webXRManager.getController(i);
+
+
+            controller.addEventListener('connected', (event) => {
+                controller.name = event.data.handedness;    // Left or right
+                controller.userData.handedness = event.data.handedness;
+                controller.gamepad = event.data.gamepad;
+                this.groupXR.add(controller);
+
+                const gripController = this.webXRManager.getControllerGrip(i);
+                gripController.name = `${controller.name}GripController`;
+                gripController.userData.handedness = event.data.handedness;
+                this.bindGripController(controllerModelFactory, gripController, this.groupXR);
+                this.controllers.push(controller);
+                this.groupXR.add(gripController);
+
+
+                // Event listeners
+                this.setupEventListeners(controller);
+            });
+
+            controller.addEventListener('disconnected', function removeCtrl() {
+                this.remove(this.children[0]);
+            });
+        }
+    }
+
+    bindGripController(controllerModelFactory, gripController, vrHeadSet) {
+        gripController.add(controllerModelFactory.createControllerModel(gripController));
+        vrHeadSet.add(gripController);
+    }
+
+
+
+
+    // Register event listeners for controllers.
+    setupEventListeners(controller) {
+        controller.addEventListener('itowns-xr-axes-changed', e => this.onAxisChanged(e));
+        controller.addEventListener('itowns-xr-axes-stop', e => this.onAxisStop(e));
+        controller.addEventListener('itowns-xr-button-pressed', e => this.onButtonPressed(e));
+        controller.addEventListener('itowns-xr-button-released', e => this.onButtonReleased(e));
+
+        controller.addEventListener('selectstart', e => this.onSelectStart(e));
+        controller.addEventListener('selectend', e => this.onSelectEnd(e));
+    }
+
+
+    /*
+Listening {XRInputSource} and emit changes for convenience user binding,
+There is NO JOYSTICK Events so we need to check it ourselves
+Adding a few internal states for reactivity
+- controller.isStickActive {boolean} true when a controller stick is not on initial state.
+*/
+
+    listenGamepad() {
+        for (const controller of this.controllers) {
+            if (!controller.gamepad) {
+                return;
+            }
+            // gamepad.axes = [0, 0, x, y];
+
+            const gamepad = controller.gamepad;
+            const activeValue = gamepad.axes.some(value => value !== 0);
+
+            // Handle stick activity state
+            if (controller.isStickActive && !activeValue && controller.gamepad.endGamePadtrackEmit) {
+                controller.dispatchEvent({
+                    type: 'itowns-xr-axes-stop',
+                    message: { controller },
+                });
+                controller.isStickActive = false;
+                return;
+            } else if (!controller.isStickActive && activeValue) {
+                controller.gamepad.endGamePadtrackEmit = false;
+                controller.isStickActive = true;
+            } else if (controller.isStickActive && !activeValue) {
+                controller.gamepad.endGamePadtrackEmit = true;
+            }
+
+            if (activeValue) {
+                controller.dispatchEvent({
+                    type: 'itowns-xr-axes-changed',
+                    message: { controller },
+                });
+            }
+
+            for (const [index, button] of gamepad.buttons.entries()) {
+                if (button.pressed) {
+                    // 0 - trigger
+                    // 1 - grip
+                    // 3 - stick pressed
+                    // 4 - bottom button
+                    // 5 - upper button
+                    controller.dispatchEvent({
+                        type: 'itowns-xr-button-pressed',
+                        message: {
+                            controller,
+                            buttonIndex: index,
+                            button,
+                        },
+                    });
+                    controller.lastButtonItem = button;
+                } else if (controller.lastButtonItem && controller.lastButtonItem === button) {
+                    controller.dispatchEvent({
+                        type: 'itowns-xr-button-released',
+                        message: {
+                            controller,
+                            buttonIndex: index,
+                            button,
+                        },
+                    });
+                    controller.lastButtonItem = undefined;
+                }
+
+                if (button.touched) {
+                    // triggered really often
+                }
+            }
+        }
+    }
+
+
+    // Clamp a translation to ground and then apply the transformation.
+    clampAndApplyTransformationToXR(trans, offsetRotation) {
+        const transClamped = this.clampToGround(trans);
+        this.applyTransformationToXR(transClamped, offsetRotation);
+    }
+
+    // Apply a translation and rotation to the XR group.
+    applyTransformationToXR(trans, offsetRotation) {
+        this.groupXR.position.copy(trans);
+        this.groupXR.quaternion.copy(offsetRotation);
+        this.groupXR.updateMatrixWorld(true);
+    }
+
+    /**
+   * Clamp the given translation vector so that the camera remains at or above ground level.
+   * @param {THREE.Vector3} trans - The translation vector.
+   * @returns {THREE.Vector3} The clamped coordinates as a Vector3.
+   */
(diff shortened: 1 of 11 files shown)


AnthonyGlt:
@Desplandis 
We could see about the animation process (especially for the 3Dtiles) since you're also working on the topic for https://github.com/iTowns/itowns/issues/2428

src/Controls/VRControls.js
```diff
@@ -0,0 +1,469 @@
...
+                // bindControllerListeners(controller, vrHeadSet);
+                controller.gamepad = event.data.gamepad;
+                this.groupXR.add(controller);
+
+                const gripController = this.webXRManager.getControllerGrip(i);
+                gripController.name = `${controller.name}GripController`;
+                gripController.userData.handedness = event.data.handedness;
+                this.bindGripController(controllerModelFactory, gripController, this.groupXR);
+                this.controllers.push(controller);
+                this.groupXR.add(controller);
```
AnthonyGlt:
```suggestion
                this.groupXR.add(gripController);
```

is it necessary ?

AnthonyGlt:
We should be able to access the vr controls from anywhere, should we not store the webxr instance directly as a member of View.js ?
this.webXR = ... (in View.js)

WDYT @Desplandis ?

packages/Main/src/Core/View.js
```diff
@@ -157,8 +157,11 @@ class View extends THREE.EventDispatcher {
      * a default one will be constructed. In this case, if options.renderer is an object, it will be used to
      * configure the renderer (see {@link c3DEngine}.  If not present, a new &lt;canvas> will be created and
      * added to viewerDiv (mutually exclusive with mainLoop)
+     * //TODO define webxr only in globelayer ?
```
Desplandis: If this is not currently working/tested on non-terrain or planar view, I would agree to only have the WebXR functionnality on GlobeView.

packages/Main/src/Core/View.js
```diff
@@ -157,8 +157,11 @@ class View extends THREE.EventDispatcher {
...
      * @param {Object} [options.webXR] - enable webxr button to switch on VR visualization.
-     * @param {number} [options.webXR.scale=1.0] - apply webxr scale tranformation.
+     * @param {number} [options.webXR.controllers=false] - enable vr controllers.
+     // * @param {number} [options.webXR.scale=1.0] - apply webxr scale tranformation.
```
Desplandis: I agree with you that changing the `scale` of the scene matrix to `1/scale` was kinda arbitrary. I suppose this was to keep the same "aspect" when switching back and forth between non-VR and VR modes?

Could you remove the options from the documentation?

packages/Main/src/Core/View.js
```diff
@@ -157,8 +157,11 @@ class View extends THREE.EventDispatcher {
...
+     @param {number} [options.webXR.callback] - rendering callback.
```
Desplandis: `options.webXR.callback` is not of type `number` but instead of type `Function` (more precisely of type `() => void`). I don't see where or why it is used. Any pointers?

packages/Main/src/Layer/OGC3DTilesLayer.js
```diff
@@ -313,6 +313,15 @@ class OGC3DTilesLayer extends GeometryLayer {
         if (config.sseThreshold) {
             this.sseThreshold = config.sseThreshold;
         }
+        // Used for custom schedule callbacks (VR)
+        this.tasks = [];
+
+        this.tilesSchedulingCB = (func) => {
+            this.tasks.push(func);
+        };
+        // // We set our scheduling callback for tiles downloading and parsing / important for VR https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/213
```
Desplandis: Okay, so it's caused by the fact that WebXR has its own `requestAnimationFrame` independent of the window-one (which may be paused by some implementations...). And the 3DTilesRenderer one is not called and the author want to schedule the tasks for the next window frame (which may never be called given the implementation).

Could you add some more description and/or link to the appropriate comment on the issue?
https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/213#issuecomment-947943386

We will have to be wary of changes on the 3DTilesRenderer `PriorityQueue` implementation.

packages/Main/src/Layer/OGC3DTilesLayer.js
```diff
@@ -436,9 +445,15 @@ class OGC3DTilesLayer extends GeometryLayer {
...
-
+    handleTasks() {
+        for (let t = 0, l = this.tasks.length; t < l; t++) {
+            this.tasks[t]();
+        }
+        this.tasks.length = 0;
+    }
     preUpdate(context) {
         this.scale = context.camera._preSSE;
+        this.handleTasks();
```
Desplandis, line 458: Since `preUpdate` is called per-frame, this does not break the invariant of 3DTilesRenderer.

packages/Main/src/Renderer/WebXR.js
```diff
@@ -1,67 +1,129 @@
 import * as THREE from 'three';
+import VRControls from 'Controls/VRControls';
+
+// TODO handle xr session end
```
Desplandis, line 4: Is there a reason that we do not handle the end of a WebXR session?

P.S.: This was not handled before this PR.

packages/Main/src/Renderer/WebXR.js
```diff
@@ -1,67 +1,129 @@
...
+        fov,
+    };
 }
 
-const initializeWebXR = (view, options) => {
-    const scale = options.scale || 1.0;
+class WebXR {
+    constructor(view, options) {
+        this.view = view;
+        this.options = options;
```
Desplandis, line 79: I would prefer to directly set `callback` as property instead of `options` which is only used with `options.callback` (easier to track attributes of the class).

packages/Main/src/Renderer/WebXR.js
```diff
@@ -1,67 +1,129 @@
...
+    // // Update the world transformation matrix, ensuring it reflects global transforms
+    view.camera3D.updateMatrixWorld(true);
+    view.notifyChange(view.camera3D, true);
+}
 
-async function shutdownXR(session) {
-    if (session) {
-        await session.end();
-    }
+function extractCameraAttributesFromProjectionMatrix(projectionMatrix) {
```
Desplandis, line 42: Correct since WebXR cameras are perspective cameras.

packages/Main/src/Controls/VRControls.js
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+
+    initControllers() {
+        //  Add a light for the controllers
+        this.groupXR.add(new THREE.HemisphereLight(0xa5a5a5, 0x898989, 3));
+
+        const controllerModelFactory = new XRControllerModelFactory();
+
+        for (let i = 0; i < 2; i++) {
```
Desplandis: Are we guaranteed that there is only 2 controllers on a given xr session? Or is this a limitation for now (who should be documented)? It seems that `three` creates a WebXRController so this should not crash but this is more for the end-user.

packages/Main/src/Controls/VRControls.js
```diff
@@ -0,0 +1,464 @@
...
+            }
+
+            if (activeValue) {
+                controller.dispatchEvent({
+                    type: 'itowns-xr-axes-changed',
+                    message: { controller },
+                });
+            }
+
+            for (const [index, button] of gamepad.buttons.entries()) {
```
Desplandis, line 127: No way to remap bindings for now, we should document it somewhere.

packages/Main/src/Controls/VRControls.js
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    // Apply a translation and rotation to the XR group.
+    applyTransformationToXR(trans, offsetRotation) {
+        if (!offsetRotation) {
+            console.error('missing rotation quaternion');
+            return;
+
+            // offsetRotation = this.groupXR.quaternion;
+        }
```
Desplandis: This should not be undefined, could you please remove this code?

packages/Main/src/Controls/VRControls.js
```diff
@@ -0,0 +1,464 @@
...
+
+        if (!trans) {
+            console.error('missing translation vector');
+            return;
+        }
```
Desplandis: This should not be undefined, could you please remove this code?

Desplandis (changes requested):
Hi and thanks for your PR. I do not have a VR headset, so I only managed to test it with the chrome extension. I have a few remarks, could you please address them?

packages/Main/src/Controls/VRControls.js
AnthonyGlt: Like you said, by calling `getController`, threejs returns a controller anyway. This controller could never be actually used, that's the reason why we wait for the `connected` event to process it.

Apparently there is no limits of how many controllers we can instantiate. I was only able to test with 0,1 and 2 controllers though...

I've chosen to loop on 2 controllers which is the most common use I think because it covers the 0,1,2 controllers.
We could put more or leave the user to chose the actual number. 
IMO, we should leave 2 and if an user needs more, we could see how to handle it then.

If we leave it like this, I will put a comment to document it

packages/Main/src/Renderer/WebXR.js
AnthonyGlt: As you said, it was not handle before also.
Ending a VR session means to have a button (on the screen or on the controller) to trigger the end of the session. I don't see a real use case for this. It's easier to reload the page. It could be an improvement that can be implemented if there is a need for that IMO

Desplandis: Yeah, let's leave it as a later improvement.

packages/Main/src/Controls/VRControls.js
Desplandis: Okay, let's just document it.

packages/Main/src/Controls/VRControls.js
AnthonyGlt: What do you mean by that ? We can not invert the value, for example
        // 4 - bottom button
                    // 5 - upper button
will always be like that.
But the user can access the class to override the different methods (onSelectStart, onButtonPressed etc)

packages/Main/src/Renderer/WebXR.js
AnthonyGlt: It's also used for:
```
if (this.options.controllers) {
                this.vrControls = new VRControls(this.view, vrHeadSet);
            }
            
 ```

It's not mandatory I guess, it's just a way of activating the controllers support
Should we just initialize the controllers everytime we enter in VR ? WDYT ?

File: packages/Main/src/Controls/VRControls.js (hunk 11 above)
Comment (Desplandis) [reply packages/Main/src/Controls/VRControls.js:127]: Okay so the buttons layout follow that of the [standard gamepad](https://w3c.github.io/gamepad/#remapping) of the Gamepad API? Nonetheless, let's leave it at that.

File: packages/Main/src/Renderer/WebXR.js (hunk 8 above)
Comment (Desplandis) [reply packages/Main/src/Renderer/WebXR.js:79]: I was talking about the callback parameter!

**Comment by AnthonyGlt** [comment]
I'm running in some issues with the unit test. It looks like the `getController` method isn't  existing in our testing context. I believe it's related to this [similar issue](https://github.com/iTowns/itowns/pull/2376)
I'll try to find a way. It should only impact the tests files so everything else in this PR is still valid

File: packages/Main/src/Renderer/WebXR.js (hunk 8 above)
Comment (AnthonyGlt) [reply packages/Main/src/Renderer/WebXR.js:79]: Like this ?

```suggestion
        this.options = options;
        this.renderCb = options.callback;    
```

AnthonyGlt: In GlobeView,
I'll also add:
            this.webXR = new WebXR(this, typeof options.webXR === 'boolean' ? {} : options.webXR);

Desplandis:
> I'm running in some issues with the unit test. It looks like the `getController` method isn't existing in our testing context. I believe it's related to this [similar issue](https://github.com/iTowns/itowns/pull/2376) I'll try to find a way. It should only impact the tests files so everything else in this PR is still valid

Okay, could you still remove the scale option documentation and move the WebXR initialisation up in GlobeView?

Desplandis (approved):

Desplandis:
Could you please fix the deepscan and commit message format issues?

Desplandis:
Moreover, could we wait for #2512 to merged?

AnthonyGlt:
@Desplandis added unit tests

Desplandis (approved):

PR #2504, opened 2025-02-28
VPC PointCloudData

ftoromanoff:
This PR aims at supporting the VPC format (vpc is a stack of ept or copc data)

New classes VpcSource and VpcLayer had been added. VpcSource beeing a class instanciating sub source (ept or copc source dependings).

The first try was to instanciate all sub sources as soon as the VpcSource was instanciate, but this lead to long delay at start.
Thus I decided to instanciate subsources only when the needed node was updated what leads to the use of mocked sources and promised sources.
(this is open for discussion)

~~(1er commit is not yet sqaushed as I feell like it's easier to check what was added for the mocked sources)~~

Other modifications:
- ~~A new id has been added to copcNode and eptNode (sId) to link the node and its subsource to ne used for the fetch.~~

- ~~The whole update of PointCloud had been copied and modified on vpcLayer, to avoid breaking the current PointCloud update, but at the end it probalby should be reunified.~~ DONE

### DIFF
--- a/examples/vpc_simple_loader.html
+++ b/examples/vpc_simple_loader.html
@@ -1,0 +1,168 @@
+<html>
+    <head>
+        <title>Itowns - Vpc simple loader</title>
+
+        <script type="importmap">
+            {
+                "imports": {
+                    "itowns": "../dist/itowns.js",
+                    "debug": "../dist/debug.js",
+                    "LoadingScreen": "./jsm/GUI/LoadingScreen.js",
+                    "itowns_widgets": "../dist/itowns_widgets.js",
+                    "three": "https://unpkg.com/three@0.170.0/build/three.module.js",
+                    "three/addons/": "https://unpkg.com/three@0.170.0/examples/jsm/",
+                    "lil": "https://unpkg.com/lil-gui@0.20.0/dist/lil-gui.esm.js"
+                }
+            }
+        </script>
+        <meta charset="UTF-8">
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <style type="text/css">
+            #description {
+                z-index: 2;
+                left: 10px;
+            }
+        </style>
+
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <script src="https://cdn.jsdelivr.net/npm/lil-gui@0.19"></script>
+    </head>
+    <body>
+        <div id="description">Specify the URL of a VPC file to load:
+            <input type="text" id="copc_url" />
+            <button id="readUrlBtn">Load</button>
+            <div>
+                <button id="loadAmiensBtn">Load Amiens VPC (copc stack)</button>
+                <button id="loadRouenBtn">Load Rouen VPC (copc stack)</button>
+                <button id="loadOrleansBtn">Load Orleans VPC (ept stack)</button>
+            <div id="share"></div>
+            </div>
+        </div>
+        <div id="viewerDiv"></div>
+
+        <script type="module">
+            import lil from 'lil';
+            import setupLoadingScreen from 'LoadingScreen';
+            import * as THREE from 'three';
+            import * as itowns from 'itowns';
+            import * as debug from 'debug';
+            import * as itowns_widgets from 'itowns_widgets';
+
+            var debugGui = new lil();
+            var viewerDiv = document.getElementById('viewerDiv');
+            viewerDiv.style.display = 'block';
+
+            function readUrl() {
+                const urlParams = new URL(location).searchParams;
+                const url = urlParams.get('vpc');
+
+                const options = {};
+                urlParams.keys().forEach(key => {
+                    if (key !== 'vpc') {
+                        options[key] = parseInt(urlParams.get(key), 10);
+                    }
+                });
+                loadVPC(url, options);
+            }
+
+            readUrl();
+
+            let view;
+            let controls;
+            function setView(crs) {
+                viewerDiv.innerHTML = '';
+                view = new itowns.View(crs, viewerDiv);
+                controls = new itowns.PlanarControls(view);
+                view.mainLoop.gfxEngine.renderer.setClearColor(0xdddddd);
+            }
+
+            var vpcLayer, vpcSource;
+
+            function onLayerReady() {
+                var lookAt = new THREE.Vector3();
+                var size = new THREE.Vector3();
+
+                const rootBbox = vpcLayer.root.children[0].bbox;
+                rootBbox.getSize(size);
+                rootBbox.getCenter(lookAt);
+
+                view.camera3D.far = 20.0 * size.length();
+
+                controls.groundLevel = rootBbox.min.z;
+                var position = rootBbox.min.clone().add(
+                    size.multiply({ x: 0, y: 0, z: size.x / size.z })
+                );
+
+                view.camera3D.position.copy(position);
+                view.camera3D.lookAt(lookAt);
+                view.camera3D.updateProjectionMatrix();
+
+                view.notifyChange(vpcLayer);
+
+                debug.PointCloudDebug.initTools(view, vpcLayer, debugGui);
+            }
+
+            function loadVPC(url, options = {}) {
+                if(!url) return;
+
+                if (vpcLayer) {
+                    vpcLayer.debugUI.destroy()
+                    view.removeLayer('VPC');
+                    view.notifyChange();
+                }
+
+                const vpcSource = new itowns.VpcSource({ url });
+
+                const config = {
+                    source: vpcSource,
+                    sseThreshold: 2,
+                    pointBudget: 3000000,
+                    ...options,
+                };
+
+                vpcSource.whenReady.then(() => {
+                    setView(vpcSource.crs);
+                    vpcLayer = new itowns.VpcLayer('VPC', config);
+                    view.addLayer(vpcLayer).then(onLayerReady);
+                })
+            }
+
+            function loadAmiens() {
+                const options = {
+                    material: {mode: 2},
+                    opacity: 0.9,
+                }
+                loadVPC('https://data.geopf.fr/chunk/telechargement/download/LiDARHD-NUALID/VPC/amiens.vpc', options);
+            }
+
+            function loadRouen() {
+                const options = {
+                    material: {mode: 2},
+                    opacity: 0.9,
+                }
+                loadVPC('https://data.geopf.fr/chunk/telechargement/download/LiDARHD-NUALID/VPC/rouen.vpc', options);
+            }
+
+            function loadOrleans() {
+                const options = {
+                    material: {mode: 2},
+                    opacity: 0.9,
+                }
+                loadVPC('https://data.geopf.fr/chunk/telechargement/download/lidarhd_fxx_ept/vpc/index.vpc', options);
+            }
+
+            const readUrlBtn = document.getElementById('readUrlBtn');
+            const loadAmiensBtn = document.getElementById('loadAmiensBtn');
+            const loadRouenBtn = document.getElementById('loadRouenBtn');
+            const loadOrleansBtn = document.getElementById('loadOrleansBtn');
+
+            readUrlBtn.addEventListener('click', readUrl);
+            loadAmiensBtn.addEventListener('click', loadAmiens);
+            loadRouenBtn.addEventListener('click', loadRouen);
+            loadOrleansBtn.addEventListener('click', loadOrleans);
+
+        </script>
+    </body>
+</html>
--- a/packages/Main/src/Layer/CopcLayer.js
+++ b/packages/Main/src/Layer/CopcLayer.js
@@ -37,14 +37,14 @@ class CopcLayer extends PointCloudLayer {
          */
         this.isCopcLayer = true;
 
-        const resolve = () => this;
+        const resolve = super.addInitializationStep();
         this.whenReady = this.source.whenReady.then((/** @type {CopcSource} */ source) => {
             const { cube } = source.info;
             const { pageOffset, pageLength } = source.info.rootHierarchyPage;
 
-            this.root = new CopcNode(0, 0, 0, 0, pageOffset, pageLength, this.source, -1);
-            this.root.bbox.min.fromArray(cube, 0);
-            this.root.bbox.max.fromArray(cube, 3);
+            const root = new CopcNode(0, 0, 0, 0, pageOffset, pageLength, this.source, -1);
+            root.bbox.min.fromArray(cube, 0);
+            root.bbox.max.fromArray(cube, 3);
 
             this.minElevationRange = this.minElevationRange ?? source.header.min[2];
             this.maxElevationRange = this.maxElevationRange ?? source.header.max[2];
@@ -52,7 +52,9 @@ class CopcLayer extends PointCloudLayer {
             this.scale = new THREE.Vector3(1.0, 1.0, 1.0);
             this.offset = new THREE.Vector3(0.0, 0.0, 0.0);
 
-            return this.root.loadOctree().then(resolve);
+            this.root = root;
+
+            return root.loadOctree().then(resolve);
         });
     }
 }
--- a/packages/Main/src/Layer/PointCloudLayer.js
+++ b/packages/Main/src/Layer/PointCloudLayer.js
@@ -279,22 +279,21 @@ class PointCloudLayer extends GeometryLayer {
         return [this.root];
     }
 
-    update(context, layer, elt) {
-        elt.visible = false;
-
-        if (this.octreeDepthLimit >= 0 && this.octreeDepthLimit < elt.depth) {
-            markForDeletion(elt);
-            return;
-        }
-
-        // pick the best bounding box
-        const bbox = (elt.tightbbox ? elt.tightbbox : elt.bbox);
-        elt.visible = context.camera.isBox3Visible(bbox, this.object3d.matrixWorld);
-        if (!elt.visible) {
-            markForDeletion(elt);
-            return;
-        }
-
+    /**
+     * Load the data of a node.
+     * Send a promise to get the data (if not already sent)
+     * and add the result to the node when resolve.
+     * Check the visiblility of children to see if the need to be updated
+     * as well.
+     *
+     * @param {PointCloudNode} elt - The element (node) to load data.
+     * @param {Object} context - The context.
+     * @param {PointCloudLayer} layer - The layer on wich the node is attach.
+     * @param {THREE.Box3} bbox - bbox of the node.
+     *
+     * @return {pointCloudNode[]} The child nodes to update (if needed).
+     */
+    loadData(elt, context, layer, bbox) {
         elt.notVisibleSince = undefined;
         point.copy(context.camera.camera3D.position).sub(this.object3d.getWorldPosition(new THREE.Vector3()));
         point.applyQuaternion(this.object3d.getWorldQuaternion(new THREE.Quaternion()).invert());
@@ -353,8 +352,38 @@ class PointCloudLayer extends GeometryLayer {
                 for (const child of elt.children) {
                     markForDeletion(child);
                 }
+                return [];
             }
         }
+    }
+
+    /**
+     * Check if the node need to be rendered. In that case it call the
+     * node.loadData() on it.
+     *
+     * @param {Object} context - The context.
+     * @param {PointCloudLayer} layer - The layer on wich the node is attach.
+     * @param {PointCloudNode} elt - The element (node) to render.
+     *
+     * @return {pointCloudNode[]} The child nodes to update or [] if ther is none.
+     */
+    update(context, layer, elt) {
+        elt.visible = false;
+
+        if (this.octreeDepthLimit >= 0 && this.octreeDepthLimit < elt.depth) {
+            markForDeletion(elt);
+            return [];
+        }
+
+        // pick the best bounding box
+        const bbox = (elt.tightbbox ? elt.tightbbox : elt.bbox);
+        elt.visible = context.camera.isBox3Visible(bbox, this.object3d.matrixWorld);
+        if (!elt.visible) {
+            markForDeletion(elt);
+            return [];
+        }
+
+        return this.loadData(elt, context, layer, bbox);
     }
 
     postUpdate() {
--- a/packages/Main/src/Layer/VpcLayer.js
+++ b/packages/Main/src/Layer/VpcLayer.js
@@ -1,0 +1,125 @@
+import * as THREE from 'three';
+import PointCloudNode from 'Core/PointCloudNode';
+import CopcNode from 'Core/CopcNode';
+import EntwinePointTileNode from 'Core/EntwinePointTileNode';
+import PointCloudLayer from 'Layer/PointCloudLayer';
+
+function _instantiateRootNode(source) {
+    let root;
+    if (source.isCopcSource) {
+        const { info } = source;
+        const { pageOffset, pageLength } = info.rootHierarchyPage;
+        root = new CopcNode(0, 0, 0, 0, pageOffset, pageLength, source, -1);
+        root.bbox.min.fromArray(info.cube, 0);
+        root.bbox.max.fromArray(info.cube, 3);
+    } else if (source.isEntwinePointTileSource) {
+        root = new EntwinePointTileNode(0, 0, 0, 0, source, -1);
+        root.bbox.min.fromArray(source.boundsConforming, 0);
+        root.bbox.max.fromArray(source.boundsConforming, 3);
+    } else {
+        const msg = '[VPCLayer]: stack point cloud format not supporter';
+        console.warn(msg);
+        PointCloudLayer.handlingError(msg);
+    }
+    return root;
+}
+
+/**
+ * A layer for [Virtual Point Clouds](https://github.com/PDAL/wrench/blob/main/vpc-spec.md) (VPC) datasets.
+ * See {@link PointCloudLayer} class for documentation on base properties.
+ *
+ * @extends {PointCloudLayer}
+ *
+ * @property {boolean} isVpcLayer - Used to checkout whether this layer
+ * is a VpcLayer. Default is `true`. You should not change this, as it is
+ * used internally for optimisation.
+ *
+ *
+ *
+ * @example
+ * // Create a new VPC layer
+ * const vpcSource = new VpcSource({
+ *     url: 'https://data.geopf.fr/chunk/telechargement/download/lidarhd_fxx_ept/vpc/index.vpc',
+ * });
+ *
+ * const vpcLayer = new VpcLayer('VPC', {
+ *     source: vpcSource,
+ * });
+ *
+ * View.prototype.addLayer.call(view, vpcLayer);
+ */
+class VpcLayer extends PointCloudLayer {
+    /**
+     * @param {string} id - Unique id of the layer.
+     * @param {Object} config - See {@link PointCloudLayer} for base pointcloud
+     * options.
+     */
+    constructor(id, config) {
+        super(id, config);
+
+        /**
+         * @type {boolean}
+         * @readonly
+         */
+        this.isVpcLayer = true;
+
+        this.scale = new THREE.Vector3(1.0, 1.0, 1.0);
+        this.offset = new THREE.Vector3(0.0, 0.0, 0.0);
+
+        const resolve = this.addInitializationStep();
+
+        // a Vpc layer should be ready when all the child sources are
+        this.whenReady = this.source.whenReady.then((/** @type {VpcSource} */ sources) => {
+            this.minElevationRange = this.minElevationRange ?? this.source.minElevation;
+            this.maxElevationRange = this.maxElevationRange ?? this.source.maxElevation;
+
+            const boundsConforming = this.source.boundsConforming;
+            this.root = new PointCloudNode(0, this.source);
+            this.root.bbox.min.fromArray(boundsConforming, 0);
+            this.root.bbox.max.fromArray(boundsConforming, 3);
+            this.root.depth = 0;
+
+            sources.forEach((source, i) => {
+                const boundsConforming = source.boundsConforming;
+                const bbox = new THREE.Box3().setFromArray(boundsConforming);
+                const mockRoot = {
+                    bbox,
+                    children: [],
+                    waitingForSource: true,
+                    source,
+                };
+
+                // As we delayed the intanciation of the source to the moment we need to render a particular node,
+                // we need to wait for the source to be instantiate to be able
+                // to instantiate a node and load the Octree associated.
+                const promise =
+                    source.whenReady.then((src) => {
+                        const root = _instantiateRootNode(src);
+                        this.root.children[i] = root;
+                        return root.loadOctree().then(resolve)
+                            .then(() => root);
+                    });
+
+                mockRoot.loadOctree = promise;
+                // when load() is called on the mockRoot, we need the associated source to be loaded
+                // as well as the octree, before calling load() on the real root.
+                mockRoot.load = () => promise.then(root => root.load());
+                this.root.children.push(mockRoot);
+            });
+            this.ready = true;
+        });
+    }
+
+    loadData(elt, context, layer, bbox) {
+        if (elt.waitingForSource) {
+            layer.source.instantiate(elt.source);
+            elt.loadOctree
+                .then(eltLoaded => super.loadData(eltLoaded, context, layer, bbox))
+                .then(() => context.view.notifyChange(layer));
+        } else {
+            return super.loadData(elt, context, layer, bbox);
+        }
+    }
+}
+
+export default VpcLayer;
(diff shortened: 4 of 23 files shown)


examples/vpc_simple_loader.html
```diff
@@ -0,0 +1,88 @@
...
+                view.camera3D.position.copy(position);
+                view.camera3D.lookAt(lookAt);
+                view.camera3D.updateProjectionMatrix();
+
+                view.notifyChange(view.camera3D);
+            }
+
+            // VPC copc
+            // const url = 'https://storage.sbg.cloud.ovh.net/v1/AUTH_63234f509d6048bca3c9fd7928720ca1/ppk-lidar/amiens.vpc';
+            // const url = 'https://storage.sbg.cloud.ovh.net/v1/AUTH_63234f509d6048bca3c9fd7928720ca1/ppk-lidar/rouen.vpc';
```
jailln: What do you think of adding buttons to the example to switch between datasets and to add an input field so a user can add its own data sets ? Similar to what's done in 3dtiles_loader.html

packages/Main/src/Core/CopcNode.js
```diff
@@ -35,8 +35,9 @@ class CopcNode extends PointCloudNode {
      * @param {number} entryLength - Size of the node entry
      * @param {CopcLayer} layer - Parent COPC layer
      * @param {number} [numPoints=0] - Number of points given by this entry
+     * @param {number} [sId] - id for multisource
```
jailln: In case we keep the sId: Could you make the comment a bit more explicit please ? For instance: `ID of the source this node belongs to. Mainly used for VPCSource (multi-source)`.

packages/Main/src/Core/CopcNode.js
```diff
@@ -64,7 +67,11 @@ class CopcNode extends PointCloudNode {
      * @param {number} size
      */
     async _fetch(offset, size) {
-        return this.layer.source.fetcher(this.layer.source.url, {
+        let sourceUrl = this.layer.source.url;
+        if (this.layer.source.urls) {
+            sourceUrl = this.layer.source.urls[this.sId];
```
jailln: what do you think about always using the array `this.layer.source.urls` with only one url in some cases? It would require more modifications but the data model would be clearer I think.

packages/Main/src/Core/CopcNode.js
```diff
@@ -48,6 +49,8 @@ class CopcNode extends PointCloudNode {
         this.y = y;
         this.z = z;
 
+        this.sId = sId;
```
jailln: In case we keep the sId: In the same idea as my comment below about using this.layer.source.urls all the time, I think we should always define this `sId` (even for single sources) to be more consistent.

packages/Main/src/Core/CopcNode.js
```diff
@@ -106,6 +113,7 @@ class CopcNode extends PointCloudNode {
             node.findAndCreateChild(depth, x,     y + 1, z + 1, hierarchy, stack);
             node.findAndCreateChild(depth, x + 1, y + 1, z + 1, hierarchy, stack);
         }
+        return this;
```
jailln: why do we need to return `this` here ?

packages/Main/src/Core/CopcNode.js
```diff
@@ -162,9 +171,18 @@ class CopcNode extends PointCloudNode {
         }
 
         const buffer = await this._fetch(this.entryOffset, this.entryLength);
+        const sources = await this.layer.source.whenReady;
```
jailln: It doesn't come from this PR but I find it weird that the `COPCNode` knows the layer; it seems it should be the opposite. Also, we only use the source from the layer so maybe we could only pass the source ? And directly the correct source, which should therefore remove the need for a `sId`. WDYT ?

packages/Main/src/Core/CopcNode.js
```diff
@@ -162,9 +171,18 @@ class CopcNode extends PointCloudNode {
...
+        let source = this.layer.source;
+        if (sources.length > 1) {
+            if (sources[this.sId].isSource) {
+                source = await sources[this.sId].whenReady;
+            } else {
+                sources[this.sId].load();
+            }
+        }
```
jailln: I don't think this logic should be put here:
* I think it's not the responsability of the CopcNode to adapt its behaviour whether there are multiple sources or not
* Could you relocate this logic inside the source and only await for `source.whenReady` instead ?

packages/Main/src/Core/EntwinePointTileNode.js
```diff
@@ -48,19 +48,28 @@ class EntwinePointTileNode extends PointCloudNode {
      * @param {number} [numPoints=0] - The number of points in this node. If
      * `-1`, it means that the octree hierarchy associated to this node needs to
      * be loaded.
+     * @param {number} [sId] - id for multisource
```
jailln: same proposals than for `CopcNode`

packages/Main/src/Core/EntwinePointTileNode.js
```diff
@@ -48,19 +48,28 @@ class EntwinePointTileNode extends PointCloudNode {
...
 
         this.voxelKey = buildVoxelKey(depth, x, y, z);
 
-        this.url = `${this.layer.source.url}/ept-data/${this.voxelKey}.${this.layer.source.extension}`;
+        let sourceUrl = this.layer.source.url;
+        let sourceExtension = this.layer.source.extension;
+        if (this.layer.source.urls) {
+            sourceUrl = this.layer.source.sources[this.sId].url;
+            sourceExtension = this.layer.source.sources[this.sId].extension;
+        }
```
jailln: same proposals than for CopcNode

packages/Main/src/Core/EntwinePointTileNode.js
```diff
@@ -72,7 +81,11 @@ class EntwinePointTileNode extends PointCloudNode {
     }
 
     loadOctree() {
-        const hierarchyUrl = `${this.layer.source.url}/ept-hierarchy/${this.voxelKey}.json`;
+        let sourceUrl = this.layer.source.url;
+        if (this.layer.source.urls) {
+            sourceUrl = this.layer.source.sources[this.sId].url;
+        }
```
jailln: same proposals than for CopcNode

packages/Main/src/Core/EntwinePointTileNode.js
```diff
@@ -95,15 +108,30 @@ class EntwinePointTileNode extends PointCloudNode {
                 node.findAndCreateChild(depth, x,     y + 1, z + 1, hierarchy, stack);
                 node.findAndCreateChild(depth, x + 1, y + 1, z + 1, hierarchy, stack);
             }
+            return this;
```
jailln: why do we need to return this here ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
+import * as THREE from 'three';
+import CopcNode from 'Core/CopcNode';
+import EntwinePointTileNode from 'Core/EntwinePointTileNode';
+import PointCloudLayer from 'Layer/PointCloudLayer';
+
+
+// PointCLoudLayer functions
```
jailln: Can you factorize these functions please?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
...
+        elt.sse = -1;
+    }
+    for (const child of elt.children) {
+        markForDeletion(child);
+    }
+}
+
+/**
+ * A layer for [Cloud Optimised Point Cloud](https://copc.io) (COPC) datasets.
+ * See {@link PointCloudLayer} class for documentation on base properties.
```
jailln, line 29: Please update coomment according to the class

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
+import * as THREE from 'three';
```
jailln, line 1: Could you write it in typescript directly since it is a new file please ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
...
+ *     url: 'https://data.geopf.fr/chunk/telechargement/download/lidarhd_fxx_ept/vpc/index.vpc',
+ * });
+ *
+ * const vpcLayer = new VpcLayer('VPC', {
+ *     source: vpcSource,
+ * });
+ *
+ * View.prototype.addLayer.call(view, vpcLayer);
+ */
+class VpcLayer extends PointCloudLayer {
```
jailln, line 51: Can you also document class properties please ? (this.roots, this.spacing, etc.)

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
...
+
+        this.roots = [];
+        this.spacing = [];
+        this.scale = new THREE.Vector3(1.0, 1.0, 1.0);
+        this.offset = new THREE.Vector3(0.0, 0.0, 0.0);
+
+        const minElevationRanges = [];
+        const maxElevationRanges = [];
+
+        // const resolve = this.addInitializationStep();
```
jailln: to remove ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
...
+                    });
+                this.loadOctrees.push(promise);
+
+                root.load = () => this.loadOctrees[i].then(res => res.load());
+                this.roots.push(root);
+            });
+            this.ready = true;
+
+            return this.loadOctrees;
+        });
```
jailln, line 110: Can you reduce the complexity of this block by creating dedicated functions please ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    // adapted from PointCloudLayer, return are differents
```
jailln: A lot of shared code with PointCloudLayer, can yoou factorize it to ease maintenance later on please ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,341 @@
...
+        if (commonAncestor) {
+            return [commonAncestor];
+        }
+
+        // Start updating from hierarchy root
+        return this.roots;
+    }
+
+    // PointCloudLayer.update separate in 2 parts: update and subUpdate
+    subUpdate(elt, context, layer, bbox) {
```
jailln: Same for update and subUpdate: there is a lot of shared code with PointCloudLayer, can you factorize them please ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,97 @@
+import LASParser from 'Parser/LASParser';
```
jailln, line 2: Could you write it directly in ts please ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,97 @@
...
+ * @property {string} url - The URL of the directory containing the whole
+ * Entwine Point Tile structure.
+ */
+class VpcSource extends Source {
+    /**
+     * @param {Object} config - The configuration, see {@link Source} for
+     * available values.
+     * @param {number|string} [config.colorDepth='auto'] - Does the color
+     * encoding is known ? Is it `8` or `16` bits ? By default it is to
+     * `'auto'`, but it will be more performant if a specific value is set.
```
jailln: If I'm not wrong it is not set to `auto` by default but to `16` instead. 
Also, why would it be more performant ? Do you mean it might not work if not provided instead ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,97 @@
...
+
+        const urlVpc = this.url;
+        this.whenReady = Fetcher.json(urlVpc, this.networkOptions).then((meta) => {
+            this.urls = meta.features.map(f => f.assets.data.href);
+
+            const boundsConformings = meta.features.map(f => f.properties['proj:bbox']);
+            this.minElevation = Math.min(...boundsConformings.map(bC => bC[2]));
+            this.maxElevation = Math.max(...boundsConformings.map(bC => bC[5]));
+
+            /* FOR ONE proj:wkt2
```
jailln: to remove ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,97 @@
...
+
+            this.sources = [];
+            this.promises = [];
+            this.urls.forEach((url, i) => {
+                const p = new Promise((re, rj) => {
+                    this.promises.push({ resolve: re, reject: rj });
+                }).then((res) => {
+                    this.sources[i] = res;
+                    return res;
+                });
```
jailln, line 103: could you add a catch and output and error in the console please ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,97 @@
...
+
+            this.crs = projCS.title || projCS.name || 'EPSG:4326';
+            if (!(this.crs in proj4.defs)) {
+                proj4.defs(this.crs, projCS);
+            }
+            */
+
+            this.sources = [];
+            this.promises = [];
+            this.urls.forEach((url, i) => {
```
jailln: Can you put this and/or what's happening in the forEach inside dedicated functions to reduce cognitive complexity please ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,97 @@
...
+            return this.sources;
+        });
+    }
+
+    load(index) {
+        const url = this.urls[index];
+        let source;
+        if (url.includes('.copc')) {
+            source = new CopcSource({ url });
+        } else {
```
jailln: can you verify the extension is `.ept` here and log an error / warn for other extensions please ?

examples/vpc_simple_loader.html
```diff
@@ -0,0 +1,88 @@
...
+
+            // VPC ept
+            const url = 'https://data.geopf.fr/chunk/telechargement/download/lidarhd_fxx_ept/vpc/index.vpc';
```
jailln: it is correctly loaded but the point cloud is all white. Is there any chance to have a colored one ?

examples/vpc_simple_loader.html
ftoromanoff: I thougth about it, but as vpc was currently only a 'IGN' used format and for simplification I made a simple example, but I can definitely do as the other what you suggest (and that I did for the other PointClouds examples)

packages/Main/src/Core/CopcNode.js
ftoromanoff: Done

packages/Main/src/Core/CopcNode.js
ftoromanoff: In this PR I choose not to modified (in deepth) the current PointCloudNode. that the reason I chose not to only used array, but if this multi source model is accepted, I can indeed make this modification (to have only 1 format)

packages/Main/src/Core/CopcNode.js
ftoromanoff: Indeed, I added at some point but it's nio needed in the end...
I removed it.

packages/Main/src/Core/CopcNode.js
ftoromanoff: I opened a new PR to answer this:
#2603

packages/Main/src/Core/CopcNode.js
ftoromanoff: See after #2603

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: done

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: see anwer for copcNode (I removed that addition)

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: same answer as for CopcNode.
Let's discuss and decide

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: same

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: done

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: Done.

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: done

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: corrected. I manage to be able to keep the addInitializationStep()

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: I did a light simplification, I don't think we need dedicated functions. The best way would be to be able to get directly what's done in the corresponding node (but seems complicated

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: done

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: done

packages/Main/src/Source/VpcSource.js
ftoromanoff: This documentation was what it was before for copc. I did update the documentation on colorDepth for the 3 Layers.

packages/Main/src/Source/VpcSource.js
ftoromanoff: I keep it to remenber that it will need to be delat with when we will try to do the reprojection.

packages/Main/src/Source/VpcSource.js
ftoromanoff: done

packages/Main/src/Source/VpcSource.js
ftoromanoff: done

examples/vpc_simple_loader.html
ftoromanoff: You just need to change the Styling in the GUI... Par default it shows 'color' thus with this dataset white as there is no color encoded...

packages/Main/src/Source/VpcSource.js
jailln: after discussion with @Desplandis , we cannot write it in `ts` now since `Sources` are not in ts yet.

examples/vpc_simple_loader.html
ftoromanoff: Done in one of the last commits.

jailln: thanks !

packages/Main/src/Core/CopcNode.js
jailln: Thanks ! I think now we can remove the sId and directly pass in the right source, that would simplify the code. WDYT ?

packages/Main/src/Core/CopcNode.js
jailln: I think it would be clearer to store urls in `this.layer.source.url` and to always make `this.layer.source.url` an array and to do it right away in this PR (so we directly go off with the right data model). What do you think? Also @Desplandis I'd like your opinion on this ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,162 @@
...
+                        if (this.source.sources[i].isCopcSource) {
+                            const { info } = src;
+                            const { pageOffset, pageLength } = info.rootHierarchyPage;
+
+                            this.spacings[i] = info.spacing;
+
+                            secondaryRoot = new CopcNode(0, 0, 0, 0, pageOffset, pageLength, this, -1, i);
+                            secondaryRoot.bbox.min.fromArray(info.cube, 0);
+                            secondaryRoot.bbox.max.fromArray(info.cube, 3);
+                        } else {
```
jailln: Could you check if it is an `EntwinePointTileSource` here and add an else statement to handle other cases and log a warning for them please ? (will allow to detect badly formed vpc and ease debug)

packages/Main/src/Layer/VpcLayer.js
jailln: Thanks, but I still think it is too complicated: there is too much imbrication and the function is too long in my opinion. Someone that doesn't know this code will take a long time to understand it. After a quick look, you could for instance:
* create a function `_initSource` for what's done in `source.forEach` on line 84
* create another method `_getPointSourcePromise` for line 94.
* maybe create a method for the CopcSource and one for the EntwinePointTileSource cases in what's done in the then after line 94
* Also, the promises manipulation seems a bit too complicated and hard to understand here (especially on line 94, 118 and 120). Maybe it could be simplified ? 
You can also try to ask an AI assistant to refactor and simplify the promises manipulations here since they are quite good at such tasks ^^

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,119 @@
...
+ */
+class VpcSource extends Source {
+    /**
+     * @param {Object} config - The configuration, see {@link Source} for
+     * available values.
+     * @param {number} [config.colorDepth] - Color depth (in bits).
+     * Either 8 or 16 bits. By defaults it will be set to 8 bits for LAS 1.2 and
+     * 16 bits for later versions (as mandatory by the specification).
+     */
+    constructor(config) {
```
jailln, line 35: similarly to VpcLayer, this constructor is too complicated in my opinion and should be broken down in simpler and smaller functions to avoid complicated promises chains with imbricated modifications.

jailln:
I get the following CORS errors for the first two datasets of the example: 
image

examples/vpc_simple_loader.html
jailln: Yep, that was my point, since this is an example that externals can see, can we either:
* get a dataset with colors encoded
* or, directly choose a styling option that seems visually relevant ?

packages/Main/src/Core/CopcNode.js
ftoromanoff: By using the source instead of the layer we can completly remove the use of sId.

packages/Main/src/Core/CopcNode.js
ftoromanoff: Same as previous. Using source instead of the layer, make the use of the main layer.urls not needed.

packages/Main/src/Core/CopcNode.js
ftoromanoff: Same as previous. sId had been removed from node. (but kept in Vpc mockSource)

packages/Main/src/Core/CopcNode.js
ftoromanoff: Indeed ! That's what I did and it simplified the code (See previous comments). The nide won't know the VPC Source but only the copc or Ept source that they need.

packages/Main/src/Core/CopcNode.js
ftoromanoff: Back to what was before as there is no need to complicated with the use of the direct source.

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: Removed

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: Removed

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: Removed

examples/vpc_simple_loader.html
ftoromanoff: I will set the 'Styling' at classification (mode: 2) by default for these datasets

ftoromanoff:
> I get the following CORS errors for the first two datasets of the example: image

I will try to see if I can check that.

packages/Main/src/Layer/CopcLayer.js
```diff
@@ -37,22 +37,24 @@ class CopcLayer extends PointCloudLayer {
          */
         this.isCopcLayer = true;
 
-        const resolve = () => this;
+        const resolve = () => PointCloudLayer.prototype.addInitializationStep.call(this);
```
jailln: could be replaced by `super.addInitializationStep();`

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -8,7 +8,7 @@ const bboxMesh = new THREE.Mesh();
 const box3 = new THREE.Box3();
 bboxMesh.geometry.boundingBox = box3;
 
-function initBoundingBox(elt, layer) {
+export function initBoundingBox(elt, layer) {
```
jailln: why do you need to export it?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -53,7 +53,7 @@ function computeSSEOrthographic(context, pointSize, pointSpacing) {
     return Math.max(0.0, distance - pointSize);
 }
 
-function computeScreenSpaceError(context, pointSize, pointSpacing, distance) {
+export function computeScreenSpaceError(context, pointSize, pointSpacing, distance) {
```
jailln: why do you need to export it?

packages/Main/src/Layer/PointCloudLayer.js
```diff
@@ -279,22 +279,8 @@ class PointCloudLayer extends GeometryLayer {
...
-
-        // pick the best bounding box
-        const bbox = (elt.tightbbox ? elt.tightbbox : elt.bbox);
-        elt.visible = context.camera.isBox3Visible(bbox, this.object3d.matrixWorld);
-        if (!elt.visible) {
-            markForDeletion(elt);
-            return;
-        }
-
+    // PointCloudLayer.update separate in 2 parts: update and loadData
```
jailln: this comment could be improved: use jsdoc, explain the general algorithm on how data is loaded

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,155 @@
+import * as THREE from 'three';
+import PointCloudNode from 'Core/PointCloudNode';
+import CopcNode from 'Core/CopcNode';
+import EntwinePointTileNode from 'Core/EntwinePointTileNode';
+import PointCloudLayer from 'Layer/PointCloudLayer';
+
+function markForDeletion(elt) {
```
jailln: same as in PointCloudLayer. Could become a method of PointCloudLayer and used in this class to avoid code duplication.

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,155 @@
...
+        // Set .sse to an invalid value
+        elt.sse = -1;
+    }
+    for (const child of elt.children) {
+        markForDeletion(child);
+    }
+}
+
+function _intanciateNode(source) {
+    let root;
```
jailln, line 8: rename method to `_instanciateRootNode` or rename this variable to `node`, depending on whether it is speicific to the root node or not.

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,155 @@
...
+        const { pageOffset, pageLength } = info.rootHierarchyPage;
+        root = new CopcNode(0, 0, 0, 0, pageOffset, pageLength, source, -1);
+        root.bbox.min.fromArray(info.cube, 0);
+        root.bbox.max.fromArray(info.cube, 3);
+    } else if (source.isEntwinePointTileSource) {
+        root = new EntwinePointTileNode(0, 0, 0, 0, source, -1);
+        root.bbox.min.fromArray(source.boundsConforming, 0);
+        root.bbox.max.fromArray(source.boundsConforming, 3);
+    } else {
+        const msg = 'Error: Stack format not supported';
```
jailln: Could you rephrase to: `[VPCLayer]: stack point cloud format not supporter` ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,155 @@
...
+         * @readonly
+         */
+        this.isVpcLayer = true;
+
+        this.scale = new THREE.Vector3(1.0, 1.0, 1.0);
+        this.offset = new THREE.Vector3(0.0, 0.0, 0.0);
+
+        const resolve = this.addInitializationStep();
+
+        this.whenReady = this.source.whenReady.then((/** @type {VpcSource} */ sources) => {
```
jailln, line 72: There are a lot of promises manipulations from here and at the end of the constructor. It's hard to understand when you are not involved in the code. Could you add a comment explaining the problem you are trying to resolve, and why you implemented like that please ?

packages/Main/src/Layer/VpcLayer.js
```diff
@@ -0,0 +1,155 @@
...
+            return [];
+        }
+
+        // pick the best bounding box
+        const bbox = (elt.tightbbox ? elt.tightbbox : elt.bbox);
+        elt.visible = context.camera.isBox3Visible(bbox, this.object3d.matrixWorld);
+        if (!elt.visible) {
+            markForDeletion(elt);
+            return [];
+        }
```
jailln: this part is also shared with PoincloudLayer, can you factorize it in a shared method please ?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,131 @@
...
+                    const mockSource = {
+                        url,
+                        boundsConforming: boundsConformings[i],
+                        whenReady,
+                        sId: i,
+                    };
+                    this.sources.push(mockSource);
+                });
+
+                return this.sources;
```
jailln, line 105: I think this then should be split into several simpler methods, it's too complex to understand like this in my opinion. You could for example have one method for the bounds, one for the crs and one for the sources manipulations.

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,131 @@
...
+            });
+    }
+
+    instanciate(source) {
```
jailln: typo: `instantiate`

Can you document it with jsdoc please?

packages/Main/src/Source/VpcSource.js
```diff
@@ -0,0 +1,131 @@
...
+        let newSource;
+        const url = source.url;
+        let srcWhenReady = cachedSrc.get(url);
```
jailln: rename variable to `cachedSrc` ? And rename the current `cachedSrc` to `cachedSources` ?

packages/Main/src/Layer/CopcLayer.js
ftoromanoff: Done

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: It was needed before the refactoring of VpcLayer.update().
I forgot to remove it afterward. Thanks !

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: same

packages/Main/src/Layer/PointCloudLayer.js
ftoromanoff: Done

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: Add export of the PointCloudLayer function to be able to use it on vpcLayer.

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: Done

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: Done (as well as in VpcSource)

packages/Main/src/Layer/VpcLayer.js
ftoromanoff: Done

packages/Main/src/Source/VpcSource.js
ftoromanoff: I move the CRS part (as It was refactorable). But for the 2 others parts, I discused with my collegues and as they are called only once (here) so they said it's not really ideal to move them to there own method...

packages/Main/src/Source/VpcSource.js
ftoromanoff: Done

packages/Main/src/Source/VpcSource.js
ftoromanoff: Done

jailln (approved):

PR #2508, opened 2025-03-05
refactor(PointCloudLayer): homogeneisation

ftoromanoff:
refactorisation of PointCloudNode:

- move EntwinePointTileNode.createChildAABB(node) and CopcNode.createChildAABB(node) to parent (PointCloudNode)
- add doc in CopcNode
- move getter in PotreeNode and Potree2Node before the methods
- move spacing property from source to layer for copc and entwine

### DIFF
--- a/packages/Main/src/Core/CopcNode.js
+++ b/packages/Main/src/Core/CopcNode.js
@@ -1,15 +1,27 @@
-import * as THREE from 'three';
 import { Hierarchy } from 'copc';
 import PointCloudNode from 'Core/PointCloudNode';
-
-const size = new THREE.Vector3();
-const position = new THREE.Vector3();
-const translation = new THREE.Vector3();
 
 function buildId(depth, x, y, z) {
     return `${depth}-${x}-${y}-${z}`;
 }
 
+/**
+ * @extends PointCloudNode
+ *
+ * @property {boolean} isCopcNode - Used to checkout whether this
+ * node is a CopcNode. Default is `true`. You should not change
+ * this, as it is used internally for optimisation.
+ * @property {number} entryOffset - Offset from the beginning of the file of
+ * the node entry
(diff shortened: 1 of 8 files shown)


packages/Main/src/Core/CopcNode.js
```diff
@@ -1,15 +1,27 @@
...
 }
 
+/**
+ * @extends PointCloudNode
+ *
+ * @property {boolean} isCopcNode - Used to checkout whether this
+ * node is a CopcNode. Default is `true`. You should not change
+ * this, as it is used internally for optimisation.
+ * @property {number} entryOffset - Offset from the beginning of the file of
+ * the node entry
```
jailln, line 15: to the node entry

packages/Main/src/Core/CopcNode.js
```diff
@@ -1,15 +1,27 @@
...
+ * @property {number} entryLength - Size of the node entry
+ * @property {CopcLayer} layer - Parent COPC layer
```
jailln: COPC Layer the nodes belongs to. BTW, could we avoid passing the layer ? Conceptually, nodes should not need to know their layer

jailln (commented):
Thanks, just small wording proposition

packages/Main/src/Core/CopcNode.js
ftoromanoff: Done

packages/Main/src/Core/CopcNode.js
ftoromanoff: Done.

For the 2nd point, It should be possible by redusing the param to the layer paroperties that we really use (mostly the source). I think this can be done when we will convert node in typescript ?

jailln: ok

jailln (approved):

PR #2515, opened 2025-03-12
Fix npm publist next after release

gchoqueux:
## Before contributing
Fix npm publist next after release

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "watch": "node ./config/concurrently.mjs watch",
     "changelog": "conventional-changelog -n ./config/conventionalChangelog/config.cjs -i changelog.md -s",
     "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm run update-packages && npm install && git add -A && git commit --amend --no-edit",
-    "prerelease-next": "npm version prerelease --preid next && npm run prerelease-next --workspaces --if-present && npm run update-packages",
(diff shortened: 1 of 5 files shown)


Desplandis:
This is still an issue (https://github.com/iTowns/itowns/actions/runs/13988222913/job/39166440004), are you working on this?

gchoqueux:
I push a new fix for npm publishing

Desplandis:
Hum... You should not use `--legacy-peer-deps` as it ignores all peer dependencies when installing... This could lead to some other bugs...

gchoqueux:
I know this isn't a good solution. Unfortunately, I've avoided using this option, but I haven't found any other solution.
The problem arises when `npm version` checks shared dependencies. npm throws an error because version `three.js (v170)` of the latest registered itowns release doesn't match that of the `pre-release (v174).`

The solution I suggest is only for pre-releases (next `version`).

NPM doesn't manage its dependencies well. You should try Yarn

Desplandis:
As much as I love `yarn`, `npm` is the default package manager provided by node and shall be supported by iTowns. We can't add another friction (and there is a lot) for devs.

gchoqueux:
Yes, absolutely. I wasn't planning on changing the manager in this PR—it's more of a proposal for later. Switching to Yarn would indeed bring its own set of challenges, but currently, NPM is causing issues.

gchoqueux:
An alternative could be to publish a new release whenever Three.js or Proj4 is updated.

gchoqueux:
fixed in #2563

PR #2517, opened 2025-03-13
feature(LayeredMaterial) LayeredMaterial expose in API

valentinMachado:
## Description

Expose `LayeredMaterial` in the API

## Motivation and Context

In order to make a custom rendering pass based on the LayeredMaterial I need this later to be exposed in the API see also => https://github.com/iTowns/itowns/discussions/2509


valentinMachado:
Extending the `LayeredMaterial` is a bit complicated because when it's created it needs some initialization
https://github.com/iTowns/itowns/blob/88aea4301cb7bd1a612946bcb258f8b8d682c69b/packages/Main/src/Converter/convertToTile.js#L53

so instead Iam going to do a shallow copy and customize it

PR #2521, opened 2025-03-14
Tile subdivision

clementcolin:
Subdivide tile is an operation that should not be in queue as is it can be blocked by incoming textures of other tile.

It can still be improve by deleting promises (therefore, pending subdivision may be deleted too)


jailln (approved):

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -430,18 +430,13 @@ class TiledGeometryLayer extends GeometryLayer {
...
-                /* specific params */
-                extentsSource: extents,
-                redraw: false,
-            };
-
-            return context.scheduler.execute(command).then((children) => {
+            const promises = [];
+
+            for (const extent of extents) {
+                promises.push(this.convert(node, extent));
```
jailln, line 436: I think we can go further and remove the promises chain of this.convert since they are not created synchronously. I'll give it a try

jailln (changes requested):

jailln (approved):

packages/Main/src/Layer/TiledGeometryLayer.js
jailln: I'll do it in another PR so we can merge this one :)

jailln:
Waiting for https://github.com/iTowns/itowns/issues/2530 to be resolved

packages/Main/src/Layer/TiledGeometryLayer.js
jailln: see #2525

Issue #2442, opened 2024-10-17
Terrain precision glitching / jumping when using lower resolution terrain with default update strategy


airnez:
## Context
Using a lower resolution terrain (world for instance) with the default strategy STRATEGY_MIN_NETWORK_TRAFFIC produces glitching effets by switching from precision levels very far apart from each other.

https://github.com/user-attachments/assets/e29a8923-5637-469b-9296-dcbd30d515dd

## Steps to Reproduce
1. Start from the view_3d_map example.
2. Modify the range of the view.controls.lookAtCoordinate function to a smaller value like 1000 to zoom closer to ground when searching a place
3. Remove the Highres France terrain to only use world terrain.
4. Remove the update strategy in WORLD_DTM.json to use default one
5. Search for the french city Sète
6. See how tiles are switching between very far appart levels

## Expected Behavior
Even if it makes sense that this strategy tries to request as few tiles as possible, it should not jump between levels too far appart. 

## Possible Cause/Fix/Solution
When zooming from the whole globe view to a close-up view, only very low-res and high res tiles are loaded, and the startegy switches betweel already loaded tiles.

AnthonyGlt:
related https://github.com/iTowns/itowns/pull/2344#issuecomment-2360364539

airnez:
I have troubles reproducing it with the terrain I mentionned, but this still applies to this one : 
Please contact me if you are interested to test the dataset

https://github.com/user-attachments/assets/5e920d02-13ff-4098-b7ac-d15810e5cdab

airnez:
@jailln I guess this is the issue we were talking about.

jailln:
I think there is indeed a visible effect that comes from jumping between levels too far appart like you said but I think there is also an additional bug: when zooming in quickly like in the video above and not moving afterwards, some tiles are not loaded at all, meaning tiles subdivision or tile fetching fails. I would expect all tiles to correctly load at some point (even if we see a big step effect).
I'll try to look into that.

Issue #2421, opened 2024-10-02
Improve terrain subdivision and raster tiles fecthing


jailln:
## Context

Several bugs are related to terrain subdivision and raster tiles fetching - for instance #2159 ; #2214

The terrain subdivision code and raster tile fetching (in [TiledGeometryLayer](https://github.com/iTowns/itowns/blob/1d9ffe9946d109d21e7ec2cf586616bb4c6670cb/src/Layer/TiledGeometryLayer.js#L21)) for instance has become quite complex, is hard to understand (see discussions in #2159 and #2344 for instance) and is buggy (see aforementioned issues).

## Proposed Solution

I think we need to:
* Refactor and simplify this part of the code that has a high cognitive complexity
* Extract and document the algorithms that control tile subdivision and raster tiles fetching
* That should eventually facilitate related bug fixing

I've started to work on a refactoring and simplification [on this branch](https://github.com/jailln/itowns/tree/refacto/tiledgeometrylayer)

airnez:
Also related to [#2442](https://github.com/iTowns/itowns/issues/2442)

AnthonyGlt:
Things to think about
https://github.com/iTowns/itowns/pull/2522#issuecomment-2809117623

Issue #2214, opened 2023-10-20
[Bug - VectorTile] Data not diplaying correctly (at the same zoom level of the associated tile).


ftoromanoff:
VectorTile Layer data are not always displayed on thesame  zoom level of the tile associated.

on branch master, example: vector_tile_raster_2d.html
![Capture](https://github.com/iTowns/itowns/assets/47628509/310aab41-9e1b-4625-84d4-129eda2dcf63)
![image](https://github.com/iTowns/itowns/assets/47628509/e6020fc5-c8e0-4d7e-9576-7c0c60cbd1f6)

As we can see above, some tiles (Rennes and Nantes) are not displayed at the same zoom level as the one around.

The PR #2135 make this bug even more visible and propose a quick fix 586f9fa5c08efa5c3497173d467fb9a152db2eed that seems to solve the probleme. But it as a quick fix and might not be well thought. A deeper analyze of the reasons of the bug should be set and a better fix should be propose.

jailln:
@ftoromanoff Can you still reproduce ? It seems fixed on the master branch

ftoromanoff:
Hello,

I did test using the PR #2522 that remove the temporary fix I introduce during the my PR #2135.
We can considere this bug solve.
Thanks !

PR #2522, opened 2025-03-14
fix(subdivision): Remove the subdividing condition in the update strategy

AnthonyGlt:
## Description

~~I've removed the method `hasEnoughTexturesToSubdivide` because it's valuable to continue to subdivide even though we don't have elevation data at a certain level.~~

~~I've added a check on the node pending subdivision to avoid blocking further requests when we are between 2 subdivision process.~~

I remove the check on the subdivision in the update strategy to allow more tiles to be updated. Also the check on the `isVectorTileSource` isn't necessary anymore (https://github.com/iTowns/itowns/issues/2214)

## Motivation and Context

We have a lot of issues linked to the subdivision:

https://github.com/iTowns/itowns/issues/2442
https://github.com/iTowns/itowns/issues/2421
https://github.com/iTowns/itowns/issues/2214

### DIFF
--- a/packages/Main/src/Layer/TiledGeometryLayer.js
+++ b/packages/Main/src/Layer/TiledGeometryLayer.js
@@ -320,7 +320,7 @@ class TiledGeometryLayer extends GeometryLayer {
             node.material.visible = true;
             this.info.update(node);
 
-            if (node.pendingSubdivision || (TiledGeometryLayer.hasEnoughTexturesToSubdivide(context, node) && this.subdivision(context, this, node))) {
+            if (this.subdivision(context, this, node)) {
                 this.subdivideNode(context, node);
                 // display iff children aren't ready
                 node.material.visible = node.pendingSubdivision;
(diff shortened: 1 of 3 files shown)


packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -320,7 +320,7 @@ class TiledGeometryLayer extends GeometryLayer {
             node.material.visible = true;
             this.info.update(node);
 
-            if (node.pendingSubdivision || (TiledGeometryLayer.hasEnoughTexturesToSubdivide(context, node) && this.subdivision(context, this, node))) {
+            if (this.subdivision(context, this, node)) {
```
Desplandis: In the code of `subdivision`, I would have added a hard-limit to not subdivide indefinitely.

Given that [common tile matrix sets](https://docs.ogc.org/is/17-083r4/17-083r4.html#toc48) covering the whole earth usually have a limit to zoom level 24, I would have added this level as limit to the subdivision. WDYT?

Desplandis (dismissed):
PR approved. However I think we should add a limit to the subdivision!

packages/Main/src/Layer/TiledGeometryLayer.js
AnthonyGlt: I've put it as a constant `MAX_SUBDIVISION_LEVEL` , should we make it as a parameter ?
IMO the constant should cover the current usage. If an issue occurs, we could add it as a parameter later.

packages/Main/src/Process/LayeredMaterialNodeProcessing.js
```diff
@@ -161,6 +164,9 @@ export function updateLayeredMaterialNodeImagery(context, layer, node, parent) {
 }
 
 export function updateLayeredMaterialNodeElevation(context, layer, node, parent) {
+    if (node.pendingSubdivision) {
+        return;
+    }
```
mgermerie: After testing your branch, I found out that this check causes an issue. Lets take an example with a `GlobeView` that displays elevation data and whose camera starting position is zoomed in. Tiles that are on the border of the viewing frustum won't have elevation data displayed - as no data from their parent tile has been loaded. Therefore, if the camera be tilted, these tiles may not intersect the viewing frustum even though they would, should elevation be considered.

That is visible on this screenshot I made on the [WFS on globe example](https://www.itowns-project.org/itowns/dev/examples/source_stream_wfs_3d.html) - on which I removed vector data as they do not interfere with the issue. You can see it on the bottom of the video:

https://github.com/user-attachments/assets/e3772201-4bee-4af0-9d90-7a9e16fdd87d

mgermerie (changes requested):
Thanks for the PR. I documented an issue I found out bellow.

And on a side note: we can see on the video bellow some 'flickering' tiles. Those tiles already display some relevant resolution texture and should not appear as these texture are 'reloaded'. This appears also on the release version and on `master` branch, but is less obvious since parent texture is displayed, showing a color that matches actual data more than the default blue of the `GlobeLayer`. It is not related to your PR, but do you have a hint on what's causing this - based on the hackathon discussions ?

AnthonyGlt:
A lot is happening based on `node.material.visible` value. If the value is at `false`, we won't do the request of data because it will be blocked in `updateLayeredMaterialNodeElevation` and in `refinementCommandCancellationFn` but also given less priority in `materialCommandQueuePriorityFunction`. If we don't request data, then the tile won't have elevation and without elevation, the bounding box will be wrong, which mean that the tile will be culled.
The value of `node.material.visible` is based on `pendingSubdivision`. So when the node has been subdivided, its visibility will be at false. 
By keeping `hasEnoughTexturesToSubdivide`, we block the subdivision until we have the data ( the important one is the elevation ), so the  `node.material.visible` stay at true until the data has been received.

IMO, we shouldn't consider the elevation as a common texture (like a colorlayer) because it's affecting the geometry of the tile, especially for the bounding box and therefore the culling process.
We should probably not based our methods only on the `material.visible` property, but on some` elevation.ready` or something...

@mgermerie as you predicted, the behavior you observed, is resulting of the missing parent elevation.

AnthonyGlt:
@mgermerie I have rollback the majority of the changes. I've only kept the edit on the strategy...

mgermerie:
Hi, Sorry for the late review and thank you for the modifications. I can see that the leftover modifications fix #2442, but I'm not sure I understand why.

What I understood is the following:
Let's say we zoom from level `1` to `19`, and then zoom out or move so that the tiles at zoom `19` are replaced by zoom `18` tiles. The new zoom `18` tiles have a `pendingSubdivision` set to `true` and will thus display data from their first ancestor with loaded data (here it should be a zoom `1` tile) instead of fetching data for their actual zoom level.

Is my understanding of what causes the issue right ? And in that case, why would the newly displayed zoom `18` tiles have a `pendingSubdivision` set to `true` ?

mgermerie (approved):

PR #2525, opened 2025-03-17
Further simplify tile subdivision mechanism

jailln:
Based on #2521

Related to #2522

* TiledGeometryLayer tiles are not built in promises anymore (useless since #2521 since they are built synchronously)
* remove TileMesh.pendingSubdivision that also becomes useless since #2521


gchoqueux:
I agree that pendingSubdivision should be removed. At the time of development, we introduced tile asynchronicity to handle streamed DEM tiles or to compute tiles in workers.

gchoqueux:
We need to remove `pendingSubdivision `and `hasEnoughTexturesToSubdivide`, which are only used during texture initialization.
A local texture, used to initialize the level 0 tile, would make these properties unnecessary.

gchoqueux:
When you remove tiles instancing from scheduler, there is no more  earlyDropFunction mechanism, to avoid unnecessary instancing.

gchoqueux:
Sorry,  `earlyDropFunction` doesn't exist for `TiledGeometryLayer`

packages/Main/src/Core/Prefab/TileBuilder.ts
```diff
@@ -100,21 +98,18 @@ export function newTileGeometry(
...
-                        index: cachedBuffers.index.array as
-                            Uint8Array | Uint16Array | Uint32Array,
-                        uv: cachedBuffers.uv.array as Float32Array,
-                    }
-                    : undefined,
-            );
-        } catch (e) {
-            return Promise.reject(e);
-        }
+        // TODO removed a try catch here that didn't seem needed after a first quick look, but needs to be re-checked
```
gchoqueux, line 101: Yes, there's no need to add a try block. Could remove comment

packages/Main/src/Layer/LayerUpdateStrategy.js
```diff
@@ -11,12 +11,7 @@ export const STRATEGY_GROUP = 1;
...
 
-function _minimizeNetworkTraffic(node, nodeLevel, currentLevel, source) {
-    // TO DO source.isVectorTileSource is a temp fix for pendingSubdivision.
-    // see issue https://github.com/iTowns/itowns/issues/2214
-    if (node.pendingSubdivision && !source.isVectorTileSource) {
-        return currentLevel;
-    }
+function _minimizeNetworkTraffic(node, nodeLevel) {
     return nodeLevel;
 }
```
gchoqueux, line 16: this method could be removed

packages/Main/src/Layer/LayerUpdateStrategy.js
```diff
@@ -74,7 +69,7 @@ export function chooseNextLevelToFetch(strategy, node, nodeLevel = node.level, c
             // default strategy
             case STRATEGY_MIN_NETWORK_TRAFFIC:
             default:
-                nextLevelToFetch = _minimizeNetworkTraffic(node, nodeLevel, currentLevel, layer.source);
+                nextLevelToFetch = _minimizeNetworkTraffic(node, nodeLevel);
```
gchoqueux, line 72: If you remove `_minimizeNetworkTraffic`

```suggestion
     nextLevelToFetch = nodeLevel
```

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -303,13 +303,6 @@ class TiledGeometryLayer extends GeometryLayer {
         if (!node.parent) {
             return ObjectRemovalHelper.removeChildrenAndCleanup(this, node);
         }
-        // early exit if parent' subdivision is in progress
-        if (node.parent.pendingSubdivision) {
-            node.visible = false;
-            node.material.visible = false;
-            this.info.update(node);
```
gchoqueux: the updating info is released in each case ?

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -320,10 +313,10 @@ class TiledGeometryLayer extends GeometryLayer {
             node.material.visible = true;
             this.info.update(node);
 
-            if (node.pendingSubdivision || (TiledGeometryLayer.hasEnoughTexturesToSubdivide(context, node) && this.subdivision(context, this, node))) {
+            if (this.subdivision(context, this, node)) {
                 this.subdivideNode(context, node);
                 // display iff children aren't ready
-                node.material.visible = node.pendingSubdivision;
+                node.material.visible = false;
                 this.info.update(node);
```
gchoqueux, line 320: `this.info.update(node);` Could this call be written only once in this method?

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -422,39 +415,16 @@ class TiledGeometryLayer extends GeometryLayer {
      * @param {Object} context - The context of the update; see the {@link
      * MainLoop} for more informations.
      * @param {TileMesh} node - The node to subdivide.
-     * @return {Promise}  { description_of_the_return_value }
      */
     subdivideNode(context, node) {
-        if (!node.pendingSubdivision && !node.children.some(n => n.layer == this)) {
+        if (!node.children.some(n => n.layer == this)) {
```
gchoqueux, line 420: may be could be write  
`if (node.children.length)`Because the children can only be on the same node's layer (to be verified).

PR #2526, opened 2025-03-21
chore(typescript): add support for __DEBUG__ and glsl imports

Desplandis:
## Description

This PR adds support on the typescript side for the following constructs:
- The `__DEBUG__` global which is not defined prior the compilation step
- The `*.glsl` imports which are not valid javascript/typescript but are inlined as string during the compilation step


AnthonyGlt (approved):
Thank you :D

PR #2527, opened 2025-03-26
feat(VR): add control event

AnthonyGlt:
## Description
Add an event when we add the control handler in a webxr context

## Motivation and Context
To access the method of [VRControls.js](https://github.com/iTowns/itowns/blob/3c1bb1c81c79c82685b459df55c3368e6fe568ff/packages/Main/src/Controls/VRControls.js) from [WebXR.js](https://github.com/iTowns/itowns/blob/3c1bb1c81c79c82685b459df55c3368e6fe568ff/packages/Main/src/Renderer/WebXR.js), we need to now when it is initialized.
By adding an event, we can know when we can access it

### DIFF
--- a/packages/Main/src/Main.js
+++ b/packages/Main/src/Main.js
@@ -25,6 +25,7 @@ export { default as FirstPersonControls } from 'Controls/FirstPersonControls';
 export { default as StreetControls } from 'Controls/StreetControls';
 export { default as PlanarControls } from 'Controls/PlanarControls';
 export { default as VRControls } from 'Controls/VRControls';
+export { VR_EVENTS } from 'Renderer/WebXR';
 export { CONTROL_EVENTS } from 'Controls/GlobeControls';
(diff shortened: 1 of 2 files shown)


mgermerie (approved):
That's fine for me. Just a question for my understanding: I could not see in iTowns repo any case where `WebXR` is instantiated without calling `WebXR.initializeWebXR()` right after it. So is there a case where the initialization could be delayed ? Such a case creates the need of an event instead of directly accessing the property.

AnthonyGlt:
> That's fine for me. Just a question for my understanding: I could not see in iTowns repo any case where `WebXR` is instantiated without calling `WebXR.initializeWebXR()` right after it. So is there a case where the initialization could be delayed ? Such a case creates the need of an event instead of directly accessing the property.

True... Maybe we could directly call the initializeWebXR in the constructor of WebXR. I don't see a case where we will need to delay it :thinking:

PR #2546, opened 2025-05-26
Migration to lil-gui library (from dat-gui)

ftoromanoff:
In the example we are currently using the dat.gui library as it was the library used in three.js.
But dat-gui is no longer maintained.

Three.js has chose to migrate to lil-gui, and it seems the best choice to follow them.

In this PR, I did migrate to lil-gui for the pointCloud example to test and show how it can be done.

If you have ay remarqs before we move all example to that library and remove dat-gui form itowns.

### DIFF
--- a/examples/copc_simple_loader.html
+++ b/examples/copc_simple_loader.html
@@ -15,7 +15,7 @@
         </style>
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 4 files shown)


AnthonyGlt (commented):
Thank you for this MR!

dat.gui.min.js is used in a lot of examples. Could you do this update everywhere ?

ftoromanoff:
Hello,
I opened an issue about that #2554.
As it is not a priority, I propose we migrate it litle by little, instead of waiting to do it all at once.

AnthonyGlt (approved):

Issue #2305, opened 2024-03-25
[PROPOSAL] Error as value instead of throwing during picking


HoloTheDrunk:
## Your Environment

* Version used: commit 9fa4cde79b138548c5b897ea78bd2a88d0946029
* Browser Name and version: Firefox 123.0 (64-bit)
* Operating System and version (desktop or mobile): Ubuntu 22.04.4 LTS x86_64

## Context

Datasets should ideally be clean of artifacts, but that's rarely the case so failing an entire picking call for a single artifact encounter is ineffective.
This happens for example when picking certain NaN-ed out points in point cloud datasets while using proj4.

## Expected Behavior

The picking routine should catch the errors and return them as values along with the valid results.

## Actual Behavior

The picking routine `pickPointsAt` in `src/Core/Picking.js` does not catch errors and instead passes them along, leading to a discard of possibly valid and unrelated results when `Coordinates::spatialEuclideanDistanceTo` throws an error.

## Possible Cause/Fix/Solution

Catching thrown errors per-point and transforming them into error values to push into the `result` array. This should not incur additional runtime costs at least on the V8 engine.

HoloTheDrunk:
It just hit me that this would be even better handled by making the function a generator and keeping the current throw mechanism.

PR #2547, opened 2025-05-26
Fix Picking - No geometry or unloaded points should not be taken for picking

Pourfex:
## Description
During point cloud picking, there is a chance that point being picked are unloaded; due to : 

- The point budget is reached
- Nodes are being loaded/unloaded
- The camera is moving quickly
- The view is being panned/zoomed

## Motivation and Context

This means that in Picking.js line 181
```
pointPos.fromBufferAttribute(o.geometry.attributes.position, candidates[i].index);
```
Will return a (0,0,0) value, thus giving incorrect compute and even crash the line 189 :

```
const dist = pointPosCoord.spatialEuclideanDistanceTo(cameraPosCoord);
```

With no try/catch mecanism; the picking fails and user cannot click anywhere, with a spatialEuclideanDistanceTo problem which is misleading.

We propose to check the geometry.attributes.position validity to ensure we have correct calculation, otherwise we should skip this unloaded point. The check could also have been been done using

```
o.userData.node.octreeIsLoaded
```

But we also need to check is the object picked is a PointCloudNode type and so on. We prefered 

```
  // Skip if geometry is invalid or missing position attribute
  if (!o.geometry || !o.geometry.attributes || !o.geometry.attributes.position) {
      return;
  }
```

as it would be more generic.

Very likely solve : 
https://github.com/iTowns/itowns/issues/2305 

Browser : firefox & chrome 136, itowns latest 2.45.1, macOS 15.

### DIFF
--- a/packages/Main/src/Core/Picking.js
+++ b/packages/Main/src/Core/Picking.js
@@ -169,6 +169,10 @@ export default {
                 // disable picking mode
                 o.material.enablePicking(false);
 
+                // Skip if geometry is invalid or missing position attribute
+                if (!o.geometry || !o.geometry.attributes || !o.geometry.attributes.position) {
+                    return;
+                }
                 // if baseId matches objId, the clicked point belongs to `o`
                 for (let i = 0; i < candidates.length; i++) {
                     if (candidates[i].objId == o.baseId) {


Desplandis (approved):
Managed to reproduce the issue...
Thanks for the fix =)

PR #2548, opened 2025-05-26
fix(OrientationUtils): add all lcc aliases

ftoromanoff:
add all lcc aliases in OrientationUtils.js

### DIFF
--- a/packages/Geographic/src/OrientationUtils.ts
+++ b/packages/Geographic/src/OrientationUtils.ts
@@ -413,7 +413,12 @@ export function quaternionFromEnuToCRS(
     const proj = typeof crsOrProj === 'string' ? proj4.defs(crsOrProj) : crsOrProj;
     switch (proj.projName) {
         case 'geocent': return quaternionFromEnuToGeocent();
-        case 'lcc': return quaternionFromEnuToLCC(proj as LCCProjection);
+        case 'lcc':
+        case 'Lambert Tangential Conformal Conic Projection':
+        case 'Lambert_Conformal_Conic':
+        case 'Lambert_Conformal_Conic_1SP': case 'Lambert_Conformal_Conic_2SP':
+        case 'Lambert Conic Conformal (1SP)': case 'Lambert Conic Conformal (2SP)':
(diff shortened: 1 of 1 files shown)


mbredif:
It looks that these projName values are indeed standard aliases in proj4js.
Can't we be a bit more generic and generalize this PR to the 4 supported projection types and use the `names` arrays exported by the projections instead of listing the aliases explicitly ?

https://github.com/proj4js/proj4js/blob/master/lib/projections/lcc.js#L130-L138
https://github.com/proj4js/proj4js/blob/master/lib/projections/geocent.js#L20
https://github.com/proj4js/proj4js/blob/master/lib/projections/tmerc.js#L160
https://github.com/proj4js/proj4js/blob/master/lib/projections/longlat.js#L10

Out of curiosity, what usage triggered the need for the lcc aliases ?

ftoromanoff:
> It looks that these projName values are indeed standard aliases in proj4js. Can't we be a bit more generic and generalize this PR to the 4 supported projection types and use the `names` arrays exported by the projections instead of listing the aliases explicitly ?
> 
> https://github.com/proj4js/proj4js/blob/master/lib/projections/lcc.js#L130-L138 https://github.com/proj4js/proj4js/blob/master/lib/projections/geocent.js#L20 https://github.com/proj4js/proj4js/blob/master/lib/projections/tmerc.js#L160 https://github.com/proj4js/proj4js/blob/master/lib/projections/longlat.js#L10
> 
> Out of curiosity, what usage triggered the need for the lcc aliases ?

I did the modification based on your idea, It's more in line with what you had in mind ?

As I am implementing reporjection 'a la volée' for point cloud data, we need to support all crs that we get from the data metadata.

ftoromanoff:
@mbredif

gchoqueux (approved):

Desplandis:
@ftoromanoff I saw that you used `@ts-expect-error`. If the `oproj` field, why not contribute to the type definition of proj4 or just extend the interface from iTowns? We should limit the use of `ts-expect-error` to the strict minimum.

PR #2549, opened 2025-05-27
fix: use a fresh page when loading an example

HoloTheDrunk:
## Description

Create and use a new puppeteer `Page` for every call to `loadExample`. This likely lengthens the functional testing process, but it's an easy way to avoid leakage that leads to bugs that are hard to debug.

## Motivation and Context

We encountered strange CI errors during the last push for the merge of #2477 that we tracked down to being an issue of page state being left over between examples.


Desplandis:
Yeah, errors when quitting the previous test/page leaked into the next test...
As for performance, I think that closing/creating another page is negligible compared to setting up a 3D context. x)

Desplandis (approved):
Tested on a functional test with a fake example URL and another one with a top-level exception on the HTML example. Both tests failed as they should do.

PR #2550, opened 2025-06-02
feat: Add gamma and ambient boost option for Point Shader

Pourfex:
To have better visualization of Point cloud under shadowed conditions (trees, underground utilities, capture device shadow...)

## Description
Had two options for point cloud vizualisation on the point cloud shader : 
- Gamma : Enhance gamma with explanation here https://pyimagesearch.com/2015/10/05/opencv-gamma-correction/ 
- Ambient boost : Have clamped illumination for pixels with explanation here : https://docs.opencv.org/4.x/d3/dc1/tutorial_basic_linear_transform.html

## Motivation and Context

In point cloud data, some acquisition techniques do have shadows in their results.
Shadows that comes from trees, side of a trench, captation device (lidar, people with camstick...) and so on.

Having those two options greatly helps for better understanding of those shadowed area, video here : 
shadow_point_cloud2

Itowns 2.45.1.

### DIFF
--- a/packages/Debug/src/PointCloudDebug.js
+++ b/packages/Debug/src/PointCloudDebug.js
@@ -156,6 +156,13 @@ export default {
         if (layer.material.shape != undefined) {
             styleUI.add(layer.material, 'shape', PNTS_SHAPE).name('Shape mode').onChange(update);
         }
+        if (layer.material.gamma != undefined) {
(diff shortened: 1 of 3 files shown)


HoloTheDrunk (approved):
Nice addition, thank you for the PR.

PR #2552, opened 2025-06-05
feat(Main): Add tsc command to generate types for typescript integration

ketourneau:
## Description
We need to have types for typescript integration

## Motivation and Context
`tsc` command is already present in Geographic package.
I think it's missing in Main package.

### DIFF
--- a/packages/Main/package.json
+++ b/packages/Main/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "lint": "eslint \"src/**/*.{js,ts,tsx}\" \"test/**/*.js\"",
-    "transpile": "cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib --extensions .js,.ts",
+    "transpile": "tsc && cross-env BABEL_DISABLE_CACHE=1 babel src --out-dir lib --extensions .js,.ts",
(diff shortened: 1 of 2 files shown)


HoloTheDrunk (approved):
Fine with this once the tests pass, thank you for the PR!

Not entirely sure what the source of the issue is, the error is a bit cryptic. Looking into it but my search has proven unfruitful so far.

ketourneau:
@HoloTheDrunk I found solution by adding to tsconfig :
```
"paths": {
            "3d-tiles-renderer": [
                "../../../node_modules/3d-tiles-renderer/src/index.js"
            ]
}
```

HoloTheDrunk (changes requested):
Seems like the issue comes from upstream where they don't specify the `"types"` entry-point in their `package.json`. Given the already complex build process, we'd rather avoid band-aid fixes using most frightening relative paths.
Best we can recommend to fix this issue while we wait for upstream to fix this is to add a `/** @type{any} */` above the `this.tilesRenderer` in `OGC3DTilesLayer.js` on line 281.

ketourneau:
Ok indeed it seems to be a better temporary solution !

HoloTheDrunk (approved):

PR #2553, opened 2025-06-05
refactor(PointCloudLayer): PointCloudNode.id rename and pointSpacing rework

ftoromanoff:
light refactorisation of PointCloudNode:
- rename id property as voxelKey for copc and entwine and as hierarchyKey for potre and potree2
  - add a getter for id at PointCloudNode level
- move pointSpacing from layer to node

### DIFF
--- a/packages/Main/src/Core/PointCloudNode.js
+++ b/packages/Main/src/Core/PointCloudNode.js
@@ -16,6 +16,10 @@ class PointCloudNode extends THREE.EventDispatcher {
         this.sse = -1;
     }
 
+    get id() {
+        throw new Error('In extended PointCloudNode, you have to implement the getter id!');
+    }
+
     add(node, indexChild) {
         this.children.push(node);
         node.parent = this;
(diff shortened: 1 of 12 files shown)


gchoqueux (commented):
Thx, 
I approved, We have, already,  discussed in peer programming

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -16,6 +16,15 @@ class PointCloudNode extends THREE.EventDispatcher {
         this.sse = -1;
     }
 
+    get pointSpacing() {
+        return this.layer.spacing / 2 ** this.depth;
+    }
+
+    get id() {
+        // hierarchyKey for potreeNode
+        return this.hierarchyKey ?? `${this.depth}${this.x}${this.y}${this.z}`;
```
mgermerie: This returns something only if in a `PotreeNode` and `Potree2Node`. This should not be defined in the more generic `PointCloudNode`.

ftoromanoff: Done in last commit (Commit that will be squashed if you agree @mgermerie )

mgermerie (approved):

PR #2556, opened 2025-06-11
Cleanup layer properties

Desplandis:
## Description
This PR introduces several changes to the base Layer and layer-related classes. The main changes includes the migration of `LayerUpdateState` and `LayerUpdateStrategy` to typescript, the reorganization of Layer properties for better separation of concerns and the addition of a startup lifecycle method as a first step to cleanup the mess that is `View#preprocessLayer`. This will not break existing layers within iTowns but could break custom layers inheriting directly from `Layer`.

## Motivation and Context
The motivation for the following changes is as follow:
- Introduction of `Layer#startup` lifecycle method for separation of concerns, allowing for more explicit initialization of layers. **Follow-up**: remove `View#preprocessLayer` and most of the complexity of `View#addLayer`.
- Migration of `LayerUpdateState` and `LayerUpdateStrategy` to typescript (I will personally rebase #2525 on this PR since there is some function paramater changes to `chooseNextLevelToFetch`).
**Follow-up**: migrate TileMesh-related modules to typescript (see #2557) / migrate `LayerNodeProcessing`
- Move `ImageryLayers` to its own module as I have short-term plan to migrate all layers to typescript (with some type hacks) and have no plan to migrate `ImageryLayers` (which given the code needs an entire refactoring...).
**Follow-up**: migrate Layer modules to typescript
- Move some layers properties to inherited classes for better separation of concerns, facilitate the typechecking of `Layer` and children.
  + `layerUpdateStrategy` used only on `RasterLayer`-derived classes
  + `style` used only on `*3DTilesLayer`, `FeatureGeometryLayer` and `ColorLayer`
  + `mergeFeatures`/`addLabel` used only on `FeatureGeometryLayer` and `ColorLayer`
**Follow-up**: migrate Layer modules to typescript

Changes considered but not done on this PR:
- Cleanup the mess in `View#preprocessLayer`
- Moving the caching mechanism in classes where it is effectively used
- Some things that I surely forgot ;)


Desplandis:
@AnthonyGlt @mgermerie @HoloTheDrunk If anyone want to review this PR. Especially if you depend on some properties here.

AnthonyGlt:
@Desplandis  Thank you for the PR.
I like this clean up. It makes the code easier to understand !

packages/Main/src/Layer/LayerUpdateStrategy.ts
```diff
@@ -21,50 +46,62 @@ function _minimizeNetworkTraffic(nodeLevel) {
...
         Math.ceil((currentLevel + nodeLevel) / 2));
 }
 
-export function chooseNextLevelToFetch(strategy, node, nodeLevel = node.level, currentLevel, layer, failureParams) {
+export function chooseNextLevelToFetch(
+    strategy: LayerUpdateStrategy,
+    nodeLevel: number,
+    currentLevel: number,
+    failureParams: { lowestLevelError: number },
+    zoom?: { min: number; max: number },
```
AnthonyGlt, line 79: `zoom` is bothering me a bit because it's only used for _dichotomy() but I don't have an idea to make it better so OK :+1:

AnthonyGlt (approved):

packages/Main/src/Layer/ColorLayer.js
```diff
@@ -107,6 +118,8 @@ class ColorLayer extends RasterLayer {
          */
         this.isColorLayer = true;
 
+        this.style = style instanceof Style ? style : new Style(style);
```
gchoqueux, line 121: Wouldn't it make more sense to move style into Layer, in order to centralize common properties (visibility, opacity/scale, etc.) that every layer can have?

packages/Main/src/Layer/Layer.js
```diff
@@ -213,6 +179,21 @@ class Layer extends THREE.EventDispatcher {
...
+     */
+    async startup(/* context */) {
+        try {
+            await Promise.all(this._promises);
+            this._resolve();
+        } catch (error) {
+            this._reject(error);
+        }
+    }
+
```
gchoqueux, line 196:
```suggestion
    async startup(/* context */) {
            await Promise.all(this._promises)
                .then(this._resolve)
                .catch (this._reject);
    }
```

packages/Main/src/Layer/ColorLayer.js
Desplandis: Since not all layers are styled (e.g. elevation layers, pointcloud layers), I decided to remove those properties from the base Layer class.

packages/Main/src/Layer/Layer.js
Desplandis: Depends on whether we want to use the async paradigm or not, but I will follow your suggestion here.

PR #2557, opened 2025-06-11
Migrate TileMesh-related modules to typescript

Desplandis:
## Description
This PR introduces TypeScript migration of the whole tile conversion process. This includes `TileMesh`, `OBB` and `convertToTile`.

Depends on #2556.

### DIFF
--- a/packages/Main/src/Converter/convertToTile.ts
+++ b/packages/Main/src/Converter/convertToTile.ts
@@ -1,15 +1,32 @@
 import * as THREE from 'three';
 import TileMesh from 'Core/TileMesh';
 import { LayeredMaterial } from 'Renderer/LayeredMaterial';
-import { newTileGeometry } from 'Core/Prefab/TileBuilder';
+import { newTileGeometry, TileBuilder, TileBuilderParams } from 'Core/Prefab/TileBuilder';
 import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
 import { geoidLayerIsVisible } from 'Layer/GeoidLayer';
 
+import type { Extent } from '@itowns/geographic';
+import type { LayeredMaterialParameters } from 'Renderer/LayeredMaterial';
+
 const dimensions = new THREE.Vector2();
 
-function setTileFromTiledLayer(tile, tileLayer) {
+// A simplified interface for TiledGeometryLayer.
+// It is used to avoid a dependency on the full TiledGeometryLayer type.
+interface TileLayerLike {
+    diffuse: THREE.Color;
+    showOutline: boolean;
+    isGlobeLayer: boolean;
+    segments: number;
+    disableSkirt: boolean;
+    hideSkirt: boolean;
+    tileMatrixSets: string[];
+    materialOptions: LayeredMaterialParameters;
+    builder: TileBuilder<TileBuilderParams>;
+}
+
+function setTileFromTiledLayer(tile: TileMesh, tileLayer: TileLayerLike) {
     if (tileLayer.diffuse) {
-        tile.material.diffuse = tileLayer.diffuse;
+        tile.material.setUniform('diffuse', tileLayer.diffuse);
     }
 
     if (__DEBUG__) {
@@ -32,7 +49,7 @@ function setTileFromTiledLayer(tile, tileLayer) {
 }
 
 export default {
-    convert(requester, extent, layer) {
+    convert(requester: TileMesh, extent: Extent, layer: TileLayerLike) {
         const builder = layer.builder;
         const parent = requester;
         const level = (parent !== undefined) ? (parent.level + 1) : 0;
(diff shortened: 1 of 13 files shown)


packages/Main/src/Core/TileMesh.ts
```diff
@@ -0,0 +1,147 @@
...
+    findCommonAncestor(tile: TileMesh): TileMesh | undefined {
+        if (!tile) {
+            return undefined;
+        }
+        if (tile.level == this.level) {
+            if (tile.id == this.id) {
+                return tile;
+            } else if (tile.level != 0) {
+                // @ts-expect-error By invariant, parent is always a TileMesh
+                return this.parent.findCommonAncestor(tile.parent);
```
Neptilo: Would a cast be better than a `ts-expect-error`?
`(this.parent as TileMesh).findCommonAncestor(tile.parent as TileMesh)`
(The comment is still useful anyway.)

packages/Main/src/Layer/ColorLayer.js
```diff
@@ -71,6 +72,13 @@ class ColorLayer extends RasterLayer {
      * * `2`: white color to invisible effect.
      * * `3`: custom shader effect (defined `ShaderChunk.customBodyColorLayer` and `ShaderChunk.customHeaderColorLayer`).
      * @param {number} [config.effect_parameter=1.0] - amount value used with effect applied on raster color.
+     * @param {(boolean|Object)} [config.addLabelLayer] - Used to tell if this layer has
+     * labels to display from its data. For example, it needs to be set to `true`
+     * for a layer with vector tiles. If it's `true` a new `LabelLayer` is added and attached to this `Layer`.
+     * You can also configure it with {@link LabelLayer} options described below such as: `addLabelLayer: { performance: true }`.
+     * @param {boolean} [config.addLabelLayer.performance=false] - In case label layer adding, so remove labels that have no chance of being visible.
```
Neptilo: The language is quite difficult to understand here. If I understood well, it could be:
"In case of adding a label layer, remove labels that have no chance of being visible.
Even in the best case, labels will never be displayed, for example, if there are many labels."

packages/Main/src/Converter/convertToTile.ts
```diff
@@ -1,15 +1,30 @@
...
 import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
 import { geoidLayerIsVisible } from 'Layer/GeoidLayer';
 
+import type { Extent } from '@itowns/geographic';
+import type { LayeredMaterialParameters } from 'Renderer/LayeredMaterial';
+
 const dimensions = new THREE.Vector2();
 
-function setTileFromTiledLayer(tile, tileLayer) {
+interface TileLayerLike {
```
Neptilo, line 15: Is this a simplified, POD version of `TiledGeometryLayer`? If so, we could mention it.

Neptilo:
This PR introduces 3 new TODOs (and removes 1, thanks for that!).
In general, I’d recommend avoiding TODOs in code unless they’re tied to a clear follow-up process (e.g. a linked issue). Otherwise, they often get forgotten and accumulate over time. A good alternative is to either address them before merging, or create an issue and link it directly in the TODO.

I also noticed a question added in a comment in `LayerUpdateStrategy`. Similar to TODOs, that kind of question is more likely to get answered if raised in chat or as part of an issue.

packages/Main/src/Core/TileMesh.ts
Desplandis: Yeah, but I think our linter does not authorize such cast, but I will do it either way (with a @ts-expect-error)

packages/Main/src/Converter/convertToTile.ts
Desplandis: Since `TiledGeometryLayer` is not typed yet, I decided to go for an sub-type for this function.

Desplandis:
@Neptilo I did all changes following your review. =)

Neptilo:
> @Neptilo I did all changes following your review. =)

Nice!

And what do you think about my previous comment?

> This PR introduces 3 new TODOs (and removes 1, thanks for that!). In general, I’d recommend avoiding TODOs in code unless they’re tied to a clear follow-up process (e.g. a linked issue). Otherwise, they often get forgotten and accumulate over time. A good alternative is to either address them before merging, or create an issue and link it directly in the TODO.

Currently I can see two added TODOs in `OBB`:
* in `updateZ`: "why not add the geoid height to the min and max parameters? The implementation of GeoidLayer is leaking here."
* in `setFromExtent`: "The crs are hardcoded here."
* a TODO in disguise ;) : "we should investigate the use of the THREE.OBB class."

Desplandis:
My bad, I forgot about those comments!

Desplandis:
@Neptilo

For the `updateZ`, it will be fixed when we'll have the possibility to combine multiple elevation layers (additive or as a mask). The geoid layers will be considered as elevation layers instead (and removing this implementation leakage). I can add a short notice to explain this.

For `setFromExtent`, it will surely not be fixed. Unless we use three OBB implementation (but this is my TODO in disguise ;)). I will remove both comments.

Neptilo (approved):

PR #2250, opened 2023-12-28
feature: Add Cloud Optimized GeoTIFF (COG) sample

ketourneau:
## Description
We need to display Cloud Optimized GeoTIFF ([COG](https://www.cogeo.org/)) with iTowns.
A [pull request](https://github.com/iTowns/itowns/pull/1307) already exist about COG but it seems to be abandoned.
We tried to used PR but we got some trouble with UTIF library to parse COG.
We think [geotiff](https://github.com/geotiffjs/geotiff.js/) is a better library to parse COG.

## Motivation and Context
We need to display heavy georeferenced tif file (geotiff) which come from CAD software.
We think the best solution is to convert geotiff to COG and display it with iTowns.
Maybe it can help [this discussion](https://github.com/iTowns/itowns/discussions/2181) ?

We will change the sample url if [this PR](https://github.com/iTowns/iTowns2-sample-data/pull/17) on itowns data is accepted (or maybe we don't need to change the sample url ?).

PR #2558, opened 2025-06-12
[Feature] Add support for GeoTIFF and COG data.

mgermerie:
## Description

This pull request adds a parser for GeoTIFF data, using [geotiff.js](https://github.com/geotiffjs/geotiff.js/) library. It also adds a parser and a source for Cloud Optimized GeoTIFF (COG) data, that use GeoTIFF parser.

Here is a detail of what is added in the PR:

#### A « node » for GeoTIFF data
 
A new class is defined, called `GeotiffNode`, that stores information on a GeoTIFF Image File Directory (IFD, basically a sub-image within a TIFF file). It also implements methods to retrieve IFD associated data. This new class is detailed bellow.

```mermaid
---
title: GeotiffNode description
---

classDiagram

class GeotiffNode~~ {
    +image geotiff.GeoTIFFImage
    +width number
    +height number
    +resolution Array<number>
    +origin: Array<number>
    +dataType THREE.TextureDataType
    +extractTexture(options) THREE.Texture
    -extentToImageWindow(extent: Extent) Array<number>
    -readImage(options) geotiff.TypedArrayWithDimensions
}

```

#### A parser for GeoTIFF data

A parser for GeoTIFF data, `GeotiffParser.parse` is created. This method instantiates a `GeotiffNode` from some GeoTIFF data, and uses node's methods to return a `TextureWithExtent` that is used in the convert methods of `RasterLayer` instances. A `TextureWithExtent` is a `THREE.DataTexture` with an `itowns.Extent`, which is a type we already use when parsing images.

#### A Source and a parser for COG

A new `CogSource` is defined. When instantiated, this source reads metadata from a COG file and creates an array containing one `GeotiffNode` for each overview found in the file.

The parser, `CogParser.parse`, selects which overview is to be displayed based on the resolution we want for the parsed data. It then retrieve the overview's image data into a `TextureWithExtent`, the same way `GeotiffParser.parse` does. 

## Resources

You can find different helpful resources here:

- Tiff format specification [here](https://www.itu.int/itudoc/itu-t/com16/tiff-fx/docs/tiff6.pdf) ;
- GeoTIFF format specification [here](https://docs.ogc.org/is/19-008r4/19-008r4.html)
- COG format specification [here](https://docs.ogc.org/is/21-026/21-026.html), and summarized [here](https://github.com/cogeotiff/cog-spec/blob/master/spec.md)

## Motivation and Context

This work was initiated by @ketourneau in #2250. Thanks to him and his team for the initial work!

## What is left to do

- [x] fix a performance issue when trying to display a small extent GeoTIFF within a big extent tile (I'll document this issue here later) ;
- [ ] add support for every tiff specified data format (only `RedFormat` and `RGBAFormat` are supported for now) ;
- [x] create a proper example ;
- [x] thoroughly test the feature and add automatic unit and functional tests.


packages/Main/src/Parser/CogParser.ts
```diff
@@ -0,0 +1,123 @@
...
+            Math.abs(overview.resolution[1]),
+        );
+
+        if (targetResolution >= sourceResolution) {
+            break;
+        }
+    }
+
+    if (!overview) {
+        throw Error('No level found. LevelTree must be empty');  // TODO: make better message
```
airnez: Remember to adress and remove this //TODO here

examples/cog_loader.html
```diff
@@ -0,0 +1,144 @@
+<html>
```
airnez: If the commit title is still right, don't forget to remove this test file before merge. I did not review it fyi.

packages/Main/test/unit/cogParser.js
```diff
@@ -0,0 +1,38 @@
...
+    describe('Fetch test data', function () {
+        it('Fetch rgba cog data', async function () {
+            source = await GeoTIFF.fromUrl('test/data/geotiff/rgba.cog.tif');
+            console.log(source);
+        });
+    });
+
+    describe('Unit tests for GeotiffParser', function () {
+        it('should parse COG correctly', async function () {
+            // do
```
airnez: Remember to implement those tests

packages/Main/src/Parser/GeotiffParser.ts
```diff
@@ -0,0 +1,316 @@
...
+                height: outputHeight,
+                resampleMethod,
+                interleave: true,
+            });
+        // TODO: Do we still need to catch this error?
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (error: any) {
+            if (error.toString() === 'AggregateError: Request failed') {
+                await new Promise((resolve) => {
+                    setTimeout(resolve, 100);
```
airnez: If you want to keep this, I suppose that you should implement some kind of retry counter, what do you think? And maybe log something to avoid silent fail?
And don't you want to check against `error.code` instead of `error.toString()`?
Don't forget to delete `//TODO` comment above

packages/Main/package.json
```diff
@@ -59,6 +59,7 @@
     "brotli-compress": "^1.3.3",
     "copc": "^0.0.8",
     "earcut": "^3.0.1",
+    "geotiff": "^2.1.4-beta.0",
```
airnez: Do we want to use a beta version ?

packages/Main/src/Parser/GeotiffParser.ts
```diff
@@ -0,0 +1,316 @@
...
+
+    const geotiffNode = new GeotiffNode({ image });
+
+    const texture = <TextureWithExtent> await geotiffNode.extractTexture({
+        // No texture extent is passed. We assume the image covers the tile as
+        // it is the case with all other raster format.
+        textureDimensions: tileRasterDimensions,
+        resampleMethod,
+        defaultAlpha,
+        // TODO: find a way to add pool
```
airnez: Remember to adress this //TODO message

packages/Main/src/Parser/GeotiffParser.ts
```diff
@@ -0,0 +1,316 @@
...
+        newBuffer[index + 2] = typedArray[oldIndex + 2];  // B
+        // Add alpha to new arrayBuffer
+        newBuffer[index + 3] = newAlpha;
+    }
+
+    return newBuffer;
+}
+
+
+class GeotiffNode {
```
airnez, line 132: You could document this class a bit, what do you think ?

packages/Main/src/Parser/CogParser.ts
```diff
@@ -0,0 +1,123 @@
...
+
+import { GeotiffNode } from './GeotiffParser';
+
+import type { TextureWithExtent } from './GeotiffParser';
+
+
+const DEFAULT_MAX_TEXTURE_SIZE = 10 * 1024 * 1024;
+
+
+function selectOverview(
```
airnez, line 26: You could maybe document this function to explain the notion of overview / resampling level

airnez (commented):
Looks good to me overall. I understand that this is still WIP, don't hesitate to request another review later if needed. 🌍🤓
Well done

packages/Main/src/Parser/GeotiffParser.ts
```diff
@@ -0,0 +1,316 @@
...
+        const {
+            window,
+            outputWidth,
+            outputHeight,
+            resampleMethod = 'nearest',
+            pool,
+        } = options;
+
+        try {
+            return <TypedArrayWithDimensions> await this.image.readRasters({
```
ketourneau: Why are you using `readRasters` instead of `readRGB` ? `readRGB` contains logic to handle [PhotometricInterpretation](https://github.com/geotiffjs/geotiff.js/blob/master/src/geotiffimage.js#L687)

examples/cog_loader.html
mgermerie: I removed the commit.

packages/Main/test/unit/cogParser.js
mgermerie: I implemented tests for `CogSource` and `CogParser`. They should be ready now.

packages/Main/package.json
mgermerie: Woops, I missed the beta release while working on it. I reverted to the `2.1.3` version.

packages/Main/src/Parser/GeotiffParser.ts
mgermerie: I tested again and could not reproduce the error. It occurred when I started working on Geotiff but the code evolution must have neutralized it. I removed this part (the whole method since it was just a wrapper around `GeoTIFFImage.readRasters` that managed the error).

packages/Main/src/Parser/GeotiffParser.ts
mgermerie: It was because `readRGB` is specific for images in RGB color space, and causes issues with for instance `Float32` single band data (which we use for elevation). I added [this part](https://github.com/iTowns/itowns/pull/2558/files#diff-62f041f5b1e7ecc708ccece245afdfbdcdde1aa9a6dce155a6a5f842baba31e9R196-R200) to use `readRGB` if data is in RGB color space and `readRasters` otherwise.

mgermerie:
Thanks for your reviews @airnez and @ketourneau! I addressed your comments, the TODOs and added some documentation.

jailln:
@mgermerie could you add document (and if possible to detect them, a warning) for tiff formats that are not supported yet please ?

mgermerie:
> @mgermerie could you add document (and if possible to detect them, a warning) for tiff formats that are not supported yet please ?

Actually after modifications and digging out, I found out that the lack of support is related to how we deal with textures in iTowns more broadly than in the geotiff case. In our shaders, we suppose textures are either `RedFormat` or `RGBAFormat` (both cases are managed in this PR). If my understanding is right (but it might not be), we don't consider cases where textures could have a different format (`RGFormat` or `RedIntegerFormat` for instance).

Therefore, adding the support for geotiff data with for instance two samples per pixels (that should be stored in an `RGFormat` texture) or any specificity corresponding with something else than the `RedFormat` and the `RGBAFormat` would require some change in our shaders.

TLDR: What would be missing for a  comprehensive support of geotiff spec (as long as what `geotiff.js` supports) would be to manage the case of images with two bands. This is however not specific to geotiff but to the way we consider textures in our shaders.

packages/Main/src/Parser/GeotiffParser.ts
airnez: LGTM

packages/Main/src/Parser/CogParser.ts
airnez: Perfect !

airnez (commented):
Looks good to me ! 🤓

airnez (approved):

jailln:
> > @mgermerie could you add document (and if possible to detect them, a warning) for tiff formats that are not supported yet please ?
> 
> Actually after modifications and digging out, I found out that the lack of support is related to how we deal with textures in iTowns more broadly than in the geotiff case. In our shaders, we suppose textures are either `RedFormat` or `RGBAFormat` (both cases are managed in this PR). If my understanding is right (but it might not be), we don't consider cases where textures could have a different format (`RGFormat` or `RedIntegerFormat` for instance).
> 
> Therefore, adding the support for geotiff data with for instance two samples per pixels (that should be stored in an `RGFormat` texture) or any specificity corresponding with something else than the `RedFormat` and the `RGBAFormat` would require some change in our shaders.
> 
> TLDR: What would be missing for a  comprehensive support of geotiff spec (as long as what `geotiff.js` supports) would be to manage the case of images with two bands. This is however not specific to geotiff but to the way we consider textures in our shaders.

Ok, thanks for the detail and the issue you open! I think we should still write in the geotiff doc which formats are supported and which are not (without explaining why), so user that will use this format also get this information. In my opinion it is a general issue in iTowns: we don't expose clearly enough what is really supported (or which parts of the formats are supported and which are not). Users can get frustrated because sometimes it doesn't work and they don't know why, and this could be simply answered by adding more information on what is really supported and what is not.

mgermerie:
> Ok, thanks for the detail and the issue you open! I think we should still write in the geotiff doc which formats are supported and which are not (without explaining why), so user that will use this format also get this information. In my opinion it is a general issue in iTowns: we don't expose clearly enough what is really supported (or which parts of the formats are supported and which are not). Users can get frustrated because sometimes it doesn't work and they don't know why, and this could be simply answered by adding more information on what is really supported and what is not.

Okay, I added the information in the documentation for `CogSource` and `GeotiffParser`.

PR #2511, opened 2025-03-10
chore(deps): bump developer dependencies

AnthonyGlt:
## Description

Bump of the following deps

## Dependencies

- [x] [3d-tiles-renderer](https://github.com/NASA-AMMOS/3DTilesRendererJS) updated to ^0.4.7  
- [x] [@maplibre/maplibre-gl-style-spec](https://github.com/maplibre/maplibre-style-spec) updated to ^23.1.0  
- [x] [@tmcw/togeojson](https://github.com/placemark/togeojson) updated to ^7.0.0  
- [x] [chalk](https://github.com/chalk/chalk) updated to ^5.4.1  
- [x] copc updated to ^0.0.8  
- [x] [earcut](https://github.com/mapbox/earcut) updated to ^3.0.1  
- [x] [lru-cache](https://github.com/isaacs/node-lru-cache) updated to ^11.0.2  

## Developer dependencies

- [x] [@babel/cli](https://github.com/babel/babel) updated to ^7.26.4  
- [x] [@babel/core](https://github.com/babel/babel) updated to ^7.26.9  
- [x] [@babel/plugin-transform-runtime](https://github.com/babel/babel) updated to ^7.26.9  
- [x] [@babel/preset-env](https://github.com/babel/babel) updated to ^7.26.9  
- [x] [@babel/preset-typescript](https://github.com/babel/babel) updated to ^7.26.0  
- [x] [@babel/register](https://github.com/babel/babel) updated to ^7.25.9  
- [x] [@types/proj4](https://github.com/DefinitelyTyped/DefinitelyTyped) updated to ^2.5.6  
- [x] [@types/three](https://github.com/DefinitelyTyped/DefinitelyTyped) updated to ^0.174.0  
- [x] [@typescript-eslint/eslint-plugin](https://github.com/typescript-eslint/typescript-eslint) updated to ^8.26.0  
- [x] [@typescript-eslint/parser](https://github.com/typescript-eslint/typescript-eslint) updated to ^8.26.0  
- [x] [@xmldom/xmldom](https://github.com/xmldom/xmldom) updated to ^0.9.8  
- [x] [babel-loader](https://github.com/babel/babel-loader) updated to ^10.0.0  
- [x] [c8](https://github.com/bcoe/c8) updated to ^10.1.3  
- [x] [chalk](https://github.com/chalk/chalk) updated to ^5.4.1  
- [x] [concurrently](https://github.com/open-cli-tools/concurrently) updated to ^9.1.2  
[...]

PR #2563, opened 2025-06-18
[CI] Fix various issues on our integration pipeline

Desplandis:
## Description

This PR aims to fix various issues on our integration pipeline. That includes:
- deprecated actions (e.g. `actions/create-release@v1`, `actions/upload-release-asset@v1`)
- deprecated GitHub actions commands (e.g. `::set-output`)
- rework of the version bumping (use of `npm version` and post scripts instead of `grunt`)
- fix the current version bumping problem

I identified some other points to fix in the future:
- concurrency problems in some jobs (e.g. two github actions trying to push on branch next)
- a need to debug coveralls problems (which could be linked to the previous point)
- try a dry run before publishing any package (which could prevent half-releases...)
- fix the security issues with our CI: use of unsanitised input at
https://github.com/iTowns/itowns/blob/fecc1f0d78e2a837d864b19b6872c4f983b06149/.github/workflows/integration.yml#L297)
- add a manual GitHub Action to create a new PR to bump itowns versions (because we're all lazy)
- use `amannn/action-semantic-pull-request` for checking that our commits are well-formed instead of our regexp
- fix the changelog script for windows users (and maybe simplify it and use the default conventional commit mode?)


package.json
```diff
@@ -29,11 +30,10 @@
...
     "changelog": "conventional-changelog -n ./config/conventionalChangelog/config.cjs -i changelog.md -s",
-    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm run update-packages && npm install && git add -A && git commit --amend --no-edit",
-    "prerelease-next": "npm version prerelease --preid next && npm run prerelease-next --workspaces --if-present && npm run update-packages",
-    "publish-next": "npm run prerelease-next && npm run prepublishOnly && npm run publish-next --workspaces --if-present",
-    "publish-latest": "npm run prepublishOnly && npm run publish-latest --workspaces --if-present",
-    "update-packages": "npm run update-package --workspaces --if-present"
+    "bump": "cross-env-shell npm version ${npm_config_level:-minor} --workspaces && npm run changelog && git add -A && git commit --amend --no-edit",
+    "prerelease-next": "npm version prerelease --workspaces --preid next",
+    "publish-next": "npm run prerelease-next && npm run prepublishOnly && npm run publish-next --workspaces",
+    "publish-latest": "npm run prepublishOnly && npm run publish-latest --workspaces"
```
gchoqueux: #### On `bump` script
* There isn't new commit, the bumped version is amended in the last commit
#### On `publish-next` and `publish-lates` scripts
* `--if-present` is missing

packages/Debug/package.json
```diff
@@ -34,8 +33,5 @@
     "chart.js": "^4.4.8",
     "dat.gui": "^0.7.9",
     "itowns": "^2.45.1"
-  },
-  "peerDependencies": {
-    "three": "^0.174.0"
```
gchoqueux, line 39: Why you remove three dependency ?

packages/Geographic/package.json
```diff
@@ -18,9 +18,7 @@
     "test-with-coverage": "c8 -n src -r html cross-env npm run test-unit",
     "test-with-coverage_lcov": "c8 -n src --reporter=lcov cross-env npm run test-unit",
     "watch": "npm run transpile -- --watch",
-    "publish-latest": "npm publish --access public --tag=latest --provenance",
-    "prerelease-next": "npm version prerelease --preid next",
-    "publish-next": "npm publish --access public --tag=next --provenance"
+    "publish-latest": "npm publish --access public --tag=latest --provenance"
```
gchoqueux: Missing `publish-next` script

packages/Main/scripts/update-version.mjs
```diff
@@ -0,0 +1,19 @@
+import fs from 'fs';
+
+// This script is used to keep the version in sync between package.json and
+// Main.js, it is called by the npm postversion script.
+
+const version = JSON.parse(fs.readFileSync('./package.json', 'utf8')).version;
+const mainPath = './src/Main.js';
```
gchoqueux: #### Note
That makes me realize there's no `REVISION `constant in `@itowns/geographic.`
We could factor it out into `@itowns/core` in the future

packages/Widgets/package.json
```diff
@@ -32,7 +31,6 @@
   },
   "homepage": "https://itowns.github.io/",
   "dependencies": {
-    "@itowns/geographic": "^2.45.1",
```
gchoqueux: `@itowns/geographic Coordinates` is used in `Minimap`
You’re removing it because it’s private and handled through the monorepo dependencies ?

gchoqueux (commented):
I successfully tested`bump ` and `publish-next` scripts

packages/Debug/package.json
Desplandis: This seems to be what caused issue with dependencies resolution. Since this package depends on a package with a peer dependency, it is implicit that all dependent packages have those peer dependencies. Only the final package (i.e. the application) shall add those peer dependencies as dependencies.

packages/Main/scripts/update-version.mjs
Desplandis: Yup, but do we really need it? I suppose for the whole `itowns` package that may make sense (I image that *geoplatforme* guards against specific itowns versions). However, I don't think we need it for those small packages. Especially if we decide sooner or later to follow semantic versioning.

packages/Widgets/package.json
Desplandis: An error but works since `itowns` depends on `@itowns/geographic`.

packages/Geographic/package.json
Desplandis: Done

package.json
Desplandis: For the **bump** script, you are right I forgot that grump was committing... It would be better to use the fact that `npm version` at the root package does the same thing! I will try to find an elegant solution for this one.

Will fix the `publish-next` and `publish-latest` scripts.

packages/Main/scripts/update-version.mjs
gchoqueux: It's quite common to include a version identifier directly in the code of an NPM package. Since we can't anticipate how users will interact with the package, it would be valuable to expose a constant or property that indicates the current revision or version. This would provide better transparency and traceability for developers relying on the package.

package.json
gchoqueux: I discovered the --workspaces-update option and wanted to give it a try.
I noticed that the bump command doesn't work on Windows (@ftoromanoff ).
I then tried it on Linux, but the dependencies were not updated as expected.

Desplandis: Pushed all fixes, I decided to overload the version command to have the same behavior as Grunt. We now have:
- A `version.mjs` script at toplevel to "augment" the default behavior of npm version by recursively updating the version of all packages in the workspace and run the changelog package script
- A `version.mjs` script at the `itowns` package to "augment" the default behavior of npm version by updating the version of the Main.js file.

Both commands add the touched files to the git index.

I tested a whole part of this PR on windows (this is the reason for the node mjs scripts instead of command on package.json). However, certain `bash` constructs like (e.g. `if [] fi;`, `${var:-default}`) are not portable to `cmd.exe` (unless we remove the default argument in the `bump` script). Moreover, the `changelog` script does not seems to work on windows (but will not be fixed on this PR).

Desplandis:
@gchoqueux Shall be all good!

Desplandis:
Added some suggestions for improving the integration GH Action

package.json
```diff
@@ -29,11 +30,11 @@
     "prepare": "cross-env NO_UPDATE_NOTIFIER=true node ./config/prepare.mjs && node ./config/replace.config.mjs",
     "watch": "node ./config/concurrently.mjs watch",
     "changelog": "conventional-changelog -n ./config/conventionalChangelog/config.cjs -i changelog.md -s",
-    "bump": "if [ -z $npm_config_level ]; then grunt bump:minor;  else grunt bump:$npm_config_level; fi && npm run changelog && npm run update-packages && npm install && git add -A && git commit --amend --no-edit",
-    "prerelease-next": "npm version prerelease --preid next && npm run prerelease-next --workspaces --if-present && npm run update-packages",
+    "bump": "npm version ${npm_config_level:-minor} -m \"release v%s\"",
+    "version": "node ./config/version.mjs",
```
gchoqueux, line 34: * Have you the issue : changelog.md is immutable file ?
* maybe rename, `version `script, to  `workspaces_version`
* ~The version  script doesn't seem to work on Windows.~

Desplandis: - Got the issue with `changelog.md` being an immuable file on windows (not on Linux) but it was already the case pre-PR
- Since I want to "override" the npm version command, I use the `version` script instead of a custom one
- The stand-alone `version` script works on windows (since this is were I developed those changes), but the `bump` script do not (for reasons explained in a previous comment)

**Edit**: Got the `changelog` issue on linux but even with the master branch... Seems to be because `conventional-changelog` was bumped at version `5.0` instead of `4.1.0`. I remember that @mgermerie encountered a similar issue when bumping our dependencies and decided to not bump `conventional-changelog`. Was bumped in #2511. Will push a fix to revert to the `4.1.0` version.

Desplandis:
@gchoqueux I fixed the changelog problem. However I will not fix the bump support for windows unless we change the way it is called (i.e. `npm run bump -- minor` instead of `npm_config_level=minor npm run bump`).

package.json
gchoqueux: I have this message on Linux : _`Error in conventional-changelog-writer: Cannot modify immutable object.`_

Desplandis: Even after a `npm install`?

gchoqueux: it works now my message was before the downgrade commit

gchoqueux: one remark but it's not very important `npm run version` doesn't work because of environment variables

Desplandis: We should be using `npm version` instead of `npm run version`. But I could add a check in both `version.mjs` to check if it is defined.

Desplandis: Does everything works tho?

Desplandis:
@gchoqueux Rebased on master (no conflicts anymore). This will be the last review/test I think! :D

gchoqueux:
@Desplandis I do the final check

gchoqueux:
Check script in progress
- [x] `bump`
- [x] `version`
- [x] `prerelease-next`
- [x] `publish-next`
- [x] `publish-latest`
- [ ] ~publish release~ for the release, I couldn't check `publish-next` in good state because I test `publish-latest` with `--dry-run option`

gchoqueux (approved):
Thx, it's good for me, see [test](https://github.com/iTowns/itowns/pull/2563#issuecomment-3023842005)

Desplandis:
Alea jacta est

Desplandis:
🥳

PR #2572, opened 2025-07-01
fix(types): resolve absolute imports to relative paths in .d.ts

Desplandis:
## Description

Since #2552, we use the `tsc` command to generate type definitions in the `itowns` sub-package. However `tsc` does not resolve absolute imports (with a base directory) when generating `.d.ts` files.

This PR uses [tsc-alias](https://www.npmjs.com/package/tsc-alias) to replace alias paths with relative paths after a first typescript compilation pass.

P.S. I saw that #2552 that the "Main" `package.json` does not have a "types" field, so the type are not available to user of the `itowns` library. Is this the intended behavior? @HoloTheDrunk @ketourneau

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "itowns",
+  "name": "itowns-1",
   "version": "2.45.1",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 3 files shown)


HoloTheDrunk (approved):
The lack of a `"types"` field in the `package.json` is not intentional but should probably be added to all the subpackages as well.

Desplandis:
Yeah but we'll have to quantify the quality of type declarations for the `itowns` subpackage (doing this now). And I also think this is better than nothing. I will add the "types" field for the `itowns` subpackage. WDYT?

P.S.: It is already available for `@itowns/geographic`.

Issue #2561, opened 2025-06-13
[REGRESSION] We can't change the opacity of the globe


AnthonyGlt:
Since https://github.com/iTowns/itowns/pull/2374/files#diff-2cbabdb42b34b51f54a3ad1eabf469e8990ea4073b43780ff5f2cc6709e08847

It seems that we can't edit the opactiy of the globe anymore

## Steps to Reproduce

1. Use the following example code
2. Click on the button "Change Tile Layer Opacity"

## Expected Behavior

![Image](https://github.com/user-attachments/assets/2a10a5b7-fea8-42ef-bd7b-522d38cd9ae1)

## Actual Behavior

![Image](https://github.com/user-attachments/assets/fd352eba-8754-42ed-9931-666fa96009f2)

```html

<html>
<head>
    <title>Itowns - Globe</title>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <link rel="stylesheet" type="text/css" href="css/example.css">
    <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
    <link rel="stylesheet" type="text/css" href="css/widgets.css">
    <script src="https://cdnjs.cloudflare.com/ajax/libs/dat-gui/0.7.6/dat.gui.min.js"></script>
    <style>
        #removeElevationBtn {
            position: absolute;
            top: 10px;
            left: 10px;
            z-index: 1000;
            padding: 10px 15px;
            background-color: #ff6b6b;
            color: white;
            border: none;
            border-radius: 5px;
            cursor: pointer;
            font-size: 14px;
        }
        #removeElevationBtn:hover {
            background-color: #ff5252;
        }
        #removeElevationBtn:disabled {
            background-color: #cccccc;
            cursor: not-allowed;
        }
        #opacityBtn {
            position: absolute;
            top: 60px;
            left: 10px;
            z-index: 1000;
            padding: 10px 15px;
            background-color: #4caf50;
            color: white;
            border: none;
            border-radius: 5px;
            cursor: pointer;
            font-size: 14px;
        }
        #opacityBtn:hover {
            background-color: #45a049;
        }
    </style>
</head>
<body>
<div id="viewerDiv"></div>
<button id="removeElevationBtn" onclick="removeElevationLayer()">Remove Elevation Layer</button>
<button id="opacityBtn" onclick="changeTileLayerOpacity()">Change Tile Layer Opacity</button>

<!-- Import iTowns source code -->
[...]

PR #2562, opened 2025-06-13
fix(TileLayer): transparence set to true by default

ftoromanoff:
Transparence of material for Tile layer set to true by default to allow opacity variation

PR #2577, opened 2025-07-07
fix(OGC3DTilesLayer): renderOrder set at 1

ftoromanoff:
Linked to reopening of the issue #2561

### DIFF
--- a/packages/Main/src/Layer/OGC3DTilesLayer.js
+++ b/packages/Main/src/Layer/OGC3DTilesLayer.js
@@ -429,6 +429,7 @@ class OGC3DTilesLayer extends GeometryLayer {
         }
 
         model.material = material;
+        model.renderOrder = 1;
     }
 
     /**


packages/Main/src/Layer/OGC3DTilesLayer.js
```diff
@@ -429,6 +429,7 @@ class OGC3DTilesLayer extends GeometryLayer {
         }
 
         model.material = material;
+        model.renderOrder = 1;
```
jailln, line 432: Why do we need to hardcode a default renderOrder for all 3DTiles layers? This seems restrictive and may lead to other bugs and limitations I think

Desplandis: We should not hardcode the `renderOrder` for each 3D objects that we draw... Or we'll have to open an issue for every 3D objects added to our scene... I think we should revert #2562 which kinda do the same bugfix and go back to the original PR #2374 that added the bug (if we can't fix it, we will have to revert it too).

Desplandis: See #2584

Desplandis:
Closing it due to #2584. Reopening the original issue.

PR #2578, opened 2025-07-08
Pack textures in texture arrays to reduce number of texture slots per layer

### LINKED ISSUE #2568: Increase number raster layers that can be added to a view
## Context

Currently, the number of layers we can add to a view is limited for several reasons:
1. This [hard limit](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Renderer/LayeredMaterial.ts#L27) of 15.
2. The max number of texture units supported by the gpu, computed [here](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Core/System/Capabilities.js#L35) (i.e. max number of textures that can be passed to the shader in one rendering pass).
3. The [following method](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Core/Prefab/Globe/GlobeLayer.js#L104) that counts the number of textures per tile.

Current GPUs with WebGL2 generally have a max number of texture units >= 32. The method 3 is quite conservative by approximating the number of textures based on the crs and by taking the worst case scenario. In this context, the limit is quickly reached: between 5 to 10 layers from my tests, which is quite limiting for users. 

## Description of the proposal

The proposal is to overcome these limits, by implementing a method to bind more textures to the layer.

### Identified use-cases

Adding more layers to the view.

## Implementation

This is only a proposition and should be discussed / specified.

Regarding the methods to pass more textures to the shader, I've found two possibilities:
* Use texture arrays (see [this tutorial](https://medium.com/better-programming/how-to-use-texture-arrays-in-webgl-921dff1c22d8) for instance)
* Create a texture atlas

Texture arrays are available since WebGL 2 and iTowns supports WebGL 2 and dropped WebGL 1 support so it seems to be a good option. However, mind the [following issue](https://github.com/mrdoob/three.js/issues/28020) (that may not be a concern since we have our [own implementation of loop unrolling](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Renderer/Shader/ShaderUtils.js#L36) in iTowns).

Regarding the above-mentionned limits:

* Limit 1. : I'm not sure it is still meaningfull. The getColorAtIdUv method is not used anymore and I don't really understand why there would be such a limit of 15. I tried removing it and it works well. I'm curious on having your opinion on this @Desplandis @HoloTheDrunk @gchoqueux 
* Limit 2. : seems legit. Could be removed if we use texture arrays though ? If we use texture atlases, we may use it to know when to start creating atlases (when the number of textures added to a tile is greater than this limit).
* Limit 3. : This computation seems quite conservative and we could remove it and compute the real number of texture per material in [LayeredMaterial.addColorTile](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Renderer/LayeredMaterial.ts#L490) if we use texture atlases. In the case of texture arrays I think we won't need it.

Neptilo:
## Description
In order to increase the number of raster layers that can be added to a `LayeredMaterial`, we assemble its textures into a texture arrays. For performance, the operation of copying the texture data to a texture array is done on the GPU.

## Motivation and Context
The number of layers we can add to a view is currently very limited, and users quickly hit that limit.
These changes fix issue #2568.
Tested on Firefox, Ubuntu.

This is work in progress: some functional tests have to be updated, and the use of the global variable `view` has to be changed.

### DIFF
--- a/packages/Main/src/Core/TileMesh.js
+++ b/packages/Main/src/Core/TileMesh.js
@@ -106,8 +106,13 @@ class TileMesh extends THREE.Mesh {
         }
     }
 
-    onBeforeRender() {
-        this.material.updateLayersUniforms();
+    /**
+     * An optional callback that is executed immediately before a 3D object is rendered.
+     *
+     * @param {THREE.WebGLRenderer} renderer - The renderer used to render textures.
+     */
+    onBeforeRender(renderer) {
+        this.material.updateLayersUniforms(renderer);
     }
 }
 
--- a/packages/Main/src/Core/TileMesh.js
+++ b/packages/Main/src/Core/TileMesh.js
@@ -112,7 +112,9 @@ class TileMesh extends THREE.Mesh {
      * @param {THREE.WebGLRenderer} renderer - The renderer used to render textures.
      */
     onBeforeRender(renderer) {
-        this.material.updateLayersUniforms(renderer);
+        if (this.material.layersNeedUpdate) {
+            this.material.updateLayersUniforms(renderer);
+        }
     }
 }
 
--- a/packages/Main/src/Core/View.js
+++ b/packages/Main/src/Core/View.js
@@ -7,7 +7,6 @@ import { COLOR_LAYERS_ORDER_CHANGED } from 'Renderer/ColorLayersOrdering';
 import c3DEngine from 'Renderer/c3DEngine';
 import RenderMode from 'Renderer/RenderMode';
 import FeaturesUtils from 'Utils/FeaturesUtils';
-import { getMaxColorSamplerUnitsCount } from 'Renderer/LayeredMaterial';
 import Scheduler from 'Core/Scheduler/Scheduler';
 import Picking from 'Core/Picking';
 import LabelLayer from 'Layer/LabelLayer';
@@ -177,6 +176,7 @@ class View extends THREE.EventDispatcher {
         // options.renderer can be 2 separate things:
         //   - an actual renderer (in this case we don't use viewerDiv)
         //   - options for the renderer to be created
+        // here renderer is passed to engine
         if (options.renderer && options.renderer.domElement) {
             engine = new c3DEngine(options.renderer);
         } else {
@@ -348,17 +348,8 @@ class View extends THREE.EventDispatcher {
             if (layer.isColorLayer) {
                 const layerColors = this.getLayers(l => l.isColorLayer);
                 layer.sequence = layerColors.length;
-
-                const sumColorLayers = parentLayer.countColorLayersTextures(...layerColors, layer);
-
-                if (sumColorLayers <= getMaxColorSamplerUnitsCount()) {
-                    parentLayer.attach(layer);
-                } else {
-                    return layer._reject(new Error(`Cant add color layer ${layer.id}: the maximum layer is reached`));
-                }
-            } else {
-                parentLayer.attach(layer);
             }
+            parentLayer.attach(layer);
         } else {
             if (typeof (layer.update) !== 'function') {
                 return layer._reject(new Error('Cant add GeometryLayer: missing a update function'));
--- a/packages/Main/src/Core/View.js
+++ b/packages/Main/src/Core/View.js
@@ -176,7 +176,6 @@ class View extends THREE.EventDispatcher {
         // options.renderer can be 2 separate things:
         //   - an actual renderer (in this case we don't use viewerDiv)
         //   - options for the renderer to be created
-        // here renderer is passed to engine
         if (options.renderer && options.renderer.domElement) {
             engine = new c3DEngine(options.renderer);
         } else {
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -153,7 +153,6 @@ function updateLayersUniforms<Type extends 'c' | 'e'>(
     if (cachedTexture) {
         uTextures.value = cachedTexture;
         uTextureCount.value = count;
-        renderer.initTexture(cachedTexture);
         return;
     }
 
@@ -223,16 +222,12 @@ interface LayeredMaterialRawUniforms {
 
     // Elevation layers
     elevationLayers: Array<StructElevationLayer>;
-    // can be [] on initialization
-    // so we don't have to instantiate an empty DataArrayTexture
     elevationTextures: THREE.DataArrayTexture | null;
     elevationOffsetScales: Array<THREE.Vector4>;
     elevationTextureCount: number;
 
     // Color layers
     colorLayers: Array<StructColorLayer>;
-    // can be [] on initialization
-    // so we don't have to instantiate an empty DataArrayTexture
     colorTextures: THREE.DataArrayTexture | null;
     colorOffsetScales: Array<THREE.Vector4>;
     colorTextureCount: number;
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -158,6 +158,13 @@ function updateLayersUniforms<Type extends 'c' | 'e'>(
         return;
     }
 
+    // If the textures are the original array obtained when loaded,
+    // they are not yet handled by our cache, so dispose of them
+    // before replacing them with data array textures.
+    if (!(uTextures.value instanceof THREE.DataArrayTexture)) {
+        for (const texture of uTextures.value) { texture.dispose(); }
+    }
+
     if (!makeDataArrayTexture(uTextures, width, height, count, tiles, max, renderer)) {
         uTextureCount.value = 0;
         return;
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -6,7 +6,7 @@ import Capabilities from 'Core/System/Capabilities';
 import RenderMode from 'Renderer/RenderMode';
 import { LRUCache } from 'lru-cache';
 import { RasterTile, RasterElevationTile, RasterColorTile } from './RasterTile';
-import { makeDataArrayTexture } from './WebGLComposer';
+import { makeDataArrayRenderTarget } from './WebGLComposer';
 
 const identityOffsetScale = new THREE.Vector4(0.0, 0.0, 1.0, 1.0);
 
@@ -78,9 +78,9 @@ const defaultStructLayers: Readonly<{
     },
 };
 
-const textureArraysCache = new LRUCache<string, THREE.DataArrayTexture>({
+const rtCache = new LRUCache<string, THREE.WebGLArrayRenderTarget>({
     max: 200,
-    dispose: (texture) => { texture.dispose(); },
+    dispose: (rt) => { rt.dispose(); },
 });
 
 /**
@@ -149,25 +149,27 @@ function updateLayersUniforms<Type extends 'c' | 'e'>(
         }
     }
 
-    const cachedTexture = textureArraysCache.get(textureSetId);
-    if (cachedTexture) {
-        uTextures.value = cachedTexture;
+    const cachedRT = rtCache.get(textureSetId);
+    if (cachedRT) {
+        uTextures.value = cachedRT.texture;
         uTextureCount.value = count;
         return;
     }
 
     // Memory management of these textures is only done by `textureArraysCache`,
     // so we don't have to dispose of them manually.
-    if (!makeDataArrayTexture(uTextures, width, height, count, tiles, max, renderer)) {
+    const rt = makeDataArrayRenderTarget(width, height, count, tiles, max, renderer);
+    if (!rt) {
         uTextureCount.value = 0;
         return;
     }
 
-    textureArraysCache.set(textureSetId, uTextures.value);
+    rtCache.set(textureSetId, rt);
+    uniforms.textures.value = rt.texture;
 
     if (count > max) {
         console.warn(
-            `LayeredMaterial: Not enough texture units (${max} < ${count}),`
+            `LayeredMaterial: Not enough texture units (${max} < ${count}), `
             + 'excess textures have been discarded.',
         );
     }
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -4,6 +4,7 @@ import TileFS from 'Renderer/Shader/TileFS.glsl';
 import ShaderUtils from 'Renderer/Shader/ShaderUtils';
 import Capabilities from 'Core/System/Capabilities';
 import RenderMode from 'Renderer/RenderMode';
+import { LRUCache } from 'lru-cache';
 import { RasterTile, RasterElevationTile, RasterColorTile } from './RasterTile';
 import { makeDataArrayTexture } from './WebGLComposer';
 
@@ -78,8 +79,10 @@ const defaultStructLayers: Readonly<{
     },
 };
 
-const textureArraysCache = new Map<string, THREE.DataArrayTexture>();
-const TEXTURE_ARRAY_CACHE_CAPACITY = 200;
+const textureArraysCache = new LRUCache<string, THREE.DataArrayTexture>({
+    max: 200,
+    dispose: (texture) => { texture.dispose(); },
+});
 
 /**
  * Updates the uniforms for layered textures,
@@ -147,10 +150,11 @@ function updateLayersUniforms<Type extends 'c' | 'e'>(
         }
     }
 
-    if (textureArraysCache.has(textureSetId)) {
-        uTextures.value = textureArraysCache.get(textureSetId);
+    const cachedTexture = textureArraysCache.get(textureSetId);
+    if (cachedTexture) {
+        uTextures.value = cachedTexture;
         uTextureCount.value = count;
-        renderer.initTexture(uTextures.value);
+        renderer.initTexture(cachedTexture);
         return;
     }
 
@@ -159,11 +163,6 @@ function updateLayersUniforms<Type extends 'c' | 'e'>(
         return;
     }
 
-    if (textureArraysCache.size >= TEXTURE_ARRAY_CACHE_CAPACITY) {
-        const oldestEntry = textureArraysCache.entries().next().value!;
-        oldestEntry[1].dispose();
-        textureArraysCache.delete(oldestEntry[0]);
-    }
     textureArraysCache.set(textureSetId, uTextures.value);
 
     if (count > max) {
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -89,13 +89,15 @@ const TEXTURE_ARRAY_CACHE_CAPACITY = 200;
  * @param uniforms - The uniforms object for your material.
  * @param tiles - An array of RasterTile objects, each containing textures.
  * @param max - The maximum number of layers for the DataArrayTexture.
+ * @param type - Layer set identifier: 'c' for color, 'e' for elevation.
+ * @param renderer - The renderer used to render textures.
  */
 function updateLayersUniforms<Type extends 'c' | 'e'>(
     uniforms: { [name: string]: THREE.IUniform },
     tiles: RasterTile[],
     max: number,
     type: Type,
-) {
+    renderer: THREE.WebGLRenderer) {
     // Aliases for readability
     const uLayers = uniforms.layers.value;
     const uTextures = uniforms.textures;
@@ -148,12 +150,11 @@ function updateLayersUniforms<Type extends 'c' | 'e'>(
     if (textureArraysCache.has(textureSetId)) {
         uTextures.value = textureArraysCache.get(textureSetId);
         uTextureCount.value = count;
-        const renderer: THREE.WebGLRenderer = view.renderer;
         renderer.initTexture(uTextures.value);
         return;
     }
 
-    if (!makeDataArrayTexture(uTextures, width, height, count, tiles, max)) {
+    if (!makeDataArrayTexture(uTextures, width, height, count, tiles, max, renderer)) {
         uTextureCount.value = 0;
         return;
     }
@@ -479,7 +480,7 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
         };
     }
 
-    public updateLayersUniforms(): void {
+    public updateLayersUniforms(renderer: THREE.WebGLRenderer): void {
         const colorlayers = this.colorTiles
             .filter(rt => rt.visible && rt.opacity > 0);
         colorlayers.sort((a, b) =>
@@ -491,6 +492,7 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
             colorlayers,
             this.defines.NUM_FS_TEXTURES,
             'c',
+            renderer,
         );
 
         if (this.elevationTileId !== undefined && this.getElevationTile()) {
@@ -500,6 +502,7 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
                     [this.elevationTile],
                     this.defines.NUM_VS_TEXTURES,
                     'e',
+                    renderer,
                 );
             }
         }
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -78,6 +78,9 @@ const defaultStructLayers: Readonly<{
     },
 };
 
+const textureArraysCache = new Map<string, THREE.DataArrayTexture>();
+const TEXTURE_ARRAY_CACHE_CAPACITY = 200;
+
 /**
  * Updates the uniforms for layered textures,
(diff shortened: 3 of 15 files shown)


packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,181 @@
...
+    max: number,
+): boolean {
+    // If no textures found, dispose previous DataArrayTexture and reset
+    if (count === 0) {
+        if (uTextures.value instanceof THREE.DataArrayTexture) { uTextures.value.dispose(); }
+        uTextures.value = new THREE.DataArrayTexture();
+        return false;
+    }
+
+    const renderer: THREE.WebGLRenderer = view.renderer;
```
Neptilo: This is a reference to a global variable. What's the best way to access the renderer without it?

packages/Main/src/Core/View.js
```diff
@@ -348,17 +348,8 @@ class View extends THREE.EventDispatcher {
             if (layer.isColorLayer) {
                 const layerColors = this.getLayers(l => l.isColorLayer);
                 layer.sequence = layerColors.length;
-
-                const sumColorLayers = parentLayer.countColorLayersTextures(...layerColors, layer);
```
Desplandis, line 352: Could you remove the `TiledGeometryLayer#countColorLayersTextures` (and `GlobeLayer` too) as well as the test?

Neptilo:
I did some comparative performance tests:

Each time, I waited for the Earth textures to load before starting the recording (because if I tried to start it as soon as possible during loading, it would be unreliable). Then I zoomed in on France and tilted the view to see the horizon.
My test results:
* On the upstream branch:
  * Calls to `renderScene` take 7.5–11% (for 1 layer) to 29% of the time (for 5 layers).
  * It spends most of its time on `setProgram`: 6.2–24% of the total time, again depending on the number of layers. (In turn, this function spends most of its time on `texSubImage2D`: 2.5 to 6.4%).
  * It spends 0.4 to 0.7% on `updateLayersUniforms`.
* On this branch:
  * `renderScene` always takes between 11 and 12%, regardless of the number of layers.
  * It spends the most time on `updateLayersUniforms`: 4.2–10% (1 layer) to 7–16% (20 layers) of total time. (In turn, this function spends the most time on my `makeDataArrayTexture` function: 2.2–7.9%).
  * It spends 3.6 to 5.6% on `setProgram`

Actually, I'm not sure to what extent these values are comparable, since the longer the textures took to load, the longer I had to wait, and the more rendering loops there were, which inevitably influenced the % of `renderScene`.

So instead, I then tested the absolute average time taken by the render function. The test scenario was now freely exploring the Earth for a long time, until the average times seemed to converge (and even then the second digit did not always stabilize, so that digit may not be significant).
Here are the results:
* Upstream branch : 2.9 ms, for 1 or 5 layers.
* This branch : 3.7 ms for 1 layer, 3.5 ms for 5, 4 ms for 10, 5 ms for 20.

Neptilo:
Code used for benchmarking:
```js
        let totalTime = 0;
        let frameCount = 0;
        this.renderView = function _(view) {
            this.renderer.clear();
            if (view._camXR) {
                this.renderer.render(view.scene, view._camXR);
            } else {
                const start = performance.now();
                this.renderer.render(view.scene, view.camera3D);
                const end = performance.now();

                totalTime += end - start;
                frameCount++;

                if (frameCount % 60 === 0) {
                    const avg = totalTime / frameCount;
                    console.log(`Average render time: ${avg.toFixed(1)} ms`);
                }
            }
            if (view.tileLayer) {
                this.label2dRenderer.render(view.tileLayer.object3d, view.camera3D);
            }
        }.bind(this);
```
This replaces the `this.renderView` block in `c3DEngine.js`. The idea is to use the program as a user would, until converging to an average render time with enough significant digits.

Desplandis:
The minimap in `view_3d_map` does not seem to work on this branch.

Neptilo:
> The minimap in `view_3d_map` does not seem to work on this branch.

OK, I figured it out. It is related to the `view` global variable problem (see my comment in `WebGLComposer.ts`). The problem is that in that example, `view` is set to the globe view, so when rendering the minimap, it also tries to use that same globe view, instead of its own planar view.
Once we decide what to do to avoid using this global, the minimap will work again.

So I'm waiting for suggestions on what is the best way to do that. Passing the view as an argument in all the intermediate functions? Passing it as part of a `context` wrapper, again passing that context as an argument to the intermediate functions? (This is what Giro3D does.) Something else?

packages/Main/src/Core/View.js
```diff
@@ -177,6 +176,7 @@ class View extends THREE.EventDispatcher {
         // options.renderer can be 2 separate things:
         //   - an actual renderer (in this case we don't use viewerDiv)
         //   - options for the renderer to be created
+        // here renderer is passed to engine
```
Desplandis: Is this a leftover of a personal comment?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -109,11 +122,49 @@ function updateLayersUniforms(
...
+                return;
+            }
         }
     }
+
+    if (textureArraysCache.has(textureSetId)) {
+        uTextures.value = textureArraysCache.get(textureSetId);
+        uTextureCount.value = count;
+        const renderer: THREE.WebGLRenderer = view.renderer;
+        renderer.initTexture(uTextures.value);
```
Desplandis: The texture shall already have been uploaded and binded at this point.

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,223 @@
...
+        void main() {
+            gl_FragColor = texture2D(sourceTexture, vUv);
+        }
+    `,
+};
+
+let renderTarget: THREE.WebGLRenderTarget | null = null;
+let material: THREE.ShaderMaterial | null = null;
+let quadScene: THREE.Scene | null = null;
+let quadCam: THREE.OrthographicCamera | null = null;
```
Desplandis: Since this is constant across the different calls to `drawTextureLayers` and is just a bunch of properties, you could make it local to the module.
```suggestion
const quadCam: THREE.OrthographicCamera = new THREE.OrthographicCamera(-1, 1, 1, -1, 0, 1);;
```

packages/Main/src/Renderer/WebGLComposer.ts
Desplandis: You can get a reference to the renderer from `Object3D#onBeforeRender` (or `TileMesh` in our case). Since `LayeredMaterial#updateLayersUniforms` is called from this method, you could pass the renderer there.

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,223 @@
...
+    // COLOR_ATTACHMENT0
+    if ('framebufferTextureLayer' in gl) {
+        gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0,
+            dataArrayTextureWebGL, 0, layerIndex);
+    } else {
+        console.error('framebufferTextureLayer function not supported');
+        return;
+    }
+
+    renderer.render(quadScene, quadCam);
```
Desplandis: There is no need to instantiate a `scene`, you can directly use the quad mesh there.

jailln:
@Neptilo regarding the view variable, we could use `_preprocessLayer` (in `View.js`) that is called by `view.addLayer` and does such things as passing the view to the layer.

We do something similar for `OGC3DTilesLayer` : [here](https://github.com/iTowns/itowns/blob/60a94655f601c273252efce54f35d4ae458af41a/packages/Main/src/Core/View.js#L104-L106) and [here](https://github.com/iTowns/itowns/blob/60a94655f601c273252efce54f35d4ae458af41a/packages/Main/src/Layer/OGC3DTilesLayer.js#L396-L422)

This way the renderer can be stored in the `Layer` and passed to `convertToTile` which instanciates the `LayeredMaterial`. 

WDYT?

Desplandis:
@Neptilo @jailln Damn, I replied through the comments but since I'm currently in review, those are hidden...

An easiest method in my opinion is to get a reference to the renderer from `Object3D#onBeforeRender` (or `TileMesh` in our case). Since `LayeredMaterial#updateLayersUniforms` is called from this method, you could pass the renderer there and have access to it in `WebGLComposer`.

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,223 @@
...
+            uniforms: {
+                // This uniform will be updated with each source 2D texture
+                sourceTexture: { value: null },
+            },
+            vertexShader: copyTextureShader.vertexShader,
+            fragmentShader: copyTextureShader.fragmentShader,
+        });
+        const quad = new THREE.Mesh(geometry, material);
+        quadScene.add(quad);
+    }
```
Desplandis: [`FullscreenQuad`](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/postprocessing/Pass.js#L138) from `three/addons/postprocessing/Pass.js` provides a wrapper over this. You could optionally use it.

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,223 @@
...
+    gl.texParameteri(gl.TEXTURE_2D_ARRAY, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+
+    // Create a temporary THREE.WebGLRenderTarget.
+    // This render target's internal framebuffer
+    // will be used to attach layers.
+    if (!renderTarget) {
+        renderTarget = new THREE.WebGLRenderTarget(width, height, {
+            depthBuffer: false, // No depth buffer needed for simple 2D texture copy
+        });
+    }
```
Desplandis: I think there may be a way to have a pure `three.js` equivalent to this. There exists [`WebGLArrayRenderTarget`](https://threejs.org/docs/index.html?q=WebGL#api/en/renderers/WebGLArrayRenderTarget) which holds a `DataArrayTexture`. The `DataArrayTexture` is [initialized](https://github.com/mrdoob/three.js/blob/0af9729d0c143a86a1d725d6e2c3ad83301f3f34/src/renderers/WebGLArrayRenderTarget.js#L39C22-L39C38), [created](https://github.com/mrdoob/three.js/blob/ae1ecc93499935619e5643667bd8921f015b6f3e/src/renderers/webgl/WebGLTextures.js#L1846), [bound](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLTextures.js#L548) and [allocated](https://github.com/mrdoob/three.js/blob/ae1ecc93499935619e5643667bd8921f015b6f3e/src/renderers/webgl/WebGLTextures.js#L1127) similarly.

We could then simplify as something like this:
```js
    const renderTarget = new THREE.WebGLArrayRenderTarget(width, height, count, {
        depthBuffer: false,
    });
    // ... Fullscreen quad initialization
    let currentLayerIndex = 0;
    for (const tile of tiles) {
        for (
            let i = 0;
            i < tile.textures.length && currentLayerIndex < max;
            ++i, ++currentLayerIndex
        ) {
            const texture = tile.textures[i];
            if (!texture) { continue; }

            // Set the current source 2D texture on the quad's material
            quad.material.uniforms.sourceTexture.value = texture;

            // render this source texture into the current layer
            drawTextureLayer(renderer, renderTarget, currentLayerIndex, sceneQuad);
        }
    }
```

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (count === 0) { return false; }
+
+    const renderer: THREE.WebGLRenderer = view.renderer;
+    const gl = renderer.getContext();
+    if (!('TEXTURE_2D_ARRAY' in gl && 'texStorage3D' in gl)) {
+        console.error('Some WebGL features are missing');
+        return false;
+    }
+
+    uTextures.value = new THREE.DataArrayTexture(null, width, height, count);
```
Desplandis: Do we dispose the previous `DataArrayTexture`?

packages/Main/src/Renderer/WebGLComposer.ts
Desplandis: Follow-up of my previous comment about the usage of `WebGLArrayRenderTarget`, when using `WebGLRenderer#setRenderTarget` we could pass a second argument `activeCubeFace` which:
> [...] Indicates the z layer to render in to when using 3D or **array render targets**.

The code in [`WebGLRenderer#setRenderTarget`](https://github.com/mrdoob/three.js/blob/eac5acce91a6d91fc256cbd99fbbdf2d392b7652/src/renderers/WebGLRenderer.js#L2827) calls `glFrameBufferTextureLayer` for `WebGLArrayRenderTarget`.

This could greatly simplifies the code:
```js
    const previousRenderTarget = renderer.getRenderTarget();
    renderer.setRenderTarget(renderTarget, layerIndex); // calls gl.framebufferTextureLayer behind the scene
    renderer.render(quadScene, quadCam);
    renderer.setRenderTarget(previousRenderTarget);
```

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -109,11 +122,49 @@ function updateLayersUniforms(
...
+        return;
+    }
+
+    if (!makeDataArrayTexture(uTextures, width, height, count, tiles, max)) {
+        uTextureCount.value = 0;
+        return;
+    }
```
Desplandis: Would it be better to separate the texture initialization (when we have the same number of textures) from the render into the texture? And what about the previous texture, is it disposed?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -109,11 +122,49 @@ function updateLayersUniforms(
...
+
+    if (textureArraysCache.size >= TEXTURE_ARRAY_CACHE_CAPACITY) {
+        const oldestEntry = textureArraysCache.entries().next().value!;
+        oldestEntry[1].dispose();
+        textureArraysCache.delete(oldestEntry[0]);
+    }
+    textureArraysCache.set(textureSetId, uTextures.value);
```
Desplandis: You could use `lru-cache` used in some places in iTowns if you wish to use a least-recently used strategy.

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -177,7 +228,7 @@ interface LayeredMaterialRawUniforms {
 
     // Color layers
     colorLayers: Array<StructColorLayer>;
-    colorTextures: Array<THREE.Texture>;
+    colorTextures: THREE.DataArrayTexture;
```
Desplandis: You should do the same for `elevationLayers` above.

Desplandis (changes requested):
Thanks for the feature, this is long overdue. =)
I took a little bit time to review this PR since I wanted to see if there was a way to write a pure `three.js` equivalent. I looked to three.js internals and I think `WebGLArrayRenderTarget` can do something similar to your WebGL calls, I documented all of this in the comments. I hope I didn't miss anything, feel free to ask me if some comments are not clear enough.

P.S.: When comparing to the render in master, it seems to me that texture are wobbly when moving the camera. Maybe an issue with the filtering?

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: I learned something, thanks!

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: Good solution. And it does fix the minimap!

packages/Main/src/Core/View.js
Neptilo: Yes. I missed it

packages/Main/src/Core/View.js
Neptilo: Also remove the test entirely? It also checks something related to the maximum GPU texture units, which still exists...

packages/Main/src/Renderer/LayeredMaterial.ts
Neptilo: This line was actually necessary. See commit abcd496bcc7b8fed1a65df216c07536ce2177eff where it was added.

packages/Main/src/Renderer/LayeredMaterial.ts
Neptilo: Not sure I understood. If you mean separating the `for` loop in `makeDataArrayTexture`, I don´t see what the advantage would be.
Regarding disposing of textures: I'm adding a conditional `dispose` before this, with an explanatory comment.
I also found a case where a texture is created without being added to the cache (if `glFormat < 0`), but it is a pathological case that shouldn´t happen, and if it happens, the layer will be broken anyway. I'm still adding a `dispose` there for good measure.

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: Let's centralize the conversation about `dispose` in the other related comment.

packages/Main/src/Core/View.js
Desplandis: No just remove the part related to `countColorLayersTextures`.

packages/Main/src/Renderer/LayeredMaterial.ts
Desplandis: I was thinking about splitting the texture allocation from the texture drawing code (i.e. have two calls here). I think we should reallocate when the textures count changes (or whatever heuristic we choose related to the textures count) but redraw our textures to the texture array when necessary. It may be premature optimization BTW.

Neptilo:
> When comparing to the render in master, it seems to me that texture are wobbly when moving the camera. Maybe an issue with the filtering?

I didn't notice anything special... What do you call "the filtering"?

packages/Main/src/Renderer/LayeredMaterial.ts
Neptilo: ...but not anymore with the use of array render targets!

packages/Main/src/Core/TileMesh.js
```diff
@@ -106,8 +106,13 @@ class TileMesh extends THREE.Mesh {
...
 
-    onBeforeRender() {
-        this.material.updateLayersUniforms();
+    /**
+     * An optional callback that is executed immediately before a 3D object is rendered.
+     *
+     * @param {THREE.WebGLRenderer} renderer - The renderer used to render textures.
+     */
+    onBeforeRender(renderer) {
+        this.material.updateLayersUniforms(renderer);
```
gchoqueux: `LayeredMaterial#updateLayersUniforms` is called every frame ?
`LayeredMaterial#layersNeedUpdate` isn't used to avoid calling ?

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,110 @@
...
+        void main() {
+            gl_FragColor = texture2D(sourceTexture, vUv);
+        }
+    `,
+};
+
+let material: THREE.ShaderMaterial | null = null;
+let quad: THREE.Mesh | null = null;
+const quadCam: THREE.OrthographicCamera = new THREE.OrthographicCamera(-1, 1, 1, -1, 0, 1);
+const emptyArrayTexture = new THREE.DataArrayTexture(null, 1, 1, 1);
```
gchoqueux: Could you use [Three post processing](https://github.com/mrdoob/three.js/tree/dev/examples/jsm/postprocessing) with `WebGLArrayRenderTarget`?  

Did you try baking all the textures into a single one? Could that be more efficient?

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,110 @@
...
+        material = new THREE.ShaderMaterial({
+            uniforms: {
+                // This uniform will be updated with each source 2D texture
+                sourceTexture: { value: null },
+            },
+            vertexShader: copyTextureShader.vertexShader,
+            fragmentShader: copyTextureShader.fragmentShader,
+        });
+        quad = new THREE.Mesh(geometry, material);
+    }
```
gchoqueux, line 67: This is similar to the [SavePass](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/postprocessing/SavePass.js) pass. 

This `quad` will be instanced, so you could do it outside of this method

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,110 @@
...
+                arrayTexture.minFilter = texture.minFilter;
+                arrayTexture.wrapS = texture.wrapS;
+                arrayTexture.wrapT = texture.wrapT;
+                arrayTexture.format = texture.format;
+                arrayTexture.type = texture.type;
+                arrayTexture.internalFormat = texture.internalFormat;
+            }
+
+            // render this source texture into the current layer
+            const previousRenderTarget = renderer.getRenderTarget();
```
gchoqueux: Could you move the render target setting outside of the loop?

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // loop through each tile and its textures
+    // to render them into DataArrayTexture layers
+    let currentLayerIndex = 0;
+    for (const tile of tiles) {
+        for (
+            let i = 0;
+            i < tile.textures.length && currentLayerIndex < max;
+            ++i, ++currentLayerIndex
```
gchoqueux, line 77: It could also be interesting in terms of performance and memory footprint to render all the textures of a layer into a single texture.
We would then have a texture array with a size equal to the number of layers.
There would only be one UV system and no more texture offsets.

packages/Main/src/Core/TileMesh.js
Neptilo: Just to be sure: are you just suggesting to add `if (this.material.layersNeedUpdate)` before this line?

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: I don't think I understood the idea. Rendering all textures of a layer into one would produce an extremely large texture that surely couldn't be handled by the GPU...

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: `SavePass` apparently can't be used with `WebGLArrayRenderTarget` because it doesn't give access to the second argument of `setRenderTarget` (the layer index).
Replying to the second question in the comment on the for loop, because I think it's referring to the same idea.

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: Replied to the comment about `SavePass` in the previous comment about Three post processing because I think it refers to the same thing.
I didn't get the second one: where would you rather instanciate it, and why?

packages/Main/src/Core/TileMesh.js
gchoqueux: yes

packages/Main/src/Renderer/WebGLComposer.ts
gchoqueux: If you use only one texture, You don’t need an `WebGLArrayRenderTarget `because all the textures are flatten into one.

packages/Main/src/Renderer/WebGLComposer.ts
gchoqueux: You only need a single instance of the quad, which can be instantiated outside the method.

packages/Main/src/Renderer/WebGLComposer.ts
gchoqueux: I didn’t express myself correctly. 

For each tile, we flat these textures into a single one.

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: Yes there is a single instance of the quad. Where would you instantiate it outside, and what would be the advantage? (Also note that I would rather not instantiate it if `makeDataArrayRenderTarget` is never called.)

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: So you're thinking of a texture atlas. This was one of the original ideas when specifying the issue. Data array textures are a modern way to do the same thing in an optimized way, using a single GPU buffer, like a texture atlas would, and without the added complexity of UV calculations.

packages/Main/src/Core/TileMesh.js
Neptilo: I added the condition.
Surprisingly, that flag was never used. I also had to set the flag to true in `TiledGeometryLayer.update` because tiles were not refreshed after being unrefined (= when their children were hidden).

packages/Main/src/Renderer/WebGLComposer.ts
Neptilo: On this topic, see my other reply on texture atlases.

Desplandis (approved):
Sounds good to me! Let's merge it! =D

PR #2580, opened 2025-07-10
fix: Remove custom fog shader to support standard Three fog

### LINKED ISSUE #2569: Fog does not work on LayeredMaterial
The fog does not apply on the globe nor the plane, i.e. it is broken in the `LayeredMaterial`.

## Context

Adding fog to the scene works well on threejs objects and on objects with the `PointsMaterial` but not on objects with the `LayeredMaterial`.

## Steps to Reproduce

In any example, for instance `3dtiles_loader.html`:

1. Import `Fog` from `three`
2. Add a basic fog: `view.scene.fog = new Fog( 0xcccccc, 2000, 2100 );`

## Expected Behavior
The fog should apply on all objects.

## Actual Behavior
The fog does not apply on the globe / the plane:

https://github.com/user-attachments/assets/608c33ed-7706-4058-a4b0-6187b8a50715

The current fog implementation is broken:

* Fog configuration should be based on the fog configuration of the scene. For instance: [USE_FOG](https://github.com/iTowns/itowns/packages/Main/src/Renderer/LayeredMaterial.ts#L260), [fogDistance and fogColor](https://github.com/iTowns/itowns/blob/049198344db809d3bad306ece81440e62ac23abe/packages/Main/src/Renderer/LayeredMaterial.ts#L299-L300)
* Two shader chunks [fog_fragment.glsl](https://github.com/iTowns/itowns/blob/049198344db809d3bad306ece81440e62ac23abe/packages/Main/src/Renderer/Shader/Chunk/fog_fragment.glsl) and [fog_pars_fragment.glsl](https://github.com/iTowns/itowns/blob/049198344db809d3bad306ece81440e62ac23abe/packages/Main/src/Renderer/Shader/Chunk/fog_pars_fragment.glsl) are imported in [LayeredMaterial shader](https://github.com/iTowns/itowns/blob/049198344db809d3bad306ece81440e62ac23abe/packages/Main/src/Renderer/Shader/TileFS.glsl#L4) but they are not working.

Note that the [PointsMaterial](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Renderer/PointsMaterial.js) has fog implemented and working (see the [shader](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Renderer/Shader/PointsFS.glsl#L7) for its implementation).

Neptilo:
## Description
Remove the custom fog shaders used in `LayeredMaterial` to bring back support for Three's standard method of adding a default fog to the scene.

## Motivation and Context
The fog defined in the scene was not applied on the globe or on the plane, but it did on other objects.
This fixes #2569.
Tested on Firefox, Ubuntu.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,6 +19,7 @@ The following people have contributed to iTowns.
   * [Anthony Gullient](https://github.com/AnthonyGlt)
   * [Jonathan Garnier](https://github.com/jogarnier)
   * [Irénée Dubourg](https://github.com/airnez)
+  * [Victor Ripplinger](https://github.com/neptilo)
 
 * [Oslandia](http://www.oslandia.com)
   * [Vincent Picavet](https://github.com/vpicavet)
--- a/packages/Main/src/Renderer/LayeredMaterial.ts
+++ b/packages/Main/src/Renderer/LayeredMaterial.ts
@@ -148,7 +148,8 @@ interface LayeredMaterialRawUniforms {
     lightPosition: THREE.Vector3;
 
     // Misc
-    fogDistance: number;
+    fogNear: number;
+    fogFar: number;
     fogColor: THREE.Color;
     overlayAlpha: number;
     overlayColor: THREE.Color;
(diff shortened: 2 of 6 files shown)


Desplandis (commented):
Sounds good to me! Have you tested it with [`FogExp2`](https://threejs.org/docs/#api/en/scenes/FogExp2) too?

Neptilo:
@Desplandis Now I have.

Desplandis (approved):

PR #2583, opened 2025-07-16
Revamp examples page

HoloTheDrunk:
Requires #2590 

## Description

Redesign of the examples page:
- CodeMirror 5 editor to edit the examples and run the editor content in the `<iframe>` instead of a direct link to the example blob on GitHub.
- Cleanup of the index HTML and CSS, moving to a flex-based layout.
- More responsive UI to accommodate smaller screens (including mobile by simply hiding the editor and example picker for now).
- Share button to improve not only the user experience of sharing examples with others but also make reporting reproducible bugs significantly easier.

## Motivation and Context

The examples page is one of the first things people will see when researching which library to use. By providing immediately editable examples, we provide a way for prospective users to toy around and see if iTowns matches their needs for a very low time investment on their end. 

## Things to be done after merge
- [ ] Add button redirecting to a gallery with a thumbnail of every example. Will be added in https://github.com/iTowns/itowns.github.io/pull/34
- [ ] Add tip in issue/PR template about using the examples page to reproduce a bug.

## Screenshots
1080p full width
Screenshot from 2025-08-20 10-52-16
Lower resolution / reduced width
Screenshot from 2025-08-20 10-52-53
Mobile ratio (editor shows up in landscape mode)
image

### DIFF
--- a/examples/css/example.css
+++ b/examples/css/example.css
@@ -42,6 +42,13 @@ body {
     }
 }
 
+@media (prefers-reduced-motion: reduce) {
+    * {
+        animation: none !important;
+        transition: none !important;
+    }
+}
+
 /* MAIN SECTION (RIGHT) */
 #main {
     width: calc(100% * 9 / 16);
@@ -64,7 +71,7 @@ body {
     transition-duration: 200ms;
 }
 
-/* NAVIGATION (LEFT) */
+/* EDITOR (LEFT) */
 #editor-panel {
     background-color: light-dark(white, #202030);
     width: calc(100% * 7 / 16);
@@ -83,6 +90,22 @@ body {
 
 #editor-panel>hr {
     margin: 0 1em;
+}
+
+#editor-collapse {
+    display: flex;
+    position: absolute;
+    background-color: #0023;
+    border-width: 0;
+    left: calc(100% * 7 / 16 - 1%);
+    width: 1%;
+    aspect-ratio: 1 / 2;
+    flex-direction: row;
+    flex: 0;
+    align-items: center;
+    justify-content: center;
+    user-select: none;
+    transition-duration: 200ms;
 }
 
 .switch input {
--- a/examples/css/example.css
+++ b/examples/css/example.css
@@ -7,8 +7,13 @@
 }
 
 /* GENERAL */
+* {
+    transition-duration: 200ms;
+}
+
 html,
 body {
+    background-color: black;
     margin: 0;
     padding: 0;
     height: 100%;
@@ -60,6 +65,7 @@ body {
 
 /* NAVIGATION (LEFT) */
 #editor-panel {
+    background-color: white;
     width: calc(100% * 7 / 16);
     height: 100%;
     position: absolute;
@@ -108,20 +114,29 @@ body {
     gap: 1em;
 }
 
+.selector {
+    border: 0;
+    background-color: #404e73;
+    color: white;
+    padding: 1em;
+    border-radius: 0.3em;
+}
+
+#example-selector {
+    width: 30%;
+    flex-grow: 3;
+}
+
+#font-size-selector {
+    width: 4em;
+}
+
 #editor-header> :nth-child(3)>div {
     display: flex;
     flex-flow: row;
     justify-content: space-around;
     align-items: center;
     gap: inherit;
-}
-
-#example-selector {
-    border: 0;
-    background-color: #404e73;
-    color: white;
-    padding: 1em;
-    border-radius: 0.3em;
 }
 
 .header-button {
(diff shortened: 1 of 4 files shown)


examples/index.html
```diff
@@ -1,125 +1,241 @@
...
-        if (document.location.hash) {
-            var url = document.location.href.replace('#', '') + '.html';
-            goTo(url);
-        } else {
-            goTo('view_3d_map.html');
+        /** @param {number} size_px - The new font size to apply */
+        function setEditorFontSize(size_px) {
+            console.info("Setting editor font size");
+            const editor = document.getElementsByClassName("CodeMirror")[0];
+            editor.style.fontSize = `${size_px}px`;
```
Neptilo: The font size input should probably be constrained/sanitized a little.
* Currently, I can type anything. When the input is invalid, there is no indicator of it.
* A font size of 0 is accepted, as well as huge values.
* We probably don't need to show (and use) more than one decimal digit. When I first loaded the page, the value was 15.5821.

examples/images/share-icon.svg
```diff
@@ -0,0 +1,43 @@
...
+     type="text/css"
+     id="style1">
+	.st0{fill:#000000;}
+</style></svg>
```
Neptilo, line 43: Should `</style>` be replaced by `</path>`?

examples/css/example.css
```diff
@@ -88,110 +60,167 @@ body {
...
+    background-size: 3em;
 }
 
-#version-switch {
-    text-align: justify;
-    line-height: 1.6;
+#share-button {
+    background: url("../images/share-icon.svg") center no-repeat;
+    background-size: 3em;
 }
```
Neptilo, line 335: Suggestion: use a pointing hand cursor when hovering the controls, and add tool tips.

examples/css/example.css
```diff
@@ -88,110 +60,167 @@ body {
...
-#view-source {
-    position: fixed;
-    right: 25px;
-    bottom: 10px;
-    width: 30px;
-    height: 30px;
-    padding: 0;
-    box-sizing: border-box;
+/* NAVIGATION (LEFT) */
+#editor-panel {
```
Neptilo, line 97: Not sure where to say this:
Scrolling fast makes the text bounce.

examples/images/share-icon.svg
HoloTheDrunk: No, the `<path>` element is self-closing and the ending followed by the `<style>` tag is actually on the far end of the `d=` line. No clue why Inkscape automatically formats files this way though, seems a bit strange...

examples/index.html
HoloTheDrunk: Kept huge values possible because I don't want to mess with the usually dormant minority of developers using 16k ultra-wide IMAX-size screens to code (actually just for accessibility purposes; big text contained in a box that won't mess with styling of the rest of the page should be the user's choice to resize without being at the whims of what the develop deems reasonable imo).

HoloTheDrunk:
Missing tooltips, dangit. I'll add that in right away.

Neptilo (approved):

jailln:
Hi, thanks for this proposal @HoloTheDrunk , it is way better and more usable ! After discussion with @clameger , we have a few modification propositions:
* We think we should add a gallery, similar to https://docs.mapbox.com/mapbox-gl-js/example/ , with a screenshot, a name and a short description for each example. When clicking on one of the examples in the gallery, it could open the current view and we could add a button in the current view to get back to the gallery. We can provide examples descriptions and take this opportunity to quickly review together the ones we really want to keep. 
* Would it be possible to add a button to set the example view in full screen ? (useful for demos for instance)
* We have some propositions to modify a little bit the interface on the upper left (above the code), I will try to provide a small model of what we have in mind

jailln:
Hi @HoloTheDrunk,

Sorry for the late feedback but I finally took some time to do some more tests. A few propositions:
* Could we change the `Custom` example by something very simple with a GlobeView and a basemap? (I think `GlobeView` is more commonly used than `PlanarView`)
*  The "latest" checkbox does not seem to work
* Could you rename "latest" into "latest version"?

A UI designer from my team will also work on a new UI, including a gallery as mentioned in my previous comment but she will only be able to work on this in February. Therefore I propose that we merge this PR when the small changes mentioned in the list above are done and we work on the UI in another PR.

jailln:
@HoloTheDrunk 

We worked on a mock-up to improve even more the UI of the sandcastle example you provide. Some screenshots:

image

image
 
We should be able to work on that on our side in the comming weeks. Since this PR has been opened a long time I propose that we merge it as it is already a nice improvement and we start this interface work on a new branch. We'll also take care of the few remaining comments. Is that ok for you? I think you just need to rebase and squash the commits before merging.

jailln (approved):

Issue #2540, opened 2025-05-06
Modifiying the visibility/opacity of a pointcloud class is affecting the other classes


AnthonyGlt:
## Context
I'm using the classification view of a cloud point. When I'm editing the value of opacity of one class, other classes are impacted and I can see through other points

## Steps to Reproduce

1. Edit the opacity of a class in the DEFAULT ClassificationScheme of PointsMaterial.js . For example, opacity 0.0 in the "building" class
2. Load Sete point cloud from the 3dtiles_loader.html
3. Switch to classification view
4. Observe that the transparency is applied also tho other points

## Expected Behavior
Just hide the specified class

## Actual Behavior
Other classes are impacted

![Image](https://github.com/user-attachments/assets/13c6e8e3-0727-49dd-9eb0-ba6fe3c53683)

AnthonyGlt:
@ftoromanoff  Could this [MR](https://github.com/iTowns/itowns/pull/2374) fix this issue ? WDYT ?

ftoromanoff:
It fixes the probleme for PointCloudLayer, 3Dtiles works a bit differently I think but if it's based on the same PointMaterial, it should fix the probleme as well.

ftoromanoff:
I added a commit #2374 to report the changes I made on PointCloudLayer to 3DTilesLayer.
1) Set transparency at true by default (https://github.com/iTowns/itowns/pull/2374/commits/cf203c572c52025e94a5a6f5b6914876cc46dba7)
2) add the properties depthWrite (https://github.com/iTowns/itowns/pull/2374/commits/1e34e24e437127126086acd8bc00b05d4ba16949)

Does this help you with the issue, or at least help you find a solution in your case ? (I havn't check performances as well)

AnthonyGlt:
@ftoromanoff With your modifications, it looks like the issue has been solved !
Thank you

AnthonyGlt:
fixed by
https://github.com/iTowns/itowns/pull/2374

jailln:
Reopen: the fix was rolled back in #2584

ftoromanoff:
I confirm that one of my bug is back as well.
Using the class visibility change the color of the point to 'white' and not to transparent:

With the points classified as 'default':
<img width="1266" height="913" alt="Image" src="https://github.com/user-attachments/assets/5baa9890-831b-48d0-8778-6d5a09129378" />

Without them:
[...]

PR #2584, opened 2025-07-17
fix(terrain): fix depth & transparency issues

### LINKED ISSUE #2561: [REGRESSION] We can't change the opacity of the globe

Desplandis:
## Description
Fixes #2561. I however had to revert #2562 and reintroduce some minor render bugs (due to transparency) when hiding points by classification.

### DIFF
--- a/packages/Main/src/Layer/OGC3DTilesLayer.js
+++ b/packages/Main/src/Layer/OGC3DTilesLayer.js
@@ -145,7 +145,16 @@ export function enableMeshoptDecoder(MeshOptDecoder) {
     itownsGLTFLoader.setMeshoptDecoder(MeshOptDecoder);
 }
 
-// TODO: rename
+/**
+ * Patches material properties to automatically update the material (uniforms
+ * and shader) when the layer properties are updated. Note that:
+ * - The transparent property is set according to the opacity. This leads
+ *   to the recompilation of the material if not cached.
+ * - The material properties cannot be set from within the material, so the
+ *   setters are not implemented and shall not be used.
+ * @param {Material} material A three.js material
+ * @param {OGC3DTilesLayer} layer An OGC3DTilesLayer
+ */
 function referMaterialProperties(material, layer) {
     Object.defineProperty(material, 'opacity', {
         get: () => layer.opacity,
@@ -171,6 +180,17 @@ function referMaterialProperties(material, layer) {
     });
 }
 
+/**
+ * Patches material properties to automatically update the material (uniforms
+ * and shader) when the layer properties are updated. Note that:
+ * - The transparent property is set according to the opacity **and** the
+ *   presence of non-opaque pixels in the classification texture. This leads
+ *   to the recompilation of the material if not cached.
+ * - The material properties cannot be set from within the material, so the
+ *   setters are not implemented and shall not be used.
+ * @param {PointsMaterial} material An itowns PointsMaterial
+ * @param {OGC3DTilesLayer} layer An OGC3DTilesLayer
+ */
 function referPointsMaterialProperties(material, layer) {
     let _transparent = material.transparent;
     Object.defineProperty(material, 'transparent', {
--- a/packages/Main/src/Layer/OGC3DTilesLayer.js
+++ b/packages/Main/src/Layer/OGC3DTilesLayer.js
@@ -16,7 +16,6 @@ import iGLTFLoader from 'Parser/iGLTFLoader';
 import { DRACOLoader } from 'three/addons/loaders/DRACOLoader.js';
 // eslint-disable-next-line import/extensions, import/no-unresolved
 import { KTX2Loader } from 'three/addons/loaders/KTX2Loader.js';
-import ReferLayerProperties from 'Layer/ReferencingLayerProperties';
 import PointsMaterial, {
     PNTS_MODE,
     PNTS_SHAPE,
@@ -144,6 +143,88 @@ export function enableMeshoptDecoder(MeshOptDecoder) {
         throw new Error('MeshOptDecoder module is mandatory');
     }
     itownsGLTFLoader.setMeshoptDecoder(MeshOptDecoder);
+}
+
+// TODO: rename
+function referMaterialProperties(material, layer) {
+    Object.defineProperty(material, 'opacity', {
+        get: () => layer.opacity,
+        set: () => {},
+    });
+
+    let _transparent = material.transparent;
+    Object.defineProperty(material, 'transparent', {
+        get: () => {
+            const transparent = material.opacity < 1.0;
+            if (transparent != _transparent) {
+                material.needsUpdate = true;
+                _transparent = transparent;
+            }
+            return transparent;
+        },
+        set: () => {},
+    });
+
+    Object.defineProperty(material, 'wireframe', {
+        get: () => layer.wireframe,
+        set: () => {},
+    });
+}
+
+function referPointsMaterialProperties(material, layer) {
+    let _transparent = material.transparent;
+    Object.defineProperty(material, 'transparent', {
+        get: () => {
+            const transparent = material.opacity < 1.0
+                || material.classificationTexture.userData.transparent;
+            if (transparent != _transparent) {
+                material.needsUpdate = true;
+                _transparent = transparent;
+            }
+            return transparent;
+        },
+        set: () => {},
+    });
+
+    Object.defineProperty(material, 'wireframe', {
+        get: () => layer.wireframe,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.opacity, 'value', {
+        get: () => layer.opacity,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.mode, 'value', {
+        get: () => layer.pntsMode,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.shape, 'value', {
+        get: () => layer.pntsShape,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.sizeMode, 'value', {
+        get: () => layer.pntsSizeMode,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.minAttenuatedSize, 'value', {
+        get: () => layer.pntsMinAttenuatedSize,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.maxAttenuatedSize, 'value', {
+        get: () => layer.pntsMaxAttenuatedSize,
+        set: () => {},
+    });
+
+    Object.defineProperty(material.uniforms.scale, 'value', {
+        get: () => layer.scale,
+        set: () => {},
+    });
 }
 
 async function getMeshFeatures(meshFeatures, options) {
@@ -429,6 +510,10 @@ class OGC3DTilesLayer extends GeometryLayer {
      * @private
      */
     _assignFinalMaterial(model) {
+        if (!model.isMesh && !model.isPoints) {
+            return;
+        }
+
         let material = model.material;
 
         if (model.isPoints) {
@@ -442,12 +527,9 @@ class OGC3DTilesLayer extends GeometryLayer {
             });
             pointsMaterial.copy(material);
             material = pointsMaterial;
-            material.depthWrite = false;
-        }
-
-        if (material) {
-            material.transparent = true;
-            ReferLayerProperties(material, this);
+            referPointsMaterialProperties(material, this);
+        } else {
+            referMaterialProperties(material, this);
         }
 
         model.material = material;
--- a/packages/Main/src/Layer/OGC3DTilesLayer.js
+++ b/packages/Main/src/Layer/OGC3DTilesLayer.js
@@ -145,6 +145,10 @@ export function enableMeshoptDecoder(MeshOptDecoder) {
     itownsGLTFLoader.setMeshoptDecoder(MeshOptDecoder);
 }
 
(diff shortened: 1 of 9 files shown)


AnthonyGlt:
Thanks :pray:

gchoqueux:
revert part of #2374

gchoqueux:
On master, the globe opacity seems to work on globe

https://github.com/user-attachments/assets/5d3564ab-9661-4185-ad09-77e917ad36e9

gchoqueux:
Do you maybe think this wasn't the right way to fix this bug #2561 ?

Desplandis:
I don't think it was the correct way to fix this bug. Unless we aim recompute the render order ourselves, but in both cases we should not hardcode the value of the render order.

Desplandis:
As for the revert on the `depthWrite`, I agree with your previous issue on this topic.

gchoqueux:
Hardcoded values are indeed not a good practice 
Layer transparency have several issues they should be listed and provide a robust solution.
What are the bugs on the master and what are the bugs with this PR @ftoromanoff @AnthonyGlt ?

AnthonyGlt:
With this fix, we can't change the opacity/visibility of a pointcloud's class.
To reproduce it:



-     Edit the opacity or the visible attribute of a class in the DEFAULT ClassificationScheme of PointsMaterial.js . For example, opacity 0.0 in the "building" class
-     Load Sete point cloud from the 3dtiles_loader.html
-     Switch to classification view
-     Observe that the transparency isn't applied


It will be applied when the layer has an opacity < 1, but then the original bug is back : https://github.com/iTowns/itowns/issues/2540

[Screencast from 23-07-2025 09:30:43.webm](https://github.com/user-attachments/assets/bb20b1ee-e452-4443-8598-44d5ccf60b42)

gchoqueux:
To renderer complex transparent meshes, a way is used [Weighted, Blended Order Independent Transparency ](https://stevinz.github.io/three-wboit/WeightedBlended.html)

Desplandis:
Unfortunately I think this is local bug compared to what was introduced on the fixes. Unless we manage to fix the original issue (@ftoromanoff) in reasonable time, I propose that we revert the fixes.

Neptilo:
Would it make sense to also add a non-regression test as part of the PR? This bug had been unnoticed for more than a year, so this would help noticing regressions earlier.
For example, the bug affected the opacity slider of the buildings in `source_stream_wfs_3d.html`, which had no effect. A test could be based on that example.

packages/Main/src/Layer/OGC3DTilesLayer.js
```diff
@@ -146,6 +145,88 @@ export function enableMeshoptDecoder(MeshOptDecoder) {
     itownsGLTFLoader.setMeshoptDecoder(MeshOptDecoder);
 }
 
+// TODO: rename
+function referMaterialProperties(material, layer) {
```
jailln, line 162: Can you add a comment explaining why we do that please ?

packages/Main/src/Layer/OGC3DTilesLayer.js
```diff
@@ -146,6 +145,88 @@ export function enableMeshoptDecoder(MeshOptDecoder) {
...
+    Object.defineProperty(material, 'opacity', {
+        get: () => layer.opacity,
+        set: () => {},
```
jailln: Why do you remove all the setters ? Why not redirecting the setter to the layer property instead ? :
```
  set: (value) => {
    layer.opacity = value;
  },
```

jailln (commented):
Thanks for the fix ! I've tested it and it fixes the terrain opacity and brings back #2540 (which I reopened). I just have a few comments. Also, the coverage test doesn't pass.

packages/Main/src/Layer/OGC3DTilesLayer.js
Desplandis: I kept the same mechanism to "override" the uniforms values by setting their getter to actually reference the layer properties (`ReferLayerProperties`). However I moved it in `OGC3DTilesLayer` because the condition for setting the `transparent` property is different for standard materials and our custom points material. We set the `transparent` property for standards materials when opacity < 1.0 and for points material when opacity < 1.0 **and** if some classification is non-opaque.

jailln: I understood all that, I meant could you add a comment **in the code** for future developers ?  

Sorry I was not specific enough. At least your comment is ready :grin:

packages/Main/src/Layer/OGC3DTilesLayer.js
Desplandis: There is no case where we directly update the material properties (I mean outside the code of the layer) and I don't think we want that an edit to material property have a side-effect to the layer. I would prefer to have some kind of group of uniforms that we could share across materials (there is `UniformGroups` in three but I never used them yet).

packages/Main/src/Layer/OGC3DTilesLayer.js
Desplandis: Yeah, at least the comment is ready x)

packages/Main/src/Layer/OGC3DTilesLayer.js
jailln: Ok, I'm fine with your justification. Could we add a warning for external itowns user though please ?

packages/Main/src/Layer/OGC3DTilesLayer.js
Desplandis: Done

packages/Main/src/Layer/OGC3DTilesLayer.js
Desplandis: Done

Desplandis:
@jailln I will add some tests to OGC3DTilesLayer to pass the coverage.

Desplandis:
@jailln Ready for another review =)

jailln (approved):
Thanks for the tests :)

PR #2586, opened 2025-07-22
Better handle near and far camera values depending on camera position

### LINKED ISSUE #2049: Better handle near and far camera values depending on camera position
The near and far plane [are fixed](https://github.com/iTowns/itowns/blob/master/src/Core/Prefab/GlobeView.js#L102) in the `GlobeView` and the `PlanarView`.

1. Make them configurable
2. Update the far distance dynamically depending on the camera position. The far plane is indeed quite far (10 times the earth diameter for the `GlobeView`) to allow zooming out pretty far from the globe or the plane. This value is not adapted when zooming in and tilting the camera: it generates unuseful tiles subdivision and external layers querying and manipulation, which has an impact on performance. Ideally, this should be configurable in some way too.

Neptilo:
## Description
Optimize the camera's near and far planes based on the camera position and the currently observable part of the view, and make them dynamic. Make some parameters configurable.

## Motivation and Context
The near and far plane were fixed in the GlobeView and the PlanarView. This loads too many unnecessary far-away tiles.
Fixes #2049.
Tested on Firefox, Ubuntu.

### DIFF
--- a/packages/Main/src/Core/Prefab/GlobeView.js
+++ b/packages/Main/src/Core/Prefab/GlobeView.js
@@ -81,9 +81,13 @@ class GlobeView extends View {
      * @param {object} [options] - See options of {@link View}.
      * @param {Object} [options.controls] - See options of {@link GlobeControls}
      * @param {Object} [options.webXR] - WebXR configuration - its presence alone
-     * enable WebXR to switch on VR visualization. (optional).
-     * @param {function} [options.webXR.callback] - WebXR rendering callback (optional).
-     * @param {boolean} [options.webXR.controllers] - Enable the webXR controllers handling (optional).
+     * enable WebXR to switch on VR visualization.
+     * @param {function} [options.webXR.callback] - WebXR rendering callback.
+     * @param {boolean} [options.webXR.controllers] - Enable the webXR controllers handling.
+     * @param {number} [options.farFactor=20] - Controls how far the camera can see.
+     * The maximum view distance is this factor times the camera’s altitude (above sea level).
+     * @param {number} [options.fogSpread=0.5] - Proportion of the visible depth range that contains fog.
+     * Between 0 and 1.
      */
     constructor(viewerDiv, placement = {}, options = {}) {
         THREE.Object3D.DEFAULT_UP.set(0, 0, 1);
@@ -93,7 +97,6 @@ class GlobeView extends View {
 
         this.camera3D.near = Math.max(15.0, 0.000002352 * ellipsoidSizes.x);
         this.camera3D.far = ellipsoidSizes.x * 10;
-
         const tileLayer = new GlobeLayer('globe', options.object3d, options);
         this.mainLoop.gfxEngine.label2dRenderer.infoTileLayer = tileLayer.info;
 
@@ -107,11 +110,45 @@ class GlobeView extends View {
             placement.range = placement.range || ellipsoidSizes.x * 2.0;
         }
 
+        this.farFactor = options.farFactor ?? 20;
+        this.fogSpread = options.fogSpread ?? 0.5;
+
         if (options.noControls) {
             CameraUtils.transformCameraToLookAtTarget(this, this.camera3D, placement);
+
+            // In this case, since the camera's near and far properties aren't
+            // dynamically computed, the default fog won't be adapted, so disable it.
+            this.scene.fog = null;
         } else {
             this.controls = new GlobeControls(this, placement, options.controls);
             this.controls.handleCollision = typeof (options.handleCollision) !== 'undefined' ? options.handleCollision : true;
+
+            const globeRadiusMin = Math.min(ellipsoidSizes.x, ellipsoidSizes.y, ellipsoidSizes.z);
+
+            this.addEventListener(VIEW_EVENTS.CAMERA_MOVED, () => {
+                // update camera's near and far
+                const originToCamSq = this.camera3D.position.lengthSq();
+
+                // maximum possible distance from ground to camera
+                const camCoordinates = new Coordinates(this.referenceCrs)
+                    .setFromVector3(this.camera3D.position);
+                camCoordinates.as(this.tileLayer.extent.crs, camCoordinates);
+                const camToSeaLevel = camCoordinates.z;
+
+                const camToGroundDistMin = camToSeaLevel - View.ALTITUDE_MAX;
+                this.camera3D.near = Math.max(1, camToGroundDistMin * this.fovDepthFactor);
+
+                // distance from camera to the horizon
+                const horizonDist = Math.sqrt(Math.max(0, originToCamSq - globeRadiusMin * globeRadiusMin));
+
+                this.camera3D.far = Math.min(this.farFactor * camToSeaLevel, horizonDist);
+                this.camera3D.updateProjectionMatrix();
+
+                const fog = this.scene.fog;
+                if (!fog) { return; }
+                fog.far = this.camera3D.far;
+                fog.near = fog.far - this.fogSpread * (fog.far - this.camera3D.near);
+            });
         }
 
         this.addLayer(new Atmosphere('atmosphere', options.atmosphere));
--- a/packages/Main/src/Core/Prefab/PlanarView.js
+++ b/packages/Main/src/Core/Prefab/PlanarView.js
@@ -1,6 +1,6 @@
 import * as THREE from 'three';
 
-import View from 'Core/View';
+import View, { VIEW_EVENTS } from 'Core/View';
 import CameraUtils from 'Utils/CameraUtils';
 
 import PlanarControls from 'Controls/PlanarControls';
@@ -25,6 +25,10 @@ class PlanarView extends View {
      * @param {CameraUtils~CameraTransformOptions|Extent} [options.placement] - The
      * {@link CameraUtils~CameraTransformOptions} to apply to view's camera or the extent it must display at
      * initialization. By default, camera will display the view's extent (given in `extent` parameter).
+     * @param {number} [options.farFactor=20] - Controls how far the camera can see.
+     * The maximum view distance is this factor times the camera’s altitude (above sea level).
+     * @param {number} [options.fogSpread=0.5] - Proportion of the visible depth range that contains fog.
+     *  Between 0 and 1.
      */
     constructor(viewerDiv, extent, options = {}) {
         THREE.Object3D.DEFAULT_UP.set(0, 0, 1);
@@ -36,14 +40,57 @@ class PlanarView extends View {
         // Configure camera
         const dim = extent.planarDimensions();
         const max = Math.max(dim.x, dim.y);
-        this.camera3D.near = 0.1;
-        this.camera3D.far = this.camera3D.isOrthographicCamera ? 2000 : 2 * max;
-        this.camera3D.updateProjectionMatrix();
 
         const tileLayer = new PlanarLayer('planar', extent, options.object3d, options);
         this.mainLoop.gfxEngine.label2dRenderer.infoTileLayer = tileLayer.info;
 
         this.addLayer(tileLayer);
+
+        this.addEventListener(VIEW_EVENTS.CAMERA_MOVED, () => {
+            // update camera's near and far
+
+            // compute layer's bounding box
+            if (!this.tileLayer) { return; }
+            const obj = this.tileLayer.object3d;
+            const box = new THREE.Box3();
+            obj.traverse((child) => {
+                if (child.isMesh && child.geometry) {
+                    child.geometry.computeBoundingBox();
+                    const childBox = new THREE.Box3().copy(child.geometry.boundingBox)
+                        .applyMatrix4(child.matrixWorld);
+                    box.union(childBox);
+                }
+            });
+            if (box.isEmpty()) { return; }
+
+            // calculate the greatest distance to the corners of the box
+            let maxDistSq = 0;
+            const camPos = this.camera3D.position;
+            for (const x of [box.min.x, box.max.x]) {
+                for (const y of [box.min.y, box.max.y]) {
+                    for (const z of [box.min.z, box.max.z]) {
+                        const distSq = camPos.distanceToSquared({ x, y, z });
+                        if (distSq > maxDistSq) {
+                            maxDistSq = distSq;
+                        }
+                    }
+                }
+            }
+            const maxDist = Math.sqrt(maxDistSq);
+
+            const camToGroundDistMin = this.camera3D.position.z - View.ALTITUDE_MAX;
+            this.camera3D.near = Math.max(1, camToGroundDistMin * this.fovDepthFactor);
+
+            const boxBottomToCam = camPos.z - box.min.z;
+            const far = this.farFactor * boxBottomToCam;
+            this.camera3D.far = Math.min(far, maxDist);
+            this.camera3D.updateProjectionMatrix();
+
+            const fog = this.scene.fog;
+            if (!fog) { return; }
+            fog.far = far;
+            fog.near = fog.far - this.fogSpread * (fog.far - this.camera3D.near);
+        });
 
         // Configure camera
         const placement = options.placement || {};
@@ -60,10 +107,17 @@ class PlanarView extends View {
         }
 
         this.tileLayer = tileLayer;
+
+        this.farFactor = options.farFactor ?? 20;
+        this.fogSpread = options.fogSpread ?? 0.5;
     }
 
     addLayer(layer) {
-        return super.addLayer(layer, this.tileLayer);
+        const p = super.addLayer(layer, this.tileLayer);
+        p.then(() => {
+            this.dispatchEvent({ type: VIEW_EVENTS.CAMERA_MOVED });
+        });
+        return p;
     }
 }
 
--- a/packages/Main/test/unit/entwine.js
+++ b/packages/Main/test/unit/entwine.js
@@ -118,8 +118,7 @@ describe('Entwine Point Tile', function () {
 
         it('tries to update on the root and succeeds', function (done) {
             view.controls.lookAtCoordinate({
-                coord: source.center,
-                range: 250,
+                range: -250,
             }, false)
                 .then(() => {
                     layer.update(context, layer, layer.root);
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -19,7 +19,7 @@ jobs:
 
       # Use specific Node.js version
       - uses: actions/checkout@v4
-      - name: Use Node.js 22.x
+      - name: Use Node.js 22.17
         uses: actions/setup-node@v4
         with:
           node-version: 22.17
@@ -91,7 +91,7 @@ jobs:
 
       # Use specific Node.js version
       - uses: actions/checkout@v4
-      - name: Use Node.js 22.x
+      - name: Use Node.js 22.17
         uses: actions/setup-node@v4
         with:
           node-version: 22.17
@@ -137,7 +137,7 @@ jobs:
 
       # Use specific Node.js version
       - uses: actions/checkout@v4
-      - name: Use Node.js 22.x
+      - name: Use Node.js 22.17
         uses: actions/setup-node@v4
         with:
           node-version: 22.17
@@ -288,7 +288,7 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - name: Use Node.js 22.x
+      - name: Use Node.js 22.17
         uses: actions/setup-node@v4
         with:
           node-version: 22.17
--- a/packages/Main/src/Controls/PlanarControls.js
+++ b/packages/Main/src/Controls/PlanarControls.js
@@ -1,5 +1,6 @@
 import * as THREE from 'three';
 import { MAIN_LOOP_EVENTS } from 'Core/MainLoop';
+import { VIEW_EVENTS } from 'Core/View';
 
 // event keycode
 export const keys = {
@@ -336,30 +337,31 @@ class PlanarControls extends THREE.EventDispatcher {
             dt = 16;
         }
         const onMovement = this.state !== STATE.NONE;
+        let camChanged = true;
         switch (this.state) {
             case STATE.TRAVEL:
                 this.handleTravel(dt);
-                this.view.notifyChange(this.camera);
                 break;
             case STATE.ORTHO_ZOOM:
                 this.handleZoomOrtho(dt);
-                this.view.notifyChange(this.camera);
                 break;
             case STATE.DRAG:
                 this.handleDragMovement();
-                this.view.notifyChange(this.camera);
                 break;
             case STATE.ROTATE:
                 this.handleRotation();
-                this.view.notifyChange(this.camera);
                 break;
             case STATE.PAN:
                 this.handlePanMovement();
-                this.view.notifyChange(this.camera);
                 break;
             case STATE.NONE:
             default:
+                camChanged = false;
                 break;
+        }
+        if (camChanged) {
+            this.view.dispatchEvent({ type: VIEW_EVENTS.CAMERA_MOVED });
+            this.view.notifyChange(this.camera);
         }
         // We test if camera collides to the geometry layer or is too close to the ground, and adjust its altitude in
         // case
--- a/packages/Main/src/Core/View.js
+++ b/packages/Main/src/Core/View.js
@@ -133,6 +133,9 @@ class View extends THREE.EventDispatcher {
     #layers = [];
     #pixelDepthBuffer = new Uint8Array(4);
     #fullSizeDepthBuffer;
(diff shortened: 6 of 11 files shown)


packages/Main/test/unit/entwine.js
```diff
@@ -118,8 +118,7 @@ describe('Entwine Point Tile', function () {
 
         it('tries to update on the root and succeeds', function (done) {
             view.controls.lookAtCoordinate({
-                coord: source.center,
```
Neptilo, line 121: `source.center` was undefined, so I removed it.

packages/Main/test/unit/entwine.js
```diff
@@ -118,8 +118,7 @@ describe('Entwine Point Tile', function () {
...
-                range: 250,
+                range: -250,
```
Neptilo, line 121: I don´t understand why and how, but reversing the range fixes the tests.

Neptilo:
With the fog now always applied by default, the "realistic lighting" object is now quite ugly and may obscure the view. Considering that the atmosphere will soon be redesigned using Three-geospatial, should I:
* disable it in all examples
* or tweak the code so that the fog isn't applied to the realistic lighting object?

examples/view_multiglobe.html
```diff
@@ -110,9 +110,14 @@
...
                     front = 1 - front;
                     t = 0;
+
                     // schedule the animation
                     view.addFrameRequester(itowns.MAIN_LOOP_EVENTS.BEFORE_RENDER, animator);
                     view.notifyChange(cameraThree);
+
+                    view.camera3D.near = 15;
+                    view.camera3D.far = 63781370;
+                    view.camera3D.updateProjectionMatrix();
```
Neptilo: This fixes a functional test.
The near/far dynamic calculation assumes that the globe is centered at the origin. If the user wants to hack that position, maybe they are supposed to hack (i.e. overwrite) the near/far as well?

packages/Main/test/unit/entwine.js
AnthonyGlt: Could you test it visually by writing a html example corresponding to this scenario  ? COuld be easier to see what's happening

examples/view_multiglobe.html
AnthonyGlt: In this case, the noControl param is set at true. Maybe we should check this value before adding the event listener

https://github.com/iTowns/itowns/blob/67ae696e4d593e64da64287367b481f85b402748/packages/Main/src/Core/Prefab/GlobeView.js#L113

It will keep the default values of near/far. If the user want to change them, he can do it in controls handlers

packages/Main/test/unit/entwine.js
Neptilo: Well you're going to have to help me because there are many test-specific things (like "withoutAutority", stubbed functions) that I don't know where they come from and what to replace them with.

AnthonyGlt: This is mock up data, no need

here is a simplified html

BUT it doesn't help that much because the range 250 is very low

By testing, I see that every negative range is working. 

Looks like it's the far value is the issue here.

```html


<html>
<head>
    <title>EntwinePointTileLayer Test - Update Root Node</title>

    <meta charset="UTF-8">

    <link rel="stylesheet" type="text/css" href="css/example.css">
    <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">

    <style type="text/css">
        #description {
            z-index: 2;
            left: 10px;
        }
    </style>
</head>
<body>
<div id="description">
    <h3>EntwinePointTileLayer Test - Update Root Node</h3>
    <p>This test replicates the unit test: "tries to update on the root and succeeds"</p>
    <div id="testStatus">Status: <span id="statusText">Initializing...</span></div>
    <button id="runTest" onclick="runTestCase()" disabled>Run Test</button>
</div>

<div id="viewerDiv"></div>

<!-- Import iTowns source code -->
<script src="../dist/debug.js"></script>
<!-- Import iTowns LoadingScreen plugin -->
<script src="js/GUI/LoadingScreen.js"></script>

    var statusText = document.getElementById('statusText');
    var runTestButton = document.getElementById('runTest');

    // Test variables (matching the unit test setup)
    var view, layer, context;

    function updateStatus(message) {
        statusText.textContent = message;
        console.log('[Test Status]', message);
    }

    // Define camera initial position (similar to test placement)
    const placement = {
        coord: new itowns.Coordinates('EPSG:4326', 0, 0),
        range: 250
    };

    // Create a GlobeView
    var view = new itowns.GlobeView(viewerDiv, placement);

    // Setup loading screen

    // // Add base layers (ortho imagery and elevation)
    // itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(config) {
    //     config.source = new itowns.WMTSSource(config.source);
    //     var orthoLayer = new itowns.ColorLayer('Ortho', config);
    //     view.addLayer(orthoLayer);
    // });
    //
    // function addElevationLayerFromConfig(config) {
    //     config.source = new itowns.WMTSSource(config.source);
    //     var elevationLayer = new itowns.ElevationLayer(config.id, config);
    //     view.addLayer(elevationLayer);
    // }
    // itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
    // itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);

    // Initialize the EntwinePointTileLayer (matching the test setup)
    function initializeTest() {
        updateStatus('Creating EntwinePointTileSource...');

        // Use the same URL as in the unit test
        var source = new itowns.EntwinePointTileSource({
            url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/entwine'
        });

        updateStatus('Creating EntwinePointTileLayer...');

        // Create layer with the same configuration as in the test
        layer = new itowns.EntwinePointTileLayer('test', { source }, view);

        updateStatus('Setting up context...');

        // Create context exactly as in the unit test
        context = {
            camera: view.camera,
            engine: view.mainLoop.gfxEngine,
            scheduler: view.mainLoop.scheduler,
            geometryLayer: layer,
            view: view,
        };

        updateStatus('Adding layer to view...');

        // Add layer to view using the same method as in the test
        itowns.View.prototype.addLayer.call(view, layer)
            .then(() => {
                updateStatus('Layer ready! Test can be executed.');
                runTestButton.disabled = false;
            })
            .catch((error) => {
                updateStatus('Error adding layer: ' + error.message);
                console.error('Layer initialization error:', error);
            });
    }

    // The actual test case: "tries to update on the root and succeeds"
    function runTestCase() {
        updateStatus('Running test: tries to update on the root and succeeds');

        runTestButton.disabled = true;

        // Execute the exact test logic from the unit test
        view.controls.lookAtCoordinate({
            range: 250,
        }, false)
            .then(() => {
                updateStatus('lookAtCoordinate completed, updating layer...');

                // Call layer.update as in the test
                layer.update(context, layer, layer.root);

                if (layer.root.promise) {
                    updateStatus('Root promise found, waiting for resolution...');

                    // Wait for the promise as in the test
                    layer.root.promise
                        .then(() => {
                            updateStatus('✅ Test PASSED: Root update succeeded!');
                            console.log('✅ Test completed successfully');

                            // Notify view to render changes
                            view.notifyChange();

                            runTestButton.disabled = false;
                        })
                        .catch((error) => {
                            updateStatus('❌ Test FAILED: Root promise rejected - ' + error.message);
                            console.error('❌ Root promise error:', error);
                            runTestButton.disabled = false;
                        });
                } else {
                    updateStatus('⚠️ Test INCONCLUSIVE: Root promise is undefined');
                    console.warn('⚠️ layer.root.promise is undefined');
                    runTestButton.disabled = false;
                }
            })
            .catch((error) => {
                updateStatus('❌ Test FAILED: lookAtCoordinate error - ' + error.message);
                console.error('❌ lookAtCoordinate error:', error);
                runTestButton.disabled = false;
            });
    }

    // Initialize test setup once the page is loaded
    window.addEventListener('load', () => {
        setTimeout(() => {
            updateStatus('Initializing test setup...');
            initializeTest();
        }, 1000);
    });
</script>
</body>
</html>


```

packages/Main/src/Core/Prefab/PlanarView.js
```diff
@@ -36,15 +40,58 @@ class PlanarView extends View {
...
+                            maxDistSq = distSq;
+                        }
+                    }
+                }
+            }
+            const maxDist = Math.sqrt(maxDistSq);
+
+            const closestPoint = new THREE.Vector3();
+            box.clampPoint(camPos, closestPoint);
+            this.camera3D.near = Math.max(1, closestPoint.distanceToSquared(camPos) / maxDist);
```
mgermerie: I don't get why you need to divide the squared minimum distance between camera and layer's bbox by the max distance. Why not simply put the following ?
```js
this.camera3D.near = Math.max(1, Math.sqrt(closestPoint.distanceToSquared(camPos)));
```

packages/Main/src/Core/Prefab/GlobeView.js
```diff
@@ -107,11 +112,50 @@ class GlobeView extends View {
...
+            const fovDepthFactor = 1 / Math.sqrt(1 + corner.x * corner.x + corner.y * corner.y);
+
+            this.addEventListener(VIEW_EVENTS.CAMERA_MOVED, () => {
+                // update camera's near and far
+                const originToCamSq = this.camera3D.position.lengthSq();
+
+                // get the minimum possible elevation (origin to ground), i.e. sea elevation, under camera
+                const camCoordinates = new Coordinates(this.referenceCrs)
+                    .setFromVector3(this.camera3D.position);
+                const layer = this.getLayers(l => l.isTiledGeometryLayer)[0];
```
mgermerie: Why not using `tileLayer` constant defined at line 102 (after your modifications), the same way you did in `PlanarView` ? You could also use `tileLayer` property in both `GlobeView` and `PlanarView`.

packages/Main/src/Core/Prefab/GlobeView.js
```diff
@@ -107,11 +112,50 @@ class GlobeView extends View {
...
+                camCoordinates.as(layer.extent.crs, camCoordinates);
+                camCoordinates.z = 0;
+                camCoordinates.as(this.referenceCrs, camCoordinates);
+                const seaElevationUnderCam = camCoordinates.toVector3().length();
+
+                // maximum possible distance from ground to camera
+                const camToSeaLevel = Math.sqrt(originToCamSq) - seaElevationUnderCam;
```
mgermerie: You could compute this distance by simply converting camera position in `EPSG:4326` and reading its `z` coordinate : 
```js
camCoordinates.as(tileLayer.extent.crs, camCoordinates);
const camToSeaLevel = camCoordinates.z;
```

packages/Main/src/Core/Prefab/GlobeView.js
mgermerie, line 1: The math behind near and far calculation in `GlobeView` theoretically good but not very robust. It involves user only displaying elevation data between `0` and `8849`.

Yet, should someone display a geoid under altitude data, elevation could overflow these limits. There are also negative altitudes in some places. Moreover, these limits imply that an `ElevationLayer` only displays altitude data. However, one should be able to display ellipsoid height, or actually any data within an `ElevationLayer`.

Finally, `ElevationLayer`s come with a `scale` property. Should it be set to something greater than 1, the whole terrain would not fit within `[0, 8849]` boundaries.

packages/Main/src/Core/Prefab/GlobeView.js
```diff
@@ -107,11 +112,50 @@ class GlobeView extends View {
...
         if (options.noControls) {
             CameraUtils.transformCameraToLookAtTarget(this, this.camera3D, placement);
         } else {
             this.controls = new GlobeControls(this, placement, options.controls);
             this.controls.handleCollision = typeof (options.handleCollision) !== 'undefined' ? options.handleCollision : true;
+
+            const globeRadiusMin = Math.min(ellipsoidSizes.x, ellipsoidSizes.y, ellipsoidSizes.z);
+            const corner = new THREE.Vector4(1, 1, -1); // a corner of the camera in NDC at near plane
+            corner.applyMatrix4(this.camera3D.projectionMatrixInverse).divideScalar(corner.z);
+            const fovDepthFactor = 1 / Math.sqrt(1 + corner.x * corner.x + corner.y * corner.y);
```
mgermerie: I get the code you have there but I admit I don't understand why. What does the field of view depth factor physically represents ?

examples/view_multiglobe.html
mgermerie, line 1: Could you remove this change, as it is not in the scope of the PR ?

packages/Main/src/Core/Prefab/PlanarView.js
Neptilo: The problem is this: setting the near plane to the nearest distance from an object to the camera only works well for the center of the camera. Elsewhere, **the near plane is further away than the `near` value**, so some parts of the object may be cropped out.
The rest will be easier to explain with a drawing, that I should add to the doc somewhere. :)

examples/view_multiglobe.html
Neptilo: Ok, I will remove it.
I'm wondering though: minor stylistic changes like this, or like typo corrections for example, will likely never be in the scope of any PR, unless the code surrounding it is also modified, so what should we do in those cases?

packages/Main/src/Core/Prefab/GlobeView.js
Neptilo: And I guess there is no way to get the maximum elevation of a data source directly, so maybe the next best thing we can do is to define the near value as the maximum between a constant value (when we're close to the ground) and the nearest point based on a bounding box (when we're really far)...
Any other idea?

packages/Main/src/Core/Prefab/GlobeView.js
Neptilo: This is to solve the same problem I described in `PlanarView.js`.

Given a plane orthogonal the camera direction (in our case, the near plane), `fovDepthFactor` represents the ratio between the distance a point on this plane at the center of the view and a point on this plane at a corner.

packages/Main/src/Core/Prefab/GlobeView.js
Neptilo: OK, I didn't notice. I will try that. Thanks!

packages/Main/src/Core/Prefab/GlobeView.js
Neptilo: I will try that. Thanks!

packages/Main/src/Core/Prefab/PlanarView.js
Neptilo: Image
Let C be the position of the camera, G its vertical projection on the ground and F the farthest point on the map. In order to be able to see the map in all cases, we decide that the near plane doesn't go beyond G. Here we represent the worst-case scenario, where the camera looks at F, so G is on the near plane. Let N be the point at the center of the camera view and on the near plane.

To calculate the near distance CN, note that triangles CNG and CGF are _similar_.
So CN / CG = CG / CF, i.e. CN = CG² / CF.

packages/Main/src/Core/Prefab/GlobeView.js
mgermerie: Ok, I get the geometry behind. Thanks!
Perhaps you could add a comment on these lines for later reference?

packages/Main/src/Core/Prefab/PlanarView.js
mgermerie: I get what you mean, and agree that the distance to the closest point should be multiplied by a factor smaller than `1` (since the near value can be smaller than the distance to the near plane). Yet, why not compute this factor the same way you did in `GlobeView` (which is an actual projection of the near value on the camera `z` axis)?

packages/Main/src/Core/Prefab/GlobeView.js
mgermerie: I see the issue with the fact we have no elevation data directly, and computing `near` value from tiles bounding box would be much more expensive than in the `PlanarView` situation.

Perhaps we can keep the value computed for a constant hypothetical max altitude (as you did), but add a parameter so user can override this default value with any chosen value? However, this solution would come with downsides. First, the `near` and `far` values should be updated when any `ElevationLayer.scale` property is changed. Second, it would cause issue in the case someone would try to display for instance a 3d model that goes beyond 8849 meters of altitude (perhaps unlikely but still...).

I would say the next solution you suggest is probably more robust but since bounding boxes are cubical and the globe is (almost) spherical, we would need to be very far for the bounding box for the solution to work (given the worst possible camera position).

Given you already changed the computing of the `far` value, would it be much worse in terms of performance to keep a constant near value (but still provide user with the possibility to change it)?

packages/Main/src/Core/Prefab/GlobeView.js
```diff
@@ -93,7 +99,6 @@ class GlobeView extends View {
 
         this.camera3D.near = Math.max(15.0, 0.000002352 * ellipsoidSizes.x);
         this.camera3D.far = ellipsoidSizes.x * 10;
```
mgermerie, line 99: Given the evolution of the PR, don't forget to remove these lines 😉

packages/Main/src/Core/Prefab/PlanarView.js
mgermerie: Also, in the case the view uses an orthographic camera, this factor is not needed. It probably does not induce a big difference though, and the value after multiplication by the factor should always be smaller than the initial value so no visible information is lost with this factor.

I don't know which would be more relevant between having an exception if camera is orthographic or applying a geometrically wrong but still functioning method to orthographic camera. What would be your opinion on this?

examples/view_multiglobe.html
mgermerie: I think that everything that already follows the formal stylistic rules that are properly stated somewhere (either in [CODING.md](https://github.com/iTowns/itowns/blob/master/CODING.md) or in linter rules) can be updated in any PR - with a specific commit though. Any other "personal preference" related style changes can sometimes come in code parts you update for some other reason. However this should be done without polluting the original scope of your modification (let's say you have a commit that changes one variable name. Changing the style of a ten lines block around this variable would be a bit too much for example).

I also agree these rules should be updated and improved, but any change should be first discussed (in an issue) and then formalized and applied everywhere.

mgermerie:
> With the fog now always applied by default, the "realistic lighting" object is now quite ugly and may obscure the view. Considering that the atmosphere will soon be redesigned using Three-geospatial, should I:
> 
>     * disable it in all examples
> 
>     * or tweak the code so that the fog isn't applied to the realistic lighting object?

My personal opinion would be to remove it until `Three-geospatial` is added. However I'd prefer some others opinion on this. WDYT @iTowns/reviewers ?

jailln:
> > With the fog now always applied by default, the "realistic lighting" object is now quite ugly and may obscure the view. Considering that the atmosphere will soon be redesigned using Three-geospatial, should I:
> > ```
> > * disable it in all examples
> > 
> > * or tweak the code so that the fog isn't applied to the realistic lighting object?
> > ```
> 
> My personal opinion would be to remove it until `Three-geospatial` is added. However I'd prefer some others opinion on this. WDYT @iTowns/reviewers ?

I'm all for removing it too, especially since the PR to add three-geospatial is already ready (#2589)

packages/Main/src/Core/Prefab/PlanarView.js
Neptilo: I think the reasons for not computing this factor the same way as in `GlobeView` was:
* to take advantage of a simpler calculation - but since `fovDepthFactor` is computed only once before the callback, that argument doesn't hold.
* because it just didn't work when I was testing it, but I just figured out it was because my bounding box calculation was missing something: it was only based on the objects' geometries, but didn't take into account the geometry modifications occurring in a shader.
I just fixed that in the upcoming commit, and used `fovDepthFactor` again.

Regarding orthographic cameras: I must have thought about it and then forgot to handle that case. I think the most correct way is to add a condition to the calculation and use of `fovDepthFactor`.

packages/Main/src/Core/Prefab/GlobeView.js
Neptilo: The performance related to keeping a constant near value is not an issue.
The reason for computing a sensible near value is that the fog uses it for its interpolation. If it is a constant value, i.e. relatively near the camera when it is far away, the entire globe will be foggy, which we don't want.

Neptilo:
> > > With the fog now always applied by default, the "realistic lighting" object is now quite ugly and may obscure the view. Considering that the atmosphere will soon be redesigned using Three-geospatial, should I:
> > > ```
> > > * disable it in all examples
> > > 
> > > * or tweak the code so that the fog isn't applied to the realistic lighting object?
> > > ```
> > 
> > 
> > My personal opinion would be to remove it until `Three-geospatial` is added. However I'd prefer some others opinion on this. WDYT @iTowns/reviewers ?
> 
> I'm all for removing it too, especially since the PR to add three-geospatial is already ready (#2589)

I actually removed it in the three-geospatial branch. :)

packages/Main/src/Core/Prefab/GlobeView.js
```diff
@@ -84,6 +86,10 @@ class GlobeView extends View {
      * enable WebXR to switch on VR visualization. (optional).
      * @param {function} [options.webXR.callback] - WebXR rendering callback (optional).
      * @param {boolean} [options.webXR.controllers] - Enable the webXR controllers handling (optional).
+     * @param {number} [options.farFactor=20] - Controls how far the camera can see. (optional)
+     * The maximum view distance is this factor times the camera’s altitude (above sea level).
+     * @param {number} [options.fogSpread=0.5] - Proportion of the visible depth range that contains fog.
+     * Between 0 and 1. (optional)
```
mgermerie: Can you remove the « (optional) » mentions (here and in `PlanarView`as well) ? I know they were added on other parameters in the documentation but that was a mistake. The documentation template already recognises optional parameters (thanks to `[...]` synthax) and adds an « (optional) » mention in the built doc. Adding it here just adds it a second time in the doc which is redundant.

Image

packages/Main/src/Core/Prefab/GlobeView.js
mgermerie: Ok I see - thanks for the explanations.

In that case perhaps we could keep the near computing you implemented, but set the `ALTITUDE_MAX` constant to a higher value. The idea would be to set it to an altitude at which the whole globe is visible. That should still hopefully fix the fog issue, and setting it much higher than the altitude of mount Everest would be less likely to cause issue with what user display as elevation data.

mgermerie (commented):
Thanks for the updates on the code. From what I tested, your modifications work almost perfectly well with a perspective camera both on a `GlobeView` and on a `PlanarView`. Well done! 😄 

There are however still few fixes to be made before merging: 

1. The orthographic camera is kinda broken with your changes. I tested [this example](https://www.itowns-project.org/itowns/examples/view_2d_map.html) and [this one](https://www.itowns-project.org/itowns/examples/misc_orthographic_camera.html) on your branch and they failed to work as intended. In the first one, the whole view is hidden after zooming out a bit. In the second one, the view is directly hidden when loading the page. On the [basic globe example](https://www.itowns-project.org/itowns/examples/view_3d_map.html), the minimap widget doesn't work either, and it also uses an orthographic camera. Could you thus add a specific case in `PlanarView` that manages orthographic camera near and far computing ?
2. The [multiple globes example](https://www.itowns-project.org/itowns/examples/view_multiglobe.html) does not work on your branch but does on master. Both `GlobeViews` in the example have a `noControl` option set to true. If I set this option on the [basic globe example](https://www.itowns-project.org/itowns/examples/view_3d_map.html) view, the same result occurs. This is probably due to the fact fog near and far are not computed if the view doesn't have controls set. This should be fixed.

Finally, can you rebase your branch on itowns master branch ? This will be necessary before merging. You will also need to clean your commit tree by squashing most of them. You could for instance keep one commit that brings the dynamic near and far computing and another one that brings the fog feature for instance. - Note that I'm only referring to commits that deal with near/far and fog. The other unrelated ones like the one on integration can stay that way.
Commit messages are used to generate changelog on each release, so it is important to keep the commit tree on master branch as clean as possible.

Neptilo:
Everything should be fixed now.
The broken cases in planar views and the broken tests were due to the recent change of the BB computation algorithm, based on the current tile extents instead of the Three object BBs. So I mostly reverted it and adapted the older algorithm to make it similar to the globe view's: adding a constant `MAX_ALTITUDE` for the comutation of the `near` value.
For the minimap, I added an event dispatch in `PlanarView.addLayer`.
For the multiple globes, I removed the fog if controls were disabled.

It's not my first PR so I was already familiar with our clean Git history policy. ;)
Actually, I was going to squash everything together, but I extracted the integration commit as you suggested.

mgermerie (approved):
I tested it again and with your changes, it all works well. Thanks again for the feature!

PR #2589, opened 2025-08-12
Three-geospatial integration

Neptilo:
## Description
Integration of the [three-geopsatial](https://github.com/takram-design-engineering/three-geospatial/) library, providing realistic rendering of the sky and atmosphere, into the `GlobeView`.
The sky includes the Sun (and the Moon) placed according to a configurable date/time, and the Earth is lit from the Sun's direction.
I also removed the previous atmosphere rendering.

## Motivation and Context
The previous implementation of the atmosphere was not visually satisfactory.
Having a lighting based on a configurable time will also allow simulating accurate shadows in the future.
Tested on Firefox, Ubuntu.

## Screenshots
globe with moon (The grey dot is the Moon.)
sunset landscape
France with atmosphere halo


packages/Main/src/Core/Prefab/Globe/SkyManager.ts
```diff
@@ -0,0 +1,123 @@
...
+        this.sunLight = new SunDirectionalLight({ distance: 300 });
+        this.sunLight.intensity = 0.1;
+        this.sunLight.target.position.copy(camera.position);
+
+        scene.add(this.sunLight);
+        scene.add(this.sunLight.target); // to update matrixWorld at each frame
+
+        this.aerialPerspective = new AerialPerspectiveEffect(camera, {
+            transmittance: false,
+            inscatter: false,
```
Neptilo: These options are what should have replaced the fog, but the shaders don´t seem to use the camera's parameters correctly, and the shaders are complex to debug, so I disabled them for now.

packages/Main/src/Renderer/OrientedImageMaterial.js
```diff
@@ -15,6 +15,7 @@ noMask.needsUpdate = true;
 const noTexture = new THREE.Texture();
 
 const shaderMaterial = new THREE.ShaderMaterial();
+shaderMaterial.name = 'OrientedImage';
```
Neptilo: I wasn´t sure whether to keep this: I named all our materials.
Naming Three.js objects is usually useful, since it helps debugging and doesn´t have any effect on user experience.
Do you have an opinion on that?

packages/Main/src/Renderer/Shader/TileVS.glsl
```diff
@@ -28,6 +31,6 @@ void main() {
         #else
         vUv = vec3(uv, 0.0);
         #endif
-        vNormal = normalize ( mat3( modelMatrix[0].xyz, modelMatrix[1].xyz, modelMatrix[2].xyz ) * normal );
+        vNormal = normalize ( mat3( normalMatrix[0].xyz, normalMatrix[1].xyz, normalMatrix[2].xyz ) * normal );
```
Neptilo, line 34: Fixes a 7-year-old mistake!

Desplandis:
@HoloTheDrunk

HoloTheDrunk (changes requested):
Would it be possible for you to get rid of the commits merged from different PRs and rebase on master instead? It makes the review quite noisy and might cause unexpected issues when merging.

Neptilo:
> Would it be possible for you to get rid of the commits merged from different PRs and rebase on master instead? It makes the review quite noisy and might cause unexpected issues when merging.

I just rebased what I could from master, but this branch directly depends on `dynamic-near-far`, so we cannot get rid of the commits from that branch. I suggest waiting for the related PR before reviewing this one.

Neptilo:
Some updates and ideas:
* I enabled and fixed the aerial perspective effect, which has a similar use to the fog, but more physically correct and realistic.
* When zooming out on the globe, it has a quite strong white overlay, because of the inscatter parameter of the aerial perspective. The same thing happens in three-geospatial's official examples. Unfortunately, the inscatter parameter is only a boolean and its strength cannot be adapted.
* Like the old _realistic rendering_, I suggest making the use of three-geospatial optional. Disabling it would allow seeing the globe lit in all directions without a night-side. In that case, the classical fog would be used. That new option could be named "realistic rendering" just like the former one, or something similar. What do you think?
* In the multiglobe example, the new integration only renders one _atmosphere_, which doesn't move. Should we try to duplicate it and move along with the globes, or deactivate it?

jailln:
Could you also updates the screenshots in the README please so it reflects the new rendering ? :)

We should also update it on the website I think @mgermerie

Neptilo:
Yes, good call. I will do it once I'm sure we're ok with my 2nd bullet point, and once we've settled for a set of parameter values (such as light intensities).

packages/Main/src/Renderer/OrientedImageMaterial.js
HoloTheDrunk: We could definitely make use of this yes, but keeping it behind the `__DEBUG__` global flag so it can be stripped on release builds would probably be better.

HoloTheDrunk:
> * When zooming out on the globe, it has a quite strong white overlay, because of the inscatter parameter of the aerial perspective. The same thing happens in three-geospatial's official examples. Unfortunately, the inscatter parameter is only a boolean and its strength cannot be adapted.

Couldn't we adjust the opacity / blending of the atmosphere effect with a uniform based on camera distance on our side? I have not looked through all the changes just yet so this might not be possible.

> * Like the old _realistic rendering_, I suggest making the use of three-geospatial optional. Disabling it would allow seeing the globe lit in all directions without a night-side. In that case, the classical fog would be used. That new option could be named "realistic rendering" just like the former one, or something similar. What do you think?

I agree with this, a "flat" view should always be available given the goals of iTowns.
 
> * In the multiglobe example, the new integration only renders one _atmosphere_, which doesn't move. Should we try to duplicate it and move along with the globes, or deactivate it?

Duplicating it and moving along with the globes would be ideal. Multiglobe capability is an advertised feature of iTowns and unless it's unreasonably difficult to make it work or requires significant hacks, we should try to avoid breaking user expectations.

Neptilo:
> Couldn't we adjust the opacity / blending of the atmosphere effect with a uniform based on camera distance on our side? I have not looked through all the changes just yet so this might not be possible.

It looks like `AerialPerspectiveEffect.blendMode.opacity` could do the trick.
I'll address this and the other tasks as soon as I can.

Neptilo:
I realized that realistic rendering was not activated in the multiglobe example, so the best way to produce a result that matches the original more closely is to leave it disabled.

Neptilo:
> Could you also updates the screenshots in the README please so it reflects the new rendering ? :)

Do you like this image for example?
image

Is the second mosaic image supposed to be created manually with an image editor?

jailln:
Yes, this one is good, regarding the mosaic image, i don't know but I think so yes. We could re-do it with current examples, with three-geospatial integration. The three screenshots in the description of your PR seems good, and we could also add a screenshot of the 3dtiles example (3dtiles_loader, zoomed on the Lille dataset), maybe one from the copc_simple_loader example, one from the vector_tile_3d_mesh example and one from the view_immersive example, WDYT ? This can also be done in another PR if you prefer ^^

packages/Main/test/unit/entwine.js
```diff
@@ -112,8 +112,12 @@ describe('Entwine Point Tile', function () {
         });
 
         it('tries to update on the root and fails', function () {
+            const cam = context.camera.camera3D;
+            const originalQuaternion = cam.quaternion.clone();
+            cam.quaternion.identity(); // look away from dataset
             layer.update(context, layer, layer.root);
             assert.strictEqual(layer.root.promise, undefined);
+            cam.quaternion.copy(originalQuaternion);
```
Neptilo, line 121: The test was (incorrectly) relying on the fact that the camera's world matrix was not updated when calling `Camera.projectBox3PointsInCameraSpace`, thus actually testing a situation where the camera's matrix would be the identity, and so where the camera would not be looking at the dataset.

This became a problem when some functions did update the camera's world matrix (as they should be allowed). In our case, `EffectComposer.render` did that.

Therefore I added the missing matrix update in `Camera.projectBox3PointsInCameraSpace` and for the test I recreated a situation where the camera looks away from the dataset.

Neptilo:
Investigations findings of the failing functional tests so far:
A long time is spent at the start of some test suites (the first test of a suite takes more than twice its normal time). Most of this time disappears if we remove the "Generate precomputed textures" part of `SkyManager.ts`, however when we time that part, it only takes 20 ms. So it must have a consequence afterwards.

Neptilo:
New findings:
* The problem of lengthy tests only occurs in headless mode. The readPixels calls stall the GPU. Probably related to https://github.com/puppeteer/puppeteer/issues/3938.
* Adding the Puppeteer flag `--disable-gl-drawing-for-tests` helps a lot.
* A few tests actually need to call the real `readPixels`, so the flag cannot be used globally.

Neptilo:
> Yes, this one is good, regarding the mosaic image, i don't know but I think so yes. We could re-do it with current examples, with three-geospatial integration. The three screenshots in the description of your PR seems good, and we could also add a screenshot of the 3dtiles example (3dtiles_loader, zoomed on the Lille dataset), maybe one from the copc_simple_loader example, one from the vector_tile_3d_mesh example and one from the view_immersive example, WDYT ? This can also be done in another PR if you prefer ^^

I was about to update the screenshots when I realized they had become a broken link in the README from master. ^^ I heard about someone writing a script to generate them automatically so I guess it has to do with that. So, if they are indeed automatic now, I suppose I have nothing more to do...

jailln:
> > Yes, this one is good, regarding the mosaic image, i don't know but I think so yes. We could re-do it with current examples, with three-geospatial integration. The three screenshots in the description of your PR seems good, and we could also add a screenshot of the 3dtiles example (3dtiles_loader, zoomed on the Lille dataset), maybe one from the copc_simple_loader example, one from the vector_tile_3d_mesh example and one from the view_immersive example, WDYT ? This can also be done in another PR if you prefer ^^
> 
> I was about to update the screenshots when I realized they had become a broken link in the README from master. ^^ I heard about someone writing a script to generate them automatically so I guess it has to do with that. So, if they are indeed automatic now, I suppose I have nothing more to do...

@mgermerie I think you were working on that ? Is it related ?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -301,7 +301,12 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
     public override defines: LayeredMaterialDefines;
 
     constructor(options: LayeredMaterialParameters = {}, crsCount: number) {
-        super(options);
+        const lambertUniforms = THREE.UniformsUtils.clone(THREE.ShaderLib.lambert.uniforms);
+        const shaderMaterialParams: THREE.ShaderMaterialParameters = { ...options };
+        shaderMaterialParams.uniforms = shaderMaterialParams.uniforms ?? {};
+        Object.assign(shaderMaterialParams.uniforms, lambertUniforms);
+        super(shaderMaterialParams);
+        this.name = 'LayeredMaterial';
```
HoloTheDrunk: Might as well make all names optional based on `__DEBUG__` no? Is there a reason we'd keep certain names in release builds?

packages/Main/test/unit/entwine.js
HoloTheDrunk: Improper application state assumptions seem to be a recurring theme in the testsuite, I fear we may run into many more like this as we refactor things... Thank you for noticing this one.

HoloTheDrunk (approved):
LGTM

packages/Main/src/Renderer/LayeredMaterial.ts
Neptilo: Not to my knowledge. I must have missed that one.

packages/Main/test/unit/entwine.js
Neptilo: And as a follow-up: https://github.com/iTowns/itowns/issues/2645

PR #2590, opened 2025-08-19
Change document.location uses to fix iframes

HoloTheDrunk:
## Description

Changed the use of `document.location` to ~~`document.baseURI`~~ `import.meta.url`.

## Motivation and Context

`document.location` can be unreliable in certain contexts, like within an iframe created with an `srcdoc` instead of `src`, where it is reported as `about:srcdoc`, while ~~`document.baseURI`~~ `import.meta.url` is generally consistent.

### DIFF
--- a/packages/Main/src/Core/Scheduler/Scheduler.js
+++ b/packages/Main/src/Core/Scheduler/Scheduler.js
@@ -155,7 +155,7 @@ Scheduler.prototype.execute = function execute(command) {
     // parse host
     const layer = command.layer;
     const host = layer.source && layer.source.url && layer.source.url !== 'none' ?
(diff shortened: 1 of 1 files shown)


Desplandis:
Or we could use [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) to have a consistent URL relative to the module! It has also the advantage to be in the ECMAScript standard.

HoloTheDrunk:
> Or we could use [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) to have a consistent URL relative to the module! It has also the advantage to be in the ECMAScript standard.

That does work fine, let's do that.

Desplandis (approved):

PR #2594, opened 2025-09-01
Allow deferred style property definition

Neptilo:
## Description
Allow assigning a function to a style property after its declaration.

## Motivation and Context
After a style is constructed from initial properties, the properties can only be (re)defined as a constant value, not assigned to a function. This can be restrictive in some cases, and the doc doesn´t mention this restriction.
Tested on Firefox, Ubuntu.

### DIFF
--- a/packages/Main/src/Core/Style.js
+++ b/packages/Main/src/Core/Style.js
@@ -138,9 +138,6 @@ function defineStyleProperty(style, category, parameter, userValue, defaultValue
             },
             set: (v) => {
                 property = v;
-                if (parameter == 'extrusion_height') {
-                    style[category].extrusionHeightSet = v != undefined;
-                }
             },
         });
 }
@@ -466,7 +463,27 @@ class Style {
         defineStyleProperty(this, 'fill', 'opacity', params.fill.opacity, 1.0);
         defineStyleProperty(this, 'fill', 'pattern', params.fill.pattern);
         defineStyleProperty(this, 'fill', 'base_altitude', params.fill.base_altitude, baseAltitudeDefault);
-        defineStyleProperty(this, 'fill', 'extrusion_height', params.fill.extrusion_height);
+
+        // define a special case for extrusion_height
+        // to be able to know if user set it or not without calling the getter
+        this._extrusionHeight = params.fill.extrusion_height;
+        Object.defineProperty(
+            this.fill,
+            'extrusion_height',
+            {
+                get: () => {
+                    if (this._extrusionHeight != undefined) {
+                        return readExpression(this._extrusionHeight, this.context);
+                    }
+                    const dataValue = this.context.featureStyle?.fill?.extrusion_height;
+                    if (dataValue != undefined) { return readExpression(dataValue, this.context); }
+                    return undefined;
+                },
+                set: (v) => {
+                    this._extrusionHeight = v;
+                },
+            },
+        );
 
         this.stroke = {};
         defineStyleProperty(this, 'stroke', 'color', params.stroke.color);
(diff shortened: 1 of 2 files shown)


packages/Main/src/Core/Style.js
```diff
@@ -139,6 +138,9 @@ function defineStyleProperty(style, category, parameter, userValue, defaultValue
             },
             set: (v) => {
                 property = v;
+                if (parameter == 'extrusion_height') {
+                    style[category].extrusionHeightSet = v != undefined;
+                }
```
mgermerie: I think I'm more found of the previous implementation even if it is also not perfect I admit (having an exception in `Style` just for conversion reasons is no the best but I don't see how it can be avoided).

I wonder if having this check in every properties setter is really necessary ? If `extrusion-height` is not set in the first place, then the check in `Feature2Mesh` line 614 ([here](https://github.com/iTowns/itowns/pull/2594/files#diff-8975429391aeeb74ee33fa889bb4f33954668f5adbcf36862f6ffae35c82046cL614)) will fail. If user later changes the style by setting `fill.extrusion_height` to something, the test will succeed. This should provide the same behavior as with your introduced `extrusionHeightSet` property, without polluting the setter for every other properties. Or perhaps I am missing some point ?

mgermerie (commented):
Except for one interrogation, it looks good to me. Thanks!

packages/Main/src/Core/Style.js
Neptilo: This is a bit old in my memory (2 months), but I guess I had to do it because of the case where you unset the extrusion (set it to `undefined`). In that case, the style still includes the `extrusion_height` property so the `includes` condition will succeed and try to extrude polygons when it shouldn't, but we also can't check for the property's value directly, because if it is defined as a function callback, calling the getter without a context (of a specific feature that would pass an argument to that callback) would throw an error. I hope my explanation is clear. Don't hesitate to ask if not.

mgermerie: Ok, thanks for the explanation. I didn't firgure the case user set extrusion to `undefined` afterwards, and the modification you make is thus indeed mandatory.

mgermerie (approved):

packages/Main/src/Core/Style.js
mgermerie: After discussing with @ftoromanoff, we believe a cleaner way would be to define a specific method equivalent to `defineStyleProperty` to handle the `extrusion_height` case. This would avoid having a specific case in a method that aims to be generic.

The idea is to store the raw value set for `extrusion_height` in a property and evaluate it only in `extrusion_height` getter. It could look like this on line 469:

```js
this._extrusionHeight = params.fill.extrusion_height;
Object.defineProperty(
    this.fill,
    'extrusion_height',
    {
        get: () => {
            if (this._extrusionHeight != undefined) { return readExpression(this._extrusionHeight, this.context); }
            const dataValue = this.context.featureStyle?.fill?.['extrusion_height'];
            if (dataValue != undefined) { return readExpression(dataValue, style.context); }
        },
        set: (v) => {
            this._extrusionHeight = v;
        },
    },
);
```

In `Feature2Mesh`, we could therefore check the value of `Style._extrusionHeight` without executing it if it is a method, so have something like:

```js
if (style.fill?._extrusionHeight !== undefined) {
   // then convert an extruded polygon...
```

mgermerie (changes requested):

packages/Main/src/Core/Style.js
Neptilo: OK, good idea.
To follow the OOP rules of encapsulation a little better, and later allow for a smoother migration to TS, I suggest considering `_extrusionHeight` as private, as the underscore suggests, and to only expose it through this:
```js
isExtruded() {
    return this._extrusionHeight != undefined;
}
```

mgermerie: Yep, good idea as well! :+1:

mgermerie (approved):

PR #2595, opened 2025-09-02
Make polygons use style opacity

Neptilo:
## Description
This makes use of the opacity defined in the style of polygon features to apply it to their material.

## Motivation and Context
Previously, only the opacity property defined in the layer was applied to polygons. The one defined in the style was not applied.
In this PR, I assumed these two opacity properties were meant to be used in association, multiplying them together to get the actual opacity.
I based it on the yet unmerged `Desplandis/fix/depth-transparency` branch to avoid any conflicts.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Use Node.js 22.x
         uses: actions/setup-node@v4
         with:
-          node-version: 22.x
+          node-version: 22.17
           cache: 'npm'
 
       # Install packages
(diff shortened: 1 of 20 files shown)


Neptilo:
@Desplandis After the recent merged fixes regarding opacity/transparency, this PR would need to change quite a bit. Since I needed it for my current work, I made the updated changes directly in the branch I'm working on. Should I:
1. close this PR and include the updated changes directly in the future branch,
2. or update this PR so it will create a separate entry in the change log?

(Option 1 would be the easier for the rebasing.)

PR #2597, opened 2025-09-10
Dynamic extruded polygon style update

Neptilo:
## Description
This allows style properties of extruded polygons to be updated after their creation, without re-instantiating geometries or recreating layers.

## Motivation and Context
To modify the style of a polygon after its creation, the only way was previously to recreate the entire layer, which is very sub-optimal, and the fact of recreating layers currently even leads to bugs.

Tested on Firefox, Ubuntu.

### DIFF
--- a/packages/Main/src/Core/Style.js
+++ b/packages/Main/src/Core/Style.js
@@ -193,6 +193,10 @@ export class StyleContext {
         return this.#localCoordinates.setFromArray(vertices, offset);
     }
 
+    getGeometry() {
+        return this.#geometry;
+    }
+
     get properties() {
         return this.#geometry.properties;
     }
--- a/packages/Main/src/Core/Style.js
+++ b/packages/Main/src/Core/Style.js
@@ -193,7 +193,7 @@ export class StyleContext {
         return this.#localCoordinates.setFromArray(vertices, offset);
     }
 
-    getGeometry() {
+    get geometry() {
         return this.#geometry;
     }
 
--- a/packages/Main/src/Converter/Feature2Mesh.js
+++ b/packages/Main/src/Converter/Feature2Mesh.js
@@ -524,7 +524,7 @@ function updateExtrudedPolygonVertices(featureMesh, options, vertices, colors, b
     coord.setCrs(context.collection.crs);
 
     // geometry range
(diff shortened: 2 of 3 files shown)


ftoromanoff:
If I undesrtand this PR correctly, you try to be able to dynamicaly change style, but only for the color and the height of extruded polygon ?
Instead of tring to tweak the current code to allow these 2 properties to be modified, shouldn't we rethink it a bit more globaly to allow all style properties to be dynamically change ?

packages/Main/src/Core/Style.js
```diff
@@ -193,6 +193,10 @@ export class StyleContext {
         return this.#localCoordinates.setFromArray(vertices, offset);
     }
 
+    getGeometry() {
+        return this.#geometry;
+    }
+
```
ftoromanoff, line 199: I think a real getter should be more appropriate?

Neptilo:
> If I undesrtand this PR correctly, you try to be able to dynamicaly change style, but only for the color and the height of extruded polygon ? Instead of tring to tweak the current code to allow these 2 properties to be modified, shouldn't we rethink it a bit more globaly to allow all style properties to be dynamically change ?

Yes. I'm actually going further with this to generalize it to all properties of all feature types, as part of my line extrusion implementation. Do you think we should close this PR and wait for the more complete implementation in the future PR for line extrusion?

ftoromanoff:
> > If I undesrtand this PR correctly, you try to be able to dynamicaly change style, but only for the color and the height of extruded polygon ? Instead of tring to tweak the current code to allow these 2 properties to be modified, shouldn't we rethink it a bit more globaly to allow all style properties to be dynamically change ?
> 
> Yes. I'm actually going further with this to generalize it to all properties of all feature types, as part of my line extrusion implementation. Do you think we should close this PR and wait for the more complete implementation in the future PR for line extrusion?

In my opion you can keep in this PR, the clean up and the refactor of cade that will be needed for the generalization, and removing everthing that's lined with the dynamic approach.
If it's too compplicated or nothing to keep, then indeed you can close it.

Neptilo:
The future implementation will change again too much, so I think it's better to close this PR.

Issue #2596, opened 2025-09-10
PointCloud raycast


ketourneau:
I am trying to get points from raycasting on point cloud (I want to color them).

## Your Environment
* Version used: 2.45.2-next.7
* Browser Name and version: firefox 142.0.1
* Operating System and version (desktop or mobile): desktop macOS 15.6.1
* Link to your project: https://itowns-pointcloud-raycast-bug.netlify.app/examples/#potree2_25d_map

## Context
I would like to get all points from raycast on pointcloud with itowns.

## Steps to Reproduce
Go to https://itowns-pointcloud-raycast-bug.netlify.app/examples/#potree2_25d_map and you need to double-click on point cloud.
Source code : https://github.com/bloc-in-bloc/itowns/blob/f/point-cloud-raycast-bug/examples/potree2_25d_map.html#L102

## Expected Behavior
Points in raycast area must be colored.

## Actual Behavior
Some points are not colored.
<img width="1134" height="768" alt="Image" src="https://github.com/user-attachments/assets/b2d94db7-920a-446e-8b13-3560910f2674" />

## Possible Cause/Fix/Solution
Maybe something is wrong with ```view.pickObjectsAt(event, 100, potreeLayer);```

Desplandis:
Managed to reproduce it on master, will look to the picking code and intermediate render of points id!

ketourneau:
@Desplandis I think I found the fix for this bug. I created a PR

PR #2598, opened 2025-09-19
fix(Picking): Fix traversePickingCircle for pickPointsAt

ketourneau:
## Description
Fix for https://github.com/iTowns/itowns/issues/2596

## Motivation and Context
In `traversePickingCircle` there's a bug in the buffer indexing calculation.
If we use the same logic at `traversePickingCircle` for `pickObjectsAt` it solve the problem.
Picking output on point cloud is now a complete circle (as expected).

## Screenshots
Capture d’écran 2025-09-19 à 16 46 35

### DIFF
--- a/packages/Main/src/Core/Picking.js
+++ b/packages/Main/src/Core/Picking.js
@@ -147,7 +147,12 @@ export default {
         const candidates = [];
 
         traversePickingCircle(radius, (x, y) => {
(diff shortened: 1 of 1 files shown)


Desplandis:
It was indeed incorrect, thanks for the fix!

Desplandis (approved):

PR #2599, opened 2025-09-23
Fix incorrectly built points URL in Potree nodes

Desplandis:
## Description

This PR fixed incorrectly built points URL in PotreeLayer. Added a test for regression.

### DIFF
--- a/packages/Main/src/Core/PotreeNode.js
+++ b/packages/Main/src/Core/PotreeNode.js
@@ -13,7 +13,7 @@ class PotreeNode extends PointCloudNode {
 
         this.depth = 0;
 
-        this.hierarchyKey = 'r';
+        this.hierarchyKey = '';
 
         this.baseurl = layer.source.baseurl;
     }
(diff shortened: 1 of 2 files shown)


jailln:
Can we replace `node.hierarchyKey = this.hierarchyKey + indexChild;` by `node.hierarchyKey = indexChild;`  in :
```
add(node, indexChild) {
        node.hierarchyKey = this.hierarchyKey + indexChild;
        node.depth = this.depth + 1;
        super.add(node, indexChild);
    }
```

Desplandis:
I don't think this is equivalent since we are assigning the hierarchy key of the child node: The key is the parent hierarchy key (here `this.hierarchyKey` post-fixed by its position in the list of children of the parent node (here `indexChild`).

I know that this code is ugly but I don't think I will fix it in this PR. I will propose some refactoring in the near future that will unify the behavior of all point-cloud layers (just a single `PointCloudLayer` with multiple compatible sources instead of the current mess). =)

jailln (approved):

jailln:
Ok, fine by me, thanks for the details :)

PR #2601, opened 2025-09-29
example(3DTiles): fix function naming error

mgermerie:
## Description

Fix function naming error in 3D Tiles loader example, which prevents loading 3D Tiles from any user given URL.


Desplandis:
Introduced in #2296, thanks for the fix!

jailln (approved):

PR #2602, opened 2025-09-30
fix(3d-tiles): per material uniform setting

Desplandis:
## Description

Let user set in 3D Tiles Layer per material uniforms.

### DIFF
--- a/packages/Main/src/Layer/OGC3DTilesLayer.js
+++ b/packages/Main/src/Layer/OGC3DTilesLayer.js
@@ -205,6 +205,7 @@ function referPointsMaterialProperties(material, layer) {
         get: () => {
             const t = material.opacity < 1.0
                 || material.classificationTexture.userData.transparent
+                || material.visibilityTexture.userData.transparent
                 || transparent;
             if (t != tPrev) {
                 material.needsUpdate = true;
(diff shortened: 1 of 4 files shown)


Desplandis:
@AnthonyGlt Does this fix your issue?

AnthonyGlt:
> @AnthonyGlt Does this fix your issue?

Thank you @Desplandis , I'll try that !
We also have the issue with the featuregeometrylayers , so should we not do the same in packages/Main/src/Layer/ReferencingLayerProperties.js ?

Desplandis:
@AnthonyGlt Done the same for geometry layers. Could you test it with your use-case?

packages/Main/src/Layer/OGC3DTilesLayer.js
```diff
@@ -196,28 +199,32 @@ function referMaterialProperties(material, layer) {
...
-    let _transparent = material.transparent;
+    let transparent = material.transparent;
+    let tPrev = transparent;
     Object.defineProperty(material, 'transparent', {
         get: () => {
-            const transparent = material.opacity < 1.0
-                || material.classificationTexture.userData.transparent;
-            if (transparent != _transparent) {
+            const t = material.opacity < 1.0
+                || material.classificationTexture.userData.transparent
```
AnthonyGlt, line 207: Should we not have a similar approach for the visibilityTexture ?

AnthonyGlt:
Thank you @Desplandis , this PR is resolving my issues with transparency/opacity control ! :+1:

AnthonyGlt (approved):

PR #2603, opened 2025-10-01
Remove reference to layer in PointCloudNode

ftoromanoff:
In reaction to PR #2504, we assume it's not the responsabilty of the node to know the layer, but the opposite.
Moreover we use mostly the layer to get his source and his metadata.

This PR aims at removing layer from the Point Cloud Node and focus only on the source.

~PS: 
The 2nd [commit](https://github.com/iTowns/itowns/pull/2603/commits/e49706e3d4eaba38052336747bcec0b619fe43e0) can't be kept, I'm waiting for an other ddataset to change it. I will remove it after the testing and the review if I don't get the new one early enought.~

### DIFF
--- a/packages/Main/src/Core/CopcNode.js
+++ b/packages/Main/src/Core/CopcNode.js
@@ -14,7 +14,6 @@ function buildVoxelKey(depth, x, y, z) {
  * @property {number} entryOffset - Offset from the beginning of the file of
  * the node entry
  * @property {number} entryLength - Size of the node entry
- * @property {CopcLayer} layer - COPC layer the node belongs to.
  * @property {number} depth - Depth within the octree
  * @property {number} x - X position within the octree
  * @property {number} y - Y position within the octree
@@ -26,18 +25,18 @@ class CopcNode extends PointCloudNode {
     /**
      * Constructs a new instance of a COPC Octree node
      *
-     * @param {number} depth - Depth within the octree
-     * @param {number} x - X position within the octree
-     * @param {number} y - Y position within the octree
-     * @param {number} z - Z position with the octree
+     * @param {number} depth - Depth within the octree.
+     * @param {number} x - X position within the octree.
+     * @param {number} y - Y position within the octree.
+     * @param {number} z - Z position with the octree.
      * @param {number} entryOffset - Offset from the beginning of the file to
-     * the node entry
-     * @param {number} entryLength - Size of the node entry
-     * @param {CopcLayer} layer - Parent COPC layer
-     * @param {number} [numPoints=0] - Number of points given by this entry
+     * the node entry.
+     * @param {number} entryLength - Size of the node entry.
+     * @param {CopcSource} source - Data source (COPC) of the node.
+     * @param {number} [numPoints=0] - Number of points given by this entry.
      */
-    constructor(depth, x, y, z, entryOffset, entryLength, layer, numPoints = 0) {
-        super(numPoints, layer);
+    constructor(depth, x, y, z, entryOffset, entryLength, source, numPoints = 0) {
+        super(numPoints, source);
         this.isCopcNode = true;
 
         this.entryOffset = entryOffset;
@@ -64,10 +63,10 @@ class CopcNode extends PointCloudNode {
      * @param {number} size
      */
     async _fetch(offset, size) {
-        return this.layer.source.fetcher(this.layer.source.url, {
-            ...this.layer.source.networkOptions,
+        return this.source.fetcher(this.source.url, {
+            ...this.source.networkOptions,
             headers: {
-                ...this.layer.source.networkOptions.headers,
+                ...this.source.networkOptions.headers,
                 range: `bytes=${offset}-${offset + size - 1}`,
             },
         });
@@ -145,7 +144,7 @@ class CopcNode extends PointCloudNode {
             z,
             offset,
             byteSize,
-            this.layer,
+            this.source,
             pointCount,
         );
         this.add(child);
@@ -162,12 +161,11 @@ class CopcNode extends PointCloudNode {
         }
 
         const buffer = await this._fetch(this.entryOffset, this.entryLength);
-        const geometry = await this.layer.source.parser(buffer, {
+        const geometry = await this.source.parser(buffer, {
             in: {
-                ...this.layer.source,
+                ...this.source,
                 pointCount: this.numPoints,
             },
-            out: this.layer,
         });
 
         return geometry;
--- a/packages/Main/src/Core/EntwinePointTileNode.js
+++ b/packages/Main/src/Core/EntwinePointTileNode.js
@@ -44,13 +44,13 @@ class EntwinePointTileNode extends PointCloudNode {
      * @param {number} z - The z coordinate of the node in the tree - see the
      * [Entwine
      * documentation](https://entwine.io/entwine-point-tile.html#ept-data)
-     * @param {EntwinePointTileLayer} layer - The layer the node is attached to.
+     * @param {EntwinePointTileSource} source - Data source (Ept) of the node.
      * @param {number} [numPoints=0] - The number of points in this node. If
      * `-1`, it means that the octree hierarchy associated to this node needs to
      * be loaded.
      */
-    constructor(depth, x, y, z, layer, numPoints = 0) {
-        super(numPoints, layer);
+    constructor(depth, x, y, z, source, numPoints = 0) {
+        super(numPoints, source);
         this.isEntwinePointTileNode = true;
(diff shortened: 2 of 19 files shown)


examples/potree2_25d_map.html
```diff
@@ -81,8 +82,10 @@
                 potreeLayer = new itowns.Potree2Layer('Lion', {
                     source: new itowns.Potree2Source({
                         file: 'metadata.json',
-                        url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/potree2.0/lion',
-                        crs: view.referenceCrs,
+                        // url: 'https://raw.githubusercontent.com/iTowns/iTowns2-sample-data/master/pointclouds/potree2.0/lion',
+                        // crs: view.referenceCrs,
+                        url: 'https://media.githubusercontent.com/media/itowns/iTowns2-sample-data/potreeData/pointclouds/potree2.0/Arene-1_L93',
```
jailln: Is that what you said should be modified before merging?

packages/Main/src/Core/CopcNode.js
```diff
@@ -14,7 +14,7 @@ function buildVoxelKey(depth, x, y, z) {
  * @property {number} entryOffset - Offset from the beginning of the file of
  * the node entry
  * @property {number} entryLength - Size of the node entry
- * @property {CopcLayer} layer - COPC layer the node belongs to.
+ * @property {CopcSource} source - COPC source the node belongs to.
```
jailln: What do you think of "Data source (COPC) of the node" instead ? + Can you use the same comment below in the constructor doc please ?

packages/Main/src/Core/CopcNode.js
```diff
@@ -162,12 +162,12 @@ class CopcNode extends PointCloudNode {
...
         const buffer = await this._fetch(this.entryOffset, this.entryLength);
-        const geometry = await this.layer.source.parser(buffer, {
+        const geometry = await this.source.parser(buffer, {
             in: {
-                ...this.layer.source,
+                ...this.source,
                 pointCount: this.numPoints,
             },
-            out: this.layer,
+            // out: this.layer,
```
jailln: to remove ? removing it won't cause a problem ?

packages/Main/src/Core/EntwinePointTileNode.js
```diff
@@ -44,13 +44,13 @@ class EntwinePointTileNode extends PointCloudNode {
      * @param {number} z - The z coordinate of the node in the tree - see the
      * [Entwine
      * documentation](https://entwine.io/entwine-point-tile.html#ept-data)
-     * @param {EntwinePointTileLayer} layer - The layer the node is attached to.
+     * @param {EntwinePointTileSource} source - The source the node is attached to.
```
jailln: Same proposition than for COPC

packages/Main/src/Core/PointCloudNode.js
```diff
@@ -57,8 +58,11 @@ class PointCloudNode extends THREE.EventDispatcher {
     }
 
     load() {
-        return this.layer.source.fetcher(this.url, this.layer.source.networkOptions)
-            .then(file => this.layer.source.parse(file, { out: this.layer, in: this.layer.source }));
+        return this.source.fetcher(this.url, this.source.networkOptions)
+            .then(file => this.source.parse(file, {
+                in: this.source,
+                // out: this.layer,
```
jailln: same here, out: this.layer is not needed anymore ?

packages/Main/src/Core/Potree2Node.js
```diff
@@ -75,14 +76,14 @@ class Potree2Node extends PotreeNode {
...
+        return this.source.fetcher(this.url, this.networkOptions(this.byteOffset, this.byteSize))
+            .then(file => this.source.parser(file, {
                 in: {
-                    source: this.layer.source,
+                    source: this.source,
                     bbox: this.bbox,
                     numPoints: this.numPoints,
                 },
-                out: this.layer,
+                // out: this.layer,
```
jailln: same here for out: this.layer

packages/Main/src/Core/PotreeNode.js
```diff
@@ -7,23 +7,23 @@ import PointCloudNode from 'Core/PointCloudNode';
...
-    constructor(numPoints = 0, childrenBitField = 0, layer) {
-        super(numPoints, layer);
+    constructor(numPoints = 0, childrenBitField = 0, source) {
+        super(numPoints, source);
         this.childrenBitField = childrenBitField;
 
         this.depth = 0;
 
-        this.hierarchyKey = '';
+        this.hierarchyKey = 'r';
```
jailln, line 16: it was changed the other way around in #2599, on which version do we settle @Desplandis @ftoromanoff ?

packages/Main/src/Core/PotreeNode.js
```diff
@@ -102,9 +102,9 @@ class PotreeNode extends PointCloudNode {
                     if (snode.childrenBitField & (1 << indexChild) && (offset + 5) <= blob.byteLength) {
                         const childrenBitField = view.getUint8(offset); offset += 1;
                         const numPoints = view.getUint32(offset, true) || this.numPoints; offset += 4;
-                        const child = new PotreeNode(numPoints, childrenBitField, this.layer);
+                        const child = new PotreeNode(numPoints, childrenBitField, this.source);
                         snode.add(child, indexChild);
-                        if ((child.id.length % this.layer.hierarchyStepSize) == 0) {
+                        if (((child.id.length - 1) % this.source.hierarchyStepSize) == 0) {
```
jailln: why this change of adding -1?

jailln (commented):
Thanks for taking the time to do these changes right away!

packages/Main/src/Core/PotreeNode.js
ftoromanoff: It was exactly why there was the bug that Desplandis fixed with the 'r'.
The question we will have to answer is to we keep the r in the PotreeNode id or not.

PS: I rethink about it and insead of using the length of the id (or Hierarchykey) I think it will be more clear to use the depth or the node (it remove the need (or not) of the -1. And i fix an other problem linnked that we weren't really able to see as we don't have currently a dataset with subfiles.

packages/Main/src/Core/CopcNode.js
ftoromanoff: Removing it doesn't cause any problem. I kept it as I thought it was a signature of all the parser, but it's seems not to be.
I will remove it.

examples/potree2_25d_map.html
ftoromanoff: Yes, The data set that I use here, can't be use in the iTowns website.

packages/Main/src/Core/PotreeNode.js
ftoromanoff: see next comment

packages/Main/src/Core/Potree2Node.js
ftoromanoff: done

packages/Main/src/Core/PointCloudNode.js
ftoromanoff: done

packages/Main/src/Core/CopcNode.js
ftoromanoff: done

packages/Main/src/Core/CopcNode.js
ftoromanoff: I move the property source documentation to PointCloudNode and change the sentence.

packages/Main/src/Core/EntwinePointTileNode.js
ftoromanoff: see previous comment

packages/Main/src/Core/PotreeNode.js
jailln: I'm not familiar enough with the context here so I trust you. @Desplandis do you have an opinion about that since you worked on it ?

examples/potree2_25d_map.html
ftoromanoff: Commit removed from the PR.

packages/Main/src/Core/PotreeNode.js
Desplandis: As long as it works, I approve!

packages/Main/src/Core/PotreeNode.js
Desplandis: As long as it works, I approve!

jailln (approved):
thanks !

Issue #2607, opened 2025-10-06
State of tree-shaking in iTowns


Desplandis:
## Context
There are some complaints from users that bundles including iTowns are too big. Using only specific parts of the library (e.g. only the 3D Tiles rendering) does not reduce the bundle size (or only by tight margins). There seems to be a problem when tree-shaking our modules (hence the bug report).

This "bug report" will act as a "meta-issue" for all tree-shaking bugs. It will includes:
- test cases used to replicate the tree-shaking issues, usually simple programs (e.g. a view, a source and a layer) which only use some sub-parts of the library. Those could be integrated into our CI pipeline to measure variation of the bundles sizes.
- a non-exhaustive list of identified non tree-shaking friendly patterns (and optionnaly a proposed solution) that we will translate into sub-issues.

## More context
A tree-shaking pass (i.e. dead code elimination) is implemented in all major bundlers (webpack, rollup, esbuild, ...) for ES modules. Since js does not have a native way to mark a statement as pure and bundlers are conservative (for correctness), the tree-shaking pass will mark some statements as having side-effects.

This is valid for the following statement in the module scope:
```js
proj4.def('EPSG:4326', ...);
```
since this have a side-effect outside of its module the `proj4.def` array (which contains the registered projection definitions).

But not valid for the following statement in the module scope:
```js
const _x = new THREE.Vector3():
```

There is two common ways to address this which could be usef in combination:
- specify files in the **"sideEffects"** field of `package.json` to indicate which modules contain side effects.
- using **`/*@__PURE__*/`** comments for pure functions to signal to the bundler that they can be safely removed when they're not used.

## Test cases
I used the `rollup` bundler as a way to analyse the tree-shaking bottleneck: `rollup --input [filename] -c rollup.config.mjs` with the following configuration file:
```js
// rollup.config.mjs
import resolve from '@rollup/plugin-node-resolve';

export default [
    {
        plugins: [
            resolve(),
        ],
        output: [
            {
                format: 'esm',
                file: 'test/treeshake/index.bundle.js',
                sourcemap: true,
            },
        ],
        experimentalLogSideEffects: true,
    },
];
```

### A - The `@itowns/geographic` and `CRS` test case
```js
import { CRS } from '@itowns/geographic`;

CRS.defs('EPSG:4326', '+title=WGS 84 (long/lat) +proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees');
```

### B - The `itowns` and 3D Tiles Layer test case
[...]

PR #2608, opened 2025-10-06
Fix tree-shaking in @itowns/geographic

Desplandis:
## Description
This PR fixes module globals tree-shaking related issues (see #2607 for more context). It also updates proj4 to the latest version.


packages/Geographic/src/Coordinates.ts
```diff
@@ -1,16 +1,15 @@
...
+import { Vector3, type Vector3Like, type Matrix4, MathUtils } from 'three';
+import proj4, { type Converter } from 'proj4';
 import Ellipsoid from './Ellipsoid';
 import * as CRS from './Crs';
 
-import type { ProjectionAlias } from './Crs';
+import type { ProjectionLike } from './Crs';
 
-const ellipsoid = new Ellipsoid();
+const ellipsoid = /* @__PURE__ */ new Ellipsoid();
```
jailln, line 8: Did you add all these annotations manually ?

Desplandis: Either mark it manually (which I could sed with a regex) or use `sideEffects` on `package.json`. However I decided to follow `three.js` there.

jailln: Since we use and manipulate global variables, `sideEffects` may be risky :( I think we should avoid marking it manually though. It seems that you only added it to the threejs instantiations, maybe we could automatize that somehow?

How does it impacts the bundle size ? Did you measure it?

Desplandis: Like I said in the linked issue, I tried it with some minimal example which indeed have some impact when using `@itowns/geographic` but not that much when using `itowns` (since we use those modules everywhere in the codebase).

Desplandis: We could automatise it (with a sed script identifying this pattern), but we'll have to add it manually for all new code. To be honest, this PR will not have that much of an impact (since `Coordinates`, `Vector3`, `Box3`, `Matrix4` & co are used everywhere) on the `itowns` bundle (and client-generated bundle). However, marking as pure some module globals used only on specific part of `itowns` (e.g. `Style`, `ShaderChunk`, ...) which cause a cascade of module imports can have an impact on client-generated bundle not using those specific parts of itowns (e.g. using only pointclouds). Same for my PR on `copc.js` and `3d-tiles-renderer` (the latter blocking tree-shaking of `three` entirely...).

jailln: ok, thanks for the explaination! Let's merge it and see if we generalize that before automatizing.

jailln (approved):

Issue #2530, opened 2025-04-07
[Proposal] A new, more unified way to identify when iTowns view is loaded and stable for testing


airnez:
## Context
During march 2025 iTowns hackaton, we identified that using the view schedueler to subdivide TiledGeometryLayer's nodes was odd. It is an almost instantaneous process, and schedueling it along network request is confusing. The PR https://github.com/iTowns/itowns/pull/2521 and https://github.com/iTowns/itowns/pull/2525 do attempt to adress issue https://github.com/iTowns/itowns/issues/2421.

However, it does break our functionnal tests for a good reason: It becomes harder to know if current iTowns view is stabilized. 

* If the test launches prematurely, some layers might be missing (the scheduler becomes empty inbetween two layers addition). For the `source_stream_wfs_raster test`, that's what happens. `VIEW_EVENTS.LAYERS_INITIALIZED` event is triggered too early.

![Image](https://github.com/user-attachments/assets/fb1c4cf8-bba6-43a8-9664-696b67382ce6)

* If a test prematurely considers that the view is stabilized, it might measure values at the wrong moment. This is the case for `view_multiglobe` test

![Image](https://github.com/user-attachments/assets/dc08a297-66f4-4d2c-95d6-e0329acd82d8)

* It a sub-test ends prematurely, it might influence upcoming tests. This is the case for the two failing `globecontrol` tests.

![Image](https://github.com/user-attachments/assets/f35704d9-c1e4-4292-bd64-ff0f819fbf19)

It is possible to handle all those specific tests. By first querying the datasources prior to view instancing, it should be possible to avoid in-between layers `VIEW_EVENTS.LAYERS_INITIALIZED` trigger. Listening to animations or frameRequesters helps identifying view stabilization. 

It however feels like we are missing a more global approach. Something easy to tell wether or not current view is stabilized or not.

## Description of the proposal
Several implementation ideas (not exclusive)  :
- A specific layer property exposing wether or not a layer is currently "working"
- A new event VIEW_EVENTS.STABILIZED that properly tells when everything is set (how?) 
- A new way to add layers to the view specifying beforehand what will be part of the view. (Passing layer loading promises for instance)

jailln:
Thanks for this analysis and proposals 🙂 

[...]

Issue #1223, opened 2019-09-07
When zooming out, detailed imagery tiles should not disappear if less detailed tiles are not available.


crubier:
When zooming out of a view where all imagery tiles are loaded, the detailed tiles disappear, even before the less detailed tiles are loaded. 

The problem is seen happening repeatedly on this screen capture:

![Screen Recording 2019-09-07 at 11 01 38 mov](https://user-images.githubusercontent.com/2954572/64473462-00618480-d15f-11e9-9773-4d2222d4f83d.gif)

If detailed tiles are available but less detailed tiles are not available, then the detailed tiles should be visible even if they are too detailed for the current zoom level. Temporarily showing a bit too much detail is better than temporarily showing nothing/very low level of detail.

We are considering moving from cesium to itowns, but this (and a few other bugs) are preventing us from doing so.

gchoqueux:
Yes, unfortunately, it is a known behaviour.  We want to improve the rendering, we could plan it in the next 3 to 6 months.

PR #2609, opened 2025-10-06
Simplification of the loading process for rasterized tiles and mesh tiles  

### LINKED ISSUE #2816: Terrain min zoom blocks globe refinement
## Your Environment
* Version used: revision 56f189d 
* Browser: Chrome & Firefox
* Operating System: Ubuntu

## Context
A globe view stops refining correctly when the elevation layer is configured with a minimum zoom. Only a few coarse tiles appear, deeper levels do not load, and camera interactions become unreliable because picking has almost no refined ground to work with.

During the regression, the console repeatedly shows subdivision-blocked for the low-resolution elevation layer, that layer stays at `tileLevel: -1`, and a later frame update can end with "Uncaught TypeError: coordinates must be finite numbers".

## Steps to Reproduce
1. In the view_3d_map.html example, right after `config.source = new itowns.WMTSSource(config.source);` and before `view.addLayer(new itowns.ElevationLayer(...`, add `config.zoom = { min: 11 };`.
2. Load the example and zoom in on the globe. The tiles don't refine.

## Expected Behavior
The globe should keep refining normally. An elevation layer with zoom.min should simply start providing data from its configured minimum zoom without blocking root-level globe subdivision.

## Actual Behavior
Globe subdivision stalls at very low resolution. Refinement stays blocked on the elevation layer, deeper tiles do not load, and navigation and picking quality degrade sharply.

The regression was introduced by commit 3c64961 on master, titled "refacto(Source): refactor extentInsideLimit() to hasData()".
This looks like a regression in the interaction between globe subdivision and elevation-layer availability checks. If an elevation layer declares coverage but cannot fetch data yet because of zoom.min, root nodes appear to wait indefinitely for elevation data instead of continuing refinement until the layer becomes eligible.

gchoqueux:
## Description

This PR proposes a possible simplification of the loading process for **rasterized tiles** and **mesh tile subdivision**.

The goals are:
* remove the `LayeredMaterialNodeProcessing.js` and `TileProvider.js` files;
* move raster processing into `RasterTile` to better align with the iTowns architecture.

### Summary of changes

- `TileMesh` instances are now created **directly in** `TiledGeometryLayer`, **without using commands or providers**.
- Asynchronous tile mesh subdivision has been **preserved** to defer child creation and break the recursive `updateElements` chain.
- `updateLayeredMaterialNodeImagery` and `updateLayeredMaterialNodeElevation` have been **merged into a single lightweight method**, and texture loading is now handled by **`RasterTile#load`**.
- `TileProvider` has been **deleted**.

fix #2816


gchoqueux:
Rebased

gchoqueux:
I’ve come up with a possible way to simplify this — what do you think of these changes?

jailln:
It seems related to #2521 and #2525 which haven't been merged yet; but it goes a bit further, is that right? These PR where not merged because of #2530 . Do you have the same issue?

gchoqueux:
> It seems related to #2521 and #2525 which haven't been merged yet; but it goes a bit further, is that right? These PR where not merged because of #2530 . Do you have the same issue?

 ~~It doesn’t seem to have the same issue in [the functional tests.]( https://github.com/iTowns/itowns/actions/runs/18463907024/job/52602206787?pr=2609)~~
However, I did change the number of displayed tiles in the functional tests — maybe because this PR is not finished yet.

gchoqueux:
~~My bad, there's a issue with  `VIEW_EVENTS.LAYERS_INITIALIZED`, but it is not the same, it's called very late, it's almost the opposite.
I'll investigate~~ [solved](https://github.com/iTowns/itowns/pull/2609#issuecomment-3402644033)

gchoqueux:
@jailln I solved my problem, which was mainly caused by excessive data loading.
The functional tests work with this PR, there isn't issue like #2530

jailln:
> @jailln I solved my problem, which was mainly caused by excessive data loading. The functional tests work with this PR, there isn't issue like #2530

Oh ok I think I know why: in #2521 tilemesh creation doesn't go through the scheduler anymore (since there is no external query, it is not required), that's why it breaks the events. We saw a perf improvement with this change so it needs to be considered in my opinion.

gchoqueux:
> Oh ok I think I know why: in #2521 tilemesh creation doesn't go through the scheduler anymore (since there is no external query, it is not required), that's why it breaks the events. We saw a perf improvement with this change so it needs to be considered in my opinion.

As mentioned in the description, `TileMesh` requests and `TileProvider` are also removed, similar to PR #2521.

I'll also perform performance measurements. What performance results were observed in PR #2521, and in which processes were they measured?

jailln:
Oh ok, sorry I missed that! Nothing was measured, it was only a visual impression (tiles refinement when zooming in seemed faster). It would be nice to measure yes.

I know it's still in draft so you may have noticed it already but currently on the globe tiles don't load until I start zooming in.

gchoqueux:
Yes, there's somes bugs.

There is also the texture loading to be validated :

* Simplification of `LayeredMaterialNodeProcessing` with only one update method for both color and elevation.
* Texture loading is now handled by `RasterTile#load`.

Neptilo:
Hi! I'm starting to look into the loading processes to get familiar with them, in order to improve them after this simplification is completed. Thanks for the work done so far!
Do you need help fixing the remaining bugs?
What is the display bug you mention in `ElevationLayer.anyVisibleData`? I can see that changing the first `current.source` to `current?.source` already makes a big difference, but I assume you are referring to a deeper bug...

gchoqueux:
## Remaining work in this PR

* **finalize** the method for verifying data visibility :  `anyVisibleData` _vs_ `extentInsideLimit` methods
* **rewrite** some unit tests 
* **finalize** `tiles` property in `RasterTile`

## FAQ

> What is the display bug you mention in ElevationLayer.anyVisibleData? 

I don't remember exactly i remove the commentary

> I can see that changing the first `current.source` to `current?.source`, you are referring to a deeper bug ?

I think it's not a bug

> the others bugs : 

- wrong tile visibility
- wrong parent texture update

gchoqueux:
### Next steps to move forward on the PR

- An initial code review by at least 2 reviewers on the architecture and the core aspects of the PR. 
**The main code of this PR hasn’t changed for weeks; only minor formatting changes and rebases have been made.**

**Assistance can be provided on the following points:**
- Validation and finalization of the development of `anyVisibleData`
- Rewriting unit tests and improving coverage
- Help testing this PR on the various itowns examples

packages/Main/src/Core/Prefab/TileBuilder.ts
```diff
@@ -87,36 +87,31 @@ export function newTileGeometry(
...
+    let geometry = cacheTile.get(key);
 
-    // build geometry if doesn't exist
-    if (!promiseGeometry) {
-        let resolve;
-        promiseGeometry = new Promise((r) => { resolve = r; });
-        cacheTile.set(key, promiseGeometry);
+    if (geometry && !geometry.index) {
+        throw new Error('Geometry found in cache without index buffer');
+    }
```
Desplandis: Is this really possible to have a geometry without index here?

packages/Main/src/Core/Prefab/TileBuilder.ts
```diff
@@ -87,36 +87,31 @@ export function newTileGeometry(
...
+            builder,
+            { ...params, center },
+            cachedBuffers !== undefined
+                ? {
+                    index: cachedBuffers.index.array as
+                        Uint8Array | Uint16Array | Uint32Array,
+                    uv: cachedBuffers.uv.array as Float32Array,
+                }
+                : undefined,
+        );
```
Desplandis, line 113: It seems that this code could fail before this PR, are those exceptions catched somewhere?

packages/Main/src/Core/Tile/Tile.ts
```diff
@@ -210,4 +210,57 @@ export function tiledCovering(e: Extent, tms: string) {
     }
 }
 
+export class TileMatrixSetLimits {
```
Desplandis, line 252: This greatly simplifies the code in `TMSSource`!

In the future (maybe for the next hackaton), I will surely try to add a tile grid module similar to openlayer's [TileGrid](https://openlayers.org/en/latest/apidoc/module-ol_tilegrid_TileGrid.html) which will incorporates your changes here.

packages/Main/src/Layer/ElevationLayer.js
```diff
@@ -115,15 +114,16 @@ class ElevationLayer extends RasterLayer {
      * @return     {RasterElevationTile}  The raster elevation node added.
      */
     setupRasterNode(node) {
-        const rasterElevationNode = new RasterElevationTile(this);
+        const tiles = node.getExtentsByProjection(this.crs);
```
Desplandis: TODO: assert this code.

packages/Main/src/Layer/ElevationLayer.js
```diff
@@ -135,11 +135,16 @@ class ElevationLayer extends RasterLayer {
...
 
+        // Init the node by parent
+        rasterElevationNode.initFromParent(node.parent.material?.getElevationTile());
+
         return rasterElevationNode;
     }
 
-    update(context, layer, node, parent) {
-        return updateLayeredMaterialNodeElevation(context, this, node, parent);
+    anyVisibleData(node) {
```
Desplandis: We should find a better name. x)

packages/Main/src/Layer/LayerUpdateStrategy.ts
```diff
@@ -1,4 +1,4 @@
-import { EMPTY_TEXTURE_ZOOM } from 'Renderer/RasterTile';
+import { EMPTY_TEXTURE_ZOOM, RasterTile } from 'Renderer/RasterTile';
```
Desplandis:
```suggestion
import { EMPTY_TEXTURE_ZOOM, type RasterTile } from 'Renderer/RasterTile';
```

packages/Main/src/Layer/RasterLayer.js
```diff
@@ -19,6 +35,7 @@ class RasterLayer extends Layer {
             cacheLifeTime,
         });
 
+        this.visible = true;
```
Desplandis, line 38: Weird that this is not defined in `Layer`.

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -1,9 +1,43 @@
...
 
 export const EMPTY_TEXTURE_ZOOM = -1;
 
+function materialCommandQueuePriorityFunction(material) {
+    // We know that 'node' is visible because commands can only be
+    // issued for visible nodes.
+    // TODO: need priorization of displayed nodes
+    // Then prefer displayed node over non-displayed one
+    return material.visible ? 100 : 10;
+}
```
Desplandis, line 16: We should prioritise either by screen-size error or screen-size radius of the tile. If equality between two tiles, we should prefer visible tiles. Let's leave it for the next PR though.

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -1,9 +1,43 @@
...
+
+function refinementCommandCancellationFn(cmd) {
+    const { requester, view } = cmd;
+    const child = requester.children[0];
+
+    // to do remove useless
+    return !cmd.force && (!view || !requester.parent || !requester.material || !requester.material.visible || child?.visible);
+}
```
Desplandis, line 26: The previous conditions for aborting a command were the following. I've marked the ones that are still present in this PR:
- [x] The `TileMesh` parent is not defined (which likely means it is marked for deletion) or the material is not visible (which means that a more precise LOD is visible).
- [ ] The `TileMesh` has a more precise elevation texture attached, so we cancel the request (e.g. SRTM vs USGS layer in levels > 8 I suppose?). This condition no longer exists, is there a reason it was removed?
- [ ] The layer was removed. This was previously detected using two fragile conditions:
  + no update state on the node for this layer
  + the `_featuresCaches` no longer exists in the source (yikes...)

  I agree that both of these conditions should be removed, but do we still have a reliable way to detect when a layer has been detached or disposed?

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -1,9 +1,43 @@
...
+    return {
+        view,
+        layer: tile.layer,
+        extentsSource,
+        extentsDestination: tile.tiles,
+        requester,
+        priority: materialCommandQueuePriorityFunction(requester.material),
+        earlyDropFunction: refinementCommandCancellationFn,
+        partialLoading: true,
+        force: tile.level == -1,
```
Desplandis: This new `force` parameter seems to prevent this command from being cancelled. Coud you explain the rationale behind the condition `tile.level == -1`?

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -37,9 +71,12 @@ export class RasterTile extends THREE.EventDispatcher {
         }
 
         this.textures = [];
+        this.tiles = tiles;
```
Desplandis, line 74:
```suggestion
        /** @type{Tile[]} */
        this.tiles = tiles;
```

packages/Main/src/Main.js
```diff
@@ -16,7 +16,6 @@ export { default as View } from 'Core/View';
 export { VIEW_EVENTS } from 'Core/View';
 export { default as FeatureProcessing } from 'Process/FeatureProcessing';
 export { default as ObjectRemovalHelper } from 'Process/ObjectRemovalHelper';
-export { updateLayeredMaterialNodeImagery, updateLayeredMaterialNodeElevation } from 'Process/LayeredMaterialNodeProcessing';
```
Desplandis, line 19: I don't think a lot of people used them but this a breaking change and shall be documented as such!

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -58,11 +95,35 @@ export class RasterTile extends THREE.EventDispatcher {
         return this.layer.visible;
     }
 
-    initFromParent(parent, extents) {
+    load(requester, view) {
```
Desplandis, line 98: Similarly to `load`, is there a method to `RasterTile` to "dispose" the textures?

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -136,13 +201,14 @@ export class RasterTile extends THREE.EventDispatcher {
...
+            return texture;
         }
     }
 
-    setTextures(textures, pitchs) {
+    setTextures(textures) {
         this.disposeRedrawnTextures(textures);
         for (let i = 0, il = textures.length; i < il; ++i) {
-            this.setTexture(i, textures[i], pitchs[i]);
+            this.setTexture(i, textures[i], this.tiles[i].offsetToParent(textures[i].extent));
```
Desplandis, line 217: So pitches/offsetScales parameters were always `offsetToParent` of texture extent?

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -313,51 +313,29 @@ class TiledGeometryLayer extends GeometryLayer {
...
-                if (!requestChildrenUpdate) {
-                    return ObjectRemovalHelper.removeChildren(this, node);
-                }
-            }
+        node.material.visible = node.visible;
 
-            return requestChildrenUpdate ? node.children.filter(n => n.layer == this) : undefined;
-        }
+        // [TEMP] find best place to update uniforms
+        node.material.layersNeedUpdate = node.material.visible;
```
Desplandis: This is similar to the old behaviour, but I agree that we should reconsider where we should request a new update of the material!

packages/Main/src/Layer/RasterLayer.js
```diff
@@ -44,6 +61,23 @@ class RasterLayer extends Layer {
...
+
+        const extents  = node.getExtentsByProjection(this.crs)
+            .map(e => e.tiledExtentParent(maxZoom));
+
+        return extents.find(e => this.source.anyVisibleData(e) && e.zoom >= this.zoom.min);
+    }
+
+    update(context, layer, node) {
+        if (layer.visible && !layer.freeze && this.anyVisibleData(node)) {
+            const rasterTile = node.material.getTile(this.id) || this.setupRasterNode(node);
```
Desplandis: So `setupRasterNode` is an abstract method of `RasterLayer`

packages/Main/src/Layer/ElevationLayer.js
```diff
@@ -135,11 +135,16 @@ class ElevationLayer extends RasterLayer {
...
+        const current = node.material.getElevationTile()?.layer;
+        return super.anyVisibleData(node)
+            && (current?.source == this.source || this.source.zoom.min > (current?.source.zoom.max || -1));
```
Desplandis: What is the meaning of `-1` here?

Desplandis (commented):
The proposed changes are fine by me. It makes the code more easy to understand (e.g. no need to track the changes of `layerUpdateState` across 4 different files). I did some high-level comments on the code (and some low-level too sorry), those are some questions about the behaviour before and after this PR.

Note that I did not tested it across our examples, nor did I run the tests.

Could you please provide a list of the remaining things to address in this PR?

packages/Main/src/Core/Prefab/TileBuilder.ts
gchoqueux: no, i remove this test

packages/Main/src/Core/Prefab/TileBuilder.ts
gchoqueux: I not sure there are failures

packages/Main/src/Core/Tile/Tile.ts
gchoqueux: yes I add news features in TileMatrixSetLimits

packages/Main/src/Layer/ElevationLayer.js
gchoqueux: yes, 
`hasData` ?

packages/Main/src/Renderer/RasterTile.js
gchoqueux: it's to force loading the first level, the loading isn't forced when the rastertile is init by the raster tile parent.

packages/Main/src/Main.js
gchoqueux: I add BC in message commit

packages/Main/src/Renderer/RasterTile.js
gchoqueux: yes

packages/Main/src/Renderer/RasterTile.js
gchoqueux: yes

packages/Main/src/Layer/RasterLayer.js
gchoqueux: yes

packages/Main/src/Layer/ElevationLayer.js
gchoqueux: it's for `this.source.zoom.min > -1 ` would be always true

packages/Main/src/Renderer/RasterTile.js
Desplandis: So tile level `-1` is the first level?

packages/Main/src/Renderer/RasterTile.js
Desplandis: Didn't see them though, could you link me the code?

packages/Main/src/Layer/ElevationLayer.js
Desplandis: I'm not good at naming things, but I'll think about it!

packages/Main/src/Layer/ElevationLayer.js
Desplandis: In that case, the data is considered visible if `source.zoom.min < layer.source.zoom.max`? Not sure to follow here.

packages/Main/src/Renderer/RasterTile.js
Desplandis: @gchoqueux

Desplandis:
@gchoqueux Are you available for a session of peer programming on this PR tomorrow or the following monday? I would like to converge on this one. =)

packages/Main/src/Renderer/RasterTile.js
gchoqueux: in layeredMaterial.dispose, ObjectRemovalHelper.js

packages/Main/src/Layer/TiledGeometryLayer.js
gchoqueux: yes, it's not the best place, because `node.material.layersNeedUpdate ` is true each time that `tileMesh` is visible

Desplandis:
I'm currently looking to the PR (hope we can finalise this), catched a regression on `view_25d_map` when moving or zooming out, we don't have textures (neither elevation nor imagery) when loading new levels. I'll try to investigate a bit during my review.

P.S.: `misc_clamp_ground` too

examples/misc_clamp_ground.html
```diff
@@ -66,10 +66,13 @@
                 config.source = new itowns.WMTSSource(config.source);
                 var layer = new itowns.ElevationLayer(config.id, config);
                 view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+
+                return layer;
             }
-            itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
+            const p = itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
```
Desplandis: I think a [top-level `await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) would be clearer. This comment applies to all examples of itowns per-se.

packages/Main/src/Renderer/RasterTile.js
```diff
@@ -131,18 +194,19 @@ export class RasterTile extends THREE.EventDispatcher {
     }
 
     setTexture(index, texture, offsetScale) {
-        if (this.shouldWriteTextureAtIndex(index, texture)) {
-            this.level = (texture && texture.extent) ? texture.extent.zoom : this.level;
-            this.textures[index] = texture || null;
+        if (texture && this.shouldWriteTextureAtIndex(index, texture)) {
```
Desplandis, line 205: Is there some places where we could pass an undefined texture there? Shall those undefined be filtered before at the call sites?

packages/Main/src/Renderer/RasterTile.js
Desplandis: @gchoqueux Are those points addressed?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -500,35 +500,37 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
      * @param renderer - The renderer used to render textures into arrays.
      */
     public updateLayersUniforms(renderer: THREE.WebGLRenderer): void {
-        const colorlayers = this.colorTiles
-            .filter(rt => rt.visible && rt.opacity > 0);
-        colorlayers.sort((a, b) =>
-            this.colorTileIds.indexOf(a.id) - this.colorTileIds.indexOf(b.id),
-        );
+        if (this.layersNeedUpdate) {
```
Desplandis: Prefer an early return with the `!this.layersNeedUpdate` condition.

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -313,51 +313,30 @@ class TiledGeometryLayer extends GeometryLayer {
...
 
-        // do proper culling
         node.visible = !this.culling(node, context.camera);
 
-        if (node.visible) {
-            let requestChildrenUpdate = false;
-
-            node.material.visible = true;
-            node.material.layersNeedUpdate = true;
+        if (node.visible && this.subdivision(context, node)) {
```
Desplandis, line 319: There are some issues with the freshly subdivided tiles which have no terrain and no imagery (e.g. `view_25dmap` and `misc_clamp_ground` issues are flagrant). Re-adding the `hasEnoughTexturesToSubvide` condition seems to fix the problem, however I think the problem is more profound.

Desplandis (changes requested):

Desplandis:
Looking at it, I think the problem is two-fold:
- Since we are not blocking subdivision with `hasEnoughTexturesToSubvide`, we do not trigger parent texture loading
- Since everything is synchronous, when we call `initFromParent`, the texture is not loaded on the parent and we don't have a mechanism to notify descendant tiles of a loaded ancestor texture.

packages/Main/src/Renderer/RasterTile.js
gchoqueux:
```js
//
export const EMPTY_TEXTURE_ZOOM = -1;
```
if `rasterTile.level  ==  EMPTY_TEXTURE_ZOOM` so tile hasn't loaded texture

packages/Main/src/Layer/ElevationLayer.js
gchoqueux: I clarified this by refactoring the code

examples/misc_clamp_ground.html
gchoqueux: done

packages/Main/src/Renderer/RasterTile.js
gchoqueux: If we filter out undefined textures, we lose the indexing needed to assign them

packages/Main/src/Renderer/LayeredMaterial.ts
gchoqueux: done

packages/Main/src/Layer/TiledGeometryLayer.js
gchoqueux: may be add hasEnoughTexturesToSubvide for the first tile ?

gchoqueux: and refactor `hasEnoughTexturesToSubvide` to `LayeredMaterial`

gchoqueux: I add a new wip commit :

5a4280925855869a60099cca9246b1321b6b42d0 fix(TiledGeometryLayer): [WIP][PROPOSAL] avoid to subdivide node mesh if raster tiles aren't loaded

packages/Main/src/Core/Prefab/TileBuilder.ts
```diff
@@ -146,15 +138,15 @@ export function newTileGeometry(
...
+        geometry = new TileGeometry(builder, params, gpuBuffers);
+        geometry.OBB =
+            new OBB(geometry.boundingBox.min, geometry.boundingBox.max);
         geometry.initRefCount(cacheTile, key);
-        (resolve as (value: TileGeometry) => void)(geometry);
 
-        return Promise.resolve({ geometry, quaternion, position });
+        cacheTile.set(key, geometry);
+
+        return { geometry, quaternion, position };
```
ftoromanoff, line 149: Useless line

packages/Main/src/Core/TileGeometry.ts
```diff
@@ -87,6 +89,7 @@ export class TileGeometry extends THREE.BufferGeometry {
             this.setAttribute(`uv_${i}`, bufferAttributes.uvs[i]);
         }
 
+        this.boundingBox = new THREE.Box3();
```
ftoromanoff: useless, the Box3 in instanciated next line.
see https://github.com/mrdoob/three.js/blob/148ef33ecb6d2502ff796d4554abd1549c95d519/src/core/BufferGeometry.js#L645-L649

If the problem come from TS on line 60 then you should add ```| null```

packages/Main/src/Renderer/RasterTile.js
ftoromanoff: I believe it is more ```/** @type{TileMesh[]} */``` and not Tile[]...

packages/Main/src/Core/Prefab/TileBuilder.ts
ftoromanoff, line 79: To rename.
It's unclear what it does...
It returns a Geometry and a position and rotation. It's somewhat closer to an Object3D, than a pure geometry.. And this parameters will be used to instanciate a TileMesh.

'getTileMeshParameters' ? or 'setTileFromGenericTile'

the class 'TileGeometry' being a GlobeTile BufferGeometry ro a generic tile or shared geometry, thus maybe the name SharedTileGeometry migth be clearer for the TileGeometry

packages/Main/src/Layer/ColorLayer.js
```diff
@@ -164,17 +163,21 @@ class ColorLayer extends RasterLayer {
      * @returns     {RasterColorTile}  The raster color node added.
      */
     setupRasterNode(node) {
-        const rasterColorTile = new RasterColorTile(this);
+        const tiles = node.getExtentsByProjection(this.crs);
```
ftoromanoff, line 166: It's more meshes than tiles, to avoid confusion...

packages/Main/src/Layer/ElevationLayer.js
```diff
@@ -115,15 +114,14 @@ class ElevationLayer extends RasterLayer {
      * @returns     {RasterElevationTile}  The raster elevation node added.
      */
     setupRasterNode(node) {
-        const rasterElevationNode = new RasterElevationTile(this);
+        const tiles = node.getExtentsByProjection(this.crs);
```
ftoromanoff, line 117: idem. 'meshes'

packages/Main/src/Layer/Layer.js
```diff
@@ -254,17 +254,19 @@ class Layer extends THREE.EventDispatcher {
...
-        const key = this.source.getDataKey(this.source.isVectorSource ? to : from);
-        let data = this.cache.get(key);
-        if (!data) {
-            data = this.source.loadData(from, this)
-                .then(feat => this.convert(feat, to))
-                .catch((err) => {
-                    this.source.handlingError(err);
-                });
-            this.cache.set(key, data);
+        if (from) {
```
ftoromanoff, line 257: Why add this test ?

packages/Main/src/Layer/LayerUpdateStrategy.ts
```diff
@@ -112,3 +112,17 @@ export function chooseNextLevelToFetch(
     }
     return nextLevelToFetch;
 }
+
+export const nextLevelToFetch = (t: RasterTile) => {
+    const zoom = {
+        min: Math.max(t.layer.zoom.min, t.layer.source.zoom?.min),
+        max: Math.min(t.layer.zoom.max, t.layer.source.zoom?.max),
+    };
+    return chooseNextLevelToFetch(
```
ftoromanoff, line 121: This function (nextLevelToFetch() returns chooseNextLevelToFetch().

And when we looked at it, it returns nextLevelToFetch. I find it quite confusing...
I would sugest to rename 'nextLevelToFetch' in chooseNextLevelToFetch() function as 'levelToFetch'

packages/Main/src/Layer/ElevationLayer.js
```diff
@@ -115,15 +114,14 @@ class ElevationLayer extends RasterLayer {
...
+
+        const rasterElevationNode = new RasterElevationTile(this, tiles);
 
-        node.material.setElevationTile(rasterElevationNode);
-        node.material.setElevationTileId(this.id);
         // bounding box initialisation
         const updateBBox = () => node.setBBoxZ({
             min: rasterElevationNode.min, max: rasterElevationNode.max, scale: this.scale,
         });
```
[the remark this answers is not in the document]
gchoqueux: The bounds are updated when the textures are applied. When the textures change, a `view.notifyChange` with a redraw is called, so the culling is recomputed.

ftoromanoff (commented):
Some remarks linked to the readabilty (in my opinion)

packages/Main/src/Renderer/RasterTile.js
gchoqueux: they are Tile or Extent

packages/Main/src/Core/Prefab/TileBuilder.ts
gchoqueux: prepareTileMesh ?

packages/Main/src/Layer/ColorLayer.js
gchoqueux: it's Tile class

packages/Main/src/Layer/Layer.js
gchoqueux: some Tile is undefined

packages/Main/src/Core/TileGeometry.ts
gchoqueux: i have this ts error https://github.com/iTowns/itowns/actions/runs/35084282273/job/104755316530?pr=2609#step:7:29

Whereas the `this.boundingBox` is never `null`.

ftoromanoff: That's where you need to add ```| null``` at line 60 in TileGoemtry.js
```
public boundingBox: THREE.Box3 | null;
```
and 
```
const bbox = geometry.boundingBox as THREE.Box3;
geometry.OBB = new OBB(bbox.min, bbox.max);
```
in TileBuilder.js line 141-143 as it was before.

packages/Main/src/Renderer/RasterTile.js
ftoromanoff: `tiles` is the 2nd argument of the constructor.

This constructor being called in:
- ColorLayer#setupRasterNode()
with `const tiles = node.getExtentsByProjection(this.crs);`
- ElevationLayer#setupRasterNode()
with `const tiles = node.getExtentsByProjection(this.crs);`

And getExtentsByProjection() is a TileMesh method that return a `Tile[]` or `undefined`.

My bad indeed.

packages/Main/src/Core/Prefab/TileBuilder.ts
ftoromanoff: It doesn't really prepare the tileMesh. It gives data to prepare (or set up) a TileMesh.

packages/Main/src/Layer/ColorLayer.js
ftoromanoff: my bad

airnez:
@Desplandis asked me to perform a functionnal review of this PR (did dot read the code in details), focusing on the tile initialization on dezoom (related to #1223).
Overall, i think it is a nice simplification and it fixes issues we had for a while around elevation loading when performing deep zooms. Testing Lyon dataset in the 3dtiles_loader example, I saw real improvements, the terrain refines properly compared to master 🥳
I reviewed several examples I thought could be impacted, here is my review: 

### ColorLayer only
When testing a globe view with an OSM TMS Color Layer, I did'nt notice any major change. It feels the same, and it does not solve #1223.

### ColorLayer + elevation
When using the regular view_3d_map example with orthoimagery + elevation, behavior definitely changed. I suppose this is related to the merge of raster data handling between elevation and color tiles. I set up a little test zooming on mountains, tilting the view to the horizon and moving backward to see elevation tile loading behavior when dezooming.
* **Pros**: It looks way smoother on this PR than on master. No more "holes" created by very coarse terrain value init. from parent.
* **Cons**: Tiles loading is way slower than before. What is the reason? Is there a quick win to fix it? See videos bellow.

 #### Master

https://github.com/user-attachments/assets/e31d2995-fb13-4e92-a4d8-5932a6ca06d6

#### This PR

https://github.com/user-attachments/assets/af1f0b70-b14d-4b08-9720-00f98737d095

### Issues found on other examples :

* view_immersive: KO
* vector_tile_raster_3d: The ortho imagery flickers when loading at the beginning (I think I also saw this in other globe examples but it is more visible here. 
* vector_tile_mapbox_raster: Some labels are disappearing at the end of the example loading process.
* source_file_cog: When loading the RGB sample, I see errors in console, it does not load properly. (Try zoom in/out)

PR #2611, opened 2025-10-08
fix(camera): don't add default fog in views where camera has static near/far

Neptilo:
## Description
Only add a default fog in planar and globe views, not in the generic parent view. In that case, the camera's near/far properties aren't updated dynamically and neither is the fog, so the fog looks much too visible in some cases.

## Motivation and Context

In the Entwine example, a simple View is used, so the fog always keeps its default parameters (with `far` set to 1000), making the entire point cloud look grey.


mgermerie (approved):
Thanks for the quick fix.

PR #2612, opened 2025-10-09
Removal of the layer count restriction

gchoqueux:
- See proposal  (#2614 )  

**WARNING : There’s still unresolved bugs:**
* tiles disappear because of an improperly initialized bbox.
* Opacity, vibility and effects haven’t been reconnected yet.

## Description

Replace all textures array by just 2 textures, **map and displacement map**. All array textures are flatten in these maps.  

## Motivation and Context

* removal of the layer count restriction
* code simplification
* first step : 
    * to replace `LayeredMaterial`  with any Three.js materials
    * to support any raster re-projection


Desplandis:
Could you open a proposal for this PR? Describing what problem you want to resolve and explain your solution. =)

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -94,86 +94,35 @@ const rtCache = new LRUCache<string, THREE.WebGLArrayRenderTarget>({
...
+            return;
+        }
+
+        renderTarget = new THREE.WebGLRenderTarget(256, 256, { depthBuffer: false });
+
+        drawMap(renderTarget, tiles, renderer, extent);
+
+        rtCache.set(textureSetId, renderTarget);
+    } else {
+        drawMap(renderTarget, tiles, renderer, extent);
```
Neptilo: If we're caching the render targets, what is the use of redrawing their textures every time we access them? Am I missing something?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -94,86 +94,35 @@ const rtCache = new LRUCache<string, THREE.WebGLArrayRenderTarget>({
...
+    if (!renderTarget) {
+        if (tiles.length == 0) {
+            return;
+        }
+
+        renderTarget = new THREE.WebGLRenderTarget(256, 256, { depthBuffer: false });
+
+        drawMap(renderTarget, tiles, renderer, extent);
+
+        rtCache.set(textureSetId, renderTarget);
```
Neptilo: I was caching array render targets instead of array textures because it was impossible to extract the textures from the RTs and manage their memory separately, but maybe now that you're using simple render targets, this becomes possible and we can just maintain a cache of textures directly, and only use one RT for everything?

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -1,106 +1,46 @@
...
-            // render this source texture into the current layer
-            renderer.setRenderTarget(renderTarget, currentLayerIndex);
-            renderer.render(quad, quadCam);
         }
     }
+
     renderer.setRenderTarget(previousRenderTarget);
+    renderer.setClearAlpha(a);
 
     return renderTarget;
```
Neptilo: The returned value is not used anymore and the documented type is now incorrect

packages/Main/src/Renderer/LayeredMaterial.ts
gchoqueux: Yes, it's temporary. I'll add a mechanism to update the texture and parent texture when the opacity changes.

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -94,86 +94,35 @@ const rtCache = new LRUCache<string, THREE.WebGLArrayRenderTarget>({
  * @param type - Layer set identifier: 'c' for color, 'e' for elevation.
  * @param renderer - The renderer used to render textures.
```
Neptilo, line 56: To be updated

packages/Main/src/Renderer/ProjectionMaterials.js
```diff
@@ -0,0 +1,110 @@
...
+
+        varying vec2 vUv;
+
+        uniform sampler2D map;
+        uniform vec2 camera_footprint;
+        uniform vec2 texture_footprint;
+        uniform vec4 offsetScale;
+        uniform float opacity;
+
+        const float CTOYL = 20037508.34 / 3.141592653589793 * 0.69314718;
```
Neptilo: Could you comment what these numbers represent?

packages/Main/src/Renderer/ProjectionMaterials.js
```diff
@@ -0,0 +1,110 @@
...
+        }
+    `,
+};
+
+export const materialMercatorToWGS84 = new THREE.ShaderMaterial({
+    transparent: true,
+    uniforms: {
+        // This uniform will be updated with each source 2D texture
+        map: { value: null },
+        texture_footprint: { value: new THREE.Vector2() },
```
Neptilo: Could you comment what these properties represent?

packages/Main/src/Renderer/LayeredMaterial.ts
gchoqueux: I'm not sure, There might be a problem with the framebuffer linked to the render target
To try

packages/Main/src/Renderer/WebGLComposer.ts
gchoqueux: yes

gchoqueux: it's draft PR, it's still evolving, no need for an in-depth review right now

packages/Main/src/Renderer/ProjectionMaterials.js
gchoqueux: i'll add some comment
* 20037508.34 : max projection in speudo mercator
* 0,69... : Ln(2)

packages/Main/src/Renderer/ProjectionMaterials.js
gchoqueux: Footprint means the area/extent covered by something on the ground.
* `map`: map to project on the ground
* `texture_footprint `: texture coverage on ground
* `camera_footprint `: camera coverage visible on ground
* `offsetScale`: the map  portion to be projected

Desplandis:
@Neptilo @HoloTheDrunk Could you take a look to this PR this week?

packages/Main/src/Converter/convertToTile.ts
```diff
@@ -89,8 +89,27 @@ export default {
             if (parent) {
                 tile.geoidHeight = parent.geoidHeight;
                 const geoidHeight = geoidLayerIsVisible(layer) ? tile.geoidHeight : 0;
-                tile.setBBoxZ({ min: parent.obb.z.min, max: parent.obb.z.max, geoidHeight });
                 tile.material.setUniform('geoidHeight', geoidHeight);
+
+                const parent_uniforms = parent.material.uniforms;
+                const child_uniforms = tile.material.uniforms;
```
Neptilo, line 95: Use camel-case to keep coding style consistent.

packages/Main/src/Converter/convertToTile.ts
```diff
@@ -89,8 +89,27 @@ export default {
...
+
+                if (parent.material.uniforms.map.value) {
```
Neptilo, line 97: You can use the previously defined `parent_uniforms` here.

packages/Main/src/Converter/convertToTile.ts
```diff
@@ -89,8 +89,27 @@ export default {
...
+                    const extentParent = parent_uniforms.map.value.extent;
+                    child_uniforms.map.value = parent_uniforms.map.value;
+                    child_uniforms.colorOffsetScales.value = tile.extent.offsetToParent(extentParent);
```
Neptilo, line 100: Minor: fix warnings for line lengths over 100 characters

packages/Main/src/Converter/convertToTile.ts
```diff
@@ -89,8 +89,27 @@ export default {
...
+                if (parent_uniforms.displacementMap.value) {
+                    const extentParent = parent_uniforms.displacementMap.value.extent.toExtent(tile.extent.crs);
+                    child_uniforms.displacementMap.value = parent_uniforms.displacementMap.value;
+                    child_uniforms.elevationOffsetScales.value = tile.extent.offsetToParent(extentParent);
+                    const rasterElevationNode = parent_uniforms.elevationLayer.value;
+                    child_uniforms.elevationLayer.value = rasterElevationNode;
+                    tile.setBBoxZ({ min: rasterElevationNode.min, max: rasterElevationNode.max, scale: rasterElevationNode.layer.scale, geoidHeight });
+                } else {
+                    tile.setBBoxZ({ min: parent.obb.z.min, max: parent.obb.z.max, geoidHeight });
+                }
```
Neptilo, line 112: An explanatory comment for all this added code would be nice.

packages/Main/src/Renderer/Shader/TileFS.glsl
```diff
@@ -31,30 +31,17 @@ void main() {
...
-        #pragma unroll_loop
-        for ( int i = 0; i < NUM_CRS; i ++) {
-            color = getOutlineColor( outlineColors[ i ], uvs[ i ].xy);
-            gl_FragColor.rgb = mix(gl_FragColor.rgb, color.rgb, color.a);
-        }
+        color = getOutlineColor( outlineColors[ 0 ], vUv.xy);
+        gl_FragColor.rgb = mix(gl_FragColor.rgb, color.rgb, color.a);
     }
-  #endif
+    #endif
```
Neptilo, line 44: Minor: make sure to keep indentation consistent

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -90,88 +90,38 @@ const defaultStructLayers: Readonly<{
...
-                return;
-            }
-        }
+    extent: Extent) {
+    if (tiles.find(tile => tile.textures.length == 0)) {
+        return;
     }
 
-    const cachedRT = renderTargetCache?.get(textureSetId);
+    const textureSetId = extent.toString();
```
Neptilo, line 103: If the cache ids are only based on the extents, will we have a problem if there are multiple views? Different textures can be used for the same extent but in different views, e.g. a globe + a minimap.

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -482,45 +422,28 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
         }> {
         return {
             layers: this.uniforms[`${type}Layers`],
-            textures: this.uniforms[`${type}Textures`],
+            textures: type == 'color' ? this.uniforms.map : this.uniforms.dislacementMap,
```
Neptilo, line 425: "dislacementMap": typo.
Could we modify the type of `uniforms` to prevent this kind of mistake? (I'm not sure myself - may be overkill)

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -482,45 +422,28 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
     }
 
-    /**
-     * Updates uniforms for both color and elevation layers.
-     * Orchestrates the processing of visible color layers and elevation tiles.
-     *
-     * @param renderer - The renderer used to render textures into arrays.
-     */
-    public updateLayersUniforms(renderer: THREE.WebGLRenderer): void {
+    public updateLayersUniforms(renderer: THREE.WebGLRenderer, extent: Extent): void {
```
Neptilo, line 431: It's always good to keep function documentation.

packages/Main/src/Renderer/ProjectionMaterials.js
Neptilo: What do you mean by "max projection"?

packages/Main/src/Renderer/ProjectionMaterials.js
```diff
@@ -0,0 +1,110 @@
...
+        uniform vec2 camera_footprint;
+        uniform vec2 texture_footprint;
+        uniform vec4 offsetScale;
+        uniform float opacity;
+
+        const float CTOYL = 20037508.34 / 3.141592653589793 * 0.69314718;
+
+        float toY(float lat) {
+            float s = sin(lat);
+            return 0.5 * log2((1.0 + s) / (1.0 - s)) * CTOYL;
```
Neptilo, line 29: Where does this formula come from?

packages/Main/src/Renderer/ProjectionMaterials.js
Neptilo, line 1: You could use this opportunity to already write this new file in TypeScript.

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -1,117 +1,52 @@
...
-            fragmentShader: copyTextureShader.fragmentShader,
-        });
-        quad = new THREE.Mesh(geometry, material);
-    }
+materials.set(forwarding(), materialUnit);
+materials.set(forwarding('EPSG:3857', 'EPSG:4326'), materialMercatorToWGS84);
+
+const quad = new THREE.Mesh(geometry, materialUnit);
 
+export function drawMap(
```
Neptilo, line 17: Can you document the function?

packages/Main/src/Renderer/WebGLComposer.ts
```diff
@@ -1,117 +1,52 @@
...
-): THREE.WebGLArrayRenderTarget | null {
-    if (count === 0) { return null; }
+const materials = new Map();
 
-    // The render target's internal framebuffer will be used to attach layers.
-    const renderTarget = new THREE.WebGLArrayRenderTarget(width, height, count, {
-        depthBuffer: false, // No depth buffer needed for simple 2D texture copy
-    });
-    const arrayTexture = renderTarget.texture;
+const forwarding = (from, to) => ((from == to) ? 'transformUnit' : `${from} => ${to}`);
```
Neptilo, line 10: Can you document what `materials` contains and give `forwarding` a more explicit name? For example, "makeProjectionId" if you think that makes sense.

packages/Main/src/Renderer/ProjectionMaterials.js
Neptilo, line 1: Could you document what the purpose of these materials is? As I understand it: running projection computations on the GPU?

Neptilo (changes requested):

packages/Main/src/Converter/convertToTile.ts
gchoqueux: Would you add camel-case in eslint rules ?

Neptilo: I think it would be good to do it eventually, but for now there would be too many variables to rename (111). In a separate PR, why not.

PR #2613, opened 2025-10-10
fix(TileLayer): fix material opacity

Desplandis:
## Description

Since #2602, it is impossible to set the opacity of the tile layer (thanks @ketourneau for the issue). This PR fixes it.

I would like to backport this fix, push it in a v2.36 branch and manually release it in npm as 2.36.1. WDYT?


ketourneau:
We try to update itowns from version 2.45.2-next.10 to version 2.46.0 and we see a breaking issue.
It's introduced by this [commit](https://github.com/iTowns/itowns/commit/90159bddb007e281928a668d8692d5dab6c54c1d).
We display point clouds with map opacity to 0 to hide blue default map but now (without this PR) set opacity on layer doesn't do anything : map_opacity

AnthonyGlt (approved):
Oh well spotted, thank you @Desplandis & @ketourneau

AnthonyGlt:
Ok to backport imo

PR #2622, opened 2025-10-31
fix(PlanarView): Add option in planarView to prevent dynamic near/far…

ketourneau:
## Description

Since this commit : https://github.com/iTowns/itowns/commit/56b722b2288bd907141dcb6162fbe80d1f937967
PlanarView now adjust automatically near and far of camera.

It break our viewer, we already compute near and far of camera outside itowns.
I would therefore like to add a option `dynamicCameraNearFar`(default true) to allow/block this behaviour.

### DIFF
--- a/packages/Main/src/Core/Prefab/PlanarView.js
+++ b/packages/Main/src/Core/Prefab/PlanarView.js
@@ -25,6 +25,7 @@ class PlanarView extends View {
      * @param {CameraUtils~CameraTransformOptions|Extent} [options.placement] - The
      * {@link CameraUtils~CameraTransformOptions} to apply to view's camera or the extent it must display at
      * initialization. By default, camera will display the view's extent (given in `extent` parameter).
+     * @param {boolean} [options.dynamicCameraNearFar=true] - The camera's near and far are automatically adjusted.
      * @param {number} [options.farFactor=20] - Controls how far the camera can see.
      * The maximum view distance is this factor times the camera’s altitude (above sea level).
      * @param {number} [options.fogSpread=0.5] - Proportion of the visible depth range that contains fog.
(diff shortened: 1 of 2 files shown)


packages/Main/src/Core/Prefab/PlanarView.js
Neptilo, line 1: Seems ok to me, but to make the feature complete, the `GlobeView` should also be able to benefit from this option.

packages/Main/src/Core/Prefab/PlanarView.js
```diff
@@ -20,6 +20,7 @@ class PlanarView extends View {
      * @param {Extent} extent - The ground extent.
      * @param {Object} [options] - See options of {@link View}.
      * @param {boolean} [options.noControls=false] - If true, no controls are associated to the view.
+     * @param {boolean} [options.dynamicCameraNearFar=true] - The camera's near and far are automatically adjusted.
```
Neptilo: Just a detail: the doc would be better organized if this option appeared right before `farFactor` and `fogSpread` which are related to dynamic near/far.

Neptilo (changes requested):

packages/Main/src/Core/Prefab/PlanarView.js
ketourneau: Ok I added option to GlobeView

packages/Main/src/Core/Prefab/PlanarView.js
ketourneau: Ok I move the option before `farFactor` and `fogSpread`

Neptilo:
One last thing: defining a default value in the function documentation doesn't actually affect the code. You'd need to add something like:
```js
if (options.dynamicCameraNearFar === undefined) {
    options.dynamicCameraNearFar = true;
}
```
for example, at the same place where the default values for `farFactor` and `fogSpread` are defined, or change the existing condition to `if (options.dynamicCameraNearFar || options.dynamicCameraNearFar === undefined)`

ketourneau:
Oh yes indeed, I added default value !

Neptilo:
> Oh yes indeed, I added default value !

Well we're almost there! If you're defining an object property, the same property has to be checked in the condition (not the one from `options`. ;)

ketourneau:
Ok my bad i edit check to use `if (options.dynamicCameraNearFar || options.dynamicCameraNearFar === undefined)`

Neptilo (approved):

Issue #2625, opened 2025-11-03
Geotiff with PlanarView bug


ketourneau:
## Context

Since https://github.com/iTowns/itowns/pull/2558 we see a bug with PlanarView + COG Layer.

We have now error `Texture dimensions mismatch`.

Live sample: https://geotiff-bug.netlify.app/examples/#view_25d_map
Code: https://github.com/bloc-in-bloc/itowns/blob/f/geotiff-bug/examples/view_25d_map.html

PR #2626, opened 2025-11-03
Fix(CogParser): Fix tile raster dimensions

ketourneau:
Fix for https://github.com/iTowns/itowns/issues/2625

Tile raster dimensions is normally 256x256, not 255x255.

### DIFF
--- a/packages/Main/src/Parser/CogParser.ts
+++ b/packages/Main/src/Parser/CogParser.ts
@@ -4,6 +4,7 @@ import {
     Vector2,
 } from 'three';
 
+import { TEXTURE_TILE_DIM } from 'Provider/Fetcher';
 import { GeotiffNode } from './GeotiffParser';
 
 import type { TextureWithExtent } from './GeotiffParser';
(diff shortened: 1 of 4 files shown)


mgermerie (commented):
Hi and hanks for spotting this mistake!

I originally planned to make this dimension a Source parameter but forgot to do so. However, the example you added in your issue (thanks for that BTW) shed light on the fact COG and Geotiff parsed textures need to be the same size as the other textures displayed with other layers.

I will update your commit to use the constant defined [here](https://github.com/iTowns/itowns/blob/c6f59647a5d429bc3205e30b1b03a7156e6b9fb7/packages/Main/src/Provider/Fetcher.js#L4) for all other texture dimensions, and set it in in `GeotiffParser` as well.

mgermerie:
I could not succeed in pushing on your branch. Perhaps the `Allow edits and access to secrets by maintainers` button was un-ticked when creating your PR? Anyways, could you pleade rebase on [this branch](https://github.com/mgermerie/itowns/tree/fix/geotiff-texture-dimension)? It includes your commit updated to use the `TEXTURE_TILE_DIM` constant both in `CogParser` and `GeotiffParser` and another commit that fixes imports in those files.

ketourneau:
Sorry for forgetting to enable the option. I rebased my branch to your branch.

ftoromanoff (approved):

PR #2635, opened 2025-11-18
fix: Potree2 and Brotli compression

ftoromanoff:
The PR #2498 added the reprojection for point cloud data, but the Potree2BrotliLoader was missed.

This PR correct that oversight.


jailln (approved):

ketourneau:
Would it be possible to add a test for Potree2 and Brotli compression, to prevent regression for future improvements ?

ftoromanoff:
> Would it be possible to add a test for Potree2 and Brotli compression, to prevent regression for future improvements ?

In my opinion, this should be done as a functional test (to see if on a data set the result is conform to what we expect, by a screen shot for example). But it's currently not implemented...

PR #2636, opened 2025-11-19
fix(test): vpc test time out

ftoromanoff:
The units test for vpc layer are currently pretty often timed out...
This PR aims to chnage them to avoid these error linked to the network.

### DIFF
--- a/packages/Main/test/unit/vpc.js
+++ b/packages/Main/test/unit/vpc.js
@@ -4,17 +4,45 @@ import GlobeView from 'Core/Prefab/GlobeView';
 import { Coordinates, CRS } from '@itowns/geographic';
 import VpcSource from 'Source/VpcSource';
 import VpcLayer from 'Layer/VpcLayer';
+import sinon from 'sinon';
+import Fetcher from 'Provider/Fetcher';
 import Renderer from './bootstrap';
 
-const vpcEptUrl = 'https://data.geopf.fr/chunk/telechargement/download/lidarhd_fxx_ept/vpc/index.vpc';
-const vpcCopcUrl = 'https://data.geopf.fr/chunk/telechargement/download/LiDARHD-NUALID/VPC/amiens.vpc';
+import eptStack from '../data/vpc/eptStack.json';
+import copcStack from '../data/vpc/copcStack.json';
+
+import eptFile from '../data/entwine/ept.json';
+import eptHierarchyFile from '../data/entwine/ept-hierarchy/0-0-0-0.json';
+
+// url of the vpc stack
+const baseurl = 'https://data.geopf.fr/chunk/telechargement/download';
+const vpcEptUrl = `${baseurl}/lidarhd_fxx_ept/vpc/index.vpc`;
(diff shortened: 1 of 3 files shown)


packages/Main/test/unit/vpc.js
```diff
@@ -4,8 +4,24 @@ import GlobeView from 'Core/Prefab/GlobeView';
...
+const baseurl = 'https://data.geopf.fr/chunk/telechargement/download';
+const resources = {
+    [`${baseurl}/lidarhd_fxx_ept/vpc/index.vpc`]: JSON.parse(eptStack),
+    [`${baseurl}/LiDARHD-NUALID/VPC/amiens.vpc`]: JSON.parse(copcStack),
+    [`${baseurl}/lidarhd_fxx_ept/kg/ept.json`]: JSON.parse(eptFile),
+    [`${baseurl}/lidarhd_fxx_ept/kg/ept-hierarchy/0-0-0-0.json`]: JSON.parse(eptHierarchyFile),
+};
+
 const vpcEptUrl = 'https://data.geopf.fr/chunk/telechargement/download/lidarhd_fxx_ept/vpc/index.vpc';
 const vpcCopcUrl = 'https://data.geopf.fr/chunk/telechargement/download/LiDARHD-NUALID/VPC/amiens.vpc';
```
jailln: you could define these two urls above and use them in `resources`

packages/Main/test/unit/vpc.js
```diff
@@ -48,30 +74,30 @@ describe('VPC', function () {
...
-                    .then(() => {
-                        assert.ok(vpcCopcSource.sources[0].isCopcSource);
-                        done();
-                    }).catch(done);
-            });
+            copcMockSource.whenReady
+                .then(() => {
+                    assert.ok(vpcCopcSource.sources[0].isCopcSource);
+                    done();
+                }).catch(done);
```
jailln, line 102: I think the catched error may be lost with `catch(done)`. Shouldn't we do something like this instead ? :
```
.catch((err) => {
    console.error(err);
    done(err);
}); 
```

jailln (changes requested):

packages/Main/test/unit/vpc.js
ftoromanoff: If I'm not worng:
``
.catch(done);
``
is the same as 
``
.catch((err) => done(err));
``
what is the same as what you advice, without the console.error().

I tried to generate errors on purpose in:
- the vpc.js file
- vpSource.instantiate()
- in the assert.ok()
- in the copcSource.whenReady (for mockSource.whenReady)

And all errors are well catch by Mocha, what is in my opinion what we aims in the units tests.


But I did forgot a 'done' as argument of the it function for the 'instanciated copc stacked source'...
So all that tests were not for nothing ;-)

packages/Main/test/unit/vpc.js
ftoromanoff: Indeed ! 
Done

packages/Main/test/unit/vpc.js
jailln: :grin:  ok thanks for the tests and the feedback :+1:

jailln (approved):

PR #2637, opened 2025-11-21
Fix interferences between array texture generation and VR

Neptilo:
## Description
The array texture generation was interfering with the VR rendering, causing flashes in the view.
Before rending the textures, we already stored the previous render target to restore it afterwards.
This fix now does the same thing for the VR state and for the viewport.

## Motivation and Context
The entire view was flashing in VR, switching from the standard camera to the XR camera between frames, making VR unusable, as well as generating wrong textures.
After the first part of the fix, only storing and restoring the XR state, the view was no longer flashing but some tiles were blinking after a new array texture was rendered, as shown in the video.
Storing and restoring the GL viewport fixed that second part.

Tested in Chrome/Linux with the Immersive Web Emulator extension, and on a Pico headset.

## Screenshots
[Screencast from 20-11-2025 17:48:49.webm](https://github.com/user-attachments/assets/2d2f557e-5596-4004-99bf-f727ff04fc32)

### DIFF
--- a/packages/Main/src/Renderer/WebGLComposer.ts
+++ b/packages/Main/src/Renderer/WebGLComposer.ts
@@ -66,10 +66,16 @@ export function makeDataArrayRenderTarget(
         quad = new THREE.Mesh(geometry, material);
     }
 
+    // Store renderer state and temporarily disable VR
+    const previousRenderTarget = renderer.getRenderTarget();
+    const gl = renderer.getContext();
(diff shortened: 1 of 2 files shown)


Desplandis (approved):
I did not find a way to do it in pure `three.js`, I approve this PR pending that we open an issue on this subject.

Neptilo:
@Desplandis Do you think we can open a bug issue in Three.js as well?

AnthonyGlt (dismissed):

PR #2638, opened 2025-11-21
fix: load sub hierarchy in entwine scheme of data

ftoromanoff:
Since the PR#2553 REfactoring on PointCloudLayer, the entwine PointClooud layer wasn't loading past the first level of hierachy.

This PR fix this regression.


packages/Main/src/Core/PointCloudNode.js
```diff
@@ -198,7 +198,11 @@ class PointCloudNode extends THREE.EventDispatcher {
         childNode.clampOBB.matrixWorldInverse = this.clampOBB.matrixWorldInverse;
     }
 
-    load(networkOptions = this.source.networkOptions) {
+    async load(networkOptions = this.source.networkOptions) {
+        // Query octree/HRC if we don't have children yet.
+        if (!this.octreeIsLoaded) {
+            await this.loadOctree();
```
airnez, line 208: I understand you're factoring this code between potree and entwine nodes. Is there any impact on VPC ?
I don't really like calling a subclass method from the parent... you maybe should declare an empty loadOctree method in PointCloudNode, WDYT ?

ftoromanoff: Done !

ftoromanoff: As there are no vpcNode, this has no impact on VPC.

airnez: ok thank, you 👍

airnez (approved):
Looks good to me ! ✌️

PR #2639, opened 2025-11-21
fix(renderer): track render target usage before cleaning them up, to fix prematurely unloading textures

Neptilo:
## Description
Track which array render targets are actually used when rendering each view.
Then, when they would normally be disposed of because the cache managing them reaches its limit, we keep them in another map to only dispose of them once they are not used for rendering in any view.

## Motivation and Context
When rendering array textures for new tiles, some tiles' textures were sometimes prematurely unloaded, leaving holes in the terrain where those tiles should be.

## Screenshots
[Screencast from 19-11-2025 09:34:36.webm](https://github.com/user-attachments/assets/f5faa5fe-a578-44cc-8843-399536f993f0)

### DIFF
--- a/packages/Main/src/Core/MainLoop.js
+++ b/packages/Main/src/Core/MainLoop.js
@@ -1,4 +1,5 @@
 import { EventDispatcher } from 'three';
+import { cleanupRts, setCurrentRenderingView } from 'Renderer/LayeredMaterial';
 
 export const RENDERING_PAUSED = 0;
 export const RENDERING_SCHEDULED = 1;
@@ -227,6 +228,8 @@ class MainLoop extends EventDispatcher {
     }
 
     #renderView(view, dt) {
+        setCurrentRenderingView(view.id); // to track render target usage
+
         view.execFrameRequesters(MAIN_LOOP_EVENTS.BEFORE_RENDER, dt, this.#updateLoopRestarted);
 
         if (view.render) {
@@ -237,6 +240,8 @@ class MainLoop extends EventDispatcher {
         }
 
         view.execFrameRequesters(MAIN_LOOP_EVENTS.AFTER_RENDER, dt, this.#updateLoopRestarted);
+
+        cleanupRts();
     }
 }
 
--- a/packages/Main/src/Core/MainLoop.js
+++ b/packages/Main/src/Core/MainLoop.js
@@ -1,5 +1,4 @@
 import { EventDispatcher } from 'three';
-import { cleanupRts, setCurrentRenderingView } from 'Renderer/LayeredMaterial';
 
 export const RENDERING_PAUSED = 0;
 export const RENDERING_SCHEDULED = 1;
@@ -228,8 +227,6 @@ class MainLoop extends EventDispatcher {
     }
 
     #renderView(view, dt) {
-        setCurrentRenderingView(view.id); // to track render target usage
-
         view.execFrameRequesters(MAIN_LOOP_EVENTS.BEFORE_RENDER, dt, this.#updateLoopRestarted);
 
         if (view.render) {
@@ -240,8 +237,6 @@ class MainLoop extends EventDispatcher {
         }
 
         view.execFrameRequesters(MAIN_LOOP_EVENTS.AFTER_RENDER, dt, this.#updateLoopRestarted);
-
-        cleanupRts();
     }
 }
 
--- a/packages/Main/src/Core/TileMesh.ts
+++ b/packages/Main/src/Core/TileMesh.ts
@@ -146,6 +146,10 @@ class TileMesh extends THREE.Mesh<TileGeometry, LayeredMaterial> {
         if (this.material.layersNeedUpdate) {
             this.material.updateLayersUniforms(renderer);
         }
+
+        // Track actual usage every time this mesh is rendered
+        // Use global current rendering view ID set by MainLoop
+        this.material.trackCurrentRenderTargetUsage();
     }
 }
 
--- a/packages/Main/src/Core/TileMesh.ts
+++ b/packages/Main/src/Core/TileMesh.ts
@@ -149,7 +149,7 @@ class TileMesh extends THREE.Mesh<TileGeometry, LayeredMaterial> {
 
         // Track actual usage every time this mesh is rendered
         // Use global current rendering view ID set by MainLoop
-        this.material.trackCurrentRenderTargetUsage();
+        this.material.markAsRendered();
     }
 }
 
--- a/packages/Main/src/Layer/TiledGeometryLayer.js
+++ b/packages/Main/src/Layer/TiledGeometryLayer.js
@@ -360,6 +360,16 @@ class TiledGeometryLayer extends GeometryLayer {
         });
     }
 
+    /**
+     * All layer's 3D objects are removed from the scene and disposed from the video device.
+     * Also disposes the render target cache.
+     * @param {boolean} [clearCache=false] Whether to clear the layer cache or not
+     */
+    delete(clearCache) {
+        super.delete(clearCache);
+        this.renderTargetCache.dispose();
+    }
+
     // eslint-disable-next-line
     culling(node, camera) {
         return !camera.isBox3Visible(node.obb.box3D, node.matrixWorld);
--- a/packages/Main/src/Layer/TiledGeometryLayer.js
+++ b/packages/Main/src/Layer/TiledGeometryLayer.js
@@ -6,6 +6,7 @@ import convertToTile from 'Converter/convertToTile';
 import ObjectRemovalHelper from 'Process/ObjectRemovalHelper';
 import { getColorLayersIdOrderedBySequence } from 'Layer/ImageryLayers';
 import { CACHE_POLICIES } from 'Core/Scheduler/Cache';
+import { LRUCache } from 'lru-cache';
 
 const subdivisionVector = new THREE.Vector3();
 const boundingSphereCenter = new THREE.Vector3();
@@ -17,6 +18,9 @@ const boundingSphereCenter = new THREE.Vector3();
  * as it is used internally for optimisation.
  * @property {boolean} hideSkirt (default false) - Used to hide the skirt (tile borders).
  * Useful when the layer opacity < 1
+ * @property {Map<string, THREE.WebGLArrayRenderTarget>} pendingRtDisposal
+ * @property {Set<string>} usedRts
+ * @property {LRUCache<string, THREE.WebGLArrayRenderTarget>} rtCache
  *
  * @extends GeometryLayer
  */
@@ -148,6 +152,25 @@ class TiledGeometryLayer extends GeometryLayer {
          */
         this.diffuse = diffuse;
 
+        /**
+         * @type {Map<string, THREE.WebGLArrayRenderTarget>}
+         */
+        this.pendingRtDisposal = new Map();
+        /**
+         * @type {Set<string>}
+         */
+        this.usedRts = new Set();
+
+        /**
+         * @type {LRUCache<string, THREE.WebGLArrayRenderTarget>}
+         */
+        this.rtCache = new LRUCache({
+            max: 200,
+            dispose: (rt, key) => {
+                this.pendingRtDisposal.set(key, rt);
+            },
+        });
+
         this.level0Nodes = [];
         const promises = [];
 
@@ -251,6 +274,18 @@ class TiledGeometryLayer extends GeometryLayer {
         // In future, the sequence must be returned by parent geometry layer.
         this.colorLayersOrder = getColorLayersIdOrderedBySequence(context.colorLayers);
 
+        // Dispose render targets that are queued for disposal and not used by this view
+        if (this.usedRts.length) { // important: only clean up if last loop did rendering
+            for (const [id, renderTarget] of this.pendingRtDisposal) {
+                if (this.usedRts.has(id)) { continue; }
+                renderTarget.dispose();
+                this.pendingRtDisposal.delete(id);
+            }
+
+            // initialize render target usage tracking for next render
+            this.usedRts.clear();
+        }
+
         let commonAncestor;
         for (const source of sources.values()) {
             if (source.isCamera) {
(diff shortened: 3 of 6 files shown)


packages/Main/src/Core/MainLoop.js
```diff
@@ -227,6 +228,8 @@ class MainLoop extends EventDispatcher {
     }
 
     #renderView(view, dt) {
+        setCurrentRenderingView(view.id); // to track render target usage
```
Neptilo: Setting a global variable here to define a temporary rendering state may not be ideal. If you have better ideas, I'm interested.

Desplandis:
Could you give an example to reproduce the issue?

packages/Main/src/Core/MainLoop.js
Desplandis: I think we should cache per-layer and not introduce a global state

Desplandis (changes requested):
Hi and thanks for the PR!
I think the fix introduces overly complex global state handling and leaks implementation detail of `TiledGeometryLayer`, `TileMesh` and `LayeredMaterial` to `MainLoop`. I think the real issue is that the `rtCache` was made a global cache rather than a cache local to a given `TiledGeometryLayer`.

If the cache was scoped to the layer, you could use the same mechanism to dispose all render targets in `TiledGeometyLayer#postUpdate` (called after all updates).

Here a suggested approach that could fix your issue:
```ts
class TiledGeometryLayer {
  rtCache: LRUCache<string, THREE.WebGLArrayRenderTarget>;
  
  convert() {
    const tilemesh = convertToTile(...); // same as current code
    convertToTile.material.cache = rtCache;
    return tilemesh;
  }
}
```

It would suppose that we also have a property `cache` in `LayeredMaterial` which would be set by the layer (whenever you're initializing a new `TileMesh`). As such all material within a layer would share the same cache and the implementation would not leak into the `MainLoop`.

Let me know if you'd like further clarification!

Neptilo:
> Could you give an example to reproduce the issue?

The issue can be observed in basic examples like `view_3d_map.html`. You can also reduce the limit of `rtCache` in `LayeredMaterial` to increase the chances of reproducing it.

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -251,6 +274,18 @@ class TiledGeometryLayer extends GeometryLayer {
         // In future, the sequence must be returned by parent geometry layer.
         this.colorLayersOrder = getColorLayersIdOrderedBySequence(context.colorLayers);
 
+        // Dispose render targets that are queued for disposal and not used by this view
+        if (this.usedRts.length) { // important: only clean up if last loop did rendering
```
Neptilo: Previously this cleanup was done after each render, and this clear was done before each render. For it to work with the new refactoring (moving it to `preUpdate`), I had to add this condition, that looks a bit hacky but is there a better solution?

Neptilo:
@ftoromanoff This PR is ready for review again

packages/Main/src/Core/TileMesh.ts
```diff
@@ -146,6 +146,10 @@ class TileMesh extends THREE.Mesh<TileGeometry, LayeredMaterial> {
         if (this.material.layersNeedUpdate) {
             this.material.updateLayersUniforms(renderer);
         }
+
+        // Track actual usage every time this mesh is rendered
+        // Use global current rendering view ID set by MainLoop
+        this.material.trackCurrentRenderTargetUsage();
```
Desplandis: TODO: this should be done in afterRender and only called when needed

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -481,14 +389,121 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
+        }
+
+        const rt = makeDataArrayRenderTarget(width, height, count, tiles, max, renderer);
+        if (!rt) {
+            uTextureCount.value = 0;
+            return;
+        }
+
+        this.rtCache?.set(textureSetId, rt);
+        rt.texture.userData = { textureSetId };
```
Desplandis: I suppose this is marking this render target as used for this frame. Why not adding it directly to the set of used render targets (i.e. `usedRts`)?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -481,14 +389,121 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
+
+        let cachedRT = this.rtCache?.get(textureSetId);
+
+        if (!cachedRT) {
+            cachedRT = this.pendingRtDisposal?.get(textureSetId);
+            if (cachedRT) {
+                this.rtCache!.set(textureSetId, cachedRT);
+                this.pendingRtDisposal!.delete(textureSetId);
+            }
+        }
```
Desplandis: From what I infer from this code:
- it fetches a render target from the cache
- if not present, it checks whether it was marked as disposed and in such case cache it again

I think this code would benefit to be factored within a class (within a `get: string -> RenderTarget` method?). Moreover, if in cache it should be marked as used.

packages/Main/src/Renderer/LayeredMaterial.ts
Desplandis: Same here, this code would benefit to be factored within a class (within a `set: string -> RenderTarget -> void` method?).

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -481,14 +389,121 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
+     * including populating the DataArrayTexture
+     * with content from individual 2D textures on the GPU.
+     *
+     * @param uniforms - The uniforms object for your material.
+     * @param tiles - An array of RasterTile objects, each containing textures.
+     * @param max - The maximum number of layers for the DataArrayTexture.
+     * @param type - Layer set identifier: 'c' for color, 'e' for elevation.
+     * @param renderer - The renderer used to render textures.
+     */
+    private _updateLayersUniformsForType<Type extends 'c' | 'e'>(
```
Desplandis: Could you just add the corresponding cache parameter instead of moving the function to a new method, this breaks the diff of the code?

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -481,14 +389,121 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
+            }
+        }
+
+        let cachedRT = this.rtCache?.get(textureSetId);
+
+        if (!cachedRT) {
+            cachedRT = this.pendingRtDisposal?.get(textureSetId);
+            if (cachedRT) {
+                this.rtCache!.set(textureSetId, cachedRT);
+                this.pendingRtDisposal!.delete(textureSetId);
```
Desplandis: A good case for factoring the two caches & marked targets, you will not have to use the `!` operator!

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -251,6 +274,18 @@ class TiledGeometryLayer extends GeometryLayer {
...
+            for (const [id, renderTarget] of this.pendingRtDisposal) {
+                if (this.usedRts.has(id)) { continue; }
+                renderTarget.dispose();
+                this.pendingRtDisposal.delete(id);
+            }
+
+            // initialize render target usage tracking for next render
+            this.usedRts.clear();
+        }
+
```
Desplandis, line 263: A good case for refactoring into a class. A `cleanup` method maybe?

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -251,6 +274,18 @@ class TiledGeometryLayer extends GeometryLayer {
         // In future, the sequence must be returned by parent geometry layer.
         this.colorLayersOrder = getColorLayersIdOrderedBySequence(context.colorLayers);
 
+        // Dispose render targets that are queued for disposal and not used by this view
+        if (this.usedRts.size) { // important: only clean up if last loop did rendering
```
Desplandis: If done correctly, we should not need this no?

I know that we don't have a method called after rendering in the layer (maybe we should add a `postRender` method?) but this should clean the render target of the previous render.

packages/Main/src/Layer/TiledGeometryLayer.js
```diff
@@ -148,6 +152,25 @@ class TiledGeometryLayer extends GeometryLayer {
...
+
+        /**
+         * @type {LRUCache<string, THREE.WebGLArrayRenderTarget>}
+         */
+        this.rtCache = new LRUCache({
+            max: 200,
+            dispose: (rt, key) => {
+                this.pendingRtDisposal.set(key, rt);
+            },
+        });
```
Desplandis: I think you should have a dedicated class for all caches, I could think of a class like:
```ts
class RenderTargetCache {
    private cache: LRUCache<string, THREE.WebGLArrayRenderTarget>;
    private pendingDisposal = new Map<string, THREE.WebGLArrayRenderTarget>();
    private used = new Set<string>();
    
    constructor(maxSize = 200) {
        this.rtCache = new LRUCache({
            max: maxSize,
            dispose: (rt, key) => this.pendingDisposal.set(key, rt),
        });
    }
    
    get(id: string): THREE.WebGLArrayRenderTarget | undefined { ... }
    set(id: string, rt: THREE.WebGLArrayRenderTarget): void { ... }
    cleanup(): void { ... }
    dispose(): void { ... }
}
```

Desplandis (changes requested):
Thanks for you PR! However, I have a few remarks:
- I think you could simplify the code by building a dedicated cache class (= cache + disposal cache + used set)
- I don't think you need to guard against the first render
- You could directly marked render target as used, instead of the code of `trackCurrentRenderTargetUsage`

packages/Main/src/Renderer/RenderTargetCache.ts
```diff
@@ -0,0 +1,111 @@
...
+                this.pendingDisposal.set(key, rt);
+            },
+        });
+    }
+
+    /**
+     * Perform cleanup of render targets that are not used in this render cycle.
+     */
+    public cleanup(): void {
+        if (this.usedIds.size) { // important: only clean up if last loop did rendering
```
Neptilo: Copy of the previous conversation on this line:
Me: 
> Previously this cleanup was done after each render, and this clear was done before each render. For it to work with the new refactoring (moving it to preUpdate), I had to add this condition, that looks a bit hacky but is there a better solution?

Desplandis:
> If done correctly, we should not need this no?
> 
> I know that we don't have a method called after rendering in the layer (maybe we should add a postRender method?) but this should clean the render target of the previous render.

The problem is that we don't always do a `#render` after an `#update`. When we don't, if we remove that condition, no RT will be marked as used, and the next `cleanup` will dispose everything.

packages/Main/src/Renderer/LayeredMaterial.ts
Neptilo: Do you mean:
1. if in cache, it should already be marked as used, or
2. if in cache, we should mark it as used?

To be clear, the `usedIds` (previously `usedRts`) refer to the textures that are used for rendering in the current render cycle. "Marking as used" (done in `trackCurrentRenderTargetUsage`) means adding a texture to that set.
In the 1st case: not necessarily, because the cache is there to store textures that are no longer visible but are likely to become visible again soon.
In the 2nd case: `trackCurrentRenderTargetUsage` will mark them anyway right after, when called in `TileMesh.onBeforeRender`, so it would be redundant to already do it here. Moreover, `trackCurrentRenderTargetUsage` is always called in `TileMesh.onBeforeRender`, but `updateLayersUniforms` not always.

packages/Main/src/Renderer/LayeredMaterial.ts
Neptilo: This only assigns an ID to a texture. What marks the RT as used for this frame is `trackCurrentRenderTargetUsage`.
That function is called right after, in TileMesh.onBeforeRender. Moreover, it is always called there but updateLayersUniforms is only called if layers need to update.

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -487,36 +487,60 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
         }
 
         this.layersNeedUpdate = false;
     }
 
+    /**
+     * Track usage of current render targets for deferred disposal.
+     * Should be called every time this material is rendered.
+     */
+    public trackCurrentRenderTargetUsage(): void {
```
Desplandis: I think you should rename it as `markAsRendered`. I think it better reflects the intent of this code.

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -487,36 +487,60 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
+        const colorTextures = this.getUniform('colorTextures');
+        if (colorTextures?.userData?.textureSetId) {
+            this.renderTargetCache!.markAsUsed(colorTextures.userData.textureSetId);
```
Desplandis:
```suggestion
            this.renderTargetCache?.markAsUsed(colorTextures.userData.textureSetId);
```

packages/Main/src/Renderer/LayeredMaterial.ts
```diff
@@ -487,36 +487,60 @@ export class LayeredMaterial extends THREE.ShaderMaterial {
...
+        }
+
+        const elevationTextures = this.getUniform('elevationTextures');
+        if (elevationTextures?.userData?.textureSetId) {
+            this.renderTargetCache!.markAsUsed(elevationTextures.userData.textureSetId);
```
Desplandis:
```suggestion
            this.renderTargetCache?.markAsUsed(elevationTextures.userData.textureSetId);
```

packages/Main/src/Renderer/RenderTargetCache.ts
```diff
@@ -0,0 +1,93 @@
...
+                this._pendingDisposal.set(key, rt);
+            },
+        });
+    }
+
+    /**
+     * Perform cleanup of render targets that are not used in this render cycle.
+     */
+    public cleanup(): void {
+        if (this._usedIds.size) { // important: only clean up if last loop did rendering
```
Desplandis: Do an early return.

```suggestion
        if (!this._usedIds.size) { // important: only clean up if last loop did rendering
            return;
        }
```

Desplandis (changes requested):
Thanks for the changes, this is much more readable. =)
However, you don't dispose the `RenderTargetCache` when the layer itself is disposed!

This is the only blocking issue for merging, aside from a few non-blocking comments.

Desplandis (approved):
Looks good to me!

PR #2641, opened 2025-11-24
Refactoring Loader 

gchoqueux:
## Description

* Remove duplicated code related to  **`three.js`**
* Remove webpack `externals` from the worker to fix imports in **`three.js`**
* Update **`three.js`** to use the new `serialize` methods

| **bundle** |      **Master**       |  **this PR** |
|----------|:-------------:|------:|
| `itowns_lasworker.js` |  214KiB | 240 KiB |

### DIFF
--- a/packages/Main/src/Loader/LASLoader.js
+++ b/packages/Main/src/Loader/LASLoader.js
@@ -1,6 +1,18 @@
+// Don’t import Three directly if you want to improve tree shaking.
+import { Vector3, Quaternion, Box3 } from 'three';
 import { LazPerf } from 'laz-perf';
 import { Las } from 'copc';
 import proj4 from 'proj4';
+// eslint-disable-next-line import/extensions
+import { LASAttributes } from './LASConstant.js';
+
+const LASAttributesName = LASAttributes.filter(a => a.size === undefined).map(a => a.name);
+
+const origin = /* @__PURE__ */ new Vector3();
+const quaternion = /* @__PURE__ */ new Quaternion();
+const box = /* @__PURE__ */ new Box3();
+const position = /* @__PURE__ */ new Vector3();
+const scalarDepth16 = 1 / 256;
 
 /**
  * @typedef {Object} Header - Partial LAS header.
@@ -18,37 +30,6 @@ import proj4 from 'proj4';
 
 function defaultColorEncoding(header) {
     return (header.majorVersion === 1 && header.minorVersion <= 2) ? 8 : 16;
-}
-
-/**
- * Applies the given Quaternion to this vector.
- * @param {Array} v - The array depicting the position [x, y, z].
- * @param {Array} q - The Array depicting the Quaternion. [x, y, z, w]
- *
- * @return {Vector3} A reference to this vector.
- */
-function _applyQuaternion(v, q) {
-    // quaternion q is assumed to have unit length
-    const vx = v[0];
-    const vy = v[1];
-    const vz = v[2];
-    const qx = q[0];
-    const qy = q[1];
-    const qz = q[2];
-    const qw = q[3];
-
-    // t = 2 * cross( q.xyz, v );
-    const tx = 2 * (qy * vz - qz * vy);
-    const ty = 2 * (qz * vx - qx * vz);
-    const tz = 2 * (qx * vy - qy * vx);
-
-    const res = [];
-    // v + q.w * t + cross( q.xyz, t );
-    res[0] = vx + qw * tx + qy * tz - qz * ty;
-    res[1] = vy + qw * ty + qz * tx - qx * tz;
-    res[2] = vz + qw * tz + qx * ty - qy * tx;
-
-    return res;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


Desplandis:
Would it be possible to bump three.js in a different PR?

gchoqueux:
> Would it be possible to bump three.js in a different PR?

done #2663

packages/Main/src/Loader/LASLoader.js
```diff
@@ -1,6 +1,28 @@
 import { LazPerf } from 'laz-perf';
 import { Las } from 'copc';
 import proj4 from 'proj4';
+// Don’t import Three directly if you want to improve tree shaking.
+// eslint-disable-next-line import/extensions
+import { Vector3 } from 'three/src/math/Vector3.js';
+// eslint-disable-next-line import/extensions
+import { Quaternion } from 'three/src/math/Quaternion.js';
+// eslint-disable-next-line import/extensions
+import { Box3 } from 'three/src/math/Box3.js';
```
ftoromanoff: Is it really needed to do the import in 3 lignes?
I understand that it's better than to do 
```import * as THREE from three```
but can't we do something like that instead ?
```import { Vector3, Quaternion, Box3 } from three/src/math```

Comment (gchoqueux) [reply packages/Main/src/Loader/LASLoader.js]: yes, it' needed to import from specific files for the tree shacking

File: packages/Main/src/Parser/LASParser.js [hunk 2]
```diff
@@ -23,29 +23,29 @@ async function loader() {
...
 
-    if (attributes.color) {
-        const colorBuffer = new THREE.BufferAttribute(attributes.color, 4, true);
+    if (attributes.colors) {
+        const colorBuffer = new THREE.BufferAttribute(attributes.colors, 4, true);
         geometry.setAttribute('color', colorBuffer);
     }
-    const scanAngle = new THREE.BufferAttribute(attributes.scanAngle, 1);
+    const scanAngle = new THREE.BufferAttribute(attributes.ScanAngle, 1);
     geometry.setAttribute('scanAngle', scanAngle);
```
ftoromanoff: Is it really a good idea to rename the attributes with a capital letter (I uinderstand it comes from the name used by copc library (https://github.com/connormanning/copc.js/blob/master/src/las/extractor.ts)
but then maybe we need to use Position instead of positions and Color instead of colors, or maybe better use a map to associate the name inplural with the name with Capital letter (Array LASAttributes in LasLoader.js)

packages/Main/src/Loader/LASLoader.js
ftoromanoff: Ok, thus that what should be done in all the other itowns files then ?

packages/Main/src/Parser/LASParser.js
gchoqueux: I refactored this part to improve maintainability and future evolution

packages/Main/src/Loader/LASLoader.js
Desplandis: Normally, `import { Vector3, Quaternion, Box3 } from three` shall be sufficient for webpack to do tree-shaking.

packages/Main/src/Parser/LASParser.js
Desplandis: Could you explain the expected future gains?

webpack.config.cjs
```diff
@@ -153,9 +156,14 @@ module.exports = () => {
...
-            three: 'three',
-        },
+        externals: [
+            ({ context, request }, callback) => {
+                if (request === 'three' && !excludesToExternals.find(p => context.includes(p))) {
+                    return callback(null, 'three');
+                }
+                callback();
+            },
+        ],
```
Desplandis, line 166: Yikes... This is pretty hacky but I don't see another way to do it.
This is in my opinion a reason why the ES bundle in webpack is experimental. We should change the bundler one day!

packages/Main/src/Parser/LASParser.js
gchoqueux: to add other attributes 
https://github.com/connormanning/copc.js/blob/master/src/las/extractor.ts#L173

Desplandis: You convinced me but I would however keep the `position` and `color` as those are standard three attributes. We could even later add a filter to filter out some las attributes!

packages/Main/src/Loader/LASLoader.js
gchoqueux: I tested the bundle is biggest  `with import { Vector3, Quaternion, Box3 } from three`

packages/Main/src/Parser/LASParser.js
gchoqueux: i.e. ?
Color isn't necessarily defined in the laz format

Desplandis: Yeah but if it is indeed defined, I would prefer to use `color` instead (in a similar manner to https://github.com/mrdoob/three.js/blob/a58e9ecf225b50e4a28a934442e854878bc2a959/examples/jsm/loaders/GLTFLoader.js#L2301C1-L2312C3).

packages/Main/src/Loader/LASLoader.js
Desplandis: Weird, are you sure you tagged all potential side-effects as pure (`/* @__PURE__ */`)? We should always use imports directly from `three` as importing from source directory is fragile.

gchoqueux: I analyze the  bundle size with : 

#### 1 • files import
```js
import { Vector3 } from 'three/src/math/Vector3.js';
import { Quaternion } from 'three/src/math/Quaternion.js';
import { Box3 } from 'three/src/math/Box3.js';
```

#### 2 • three package import
```js
import { Vector3, Quaternion, Box3 } from 'three';
```



|  1 • files import |  2 • three package import  |
|---|---|
| novanilla| vanilla |
| 660 KB |  1,90 MB |

gchoqueux: @Desplandis I rool back to `import { Vector3, Quaternion, Box3 } from three`

Desplandis: Tested with `webpack serve` in production mode, I got for `itowns_lasworker.js`:
| **bundle** |      **Master**       |  **this PR** |
|----------|:-------------:|------:|
| Transfer |  83KiB | 107KiB |
| Size | 263kiB | 345KiB |

I think that development mode disables part of tree-shaking. This is however a +25% diff but this could be due to webpack... I think we should change our bundler and add a test suite with common bundlers (+ tutorials)

Desplandis (approved):
Looks good to me!

Tested with `webpack serve` in production mode, I got for `itowns_lasworker.js`:
| **bundle** |      **Master**       |  **this PR** |
|----------|:-------------:|------:|
| Transfer |  83KiB | 107KiB |
| Size | 263kiB | 345KiB |

I think that development mode disables part of tree-shaking. This is however a +25% diff but this could be due to webpack... I think we should change our bundler and add a test suite with common bundlers (+ tutorials)

packages/Main/src/Loader/LASLoader.js
gchoqueux: Which type of import did you try?

#### 1 • files import
```js
import { Vector3 } from 'three/src/math/Vector3.js';
import { Quaternion } from 'three/src/math/Quaternion.js';
import { Box3 } from 'three/src/math/Box3.js';
```
OR
#### 2 • three package import
```js
import { Vector3, Quaternion, Box3 } from 'three';
```

PR #2648, opened 2025-12-04
Dynamic style update for feature meshes

Neptilo:
Addresses #2649

## Description

Enables dynamic style updates for feature meshes. Refactors buffer logic into reusable functions and adds style versioning to track property changes.

## Motivation and Context

Allows real-time style updates (colors, heights) without rebuilding meshes, improving performance for interactive applications.

### DIFF
--- a/examples/source_stream_wfs_3d.html
+++ b/examples/source_stream_wfs_3d.html
@@ -7,8 +7,7 @@
             "imports": {
                 "itowns": "../dist/itowns.js",
                 "debug": "../dist/debug.js",
-                "dat": "https://unpkg.com/dat.gui@0.7.9/build/dat.gui.module.js",
-                "GuiTools": "./jsm/GUI/GuiTools.js",
+                "lil": "https://unpkg.com/lil-gui@0.19.2/dist/lil-gui.esm.min.js",
                 "LoadingScreen": "./jsm/GUI/LoadingScreen.js",
                 "itowns_widgets": "../dist/itowns_widgets.js",
                 "three": "https://unpkg.com/three@0.182.0/build/three.module.js",
@@ -16,6 +15,12 @@
             }
         }
         </script>
+
+        <style type="text/css">
+            .lil-gui {
+                left: 15px;
+            }
+        </style>
 
         <meta charset="UTF-8">
         <link rel="stylesheet" type="text/css" href="css/example.css">
@@ -32,7 +37,7 @@
         </div>
         <script type="module">
 
-            import GuiTools from 'GuiTools';
+            import lil from 'lil';
             import setupLoadingScreen from 'LoadingScreen';
             import * as THREE from 'three';
             import * as itowns from 'itowns';
@@ -67,7 +72,7 @@
             itowns.Fetcher.json('./layers/JSONLayers/Ortho.json').then(function _(config) {
                 config.source = new itowns.WMTSSource(config.source);
                 var layer = new itowns.ColorLayer('Ortho', config);
-                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+                view.addLayer(layer);
             });
 
             // Add two elevation layers.
@@ -75,7 +80,7 @@
             function addElevationLayerFromConfig(config) {
                 config.source = new itowns.WMTSSource(config.source);
                 var layer = new itowns.ElevationLayer(config.id, config);
-                view.addLayer(layer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
+                view.addLayer(layer);
             }
             itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
             itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
@@ -225,7 +230,7 @@
             });
             view.addLayer(wfsBuildingLayer);
 
-            var menuGlobe = new GuiTools('menuDiv', view);
+            var debugGui = new lil();
 
             // Add buttons to toggle extrusion height and color
             const extrusionToggle = {
@@ -240,11 +245,6 @@
                     view.notifyChange(wfsBuildingLayer);
                 }
             };
-            menuGlobe.gui.add(extrusionToggle, 'toggleHeight').name('Toggle Extrusion (Property/8)');
-            menuGlobe.gui.add(extrusionToggle, 'toggleColor').name('Toggle Color');
-
-            extrusionToggle.toggleHeight();
-            extrusionToggle.toggleColor();
 
             // Listen for globe full initialisation event
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
@@ -252,7 +252,7 @@
                 console.info('Globe initialized');
             });
 
-            debug.createTileDebugUI(menuGlobe.gui, view);
+            debug.createTileDebugUI(debugGui, view);
 
             function picking(event) {
                 if(view.controls.isPaused) {
@@ -285,24 +285,38 @@
             for (var layer of view.getLayers()) {
                 if (layer.id === 'WFS Bus lines') {
                     layer.whenReady.then( function _(layer) {
-                        var gui = debug.GeometryDebug.createGeometryDebugUI(menuGlobe.gui, view, layer);
+                        var gui = debug.GeometryDebug.createGeometryDebugUI(debugGui, view, layer);
                         debug.GeometryDebug.addMaterialLineWidth(gui, view, layer, 1, 10);
+                        gui.close();
                     });
                 }
                 if (layer.id === 'WFS Building') {
                     layer.whenReady.then( function _(layer) {
-                        var gui = debug.GeometryDebug.createGeometryDebugUI(menuGlobe.gui, view, layer);
+                        var gui = debug.GeometryDebug.createGeometryDebugUI(debugGui, view, layer);
                         debug.GeometryDebug.addWireFrameCheckbox(gui, view, layer);
+
+                        // Buttons to demonstrate dynamic style updates
+                        gui.add(extrusionToggle, 'toggleHeight').name('Toggle Extrusion (Property/8)');
+                        gui.add(extrusionToggle, 'toggleColor').name('Toggle Color');
+                        extrusionToggle.toggleHeight();
+                        extrusionToggle.toggleColor();
+
                         window.addEventListener('mousemove', picking, false);
+                        gui.close();
                     });
                 }
                 if (layer.id === 'WFS Route points') {
                     layer.whenReady.then( function _(layer) {
-                        var gui = debug.GeometryDebug.createGeometryDebugUI(menuGlobe.gui, view, layer);
+                        var gui = debug.GeometryDebug.createGeometryDebugUI(debugGui, view, layer);
                         debug.GeometryDebug.addMaterialSize(gui, view, layer, 1, 200);
+                        gui.close();
                     });
                 }
             }
+
+            // Close all GUI folders on start
+            debugGui.foldersRecursive().forEach(folder => folder.close());
+
             window.view = view;
             window.itowns = itowns;
             window.THREE = THREE;
--- a/examples/source_stream_wfs_3d.html
+++ b/examples/source_stream_wfs_3d.html
@@ -169,6 +169,9 @@
                 return properties.hauteur;
             }
 
+            var useFixedHeight = true;
+            var useFixedColor = true;
+
             function acceptFeature(properties) {
                 return !!properties.hauteur;
             }
@@ -218,18 +221,31 @@
                 filter: acceptFeature,
                 source: wfsBuildingSource,
                 zoom: { min: 14 },
-
-                style: {
-                    fill: {
-                        color: colorBuildings,
-                        base_altitude: altitudeBuildings,
-                        extrusion_height: extrudeBuildings,
-                    }
-                }
+                style: { fill: { base_altitude: altitudeBuildings } }
             });
             view.addLayer(wfsBuildingLayer);
 
             var menuGlobe = new GuiTools('menuDiv', view);
+
+            // Add buttons to toggle extrusion height and color
+            const extrusionToggle = {
+                toggleHeight: function () {
+                    useFixedHeight = !useFixedHeight;
+                    wfsBuildingLayer.style.fill.extrusion_height = useFixedHeight ? 8 : extrudeBuildings
+                    view.notifyChange(wfsBuildingLayer);
+                },
+                toggleColor: function () {
+                    useFixedColor = !useFixedColor;
+                    wfsBuildingLayer.style.fill.color = useFixedColor ? color.set(0x00cc00) : colorBuildings
+                    view.notifyChange(wfsBuildingLayer);
+                }
+            };
+            menuGlobe.gui.add(extrusionToggle, 'toggleHeight').name('Toggle Extrusion (Property/8)');
+            menuGlobe.gui.add(extrusionToggle, 'toggleColor').name('Toggle Color');
+
+            extrusionToggle.toggleHeight();
+            extrusionToggle.toggleColor();
+
             // Listen for globe full initialisation event
             view.addEventListener(itowns.GLOBE_VIEW_EVENTS.GLOBE_INITIALIZED, function () {
                 // eslint-disable-next-line no-console
--- a/examples/source_stream_wfs_3d.html
+++ b/examples/source_stream_wfs_3d.html
@@ -152,18 +152,18 @@
 
             function colorBuildings(properties) {
                 if (properties.usage_1 === 'Résidentiel') {
-                    return color.set(0xFDFDFF);
+                    return '#FDFDFF';
                 } else if (properties.usage_1 === 'Annexe') {
-                    return color.set(0xC6C5B9);
+                    return '#C6C5B9';
                 } else if (properties.usage_1 === 'Commercial et services') {
-                    return color.set(0x62929E);
+                    return '#62929E';
                 } else if (properties.usage_1 === 'Religieux') {
-                    return color.set(0x393D3F);
+                    return '#393D3F';
                 } else if (properties.usage_1 === 'Sportif') {
-                    return color.set(0x546A7B);
+                    return '#546A7B';
                 }
 
-                return color.set(0x555555);
+                return '#555555';
             }
 
             function altitudeBuildings(properties) {
@@ -241,7 +241,7 @@
                 },
                 toggleColor: function () {
                     useFixedColor = !useFixedColor;
-                    wfsBuildingLayer.style.fill.color = useFixedColor ? color.set(0x00cc00) : colorBuildings
+                    wfsBuildingLayer.style.fill.color = useFixedColor ? '#00cc00' : colorBuildings
                     view.notifyChange(wfsBuildingLayer);
                 }
             };
--- a/examples/source_stream_wfs_3d.html
+++ b/examples/source_stream_wfs_3d.html
@@ -236,12 +236,12 @@
             const extrusionToggle = {
                 toggleHeight: function () {
                     useFixedHeight = !useFixedHeight;
-                    wfsBuildingLayer.style.fill.extrusion_height = useFixedHeight ? 8 : extrudeBuildings
+                    wfsBuildingLayer.style.fill.extrusion_height = useFixedHeight ? 8 : extrudeBuildings;
                     view.notifyChange(wfsBuildingLayer);
                 },
                 toggleColor: function () {
                     useFixedColor = !useFixedColor;
-                    wfsBuildingLayer.style.fill.color = useFixedColor ? '#00cc00' : colorBuildings
+                    wfsBuildingLayer.style.fill.color = useFixedColor ? '#00cc00' : colorBuildings;
                     view.notifyChange(wfsBuildingLayer);
                 }
             };
@@ -295,7 +295,7 @@
                         var gui = debug.GeometryDebug.createGeometryDebugUI(debugGui, view, layer);
                         debug.GeometryDebug.addWireFrameCheckbox(gui, view, layer);
 
-                        // Buttons to demonstrate dynamic style updates
+                        // Buttons to demonstrate style changes after initialization
                         gui.add(extrusionToggle, 'toggleHeight').name('Toggle Extrusion (Property/8)');
                         gui.add(extrusionToggle, 'toggleColor').name('Toggle Color');
                         extrusionToggle.toggleHeight();
--- a/packages/Main/src/Converter/Feature2Mesh.js
+++ b/packages/Main/src/Converter/Feature2Mesh.js
@@ -925,38 +925,42 @@ export default {
                 console.error(`Trying to update style of unsupported feature type: ${feature.type}`);
         }
         let colors;
-        if (newColor && featureMesh.oldColor && !featureMesh.oldColor.equals(newColor)) {
-            featureMesh.oldColor = newColor;
-            colors = new Uint8Array(vertSize);
-            featureMesh.geometry.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
-        }
-        if (newExtrusionHeight !== null && featureMesh.oldExtrusionHeight !== newExtrusionHeight) {
-            featureMesh.oldExtrusionHeight = newExtrusionHeight;
-            posAttr = featureMesh.geometry.getAttribute('position');
-            for (const geometry of feature.geometries) {
-                context.setGeometry(geometry);
-                switch (feature.type) {
-                    case FEATURE_TYPES.LINE: {
-                        if (newExtrusionHeight !== null) {
-                            updateExtrudedLineVertices(featureMesh, options, posAttr?.array, colors);
-                        } else {
-                            console.warn('Updating unextruded line style not yet implemented');
-                        }
-                        break;
-                    }
-                    case FEATURE_TYPES.POLYGON: {
-                        if (newExtrusionHeight !== null) {
-                            updateExtrudedPolygonVertices(featureMesh, options, posAttr?.array, colors);
-                        } else {
-                            console.warn('Updating unextruded polygon style not yet implemented');
-                        }
-                        break;
-                    }
-                    default:
-                        console.warn('Updating point style not yet implemented');
-                }
+        if (newColor) {
+            if (featureMesh.oldColor && !featureMesh.oldColor.equals(newColor)) {
+                colors = new Uint8Array(vertSize);
+                featureMesh.geometry.setAttribute('color', new THREE.BufferAttribute(colors, 3, true));
             }
-            posAttr.needsUpdate = true;
+            featureMesh.oldColor = newColor;
+        }
+        if (newExtrusionHeight !== undefined) {
+            if (featureMesh.oldExtrusionHeight !== undefined && featureMesh.oldExtrusionHeight !== newExtrusionHeight) {
+                posAttr = featureMesh.geometry.getAttribute('position');
+                for (const geometry of feature.geometries) {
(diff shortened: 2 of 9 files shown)


ftoromanoff:
Thanks for the PR ! 

Would it be possible to add unit tests for the new behavior that you add here ? (Mostly for Feature2Mesh.js)
image

And maybe an example where we can play with this  ?

Neptilo:
@ftoromanoff @mgermerie PR ready again

jailln:
ping @ftoromanoff @mgermerie

ftoromanoff:
I don't know if it's linked with that PR, but when I play with the Layer WFS Building' opacity and that I tilt the view I get some strange results:
image
(This doesn't seem to appear in ortho view)

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -702,4 +871,69 @@ export default {
...
+            console.error('Cannot update style: collection not provided');
+            return;
+        }
+
+        context.setCollection(collection);
+        collection.updateMatrixWorld(true);
+        collection.matrixWorld.copy(collection.origMatrixWorld);
+
+        // only define attributes that need an update
+        featureMesh.stylePropVersions ||= {};
```
ftoromanoff: I don't really understand the use of the propertie 'stylePropVersions'.
If it's just to see if it need an update, why not just use a 'needUpdate' as boolean ?
Moreover, I fell like that if we switch way and back the toggling, we will create a new stylePropVersions each time even if it will be the same style... (kind of a memory leak?)

packages/Main/src/Core/Style.js
```diff
@@ -192,6 +193,10 @@ export class StyleContext {
         return this.#localCoordinates.setFromArray(vertices, offset);
     }
 
+    getGeometry() {
+        return this.#geometry;
+    }
+
```
mgermerie: Maybe a proper `get geometry` would be more alike what's done with `properties` getter. I know setters are not defined as so but we can try at least to have coherence in our getters.

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -514,24 +620,87 @@ function featureToExtrudedPolygon(feature, options) {
...
+ * @param {Object} featureMesh - Object carrying the feature (expects { feature }).
+ * @param {Object} buffers - Buffer management object.
+ * @param {Float32Array} [buffers.vertices] - Target positions buffer to write into.
+ * @param {Uint8Array} [buffers.colors] - Target color buffer (rgb Uint8, normalized).
+ * @param {Uint32Array} [buffers.batchIds] - Target per-vertex batch id buffer.
+ * @param {TypedArray} [buffers.indices] - Target index buffer to write line segment indices to.
+ * @param {number} buffers.indexPtr - Current write position in index buffer (incremented by function).
+ * @param {number} [id] - Batch id value to assign to written vertices when batchIds is provided.
+ */
+function updateLineBuffers(featureMesh, buffers, id) {
```
mgermerie: For more code navigability and consistency with your other updates, can you either define this method right after `featureToLine` or define all other `update{shape}Buffer` methods here along side this one ?

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -187,47 +195,24 @@ function featureToPoint(feature, options) {
     let featureId = 0;
     const vertices = new Float32Array(ptsIn);
     inverseScale.setFromMatrixScale(context.collection.matrixWorldInverse);
-    normal.set(0, 0, 1).multiply(inverseScale);
+    up.set(0, 0, 1).multiply(inverseScale);
 
     const pointMaterialSize = [];
     context.setFeature(feature);
```
mgermerie: You re-call this inside the `updatePointBuffer` method. I get the need of it inside `updatePointBuffer` method (which can be called by `updateStyle`) but it thus should be removed from here. Same goes for `context.setGeometry` on line 205 and in other feature type methods.

mgermerie (changes requested):
Thanks for the PR, and nice idea for the style properties versioning implementation!

A few comments on the code other than which there is an issue with line altitude: it seems you have an accidental increment somewhere, making each line appear higher than the previous one. You can see this behavior on this screenshot:

Image

mgermerie:
> I don't know if it's linked with that PR, but when I play with the Layer WFS Building' opacity and that I tilt the view I get some strange results: image (This doesn't seem to appear in ortho view)

From my personal testing, I did not have this issue, buildings opacity was correctly computed. I tried on the [WFS to 3D objects](https://www.itowns-project.org/itowns/examples/source_stream_wfs_3d.html) example, with custom or identical color for each buildings, before and after toggling their extrusion height. Maybe you found this in another test condition @ftoromanoff ?

Neptilo:
> A few comments on the code other than which there is an issue with line altitude: it seems you have an accidental increment somewhere, making each line appear higher than the previous one.

Found it! I had oversimplified by getting `base_altitude`  before the for loop, but actually, the call to `setLocalCoordinatesFromArray` inside the loop could modify the context, which was then used in `base_altitude`'s getter.

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: Good catch. For `context.setGeometry`, I don't see the redundancy. All the `update*Buffers` seem to access `context.geometry` and therefore need the geometry to be defined before.

Neptilo:
> I don't know if it's linked with that PR, but when I play with the Layer WFS Building' opacity and that I tilt the view I get some strange results

I can´t reproduce either. In the image it looks like the terrain also has transparency.

ftoromanoff:
> > I don't know if it's linked with that PR, but when I play with the Layer WFS Building' opacity and that I tilt the view I get some strange results
> 
> I can´t reproduce either. In the image it looks like the terrain also has transparency.

I tested again and don't have it anymore.... Might have been a wrong version on my side...
I will just advice, for a better understanding in the example to have the 'Toggle Extrusion' and 'Toggle Color' after the Layer WFS Building (or in the folder), you can also migrate that exemple to lil-gui as I did for the PointCloud examples (way back^^)

packages/Main/src/Converter/Feature2Mesh.js
mgermerie: Reading the code again, there is indeed no issue with `context.geometry`. What initially concerned me is that `context.geometry` is always set by the method calling `update{Shape}Buffers`, but is only used within `update{Shape}Buffers`. This could lead to side effects if `update{Shape}Buffers` is called improperly (i.e., without setting `context.geometry` first). However, I don’t see a clean way to avoid this without looping through each geometry inside `update{Style}Method`, which wouldn’t be ideal.

ftoromanoff (commented):
Sorry, it seems that i forgot to submit this commentary...

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: It was a long time ago so I'll do my best to remember...
If we use a simple boolean, the idea would be that we reset it as soon as the new style property has been "consumed" by an object, but in this case we have many objects consuming the style change, so we don't know when it is finished.
The next idea was to compare the values between the source style and the target objects, and only update them if they have changed, but we have the problem that the style value may be defined as a callback, but the getter doesn't return that function but tries to call it.
Hence the next idea of storing a property version number. Note that this is a common technique also used in Three internals.
If we keep toggling the switch, the property version will increase each time indeed, but since it is just a number, it will not use more memory.

ftoromanoff: I feel it strange to add a 'style' property in featureMesh when there is already a style object exactly for that..

Neptilo: Are you still referring to `stylePropVersions`? That only contains numbers, not styles per se. Would you like to discuss this orally so we can move forward?

examples/source_stream_wfs_3d.html
```diff
@@ -218,25 +226,33 @@
...
+                    wfsBuildingLayer.style.fill.extrusion_height = useFixedHeight ? 8 : extrudeBuildings
+                    view.notifyChange(wfsBuildingLayer);
+                },
+                toggleColor: function () {
+                    useFixedColor = !useFixedColor;
+                    wfsBuildingLayer.style.fill.color = useFixedColor ? color.set(0x00cc00) : colorBuildings
+                    view.notifyChange(wfsBuildingLayer);
+                }
+            };
+
```
gchoqueux, line 248: Can you use the callbacks (`colorBuildings, extrudeBuildings`) already present in the example to specify colors and height based on properties and context?

packages/Main/src/Layer/ReferencingLayerProperties.js
```diff
@@ -4,18 +4,21 @@ function ReferLayerProperties(material, layer) {
     if (layer && layer.isGeometryLayer) {
         material.layer = layer;
 
+        const getOpacity = (opacity) => {
+            const styleOpacity = material.layer.style?.fill?.opacity ?? 1;
+            const layerOpacity = material.layer.opacity;
+            return styleOpacity * layerOpacity * (opacity ?? 1);
+        };
```
gchoqueux, line 11: `opacity` and `styleOpacity` are different ?

gchoqueux (dismissed):
## Architecture

### Pattern
A robust way possible is use  [Model–view–controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) to apply dynamic style applies a style to data.
Could you reinforce this pattern within your development?

### Update Style 

When the building height modification is applied, it seems to me that all buildings are rebuilt: reconstruction of the base and roof vertices, face computation, and repopulation of the colors. Is that correct?  If so, what would be the difference compared to removing the `featureMesh` and recreating it ?

####  Proposed solutions

1. Create a controller/updater to modify only the relevant attribute
2.  Link `batchId` attributes to style buffer attributes. You would only need to modify the attributes. wdyt ?
3. use [BatchedMesh](https://threejs.org/docs/?q=batch#BatchedMesh) see [example ](https://threejs.org/examples/?q=bat#webgl_batch_lod_bvh)

## Example

Could add a color picker to specify the color building ?
Could you add a slider to change height building in real  time ? 
Is it possible to have a style property that changes over time — that is, one that can be animated?

## Issue

There's a issue with one color mode, it's not just the color green

Image

packages/Main/src/Process/FeatureProcessing.js
```diff
@@ -25,6 +26,11 @@ export default {
                 f.layer.object3d.add(f);
                 f.meshes.position.z = geoidLayerIsVisible(layer.parent) ? node.geoidHeight : 0;
                 f.meshes.updateMatrixWorld();
+
+                // update existing features
+                for (const featureMesh of f.meshes.children) {
+                    Feature2Mesh.updateStyle(featureMesh, f.collection, layer.style);
```
ftoromanoff: Why not a featureMesh.applyStyle(), as this don't update a style but a featureMesh.

The style actually if i recall corectly, fixed a value (for each style property) at start between the value entered by the user (the layer.value) then the value from the source, then the default value. Maybe we just need an other level, or a way to have the userValue modified, As i can see ther is a set property in defineStyleProperty than can have a function. Can you have a function that use the value from your toggle button ?

examples/source_stream_wfs_3d.html
Neptilo: Not sure what you mean. Those callbacks are already used.

packages/Main/src/Layer/ReferencingLayerProperties.js
Neptilo: Updated. They are different but they are never both defined at the same time.

examples/source_stream_wfs_3d.html
gchoqueux: Could you modify  the callbacks `colorBuildings` and  `extrudeBuildings`  to specify by example : 

```js
function colorBuildings(properties) {
    if (toogle) { 
         return '#00ff00';
    }
    if (properties.usage_1 === 'Résidentiel') {
        return '#FDFDFF';
    } else if (properties.usage_1 === 'Annexe') {
        return '#C6C5B9';
    } else if (properties.usage_1 === 'Commercial et services') {
        return '#62929E';
    } else if (properties.usage_1 === 'Religieux') {
        return '#393D3F';
    } else if (properties.usage_1 === 'Sportif') {
        return '#546A7B';
    }

    return '#555555';
}
```

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: After discussion, we decided to trigger style updates from the setter of the style properties (done by the iTowns user) instead of doing it on the layer updates, using an event mechanism instead of a property version mechanism.

examples/source_stream_wfs_3d.html
Neptilo: In that case, that would technically be the same style, not a style change, and the toggle wouldn't trigger an update.

packages/Main/src/Process/FeatureProcessing.js
Neptilo: I think the second comment is the same as https://github.com/iTowns/itowns/pull/2648#discussion_r2864344681. See the discussions there.

examples/source_stream_wfs_3d.html
ftoromanoff: But do we really need a 'style change' ?

packages/Main/src/Process/FeatureProcessing.js
ftoromanoff: Indeed, it's linked.

examples/source_stream_wfs_3d.html
Neptilo: As I said, without it, the toggle wouldn't trigger an update, so the objects would keep the same appearance as before. And we can´t reasonably defer the reponsibility of calling `style.dispatchEvent({ type: 'style-property-changed', style, parameter });` to the user. Even with a wrapper around it, the user would need to pass the style and changed parameter name as arguments (to keep the benefit of not updating all buffers).
And even if we implemented such a user-called update signal, as a user I would still exepect `style.property = ...` to work out-of-the-box. So the current example is useful to at least demonstrate that this technique works.

examples/source_stream_wfs_3d.html
```diff
@@ -269,24 +285,38 @@
...
+                        gui.close();
                     });
                 }
                 if (layer.id === 'WFS Building') {
                     layer.whenReady.then( function _(layer) {
-                        var gui = debug.GeometryDebug.createGeometryDebugUI(menuGlobe.gui, view, layer);
+                        var gui = debug.GeometryDebug.createGeometryDebugUI(debugGui, view, layer);
                         debug.GeometryDebug.addWireFrameCheckbox(gui, view, layer);
+
+                        // Buttons to demonstrate dynamic style updates
```
ftoromanoff: As it's not really a 'dynamic style' update, I would suggest to change the commentary to 'Buttons to demonstrate a dynamic update of style.' or 'Buttons to demonstrate  the feasibilty of a change of style after initialization.'

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -13,7 +13,7 @@ let style;
 
 const dim_ref = new THREE.Vector2();
 const dim = new THREE.Vector2();
-const normal = new THREE.Vector3();
+const up = new THREE.Vector3();
```
ftoromanoff, line 16: Maybe we should go back to 'normal' or find an other expliciter name ?

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -38,17 +38,25 @@ class FeatureMesh extends THREE.Group {
         this.#collection = new THREE.Group().add(this.meshes);
         this.#collection.quaternion.copy(collection.quaternion);
         this.#collection.position.copy(collection.position);
-
         this.#collection.scale.copy(collection.scale);
         this.#collection.updateMatrix();
+        this.#collection.matrixWorldInverse = this.#collection.matrixWorld.clone().invert();
+        this.#collection.origMatrixWorld = collection.matrixWorld.clone();
```
ftoromanoff: Why do you need to save the matrixWorld ?

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -38,17 +38,25 @@ class FeatureMesh extends THREE.Group {
...
+
+        this.#collection.crs = collection.crs;
 
         this.#originalCrs = collection.crs;
         this.#currentCrs = this.#originalCrs;
         this.extent = collection.extent;
+        this.isFeatureMesh = true;
```
ftoromanoff, line 50: Do you need that property ?

ftoromanoff (dismissed):

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -698,8 +860,78 @@ export default {
...
+function _applyStyle(featureMesh, collection, parameter) {
+    const feature = featureMesh.feature;
+    if (!collection) {
+        console.error('Cannot update style: collection not provided');
+        return;
+    }
+
+    context.setCollection(collection);
+    collection.updateMatrixWorld(true);
+    collection.matrixWorld.copy(collection.origMatrixWorld);
```
ftoromanoff: You update the MatrixWorld to overwrite the next line. I feel there is somethings weird here...

examples/source_stream_wfs_3d.html
```diff
@@ -218,25 +226,33 @@
...
             view.addLayer(wfsBuildingLayer);
 
-            var menuGlobe = new GuiTools('menuDiv', view);
+            var debugGui = new lil();
+
+            // Add buttons to toggle extrusion height and color
+            const extrusionToggle = {
+                toggleHeight: function () {
+                    useFixedHeight = !useFixedHeight;
+                    wfsBuildingLayer.style.fill.extrusion_height = useFixedHeight ? 8 : extrudeBuildings
```
mgermerie: Missed it on the first review, but could you just add the `;` at the end of this instruction as well as on line 244 ?

mgermerie (dismissed):
Thanks for the architecture updates on the PR!

The current approach to updating styles does not account for changes in line width, which I missed during my initial review. Was this an oversight, or is it out of scope for this PR?

If it’s the latter, could you please:
- add a console warning to inform users that changing `style.stroke.width` is not yet supported?
- open an issue to track support for this feature?

packages/Main/src/Converter/Feature2Mesh.js
ftoromanoff: up is indeed a property on THREE.object3D that "Defines the up direction of the 3D object which influences the orientation".
But is it the same property here ? just to be sure we don't get confused...

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: For information, the refactoring works for **WFS** datasets but broke the style updates for **MVT** dataset (e.g. `vector_tile_3d_mesh`). I've tried debugging it but haven't been able to find a solution so far.

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: This is indeed the idea I had in mind.
In practice, `up` is mostly used in Three for objects on which it is meaningful to define the action of "looking at" something: cameras, lights, or even sprites (that must keep facing the camera).
Here, `up` is used as the direction of extrusion, so not the same use case, but it does represent the same vector: the `up` of the extruded polygons represents the same vector as the `up` of the camera.

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: To reapply it here: https://github.com/iTowns/itowns/pull/2648#discussion_r2980118683.
I can't really explain it, but without it the result was incorrect (before the event refactoring - and after the refactoring it is more broken).

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: It is useful when debugging, to check if an object is of this type. It was already done in some classes (and everywhere within Three's codebase) so it seemed reasonable to keep it here too.

packages/Main/src/Converter/Feature2Mesh.js
Neptilo: See https://github.com/iTowns/itowns/pull/2648#discussion_r2996411147.
The results were incorrect if we removed one of those lines. A subtlety about `updateMatrixWorld` is that it also updates its children's matrices.

jailln:
@mgermerie @ftoromanoff  I took a look into the two remaining topics of this PR:

1. `collection` `matrixWorld` that was saved in `origMatrixWorld` in `convert` and reused in `applyStyle`. 

It was related to an issue that was already present but not visible before: We used `collection.matrixWorld` instead of `collection.matrix` - which was not an issue in `convert` since they were the same but is an issue in `applyStyle` since the collection has been added to the scene at that point. This is fixed in [the following commit](https://github.com/iTowns/itowns/pull/2648/changes/2e9b80bd268bd1a09fd87fa5fc3c33c3760d9a2d)

2. The need to store a styleVersion in each Layer node (instead of having a layer.needsUpdate property on the layer only). 

When not using style versions in each layer node, there was a bug where some tiles didn't get the correct style modifications. I found out that this was the case for cached tiles. When the user triggered a style modification, all displayed tiles were correctly updated and and then layer.needsUpdate was reset to false (but cached tiles style was not updated). Then, when a cached tile was retrieved to be displayed, it contained the wrong style and did not went through `applyStyle` because layer.needsUpdate was already false. Applying style updates to all cached tiles whenever a user changes a style property would be a waste of time since cached tiles may not ever be displayed. However I tried an implementation without the layer node style versioning system (see [this commit](https://github.com/iTowns/itowns/commit/047d097f615503d71da52ae43c2d8a1e9e3281b2)). The main idea was to apply the style modification to tiles in the scheduler callback ([here](https://github.com/iTowns/itowns/commit/047d097f615503d71da52ae43c2d8a1e9e3281b2#diff-073c93060804b9a6a67da1185aeb54d5f9613ae80abe7e84834563e2d7c688efR89-R100)). However, this also applies it to newly loaded tiles, whether they are retrieved from the cache or not, which is not optimized at all (newly loaded tiles buffers are always computed twice). Therefore, we think that the best implementation is to have a style versionning mechanism on tiles to know if they need to be updated or not to correctly account for cached tiles, but to also use events to mark layers that needs update (which is an optimization comparing to the first proposed implementation). This is implemented in the last commits pushed on the branch of this PR.

We are available to explain this orally since it may be difficult to follow in written.

examples/source_stream_wfs_3d.html
```diff
@@ -218,25 +226,33 @@
...
+            // Add buttons to toggle extrusion height and color
+            const extrusionToggle = {
+                toggleHeight: function () {
+                    useFixedHeight = !useFixedHeight;
+                    wfsBuildingLayer.style.fill.extrusion_height = useFixedHeight ? 8 : extrudeBuildings;
+                    view.notifyChange(wfsBuildingLayer);
+                },
+                toggleColor: function () {
+                    useFixedColor = !useFixedColor;
+                    wfsBuildingLayer.style.fill.color = useFixedColor ? '#00cc00' : colorBuildings;
```
Desplandis, line 244: Optionally, I think this example would better demonstrate the style-changing feature if the buildings were coloured based on a different property. This is typically what users expect and would help clarify that the proposed functionality is more of a style change than a dynamically updated style.

packages/Main/src/Core/Style.js
```diff
@@ -138,6 +138,7 @@ function defineStyleProperty(style, category, parameter, userValue, defaultValue
             },
             set: (v) => {
                 property = v;
+                style.dispatchEvent({ type: 'style-property-changed', style, parameter });
```
Desplandis, line 141: This code limits where the style change can be applied no?

I can trigger the style change through property set. For example:
```js
wfsBuildingLayer.style.fill.color = useFixedColor ? '#00cc00' : colorBuildings;

```

But the above code limits certain valid use-case for the end-user:
- Setting the `fill` property directly. The above could be rewritten as:
   ```js
   wfsBuildingLayer.style.fill = {
     ...wfsBuildingLayer.style.fill,
     color: useFixedColor ? '#00cc00' : colorBuildings,
   };
   ```
- Setting the style directly. The above could be rewritten as:
   ```js
   wfsBuildingLayer.style= {
       ...wfsBuildingLayer.style,
       fill: {
           ...wfsBuildingLayer.style.fill,
           color: useFixedColor ? '#00cc00' : colorBuildings,
       }
   };
   ```

How do we support such valid use-cases?

packages/Main/src/Converter/Feature2Mesh.js
```diff
@@ -698,8 +859,90 @@ export default {
...
+                this._stylePositionVersion = this._stylePositionVersion ?? 0;
+                style.addEventListener('style-property-changed', (event) => {
+                    if (event.parameter === 'color') {
+                        this._styleColorVersion++;
+                    }
+                    if (event.parameter === 'extrusion_height' || event.parameter === 'base_altitude') {
+                        this._stylePositionVersion++;
+                    }
+                });
+            }
```
Desplandis, line 879: I think this code should be in the layer (i.e. where `Feature2Mesh.convert` is called). Moreover, I think we could more easily track the color and position changes using a `Map` or `Set` of difference between the old and new style in the layer.

packages/Main/src/Process/FeatureProcessing.js
```diff
@@ -25,6 +26,23 @@ export default {
                 f.layer.object3d.add(f);
                 f.meshes.position.z = geoidLayerIsVisible(layer.parent) ? node.geoidHeight : 0;
                 f.meshes.updateMatrixWorld();
+                const updateColor = f.styleColorVersion !== (layer._styleColorVersion ?? 0);
+                const updatePosition = f.stylePositionVersion !== (layer._stylePositionVersion ?? 0);
```
Desplandis, line 30: A simple flag or a set that tracks which `FeatureMesh` instances needs updating should be sufficient. The purely event-based approach could not work here without tracking since we don't always revisit the nodes that requires updates (culling, ...).

Desplandis (changes requested):
From my understanding, this PR has two main goals:
- First, a refactoring of the `Feature2Mesh` module into smaller, composable components which I fully welcome. I didn't notice any issues here except in my opinion a lack of tests (was already the case on master)
- Second, a style property change feature and not a dynamic style per se (if you could update the PR title accordingly). I have a few concerns, primarily regarding the style-change API, which currently does not support some expected usage patterns (see the dedicated comment).

Regarding the change notification process, I understand why and event-based approach alone isn't sufficient, since buffer updates of a loaded feature mesh only occurs when it is updated by the layer (which may not occur if it was culled for this frame). However, a simpler mechanism (such as tracking nodes pending updates within the layer) might be easier to reason about, maintain and test. I don't see the need for versioning when a simple flag or tracked node set could do the job.

examples/source_stream_wfs_3d.html
jailln: agreed but can be improved later as discussed orally

packages/Main/src/Core/Style.js
jailln: Done in the last commit, although it does not work with the spread operator (see my comment in Style.js, which will also be in the jsdoc). If it raises to much discussion I propose that we move this into a new issue / PR opened right away so we can finally move on this 5 month old PR.

packages/Main/src/Converter/Feature2Mesh.js
jailln: can be improved later as agreed orally. We thought of it but it wasn't that easy to implement without many modifications

packages/Main/src/Process/FeatureProcessing.js
jailln: it seems that it was necessary to store versions but I don't totally recollect why. Again we can improve this afterwards as agreed orally.

Desplandis (approved):

PR #2650, opened 2025-12-08
Fix loading issues with potree2

Desplandis:
## Description
Potree 2 builds its request using information from the loaded hierarchy. However, #2638 introduced a regression where request building was done before the hierarchy was loaded, introducing invalid values for the request. This PR fixes this regression.

I did not add a new test since I'm currently doing a refactor of all pointcloud related modules.


ftoromanoff:
I propose here a simplier way to fix it, more in adequation with the refactoring of all pointCloudNode (in my opinion)
https://github.com/iTowns/itowns/commit/c7dc6b36ef40fc74347a614c788f4fc865d8b876

jailln (approved):

PR #2651, opened 2025-12-08
Migrate pointcloud-related modules to typescript

Desplandis:
## Description

This PR migrates most of pointcloud-related modules (i.e. nodes, sources and layers) to typescript. It also set what is the public API of both pointcloud sources and layers. This a first step towards a refactoring of all pointclouds.

### DIFF
--- a/packages/Main/src/Core/Potree2Node.ts
+++ b/packages/Main/src/Core/Potree2Node.ts
@@ -33,9 +33,8 @@ of the authors and should not be interpreted as representing official policies,
     either expressed or implied, of the FreeBSD Project.
  */
 
-import PointCloudNode from 'Core/PointCloudNode';
 import type Potree2Source from 'Source/Potree2Source';
-import { computeChildBBox } from 'Core/PotreeNode';
+import { PotreeNodeBase } from 'Core/PotreeNode';
 
 const NODE_TYPE = {
     NORMAL: 0,
@@ -45,13 +44,8 @@ const NODE_TYPE = {
 
 type NodeType = typeof NODE_TYPE[keyof typeof NODE_TYPE];
 
-class Potree2Node extends PointCloudNode {
+class Potree2Node extends PotreeNodeBase {
     source: Potree2Source;
-
-    childrenBitField: number;
-    hierarchyKey: string;
-    baseurl: string;
-    crs: string;
 
     loaded: boolean;
     loading: boolean;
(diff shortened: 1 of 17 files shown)


packages/Main/src/Core/Potree2Node.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const NODE_TYPE = {
+    NORMAL: 0,
+    LEAF: 1,
+    PROXY: 2,
+} as const;
+
+type NodeType = typeof NODE_TYPE[keyof typeof NODE_TYPE];
+
+class Potree2Node extends PointCloudNode {
```
ftoromanoff: Why remove the extends from PotreeNode ?
I would advice creating a new abstract Potree0Node from where PotreeNode and Potree2Node are children as there is really a lot in commun...

Desplandis: Because `PotreeNode` is not a subset of `Potree2Node`. However, they share a common subset that I extracted as `PotreeBaseNode`.

Desplandis:
@ftoromanoff @jailln This PR is ready for review, there is still things that I can't type yet (which depends on either `Source`, `Layer` or `PointsMaterial`) but for the most part it is. I will let you choose which one of you review this PR. x)

P.S.: I will open another PR to write a better documentation since the addition of the documentation commits caused the moved files to be considered as new files when @ftoromanoff was rebasing his PR into my branch...

ftoromanoff:
I will propose this commit:
[0f7659ee8217e206f621ae8523837089e9060a41](https://github.com/ftoromanoff/itowns/pull/22/commits/0f7659ee8217e206f621ae8523837089e9060a41)
to keep the factorization of the load method

packages/Main/src/Core/Potree2Node.ts
ftoromanoff: thanks

ftoromanoff:
Then to simplified the typing there is properties that can be moved to constructor: (depth and index).
See commit 
https://github.com/ftoromanoff/itowns/pull/22/commits/28067a5a345999bc04b1ca8aab0cc81716415a75
and 
https://github.com/ftoromanoff/itowns/pull/22/commits/1c56aaab8c638155182a46ed65a7be8f53a3bbdd

ftoromanoff:
there is many warning on copcNode that should be adressed, or they are corrected on your documentation commit that are following ?

ftoromanoff:
reopened under #2657. To go further...

PR #2653, opened 2025-12-09
[Demo] added an end-user oriented demonstration of itowns capabilities

Nynjin:
## Description
This pull request introduces a new demonstration example for iTowns, adding a complete demo application under `examples/demo`, integrated amongst the other Itowns examples under the section `Demo`.

This demo section features multiple self-contained scenes with navigation, camera transitions, and layer management (visibility, controls, feature picking, recreation of view and layers from scratch, ...).

The `examples/demo` package can be described as such :
> `css/` : style for the UI elements
>
> `src/` : Typescript demo code
>
>> `Config/` : global constants such as scene transition duration or distance from globe where to toggle on/off atmosphere
>>
>> `Layers/` : objects containing promises for itowns layered shared between multiple scenes to avoid refetching/re-instantiation, as well as methods such as feature picking which is specific to each layer
>>
>> `Sources/` : promises containing sources that are used by multiple layers to avoid refetching
>>
>> `Views/` : singletons for the different views shared between layers (globe, immersive, planar, ...) that initialize itowns views and expose extra methods/properties (visibility toggle, viewerDiv, layer insertion, instance clearing, ...)
>>
>> `Scenes/` : scenes used in the demo containing their methods and properties (title, description, layers, onCreate, onEnter, onExit, view)
>>
>> `Repositories/` : shared singleton containers such as SceneRepository (array of all used scenes)
>>
>> `Types/` : common types and wrappers (SceneType, or LayerType containing all types of itowns layers used)
>>
>> `Utils/` : global methods such as scene transition, ui update, camera movement, event propagation block, ...
>>
>> `Services/` : application-level services such as feature picking on click
>>
>> `index.ts` : main demo entrypoint
>
> `index.html` : entrypoint for the demo. Hosts the root containers for the UI

## Motivation and Context
iTowns currently provides an unstructured collection of examples that are mainly designed for developers and are kind of tutorials in the end. This PR adds an end-user oriented demonstration of main iTowns features.
See [related issue](https://github.com/iTowns/itowns.github.io/issues/35).

### Environment

- OS: Windows 11
- Browser: Brave
- Node: v24.11.0
- npm: 11.6.1

## Screenshots

https://github.com/user-attachments/assets/d7fa247e-3531-4450-a7db-583abb61ca0e

| Globe Scene | Planar Scene | Terrain Elevation Scene |
| --- | --- | --- |
|  |  |  |

| Instanced Data Scene | Point Cloud Scene | BIM Scene |
| --- | --- | --- |
|  |  |  |

| Textured Meshes Scene | Immersive Scene |
| --- | --- |
|  |  |

## Open questions / TODO
* Find a better 3DTiles dataset
* Find a better point cloud (bigger, with classification and RGB if possible). Which dataset can we use from the LIDAR HD ? 
* Immersive and planar scene tend to be unstable (white screen sometimes, ground on immersive scene may not appear), mostly fixed with scene reset
* Add extruded lines when ready
* Move light on globe scene to not obscure the center
* Camera movement based on duration can be troublesome when switching between scenes that have the same placement, or if resetting camera position when it hasn't moved, or has barely moved (waiting 2 seconds each time and blocking the UI)


Desplandis:
I would not be able to do a review until new year's eve (vacation time :D), @alavenant @Calvina24 could you review the functionality of this PR and dispatch the technical part to a member of the team?

Desplandis:
@Nynjin I saw that there is a 10mB binary file in the changes, would it be possible to remove it from the commits and host it elsewhere to not pollute the main itowns repository? We usually use [itowns2-sample-data](https://github.com/iTowns/iTowns2-sample-data) for the datasets.

Nynjin:
> @Nynjin I saw that there is a 10mB binary file in the changes, would it be possible to remove it from the commits and host it elsewhere to not pollute the main itowns repository? We usually use [itowns2-sample-data](https://github.com/iTowns/iTowns2-sample-data) for the datasets.

Got it thanks, I uploaded my models there and made a [PR](https://github.com/iTowns/iTowns2-sample-data/pull/21).
Once that's done I'll update this one to remove the local models and use the uploaded ones instead.

alavenant:
Thanks for the work. 

The transition from one demo to another is smooth and enjoyable.

I suggest to put the "go local" next to "bring your data to life". 

In "Feel the Relief", you could also talk about the ortho-images.

"go local" : unzoom so that we can fully appreciate the scene as a whole

Why not add a view "combine your data" with lidar, 3D objects, buildings, bim, ... ?

Nynjin:
> Thanks for the work.
> 
> The transition from one demo to another is smooth and enjoyable.
> 
> I suggest to put the "go local" next to "bring your data to life".
> 
> In "Feel the Relief", you could also talk about the ortho-images.
> 
> "go local" : unzoom so that we can fully appreciate the scene as a whole
> 
> Why not add a view "combine your data" with lidar, 3D objects, buildings, bim, ... ?

Thank you for you feedback.

I did consider putting "go local" before "bring your data to life" but having a planar view right in the middle of globe views looks a bit off, that's mostly why planar and street views are at the very end.

And when it comes to combining the data, it's an interesting idea but I'd have to check how it holds up performance-wise since clicking directly on that scene would attempt to instantiate all of these elements at the same time.

examples/demo/index.html
```diff
@@ -0,0 +1,162 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Itowns - Demonstration</title>
```
mgermerie: A minor detail but maybe "Showcase" would be more suited than "Demonstration" ?

examples/demo/src/Scenes/Globe3dScene.ts
```diff
@@ -0,0 +1,33 @@
...
+    description: 'Visualize the Earth in 3D with high-resolution orthophoto imagery. '
+    + 'Navigate freely around the globe with intuitive controls.',
+    placement: {
+        coord: new itowns.Coordinates('EPSG:4326', 2.351323, 48.856712),
+        range: 25000000,
+        tilt: 89.5,
+        heading: 0,
+    },
+    layers: [],
+    view: new View3D(),
```
mgermerie: This should not be necessarily initialized with an instance of `View3D` since it is later overwritten within `onCreate` method (on line 23).

mgermerie: And the same goes for all other scenes.

examples/demo/src/Views/View.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    addLayers(layers: LayerType[]) {
+        if (!this.view) {
+            throw new Error(`view '${this.id}' is not defined`);
+        }
+
+        const layerPromises = layers.map(async (layer) => {
+            if (this.view!.getLayerById(layer.id)) {
```
mgermerie: The non-null assertion here should not be necessary since `this.view` was already checked on line 58.

mgermerie: The same goes for line 67.

examples/demo/src/Scenes/InstancedDataScene.ts
```diff
@@ -0,0 +1,79 @@
...
+                .notifyChange(InstancedDataScene.view.getView().camera3D, true);
+        }
+    },
+    onCreate: async () => {
+        if (InstancedDataScene.ready) {
+            return;
+        }
+        InstancedDataScene.view = new View3D();
+
+        const view = InstancedDataScene.view.getView();
```
mgermerie: For clarity, perhaps you could rename this constant to `itownsView`, for example. This would help distinguish between the `View3D` and `itowns.GlobeView` types.

examples/demo/src/Sources/index.ts
```diff
@@ -0,0 +1 @@
+export * as BuildingsSource from './BuildingsSource';
```
mgermerie, line 1: I understand that this data source is the only one used in multiple layers, which is probably why it’s the only one exposed in `Sources/`.

However, for better code consistency, I suggest declaring all sources in `Sources/`, even if they are used by only a single layer. In my opinion, this would serve two purposes:
- If a source needs updating, it’s more straightforward to find the line to change in the `Sources/` folder rather than inside a specific file in `Layers/`.
- The showcase might evolve and eventually display the same 3D tiles set in multiple layers across different scenes, for instance.

Then there is the issue of ortho imagery and DEM sources, whose parameters are obtained from fetching JSON files. This approach is used in the current examples, but I don’t think it should be replicated elsewhere - it only serves to make the example code shorter by removing lines that aren’t relevant to the feature being demonstrated.
In my opinion, it would be better to copy these source configurations into specific files in the `Sources/` folder, keeping the `demo` and `examples` packages self-contained.

examples/demo/src/Utils/SceneTransitionUtils.ts
```diff
@@ -0,0 +1,250 @@
...
+        for (const layer of nextScene.layers) {
+            layer.visible = true;
+            if (FeaturePickerService.layers.find(l => l.id === layer.id) &&
+            nextScene.view instanceof View3D) {
+                FeaturePickerService.enable(nextScene.view);
+                featurePickingInScene = true;
+            }
+        }
+
+        FeaturePickerService.container!.style.display =
```
mgermerie: On some of the transitions I tested, `FeaturePickerService.container!.style` was `null`, causing an error when reading `display`.

examples/demo/src/Views/View.ts
```diff
@@ -0,0 +1,77 @@
...
+        return this.viewerDiv;
+    }
+
+    setVisible(visible: boolean) {
+        if (!this.viewerDiv) {
+            throw new Error('viewerDiv is not defined');
+        }
+
+        this.viewerDiv.setAttribute('id', visible ? 'viewerDiv' : this.id);
+        this.viewerDiv.style.display = visible ? 'block' : 'none';
```
mgermerie, line 38: From my understanding, simply setting `display: none` on a three.js canvas container does not prevent animation frames from being requested. It therefore continues to use resources unnecessarily.

You could try:
```js
this.view.tileLayer.frozen = !visible;
this.view.notifyChange();
```
but I’m not sure whether there is a better way to prevent three.js from rendering. @Desplandis or @HoloTheDrunk, do you have any insight on this ?

webpack.config.cjs
```diff
@@ -158,8 +159,36 @@ module.exports = () => {
...
+                import: './examples/demo/src/index.ts',
+                dependOn: 'itowns',
+            },
+        },
+        output: {
+            ...configESM.output,
+        },
+        resolve: {
+            ...configESM.resolve,
+        },
```
mgermerie: As `output` and `resolve` are already properties of `configESM`, you should not need to overwrite them here. They should already be accounted for with the spread on line 163.

mgermerie (commented):
Thank you for this PR! I found it well documented, with clearly written and well-architected code, which made it pleasant to review.

I left some comments below. If I understand correctly, the UI of the demo will change. Will this be done in this PR or in another one?

examples/demo/index.html
Nynjin: Showcase does sound better, I renamed it

examples/demo/src/Views/View.ts
Nynjin: It happens because the non-null assertions was outside of the promise. And by the time the promise is executed, typescript considers it could be null again. I just moved the assertion inside the promise

examples/demo/src/Scenes/Globe3dScene.ts
Nynjin: You're right, I changed it to undefined until onCreate initializes it and I added getters to retrieve the view

examples/demo/src/Sources/index.ts
Nynjin: Got it, I separated source loading, 3D model loading and layers
In my opinion the JSON fetcher can still be interesting for code examples on using config files so for now I just copied the json configuration to be inside the demo in order to keep it self contained
But it could be removed to use the configuration directly in the source

examples/demo/src/Utils/SceneTransitionUtils.ts
Nynjin: I added an extra check to make sure FeaturePickerService container is defined before using it

webpack.config.cjs
Nynjin: You're right, I removed the redundant overwrite

Nynjin:
> > Thanks for the work.
> > The transition from one demo to another is smooth and enjoyable.
> > I suggest to put the "go local" next to "bring your data to life".
> > In "Feel the Relief", you could also talk about the ortho-images.
> > "go local" : unzoom so that we can fully appreciate the scene as a whole
> > Why not add a view "combine your data" with lidar, 3D objects, buildings, bim, ... ?
> 
> Thank you for you feedback.
> 
> I did consider putting "go local" before "bring your data to life" but having a planar view right in the middle of globe views looks a bit off, that's mostly why planar and street views are at the very end.
> 
> And when it comes to combining the data, it's an interesting idea but I'd have to check how it holds up performance-wise since clicking directly on that scene would attempt to instantiate all of these elements at the same time.

@alavenant I added a `Combine Your Data` scene with the data from parks, point cloud, buildings, 3D trees and BIM
It's pretty slow when using performance and network throttling but if you navigate the scenes one by one then all of the layers are already cached and it's nearly instant
I just used the data I already had to see how it'd look so the BIM model is out of place. If there's a proper dataset it could be interesting

image

examples/demo/src/Scenes/InstancedDataScene.ts
Nynjin: You're right it's clearer. I renamed the "getView" method to "getItownsView" as well

examples/demo/src/Views/View.ts
Desplandis: Since we lazily request animation frames, if we do not trigger update (no call to `view#notifyChange` & co) we should not update / render another frame!

jailln:
@mgermerie could you take a look at the last changes and comments of @Nynjin please ? I would like to converge soon on this one

examples/demo/src/Sources/index.ts
mgermerie: Thanks. Having the JSON config files as you suggest in your latest implementation is fine for me.

mgermerie (commented):
Thanks for the updates.

I just tested the demo again and found a small glitch: if you change the point cloud color option and then switch scenes, the UI for that option is not updated accordingly. For example, you can toggle the option to "Color" in `PointCloudScene` and then switch to `CombinedDataScene`. The UI will reset to "Classification", while the points themselves remain in "Color" mode.

Other than that, it looks all good to me.

Would you mind squashing the commits as well ? Since this PR introduce a single large feature (the showcase addition), it might be clearer from a changelog/history perspective to keep it as one commit. That said, it would make for quite a large commit, so I'd be interested in hearing your thoughs on this as well, @Desplandis and @jailln.

jailln:
> Thanks for the updates.
> 
> I just tested the demo again and found a small glitch: if you change the point cloud color option and then switch scenes, the UI for that option is not updated accordingly. For example, you can toggle the option to "Color" in `PointCloudScene` and then switch to `CombinedDataScene`. The UI will reset to "Classification", while the points themselves remain in "Color" mode.
> 
> Other than that, it looks all good to me.
> 
> Would you mind squashing the commits as well ? Since this PR introduce a single large feature (the showcase addition), it might be clearer from a changelog/history perspective to keep it as one commit. That said, it would make for quite a large commit, so I'd be interested in hearing your thoughs on this as well, @Desplandis and @jailln.

One commit is fine by me :+1:

Nynjin:
> Thanks for the updates.
> 
> I just tested the demo again and found a small glitch: if you change the point cloud color option and then switch scenes, the UI for that option is not updated accordingly. For example, you can toggle the option to "Color" in `PointCloudScene` and then switch to `CombinedDataScene`. The UI will reset to "Classification", while the points themselves remain in "Color" mode.
> 
> Other than that, it looks all good to me.
> 
> Would you mind squashing the commits as well ? Since this PR introduce a single large feature (the showcase addition), it might be clearer from a changelog/history perspective to keep it as one commit. That said, it would make for quite a large commit, so I'd be interested in hearing your thoughs on this as well, @Desplandis and @jailln.

Thank you for the review.

I fixed the UI to stay synced with the point cloud layer + merge conflicts as I just rebased.
I can squash the commits during the merge.

AnthonyGlt:
@tim-btrn

Nynjin:
Regarding the Point Cloud view, I've switched to the dataset used in `entwine-simple-loader` which isn't lidar-hd and has stray points in the sky, though it seems to look a little better as it is much larger and has colors. 
It would probably be a lot better with EDL but for now it looks like this :
image
image

jailln (approved):

mgermerie (approved):
Thanks for this much-needed addition! I understand that the UI changes you discussed with your designers will come later in another PR. Is that right ?

Nynjin:
> Thanks for this much-needed addition! I understand that the UI changes you discussed with your designers will come later in another PR. Is that right ?

That's right. @tim-btrn will be the one working on it.

PR #2654, opened 2025-12-10
[Pointcloud] Add per-node priority update in pointclouds

Desplandis:
## Description

This PR introduces a priority-based update mechanism for point cloud layers.

Before this PR, the update logic traversed the tree in a depth-first fashion, issued requests during traversal, and only afterward limited the number of point clouds rendered. This often resulted in unnecessary requests and complex update logic.

The new approach uses a priority queue (à la potree) to traverse only the most relevant nodes, sorted by their screen-space size.

Depends #2685.
Blocked by issue #2694 .

Benefits:
- Limit network requests up front
- Avoid traversing unimportant parts of the tree
- Simplify the overall update logic
- Improve performance and predictability

### DIFF
--- a/packages/Main/src/Layer/PointCloudLayer.ts
+++ b/packages/Main/src/Layer/PointCloudLayer.ts
@@ -418,53 +418,62 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
      *
      * @returns The child nodes to update or [] if there is none.
      */
-    update(context: Context, layer: this, elt: PointCloudNode): PointCloudNode[] {
-        if (this.octreeDepthLimit >= 0 && this.octreeDepthLimit < elt.depth) {
-            return [];
-        }
-
-        // get object on which to measure distance
-        let bbox;
-        let object3d;
-        if (elt.obj) {
-            object3d = elt.obj;
-            bbox = object3d.geometry.boundingBox as THREE.Box3;
-        } else {
-            object3d = elt.clampOBB;
-            bbox = object3d.box3D;
-            if (!object3d.parent) {
-                this.obbes.add(object3d);
-                object3d.updateMatrixWorld(true);
+    update(context: Context, layer: this, root: PointCloudNode): PointCloudNode[] {
+        const rootWithWeight = { node: root, weight: Infinity };
+        const queue = new TinyQueue([rootWithWeight], (a, b) => b.weight - a.weight);
+        let numVisiblePoints = 0;
+        while (queue.length > 0 && numVisiblePoints < this.pointBudget) {
+            const { node } = queue.pop() as { node: PointCloudNode };
+            // get object on which to measure distance
+            let bbox;
+            let object3d;
+            if (node.obj) {
+                object3d = node.obj;
+                bbox = object3d.geometry.boundingBox as THREE.Box3;
+            } else {
+                object3d = node.clampOBB;
+                bbox = object3d.box3D;
+                if (!object3d.parent) {
+                    this.obbes.add(object3d);
+                    object3d.updateMatrixWorld(true);
+                }
             }
-        }
 
-        const visible = context.camera.isBox3Visible(bbox, object3d.matrixWorld);
+            if (this.octreeDepthLimit >= 0 && this.octreeDepthLimit < node.depth) {
+                continue;
+            }
 
-        if (!visible) {
-            return [];
-        }
+            const visible = context.camera.isBox3Visible(bbox, object3d.matrixWorld);
 
-        point.copy(context.camera.camera3D.position)
-            .applyMatrix4(object3d.matrixWorldInverse as THREE.Matrix4);
+            if (!visible) {
+                continue;
+            }
 
-        const distanceToCamera = bbox.distanceToPoint(point);
+            numVisiblePoints += node.numPoints;
 
-        elt.sse = computeScreenSpaceError(
-            context,
-            layer.pointSize,
-            elt.pointSpacing,
-            distanceToCamera,
-        ) / this.sseThreshold;
+            point.copy(context.camera.camera3D.position)
+                .applyMatrix4(object3d.matrixWorldInverse as THREE.Matrix4);
 
-        // The visibility here is not definitive, as it will be updated
-        // in the postUpdate phase
-        elt.visible = visible;
-        this._candidateNodes.push(elt);
-        this.loadData(elt, context, layer, distanceToCamera);
+            const distanceToCamera = bbox.distanceToPoint(point);
 
-        if (elt.children && elt.children.length) {
-            if (elt.sse >= 1) {
-                return elt.children;
+            node.sse = computeScreenSpaceError(
+                context,
+                layer.pointSize,
+                node.pointSpacing,
+                distanceToCamera,
+            ) / this.sseThreshold;
+
+            node.visible = visible;
+            node.notVisibleSince = undefined;
+            this._candidateNodes.push(node);
+            this.loadData(node, context, layer, distanceToCamera);
+
+            if (node.children && node.children.length) {
+                if (node.sse >= 1) {
+                    for (const child of node.children) {
+                        queue.push({ node: child, weight: node.sse });
+                    }
+                }
             }
         }
 
@@ -472,17 +481,13 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
     }
 
     postUpdate() {
+        // TODO: Need to only change group meshes here + visibility event
         const visibleLastUpdate = this._visibleNodes;
         this._visibleNodes = new Set();
         this.displayedCount = 0;
 
-        // Push visible nodes until the point budget is reached
         while (this._candidateNodes.length > 0) {
             const node = this._candidateNodes.pop() as PointCloudNode;
-            if (this.displayedCount + node.numPoints > this.pointBudget) {
-                this._candidateNodes.push(node);
-                break;
-            }
             this.displayedCount += node.numPoints;
             this._visibleNodes.add(node);
             node.visible = true;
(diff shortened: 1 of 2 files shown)


packages/Main/src/Core/PointCloudNode.ts
```diff
@@ -181,8 +179,15 @@ abstract class PointCloudNode extends THREE.EventDispatcher {
         root.voxelOBB.quaternion.copy(rotation).invert();
 
         root.voxelOBB.updateMatrix();
+        // Update matrixWorld since OBB is not part of scene hierarchy
```
ftoromanoff: It is has there is this line
```
this.object3d.add(this.root.clampOBB);
this.root.clampOBB.updateMatrixWorld(true);
```
In all PointCloudLayer after we call setOBBes(). So I decided not to call it 2 times...

packages/Main/src/Renderer/OBB.ts
```diff
@@ -114,6 +114,79 @@ class OBB extends THREE.Object3D {
...
+        this.updateMatrixWorld(true);
+
+        const invMatrix = this.matrixWorldInverse;
+
+        const min = this.box3D.min;
+        const max = this.box3D.max;
+
+        const center = new THREE.Vector3()
+            .addVectors(min, max)
+            .multiplyScalar(0.5);
```
ftoromanoff:
```
const center = new THREE.Vector3();
this.box3D.getCenter(center);
```

packages/Main/src/Renderer/OBB.ts
```diff
@@ -114,6 +114,79 @@ class OBB extends THREE.Object3D {
...
+
+        const extents = new THREE.Vector3()
```
ftoromanoff: Is this really an extents ? I feel like it's more the halfed size of the Box3 then:
```
const size  = new THREE.Vector3();
const halfSize = this.box3D.getSize(size).multiplyScalar(0.5);
```

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -71,6 +73,51 @@ interface Context {
...
 }
 
+function computeObjectScreenSpace(context: Context, object: OBB) {
+    // Update camera matrices
+    context.camera.camera3D.updateMatrixWorld();
+
+    // Get bounding sphere from the OBB's box3D
+    const box3 = object.box3D;
+    const sphere = new THREE.Sphere();
+    box3.getBoundingSphere(sphere);
```
ftoromanoff: Why using the sphere of the cubique OBB ? there is a clampOBB closer to the data...

packages/Main/src/Renderer/OBB.ts
```diff
@@ -114,6 +114,79 @@ class OBB extends THREE.Object3D {
...
+
+            if (d + r < 0) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    getBoundingSphere(sphere: THREE.Sphere = new THREE.Sphere()): THREE.Sphere {
```
ftoromanoff: I don't see where you use it. Is it just for having it available ?

packages/Debug/src/PointCloudDebug.js
```diff
@@ -49,119 +49,62 @@ function setupControllerVisibily(gui, displayMode, sizeMode) {
...
-                    }
-                });
-            }
+        if (!debugState.displayParentBounds && node.parent?.visible && node.parent?._obbHelper) {
+            node.parent._obbHelper.visible = isVisibleLeaf(node.parent);
+        }
+    }
 
-            node.link[layer.id] = helper;
+    layer.addEventListener('tile-visibility-change', (event) => {
```
ftoromanoff: I would say that it's not a bad idea to remove the layer and use event, But I think it should be done in antoher PR and done for the other debuger as well...

ftoromanoff (changes requested):

ftoromanoff:
@Desplandis The checks have not all passed...

packages/Main/src/Renderer/OBB.ts
Desplandis: I finally decided to do without changes to OBB.

packages/Main/src/Renderer/OBB.ts
Desplandis: I finally decided to do without changes to OBB.

packages/Main/src/Layer/PointCloudLayer.ts
Desplandis: I'm using the current sse-based approach now.

packages/Main/src/Renderer/OBB.ts
Desplandis: I removed it.

packages/Debug/src/PointCloudDebug.js
Desplandis: Did it in #2721!

Desplandis:
I'm waiting for the tests to passed (they did locally), but @ftoromanoff could you review this PR today?

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -418,71 +418,77 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
+            if (!visible) {
+                continue;
+            }
 
-        point.copy(context.camera.camera3D.position)
-            .applyMatrix4(object3d.matrixWorldInverse as THREE.Matrix4);
+            numVisiblePoints += node.numPoints;
 
-        const distanceToCamera = bbox.distanceToPoint(point);
+            // TODO: See if we can limit the calcul of the matrixWorlInverse.
```
ftoromanoff: to remove

packages/Main/test/unit/pointcloudlayer.js
```diff
@@ -6,7 +6,7 @@ import PointCloudLayer from 'Layer/PointCloudLayer';
 const TTL = 10000;
 const EPSILON = 10;
 
-function createClampOBB(
+function createVoxelOBB(
```
ftoromanoff: I could say you need ClampOBB instead of voxellOBB

packages/Main/test/unit/pointcloudlayer.js
```diff
@@ -27,7 +27,7 @@ function createNode(params = {}) {
         obj: undefined,
         children: [],
         pointSpacing: 1,
-        clampOBB: createClampOBB(),
+        voxelOBB: createVoxelOBB(),
```
ftoromanoff: change to revert

packages/Main/test/unit/pointcloudlayer.js
```diff
@@ -170,7 +170,7 @@ describe('PointCloudLayer', function () {
             const parent = createNode({ depth: 0, children: [child] });
             const context = createContext({
                 // isBox3Visible rejects only the child's bbox
-                camera: { isBox3Visible: bbox => bbox !== child.clampOBB.box3D },
+                camera: { isBox3Visible: bbox => bbox !== child.voxelOBB.box3D },
```
ftoromanoff: to revert

ftoromanoff (changes requested):

packages/Main/test/unit/pointcloudlayer.js
Desplandis: Fixed!

packages/Main/src/Layer/PointCloudLayer.ts
Desplandis: Fixed!

packages/Main/test/unit/pointcloudlayer.js
Desplandis: Fixed!

packages/Main/test/unit/pointcloudlayer.js
Desplandis: Fixed!

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -418,71 +418,76 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
-        if (!visible) {
-            return [];
-        }
+            if (!visible) {
+                continue;
+            }
 
-        point.copy(context.camera.camera3D.position)
-            .applyMatrix4(object3d.matrixWorldInverse as THREE.Matrix4);
+            numVisiblePoints += node.numPoints;
```
ftoromanoff, line 452: To be exact, it's not really the number of point visible (by the user) but the number of point sent to the shader, is it ?
Maybe we shoudl try to find a way to be more accurate, to display a specific number of point. But probalby not here anyway.

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -418,71 +418,76 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
-                return elt.children;
+            node.visible = visible;
+            node.notVisibleSince = undefined;
+            this._candidateNodes.push(node);
+            this.loadData(node, context, layer, distanceToCamera);
+
+            if (node.children && node.children.length) {
+                if (node.sse >= 1) {
+                    for (const child of node.children) {
+                        queue.push({ node: child, weight: node.sse });
```
ftoromanoff, line 474: Shouldn't we calculate the sse of the child and not the sse of the parent ?

Desplandis: Since we subdivide due to parent `sse` being too large, I wanted to keep the same behaviour and push the children with the same priority.

packages/Main/src/Layer/PointCloudLayer.ts
Desplandis: This is number of points that the user may see if all data is loaded. If we were to count from the number of loaded and visible points, we'll have pop-in/pop-out when data arrives. I think it is more stable this way.

packages/Main/src/Layer/PointCloudLayer.ts
ftoromanoff: We should keep that in mind for improvment.

packages/Main/src/Layer/PointCloudLayer.ts
ftoromanoff: Ok

ftoromanoff (approved):
Open issue on what still to improvement for PointCloudLayer:
- priority for children (see or distance)
- nb of displayed point rather as rendered

Issue #2633, opened 2025-11-17
Potree2 PointCloud doesn't appear with PlanarView


ketourneau:
Since https://github.com/iTowns/itowns/pull/2498

Potree2 point clouds no longer appear with PlanarView.

Source code : https://github.com/bloc-in-bloc/itowns/blob/f/potree2-planarView-bug/examples/view_25d_map.html#L108

Before https://github.com/iTowns/itowns/pull/2498
<img width="1417" height="771" alt="Image" src="https://github.com/user-attachments/assets/55573f43-2e92-4d80-9356-0923f580b557" />

After https://github.com/iTowns/itowns/pull/2498
<img width="1251" height="779" alt="Image" src="https://github.com/user-attachments/assets/413cb36c-94be-4bcb-9512-ca66c7805082" />

ftoromanoff:
I indeed forgot to update the Potree2BrotliLoader...
I open a PR #2635 to correct this oversight.

<img width="861" height="779" alt="Image" src="https://github.com/user-attachments/assets/e42b66b9-37d1-41c6-babc-ffd42f407a58" />


A refactorization between the commun part of the 2 Potree2Loader, should be done, to avoid similar oversight later on...

ftoromanoff:
The PR #2635 fixing this had been merged into master. Can you confirme that the bug is fixed ? @ketourneau Thanks in advance !

ketourneau:
Thanks for the fix ! It's work now in the simple example but it still doesn't work in our use case (custom controls, custom renderer, custom camera) https://github.com/bloc-in-bloc/itowns/blob/f/custom-pointcloud-sample/examples/demo_hackathon_orvault_planarView.html

ketourneau:
On master branch (demo_hackathon_orvault_planarView.html) :
<img width="1916" height="932" alt="Image" src="https://github.com/user-attachments/assets/0348a53b-92a5-4e15-90b8-dba97d05f46c" />

Before the changes on point cloud (demo_hackathon_orvault_planarView.html) :
<img width="1914" height="931" alt="Image" src="https://github.com/user-attachments/assets/74860726-56a3-4933-9257-9577ac63136d" />

ftoromanoff:
From what I started to check, it's a problem with the matrixWorld of the camera3D of the custom camera.
The update loop calculate that the bbox of the node are not in the cameraSpace...
[...]

PR #2656, opened 2025-12-12
fix: Custom example with custom projection

ftoromanoff:
linked to the issue #2633.

A proposal to solve the problemethat appeared with the custom example 
https://github.com/bloc-in-bloc/itowns/blob/f/custom-pointcloud-sample/examples/demo_hackathon_orvault_planarView.html

The custome transformation apply to the global object3d wasn't apply to the newly generated OBB, thus considering that they were elsewhere.

### DIFF
--- a/packages/Main/src/Core/LasNodeBase.ts
+++ b/packages/Main/src/Core/LasNodeBase.ts
@@ -1,4 +1,4 @@
-import { Vector3 } from 'three';
+import { Vector3, type Group } from 'three';
 import type { Hierarchy } from 'copc';
 import PointCloudNode, { PointCloudSource } from 'Core/PointCloudNode';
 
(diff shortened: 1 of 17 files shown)


jailln (dismissed):

packages/Main/src/Core/LasNodeBase.ts
```diff
@@ -106,8 +106,8 @@ abstract class LasNodeBase extends PointCloudNode {
             childClampBBox.min.z = Math.max(childClampBBox.min.z, this.source.zmin);
         }
 
-        childNode.voxelOBB.matrixWorldInverse = this.voxelOBB.matrixWorldInverse;
-        childNode.clampOBB.matrixWorldInverse = this.clampOBB.matrixWorldInverse;
+        this.clampOBB.parent!.add(childNode.clampOBB);
```
jailln: what does the `!.` syntax means ? Do you mean `?.` instead ? (there is other occurences in the MR)

ftoromanoff: the !. is a TS syntax. it lmight be in a way similar to the ?. but on the type and not on the value.
In this case, I know that parent is always define even if parent on a generic clampOBB can be definied or not, as we call the createChildOBB, only on a child node (thus with a parent)

ftoromanoff: Fixed ! I used ```as Group``` instead

jailln (approved):

PR #2657, opened 2025-12-15
Migrate pointcloud-related modules

ftoromanoff:
#2651 reopened with deeper migration:
- move subclasses to own file
- add keywords override
- add return type in all functions and child functions

## Description

This PR migrates most of pointcloud-related modules (i.e. nodes, sources and layers) to typescript. It also set what is the public API of both pointcloud sources and layers. This a first step towards a refactoring of all pointclouds.


packages/Main/src/Core/EntwinePointTileNode.ts
```diff
@@ -0,0 +1,226 @@
...
+
+const size = new Vector3();
+const position = new Vector3();
+const translation = new Vector3();
+
+function buildVoxelKey(depth: number, x: number, y: number, z: number): string {
+    return `${depth}-${x}-${y}-${z}`;
+}
+
+export abstract class LasNodeBase extends PointCloudNode {
```
HoloTheDrunk: Move this to its own file since it's a superclass of multiple point cloud node variants.

packages/Main/src/Core/CopcNode.ts
```diff
@@ -0,0 +1,137 @@
...
+
+        this.url = this.source.url;
+
+        this.crs = crs;
+
+        this.entryOffset = entryOffset;
+        this.entryLength = entryLength;
+    }
+
+    private async _fetch(offset: number, size: number) {
```
HoloTheDrunk: Indicate the return type. It's a promise yes, but we need to know what it contains.

packages/Main/src/Core/CopcNode.ts
```diff
@@ -0,0 +1,137 @@
...
+            node.findAndCreateChild(depth, x,     y + 1, z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y + 1, z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y,     z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y,     z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y + 1, z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y + 1, z + 1, hierarchy, stack);
+        }
+    }
+
+    findAndCreateChild(
```
HoloTheDrunk: Choosing this method at random for this suggestion: please add a short bit of TSDoc to each method, explaining what it does, giving a minimal 2-3 line example and presenting common pitfalls if there are any. Classes should get the same treatment with a usage snippet and property descriptions.

Also, make a habit of indicating the return type even if it's void.

packages/Main/src/Core/CopcNode.ts
```diff
@@ -0,0 +1,137 @@
...
+        return this.source.fetcher(this.source.url, {
+            ...this.networkOptions,
+            headers: {
+                ...this.networkOptions.headers,
+                range: `bytes=${offset}-${offset + size - 1}`,
+            },
+        });
+    }
+
+    async loadOctree() {
```
HoloTheDrunk: Indicate the return type.

packages/Main/src/Core/CopcNode.ts
```diff
@@ -0,0 +1,137 @@
...
+            byteSize,
+            this.source,
+            pointCount,
+            this.crs,
+        );
+        this.add(child as this, 0);
+        stack.push(child);
+    }
+
+    async load() {
```
HoloTheDrunk: Return type. I'll limit comments about it to this file, but please do the same for every other function in this PR, including getters. Having an extra check helps catch bugs when refactoring.

packages/Main/src/Core/EntwinePointTileNode.ts
```diff
@@ -0,0 +1,226 @@
...
+            node.findAndCreateChild(depth, x,     y + 1, z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y + 1, z,     hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y,     z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y,     z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x,     y + 1, z + 1, hierarchy, stack);
+            node.findAndCreateChild(depth, x + 1, y + 1, z + 1, hierarchy, stack);
+        }
+    }
+
+    findAndCreateChild(
```
HoloTheDrunk: Mark overridden methods as such with the `override` keyword, it can't hurt.

HoloTheDrunk (changes requested):

packages/Main/src/Core/CopcNode.ts
ftoromanoff: I think @Desplandis had some commits waiting, that was dealing with this part.
If it's not the case or not completly answering your demand, I will do it in another PR.

packages/Main/src/Core/CopcNode.ts
ftoromanoff: Done.
(and for function/getter)

packages/Main/src/Core/CopcNode.ts
ftoromanoff: Done.
(and for function/getter)

packages/Main/src/Core/CopcNode.ts
ftoromanoff: Done.
(and for function/getter)

packages/Main/src/Core/EntwinePointTileNode.ts
ftoromanoff: Done. For all functions that needs this keyword.

packages/Main/src/Core/EntwinePointTileNode.ts
ftoromanoff: Done (as well as for PotreeNodeBase)

HoloTheDrunk (approved):
LGTM

PR #2660, opened 2025-12-18
fix(docs): exclude sub-folder node_modules to fix doc build

Neptilo:
## Description
Exclude sub-folder node_modules from doc build script

## Motivation and Context
The doc build job is currently broken, preventing new version of iTowns from being published.
Since the integration of three-geospatial, that module is both directly included and included by 3d-tiles-renderer, in different versions, which makes the npm install scripts create a node_modules in the Main package sub-folder, no matter what we do.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ potree
 /docs/out/
 **/coverage
 .nyc_output/
+buildDocs/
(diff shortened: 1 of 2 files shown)


HoloTheDrunk (approved):
I don't see an immediate issue this could cause, so tentative LGTM? Worst case we just roll it back and find another fix.

Issue #2655, opened 2025-12-10
PMTiles MVT


ymoisan:
# Add possibility to read PMTiles MVT

This issue is a **feature proposal** for which I have a branch : https://github.com/ymoisan/itowns/tree/fix/pmtiles-mvt. I see the PR would imply a merge conflict but I doubt it is a very big one since most code is additive rather than replaces current code.

## Context

As I mention in #2542 I believe users should have an easier way to use their own basemaps and elevation data. I see most of the calls related to OGC services are hardcoded (IGN services?).  At first I thought of a PR to enable users to select their own OGC web services but I found the specificity and breadth of web services currently offered may not be available to other users.

In that regard, I believe it would be very interesting to replace OGC web services with PMTiles not only because it would give extra opportunities for users to use their own data but also because PMTiles are much easier to use than OGC web services from an infrastructure point of view : no [WMS|WMTS|WFS|etc] server needed; just files on object storage accessible via http range requests ("cloud native").

This is the first part of a plan to integrate PMTiles as a replacement of OGC services starting with vector tiles (MVT).  As a function of discussion with the commuinity, we'll see if raster PMTiles and "elevation" PMTiles could follow suit.  I'm leaving the following sections empty because iTowns developers can see what I have done in my branch.  I'll be happy to continue discussing in #2542.

Desplandis:
Hi, and thank you for your contribution.

I’m in favour of adding support for PMTiles, so you have my vote on that!

However, I don't support replacing all OGC web services with PMTiles. iTowns should be modular enough to support new formats without removing the existing ones, especially OGC standards. There may be some caveats for vector layers (I don't know if the styling is generic enough) but we should tend to modularity.

Regarding the attached implementation, I think we could use it as a starting point for a technical discussion, but I noticed you forgot to commit both `PMTilesSource` and `PMTilesVectorSource`. x)

I will not be able to review a PR before Jan. 2026 however.

P.S.: Just to clarify, iTowns isn’t hardcoded to use IGN web services, although our examples are quite IGN-centric. This is something we should address at some point. @jailln @Calvina24 @alavenant

P.P.S.: Would it be possible to precise your use-cases? Why did you choose PMTiles and not of OGC/MVT Vector Tiles?

ymoisan:
Thank you for your comments. 

>  iTowns should be modular enough to support new formats without removing the existing ones, especially OGC standards. 

[...]

PR #2662, opened 2025-12-19
Add PMTiles basemaps to iTowns

ymoisan:
I open this PR as per [this issue](https://github.com/iTowns/itowns/issues/2655) and [this discussion](https://github.com/iTowns/itowns/discussions/2542).

As I stated in the discussion item, I would like to be able to use PMTiles instead of OGC web services, provided that all "flavours" of PMTiles [(MVT, raster, elevation)](https://docs.protomaps.com/pmtiles/) cover the requirements for iTowns to work.  If so, it is my opinion that functional requirements for iTowns could be dramatically relaxed : no web server needed, just simple http access to files.  Simplicity in using different basemaps and other map artifacts than what is currently provided would make iTowns easier to personnalize for other organizations, especially those that do not have the breadth of OGC web services currently needed to accommodate iTowns.

I would like this PR to serve as a starting point for discussing the integration of "cloud-native" file formats in iTowns.  The example provided is just to show PMTiles of MVT type can be overlaid in iTowns.  My hope is that a well prepared PMTiles could provide the same output than OGC services and therefore could be a substitue rather than mererly an overlay.

### DIFF
--- a/examples/source_pmtiles_vector.html
+++ b/examples/source_pmtiles_vector.html
@@ -1,0 +1,322 @@
+<html>
+    <head>
+        <title>Itowns - PMTiles Vector Source</title>
+
+        <script type="importmap">
+        {
+            "imports": {
+                "itowns": "../dist/itowns.js",
+                "debug": "../dist/debug.js",
+                "GuiTools": "./jsm/GUI/GuiTools.js",
+                "LoadingScreen": "./jsm/GUI/LoadingScreen.js",
+                "three": "https://unpkg.com/three@0.170.0/build/three.module.js",
+                "three/addons/": "https://unpkg.com/three@0.170.0/examples/jsm/",
+                "lil-gui": "https://unpkg.com/lil-gui@0.19.2/dist/lil-gui.esm.js"
+            }
+        }
+        </script>
+
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+        <link rel="stylesheet" type="text/css" href="css/example.css">
+        <link rel="stylesheet" type="text/css" href="css/LoadingScreen.css">
+
+        <style type="text/css">
+            #description {
+                z-index: 2;
+                left: 10px;
+                max-width: 400px;
+                background: rgba(255, 255, 255, 0.95);
+                padding: 10px;
+                border-radius: 5px;
+                box-shadow: 0 2px 10px rgba(0,0,0,0.2);
+            }
+
+            #description .marg {
+                margin: 10px;
+            }
+
+            #layerInfo {
+                font-size: 12px;
+                color: #555;
+                margin-top: 10px;
+            }
+        </style>
+
+    </head>
+    <body>
+        <div id="viewerDiv"></div>
+        <div id="description">
+            <h3>PMTiles Vector Source</h3>
+            <p>Load vector tiles from a PMTiles archive:</p>
+            <input type="text" id="url" style="width: 100%; box-sizing: border-box;" placeholder="Enter PMTiles URL" />
+            <button id="loadButton" style="margin-top: 5px;">Load PMTiles</button>
+
+            <!-- Info -->
+            <div id="layerInfo" class="marg"></div>
+        </div>
+
+        <script type="module">
+            import * as THREE from 'three';
+            import { GUI } from 'lil-gui';
+            import setupLoadingScreen from 'LoadingScreen';
+            import * as itowns from 'itowns';
+
+            const {
+                TMSSource, WMTSSource, PMTilesVectorSource,
+                ColorLayer, ElevationLayer,
+                GlobeView, Coordinates, Fetcher, Style,
+            } = itowns;
+
+            // ---- Create a GlobeView ----
+
+            // Define camera initial position (matching 3dtiles_loader.html)
+            const placement = {
+                coord: new Coordinates('EPSG:4326', 2.351323, 48.856712),
+                range: 12500000,
+            };
+
+            const viewerDiv = document.getElementById('viewerDiv');
+
+            const view = new GlobeView(viewerDiv, placement, {
+                controls: { minDistance: 100 },
+            });
+
+            setupLoadingScreen(viewerDiv, view);
+
+            // ---- Add a basemap ----
+
+            Fetcher.json('./layers/JSONLayers/OPENSM.json').then((config) => {
+                const colorLayer = new ColorLayer('Ortho', {
+                    ...config,
+                    source: new TMSSource(config.source),
+                });
+                view.addLayer(colorLayer);
+            });
+
+            // ---- Add 3D terrain ----
+
+            function addElevationLayerFromConfig(config) {
+                config.source = new itowns.WMTSSource(config.source);
+                const elevationLayer = new itowns.ElevationLayer(config.id, config);
+                view.addLayer(elevationLayer);
+            }
+            itowns.Fetcher.json('./layers/JSONLayers/IGN_MNT_HIGHRES.json').then(addElevationLayerFromConfig);
+            itowns.Fetcher.json('./layers/JSONLayers/WORLD_DTM.json').then(addElevationLayerFromConfig);
+
+            // ---- PMTiles Vector Layer ----
+
+            let pmtilesLayer = null;
+            let currentPmtilesSource = null;
+            let gui = null;
+            let layersFolder = null;
+
+            // User-controlled opacity (persists through layer refreshes)
+            let userOpacity = 0.5;
+
+            // Create a simple layer style
+            function createLayerStyle() {
+                return new Style({
+                    fill: {
+                        color: '#4682b4',
+                        opacity: 1.0,
+                    },
+                    stroke: {
+                        color: '#ffffff',
+                        width: 1,
+                    },
+                    point: {
+                        color: '#4682b4',
+                        radius: 4,
+                        line: '#ffffff',
+                    },
+                });
+            }
+
+            // Refresh the layer by removing and re-adding it
+            async function refreshLayer(pmtilesSource) {
+                if (!pmtilesLayer || !pmtilesSource) {
+                    return;
+                }
+
+                const opacityToUse = userOpacity;
+
+                pmtilesSource.clearCache();
+                view.removeLayer('pmtiles-vectors');
+
+                pmtilesLayer = new ColorLayer('pmtiles-vectors', {
+                    source: pmtilesSource,
+                    transparent: true,
+                    opacity: opacityToUse,
+                    noTextureParentOutsideLimit: true,
+                    style: createLayerStyle(),
+                });
+
+                await view.addLayer(pmtilesLayer);
+                pmtilesLayer.opacity = opacityToUse;
+                view.notifyChange(pmtilesLayer);
+            }
+
+            // Setup lil-gui controls
+            function setupGui(pmtilesSource, names) {
+                if (gui) {
+                    gui.destroy();
+                }
+
+                gui = new GUI({ title: 'PMTiles Controls' });
+
+                // Opacity control
+                const layerSettings = { opacity: userOpacity };
+                gui.add(layerSettings, 'opacity', 0, 1, 0.01)
+                    .name('Opacity')
+                    .onChange((value) => {
+                        userOpacity = value;
+                        if (pmtilesLayer) {
+                            pmtilesLayer.opacity = value;
+                            view.notifyChange(pmtilesLayer);
+                        }
+                    });
+
+                // Layer visibility controls
+                if (names.length > 0) {
+                    layersFolder = gui.addFolder('Source Layers');
+
+                    const allLayersObj = { visible: true };
+                    layersFolder.add(allLayersObj, 'visible')
+                        .name('All Layers')
+                        .onChange((value) => {
+                            pmtilesSource.setAllLayersVisibility(value);
+                            layerControllers.forEach((ctrl) => {
+                                ctrl.setValue(value);
+                            });
+                            refreshLayer(pmtilesSource);
+                        });
+
+                    const layerControllers = [];
+                    names.forEach((layerName) => {
+                        const layerObj = { visible: pmtilesSource.getLayerVisibility(layerName) };
+                        const ctrl = layersFolder.add(layerObj, 'visible')
+                            .name(layerName)
+                            .onChange((value) => {
+                                pmtilesSource.setLayerVisibility(layerName, value);
+                                refreshLayer(pmtilesSource);
+                            });
+                        layerControllers.push(ctrl);
+                    });
+
+                    layersFolder.open();
+                }
+            }
+
+            function loadPMTiles(url) {
+                if (!url) {
+                    console.warn('No PMTiles URL provided');
+                    return;
+                }
+
+                // Remove existing PMTiles layer and GUI
+                if (pmtilesLayer) {
+                    view.removeLayer('pmtiles-vectors');
+                    pmtilesLayer = null;
+                    currentPmtilesSource = null;
+                }
+                if (gui) {
+                    gui.destroy();
+                    gui = null;
+                }
+
+                console.log('Loading PMTiles from:', url);
+                document.getElementById('layerInfo').innerHTML = '<p>Loading PMTiles header...</p>';
+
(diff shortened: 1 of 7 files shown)


ymoisan:
The question of "why PMTiles" was raised in a previous comment.  I think a nice answer lies in this article [Open Source Mapping for News: Reuters](https://scottreinhardmaps.com/blogs/custom-mapping-projects/custom-reuters-open-source-mapping).

airnez:
Hi @ymoisan
Thank you for this interesting contribution :) 
I'll start reviewing this PR in the next few days.

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+        // Set format for vector tiles
+        source.format = 'application/x-protobuf;type=mapbox-vector';
+
+        // PMTiles vector tiles are typically EPSG:3857
+        source.crs = source.crs || 'EPSG:3857';
+
+        super(source);
+
+        this.isPMTilesVectorSource = true;
+        this.isVectorSource = true;
```
airnez: Is setting `isVectorSource` necessary ? `Source.js` defines
```
this.isVectorSource = (source.parser || supportedParsers.get(source.format)) != undefined;
```
I guess this is enough considering that you set
```
source.format = 'application/x-protobuf;type=mapbox-vector';
```
What do you think ?

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+     *
+     * @param {Extent|Object} extent - The extent or tile coordinates
+     * @returns {string} Cache key
+     */
+    getDataKey(extent) {
+        if (extent.isTile) {
+            return `z${extent.zoom}r${extent.row}c${extent.col}`;
+        }
+        return `z${extent.zoom}r${extent.row}c${extent.col}`;
+    }
```
airnez: This might not be necessary. It always returns the same value whether extent.isTile is true or false. You could rely on `Source.js` code that does use the same key scheme.
```
`z${extent.zoom}r${extent.row}c${extent.col}`
```

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+        const z = extent.zoom;
+        const x = extent.col;
+        const y = extent.row;
+
+        // Fetch and parse the tile
+        features = this.getTile(z, x, y)
+            .then((data) => {
+                if (!data) {
+                    // No tile data - return empty collection
+                    return Promise.resolve(null);
```
airnez: I think it could be safer to return an empty FeatureCollection here.
```
Promise.resolve(new FeatureCollection(out));
```

examples/source_pmtiles_vector.html
```diff
@@ -0,0 +1,322 @@
...
+
+            // Refresh the layer by removing and re-adding it
+            async function refreshLayer(pmtilesSource) {
+                if (!pmtilesLayer || !pmtilesSource) {
+                    return;
+                }
+
+                const opacityToUse = userOpacity;
+
+                pmtilesSource.clearCache();
```
airnez: I don't think that you need to empty the source cache to reload the layer. It seems to work after commenting this line. In Itowns, it seems that clearing the cache is more a View / Layer responsability than source.

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+     * @param {Object} options - Options
+     */
+    onLayerAdded(options) {
+        super.onLayerAdded(options);
+
+        // Setup cache for features
+        if (!this._featuresCaches[options.out.crs]) {
+            this._featuresCaches[options.out.crs] = new LRUCache({ max: 500 });
+        }
+    }
```
airnez: I think that `Source.js` already handles that for you. Calling super will always instanciate features cache for this layer. You could simpy not implement this method, what do you think ?

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+     */
+    clearCache() {
+        if (this._featuresCaches) {
+            Object.values(this._featuresCaches).forEach((cache) => {
+                if (cache.clear) {
+                    cache.clear();
+                }
+            });
+        }
+    }
```
airnez: Is this really needed ? It looks too low-level for a Source implementation class. I commented
```
// pmtilesSource.clearCache();
```
in the example, and it seemed to work just fine

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+
+                // Parse the MVT data
+                return this.parser(data, {
+                    extent,
+                    in: this,
+                    out,
+                });
+            })
+            .catch((err) => {
+                console.error('PMTilesVectorSource: Error loading tile', z, x, y, err);
```
airnez: You could use `this.handlingError` here I guess. I think loadData is expected to either throw an error or return a FeatureCollection, even empty. (This contract could be more explicit tho)

examples/source_pmtiles_vector.html
```diff
@@ -0,0 +1,322 @@
...
+                        color: '#ffffff',
+                        width: 1,
+                    },
+                    point: {
+                        color: '#4682b4',
+                        radius: 4,
+                        line: '#ffffff',
+                    },
+                });
+            }
```
airnez: This style never changes. You could maybe instanciate it only once as a constant, what do you think ?

packages/Main/src/Source/PMTilesSource.js
```diff
@@ -0,0 +1,128 @@
...
+ * @property {boolean} isPMTilesSource - Used to checkout whether this source is
+ * a PMTilesSource. Default is true. You should not change this, as it is used
+ * internally for optimisation.
+ * @property {PMTiles} pmtiles - The PMTiles instance for accessing the archive.
+ * @property {Object} zoom - Object containing the minimum and maximum values of
+ * the level, to zoom in the source.
+ * @property {number} zoom.min - The minimum zoom level of the source.
+ * @property {number} zoom.max - The maximum zoom level of the source.
+ */
+class PMTilesSource extends Source {
```
airnez, line 25: I understand that this parent source will be useful later to specify a  Raster PMTiles Source ?

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+
+        // Y coordinate inversion (standard for web tiles)
+        this.isInverted = source.isInverted !== undefined ? source.isInverted : true;
+
+        // Initialize layer and style storage
+        this.layers = source.layers || {};
+        this.styles = source.styles || {};
+
+        // Parser for MVT tiles
+        this.parser = VectorTileParser.parse;
```
airnez: I dont think this is needed, as `Source.js` already sets this parser for `application/x-protobuf;type=mapbox-vector` WDYT ?

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * This source is designed to work with PMTiles archives containing MVT
+ * (Mapbox Vector Tiles) data.
+ *
+ * @extends PMTilesSource
+ *
+ * @property {boolean} isPMTilesVectorSource - Used to checkout whether this
+ * source is a PMTilesVectorSource. Default is true. You should not change this,
+ * as it is used internally for optimisation.
+ * @property {Object} layers - Object containing layer definitions for styling.
```
airnez, line 16: This differs from VectorTilesSource, but I think this makes sense as PMTiles are not advertised through mapbox styles. This directly corresponds to the format expected by the parser

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * Get visibility state of a layer.
+     *
+     * @param {string} layerName - The name of the source layer
+     * @returns {boolean} Whether the layer is visible
+     */
+    getLayerVisibility(layerName) {
+        return this._layerVisibility ? this._layerVisibility[layerName] !== false : true;
+    }
```
airnez: I understand that this code is very helpful to build your example page (which is great by the way!). I think that this is something that should be managed by the Layer, not the source. What you are actually doing here is modifying `source.layers` in order that the parser filters the unwanted layers. You are updating the layer filtering at the source and re-instanciating the layer, not updating visibilities. There are discussions about off-screen GPU rendering for rasterized tiles. I think this is the way to implement true layer visibility switching. For now, I suggest that you could keep all this code by moving it to your example page. What do you think ? @Desplandis I can't find any issue related to off-screen ColorLayer rendering, is there one existing already ? I think this PR gives another reason to work on this great idea.

packages/Main/src/Source/PMTilesVectorSource.js
```diff
@@ -0,0 +1,284 @@
...
+        this.whenReady = this.whenReady.then(async (header) => {
+            // Try to get metadata which may contain layer information
+            try {
+                const metadata = await this.pmtiles.getMetadata();
+                if (metadata) {
+                    this._metadata = metadata;
+
+                    // If no layers were provided, try to auto-detect from metadata
+                    if (Object.keys(this.layers).length === 0 && metadata.vector_layers) {
+                        this._setupLayersFromMetadata(metadata.vector_layers);
```
airnez, line 95: It's great to be able to parse headers directly. Thank you for implementing that part. It makes iTowns easier to use :)

airnez (changes requested):
Thank you for this contribution ! PMTiles is indeed a very interesting format.
I tested your branch, and it looked fine to me.
With one data source, it seems that some tiles are broken, do you know why ? I suspect an iTowns-related problem, after reading your code.  
https://r2-public.protomaps.com/protomaps-sample-datasets/tilezen.pmtiles

ymoisan:
Thanx for the thorough review @airnez.  I'll be honest up front : although I have coding experience, everything in this PR is "vibe coded".  I could never think of implementing all this in JS, a language I don't work with normally.  However, I knew what I wanted so I guided the implementation best I could.  

So, I submitted your comments to AI : 

**\<begin-AI>**
I completely agree with all your comments. Here's what I plan to do and a few questions:

### What I'll fix:

1. **Redundant code (lines 76, 86, 221, 234):** I'll remove `isVectorSource`, parser assignment, feature cache initialization, and `getDataKey()` - you're absolutely right that `Source.js` already handles all of this.

2. **Layer visibility management (lines 140-184):** I completely agree this belongs in the example, not the Source. As you suggested, I'll move all the layer visibility management code (`setLayerVisibility`, `setAllLayersVisibility`, `getLayerVisibility`, `getLayerNames`, and the `_originalLayers`/`_layerVisibility` storage) from the Source to the example file.

3. **Error handling (lines 263, 274):** I'll fix both - return `FeatureCollection` instead of `null` and use `this.handlingError(err)` for errors.

4. **Example optimizations (lines 135, 145):** I'll create the style object once as a constant and remove the `clearCache()` call.

### Questions:

**1. Layer visibility UI implementation (addressing your comment on lines 140-184):** I'd like to keep the layer visibility UI in the example since it's helpful for demonstrating PMTiles capabilities. As you suggested, I'll move all that code from the Source to the example. However, I'm wondering about the best approach for implementing this in the example - should I:

- Create multiple `ColorLayer` instances (one per source layer from the PMTiles) and toggle their `visible` property?
- Or modify the `source.layers` configuration in the example and re-instantiate the layer (which is what I'm currently doing, but moved to the example)?

Since you mentioned that true layer visibility switching will come with off-screen GPU rendering, what would be an acceptable interim approach for the example?

**2. Error handling strategy (addressing your comment on line 274):** You suggested using `this.handlingError()`. Should I always throw on error, or would returning an empty `FeatureCollection` be acceptable in some cases (e.g., when a tile doesn't exist)? The contract seems to be that `loadData` should either throw or return a FeatureCollection - is throwing always preferred?

**3. Metadata parsing:** Should I keep the automatic layer setup from metadata (`_setupLayersFromMetadata`) for initial layer configuration, even though I'm removing the visibility toggling parts? It's still useful for auto-detecting available layers when none are provided.

**4. PMTilesSource parent class (addressing your comment on line 25):** Yes, exactly! The parent `PMTilesSource` is intended to be reusable for future raster and elevation PMTiles implementations.

**5. Broken tiles issue (addressing your comment about https://r2-public.protomaps.com/protomaps-sample-datasets/tilezen.pmtiles):** I suspect the broken tiles might be related to the error handling issues you identified. Currently, the code:

- Returns `null` when tiles don't exist (line 263) or when errors occur (line 275)
- Caches these `null` promises (line 279), which might cause issues downstream if the code expects a `FeatureCollection`
- Swallows errors instead of propagating them, making debugging difficult

Looking at how `VectorTilesSource` handles this (line 183), it uses `this.handlingError(err)` which throws, rather than returning `null`. Also, `getTile()` returns `undefined` when a tile doesn't exist (PMTilesSource.js line 88), but the check uses `!data` which catches both `null` and `undefined`, then returns `null` which might be inconsistent.

I suspect the broken tiles might be caused by:

1. **Silent failures**: Errors during parsing or fetching are caught and return `null`, but downstream code might not handle `null` properly
2. **Cached null values**: Once a tile is cached as `null`, it will always return `null` even if it becomes available later
3. **Missing tiles**: When a tile doesn't exist, returning `null` instead of an empty `FeatureCollection` might break rendering expectations

With the error handling fixes I mentioned above (returning empty `FeatureCollection` instead of `null` and using `this.handlingError()` for actual errors), this should be resolved. However, I'm wondering:

- Should missing tiles (when `getTile()` returns `undefined`) return an empty `FeatureCollection` or throw?
- Are there any specific error patterns you've observed with the protomaps tilezen.pmtiles file that might help me debug this?

**\<end-AI>**

If that suits you I can provide code fixes.  Otherwise let's continue the discussion.

ymoisan:
I tested my example and it works fine.  

## Summary of Changes

### PMTilesVectorSource.js (simplified from 284 to 181 lines)

| Reviewer Concern | Change Made |
|------------------|-------------|
| `isVectorSource = true` redundant | Removed - `Source.js` sets this based on format |
| `parser = VectorTileParser.parse` redundant | Removed - `Source.js` sets parser from `supportedParsers` |
| `getDataKey()` duplicates `Source.js` | Removed - uses inherited method |
| `onLayerAdded()` redundant | Removed - `Source.js` handles cache setup |
| `clearCache()` too low-level for Source | Removed - View/Layer responsibility |
| Visibility methods belong at Layer level | Moved to example page |
| Return `null` unsafe | Now returns `new FeatureCollection(out)` |
| Use `this.handlingError` | Now uses proper error handling pattern |

airnez:
@ymoisan Sorry for taking this long to answer. We talked about it a few weeks ago with other contributors, and I forgot to keep you in touch.
You're the first (I think) contributor that submitted an entirely "vibe-coded" PR to this project. This is ok (Progress doesn't stand still, they say). It is however important to us that it is mentionned in the PR message. We also think that contributors should be able to understand and apply PR review comments by themselves. We'll be updating our contributing guidelines soon.

For this one, I'll review your code and provide the code modifications if needed. Your contribution seems safe to merge because it's a new independant feature. I already spend time on it, let's push it. I'll then hand the final review to another maintainer. I'll keep you in touch :)

Thank you for contributing. You are welcome to contribute again, at the sole condition that you are able to understand provided code and apply review suggestions by yourself.

ymoisan:
Not explicitly related to this PR, but still food for thought : [You Don’t Write the Code. You Don’t Read the Code Either](https://www.danshapiro.com/blog/2026/02/you-dont-write-the-code/).  One excerpt : 

"First, you recognize that, if you want to move quickly, you’re not the person best qualified to be writing code anymore. The AI writes the code.

Second, you recognize that if you’re not writing the code, and you’re still reviewing every pull request, *you* are the bottleneck. So you have to stop reading the code, too.

Third, you realize this creates an enormous pile of terrifying problems. If nobody’s writing code, who understands it? If nobody’s reading the code, how do you know it works? How do you know it’s getting better instead of worse?

Finally – and this is the part that takes a minute to land – you realize that solving those problems is your actual job now."

Progress doesn't stand still, indeed.

Issue #2618, opened 2025-10-22
[Proposal] Add adpative point size mode (as potree) or something similar


ketourneau:
This issue is a **feature proposal**. Feel free to upvote (with :+1: ), comment and provide your use-cases if you're interested by this feature.

## Context
We have user feedback that itowns doesn't have the same point cloud display as potree viewer.
Potree viewer use [adaptive point size mode ](https://github.com/potree/potree/blob/c53cf7f7e692ee27bc4c2c623fe17bd678d25558/src/materials/shaders/pointcloud.vs#L689) that seems to better suit certain needs.

### Identified use-cases
Actually our users are accustomed to the [adaptive point size mode ](https://github.com/potree/potree/blob/c53cf7f7e692ee27bc4c2c623fe17bd678d25558/src/materials/shaders/pointcloud.vs#L689) in potree. We try to replace old software (Potree based solution) by itowns and the attenuated mode does not meet expectations. iTowns attenuated mode is not as good as potree adaptive point size mode if for exemple, user want to draw lines on pointcloud when he zoom very close to object.

### iTowns attenuated mode

<img width="1571" height="863" alt="Image" src="https://github.com/user-attachments/assets/a109008c-f9ac-4424-b3c7-c60aee721aa4" />

### potree adaptive size mode

<img width="1704" height="935" alt="Image" src="https://github.com/user-attachments/assets/ad4ef9a1-7192-4db9-982c-20e2fbbe0459" />

Issue #2665, opened 2025-12-22
[Bug]: Point Cloud postUpdate


ketourneau:
## Context

I'm trying to set up adaptive size mode for pointcloud in postUpate on PointCloudLayer.

I am currently having trouble with initialization, pointcloud postUpdate is called before points are loaded.

In [MainLoop](https://github.com/iTowns/itowns/blob/master/packages/Main/src/Core/MainLoop.js#L157) it's written : 

> `postUpdate` is called when this geom layer update process is finished

But in practice, the behavior is not as described in the comment, postUpdate is called before the pointcloud nodes is loaded at initialization.

## Expected Behavior
postUpdate must be called each time a pointcloud node is loaded.

## Actual Behavior
postUpdate is called before the pointcloud nodes is loaded at initialization.

See logs in video :

https://github.com/user-attachments/assets/4b7cb655-a4fa-486a-b1e8-3403c986f267

## Possible Cause/Fix/Solution
Maybe call postUpdate each time a data is loaded ?

Desplandis:
The `postUpdate` method is called after each call of `update`.

However, since loading in update is asynchronous, it will be called before the data is loaded. When some data is loaded, it re-triggers an update loop. I think this is more of a bug in the re-trigger of the update loop + 3d objects added to the scene before re-trigger.

jailln:
> The `postUpdate` method is called after each call of `update`.
> 
> However, since loading in update is asynchronous, it will be called before the data is loaded. When some data is loaded, it re-triggers an update loop. I think this is more of a bug in the re-trigger of the update loop + 3d objects added to the scene before re-trigger.

Is there a point in having a hook (`postUpdate`) that is triggered before the data is loaded ?  Do you have a usecase in mind?

Desplandis:
The `postUpdate` hook is called after iterating over the whole octree (one `update` per node). We may want to apply some effects there on the already loaded nodes (without waiting for all nodes to be loaded). Currently this is just used to limit the number of drawn nodes and free culled 3D objects.

Per the discussion at #2630, I think we should remove entirely the distinction between `preUpdate`, `update` and `postUpdate`.

PR #2666, opened 2025-12-22
chore(pointcloud): Add adaptive size mode

ketourneau:
## Description
We try to implement [adaptive size mode](https://github.com/iTowns/itowns/issues/2618).

## Motivation and Context
Our customer needs it to draw on pointcloud.

https://github.com/user-attachments/assets/c834f131-c0fe-4393-bba2-1260de37ecf6

### DIFF
--- a/packages/Debug/src/PointCloudDebug.js
+++ b/packages/Debug/src/PointCloudDebug.js
@@ -299,7 +299,7 @@ export default {
         styleUI.add(layer, 'opacity', 0, 1).name('Layer opacity').onChange(update);
         styleUI.add(layer, 'pointSize', 0, 15).name('Point size').onChange(update);
         if (layer.material.sizeMode != undefined && view.camera.camera3D.isPerspectiveCamera) {
-            styleUI.add(layer.material, 'sizeAttenuation').name('Size attenuation')
+            styleUI.add(layer.material, 'sizeMode', PNTS_SIZE_MODE).name('Size mode')
                 .onChange(update);
             styleUI.add(layer.material, 'minAttenuatedSize', 0, 15).name('Min size')
                 .onChange((value) => {
--- a/packages/Debug/src/PointCloudDebug.js
+++ b/packages/Debug/src/PointCloudDebug.js
@@ -40,7 +40,7 @@ function setupControllerVisibily(gui, displayMode, sizeMode) {
     }
 
     sizeMode =  parseInt(sizeMode, 10);
-    if (sizeMode === PNTS_SIZE_MODE.VALUE) {
+    if (sizeMode === PNTS_SIZE_MODE.VALUE || sizeMode === PNTS_SIZE_MODE.ADAPTIVE) {
         getController(gui, 'minAttenuatedSize').hide();
         getController(gui, 'maxAttenuatedSize').hide();
     } else {
--- a/packages/Debug/src/PointCloudDebug.js
+++ b/packages/Debug/src/PointCloudDebug.js
@@ -40,10 +40,12 @@ function setupControllerVisibily(gui, displayMode, sizeMode) {
     }
 
     sizeMode =  parseInt(sizeMode, 10);
-    if (sizeMode === PNTS_SIZE_MODE.VALUE || sizeMode === PNTS_SIZE_MODE.ADAPTIVE) {
+    if ([PNTS_SIZE_MODE.VALUE, PNTS_SIZE_MODE.ADAPTIVE].includes(sizeMode)) {
+        getController(gui, 'pointSize').show();
         getController(gui, 'minAttenuatedSize').hide();
         getController(gui, 'maxAttenuatedSize').hide();
     } else {
+        getController(gui, 'pointSize').hide();
         getController(gui, 'minAttenuatedSize').show();
         getController(gui, 'maxAttenuatedSize').show();
     }
@@ -297,7 +299,6 @@ export default {
         }
 
         styleUI.add(layer, 'opacity', 0, 1).name('Layer opacity').onChange(update);
-        styleUI.add(layer, 'pointSize', 0, 15).name('Point size').onChange(update);
         if (layer.material.sizeMode != undefined && view.camera.camera3D.isPerspectiveCamera) {
             styleUI.add(layer.material, 'sizeMode', PNTS_SIZE_MODE).name('Size mode')
                 .onChange(update);
@@ -318,6 +319,7 @@ export default {
                     update();
                 });
         }
+        styleUI.add(layer, 'pointSize', 0, 15).name('Point size').onChange(update);
 
         if (layer.material.picking != undefined) {
             styleUI.add(layer.material, 'picking').name('Display picking id').onChange(update);
--- a/packages/Main/src/Layer/PointCloudLayer.ts
+++ b/packages/Main/src/Layer/PointCloudLayer.ts
@@ -557,9 +557,14 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
             data.set(this._visibilityTextureData.data);
             vnt.needsUpdate = true;
 
-            // Use natBox (native octree space) for octreeSize, as the
-            // octree hierarchy is defined in the source CRS coordinate system
-            const rootSize = this.root!.voxelOBB.natBox.getSize(new THREE.Vector3());
+            // Use box3D (projected in the local geometry coordinate frame) for
+            // octreeSize and bboxMin, because the vertex shader positions are
+            // expressed in that same local frame after CRS reprojection and
+            // quaternion rotation applied by the parsers.
+            // Using natBox (native source CRS) would mix coordinate systems
+            // (e.g. Lambert93 metres vs. ECEF metres) and produce wrong LODs
+            // for point clouds whose source CRS differs from the scene CRS.
+            const rootSize = this.root!.voxelOBB.box3D.getSize(new THREE.Vector3());
             const octreeSize = Math.max(rootSize.x, rootSize.y, rootSize.z);
 
             for (const pts of this.group.children) {
@@ -568,18 +573,16 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
                 const nodeStartOffset = this._visibilityTextureData.nodeToIndex.get(node);
                 const octreeSpacing = node.source.spacing;
 
-                // Compute the bounding box min of the node's octree cell
-                // in the local space of the THREE.Points geometry.
-                // We must use voxelOBB.natBox.min (the octree cell boundary
-                // in source CRS) rather than geomBBox.min (which only covers
-                // the actual points, not the full octree cell).
-                const natMin = node.voxelOBB.natBox.min;
-                const origin = node.obj.position;
-                const bboxMin = new THREE.Vector3(
-                    natMin.x - origin.x,
-                    natMin.y - origin.y,
-                    natMin.z - origin.z,
-                );
+                // Compute the bounding box min of the node's octree cell in
+                // the local coordinate frame used by the THREE.Points geometry.
+                // The parsers store each point as:
+                //   quat * (forward(nativePos) - centerZ0)
+                // where centerZ0 = voxelOBB.position and quat is computed at
+                // that same position.  projOBB() builds box3D with the exact
+                // same origin and the same quaternion, so box3D.min is already
+                // expressed in the geometry's local frame — no further
+                // transformation is needed.
+                const bboxMin = node.voxelOBB.box3D.min.clone();
 
                 pts.onBeforeRender = (_renderer, _scene, _camera, _geometry, material) => {
                     // @ts-expect-error Material is not typed yet
--- a/packages/Main/src/Layer/PointCloudLayer.ts
+++ b/packages/Main/src/Layer/PointCloudLayer.ts
@@ -554,6 +554,7 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
             // @ts-expect-error PointsMaterial is not typed yet
             const vnt = this.material.visibleNodes;
             const data = vnt.image.data;
+            data.fill(0);
             data.set(this._visibilityTextureData.data);
             vnt.needsUpdate = true;
 
--- a/packages/Main/src/Layer/PointCloudLayer.ts
+++ b/packages/Main/src/Layer/PointCloudLayer.ts
@@ -566,7 +566,6 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
             // (e.g. Lambert93 metres vs. ECEF metres) and produce wrong LODs
             // for point clouds whose source CRS differs from the scene CRS.
             const rootSize = this.root!.voxelOBB.box3D.getSize(new THREE.Vector3());
-            const octreeSize = Math.max(rootSize.x, rootSize.y, rootSize.z);
 
             for (const pts of this.group.children) {
                 const node = pts.userData.node;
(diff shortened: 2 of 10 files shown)


ketourneau:
I see an issue with postUpdate, pointcloud size is wrong at initialisation (postUpdate is called before points loaded) : https://github.com/iTowns/itowns/issues/2665

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -506,6 +506,38 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
+
+                pts.onBeforeRender = (_renderer, _scene, _camera, _geometry, material) => {
+                    // @ts-expect-error Material is not typed yet
+                    material.uniforms.nodeStartOffset.value = nodeStartOffset;
+                    // @ts-expect-error Material is not typed yet
+                    material.uniforms.octreeSize.value = octreeSize;
+                    // @ts-expect-error Material is not typed yet
+                    material.uniforms.octreeSpacing.value = octreeSpacing;
+                    // @ts-expect-error Material is not typed yet
+                    material.uniforms.nodeDepth.value = depth;
```
Desplandis, line 599: Those uniforms will never be transferred to the GPU program before the render if you do not call [`material.uniformsNeedUpdate`](https://threejs.org/docs/#ShaderMaterial.uniformsNeedUpdate). This leads to "instability" since the uniforms will be updated at the next frame with the values of the last call to `onBeforeRender`.

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -506,6 +506,38 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
                 this.dispatchEvent({ type: 'dispose-model', scene: obj, tile: obj.userData.node });
             }
         }
+
+        // @ts-expect-error PointsMaterial is not typed yet
+        if (this.material.sizeMode === PNTS_SIZE_MODE.ADAPTIVE) {
+            const nodes = this.getNodes(this.group.children);
+            const visibilityTextureData = this.computeVisibilityTextureData(nodes);
```
Desplandis: It would be nice to compute the visibility texture lazily (if a node change visibility during an update). I have a PR on review which introduces such mechanism (#2721).

For now, let's wait until you've stabilised the point size and my PR is merged.

packages/Main/src/Layer/PointCloudLayer.ts
ketourneau: Thanks, that's better. Now points have different size. I will continue to investigate what is wrong in the computeVisibilityTextureData

Desplandis:
Currently testing on an USGS LiDAR dataset which covers part of Illinois (https://s3-us-west-2.amazonaws.com/usgs-lidar-public/USGS_LPC_IL_4County_Cook_2017_LAS_2019/ept.json). They are still some artefacts where points (of different octree levels) sharing a similar extent do not have the same point size. Maybe an issue in the `getLOD()` function? I think we can test it by colouring the point cloud by the LOD inferred.

image

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -559,6 +620,82 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
             }
         }
     }
+
+    // Encoding the octree hierarchy in breadth-first order
+    // into a texture for adaptive point size rendering
+    // Explanation p36: https://www.cg.tuwien.ac.at/research/publications/2016/SCHUETZ-2016-POT/SCHUETZ-2016-POT-thesis.pdf
+    computeVisibilityTextureData(nodes: PointCloudNode[]) {
```
Desplandis, line 627: Could you add a test which kinds of do the same thing in reverse (i.e. building the octree from the texture)?

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -559,6 +620,82 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
+        // sort by level and hierarchy order
+        const sort = function sortNodes(a: PointCloudNode, b: PointCloudNode) {
+            if (a.depth !== b.depth) { return a.depth - b.depth; }
+            // @ts-expect-error PointCloudNode has x properties
+            if (a.x !== b.x) { return a.x - b.x; }
+            // @ts-expect-error PointCloudNode has y properties
+            if (a.y !== b.y) { return a.y - b.y; }
+            // @ts-expect-error PointCloudNode has z properties
+            return a.z - b.z;
```
Desplandis, line 636: Weird that those are not defined in `PointCloudNode`. @ftoromanoff?

packages/Main/src/Layer/PointCloudLayer.ts
Desplandis, line 627: Since we are not using a reference to `this`, I think this could be a function.

packages/Main/src/Layer/PointCloudLayer.ts
ftoromanoff: These properties are indeed not defined at PointCloudNode level,but only at LasNodeBase level....
Currently for all Potree Node (1&2) these properties doesn't exist... These part might then have strange behaviour for potree...
My PR https://github.com/iTowns/itowns/pull/2710 will homogenize potreeNode and LasNode and thus add these properties (x, y and z) for PotreeNode.

An other consern is that, why would a node with an x higher be higher in priorization than a lower one ? (same for y and z) Shouldn't we order on some other properties ?

ketourneau: Maybe it would be better to have `visibleNodes` already sort in breadth-first order ?
I'm not really sure how to sort them without using x/y/z, maybe we need to create a index on PointCloudNode like Potree ? (r01, r02 ...)

ftoromanoff: I don't think sorting them follwing the potree name would be a better ordering, why would the child 0 be before the child 1 ?
There is no need to sort them apart from the depth level, if i undertand what is a breadth-first order..

ketourneau: For each parent node, it stores:
1. A bitmask (red channel) indicating which children (out of 8) are visible
2. A relative offset (green+blue channels) to the first child in the array

The GPU shader finds a specific child by doing: firstChildOffset + popcount(bitmask & ((1 << childIndex) - 1)), i.e., it counts the active bits before the desired child. This assumes that all siblings (children of the same parent) are contiguous in the array.

If we sort only by depth, children of different parents at the same level can be interleaved.

ftoromanoff: This only applies to potree... Or from what I understand you want a way to sort PointCloudNode follwing a breadth-first order, meaning parent before child and then all children of the first parent before all children of the 2nd parent (without ordering between them). I don't fell using x, y and z which was the first solution (moreover the fact that it was currently only applyable on lasNode) was soing that.
For potree using the 'hierarchyKey' or the id might indeed go in the best direction, but thus is not applyable to LasNode.
We need something more general I would say...

Desplandis: @ketourneau Thinking a little bit more about this, would it be simpler to just do a breadth-first traversal starting from the root node? We have the set of visible nodes `visibleNodes`, so we can determine during traversal whether a node is a candidate for encoding in the texture.

I think this can be built in two passes:
- First pass, just do a BFS over the visible portion of the octree. For each `node`, record its position in the traversal order (i.e. the pixel index of the final texture). You could optionality fill the red channel from its children `node.children` or do it in the second pass.
- Second pass, you could for each visible nodes you can then encode the red and green channel
  + You can fill the red channel from `node.children`
  + You can fill the green channel from the current `node` index as well as its first visible `node.children` using the position map

Desplandis: Something like the following pseudo-code.

First pass:
```js
  nodeToIndex = new Map()
  queue = [root]
  while queue is not empty:
    node = queue.dequeue()
    if not node in visibleNodes { continue }
    nodeToIndex[node] = index++
    queue.enqueue(node.children)
```

Second pass:
```js
  for node of nodeToIndex {
    red = encodeChildrenVisibility(node.children)
    green = encodeFirstVisibleChildOffset(node.children, nodeToIndex)
  }
```

ketourneau:
@Desplandis I have fixed the issue with your dataset and I apply your suggestion (added test & refactor with 2 pass)

jailln:
@ketourneau  nice addition thanks ! Tested ok on 3D Tiles datasets :slightly_smiling_face: 

I think we should modify the point clouds examples (including 3dtiles_loader.html) to use this display mode. WDYT @Desplandis ?

jailln:
> @ketourneau nice addition thanks ! Tested ok on 3D Tiles datasets 🙂
> 
> I think we should modify the point clouds examples (including 3dtiles_loader.html) to use this display mode. WDYT @Desplandis ?

After discussion with @ketourneau it is not applied to 3D Tiles layers because the implementation is based on the assumption that the data is organized in an octree. I'm not familiar enough with this algorithm to know if and how it could be adapted to 3D tiles but it should not block this PR. We should however add in the doc and maybe in a warning that it is not supported for 3D Tiles layers.

ftoromanoff:
Small remarks for the PointCloudDebug. The 'Min size' and 'Max size' are not use for the 'new 'ADAPTIVE' 'Size mode', are they ? In this case the should be hidden as for when we set the 'Size mode' on 'VALUE'.

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -535,6 +539,67 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
+            }
+
+            if (this._visibilityTextureData === undefined) {
+                return;
+            }
+
+            // @ts-expect-error PointsMaterial is not typed yet
+            const vnt = this.material.visibleNodes;
+            const data = vnt.image.data;
+            data.set(this._visibilityTextureData.data);
```
Desplandis, line 558:
```suggestion
            const data = vnt.image.data;
            data.fill(0);
            data.set(this._visibilityTextureData.data);
```

I think there are still some stale entries from the previous updates, I recommend you to zero out all the buffer. It is weird however that we access those part of the buffer but it shall fix the issue for now.

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -535,6 +539,67 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
+            // Using natBox (native source CRS) would mix coordinate systems
+            // (e.g. Lambert93 metres vs. ECEF metres) and produce wrong LODs
+            // for point clouds whose source CRS differs from the scene CRS.
+            const rootSize = this.root!.voxelOBB.box3D.getSize(new THREE.Vector3());
+            const octreeSize = Math.max(rootSize.x, rootSize.y, rootSize.z);
+
+            for (const pts of this.group.children) {
+                const node = pts.userData.node;
+                const depth = node.depth;
+                const nodeStartOffset = this._visibilityTextureData.nodeToIndex.get(node);
```
Desplandis, line 573: Beware that `nodeStartOffset` may be undefined! This is can happen because the rendered group `this.group.children` do not guarantee that all points are visible. They are in the rendered group (were visible once) but remain hidden until disposal (10s later).

packages/Main/src/Renderer/Shader/PointsVS.glsl
```diff
@@ -40,6 +48,61 @@ attribute float returnNumber;
...
+    // to octree space (relative to node bbox min)
+    vec3 pos = position - nodeBBoxMin;
+    vec3 offset = vec3(0.0, 0.0, 0.0);
+    int iOffset = int(nodeStartOffset);
+    float depth = nodeDepth;
+    for (float i = 0.0; i <= 30.0; i++) {
+        float nodeSizeAtLevel = octreeSize / pow(2.0, i + nodeDepth);
+
+        vec3 index3d = (pos - offset) / nodeSizeAtLevel;
+        index3d = floor(index3d + 0.5);
```
Desplandis: I think the more important issue of all. It seems that `index3d` may be outside the range $[0, 1]$. Clamping here fix the big points issues. It may be because of floating point error accumulation or because we are outside the bounds...

packages/Main/src/Renderer/Shader/PointsVS.glsl
```diff
@@ -40,6 +48,61 @@ attribute float returnNumber;
...
+        int index = int(round(4.0 * index3d.x + 2.0 * index3d.y + index3d.z));
+
+        vec4 value = texture2D(visibleNodes, vec2(float(iOffset) / 2048.0, 0.0));
```
Desplandis:
```suggestion
        vec4 value = texture2D(visibleNodes, vec2((float(iOffset) + 0.5) / 2048.0, 0.5));
```

I suggest to sample at mid-pixel in case of a small floating-point error to not sample the neighbour texel.

ketourneau:
> Small remarks for the PointCloudDebug. The 'Min size' and 'Max size' are not use for the 'new 'ADAPTIVE' 'Size mode', are they ? In this case the should be hidden as for when we set the 'Size mode' on 'VALUE'.

Yes, indeed, I had asked myself the same question. I hided the 'Min size' and 'Max size'

packages/Main/src/Layer/PointCloudLayer.ts
ketourneau: Indeed, good point.

packages/Main/src/Renderer/Shader/PointsVS.glsl
ketourneau: Okay, I applied your suggestion.

packages/Main/src/Renderer/Shader/PointsVS.glsl
ketourneau: Okay, I applied your suggestion.

packages/Main/src/Layer/PointCloudLayer.ts
ketourneau: Ok I continue to next iteration if nodeStartOffset is undefined

packages/Debug/src/PointCloudDebug.js
```diff
@@ -299,7 +299,7 @@ export default {
         styleUI.add(layer, 'opacity', 0, 1).name('Layer opacity').onChange(update);
         styleUI.add(layer, 'pointSize', 0, 15).name('Point size').onChange(update);
```
ftoromanoff: This parameter initialy only have an impact on Size mode = VALUE (and not when Size mode = ATTENUATED) thus it should be put under 'Size mode' and be hidden when we are 'ATTENUATED' ?
Now i see that in the mode = ADAPTIVE, it's also has an impact even if i don't really understand to what exactly the value will refer to.

ketourneau: In ADAPTIVE mode this parameter can be use to apply a scale factor. But normally, 1.0 factor value should be enough in most cases

ftoromanoff: I would then still suggest to move this parameter after the 'Size mode' button and hide it when it has no impact. It is just purely ergonic and not algorythmic.

ketourneau: Ok done

ftoromanoff: It's fine by me. I don't validate the PR as I didn't review the rest of the code. I'll let @Desplandis do that.

Desplandis (approved):

Issue #2652, opened 2025-12-08
[Bug]: Point Cloud in globe view and colorimetry


ftoromanoff:
There is something strange with the colorimetry of the point cloud data following the merge of #2589:
- the colorimetry of the point cloud is now very white
- the globe is very grey

Before the merge we had:

<img width="1071" height="894" alt="Image" src="https://github.com/user-attachments/assets/fbe72099-a3f0-4778-a2c6-45566a5ac0ea" />
and 

<img width="1078" height="887" alt="Image" src="https://github.com/user-attachments/assets/f087d27a-8b59-4815-9f52-d16318edb49d" />

now we have:

<img width="1075" height="906" alt="Image" src="https://github.com/user-attachments/assets/b10acb32-40a5-430c-a906-b50d324db037" />
and 

<img width="1075" height="914" alt="Image" src="https://github.com/user-attachments/assets/bb0fcdf0-6991-465b-8d94-7fb690c86c61" />

(example/entwine_3d_loader view of the data and unzoomed to see the full globe)

ftoromanoff:
Adding an empty placement and option realisticLighting in the constructor of the view
```
var view = new itowns.GlobeView(viewerDiv, {}, { realisticLighting: false});
```
seems to almost do the trick: The equador and the pole are shown white...

<img width="1062" height="895" alt="Image" src="https://github.com/user-attachments/assets/d4868cd9-d270-430d-bc13-7d15d3b67433" />

Desplandis:
It seems to be caused by the use of `realisticLighting` by default which breaks the pointclouds. IMO it should be optional until we:
- have a way to set the time in the API and in the example
- are sure that this does not break other things

HoloTheDrunk:
I should've caught this during review, my apologies. Though even if off by default, the end result feels too desaturated and lacking in contrast, especially when zooming in close to the ground.
We could maybe try solving this by adjusting the atmosphere effect's intensity based on a function of camera angle, ~~surface normal (before elevation, so the surface normal at the current point on the tile below)~~ and elevation? Perhaps even providing different such functions as options that can be passed into the settings of the `GlobeView`, allowing users to pick from presets or pass in their own?

Update as of 11/12: I'm currently working on implementing this.

Neptilo:
For background information: the whitening effect was already identified during developement and somewhat attenuated (see https://github.com/iTowns/itowns/pull/2589#issuecomment-3381620199 and HoloTheDrunk's response, and mine).
[...]

PR #2670, opened 2026-01-14
Disable atmosphere by default

Desplandis:
## Description

#2589 replaced our own baked atmosphere by `three-geospatial` and enabled atmosphere by default.

However, this leads to :
- incorrect lighting for pointclouds (see #2652)
- jankiness on windows at startup (I will post a reproducible example in an issue)

We also do not have a widget to control the date (i.e. day + time of the day).

Until those points have been addressed, I propose to revert to the old disabled by default behaviour.

P.S.: I also fixed the mars example, the wmts stream URL has changed.

### DIFF
--- a/examples/mars.html
+++ b/examples/mars.html
@@ -61,7 +61,8 @@
                     outerRadius: 6600000,
                     wavelength: [0.350, 0.470, 0.575],
                     scaleDepth: 0.38,
-                }
+                },
+                realisticLighting: true,
             });
             setupLoadingScreen(viewerDiv, view);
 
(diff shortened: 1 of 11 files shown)


Neptilo:
For the decreased test coverage, maybe we could simply add a test in `globeview.js` that would do `viewer.skyManager.enabled = true`, render and set it back to false, to check there's no error.
This would mean activating `realisticRendering` during construction and setting `skyManager.enabled` to false right after, so that the other test functions run without it.

Neptilo (approved):

PR #2672, opened 2026-01-19
feat: add Docker support for running examples locally

Desplandis:
## Description

This PR adds Docker support for running iTowns examples locally by adding a `Dockerfile` using node LTS alpine.

```bash
# Locally
docker build -t itowns .
docker run -p 8080:8080 itowns

# From a remote branch
docker run -p 8080:8080 $(docker build -q https://github.com/itowns/itowns.git#<branch>)
```
where `<branch>` is the desired branch name.

Note that this is not limited to the itowns repository but can be applied on all branches rebased on this commit.

## Motivation and Context
Provides an alternative method to run iTowns examples locally without requiring a local `node` installation. This aims to:
- simplify onboarding of users who want to test features of itowns without having to setup a local node environment
- get reproducible builds (e.g. for some poc or demos)
- easily test different branches using a git URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,43 @@ here](http://www.itowns-project.org/itowns/examples/). Some examples available:
 
 [![iTowns examples](http://www.itowns-project.org/images/montage.jpg)](http://www.itowns-project.org/itowns/examples/)
 
-## How to run it locally?
+## Running locally
 
-Clone the repo and then run:
+Examples can be run locally using either npm or Docker. Both methods serve them on port `8080`.
 
-```
+### Using npm
+
+Requires [Node.js](https://nodejs.org/) (LTS version recommended).
+
+```bash
+git clone https://github.com/itowns/itowns.git
+cd itowns
 npm install
 npm start
 ```
 
-Try out the examples at http://localhost:8080/examples
+### Using Docker
+
+Requires [Docker](https://www.docker.com/).
+
+```bash
+git clone https://github.com/itowns/itowns.git
+cd itowns
+docker build -t itowns .
+docker run -p 8080:8080 itowns
+```
+
+Alternatively, you can build and run directly from a remote branch of itowns in a single command:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -39,16 +39,43 @@ here](http://www.itowns-project.org/itowns/examples/). Some examples available:
...
+
+```bash
+docker run -p 8080:8080 $(docker build -q https://github.com/itowns/itowns.git#<branch>)
+```
+
+> Replace `<branch>` with the desired branch name (e.g. `master`, `dev`).
+
+### Accessing the examples
+
+Once the server is running, the examples are available at http://localhost:8080/
```
HoloTheDrunk, line 78: It's all working on my end but you might want to amend this to tell the user to look at the "On Your Network" line of the Webpack output, as it had the server exposed on `172.17.0.2` (reserved IP address for private network communication) and not localhost.

HoloTheDrunk (changes requested):

README.md
Desplandis: The [`-p`](https://docs.docker.com/reference/cli/docker/container/run/#publish) parameter of the `docker run` command here should map the 8080 port of the container to port 8080 of `127.0.0.1` (i.e. `localhost`). Or did I miss something?

Edit :
> If you don't specify an IP address (i.e., -p 80:80 instead of -p 127.0.0.1:80:80) when publishing a container's ports, Docker publishes the port on all interfaces (address 0.0.0.0) by default. These ports are externally accessible.

So it binds the container port to all network interfaces (localhost, machine IP, ...) so more for exposing it to external users (which is usually my use-case of docker :P).

HoloTheDrunk: For some reason localhost now works..? I'm unsure what caused it to not work as expected before but problem solved I guess.

HoloTheDrunk (approved):

PR #2673, opened 2026-01-19
Fix(pointcloud): incorrect loading mechanism & add layer events

Desplandis:
## Description
This PR fixes the current loading mechanism of pointclouds. Those were added directly visible on the scene on load and bypassed the `notifyChange` mechanism.

This PR also adds layer events for model load, error and dispose for end-users and debugging purpose (the events types were based on those of `OGC3DTilesLayer`).

## Motivation and Context
This could maybe help #2665 #2666 (@ketourneau if you can rebase on this and see if it fixes at least the points pop?)

Based on #2672

### DIFF
--- a/packages/Main/src/Layer/PointCloudLayer.ts
+++ b/packages/Main/src/Layer/PointCloudLayer.ts
@@ -67,7 +67,9 @@ interface Context {
             earlyDropFunction?: (cmd: { requester: PointCloudNode }) => boolean;
         }) => Promise<THREE.Points>;
     };
-    view: object;
+    view: {
+        notifyChange: (elt: PointCloudLayer) => void;
+    };
 }
 
 function computeSSEPerspective(
@@ -323,8 +325,7 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
         this.maxElevationRange = this.maxElevationRange ?? this.source.zmax;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    preUpdate(context: Context, changeSources: any) {
+    preUpdate(context: Context) {
         // See https://cesiumjs.org/hosted-apps/massiveworlds/downloads/Ring/WorldScaleTerrainRendering.pptx
         // slide 17
         context.camera.preSSE =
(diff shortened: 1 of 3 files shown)


ketourneau:
Looks good for me, it solve https://github.com/iTowns/itowns/issues/2665 but it confirm that https://github.com/iTowns/itowns/pull/2666 it's not ready. There is a minor bug, probably when calculating the size of the points at certain times.

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -344,36 +345,6 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
-                        return [this.root];
-                    }
-                }
-            }
-        }
-
-        if (commonAncestor) {
-            return [commonAncestor];
-        }
-
```
ftoromanoff, line 376: Are we sure we want to removed that part (as you said it's never called) but an other solution might have been to try to make this part used.

packages/Main/src/Layer/PointCloudLayer.ts
```diff
@@ -414,12 +390,15 @@ abstract class PointCloudLayer<S extends PointCloudSource = PointCloudSource>
...
                 }).then((pts: THREE.Points) => {
                     elt.obj = pts;
-
+                    elt.obj.visible = false;
                     // make sure to add it here, otherwise it might never
                     // be added nor cleaned
                     this.group.add(elt.obj);
                     elt.obj.updateMatrixWorld(true);
+                    context.view.notifyChange(this);
+                    this.dispatchEvent({ type: 'load-model', scene: pts, tile: elt });
```
ftoromanoff, line 399: Currently this event type is only listened in OGC3DTilesLayer.js and OGC3DTilesDebug.js.
And I don't see here any addEventListener for these... Is it really useful to add them then ?

packages/Main/src/Layer/PointCloudLayer.ts
Desplandis: I would prefer to remove unused (pre-mature) optimisation for now, until we manage to have a perfectly stable implementation of the point cloud update mechanism. Moreover, I do not use this code in the future priority-based mechanism.

packages/Main/src/Layer/PointCloudLayer.ts
Desplandis: I use them for debugging purpose and I think this would be a nice addition to users who want to change the material (model-load) or just if we wish to implement some sort of benchmarking.

jailln: agreed with @Desplandis , always useful for users to have access to such events

packages/Main/src/Layer/PointCloudLayer.ts
ftoromanoff: I see, but we might ask @jailln and @gchoqueux to agree ? Just to be sure ?

packages/Main/src/Layer/PointCloudLayer.ts
ftoromanoff: Ok, I just currently don't see any use of them, we just need to be sure that they will be documented ;-)

packages/Main/src/Layer/PointCloudLayer.ts
jailln: Ok for me to remove it

packages/Main/src/Layer/PointCloudLayer.ts
jailln: Yes they indeed need to be documented :)

jailln (approved):

packages/Main/src/Layer/PointCloudLayer.ts
ftoromanoff: @Desplandis Was this done ?

Issue #2630, opened 2025-11-05
[Proposal][Architecture] Per-update custom update traversal


Desplandis:
This issue is a **feature proposal**. Feel free to upvote (with :+1: ), comment and provide your use-cases if you're interested by this feature.

## Context
This proposal aims to remove the [hard-coded depth-first traversal](https://github.com/iTowns/itowns/blob/67832d61dfe230ffebb408e3fe3f2ab8f786d9f8/packages/Main/src/Core/MainLoop.js#L31) of the quadtree/octree when updating a geometry layer and replace it with per-layer customizable traversal strategies. The depth-first strategy is not always optimal and leads to performance issues in some identified use-cases (see below).

## Description of the proposal
This proposal will describe the current architecture as well as different possible replacements (from non-breaking to hard breaking changes) to allow different geometry layers to use different traversal strategies. Each proposed implementation will be linked to a "proof of concept" branch to assert its feasibility.

### Identified use-cases
A non-exhaustive (feel free to complete) list :
- `PointCloudLayer` due to the current traversal strategy do not prioritize the expected nodes when querying plus queries too much nodes. The `postUpdate` step limits _a posteriori_ the rendering to the greatest priority loaded nodes (within the point budget). Since the priority changes with respect to the loaded nodes, this causes pop-in/pop-out effects (especially for huge datasets). We could use a priority queue to limit _a priori_ during the `update` step(s) the queried nodes (with respect to the point budget).

## Implementation
In the following, let:
- `G` denote the current _geometry layer_ updated (this does not include `FeatureGeometryLayer` BTW)
- `l` an _attached layer_ to $G$ (usually a `RasterLayer`/`FeatureGeometryLayer`)
- `S` the _set of changes_ (i.e. nodes of `L`, `L` itself, the camera and/or the view) triggering the update

For ease of understanding, we made the following assumptions:
- the current _context_ is implicitly passed to the methods
- the attached layers are all ready

### Current implementation
Let's start with our current implementation. The geometry layer interface could be defined as:
```ts
interface GeometryLayer<Node, O extends Object3D>  {
     attached: AttachedLayer[]
     preUpdate(changes: S) -> Node[]
     update(n: Node) -> Node[]
     postUpdate(changes: S)
     getObjectToUpdateForAttached(n: Node) -> O[]
}
```
where `Node` is either a:
- `TileMesh` for `TiledGeometryLayer`
- `PointCloudNode` for `PointCloudLayer`
[...]

Issue #2274, opened 2024-02-16
[Proposal - Coding style] Unify private methods and fields convention


jailln:
## Context

There are at least two conventions used in the code to indicate private fields and methods: 
* [Javascript private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)
* `_` prefix

We should choose one and harmonize the code.

## Proposal

In my opinion Javascript private properties implementation is not stable enough:
* transpiled code is hard to debug
* classes using them cannot be proxied, meaning you [cannot use them in vuejs apps](https://lea.verou.me/blog/2023/04/private-fields-considered-harmful/), which should clearly be a no go
* not really adopted in the javascript world

Sadly and since there is no solid implementation of private fields in js, I propose to go back to good old prefixes ( :cry: ). I propose using `__` which is in my opinion clearer than `_` and also commonly used.

## Implementation

* [ ] Update the code to harmonize private fields implementation
* [ ] update doc generation rules to not generate jsdoc in the public API for such fields
* [ ] add these rules in the [coding](https://github.com/iTowns/itowns/blob/master/CODING.md) file

jailln:
@Desplandis @mgermerie @ftoromanoff @AnthonyGlt what is your opinion on this ?

ftoromanoff:
I don't really know about the double _ or single _ but i do agree that we should use these method over the other one. (At least up to the time the private properties might be well implemented).

Desplandis:
It seems that the blog post you mentioned above presents a compelling argument for library authors to not use the `private` keywords. It restricts the use of proxies and thus makes `itowns` incompatible with popular reactive libraries (vue, react, ...) which intensively uses proxying.

For this reason alone, I agree with the current proposal, but it should be clear that we, as library authors, could break anything that is prefixed by `_`.

I vote for the use of `_` as it is personally clearer than `__` (double underscore) and already used in the `itowns` codebase. x)

AnthonyGlt:
It sounds that the Javascript private properties convention is causing too much trouble, so we should revert it.
Concerning the distinction between `__` and `_`, do you have some kind of documentation saying that the double underscore is clearer and more commonly used ?
It looks like it's more common to use single underscore than double and, in my opinion, using double underscore could encourage to also use single one in some declaration, which could lead to misunderstanding in the code.
I vote for the single underscore `_` :+1:

jailln:
Ok, let's go for single `_` then :) thanks!
[...]

PR #2675, opened 2026-01-20
refactor: migrate MainLoop to TypeScript

HoloTheDrunk:
## Description

Migrates MainLoop to TypeScript and cleans up some of the outdated and/or unused code.

## Motivation and Context

Part of the project-wide migration to TypeScript, I needed this file's migration to happen for other PRs I'm working on.

### DIFF
--- a/examples/vector_tile_raster_2d.html
+++ b/examples/vector_tile_raster_2d.html
@@ -78,7 +78,7 @@
             view.addLayer(mvtLayer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
 
             // Request redraw
-            view.notifyChange(true);
+            view.notifyChange({});
             debug.createTileDebugUI(menuGlobe.gui, view);
             window.view = view;
             window.itowns = itowns;
--- a/packages/Main/src/Core/MainLoop.ts
+++ b/packages/Main/src/Core/MainLoop.ts
@@ -13,21 +13,21 @@ export const RENDERING_SCHEDULED = 1;
  */
 export const MAIN_LOOP_EVENTS = {
     /** Fired at the start of the update */
-    UPDATE_START: 'update_start',
+    UPDATE_START: 'update_start' as const,
     /** Fired before the camera update */
-    BEFORE_CAMERA_UPDATE: 'before_camera_update',
+    BEFORE_CAMERA_UPDATE: 'before_camera_update' as const,
     /** Fired after the camera update */
-    AFTER_CAMERA_UPDATE: 'after_camera_update',
+    AFTER_CAMERA_UPDATE: 'after_camera_update' as const,
     /** Fired before the layer update */
-    BEFORE_LAYER_UPDATE: 'before_layer_update',
+    BEFORE_LAYER_UPDATE: 'before_layer_update' as const,
     /** Fired after the layer update */
-    AFTER_LAYER_UPDATE: 'after_layer_update',
+    AFTER_LAYER_UPDATE: 'after_layer_update' as const,
     /** Fired before the render */
-    BEFORE_RENDER: 'before_render',
+    BEFORE_RENDER: 'before_render' as const,
     /** Fired after the render */
-    AFTER_RENDER: 'after_render',
+    AFTER_RENDER: 'after_render' as const,
     /** Fired at the end of the update */
-    UPDATE_END: 'update_end',
+    UPDATE_END: 'update_end' as const,
 };
 
 type Context = {
--- a/packages/Main/src/Core/MainLoop.ts
+++ b/packages/Main/src/Core/MainLoop.ts
@@ -1,4 +1,8 @@
-import { EventDispatcher } from 'three';
+import { GeometryLayer, Layer, View } from 'Main';
+import { EventDispatcher, Camera as ThreeCamera } from 'three';
+import Camera from 'Renderer/Camera';
+import c3DEngine from 'Renderer/c3DEngine';
+import Scheduler from './Scheduler/Scheduler';
 
 export const RENDERING_PAUSED = 0;
 export const RENDERING_SCHEDULED = 1;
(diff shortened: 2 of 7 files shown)


Desplandis:
I did some brief explanation of the main loop mechanism in #https://github.com/iTowns/itowns/issues/2630 if that helps!

packages/Main/src/Core/MainLoop.ts
```diff
@@ -1,108 +1,128 @@
...
     BEFORE_LAYER_UPDATE: 'before_layer_update',
+    /** Fired after the layer update */
     AFTER_LAYER_UPDATE: 'after_layer_update',
+    /** Fired before the render */
     BEFORE_RENDER: 'before_render',
+    /** Fired after the render */
     AFTER_RENDER: 'after_render',
+    /** Fired at the end of the update */
     UPDATE_END: 'update_end',
 };
```
Desplandis, line 32: Since they are used as a constant set of events, you could refine the type by marking it as `as const`. As such the inferred type would be
```ts
{
    UPDATE_START: 'update_start';
    ...
}
```
instead of
```ts
{
    UPDATE_START: string;
    ...
}
```

packages/Main/src/Core/MainLoop.ts
```diff
@@ -1,108 +1,128 @@
-import { EventDispatcher } from 'three';
+import { GeometryLayer, Layer, View } from 'Main';
+import { EventDispatcher, Object3D, Camera as ThreeCamera } from 'three';
+import Camera from 'Renderer/Camera';
+import c3DEngine from 'Renderer/c3DEngine';
+import Scheduler from './Scheduler/Scheduler';
```
Desplandis: Could you mark some of the imports here as `import { type ... }`? I think only the `EventDispatcher` is instantiated (through `super()`).

packages/Main/src/Core/MainLoop.ts
```diff
@@ -1,108 +1,128 @@
...
+function filterChangeSources(
+    updateSources: Set<UpdateSource>,
+    geometryLayer: GeometryLayer,
+): Set<UpdateSource> {
     let fullUpdate = false;
-    const filtered = new Set();
+    const filtered = new Set<UpdateSource>();
     updateSources.forEach((src) => {
-        if (src === geometryLayer || src.isCamera) {
+        if (src === geometryLayer || (src as ThreeCamera).isCamera) {
```
Desplandis: I would prefer the use of `isCamera` in src, as it narrows the type!

packages/Main/src/Core/MainLoop.ts
```diff
@@ -1,108 +1,128 @@
...
                 }
             }
         }
+
         updateElements(context, geometryLayer, newElementsToUpdate);
     }
 }
 
-function filterChangeSources(updateSources, geometryLayer) {
+// TODO: Figure out what that last type actually is
```
Desplandis: Could you move the type on top of the file, I find it a little more readable. The last type is actually one of the numerous node type of itowns: `TileMesh` nodes, previously `PointCloudNode` nodes until we removed the `layer` property (whoopsie) and deprecated `C3DTilesLayer` ones.

So just as you inferred just an object with a `layer` property.

packages/Main/src/Core/MainLoop.ts
```diff
@@ -1,108 +1,128 @@
...
             filtered.add(src);
         }
     });
     return fullUpdate ? new Set([geometryLayer]) : filtered;
 }
 
+type MainLoopEvents = {
+    // An unknown body indicates an empty event
+    'command-queue-empty': unknown;
+};
```
Desplandis: Move the type at the top of the file.
What is the difference between using `unknown` or `object` there?

packages/Main/src/Core/MainLoop.ts
```diff
@@ -174,19 +199,18 @@ class MainLoop extends EventDispatcher {
         const willRedraw = this.#needsRedraw;
         this.#lastTimestamp = timestamp;
 
-        // Reset internal state before calling _update (so future calls to View.notifyChange()
-        // can properly change it)
+        // Reset internal state before calling _update (so future calls to
+        // View.notifyChange() can properly change it)
         this.#needsRedraw = false;
         this.renderingState = RENDERING_PAUSED;
         const updateSources = new Set(view._changeSources);
```
Desplandis: `view._changesSources` is `any` there. Why did the strict mode not catch this, is this because we are mixing untyped js & ts files?

packages/Main/src/Core/MainLoop.ts
```diff
@@ -120,26 +140,29 @@ class MainLoop extends EventDispatcher {
...
 
         // replace layer with their parent where needed
-        updateSources.forEach((src) => {
+        updateSources.forEach((src: UpdateSource) => {
+            // @ts-expect-error True JS shenanigans
             const layer = src.layer || src;
             if (layer.isLayer && layer.parent) {
                 updateSources.add(layer.parent);
             }
         });
```
Desplandis, line 158: Would be better with a reduce? This is one ugly mutation in a for-each...

packages/Main/src/Core/MainLoop.ts
```diff
@@ -206,30 +230,33 @@ class MainLoop extends EventDispatcher {
...
         view.execFrameRequesters(MAIN_LOOP_EVENTS.UPDATE_END, dt, this.#updateLoopRestarted);
     }
 
-    #renderView(view, dt) {
+    #renderView(view: View, dt: number) {
         view.execFrameRequesters(MAIN_LOOP_EVENTS.BEFORE_RENDER, dt, this.#updateLoopRestarted);
 
+        // @ts-expect-error Checking dynamically-added method
         if (view.render) {
+            // @ts-expect-error Result of the above
```
Desplandis: Maybe we could use an clean interface for `View` until `View` is typed? Or JSDoc the necessary parts of `View`.

packages/Main/src/Core/MainLoop.ts
```diff
@@ -1,108 +1,128 @@
...
+
+    public renderingState: typeof RENDERING_PAUSED | typeof RENDERING_SCHEDULED;
+    public scheduler: Scheduler;
+    public gfxEngine: c3DEngine;
+
+    constructor(scheduler: Scheduler, engine: c3DEngine) {
         super();
         this.renderingState = RENDERING_PAUSED;
         this.scheduler = scheduler;
         this.gfxEngine = engine; // TODO: remove me
```
Desplandis, line 121: Not until we remove all dozens of references hidden in itowns my friend.

packages/Main/src/Core/MainLoop.ts
```diff
@@ -120,26 +140,29 @@ class MainLoop extends EventDispatcher {
         }
     }
 
-    #update(view, updateSources, dt) {
-        const context = {
+    #update(view: View, updateSources: Set<UpdateSource>, dt: number) {
```
Desplandis: While we are still at it, could you replace `#` by `private _`? I can't remember the reasons but I think it causes some conflicts with the use of `Proxy` (and so with every reactive lib out there, nice standard we got there)? @jailln ?

Desplandis (commented):
I have some comments but +1 for typechecking this module + simplify `sub.elements` mess.

packages/Main/src/Core/MainLoop.ts
jailln: @Desplandis you remember well, see #2274

packages/Main/src/Core/MainLoop.ts
HoloTheDrunk: `TileMesh.layer` is of type `TileLayerLike` which is defined as
```ts
interface TileLayerLike {
    tileMatrixSets: string[];
}
```
and this understandably makes the comparison to a `GeometryLayer` invalid. When you say "`TileMesh` nodes", are you referring to a type different from `TileMesh`?

Desplandis: I refer to `TileMesh` (which is both a quadtree node and a mesh) of `TiledGeometryLayer`. Since we only access the `tileMatrixSets` property `layer` in `TileMesh` we only partially typed it. Keep it as `{ layer: Layer }` in the main loop, we'll do the glue later on!

packages/Main/src/Core/MainLoop.ts
HoloTheDrunk: After a bit of testing, I don't believe there's a difference in this case.

packages/Main/src/Core/MainLoop.ts
HoloTheDrunk: It's because it's an inferred type from View.js and not an explicit TS one yes.

Desplandis: It will bite us for sure. If you have a way to easily type it in jsdoc, it would be better!

packages/Main/src/Core/MainLoop.ts
HoloTheDrunk: Given that we have ts-expect-error there, we'll be reminded by the linter to change those lines when we do type View, so I think it's fine to leave this as is.

packages/Main/src/Core/MainLoop.ts
HoloTheDrunk: It's not ideal but it works and is likely more efficient than keeping a separate data structure. I think it's one of the rare good cases for modifying a data structure during iteration. `reduce` isn't defined on sets and I'm not exactly sure how you'd use it to achieve the same effect apart from treating it like a `forEach` and discarding the result..?

Desplandis: Arf.. No `reduce` nor `map` on sets, that sucks! Since the order is not defined, let's hope we never encounter a double level of ancestry on the layers. x)

packages/Main/src/Core/MainLoop.ts
Desplandis: Hmm... We could also add a `render` property in `View` initialized to undefined and type it as `@type {undefined | function():void}`?

packages/Main/src/Core/MainLoop.ts
Desplandis: You did not take my comment into account.

```suggestion
} as const;
```

examples/vector_tile_raster_2d.html
```diff
@@ -78,7 +78,7 @@
             view.addLayer(mvtLayer).then(menuGlobe.addLayerGUI.bind(menuGlobe));
 
             // Request redraw
-            view.notifyChange(true);
+            view.notifyChange({});
```
Desplandis, line 81: Prefer calling it without parameters or with `layer` or `view` as parameter if a layer or a set of layers have been updated!

Desplandis:
@HoloTheDrunk One of my comment was not resolved

packages/Main/src/Core/MainLoop.ts
HoloTheDrunk: I distinctly remember making that change, not sure what happened...

Desplandis (approved):

Issue #2678, opened 2026-01-22
Picking for GLTF isn't working correctly


AnthonyGlt:
When using the picking on a gtlf, we don't get any result.
This is the case when the background color is left at the default value (#000000 <- black)

view.mainLoop.gfxEngine.renderer.setClearColor(

## Your Environment
branch master

## Context
When calling `view.pickObjectsAt(event, 0, gtlfObj);`, the return value is always empty.
Even though i'm clicking on the object.
It works if I set the background color to a different value, using `view.mainLoop.gfxEngine.renderer.setClearColor(...)`

## Steps to Reproduce
1. Don't call `view.mainLoop.gfxEngine.renderer.setClearColor()`
2. Import GLTF
3. Enable Picking on the gltf by click with a radius >= 0
4. Click and see that no value is returned

## Expected Behavior
Return the gltf object

## Actual Behavior
Nothing is returned

## Possible Cause/Fix/Solution
When calling `view.pickObjectsAt()` for a gltf with a radius >= 0, we go directly to Picking.js  `pickObjectsAt`
In this method, we render a circular zone to a buffer and reads the pixel colors back based on the 3d object (gltf). BUT it always returns RGB values closed to 0 and then we compare those values with the background color ` view.mainLoop.gfxEngine.renderer.getClearColor(clearColor);` which is, by default, black. So the picking area is considered as being part of the background because they are both black ...

The issue is that when we read the pixels colors values from the rendering, we shouldn't not get the color black. Currently, I think we get this wrong values because in the rendering, we are missing a light, so the scene is just black...

To correct this, we could either create a temporary scene containing only the object and a light and use it for the render.
Or just render the full scene and not only the object.

IMO the second option is the way to go

## Reprod Script
```
<html>
    <head>
        <title>Itowns - Instancing</title>

        <script type="importmap">
        {
            "imports": {
                "itowns": "../dist/itowns.js",
                "three": "https://unpkg.com/three@0.182.0/build/three.module.js",
                "three/addons/": "https://unpkg.com/three@0.182.0/examples/jsm/"
            }
        }
        </script>
        <meta charset="UTF-8" />
        <link rel="stylesheet" type="text/css" href="css/example.css" />

        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    </head>
    <body>
        <div id="description">
            <p>
                <b>
                    Render a large number of objects with the same geometry<br/>
                    by importing your 3D models
                </b>
            </p>
        </div>
[...]

PR #2679, opened 2026-01-22
fix(Picking): use full scene for rendering to buffer

AnthonyGlt:
## Description
Use the full scene for the rendering used in the picking

## Motivation and Context
Currently the rendering is done with only the selected object. So there is no light associated, then the rendering is completely black.

cf https://github.com/iTowns/itowns/issues/2678

The full scene includes the light. Then the `intersect` is still done with only the object

### DIFF
--- a/packages/Main/src/Core/Picking.js
+++ b/packages/Main/src/Core/Picking.js
@@ -234,7 +234,7 @@ export default {
             height: 1 + radius * 2,
         };
         const pixels = view.mainLoop.gfxEngine.renderViewToBuffer(
-            { scene: object, camera: view.camera },
+            { scene: view.scene, camera: view.camera },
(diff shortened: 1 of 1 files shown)


Desplandis (approved):
This is a temporary fix for #2678, we opened an issue to fix it permanently in itowns in #2682

gchoqueux (commented):
The depth buffer is used to pick the mesh with displacement map. In itowns there are other features to pick and filter meshes.
This fix may break the other pickings — have you checked?

AnthonyGlt:
@gchoqueux  I don't get what you mean sorry. Could you provide an example of what you mean ? 
Thank you

jailln:
Closed in favor of #2705

PR #2681, opened 2026-01-23
fix(examples): remove obsolete osm urls

### LINKED ISSUE #2664: Use preferred tile.openstreetmap.org URL
See https://github.com/openstreetmap/operations/issues/737

Use `https://tile.openstreetmap.org/` (Change from HTTP to HTTPS, no `{s}.`/`a.`/...) in all repositories for your organization.
Change regular `www.openstreetmap.org/`... URLs also preferably from HTTP to HTTPS.

Use HTTPS unless your application/browser/device/user does not support it.

jailln:
Closes #2664


Desplandis (approved):
Is there any use-case left for the `subdomains` scheme?

jailln:
@Desplandis  not sure, I think osm was the main usecase. I hesitated to remove it totally but I decided to leave it since it is a simple implementation that works.

